PR #173, opened 2017-12-03
Improve working with mmj2

jkingdon:
There's background at https://github.com/digama0/mmj2/issues/6#issuecomment-348714648 but the short summary is that adding some global variables and a dummy mathbox theorem will make mmj2 work better.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -14719,6 +14719,40 @@ $)
 
 $(
 #*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
+Constructive set theory
+#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
+$)
+
+  $( This is a very abbreviated version, enough to keep mmj2 happy. $)
+
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+                   Class abstractions (a.k.a. class builders)
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  $( Declare variable symbols that will be used to represent classes.  Note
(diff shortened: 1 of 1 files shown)


nmegill:
When you add a new symbol ($c or $v), you need to add an 
htmldef/althtmldef/latexdef for it in the special comment containing $t. 
  This can be checked with 'verify markup */f', which is useful to run 
before submitting.  Most or all of the htmldefs can be found in set.mm.

While I don't quite understand the mmj2 issues, I don't think you need 
more than 3 or 4 class variables at this stage (there is no set theory), 
meaning fewer htmldefs to add.  I'll leave that up to you.

I'll wait for you to fix the 'verify markup' problems before merging.

Norm

digama0:
The issue is that mmj2 doesn't like when variables are declared locally (like `wceq.cA`) but are never given a global definition. This doesn't happen in set.mm - all local variables are later declared globally. I think the only local class variables in set.mm (and iset.mm) are `A` and `B`, but there are quite a few local set variables like e,f,g.

jkingdon:
Dang! I'd been running `verify markup */f` reliably, until... well I forgot this time. Oh well, some time someone will put it into travis.

I'll fix that, and I guess while I'm at it I'll trim it down to the minimal variables needed (which I think is about four of them).

jkingdon:
The `verify markup */f` issue seems to be about the mathbox theorem, not the new variables. The error I am getting is:

```
MM> verify markup */f
?Warning: In the $t comment, extHtmlHome has no 'HREF="'.
?Warning: In the $t comment, extHtmlHome has no 'IMG SRC="'.
?Warning: In the $t comment, extHtmlHome has no 'HREF="'.
?Warning: In the $t comment, extHtmlHome has no 'IMG SRC="'.
?Warning: In the $t comment, extHtmlHome has no 'HREF="'.
?Warning: In the $t comment, extHtmlHome has no 'IMG SRC="'.
MM> 
```
If I remove the `mathbox` theorem, it goes away. Messing around with the details of the mathbox theorem and whether it is before or after the $t section didn't seem to help. So I'm a bit at a loss about what the real problem here is.

nmegill:
This seems like a bug in metamath.exe, maybe related to adding the 
mathbox theorem, which was designed for the set.mm structure.  I'll look 
at it soon.  In the meantime you can ignore these messages.

Mario, why is the mathbox theorem needed?

Norm

jkingdon:
Leaving off the mathbox theorem isn't particularly bad for mmj2, it just pops up a dialog but it continues to work. So I've revised this pull request to just have the variable changes.

jkingdon:
Thanks for fixing the bug. Good to know we can add the mathbox pseudo-theorem if mmj2 keeps that warning.

PR #175, opened 2017-12-04
remove dependencies on ax-8 in predicate logic

wlammen:


wlammen:
I added two new theorems wl-ax17e and wl-exiftru, that play a key role in removing unnecessary dependencies on ax-8. As a compensation, there are now a couple of orphaned sp* theorems either nowhere referenced or only in OLD proofs. WL

nmegill:
On 12/4/17 1:00 PM, David A. Wheeler wrote:
> *@david-a-wheeler* commented on this pull request.
> 
> Wait a moment, that seems to remove the entire discussion about the 
> indemonstratables. That is a key Axiom of stoic logic. Am I 
> misunderstanding this request?

Which theorem in set.mm are you concerned about?

Wolf is showing that certain theorems can be proved without invoking 
ax-8, whereas the existing proof did depend on it.  To me it is usually 
interesting to show that a theorem doesn't depend on a particular axiom 
(for example, eliminating unnecessary dependence on ax-ac).  This 
doesn't mean that ax-8 will go away, just that for certain theorems it 
isn't necessary.

Norm

nmegill:
Now I see you probably mean mpto2OLD.  Wolf shortened the proof of 
mpto2, which still has your original description but a shorter proof. 
Traditionally, we have reproduced the the theorem with the original 
proof and renamed it to *OLD, so the new and old proofs can be compared. 
  In the *OLD version, we usually replace the original comment with 
"Obsolete version of...." like Wolf has done in mpto2OLD.  We add "(New 
usage is discouraged.)" so the *OLD won't accidentally be picked up in 
in later proofs.

Periodically we (usually I) go through and purge *OLDs (with the 
"Obsolete" comment) that are more than a year old.

Norm

wlammen:
Hi David,

On the outside, you should see no difference, as theorems stay as they are. At most, and on rare occasions, you will find a reduced set of dependent axioms, or a reduced number of distinct variable conditions. I don't think this messes with your contributions and intentions.

WL

david-a-wheeler:
Yes, I meant mpto2OLD.  I saw that in the *changes* a lot of text was being removed.  However, you're right, it looks like mpto2 is kept, so I don't see a problem with this change!

Thanks for explaining this.

PR #181, opened 2017-12-08
rm wl prefix from two recently created theorems

wlammen:
I left the prefixes because I thought the name of the theorems undergo some renaming anyway. This seems not be the case.


wlammen:
This was a test for an alternative way of contributing: Using file upload instead of the git add - git commit - git push cycle. Unfortunately, GitHub imposes a size limit on these files (25MB), which is well exceeded by set.mm. So I think, we cannot replace the procedure in the 'Getting started with contribution' wiki entry with something simpler.

PR #188, opened 2017-12-10
Remove mt4, mt4d, and mt4i.

jkingdon:
They are classical results which are pretty straightforward consequences of
more fundamental theorems, and we don't use them anywhere.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -5889,31 +5889,6 @@ $( This section makes our first use of the third axiom of propositonal
     ( wi pm2.21 con1i ) AABCABDE $.
 
   ${
-    mt4.1 $e |- ph $.
-    mt4.2 $e |- ( -. ps -> -. ph ) $.
-    $( The rule of modus tollens.  (Contributed by Wolf Lammen,
-       12-May-2013.) $)
-    mt4 $p |- ps $=
-      ( con4i ax-mp ) ABCBADEF $.
-  $}
-
-  ${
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
I'm not as sure we want to remove these.  These theorems are obviously trivial, but in many cases we want to encourage the use of deductive form, because that lets us avoid having to use the harder-to-deal-with deduction theorem.

nmegill:
I'm not sure that these are valid intuitionistically (Jim would know 
better).  But if not, I think he is trimming down the classical section 
to include mainly classical theorems cross-referenced from the 
intuitionistic section comments, and there's no point duplicating all of 
set.mm in the classical section.

Norm

jkingdon:
Norm is correct, these are not valid intuitionistically (they are removing negation in a contrapositive sort of way). David, did you notice this was iset.mm not set.mm (I didn't make that as prominent as I probably should have).

nmegill:
I've been using the informal convention "file.name - description of 
change" for commits that involve a single file, e.g. "set.mm - shorter 
proof of abc".

Norm

david-a-wheeler:
@jkingdon : You're right, I didn't notice that this was for iset.mm, not for set.mm.  I thought it was getting removed from set.mm.  Never mind!  I'll double-check that from now on.

These aren't valid intuitionistically, so removing them makes sense!!

jkingdon:
In the future, I'll try to make it more prominent which file it is. Thanks everyone for the comments.

PR #192, opened 2017-12-15
Replace proof of ax11v with an intuitionistic one.

jkingdon:
- Move ax11v, sb56, sb5, and sb6 a bit later in the file (after we've
  proved a few theorems we need in this proof).

As for the idea of this proof, it is a pretty straightforward adaptation of the ax11v2 proof but uses distinct variable constraints directly, rather than distinctors (which don't work especially well intuitionistically, at least in situations like separating into distinct and non-distinct cases and then using case elimination).

Although this pull request is only for iset.mm, I'm wondering if set.mm would prefer this proof as well, because:

a) it avoids the odd theorem ax16

b) it proves ax11v directly from the current ax-11 axiom rather than from ax11o

Anyway, I don't know all the considerations in deciding whether to do that or not, but I'd be glad to make a pull request for it if it sounds like a good idea.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 9-Dec-2017
+$( iset.mm - Version of 15-Dec-2017
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm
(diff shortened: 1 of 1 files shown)


digama0:
I forget if we added the axiom of LEM for distinctors. There was a discussion a while back about this where it was argued that this is reasonable assuming that the underlying set of variables in the model has decidable equality. It looks like this:

    |- ( A. x x = y \/ -. A. x x = y )

jkingdon:
I'm no logician, but decidable equality for all sets seems like a bit much to assume (as far as I can tell from a little web searching and looking through the HoTT book, not a deep understanding).

digama0:
Not decidable equality for the universe, decidable equality for the set of variables. Usually, the set of variables is the natural numbers or something similar, so decidable equality is intuitionistically provable in the metalanguage. In standard logic, every instance of |- ( A. x x = y \/ F/_ x y ) is intuitionistically provable, because it will either be |- ( A. vn vn = vn \/ F/_ vn vn ) where the first disjunct is provable, or |- ( A. vn vn = vm \/ F/_ vn vm ) for m != n, and then the second disjunct is provable.

nmegill:
On 12/15/17 4:00 AM, Jim Kingdon wrote:
>   * Move ax11v, sb56, sb5, and sb6 a bit later in the file (after we've
>     proved a few theorems we need in this proof).
> 
> As for the idea of this proof, it is a pretty straightforward adaptation 
> of the ax11v2 proof but uses distinct variable constraints directly, 
> rather than distinctors (which don't work especially well 
> intuitionistically, at least in situations like separating into distinct 
> and non-distinct cases and then using case elimination).
> 
> Although this pull request is only for iset.mm, I'm wondering if set.mm 
> would prefer this proof as well, because:
> 
> a) it avoids the odd theorem ax16
> 
> b) it proves ax11v directly from the current ax-11 axiom rather than 
> from ax11o
> 
> Anyway, I don't know all the considerations in deciding whether to do 
> that or not, but I'd be glad to make a pull request for it if it sounds 
> like a good idea.

The current ax11v proof is somewhat shorter, which is one of our 
criteria.  However, your proof is interesting, and I propose to add it 
as ax11vALT, which is a convention we typically use for alternate proofs.

I add this one myself (with your name attached) since I want the 
description to be worded in a particular way, which you can use as an 
example for future pull requests.

Norm

nmegill:
Hi Mario,

In this email the backslashes in the 'or' symbol were deleted (oddly, 
just in the email but not on the github site).  But in your previous 
email the backslashes came through.  Do you recall what you did 
differently, since I've had the same problem?

Norm

nmegill:
On 12/15/17 12:40 PM, N Megill (please reply to nm@alum.mit.edu) wrote:

> I add this one myself (with your name attached) since I want the
   ^^^^^
I will add

jkingdon:
Having you add it to set.mm as ax11vALT sounds good. Thanks for doing that.

Feel free to speak up about description wording. I have my own desires (like wanting to link to related theorems) but if I'm not following existing conventions it is probably mostly due to ignorance.

david-a-wheeler:
On Fri, 15 Dec 2017 10:38:27 -0800, Jim Kingdon <notifications@github.com> wrote:
> ... if I'm not following existing conventions it is probably mostly due to ignorance.

I tried to capture at least some of the important set.mm conventions here:
  http://us.metamath.org/mpeuni/conventions.html

Please let us know about any important omissions in the documentation on conventions.
It's often hard to notice what *should* be there but isn't.

--- David A. Wheeler

nmegill:
On 12/15/17 6:01 PM, David A. Wheeler wrote:
> On Fri, 15 Dec 2017 10:38:27 -0800, Jim Kingdon 
> <notifications@github.com> wrote:
>  > ... if I'm not following existing conventions it is probably mostly 
> due to ignorance.

David's advice is useful, but overall I think you are doing fine.  The 
reason I wanted to apply your ax11v to set.mm myself was that we 
typically use "discouraged" tags in ALT versions of an existing theorem 
to help ensure consistent usage of the main theorem and also to prevent 
the proof of the ALT version from accidentally being "minimize"d to 1 
step that references the main version.  Since these are not used in 
iset.mm, I assumed you are probably unfamiliar with them.

I updated 
https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md to 
document these tags and how to regenerate the "discouraged" file when 
these tags are used, so that you will know how to do it in the future.


> 
> I tried to capture at least some of the important set.mm conventions here:
> http://us.metamath.org/mpeuni/conventions.html
> 
> Please let us know about any important omissions in the documentation on 
> conventions.
> It's often hard to notice what *should* be there but isn't.
> 
> --- David A. Wheeler

PR #199, opened 2017-12-18
Add sbnv to iset.mm

jkingdon:
This is an intuitionistically valid proof, but requires that x and y be distinct.

The thought process here is that we'll be able to prove the full sbn from this one by introducing an intermediate variable. But we don't have enough theorems concerning substitution composition (sbco2 or similar) to do that yet. So if you'd like to hold off on merging sbnv until we figure out whether this can work, I'm OK with that. But I thought I should at least show what I have so far.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 16-Dec-2017
+$( iset.mm - Version of 18-Dec-2017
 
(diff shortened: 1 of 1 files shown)


nmegill:
I see no problem with putting it in now.  You can always change it later.

jkingdon:
Good point about deleting it later if it turns out to be a dead end. Sounds good.

PR #200, opened 2017-12-19
Full reorganization of my mathbox using subsections, following Gerard Lang's advice

tirix:
This pull request includes a major reorganization of my mathbox, with two layers of sections/subsections, like in Alexanders, following Gerard Lang's advice.

It also includes an additional axiom redefining ` CCfld ` to add its "Uniform Structure" component (~ df-cnfldNEW). This allows to prove that the resulting structure is a Complete Uniform Space (~ cnfldcusp), as is its restriction to the real numbers (~ recusp).
However that last additional change in comments fails the Travis check: the consistency check detects that I'm redefining a symbol already used, so I had to put it in comment. Would you accept to change the official ` CCfld ` definition? How would we proceed? (I would likely do the same change for ` RR*s ` too)

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 16-Dec-2017
+$( iset.mm - Version of 18-Dec-2017
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm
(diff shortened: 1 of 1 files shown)


nmegill:
There is a problem in that Wolf Lammen's recent shortened proof was reverted.  Can you fix this and resubmit?

As for CCfld, I haven't followed the discussions that closely.  Mario or Stefan should probably comment on this.  But if you are revising the CCfld definition, then it has to be done in the main part of set.mm and all of the old theorems that are affected need to be revised.  Any new CCfld theorems can remain in your mathbox, or if they seem reasonably general purpose, you may as well move them to the main part of set.mm.  (All this can be done in a future submission if you keep df-cnfldNEW commented out for now.)

tirix:
I've taken care of Wolf Lammen's shortened proof in the last commit, it shall be back.

For the other theorems on which CCfld depend, the commit also includes proofs for them (cnfldstrNEW, cnfldbasNEW, cnfldaddNEW, cnfldmulNEW, cnfldcjNEW, cnfldtsetNEW, cnfldleNEW, cnflddsNEW).

When updating the definition of CCfld in the main part of set.mm, df-metu will also have to be moved, since that new definition uses it. That could be the only one, if you prefer to limit those additions.

PR #201, opened 2017-12-19
Add sban-1 (forward direction of sban).

jkingdon:
Although the idea in this proof (just expanding df-sb, moving things around, and then applying df-sb again) doesn't work for the reverse direction, having the forward direction is progress.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 18-Dec-2017
+$( iset.mm - Version of 19-Dec-2017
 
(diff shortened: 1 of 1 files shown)


nmegill:
By the way I'm not sure if you are using or are aware of the 'expand' command that was added relatively recently to metamath.exe.  For some things like this it can save time.

jkingdon:
Thanks for the tip on `expand`. I'm still learning basic mmj2 usage (haven't done as much with the metamath.exe proof assistant yet), so I'm probably a little early on the learning curve in terms of knowing how to take advantage of it, but I looked at the description and can see how it might be useful.

PR #203, opened 2017-12-21
Updating the definition of ` CCfld ` with its Uniform Structure

tirix:
This is what the changes of the ` CCfld ` structure would look like in the main `set.mm`, so that the Uniform Structure component is added.

- There is one new (discouraged) usage of ~ df-cnfld, it's required to access the new component,
- In addition to the definition of ` metUnif `, I actually also had to pull the definitions of ` fBas ` and ` filGen ` earlier in set.mm
- I've not updated the structure product `Xs_` and the strucutre image `"s`, this could be done in a second step.
- Some reformatting also appears in the commit, it's not related - sorry it makes reviewing confusing.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4220,6 +4220,7 @@
 "df-cnfld" is used by "cnfldmul".
 "df-cnfld" is used by "cnfldstr".
 "df-cnfld" is used by "cnfldtset".
+"df-cnfld" is used by "cnfldunif".
 "df-cnfn" is used by "elcnfn".
 "df-cnfn" is used by "hhcnf".
 "df-cnop" is used by "elcnop".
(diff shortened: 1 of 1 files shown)


nmegill:
This seems OK to me.  I'll wait 24 hrs for any objection (Mario? 
Stefan?) before merging.  Meanwhile you can add any new changes you have 
to this pull request, if you want.

Norm

digama0:
LGTM. Although this seems a little silly without a definition of uniform spaces in main. You should prepare a PR for moving uniform spaces up, and then we can start retrofitting MetSp and TopGrp to work with uniformities.

tirix:
Full section on uniform spaces is now in main, since pull request #297!

PR #218, opened 2017-12-29
shortening cbv3hv & nfbid

wlammen:


wlammen:
I moved some misplaced theorems to what I think are their proper places. Some descriptions in the nf* range did not match the theorems; fixed this

PR #219, opened 2017-12-29
Replace a4sbe proof with intuitionistic one.

jkingdon:
Also, cite this theorem in Margaris.


jkingdon:
This reduces the `show usage ax-3/recursive` count by one.

PR #221, opened 2017-12-30
 Graph Theory: Vertext Degree, Hypergraphs, 1. Claim in Proof of Huneke

avekens:
Additional definitions and theorema for Graph Theory in the mathbox of Alexander van der Vekens: alternative definition of the Vertext Degree, definition of Hypergraphs, 1. claim in proof of the friendship theorem by Huneke.

**Note**: Theorem ~ rabfi  from Mathbox of Thierry Arnoux is used (by ~ vdgrefinf ). This theorem should be moved to the main part of set.mm.


nmegill:
I moved rabni to the main part of set.mm as requested.
Norm

PR #223, opened 2017-12-30
shorten 19.9 & 19.9t

wlammen:


wlammen:
shortening nfimd is a trade off: Proving 19.21t from elementary theorems increased it by 18 bytes, but nfimd shrank in contrast by 88 bytes.

The idea of deriving A. x (ph->ps) from ax-1, pm2.21 and jad and some other preconditions is found at least 4 times, always leading to long proofs. I decided on 19.21t to take the burden, and let the others depend on it with much shorter proofs. nfimd is an example. Others follow, once the nf* block is moved closer to sp.

PR #225, opened 2017-12-31
What is wrong with iset.mm and this proof?

jkingdon:
Here's a proof of a theorem called `sbrequv-trouble`, which passes `VERIFY PROOF *` and which reads, in its entirety, `x = y`. I take this to mean there is only one object in the universe, which is a contradiction (rather, would be if combined with set theory axioms) or at least a problem.

Am I confused about what this theorem means? Is there a problem with the iset.mm axioms? Is there a bug in the proof verifier? I suppose I'd guess the axioms (they aren't from a published source, are they?) but I'm not sure I have any especially concrete ideas about which ones to dig into and how. As for the verifier, I also tried smetamath.

Unless someone else can just see what is going on here, I suppose my next step might be seeing if set.mm can do the same thing.

(I wasn't sure whether to make this a pull request or an issue or what, but I'm not necessarily suggesting this be merged as-is).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -12508,6 +12508,25 @@ $)
   $}
 
   ${
+    $d x y z $.  $d ph x $.
+
+    $( Fishy-looking theorem.  (Contributed by Jim Kingdon, 30-Dec-2017.) $)
+    sbrequv-paradox $p |- ( [ z / y ] ph <-> ph ) $=
+      ( vx cv wsbc ax-17 sbrequv sbie sbf sbbii bitr3i bitri ) ABCEZFZA
+      DBEZFZAOADNFZDPFQRODBODGADBCHIRADBADCADGZJKLADBSJM $.
+  $}
+
+  ${
+    $d x y $.
+    $( There is only one object in the universe.  Not completely sure we have
+       the ability to turn ~ sbrequv-paradox an outright contradiction without
+       set theory axioms (or perhaps we do), but this should be enough to
+       illustrate a problem.  (Contributed by Jim Kingdon, 30-Dec-2017.) $)
+    sbrequv-trouble $p |- x = y $=
+      ( cv wceq wsbc equsb2 sbrequv-paradox mpbi ) ACZBCDZBIEJBAFJBAGH $.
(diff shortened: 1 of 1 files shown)


digama0:
You're right, this is a flaw in the axioms, specifically `ax-i11e`. When `x` and `y` are distinct, the statement says:

    ⊢ x = y → (∃ y', ph(x, y') → ∃ x' (x' = y ∧ ph(x', y))

which amounts to asserting `E. y ph(x, y) -> ph(x, x)` which is false in general.

digama0:
I think the correct fix is to reverse the arrow:

    ax-i11e $a |- ( x = y -> ( E. x ( x = y /\ ph ) -> E. y ph ) ) $.

digama0:
(As an aside, we should probably derive all the intuitionistic axioms in set.mm, so that this doesn't happen again. That will give confidence that set.mm is at least equiconsistent with iset.mm.)

jkingdon:
OK, I'll try reversing the arrow and seeing what changes.

I had gotten as far as noticing it was reversed, but not as far as working out the consequences of changing that. Thanks for taking a look at this.

And, yes, deriving the new axioms makes sense (I guess in set.mm, or perhaps in iset.mm plus ax-3 and being careful not to use the new axioms, although with the latter approach it is probably easier to get confused about what axioms we are and are not using at any given moment).

```
  exmid-ax11e $p |- ( x = y -> ( E. x ( x = y /\ ph ) -> E. y ph ) ) $=
    ( cv wex wa wn wal wi ax-11 alnex imnan albii bitri imbi12i sylib
    wceq con4d ) BDCDQZACEZSAFZBEZSAGZCHZSUCIZBHZITGZUBGZIUCBCJUDUGUF
    UHACKUFUAGZBHUHUEUIBSALMUABKNOPR $.
```

jkingdon:
I've looked over the proof of sbrequiv as found on this branch, and it indeed falls apart with the implication in the correct direction in `ax-i11e`.

So I'm going to consider this issue resolved by #226 and #227 .

PR #226, opened 2017-12-31
Prove intuitionistic axioms.

jkingdon:
This was suggested in a comment to #225 and is a way of reassuring ourselves that we don't have additional problems in the iset.mm axioms which render iset.mm inconsistent.

The formatting changes outside the new section are just running `write source set.mm/rewrap`.


set.mm
```diff
@@ -29066,6 +29066,84 @@ existence hypothesis (to prove the "some" in the conclusion).  Example
...
+     to be stated using the existential quantifier.  (Contributed by Jim
+     Kingdon, 31-Dec-2017.) $)
+  axi9 $p |- E. x x = y $=
+    ( a9e ) ABC $.
+
+  $( Axiom of Quantifier Substitution (intuitionistic logic axiom ax-10).  This
+     is just ~ ax10 by another name.  (Contributed by Jim Kingdon,
+     31-Dec-2017.) $)
+  axi10 $p |- ( A. x x = y -> A. y y = x ) $=
+    ( ax10 ) ABC $.
```
digama0: I think this should be `ax-10` not `ax10`.

nmegill: Actually since ax-10 is redundant, we have deprecated its use and have a "New usage is discouraged" tag.   ax10 (which someday will probably be named something more meaningful) is the one to use, being derived from the official axioms.

Norm

nmegill: I meant, of course, classically redundant.  I have no idea if it can be derived from the others intuitionistically, so for iset.mm we still should keep ax-i10 unless and until it can be shown redundant there.

Norm

PR #234, opened 2018-01-03
shorten hbimd further

wlammen:


wlammen:
These are not my fault!
===================================

MM> verify markup *
?Warning: Line number 23096 has 87 characters (should have 79 or less):
      $( Axiom of Left Equality for Binary Predicate.  One of the equality a...
?Warning: Line number 23099 has 81 characters (should have 79 or less):
         arbitrary binary predicate ` e. ` , which we will use for the set m...
?Warning: Line number 23100 has 88 characters (should have 79 or less):
         relation when set theory is introduced.  This axiom scheme is a sub...
?Warning: Line number 23124 has 88 characters (should have 79 or less):
      $( Axiom of Right Equality for Binary Predicate.  One of the equality ...
?Warning: Line number 23127 has 81 characters (should have 79 or less):
         arbitrary binary predicate ` e. ` , which we will use for the set m...
?Warning: Line number 23128 has 88 characters (should have 79 or less):
         relation when set theory is introduced.  This axiom scheme is a sub...

nmegill:
It is my fault.  I transferred the wrong version of set.mm for my update this morning.  I'll fix it.
Norm

PR #235, opened 2018-01-03
shortening 19.12

wlammen:
A 2nd order shortening because the proof in compressed mode is exactly the same, but the HTML display uses less symbols.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12968,6 +12968,7 @@ New usage of "0psubclN" is discouraged (1 uses).
 New usage of "0r" is discouraged (11 uses).
 New usage of "0reALT" is discouraged (0 uses).
 New usage of "0vfval" is discouraged (9 uses).
+New usage of "19.12OLD" is discouraged (0 uses).
 New usage of "19.21a3con13vVD" is discouraged (0 uses).
 New usage of "19.21hOLD" is discouraged (0 uses).
 New usage of "19.21tOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
Using the rewrap procedure is really a pain. It affects lines I have not touched. And does not only correct indentations, but changes proofs (-styles) as well. Example:

@@ -24135,8 +24135,7 @@ $)
     $( If ` x ` is not free in ` ph ` and ` ps ` , then it is not free in
        ` ( ph -/\ ps ) ` .  (Contributed by Scott Fenton, 2-Jan-2018.) $)
     nfnan $p |- F/ x ( ph -/\ ps ) $=
-      wph wps wnan wph wps wa wn vx wph wps df-nan wph wps wa vx wph wps vx
-      nfan.1 nfan.2 nfan nfn nfxfr $.
+      ( wnan wa wn df-nan nfan nfn nfxfr ) ABFABGHCABIABGCABCDEJKL $.
 
So I had to do a partial revert just keeping my lines tidy. Which is really a mess, because I needed all my git know-how to finally get the fork on github and my clone synchronized.

nmegill:
The rewrap changes you saw is because Scott didn't rewrap his changes.  Also, our convention for set.mm is to use only compressed proofs.

Rewrapping is important to maintain a consistent style throughout set.mm.  But there is no requirement that you must do a rewrap.  It is just a suggestion that may help prevent future merge conflicts, because a rewrap will eventually be done by someone and may change an unwrapped section you are still working on.  If you don't plan to continue working on a section you modified, then you don't have to do a rewrap.

I considered making automatic wrapping a part of the Travis build.  but sometimes (such as making global label change) it will cause hundreds of proof lines to change, obscuring the "real" change that you could otherwise make into a separate commit.

Norm

sctfn:
FWIW, I try to to a rewrap each time, but I'd gotten lost in the new /split
work (that theorem was to support things over in nf.mm).

PR #237, opened 2018-01-04
Move modal-b and ax9 to classical logic section.

jkingdon:
I keep going back and forth about how to deal with all the predicate logic axioms (or former axioms) in iset.mm. But I guess for the most part I'm continuing to move the proofs which depend on ax-3 to the end of the file. This pull request moves one more, ax9 (and modal-b, which at least in terms of its current usages can be considered a lemma for ax9).

We can't move ax11o until we figure out substitution, but I haven't made any new progress on that in a while, so that'll have to stay where it is for the moment.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -10029,11 +10029,6 @@ $)
   modal-5 $p |- ( -. A. x -. ph -> A. x -. A. x -. ph ) $=
     ( wn hbn1 ) ACBD $.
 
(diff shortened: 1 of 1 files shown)


nmegill:
Hi Jim,
What is your email address?
Norm

jkingdon:
Email address is Jim Kingdon <kingdon@panix.com>.

PR #246, opened 2018-01-08
rm dependencies from excom

wlammen:
The current proof of excom is short, but depends on a lot of axioms. In fact, the straightforward proof uses just ax-gen, ax-5 and ax-7.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14923,6 +14923,8 @@ New usage of "exatleN" is discouraged (1 uses).
 New usage of "exbir" is discouraged (0 uses).
 New usage of "exbirVD" is discouraged (0 uses).
 New usage of "exbiriVD" is discouraged (0 uses).
+New usage of "excomOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
Regarding "metamath does not like commas in Revised... annotations":

The "Revised by" markup tag format was intended to be "(Revised by &lt;author>, &lt;date>.)"  While the &lt;author> is not enforced by the program (e.g. by matching the contributor list), it does check to see that there is one comma and (IIRC) that the date has the right format.

In the past, I've used "(Revised by Wolf Lammen, 8-Jan-2018.)  Removed dependency on ~ ax-11 , ~ ax-6 , ~ ax-9 ,   ~ ax-8 , and ~ ax-17 ." instead of "(Revised by Wolf Lammen to remove dependency on ~ ax-11 ~ ax-6 ~ ax-9  ~ ax-8 and ~ ax-17 , 8-Jan-2018.)"

Norm

PR #249, opened 2018-01-11
nf.mm - reworking for split files

sctfn:
This merge contains my work to bring nf.mm up to date with split files. As such, it also changes the hb* theorems to nf* ones. 

At the moment, it inlines set-pred.mm instead of leaving it as an include. I'm not sure about this choice. What do you think?


nmegill:
I think nf.mm should be left in-line for the time being like  set.mm, since it is the simplest thing to do and it is what we will be doing with set.mm initially.  Also, the new metamath.exe feature is still experimental.  We can occasionally update set-pred.mm from set.mm since the set.mm version changes over time whereas nf.mm has low activity (although I'm not sure what your plans are for nf.mm).  It may be worthwhile at some point to update the Travis script to do basic error checks on nf.mm.

sctfn:
Oh gods, please don't turn on Travis checks yet. I've got 800+ "?who?"
entries lying around to clean up :-)

nmegill:
There's no urgency to do this; it was just a suggestion for the longer term.  Also, it could start with just a basic 'verify proof *' and nothing else.

Norm

sctfn:
Actually, I just figured it out. I got lazy/clever and set all the
contributors to "set.mm contributor" when it wasn't an NF specific theorem.
I feel this is a fair solution since I want to emphasize the common parts
of the development anyway.

PR #257, opened 2018-01-13
prove sp via 19.8a

wlammen:
Don't know whether you like this one. It puts sp second in row after 19.8a, which is easier to prove from ax-11.


wlammen:
Hold on

PR #261, opened 2018-01-15
Don't try to list axioms in section headers (iset.mm).

jkingdon:
These headers might still be accurate for set.mm but not for
iset.mm.  The most obvious issue is that the Axiom of Existence
is ax-i9 in iset.mm, but the header about "ax-8 through ax-14
except ax-9" isn't an accurate list either.


jkingdon:
Sorry for the close/reopen, that was just fat fingering on my part. This is ready to be merged if people think it is a good idea.

jkingdon:
@nmegill This is another pull request which apparently slipped through the cracks.

PR #264, opened 2018-01-16
implication distributes over disjunction

jkingdon:
imordi is a theorem of propositional logic which seems interesting on its own merits, although it is really in intuitionistic logic (in which it isn't just enough to convert all the disjunctions to implication and negation and work that way) that it would seem most worthy of full investigation.

I did find one reference to this theorem: https://proofwiki.org/wiki/Implication_is_Left_Distributive_over_Disjunction/Formulation_2 which cites Theorem T55 of Donald Kalish and Richard Montague (1964), Logic: Techniques of Formal Reasoning. I've ordered a copy of that book (turns out it is much cheaper than the 1980 edition), but of course I won't have it for a while.

There's also a minor cleanup to sborr. Wondering if that proof can be biconditionalized (especially in light of 19.43 being a biconditional) is what got me onto the topic of disjunctions in the first place, but this part of the investigation didn't get very far.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -11743,11 +11743,11 @@ $)
   $( Reverse direction of ~ sbor .  (Contributed by Jim Kingdon,
      20-Dec-2017.) $)
   sborr $p |- ( ( [ y / x ] ph \/ [ y / x ] ps ) -> [ y / x ] ( ph \/ ps ) ) $=
-    ( wsb wo weq wi wex df-sb biimpi orim12i orc imim2i anim1i andi exbii bitri
-    wa olc 19.43 anbi2i sylibr syl ) ACDEZBCDEZFCDGZAHZUGASZCIZSZUGBHZUGBSZCIZS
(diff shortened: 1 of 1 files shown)


wlammen:
imordi is (more or less) a duplicate of pm4.78

jkingdon:
Ah, good catch on pm4.78. I was looking in roughly that part of the file, but didn't find it.

I'll revise the pull request (there's still a little work to do at least in iset.mm) but there is no need for a new theorem which duplicates one we already have.

nmegill:
I didn't see this message before I merged, so your revision would need a new pull request.

BTW to search for an expression in set.mm with metamath.exe, not knowing which is ph and which is ps, you can use:

MM> search * '( ?? -> ?? ) \\/ ( ?? -> ?? )'

nmegill:
"I guess next time I make a comment saying "I plan on revising this", I'll close the pull request. That might be the best way to express "don't merge this yet" in github. (Not that I'm too worried about a followup pull request; that's now at #267 )."

Hi Jim,

Don't worry about it; I was just careless.  Usually I do check all notices for a particular pull request number before merging it.

PR #268, opened 2018-01-18
Move/remove classical theorems in iset.mm

jkingdon:
Move excluded middle and related theorems to classical logic section.

Remove `impt` from iset.mm. `-.` and `->` do not form a set of primitive connectives in
intuitionistic logic. And we already have `pm3.31`.

Remove unused classical theorems from iset.mm. Specifically `mt3i`, `pm2.61ddan`, `pm2.61dda`, `pm2.61nii`, and `pm2.61iii`. They don't do a better job of contrasting with intuitionistic logic than, say, `pm2.61` itself.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -6200,24 +6200,6 @@ $( Although the Intuitionistic Logic Explorer is primarily for theorems which
       ( wi wn wo imor mpbi ) ABDAEBFCABGH $.
   $}
 
-  $( Law of excluded middle, also called the principle of _tertium non datur_.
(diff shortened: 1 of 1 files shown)


jkingdon:
@nmegill Have you had a chance to look at this?

Not that my future development especially depends on it, just that the more I can dispose of the easy cases, the easier it is to see what work remains.

nmegill:
Sorry, I had overlooked it.  It is now merged.
Norm

PR #269, opened 2018-01-18
Develop -  graph theory in Mathbox of Alexander van der Vekens

avekens:
limit for number of edges in finite graphs, rework/enhancement glossary, generalization of auxiliary theorems.

@Norman: Please move the following theorems currently in the Mathbox of Thierry Arnoux into the main part of set.mm, because I use them in the proofs of my theorems: 
   + prelpwi, used in cusgraexi
   + hashgt0, used in hashgt0elexb


nmegill:
OK, I will move the theorems you requested.
Norm

PR #273, opened 2018-01-20
Develop

wlammen:


wlammen:
still a problem with distinct variable conditions...

PR #283, opened 2018-01-23
Develop - Moving graph theory into main part of set.mm (5)

avekens:
Moving definitions of Walks, Trails, Paths, Circuits, Cycles, Connected graphs and corresponding theorems from Alexander's Mathboxes into the main part for graph theory within set.mm. Also examples for undirected simple graphs.

Remarks:
- By this pull request, moving graph theory to the main part of set.mm is finished (according to the friendship theorem, I'll continue wiorking in my Mathbox)
- The alignment of Eulerian Path with the general definition of a trail is still to be done (see ~ eupatrl). However, I think this is not so important/urgent.
- I moved ~ ftp from SO's mathbox to main set.mm
- Again, git has massive  problems to show the differences (although it is only a big block which I moved...)


nmegill:
You need to reissue the pull request against develop, not master.

Norm

avekens:
OK, sorry - but the github webpage often switches back to master automatically...

PR #285, opened 2018-01-25
Additions related to uniform spaces and ordered fields.

tirix:
This is for merging recents changes, only in my mathbox in set.mm (plus regeneration of "discouraged" to fix the travis check which failed on develop).
The diff seems to be huge, I guess it's not due to changes on the Proba branch, but to one single commit on the develop branch: 356d6da1ed06285374b0b2677714cae2358f124a

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16706,6 +16706,8 @@ New usage of "pl42lem3N" is discouraged (1 uses).
 New usage of "pl42lem4N" is discouraged (1 uses).
 New usage of "plpv" is discouraged (1 uses).
 New usage of "ply1rclOLD" is discouraged (0 uses).
+New usage of "pm11.07" is discouraged (0 uses).
+New usage of "pm11.07OLD" is discouraged (0 uses).
 New usage of "pm110.643ALT" is discouraged (0 uses).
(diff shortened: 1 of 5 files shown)


nmegill:
Something went very wrong here, I don't understand what.  In particular, you shouldn't be making changes to iset.mm and nf.mm.

nmegill:
In iset.mm, your pull request claims to make the following change:

-$( iset.mm - Version of 9-Jan-2018
+$( iset.mm - Version of 21-Jan-2018

Yet the latest iset.mm ca2449f already has 21-Jan-2018.  It seems like it is trying to update an old version of iset.mm, which makes no sense to me.

I hope someone can figure out what went wrong, but if not, email me your mathbox and $t section as attachments (if those are all you changed) and I'll update set.mm by hand.

Norm

tirix:
Hi Norm, indeed, it looks like I did not merge properly. This now looks much better. Could you please have another look?

nmegill:
It looks good.

The annual 'minimize' run, which I'll probably apply to set.mm tomorrow, will affect your mathbox; so after that you probably want to update your proba branch to prevent future merge conflicts.

tirix:
Sure. Thank you!

PR #286, opened 2018-01-25
Copy many F/ theorems from set.mm to iset.mm.

jkingdon:
This is mostly verbatim copy-paste, with a few adjustments to some
theorems.

Adds one to the show usage ax-3/recursive count, which is nf4, and
which seems worth adding (to the classical section) to illustrate
that it is a classical result.

As for the larger picture, basically twofold. One is just a general desire to have iset.mm catch up to what has been going on in set.mm since they forked. But the other is that I'm thinking about how to prove hbsb (without the extra distinct variable constraint in http://us2.metamath.org:88/ileuni/hbsbv.html ), and it may involve more closed form usages of freeness, which seem more natural to me with `F/`.

At the risk of describing something which isn't fully formed, the techniques I've been using to prove substitution theorems for iset.mm may be running out of steam, and I think the next step may be to get more serious about ax-i12. I'm now starting to catch up to what Mario Carneiro was saying in #192 about whether distinctness is decidable (my wording) or whether the "underlying set of variables in the model has decidable equality" (his wording from the discussion there). Seems like ax-i12 is the form of decidability (or something akin to it) which we have been assuming so far, and in things like the http://us2.metamath.org:88/ileuni/equvini.html proof, it plays a similar role to the "consider the distinct and non-distinct cases and then use case elimination" type proofs in set.mm.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -9091,6 +9091,75 @@ $)
       ( wal ex al2imi imp ) ADFBDFCDFABCDABCEGHI $.
   $}
 
+  $c F/ $.  $( The not-free symbol. $)
+
+  $( Extend wff definition to include the not-free predicate. $)
+  wnf $a wff F/ x ph $.
+
+  $( Define the not-free predicate for wffs.  This is read " ` x ` is not free
+     in ` ph ` ".  Not-free means that the value of ` x ` cannot affect the
+     value of ` ph ` , e.g., any occurrence of ` x ` in ` ph ` is effectively
+     bound by a "for all" or something that expands to one (such as "there
+     exists").  In particular, substitution for a variable not free in a wff
+     does not affect its value ( ~ sbf ).  An example of where this is used is
+     ~ stdpc5 .  See ~ nf2 for an alternative definition which does not involve
+     nested quantifiers on the same variable.
+
+     Not-free is a commonly used constraint, so it is useful to have a notation
+     for it.  Surprisingly, there is no common formal notation for it, so here
+     we devise one.  Our definition lets us work with the not-free notion
+     within the logic itself rather than as a metalogical side condition.
+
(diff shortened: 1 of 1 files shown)


nmegill:
Jim said: "Seems like ax-i12 is the form of decidability (or something akin to it) which we have been assuming so far, and in things like the http://us2.metamath.org:88/ileuni/equvini.html proof, it plays a similar role to the "consider the distinct and non-distinct cases and then use case elimination" type proofs in set.mm."

We have been using the term "bundling" to describe the merging of distinct and non-distinct cases.  See:
http://us.metamath.org/mpeuni/mmtheorems16.html#mm1545b
http://us.metamath.org/mpeuni/mmtheorems18.html#mm1734b
http://us.metamath.org/mpeuni/mmset.html#axiomnote
as well as Google Group messages discussing it.
In set.mm we often achieve bundling using pm2.61i (which is not intuitionistically valid, I don't think) to combine cases with antecedents A. x x = y and -. A. x x = y resp.  Theorems dvelim and friends assist the conversion of $d to -. A x x = y.  Theorems dr* (dral1, dral2,...) assist construction of the A. x x = y part.  I have wondered what techniques might be available for the iset.mm system and whether it is what we call "metalogically complete" i.e. whether we can always eliminate unnecessary $d's.

Norm

jkingdon:
Norm's summary of bundling matches my understanding. As for how intuitionistic logic affects this, it is correct that we don't have pm2.61i but instead just have things like ecased.

That's why changing `ax-12` to `ax-i12` is a big deal (in the intuitionistic context). It is making a fairly fundamental assumption about what a variable is and what models are possible. Assuming as an axiom `A. x x = y \/ -. A. x x = y` would be another choice. In iset.mm, we have not made the latter choice but we have adopted `ax-i12`.

PR #291, opened 2018-01-28
Cleanup following minimize

tirix:
- recovered 2 ~ALT proofs which were removed by minimize
- removed 3 theorems having a one-step proof, replaced them by the originals
- removed theorems already moved into the main section
- regen discouraged (several new "proof modification discouraged" were added)

_Note:_ There is another ~ALT proof, cleljustALTNEW7, which was probably unduly minimized. That one is in NM mathbox. I suppose it shall be recovered too.


nmegill:
Thanks for reviewing these and fixing them.  One of my hopes with the "discouraged" markup was to avoid having to do this review after "minimize"; we're almost there but not quite yet.

Norm

PR #292, opened 2018-01-28
eliminate ax12olem3

wlammen:
and remove dependency of ax12olem4 on ax-6


wlammen:
There is something wrong with the discouraged file, but I haven't done anything needing an update of his file

nmegill:
The discouraged is my fault, sorry.  I'll fix it.
Norm

PR #294, opened 2018-01-29
More F/ in iset.mm: nfand ax16nf 19.9t and friends

jkingdon:
So the thing I'm really working on (substitution theorems) is going kind of slowly, so I thought it would be a good idea to send in the F/ related changes which I've made in the last few days.

I think all of this is just catching up to set.mm except for ax16nf which set.mm doesn't seem to yet have (as far as I noticed, anyway). The real motivation for ax16nf (how I want to use it, that is) is off on the other branch ( which is at https://github.com/jkingdon/set.mm/tree/sbco2rv plus some `.mmp` files sitting on my laptop's hard drive, and subject to change before it is ready for a pull request).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -12503,6 +12503,15 @@ $)
   $}
 
   ${
+    $d x y $.
+    $( Theorem of Distinct Variables ( ` F/ ` variant).  Like ~ ax16 , but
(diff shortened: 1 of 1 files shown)


jkingdon:
@nmegill Did you see this one? I'm fairly close to submitting some further work which depends on it and #298 so if you get the chance to look at them before I do, things may be a little less confusing.

PR #295, opened 2018-01-29
Added proof on uniform continuity, and RR canonical embedding.

tirix:
This should complete the chapter on uniform structures / uniform spaces / uniform continuity and Cauchy filters: its content allowed me to complete proofs on extension by continuity, and on the canonical embedding into RR (which is the extension by continuity of the canonical embedding into QQ).
Next, I would like to propose to move it to the main part of set.mm as suggested by Mario.


nmegill:
Please go ahead and move it to the main set.mm.
Norm

PR #296, opened 2018-01-30
provide an alternative, shortened proof of ax12o

wlammen:
a replacement for ax12o, ax12olem1 - ax12olem7, dvelimhw


nmegill:
@wlammen These are nice proofs.  I propose you do this:  

(1) delete ax12olem1OLD, ax12olem2OLD; 
(2) rename ax12olem1-ax12olem7 and ax12o to ax12olem1OLD-ax12olem7OLD and ax12oOLD, adding "Obsolete version of ... as of ..." but also keeping the original comments in the descriptions since they don't match exactly your ALT versions;
(3) rename ax12olem1ALT-ax12olem4ALT and ax12oALT to ax12olem1-ax12olem4, ax12o.

For (1), you can alternately rename them to ax12olem1OLDOLD and ax12olem2OLDOLD.  I'll leave it up to you.  For (2), obviously their proofs will have references to the *OLD versions, but that is OK; when they expire, we will delete them as a group.

Norm

nmegill:
@wlammen ... and don't forget the usage and proof mod discouraged tags on the OLDs. :)
Norm

PR #297, opened 2018-01-30
 Moving section on uniform spaces into the main body of set.mm

tirix:
I'm again having issues with git showing huge changes, though I'm pretty sure I did everything right this time. Checking the changes using the command `git diff 8e9b28e 6186ac0 set.mm`, I can see that starting from metrics, diff seems to lose synchronization.


tirix:
`git diff --minimal 8e9b28e e2aea14 set.mm` actually produces a better result!

tirix:
Thank you!

PR #298, opened 2018-01-30
Modify iset.mm proof of dveeq2 to be intuitionistic

jkingdon:
Turns out that distinctors can function in intuitionistic logic, at
least given `ax-i12` and intuitionistic propositional logic
such as `orel2`.

Reduces "show usage ax-3/recursive" count by 10.


jkingdon:
@nmegill Did you see this one? I'm fairly close to submitting some further work which depends on it and #294 so if you get the chance to look at this first, things may be a little less confusing.

PR #299, opened 2018-01-30
 former ax12oALT is now the proof for ax12o 

wlammen:


wlammen:
@nmegill Note that your mathbox now references a couple of the ax12lem*OLD lemmas.

PR #337, opened 2018-02-24
Add some truth table entries and related theorems to iset.mm

david-a-wheeler:
Most of these proofs are from set.mm; credit is retained (unmodified)
in those cases.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -8815,6 +8815,32 @@ $)
       ( wtru tru ax-mp ) CADBE $.
   $}
 
+  $( If something is true, it outputs ` T. ` .  (Contributed by Anthony Hart,
+     14-Aug-2011.) $)
+  tbtru $p |- ( ph <-> ( ph <-> T. ) ) $=
+    ( wtru tru tbt ) BACD $.
+
+  $( If something is not true, it outputs ` F. ` .  (Contributed by Anthony
+     Hart, 14-Aug-2011.) $)
+  nbfal $p |- ( -. ph <-> ( ph <-> F. ) ) $=
+    ( wfal fal nbn ) BACD $.
+
+  ${
+    bitru.1 $e |- ph $.
+    $( A theorem is equivalent to truth.  (Contributed by Mario Carneiro,
+       9-May-2015.) $)
+    bitru $p |- ( ph <-> T. ) $=
+      ( wtru tru 2th ) ACBDE $.
+  $}
+
+  ${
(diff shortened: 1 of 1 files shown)


jkingdon:
Basically looks good, a few minor points:

1. Travis is failing because the "Version of" date at the top of the file needs to be updated.

2. Looks like `iset.mm` already has `dfnot` under the name `dfneg` . My preference, where theorems exist in `set.mm`, is to use the `set.mm` names. So I'd prefer `dfnot` (and it looks like `alnex` is the only proof which would need to be updated in order to drop `dfneg`).

3. Similarly, `notfalbitru`, `notfal`, and `fal` ideally would follow their `set.mm` counterparts. None of this has anything to do with intuitionistic versus classical logic, just that `set.mm` has changed since the version which `iset.mm` forked from, and ideally we would move over the recent changes in `set.mm` (those that make sense intuitionistically, that is). In this case the renames section of `set.mm` says `4-Jun-15 notfal    fal`.

4. I'm not sure exactly how the comments shoudl read in introducing "truth tables". Truth tables don't have the same role in the semantics of intuitionistic logic that they do for classical logic, but I guess they are a thing at least in the sense that the theorems which are part of this pull request are still valid.

5. The `show usage ax-3/recursive` count goes from 256 to 258 with this pull request. Do you know which two proofs aren't intuitionistic? I didn't spot them, but I was only sometimes looking at the "This theorem was proved from axioms" text at the bottom of the page.

We don't want to be adding new usages of `ax-3` (unless perhaps if it is in the classical logic section at the end of the file, and even then generally only if we call out the non-intuitionistic nature of the result in the comments).

jkingdon:
Found the `ax-3` usages. The proof of `dfnot` is not an intuitionistic one (it uses `ja`). Replacing that with the `dfneg` proof should take care of both of them (the proof of `inegd` looks fine except for the `dfneg` problem).

david-a-wheeler:
I think it's all resolved now (except for the expected difference in dates at the top of iset.mm).

david-a-wheeler:
This is probably best pulled in as a "squash and merge", since I don't think people will want to see the intermediate steps.

jkingdon:
I went ahead and deleted the `itruth` branch on the theory that we're done with it (and that deleting branches as soon as they are merged tends to be a good time to clean them up).

david-a-wheeler:
Sure! Sounds good, I do the same on other projects.

PR #344, opened 2018-02-26
shorten ax12b

wlammen:


wlammen:
The Travis check has passed (click on yellow bullet to verify), but the web page is not updated accordingly

nmegill:
@wlammen If you mean the us2.metamath.org pages, those get updated after a site rebuild that takes around 7 hrs to run.  I usually start the site rebuild after I merge a commit.  If it is still running from a previous commit, I may postpone starting it for the new commit until the next day.  I also have a procedure to update only selected theorems more quickly, but it isn't automatic (it requires specifying the theorems by hand) so I don't use it except in special cases.

wlammen:
@nmegill No, the yellow bullet indicating a running Travis check did not turn into a green check mark, even after checks have passed and after reloading this page. Now everything is fine, so it was some temporary hiccup.

Issue #316, opened 2018-02-10
Add more T. (true) and F. (false) theorems to iset.mm


jkingdon:
I ran `search * 'T.'` and `search * 'F.'` in `set.mm` and found a lot of theorems which are not in `iset.mm` yet. The vast majority hold intuitionistically, and the proofs probably aren't especially hard. In some cases the proofs from `set.mm` will work unmodified.

This would be a good beginner project for someone who is learning logic or metamath.

david-a-wheeler:
Well, I'm not *exactly* a beginner, but I'm definitely a beginner at intuitionistic logic.  Besides, it looked like fun, so I did a few.  It turns out that many of the same theorems work just fine unchanged, so I just copied them from set.mm with credit retained.  See #337.

jkingdon:
@david-a-wheeler Now that #337 is merged, are we ready to close this issue? I didn't see any other theorems which caught my eye, but if you are aware of some, we can list them here.

david-a-wheeler:
I want to include exclusive or and nand. In particular, exclusive or is the basis of some ancient propositional logic that I would really like to prove is also valid in intuitionistic logic. However, I'm not sure I understand how exactly to represent exclusive or in intuitionistic logic. It appears to me that several different expressions for exclusive or that are exactly the same in classical logic are not the same in intuitionistic logic, and I do not understand which one is "correct" in intuitionistic logic. I started looking for sources about this, but I am probably looking in the wrong places. Any suggestions?

jkingdon:
There is a section about exclusive or at http://us2.metamath.org:88/ileuni/mmtheorems19.html#mm1847s but there is more that could be done here.

See https://github.com/metamath/set.mm/pull/236 which includes a web page link which maybe should be in our bibliography. That web page also mentions existential uniqueness in the same breath as exclusive or, and no one has even started on existential uniqueness in iset.mm.

nmegill:
A Google search for "exclusive or in intuitionistic logic" returns ileuni/xor.html as the first hit, which isn't helpful, but some way down the page it shows https://ncatlab.org/nlab/show/exclusive+disjunction that comments on it, saying that some classical equivalents aren't equivalent intuitionistically but not committing to which one is the right one.  Based on the usage in English, I might go for "p or q but not both" expressed as "(p or q) and not (p and q)", which isn't mentioned there, and I don't know if it is equivalent to any of theirs.

david-a-wheeler:
Thanks for the reference!

[...]

Issue #339, opened 2018-02-25
Prove sbco2, hbsb, sbco3 and related theorems in iset.mm


jkingdon:
We are one distinct variable constraint away from having these. See [sbco2v](http://us2.metamath.org:88/ileuni/sbco2v.html), [hbsbv](http://us2.metamath.org:88/ileuni/hbsbv.html), and [sbco3v](http://us2.metamath.org:88/ileuni/sbco3v.html) for versions with the extra constraint. In classical logic, it is possible to prove two cases, where a pair of variables are identical or where they are distinct, and use [pm2.61](http://us2.metamath.org:88/ileuni/pm2.61.html) to get the desired result. In intuitionistic logic, one needs a disjunction (a la [nfsb2or](http://us2.metamath.org:88/ileuni/nfsb2or.html)), direct application of [ax-i12](http://us2.metamath.org:88/ileuni/ax-i12.html), or some other technique.

We can also add `sbcom` to the list of theorems in this family; see #348 for the connection with `sbco3`.

Feel free to ask me (@jkingdon) for more background on what I have tried or discovered. But also feel free to try new things, because so far the breakthroughs have come in ways that have been hard for me, at least, to anticipate.

Oh, and if you do make progress, submitting frequent pull requests is encouraged. There may be other people working on this or closely related things.

jkingdon:
This is done, at least assuming [ax-bnd](http://us2.metamath.org:88/ileuni/ax-bnd.html). See #364 and related issues and pull requests.

And note that the theorems in the subject line are (along with every predicate logic theorem which hasn't been moved to the end of the file) proved intuitionistically in iset.mm, as can be seen by running `show usage ax-3/recursive`.

PR #349, opened 2018-03-01
Clarify some xor results are not intuitionistic

jkingdon:
That is, update the comments to say (at least in some of the places)
when we are talking about classical exclusive-or and not intuitionistic
exclusive-or.

There's more to do here and I think @david-a-wheeler wanted to work on it, but since
@nmegill noticed (in #316 ) that http://us.metamath.org/ileuni/xor.html was at the
top of the google search results for "exclusive or intuitionistic logic" we need a quick
fix to avoid looking bad. If people arrive at this section via the theorem list, they'll
see the introduction (right under the "Exclusive or and related theorems (classical)"
section header) which clarifies classical versus intuitionistic, but if they deep link
into xor, they won't.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -14934,9 +14934,10 @@ also part of this exclusive-or family. Many are expressed in terms
 of the biconditional and negation.
 $)
 
-  $( Two ways to express "exclusive or."  Theorem *5.22 of [WhiteheadRussell]
-     p. 124.  (Contributed by NM, 3-Jan-2005.)  (Proof shortened by Wolf
-     Lammen, 22-Jan-2013.) $)
+  $( Classical ways to express "exclusive or."  Theorem *5.22 of
+     [WhiteheadRussell] p. 124.  In classical logic these are equivalent, but
+     in intuitionistic logic they are not.  (Contributed by NM, 3-Jan-2005.)
+     (Proof shortened by Wolf Lammen, 22-Jan-2013.) $)
   xor $p |- ( -. ( ph <-> ps ) <->
                 ( ( ph /\ -. ps ) \/ ( ps /\ -. ph ) ) ) $=
     ( wn wa wo wb wi iman anbi12i dfbi2 ioran 3bitr4ri con1bii ) ABCDZBACDZEZAB
@@ -14963,7 +14964,8 @@ $)
                 ( ( ph /\ -. ps ) \/ ( ps /\ -. ph ) ) ) $=
     ( wb wn wa wo xor dfbi3 xchnxbi ) ABCABDZEBADZEFABEKJEFABGABHI $.
 
-  $( Two ways to express "exclusive or."  (Contributed by NM, 1-Jan-2006.) $)
+  $( Two ways to express classical "exclusive or."  (Contributed by NM,
+     1-Jan-2006.) $)
   xor3 $p |- ( -. ( ph <-> ps ) <-> ( ph <-> -. ps ) ) $=
     ( wn wb pm5.18 con2bii bicomi ) ABCDZABDZCIHABEFG $.
 
@@ -14972,8 +14974,8 @@ $)
   pm5.15 $p |- ( ( ph <-> ps ) \/ ( ph <-> -. ps ) ) $=
     ( wb wn xor3 biimpi orri ) ABCZABDCZHDIABEFG $.
 
-  $( Two ways to express "exclusive or."  (Contributed by NM, 3-Jan-2005.)
-     (Proof shortened by Wolf Lammen, 24-Jan-2013.) $)
+  $( Classical ways to express "exclusive or."  (Contributed by NM,
+     3-Jan-2005.)  (Proof shortened by Wolf Lammen, 24-Jan-2013.) $)
   xor2 $p |- ( -. ( ph <-> ps ) <-> ( ( ph \/ ps ) /\ -. ( ph /\ ps ) ) ) $=
     ( wb wn wo wa xor3 pm5.17 bitr4i ) ABCDABDCABEABFDFABGABHI $.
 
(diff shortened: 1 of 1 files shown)


digama0:
You stated in the other thread that `(p \/ q) /\ -. (p /\ q)` is weaker than `(p \/ q) /\ (-. p \/ -. q)`, but this is not correct. We know `p \/ q`. If `p`, then since `q` implies a contradiction to `-. (p /\ q)`, we have `-. q`, so `-. p \/ -. q`. Similarly, if `q` then `-. p` and hence `-. p \/ -. q`. Thus `(p \/ q) /\ -. (p /\ q)` implies `-. p \/ -. q`, and of course it implies `p \/ q`. The other direction of implication is trivial as you observed.

digama0:
In set.mm, there was not a definition for xor for a while, because of the short alternative `-. ( ph <-> ps )`. This is not suitable in ile for the reasons I've mentioned, but set.mm has also since adopted a notation for xor, `( ph \/_ ps )`, and I would prefer to see this notation used in the basic theorems.

jkingdon:
@digama0 good catch on (p \/ q) /\ -. (p /\ q) not being weaker than (p \/ q) /\ (-. p \/ -. q) . I'll try to write that up as a proof (together with the xor notation as you suggest) unless you or @david-a-wheeler beats me to it.

david-a-wheeler:
@jkingdon - please go ahead. In particular I would really like to see some sort of official definition of intuitionistic xor, that is, for:
( ph \/_ ps )

I would be happy to help, but right now I am not entirely clear about which theorems I can use that are truly intuitionistic. Intuitionistic logic is interesting, but it is certainly not what I grew up on, so I feel a little uncertain starting out with it.

jkingdon:
@david-a-wheeler You can use anything which doesn't depend on ax-3 (as shown in the axioms at the bottom of the web page or `show usage ax-3/recursive` in metamath.exe). We want to move ax-3 much later in the file, to better segregate the classical stuff, and are most of the way there except #339

My proposed official definition of exclusive or is at #350 (right now a link to a branch, will make it a pull request once I add a few proofs and fix a few formatting issues).

As for how to learn intuitionistic logic, the Bauer paper in the bibliography is a fun read (even though it is hardly nuts and bolts stuff), iset.mm is hopefully helpful (once you understand the "don't use ax-3" caveat and perhaps a few other rough edges we haven't fixed yet), other than that not too sure. I think there are textbooks but I don't know them well enough to make a recommendation.

david-a-wheeler:
The Bauer paper is delightful, but I feel like I'm walking through a house of mirrors while blindfolded.  I think I need to try to actually *create* some intuitionistic proofs to really get it.  Thanks for the "don't use ax-3" rule.  Once the definition of intuitionistic xor is set, I plan to give it a try.  In particular, I intend to try to prove some of the Stoic logic indemonstrables (as developed by Chrysippus of Soli), or at least try to.  Specifically I'd like to try mpto2 and mtp-xor.  The definition of xor will be different (I believe), so a new proof will be needed if it's provable at all.

david-a-wheeler:
@jkingdon - I propose that you merge df-xor with the unproven theorems as comments, if you don't have time to prove the rest.

I think this proof:

~~~~
  $( One way of defining exclusive or.  Equivalent to ~ df-xor .  (Contributed
     by Jim Kingdon, 1-Mar-2018.) $)
  xoranor $p |- ( ( ph \/_ ps ) <-> ( ph \/ ps ) /\ ( -. ph \/ -. ps ) ) $= ? $.
~~~~

... can be easily proved by expanding the LHS using df-xor, and using ioran to show equivalence.

digama0:
@david-a-wheeler ioran is not intuitionistically valid (most of the de morgan laws are only valid in one direction). You have to use the left conjunct in the proof of equivalence, as I indicated here https://github.com/metamath/set.mm/pull/349#issuecomment-369644423 .

digama0:
Re: your plan to prove the Stoic logic theorems, I can tell you that `mpto1`, `mpto2`, `mtp-xor`, and `mtp-or` are all intuitionistically valid. Usually in intuitionistic logic, a theorem is either "true for obvious reasons" or not provable. So unless you try to use a "slick" proof that uses some invalid equivalences, the natural proof that presents itself will be intuitionistically correct. Just try not to think in terms of truth tables, think of `-. ph` as `ph` implies a contradiction, and think of `ph \/ ps` as knowing "positively" that one of the two is true. (It may help to use the disjunction property as a guide: if intuitionistic logic proves `|- ph \/ ps`, then either it proves `|- ph` or it proves `|- ps`. That is, an or can only be true for "trivial reasons". Classical logic of course fails this.) Can you see the difference between "if I show you that `ph` is contradictory, then you can show me that `ps` is true" and "one of `ph` or `ps` is known to be true"? These are the readings of the intuitionistically non-equivalent statements `-. ph -> ps` and `ph \/ ps` respectively.

jkingdon:
The writeup of @digama0 's proof above is at #351 .

I've recorded future work at #350 but we don't have to wait for that to merge this pull request and #351

jkingdon:
@david-a-wheeler Totally understand about the "house of mirrors while blindfolded" - I still make mistakes (for example, I claimed that xoranor was only valid in one direction, but then @digama0 corrected me and sketched out the proof of the other direction which I then wrote up for #351 ). Others that really shocked me the first time were con2i and alnex (which are intuitionistically provable, but look like things which aren't).

That's the best thing about doing this in metamath - you have the verifier to tell you whether you are getting it right.

david-a-wheeler:
I've started with some proofs.  I am shocked that "notfal" is provable in intuitionistic logic.  Are we accurately capturing intuitionistic logic?  If so, a lot of things I thought would be unprovable in intuitionistic logic turn out to be "no problem".

digama0:
@david-a-wheeler Remember what I said about `-. ph` meaning `ph` implies a contradiction. Of course `F.` implies a contradiction, since it is a contradiction. More generally, all truth table facts are provable in intuitionistic logic, because if you restrict your attention to formulas that satisfy excluded middle, intuitionistic and classical logic coincide. And of course both `T.` and `F.` satisfy excluded middle, which is to say that each of them is either true or false - since they *are* true and false (respectively). The difference only comes in when you start talking about *general* propositions `ph` and assert that they must be either true or false. If you want an example, think of the fact that some turing machine halts - it may not be either provable or disprovable, and for such statements you will not be able to prove excluded middle.

A terminological note: propositions that satisfy excluded middle are called decidable propositions. True and false are decidable, as indicated; the haltingness of (some) turing machines is not decidable.

david-a-wheeler:
Thanks.  The Stoic logic indemonstrables require handling more than just raw "false" but they were easier for me than I expected.

[Pull request 352](https://github.com/metamath/set.mm/pull/352) contains some work I've done to prove a few small theorems in intuitionistic logic.

Quite illuminating.

jkingdon:
The changes in this pull request itself are no longer needed since #405 (and other work in the last month or two) has cleaned up the exclusive-or theorems of iset.mm to be free of `ax-3`.

As for the other topics discussed in the comments above, I think they are all resolved as well but feel free to open specific issues with particular topics as needed.

PR #355, opened 2018-03-05
updating mathbox

Achaemenid:
Adding two theorems on measurability.


nmegill:
@Achaemenid 
Hi Brendan,

A comment about the size of the ismblfin proof.

In an ideal world, proof size shouldn't matter, but unfortunately some of our software functions have run time that increases nonlinearly, possibly exponentially in some cases, with proof size.  In particular, the 'minimize' and 'save proof.../compressed' functions in metamath.exe become very slow with extremely large proofs.  Maybe the algorithms can be improved, but for now they are what we have to live with.

Very roughly, the 'minimize' algorithm attempts to match each hypothesis of a trial theorem with every step of the proof.  If the proof has s steps and the trial theorem has n hypotheses, this amounts to order of s^n matching tests.

While it is an impressive result, ismblfin is now the largest proof in the database (around 200K uncompressed steps including syntax building).  Previously, your itg2addnc (about 145K steps) was the largest and was a bottleneck for our full set.mm 'minimize' run a couple of months ago, requiring a separate job just for it running on a dedicated core.  'save proof itg2addnc/compressed' takes 24 sec and 'save proof ismblfin/compressed' takes over 100 sec, showing the nonlinear behavior vs. proof size.

While we don't have a hard and fast rule, in the past we have broken up large proofs into smaller lemmas with a goal of making 'save proof.../compressed' take at most 2 to 3 sec.  When you get a chance, it would be good to break up ismblfin (and maybe eventually some other large ones like itg2addnc).  If you don't have time to break it up then eventually I or someone else probably will, but you understand the proof structure and might know better what lemmas might make the most sense.  If there are lemmas that can be reused several times with different substitution instances by the final proof, it can even make the net size of the proof+lemmas smaller in some cases.

FYI the 'minimize' and 'save proof' commands in metamath.exe have a /time qualifier that displays the run time.

A secondary issue is human understanding of the proof.  It can be hard to follow a very long standalone proof.  But if broken into well-chosen lemmas, the final proof that ties the lemmas together can provide a better understanding of the big picture.

Thanks,
Norm

Achaemenid:
Honestly, there's a part of the proof I probably should have separated out
anyway, if only because it owes too much to lecture 7.  That'll probably
mean essentially eliminating the web page's steps 1-398 and 651-1223 for
three uses of a new lemma (although there'll certainly be some doing
getting the ducks in a row for it).

itg2addnc, on the other hand, I actually was working on when I saw this,
but right now all I'm doing is making the statement of the first lemma less
unwieldy (in a fairly obvious way).  That should pare it down a tiny bit,
but I'm not sure there's really a natural cut, although it might be clearer
as I work on it.  When I first sat down to it, I had genuinely thought most
of the legwork was done; you'll notice the description is essentially
describing the second lemma more than the main part of the proof.

You didn't mention itgaddnclem2, but that one comes in at 153K and I think
there might be a fairly clean cut to be made, so I'll probably start on
that after these two.  I don't think there's really a whole lot I can do
with itg2addnc, though.

PR #357, opened 2018-03-09
iset.mm: Add a relationship between xor and other possible definitions

jkingdon:
I don't know for sure that `ph <-> -. ps` is weaker than `ph \/_ ps`, but here's an intuitionistic proof that the latter implies the former.


digama0:
`ph <-> -. ps` is strictly weaker than `ph \/_ ps`. Assuming the former, this is equivalent to `-. ps \/_ ps`, or `(-. ps /\ -. ps) \/ (ps /\ ps)`, which is clearly equivalent to decidability of `ps`.

PR #361, opened 2018-03-11
Refactoring: Uniform spaces are now generated by pseudo metrics

tirix:
Beyond changes in my mathbox, mainly for the Bochner Integral, this commit introduces the following changes in the main body of set.mm:

- Uniform spaces generated by metrics, ` metUnif ` , now use pseudo metrics instead of extended metrics. The old version of the theorems are kept, but renamed ~OLD, and marked as change and usage discouraged. They can eventually be removed.
- This prompted changing the definition of balls, with were previously defined on extended metrics, and which I've generalized as defined on ` _V ` as suggested by Mario. For several theorems related to balls, there are now new versions, valid for pseudo metrics, which names are suffixed with ~ps. For the moment, I've kept the older version unchanged, with very similar proofs. Since all extended metrics are also pseudo metrics, one possibility would be to keep them, but shorten them by using their pseudo-metric counterpart. Another possibility would be to altogether replace them by their pseudo metrics counterparts, using ~OLD prefixes like for ` metUnif `, but this would mean revising all theorems using them. Any advice?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2879,6 +2879,9 @@
 "cdleml2N" is used by "cdleml3N".
 "cdleml3N" is used by "cdleml4N".
 "cdleml4N" is used by "cdleml5N".
+"cfilucfil2OLD" is used by "cfilucfil3OLD".
+"cfilucfil2OLD" is used by "cmetcuspOLD".
+"cfilucfilOLD" is used by "cfilucfil2OLD".
 "cgcdOLD" is used by "ee7.2aOLD".
 "ch0" is used by "nonbooli".
 "ch0" is used by "omlsii".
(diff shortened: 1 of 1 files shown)


tirix:
[Wikipedia](https://en.wikipedia.org/wiki/Metric_(mathematics)#Extended_metrics) indeed notes "∞-metric", but I don't know of official/academic sources for that notation.

tirix:
My next steps would then be the following:

- prove that the "integral metric" (`sibm`) on simple functions is a pseudo metric, inducing a uniform space,
- define the space of integrable functions as the Hausdorff completion of that space (like `RR` to `QQ`) - the concept of Hausdorff completion still needs to be developped,
- define the (Bochner) integral as the extension by continuity of the integral for simple functions (`sitg`) to that space.

PR #372, opened 2018-03-22
Add axbnd to set.mm.

jkingdon:
Since we recently added ax-bnd to iset.mm, we should prove it as
a theorem of classical logic (as we do with the other axioms of
intuitionistic predicate logic which are not axioms of set.mm).

I just realized I forgot to do this when I submitted #369


nmegill:
@jkingdon The Travis 'discouraged' check fails because ax-12o is obsolete (shouldn't be used in new proofs).  Instead, use ax12o.

PR #385, opened 2018-04-08
Auxiliary theorems, especially for Words

avekens:
Auxiliary theorems (required for graph theory), especially for Words: subwords of subwords, subwords of concatenated words - in my Mathbox.
Two theorems in Mathbox of  Thierry Arnoux are used in my Mathbox, so I moved them into the main part.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14843,6 +14843,7 @@ New usage of "equs5eOLD" is discouraged (0 uses).
 New usage of "equsalOLD" is discouraged (0 uses).
 New usage of "equsalhwOLD" is discouraged (0 uses).
 New usage of "equsexOLD" is discouraged (0 uses).
+New usage of "equviniOLD" is discouraged (0 uses).
 New usage of "erngbase-rN" is discouraged (4 uses).
 New usage of "erngdv-rN" is discouraged (0 uses).
 New usage of "erngdvlem1-rN" is discouraged (3 uses).
@@ -17700,6 +17701,7 @@ Proof modification of "equs5eOLD" is discouraged (50 steps).
 Proof modification of "equsalOLD" is discouraged (54 steps).
 Proof modification of "equsalhwOLD" is discouraged (79 steps).
 Proof modification of "equsexOLD" is discouraged (48 steps).
+Proof modification of "equviniOLD" is discouraged (122 steps).
 Proof modification of "eujustALT" is discouraged (188 steps).
 Proof modification of "ex-natded5.13" is discouraged (67 steps).
 Proof modification of "ex-natded5.2" is discouraged (18 steps).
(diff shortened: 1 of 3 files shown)


nmegill:
@avekens a conflict arose after I merged @wlammen's pull request

avekens:
OK, I solved the conflict - it was again only the date at the beginning of set.mm: $( set.mm - Version of 8-Apr-2018 $). It was the second time I had this conflict, and it takes about 1 hour to solve this trivial problem. Is there a way to avoid this in the future?

This problem should occur every time changes in two forks are to be merged into one version of the main-branch..

Alexander

wlammen:
The line in question is a sort of a revision number. See here https://stackoverflow.com/questions/4120001/what-is-the-git-equivalent-for-revision-number for reasons why git, by its decentralized nature, cannot support these automatically.

Usually the date will match the latest date in an annotation, and, thus, is often redundant, but not always, as typo fixing, editorial stuff often goes without an explicit 'Revised by' tag. Within a git repository "git log set.mm" displays the time stamp of latest commit, which could be used as a revision number replacement, but that won't help users, who keep set.mm outside of repositories.

I wonder why you needed 1h work to fix this issue. This seems to be an unduly long time amount. What exactly were your problems, even doing it the second time? By contrast, the last merge conflict I suffered from (due to an external rewrap) was resolved in less than 15 minutes. 

Wolf

jkingdon:
I do need to resolve that conflict from time to time but I guess I am more used to resolving git conflicts so it doesn't seem like a big deal (although if I added up all the time to notice it, fix it, wait for Travis, etc maybe it could approach an hour).

I'm probably not going to object to whatever people decide, though.

wlammen:
@nmegill Fine with me.

avekens:
Thanks for the discussion of my concern. Since I am not very experienced with Git, especially in working with forks (only dealing with branches seems to be easier), I am always afraid to do something wrong. 

If it is easily possible to remove the date check from the Travis script, and to maintain the date in the metamath-branch, I would appreciate that very much. So changes only in one's mathbox should be possible (almost) without conflicts.

nmegill:
I modified .travis.yml to no longer check the version date on set.mm.

I put the following note at the top of set.mm:
$( To prevent GitHub merge conflicts, please do not change the above date
 unless there are no other pull requests in the queue.  Travis has been
 modified to use "verify markup * /skip_date" to prevent a non-updated date
 from causing an error.  I will maintain the date when necessary.  --NM $)

The rule is that if there are pending pull requests for set.mm, don't update the version date.  Otherwise you are encouraged to update it (to help keep it current and reduce my maintenance effort), but if there is any doubt, feel free to leave it alone.

Norm

PR #387, opened 2018-04-16
Some exclusive-or for decidable propositions in iset.mm

jkingdon:
The general course of this project is to go through analogues of `xor3`, `pm5.17`, `xor2`, and `nbbn` for decidable propositions.

I haven't had a lot of time for this, but figured I should sent what I have so far.

cc @david-a-wheeler because this is about xor in iset.mm (even though my angle isn't really developing xor as such, so much as just trying to get rid of the current uses of `ax-3`).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -9107,6 +9107,14 @@ $)
   xornbi $p  |- ( ( ph \/_ ps ) -> -. ( ph <-> ps ) ) $=
     ( wxo wn wb xorbin pm5.18im con2i syl ) ABCABDEZABEZDABFKJABGHI $.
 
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Thanks for the cc, makes sense to me.  Clearly "DECID ..." is going to show up in many places, and that makes sense.

jkingdon:
Yes, `DECID` is currently our main tool for handling theorems which don't hold unconditionally in intuitionistic logic.

There might be a few places where we have more `DECID` than necessary (for example on both `ph` and `ps` when only one is needed), and if so I'd welcome pull requests to tighten those results.

PR #389, opened 2018-04-21
shorten ax11v2, dvelimf, ax10o, hbae

wlammen:
@nmegill ax10lem3 could simplify the proofs of ax10o and hbae. Why is it forbidden to use it there?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14568,6 +14568,7 @@ New usage of "dveeq2OLD" is discouraged (2 uses).
 New usage of "dvelimALT" is discouraged (1 uses).
 New usage of "dvelimALTNEW7" is discouraged (0 uses).
 New usage of "dvelimf-o" is discouraged (3 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
@wlammen "ax10lem3 could simplify the proofs of ax10o and hbae. Why is it forbidden to use it there?"

At one time maybe I wanted to keep the uses of ax10/ax-10 "pure" or something, but there isn't a need for that since so many things have changed, and ax10 is now just another theorem.  Go ahead and use ax10lem3 and delete its "new usage is discouraged".  I will wait for that before I merge your pull request.

"The text in ax11o is slightly incorrect, as it is already proven without ax-16."

I think I meant "neither ax-16 nor ax-17", but it refers to an obsolete axiomatization that is no longer of interest.  I will fix or remove this and some other text after you are done.

Norm

wlammen:
That's it for today.

PR #390, opened 2018-04-22
shorten dral1, chvarv, cbval, cbval2, sbft

wlammen:
@nmegill dral1 can be further shortened by ax10lem3. I discouraged the lemma's new usage for now, to prevent Metamath's minimizer spread a lemma across the whole file. ax10o and hbae, which I simplified with the lemma recently, are very fundamental results, dral1 is not. However, one should consider turning ax10lem3 into an ordinary theorem, because of its wider application range.

Wolf

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1187,13 +1187,6 @@
 "ax-pre-mulgt0" is used by "axmulgt0".
 "ax-pre-sup" is used by "axsup".
 "ax10lem2OLD" is used by "ax10lem3OLD".
-"ax10lem3" is used by "ax10".
-"ax10lem3" is used by "ax10NEW7".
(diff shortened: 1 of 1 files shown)


nmegill:
@wlammen You can rename ax10lem3 to ax10o2, with the new description "Same as ~ ax10o but with reversed antecedent.  (Contributed by NM, 25-Jul-2015.)".   Add an entry to the "Recent label changes" list at the top of set.mm.  Post a note here to let me know you're done for the day. :)

nmegill:
Oops - clicked the wrong button.  Pull request has been reopened. - Norm

wlammen:
That's it for today

PR #396, opened 2018-04-29
shorten a couple of theorems about ax12o

wlammen:
I added two theorems to my mathbox that explain what A. x y = z and E. x y=z exactly mean. I used these results sometimes to better understand intermediate proof steps. I don't know whether they are worth being included into the main part.


wlammen:
@nmegill us2.metamath.org:88 is currently down, so "wget -N -q http://us2.metamath.org:88/ocat/mmj2/mmj2jar.zip" fails during the travis checks. Cannot do anything about it

nmegill:
@wlammen The us2 server shouldn't have been in the Travis script.  I changed it to us.metamath.org which is more reliable.  Your commit merged with no failures.
Norm

PR #403, opened 2018-05-05
shorten drnf2

wlammen:
1. shorten proof of drnf2
2. remove an internally duplicated proof step resulting in really small proof  shortening
3. add wl-nfnbi to my mathbox, an obvious generalization of nfimd


wlammen:
closed because the destination is not the development branch

PR #404, opened 2018-05-05
shorten drnf2 &nfsb4t

wlammen:


wlammen:
1. shorten drnf2
2. remove an internally duplicated proof step resulting in a very small improvement on sbequi
3. add wl-nfnbi to my mathbox, a more complete theorem than nfnd

PR #409, opened 2018-05-07
Add extensionality and class abstraction to iset.mm

jkingdon:
There are some comments here about the direction of adding IZF to iset.mm, which strikes me as the obvious next step (in the comment there's some discussion of IZF versus CZF). Of course, if there is something I'm not aware of, let me know.

What is actually in this pull request is the axiom of extensionality (which is identical in ZF and IZF) and some of the things which can be proved given only it and logic. This includes the whole class abstraction section. There is a certain amount of getting iset.mm up to date with the latest set.mm conventions to make this work, but that's a good thing in general as well.

If this looks good, I'll continue on to the other IZF axioms.


jkingdon:
I added the missing bibliographic entries (not sure why I didn't notice them before) and the checks are now passing.

PR #410, opened 2018-05-08
Expand iset.mm intro to include ax-bnd

jkingdon:
This is the web page which provides an overview of the axioms,
and which was not updated for ax-bnd until now.

The wording is mine but I make this change at the request of Gérard Lang.


wlammen:
"...just a slight variable of the classical..."
Suggest to replace variable with variation.

Wolf

jkingdon:
Thanks for catching that typo, @wlammen . I have added applied that fix to this pull request.

PR #412, opened 2018-05-12
substitution improvements for iset.mm (nfsb4or and more)

jkingdon:
Many of these will soon be useful in the set theory. The summary is making
iset.mm more in line with set.mm and making various theorems available
in `A. x x = y \/ something` format rather than `-. A. x x = y -> something`.

- Add nfsb4or, an analogue to hbsb4 but stronger in that it is stated
with a disjunction rather than with negation and implication.

- add dvelimor

    This is a version of dvelim which uses a disjunction instead
    of negation and implication, making it stronger (in intuitionistic
    logic). The proof is based off the one for dvelimALT, with the
    main change needed being changing ax-i12 to ax-bnd.

- Add nfalt
    
    This is a closed form of nfal and follows from df-nf and the
    usual universal quantification theorems (alim etc).

- Make sbf more like set.mm
    
    - Rename sbf to sbh
    - Add sbf which uses F/

- Add naecoms
    
    The proof is different from set.mm but is straightforward.

- Make nfsb4t, etc, more like set.mm
    
    - Rename sbied to sbiedh
    - Add sbied with the same statement (not the same proof) as
      the one from set.mm
    - Add nfsb4t
    - Copy dvelimdf from set.mm (both statement and proof)

- Add nfsbd

    - add nfsbxyt, a lemma for nfsbd
    - add nfsbt, a lemma for nfsbd
    - add nfsbd, which is like hbsbd but is proved in a somewhat
      different way

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 6-May-2018
+$( iset.mm - Version of 9-May-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm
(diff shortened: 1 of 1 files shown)


wlammen:
@jkingdon dvelimdf: My name and shortened is misspelled. These typos existed in set.mm for a couple of days before my latest pull request fixed them. Unfortunately, you copied them to iset.mm during their short lifetime. Murphy's law.

jkingdon:
@wlammen Thanks for noticing that. I pushed another commit to this branch which pulls in the fix from set.mm

PR #413, opened 2018-05-12
shorten spei and cbv3/cbv3hALT

wlammen:
I'm a bit puzzled about the cbv* section.

1. What is the reason to block cbv3h from modification? It's proof is not very efficient and can be shortened quite a bit, as demonstrated in cbv3hALT.

2. If you provide a theorem in deduction form, it seems a good idea to have all necessary F/ declarations attached to the common antecedent.  Otherwise your proof winds up with extra for-all quantifiers in the consequence, as demonstrated in cbv1, for example.  More, the theorem is less versatile than can be, since the hypotheses are stricter than necessary.  For example, any leading existential quantifier provides the F/ prerequisite, but is not directly supported.  On the other hand, if in a concrete application a F/ prerequisite is missing from the common antecedent, sps provides an easy means to supply it.  Of course, if there is a vast number of usages, convenience theorems can save some writing effort.  But the cbv* theorems are not frequently called.

Any suggestions on how to proceed?


nmegill:
@wlammen 

1. The "(Proof modification is discouraged.)" in cbv3h is a mistake.  I think what happened was approximately the following.  Before the introduction of F/, there was a cbv3 that avoided ax-12o (one of many experiments that eventually led to discovering the present ax-12).  Below it, there was a cbv3ALT with a shorter proof, and the "discouraged" was to prevent "minimize" from wiping it out using cbv3, in order for the shorter cbv3ALT proof to be available if later we decided it wasn't necessary to avoid ax-12o.  Then, Mario revised cbv3 to use F/, and cbv3ALT was renamed cbv3h.<P>Solution:  remove "(Proof modification is discouraged.)" from cbv3h and shorten its proof however you would like.  I think you added cbv3hALT, which of course can be removed.

2. The way to fix cbv1 to avoid the extra foralls is to add two additional hypotheses, F/ x ph and F/ y ph.   [These hypotheses can be equivalently expressed as (ph -> F/ x ph) and (ph -> F/ y ph) to achieve a "pure" deduction form, but slightly longer proofs would result so we probably don't want to do that.]  Then all antecedents can be pure ph with no foralls, so that ph will be free for any substitution when used as part of a deduction.<P>I agree that the early cbv's before cbval are somewhat awkward with somewhat inconsistent forms.  These were written very early in set.mm development before I knew what the best forms were.  One feature of cbv1 is that cbv1.3 requires satisfying only one direction, which occasionally might be useful.  If you want to revise or replace the early cbv's, that would be fine with me.

Norm

wlammen:
@nmegill 
> The way to fix cbv1 to avoid the extra foralls is to add two additional hypotheses, F/ x ph and F/ y ph. 

That was what I meant when I wrote

> it seems a good idea to have all necessary F/ declarations attached to the common antecedent

Valid not only for cbv1, but as a general rule for theorems in deduction form.

I reworked cbv3(h) accordingly and will file a new pull request soon.

Wolf

PR #417, opened 2018-05-13
rework cbv* section

wlammen:
1. I changed the hypotheses block of theorem cbv1/cbv2 to better match our usual style, so in these cases an incompatible change occurs.  Fortunately, these theorems are hardly used, so within set.mm this is no problem.  I still wonder whether this incompatibility must be noted somewhere outside of the comment section of the theorem. Another issue are the comments in the corresponding *OLD theorems, that start with "Obsolete version..." instead of "Obsolete proof ...". I don't know whether the changed text interferes with some automatic procedures.

2. The cgv*h theorems are left as they are, although their hypotheses block is in uncommon style, too. Again, their usage is rare, and I decided against reworking them, mostly because I don't expect much effect from the efforts. Of course, shortenings are applied where appropriate.

3. An astute reader will notice, that the requirements of cbv1 for example can be slightly weakened from F/x ph & ph -> F/x ch to F/x (ph -> ch). The second is a straightforward consequence of the first couple via nfim1, but the proof requires only the latter.  I occasionally stumble over such possible generalizations, and wonder what to do then.

4. I clustered all cbv* theorems and moved them right after the spim* section.  In my eyes they belong there, given their dependencies and sophistication level.

Wolf

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2795,6 +2795,8 @@
 "cbncms" is used by "minvecolem4a".
 "cbncms" is used by "ubthlem1".
 "cbncms" is used by "ubthlem2".
+"cbv1OLD" is used by "cbv3OLD".
+"cbv2OLD" is used by "cbvaldOLD".
 "cbvexsv" is used by "onfrALTlem1".
 "cbvexsv" is used by "onfrALTlem1VD".
 "cdj3lem1" is used by "cdj3i".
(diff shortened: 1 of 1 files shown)


nmegill:
@wlammen 

1. There are currently no tools that recognize this.  When purging expired *OLDs by hand, I often just search for e.g. "-Apr-2017.[space]" which 99% of the time matches a *OLD.  I would suggest continuing to use "Obsolete proof" for shortened proofs and "Obsolete version" for other revisions.

2. OK

3. I think they should be left as separate hypotheses, because (1) that is what the user normally expects and (2) it may require an extra ~ nfim in a proof that references it.  I think it would be rare that the hypothesis weakening would have a benefit, but if such a case does arise we can decide on a case by case basis.

4. OK.

Norm

PR #425, opened 2018-05-24
Start the process of intuitionizing existential uniqueness in iset.mm

jkingdon:
This is only a start, but I thought I should send what I have so far.

It doesn't look like existential uniqueness needs any fundamental change for intuitionistic logic. The definition at https://ncatlab.org/nlab/show/quantification#quantification_in_firstorder_logic_with_equality corresponds to http://us.metamath.org/ileuni/eu1.html which already is proved without ax-3.

There's also a (small) update to some comments in set.mm which I found as I was looking through this.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 19-May-2018
+$( iset.mm - Version of 23-May-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm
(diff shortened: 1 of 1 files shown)


nmegill:
@jkingdon I see you are calling E* (at most one) "uniqueness", and E! "existential uniqueness".  I don't recall seeing E* called "uniqueness".  It is called "at most one" in Bell & Machover, and the word "existential" is sometimes omitted to describe E!.   For example, the ncatlab reference you gave uses "uniqueness quantification" for E! and not "existential uniqueness quantification".  I know that "at most one" can be an awkward expression sometimes, but it is very clear, and I am wondering if it might be confusing to use "uniqueness" for E*.  Do you have a reference?

nmegill:
@jkingdon Ignore my previous message.  I was swapping red and green in my head (I wasn't expecting that "uniqueness" was incorrectly used for "at most one" in set.mm).  Thanks for making the corrections.

jkingdon:
I too was a bit surprised the E! versus E* errors stayed there as long as they did. Thanks for checking my work rather than blindly merging it (and if anyone wants to do some more proofreading, go for it. I was fixing the ones I noticed rather than trying to check every comment).

PR #428, opened 2018-06-09
Define wrecs

sctfn:
By analogy with the recs functor, I've added wrecs for well-founded recursion

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4431,6 +4431,7 @@
 "df-recs" is used by "recseq".
 "df-recs" is used by "recsfval".
 "df-recs" is used by "tfrlem9".
+"df-recs" is used by "tfrALTlem".
 "df-rngo" is used by "isrngo".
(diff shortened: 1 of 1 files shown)


nmegill:
@sctfn The Travis check failed because of mismatches in the 'discouraged' file.  E.g. tfr1ALT has "(Proof modification discouraged.)" because we don't want it to be wiped out by tfr1 during a 'minimize' run.  This means, though, if you change the tfr1ALT proof you will have to regenerate 'discouraged'.

To regen 'discouraged' w/ metamath.exe, use
<pre>
./metamath 'read set.mm' 'set width 9999' 'show discouraged' quit | tr -d '\015' | grep '^SHOW DISCOURAGED.' | sed -E -e 's/^SHOW DISCOURAGED.  ?//' | LC_ALL=C sort > discouraged
</pre>
(mmj2 is faster but I don't have the command handy.)

PR #452, opened 2018-07-15
iset.mm: add "Proper substitution of classes for sets" section

jkingdon:
Reading over this change, I realize it looks like a lot of changes. But a lot of the changed lines are easy updates to the way set.mm now does things, like using `wsb` instead of `cv wsbc` and using `F/` more places.

As for the section itself, it was mostly just the usual kinds of changes. `sbc2or` and `rmo2` didn't make it (for multiple reasons in the former case, for lack of `mo2` in the latter). But those are not heavily used.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -22605,13 +22605,9 @@ $)
 
      Our definition also does not produce the same results as discussed in the
      proof of Theorem 6.6 of [Quine] p. 42 (although Theorem 6.6 itself does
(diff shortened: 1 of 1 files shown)


jkingdon:
Thanks for the reminder on the brackets (fortunately that only affects typesetting so it does not need any intuitionizing in terms of the proofs, $c, etc).

There are a number of places where iset.mm is behind set.mm on typesetting. Hopefully someone will have some time to pull over all the things which apply. I made an issue at https://github.com/metamath/set.mm/issues/454

PR #455, opened 2018-07-15
Remove dfsb2 and dfsb3 from iset.mm

jkingdon:
There doesn't seem to be any need to say anything about dfsb3
because the point is to use primitive connectives (and what it
considers to be primitive is not the same as iset.mm).

As for dfsb2, it would probably not be especially easy to
come up with a version which uses DECID or other conditions,
and if we did, the extra conditions might be clutter as much
as something which shows the contrast between classical and
intuitionistic logic. So putting the definition in a comment
and not trying to prove anything about it may be the best
option.

Reduces "show usage ax-3/recursive" count by two.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -12249,9 +12249,9 @@ $)
      When ` x ` and ` y ` are distinct, we can express proper substitution with
      the simpler expressions of ~ sb5 and ~ sb6 .
 
-     In classical logic, we have additional equivalent definitions ~ dfsb2 and
-     ~ dfsb3 , but we do not have intuitionistic proofs that those are
-     equivalent.
+     In classical logic, another possible definition is
+     ` ( x = y /\ ph ) \/ A. x ( x = y -> ph ) ` but we do not have an
+     intuitionistic proof that this is equivalent.
 
      There are no restrictions on any of the variables, including what
      variables may occur in wff ` ph ` .  (Contributed by NM, 5-Aug-1993.) $)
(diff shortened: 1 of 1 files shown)


jkingdon:
@nmegill @digama0 Any reactions? I'd like to continue getting rid of `ax-3` (this pull request gets it down to 69 usages).

nmegill:
I think removing them is fine.   They aren't used for anything even in set.mm (although Wolf Lammen's recent shortening of sbn uses dfsbc3, but that is easily reverted if it is an issue).

I'm not sure why I missed this pull request, sorry.

jkingdon:
I think we've seen for a while that the shortest `set.mm` proofs are sometimes non-intuitionistic (even in cases where an intuitionistic proof exists). And unless we change something moderately fundamental about our criteria, that just means we'll need to live with `set.mm` sometimes diverging from `iset.mm`. As for the specific proof at http://us.metamath.org/mpeuni/sbn.html it uses `exanali` and `annim` in addition to `dfsb3`, so it is a pretty thoroughly non-intuitionistic proof. And `iset.mm` has taken a different path to `sbn` for a while.

As an aside, thanks for importing so much old history into git. On at least one occasion, I was able to go through the set.mm git history to find an old intuitionistic proof, which had since been replaced by a non-intuitionistic proof (which was shorter or otherwise more appealing).

No worries about missing pull requests occasionally. I have a loose policy of asking again after about a week (and just branching off a branch where a non-merged pull request is something I need to build on top of for my next chunk of work).

PR #457, opened 2018-07-18
Add basic set operations and subsets to iset.mm

jkingdon:
This adds the "Define basic set operations and relations" and "Subclasses and subsets" sections to iset.mm.

As long as #456 is not merged, this pull request will also include all the commits from that pull request. If the decision is to merge them both as-is, it should just work, but if there are revisions to be done, I can revise them accordingly.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -24374,6 +24374,1000 @@ $)
   $}
 
 $(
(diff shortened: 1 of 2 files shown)


nmegill:
@jkingdon There are some missing refs in mmil.html causing Travis to fail. - Norm

jkingdon:
@nmegill Thanks for the heads-up on the failed Travis build. Fixed.

PR #464, opened 2018-07-28
rm dependencies of sbequ8

wlammen:
This contribution is second in a row to show that the proof of sbiedALT does, in fact, not depend on ax-7. It uses theorems only that are now known to be provable without ax-7.

This proof of sbequ8 has its dependencies dramatically reduced, and it can be moved next to df-sb.

As usual, this flexibility has its costs: The proof appears unwieldily long, compared to the legacy one.


nmegill:
@wlammen Nice work!

We should revert  sbiedALT, sbieALT, sbied, sbie back to sbied, sbie, sbiedOLD, sbieOLD.  Just copy/paste them unchanged from the old set.mm; no need to mention ax-7 since if we adopt your new proofs (which we should if we are concerned with ax-7) then none of these will use ax-7.

I think we should use your new mathbox proofs as the official ones in spite of their longer lengths, and rename the shorter original ones to *ALT instead of *OLD so their shorter proofs will be available to people who don't care about ax-7.

Norm

PR #465, opened 2018-07-28
iset.mm: add more of "The difference, union, and intersection of two classes"

jkingdon:
The sections "The union of two classes" and "The intersection of two classes" went in easily.

The section "Combinations of difference, union, and intersection of two classes" was a whole lot of work. It is full of class equivalents of DeMorgan's law and double negation, and similar theorems. Many of the theorems there turned from equalities to subset relationships; others still hold but needed different proofs.

A number of these changes were best done by improving the propositional logic section (for example, by adding implications which are in classical logic biconditionals). This should make it easier to get rid of ax-3 later.

The section "Class abstractions with difference, union, and intersection of two classes" is a short one and it needed only one change, which was for `symdif2` to specify exclusive-or using our exclusive-or notation, rather than a version of exclusive-or which is only equivalent classically. This might be a good change for set.mm as well, if people think it is a good thing to use the exclusive-or notation more widely.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -25953,6 +25953,304 @@ $)
     ( cun wral wa ralunb biimpri ) ABCDEFABCFABDFGABCDHI $.
 
 $(
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+      The intersection of two classes
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  ${
+    $d x A $.  $d x B $.  $d x C $.
+    $( Expansion of membership in an intersection of two classes.  Theorem 12
+       of [Suppes] p. 25.  (Contributed by NM, 29-Apr-1994.) $)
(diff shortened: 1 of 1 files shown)


nmegill:
@jkingdon Travis shows a failure, but it isn't reported in the log.  The log seems to show metamath dying silently after "Checking statement comments...":

MM> verify markup *

I ran it on my system and got this:

MM> v m *
?Error: The bibliographic reference "[Monk1]" in statement "ssin" was not found
as an an <A NAME="Monk1"></A> tag in the file "mmil.html".

I tried 2 different compilers, and both gave the above i.e. neither was dying after "Checking statement comments...".  Anyway this is the error that has to be fixed, but it would be good to know why it isn't showing in the Travis log.

Norm

jkingdon:
Yeah, I'm not too sure about exactly how all this `VERIFY MARKUP` stuff works. I personally don't see it all locally because I run `VERIFY MARKUP */F` (because I don't have the gif files around, mainly).

Thanks for looking into this; I've pushed up an addition of Monk1 to the mmil.html bibliography.

nmegill:
@jkingdon  A quick way to get all the symbols is to unzip http://us.metamath.org/downloads/symbols.zip 
It should have all of the symbols you need (plus many others that you don't).

Norm

jkingdon:
OK, I have the symbols, messed around with my git ignores, and am now running VERIFY MARKUP * without having to supply /F.

PR #470, opened 2018-08-03
Rewrite naming section of iset.mm

jkingdon:
Remove list of theorem renames.  It was far from complete, and if
people don't find the git history or other mechanisms sufficient,
we'd need to make a much more concerted effort to update this
list and keep it up to date.

When saying that iset.mm forked from set.mm, also mention the
large number of copies from set.mm since then.

Describe how we name things, particularly how naming relates
to set.mm names.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,7 +1,8 @@
 $( iset.mm - Version of 2-Aug-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
-set.mm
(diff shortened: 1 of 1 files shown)


jkingdon:
@nmegill @digama0 @david-a-wheeler Any reactions? Pretty much everything in here is negotiable if we'd rather do something differently.

nmegill:
I seem to have overlooked it.  It seems fine to me, and I will merge it.

PR #476, opened 2018-08-11
Add alternative `axiom` in formal system developed by Heyting

sofigern:
I'm new one publisher and enthusiast of Metamath project and I've chosen this missed theorem to add as a training example.
I'd like to contribute more but I believe that due to lack of experience and language my changes and naming of those changes can be the complete disaster.
That's why your review would be especially appreciated, because any feedback will be useful!
Also I've only russian edition of Heyting introduction to Intuionistic so page link could be misleading. If you can help me with this I'd really appreciate that!
Thanks!

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 6-Aug-2018
+$( iset.mm - Version of 11-Aug-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm
@@ -27,6 +27,7 @@ PC  Paul Chapman         FL  Frederic Line          AS  Alan Sare
 SF  Scott Fenton         RFL Roy F. Longton         ES  Eric Schmidt
 JGH Jeffrey Hankins      JM  Jeff Madsen            DAW David A. Wheeler
 AH  Anthony Hart         RM  Rodolfo Medina         GD  Georgy Dunaev
+MM  Mykola Mostovenko
 
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
                           Contents of this header
@@ -1331,6 +1332,11 @@ $)
      Lammen, 13-Nov-2012.) $)
   simpr $p |- ( ( ph /\ ps ) -> ps ) $=
     ( ax-ia2 ) ABC $.
+
+  $( Heyting's formal system Axiom #1 from [Heyting] p. 127.
+     (Contributed by MM, 11-Aug-2018.) $)
+  ax1hfs $p |- ( ph -> ( ph /\ ph ) ) $=
+    ( wa ax-ia3 pm2.43i ) AAABAACD $.
 
   ${
     simpli.1 $e |- ( ph /\ ps ) $.
@@ -28370,4 +28376,3 @@ htmldef "DECID" as "<SMALL>DECID</SMALL> ";
 /* End of typesetting definition section */
 
 $)
-
--- a/mmset.html
+++ b/mmset.html
@@ -5845,6 +5845,10 @@ available at
 http://www.math.cornell.edu/~hatcher/AT/ATpage.html</A>
 (retrieved 11 Nov 2014).</LI>
 
+<LI><A NAME="Heyting"></A> [Heyting] Heyting, A., <I>Intuitionism: An
+introduction</I>, North-Holland publishing company, Amsterdam, second
+edition (1966).</LI>
+
 <LI><A NAME="Herstein"></A> [Herstein] Herstein, I. N., <I>Abstract
 Algebra,</I> Macmillan Publishing Company, New York (1986) [QA162.H47
 1986].</LI>


sofigern:
Ugh, also there are some bibliography issues with set.mm so tests fails and for now I don't understand how to fix that =(

david-a-wheeler:
No problem! If you click on the "details" to see why the build failed, you'll see this:

~~~~
MM> write bibliography mmbiblio.html
The old file will be renamed mmbiblio.html~1.
Reading definitions from $t statement of set.mm...
1362 typesetting statements were read from "set.mm".
?Warning: No page number after [<author>] bib ref in "ax1hfs".
3748 references were processed.
2005 table rows were written.
MM> verify markup */date_skip
?Error: The bibliographic reference "[Heyting]" in statement "ax1hfs" was not
found as an <A NAME="Heyting"></A> anchor in the file "mmset.html".
?Warning: No page number after [<author>] bib ref in "ax1hfs".
MM> exit
~~~~

Basically, we want to make sure we "cite our sources" and we check that in a picky way.  The "conventions" section of set.mm at <http://us.metamath.org/mpegif/conventions.html> describes the conventions used in iset.mm as well:

Bibliography references. Please include a bibliographic reference to any external material used. A name in square brackets in a comment indicates a bibliographic reference. The full reference must be of the form KEYWORD IDENTIFIER? NOISEWORD(S)* [AUTHOR(S)] p. NUMBER - note that this is a very specific form that requires a page number. There should be no comma between the author reference and the "p." (a constant indicator). Whitespace, comma, period, or semicolon should follow NUMBER. An example is Theorem 3.1 of [Monk1] p. 22, The KEYWORD, which is not case-sensitive, must be one of the following: Axiom, Chapter, Compare, Condition, Corollary, Definition, Equation, Example, Exercise, Figure, Item, Lemma, Lemmas, Line, Lines, Notation, Part, Postulate, Problem, Property, Proposition, Remark, Rule, Scheme, Section, or Theorem. The IDENTIFIER is optional, as in for example "Remark in [Monk1] p. 22". The NOISEWORDS(S) are zero or more from the list: from, in, of, on. The AUTHOR(S) must be present in the file identified with the htmlbibliography assignment (e.g. mmset.html) as a named anchor (NAME=). If there is more than one document by the same author(s), add a numeric suffix (as shown here). The NUMBER is a page number, and may be any alphanumeric string such as an integer or Roman numeral. Note that we require page numbers in comments for individual $a or $p statements. We allow names in square brackets without page numbers (a reference to an entire document) in heading comments. If this is a new reference, please also add it to the "Bibliography" section of mmset.html. (The file mmbiblio.html is automatically rebuilt, e.g., using the metamath.exe "write bibliography" command.)

In the set case of iset.mm, the corresponding bibliography entry would be in "mmiset.html" instead.

Does that help enough?

jkingdon:
Hey, great to see more people interested in `iset.mm`. One of the things I've been not quite getting around to is spending more time with the intuitionistic/constructive literature. There may well be theorems and proofs worth adding, or even if it is just existing theorems, it is always good to have literature citations.

You ask about naming. In most cases we name things according to our own conventions. For example, conjunction corresponds to `an` and idempotence corresponds to `idm`. In fact there is currently a theorem `anidm` in `set.mm` and `iset.mm` which is `( ph /\ ph ) <-> ph`. If we also want `ph -> ( ph /\ ph )` we would tend to name it `anidmr` (that is, the reverse direction of `anidm`. If it were the forward direction we'd add `im` instead of `r`). In the case of a biconditional, often we'll just have the biconditional version (but being able to cite other authors verbatim can be a consideration, so this might be a case where we would want the reverse direction as well as `anidm`). 

In a few cases, like Principia and Margaris, we name theorems according to the numbering of the source we take them from. We tend to do that for only a handful of sources though, and I'm not sure whether I'd put Heyting in that category.

The way we name things isn't an exact science, though, so I hope trying to find the perfect name doesn't stop you from contributing.

Please ask if you have any other questions about how to contribute to `iset.mm`.

sofigern:
wow, that was helpful. I really missed `anidm` theorem, most probably I haven't used to the search abilities of a software and the Metamath language itself. I'm definitely not insisting on adding redundant  theorems to the database it was just a nice example I was not able to found in `iset.mm` and it was easy to start and find useful feedback. About the very this theorem, I'm not familiar with literature enough to decide if it will be useful later, so probably something will be changed when I'll just know more. 

(Not sure probably google group is a better place for this question)
What about intuitionistic literature, I just starting to dig into it, but mostly I see the lack of formalization or I see some formal theory is built on top of assumptions like (let's assume that rational numbers theory is built so let's build the theory of real numbers/integration/whatever). My question is if I can add somewhere, will say into my `Mathbox`, some unproven stuff as axioms, so I or community can prove those later because to translate some known theory is much easier work than built new theory and prove that by yourself. And I'm not sure that I won't be stucked on the very beginning if I'll try to prove some things, which I don't know how to prove yet. Or probably there is some other way to build things from results I want and digging into details during the process. Maybe it just should be the different branch or so. Your advice on this purpose will be very helpful for me. Thanks!

jkingdon:
@sofigern Don't worry about not finding the existing theorem, I didn't know about it at first either. Some theorems I remember (although the database is too large to keep them all in my head), some theorems I find with SEARCH or wondering where they might be used, some I can find by browsing particular sections. In this case I found it by thinking of "what would I name that theorem in this pull request" and found we already had one.

As for writing your own proofs or axioms, I definitely encourage that. Making a branch (maybe using the github fork button) is one good way. Mathboxes are a way to do somewhat the same thing.

Your idea of writing some axioms is not a bad one. If you look at something like real numbers, or rational numbers, in `set.mm` it builds on some concepts which are, at least to me, kind of unfamiliar (having to do with functions and relations and things like iota). But writing out some axioms in whatever language you are familiar with might be a good way to figure out what is possible (and don't worry if you don't get them right the first time, there will be time to correct them later if need be).

The harder part of course is deciding what to work on when getting started. Shortening or otherwise improving existing proofs has been an activity for some which is fairly concrete. In my case, moving theorems from `set.mm` to `iset.mm` and seeing what still works has been a thing (I'd certainly welcome help with that so feel free to ask if you have questions as you start trying that out).

jkingdon:
@david-a-wheeler Just FYI the bibliography for Intuitionistic Logic Explorer is in `mmil.html` (you might already know that, but giving the correct filename might reduce confusion). Thanks for the explanations of how the bibliography stuff works.

sofigern:
Thanx a lot for all the people in the thread! Communication was very encouraging and helpful, I eager to deliver some useful stuff! Now my weekend is over, so I hope to find more time and work around this topic as soon as possible =) I hope I'll be able to help! Have a nice week=)

nmegill:
Thank you for your contribution.  I'm traveling today and will review this tomorrow.

My first reaction is that this belongs in iset.mm but not set.mm.  In particular, since since set.mm proves the iset.mm axioms in axia1 through axio, any prop. calc. theorem in iset.mm is automatically provable from the set.mm axioms, so in a sense it is redundant to have it in set.mm also (unless there is a separate use case for it being a frequently-used utility theorem, which I don't think it would be since we already have anidm).

On the other hand, showing the equivalence of alternate axiomatizations of intuitionistic prop. calc. would be a welcome contribution to iset.mm, just as we show the equivalence of various alternate axiomatizations of classical prop. calc. in set.mm.

Norm

sofigern:
For now, I'm closing the PR as not useful. Probably as it was mentioned it will be useful to prove equivalence of axiom systems later.

jkingdon:
@sofigern Sounds good. Please do open a new PR if some part of this (or related work) seems worth reviving.

PR #477, opened 2018-08-13
Add "Unordered and ordered pairs" to iset.mm

jkingdon:
One decision here was to rewrite the proofs which had been using the conditional operator `if`. I've been assuming that the conditional operator is hard to intuitionize and not especially valuable if its main use is the weak deduction theorem. If `dfopif` and/or future things like it (which do not involve the weak deduction theorem) are more numerous than I thought, perhaps this topic is worth a closer look.

This is the usual collection of some proofs which just need some rearrangement, some results which are weakened in various ways (usually the obvious way implied by the current proof; I don't in all cases know that the non-weakened theorem could not be proven in intuitionistic logic).

There's a bit more inhabited versus nonempty class stuff here too.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 11-Aug-2018
+$( iset.mm - Version of 12-Aug-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm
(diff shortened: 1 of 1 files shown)


jkingdon:
Sounds like `if` doesn't have a lot of usages (at least, not until we get to things like proving that natural number equality is decidable, which a ways off if we continue taking things in `set.mm` order).

Thanks for the tips on `A e. _V` case elimination.  It will take me a while to fully grasp what proving `( ( A e. _V -> ph ) -> ph )` is like, but it does seem like something which suffices for this purpose and may well be easier to prove than `A e. _V \/ ph`.

As for notation for inhabited sets, I've been using `∃𝑥 𝑥 ∈ 𝐴` so far because (a) there are existing theorems in `set.mm` which use it (although of course they don't have the same need to distinguish it from `𝐴 ≠ ∅` so they tend to fall back to the latter pretty soon after `n0` enters the picture), (b) a new notation like `INHABITED A` is more of a last resort than a preference most of the time, (c) `∃𝑥 𝑥 ∈ 𝐴` seems fairly concise (only mentions `A` once; no need to substitute anything larger than a single variable for `x`).

Granted, I had not thought of the possibility of redefining `≠`. It is an intriguing idea and I wonder if there would be applications other than comparing with `∅`. I'd be reluctant to pick `≠` as the notation for this predicate, though, for much the same reasons that constructive mathematicians tend to use `#` for apartness and contrast it with `≠` which is negated equality at places like https://en.wikipedia.org/wiki/Apartness_relation or the HoTT Book (section 1.12.3 Disequality).

PR #478, opened 2018-08-13
updating mathbox

Achaemenid:
Heine-Cantor theorem


Achaemenid:
Minimization didn't save somehow.  New request in half an hour or so.

PR #479, opened 2018-08-13
Add iset constructs for stability and testability

david-a-wheeler:
This adds "stable" and "testable" constructs to iset.mm,
along with a citation that explains them.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 6-Aug-2018
+$( iset.mm - Version of 13-Aug-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm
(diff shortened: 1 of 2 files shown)


jkingdon:
Looks good to me. I assume the next step will be to prove things like "Decidability implies stability . . . The conjunction of stability and testability is equivalent to decidability."

jkingdon:
@david-a-wheeler Unlike set.mm, you should update the date at the top of the file.

david-a-wheeler:
> Looks good to me. I assume the next step will be to prove things like "Decidability implies stability . . . The conjunction of stability and testability is equivalent to decidability."

Yes, but I thought it'd be wise to make sure we agree on the definitions first :-).

I used long names for w... and df-... to avoid matching anything in set.mm.

digama0:
Isn't `TESTABLE ph` just the same as `DECID -. ph`? I don't have a good sense for what properties could be stable but not decidable, but another interesting theorem I would recommend in this space is `STABLE (DECID ph) -> DECID ph`.

PR #483, opened 2018-08-15
Add to iset basic theorems about decidability and testability

david-a-wheeler:


david-a-wheeler:
@jkingdon - in the longer run I suspect some theorems requiring decidability really only require stability or testability.  But first things first... we need to define those conditions and their relationship to decidability!

jkingdon:
First things first isn't a bad motto for much of what we do. Glad to see you looking into the propositional logic topics, because I haven't really done a lot with it since roughly when I started set theory.

PR #489, opened 2018-08-25
Add some theorems to iset.mm for natural deduction

david-a-wheeler:
The set.mm documentation theorem "natded" lists set.mm theorems
that are equivalent to various natural deduction rules.
This commit adds several of these theorems that were previously
not in iset.mm (by copying them from set.mm, including their credits).
As a result, this commit makes it easier to apply
natural deduction approaches within iset.mm.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -4594,6 +4594,14 @@ $)
       ( wn con3d imp ) ACEBEABCDFG $.
   $}
 
+  ${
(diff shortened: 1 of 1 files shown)


jkingdon:
Nice.

david-a-wheeler:
Thanks! I think pm2.18da should be possible too, but it did not just copy over. See: https://github.com/metamath/set.mm/issues/490

PR #494, opened 2018-08-28
Add "Theorems requiring subset and intersection existence" section to iset.mm

jkingdon:
Add  `alexnim` (one direction of `alexn` from set.mm).

Break `intex` into two theorems: `inteximm` is the forward direction (but for inhabited rather than non-empty classes) and `intexr` is the reverse direction.

The reverse direction of `intnex` is the one we already have a proof for. Although it is pretty much trivial, just include that one and name it `intnexr` .

Intuitionize `intexab` and `intexrab` in iset.mm. In both cases, prove the forward direction and accordingly
rename to `intexabim` and `intexrabim` respectively.

Rename `iinexg` to `iinexgm` and change non-empty to inhabited. Also, the proof, although basically the same in structure, needs changes throughout for intuitionistic logic. Specifically, `r19.35-1` in place of `r19.35` , `r19.2m` in place of `r19.2z` , `abid` in place of `abn0` , and `inteximm` in place of `intex` (which also means we need a variable change).

Remove `intabs` from iset.mm. The proof is not easily intuitionized and it is only used once
in set.mm (some distance in the future).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -32718,21 +32718,6 @@ $)
   $}
 
   ${
-    $d x y $.  $d x A $.  $d y ph $.  $d x ps $.  $d x ch $.
-    intabs.1 $e |- ( x = y -> ( ph <-> ps ) ) $.
-    intabs.2 $e |- ( x = |^| { y | ps } -> ( ph <-> ch ) ) $.
-    intabs.3 $e |- ( |^| { y | ps } C_ A /\ ch ) $.
-    $( Absorption of a redundant conjunct in the intersection of a class
-       abstraction.  (Contributed by NM, 3-Jul-2005.) $)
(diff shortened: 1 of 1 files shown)


nmegill:
Travis failed because the us.metamath.org mirror is down.  So, I pointed the us.metamath.org DNS to the China mirror (AFAIK there aren't any tariffs on that yet).  Try again.

jkingdon:
I was having trouble finding the Travis retry button (I thought I did at one point but now I can't find what screen that would have been on).

I see you already merged this, so I guess it is a moot point.

nmegill:
@jkingdon The problem was that the mirror temporarily replacing us.metamath.org doesn't allow directory listing of .../metamath/.  So I changed wget in .travis.yml to fetch metamath.tar.bz2 instead and untar it.

Norm

PR #498, opened 2018-09-02
A proof of Euler's partition theorem

tirix:
This finally completes the proof for Euler's partition theorem in my mathbox.
This number 45 in the Metamath 100 list!


nmegill:
Congratulations, well done!
Norm

david-a-wheeler:
On September 2, 2018 7:01:33 AM EDT, nmegill <notifications@github.com> wrote:
>Congratulations, well done!
>Norm
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/metamath/set.mm/pull/498#issuecomment-417921931

That is fantastic, and the big congratulations! Unless Norm says otherwise, I will tweak the various pages and send an email to Freek.
--- David A.Wheeler

tirix:
I actually started the proof last year (based on Mario's starting point)! 
The final proof dates back from then, but I often work backwards, so the date at which I complete the different lemmas is not linear. For example I only completed the proof for the ~ eulerpartlemn lemma this Friday. If you want to retain a date for the overall completion, that shall be August 30th, 2018. We could log that date for the final theorem, if that's clearer.

david-a-wheeler:
Ok. I think it would be best if the comments in eulerpart included the 30-Aug-2018 date somewhere ( it doesn't have to be the final listed date, just somewhere noting that that was the date when all the sub pieces were completed). In the meantime, I will update various files to use that 30 August 2018 date.

digama0:
I agree that `eulerpart` should have the final date of completion of the whole sequence, rather than the date you finished that part of the proof (presumably relatively trivial). Also, congratulations on finally finishing the proof!

PS: I use the `lemx` naming style (with letters) for unfinished proofs, labeled in the order I invent the lemma statements, to prevent too much theorem renaming, and when I finish the proof I label them all sequentially in file order with numbers like `lem5`.

tirix:
Thank you al! 
And thanks again to Mario too for the bootstrapping. 
Yes, some sequential renaming of the lemmas could help... and some of lemmas proven might actually not be used in the proof!

PR #507, opened 2018-09-06
Merge WIP from branch Proba

tirix:
This merges mostly work in progress in my mathbox from the "proba" branch.
Introduces semirings, ordered monoids, ordered groups, ordered rings and ordered fields.
The only valuable theorem is probably the proof that ordered Archimedean groups are abelian.


nmegill:
@tirix @david-a-wheeler @digama0 Anyone know why the Travis failed here?
What is continuous-integration/travis-ci/push  vs. continuous-integration/travis-ci/pr?
The first one says there was a 10min timeout.  The second one passed all the normal checks.

Norm

tirix:
Thank you!
I guess "pr" stands for "pull request", while "push" is just for the set of commits. The travis check would actually run on the same data.

PR #518, opened 2018-09-15
shorten sbequi again

wlammen:


wlammen:
I woke up in the middle of the night, this idea sprang to my mind, and I had to write it down. Weird. Hopefully this won't become a habit. Go to sleep again.

PR #519, opened 2018-09-15
shorten equvini again

wlammen:
@nmegill I cleaned up my Mathbox and removed some old studies. One of the theorems, wl-bitr1, is elsewhere in use (one hit). Should it be put in the main part then?

As requested, I provide a short proof of sbcom3 in my Mathbox. Apart from the much shorter proof, the unnecessary distinct variable constraints have been removed.
An alternative, slightly modified form (wl-sbcom3a) avoids references to ax-7 (if that still matters).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13146,19 +13146,8 @@
 "w-bnj19" is used by "bnj978".
 "watfvalN" is used by "watvalN".
 "watvalN" is used by "iswatN".
-"wl-bibi1" is used by "wl-bibi1d".
-"wl-bibi1" is used by "wl-bibi1i".
-"wl-bibi1d" is used by "wl-bibi2d".
 "wl-bitr1" is used by "e2ebindALT".
-"wl-bitr1" is used by "wl-bibi1".
-"wl-bitr1" is used by "wl-bitr".
-"wl-bitr1" is used by "wl-bitrd".
-"wl-bitr1" is used by "wl-bitri".
-"wl-jarri" is used by "wl-dedlem0a".
-"wl-jarri" is used by "wl-pm2.86i".
 "wl-mps" is used by "wl-syls1".
-"wl-pm5.74lem" is used by "wl-pm5.32".
-"wl-pm5.74lem" is used by "wl-pm5.74".
 "wvd2" is used by "dfvd2".
 "wvd2" is used by "dfvd2i".
 "wvd2" is used by "dfvd2imp".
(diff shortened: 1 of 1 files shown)


benjub:
The statement of wl-sbcom3a is more natural; I think that wl-sbcom3a should become the main statement and be labeled sbcom3 and put in the main part.

nmegill:
@wlammen I don't think putting wl-bitr1 will be of much benefit if it only shortens one proof.  wl-bitr1 is just ex applied to bitr, and doing that in AS's e2ebindALT increases the length by only 7 bytes.  You can do that directly in his mathbox, which shouldn't be referencing other mathboxes anyway.   (At you convenience.  I would do it myself but am deep into the df-riota change at the moment.  No need to make a "revised" comment about it.)  Also, if we run minimize on e2ebindALT, the proof length decreases from 283 bytes to 136 bytes with no use of wl-bitr1, but I think he wants the proof to illustrate his virtual deduction tool, so other than wl-bitr1 elimination we should leave it alone for now.

Great job on sbcom3(a)!

@benjub I agree that wl-sbcom3a should be moved up to sbcom3. It is on my to-do lilst, but first I want to decide with GM what to do about his apparently automatically-generated, 4,731,945-step (uncompressed) sbcom3 proof.  It can't stay in set.mm in its present form because it is a burden on some tools, such as taking 6 hrs to enter MM-PA (the cause of which I am investigating separately in metamath.exe).

Norm

david-a-wheeler:
@nmegill : Okay, I'll rise to the bait.  Other than as a stress test, why would we want a  4,731,945-step (uncompressed) proof of sbcom3 in set.mm???  It might be an interesting demo, but in that case, put it in a separate .mm file.

nmegill:
@david-a-wheeler We don't want it.  It will go away in due time after I get a chance to discuss its disposition with GM.  It has been there for a while in GM's mathbox, and O'Cat "discovered" it a few days ago while trying to "save proof */normal" to test his old mmj2 version.  It doesn't hurt anything too badly except things like MM-PA and "minimize".  But first I wanted WL to find a shorter proof, which he did, to show that a shorter one existed.

In spite of its length, in principle I find it fascinating that GM apparently has an automated proof-generation tool that can prove something that is somewhat advanced in predicate logic (with the df-sb definition and $d conditions).  In practice, it has the size problem as well as being nearly impossible for a human to follow.

I also wonder if this hints that the reason most other automated provers (actually all that I am aware of) never produce actual formal proofs (in the textbook sense) is that they would be gigantic and incomprehensible.

Norm

david-a-wheeler:
I'm impressed with a tool that can automatically generates such long proofs. But I hope that they would be minimized easily, and we would use the minimized the version.

PR #521, opened 2018-09-16
Order topology on ordered spaces & misc.

tirix:


david-a-wheeler:
I look forward to seeing the filled in proof of Descartes rule of signs.

tirix:
It may take a while! :D 
I started the proof having in mind to do it for a more general ordered field, since I recently introduced this notion([~df-ofld](http://us2.metamath.org/mpeuni/df-ofld.html)). It led me to consider the "intermediate value theorem" [~ivth](http://us2.metamath.org/mpeuni/ivth.html) for more generic ordered spaces, which then takes the form of [~conima](http://us2.metamath.org/mpeuni/conima.html), and so I started to try to express connectedness for a generic ordered space - this led me to write [~ordtcomlem1](http://us2.metamath.org/mpeuni/ordtconlem1.html). This also requires the space to be complete. And since any complete ordered field is isomorphic to the real numbers, I might as well do the proof specifically for the real numbers! So I will probably rewrite it simply for real numbers, in a first step.

Now I wonder if I shall use ` [( Poly1 ` ( CCfld |`s RR ) )](http://us2.metamath.org/mpeuni/df-ply1.html) ` as a frame for the polynomials, as I originally planned, or rather ` [( Poly ` RR )](http://us2.metamath.org/mpeuni/df-ply.html) ` which is specific to ` RR `. @nmegill , @digama0 , what would you suggest?

tirix:
I believe I will go with ` Poly `... but I'm still interested in your opinion, and I still believe we shall have common clear guidelines!

nmegill:
For PosetRel vs. Poset, PosetRel is much older and doesn't generalize easily to more specialized structures.  PosetRel was introduced several years before the extensible structure mechanism, and Poset was intended as its replacement.  If you look at the "Recent Label Changes" at the top of set.mm, on 4-Nov-12 we changed many definitions to use extensible structures, and the idea was that the ones it replaced would become obsolete.  Most of the old definitions went into a "deprecated" section of set.mm, although a small number such as Top were kept for general use simply because they were so convenient to use.  It appears that PosetRel was also kept out of the "deprecated" section, although I can't recall why we were led to that decision.  Maybe Mario remembers.  I am inclined to prefer Poset, but if PosetRel will lead to significant proof size savings, perhaps it would be better to use.  For some of the older definitions we have conversions to the newer ones, as in grpo2grp, but I don't see one for Poset vs. PosetRel.

As for A e/ B vs -. A e. B:   so far it has been relatively rare that A e/ B would be advantageous (unlike A =/= B vs. -. A = B, which is ubiquitous).  My practice has been not to use A e/ B generally, because converting back and forth tends to make proofs longer, since we don't have a large library of theorems for it like we have for A =/= B.   And A e/ B would be used too seldomly to justify a large library. My use of A e/ B has mainly been restricted to "display purposes" to make an occasional final theorem look prettier, such as in sqr2irr.

As for Poly1 vs. Poly, I haven't used either and lack enough experience to comment.

david-a-wheeler:
If Poset is preferred over PosetRel then that should be expressly documented.  At least document in PosetRel's comments that it is deprecated (and that Poset is preferred instead).

I disagree about A e/ B. I think it's inconsistent that we have good support for A =/= B but not for A e/ B.  There's no fundamental reason for a lack of a large library of theorems for A e/ B; we just don't have a large library because we don't have a large library :-).  It would be easy to add more theorems about A e/ B, after all, and I think A e/ B is pretty common claim.

benjub:
Regarding A e/ B versus -. A e. B, there is indeed a choice to be made (and documented, so that we can be consistent and stick to it).  I would tend to agree with DAW (though not strongly).  An argument for having more support for =/= than for e/ is that equality is the only logical predicate, but one can also point out that = and e. are the only two predicates in the language of the theory (the only two non-defined predicates). I contributed a few basic lemmas in this direction in July:
for equality: neii, neir, nesym, nesymi, nesymir
for membership: neli, nelir

nmegill:
Some years ago I did a big conversion to =/= after noticing that many proofs would be reduced in size by eliminating extra steps for the "-." part, such as an extra notbid step.  "Necessity" (shorter proofs) is what drove A =/= B to become widely adopted.

In proving thousands of theorems, I haven't observed a similar situation with -. A e. B.  Any savings with A e/ B tend to be cancelled by conversions back to -. A e. B, making proofs longer instead of shorter.  I've rarely if ever felt "I wish I had an A e/ B library theorem for this".

People are free to use A e/ B when they see a benefit.  Library theorems are trivial and can be added as the need arises, with names analogous to the A =/= B theorems.

PR #535, opened 2018-09-23
Let sbcom3 unfold its value

wlammen:


wlammen:
@nmegill Has somebody found a value in pm11.07? The comment says it is bound to be deleted.

nmegill:
@wlammen I was going to try to decipher *11.07 in PM but then forgot about it.  On Sep 16, Benoit brought up his interpretation, which I want to think about, so the issue is still of interest.  Maybe change the "delete" date to 16-Sep-2019.

PR #537, opened 2018-09-24
Latest additions

tirix:
Note I've moved David's definition of the signum ` sgn ` into my mathbox so that I could use it for Descartes' Rule of Signs.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -377,6 +377,7 @@
 "4syl" is used by "pf1subrg".
 "4syl" is used by "pgpfac1lem1".
 "4syl" is used by "pgpfaclem2".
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Moving into your mailbox is fine short-term, but since there are now multiple users I think it should move into the main body. Our usual rule is that once there are multiple users, it should move into the main body.

tirix:
Sure, moving it to the main body would indeed be the best in the long term. 
I suppose I might add more generic theorems using it in the course of proving Descartes' Rule of Signs (the future will tell) - then maybe they can all be moved at once once it is complete.

PR #541, opened 2018-09-25
Add sbco4

jkingdon:
This is both set.mm and iset.mm and is a theorem suggested by Norm Megill on the mailing list.

I haven't yet spent any effort on getting rid of the distinct variable constraint between w and v. At least in the iset.mm case, it might be fairly difficult or relatively easy (I went through a similar set of problems with sbco2). For set.mm there are more techniques available but I haven't done any investigation there either.

It seemed better to send what I have so far, however.

The commit about sbid2 is iset.mm only and is bringing iset.mm in line with set.mm with respect to `F/` (and that allows the same `sbco4` proof to work for both).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -13324,10 +13324,18 @@ $)
     BDZABCFZARSPBCFZAAPBCGTPAABCHABIJKQASEBPSAABCLMNO $.
 
   ${
-    sbid2.1 $e |- ( ph -> A. x ph ) $.
+    sbid2h.1 $e |- ( ph -> A. x ph ) $.
     $( An identity law for substitution.  (Contributed by NM, 5-Aug-1993.) $)
-    sbid2 $p |- ( [ y / x ] [ x / y ] ph <-> ph ) $=
+    sbid2h $p |- ( [ y / x ] [ x / y ] ph <-> ph ) $=
       ( wsb sbcof2 sbh bitri ) ACBEBCEABCEAABCDFABCDGH $.
+  $}
+
+  ${
+    sbid2.1 $e |- F/ x ph $.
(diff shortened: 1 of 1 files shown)


nmegill:
Nice work.  To eliminate $d v w, you can make (web page) step 16 of sbco4 your lemma:

 ( [ x / v ] [ y / x ] [ v / y ] ph <-> [ x / w ] [ y / x ] [ w / y ] ph )     (new sbco4lem)

Then chain this with a dummy variable t in the middle:

... (steps from old sbco4lem) ...
 ( [ y / u ] [ x / v ] [ u / x ] [ v / y ] ph <-> [ x / v ] [ y / x ] [ v / y ] ph ) (was sbco4lem, now in main proof)
 ( [ x / v ] [ y / x ] [ v / y ] ph <-> [ x / t ] [ y / x ] [ t / y ] ph )     (new sbco4lem)
 ( [ x / t ] [ y / x ] [ t / y ] ph <-> [ x / w ] [ y / x ] [ w / y ] ph )   (new sbco4lem)

then apply 3bitri to the above to isolate v from w - thankfully $d isn't transitive :) - and obtain the final sbco4 without $d v w.

Norm

jkingdon:
Ah, thanks for the suggestion on eliminating `$d v w` . The only part I'm unsure about is whether to inline `sbco4lem` (as it is now called, might get renamed to `sbco4lem1` if there are two lemmas) into the main proof or not. I'll see what things look like and see if I have an opinion (Norm of course has the final call, so if he wants to insist I'll make that happen).

jkingdon:
Follow up pull request is https://github.com/metamath/set.mm/pull/543

PR #542, opened 2018-09-26
Add ifbieq1d and minimize 85 theorems with it.

jamesjer:
I would like to suggest adding ifbieq1d.  85 theorems can be minimized with it, resulting in a net reduction in size of set.mm of 1204 bytes.  That includes the comment.  Without the comment, there is net reduction of 1314 bytes.


nmegill:
I like it.  Thanks.

I added your name to the contributor list at the top of set.mm.

Norm

PR #543, opened 2018-09-26
Remove distinct variable constraint from sbco4

jkingdon:
This is for both set.mm and iset.mm (same proof both places).

The technique, suggested by Norm Megill, is to add a lemma sbco4lem2 which lets us use an additional temporary variable to insulate v from w (similarly to, say, dfsb7 ).

The only real question in my mind is whether to keep `sbco4lem1` as a separate lemma (advantages: it is concisely stated and lets the reader follow its proof without being distracted by the variable `t`) or inline it into the proof of `sbco4` (advantages: avoids an extra theorem in the theorem list, avoids the cognitive overhead of `sbco4lem1` being completely redundant once `sbco4` is proved).

I would somewhat lean towards keeping `sbco4lem1` separate, which is how this pull request is written now, but am interested in hearing other people's opinions.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 25-Sep-2018
+$( iset.mm - Version of 26-Sep-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm (with updates since then, including copying entire theorems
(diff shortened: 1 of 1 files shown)


jkingdon:
Travis is failing because either we need to inline `sbco4lem1` (as discussed above), regenerate `discouraged` (glad to do that but I've never done that so a pointer to the right part of the documentation would be helpful), or not mark it discouraged.

david-a-wheeler:
I have not looked at this in detail, but in general I recommend keeping lemmas separate.  Breaking things down makes each theorem easier to follow, and the lemmas can sometimes be reused in surprising ways. Regenerating the discouraged list is easy.

david-a-wheeler:
Okay, then I would suggest still leaving it as a separate lemma with the discouraged mark as you've done.

david-a-wheeler:
One way to regenerate the (updated) discouraged) file is to run this on a Unix-like system:

````
metamath 'read set.mm' 'set width 9999' 'show discouraged' quit | grep '^SHOW DISCOURAGED.' | sed -E -e 's/^SHOW DISCOURAGED.  ?//' | LC_ALL=C sort > discouraged
````

david-a-wheeler:
Oh, I didn't realize that the inlined version would be so small.  8 steps is trivial; even 2+2=4 is longer.  In that case, I think this should be inlined and NOT have a separate lemma.

I still think that we should try to break down theorems into smaller chunks so that (1) each theorem easier to follow (by not being very long), and (2) the lemmas can be reused.  But #2 doesn't apply (because it's leading to the more general form), and #1 doesn't really apply because the inline version is absurdly short.  Clearly there's a point of diminishing returns, and I think that point has been crossed here :-).

david-a-wheeler:
More clearly: I now think that sbco4lem1 should be inlined, because it's completely redundant once you get to the main proof, and the main proof even when inlined is so short that breaking it down doesn't seem to really buy anything.

wlammen:
> ... and 313 have "(New usage is discouraged.)"

3 of them are from me (ax12lem*). There is/was good reason to do so. For one, during active development, these lemmas should be easily modifiable, without being locked by random callers outside the scope of interest. After things settled, 2 of them are now just specializations of later theorems, and the 3rd, ax12lem3, is easily recoverable from the more flexible dveeq2.

In my eyes, these are valid reasons to keep the tag.

Wolf

jkingdon:
Although I still am interested in the question of how fine-grained lemmas should be, this doesn't seem to be a particularly good test case. That's probably most strongly because `sbco4lem1` was a special case of `sbco4` (and in fact that's why I bothered with `New usage is discouraged` - I wanted the minimizer or any similar tools to pick `sbco4` not `sbco4lem1` if it had the choice).

As for length of the proofs, at least for me that isn't as convincing a reason to inline. Although number of steps is one way to measure it, there is also total number of symbols or total number of variables or however else we might measure what is involved in getting the proof into one's head. Very much reminds me of debates people have in software style around the length of a method/procedure/function/whatever your language calls them - https://www.cqse.eu/en/blog/the-real-benefits-of-short-methods/ is as good a citation as any.

Anyway, there's a consensus in this case so I've pushed up the change to inline `sbco4lem1`.

nmegill:
I agree with Jim and Wolf that it is a good idea to put the "(New usage is discouraged.)" tag in lemmas that are instances of the final theorem.  (I think this happens mostly in early set.mm and becomes somewhat rare later on.)  In general I would say if there is a specific reason to avoid the lemma (including being under active development as Wolf mentions) then go ahead and use the tag, but in most cases the tag shouldn't be needed.  In particular I don't think we want to blindly add the tag to all lemmas, as it would dilute its usefulness in communicating that there is a reason not to use the lemma.

An example of a lemma that recently became useful for 2 theorems is ~ tosglblem where I indicated both theorems in the comment.  Actually, I broke this out from the original ~ tosglb proof because I needed it to fix the ~ xrsclat proof as part of the df-riota revision.

As for whether to in-line a lemma or not, I don't have a good rule of thumb.  In the case of sbco4, I think the in-lined lemma is easy to see as a block if you pay attention to the indentation levels.   Tools like Thierry's structured proof display may help also.  Going to an extreme of breaking things down could even be detrimental because you can't see everything at once but have to flip back and forth between pages.  At the other extreme, there are are a number of very large proofs in set.mm with varying degrees of readability, some of which really should be broken down into smaller lemmas.  A rough rule I've often followed is to make a lemma correspond to one English sentence (or independent clause in a sentence) in a textbook proof.  Even that can sometimes lead to proofs a hundred or more steps long, and I've sometimes broken a single sentence into a dozen lemmas.

Guidelines for structuring programming code into small pieces may or may not be applicable depending on the situation.  In particular, a computer-verified proof won't have any bugs, so partitioning for easier unit tests isn't relevant.  Four goals I can think of are making the proofs palatable for humans to read, identifying subproofs that might be useful as library theorems. providing proofs of a size that isn't too big for our tools to handle easily and quickly (some very large proofs in set.mm have this problem), and minimizing the file size of set.mm.  These goals can conflict with each other.

PR #544, opened 2018-09-26
rm unnecessary variable constraints from 2sb6rf

wlammen:


nmegill:
@wlammen It seems that the previous merge caused a conflict.  Date perhaps?

wlammen:
@nmegill  ok, i revert it to yesterday

wlammen:
@nmegill No executing rewrap on my side caused the problem

PR #547, opened 2018-09-28
prove 2sb6rf with minimal variable constraints

wlammen:
Lemma 2sb6rflem2 could be slightly shortened by 19.8ad in DAW's mathbox. I don't know whether this is sufficient to include it in main. The effect here is small, but I remember I had looked for it in the past occasionally.


david-a-wheeler:
I created 19.8ad for use, so I'd love to see it moved into main & used.

benjub:
I think it's more than enough reason to move 19.8ad to the main part.
Maybe add to its comment: "Also, dual form of ~ sps ."  And also add to
19.8a's comment that it is dual to ~ sp (and similarly add these facts to ~
sp and ~ sps; I think this type of links from one theorem to another is
useful).

By the way: all this work by WL it great!

wlammen:
@benjub Thanks for the kudos, Benoit

PR #548, opened 2018-09-28
Discourage depending on the Kuratowski definition for ordered pairs

david-a-wheeler:
We use the Kuratowski definition for ordered pairs, but
there are many other ways to define ordered pairs.
For example, the Kuratowski definition is very inconvenient in
New Foundations theory because it is not type-level, so
the Definition from [Rosser] p. 281 is often used instead in that context.

This commit uses "(New usage is discouraged.)" markings to discourage
depending on the particular definition/construction of ordered pairs.

This way, most theorems that use ordered pairs will not depend
on this particular Kuratowski construction, but will instead only
depend on more abstract properties of ordered pairs.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4611,6 +4611,7 @@
 "df-nv" is used by "isnvlem".
 "df-nv" is used by "nvss".
 "df-oc" is used by "ocval".
+"df-op" is used by "dfopif".
 "df-ph" is used by "isphg".
 "df-ph" is used by "phnv".
 "df-pjh" is used by "pjhfval".
@@ -4715,6 +4716,34 @@
 "dfiop2" is used by "hoico1".
 "dfiop2" is used by "hoico2".
 "dfiop2" is used by "hoif".
+"dfop" is used by "elop".
+"dfop" is used by "funopg".
+"dfop" is used by "op1stb".
+"dfop" is used by "opeqpr".
+"dfop" is used by "opeqsn".
+"dfop" is used by "opi1".
+"dfop" is used by "opi2".
+"dfop" is used by "opid".
+"dfop" is used by "relop".
+"dfop" is used by "uniop".
+"dfop" is used by "xpsspw".
+"dfop" is used by "xpsspwOLD".
+"dfopg" is used by "0nelop".
+"dfopg" is used by "dfop".
+"dfopg" is used by "gruop".
+"dfopg" is used by "opnz".
+"dfopg" is used by "opth".
+"dfopg" is used by "opth1".
+"dfopg" is used by "opwf".
+"dfopg" is used by "rankopb".
+"dfopg" is used by "tskop".
+"dfopg" is used by "wunop".
+"dfopif" is used by "dfopg".
+"dfopif" is used by "nfop".
+"dfopif" is used by "opeq1".
+"dfopif" is used by "opeq2".
+"dfopif" is used by "opex".
+"dfopif" is used by "opprc".
 "dfpjop" is used by "elpjch".
 "dfpjop" is used by "elpjhmop".
 "dfpjop" is used by "elpjidm".
@@ -5690,6 +5719,7 @@
 "elnpi" is used by "prn0".
 "elnpi" is used by "prnmax".
 "elnpi" is used by "prpssnq".
+"elop" is used by "relop".
 "elpaddatiN" is used by "osumcllem7N".
 "elpaddatiN" is used by "pexmidlem4N".
 "elpclN" is used by "pclfinclN".
@@ -10919,6 +10949,14 @@
 "opelreal" is used by "axrnegex".
 "opelreal" is used by "axrrecex".
 "opelreal" is used by "ltresr".
+"opeqpr" is used by "relop".
+"opi1" is used by "opth".
+"opi1" is used by "opth1".
+"opi2" is used by "elvvuni".
+"opi2" is used by "opeluu".
+"opi2" is used by "uniopel".
+"opid" is used by "dmsnsnsn".
+"opid" is used by "funopg".
 "opidon" is used by "opidon2".
 "opidon" is used by "rngopid".
 "opidon2" is used by "exidreslem".
@@ -11753,6 +11791,7 @@
 "recmulnq" is used by "recrecnq".
 "recrecnq" is used by "reclem2pr".
 "reexALT" is used by "cnexALT".
+"relop" is used by "funopg".
 "relrngo" is used by "iscrngo2".
 "relrngo" is used by "isdrngo1".
 "relrngo" is used by "isrngo".
@@ -14627,6 +14666,7 @@ New usage of "df-nq" is discouraged (5 uses).
 New usage of "df-nr" is discouraged (24 uses).
 New usage of "df-nv" is discouraged (2 uses).
 New usage of "df-oc" is discouraged (1 uses).
+New usage of "df-op" is discouraged (1 uses).
 New usage of "df-ph" is discouraged (2 uses).
 New usage of "df-pjh" is discouraged (2 uses).
 New usage of "df-pli" is discouraged (2 uses).
@@ -14669,6 +14709,9 @@ New usage of "dfch2" is discouraged (0 uses).
 New usage of "dfcnqs" is discouraged (4 uses).
 New usage of "dfhnorm2" is discouraged (3 uses).
 New usage of "dfiop2" is discouraged (3 uses).
+New usage of "dfop" is discouraged (12 uses).
+New usage of "dfopg" is discouraged (10 uses).
+New usage of "dfopif" is discouraged (6 uses).
 New usage of "dfpjop" is discouraged (4 uses).
 New usage of "dfsup2OLD" is discouraged (1 uses).
 New usage of "dfsup3OLD" is discouraged (0 uses).
@@ -15088,6 +15131,7 @@ New usage of "elnlfn" is discouraged (4 uses).
 New usage of "elnlfn2" is discouraged (2 uses).
 New usage of "elnp" is discouraged (5 uses).
 New usage of "elnpi" is discouraged (4 uses).
+New usage of "elop" is discouraged (1 uses).
 New usage of "elpaddatiN" is discouraged (2 uses).
 New usage of "elpaddatriN" is discouraged (0 uses).
 New usage of "elpclN" is discouraged (1 uses).
@@ -16675,6 +16719,10 @@ New usage of "opelopab4" is discouraged (0 uses).
 New usage of "opelopabsbOLD" is discouraged (4 uses).
 New usage of "opelreal" is discouraged (8 uses).
 New usage of "opelresiOLD" is discouraged (0 uses).
+New usage of "opeqpr" is discouraged (1 uses).
+New usage of "opi1" is discouraged (2 uses).
+New usage of "opi2" is discouraged (3 uses).
+New usage of "opid" is discouraged (2 uses).
 New usage of "opidon" is discouraged (2 uses).
 New usage of "opidon2" is discouraged (1 uses).
 New usage of "opsqrlem1" is discouraged (0 uses).
@@ -17033,6 +17081,7 @@ New usage of "reclem4pr" is discouraged (1 uses).
 New usage of "recmulnq" is discouraged (3 uses).
 New usage of "recrecnq" is discouraged (1 uses).
 New usage of "reexALT" is discouraged (1 uses).
+New usage of "relop" is discouraged (1 uses).
 New usage of "relopabVD" is discouraged (0 uses).
 New usage of "relrngo" is discouraged (6 uses).
 New usage of "renicax" is discouraged (0 uses).
--- a/mmset.html
+++ b/mmset.html
@@ -6066,6 +6066,8 @@ Analysis, Academic Press, New York (1972) [QC20.7.F84.R43].  </LI>
 <LI><A NAME="Rosenlicht"></a> [Rosenlicht] Maxwell Rosenlicht, <I>Introduction
 to Analysis,</I> Dover Publications, Inc., New York (1986) [QA300.R63]. </LI>
 
+<LI><A NAME="Rosser"></a> [Rosser] Rosser, John B., <I>Logic for Mathematicians</I>, Dover Publications, Mineola, N.Y. (2008) [BC135.R58 2008]. </LI>
+
 <LI><A NAME="Rudin"></A> [Rudin] Rudin, Walter, <I>Principles of Mathematical
 Analysis,</I> McGraw-Hill, New York, second edition (1964) [QA300.R916
 1964].</LI>


david-a-wheeler:
I don't know what you'll think about this, but I figured it was easier to create this proposed change so it could be discussed instead of just having an abstract discussion.

I'm used to writing computer programs, and it's generally wise to hide abstractions if you know there are alternatives.  That is *clearly* the case for df-op... so I thought I'd take a stab at doing the abstraction.

david-a-wheeler:
If we do this, then we should probably move most of the theorems that *do* depend on the construction details together (they're the ones being allowed to access the internal details).  But there's no point in moving theorems until there's an agreement that we want to hide the construction of ordered pairs.

david-a-wheeler:
Oh I definitely do not want it pure in that sense. That would be ridiculous, because it would require too much work. For example, noticed that opex is NOT discouraged for use. I intend for it to stay available for use.

The idea is to try to not leak what most of the things do not depend on, but go ahead and guarantee properties that are extremely useful for ordered pairs.  It's true that strictly speaking opex will prohibit some definitions of ordered pairs, but I suspect that we do not want to ever use any definition of ordered pairs that does not provide the guarantee of Opex.... so let us not try to prohibit it.

If you want an analogy, I might point to the example that many sort libraries guarantee that they are stable sorts. It's true that guaranteeing that a sort is stable does limit the set of permitted algorithms, but adding a few guarantees makes the library much easier to use, while still providing some flexibility on exactly what algorithm is used.

There is always a trade-off between information hiding and other useful attributes.

My goal in this commit is to try to hide some information to provide some abstraction, without being ridiculous. I'm not sure if I actually reached that goal, which is why I would very much like comments and frankly this is something of an experiment. But I think this experiment actually worked out, I think it hides the exact definition while still being useful and not too crazy to use in practice.

david-a-wheeler:
The idea of this commit is to hide the detail from most proofs that ordered pairs are represented as { { A } , { A , B } }.

It does NOT try to hide all assumptions.  In particular, too many proofs depend on the assumption that proper classes in a pair produce an empty set. Therefore, I do not try to hide that. So opprc1, opprc2, and 0nelop continue to be available. Since the empty set is a class, and in axiomatic set theory is everything ends up being just a set, it should be unsurprising that a pair is always a set. And so on. This means that not all possible definitions one might dream of are acceptable, but we still hide some details that most proofs should not depend on.

It's a compromise, but I think it's a reasonable one.  It's okay to make stronger guarantees, if they make it easier for later proofs.

nmegill:
Here are some objections.  I would also like to hear what @digama0 thinks.

1. We are not doing anything different from many (most?) set theory books, which don't try to hide the details of the definition.  For example, textbook theorems tied directly to the Kuratowski details are op1stb (Suppes), op2ndb (Monk), and rankop (Kunen).  One purpose of set.mm is to document published literature.

2. For set theory beginners, playing around with the df-op details to find interesting relationships can be fun (at least it was for me) and a good way to get exposure to some early concepts.  I think this should be encouraged, which it won't be if the user has to constantly override the discouragements and wade through the warnings that will litter the screen.

3. I can't think of any theorem where accidentally using df-op details has produced a misleading result.  I also can't think of any case where a definition-dependent result was used accidentally when opth would sufiice.  This could be determined definitely by scanning the uses of definition-dependent theorems, but until then it would seem like a straw man argument.

4. Perhaps contrary to common belief, opth is insufficient for certain standard results.  For example, it is impossible to compute the rank of an ordered pair from just opth (see rankop via rankopb), and I think also some theorems related to universes (wunop, tskop, gruop).

Overall, I think "New usage is discouraged" tags on these would be overkill.  I don't see an advantage, and they will just make life a little more inconvenient.

Here is a compromise proposal:  instead of a discouraged tag, just add a comment saying that the theorem depends on the details of the Kuratowski definition.  We already do that in a few theorems such as ovprc.

david-a-wheeler:
That's a reasonable compromise.  Perhaps some note like this: "Note that this is directly tied to the Kuratowki definition of ordered pairs".  That note would be added to the theorems that this commit discouraged (other than the original definition) and also to at least op1stb (Suppes), op2ndb (Monk), and rankop (Kunen).

david-a-wheeler:
I'd also love to hear @digama0 's thoughts.

david-a-wheeler:
At the moment I plan to create another branch with additional comments on a few theorems, per the discussion above.  But I'll wait to see if there are further comments first; then I'll close this branch.

digama0:
I think that hiding the definition of ordered pairs is on a slippery slope, because you can say similar things about most definitions. As pointed out, opth does not give you enough information to prove rankop, wunop or opex; these are the sorts of theorems that depend on "where" a set is situated in the universe and not just what its universal property is. Then again, a sometimes leaky abstraction is more or less what "new usage is discouraged" means - it doesn't mean new usage is forbidden, but you should have a good reason to do so.

The argument that new users might want to play around with e.g. proving opth for themselves doesn't seem to be a big deal; users will not be bugged about this unless they contribute to set.mm, although I guess they will have to call up df-op explicitly in a proof assistant since usually these discouraged theorems are not subject to automatic theory searches. But that seems fine to me.

While I can see good arguments for both sides, I am inclined to follow tradition and mark these as discouraged theorems. (Tradition, in the sense that every book on the subject tends to take a moment to remark about the idea of hiding the definition after opth is proven.)

nmegill:
@benjub  _By the way, the main problem I see with the Kuratowski definition of a couple is not the type-level issue, but the fact that it is not suitable for classes._

What definition do you propose?  The definition implied by opthprc requires cross-products, which in turn requires ordered set pairs of the Kuratowski type.  Do you have a definition that bypasses this and can be used everywhere?

How would you define functions with your class version of df-op?  Would these functions allow proper classes as arguments?

_There are natural definitions for which opth is true without sethood hypotheses, and for which the statement of opex becomes more natural: a couple of sets is a set._

I think that not having unconditional opex would make many proofs significantly longer and more difficult.  It is similar to the discussion sometime back about Alexander van der Vekens proposed df-afv in which out-of-domain function values evaluate to _V.

_I also understand Norm's arguments (even though I find op1sta, op1stb, op2nda, op2ndb very unnatural_

I agree about op1stb and op2ndb.  What do you find unnatural about op1sta and op2nda?  What would be your _version_ of these?

_, and ovprc very unnatural and dangerous in the sense that it could prove misleading results)._

What misleading results do you mean?  Do you have examples?

_And of course, rank related matters will be linked to a concrete definition (though not in an essential way for non-contrived definitions)._

I don't understand what you mean.  Won't this prevent the purity you seek?

nmegill:
@digama0 _I think that hiding the definition of ordered pairs is on a slippery slope, because you can say similar things about most definitions._

Yes, to be consistent we would need to apply the philosophy everywhere.

An important example is the definition of function.  There is an abstract idea behind it, a mapping of one set to another, and then there is the concrete definition we use, which is a collection of ordered pairs.  There are a very large number of theorems that are dependent on the ordered-pair definition for their statement or proof. 

Similarly we have relations, cross products, domains, ranges, restrictions, images, etc. all of which have many theorems derived from their ordered-pair constructions, and thus aren't "pure" embodiments the abstract idea behinds them.  Would all of these have a "discouraged" tag?

digama0:
> Similarly we have relations, cross products, domains, ranges, restrictions, images, etc. all of which have many theorems derived from their ordered-pair constructions, and thus aren't "pure" embodiments the abstract idea behinds them. Would all of these have a "discouraged" tag?

I don't think so. While I think there is a solid case for function application being overly "specific" in its definition (in that it specifies some behavior that we would like to regard as "junk"), I'm not sure you can say the same about relations and cross products (which really are sets of ordered pairs). Domain and range, and image, should probably only be considered when the input is a relation. Functions are on the border; for much of set theory it is important that functions be identified with functional relations, but after that you kind of want to think of them as primitives as in type theory.

In short, this constitutes an analysis of "well typing" on class expressions. If an expression makes sense for any class input at all, because it is treated as a collection of elements, like `( A u. B )` or `( A X. B )`, then its definition should probably be "exposed" and all theorems are okay. If there is some precondition on valid inputs because they should be structured in some way, then any theorems that *directly* talk about these out of domain properties should be discouraged. Individual exceptions can be made to this, when we decide that a property of the definition is important enough to be used generally and any redefinition should preserve it, i.e. fvex or ndmfv. Theorems that use one of these "ill typed" theorems to e.g. remove a hypothesis need no discouraged marker, but of course they will appear in the discouraged file for using a discouraged theorem.

benjub:
@nmegill
I will try and put an alternate definition in my mathbox in the following
days.

You are right that not having unconditional opex would make many proofs
longer.  Actually, several proposals I would like to make to increase
"purity" would make proofs longer, and conversely, I understand that many
things I consider like "hacks" are useful to simplify proofs.  This reminds
me of Mario saying something along the lines that proof-formalization is in
some ways closer to engineering than to mathematics (if I recall correctly,
and of course without any value judgment).  This is a case of contradicting
goals, and I reiterate that establishing a list of goals, objectives,
priorities, etc. of the MPE would be useful (not only having it would be
useful, but also making it would be useful since it would force us to think
of the MPE as a whole).

I find op1sta/b and op2nda/b unnatural because... they depend on the
"Kuratowski hack".  I agree it's a bit tautological!

As for opvrc being dangerous: I'm not positive, but I could imagine someone
new to metamath thinking "What?  If I take "F" to be "union" (which is
indeed a binary operation on classes), and if I take A and B to be the
universal class, then opvrc tells me that the union of the universal class
with itself is the empty class???"  Of course, one cannot take F to be
"union"...

As for ranks, I mean that the specific concrete definition of a couple will
indeed influence the rank.  But it will influence it slightly, in amounts
irrelevant to the important theorems.  I have no example, but here is an
analogy: the specific model of computation (e.g. Turing machine with one or
many tapes) will change the speed of algorithms, but only in a quadratic
way, so important theorems (e.g. an algorithm being polynomial or not) will
not be affected by the chosen model of computation.

nmegill:
@digama0 _I don't think so. While I think there is a solid case for function application being overly "specific" in its definition (in that it specifies some behavior that we would like to regard as "junk"), I'm not sure you can say the same about relations and cross products (which really are sets of ordered pairs). Domain and range, and image, should probably only be considered when the input is a relation. Functions are on the border; for much of set theory it is important that functions be identified with functional relations, but after that you kind of want to think of them as primitives as in type theory._

When you say relations "really are" sets of ordered pairs, that isn't the attitude of Quine (Set Theory and Its Logic).  He introduces the concept of a relation in an abstract way on p. 21:  'z{xy : Fxy}w' for 'Fzw'.  He waits until p. 58 to introduce ordered pairs, then says "The principal virtue of ordered pairs is that they make it possible to _simulate_ relations within the theory of classes." (p. 59) (emphasis added).

The relation "<" in high-school algebra is not ordinarily thought of as a set of ordered pairs but as something more abstract, defined by rules such as transitivity but otherwise left unspecified.

So as we extend this philosophy about df-op, "<. 1, 2 >. e. <" will also need to be "discouraged", and so on.

I think this discussion boils down to a tension between strict formalism and Platonism (not sure if those are the right terms).  To the formalist, an ordered pair _is_ df-op, nothing more and nothing less.  All theorems about it are equally valid theorems of set theory, even op1stb.  This makes life simple and all proofs rigorous.  If someone adopts a different df-op2, to the formalist it will be a different definition, although it will have similar properties and many theorems will remain the same.   Once you get into what properties of a definition you are allowed to use, it starts to fall into the realm of philosophy, where opinions may subtly differ.

On @benjub's wish to have ordered pairs apply to proper classes:  while opthprc can be useful in some specialized applications, we should remember that class notation (to the formalist at least) is a fictional defined notation to make expressions easier to work with and that proper classes aren't actual things in ZFC.  It can be enlightening to take statements in class notation and fully expand them to primitive set theory language.  As a simple example, the class expression A = A ends up translating to the scheme ph <-> ph, which has nothing to do with set theory.

For the purpose of this pull request, I suggest that @david-a-wheeler use a fixed phrase in place of "(New usage is discouraged.)" that we can match to replace it with "(New usage is discouraged.)" with a trivial global substitution if later on we agree it's appropriate. That way his hard work of identifying "improper" uses of df-op won't be lost, and I hope that this compromise will be satisfactory as this debate is continued.

The original intent of David's "discouraged" proposal was a prohibition, which after some discussions we softened to use the word "discouraged", but at least in metamath.exe, in practice it is almost a prohibition.  Even if you override it, there are constant in-your-face warnings that interfere with everything you try to do.  I could add an option to 'set discouragements off' to suppress them, but I worry that as we put more discouragements where they serve a debatable philosophical purpose, it will start to dilute the intent of "(New usage is discouraged.)" where we really need it, such as avoiding ax-meredith and the like.

nmegill:
@benjub _I find op1sta/b and op2nda/b unnatural because... they depend on the "Kuratowski hack"._

I don't see that for op1sta.  It should work fine with any other df-op, as long as we accept the definitions of relation and doman.  Can you explain further?

benjub:
> I don't see that for op1sta. It should work fine with any other df-op, as
long as we accept the definitions of relation and doman. Can you explain
further?

You're completely right, my bad, I misread op1sta and op2nda.

david-a-wheeler:
> I think this discussion boils down to a tension between strict formalism and Platonism (not sure if those are the right terms). To the formalist, an ordered pair is df-op, nothing more and nothing less. All theorems about it are equally valid theorems of set theory, even op1stb. This makes life simple and all proofs rigorous. If someone adopts a different df-op2, to the formalist it will be a different definition, although it will have similar properties and many theorems will remain the same. Once you get into what properties of a definition you are allowed to use, it starts to fall into the realm of philosophy, where opinions may subtly differ.

In *practice* people & theorems often treat "ordered pair" as a special construct, not as a specific definition, and there are multiple ways that ordered pairs can be implemented.  There are even systems where ordered pairs are primitives.  It can be useful to focus on the properties that you care about, and try to avoid depending on accidental properties you don't care about (unless you actually do care about that).  If someone wanted to use a different definition of ordered pairs (as must happen in New Foundations), it'd be really nice if *most* theorems could be trivially reused.

Historically many mathematicians generally focus on *specific* proofs, and don't worry about how to "scale up to prove all of mathematics within a single framework".  Metamath's set.mm attempts to put everything in one place, which is a different matter.  I think information hiding / abstraction, a key approach to scaling up in software development, is a useful tool in set.mm for the same reasons.

> For the purpose of this pull request, I suggest that @david-a-wheeler use a fixed phrase in place of "(New usage is discouraged.)" that we can match to replace it with "(New usage is discouraged.)" with a trivial global substitution if later on we agree it's appropriate. That way his hard work of identifying "improper" uses of df-op won't be lost, and I hope that this compromise will be satisfactory as this debate is continued.

How about: "(Avoid depending on this detail.)"

benjub:
I hadn't seen opthprc, but that's what I had in mind (this is Anthony
Morse's definition).  I know that classes are only virtual in set.mm, but
it looks like a definition as in opthprc could be used for instance to
define functions as triples (A,G,B) where $G \subseteq A \times B$ such
that...

For discussions on definitions of couples (ordered pairs), see
McCarty, Scott, Reconsidering ordered pairs,
https://projecteuclid.org/euclid.bsl/1231081372
Kanamori, The empty set, the singleton, and the ordered pair,
http://math.bu.edu/people/aki/8.pdf

As for your discussion of formalism versus platonism, I would add the
distinction materialist/structuralist.  Of course, ZF is a material set
theory, so it's hard to push structuralist ideas too much inside it. See
for instance
McLarty, Numbers can be just what they have to,
https://www.jstor.org/stable/2215789?seq=1#page_scan_tab_contents

nmegill:
@david-a-wheeler _How about: "(Avoid depending on this detail.)"_

Sounds good to me.

nmegill:
@benjub  _I hadn't seen opthprc, but that's what I had in mind (this is Anthony Morse's definition)._ 

If you want to add this to the opthprc comment, feel free.  A reference would also be welcome if you have one.  (Our tradition on things that are referenced once is to state the reference in-line rather than an mmset.html bibliography entry, such as the C.A. Meredith article reference in theorem meredith.  But this isn't a hard and fast rule, and it is up to you.)

_As for your discussion of formalism versus platonism, I would add the distinction materialist/structuralist. Of course, ZF is a material set theory, so it's hard to push structuralist ideas too much inside it. See for instance McLarty, Numbers can be just what they have to, https://www.jstor.org/stable/2215789?seq=1#page_scan_tab_contents_

We do take the separation of the complex numbers from their construction seriously as you can see from their re-axiomatization after the construction.  (In the past we've had to resist attempts to use 0ncn, which would make many things more convenient.)

Going further, all of our theorems using extensible structures, such as those on groups, do not concern any particular construction but talk about properties that any member would have.  So I think that would almost qualify us as structuralist for more advanced topics.  We could in principle even have a hypothesis "Let CC be a member of the class of complex number systems" (where CC is a class variable) for every complex number theorem, dispensing with the construction aside from showing that one such system exists, but at a cost of more complexity and longer proofs, and I think our compromise is acceptable.

david-a-wheeler:
Based on this discussion, I've created a new pull request https://github.com/metamath/set.mm/pull/559 and I'm closing this one.

PR #550, opened 2018-09-29
shorten sbel2x

wlammen:
Thanks to the recent reworking of 2sb5rf, the proof is not only shortened, but the necessary variable constraints are reduced to a minimum.
I removed a note, that is
- obsolete now;
- is incorrect, as I think x and y must still be distinct. Otherwise I arrive at something like w=z -> Ew Ez ph <-> ph, which does not hold generally. I did not bother formally checking this, though.
@nmegill For the second reason, you might want to remove this comment from master as well.


wlammen:
Nothing more for now

PR #552, opened 2018-09-30
shorten sbid & 2exsb

wlammen:


benjub:
@wlammen It looks like the minimal dv condition for 2exsb is {(x,y), (x,z), (y,w), (w,z,ph)}, that is, y and z need not be distinct.  Indeed, in the current version, x and w need not be distinct, and they play symmetric roles.  It could be the case that both pairs cannot be simultaneously equal, but this is unlikely.  This works because ph is distinct from w, so also from x if x==w, and similarly it is distinct from z, so also from y if z==y.  Maybe you can prove this in metamath ?

wlammen:
@benjub hmm. If I for the moment identify z and y, then I can reorder the quantifiers on the right side, arriving after some simplifications at Ew Ey Ay Ax (x=w -> ph). Here Ey Ay annihilate (a6e), so I finally have Ew [x/w] ph. But w is not free in ph, so the expression further simplifies to ph. On the rhs the expression reduces to Ex ph. Both sides are not equivalent without F/ x ph.
I do not bother formally proving this idea is correct.
As for minimizing dv conditions, this is not an easy task. You have to provide a distinctor -. Ay y=z -> ...
version of 2sb6, for example. Please look at the proof of 2sb6rf, where I underwent this tedious task, to see how much effort this costs. 2sb6 had excessive many unnecessary dv conditions, so in this case I found the work worthwhile.
Even if I am convinced (which I am not here), that a dv condition can be dropped, I won't elaborate on this, unless the theorem is very fundamental.
But feel free to help out and add a (Revised by...) tag referring to you.

Wolf

wlammen:
It is not too difficult. The proof of 2sb6 requires y different from z only in line 2, where 19.21v can be replaced with 19.21t, so a proof with distinctor -. A y y = z is not too difficult

PR #553, opened 2018-09-30
mathbox-update with equsal-like theorems

benjub:
I used a theorem in WL's mathbox to prove more precise versions of equsal.  Maybe this was his original motivation for proving wl-equsal1t?

@WL I think that your wl-equsal1i has a shorter proof avoiding ax-7 and df-or, using my exlimiieq1 and exlimiieq2.  But I think the latter two are actually sufficient.


nmegill:
@benjub The comment under a header respects the back quote notation surrounding math expressions.  So if you use 

` |- ( ph <-> ps ) ` => ` | - ( ch <-> th ) `

instead of

|- ( PH <-> PS ) => |- ( CH <-> TH )

it will look a little nicer.  Could you do this, and then I will merge this?  Thanks.

Note that => is not a set.mm symbol, so we have to keep it outside of the back quotes.  You can use double back quote to escape the back quote in function value e.g. ` ( f `` x ) = 1 `.

nmegill:
Github erased my back quotes.  I meant:

\` |- ( ph <-> ps ) \` => \` | - ( ch <-> th ) \`

wlammen:
My daily work is programming, and I am used to think along the lines of interfaces.  I learned, that to be most useful, interfaces are orthogonal, complete and each offered function (or theorem in this context) should do just one thing at best.  They are usually accompanied by a set of convenience functions (theorems) adapted to frequent use cases.
This thinking lets me constantly look out for 'primitives', and wl-equsal1t is one, because it gives a criterion by which you can tell, whether an antecedent x = y is redundant.
Over the time I have become reluctant to request adding new theorems to main, unless I can show they have at least a couple of applications. Norman speaks of seven as a thumb of rule, if I remember right.
In this particular case, I added the theorem to my mathbox, as a reminder, because in a few instances equsal in deduction form would have come in handy.
BTW you should use (at)wlammen to address me. I will be notified by email then. It was pure accident that I found your remark.

Wolf

PR #556, opened 2018-10-02
Add another power set section, and epsilon section, to iset.mm

jkingdon:
These are the sections "Power class of union and intersection" and "Epsilon and identity relations". Both sections are short and need light intuitionizing:

- replace equality/biconditional with subset/implication three places
- remove dfid3 and dfid2


jkingdon:
@nmegill Did you get a chance to look at this one?

PR #559, opened 2018-10-06
Mark ordered pair definition details with (Avoid depending on this detail.)

david-a-wheeler:
Set.mm uses Kuratowski's ordered pair definition,
which is standard for ZFC set theory, but is very inconvenient to use
in some other systems (such as New Foundations theory)
and there are other ways to define ordered pairs.

One solution would be to make uses that directly depended on
the details of Kuratowski's ordered pair definition
"(New usage is discouraged.)".  However,
there was some controversy about doing that.

Here we mark some of the key items eposing the details with
"(Avoid depending on this detail.)".  This isn't "discouraged", but
it marks those items so that they could be discouraged in the future.


benjub:
@david-a-wheeler There is a typo: "is not is not" in a comment.

PR #563, opened 2018-10-06
Add Aristotelian logic to iset.mm

david-a-wheeler:
This commit adds proofs of Aristotelian logic to iset.mm.
Remarkably, once we copy alsyl from set.mm to iset.mm
we can reuse all the proofs of Aristotelian logic in set.mm.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -11552,6 +11552,12 @@ $)
   exintr $p |- ( A. x ( ph -> ps ) -> ( E. x ph -> E. x ( ph /\ ps ) ) ) $=
     ( wi wal wex wa exintrbi biimpd ) ABDCEACFABGCFABCHI $.
 
+  $( Theorem *10.3 in [WhiteheadRussell] p. 150.  (Contributed by Andrew
+     Salmon, 8-Jun-2011.) $)
+  alsyl $p |- ( ( A. x ( ph -> ps ) /\ A. x ( ps -> ch ) ) ->
+        A. x ( ph -> ch ) ) $=
+    ( wi pm3.33 alanimi ) ABEBCEACEDABCFG $.
+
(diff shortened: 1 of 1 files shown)


jkingdon:
That is indeed intriguing that the Aristotelian syllogisms work in intuitionistic logic (that is, do not entail case elimination or the other things which depend on excluded middle). To put it in a sweeping form, "the roots of intuitionistic logic are as old as the roots of classical logic". This may not be fair - Frege preceded Brouwer by a few decades, Cauchy and Dedekind preceded Bishop, etc - but it is interesting to look for precursors to the constructive mathematics of today.

david-a-wheeler:
I was frankly surprised. I feel like I've learned something!

david-a-wheeler:
To be fair, the law of the excluded middle that is not allowed in intuitionistic logic is also from Aristotle. Here is a citation from Wikipedia's article about the law of the excluded middle:
"The earliest known formulation is in Aristotle's discussion of the principle of non-contradiction, first proposed in On Interpretation,[1] where he says that of two contradictory propositions (i.e. where one proposition is the negation of the other) one must be true, and the other false.[2] He also states it as a principle in the Metaphysics book 3, saying that it is necessary in every case to affirm or deny,[3] and that it is impossible that there should be anything between the two parts of a contradiction."

jkingdon:
Agreed that something like "Aristotle invented intuitionistic logic" would be way over the top, and I don't know enough about the syllogisms (either in Aristotle directly, or the medieval development thereof) to know how they fit into logic more generally in that era. Part of what we are seeing here is just "a lot of classical logic theorems still hold in intutionistic logic" (as can been seen by how many proofs can be copied from set.mm without modification).

Issue #406, opened 2018-05-06
Flesh out exclusive-or in iset.mm


jkingdon:
We now have a definition (`df-xor`) and a number of theorems concerning exclusive or.

I'm tagging this as `good first issue` since some of it is definitely easy (and the rest probably is, but might involve a little more figuring out what the theorems should be).

Remaining known tasks are:

* bring `xorass` and `xorcom` over from set.mm (need new proofs but those should be straightforward)
* figure out what to do with `xordi`. Does `( ( ph /\ ( ps \/_ ch ) ) <-> ( ( ph /\ ps ) \/_ ( ph /\ ch ) ) )` hold or is this theorem, or some variation thereof, dependent on some propositions being decidable?
* go through xor related theorems in set.mm and see if there are any more which should be brought over, either as-is or with decidability conditions added.

Oh, and in case #405 isn't merged by the time you read this, the discussion above is assuming the xor related changes in 405.

This issue replaces #350

david-a-wheeler:
I did xorcom. xorass is harder; the approaches that are "obvious" to me don't appear to be valid under intuitionistic logic.

david-a-wheeler:
@digama0 : Are you sure `xordi` is intuitionistically provable?  @jkingdon `annimdc` and `xordidc` for *decidable* propositions, but not when there's no guarantee that something is decidable.

david-a-wheeler:
I'm fiddling with iset.mm in part to try to get my head wrapped around intuitionistic logic, which is decidedly different from classical logic.  As such, I'm not going to be the best person to find subtle proofs.

digama0:
Suppose `( ph /\ ( ps \/_ ch ) )`, then `ph` and one of `ps` and `ch` is true, suppose `ps` is true and `ch` is false (WLOG). Then `ph /\ ps` is true, and `ph /\ ch` is equivalent to `ch`, so `ph /\ ch` is false and hence `( ph /\ ps ) \/_ ( ph /\ ch )`.

Conversely, suppose `ph /\ ps` is true and `ph /\ ch` is false; then since `ph`, `ph /\ ch` is equivalent to `ch` so `ch` is false. Thus `ph` and `ps \/_ ch`.

digama0:
[...]

PR #564, opened 2018-10-07
Add anxordi to iset.mm

jkingdon:
- xorbi12d is analogous to anbi12d and similar theorems for other connectives. We also get xorbi1d and xorbi2d because those are used in the proof of xorbi12d .

- anxordi is distributivity of conjunction over exclusive-or, and is similar to xordi in set.mm except that it is stated in terms of the exclusive-or notation. The proof was sketched out by Mario Carneiro in https://github.com/metamath/set.mm/issues/406#issuecomment-427619926 and turned into a metamath proof by Jim Kingdon.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 6-Oct-2018
+$( iset.mm - Version of 7-Oct-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm (with updates since then, including copying entire theorems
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Cool.  I suspect anxordi should be in set.mm as well (if it isn't already).

jkingdon:
> Cool. I suspect anxordi should be in set.mm as well (if it isn't already).

Agreed. I thought about making that part of this pull request but I'll let others get some of the fun too. If you look at xordi in set.mm you'll see it is close but not quite the same.

david-a-wheeler:
Okay, I just created commit 6c81a88f4dd4fa1e8651967b417834435d69d95e  in branch add-anxordi to add anxordi to set.mm.

PR #565, opened 2018-10-07
add two inferences that minimize around 44 proofs

benjub:
Added the inferences iftruei and iffalsei which minimize around 44 proofs.


nmegill:
@benjub I'm guessing these probably "pay for themselves", but in the future it would be helpful to know how many bytes set.mm is reduced by.

PR #570, opened 2018-10-09
Additional auxiliary theorems for graph theory

avekens:
Changes in my Mathbox only: Additional auxiliary theorems for graph theory, especially about integers, hash function, words, singleton words, last symbol of a word.


nmegill:
@avekens Travis failed because the use of axia1 is now discouraged (because it duplicates simpl, which is the preferred theorem).  All you need to do is replace your "axia1" reference with "simpl" and resubmit.

avekens:
OK, I fixed it - I do not know how this came in there - maybe by "minimize"?

PR #572, opened 2018-10-10
Move some commands in .travis.yml into separate scripts

david-a-wheeler:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,18 +9,13 @@
 # Some of these verifiers also check for style,
 # so using multiple verifiers can also detect many different style problems.
 #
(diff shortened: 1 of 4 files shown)


david-a-wheeler:
Don't pull unless the Travis run passes, obviously.

I'm trying to pull out the mishmash of Travis commands into simple clean scripts that do one thing.  This is the first step, just pulling out the verification commands for metamath.exe and mmj2.  Once I get this working, I can extract some more things.  The eventual goal is for the Travis commands to be mostly short and obvious, with a bunch of reusable scripts in the "scripts/" directory that each do reasonable things.

PR #575, opened 2018-10-11
DO NOT MERGE: Intentionally put errors in set.mm and iset.mm

david-a-wheeler:
This should cause the Travis system to fail - make sure it does.


david-a-wheeler:
@nmegill : Good news, the tweaked Travis commands reject mm files with serious problems.  That's the expected outcome, but it's wise to verify it.

PR #578, opened 2018-10-11
Minor formatting cleanup: Remove trailing blanks on lines

david-a-wheeler:
Trailing blanks on a line are not visible, and that can cause
long-term problems.  Remove all trailing blanks in set.mm.


avekens:
Thanks a lot, David, for the cleanup. I didn't thought that it was so much. Maybe the removal of trailing blanks should be automated in metamath.exe: sometimes trailing blanks ar removed (I assume by "write source"), but not always, as can be seen now.

david-a-wheeler:
They probably should be automated and/or included in the CI test suite (they're easy to detect).  But fixing them is trivial; you did the hard part!

nmegill:
I thought they were removed by 'write source.../rewrap' but I may be mistaken, I'll have to check.  I think I will add a check to 'verify markup'.

They cause no fundamental problem, of course.  The minor annoyance is that they result in spurious git differences when the next person strips trailing spaces.

nmegill:
That said, do not let this issue distract you from proving new theorems. :)  It is by no means a big deal, hardly even worth worrying about.  Trailing spaces will eventually be removed by someone.  And congratulations @avekens on you proof!

PR #580, opened 2018-10-12
Add algebra helpers and "i2lines" to demo it

david-a-wheeler:
This commit adds to David A. Wheeler's mathbox a collection of
"algebra helpers" that I suspect will make it easier to do
a number of simple algebra operations on equations.
It also includes "i2lines", which demonstrates how to use them.

This is an experiment in a different approach, but since all theorems are
proven and no new definitions are added, everything is of course proven.


david-a-wheeler:
@benjub - Thanks! You're right, I should definitely use addcli, mulcli, subcli to reduce $e in the helpers.

In i2lines I was trying to keep the number of displayed lines shorter, but it clearly would be easy to simply make a statement on each line. I guess I don't really care either way. If anyone else has a preference, let me know. Otherwise I will probably just change it as suggested.

nmegill:
The theorem might be more generally useful in Mario's "deduction" format:  |- ( ph -> A e. CC ) ... |- ( ph -> (conclusion).  It's only in early CC that we use inferences with A e. CC etc. as hypotheses.  More advanced theorems often involve existential quantification (to emulate Rule C; see comment for exlimiv) and require library theorems in closed (or deduction) form, and we would need to apply dedth and friends to your theorems.  While turning A e. CC into an antecedent isn't too hard with dedth, making hypotheses like Y = ( A x. X ) + B become antecedents is somewhat awkward, and we tend to end up with proofs longer than they would be if we just proved the deduction form directly.

david-a-wheeler:
@nmegill : I agree, it'd be more general in "deduction" form, so that definitely should happen.  I think it'd be useful to have the non-deduction forms as well.  I imagine the "final" versions would be first proven in deduction form, and then the non-deduction forms would be proven from them.

As expected, the disadvantage of this approach is that (1) you have to create a lot of permutations of various theorems and (2) it only handles "simple" cases.  However, each permutation is short & easy to prove, and when you *do* have simple cases, the step is higher-level and thus clearer/easier.

The point of the experiment, of course, is to get a sense of whether or not it's worth it.  As long as we don't go crazy (i.e., only do a few operations) this is starting to look like it'll work out!  For "simple cases" the resulting proofs are shorter - and in my opinion they're easier to understand because they look a little more like human proofs.  E.g., "move + X from the RHS to the LHS" is easy enough to grasp.

The naming gets a little crazy :-).

david-a-wheeler:
I just posted another version that removes some of the $e's.  I didn't use deduction form (yet).

PR #582, opened 2018-10-12
mathbox update with alternate def for couples

benjub:
I added a definition of couples which works for classes (as promise on the google group). Details are in the header comment of the corresponding subsection. Notation can be changed.

The diffs outside of my mathbox are just artifacts due to line wrapping.


nmegill:
@benjub You need to update the 'discouraged' file.

<pre>
./metamath 'read set.mm' 'set width 9999' 'show discouraged' quit \
    | tr -d '\015' | grep '^SHOW DISCOURAGED.' \
    | sed -E -e 's/^SHOW DISCOURAGED.  ?//' | LC_ALL=C sort > discouraged

</pre>

PR #585, opened 2018-10-13
Add comraddd to Algebra helpers in DAW's mathbox

david-a-wheeler:
Remove some unnecessary $e hypotheses in the Algebra helpers
and modify the corresponding i2lines example.
This is in reponse to some helpful comments by Benoit (@benjub).
This is experimental work in David A. Wheeler's mathbox to
develop a higher-level and easier way to express simple algebra
manipulations of equations.


david-a-wheeler:
I proved comraddd directly.  Might it be better to prove 1 version, and then use that to prove the other?  What would be the best way to do that?  I presume it'd be best to prove deduction form first, and then create the others, but I'm not sure of the cleanest way to do that (it's probably obvious to everyone else, I think I'm sleepy this morning).

david-a-wheeler:
One approach: If I have the deduction theorem form, I can prove the version without the "ph ->" prefixes by prefixing each one with "T. -> " using `a1i`, using the deduction theorem form to prove the result, and then removing "T. ->" using `trud` .

nmegill:
@david-a-wheeler We usually call "the version without the ph" the inference form.  Your method will work, but I don't think it's necessary to prove the inference form - I think it would rarely be used, and it is trivial to derive if needed.  Of course it's your mathbox, and what you do there is up to you, and I may be wrong.

PR #589, opened 2018-10-14
add sp2 to minimize 9 proofs

benjub:
Added sp2, which was in my mathbox, to minimize 9 proofs.  Norm's script min.cmd gave the following:

Proof of "cbv1h" decreased from 119 to 112 bytes using "sp2".
Proof of "mopick" decreased from 321 to 311 bytes using "sp2".
Proof of "csbie2t" decreased from 196 to 187 bytes using "sp2".
Proof of "copsex2t" decreased from 432 to 419 bytes using "sp2".
Proof of "fundmpss" decreased from 1141 to 1134 bytes using "sp2".
Proof of "wfrlem5" decreased from 1217 to 1209 bytes using "sp2".
Proof of "frrlem5" decreased from 1228 to 1220 bytes using "sp2".
Proof of "mbfresfi" decreased from 4670 to 4657 bytes using "sp2".
Proof of "pm14.123b" decreased from 270 to 261 bytes using "sp2".

Hence a total of 84 bytes reduction, while

---------The proof of "sp2" (24 bytes) ends above this line.


nmegill:
I'll merge this one somewhat reluctantly, but counting bytes in an added proof vs. bytes reduced in the minimized proofs is not the criterion we use.  The criterion is "Does the new theorem reduce the net file size of set.mm?"  In this case, you added 271 bytes to free up 84 bytes, or a net increase in set.mm size of 187 bytes.  The "rules" allow ignoring the "Contributed by", but even with that the set.mm size increases.

From set.mm:
<pre>
Usually I will also automatically accept a _new_ theorem that is used to
the new theorem, not including the acknowledgment) decreases as a result.
</pre>

A secondary consideration is whether the theorem will be useful or interesting for other reasons, and I would say that probably not.  It will hardly ever used, so people will tend not to remember it for everyday use (unlike sp) and therefore will probably chain sp's anyway.  Since it is trivial to chain to sp's with syl, it doesn't save much proof writing time.  Therefore IMO its primary purpose would be to minimize proofs, and it doesn't do that well enough to reduce the size of set.mm.

Will it be used in the future enough to finally justify itself in terms of a net reduction of set.mm size?  No one can say for sure, but given only 9 uses out of 30K theorems it will be a very long time.

benjub:
@nmegill Ok, I understand.  I used the command "MM>show proof sp2
/compressed" to get the size of the proof.  Is there a command to get the
size to be considered here (comment - acknowledgments + statement + proof) ?

Another reason why I added sp2 is that it is in PM (as *11.1), or more
precisely, it is the analog of PM*11.1 in the metamath setting.  The
precise PM*11.1 is stdpc4-2, currently in AS's mathbox.  Since I use the
latter in my mathbox (ax7-pm2), maybe it is worth adding to the main part ?

nmegill:
_Is there a command to get the size to be considered here (comment - acknowledgments + statement + proof) ?_

No.  It is a difficult problem because there is no clear way to account for shared $e's, $d's, etc.  As a rule of thumb you typically need another 200 or so bytes for simple theorems (depending on the comment of course).  In all of set.mm, the average  (comment - acknowledgments + statement + proof)  is 421 bytes.  As you can see, a trivial theorem that saves a single proof step may need many dozens of uses to "pay for itsefl".  A theorem that saves a large subproof may need only 3 or 4 uses to do so.

_ Since I use the latter in my mathbox (ax7-pm2), maybe it is worth adding to the main part ?_

Our policy is to move up theorems that are used by more than one mathbox, so go ahead at your convenience.

PR #591, opened 2018-10-14
Add naming and conversion discussion to mmnatded.html

david-a-wheeler:
Improve the documentation in mmnatded.html by adding a more
detailed discussion about naming conventions and
how to do conversions.

### DIFF
--- a/mmnatded.html
+++ b/mmnatded.html
@@ -293,17 +293,14 @@ of Td in turn by eliminating conjuncts
 <A HREF="simp1.html">simp1</A>,
 <A HREF="simp2.html">simp2</A>, and
 <A HREF="simp3.html">simp3</A>).
-If the expression is a conjunction of conjunctions, it may sometimes
-take multiple steps to prove some of the hypotheses
-(to break the expression into its parts).
 If the expression has nested conjunctions,
 inner conjuncts can be broken out by chaining the above theorems
 with <A HREF="syl.html">syl</A>.
 (There are actually many theorems (labeled simp* such as
 <A HREF="simp333.html">simp333</A>)
 that break out inner conjuncts in one step,
-but rather than learning them you can just use simple sequences to
-prove them, and then let the
+but rather than learning them you can just use the chaining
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
I think some documentation like this is necessary.  I think it's correct, but if it's wrong, please fix or tell me what the problem is!

david-a-wheeler:
@digama0 @nmegill - If I got something wrong, let me know!

david-a-wheeler:
I'm okay with the phrase closed form, but if we do that, we will need to modify the conventions text to match. I would like to hear from Norm or Mario before we do that.

Can you give me precisely the sentences you'd like to see in that last part?

nmegill:
I agree with Benoit that "tautology form" should be "closed form", and "induction form" should be "inference form".  BTW "tautology" usually refers to propositional calculus, occasionally to logic in general, but never set theory.

For "If the expression is a conjunction of conjunctions, it may sometimes
take multiple steps to prove some of the hypotheses
(to break the expression into its parts)."  I might say instead, "If the expression has nested conjunctions, inner conjuncts can be broken out by chaining the above theorems with ~ syl.  There are actually  many theorems (labeled simp* such as ~ simp333) that break out inner conjuncts in one step, but rather than learning them you can just let the metamath.exe command "minimize_with" figure out the right ones needed to collapse your ~syl's."

Line 285: "eliminating conjunctions" -> "eliminating conjuncts"

david-a-wheeler:
Thanks for the comments!  Here's a new version that I *hope* reflects them.

nmegill:
Lines 296-298 are ones I meant to replace with my suggested text.  As it is now, it more or less repeats what's below it.  I would suggest removing lines 296-298.

In line 305, I don't know if the reader will know what a "simple sequence" is (sounds like a technical term) or that it refers to the chaining described above it.  Perhaps change "simple sequence" to "the chaining we just described"

david-a-wheeler:
Your wish is my command!  Changed as requested.

PR #592, opened 2018-10-15
Add ordinals to iset.mm

jkingdon:
Ordinals are (according to Crosilla) less important in constructive mathematics than classical mathematics, so it perhaps isn't surprising that most of this section didn't make it over.

However, we do have a few important pieces: the definition of ordinal from Crosilla (df-iord which replaces df-ord) and a few theorems which either still work as-is, or work with some changes.

Some parts of this section will probably need to be revisited and somehow re-proven (either for all ordinals, or at least for elements of 𝜔) once we get to Peano arithmetic. But for now add the parts which work without too much rework.


jkingdon:
@nmegill Did you get a chance to look at this?

PR #600, opened 2018-10-18
Minor fix in documentation of empty-surprise

david-a-wheeler:
Fix the documentation of empty-surprise by
removing an unintended "not" symbol.

My thanks to Gérard Lang for reporting the problem!


nmegill:
A technical detail on the comment of empty-surprise:  in ZF set theory, there is only one empty set (theorem zfnuleu).  You may want to check some literature examples, but I don't think it is customary to speak of "empty sets", even if the symbol (/) occurs multiple times in a formula.

PR #606, opened 2018-10-21
Add experimental "allsome" quantifier to DAW's mailbox

david-a-wheeler:


nmegill:
Not sure I'm warm to this, but it's your mathbox. :)  BTW I think the notation "A! x ph -> ps" will be confusing as part of complex wffs, but you can think about that later.

PR #609, opened 2018-10-22
Add Axiom of ∈-induction to iset.mm

jkingdon:
This is our replacement for the Axiom of Regularity.

The larger picture here is that might as well add ∈-induction because we'll want it. In particular, there are a lot of theorems in set.mm which are about delaying the use of Regularity, and some of those we have been having trouble intuitionizing. So as we keep exploring this, we'd like to have ∈-induction available in case we decide that's what we want.

As for "why ∈-induction?", as explained in Crosilla, regularity implies undesirable forms of the law of the excluded middle.

We don't yet have anything proved from ∈-induction, but
once we figure out well-foundedness (or proofs directly
from ∈-induction), we may be able to bring in some of
the theorems which had been using regularity (and/or
some of the ones proved without regularity in set.mm
but which we don't have yet in iset.mm).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 16-Oct-2018
+$( iset.mm - Version of 22-Oct-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm (with updates since then, including copying entire theorems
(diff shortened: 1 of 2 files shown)


jkingdon:
This was merged as part of #611 .

PR #614, opened 2018-10-28
Prove that there are no surprises in some cases (DAW's mathbox)

david-a-wheeler:
We earlier proved that there are, in some cases, "surprises"
for the novice.  Here we prove that those surprises cannot occur
in other situations (where "there exists" is specifically asserted
or where the allsome quantifier is used).  It also cleans up
some comments in the surprises and allsome sections.


david-a-wheeler:
It's actually my *mathbox*, not my *mailbox*.  :-).

PR #618, opened 2018-10-31
Fix year "1017" to be "2017" in a few places

david-a-wheeler:
Metamath was not around before the Norman Conquest.  This fixes
the dates so the years coincide with a different Norman.


digama0:
I recall Norm saying that this was deliberate, to avoid some theorems appearing on the recent list. @nmegill are you still doing this?

david-a-wheeler:
It should be okay to change them to 2017 now, though.  I assume that something from 2017 won't appear on the "recent" list :-).

nmegill:
@digama0 I am no longer doing this and thought I had fixed them all.  I don't even recall the ones David fixed - maybe it was an actual typo that propagated w/ copy/paste.

david-a-wheeler:
Thanks for merging!  I only noticed this because I sorted the theorems based on date; I discovered that some of them were older than I was expecting :-).

PR #619, opened 2018-10-31
Add some conjunction theorems to David A. Wheeler's mathbox

david-a-wheeler:


nmegill:
@david-a-wheeler Often we use "anabs..," (absorption) instead of "anandi...".  The latter is typically used when the structure resembles a distributive law like "andi" but with all conjunctions.  The distinction between the two isn't always clear or even matters that much, just letting you know our informal convention.  I'm fine with your labels if you like them.

david-a-wheeler:
Thanks for the merge!  I'd be happy to rename them to something clearer, and calling them "absorption" seems clearer too.  Any suggestions on the names I should use?

david-a-wheeler:
Also: What do you think about moving these theorems after the existing absorption theorems?

I noticed that these absorption theorems were missing when I tried to prove alsi-no-surprise - that theorem should have been super-trivial, but I had to spend a number of steps to do what this theorem can do in one step.  I suspect it would be useful elsewhere.

nmegill:
@david-a-wheeler Can it reduce the size of existing proofs?  That would provide an argument for moving it up.

david-a-wheeler:
I suspect it can reduce database size, but I also suspect that minimize will not be able to figure it out in many cases. I suggest that we figure out a good name for it first, and then I can try to figure out how to show if it can actually reduce sizes.

nmegill:
@david-a-wheeler These can be proved more simply by chaining anandi(r) and a df-3an variant e.g. for anandisr:

    ( w3a wa 3anan32 anandir bitr2i ) ABCDACEBEABECBEEABCFACBGH $.

Regarding the name, the "s" suffix in e.g. anandirs means "eliminate a syl" and is often used to suffix theorems that manipulate antecedents like eqcoms.

I might have called your anandisr, anandisl with the names anandi3r, anandi3 to match the anandir, anandi pattern, where the "r" means "reversed" and the "3" suggests a 3an usage.    A small issue is that the lhs and rhs are reversed compared to anandi(r), and anandisr isn't quite an anandir variant since the distributed "ps" is in the middle of the rhs.

Norm

nmegill:
@david-a-wheeler _I could swap the lhs and rhs, would that be preferred?_

If we call them anandi* (which I prefer), it probably would be a good idea to swap them to match anandi, where the lhs is before distributing and the rhs is after.

david-a-wheeler:
Very good.  I'll create a separate pull request to do that.

PR #625, opened 2018-11-05
Add tfi to iset.mm

jkingdon:
This proof, although long, is fairly straightforward from
e.-induction. By contrast, the set.mm proof does not depend
on e.-induction (ax-reg in that context).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -36177,6 +36177,34 @@ $)
   $}
 
 $(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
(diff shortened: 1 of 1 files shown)


digama0:
I think you should be assuming as part of the definition that E. is founded on an ordinal. This is part of the definition in ZF, and it is the reason you usually don't need epsilon induction in ordinal proofs.

jkingdon:
As for definitions, I tried to prove some things about definitions related to well-foundedness, and progress was slow to non-existent. The existing definition of ordinal is taken directly from the cited source which itself claims that it is widely adopted (I'm not really qualified to evaluate this claim, but in the absence of citations saying otherwise....). I'm guessing we could prove things related to well-foundedness with that definition but, like I say, when I tried I didn't get very far.

PR #626, opened 2018-11-05
Expand "Transfinite induction" section in iset.mm

jkingdon:
This is part of the "Transfinite induction" section from set.mm. It does not include the theorems which don't hold in intuitionistic logic (those which rely on treating successor and limit cases separately).

In intuitionistic logic we only get the ones which rely on an induction step of "the property holds for all ordinals less than this one implies that it holds for this one".


digama0:
It sounds like you've hit a major roadblock, that is going to cause more problems going forward. Could you give an example of the kind of proof you don't see how to intuitionize?

digama0:
Here's my pass at "Ordinals (continued)":

* ordon: provable with your definition (should also be provable if you throw in E. Fr A in the definition)
* epweon: not provable since you aren't assuming We as part of the definition, but `epfron` should be analogously provable
* onprc: provable (and this is an important one, the Burali-Forti paradox). Follows easily from ordon
* ssordi, etc: provable (not sure why it's not proven... maybe ordon tripped you up)
* ordeleqon: not provable (kind of obviously)
* ordsson, ordss: provable, easy
* ssonprc: not provable, but interesting enough to intuitionize anyway. U. A = On -> A e/ V is provable, and B e. On /\ U. A C_ B -> A e. V is provable. (Why isn't df-pss stated so that the set difference is inhabited? If so, you could prove U. A C. On -> A e. V.)
* onuni, etc: provable, and important
* onint: provable if you assume A is inhabited
* onssmin: provable if you assume A is inhabited (this will substitute for most proofs by well foundedness)
* onminesb, onminsb: provable
* oninton: provable if you assume A is inhabited
* onintrab: provable

PR #628, opened 2018-11-08
Document wff symbol variations in conventions

david-a-wheeler:
Document in "conventions" that
we try to follow mathematical conventions, but in many cases
different texts use different conventions.
In those cases we pick some reasonably common convention and stick to it.
Provide specific details about the variables used to represent
arbitrary well-formed formulas (wffs), showing that there is a wide
variety in practice.


david-a-wheeler:
The information here is a combination of some things found by Norman Megill plus others that I found.

Issue #629, opened 2018-11-08
Expand ordinals in iset.mm


jkingdon:
This is based on https://github.com/metamath/set.mm/pull/626#issuecomment-436090878 but omits most theorems which are described there as not provable. It also omits theorems which have been proved since that comment was written.

* ~~`onprc: provable (and this is an important one, the Burali-Forti paradox). Follows easily from ordon.` (haven't spotted this proof yet. The set.mm proof of onprc depends on ordon and ordirr and we don't yet have the latter in iset.mm)~~ has been proved

* ~~`onint: provable if you assume A is inhabited` (looks like we need a variant of tz7.5 for inhabited rather than non-empty)~~ not provable

* ~~onint0: not in original comment but looks like in addition to onint it may need a bit more work~~ not provable

* ~~`onssmin: provable if you assume A is inhabited (this will substitute for most proofs by well foundedness)` set.mm proof uses onint~~ not provable

* ~~`onminesb, onminsb: provable` set.mm proof uses onint~~ not provable

* `oninton` (with non-empty changed to inhabited): This one I think can still be salvaged though. From the fact that it is inhabited you get that it exists, and is a subset of an ordinal x. It is an intersection of transitive sets so it is transitive, and of course all its members are members of x so they are transitive too. And E. Fr A falls to subsets.

* `onintrab, onintrab2: provable` set.mm proof uses oninton

* `onnmin`: May be able to replace trichotomy with irreflexibility. If B e. A and B e. |^| A, then |^| A C_ B so B e. B, violating ordirr.

* `onminsb` set.mm proof relies on trichotomy

* `oneqmin: provable if B is inhabited, and you replace -. x e. B with B C_ x`

* `onminex` falls as written because it implies onint, but it might be useful to keep the reverse direction for subsets that do have a minimum.

* `sucon: provable` The set.mm proof will work if we get `onprc`

* ~~ordsuc. The set.mm proof relies on case elimination (on whether A is a proper class).~~ Proved for the  `A e. _V` case in #641

* `ordpwsuc: not provable, but interesting. I think suc' A = ( ~P A i^i On ) may play a role later on, it might be worth having both notions. suc A C_ suc' A, and both are ordinals.` Not sure I have an opinion yet about introducing a notation for suc' but at a minimum we can prove a variant of ordpwsuc which has subset rather than equality in the consequent.

* ~~`sucelon`~~ see #641

* `onpsssuc: provable (for both suc's)`

[...]

PR #633, opened 2018-11-12
Change limit ordinal definition in iset.mm

jkingdon:
We'd expect at least the usual change of non-empty to inhabited. However, we take it one step further, and adopt a definition ( dflim2 ) that to be a limit ordinal an ordinal must contain the empty set, not just any element. Being inhabited might imply containing the empty set, but I was not able to find a proof so there isn't any particular problem in using the "contain the empty set" definition.

Adjust theorems which had been using df-lim to use dflim2 instead ( limeq , nlim0 , limord , limuni ). Add 0ellim and prove in terms of dflim2 .

Once we have ordunisuc2 (using the proof from #629 presumably) we'll also get `dflim4` but I'll leave that for a follow-up pull request.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 8-Nov-2018
+$( iset.mm - Version of 11-Nov-2018
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm (with updates since then, including copying entire theorems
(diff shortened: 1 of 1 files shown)


jkingdon:
Ah, yes thanks for the reminder about inhabited versus containing the empty set. Sort of smells like "a subset of a finite set is finite".

PR #655, opened 2018-12-02
Flesh out natural numbers in iset.mm

jkingdon:
This pull request should (unless something else comes up) complete the development of the natural numbers (omega) in iset.mm.

This is elnn , ordom , omelon2 , omelon , nnon , nnoni , nnord , omsson ,
peano2b , and nnsuc .

We now have pretty much everything from the "The natural numbers
(i.e. finite ordinals)" section in set.mm except for limit
ordinals - elom , limomss , nnlim , omssnlim , limom , and ssnlim .
We haven't done much with limit ordinals and it is not clear
they are as important in IZF as in ZF. But if we do want to add
them back later, they will need new proofs.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -36425,7 +36425,7 @@ $)
   ${
     $d A x $.
     $( A class is equal to its successor iff it is a proper class (assuming the
-       Axiom of Regularity).  (Contributed by NM, 9-Jul-2004.) $)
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@jkingdon : By the way - a big congrats on getting to this point in iset.mm!!

jkingdon:
Thanks @david-a-wheeler for the kind words. Always good for a reminder to take a moment to celebrate. Especially at a milestone like this.

PR #656, opened 2018-12-02
Add hyperlink from mmset.html to mmtopstr.html

david-a-wheeler:
Add a hypertext link from mmset.html to mmtopstr.html so that
viewers of the "front page" can easily discover the page explaining
algebraic and topological structures.

### DIFF
--- a/mmset.html
+++ b/mmset.html
@@ -297,6 +297,12 @@ SIZE=-1><I>15-Jan-2017</I></FONT>
 -->
 
     </LI>
+
+<LI> <A HREF="conventions.html">Algebraic and Topoologic Structures</A>
+    <FONT SIZE=-2 FACE="Comic Sans MS" COLOR=ORANGE>Added</FONT> <FONT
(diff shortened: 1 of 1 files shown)


nmegill:
@david-a-wheeler If you look at the current mmset.html on us2, it already has a link to mmtopstr.html called "Topological Structures".  It looks like I forgot to push it, sorry.  Also, yours links to conventions.html rather than mmtopstr.html.

I will do this:

1. I'll close out your pull request without merging.

2. I'll add "Algebraic" to the link description.  I won't change "Topological" to "Topologic" (which I've never heard before that I can recall).  "Topologic" does seem to be a legal English word, but Google says it's used 200 times less frequently than "Topological".  If you feel strongly about it, change it to "Topologic".

3. Feel free to make any other changes starting from my version.

Norm

david-a-wheeler:
@nmegill - that sounds perfect. I just noticed that there was no way to get to the new page from the front page in the develop branch... as long as that's fixed, that is great!

PR #659, opened 2018-12-03
Move anandi3 and anadi3r into main body

david-a-wheeler:
Move the theorems anandi3 and anandi3r into the main body.
These provide theorems of distribution of triple conjunction over
conjunction, similar to the existing anandi and anandir but for triple
conjunction.  This immediately shortens ndmovdistr.


david-a-wheeler:
I propose moving the theorems anandi3 and anandi3r into the main body.  We already have anandi and anandir; these are the equivalents for triple conjunction.

This immediately reduces ndmovdistr by 18 bytes.  There are probably other reductions this addition enables, but I don't think minimize can figure them out.

david-a-wheeler:
I'm almost certain that alsi-no-surprise can be shortened with these, but I have to rewrite the proof to use them.

PR #671, opened 2018-12-06
Fix various misspelled words and typos.

jamesjer:
I stumbled across 3 misspelled words last weekend.  As I was about to make a
pull request to fix them, I thought it would be great if I could run a spell
checker to see if there were more.  I threw together a very basic Emacs
metamath mode, which basically just distinguishes comments from non-comments,
then used the Emacs function ispell-comments-and-strings to spell check just
the comments (since the theorems and proofs are guaranteed to fail a spell
check!).  It took a few hours to pore through the results.  Note the following:

- I ran the check with an American English dictionary, which flagged words
  spelled the British way (e.g., "neighbourhood").  I made no attempt to do
  anything about such words, although my personal opinion is that standardizing
  on either American or British spelling would be helpful, as then the same
  word would be spelled the same way throughout.

- I stopped looking at the end of set-main.mm.  There are three uses of
  "antecendent" in the mathboxes, but I do no think it is a good idea for me
  to correct somebody else's mathbox.

- In a few cases, the text had to be rewrapped due to adding a letter or two to
  a line that was already 79 characters long.

- I did not correct grammar, just spelling.  If you look through the changes,
  you can see a few sentences with incorrect grammar.

- The change at line 172709 for 4001lem2 is not, strictly speaking, a spelling
  correction, but since I noticed the error I figured it should be corrected,
  too.

- Note the comments above supsr.  There were two "Contributed by"
  comments.  The first one misspelled "contributed", which is how I found it.
  I believe the first one is just wrong and removed it, in favor of the
  "Revised by" comment with the same date and author.


digama0:
LGTM

nmegill:
Thank you for fixing these errors.

Your fixes to misspelled "(Proof modification is discouraged.)" and "(New usage is discouraged.)" caused the Travis check to fail because the "discouraged" file is now different.  I regenerated it for you, so it is OK now.  In the future, you can regenerate it as follows:

<pre>
./metamath 'read set.mm' 'set width 9999' 'show discouraged' quit \
    | tr -d '\015' | grep '^SHOW DISCOURAGED.' \
    | sed -E -e 's/^SHOW DISCOURAGED.  ?//' | LC_ALL=C sort > discouraged
</PRE>

(I hope to have the "show discouraged" command generate it directly in the correct format, but for now the above is what you have to do.)

> * I ran the check with an American English dictionary, which flagged words
>   spelled the British way (e.g., "neighbourhood").  I made no attempt to do
>   anything about such words, although my personal opinion is that standardizing
>   on either American or British spelling would be helpful, as then the same
>   word would be spelled the same way throughout.

I think that American English currently dominates (and is used elsewhere on the web site) so I would suggest that be the standard used throughout.  E.g. "neighbourhood" is used one time in set.mm, whereas "neighborhood" is used 176 times.

> * I stopped looking at the end of set-main.mm.  There are three uses of
>   "antecendent" in the mathboxes, but I do no think it is a good idea for me
>   to correct somebody else's mathbox.

I think most mathbox users would welcome spelling error corrections as well as obvious grammar error corrections such as subject-verb agreement.  A lot of mathbox theorems will be imported into the main set.mm some day, so correcting them now would be a good idea.  We explicitly say that we "may make changes to mathboxes to maintain the overall quality of set.mm" in the comment of theorem mathbox.

> * In a few cases, the text had to be rewrapped due to adding a letter or two to
>   a line that was already 79 characters long.

This can be done automatically by reading set.mm into metamath.exe then 'write source set.mm /rewrap'.

> * I did not correct grammar, just spelling.  If you look through the changes,
>   you can see a few sentences with incorrect grammar.

Fixing these would be welcome whenever you find them.

> * The change at line 172709 for 4001lem2 is not, strictly speaking, a spelling
>   correction, but since I noticed the error I figured it should be corrected,
>   too.

Good.

> * Note the comments above supsr.  There were two "Contributed by"
>   comments.  The first one misspelled "contributed", which is how I found it.
>   I believe the first one is just wrong and removed it, in favor of the
>   "Revised by" comment with the same date and author.

That is correct.  The "verify markup */f" command complains whenever there are two "(Contributed by"s or a missing "(Contributed by" but of course it will miss this if "Contributed" is misspelled.

Norm

david-a-wheeler:
> In the future, you can regenerate it as follows...

The much simpler way to do it is to run:
scripts/regen-discouraged

On Windows you can run this via Cygwin.

PR #672, opened 2018-12-07
Remove repeats

david-a-wheeler:


david-a-wheeler:
This sequence of commits reduces set.mm's size by 5208 bytes.

I ran scripts/find-repeats and found several statements that are repeatedly proved a very large number of times.  I proved them separately (they are all small), then minimized all statements not in a mathbox. This resulted in a large (distributed) savings by only proving these theorems once.

I rewrapped set.mm first so I would have accurate counts.  I then found that "wc -c set.mm" reported:
33627507 set.mm

After minimizing on these new repeats, we have:
33622299 set.mm

david-a-wheeler:
A nice side-effect of having these proved directly is that this slightly improves automation from the user's point of view (because these common sequences are now instantly and directly available).

david-a-wheeler:
Fixing the discouraged proof means this saves 5207 bytes, not 5208 bytes.  That's still pretty impressive.

I should note that this approach's returns quickly diminish.  But this was an easy step to take.

If I applied these to the mathboxes I expect much larger returns,. but I wasn't sure that I should do that.

david-a-wheeler:
The new theorems are 0red,1red,1cnd,neg1rr,2t1e2,1zzd,picn.  These are trivial, e.g. 0red is:

> 0red $p |- ( ph -> 0 e. RR ) $=

The minimization I used was:

> min 0red,1red,1cnd,neg1rr,2t1e2,1zzd,picn/no_new ax-*

nmegill:
@david-a-wheeler 

> Fixing the discouraged proof means this saves 5207 bytes, not 5208 bytes. That's still pretty impressive.

I'm not sure what you mean.  Why was a discouraged proof affected?

> If I applied these to the mathboxes I expect much larger returns,. but I wasn't sure that I should do that.

We include mathboxes in our annual full minimize runs, so I think that would be OK.  The only negative I can think of is that a mathbox may acquire large diffs that the mathbox owner could find confusing,  but unless the owner is working on rewriting a proof that became minimized, I don't see it leading to merge conflicts.  It hasn't  been a problem for our annual full minimizations.  So I would suggest doing this in a separate pull request.

david-a-wheeler:
> I'm not sure what you mean. Why was a discouraged proof affected?

Sorry, I did not explain that clearly. I didn't affect any discouraged proof.  The problem was that my original proof of 1cnd used ax1cn but further use if that is discouraged. I replaced that reference with ax-1cn ... that fixed the problem but took an extra byte because the name is longer. It's no big deal.

nmegill:
BTW nice work!

david-a-wheeler:
> BTW nice work!

Thanks! I think I can do one more round while being useful.  This approach only works when there are very large number of repeats, so the point of diminishing returns gets reached pretty quickly.  Still, it shortens things, and it also means that a number of trivial claims do not need to be reproven, which I think makes the proofs a little easier to read.

tirix:
Nice work!
I'm certain this will also greatly reduce the size of the mathboxes.

PR #675, opened 2018-12-09
More repeats

david-a-wheeler:
This branch saves 10326 bytes in set.mm by proving *once*
a large number of small facts that have previously been repeatedly reproved.


david-a-wheeler:
This took longer than I expected, but I think it turned out well!

PR #677, opened 2018-12-10
Add selpw and ssnid to iset.mm (common set.mm repeats)

david-a-wheeler:
This adds selpw and ssnid to iset.mm.
These are common cases in set.mm, and we expect them to become
common in iset.mm as well.  Adding these will also ease moving
some theorems from set.mm to iset.mm, since we expect that
over time many theorems will use them.


jkingdon:
Thanks, looks good.

PR #679, opened 2018-12-11
Add scripts/typeset.py

david-a-wheeler:
Add a simple script to apply an mmfile typeset rules to an arbitrary file.

### DIFF
--- a/scripts/typeset.py
+++ b/scripts/typeset.py
@@ -1,0 +1,159 @@
+#!/bin/python
+# Process a file's `...` instructions using an mmfile's typographical commands
+# Reads from stdin and sends to stdout.
+# Use the "--help" option to see all the options.
+
+# Copyright 2018, David A. Wheeler
+# SPDX-License-Identifier: MIT
+
+from __future__ import print_function
+
+import sys
+import re
+import fileinput
+import argparse
+
+# TODO: Currently each `...` must not cross lines.
+# TODO: Currently this doesn't handle 'no space if followed by punctuation'.
+
+# Apply typographical definitions in MMFILE which look like this:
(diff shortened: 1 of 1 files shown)


nmegill:
Is this meant to replace the 'markup' command we discussed for metamath.exe?  Do you still want that command?

david-a-wheeler:
It *was* meant to replace the markup command.  However, I've now realized that mmset.html does a lot more than this. In particular, in its display of this:

~~~~
2p2e4 <- 2cn <- 2re <- 1re <- axrrecex <- recexsr <- sqgt0sr <- pn0sr <- distrsr <- dmaddsr <- addclsr <- addsrpr <- enrer <- addcanpr <- ltapr <- ltaprlem <- ltexpri <- ltexprlem7 <- ltaddpr <- addclpr <- addclprlem2 <- addclprlem1 <- ltrnq <
~~~~

It uses backquotes (yikes) and refers to summarized titles.

I think typeset.py is useful... but if we want *that* information, a little Python script won't cut it. Hmm.  Thinking.

nmegill:
These are just a blind copy of the truncated description to display in the tooltip, to give the user a quick hint about the linked theorem.  There is no attempt to translate the markup, nor do we have any plans to do that.

You can just permanently change them to ' (regular quote), before adding any markup, so as not to interfere with the markup translation.  When I regen the 2+2 trivia list I'll just have to remember to do that.

Norm

david-a-wheeler:
Oh, I see. Instead of hand doing it, I suggest copying that part into a file.  We can then translate pieces and concatenate them all together. That way it can be done automatically.

I think for my test cases I should start with something simpler than mmset.html. That is a big file.

PR #680, opened 2018-12-11
Tweak scripts/typeset.py

david-a-wheeler:

### DIFF
--- a/scripts/typeset.py
+++ b/scripts/typeset.py
@@ -1,7 +1,9 @@
 #!/bin/python
 # Process a file's `...` instructions using an mmfile's typographical commands
-# Reads from stdin and sends to stdout.
+# Reads from stdin and sends to stdout.  By default the mmfile is "set.mm".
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
I've been testing this on mmcomplex.html.
Interestingly, the current "old HTML" generation is different, because
we now include TITLE= values. E.g.:

~~~~
 To do this, we introduce two classes
-<IMG SRC='bbc.gif' WIDTH=12 HEIGHT=19 ALT='CC'>
+<IMG SRC='bbc.gif' WIDTH=12 HEIGHT=19 ALT=' CC' TITLE='CC'>
 (set of complex numbers) and
-<IMG SRC='bbr.gif' WIDTH=13 HEIGHT=19 ALT='RR'>
+<IMG SRC='bbr.gif' WIDTH=13 HEIGHT=19 ALT=' RR' TITLE='RR'>
~~~~

If we auto-generated these HTML files, these omissions and differences would not exist :-).

PR #684, opened 2018-12-12
Add note about part-whole theory in Aristotle comments

david-a-wheeler:
Add a note about part-whole theory in the discussion about
Aristotle.  Basically, older logic systems didn't fuss as much
about the difference between "is member of" and "is subset of".
Modern logic systems generally do care about this difference, of course.

These older systems are often called "part-whole theory".
If you don't know its name, you can't really
learn more about it, and its existence also helps justify why
we can just use wffs for these syllogistic forms.
Basically, we shouldn't worry about these distinctions, since
Aristotle did not.


nmegill:
Note that this won't go on the us2 site until I've finished a regression test, sometime tomorrow (or later if it fails), to make sure the 'markup' command didn't have any side effects on the site build.

david-a-wheeler:
No problem! I'm very happy with the new "markup" command (now that I managed to get it running).

nmegill:
> Note that this won't go on the us2 site until I've finished a regression test, sometime tomorrow (or later if it fails), to make sure the 'markup' command didn't have any side effects on the site build.

Sure enough, the regression test failed.  The corrected program version is 0.170 on us2; please use that for any further testing.  The regression test has been restarted.

PR #685, opened 2018-12-12
Fix mmcomplex.html and mmcomplex.raw.html

david-a-wheeler:
This fixes mmcomplex.html and its corresponding parts in
mmcomplex.raw.html.  It makes these changes:

* (. becomes C.
* (_ becomes C_
* P~ becomes ~P
* seq1 becomes seq
* df-seq1 becomes df-seq

At the moment, "mmcomplex.html" is the master version, so I edited it
and then generated mmcomplex.raw.html using:

> python scripts/unjunk.py < mmcomplex.html > mmcomplex.raw.html

In the long term I hope to make mmcomplex.raw.html the master
version, which then generates mmcomplex.html, but that potential
result is for another time.

The discussion about df-seq may be obsolete, but currently
I'm just trying to fix the formatting.  Updating the actual
discussion there is for another time.

### DIFF
--- a/mmcomplex.html
+++ b/mmcomplex.html
@@ -201,13 +201,13 @@ numbers</A> as specific subsets of
 <IMG SRC='bbc.gif' WIDTH=12 HEIGHT=19 ALT='CC'>,
 leading to the nice <A HREF="nthruc.html">relationships</A>
 <IMG SRC='bbn.gif' WIDTH=12 HEIGHT=19 ALT='NN'>
-<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='(.'>
+<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='C.'>
 <IMG SRC='bbz.gif' WIDTH=11 HEIGHT=19 ALT='ZZ'>
-<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='(.'>
+<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='C.'>
 <IMG SRC='bbq.gif' WIDTH=13 HEIGHT=19 ALT='QQ'>
-<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='(.'>
+<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='C.'>
 <IMG SRC='bbr.gif' WIDTH=13 HEIGHT=19 ALT='RR'>
-<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='(.'>
+<IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT='C.'>
 <IMG SRC='bbc.gif' WIDTH=12 HEIGHT=19 ALT='CC'>.
 We later define the set of extended reals
 (real numbers, positive infinity, and negative infinity), and once those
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
Once these fixes are in mmcomplex.html there are no reported warnings when running markup.  I did a test run and generated a file with the /althtml option (Unicode).  No problems at all once these changes are made, and it's a really nice improvement. Hooray!

david-a-wheeler:
Excellent. I intend in an upcoming commit to delete mmcomplex.html, and create a short script for generating selected HTML files from .raw.html. Obviously you will have to change the generation process to use them. It produces an immediate upgrade in the Unicode display, which becomes significantly prettier to read.

nmegill:
> Excellent. I intend in an upcoming commit to delete mmcomplex.html, and create a short script for generating selected HTML files from .raw.html. Obviously you will have to change the generation process to use them. It produces an immediate upgrade in the Unicode display, which becomes significantly prettier to read.

Don't delete mmcomplex.html yet.  One issue is where we generate it.  If it is generated by the build script, then even though errors will get logged I don't look at the log very often.  I just do an occasional sanity check now and then because 'verify markup' will have caught essentially all errors before the build is run.  So if the build generates things that may have errors not previously checked for, they may get overlooked for a period of time.

So it might be better to generate it at the Travis level so they would be caught immediately.  What do you think?

It also might be wise to use ~ label tags for theorem references in mm*.raw.html files, because there have been cases where the set.mm label was changed but some of the mm*.html files weren't updated (this could even be the case right now).  But running 'markup' without '/symbols_only' can be a nuisance due to literal '[' that frequently occurs in normal text.  We could escape them, but people editing .raw.html will probably forget to do that.  I don't think literal '~' occurs very often.  So I'm considering adding '/labels_and_symbols_only' (abbreviated with /l of course) that would ignore bib refs.   '/labels_and_symbols_only' would also suppress the colored number after the label, which is distracting clutter in normal prose text.

Norm

david-a-wheeler:
Adding Travis checks is an excellent idea. (Indeed, adding CI/CD checks is almost always an excellent idea.)  I can do that later today.

I like /labels_and_symbols_only , since that lets us automatically detect problems. Indeed, mmcomplex.html had an example of that - it referred to df-seq1 which no longer existed.  Omitting the colored number also sounds like a good idea.  Maybe "omit colored number" should be controlled with a separate option, because I can imagine that being useful in other situations as well.

nmegill:
If we're going to add a bunch of options, maybe it's better to specify what you want rather than what you don't want.

Example:
/symbols - process \`...\`
/labels - process ~ label
/colored_numbers - add colored number after label
/bibliography - process [bibref]
/underscores - process _emphasized text_ and x_n

If none of these are specified then the output file will equal the input file i.e. no processing occurs.

I don't think /underscores would ever be used since it's better to do it in native HTML, but if we're going to add the others we might as well be complete.  I think only the first two would typically be used.  /bib could be handy on occasion but is also a nuisance because it requires mmset.html or equivalent to be present.

Alternately, it could default to everything, with the negative of each above option instructing what to drop.  Although we do that in 'verify markup' with /date_skip and /file_skip, somehow it feels slightly inelegant to have to tell the computer what you don't want it to do.

Norm

david-a-wheeler:
I agree, it's much better to have *positive* options instead of *negative* ones; it's easier understand the positive options. So I like your proposed options.  /labels in particular will be useful... but we don't need to wait for that before switching to .raw.html (that can be done later).  I'll create a separate issue on switching to the .raw.html files.

PR #686, opened 2018-12-12
Add scripts/check-raw-html and use it in .travis.yml

david-a-wheeler:
This adds a check of "raw" html files to detect markup problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,12 +48,13 @@ install:
 script:
   - scripts/verify --date_skip --extra 'write bibliography mmbiblio.html' set.mm
   - scripts/verify iset.mm
-  # Verify and also also generate 'discouraged.new':
+  # Verify and also generate 'discouraged.new':
   - scripts/verify-mmj2 --extra 'RunMacro,showDiscouraged,discouraged.new' set.mm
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
For this commit to *work* as expected, the metamath at http://us.metamath.org/downloads/metamath.tar.bz2 will need to be updated to include the markup command.

nmegill:
OK, metamath.tar.bz2 is on the us server.

On 30 Dec 2016 we talked about a download of just the program without *.mm, which I made and is called metamath-program.zip (which I also put on the us server).  I thought it was going to be used for Travis; if so, has that idea been abandoned?

david-a-wheeler:
> On 30 Dec 2016 we talked about a download of just the program without *.mm, which I made and is called metamath-program.zip (which I also put on the us server). I thought it was going to be used for Travis; if so, has that idea been abandoned?

I didn't remember that you did it.  I will chalk that up to "it was the end of the year and I was confused".

I will make a pull request so that Travis uses that instead.  It'll save some time and bandwidth.

david-a-wheeler:
We can't switch to metamath-program.zip right now because it's missing the configure.ac file.  Can you please add that?  I created a separate issue to track this: https://github.com/metamath/set.mm/issues/687

PR #688, opened 2018-12-13
Download metamath-program.zip instead of metamath.tar.bz2

david-a-wheeler:
Modify scripts so they download and install only what they need to.


david-a-wheeler:
This speeds up things, since instead of downloading an 11MiB .bz2 file, we download a 633KiB file.

PR #690, opened 2018-12-13
Add & and => to set.mm constants

david-a-wheeler:
Declare typographical constant symbols "&" and "=>" to set.mm.

This may seem extremely odd, since they are not used directly
in the formalism.  However, they *are* symbols we find useful when
explaining the formalism.  Their omission makes it harder to
generate text for different circumstances, in particular, for
HTML where GIFs must be used.

It is much easier to consistently use
a single typographical system when generating text, and it does
no harm to define constants with typographical definitions but
no valid syntax within the formalism.  Rather than work around
their omission, this commit simply adds them so they can be
used in text wherever it is convenient.


david-a-wheeler:
The lack of these symbols makes it trickier to translate some mm*.html files...
I think it'd be easier to just define them!

Issue #691, opened 2018-12-13
Delete mmcomplex.html and mmnatded.html - generate them from .raw.html instead


david-a-wheeler:
I propose that we delete mmcomplex.html and mmnatded.html, and instead generate them from their .raw.html files instead (mmcomplex.raw.html and mmnatded.raw.html).

You can regenerate the "old HTML" format this way:

scripts/regen-from-raw

And you can regenerate the "new HTML" (ALTHTML) formats this way:
scripts/regen-from-raw "/ALTHTML"

Currently that script generates .demo.html files; just delete ".demo" to generate the real things.

The ALTHTML versions are a *big* improvement over our current files when viewing in Unicode (the default).  mmcomplex.html, in particular, has a much prettier display.

There was discussion about this earlier in https://github.com/metamath/set.mm/pull/685 :

>>    Excellent. I intend in an upcoming commit to delete mmcomplex.html, and create a short script for generating selected HTML files from .raw.html. Obviously you will have to change the generation process to use them. It produces an immediate upgrade in the Unicode display, which becomes significantly prettier to read.

> Don't delete mmcomplex.html yet. One issue is where we generate it. If it is generated by the build script, then even though errors will get logged I don't look at the log very often. I just do an occasional sanity check now and then because 'verify markup' will have caught essentially all errors before the build is run. So if the build generates things that may have errors not previously checked for, they may get overlooked for a period of time.
> So it might be better to generate it at the Travis level so they would be caught immediately. What do you think?

I agree, but that has since been fixed.  The Travis build now checks the .raw.html files.

> It also might be wise to use ~ label tags for theorem references in mm*.raw.html files, because there have been cases where the set.mm label was changed but some of the mm*.html files weren't updated (this could even be the case right now). But running 'markup' without '/symbols_only' can be a nuisance due to literal '[' that frequently occurs in normal text. We could escape them, but people editing .raw.html will probably forget to do that. I don't think literal '~' occurs very often. So I'm considering adding '/labels_and_symbols_only' (abbreviated with /l of course) that would ignore bib refs. '/labels_and_symbols_only' would also suppress the colored number after the label, which is distracting clutter in normal prose text.

I think that's a good long-term plan, but we don't need to wait for that step.  Let's start the switchover with just symbols, and we can change the labels later.  Switching to generated symbols is a significant improvement in the display.  Switching labels has advantages, but it's already easy in HTML to create a hyperlink, so the benefits aren't as big a deal.

[...]

PR #692, opened 2018-12-13
Use => in mmnatded.raw.html, now that it's available

david-a-wheeler:
Add the first use of ` => ` in a .raw.html file (mmnatded.raw.html).
Everything seems to work correctly.


david-a-wheeler:
This is in support of eventually implementing issue https://github.com/metamath/set.mm/issues/691

PR #693, opened 2018-12-14
Modify Travis build to generate HTML from .raw.html

david-a-wheeler:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,3 +59,10 @@ script:
   - echo 'Looking for tabs (not allowed)...' && ! grep "$(printf '\t')" iset.mm
   - ./checkmm set.mm
   - ./checkmm iset.mm
+  # Generate HTML from raw files into mpegif-html/ and mpeuni-html/
+  - scripts/regen-from-raw '/HTML'
+  - mkdir -p mpegif-html/
+  - for f in *.raw.html; do echo mv $f mpegif-html/${f%.raw.html}.html ; done
+  - scripts/regen-from-raw '/ALTHTML'
+  - mkdir -p mpeuni-html/
+  - for f in *.raw.html; do echo mv $f mpeuni-html/${f%.raw.html}.html ; done


david-a-wheeler:
I'm not sure if this is what you had in mind for issue https://github.com/metamath/set.mm/issues/691 - is it?

nmegill:
BTW the mpeuni version will need the css script to load the XITS font and set the overridable colors, e.g. for set.mm:
```
  <STYLE TYPE="text/css">
  <!--
  .set { color: red; }
  .wff { color: blue; }
  .class { color: #C3C; }
  .symvar { border-bottom:1px dotted;color:#C3C}
  .typecode { color: gray }
  .hidden { color: gray }
  @font-face {
    font-family: XITSMath-Regular;
    src: url(xits-math.woff);
  }
  .math { font-family: XITSMath-Regular }
  -->
  </STYLE>
  <LINK href="mmset.css" title="mmset"
    rel="stylesheet" type="text/css">
  <LINK href="mmsetalt.css" title="mmsetalt"
    rel="alternate stylesheet" type="text/css">
```

You may have to put this in temporarily by hand for testing, but I plan to have the 'markup' command automatically insert the correct css specified in the $t statement, with a qualifier I'm adding (it will be '/css' or something).

david-a-wheeler:
The current Travis scripts only use /HTML or /ALT_HTML. I think if we just add /CSS , and remove the current CSS materials, things should work. Later on we can modify the files to use ~ and then use /LABELS , but I would rather make changes in small stages. Best to make sure we get the basic formulas working all the way through, and then we can extend what the processing does.

david-a-wheeler:
The fixups to Travis are here:

https://github.com/metamath/set.mm/pull/695

I forgot to do the /CSS part; I guess I'll do that as a separate pull request.  In any case, making sure it works at all is job 1.

nmegill:
I've run into a problem.  The symbol spacing tweaks are not currently implemented for symbols in comments, only for actual theorem and proof step displays, and fixing this is harder than I anticipated.

The problem is the following.  The comment processing algorithm simply scans linearly through the comment text, translating symbols one by one when in the math symbol mode triggered by "\`"; there is no look-ahead or look-back that could determine context.  The theorem and proof processing algorithm, on the other hand, is different - each formula consists of an array of symbols, and context can be determined by scanning back and forth through the array.

To do this for comments, I will have to parse the comment's math symbol strings into proof-step-like arrays and call the function that does the context-dependent tweaks.  This is a somewhat larger mod than I originally anticipated for the 'markup' command.  I'll add it to my to-do list, but it might not happen right away.

However, the cases that need tweaks seem relatively rare on the mm*.html pages.  I think the worst is mmset.html where I see 8 places from a cursory glance, the first being the x and y spacing in the \` A. x y = z \` part of ax-12.  An easy workaround is to break up the expression into \` A. x \` \` y = z \`.

In any case, the space-tweaking mod wouldn't affect the above manual workaround since each \`...\` grouping will be considered an independent symbol string.  So there is no reason not to do the manual fixes whenever bad spacing is found, if we find it bothersome.  (If we are truly obsessed, we can just do a global deletion of "\` \`" after the space-tweaking code is fixed.  But that won't affect anything the reader sees.)

david-a-wheeler:
I did not notice the spacing issues so go ahead and make the changes in the .raw.hrml files to fix it.

I think the plan you've outlined is the best course. Tweaked things for now, and at some point metamath will do a better job at it. I could try to modify my python script to adjust for spacing, but it doesn't even have the adjustments you already have.

nmegill:
> > The theorem and proof processing algorithm, on the other hand, is different - each formula consists of an array of symbols, and context can be determined by scanning back and forth through the array.
> 
> In the long-term, is there a way to put that kind of information inside the mm file instead of trying to do special cases within the code? Each symbol already has a translation for it in the mm file. I don't know if that would make any sense, I haven't looked extensively at the algorithm.

I don't think so, not without inventing a typesetting language that would be needlessly complex.  I don't want to make the markup language and $t any more complex than they already are, if nothing else as a favor to users who otherwise would have one more thing to learn.

The number of cases is quite limited and so far not terribly hard to add to the code when they crop up.  It is just a minor and _optional_ cosmetic adjustment.  For now, let's let it happen invisibly behind the scene inside metamath.exe.

PR #694, opened 2018-12-14
Delete mmcomplex.html and mmnatded.html (use .raw.html instead)

david-a-wheeler:
Delete the files mmcomplex.html and mmnatded.html, as they are
no longer source files.  We will instead edit their .raw.html
files and generate the .html files as needed.

This avoids the risk of editing non-source files (since those changes
will be discarded during regeneration).

Generating the .html files has several advantages - the most obvious
is that we can now truly serve better-looking Unicode versions of
those files.  More broadly, supporting .raw.html makes it much easier
to edit those files and keep their formatting in sync with the
actual formatting of the corresponding .mm file.

This does *not* modify mmset.html.  That file is longer and more
complicated, and it is probably better to get these "simpler" cases
working first.

This fixes issue #691.


david-a-wheeler:
This commit does *not* delete the files from version control, of course, so if we ever want to see them they're trivial to recover (using git checkout).  But "normal users" will just see the source files & won't be tempted to edit generated files - and that's a *good* thing.

PR #702, opened 2018-12-15
Create mmzfcnd.raw.html

david-a-wheeler:


david-a-wheeler:
I don't want to modify the generator script until the other commit is accepted, because that would cause a merge conflict.  I can later modify the script to generate .html files, and remove the original .html.

PR #706, opened 2018-12-18
Add scripts/linkify.py

david-a-wheeler:
Add the "scripts/linkify.py" script. This script converts
<A HREF="NAME.html">NAME</A> into ~ NAME .
It reads from stdin, writes to stdout

This will make it easy to convert existing HTML files to
use ~ NAME format.  This format can be more easily be checked
(they should *only* refer to valid labels, so if they
refer to something else then there is a problem).

### DIFF
--- a/scripts/linkify.py
+++ b/scripts/linkify.py
@@ -1,0 +1,43 @@
+#!/bin/python
+# linkify - convert <A HREF="NAME.html">NAME</A> into ~ NAME .
+# Reads from stdin, writes to stdout
+
+# Copyright 2018, David A. Wheeler
+# SPDX-License-Identifier: MIT
+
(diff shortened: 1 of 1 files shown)


nmegill:
@david-a-wheeler Is this ready to use (as far as you know) or is it still in testing?  It seems to work on mmset.raw.html (after escaping all ~ with ~~ in mmset.raw.html before running, which should be done anyway before 'markup ... /labels').  'markup ... /a/s/l/c' runs with no errors, mmset.html passes html validation, and overall it looks ok.  Since the line wrapping is different (as expected), diff between old and new mmset.html is harder to interpret but spot-checking a few diffs I didn't see any real differences.

PR #707, opened 2018-12-19
Add more repeatedly-proven statements, e.g., 2re0lt2

david-a-wheeler:
This commit adds common repeatedly-proven statements to set.mm.

Here are the statements this commit adds, including (1) the number of times
the statement is reproven in commit 1b70cbbf1a27daf56, (2) the name of the
new theorem, and (3) the statement itself.  These repeats were found
using "scripts/find-repeats".  These are all relatively simple statements.

~~~~
26 2re0lt2        |- ( 2 e. RR /\ 0 < 2 )
20 negpire        |- -u pi e. RR
20 cnunitopncnfld |- CC = U. ( TopOpen ` CCfld )
13 subgimpssbas   |- ( S e. ( SubGrp ` G ) -> S C_ ( Base ` G ) )
13 rngimpidbas    |- ( R e. Ring -> ( 1r ` R ) e. ( Base ` R ) )
12 nulssnid       |- (/) e. { (/) }
11 0gmgpcnfld     |- 1 = ( 0g ` ( mulGrp ` CCfld ) )
11 leinressxr     |- ( <_ i^i ( RR X. RR ) ) C_ ( RR* X. RR* )
10 pnflepnf       |- +oo <_ +oo
10 0icopnfimpicc  |- ( x e. ( 0 [,) +oo ) -> x e. ( 0 [,] +oo ) )
~~~~

This commit should reduce the database size (by eliminating duplicates),
simplify proofs (by reducing the number of steps they use), and make
it slightly easier to automatically generate proofs (by directly providing
common leaves instead of requiring a system to figure them out).


david-a-wheeler:
I found a few more statements that are repeatedly proven, and separately proved the most common ones that didn't have hypotheses.

This time, based on previous comments, I'm documenting how many times each expression was reproven.  That seems like useful information and I think justifies their inclusion. Instead of re-proving a simple statement 26 times, I think we should prove it once.

david-a-wheeler:
I listed the current repeat counts.  However, note that it's likely that these counts would increase over time unless we separately re-prove them, since they're obvious common things to need.

nmegill:
@david-a-wheeler This pull request increases the set.mm size by 1789 bytes, which means there isn't a size saving.  A 2-step proof like negpicn will save only a byte or two per proof, and it would require a very large number of references to make up for the overhead of 300 or so bytes for the statement.

In some cases it might make the bytes in the proof longer if it adds an additional label (such as if picn and negcli are already used elsewhere in the proof). I think 'minimize' accounts for this, but it will make the usefulness of negpicn somewhat less than one might think.

As for making it easier to automatically generate proofs, MM-PA will prove negpicn automatically with no noticeable increase in CPU time, and I think mmj2 will do the same.  A drawback of adding little theorems like this is they add to human overhead of scanning through the theorem list.

Can you identify which of these theorems actually reduce the size of set.mm?

david-a-wheeler:
I'm closing for now and will investigate separately later.

PR #710, opened 2018-12-21
possible oversight in recent renaming

wlammen:


wlammen:
Travis complains, because my renaming violates the 'New usage is discouraged' constraint. I don't know what to do about it.

PR #711, opened 2018-12-22
Iset iota

jkingdon:
Add "Definite description binder (inverted iota)" section to iset.mm.

The intuitionizing needed is about typical. The biggest one is adding the `E! x ph` condition to `iotaex`. But given how intuitionistic logic works, I'd be a bit surprised if there were a way to "bundle" the `E! x ph` and `-. E! x ph` cases, as set.mm does here.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -41685,8 +41685,7 @@ $)
        Sometimes proofs need to expand an iota-based definition.  That is,
        given "X = the x for which ... x ... x ..." holds, the proof needs to
        get to "...  X ...  X ...".  A general strategy to do this is to use
-       ~ riotacl2 (or ~ iotacl for unbounded iota), as demonstrated in the
-       proof of ~ supub .  This can be easier than applying ~ riotasbc or a
+       ~ iotacl (for unbounded iota).  This can be easier than applying a
        version that applies an explicit substitution, because substituting an
        iota into its own property always has a bound variable clash which must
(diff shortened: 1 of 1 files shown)


digama0:
There is something nontrivial you can prove about iota that doesn't require the uniqueness assumption: If `A. x ( ph -> x C_ A )`, then `( iota x ph ) C_ A`. So an iota over a collection of things regarded as sets (not atoms) actually makes sense and is something like the union of all the satisfying instances.

jkingdon:
Hmm, not sure I follow "union of all the satisfying instances" given the definition of iota as `U. { y | { x | ph } = { y } }` . Nor did I see a quick proof for `( A. x ( ph -> x C_ A ) -> ( iota x ph ) C_ A )` . Unless someone else wants to follow up on this, it sounds like a matter to come back to if/when it seems like it might help with intuitionizing the many uses of `iotaex`.

digama0:
Well it's not quite the union of all the satisfying instances, but it is (literally) the union of a subset of all satisfying instances. The proof is only slightly roundabout: just use sspwuni on the goal, so you have to prove `{ x | ph } = { y } -> y C_ A`, and this follows from `y = U. { y } = U. { x | ph }` and `U. { x | ph } C_ A`, which again boils down to sspwuni + the assumption.

jkingdon:
OK, submitted at #715 .

PR #712, opened 2018-12-22
Theorems about Matrices and Determinants

avekens:
First steps towards proving Cramer's rule.

Remarks:
- The theorems are based on the definitions and theorems in Stefan O'Rear's Mathbox (so my Mathbox is strongly depending on Stefan's in this part).
- I surrounded Stefan's Mathbox with Begin/End-Tags (to use it in my Mathbox)
- I moved 2 theorems from TA's and 2 theorems from my mathbox to main set.mm


savask:
I want to make a small nitpick - the correct term is "symmetric group", not "symmetry group" (at least in this case). Other than that - awesome work!

avekens:
> 
> 
> I want to make a small nitpick - the correct term is "symmetric group", not "symmetry group" (at least in this case). Other than that - awesome work!

done - thanks!

PR #713, opened 2018-12-22
shorten morimv

wlammen:
@nmegill I am actually a bit puzzled about this theorem. Without further ado E\* x(ph -> ps) -> E \* x ps is provable from ax-1. morimv in its current form is overly specific, both in expression length and constraints. I suggest to remove it from the database, or at least rename it, since no distinct variable constraints are required.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16333,6 +16333,7 @@ New usage of "mndoissmgrp" is discouraged (1 uses).
 New usage of "mndomgmid" is discouraged (3 uses).
 New usage of "mo3OLD" is discouraged (0 uses).
 New usage of "moOLD" is discouraged (0 uses).
+New usage of "morimOLD" is discouraged (0 uses).
 New usage of "morimvOLD" is discouraged (0 uses).
 New usage of "mp2ALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> @nmegill I am actually a bit puzzled about this theorem. Without further ado E* x(ph -> ps) -> E * x ps is provable from ax-1. morimv in its current form is overly specific, both in expression length and constraints. I suggest to remove it from the database, or at least rename it, since no distinct variable constraints are required.

Yes, your proof is much better.  At the time it was proved, moimi didn't exist.  The purpose of the theorem was to provide an analogue of one direction of 19.37v, pushing E* into an implication.  The "r" means "right side" as in the "r" of alrimiv.  Since it has never been used, I have no objection to deleting it.  Please go ahead and do so as part of this P.R..  Name your version morimOLD so as not to conflict with morimvOLD.  For the deletion entry at the top of set.mm, use "22-Dec-18  morimv --- obsolete; deleted - use morimi applied to ax-1".

nmegill:
>  "22-Dec-18 morimv --- obsolete; deleted - use morimi applied to ax-1".

I meant  "22-Dec-18 morimv --- obsolete; deleted - use moimi applied to ax-1".

PR #714, opened 2018-12-23
shorten dvelimhw

wlammen:
While I was looking at dvelimhw I noticed this small optimization.


nmegill:
It looks like I forgot about the hypotheses in the ax-12/13 swap. (:  I'll check the rest.  Thanks.

PR #729, opened 2019-01-08
replace an OLD theorem in proof

wlammen:


wlammen:
Shouldn't uniop be tagged with 'avoid depending on this detail'?

nmegill:
> Shouldn't uniop, uniopel be tagged with 'avoid depending on this detail'?

Of course it is preferable not to depend on the construction of ordered pairs when possible, but early theorems depend on it in an essential way.  In particular, we cannot compute the rank of some things without knowing exactly how ordered pairs are defined.  dmrnssfld and some subsequent existence proofs depend on these two theorems.  These theorems and proofs would of course have to be rewritten if we switched to an alternate ordered pair definition.

If you want to put in a comment like "We prefer to avoid this theorem when possible since it depends on the construction details of the ordered pair.", please go ahead.  I don't think we need to go all the way to "(New usage is discouraged.)" but if people feel strongly about it I won't object.

wlammen:
I noticed that some theorems (df-op, opid for instance) are marked this way, so I thought the above mentioned theorems should follow suit.

nmegill:
> I noticed that some theorems (df-op, opid for instance) are marked this way, so I thought the above mentioned theorems should follow suit.

This mark was added on Sep. 28, 2018 by David:
https://github.com/metamath/set.mm/pull/548
to 6 statements: df-op dfopif dfopg dfop opid elop opi1 opi2 opeqpr relop

It was initially "(New usage is discouraged.)" but was changed to the milder "(Avoid depending on this detail.)" after some discussion.  The "Avoid" mark isn't recognized by any software but is for humans.

If you want to put that phrase in uniop and uniopel, go ahead.  But as Mario indicates in that discussion, it is a "slippery slope".  E.g. 'show usage uniop/rec' lists 18610 proofs depending on it directly or indirectly.  Certainly we don't want to mark them all with "Avoid".  I suppose we could determine some "reasonable" properties other than opth that any ordered pair definition is expected to provide, such as existence when the arguments are proper classes.  This is a somewhat subjective judgment call, though.  I think a theorem such as dmex that depends on uniop would be considered a "reasonable" property that shouldn't be marked "Avoid", even though it can't be proved from opth alone.

Although I don't think the issue is terribly important, if someone wants to propose  additional theorems to mark with "Avoid" that's OK with me, as long as it isn't more than perhaps a dozen or so.

wlammen:
> ... if someone wants to propose additional theorems to mark with "Avoid" ...

No, I don't, it was just an observation.

I fell to sleep some three hours ago, nothing will follow here.

Issue #594, opened 2018-10-15
Request: Document how to "Intuitionize" in iset.mm's documentation


david-a-wheeler:
I would love to see the iset.mm documentation explain more about how to "intuitionize", including citations pointing to especially good books/papers and specific explanations of how to do it in metamath ("when you see X, you often need to convert to Y by following steps Z").  Pointing to a few examples of theorems that are especially good at explaining the changes (and how to do something similar) would be great.

This is probably especially a request to @jkingdon ... but anyone else's input would be welcome.

I am intrigued by intuitionistic logic, but I am hopelessly steeped in classical logic :-).  A little more explanation of what you have to change would be helpful for me, and probably for others.

jkingdon:
Hmm, if I gave some sketchy outlines, would that be enough for you to find references in iset.mm and turn it into documentation?

1. If you see pm2.61 you'll probably end up with two theorems for the two cases. In particular, if the cases were A e. _V and -. A e. _V you probably just care about the A e. _V case.

2. Non-empty almost always needs to be changed to inhabited.

3. If the original proof relied on propositional/predicate logic which isn't a theorem of intuitionistic logic, maybe there is a way of expressing it more directly (for example, keep things as E. rather than translating to A. and negation, e.g. cbvrexcsf , sbcrext ). This is perhaps the hardest one to put in cookbook form: sometimes I just try some things and see if anything works.

4. If you are dealing with a definition, try to find the best constructive definition from the literature (HoTT book, Stanford Encyclopedia of Philosophy, Brauer, etc). Once you pick a definition, that'll affect the proofs which rely on that definition.

5. If there is case elimination on whether variables are distinct, most of the time you just need the variant with distinct variables. Sometimes you can then remove the constraint with a temporary variable (e.g. the various sbco2* variants, nfrabya).

6. Sometimes one direction of a biconditional holds, or subset holds instead of equality. You might be able to keep the easy direction and worry about the other one later.

7. If your proof relies on dveeq2 try dveeq2or and likewise for the other things downstream of ax-i12 or ax-bnd . In these cases, and perhaps others involving disjunction, jaoi is your friend.

8. If your proof is doing tricky things perhaps in the interest of shortness, try just expanding the definitions and applying logic in a straightforward way. See if this gets you a working (although perhaps longer) proof.

9. If your proof relies on a biconditional in set.mm which isn't in iset.mm, see if one direction is in iset.mm and see which direction your proof is using. For example 19.35-1 or exnalim .

[...]

PR #732, opened 2019-01-14
Add intuitionization section to mmil.html

jkingdon:
This is the text which has been hanging out at #594 for a while.

@david-a-wheeler Do you have any comments on how to format this? I know you've been doing things with `.raw.html` files but I didn't fully understand how to work with them.

There's also a small change to add a few entries to the list of sample theorems.

### DIFF
--- a/mmil.html
+++ b/mmil.html
@@ -782,6 +782,18 @@ variables</A></LI>
 <LI>
 <A HREF="ordtriexmid.html">Ordinal trichotomy implies excluded middle</A></LI>
 
+<LI>
+<A HREF="findes.html">Mathematical (finite) induction</A></LI>
+
+<LI>Peano's postulates for arithmetic:
+<A HREF="peano1.html">1</A>
+<A HREF="peano2.html">2</A>
+<A HREF="peano3.html">3</A>
+<A HREF="peano4.html">4</A>
+<A HREF="peano5.html">5</A></LI>
+
+<LI><A HREF="nndc.html">A natural number is either zero or nonzero</A> (a special case of the law of the excluded middle which we can prove).
+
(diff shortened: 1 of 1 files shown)


nmegill:
Nice writeup!

jkingdon:
> Nice writeup!

The perfectionist in me keeps wanting to tweak it (maybe make it more explicit in places, maybe more references to examples in others). But on the whole, I agree, I'm relatively satisfied with this as a summary. I'd be glad to hear reactions, especially from people who are trying to use it to write proofs.

nmegill:
@jkingdon Since you have a lot of labels mentioned in your writeup, you may want to use @david-a-wheeler 's scripts (that are run once) to convert mmil.html to mmil.raw.html, then use the 'markup' command in metamath.exe to create mmil.html from it, keeping mmil.raw.html as the master.  David did this for mmset.html, letting us maintain with \` math \` and ~ label markup, and 'markup' will also make sure all labels and math symbols are consistent with iset.mm.  I'm sure you'll find mmil.html much easier to maintain with this, as I did for mmset.html (e.g. the recent axiom renumbering).

The 'linkify.py' script will convert existing label hyperlinks to ~ label.  David, what is the script to convert math symbols to \` ... \` ?  I notice that mmil.html uses Unicode symbols; does it work for them?  (If not, it shouldn't be too hard to do manual search and replace, but existing ones can be left as-is I suppose, at least temporarily.  An advantage to converting is that it will use the same XITS font as on the theorem pages.)

jkingdon:
> @jkingdon Since you have a lot of labels mentioned in your writeup, you may want to use @david-a-wheeler 's scripts (that are run once) to convert mmil.html to mmil.raw.html

I've made a stab at this at https://github.com/metamath/set.mm/pull/740

PR #733, opened 2019-01-19
shorten 2mo

wlammen:
Only looked at the 2nd half of the proof (from line 47 on, web page)


wlammen:
Credits to whom it belongs: I usually prove a theorem without looking at the current proof, to give completely different ideas a chance.  2mo was no exception.  I tried 2, 3 proof ideas on the reverse side of the equivalence, but ALL of them were longer than the current one. I only achieved this shortening by merging the best of all ideas.
Won't commit anything more today.
Wolf

PR #737, opened 2019-01-27
Add "First and second members of an ordered pair" to iset.mm

jkingdon:
The intuitionizing here is not difficult. Some theorems don't make it, but most of the needed changes are the usual intuitionizations.

Add `opelrng`, a counterpart to `opelrn`.

Also intutionize the section from `resfunexgALT` to `ofmresex`.  In set.mm these are in a section called "Finite induction (for finite ordinals)" (although I'm not sure they depend on finite induction). Some other parts of that section are in iset.mm already, but these ones seem to have slipped through the cracks.


david-a-wheeler:
Wow, that is an impressive amount of work.

I have no idea what "less curry" means, but my compliments to the chef. As in Curry-Howard?

nmegill:
Quotation to ponder:

The intuitionist overhearing with dismay the meanderings of some classicalist can always say:  "Poor fellow!  He actually thinks he is reasoning.  Still, there is some sense that can be made of his musings.  What he seems to be doing is assuming (without warrant) a bunch of excluded middles.  So I can charitably interpret him as constructing an enthymematic argument which can be made (intuitionistically) correct by adding the appropriate excluded middles as premisses."
- Anderson, Belnap, and Dunn, _Entailment_, vol. II (1992), p. 503

PR #740, opened 2019-01-28
convert mmil.html to mmil.raw.html

jkingdon:
Did I get this right? In particular, the file needs to be processed with the `/LABELS` option to the `MARKUP` command (or else the `~~` won't get unescaped into `~`), yet `regen-from-raw` does not specify that option unless it is passed in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *~*
 mm.log
+*.demo.html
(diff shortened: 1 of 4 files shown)


david-a-wheeler:
I plan to take a look, but today is the first day of class (I teach graduate classes at George Mason University).  So I may not have a chance to look today.

david-a-wheeler:
On a quick glance it mostly looks good.

The CI fails with:

~~~~
?Warning:  Couldn't open or read the file "mmil.html".  The bibliographic
hyperlinks will not be checked for correctness.  The first one is "[Margaris]"
~~~~

You might want to modify .travis.yml to generate mmil.html early, the same way that mmset.html is generated early.

PR #749, opened 2019-02-09
Definitions and theorems for determinants and adjuncts

avekens:
- splitting section "The determinant" into sections "The determinant" and "The matrix adjugate/adjunct" (in SO's mathbox)
- moving theorems of SO's determinants-1 branch from AV's mathbox into SO's mathbox
- moving more theorems from SO's determinants-1 branch into SO's mathbox
- moving some general theorems to the main part
- adding/updating comments for definitions and theorems for determinants, especially referencing Weierstrass.

@sorear : Again, very great work! Especially to prove ~ mdetuni is enormous, this should already have been contained in the 100 theorems list by itself (unfortunately we cannot change the list). I hope you are happy with my work, moving these valuable results from your determinants-1 branch, restructuring it a little bit and adding the comments. Of course you are free to modify them (it's your mathbox) as you like.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1394,6 +1394,9 @@
 "axc9lem5OLD" is used by "axc9lem7OLD".
 "axc9lem6OLD" is used by "axc9lem7OLD".
 "axc9lem7OLD" is used by "axc9OLD".
+"axia1" is used by "mdetunilem8".
+"axia2" is used by "mdet1".
+"axia2" is used by "mdetmul".
 "axia2" is used by "mdetleib2".
 "axia2" is used by "mdetrlin".
 "axia2" is used by "psgnodpm".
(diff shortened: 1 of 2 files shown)


avekens:
Unfortunately, some "discouraged" theorems came into set.mm by the theorems from Stefan's determinants-1 branch. How should these be handled?

david-a-wheeler:
If you are on a Unix-like, run scripts/regen-discouraged to fix the discouraged file.

nmegill:
Actually, the "discouraged" theorems in this case were discouraged for a reason, and references to them should have been corrected rather than regenerating "discouraged".  Specifically:  axia1 and axia2 are aliases for simpl and simpr as part of the derivation of intuitionistic axioms, and they aren't intended for regular use (otherwise it might be somewhat confusing to the reader to see axia1 in some places and simpr in others).  eqid1 is a joke alias for eqid that has some "April fools" text in its comment, and we prefer eqid for official use.

To fix these, replace all references to axia1,axia2,eqid1 (which should not be used anywhere) with simpl,simpr,eqid respectively.  You might as well change these in entire set.mm; sorear's stuff and maybe others could have used them before we implemented the "discouraged" checks in the metamath.exe and mmj2 proof assistants.

Then regenerate "discouraged" again to get everything back in sync.

PR #752, opened 2019-02-16
Verify lrparser

david-a-wheeler:


david-a-wheeler:
This modifies our CI test pipeline to use mmj2's verify.LRParser check on set.mm.  That should significantly reduce the risk of bad definitions leading to nonsense "proofs" in set.mm.

PR #753, opened 2019-02-17
Add the first thema of the 4 Stoic logic themata

david-a-wheeler:
Stoic logic at its basics had five axiomatic arguments
(the indemonstratables) and four inference rules (the themata).

This commit adds proofs of the first thema (of the four themata).
We already have the indemonstrables.

The idea is to show that the axiomatic system used in MPE is
at least as strong as this ancient logic system.
We can prove this if we can prove (or accept) its axioms and rules.
The statements and results are interesting
in their own right.  They may also be useful in other theorems,
since these were "obvious" enough to be considered the basis of a
logic system thousands of years ago.


david-a-wheeler:
It makes sense to add the Stoic themata - which are basically its inference rules - after the Stoic indemonstratables.  Here's the first one.  I intend to eventually formalize and prove the other 3 themata, but I thought it'd be best to work out the first one first.

Note: "themata" is plural, "thema" is the singular.

nmegill:
You need a tilde in front of https://... in order to produce hyperlinks.

david-a-wheeler:
Excellent point.  Fixed.

PR #754, opened 2019-02-17
Add a "--setparser" option to verify-mmj2 and use it

david-a-wheeler:
The mmj2 setparser option is a little tricky to use - it must be set
*very* early to have the intended effect.

So add a new --setparser option to verify-mmj2, and modify .travis.yml
so that we use it when checking set.mm.
If you use --setparser 'mmj.verify.LRParser' you can immediately
detect and counter "junk" definitions like this devious
malformed definition by saueran at oregonstate dot edu posted on
Mon, 11 Feb 2019 17:32:32 -0800 (PST):

```
wleftp $a wff ( ( ph ) $.
wbothp $a wff ( ph ) $.
df-leftp $a |- ( ( ( ph ) <-> -. ph ) $.
df-bothp $a |- ( ( ph ) <-> ph ) $.

anything $p |- ph $=
  ( wbothp wn wi wleftp df-leftp biimpi df-bothp mpbir mpbi
  simplim ax-mp ) ABZAMACZDZCZMOEZOCQAEZNDZRNAFGSHIOFJMNKLAHJ $.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ script:
   - scripts/verify iset.mm
   # Verify and also generate 'discouraged.new', and do extra checking
   # (the 'printf' lets us insert multiple lines):
-  - scripts/verify-mmj2 --setparser 'mmj.verify.LRParser' --extra 'RunMacro,showDiscouraged,discouraged.new' set.mm
+  - scripts/verify-mmj2 --setparser 'mmj.verify.LRParser' --extra "$(printf 'RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel,df-sbc\nRunMacro,showDiscouraged,discouraged.new')" set.mm
   - echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
   - scripts/verify-mmj2 iset.mm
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
Sigh.  You'd *think* just adding one line to mmj2 would be easy, but setparser is a little tricky - it's happy to "accept" it in a way that's useless.  I've confirmed that this order *DOES* detect and counter the malicious definition shared earlier.

david-a-wheeler:
@digama0 - it looks like when I tell mmj2 to use mmj.verify.LRParser, it correctly determines that it is syntactically fine - but then returns an error code (!).

Do I understand that correctly?  If so, can we fix mmj2 to return "no error" if there is actually no error?  I could work around with with a complicated script, but there'd be a big overhead (I'd need to run things twice), and it really doesn't make any sense.

david-a-wheeler:
@digama0 - never mind, I found it.  Yet another stupid bug. Wow.

david-a-wheeler:
Well, that was an adventure.  How come "simple" things are hard? :-).

david-a-wheeler:
It works!  Not only do correct files pass, but the sneaky broken definitions are now detected as shown in: https://github.com/metamath/set.mm/pull/755

PR #758, opened 2019-02-20
Check that iset.mm doesn't have syntax ambiguities

david-a-wheeler:
Modify CI checks so that we also check that the definitions
in iset.mm don't create syntax ambiguities.


nmegill:
@david-a-wheeler I'll let you merge this when you're ready.

david-a-wheeler:
@nmegill - thanks!!

@jkingdon - I don't know of any reason this would be a problem, I'm adding a check that is already applied to set.mm, and I presume that nobody wants the iset.mm syntax to be ambiguous.  That said, please speak up soon if there's a problem, else I'll merge it relatively soon. Thanks!

jkingdon:
Some day perhaps we'll try to factor out this kind of thing so it automatically applies to both `set.mm` and `iset.mm` without having to make an edit two places.

Anyway, looks good to me.

david-a-wheeler:
@jkingdon - it's an edit in the same file, one line down.  Further factoring sounds good, but these sorts of changes are relatively rare, so it's not a big deal.

PR #760, opened 2019-02-20
add tau (ratio of a circle's circumference to its radius)

jkingdon:
So I don't know for sure where this should go, but in the run-up to Tau Day last year, I found the topic of tau in set.mm somewhat irrestistible. I started out with a definition and didn't really get much further (partly because I understood less about set.mm and metamath tools than I do now). Well, for better or worse I finished off the development of this, so this pull request constitutes a "proper" definition of tau (in the sense of being one you might adopt if you weren't going to define pi) through to a proof that tau equals two times pi (relatively straightforward, and quite possibly could be even shorter if I were better at these kinds of proofs).

If you haven't heard of tau (as the ratio of a circle's circumference to its radius) and are scratching your head, there is lots more at https://tauday.com including some literature citations. Lots more that could be said about various aspects of tau, but for me it was more than anything else an exercise in seeing whether I could do some things with real numbers and prove a result which is not a huge leap from what is in set.mm already but which also isn't there in quite this form.


nmegill:
While this is nice work, typically a primary purpose of a new definition is to shorten theorems and proofs so that eventually the definition "pays for itself" in terms of smaller set.mm size,  Unless we commit to either pi or tau exclusively, we will require more space to have parallel developments of both.

A similar issue arose as to why we only have < instead of both > and < in the main set.mm, even though for some theorems it may seem more natural to state them with >.  We rejected this for the main set.mm because we would need twice the number of theorems, and perhaps worse if we wanted all possible combinations of < and > in a theorem.

My preference would be for this to go in a mathbox as a specialized topic.  If at some point in the future tau  becomes sufficiently advantageous, we will have your development available to move into the main set.mm.

Besides, you don't have a mathbox, and this provides a good opportunity to create one. :)

jkingdon:
Mathbox sounds right then. I'll revise the pull request.

david-a-wheeler:
@nmegill - you can add `>` to the main area without doubling the number of theorems.  Just declare (as a style issue) that `<` and `<=` are used instead of `>` and `>=`, and then declare only the basics about `>` in the main body so that people can see its formal definition.  You can even mark `>` as "discouraged" (a capability added later) so people won't use it accidentally.

david-a-wheeler:
Adding a mathbox for @jkingdon is definitely a good idea no matter what :-).

david-a-wheeler:
I wonder if it's too hard to distinguish tau the logical variable from tau the constant.  But that's a typographical issue that can be tweaked later.

nmegill:
> I wonder if it's too hard to distinguish tau the logical variable from tau the constant. But that's a typographical issue that can be tweaked later.

Our general convention has been to use math italic for variables and non-italic for constants, such as "V' vs "_V", "i" vs. "_i", "e" vs. "_e", "E" vs. "_E", etc.  The Greek variables for wffs are math italic, so a non-italic tau would not conflict.  It seems we are not following the convention for pi but since there's no conflict I don't think it's worth worrying about at this point.  BTW I think books have started to use this convention; I've seen non-italic i for sqrt(-1) and pretty sure I've seen non-italic pi recently

jkingdon:
I have revised this pull request to use a mathbox. I also changed the typography as suggested.

Ready for re-review.

david-a-wheeler:
Cool! This should be posted long before tau day (6/28). Maybe we can have people point to these proofs.

jkingdon:
> Cool! This should be posted long before tau day (6/28). Maybe we can have people point to these proofs.

Totally. I'm not sure how to publicize this in the wider tau community. I tweeted at tauday: https://twitter.com/jkingdon2000/status/1098848506142351360 but I don't know whether that is enough to get noticed and whether there is a better place to send such a statement.

PR #772, opened 2019-03-08
Add discussion on Frege to DAW's mathbox

david-a-wheeler:
Add to David A. Wheeler's mathbox some information to show
that MPE covers everything in Frege's Begriffsschrift.
At the least, we prove or include all its axioms.


nmegill:
The header text should go after the 2nd =-=-=-=-=-=-=, not before it (compare other examples).

PR #775, opened 2019-03-09
Clarify df-sb

david-a-wheeler:
Not everyone uses Curry's substitution notation, so modify the
comment of df-sb to make it *especially* clear
what gets substituted for what.


nmegill:
Could you change [ x / y ] to [ y / x ] here too?

And maybe fix elsb4 to swap them so the comment will be consistent.  (I do it like this:  Rename elsb4 to elsb4OLD everywhere, create and prove a new elsb4 that swaps x and y, use 'minimize elsb4/allow' on each proof using elsb4OLD, delete elsb4OLD, and add "(Revised by David A. Wheeler...)" to elsb4.)

[ y / x ]  is almost always used except in a few places, and someday we should make them consistent:

MM> sea * '|- ( [ x / y
6103 equsb3lem $p |- ( [ x / y ] y = z <-> x = z )
6108 equsb3 $p |- ( [ x / y ] y = z <-> x = z )
6113 equsb3ALT $p |- ( [ x / y ] y = z <-> x = z )
6118 elsb3 $p |- ( [ x / y ] y e. z <-> x e. z )
6123 elsb4 $p |- ( [ x / y ] z e. y <-> z e. x )
7667 eqsb3lem $p |- ( [ x / y ] y = A <-> x = A )
7674 eqsb3 $p |- ( [ x / y ] y = A <-> x = A )
7681 clelsb3 $p |- ( [ x / y ] y e. A <-> x e. A )
9319 sbralie $p |- ( [ x / y ] A. x e. y ph <-> A. y e. x ps )
95855 clelsb3f $p |- ( [ x / y ] y e. A <-> x e. A )
109959 subsym1 $p |- ( [ x / y ] [ x / y ] F. -> [ x / y ] ph )
(the ones in my mathbox aren't important - they will be deleted someday)

david-a-wheeler:
I certainly could, but normally we order things alphabetically.  If we use `[ y / x ]` everywhere, I think we should document that exception in `conventions`.

So: do you want to use y...x and document that in conventions, or switch to x...y ?

david-a-wheeler:
We're using `[ y / x ]` as the convention!

david-a-wheeler:
I'll fix elsb4 as a separate change.

PR #785, opened 2019-03-17
Symbol spaces and typos

benjub:
Fixed some typos:
- removed consecutive periods (searching for the string ". . ").

In (alt)htmldef:
- removed spaces before +oo and -oo, so that they behave like a normal term (to avoid unwanted space in e.g. http://us2.metamath.org/mpeuni/df-xr.html);
- removed spaces around T. and F., so that they behave like a normal wff (to avoid unwanted space in e.g. http://us2.metamath.org/mpeuni/df-tru.html);
- added a trailing space after, and removed a leading space before, the comma.

PR #786, opened 2019-03-17
Symbol spaces fixed

david-a-wheeler:


david-a-wheeler:
Pull this *instead* of #785 - I simply fixed the merge conflicts.

david-a-wheeler:
@nmegill - merge this instead of #785 .

@benjub - this is just your changes PLUS fixes to the merge conflicts.

benjub:
@david-a-wheeler Thanks a lot.  I'll try to follow your advice in #785.  It sounds like I should create a new branch each time?  How to delete them once the commit is done?  I would like to have as little divergence with metamath/set.mm as possible, and always do single, simple and small commits each time (I would have no use in having parallel branches for two different projects, for example).

david-a-wheeler:
Yes, create a branch each time for each different change.  That makes it much easier to handle them separately (e.g., maybe one is accepted and the other is not).  Branches are basically free in git, so use as many as you want.  You can delete a branch (once merged) using `git branch -d BRANCHNAME` .  To be honest, I often don't bother, but you can certainly do that.

david-a-wheeler:
You can also delete branched in the GitHub API, though that only deletes it on GItHub's end.

PR #791, opened 2019-03-19
Add Jon Pennant mathbox and arearect theorem

TjWhale:
Here is a pull request for my area of a rectangle theorem. I have followed the naming convention of areacirc so I chose arearect. I have run the rewrap commands on my set.mm. After an error with the last pull request I have also replace uses of discouraged theorems.

Let me know if there are any problems or changes required, I'm happy to alter things :)

Two questions: I think I may have not updated the main version date, as my proof says 19-Mar-2019, so the main date of the repo may need updating too.

Also this line "$( End $[ set-mbox-so.mm $] $)" appears just before my math box. It was there when I found it so I have not altered it but I'm not sure what it is. It's on line 483809.


benjub:
Congratulations.  I would suggest:
it's --> its (twice)
rectangle --> rectangle whose sides are parallel to the coordinate axes
Similarly as for other mathboxes, add before/after yours:
$( Begin $[ set-mbox-jp.mm $] $)
$( End $[ set-mbox-jp.mm $] $)

TjWhale:
Thanks for the feedback, I've made those changes :)

nmegill:
$( Begin $[ set-mbox-jp.mm $] $) ... $( End $[ set-mbox-jp.mm $] $)

is optional, and not everyone uses it.  It allows you to work on your mathbox as the standalone file set-mbox-jp.mm, which you create with 'write source set.mm/split'.  When done just 'read set.mm' as normal for verification etc.  'write source set.mm' (without /split) will reassemble the split pieces for the github pull request.

http://us.metamath.org/mpeuni/mmnotes.txt 21-Dec-2017 entry has the details on this syntax and how it behaves.

TjWhale:
Thanks Norm, that's helpful info :)

PR #793, opened 2019-03-23
 Added proofs around the Tarski Geometry

tirix:


benjub:
In the comment, shouldn't it be .- instead of -. ?

tirix:
Indeed! Thanks for the careful review, @benjub . I fixed this in the latest commit.
I've chosen ` .- `  for distance because the ` D ` is commonly used as a point name. It's not perfect, but it's the nearest I found the the idea. I'm opened to other suggestions if you have any!

PR #801, opened 2019-03-27
Improved mmj2 workaround for setvar

david-a-wheeler:
mmj2 doesn't currently know about "setvar"; it expects the older
"set" constant in some checks inside "macros/definitionCheck.js".
We've worked around that by modifying set.mm.  However,
it's not really a good practice to modify the file you're checking
before you check it.

This commit modifies the source code of mmj2's
"macros/definitionCheck.js" after downloading it so it will look for
"setvar" instead of "set".  That means we can check set.mm
without modifying it.

NOTE: This will currently cause a failure because iset.mm
uses "set" instead of "setvar".  Normally iset.mm tries to
be consistent with set.mm, so the solution is to modify iset.mm
to use the same convention.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,9 @@ install:
   - wget -N -q http://us.metamath.org/ocat/mmj2/mmj2jar.zip
   - unzip -q mmj2jar.zip
   - jdk_switcher use oraclejdk8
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Do not merge this until iset.mm is modified to *also* use "setvar" instead of "set"!

PR #806, opened 2019-03-31
add babygodel

benjub:
Add Theorem ~ babygodel to my mathbox.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17874,7 +17874,7 @@ Proof modification of "axc9lem6OLD" is discouraged (78 steps).
 Proof modification of "axc9lem7OLD" is discouraged (15 steps).
 Proof modification of "axnul" is discouraged (52 steps).
 Proof modification of "axnulALT" is discouraged (95 steps).
-Proof modification of "babygodel" is discouraged (64 steps).
+Proof modification of "babygodel" is discouraged (54 steps).
 Proof modification of "binom2aiOLD" is discouraged (171 steps).
 Proof modification of "bitr3" is discouraged (14 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).
--- a/discouraged
+++ b/discouraged
@@ -17874,6 +17874,7 @@ Proof modification of "axc9lem6OLD" is discouraged (78 steps).
 Proof modification of "axc9lem7OLD" is discouraged (15 steps).
 Proof modification of "axnul" is discouraged (52 steps).
 Proof modification of "axnulALT" is discouraged (95 steps).
+Proof modification of "babygodel" is discouraged (64 steps).
 Proof modification of "binom2aiOLD" is discouraged (171 steps).
 Proof modification of "bitr3" is discouraged (14 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).


benjub:
Thanks @david-a-wheeler !  I only copied the 10-line propositional-calculus proof contained in the cited article by Boolos.  As I wrote, the (much) harder part of the proof is in the definition of the proof predicate using Godel numbering and in the definition of the Godel-Rosser sentence, using Godel diagonalization.  One would have to model the theory T itself...  As you see, I use "tricks" here, interpreting both ` A. x ` and ` |- ` as "it is provable in T that".
I would welcome any comments, especially from @nmegill and @digama0

david-a-wheeler:
I think @digama0 had started working on that larger proof, but I don't know how far he got.

nmegill:
Before I attempt to understand the description or the referenced paper, let me just mention that the $d is unnecessary, and the hypothesis |- -. A. x F. is also not necessary.  Here is a proof without them:
```
 1           albii.1=babygodel.1 $e |- ( ph <-> -. A. x ph )
 2         bitri.1=albii       $p |- ( A. x ph <-> A. x -. A. x ph )
 3           impbii.1=sp         $p |- ( A. x -. A. x ph -> -. A. x ph )
 4           impbii.2=hbn1       $p |- ( -. A. x ph -> A. x -. A. x ph )
 5         bitri.2=impbii      $p |- ( A. x -. A. x ph <-> -. A. x ph )
 6       mpbir.min=bitri     $p |- ( A. x ph <-> -. A. x ph )
 7       mpbir.maj=xor3      $p |- ( -. ( A. x ph <-> A. x ph ) <-> ( A. x ph
                                                             <-> -. A. x ph ) )
 8     mtbir.1=mpbir       $p |- -. ( A. x ph <-> A. x ph )
 9     mtbir.2=df-tru      $a |- ( T. <-> ( A. x ph <-> A. x ph ) )
10   mpbir.min=mtbir     $p |- -. T.
11   mpbir.maj=df-fal    $a |- ( F. <-> -. T. )
12 babygodel=mpbir     $p |- F.

benjub:
@nmegill  Here, I use "A. x" as a device to mean "it is provable in the theory T that", so I can only use the three properties of "A. x" I mentioned in the comments.  In particular, I cannot use ~ sp , which would mean in this context that a provable-in-T sentence is true, which I do not want to assume.  That is why I restate |- -. A. x F. as a hypothesis (I mention in the comment that it is not, in this context, ~ nalf ).  As for the dv hypothesis, I stated it for that same reason.
In his article, Boolos uses \box to mean provability, since it has the same properties as the necessity modality (in e.g. the modal logic K4).  I did not want to define a new symbol and associated properties, so I decided to emulate this with ` A. x ` .

benjub:
On second thought, I will remove the dv hypothesis (but keep |- -. A. x F. as an explicit $e hypothesis).

benjub:
I fixed the typo found by David and added the tag "Proof modification discouraged" since the comment refers to a special feature of the proof, which should be preserved by a potential proof modification.

digama0:
How about doing this proof in modal logic instead? Like Norm says, this theorem looks a bit silly as is unless you make it clear what you are using about `A. x`. Maybe just have a `Proves` constant that you axiomatize in the section.

benjub:
> ... unless you make it clear what you are using about `A. x`.

This is precisely what I do in the first paragraph of the comment, especially its last sentence.

> Maybe just have a `Proves` constant that you axiomatize in the section.

I used ` A. x ` to avoid this extra work.  But if you think it is better, then I can do it...  Ok, I'll do it, so do not merge yet.

PS: I can email Boolos's article to anyone interested.

PR #808, opened 2019-04-05
LINEAR ALGEBRA: Enhanced comments and additions in AV's Mathbox

avekens:
- References to [Lang] added
- comments for the following sections added:
-- Direct sum of left modules
-- Free modules
-- Independent sets and families
-- Matrices
- Definitions and theorems for the support of functions and linear combinations in AV's Mathbox.


nmegill:
Trivia of the day:  Serge Lang is a distant relative of our friend Gérard.

PR #815, opened 2019-04-16
minimize sucprcreg

benjub:
Shorten the proof of sucprcreg (and keep the old one as sucprcregOLD with usage discouraged).  The new shorter proof does not use any dummy variables.


benjub:
I had forgotten the "proof modif. disc." tag, which I just added.

As for the minimization of ~ruALT, it keeps the same kind of proof, which is why I overrode the tag. The shortening simply involves replacing "df-nel mpbir" with "nelir", which could have been done running "minimize */except ru", so I did not add any credit to the comment.

I think it is now ready for merging.

PR #817, opened 2019-04-19
add theorems about xor and shorten associated modi ponentes

benjub:
There is some overlap with my previous (pending) PR.  I hope this will not impair merging.

Maybe bj-cases, bj-nbior, bj-xornan, bj-xoror could be moved to the main part: they look rather natural and important theorems.  I can insert them at the right place if @nmegill agrees.  Then, I could shorten the proofs of mtp-xor and mpto2 as indicated.  @david-a-wheeler might be interested.

### DIFF
--- a/scripts/min.cmd
+++ b/scripts/min.cmd
@@ -29,6 +29,7 @@
 ! MM> open log log.txt                (recommended, to review result later)
 ! MM> submit 1.tmp
 ! MM> write source set.mm
+! MM> close log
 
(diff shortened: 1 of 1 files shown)


nmegill:
You can go ahead and move the 4 theorems you mention.

nmegill:
I'll let you do this as part of this PR so you can just make the edit directly and not be concerned about the label change list etc.

PR #824, opened 2019-04-25
Propcalc

benjub:
* Message of the first commit is self-explanatory.

* As for the second commit, I noticed that the proof of 'elimif' was almost only propcalc (and actually a nice statement in propcalc), so I "extracted" it to a statement I called 'cases'.  I think this makes things clearer.  As for the credits, since the proof was already contained in that of 'elimif', I credited NM.

* Do you think that bj-ifval could be moved right after eqif ?


nmegill:
You can move bj-ifval after eqif in your next PR (removing "bj-" of course).

You can credit yourself for cases.  The attribution is for the theorem + proof, not just the proof.  Many proofs informally borrow ideas from other proofs without getting nitpicking about attributions, and that is fine for routine proofs like this.

PR #832, opened 2019-04-28
Remove dv

benjub:
@nmegill You might require further changes before merging.

* Edited comment of sb56 and two others to signal that one direction of the biconditional is also in the database, with no dv condition.  Can I do the same edits for the three more statements I mention in the header comment of my just added section "Removing some dv conditions"? By the way, is there a way to see which statements have a comment which links to a given statement/statement pattern (i.e., crossrefs)? I think that (reasonable) hyperlinks among statement comments are useful.

* bj-eqvisset may be interesting since it has no dv condition. Move to main?

* bj-alexeqg  is a closed form of ~ alexeq , which allows to shorten the proof of ceqex by almost half. Move to main?

*  bj-rabbida2  bj-rabbida  bj-rabbid are versions of existing theorems with dv conditions replaced by F/ hypotheses. They are used for the following item.

* bj-rabeqd is the deduction form of ~ rabeq and contrary to it, it has no dv condition. This allows to remove two dv conditions in each ~ rabeqbidv and ~ rabeqbidva . I give formulations with the remaining (third) dv condition replaced by a F/ hypothesis.  bj-rabeqd also allows to remove the dv in bj-inrab2, which therefore is the same as inrab with a shorter proof (since it is a special case of bj-inrab).  Move all/some to main? Reprove the dv versions from them? If yes, keep the older proofs?

Thanks


nmegill:
> * Edited comment of sb56 and two others to signal that one direction of the biconditional is also in the database, with no dv condition.  Can I do the same edits for the three more statements I mention in the header comment of my just added section "Removing some dv conditions"? 

Go ahead.  Watch out for the anti-bundling police. :)

> By the way, is there a way to see which statements have a comment which links to a given statement/statement pattern (i.e., crossrefs)? I think that (reasonable) hyperlinks among statement comments are useful.

The only ways I know of are (1) search * '~ abc' /comment in metamath.exe and (2) grep -A 5 '~ abc' set.mm (sometimes you may need more -A context lines to see statement label).  Note that /rewrap prevents a line break between ~ and the label.

I don't have plans to put a cross-reference on the web pages because I don't think there is that much demand for this, especially given that it's easy enough to do at the source level in (I think) the relatively rare cases when it is needed.

> * bj-eqvisset may be interesting since it has no dv condition. Move to main?

I looked at several uses of isset* and didn't see any proofs that would have benefited i.e. the E. x seems naturally part of any proof that uses them.  Since the bj-eqvisset proof is somewhat obvious, my feeling is that it should stay in your mathbox.

> * bj-alexeqg  is a closed form of ~ alexeq , which allows to shorten the proof of ceqex by almost half. Move to main?

OK.  Probably we should make alexeq obsolete in that case (only ceqex uses it).  To do that, rename alexeq to alexeqOLD with the comment "Obsolete as of 28-Apr-2019. " (this will cause it to get deleted after a year) and "(New usage is discouraged.)";  and put "28-Apr-19 alexeq --- deleted; use alexeqg" in the label change list.

> * bj-rabbida2  bj-rabbida  bj-rabbid are versions of existing theorems with dv conditions replaced by F/ hypotheses. They are used for the following item.
> * bj-rabeqd is the deduction form of ~ rabeq and contrary to it, it has no dv condition. This allows to remove two dv conditions in each ~ rabeqbidv and ~ rabeqbidva . I give formulations with the remaining (third) dv condition replaced by a F/ hypothesis.  bj-rabeqd also allows to remove the dv in bj-inrab2, which therefore is the same as inrab with a shorter proof (since it is a special case of bj-inrab).  Move all/some to main? Reprove the dv versions from them?

I have mixed feelings; typically we add F/ versions only when they become needed, which isn't that often.  If we did it routinely everywhere we'd have a huge number of theorems cluttering up set.mm that would never be used.  However, removing two $ds on rabeqbidv would be nice.  So go ahead and add at least bj-rabeqd.  Use your judgment on the rest if you think they will be used in the future.

BTW there are 2 conventions in use:  the $d version suffixed with "v" or the F/ version suffixed with "f".  Since the rab* theorems use the "v" convention, it seems your naming (with "bj-" removed of course) is OK.

> If yes, keep the older proofs?

It is optional.  Often when people find shorter proofs, they rename the original to *OLD, add an "Obsolete as of" comment and "(New usage is discouraged.) (Proof modification is discouraged.)" to the *OLD, and "(Proof shortened by...)" to the original; this way the proof shortening can be easily seen. 

When a $d is removed, typically we just add "(Revised by...)" to the original and don't create a *OLD, although if the proof is revised significantly to accomplish that, maybe create a *OLD.

You can make these changes in your next PR.  I'm merging so I will be able to see your proofs more conveniently.

david-a-wheeler:
> Go ahead. Watch out for the anti-bundling police. :)

Our eyes are everywhere! There's no escape.  Especially since they can be detected automatically by analyzing the public .mm file.... :-).

In all seriousness, even if we specially mark bundles or reduce their use, I would expect such a transition to be handled "all at once".  There's no agreement on whether or not we should, and if so how, so let's just keep working normally (as has been done here) and we can discuss that as a separate item.

benjub:
@david-a-wheeler Agreed!  Anyway, my current work is just about documenting the existing theorems and adding links among comments (which I think is always useful). And the work I plan to do (#834) with removing some dv conditions is about dv conditions mainly of the form var-wff or var-class, and not var-var.
Since you're here: what do you think of adding somewhere in the metamath book Norm's way to get "crossrefs" (see above) ?

PR #844, opened 2019-05-07
 Recovered the proof that the Euclidean space is a Tarskian geometry.

tirix:
And corrected the chapter title!


david-a-wheeler:
Fantastic!

PR #845, opened 2019-05-07
Discourage direct use of df-tru; use tru instead

david-a-wheeler:
There's been a lot of discussion about the definition of T.
in df-tru.  There are many ways to define T., and proofs should
generally not depend on the specific definition.

This commit discourages directly using df-tru, so that people
will use tru instead.  Theorem tru uses df-tru, but if all other
theorems use tru, that means the irrelevant details of df-tru
won't matter to anything else.

This commit regenerates "discouraged" to match.
It also comments out a few theorems by Alan Sare that used df-tru.
If someone wants to re-prove them using tru that could be done later.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4646,6 +4646,7 @@
 "df-topspOLD" is used by "eltopspOLD".
 "df-topspOLD" is used by "istpsOLD".
 "df-topspOLD" is used by "tpsexOLD".
+"df-tru" is used by "tru".
 "df-unop" is used by "elunop".
 "df-va" is used by "0vfval".
 "df-va" is used by "vafval".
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@nmegill - this seems like a good idea no matter what.

nmegill:
We don't need to comment out uunT1 since fixing its proof is easy:  replace the reference to df-tru to become tru, biid, 2th.  Specifically, the modified uunT1 proof without df-tru is:

      ( wtru wn wo orc tru biid 2th exmid a1i biidd impbii bitri sylibr mpancom
      wb ) DABAAAEZFZDASGDAARZTDUAHAIJUATTUAAKLTAMNOPCQ $.

Could you uncomment uunT1 and the theorems depending on it, and replace the uunT1 proof with this one?

The comment change in df-tru is OK.

Thanks.

david-a-wheeler:
Sure, I'll put the proof in.

PR #854, opened 2019-05-16
Recovered the function converting Hilbert spaces to Geometries

tirix:
So that they include the additional definition of lines.


tirix:
The news entry is maybe a bit too optimistic. Actually, the link between general Hilbert spaces and Tarski geometries is not completely done yet, neither is the link with the complex plane.

I believe we shall shorten the news to:

> Set.mm now has a major new section on elementary geometry. This begins with definitions that implement Tarski's axioms of geometry (including concepts such as congruence and betweenness). This uses set.mm's extensible structures, making them easier to use for many circumstances. The section then connects Tarski geometry with geometry in Euclidean places. Most of the work in this section is due to Thierry Arnoux, with earlier work by Mario Carneiro and Scott Fenton.

david-a-wheeler:
@tirix @nmegill - shortening the news entry makes sense to me!

nmegill:
I shortened the news item accordingly.

PR #855, opened 2019-05-16
add elementary thms of modal logic

benjub:
Add to my mathbox elementary theorems of modal logic, concerning axioms (T) and (D).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17756,6 +17756,11 @@ Proof modification of "binom2aiOLD" is discouraged (171 steps).
 Proof modification of "bitr3" is discouraged (14 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).
 Proof modification of "bj-19.21t" is discouraged (39 steps).
+Proof modification of "bj-ala1" is discouraged (9 steps).
+Proof modification of "bj-ala1BIS" is discouraged (9 steps).
+Proof modification of "bj-axd2d" is discouraged (16 steps).
+Proof modification of "bj-axdd2" is discouraged (28 steps).
+Proof modification of "bj-axtd" is discouraged (24 steps).
 Proof modification of "bj-denotes" is discouraged (69 steps).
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
There are usual notations for modal logic (diamond and box).  I think we should define them and use them.

FL had a number of things in his now-defunct mathbox, starting at "Linear temporal logic".  Perhaps they could used/generalized to support any kind of modal logic?

benjub:
Yes, I mentioned these notations in the head comment of the subsection. I
think that if modal logic is to be developed, it should be in a new modal.mm
whose first line would be $[ propcalc.mm $]

The tiny theorems I just added were actually to answer a question asked to
me by Gérard Lang.

I'll have a look at FL's work, thanks for pointing it.

nmegill:
I have mixed feelings about David's suggestion, just as I had mixed feelings about FL's adding new modal symbols and new axioms for them.  For one thing, set.mm is generally supposed to be about "set theory" as its name implies.

Indeed I have wondered whether our numerous axiomatizations of prop calc really belong in set.mm.  For one thing, there are issues with knowing whether by accident the proofs are "cheating", which now can only be determined by manual inspection.  But at least they are ultimately the same classical prop calc that set theory uses and in principle could serve as alternate axiomatizations in place of the official one.

Since the correspondence between [] and A. x is immediate and direct, and since as Benoit says it's for a limited purpose of addressing an email of Gerard, I don't think that it is necessary or worthwhile to introduce the modal symbols and axioms for them.  I'll leave that up to Benoit, but it is unlikely it would ever be moved out of his mathbox.

BTW the MPE Home Page mentions that ax-c5, ax-4, ax-10, and ax-gen correspond to modal system S5 under the interpretation of A. x meaning [].

david-a-wheeler:
Modal logics are way larger than just answering one email from Gerard, though, especially since it's a whole field.  This is probably better discussed in a different (live) issue, though, so at some point soon I think I'll open a new issue so we can discuss it separately.

PR #862, opened 2019-05-20
Unbundling 3 theorems and alternative definition of linear dependency for vector spaces

avekens:
Changes in AV's Mathbox only:
Unbundling ~ ovmpt3rab1, ~ elovmpt3rab1 and 2wlkeq:
- ~ ovmpt3rab1 and ~ elovmpt3rab1 revised using implicit substitution
-- problematic set variables not existing/needed anymore
- ~ 2wlkeq split into two variants:
-- ~ 2wlkeq ("degenerated instance")
-- ~ 2wlkeqdv ("principal instance")

Alternative definition of linear dependency for vector spaces:
- Main theorem ~ isldepslvec2 with some intermediate results also valid for left modules.
- 2 new theorems for the support of functions and nonzero rings

### DIFF
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -5219,6 +5219,10 @@ revised edition (1969) [QA248.Q7 1969].</LI>
 <I>Methods of Modern Mathematical Physics</I>, Vol.  I:  Functional
 Analysis, Academic Press, New York (1972) [QC20.7.F84.R43].  </LI>
 
+<LI><A NAME="Roman"></A> [Roman] Steven Roman, <I>Advanced Linear Algebra</I>,
+3rd edition, Graduate Texts in Mathematics, Springer Science+Business Media,
+New York (2008) [QA184.R65 1992].  </LI>
+
 <LI><A NAME="Rosenlicht"></a> [Rosenlicht] Maxwell Rosenlicht, <I>Introduction
 to Analysis,</I> Dover Publications, Inc., New York (1986) [QA300.R63]. </LI>
 


benjub:
@avekens  So you added extra unnecessary dv conditions to what was 2wlkeq (now 2wlkeqdv) in order to weaken it and make it less widely applicable? I'm not sure this is for the better. (I know that in this case this is not important, but I'm afraid we might be on a slippery slope.)

avekens:
@benjub I did this due to the discussion about bundled theorems in the Metamath mailing list, see https://groups.google.com/d/msg/metamath/A2p5GPDwXxw/Ft0Y_MmxAAAJ, so that 2wlkeq would not appear on Mario's list anymore. I think 2wlkeqdv and (the new) 2wlkeq together have the same applicability as the old 2wlkeq before.
@digama0 What do you think/propose? Can I leave the two "variants" 2wlkeq and 2wlkeqdv as in this pull request, or should I remove 2wlkeqdv  (it is currently only used to prove the new 2wlkeq, so I have to prove the new 2wlkeq directly)?

digama0:
@benjub I think "less widely applicable" is a strong claim. It's still logically applicable, but it might require some alpha renaming which should only be a few steps. Importantly, this is only a question of total length of proof, and that is informed by actual use of the theorem. This theorem is only used in one place, so tailoring it for use there is not necessarily a bad thing. If and when it acquires new uses, the statement may change to support the variation in use case. This is the way it's always been. I knew that someone would make a remark on these lines as soon as any part of my proposal bore fruit in actual set.mm practice, but I explicitly indicate the exact length of this "slippery slope" so it is not a strong argument IMO.

digama0:
@avekens I suggest not having `2wlkeqdv` at all. If you do need both versions, with x,y and x,x, then it should just be the original bundled version. If you only need the x,y version, then you can either have the original bundled version or you can add the x,y DV condition (for your own convenience and to make the $d list shorter). If you only need the x,x version, then you should have `2wlkeq` with x,x and no x,y version. If you literally have two versions of the same theorem which differ only in bundling, I would suggest having one bundled theorem, unless one of the versions is notably different in its proof (for example, if it is trivial).

The goal here is to not bloat set.mm too much with unbundling concerns, but doing some low cost optimizations when reasonable.

avekens:
@digama0 thank you for your reply. 
Since I only need ~ 2wlkeq (the x,x version) at the moment (for ~ usg2wlkeq , and also generally see no need for ~ 2wlkeqdv, I will delete ~ 2wlkeqdv and change the prove of ~ 2wlkeq not using ~ 2wlkeqdv.

benjub:
@digama0 I agree on this theorem, and I agree that I was voluntarily
provocative.  Though I had understood from the mailing list that bundled
theorems are not "culprits" by themselves, it is only the uses of their
degenerate instances which are "bad". E.g.,
  pred $p |- Pred x y $= ? $.
is fine, but according to you, one should not use, in a later proof, an
instance of ~ pred of the form ` Pred x x ` . Therefore, @avekens 's remark
"so that 2wlkeq would not appear on Mario's list anymore" above surprises
me.

avekens:
2wlkeqdv is removed now.

avekens:
@benjub 2wlkeq in its previous, bundled form was used as degenerate instances in usg2wlkeq, so it appearded in Mario's list (as far as I understood).

PR #863, opened 2019-05-21
Add ancomass.

giomasce:
A little theorem that at some point will be used by my program mmpp.


benjub:
@giomasce This looks like a useful lemma. Did you try min.cmd with it ?

nmegill:
By the way the closed form of your ancomass is an12.  There are several theorems of the form anNN where the first N (position) is in an outer scope position and is swapped with the second N in an inner scope position.  Other examples are an32, an31, an13.

giomasce:
@nmegill Ah, I had not found `an12`. I will update the proof using it. The reason why I need this in inference form is that I plan to use it as a step for a proof generator in mmpp, and since mmpp is already quite verbose in generating proofs I try to make standard steps as tailored over its need as possible.

Issue #847, opened 2019-05-09
Define iset.mm recursive definition generator


jkingdon:
Now that iset.mm has http://us2.metamath.org/ileuni/tfri1.html , http://us2.metamath.org/ileuni/tfri2.html , and http://us2.metamath.org/ileuni/tfri3.html , the next step is the recursive definition generator which makes it easier to define recursive functions given an initial value and characteristic function (that is, function for giving the value at a successor).

The current generator in set.mm, http://us2.metamath.org/mpeuni/df-rdg.html , is based on case elimination, in particular between the limit ordinal and successor ordinal cases. The initial value could presumably be handled with `( <. (/) , I >. u. something )` where `something` is what we end up using for successor/limit.

So the basic question is: is there a practical way to come up with an expression which combines limit and successor? It would presumably be something along the lines of "apply the characteristic function for all ordinals less than the current one, then take the union of all the results".

Another path would be to go down the path of trying to define finite recursion next, as it would only need a successor case not a limit case. Presumably this would be via a new notation such as `frec(𝐹, 𝐴)` rather than using `(rec(𝐹, 𝐴) ↾ 𝜔)` as in set.mm finite recursion theorems such as http://us2.metamath.org/mpeuni/frsuc.html . And if this is the chosen path, is there a way to avoid re-proving http://us2.metamath.org/ileuni/tfri1.html and the like for `𝜔` (or maybe for arbitrary ordinals) in place of `On`?

By way of motivation, the key thing we want out of this is positive integer addition - http://us2.metamath.org/mpeuni/df-pli.html - which is part of the construction of real numbers.

digama0:
For increasing and continuous ordinal functions, it's not so hard; it's like you said - define `G(x) = sup y<x F(G(y))`. That should be enough for most of the applications in set.mm, although there is a caveat for ordinal exponentiation, which is not increasing when the base is 0 and is handled by a case analysis. If you skip the case analysis, then you get 0^alpha = 1 for all alpha. Maybe this is fine, although you can still basically recover the traditional function using `a^b = { y | ( E. x x e. a \/ a = (/) ) /\ y e. a ^' b }` where `a ^' b` is the inductive definition.

[...]

PR #865, opened 2019-05-24
shorten proof of tfri, iset.mm simplifications + additions

digama0:
@jkingdon This PR has some new ways of proving that a function is defined everywhere, along the lines of https://github.com/metamath/set.mm/issues/847#issuecomment-495489108 . Specifically, if the range of a function is a set and the input is a set, then the function value exists (relrnfvex). In particular this applies to show that ``( f ` x )`` is unconditionally a set.  Also, for proper class functions there is a new predicate (or rather a new use of an old predicate): `sefvex` says

    |- ( ( `' F Se _V /\ A e. _V ) -> ( F ` A ) e. _V )

and the assumption `` `' F Se _V``, read "F is set-like", means that F is a class function that sends sets to sets. It's equivalent to the assertion ``A. x ( F ` x ) e. _V``, and it will be provable in basically every case when you have a mapping abstraction with some expression that's always a set depending on the input (for example `x |-> suc x` or `x |-> ( ~P ~P x u. I )` etc.)

Given this, I've used the assumption ``( F ` x ) e. _V`` as the precondition to `tfri`, in addition to simplifying the proof a bit. If you prefer the original assumption I can restore it; I don't think the proof simplifications rely on this part of it.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -33977,6 +33977,15 @@ $)
     ( wcel cpw cvv csn pwexg wss snsspw ssexg mpan syl ) ABCADZECZAFZ
     ECZABGOMHNPAIOMEJKL $.
 
+  ${
+    snex.1 $e |- A e. _V $.
+    $( A singleton whose element exists is a set.
+       (Contributed by NM, 7-Aug-1994.)
+       (Revised by Mario Carneiro, 24-May-2019.) $)
+    snex $p |- { A } e. _V $=
+      ( cvv wcel csn snexg ax-mp ) ACDAECDBACFG $.
+  $}
+
   $( A singleton whose element is a proper class is a set.  The ` -. A e. _V `
      case of Theorem 7.12 of [Quine] p. 51, proved using only Extensionality,
      Power Set, and Separation.  Replacement is not needed.  (Contributed by
@@ -34225,10 +34234,14 @@ $)
     wa ) ACDZBCDZOZABEAFZABGZGZCABCCHRSCDZTCDZOUACDRUBUCPUBQAIJABKLST
     KMN $.
 
-  $( An ordered pair of setvar variables is a set.  (Contributed by Jim
-     Kingdon, 24-Sep-2018.) $)
-  sopex $p |- <. x , y >. e. _V $=
-    ( cv cvv wcel cop vex opexgOLD mp2an ) ACZDEBCZDEJKFDEAGBGJKHI $.
+  ${
+    opex.1 $e |- A e. _V $.
+    opex.2 $e |- B e. _V $.
+    $( An ordered pair of sets is a set.  (Contributed by Jim
+      Kingdon, 24-Sep-2018.) (Revised by Mario Carneiro, 24-May-2019.) $)
+    opex $p |- <. A , B >. e. _V $=
+      ( cvv wcel cop opexg mp2an ) AEFBEFABGEFCDABEEHI $.
+  $}
 
   $( An ordered triple of sets is a set.  (Contributed by Jim Kingdon,
      19-Sep-2018.) $)
@@ -34585,8 +34598,8 @@ $)
        (Contributed by NM, 14-Apr-1995.)  (Proof shortened by Andrew Salmon,
        25-Jul-2011.) $)
     opabid $p |- ( <. x , y >. e. { <. x , y >. | ph } <-> ph ) $=
-      ( vz cv cop wceq wa wex copab sopex copsexg bicomd df-opab elab2 ) DEZBEC
-      EFZGZAHCIBIZADQABCJBCKRASABCPLMABCDNO $.
+      ( vz cv cop wceq wa wex copab vex opex copsexg bicomd df-opab elab2 ) DEZ
+      BEZCEZFZGZAHCIBIZADTABCJRSBKCKLUAAUBABCQMNABCDOP $.
   $}
 
   ${
@@ -34595,10 +34608,10 @@ $)
        24-Mar-1998.) $)
     elopab $p |- ( A e. { <. x , y >. | ph } <->
                  E. x E. y ( A = <. x , y >. /\ ph ) ) $=
-      ( vz copab wcel cvv cv cop wceq wa wex elex sopex mpbiri adantr
-      eleq1 exlimivv eqeq1 anbi1d 2exbidv df-opab elab2g pm5.21nii )
-      DABCFZGDHGZDBICIJZKZALZCMBMZDUFNUJUGBCUIUGAUIUGUHHGBCODUHHRPQSE
-      IZUHKZALZCMBMUKEDUFHULDKZUNUJBCUOUMUIAULDUHTUAUBABCEUCUDUE $.
+      ( vz copab wcel cvv cv cop wceq wex elex vex eleq1 mpbiri adantr exlimivv
+      wa opex eqeq1 anbi1d 2exbidv df-opab elab2g pm5.21nii ) DABCFZGDHGZDBIZCI
+      ZJZKZASZCLBLZDUGMUMUHBCULUHAULUHUKHGUIUJBNCNTDUKHOPQREIZUKKZASZCLBLUNEDUG
+      HUODKZUQUMBCURUPULAUODUKUAUBUCABCEUDUEUF $.
   $}
 
   ${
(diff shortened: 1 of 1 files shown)


jkingdon:
Really glad to see this. Will take me a bit to look at it closer (probably later today) and I see a few little style points (for example, since this `opex` has different hypotheses than the set.mm `opex` our policy has been that we would give it a different name). But generally, anything to simplify transfinite recursion, and make it easier to show the existence of function values, strikes me as very much useful.

As for `` ( F ` x ) e. _V `` versus the other way of stating that condition, not sure I have an opinion yet, will see if anything comes up when I take a closer look.

digama0:
For `opex`: That's the usual naming convention from set.mm, which is to say, if a similar operation required sethood assumptions and was taking them in inference form, then it would be called `<operation>ex`, for example `unex`. I'm a bit skeptical of a naming convention that avoids the natural names for things when there is no theorem in that spot. I know while working on this there were several cases where I tried and failed to guess the name based on my set.mm knowledge.

nmegill:
There is a minor Travis failure (date at top of file needs to be updated).

digama0:
I thought the policy was to not touch that. Or is iset.mm different in that respect?

nmegill:
The problem with set.mm was that two PRs in the queue were often edited with different dates at the top mainly due to US vs. EU time zone shifts, causing merge conflicts.  This happened often enough that we decided to ignore the top date, with a note not to change it if there were other PRs in the queue.  We changed the script to use 'verify markup * /top_date_skip' to prevent Travis failures resulting from this.

This change wasn't done for iset.mm because there are fewer users and it hasn't been a problem.  I suppose we should do 'verify markup * /top_date_skip' for iset.mm also for consistency.  I took a look at the scripts, which are a little bit obscure because they call other scripts with arguments, and without a careful analysis couldn't figure out confidently what I needed to change to make this happen.  Perhaps @david-a-wheeler could make this change.  In the meantime, could you update the date at the top of iset.mm?  Thanks.

jkingdon:
The rule about changing names when we change set.mm theorems has been stated
in iset.mm for a while now: "Theorems which are different [than set.mm] should be named
differently." Its strongest proponent was @david-a-wheeler so I'll ask him:
do you have an opinion about whether the rule should be relaxed by saying
that `added or removed sethood hypotheses "don't count" in this accounting`?
I could live with what the group decides either way.

On a semi-related naming topic, in case you didn't notice it there is a table of "missing" set.mm theorems at http://us2.metamath.org/ileuni/mmil.html#setmm .  Quite aside from what we decide about snex naming, that strikes me as a useful thing.

With respect to sopex, I assume this is just a matter of applying the rule
"Usually I will also automatically accept a _new_ theorem that is used to
the new theorem, not including the acknowledgment) decreases as a result."
I haven't tried it both ways, so I don't know what answer this gives
with respect to sopex (either with or without the inference version of
opexg).

Anyway, on to more important topics. Love the addition of the new ways
of showing that a function value exists (`fvss`, `fvssunirng`, `relfvssunirn`,
`relrnfvex`, and `sefvex`). Up until now the only way has been to show
we are evaluating the function within its domain. I suppose that will
continue to be the most common way, since it does have a nice obviousness
to it (and is often quite easy), but as we've seen with transfinite
recursion, having more ways to show that function values exist is likely
to be useful.

The shortenings of the transfinite recursion theorems I only looked at
briefly but they look good (using deduction form more, for example, makes
a lot of sense in a set of lemmas this complicated). One question: since
tfrlem[1-9] (plus tfrlem3a) are taken verbatim from set.mm, should we make
those changes in set.mm as well?

That leaves the change from `∀x ∈ On ∀f(f Fn x → f ∈ dom 𝐺)` to `(𝐺‘x) ∈ V` as the condition for what a recursion rule can look like. Because the former is not as severe a condition (requires only that the recursion rule be defined for ordinals, not for all sets) it strikes me as more intuitive. But I must admit I don't really see the implications of choosing one way or the other. If we can show `(𝐺‘x) ∈ V` for all the recursion rules we care about, well it is shorter to state. If we do want to go forward with this, there are a few nits: (1) restate the comment "condition that the recursion rule 𝐺 is defined for all ordinals" which is no longer accurate, and (2) seems like `x` would be more clearly expressed as `f` or something else which looks like a function. I also find it slightly jarring to not have the `A. x` expressed explicitly, but maybe that is just something I should get used to, as I do understand that the quantifier is unnecessary here.

david-a-wheeler:
@jkingdon - as far as naming goes, I think the key is the naming convention (whatever it is) should make it easy to work with over time, including transfers from set.mm to iset.mm.  I can see an argument either way.

If there's a proof in set.mm, and I copy it over to iset.mm, what do we think should happen?  If the names in what's used are all the same, I would think it should "just work".  I think that if there are different hypotheses there should be some tweak in the name, to indicate that a simple copy *won't* just work.  But maybe that's not the right way to think about it... I'm torn myself.

digama0:
@david-a-wheeler I don't think it's reasonable to expect "binary compatibility" for all but the smallest proofs. I think a better approach is if you took the mmj2 worksheet for a set.mm theorem and copied it to iset.mm. In this case added hypotheses would be inserted as steps (and hopefully proven automatically), so that it would still "just work" in this case. If the statements change slightly, the proof would break, but at least the fix would be relatively localized and mmj2 would already be "trying to apply" the right theorem. If a theorem with that name doesn't exist at all, then you have no assistance on finding the replacement, and you just have to continue the proof from that point.

So my preference would be to have iset.mm theorems be named as if they were in set.mm and we were coming across these theorems the first time. A theorem like one direction of a de Morgan law would not be named as if it was both directions, but maybe there would be an `r` stuck in one of them. I don't think there should be a naming trace for assumptions that "automatically" go with a theorem, i.e. there is only one reasonable option for them. If a theorem holds under a variety of assumptions, like the variations on `fvex` in this PR, then the naming should indicate how they differ. (Although I didn't prove it in this PR, I think there is room for a theorem actually called `fvex`, namely `|- F e. _V & |- A e. _V => |- ( F ` A ) e. _V`.)

jkingdon:
As for the naming thing, in the work I have been doing I have been surprised at how many proofs (and whole sections, sometimes) can be copy-pasted from set.mm to iset.mm without any changes at all. So there is some degree of "binary compatibility", although that isn't exactly relevant to a case like snex, since that won't exist exactly as stated in set.mm no matter what we do.

I think I find the argument about how mmj2 will be able to fill in hypotheses (and sometimes their proofs) fairly convincing. So I would suggest changing "Theorems which are different should be named differently." to "Theorems which are different should be named differently (although if additional hypotheses are needed in iset.mm that does not call for changing the name)."

One more point which is perhaps understood or perhaps worth stating more explicitly is that it is my preference to keep set.mm and iset.mm similar where we can. For example, if a theorem exists both places but the name in set.mm isn't great, I'd rather either just keep that name for iset.mm, or change it in both files together. This is mainly to make it easier to copy theorems, but also to generally help people browse proofs, find theorems, write proofs, etc.

david-a-wheeler:
As for the naming thing, in the work I have been doing I have been surprised at how many proofs (and whole sections, sometimes) can be copy-pasted from set.mm to iset.mm without any changes at all. So there is some degree of "binary compatibility", although that isn't exactly relevant to a case like snex, since that won't exist exactly as stated in set.mm no matter what we do.

I think I find the argument about how mmj2 will be able to fill in hypotheses (and sometimes their proofs) fairly convincing. So I would suggest changing "Theorems which are different should be named differently." to "Theorems which are different should be named differently (although if additional hypotheses are needed in iset.mm that does not call for changing the name)."

david-a-wheeler:
(Sorry for weird first version, the comment on GitHub is now correct.)

digama0:
@jkingdon Feel free to make any changes you see fit to the theorems here. Sorry I didn't get around to fixing the tfri comment; my metamath time these days is limited.

nmegill:
I updated the date at the top of iset.mm in 19c5c82 to fix the Travis verify markup faillure.

jkingdon:
@digama0 no worries, I'll make myself a note to fix the comments and maybe also the other things I mentioned. I very much appreciate the time that you do have, and also your suggestions of what might make it easier or harder to work with iset.mm.

PR #868, opened 2019-05-25
 Added new geometry proofs

tirix:
With this pull request, I'm adding a notion of complex vector spaces. 
I expect to grow the number of theorems for complex vector spaces as the needs for geometry grows.


set.mm
```diff
@@ -281762,6 +281845,115 @@ a tuple of a topological space (a member of ` TopSp ` , not ` Top ` )
...
+                         Complex vector spaces
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  $c CVec $. $( Complex vector space $)
+
+  $( Complex vector space. $)
+  ccvs $a class CVec $.
+
+  $( Define a complex vector space, which is just a complext left module and a
```
david-a-wheeler: Nit.. complext should be complex.

set.mm
```diff
@@ -345636,15 +346052,167 @@ the identity of betweenness (Axiom A6).  (Contributed by Thierry
...
+        BNVCZVVKNVCZUWBALMVWFVCZVWNGVCVWOVXDNVCVXFVXMALMVWFGVUPFNDIVWNTUBVUQUEU
+        AVWIVWKVUNVUOVWSYMAUYJVXNVWNGAVUEVUMVULUYJVXNVFVUIVUNVUOLMVUPFIVUQUEYNX
+        BYHAVXLVWOVVKVXDNAVXLVWPBNVCZVWOACVWPBNVAYHAIUVMVDZVVLVWQVXOVWOVFAVWJVX
+        PVWKIUVNYDUIVWTDEINBVWOTUGUAUVOXBYKUTYGYOAVVEVWMVVGVVLVXMUWBVFVVHVXAUCU
+        IDINCPBTUAYSYCYKYIYPYQYRUVQYRUVPUWFUWQVBUWNUWGUWCUWNVFUWEUWPUWBUWCUWNUW
+        DGUVRUVSUVTXLADGVBHIJNXQPQCRSTUAUBUCUDVUIVXAUWAXM $.
     $}
 
 $(
+    @{
```
david-a-wheeler: Why is this commented out?

set.mm
```diff
@@ -345171,6 +345369,224 @@ The second section ("Tarskian geometry") develops the synthetic treatment of
...
+      $( Commutativity law for lines.  Part of theorem 6.17 of [Schwabhauser]
+         p. 45. (Contributed by Thierry Arnoux, 17-May-2019.) $)
+      tglinecom $p |- ( ph -> ( P L Q ) = ( Q L P ) ) $=
+        ( vx co wcel wa adantr tglnssp sselda cstrkg simpr lncom impbida eqrdv
+        cv ) ANCDGOZDCGOZANUFZUGPZUIUHPZAUJQBEFGCDUIHJIAEUAPZUJKRACBPZUJLRADBPZ
+        UJMRAUGBUIABEFGCDHJIKLMSTAUJUBUCAUKQBEFGDCUIHJIAULUKKRAUNUKMRAUMUKLRAUH
+        BUIABEFGDCHJIKMLSTAUKUBUCUDUE $.
+
+      $d A x y $.  $d B x y $.  $d G y x $. $d I x y $.  $d L x y $.
+      $d P x y $.  $d Q x y $.  $d ph x y $.
```
david-a-wheeler: I love that you already have a nontrivial amount in chapter 6.

tirix: Scott Fenton went up to 6.21, so with 6.18, I've already almost covered what he had, but what I really want to cover is the chapter 14... that's still a long way to go.

set.mm
tirix: Because I still need to work on that proof. Corner cases are not handled, and one of the conditions to apply the lemma still evades me.

set.mm
tirix: This will be fixed with the next commit!

nmegill: > This will be fixed with the next commit!

I  fixed complext -> complex in 57b7d75

Issue #866, opened 2019-05-24
Ring of integers


avekens:
I need the ring of integers ` ( CCfld |`s ZZ ) ` as (counter) examples for some theorems in linear algebra. There is already some material in different places (esp. mathboxes):

- First appearance in section "10.11 The complex numbers as an algebraic extensible structure"
- Section "The ordered commutative ring of integers" in TA's mathbox contains basic properties.

I would like to move section "The ordered commutative ring of integers" in TA's mathbox into the main part, maybe as section "10.8.6 Ring of integers" after "10.8.5  Left regular elements. More kinds of rings" (since ` ( CCfld |`s ZZ ) e. LPIR `, see ~ zlpir), and to add more theorems I have not commited yet.

Any objections?

nmegill:
I have no objections.

tirix:
No objections here either!

avekens:
With PR #872/#876 I moved the section "The ordered commutative ring of integers" in TA's mathbox into the main part, as section "10.11.2  Ring of integers" (in contrast to the expected/planned section 10.8.6) , because CCfld is defined in section 10.11.1 only. 

With hindsight  I think it would be practicable to give it an own definition, maybe 
df-rngz $a |- RingZ = ( CCfld |s ZZ ) 

Unfortunately, there is already a definition df-zrng $a |- ZRing = ... in Mario's Mathbox (subring of integral elements in a ring), so ZRing cannot be used for the "Ring of Integers" (or Mario's definition must be changed).

What do you think about this?

@benjub I added the word "unital" in the comment of the header of section "Ring of integers", as you propoed in #872. I put it in square brackets, however, because my text is a citation taken from Wikipedia.

benjub:
@avekens As for the HTML symbol (and also in the tokens), I would use "Ring/ring" instead of "Rng/rng".  The reason is that the term "rng" has become quite popular for "non-necessarily unital rings" (they are "rings" that lost the "i" of "identity").  I don't know what @nmegill and @digama0 think of it.

avekens:
@benjub Thank you for this hint - I didn't know that. But also in Wikipedia the differentiation is made, see [https://en.wikipedia.org/wiki/Rng_(algebra)](https://en.wikipedia.org/wiki/Rng_(algebra)). So my suggestion woud be the following: 

Definition: `df-zzring $a |- ZZRing = ( CCfld |``s ZZ )`
[...]

PR #872, opened 2019-05-25
Develop

avekens:
Theorems of TA's and AV's Mathox moved, esp. for ring of integers (see also #866):
* ress0g, mul2lt0rlt0, mul2lt0rgt0, mul2lt0llt0, mul2lt0lgt0, mul2lt0bi moved from TA's mathbox to main set.mm
* zzsinvg/zrnginvg, zrngrng moved from AV's mathbox to main set.mm
* zzsbase, zzsplusg, zzsmulg, zzsmulr, zzs0, zzs1 renamed and moved from TA's mathbox to main set.mm

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15182,6 +15182,7 @@ New usage of "erngset-rN" is discouraged (3 uses).
 New usage of "eu1OLD" is discouraged (0 uses).
 New usage of "eu2OLD" is discouraged (0 uses).
 New usage of "eu3OLD" is discouraged (0 uses).
+New usage of "eu5OLD" is discouraged (0 uses).
 New usage of "euanOLD" is discouraged (0 uses).
 New usage of "euexOLD" is discouraged (0 uses).
 New usage of "eufOLD" is discouraged (0 uses).
@@ -18056,6 +18057,7 @@ Proof modification of "equncomiVD" is discouraged (19 steps).
 Proof modification of "eu1OLD" is discouraged (92 steps).
 Proof modification of "eu2OLD" is discouraged (118 steps).
 Proof modification of "eu3OLD" is discouraged (45 steps).
+Proof modification of "eu5OLD" is discouraged (38 steps).
 Proof modification of "euanOLD" is discouraged (106 steps).
 Proof modification of "euexOLD" is discouraged (32 steps).
 Proof modification of "eufOLD" is discouraged (73 steps).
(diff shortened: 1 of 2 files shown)


nmegill:
@avekens A merge conflict arose, could you look at it?  Thanks.

avekens:
Puh, there was an accepted pull request from Wolf Lammen, in which a lot of reformattings within my Mathbox were perfomed! Since I also did a rewrap before my pull request, I fear a lot of conflicts are caused by this. I try to find an easy solution to merge these changes, but maybe I have to through away (and redo them in a clean repository) my changes and discard this pull request at worst.

But maybe someone with write access to metamath/set.mm could do it more easily (directly) than I can do (going via a fork).

wlammen:
I am awfully sorry for this. For rewrapping I use

./metamath 'read set.mm' 'write source set.mm /rewrap' \
  erase 'read set.mm' 'save proof */compressed/fast' \
  'verify markup */file_skip' 'verify proof *' \
  'write source set.mm' quit

I thought this is a standard procedure, maybe I am not quite up-to-date here. Hints are welcome.
git allows reverting a particular commit by
git revert d8e06f4
(the number refers to the rewrap commit)

avekens:
@wlammen it's not you fault - the rewrapping you performed is absolutely fine (I do it the same way). I think the general problem is that we are working on one big file (set.mm) in parallel, and this time there were 3-4 parallel pull requests (usually we have 1-2 at the same time). I do not now how this problem can be solved in general...

benjub:
@avekens It might be safer to specify that \Z is the initial object in the category of *unital* rings (you didn't write "unital") Remark: in the category of non-necessarily unital rings, the zero ring is the zero object (i.e. both initial and terminal).

nmegill:
@wlammen @avekens I'm not sure just where the conflicts arose, but if we were to accept set.mm only if it is rewrapped - i.e. have Travis do a local rewrapping and make sure no differences arose - would that have solved this problem?

avekens:
Although there were several places with merge conflicts (about 5-10? I didn't count them), it was relatively easy to solve them. However, the whole procedure cost me mor than half an hour...

@nmegill My assumption was false, most of the conflicts didn't come from the rewrapped comments, but had seveersal causes (conflicts in proves, headers, etc.). So I think from this case we cannot derive an improvement.  To check for rewrappings, as you suggested, might be counterproductive (many rejections because of inconsostent rewraps).

Fortunately, everything loks fine now.

PR #874, opened 2019-05-26
replacement for mo2

wlammen:


benjub:
I would keep both theorems in the main part, labelling them either mo2 and mo2v, or mo2f and mo2 (both conventions exist: the suffixed "v" is for "disjoint Variable condition" and "f" is for "non-Freeness hypothesis").

nmegill:
> I would keep both theorems in the main part...

In general we include both dv and nf versions only when there is a specific need - e.g. when we have only a dv version (with some usage) but a specific need for the nf version arises, or when we have only an nf version (with heavy usage) but we notice that most uses require only the dv version, which we then add especially if the resulting proof shortening is significant.

Typically we don't start off including both versions if only one is needed.  This explains -/v vs. f/- conventions - the missing one gets the v or f when it is added later.  Maybe for consistency, a single convention could be adopted some day via a big renaming project, but I din't think it's urgent right now.

wl-mo1 (which I would rename mo2v) is a special case that is interesting because it avoids ax-11, so for that reason might be worthwhile to move to main even if it isn't used.  In that case there may be uses of mo2 that don't require the nf and whose proofs could be shortened with to mo2v.  I'll let Wolf decide.

wlammen:
All usages of eu3 and by far the most usages of mo2 benefit from a dv version of these theorems.  You will find extra lines of nfv to provide the necessary F/ y ph for the current theorems, so the line count in proofs will reduce by 20 or so. The reduction in proof size is less impressive, since nfv is cheap.  The total size will likely not decrease, because the proof of wl-mo1 eats up all the small shortenings.

Apart from ax-11, there is another idea that appeals me.  I was actually at first a bit disappointed to learn that this section starts out with E! and not with E* as the foundational definition.  Since E! x ph -> E* x ph, E! appears as a specialized version of E*.  And a usual procedure in mathematics is to lay foundations with the general ideas before starting to specialize.  Of course, E! and E* are tightly coupled via E. x ph, so in this particular case no harm arises from proceeding the other way round.  However, I still sense some nagging feeling when browsing this section.  wl-mo1 is exactly the definition of E* that you could have started with.  So this theorem balances the role of E! and E* again.

I am currently very busy at work, and will be working on metamath only at weekends.  Don't expect quick solutions.

Finally I'd like to thank you both for taking time to review my ideas.

@nmegill bisym and bi3ant are two theorems I once introduced in the hope to profit from some systematic approach in proving nearby theorems.  This has never worked out.  So these theorems are leftovers barely understandable to any 'uninitiated'.  I propose to remove them from the database.

nmegill:
> bisym and bi3ant .... I propose to remove them from the database.

@wlammen Feel free to do so (by renaming them to *OLD).  Since they aren't used, it's probably optional to list them at the top of set.mm.

PR #880, opened 2019-05-28
Rename "cross product" to "Cartesian product"

david-a-wheeler:
The term "cross product" is confusing and misleading.
I think the far more common term for what is meant in df-xp
is "Cartesian product"; for example, see:
https://en.wikipedia.org/wiki/Cartesian_product

The term "cross product" is more typically used to mean
"a binary operation on two vectors in three-dimensional space ( R 3 )".
https://en.wikipedia.org/wiki/Cross_product

Even set.mm uses the term "Cartesian product" in some places.

Let's use the single term "Cartesian product" when that's what
we mean, to reduce the possibility of confusion and to be more consistent.

Note that an "infinite Cartesian product" is something else;
for that, see df-ixp.


david-a-wheeler:
@nmegill - this is consistent with the book.

tirix:
“Cartesian product” is also the naming I’m used to.

PR #882, opened 2019-05-29
eu3v will finally replace eu3

wlammen:
There is no usage of eu3 that cannot provide proper dv conditions. So go for this slightly modified one, and let eu3 finally die as eu3OLD without any reference. The advantage is
- shortened proofs;
- fewer dependencies on axioms.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15182,6 +15182,7 @@ New usage of "erngset-rN" is discouraged (3 uses).
 New usage of "eu1OLD" is discouraged (0 uses).
 New usage of "eu2OLD" is discouraged (0 uses).
 New usage of "eu3OLD" is discouraged (0 uses).
+New usage of "eu3OLD" is discouraged (0 uses).
 New usage of "eu5OLD" is discouraged (0 uses).
 New usage of "euanOLD" is discouraged (0 uses).
 New usage of "euexOLD" is discouraged (0 uses).
@@ -18057,6 +18058,7 @@ Proof modification of "equncomVD" is discouraged (53 steps).
 Proof modification of "equncomiVD" is discouraged (19 steps).
 Proof modification of "eu1OLD" is discouraged (92 steps).
 Proof modification of "eu2OLD" is discouraged (118 steps).
+Proof modification of "eu3OLD" is discouraged (108 steps).
 Proof modification of "eu3OLD" is discouraged (45 steps).
 Proof modification of "eu5OLD" is discouraged (38 steps).
 Proof modification of "euanOLD" is discouraged (106 steps).
--- a/discouraged
+++ b/discouraged
@@ -5838,8 +5838,6 @@
 "erngset-rN" is used by "erngbase-rN".
 "erngset-rN" is used by "erngfmul-rN".
 "erngset-rN" is used by "erngfplus-rN".
-"eu3" is used by "eu5OLD".
-"eu3" is used by "mo2OLD".
 "exatleN" is used by "cdlema2N".
 "exidu1" is used by "cmpidelt".
 "exidu1" is used by "exidresid".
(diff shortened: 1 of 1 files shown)


wlammen:
Travis won't allow me to mark eu3 as old as long as I have to update the name in mo2OLD and eu5OLD as well.  Thinks this is a 'new' usage of a theorem that has a "New usage is discouraged." annotation attached to it.
Is there a workaround?

wlammen:
Another mystery is that eunex is killed all the time. Commit 20d2732 shows that I clearly added a proof, which is silently removed the next commit f7e3747 again. What is going on here?

benjub:
Im a bit sad that the more general theorem disappears. Why not keep both ?
By the way @wlammen you might want to chime in the associated thread on the
google group.

nmegill:
One strange thing going on is that in commit 4534bb8, the discouraged file has:
Proof modification of "eu3OLD" is discouraged (108 steps).
Proof modification of "eu3OLD" is discouraged (45 steps).
At most one line per theorem should be there, and certainly not 2 lines with different # steps.
I regenerated the discouraged file from the 4534bb8 set.mm locally (using metamath.exe) and only the 2nd line above appeared.  How did you generate the discouraged file?

wlammen:
@benjub In my eyes, eu3 is a sort of a technical theorem, where E* x ph is replaced with a complex, equivalent expression.  This helps perhaps in some sophisticated proofs, but it does not really say more than eu5 in conjunction with either mo2/mo2v does.  If there is no application of eu3 anyway, why keep it then?  The comment in df-eu hints to eu3 as one alternative definition.  If there is actually literature based upon such a definition, then I'd be convinced to keep eu3, perhaps with a reference to a book, or so...

There is some other feeling I have about mo2/euf vs. mo2v/df-eu.  For me, the latter have a cleaner structure. They keep the set variables and the wff ph as much as possible apart. This is an advantage if you want to explore the properties of existential uniqueness. So I view the former pair as a reminder of certain wff subexpressions that one might come across accidentially.  Then it is nice to be able to recognize them and know their semantics, without being too much bothered with technical details. But again: for this mo2 is sufficient.

[chime in...] I simply don't have the time to take part in any detailed discussion.  I usually get the idea of what is going on, but elaborating on any detailed question is beyond my capacities. I earn my money from work not particularly related to logic (or mathematics).

wlammen:
@nmegill I used this script:

./metamath 'read set.mm' 'set width 9999' 'show discouraged' quit | grep '^SHOW DISCOURAGED.' | sed -E -e 's/^SHOW DISCOURAGED.  ?//' | LC_ALL=C sort > discouraged

nmegill:
@wlammen 

> @nmegill I used this script:
> 
> ./metamath 'read set.mm' 'set width 9999' 'show discouraged' quit | grep '^SHOW DISCOURAGED.' | sed -E -e 's/^SHOW DISCOURAGED. ?//' | LC_ALL=C sort > discouraged

This is also what I used, and it gives something different:

$ diff discouraged-wl discouraged-nm
5840a5841,5842
> "eu3OLD" is used by "eu5OLD".
> "eu3OLD" is used by "mo2OLD".
15184,15185c15186
< New usage of "eu3OLD" is discouraged (0 uses).
< New usage of "eu3OLD" is discouraged (0 uses).
---
> New usage of "eu3OLD" is discouraged (2 uses).
18061d18061
< Proof modification of "eu3OLD" is discouraged (108 steps).

I used version 0.177 of metamath.exe but there have been no recent changes in the discouraged generation code.

Can you run this again on set.mm 4534bb8 to see if it is repeatable?  Something is very wrong.  In particular, a double occurrence of "New usage of "eu3OLD" is discouraged (0 uses)." should be impossible based on what the code is supposed to do.

nmegill:
@wlammen @benjub I don't have a problem making eu3 obsolete.  If we really need it in the future, it's relatively simple to recreate.  While we do have a number of cases of both 'v' and 'f' versions where one is unused, I have also purged unused ones in the past when I've felt that they would never be needed,  so we don't clutter up set.mm pointlessly.  While it is a judgment call, we need to weigh how hard it is to prove the more general version if it becomes needed (simple in this case, with many examples of how it is done) vs. the likelihood that it will ever be useful (not likely in this case), together with whether eu3 is intrinsically interesting for some reason (I don't think it is).  So I will support WL's decision.

If you strongly feel we should keep eu3, perhaps you can move it to your mathbox for later discussion, but I'd like to get this PR resolved.

wlammen:
@nmegill I played around a bit.  The most likely explanation is that I updated the discouraged file while there were temporarily 2 eu3OLD theorems in the file. I remember I renamed eu3 to eu3OLD before deleting it finally

benjub:
@nmegill Ok. I'll have a second look and I might make a "backup" of "eu3f" in my mathbox.

PR #890, opened 2019-05-31
added three new proofs to Jon Pennant Mathbox, moved several other pr…

TjWhale:
…oofs which were used.

I hope I have moved the proofs required to reasonable places. Let me know if any changes or additions are needed :)


benjub:
@TjWhale I would rephrase the following comment:
The integral of an exponent --> The integral of a monomial

Typos:
it's higher edge --> its
it's lower edge --> its

Its looks like some comments are not rewrapped.  I recommend that you use the command given at
https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md

In this case, the two Travis errors are dates written "2019)." instead of "2019.)"

TjWhale:
Ok I think it's in reasonable shape now. I am not sure if I put mulcncf in a reasonable place, it seems to be floating a bit.

PR #893, opened 2019-05-31
Document how to convert $d to F/_

david-a-wheeler:
Mario Carneiro posted on the Metamath mailing list
on Tue, 28 May 2019 16:34:40 -0400, a general approach
for converting $d statements into F/_ statements
(see "Re: [Metamath] Replacing $d with F/ and F/_").

This commit adds that information to the "conventions"
discussion about freeness, so that others will be able to
find that information later.


benjub:
Beginning of line 358309, you forgot something like "Imagine you have proved".  You could also remove "and A for A".
Maybe to be more explicit: "The side goal is chvar.2 |-..."

giomasce:
Good idea. I wanted to do the same!

PR #894, opened 2019-05-31
typos and add (to my mathbox) versions of theorems of the main part w…

benjub:
…ith one additionnal dv and not using ax-13

Experiment: reproduce some theorems of the main part with one additional dv and no dependency on ax-13. Many are utility theorems that are typically used with dummy variables, so the dv conditions are often innocuous. They are not meant to replace the more general version, at most be added to them.


benjub:
@nmegill Shouldn't there be a "new usage discouraged" tag for ~ dummylink and ~ idi ? They are sometimes (albeit rarely) accidentally used by "min *" or "improve all". I'll add the tag in a future PR if you agree.

nmegill:
> @nmegill Shouldn't there be a "new usage discouraged" tag for ~ dummylink and ~ idi ? They are sometimes (albeit rarely) accidentally used by "min *" or "improve all". I'll add the tag in a future PR if you agree.

No, because they are sometimes used temporarily to assist constructing proofs.  The 'minimize' command will always remove them, so they cause no permanent harm.  

dummylink has no uses currently.

It seems there are some uses of idi, which are either new proofs for which 'minimize' hasn't been run or are proofs generated by Alan Sare's virtual deduction tool.  For now, we should not change the uses of idi in AS's mathbox (they are all marked "New usage discouraged" and "Proof modification discourged" anyway), but I think all other uses should have idi removed using 'minimize'.  You are welcome to do that if you feel like it, although every few years we have typically done a 'minimize' run on the whole set.mm.  If you do 'minimize' a proof, we normally do not change the description by adding "Proof shortened by" since it is an automated step that we ordinarily expect to be done eventually anyway.

I see that suctr uses idi and suctrALT doesn't.  This was a mistake, and the labels should be swapped:  suctr should be called suctrALT and vice-versa, and the resulting suctrALT (the one using idi) should be moved to AS's mathbox.  You can do that if you want to.

nmegill:
BTW yes, you will sometimes see dummylink or idi appear during a 'minimize' run, but it is only temporary, and the final proof that results will not use either of them.

PR #895, opened 2019-06-02
Clean

benjub:
As per #894 :
* swap suctr and suctrALT, and put the latter in Alan Sare's mathbox.
* minimize/no_new_ax * cshwshashnsame , which was the only theorem of the main part using ~ idi ; the other ones are in mathboxes so I didn't minimize them.

@nmegill *Before* merging, would you let me do the following changes:
* Add both discouragement tags to isosctrlem1ALT , iunconlem2 , sineq0ALT in Alan Sare's mathbox (they use ~ idi and rely on Virtual Deduction), as well as iunconALT , sspwimpALT2 , notnot2ALT2 .  All other nearby theorems have both discouragement tags.
* set.mm currently has 435 double blank lines (found by searching the string "\n\n\n").  They don't necessarily correspond to end of (sub)sections/parts/chapters.  Can I replace them with single blank lines (doing a global \n\n\n --> \n\n) ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -8667,6 +8667,7 @@
 "isvci" is used by "hilvc".
 "isvclem" is used by "isvc".
 "isvclem" is used by "vcoprnelem".
+"iunconlem2" is used by "iunconALT".
 "iunonOLD" is used by "alephon".
 "iunonOLD" is used by "cfsmolem".
 "iunonOLD" is used by "hsmexlem5".
@@ -15901,6 +15902,7 @@ New usage of "isnvi" is discouraged (3 uses).
 New usage of "isnvlem" is discouraged (1 uses).
 New usage of "isolatiN" is discouraged (0 uses).
 New usage of "isomliN" is discouraged (0 uses).
+New usage of "isosctrlem1ALT" is discouraged (0 uses).
 New usage of "ispautN" is discouraged (0 uses).
 New usage of "isph" is discouraged (2 uses).
 New usage of "isphg" is discouraged (4 uses).
@@ -15929,6 +15931,8 @@ New usage of "isvc" is discouraged (1 uses).
 New usage of "isvci" is discouraged (3 uses).
 New usage of "isvclem" is discouraged (2 uses).
 New usage of "iswatN" is discouraged (0 uses).
+New usage of "iunconALT" is discouraged (0 uses).
+New usage of "iunconlem2" is discouraged (1 uses).
 New usage of "iunonOLD" is discouraged (4 uses).
 New usage of "jaoded" is discouraged (1 uses).
 New usage of "jaoi2OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> As per #894 :
>
> swap suctr and suctrALT, and put the latter in Alan Sare's mathbox.
minimize/no_new_ax * cshwshashnsame , which was the only theorem of the main part using ~ idi ; the other ones are in mathboxes so I didn't minimize them.

Actually I don't see any harm in minimizing proofs in mathboxes - our periodic global 'minimize' runs will do that anyway, so it would get us one step ahead.  I used to require it for mathbox submissions, but people complained it was too much trouble.

So you can go ahead and do that for idi uses, except AS's mathbox (where the discouraged tags will prevent it anyway).

Even manual proof shortenings (with a "Proof shortened by..." acknowledgment) have often been done inside others' people's mathboxes, and I have encouraged that.  If someone doesn't want their mathbox proof touched for some reason, they can add "Proof modification is discouraged".

The only exception I can think of is when someone is actively working on a section with proofs in progress, which can often be determined by the date, which might lead to merge conflicts if they also shorten their proof (in a different way).  In practice this hasn't been a problem, but if in doubt the proof shortening can be coordinated with the mathbox owner.

> @nmegill _Before_ merging, would you let me do the following changes:
> 
> * Add both discouragement tags to isosctrlem1ALT , iunconlem2 , sineq0ALT in Alan Sare's mathbox (they use ~ idi and rely on Virtual Deduction), as well as iunconALT , sspwimpALT2 , notnot2ALT2 .  All other nearby theorems have both discouragement tags.

Yes, please do.

Eventually we will want to think about how we might want to clean up or reorganize AS's mathbox since he is no longer with us to do that, but there's no urgency.  His daughter has expressed a desire for his mathbox to remain as his legacy.

> * set.mm currently has 435 double blank lines (found by searching the string "\n\n\n").  They don't necessarily correspond to end of (sub)sections/parts/chapters.  Can I replace them with single blank lines (doing a global \n\n\n --> \n\n) ?

I have been placing double blank lines between major sections (e.g. between mathboxes) to better delineate them visually, although which level (section, subsection, etc.) is called "major" is not defined and I've been leaving that up to the user e.g. inside of mathboxes.  We can discuss removing double blank lines everywhere, but for now I think it is best to leave the ones between sections alone.

I agree that double blank lines elsewhere should be removed.  Since section headings begin with "$(" at the beginning of the line, you can go ahead and change "\n\n\n " to "\n\n " globally.  I think that will work.

"write source .../rewrap" changes triple blank lines to double ones, so triple blank lines shouldn't be a concern.  I suppose that I could have it do change "\n\n\n " to "\n\n " as well, but for now that is low priority.  I added a to-do item to my list.

benjub:
The previous comments have been addressed.  I think it's now ready to merge.

PR #897, opened 2019-06-04
Theorems for commuting class substitution.

giomasce:


benjub:
@giomasce, I just noticed that your sbdel was already in my mathbox as bj-sbtr.  @nmegill, is it reason enough to move it to the main part, right after sbt ?  As an approximate converse of sbt, it looks like a natural statement to add.

giomasce:
No problem at all, thanks for noticing it. I do not even care at what its name becomes.

nmegill:
> @nmegill, is it reason enough to move it to the main part, right after sbt ?

Strictly speaking, no, because it isn't needed for anything.  But it is a little interesting, so go ahead and move it.

GM's proof becomes identical with BJ's after minimizing.  Note that it is conventional to put not-free hypotheses before the others, like BJ does.  The name "sbdel" isn't bad, but I would go with BJ's "sbtr" since an "r" suffix is often used to indicate "reverse".  Actually either name is OK.

As for where to place it, the proof needs sbid2, which comes after sbt.  This can be seen with "show trace_back bj-sbtr", which I often  use to guide theorem placement.  Alternately, the first use of sbt is after sbid2, so we could also move sbt after sbid2 if we want them to be together.  I'll let you decide.

giomasce:
That theorem is used by my generator which converts ND proofs generated by GAPT (which in turns uses tools like E, prover9 and similar) to Metamath format. The generator is not yet ready, but at some point I will definitely use it to prove new things, so that theorem too will be used. The generated proofs will probably have the usual problem of being much larger than what a human would create by hand, so it's not clear whether it will be sensible to add them to `set.mm`.

As for the order of the hypotheses, I fully agree, but unfortunately for the moment `mmpp` does not make me able to choose their order, so sometimes it is not really the most sensible one. Does `metamath` have a tool to reorder the hypotheses in a theorem?

nmegill:
> Does `metamath` have a tool to reorder the hypotheses in a theorem?

From "help save proof":  "SAVE PROOF * / EXPLICIT / PACKED / FAST will allow the order of $e and $f hypotheses to be changed without affecting any proofs."

benjub:
Ok, I'm going to add bj-sbtr (as sbtr).  In the meantime, I have actually added a closed form of it (bj-sbtrt) and proved bj-sbtr from it.  I find that it makes things clearer.  May I add both?  Or do you prefer I only add sbtr ?

nmegill:
OK, add both.

PR #900, opened 2019-06-06
moving sbtrt and sbtr to main part

benjub:
As per Issue #897, move sbtrt and sbtr to the main part (and move sbt slightly later). Minimize sbdel with sbtr, so that sbdel can probably be removed (I let @giomasce do it since I'm not sure how his external program uses set.mm and don't want to break things).

@nmegill I noticed something strange: it looks like the us2 server is faster to update html pages than to "serve" newly created ones. For instance, in the previous pull request, I changed the comment of bj-sbtr to link to bj-sbtrt and I added bj-sbtrt, but right now, the page bj-sbtr.html indeed has a link to bj-sbtrt.html, but that page does not exist so I have a 404 error.
More generally, is there a rule as to when the us2 and us pages are updated ?


benjub:
I went ahead and fixed the althtmldef and latexdef of "extended unary minus", to be consistent with the htmldef (and also the (alt)htmldef's of +e and .e), and to differentiate it from "unary minus".

giomasce:
My program can find the theorems it uses even if hypotheses are permuted or variables are renamed, so no problem with deleting `sbdel` directly. You can do it if you want, otherwise I will do it next time I touch again that section.

nmegill:
> @nmegill I noticed something strange: it looks like the us2 server is faster to update html pages than to "serve" newly created ones. For instance, in the previous pull request, I changed the comment of bj-sbtr to link to bj-sbtrt and I added bj-sbtrt, but right now, the page bj-sbtr.html indeed has a link to bj-sbtrt.html, but that page does not exist so I have a 404 error.

The site is built in a temporary directory.  I have modified the build script to go though a pass that first copies over any files that don't exist yet to the live directory, before copying over any changed files.  I will test it over the next day or two.  This should solve the problem you saw.

> More generally, is there a rule as to when the us2 and us pages are updated ?

When the build is complete, each file in the temporary directory is compared (with diff) to the one in the live directory (if it exists), and only the new ones and ones that changed are copied to the live directory (so as not to trigger rsync or lftp unnecessarily).

It might have been better if I had changed the file dates inside the temporary directory, using the live directory as a reference.  Then when done I would simply rename the temporary directory to become the live directory, so all changes would  be simultaneous.  I'll think about doing it this way.  My thinking a long time ago was probably that the current way lets people see changes a little sooner, but it's not worth it if it causes problems or confusion.

PR #901, opened 2019-06-07
Subset

benjub:
Change gif, unicode and latex symbols for strict subsets, to remove ambiguity, as per https://groups.google.com/d/msg/metamath/_OqalbIKkgE/MNNefH20BAAJ
(I had let this on hold, but after recently running into the symbol again, I think it's just too confusing)

@nmegill : after agreeing in https://groups.google.com/d/msg/metamath/_OqalbIKkgE/p-iW2wIGBQAJ
you wrote:

> Therefore, Benoit, you may issue a pull request with the htmldef and althtmldef updated.  I don't think we should change the ASCII token "C.".  For the htmldef, use the symbol subsetneq.gif
> http://us.metamath.org/symbols/subsetneq.gif  (12 x 19)
> and I will modify install.sh so it will get copied to the mpegif directory during site build.  (If you prefer another gif, let me know.)
> 
> Leave the old htmldef and althtmldef commented out, and add your initials to a dated comment above the new one.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -54579,11 +54579,14 @@ htmldef "C_" as
   althtmldef "C_" as ' &#8838; '; /* &subE; */
     /* 2-Jan-2016 reverted sans-serif */
   latexdef "C_" as "\subseteq";
+    /* 7-Jun-2019 changed gif, unicode and latex def of "C." from subset to
+    subsetneq (BJ) */
 htmldef "C." as
-    " <IMG SRC='subset.gif' WIDTH=12 HEIGHT=19 ALT=' C.' TITLE='C.'> ";
-  althtmldef "C." as ' &sub; ';
+    " <IMG SRC='subsetneq.gif' WIDTH=12 HEIGHT=19 ALT=' C.' TITLE='C.'> ";
+    /* subset.gif */
+  althtmldef "C." as ' &#x228a; '; /* &sub; */
     /* 2-Jan-2016 reverted sans-serif */
-  latexdef "C." as "\subset";
+  latexdef "C." as "\subsetneq"; /* \subset */
 htmldef "(/)" as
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
Thanks for creating the pull request!  I'm absolutely fine with this change, since it's clear that different people have different conventions.  We also have the example of Wikipedia, who chose to do the same thing so that they'd have an unambiguous notation.

nmegill:
@benjub For consistency, could you make the same change to nf.mm and iset.mm, before I merge the PR?  Thanks.

benjub:
I noticed two other discrepancies in nf.mm (but not in iset.mm) compared to set.mm:
* the "sans-serif reversion" of 2016-01-02 was not done in nf.mm
* the  tag misses and ALT attribute and has an ALIGN=TOP attribute
Should they be fixed ?

nmegill:
> I noticed two other discrepancies in nf.mm (but not in iset.mm) compared to set.mm:
> 
> * the "sans-serif reversion" of 2016-01-02 was not done in nf.mm
> * the IMG tag misses an ALT attribute and has an ALIGN=TOP attribute
>   Should they be fixed ?

Yes, it would be nice to fix these.  It is lower priority and can be done at your convenience if you wish to do it (in a separate PR).  nf.mm has achieved its primary goal of proving the main important features of NF (derive infinity and show that Cantor's theorem and AC don't hold in general) and isn't being actively developed right now.

BTW I find Scott Fenton's development of nf.mm very impressive.  AFAIK this is the only computer formalization of NF set theory in existence.  Scott was unable to derive what he calls the "axiom of singleton" from Halperin's NF formalization, and whether Halperin's in incomplete as a result is an open problem.  If it is incomplete, that would certainly merit publication I think.

PR #902, opened 2019-06-07
mark eumo0 as OLD

wlammen:
eumo0 is an auxiliary theorem with little significance.  It became redundant after mo2v was introduced.  Mark it as OLD.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16330,6 +16330,7 @@ New usage of "mndoisexid" is discouraged (2 uses).
 New usage of "mndoismgm" is discouraged (3 uses).
 New usage of "mndoissmgrp" is discouraged (1 uses).
 New usage of "mndomgmid" is discouraged (3 uses).
+New usage of "mo2OLD" is discouraged (0 uses).
 New usage of "mo3OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
@benjub Anticipating that you may be concerned about this, I will go along with WL's deletion, particularly since eumo0 can be proved with a trivial chaining of eumo and mo2.  I'm not sure why it wasn't shortened that way before.  Maybe WL is losing his magic touch? :)

@wlammen Could you add the following line to the top of set.mm?  Then I'll merge the PR.
```
  6-Jun-19 eumo0     ---         deleted; use eumo + mo2
```

nmegill:
@wlammen Also, mo2OLD seems to be missing "Obsolete as of..." and discouragement tags.

wlammen:
>  I'm not sure why it wasn't shortened that way before. Maybe WL is losing his magic touch?

Rest assured I noticed it. But it wasn't worth the effort, because it was bounded for deletion anyway.  That's why the Proof modification.. tag was missing in the commits so far.  Added it now.
> Could you add the following line to the top of set.mm?
done
> Also, mo2OLD seems to be missing...
done

PR #903, opened 2019-06-08
added two new union of intervals proofs

TjWhale:
Added two new proofs about real intervals ioounsn and iocunico. The proof ioounsn  is the partner of http://us.metamath.org/mpegif/snunioo.html


benjub:
I suggest:
the other end --> the upper end

tirix:
Looks good! 
I think the comment of ~ iocunico may have to be updated, the ` Q ` does not apply anymore.

PR #904, opened 2019-06-08
More geometry

tirix:
- Added definition of 2D geometry, lower and upper axioms,
- Completed chapter 6 and started chapter 7 of [Schwabhauser]
- Moved 4 theorems from my mathbox into the main part of set.mm

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -264,6 +264,7 @@
 "4syl" is used by "baerlem5blem2".
 "4syl" is used by "bcm1k".
 "4syl" is used by "binomfallfaclem2".
+"4syl" is used by "bj-aevlem1v".
 "4syl" is used by "bnj1098".
 "4syl" is used by "canthp1lem2".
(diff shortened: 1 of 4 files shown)


benjub:
While I'm here... could you move your snunioc to the main part, right after snunico ?

tirix:
@benjub added snunioc !

tirix:
Note that I believe I shall create extra definitions for "midpoint" and "point mirroring/point reflexion/point inversion", as I assume they may be widely used in the future.

I've originally used "point mirroring" because this is the direct translation of the reference text (Punktspiegelung), however I see it the Wikipedia page prefers "point inversion".

I'll probably propose "midG" for midpoint and "invG" for point inversion.

benjub:
"central symmetry" sounds less old-fashioned

tirix:
It looks like "central symmetry" would rather be the property of a set of points to be invariant under such a transformation...

PR #909, opened 2019-06-09
provide nfeqf2 ... completed

wlammen:
The hypothesis builder form of theorem dveeq2 is actually not the best choice. Mostly a modified version having a F/ expression as a consequent is required.  So the clients repeatedly derive the modified version from dveeq2 themselves.

This commit provides the modified version named as nfeqf2.  Its proof is in fact contained in that of 
dveeq2 and need just be extracted.  The name resembles that of nfeqf, that can be seen as a generalization.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1374,8 +1374,8 @@
 "axc9lem1" is used by "nfeqf".
 "axc9lem2" is used by "ax6e".
 "axc9lem2" is used by "axc9lem3".
-"axc9lem2" is used by "dveeq2".
 "axc9lem2" is used by "nfeqf2".
+"axc9lem3" is used by "nfeqf2".
 "axicn" is used by "sineq0ALT".
 "axinfnd" is used by "axinfprim".
 "axinfnd" is used by "zfcndinf".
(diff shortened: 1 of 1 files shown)


wlammen:
copsexg has a very tiny correction added, so its not just minimized, and an OLD version is kept.

wlammen:
In order to facilitate the usage of nfeqf2 in lemmas of axpownd, I shortened two of them notably.  There is more headroom for shortenings, but I am not going to replay the proofs again for now.

wlammen:
The only theorem needing dveeq2 in its original form is ax12v.  So this proof becomes one line/7 bytes longer than the old one.  For now keep it in my mathbox as wl-ax12v.

dveeq2 now has just one application left, where it is actually better than nfeq2.  If you removed dveeq2, the proof of ax12v would become 7 bytes longer. That is more than compensated for by the drop of proof bytes in dveeq2.

Very similar considerations hold for dveeq1.

Keep them or drop them? My vote goes for dropping

benjub:
> Keep them or drop them? My vote goes for dropping.

@wlammen @nmegill As you can expect, I vote for keeping dveeq1 and dveeq2, or at least tell me before you remove them, so that I can duplicate them in my mathbox.

wlammen:
@benjub Thanks for your answer.

I did not expect you to keep dveeq*. Is there a good reason for doing so? Both competetors have just a few applications, so merging them into the most versatile form (which is nfeqf*) makes sense for me to keep the database free of just small variations of theorems.

benjub:
I plan to change the definition of df-nf soon, as was discussed recently on the discussion group, and I'd feel safer to be able to fall back on dveeq* if something fishy happens with nfeqf*.

Also, I think keeping even small variations does not clutter the database.  In other words, it's pretty costless to keep them.

PR #919, opened 2019-06-10
Hull

benjub:
Add to mathbox:
* finite summation in commutative monoids,
* field of real numbers, class of real vector spaces.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17842,6 +17842,8 @@ Proof modification of "binom2aiOLD" is discouraged (171 steps).
 Proof modification of "bitr3" is discouraged (14 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).
 Proof modification of "bj-19.21t" is discouraged (39 steps).
+Proof modification of "bj-abelcmnd" is discouraged (10 steps).
+Proof modification of "bj-abelcmnda" is discouraged (5 steps).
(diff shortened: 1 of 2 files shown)


tirix:
@benjub I believe we already have ` CMnd ` for the commutative monoids.

I did not read your finite sum in detail. Why couln't ` gsum ` be used for that?

benjub:
@tirix I did use CMnd (I had added ComMnd in an intermediate commit because I had missed CMmnd, but then I saw it and used it in subsequent commits).
As for df-gsum, I missed it (I google-searched "site/us.metamath.org monoid" and things like that but did not find it).  I'll have to have a look. It should be equivalent (the goal is indeed the same) but df-gsum looks harder to understand...

tirix:
Great! 
I think ` gsum ` is a bit more general, as it handles both commutative and non-commutative monoids.

Also, maybe you could also consider using ` CVec ` for your ` RRVec `. There are already many theorems which can be reused.

(PS. Yes, I also saw you backed up on the ` CMnd ` , and edited my comment!)

nmegill:
> As for df-gsum, I missed it (I google-searched "site/us.metamath.org monoid" and things like that

I assume you meant "site:us.metamath.org monoid" which gives much better results.

Also, in metamath.exe you can type "search * monoid /comment" which gives about 250 results, the 2nd of which is df-gsum.

PR #924, opened 2019-06-14
Replacing ( CCfld |`s ZZ ) by ZZring (3)

avekens:
See also #866:
- Theorems/definions still using ( CCfld |`s ZZ ) duplicated (with suffix OLD) and marked as obsolete.
- all remaining occurences of ( CCfld |`s ZZ ) replaced by ZZring (except in obsolete theorems/definitions)
- many usages of obsolete theorems/definitions in proofs replaced by corresponding theorems/definitions using ZZring (remaining usages of "zlmvalOLD", "zrhmulgOLD" and "zrhrhmOLD" will be replaced soon)
- some new basic theorems for ZZring added
- some minor improvements of comments

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4676,6 +4676,7 @@
 "df-wrecs" is used by "wfrlem9".
 "df-wrecs" is used by "wrecseq123".
 "df-zlmOLD" is used by "zlmvalOLD".
+"df-znOLD" is used by "znvalOLD".
 "df-zrhOLD" is used by "zrhvalOLD".
 "df0op2" is used by "hh0oi".
 "df0op2" is used by "ho01i".
--- a/discouraged
+++ b/discouraged
@@ -430,9 +430,11 @@
 "4syl" is used by "vieta1lem2".
 "4syl" is used by "wfrlem5".
 "4syl" is used by "znf1o".
+"4syl" is used by "znf1oOLD".
 "4syl" is used by "znfld".
 "4syl" is used by "znhash".
 "4syl" is used by "znzrh2".
+"4syl" is used by "znzrh2OLD".
 "4syl" is used by "zrhunitpreima".
 "5oalem1" is used by "5oalem6".
 "5oalem2" is used by "5oalem3".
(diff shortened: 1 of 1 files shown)


avekens:
Unfortunately, the Travis check failed because of:

The command "scripts/verify-mmj2 --setparser 'mmj.verify.LRParser' --extra "$(printf 'RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel,df-sbc
RunMacro,showDiscouraged,discouraged.new')" set.mm" exited with 1.

Running mmj2 locally I do not have such problems. Can this be caused by Mario's latest changes?

avekens:
The obsolete definitions df-zrhOLD, df-zlmOLD and df-znOLD seem to cause the problem, e.g.:
I-PA-0202 Axiom df-zrhOLD has failed the definitional soundness check. The previous axiom df-zn uses the symbol being defined in this axiom.

But I need them for the other OLD theorems based on the old definition (recursively!). So what can I do? The only way I see is to delete all OLD theorems based on these definitions (in contrast to the procedure described by Norm in #866). Or I have to introduce new symbols ZRHomOLD,  ZModOLD and Z/nZOLD and use them to replace the current symbols in all OLD definitions/theorems...

nmegill:
Unfortunately you can't have two definitions of the same symbol.  To work around this, you have to use a different name for the symbol like ZModOLD for df-zlmOLD, which is annoying and also requires a $t update.  (Another workaround is to rename df-zlmOLD as ax-zlmOLD, but that is ugly.)

In the case of a modified definition, I think it is often best just to delete the old definition and the *OLD theorems depending on it, rather than keeping them around as "obsolete" for a year as I described earlier; sometimes it isn't worth the effort.  Worst case, GitHub still has the old definition if someone needs to look it up.

To work around Travis, the easiest thing to do here is rename df-zl.mOLD to ax-zlmOLD etc. temporarily for Travis to pass,, then soon (within a few days) deleting ax-zlmOLD and everything depending on it.  Or, just delete the old definitions and their dependencies now if you want.

avekens:
@nmegill  Thanks for the hint: I'll take the quick&dirty solution for the moment (renamed df-**OLD by ax-**OLD) to get thinks done already into GIT (to avoid later merge conflicts). It is, however, a lot to do to resolve all remaining dependencies, so this will take several hours. I promise that I will provide a clean solution (removing ax-**OLD and all dependent theorems) soon (hopefully until Monday latest).

david-a-wheeler:
@avekins - I think that's the right choice, it's just not worth trying to keep the old versions around.

benjub:
@avekens In the comments, you can also write ` ZZring `-module instead of ` ZZ `-module.  And thanks for this work!

PR #926, opened 2019-06-14
Extensionality

benjub:
Mainly typos in the main part and cleanup of my mathbox.

You can merge now, because I'm afraid of merge conflicts, but for the next PR, may I:
* remove ax12a2 and ax12v2 and "inline" them in the proof of axc15
* replace the occurrences of "~~? " with "~ "
* remove the TODO note and the discouraged tag from exp3acom23g (if it keeps the discouraged tag, it will necessarily appear as a seldom used theorem!)
* remove pm2.21fal, since this is an instance of pm2.21


nmegill:
> * remove ax12a2 and ax12v2 and "inline" them in the proof of axc15

My plan was to make sure we can derive ax-12 from
  ( x = y -> ( ph -> A. x ( x = y -> ph ) ) )  with $d x y
without using ax-12 and vice-versa, to show that in principle it can serve as an alternative to ax-12.  The above theorem is the version of ax-12 used in the literature (mentioned and proved several times in the Tarski/Kalish/Monk trilogy of papers.  (I did that with the old ax-c15 but never got around to doing it with the newer ax-12.)   If you want to attempt that, feel free to do so.  I kept these theorems in case their proofs contained hints towards that goal.

> * replace the occurrences of "~~? " with "~ "

OK.  (Make sure the ~~? theorem now exists of course.)

> * remove the TODO note and the discouraged tag from exp3acom23g (if it keeps the discouraged tag, it will necessarily appear as a seldom used theorem!)

It is useless.  Just to be sure, I 'minimized' all of set.mm with it against the whole set.mm (several years ago).  My plan was to give AS the opportunity to justify it but of course that it not possible now.  (I discussed with him and purged a number of theorems of this nature, or moved them to his mathbox, but this one apparently got overlooked.)  In particular, its last step of imbi2i makes it even more specialized and less useful, which is why we don't have 1000 theorems applying imbi2i to every existing equivalence.

We have to draw the line somewhere about adding trivial combinations of existing theorems for no reason, otherwise we could potentially have millions of silly useless theorems.

There is a very slight possibility that his virtual deduction tool has a dependency on this theorem.  So rather than deleting it, you can move it to AS's mathbox and remove the TODO note.

> * remove pm2.21fal, since this is an instance of pm2.21

OK.  (It is used by dihglblem6, so modify that proof to eliminate it.  'expand pm2.21fal' in MM-PA should do that.)

nmegill:
> My plan was to make sure we can derive ax-12 from
> ( x = y -> ( ph -> A. x ( x = y -> ph ) ) ) with $d x y    (*)
> without using ax-12 and vice-versa

Thinking about this some more, the elimination of the old ax-c11 seems to make essential use of the bundling in ax-12, so I don't think (*) can replace ax-12 unless we add ax-c11 back as an axiom.  The old work in theorems ax12*,axc15* primarily relates (*) and axc15, and my vague plan was to eventually clean these up to use ax-12 more directly instead of via axc15.  So maybe leave ax12a2 and ax12v2 in for now, and I'll add a to-do item to my personal list to look at this more carefully.

nmegill:
> so I don't think (_) can replace ax-12 unless we add ax-c11 back as an axiom. The old work in theorems ax12_,axc15*

with markdown mangling fixed:

so I don't think (\*) can replace ax-12 unless we add ax-c11 back as an axiom. The old work in theorems ax12\*,axc15\*

PR #929, opened 2019-06-15
Extensionality

benjub:
* move exp3acom23g to AS's mathbox
* many tiny typos related to tildes, spaces, wrapping...

@nmegill merge now since the tiny changes are scattered through all of set.mm and I fear merge conflicts.  I'll do the rest of #926 tomorrow but I have to go now.


nmegill:
@benjub I noticed you have changed the formatting of some theorems.  I'll mention a convention that I didn't follow early on but now try to when I remember.  The convention is that when a line is broken at a connective, I put the connective at the start of the next line instead of at the end of current line.  E.g.
```
1 + 1
 = 2
```
rather than
```
1 + 1 =
  2
```
Someone actually emailed me about this a few years ago, saying that in the literature, the practice was for usually for connectives to start a new line instead of ending the previous line, especially in a long chain of equalities or implications.  I think he was right and agree with him that it generally looks better, so I have started doing that, and over time have reformatted some older theorems but nowhere near all of them.

benjub:
@nmegill Oops!  I'm sorry, I didn't know the convention about the line breaking, and I did break several statements *after* binary operators.  But I'm not sure any convention should be enforced on this topic... many special cases should be dealt with and I'm not sure it's worth it.
I used to break lines before binary operators, but know I break them after so that it is obvious to the human eye that this is not the end of the formula and another line will follow.
Other conventions I generally and loosely use:
* if the formula can be put in one line, then do not break it, and more generally, use as few lines as possible (but use common sense to make exceptions);
* if the formula has line breaks, then "flush right" the last line, i.e. the final "$=" or "$." should be the 78th-79th characters.

benjub:
@nmegill Oops!  I'm sorry, I didn't know the convention about the line breaking, and I did break several statements *after* binary operators.  But I'm not sure any convention should be enforced on this topic... many special cases should be dealt with and I'm not sure it's worth it.
I used to break lines before binary operators, but know I break them after so that it is obvious to the human eye that this is not the end of the formula and another line will follow.
Other conventions I generally and loosely use:
* if the formula can be put in one line, then do not break it, and more generally, use as few lines as possible (but use common sense to make exceptions);
* if the formula has line breaks, then "flush right" the last line, i.e. the final "$=" or "$." should be the 78th-79th characters.

nmegill:
> @nmegill Oops! I'm sorry, I didn't know the convention about the line breaking, and I did break several statements _after_ binary operators. But I'm not sure any convention should be enforced on this topic... many special cases should be dealt with and I'm not sure it's worth it.

It is mainly a stylistic preference that we don't enforce.

> I used to break lines before binary operators, but know I break them after so that it is obvious to the human eye that this is not the end of the formula and another line will follow.

I am aware of this argument, and it's the reason I originally did it in older theorems.  But in the case of set.mm, the absence of "$=" indicates there is more to follow.

I think the primary motivation for a break before connective is the following.  In a book it would be unusual to see
```
1+2+3+4= 
3+3+4=
6+4=
10
```
rather than
```
1+2+3+4  
  =3+3+4 
  =6+4 
  =10
```
because in the latter case, the ='s line up and it's easy to see that we have a chain of equalities.  Of course in set.mm, chains of equalities are not usually part of a statement, only a proof, so strictly speaking that argument doesn't apply too well.

I wish I could quote the email that convinced me to switch to line break before connective, but I was unable to find it.  I'll keep looking.


> Other conventions I generally and loosely use:
> 
> * if the formula can be put in one line, then do not break it, and more generally, use as few lines as possible (but use common sense to make exceptions);

Agree.

> * if the formula has line breaks, then "flush right" the last line, i.e. the final "$=" or "$." should be the 78th-79th characters.

Maybe sometimes, when it helps clarity.  It is done 'show proof' screen displays in metamath.exe, in particular because it has arbitrary (often bad) line breaks anyway, and pushing the leftover piece to the end of the line can help to visually continue the line,

I haven't been doing it routinely in set.mm because it wastes file space.  (In principle we could use tabs to save space, but it was a mess when I allowed tabs a long time ago since everyone seemed to have their own tab stop preferences.)  But when manual formatting is done right, it isn't too common to have a small left-over piece of the equation.

As you said, we should let common sense be the guide.  I don't think we need to enforce any of this, although it might be nice to have some recommended style guidelines.

Issue #52, opened 2016-07-04
Merge developments of R^n


digama0:
On Thursday, April 7, 2016 at 8:34:08 AM UTC-4, fl [wrote](https://groups.google.com/d/msg/metamath/ri4VTOCkXeg/z6sAQ-U_BwAJ):
> I think that ( RR ^N ) and ( CC ^ N )  in my mathbox or what is related to euclidean spaces in Madsen's mathbox deserve to be made official.

The version I would prefer to see (personal taste only) is R^n and C^n as Hilbert spaces, meaning the structure product of n copies of RR or CC with an inner product and induced norm. This will allow for reusing structure theorems for groups, vector spaces etc. (which FL's mathbox version doesn't do) and will also allow dealing with the inner product structure (which supercedes JM's definition of the norm).

If you like Madsen's mathbox version, an easy way to get it into main is just to use it in your mathbox. Since mathboxes are independent, any common work will be moved up.

That said, I'd really like to have some theorem that depends essentially on the structure being R^n / C^n to justify the definition. So far it's all statements that apply to all vector spaces or all Hilbert spaces. Madsen's proof that boundedness is equivalent to total boundedness in R^n is a good example, but if it's just the one theorem it is not really worth it.

As you say, we currently have three versions of R^n in the database. The only way this will be moved up is if someone takes it upon themselves to unify all three developments, meaning that a new development (or extension of one of these three) must encompass all the major results of the other two developments.

The Tarskian geometry approach is interesting, but I think your "planar incidence geometry" approach is the better one. Essentially, define the geometry axioms in general and form a new structure for them (and prove what results you can from this structure), then show that R^n has a compatible structure, so that R^n acquires the geometry theorems.

sorear:
Side note: `df-plig` has a broken link, should be http://public.csusm.edu/aitken_html/m410/betweenness.08.pdf

digama0:
@sorear Fixed in https://github.com/metamath/set.mm/commit/04f3414a460ea4e8e06ba0167f0cf690a52ebc09.

tirix:
I think it's worth mentioning in this issue this definition proposed in [a post by Mario](https://groups.google.com/d/msg/metamath/fAw0tIDacrw/KuYXrLGCBQAJ):
` df-rrx $a |- RR^ = ( i e. _V |-> ( toCHil ' ( ( ringLMod ' ( CCfld |'s RR ) ) freeLMod i ) ) ) $. `
[...]

PR #931, opened 2019-06-15
Added inner product to the "image structure" imas and "product struct…

tirix:
This pull request changes the ~ df-imas and ~ df-prds definitions to also construct the inner product fields.

The idea is to be able to use this to advance towards #52 : the development of RR^n.
I had to move gsum and 0g all the way up before ~ df-prds is defined. Is that an issue? Any hint about how to avoid that this kind of definitions land far away from the chapter they belong to?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1443,7 +1443,6 @@
 "bafval" is used by "sspba".
 "basendx" is used by "2strstr".
 "basendx" is used by "catstr".
-"basendx" is used by "cchhllem".
 "basendx" is used by "eengstr".
 "basendx" is used by "grpbasex".
 "basendx" is used by "grpplusgx".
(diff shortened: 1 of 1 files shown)


tirix:
I realise now my definition of the inner product for the structure image does not make any sense. 
I may update it by requiring < f(p) | f(q) > = < p | q >, i.e. that f is an isometry.

tirix:
This pull request is now fixed and can be merged from my point of view, but review is welcome!

benjub:
@tirix Skimming through your changes, I noticed that the term "scalar product" is sometimes used for "scalar multiplication".  There are many occurrences of "scalar product" in set.mm, sometimes to mean "inner product" (it is a legit synonym), and sometimes, ambiguously, to mean "scalar multiplication".  Maybe we should retain from using "scalar product", and use only "inner product" and "scalar multiplication".  Do @nmegill and @digama0 agree ?

benjub:
@tirix As for the image structure: inner products, being forms, are things that you "pull back", not that you "push forward", so you can define the inverse image but not the direct image.  I don't know what metamath requires for inner products, but for instance, the pullback of a definite form is definite if and only if the function f is injective.  In particular, if f is a linear bijection, then you can also define the direct image (pullback by the inverse of f).

tirix:
@benjub I see, thanks! Actually, the reason why I also changed the "image structure" is that the theorem ~ imasvalstr is used both both for the image structure and the product structure, because both structure have the same arguments.
I thought it would be best to keep the parallel. If it does not make sense, I can roll back the changes for the image product (df-imas). The theorem ~ imasval will have to be duplicated for the product structure.

PR #932, opened 2019-06-16
rm an 18 month old outdated proof

wlammen:
I think this fossil (hba1OLD) was an oversight.  I don't see any point in keeping this any longer.


nmegill:
The reason I overlooked this was that I usually search for "mmm-yyyy. " (note the period and the space after the period) when I remove the expired ones.

PR #937, opened 2019-06-16
Added a definition of Euclidean spaces

tirix:


benjub:
A few comments:
* Eulidean --> Euclidean (several times)
* The symbol RR^ makes one think of a direct product while your definition defines a direct sum.  I think both objects are important and both should be introduced, with matching symbols.  Proposals for a symbol for the direct sum are welcome...
* I would rename the section "The Archimedean ordered field of real numbers" to "The ordered field of real numbers", since "Archimedean" is a property and not a (piece of) structure.  I would also introduce the symbol RRfld (see ~ df-bj-rrfld ).  It's a good thing that you moved this section to the main part: I might need it soon for real vector spaces.

tirix:
Thank you Benoît, I fixed the easy typos. 
For the more ambitious changes (like ` RRfld `), they can be done in a different PR.
For direct sum/direct product, feel free to suggest a better symbol, I'm opened to suggestions but can't find any myself either!

One question about the indexes for vectors: I chose ` ( 1 ... n ) ` to be later able to integrate with Scott's ` EE `  ( ~ df-ee ), and because it's the usual convention in mathematics (and physics) to start indexing vectors coordinates with 1. 
For proving ` RR^m ⊕ RR^n = RR^(m+n) `, it might be interesting to use the ` Word ` concept, which has indices starting at zero ` ( 0 ..^ n ) `, but I still don't believe it a reason enough to choose this convention.

nmegill:
I'm not sufficiently well-versed in this to give a useful opinion, but it seems ok and I will merge it.

One little thing:  you are using 'I e. _V' in the antecedent of rrxval and others.  We generally prefer to use the more general 'I e. V' (or another variable if V is not available) to make it more general; that way we can often avoid needing extra elex and syl in case I is already a member of something.  You may want to change that in a future PR.

Actually, looking thru set.mm with 'search * '|- ( ? e. V_ -> ' it seems there are about 20 other theorems with this same issue.  Someone may want to volunteer to clean these up.

tirix:
@nmegill right, thanks! I've fixed this in #939.

Your search might be too conservative, though. Even just ` |- ( ( ? e. V_ -> ` (with double parens) will account for theorems with more than one condition and give more/different results.

benjub:
@tirix Another remark: The term "generalized Euclidean space" is not standard.  Did you find it somewhere?  I don't like this terminology much since I would expect a Euclidean space, even a generalized one, to be complete, and this is not the case for an infinite countable direct sum of copies of R. Also, I think the phrase "direct sum of copies of the field of real numbers" should appear somewhere; maybe in the comment of df-rrx:

> Define the function associating with a set the free real vector space on that set, equipped with the natural inner product.  This is the direct sum of copies of the field of real numbers indexed by that set.  We call it here a "generalized Euclidean space", but note that it need not be complete (for instance if the given set is infinite countable).

PR #938, opened 2019-06-16
Extensionality

benjub:
(see commits summaries)
@nmegill :
* Since I would like to keep cbv1hOLD cbv1OLD cbv3hOLD cbv2OLD until I make the
changes to df-nf agreed upon in the discussion group, can I change the date after "obsolete as of" from 13-May 2018 to, say, 13-Dec-2018 ?
* I'm not forgetting the few things left at #926 and #929 .
* aecom and axc11n are the same statement; discourage usage of one of them?
* it looks like parts of the comments of renegcl and renegcli really belong to renegclALT

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17860,6 +17860,7 @@ Proof modification of "bj-cbvexdv" is discouraged (65 steps).
 Proof modification of "bj-cbvexdvav" is discouraged (22 steps).
 Proof modification of "bj-cbvexv" is discouraged (46 steps).
 Proof modification of "bj-cbvexvv" is discouraged (12 steps).
+Proof modification of "bj-ceqsalt" is discouraged (126 steps).
 Proof modification of "bj-chvarv" is discouraged (18 steps).
 Proof modification of "bj-chvarvv" is discouraged (10 steps).
 Proof modification of "bj-cleqh" is discouraged (108 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
> (see commits summaries)
> @nmegill :
> 
> * Since I would like to keep cbv1hOLD cbv1OLD cbv3hOLD cbv2OLD until I make the
>   changes to df-nf agreed upon in the discussion group, can I change the date after "obsolete as of" from 13-May 2018 to, say, 13-Dec-2018 ?

Sure, make them whatever you think you need, even into the future.

> * I'm not forgetting the few things left at #926 and #929 .
> * aecom and axc11n are the same statement; discourage usage of one of them?

I think aecom should come first, followed by axc11n.  Of course the proofs would also be swapped since one depends on the other.  The purpose of axc11n is simply to provide a match to ax-c11n for 'verify markup', and that is the one that should be marked 'discouraged'.  But we also want to keep its alias aecom because it is a more meaningful name.

> * it looks like parts of the comments of renegcl and renegcli really belong to renegclALT

renegclALT is just an experiment in my mathbox.  Its usage is discouraged anyway.  However, the 'deprecated' comment in renegcli may be a little severe; I'll think about it.  We don't want to encourage it where it's not appropriate, but dedth will always exist (it is currently used 88 times) and there are occasional valid uses of it that make overall proofs shorter.  See https://groups.google.com/d/msg/metamath/Nv5ZSEsKueg/Opf41fChAgAJ
I may make a google group post about dedths and elimhyps in a couple of days - these are not meant to be used but they (along with a to-be-written description) may provide a better understanding of the weak deduction theorem, which some people have found confusing.

PR #941, opened 2019-06-17
change obsolete dates of some theorems to keep them a few more months…

benjub:
Changes:
* I removed drnf2OLD since it has been obsolete for more than a year and its proof has no advantage whatsoever (axioms, definitions, syntax used) over drnf2.
* Concerning axc11n/aecom: I would make aecom a biconditional (see bj-aecom) and move the part of the comment of aecom citing [Megill] to the comment of axc11n.
* Minor changes.

Questions, remarks for @nmegill :
* Can I move axc11nlem1 earlier, since it uses only axioms up to ax-7 ? Maybe you can propose a better label, without "lem" in it ?
* On second thought, I will not delete pm2.21fal since it is mentioned in the comment of ~ natded.
* I leave ax12a2 and ax12v2 as they are, and let you add them on your todo list.
* For the moment, I will not touch theorems in "1.2.13  Auxiliary theorems for Alan Sare's virtual deduction tool, part 1", but I agree that eventually, the theorems therein will go either to previous sections of propcalc, or to AS's mathbox.  By the way, is there a "part 2" ?
* Looking at your section "21.30.1  Experiments with weak deduction theorem", it looks like elimhyps (resp. dedths) could be proved from elimhyps2 (resp. dedths2) more simply? And maybe it is worth proving the closed forms as well (as well as the (partially) closed forms of elimhyp and dedth)?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14906,7 +14906,6 @@ New usage of "dochspocN" is discouraged (0 uses).
 New usage of "dral1-o" is discouraged (4 uses).
 New usage of "dral1ALT" is discouraged (0 uses).
 New usage of "dral2-o" is discouraged (4 uses).
-New usage of "drnf2OLD" is discouraged (0 uses).
 New usage of "drngoi" is discouraged (2 uses).
 New usage of "dtruALT" is discouraged (0 uses).
 New usage of "dtruALT2" is discouraged (0 uses).
@@ -17766,6 +17765,7 @@ Proof modification of "ax2" is discouraged (46 steps).
 Proof modification of "ax3" is discouraged (24 steps).
 Proof modification of "ax4" is discouraged (45 steps).
 Proof modification of "ax52" is discouraged (14 steps).
+Proof modification of "ax5ALT" is discouraged (3 steps).
 Proof modification of "ax5el" is discouraged (30 steps).
 Proof modification of "ax5eq" is discouraged (30 steps).
 Proof modification of "ax6e2eq" is discouraged (111 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
> * Concerning axc11n/aecom: I would make aecom a biconditional (see bj-aecom)...

It depends on what the benefit would be.  In that section there may be a lot of implications that are chained in proofs, and making aecom biconditional might break up a 3syl for example.  But if we use axc11n to keep the old proofs, the name is cryptic.

 
> * Can I move axc11nlem1 earlier, since it uses only axioms up to ax-7 ? 

Yes.

>Maybe you can propose a better label, without "lem" in it ?

Maybe aevw, "weak aev", where aev means "forAll Equality dV".  But if you like something else that's fine too.

> * On second thought, I will not delete pm2.21fal since it is mentioned in the comment of ~ natded.
> * I leave ax12a2 and ax12v2 as they are, and let you add them on your todo list.

They are on my list.

> * For the moment, I will not touch theorems in "1.2.13  Auxiliary theorems for Alan Sare's virtual deduction tool, part 1", but I agree that eventually, the theorems therein will go either to previous sections of propcalc, or to AS's mathbox.

You can move them to AS's mathbox unless they are used, in which case they can go to the proper place.  You can discard the section heading.

> By the way, is there a "part 2" ?

It used to be in early set theory but no longer exists.  It has already been moved to his mathbox or to the proper place.

> * Looking at your section "21.30.1  Experiments with weak deduction theorem", it looks like elimhyps (resp. dedths) could be proved from elimhyps2 (resp. dedths2) more simply?

Yes, I know that, I'm just playing around with some ideas.

> And maybe it is worth proving the closed forms as well (as well as the (partially) closed forms of elimhyp and dedth)?

I haven't thought about that, could it have an interesting use?  I think the weak deduction theorem is intriguing, and I wonder if there is potential for the technique to be more efficient and less cryptic.

PR #945, opened 2019-06-18
Extensionality

benjub:
* move axc11nlem1 earlier and relabel it cbvequv (for "Change Bound Variable in an EQUality with dV conditions); relabel axc11nlem2 to axc11nlem
* mathbox: minor updates

From #941 :
* regarding closed forms of elimhyp and dedth and your generalizations: no, I haven't thought of particular uses.
* axc11n/aecom: making aecom a biconditional (see bj-aecom) would render it consistent with ax11 versus alcom; I checked that it would not make later proofs longer (when axc11n is used, it is chained with syl, which would become something like sylib, and similar cases where the inference using a biconditional exists in set.mm); the only exception is axi10, which would have to reference axc11n or use biimpi.

Other questions:
* There is a "TODO" in the comment of spfw asking if it should be removed.  I think keeping it makes things clearer, so I would remove the TODO if you agree.
* The proofs of 19.8w, speimfw, spimfw, ax12i actually prove more general theorem schemes (see bj-19.8w, bj-speimfw, bj-spimfw, bj-ax12i).  The first one could be renamed 19.2d and the other three could be moved up one section: they have actually nothing to do with equality.  Can I restore the general statements and do this ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17826,7 +17826,7 @@ Proof modification of "bj-ala1BIS" is discouraged (9 steps).
 Proof modification of "bj-ax16gb" is discouraged (18 steps).
 Proof modification of "bj-ax16nf" is discouraged (31 steps).
 Proof modification of "bj-axc10v" is discouraged (37 steps).
-Proof modification of "bj-axc11nlem2v" is discouraged (58 steps).
+Proof modification of "bj-axc11nlemv" is discouraged (58 steps).
 Proof modification of "bj-axc11nv" is discouraged (71 steps).
 Proof modification of "bj-axc11v" is discouraged (14 steps).
 Proof modification of "bj-axc15v" is discouraged (93 steps).


nmegill:
> Other questions:
> 
> * There is a "TODO" in the comment of spfw asking if it should be removed.  I think keeping it makes things clearer, so I would remove the TODO if you agree.

I agree.

> * The proofs of 19.8w, speimfw, spimfw, ax12i actually prove more general theorem schemes (see bj-19.8w, bj-speimfw, bj-spimfw, bj-ax12i).  The first one could be renamed 19.2d and the other three could be moved up one section: they have actually nothing to do with equality.  Can I restore the general statements and do this ?

This is a little problematic.  Often we do prefer the most general possible statement of a theorem, but not always.  The point of the ones you mentioned is not to show the most general theorem possible from their proofs but to illustrate a property described in the comment.  In the case of bj-ax12i, for example, it visually bears no obvious relationship to ax-12, whereas in ax12i it is obvious.  The substitution instance used is part of communicating the theorem to the reader, and by showing only the most general instance we would be discarding information we had before.  I might go along with this (or perhaps keep both versions, yours with a "g" suffix) in case the more general version was useful for some other purposes, but I don't see that in this case.

There are probably many cases like this in set.mm.  In principle we could write a program that identifies proofs that result in a more general theorem, but I don't think that there wouldn't be many where it would be beneficial to discard the information provided by displaying the special case.  (TBH I'm just guessing; it might be an interesting project someday.)

A very minor point is that using bj-ax12i in a proof in place of ax12i will result in a slightly longer proof, since the substitution instances are less direct.  (This is not a reason in itself to reject the more general version, though.)

Some years ago, 19.2g arose as test case that ocat liked while was writing mmj2, and I somewhat reluctantly added it (with a "g" suffix for "more general") to set.mm.  I thought it would never get used, and so far I've been right.  Even if you explicitly use 19.2g in a proof, 'minimize' will always replace it with 19.2 when the x and y are the same variable (which so far has always been the case).

wlammen:
@benjub bj-hbaev is not independent of ax-13. In fact the proof uses it quite directly in step 2.

benjub:
Thanks @wlammen. I'm going to correct that.  To prove it without ax-13, I would need to completely unbundle it, but then it would be a trivial consequence of ax-5 (and similarly with theorems referencing it).  You certainly noticed that this work of mine is very basic (probably automatizable), and is an experiment to see if removing dependencies on ax-13 from later theorems would lengthen the proofs much.  And you: are you still trying to remove some dependencies? To you have a plan or program?

wlammen:
@benjub I do not have any plans whatsoever, because I am working full time in a job unrelated to the topic of Metamath. This leaves just some of my spare time to irregularly take a look at set.mm. It is out of my scope to follow an agenda, or to delve deeper into the background of something, at least on a larger scale.
hbae seems to be very fundamental in that it combines most, if not all, axioms of predicate logic, into a a single property of A. x x = y. Untangling the ingredients looks promising to me, that's why I was curious about your result.

PR #946, opened 2019-06-19
Extensionality

benjub:
* mathbox: fixed some theorems thanks to @wlammen's comment in #945 
* removed the TODO of spfw as agreed in #945 

@nmegill : may I go on with the change to making aecom a biconditional, given that:
> axc11n/aecom: making aecom a biconditional (see bj-aecom) would render it consistent with ax11 versus alcom; I checked that it would not make later proofs longer (when axc11n is used, it is chained with syl, which would become something like sylib, and similar cases where the inference using a biconditional exists in set.mm); the only exception is axi10, which would have to reference axc11n or use biimpi.

* Note to myself: I still have to do this from #945 (see exchanges there):
> The proofs of 19.8w, speimfw, spimfw, ax12i actually prove more general theorem schemes (see bj-19.8w, bj-speimfw, bj-spimfw, bj-ax12i). The first one could be renamed 19.2d and the other three could be moved up one section: they have actually nothing to do with equality.


nmegill:
> @nmegill : may I go on with the change to making aecom a biconditional, given that:

OK, go ahead.

Issue #949, opened 2019-06-20
Conventions: Update table of abbreviations, theorem forms


avekens:
I just committed an enhancement of the comment for ~ conventions (table of abbreviations used in labels) and an update of the section comment for "Logical implication", see PR #948 (@nmegill you were too fast merging my PR into metamath:develop - I originally wanted you and @david-a-wheeler  to review my changes first...).

Here some remarks on and questions about these changes:

- **Conventions/Table of abbreviations used in labels**: Although the number of entries in the table is more than doubled now, I think it is far from being complete. I wonder what should actually be contained in this list (what does "complete" mean?): all exceptions to the general conventions (of course), prefixes/suffixes not being names of definitions (df-NAME), e.g. ALT, com, simp, and the names (NAME part of df-NAME) of (some?/all?) definitions. In question is the last point: which definitions should be included? Only "important"/"often used"/"familiar" ones (this might be very subjective)? All (this will be a lot!)? Or are there certain rules? For example, I did not include an entry for "oprab", although there is a definition df-oprab, because this name is composed of "op" and "rab" for which entries already exist. Similar with "gra" (for graph): there is no definition df-gra for a "graph", but definitions df-usgra, df-umgra etc. for special kinds of graphs, containing this abbreviation. Nevertheless, I added most entries relying on definitions on personal taste. @david-a-wheeler please feel free to remove some of the entries I added if they are not appropriate for the table!
- **Section comment for "Logical implication"**: I replaced the paragraph about "forms of theorems" by a (slightly modified) text from the (new version of the) Metamath book. I wonder, however, if it should be placed there, because there is a similar paragraph ("Theorem forms") in the "conventions". Maybe the paragraph in the comment for "Logical implication" can be removed completely, and the paragraph "Theorem forms" in the "conventions" can be revised accordingly.

@nmegill @david-a-wheeler what is your opinion?

nmegill:
> I just committed an enhancement of the comment for ~ conventions (table of abbreviations used in labels) and an update of the section comment for "Logical implication", see PR #948 (@nmegill you were too fast merging my PR into metamath:develop - I originally wanted you and @david-a-wheeler to review my changes first...).

I did read "logical implication", which is fine, but I only checked a few in the conventions table.  I find it hard to render HTML source code in my head, :) so actually regenerating it on us2 is helpful to review it.

> 
> Here some remarks on and questions about these changes:
> 
[...]

PR #951, opened 2019-06-21
 Revision of "theorem forms" and suffixes

avekens:
Revision of "Header" comments:
- Item "4. Metamath syntax summary" revised (esp. clarification of terms "token" and "symbol")
- Item "5. Other notes" enhanced by subitems 2. and 3. , esp. link to ~ conventions

Revision of other comments:
- conventions about theorem forms removed completely in section header of "Logical implication" (hint in header item "5. Other notes"  shpould be sufficient)
- conventions about theorem forms removed completely in theorem comment of  ~ a1d (hint in header item "5. Other notes"  shpould be sufficient)
- Item "Theorem forms" in comment of ~ conventions completely  revised (mainly according to the new Metamath book)
- Item "Suffixes" in comment of ~ conventions partially revised

See also issue #949.


avekens:
@nmegill @david-a-wheeler could you review my changes first, please? Do you know how I can make you official reviewers of a PR in GitHub? There is the title "reviewers" at the right of the comments, but I do not know how to add somebody there (according to the Help of GitHub, there should be something to be clicked, but there isn't anything for me...).

david-a-wheeler:
Nit:

>We prefer to write theorems in "deduction form" and avoid theorems in "inference form".  We call this approach "deduction style".

The antecedent is misleading. The phrase "this approach" seems to refer to inference form, and of course that is not what you meant.

david-a-wheeler:
Overall I like this change however:

> In contrast, we avoid theorems in "inference form".

Seems a little strong. We try to avoid using inference form when that leads us to use the deduction theorem. Other than that, there certainly isn't a Prohibition against using inference form, and there are clearly many theorems in that form. Indeed, I think the week deduction theorem is kind of interesting, so I would hate to see its loss entirely.

So I would tweak that to say

> In contrast, we usually avoid theorems in "inference form" when that would end up would require us to use the deduction theorem.

avekens:
I changed the texts a little bit, according to David's remarks. The "deduction style" should have been related to both: preferring (deduction form) AND avoiding (inference form). But if this is misleading, I agree with changing it (as I have done already).
Concering "avoiding the inference form": "to avoid" does not mean "to forbid", so it would still be allowed to use the inference form. Since in most cases, however, we do not know _in advance_ that we have to use the deduction theorem (as soon as we want/have to use the theorem in deduction form), the deduction form should be proven instead of the corresponding inference form. I made the experience several times (and Jon made the same experience recently, see Google group https://groups.google.com/d/msg/metamath/DDN4noc0TYU/GTq6zioxBwAJ ) that I needed an existing theorem to proof a new theorem but this existing theorem is only available in inference form, so that I could not use it immediately. Therefore, the inference form should only be used if there are good reasons for it (e.g. week deduction theorem).

nmegill:
I am fine with these changes.

david-a-wheeler:
@avekens - thanks!!

Issue #947, opened 2019-06-19
Unicode symbols for CHil and friends


benjub:
* May I change unicode symbols (i.e. althtmldef) to make the "C" blackboard bold in CMod, CVec, CPreHil, toCHil, CHil, CPreHilOLD, CBan, CHilOLD, and others that I may have forgotten?

* May I change the tokens
pi to _pi
om to _om
tau to _tau
iota to _iota
iota _ to _ iota _  (without the spaces: it is to avoid markdown)
in order to avoid confusion with wff metavariables (and in the case of pi and om, to make the non-underscored versions usable as wff metavariables), and others that I may have forgotten?

nmegill:
> * May I change unicode symbols (i.e. althtmldef) to make the "C" blackboard bold in CMod, CVec, CPreHil, toCHil, CHil, CPreHilOLD, CBan, CHilOLD, and others that I may have forgotten?

I'm not sure if I like that - I think it would look weird and might even be confusing (e.g. the CC and Mod in CMod might look like two juxtaposed symbols that someone forgot to put space between).  If anything, making the CC a subscript might be better.  Maybe someone else has an opinion.

However, the whole section containing these is deprecated and will be deleted someday.  So it will be wasted effort.  This was all done  before the extensible structure builder and some definitions ended up awkward and inflexible.  We are keeping primarily to use as a possible reference for the structured and possibly more general versions.  Maybe that time has even passed or will soon.

> * May I change the tokens
>   pi to _pi
>   om to _om
>   tau to _tau
>   iota to _iota
>   iota _ to _ iota _  (without the spaces: it is to avoid markdown)
>   in order to avoid confusion with wff metavariables (and in the case of pi and om, to make the non-underscored versions usable as wff metavariables), and others that I may have forgotten?

OK except iota, see below.  Also, we have been using non-italic for Roman constants e.g. _e and _i, and probably should use non-italic (non-slanted) for Greek constants.  I have seen non-italic letters for constants used more often in recent literature, and I think I've even seen a non-italic pi used for 3.14.

I don't think we should add them as wff variables until there's an actual need.  Perhaps that is what you meant.

[...]

PR #952, opened 2019-06-21
Labels

benjub:
Second item of #947 : 
* tokens pi , om , tau ----> _pi , _om , _tau
* unicode of _om put upright; unicode of _tau was already upright; unicode of _pi unchanged since the upright looked ugly (although maybe I did not have the correct font installed)


nmegill:
For math token changes, typically we have been documenting it with a single line rather than listing the affected theorems.  E.g.
```
 8-Dec-17 ---       ---         changed math symbol from "cprod_" to "prod_"
```
You may want to simplify the "label changes" list this way in a future PR.

PR #955, opened 2019-06-23
Extended complex numbers

benjub:
* document token changes pi, om, tau --> _pi, _om, _tau
* move en3lp where it belongs
* for some theorems: when the theorem has "New usage is discouraged.", add to the comment: "Use ... instead." (this should be more systematic).
* mathbox: add extended complex numbers, addition, opposite

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -49,6 +49,7 @@
 "0lt1sr" is used by "supsrlem".
 "0ncn" is used by "axaddf".
 "0ncn" is used by "axmulf".
+"0ncn" is used by "bj-inftyexpidisj".
 "0ngrp" is used by "vsfval".
 "0ngrp" is used by "zrdivrng".
 "0nnq" is used by "addassnq".
@@ -1750,6 +1751,7 @@
 "bnj1213" is used by "bnj1452".
 "bnj1213" is used by "bnj1523".
 "bnj1219" is used by "bnj1379".
+"bnj1224" is used by "bj-inftyexpidisj".
 "bnj1224" is used by "bnj1204".
 "bnj1224" is used by "bnj1279".
 "bnj1228" is used by "bnj1204".
@@ -13382,7 +13384,7 @@ New usage of "0lnfn" is discouraged (9 uses).
 New usage of "0lno" is discouraged (3 uses).
 New usage of "0lnop" is discouraged (6 uses).
 New usage of "0lt1sr" is discouraged (2 uses).
-New usage of "0ncn" is discouraged (2 uses).
+New usage of "0ncn" is discouraged (3 uses).
 New usage of "0ngrp" is discouraged (2 uses).
 New usage of "0nnq" is discouraged (22 uses).
 New usage of "0npi" is discouraged (9 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -542114,6 +542146,189 @@ sethood hypotheses (compare ~ opth ).  (Contributed by BJ, 6-Oct-2018.) $)
...
+
+  $( Syntax for the opposite of extended complex numbers. $)
+  coppcc $a class -cc $.
+
+  $( Define the opposite of an extended complex number (or more precisely, the
+     operation giving the opposite, sometimes called negation).  (Contributed
+     by BJ, 22-Jun-2019.) $)
+  df-bj-oppc $a |- -cc = ( x e. CCbar |->
+      if ( x e. CC , -u x , ( inftyexpi `
+                          ( if ( 0 < x , ( x - _pi ) , ( x + _pi ) ) ) ) ) ) $.
```
tirix: I assume you mean ` ( 1st ' x ) ` here instead of ` x `?

benjub: Bien vu !  commit coming...

tirix:
Interesting! I'm curious about two things:

- Do you also have plans to change the definition of +oo and -oo (df-pnf, df-mnf) so that we have ` RR* C_ CCbar ` ?
- Do you have plans to do geometry in this set?

benjub:
Thanks Thierry!  Yes, the plan is to change the definitions of +oo and -oo to have a genuine inclusion of \overline{\R} into \overline{\C}, as written in the head comment of the section "Extended complex numbers".  I have to define multiplication in the obvious way (I think I will define complex arguments to make it clearer), inversion, and complex conjugation.  I'm not sure if it's possible to define this early enough in set.mm so as to "overload" the operations, to have unified operations in \overline{\C}.
No plans beyond that: as I wrote, it's not a very important space, but it may help unify some operations. Much more important are the real/complex projective lines and the real projective plane, that can be obtained as quotients of \overline{\C}.

set.mm
```diff
@@ -542114,6 +542158,237 @@ sethood hypotheses (compare ~ opth ).  (Contributed by BJ, 6-Oct-2018.) $)
...
+  df-bj-ccbar $a |- CCbar = ( CC u. CCinfty ) $.
+
+  $c +cc -cc prcpal Arg .cc invc $.
+
+  $( Syntax for the function ` prcpal ` . $)
+  cprcpal $a class prcpal $.
+
+  $( Define the function ` prcpal ` .  (Contributed by BJ, 22-Jun-2019.) $)
+  df-bj-prcpal $a |- prcpal = ( x e. RR |-> ( ( x mod ( 2 x. _pi ) ) -
+                 if ( ( x mod ( 2 x. _pi ) ) <_ _pi , 0 , ( 2 x. _pi ) ) ) ) $.
```
tirix: With this definition, you will probably have to handle many cases. Maybe it would be more straightforward if you define it as ` Im(log(exp(iθ))) ` (where `θ` is your `x`). That way, you could benefit of all the properties and theorems already available for ` Im(log x) `.

benjub:
@tirix Thanks.  I contemplated doing that, but I prefer not to use such an
advanced function (logarithm) to define this simple function, which is
basically a "modulo".  But indeed, the link with the argument is clear and
shall be formalized.  My next tasks are something like

$p |- ( A e. ( - _pi (,] _pi ) -> ( prcpal ` A ) = A $= ? $.

$p |- ( A e. ( - _pi (,] _pi ) -> ( prcpal ` ( A + _pi ) ) = if ( A <_ 0 ,
( A + _pi ) , ( A - _pi ) ) $= ? $.  (and use the LHS instead of the RHS in
the definition of the opposite)

$p |- ( ( A e. RR+ /\ B e. RR ) -> ( Arg ` ( A x. ( exp ` ( _i x. B ) ) ) )
= ( prcpal ` B ) ) $= ? $.

But for now, it is ready for @nmegill to merge.

benjub:
And of course, I'll have to correct the typo in the domain of the
multiplication (df-bj-mulc) due to a copy-paste error from a previous
version.

PR #956, opened 2019-06-23
 A real Euclidean space is a complex pre-Hilbert space.

tirix:


nmegill:
It seems there was a conflict with Benoit's PR.  My guess is the label change list at the top.

nmegill:
While it's not a requirement, typically we omit the "d" suffix meaning "deduction version" (as I assume is the case for suppofss1d) for more advanced theorems when we don't expect to have an inference or closed theorem version.

I have no other comments; nice work.

PR #958, opened 2019-06-25
optimize aevlem1

wlammen:
The idea behind this optimization is to recycle setvar rather than introduce new ones.  This keeps both the list of referenced theorems and the distinct variable constraints shorter.  The proof logic does not change.  Therefore I did not bother adding a Revised... tag, or even create an ...OLD version.


benjub:
I think there was a thread on the google group about the practice of using different dummy variables versus recycling them, with @digama0 saying he had changed his mind about this over time.  You (Wolf) gave the arguments in favor of recycling them, but an argument against is that it makes proofs harder to follow.
I'm not saying which should be chosen, or if it should depend on proof length or "difficulty"; I'm just saying that there apparently is a debate. Maybe @nmegill has some thoughts about it.

wlammen:
I do not find the relevant groups post.

Here, a variable is not really recycled in different contexts, but kept longer in use. This narrows down a more general intermediate result, that in turn is a specialization of the end result. That is how the whole proof is organized: successively dropping constraint after constraint until finally reaching the most versatile form, an expression allowing for bundling.

There is perhaps some merit in demonstrating how to escape narrowing conditions at a maximum pace. In the particular instance of aevlem1 the 'high speed' can be throttled without an effect other than reducing the amount of resources needed.

nmegill:
We have no rule on recycling dummy variables, but I'm not opposed to it generally, and I think it depends on context.  I've done both recycling and not recycling depending on whether I feel that recycling would interfere with understanding the proof.  Sometimes recycling is clearly desirable; for example, using A. x phi(x), A. y phi(y), and A. z phi(z) in different parts of the proof might be more confusing than just using A. x phi(x) everywhere, in addition to making both the compressed proof and the web page proof significantly longer since we have to reprove the same thing with different variables.

There is a another issue which I will call "loss of information".  It would be relatively straightforward to write a program to eliminate all variable recycling in set.mm; it can be done manually in metamath.exe with 'delete floating', 'ini all', 'unify all/int', then just assign a different letter to each work variable.  OTOH it can be hard, and may involve some human judgment, to go the other way.  So that is an argument in favor of erring on the side of recycling when in doubt.

PR #959, opened 2019-06-26
Axext

benjub:
Commit messages say it all. Ready to merge.


nmegill:
Good find on df-cleq.  Please go ahead and replace the existing definition at your convenience.  BTW we can't have a *OLD for a $a.  You could put the *OLD in commented out form if you wish.

PR #961, opened 2019-06-26
generalize wl-equsb4

wlammen:


benjub:
@wlammen In the comment of wl-equsb4, you wrote "The distinctor antecedent is more general than a distinct variable constraint".  It looks like this requires dtru, which itself depends on many axioms, or am I mistaken?  Isn't it worth keeping both versions of wl-equsb4 ?

nmegill:
@benjub The distinctor antecedent replaces the need for $d x z.  To obtain the $d x z version, we would normally make use of axc16g to derive the A. x x = z -> ... version, then apply pm2.61i.  I don't know if that uses more logic axioms, but unlike dtru it would not use any set theory axioms.  As for "more general", I guess it depends on one's viewpoint.  But I think the distinctor version is often harder to prove than the $d version, so the distinctor version is "more general" in that it provides both the distinctor version (itself) immediately and the $d version relatively easily.

wlammen:
@benjub It is actually more specific than a distinct variable condition, because it automatically rules out the (boring) case of a one-element-universe.

What I had in my mind when I wrote the comment was still something different: A distinctor expression is directly accessible to logical manipulation. On application it is more or less easily dropped via pm2.61, ax16g, as Norm wrote, quite standard, I routinely check for this. On the other hand, try to drop an unnecessary distinct variable condition... They persist in an obnoxious way. The infamous ax-11 unbundling problem is an infamous example: Were the distinct variable condition available as a distinctor, I could derive ax-11 in 5 minutes from it. By contrast, it took me (and others) weeks of effort to first clear hundreds of theorems of ax-11, then fruitlessly struggle to turn the dv condition into a distinctor. Sigh. 

It is true that handling a distinctor depends on axioms, in particular ax-13 is usually involved. Sad. I still prefer distinctors over dv conditions because of the reasons I mentioned. As much as my time allows, I will put some basic distinctor applications like wl-equsb3 in my mathbox, in the faint hope, that either I or sb else will manage to prove ax-11 from ax-11v, via distinctor possibly,or the reason why this must fail, becomes apparent.

wlammen:
@nmegill I don't know whether this unbundling thing is still open. I like to develop a common technique for this along the following lines:
If we have an expression ph(u, y) which we can prove under the assumption d(u, y) and -. A. u u = y, then by aev and dv(u, x) we have -. A.x x = y -> -. A. u u = y, and in total
-. A.x x = y -> A. u u = y -> ph(u, y)
Applying [x / u] to this implication chain yields
-. A. x x = y -> -. A. x x = y -> ph(x, y)
This leaves us with proving the special case
A. x x=y -> ph(x, y),
for many ph a very easy task.
And of course [x/u] A. x x= y <-> A. x x= y along with [x/u] A. u u = y <-> A. x x = y, for which I prepared wl-equsb3 / wl-equsb4 already.

My problem is my limited time. If you are impatient enough (unless you know this leads to nothing), you might want to have a closer look into this idea.

PR #965, opened 2019-06-27
mathbox: extended complex numbers, df-clel tweak

benjub:
@nmegill I will do as you advised in #959 and actually propose the same modification for both df-cleq and df-clel (see the present PR).
One thing I wanted to explore is to split df-cleq in two halves (the respective two implications of the biconditional): one is extensionality and the other is the substitution property for the membership predicate.  With the latter, it would already be possible to prove that equality of classes is an equivalence relation, and many other theorems, without extensionality.  But then, these two halves of df-cleq would not look like a definition anymore ( df-xxx $a |- ( definiendum <-> definiens ) $. )  So I'm not sure you would like it.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1525,8 +1525,11 @@
 "bitr3" is used by "sbc3orgVD".
 "bitr3" is used by "trsbcVD".
 "bj-csbsnlem" is used by "bj-csbsn".
+"bj-df-clel" is used by "bj-dfclel".
 "bj-df-cleq" is used by "bj-dfcleq".
 "bj-inftyexpidisj" is used by "bj-ccinftydisj".
+"bj-inftyexpidisj" is used by "bj-minftynrr".
+"bj-inftyexpidisj" is used by "bj-pinftynrr".
 "blo3i" is used by "ipblnfi".
 "blocn" is used by "blocn2".
 "blocn2" is used by "ubthlem1".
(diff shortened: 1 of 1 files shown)


nmegill:
> @nmegill I will do as you advised in #959 and actually propose the same modification for both df-cleq and df-clel (see the present PR).

While I agree that your df-cleq version may be nicer in some respects, the modification comes inexpensively, requiring no more symbols while eliminating an implicit dependence on an FOL axiom.  It eliminates the need for a cleqjust theorem (which we should probably have had for the existing definition).  And it doesn't require any changes to our writeups like the class theory section of mmset.html.

However, I see this as a minor cosmetic improvement.  I think the the existing df-cleq version is fine in principle.  Arguably the existing version is even better, since it makes the ax-ext dependence explicit by reproducing the actual ax-ext.  The point of the hypothesis is to prevent the definition from being creative from the point of view of set theory i.e. not implying ax-ext.  At that point we have FOL as a given, and the existing df-cleq does not add anything to FOL.  If we want to go to extremes, all FOL definitions imply an instance of biid and would add something to a prop calc fragment containing say only ax-3.

The existing df-clel doesn't add anything to FOL as cleljust shows.  Even though it may add something to weak fragments of FOL (i.e. cleljust needs the FOL axioms that prove it), that isn't a concern once we are in set theory.  For example, it is rare that we go out of our way to avoid an FOL axiom in a set theory proof.  Adding a hypothesis sends the wrong message and gives the impression that something deeper is at stake.  If you want to explain in the df-clel description that the definition presupposes those FOL axioms needed to prove cleljust, go ahead.

But at this point I don't want to replace df-clel.

> One thing I wanted to explore is to split df-cleq in two halves (the respective two implications of the biconditional): one is extensionality and the other is the substitution property for the membership predicate. With the latter, it would already be possible to prove that equality of classes is an equivalence relation, and many other theorems, without extensionality. But then, these two halves of df-cleq would not look like a definition anymore ( df-xxx $a |- ( definiendum <-> definiens ) $. ) So I'm not sure you would like it.

While I see your point, I don't think it is a good idea to do that because it would require changing everything we've written about df-cleq, it does not match the definition in textbooks, and it probably confuses things for the reader not interested in such subtleties.

If you want to do a side study of what can be derived from one direction of df-cleq (stated as a theorem and used as a starting point for the study), that is fine of course.

PR #967, opened 2019-06-28
trying to prove ax-11 from ax-11v, discontinued

wlammen:
Now A. y ph <-> A. u A. y [ u / x ] ph needs a proof.  This is harder, one has to make a distinction A. x x= y and -. A. x x = y.  The latter is the complex case, not as easy as the committed case here


wlammen:
I did some pen and paper work yesterday, and thought I found some way to prove ax-11 from ax-11v. Now, after work, I looked into this again, and came to the same opinion as 6 months ago: as I now remember, ax-12 and ax-gen do not allow this, accept in some degenerate cases like identical variables, or unconditional true expressions. I discontinue this attempt, sorry for any misdirection.

PR #970, opened 2019-06-30
( CCfld |`s RR ) replaced by RRfld

tirix:
Also added the geometrical "less than" relationship


tirix:
In most cases, changes were minor, so I have not added the "Revised by" at the end of changed theorems. Shall I?

nmegill:
> In most cases, changes were minor, so I have not added the "Revised by" at the end of changed theorems. Shall I?

If only the proof changes, normally we omit "Revised by"; only when the proof is changed manually specifically to shorten it (not with 'minimize' or merely as a result of a changed definition) do we mention "Proof was shortened by...".

If the theorem changes, it is less clear.  If many theorems are slightly modified to accommodate a new definition, it may be optional.  For example, when I switched to the new definition of df-sbc, I didn't add "Revised by" (except in the definition itself) but added to the label change list,

```
 7-Jan-17 df-sbc    [--same--]  df-sbc and 521 uses modified (statement and/or
                                  proof) to use separate notation from df-sb.
```

A primary purpose of the label change list and "Revised by..." is to inform people who may be working in those areas or using those theorems.  It partly depends on the audience and how old the original theorem is.  If it was recently added and you're pretty sure no one else is working in the area (or if they are and you know they are aware of what is going on) then it's probably less important.

Overall, a common-sense judgment call should be made.  In this case, I think it is OK not to put explicit "Revised by..." if you don't care about being acknowledged for the revision.  However, it would be good to put at least a generic comment about the change like the 7-Jan-17 one above.

tirix:
Got it, thanks!

I think others (Benoît and Alexander) are already largely aware about the change.

Most of the changes are to my own, so the acknowledgment is already there. I some cases I did change the theorem statement, I might add the “Revised by” for those.

PR #973, opened 2019-06-30
ax-11v study

wlammen:
For me this settles now the question whether ax-11 can be derived from a version with distinct variable conditions.  Those interested in this topic might want to have a look at wl-ax11-lem10 where I put some notes into the comment.


benjub:
Nice development, @wlammen!  Maybe you could actually do what you say in the comment, i.e. prove
${ $d x u $.  wl-ax11-lem11 $p |-  ( A. y y = u -> ( A. y A. x ph -> A. x A. y ph ) ) $= ? $. $}

wlammen:
> Nice development, @wlammen! Maybe you could actually do what you say in the comment, i.e. prove
> ${ $d x u $. wl-ax11-lem11 $p |- ( A. y y = u -> ( A. y A. x ph -> A. x A. y ph ) ) $= ? $. $}

This wont help. The distinct variable condition on u and x would reduce the result at once to ax-11v,

benjub:
I noticed that. But since you say it in the comment, you may as well do it.  Anyway, I think your development, even if it does not prove ax-11 from ax11v (I suspect it is not possible), is interesting in showing more precisely what is lacking.

wlammen:
Need to check what happens, if I use [ y / u ] on wl-ax11-lem10

nmegill:
@wlammen I was looking with interest at some of your ax-11v results.  As I think I mentioned some time ago, I was also trying to see if ax-11v sufficed, but I gave up.  Because it reproduced most of pred calc and caused confusing duplicated search matches, I deleted this work from my mathbox in commit c342609.

I was able to achieve ax-11 from ax-11v indirectly via substitution:
```
    $d u v x $.  $d u v y $.  $d x ph $.  $d y ph $.
    ax11w12AUX11 $p |- ( A. x A. y [ x / u ] [ y / v ] ph -> A. y A. x [ x / u ] [ y / v ] ph ) 
```
and the other direction can be obtained using ~ sbcom2NEW11 if I'm not mistaken.  Using this, I think ax-11 from ax-11v can be achieved for wffs not involving wff variables or quantifiers because we can eliminate the substitutions e.g.
```
    $d v u $.  $d y u $.
    elsb34AUX11 $p |- ( [ x / u ] [ y / v ] u e. v <-> x e. y )
```
or if one of the variables is not x or y, e.g.
```
    $d u v z $.
    ax11w17lem1AUX11 $p |- ( [ x / u ] [ y / v ] u = z <-> x = z )
```
and we should be able to build up wffs using ~ sbn and ~ sbim.  Being able to derive these weak instances of ax-11 is still far from ax-11 itself, but it might be useful for some simple theorems without wff variables or quantifiers.

I didn't get far with moving quantifiers to the outside of substitution (sbal*)  and in particular wasn't able to derive the elusive ~ hbae.  Of the sbal* theorems requiring ax-11, I was able to prove from ax-11v
```
    $d x y $.
    sbal1NEW11 $p |- ( -. A. x x = z -> ( [ z / y ] A. x ph <-> A. x [ z / y ] ph ) ) 
-------------------------------------------------------------------------------
    $d x y $.  $d x z $.
    sbalNEW11 $p |- ( [ z / y ] A. x ph <-> A. x [ z / y ] ph ) 
-------------------------------------------------------------------------------
    $d x z $.  $d y z $.
    sbalvNEW.1 $e |- ( [ y / x ] ph <-> ps ) 
    sbalvNEW11 $p |- ( [ y / x ] A. z ph <-> A. z ps ) 
```
but was unable to prove from ax-11v
```
    $d x z $.
    sbal2 $p |- ( -. A. x x = y -> ( [ z / y ] A. x ph <-> A. x [ z / y ] ph )  )
```

wlammen:
@nmegill It is impressive how close you come to the final goal, and anyone (well, with some knowledge) interpreting the formulas "feels" or "wants" that goal be true.  In particular the theorem sbal1NEW11 was either not familiar to me, or I had forgotten about it over the time.  Nevertheless, I think substitution does not really lead to more progress.  If we, for a moment, put aside the boring case of a one-object-universe, then all non-trivial substitution boils down to A.x (x = y -> ph).  And there you are, mixing quantifiers and substitution, or recurred substitution in one formula, is just another sequence of quantifiers in disguise, whose exchange imposes exactly the same distinct variable mess we want to eliminate.  At least that is my impression after so many futile efforts to get around this.  I cannot provide a thorough proof, or at least suggestive explanation, why this impression holds, though.

I posted the study to direct attention to a different problem, that is perhaps more at the core of the problem.  At least I think so.  In a situation with two separate implications A.x x=y -> ph and -. A.x x=y -> ph, with the goal to show a bundled theorem, you are left in the -. A.x x=y branch with a particular problem: The antecedent implies a dv condition, but you must not state it explicitly.  The idea behind the study is to alias y to a variable u different from x, with an appropriate dv condition.  The introduction is possible, because (A.u u=y /\ -. A.u u=x) -> -. A.x x=y is provable. Unfortunately the reverse direction not, although a variable u exists for sure (namely y), but any such substitution is rejected by Metamath due to dv condition violations.  You need the reverse direction to get rid of u again.

PR #974, opened 2019-07-01
add fnoa to iset.mm

jkingdon:
The proof ends up being longer than the set.mm one because we
have additional steps to show that we are evaluating functions
within their domains. But because the characteristic function
`( z e. _V |-> suc z )` is defined on all sets, it is relatively
straightforward (by contrast, `fnom` may require a version of
`rdgifnon` which only requires a characteristic function defined
on ordinals, or some other approach).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -54229,6 +54229,18 @@ $)
       ZHUBJJCUDTNDOPRS $.
   $}
 
+  ${
+    $d x y z $.
+    $( Functionality and domain of ordinal addition.  (Contributed by NM,
+       26-Aug-1995.)  (Revised by Mario Carneiro, 8-Sep-2013.) $)
+    fnoa $p |- +o Fn ( On X. On ) $=
+      ( vy vz vx cvv csuc cmpt crdg cfv wcel con0 wral coa cxp wfn cdm wceq vex
+      cv sucex ax-mp eqid fnmpti rdgifnon fndm eleq2i wfun rdgfun funfvex rgenw
+      mpan sylbir rgen df-oadd fnmpt2 ) ARZBDBRZEZFZCRZGZHZDIZAJKZCJKLJJMNVCCJV
+      BAJUOJIUOUTOZIZVBVDJUOUTJNVDJPUSURCQBDUQURUPBQSURUAUBUCJUTUDTUEUTUFVEVBUS
+      URUGUOUTUHUJUKULUICAJJVALDCABUMUNT $.
+  $}
+
 $(
 #*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
        Appendix:  Typesetting definitions for the tokens in this file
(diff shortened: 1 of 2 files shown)


digama0:
Hm, I'd hoped we could put this "prove it is in domain" stuff behind us. I realize that there is a missing theorem: We need to know that `recs ( F )` (and `rec` by consequence) is set-like when F is. The main theorems tfri* are already assuming that F is a set-like function, but to show that `recs ( F )` is also set-like some of the intermediate lemmas (specifically tfrlemi1) must be generalized to the case when `z` is not an ordinal.

I think the following argument works: Suppose we are given a set `x`, and want to show that ``( recs ( F ) ` x)`` exists, which reduces to proving that there is some set `U` such that for all y, `x A y -> y e. U`. Using tfrlemi1 on the ordinal `( x i^i On )` we obtain an acceptable function `g` on `( x i^i On )`. Let `U := ran g`. Then for any `y` such that `x A y`, we deduce that `x h y` for some acceptable ordinal function `h`, and hence `x` is an ordinal; thus `x = ( x i^i On )` and `x g y` by uniqueness of acceptable functions, so `y e. ran g`.

jkingdon:
> Hm, I'd hoped we could put this "prove it is in domain" stuff behind us.

I will admit that evaluating functions within their domains strikes me as more elegant than coming up with ways to evaluate them outside, but I'm also pragmatic enough to be willing to pursue anything which might get the theorems we want proved. It seems like a variety of recursion hypotheses could coexist (including different ways to state "increasing and continuous" although I'm not sure I yet have anything to propose other than the hypothesis of http://us2.metamath.org/ileuni/rdgisuc2.html ).

Anyway, to get back to function values existing, I took a bit of a look at formalizing the "missing" theorem mentioned. I'm afraid I didn't get much further than stating it; the statement I had was:

```
  ${
    tfrex.1 $e |- F = recs ( G ) $.
    tfrex.2 $e |- ( Fun G /\ ( G ` x ) e. _V ) $.

    $( Existence of transfinite recursion evaluated at any set. $)
    tfrex $p |- ( F ` u ) e. _V $=
      ? $.
  $}
```

At the moment I'm probably most focused on seeing how far I can get with ordinal addition, but it won't be long before I or someone else needs to dig into a number of the more difficult topics, such as the function existence one here, the fact that http://us2.metamath.org/ileuni/rdgisuc2.html isn't yet as powerful as http://us2.metamath.org/mpeuni/rdgsuc.html , and maybe some others in the general area of recursion.

digama0:
To clarify re: eliminating domain checks, what I mean is that I want to recover as much as possible the ease of use of `fvex`, which says nothing about domain checks. Of course for proving that ``( F ` A ) e. B`` we will usually want some domain conditions, but for pure sethood there is more or less a metatheorem that says that any class expression you write down will be set-like, and hence all function values exist if the input does. (An example of a non set-like class function is ``A |-> |^| A``, but it's also not a set for certain input values.)

Re: your formulation, I would suggest changing it to ``( A e. V -> ( F ` A ) e. _V )`` for practical use. It can also be stated as `` `' F Se _V `` for use with `sefvex`, but that's a purely cosmetic matter and you will want the first version anyway.

PR #977, opened 2019-07-02
Added constants used in "structured" typesetting

tirix:


wlammen:
OT: I noticed that leg0 has a German comment: Nullstrecken sind minimal. A translation would be : A degenerated (zero-length) leg is minimal

nmegill:
How are these constants going to be used?  Can you show an example?

tirix:
@nmegill My implementation of the structured typesetting output is a modification of the Metamath program, with an additional output format.

It reads and parses format files (currently only set-mathml.mmts is maintained) using the Metamath program’s pre-existing functions for building strings of math symbols, therefore operates on math symbols.

I needed to extend the existing type constant symbols (wff, setvar, class) with contextual information about how to represent them. There are two such extensions in set-mathml.mmts:

- one for classes “on their own lines”, for which parens can be omitted like when below a square root radical or as a numerator or denominator of a fraction. These get a new type “class-o”.
- one for classes representing numerals, which can be concatenated so that e.g. ; 1 0 is represented as 10. These get “class-n”.

This commit declares these new type constants, so that they are read and understood by the modified Metamath program. You won’t see them referenced elsewhere in set.mm, but only in set-mathml.mmts.

Previously, I’ve been manually modifying set.mm each time I generated the structured typesetting pages. I would like to ease my job and get rid of that step by having the definition in the file.

This was a bit long, I hope I’ve been clear! 😊

tirix:
@wlammen Thanks! I fixed that in #979.
@nmegill Thanks for merging, I guess I was clear enough then!

PR #980, opened 2019-07-03
Add more set-like lemmas to iset.mm

digama0:
Based on https://github.com/metamath/set.mm/pull/974#issuecomment-507122575 . Proves that `recs` and `rec` are set-like, as well as adding some more lemmas for deducing that mpt and mpt2 are set-like. *Note:* It might seem peculiar that `mptfvex` does not have a domain condition, but this is necessary. It's not sufficient to prove that `A. x e. A B e. _V` to show that `( x e. A |-> B )` is set-like, because the existence `A. x ( x e. A -> E. y y = B )` is in the wrong place relative to the implication. What is needed is `A. x E. y ( x e. A -> y = B )`, and in most cases it's easier to just prove the stronger statement `A. x E. y y = B`.

This also adds a variant on df-oexp, df-oexpi, that just drops the if condition. I thought about putting a constraint to make it classically equal to df-oexp, but the constraint will not behave very well on general ordinal numbers. Once we get to `om` we will have discreteness which will make these sorts of adjustments more well behaved. So I would propose to not bother with df-oexp and insert the if statement when defining the finite version. (In fact, the ordinal exponential is not even needed for peano arithmetic, so it could just be skipped entirely until df-exp.)

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -45478,6 +45478,24 @@ $)
   $}
 
   ${
+    $d x y A $.  $d y B $.  $d x y C $.  $d x y D $.  $d y F $.
+    fvmptss2.1 $e |- ( x = D -> B = C ) $.
+    fvmptss2.2 $e |- F = ( x e. A |-> B ) $.
+    $( A mapping always evaluates to a subset of the substituted expression in
+       the mapping, even if this is a proper class, or we are out of the
+       domain.  (Contributed by Mario Carneiro, 13-Feb-2015.)
+       (Revised by Mario Carneiro, 3-Jul-2019.) $)
+    fvmptss2 $p |- ( F ` D ) C_ C $=
+      ( vy cv wbr wss wi cfv fvss cvv wcel nfcv wceq sylbi funmpt2 funrel ax-mp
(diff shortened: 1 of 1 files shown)


nmegill:
The Travis check failed:
```
Created at the top of file "iset.mm" is less recent than the date 3-Jul-2019 in
    the description of statement "fvmptss2".
```

jkingdon:
This is great work. Thanks for fleshing out those comments at #974 ; if I had tried to prove this stuff based on those comments alone I think I would have gone down several dead ends and generally not known what to do.

As for exponentiation, I'll confess I haven't really been playing close attention to exponentiation yet, but `df-oexpi` as defined here seems like the plausible way to go about it (with https://github.com/metamath/set.mm/issues/847#issuecomment-490882770 having a slight elaboration of this issue).

jkingdon:
> The Travis check failed:
> 
> ```
> Created at the top of file "iset.mm" is less recent than the date 3-Jul-2019 in
>     the description of statement "fvmptss2".
> ```

I have submitted #981 to make this date check work as in set.mm.

PR #981, opened 2019-07-04
Also specify --top_date_skip for iset.mm

### LINKED ISSUE #867: Modify iset.mm verification to skip top date
Modify the iset.mm verification to work like set.mm.

See https://github.com/metamath/set.mm/pull/865#issuecomment-495820216

jkingdon:
This means that pull requests don't have to update the date at
the top of the file. It seems like having one rule for set.mm
and a different one for iset.mm has been both unnecessary and
confusing.

This resolves #867

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ script:
   - metamath 'read set.mm' 'markup mmset.raw.html mmset.html /ALT /CSS' quit
   - metamath 'read iset.mm' 'markup mmil.raw.html mmil.html /ALT /CSS' quit
   - scripts/verify --top_date_skip --extra 'write bibliography mmbiblio.html' set.mm
-  - scripts/verify iset.mm
+  - scripts/verify --top_date_skip iset.mm
   # Verify and also generate 'discouraged.new', and do extra checking
   # (the 'printf' lets us insert multiple lines):
   - scripts/verify-mmj2 --setparser 'mmj.verify.LRParser' --extra "$(printf 'RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel,df-sbc\nRunMacro,showDiscouraged,discouraged.new')" set.mm


nmegill:
I have mixed feelings about this.

The top date is part of a complete 'verify markup' check.   The top_date_skip for set.mm is a compromise that I made somewhat reluctantly for a practical reason, which is that several times we have had 2 PRs from different time zones with different dates, leading to a merge conflict in the 2nd PR after the 1st PR was merged.  Because of this compromise, the top date is often stale, maintained only whenever someone thinks of it, and you can no longer depend on it to  indicate the version date of a set.mm copy.

There has never been a merge confllict with iset.mm caused by editing the top date, and it is unlikely to happen with only a couple of developers.  Unlike merge conflicts in set.mm,  which can take time to pin down, there is nothing confusing about the Travis date error:  Travis tells you exactly what to fix.  It just means that the PR submitter didn't run 'verify markup/f' or didn't check to see if the Travis check passed.

For PR #980, I usually would have merged the PR anyway and fixed it with a separate commit, which I sometimes to for such simple corrections.  Unfortunately I was rushed last night and didn't have time.  I just fixed it so we can discuss this without holding up iset.mm.

How do other people feel about dropping the top date check for iset.mm?

jkingdon:
My ideal solution would involve tracking versions via something other than a date/version stored in the file in git. For example, every time there is a push to the develop branch, a script could automatically add a version/date and publish the result to somewhere on metamath.org.

However, that's a bigger change.

I guess I don't really have much of an agenda here one way or another beyond wanting to make it easy for @digama0 to contribute to iset.mm. If the `--top_date_skip` cure is seen as being worse than the disease, so be it.

digama0:
I think we should drop the top date entirely. It's a high contention target, causing extra busywork and failed merges. Git makes it completely unnecessary to put the date in the file itself, and it can manage the metadata much better than we can. If you are worried about some software archaeology situation where someone has come across the .mm file without the metadata, it's not hard to get a fairly accurate date by looking at the most recent dates on a theorem.

nmegill:
The problem with depending on GitHub is that most of the infrequent contributors don't know how to use git and don't care to learn because it's not important to them.  Instead, I get updates to their mathboxes, patches to comment wording, etc. as email attachments.  While of course we want to encourage GitHub use, I am fine with handling such submissions manually because they are not that frequent.

Usually they download the current set.mm from the Metamath site and put it into a local directory and make their edits.  Sometimes their changed set.mm lags the GitHub version.  Having the date at the top is useful for these situations so that I can know instantly (within a day at least) what version  they were working from.

I try to keep the set.mm date updated for this reason.  Since it's the only .mm file where the date can be stale, I don't mind doing that.  The more diligent contributors usually do it for me anyway when their set.mm PR is the only one in the queue.

If I also have to keep iset.mm, nf.mm, etc. dates updated by hand, that is more work for me.  Since these have never had the merge conflict problem and probably never will, I don't think that having users to run 'verify markup */f' and fixing any issues reported is asking too much.  They should run the markup check anyway for all the many other things it does, so the report of a stale top date will automatically happen when they do this.

Finally, if people really don't want to bother with the 'verify markup' date check in iset.mm, I am happy to fix the Travis failure myself, like I did for this PR.  The Travis check tells me the date is wrong and should be fixed.  That is much better than Travis not checking the date and hoping that I remember to check it by hand occasionally.

benjub:
I think I fall in the "archaeology" case mentioned by @digama0.  I'd prefer to have a date in plain text within the file, as currently done, as a safety future-proof measure.  If a tool could automatically update the date on the first line, as @jkingdon seems to hint, that would be, I think, the best solution.

digama0:
> Usually they download the current set.mm from the Metamath site and put it into a local directory and make their edits. Sometimes their changed set.mm lags the GitHub version. Having the date at the top is useful for these situations so that I can know instantly (within a day at least) what version they were working from.

The web site version of set.mm qualifies as a build artifact here. It is fairly easy to get a script to download the current develop commit, put the commit date on the first line, and upload it to the web site. It can even be run as a post-commit hook so that the web site always has the latest version, with the right date.

But I think it is a very bad idea to have dates on the develop branch. I have never seen this practice in any software project, and it is easy to see why. Changelogs are similar, but they have more useful content and are append only, which makes them a bit easier to handle. (They are also only updated every new release rather than every commit.)

benjub:
I think that the solution proposed by @digama0 is best (a script putting the commit date on the master, while the develop branch does not have a top date).

nmegill:
> I think I fall in the "archaeology" case mentioned by @digama0. I'd prefer to have a date in plain text within the file, as currently done, as a safety future-proof measure. If a tool could automatically update the date on the first line, as @jkingdon seems to hint, that would be, I think, the best solution.

'verify markup */f' is almost such a tool, requiring a manual 5-second edit to line 1 when it tells you what the top date should be.  Is this really asking too much?  Sure, the edit could be automated, but so far the problem has been forgetting to run 'verify markup */f' in the first place.  I don't think people ignore its error messages when they do run it.  Most other markup error messages require manual edits to fix; why is this one different?

As I described, I am probably the main user of this date.  I'm happy to do the 5-second edit myself when people forget.  I should have shut up yesterday and just done it.  :)  I would prefer to leave Travis the way it is so that it notifies me that the edit is required for iset.mm.

jkingdon:
Glad to handle this however is most convenient for @nmegill

nmegill:
@digama0 I think this is being blown up out of proportion and is making a mountain out of a molehill.  The top date is useful for my purposes, and I want to keep it.  I also don't to have different versions of set.mm with and without a top date - that is just asking for more complexity, trouble, and confusion than it is worth.

As I indicated, I am happy to be responsible for the top date in the case of set.mm, or when people forget to run 'verify markup' for the other .mm's.  It is just not a big deal.

PR #985, opened 2019-07-07
RR^I is a metric space. 

tirix:
And additional theorems about the less-than geometric relation


benjub:
@tirix Do you think you could change the token from RR^ to RR^() and accordingly for unicode, GIF and LaTeX ?
This was @digama0's proposal, since RR^ should denote the direct product.  This reflects the standard notation $A^{(I}}$ versus $A^I$ for direct sum versus direct product.

Also, do you confirm that EEhil will eventually be simply EE, once you merge the other approaches in mathboxes with yours ?

tirix:
@benjub Yes, I can do the ‘RR^()’ change in my next PR.

Concerning the naming ‘EEhil’, the idea was to follow the same model as ‘CCfld’, ‘RRfld’, and ‘ZZring’, namely, including the name of the base set ‘EE’, and the structure we’re giving it, in this case, the structure of a Hilbert space.

‘EE’ is already the name of the base set of that structure (changing the definition of EE so that it includes the 0-dimension Euclidean space, ‘{ 0 }’, is also on my list)

benjub:
@tirix The thing is that "E" stands for "Euclidean", which means "finite-dimensional real Hilbert", so "EEhil" sounds redundant. But this is nitpicking, and I'm fine with whatever you decide.  Is EE exactly the underlying space of EEhil ? If yes, that's all good... and it should be proved... something like
  |- ( Base ` EEhil ) = EE
Thanks.

tirix:
Actually, there is a very minor difference, since ` EE ` is defined for dimensions 1 and above, and ` EEhil ` for dimensions 0 and above. As a matter of fact, they agree on dimensions 1 and above. We could prove:
` |- ( N e. NN -> ( Base ' ( EEhil ' N ) ) = ( EE ' N ) `
Right now the definition of ` EE ` is well after ` EEhil `.

benjub:
I see that you wrote above that including dimension 0 for EE was on your todo list, which is a good idea, and then you can prove
  |- ( N e. NN0 -> ( Base ' ( EEhil ' N ) ) = ( EE ' N ) )
once both are defined. That works!

PR #991, opened 2019-07-08
Misc

benjub:
Mathbox and typos; the two commits say it all (and also: set of upper integers --> upper set of integers).


benjub:
I'll fix this tomorrow...

PR #999, opened 2019-07-11
modify df-tru to use implication (id) instead of biconditional (biid)

benjub:
* Modify df-tru to use implication (id) instead of biconditional (biid) as definiens (with @digama0's agreement, https://groups.google.com/d/msg/metamath/t68Dkr0euMw/j6eq8FiOCAAJ).
* Prove the older definition as df-tru2.
* Adapt trujust and relabel the previous trujust as trujustOLD.
* Slightly reorder statements to have the two consecutive triples ($c T. ; wtru ; df-tru) and ($c F. ; wfal ; df-fal).

@nmegill As for trujustOLD and/or dftru2: I think they can be removed right now without waiting for a year.  Tell me what you prefer.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1534,6 +1534,7 @@
 "bj-inftyexpidisj" is used by "bj-minftynrr".
 "bj-inftyexpidisj" is used by "bj-pinftynrr".
 "bj-inftyexpiinv" is used by "bj-inftyexpiinj".
+"bj-vexw" is used by "bj-ralvw".
 "blo3i" is used by "ipblnfi".
 "blocn" is used by "blocn2".
 "blocn2" is used by "ubthlem1".
@@ -14215,6 +14216,7 @@ New usage of "bj-inftyexpiinv" is discouraged (1 uses).
 New usage of "bj-rabtrALT" is discouraged (0 uses).
 New usage of "bj-rabtrAUTO" is discouraged (0 uses).
 New usage of "bj-sbeqALT" is discouraged (0 uses).
+New usage of "bj-vexw" is discouraged (1 uses).
 New usage of "bj-xpima1snALT" is discouraged (0 uses).
 New usage of "blo3i" is discouraged (1 uses).
 New usage of "blocn" is discouraged (1 uses).
--- a/discouraged
+++ b/discouraged
@@ -15265,6 +15265,7 @@ New usage of "dfpjop" is discouraged (4 uses).
 New usage of "dfprm2OLD" is discouraged (0 uses).
 New usage of "dfsup2OLD" is discouraged (1 uses).
 New usage of "dfsup3OLD" is discouraged (0 uses).
+New usage of "dftru2" is discouraged (0 uses).
 New usage of "dfvd1imp" is discouraged (1 uses).
 New usage of "dfvd1impr" is discouraged (1 uses).
 New usage of "dfvd1ir" is discouraged (16 uses).
--- a/discouraged
+++ b/discouraged
@@ -15267,7 +15267,6 @@ New usage of "dfpjop" is discouraged (4 uses).
 New usage of "dfprm2OLD" is discouraged (0 uses).
 New usage of "dfsup2OLD" is discouraged (1 uses).
 New usage of "dfsup3OLD" is discouraged (0 uses).
-New usage of "dftru2" is discouraged (0 uses).
 New usage of "dfvd1imp" is discouraged (1 uses).
 New usage of "dfvd1impr" is discouraged (1 uses).
 New usage of "dfvd1ir" is discouraged (16 uses).
--- a/discouraged
+++ b/discouraged
@@ -18457,6 +18457,7 @@ Proof modification of "bj-axc16" is discouraged (5 steps).
 Proof modification of "bj-axc16g" is discouraged (30 steps).
 Proof modification of "bj-axd2d" is discouraged (16 steps).
 Proof modification of "bj-axdd2" is discouraged (28 steps).
+Proof modification of "bj-axext3" is discouraged (57 steps).
 Proof modification of "bj-axrep1" is discouraged (190 steps).
 Proof modification of "bj-axrep2" is discouraged (181 steps).
 Proof modification of "bj-axrep3" is discouraged (122 steps).


nmegill:
@benjub We are going to make df-tru become ( A. x x = x -> A. x x = x ) so I'm not sure I understand the benefit of this.

benjub:
I did not know the decision was taken, but if indeed you plan to make the change you propose soon, then the present PR is not useful.

david-a-wheeler:
I also thought we were going to use ` ( A. x x = x -> A. x x = x ) ` instead.

Note that there are a number of *other* unrelated changes in this patch, e.g., "distinguish" to "different".  I think it'd be better to move those potential changes (if still desired) into different pull requests so they could be considered separately.  Some don't seem useful, some *do* seem useful.  E.g.:

- is changing "distinguished" to "different" a good thing?
- the added comment for infmxrlb looks helpful to me and we *should* do that.

benjub:
@david-a-wheeler :
* "distinguished" may have other meanings, for instance in a topological space, two points are "distinguished" if they do not have the same neighborhood systems.  On the other hand, "different" has a clear meaning.
* Since I'm not well versed into github, I do not always know if it's better to group the small changes or not.  I don't want to cause too much work to the person doing the merge...

digama0:
Regarding "distinguished" v. "different", they are not synonyms, but in the statement it is proving that +oo and -oo are not equal, which is accurately described as "different" rather than "distinguished" (which may also be applicable but is not relevant to the theorem).

digama0:
As long as everything in the PR is uncontroversial, there isn't much of a problem with disparate stuff in it, although it's less organized. But if there is contention, then it's hard to accept only part of a PR, so in that case it's best to split up the parts so that some stuff gets in and other stuff doesn't. In this case, I would want the stuff that involves changes to df-tru to be separate from anything else, because changing axioms is always contentious.

benjub:
I reverted the commit regarding df-tru (git is not that bad after all!).

nmegill:
I have already done some work defining df-tru with  ( A. x x = x -> A. x x = x ).  I tried to make the pred calc intrusion into prop calc as "invisible" as possible by introduce the $v for x in the scope of df-tru (and the first theorem) so that it can't be referenced outside that scope.  The $v for x is later re-introduced globally at the start of pred calc like it is now.  I will try to merge this work into set.mm tonight.

I haven't yet moved up any prop calc theorems after df-tru that don't need df-tru.  That needs to be done carefully because we don't won't to break up studies of prop calc axiomatizations that use "F.".  But that is lower priority and can be done later.

nmegill:
I put the definition of df-tru as ( A. x x = x -> A. x x = x ) in commit 4653b13.

nmegill:
@benjub If you want to change dftru2 (the renamed df-tru) to use ph -> ph instead of ph <-> ph, feel free to do so at your convenience.

benjub:
@nmegill I propose to:
* change dftru2 to use -> instead of <-> (this is very minor)
* currently, dftru2 is the only theorem that references df-tru, contrary to what is written in the comment of df-tru.  I propose to prove dftru2 from tru (via 2th), istelf proved from df-tru.
* Rearrange the order so as to have: trujust ; $c T. ; wtru ; df-tru ; $c F. ;  wfal ; df-fal ; tru ; fal ; dftru2 ...
* Rearrange the sectioning to have:
1.2.9. Other logical connectives
1.2.9.1. Logical NAND
1.2.9.2. Logical XOR
1.2.10. Theorems for VD... <--- I volunteer to progressively move all theorems either in earlier sections or to AS's mathbox, so as to remove this subsection
1.2.11. Half-adder and full-adder in propositional calculus <-- replace cadtru with |- ps => |- cadd(ps,ps,ph)
1.2.12. Propositional calculus with logical constants
1.2.12.1. Universal quantifier for use by df-tru
1.2.12.2. Equality predicate for use by df-tru
1.2.12.3. The logical constants True and False
1.2.12.4. Truth tables
1.3....

nmegill:
> @nmegill I propose to:
> 
> * change dftru2 to use -> instead of <-> (this is very minor)
> * currently, dftru2 is the only theorem that references df-tru, contrary to what is written in the comment of df-tru.  I propose to prove dftru2 from tru (via 2th), istelf proved from df-tru.
> * Rearrange the order so as to have: trujust ; $c T. ; wtru ; df-tru ; $c F. ;  wfal ; df-fal ; tru ; fal ; dftru2 ...

The above is all OK.

> * Rearrange the sectioning to have:
>   1.2.9. Other logical connectives
>   1.2.9.1. Logical NAND
>   1.2.9.2. Logical XOR
>   1.2.10. Theorems for VD... <--- I volunteer to progressively move all theorems either in earlier sections or to AS's mathbox, so as to remove this subsection
>   1.2.11. Half-adder and full-adder in propositional calculus <-- replace cadtru with |- ps => |- cadd(ps,ps,ph) ; remove cadanOLD
>   1.2.12. Propositional calculus with logical constants
>   1.2.12.1. Universal quantifier for use by df-tru
>   1.2.12.2. Equality predicate for use by df-tru
>   1.2.12.3. The logical constants True and False
>   1.2.12.4. Truth tables
>   1.3....

The above content changes are OK, but I object to changing the section levels and section titles.

In particular, I used subsubsections to group A. , =, and df-tru together as "pieces" of the final df-tru definition, so I indented them all together so these three would stand out as building up to df-tru.  Obviously we don't want to use A. and = outside of df-tru.  But by making everything around it subsubsections also, we lose the obvious grouping of A. , =, and df-tru.   However, I also don't want to add a a subsubsubsection level just to handle this one case - it is a lot of work to add a new level to metamath.exe.

In other words I would like to keep the present titles and levels starting from the present "1.2.11  True and false constants", and would like to keep that title rather than change it to "Propositional calculus with logical constants" (there are never going to be more than 2 anyway).  "1.2.12 Truth tables" should start a new subsection (not subsubsection) like it does now, so it becomes obviously separated from the {A., =, df-tru} grouping.  I think it is important to emphasize that grouping, because the mixture of prop and pred calc in that area is bound to be confusing to some people.

I also don't see the point of wasting a subsubsection level for "Other logical connectives".  I don't feel as strongly about this one if you really want to do that, but all the other connectives (AND, OR,...) have their own subsection level, and the new title "Other logical connectives" is an extra line in the TOC that seems unnecessary just to announce two new connectives.

nmegill:
> * Rearrange the order so as to have: trujust ; $c T. ; wtru ; df-tru ; $c F. ;  wfal ; df-fal ; tru ; fal ; dftru2 ...

This will be somewhat less elegant because we will have to break up the scoping ${ ... $} into 2 pieces in order to introduce $c T. in the middle (a constant can be declared only in the outermost scope).  Moreover, it requires reintroducing the local setvar x again in the 2nd scope, with a different label.

I won't object if you have a good reason for doing this, but keep in mind that it breaks up the local setvar scope, which I tried to keep in one place to minimize the introduction of local setvars.

nmegill:
@benjub Here is my proposed text for the df-tru comment.
```
    $( Definition of the truth value "true", or "verum", denoted by ` T. ` .
       This is a tautology, as proved by ~ tru .  In this definition, an
       instance of ~ id is used as the definiens, although any tautology, such
       as an axiom, can be used in its place.  This particular ~ id instance
       was chosen so this definition can be checked by the same algorithm as is
       used for predicate calculus.  This definition should be referenced
       directly only by ~ tru , and other proofs should depend on ~ tru
       (directly or indirectly) instead of this definition, since there are
       many alternative ways to define ` T. ` .  (Contributed by Anthony Hart,
       13-Oct-2010.)  (Revised by NM, 11-Jul-2019.)

```
Of course tru would go where dftru2 is now (i.e. inside of the scope with setvar x), and dftru2 would go where tru is now.  Whether to "discourage" the use of dftru2 is optional and probably unnecessary, but you can decide.

benjub:
So to be sure: is the following ok?
1.2.9.  Logical 'nand' (Sheffer stroke)
1.2.10.  Logical 'xor'
1.2.11. Theorems for VD... [to be remove progressively]
1.2.12. Half-adder and full-adder in propositional calculus
1.2.13. True and false constants
1.2.13.1. Universal quantifier for use by df-tru
1.2.13.2. Equality predicate for use by df-tru
1.2.13.3. Define the true and false constants
1.2.14. Truth tables
1.3....

digama0:
I object to the change to cadtru. The "generalized" cadtru already exists, it's called cad11. If df-tru is moving after df-cad, then this theorem should move with it, perhaps near the truth tables.

benjub:
As for the order, can I simply move tru between df-tru and dftru2, inside the scope?

benjub:
@digama0 ok to move cadtru as the last theorem in the subsection "Truth tables"

nmegill:
> So to be sure: is the following ok?
> 1.2.9. Logical 'nand' (Sheffer stroke)
> 1.2.10. Logical 'xor'
> 1.2.11. Theorems for VD... [to be remove progressively]
> 1.2.12. Half-adder and full-adder in propositional calculus
> 1.2.13. True and false constants
> 1.2.13.1. Universal quantifier for use by df-tru
> 1.2.13.2. Equality predicate for use by df-tru
> 1.2.13.3. Define the true and false constants
> 1.2.14. Truth tables
> 1.3....

This is fine with me (subject to Mario's later comments).

nmegill:
> As for the order, can I simply move tru between df-tru and dftru2, inside the scope?

I think dftru2 should be outside the scope because its proof from tru will not need any setvar.

Basically:

$c T.
${
  $v...
  trujust $p...
  df-tru $a...
  tru $p... (proved from df-tru)
$}
dftru2 $p... (proved from tru)

dftru2 is probably not useful for anything.  It exists to illustrate how df-tru could be defined without pred calc machinery (but w/ more complex justification).

PR #1000, opened 2019-07-12
True

benjub:
Following #999; commit messages are descriptive. DO NOT merge yet, since I have to deal with hadbi123i and cadbi123i.
@digama0 What should I do with hadbi123i and cadbi123i, which use dftru in their proofs? (for the moment, they are commented out)
- Delete them? (since they are not used by any theorems and cadbi123d and hadbi123d might suffice)
- Reprove them: directly? or from the decution version, using (ph -> ph) as an antecedent?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4752,7 +4752,7 @@
 "df-topspOLD" is used by "eltopspOLD".
 "df-topspOLD" is used by "istpsOLD".
 "df-topspOLD" is used by "tpsexOLD".
-"df-tru" is used by "dftru2".
+"df-tru" is used by "tru".
 "df-unop" is used by "elunop".
 "df-va" is used by "0vfval".
 "df-va" is used by "vafval".
@@ -4804,7 +4804,6 @@
 "dfpjop" is used by "elpjidm".
 "dfpjop" is used by "pjhmopidm".
 "dfsup2OLD" is used by "dfsup3OLD".
-"dftru2" is used by "tru".
 "dfvd1imp" is used by "gen11".
 "dfvd1impr" is used by "gen11".
 "dfvd1ir" is used by "2uasbanhVD".
@@ -15268,7 +15267,7 @@ New usage of "dfpjop" is discouraged (4 uses).
 New usage of "dfprm2OLD" is discouraged (0 uses).
 New usage of "dfsup2OLD" is discouraged (1 uses).
 New usage of "dfsup3OLD" is discouraged (0 uses).
-New usage of "dftru2" is discouraged (1 uses).
+New usage of "dftru2" is discouraged (0 uses).
 New usage of "dfvd1imp" is discouraged (1 uses).
 New usage of "dfvd1impr" is discouraged (1 uses).
 New usage of "dfvd1ir" is discouraged (16 uses).


digama0:
`trud` is very important for doing anything in deduction style. I'm not happy to hear about these changes, and I'm not clear on why proofs need to be rewritten. I thought the idea was to move the definitions of the syntax of forall and equals way back into prop calc, so that df-tru and tru can stay where they are. Hence most theorems should not need to be changed or reordered, so what is this doing?

digama0:
Also, congratulations for PR 1000!

benjub:
I can revert these commits if you want. I just thought moving the logical constants as late as possible in the prop calc part was better, since the new definition uses syntax foreign to prop calc. But we can keep the present sectioning if you prefer.

digama0:
What I want to avoid is the suggestion that df-tru is "foreign to prop calc" in the first place. It's clearly an important part of prop calc, and our choice to implement it using forall should not demote it to second class. Rather, we should see forall and equals (and class variables) as being displaced from their appropriate locations, and I don't know whether it is necessary to have a bunch of sections to discuss these displaced constants. I'd prefer to have only one, with references to the beginning of the sections where forall and classes are taken up for real. (I think for the longest time the introduction of class variables for = was only discussed in comments in the metamath file, not on the web pages, but I see it's since been turned into section headers, which gives them more exposure. While there are some subtle things going on, I am of mixed opinion whether we should give them a lot of air-time in the middle of a prop calc section, where it distracts from standard prop calc.)

benjub:
I'm going to revert the contentious second commit.  I agree with Mario's latest message. If we don't want to make |- T. an axiom, then there is no optimal solution, unfortunately.  I'll leave it up to @digama0 and @nmegill in case they want to make further modifications.

nmegill:
> What I want to avoid is the suggestion that df-tru is "foreign to prop calc" in the first place. It's clearly an important part of prop calc, and our choice to implement it using forall should not demote it to second class. Rather, we should see forall and equals (and class variables) as being displaced from their appropriate locations, and I don't know whether it is necessary to have a bunch of sections to discuss these displaced constants. I'd prefer to have only one, with references to the beginning of the sections where forall and classes are taken up for real.
> 
> (I think for the longest time the introduction of class variables for = was only discussed in comments in the metamath file, not on the web pages, but I see it's since been turned into section headers, which gives them more exposure. While there are some subtle things going on, I am of mixed opinion whether we should give them a lot of air-time in the middle of a prop calc section, where it distracts from standard prop calc.)

I'm open to changes to address this.  Mainly I want to make things clear for the reader.  Maybe we should delete the historical negative approach about (theorem) weq being a "patch" to correct an ambiguous parsing problem and instead present it in a straightforward positive way along the lines of "we do it this way in order to provide unambiguous parsing..."

I don't mind moving the description of these details down to the pred calc section.  It's just that the natural place for them to go is in the description sections of wal and w(c)eq, which are now located in the prop calc section.

However, we can postpone introducing theorem weq until the pred calc section and have the trujust and tru proofs reference cv and wceq directly.  Then the details of w(c)eq can be explained in the weq comment, which will be properly located in pred calc.

benjub:
@nmegill Your suggestion of moving weq to the pred calc section sounds good to me... but it looks like you just did it anyway?  Would it be possible for you to modify the files using github PRs?  Even if you directly merge them, it is easier for us to see what has changed.  Or maybe I am missing something?

Is it possible to remove the lines around "@c class-n @."?  I know they are commented out anyway, but they are a bit confusing

nmegill:
> @nmegill Your suggestion of moving weq to the pred calc section sounds good to me... but it looks like you just did it anyway? Would it be possible for you to modify the files using github PRs? Even if you directly merge them, it is easier for us to see what has changed. Or maybe I am missing something?

@benjub Let me look into doing that.
 
> Is it possible to remove the lines around "@c class-n @."? I know they are commented out anyway, but they are a bit confusing

@tirix Do you need this commented-out version of class-n and class-o, since you have the real versions later on?

tirix:
Yes, no problem to remove the commented-out versions of class-n and class-o and related stuff, all I need is the uncommented versions in my mathbox.

david-a-wheeler:
@digama0 :
> What I want to avoid is the suggestion that df-tru is "foreign to prop calc" in the first place.

Perhaps we need to not be so fussy about where the *syntax* declarations go.

There are systems where you have to declare all allowed syntax *first*, and then you can later use it.  I don't think set.mm should be organized that way, but there's really nothing wrong with declaring an allowed *syntax* early.  After all, you could have put all the syntax declarations like ` wal ` in one place.

So I guess I'm arguing for where we seem to be going: *usually* we put together new symbol constant declarations, syntax declarations, and the first axiom/definition that uses it... but if needed, we can put the symbol constant and syntax declarations earlier (without embarrassment).

digama0:
@david-a-wheeler This would go against my proposal to keep definitions next to syntax declarations, though. In this particular case it's not a problem because the relevant syntax - wal, wceq - are primitives so there is no definition to go along with it. For definitions I think the definition should stay close to the syntax declaration, because in the interval between the two it's an axiomatic constant, and I don't like the idea of axiomatic constants transforming into definitions later.

david-a-wheeler:
@digama0 - I agree that it's better to keep the definitions next to syntax declarations.  But I think making an exception for a *reason* - as is done on in this case - is just fine.

PR #1002, opened 2019-07-13
Regen from raw

david-a-wheeler:


david-a-wheeler:
This branch updates scripts/regen-from-raw, and then switches mmnf.html to use mmnf.raw.html instead.

david-a-wheeler:
Build failed, not sure why.  WIll try to check later.

david-a-wheeler:
@nmegill ... please merge!

nmegill:
A minor issue is that the ~'s in the original mmnf.html weren't escaped with ~~ in mmnf.raw.html.  For example, http://math.boisestate.edu/~holmes/holmes/nf.html should be http://math.boisestate.edu/~~holmes/holmes/nf.html, otherwise the 'markup' command thinks 'holmes...' is a label in nf.mm and gives an error message since it doesn't exist.  I fixed these by hand in mmnf.raw.html in commit 06727e0.

PR #1007, opened 2019-07-16
Add imim12 and imbi12

benjub:
Mainly, added imim12 and imbi12 (the latter from AS's mathbox). Smaller changes are described in commit messages.
Can I move bj-imim2ALT to the main part? This proof is actually simpler (see MM>show trace /count).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17637,6 +17637,7 @@ New usage of "pm11.07" is discouraged (0 uses).
 New usage of "pm110.643ALT" is discouraged (0 uses).
 New usage of "pm2.43bgbi" is discouraged (0 uses).
 New usage of "pm2.43cbi" is discouraged (2 uses).
+New usage of "pm2.86iALT" is discouraged (0 uses).
 New usage of "pmap1N" is discouraged (2 uses).
 New usage of "pmapglb2N" is discouraged (0 uses).
 New usage of "pmapglb2xN" is discouraged (1 uses).
@@ -19256,6 +19257,7 @@ Proof modification of "pm110.643" is discouraged (72 steps).
 Proof modification of "pm110.643ALT" is discouraged (35 steps).
 Proof modification of "pm2.43bgbi" is discouraged (16 steps).
 Proof modification of "pm2.43cbi" is discouraged (34 steps).
+Proof modification of "pm2.86iALT" is discouraged (16 steps).
 Proof modification of "prdsgsumOLD" is discouraged (415 steps).
 Proof modification of "probfinmeasbOLD" is discouraged (225 steps).
 Proof modification of "problem1" is discouraged (20 steps).
--- a/discouraged
+++ b/discouraged
@@ -8589,9 +8589,6 @@
 "iin1" is used by "sspwimpcf".
 "iin1" is used by "suctrALTcf".
 "imaelshi" is used by "rnelshi".
-"imbi12" is used by "imbi13".
-"imbi12" is used by "imbi13VD".
-"imbi12" is used by "sbcssgVD".
 "imbi13" is used by "trsbc".
 "imbi13" is used by "trsbcVD".
 "impexp3a" is used by "impexp3acom3r".
@@ -16472,7 +16469,6 @@ New usage of "iin1" is discouraged (3 uses).
 New usage of "iin2" is discouraged (0 uses).
 New usage of "iin3" is discouraged (0 uses).
 New usage of "imaelshi" is discouraged (1 uses).
-New usage of "imbi12" is discouraged (3 uses).
 New usage of "imbi12VD" is discouraged (0 uses).
 New usage of "imbi13" is discouraged (2 uses).
 New usage of "imbi13VD" is discouraged (0 uses).
@@ -18615,6 +18611,7 @@ Proof modification of "bj-hbab1" is discouraged (20 steps).
 Proof modification of "bj-hblem" is discouraged (33 steps).
 Proof modification of "bj-hbs1" is discouraged (22 steps).
 Proof modification of "bj-hbsb2v" is discouraged (31 steps).
+Proof modification of "bj-imim2ALT" is discouraged (10 steps).
 Proof modification of "bj-inrab2" is discouraged (48 steps).
 Proof modification of "bj-isseti" is discouraged (15 steps).
 Proof modification of "bj-issetw" is discouraged (53 steps).
@@ -19069,7 +19066,6 @@ Proof modification of "iidn3" is discouraged (8 steps).
 Proof modification of "iin1" is discouraged (1 steps).
 Proof modification of "iin2" is discouraged (1 steps).
 Proof modification of "iin3" is discouraged (1 steps).
-Proof modification of "imbi12" is discouraged (30 steps).
 Proof modification of "imbi12VD" is discouraged (121 steps).
 Proof modification of "imbi13" is discouraged (34 steps).
 Proof modification of "imbi13VD" is discouraged (67 steps).


benjub:
Strange Travis fail:
```
$ jdk_switcher use oraclejdk8
jdk_switcher: command not found
The command "jdk_switcher use oraclejdk8" failed and exited with 127 during .
```

david-a-wheeler:
That is a bizarre error message that should not happen. I suspect that there was a temporary problem in Travis. I have restarted the build, hopefully that will fix it.

benjub:
Thanks David.  How did you restart the build, by the way?
Looks like it's not working. Possibly related:
https://github.com/apache/rocketmq/pull/1242
https://github.com/michaelklishin/jdk_switcher/issues/19

david-a-wheeler:
Weird. The changes made to .travis.yml may help:
https://github.com/apache/rocketmq/pull/1242/files

wlammen:
There are some criteria by which a proof formally qualifies as shorter than another.  I think you find them in set.mm itself close to the beginning.  Or somewhere on the Metamath web pages... I am at work, and cannot look them up right now.  If I remember right, first comes the dependencies on axioms, then the number of bytes used in compressed mode, if tied, the count of symbols used in the display of the proof, and then... I don't know, perhaps Norms personal liking.
I am sorry, by these criteria the legacy proof of pm2.86 is shorter than the proposed one.  During my course of shortenings including FOL layer, I occasionally found proofs that for various reasons appealed me more than the finally picked one.  Reasons: From a pedagogical perspective, I would have preferred the looser, or one proof is hampered by a less developed foundation, though more instructive, and so on, and so forth.
Although I concede I every now and then was tempted to violate the rules myself, I always resisted. In total, the criteria are not perfect, but usually work good enough.  Meaning, on the whole, in this process basic theorems slowly move upwards where they can assist the further development of theorems, and this holds surprisingly often. 
So from my point of view, I'd rather see a discussion first before changing the rules.

Wolf

benjub:
Reply to @wlammen :

It goes without saying, and it is probably not your concern either in the present case, but to be clear: I do not care, here, about the credits, and I'll be happy to change these theorems to "Contributed by NM" or "Contributed by Wolf Lammen".

I think both proofs should stay in set.mm since both have advantages.  One of them should be "ALT"; you or @nmegill decide which one.  It looks to me that the one I proposed is "shorter" since it has the same compressed size as the other one:
```
  pm2.86i $p |- ( ph -> ( ps -> ch ) ) $=
     ( wi pm2.86 ax-mp ) ABEACEEABCEEDABCFG $.
  pm2.86iALT $p |- ( ph -> ( ps -> ch ) ) $=
     ( wi ax-1 syl com12 ) BACBABEACEBAFDGH $.
```
but only 2 essential steps versus 3.  (By the way: I think the number of essential steps should come first in determining the shorter proof.)  Or maybe you count the length of the block of capital letters?  Anyway, this is not very important and I'm fine with either one being the "ALT" one.

I think that when several proofs have advantages (e.g., one is shorter, another one is more direct, a third one is more enlightening for whatever reason), then all should be in set.mm.  This is a reason why I'm asking to move bj-imim2ALT to the main part.  (And again, you had probably already thought of it, so I'll be happy to make it "Contributed by Wolf Lammen", or "by NM".)

nmegill:
(@benjub & @wlammen I am busy with another project but will try to respond in a day or so.)

avekens:
Is somebody looking after the Travis problem? I want to create a PR by myself, but if this problem still occurs I have to wait...

benjub:
Could this problem be due to the recent changes made by @david-a-wheeler to
the scripts?  I see that .travis.yml was changed 5 days ago (although those
changes look at first sight unrelated to the "jdk-switcher" thing).

david-a-wheeler:
@nmegill : I'm not certain what the problem is.  I don't think my recent changes caused this.  Instead, I think there's a change in Travis that is causing the problem.

I verified that any Travis run fails now:
https://travis-ci.org/metamath/set.mm/builds/560595979?utm_source=github_status&utm_medium=notification

This change looks like a solution for a suspiciously similar problem:
https://github.com/apache/rocketmq/pull/1242/files

I created a new branch to try this solution out:
https://github.com/metamath/set.mm/tree/fix-travis

We'll see if it works here:
https://travis-ci.org/metamath/set.mm/builds/560599963?utm_source=github_status&utm_medium=notification

david-a-wheeler:
It seems to have worked!  Before the fix-travis branch, even useless changes were failing.  Once the fix-travis branch was added, the testing worked again.

Let's wait for @nmegill to check on the new fix-travis branch.  If he merges it into "develop" (& I hope he will), you can then just "git merge" the latest "develop" branch into your changes & all should work again.  Hopefully :-).

david-a-wheeler:
@nmegill - see pull request https://github.com/metamath/set.mm/pull/1008 that I believe will fix the problem.

nmegill:
> (@wlammen :) There are some criteria by which a proof formally qualifies as shorter than another. I think you find them in set.mm itself close to the beginning. Or somewhere on the Metamath web pages... 

Look for the section near the beginning of set.mm called "6.  Acceptable shorter proofs".  This was written many years ago where I proposed some criteria for what counts as "shorter".  In the end it is just my opinion of what I thought reasonable in case a dispute arose, but certainly there can be borderline cases or other opinions that can also be valid.

> I am sorry, by these criteria the legacy proof of pm2.86i is shorter than the proposed one. During my course of shortenings including FOL layer, I occasionally found proofs that for various reasons appealed me more than the finally picked one. Reasons: From a pedagogical perspective, I would have preferred the looser, or one proof is hampered by a less developed foundation, though more instructive, and so on, and so forth.

From the strict perspective of the stated rules, I think @wlammen is right.

Here is a quote from the mentioned section in set.mm:
```
In borderline cases, I typically place more importance on the number of
compressed proof steps and less on the length of the label section (since the
names are in principal arbitrary).  If two proofs have the same number of
compressed proof steps, I will typically give preference to the one with the
smaller number of different labels, or if these numbers are the same, the proof
with the fewest number of characters that the proofs happen to have by chance
when label lengths are included.

A few theorems have a longer proof than necessary in order to avoid the use of
certain axioms, for pedagogical purposes, and for other reasons.  These
theorems will (or should) have a "(Proof modification is discouraged.)" tag in
their description.  For example, id1 shows a proof directly from axioms.
Shorter proofs for such cases won't be accepted, of course, unless the criteria
described continues to be satisfied.
```

> Although I concede I every now and then was tempted to violate the rules myself, I always resisted. In total, the criteria are not perfect, but usually work good enough. Meaning, on the whole, in this process basic theorems slowly move upwards where they can assist the further development of theorems, and this holds surprisingly often.

 
> (@benjub :) It goes without saying, and it is probably not your concern either in the present case, but to be clear: I do not care, here, about the credits, and I'll be happy to change these theorems to "Contributed by NM" or "Contributed by Wolf Lammen".
> 
> I think both proofs should stay in set.mm since both have advantages. One of them should be "ALT"; you or @nmegill decide which one. It looks to me that the one I proposed is "shorter" since it has the same compressed size as the other one:
> 
> ```
>   pm2.86i $p |- ( ph -> ( ps -> ch ) ) $=
>      ( wi pm2.86 ax-mp ) ABEACEEABCEEDABCFG $.
>   pm2.86iALT $p |- ( ph -> ( ps -> ch ) ) $=
>      ( wi ax-1 syl com12 ) BACBABEACEBAFDGH $.
> ```
> 
> but only 2 essential steps versus 3. (By the way: I think the number of essential steps should come first in determining the shorter proof.) Or maybe you count the length of the block of capital letters?

Yes, including the length of the labels is somewhat meaningless since the length of a label is an arbitrary human choice.  The first one above should be the "ALT" one.  The _number_ of different labels can be a criterion (obviously the fewer the better) but just so that we have a definite rule, I would say the length of the capital letter block should be the primary determining factor for "shorter" even if, including labels, it has more characters in total.

> Anyway, this is not very important and I'm fine with either one being the "ALT" one.
> 
> I think that when several proofs have advantages (e.g., one is shorter, another one is more direct, a third one is more enlightening for whatever reason), then all should be in set.mm. This is a reason why I'm asking to move bj-imim2ALT to the main part. (And again, you had probably already thought of it, so I'll be happy to make it "Contributed by Wolf Lammen", or "by NM".)

For your (Benoit's) version, which should become the ALT version and go after the original non-ALT version, it should probably be "(Contributed by NM, 5-Aug-1993.) (Revised by BJ, xx-Jul-2019.)".  You should also have "(Proof modification discouraged.)" so that 'minimize' won't reduce it to a 1-step proof.  There should also be a comment as to what the advantage of the ALT version is.  Unfortunately there is no "Proof revised by" tag, since I never anticipated anything other than shortening, but I suppose I should add it since "Revised by" normally means a change to the theorem itself.

As for having several proofs for the same theorem, I have mixed feelings about that.  Almost any proof could be done in multiple ways, and adding multiple versions bloats set.mm and adds unnecessary matches when searching for theorems we need for a proof.  I suppose a few ALTs here and there won't do any great harm, but it isn't something I would want to do routinely.  There should be a clear purpose to the ALT version like we have for id vs. id1.  Or hartogs vs. ondomon where the latter is much shorter by allowing the use of AC.  I think pm2.81i vs. pm2.81iALT is a really marginal case, but you have my OK to move it up if you feel strongly about it.

If prop calc proofs are assigned to a classroom, there could be almost as many different proofs as there are students, and each version might be the one its creator thinks is the most natural (or more realistically the first one the creator obtained after a great deal of struggle, then immediately stopped working on in order to continue with the next problem).  Picking the shortest one provides a completely objective criterion, and I would argue that in most cases it would also be the easiest to understand or even the most "beautiful" in some subjective sense.

Finally, if we give the student many variations of many proofs, is that really helpful, outside of maybe a small number of examples showing alternate proofs are possible?  Students learn manly by working out proofs themselves, and it is already bad enough that they can cheat by consulting the set.mm version.  I think most shorter proofs were found by motivated people who reproved the theorems from scratch (hiding the set.mm version) then kept the ones that ended up shorter.  Wolf, the expert, can confirm or deny that. :)

nmegill:
@david-a-wheeler It looks like the same problem is there.  Did I correctly merge what you asked?

david-a-wheeler:
@nmegill ... now that the fix has been merged into the develop branch, I hope that they can simply do a git merge from develop into this branch and fix the problem.

nmegill:
@david-a-wheeler What I meant in my previous message was that I re-ran Travis on this PR, after I merged your fix, and the failure was still there.  But I'll go ahead and merge despite the Travis error.

david-a-wheeler:
I did not mean to merge this into develop. I meant to merge develop into this, that is, exactly the reverse:

~~~~
git checkout develop
git pull
git checkout misc
git merge develop
~~~~

But no matter, it looks like everything worked out.

PR #1008, opened 2019-07-18
Attempt to fix Travis problems with Java

david-a-wheeler:
Attempt to fix the Travis check, which is currently failing.
The failure manifests after the "unzip -q mmj2jar.zip" as:

~~~~
$ jdk_switcher use oraclejdk8
jdk_switcher: command not found
The command "jdk_switcher use oraclejdk8" failed and exited with 127 during .
~~~~

I found that rocketmq had a similar problem, so I quickly
tried reusing their solution:
https://github.com/apache/rocketmq/pull/1242/files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,11 @@
 #
 # We typically rewrap the file using scripts/rewrap.
 #
+dist: trusty
+
 language: rust
+jdk:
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@nmegill - this seems to fix the problem, please merge

PR #1012, opened 2019-07-20
shorten mo3

wlammen:


benjub:
@wlammen  Why is mo3OLDOLD obsolete?  The proof looks shorter.  Am I missing something?

wlammen:
Sometimes a shortening involves two or more theorems. One is growing, but this is more than compensated for by shrinking the other.  In total, the number of compressed proof bytes always become less.  In this case mo3 has been made a fundamental theorem, and moved upwards.  In such a process, the fundamental theorem may capture much of the proof logic, while the rest degenerate to simple applications of it.  To learn more about the details, simply check out the historic state in git.  If you dont know how to proceed, feel free to come back on me.

benjub:
Thanks @wlammen !  Indeed, I hadn't seen that mo3OLDOLD uses mo, which itself uses mo3.  I should have looked more carefully.

PR #1015, opened 2019-07-21
shorten truan

wlammen:
Apart from the shortening, I marked mp2ALT and a1iiALT as obsolete.  If I remember right, these were once overridden by a proof that was marginally longer, but refers to less axioms.  Such a proof has priority anyway, and could have just replaced the other.  I do not see any value in keeping the older one.  The procedure is the same as with bisym.


nmegill:
The deletion of mp2ALT and a1iiALT is ok with me.

PR #1018, opened 2019-07-22
shorten pm2.21dd

wlammen:
The appearance of this theorem reminds me more of pm2.65i than of pm2.21d, so I think its name is a bit unfortunate.  Following similar pattern (pm2.621 e.g.) elsewhere, I suggest to rename it to pm2.651i, indicating a variant of pm2.65i.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17591,6 +17591,7 @@ New usage of "plpv" is discouraged (1 uses).
 New usage of "pm10.252" is discouraged (0 uses).
 New usage of "pm11.07" is discouraged (0 uses).
 New usage of "pm110.643ALT" is discouraged (0 uses).
+New usage of "pm2.21ddOLD" is discouraged (0 uses).
 New usage of "pm2.43bgbi" is discouraged (0 uses).
 New usage of "pm2.43cbi" is discouraged (2 uses).
 New usage of "pm2.86iALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> The appearance of this theorem reminds me more of pm2.65i than of pm2.21d, so I think its name is a bit unfortunate. Following similar pattern (pm2.621 e.g.) elsewhere, I suggest to rename it to pm2.651i, indicating a variant of pm2.65i.

I think we should leave it as it is.  pm2.21dd is clearly a deduction form of pm2.21, going one step further than pm2.21d in breaking it down.  Also we have pm2.21fal that's a substitution instance.  Also, pm2.65i seems intrinsically different because it uses the negation of the antecedent to serve as the conclusion.

I also don't really like the name pm2.621, which I find hard to remember.  Perhaps I'm wrong, but based on other 4-digit names in PM, I suspect that Russell slipped it into into his index cards in as an afterthought when he found he needed it later.  He added the extra decimal to make it sort in the right place rather then renumbering the ones that followed (a big and error-prone effort if you don't have a computer to search through the book).  I find pm2.21... easier to remember because it is used a lot and it has fewer digits.

For the very long term, it would be nice to devise more meaningful names for the pm* series (and the 19.* series) when we can.  I have been occasionally doing that when I think of a reasonable name; for example, pm2.06 became imim1.  But it is hard to come up with short names that fit into a pattern and are easy to remember.  So this effort will probably continue very slowly.

PR #1028, opened 2019-07-27
mathbox update

wlammen:


nmegill:
@wlammen The top date shouldn't be changed if there are other PRs in the queue.  Could you change it to 27-Jul-2018 to fix the conflict?  Thanks.

wlammen:
I did not change the date, since even in the western parts of US it is the 28th by now, but resolved the conflict instead (instructed git which date to use).

PR #1044, opened 2019-08-08
The axiom of choice implies excluded middle

jkingdon:
In the process of proving various ordinal theorems imply excluded middle, at some point I realized I was heading towards a larger prize: formalizing the well-known proof that the axiom of choice implies excluded middle. Although this formalization is for IZF (as with the rest of iset.mm), the result also holds for most forms of constructive set theory (see the Stanford Encyclopedia of Philosophy for details about which ones).

This theorem is a significant one and I suspect that if someone made a Top 100 list for constructive mathematics, it might make the list.

Anyway, enough self-promotion for now. The proof isn't super short (partly due to the complexities of stating and working with a suitable version of the axiom of choice), but it naturally divides into lemmas and follows the proof in Bauer pretty closely. So starting by reading that informal proof is where I'd start if I wanted to study this proof.

The pull request has a few minor cleanups and comment changes outside the main theorem, but they are small.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -36602,8 +36602,12 @@ $)
     $d x y z w s $.  $d z w s ph $.  $d x ph $.
     ordtriexmid.1 $e |- A. x e. On A. y e. On ( x e. y \/ x = y \/ y e. x ) $.
     $( Ordinal trichotomy implies the law of the excluded middle (that is,
-       decidability of an arbitrary proposition).  (Contributed by Mario
-       Carneiro and Jim Kingdon, 14-Nov-2018.) $)
+       decidability of an arbitrary proposition).
+
+       This theorem is stated in "Constructive ordinals", [Crosilla], p.
+       "Set-theoretic principles incompatible with intuitionistic logic".
+
+       (Contributed by Mario Carneiro and Jim Kingdon, 14-Nov-2018.) $)
     ordtriexmid $p |- ( ph \/ -. ph ) $=
       ( vz wn wo c0 wceq wcel w3o con0 cv eleq1 eleq2 3orbi123d wa wi 0ex ax-mp
       csn crab noel ordtriexmidlem eqeq1 0elon anbi2d eqeq2 imbi12d vtocl mpan2
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 2-Aug-2019
+$( iset.mm - Version of 7-Aug-2019
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm (with updates since then, including copying entire theorems
@@ -36666,10 +36666,10 @@ $)
 
     $( Lemma for ~ onsucelsucexmid .  The set
        ` { x e. { (/) , { (/) } } | ( x = (/) \/ ph ) } ` appears as ` A ` in
-       the proof of Theorem 1.3 in [Bauer] p. 483, and similar sets also appear
-       in other proofs that various propositions imply excluded middle, for
-       example in ~ ordtriexmidlem .  (Contributed by Jim Kingdon,
-       2-Aug-2019.) $)
+       the proof of Theorem 1.3 in [Bauer] p. 483 (see ~ acexmidlema ), and
+       similar sets also appear in other proofs that various propositions imply
+       excluded middle, for example in ~ ordtriexmidlem .  (Contributed by Jim
+       Kingdon, 2-Aug-2019.) $)
     onsucelsucexmidlem $p |- { x e. { (/) , { (/) } } | ( x = (/) \/ ph ) }
         e. On $=
       ( vy vz cv c0 wceq wo csn wcel word wa wi wal eleq2 adantl ex sylib mpbir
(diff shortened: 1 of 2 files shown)


nmegill:
Nice work!

david-a-wheeler:
Impressive!

jkingdon:
Thanks for the kind words, @nmegill and @david-a-wheeler . The key insight was that riota in general, and http://us2.metamath.org/ileuni/riotaeqbidv.html in particular, were going to take care of a lot of quantifier manipulation which was going to be pretty tedious if I tried to do it with theorems like http://us2.metamath.org/ileuni/euanv.html and http://us2.metamath.org/mpeuni/reuan.html . So http://us2.metamath.org/ileuni/acexmidlemab.html is the heart of this proof (at least in the sense of having been the hardest to come up with). Although it does have that nice property of looking nice and easy and obvious after the fact.

nmegill:
There is a subtlety that I thought I would mention.  You are assuming AC in the form of ac2, which is equivalent to ax-ac in set.mm.  The "standard" textbook AC is ax-ac2, and a weak component of ax-reg is needed to derive ax-ac from ax-ac2.  (This is mentioned below the axiom list at http://us.metamath.org/mpeuni/mmset.html#staxioms.)

I think that elirrv and en2lp are the only consequences of ax-reg that are needed for ax-ac2 -> ax-ac, and  I see that you derive these from ax-setind.  I don't know whether ax-ac2 -> ax-ac requires anything nonconstructive.

jkingdon:
Hey, thanks for the thoughts on various ways of stating choice.

I have made no attempt to avoid ax-setind in iset.mm. This is largely based on the sources I have to work with (for example, https://plato.stanford.edu/entries/nonwellfounded-set-theory/ does not seem to mention constructive set theory at all, and likewise the constructive sources don't seem to discuss non-well-foundedness), or maybe a better wording is "Because iset.mm assumes ax-setind without reluctance, all sets are well-founded. We could adopt a treatment more like set.mm if people want to investigate set theories which are constructive but which do not assume ax-setind." (currently in mmil.html under df-fr in the missing theorems list). I vaguely remember seeing some papers about subsystems of CZF, but I'm not really up on whether they include ax-setind, or some weak version thereof.

The language about inhabited versus non-empty is aimed at things like set.mm's ac3/ac5/ac8 (without getting too deep into listing theorem names which would need to be updated if set.mm changes). So that's what I was thinking of, more than ax-setind (or other formulations of regularity), when I thought about which one to pick. Whether the usual axiom of choice formulations (including Zorn's lemma, well-ordering theorem, etc) all imply excluded middle (and thus all the familiar equivalences) is not something I know (although I would conjecture yes for most/all).

jkingdon:
> I have made no attempt to avoid ax-setind in iset.mm. This is largely based on the sources I have to work with (for example, https://plato.stanford.edu/entries/nonwellfounded-set-theory/ does not seem to mention constructive set theory at all, and likewise the constructive sources don't seem to discuss non-well-foundedness)

Don't know if anyone is browsing old pull request comments, but in case they are I'll add a pointer to current thinking:

* The entry for df-fr at http://us.metamath.org/ileuni/mmil.html describes our current thinking on what life without ax-setind would be like
* There is a whole chapter on non-wellfounded constructive set theory in Chapter 20 of Aczel, Peter, and Rathjen, Michael, "Constructive Set Theory", 2018, http://www1.maths.leeds.ac.uk/~rathjen/book.pdf (I'll add a mention of this to the df-fr entry of mmil.html in a future pull request).

digama0:
Regarding ax-setind, which is the equivalent to set.mm's ax-reg, my recollection is that after a pass to eliminate unnecessary uses, it is used almost nowhere. It is used to derive the equivalence of ax-ac2 and ax-ac, but ax-ac isn't used anywhere else so it doesn't matter so much. Other than that, most things that need to be well founded can be explicitly assumed to be so, and ordinals are well founded by construction so all "normal" sets based on them are also well founded, like the construction of real numbers. It is only when you have statements of the form "all sets have such and such property" where the property requires well foundedness that ax-reg comes up, but you can always replace such a statement with "all well-founded sets have such and such property" and set.mm has a `WF` class for this purpose.

What theorems require ax-setind? If it is easier for you, there isn't any issue with assuming it freely, but it may not be a large gap from that to eliminating it entirely if the uses are all nonessential.

jkingdon:
> What theorems require ax-setind? If it is easier for you, there isn't any issue with assuming it freely, but it may not be a large gap from that to eliminating it entirely if the uses are all nonessential.

At the moment, lots of stuff uses it:

```
MM> show usage ax-setind/recursive
Statement "ax-setind" directly or indirectly affects the proofs of 3005
statements:
```

I suppose a lot of those are "nonessential" but I haven't taken a close look :shrug: . I don't really have much of an opinion about whether to remove the nonessential usages or not, I just figured I should describe the situation at http://us2.metamath.org/ileuni/mmil.html (search for "df-fr").

digama0:
Here are the theorems that directly reference a statement that depends on ax-setind, that do not themselves need to depend on ax-setind: (This is a complete list, since it is a cut through the dependency graph rooted at ax-setind.)

```
ordirr <- elirr
  tfrlemisucfn <- elirrv
  tfrlemisucaccv <- elirrv
  ltsopi <- elirrv
  dtruex <- elirr
  ordsoexmid <- elirr
  onnmin <- elirr
  nntri1 <- ssnel
  onpsssuc <- elirr
  ordtri2or2exmid <- elirr
  nntri2 <- elirr, en2lp
  nntri3 <- elirr
  nndceq <- elirr
  nndcel <- elirr, en2lp
  phpelm <- elirr
  fiunsnnn <- elirr
suc11 <- suc11g
  peano4 <- suc11g
  frecsuclem3 <- suc11g
ordsuc <- en2lp
2pwuninelg <- en2lp
tfi <- ax-setind
```
The unindented lines are "root causes", and the indented ones can be fixed by referencing the parent theorem instead. The main cause of unnecessary uses is elirr, which can be proven for ordinals without ax-reg, and most applications only need ordirr instead of elirr. Similarly, en2lp holds for ordinals and that is how you prove suc11 and ordsuc. I'm not entirely sure, but I believe that the set.mm proof of 2pwuninel using cantor's theorem should still apply to prove 2pwuninelg instead of the current regularity proof. Finally tfi is a theorem about ordinals so you can just use that set induction holds for ordinals by definition.

PR #1048, opened 2019-08-11
mathbox update

wlammen:


wlammen:
>> The command "scripts/verify iset.mm" exited with 1.

I did not modify iset.mm.  It must have been in a broken state before.

nmegill:
Last night I uploaded a new version of metamath.exe (0.178) that fixes detection of too-long lines.  I thought we had fixed iset.mm but one line must have slipped through.  Anyway, 'verify markup' with 0.178 will detect this problem from now on, so this shouldn't be a problem going forward.

jkingdon:
@nmegill Thanks for fixing that line - because it is very new in iset.mm I suspect that's why it slipped through our earlier pass looking for long lines. And I agree that having `verify markup` check should handle this case in the future. I just upgraded the machine I develop on to Version 0.178 10-Aug-2019 and it does indeed detect that long line for me.

PR #1049, opened 2019-08-11
add finite recursion to iset.mm

jkingdon:
This defines finite recursion as suggested in #847 . It includes two of the three important finite recursion theorems (frfnom and fr0g in set.mm, renamed here due to different notation).

The third one, frsuc, will presumably (1) be harder, and (2) be possible.

Oh, and renaming `om` to `_om` is to bring iset.mm closer to set.mm.


jkingdon:
I changed `.travis.yml` so that it checks mmil.raw.html against iset.mm and has a list of .raw.html files (all the rest, currently) which get checked against set.mm. Not attached to this solution if @david-a-wheeler or @nmegill want to solve it another way.

nmegill:
@david-a-wheeler wrote .travis.yml, and if he agrees with your change I will agree also.

david-a-wheeler:
The .travis.yml change looks good to me.

PR #1055, opened 2019-08-18
shorten mo3; update mathbox

wlammen:


wlammen:
axregnd: an occasional shortening when investigating the usage of sps, nothing spectacular.

wlammen:
I suggest to include wl-aleximd to the main part.  I know of at least seven (the minimum count) occasions for replacing a sps eximd sequences, two other alimi exim syl sequences in the main part alone. There is one more in my mathbox.

benjub:
I second @wlammen's suggestion of adding wl-aleximd to the main part.  I
would label it "aleximi" since it is an inference and not a deduction (even
though it is "deduction-like").
Since Wolf hints at a shorter proof using more axioms, maybe it can be
added as aleximiALT, with both "discourage" tags and cross-references in
comments.

PR #1057, opened 2019-08-19
add natural number addition theorems to iset.mm

jkingdon:
Turns out a number of theorems in natural number arithmetic (addition, mostly) are ready to go. Many of them were unlocked by oasuc.


david-a-wheeler:
Very impressive! Looks like some streams have finally broken through.

jkingdon:
Thanks for the kind words @david-a-wheeler . As usual @digama0 gets a lot of the credit (in this case for `oasuc` which had been stuck).

What I'm most worried about is the order stuff like http://us.metamath.org/mpeuni/nnaord.html . Hopefully we'll figure that out (at least for natural numbers where more propositions are decidable, general ordinals might be a different matter).

PR #1059, opened 2019-08-22
Diagonal matrices and two new theorems for propositional logic

avekens:
* two new theorems for propositional logic, according to discussion in Google-Group:https://groups.google.com/d/topic/metamath/f-L91-1jI24/discussion
* theorems for diagonal matrices (AV's Mathbox)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -5983,6 +5983,8 @@
 "eu3OLD" is used by "mo2OLD".
 "eumo0OLD" is used by "eu2OLD".
 "eumo0OLD" is used by "mo2OLD".
+"ex-natded5.3i" is used by "stoweidlem34".
+"ex-natded9.20" is used by "nobnddown".
 "exatleN" is used by "cdlema2N".
 "exidu1" is used by "cmpidelt".
 "exidu1" is used by "exidresid".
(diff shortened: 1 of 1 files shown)


wlammen:
@avekens  Hint: Line 15547 contains an unusual comment.

benjub:
Thanks Alexander.  Your proof of jccil is actually shorter when counting
the number of (essential and non-essential) steps, which is the length of
the block of capital letters.
So I would remove the commented out proof and would change the second
sentend of the comment to: "One can also prove this theorem using ~ syl and
~jca ." for the sake of concision.  You can also remove "/BJ" to have a
more standard form of credit.

I think the more standard term for "cell" of a matrix is "entry". For
instance, the comment of dmatelnd could be "An extradiagonal entry of a
diagonal matrix is equal to zero." Maybe this occurs in other theorems too.

dmatcrng: what do you mean by "over a commutative ring" ? This looks
extraneous.

A matrix which is a (scalar) multiple of the identity (equivalently, a
diagonal matrix with identical diagonal entries) is called a scalar matrix.
So for d1matid, you could write "The identity matrix is a scalar matrix
(that is, a diagonal matrix with identical diagonal entries)." (You forgot
the "diagonal" after "identical".)  For the following theorems, you can
then simply use "scalar matrix", and maybe relabel d1mat* to scalmat* or
scmat* or smat* ?

avekens:
Thank you very much for the review. I revised my contribution accordingly.

PR #1061, opened 2019-08-24
shorten sb8eu & mo2v

wlammen:


wlammen:
@nmegill I still wonder what to do with wl-aleximi. It could be applied 10+ times in the main part, each time saving roughly two proof lines.  You once told me that seven such occurrences are sufficient for an inclusion into the main part.  Is there anything speaking against this?

benjub:
I agree with @wlammen that wl-aleximi could be moved to the main part (e.g. right after eximii) and used to minimize enough theorems that it justifies its addition.

nmegill:
I'm fine with moving  up wl-aleximi.

I mentioned 7 uses as a rough guide, but a trivial variant of an existing theorem may take a dozen uses or more before it pays for itself in set.mm size reduction.  There is no easy way to determine that except by doing it.

Another subjective criterion is whether it fills in a missing gap that one might expect to have available, and it does seem to fit in as the existential version of al2imi (which has about 30 uses).

nmegill:
BTW I think know this but we usually don't acknowledge proof shortenings achieved by 'minimize' e.g. for the individual proofs that wl-aleximi will shorten.

PR #1064, opened 2019-08-25
 Add theorems on mirror points, up to the crib lemma.

tirix:


nmegill:
Looks good.

There doesn't seem to be anything called "crib lemma" or "Crib's Lemma" in an English Google search.  The work of Micheal Beeson (in English) just uses "the Krippenlemma" or "Gupta's Krippenlemma".  Perhaps call it one of these?

(BTW Micheal Beeson has expressed an interest in Metamath and says he might try to learn about it later this year.)

tirix:
I've updated to simply "Krippenlemma", but added the reference to Gupta's paper in the comment. I'll maybe add the reference in the bibliography if I use it elsewhere.

I'm glad Michael Beeson is interested in this work! I'd be happy to explain my approach.

PR #1065, opened 2019-08-25
Mostly: introduce aleximi and minimize with it

wlammen:
I did not touch mathboxes (except my own) during the minimizing process.

Statistics: in main 13 theorems are affected by 21 individual replacements. Each application saves roughly 15 characters in compressed proof size, along with two proof lines, whereas aleximi adds 26 characters and 4 proof lines to the pool. Mostly minimize_with in the metamath program did the work. On 4 occasions proofs were manually reworked, most notably in 19.30, where the proof structure changed considerably.

@benjub A shorter proof of aleximi uses nfa1, sps and eximd, but depends on much more axioms. This inhibits its usage in early theorems like exbi, for example.  As a rule: fewer dependencies usually enforce larger proofs.  I encountered several instances of this rule when reducing the dependency on ax-11. I marked such theorems with (Revised by...) tags, but generally did not bother to keep the old proof as an alternative.  If you feel this should be done in the present case, please introduce it on your own.


nmegill:
Actually our policy is to allow minimization (and even manual proof shortening) in other people's mathboxes, unless it is explicitly stated "(Proof modification is discouraged.)".  So feel free to do so if you wish.

Issue #809, opened 2019-04-06
Definition for the support of a function


avekens:
While working for Linear Algebra, I often had/have to use functions with finite support. Currently, this is expressed like ( `' F " ( _V \ { Z } ) ). Although I used it a lot of times, I am still always puzzled by this cryptic expression. Therefore, I would like to propose a special definition for the support of  functions, as I created already in my Mathbox:

df-supp $a |- supp = ( x e. _V , z e. _V |-> { i e. dom x | ( x ` i ) =/= z } ) $.

By this, ( F supp Z ) can be used instead of ( `' F " ( _V \ { Z } ) ) (under certain, but usually given conditions), which is shorter and more concise. The equivalence of both terms is shown in ~ suppimacnv.

If we can agree on this definition, I would suggest to place it in section "2.3.12  Functions", and to replace all occurences of terms like ( `' F " ( _V \ { Z } ) )  by ( F supp Z ) in the following theorems of (the main part of) set.mm. If I counted correctly, there are about 80 such terms to be replaced.

Would this be OK?

benjub:
Maybe you could replace ( x ` i ) =/= z with ( x " i ) =/= { z } and then prove ~ suppimacnv with the hypothesis relaxed from ` Fun F ` to ` Rel F ` ?

In the comment of df-supp, there is an extra closing parenthesis in the link to the Wikipedia article.

david-a-wheeler:
I like the idea of replacing a cryptic sequence with a clear definition. So I like this one.

avekens:
Good idea, Benoit! Although it was not easy, I could prove ~ suppimacnv using your version of the definition, even without additional hypotheses (except assuring sethood):
      df-supp $a |- supp = ( x e. _V , z e. _V |-> { i e. dom x | ( x " { i } ) =/= { z } } ) $.
      suppimacnv $p |- ( ( R e. V /\ Z e. W ) -> ( R supp Z ) = ( `' R " ( _V \ { Z } ) ) ) $=
I'll update my mathbox accordingly soon.

benjub:
@avekens  Nice !  There's a typo I'll let you fix in a future PR, since it is in your mathbox: to fix the link in the comment, you need to add a space:
~ https://en.wikipedia.org/wiki/Support_(mathematics))
to
~ https://en.wikipedia.org/wiki/Support_(mathematics) )

avekens:
[...]

PR #1076, opened 2019-09-02
Revised Euler Partitions theorems which were using obsolete theorems on the support of functions.

tirix:
This is related with issue #809 , to generalise the use of ` supp `.


nmegill:
This is fine with me.

PR #1081, opened 2019-09-07
Remove

benjub:
Commit messages say it all (I also moved axc16ALT, axc16ALT2 and axc16gALT earlier to a more fitting place).
I think that the relabeling ax16gALT --> axc16gALT need not be documented in the preamble since it is never used.
If @nmegill agrees, I suggest removing axc16ALT2, which does not add anything, it seems, to axc16 and axc16ALT.
I also suggest removing what looks like old non-visible comments about dedekind.mm: lines 111373--111376, 115919--115929, 385273--385274.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1360,7 +1360,6 @@
 "axacndlem4" is used by "axacndlem5".
 "axacndlem5" is used by "axacnd".
 "axaddf" is used by "axaddcl".
-"axc16ALT2" is used by "ax16gALT".
 "axc4i-o" is used by "aev-o".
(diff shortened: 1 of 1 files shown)


nmegill:
> I think that the relabeling ax16gALT --> axc16gALT need not be documented in the preamble since it is never used.

Fine.

> If @nmegill agrees, I suggest removing axc16ALT2, which does not add anything, it seems, to axc16 and axc16ALT.

I agree.

> I also suggest removing what looks like old non-visible comments about dedekind.mm: lines 111373--111376, 115919--115929, 385273--385274.

OK.

PR #1082, opened 2019-09-07
Remove

benjub:
I did the modifications agreed in PR #1081.

I standardized more comments of ALT proofs, i.e., the comment of xxxALT begins with "Alternate proof of ~ xxx ." and then explains the differences. Some work remains to be done in a future PR.

By the way, shouldn't elunirn and elunirnALT be swapped? They use the same axioms except that the ALT proof does not require ax-pow (nor df-iun df-res df-ima) contrary to the (currently) main proof. If you agree, I would do this in a future PR.

Note: some couples (xxx/xxxALT) seem to differ by the requirement or not of ax-8, but this is an artifact which will disappear when I modify df-clel to bj-df-clel.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13958,7 +13958,6 @@ New usage of "axc11n-16" is discouraged (0 uses).
 New usage of "axc11next" is discouraged (0 uses).
 New usage of "axc11nfromc11" is discouraged (0 uses).
 New usage of "axc16ALT" is discouraged (0 uses).
-New usage of "axc16ALT2" is discouraged (0 uses).
 New usage of "axc16b" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
@benjub I will be traveling and will address your questions in a day or so.

nmegill:
> By the way, shouldn't elunirn and elunirnALT be swapped? They use the same axioms except that the ALT proof does not require ax-pow (nor df-iun df-res df-ima) contrary to the (currently) main proof. If you agree, I would do this in a future PR.

elunirnALT has a significantly longer proof, but if that's what it takes to avoid ax-pow then so be it.  Go ahead and swap them if you want.

PR #1086, opened 2019-09-08
shorten euequ1

wlammen:
This shortening saves just a single character, but the proof does not use more from the existential quantifier section than the definitions.  That is why it can be placed near the beginning.  Other theorems like cbveu, cbvmo have been moved up too, because of their fundamental value.

I have expected euequ1 much earlier, because, from my point of view, df-eu checks arbitrary expressions ph against a simple equation.  For this reason alone, I think, it is worth to see how an equation itself behaves under definition df-eu.

There is more to it, as can be seen from the following.  Ever since I started proving something in the prepositional section, I often rely on a geometrical interpretation as a source of inspiration.  I envision the objects in the universe being mapped to, say, the real numbers, so I can line them up on a ray, and I view a logical expression ph(x, y, z) as a function in three dimensions, whose result (graph) determines, how a three-dimensional point is colored.  White for false, black for true, so you finally end up, at least symbolically, with Venn like diagrams. If you look at a simple equation x = u with some fixed u (I know, I know ...) then this equation, being free of y and z, describes a plane (more general a hyperplane in some n-dimensional universe space) in the three-dimensional domain of ph.  Adding more equations in other variables yields intersections of the planes, finally ending up in a single point.  This process can be viewed as the driving motivation for expressions like E. u E. v E. w A. x A. y A. z ( ph <-> (x=u /\ y = v /\ z = w )). Unfortunately, a theorem like 2mo, and its proof, does not really help understanding this.  I would rather see expressions like E! x E. y E. z ph /\ E! y E. x E. z ph /\ E! z E. x E. y ph as a representation of multi-dimensional uniqueness. In addition, I expect these to hint how to proceed to more dimensions, without being stuck with unwieldy intermediate expressions.

This is the point where I miss not having an education in logic, nor the time or opportunity to access and read literature.  The only thing I do is proving hundreds of theorems on my own, so I finally get an idea of what is going on.  Maybe I am wrong here.  Please let me know.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15620,6 +15620,7 @@ New usage of "eu2OLD" is discouraged (0 uses).
 New usage of "eu3OLD" is discouraged (2 uses).
 New usage of "eu5OLD" is discouraged (0 uses).
 New usage of "euanOLD" is discouraged (0 uses).
+New usage of "euequ1OLD" is discouraged (0 uses).
 New usage of "euexOLD" is discouraged (0 uses).
 New usage of "eufOLD" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


digama0:
To me, the expression E. u E. v E. w A. x A. y A. z ( ph <-> (x=u /\ y = v /\ z = w )) has the advantage that it's obviously correct, whereas the various variations on E! x E. y E. z ph look a lot less clear to me, even if they happen to work out right. Once we have pairs we can express this with a single E!, but I don't see any easy compositional way to build multidimensional exists unique.

PR #1089, opened 2019-09-09
Bust verifier -- do not accept

wlammen:
This is a test to check whether travis accepts certain definition pattern. This pull request will eventually be deleted. DO NOT ACCEPT

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15620,6 +15620,7 @@ New usage of "eu2OLD" is discouraged (0 uses).
 New usage of "eu3OLD" is discouraged (2 uses).
 New usage of "eu5OLD" is discouraged (0 uses).
 New usage of "euanOLD" is discouraged (0 uses).
+New usage of "euequ1OLD" is discouraged (0 uses).
 New usage of "euexOLD" is discouraged (0 uses).
 New usage of "eufOLD" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
@@ -18809,6 +18810,7 @@ Proof modification of "eu2OLD" is discouraged (118 steps).
 Proof modification of "eu3OLD" is discouraged (45 steps).
 Proof modification of "eu5OLD" is discouraged (38 steps).
 Proof modification of "euanOLD" is discouraged (106 steps).
+Proof modification of "euequ1OLD" is discouraged (42 steps).
 Proof modification of "euexOLD" is discouraged (32 steps).
 Proof modification of "eufOLD" is discouraged (73 steps).
 Proof modification of "eujustALT" is discouraged (188 steps).


wlammen:
Unfortunately, it is possible to fool both the Metamath proof verifier and Travis' checks to accept a carefully crafted definition, that ultimately allows to arrive at a contradiction. There are still checks failing, but they are related to missing HTML instructions.  I did not bother adding these for this test, that otherwise is accepted.
I will keep this pull request as long open, as discussion demands. It will be eventually be deleted.

digama0:
@wlammen Is the problem fixed?

wlammen:
@digama0 The problem has not been addressed by me, so if you haven't done anything, I think it is still unresolved.  I closed this pull request, because I announced to do so, and its sole purpose was to provide evidence, that messed up definitions can pass undetected.

I actually think, that improving on Travis checks is not sufficient.  Sure, it protects the main repository and hence covers some 90% of Metamath applications.  But the Metamath book promises something different: If you let a proof verifier check your (possibly specific/extended) source, you can rest assured to the extent you trust your computer, that your proofs are correct.  In the light of the presented case this is simply not correct.  In addition, the book voices at several occasions the rigour by which mathematics is developed from scratch on. Since in a definition like df-eu, for example, a proof is missing (but possible), that the dummy variable y is not free in the expression to the right of the bi-conditional (a one liner, nfe1, but missing, though), I feel this is just to 99.99% true.  (The process of checking the soundness of definitions must further be able to filter out set variables appearing on just one side, which involves some text processing outside of logic -- simple enough to believe an automaton succeeds on this, but it is worth a mention, though).

Maybe, I will copy this text to the google group later, currently I am busy at work.

benjub:
To be fair, as far as I recall, the Metamath book does not promise to check
definitions.  I think it explicitly says that $a-statements are the user's
responsibility, and the Metamath program only guarantees that the proofs
from these $a-statements are correct.  Apart from that, I agree with
@wlammen that it would be nice to have mmj2 (or another tool) check for the
appearance of free variables in the definiendum that are not free in the
definiens.
Concerning df-eu, there is right before it the required theorem eujust.

benjub:
Erratum: swap "definiendum" and "definiens" in my previous post, of course.

digama0:
In that case let's either keep this PR open or open an issue until it's fixed, lest we forget.

digama0:
I agree with the thrust of @wlammen's critique, that it is a problem that there is soundness critical code outside the MM verifier (this is one of the main reasons why MM0 differs from MM), but as @benjub says it's not technically a problem with the specification, which checks exactly what it claims.

The problem is that one cannot simultaneously claim that MM checks everything down to the axioms, and all of set theory can be developed in MM from a small enumerated list of ~20 axioms. Metamath in actual practice violates the second claim, because it is actually a proof relative to a few thousand axioms, all of which are soundness critical, and MM + defcheck violates the first claim, since now the defchecker has to be trusted in addition to the verifier. (MM0 folds the defchecker into the verifier, at the cost of additional complexity.)

One social problem with the defchecker is that it has no formal spec and no reimplementations, unlike the main verifier which has several dozen implementations, which gives it a sort of "second class" status. How do people feel about hooking up the MM -> MM0 translation and the MM0 verifier applied to the result to Travis? It is slightly heuristic, but it would provide an alternative defchecker and verifier, and since the algorithm is different from the canonical MM algorithm, it may notice bugs that are correlated across MM verifiers.

david-a-wheeler:
> How do people feel about hooking up the MM -> MM0 translation and the MM0 verifier applied to the result to Travis? It is slightly heuristic, but it would provide an alternative defchecker and verifier, and since the algorithm is different from the canonical MM algorithm, it may notice bugs that are correlated across MM verifiers.

I'm okay with adding it for now, since there are other heuristic checks (like formatting ones) that are already applied.

In the *longer* term, I'd like for there to be a formal definition for definitions (!) - at least within set.mm - so that databases that are "like" set.mm can be checked by multiple verifiers.

digama0:
To clarify, the heuristic is not in the "definition of definitions"; I am sure that if it passes the translation and verification in MM0 then it is a good definition, unless there is a bug in MM0 theory or implementation (and both of these will be formally verified in the near to mid term). The heuristic is that there are several things that are necessary for definitional checking that are not marked in the database, and I am inferring stuff when it can reasonably be done and using $j annotations for the rest (which exist in set.mm but not necessarily in MM conformant offshoot libraries). I think we should get more particular about which annotations are required for correctness.

The areas where the translation is heuristic includes things like inferring binder structure of term constructors, which is not marked in stock MM style. Both `wal` and `wel` look like binary relations from the point of view of MM, but MM0 wants to know that `wal` binds occurrences of the first argument in the second. There are $j annotations for this, and they are mandatory for primitive term constructors and inferred for definitions.

Grammar unambiguity is not checked, but MM0 uses a general (backtracking) CFG parser and works with the first valid parse it finds, so any proof that introduces and exploits an ambiguity will not pass the translation stage.

wlammen:
@benjub  eujust is not enough. You need explicitely F/ z E. z ...

digama0:
By the way, I tried to replicate this locally and mmj2 (master) worked fine. Can anyone else replicate? As far as I can tell the only change between the downloaded mmj2 definitionCheck.js and the master branch is the set/setvar change, which is currently being patched in via a sed script in travis, so I don't see any reason for it to be different.

benjub:
@wlammen Can you elaborate?  I would think it's the opposite: we are content with "F/ x ph(x)" because we know that it implies that "ph(x) <-> ph(y)", but this is this latter formula that we really need to prove eliminability.  Can @digama0 or @nmegill shed some light?

wlammen:
Being able to exchange the dummy variable does not tell anything formally about the freedom of it. In my eyes eujust suggests that the dummy variable will not play a role during unfolding the definition. Such a variable is not free by all one can expect, but the same holds for 1 + 1 = 2.  After having looked at my fingers, I am instantly convinced, this is true.  In Metamath you need a formal proof for that.
Obviously, I do not have a theoretical background, so this is it from my side.

benjub:
@digama0 I tried it locally and my mmj2 (v2.5.1) did not complain at the new definition... But then I saw that the RunParms.txt file had not definition check, so I added the line
> SetMMDefinitionsCheckWithExclusions,ax-*,df-bi,df-clab,df-cleq,df-clel
and then all definitions generated error messages like
> I-PA-0205 Axiom df-bust-verifier has failed the definitional soundness check. Non-set dummy variable found, and no justification theorem is available.
This is probably because of the change var / setvar...
I can't do more; the installation procedures of mmj2 are not very easy, unfortunately.

digama0:
The only relevant file is `mmj2jar/macros/definitionCheck.js`; you can just download it from the [github master branch](https://github.com/digama0/mmj2/blob/master/mmj2jar/macros/definitionCheck.js) and replace your local copy without recompiling anything (this is why the JS macro system exists, so that users can edit these files themselves if they want to extend mmj2).

digama0:
> How do people feel about hooking up the MM -> MM0 translation and the MM0 verifier applied to the result to Travis? It is slightly heuristic, but it would provide an alternative defchecker and verifier, and since the algorithm is different from the canonical MM algorithm, it may notice bugs that are correlated across MM verifiers.

Actually, this won't work at present. It would function as an alternative MM verifier, but the MM -> MM0 translation does not currently turn definitional axioms into `def`s in MM0 (they remain axioms as in MM), so it would not do any defchecking just like every other verifier. (The groundwork is laid for this to work in the future, but it isn't being done yet.)

> Being able to exchange the dummy variable does not tell anything formally about the freedom of it.

Strictly speaking, the fact that we need for conservativity is that you can change the names of bound variables. This follows from the appropriate `F/` theorem for the target bound variable, so you need one of these theorems for each dummy variable in the term. The problem with baking this into a MM verifier is that MM doesn't even know what `->` means, let alone binders, quantifiers, and the not-free predicate, so it's not obvious what this theorem has to do with the claimed conservativity result (which depends on the logic acting essentially like a first order logic with a sort for variables and a sort for wffs etc). I have written a number of `$j` annotations that should be sufficient in principle to get from the primitives to the conservativity meta-proof, which involves essentially an algorithm for proving theorems of the form `P(x1,..,xk) <-> P(y1,...,yk)` for lots of predicates `P` with the marked dummy variables (and possibly additional free and bound variables), but currently there is no implementation that actually uses this information, and mmj2 uses a general purpose heuristic prover to prove these facts rather than any principled system (so it can potentially fail spuriously if it can't prove the not-free theorems).

benjub:
@digama0 I downloaded mmj2jar/macros/definitionCheck.js from the github master branch as you indicated, but it still has "set" in the code, so I replaced the five occurrences with "setvar", and then it passed:
> I-UT-0015 **** Processing RunParmFile Command #4 = SetMMDefinitionsCheckWithExclusions,ax-*,df-bi,df-clab,df-cleq,df-clel

By which I mean: the definition df-bust-verifier caused no problem, unlike in your case (my version is v2.5.1).

By the way, do you know why Travis excludes df-sbc from the definition check? (it says: RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel,df-sbc)

digama0:
Oops, I should have referred you to the [develop](https://github.com/digama0/mmj2/blob/develop/mmj2jar/macros/definitionCheck.js) branch version, which is more recent (and includes the set/setvar fix). 

Oh! You said you are using v2.5.1 but the latest stable version is v2.5.2, and there is even a commit in the interim suspiciously labeled "fix definition bug", which may be related. (I am working on releasing v2.5.3 now with these bugfixes.) But I don't see how it could be, because it only affects error I-PA-0204, and df-bust-verifier should be reporting error I-PA-0206. If you downloaded the master branch version and changed "set" to "setvar" everywhere, it should be identical to the develop version.

benjub:
Wow, the file on the develop branch is *identical* to the one I modified !  I'm going to install v2.5.2 and I'll let you know.  As asked above: do you know why Travis excludes df-sbc from the definition check?

digama0:
df-sbc is excluded because it reuses a constant that has already been introduced, namely `[ A / x ] ph`. This was "punned" earlier in order to define `[ x / y ] ph` in df-sb without having a grammatical ambiguity. As a result, neither df-sb nor df-sbc follows the rules of introduction of new constants, and the non-logical theorem `[ x / y ] ph <-> x e. { y | ph }` (with df-sb on the left) is derivable purely by using the "definition" df-sbc.

benjub:
I downloaded the master branch, didn't have to compile anything, and obtained v2.5.3!  And it correctly rejects the spurious definition:
```
I-UT-0015 **** Processing RunParmFile Command #4 = SetMMDefinitionsCheckWithExclusions,ax-*,df-bi,df-clab,df-cleq,df-clel
I-PA-0206 Axiom df-bust-verifier has failed the definitional soundness check. Dummy variables [y] are possibly free in the definiendum, and no justification is available.
```

Thanks for the df-sbc explanation.  I guess it's therefore excluded for when a future algorithm is able to reject such a definition, because for the moment, it passes. For the record:
```
I-UT-0015 **** Processing RunParmFile Command #4 = SetMMDefinitionsCheckWithExclusions,ax-*
I-PA-0201 Axiom df-bi has failed the definitional soundness check. The root symbol is not = or <->.
I-PA-0202 Axiom df-clab has failed the definitional soundness check. The previous axiom ax-c14 uses the symbol being defined in this axiom.
I-PA-0202 Axiom df-cleq has failed the definitional soundness check. The previous axiom ax-8d uses the symbol being defined in this axiom.
I-PA-0202 Axiom df-clel has failed the definitional soundness check. The previous axiom df-clab uses the symbol being defined in this axiom.
I-PA-0206 Axiom df-bust-verifier has failed the definitional soundness check. Dummy variables [y] are possibly free in the definiendum, and no justification is available.
```

digama0:
mmj2.zip has been updated with v2.5.3 on us2, so Travis should be picking it up now. I can't seem to rerun this PR, though, possibly because it's been closed and reopened?

nmegill:
> df-sbc is excluded because it reuses a constant that has already been introduced, namely `[ A / x ] ph`.

This used to be the case but I changed it to use new constants in Jan 2017.  df-sb is '[ y / x ] ph' and df-sbc is now '[. A / x ]. ph'.  So df-sbc can and should be removed from the exception list.

wlammen:
I started a rerun...

digama0:
Hm, it picked up v2.5.1. I think it's because it's reading the us server, not us2, and it might be a few days before the change propagates.

nmegill:
> Hm, it picked up v2.5.1. I think it's because it's reading the us server, not us2, and it might be a few days before the change propagates.

I just transferred it directly to the us server, so the new version should be there if I didn't make a mistake.

wlammen:
no Travis checks visible. I cannot attend this any longer, have to get up early next morning.

digama0:
I'm not sure why the last two commits didn't trigger a travis rebuild, but I restarted the last build and it read v2.5.3 correctly this time and the check worked, so I think this issue is fixed.

PR #1096, opened 2019-09-16
Construction of outer measures

tirix:


nmegill:
supssd, xrge0infss, and xrge0infssd look like they can be moved the main set.mm at your convenience.  I think xrge0infss might be better named xrge0infex.   xrge0infssd could be named just xrge0infss (because the deduction form for advanced theorems is becoming a frequent convention for easier use, and typically we omit the "d" to shorten the name because there is no purpose in also having a closed theorem form; whether a theorem is a deduction form or not is obvious by inspection, which needs to be done anyway due to the complexity of most advanced theorems).

tirix:
Thanks! 
I'll move them with my next PR. I also noticed a mistake in the statement for the outer measure construction statements, which I will need to fix.
Right now I'm wondering whether it is worth defining a class of its own for outer measures or not.

PR #1098, opened 2019-09-19
Misc

benjub:
* update scripts/min.cmd and scripts/minimize to work with the new 'minimize_with' behavior
* move theorems from the 'AS subsection' to either main part or AS's mathbox, as agreed with @nmegill (see below)
* some relabelings, documented at the top of set.mm ("Recent label changes"), and for exp3a --> expd and imp3a --> impd, also documented in the "Important recent changes" paragraph.
----
Details about the theorems in the former section "1.2.13  Auxiliary theorems for Alan Sare's virtual deduction tool":
* exbir : move to AS's mathbox; no discouragement tag (in its comment, "head" --> "consequent", "conditional" --> "implication").
* 3impexp, 3impexpbicom, 3impexpbicomi: move to AS's mathbox (since using "triple conjunction" is not favored), no discouragement tag.
* ancomsimp: keep (since several uses and is the closed form of a theorem in the main part) and relabel ancomst.
* exp3acom3r : keep (since many uses) and relabel "expdcom" and add to comment "Commuted form of ~expd."
* exp3acom23 : keep (since many uses) and relabel expcomd and add to comment "Deduction form of ~expcom."
* simplbi2comg : keep (as closed form of a theorem in the main part) and relabel simplbi2comt; no discouragement tag.
* simplbi2com : keep (since many uses).
* ee02 : move to AS's mathbox (too artificial to be in main part).
* Comment of simplbi2: remove "Automatically..." in order to avoid references to mathboxes from the main part.  Add to comment of simplbi2VD: "The proof of ~simplbi2 was automatically derived from it."
* Other relabeling suggestions:
    * exp3a --> expd (and add to comment "Deduction form of ~ex.")
    * imp3a --> impd (and add to comment "Deduction form of ~imp.")
    * con3and --> con3dimp (since it is not a deduction itself, but simply is ~con3d followed by ~imp)

PR #1099, opened 2019-09-20
Misc

benjub:
* Relabeling as agreed in PR #1098 
* Typos
* Mathbox: "if" operator for propositions


nmegill:
@avekens @tirix (and anyone else) Let me know whether the previous imp3a, exp3a -> impd, expd label change causes problems for your work in progress, before I merge this, so we don't get too far ahead in case we need to revert.

avekens:
I synchronized my fork and updated my modifications - no problems with  imp3a, exp3a -> impd, expd  (see PR #1100).

tirix:
Hi Norm, 
I usually try to commit as early as possible, so the amount of proofs in progress is not too high. In any case I'll adapt, you can go ahead with the change. If it's only a label change it will also not be hard to catch up.

PR #1101, opened 2019-09-21
shorten 2eu6

wlammen:
- 2sb6rflem2 has now a usage outside of the proof of 2sb6rf. I think this is a reason to promote it to a usual theorem, perhaps named as 2ax6e.
- 2eu6 and mo2v have a subproof of E. x ph -> A. x ( ph -> x = y ) -> A. x ( x= y -> ph ) in common. I suggest to make it a lemma used by both theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13644,6 +13644,7 @@ New usage of "2atm2atN" is discouraged (0 uses).
 New usage of "2atnelvolN" is discouraged (0 uses).
 New usage of "2bornot2b" is discouraged (0 uses).
 New usage of "2eu4OLD" is discouraged (0 uses).
+New usage of "2eu6OLD" is discouraged (0 uses).
 New usage of "2exsbOLD" is discouraged (0 uses).
 New usage of "2llnjN" is discouraged (1 uses).
 New usage of "2llnjaN" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


benjub:
* I second @wlammen's proposal, and the label 2ax6e (or maybe ax6e2 ?).  By the way: 2sb6rflem1 (to be relabeled 2ax6elem or ax6e2lem ?) has a precursor, ax6e2nd by Alan Sare. It subsumes it since it has no dv conditions, but maybe credit to Alan Sare could be added to ax6e2lem.
* Do you mean, ( E. x ph -> ( A. x ( ph -> x = y ) -> A. x ( ph <-> x= y ) ) ) ?  I agree, but do not have a label suggestion.

wlammen:
No, I meant E. x ph -> ( A. x ( ph -> x = y ) -> A. x ( x= y -> ph ) ). There are 3 usages of this pattern, one in mo2v and two in 2eu6.  I don't know whether your extension provides an even better result

nmegill:
> * I second @wlammen's proposal, and the label 2ax6e

I'll go along with that.

> (or maybe ax6e2 ?).

Probably not - a 2 in front usually means double quantifier etc. e.g. 2al*, 2eu*, whereas 2 at the end usually means an alternate version or variation e.g. dfss2.

PR #1102, opened 2019-09-22
iset.mm: develop positive fractions through ltsonq

### LINKED ISSUE #1085: iset.mm: construct positive rationals via equivalence classes
The current definition, of choosing a canonical fraction (the one with the smallest possible denominator), as seen in http://us2.metamath.org/ileuni/df-nq.html (copied from set.mm) seems to rely on "every set of natural numbers has a minimal element" (which we don't have as shown at https://github.com/metamath/set.mm/pull/1084 ) or similar propositions. This is seen for example in the use of rexnal at http://us2.metamath.org/mpeuni/nqereu.html .

Using quotients or equivalence classes is what the HoTT book does (and even what pre-2013 set.mm did if I remember right).

I'm putting this into a github issue in case someone wants to weigh in or sees a solution which is not apparent to me.

jkingdon:
This is fleshing out most of the section on positive fractions (ratios of two positive integers). There are about 10 theorems left (beyond what is in this pull request) before that section is done and we are ready to try our hand at Dedekind cuts.

There's a lot of intuitionizing to do here, but it goes fairly easily given a few basic parameters:

1. We are using the equivalence class approach (as in set.mm from 2012)

2. Various theorems we rely on work slightly different (e.g. mulcompig instead of mulcompi, caov4d instead of caov4, etc)

2a. Some of that machinery is extended here (e.g. adding `caovimo` , `ecovidi`, `caovdilemd`, `ecoviass`, `ecovicom`, `oviec`, `ovi3`

3. Trichotomy is strong than `Or` (which in iset.mm is defined to mean weak linearity, as discussed at http://us2.metamath.org/ileuni/df-iso.html ). However, we can and do prove both for positive fractions (it will be when we get to reals that we will only get weak linearity).

4. I rearranged some of the reciprocal theorems and what is proved in terms of what. This is mostly to avoid reverse closure. Reverse closure as practiced in set.mm relies on excluded middle; I haven't been clear whether there is some form of reverse closure which would work without it, but at least to date we don't have any kind of reverse closure in iset.mm. Anyway, what is in this pull request for reciprocals does add a few (inocuous-seeming) conditions to some theorems but on the whole ends up proving the same things that set.mm does.

Fixes #1085


nmegill:
@jkingdon Travis reports that lines 58275 58276 58302 are 80 chars long (>79) and need to be shortened. - N

jkingdon:
Ah, I figured out why the long lines slipped through. I was accidentally using an older version of metamath on my machine than I thought I was.

Those responsible have been sacked (er, just kidding I'm not going to quit metamath over 80 versus 79 characters).

david-a-wheeler:
Those responsible for sacking those who were sacked, have been sacked. :-).

Seriously, it shows an advantage of CI.. even if your local system doesn't have sone check, it will be checked in CI.

I'm looking forward to the Dedekind cut material... good luck!!

jkingdon:
Thanks for the kind words.

How to do Dedekind cuts constructively seems to be pretty well laid out in the HoTT book (as described at http://us.metamath.org/ileuni/mmtheorems62.html#mm6102h where I also try to summarize the three main things which, based on that source anyway, will be different without excluded middle). Unlike some other parts of the HoTT book, this section is relatively readable if you just know set theory and not type theory.

So on general approach I feel more confident than I was with, say, ordinals. On the flip side, there probably will be a fair bit of work to formalize the informal treatment in the HoTT book and a moderately large divergence from set.mm based on the definitions which are changed (or in the case of apartness, added anew).

benjub:
I'm not following the development of iset.mm, but I see on the home page that you use IZF.  If that's the case, your use of "constructive" is debatable, I think. Have you thought of using CZF (concurrently)?

jkingdon:
> I'm not following the development of iset.mm, but I see on the home page that you use IZF. If that's the case, your use of "constructive" is debatable, I think. Have you thought of using CZF (concurrently)?

Hey thanks for noticing and asking about the project. I love to hear from people who are interested, whether following closely or not.

We mention CZF at http://us2.metamath.org:88/ileuni/mmtheorems21.html#mm2005h . I think it would be possible to formalize it in metamath, but would require a decent amount of machinery to define bounded formulas and provide mechanisms for proving that particular formulas are bounded.

Having done that, my guess (without knowing CZF vs IZF especially well) is that most of iset.mm would probably work without a lot of changes. Places where we use unrestricted abstraction to define a class could remain, but that proving that class exists (is a set) would not always be possible. Theorems such as http://us2.metamath.org/ileuni/ordtriexmid.html would only apply where `ph` is a bounded formula. Definitions such as http://us2.metamath.org/ileuni/df-enq.html would need to be restated slightly so that `E. z` is changed to `E. z e. N.`.

As for terminology, my preferred terminology most of the time is to say "with excluded middle" or "without excluded middle" rather than words like "classical", "intuitionistic", etc. But when I do use "constructive" I mean it as a big-tent word for a system which may or may not include countable choice or dependent choice, may be set theory or type theory, may be predicative or impredicative, etc. Here I'm following Bauer, whose definition is "Constructive mathematics is mathematics done without the law of excluded middle" (page 481 of http://dx.doi.org/10.1090/bull/1556 ).

Thanks again for the interest and I'd be glad to help if there is anything you've been thinking of contributing to iset.mm (I see for example you have done work in set.mm propositional logic, and most of what is in iset.mm in that area is based on set.mm as of 2015, so updating iset.mm and figuring out what does and does not need excluded middle is just one example of something that I probably won't get to soon, but would make iset.mm more complete).

benjub:
Thanks for your answer. I think you should put that second paragraph of http://us2.metamath.org:88/ileuni/mmtheorems21.html#mm2005h about CZF also on the home page after the IZF table, since I think it's one of the first things a newcomer will wonder.  (Remarks in passing: the paragraph before says "contained in" but I think "an element of" would be clearer, since "be contained" could be interpreted equally as "be included". The following comment has an unnecessary "n" in "unnnecessary".)

I'm not sure about your guess for the CZF development not differing too much from the IZF one.  In terms of proof-theoretic strengh, IIRC, IZF is closer to ZF than to CZF.  And conventional wisdom is that CZF captures the vague notion of "constructive mathematics" better (I'm no proof theorist or logician so I cannot tell).

Would it be really too hard to MMize CZF?  Maybe I'll give it a try in a new mathbox, along the lines:
$c wwf0 $.
$v ph0 ps0 $.
w0ph $f wwf0 ph0 $.
w0ps $f wwf0 ps0 $.
wwwf0 $a wwf ph0 $.
w0im $a wwf0 ( ph0 -> ps0 ) $.
...
w0al $a wwf0 A. x e. X ph0 $.  <--- maybe here I'll need a type "set" ?

The few things I did in the propositional calculus part of set.mm are merely fine tunings and small lemmas, so not worth the little trouble porting to iset.mm. As you know, I regret that the axiomatization I proposed to deal with the "big three" propositional calculi (minimal, intuitionistic, classical) simultaneously was not adopted, but let's not come back to it.

Even though I'm not contributing to iset.mm, I still try to be compatible with it, as you might remember with the proposed new definition of df-bj-nf discussed in the google group! (https://groups.google.com/d/msg/metamath/Ovxv2aXJOIM/6LUxnh9hBAAJ)

jkingdon:
> I think you should put that second paragraph

Good ideas all. I've submitted these as https://github.com/metamath/set.mm/pull/1108 - if you see anything else like that please let us know (via github issue, pull request, whatever).

As for how to formalize CZF, I suspect that a second `$c` for bounded formulas is likely to be pretty difficult. Not only would you need to duplicate most of the propositional and predicate logic sections for `wwf0`, you'd also need to develop a way to connect `wff` with `wwf0`. What I had in mind instead was a definition like `isBounded` and then `$a` axioms for `isBounded ph -> isBounded -. ph`, `isBounded ph -> isBounded E. x e. y ph` etc. Then restricted separation would look something like `isBounded φ → ∃y∀x(x ∈ y ↔ (x ∈ z ∧ φ))`. Of course the real way to find out would be to start trying and see where it gets one. I like the mathbox idea; you could look at what is downstream of ax-sep to figure out what theorems are worth reproving/modifying with only restricted separation. And that would be a good way to figure out how much of the strength of IZF we are actually using versus how much is just "sitting there" but unused in iset.mm.

And no worries about contributing - you know better than I what you are able to contribute. But if you do see something (e.g. those web page fixes in #1108 ), don't hold back.

benjub:
You're right, there are basically two ways to formalize bounded formulas: either a new type wwf0 or a predicate isBounded.  Maybe the second is more in the spirit of existing mm databases.  For the first way: note that my above axiom "wwwf0 $a wwf ph0 $." says that any bounded formula is a formula (similar to "every setvar is a class", ~cv in set.mm), therefore I can "reuse" already proved propcalc theorems for the bounded formulas. What are other advantages/drawbacks of either approach?

With either method (for the sake of concreteness, I'm using the predicate way), is the axiom
  $a |- ( isBounded ph -> isBounded A. x e. y ph ) $.
enough? Don't we need, rather,
  $a |- ( isBounded ph -> isBounded A. x e. X ph ) $.
where X is of a new type, the type "set" (i.e. a set term, not a set var)?

Remark: I noticed another thing in iset.mm: the comment of Diaconecu's theorem (acexmid) does not even mention R. Diaconescu (rediscovered independently by N. D. Goodman and J. Myhill).

jkingdon:
Ah, OK, `"wwwf0 $a wwf ph0 $." says that any bounded formula is a formula` might change things.

As for how to say "is a set" in an axiom, we usually just use a set variable. We can convert it to a class with a `A e. _V` condition using the `vtocl` family of theorems. At least, that's what I'd try to make work if it can be done (I don't know if there are any additional difficulties relating to restricted separation in particular).

> Remark: I noticed another thing in iset.mm: the comment of Diaconecu's theorem (acexmid) does not even mention R. Diaconescu (rediscovered independently by N. D. Goodman and J. Myhill).

Well and posed as an exercise by Bishop some years before Diaconescu published it.

My instinct is just to cite a few sources which explain it well rather than try to recap the history ourself (as we do for most theorems, where we cite a textbook rather than the first publication), but I'm not firmly opposed to some kind of expansion of the origin of this if we can figure out a suitably concise way to sum it up.

digama0:
I agree with @jkingdon that it would be better to have a predicate on wffs rather than a new sort, which would require having either a syntax ambiguity or a second version of each wff connective. But one has to be careful with the axioms about `isBounded`, because you can't view it as a function from bool to bool, so it cannot act simply as a function on wffs like `-.` does. That means that something like

    |- ( ( ph <-> ps ) -> ( isBounded ph <->  isBounded ps ) )

is a non-theorem, while

    $e |- ( ph <-> ps )
    $p |- ( isBounded ph <->  isBounded ps )

would be okay (and would be important for being able to show that composite expressions containing definitions are bounded, not just expressions using the primitives).

benjub:
The thing is that being bounded is a syntaxic, not a semantic, property. Or am I mistaken on its definition?
For instance, E. x x = 1 is not bounded while E. x e. NN x = 1 is bounded. Is it correct ? Or do we define a bounded formula as one equivalent to a syntactically-bounded one ?

digama0:
But that does mean that you can't necessarily apply vtocl as suggested above, because that requires you to prove some theorem like

    |- ( x = A -> ( isBounded ph[x] <-> isBounded ph[A] ) )

which isn't going to work for the reasons stated above.

@benjub, Being Delta_0 is usually a mixture of the two - traditionally, one defines a Delta_0 formula to be one equivalent to a syntactically bounded formula.

digama0:
For completeness, it suffices to characterize syntactically bounded formulas, since from the comprehension rule for syntactically bounded formulas you can derive comprehension for Delta_0 formulas. But it is inconvenient to deal with only syntactically bounded formulas because you can't introduce any definitions so you will end up with huge expressions when proving that some formula is Delta_0 by constructing an equivalent syntactically bounded formula.

benjub:
Therefore,
```
$e |- ( ph <-> ps )
$p |- ( isBounded ph <->  isBounded ps )
```
should rather be an axiom? Or maybe one even needs something similar with the pattern of a vtocl* theorem to be an axiom?

nmegill:
@benjub:
> (Remarks in passing: the paragraph before says "contained in" but I think "an element of" would be clearer, since "be contained" could be interpreted equally as "be included". The following comment has an unnecessary "n" in "unnnecessary".)

This is fine.  I've also seen 'contains' and 'includes' used both ways, sometimes making assertions hard to interpret correctly, having to go back carefully through several definitions to figure out the usage that was meant.  Munkre's Topology was like this, at least the first edition I learned from.  IIRC he sometimes used both senses of 'contains' on the same page!

~ wel in set.mm has the following comment, which specifies the convention we should use (maybe it should be moved to ~ conventions?):
```
  Extend wff definition to include atomic formulas with the epsilon
  (membership) predicate.  This is read " ` x ` is an element of
  ` y ` ," " ` x ` is a member of ` y ` ," " ` x ` belongs to ` y ` ,"
  or " ` y ` contains ` x ` ."  Note:  The phrase " ` y ` includes
  ` x ` " means " ` x ` is a subset of ` y ` ;" to use it also for
  ` x e. y ` , as some authors occasionally do, is poor form and causes
  confusion, according to George Boolos (1992 lecture at MIT).
```

digama0:
I don't think you can use the vtocl theorem version as an axiom, because you can derive the implication version from it. Actually it's tricky to show this is a problem intuitionistically, but using LEM you can prove `( ph -> isBounded ph )` and `( -. ph -> isBounded ph )` and deduce that every formula is bounded, since in the first case it's equivalent to `T.` and in the second case it's equivalent to `F.`.

One way to encode this without using any dubious semi-functions like `isBounded` is to use a regular predicate on classes. If we admit that pairing is bounded, then we can express any n-ary predicate (i.e. a wff with n free variables) by tupling together all the arguments and representing it as a (closed) class. Then you can simply say that such and such class is bounded (with respect to its arguments). You get axioms like this:

    |- ( A = B -> ( isBounded A <-> isBounded B ) )
    |- ( ( isBounded A /\ isBounded B ) -> isBounded ( A \ B ) )
    |- ( isBounded A -> isBounded { <. x , y >. | A. z e. x <. y , z >. e. A } )
    |- isBounded { <. x , y >. | x = y }
    |- isBounded { <. x , y >. | x e. y }

There are a few more to do with tupling variables, composition, and projections, but you get the idea.

benjub:
Thanks. I began a mathbox (see PR #1112 ) and obviously more questions will come up.  When that's the case, I'll open a new issue.

PR #1103, opened 2019-09-22
rename 2sb6rflem2 -> 2ax6e

wlammen:
discussed in pull request #1101


wlammen:
The proof idea of 2eu6 is essentially kept, but some clean up removes another 7 proof lines.

I wonder whether I should explain the proof in some way in the comment. It is long and littered with small adjustments, so it takes time to figure its concept out.

benjub:
I'm in favor of a short explanation in the comment. The bare minimum to get the main idea(s) of the proof.

PR #1105, opened 2019-09-22
Misc

benjub:
Commit messages say the most important.
I see that idi is used again by several theorems, so I removed some uses (I had done this a few months ago, but new usages reappeared).  Sometimes a
  MM-PA>min */all *
does not remove a superfluous idi...
I propose to add usage discouragement tags to dummylink and idi to (partly) avoid this from repeating.
Since some theorems in AS's mathbox use idi for pedagogical purposes, I propose adding a theorem "idiALT" at the beginning of his mathbox and use it instead of idi.
@nmegill , tell me if you want a different modification for mmset.html

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1533,6 +1533,7 @@
 "bj-inftyexpidisj" is used by "bj-pinftynrr".
 "bj-inftyexpiinv" is used by "bj-inftyexpiinj".
 "bj-vexw" is used by "bj-ralvw".
+"bj-vexwt" is used by "bj-vexw".
 "blo3i" is used by "ipblnfi".
 "blocn" is used by "blocn2".
 "blocn2" is used by "ubthlem1".
@@ -13922,7 +13923,6 @@ New usage of "ax1rid" is discouraged (0 uses).
 New usage of "ax2" is discouraged (0 uses).
 New usage of "ax3" is discouraged (0 uses).
 New usage of "ax4" is discouraged (0 uses).
-New usage of "ax52" is discouraged (0 uses).
 New usage of "ax5ALT" is discouraged (0 uses).
 New usage of "ax5el" is discouraged (1 uses).
 New usage of "ax5eq" is discouraged (1 uses).
@@ -14068,6 +14068,7 @@ New usage of "bj-rabtrALT" is discouraged (0 uses).
 New usage of "bj-rabtrAUTO" is discouraged (0 uses).
 New usage of "bj-sbeqALT" is discouraged (0 uses).
 New usage of "bj-vexw" is discouraged (1 uses).
+New usage of "bj-vexwt" is discouraged (1 uses).
 New usage of "bj-xpima1snALT" is discouraged (0 uses).
 New usage of "blo3i" is discouraged (1 uses).
 New usage of "blocn" is discouraged (1 uses).
@@ -18241,6 +18242,7 @@ Proof modification of "2sb5ndVD" is discouraged (230 steps).
 Proof modification of "2sb5rfOLD" is discouraged (98 steps).
 Proof modification of "2sb6rfOLD" is discouraged (98 steps).
 Proof modification of "2sbcrexOLD" is discouraged (63 steps).
+Proof modification of "2stdpc5" is discouraged (34 steps).
 Proof modification of "2uasban" is discouraged (34 steps).
 Proof modification of "2uasbanh" is discouraged (214 steps).
 Proof modification of "2uasbanhVD" is discouraged (313 steps).
@@ -18298,7 +18300,6 @@ Proof modification of "ax2" is discouraged (46 steps).
 Proof modification of "ax3" is discouraged (24 steps).
 Proof modification of "ax3h" is discouraged (21 steps).
 Proof modification of "ax4" is discouraged (45 steps).
-Proof modification of "ax52" is discouraged (14 steps).
 Proof modification of "ax5ALT" is discouraged (3 steps).
 Proof modification of "ax5el" is discouraged (30 steps).
 Proof modification of "ax5eq" is discouraged (30 steps).
@@ -18351,6 +18352,7 @@ Proof modification of "binom2aiOLD" is discouraged (171 steps).
 Proof modification of "bitr3" is discouraged (14 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).
 Proof modification of "bj-19.21t" is discouraged (39 steps).
+Proof modification of "bj-2stdpc4v" is discouraged (28 steps).
 Proof modification of "bj-abbi" is discouraged (87 steps).
 Proof modification of "bj-abbi1dv" is discouraged (24 steps).
 Proof modification of "bj-abbi2dv" is discouraged (24 steps).
@@ -18362,7 +18364,9 @@ Proof modification of "bj-abelcmnd" is discouraged (10 steps).
 Proof modification of "bj-abelcmnda" is discouraged (5 steps).
 Proof modification of "bj-abeq1" is discouraged (36 steps).
 Proof modification of "bj-abeq2" is discouraged (47 steps).
+Proof modification of "bj-abfal" is discouraged (54 steps).
 Proof modification of "bj-abid2" is discouraged (14 steps).
+Proof modification of "bj-abtru" is discouraged (29 steps).
 Proof modification of "bj-aecomsv" is discouraged (16 steps).
 Proof modification of "bj-aev" is discouraged (39 steps).
 Proof modification of "bj-aevlem1v" is discouraged (42 steps).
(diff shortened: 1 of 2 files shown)


nmegill:
@benjub I have been unsuccessful at resolving the merge conflict that arose due to wlammen and avekens PRs.  I thought it was due to editing the label change list (as it was with a wlammen/avekens conflict) but apparently it's not.  Could you take a look?

> Commit messages say the most important.
> I see that idi is used again by several theorems, so I removed some uses (I had done this a few months ago, but new usages reappeared). Sometimes a
> MM-PA>min */all *
> does not remove a superfluous idi...

Strange.  Can you give me a test case?

> I propose to add usage discouragement tags to dummylink and idi to (partly) avoid this from repeating.

I think mmj2 people often use idi while constructing proofs, so it's probably not a good idea.  Anyway, 'min' should remove them unless there's a bug.

> Since some theorems in AS's mathbox use idi for pedagogical purposes, I propose adding a theorem "idiALT" at the beginning of his mathbox and use it instead of idi.

OK

> @nmegill , tell me if you want a different modification for mmset.html

I made a quick change on us2 (independently of GitHub just to see its appearance).  The text I added immediately after "an overview of Metamath and download links." was "If you wish to contribute your own proofs to the Metamath project, see <A HREF="../index.html#contribute">How can I contribute to Metamath?</A>"  Perhaps you can add that text to your PR.

Thanks,
Norm

nmegill:
> I made a quick change on us2 (independently of GitHub just to see its appearance). The text I added immediately after "an overview of Metamath and download links." was "If you wish to contribute your own proofs to the Metamath project, see [How can I contribute to Metamath?](../index.html#contribute)" Perhaps you can add that text to your PR.

The link was mangled.  Here is the text I added:

```
If you wish to contribute your own proofs to the Metamath project, see
<A HREF="../index.html#contribute">How can I contribute to Metamath?</A>
```

benjub:
Unfortunately, I cannot reproduce the "failed removal of idi under MM-PA>min". It was with a former proof of bj-issetw, that I got back thanks to your new "old" repository (http://us2.metamath.org/old/mpeuni/bj-issetw.html). The only thing I can reproduce is the following where I "/except" the label that would take the place of idi.  I'm pretty sure I did the same thing earlier without the "/except" and idi was not removed... I'll try to track it better if/when this happens again.
```
MM-PA> sh n
36   3bitri.1=df-clel    $a |- ( A e. { x | ph } <-> E. z ( z = A /\ z e. { x | ph } ) )
82             idi.1=bj-issetw.1   $e |- ph
83           bj-vexwv.1=idi      $p |- ph
84         biantru.1=bj-vexwv  $p |- z e. { x | ph }
85       bicomi.1=biantru    $p |- ( z = A <-> ( z = A /\ z e. { x | ph } ) )
86     exbii.1=bicomi      $p |- ( ( z = A /\ z e. { x | ph } ) <-> z = A )
87   3bitri.2=exbii      $p |- ( E. z ( z = A /\ z e. { x | ph } ) <-> E. z z = A )
91   3bitri.3=bj-denotes $p |- ( E. z z = A <-> E. y y = A )
92 bj-issetw=3bitri    $p |- ( A e. { x | ph } <-> E. y y = A )
MM-PA> min * /allow */except bj-vexwv
No shorter proof was found.
```

nmegill:
> 82             idi.1=bj-issetw.1   $e |- ph
83           bj-vexwv.1=idi      $p |- ph
84         biantru.1=bj-vexwv  $p |- z e. { x | ph }
...
> MM-PA> min * /allow */except bj-vexwv
...
>No shorter proof was found.

This is expected.

The way minimize works is that it takes a $p or $a that is allowed (by the 'min' arguments) and tries matching it to existing proof steps, then checks to see whether the proof was shortened.  In this case, if bj-vexwv were not blocked, it would match the bj-vexwv conclusion to step 84 and the hypothesis bj-vexwv.1 to step 82, and if the proof was shortened (as it would be with idi eliminated) it replaces the old proof section with the new bj-vexwv match.

But since you have blocked bj-vexwv, it is unable to "use" bj-vexwv in this way.

The message would normally say the proof was shortened using bj-vexwv, and nothing about idi.  It doesn't care about idi and isn't even really aware of it - it just discards the resulting unused proof stub consisting of idi and possibly others (such as a long chain of idi's), but bj-vexwv is what was actually "used" to shorten the proof.

BTW if you have a long chain of idi's, it's possible that idi would be mentioned as one of the theorems it used to shorten the proof (because one idi could replace 2 idi's and thus shorten the proof).  However, the algorithm keeps trying until nothing changes, and in the end all idi's should be gone.  Or by coincidence, depending on scan order, it could also remove the entire idi chain at once and make no mention of idi.

benjub:
@nmegill I understood that this result is expected, but it had happened without me using "/except".  I guess I know why, now: it's because the said theorem had discouraged usage. To be sure, I added this to set.mm:
```
$( Test for elimination of ~ idi from a proof using "MM-PA> minmize". $)
  bj-merlem1 $p |- ( ( ( ch -> ( -. ph -> ps ) ) -> ta ) -> ( ph -> ta ) ) $=
     ( wn wi merlem1 idi ) CAEBFFDFADFFABCDGH $.
```
and then,
```
MM> pr bj-merlem1
You will be working on statement (from "SHOW STATEMENT bj-merlem1"):
137194 bj-merlem1 $p |- ( ( ( ch -> ( -. ph -> ps ) ) -> ta ) -> ( ph -> ta ) ) $= ... $.
MM-PA> sh n
17   idi.1=merlem1  $p |- ( ( ( ch -> ( -. ph -> ps ) ) -> ta ) -> ( ph -> ta ) )
18 bj-merlem1=idi $p |- ( ( ( ch -> ( -. ph -> ps ) ) -> ta ) -> ( ph -> ta ) )
MM-PA> min */all *
No shorter proof was found.
(Other mathboxes were not checked.  Use / INCLUDE_MATHBOXES to include them.)
```
This is also "expected" since ~merlem1 has discouraged usage (and the command MM-PA> min */ov correctly removes ~idi). I think it is not worth modifying the code of "minimize_with" for such special cases.  How to manually remove the superfluous ~idi step?  Generally, I would do "DEL STEP ..." and retype the corresponding subproof, but it could be long.  Maybe it's better to save the proof in "normal" format, then simply erase "idi" from "set.mm", and then resave in compressed format? Or another method? Thanks.

nmegill:
@benjub 
> How to manually remove the superfluous ~idi step? Generally, I would do "DEL STEP ..." and retype the corresponding subproof, but it could be long.

It should be trivial in mmj2 to do this kind of editing.  Realistically, mmj2 is what you should use for this sort of thing.

It's harder in MM-PA as you guessed.  To avoid retyping the subproof, I have a quick and dirty ad-hoc script that turns 'show new_proof' into a proof-building script with successive 'assign's (in reverse order of the proof steps).  To remove idi manually I would first generate the proof-building script, delete the idi step (and thus its subproof), and finally extract and run the subproof portion of the proof-building script (instead of retyping the subproof).

However, I have never actually removed idi manually because it can always be done automatically, worst case forcing it with overrides as you indicate.  For safety when overriding, I would type "min merlem1/a */ov" rather than "min */a */ov".  That runs much faster as well.

For the record, here is my script to generate a proof-building script from 'show new_proof'.  If there are dummy variables, those have to be assigned by hand in the generated proof.  If the original proof is not complete, the generated proof-building script is quirky and unusable by anyone but me, and I can't offer much support.  I hope I won't regret publishing this. :)

```
set width 9999
open log 1.tmp
show new_proof
close log
set width 79
tools
match 1.tmp $ y
clean 1.tmp ber
substitute 1.tmp ' ' '%^' 1 ''
add 1.tmp ! ''
delete 1.tmp ^ =
delete 1.tmp " " ""
!substitute 1.tmp obs "" a ""
substitute 1.tmp '%' '%assign last ' 1 ''
add 1.tmp "" " /no_unify"
reverse 1.tmp
substitute 1.tmp % \n a ''
quit
quit
```

BTW this assumes you are already in MM-PA of a known good proof.  The 2 quits at the end go back to MM>, under the assumption you will use the generated proof-building script to build a subproof of some other proof, or maybe regenerate the proof for a copy of the theorem with different variable names, that sort of thing.  (The script tends to change depending on what I need at the moment.)

benjub:
Thanks for these explanations.  Note: it could happen that, in some proof, "MM-PA> min thm/ov" might contract "idi thm" to "thm" as wanted, but may also introduce other unwanted uses of "thm" at other places in the proof. Therefore, manual removal of idi might be safer.
Actually, MM is more stable than mmj2 and I always use MM for small (<30 essential steps) proofs.  I use mmj2 for the rare larger proofs I do. Apparently, I was using an older version, so hopefully I'll have a better experience with v2.5.3.

digama0:
Regarding idi use in mmj2: In recent-ish versions of mmj2, idi is no longer necessary. If you write `#` in place of a theorem name, it will search all previous steps for one with the same statement as the current one, and delete the current step and edit all references to point to the previous one instead. You can also be more explicit about it, and write `#205` say to delete the current step and replace it with step 205. This will succeed where `#` fails, if it is not an exact match but requires some metavariable unification. Furthermore, the `#n` style can even resolve a forward reference (if step n is after the current step), and it will reorder the steps as necessary (provided it is not actually a cyclic proof), which is super useful for doing step reordering. (You can also just reorder the steps in the worksheet, and mmj2 will restore the order.)

benjub:
Thanks @digama0, this feature of mmj2 that I didn't know looks very useful !

PR #1107, opened 2019-09-23
shorten 2eu6 yet again

wlammen:
It's not that I look through 2eu6's proof again and again, hunting for shortenings there.  I just look at nearby theorems to see how they might profit from techniques I used in 2eu6. And then it works the other way round.  This time it was mo3 that hinted me to a shortening in 2eu6, helping to shrink the proof by another three lines.  Sorry for keeping you busy, Norm.


nmegill:
> Sorry for keeping you busy, Norm.

Not a problem, don't worry about it.  Shorter proofs are always welcome.

PR #1108, opened 2019-09-24
documentation fixes especially CZF

jkingdon:
As suggested by @benjub at https://github.com/metamath/set.mm/pull/1102#issuecomment-534292550 this makes the following documentation fixes:

1. Takes a paragraph about CZF which had been in iset.mm (as a section header for the set theory section) and copies it to the web page where we list the IZF axioms

2. changes "contained in" to "an element of" in one sentence (which appears four places). This is more precise, and in this context at least, reads just as naturally.

3. Fixes a typo "unnnecessary". Apparently one of my keyboards has a sticky "n" key, and/or I was pretty prolific in copy-pasting this typo.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -16598,7 +16598,8 @@ $(
 
   Set theory uses the formalism of propositional and predicate calculus to
   assert properties of arbitrary mathematical objects called "sets."  A set can
-  be an element of another set, and this relationship is indicated by the ` e. `
+  be an element of another set, and this relationship is indicated by the
+  ` e. `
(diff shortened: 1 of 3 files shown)


nmegill:
@jkingdon Another 80-character line problem...

jkingdon:
> @jkingdon Another 80-character line problem...

Ha! 4 minutes before you posted that I had already pushed up the fix. Not that there's anything wrong with pointing it out, in case I didn't see the travis output.

PR #1111, opened 2019-09-25
2eu6 again

wlammen:


benjub:
Nice, @wlammen ! Once the proof is stabilized, a short explanation of the proof's main idea in the comment would be great.

PR #1112, opened 2019-09-25
Misc

benjub:
* iset.mm : added comment to acexmid. The main reason is not so much historical, but the theorem is usually referred to as Diaconescu's (wikipedia, nlab, Bauer's article Jim cited), and it's so that the theorem acexmid "pops up" when one searches for "Diaconescu" (in e.g. google or MM>search xxx/comment). I hope @jkingdon is OK.
* iset.mm: created my mathbox to begin work on bounded formulas following discussion in PR #1102  I hope @jkingdon and @digama0 are OK with that (especially that I spelled out your names in full in the comment). I'm now realizing that I wrote "contributed by BJ" for the axioms, but I could totally change it to JK or MC.
* set.mm: some tidying in the $t / mathbox part.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -48865,7 +48865,7 @@ $)
     $d x y z w v u s A $.  $d x y z w v u s B $.  $d x y z w v u s C $.
     $d x y z w v u s ph $.
     acexmidlem.a $e |- A = { x e. { (/) , { (/) } } | ( x = (/) \/ ph ) } $.
-    acexmidlem.b $e |- B = { x e. { (/) , { (/) } } 
+    acexmidlem.b $e |- B = { x e. { (/) , { (/) } }
                                                 | ( x = { (/) } \/ ph ) } $.
     acexmidlem.c $e |- C = { A , B } $.
     $( Lemma for ~ acexmid .  (Contributed by Jim Kingdon, 6-Aug-2019.) $)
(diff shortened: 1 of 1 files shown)


benjub:
* I added DV conditions to df-frec to pass mmj2's definition check.  I hope it's OK. It's strange, though, that it had passed before this PR.  Was it because of the mmj2 definition check bug that was mentioned in PR #1089 ?

benjub:
* I updated the top date of iset.mm because it did not pass Travis, but now, there are conflicts... I hope it can be sorted out.

nmegill:
> * I updated the top date of iset.mm because it did not pass Travis, but now, there are conflicts... I hope it can be sorted out.

Put the date back the way it was, to solve the merge conflict, and let Travis fail.  Then I can merge, and I'll update the date for you.

nmegill:
It is merged and fixed now.  (I didn't do a PR for it.)

jkingdon:
Thanks for all this. The missing distinct variable constraint is of course the most severe (especially if it leads to unsoundness which would be my default assumption in such cases). I'll need to take a look at the Dedekind cut definition I'm working on now to avoid the same problem.

The text regarding Diaconescu and the others looks good. It sums up the situation nicely.

No worries about the attribution. My general preference is to just list who wrote the theorem for metamath, not their predecessors (e.g. people who wrote theorems that this one may be a modification of, etc) or successors (e.g. people who modified the proof). Hopefully I haven't offended anyone by following this practice (sometimes it leads to me getting "too much" credit, sometimes "too little").

But more to the point, congratulations on getting started on the mathbox. I was just thinking of CZF when I read "the Dedekind reals work better with power sets or some other form of impredicativity" in the HoTT book (chapter 11 intro), and realized I'm about to use the power set axiom to prove real numbers exist. Will be interesting to see what you find.

benjub:
Indeed, the axiom of subset collection is required in order to prove that the Dedekind reals form a set in CZF, https://arxiv.org/abs/1510.00641

PR #1113, opened 2019-09-26
Document extra set.mm checks (definitionCheck.js and LRParser.java)

david-a-wheeler:
The set.mm database undergoes a set of extra checks;
this commit documents those checks in the "conventions" secion.
These are:

* definitionCheck.js in mmj2, which adds requirements on
  (most) definitions to keep them non-creative and eliminable.
* LRParser.java in mmj2, which requires an unambiguous tree parse.


benjub:
Thanks. I would have kept the phrase "If [the definition] passes Rule 1, then we can define..."
Also, I would remove the digression after Rule 3. If you want to keep it, then maybe add it as a remark after the list of rules (and eliminatible --> eliminable).
Also, I prefer "disjoint variable" over "distinct variable" (and I think @nmegill does as well?).

Remark: for the moment, the definition check done by Travis still excludes df-sbc, which should not be the case.

david-a-wheeler:
@benjub - I kept the extra material in rule 3, because I think it's important for explaining why the rule exists. I plan to make the other changes.

benjub:
So maybe put it after the list of rules?
I forgot: there are four "set" which should be "setvar".

david-a-wheeler:
@benjub - good point about setvar, I will make that change.

I think the details about rule 3 belong right there with rule 3.  It is not at all obvious why rule 3 is important, indeed, it seems like nonsense, so I think an explanation of the problem it counters belongs immediately with the rule.

david-a-wheeler:
@benjub :
> Remark: for the moment, the definition check done by Travis still excludes df-sbc, which should not be the case.

Great point. I'll make that a separate pull request.

set.mm
```diff
@@ -369106,6 +369106,95 @@ varying definitions (some start from 0, others start from 1), but
        renamer theorem like ~ cbval . The section
        ~ mmtheorems32.html#mm3146s also describes the
        metatheorem that underlies this.
+       </p>
+
+       <p>
+       Standard Metamath verifiers do not distinguish between axioms and
+       definitions (both are $a statements).
+       In practice, we require that definitions (1) not be creative
```
digama0: s/creative/conservative/

set.mm
```diff
@@ -369106,6 +369106,95 @@ varying definitions (some start from 0, others start from 1), but
...
+       a logically equivalent expression that previously qualified as a wff).
+       To ensure this, we have additional rules on almost all definitions
+       ($a statements with a label that does not begin with ax-).
+       These additional rules are not applied in a few cases where they
+       are too strict ( ~ df-bi , ~ df-clab , ~ df-cleq , and ~ df-clel );
+       see those definitions for more information.
+       These additional rules for definitions are checked by at least
+       mmj2's definition check (see
+       <A HREF="https://github.com/digama0/mmj2/blob/master/"
+       >mmj2 master</a> file mmj2jar/macros/definitionCheck.js).
```
digama0: This could be a direct link? Is it because you want to stay under 80 chars?

set.mm
```diff
@@ -369106,6 +369106,95 @@ varying definitions (some start from 0, others start from 1), but
...
+       This definition check relies on the database being very much like
+       set.mm, down to the names of certain constants and types, so it
+       cannot apply to all Metamath databases... but it is useful in set.mm.
+       In this definition check, a $a-statement with a given label and
+       typecode ` |- ` passes the test if and only if it
+       respects the following rules (these rules imply that we have
+       an unambiguous tree parse, which is also checked):
```
digama0: These rules don't ensure an unambiguous parse; rather, they presume that the expression has a unique parse, and unambiguity and parse existence is checked separately.

set.mm
```diff
@@ -369106,6 +369106,95 @@ varying definitions (some start from 0, others start from 1), but
...
+           its definiens as its right hand side (RHS).
+           We define the *defined symbol* as the root-symbol of the LHS.
+           We define a *dummy variable* as a variable occurring
+           in the RHS but not in the LHS.
+           Note that the "root-symbol" is the root of the considered tree;
+           it need not correspond to a single token in the database
+           (e.g., see ~ w3o or ~ wsb ).
+       <li>The defined symbol should not occur in an earlier
+           $a-statement with typecode ` |- ` or in any another place
+           in this labelled expression.
```
digama0: I would rephrase this to say that it should not appear at all in any statement strictly between the syntax axiom and the definition. (If we institute the rule that the syntax axiom always immediately precedes the definition then this becomes moot.)

set.mm
```diff
@@ -369106,6 +369106,95 @@ varying definitions (some start from 0, others start from 1), but
...
+       <li>No two variables occurring in the LHS should share a
+           disjoint variable (DV) condition.
+           Without this rule, you can define something like
```
digama0: I agree with Benoit that the justification for the rules should come after the list. I'm not sure if we even need the example here, but please rename Bar to Foo if so.

nmegill:
> Also, I prefer "disjoint variable" over "distinct variable" (and I think @nmegill does as well?).

"Disjoint variable" is the technically correct description for the general case, but I worry that an average reader might be confused by it.  Moreover, set.mm doesn't have the kind of $d e.g. $d ph ps $. that would need "disjoint" to understand it.   Instead, we have $d x y $. (x must be distinct from y) and $d x ph $. (x must be distinct from all variables in ph i.e. x doesn't occur in ph).

I often use "distinct variable" informally for this reason, hoping it would make sense to more readers.  For example, "distinct" is used almost always on the MPE home page.  However, I tend to use "disjoint" in a technical description or spec.  I could go either way here - if you (David) have no preference, maybe use "disjoint" since Benoit does have a preference.

I'll note that I'm not aware of any literature outside of Metamath that uses "disjoint" - it's always "distinct" or "does not occur in".

set.mm
```diff
@@ -369106,6 +369106,95 @@ varying definitions (some start from 0, others start from 1), but
...
+           The justification theorem must be of form
+           ` |- ( ` definiens root-symbol definiens' ` ) `
+           where definiens' is definiens but the dummy variables are all
+           replaced with other unused dummy variables of the same type.
+           Note that root-symbol is ` <-> ` or ` = ` , and that setvar
+           variables are simply variables with the ` setvar ` typecode.
+       <li>One of the following must be true:
+           (a) there must be no setvar dummy variables,
+           (b) there must be a justification theorem as described in rule 5, or
+           (c) we must prove that each dummy setvar variable is not free,
```
digama0: Technically (c) subsumes (a)

digama0: The modality here is wrong. This is a correctness condition, so you should drop "we must prove that". Probably we want to prove this, but the condition itself simply asserts that each dummy variable is not free.

set.mm
david-a-wheeler: Right.  It was originally a direct link but then it can't commit because Travis requires all lines to be under 80 chars.

digama0: @nmegill Thoughts? I could have sworn this has come up before.

set.mm
david-a-wheeler: Let's do s/not be creative/be conservative/ instead :-).

set.mm
david-a-wheeler: You're absolutely right, will fix.  What I meant != what I said :-).

set.mm
david-a-wheeler: Okay, but it also can't occur in the definiens, and as written that might forbid df-3an. How about this:

The defined expression must not appear in any statement between its syntax axiom ( ` $a wff ` ) and its definition, and the defined expression must be not be used in its definiens. See ~ df-3an for an example where the same symbol is used in different ways (this is allowed).

set.mm
david-a-wheeler: I think it's better where it is, but 2 people disagree with me & I don't care *that* much. So I'll move it.  I'll rename the example to foo.

set.mm
david-a-wheeler: Okay, will tweak to clarify.

david-a-wheeler:
I've pushed another version that I *hope* resolves all comments up to this point, as commit 4cdc7b94a1d3dfe7bac806fbbe76f2019b1e7673.

nmegill:
@digama0 Could you do the merge for this PR when you are satisfied with the changes? Thanks.

david-a-wheeler:
@nmegill - sounds good. @digama0 - I *think* got them all, let me know of anything I'm missing.

set.mm
david-a-wheeler: @digama0 - it's a really long URL, I don't remember there being a "nice" way to deal with this.

PR #1115, opened 2019-09-26
Check df-sbc along with other definitions

david-a-wheeler:
At one time df-sbc had to be excluded from the set.mm
definition check, but it has since been modified to
follow the conventions of the other definitions.

We want to minimize what we exclude from checking, so
remove the exclusion of df-sbc.


david-a-wheeler:
This is thanks to a reminder from @benjub .

PR #1118, opened 2019-09-27
Run definition checker on iset.mm in travis

jkingdon:
The definition checker is just as useful on iset.mm as on set.mm, and in fact would have caught a recently omitted distinct variable constraint in a definition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,7 @@ script:
   # Verify and also generate 'discouraged.new', and do extra checking
   # (the 'printf' lets us insert multiple lines):
   - scripts/verify-mmj2 --setparser 'mmj.verify.LRParser' --extra "$(printf 'RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel\nRunMacro,showDiscouraged,discouraged.new')" set.mm
+  - scripts/verify-mmj2 --setparser 'mmj.verify.LRParser' --extra "$(printf 'RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel\nRunMacro,showDiscouraged,iset-discouraged.new')" iset.mm
   - echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
   # Use setparser to check that definitions don't create syntax ambiguities
   - scripts/verify-mmj2 --setparser 'mmj.verify.LRParser' iset.mm


david-a-wheeler:
Looks great!

Eventually you will probably want to add an iset-discouraged file and a test to ensure that the file has the expected contents (just like set.mm does).  But doing that in a separate pull request would be sensible.

benjub:
A few things in .travis.yml look strange to me.  I don't know the language so this might be correct, but maybe @jkingdon or @david-a-wheeler can have a look.

l.48--50: now obsolete ?
l.58: change to " # Create mmset.html and mmil.html so we can verify against them." ?
l.62: why are the options different from l.61 ?
l.67: why no analogue for some "idiscouraged" ?
l.68: comment is actually describing l.65,66 ?
l.69: now redundant with l.66 ?
l.78-84: this generates mpe but not ile ?

nmegill:
@jkingdon The definition check you added gave the following errors:
```
I-UT-0015 **** Processing RunParmFile Command #5 = RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel
I-PA-0204 Axiom df-i1p has failed the definitional soundness check. All dummy variables in the definiens are required to be distinct from each other and from variables in the definiendum. The following DJ conditions need to be added:
  $d l u $.
I-PA-0204 Axiom df-iplp has failed the definitional soundness check. All dummy variables in the definiens are required to be distinct from each other and from variables in the definiendum. The following DJ conditions need to be added:
  $d q r s x y $.
```

jkingdon:
> @jkingdon The definition check you added gave the following errors:
> 
> ```
> I-UT-0015 **** Processing RunParmFile Command #5 = RunMacro,definitionCheck,ax-*,df-bi,df-clab,df-cleq,df-clel
> I-PA-0204 Axiom df-i1p has failed the definitional soundness check. All dummy variables in the definiens are required to be distinct from each other and from variables in the definiendum. The following DJ conditions need to be added:
>   $d l u $.
> I-PA-0204 Axiom df-iplp has failed the definitional soundness check. All dummy variables in the definiens are required to be distinct from each other and from variables in the definiendum. The following DJ conditions need to be added:
>   $d q r s x y $.
> ```
OK, I'll make a follow up pull request to add those constraints (I'm not completely sure they lead to unsoundness in this case, but it is better to play it safe).

david-a-wheeler:
@benjub - you're right, the .travis file needs some scrubbing.  Let's discuss follow-on work in issue #1120 since this pull request is specifically about adding the definition checker to iset.mm.

david-a-wheeler:
@jkingdon - please make that follow-on fix as soon as you reasonably can.  I've made additional tweaks to .travis.yml, and can't really confirm that everything works while the branch basis fails :-).

PR #1122, opened 2019-09-28
Fix .travis.yml, regen-discouraged, add iset-discouraged, fixes #1120

### LINKED ISSUE #1120: Scrub .travis.yml file
@benjub looked at the .travis.yml file and noted a number of things that probably need fixing:

~~~~
l.48--50: now obsolete ?
l.58: change to " # Create mmset.html and mmil.html so we can verify against them." ?
l.62: why are the options different from l.61 ?
l.67: why no analogue for some "idiscouraged" ?
l.68: comment is actually describing l.65,66 ?
l.69: now redundant with l.66 ?
l.78-84: this generates mpe but not ile ?
~~~~

david-a-wheeler:
Tweak .travis.yml per issue #1120 from @benjub;
the file had gotten a little cruft and no-longer-needed statements.

This adds iset-discouraged, so iset.mm undergoes the same
"discourage checking" that set.mm already undergoes.
If you *do* want to use something that is discouraged in iset.mm,
you can use this to regenerate the file:

> scripts/regen-discouraged --results iset-discouraged iset.mm

This commit also fixes scripts/regen-discouraged; it was
supposed to already work this way, but due to a bug it didn't
actually write to the --results file.  That's now fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 # Only one verifier is necessary, but using multiple verifiers counters
 # the risk that a verifier might have a defect in its implementation.
 # With 4 verifiers, all 4 would have to have a defect involving an
-# invalid theorem # for that invalid theorem to be accepted as valid.
+# invalid theorem for that invalid theorem to be accepted as valid.
 # This is unlikely, because the verifiers are written by different people
 # in different languages.  This can be considered extreme peer review, where
 # every reviewer is separately checking each step.
@@ -45,9 +45,6 @@ install:
   - wget -N -q http://us.metamath.org/ocat/mmj2/mmj2jar.zip
   - unzip -q mmj2jar.zip
   - jdk_switcher use oraclejdk8
-  ##### 27-Mar-2019 daw TEMPORARY WORKAROUND - modify mmj2 definitionCheck.js
-  ##### so that it looks for "setvar" instead of "set":
-  - sed -i -e 's/"set"/"setvar"/g' macros/definitionCheck.js
   # Install smetamath-rs (smm3), a Rust verifier by Stefan O'Rear
   # See: https://github.com/sorear/smetamath-rs
   - "[ -x  ~/.cargo/bin/smetamath ] || cargo install smetamath --vers 3.0.0"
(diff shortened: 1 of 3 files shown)


david-a-wheeler:
@benjub - this should resolve many of the issues you noticed.

@jkingdon - this implements the "discouraged" checking for iset.mm, just like it already exists in set.mm.  It also fixes the regen-discouraged script so it will actually *work* for iset.mm.  Once this is merged you'll need to occasionally regenerate iset-discouraged, e.g., using:

~~~~
  scripts/regen-discouraged --results iset-discouraged iset.mm
~~~~~

In particular, while you're working on Dedekind cuts you'll do a lot with temporary values that I suspect you'll be discouraging, so you'll probably have to regenerate them often as you add discouraged things or use discouraged things.  That should be a good set of tests to ensure that the detection system for discouraged things works properly.

david-a-wheeler:
Hmm, we have substantive failures from iset.mm (copied below).  I'll wait until those are fixed, and then merge the fix into this branch.

~~~~
I-PA-0204 Axiom df-i1p has failed the definitional soundness check. All dummy variables in the definiens are required to be distinct from each other and from variables in the definiendum. The following DJ conditions need to be added:

  $d l u $.

I-PA-0204 Axiom df-iplp has failed the definitional soundness check. All dummy variables in the definiens are required to be distinct from each other and from variables in the definiendum. The following DJ conditions need to be added:

  $d q r s x y $.
~~~~

nmegill:
@david-a-wheeler Jim's PR to fix the iset.mm definition errors was merged.

I noticed there were many warnings for the mmdeduction.raw.html conversion although I didn't check what is going on.

jkingdon:
@david-a-wheeler @nmegill I took the liberty of checking out this branch, running `git merge origin/develop`, and pushing up the result. That should bring in the fixes to df-i1p and df-iplp .

As for the discouraged checking, thanks. You are correct that it'll get heavy use in the section on real number construction, since every theorem is discouraged there.

jkingdon (approved):

nmegill:
> In particular, while you're working on Dedekind cuts you'll do a lot with temporary values that I suspect you'll be discouraging, so you'll probably have to regenerate them often as you add discouraged things or use discouraged things. That should be a good set of tests to ensure that the detection system for discouraged things works properly.

It probably isn't necessary to discourage the Dedekind stuff (like is done in set.mm) until the section is finished, otherwise it will be rather annoying constantly having to override things during development.  However, there are discouragements on e.g. *ALTs already in iset.mm that will do some checking.

jkingdon:
....and of course discouraged needs re-generating. I've pushed up a commit to do that.

As for whether to discourage the real number construction as I go, or at the end once I've introduced `NN` , `QQ` , etc, I'm not sure I have a super-strong opinion but in general I find it easier to keep things similar to set.mm. If I get tired of regenerating the iset-discouraged file, I can always change my mind.

PR #1123, opened 2019-09-28
shorten 2eu6

wlammen:
Business as usual as it seems, 2eu6 is shortened by another 3 proof lines.

Some formatting changes are caused by rewrap, which mercilessly throws indentations away, even if they improve readability.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15617,6 +15617,7 @@ New usage of "eu3OLD" is discouraged (2 uses).
 New usage of "eu5OLD" is discouraged (0 uses).
 New usage of "euanOLD" is discouraged (0 uses).
 New usage of "euequ1OLD" is discouraged (0 uses).
+New usage of "euexOLD" is discouraged (0 uses).
 New usage of "eufOLD" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
 New usage of "eumo0OLD" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


benjub:
From the current http://us2.metamath.org/mpeuni/euex.html versus http://us2.metamath.org/mpeuni/euexOLD.html , I see that euexOLD is shorter but uses more axioms (ax-1--13 versus ax-1--6) and more definitions.  If this is the case, then euex is correctly the main proof, and euexOLD could be relabelled euexALT and its comment should read something like "Atlernatre proof of ~ euex . Shorter but uses more axioms." while the comment of euex could add "For a shorter proof using more axioms, see ~ euexALT ."

wlammen:
Thanks for the hint @benjub, I will restore the old proof.  I did not look at the axioms.  I do not really care about an ALT version, though, as I find it not that inspiring.

nmegill:
> Some formatting changes are caused by rewrap, which mercilessly throws indentations away, even if they improve readability.

rewrap doesn't touch any statement's comment containing the markup keyword "<HTML>", _except_ that it starts all lines in the correct column according to the indentation.  This is to give the file as a whole a consistent appearance.

The 'conventions' comment was originally put into set.mm rather than made into a separate file because we needed math symbols and label hyperlinks.  Because of its huge size, some indentation might make it prettier, but  I don't think it is worth adding more complexity with special markup to just to control indentation for that one comment.  Because it is HTML code, it is already rather awkward to read in a text editor, and I don't think many people do except to edit it.  At some point it might make sense to make it a separate HTML file since we now have the 'markup' command; then the indentation could be controlled any way the author wants.

nmegill:
> markup keyword ""

markup keyword "`<HTML>`"

PR #1124, opened 2019-09-28
Misc

benjub:
Modified some scripts:
* fixed the ones using "minimize_with" with option "allow"
* changed metamath to ./metamath for safety
* made "list-theorems" write to "output" which defaults to 1.tmp (instead of hardcoded 1.tmp) and made minimize-all, which uses list-theorems, specify the argument 1.tmp, for robustness. Is this correct, @david-a-wheeler ?
* Can one (I think one needs privileges) rename scripts/NOTES to scripts/README.txt (or .md) so that it appears on the GitHub page?
* I noticed that some scripts use a temporary file 0.tmp. Is there a simple way to make sure it avoids overwriting an existing file? (i.e. make sure that the "dummy file" used in the script is a new file and avoids clashes?)  Idem with the default output 1.tmp ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ install:
   - wget -N -q http://us.metamath.org/downloads/checkmm.cpp
   - g++ -O2 -o checkmm checkmm.cpp
 script:
-  # Create mmset.html so we can verify against it.
+  # Create mmset.html and mmil.html so we can verify against them.
   - metamath 'read set.mm' 'markup mmset.raw.html mmset.html /ALT /CSS' quit
   - metamath 'read iset.mm' 'markup mmil.raw.html mmil.html /ALT /CSS' quit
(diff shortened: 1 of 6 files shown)


benjub:
Now I'm not sure anymore: maybe having "metamath" instead of "./metamath" makes it possible to use it from any directory, provided that one added the metamath directory to the $PATH variable ? In which case I'll have to revert the scripts to "metamath" and the NOTES file to remove "and is in the current directory".  I'll wait for @david-a-wheeler 's advice.

david-a-wheeler:
I recommend *NOT* using "./metamath", indeed, it won't work on my setup.  The point of using just the name "metamath" is that you can control which one is used via PATH (which is normally how that's done).

benjub:
@david-a-wheeler OK, I'll revert it, sorry for that.  After that, can you add to "NOTES" how to set PATH so that it works correctly? (and rename NOTES to README.txt so that it appears on the GitHub page?)

PR #1125, opened 2019-09-28
Various discouraged-related updates (set.mm and iset.mm)

jkingdon:
This pull request is somewhat miscellaneous; let me know if there is any reason to split it up.

1. A bit more development of real number construction in iset.mm: prop , prml , prmu , prssnql , prssnqu .

2. Upon reflection, I'll try taking @nmegill 's suggestion and remove the discouraged tags from iset.mm real number construction, at least until that construction is done.

3. Remove various discouraged theorems which have no usages. This is iset.mm-only except for brabsbOLD and opelresiOLD . Hopefully those have been discouraged long enough that people with copies of set.mm (that is, doing work outside mathboxes) have had a chance to update them.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -11198,7 +11198,6 @@
 "onfrALTlem5" is used by "onfrALTlem3VD".
 "opelcn" is used by "axicn".
 "opelopabsbOLD" is used by "brabsb2".
-"opelopabsbOLD" is used by "brabsbOLD".
 "opelopabsbOLD" is used by "cnvopab".
 "opelopabsbOLD" is used by "inopab".
 "opelreal" is used by "ax1cn".
@@ -14468,7 +14467,6 @@ New usage of "bra0" is discouraged (1 uses).
 New usage of "bra11" is discouraged (6 uses).
 New usage of "braadd" is discouraged (1 uses).
 New usage of "brabn" is discouraged (1 uses).
-New usage of "brabsbOLD" is discouraged (0 uses).
 New usage of "bracl" is discouraged (4 uses).
 New usage of "bracnln" is discouraged (1 uses).
 New usage of "bracnlnval" is discouraged (0 uses).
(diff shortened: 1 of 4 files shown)


nmegill:
> * Upon reflection, I'll try taking @nmegill 's suggestion and remove the discouraged tags from iset.mm real number construction, at least until that construction is done.

Another way to do this is to change "(New usage is discouraged.)" to a non-matching string such as "(?New usage is discouraged.)" that you can then change back with a global replacement at the end.  But any way you do it is fine of course.

BTW the match strings are in mmdata.h, e.g. 
#define USAGE_DISCOURAGED_MARKUP "(New usage is discouraged.)"

> * Remove various discouraged theorems which have no usages. This is iset.mm-only except for brabsbOLD and opelresiOLD . Hopefully those have been discouraged long enough that people with copies of set.mm (that is, doing work outside mathboxes) have had a chance to update them.

Usually we say "Obsolete as of dd-mmm-yyyy.' and periodically delete those over a year old.  These two seem to have gotten overlooked for a decade, and yes they should be deleted.

jkingdon:
> Usually we say "Obsolete as of dd-mmm-yyyy.' and periodically delete those over a year old. These two seem to have gotten overlooked for a decade, and yes they should be deleted.

Speaking of which, is the date of 2-Sep-2019 at riotaiotaOLD correct or should it be 2-Sep-2018 (which is the date given for the last revision at df-riota)?

Or is the "Obsolete as of" supposed to already have the year grace period added?

As for detecting overlooked old theorems, looking for "0 uses" in the discouraged file seems to have been a fruitful place to search. Maybe I'll do a bit more of that if I get sick of real number construction.

nmegill:
> Speaking of which, is the date of 2-Sep-2019 at riotaiotaOLD correct or should it be 2-Sep-2018 (which is the date given for the last revision at df-riota)?
Or is the "Obsolete as of" supposed to already have the year grace period added?

That was a typo.  It should be 2-Sep-2018, the day the theorem was renamed to *OLD and declared obsolete (which was when df-riota was modified).  You can delete it since a year has gone by.

PR #1126, opened 2019-09-28
Misc2

benjub:
Commit messages say it all.
@david-a-wheeler I noticed that there is no mmhol.raw.html in the GitHub repository (corresponding to http://us2.metamath.org/holuni/mmhol.html).  Is it voluntary?  Also, maybe the .gitignore file has to be updated?


nmegill:
> I noticed that there is no mmhol.raw.html

That is because there is no activity with it.  There are other mm*.html like that also.  I don't think it's necessary to create them until they are needed, unless someone wants to spend time doing it.

PR #1128, opened 2019-09-29
matrices over a polynomial ring

avekens:
Main part:
* expired obsolete ~ csbopabgOLD , ~ csbcnvgOLD removed
* ~ csbmpt12 , ~ csbmpt2 added
* ~ ply1coe revised (previous version as ply1coeOLD still available)
* ~ ply1coefsupp and ~ mptscmfsuppd moved from AV's mathbox to main set.mm
* proof of ~ ply1coe and ~ plypf1 shortened

AV's Mathbox:
* theorems for matrices in general
* theorems for matrices over a polynomial ring

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4251,7 +4251,6 @@
 "csbingOLD" is used by "onfrALTlem4VD".
 "csbingOLD" is used by "onfrALTlem5VD".
 "csbiotagOLD" is used by "csbfv12gOLD".
-"csbopabgOLD" is used by "csbcnvgOLD".
 "csbresgOLD" is used by "csbima12gALTOLD".
 "csbresgOLD" is used by "csbima12gALTVD".
 "csbrngOLD" is used by "csbima12gALTOLD".
@@ -14890,7 +14889,6 @@ New usage of "conventions" is discouraged (0 uses).
 New usage of "copsexgOLD" is discouraged (0 uses).
 New usage of "counop" is discouraged (0 uses).
 New usage of "csbabgOLD" is discouraged (10 uses).
-New usage of "csbcnvgOLD" is discouraged (0 uses).
 New usage of "csbcomgOLD" is discouraged (0 uses).
 New usage of "csbeq2gOLD" is discouraged (6 uses).
 New usage of "csbeq2gVD" is discouraged (0 uses).
@@ -14908,7 +14906,6 @@ New usage of "csbima12gOLD" is discouraged (2 uses).
 New usage of "csbingOLD" is discouraged (4 uses).
 New usage of "csbingVD" is discouraged (0 uses).
 New usage of "csbiotagOLD" is discouraged (1 uses).
-New usage of "csbopabgOLD" is discouraged (1 uses).
 New usage of "csbovgOLD" is discouraged (0 uses).
 New usage of "csbresgOLD" is discouraged (2 uses).
 New usage of "csbresgVD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
I'm not sure about the removal of csbopabgOLD. Its comment says to use csbopab instead, but the latter depends on ax-sep, ax-nul, ax-pr, contrary to csbopab. I would propose to relabel csbopabgOLD to csbopabgALT and comment "Version of csbopab with a sethood antecedent but depending on fewer axioms." and a similar comment for csbopab.
Idem with csbcnvgOLD.

avekens:
On resolving merge conflicts, I removed the entries for ~ brabsbOLD and ~ opelresiOLD from the history - I think removing OLD theorems needs not and should not be reported here (anymore)...

benjub:
Similarly as in my previous comment, opelopabsbOLD is less general than opelopabsb but has a much shorter proof; keep it as ALT ?
brabsbOLD: OK to remove (even though it depends on fewer definitions)
opelresiOLD: OK to remove
@avekens @nmegill I agree that it's not necessary to signal the removals at the top of set.mm, but on the other hand, one should be sure before remove them that the proofs do not add anything in terms of: proof length, dependency on axioms (and df-clab, df-cleq, df-clel), interest of alternate proofs... And maybe give a "final call before deletion" to give some time for contributors to check the above.

avekens:
The three OLD theorems mentioned by Benoit are now (again) available as ALT theorems.

@benjub I think as soon as a theorem is declared OLD with "obsolete as of" within the comment, it is marked to be deleted after one year. This should be enough time to decide if this theorem should be kept (and then renamed again!) or not. I recently created a lot of OLD theorems introducing "supp" and "finSupp", and all these theorems should/will be deleted next year - I see no relevance for them to live any longer...

benjub:
@avekens Agreed, and thanks for the ALT theorems.

nmegill:
> On resolving merge conflicts, I removed the entries for ~ brabsbOLD and ~ opelresiOLD from the history - I think removing OLD theorems needs not and should not be reported here (anymore)...

That's right, theorems with the suffix OLD should be never be listed in the history.  For one thing, they aren't (or shouldn't be) used, so no one is affected by their removal.

In case brabsb was "deleted" by renaming it OLD with no newer version, the history should say: "dd-mmm-yy brabsb  --- deleted; use xyz instead" where "use xyz instead" is an (optional) hint for the user.  If brabsb was replaced by a newer version that isn't directly compatible (e.g. swapped biconditional or reordered hypotheses), say: "dd-mmm-yy brabsb  [--same--] revised".  If brabsb was replaced by one with a shorter proof but no other changes, no history entry is needed.

It does seem there are some *OLDs in the history in 2015 and before.  Not sure what happened.  Sometimes people do a global change of a label to *OLD without remembering it may also change the history listing.  This is probably what happened here.  It's not worth worrying about too much though.  I think the string "OLD" can be deleted from the history as part of a cleanup someday.

nmegill:
> Similarly as in my previous comment, opelopabsbOLD is less general than opelopabsb but has a much shorter proof; keep it as ALT ?
> brabsbOLD: OK to remove (even though it depends on fewer definitions)
> opelresiOLD: OK to remove
> @avekens @nmegill I agree that it's not necessary to signal the removals at the top of set.mm, but on the other hand, one should be sure before removing them that the proofs do not add anything in terms of: proof length, dependency on axioms (and df-clab, df-cleq, df-clel), interest of alternate proofs... And maybe give a "final call before deletion" to give some time for contributors to check the above.

I think this is carrying things too far.  A year is plenty of time for a grace period, and it would be rather inconvenient for someone cleaning out *OLDs to have to announce each one, wait another month or whatever, then finally delete them.  If someone is really concerned, they can periodically check the *OLDs due for deletion in the next month or two.  Even in the worst case, the old work is still available on github.

We normally use the comment "Obsolete [version/proof of ~ xxx] as of dd-mmm-yyyy. " to indicate the start of the grace period.   For most theorems I can find the ones due for deletion in e.g. Oct 2019 by grepping for "Oct-2018" followed by period and space, since contribution dates are followed by period and parenthesis.

PR #1136, opened 2019-10-01
Fix date of Fenton's trf1ALT to 22-April-2011

david-a-wheeler:
Theorem tfr1ALT was claimed to be contributed by Scott Fenton
on 17-Aug-1994, but that is clearly an error
(it wasn't in the repo at that time).

Change its contribution date to 22-Apr-2011, the best available
estimate by Norman Megill per a private email from him
(Subject "Re: Unusually early set.mm contribution: tfr1ALT"
sent Tue, 1 Oct 2019 10:44:30 -0400).

This means that the earliest contribution recorded in set.mm
by someone other than Norman Megill is 1999-09-05,
a proof shortening by David Harvey of "notnot2".
This appears to be accurate information.

We want all credits and dates to be accurate, of course.
That sais, fixing this one is especially important for the
Gource visualization of set.mm's history, because that
visualization highlights early contributions by others.


nmegill:
My browser says "Some checks haven’t completed yet" after 1/2 hour, but the Travis checks seem to have passed.  Any idea what is happening?

david-a-wheeler:
@nmegill - every once in a while a build completes but somehow GitHub doesn't get the info that it's completed, so it gets hung forever.  It's not clear if the builder (Travis) failed to send the message, if the message was lost on the net after repeated resends, or if GitHub lost the info once it received it.

The problem is rare, but annoying.

My solution: Just restart the build (you can go to the build and force a restart).  It's the build equivalent of turning it off & back on again.  Wasteful, but it should resolve the problem.

david-a-wheeler:
@nmegill - looks like the build restarts resolved the problem.

benjub:
@david-a-wheeler A related thing that I mentioned on the google group is the very early appearance of "Supplementary material... for BJ".  It is due to me having "backed up" an early theorem by @nmegill that would have been deleted otherwise.  However, I think the attribution in set.mm is fine, even if as a result the visualization is a bit strange, and the cure may be worse than the (small) problem.

benjub:
Another glitch is the presence of Loof Lirpa among the contributors.  To be honest, I'm not sure the section "Humor" has its place in set.mm.

david-a-wheeler:
@benjub - I noticed that too, but I liked it so I left it in. The world has enough sadness, I think brightness like that is fun.

PR #1141, opened 2019-10-02
Fix credits for nnacli, nnmcli, and omopth

david-a-wheeler:
Fix credits for nnacli, nnmcli, and omopth. These are credited to
Mario, but on review of commit f9ba34347e012f6fbbbcd5a39eee0a4aae57b3c9
(stable release, archive name set.mm.2012-11-04.last-before-grp2grpop)
I believe these should have been credited to Norman Megill.


david-a-wheeler:
@digama0 @nmegill - I believe there's an error in the credits and that this fixes it.
Justification is in the commit log.

If my correction is wrong, let me know!

nmegill:
Actually both of the dates apply to Scott Fenton.

In set.mm.2012-05-21, nnacli is in Scott's mathbox with 2 dates, 11-Apr-2012 and 28-Mar-2012.  One reason for 2 dates was that people would sometimes add a 2nd date when submitting their mathbox, to make the theorem bubble to the top of Most Recent Proofs.  You might say it was created on the earlier date and the mathbox was added to set.mm on the later date.

Compounding this problem, for a while metamath.exe would add or update the later date upon 'save new_proof'.  This feature was later removed.

At one point, not sure when, I went through a big purge where if two dates were "too close" (say a month or two), I removed the later date, in order to clean up the damage cause by the 'save new_proof' feature.  This was done mostly manually, maybe with a little automated assistance to identify them, and it's possible I may have missed some.

In principle, 28-Mar-2012 would be Scott's "Contributed by" date and 11-Apr-2012 would be his "Revised by" date.   However, in this case and whenever the dates are very close like this, I would make it just "(Contributed by Scott Fenton, 28-Mar-2012.)" and simply delete the second date.  Especially if the theorem is or used to be in a mathbox and the dates are close, it would be very rare that I would revise the theorem, and even if I did it's pretty inconsequential at this point.

david-a-wheeler:
@nmegill - okay!  I'll make a different pull request that simply removes the too-close-in-time revision.

david-a-wheeler:
Please look at pull request https://github.com/metamath/set.mm/pull/1143 which resolves the problem in the way recommended by Norm.

benjub:
Given the comment explaining the closing of this PR (as subsumed by #1143), I also delete the branch.

PR #1145, opened 2019-10-03
shorten 2eu6

wlammen:
I'd rather see the proof of 2eu6 based directly on df-eu, instead via 2eu4. I suspect a recursive scheme would be visible then, indicating how to arrive at even higher dimensions. I occasionally explore this idea, and as a by-catch, optimizations of the current proof pop into my mind.

I am pretty sure that numerous people have proven a theorem like 2eu6 more or less the same way I did, but I don't know.  I do not use textbooks or literature for my proofs.  I usually even avoid looking at current proofs in Metamath, so I can explore ideas completely unbiased.  This imposes a particular problem. The following citation is publicly visible in mmset.html:

> Reading Suggestions    Logic and set theory provide a foundation for all of mathematics. One possible way to start to learn about them is to use the Metamath Proof Explorer in conjunction with one or more textbooks listed in the Bibliography of the next section.

Since my proofs are not taken from published sources, they may not be well received by students working along some learning material, as indicated in the quote.  Any suggestions?  No problem at all?  Should there be a proof layout paragraph in case of longer proofs?


benjub:
@wlammen you're right: sometime, a proof which has been over-optimized for length might be shorter, but less systematic and less straightforward to understand than the standard and almost mechanical proof.  In that case, you may relabel the standard proof as ALT instead of OLD.  I let you and @nmegill judge when it is necessary.

nmegill:
> I am pretty sure that numerous people have proven a theorem like 2eu6 more or less the same way I did, but I don't know.

I have never seen double or higher uniqueness stated correctly with symbols in a published work, much less proved formally.  The only one I have found stated with symbols is in Takeuti-Zaring, incorrectly using E! x E! y ph (see 2eu5 comment).  Otherwise, I've seen it stated only in informal English, such as "the unique x and y such that...", without a precise definition.

 > I do not use textbooks or literature for my proofs. I usually even avoid looking at current proofs in Metamath, so I can explore ideas completely unbiased. This imposes a particular problem. The following citation is publicly visible in mmset.html:
 
> > Reading Suggestions    Logic and set theory provide a foundation for all of mathematics. One possible way to start to learn about them is to use the Metamath Proof Explorer in conjunction with one or more textbooks listed in the Bibliography of the next section.

What I intended by this is that the theorem statements can often be compared to the literature in order to become familiar with the notation used by both.  Each offers advantages the other doesn't have - the literature provides motivation, and Metamath provides complete proofs for someone who might be stuck.

I didn't mean that the proofs should be compared, because typically that's not possible anyway.  The Metamath proofs are usually, if not almost always, different from the literature proofs, if the literature even gives a proof, and certainly formal proofs are quite rare.

> Since my proofs are not taken from published sources, they may not be well received by students working along some learning material, as indicated in the quote. Any suggestions? No problem at all? Should there be a proof layout paragraph in case of longer proofs?

Very long proofs are typically partitioned into lemmas, and often the description for each lemma corresponds to a sentence or "step" in the textbook version.   Metamath can be consulted to see the details, which often aren't provided in the textbook version.  I think this has worked out well for many proofs.  A dry and boring verbose breakdown describing the Metamath proof details in English might even be counterproductive since few would read it and the important things might be overlooked.  Sometimes it is useful to describe one or more key or important steps or techniques in a long proof; but to what extent is highly subjective, and we leave it up to the person writing the proof to describe in the comment what if anything about the proof is important.

PR #1149, opened 2019-10-03
Add basic infastructure for "people/" subdirectory

david-a-wheeler:
Add the basics for a new "people/" subdirectory, which
will store public avatars of contributors (where the contributor
has indicated that such a public avatar is okay).

The immediate use is for "Gource" visualizations, but there
may other occasions where an avatar may be useful.

Note: Avatars are an image of the contributor's choosing;
they do not need to be a face (and often aren't).
We do not require contributors to give their real names, to respect
people's privacy; contributors may use a pseudonym of their choosing.
Of course, we may choose to not accept a pseudonym or avatar
for our purposes :-).

### DIFF
--- a/people/000-README
+++ b/people/000-README
@@ -1,0 +1,1 @@
+Please modify the "urls" file to record the source of a new avatar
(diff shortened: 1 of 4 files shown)


david-a-wheeler:
@nmegill - hopefully you'll accept this.

The next step would be to add the actual current avatars.  This doesn't take a lot of space, and would mean that the avatars would keep working no matter what.

PR #1151, opened 2019-10-03
Add the actual avatar images to the people/ directory

david-a-wheeler:


david-a-wheeler:
This is the commit that actually adds the images (not just instructions to load the images).

The advantage of doing this is that even when websites disappear or reorganize the images will still be available.  I waited a few days because I wanted to see if there were any objections, but I've tried to ensure that the images are either already publicly available or were provided by the contributor (and thus are okay).

PR #1152, opened 2019-10-03
Misc

benjub:
Miscellaneous theorems in mathbox.

@digama0 @jkingdon : Is what I just wrote in the head comment of my mathbox in iset.mm (regarding bounded formulas) correct ?  Do you think the current axioms are enough to prove that [ x / y ] ph is bounded when ph is ?  It looks like not and that we are going to posit it as an axiom.

@jkingdon : The definition of the truth constant TRUE has been changed recently in set.mm.  Can you replicate the change in iset.mm ?

@nmegill @digama0 : Is what I wrote in the comment of bj-1 a few days ago (http://us2.metamath.org/mpeuni/bj-1.html) correct ?

Thanks !


benjub:
Note: Travis failed because of the top date in iset.mm, which I'm going to fix, but there are also many strange warnings...

digama0:
@benjub The parenthetical in bj-1 is not correct; weq contains proper nonatomic subformulas, namely "class x". That is, it's possible to never derive "class x" directly in a proof that uses weq, even though it should properly appear in the parses of formulas in the statement. This is a general fact about syntax theorems. (You should think of "cv x" as being a subterm of the expression, even though the "cv" is silent.)

A downside of syntax theorems is that they mean that a parse tree cannot be identified with a syntax proof. If I use idi twice to derive `|- x = y` from `|- x = y` from `|- x = y`, where in the first application the syntax proof is `vx vy weq` and in the second application it is `vx cv vy cv wceq`, this is nevertheless a valid proof, even though the two theorems disagree on the parse of the middle statement. It is for this reason that I have generally tried not to use syntax theorems in parsers, and MM0 has no such facility.

nmegill:
> @nmegill @digama0 : Is what I wrote in the comment of bj-1 a few days ago (http://us2.metamath.org/mpeuni/bj-1.html) correct ?

It's correct that bj-0 saves only 1 step in this case, but that's mainly because bj-0 doesn't do much.

If you make bj-1 a very complex instance of id with ph the only wff variable, and prove the corresponding bj-0, the bj-0 might have a very long proof, whereas bj-1 would have 3 steps no matter what.
```
1     wph=wph   $f wff ph
2   wph=bj-0 $p wff [very complex expression with ph]
3 bj-1=id  $p |- ( [very complex expression with ph] -> [very complex expression with ph] )
```

I'm not sure if this addresses what you are asking.  There are other reasons syntactic proofs aren't useful.  A syntactic proof is rather short, and only needs to be proved once in a compressed non-syntactic proof.  Indeed, a compressed proof takes advantage of sub-wffs occurring during the construction of the the big wff (with a separate syntactic proof), and you would still have to repeat the sub-wff constructions anyway, so a syntactic proof in an ordinary proof might save nothing in the compressed proof.  The bj-1 example is contrived and doesn't manipulate sub-wffs throughout like an ordinary proof would.

More importantly, though, I don't think we want to encourage widespread use of them because it will likely just cause confusion, with very tiny savings if any.  Indeed, so far we have restricted them to weq and wel for the special purpose explained in their comments, and those are confusing enough. :)

benjub:
Thanks @digama0 and @nmegill , I will ponder your comments and fix my text accordingly. Note that the comment of bj-1 goes in your sense, Norm, since it says "This explains why syntactic theorems are not useful in set.mm."  The goal of my comment is precisely to explain why.

digama0:
Regarding `Bdd`: You don't need all those axioms. I would recommend:

    |- Bdd x e. y
    $d x y |- Bdd ph -> Bdd A. x e. y ph
    |- ( ( Bdd ph /\ Bdd ps ) -> Bdd ( ph -> ps ) )
    |- ( Bdd ph -> Bdd -. ph )
    ( ph <-> ps ) |- ( Bdd ph -> Bdd ps )

You should be able to derive everything else: boundedness of the other propositional connectives (true, false, and, or, iff), and boundedness of bounded exists.

As for substitution, I think you are right that it's not derivable (it is derivable for every concrete choice of `ph`), but it's not clear to me that this is the best place to put it. `Bdd` is not really a predicate, and we don't want to pack all sorts of logic into it. If the point is to have a precondition for a separation axiom, then you can do the logic after you've constructed the set of interest. An equality bound is the same thing as a bound like `A. x e. { y } ph`, and there are several other kinds of quantifier like this that you might want to have, but I don't think they are all admissible.

digama0:
> Finally, note that the axioms have to be written in closed form, rather than
as inferences, since formulas may have free variables and be semantically
bounded for some values (but not all) of these variables.

This is not true. You should think of `Bdd ph` as binding all variables in `ph`. That is:

    |- F/ x Bdd ph

with no side conditions. It basically takes as input an n-ary predicate `ph(x,y,z,...)` and tells you if every variable in an atomic subformula of `ph` is derived from the input variables `x,y,z,...`. This is what ensures predicativity.

Actually, it's not clear to me whether there is a good reason to have `Bdd` be available not just as an inference, but I don't think it hurts to do so. It's actually pretty easy to see, based on the axioms that I gave above (yours would also do for this), that the only way to derive a `Bdd` is using other `Bdd` theorems, and moreover there is no way to sneak information under a `Bdd` other than by using the equivalence rule, which does not have a context. More formally:

> If `ph`, `ps` are formulas such that `|- ( ph -> Bdd ps )` is derivable, then either `|- -. ph`, or `|- ( ( Bdd ph1 /\ ... /\ Bdd phn ) -> Bdd ps )` is also derivable for some formulas `ph1` ... `phn`.

(This can be proven using the cut-elimination theorem.)

jkingdon:
@benjub I've submitted the T. one at https://github.com/metamath/set.mm/issues/1153 . There's no reason to why set.mm and iset.mm need to be different here, so hopefully someone will get around to it before long.

As for boundedness, I think that's covered well above.

PR #1155, opened 2019-10-04
more real numbers in iset.mm

jkingdon:
These are the next few theorems, taken from set.mm and modestly intuitionized.

After this, well, it is building up to closure of positive real addition and multiplication.

One hurdle which will need to be tackled (again in a future pull request) will be the converse of http://us2.metamath.org:88/ileuni/prop.html (or something sorta along those lines, the point is we don't quite have the machinery to handle the analogue to the reverse direction of elnp in http://us2.metamath.org:88/mpeuni/genpcl.html ). Seems like if ` ( 1st `` A ) ` is inhabited, then there must be an ordered pair which .... well I haven't fully fleshed out the argument and there don't seem to be obvious jumping off points in set.mm, but it seems like something should be possible here.

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -59142,10 +59142,15 @@ $)
 
   ${
     $d x y q r s $.
-    $( Define addition on positive reals.  This is a "temporary" set used in
-       the construction of complex numbers, and is intended to be used only by
-       the construction.  From Section 11.2.1 of [HoTT], p.  (varies).
-       (Contributed by Jim Kingdon, 26-Sep-2019.) $)
+    $( Define addition on positive reals.  From Section 11.2.1 of [HoTT], p.
+       (varies).  We write this definition to closely resemble the definition
+       in HoTT although some of the conditions (for example, ` r e. Q. ` and
+       ` r e. ( 1st `` x ) ` ) conditions are redundant and can be simplified
+       as shown at ~ genpdf .
+
+       This is a "temporary" set used in the construction of complex numbers,
+       and is intended to be used only by the construction.  (Contributed by
+       Jim Kingdon, 26-Sep-2019.) $)
     df-iplp $a |- +P. = ( x e. P. , y e. P. |->
       <. { q e. Q. | E. r e. Q. E. s e. Q. ( r e. ( 1st ` x ) /\
         s e. ( 1st ` y ) /\ q = ( r +Q s ) ) } ,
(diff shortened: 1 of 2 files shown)


digama0:
I'm afraid I don't understand the sticking point. It seems like the converse of `prop` is trivial: `P.` is an ordered pair abstraction, hence it is a relation, hence every element of it satisfies theorem `1st2nd`.

jkingdon:
> I'm afraid I don't understand the sticking point. It seems like the converse of `prop` is trivial: `P.` is an ordered pair abstraction, hence it is a relation, hence every element of it satisfies theorem `1st2nd`.

We'd get prop itself, not the converse, from `1st2nd`, wouldn't we? I don't think the converse is a huge sticking point (for one thing, http://us2.metamath.org:88/ileuni/df-iplp.html produces ordered pairs, and proving closure of addition and multiplication is the goal of all this), but I haven't seen a trivial solution and I'm not sure whether http://us2.metamath.org:88/ileuni/elinp.html (where `L` is substituted by `( 1st `` A )`` and `U` is substituted by `( 2nd `` A )`) is what we'll need as-is or whether the right hand side needs a condition added (in the case where we want the left hand side to be `A e. P.`, that is).

jkingdon:
As for the current pull request, I pushed up a fix to discouraged comments to make travis happy. So this should be ready to merge unless someone sees a problem with it.

digama0:
@jkingdon I see, you want to derive from the fact that ``<. ( 1st ` P ) , ( 2nd ` P ) >. e. R`` (where `R` is a relation) that ``<. ( 1st ` P ) , ( 2nd ` P ) >. = P``? I think that's not going to be possible. It's not even true classically because `P` could be a proper class, in which case the LHS evaluates to `<. (/) , (/) >.` and the RHS to `P`. Why do you want this? These kinds of reverse closure rules are almost never true in iset.mm because ndmfv can't be used in contrapositive as in set.mm.

digama0:
You say you are going for the reverse direction of enlp, but you changed the definition of `P.` so it doesn't hold as stated anymore; you have to talk about the right side of the cut as well. The analogue of enlp would be what you are calling enilp, and it appears to have been no trouble to prove.

jkingdon:
> Why do you want this?

The context is that we'll be wanting to prove ` ( A e. P /\ B e. P. ) -> ( A F B ) e. P. ` . The piece related to `1st` and `2nd` is already solved on the https://github.com/jkingdon/set.mm/tree/iset-npcl branch and the solution is

```
  ${
    $d l u q r A $.
    $( Membership in positive reals, using ` 1st ` and ` 2nd ` to refer to the
       lower and upper cut.  (Contributed by Jim Kingdon, 3-Oct-2019.) $)
    elnp1st2nd $p |- ( A e. P. <-> ( (
        A e. ( ~P Q. X. ~P Q. ) /\
        ( E. q e. Q. q e. ( 1st ` A ) /\ E. r e. Q. r e. ( 2nd ` A ) ) ) /\ (
        ( A. q e. Q. ( q e. ( 1st ` A ) <->
            E. r e. Q. ( q <Q r /\ r e. ( 1st ` A ) ) ) /\
          A. r e. Q. ( r e. ( 2nd ` A ) <->
            E. q e. Q. ( q <Q r /\ q e. ( 2nd ` A ) ) ) ) /\
        A. q e. Q. -. ( q e. ( 1st ` A ) /\ q e. ( 2nd ` A ) ) /\
        A. q e. Q. A. r e. Q. ( q <Q r ->
          ( q e. ( 1st ` A ) \/ r e. ( 2nd ` A ) ) )
        ) ) ) $=
```

where the key thing is adding `A e. ( ~P Q. X. ~P Q. )` to the right hand side. This goes with
```
genpelpw $p |- ( ( A e. P. /\ B e. P. ) ->
        ( A F B ) e. ( ~P Q. X. ~P Q. ) ) $=
```
(given the usual genp* hypotheses) which is also proved on that branch.

So the other pieces of `elnp1st2nd` are inhabited, rounded, disjoint, and located (with the first two having separate conjuncts for lower and upper cuts). So far that branch has inhabited for lower cuts:
```
    genpml $p |- ( ( A e. P. /\ B e. P. ) ->
        E. q e. Q. q e. ( 1st ` ( A F B ) ) ) $=
```
(given those same genp* hypotheses) and I'm now working on rounded. After that will be disjoint and located and the upper cut versions of inhabited and rounded (straightforward once we have the lower cut versions).

jkingdon:
The work described above, plus roundedness, is at #1192 .

PR #1165, opened 2019-10-04
Update iset.mm to define tru the same way as set.mm

### LINKED ISSUE #1153: update iset.mm definition of T. (true)
In July 2019 set.mm changed to ⊤ ↔ (∀𝑥 𝑥 = 𝑥 → ∀𝑥 𝑥 = 𝑥 . Bring this definition over to iset.mm and whatever else is required to get it to work.

jkingdon:
There's no reason to diverge here, and set.mm material can be copied over with very minor updates.

@benjub I found this as best as I could from latest set.mm and git. But speak up if you see anything missing here. Hopefully this will unblock whatever CZF issues you had with this definition.

Fixes #1153

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -38,6 +38,7 @@
 "alrimih" is used by "sb6rf".
 "alrimih" is used by "sbbid".
 "ax-11o" is used by "ax11b".
+"df-tru" is used by "tru".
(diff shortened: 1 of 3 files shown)


benjub:
Thanks @jkingdon (I think regenerating "discouraged" files will be enough to make Travis pass).

PR #1167, opened 2019-10-05
Use 'finds' for Metamath 100 #74, fixes #1159

### LINKED ISSUE #1159: Metamath 100 proof #74 (2/100) - use nnind by Norman Megill, not findes by Raph Levien
We need to straighten out the entry of Metamath 100 proof # 74 (2/100), that is, the second proven proof in the Metamath 100. I think we should refer to nnind everywhere, and credit Norman Megill for it.  In notes we can note that there is a related proof by Raph Levien, findes, that was later contributed on 9-Jul-03.

This is part of an overall careful review of the Metamath 100 (I've found 6 things that are odd about the author data I need to check through, and this is the first one).

@nmegill - do you object to this change?  Here are the details.  I can make a proposed fix, but I first want to confirm that this is the right thing to do.

Currently we have a muddle, where nnind is referenced & is claimed to be done by Raph, which isn't quite true.

In https://docs.google.com/spreadsheets/d/1jcLOp_jF4sPrVPdPedL_ZWebXp8oYEQOSmbWb168YHE/edit#gid=0
It is currently listed as this:
1997-01-10 | 2 |   | 74 | The Principle of Mathematical Induction | Raph Levien | http://us.metamath.org/mpegif/nnind.html |  
-- | -- | -- | -- | -- | -- | -- | --

And in the page http://us.metamath.org/mm_100.html it says;
> 74. The Principle of Mathematical Induction (findes, by Raph Levien, 2003-07-09) - there are many versions of Mathematical Induction in set.mm; another version is nnind (by Norman Megill, 1997-01-10).

But when I do a git checkout ba18c14ef9713308a7efe1dfbc8049975f52f56a (archive name set.mm.2007-04-29-grothprim)

We see:

~~~~
  ${
    $d x y z $.  $d y z ph $.
    findes.1 $e |- [ (/) / x ] ph $.
    findes.2 $e |- ( x e. om -> ( ph -> [ suc x / x ] ph ) ) $.
    $( Finite induction with explicit substitution.  The first hypothesis is
       the basis and the second is the induction hypothesis.  Theorem Schema
       22 of [Suppes] p. 136. (Contributed by Raph Levien, 9-Jul-03.) $)
    findes $p |- ( x e. om -> ph ) $= .... $.      $( [29-Jul-2003] $)
  $}
~~~~

~~~~
    $( Principle of Mathematical Induction (inference schema).  The first
       four hypotheses give us the substitution instances we need; the last
       two are the basis and the induction hypothesis.  See ~ nnaddclt for
       an example of its use.  See ~ nn0ind for induction on nonnegative
       integers and ~ uzind , ~ uzind4 for induction on upper partitions of
       integers.  See ~ indstr for strong induction. $)
    nnind $p |- ( A e. NN -> ta ) $= ... $.
      $( [10-Jan-1997] $)
  $}
~~~~

When you checkout the current version you see that nnind was (Revised by Mario Carneiro, 16-Jun-2013.).

david-a-wheeler:
Straighten out peculiarities in the entry for
Metamath 100 proof # 74 (chronologically 2/100), by
using 'finds' (by Norman Megill, 1995-04-14) as the primary proof,
but also noting theorem 'findes' (by Raph Levien, 9-Jul-2003) and
'nnind' (by Norman Megill, 10-Jan-1997).

We had been listing 'findes', but theorem 'finds' came first,
so it makes more sense to list 'finds' instead.

If this commit is accepted, it implies that:
* the Google docs spreadsheet for Metamath 100 will need modification,
  for entry 2 it becomes Norman Megill, 1995-04-14, 'finds'.
* Freek will need to be told that in his list entry #74 needs
  'Raph Levien' changed to 'Norman Megill'.
If this commit is accepted I plan to do that.

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -434,10 +434,14 @@ by Mario Carneiro, 2015-01-28)</li>
 
 <!-- 2nd added to list -->
 <li><a name="74">74</a>.  The Principle of Mathematical Induction (<a
-href="mpeuni/findes.html">findes</a>, by Raph Levien, 2003-07-09) - there
-are many versions of Mathematical Induction in set.mm; another version
-is <a href="mpeuni/nnind.html">nnind</a> (by Norman
-Megill, 1997-01-10).</li>
+href="mpeuni/finds.html">finds</a>, by Norman Megill, 1995-04-14).
+There are many versions of Mathematical Induction in set.mm. Theorem <a
+href="mpeuni/findes.html">findes</a> (by Raph Levien, 9-Jul-2003)
+is a compact and more traditional-looking version with explicit substitution.
+Another alternative is <a
+href="mpeuni/nnind.html">nnind</a> (by Norman Megill, 10-Jan-1997),
+which uses natural numbers instead of ordinal numbers and might be
+less obscure for non-mathematicians.</li>
 
 <!-- 36th added to list -->
 <li><a name="75">75</a>. The Mean Value Theorem (<a


nmegill:
> * Freek will need to be told that in his list entry #74 needs
>   'Raph Levien' changed to 'Norman Megill'.
>   If this commit is accepted I plan to do that.

I think bothering Freek with such minutia may annoy him.  It could be mentioned the next time we have a new theorem to add.

PR #1185, opened 2019-10-08
Misc

benjub:
@nmegill @digama0 I added a short description for half and full adders. Are you fine with this version ? (I think that some explanation is necessary, but that it is not necessary to write an algorithm in pseudocode.)

@jkingdon In the course of proving that singletons and pairs exist, I had to add the theorem nfrex which is in set.mm.  You can move it to the main part if you think it's ok. Also, there are several theorems using snexgOLD and prexgOLD. It should probably be replaced with snexg/prexg in the proofs, but since I was not sure, I let it as is.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18401,7 +18401,7 @@ Proof modification of "bj-axrep3" is discouraged (122 steps).
 Proof modification of "bj-axrep4" is discouraged (130 steps).
 Proof modification of "bj-axrep5" is discouraged (125 steps).
 Proof modification of "bj-axsep" is discouraged (135 steps).
-Proof modification of "bj-axsep2" is discouraged (68 steps).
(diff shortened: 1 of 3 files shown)


jkingdon:
> I had to add the theorem nfrex which is in set.mm. You can move it to the main part if you think it's ok

The main part already has nfrexxy and nfrexya which are the same but with more distinct variable constraints than in set.mm. I've submitted #1186 to add these and various others to the list of "missing" (present in set.mm but not in iset.mm) theorems.

> Also, there are several theorems using snexgOLD and prexgOLD. It should probably be replaced with snexg/prexg in the proofs

Where "several" is 29 for `snexgOLD` and 14 for `prexgOLD`. Yes, I would like to see all of those replaced with snexg/prexg.

PR #1187, opened 2019-10-09
Misc

benjub:
@jkingdon What is your opinion about my comment of ~bj-df-test ?  Also, I would move all the theorems in the "misc" section of my mathbox to the main part if you agree, or you can do it if you prefer.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -37,7 +37,7 @@
 "alrimih" is used by "sb4or".
 "alrimih" is used by "sb6rf".
 "alrimih" is used by "sbbid".
-"ax-11o" is used by "ax11b".
+"ax-10o" is used by "ax10".
 "df-tru" is used by "tru".
 "equsalh" is used by "dvelimALT".
 "equsalh" is used by "dvelimfALT2".
@@ -142,8 +142,10 @@ New usage of "19.9hd" is discouraged (1 uses).
 New usage of "4syl" is discouraged (6 uses).
 New usage of "a9evsep" is discouraged (1 uses).
 New usage of "alrimih" is discouraged (25 uses).
-New usage of "ax-11o" is discouraged (1 uses).
+New usage of "ax-10o" is discouraged (1 uses).
+New usage of "ax-11o" is discouraged (0 uses).
 New usage of "ax-16" is discouraged (0 uses).
+New usage of "ax10" is discouraged (0 uses).
 New usage of "ax9vsep" is discouraged (0 uses).
 New usage of "axnul" is discouraged (0 uses).
 New usage of "bdnthALT" is discouraged (0 uses).
--- a/iset.mm
+++ b/iset.mm
@@ -60688,6 +60688,42 @@ $)
 
 $(
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+                 Miscellaneous
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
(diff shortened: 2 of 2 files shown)


jkingdon:
> What is your opinion about my comment of ~bj-df-test ?

I'd be fine with replacing `TEST ph` with `DECID ~ ph` although I suppose we should see whether @david-a-wheeler has an opinion on the topic.

Even `DECID` is something I'm kind of on the fence about. I think the only "real" theorem (that is, a theorem which isn't about decidability itself) we prove with it is http://us.metamath.org/ileuni/elni2.html , and it isn't clear that `DECID ph` works any better than `ph \/ -. ph` for this purpose.

>  I would move all the theorems in the "misc" section of my mathbox to the main part if you agree

Looks like this is bj-notnotdc1 , bj-notnotdc , and bj-dcdc (plus whatever we do about `TEST`).

Unless we would use these in proofs outside your mathbox I'm not sure where to draw the line between "these are important/interesting and go in the main part for this reason" or "mathboxes can have theorems of general interest, the real threshold for moving them out is when we have reason to use them outside the mathbox they are in". Perhaps @nmegill or @david-a-wheeler has a guideline to point to.

digama0:
In lean, decidability is used as a hypothesis in all the "missing" theorems of propositional logic. Stuff like peirce's law, case analysis, double negation elimination and proof by contradiction; De Morgan laws, and/or distribution theorems. These are useful in general purpose proving contexts, and can be satisfied when the proposition is, well, decidable; generally anything involving countable structures will be decidable here.

It also propagates into several algebraic structures; for example, you can define the max function on any decidable linear order, where a decidable order is one where `A. x A. y DECID (x <= y)`.

jkingdon:
> These are useful in general purpose proving contexts, and can be satisfied when the proposition is, well, decidable; generally anything involving countable structures

OK, cool, that stuff can stay. For example, we have http://us2.metamath.org:88/ileuni/enqdc.html . And it is most easily expressed using `DECID`.

> a decidable order is one where A. x A. y DECID (x <= y)

That raises a question I've been wondering about. We usually express trichotomy as `x < y \/ x = y \/ y < x` (for example http://us2.metamath.org:88/ileuni/nqtri3or.html ). And that implies either `x <= y \/ y <= x` (which I'm not sure we've done much with) and `x < y <-> -. ( x = y \/ y < x )` (e.g. http://us2.metamath.org:88/ileuni/nqtric.html ). But do the latter two have the usual consequences of trichotomy? For example, analogoues to http://us2.metamath.org:88/ileuni/ordtriexmid.html ? Last time I tried to prove " If you apply onint to a pair you can derive totality of the order." I got stuck at the stage of showing that 'A C_ B \/ B C_ A` for ordinals implies excluded middle, although it seems like it should be straightforward.

benjub:
I agree that DECID should stay (I never proposed its deletion).  For another example, it could be used to state transparently the axiom for Delta_0-classical logic: ` |- Bdd ph => |- DECID ph ` .

As for TEST, it is definitely less used (and less known) than DECID, and more importantly, having a definition for TEST only saves 1 token ( ` TEST ph ` versus ` DECID -. ph ` ) so actually ` TEST ` is really an abbreviation for ` DECID -. ` , while DECID is useful since it saves n+3 tokens if the formula ph under consideration has n tokens. Furthermore, TEST would require the reader to know one more definition (and a much less known one than that of DECID). For instance, far more readers would readily understand ` |- ( DECID ph -> DECID -. ph ) ` than ` |- ( DECID ph -> TEST ph ) ` .

david-a-wheeler:
I created TEST because "testability" is an important concept discussed all by itself in intuitive logic.  In particular, some theorems only require testability  ( TEST ph ), not full decidability ( DECID ph ).  The same is true for stability, too.

If it's determined that testability doesn't necessarily warrant its own definition, and some other notational convention should be used instead, then I think it's really important to *document* that convention *and* provide proofs that the convention "does what it's supposed to do".

PR #1189, opened 2019-10-10
mmset.raw.html: Add getting started link

david-a-wheeler:
Nowhere on the set.mm front page do we provide a hint of
how to start contributing something else.  That should be fixed.

This adds a single link on the top right to the GitHub page on
"Getting started with contributing (to Metamath set.mm)".
That will at least help people find that information.

### DIFF
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -312,7 +312,7 @@ SIZE=-1><I>12-Apr-2008</I></FONT>
 </LI>
 
 <LI>
-<A HREF="https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing">Getting started with contributing [to Metamath set.mm]</A>
+<A HREF="../index.html#contribute">How can I contribute to Metamath?</A>
 </LI>
 
 <!-- <LI> <A HREF="mmhil.html">Hilbert Space Explorer Home Page</A></LI> -->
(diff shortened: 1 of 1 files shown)


nmegill:
We already have a link at the top on the page saying, "If you wish to contribute your own proofs to the Metamath project, see 'How can I contribute to Metamath?'"  Is that not sufficient?

benjub:
In addition to the link 'How can I contribute to Metamath?' that @nmgeill
added recently, I also added recently the link "GitHub repository (contains
the database as well as help on contributing)" (and I created the "External
links" list for that purpose).

nmegill:
@david-a-wheeler Given the comments above, do you think this is still necessary?

Also, if we do add a your link, it should go in the "External links" section because it directs to a site outside of metamath.org.

david-a-wheeler:
@nmegill - I didn't notice that sentence earlier, which suggests to me that perhaps it's not enough :-).  I think the link of major links should also note it.  You're right that it should be an "external links" entry if we use my text.

However, now that I see "How can I contribute to Metamath?", I think we should simply add *that* in our list of big links.

nmegill:
OK, I suggest you use the "How can I contribute to Metamath?" link to replace the link you added.  It can stay in the "Related pages" area since it links to a metamath.org page.

PR #1207, opened 2019-10-19
PLEASE MERGE - Update travis from trusty (Ubuntu 14) to xenial (16)

david-a-wheeler:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ jobs:
   include:
     ###########################################################
     - # smetamath-rs (smm3), a Rust verifier by Stefan O'Rear.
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
This attempts to update to a newer version of Ubuntu (16 instead of 14) in our Travis verifications.

david-a-wheeler:
Good news, I've updated our Travis verifications to run on Ubuntu 16 instead of Ubuntu 14.  Ubuntu 14 is old, so it's really time to upgrade.  Eventually we should update to Ubuntu 18, but this is a good step and should make the transition to Ubuntu 18 easier.

PR #1209, opened 2019-10-20
Fix 'quad3' to replace ax1cn with ax-1cn and rewrap

david-a-wheeler:
Theorem ax1cn is discouraged, which causes discouragement checks to fail.
There's no reason to use ax1cn in this case, so change it to ax-1cn.


benjub:
By the way, @david-a-wheeler : do you know how/when ~quad3 arrived in set.mm ? It has "Contributed... 19-Oct-2019" but I do not see a pull request corresponding to it ?

nmegill:
Good.  Sorry about that. :)

nmegill:
@benjub It was intended as a quick update to FC's mathbox, and I didn't do a PR since it wasn't expected to affect anyone else.  (The SSD on my PC is almost full, and I have to keep making room for things like another fork.  Hopefully it will be addressed soon.)

PR #1211, opened 2019-10-20
READY FOR MERGE: If regen-from-raw finds warnings, report as a failure

david-a-wheeler:

### DIFF
--- a/mmdeduction.raw.html
+++ b/mmdeduction.raw.html
@@ -7,7 +7,7 @@
 <META NAME="viewport" CONTENT="width=device-width, initial-scale=1.0">
 
 <META HTTP-EQUIV="Content-Type"
- CONTENT="text/html; charset=iso-8859-1">
+ CONTENT="text/html; charset=UTF-8">
  <TITLE>The Weak Deduction Theorem - Metamath Proof Explorer</TITLE>
 <LINK REL="shortcut icon" HREF="favicon.ico" TYPE="image/x-icon">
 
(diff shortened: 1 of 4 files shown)


david-a-wheeler:
This fixes our detection mechanism - it now reports failures on problems it didn't report before.  Now we need to fix the problems it reports...!

david-a-wheeler:
I fixed our mechanism for detecting problems when generating HTML from raw html files, and then fixed all the new errors it was reporting.  In mmil.raw.html I changed "prnadd: prnaddl" (which is clearly wrong) to "prnmadd: prnmaddl" (which I think is what was intended).  The "email quoting" table in mmdeduction.raw.html should in the long term be rewritten, but that would be a lot more work - I instead tweaked it so it at least displays better than before without errors.

benjub:
Maybe it's safer to keep the page mmdeduction.raw.html ascii, in which case, replacing ¬ with &not; or &#172; is enough.

david-a-wheeler:
> Maybe it's safer to keep the page mmdeduction.raw.html ascii, in which case, replacing ¬ with & not ; or & # 172 ; is enough (spaces added to prevent markdown).

I'm not sure that's true any more, since in HTML5 the default character encoding is UTF-8. (This has not always been the case.)

But better safe than sorry. I'll add a commit switch to & not ; - and for good measure, change the character encoding to UTF-8 so that if a UTF-8 character *does* slip through it should be interpreted correctly.

david-a-wheeler:
I think it's all set to go! Let me know of any other issues.

nmegill:
> I'm not sure that's true any more, since in HTML5 the default character encoding is UTF-8. (This has not always been the case.)

The standard we've been using for all web pages is HTML 4.01 Transitional, as indicated in the DOCTYPE at the top of each page.  This does everything that we have needed so far, and I don't see any reason to change as long as validator.w3.org accepts it.

I don't know what 4.01 says about UTF-8, but personally I strongly prefer to have ASCII in the source code so I know what I'm looking at.  I can also type ASCII from the keyboard. :)

Anyway, the most important criterion for me is that our pages pass validator.w3.org.  It has caught many careless errors that would have led to corrupted tables etc. that otherwise would have gone unnoticed .   Most of our pages have a link to it in the lower right corner.  [BTW iset.mm and nf.mm need ALT tags (as in set.mm) in their htmldefs to pass, a low priority thing to fix someday.]

nmegill:
FYI 'markup ... /labels' will treat ~~ as an escape for ~ and convert it to a single tilde.
Your solution is fine too.

jkingdon:
Thanks for fixing mmil.raw.html, @david-a-wheeler . Your fix is indeed correct (if you look at e3d3b29f2dc2ab594bc539688b2fc89e0c9bb34e you'll even see that the git commit message had the correct name, even if mmil.raw.html didn't).

PR #1212, opened 2019-10-20
Misc

benjub:
* quantriv happens to be a duplicate (in a mathbox) of 19.3v ; may I remove it ?

* relabel ax10e --> axc7e (since it is axc7 except for using E.)

* swap positions of nfn and nfnd (to make sure nfn is not proved from nfnd and to put the simpler statement first); it is of course part of the trinity nfnt/nfn/nfnd (closed form/inference/deduction).

* mathbox: added some more closed forms, notably bj-nfext

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1339,7 +1339,6 @@
 "ax12indn" is used by "ax12indi".
 "ax12v2-o" is used by "ax12a2-o".
 "ax16g-o" is used by "ax12inda2".
-"ax1cn" is used by "quad3".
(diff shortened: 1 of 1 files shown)


nmegill:
> * quantriv happens to be a duplicate (in a mathbox) of 19.3v ; may I remove it ?

Yes, and also make the attribution of 19.3v "(Contributed by Anthony Hart, 13-Sep-2011.)".  I wasn't aware of quantriv when I added 19.3v.

PR #1214, opened 2019-10-21
DO NOT MERGE: Try out executing Python with pypy3

david-a-wheeler:


david-a-wheeler:
CPython is really slow. Try out pypy3 and see if that helps.

david-a-wheeler:
I give up.  The docs don't seem to match reality.

PR #1220, opened 2019-10-23
Modify .travis so set.mm verified by mmverify.py

david-a-wheeler:
Python is slow, so to use mmverify.py we'll use 3 jobs:
set.mm pre-mathboxes, set.mm mathboxes, and iset.mm.
I've modified mmverify.py so that it can do this.

With this commit, we now have 5 independent verifiers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,34 +128,46 @@ jobs:
         - ./checkmm set.mm
         - ./checkmm iset.mm
     ###########################################################
-    - # mmverify.py, a Python3 verifier by Raph Levien, for iset.mm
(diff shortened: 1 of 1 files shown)


nmegill:
@david-a-wheeler This PR has a merge conflict that I think is best if you fix it.

Also, we can change the temporary us2 IP address back to us.metamath.org.

david-a-wheeler:
@nmegill - the conflict is *because* of the temporary change to IP addresses. I'll create a pull request to restore the names (instead of IP addresses), once that's accepted it will fix the conflict & you can merge this as well.

david-a-wheeler:
@nmegill .. we can now easily merge this.

PR #1221, opened 2019-10-23
Theorems for matrices consisting of polynomials

avekens:
Main part:
* theorems about monoid/ring homomorphisms

AV's Mathbox:
* The product of two polynomials expressed as group sum of scaled monomials
* Matrices consisting of polynomials: Isomorphism between the ring of matrices having polynomials as entries and the ring of polynomials having matrices as coefficients


avekens:
Something went wrong with the trevis jobs! Could anybody have a look at it?

david-a-wheeler:
The Travis jobs are unusually slow to execute right now. I'm not seeing any particular symptom as to why. Hopefully this is just a transient problem. If the problem persists then we may have to investigate further.

david-a-wheeler:
The jobs took a ridiculously long time, but they completed normally.

avekens:
At least it worked now ;-).

PR #1222, opened 2019-10-23
shorten 2mo

wlammen:
I don't have that much time recently for Metamath, so push this before it gets lost.

A small optimization in 2mo, nothing exciting.


wlammen:
us.metamath.org is not accessible for the verifiction scripts, this is why the check fails.  Icannot do anything about this.

PR #1225, opened 2019-10-25
shorten 2mo

wlammen:
Considerably more proof lines, but mostly elementary steps, so in total this proof is shorter than the recently posted one.  The proof suffers a bit from missing infrastructure wrt double E*.  I suggest to single out (more or less) repeated proof steps as lemmas, working tentatively towards this infrastructure.

There are two more remarks:
1. Expressing exmo in terms of an \\/ might be better readable for beginners, but is less application friendly.  All but two usages need an ori to unpack the \\/, and the remaining two can be readily transscripted without proof lengthening effects.

2. 2exsb should be replaced with E. x E. y ph <-> E. z E. w [ z / x ] [ w / y ] ph, which suits its users best, and is better understandable, to boot.  Maybe the name should be changed to 2sb8e then.


benjub:
> Considerably more proof lines, but mostly elementary steps, so in total this proof is shorter than the recently posted one.

Then maybe keep both, the longer as "ALT" ?

> The proof suffers a bit from missing infrastructure wrt double E*. I suggest to single out (more or less) repeated proof steps as lemmas, working tentatively towards this infrastructure.

Maybe yes.  Even if double E* is probably not used too much.

>     1. Expressing exmo in terms of an \/ might be better readable for beginners, but is less application friendly.  All but two usages need an ori to unpack the \/, and the remaining two can be readily transscripted without proof lengthening effects.

Maybe have both exmo and the version ( -. E. x ph -> E* x ph ) as exmo2 ?

>     2. 2exsb should be replaced with E. x E. y ph <-> E. z E. w [ z / x ] [ w / y ] ph, which suits its users best, and is better understandable, to boot.  Maybe the name should be changed to 2sb8e then.

I agree that this formulation is more readable. The current label looks fine, though. The same is true with exsb, I think.

PR #1226, opened 2019-10-26
introduce 2mo2 

wlammen:
This characterization of double E* is very basic.  Two theorems, 2mo and 2eu4 had it already as an intermediate result within their proof, so factoring it out helps keeping the proofs shorter (saves in total ~ 90 proof bytes), and more comprehensible.


benjub:
uniqeness --> uniqueness
I think 2mo2 is a nice addition.  Will you be able to take into account my remarks in PR #1225 ?

wlammen:
@benjub Thanks for pointing me to the typo.  I read your comment in #1225 , but I am still not quite decided, how far I will take this. There are more already finished changes waiting in the queue.  I will release them one by one in order to allow others to object, if/when they think, I am overdoing it.

nmegill:
I like your 2mo2.  I wonder what other "simple" expressions work for "at most one pair".  For example, for "exactly one pair" we have via 2eu2 to 2eu8 the equivalent expressions:

```
( E! x E. y ph /\ E! y E. x ph )
( E! x E! y ph /\ E! y E. x ph )
( E! x E! y ph /\ A. x E* y ph )
E! x E! y ( E. x ph /\ E. y ph ) )
E! x E! y ( E! x ph /\ E. y ph ) )
E! x E! y ( E. x ph /\ E! y ph ) )
```
Although from a practical point of view these probably have little value, I find them very non-intuitive and thus interesting as puzzles. :)  Also, it seems likely that none of these have been published, including your mo2.

wlammen:
@nmegill 2mo2 is IMO a key theorem to extending "at most" to more dimensions, i.e. for ph(x,y,z...). Note that E. x ph is dependent only on y, z ... The formula thus opens the door to a recursive proof scheme.  So I am a bit skeptic about others not having detected it before.

I already sketched how to proof the 2-dimensional case in a way that it easily extends to more dimensions. This still needs some spare time which my job currently permits in small doses only.

Of the above expressions, I believe, most will carry over to the E* case. You just have to check whether ph <-> F. is possible in the modified formulas.

The expression A. x E* y ph caught my special attention, because it expresses the idea of a function so closely.  This could/should have applications.

PR #1227, opened 2019-10-26
Revert "Temporarily use IP address for DNS name in .travis.yml (#1224)"

david-a-wheeler:
This reverts commit c45d34d850b5421a7fab80f6aaf6f15233a532b7.

We had to temporarily use IP addresses because of a DNS problem.
The DNS problem has been fixed, so restore the names.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,6 @@
 #
 # We typically rewrap the file using scripts/rewrap.
 #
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
@nmegill - it looks like we're ready to revert the raw IP address, so here's the pull request to do that.

PR #1233, opened 2019-10-30
Misc

benjub:
Mainly cleanup work about OLD theorems in iset.mm.  Updated some proofs and statements silently (i.e. no credits); I hope @nmegill and @jkingdon are ok with that.

I began dealing with uses of snexgOLD and prexgOLD, and generalized tpexg and otexg so that the same mishaps do not happen.

@jkingdon : the statements dflim2/df-ilim and dfom3/df-iom are identical.  I don't know what the reason is (it's probably because set.mm uses another definition which is not correct intuitionistically but proves dflim2 to be equivalent to it).  In any case, one of the identical statements should have a "New usage discouraged" tag.  There may be other cases.  A better solution would be to adopt the same intuitionistically correct definition for both set.mm and iset.mm.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -59,8 +59,8 @@
 "mo3h" is used by "moim".
 "mo3h" is used by "moimv".
 "mo3h" is used by "mopick".
-"opelopabsbOLD" is used by "cnvopab".
-"opelopabsbOLD" is used by "inopab".
+"opelopabsbALT" is used by "cnvopab".
+"opelopabsbALT" is used by "inopab".
 "opexgOLD" is used by "elxp4".
 "opexgOLD" is used by "elxp5".
 "opexgOLD" is used by "fliftel".
@@ -69,7 +69,6 @@
 "opexgOLD" is used by "opbrop".
 "opexgOLD" is used by "opeliunxp".
 "opexgOLD" is used by "opswapg".
-"opexgOLD" is used by "otexg".
 "opexgOLD" is used by "otth2".
 "opexgOLD" is used by "relsnop".
 "opexgOLD" is used by "resfunexg".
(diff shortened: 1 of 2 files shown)


jkingdon:
> Mainly cleanup work about OLD theorems in iset.mm. Updated some proofs and statements silently (i.e. no credits); I hope @nmegill and @jkingdon are ok with that.

I haven't been modifying credits for this kind of change. So I'm fine with it.

> I began dealing with uses of snexgOLD and prexgOLD, and generalized tpexg and otexg so that the same mishaps do not happen.

Yup, thanks for this work. That was my intention when I first became aware that I had (because I didn't know better) overused `e. _V` in places.

> @jkingdon : the statements dflim2/df-ilim and dfom3/df-iom are identical. I don't know what the reason is (it's probably because set.mm uses another definition which is not correct intuitionistically but proves dflim2 to be equivalent to it).

Yes, exactly. http://us2.metamath.org/mpeuni/df-lim.html is the set.mm definition. My thinking was that we should have dflim2 under the name `dflim2` in iset.mm because we try to keep the name when the same statement is proved in both set.mm and iset.mm. And that the `$a` itself has to start with `df-`. So if both of those reasons are true, we need both `dflim2` and `df-ilim`.

> In any case, one of the identical statements should have a "New usage discouraged" tag.

Not sure I have an opinion about this one way or the other.

> There may be other cases. A better solution would be to adopt the same intuitionistically correct definition for both set.mm and iset.mm.

In the case of `df-lim` that might work. In the case of `df-om` I suspect it would run into different ideas about how to treat the axiom of infinity (the short summary is that iset.mm relies on the axiom of infinity more, and earlier, than set.mm).

PR #1245, opened 2019-11-11
shorten 2eu1

wlammen:


wlammen:
What I don't understand is why the file discouraged updated (at least it is not in the File changed list), but the Travis check lets this commit pass

PR #1248, opened 2019-11-13
shorten eqneltrrd & neleqtrrd

wlammen:
@jkingdon Most of the intuitionistic logic axioms in set.mm are secured with a "New usage is discouraged" tag, but few are not, like axie2, axi5r. Is this intentional?


jkingdon:
> @jkingdon Most of the intuitionistic logic axioms in set.mm are secured with a "New usage is discouraged" tag, but few are not, like axie2, axi5r. Is this intentional?

I don't have an opinion. If that's what we usually do for axioms, sure.

PR #1249, opened 2019-11-13
Misc

benjub:
iset: following #1233 , I discouraged usage of df-iom and df-ilim in favor of their aliases, and added an explanation for theses aliases in the associated comments.


jkingdon:
> iset: following #1233 , I discouraged usage of df-iom and df-ilim in favor of their aliases, and added an explanation for theses aliases in the associated comments.

Sounds good to me.

PR #1250, opened 2019-11-14
iset/mathbox: omex from axiom of infinity

benjub:


benjub:
It would be nice to have the same policy regarding the top date in set.mm and iset.mm (my preference is to the policy for set.mm).

wlammen:
> (Proof omdification ...

Suggest to fix the typo and update discouraged.

benjub:
Thanks @wlammen .  I'll fix it in my next PR.

jkingdon:
> It would be nice to have the same policy regarding the top date in set.mm and iset.mm (my preference is to the policy for set.mm).

There's a long (arguably longer than necessary) discussion of this topic and related ones at #981 . I'm willing to leave things as they are and not worry about it for a while.

nmegill:
> > It would be nice to have the same policy regarding the top date in set.mm and iset.mm (my preference is to the policy for set.mm).
> 
> There's a long (arguably longer than necessary) discussion of this topic and related ones at #981 . I'm willing to leave things as they are and not worry about it for a while.

Merge conflicts still happen with set.mm because people sometimes overlook the notice at the top of set.mm, or other things happen like 2 PRs that come in almost simultaneously from 2 different time zones.  I have been silently fixing these myself - it isn't a big deal - so it's not something to be too concerned with.  I am fine leaving iset.mm the way it is for now, and I will handle any date-related merge conflicts.  If it becomes too much of a problem for me then we can switch to the set.mm approach.

david-a-wheeler:
Another approach would be to remove the date at the top. Now that we have version control it is not needed.

nmegill:
> Another approach would be to remove the date at the top. Now that we have version control it is not needed.

It is for my benefit.  Whenever I find a set.mm version in an old archive I can do 'head -1 set.mm' and instantly determine its date (directory date stamps are not always reliable).  Since I have committed to maintaining merge conflicts caused by the date when necessary, it shouldn't affect anyone else.

PR #1266, opened 2019-11-23
Geometry : Lower dimension Axiom, existence of perpendicular

tirix:


tirix:
I fixed the definition of the "dimension greater or equal than" relation, so this is good to be merged, if you're Ok too!

PR #1277, opened 2019-11-28
Revision of sections "Polynomial matrices - Basic properties" and "Constant polynomial matrix"

avekens:
* new theorems in main part: ~ ralimdvva , ~ rspcedvd  , ~ ply1sclcl  ,~ matvscl
* Revision of section "Polynomial matrices - Basic properties"
* Revision of section "Constant polynomial matrix"


benjub:
@avekens I recently pushed to iset.mm a more general version of your rspcedvd (see my comment to PR #1275); it is bj-rspgt, which is in closed form and replaces DV conditions with non-freeness hypotheses, and from which rspcedvd can be proved.  You may also find some other theorems I mention in that comment useful.  Feel free to copy them to set.mm and use them.  If you are the one to propose to move them to the main part, @nmegill will be more likely to accept.

PR #1278, opened 2019-11-29
Misc

benjub:
@jkingdon I noticed that iset.mm uses df-iord which is set.mm/dford3 which is currently in a mathbox.  Therefore, I think dford3 should be moved to set.mm/main (noting in its comment that it is the correct intuitionistic definition, or even making it the official definition), and that it should be added to iset.mm as an alias for df-iord and the latter's usage should be discouraged (like I did for df-ilim/dflim2).

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -215,10 +215,13 @@ Proof modification of "bj-nalset" is discouraged (105 steps).
 Proof modification of "bj-nn0suc" is discouraged (148 steps).
 Proof modification of "bj-nn0suc0" is discouraged (129 steps).
 Proof modification of "bj-nnelirr" is discouraged (261 steps).
+Proof modification of "bj-nnen2lp" is discouraged (52 steps).
 Proof modification of "bj-nntrans" is discouraged (222 steps).
+Proof modification of "bj-nntrans2" is discouraged (33 steps).
 Proof modification of "bj-nvel" is discouraged (11 steps).
 Proof modification of "bj-omex" is discouraged (144 steps).
 Proof modification of "bj-peano2" is discouraged (283 steps).
+Proof modification of "bj-peano4" is discouraged (168 steps).
 Proof modification of "bj-prexg" is discouraged (65 steps).
 Proof modification of "bj-snex" is discouraged (12 steps).
 Proof modification of "bj-snexg" is discouraged (22 steps).
--- a/iset.mm
+++ b/iset.mm
@@ -1,4 +1,4 @@
-$( iset.mm - Version of 25-Nov-2019
+$( iset.mm - Version of 28-Nov-2019
 
 Created by Mario Carneiro, starting from the 21-Jan-2015 version of
 set.mm (with updates since then, including copying entire theorems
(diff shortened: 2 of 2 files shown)


benjub:
Not sure why Travis failed: "The job exceeded the maximum log length, and has been terminated."

jkingdon:
I closed and reopened the pull request to re-run Travis. We'll see if it works better the next time.

jkingdon:
> @jkingdon I noticed that iset.mm uses df-iord which is set.mm/dford3 which is currently in a mathbox. Therefore, I think dford3 should be moved to set.mm/main (noting in its comment that it is the correct intuitionistic definition, or even making it the official definition), and that it should be added to iset.mm as an alias for df-iord and the latter's usage should be discouraged (like I did for df-ilim/dflim2).

I have no objection to any of this provided we keep the sentence `Definition of ordinal in [Crosilla], p. "Set-theoretic principles incompatible with intuitionistic logic".`, or at least some cite(s) to sources which indicate this is an/the appropriate definition. If I'm correctly reading that source, this definition is equally applicable to IZF or CZF.

jkingdon:
> Not sure why Travis failed: "The job exceeded the maximum log length, and has been terminated."

Interesting. The failure (at https://travis-ci.org/metamath/set.mm/builds/618407827 and again https://travis-ci.org/metamath/set.mm/builds/618422828 ) is after a bunch of output from `scripts/report-changes.py --gource` step. The last line before `The job exceeded the maximum log length, and has been terminated.` is `1435536000|Mario Carneiro|M|SUPPLEMENTARY MATERIAL (USER'S MATHBOXES)/Mathbox for Norm Megill/hlhil0|3eb9e4`. @david-a-wheeler have you seen this sort of failure? Is `report-changes.py --gource` intended to output that much data to the logfile? Based on the wording of the error message, I assume the large amount of output is the cause of the failure.

david-a-wheeler:
No, I have not seen that error message before. Quite bizarre. I will need to investigate it.

david-a-wheeler:
Yes, report-changes.py --gource is intended to output that much data to the logfile. It never caused problems before, and by showing all that detail it made it easier to see what the problem is in case of a mistake.

Based on the wording of the error message, I also assume the large amount of output is the cause of the failure. I suspect that Travis has reduced the maximum number of lines in the log output, or that the files have increased so much that we've accidentally hit the limit.

I think that to fix it, all we need to do is redirect the output of that command to /dev/null within .travis.yml. The script is checking dates, a side effect of trying to gather the course data is that it detects screwed up dates and errors out on them. Things like February 31.

david-a-wheeler:
Give me a moment, I need to get to a keyboard and I think I can fix this quickly. Sorry about the trouble.

david-a-wheeler:
I just created pull request https://github.com/metamath/set.mm/pull/1279 - which I *think* will fix the problem.

Sorry for the trouble!

nmegill:
@david-a-wheeler Travis still fails.  Is that because the old version gets used for this PR?  Should I merge this anyway?

david-a-wheeler:
@nmegill -yes, this pull request still uses the old Travis configuration.

The safest thing to do is for @benjub to run

~~~~
git checkout develop
git pull
git checkout misc
git merge develop
git push
~~~~

That will trigger Travis with the new configuration, and hopefully everything will then just work.

david-a-wheeler:
Thanks for merging. Sorry for the trouble, I had no idea that there be such a strict log limit. Lesson learned I guess. There shouldn't be any problems like that from here on.

PR #1280, opened 2019-11-30
axc16g, aev: rm dependency on ax-13

wlammen:
I recently stumbled over a theorem in BJ's mathbox (forgot which) while shortening the class equal/not equal section.  He removed a dependency on ax-13 there, and since then I thought how this can be incorporated in the main part without too heavy changes. To give it a head start, I aimed at a low hanging fruit, axc16g and aev, and managed to get this into main at low costs.

I renamed cbvequv to cbvaev, since ae (all equal, I think) is the usual abbreviation for A. x x = y.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -256,6 +256,7 @@
 "4syl" is used by "ackbij2lem3".
 "4syl" is used by "acndom".
 "4syl" is used by "aevlem1".
+"4syl" is used by "aevlem1OLD".
 "4syl" is used by "afv0nbfvbi".
 "4syl" is used by "aomclem6".
 "4syl" is used by "axdc3lem2".
@@ -1358,6 +1359,8 @@
 "axacndlem4" is used by "axacndlem5".
 "axacndlem5" is used by "axacnd".
 "axaddf" is used by "axaddcl".
+"axc11nlemOLD" is used by "aevlem1OLD".
+"axc11nlemOLD" is used by "axc11nOLD".
 "axc4i-o" is used by "aev-o".
 "axc4i-o" is used by "ax12inda2ALT".
 "axc4i-o" is used by "ax12indalem".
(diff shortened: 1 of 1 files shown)


benjub:
Thanks @wlammen , I'm glad my mathbox proved useful.  Indeed, the statements are

bj-axc16g : Remove dependency on ax-13 from axc16g
bj-aev : Remove dependency on ax-13 from aev

The proofs are basically the same:
* for axc16g: I use bj-axc11v which is basically the same theorem as axc112, except for the DV condition: so I think the proof you propose is better, and I should delete bj-axc11v.
* for aev : First, I use bj-aevlem1v which is your new aevlem1, but you added a dv condition; I would keep the older aevlem1 as it is and introduce aevlem1v as a separate theorem, since it has one more dv condition. Second, I use bj-spimvv and other lemmas which I think are worth introducing.

I have tried to be consistent in the section of my mathbox titled "21.29.4.12  Removing dependencies on ax-13 (and  ax-11)" which has a detailed section comment. There, I remove dependencies from ax-13 from many theorems, and provide the technical lemmas to remove these dependencies from most later theorems (since they use these lemmas with dummy variables, which are always disjoint). I would propose @nmegill to move most of them to the main part, but I'm afraid I already know the answer.

wlammen:
@benjub aevlem1 is a lemma used to prove aev, nothing more. Given this sole purpose, I think,
there is no need to keep several versions around. aev is the better replacement for aevlem1 anyway, once it is established.
As for spimvv, I try to use a minimal invasive strategy here. That is I aim at proof replacements only for already existing theorems. Without a couple of applications at hand, I simply expand a theorem like spimvv.

benjub:
> @benjub aevlem1 is a lemma used to prove aev, nothing more. Given this sole purpose, I think,
> there is no need to keep several versions around. aev is the better replacement for aevlem1 anyway, once it is established.

If you're sure, then OK (I haven't checked).

> As for spimvv, I try to use a minimal invasive strategy here. That is I aim at proof replacements only for already existing theorems. Without a couple of applications at hand, I simply expand a theorem like spimvv.

I understand your strategy. Mine was to "semi-automatically" replace all lemmas (say, "xxx") to make ax-13 unnecessary in a trivial way, and then to use bj-xxxv to prove later theorems. Once this is done for the hundred or so technical lemmas, most of the later theorems won't depend on ax-13 automatically. So there is less human thinking involved.  I'll try to post on the google group about that, to have more opinions.

PR #1281, opened 2019-12-01
Provide a simple English definition for poset

david-a-wheeler:
Modify the df-poset comment to provide the full name of posets
(partially ordered sets) and add a simple English definition
for the term "poset".


benjub:
I do not know if it is simpler, but it certainly is less precise. Maybe you can also put a precise definition, like:
> A poset is a set equipped with a partial order, that is, a binary relation which is reflexive, antisymmetric, and transitive.

david-a-wheeler:
@benjub - I'm sure we could improve the definition, my concern was that there wasn't one at all.

I propose working out an improvement in issue https://github.com/metamath/set.mm/issues/1285 , and once the improvement is worked out, create a pull request to implement it.  That way we don't have endless pull requests to tweak that text.

PR #1286, opened 2019-12-01
Add "Reflexive" etc. definition to Wheeler's mathbox

david-a-wheeler:
As an experiment, define "Reflexive" and "Irreflexive"
to the mathbox of David A. Wheeler.
This defines the terms generally, so that there's a formal
definition of these terms somewhere.


david-a-wheeler:
The definition check isn't happy with that, I'll need to see how to revise them.

benjub:
I think it's a good idea.  I have two proposals:

* I would really like that this defines reflexive *binary relations*, that is,
( R Reflexive A <-> ( R C_ ( A X. A ) /\ A. x e. A x R x ) )
(note also the infix notation).

* Maybe it's better to define this as a function (then it does not work for proper classes, but I think it's not a big loss):
Reflexive = ( x e. \_V |-> { r | ( r C_ ( x X. x ) /\ A. y e. x y r y ) } )

Then, one can define
PreOrder = ( x e. \_V |-> ( ( Reflexive \` x ) i^i ( Transitive \` x ) ) )
PartialOrder = ( x e. \_V |-> ( ( PreOrder \` x ) i^i ( Antisymmetric \` x ) ) )
Equivalence = ( x e. \_V |-> ( ( PreOrder \` x ) i^i ( Symmetric \` x ) ) )
etc.

and this makes for very readable definitions, and basically what appears in textbooks.

david-a-wheeler:
@benjub - Thanks! I need to think about it, but either of the syntaxes you noted are better than what I came up with.

I was trying to make it also work for proper classes, but that may be wasted effort.

david-a-wheeler:
(I need to go do "real work" right now, but I plan to return to this later. Other ideas to improve this are welcome.)

tirix:
I also believe this is a very good idea.
Note that there are interesting variable-free definitions (I believe already present in set.mm), like:
` ( R Reflexive A <-> ( R C_ ( A X. A ) /\ ( _I |' A ) C_ R ) ) `
` ( R Transitive A <-> ( R C_ ( A X. A ) /\ ( R o. R ) C_ R ) ) `
` ( R Symmetric A <-> ( R C_ ( A X. A ) /\ `` R C_ R ) ) `
They could be given as alternate definitions, keeping the math book definition as the default one.

By restricting the definitions to proper sets, we will not be able to provide statements about `_I`, `[C.]`, ` tpos `, ` recs `, ` ~~ `, `~<_`, `~<`, `~<_* `, except if restricted to sets. I would not mind.

set.mm
```diff
@@ -520225,6 +520231,62 @@ _modus ponens_ (our ~ ax-mp ), but he later qualifies this.
...
+  $( Extend wff definition to include "Reflexive" applied to a class, which is
+     true iff class R is a reflexive relationship over the set A. See
+     ~ df-reflexive .  (Contributed by David A. Wheeler, 1-Dec-2019.) $)
+  wreflexive $a wff Reflexive A R $.
+
+  ${
+    $d A R x $.
+    $( Define relexive relationship; R is reflexive over the set A iff
+       ` A. x e. A x R x ` .  (Contributed by David A. Wheeler, 1-Dec-2019.) $)
+    df-reflexive $a |- ( Reflexive A R <-> ( A. x e. A x R x ) ) $.
```
tirix: I believe you don't need the parenthesis around ` A. x e. A x R x ` :)

set.mm
```diff
@@ -520225,6 +520231,62 @@ _modus ponens_ (our ~ ax-mp ), but he later qualifies this.
...
+     ~ df-irreflexive .  (Contributed by David A. Wheeler, 1-Dec-2019.) $)
+  wirreflexive $a wff Irreflexive A R $.
+
+  ${
+    $d A R x $.
+    $( Define irrelexive relationship; R is irreflexive over the set A iff
+       ` A. x e. A -. x R x ` .  Note that a relationship can be neither
+       reflexive nor irreflexive.  (Contributed by David A. Wheeler,
+       1-Dec-2019.) $)
+    df-irreflexive $a |- ( Irreflexive A R <-> ( A. x e. A -. x R x ) ) $.
```
tirix: Same, no need for the parenthesis around ` A. x e. A ... `. set.mm likes them so much we tend to put too many!

david-a-wheeler:
Thanks for the review! The infix form is more flexible, because you can then say "over the set of...".

However, that leads to a problem. I can imagine saying R Reflexive _V , that is, R is reflexive over all sets, but _V is of course a class and not itself a set. The definition of X. only works with sets though.

One easy solution is to just omit R C_ ( A X. A ). I think it still works as a definition, yes?

benjub:
> One easy solution is to just omit R C_ ( A X. A ). I think it still works
as a definition, yes?

It still works as *a* definition, just not the definition of a reflexive
binary relation, but rather of a "reflexive thing".  For instance, with
that definition, any class would be a reflexive relation on the empty set,
whereas we all know that there is a unique reflexive relation on the empty
set: the empty relation.  So I really think that my proposal to require R
C_ ( A X. A ) makes things cleaner.  (An argument to not require it is that
"we don't care about what happens outside of the domain we consider", but I
think this is sloppy.)

I think I would actually choose my second proposal, now: making "Reflexive"
the function which associates with a set the set of reflexive relations on
it. This does not work for proper classes but I think this is not a big
loss, and it has more advantages (not introducing a new syntax being among
them), and it looks cleaner to "internalize" definitions like this.  For an
example, see my section "21.29.6.1  Diagonal in a Cartesian square" around
the theorem ~df-bj-diag.

Coming from category theory, I obviously prefer the element-free
definitions proposed by Thierry (which could use my ~df-bj-diag for maximum
clarity and proximity with textbooks), if they are accepted (I actually use
them in my work, e.g. https://arxiv.org/abs/1510.01499).

tirix:
Another interesting idea would be to define:
` RefRel = { r | ( Rel r /\ ( _I |' dom r ) C_ r ) } `
` TransRel = { r | ( Rel r /\ ( r o. r ) C_ r ) } `
` SymRel = { r | ( Rel r /\ '' r C_ r ) } `
This avoid defining a new "notation", i.e. we have only new constant classes, this but works for relations over proper sets. One would be able to write for example ` _I e. RefRel ` or ` ~< e. TransRel ` without any notion of the "base" class.
And of course, definition variants with ` A. x e. dom r x r x ` also work.

tirix:
Forget my previous post, it indeed cannot work like that!
If the predicate has to work on proper classes, we _need_ new notations.
If we can restrict ourselves to sets, we can use functions, relations or abstract builders.

tirix:
David, you have a lot of options to choose from!
I still believe (like Benoît), that it's not a big problem if the solution chosen does not work for proper classes. If you choose for example a relation, we might not be able to prove ` _I reflexive _V ` , but we would still be able to prove ` ( A e. V -> ( _I |' A ) reflexive A ) ` or similar, which I believe is sufficient.

nmegill:
While of course you are encouraged to experiment in your mathbox, I have mixed feelings about making these part of the main set.mm.

Re:  "This defines the terms generally, so that there's a formal definition of these terms somewhere."  Adding a set.mm definition that probably won't be used much if at all, just to clarify a concept, seems slightly abusive of our definition mechanism, since there is no technical need driving it (like shorter proofs and theorem statements).  If we need to make a concept precise for the reader, why can't it be stated in precise English or, if necessary, as a string of backquoted math symbols in a comment?

Also, if these concepts are needed for some purpose, it would be more in keeping with our conventional practice to introduce them in the framework of our extensible structure builders.  So instead of  'A Reflexive B', we would define the class of all reflexive structures, Refl or something, which would be identical to Poset (~ df-poset) except that the quantified condition would be replaced with simply 'A. x e. b  x r x'.  But just adding Refl when there's not a practical need for it seems somewhat wasteful.  Indeed, the only simpler structure than poset defined in Wikipedia is a preorder (~ df-preset), which is reflexive and transitive, and the reflexive part is broken out with ~prsref should we need it.  Maybe there is a need for Refl in some application I'm not familiar with, and if so we can add it when the need arises.

Requiring the field of the relation to be restricted to the base set means some extra complexity of cutting down the relation when working with substructures with a subset of the base set.  We don't require that for our existing structures such as Poset, and it hasn't been a problem.  I don't see the practical need, even if in some sense as Benoit says it would be less "sloppy".  One purpose of having a separate base set is to allow the use of a possibly larger relation for convenience, otherwise we might as well abandon the base set and just use the field of the relation.

david-a-wheeler:
I haven't proposed making these part of the main set.mm; they're clearly marked as experimental. Of course, if they end up being useful then we talk about moving them into the main body. But that's putting the cart before the horse; let's see if there's anything to see first :-).

david-a-wheeler:
Even if a definition isn't used elsewhere, I think it's valuable to have a *formal* definition of key terms (especially for terms that are widely used in the set.mm comments or in external math literature). We can always mark them them "New usage is discouraged" to prevent use if we want.  But that's prejudging anyway.  Once such definitions are created, they might be quite useful directly within set.mm.  Hence, I'm experimenting in a way that others can see & kibbitz.

I sometimes find it useful to just "try and see",  I can only see so many moves ahead on the chess board, and others are likely to see things I don't.  I don't agree at all with Yoda - there *is* a try :-).

benjub:
>  One purpose of having a separate base set is to allow the use of a possibly larger relation for convenience, otherwise we might as well abandon the base set and just use the field of the relation.

@nmegill : It is not always possible to recover the base set from the relation: at the risk of being pedantic, I would say that the empty relation on { (/) } is another object than the empty relation on the empty set. A bit less artificial is the fact that the exponential function from \R to \R is another object than the exponential function from \R to \R_{>0}. For instance, the latter is surjective whereas the former is not.  The current coding of functions and relations in set.mm does not permit to distinguish these objects, and for the same reason, the notion of "corestriction" is hard to define. As you know, another convention is to include the domain and codomain in the function, that it, a function (or relation) from A to B is a triple f = (A, C, B) such that C C_ ( A X. B ) and...
I do not mean to propose a change in the set.mm convention, but one has to be aware of that.

I would encourage @david-a-wheeler to define them in his mathbox (or some of them, time permitting), in order to give a proof-of-concept.  In my opinion, the clarity of the definitions would speak for itself.  My preferred choice would be:

BinRel = ( x e. \_V |-> { r | r C\_ ( x X. x ) } )
ReflBinRel = ( x e. \_V |-> { r e. ( BinRel \`x ) | ( Diag \` x ) C\_ r } )
IrreflBinRel = ( x e. \_V |-> { r e. ( BinRel \`x ) | ( r i^i ( Diag \` x ) ) = (/) } )
TransBinRel = ( x e. \_V |-> { r e. ( BinRel \`x ) | ( r o. r ) C\_ r } )
SymBinRel = ( x e. \_V |-> { r e. ( BinRel \`x ) | \`' r C\_ r } )
AntisymBinRel = ( x e. \_V |-> { r e. ( BinRel \`x ) | ( \`' r i^i r ) C\_ ( Diag \`x ) } )
AsymBinRel = ( x e. \_V |-> { r e. ( BinRel \`x ) | ( \`' r i^i r ) = (/) } )
(you can also define similarly left-Euclidean, right-Euclidean...)

Preorder = ( x e. \_V |-> ( ( ReflBinRel \` x ) i^i ( TransBinRel \` x ) ) )
Equivalence = ( x e. \_V |-> ( ( Preorder \` x ) i^i ( SymBinRel \` x ) ) )
Order = ( x e. \_V |-> ( ( Preorder \` x ) i^i ( AntiSymBinRel \` x ) ) )
StrictOrder = ( x e. \_V |-> ( ( TransBinRel \` x ) i^i ( ASymBinRel \` x ) ) )

One can then prove several theorems, like for instance the fact that an equivalence can also be defined as a reflexive Euclidean binary relation.

You can also define the reflexive, symmetric and transitive closures of a binary relation.  I would first prove that the sets defined above are closed under arbitrary intersections (or only non-empty intersections for irreflexive, antisymmetric and asymmetric ones), and then define the transitive closure of r e. ( BinRel `x ) as the intersection of all the transitive relations on x containing r.  Combining these closures, you would define the equivalence, or the order, generated by a binary relation, etc.

nmegill:
@benjub 

> It is not always possible to recover the base set from the relation

Sorry, I meant the Reflexive relation; I should have been specific.  I'm not proposing doing this. :)

For the definitions you propose, yes, we can prove some theorems relating them to each other, but I don't see a wide applicability outside of the collection itself.

Wouldn't they be more useful as extensible structures?  We have some of them already defined e.g. preorder (df-preset), and it is easy to introduce new ones when they become needed for some application.  I don't think we need to define two versions of the same concept, with two acronyms to remember and possibly some confusion about which one to use.

To me these would be better just stated as a convenient glossary in a comment that defines them all in one place in English, e.g. "A binary relation \` R \`  is \_reflexive\_ over a base set \` A \` iff \` x R x \` holds for all \` x e. A \`.".  A hyperlink could be included whenever a corresponding extensible structure definition exists.  To me, that seems a little less cryptic than "ReflBinRel = ( x e. \_V |-> { r e. ( BinRel \` x ) | ( Diag \`  x ) C\_ r } )".

tirix:
I've also thought about providing the definitions through extensible structures. As noted by Norm, we already have Preset, Poset, Toset, Dirset, Lattices... (see [the schema of the algebraic structures](http://us.metamath.org/mpeuni/mmtopstr.html))

However I think this would be too restrictive. For example, in geometry, there is [a "less-than" relation](http://us.metamath.org/mpeuni/df-leg.html), but while the Geometry itself is an extensible structure, I would have to build a new structure to state "[the geometric less-than relation is transitive](http://us.metamath.org/mpeuni/legtrd.html)". I don't think it's worth it. 

In any case, I think the difference of opinion about whether these definitions shall be included or not lies in what one would like to see in set.mm: if it's to be an encyclopedia of Mathematics, including high school math and more, then definitions for the different types of relations belong in it.

benjub:
First, a side remark (but important nonetheless): I advise to change
everywhere (i.e. in comments, labels, symbols) "preset" to "proset".
Indeed, "proset" can only mean "preordered set", whereas "preset" can mean
several things (including "set without an equality relation"). Plus,
"proset" is apparently more widely used.  If you don't like "proset"
(personally, I'm not fond of it and I think it is dangerously close to
"poset"), then only use the non-abbreviated "preordered set" in comments
(and still "proset" in labels and symbols).

Second, I am aware that many of these questions about encoding mathematical
objects in set.mm are less about mathematics than about engineering, and I
admit that I do not have much experience in the latter (like building long
proofs, for instance).

I do not like extensible structures much, since in my opinion they are too
far away from the mathematical object, and not that flexible (since "slots"
are hard-coded). But again, they might be more convenient for proof
building and this may overcome that drawback, I cannot tell. For instance,
the operation of quotienting by an equivalence relation is ubiquitous
mathematics. So should one create an extra "slot" to host a binary relation
which is meant to be an equivalence relation? Thierry gives an example where
this would be overkill.

As for "defining two versions of the same concept": it is not exactly the
same concept, and both are omnipresent in books. For instance, I think the
definition
  Poset = { x | (le \` x ) e. ( Order \` ( Base \` x ) ) }
(with my above definition of Order) is a transparent encoding of the
textbook definition "A partially ordered set is a set equipped with a
partial order."

In any case, I am only saying that experimenting with this kind of things
in a mathbox is interesting to me, nothing more.

As for the element-free version, i.e. " ( Diag \` x ) C_ r " versus " A. y
e. x y r y " : it is my category theory bias which makes me prefer the
former, and it is a detail since set.mm is firmly grounded on set theory.
Element-free definitions make it easier to "internalize", as I wrote above,
the concept of a reflexive/transitive/etc. relation on an object of a
category. More precisely, as soon as you have a concept of element and
membership, you can define inclusion and composition, but in the opposite
direction, it can happen that you have a category with products, with a
notion of inclusion (subobject) and composition, but without elements (or
where defining "element" and "membership" would be artificial). On objects
of that category, you can still define a reflexive/transitive/etc. relation
using the element-free definition.

david-a-wheeler:
@benjub : your comment about "preset" vs. "proset" is interesting, but it's a different topic.

I've pealed that off into issue https://github.com/metamath/set.mm/issues/1289 ; that way, it won't be lost & it can be discussed separately.

david-a-wheeler:
@tirix : I completely agree that there a lot of options to choose from if it only needs to handle sets. I had originally intended to also handle proper classes, but that may not be a reasonable goal.

benjub:
relexive --> reflexive
irrelexive --> irreflexive
relationship --> relation (in the present case)
Your definitions are reasonable, and I think this can be merged.
A side remark on "making it work also for proper classes": as you know, proper classes are only "virtual objects" in the axiomatization of set.mm, and being a set is equivalent to existing (this is ~isset). Therefore, the need of making it work for proper classes is important only for the most fundamental concepts. I'm not sure (genuinely not sure: this is not rhetorical!) about the present case.
By the way, in my opinion, couples (ordered pairs) are such a fundamental concept, and I describe a definition of couples which works for proper classes, in my mathbox (see bj-2uplth and bj-2uplex).

david-a-wheeler:
@benjub - thanks for the spelling notes, I'll fix it.

Yes, I'm quite aware that classes are only "virtual objects". I intend to merge this into my mathbox and (1) play with and (2) see if others have ideas.  I suspect what will happen long-term is that I switch to something like your definitions; once things are limited to sets it's a lot easier to handle them in set.mm (who would have thought?).

nmegill:
@benjub 
> I do not like extensible structures much, since in my opinion they are too far away from the mathematical object, and not that flexible (since "slots" are hard-coded).

I don't know whether there is a better approach (we'd like to hear your ideas), but it was clear that the now deprecated previous approach (GrpOp, RingOps, etc.) was awkward and too inflexible for practical development of more complex structures.

Extensible structures were discussed around 2011 for a long time, going through several iterations before they were added to set.mm.  These discussions are archived in the AsteroidMeta wiki, and I can look them up if you're interested.  Even after that, I kept the early development (that converted theorems about GrpOp to Grp etc.) in my mathbox for many months before I finally felt comfortable importing it.  More recently Mario and Stefan have fine-tuned the mechanism by adding features like Slot (df-slot) and |\`s (df-ress).

If you have an idea for improving or replacing the extensible structure mechanism, we'd like to hear it.  It would be a large project to replace existing definitions if a better approach is found, but it is probably feasible as long as the theorems remain the same.  That way we'd only have to change a few initial theorems using the new definition.  E.g.  ablcom would hopefully still be
```
ablcom.b $e |- B = ( Base ` G ) $. 
ablcom.p $e |- .+ = ( +g ` G ) $. 
ablcom $p |- ( ( G e. Abel /\ X e. B /\ Y e. B ) -> ( X .+ Y ) = ( Y .+ X ) ) $=
```
even though the details of what "Abel", "Base", and "+g" mean might differ at the definitional level.  BTW we use ( Base \` G ) instead of ( G \` Base ) so that we can make an operation such as group inverse be either primitive (given as a structure member) or defined, without affecting later theorems, and we'd want to keep that feature.


> In any case, I am only saying that experimenting with this kind of things
in a mathbox is interesting to me, nothing more.

I have no issues with putting (almost) anything you want in your mathbox.  My main concern is when a long development is done from a definition that we decide not to import, there may be resentment that the work was "wasted" (even though it can still stay in the mathbox indefinitely).  So it's good to get a consensus on a new mathbox definition before going too far with it.

nmegill:
@david-a-wheeler I didn't notice you changed the title to DO NOT MERGE.  I will try to revert it.

david-a-wheeler:
Do not bother reverting it. They are just typos, I can fix it with later commits.

david-a-wheeler:
Oh I see you did revert it. No problem. I am dealing with some family issues, so I may be unavailable for a little while, but I will get back this and complete. Thank you so much.

benjub:
@nmegill I do not know either if there is a better approach, and as I wrote, I feel that I do not have enough experience with building longer proofs in order to judge which is better in practice. The fact that I am not satisfied with extensible structures is not criticism.

I would maybe define things like the category of groups (or possibly $u$-small groups where $u$ is a universe), rings, etc. and then forgetful functors between them (e.g. the underlying group of a ring, the underlying set of a topological space, etc.).  But maybe all these forgetful functors would hinder the readability of the statements?

For instance, is there in set.mm the theorem "the multiplicative group of a finite field is cyclic"? How to state it with extensible structures ? Do we have to "put the multiplication in the slot corresponding to the group law" ? Not a very mathematical operation.
If we define the category Field and the functor MulGroup : Field ---> Group (or Field ---> Ab ?), we could state something like
|- ( ( F e. ( Obj \` Field ) /\ ( Base \` F ) e. Fin ) -> ( MulGroup \` F ) e. CyclicGroup )

Of course, there is also a functor MulGroup : Ring ---> Group.  Maybe identify them, and make Field a genuine subcategory of Ring (and similarly for every full subcategory)? I do not know what is best.

These are only ideas, and I am not advocating any change soon. Spending more time on this right now would be premature.

I'll try to look at the discussion you hinted at (do you have link)?

tirix:
Hi Benoît,

To answer your questions about the extensible structures:
I don't think we have the theorem "the multiplicative group of a finite field is cyclic", but it would be stated like this:
` fifldmgpcyg $p |- ( ( F e. Field /\ ( Base ' F ) e. Fin ) -> ( mulGrp ' F ) e. CycGrp ) $. `

Concerning "Field as a subcategory of Ring", we do have in set.mm:
` isfld $p |- ( R e. Field <-> ( R e. DivRing /\ R e. CRing ) ) $. `
` drngrng $p |- ( R e. DivRing -> R e. Ring ) $. `
From which it follows:
` Field C_ Ring `
(even if there is no theorem proving that)

nmegill:
@benjub 
> I'll try to look at the discussion you hinted at (do you have link)?

It seems to be lost forever.  (: The last backup that I made of AsteroidMeta (at http://us2.metamath.org/other/AsteroidMeta/metamath) was prior to that discussion, and archive.org didn't save it:  at https://web.archive.org/web/20130625024230/http://wiki.planetmath.org:80/cgi-bin/wiki.pl/metamath the "Extensible structure builder" link was not archived.

PR #1292, opened 2019-12-04
shorten ralimdaa, ralbida, ralbii

wlammen:
... and cluster more theorems around ralim*
... and move aev + family before the introduction of ax-13 
@nmegill I need your advice here: I introduced a hb version of ralrimi for the sole purpose of reducing axiom usage in ralrimiv without fully duplicating the proof of ralrimi.  I suspect there are much more occasions where a usage of nfv before calling a F/ based theorem drags in extra axioms.  To avoid both a proof duplication and unnecessary axiom references, you need to create a hb form in each such instance, from which you fork both a normal and a -v version.

Is this really what we want? Or isn't it better to say at some level of abstraction, all axioms of FOL are assumed to hold, and one does not care about fine tuned axiom usage any more? What do you think?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17676,7 +17676,6 @@ New usage of "r1pwOLD" is discouraged (0 uses).
 New usage of "ralbii2OLD" is discouraged (0 uses).
 New usage of "ralimOLD" is discouraged (1 uses).
 New usage of "ralimdaaOLD" is discouraged (0 uses).
-New usage of "ralrimiALT" is discouraged (0 uses).
 New usage of "ralxfrALT" is discouraged (0 uses).
 New usage of "rb-ax1" is discouraged (5 uses).
 New usage of "rb-ax2" is discouraged (9 uses).
--- a/discouraged
+++ b/discouraged
@@ -17673,6 +17673,7 @@ New usage of "qlaxr5i" is discouraged (0 uses).
 New usage of "quoremnn0ALT" is discouraged (0 uses).
 New usage of "r1omALT" is discouraged (0 uses).
 New usage of "r1pwOLD" is discouraged (0 uses).
+New usage of "ralbidaOLD" is discouraged (0 uses).
 New usage of "ralbii2OLD" is discouraged (0 uses).
 New usage of "ralimOLD" is discouraged (1 uses).
 New usage of "ralimdaaOLD" is discouraged (0 uses).
@@ -19421,6 +19422,7 @@ Proof modification of "qexALT" is discouraged (64 steps).
 Proof modification of "quoremnn0ALT" is discouraged (360 steps).
 Proof modification of "r1omALT" is discouraged (13 steps).
 Proof modification of "r1pwOLD" is discouraged (151 steps).
+Proof modification of "ralbidaOLD" is discouraged (43 steps).
 Proof modification of "ralbii2OLD" is discouraged (37 steps).
 Proof modification of "ralimOLD" is discouraged (59 steps).
 Proof modification of "ralimdaaOLD" is discouraged (49 steps).
--- a/discouraged
+++ b/discouraged
@@ -17677,6 +17677,8 @@ New usage of "ralbidaOLD" is discouraged (0 uses).
 New usage of "ralbiiOLD" is discouraged (0 uses).
 New usage of "ralimOLD" is discouraged (1 uses).
 New usage of "ralimdaaOLD" is discouraged (0 uses).
+New usage of "ralrimiOLD" is discouraged (0 uses).
+New usage of "ralrimivOLD" is discouraged (0 uses).
 New usage of "ralxfrALT" is discouraged (0 uses).
 New usage of "rb-ax1" is discouraged (5 uses).
 New usage of "rb-ax2" is discouraged (9 uses).
@@ -19425,6 +19427,8 @@ Proof modification of "r1pwOLD" is discouraged (151 steps).
 Proof modification of "ralbidaOLD" is discouraged (43 steps).
 Proof modification of "ralimOLD" is discouraged (59 steps).
 Proof modification of "ralimdaaOLD" is discouraged (49 steps).
+Proof modification of "ralrimiOLD" is discouraged (24 steps).
+Proof modification of "ralrimivOLD" is discouraged (9 steps).
 Proof modification of "ralxfrALT" is discouraged (85 steps).
 Proof modification of "rb-ax1" is discouraged (32 steps).
 Proof modification of "rb-ax2" is discouraged (17 steps).
--- a/discouraged
+++ b/discouraged
@@ -17674,7 +17674,7 @@ New usage of "quoremnn0ALT" is discouraged (0 uses).
 New usage of "r1omALT" is discouraged (0 uses).
 New usage of "r1pwOLD" is discouraged (0 uses).
 New usage of "ralbidaOLD" is discouraged (0 uses).
-New usage of "ralbii2OLD" is discouraged (0 uses).
+New usage of "ralbiiALT" is discouraged (0 uses).
 New usage of "ralimOLD" is discouraged (1 uses).
 New usage of "ralimdaaOLD" is discouraged (0 uses).
 New usage of "ralxfrALT" is discouraged (0 uses).
@@ -19423,7 +19423,6 @@ Proof modification of "quoremnn0ALT" is discouraged (360 steps).
 Proof modification of "r1omALT" is discouraged (13 steps).
 Proof modification of "r1pwOLD" is discouraged (151 steps).
 Proof modification of "ralbidaOLD" is discouraged (43 steps).
-Proof modification of "ralbii2OLD" is discouraged (37 steps).
 Proof modification of "ralimOLD" is discouraged (59 steps).
 Proof modification of "ralimdaaOLD" is discouraged (49 steps).
 Proof modification of "ralxfrALT" is discouraged (85 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
> @nmegill I need your advice here: I introduced a hb version of ralrimi for the sole purpose of reducing axiom usage in ralrimiv without fully duplicating the proof of ralrimi. I suspect there are much more occasions where a usage of nfv before calling a F/ based theorem drags in extra axioms. To avoid both a proof duplication and unnecessary axiom references, you need to create a hb form in each such instance, from which you fork both a normal and a -v version.

I think it is fine to add an occasional hb* in this area, especially if it eliminates the use of an axiom.  I've felt that hb* is sometimes a better fit than nf* for early work close to the axioms, although it's clear that nf* is better later on.  As you can see, we already have a number of hb* in the earlier sections that we kept even after adding nf*.

> Is this really what we want? Or isn't it better to say at some level of abstraction, all axioms of FOL are assumed to hold, and one does not care about fine tuned axiom usage any more? What do you think?

When we get to set theory, I think we can stop caring about which pred calc axioms are used.  Even as early as E! I don't think it is worth worrying too much about what axioms are used.

In the early and main pred calc develoment, it can be advantageous to have theorems depend on fewer axioms when possible.  The main reason it that we have one more theorem that could be used if we are attempting to show that the eliminated axiom is redundant.  For example, we still do not know whether ax-13 is redundant, and it would certainly be nice to get rid of it.  It is the only axiom that Mario's mm0 doesn't use in some form (because mm0 unbundles everythinig), making it particularly weird for those not used to bundling.

Also, if a theorem can be proved two ways with different axioms, say one way uses ax-12 but avoids ax-13 and the other does the opposite, it would be worth adding both, one as an *ALT.  That might give some insight into the role each axiom is playing and what property ax-12 and ax-13 have in common that is allowing each proof to go through - with a longer term goal of perhaps simplifying the axiom system by modifying one of them to eliminate another.

The FOL axioms beyond ax-9 are non-standard in the literature as well as logically redundant, and the fewer and simpler the better, especially in terms of appeal to outsiders who may find them strange.  Over the years we have shown some to be redundant and simplified others.  In particular, I was very pleased to find the ax-13 replacement for the ugly ax-c9.  But if ax-13 could be eliminated, that would be even better of course.

benjub:
@nmegill I agree with everything, including having both an hb* and an nf* versions of the same theorems in the early treatment. Similarly, having both closed form and deduction form of the same theorems, etc. The early part of set.mm should provide a full and versatile toolbox adapted to different kinds of proving styles, and it's better to have statements which from time to time may prove redundant than lacking the precise statement that would fit best a given situation.

I referred to the "early" treatment above, and you also say that care concerning dependencies on FOL axioms should be taken in the early phase.  I think my definition of the "early" phase is a bit more inclusive than yours: I would include most of the things before Replacement.  Most of the set theory before Replacement, that is, using only Extensionality, is really "staple statements" and I think that minimizing FOL axiom dependencies there is still relevant (FOL is not yet "overwhelmed" by the complexity of set theory). An example is my bj-ru (and bj-ru0/1) which you might enjoy.

nmegill:
@benjub I have no objection to to reducing the FOL axiom usage in set theory before Replacement, especially if it comes for "free" or almost so.  But I question its value.

A primary purpose of doing it for FOL theorems is that it makes those theorems more useful as tools to help with studies of the FOL axioms.  For example, if ax-13 is eliminated from a theorem, then that theorem becomes available for use in an attempt to prove ax-13 redundant.

In set theory before Replacement, I don't see what purpose it would have.  It doesn't provide us with any new tools for study of FOL axioms.  I don't think anyone outside of the set.mm community cares which of our FOL axioms are used in set theory.

What do you see as the purpose for doing this?

I agree that bj-ru0 is nice, but it is pure FOL and could go in the FOL section instead of set theory.  bj-ru1 and bj-ru can reference it in the set theory part, but the "work" to minimize FOL axiom usage will then be done in the FOL section, and the set theory section would simply reference it.

Challenge:  can bj-ru0 be proved from ax-1 through ax-9 only?  Hint: use ax10w through ax13w instead of ax-10 through ax-13.  See ax12wdemo for an example of the techniques to use.  Indeed, I think any FOL statement with only setvar variables, all of which are mutually distinct, should be provable from ax-1 through ax-9 only.

benjub:
@nmegill You're right, removing dependencies on FOL axioms after ax-ext is
introduced is less important.  I still removed dependencies on ax-13 for
some of these theorems, since it will imply that most later theorems will
not require ax-13 anymore, and that axiom is the most elusive of FOL
axioms. Note that I also removed many dependencies on ax-ext, some of which
may come as a surprise (see bj-axsep2 and its comment, and actually also
the whole section "21.29.5.2  Classes without extensionality"; I plan to
post to the discussion group about that).

As for bj-ru0, the current description reads "The FOL part of Russell's
paradox", which I think is fit, and I would indeed put it in the FOL part
if you agree.

Concerning your challenge: I will have a look, but to me, the present proof
of bj-ru0 is already dependent only on classical FOL without the
set.mm-specific axioms ax-10--13.  Indeed, the proof requires ax-1--9
together with ax-10 and ax-12, and:
* the dependency on ax-12 is only because of the use of (an equivalent form
of) sp, which should be an axiom anyway;
* the dependency on ax-10 is because of the current definition of nf, and
it will disappear with my proposal to define F/ x ph as ( E. x ph -> A. x
ph ) which I made on the discussion group a few months ago.  Removing
dependencies on ax-10 was actually the main motivation for this proposal,
which removes from the definition of nf nested quantifications on the same
variable (which is precisely what ax-10 takes care of).  Hopefully I'll
make the definition change during the Christmas break.

nmegill:
@benjub 
> * the dependency on ax-12 is only because of the use of (an equivalent form of) sp, which should be an axiom anyway; 

Theorem spw provides the ax-1 to ax-9 version.  You have to eliminate the hypothesis, but that can always be done when there are no wff variables and all setvars are mutually distinct.

> * the dependency on ax-10 is because of the current definition of nf, and it will disappear with my proposal to define F/ x ph as ( E. x ph -> A. x ph )

If you use ax10w, there won't be any dependence on how F/ is defined.

benjub:
Thanks: you essentially provided the proof of bj-ru0 from ax-1--9, using
ax10w and spw.

What I meant is the following: tracing back the use of ax-10 and ax-12 in
the proof of bj-ru0, we obtain:
bj-ru0 -> bj-spvv -> bj-spimvv -> bj-spimv -> 19.36i -> 19.36 -> 19.9 ->
nfri -> nfr -> sp
and the bifurcation: 19.9 -> 19.9h -> 19.9t -> 19.8a
I claim that ALL these theorems depend only on ax-1--10 plus sp (which
should be an axiom), and that once the definition df-nf is changed
according to my proposal, they will automatically depend only on ax-1--9
plus sp (this definition change has an influence since some of the
statements above contain F/). So after these two modifications (adding sp
as an axiom and changing df-nf), the reduced dependencies come for free.

As to adding sp as an axiom: one should accordingly weaken ax-6 to A. x E.
y y = x in order to maintain independence of the axiom system (and changing
the open formula ax-6 to the closed formula above makes it easier to
understand).  This would have the added benefit of finer modularity:
* one could consider the system {ax-1--5,10,11 + sp + ax-mp + ax-gen} of
"pure" predicate calculus;
* the axiom sp would be the only one not valid in inclusive logic (i.e.,
valid on an empty domain).
However, sp would still not be the axiom pinning down non-emptiness of the
domain of discourse. For this, one would need (using the analogy with modal
logic) to replace the axiomatization T5 with DB4 (I began providing ground
work in my mathbox section on modal logic). Both axiomatizations have their
benefits and drawbacks and I'm not set.

nmegill:
@benjub Our FOL axiom system is primarily intended for ZFC, which of course has a non-empty domain.  In addition, it proves exactly the theorems of the "standard" FOL axioms in almost all textbooks, which assume a non-empty domain and is what most people expect who've had some introductory logic.

Systems valid in the empty domain are a somewhat specialized topic.  Just as we don't break out various sub-logics (minimal, intuitionistic, ...), I don't see that breaking out axioms valid in the empty domain is in the spirit of simplicity we want to achieve with set.mm's FOL axiomatization, because it doesn't serve any purpose once we commit to the ZFC axioms.  That isn't to say it shouldn't be studied, but set.mm (outside of a mathbox) is the wrong place I think.

Moreover, there isn't universal agreement about the behavior of quantifiers in the empty domain:  in Hailperin's version A. x ph is always true and E. x ph always false regardless of ph, whereas in Mostowski's version both A. x ph and E. x ph are equivalent to ph when x is not free in ph.

Hugues LeBlanc shows how to modify Tarski's system (our axioms ax-4 through ax-9) for empty domains for both Hailperin's and Mostowski's versions.  See "On Meyer and Lambert's Quantificational Calculus FQ", J. Symb.  Logic 33 1968 275-280.  For Hailperin's version, it is basically changing ax-6 to "A. x E. y y = x" like you suggested and also adding the axiom "x = x".

benjub:
@nmegill I understand your concerns.  I realize that I have different interests from the main objectives of set.mm.  I have been considering on and off to start another mm-database with more focus on axiomatizations, but this would be a huge work, very largely redundant with that of set.mm, and it's currently not in my plans.

In the present case, having sp as an axiom would allow a clearer exposition of FOL, in my opinion, since currently many theorems not involving equality still require ax-6, 7, 12 and have to be postponed until these are available.  The fact that sp follows from ax-6, 7, 12 is an interesting fact and a tour-de-force of Kalish-Montague, but to me, it is rather a curiosity.

Concerning the semantics on the empty domain: they show precisely that open formulas are harder to grasp than closed formulas, and therefore that "A. x E. y y = x" is preferable to "E. y y = x".

PR #1295, opened 2019-12-05
shorten speimfw, ralbii2

wlammen:
This shortening was made possible through the introduction of aleximi in August.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17993,6 +17993,7 @@ New usage of "spanunsni" is discouraged (1 uses).
 New usage of "spanval" is discouraged (6 uses).
 New usage of "speccl" is discouraged (0 uses).
 New usage of "specval" is discouraged (1 uses).
+New usage of "speimfwOLD" is discouraged (0 uses).
 New usage of "sps-o" is discouraged (7 uses).
 New usage of "sqgt0sr" is discouraged (1 uses).
 New usage of "ssdmd1" is discouraged (1 uses).
@@ -19554,6 +19555,7 @@ Proof modification of "snsslVD" is discouraged (24 steps).
 Proof modification of "soirriOLD" is discouraged (36 steps).
 Proof modification of "son2lpiOLD" is discouraged (28 steps).
 Proof modification of "sotriOLD" is discouraged (83 steps).
+Proof modification of "speimfwOLD" is discouraged (35 steps).
 Proof modification of "sps-o" is discouraged (10 steps).
 Proof modification of "sseliALT" is discouraged (152 steps).
 Proof modification of "sspwimp" is discouraged (87 steps).
--- a/discouraged
+++ b/discouraged
@@ -17675,6 +17675,7 @@ New usage of "r1omALT" is discouraged (0 uses).
 New usage of "r1pwOLD" is discouraged (0 uses).
 New usage of "ralbidaOLD" is discouraged (0 uses).
 New usage of "ralbidvaOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@wlammen and @nmegill : I don't understand the comments of speimfw and spimfw: "additionnal weakening" compared to what ? Is it possible to expand a bit the comments ?  Also, wouldn't it be clearer to state these using E. instead of -. A. -. ? Thanks

wlammen:
This theorem is related to 19.2.  Unlike 19.2, it does not require ax-6 be valid (that is its stronghold), but still = must be defined, and it needs two prerequisites, the hypothesis and the first antecedent.  That is, why it is called a weak version of 19.2.

All of this is refers to some work called Tarski's FOL, based on fewer axioms, but I guess, it is likely in need of some meta rules, in order to arrive at a similar richness of theorems as the Metamath axioms allow.  I am not familiar with this Tarski's theory, please refer to literature or ask Norm for details.

wlammen:
It seems the travis checks are not executed in parallel anymore.  Is this a new feature or a bug?

benjub:
@wlammen Thanks. In the comment of speimfw, can you add:
...with additional weakening (compared to ~ 19.2 ) to allow...
and for spimfw:
...with additional weakening (compared to ~ sp ) to allow...
(if it is indded ~ sp ).

What do you thing about using E. instead of -. A. -. in the statements ? (this would be more readable and shorten the proof of speimfw

nmegill:
> @wlammen Thanks. In the comment of speimfw, can you add:
> ...with additional weakening (compared to ~ 19.2 ) to allow...
> and for spimfw:
> ...with additional weakening (compared to ~ sp ) to allow...
> (if it is indded ~ sp ).

That sounds OK.

> What do you thing about using E. instead of -. A. -. in the statements ? (this would be more readable and shorten the proof of speimfw

The \*w stuff in the section prior to ax-10 is meant to show what is possible to prove using only Tarski's schemes ax-4 through ax-9.  The 1965 papers of Tarski and Kalish/Montague use -. A. -. exclusively instead of E. (they don't even define E.), and I tried to follow that work as closely as I could.  These are proofs directly from their axiom schemes, and while I can't judge what they were thinking, my guess is that it makes some of their proofs shorter by not having to convert back and forth between -. A -., and E.  In addition, the -. A. -. form has the initial -. conveniently available to be eliminated with contraposition.  Note that speimfw (which is really just a lemma for later theorems in that section) has ax-6 as its antecedent, and ax6v is later used to eliminate it in spimw, again staying closer to the starting axioms.

(The 3rd paper by Monk in that 1965 series finally does introduce E..)

The main goal of the \*w stuff is to prove schemes ax10w, ax11w, ax12w, and ax13w thus showing that our ax-10 through ax-13 do not strengthen Tarksi's system at the object language level.  Many of the proofs are based on the Kalish/Montague proofs, which use induction on formula length, some of which were (for me) somewhat opaque and difficult to follow.  So I was pleased that I was able to express their (meta)theorems using implicit substitution hypotheses, which I think makes them much easier to follow.

The bottom line is that I think our use of -. A. -. in that section keeps it closer to the literature and thus possibly easier to correlate with the literature proofs.  These are all technical lemmas and theorems that are unlikely to be used outside of that section.  In particular, no one would use sp\*w when the much simpler sp\* are available, unless it is for a technical reason like to avoid one of the axioms >= ax-10.  Same for ax10w-ax13w since ax-10-ax-13 are eventually available.

If people feel strongly about it I could accept converting -. A. -. to E. but personally I would prefer to keep it closer to the literature.

PR #1297, opened 2019-12-05
reduce dependencies on axioms on a larger scale

wlammen:
This affects dozens of theorems in equality section.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -308,6 +308,7 @@
 "4syl" is used by "efgsdmi".
 "4syl" is used by "el2spthonot".
 "4syl" is used by "elaa".
+"4syl" is used by "eqeq1d".
 "4syl" is used by "erdsze2lem1".
 "4syl" is used by "f1ocnvfvrneq".
 "4syl" is used by "f1otrg".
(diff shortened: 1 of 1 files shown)


nmegill:
Note that metamath.exe accepts acknowledgment tags anywhere in the comment, so 

"Revised to remove dependencies on axioms.  (Contributed by NM, 27-Dec-1993.)  (Revised by  Wolf Lammen, 5-Dec-2019.)"

could also be

"(Contributed by NM, 27-Dec-1993.)  Revised to remove dependencies on axioms.  (Revised by Wolf Lammen, 5-Dec-2019.)"

Although this is a stylistic choice and not a requirement (there is no need for you to change existing comments), it more closely associates what is revised with the "(Revised by...)" tag.  In particular, if there are 2 revisions, it can make clearer who did what.

benjub:
@nmegill I agree with this convention, which I was applying without knowing it.  A similar convention (@david-a-wheeler : conventions!) I try to consistently apply is to write:
* just before a tag "New usage is discouraged." something like "Use ~ xxx instead." (and/or an explanation why, as in 4syl);
* just before a tag "Proof modification is discouraged." an explanation why (if it's not obvious), like in 
bj-nfcsym to give an example.
Finally, concerning the ordering of the tags when both are present, I try to consistently put "Proof..." first because usage concerns later proofs.

PR #1304, opened 2019-12-08
Shorten 19.8a

wlammen:
Jim Kingdon s version of ax12v does not depend on ax-13, so prefer it over the current one.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13655,6 +13655,7 @@ New usage of "19.35OLD" is discouraged (0 uses).
 New usage of "19.41rg" is discouraged (3 uses).
 New usage of "19.41rgVD" is discouraged (0 uses).
 New usage of "19.43OLD" is discouraged (0 uses).
+New usage of "19.8aOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
@jkingdon I upgraded your version of ax12v to be the main version, because it depends on less axioms.  Check whether its comment is ok for you.

jkingdon:
> @jkingdon I upgraded your version of ax12v to be the main one, because it depends on less axioms. Check whether its comment is ok for you.

Looks good.

My work on ax12v made it into set.mm - see #192 if the background is relevant. There might be others which are only in iset.mm, which in general develops predicate logic with more use of distinct variables and less use of distinctors or http://us.metamath.org/mpeuni/ax-12.html (not due to a concerted effort to minimize axiom usage, more because of the inability to do case elimination on distinctors).

wlammen:
@jkingdon Thanks for your hint.  I certainly want distinct variable versions, because they help reducing dependencies on axioms.
By the way, ax12v was shortened by me, the pull request is pending.

PR #1307, opened 2019-12-09
revert ralbii; minimal axiom usage in axext3, axext4

wlammen:
This is an interesting move.  I had to revert ralbii, because eqid is artificially based on ax-ext.  But ralbii is provable without this axiom.  For the particular class equation A = A, the reference to ax-ext is questionable.
@benjub I later saw that your mathbox has contents related to axext3. I proved this version independently, was actually inspired by the recent proof of ax12v.  As far as I am concerned, feel free to add your name to the contributors of axext3.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14024,6 +14024,7 @@ New usage of "axcnex" is discouraged (0 uses).
 New usage of "axcnre" is discouraged (0 uses).
 New usage of "axdc" is discouraged (0 uses).
 New usage of "axdistr" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@wlammen That's fine, thanks. I'll have to look at the proof of bj-axext3 (and bj-axext4) again, since it may have independent interest, not sure. Could you be more precise in the comment, e.g.
> Revised to remove dependency on ~ ax-13 [ , ~ ax-xx ... ]

instead of
> Revised to minimize axiom dependencies

Thanks.

Issue #76, opened 2016-08-31
Rewrite "no ph is ps" in Aristotelian section


david-a-wheeler:
Currently "no ps is ps" is written as -. E. ( ph /\ ps ) in the Aristotelian section. I suggest rewriting as A. ( ph -> -. ps ) ; it would make the proofs shorter, and that suggests that this would be the more natural representation in metamath anyway.

david-a-wheeler:
I plan to do this soon. I created this issue just to see if there were objections.

david-a-wheeler:
Note that this is done in Adriane Rini's "Aristotle's Modal Proofs", Springer, 2011.  So this modeling of No.... using all.. is already in the literature.

PR #1317, opened 2019-12-13
Fourier series convergence, #76 in the 100 theorems list.

glacode:
The main result in this commit is ~ fourier, the proof of the convergence of Fourier series for piecewise smooth functions. This is #76 in the 100 theorems list.
As an immediate application, the well known Fourier expansion for the (discontinuous) square wave function is proven in ~ fouriersw .
The deduction form of the main theorem is ~ fourierd: the series converges to the average value of the left and the right limit of the function. Thus, if the function is continuous at a given point, the series converges exactly to the function value, see ~ fouriercnp . Notice that for a piecewise smooth function, the left and right limits always exist, see ~ fourier2 for an alternative form of the theorem that makes this fact explicit. When the first derivative is defined on the entire periodic interval, a simpler version of the theorem can be stated, see ~ fouriercn .


david-a-wheeler:
CONGRATULATIONS!  Nice work.

I think you need to regenerate the "discouraged" file to make all the checks pass.

I'm dealing with some personal matters, so it may be a little while before I can do my normal update of the Metamath 100. But I'll be delighted to do that.

Again, congrats!!

glacode:
Hi Dave,

I guess I have to study the discouraged "stuff"  :-)

I'm pretty sure mmj2 autofilled in some discouraged label and 
unfortunately it didn't occur to me the (obvious) need for this 
additional check.

A bit late here in Italy, I need a good sleep and tomorrow I will start 
to investigate how to solve the issue...

Thanks for the wonderful job that you do for this amazing framework.

Glauco

benjub:
Simply copypaste the relevant line(s) at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md

Congratulations, and I hope you post about it on the google discussion group and give a brief high level summary of the proofs.

david-a-wheeler:
@glacode - no worries, you've certainly earned a rest.

The big issue is please check to see if you're using something you shouldn't. If the uses are intentional, regenerating the discouraged entries is easy.

nmegill:
@glacode Congratulations on this work!

The "discouraged" file mismatches because you reference the complex number construction of the  instead of their axioms.  For example, you should use ax-1cn instead of ax1cn.   This keeps our complex number section independent of how they are constructed.   See the differences here:
https://travis-ci.org/metamath/set.mm/jobs/624853957?utm_medium=notification&utm_source=github_status

Fixing these should be just simple label substitutions in your proofs, but it would be good to fix them before we merge to prevent future confusion.  Please **don't** regenerate the 'discouraged' file (that would defeat its purpose, which is to detect exactly this kind of problem).  It looks like you don't need to make any changes at all to 'discouraged'.

After you make the label changes, do 'save proof * /compressed /fast' in metamath.exe to prevent > 79 char lines (since the axioms add 1 more character to the label).

I'll look forward to seeing these fixed.  Thanks, Norm

nmegill:
Unfortunately there is another complication.

In order to keep labels more uniform from the reader's point of view, some of the complex number axioms have aliases that must be used, such as mulcom instead of ax-mulcom (as described in the ax-mulcom description).  Here is a list of the labels and what you should change them to:

```
ax1cn       use ax-1cn
ax1ne0      use ax-1ne0
axaddf      use ax-addf
axcnex      use cnex
axicn       use ax-icn
axmulcom    use mulcom
axresscn    use ax-resscn
```

Furthermore, ax-addf has "(New usage is discouraged.)" with no alternate alias.  The reason is that historically ax-addf was added later (a decade after the others), and at that time I felt that it would be cleaner to use only the properties given by the earlier axioms when possible and not depend on ax-addf.  Whether or not that is a valid concern, you can continue to use ax-addf as you are now.   However, you **will** have to regenerate the 'discouraged' file to accommodate the ax-addf usage.

Thanks,
Norm

david-a-wheeler:
How hard would it be to switch from using ax-addf to using addcl?

nmegill:
> How hard would it be to switch from using ax-addf to using addcl?

It probably isn't worth worrying about at this point in order not to delay the PR.  We can look at it separately if we want.  It's already been used 18 times before this PR.

glacode:
I knew pretty well about the discouraged concept. Unfortunately, mmj2 
sometimes autocompletes using discouraged labels (I believe, by default, 
it filters them out in searches, but not in autocomplete).

Instead I didn't know much about the discouraged file. Now I'm cleaning 
up the "code" (a cycle of relabeling, creating a new local discouraged 
file, filediffing against the online discouraged file last version).

Main concern, I'm not that good with set.mm reformatting with 
metamath.exe (I spend all the time in mmj2), but I'm confident I can do 
it :-)

When I'll retry a pull request, I'll look at mmj2 Travis if any 
discouraged is left, and go on cleaning and resubmitting if needed (if 
in doubt and in need of help, I'll reply to this message).

Thanks for all the precious feedback.

Glauco

david-a-wheeler:
As a separate issue we should probably  remove uses of ax-addf when they are not necessary. But I agree that should not hold this result up, we can do that separately.

We keep getting unintentional uses of discouraged theorems. I thought mmj2 did not fill in with discourage their on this. That sounds like a bug in its autocomplete. Not a serious bug obviously, but still not what we want.

tirix:
Hi Glauco,

Congratulations for the proof!

I found just now the parameter [ProofAsstExcludeDiscouraged](https://htmlpreview.github.io/?https://github.com/digama0/mmj2/blob/master/doc/BatchCommandDocumentation.html#command50) to exclude discouraged theorems from unification in MMJ2, I assume that's what you need!

I actually did not know about it, I've been using [ProofAsstUnifySearchExclude](https://htmlpreview.github.io/?https://github.com/digama0/mmj2/blob/master/doc/BatchCommandDocumentation.html#command80) the past, listing there theorems like ~ ax-1cn.

david-a-wheeler:
On Sat, 14 Dec 2019 08:32:15 -0800, tirix <notifications@github.com> wrote:
> I found just now the parameter [ProofAsstExcludeDiscouraged](https://htmlpreview.github.io/?https://github.com/digama0/mmj2/blob/master/doc/BatchCommandDocumentation.html#command50) to exclude discouraged theorems from unification in MMJ2, I assume that's what you need!
> 
> I actually did not know about it, I've been using [ProofAsstUnifySearchExclude](https://htmlpreview.github.io/?https://github.com/digama0/mmj2/blob/master/doc/BatchCommandDocumentation.html#command80) the past, listing there theorems like ~ ax-1cn.

That should probably be the default, but until then, we definitely need to document that better.

--- David A. Wheeler

david-a-wheeler:
I think this PR is obsoleted by the newer pull request #1322

glacode:
> I think this PR is obsoleted by the newer pull request #1322

Yes Dave, #1322 fixes #1317  issues.

david-a-wheeler:
Ok, I'm closing this pull request since it's obsolete.

(BTW, I prefer "David", but I'll answer to almost anything...!).

david-a-wheeler:
FYI, I looked at the documentation on the parameter [ProofAsstExcludeDiscouraged](https://htmlpreview.github.io/?https://github.com/digama0/mmj2/blob/master/doc/BatchCommandDocumentation.html#command50) to exclude discouraged theorems from unification in MMJ2.

It now defaults to true (as it should!).

digama0:
Huh, I didn't know about that option either. (I was going to suggest ProofAsstUnifySearchExclude.) Did I write it?

PR #1320, opened 2019-12-14
Revision of subsection "Constant polynomial matrix"

avekens:
Main Part:
* new theorem ~ 2fvidf1o

AV`s Mathbox:
* Revision of subsection "Constant polynomial matrix": Preparation for the definition of the inverse transformation cPolyMatToMat


benjub:
@avekens , @nmegill : Wouldn't it be more flexible to state ~2fvidf1o as two theorems? I think it should be most often the case when the (consequent of the conclusion of the) theorem is a conjunction ?

PR #1322, opened 2019-12-14
Fourier series convergence, #76 in the 100 theorems list.

glacode:
The main result in this commit is ~ fourier, the proof of the convergence of Fourier series for piecewise smooth functions. This is #76 in the 100 theorems list.
As an immediate application, the well known Fourier expansion for the (discontinuous) square wave function is proven in ~ fouriersw .
The deduction form of the main theorem is ~ fourierd: the series converges to the average value of the left and the right limit of the function. Thus, if the function is continuous at a given point, the series converges exactly to the function value, see ~ fouriercnp . Notice that for a piecewise smooth function, the left and right limits always exist, see ~ fourier2 for an alternative form of the theorem that makes this fact explicit. When the first derivative is defined on the entire periodic interval, a simpler version of the theorem can be stated, see ~ fouriercn .


david-a-wheeler:
I presume that this pull request replaces PR #1317, so we can close that one, correct?

PR #1326, opened 2019-12-16
Congruence of right triangles

tirix:


nmegill:
Looks fine.  I had to look up "cathetus", never heard that word before.  I'm tutoring my nephew in geometry, and I'll have to tell him that word to impress his teacher. :)  His book calls it a "leg", and a geometry book I have from 1927 calls it an "arm".

Issue #1314, opened 2019-12-12
Moving theorems for basic linear algebra from mathoxes into main part


avekens:
I almost finished the revisions of the sections with theorems in the context of the Cayley Hamilton theorem in my mathbox. The next step would be to move these sections (and after the definition of diagonal/scalar matrices is clarified, see the discussion in the Google group https://groups.google.com/d/msg/metamath/4_ZOT4Gddwo/k-lxGMxGBgAJ , also subsection "The subalgebras of diagonal and scalar matrices " ) into the main part, Part "11. BASIC LINEAR ALGEBRA").

I would propose to place subsection "The subalgebras of diagonal and scalar matrices" after subsection "11.2.1 The matrix algebra", and sections "Polynomial matrices" and "The characteristic polynomial " after "11.3 The determinant".

Would this be OK? Of course there will be a lot of auxiliary theorems currently in my mathbox, which are used in the proofs of these sections, to be moved to the main part, too.

avekens:
According to David's Google group post https://groups.google.com/d/msg/metamath/jINlwEjvwOc/IINkUI1sBAAJ, I want to move the Cayley Hamilton theorem and all the theorems required to prove it from my mathbox into the main part of set.mm, as indicated in this issue. Before I can do this, the question how to define diagonal and scalar matrices must be answered, see my Google group post https://groups.google.com/d/msg/metamath/4_ZOT4Gddwo/k-lxGMxGBgAJ . Unfortunately, there is no response to this post until now. So I want to ask (some of) you directly:

@nmegill @benjub @tirix @david-a-wheeler @digama0 : do you have any hint/opinion/preference which of the alternatives to define diagonal and scalar matrices described in the Google group post https://groups.google.com/d/msg/metamath/4_ZOT4Gddwo/k-lxGMxGBgAJ should be used?

As soon as we have a decision, I'll start to revise section "The subalgebras of diagonal and scalar matrices" in my mathbox. Afterwards, I can move this and other subsections concerning the Cayley Hamilton theorem into the main part as proposed in this issue.

tirix:
﻿Hi Alexander,

Both solutions described in your post seem to have their advantages and drawbacks.

I think my vote may go to your current ‘ DMatALT ‘, since its definition is simpler, the operations are identical, and I would assume that later developments would use more often statements like “M is a diagonal matrix” rather than “the set of diagonal matrices forms a ring”.

However, I have no strong opinion and would agree with any solution.
BR,
_
Thierry

[...]

PR #1335, opened 2019-12-21
Add more information about extensible structures

david-a-wheeler:
Add information from discussions into comments. The text is mostly by
Mario Carneiro and @Benjub. See:
https://github.com/metamath/set.mm/issues/1314


benjub:
Thanks @david-a-wheeler for documenting this. While I see nothing false in your proposal, I would like to modify it a bit but I won't have the time to do it before Christmas. Also, @digama0 is probably the best person for such a summary, if he has time.  I don't know if you can wait, or if you want to merge soon and that after, I propose a new PR ? @nmegill ?
In any case:
* exensible --> extensible
* the hypothetical case of NNmnd has an actual analogue in set.mm, which is preferable to use in these explanations: this is df-zring (by the way: the underlying set of NNmnd is NN0 in the set.mm notation).
* I would remove the reference to manifolds, or at least put it in a more general remark that, roughly, extensible structures work well for "sets with extra structure" (roughly: the objects of concrete categories), but have limitations for more general structures (for instance, for manifolds, where |\`s would not work). However, extensible structures are sufficient for the structures that set.mm currently considers, and offer a good compromise for a goal-oriented formalization.

set.mm
```diff
@@ -179333,6 +179333,73 @@ other components such as a multiplicative operation for rings (and still
...
+  Should we come up with some new structure in the future that we
+  wish ` CC ` to inherit, then we change the definition of ` CCfld ` ,
+  reprove all the slot extraction theorems, add a new one,
+  and that's it. None of the other downstream theorems have to change.
+
+  The restriction operation ` |`s ` is not very useful for manifolds.
+  Manifold embeddings or immersions are something else entirely; you
+  clearly need a function to perform that mapping so the implicit subset
+  morphism doesn't cut it. In that case, the typical solution would be to
+  have a set ( M ManifoldImmersion N ) for encoding all these details.
```
digama0: This paragraph doesn't make much sense out of context. I suggest dropping it.

set.mm
```diff
@@ -179333,6 +179333,73 @@ other components such as a multiplicative operation for rings (and still
...
+  X is a Y via a bunch of forgetful functors) can
+  cause serious performance problems for automated tooling,
+  and the resulting proofs would be painful to look at
+  directly (in the case of Lean, they are long past the level where
+  people would find it acceptable to look at them directly).
+  Metamath is working under much stricter conditions than this,
+  and it has still managed to achieve about the same level of flexibility.
+
+  To create a substructure of a given exensible structure, you can simply
+  use the multifunction restriction operator for extensible structures ` |`s `
```
tirix: Warning, you have to double the ` within math formula embedded in a comment!
(same in other instances below)

david-a-wheeler:
@digana0 - I'm happy to drop it for now, but I'd really like to document more rationale in general. Can you suggest other text instead?

david-a-wheeler:
Thanks for your comments all!  I tried to respond to all comments, and I expanded the description further.

Extensible structures are used all over the place, but the way they're done in set.mm is not something you'll find in a math textbook. Thus, while I think documenting is good everywhere, I think it's especially important that extensible structures be explained (both what they are and the rationale for the approach).

set.mm
david-a-wheeler: Fixed!

set.mm
david-a-wheeler: Dropped.

david-a-wheeler:
(Travis is *really* unusually slow to get going right now, looks like they are working off a backlog.)

david-a-wheeler:
Everything passes, I think this is ready for now. We can always add further improvements later.

benjub:
Thanks @david-a-wheeler for these explanations.  Everything is correct, but I think there are a few things where the emphasis is not put at the right place (for instance the passage about Moore collections). I still think @digama0 is the best person to edit these paragraphs if he has time. Maybe I'll try to propose something in a subsequent PR, but not in the coming days.

david-a-wheeler:
@benjub - if anyone has further improvements, I greatly welcome them. Just create a new pull request with those changes.  I certainly agree that Mario could add much more. But I figured that some improvement was better than none.

benjub:
By the way, @david-a-wheeler, since you wrote about Moore collections: they could be used in your study of binary relations (I know that for the moment, it's only to have the definition written, but maybe you'll develop it), see the last paragraph of my message https://github.com/metamath/set.mm/pull/1286#issuecomment-560756122.  This is a typical process: you first prove that a collection is stable under intersections, and then you can take the "closure", or the "group/ring/etc. generated by" as the intersection of all the sets containing the given set: it is guaranteed to be of the same sort, by stability under intersections, and to be the smallest one. Same thing for closed subsets, convex subsets, sigma-algebras, etc.

PR #1339, opened 2019-12-22
Revision of "Matrices of dimension 0 and 1"

avekens:
* Revision of section "Matrices of dimension 0 and 1" (Section header, new theorems)
* Showing that matrices of dimension 0 and 1 are scalar matrices


benjub:
Fun fact: the ring of scalar matrices is isomorphic to the ring of scalars in strictly positive dimension, but not in dimension 0 (where it is the zero ring). I don't know if this appears in set.mm.

avekens:
@benjub yes, you are right - I could have proven this for all scalar matrices (with dimension greater 0), which includes the case for dimension 1 I have proven now. I put it on my todo list ...

PR #1348, opened 2019-12-25
Revert "Revert "Add ltexpri to iset.mm""

nmegill:
Reverts metamath/set.mm#1347
(fix top date conflict)

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -1,4 +1,3 @@
-"19.21h" is used by "19.21-2".
 "19.21h" is used by "19.21v".
 "19.21h" is used by "hbim1".
 "19.21h" is used by "nf3".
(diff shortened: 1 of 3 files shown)


nmegill:
Sorry for the multiple reversions.  I was in a rush (holidays) and made a mistake.  All should be back to normal, and @jkingdon's #1346 is now merged.

jkingdon:
> Sorry for the multiple reversions. I was in a rush (holidays) and made a mistake. All should be back to normal, and @jkingdon's #1346 is now merged.

Yay! Looks like I have about three more theorems (some with lemmas) in the positive reals (well excluding the least upper bound property, but we only get restricted versions of that anyway so I might put it off for a little while).

After that is signed reals.

PR #1350, opened 2019-12-26
shorten sbabel

wlammen:


wlammen:
Travis checked this within only 3 minutes! Usually it takes more than 10 minutes.

PR #1355, opened 2019-12-28
shorten ralrimdva

wlammen:
The shortening is barely noticeable.  This is more about reducing dependencies on axioms.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17736,6 +17736,8 @@ New usage of "ralbiiOLD" is discouraged (0 uses).
 New usage of "ralimOLD" is discouraged (1 uses).
 New usage of "ralimdaaOLD" is discouraged (0 uses).
 New usage of "ralimdvaOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@wlammen Can you be more specific in the comment, e.g.
"Remove dependency on ~ ax-6 , ~ ax-7 and ~ ax-12 ."
(I'm not sure these are the correct axioms.)

Why did you move ralimdv2 up ?

wlammen:
Can you be more specific in the comment, e.g.
"Remove dependency on ~ ax-6 , ~ ax-7 and ~ ax-12 ."
@benjub For what reason? This is a fundamental change in proof strategy, not some fine tuning.  Once the OLD theorems are removed, the details become irrelevant anyway.

(I'm not sure these are the correct axioms.)
Neither am I.  It takes (a little) attention to get these right.

Why did you move ralimdv2 up ?
In preparation of possible future shortenings through specialization, and to cluster theorems of similar kind.

Issue #1333, opened 2019-12-20
Moving theorems for basic linear algebra from mathboxes into main part 


avekens:
(This issue was originally #1314, but was copied because the discussion in #1314 has gone into a differnt direction)

I almost finished the revisions of the sections with theorems in the context of the Cayley Hamilton theorem in my mathbox. The next step would be to move these sections (and after the definition of diagonal/scalar matrices is clarified, see the discussion in the Google group https://groups.google.com/d/msg/metamath/4_ZOT4Gddwo/k-lxGMxGBgAJ , also subsection "The subalgebras of diagonal and scalar matrices " ) into the main part, Part "11. BASIC LINEAR ALGEBRA").

I would propose to place subsection "The subalgebras of diagonal and scalar matrices" after subsection "11.2.1 The matrix algebra", and sections "Polynomial matrices" and "The characteristic polynomial " after "11.3 The determinant".

Would this be OK? Of course there will be a lot of auxiliary theorems currently in my mathbox, which are used in the proofs of these sections, to be moved to the main part, too.

avekens:
Post to #1314 on 17-Dec-2019, from @avekens:
According to David's Google group post https://groups.google.com/d/msg/metamath/jINlwEjvwOc/IINkUI1sBAAJ, I want to move the Cayley Hamilton theorem and all the theorems required to prove it from my mathbox into the main part of set.mm, as indicated in this issue. Before I can do this, the question how to define diagonal and scalar matrices must be answered, see my Google group post https://groups.google.com/d/msg/metamath/4_ZOT4Gddwo/k-lxGMxGBgAJ . Unfortunately, there is no response to this post until now. So I want to ask (some of) you directly:

@nmegill @benjub @tirix @david-a-wheeler @digama0 : do you have any hint/opinion/preference which of the alternatives to define diagonal and scalar matrices described in the Google group post https://groups.google.com/d/msg/metamath/4_ZOT4Gddwo/k-lxGMxGBgAJ should be used?

As soon as we have a decision, I'll start to revise section "The subalgebras of diagonal and scalar matrices" in my mathbox. Afterwards, I can move this and other subsections concerning the Cayley Hamilton theorem into the main part as proposed in this issue.

avekens:
Post to #1314 on 17-Dec-2019, from @tirix:

> Both solutions described in your post seem to have their advantages and drawbacks.
> 
[...]

PR #1361, opened 2019-12-30
Moving section "The characteristic polynomial" to main part of set.mm

avekens:
Moving the theorems in the context of the Cayley Hamilton theorem to the main part  of set.mm (see issue #1333) finished:
* Section "The characteristic polynomial" moved
* auxiliary theorems moved

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -979,6 +979,30 @@
 "ax-cnex" is used by "cnex".
 "ax-cnre" is used by "cnre".
 "ax-distr" is used by "adddi".
+"ax-frege1" is used by "frege11".
+"ax-frege1" is used by "frege24".
+"ax-frege1" is used by "frege26".
+"ax-frege1" is used by "frege27".
+"ax-frege1" is used by "frege3".
+"ax-frege1" is used by "frege36".
+"ax-frege1" is used by "frege5".
+"ax-frege1" is used by "frege55lem1c".
+"ax-frege1" is used by "rp-6frege".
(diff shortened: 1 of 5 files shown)


nmegill:
I think "telescoping" can be abbreviated "tel", especially since no other label uses that prefix.  So "telescfzgsum" would be "telfzgsum".

avekens:
@nmegill: the corresponding "fsum" theorems use the abbreviation "tscop" - should they also be renamed (e.g.  fsumtscop -> telfsum)? Or shall we use this abbreviation also for my "gsum" variants:
 
* telescfzgsum -> -> gsumtscopfz  (analogous to fsumtscopo )
* telescfzgsumis -> gsumtscopfzis
* telescfzgsum0 -> gsumtscopfz0
* telescfzgsum0is -> gsumtscopfz0is
* telescgsum ->gsumtscop (analogous to fsumtscop )

nmegill:
I think both "telesc" and "tscop" make the label a little long, given that there are several other pieces to the label.  Mario uses "tscop" after "sum", which is somewhat understandable if "sum" is the main focus and "tscop" is just an attribute.  But I still like "tel" as a prefix since in English we say "telescoping sum" and there are no other labels with this prefix (so "show label tel*" will give us only them).  I personally like "telfsum" better than "fsumtscop"; to me it looks easier to remember.  If you and others have no objections, I would like to go with the "tel" prefix for both your and Mario's labels.

Issue #1352, opened 2019-12-27
Add signed reals to iset.mm


jkingdon:
Now that #1351 is submitted, I'm very close to starting signed reals.

The reason I add this as an issue is that I see several choices:

* Dispense with equivalence classes but instead pick a canonical representation of each real (say, the one in which the first number is as close to zero as possible). Partly because set.mm doesn't do this for signed reals, and partly because I'm not even sure it would be possible (in iset.mm) for reals as it would be for fractions, I'm not considering this one any further.

* Use an approach similar to set.mm in which we define notations for "pre-addition" and "pre-multiplication" and use theorems such as http://us.metamath.org/ileuni/oviec.html in key places.

* Use an equivalence class approach which is generally similar to set.mm but which avoids some of the abstractions in favor of more concrete steps (that is, steps which hardcode operations such as positive real addition and positive real multiplication). This is based on the way non-negative fractions are currently developed in iset.mm: see http://us.metamath.org/ileuni/df-nq0.html and following theorems.

I'm leaning towards the third approach although I'm not sure I have a lot of concrete reasons other than "I find it really hard to read and write theorems like http://us.metamath.org/ileuni/oviec.html and the steps which use them". If anyone has any reactions (for example, comparing the non-negative fraction development with, say, the positive fraction development in iset.mm or the signed real development in set.mm), speak up. Perhaps there are even ways to refine some of this further and avoid some of the downsides to the "abstraction-heavy" approach (the fact that oviec has no fewer than 15 hypotheses perhaps is one illustration of why I wonder whether there is another way to organize this) as well as some of the not-great aspects of the non-negative fractions in iset.mm (a certain amount of repetition is what first springs to mind - but whether that is a problem or not is less clear to me - and I haven't really read these proofs since I first wrote them so I may lack perspective here too).

jkingdon:
Signed reals were added by several pull requests leading up to https://github.com/metamath/set.mm/pull/1387

PR #1364, opened 2019-12-31
Replace ovec and pre-addition/multiplication in set.mm

jkingdon:
Over at #1352 I wrote some thoughts concerning how we define signed reals (given positive reals). The immediate prompt was that I will soon be adding signed reals to iset.mm.

But upon reflection: (1) this has almost nothing to do with set.mm versus iset.mm (well a few of the details do but the overall approach is little different between the two), and (2) the question marks are really just a few issues, which are basically http://us.metamath.org/mpeuni/ovec.html and http://us.metamath.org/mpeuni/df-plpr.html (and its analogue for multiplication).

Because of the large number of hypotheses and (related) fact that it is doing a lot, ovec is a bit unwieldy. The pre-operations are also kind of confusing, partly because we only use them this one place - in other places the expression for the pre-operation is expanded out. This change basically just expands it in the remaining places.

The result of this pull request is that set.mm gets 5271 bytes shorter (that's raw file size, so it includes changes to comments and whitespace - although I did this comparison before I added to the list of changes at the top of set.mm).

Because this is only used in the construction of real numbers, it should not disturb anything which merely uses real numbers. The statement of http://us.metamath.org/mpeuni/addsrpr.html is the same (although its proof is different), so everything after that is unaffected.

The pull request also includes two small changes to iset.mm, which are just things I noticed while doing this.


nmegill:
I like this.  Nice work!

jkingdon:
> I like this. Nice work!

Cool. There might be further fine tuning possible (like maybe there are further lemmas which could be shared between the addition and multiplication cases), if someone wants to look at this further. But I'm happy in terms of having something that I think will be easier to start from as I add this to iset.mm.

PR #1369, opened 2020-01-01
One New, Ten Minimized sandbox theorems

arpie-steele:


arpie-steele:
Thanks, @benjub for reminding me to minimize.

PR #1370, opened 2020-01-01
shorten r19.21bi

wlammen:
This pull request also contains a shortening of rgen2a.  This theorem must no be shortened, but it has room for technical corrections, though, even without compromising the underlying proof idea. I publish the shortening in my mathbox as wl-rgen2a.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17724,6 +17724,7 @@ New usage of "qlaxr3i" is discouraged (0 uses).
 New usage of "qlaxr4i" is discouraged (0 uses).
 New usage of "qlaxr5i" is discouraged (0 uses).
 New usage of "quoremnn0ALT" is discouraged (0 uses).
+New usage of "r19.21biOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@wlammen @nmegill : ~wl-rgen2a indeed preserves the idea of the proof, so I think it can be promoted as the new ~rgen2a.

nmegill:
> ~wl-rgen2a indeed preserves the idea of the proof, so I think it can be promoted as the new ~rgen2a.

I agree.  My primary concern was to have a practical  example of using ~dvelim* to convert a $d to "distinctor" antecedent (other than for variations of ~dve* theorems), then using ~pm2.61* to eliminate the $d entirely.  That is the reason for the "(Proof modification is discouraged.)".   I have been surprised at how seldom ~dvelim* ended up being used.

wlammen:
@nmegill This is in part due to the fact that dvelim* uses ax-13, which rules against its usage prior to this section. I always kept an eye on it.  And many basic theorems already have a distinctor version, so its not needed then.

PR #1372, opened 2020-01-02
Finite Intersection property of a class.

arpie-steele:
Some exposition.
Four sandbox theorems, fiinfi moved earlier as dependencies clarified.


benjub:
@arpie-steele Nice additions.  The nonempty finite intersection property (equivalently, the pairwise intersection property), as well as the intersection property, are fundamental and ubiquitous concepts. Apparently, collections with the latter property are sometimes called "Moore collections", which have a dedicated subsection in set.mm which could be helpful to you (the phrase "Moore collection" is not very common, so I prefer "collection with the intersection property", or more explicitly, "collection closed under intersections").
Maybe for some of your theorems, you can prove directly the intersection property instead of the pairwise version. Also, set.mm may contain a theorem on general associativity of intersections (or else it should be added), which could be useful (for instance for ~fiinfi). @avekens @tirix @digama0 may have useful advice on this.
Recently, I needed a quick way to formulate the intersection property, and I used: http://us2.metamath.org/ileuni/bj-indint.html from which it follows directly that the intersection of all inductive classes is the smallest inductive class. This is probably your plan for the transitive closure.

nmegill:
@benjub There are many theorems in set.mm about Moore collections/closures/systems ('sea * moore/c') mostly by Stefan O'Rear.  Is this what you are describing?

benjub:
@nmegill Yes, I think these could be useful.  These are collections of sets, and not of classes, so for very specific tasks they may not be sufficient (in which case, things like iset/bj-indint linked above may be used), but in most cases, these should be enough. There may be some overhead in proving that a given collection of sets is a Moore collection on a set A, but I think it is worth it.  After all, it is often sufficient to consider collections of sets, e.g. "the set of transitive relations on the set x" versus "the class of transitive relations". This is what I called a few times "internalization", and it works well with the definition of BinRel, TransBinRel etc. that I proposed in https://github.com/metamath/set.mm/pull/1286#issuecomment-560756122

PR #1375, opened 2020-01-02
shorten r19.21t

wlammen:
19.21v can be proven from a significantly reduced set of axioms.  So far doing so had no impact on other theorems.  Users of 19.21v somehow depend on all of predicate logic axioms anyway.  This pull request showcases the first theorems (r19.21v, r19.32v) where the smaller set of axioms is visible.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17727,6 +17727,7 @@ New usage of "qlaxr5i" is discouraged (0 uses).
 New usage of "quoremnn0ALT" is discouraged (0 uses).
 New usage of "r19.21biOLD" is discouraged (0 uses).
 New usage of "r19.21tOLD" is discouraged (0 uses).
+New usage of "r19.21vOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
@benjub I actually wait a bit for your reworking of predicate logic.  Is it still planned/under way?  I hope rewriting 19.21v does not disturb the process.

benjub:
Thanks for the heads-up @wlammen. Your shortening of 19.21v looks "future-proof".

My reworking of df-nf (i.e., define F/ x ph as ( E. x ph -> A. x ph ) , that is, use ~nf2 as the definition) is indeed still planned, but unfortunately I do not have the time right now. Maybe by the end of January (depending on my day job...).

By the way, you might be able to shorten the proofs of ~cases and ~cases2, which I added recently (~cases is credited to @nmegill because I simply extracted it as the FOL part of the proof of ~elimif).

PR #1376, opened 2020-01-02
Misc

benjub:
Delta_0 classical logic and separation implies classical logic; this might be of interest to @jkingdon


jkingdon:
> Delta_0 classical logic and separation implies classical logic; this might be of interest to @jkingdon

Ah, cool stuff.

We also might get CZF plus excluded middle equals ZF, or IZF plus excluded middle equals ZF, some; day.

PR #1382, opened 2020-01-05
iset.mm: Addition and multiplication of signed reals are commutative, assocative, and distributive

jkingdon:
We developed most of the machinery for this when we were doing positive fractions.

The most complicated thing, perhaps, in this pull request is intuitionizing `caovlem2` (which is used only in `mulasssr` in set.mm). But that works more or less as expected (modulo a bit of extra work using commutativity to convert between right distributivity and left distributivity).


jkingdon:
Travis failed the first time (difficulty downloading metamath or something like that), so I closed and reopened the pull request, and then it passed.

PR #1386, opened 2020-01-06
Manifolds

tirix:
Proposing a definition of topological manifolds.


benjub:
Hi @tirix .  Here are some comments, emphasizing the mathematical point of view but ignoring the ease of "codability" in set.mm and usefulness there, which I leave to your judgment.

First, I would rename "ManTop" to "TopMan", which is more common (for many categories, take the first three letters of each word and concatenate -- then, the classical joke is about Banach analytic manifolds).

As for the definition itself, you have been wise to concentrate on the topological case first (so no need for atlases, which would require to make many more choices, and there are already enough choices to make).  Side remark: a topological manifold can also be seen as a manifold-with-atlas of type "Top", but this could be a later theorem once manifolds-with-atlases are defined.

At its most general: a (topological) manifold is a topological space locally homeomorphic to a topological vector space, with possibly additional properties.  Some have considered topological modules instead, but let's not go there.

These properties are Hausdorffness, paracompactness, second countability.  I advise to not use second countability but to define manifolds as paracompact Hausdorff spaces.  There is no big difference in the finite-dimensional case: a paracompact Hausdorff n-manifold is second countable if and only if it has a countable number of connected components (and a second countable Hausdorff n-manifold is paracompact).  Importantly, every discrete topological space is a paracompact Hausdorff manifold (of dimension 0).  I know that many textbooks use second countability, but this is mostly to avoid a supposedly more abstract notion (paracompactness). This could be coded later, and you could simply put a "TODO: replace second countability with paracompactness".

You define real n-dimensional manifolds, but maybe they should be defined in terms of more general manifolds ?  This could of course come later, and maybe it's better at first to focus on this subclass, which is by far the most important (with complex n-manifolds).  The definition I gave above is more general in two ways: first, the model spaces are any topological vector spaces (and not necessarily real n-dimensional ones), and second, the model space may vary with the point (it is then a theorem that it can be chosen to be locally constant) -- when the model space can be chosen constant, the manifold is sometimes called "pure", and any real or complex n-manifold is pure.

The precise definition should be: "each point has an open neighborhood which is homeomorphic to an open subset of a topological vector space" -- and not, "to a topological vector space", although in the n-dimensional case, this is equivalent.

I see that your first commented-out theorem is that the real line is a manifold.  Mabye deduce it from the theorem that a (paracompact Hausdorff) topological vector space is a topological manifold (modeled on itself); in your case, ( TopOpen \` ( EEhil \` N ) ) is an N-manifold.  From which you deduce that RRfld is a 1-manifold.

tirix:
Hi Benoît,
Thank you very much for your comments!
About the naming, I chose ` ManTop ` because the class actually really defines manifold topologies, and not topological manifolds. These are topologies i.e. just collections of open sets, with no additional structure.
The naming ` TopMan ` could be used for topological spaces (i.e. structures) which topologies are manifold topologies. This way, topological vector spaces ` TopVec ` could also be seen as ` TopMan ` once they fulfill the required additional conditions.
I'm not sure yet if I shall define both ` ManTop ` (bare topologies) and ` TopMan ` (structures), as this is probably a confusing distinction, or move on directly to structures. It seemed to me that there was already much to be said about the topologies only.

Paracompactness is not defined yet in set.mm, so as you suggest, taking the shortcut to define it later would help speeding up, and it seems it would not harm if it is possible to painlessly change the definition of manifolds once paracompactness is defined

I will see how I can change the definition to be more general and specify homeomorphism to an open subset of a topological vector space. In that case, the dimension would not be specified, as the manifold would simply inherit the dimension of the vector spaces (yet to be proved that they all have to be the same, whatever the chosen neighbourhood)

nmegill:
@benjub thank you for looking at this.

@tirix I don't know much about manifolds and can't really provide a meaningful comment.  Would you prefer to merge your PR now and make the changes you propose later, or should we wait for your changes?

tirix:
@nmegill this new definition is in my Mathbox, so I think we can treat this as a scratchpad/not mandatorily finalized, and merge. 
This will also allow others, such as Gérard, to see the proposed changes and comment.

Thank you!

tirix:
@digama0  I actually started to write a definition of paracompact spaces (head first as always) when I discovered that Jeff Hankins already had a definition of locally finite covers, [LocFin](http://us2.metamath.org/mpeuni/islocfin.html), and of refinements, [Ref](http://us2.metamath.org/mpeuni/isref.html).

Also, it appears to me that we could write "locally finite" in set.mm as ` N-Locally Fin ` . Am I correct?

What is your opinion? Shall I move Jeff's definition of refinements to the main part of set.mm?

tirix:
Another question: I see that Bourbaki requires paracompact spaces to be Hausdorff, but that does not seem to be the case in general. What do you suggest @digama0 @benjub ? 
Of course it does not matter for manifolds which are already required to be Hausdorff.

benjub:
I suggest to define paracompactness without Hausdorffness assumption.
Indeed, Bourbaki assumes Hausdorffness (like they do for "compact") since
without Hausdorffness, paracompactness is not a very useful property (even
more so than for compactness), but they are pretty alone to do that.  Since
compactness in set.mm does not (I presume) assume Hausdorffness, and since
you want "compact implies paracompact", you should not assume Hausdorffness
for paracompact spaces either. Additionally, this is better to fight
ambiguity: in most theorems, you will write "paracompact Hausdorff space"
(either in plain English in the comments or in the formal language), which
is non-ambiguous. If you had defined paracompactness assuming
Hausdorffness, then you could write "paracompact space", but this would be
ambiguous (i.e., at best, the reader would have to go back and check the
definition you are using, and at worst, he would assume Hausdorffness is
not necessary).

Sorry for the long verbiage !

tirix:
Indeed, in set.mm, compact does not imply Hausdorff. I'll follow the same logic for paracompactness.

tirix:
Hmm, after consideration, the definition of paracompactness requires a _covering_ to be locally finite, so my idea to use ` N-Locally Fin ` would not work, since that works only for topologies, and the covering itself is not a topology, but simply a collection of open sets.

On the other hand, maybe we could define a generic term for the existence of a refinement of any open cover, with given properties. This may allow to simplify definitions of compact, paracompact, Lindelöf, and metacompact.

PR #1387, opened 2020-01-07
iset.mm signed reals from 00sr to mulgt0sr

jkingdon:
The intuitionizing ranges from zero to a whole lot of adding `something e. P.` to a lot of lines. But even in the latter case, it is pretty much all the same kind of intuitionizing, in which we need to evaluate functions (in this case operations) within their domains (in this case `P.`).


jkingdon:
I've pushed changes to expand this pull request to finish the signed real section. Turns out the signed real theorems after mulgt0sr are best dealt with later (after df-c). The reason (as described in the text to mmil.html which is now part of this pull request) is that they relate to supremums, a topic which we've been putting off (and which is different without excluded middle).

Issue #1368, opened 2020-01-01
Move friendship theorem into main part of set.mm


avekens:
According to Google group discussion https://groups.google.com/d/msg/metamath/jINlwEjvwOc/IINkUI1sBAAJ, the friendship theorem and related theorems should be moved from AV's mathbox into the main part of set.mm.
More concretely, section "21.25.6 The Friendship Theorem" and most of section "21.25.5  Graph theory" should be moved.

avekens:
Task done with PR #1415.

PR #1388, opened 2020-01-07
Subsection "Closed walks" (1. part) moved

avekens:
Main part - see issue #1368:
* auxiliary theorems moved from AV's mathbox to main part of set.mm
* fzossrbm1 revised (fzossrbm1OLD still to be replaced later)
* subsection "Closed walks" (1. part) moved from AV's mathbox to main part of set.mm
* new class symbol for a connector (small circle) - to abstract from the meaning "plus" or "times"

AV's Mathbox:
* new subsection "Magmas and semigroups" with new definitions for Mgm and SGrp and related theorems
* new subsection "General rings ("rngs")" with new definitions for Rng and related theorems (TODO: replace placeholder 9999 for page number in [BourbakiAlg1] @benjub do you have (access to) the french version of Bourbaki's "Algèbre" to tell me the page number, please?)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -6046,6 +6046,11 @@
 "funadj" is used by "adjeq".
 "funadj" is used by "funcnvadj".
 "funcnvadj" is used by "adj1o".
+"fzossrbm1OLD" is used by "clwlkisclwwlklem1".
+"fzossrbm1OLD" is used by "redwlk".
+"fzossrbm1OLD" is used by "redwlklem".
+"fzossrbm1OLD" is used by "signsvtn0".
+"fzossrbm1OLD" is used by "wwlkm1edg".
 "gen11" is used by "ax6e2eqVD".
 "gen11" is used by "ax6e2ndeqVD".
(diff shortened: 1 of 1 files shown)


benjub:
You got the page references from Gérard. Another remark concerning non-unital rings: I would choose this term ("non-unital ring") as the main term, and only mention rng and pseudoring as variants. The reason is that "non-unital ring" is self-explanatory. On the contrary, if I see "pseudoring", I know it is a ring with some property removed, but I hesitate between non-unital, non-associative, no negatives...  (additionally, "pseudo" is over used: basically, for any mathematical object "smurf", you have authors defining pseudosmurf, almost-smurf, semismurf, hemismurfn quasismurf... with often no consistency at all... and some authors chain these: almost-pseudosmurf, etc.).

Here, this would imply:
General rings ("rngs")  --> Non-unital rings
a _rng_ (or pseudo-ring or non-unital ring) --> a _non-unital ring_ (or rng, or pseudoring)

And you can add: As almost always in mathematics, "non-unital" means "not necessarily unital".

PR #1390, opened 2020-01-07
Modify conventions text, e.g., focus on important part

david-a-wheeler:
This modifies the label conventions text per the
mailing list conversation to emphasize the important part.
While we're at it, it improves the text about syntax fragments
so that it flows better (including with the new text).


avekens:
Hi David,
great, a very good enhancement of the conventions. I have only two small remarks/proposals for improvement:

- "Focus on the important part of the conclusion." should be placed after "Syntax label fragments." (because the first is a specialization of the second)
- a link to theorem cos0 (~ cos0) should be added, or the first "cos0" should be replaced by ~ cos0.

Alexander

david-a-wheeler:
@avenkens - thanks, and both of your suggestions make sense to me. I am away from a useful keyboard, but when I get a chance I plan to do that.

benjub:
Thanks @david-a-wheeler for the documentation. In addition to @avekens 's remarks, there is a typo:
the conclusion and not --> the conclusion and does not
(or alternately, remove the "s" in "provides"!).

david-a-wheeler:
@avenkens @benjub - Thanks for the additional comments. I'll work on another pull request to try to make those refinements.

david-a-wheeler:
I added more tweaks here: https://github.com/metamath/set.mm/pull/1392

My hope is that label names will increasingly not be a mystery :-).

PR #1392, opened 2020-01-08
More improvements on label convention info

david-a-wheeler:
Further improve the "conventions" section that
explains the label conventions. Most changes are due
to comments from @avekens and @benjub. See:
https://github.com/metamath/set.mm/pull/1390


avekens:
Hi David,
thanks for considering our remarks in #1390.
Concerning your additional convention for the unique ids for hypotheses ("these ids are usually consecutive integers starting with 1"), however, I think this is not the currently used and recommened style (anymore). Instead (or additionally), a letter is used, corresponding to or indicating the (main) class within the hypothesis. See, for example,  ~lspprss:

    lspprss.s $e |- S = ( LSubSp ` W ) $.
    lspprss.n $e |- N = ( LSpan ` W ) $.
    lspprss.w $e |- ( ph -> W e. LMod ) $.
    lspprss.u $e |- ( ph -> U e. S ) $.
    lspprss.x $e |- ( ph -> X e. U ) $.
    lspprss.y $e |- ( ph -> Y e. U ) $.

I remember darkly that there was already a discussion somewhere about this...

Alexander

nmegill:
> See, for example, ~lspprss:
lspprss.s $e |- S = ( LSubSp ` W ) $.
lspprss.n $e |- N = ( LSpan ` W ) $.
lspprss.w $e |- ( ph -> W e. LMod ) $.
lspprss.u $e |- ( ph -> U e. S ) $.
lspprss.x $e |- ( ph -> X e. U ) $.
lspprss.y $e |- ( ph -> Y e. U ) $.
I remember darkly that there was already a discussion somewhere about this...

I couldn't find the discussion, but I started doing this a few years ago just to make it easier to remember the hypotheses while entering the proof of this kind of theorem.  Some theorems have obvious label letters like this one; others don't, and numbers are fine for them.  Once a proof is complete, hypothesis labels can be changed without side effect or any impact on anyone else, and anyone could tidy them up someday if they have a strong enough urge. :)   This probably isn't important enough to be a requirement, but it can be a helpful suggestion.

I'm merging this PR and any guidelines like the above can be tweaked in the next PR.

avekens:
I found the corresponding Google group discussion: https://groups.google.com/d/msg/metamath/NQTyotIp6V4/ZlEmuWlaDQAJ. There is also another helpful convention for the labels of hypotheses:

- In the case of a $e shared by multiple theorems, we've usually chosen the label of the first $p after it for the prefix.  
- Although when a $e is shared by both a lemma and its main theorem, we typically choose the main theorem label for the prefix.

benjub:
I agree with these more explicit labels for $e-hypotheses. Plus, they are written only once in set.mm (since proofs are in the compressed format), so longer labels are less of a problem.
I've been applying this for some time, for instance:
```
2spim.nfx | ⊢ Ⅎxχ
2spim.nfz | ⊢ Ⅎzχ
2spim.1 | ⊢ ((x = y ∧ z = 𝑡) → (ψ → χ))
2spim | ⊢ (∀z∀xψ → χ)
```
or
```
bdzfauscl.bd | ⊢ BOUNDED φ
bdzfauscl | ⊢ (A ∈ 𝑉 → ∃y∀x(x ∈ y ↔ (x ∈ A ∧ φ)))
```
both in iset.mm

PR #1397, opened 2020-01-09
Add more explanation about dvmulbr / dvmul

david-a-wheeler:
These explanations are derived from a mailing list
discusion on 2020-01-09.


benjub:
Looks strange... I'm replying in the discussion group. (Note that "equals x" might be "at x equals A"...)

david-a-wheeler:
@benjub - okay. I took Mario at his word. If it turns out it's a subtle error, then we clearly need to fix it and maybe this text is wrong.

digama0:
While I'm happy to have my words copied about, you should keep in mind that different contexts require different phrasing, and sometimes this kind of documentation gets a bit "copy-pasta". @benjub's correction is accurate.

As for this edit, it's a bit odd. Why here and not on some other statement? I responded to Filip's question about dvmulbr but you could ask exactly the same question about dvaddbr, or really any dv theorem that uses relations (of which there are several). I think this comment shouldn't be here at all. If you want to make a remark about the relational use of df-dv, it should be on eldv, but it's not clear to me that anything needs saying in the first place. So I'm inclined to just reject this edit.

david-a-wheeler:
I did tweak the text to reduce "copy-pasta"ness.

It's not obvious where to put this kind of comment, and I thought it'd be useful to have a comment like this *somewhere* than nowhere at all. But if people don't think it's useful, we can remove the verbage.

Even if we don't include this, I *do* think it'd be useful to have hyperlinks like this between the "br" and non-"br" forms, like between dvmulbr / dvmul. There are many ways you can end up at one theorem and not realize that there's a related theorem. But if you don't want this comment, I can try to generate such hyperlinks separately.

david-a-wheeler:
If you don't want the detail, how about something like this:

In dvmulbr, say:

> For the (simpler but more limited) function version, see ~ dvmul .

In dvmul, say:

> For the (more general) relation version, dvmulbr .

And then do the same for the other br / not-br theorems?

If you just see one theorem, it's not necessarily obvious what the "other" theorem is, and in general I think we should be adding hypertext links between related theorems to ease navigation. As set.mm gets bigger, these aids become increasingly helpful.

benjub:
I agree with @david-a-wheeler's latest proposals:
> > For the (simpler but more limited) function version, see ~ dvmul .

> > For the (more general) relation version, see ~ dvmulbr .

> in general I think we should be adding hypertext links between related theorems to ease navigation

I agree (without excess: for instance, linking to the 10 related theorems will be less useful than linking to the main theorem of the family).

david-a-wheeler:
@benjub said:

> I agree (without excess ...)

Well sure, anything can be overdone. But I don't think what I have in mind is excessive. What I *meant* was that:

* dvmul and dvmulbr would cross-reference each other
* dvadd and dvaddbr would cross-reference each other
* dvco and dvcobr would cross-reference each other
* dvcj and dvcjbr would cross-reference each other

All using the text format listed above.

I think it's important to help people find related statements.

david-a-wheeler:
Okay, I'm closing this pull request & will open a new pull request with this change instead.

PR #1400, opened 2020-01-11
Revise 19.23v

wlammen:


avekens:
I wonder why ~ 9.23vOLD should be removed (in the future), although it has a much shorter proof than  ~19.23v. And did we really agree that reduced dependencies on axioms has higher priority than a shorter proof (if so, then only in PART 1 of set.mm)? This is, at least, not documented in the header of set.mm ("6. Acceptable shorter proofs").   There, it is written "A few theorems have a longer proof than necessary in order to avoid the use of certain axioms, for pedagogical purposes, and for other reasons.  These theorems will (or should) have a "(Proof modification is discouraged.)" tag in
their description. " - by this, ~ 9.23v should be tagged, and not ~ 9.23vOLD...

I think (and @benjub too - as far as I can remember) ~ 9.23vOLD should be renamed to ~ 9.23vALT and should be kept (and not deleted after a year, as it would be done with  ~ 9.23vOLD.

Maybe there are other OLD theorems (resulted from reducing the dependencies on axiomst) which should be kept...

wlammen:
@avekens 
> And did we really agree that reduced dependencies on axioms has higher priority than a shorter proof
See http://us2.metamath.org:88/mpeuni/conventions.html "Minimizing axioms..."
The same paragraph allows keeping the old proof of 19.23v as an alternative, but in this particular case I am opposed to that, because deriving a *v from * is pretty standard (simply use nfv), and regularly doing so will produce just a fallout of ALT proofs (often far separated from the main proof)  without much extra insight.  If it lends a helping hand to beginners, a short reference to an example
implementation in the comment should suffice IMO.

avekens:
The convention "Minimizing axioms and the axiom of choice" does not say anything about the length of the proof. So there is a small inconsistency between this convention and "6. Acceptable shorter proofs". 

Concerning ~19.23v, I agree with you that ~19.23vALT would not be very useful, after I had a closer look at the related theorems. I would suggest, however, to add a link to ~19.23 in the (new) comment  of ~19.23v, for example:
`Special case of Theorem 19.23 of [Margaris] p. 90, see ~ 19.23 .  (Contributed by NM, 28-Jun-1998.)  Reduce dependencies on axioms.  (Revised by Wolf Lammen, 11-Jan-2020.) $)`

wlammen:
> The convention "Minimizing axioms and the axiom of choice" does not say anything about the length of the proof.

Oversight? "We prefer proofs that depend on fewer and/or weaker axioms, even if the proofs are longer."

> So there is a small inconsistency between this convention and "6. Acceptable shorter proofs".

I don't see a substantial contradiction: "Usually I will automatically accept shorter proofs that... (4) use no additional axioms."

> I would suggest, however, to add a link to ~19.23 in the (new) comment of ~19.23v, for example:
> `Special case of Theorem 19.23 of [Margaris] p. 90, see ~ 19.23 . (Contributed by NM, 28-Jun-1998.) Reduce dependencies on axioms. (Revised by Wolf Lammen, 11-Jan-2020.) $)`

This opens a can of worms. In parts I covered so far, we have lots families of related theorems. The ralim* part, for example, covers 26 highly specialized theorems of similar kind.  So far I did not bother linking them all to a base theorem explicitly.  Instead, the relation is often implicitly given, because the base theorem appears first in a bulk, is used in the proof, or has the simplest name of a kind.

I don't know how much user experience is improved, if a direct link to the base theorem(s) is always provided in the comment.  If one agrees on such a feature, I will follow suit, but then it should be embedded in a standard phrase.

nmegill:
> The convention "Minimizing axioms and the axiom of choice" does not say anything about the length of the proof. So there is a small inconsistency between this convention and "6. Acceptable shorter proofs".

While it is important to have such conventions to try to maintain some consistency in set,mm, there are borderline cases like this where it's not all that black and white.  (And the conventions were written by me, and I am not perfect nor could anticipate all possible cases.)

As a general rule, I think that the use of fewer axioms is more interesting and would have precedence over shorter proofs.  This is almost certainly the case when we can avoid the axiom of choice.  An example is ~ hartogs vs. ~ ondomon .  (Actually ~ ondomon should probably have "New usage discouraged".  I didn't name it hartogsALT - although perhaps we could - because the important contribution of Hartogs is the AC-free proof more than the theorem.  I chose not to delete ~ ondomon because it is a good illustration of how using AC can dramatically shorten some proofs.)

I almost always would prefer a longer proof in FOL that uses fewer axioms, because it makes the theorems more useful when we want to study the independence of axioms, or the possibility of using fewer or weaker or alternate axioms (although I think the current axiomatization, which evolved over many years, will be stable for a long time).

However, in early set theory, such as the r19.* stuff, it seems less important to focus on using fewer FOL axioms, because they would rarely be helpful for studying FOL axiomatizations, since they cannot be "imported" back into FOL.  That isn't to say that in principle they might help those borrow from set.mm for weaker axiomatizations like iset.mm, but it seems unlikely because those use such different FOL axiomatizations from the start.  So in early set theory my personal preference would usually be to have shorter proofs take precedence over fewer FOL axioms.  But I won't insist on that and leave it to the individual contributors to make their own judgment calls; both cases can be interesting.  

> 
> Concerning ~19.23v, I agree with you that ~19.23vALT would not be very useful, after I had a closer look at the related theorems. I would suggest, however, to add a link to ~19.23 in the (new) comment of ~19.23v, for example:
> `Special case of Theorem 19.23 of [Margaris] p. 90, see ~ 19.23 . (Contributed by NM, 28-Jun-1998.) Reduce dependencies on axioms. (Revised by Wolf Lammen, 11-Jan-2020.) $)`

Sounds reasonable.

wlammen:
This pull request is all about a FOL theorem 19.23v, not about r19.23v.

avekens:
Thanks, @nmegill and @wlammen for your explanations. To satisfy me entirely, it should be mentioned in "Minimizing axioms and the axiom of choice" that this only/mainly holds for the FOL part, e.g. "We prefer proofs that depend on fewer and/or weaker axioms, even if the proofs are longer, at least in the FOL part."

Concerning the links, I would like to have them, because they help to find the related theorem more easily. They need not to be added in a big bang, but if a theorem is modified anyway they should be added. An alternative for the given example could be 
`Special case of Theorem ~ 19.23 of [Margaris] p. 90. (Contributed by NM, 28-Jun-1998.) Reduce dependencies on axioms. (Revised by Wolf Lammen, 11-Jan-2020.) $)`
i.e. only adding a single "~"!

nmegill:
> This pull request is all about a FOL theorem 19.23v, not about r19.23v.

Sorry, I meant to discuss the issue in general and am not sure why I typed "borderline cases like this" - it should have been just "borderline cases".  I edited my comment.

Yes, there is no issue for 19.23v, and a longer proof with fewer axioms is always preferable there.

benjub:
I would always give precedence to fewer axioms, but if in later parts of set.mm, @nmegill prioritizes shorter proofs, at least keep proofs using fewer axioms as *ALT.

By the way, when we say "fewer axioms", there are also cases where a theorem is provable from axioms A and B, and also from axioms A and C.  In these cases, I also think both proofs should be kept (using *ALT).

PR #1405, opened 2020-01-12
Document why some sections are deprecated

david-a-wheeler:


nmegill:
Use double tilde to escape tilde i.e. cgr~circgrp s.b. cgr~~circgrp

david-a-wheeler:
@nmegill - you're absolutely right. Fixed, hopefully that's the only problem :-).

PR #1406, opened 2020-01-12
Explain how we handle "undefined/junk" cases

david-a-wheeler:
Modify the "conventions" section to explain how we handle
undefined/junk cases, per the Metamath mailing list
discussion in January 2020 title 'Handling "undefined terms"'.


nmegill:
I think this is premature, pending more discussion on the issue.  I don't necessarily agree that "junk" consequences are bad, and they can significantly shorten proofs in some cases.  set.mm would be much larger if, for example, we had to condition ~ fvex on the argument being in the domain of the function.  It is impossible to derive a contradiction from sound definitions (i.e. that pass the definition check), assuming ZFC is consistent, and I don't see the point of all the extra busy work and huge increase in set.mm size that would result from restricting all definitions in the way proposed.

In the case of iota, we are defining it exactly as Quine does, whose definition also evaluates to the empty set when uniqueness is not satisfied.  He has no problem with that and I don't see why we should.

david-a-wheeler:
> I think this is premature, pending more discussion on the issue.

Ok, but I think the question will be continuously re-raised by new people, so we ought to write down *something* when we have something to say.

> I don't necessarily agree that "junk" consequences are bad, and they can significantly shorten proofs in some cases. set.mm would be much larger if, for example, we had to condition ~ fvex on the argument being in the domain of the function. It is impossible to derive a contradiction from sound definitions (i.e. that pass the definition check), assuming ZFC is consistent, and I don't see the point of all the extra busy work and huge increase in set.mm size that would result from restricting all definitions in the way proposed.

A fair point.

I think it *is* weird in some cases, though, since humans might misunderstand what's being proven. So at least *noting* an alternative (using discouraged) is, I think, a good idea.

> In the case of iota, we are defining it exactly as Quine does, whose definition also evaluates to the empty set when uniqueness is not satisfied. He has no problem with that and I don't see why we should.

In that case, Quine is taking approach #1 (make it null). So in that case, the argument may be "others to it too". Perhaps there's a better example.

nmegill:
OK, I'll accept this.  If some rewording is needed we can do that later.

benjub:
I appreciate @david-a-wheeler documenting these things (as I wrote, an objective of my post was indeed to suscitate informative answers).  But in that case (as in another recent one), I think it was indeed a bit premature, since the discussion on the google group was not even over.

>  which is useful when \` F \` is a family of sets. (by this we mean it's a set set even in a type theoretic interpretation.)

I think it would be better to write "which is useful when the values of \` F \` are nonempty sets."

> definitions like ${ $d \` y x \` $. $d \` y ph \` $. df-iota $a \` |- ( E! x ph -> ( iota x ph ) = U. { x | ph } ) \` $. $}.

No need for the dv conditions here (since y does not appear anymore), so
> df-iota $a \` |- ( E! x ph -> ( iota x ph ) = U. { x | ph } ) \` $.

is enough. My bad because the typo originated from me, but your bad for not checking it !

definition check --> definition soundness check (I think it is clearer)

I think these paragraphs are a bit colloquial, and phrases like "knowledgeable people such as @digama0 have never seen this happen" or "@nmegill does not see this as..." feel a bit strange. It is becoming a verbatim of the discussion group.

But I reiterate the main part: overall, I think that you're documenting this is useful.

digama0:
> > which is useful when ` F ` is a family of sets. (by this we mean it's a set set even in a type theoretic interpretation.)
> 
> I think it would be better to write "which is useful when the values of ` F ` are nonempty sets."

Actually, they need not be nonempty sets, for instance `( M ... N )` is the set of integers between M and N and may be empty even within the domain, if M is greater than N, but reverse closure ensures that it makes sense (and evaluates to the empty set) when applied to non-integers.

benjub:
Of course they may be empty.  It's just more useful when values are nonempty (since then, for all B in the domain, you will has an A such that A e. ( F \` B ), which is the antecedent in the closure theorem. Anyway, the important part of my proposal was adding the phrase "the values".

PR #1407, opened 2020-01-13
Add more about undefined/junk cases

david-a-wheeler:
This commit adds some of information about undefined/junk
cases that Norm raised in:
https://github.com/metamath/set.mm/pull/1406 ;

Over the years we have repeatedly had questions/discussions
about undefined/junk cases. This tries to document *why* things
are the way they are, to answer those questions ahead of time.


benjub:
See my comment at #1406

david-a-wheeler:
@benjub - I'll try to respond to some of your comments on #1406.

david-a-wheeler:
In #1406 @benjub said:
> I appreciate @david-a-wheeler documenting these things (as I wrote, an objective of my post was indeed to suscitate informative answers). But in that case (as in another recent one), I think it was indeed a bit premature...

This discussion has been reoccurring every few years for many years (not always with the same players). So it's time to try to capture *something*. If you have suggestions on how to make it "less colloquial", let me know.

I make no apologies that some of it sounds like the mailing list discussions (since that's obviously where it's from). My goal is to move some of the key points from the discussions into something "more permanent" so that people will get these questions answered ahead-of time. In this case, the issue of "how to handle undefined/junk cases" is a common issue in all formal mathematics systems, so it's an obvious question to ask; we need to have an answer for it.

benjub:
I wrote my appreciation of your documentation endeavor twice in my comment, so I certainly require no apology. I was simply giving suggestions (probably some of them good and some others bad) to improve parts of it. As for the colloquial/formal style, you certainly have a better grasp than me, so if after re-reading it, you deem it fine, I certainly trust you.

As for the "a bit premature", I was simply meaning that you may next time simply wait for the discussion to settle down (say, one or two days of inactivity on the thread). This precise topic (undefined terms/junk theorems) occurred a few times indeed, and I gave links in my initial post to two previous discussions.

PR #1409, opened 2020-01-14
revert to old proof of nrex not depending on ax-5

wlammen:


nmegill:
I noticed you have reverted several of these to eliminate dependencies.  As I said, you can use your own judgment for r19.*, shorter proof vs. fewer axioms,  I'm just curious if you have a specific goal.

wlammen:
@nmegill I used shortening and revision as a way to learn about a topic. This is a sort of homework for me.  So I usually have no specific goal in the beginning, simply because I do not overlook a particular theory then, and cannot tell important/fundamental theorems from accessory/derived ones. Of course it helps when I encountered a notation before, say in school, or in my calculus book. But any reasoning about what I did, is hindsight thinking, nothing planned.

So what can I say in favour of my recent work? A majority of theorems of the restricted quantifier section do not depend on ax-8, ax-9 or ax-ext, which means, A e. B need not have the usual meaning, but can be any arbitrary binary relation between classes.  df-ral is then just a shorthand notation for a unspecific expression. In this light compare df-clel with df-rex, and you see that many proofs from the rex section must carry over to clel using some primitive translation.  Is it possible to formalize this, using a new variable type as a placeholder for binary operations?  I am pretty sure this has been decided before, but for me this is a reason to have a closer look.  In the light of these ideas, df-ral/df-rex appear to be more primitive concepts than df-clel, and so IMO should come before it.  If it is closer to FOL, then be more careful about where the roots are.

Formally, I followed the rules on the conventions.html page blindly. In the case of nrex, I accidently violated them by adding a dependency on ax-5 that wasn't there before.

As for the future, I will finish the restricted quantifier section, and will then abandon the idea of reducing FOL axiom usage outside of strict FOL.

benjub:
Regarding restricted quantifiers: I agree with @wlammen that they can be considered part of FOL-with-one-binary-predicate ("the first order theory of one binary predicate with no axiom") and be placed earlier in set.mm.
I think there should be, either at the very end of Part 1 or at the very beginning of Part 2, a section for FOL-with-one-binary-predicate, containing wcel, wel, ax-8, ax-9, df-ral, df-rex, df-reu, df-rmo. See also my section comment for "21.29.5.2  Classes without extensionality".  Note that I also made some pretty dramatic axiom dependency reductions in bj-ralcom4, bj-rexcom4.  What do @nmegill and @digama0 think about it ?

@wlammen : I see that you recently reverted several proof shortenings you made, because you had accidentally introduced new axiom dependencies (am I correct?).  I cannot help but fear that some other proof shortenings did add some dependencies but were not caught?  I hope you check them carefully ?  (Once again: this is not a criticism; I am very appreciative of your work.)

wlammen:
@benjub 
> I cannot help but fear that some other proof shortenings did add some dependencies but were not caught? I hope you check them carefully ?

I do my very best, in particular in the restricted quantifier section, where I can spell out rules when certain axioms are needed.  I am less confident about shortenings in the FOL section I did some years ago, when I was less aware of this restriction. I do not expect more than a couple of failures, though.

This points out the need of an automatic check during verification.

nmegill:
> Regarding restricted quantifiers: I agree with @wlammen that they can be considered part of FOL-with-one-binary-predicate ("the first order theory of one binary predicate with no axiom") and be placed earlier in set.mm.

The issue is that these use class variables, which are not part of FOL, so I don't consider them "pure" FOL theorems.  In particular, they require the syntax axiom wcel via df-ral and df-rex.

It is true that we introduce class variables artificially in FOL so as not to overload the e. and = predicates and have an unambiguous grammar for mmj2 and other tools.  This is done as a practical workaround.

But we try to avoid the use of wceq, wcel, and cv in the FOL section so that if someone wanted to have a "pure" FOL for another purpose, all they would have to do is remove wceq, wcel, and cv, and restate weq, wel as $a's.  If this is done, nrex etc. could not be expressed, and in that sense I would consider nrex as not part of FOL.

We could define setvar-only versions of df-ral/rex and have the FOL version of nrex use "x e. y" instead of "x e. A", and the essential steps in the proof would be the same, so in theory there could be some theoretical FOL benefit for the nrex proof using fewer axioms.  It is a stretch though, and I don't think it is something we need to be obsessed with.

BTW the avoidance of wceq, wcel, cv can't be controlled by 'discouraged' tags, so in principle it could be and sometimes is violated.  But avoidance usually happens more or less automatically, is easy to check, and 'minimize' will remove any violations, so we don't make a big deal about it.  I consider any violation a very minor cosmetic issue that is easy to fix for anyone concerned about it and in any case will get fixed by the next 'minimize' run.

(On second thought, there is a way to 'discourage' them:  we add 'usage discouraged' to wceq, wcel, cv, then at the beginning of set theory re-introduce them as theorems (aliases) that aren't discouraged.)

> I think there should be, either at the very end of Part 1 or at the very beginning of Part 2, a section for FOL-with-one-binary-predicate, containing wcel, wel, ax-8, ax-9, df-ral, df-rex, df-reu, df-rmo. See also my section comment for "21.29.5.2 Classes without extensionality". Note that I also made some pretty dramatic axiom dependency reductions in bj-ralcom4, bj-rexcom4. What do @nmegill and @digama0 think about it ?

They are nice.  I have no objection to making them the official ralcom4, rexcom4 in the main set.mm.

nmegill:
I had to rush off to an appointment and answered only the rexcom4/ralcom4 question, sorry.  Hopefully I've reread your proposal more carefully now....

> Regarding restricted quantifiers: I agree with @wlammen that they can be considered part of FOL-with-one-binary-predicate ("the first order theory of one binary predicate with no axiom") and be placed earlier in set.mm.

I don't think it belongs in the FOL section for the reason I gave (FOL doesn't have class variables).  I'm not even comfortable moving it before ax-ext and the class definitions.  We are already overwhelming the reader with df-clab,cleq,clel after the introduction of ax-ext.  Moreover, df-cleq doesn't hold before ax-ext (unless we drag around its hypothesis), and at that point the reader has no understanding of what class variables are even about.  I think from the reader's point of view, we would be introducing df-ral,rex too early even though it's technically possible, since the reader has no idea what a class variable is yet.

df-ral,rex are just simple "convenience" definitions that directly abbreviate a common pattern.  To me, the fact that a few early theorems using them don't need ax-ext is just a coincidence that doesn't have a deeper meaning.  There are likely many set theory definitions where we can use just prop calc and maybe some FOL to show alternate versions or simple consequences, but that doesn't mean we should consider them to be "FOL theorems" and move the definitions and those early theorems into the FOL section.

> I think there should be, either at the very end of Part 1 or at the very beginning of Part 2, a section for FOL-with-one-binary-predicate, containing wcel, wel, ax-8, ax-9, df-ral, df-rex, df-reu, df-rmo. See also my section comment for "21.29.5.2 Classes without extensionality".

I think it is better to introduce ax-ext first (gently introducing an axiom in the pure language of FOL with no class variables), then introduce df-clab,cleq,clel together like we are doing now.  I think these 3 need to be understood as an inseparable set of definitions that play together, for class theory to be understood by the reader.  All books using class theory, that I can recall, introduce these together. The algorithm to convert a class expression back to the language of pure FOL requires all 3 definitions.  Breaking out pieces like placing df-clel before ax-ext (if I understand you correctly) seems ugly and probably confusing.  Indeed, introducing any theorems or definitions at all that have class variables (including df-ral,rex) before ax-ext and df-cl* seems like the wrong thing to do.  It is a rather drastic change in the presentation philosophy to make, just because some early theorems with df-ral,rex don't need ax-ext.  So I'm not in favor of such a reorganization.

df-ral,rex are deceptively simple in terms of what they really represent when converted to the actual language of FOL  In particular, the "A" would become an expression involving a wff variable.

nmegill:
> df-ral,rex are deceptively simple in terms of what they really represent when converted to the actual language of FOL In particular, the "A" would become an expression involving a wff variable.

It might be useful to look at an FOL scheme corresponding to ~ nrex, in order to see what ~ nrex is really saying in FOL language:

```
|- ( x e. A -> -. ps ) => |- -. E. x e. A ps
      ~ nrex
|- ( x e. A -> -. ps ) => |- -. E. x ( x e. A /\ ps )
      Expand definition ~ df-rex
|- ( x e. { x | ph } -> -. ps ) => |- -. E. x ( x e. { x | ph } /\ ps )
      Substitute "{ x | ph }" for "A", where "ph" is a fresh wff variable
|- ( ph -> -. ps ) => |- -. E. x ( ph /\ ps )
      Use ~ abid
```

The corresponding FOL scheme is thus "|- ( ph -> -. ps )" => "|- -. E. x ( ph /\\ ps )".  From it we can recover the original ~ nrex:

```
|- ( ph -> -. ps ) => |- -. E. x ( ph /\ ps )
      Given
|- ( x e. A -> -. ps ) => |- -. E. x ( x e. A /\ ps )
      Substitute "x e. A" for "ph", where "A" is a fresh class variable
|- ( x e. A -> -. ps ) => |- -. E. x e. A ps
      Introduce definition ~ df-rex
```

Thus we end up with ~ nrex (possibly mod variable names), showing the equivalence of the translation to the FOL scheme.  In general we may need FOL+ax-ext+df-clab,cleq,clel for such a proof, although only FOL+df-clab are needed in this example.

PR #1411, opened 2020-01-14
section "The Friendship Theorem" moved from AV's mathbox to main set.mm

avekens:
Main body of set.mm - see issue #1368:
* section "The Friendship Theorem" (except subsection "Huneke's Proof of the Friendship Theorem") moved from AV's mathbox to main set.mm


david-a-wheeler:
Excellent! We now have more useful material in the main body, and I think that's great. Thank you so much!

Issue #1412, opened 2020-01-14
Label change proposals


benjub:
I propose the three label changes:
> cabel --> cabl
> pgrple2abel --> pgrple2abl
> pgrpgt2nabel --> pgrpgt2nabl

(these last two are in @avekens's mathbox).

Rational: all other theorems about abelian groups use "abl" in their labels (beginning with df-abl).  In addition, "abel" is often used in theorems related to "ELements of class ABstractions".

Then, there is Abel's theorem abelth and its lemmas. I think they can stay as is.

If @nmegill and @avekens give me the green light, I will do these three changes and document only the first one at the top of set.mm, and then close the issue.

nmegill:
I agree.

avekens:
I also agree. I will change the three labels in my next update.

avekens:
With PR #1415, the labelspgrple2abel and pgrpgt2nabel were changed.
Because cabel is used very often (219 times), and I had severe merge conflicts anyway, I did not change cabel yet. Maybe it would be better if a GitHub-Admin will do it (@nmegill ?).

benjub:
Thanks @avekens .  @nmegill : is it better that you do the change cabel --> cabl as @avekens suggest, for fear of merge conflicts, or should I do it ?

nmegill:
@benjub - OK, I will do this soon.

PR #1415, opened 2020-01-15
subsection "Huneke's Proof of the Friendship Theorem") moved from AV's mathbox to main set.mm

avekens:
Main body of set.mm:
* subsection "Huneke's Proof of the Friendship Theorem") moved from AV's mathbox to main set.mm. By this, #1368 is done! (Unfortunately, this subsection is intermingled with part "GUIDES AND MISCELLANEA" in the difference view below)
* ccats1rev renamed to reuccats1lem

AV's mathbox:
* pgrple2abel, pgrple2nabel renamed (see #1412)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ contributing changes to set.mm are described in the wiki,
 (Feel free to enhance the wiki.)
 
 The rest of this note shows how to achieve the recommended
-formatting for set.mm prior to submitting your changes.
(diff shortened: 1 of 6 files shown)


avekens:
I had to reconstruct some parts of "GUIDES AND MISCELLANEA" while I was solving merge conflicts. I compared this part with set.mm of the main branch, and it should be OK now. The only difference is in bullet point "Hypotheses", which I enhanced intentionally!

PR #1419, opened 2020-01-18
reduce dependencies on axioms in 19.36v

wlammen:


benjub:
@wlammen : very nice !  Maybe move 19.36v and 19.36aiv earlier in set.mm,
somewhere strictly between 19.9v and ax-7 ?
Also, relabel 19.36aiv to 19.36iv (or 19.36vi) ?  @nmegill ?  (I do not
understand the "a" in the label.)

benjub:
Same with 19.23v and 19.23vv ? Also 19.21v, which should be moved before
weq.
I noticed these by searching for "depend" in
http://us2.metamath.org/mpeuni/mmrecent1000.html
<http://us2.metamath.org/mpeuni/mmrecent1000.html#thlist> but this does not
go very far in the past.  Maybe there are other cases. (That is: respective
to $a-statements, a $p-statement should be moved as early as possible.)

nmegill:
> Also, relabel 19.36aiv to 19.36iv (or 19.36vi) ? @nmegill ? (I do not understand the "a" in the label.)

A long time ago the convention I used for the inference versions of biconditional 19.\* theorems was to use "a" for the forward inference and "b" for the reverse.  E.g. 19.21 had inferences 19.21ai (now renamed alrimi) and 19.21bi.

In prop calc we sometimes have used no suffix for the forward inference and "r" for reverse (pm5.74i and pm5.74ri) although that inconsistently has also meant swapping an internal conjunction (pm4.71i and pm4.71ri).

The "19.\*" (and "pm\*") labels aren't ideal since they are obscure, but for many of them I haven't thought of a better name and welcome suggestions.  To see the renames done so far, do `grep ' 19\.' set.mm | more`.

As for \*vi vs. \*iv, in most cases we've put the v for the distinct variable version after the i (\*iv, \*dv).

For the specific case of 19.36aiv, the need for a "b" version seems to have never arisen.  If you think we should take out the "a", I won't object.

Another inconsistent convention is using "vv" (19.23vv) and "bb" (19.21bbi) for double quantifier versions, I guess because there was no "v" to use in the 2nd case.  Again I've not been able to come up with a good consistent naming scheme for these.  When they've been renamed sometimes "2" was added in front of "al" (i.e. 2 foralls), like 2alimi (which used to be 19.20i2 - yet another inconsistent convention).  There is also 19.21-2.

wlammen:
@nmegill @benjub Thanks for the kudos.  I am not sure about the effects on usability if you keep *v theorems separate from their * counterpart.  Looking at 'nearby theorems' would not point to relevant variants any longer.  I use it a lot.  On the other hand, a separate development of dv aware theorems hints towards how FOL would develop if sp was an axiom.  spv is provable without ax-12.

This week I will be pretty busy at work, so do not expect too much input from my side.  Feel free to reorder/rename theorems at your discretion.  It won't interfere with my contributions.

PR #1423, opened 2020-01-19
Rename 'sqr' to 'sqrt'

david-a-wheeler:


david-a-wheeler:
The rename seems to have worked out, but I may need to be more aggressive about rewrapping lines.

david-a-wheeler:
Weird. Even after running scripts/rewrap there are many lines with more than 79 characters, e.g.:
~~~~
?Warning: Line number 156341 has 81 characters (should have 79 or less):
          sqrtmulii $p |- ( sqrt ` ( A x. B ) ) = ( ( sqrt ` A ) x. ( sqrt `...
?Warning: Line number 174854 has 81 characters (should have 79 or less):
        pythagtriplem11.1 $e |- M = ( ( ( sqrt ` ( C + B ) ) + ( sqrt ` ( C ...
?Warning: Line number 174916 has 81 characters (should have 79 or less):
        pythagtriplem13.1 $e |- N = ( ( ( sqrt ` ( C + B ) ) - ( sqrt ` ( C ...
?Warning: Line number 174987 has 81 characters (should have 79 or less):
        pythagtriplem15.1 $e |- M = ( ( ( sqrt ` ( C + B ) ) + ( sqrt ` ( C ...
?Warning: Line number 174989 has 81 characters (should have 79 or less):
        pythagtriplem15.2 $e |- N = ( ( ( sqrt ` ( C + B ) ) - ( sqrt ` ( C ...
?Warning: Line number 177034 has 80 characters (should have 79 or less):
            prmreclem3 $p |- ( ph -> ( # ` M ) <_ ( ( 2 ^ K ) x. ( sqrt ` N ...
?Warning: Line number 306261 has 80 characters (should have 79 or less):
       ( norm ` w ) = ( x e. ( Base ` w ) |-> ( sqrt ` ( x ( .i ` w ) x ) ) ...
~~~~

benjub:
Thanks for the rename.  I think that indeed the rewrapping does not rewrap within a $p ... $= and similarly with $e and $a.

When possible (e.g., in all but the last cases you quoted above), I would break the line after |- and right-align the second line.

david-a-wheeler:
> I would break the line after |- and right-align the second line.

Unfortunately I only saw that comment after I completed hand-fixing it.

In any case, I hand-wrapped everything so we stay within the column limit.
If people don't like it, the wrappings can be changed later as a separate step,
but I want to do the renames "all at once" to minimize the impact on others.

nmegill:
'/rewrap' doesn't modify the math content of statements.  It used to, but there were complaints that it chose bad line breaks, so I took it out to let humans choose the breaks they prefer.

Breaking after "|-" hasn't been our convention.  Usually we try to break before an outermost connective e.g. the too-long
```
sqrtmulii $p |- ( sqrt ` ( A x. B ) ) = ( ( sqrt ` A ) x. ( sqrt ` B ) )
```
would become
```
sqrtmulii $p |- ( sqrt ` ( A x. B ) ) 
                  = ( ( sqrt ` A ) x. ( sqrt ` B ) )
```
This convention is used almost universally in set.mm except in some equations in some mathboxes.  While breaking after "|-" allows the special case of equations with between ~60 to 78 characters to appear on one line with a blank line above it, for other equations it will end up using up more lines because we are wasting the blank line after "|-":
```
longeqn $p |- [20 char formula] = [70 char formula]
```
will require 3 lines if we break after "|-" but will require only 2 lines if we break before "=".
While an extra line doesn't impact file space much, it does reduce the amount that you can see in an editor window.  So we should probably discuss this before adopting a rule of always breaking after "|-".  If we make an exception just for the special case of ~60 to 78 character lines, it is slightly complicating the simple rule of "try to break before an outermost connective".

benjub:
Agreed for minimizig the number of lines. But isn't |- the outermost
connective? It is the root symbol.

nmegill:
It also occurred to me that
```
sqrtmulii $p |- ( sqrt ` ( A x. B ) ) 
                  = ( ( sqrt ` A ) x. ( sqrt ` B ) )
``` 
can be argued to convey more visual information to the reader at a glance than
```
sqrtmulii $p |-
           ( sqrt ` ( A x. B ) ) = ( ( sqrt ` A ) x. ( sqrt ` B ) )
```
because the former shows explicitly where the outermost connective is.  For very short equations it doesn't matter, so we put the whole equation on the same line as the label, but as the equation grows longer it becomes harder to parse it visually.  The threshold for that isn't well-defined, of course, but it seems to me it might as well be the point where we start to need more than one line.

nmegill:
> Agreed for minimizig the number of lines. But isn't |- the outermost connective? It is the root symbol.

If you want to be technical, yes, but to most people the "=" in an equation is what breaks the equation into 2 pieces.  By "outermost" I mean whatever provides the best human presentation, not in a strict technical sense.

nmegill:
I saved the diffs from David's commit and will review them for better line breaks when I get a chance.

BTW '/rewrap' doesn't like line breaks in the middle of math strings in comments (whether we should allow them is another discussion - the original intent was to make it easier for quick one-off scripts to parse comment markup), and it reassembles a broken math string back into one line.  If the resulting line is >79 chars, it will push the beginning of the line to column 1.  If it is still too long, the line will fail 'verify markup' and the user will have to adjust it by hand.  Fortunately in the case of David's edits, it just barely made it by pushing some of these to column 1.  I saved the diffs and will fix those later so they look a little nicer.  But as a general rule, a math string in a comment should be surrounded by backquotes on the same line, and if it is too long it should be broken into 2 mathstrings e.g.
```
$( ... 
    ` [long math string] ` 
    ` [continuation of long math string] `
   ... $)
```
This also gives the user a little bit more control over where they want to see long math expressions broken up, so that the breaks will survive in the source file after /rewrap.

<hr>

Going back to "outermost connective" - maybe I should have called it "the important connective" or something - another subtlety is when the whole expression has an antecedent.  E.g. in "|- ( A e. V -> [20-char lhs] = [60-char rhs] )", to a human the important part of the expression might be the consequent, with "=" as the principal symbol connecting its parts.  We could break this into
```
xyz $p |- ( A e. V
        -> [20-char lhs]
             = [60-char rhs] ) $=
```
but I think it would be acceptable to use
```
xyz $p |- ( A e. V -> [20-char lhs]
                = [60-char rhs] ) $=
```
I believe we do that a lot in set.mm e.g. ~ absmax.  Maybe it's not the prettiest presentation, but I don't think it's confusing once you get used to it, and it saves a lot of extra lines overall.

Admittedly there are also many that don't follow any pattern such as ~ binom21, and I would like to see them improved, but that is a project for another day.

There are also many early theorems that break after instead of before the "important" connective, such as ~ binom2, and I would also like to have those fixed someday.  Originally breaking after the connective was the convention, until a decade or so ago someone convinced me that books almost always break before, such as in a long chain of equalities.

nmegill:
I updated the sqrt line breaks to my satisfaction, at least, in commit fb4b0bf .  (I updated it directly; couldn't create a PR because of disk space problems creating a fork.)

@david-a-wheeler FYI I noticed you adjusted some long URLs by putting a line break after the \~, I think in an attempt to align it to the indentation level.  This was undone by /rewrap, which doesn't like a line break between a ~ and the label/URL.  It took out the line break, then moved the long "~ http..." back to column 1 so it would fit in 79 columns.  This is by design as a workaround to allow longer URLs.

david-a-wheeler:
Thanks!

Is there something here worth documenting in the conventions?

> rewrap' doesn't modify the math content of statements. It used to, but there were complaints that it chose bad line breaks, so I took it out to let humans choose the breaks they prefer.

It would be nice to have wrapping statements an option in metamath.exe. Hint hint hint. When you have to do a large number of them, the risk of making a stupid mistake goes up.

nmegill:
> Is there something here worth documenting in the conventions?
> 
> > rewrap' doesn't modify the math content of statements. It used to, but there were complaints that it chose bad line breaks, so I took it out to let humans choose the breaks they prefer.
> 

It is documented in 'help write source'.  I'm not sure if adding more to the conventions is really going to help - the more people have to read the less they will remember.  As you've seen, you will quickly find out about it with error messages in 'verify markup *' - which BTW you apparently didn't do before submission. :)

> It would be nice to have wrapping statements an option in metamath.exe. Hint hint hint. When you have to do a large number of them, the risk of making a stupid mistake goes up.

This seems like a dangerous thing to do.  The current method of not wrapping forces you to look at each case and hopefully handle it in a reasonable way.  Note that all of the "stupid mistakes" were caught by "verify...." commands so no permanent harm was done.

It isn't just your work affected by the label change.  If you had used such an option (which it sounds like you might have in this case), I think people would dislike even more to see their carefully formatted wrapping destroyed, possibly hidden and going unnoticed for years.

david-a-wheeler:
> I'm not sure if adding more to the conventions is really going to help - the more people have to read the less they will remember.

That's an issue, but the best way to address that is to organize things so that they see exactly what they need when they need it.

The chance of them seeing or finding something undocumented is 0% :-).

> As you've seen, you will quickly find out about it with error messages in 'verify markup *' - which BTW you apparently didn't do before submission. :)

I did verify things. I saw the rewrap warnings, and I then ran scripts/rewrap. I didn't reverify after that because I foolishly trusted metamath to rewrap the text correctly :-) :-). And I'm willing to directly trust the CI tool; it tests more than I can easily test locally, I think it's unlikely you'll merge anything that doesn't pass the tests.

That said, all's well that ends well, and things were learned.

david-a-wheeler:
> It isn't just your work affected by the label change. If you had used such an option (which it sounds like you might have in this case), I think people would dislike even more to see their carefully formatted wrapping destroyed, possibly hidden and going unnoticed for years.

Maybe there could be an option to force rewrapping for *only* constructs that exceeded the
line length? That wouldn't necessarily make things "pretty" of course, but on mass restructures
that may be too much to ask.

nmegill:
> I didn't reverify after that because I foolishly trusted metamath to rewrap the text correctly :-)

/rewrap wraps only content without an \<HTML\> tag in a $p/$a description and nowhere else.  So there are many places and situations where lines could become too long.  (I plan to rewrap the text below section headers in a future release, but right now it isn't wrapped.)  From 'help write source':
```
        SET WIDTH parameter currently in effect.  Note 1:  A comment
        containing an <HTML> tag is not rewrapped.  Note 2:  Comments not
        immediately preceding a $a or $p statement are not rewrapped.
        Note 3:  Math symbol strings are not rewrapped.
```

> Maybe there could be an option to force rewrapping for _only_ constructs that exceeded the
> line length?

That is how is used to work.  There were still complaints.

> on mass restructures that may be too much to ask.

Changes have consequences when the label or token length increases.  There is no use case I can think of where we want to allow people to blindly wrap formulas, if we value keeping the set.mm content neat and well-structured.  Such a feature would almost certainly be abused, and the formula structure in set.mm would slowly become worse.  I think we _must_ force people to look at formulas needing wrapping.  An error message from 'verify markup' is the only way I can think of to enforce it.

What I could do, perhaps, is to have /rewrap issue a warning when there are still long lines that need to be identified with 'verify markup' and adjusted manually.  I'll add that to my todo list.

If I counted correctly, there were 36 $a, $p, and $e formulas that were too long.  That isn't a big number given the huge size and scope of the sqrt mod.  Forcing these formulas to be adjusted by hand seems a small price to pay if we want to keep the formulas structured.  Indeed, for the ones I readjusted, I saw it as an opportunity to improve the original formatting in some cases.

Maybe in the future an algorithm can figure out how to optimally reformat formulas in the same way as we do now by hand, ideally having meaningful line breaks but knowing where to compromise in order to keep line count to a minimum.  But adding that to metamath.exe is beyond the scope of what I currently have planned.

david-a-wheeler:
> If I counted correctly, there were 36 $a, $p, and $e formulas that were too long. That isn't a big number given the huge size and scope of the sqrt mod. Forcing these formulas to be adjusted by hand seems a small price to pay if we want to keep the formulas structured. Indeed, for the ones I readjusted, I saw it as an opportunity to improve the original formatting in some cases.

Fair enough. I think part of the issue is that I was surprised I had to do the rewrap by hand at all. I think I had used metmath's rewrap back when it had the older behavior, and I hadn't noticed that the behavior changed.

In any case, I now know about it,  and the conventions section has better info and pointers to even more info.

PR #1425, opened 2020-01-20
Add section "multiplication" to iset.mm

jkingdon:
That is, from kcnktkm1cn through mulsubfacd .

We start with mul02 and its consequences - which are defined earlier in set.mm but which we're ready to tackle now.

There's a little intuitionizing in ine0 .

Other than those two topics, the whole section is unmodified from set.mm.


david-a-wheeler:
Wow. You appear to have crossed some the hardest problems in getting started and you're now up to the math recognized by kids everywhere. Congratulations!

jkingdon:
> Wow. You appear to have crossed some the hardest problems in getting started and you're now up to the math recognized by kids everywhere. Congratulations!

Thanks for the kind words. Always good to keep in mind all the progress when it isn't clear how to make things work (the struggle at the moment is how to prove "a ≤ b and b < c implies a < c"), or the various other tough bits to come.

Issue #1418, opened 2020-01-17
Revision of Graph Theory


avekens:
While moving the friendship theorem into the main part of set.mm (see #1389), I noticed some intricatenesses which bother me a lot:

1.  Edges:

To be compatible with the definition of multigraphs, simple graphs were defined as pairs of a set (vertices) and an edge function (or indexed edges) instead of edges (as pairs of vertices), which would have been sufficient for simple graphs. To consider the "edges" always as functions makes many theorems clumsy and their proofs longer.  Therefore, I introduced a function **`Edges`** which provides the edges as a set of pairs of vertices for a simple graph (already available in the main body of set.mm):

`( Edges `` <. V , E >. ) = ran E` or `( Edges `` G ) = ran ( 2nd `` G )`
[Would **`Edge`** be a better symbol instead of `Edges`? We also have, for example, the symbol `Ring` to denote the class of all (unital) ring**s**, and not `Rings` . Another alternative would be to use **`Edg`**.]

This is the "usual" way to talk about edges (of simple graphs), see [Bollobas] p.1: "... such that E is a subset of the set V^2 of unordered pairs of V." Using this definition, many theorems can be expressed in a better (that is shorter and more concise and comprehensible) way, e.g.

usgvincvadeu: `( ( G e. USGrph /\ C e. ( Edges `` G ) /\ X e. C ) -> E! y e. V C = { X , y } )`

instead of

usgraedgreu: `( ( V USGrph E /\ X e. dom E /\ Y e. ( E `` X ) ) -> E! y e. V ( E `` X ) = { Y , y } )`

Therefore, I want to propose to use `Edges` in more theorems in the context of simple graphs, mainly replacing `dom E` and `ran E` in the currently available theorems, making those obsolete. I started this endeavor already in my mathbox, see subsection "Edges of undirected simple graphs without loops". 
 
2. A graph represented by a class variable `G`: 

We started talking about graphs regarding them always as pairs `<. V , E >.` Instead of writing 
`<. V , E >. e. USGrph`, we used to use the representation as binary relation `V USGrph E`, which is shorter (saves 7 characters), but does not reflect the original meaning "A graph is a pair of vertices and edges" (instead, "V USGrph E" would literally stand for "The vertices and edges are related by a graph", which is not wrong, but not the way how graphs are seen).

[...]

Issue #1389, opened 2020-01-07
Ring vs. Rng (deferred)


avekens:
According to the Google group discussion https://groups.google.com/d/msg/metamath/XKl7K0qx4uc/hRykryMhDgAJ (see also issue #866), the currently used abbreviation "rng" for a unital ring `Ring` should be replaced by "ring".

As a first step, I provided the definition of a general, i.e. not necessarily unital , ring `Rng` (see ~df-rng0 in my mathbox). For this, I needed the definition of a "Semigroup" (the mutiplicative group of a general ring is a semigroup), so I added this, too (`SGrp`, see ~df-sgrp). Doing this, I also added a definition of a "Magma" (`Mgm`, see ~df-mgmALT). Here are conflicts with the deprecated definition of a magma which have to be cleaned up. For all changes see PR # 1388.

If there is no objection, I would move these definitions and the related theorems into the main part of set.mm in the next step, changing the labels: df-rng -> df-ring, df-rng0 -> df-rng (and df-mgm -> df-magma, df-mgmALT -> df-mgm, etc.). I would also update the table of abbreviations in section "17.1.1  Conventions" accordingly. Maybe some of the current "rng" theorems are already valid for Rngs, so that they can be moved up into the corresponding subsection (and need not to be renamed).

Afterwards, the renaming of the labels containing "rng" could start.

@benjub @nmegill @digama0 @david-a-wheeler @tirix OK?

benjub:
Sounds good to me. Thanks @avekens

avekens:
After I bought Bourbaki ;-) I decided to go 2 steps back beyond the definition of magmas and start the definitions and theorems for algebraic structures with "laws" and "internal binary operations", according to Bourbaki, but also to Marshall Hall Jr. in "The Theory of Groups". This seems to be also the approach in the deprecated section "18.1.5 Group-like structures", which I will clean up on moving my definitions to the main body of set.mm.

I defined 3 laws:
- clLaw
- assLaw (corresponds to deprecated `Ass`)
- comLaw

and 3 kinds of internal binary operations:
- intOp
- clIntOp (corresponds to deprecated `Magma`)
- assIntOp (corresponds to deprecated `SemiGrp`)

The laws can be used to specify the operation slots of extensible structures, resulting in slightly modified definitions for magmas and semigroups:

-    df-mgm2 $a |- MgmALT2 = { m | ( +g ` m ) clLaw ( Base ` m ) } $.
[...]

PR #1428, opened 2020-01-20
Graph theory; basic algebraic structures

avekens:
main set.mm (see also issue #1418):
* bibliographic reference added to header comment of part "GRAPH THEORY"
* new entry "Undirected simple hypergraph", revision of entry  "Undirected hypergraph" in glossary.
* new definition for USHGrph and related theorems
* htmldef/althtmldef/latexdef for new symbols added

AV's mathbox:
* Showing the equivalence of definitions for undirected hypergraphs
* new section "Internal binary operations" - this should make many of the definitions/theorems in section "18.1 Additional material on group theory" obsolete (for the moment, the "discouraged" counter was increased for som of these theorems) . See also issue #1389

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4069,6 +4069,7 @@
 "chub2i" is used by "qlaxr3i".
 "chub2i" is used by "sumdmdlem2".
 "clmgm" is used by "exidcl".
+"clmgm" is used by "mgmcllaw".
 "cm0" is used by "chirred".
 "cm2j" is used by "cm2ji".
 "cm2ji" is used by "cm2mi".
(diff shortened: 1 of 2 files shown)


nmegill:
The section numbers change when new earlier sections are added.  So in the comment 'In subsection "18.1.5 Group-like structures",...', the 18.1.5 will not be permanent.  It might be better to say, 'In the subsection "Group-like structures" beginning at ~  cmagm ,...'.

Unless someone has a better suggestion.

While not as convenient as "18.1.5", at least the reader can search for "cmagm" on the TOC contents page.  Alternately the reader can click the cmagm hyperlink then click "Nearby theorems" to see the section heading (after scrolling up a couple of lines).

nmegill:
BTW Part 18 of the TOC is deprecated.  I think you mean Part 10.

avekens:
No, I actually mean Part 18. According to the Google group discussion (see https://groups.google.com/d/msg/metamath/OdIB-vF-Qo0/kKnAOFe6AQAJ), I think I can remove a lot of this part after I have updated Part 10 accordingly. Therefore, the references to Part 18 in my comments  are only temporary and will be removed soon. See also issue #1389.

PR #1433, opened 2020-01-23
Magmas and Rings

avekens:
Main set.mm:
- proposal to rename ~proplem, ~proplem2, ~proplem3 added in header section "1. Recent label changes"
- ~sqxpeqd and ~sqxpexg moved from AV's mathbox to main set.mm
- new theorem ~bropaex12
- new connective symbol .o. (circle)

Cleanup "Part 18" (see #1423):
- new theorem ~mndpfo in main set.mm (variant of ~mndfo)
- ~mnd1, ~mnd1id, ~grp1, ~abl1, ~abln0 moved from AV's mathbox to main set.mm
- ~rng1ne0, ~rng1, ~rngn0, ~isnzr2hash, ~0rngnnzr, ~rng1nnzr and ~rng1nfld moved from AV's mathbox to main set.mm
- ~opidon, ~rngopid, ~opidon2, ~ablosn, ~gidsn, ~cnaddablo in Part 18 marked as obsolete

AV's Mathbox:
- section "Monoids (extension)" renamed to "Magmas and semigroups (1. approach)"
- note regarding the term "groupoid" added (according to remarks/proposals of Benoit and Gerard)
- subsection "References to subsection "Group-like structures"" added
- section "Internal binary operations" renamed to "Magmas and internal binary operations (2. approach)"
- subsections introduced
- revision of comments, new theorems (inspired by theorems of Part 18)

### DIFF
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -5070,7 +5070,6 @@ Sol&egrave;r,&quot; <I>Bull.  Am.  Math.  Soc.</I> 32:205-234 (1995)
 [QA.A513]; available at <A HREF="http://arxiv.org/abs/math/9504224v1">
 http://arxiv.org/abs/math/9504224v1</A> (retrieved 11 Nov 2014).</LI>
 
-
 <LI><A NAME="Huneke"></A> [Huneke] Huneke, Craig L.,
 &quot;The Friendship Theorem&quot; <I>American Mathematical Monthly</I>
 109:192-194 (2002); available at <A


nmegill:
I am fine with your proplem, proplem2, proplem3 proposal.

set.mm
```diff
@@ -74469,6 +74501,22 @@ Finite induction (for finite ordinals)
...
+  $}
+
+  ${
+    sqxpeqd.1 $e |- ( ph -> A = B ) $.
+    $( Equality deduction for a Cartesian square.  (Contributed by AV,
+       13-Jan-2020.) $)
+    sqxpeqd $p |- ( ph -> ( A X. A ) = ( B X. B ) ) $=
+      ( xpeq12d ) ABCBCDDE $.
+  $}
+
```
tirix: Isn't basically ~ sqxpeqd an alias for ~ xpeq12d ?

avekens: In principle yes, but I had 3 reasons to add this theorem:
- it is shorter to use typing a proof (at least with metamath.exe): 2 steps instead of 3
- the proofs will become a little bit shorter: see, for example ~isfin6: minimize will replace ~xpeq12d by ~sqxpeqd, decreasing its proof from 266 to 261 bytes and from 189 to 181 steps (1 essential step).
- I wanted to popularize the term "Cartesian square" (instead of "square Cartesian product"), see also ~sqxpexg.

benjub:
>
>
>    - I wanted to popularize the term "Cartesian square" (instead of
>    "square Cartesian product"), see also ~sqxpexg.
>
> This is in my opinion the most natural term; see my section "21.29.6.1
 Diagonal in a Cartesian square".
As for sqxpeqd, I think its presence is justified.

digama0:
`sqxpeqd` looks questionable to me. I would judge it based on pure proof shortening power. Losing 5 bytes in a proof is something but probably not enough to recoup the loss of the statement itself, unless "cartesian squares" are used a LOT more than I thought.

avekens:
> `sqxpeqd` looks questionable to me. I would judge it based on pure proof shortening power. Losing 5 bytes in a proof is something but probably not enough to recoup the loss of the statement itself, unless "cartesian squares" are used a LOT more than I thought.

After the recent global minimization, ~sqxpeqd is used in 121 proofs!

david-a-wheeler:
> ~sqxpeqd is used in 121 proofs!

That's remarkable, I would never have guessed it would be used that often. Very cool.

benjub:
That's probably because binary laws in algebraic structures and binary relations are defined on Cartesian squares?

PR #1437, opened 2020-01-25
Two changes to iset.mm definitions

jkingdon:
(quick summary: this pull request removes `TEST`, `+pR`, and `-pR` . cc @david-a-wheeler and @benjub ).

In light of the recent discussion of definitions, I looked at definitions in iset.mm (especially ones that do not also appear in set.mm).

Here's what I found:

- STAB . This is mostly unused in iset.mm, but since ` ph ` appears twice in the definiens, it might be awkward to do without it at some point in the future. My suggestion: keep it for now.

- TEST . This is easily replaced (as in, `TEST ph <-> DECID -. ph` ), and this pull request replaces it in just that way. For example, `stabtestimpdc` changes from ` ( ( STAB ph /\ TEST ph ) <-> DECID ph )` to
`( ( STAB ph /\ DECID -. ph ) <-> DECID ph )` (and the comment describes the left hand side as "ph is stable and testable" or something like that).

- DECID . This one does have one concrete application of being used to prove something on a topic other than decidability itself: http://us.metamath.org/ileuni/distrlem4prl.html . Suggestion: keep.

- `Q0.` and friends (non-negative fractions). These are used in the construction of real numbers (for the theorem `prarloc` ). Whether there is another way to prove this theorem, I don't know, but given that we are using this technique, it seems hard to see how we'd do it without the definitions.

- `+pR` and `.pR` . These were copied from an old version of set.mm but are no longer needed (in either set.mm or iset.mm) so this pull request removes them.

Upcoming:
- In order to state things like division and reciprocal theorems, we'll want apartness. Although for real number apartness, it might be OK to say `A < B \/ B < A` (or, depending on the context, just `A < B`), I think that will get awkward and most texts define a notation (usually `A # B`) for this. That's even more true for complex number apartness. Probably we'll define `#` to be complex number apartness (not sure we'll need a separate definition for reals).

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -58140,9 +58140,6 @@ $)
   $c .P. $. $( Positive real multiplication $)
   $c <P $. $( Positive real ordering relation $)
 
-  $c +pR $. $( Signed real pre-addition $)
-  $c .pR $. $( Signed real pre-multiplication $)
-
   $c ~R $. $( Equivalence relation to construct signed reals $)
   $c R. $. $( Set of signed reals $)
   $c 0R $. $( Signed real constant 0 $)
@@ -58229,10 +58226,6 @@ $)
   $( Extend class notation with new sets (constants) used in construction of
     signed real numbers: $)
 
-  $( Signed real pre-addition. $)
-  cplpr $a class +pR $.
-  $( Signed real pre-multiplication. $)
-  cmpr $a class .pR $.
   $( Equivalence class used to construct signed reals. $)
   cer $a class ~R $.
   $( Set of signed reals. $)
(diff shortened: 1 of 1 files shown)


benjub:
Thanks @jkingdon for the heads-up.  This is in line with my comment in PR https://github.com/metamath/set.mm/pull/1187#issuecomment-540692790
As I noted there, the important point is that "ph" appears only once in the definiens of TEST.  In this respect, STAB, where "ph" appears twice in the definiens, has *some* justification.

The apartness relation should in my opinion have a defined symbol as you propose, like A # B, since ( A < B \\/ B < A ) would become unreadable after A and B are substituted with long strings.

By the way: my ~dcnbidcnn is now exactly testbitestn, except that the proofs seem to differ (use of orbi1i versus orbi2i).  Can you pick the shortest proof and remove the other one ? (I'm away from my main computer). (Of course, credit is to @david-a-wheeler, not me.)

jkingdon:
> The apartness relation should in my opinion have a defined symbol as you propose, like A # B, since ( A < B \/ B < A ) would become unreadable after A and B are substituted with long strings.

Agreed. The only question is whether we'll also need a `#R` which we use temporarily for real number apartness until we can define the full complex number `#`. But that should become apparent once we get to apartness.

digama0:
In set.mm we've pretty universally used "maximally defined symbols" for these situations. + is defined on complexes and restricted to reals. < is defined on extended reals and restricted to reals. So as long as complex apartness specializes correctly to real apartness, I would be inclined to have just one apartness.

Given that not equal is written =/= in set.mm, how about using =//= for apartness? I think # is already used by the cardinality function. (This is only a thought; it's not clear if this is worth it if apartness is common enough.)

jkingdon:
> In set.mm we've pretty universally used "maximally defined symbols" for these situations. + is defined on complexes and restricted to reals. < is defined on extended reals and restricted to reals. So as long as complex apartness specializes correctly to real apartness, I would be inclined to have just one apartness.

Yup. In the process of defining `<`, we have a `<RR` but it isn't used once we have `<` and that's what I would see here. That is, `#R` would either not exist at all, or would only get used in a small section of iset.mm.

And yes, I'm assuming that complex apartness specializes correctly to real apartness (see section 6 of Skeleton for the Proof development leading to the. Fundamental Theorem of Algebra. Herman Geuvers, Randy Pollack, Freek Wiedijk, Jan Zwanenburg, October 2, 2000 for some definitions on complex apartness).

> Given that not equal is written =/= in set.mm, how about using =//= for apartness? I think # is already used by the cardinality function. (This is only a thought; it's not clear if this is worth it if apartness is common enough.)

Thanks for pointing out the conflict. The constructive literature as far as I've noticed has only two notations for apartness: `#` and `≠`. Picking the latter, of course, means that ≠ does not mean negated equality any more, which especially in our context makes me reluctant to go down that road.

For the ASCII tokens, `=//=` might not as nice as `#` but we could presumably cope. But the issue with how we typeset these is harder to resolve. Most of the possible solutions strike me as having some downside.

Issue #1436, opened 2020-01-24
iscatd2's description is self-referential 


Prosfilaes:
Theorem | iscatd2 14932* | Version of iscatd2 14932 with a uniform assumption list, for increased proof sharing capabilities.

I'm guessing this is supposed to say "Version of iscatd..."? Otherwise, the definition of what iscatd2 is supposed to be was lost somewhere along the line.

tirix:
You're right, thanks! I've fixed that in #1440 .

PR #1440, opened 2020-01-25
Paracompactness, image of a group by an homomorphism

tirix:


set.mm
```diff
@@ -64215,6 +64217,14 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
       HUJCKZABLZUFUGUMCAEBDNOULUOECUNCPHZABUNUJPHUPUIDQUJCPRSUAUHCKZUKUNABUQUKC
       UJKUNUHCUJUBCUJUCTUDUET $.
 
+    $( A member of a surjective function's ran is a value of the function.
```
benjub: To me, the range is the set of values of a function, so this comment seems a bit tautological. If F is a function from A to B, I would call B the codomain. Of course this is not definable in the current set.mm because of the way functions are defined. This does not forbid using the term "codomain" in comments.

set.mm
```diff
@@ -204730,6 +204740,118 @@ with the same (relevant) properties is also a group.  (Contributed by
...
+    ghmgrp.p $e |- .+ = ( +g ` G ) $.
+    ghmgrp.q $e |- .+^ = ( +g ` H ) $.
+    ghmgrp.1 $e |- ( ph -> F : X -onto-> Y ) $.
+    ${
+      mhmmnd.3 $e |- ( ph -> G e. Mnd ) $.
+      ${
+        $d .0. a d i x y $.
+        mhmid.0 $e |- .0. = ( 0g ` G ) $.
+        $( The neutral element of the image of a monoid by an homomorphism.
+           (Contributed by Thierry Arnoux, 25-Jan-2020.) $)
```
benjub: Do you mean: "A surjective monoid morphism preserves units."

set.mm
```diff
@@ -275107,6 +275229,741 @@ arbitrary neighborhoods (such as "locally compact", which is actually
...
+
+  $c LocFin $.
+  $( Extend class definition to include the class of locally finite covers. $)
+  clocfin $a class LocFin $.
+
+  ${
+    $d x y z $.
+    $( Define the fineness relation for covers.  (Contributed by Jeff Hankins,
+       28-Sep-2009.) $)
+    df-fne $a |- Fne = { <. x , y >. | ( U. x = U. y /\ A. z e. x z C_ U. ( y
```
benjub: Is this a standard notion ?

set.mm
```diff
@@ -275107,6 +275229,741 @@ arbitrary neighborhoods (such as "locally compact", which is actually
...
+
+  ${
+    $d a b x y A $.  $d a b x B $.  $d a b C $.  $d a b X $.  $d a b Y $.
+    isref.1 $e |- X = U. A $.
+    isref.2 $e |- Y = U. B $.
+    $( The property of being a refinement of a cover.  Dr.  Nyikos once
+       commented in class that the term "refinement" is actually misleading and
+       that people are inclined to confuse it with the notion defined in
+       ~ isfne .  On the other hand, the two concepts do seem to have a dual
+       relationship.  (Contributed by Jeff Hankins, 18-Jan-2010.) $)
```
benjub: I would only keep: "The property of being a refinement of a cover."

set.mm
nmegill: I agree, "surjective function's ran" should be changed to "surjective function's  codomain"

nmegill:
I'm fine with this generally and will merge it.  You can make the "ran" to "codomain" change and other changes you think are needed based on @benjub's comments, in your next PR.

set.mm
tirix: I'll keep your wording, but write "A surjective monoid morphism preserves _identity elements_.", since this is the way it is named elsewhere.

set.mm
tirix: Thanks! I've changed to "codomain".

tirix:
PS. This PR also fixes issue #1436

set.mm
tirix: Here I chose to move the whole section from Jeff Hankins' mathbox to the main part. You're right, I think while "refinement" is a standard notion, the "finer" relation for covers is not. What do others think? I could move this back into Jeff's mathbox.

benjub:
Yes @tirix , I would remove all theorems involving Fne (put them back in the mathbox).  It is strange that @nmegill let them go in the main part.

nmegill:
I am not too familiar with Fne but noticed that JH was the contributor.  He was usually pretty reliable (I think he was taking a topology course at the time) so that was the reason I didn't object.  They can be moved back.

PR #1442, opened 2020-01-26
Extensible structure versions of proofs from section 18

tirix:


set.mm
```diff
@@ -217908,6 +217908,46 @@ an extension of the previous (inserting an element and its inverse at
...
+    ghmabl.q $e |- .+^ = ( +g ` H ) $.
+    ghmabl.f $e |- ( ( ph /\ x e. X /\ y e. X )
+      -> ( F ` ( x .+ y ) ) = ( ( F ` x ) .+^ ( F ` y ) ) ) $.
+    ghmabl.1 $e |- ( ph -> F : X -onto-> Y ) $.
+    ${
+      ghmcmn.3 $e |- ( ph -> G e. CMnd ) $.
+      $( The image of a commutative monoid` G ` under a group homomorphism
+         ` F ` is a commutative monoid.  (Contributed by Thierry Arnoux,
+         26-Jan-2020.) $)
+      ghmcmn $p |- ( ph -> H e. CMnd ) $=
```
digama0: Is there a version of this for monoids/groups?

digama0: Also, it would be nice to conclude that `F` is also a group hom.

set.mm
```diff
@@ -341639,6 +341684,51 @@ mathematician Ptolemy (Claudius Ptolemaeus).  This particular version is
...
+    $d A x y z $.  $d F x y $.  $d G x y $.  $d X x y z $.  $d ph x y z $.
+    efabl.1 $e |- F = ( x e. X |-> ( exp ` ( A x. x ) ) ) $.
+    efabl.2 $e |- G = ( ( mulGrp ` CCfld ) |`s ran F ) $.
+    efabl.3 $e |- ( ph -> A e. CC ) $.
+    efabl.4 $e |- ( ph -> X e. ( SubGrp ` CCfld ) ) $.
+    $( The image of a subgroup of the group ` + ` , under the exponential
+       function of a scaled complex number, is an Abelian group.  (Contributed
+       by Paul Chapman, 25-Apr-2008.)  (Revised by Mario Carneiro,
+       12-May-2014.)  (Revised by Thierry Arnoux, 26-Jan-2020.) $)
+    efabl $p |- ( ph -> G e. Abel ) $=
```
digama0: This theorem is jumping the gun a bit. Before concluding that G is an abelian group, you want to know that `ran F` is a subgroup of the unit group of CCfld (or a submonoid of multiplicative CCfld).

set.mm
```diff
@@ -341639,6 +341684,51 @@ mathematician Ptolemy (Claudius Ptolemaeus).  This particular version is
...
+  $}
+
+  ${
+    $d x y C $.  $d y T $.
+    circgrp.1 $e |- C = ( `' abs " { 1 } ) $.
+    circgrp.2 $e |- T = ( ( mulGrp ` CCfld ) |`s C ) $.
+    $( The circle group ` T ` is an Abelian group.  (Contributed by Paul
+       Chapman, 25-Mar-2008.)  (Revised by Mario Carneiro, 13-May-2014.)
+       (Revised by Thierry Arnoux, 26-Jan-2020.) $)
+    circgrp $p |- T e. Abel $=
```
digama0: ditto here: the statement that C is a subgroup of the unit group of CC should be exported, this is just a corollary

set.mm
tirix: Yes, [~mhmmnd](http://us2.metamath.org/mpeuni/mhmmnd.html) is the version for monoids, and [~ghmgrp](http://us2.metamath.org/mpeuni/ghmgrp.html) for groups.

set.mm
tirix: Can do! I've added that proof.

set.mm
tirix: Done in the latest commit!

set.mm
tirix: The proof that `F` is a group homomorphism is in the [last commit](https://github.com/metamath/set.mm/pull/1442/commits/a06956372243979818c3a30cef91945ec9cb2bb6).

PR #1446, opened 2020-01-27
iset/mathbox minor corrections

benjub:
@jkingdon :
* In this PR, I changed the conclusion of bj-d0clsepcl from |- ( ph \\/ -. ph ) to |- DECID ph
It looks clearer, and since we have a definition, we could as well use it. Maybe do the same with acexmid and friends ?
* I added to my mathbox closed forms of notbid and dcbid.  I know deduction form is sufficient and easier to manipulate in longer proofs, but with fundamental operators like -. and DECID, having a closed form has the potential to minimize many proofs. Can I move them to main and do the minimizations ?

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -290,7 +290,8 @@ Proof modification of "bj-axun2" is discouraged (47 steps).
 Proof modification of "bj-bdfindes" is discouraged (121 steps).
 Proof modification of "bj-bdfindis" is discouraged (84 steps).
 Proof modification of "bj-bdfindisg" is discouraged (44 steps).
-Proof modification of "bj-d0clsepcl" is discouraged (125 steps).
+Proof modification of "bj-d0clsepcl" is discouraged (111 steps).
+Proof modification of "bj-dcbi" is discouraged (31 steps).
 Proof modification of "bj-ex" is discouraged (31 steps).
 Proof modification of "bj-exlimmp" is discouraged (34 steps).
 Proof modification of "bj-exlimmpi" is discouraged (11 steps).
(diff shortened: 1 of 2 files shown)


benjub:
Fixed top date... again... Having two different conventions for the main two databases is confusing.

jkingdon:
>     * I changed the conclusion of bj-d0clsepcl from |- ( ph \/ -. ph ) to |- DECID ph
>       It looks clearer, and since we have a definition, we could as well use it. Maybe do the same with acexmid and friends ?

I thought about this when I first introduced `acexmid` and the like. And I guess my reasoning was that it is clearer to say `ph \/ -. ph` than "ph is decidable" when this isn't as familiar a concept as many others. And `DECID ph` is worse yet, as this is not found in any works except iset.mm as far as I know. Now if people want to make the case for the other choice, we can consider it. But at least in the works I've been reading, I don't see "xxx is decidable" all that often.

>     * I added to my mathbox closed forms of notbid and dcbid.

Definitely should add bj-notbi because set.mm has http://us.metamath.org/mpeuni/notbi.html and I'll believe you that it would minimize multiple proofs. I guess we should name it `notbiim` (by analogy to `imanim`, `annimim`, `oranim` and probably other cases where set.mm has a biconditional and we just have one direction).

As for `bj-dcbi`, I don't object to having a closed form but that seems to be a bit thornier in terms of naming (as you already noticed). The thing about http://us.metamath.org/ileuni/dcbi.html is that we also have http://us.metamath.org/ileuni/dcim.html , http://us.metamath.org/ileuni/dcor.html , etc. I suppose these also might be better stated in imported rather than exported form (for example, `( ( DECID ph /\ DECID ps ) -> DECID ( ph \/ ps ) ) )` rather than `( DECID ph -> ( DECID ps -> DECID ( ph \/ ps ) ) )` although that's more of an aside, not something we'd necessarily consider at the same time we consider the naming conflict over `dcbi`.

jkingdon:
> Fixed top date... again... Having two different conventions for the main two databases is confusing.

Yeah, I too mix this up even though I know about both conventions. I haven't tried to raise a stink because.... well, Norm spends a lot more time doing things like merging so I'm willing to go with what he finds convenient.

nmegill:
> Fixed top date... again... Having two different conventions for the main two databases is confusing.

OK, let's add /top_date_skip to the 'verify markup' run so they will be the same.  @jkingdon I think you did a PR for this that wasn't merged, could you do it again?  Thanks.

PR #1448, opened 2020-01-29
Locally finite refinement of an open cover

tirix:
induces a locally finite cover with the original index set.


benjub:
Does this really prove what the comment claims ?  This may be the case, but all these encodings are getting very cryptic...

* an union --> a union (since the "u" is pronouced "you", sounding like a consonant)

* "The function associating each element of a refinement to its superset." : shouldn't it be "For any cover refinement, there exists a function associating with each set in the refinement a set in the original cover containing it (and conversely, the existence of such a function implies that one has a cover refinement)."
[Indeed, you cannot write "the function" since there is no uniquess -- and crucially so since the axiom of choice is used as you notice.] Of course keep your remark about AC.

Two asides:
* can you put the theorems about "finer covers" back to the mathbox, or is it too much work ?
* as a big contributor, your take on the discussion group thread "Philosophy and goals for set.mm" could be interesting.

nmegill:
Could you comment on @benjub's remarks?

benjub:
In particular: locfinref says "E. w..." but it looks to me that it suffices to take w = V.

tirix:
Thanks a lot for your comments, Benoît!

Following them, I pushed commit [d3f5a6f](https://github.com/metamath/set.mm/pull/1448/commits/d3f5a6f75894544532b3a0c2f54e716357afac32).

> As I understand it, you define covers as sets of sets, so you do not have an indexing to begin with. If you want to prove something saying what your initial comment means, then you should prove that there exists a refinement with a function as in ~reff which is bijective.

Yes, as you noticed, as it was worded, ~ locfinref was trivial. I've corrected the statement to provide the indexing function ` f `, which domain is a subset of the original cover:
` ( ph -> E. f ( ( Fun f /\ dom f C_ U ) /\ ( U Ref ran f /\ ran f e. ( LocFin ' J ) ) ) ) `
You could actually extend that function ` f ` to the whole of ` U `, by giving it the value `(/)` "empty set" outside of its current domain, with the extended index function still having the same properties: we might need this extension, but I'll keep the theorem in this form for now (local finiteness was not easy to prove!) 
I hope that this version is correct - please tell if you think additional explanation is needed.

As for your other points:
- I've corrected the two descriptions you mention.
- I agree with the fact that ` Ref ` is backwards: ` U Ref V ` means currently "` V ` is a refinement of ` U `". This is Jeff Madsen's definition and I kept it as is, but my first intuition was also the reverse. Also, I read that in some cases, "refines" is written with the symbol "<", which would also be the reverse direction. There are currently less than 20 theorems using this definition, so I can still change the direction if there is a consensus. @nmegill @digama0 what is your opinion? I think Norm did not want to disturb Jeff for this kind of questions. 
- I'm OK to move back the "finer" relation to Jeff's Mathbox, I'll do it in the next commit/PR. I just wanted to finish this long proof before refactoring.
- Sure, I'll give my opinion on the "philosophy and goals of set.mm" thread! :)

set.mm
```diff
@@ -440751,22 +440862,92 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+    locfinref.3 $e |- ( ph -> V C_ J ) $.
+    locfinref.4 $e |- ( ph -> U Ref V ) $.
+    locfinref.5 $e |- ( ph -> V e. ( LocFin ` J ) ) $.
+    $( A locally finite refinement of an open cover induces a locally finite
+       cover with the original index set.  This is fact 2 of
+       http://at.yorku.ca/p/a/c/a/02.pdf , it is expressed by exposing a
+       function ` f ` from the original cover ` U ` , which is taken as the
+       index set.  (Contributed by Thierry Arnoux, 29-Jan-2020.) $)
+    locfinref $p |- ( ph -> E. f ( ( Fun f /\ dom f C_ U )
+      /\ ( U Ref ran f /\ ran f e. ( LocFin ` J ) ) ) ) $=
```
digama0: This looks like a strange way to state this theorem. What is the purpose of the function `f`? It seems like all that matters is `ran f`, and the only thing about the refinement one can derive from the fact that `dom f C_ U` is `ran f ~<= U`. So why not say this instead?

benjub:
I think the new formulation is not precise enough: you need somehow to
relate $f$ to the original refinement $V$.  In your formulation, you need
to add to the conclusion "ran f = V", but you also need to more than that.
The solution I proposed above is to state that there is function as in
~reff which is a surjection (I mistakenly wrote above "bijection"). This
could be stated as
  E. f ( f : V -surjection-> U  /\ A. x e. V x C_ ( f ` x ) )

As for reversing Ref, your remark mentioning the order relation is on
point.  If you also want a more "high tech" rationale, you can look at
"Grothendieck topologies" for which refinements are always written in the
direction "A refines B", not "B is refined by A".  Hopefully @digama0 and
@nmegill will agree with this "swap".

benjub:
EDIT: my above remark is wrong too.  I'll have to find a pen and paper and
think a bit... maybe tonight !

benjub:
Ok, I was thinking of something else.  I just had a look at the paper you
cited. I would write:
  E. f ( f : U --> ~P V /\ U Ref ran f /\ ran f e. ( LocFin ` J ) )
That is, Thierry's formulation but being a bit more precise about f, which
I think could be useful.

tirix:
Actually I think Mario's proposal is the best (same statement power, simpler formula). But still, I would like to see how this theorem is used to find out the best formulation.

@benjub Giving the codomain ` ~P V ` of `f` is redundant with ` ran f ` being locally finite, so I think ` f Fn U ` would be sufficient in your case. I think the formula you propose can be proved, but the function I have right now would have to be completed by "padding with the empty set" outside of its current domain, which is smaller than `U`.

benjub:
Ok.  I think it's useful to be as precise as possible in lemmas, but as you say, applications will tell what formulation is best suited.  By the way, I think my proposal should have read
  f : U --> { x | E. y e. ~P V x = U. y }

nmegill:
@benjub's - are you comfortable with merging this PR?  If all your comments haven't all been addressed, they can be done in new PR.  (As I mentioned I'm not an expert on what refinement terminology is standard.)

benjub:
@nmegill : Yes, good to merge !

PR #1450, opened 2020-01-30
Examples for magmas and semigroups

avekens:
In AV's mathbox only:
* alternate predicate "is a magma"
* (generalized) relationships between +g and +f (for magmas)
* example of a magma which is not a semigroup
* example of a semigroup which is not a monoid
* proper inclusion chain Mnd C. SGrp C. Mgm
* auxiliary theorems to prove the previously mentioned results


avekens:
In addition, I have shown that RR*s is a magma, but not a semiring, as indicated in the comments of ~xrsmcmn and ~xrs1mnd. I pointed out in my Google group post https://groups.google.com/d/msg/metamath/mnkBQV1_cXc/CEEnCeHyAQAJ.

tirix:
I may be a bit too late on my remarks, since the PR is already closed :)
- I think RR*s is a very good example of magma / not semiring!
- Your example ` ( x e. S , y e. S |-> if ( ( x = A /\ y = A ) , B , A ) ) ` is actually the boolean AND operation, isn't it? It would be nice to use the already defined ` T. ` and ` F. ` and to state it explicitly.
- Why have you renamed "operation" into "function" in your last commit? I think operation is the right term / is used throughout the rest of set.mm.

avekens:
It's never too late for remarks and questions ;)

> * I think RR*s is a very good example of magma / not semiring!

Thanks!

> * Your example `( x e. S , y e. S |-> if ( ( x = A /\ y = A ) , B , A ) )` is actually the boolean AND operation, isn't it? It would be nice to use the already defined `T.` and `F.` and to state it explicitly.

No it's not AND, because A+A=B. Boolean AND would be associative, thus a semigroup.

> * Why have you renamed "operation" into "function" in your last commit? I think operation is the right term / is used throughout the rest of set.mm.

The comment refers to the symbol +f, which is called "group addition function" (see ~dfplusf), (maybe) to distinguish it from +g, the "group operation". Of course also +f is an operation, but its property to be a function is important, that's why "group addition function" is a good choice (not invented by me).

tirix:
Got it, thank you very much!

PR #1453, opened 2020-02-02
More locally finite refinements

tirix:
And a proof of the uniqueness of solutions to the two-square theorem.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -350,6 +350,7 @@
 "4syl" is used by "isumsup2".
 "4syl" is used by "kelac1".
 "4syl" is used by "kelac2".
+"4syl" is used by "locfinref".
 "4syl" is used by "lpssat".
 "4syl" is used by "lsmmod".
 "4syl" is used by "lssat".
@@ -13781,7 +13782,7 @@ New usage of "4atex2-0bOLDN" is discouraged (0 uses).
 New usage of "4atex2-0cOLDN" is discouraged (0 uses).
 New usage of "4ipval2" is discouraged (2 uses).
 New usage of "4ipval3" is discouraged (0 uses).
-New usage of "4syl" is discouraged (195 uses).
+New usage of "4syl" is discouraged (196 uses).
 New usage of "5oai" is discouraged (0 uses).
 New usage of "5oalem1" is discouraged (1 uses).
 New usage of "5oalem2" is discouraged (2 uses).


nmegill:
> And a proof of the uniqueness of solutions to the two-square theorem.

Although uniqueness isn't an explicit requirement of the MM100 two-square theorem (already added by Mario), @david-a-wheeler I wonder if it would make sense to mention it there.  It seems like a nice bonus point for the list even if not officially recognized.

benjub:
@nmegill: so, is it ok to "reverse" Ref, so that A Ref B means the former B Ref A, as discussed in the previous PR ?

A few comments to @tirix about possible improvements:

lfinun0: adding any set preserves local finiteness (finite + 1 is finite), so you could replace (/) with B (and relabel lfinunsn). Even if the theorem is only used for the moment with B = (/), I think it is worth to state it generally.

gcdnncl: relax to N e. ZZ ?

2sqcoprm: relax to A, B e. ZZ ?

2sqmod: relax to A, B, C, D e. NN0 ?  I do not know the best way to state uniqueness: either as you did, you impose A <_ B and C <_ D, or another solution is to not impose these two conditions and to state { A , B } = { C , D }.  Maybe an intermediate result worth stating is to impose only A, B, C, D e. ZZ and to state { ( A ^ 2 ) , ( B ^ 2 ) } = { ( C ^ 2 ) , ( D ^ 2 ) } ?

By the way: do we have in set.mm something like |- ( A ^ 2 e. Prime -> -. A e. QQ ) ?  Derived for instance from |- ( ( A ^ 2 e. ZZ /\ A e. QQ ) -> A e. ZZ ) if it is in set.mm ?

2sqmo: relax to "e. NN0" twice ?

locfinreflem: maybe add to the comment that the images of f are unions of sets in V ?

to used --> to use

pcmplfin, pcmplfinf: A paracompact space \` J \` --> A paracompact topology \` J \`

nmegill:
> @nmegill: so, is it ok to "reverse" Ref, so that A Ref B means the former B Ref A, as discussed in the previous PR ?

Yes, I have no objection.

david-a-wheeler:
@nmegill :
> Although uniqueness isn't an explicit requirement of the MM100 two-square theorem (already added by Mario), @david-a-wheeler I wonder if it would make sense to mention it there. It seems like a nice bonus point for the list even if not officially recognized.

It makes sense to me, and it's easy to make that an additional comment in the mm100 list.  I'll try to do that soon, though I don't know if I'll have time to do that today.

tirix:
@benjub I agree with your comments, I'll try to implement them later. 
For ~2sqmo, I would rather relax it only to NN0 than to ZZ, as the statement is much clearer in that case.

tirix:
Let's answer here to @benjub's comments:

> lfinun0: adding any set preserves local finiteness (finite + 1 is finite), so you could replace (/) with B (and relabel lfinunsn). Even if the theorem is only used for the moment with B = (/), I think it is worth to state it generally.

Yes, I've actually changed the proof to adding _any_ finite set, see #1460 

> gcdnncl: relax to N e. ZZ ?

Actually, there is already [~ gcdcl](http://us2.metamath.org/mpeuni/gcdcl.html) for that. My theorem was especially for the non-zero GCD. 

> 2sqcoprm: relax to A, B e. ZZ ?

Done in #1460 !

> 2sqmod: relax to A, B, C, D e. NN0 ? I do not know the best way to state uniqueness: either as you did, you impose A <_ B and C <_ D, or another solution is to not impose these two conditions and to state { A , B } = { C , D }. Maybe an intermediate result worth stating is to impose only A, B, C, D e. ZZ and to state { ( A ^ 2 ) , ( B ^ 2 ) } = { ( C ^ 2 ) , ( D ^ 2 ) } ?

As mentioned, I prefer to relax to NN0. Done in #1460 

> By the way: do we have in set.mm something like |- ( A ^ 2 e. Prime -> -. A e. QQ ) ? Derived for instance from |- ( ( A ^ 2 e. ZZ /\ A e. QQ ) -> A e. ZZ ) if it is in set.mm ?

Yes, there is [~expnprm](http://us2.metamath.org/mpeuni/expnprm.html), valid for any power.

> 2sqmo: relax to "e. NN0" twice ?

Done in #1460 

> locfinreflem: maybe add to the comment that the images of f are unions of sets in V ?
to used --> to use
pcmplfin, pcmplfinf: A paracompact space ` J ` --> A paracompact topology ` J `

All done in #1460

benjub:
Thanks @tirix !
> > gcdnncl: relax to N e. ZZ ?
> 
> Actually, there is already [~ gcdcl](http://us2.metamath.org/mpeuni/gcdcl.html) for that. My theorem was especially for the non-zero GCD.

What I meant was: in gcdnncl, relax to N e. ZZ, but keep the rest unchanged.

> > 2sqmod: relax to A, B, C, D e. NN0 ? I do not know the best way to state uniqueness: either as you did, you impose A <_ B and C <_ D, or another solution is to not impose these two conditions and to state { A , B } = { C , D }. Maybe an intermediate result worth stating is to impose only A, B, C, D e. ZZ and to state { ( A ^ 2 ) , ( B ^ 2 ) } = { ( C ^ 2 ) , ( D ^ 2 ) } ?
> 
> As mentioned, I prefer to relax to NN0. Done in #1460

Ok. And you also kept the two hypotheses A <_ B and C <_ D, so that you are able to deduce A = C and B = D, as opposed to my proposal of no hypotheses and { A , B } = { C , D }.  This is very reasonable, and my proposal may be an intermediate result in your proof, I haven't checked. You have probably noticed that your conclusion also follows when replacing the two hypotheses above with 0 <_ ( ( B - A ) x. ( D - C ) ) and that this latter statement is probably easily proved from two applications of your statement, provided one has in set.mm (is it there?) a theorem like 0 <_ ( A x. B ) <-> ( ( 0 <_ A /\ 0 <_ B ) \\/ ( A <_ 0 /\ B <_ 0 ) ).

> > By the way: do we have in set.mm something like |- ( A ^ 2 e. Prime -> -. A e. QQ ) ? Derived for instance from |- ( ( A ^ 2 e. ZZ /\ A e. QQ ) -> A e. ZZ ) if it is in set.mm ?
> 
> Yes, there is [~expnprm](http://us2.metamath.org/mpeuni/expnprm.html), valid for any power.

Thanks ! And the other theorem I mentioned is ~qexpz.  But ~expnprm is not derived from ~qexpz. As I wrote above, the derivation would probably give a shorter proof: if A \in \Q and N \geq 2, then: if x^n is not in \Z, then it is not in \P; else x^n \in \Z so by ~qexpz, we have x \in \Z, so x | x^n. Now if x = 1, then x^n = 1 \notin \P; also, if x = 0, then x^n = 0 \notin \P, so x \neq 0, so x \neq x^n, therefore x^n has a divisor other than 1 and itself, so it is not prime.
This is a bit long to formalize, but probably shorter than the current proof of ~expnprm ? Or did you check that it isn't, @digama0 ?

By the way: I haven't looked at the unicode for \` Ref \` in set.mm but it looks like it lacks surrounded spaces: since it is mainly used infixed, I would define it as " Ref ".

PR #1455, opened 2020-02-03
Swap definition of Ref

tirix:
As per Benoît's request and Norm's [comment](https://github.com/metamath/set.mm/pull/1453#issuecomment-581156560) in #1453 : 

> >@nmegill: so, is it ok to "reverse" Ref, so that A Ref B means the former B Ref A, as discussed in the previous PR ?

> Yes, I have no objection.


tirix:
As a note, I have *only* changed the definition of `Ref`, and not the definition of `Fne`, which has stayed (was returned) in Jeff's Mathbox. There were two theorems where Jeff introduced the relationship between `Fne` and `Ref`, and expressed that both relation held as `A ( Fne i^i Ref ) B`. With the inversion, this became `(A Fne B /\ B Ref A)`, which is less elegant. Those theorems are `~fnessref` and `~refssfne`

Issue #1457, opened 2020-02-03
Magmas and Semigroups


avekens:
This issue is a spin-off of #1389. According to the Google group discussion, see https://groups.google.com/d/topic/metamath/mnkBQV1_cXc/discussion, the definitions and theorems for magmas and semigroups (but not for "laws" and internal binary operations) can/should be moved from AV's mathbox to the main body of set.mm.

The definition and basic theorems for monoids should be cleaned up and adapted accordingly. Impacts of other sections shoud be avoided/minimized.

avekens:
After PR #1465 was accepted, this issue is almost done. Two points remain open:

- According to Norm's comment in PR #1465, the justification for mndmgm, grpmgm, grpsgrp should be discussed:
> It seems like you are adding all possible links from less general to more general structures (which will grow quadratically with the number of layers). I don't see mndmgm, grpmgm, grpsgrp ever being needed for more than 1 or 2 theorems if at all. I think we only need sgrpmgm, mndsgrp, grpmnd, which can be chained with syl when necessary.
> Occasionally we have a frequent need to go down more than one layer, e.g. rngmnd, but that is justified because it is used 80 times. As with definitions, we should try to add theorems based on need or at least anticipated need.

- Should subsection "Examples and counterexamples for magmas, semigroups and monoids" be moved from my mathbox into main set.mm? I would recomment this, putting this subsection after subsection "Free monoids" (last subsection in section "Monoids").

@nmegill, @digama0, @david-a-wheeler , @benjub, @tirix : what's your opinion?

avekens:
Current usage:
- mndmgm - 3 statements:  mndcl mndplusf srg1zr
- grpmgm - 1 statement: ringmgm
- grpsgrp -1 statement: grpmgm

Therefore, I think ~mndmgm should be kept, ~grpmgm and ~grpsgrp can be removed

benjub:
I do not know about the examples section.
The TOC is currently:

PART 10  BASIC ALGEBRAIC STRUCTURES
- ​10.1  Monoids
- - 10.1.1  Magmas   cplusf 15738
- - 10.1.2  Semigroups   csgrp 15754
- - 10.1.3  Definition and basic properties of monoids   cmnd 15762
- - 10.1.4  Monoid homomorphisms and submonoids   cmhm 15812
- - 10.1.5  Ordered group sum operation   gsumvallem1 15850
- - 10.1.6  Free monoids   cfrmd 15875
- ​10.2  Groups

[...]

PR #1459, opened 2020-02-03
Magmas and semigroups

avekens:
This is the first PR for #1457:
* auxiliary theorems moved from AV's and GS's mathboxes to main set.mm
* subsections "Magmas" and "Semigroups" moved from AV's mathbox to main set.mm

Subsection "Definition and basic theorems (for monoids)" revised:
* Symbol definitions for `Grp`, `invg`, `-g` and `.g` moved to section "Groups"
* Theorems ~mndmgm and ~ mndsgrp moved from AV's mathbox
* Theorems about `+f` moved to subsection "Magmas"
* proof of ~mndplusf shortened

Furthermore, some definitions/theorems of part 18 are marked as OLD/obsolete.

Before I resume the work, I want to ask if this approach is OK now, or if there are any comments/suggestions for improvements in detail so far (not to start the generall discussion again).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4068,8 +4068,8 @@
 "chub2i" is used by "pjclem1".
 "chub2i" is used by "qlaxr3i".
 "chub2i" is used by "sumdmdlem2".
-"clmgm" is used by "exidcl".
-"clmgm" is used by "mgmcllaw".
+"clmgmOLD" is used by "exidcl".
+"clmgmOLD" is used by "mgmcllaw".
 "cm0" is used by "chirred".
 "cm2j" is used by "cm2ji".
 "cm2ji" is used by "cm2mi".
@@ -4644,13 +4644,13 @@
 "df-m1r" is used by "mappsrpr".
 "df-md" is used by "mdbr".
 "df-metuOLD" is used by "metuvalOLD".
-"df-mgm" is used by "iopmapxp".
-"df-mgm" is used by "ismgm".
+"df-mgmOLD" is used by "iopmapxp".
+"df-mgmOLD" is used by "ismgmOLD".
 "df-mi" is used by "dmmulpi".
 "df-mi" is used by "mulpiord".
 "df-mndo" is used by "ismndo".
 "df-mndo" is used by "mndoisexid".
-"df-mndo" is used by "mndoissmgrp".
+"df-mndo" is used by "mndoissmgrpOLD".
 "df-mnf" is used by "mnfnre".
 "df-mnf" is used by "mnfxr".
 "df-mnf" is used by "pnfnemnf".
@@ -4765,9 +4765,9 @@
 "df-rq" is used by "dmrecnq".
 "df-rq" is used by "recmulnq".
 "df-scon" is used by "isscon".
-"df-sgr" is used by "issmgrp".
-"df-sgr" is used by "smgrpisass".
-"df-sgr" is used by "smgrpismgm".
+"df-sgrOLD" is used by "issmgrpOLD".
+"df-sgrOLD" is used by "smgrpisass".
+"df-sgrOLD" is used by "smgrpismgmOLD".
 "df-sh" is used by "issh".
 "df-shs" is used by "shsval".
 "df-sm" is used by "smfval".
@@ -8784,7 +8784,7 @@
 "isabloi" is used by "hhssabloi".
 "isabloi" is used by "hilablo".
 "isass" is used by "assasslaw".
-"isass" is used by "issmgrp".
+"isass" is used by "issmgrpOLD".
 "isass" is used by "sgrpplusgaop".
 "isblo" is used by "bloln".
 "isblo" is used by "htthlem".
@@ -8819,7 +8819,7 @@
 "isexid" is used by "exidres".
 "isexid" is used by "isexid2".
 "isexid" is used by "ismndo".
-"isexid" is used by "opidon".
+"isexid" is used by "opidonOLD".
 "isexid2" is used by "exidu1".
 "isgrp2d" is used by "ghgrp".
 "isgrp2d" is used by "isgrp2i".
@@ -8863,11 +8863,11 @@
 "islpolN" is used by "lpolsatN".
 "islpolN" is used by "lpolvN".
 "islpoldN" is used by "dochpolN".
-"ismgm" is used by "clmgm".
-"ismgm" is used by "iopmapxp".
-"ismgm" is used by "issmgrp".
-"ismgm" is used by "mgmplusgiop".
-"ismgm" is used by "opidon".
+"ismgmOLD" is used by "clmgmOLD".
+"ismgmOLD" is used by "iopmapxp".
+"ismgmOLD" is used by "issmgrpOLD".
+"ismgmOLD" is used by "mgmplusgiop".
+"ismgmOLD" is used by "opidonOLD".
 "ismndo" is used by "ismndo1".
 "ismndo1" is used by "ismndo2".
 "ismndo1" is used by "rngomndo".
@@ -8916,9 +8916,9 @@
 "issh2" is used by "shmulclOLD".
 "issh2" is used by "shscli".
 "issh3" is used by "nlelshi".
-"issmgrp" is used by "ismndo1".
-"issmgrp" is used by "smgrpass".
-"issmgrp" is used by "smgrpmgm".
+"issmgrpOLD" is used by "ismndo1".
+"issmgrpOLD" is used by "smgrpassOLD".
+"issmgrpOLD" is used by "smgrpmgm".
 "isssp" is used by "hhsssh2".
 "isssp" is used by "hhsst".
 "isssp" is used by "sspba".
@@ -9799,10 +9799,10 @@
 "minvecolem7" is used by "minveco".
 "mndoisexid" is used by "mndomgmid".
 "mndoisexid" is used by "rngo1cl".
-"mndoismgm" is used by "isdrngo2".
-"mndoismgm" is used by "mndomgmid".
-"mndoismgm" is used by "rngo1cl".
-"mndoissmgrp" is used by "mndoismgm".
+"mndoismgmOLD" is used by "isdrngo2".
+"mndoismgmOLD" is used by "mndomgmid".
+"mndoismgmOLD" is used by "rngo1cl".
+"mndoissmgrpOLD" is used by "mndoismgmOLD".
 "mndomgmid" is used by "isdrngo2".
 "mndomgmid" is used by "ismndo2".
 "mndomgmid" is used by "rngoidmlem".
@@ -11266,9 +11266,9 @@
 "opelreal" is used by "axrnegex".
 "opelreal" is used by "axrrecex".
 "opelreal" is used by "ltresr".
-"opidon" is used by "opidon2".
-"opidon" is used by "rngopid".
-"opidon2" is used by "exidreslem".
+"opidon2OLD" is used by "exidreslem".
+"opidonOLD" is used by "opidon2OLD".
+"opidonOLD" is used by "rngopidOLD".
 "opsqrlem2" is used by "opsqrlem6".
 "opsqrlem3" is used by "opsqrlem4".
 "opsqrlem3" is used by "opsqrlem5".
@@ -12256,10 +12256,10 @@
 "rngomndo" is used by "rngo1cl".
 "rngomndo" is used by "rngoidmlem".
 "rngon0" is used by "rngoueqz".
-"rngopid" is used by "exidcl".
-"rngopid" is used by "exidresid".
-"rngopid" is used by "isexid2".
-"rngopid" is used by "ismndo2".
+"rngopidOLD" is used by "exidcl".
+"rngopidOLD" is used by "exidresid".
+"rngopidOLD" is used by "isexid2".
+"rngopidOLD" is used by "ismndo2".
 "rngoridm" is used by "rngonegmn1r".
 "rngoridm" is used by "rngoridfz".
 "rngoridm" is used by "rngoueqz".
@@ -12755,8 +12755,8 @@
 "smfval" is used by "nvvop".
 "smfval" is used by "phop".
 "smfval" is used by "phpar".
-"smgrpass" is used by "ismndo1".
-"smgrpismgm" is used by "mndoismgm".
+"smgrpassOLD" is used by "ismndo1".
+"smgrpismgmOLD" is used by "mndoismgmOLD".
 "smgrpmgm" is used by "ismndo1".
 "snatpsubN" is used by "pclfinN".
 "snatpsubN" is used by "pclfinclN".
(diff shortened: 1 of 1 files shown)


nmegill:
I agreed with importing semigroups, not magmas.  In other words, I meant the definition
```
 df-sgrp $a |- SGrp = { g |
                           [. ( Base ` g ) / b ]. [. ( +g ` g ) / o ].
                           A. x e. b A. y e. b A. z e. b ( ( x o y ) e. b
                         /\ ( ( x o y ) o z ) = ( x o ( y o z ) ) ) } $.
```
not
```
 df-sgrp $a |- SGrp = { g e. Mgm |
                           [. ( Base ` g ) / b ]. [. ( +g ` g ) / o ].
                           A. x e. b A. y e. b A. z e. b
                           ( ( x o y ) o z ) = ( x o ( y o z ) ) } $.
```
I agreed with importing semigroups to address the complaint that df-mnd was hard to read because it was too complex.  Magmas are completely useless for set.mm purposes, because we already have a well-developed theory of binary operations, and it makes things more complex not simpler. 

On 2/1 I wrote "I might agree to [semigroups] it anyway just to get the Bourbaki people off my back.  :)  (I think magma would be going to far, though.)"  Perhaps I should have been clearer.  I still don't like semigroups, but agreed to it as a concession.

@digama0 how do you feel about this?

avekens:
> On 2/1 I wrote "I might agree to [semigroups] it anyway just to get the Bourbaki people off my back. :) (I think magma would be going to far, though.)" Perhaps I should have been clearer. I still don't like semigroups, but agreed to it as a concession.

I was not sure, but I had feared it. In my opinion, however, it makes no sense to introduce a definition only for semigroups and not for magmas, or of internal binary operations at least (whereby the definition of magmas would better fit into the extensible structure framework, therefore I took this approach) - it would be half-assed. For me the fact that df-mnd is hard to read is not sufficient for a new definition - it's their (formal) applications (giving examples for non-magmas, magmas which are no semigroups, semigroups which are no monoids, etc. It may be useful to know that the positive integers are no monoid, and that the odd numbers are not even a magma under addition...). 

But as I said before, I do not want to restart the general discussion again. Therefore there are only two options: to accept this PR or to reject it completely.   

PS: I would not take it personally if this PR is rejected, it is the unrestricted right of the owner(s) of set.mm to decide what should be included. But I would stop contributing in this area after I tidied up my mathbox a little bit...

nmegill:
I am admittedly quite torn about this PR, with my opinions and philosophy on definitions now well-known here and elsewhere.  But like I said it is only a tiny part of set.mm.  It isn't worth creating discord among important contributors.

I will accept this PR if Mario agrees (or maybe if he doesn't).  In the worst case, because magma and semigroup have no deep consequences that I know of, they could probably be removed without a huge amount of effort, if we change our collective minds in the future.

Can you let us know what you think, @digama0 ?  Thanks.

david-a-wheeler:
The Wikipedia page on https://en.wikipedia.org/wiki/Semigroup prominently shows a figure relating various kinds of algebraic structures, including semigroup, and magma is essentially the root of them all. That doesn't mean set.mm has to include magma, but including a definition because it's the basis of many other things is a defensible exception to the general policy.

digama0:
I noticed this "scope creep" as well, and I am not comfortable with it at all. Regarding making definitions easier to read, I think this is mostly a problem with the HTML printing - line breaks would do most of the job of breaking it up into bite size chunks. Alternatively, the comment can give a detailed breakdown of the definition parts and give them names, like so:

> A monoid is a structure with a binary operation which is:
> * closed in the structure (`( x o y ) e. b`)
> * is associative (`( x o ( y o z ) ) = ( ( x o y ) o z )`)
> * and has an element `e` which is a left (`( e o x ) = x`) and right (`( x o e ) = x`) identity

I think this (or a modification thereof) is sufficient to quell any concerns about unreadability of the definition itself, and I would prefer this to any formal modifications.

digama0:
> The Wikipedia page on https://en.wikipedia.org/wiki/Semigroup prominently shows a figure relating various kinds of algebraic structures, including semigroup, and magma is essentially the root of them all. That doesn't mean set.mm has to include magma, but including a definition because it's the basis of many other things is a defensible exception to the general policy.

The reason set.mm has *monoids* is because they are "the root of them all". This is a relative notion, but the greatest lower bound of all algebraic structures in set.mm is monoids, not magmas.

nmegill:
@digama0  - I agree this PR is "scope creep" and like you would prefer not to import it.  But given the friction that it is causing, I think we should make an exception for this special case.  We can make it clear that we don't plan to accept these kinds of definitions (not driven by mathematical need) on a regular basis.

Let me note that df-mgm involves the importation of only 8 associated theorems, not a large number.  Unless I missed one, they are: ismgm, ismgmn0, mgmimp, mgmcl, isnmgm, mgmplusf, plusfreseq, mgmplusfreseq.  At this point it would be rather easy to remove the df-mgm stuff with a few edits related to df-sgrp.  I don't know how deeply it will become embedded in the long run, but I can't see how it would be useful in general.  I suppose we could tag the definition and theorems as "discouraged" for now, to prevent their casual or accidental use, if that would make it more agreeable to you.

Actually I think only 3 or 4 of the 8 theorems are really necessary.  For the others, here are my comments.

ismgmn0 - I'm not sure whether this one or ismgm will be more useful, but requiring the base set to be nonempty to avoid "M e. V" makes it somewhat less general.  Of course, in practice empty base sets can't exist in more specialized structures such as monoids that require existence of a special element like an identity element.

mgmimp - I don't think this will be used enough to justify its overhead.  If needed in a proof, it can be obtained from mgmcl using ralrimivva.

isnmgm - This is kind of an odd theorem, which I don't see being used enough to justify its overhead.  We don't have, for example, "is not a group"; we would just prove that directly for the special case we want to show.  BTW here's a shorter proof: 
```
      ( wcel co wnel cmgm wa mgmcl 3expib com12 nelcon3d 3impia ) CAHZDAHZCDEIZ
      AJBKJRSLZBKTABKHZUATAHZUBRSUCABCDEFGMNOPQ $.
```

(A side comment on "e/":  while we don't prohibit or even "discourage" it, unlike the ubiquitous "=/=" it has never really paid off; proofs are almost always longer because of converting back and forth to "-.  A e. B".  For theorems intended to be used often, "-.  A e. B" almost always results in shorter proofs.  I've sometimes thought that adding "e/" was a mistake, but sometimes it looks nice in endpoint theorems like sqrt2irr.)

plusfreseq, mgmplusfreseq - dependence on the empty set not being in the base set isn't something we often find useful and might even be considered poor style, since it would make theorems less general.  Right now I don't see an application for these.

avekens:
> @digama0 - I agree this PR is "scope creep" 

If it is a "scope creep" depends on the view on it: for me (and others) it is a necessity to provide a complete picture of this topic.

> and like you would prefer not to import it. But given the friction that it is causing, I think we should make an exception for this special case. We can make it clear that we don't plan to accept these kinds of definitions (not driven by mathematical need) on a regular basis.

Of course I agree to this in general. The "mathematical need", however, is to be discussed in each single case.

> Let me note that df-mgm involves the importation of only 8 associated theorems, not a large number. Unless I missed one, they are: ismgm, ismgmn0, mgmimp, mgmcl, isnmgm, mgmplusf, plusfreseq, mgmplusfreseq. 

~plusfreseq does not depend on the definition of a magma.

> At this point it would be rather easy to remove the df-mgm stuff with a few edits related to df-sgrp. I don't know how deeply it will become embedded in the long run, but I can't see how it would be useful in general. I suppose we could tag the definition and theorems as "discouraged" for now, to prevent their casual or accidental use, if that would make it more agreeable to you.

If this (to tag the definition and theorems as "discouraged") is the price to be paid that they are accepted, I would agree. From my point of view, however, this seems to cause a self-fulfilling prophecy - in two years or so you can say: "I was right, the definition of magmas is useless, because nobody used it...". In mathboxes their usage should be allowed, at least.

> Actually I think only 3 or 4 of the 8 theorems are really necessary. For the others, here are my comments.
> 
> ismgmn0 - I'm not sure whether this one or ismgm will be more useful, but requiring the base set to be nonempty to avoid "M e. V" makes it somewhat less general. Of course, in practice empty base sets can't exist in more specialized structures such as monoids that require existence of a special element like an identity element.

This theorem is useful for proving that a concretely given structure is a magma (see ~mgm2nsgrplem1 and ~sgrp2nmndlem1 - I have additional cases in mind...). I do not know if there is already a theorem providing conditions for an extensible structure to be a set, but to show that the base set is not empty is almost always easy.

> mgmimp - I don't think this will be used enough to justify its overhead. If needed in a proof, it can be obtained from mgmcl using ralrimivva.

Currently, ~mgmimp is (essentially) only used to prove ~mgmcl. I do not mind if it is removed.

> isnmgm - This is kind of an odd theorem, which I don't see being used enough to justify its overhead. We don't have, for example, "is not a group"; we would just prove that directly for the special case we want to show. 

Like ~ ismgmn0, this theorem is useful for proving that a concretely given structure is **not** a magma. I think it would pay off if there are 2-3 cases in which it is used (currently, it isn't used, but I have some cases in mind).

> BTW here's a shorter proof:
> 
> ```
>       ( wcel co wnel cmgm wa mgmcl 3expib com12 nelcon3d 3impia ) CAHZDAHZCDEIZ
>       AJBKJRSLZBKTABKHZUATAHZUBRSUCABCDEFGMNOPQ $.
> ```
> 

That's great - of course the shorter proof should be in set.mm.

> (A side comment on "e/": while we don't prohibit or even "discourage" it, unlike the ubiquitous "=/=" it has never really paid off; proofs are almost always longer because of converting back and forth to "-. A e. B". For theorems intended to be used often, "-. A e. B" almost always results in shorter proofs. I've sometimes thought that adding "e/" was a mistake, but sometimes it looks nice in endpoint theorems like sqrt2irr.)

"e/" is not used often, and makes proofs longer, because there were only a few theorems supporting it (about 8 I think) - I added 3 theorems which were useful (and used) for my recent proofs. I think using "e/" more often could reduce size (it saves 3 characters) and improve readability.
 
> plusfreseq, mgmplusfreseq - dependence on the empty set not being in the base set isn't something we often find useful and might even be considered poor style, since it would make theorems less general. Right now I don't see an application for these.

I had a use case for these theorems, but after I had proven them, I found a simpler way for my initial problem. I kept these theorems because I thought they are interesting, explaining the differences between +g and +f (these theorems are based on weaker conditions than the available ~plusfeq). The dependency on the containment of the empty set is a consequence of our usage of the empty set for meaning "not defined".    Nevertheless, I have no problems with moving these 2 theorems back into my mathbox.

nmegill:
Unless @digama0 wants to make another comment, I will be merging this in about an hour so we can put this issue to rest.  I'm sorry if we can't make everyone happy.

nmegill:
@avekens With my comments in mind, use your own judgment to choose which magma theorems we should keep in the main set.mm.  I don't think there is a need to mark anything "discouraged" at this point.

nmegill:
Currently df-mnd is not dependent on semigroups and magmas.  While I'm sure you want to swap df-mnd and df-mndALT, I would prefer to do that in small steps.  First, add a theorem dfmndlem that says "{ g | [. ( Base \` g ) / b ]...} = { g e. SGrp |  [. ( Base \` g ) / b ]...}" to connect the definiens of df-mnd and df-mndALT (that uses neither of those definitions).  Use it to prove df-mndALT from df-mnd as theorem dfmnd2.  At that point you can start adding theorems such as "a monoid is a semigroup".  Later, it will be trivial to swap df-mnd and dfmnd2 using dfmndlem without affecting any other theorems. In the meantime, we can think about how the comments of the old and new definitions should read.  For example, few college textbooks define magma and semigroup, and I would like to describe to the reader, via the comments, how to understand monoid without necessarily learning semigroup and magma (if that is their choice).   After we are happy with that, we can do the final swap of definitions.  Fair enough?

digama0:
(Sorry about non-response, I was on a flight most of the day.) I'm deferring to Norm's decision here, however that goes, but if course I have opinions about all this and you've already heard them for the most part. 

> > I think we should make an exception for this special case. We can make it clear that we don't plan to accept these kinds of definitions (not driven by mathematical need) on a regular basis.
> 
> Of course I agree to this in general. The "mathematical need", however, is to be discussed in each single case.

I still think that "two people need it" is a clean, simple, and objective metric for inclusion. Mathematical interestingness is subjective, and there is no harm in having definitions in a mathbox.

> > At this point it would be rather easy to remove the df-mgm stuff with a few edits related to df-sgrp. I don't know how deeply it will become embedded in the long run, but I can't see how it would be useful in general. I suppose we could tag the definition and theorems as "discouraged" for now, to prevent their casual or accidental use, if that would make it more agreeable to you.
> 
> If this (to tag the definition and theorems as "discouraged") is the price to be paid that they are accepted, I would agree. From my point of view, however, this seems to cause a self-fulfilling prophecy - in two years or so you can say: "I was right, the definition of magmas is useless, because nobody used it...". In mathboxes their usage should be allowed, at least.

I don't think magmas or semigroups should be discouraged, for exactly this reason. I'm still hoping that some actual use case crops up so I can support this more wholeheartedly. I have nothing against semigroups, but no one is doing actual "semigroup theory" here, so I don't see the point. (Yes, there is such a thing as semigroup theory, it's not a completely trivial subject, but so far none of the theorems in this area have been formalized.)

> > Actually I think only 3 or 4 of the 8 theorems are really necessary. For the others, here are my comments.
> > ismgmn0 - I'm not sure whether this one or ismgm will be more useful, but requiring the base set to be nonempty to avoid "M e. V" makes it somewhat less general. Of course, in practice empty base sets can't exist in more specialized structures such as monoids that require existence of a special element like an identity element.
> 
> This theorem is useful for proving that a concretely given structure is a magma (see ~mgm2nsgrplem1 and ~sgrp2nmndlem1 - I have additional cases in mind...). I do not know if there is already a theorem providing conditions for an extensible structure to be a set, but to show that the base set is not empty is almost always easy.
> [...]
> > plusfreseq, mgmplusfreseq - dependence on the empty set not being in the base set isn't something we often find useful and might even be considered poor style, since it would make theorems less general. Right now I don't see an application for these.
> 
> I had a use case for these theorems, but after I had proven them, I found a simpler way for my initial problem. I kept these theorems because I thought they are interesting, explaining the differences between +g and +f (these theorems are based on weaker conditions than the available ~plusfeq). The dependency on the containment of the empty set is a consequence of our usage of the empty set for meaning "not defined". Nevertheless, I have no problems with moving these 2 theorems back into my mathbox.

There are theorems showing that an extensible structure is a set. In fact, there is a "general wellformedness" predicate [~ df-struct](http://us.metamath.org/mpeuni/df-struct.html) for structures, which implies that it is a set and a partial function and can be proven efficiently via [~ strleun](http://us.metamath.org/mpeuni/strleun.html) et al.

I don't like theorems that make assumptions like "`(/) e/ A`" as a result of exploiting function value out of domain because this breaks our soft type system. General structures can have any set as base, there is no prohibition on the use of the empty set, so these theorems become awkwardly hobbled.

> > isnmgm - This is kind of an odd theorem, which I don't see being used enough to justify its overhead. We don't have, for example, "is not a group"; we would just prove that directly for the special case we want to show.
> 
> Like ~ ismgmn0, this theorem is useful for proving that a concretely given structure is **not** a magma. I think it would pay off if there are 2-3 cases in which it is used (currently, it isn't used, but I have some cases in mind).

There is no such thing as "not a magma". The +g slot should always contain a binary operation satisfying the constraints of a magma, unless it is not used at all. Otherwise, it's not a well formed structure, and operations like df-ress will not work properly on it. This is one more reason why "Magma" is not a useful class, because it is equivalent to "has a correct +g slot" with no constraints.

avekens:
Thank you very much, Norm and Mario, for finally accepting this PR, and for your obligingness towards me and the others (there are at least 3). And also many thanks for your remarks. I'll come back to them in my next PRs (I started this already, see the current PR #1461).

avekens:
> @avekens With my comments in mind, use your own judgment to choose which magma theorems we should keep in the main set.mm. I don't think there is a need to mark anything "discouraged" at this point.

See PR #1461 .

PR #1461, opened 2020-02-06
Revision of definition of monoids

avekens:
See also issue #1457 and PR #1459.

Improvements for magmas and semigroups (see discussion in PR #1459):
* ~mgmimp replaced in proofs of other theorems and deleted
* proof of ~isnmgm shortened
* ~plusfreseq and ~mgmplusfreseq moved back into AV's mathbox

Revision of definition of monoids:
* marking old definition (and the only theorem ~ismnd using it) as obsolete
* new definition and theorem ~ismndALT moved from AV's mathbox
* new theorem ~ismnddef
* proof of ~ismnd revised

Remark: ~ismnd contains an imperfection, as detected by Benoît (the ` A. c e. B ` is unnecessarily placed before ` ( a .+ b ) e. B `). In a next step, ~ismnd will therefore be replaced by ~ismndALT (7 occurences), deleted afterwards and finally ~ismndALT will be renamed to "ismnd".

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4070,7 +4070,6 @@
 "chub2i" is used by "qlaxr3i".
 "chub2i" is used by "sumdmdlem2".
 "clmgmOLD" is used by "exidcl".
-"clmgmOLD" is used by "mgmcllaw".
 "cm0" is used by "chirred".
 "cm2j" is used by "cm2ji".
 "cm2ji" is used by "cm2mi".
@@ -4645,10 +4644,10 @@
 "df-m1r" is used by "mappsrpr".
 "df-md" is used by "mdbr".
 "df-metuOLD" is used by "metuvalOLD".
-"df-mgmOLD" is used by "iopmapxp".
 "df-mgmOLD" is used by "ismgmOLD".
 "df-mi" is used by "dmmulpi".
 "df-mi" is used by "mulpiord".
+"df-mndOLD" is used by "ismndOLD".
 "df-mndo" is used by "ismndo".
 "df-mndo" is used by "mndoisexid".
 "df-mndo" is used by "mndoissmgrpOLD".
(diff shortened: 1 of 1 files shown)


avekens:
@nmegill : Concrning your comment in #1459:
> Currently df-mnd is not dependent on semigroups and magmas. While I'm sure you want to swap df-mnd and df-mndALT, I would prefer to do that in small steps. First, add a theorem dfmndlem that says "{ g | [. ( Base ` g ) / b ]...} = { g e. SGrp | [. ( Base ` g ) / b ]...}" to connect the definiens of df-mnd and df-mndALT (that uses neither of those definitions). Use it to prove df-mndALT from df-mnd as theorem dfmnd2. At that point you can start adding theorems such as "a monoid is a semigroup". Later, it will be trivial to swap df-mnd and dfmnd2 using dfmndlem without affecting any other theorems. In the meantime, we can think about how the comments of the old and new definitions should read. For example, few college textbooks define magma and semigroup, and I would like to describe to the reader, via the comments, how to understand monoid without necessarily learning semigroup and magma (if that is their choice). After we are happy with that, we can do the final swap of definitions. Fair enough?

I think it is not necessary to perform one (or more) intermediate steps. As you can see in set.mm of this PR, I really had to exchange the definitions and to modify the proof of ~ismnd only. There is actually no impact on other existing theorems. Furthermore, I kept the previous definition and version of ismnd (marked as OLD/obsolete). Therefore, it would be very easy to switch back to the old definition if required.

Finally, I added a lot of comments, explaining that a monoid can be defined in two ways: either using semigroups or providing the 3 rules directly. I think it cannot be done more.

 If you have concerns anyway, you should leave this PR open, we should discuss improvements here, and I could revise my contribution accordingly.

PR #1465, opened 2020-02-08
Subsection "Identity element" and applications for magmas

avekens:
See also #1457:
*  Theorems related to the "group identity element" `( 0g `` M )`, see ~ df-0g, which are already valid for magmas moved into a new subsection "Identity elements".
* Shorter proofs for ~mndcl and ~mndass using theorems for magams and semigroups. Old versions kept, marked as OLD/obsolete.
* ~grpsgrp, ~grpmgm, ~ringmgm moved to main set.mm
* proof of ~ gsumply1subr shortened (as application of the definition of a magma)
* ~ gsummgmpropd moved to main set.mm (as application of the definition of a magma)
* ~xrsmgm, ~xrsnsgrp, ~xrsmgmdifsgrp moved to main set.mm (as application of the definition of a magma and a semigroup)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15927,7 +15927,6 @@ New usage of "gsummhmOLD" is discouraged (2 uses).
 New usage of "gsummptfsaddOLD" is discouraged (0 uses).
 New usage of "gsummulc1OLD" is discouraged (1 uses).
 New usage of "gsummulc2OLD" is discouraged (1 uses).
-New usage of "gsumply1subrALT" is discouraged (0 uses).
 New usage of "gsumptOLD" is discouraged (1 uses).
 New usage of "gsumresOLD" is discouraged (6 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
It seems like you are adding all possible links from less general to more general structures (which will grow quadratically with the number of layers).  I don't see mndmgm, grpmgm, grpsgrp ever being needed for more than 1 or 2 theorems if at all.  I think we only need sgrpmgm, mndsgrp, grpmnd, which can be chained with syl when necessary.

Occasionally we have a frequent need to go down more than one layer, e.g. rngmnd, but that is justified because it is used 80 times.  As with definitions, we should try to add theorems based on need or at least anticipated need.

nmegill:
I'm merging this so we can move forward, but we should probably discuss the justification for mndmgm, grpmgm, grpsgrp.

PR #1467, opened 2020-02-10
Rename label fragments 'rng' to 'ring' if they are Ring (NOTE: might create merge conflicts for some unmerged changes)

david-a-wheeler:
NOTE: Because this renames a lot of labels and then rewraps,
many external changes will *NOT* merge cleanly with this.
If you have a not-yet-merged change, it needs to first
apply the renames on this branch, then rewrap.

This commit implments the rename of label fragments from 'rng' to 'ring'
when they involve a Ring, per:
https://github.com/metamath/set.mm/issues/1389

This also provides a useful test/demo of our scripts to enable
mass renames. These scripts provide a mechanism so we *can*
rename labels in a safer and more automated way.

These renames were developed by using an improved
"scripts/rename-proposals", reviewed, and then applied with an improved
"scripts/mass-rename".  The proposed script improvements have been
submitted as a separate commit (they are not in this one).

Signed-off-by: David A. Wheeler <dwheeler@dwheeler.com>

PR #1469, opened 2020-02-11
Fix label changes table

david-a-wheeler:
Fix the label changes table so it's more consistent.
In particular: dates with leading "0" become leading space,
renames are given in reverse order, and $e renames are not noted.
See the discussion here:
https://github.com/metamath/set.mm/pull/1467

These changes are based on the modified script scripts/mass-rename
in https://github.com/metamath/set.mm/pull/1468


avekens:
Great, and thank you for cleaning up my leading 0s...

PR #1472, opened 2020-02-13
Add scripts/jobs* to automate minimizing

david-a-wheeler:
The file scripts/jobs and its supporting file scripts/jobs.makefile
provide automation of mass minimizing. It uses GNU make's
job tracking mechanisms so that when a job is done, another
one is run until all the jobs are done.


david-a-wheeler:
Here are the scripts I just posted in the mailing list to automate
execution of jobs. It should work on Linux and Cygwin at least.

PR #1474, opened 2020-02-13
Improve job handling in scripts/jobs

david-a-wheeler:
Improve job handling in scripts/jobs. In particular:

* Allow users to enter a set of multiple jobs with ranges.
  For example, `scripts/jobs 112 120-130 140-150`
  does jobs 112, 120 through 130 (inclusive), and
  140 through 150 (inclusive).
  This is a change in the API, but a useful one.
* Use the log command as recommended by Norm.
  So the resulting log files for individual jobs should be
  exactly what was expected.

As before, it will run multiple processes in parallel, up to some
limit (by default the number of processors unless you set NPROC).
When one job is done it will start another one until all work is done.
You can view various log files to see its progress.

### DIFF
--- a/scripts/jobs
+++ b/scripts/jobs
@@ -1,7 +1,15 @@
 #!/bin/sh
 
-# Run set of Metamath minimization jobs from $1 (start) through $2 (end).
-# The number of jobs is the environment variable NPROC if set, else # CPUs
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
@nmegill - I think this version generates the log file format you were expecting.

This version supports multiple disjoint ranges.

PR #1477, opened 2020-02-16
Magmas, semigroups, monoids, rings; category of rings

avekens:
AV's mathbox only (all proofs are minimized):
* auxiliary theorems for subrings of the field of complex numbers
* theorem for showing that a structure is not a monoid
* the additive structure of all odd integers is not a magma
* the additive structure of positive integers is not a monoid
* auxiliary theorems for ring homomorphisms
* the non-unital ring of even integers
* the category of rings


benjub:
I haven't looked at the details of your mathbox, but you can prove (maybe that's your plan) that any ideal in a ring is a nonunital ring (and actually, any ideal in a nonunital ring is a nonunital ring), so in particular n\Z for any n \in \Z.

PR #1479, opened 2020-02-19
non-unital rings

avekens:
AV's mathbox only:
* theorems for non-unital rings
* theorems for structures with a constant group addition operation
* the ring constructed from a ` Z/nZ ` structure by replacing the (multiplicative) ring operation by a constant operation
* ideals as non-unital rings (thanks to @benjub for his hints, see PR #1477)


savask:
Looks very cool! Although, is there a specific reason why `lidlrng` has `R e. Ring` as a hypothesis, rather than `R e. Rng`? It seems that many theorems about rings can be generalized to rngs anyway.

avekens:
> Looks very cool! Although, is there a specific reason why `lidlrng` has `R e. Ring` as a hypothesis, rather than `R e. Rng`? It seems that many theorems about rings can be generalized to rngs anyway.

Yes, I plan to provide theorems for the category for non-unital rings shortly. I started, however, with unital rings to become familiar with categories in set.mm, because most of the required theorems are available, which is not the case for non-unital rings.

benjub:
@savask about ~lidlrng: you are right; that was noted in my comment of PR
#1477 mentioned above.

@avekens Suggestions:
an multiplicative --> a multiplicative
~lidlmmgm: somehow, it should be stated that this is a lemma for ~lidlmsgrp
unital ring --> (unital) ring  [[[to emphasize that in set.mm, "ring"
always means "unital ring" anyway]]]
(left) ideal --> left ideal
last two theorems: why are you assuming that R is a domain ? A ring is
enough. Side remark: as a rule, theorems like ~uzlidlring are probably more
usable than theorems like ~lidldomnnring, and are probably enough. Maybe I
would replace I e. Ring by I e. ( SubRing ` R ) .

avekens:
> Suggestions: 
an multiplicative --> a multiplicative 

I fixed that

> ~lidlmmgm: somehow, it should be stated that this is a lemma for ~lidlmsgrp 

why? I think this theorem stands for itself (and emphasizes the usefulness of a magma)

> unital ring --> (unital) ring [[[to emphasize that in set.mm, "ring" always means "unital ring" anyway]]] 

There are already many comments explicitly using "unital ring" (e.g. ~ringi, ~ringlz). I would prefer to use the term "unital rings" in comments for theorems for rings themselves.

> (left) ideal --> left ideal 

The term "left ideal" is almost never used! Even in comments for LIdeal only "ideal" is used (e.g. ~lidlss: "An ideal is a subset ..."). Therefore, I think I could have omitted also the "(left)".

> last two theorems: why are you assuming that R is a domain ? A ring is enough. 

I needed the property that there is no zero divisor to prove ~lidldomn1 (to show that the identity of an ideal is the identity of the ring), and this result was needed to prove ~uzlidlring. I do not know if this theorem was also valid if there are zero divisors...

> Side remark: as a rule, theorems like ~uzlidlring are probably more usable than theorems like ~lidldomnnring, and are probably enough. Maybe I would replace I e. Ring by I e. ( SubRing ` R ) .

You are right, such theorems would not be used in proofs of other theorems, but I wanted to emphasize that there are "Rng"s which are not "Ring"s.

benjub:
I think that the term "ideal", without any qualification, means by default,
"bilateral ideal". But more importantly, conventions may vary, so "ideal"
alone would be ambiguous, so stating explicitly "left ideal" is preferable
IMHO.

For the last two theorems: if R is a ring, L is a left ideal and 1 \in L,
then for all x \in R, one has x = x.1 \in L. So in your notation, you can
assume only R e. Ring, and you can conclude U = B (if U = {0}, then B is
the zero ring, so U = B in any case).

For ~lidldomn1 : it is actually sufficient to assume that R is a prime
ring, that L is a nonzero left ideal, that I in R (no need to be in L), and
that for all x in L, one has I x = x (no need for the other side). Indeed,
let u in L be different from 0 and let r in R. Then, (I - 1)ru = Iru - ru =
0 since ru \in L. Note that this is true for all r \in R. Therefore, since
R is a prime ring and u is not 0, then I - 1 = 0, so I = 1.

Also: in the previous PR, you proved that even integers form a non-unital
ring.  You can now deduce this fact from ~lidlrng (maybe you did it
already, I haven't checked).

avekens:
> I think that the term "ideal", without any qualification, means by default,
"bilateral ideal". But more importantly, conventions may vary, so "ideal"
alone would be ambiguous, so stating explicitly "left ideal" is preferable
IMHO.

Of course you are right, and therefore I added the "(left)", but as long as "ideal" is used in set.mm to mean "left ideal", I think this is sufficiently clear (BTW this is the analogous situation as for "unital rings", for which you proposed to use "(unital) ring"...).
 
> For the last two theorems: if R is a ring, L is a left ideal and 1 \in L,
then for all x \in R, one has x = x.1 \in L. So in your notation, you can
assume only R e. Ring, and you can conclude U = B (if U = {0}, then B is
the zero ring, so U = B in any case).

From I e. Ring I cannot always conclude that 1_R e. I. I only know that 1_I e. I. 
Therefore I had to show 1_I = 1_R first by using ~lidldomn1.

> For ~lidldomn1 : it is actually sufficient to assume that R is a prime
ring, that L is a nonzero left ideal, that I in R (no need to be in L), and
that for all x in L, one has I x = x (no need for the other side). Indeed,
let u in L be different from 0 and let r in R. Then, (I - 1)ru = Iru - ru =
0 since ru \in L. Note that this is true for all r \in R. Therefore, since
R is a prime ring and u is not 0, then I - 1 = 0, so I = 1.

Unfortunately, prime rings (`PrRing`) are defined in a mathbox only, and based on the old definition of rings (not as extensible structure). Therefore, a lot of preliminary work has to be done before...
 
>Also: in the previous PR, you proved that even integers form a non-unital
ring. You can now deduce this fact from ~lidlrng (maybe you did it
already, I haven't checked).

I haven't planned this yet, but this would be a good idea.

benjub:
Regarding uzlidlring: I see, thanks.  So what I proposed above, ( I e.
SubRing `R ), is actually necessary.  That is, the statement to prove is:
  |- ( R e. Ring -> ( ( LIdeal ` R ) i^i ( SubRing ` R ) ) = { ( Base ` R )
} )

Remark: I remember of a debate a few weeks ago about whether defining
substructures as sets or structured sets, and the former was chosen (I was
in favor of the latter).  Note that with the latter convention, the above
theorem would read:
  |- ( R e. Ring -> ( ( LIdeal ` R ) i^i ( SubRing ` R ) ) = { R } )

PR #1480, opened 2020-02-19
Half planes, split ~ mideulem into 3

tirix:
As discussed for the minimisation, this splits ~mideulem into 3 parts.
The proof for ~ mideulem was originally around 7300 chars. It was split into:

- ~ mideulem2 : 3835 chars
- ~ opphllem : 3752 chars (which is a lemma in [Schwabhauser])
- ~ mideulem : 450 chars

Minimise was not run yet on the separated lemmas.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,11 @@
 *~*
 mm.log
 *.demo.html
+,errs
+mmcomplex.html
+mmdeduction.html
 mmil.html
+mmnatded.html
+mmnf.html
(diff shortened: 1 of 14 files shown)


nmegill:
A merge conflict arose after I merged #1479.  I don't see anything obvious like the date.

tirix:
Actually it was only the date - I usually do the merges and resolve the conflicts, I just did not have time earlier.

PR #1489, opened 2020-02-25
Adjust iset.mm for new OLD and ALT rules

jkingdon:
This is based on requiring both discouragement tags for *OLD and *ALT theorems as described on the mailing list.

In many cases, we just get rid of theorems we don't need any more. Other cases may need a closer look later (for example, dvelimALT probably shouldn't be an ALT), but for now just add the tags.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -60,6 +60,8 @@
 "ax-mulcl" is used by "mulcl".
 "ax-mulcom" is used by "mulcom".
 "ax-mulrcl" is used by "remulcl".
+"ax16ALT" is used by "dvelimALT".
+"ax16ALT" is used by "dvelimfv".
 "axltirr" is used by "ltnr".
 "axlttrn" is used by "lttr".
 "axresscn" is used by "ax1cn".
@@ -71,6 +73,7 @@
 "df-pnf" is used by "mnfnre".
 "df-pnf" is used by "pnfnre".
 "df-tru" is used by "tru".
+"dvelimALT" is used by "hbsb4".
 "equsalh" is used by "dvelimALT".
 "equsalh" is used by "dvelimfALT2".
 "equsalh" is used by "dvelimfv".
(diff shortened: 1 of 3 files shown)


benjub:
Proposal: dvelimALT --> dvelimw (for "weak", and for similarity with ~dvelimhw in set.mm).

Why did you remove the three TODOs I had added: "remove in favor of ~ snexg / prexg / opexg"?  Maybe I should have phrased them: "TODO: replace its uses by uses of ~snexg and then remove it".

Maybe add to ~csbexga and ~csbexa something like: "TODO: if possible, remove the hypothesis \`A e. V \` as in set.mm (see ~csbexg there)." Or if you know that it is not possible, or probably not possible, add it as a remark ?

By the way: it has happened in the past that I searched for the string "TODO" in set.mm (and maybe also iset.mm) and actually did the thing, so it is not completely hopeless that these things will be done in the future!

jkingdon:
> Why did you remove the three TODOs I had added: "remove in favor of ~ snexg / prexg / opexg"? Maybe I should have phrased them: "TODO: replace its uses by uses of ~snexg and then remove it".

I thought it was clear from the rest of the comment, but we totally can bring the TODO wording back if it will increase the chances of it being done.

> Maybe add to ~csbexga and ~csbexa something like: "TODO: if possible, remove the hypothesis `A e. V ` as in set.mm (see ~csbexg there)." Or if you know that it is not possible, or probably not possible, add it as a remark ?

I suspect it isn't possible, but "know" would be overstating my confidence on this point.

jkingdon:
@benjub I've restored the TODO comments (with your new proposed wording).

I also expanded the discussion of set.mm versus iset.mm in mmil.html (which is where most of the comparison between the two goes) to reflect the issue of the A e. _V condition in csbex. My preference generally is to say what we know (that the set.mm proof uses case elimination) and not say as much about exactly what is possible (unless, of course, we have a proof of something being impossible as in http://us.metamath.org/ileuni/ordtriexmid.html ).

PR #1491, opened 2020-02-26
Fake associativity examples and natural number finiteness check

arpie-steele:
I was reading a textbook which stressed that associativity was a property of union and intersection, but not mixtures of union and intersection, so I was inspired to seek an example of when such a misunderstanding would be accidentally confirmed.

The hash function mixes the concept of cardinality of finite sets with that of the natural numbers (as a subset of the complex numbers). So I thought it would be nice to have definition of finiteness in terms of the natural-number-based hash function.


nmegill:
@arpie-steele Since you're here, let me note this somewhat unrelated observation mentioned by @benjub .  I had noticed it too but he mentioned it publicly first, so he gets credit. :)

While reviewing the recent minimizations that were done, these appeared in the "discouraged" file:
```
$ diff discouraged discouraged.old
...
984a985
> "ax-frege1" is used by "frege24".
989a991,992
> "ax-frege1" is used by "frege55lem1c".
> "ax-frege1" is used by "rp-6frege".
990a994
> "ax-frege1" is used by "rp-frege3g".
992a997
> "ax-frege2" is used by "frege39".
994a1000,1001
> "ax-frege2" is used by "rp-8frege".
> "ax-frege2" is used by "rp-frege24".
995a1003,1004
> "ax-frege2" is used by "rp-frege2ii".
> "ax-frege2" is used by "rp-frege2iii".
996a1006
> "ax-frege2" is used by "rp-frege4g".
13970,13971c14003,14004
< New usage of "ax-frege1" is discouraged (8 uses).
< New usage of "ax-frege2" is discouraged (6 uses).
---
> New usage of "ax-frege1" is discouraged (12 uses).
> New usage of "ax-frege2" is discouraged (12 uses).
```
I'm not sure whether you wanted these "minimized" or not.

However, I think these minimizations are legitimate, in the sense that no *new* axioms (such as ax-1 through ax-3) were brought in as a result.  For example, the proof of rp-6frege was shortened by using rp-a1i, thus eliminating the direct reference to ax-frege1 and instead making the reference indirect.

The only issue is whether you intended to show proofs directly from axioms.  If so, we will have to revert these from a previous set.mm, and you should add "(Proof modification is discouraged.)" to the comment in each of the theorems.

You can respond here to let me know if you want to do anything about this.  If so, let me know if you want me to revert them or if you will do it.

To make this analysis easier in the future, especially since you are using non-standard axioms, you may want to run 'minimize' in MM-PA on your proofs, using the command "minimize * /allow * /no_new ax-*".  It is optional though.

Norm

arpie-steele:
@nmegill Thank you for bringing this to my attention.

Sorry for not replying earlier. Shortly after submitting this pull request, I experienced a hardware failure that broke my home network. A new pull request reverts most of the NM improvements of *frege* proofs and moves *frege* axioms and theorems into the discouraged list.

Because Frege1879 is not concise there are missed opportunities for minimization, but my intent was to try to preserve as much of the proof structure from Frege1879 as is compatible with the necessary syntax changes and logical pitfalls of Frege's pre-set theory approach.

PR #1495, opened 2020-02-27
revert some accidental automatic minimizations and add discouragement…

benjub:
…s tags (only in my and WL's mathboxes)

Reverted to previous state proofs which were accidentally minimized in the global minimization, and added discouragement tags to:
bj-abf, bj-hbsb3, bj-nfs1, bj-csbprc,
and, in @wlammen 's mathbox, to:
wl-equsal, wl-sbal1, wl-sbal2.
For these three, I also minimized the proofs excepting use of the corresponding theorem (e.g. MM-PA>minminize */except equsal) and I added a link in the comment (e.g. "See also ~  equsal").

I think @wlammen 's wl-sb6nae can be deleted ? Even before the minimization, its proof was identical to that of sb4b. Although the label proposed by WL, i.e. "sb6nae" as "sb6 with a distinctor antecedent", also makes sense, as well as "sb4b" ("b" for "B"idirectional strengthening of sb4). Or maybe the current sb6 should be "sb6v".  Up to @nmegill

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18734,6 +18734,7 @@ Proof modification of "bj-abbidv" is discouraged (9 steps).
 Proof modification of "bj-abbii" is discouraged (17 steps).
 Proof modification of "bj-abeq1" is discouraged (36 steps).
 Proof modification of "bj-abeq2" is discouraged (47 steps).
+Proof modification of "bj-abf" is discouraged (13 steps).
 Proof modification of "bj-abfal" is discouraged (54 steps).
 Proof modification of "bj-abid2" is discouraged (14 steps).
 Proof modification of "bj-ablsscmn" is discouraged (10 steps).
@@ -18814,6 +18815,7 @@ Proof modification of "bj-cleqh" is discouraged (108 steps).
 Proof modification of "bj-cmnssmnd" is discouraged (36 steps).
 Proof modification of "bj-cmnssmndel" is discouraged (5 steps).
 Proof modification of "bj-con3ALT" is discouraged (63 steps).
+Proof modification of "bj-csbprc" is discouraged (47 steps).
 Proof modification of "bj-denotes" is discouraged (76 steps).
 Proof modification of "bj-df-clel" is discouraged (4 steps).
 Proof modification of "bj-df-cleq" is discouraged (4 steps).
@@ -18871,6 +18873,7 @@ Proof modification of "bj-hbntbi" is discouraged (31 steps).
 Proof modification of "bj-hbs1" is discouraged (22 steps).
 Proof modification of "bj-hbsb2av" is discouraged (32 steps).
 Proof modification of "bj-hbsb2v" is discouraged (31 steps).
+Proof modification of "bj-hbsb3" is discouraged (20 steps).
 Proof modification of "bj-hbsb3v" is discouraged (24 steps).
 Proof modification of "bj-imim2ALT" is discouraged (10 steps).
 Proof modification of "bj-imn3ani" is discouraged (18 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
I'm fine with deleting wl-sb6nae but that should be @wlammen's decision I think.  Maybe he has a reason for duplicating sb4b in his mathbox.

I like the *nae (Not-forAll-Equal) suffix for distinctors since it's consistent with other uses of "ae" and "nae".  The sb* theorems were mostly done very early (1993) and there was no consistent labeling convention back then.  Someone may want to go through the theorems with a "-. A. x x = y" antecedent and propose renames with "nae" suffixes - like we have for "v" (dv condition), "f" (not-free condition) suffixes.  When we have both "v" and "f" versions, usually one of them has the "v" or "f" suffix but not both - maybe that's not ideal, but it has usually been our practice, with the version added later or sometimes when less commonly used (whether "v" or "f") acquiring a suffix.  So we would have sb6, sb6f, sb6nae (in place of sb4b).  We also have sb6x with an alternate not-free condition - not sure what to call it.  Maybe sb6x and sb6f should be called sb6xf and sb6yf?   The "6" is kind of arbitrary, but at least sb*5* and sb*6* are consistent into class theory e.g. sbc5, sbc6.

The dvelim* theorems have both a distinctor antecedent and a $d, but introducing the distinctor to eliminate a $d is its purpose, so I think that name is ok as is.  We also have an "f" version, dvelimf, that uses a distinctor to eliminate a not-free condition (so it really should be named nfelim?).  I see that Mario changed the "ps -> A. x. ps" in the dvelimf conclusion to F/ x ps, which I think defeats the purpose of the theorem, which is to add "A. x" inside of a wff construction by using the distinctor antecedent in place of the not-free condition.  Anyway dvelim* are used much less frequently than I thought they would be, so maybe it's not worth worrying about.  I wonder if dvelim* would be easier to use if we made the conclusion "ps <-> A. x. ps", but I'm getting off-topic...

nmegill:
> I see that Mario changed the "ps -> A. x. ps" in the dvelimf conclusion to F/ x ps, which I think defeats the purpose of the theorem

I forget that the original dvelimf was renamed dvelimh, which has the conclusion "ps -> A. x. ps".  So the original is still there.

digama0:
Is

On Thu, Feb 27, 2020 at 7:13 PM nmegill <notifications@github.com> wrote:

> I'm fine with deleting wl-sb6nae but that should be @wlammen
> <https://github.com/wlammen>'s decision I think. Maybe he has a reason
> for duplicating sb4b in his mathbox.
>
> I like the *nae (Not-forAll-Equal) suffix for distinctors since it's
> consistent with other uses of "ae" and "nae". The sb* theorems were
> mostly done very early (1993) and there was no consistent labeling
> convention back then. Someone may want to go through the theorems with a
> "-. A. x x = y" antecedent and propose renames with "nae" suffixes - like
> we have for "v" (dv condition), "f" (not-free condition) suffixes. When we
> have both "v" and "f" versions, usually one of them has the "v" or "f"
> suffix but not both - maybe that's not ideal, but it has usually been our
> practice, with the version added later or sometimes when less commonly used
> (whether "v" or "f") acquiring a suffix. So we would have sb6, sb6f, sb6nae
> (in place of sb4b). We also have sb6x with an alternate not-free condition
> - not sure what to call it. Maybe sb6x and sb6f should be called sb6xf and
> sb6yf? The "6" is kind of arbitrary, but at least sb*5* and sb*6* are
> consistent into class theory e.g. sbc5, sbc6.
>
(At first I thought you meant it used the ph -> A. x ph version of
not-free, to which I would have suggested the "h" suffix.) For sb6x I would
suggest sb6f2 to indicate that it is another variation on sb6f (which is
itself a variation on sb6).

> The dvelim* theorems have both a distinctor antecedent and a $d, but
> introducing the distinctor to eliminate a $d is its purpose, so I think
> that name is ok as is. We also have an "f" version, dvelimf, that uses a
> distinctor to eliminate a not-free condition (so it really should be named
> nfelim?). I see that Mario changed the "ps -> A. x. ps" in the dvelimf
> conclusion to F/ x ps, which I think defeats the purpose of the theorem,
> which is to add "A. x" inside of a wff construction by using the distinctor
> antecedent in place of the not-free condition.
>
Are you sure? Given the proof of dvelimh, it seems that it's as easy as one
can expect to derive the ps -> A. x ps version from the original.

PR #1499, opened 2020-02-28
Expand apartness comments in iset.mm

### LINKED ISSUE #1498: iset.mm should briefly explain why apartness is needed
In iset.mm I see that `#` is defined using `#RR`, and `#RR = { <. x , y >. | ( ( x e. RR /\ y e. RR ) /\  ( x < y \/ y < x ) ) }`. But I didn't see any text in comments that explain *why* iset.mm needs to use "apartness" as something different from "not equal to". I guess I shouldn't be surprised that "not" causes complications, but I don't see any text briefly explaining the issue.

I think there should be a little text in iset.mm that explains iset.mm needs "apartness" to be different from "not equal". I'm sure the cited texts go into a lot of detail, but a hint within the text of iset.mm itself would be helpful.

This probably really a request for @jkingdon - but others are welcome to pipe in :-).

jkingdon:
Although we do cite some sources which describe this in more
detail, it is helpful to summarize with links to the relevant
theorems in iset.mm.

Fixes #1498

### DIFF
--- a/iset.mm
+++ b/iset.mm
@@ -69361,7 +69361,11 @@ $)
   ${
     $d x y $.
     $( Define real apartness.  Definition in Section 11.2.1 of [HoTT], p.
-       (varies).  (Contributed by Jim Kingdon, 26-Jan-2020.) $)
+       (varies).  Although ` #RR ` is an apartness relation on the reals (see
+       ~ df-ap for more discussion of apartness relations), for our purposes it
+       is just a stepping stone to defining ` # ` which is an apartness
+       relation on complex numbers.  On the reals, ` #RR ` and ` # ` agree
+       ( ~ apreap ).  (Contributed by Jim Kingdon, 26-Jan-2020.) $)
     df-reap $a |- #RR = { <. x , y >. | ( ( x e. RR /\ y e. RR ) /\
       ( x < y \/ y < x ) ) } $.
   $}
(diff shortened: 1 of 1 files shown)


benjub:
@jkingdon Maybe you can add to that second comment that in intuitionistic logic, one cannot prove ( -. x = y -> x # y ) , contrary to classical logic ?

[Maybe you will even formalize one day that this implication would imply excluded middle or something like that?]

david-a-wheeler:
Excellent, thanks for the explanation!

So where possible, ideally we'd prove that things are true when they are merely "not equal", since that is a weaker claim than "apart", but in many situations you need the stronger statement "apart". Is that right?

jkingdon:
> @jkingdon Maybe you can add to that second comment that in intuitionistic logic, one cannot prove ( -. x = y -> x # y ) , contrary to classical logic ?

I think this is covered for now by "Apartness implies negated equality, as seen at ~ apne , and the converse would also follow if we assumed excluded middle."
> 
> [Maybe you will even formalize one day that this implication would imply excluded middle or something like that?]

Don't know if it would be me or someone else but yeah maybe someone will contribute this, or some other proof that it cannot be proven. Not sure I particularly know how one would go about this, although it sort of does remind me of Remark 2.19 in [Geuvers].

jkingdon:
> "not equal" . . . is a weaker claim than "apart", but in many situations you need the stronger statement "apart". Is that right?

Not sure I follow the entire sentence, but the part I quoted here is correct.

PR #1500, opened 2020-02-29
Mgm/Rng homomorphisms; the category of non-unital rings

avekens:
AV's mathbox only:
* The empty magma: ~0mgm
* more auxiliary theorems for magmas: ~mgmpropd, ~ismgmd
* Subsection "Magma homomorphisms and submagmas" with definitions of MgmHom and SubMgm
* Subsection "Rng homomorphisms" with definitions of RngHomo and RngIsom (unfortunately, the symbols RngHom and RngIso are used already in another mathbox)
* Subsection "The category of non-unital rings" with definition of RngCat


benjub:
For the unicode/gif of SubMgm, RingCat, RngCat, I would not surround them by spaces, (e.g., I would use: htmldef "SubMgm" as "SubMgm";).
On the other hand, MgmHom, for instance, is typically used infix, so it should indeed be surrounded by spaces as you propose.
Of course, this is purely aesthetical and subjective.

avekens:
> For the unicode/gif of SubMgm, RingCat, RngCat, I would not surround them by spaces, (e.g., I would use: htmldef "SubMgm" as "SubMgm";).
OK, I removed the spaces...

nmegill:
What is the goal with your new definitions?  What are some theorems that you plan to use them for?

avekens:
> What is the goal with your new definitions? What are some theorems that you plan to use them for?

Gérard and Benoît asked for the category of non-unital rings. In a next step, I want to prove that  the category of unital rings is a subcategory of the category of non-unital rings, and that the zero ring is a terminal object in both categories, but an initial object only in the category of non-unital rings.

benjub:
In addition to @avekens's note: In the category of rings, the initial object is of course \Z (this is probably already in set.mm). Therefore, in RingCat, the initial object is different from the terminal object, whereas in RngCat, these are the same, hence the zero object (by definition, a zero object is an object which is both initial and terminal).

Note: Here, "the" is the "categorical the", that is, it means that the object is unique up to unique isomorphism. Since terminal and initial objects are unique up to unique isomorphism (almost by definition), the use of "the" is legitimate.

PR #1501, opened 2020-02-29
set.mm - update SRs mathbox

nmegill:


nmegill:
Something has gone badly wrong.  I'm pretty sure everything is up to date locally, but it's undoing the previous PR.  I give up and will commit/push this directly.

nmegill:
Has anyone seen this error?
! [remote rejected] develop -> develop (shallow update not allowed)
               error: failed to push some refs to 'https://github.com/nmegill/nmfork'

david-a-wheeler:
@nmegill - Yes. It means that something has a "shallow" copy of a repo instead of a "full" copy.

Git merges information by looking for a "common ancestor", and then looking for the changes since then. If a repo doesn't have all versions, it's possible to not be able to find the common ancestor.

However, making a full copy of a repo can take a while. So in some cases a repo clone will be created with a "--depth" parameter, to only copy a few versions back. We do this on Travis; the CI system isn't merging anything, it's just testing the current version, so there's no reason to copy the entire history of it just to create the current version.

I don't know exactly what command you used that caused the error, and I'm about to be mostly-inaccessible for a bit. You probably want to use `git fetch --unshallow REMOTE_BRANCH` on a repo, where `REMOTE_BRANCH` is `origin` or `upstream`. Hope that helps.

nmegill:
I deleted the fork, regenerated it, and re-cloned without --depth.  That seemed to fix the problem.

PR #1503, opened 2020-02-29
Mathbox fake associativity

arpie-steele:
Renamed two, added two principle proofs, added two corollaries.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -8456,7 +8456,6 @@
 "ifchhv" is used by "spansnj".
 "ifeq123d" is used by "fourierdlem103".
 "ifeq123d" is used by "fourierdlem104".
-"ifeq123d" is used by "fourierdlem112".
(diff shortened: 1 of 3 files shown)


arpie-steele:
"added two principle proofs" -- Hmm, this seems right on the dividing line between "principal" and "principle"

avekens:
I think changes in mathboxes should not be documented in the history at the top of set.mm.

nmegill:
@avekens is right, we don't normally document changes inside mathboxes, since often they are work in progress that is expected to change anyway.  They aren't intended to be used by others unless they are exported to the main set.mm (and we do document the exportation).  Could you remove them from the top of set.mm?  Thanks.

arpie-steele:
@avekens @nmegill Thanks for the tip. Those lines have been removed.

PR #1512, opened 2020-03-05
paired up three theorems from 18.1.7

TjWhale:
I changed 3 theorems from the end of 18.1.7 to have OLD at the end of the their names and added links to their pairs with dates. 

I think also some formatting changes from the metamath rewrap has got mixed up in there too.

Happy to make any changes required :)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -487,7 +487,7 @@
 "ablogrpo" is used by "cnnv".
 "ablogrpo" is used by "cnnvba".
 "ablogrpo" is used by "cnrngo".
-"ablogrpo" is used by "efghgrp".
+"ablogrpo" is used by "efghgrpOLD".
 "ablogrpo" is used by "ghablo".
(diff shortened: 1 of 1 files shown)


TjWhale:
It looks like the check that failed is that I renamed ghsubablo to ghsubabloOLD and because it is used in 1 proof that is counted as a new use of a discouraged theorem. However this isn't actually a new use, it's just a rename of where it is used in a proof already.

tirix:
As the new use of a discouraged theorem is the intended behaviour, then you can regenerate the "discouraged" file by yourself.
How to do it is explained [here](https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md#regenerating-the-discouraged-file).

TjWhale:
Thanks, I regenerated it. I did get this warning which I hope is ok

"warning: LF will be replaced by CRLF in discouraged.
The file will have its original line endings in your working directory."

PR #1518, opened 2020-03-07
Rename labels thema* to stoic*

david-a-wheeler:
Stoic logic has a number of themata (singular: thema).
However, the word "thema" is very general - and thus vague.
We originally used label names of thema*, but that also
made the names vague.

For clarity, this commit renames the labels from "thema" to "stoic",
e.g., Stoic logic thema 3 is renamed from "thema3" to "stoic3".

Many of these, especially stoic3, have been found to be useful
elsewhere. I find that unsurprising, since they form part of
the basis of Stoic logic. These have very old names, so I think
it makes sense to use their old names, just like we use the
old names "modus ponens" in ~ ax-mp or "modus tollens" in ~ mto .
The goal of this commit is clarity, to make it obvious that these
labels refer to the *Stoic logic* themata instead of other themata.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18244,6 +18244,8 @@ New usage of "stm1i" is discouraged (3 uses).
 New usage of "stm1ri" is discouraged (2 uses).
 New usage of "sto1i" is discouraged (3 uses).
 New usage of "sto2i" is discouraged (1 uses).
+New usage of "stoic2a" is discouraged (0 uses).
+New usage of "stoic2b" is discouraged (0 uses).
 New usage of "strb" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


arpie-steele:
@david-a-wheeler It appears you may need to regenerate the discouraged file, as well as determine if the "(New usage is discouraged.)" on stoic2a (identical to syldan) /stoic2b (identical to mpd3an3) is still wanted (I would argue that they are still discouraged in favor of the more methological, less historical naming of their aliases).

david-a-wheeler:
@arpie-steele : You're exactly right, I forgot the update the discouraged entries due to the renames. All should be well now.

PR #1521, opened 2020-03-09
fix a typo in nanbi

wlammen:
@nmegill  euanOLD needs special attention. It was mentioned on a web page. I don't know whether this necessitates an entry in the "Recent label changes" section.


nmegill:
~euanOLD should not be added to the label change list and shouldn't have been referenced in the ~conventions comments in the first place.  The only *OLD exception is ~19.43OLD, which is referenced on the mmrecent.html page, and it has a note in its comment not to delete it.  It should replace ~bi3antOLD in the ~conventions comment.

PR #1522, opened 2020-03-09
IMO 1972 B2

spolu:
Formalization of http://www.cs.ru.nl/~freek/demos/exercise/exercise.pdf

Intersting as a comparison to http://www.cs.ru.nl/~freek/demos/index.html


spolu:
Also includes a simple induction example (formalized to educate mysefl):

```
inductionexd $p |- ( N e. NN -> 3 || ( ( 4 ^ N ) + 5 ) ) $=
```

spolu:
(de Bruijn factor of the proof: ~12)

jkingdon:
Congratulations on finishing this proof 🎉

spolu:
\o/ thanks!

digama0:
By the way, for the de Bruijn factor calculation I think it is more accurate to count only the left column of the worksheet, skipping the theorem statements on each line. For the set.mm count you can approximate this by looking at the length of the label list, although that doesn't have quite enough to reconstruct the proof. If you count that way I think you will find the de Bruijn factor closer to 2-4. (Alternatively, it could be an artifact of this being your first proof, I will have a look at the proof more carefully later.)

digama0:
Actually, maybe for this particular proof the de Bruijn factor is just unusually high because that proof is really short and dense. (It reminds me a bit of [Zagier's one-sentence proof that every prime p = 4k+1 is a sum of two squares](https://people.mpim-bonn.mpg.de/zagier/files/doi/10.2307/2323918/fulltext.pdf).)

PR #1525, opened 2020-03-13
Move earlier

benjub:
Continuation of PR #1516 : some reordering, some comment editing (mainly adding cross-references).
Three new theorems: two moved from my mathbox: nffal and equs4v, and 19.8v added for clarity of the development of nearby theorems.


nmegill:
When I look at the "Files changed", set.mm says: "14,008 additions, 14,031 deletions not shown because the diff is too large. Please use a local Git client to view these changes."

benjub:
They are due to the label change cdivides --> cdvds. In this case, this shortens the label and makes it consistent with df-dvds, so there is no trade-off. I documented the change at the top of set.mm.

Can you see it with git?

PR #1535, opened 2020-03-17
added old labels to several theorems in 18.1.7

TjWhale:
I think I have completed 18.1.7 with adding old labels and pairing the theorems up with their new equivalents.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -459,7 +459,7 @@
 "ablocom" is used by "ablo32".
 "ablocom" is used by "ablodiv32".
 "ablocom" is used by "ablomuldiv".
-"ablocom" is used by "ghablo".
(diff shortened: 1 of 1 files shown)


benjub:
To catch these problems before PR, I advise you simply copy and paste the two lines at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md (the ones following the phrase "for ease of copypasting"), possibly changing the directory of the metamath program. Before that, update to the most recent metamath version.

TjWhale:
Thanks for the advice. I did run those lines but the old version of metamath tripped me up, I have updated now. :)

nmegill:
Thanks for your analysis.

PR #1536, opened 2020-03-17
spolu-mathbox: remove extraneous hyps and fix comments

spolu:
As per email discussion with Gérard Lang


benjub:
Does the following work:

To catch these problems before PR, I advise you simply copy and paste the two lines at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md (the ones following the phrase "for ease of copypasting"), possibly changing the directory of the metamath program. Before that, update to the most recent metamath version.

benjub:
By the way: when you mention a theorem in a comment, you can prepend a tilde, as in " ~ ax-1 " to create a link in the webpage (spaces after the tilde and after the label).

spolu:
(sorry for the premature submission still broken. will fix and follow-up)

spolu:
@benjub thanks for the `~` tip!

nmegill:
An error was introduced in set.mm that prevents the Travis check from passing.
```
MM> verify proof *
?Error on line 641003 of file "set.mm" at statement 175548, label
"imo72b2lem0", type "$p":
      eqeltrrd leeq2d leeqd 0le2 absmulrposd 2pos wwlemuld absmuld ) ACENZDFNZU
                      ^^^^^
This token is not the label of an assertion or optional hypothesis.

**Comment by spolu** [comment]
@benjub I generally run only:
```
./metamath 'read set.mm' 'write source set.mm /rewrap' erase 'read set.mm' 'save proof */compressed/fast' 'verify proof *' 'write source set.mm' quit
```
as `'verify markup */file_skip/top_date_skip'` crashes on my machine with an illegal hardware instruction.

spolu:
@nmegill should be fixed and ready (sorry for the premature pull request)

PR #1528, opened 2020-03-13
Move earlier

benjub:
Continuation of #1525 .  I mainly add links/precisions to existing comments. @nmegill : I hope you're ok with this style (see also commit message)?  I plan to continue in the same way.

Can I move from my mathbox to main: bj-eximal, bj-alimex, bj-ala1, bj-exa1. Indeed, they might be useful lemmas (although I haven't tried a global "minimize" with them), and they depend on very few axioms, so they would help me in my next two tasks: 1) removing some axiom dependencies in 19.xx(v) statements, and (later) 2) implement the new definition df-bj-nf.

Also, can I move from Glauco's mathbox pirp2: this is also Jim's pirp (which I had to duplicate at some point as bj-pirp). As for credits, I would write "contributed by @jkingdon , shortened by @glacode " if you see fit and they agree.

PR #1537, opened 2020-03-18
Move earlier

benjub:
Commit messages say it all. As agreed in PR #1528 

@nmegill : in the fifth commit (see commit message), I manually shortened the proof of ax5e because "minimize_with" did not do it. So maybe there are other minimizations that are possible but are a bit too complex for "minimize_with"...

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18692,7 +18692,6 @@ Proof modification of "bj-abtru" is discouraged (29 steps).
 Proof modification of "bj-aecomsv" is discouraged (16 steps).
 Proof modification of "bj-aev" is discouraged (39 steps).
 Proof modification of "bj-aevlem1v" is discouraged (42 steps).
-Proof modification of "bj-ala1" is discouraged (9 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
> Commit messages say it all. As agreed in PR #1528
> 
> @nmegill : in the fifth commit (see commit message), I manually shortened the proof of ax5e because "minimize_with" did not do it. So maybe there are other minimizations that are possible but are a bit too complex for "minimize_with"...

I didn't analyze it in detail, but your new proof of ax5e uses both eximal and mpbir, which weren't in the old proof.  "minimize_with" tries only one statement at a time, and if a combination of two statements are needed, it can't do that.

nmegill:
> So maybe there are other minimizations that are possible but are a bit too complex for "minimize_with"...

Yes, there are probably many such cases.  Clever humans can very often do better than 'minimize_with' and won't become obsolete for quite a while.  There are nearly 2000 'Proof shortened by...' attributions, and all were done by humans after minimize was run - unless they cheated and took credit for the minimize run. :)

Issue #1507, opened 2020-03-04
Categories of Rngs and Rings


avekens:
With PR #1506 I proved that the set of unital ring homomorphisms is a "subcategory" of the category of non-unital rings (see ~rhmsubc) and that the restriction of the category of non-unital rings to the set of unital ring homomorphisms is a category (see ~rhmsubccat):

    rngcrescrhm.u $e |- ( ph -> U e. V ) $.
    rngcrescrhm.c $e |- C = ( RngCat ` U ) $.
    rngcrescrhm.r $e |- ( ph -> R = ( Ring i^i U ) ) $.
    rngcrescrhm.h $e |- H = ( RingHom |` ( R X. R ) ) $.

    rhmsubc $p |- ( ph -> H e. ( Subcat ` ( RngCat ` U ) ) ) $=
    rhmsubccat $p |- ( ph -> ( ( RngCat ` U ) |`cat H ) e. Cat ) $=

I also wanted to show

   rngcresringcat $p |- ( ph -> ( C |`cat H ) = ( RingCat ` U ) )

but this is not possible because the compositions are different. Therefore, I think my definions of the Categories of Rngs and Rings are not adequate, i.e. too restrictive, e.g.:

    df-rngc $a |- RngCat = ( u e. _V |-> [_ ( u i^i Rng ) / b ]_
              { <. ( Base ` ndx ) , b >. ,
                <. ( Hom ` ndx ) , ( x e. b , y e. b |-> ( x RngHomo y ) ) >. ,
                <. ( comp ` ndx ) , ( v e. ( b X. b ) , z e. b |->
                                     ( g e. ( ( 2nd ` v ) RngHomo z ) ,
                                       f e. ( ( 1st ` v ) RngHomo ( 2nd ` v ) )
                                       |-> ( g o. f ) ) ) >. } ) $.

I plan to replace the comp slot by the composition used for the category of sets (see ~df-setc). 
Before that, I want to provide a definition for such a composition:

df-mapco $a |- mapCo = ( u e. _V |-> ( v e. ( u X. u ) , z e. u 
          |-> ( g e. ( z ^m ( 2nd ` v ) ) , f e. ( ( 2nd ` v ) ^m ( 1st ` v ) ) |-> ( g o. f ) ) ) ) $.

With this definition, I plan to change the definition of categories of Rngs (and Rings):

     df-rngc $a |- RngCat = ( u e. _V |-> [_ ( u i^i Rng ) / b ]_
[...]

PR #1539, opened 2020-03-21
Categories of ext. structures, non-unital rings and unital rings

avekens:
See also issue #1507.

AV's mathbox:
* the category of sets as extensible structures
* the category of non-unital rings revised (previoud definitions and theorems marked with OLD)
* the category of unital rings revised (previoud definitions and theorems marked with OLD)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4290,6 +4290,7 @@
 "con5" is used by "con5i".
 "con5i" is used by "vk15.4j".
 "con5i" is used by "vk15.4jVD".
+"crhmsubcOLD" is used by "cringcatOLD".
 "csbabgOLD" is used by "csbfv12gALTOLD".
 "csbabgOLD" is used by "csbfv12gALTVD".
 "csbabgOLD" is used by "csbingVD".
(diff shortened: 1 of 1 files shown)


avekens:
Why is the travis check not performed?

arpie-steele:
> Why is the travis check not performed?

Travis is under maintenance : https://travis-ci.org/metamath/set.mm

avekens:
All Travis jobs are finished successfully, but Travis seems to be hanging somehow...

david-a-wheeler:
I cannot be certain, but it looks to me that the Travis job completed, but the message about the completion got dropped. So the system is waiting for the completion that has already occurred. The obvious solution is to restart the job, but I can't see a way to restart a job that Travis believes has completed. The whole thing is in a funky state.

I recommend that you make a completely innocuous change, commit that, and push. Then there will be an actual change, which will trigger the whole thing again. Hopefully that will unstick everything. If it doesn't, then we'll have to start figuring things out for real.

avekens:
OK, I added another section header comment (which I've planned anyway), so let's wait and see...

david-a-wheeler:
That seems to have worked! Hopefully we won't have to do that often.

PR #1546, opened 2020-03-24
Example

benjub:
Add to my mathbox a revision of @spolu 's example. Two changes:
- it is on NN0 rather than on NN
- the RHS is 4^n + 2 instead of 4^n + 5 in order to have smaller (and in particular, only single-digit) numbers in the base case and induction step.

What do @nmegill and @spolu think about moving it to the main part in the Example section 17.1.5 "Other examples", as well as @spolu's two deduction forms which are used in his proof ?


spolu:
(Bullish as I think it covers the kind of exercise that are not well covered my Metamath or any other formal math database. I'll gladly remove it from my mathbox if that happens)

nmegill:
Yes, you can go ahead and do the moves you propose.

PR #1549, opened 2020-03-27
Move OLD versions (of OpenAI shortened proofs) before current

spolu:
Also applies metamath-exe rewrap.

See: https://github.com/metamath/set.mm/pull/1547


spolu:
(feel free to close if not needed)

PR #1551, opened 2020-03-28
Misc

benjub:
The command at CONTRIBUTING.md to regenerate the "discouraged" file added the two lines:
  "2llnma2rN" is used by "cdleme20y".
  New usage of "2llnma2rN" is discouraged (1 uses).
because OpenAI used it. Did it use other discouraged results ? (@nmegill and @spolu)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -192,6 +192,7 @@
 "2llnjN" is used by "2llnm2N".
 "2llnjaN" is used by "2llnjN".
 "2llnm2N" is used by "2llnm3N".
+"2llnma2rN" is used by "cdleme20y".
 "2llnne2N" is used by "2llnneN".
 "2pm13.193" is used by "2sb5nd".
 "2pm13.193" is used by "2sb5ndALT".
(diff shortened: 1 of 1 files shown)


spolu:
Should we remove the new cdleme20y ? Or attempt to amend it manually if we have an equivalent of 2llnma2rN ?

benjub:
I do not know about this particular case. It is strange that the PR with your shortenings did not modify the "discouraged" file... Let's wait for @nmegill's input.

spolu:
I may have missed a required pre processing. My bad.

benjub:
The strange thing is that it passed the Travis test... Not your bad !

nmegill:
Something has gone very wrong with the Travis linkage - it seems it is not checking anything anymore.

Look at the commits summary https://github.com/metamath/set.mm/commits/develop - Mar 26 and earlier, there is a green check mark next to each merged commit, but on Mar 27, the green check has disappeared.  This seems kind of disastrous, especially with the OpenAI stuff still under debugging, and the 'discouraged' file now way out of sync.

Any suggestions?  Should we revert everything back to Mar 26 and start over?  @david-a-wheeler any ideas?

benjub:
Looking at the changes in "discouraged" in this PR, it looks like the only discouraged use is the one I mentionned above, "2llnma2rN" is used by "cdleme20y". And I did run both recommended commands in CONTRIBUTED.md, so it does not look that bad.

nmegill:
A cursory check of the 'discouraged' file does seem to indicate the only problem is

"2llnma2rN" is used by "cdleme20y".
New usage of "2llnma2rN" is discouraged (1 uses).

as @benjub suggested.  Since that's in my mathbox, let me study the shortened proof, and I will make any adjustments needed myself (going back to the *OLD if I decide it should be done).  So let's not revert anything for now.

So I think this is ready to merge.  But I would like to get the Travis problem resolved before we merge any more PRs.

david-a-wheeler:
@nmegill said:

> Something has gone very wrong with the Travis linkage - it seems it is not checking anything anymore.

Something is wrong, that's for sure.

Travis seems to still be doing the checks, you can see them here: https://travis-ci.org/github/metamath/set.mm/builds

But our GitHub page isn't showing them.

Let's track the Travis/Github continuous integration problem as a separate issue, [1555](https://github.com/metamath/set.mm/issues/1555)

nmegill:
Since Travis is working again, could you close and resubmit this PR?  Thanks.

PR #1552, opened 2020-03-28
Frege's hereditary relations

arpie-steele:
New $c hereditary
3 definitions for it.
13 theorems based on those definitions.
Revised frege58b, added frege58bid, removing many dvs.
Added dffrege69 restatement of definition of hereditary to match Frege1879.
7 theorems from Frege1879.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -192,6 +192,7 @@
 "2llnjN" is used by "2llnm2N".
 "2llnjaN" is used by "2llnjN".
 "2llnm2N" is used by "2llnm3N".
+"2llnma2rN" is used by "cdleme20y".
 "2llnne2N" is used by "2llnneN".
 "2pm13.193" is used by "2sb5nd".
 "2pm13.193" is used by "2sb5ndALT".
@@ -995,6 +996,7 @@
 "ax-frege2" is used by "frege3".
 "ax-frege2" is used by "frege39".
 "ax-frege2" is used by "frege4".
+"ax-frege2" is used by "frege73".
 "ax-frege2" is used by "rp-7frege".
 "ax-frege2" is used by "rp-8frege".
 "ax-frege2" is used by "rp-frege24".
@@ -1022,6 +1024,7 @@
 "ax-frege8" is used by "frege62newc".
 "ax-frege8" is used by "frege66b".
 "ax-frege8" is used by "frege66newc".
+"ax-frege8" is used by "frege74".
 "ax-frege8" is used by "frege9".
 "ax-hcompl" is used by "chscllem2".
 "ax-hcompl" is used by "hhcms".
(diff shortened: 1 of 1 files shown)


arpie-steele:
@nmegill @benjub Frege's Definition 69 introduces a two-dimensional notation relating wffs "F(α)" and "f(δ, α)" where δ and α are dummy variables. Here I interpret δ and α as sets, which renders the "F(α)" as "α ∈ A" and "f(δ, α)" as "δ R α" and eliminating dummy variables, we have the df-he as ` ( R |`` A ) C_ A ` . While this is compact, it has 3 floating hypotheses when we want only 2, causing me to introduce ` R hereditary A ` .

Now since there is no standard notation (to my limited experience) I'm torn on the question of notation. I have used "R he A" as the display form and `R hereditary A ` as the source form. We almost certainly don't want to code the latex for Frege's 2-dimensional notation with dummy variables.

benjub:
I would use, as often as possible, the same for the "source form" and the "display form", so "R hereditary A" for both.  The standard meaning of "hereditary" is different, though, so at least put a caveat in the comment of df-he ?
I am not certain of the usefulness of this definition, but I do not know which theorems you plan to prove, so I cannot say for sure.

nmegill:
Regarding "R hereditary A" I agree with @benjub 's comments.

Also, regarding the definition form
```
df-he $a |- ( R hereditary A <-> ( R " A ) C_ A ) $.
```
in general we prefer to define new relations as constants e.g.
```
df-dom $a |- ~<_ = { <. x , y >. | E. f f : x -1-1-> y } $.
```
and yours would be
```
df-he $a |- hereditary = { <. r , a >. | ( r " a ) C_ a ) } $.
```
The reason for this is that we have a large collection of library theorems about binary relations, and it saves having to prove separately things like equality theorems, not-free-in theorems, etc.  In addition, it allows the symbol "hereditary" to be manipulated as a stand-alone class for various purposes.

The main exception to this is if one or both arguments could be proper classes e.g.
```
df-ss $a |- ( A C_ B <-> ( A i^i B ) = A ) $.
```
but this becomes rare after early set theory.  As for your application, I don't know whether you will be requiring proper classes as arguments; if so then your definition is acceptable as is.

(I am going to merge this anyway, and you can make any adjustments in your next PR.)

nmegill:
Another minor thing, above your htmldef section, the comment should be
```
/* Mathbox of Richard Penner */
```
rather than
```
/* Richard Penner's Sandbox */
```
This is for consistency to assist searching (see the first comment containing "Mathbox of" in the htmldefs).

Regarding my previous comment, you'll also need $d r a $. if you use the alternate definition I suggested.

A general remark about comments:  You have several comments with "PLEASE PUT DESCRIPTION HERE." followed by a blank line, that should be deleted.  Also, the "(Contributed by..)" shouldn't be in a separate paragraph i.e. shouldn't be preceded by a blank line.  In general, our stylistic practice has been to avoid paragraph breaks unless they are important to delineate things like auxiliary notes.  Otherwise it tends to put a lot of excessive blank space on things like the Theorem List, showing less information on the screen and requiring extra scrolling.  For example, normally we would put the 4 paragraphs in ~ frege5 into a single paragraph.  OTOH the ~ rp-frege3g uses a 2nd paragraph to introduce a side note, which is fine as is.  While this is a just general guideline and not a requirement, we like to have a consistent style, and you may want to look at how others have formatted their comments.

PR #1553, opened 2020-03-28
Break long line resulting from added OLD to label.

arpie-steele:


arpie-steele:
Merging either this or #1552 should fix the OpenAI long line and need to rebuild discouraged for set.mm.

@spolu in the CONTRIBUTING.md there are two long metamath command lines given as examples to a) re-indent statements, verify all proofs and markup and complain of long lines (it however does not break long lines in comments or math symbols which must be fixed manually) and b) rebuild the discouraged file based on set.mm statements where either future use or proof modification is discouraged. These two changes manually fixes on of the long lines in the math symbols and automatically rebuild the discouraged file. Since the choice of where to break the math symbols and how to indent are manual, this type of change is more likely to lead to merge conflicts so I put it in early even though as it turns out the automatic reporting of build status was broken ( Issue #1555 ) but the build status tested OK in https://travis-ci.org/github/metamath/set.mm/builds/668072840

spolu:
@arpie-steele duly noted. I missed a few of these commands because my metamath-exe crashes on them. I will update it and fix the issue to make sure our future contributions are aligned with the CONTRIBUTING standards.

Sorry for the mishap!

PR #1554, opened 2020-03-28
Escape question mark in egrep pattern.

arpie-steele:
Question mark is a regular expression metacharacter and the MacOS
egrep was complaining about it being used as a literal character.

@david-a-wheeler I think this is not controversial and may prevent problems if the unix platform updates the behavior of egrep to be more like my MacOS.

### DIFF
--- a/scripts/verify
+++ b/scripts/verify
@@ -33,7 +33,7 @@ COMMANDS
 run_verify | tee mm$$.log
 
 # Get status, which is false (nonzero) if an error or warning was found.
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Looks good to me!

nmegill:
Since Travis is working again, could you close and resubmit this PR?  Thanks.  (Unless you know a way to get Travis activated on this PR.)

nmegill:
Since this doesn't involve any database checks, Travis is not relevant, so I'm merging anyway.

PR #1557, opened 2020-03-28
Add clarifications on ALT and OLD statements

david-a-wheeler:
This modifies the conventions text to add missing information
about ALT and OLD statements, based on interactions with the
OpenAI group.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -192,6 +192,7 @@
 "2llnjN" is used by "2llnm2N".
 "2llnjaN" is used by "2llnjN".
 "2llnm2N" is used by "2llnm3N".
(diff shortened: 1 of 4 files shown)


david-a-wheeler:
This fails because of a pre-existing problem in develop (line too long).

Once those other problems are addressed, I'll merge those fixes into this branch to fix it.

nmegill:
The Travis check is working again, with a failure likely due to a previous unchecked PR.  Could you fix it?  Thanks. -Norm

david-a-wheeler:
This *should* be fine, I merged from the new tip of "develop". Let's see what the Travis test (which is now working!) says...

PR #1561, opened 2020-03-30
OpenAI prover's shortening of supmax, dprdwd, unitg, bnj538

spolu:
A few more manually curated shorter proofs found by OpenAI.

Note that supmax is not only shorter but it also removes the need for supmaxlem entirely, so I renamed supmaxlem as supmaxlemOLD since it's not used anywhere. Not sure if it's the right strategy, maybe it should not be renamed but only marked as discouraged?

(as before, no impact on axiom references)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2271,7 +2271,7 @@
 "bnj523" is used by "bnj934".
 "bnj525" is used by "bnj1040".
 "bnj525" is used by "bnj523".
-"bnj525" is used by "bnj538".
+"bnj525" is used by "bnj538OLD".
 "bnj525" is used by "bnj539".
 "bnj525" is used by "bnj540".
 "bnj525" is used by "bnj91".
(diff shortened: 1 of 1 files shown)


nmegill:
> Note that supmax is not only shorter but it also removes the need for supmaxlem entirely, so I renamed supmaxlem as supmaxlemOLD since it's not used anywhere. Not sure if it's the right strategy, maybe it should not be renamed but only marked as discouraged?

This is a tough one.  The comment for supmaxlemOLD says "This allows  sup ( A , B , R )  to be used in some  situations without the completeness axiom" suggesting it might be useful for other situations even though it currently isn't.  So maybe it should be renamed back to supmaxlem (or a more meaningful name), and shouldn't be marked discouraged since the comment implies it could have other uses.

Most lemmas have comments like "Lemma for ~ xyz" implying that they are specialized for that purpose, and those can almost always be changed to *OLD like you did.

I'll accept this as-is for now, you needn't do anything else.  I put a note in my todo list to think about whether it should be left as *OLD or renamed to something else.

spolu:
Thanks!

PR #1562, opened 2020-03-30
Misc

benjub:
Shortening ~ pm4.81 and some comment edits, adding cross links between several inference and closed forms (including ~ dummylink and ~ idi, and also contrapositions and "modi tollentes", if I remember my Latin).

As discussed on the discussion group with @spolu, I had a look at the proof of ~ midexlem by @tirix. I had selected it because it uses a lot of chained ~ ad?ant(l|r)* and ~ simp(l|r)* . I noticed that the proof could be shortened by using some ~ simprrd which actually was already in a mathbox, @glacode's. So I moved both ~ simprrd and ~ simplld to main and it automatically minimized 34 proofs.  I accidentally noticed ~ ne0ii is his mathbox, which looked promising, and indeed it minimized 58 proofs. Other useful lemmas are probably his theorems of the form ~ ad?ant(l|r)* and ~ simp(l|r)* which I could try next ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18909,7 +18909,7 @@ Proof modification of "bj-axrep2" is discouraged (181 steps).
 Proof modification of "bj-axrep3" is discouraged (122 steps).
 Proof modification of "bj-axrep4" is discouraged (130 steps).
 Proof modification of "bj-axrep5" is discouraged (125 steps).
-Proof modification of "bj-axsep" is discouraged (135 steps).
+Proof modification of "bj-axsep" is discouraged (132 steps).
 Proof modification of "bj-axsep2" is discouraged (58 steps).
 Proof modification of "bj-axtd" is discouraged (26 steps).
 Proof modification of "bj-bibibi" is discouraged (34 steps).
@@ -18986,9 +18986,11 @@ Proof modification of "bj-equs5v" is discouraged (41 steps).
 Proof modification of "bj-equsal" is discouraged (31 steps).
 Proof modification of "bj-equsalhv" is discouraged (10 steps).
 Proof modification of "bj-equsalv" is discouraged (39 steps).
+Proof modification of "bj-equsalvv" is discouraged (38 steps).
 Proof modification of "bj-equsb1v" is discouraged (17 steps).
 Proof modification of "bj-equsexhv" is discouraged (10 steps).
 Proof modification of "bj-equsexv" is discouraged (38 steps).
+Proof modification of "bj-equsexvv" is discouraged (36 steps).
 Proof modification of "bj-eunex" is discouraged (53 steps).
 Proof modification of "bj-exlimmpbi" is discouraged (11 steps).
 Proof modification of "bj-exlimmpbir" is discouraged (11 steps).


nmegill:
While I haven't quite figured out what's going on with midexlem that requires so many ad\*ant\*, in the past I've observed that sequences of ad\*ant\* can make proofs inefficient.  A long time ago I was able to shorten many proofs with sequences of ad\*ant\* by replacing them with syl\*anc and using simp\* to satisfy the antecedents of syl\*anc.  I described this in a note here: http://us.metamath.org/mpeuni/mmnotes.txt - search for "(19-Mar-2012) A new approach to antecedent manipulation".  In particular, look for the divdivdiv example at the end of that note.  I don't know if that method is applicable to midexlem but it might be worth looking at if someone has time.

tirix:
I appreciate the increased scrutiny around my proofs! :-)
Some insights which may help understanding the proof structures:

I've designed the whole chapter "Tarskian Geometry" in (Mario's) deductive form, which I find a very comfortable and effective way of working - this is very near to "natural deduction". I believe this shortens many proofs in general.

The proofs start with a context `ph`, but there are several cases when this context grows, and in those cases I've been using the same "recipes":

- whenever the proof handles different cases. For example, if there are two cases ` ps ` and not ` ps `, I use ~ [pm2.61dan](http://us.metamath.org/mpeuni/pm2.61dan.html), and the context grows into ` ph /\ ps ` and ` ph /\ -. ps `.
- whenever the proof needs intermediate constructions, which happens often in elementary geometry (mostly intermediate points). For example, if the proof constructs an ` x ` such that `ps ` ("let x such that ps"), I use ~ [r19.29a](http://us.metamath.org/mpeuni/r19.29a.html), and the context grows into ` ( ph /\ x e. A ) /\ ps `
- whenever the proof uses contradiction. For example, if the proof states something like " `ps` cannot be, otherwise `ch` would not hold, which contradicts our hypothesis", I would use [pm2.65da](pm2.65da), and the context would grow into ` ( ph /\ ps ) ` until the contradiction is proved.

When several constructions/assumptions are done, the context grows and grows.

In the proof of ~midexlem, for example, if you look at the extreme [step 142](http://us.metamath.org/mpeuni/midexlem.html#142), we are in the case where `A`, `B` and `C` are not colinear, we have constructed `p`, `q`, `r`, `x` and `s`, and we are in a case where `r =/= A`. This is a total of seven local assumptions which have been stacked onto the initial  "context" `ph`.

With this structure, whenever we need to grab a statement from the context,  a variant of ~simpr (simp-?l) is used. Whenever we need to grab some statement from a hypothesis or from a previous proven statement with a shorter context, a variant of ~adantr (ad?antr) is used. That's why these are used very often in the proofs.

There is a trade-off to find between first building the full context, pulling all needed statements at that level, and doing all proofs there, which keeps a longer context, but pulls the statements only once; and doing sub-proofs with the smallest possible context, which keeps a smaller context but requires to pull the statements repeatedly. The proof of ~ midexlem is more in the second flavor. It is sometimes anyway necessary because to instanciate a constructed `x`, you first need to prove its existence, before you can "add it to the context".

I'd be happy to hear if you find ways to shorten proofs like ~midexlem and others! To be honest, I've not been zealously trying to find the shortest possible proofs, I just did the best I could think of, and I've been relying on minimation runs for the last mile.

wlammen:
Looking at the diff, I'd say, the proof of pm.4.81 is longer than the old one.

spolu:
@benjub out of curiosity, how does one run the minimization process?

benjub:
@wlammen You're right, I'll revert to the old proof. The metamath program displays
    ( wn wi pm2.18 ax-1 impbii ) ABZACAADAGEF $.
---------The proof of "pm4.81" (41 bytes) ends above this line.
and the previous proof said "42 bytes", with a line one-character shorter. But the block of capital letters is one block longer, as you noticed.

benjub:
@spolu I use the script "scripts/min.cmd", which contains instructions on how to use it.

nmegill:
@spolu To give a concrete example to supplement what @benjub wrote, let me start with a simple existing proof:
```
MM> prove condan
You will be working on statement (from "SHOW STATEMENT condan"):
2432 condan.1 $e |- ( ( ph /\ -. ps ) -> ch ) $.
2433 condan.2 $e |- ( ( ph /\ -. ps ) -> -. ch ) $.
2434 condan $p |- ( ph -> ps ) $= ... $.
MM-PA> show new_proof
 7     pm2.65da.1=condan.1 $e |- ( ( ph /\ -. ps ) -> ch )
 8     pm2.65da.2=condan.2 $e |- ( ( ph /\ -. ps ) -> -. ch )
 9   notnotrd.1=pm2.65da $p |- ( ph -> -. -. ps )
10 condan=notnotrd     $p |- ( ph -> ps )
```
For the demonstration, first I'll eliminate notnotrd (i.e. replace the reference to notnotrd with its proof):
```
MM-PA> expand notnotrd
Proof of "condan" increased from 35 to 43 bytes after expanding "notnotrd".
MM-PA> show new_proof
10     pm2.65da.1=condan.1 $e |- ( ( ph /\ -. ps ) -> ch )
11     pm2.65da.2=condan.2 $e |- ( ( ph /\ -. ps ) -> -. ch )
12   syl.1=pm2.65da      $p |- ( ph -> -. -. ps )
14   syl.2=notnot2       $p |- ( -. -. ps -> ps )
15 condan=syl          $p |- ( ph -> ps )
```
Now we "minimize", which will scan all earlier statements (the \* argument):
```
MM-PA> minimize_with *
Proof of "condan" decreased from 43 to 35 bytes using "notnotrd".
Scanning backward through statements...
Proof of "condan" decreased from 43 to 35 bytes using "notnotrd".
The forward scan results were used.
MM-PA> show new_proof
 7     pm2.65da.1=condan.1 $e |- ( ( ph /\ -. ps ) -> ch )
 8     pm2.65da.2=condan.2 $e |- ( ( ph /\ -. ps ) -> -. ch )
 9   notnotrd.1=pm2.65da $p |- ( ph -> -. -. ps )
10 condan=notnotrd     $p |- ( ph -> ps )
MM-PA>
```
The "minimize_with" algorithm is conceptually trivial:  it simply tests each statement earlier than condan, one at a time, to see if the hypotheses and conclusion match existing proof steps.  If so, and the proof size became smaller, it replaces those steps with the trial statement.

It scans in both the forward and reverse directions and picks the one with the shortest result to use for the final proof.

The "minimize_with" algorithm tries only one statement at a time.  It cannot combine two statements.  Despite this "brain dead" simplicity, it has still proved very useful to clean up some kinds of sloppily written proofs.  After one becomes familiar with what it can do, it can make proof writing a little easier.  For example, I usually don't bother with perfect matches to the \*bitr\*d (or even remember all of them) but just use inefficient chains of bicomd and bitrd as needed and let "minimize_with" find the best fit.

By default, "minimize_with" will not bring in the use of new axiomatic assertions ($a statements) i.e. ones the proof doesn't already depend on.  We usually run it using "minimize_with \* /allow_new_axioms \* /no_new_axioms_from ax-\*" which will allow the use of new definitions (df-\*) but still forbid the use of new axioms (ax-\*).  "help minimize_with" will show the various options available.

benjub:
Question for @nmegill : I just tried "MM-PA>minimize * " on pm4.81 and I do not understand why it did not minimize the proof by using pm2.24.  It looks like pm2.24 corresponds to the second step of the proof, so it should have been used. Why not ?

nmegill:
@benjub Yes, the compressed size takes precedence, since that is what affects the set.mm file size.

benjub:
On the other hand, when I prove pm4.81 by using pm2.24 and then I type "MM-PA>minimize *", then it does not minimize the proof by using ax-1...

nmegill:
> On the other hand, when I prove pm4.81 by using pm2.24 and then I type "MM-PA>minimize *", then it does not minimize the proof by using ax-1...

The algorithm first sees if the uncompressed proof size decreases.  If so, it checks to see if the compressed proof size decreases or stays the same.  If both conditions don't hold, it will not use the trial statement.

The basic algorithm was written many years ago before compressed proofs existed.  Checking the compressed proof size (after a trial minimization of the uncompressed proof) was added on later.  Ideally it would check just the compressed proof, but that would mean a rewrite of everything.  I decided that rare edge cases like this that would only save a character or two didn't justify a major effort.  Perhaps someday it can be rewritten, but it is low priority.

benjub:
Regarding the proof of midexlem and @nmegill's and @tirix's remarks: it looks like @spolu's OpenAI minimizations did some shortenings by not "carrying" all the contexts for many proof steps but merging them only at the end, which may be similar to Norm's idea of using syl*anc. But I'm saying this from vague memory, and I'm not sure. Actually, in the proof of midexlem, several steps are used multiple times, so shortening is not that easy: I found the one I mentioned above after several attempts.
In some cases, applying the systematic process "from the book of deduction style" can be longer than some slicker proof, and this may be one of the few cases where a shorter proof is actually harder to understand because it is less systematic.

PR #1565, opened 2020-04-01
Rpenner frege comments

arpie-steele:
@nmegill Continuing from your comments in #1552 . At this point in time I think it makes more sense to develop Frege1879 theorems in terms of classes, not sets, where possible, because it would be useful to see how far the logic could be followed before changes must be made.

I have updated the htmldef section comments to comport with other Mathbox examples, greatly reduced the number of paragraphs in the Frege1879 propositions and fixed some label typos.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1021,9 +1021,9 @@
 "ax-frege8" is used by "frege53b".
 "ax-frege8" is used by "frege53c".
 "ax-frege8" is used by "frege62b".
-"ax-frege8" is used by "frege62newc".
+"ax-frege8" is used by "frege62c".
 "ax-frege8" is used by "frege66b".
-"ax-frege8" is used by "frege66newc".
+"ax-frege8" is used by "frege66c".
 "ax-frege8" is used by "frege74".
 "ax-frege8" is used by "frege9".
 "ax-hcompl" is used by "chscllem2".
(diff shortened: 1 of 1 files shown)


benjub:
Side note to @arpie-steele : the statement of rp-frege2iii is identical to that of bj-mp2c, but note that the latter's proof uses no axioms (only the rule of inference). Together with mp2, mp2b, bj-mp2d, they are the four ways to chain two modus ponens. As such, I had proposed @nmegill to move them all at the very beginning of set.mm, even though last time I tried, they didn't minimize many proofs. There is no hurry in making a choice: they are there at the beginning of my mathbox anyway.

arpie-steele:
@benjub While trying to trace usage and describe the theorems, I did take note of how little of ax-2 (actually ax-frege2) remained in the statement of rp-frege2iii. I was close to deciding to deleting it as a waste of space. In another direction I was thinking about building a library of minimal logic to basically capture the content of https://arxiv.org/abs/1606.08092 which if set.mm and iset.mm ever go modular could be a preamble/library.

benjub:
@arpie-steele I built an MM-database for minimal logic and posted it as an attachment on the google discussion group a few years ago. Search "benoit minimal logic" and you'll probably find it. Indeed, I strongly advocated a modular axiomatization for the "big three": minimal, intuitionistic, and classical logics (for propositional calculus). Miraculously, such a piecemeal axiomatization is possible, with additional benefits like the separation theorem (nothing is due to me, but to T. Thacher Robinson). Unfortunately, it was not accepted by @nmegill for set.mm nor by @jkingdon for iset.mm.

jkingdon:
> Unfortunately, it was not accepted by @nmegill for set.mm nor by @jkingdon for iset.mm

I remember Norm making such a statement for set.mm, but I don't remember having a particularly strong opinion about what should be done in iset.mm.

Is what is needed to make it minimal just to omit ax-in2 or are other changes needed?

If so, perhaps delaying the introduction of ax-in2 until a bit later in the propositional logic section would be a good first step? Or is there another approach being considered?

benjub:
Sorry @jkingdon if I misremembered. In any case, you and/or @nmegill gave reasonable arguments to stay as is.  The thread is https://groups.google.com/d/topic/metamath/mYchAks747s/discussion .  I haven't taken time to reread it. The mm file attached in that thread is old and I have a newer version that I'll try and find on my computer and post in the discussion group.

PR #1568, opened 2020-04-04
Add real number intervals to iset.mm

jkingdon:
This the set.mm section "Real number intervals". It needs a lot of intutionizing, but perhaps surprisingly only a bit of that has to do with differences in how order works in the reals (lack of trichotomy and the like).

Probably the larger amount of intuitionizing has to do with evaluating functions outside their domains, which we mostly solve by ensuring that we are within the domain.  For example, in a theorem which refers to `( A (,) B )` we add conditions that `A e. RR*` and `B e. RR*`.

There are three changes to set.mm, which correct comments in which the comment does not match the theorem being proved.


david-a-wheeler:
Should we add some of those conditions in set.mm? I realize it's sometimes convenient to be able to proof statements without the conditions, and there is a diminishing point of return I'm sure, yet I presume that we normally expect these theorems to only be used in certain conditions. Maybe it's time to add some of those conditions in set.mm. I do not expect everything in iset.mm and set.mm to be identical, but maybe this is an example of an unnecessary difference.

jkingdon:
> Should we add some of those conditions in set.mm?

This is discussed at length in the "Junk/undefined results" section of http://us.metamath.org/mpeuni/conventions.html and the section with the same name at http://us.metamath.org/ileuni/conventions.html

There have been arguments over the years that it is kind of ugly, or confusing, or otherwise bad, to rely on the value of functions outside their domains. And there are various arguments on both sides but "proofs get shorter" is generally seen as a pretty strong argument most of the time. :man_shrugging: If we want to revisit this, no doubt we could, but I'm going to treat the current situation, as documented in those two conventions pages, as the way we do things unless something changes.

digama0:
Are you sure that this isn't still true? My gut says that reverse closure theorems for "set-like" functions like `( A (,) B )` are actually true in iset.mm. Because the constructive version of ndmfv is `A e. ( F ` B ) -> B e. dom F` which should be enough to prove eliooxr.

jkingdon:
> Are you sure that this isn't still true? My gut says that reverse closure theorems for "set-like" functions like `( A (,) B )` are actually true in iset.mm. Because the constructive version of ndmfv is `` A e. ( F ` B ) -> B e. dom F `` which should be enough to prove eliooxr.

Ah good catch. The "constructive version of ndmfv" you refer to is http://us.metamath.org/ileuni/relelfvdm.html and I was able to use that to prove eliooxr . But then the minimizer found something for me: http://us.metamath.org/ileuni/elmpt2cl.html proves eliooxr in one step (and doesn't even seem to care whether it is set-like or not).

I did take advanage of `(,)` being set-like (well, of it being a set) in http://us2.metamath.org/ileuni/ioof.html . There might be more chances to do things like this.

I'll send in a pull request re-working some of this section to take advantage of this insight.

PR #1577, opened 2020-04-09
Minimize etransc and its lemmas

david-a-wheeler:
Minimize etransc and its lemmas using scripts/minimize.


david-a-wheeler:
@glacode - we normally do minimize runs every once in a while, but if there's something in this minimization you object to, please let me know!

benjub:
By the way: itgeq2dv is identical to itgeq2d. It's strange that "minimize_with" did not catch it.

david-a-wheeler:
@benjub - I searched the output of the minimization run, neither itgeq2d nor itgeq2dv are mentioned.

benjub:
Yes: itgeq2dv should have been caught by minimize_with to provide a 1-step proof of itgeq2d. Unless you only ran minimize_with on etransc* and not on the other theorems introduced in that PR?

david-a-wheeler:
I *only* ran minimize on theorems named etransc* - not on all theorems introduced in that PR.

nmegill:
@benjub minimize doesn't look at mathboxes unless you use /include_mathboxes.  If you do, the minimization should take place.

benjub:
@nmegill itgeq2d is in the main part. It's just that "minimize" was not run on itgeq2dv

david-a-wheeler:
@benjub  -  I ran `scripts/minimize itgeq2d` ; the proof of "itgeq2d" decreased from 69 to 20 bytes using "itgeq2dv". I can just contribute that minimization as another pull request. Are there other theorems that should be minimized?

david-a-wheeler:
@benjub @nmegill  I ran "minimize" on `itgeq2d`, but the result isn't just *shorter* - it appears to me that `itgeq2d` is a *repeat* of `itgeq2dv`. I could create a pull request of the minimized version, but removing `itgeq2d` entirely (and just using `itgeq2dv` instead) would be even *shorter* :-).

@glacode - can we just replace all uses of `itgeq2d` with `itgeq2dv`? Is there a distinction I'm not noticing, or a reason that you intended for this theorem to have both names? I'm guessing this was accidental.

benjub:
@david-a-wheeler : Yes it is a repeat: that's the meaning of "identical" which I wrote in my first comment of this thread ;-)

@glacode: Minimizing before a PR is not mandatory, but if you have enough computing power, it's a nice plus, which would detect such duplicates.

PR #1583, opened 2020-04-11
Add additional avatars for contributors

david-a-wheeler:
Add the avatar images for contributors Gérard Lang,
Jon Pennant, and Thierry Arnoux.png.
These images are used when generating the Gource visualization
of contributions, and may be useful for other purposes
when representing the work of various contributors.


david-a-wheeler:
We got 3 more avatars!

david-a-wheeler:
This means we'll have a filename with UTF-8 encoding. That's the standard way to encode filenames, and Lang's first name *does* have an accent. This would have been a problem 20 years ago, but I think it should be fine today.

PR #1592, opened 2020-04-14
Misc

benjub:


nmegill:
Note that strings can't contain newlines.  Instead, they must be broken up into strings that span a line and are connected with + e.g.

```
althtmldef "-Set->" as ' <span style="' +
  ' overflow: visible; width: 0px; height: 0px; display: inline-block;' +
  ' ...
```

nmegill:
It seems a little redundant to have both df-map and df-bj-fset.  What is the main reason for this?  

Also, as you can see in the comment of df-map, some authors define it with the arguments reversed.  I picked the order we use mainly because it's the same argument order as in the cardinal exponentiation it can represent, but that's not strictly necessary and doesn't seem to bother authors who reverse the order.  Would  it make sense to use your definition to replace df-map?

benjub:
@nmegill  Indeed it is redundant to have both.  The notation $B^A$ is indeed classical (although I think Tarski used ${}^A B$) and matches with cardinality: card(B^A) = card(B)^card(A). Most people will understand it correctly.

Two advantages I see for ( A -Set-> B ) are that the order or the arguments is not ambiguous and also that we can have as well -Grp-> , -Ring-> , -Top-> , etc.  I'm going to write on the discussion group to have more opinions.

PR #1594, opened 2020-04-15
Add equinumerosity to iset.mm

jkingdon:
Although the constructive literature doesn't talk a lot about equinumerosity, as far as I could tell the set.mm definition, of whether there is a 1-1 and onto mapping, is the appropriate one.

Large parts of this section intuitionize. Other parts, like strict dominance, are omitted.

As for the motivation, I wasn't necessarily thinking more deeply than a few finite set theorems which come up as I work through set.mm sections on integer ranges. If you want to develop cardinality more fully... well go for it because I'm probably not going to look very much at this soon (I did look through the section on cardinality in the HoTT book and wasn't sure how, or whether, it applies to iset.mm).


david-a-wheeler:
That's interesting! I notice that the Stoic logic theorems are valid in intuitionistic logic. That's.. interesting! They also accepted the law of excluded middle, but you can obviously.. exclude that :-).

jkingdon:
> That's interesting! I notice that the Stoic logic theorems are valid in intuitionistic logic. That's.. interesting! They also accepted the law of excluded middle, but you can obviously.. exclude that :-).

Apparently :shrug: . You probably know more than I do about the Stoic-ness of these theorems, but I just pulled them over because.... well that's what we do with anything in set.mm which can be pulled over.

PR #1595, opened 2020-04-16
More df-bj-if and two statements from Frege

arpie-steele:
The Frege xor statements come in Chapter I, Section 7 and have no label. Should they be labeled for the bibliography? If so, how? It wouldn't be too hard to find their page number as I did with all the labeled theorems.

@benjub is free to move the df-bj-if theorems to his mathbox if desired.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1023,7 +1023,7 @@ is also included.
 <A HREF="downloads/metamath.zip">metamath.zip</A> (12 MB)
   <UL>
   <LI><FONT COLOR="#006633"><I>Description:</I></FONT> The
-    <B>metamath program</B> (version 0.182 12-Apr-2019), which is an
+    <B>metamath program</B> (version 0.182 12-Apr-2020), which is an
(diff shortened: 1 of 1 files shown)


benjub:
@arpie-steele Yes, I'll try to use the current lockdown to put df-bj-if in Main. @nmegill , @digama0 and @jkingdon had expressed their preference for bj-dfif5 to be the "official" definition, so I just have to make this change before. The label to be used for the official definition could be df-ifp (the "p" is for "proposition").

jkingdon:
> @arpie-steele Yes, I'll try to use the current lockdown to put df-bj-if in Main. @nmegill , @digama0 and @jkingdon had expressed their preference for bj-dfif5 to be the "official" definition, so I just have to make this change before. The label to be used for the official definition could be df-ifp (the "p" is for "proposition").

That seems like a sensible preference by my past self, so sure.

Not sure I have a big reaction to `if-` in general, but I don't object to it.

PR #1599, opened 2020-04-19
Initial, terminal, zero objects of RngCat and RingCat

avekens:
main set.mm:
* The ring of integers is a nonzero ring.

AV's mathbox:
* Property of a monoid homomorphism, expressed by a magma homomorphism.
* additional theorems for terminal and zero objects
* revision of showing that initial objects are essentially unique
* proof that terminal objects are essentially unique
* additional theorems for zero and nonzero rings
* additional theorems for nonunital rings
* The constant mapping to zero is a nonunital/unital ring homomorphism from any nonunital/unital ring to the zero ring.
* The constant mapping to zero is a nonunital ring homomorphism from the zero ring to any nonunital ring.
* The zero ring is an initial/terminal/zero object in the category of nonunital rings.
* The zero ring is a terminal object in the category of unital rings.
* There is no zero object in the category of unital rings


savask:
@avekens Nice work. I just want to mention that [Lang] also defines initial and terminal objects, but he uses terms "universally attracting" and "universally repelling" (see section 11 in chapter I). He explicitly states the fact that two universal objects have a unique isomorphism between them, so one can use [Lang] as a reference in `initoeu1`, instead of wikipedia (which may change over time).

avekens:
@savask thank you very much for your hints - I will consider them in my next PR.

PR #1604, opened 2020-04-23
move df-ifp to Main

benjub:
@nmegill Before merging, I would like to modify the unicodes as follows.  I am also wondering about changing tokens (and maybe even labels? although this is a bigger change...) to have more consistency.  What do you and @digama0 think ?
```
labels   token   unicode
ifp      if-     if-        CURRENT
if       if      if         CURRENT
if       if      if         PROPOSED
ifc      if_     <U>if</U>  PROPOSED
nf       F/      &#8498;
nfc      F/_     <U>&#8498;</U>
sb       [       [
csb      [_      <B>&#x298B;</B>' /* left sq brack w underbar */
```

@arpie-steele If you find shorter proofs, feel free to shorten mine and add the corresponding credits (once this PR is merged).

After this is merged, I will adapt (non-disruptively) the section on the conditional operator for classes, and maybe also move some staple theorems by @arpie-steele to Main.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14795,7 +14795,6 @@ New usage of "cantnfsOLD" is discouraged (19 uses).
 New usage of "cantnfsucOLD" is discouraged (6 uses).
 New usage of "cantnfval2OLD" is discouraged (0 uses).
 New usage of "cantnfvalOLD" is discouraged (7 uses).
-New usage of "cases2OLD" is discouraged (0 uses).
 New usage of "cayleyhamiltonALT" is discouraged (0 uses).
 New usage of "cba" is discouraged (88 uses).
 New usage of "cbncms" is discouraged (5 uses).
@@ -19149,7 +19148,6 @@ Proof modification of "cantnfsOLD" is discouraged (112 steps).
 Proof modification of "cantnfsucOLD" is discouraged (204 steps).
 Proof modification of "cantnfval2OLD" is discouraged (531 steps).
 Proof modification of "cantnfvalOLD" is discouraged (318 steps).
-Proof modification of "cases2OLD" is discouraged (103 steps).
 Proof modification of "cayleyhamiltonALT" is discouraged (657 steps).
 Proof modification of "cbvabOLD" is discouraged (77 steps).
 Proof modification of "cbvex2OLD" is discouraged (94 steps).


nmegill:
Regarding the PR, are all of the theorems you are moving up actually used by two mathboxes?  I don't see the usefulness of many or most of them (or even the if- itself outside of very specialized applications).  Most seem to be minor variations that can be easily derived as needed.

The Frege stuff, while it is interesting, will never be moved out of a mathbox because it adds additional axioms to set.mm.  Ideally it should be in its own .mm database, but we don't have a convenient mechanism for adding new databases to the web site right now - it involves a new directory with associated infrastructure like iset.mm and nf.mm.  But perhaps we should consider that (although that would mean less motivation for moving up if- in set.mm).



Regarding changing the tokens, perhaps there should be a discussion on the group.

```
labels   token   unicode
ifp      if-     if-        CURRENT
if       if      if         CURRENT
if       if      if         PROPOSED
```
What is the difference between current and proposed above?
```
ifc      if_     <U>if</U>  PROPOSED
```
What are ifc and if_ ?
```
nf       F/      &#8498;
nfc      F/_     <U>&#8498;</U>
```
Isn't this what we have now?  What are you proposing?
```
sb       [       [
csb      [_      <B>&#x298B;</B>' /* left sq brack w underbar */
```
Adding an accent seems it would be more confusing or "busy", not less so.  And wouldn't we want one after the right bracket as well to be consistent?

benjub:
Concerning the proposals for tokens/unicodes: I was not clear at all, sorry. What I meant is that we currently have:
```
labels   token   unicode
nf       F/      &#8498;          - for propositions
nfc      F/_     <U>&#8498;</U>   - for classes
sb       [       [                - for propositions
csb      [_      <B>&#x298B;</B>  - for classes
```
so for the sake of consistency, I propose to also have:
```
if       if      if         PROPOSED - for propositions
ifc      if_     <U>if</U>  PROPOSED - for classes
```
whereas we currently have:
```
ifp      if-     if-        CURRENT - for propositions
if       if      if         CURRENT - for classes
```

benjub:
Concerning the move to Main, I thought this was settled by your post https://groups.google.com/d/msg/metamath/loIGujrD5Sg/82bkMG6oCwAJ (see also that thread and the previous thread https://groups.google.com/d/topic/metamath/zq_oZ15SFc8/discussion).  By the way, you proposed to move it right before the section introducing True/False, which might be a better place than my current proposal (right before the section introduction triple conjunction/disjunction, a bit earlier).  I can change that if you want.

I think not all alternate definitions are used by @arpie-steele, but it would feel strange to avoid some of them: it would look like "omissions".

nmegill:
For the token changes, let's postpone that for now, and maybe open a group discussion later.  I don't think there is any urgency.

> Concerning the move to Main, I thought this was settled by your post https://groups.google.com/d/msg/metamath/loIGujrD5Sg/82bkMG6oCwAJ (see also that thread and the previous thread https://groups.google.com/d/topic/metamath/zq_oZ15SFc8/discussion). By the way, you proposed to move it right before the section introducing True/False, which might be a better place than my current proposal (right before the section introduction triple conjunction/disjunction, a bit earlier). I can change that if you want.

Yes, I would prefer that it be placed above the T/F section.  Could you do that?

> I think not all alternate definitions are used by @arpie-steele, but it would feel strange to avoid some of them: it would look like "omissions".

OK, we can keep the alternate definitions.

While I haven't checked them all, there are other theorems like ifpor that aren't used anywhere.  Maybe I'm wrong, but this seems like a somewhat useless theorem, or if needed (which I doubt would be more than once) has a trivial proof that would be better done in-line.  To me it seems to waste space in set.mm.

I noticed that the analogous anifp is used once in axfrege58a (with a 1-step proof referencing it), and again I can't see that this theorem would ever be used for anything else.  It would be better IMO to simply replace the proof of axfrege58a with the proof of anifp, making anifp unnecessary.

So I would prefer that ifpor and anifp not moved up, and axfreg58a proved directly from dfifp5.  (This can be done with 'expand anifp' in MM-PA for axfrege58a.)

I glanced at the others and I guess they aren't unreasonable for a basic library, so they can stay.

PR #1610, opened 2020-04-25
Stop duplicate checking on Travis

david-a-wheeler:
Travis runs the complete set of checks twice when there
is a pull request. This commit attempts to eliminate the duplicated checks.
A branch must accept this change for the duplication to be eliminated,
since *their* version of .travis.yml is what Travis uses.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,15 @@
 # "bionic" is Ubuntu 18.04 LTS, "xenial" is Ubuntu 16, "trusty" is Ubuntu 14.
 dist: bionic
 
+# Avoid checking twice the pull requests to "develop" or "master".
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
An awkward problem about commits like this is that they're hard to test - the only easy way to check this commit is to actually use it.

But this commit is also trivial to undo, and it *looks* like it does the right thing. When I created this pull request, there was only *one* pending set of checks instead of two.

This should speed development slightly, as well eliminating some energy waste (and that's a good thing).

david-a-wheeler:
Hmm, maybe not a good idea. It looks like they aren't necessarily dups. One appears to be this pr, the other is if merged. If the base is the same it's a dup, but that's harder to check.

PR #1611, opened 2020-04-26
Gource iso dates

david-a-wheeler:


david-a-wheeler:
This modifies how the Gource captions work and adds more captions.

In particular, the file we edit for captions now uses ISO dates. It's too hard to use Unix timestamps directly. Once that was fixed, I added more captions so that future visualizations will have more information.

PR #1612, opened 2020-04-26
Switch to ++ everywhere for word concatenation

david-a-wheeler:
Use ++ everywhere for word concatenation, per a
poll on the Metamath mailing list.

In the ASCII input representation we've been using "concat";
this is clear but long.  Many presentation formats are in use
in the literature; on the mailing list "++" (as two plusses
or the special Unicode symbol) or frown were preferred.
There was no universal consensus, but this result seemed to
be slightly preferred by the group and everyone could live with it.


avekens:
The ++ looks fine, at least in ASCII. Let's wait and see how it looks in HTML...

david-a-wheeler:
@avekens - I didn't change the HTML, that was already displaying concat as "++".

PR #1613, opened 2020-04-26
Add elsymdifxor to David A. Wheeler's mathbox

david-a-wheeler:


benjub:
Thanks for formalizing this connection between symdif and XOR, following https://groups.google.com/d/msg/metamath/UwTUuNPgaB0/PyQnaD6xAgAJ

In the same spirit, can you also add
`|- ( A /_\ B ) = { x | ( x e. A \/_ x e. B ) }`
and
```
althtmldef "/_\" as " &#x25b3; ";
latexdef "/_\" as " \triangle ";
```

nmegill:
> Thanks for formalizing this connection between symdif and XOR, following https://groups.google.com/d/msg/metamath/UwTUuNPgaB0/PyQnaD6xAgAJ
> 
> In the same spirit, can you also add
> `|- ( A /_\ B ) = { x | ( x e. A \/_ x e. B ) }`

Fine with me.  BTW I would prefer to keep the existing definition ~df-symdif if there are any ideas in that direction.

> and
> 
> ```
> althtmldef "/_\" as " &#x25b3; ";
> latexdef "/_\" as " \triangle ";
> ```

Don't forget the htmldef (could be same as althtmldif or you could use http://us2.metamath.org/symbols/bigtriangleup.gif i.e. 
```
htmldef  "/_\" as " <IMG SRC='bigtriangleup.gif' WIDTH=13 HEIGHT=19" +
    " ALT=' /_\' TITLE='/_\'> ";
```
The spaces around "" duplicate the surrounding spaces in the althtmldef.  We always put a leading space in the "ALT=" argument so that copy/paste from screen will result in space-separated tokens.

david-a-wheeler:
Thanks!

I also think the triangle is the right symbol for symdif, but I would like that to be a separate commit on its own, since it's logically a different change.

PR #1614, opened 2020-04-26
Change symbol for symmetric difference to triangle

david-a-wheeler:
Change the symbol for symmetric difference from "(++)"
to the triangle "/_\"  per mailing list discussion.
The triangle seems to be a relatively common symbol for it.


david-a-wheeler:
This implements the symbol change of symmetric difference per mailing list discussion.

nmegill:
There should be a leading space in front of `<IMG SRC=...` (to match the althtmldef spacing).  Although is there a reason @benjub put in the althtmldef spacing?  The original "(++)" had no surrounding space.  In any case, there should be space on both sides or neither side, otherwise the symbol won't be centered.

BTW the surrounding space in the latexdef is ignored by LaTeX, which uses its own spacing rules.  To force a space (rarely needed) we would use "\\ ", but I don't think it is needed here.

benjub:
@nmegill I did put two surrounding spaces (for unicode) since the symbol is infixed, so it will look better.  As for LaTeX, the surrounding spaces will not be rendered of course, but I think they make the TeX code itself more readable (not a big deal).

PR #1616, opened 2020-04-28
Move symmetric difference into main body

david-a-wheeler:
Move the symmetric difference operation on sets
into the main body, along with a few proofs about it.

Symmetric difference is a standard operation on sets.
There's even a Wikipedia article specifically about it
(<https://en.wikipedia.org/wiki/Symmetric_difference>).
There was a long discussion on the Metamath mailing list
on how to represent it typographically, which also suggests
that people care about this operation. Our usual rule moving
things is when more than one mathbox uses it, and now we have
multiple users.  Thus, it makes sense to move it to the main body.


benjub:
Thanks for the move. Why are the syntax and definition preponed outside of the dedicated section ?

Following PR #1613 , could you add
` |- ( A /_\ B ) = { x | ( x e. A \/_ x e. B ) } `
as dfsymdif2 ? (proof using elsymdif(xor) )

david-a-wheeler:
@benjub - I split it up that way because the other similar definitions did the same thing. But i you still prefer the definitions moved down, then sure, I'd be happy to do that.

I'd prefer to make the move commit separate from any additions.

benjub:
Oh, I see.  Actually, I would still prefer to have the definition in the dedicated section, if you have time to do it.

For instance, df-un (3480) is first used for elun (3644) in the dedicated section. I think having the definition preponed this far from its section makes it harder to locate. (But this is for another PR, if any.)

david-a-wheeler:
Okay, I can easily move the definition later to the dedicated section. It'll have to be a little later today.

nmegill:
I agree with @benjub that the syntax and definition should go in the dedicated section (you can do it in another PR).

benjub:
@david-a-wheeler : I'm going to add dfsymdif2

PR #1617, opened 2020-04-28
Reserve a set of constants for ASCII characters

david-a-wheeler:
Reserve a set of constants for ASCII characters
in David A. Wheeler's mathbox.

We may someday want to represent ASCII characters as constants
(e.g., for "Metamath in Metamath"). Mario is already doing
this in an MM0 database.

This commit creates a proposed set of constant names
to "reserve" these names for characters in set.mm.
That way, if we choose to use them, this set of symbols
will be reserved and ready for their use.


nmegill:
We have never had any policy for "reserving" symbols this way, and it seems to be bloating set.mm with unnecessary tokens that will may never be used.  It will add unnecessary scrolling on the mmascii.html page when browsing symbols.  And if we set this precedent, where does it stop?  Do we for example add all unused _<letter> to force them to be non-italic letters, all unused ~<letter> to force them to be script font, etc.?

I think just a mention in the "ASCII representation of symbols" section of the conventions, like we do for our other token conventions, would suffice.  That is what that section is for.  If someone adds a token clashing with the convention, the amount of work to change it to something else if it later clashes is trivial.

So I am uncomfortable with this PR.  Does anyone else have an opinion?

david-a-wheeler:
> I think just a mention in the "ASCII representation of symbols" section of the conventions, like we do for our other token conventions, would suffice. That is what that section is for. If someone adds a token clashing with the convention, the amount of work to change it to something else if it later clashes is trivial.

We could certainly do that, but I think that's second-best.

The *best* way to tell a user to do or not do something is via an automated check. Then you know that it will or won't happen. Done! Sadly, you can't automate many things, and for those you need to document conventions. But convention documentation should be done when you *can't* automate the check, and this is one of the few checks that *can* be automated.

> We have never had any policy for "reserving" symbols this way, and it seems to be bloating set.mm with unnecessary tokens that may never be used. It will add unnecessary scrolling on the mmascii.html page when browsing symbols. And if we set this precedent, where does it stop? Do we for example add all unused _<letter> to force them to be non-italic letters, all unused ~<letter> to force them to be script font, etc.?

I think this is a special case, because there are a *lot*  of characters (98), not just A-Z, and it would be much better if we had a standard convention for them. What's more, that convention should have a short ASCII representation. It's true that we could rename some operations later, but that's a pain to do.

As far as the mmascii.html page goes, it's already long, and it's easy for humans to understand this particular set (in part because they follow a simple pattern). So I don't think that's a real problem.

As far as "bloat" goes, it doesn't add much to the file size. The "main body" addition is just a few lines of $c statements, and while there are a number of typographic lines, in the grand scheme it's still small and it compresses beautifully. A $300 laptop at Walmart has 4GBRAM and 128GB storage, so even a low-end laptop can easily handle orders of magnitude larger set.mm files. The processing time is trivial, especially since most Metamath tools ignore typography commands (which is what most of it is).

So yes, we *could* document this as a convention. But tis one of the few cases where we can do better than that.... so I think we should.

digama0:
I think I've already said that I don't think the entire set should be declared up-front. Yes, the ideal situation would be to detect a token of the form `'a` and give a warning, but I would much rather that be implemented either as a parameterized family or through a conventions note, rather than case splitting on all characters that exist. This *is* a departure from our previous practices of defining things based on need, even if you are only declaring the `$c` part (which already requires a bunch of htmldef boilerplate).

david-a-wheeler:
Okay, I'll have to create that other pull request (on conventions) later. I'll leave this PR request until that happens so I remember to do it.

david-a-wheeler:
See: https://github.com/metamath/set.mm/pull/1622

PR #1624, opened 2020-04-30
add dfsymdif2

benjub:
As per PR #1616

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18438,6 +18438,7 @@ New usage of "supsrlem" is discouraged (1 uses).
 New usage of "syl5eqnerOLD" is discouraged (0 uses).
 New usage of "syl5imp" is discouraged (0 uses).
 New usage of "syl5impVD" is discouraged (0 uses).
+New usage of "symdif1OLD" is discouraged (0 uses).
 New usage of "tb-ax1" is discouraged (3 uses).
 New usage of "tb-ax2" is discouraged (1 uses).
 New usage of "tb-ax3" is discouraged (2 uses).


benjub:
@nmegill Theorem ~symdifV uses a capital letter.  I think the only capital letters allowed in labels (in Main) are OLD and ALT. What label do you propose ? Is there a standard way to denote the universal class _V in a label ? @david-a-wheeler ?

david-a-wheeler:
@benjub - thanks for adding dfsymdif2!

benjub:
@david-a-wheeler Thanks for your work as well ! Do you have a proposal for the label of symdifV ? Is there a convention for denoting _V in a label ?

david-a-wheeler:
We tend to use lowercase 'v' when referring to `_V`, because `_V` is defined in `df-v` and we usually the part after `df-` as the label fragment. For example, `ralv` follows this convention.

We *also* sometimes the suffix "v" in the label of a theorem eliminating a hypothesis such as ` F/ x ph ` in ~ 19.21 via the use of distinct variable conditions combined with ~ nfv . But both uses are uncommon enough that I think the overlap is okay.

So I think switching to lowercase 'v' is fine. I simply moved the existing theorems, I didn't try to rename them.

benjub:
Relabeling done. I was afraid that *v was only to denote dv conditions, but your example shows that it is also used for _V, so all is good.

david-a-wheeler:
@benjub - understood, I had to look myself. The label conventions don't guarantee uniqueness at all anyway. We don't have separators between the labels in a name, and there can be multiple theorems with different hypotheses and the same proof result.

BTW, ~ abeq2 discusses how to convert classes like `A` to set variables like `x`, but it only lists `cp` as an example, and `cp` is complicated. I think we ought to add your `dfsymdif2` as a second example (in addition to `cp`) so that readers will have a simpler example they can jump to.

david-a-wheeler:
By the way, we use 'v' as the label fragment for `_V` convention often, so don't feel bad about it. I found 62 uses via `grep 'v $p .* _V' set.mm | wc -l`. Here are a few examples:

~~~~ 
  ralv $p |- ( A. x e. _V ph <-> A. x ph ) $=
  rexv $p |- ( E. x e. _V ph <-> E. x ph ) $=
  reuv $p |- ( E! x e. _V ph <-> E! x ph ) $=
  rmov $p |- ( E* x e. _V ph <-> E* x ph ) $=
    ssv $p |- A C_ _V $=
    eqv $p |- ( A = _V <-> A. x x e. A ) $=
  unv $p |- ( A u. _V ) = _V $=
  pssv $p |- ( A C. _V <-> -. A = _V ) $=
  pwv $p |- ~P _V = _V $=
    mptv $p |- ( x e. _V |-> B ) = { <. x , y >. | y = B } $=
  trv $p |- Tr _V $=
  intv $p |- |^| _V = (/) $=
  univ $p |- U. _V = _V $=
    opelvv $p |- <. A , B >. e. ( _V X. _V ) $=
    elvv $p |- ( A e. ( _V X. _V ) <-> E. x E. y A = <. x , y >. ) $=
    elvvv $p |- ( A e. ( ( _V X. _V ) X. _V )
  onnev $p |- On =/= _V $=
~~~~

benjub:
I just noticed the older theorems symdif1 and symdif2. I am going to merge them with the current formalism.

benjub:
Incidentally, I did not obsolete symdif2 only because it is used by mbfeqalem.  In my opinion, this probably tells that the latter's proof is too convoluted.  Maybe @digama0 , who proved it, sees a way to streamline its proof.

digama0:
To be sure, there are ways to significantly shorten the proof using symmetric difference. That would require several more theorems about symmetric difference, connecting it to other things, like ` ( A /_\ B ) e. dom mbl -> ( A e. dom mbl <-> B e. dom mbl )`.

nmegill:
@benjub As you and David noted, lower-case "v" is the convention for labels about _V (as well as "dv version" and rarely as "variable" e.g. visset).  It is essentially impossible to have unique abbreviations for everything without making labels very long and harder to remember, which we (or at least I) don't want.  For searching, things can be narrowed down with additional context, such as "show label \*ex\*v" or "search \*v\* '= _V'".

david-a-wheeler:
@digama0 - I don't see any definition of `mbl` so I don't know what you mean by `( A /_\ B ) e. dom mbl -> ( A e. dom mbl <-> B e. dom mbl )`.

Did you mean `( A /_\ B ) e. dom vol -> ( A e. dom vol <-> B e. dom vol )` ?

digama0:
yes

benjub:
That result is interesting in itself. Who volunteers to add it to set.mm ? ;-)
I think symmetric difference is more widely used in some treatments of measure theory, in connection with Dynkin systems.

PR #1639, opened 2020-05-03
Add Archimedean property to iset.mm

jkingdon:
This is a proof of `arch`. In set.mm, this is proved from the least upper bound property. In iset.mm, we cannot have the least upper bound property. Therefore `arch` is instead proved from http://us.metamath.org/ileuni/archnqq.html (which we have had for a while).

The biggest complexities have to do with embedding positive integers (`N.`) into, in order, `Q.`, `P.`, `R.`, and then ` RR `. In particular, `addnqpr1` is the hardest part.

Also includes `indpi` because we use induction to connect `N.` to `NN`.

~~Unrelated, but in this pull request, is a small change to mmil.html adding a few thoughts on exclusive or. If there is anything controversial about this, I can split it into a separate pull request, but I don't think it is especially definitive about whether to use exclusive or more or not, so maybe it is no big deal.~~ This part is no longer in this pull request - moved to #1641


david-a-wheeler:
It would be nice if set.mm and iset.mm had the "same" definition of exclusive-or so that at least *some* proofs about exclusive-or also held in iset.mm. That may be asking too much.

jkingdon:
> It would be nice if set.mm and iset.mm had the "same" definition of exclusive-or so that at least _some_ proofs about exclusive-or also held in iset.mm. That may be asking too much.

Moving this part of the pull request to #1641 , will respond there.

PR #1641, opened 2020-05-03
Add note about exclusive or to mmil.html

jkingdon:
This isn't really intended to encourage or discourage wider use of exclusive or, just to point to how it may have a somewhat different role in iset.mm than set.mm.


jkingdon:
In response to https://github.com/metamath/set.mm/pull/1639#issuecomment-623050170 I've moved that to #1642

Issue #1509, opened 2020-03-04
Prove Morley's Theorem (Metamath 100 #84)


tirix:
There are nice purely geometrical proofs for Morley's theorem that may become possible with the development of elementary geometry from Tarski's axioms.

savask:
As far as I can tell, you've recently added a "congruence of angles" predicate, so it's now possible to formally state Morley's theorem. Maybe it's worth doing so as a comment in set.mm or even put the exact statement here?

I peeked at several available geometric proofs of Morley's theorem and they all used the property that triangle's angles sum to pi. If I recall correctly, angle and distances sums are somewhat tricky in Tarskian geometry (I think they require introducing arithmetic on equivalence classes). Is formalization of the sum of angles property in set.mm still far away?

Also, I can't help noting that you came close to proving SAS-type theorems for triangles (congrats!). Should we now reprove isosceles triangles theorem in the new geometric framework?

tirix:
I've committed [here](https://github.com/metamath/set.mm/commit/bb4c3896a9f075ca7cddc9b694378dabfca876b3) what I think could be a formulation of the statement.

The first thing to prove with congruence of angles is indeed that it is an equivalence relation, and then to define a law for combining (adding) angles. I don't think we need to build a full arithmetic, but it would be nice to show that this is a group law.

Interesting geometric proofs of Morley's Theorem use "adding 60 degree" to angle. It's possible to build that with combining equilateral angles. As you point out, all proofs of use the property that triangle's angles sum to pi: I think it shall be possible to prove that using supplementary angles properties.

I'm not quite there yet, but it's nice to set some exciting targets, and SAS-type theorems are the first step!

savask:
Your statement of Morley's theorem looks nice, although I think it's worth adding something like `ph -> G TarskiGDim>= 2` since the result doesn't seem to be true for a flat triangle.

I looked for Morley's theorem in Coq (they have also formalized Tarskian geometry), but apparently their proof relies on trigonometry. I guess Metamath could be the first system to formalize a purely geometric proof.

tirix:
Hm, ensuring that we are in dimension 2 or more will not guarantee that the given triangle is not flat: I believe the condition to add is that the 3 input points A, B, C are not colinear (which is only possible in dimension 2 or higher)

savask:
[...]

PR #1655, opened 2020-05-17
Constructed outer measures are countably sub-additive.

tirix:
And other changes which were pending in my mathbox.


david-a-wheeler:
Hi! Sorry I didn't comment, I teach graduate classes and I'm in the middle of the end of semester rush.

savask:
@tirix Nice work on outer measures! Judging by the table of contents in Bogachev's book, it is near the end of the chapter on constructions and extensions of measures. Are you looking forward to formalizing the Lebesgue integral next? :-)

Also I wanted to comment on your Morley theorem statement. As discussed in #1509, the statement is not correct for "flat" triangles. Maybe it makes sense to add the requirement that points A, B and C are not collinear? As far as I understand, one could add ```( phi -> L = ( LineG ` G ) )``` and `( phi -> -. ( A e. ( B L C ) \/ B = C ) )` to the statement.

tirix:
@savask Thanks! I'm not quite done yet though but I'd like to continue all the way to Caratheodory's Extension Theorem. Better than formalizing Lebesgue integral, my goal has been to formalize Bochner's integral (which is merely a generalization of Lebesgue integral to any Banach space ;-) ). You can find it in my Mathbox: [df-sitg](http://us.metamath.org/mpeuni/df-sitg.html) and following theorems, and the (commented out) ~ df-itgm. Both goals complement each other, since Caratheodory's extension can provide measures to use in the Bochner (or Lebesgue) integral. Unfortunately I'm stalled on that track, my attention moves easily!

You're of course right about the non-collinear condition for the statement, I need to add it!
BTW, if you're interested, Gérard pointed me to [a recent Arxiv submission](https://arxiv.org/abs/2005.08723) with a new proof of that Theorem!

tirix:
As a note, I've added the condition that A, B, and C are not collinear in Morley's theorem as part of #1658.

savask:
> Better than formalizing Lebesgue integral, my goal has been to formalize Bochner's integral (which is merely a generalization of Lebesgue integral to any Banach space ;-) )

Now that you mentioned it, I remember the idea being discussed in the google group years ago. It's good to know that it wasn't abandoned. Hopefully your work on measures will finally enable us to define area on CC properly, so we can replace df-area.

> BTW, if you're interested, Gérard pointed me to a recent Arxiv submission with a new proof of that Theorem!

It looks really nice, but the proof still involves some angle-arithmetic ("angle chasing" as they call it). I'm not sure how easy these kinds of arguments are in Tarski geometry, but the proof indeed is very short.

PR #1664, opened 2020-05-31
ifp, df-relexpNEW, df-trcl and misc

arpie-steele:
Although all proofs are either proved or commented out, there are some rough spots.

Notably df-relexpNEW extends (and therefore contradicts) the 2015 sandbox df-relexp in order to be more compatible with df-trcl in that both df-relexp and df-trcl take arbitrary sets, not limited to being a pure relation, but the old df-relexp returns only the relation part when raised to the 1st power. This prevents one from writing 
```
( t+ ` R ) = ( ( r e. _V |-> U_ n e. NN ( r ^r n ) ) ` R ) 
```
which would lead to a simpler application. Since there has been no development on df-relexp since 2015 it was a short job to prepare candidate replacements for all downstream usage of df-relexp so if approved the replacement may be easily made. Some of them have shorter hypotheses (see commit eec5110 for details).

Also, at Benoit's request I have replaced all the bj-if prefixes in my sandbox with ifp.


arpie-steele:
@benjub The renaming of the "bj-if" theorems in my sandbox is in 26fe1ae

arpie-steele:
I don't have mmj2 installed, but almost certainly the conflict between the 2015 df-relexp and my df-relexpNEW defining the same symbol triggered that error.

arpie-steele:
Also, regarding symbols, I'm thinking ^r should have althtmldef of `" &uarr;<SUB>&#x1D45F;</SUB> "` , htmldef of `" r "` , and a latexdef of `"\uparrow_r"`

avekens:
The problem is "I-PA-0202 Axiom df-relexpNEW has failed the definitional soundness check. The previous axiom df-relexp uses the symbol being defined in this axiom."

A symbol must not be defined twice. Here, it is ^r which is defined in df-relexpNEW as well as in df-relexp. The symbol used by you must be changed ( e.g ^R ) to pass the Travis check.

arpie-steele:
> A symbol must not be defined twice. Here, it is ^r which is defined in df-relexpNEW as well as in df-relexp. 

I was aware that different definitions would be a problem, mathematically, but when metamath didn't complain thought the potential difficulties with the long dormant operator could be handled by a usage note until a decision on the final fate of ^r was made.

> The symbol used by you must be changed ( e.g ^R ) to pass the Travis check.

I did experiment with just such a placeholder for a new symbol for the new definition. (likewise cRelexp for the class declaration) which had the advantage that they could be replaced inline without need to change other elements or layout of the proofs.

Unfortunately, the Drahflow definition of `t*rec` uses `^r` which means trying to use `^R` as a holding place for the re-definition and updated versions of proofs necessarily breaks those proofs which use both. I'm currently working on moving the new definition and the updated proofs to replace those in Drahflow's Sandbox so that the t*rec usage remains in place. Having the necessary theorems together will facilitate any future decision to move `^r` and/or `t*rec` into the main body of set.mm.

I'm working on almost no sleep so please bear with me a little longer as I double check everything before updating the PR.

nmegill:
The original df-relexp affected only 18 proofs, so I think it is fine to replace it with your apparently more useful version.

PR #1670, opened 2020-06-08
Section "Complexity theory" added in AV's mathbox

avekens:
Main set.mm:
* some theorems of other mathboxes used in AV's mathbox moved to main set.mm

AV's mathbox:
* Auxiliary theorems (esp. for the modulo operation, even/odd integers, natural logarithm, general logarithm, binary logarithm)
* Division of functions: definition and related theorems
* Upper bounds (big-O): definition and related theorems
* Binary length: definition and related theorems
* Digits: definition and related theorems
* Representation of a nonnegative integer as sum of its shifted digits
* Algorithms for the multiplication of nonnegative integers


savask:
Looks cool! Why is it named "complexity theory", though? Do you have some further results in mind?

avekens:
@savask yes, I plan to define a measure to "calculate" the time complexity of the two multiplication algorithms ~ nn0mulfsum and ~ nn0mullong (and maybe later of the Karatsuba multiplication). I do not know, however, if this is possible with Metamath, but I think the results I achived until now are useful anyway.

savask:
@avekens I see, just the title seemed a little bit misleading. When I read "complexity theory", I was expecting to see Turing machines and complexity classes :-)

I don't know if it's possible to meaningfully study complexity of algorithms without fleshing out a relevant computation theory. It is quite interesting, though, if fast multiplication algorithms in Metamath imply shorter proofs of arithmetic identities involving multiplication, for example, how shorter the proof of "123456789*987654321 = 121932631112635269" is if one uses Karatsuba multiplication instead of the traditional algorithm.

digama0:
@savask Faster algorithms definitely imply shorter proofs. Unfortunately, the correct setting for algorithmic complexity compared to proof length uses *nondeterministic* TMs, which are not well studied because they are "unrealistic", except for the complexity class NP.

For multiplication, as far as I know there is no faster algorithm to do multiplication on a NTM than on a TM. Verifying a multiplication is as fast as evaluating the multiplication and checking you got the right answer.

For alternative multiplication methods like Karatsuba, the question is really when they pay off. The metamath cost metric is quite noisy because there are lots of miniature optimizations that appear when a digit happens to be 0 or 1, so perhaps to get a good result you just have to try Karatsuba steps in lots of ways and switch if it looks like it helps on the particular input. But it would be nice to at least have a ballpark figure for where the switch-over part is. For computers it is quite large, something like 400 bits, but without hardware accelerated 64-bit multiplication the naive method is probably at greater disadvantage.

PR #1674, opened 2020-06-08
answering 2 TODO-NMs

wlammen:
If there are no objections, I will remove bi3antOLD and bisymOLD in near future, as their lifespan expired already.
I introduced both theorems in the hope to be useful in shortening theorems like ~ bibi2i , that have a biconditional in antecedent position, or need so throughout a proof.  This turned out to be fruitless, as there was no single application over several years.  So I asked permission to remove these theorems from the database, which was accepted on 27-May-2019 by NM.  In the wake only BJ objected to the complete removal, and saved a copy to his mathbox.  ~ bija is better suited to build up a biconditional, and it is more precise, as its hypotheses are exactly equivalent to the consequence.  By contrast, the theorems in question appear overly complex to me to be of principal value.


nmegill:
I have no objections.

PR #1677, opened 2020-06-13
rm OLD theorems

wlammen:
As announced a couple of days ago I remove biant3OLD and bisymOLD now.  While I was at it, I tried to discard more expired OLD theorems.  This turned out to be a pain.  OLD theorems are heavily used both in the main part, and in mathboxes, and so I was bitten by a lot of cascading effects.  Some forced me to move OLD theorems to user mathboxes (namely AS and SO). I do not like modifying mathboxes of others without their consent, but I do not know how to do better without keeping OLD theorems forever.  Conclusion: There seems to be some headroom for improving on maintainability of set.mm.

I did not finish the job completely (i.e. you will still find expired OLD theorems left), due to limited time 

WL

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1590,7 +1590,6 @@
 "bdopssadj" is used by "adjeq0".
 "bdopssadj" is used by "bdopadj".
 "bdopssadj" is used by "nmopadjlei".
-"bi3antOLD" is used by "bisymOLD".
 "bitr3" is used by "3orbi123VD".
 "bitr3" is used by "csbrngVD".
 "bitr3" is used by "e2ebindALT".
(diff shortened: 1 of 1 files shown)


avekens:
Hi Wolf,
I want to ask you to stop removing OLD theorems for the moment, because I also started it...
I would propose that we wait until our current PRs are accepted (these should not be in conflict with each other), then I will merge my current changes (not in Git yet) and ask to accept them, and afterwards you can continue your work. I started removing only OLD theorems which are not used anymore.

Would this be OK?

Best regards,
Alexander

wlammen:
Hi Alexander,

That's fine. I currently have limited time for Metamath anyway, so go ahead as you like.

Best Regards
Wolf

PR #1680, opened 2020-06-14
Move remaining log proofs from TA to DAW's mathbox

david-a-wheeler:
Move some of DAW's proofs back to DAW's mathbox.


avekens:
Do you really still need ~logccne0OLD? This theorem is already available as ~logccne0 (and I moved it already to the main part) - only the proof is different.

By the way: ~logccne0 (and also ~logeq0im1) belong to section "14.3.4  The natural logarithm on complex numbers". I can/will move them with my next PR.

PR #1686, opened 2020-06-16
Remove duplicate logblogfv and improve elogb

david-a-wheeler:
The proofs logblogfv and elogb duplicate each other; we only need one.

The elogb proof was first (contributed in 2017) had shorter proofs
for the cases e =/= 1 and e =/= 0 and made those proofs reusable.
However, logblogfv had a shorter proof for building on those facts than
the original elogb proof.

This commit removes logblogfv and modifies elogb so it has the
best traits of the two original proofs (it's one step shorter now).
It also combines the credits (since it's now a combined work of DAW and AV).

This commit also rewraps set.mm using "scripts/rewrap".


david-a-wheeler:
@avekens - I combined my  ` elogb ` and your ` logblogfv ` into one proof to rule them all :-). It didn't make sense to have 2 proofs of the same thing with such small differences. The new proof credits both of us. I hope that seems reasonable!

avekens:
That's fine, thanks.

PR #1688, opened 2020-06-20
my mathbox: more on normalized propositional expressions

wlammen:


david-a-wheeler:
Some of your proofs like `wl-impchain-com-1.x` appear to really be just heading comments. Perhaps you could turn those into headings with comments instead?

Make sure you have "(New usage is discouraged.)" for examples that aren't to be further used. I did a spot-check and it looks like you did!

wlammen:
The theorem you mentioned is certainly a heading comments.  Including it in a dummy proof makes the comment visible on the web pages as a guidance for the theorems to come.  As far as I know there is no procedure to create a web page with just a comment, so I fell back to this mechanism.  The comment in set.mm is a poor replacement, because it is not exported to the web space (to my knowledge!).  Since this all is contained in my mathbox, I felt free to explore new ways, and to suggest possible solutions for guiding pages.  If this technique finds broader acceptance, one might think about a general and more appropriate replacement for it.  Even better, if you know a good way to achieve the same effect right away, I'd be glad to learn from you.

>> Make sure you have "(New usage is discouraged.)" for examples that aren't to be further used.

It is certainly a bad idea to refer to theorems in other people's mathbox.  For one, this is unjust, because an early mathbox suffers from restricted access to others.  And second, I tend to change the contents of my mathbox without further notice. This will possibly break theorems depending on a previous state.  And I would frown upon a need to repair other mathboxes to account for any (mis-)uses there.  If someone finds a particular theorem appealing, he is allowed to copy it to his mathbox, to create a self-contained environment.

To get back to the tags, they are there for my personal convenience, and I use them without obligation to rules established elsewhere.  For example, many theorems, namely the ones you noticed, are written to contribute to a more structured approach to propositional logic, with normalized expressions, and so on.  The "(New usage is discouraged.)" actually blocks me from nothing, and I deliberately act against their content, if I want to.  That said, I try to apply them with consideration.

OK, I don't know whether this provokes further comments, or entices hints to some written rules of conduct.  Should I have violated one, I did so unsuspectingly, and I promise to fix them soon.

Wolf

david-a-wheeler:
> The theorem you mentioned is certainly a heading comments. Including it in a dummy proof makes the comment visible on the web pages as a guidance for the theorems to come. As far as I know there is no procedure to create a web page with just a comment, so I fell back to this mechanism.

If you want it as an independent web page, then what you're doing it just fine; carry on! We do this in other places, e.g., "conventions". However, if you don't actually want a web page, and just want to group theorems, you can using headings and add text after the heading within the comments after the heading. Your choice.

> It is certainly a bad idea to refer to theorems in other people's mathbox.

I think it depends on the material. We often don't promote theorems to the main body until someone else uses them, so I would hate to *discourage* the use of mathbox material by others :-). For example, we just promoted a  "general logarithm" definition and multiple theorems. I had originally developed 2 possible definitions which I put in my mathbox. Others picked one as the "preferred" definition & built on that work. The result just got promoted into the main body because we had multiple people all working on them within their mathboxes, resulting in multiple users of the work.

Of course, the mathbox material might change, but I think that's the point. Mathbox material tends to be more experimental, allowing experiments to be shared with others.

tirix:
@wlammen As a note, I've been considering that a natural normalisation of expressions was this form :
`
( ( ( ( ( ph /\ ps ) /\ ch ) /\ th ) /\ et ) -> la )
`
Like I explained in [this comment](https://github.com/metamath/set.mm/pull/1562#issuecomment-606395402), I found out that this form "grows naturally" when adding more antecedents. The corresponding theorems like [~ ad6antr](http://us2.metamath.org:88/mpeuni/ad6antr.html) already exist. I saw several others use that form too.

[~ wl-impchain-a1-3](http://us2.metamath.org:88/mpeuni/wl-impchain-a1-3.html) is very different, are you using it as a normalized form in your proofs?

wlammen:
@tirix I think, there are several possible normal forms, that cover all of propositional logic, besides yours and mine, for example conjunctive normal form (https://en.wikipedia.org/wiki/Conjunctive_normal_form). Some are particular useful for automated proving. My choice is adapted to Metamath by requiring -> and -. only, such rooting in its current axiomisation.  This comes with a small price, though: It is not unambiguous, as can be seen by con2i.  Yours need three operators ->, /\ and -. . My choice stacks up consequents, yours antecedents.  I don't see a clear line to favour one over the other.  I think much depends on the intended range of application.

Disclaimer: I do not work as a logic expert, I consider myself an amateur in this field.  I earn my money as a programmer, which, more often than I like, leaves me with little time for contributing to Metamath, or do some background research.  So take any statement issued here with due caution.

>  are you using it as a normalized form in your proofs?

I quite automatically dissect any expression of propositional logic into chains of implication in my brain, and that helps me sort out matters, but I haven't used it outside a narrow scope in my mathbox, as you can see by the listed references.  If I only had time...  This field is developing at a really low pace.  

But while I am at a wish list, these are my favourites:

* Convert metamath.exe into a library of utility functions, so others can build their own applications on top of a well tested API.
* Provide a more systematic approach to propositional logic.
* Provide a means of modularisation of set.mm, much along the lines a compiler assembles units and link it to a final result.  Needs some kind of interface technique developed first.
* Provide a machine ready description of the contents of set.mm in some grammar like fashion.

If only I had time...

digama0:
I have a strong preference for the `(a /\ ...) /\ b -> c` style. The reason is that this best approximates the `a, ... , b |- c` style of natural deduction, where `b` and `c` are the "active components" of the current sequent. Most of the time you will be proving things in a context of assumptions, and so it is important for them to be near the root of the tree because of the way metamath theorem matching works. This is most important for steps in a large proof; for complete statements, where the user will be in a completely different context, you can either use full deduction form like `G -> a, G -> b |- G -> c` at the hypothesis level, or a theorem like `a -> b -> .. -> c` or `a /\ .. /\ b -> c`; it doesn't really matter here what the normal form is because you need a syl*anc like theorem anyway to apply it in the new context.

wlammen:
@digama0 I see your point, converting a rule with hypotheses into a tautology. I deliberately picked primitive expressions as a starting point, not aiming at a full context.

digama0:
For the logic puzzle style propositional theorems that you seem to like working on, I'm not sure there is any point in having a normal form. Perhaps it's still a bit too much for brute force optimal proofs, but that's mostly the way I think about them. For slightly more advanced theorems, I will use deduction form as an organizational device.

You can see my version of the "propositional logic toolbox" as adapted for MM1 at https://github.com/digama0/mm0/blob/master/examples/peano.mm1#L243-L533 . In most cases the proofs are so short (composition of two theorems) that there is no point in putting things in a normal form, but for more complicated theorems like `andi` you will see a little bit of organization into deduction form (using `exp` aka `ex`) before using deduction style theorems like `eor` (or elimination, aka `jao`). Notice also that `eor` is written in iterated implication form because there is a macro of sorts that applies `ax_mp` as necessary so that it can also be used as `jaoi` (and in general I don't usually bother with "inference form" versions of a theorem for this reason). I don't know how helpful this is to you but I thought it might help to see an alternative metamath style propositional logic development organized around deduction form.

wlammen:
I am not sure whether I am criticized for my contributions here.  Just in case, a short reply here:

I am not fond of puzzle style theorems, as much as I am not fond of puzzle style utility libraries in my profession.  I'd rather see a theorem generator at my hand, that drops out an automatically proven theorem in .mm style, whenever I need one on a higher level.  But given the current situation, where it is hard to restructure main itself, it is kind of fun to run through a proof for redundancies, and remove them.  This had in the past some benefits, because basic and important theorems tend to bubble up to the front. And in certain areas, e.g. nfeqf2 or the mo* theorems, the proofs changed totally, for the better, if you ask me.

>I'm not sure there is any point in having a normal form...that there is no point in putting things in a normal form

I am not sure how to interpret this. Should you ever come across a theorem with lots of cases (My wild guess here for extremes: Four colours, Characterization of finite groups), how do you get along without having a normal form at hand?  The scaling necessity is so pressing, I personally see no way around it. Maybe here I am not educated enough to see slick solutions.

The same holds for any automated proving.  Your programs become more efficient, if you can reduce input to normal forms.  Do your MM1 processor has some automation built in?  OK then you're ahead of Metamath I'd say.

>I thought it might help to see an alternative metamath style propositional logic development organized around deduction form

Exactly.  If I was to create propositional logic from scratch, I would have centred theorems around the deduction form. No need for help here.

digama0:
> I am not sure whether I am criticized for my contributions here.

Not at all. I'm merely giving some advice regarding normal forms and algorithmic proof from my personal experience.

> > I'm not sure there is any point in having a normal form...that there is no point in putting things in a normal form
>
> I am not sure how to interpret this.

I mean when the goal is to find the optimal proof for a bunch of basic theorems, there comes a point at which the "standard algorithm" with some normal form becomes less efficient than an outside-the-box solution using all theorems that are available without any particular added structure. That's also what I meant by "puzzle style"; I may have misinterpreted your work but the task of finding the shortest proof of the first 500 set.mm theorems looks much like a puzzle to me. I don't mean puzzle in the sense of a logic puzzle like blue eyed islanders or something; these generally translate to rather useless propositional theorems.

> Should you ever come across a theorem with lots of cases (My wild guess here for extremes: Four colours, Characterization of finite groups), how do you get along without having a normal form at hand? The scaling necessity is so pressing, I personally see no way around it. Maybe here I am not educated enough to see slick solutions.

When the scale increases, the brute force "clever" way tends to lose out to a regimented technique with a well defined normal form. Most of my contributions in set.mm for the first 2 years or so demonstrated this to me; I would rewrite a theorem proven using lots of clever propositional lemmas connected together with no overarching structure and replace it with one using straight deduction form, and it would almost always drop in size by almost 50%, even if the high level proof approach remains the same. It simply scales better. Of course the optimal proof is smaller still, but at the scale of 80-800 line proofs the brute force approach is simply not feasible anymore.

For large computations, however, I find it best if one picks a bespoke normal form. Deduction form is good for general reasoning but for something like enumerating a large set, calculating primes and so on, the specific normal form (and associated set of theorems) should be tailored for the theorem generation algorithm.

> The same holds for any automated proving. Your programs become more efficient, if you can reduce input to normal forms. Do your MM1 processor has some automation built in? OK then you're ahead of Metamath I'd say.

I haven't (yet) written much MM1 automation for working with propositional logic specifically yet, but I've been planning to do so for a while. Ideally you could manipulate the context much as one does in HOL or DTT, where the context consists of named hypotheses you can apply and you can just use imp intro and add and remove things from the context. Until then, I just do the same thing manually using appropriately crafted theorems, which is more or less what I did in metamath as well.

> PS: Of course, an implication only theorem needs no deduction form, since you can add the context to the consequent, if needed.

If the only form of a theorem is `a -> b -> ... -> z`, then you at least need a `sylNc` type theorem to be able to deduce `G -> z` from `G -> a`, `G -> b`, etc. Furthermore in the proof of that theorem you will probably want to first turn it into the form `a /\ b /\ ... -> z` so that you can work on `z` (unless you are going to have lots of theorems like `mpddddd` and `sylddddd` and such).

wlammen:
Your answer settles a couple of questions I had, puzzles so to say.  Thanks for that.

> ... with some normal form becomes less efficient than an outside-the-box solution ...

This all boils down to what goal you have in mind.  The metric changes: simplicity in terms of depending only on a couple of principles, or computationally feasible, or, as it in effect now, short in terms of proof bytes.  The latter hopes for better proof elegance and human readability of a somewhat non-standard notation.

> unless you are going to have lots of theorems like mpddddd and sylddddd and such

Yes, I am, but not outside a computer generated proof.  The idea here is to showcase an "engineered" process, i.e. one that inevitably leads to a result, in moderate time and using current resources like a computer. This is not meant to be an optimal output in terms of a metric like human readability.

digama0:
> > unless you are going to have lots of theorems like mpddddd and sylddddd and such
> 
> Yes, I am, but not outside a computer generated proof. The idea here is to showcase an "engineered" process, i.e. one that inevitably leads to a result, in moderate time and using current resources like a computer. This is not meant to be an optimal output in terms of a metric like human readability.

My contention is that you get better bang for the buck here by only doing this kind of deep iteration for context extraction theorems like adant5lr, because this is a much more limited set than all the theorems you want for working on the right side of the sequent, like syl6 and alrimiv and rexlimivw and pm2.61dan and many others. The "engineered" process knows how to select the n-th item from the sequent by using a combination of simpr and adantr (or more optimized versions using adant5lr et al), and then everything else only needs to support one context variable on the left, which gets substituted for a big conjunction.

wlammen:
I agree.  This is just the second step after taking the first one.  The first reduces to a minimum of principles, which I put some emphasis on.  It is of educational value, so to say, and best left after implications grow beyond, say 4 nodes.  OK, getting back to @tirix question, from the point of usability, 'anding' antecedents is superior to using implication only, in 'real' applications.

tirix:
I'm glad I triggered an interesting discussion! I in no way wanted to criticise anyone!
As far as normal forms are concerned, just to name a few others I noticed:

- I think [Giovanni Mascellani used a similar normal form to the one I've mentioned, but with antecedents grouped to the right](https://groups.google.com/d/msgid/metamath/bcb3fd64-bc52-4892-9777-0fb331c0dbd5%40gmail.com.). 
- Giovanni Mascellani @giomasce also uses conjunctive normal form (CNF) and Skolemization in his automatic prover [mmpp](https://github.com/giomasce/mmpp).

Issue #1671, opened 2020-06-08
dots in labels


avekens:
I saw that we have theorems ($p statements) in main set.mm with dots in their labels (e.g. rtrclreclem.min), coming from Drahflow's mathbox recently. Is this according to the conventions of set.mm? I thought that dots "." should be used only in labels for hypotheses ($e statements).

These labels should be cleaned up, shouldn't they?

digama0:
Well, it is legal, but against set.mm conventions. But it's in a mathbox so I wouldn't do much more than inform @Drahflow of the convention and let them do what they like with that information. We don't usually enforce naming conventions until they are getting merged into the main part.

avekens:
@digama0 maybe you looked at an older version of set.mm, but such theorems were moved to main set.mm on 5-Jun-20! That's why I complained about it.

nmegill:
I believe these were moved to main by @arpie-steele. 

I agree that our informal practice has been to reserve the dot (period) for use in hypotheses.  While there is no specific requirement for this, it is nice to be consistent. 

To keep the names to a reasonable length, I would suggest rtrclreclem.refl be changed to  rtrclreclem1, rtrclreclem.subset changed to rtrclreclem2, etc. as is our custom with lemmas.  Although in a sense we are losing information, the comment of the lemmas already indicate that they are about reflexive properties, etc.  In ~ conventions, we state, "Labels are not a 1-to-1 mapping, because that would create long names that would be difficult to remember and tedious to type. Instead, label names are relatively short while suggesting their purpose."

@arpie-steele or anyone who wants to volunteer can rename these in a future PR.  Since they are already in main and will have the renames documented in the list at the top of set.mm, there is no need to notify Drahflow.

Thanks.

nmegill:
BTW we don't want to make this a formal requirement because the early theorems ~ pm\*.\* and ~ 19.\* would violate it.  Over the years we have slowly been renaming a few of them more meaningfully, like ~ notnot1 instead of ~ pm2.12, but that is hard to do in general (with names that are easy to remember).

digama0:
Don't forget the colorfully-named `0.999...`

arpie-steele:
The naming change was adopted in #1689

david-a-wheeler:
I think we can make an exception for `0.999...`. It's memorable, and since it *ends* in a period, it can't conflict with the "normal" user.

avekens:
Done with  PR #1691.

PR #1691, opened 2020-06-24
Convention for characters to be used for labels

avekens:
Convention for characters to be used for labels added (see also issue #1671).

In addition to the convention about the usage of "." as discussed in #1671, I added that the usage of underscore " _ " is also not recommended. There are currently only four labels (for $p statements) using " _ ": ~r19.29_2a, ~tghilbert1_1, ~tghilbert1_2 and ~log_val - should those be changed?

Additional minor changes (in AV's mathbox):
* Header comment of section "Even and odd numbers" adjusted
* Some theorems about even/odd numbers rearranged


digama0:
Ooh, banning underscore sounds nice, because MM0 uses underscore in place of hyphen and without a rule like this we can get more name collisions.

nmegill:
I agree w/ this convention.  I also agree that the existing labels should be changed.  Their contributors are @tirix and @david-a-wheeler; could they acknowledge their approval?  The non-mathbox ones should have the label changes documented at the top of set.mm of course.
```
MM> sea *_*,mathbox contr/c
9096 r19.29_2a $p "...wo restricted quantifiers (Contributed by Thierry Arn..."
99893 tghilbert1_1 $p "...axiom I.1 for geometry. (Contributed by Thierry A..."
99894 tghilbert1_2 $p "...axiom I.2 for geometry. (Contributed by Thierry A..."
112232 mathbox $p "...n_.) A "mathbox" is a user-contributed section that i..."
147950 df-log_ $a "...gested by Mario Carneiro. (Contributed by David A. Wh..."
```
It also looks like the r19.29_2a comment is missing a period after "quantifiers". :)

This could be enforced with a 'verify markup' warning.  Should we go that far?  Maybe we should if it prevents label collisions in MM0.

david-a-wheeler:
I've been thinking about removing `df-log_`. At the time it wasn't clear if it or logb would be best, but logb turned out to be nicer and is now in the main section. I might try to rescue its one use, because it's an interesting example of currying, but it's not clear it's worth it.

tirix:
The "2" in "r19.29_2a" is because this is a double-quantifier version, but if I sticked it together with the rest of the label, it would have resulted in "r19.292a" which suggests a reference to a non-existing chapter 19.292. Do you have any better proposal?
For double-quantifier versions, sometimes one letter is doubled (like in ~ ralrimivva), and sometimes a "2" is appended (like in ~ cbvral2v). Shall we also have a convention about which one is preferred?

avekens:
@tirix There is already the convention that vv should be used if there are two distinct variables. Since this is the case with r19.29_2a, I think it can be renamed r19.29vva. This would also correspond to ~r19.28v. If you don't disagree, I would change it (also changing tghilbert1_1 -> tghilberti1 and tghilbert1_2 -> tghilberti2).
@david-a-wheeler I would rename df-log_ to df-logbALT, at least temporarily.

tirix:
@avakens I'm fine with r19.29vva!

I think this "suffix doubling" would be worth mentioning [on our conventions page](http://us2.metamath.org:88/mpeuni/conventions.html) (chapter "Label Naming Conventions", section "Suffixes"). Actually it would be nice if there was a full list of suffixes and the order in which to expect them...

PR #1692, opened 2020-06-25
"_" in labels replaced

avekens:
See discussion in PR #1691.

As suggested by @tirix, I've enhanced the conventions, especially I provided a list of common suffixes. Finally, I added the 19.x series of theorems (Margaris) as exceptions for using a period in labels.


tirix:
Excellent, thank you very much Alexander! @avekens

PR #1696, opened 2020-06-26
RP Mathbox Shorten proofs

arpie-steele:
@nmegill Should I place (Proof shortened by RP, XX-Jun-2020.) on these 5 proofs in my mathbox or is it OK to leave them as is?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16552,6 +16552,7 @@ New usage of "idunop" is discouraged (1 uses).
 New usage of "ifchhv" is discouraged (22 uses).
 New usage of "ifeq123d" is discouraged (3 uses).
 New usage of "ifhvhv0" is discouraged (48 uses).
+New usage of "ifpfalOLD" is discouraged (0 uses).
 New usage of "iidn3" is discouraged (1 uses).
 New usage of "iin1" is discouraged (3 uses).
 New usage of "iin2" is discouraged (0 uses).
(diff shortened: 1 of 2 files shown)


nmegill:
(Sorry for the delay in merging - I somehow overlooked this.)

> Should I place (Proof shortened by RP, XX-Jun-2020.) on these 5 proofs in my mathbox or is it OK to leave them as is?

It is OK to leave them as is.  Ordinarily I, and I think most people, don't add a "Proof shortened by..." tag when shortening their own proofs (although we don't have a formal policy on it).

Also, we don't add "Proof shortened..." when the shortening was done using 'minimize_with', even if it is someone else's proof.

nmegill:
A possible exception I can think of is if the proof shortening of your own proof is significant in terms of demonstrating, say, a novel proof technique, and you want people to notice it.  The "Proof shortened..." date will cause it to be placed at the top of the "Most Recent Proofs" page.  In this case, you would also want to reproduce the older proof with a "OLD" appended to its label so that the reader can compare the old and new proofs.

In general, though, we always welcome shorter proofs whether or not they are marked with "Proof shortened...".  This was discussed here:
https://groups.google.com/d/msg/metamath/CMKRwV0CX3w/0-flqYbyAgAJ
(I see you contributed to that discussion.)

arpie-steele:
Thank you for the clarifications. The proofs where I made large percentage reductions were simply indications I should have been more familiar with the indexed union theorems before starting. I'm glad I got to shrink them before someone else did.

Issue #1682, opened 2020-06-14
Even and odd integers


avekens:
In one of my recent PRs, I needed/provided a lot of theorems about even and odd integers. I realized (again) that even and odd integers are mentioned in set.mm at several places, but in different ways:

Even integers:
* ( N / 2 ) e. ZZ 
* 2 || N 
* E. i e. ZZ N = ( 2 x. i )

Odd integers:
* ( ( N + 1 ) / 2 ) e. ZZ
* ( ( N - 1 ) / 2 ) e. ZZ
* -. ( N / 2 ) e. ZZ 
* 2 || ( N + 1 ) 
* 2 || ( N - 1 ) 
* -. 2 || N 
* E. i e. ZZ N = ( ( 2 x. i ) + 1 )

In all cases, it must be usually assumed that N is at least a number (N e. CC) , because we cannot derive from ( N / 2 ) e. ZZ that N e. ZZ (Gérard made a corresponding remark recently...).

In Februrary, I already provided definitions (as hypothesis/$e statement) for the set of even and odd integers:
* E = { z e. ZZ | E. x e. ZZ z = ( 2 x. x ) }
* O = { z e. ZZ | E. x e. ZZ z = ( ( 2 x. x ) + 1 ) } 

Of cource, any of the equivalent definitions could be usedfor a definition, e.g. E = { z e. ZZ | ( z / 2 ) e. ZZ }.

Since the concepts of even and odd integers are very commen and frequently used, I want to propose to provide explicit definitions `Even` or `Odd` for them. By such definitions, theorems could be stated without choosing a special representation, e.g. ~nneo (currently ( N e. NN -> ( ( N / 2 ) e. NN <-> -. ( ( N + 1 ) / 2 ) e. NN ) ) ) could be written as ( N e. NN -> ( N e. Even <-> -. N e. Odd ) ).
Instead of `Even` or `Odd`, we could choose the symbols `ZZ_even` and `ZZ_odd` or, shorter, `ZZ_e` and `ZZ_o`, using the representation of `ZZ`with corresponding subscripts for typesetting.

What do you think about such definitions? Such definitions and related theorems could be placed in a section "Even and odd integers" after section "5.4.8 Integers (as a subset of complex numbers)".

In a more general definition, we could restrict the range of integers, for example

odd = { <. x , y >. | ( x C_ ZZ /\ y e. x /\ ( ( y - 1 ) / 2 ) e. ZZ ) },

[...]

PR #1699, opened 2020-06-27
Convention for alternative versions (*ALTV) added

avekens:
Main set.mm:
* swrdsymbeq deleted (swrdeq to be used instead)

Conventions:
* Convention for alternative versions (*ALTV) added
* Convention for old versions/proofs (*OLD) and alternative proofs  (*ALT) revised
* additional entries in list of common suffixes

AV's Mathbox:
*   pfxsymbeq deleted (pfxeq to be used instead)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -10,6 +10,7 @@
 "0cnfn" is used by "riesz1".
 "0cnfn" is used by "riesz4".
 "0cnop" is used by "cnlnadjeu".
+"0evenALT" is used by "0noddALT".
 "0hmop" is used by "0leop".
 "0hmop" is used by "0lnop".
 "0hmop" is used by "idleop".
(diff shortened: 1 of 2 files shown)


avekens:
With PR #1698, @nmegill added "(New usage is discouraged.) (Proof modification is discouraged.)" to the ALT theorems in my mathbox. This is according to the discussion in the Google group: https://groups.google.com/d/topic/metamath/NhPM9XNNh1E/discussion . There, I also mentioned that I use ALT not only for alternate proofs, but also for variants of theorems/definitions. We decided (informally) that for such cases a different suffix (like ALTV) should be used. Unfortunately, I did not consider this decision in my recent PRs...

With this PR, I enhanced the conventions accordingly (I hope this is fine for everybody). In a later PR, I will replace ALT by ALTV in the concerned labels in my mathbox, and will remove "(New usage is discouraged.) (Proof modification is discouraged.)" again.

By the way, if the definitons for even/odd numbers and related theorems can be moved to main set.mm (see #1682), I can save myself this work ;-).

PR #1702, opened 2020-06-30
Add actual Schwabhauser 2.5

david-a-wheeler:
Although tgcgrcomlr is true (congruence commutes on both sides),
that is not Schwabhauser theorem 2.5 (congruence commutes on the RHS).

This commit proves Schwabhauser theorem 2.5 (implication that
congruence commutes on the RHS) as it is stated in the book.
This commit also adds two simple related theorems that commute
congruence on the RHS and LHS; congruence is an extremely important
relationship in geometry, so it's convenient to be able
to commute congruence in one step. This commit also fixes
the commment of tgcgrcomlr so that it no longer says it is theorem 2.5.


david-a-wheeler:
@tirix - per my email to you yesterday, this fixes the problem that `tgcgrcomlr` is not Schwabhauser theorem 2.5 (the statement is *true*, it's just not the same claim). I've added a basic proof of the actual theorem 2.5, as well as 2 simple theorems about the the commutativity of congruence. Congruence is incredibly important in geometry, so I think these are simple but worthwhile. Hopefully that fixes things!

tirix:
It definitely helps, thank you very much!

A minor remark: I believe theorem ~ tgcgrcoml should refer to Schwabhäuser's theorem **2.4**. (you have 2.5 there).

Your comments about ~ tgcgrcomimp being actually the theorem in Schwabhäuser's book is also correct, and actually apply to all of our formalization. I've stated the theorems in a form in which I believe they are most easily applicable, and trivially equivalent: I wrote everywhere ( ph -> A ) |- ( ph -> B ) instead of ( A -> B ). This could e.g. be mentioned in the chapter header.

As another note: I tried to stick to the book's variable names, but could not always. For example, in some cases, the book uses the same letter in both capital and lowercase letters to identify two different variable points, so I had to rename one of them. In other cases, the point names would collide with variable names I've used (like I for intervals, P for the set of all points, G for the geometry structure, etc.)

Congratulations to your first contribution in Metamath set.mm's geometry chapter! (the first one of many I hope ;-) )

PR #1704, opened 2020-07-01
Document how to recreate implications from deduction form

david-a-wheeler:
Document conventions and approaches that are important for
(at least) the elementary geometry section. In particular, document
how to easily recreate implications from an assertion in deduction form.
This is especially important for the elementary geometry section,
because its Metamath representation at first glance might look like they
fail to capture the theorems stated in [Schwabhauser].

mmnatded.raw.html: Document in detail how to convert THM
` ( ph -> ps ) |- ( ps -> ch ) ` into
` ( ph -> ( ps -> ch ) ) ` , where ` ps ` and ` ch ` are any proposition.

set.mm: Modify the elementary geometry section to briefly note
key conventions and possible conversions. Otherwise it might
look like the theorems in this section do not properly prove the
theorems stated in [Schwabhauser].

My thanks to Thierry Arnoux for setting me straight on this
(hopefully I've captured the essence of our discussion correctly).

### DIFF
--- a/mmnatded.raw.html
+++ b/mmnatded.raw.html
@@ -313,10 +313,10 @@ Td in deduction form.
 An assertion in deduction form that has a hypothesis
 can be easily converted so that the hypothesis becomes an implication.
 More formally, given an assertion THM in deduction form with
-hypothesis $e ` ( ph -> ps ) ` and a conclusion ` ( ps -> ch ) `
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@tirix - look okay?

@nmegill - I had some questions for @tirix about how he represented Schwabhauser using Metamath. There was an undocumented convention that I think is important to document; this commit tries to do that.

tirix:
Yes, this looks good to me.

PR #1710, opened 2020-07-07
OpenAI: proofs of some unproven assertions of set.mm.

augustepoiroux:
As suggested by @david-a-wheeler , we (OpenAI @spolu ) ran our model on some of the unproven statement in set.mm. This PR proposes a few proofs we found:
- lt2sqtROY by Roy F. Longton
- ntrivcvgcaulem1 @sctfn 
- axpowndNEWlem1 @nmegill 
- bj-inrab3 @benjub 
- sqrltROY by Roy F. Longton
- fltsval1 by @glacode (Glauco Siliprandi)
- gsummptadd @david-a-wheeler 
- rntrclfv by Richard Penner
- zfcndpowNEW @nmegill 

The model also detected that some are already proven under another name
- sqr2gt1lt2ROY -> sqrt2gt1lt2
- relogbcl is set twice
- dfoprab2f -> dfoprab2
- alephprc2 -> alephprc

PR #1618, opened 2020-04-28
Cleanup OLD theorems

avekens:
I have looked for theorems with suffix OLD which are not used by other theorems (by checking the discouraged file). There were 320 of them. Many of them are not "expired" (232) or in mathboxes (28), so I left them untouched. I removed 18 theorems with an exceeded expiry date (tagged by "Obsolete ... as of  .. " with a date before April 2019), and 11 with no expiry date, but with a current version older than 1 year (see lists below). Some theorems (3) I found to be worth to be kept, so I changed their suffix to ALT. Then I found some theorems (19) for which it is not clear for me what to do, so I marked them with TODO-NM in their comments. These therorems are listed in the following (@nmegill could you have a look at them?):
* bi3antOLD
* bisymOLD
* morimOLD
* reusv5OLD
* reusv6OLD
* reusv7OLD
* xpsspwOLD
* riotassuniOLD
* sucdomiOLD
* fisucdomOLD
* noinfepOLD
* mulge0OLD
* uzind3OLD
* uzwoOLD
* binom2aiOLD
* xpnnenOLD
* xpomenOLD
* mayete3iOLD
* axtglowdim2OLD

Removed (expiry date exceeded):
* morimvOLD
* sbc3angOLD
* sbcimdvOLD
* sbcrextOLD
* sbcreugOLD
* sbcne12gOLD
* csbcomgOLD
* sbccsbgOLD
* sbccsb2gOLD
* csbungOLD
* csbuni2OLD
* sbcsngOLD
* csbexOLD
* csbfv12gOLD
* fnprOLD
* csbriotagOLD
* csbovgOLD
* bwthOLD

Removed (no expiry date, but current version older than 1 year):
* disjmoOLD
* disjiunOLD
* soirriOLD
* sotriOLD
* son2lpiOLD
* xpimasnOLD
* disjmoOLD
* dif1enOLD
* dfsup3OLD
* sucprcregOLD
* ledivmulOLD

Suffix changed from OLD to ALT
* r1pwOLD -> r1pwALT
* nmounbseqiOLD -> nmounbseqiALT
* nmobndseqiOLD -> nmobndseqiALT

Additional changes:
* minor adjustments in comments 
* proof of ~wrdeqcats1 shortened

PR #1716, opened 2020-07-10
set.mm - cleanup to delete some OLD and commented-out stuff

nmegill:
This cleans up commented out sections mentioned in PR #1710 and some \*OLDs mentioned in PR #1618.

@tirix - could you decide what to do with ~ axtglowdim2OLD (i.e. delete, or keep/rename, or leave it alone for now)?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1331,7 +1331,6 @@
 "ax-hvmulass" is used by "kbass5".
 "ax-hvmulass" is used by "kbmul".
 "ax-hvmulass" is used by "mayete3i".
-"ax-hvmulass" is used by "mayete3iOLD".
 "ax-hvmulass" is used by "spansncol".
 "ax-hvmulass" is used by "strlem1".
 "ax-hvmulid" is used by "h1datomi".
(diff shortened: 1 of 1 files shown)


tirix:
Hi Norm!

Thanks for pointing to this, this indeed requires some clean-up.

Here is the story behind ~ axtglowdim2OLD:
I originally defined `G e. TarskiG2D` for the Tarski Geometries of dimension 2, but later felt it would be more convenient and powerful to define `G TarskiGDim>= N`, for the "Geometries of dimension greater or equal to N".
As a matter of fact, this is what theorems often require ("at least two different points" for dimension greater or equal to 1, "at least 3 non-colinear points" for dimension greater or equal to 2), so this allows to state theorems in their most general form, rather than only in the plane.
With this definition, it is still possible to define the usual planar geometries by using ` ( G TarskiGDim>= 2 /\ -. G TarskiGDim>= 3 ) `, i.e. a geometry of dimension at least 2, but not 3 or more.
"Dimension at least 2" is what Schwabhaeuser's lower dimension axiom for dimension 2, ~ axtglowdim2, states. I have rewritten it and renamed the old version to ~ axtglowdim2OLD.
There is another axiom, "Dimension at most 2", ~ axtgupdim2, which I have not rewritten and is still written with ` G e. TarskiG2D `.

So I would need to rewrite ~ axtgupdim2 using the new convention, and probably remove ` TarskiG2D ` to the profit of `TarskiGDim>=` only.
Because the OLD versions of `axtglowdim` and `axtgupdim` will rely on the obsolete ` TarskiG2D `, they will not be used/needed anywhere anymore. But instead of completely deleting them, maybe I would move them to my mathbox instead (like a presidential pardon, but for theorems). 

Let me try to do that soon.

tirix:
(@david-a-wheeler, you might find this post interesting, as you were studying the Elementary Geometry section of set.mm - maybe you have an opinion about the fate of the old version of the theorems?)

PR #1721, opened 2020-07-13
nexdv: reduce dependencies on axioms

wlammen:


wlammen:
Travis complains, because curl cannot load a metamath binary.  That is nothing I can do anything about.

nmegill:
I reran Travis and it passed.  No idea what was happening.

PR #1733, opened 2020-07-19
Make minor improvements to elementary geometry comments

david-a-wheeler:
Modify the Tarski elementary geometry section to
fix typos and add a few cross-references. This does *not*
rewrap; there are many other sections that need rewrapping
and I didn't want to mess those up.


nmegill:
> This does _not_
> rewrap; there are many other sections that need rewrapping
> and I didn't want to mess those up.

I rewrapped in the previous PR #1732, so if you rewrap it should affect only your changes.  Note that comments with an \<HTML\> tag are not touched by /rewrap.  See https://github.com/metamath/set.mm/pull/1695#issuecomment-652129129

Also, comments below section headers are also not currently rewrapped whether or not there is an \<HTML\> tag.  It's actually not done because it's somewhat difficult the way the code is currently written, although I may do it someday for comments below headers without an \<HTML\> tag.  (When that happens I'll inspect each comment below headers and add an \<HTML\> tag if it seems like the comment shouldn't be wrapped.)

david-a-wheeler:
Weird, I'm getting a *lot* more changes from a rewrap.

To rewrap I run `scripts/rewrap`. That in turn runs:

~~~~
metamath "read ${source}" 'save proof */c/f'  "write source ${source} /rewrap" quit
~~~~

I'm running Metamath Version 0.182 12-Apr-2020.

Any idea why I'm getting so many more changes from a rewrap?

nmegill:
OK, the changes are the proof rewrapping by 'save proof */c/f' which it looks like I forgot to do after changing some labels.  (My comment was about /rewrap.)  You can go ahead and do the rewrapping.

nmegill:
Actually I'm going to merge this now to prevent possible conflicts w/ AV's PR in the queue, and I'll do the proof rewrapping separately.

PR #1736, opened 2020-07-20
First pull request

ml-2:
- Add a mathbox for myself at a randomly chosen location (I didn't know how to choose, or if it mattered).

- Add csbdif. I put "Contributed by ML" in its comment, but really it's exactly the same proof as csbun, except all the \*un\* theorems are replaced by the equivalent \*dif\* theorems, so maybe it would be better to say "(Contributed by Drahflow, 23-Sep-2015.) [...] (Revised my ML, 14-Jul-2020.)". I didn't use this theorem, but I think it's worth having?

- Add exlimim\* and exellim* theorems. These are all kind of redundant because of the existing exlimd\*, the real reason I added them is that I didn't find out about exlimd\* until after I had already proven exellimddv and used it. But I think they might still have some value, I think exellim is a special case of interest, and exlimim might save some time for any others who, like me, have an implication to get rid of and don't think to search for the conjunction form.

- Add the \*mptsn\* family, dissneq, and the topdifinf\* family.

- I tried rewrapping, but this also changed lines in someone else's mathbox, so I reverted back.

- The contribution dates are the days I actually proved the theorems, is that the correct thing to do or should I date them on the same day as my pull request?

Thanks!


arpie-steele:
> * The contribution dates are the days I actually proved the theorems, is that the correct thing to do or should I date them on the same day as my pull request?
- I'm the third newest mathbox before you, I believe, so I'm not the most experienced contributor. But as far as I can tell, it is absolutely fine to put the actual date (in local time) that you proved the statement on the contributed line.
- I don't think you are obligated to indicate that you copied the skeleton of the proof from another with a Contributed by ... credit. I believe the first proof of the statement (modulo reassignment of variables) deserves the Contributed by credit. In this particular case `csbin` was proved before `csbun` and has a similar structure. Drahflow also gave us theorems on the transitive closure of a relation which only migrated to the main body of set.mm last month.
- Currently there is a new version of the metamath program ( 0.184 ) which more strictly enforces the idea that mathboxes should be independent of each other and therefore order-independent. Cross-mathbox submissions in the future will have to promote theorems into the main body of set.mm.

ml-2:
@arpie-steele 

Thank you. I've upgraded my metamath.exe version to 0.184 now.

nmegill:
> > * The contribution dates are the days I actually proved the theorems, is that the correct thing to do or should I date them on the same day as my pull request?
> 
> * I'm the third newest mathbox before you, I believe, so I'm not the most experienced contributor. But as far as I can tell, it is absolutely fine to put the actual date (in local time) that you proved the statement on the contributed line.

Either the date you actually proved the theorem locally or the date of the PR is fine.  Some people use the date of the PR so that the theorem will show up at the top of the Most Recent Proofs page.  (One contributor used to make it one day into the future to guarantee his theorems would sort above everyone else's, but that is discouraged.)

> * I don't think you are obligated to indicate that you copied the skeleton of the proof from another with a Contributed by ... credit. I believe the first proof of the statement (modulo reassignment of variables) deserves the Contributed by credit. In this particular case `csbin` was proved before `csbun` and has a similar structure. Drahflow also gave us theorems on the transitive closure of a relation which only migrated to the main body of set.mm last month.

Correct.  We encourage borrowing techniques and methods from other proofs, and it is customary not to acknowledge this.  In the worst case, if someone sees an identical skeleton in 2 proofs, the first contributor can easily be determined from the date if there is a dispute of originality (which has never happened).

> * Currently there is a new version of the metamath program ( 0.184 ) which more strictly enforces the idea that mathboxes should be independent of each other and therefore order-independent. Cross-mathbox submissions in the future will have to promote theorems into the main body of set.mm.

One small issue is that some mathboxes have trivial variations of existing theorems with proofs that have only a couple of steps.  While some judgment is required, we typically don't promote those unless they shorten enough other proofs to "pay for themselves" with a net reduction in the byte count of set.mm.  Instead, we would typically eliminate the reference to such a theorem by expanding its 2-step proof (e.g. using the metamath.exe 'expand' command) to avoid the need for promotion.

PR #1738, opened 2020-07-21
move ifpor to Main, and misc

benjub:
* Moved ifpor from my mathbox to Main, since anifp was moved recently, and both statements are dual to each other and their comments cross-link each other.

* Some typos (I hope @avekens is fine with the (alt)htmldef: no spaces around Repart since it is mostly used as a function).

* Several diffs are merely due to rewrap. I used the "canonical" sequence of instructions at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md and metamath version 0.184.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18874,7 +18874,7 @@ Proof modification of "bitr3VD" is discouraged (36 steps).
 Proof modification of "bj-19.21t" is discouraged (39 steps).
 Proof modification of "bj-2stdpc4v" is discouraged (28 steps).
 Proof modification of "bj-abbi" is discouraged (87 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
> * Several diffs are merely due to rewrap. I used the "canonical" sequence of instructions at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md and metamath version 0.184.

This was my fault, sorry.  Your rewrap fixes it, so everything is good.

avekens:
@benjub you are absolutely right - class symbols for functions (like abs, sin, etc) should usually not have leading/trailing blanks in their representation (in contrast to class symbols for operations, like gcd, sadd, eval, etc.). Maybe this should be mentioned in our conventions

PR #1742, opened 2020-07-28
Forward difference and other theorems

sctfn:


savask:
Great work!

Am I right in assuming that `fwddifnp1` can be restated as something like
```|- ( ph -> ( ( ( N + 1 ) _/_\^n F ) ` X ) = ( _/_\ ( N _/_\^n F ) ) ` X )```,
i.e. n-th forward difference is the iteration of the usual forward difference (maybe one can even omit the `X`)? That was mentioned in the comment to `df-fwddifn`, but I couldn't find the relevant statement.

By the way, you probably already know this, but iterated forward difference can be used to solve the "easier Waring problem", i.e. to show that every integer can be written as a sum or difference of a bounded number of k-th powers. It is not a metamath 100 theorem or anything, but I guess it's still a fun application of forward difference worth formalizing.

nmegill:
Looks like you need to regen the 'discouraged' file before I can merge.

sctfn:
Soon as I get home today.

Issue #1741, opened 2020-07-26
Sort mathboxes by user name?


nmegill:
5 years ago or so, I kept the mathboxes sorted by user surname (except for mine, which was at the end).  This ordering hasn't been maintained for a while.  Now that 'verify markup' requires that mathboxes be independent, we can go back to doing that (and also put mine in its proper alphabetic position).  This will make mathbox users easier to lookup up on the theorem list table of contents.  The ordering will be the same as the "Contributor list" at the top of set.mm.

Before I go ahead and do this in a few days, this Issue provides an opportunity to object or offer other comments.

Norm

avekens:
After the mathboxes are resorted now, my mathbox is near the end now. My previous observeration, that some "frege" theorems of RP are used (automatically) during unification (not minimization), has become worse now. Is there a way to exclude theorems of other mathboxes from unification? On the other hand, the chance that my theorems are used by others (in their mathboxes) are decreased by this sorting. So it has some disadvantages if one's surname starts with a letter of the end of the alphabet...

benjub:
I understand @avekens's concerns. Ideally, the order of the mathboxes should be completely irrelevant. How to do it ? Would it be easily doable, for instance, to have a script (e.g. using metamath.exe's "/split" option) to position the wanted mathbox first or last at eh beginning of a working session (during which one can use /allow_mathboxes), see what theorems are used from other mathboxes and duplicate them in the working mathbox, and then put the mathbox back in its place ? This could foster cooperation. Maybe the script wizzard @david-a-wheeler can see a way ?

david-a-wheeler:
I like the title script wizard, do I get a pointy hat?

It would be possible to do that with a script, but I I suspect there are better ways to do that. Are you using MMJ2, or C metamath, or something else?

benjub:
I was not centering on my use case, but was thinking of a way to foster inter-mathbox cooperation. Therefore, I think something suitable for both the metamath program and mmj2 (and others) would be ideal. Here is my initial thought, using @nmegill's /split option. Roughly:
* Right now, add at the end of set.mm the lines 
```
$( The following two lines are here to help the temporary positioning
   of a given mathbox at the last position, using the script xxx. $)
$( Begin $[ set-mbox-last.mm $] $)
$( End $[ set-mbox-last.mm $] $)
```
* The script xxx should do the following:
0) It is called with one argument $NAME, which is the name of the mathbox you want to work on (for me, it would be "bj").
[...]

PR #1743, opened 2020-07-30
set.mm - put mathboxes in alphabetical order by surname

nmegill:
This implements issue #1741.  Note that there are around 100K insertions and 100K deletions.  Until your local repository is updated, I am guessing that git might be unable to merge your local mathbox edits when the entire mathbox was moved somewhere else.  So  manual replacement of mathbox in an editor might be required.  I will leave this PR open for 24 hrs in case there are objections or other comments.


david-a-wheeler:
If you *only* moved files (and didn't edit them), and each other changes are localized, merging other changes *might* out better than you expect (depending on what's being merged). Git is often good at noticing where the "line is now". That's a core git functionality; it was designed to emphasize ease-of-integration. But obviously there are limits to that :-).

PR #1744, opened 2020-08-01
Changed the definition of angle congruence in geometry.

tirix:
Also, added rings of sets, and made progress towards Caratheodory's Extension Theorem.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -990,9 +990,7 @@
 "ax-frege1" is used by "frege3".
 "ax-frege1" is used by "frege36".
 "ax-frege1" is used by "frege5".
-"ax-frege1" is used by "frege55lem1c".
 "ax-frege1" is used by "rp-6frege".
-"ax-frege1" is used by "rp-a1i".
 "ax-frege1" is used by "rp-frege3g".
 "ax-frege1" is used by "rp-simp2-frege".
 "ax-frege2" is used by "frege3".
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
I primarily looked at iscgra. It seems to be fine to me & ready for a merge.

That said, its expression is so different from Schaubhauser's that I really had to look at it multiple times. Don't slavishly follow Schaubhauser, I think that would be a bad idea!  But where definitions ($a) differ, you might want to explain a little more in a comment about what the differences are & why. So I think it'd be really helpful to add comments, especially in definitions, about the differences & why. For example, the original used 4 "there exists"; why the variation, why does that not matter, etc?

tirix:
Indeed, I took some liberties. Like in other areas of set.mm, we could certainly prove the exact statement of Schwabhäuser's 11.2 / 11.3 / 11.4, in order to show that the definition I chose is equivalent.

And explanations are nice, of course, but I'm not always comfortable with being very verbose.

tirix:
For the third congruence theorem, SSS, Schwabhäuser has 3 versions, and I have provided all three, even though the second and third ones follow trivially from the first. Shall we keep them anyway?

david-a-wheeler:
> Indeed, I took some liberties. Like in other areas of set.mm, we could certainly prove the exact statement of Schwabhäuser's 11.2 / 11.3 / 11.4, in order to show that the definition I chose is equivalent.

I think that'd be a good idea in those cases (and several others). Metamath is *excellent* for showing that a proof of an expression follows from definitions, but it's a very valid complaint to say "these definitions are different, how do I know that they express the same thing?". That would completely eliminate that kind of complaint.

> And explanations are nice, of course, but I'm not always comfortable with being very verbose.

It's easier to simplify statements that are there than to come up with the statements in the first place. So where there's a risk that the expression might be unclear or where it's (non-trivially) different from the source, a sentence or two of explanation would be great. I'd be happy to propose edits to such text if that would be helpful.

> For the third congruence theorem, SSS, Schwabhäuser has 3 versions, and I have provided all three, even though the second and third ones follow trivially from the first. Shall we keep them anyway?

Yes, I think we should. SSS is *tremendously* important. Besides, since this formalizes Schwabhäuser, the easiest way to show that all versions are covered is to actually prove them :-).

PR #1745, opened 2020-08-01
Misc

benjub:
Mainly cleaning, see commit messages.

I tried to add consistency and cross-links regarding associated inferences.  I hope the new paragraph in ~conventions is ok (@david-a-wheeler).

@nmegill : I moved to the top the theorems requiring only ax-1 and only ax-2 (right after the derived rules of inference mp2 and mp2b which require no axioms and which I had moved up earlier).  Also, I think that a more logical label for a1ii, which is not the inference associated with a1i, but is "twice a1i", would be 2a1i or a1i2. What do you think ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -10431,10 +10431,6 @@
 "nn0suppOLD" is used by "psrlidmOLD".
 "nn0suppOLD" is used by "psrridmOLD".
 "nnexALT" is used by "qexALT".
-"nnexALT" is used by "rpnnen1lem1".
-"nnexALT" is used by "rpnnen1lem3".
-"nnexALT" is used by "rpnnen1lem4".
-"nnexALT" is used by "rpnnen1lem5".
 "nnexALT" is used by "zexALT".
 "norm-i" is used by "cdj3lem1".
 "norm-i" is used by "hhnv".
@@ -12068,10 +12064,6 @@
 "psubspi2N" is used by "pclfinclN".
 "pwfi2f1oOLD" is used by "pwfi2enOLD".
 "pwsgsumOLD" is used by "frlmgsumOLD".
-"qexALT" is used by "rpnnen1lem1".
-"qexALT" is used by "rpnnen1lem3".
-"qexALT" is used by "rpnnen1lem4".
-"qexALT" is used by "rpnnen1lem5".
 "ralimOLD" is used by "ral2imiOLD".
 "rb-ax1" is used by "rblem1".
 "rb-ax1" is used by "rblem2".
@@ -13954,7 +13946,6 @@ New usage of "5oalem4" is discouraged (1 uses).
 New usage of "5oalem5" is discouraged (1 uses).
 New usage of "5oalem6" is discouraged (1 uses).
 New usage of "5oalem7" is discouraged (1 uses).
-New usage of "a1iiOLD" is discouraged (0 uses).
 New usage of "a2andOLD" is discouraged (0 uses).
 New usage of "abbi1dvOLD" is discouraged (0 uses).
 New usage of "abbiOLD" is discouraged (0 uses).
@@ -17187,7 +17178,6 @@ New usage of "mo3OLD" is discouraged (0 uses).
 New usage of "modidmul0OLD" is discouraged (0 uses).
 New usage of "mopickOLD" is discouraged (0 uses).
 New usage of "morimOLD" is discouraged (0 uses).
-New usage of "mp2OLD" is discouraged (0 uses).
 New usage of "mplbas2OLD" is discouraged (0 uses).
 New usage of "mplbasOLD" is discouraged (1 uses).
 New usage of "mplcoe2OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@nmegill : I suggest keeping pm2.21ddOLD as pm2.21ddALT since it shows an archetypal use of ~mpd. Are you OK ?

benjub:
Relabeling proposals:
bi1 --> biimp (for consistency with biimpi, biimpd; find another label for the biimp currently in a mathbox)
bi3 --> impbi (for consistency with impbii, impbid, impbidd, impbid21d)

nmegill:
> @nmegill : I suggest keeping pm2.21ddOLD as pm2.21ddALT since it shows an archetypal use of ~mpd. Are you OK ?

Yes, go ahead.

pm2.21dd vs. pm2.21ddOLD presents a curious dilemma.  On the one hand, the compressed proof of pm2.21dd is 2 bytes shorter (disregarding labels), so according to our rules it is preferred for set.mm.  On the other hand, pm2.21ddOLD involves 7 earlier theorems to derive from axioms, whereas pm2.21dd requires 39 earlier theorems.  So someone wanting to understand how pm2.21dd follows from the axioms is going to have to work through a lot more intermediate material. :)

nmegill:
> @nmegill : I moved to the top the theorems requiring only ax-1 and only ax-2 (right after the derived rules of inference mp2 and mp2b which require no axioms and which I had moved up earlier). Also, I think that a more logical label for a1ii, which is not the inference associated with a1i, but is "twice a1i", would be 2a1i or a1i2. What do you think ?

I suppose you're right, although it would take me a while to get used to it.  I would pick a1i2 analogous to mp2.  Using 2 as a prefix meaning 2 applications seems to be absent in prop calc, showing up only in pred calc in 2alimi etc.

nmegill:
> @nmegill:
> 
> As for a1ii, I'll follow your preference of a1i2, though I was thinking of 2a1i since I think 2 is used as a label prefix when a statement has two quantifiers (2alimi as you mentioned, and many others), and numeric prefixes are also used in 3syl, 4syl.
> 

I hadn't considered 3syl, 4syl.  Actually, when these were added, syl3 and syl4 were already taken, although I'm not saying that's the reason for the choice (back then it was simply what I pulled out of the air at that moment).  I did 'show label 2\*' and only 4 showed up in prop calc: 2th, 2thd, 2false, 2falsed, so I assumed that a prefix of 2 was not a convention for "two applications" in prop calc.

For the labels \*2 in prop calc, other than the pm\* ones most seem to refer to "second version" or "second argument".  Note that we do have a1dd2 that is almost exactly analogous to a1ii, but I think the "2" there was added because a1dd already existed.

My suggestion was meant to follow existing convention for prop calc, but after a closer look it seems there actually is no consistent one, so I'll retract my suggestion for a1i2.  Maybe we could start a convention where a prefix of 2 in prop calc typically means two applications of a theorem, so a1ii would become 2a1i as you suggest, and we'd change a1dd2 to 2a1dd (the two d's meaning nested twice, not two applications of a1d).  I'm not sure if there are others involving 2 applications that should be renamed.  Of course we have many 3bi\*, 3im\* which means chaining 3 things analogous to 3syl.

To summarize, I now think we should change "a1ii" to "2a1i" and "a1dd2" to "2a1dd".

I'm a little torn about changing "mp2" to "2mp" (although it would be consistent with this scheme) because of variations like "mp2an" - should we also change that to "2mpan"?  What about "mp3an", where the "3" refers to triple conjunction as well as to 3 applications of ax-mp?  We would need further discussion for those.

> What do you think of my above proposal:
> bi1 --> biimp (for consistency with biimpi, biimpd; find another label for the biimp currently in a mathbox)
> bi3 --> impbi (for consistency with impbii, impbid, impbidd, impbid21d)

Both are fine with me; go ahead.

In the mathbox theorem, "imp" means "import".  Maybe change "biimp" there to "impbi", or whatever you like.  (Since this is in AS's mathbox, the rest of us are maintaining it here on.)

> 
> Also, we had seen that id1 would be more consistently labeld as idALT, but there are some links to that webpage, so the change could break some links. Do you see a way of smoothly transitioning ? For instance, having both id1 and idALT for some time, while locating and changing the links ?

Yes, we can have both; they don't take up much space.  It looks like id1 is linked to in mmset.html, mmil.html, iset.mm, and mmmusic.html, and we can change those to idALT.  A potential problem is that id1 is mentioned multiple times in the Metamath book and also in at least one slide presentation.  So I think we need to keep the id1 alias indefinitely or at least until the next printed edition of the book.

benjub:
Ok. Next week, when I'm back to my main computer, I'll do the changes:
* a1ii --> 2a1i
* a1dd2 --> 2a1dd
* bi1 --> biimp
* bi3 --> impbi
* biimp (in mathbox) --> (the label you suggested, impbi, will be taken... Maybe bimp or impb ?)

I think keeping mp2 and friends is fine (and I don't despair of finding some use for bj-mp2c and bj-mp2d one day!).  And we keep in our head the idea of having id1 as an alias for a new idALT, for when the next edition of the Metamath book is in sight !

PR #1719, opened 2020-07-11
`gsummptres` proved using OpenAI's proof assistant, other theorems

tirix:
* lemmas for Lambda-Pi theorem
* clean-up of usused ~ gsummptadd
* first step of cleaning up unused `TarskiG2D` (incl. axtglowdim2OLD and axtgupdim2OLD)

PR #1747, opened 2020-08-02
Prove definition of angle congruence

tirix:
As discussed in #1744, this shows that the definition chosen for angle congruence is equivalent to the one of Schwabhäuser.
Also fixes an issue with ~cgratr where ` I ` was reused.
Also, as discussed in #1716 and #1719, this moves TarskiG2D and related theorems to my mathbox for "archiving".

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4864,6 +4864,7 @@
 "df-topspOLD" is used by "eltopspOLD".
 "df-topspOLD" is used by "istpsOLD".
 "df-topspOLD" is used by "tpsexOLD".
+"df-trkg2d" is used by "istrkg2d".
 "df-tru" is used by "tru".
 "df-unop" is used by "elunop".
 "df-va" is used by "0vfval".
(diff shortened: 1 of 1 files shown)


tirix:
Note: With this PR, I'm also moving ~ simplrd and ~ simprld from GS Mathbox into the main part of set.mm.
As a matter of fact, ~ simplld and ~ simprrd are already there, I believe we shall have either none or all the 4 combinations.

avekens:
Hi Thierry, 
as I saw ~ topnfbey I was really shocked at first, destroying my faith in the correctness of Metamath and the axioms of set.mm, because it seemed to be possible that something impossible like ~ topnfbey can be proven. A closer look at the proof discovered my error in reasoning: ~ topnfbey is valid because ( 0 ... +oo ) is simply "not defined" ( ... is an operation on integers, and +oo is not an integer). Maybe this could be added as remark to the comment of the theorem - it's not so obvious when you see this theorem not on April 1st ;-).

Best regards,
Alexander

tirix:
@avekens Aha I got you ! ;-)

I can certainly mention it's a candidate to the "humor" section... but the warning will have to be removed if it's actually moved there!

_To infinity... and beyond!_

nmegill:
> @avekens Aha I got you ! ;-)
> 
> I will mention it's a candidate to the "humor" section... but the warning will have to be removed if it's actually moved there!
> 
> _To infinity... and beyond!_

Go ahead and move it to "humor".  Next year we'll re-date it to appear at the top on Apr. 1. :)

tirix:
> Go ahead and move it to "humor". Next year we'll re-date it to appear at the top on Apr. 1. :)

Done in commit [070019e](https://github.com/metamath/set.mm/pull/1755/commits/070019eb8193ed53d40fb319c32ca956b91b333a), PR #1755!

PR #1749, opened 2020-08-03
Move well-founded induction, Bernoulli polynomials into main body

sctfn:


benjub:
I saw in the TOC http://us.metamath.org/mpeuni/mmtheorems.html that there is a treatment of well-founded recursion in another mathbox as well.  Did you look at how they compare ?

sctfn:
I looked at it. He's pulling from the same notes I did. The difference is
that he's got a sethood condition on the base set, while I've only got a
set-like condition on the relationship.

benjub:
Thanks @scott-fenton for having looked at it and confirmed that your treatment subsumes the other. I thought moving to Main was a good time to make sure we have the best of both worlds in Main.

benjub:
By the way, are these notes
> Lectures on Set Theory, J. Donald Monk, 2019.

? Could you mention it, maybe in the section comment ? (sorry if it's already somewhere and I missed it)

There is a link from ~frmin to http://euclid.colorado.edu/~monkd/settheory which seems to be broken.  I found that http://euclid.colorado.edu/~monkd/setth.pdf works. @nmegill: Is it possible to make a local copy in metamath.org ? (I do not know the policy for this, but for instance the OEIS favors local copies to avoid broken links as time goes by)

nmegill:
Since this is a rather large change, I updated the master with the version before this merge, just in case...

nmegill:
> There is a link from ~frmin to http://euclid.colorado.edu/~monkd/settheory which seems to be broken. I found that http://euclid.colorado.edu/~monkd/setth.pdf works. @nmegill: Is it possible to make a local copy in metamath.org ? (I do not know the policy for this, but for instance the OEIS favors local copies to avoid broken links as time goes by)

Of course the broken link above should be changed to the working one.  However, a policy of storing local pdf copies raises potential copyright concerns that I am not qualified to deal with.

There are many links on various Metamath web pages that have broken over time, and they can almost always be found on archive.org.   I change broken links to the archive.org link when I find them.

PR #1751, opened 2020-08-06
Add David Moews to the contributor list

ml-2:
I noticed that David Moews contributed some theorems, but is not on the contributor list. I'm not sure if this is intentional or an oversight, and this is my way of asking. (Feel free to reject this pull request if it's intentional, of course.)


nmegill:
Good catch, thanks.

PR #1756, opened 2020-08-08
Misc

benjub:
Part of the modifications agreed in #1745.

@david-a-wheeler : I think it would be clearer to split ~conventions, which is getting long, by creating a new statement ~conventions-labels.  I'll do it if you agree. What do you think?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17852,7 +17852,7 @@ New usage of "ply1coeOLD" is discouraged (0 uses).
 New usage of "pm10.252" is discouraged (0 uses).
 New usage of "pm11.07" is discouraged (0 uses).
 New usage of "pm110.643ALT" is discouraged (0 uses).
-New usage of "pm2.21ddOLD" is discouraged (0 uses).
+New usage of "pm2.21ddALT" is discouraged (0 uses).
 New usage of "pm2.43bgbi" is discouraged (0 uses).
 New usage of "pm2.43cbi" is discouraged (2 uses).
 New usage of "pm2.61da3neOLD" is discouraged (0 uses).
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
@benjub - I think having a separate region for just the label conventions is a great idea. Go for it!

nmegill:
> I think it would be clearer to split ~conventions, which is getting long, by creating a new statement ~conventions-labels. I'll do it if you agree. What do you think?

I'm fine with this.  Of course don't change the main ~conventions label because it's referenced externally in several places.  And I assume you will link to ~conventions-labels from within ~conventions.

benjub:
@nmegill After splitting ~conventions, MM> verify markup */file gives:
```
?Warning: No page number after [<author>] bib ref in "conventions-label".
```
I do not know where this comes from.  There are two references to [Margaris] not followed by page numbers, but they were already in ~conventions before, and not triggering warnings. The content in conventions-label is identical to what was in ~conventions... Do you have any idea ?

nmegill:
It's not obvious to me.  I'll have to look at the actual comment.  Can you email me the conventions-label giving the error?

nmegill:
The original ~conventions has a lot of [author] references that have no page numbers e.g. "...section 19 of [Margaris] often use a special naming...".  This is a problem and has messed up the bib ref page http://us2.metamath.org/mpeuni/mmbiblio.html (search for 'Margaris') even when it didn't produce an error message (I guess it found a page number somewhere later in the long comment).

With the exception of section header comments (which this is not), anything without spaces in [ ] must follow the syntax described in 'help write bib' in metamath.exe.  We can't have an isolated "[Margaris]" without a page number.

I think all words in square brackets need to be reviewed and fixed in ~conventions when they don't have page numbers.  An easy way to fix them is change isolated "[Margaris]" to "[[Margaris]".

PR #1758, opened 2020-08-09
fix a comment in my mathbox

wlammen:


nmegill:
@wlammen - In ~wl-ax12 you show that ~ax-12 can be derived from ~ax12v i.e.
```
ax12v $p |- ( x = y -> ( ph -> A. x ( x = y -> ph ) ) ) where $d x y
```
I am wondering if it is also possible to derive ~ax-12 from
```
ax12va $p |- ( x = y -> ( A. y ph -> A. x ( x = y -> ph ) ) ) where $d x y
```
(which is ~ax-12 with $d x y applied).

wlammen:
@nmegill The presented proof is not useful, because it is still based on ax-12.  This creeps in via dral1.  I just copied and adapted the proof from your mathbox as a working idea.  Maybe it should be marked as such.
Right now, I am not convinced any more, that ax-12 is provable from ax12v or wl-ax12v.  A proof idea did not work out.

What is left? Actually, ax-12 has two different aspects merged into one expression: sp (free of = and bundling) and axc112, explicitly based on bundling, and stating a simple property of an identical variable specifier.  I currently think, ax-12 is rederivable from these aspects.  If so, I am going to work this out in my mathbox, at the pace my time allows. One might discuss then whether one should dissect complex axioms into their fundamental ideas, at the expense of the count of axioms.  Something that is already decided in the case of ax-meredith.

What can be done right now is replacing ax12v with wl-ax12v.  My comment in wl-ax12v says why I am in favour of adding the extra dv constraint.

benjub:
> I currently think, ax-12 is rederivable from these aspects.

I do not think so. Here is a model, which is a variation on Monk's:
* The language consists of one non-logical predicate, say P, which is unary.
* The domain of discourse has three objects, say 0, 1, 2.
* Equality is interpreted as the equivalence relation with equivalence classes {0, 1} and {2}.
* The predicate P holds for exactly 0.

Then, ax-1--11, ax-13, axc112, sp hold, but ax-12 does not hold. Can you double-check ?

> What can be done right now is replacing ax12v with wl-ax12v. My comment in wl-ax12v says why I am in favour of adding the extra dv constraint.

If wl-ax12v is to be moved to Main, then I would label it ax12vv (because of the two dv conditions), and have it **in addition to** the existing ax12v (i.e., **do not remove** ax12v), and use ax12vv in the proof of 19.8a (and ax12a2, exsb, mo2v, 2eu6, rexsb) as you did in your mathbox.

By the way: the statement of wl-ax12v2 seems to be identical to the statement of axc15.

wlammen:
@benjub You are right, I have come to the same conclusion and abandoned the idea of working from sp.  Instead one can let ax12v (or wl-ax12v) and axc112 be the only theorems referring to ax-12 directly, from which it can be rederived as wl-ax12.  This gives a good idea of how bundling in ax-12 affects the whole first order logic, and points out the two different aspects of ax-12 more clearly than before.

@nmegill The hypothesis of ax12a2 looks pretty much the same as wl-ax12v.  The quantifier A. z in front of ph seems pointless, since ph and z are assumed too be distinct.
ax12va follows from wl-ax12v as can essentially be seen in the proof of wl-ax12v2.  I doubt that axc112 follows from ax12va only, and I think you need ax-12 to prove axc112, but I have no profound confirmation of that.  

Wolf

PR #1759, opened 2020-08-09
Angle construction, existence and uniqueness

tirix:


david-a-wheeler:
Unfortunately I've been swamped with work & haven't had a chance to seriously review this. So I've only taken a brief look so far :-(.

However, a quick note: it's really challenging for me to verify that `acopyeu` is a representation of Theorem 11.15 of [Schwabhauser] p. 98. The set.mm is so different! I understand why, in particular, it's clear you're trying to minimize the creation of new definitions. But it's still a little tricky. It might be useful to put, near the beginning, a "master translation list" of expressions in Schwabhauser and their equivalent in set.mm. Then it'd be easier for anyone confirm that proofs map one-to-one when they're supposed to.

PR #1760, opened 2020-08-10
Misc

benjub:
Rest of the modifications announced in #1745 and #1756.


benjub:
@nmegill Here is the commit with the bibliography problem. I can also email you set.mm or only the problematic comment of ~conventions-label.

nmegill:
> @nmegill Here is the commit with the bibliography problem. I can also email you set.mm or only the problematic comment of ~conventions-label.

No need to email; I found the problem.  There are 2 situations where [Margaris] is not followed by a page number.  Line 406820 reads:
```
       theorems from section 19 of [Margaris] often use a special naming
```
and line 406832 reads,
```
       19.x series of theorems from [Margaris] (see above) and ~ 0.999... .
```
To fix both of these, change "[Margaris]" to "[[Margaris]" in these 2 cases.  That can be our permanent fix (the [[ will become a single [ on the web page).  I verified that the corrupted 'write bibliography' output (mmbiblio.html) will become resolved with this fix.

The error message unfortunately isn't very informative.  The metamath.exe code was written anticipating short comments with typically one bib ref where the error would be obvious.  And I don't understand why it didn't complain before you split off ~conventions-label, which I'll have to look into.  That particular code was hastily written years ago and frankly is ugly and not easy to maintain.  I have vague long-term plans to rewrite much of the web page generation stuff with cleaner and more maintainable code, with better error messages I hope.

david-a-wheeler:
On August 10, 2020 7:39:54 PM EDT, nmegill <notifications@github.com> wrote:
>> @nmegill Here is the commit with the bibliography problem. I can also
>email you set.mm or only the problematic comment of ~conventions-label.
>
>No need to email; I found the problem.  There are 2 situations where
>[Margaris] is not followed by a page number.  Line 406820 reads:
>```
>      theorems from section 19 of [Margaris] often use a special naming
>```
>and line 406832 reads,
>```
>   19.x series of theorems from [Margaris] (see above) and ~ 0.999... .
>```
>To fix both of these, change "[Margaris]" to "[[Margaris]" in these 2
>cases.  That can be our permanent fix (the [[ will become a single [ on
>the web page).  I verified that the corrupted 'write bibliography'
>output (mmbiblio.html) will become resolved with this fix.
>
>The error message unfortunately isn't very informative.  The
>metamath.exe code was written anticipating short comments with
>typically one bib ref where the error would be obvious.  And I don't
>understand why it didn't complain before you split off
>~conventions-label, which I'll have to look into.  That particular code
>was hastily written years ago and frankly is ugly and not easy to
>maintain.  I have vague long-term plans to rewrite much of the web page
>generation stuff with cleaner and more maintainable code, with better
>error messages I hope.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/metamath/set.mm/pull/1760#issuecomment-671642595

Can we make the special form [[Margaris]] , that is, balanced square brackets? Unbalanced markers bug me. I have spent a lifetime looking for them, so they look hideously ugly when that happen intentionally.
--- David A.Wheeler

nmegill:
> Can we make the special form [[Margaris]] , that is, balanced square brackets? Unbalanced markers bug me. I have spent a lifetime looking for them, so they look hideously ugly when that happen intentionally. --- David A.Wheeler

Better, how about just removing the square brackets completely?  I see no reason to have them when it's not a link to a bibliography page.  Margaris has already been linked to earlier, so there's no issue about what work we're talking about.  And that would eliminate yet another special form the user has to learn.

david-a-wheeler:
> Better, how about just removing the square brackets completely?

+1.

It's hard to beat that solution for simplicity :-).

benjub:
Thanks @nmegill and @david-a-wheeler. I fixed the two occurrences by using "[[Margaris]'' since I am not sure if "[[Margaris]]" is accepted yet.  I would prefer "see [[Margaris]]" over "see Margaris" with no brackets, since the former clearly refers to the book referred to as [Margaris] in the bibliography page, whereas "Margaris" could refer to the person ("go see Margaris at the next department tea, and ask him for details" ;-) )
But I'll use the form @nmegill prefers.

nmegill:
To be honest I'm not really happy with either [[Margaris] or [[Margaris]].  The "[[" escape has never been needed so far.  It wasn't really intended for general use but more as an emergency way to bypass a missing bib entry temporarily to avoid an error message, that kind of thing.  It also supplements the ~~ and \`\` escapes for labels and math symbols.

In set.mm, we've always been able to word things to avoid a non-bib ref bracket.  I'd prefer to do the same here.

Specifically:
```
       theorems from section 19 of [Margaris] often use a special naming
```
could be changed to (note that "section" is a keyword):
```
       theorems from section 19 of [Margaris] p. 90 often use a special naming
```
and
```
       19.x series of theorems from [Margaris] (see above) and ~ 0.999... .
```
could be changed to 
```
       19.x series of theorems (see above) and ~ 0.999... .
```
I am traveling today and won't be able to get back to this until late tonight.  So I'm going ahead and accepting the PR for now, and we can change it later or continue the discussion.

PR #1761, opened 2020-08-11
Misc

benjub:
* Format the bibliographic references according to https://github.com/metamath/set.mm/pull/1760#issuecomment-671905131
* @nmegill : I added a new reference to [Church] in a comment in set.mm. Where do I add:
>   [Church] Alonzo Church, Introduction to Mathematical Logic, Princeton University Press, 1956.

"MM>help write bib" mentions mmbiblio.html, which I cannot find...
And then, do I have to run "write bib" or something similar ?

### DIFF
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -4925,6 +4925,9 @@ DeWitt-Morette, with Margaret Dillard-Bleick, <I>Analysis, Manifolds and
 Physics,</I> Elsevier Science B.V., Amsterdam (1982) [QC20.7.A5C48
 1981].</LI>
 
+<LI><A NAME="Church"></A> [Church] Church, Alonzo, <I>Introduction to
+Mathematical Logic,</I> Princeton University Press, 1956.</LI>
+
 <LI><A NAME="Clemente"></A> [Clemente] Clemente Laboreo, Daniel
 <I>Introduction to natural deduction</I> (2014); available
 at <A HREF="http://www.danielclemente.com/logica/dn.en.pdf">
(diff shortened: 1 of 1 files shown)


nmegill:
> * @nmegill : I added a new reference to [Church] in a comment in set.mm. Where do I add:
> 
> > [Church] Alonzo Church, Introduction to Mathematical Logic, Princeton University Press, 1956.

You add it to mmset.raw.html in the bibliography section (search for "bib" in double quotes).  In the current version of that file, you would add the Church entry alphabetically below line 4927.  For the format, use the surrounding examples as a guide.  In particular, the anchor '\<A NAME="Church"\>' (case sensitive) is what 'verify markup' will look for.  Then include the modified mmset.raw.html in your PR.

BTW in the set.mm $t statement, 'htmlbibliography "mmset.html";' specifies the file where metamath.exe looks for and links to these anchors.  iset.mm would point to mmil.html and so on.

> 
> "MM>help write bib" mentions mmbiblio.html, which I cannot find...
> And then, do I have to run "write bib" or something similar ?

You don't have to worry about mmbiblio.html unless you are generating the site locally.  In that case, it will be part of the site download.  You can also save the web page from the browser if you just want to test 'write bib'.

mmbiblio.html is a template file that is automatically filled in (or refreshed) by 'write bib' during the site build.  Since it isn't normally edited by users (and is dangerous to do so if the string matches expected by 'write bib' are tampered with), it isn't currently under revision control on GitHub.  I have no objection to putting it there if there is a need to edit it, but it hasn't been touched by a human in many years, and metamath.exe is the only program that modifies it.

PR #1763, opened 2020-08-12
Misc

benjub:
@nmegill : I added to some comments the sentence "Inference associated with ~ ..." because I think this provides hyperlinks which may be useful. Do not worry, I will not do this for the whole database: only some less obvious cases and the subsection before the first use of ax-3.


wlammen:
@benjub syld contains a typo 'decution'

PR #1764, opened 2020-08-13
Add conventions info

david-a-wheeler:
Note that we try to have:

* Number construction independence
* Minimize hypotheses (except for construction independence
  and number theorem domains)


david-a-wheeler:
This attempts to capture (in a summary way) the mailing list discussion.

benjub:
to independent --> to be independent (and a period instead of a comma at the end of that sentence)
(and "reintroduce" without dash?)

nmegill:
@david-a-wheeler There seems to be a Travis error.

david-a-wheeler:
@nmegill - I accidentally used some tabs instead of spaces. Easily fixed. I will fix that and also address the issues identified by @benjub , but I need to do some work first :-).

avekens:
@david-a-wheeler if you revise the conventions anyway, could you add also the following convention:

```
 class symbols for functions (like abs, sin, etc) should usually not have
 leading/trailing blanks in their html/latex representation (in contrast to class symbols
 for operations, like gcd, sadd, eval, etc.).

```
I made this mistake the other day, and @benjub pointed it out to me.

nmegill:
` om ` s.b. `_om `

avekens:
@david-a-wheeler great - thank you!

PR #1765, opened 2020-08-13
Misc

benjub:


benjub:
@digama0 : I had a look at the section on half and full adders.  It seems to me that what is defined in set.mm is actually the two outputs (sum and carry) of the full adder, and not the "half and full adders". Do you agree that I edit the section accordingly ? (I had already edited the section comment to add an explanation)

Further explanation: in https://en.wikipedia.org/wiki/Adder_(electronics), the half adder is defined as having two inputs (it sums two 1-bit numbers) and the full adder as having three inputs (it takes a carry into account), while both adders have two outputs (sum and carry).

digama0:
You are right, that name is nonsense. The outputs of the half adder are XOR and AND, so they don't need definition, while the outputs of the full adder are df-had (maybe better called XOR3) and df-cad (maybe better called the majority function).

PR #1767, opened 2020-08-15
Second congruence theorem: ASA. (Angle-Side-Angle)

tirix:
Also, reductions in distinct variable restrictions.


david-a-wheeler:
Spectacular! Congratulations!

PR #1768, opened 2020-08-16
Refactoring of the "Half-Line" definition, Axiom of Pasch.

tirix:
A new global definition is introduced for the "half-line" concept. I originally wanted to avoid introducing too many definitions, but this one ended up being used very often. The explicit definition overall saved ~ 100 lines of set.mm.

A local definition is still used for the concept of "lying on opposed sides of a line".


tirix:
I'm not sure why the Travis build is not launched...

david-a-wheeler:
I think every once in a great while the messages get dropped between GitHub and Travis. You're absolutely right that it shouldn't happen, it should really be retransmitted, but it seems to happen anyway.

I suggest making a completely trivial change in this branch, such as a tweak in a comment, and push that.

david-a-wheeler (approved):
This looks good! I think adding this half-line definition is a good idea. It clarifies things, makes the work closer to the original, and shortens the results in set.mm. What's not to like?

david-a-wheeler:
The Travis builds were triggered but aren't being shown here. I have reasons to hope that this is temporary.

There appears to be a **failed** Travis build on this, per: https://travis-ci.org/github/metamath/set.mm/builds/718315621
I think this is from the latest version, though commit id mismatch makes that hard to be certain.

All tests passed except metamath.exe:

~~~~
?Warning: Line number 383476 ends with a space character, which is discouraged.
?Warning: Line number 383486 ends with a space character, which is discouraged.
~~~~

I've verified those problems. Would you please fix that?

tirix:
Great!
It looks like the build was triggered for my last commit (which has not much to do with geometry, just a trial in my mathbox)
Sorry the spaces were left in the previous one.

PR #1771, opened 2020-08-17
Theorems moved to the main part of set.mm : supadd , inisegn0 , ad4an…

glacode:
Theorems moved to the main part of set.mm : supadd , inisegn0 , ad4ant* and ad5ant*


nmegill:
Thanks for re-doing this.

PR #1773, opened 2020-08-18
clean up 'adders' section

benjub:
Clean up 'adders' section, according to #1765 

Two questions to @nmegill and @digama0:
* I kept the current label and token for "had(d)", although it is not the "half adder".  It would be more consistent to use "sad(d)" but it's already used, and there is the "cost of change". So I let it this way for the moment, tell me if you see a better option.
* I checked the *OLD proofs of this section (following @wlammen 's shortenings) and I think they can be safely deleted; I can delete them in this PR if you agree.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14764,10 +14764,6 @@ New usage of "branmfn" is discouraged (1 uses).
 New usage of "braval" is discouraged (10 uses).
 New usage of "c-bnj14" is discouraged (131 uses).
 New usage of "c-bnj18" is discouraged (58 uses).
-New usage of "cad1OLD" is discouraged (0 uses).
-New usage of "cadcombOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> * I checked the *OLD proofs of this section (following @wlammen 's shortenings) and I think they can be safely deleted; I can delete them in this PR if you agree.

That's OK with me.

On a related question, does the adder stuff belong in the main set.mm rather than a mathbox?  Is there any future work that might depend on it?

benjub:
> On a related question, does the adder stuff belong in the main set.mm rather than a mathbox? Is there any future work that might depend on it?

There is at least the section "6.1.5  Bit sequences" (to define and prove some properties of df-sad).

nmegill:
> There is at least the section "6.1.5 Bit sequences" (to define and prove some properties of df-sad).

OK, I didn't see that.

PR #1774, opened 2020-08-20
Relationship between factorial, lcm and primorial

avekens:
Theorems showing that the primorial of  n is less than or equal to the least common multiple of the numbers 1...n (~prmorlelcms), which itself is less than or equal to the factorial of n (~lcmslefac).

AV's Mathbox: Bibliographic references adjusted


avekens:
@david-a-wheeler There are still problems with Travis. It didn's start on the first try, and it didn't start after another push with small changes...

avekens:
As observed by @tirix , Travis-Jobs are executed (for this PR see https://travis-ci.org/github/metamath/set.mm/builds/719500019), although they are not shown here...

PR #1775, opened 2020-08-20
Added a definition for sub-matrices in my mathbox.

tirix:
This definition allows to compute the determinant of submatrices, unlike the current definition in the main part of set.mm.


tirix:
Travis builds are [here](https://travis-ci.org/github/metamath/set.mm/builds/719532058) and [here](https://travis-ci.org/github/metamath/set.mm/builds/719532093). Both are green.
Not sure why they are not shown again.

david-a-wheeler:
I don't know either. This is pretty clearly a bug in Travis or GitHub, not in what we're doing.

I have to do work stuff for now. Maybe we can report this to Travis.

PR #1776, opened 2020-08-20
shorten ~equsex and keep the longer proof as equsexALT since it may g…

benjub:
…ive additional insight

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18880,7 +18880,7 @@ Proof modification of "bj-aevlem1v" is discouraged (42 steps).
 Proof modification of "bj-ax12v" is discouraged (35 steps).
 Proof modification of "bj-ax16gb" is discouraged (18 steps).
 Proof modification of "bj-ax16nf" is discouraged (31 steps).
-Proof modification of "bj-ax8" is discouraged (17 steps).
+Proof modification of "bj-ax8" is discouraged (55 steps).
 Proof modification of "bj-ax9" is discouraged (35 steps).
 Proof modification of "bj-axc10v" is discouraged (37 steps).
 Proof modification of "bj-axc11nlemv" is discouraged (58 steps).
--- a/discouraged
+++ b/discouraged
@@ -14330,7 +14330,6 @@ New usage of "bj-axdd2" is discouraged (0 uses).
 New usage of "bj-axtd" is discouraged (0 uses).
 New usage of "bj-ceqsalgALT" is discouraged (0 uses).
 New usage of "bj-ceqsalgvALT" is discouraged (0 uses).
-New usage of "bj-cleljustOLD" is discouraged (0 uses).
 New usage of "bj-con3thALT" is discouraged (0 uses).
 New usage of "bj-con4iALT" is discouraged (0 uses).
 New usage of "bj-consensusALT" is discouraged (0 uses).
@@ -15907,6 +15906,7 @@ New usage of "equidqe" is discouraged (2 uses).
 New usage of "equncomVD" is discouraged (0 uses).
 New usage of "equncomiVD" is discouraged (0 uses).
 New usage of "equsb3ALT" is discouraged (0 uses).
+New usage of "equsexALT" is discouraged (0 uses).
 New usage of "erngbase-rN" is discouraged (4 uses).
 New usage of "erngdv-rN" is discouraged (0 uses).
 New usage of "erngdvlem1-rN" is discouraged (3 uses).
@@ -18940,8 +18940,7 @@ Proof modification of "bj-ceqsalv" is discouraged (10 steps).
 Proof modification of "bj-chvarv" is discouraged (18 steps).
 Proof modification of "bj-chvarvv" is discouraged (10 steps).
 Proof modification of "bj-clabel" is discouraged (43 steps).
-Proof modification of "bj-cleljust" is discouraged (51 steps).
-Proof modification of "bj-cleljustOLD" is discouraged (25 steps).
+Proof modification of "bj-cleljust" is discouraged (22 steps).
 Proof modification of "bj-clelsb3" is discouraged (63 steps).
 Proof modification of "bj-cleqh" is discouraged (108 steps).
 Proof modification of "bj-cleqhyp" is discouraged (21 steps).
@@ -18982,7 +18981,7 @@ Proof modification of "bj-equsalv" is discouraged (39 steps).
 Proof modification of "bj-equsalvv" is discouraged (38 steps).
 Proof modification of "bj-equsb1v" is discouraged (17 steps).
 Proof modification of "bj-equsexhv" is discouraged (10 steps).
-Proof modification of "bj-equsexv" is discouraged (38 steps).
+Proof modification of "bj-equsexv" is discouraged (37 steps).
 Proof modification of "bj-equsexvv" is discouraged (36 steps).
 Proof modification of "bj-eunex" is discouraged (53 steps).
 Proof modification of "bj-exlimmpbi" is discouraged (11 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
Question:  what is the reason for keeping ~equsexALT?

wlammen:
@benjub By what criteria is the new proof of equsex an improvement over the old one?

benjub:
To @nmegill and @wlammen : both proofs are now on us2: http://us2.metamath.org/mpeuni/equsex.html and http://us2.metamath.org/mpeuni/equsexALT.html.  Also, the Metamath program gives:
```
MM> sh pr equsex/co
Proof of "equsex":
      ( weq wa wex pm5.32i exbii ax6e 19.41 mpbiran bitri ) CDGZAHZCIPBHZCIZBQR
      CPABFJKSPCIBCDLPBCEMNO $.
---------The proof of "equsex" (95 bytes) ends above this line.
MM> sh pr equsexALT/co
Proof of "equsexALT":
      ( weq wa wex biimpa exlimi wi wal equsal equs4 sylbir impbii ) CDGZAHZCIZ
      BSBCERABFJKBRALCMTABCDEFNACDOPQ $.
---------The proof of "equsexALT" (104 bytes) ends above this line.
```
I think it's useful to keep both proofs.  The statements ~equsal and ~equsex are in some sense dual, and the former is more fundamental, since \forall is primitive while \exists is defined from it.  Therefore, there are two ways to prove ~equsex: either as a corollary of ~equsal (that is what equsexALT is doing, using ~equs4 to "translate"), or by proving it directly, with a proof parallel to that of ~equsal (that is what ~equsex is doing: note that 19.41 is akin to 19.23, importing a quantifier inside a conjunction instead of an implication; similarly mpbiran versus a1bi, and pm5.32i versus pm5.74i, and of course exbii versus albii).

Most of the time, if there are two similar "hard" theorems, one prefers to prove only one, and prove the other as a corollary.  But here, the translation (via ~equs4) is actually as hard (in that it requires ~ax-13) as each theorem, so that ~equsexALT relies on more "harder theorems" than ~equsex (for instance, looking at theorems requiring ax-13, ~equsexALT uses ~equsal and ~equs4 while ~equsex uses only ~ax6e). Therefore, I think ~equsex should be the main proof and ~equsexALT the alternate one (in addition to the above difference, 95 versus 104 bytes). There is also the consideration that it will adapt better to intuitionistic logic, where \forall and \exists are not inter-definable.

**Edit:** another manifestation of the above is that the proof of ~equsex has only existential quanfiers, while the proof of ~equsexALT has both existential and universal quantifiers.

(I'm going to add to the comment of ~equsex the sentence "See ~ equsexALT for an alternate proof from ~ equsal .")

wlammen:
I consider the ALT proof shorter than the newly introduced one.  The rules by which a proof qualifies as shorter than another are in preference order: 1. Number of proof bytes in compressed format. This does not include the labels, that are arbitrarily chosen anyway.  Both proofs consist of 41 Bytes1 2. The number of proof lines.  This is the same for both proofs (8 lines)  3.  The number of symbols used to represent the proof.  And here is where the ALT proof wins.  These criteria are somewhere written down on the Metamath web site, but I forgot where I saw them.

benjub:
I'll modify set.mm as @nmegill decides. What is important is to have both; which one is ALT is not too important.

Of course the length of labels is arbitrary, but the label count is 9 versus 11 in favor of ~equsex. Also, the symbols used are -> , <-> , /\ , E. , F/ (and =) for both proofs, and ~equsexALT uses A. too, so it's one more.

Also, I'm pretty sure that you had already seen the possibility of proving ~equsex directly, so I'd be happy to remove the credit to me and add it to you, if you can give me an approximate date (or else 20-Aug-2020).


**Edit:** as for the "normal" proof format:
```
MM> sh pr equsex/normal
Proof of "equsex":
      vx vy weq wph wa vx wex vx vy weq wps wa vx wex wps vx vy weq wph wa vx
      vy weq wps wa vx vx vy weq wph wps equsex.1 pm5.32i exbii vx vy weq wps
      wa vx wex vx vy weq vx wex wps vx vy ax6e vx vy weq wps vx equsex.nf
      19.41 mpbiran bitri $.
---------The proof of "equsex" (232 bytes) ends above this line.
MM> sh pr equsexALT/normal
Proof of "equsexALT":
      vx vy weq wph wa vx wex wps vx vy weq wph wa wps vx equsex.nf vx vy weq
      wph wps equsex.1 biimpa exlimi wps vx vy weq wph wi vx wal vx vy weq wph
      wa vx wex wph wps vx vy equsex.nf equsex.1 equsal wph vx vy equs4 sylbir
      impbii $.
---------The proof of "equsexALT" (224 bytes) ends above this line.
```

wlammen:
The above three rules are those by which I decide borderline cases like this one for a decade by now.  One can discuss whether one should enhance these to better cover aspects like those mentioned by you.  But I am in favour of having a fixed set of rules, to avoid any contention arising from different point of views. I saw the above rules written down somewhere, they are definitely not of my origin.  Unfortunately I am currently unable to retrieve them.  Next time I come across them again, I will jot down their web address to be better prepared.  As for the credits, I am not interested in appearing as a contributor to a proof unless I am one.

wlammen:
I found the reference. It is set.mm itself, around line 12200.

Shorter proofs are welcome, and any shorter proof I accept will be acknowledged
in the theorem's description.  However, in some cases a proof may be "shorter"
or not depending on how it is formatted.  This section provides general
guidelines.

Usually I will automatically accept shorter proofs that (1) shorten the set.mm
file (with compressed proofs), (2) reduce the size of the HTML file generated
with SHOW STATEMENT xx / HTML, (3) use only existing, unmodified theorems in
the database (the order of theorems may be changed, though), and (4) use no
additional axioms.

Usually I will also automatically accept a _new_ theorem that is used to
the new theorem, not including the acknowledgment) decreases as a result.

In borderline cases, I typically place more importance on the number of
compressed proof steps and less on the length of the label section (since the
names are in principle arbitrary).  If two proofs have the same number of
compressed proof steps, I will typically give preference to the one with the
smaller number of different labels, or if these numbers are the same, the proof
with the fewest number of characters that the proofs happen to have by chance
when label lengths are included.

nmegill:
You can search for "Acceptable shorter proofs" in set.mm (or just the word "shorter") to find the text @wlammen mentions.

It seems I added this text around 2004, probably in response to someone's question.  I don't know whether the rules for distinguishing borderline cases are optimal, but for now at least it establishes concrete rules.  We can discuss possible changes to the rules for the future if it makes sense to do so.

Since this is a borderline case where both opinions are probably valid, I propose swapping back the equsex and equsexALT labels, restoring the original eqsex and its original comment and proof.  I would probably prefer eqsexALT (BJ's proof) in his mathbox, but if he feels it is useful to have in the main set.mm, I will accept that.  For eqsexALT, because its proof is completely different, BJ should probably take out both "Proof shortened by" tags and just say "... (Revised by BJ, 20-Aug-2020.) Alternate proof for ~ eqsex ." (and/or anything else you want to say about the proof).  Note that having the comment next to the "Revised" tag may help the reader know that the comment is about the revision.

benjub:
Ok, I'm going to do that in my next PR, but then I will let @nmegill change, if he sees fit, "Acceptable shorter proofs" accordingly, since it currently states:
> If two proofs have the same number of compressed proof steps, I will typically give preference to the one with the smaller number of different labels

(although the "typically" may dispense from doing so).

The same thing (two proofs of equal size) applies to bj-equsexv(v) and I will see what I do about them.

On a related note: I will probably propose to move bj-equsexvv to Main, since it allows to remove from ~cleljust dependencies on ax-10, ax-12 and ax-13 (and also df-nf), see ~bj-cleljust.  I think it is important to show that this important theorem can be proved within Tarski's original FOL axiom schemes, without the auxiliary axioms.

PR #1777, opened 2020-08-21
Start on real number completeness in iset.mm

jkingdon:
We've been putting off dealing with real number completeness in iset.mm. But with square root being the next section to formalize, this cannot be delayed any longer.

 The least upper bound property, as generally formulated, needs excluded middle and is not available. Although modified least upper bound properties may be possible, it seems like the easiest way to address completeness may be via Cauchy sequences. We'll need them eventually for complex exponentiation (and other things).

This pull request contains a first step, which is showing that a Cauchy approximation defined on `Q.` converges to a limit in `P.` (see the comments in the pull request, or the cited parts of the HoTT book, for more on Cauchy approximations versus Cauchy sequences and other details). The eventual goal is going to be something like http://us.metamath.org/mpeuni/caucvg.html (and most of the other theorems in that section). My hope is that this pull request is the hardest part, as it establishes the basic idea of real number completeness. Time will tell on that, of course.

The pull request contains a few small cleanups but nothing large except for the Cauchy approximation work itself.


jkingdon:
I don't know if there is anything I can do to get travis to report its results here, but it seems like the travis build did happen: https://travis-ci.org/github/metamath/set.mm/builds/719850189

david-a-wheeler:
This seems to be a continuing problem, where GitHub does not tell us about Travis results. I was hoping this problem would go away, but that hope seems to be unfounded. I'll look into this further when I get a chance. I have no idea why this is happening.

PR #1779, opened 2020-08-22
shorten equid

wlammen:
Most of the diff not related to equid is caused by rewrap, that indents proof blocks differently to what was submitted in pull request #1778.  In addition it complains about 
Warning: Bibliography keyword missing in comment for "ax-bgbltosilva".
    (See HELP WRITE BIBLIOGRAPHY for list of keywords.)
Likely, this is related to
"see result of [OeSilva] p. ?."
where in addition the page number is missing.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15906,6 +15906,7 @@ New usage of "eqresr" is discouraged (4 uses).
 New usage of "eqsbc3rVD" is discouraged (0 uses).
 New usage of "equid1" is discouraged (0 uses).
 New usage of "equid1ALT" is discouraged (0 uses).
+New usage of "equidOLD" is discouraged (0 uses).
 New usage of "equidq" is discouraged (0 uses).
 New usage of "equidqe" is discouraged (2 uses).
 New usage of "equncomVD" is discouraged (0 uses).
@@ -19444,6 +19445,7 @@ Proof modification of "eqrdavOLD" is discouraged (59 steps).
 Proof modification of "eqsbc3rVD" is discouraged (129 steps).
 Proof modification of "equid1" is discouraged (50 steps).
 Proof modification of "equid1ALT" is discouraged (36 steps).
+Proof modification of "equidOLD" is discouraged (26 steps).
 Proof modification of "equidq" is discouraged (27 steps).
 Proof modification of "equidqe" is discouraged (30 steps).
 Proof modification of "equncomVD" is discouraged (53 steps).


wlammen:
No Travis checks? Everything ok on the servers?

wlammen:
Another unrelated note:
http://us2.metamath.org:88/mpeuni/ax-luk1.html shows an undetected cross-mathbox reference. Apart from what git blame will reveal how this came about, shouldn't this be marked by the verifiers?

benjub:
> Most of the diff not related to equid is caused by rewrap, that indents proof blocks differently to what was submitted in pull request #1778.

Indeed. I've been suggesting that submitters systematically execute the two commands given in https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md.  Since the suggestion was not enough (I'm not blaming anyone, and I have sometimes forgotten myself to execute them), is it possible to integrate it in github or Travis ?  That is, these two commands would be automatically executed on any PR ? @nmegill or the Travis wizard @david-a-wheeler may know a way ?

This is not a theoretical problem: more modifications means potentially more merge conflicts (and I'm pretty sure I will have some merge conflicts when I merge this with the PR I am preparing).

> In addition it complains about
> Warning: Bibliography keyword missing in comment for "ax-bgbltosilva".
> (See HELP WRITE BIBLIOGRAPHY for list of keywords.)
> Likely, this is related to
> "see result of [OeSilva] p. ?."
> where in addition the page number is missing.

Yes. In the PR I am preparing, I changed it to "see Table 1 of [OeSilva] p. 2034." (I checked the paper referenced by @avekens).  This would still not work since "Table" is for the moment not a "bibliography keyword".  May I suggest to add to the Metamath program (some of) the following bibliography keywords: Conjecture, Fact, Table, Scolia (or Scolion) (used by Bourbaki), and maybe also Introduction, Conclusion, Paragraph, Subsection, which may be useful when pointing to a specific part of a paper.

> Another unrelated note:
> http://us2.metamath.org:88/mpeuni/ax-luk1.html shows an undetected cross-mathbox reference. Apart from what git blame will reveal how this came about, shouldn't this be marked by the verifiers?

That's interesting. The statement bj-0 has a discouragement tag, so apparently the discouragement tag is not taken into account (in addition to the verification not catching this use, as @wlammen noticed).  This is probably because bj-0 does not appear as being referenced by ax-luk1 in set.mm.  This is not a proof, since ax-luk1 is an axiom, but a "syntax breakdown", which is not explicitly in set.mm. I do not know how syntax breakdowns are generated...

avekens:
> In addition it complains about
Warning: Bibliography keyword missing in comment for "ax-bgbltosilva".
(See HELP WRITE BIBLIOGRAPHY for list of keywords.)
Likely, this is related to
"see result of [OeSilva] p. ?."
where in addition the page number is missing.

Who/what complains? The keyword "result" was added by Norm in the latest version 0.187 of metamath.exe, and I cannot find the mentioned warning neither in the Travis logs nor in my personal checks.

wlammen:
@avekens In metamath(.exe) load set.mm and run "verify markup * /file_skip", and you will see the warning.
OK, my version is 0.183, maybe the warning will be gone then.
Confirmation: after upgrading to 0.187 the warning is gone. So my apologies, this is my fault.
It is perhaps a bit unfortunate that it wasn't announced. @benjub check whether the keywords you like to see as well are now added, or how your pull request can be adapted to recent changes.

nmegill:
> Indeed. I've been suggesting that submitters systematically execute the two commands given in https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md. Since the suggestion was not enough (I'm not blaming anyone, and I have sometimes forgotten myself to execute them), is it possible to integrate it in github or Travis ? That is, these two commands would be automatically executed on any PR ? @nmegill or the Travis wizard @david-a-wheeler may know a way ?

I don't think Travis has the ability to permanently modify a file in a PR submission, and I'm not sure we would want it to even if we could.  I suppose we could regenerate set.mm locally with these two commands applied, then do a diff to report a Travis failure if the regenerated set.mm doesn't match the one submitted for the PR.  Both of these commands are somewhat CPU-intensive (although I have a long-term goal to speed up /rewrap) and would put extra burden on the Travis computers;  I don't know if that's a problem.

> May I suggest to add to the Metamath program (some of) the following bibliography keywords: Conjecture, Fact, Table, Scolia (or Scolion) (used by Bourbaki), and maybe also Introduction, Conclusion, Paragraph, Subsection, which may be useful when pointing to a specific part of a paper.

I will add these keywords to the next metamath.exe release.  (A proper fix would be to make the list part of the $t statement, which is on my todo list and may happen eventually.)

> That's interesting. The statement bj-0 has a discouragement tag, so apparently the discouragement tag is not taken into account (in addition to the verification not catching this use, as @wlammen noticed). This is probably because bj-0 does not appear as being referenced by ax-luk1 in set.mm. This is not a proof, since ax-luk1 is an axiom, but a "syntax breakdown", which is not explicitly in set.mm. I do not know how syntax breakdowns are generated...

The syntax breakdowns for the axiom/definition displays are done using the "improve" command's algorithm and (as I recall) without checking "discouraged" tags.  What you see is a display artifact that is harmless and doesn't affect anything stored in set.mm.  If we honor "discouraged" tags, that might be a problem since they aren't prohibitions, and the program doesn't know whether the use of a discouraged syntax was intentional.

I could rewrite the algorithm to first check non-discouraged statements for the syntax breakdown then if it fails add in discouraged ones, but that seems like a lot of work to do just for ~bj-0.  Stand-alone syntax proofs such as bj-0 weren't really intended for general use - ~wel and ~weq are the only ones other than ~bj-0, and they have a specific purpose, which is to provide a way to overload the = and e. syntax. 

So unless people are really bothered by this, I won't do anything about it.

nmegill:
It looks like the Travis script had a problem.  Maybe close and reopen the PR?  (There used to be a button to restart on the Travis page but I couldn't find it.)

```
The command "scripts/download-metamath" failed and exited with 3 during .
```

wlammen:
@nmegill I can certainly live (to some extent) with web pages in my mathbox that would look differently should the underlying theorem/axiom be moved to main. In particular when only syntax breakdown is affected.  It is up to you to live with noise produced by unwarned people like me then.

I am going to close and reopen the pull request now.

benjub:
> Stand-alone syntax proofs such as bj-0 weren't really intended for general use - ~wel and ~weq are the only ones other than ~bj-0, and they have a specific purpose, which is to provide a way to overload the = and e. syntax.

Indeed, ~bj-0 is not intended for general use.  I give a proof, in the comment of ~bj-1, that syntactic theorems are generally useless, and I mention there that the only two syntactic theorems in Main are ~wel and ~weq indeed.  I think it is a nice curiosity to have them, but if @nmegill and/or @wlammen prefer, I can remove them from my mathbox.

nmegill:
> May I suggest to add to the Metamath program (some of) the following bibliography keywords: Conjecture, Fact, Table, Scolia (or Scolion) (used by Bourbaki), and maybe also Introduction, Conclusion, Paragraph, Subsection, which may be useful when pointing to a specific part of a paper.

These are now in metamath.exe version 0.188, uploaded yesterday.  As a reminder, 'help write bib' will show the current list of keywords.

PR #1784, opened 2020-08-23
Misc

benjub:
Switched equsex and equsexALT according to #1776

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18855,6 +18855,8 @@ Proof modification of "bcsiALT" is discouraged (444 steps).
 Proof modification of "bianirOLD" is discouraged (19 steps).
 Proof modification of "bitr3" is discouraged (14 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).
+Proof modification of "bj-0" is discouraged (5 steps).
+Proof modification of "bj-1" is discouraged (5 steps).
 Proof modification of "bj-19.21t" is discouraged (39 steps).
 Proof modification of "bj-2stdpc4v" is discouraged (28 steps).
 Proof modification of "bj-abbi" is discouraged (87 steps).
--- a/discouraged
+++ b/discouraged
@@ -19452,7 +19452,7 @@ Proof modification of "equidqe" is discouraged (30 steps).
 Proof modification of "equncomVD" is discouraged (53 steps).
 Proof modification of "equncomiVD" is discouraged (19 steps).
 Proof modification of "equsb3ALT" is discouraged (44 steps).
-Proof modification of "equsexALT" is discouraged (38 steps).
+Proof modification of "equsexALT" is discouraged (37 steps).
 Proof modification of "euequ1OLD" is discouraged (42 steps).
 Proof modification of "euexALT" is discouraged (32 steps).
 Proof modification of "eujustALT" is discouraged (188 steps).
--- a/discouraged
+++ b/discouraged
@@ -14342,13 +14342,17 @@ New usage of "bj-godellob" is discouraged (0 uses).
 New usage of "bj-imim2ALT" is discouraged (0 uses).
 New usage of "bj-inftyexpidisj" is discouraged (3 uses).
 New usage of "bj-inftyexpiinv" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@nmegill : I checked that the following OLD theorems can be deleted safely (the proofs do not have special properties or offer additional insight which makes them worth keeping).  Most, if not all, are shortenings due to @wlammen.  If you agree, I can remove them in this or my next PR:

CAN BE DELETED:
a2andOLD, nbiorOLD, otnldOLD, rbaibOLD, rbaibrOLD, rnlemOLD, bianirOLD
nancomOLD, nanOLD, nanbiOLD, nanbiOLDOLD
xorassOLD, xorneg1OLD, xorneg2OLD, truanOLD, dfnotOLD
stoic1aOLD
eximOLD, aleximiOLD, 19.35OLD, 19.21vOLDOLD, 19.21vOLD, stdpc5vOLD
ax12vOLD (essentially same proof)
19.8aOLD
nexdvOLD (requires more axioms)
19.9tOLD, 19.9hOLD, 19.9dOLD, 19.9OLD (essentially same proofs)
19.23vOLD, 19.36vOLD, 19.36aivOLD (require more axioms)
sb8euOLD (essentially same proof)
mopickOLD
2moOLD (essentially same proof)
2eu1OLD
2eu4OLD
2eu6OLD
euequ1OLD (requires more axioms)

nmegill:
> @nmegill : I checked that the following OLD theorems can be deleted safely (the proofs do not have special properties or offer additional insight which makes them worth keeping). Most, if not all, are shortenings due to @wlammen. If you agree, I can remove them in this or my next PR:

Some of these OLDs are not yet a year old, e.g. ax12v vs. ax12vOLD, which was obsolete as of 8-Dec-2019.  I think @wlammen's next PR #1785 removes the right OLDs (I didn't check in detail).

benjub:
> > @nmegill : I checked that the following OLD theorems can be deleted safely (the proofs do not have special properties or offer additional insight which makes them worth keeping). Most, if not all, are shortenings due to @wlammen. If you agree, I can remove them in this or my next PR:
> 
> Some of these OLDs are not yet a year old, e.g. ax12v vs. ax12vOLD, which was obsolete as of 8-Dec-2019. I think @wlammen's next PR #1785 removes the right OLDs (I didn't check in detail).

Ok. Then it's ready to merge. Maybe save the list somewhere ? So that when their time comes, one does not have to carefully check all the proofs again.

nmegill:
It looks like GS didn't do a rewrap in his last PR, so when BJ did it, a 6000-line diff resulted that is somewhat confusing.  I would guess that this also puts an unnecessary disk space burden on the git tracking (although I don't know the details of how it stores diffs).  So maybe we should implement the suggestion in the 1st paragraph of https://github.com/metamath/set.mm/pull/1779#issuecomment-678657876 to have Travis do a local rewrap ('write souce.../rewrap' and 'save proof */c/f') and fail if it is different from the submitted set.mm.

@david-a-wheeler Could you put this on your list to look into this when you have some time?  I don't think it's high priority but it would make it easier to see the real changes.  set.mm is probably the only db we care about initially; the others haven't been a problem.  Thanks.

nmegill:
> > Some of these OLDs are not yet a year old, e.g. ax12v vs. ax12vOLD, which was obsolete as of 8-Dec-2019. I think @wlammen's next PR #1785 removes the right OLDs (I didn't check in detail).
> 
> Ok. Then it's ready to merge. Maybe save the list somewhere ? So that when their time comes, one does not have to carefully check all the proofs again.

Having a separate list means it has to be kept updated, which probably won't be done consistently because people forget.

Maybe I'm misunderstanding, but checking the proofs isn't necessary.  All that matters is the date in the "Obsolete..." comment and whether a year has elapsed.

Instead, we've been keeping the the "Obsolete..." comment formatted so that it ends with "dd-mmm-yyyy. " with a space after the period.  Since this format is otherwise very rare in comments, we can search for e.g. "-Jul-2019. " to identify all the ones that expired in July 2020.  Of course we have to check visually that it really is an "Obsolete..." comment, but I've done that many times in the past and it takes very little time and effort even though it's a manual procedure.

PR #1785, opened 2020-08-23
remove outdated OLD theorems

wlammen:
Two of them had no "Obsolete as of..." comment, so I checked out the develop version of 20-Aug-2019 and verified they were present in the file at that time.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -9887,7 +9887,6 @@
 "mplelbasOLD" is used by "mplelsfiOLD".
 "mplelbasOLD" is used by "mplsubrglemOLD".
 "mplelsfiOLD" is used by "coe1sfiOLD".
-"mplelsfiOLD" is used by "evlslem6OLD".
 "mplsubglemOLD" is used by "mpllsslemOLD".
 "mplvalOLD" is used by "mplbasOLD".
 "mpv" is used by "mulcompr".
@@ -13226,9 +13225,7 @@
 "suppssOLD" is used by "psrbaglesuppOLD".
 "suppssOLD" is used by "psrlidmOLD".
 "suppssOLD" is used by "psrridmOLD".
-"suppssfvOLD" is used by "evlslem6OLD".
 "suppssof1OLD" is used by "psrbagev1OLD".
-"suppssov1OLD" is used by "evlslem6OLD".
 "suppssov1OLD" is used by "suppssof1OLD".
 "suppssrOLD" is used by "cantnflem1OLD".
 "suppssrOLD" is used by "cantnflem1dOLD".
@@ -15929,7 +15926,6 @@ New usage of "euequ1OLD" is discouraged (0 uses).
 New usage of "euexALT" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
 New usage of "evlslem4OLD" is discouraged (0 uses).
-New usage of "evlslem6OLD" is discouraged (0 uses).
 New usage of "ex-gt" is discouraged (0 uses).
 New usage of "ex-gte" is discouraged (0 uses).
 New usage of "ex-natded5.13" is discouraged (0 uses).
@@ -17179,7 +17175,7 @@ New usage of "mplbasOLD" is discouraged (1 uses).
 New usage of "mplcoe2OLD" is discouraged (0 uses).
 New usage of "mplcoe3OLD" is discouraged (0 uses).
 New usage of "mplelbasOLD" is discouraged (3 uses).
-New usage of "mplelsfiOLD" is discouraged (2 uses).
+New usage of "mplelsfiOLD" is discouraged (1 uses).
 New usage of "mpllsslemOLD" is discouraged (0 uses).
 New usage of "mplsubglemOLD" is discouraged (1 uses).
 New usage of "mplsubrglemOLD" is discouraged (0 uses).
@@ -18428,9 +18424,9 @@ New usage of "supmaxOLD" is discouraged (0 uses).
 New usage of "supmaxlemOLD" is discouraged (1 uses).
 New usage of "suppss2OLD" is discouraged (18 uses).
 New usage of "suppssOLD" is discouraged (12 uses).
-New usage of "suppssfvOLD" is discouraged (1 uses).
+New usage of "suppssfvOLD" is discouraged (0 uses).
 New usage of "suppssof1OLD" is discouraged (1 uses).
-New usage of "suppssov1OLD" is discouraged (2 uses).
+New usage of "suppssov1OLD" is discouraged (1 uses).
 New usage of "suppssrOLD" is discouraged (28 uses).
 New usage of "supsr" is discouraged (1 uses).
 New usage of "supsrlem" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
It looks like this one has a 6000-line diff that reverts back to GS's unwrapped version.  I guess I should have merged this one first.  Oh well, the next PR should resolve that with another 6000-line diff. :)

wlammen:
There are a LOT of outdated OLD theorems in Main, e.g. the old Cantor stuff from Mario, dating from May to July 2019, that are bound for deletion now.  Unfortunately, many are still referenced, sometimes even in mathboxes, and their comments often differ from the preferred "Obsolete proof ... as of..." style, or even have no such hint at all.  Fortunately, git allows checking out an old version of the branch develop by "git checkout develop@{2019-08-20}" for example, and you can load set.mm into an editor then.  At the same time, I open the web page of the OLD theorem, say from us2.metamath.org.  If the editor has the OLD theorem present, it existed 1 year ago, and a quick glance at the proof vs. referenced theorem list gives an overall idea whether meanwhile something has changed in the proof, that mandates keeping it longer.  In any case, I restore the current state by "git checkout develop" again.  If the OLD proof's lifetime has expired, I go through the typical cycle, check discouraged, it has no references, remove it from set.mm, run a verification to see the file is in a consistent state, update discouraged, and commit/push the change.  The latter step is necessary, otherwise you might lose some of your work, when looking at the past year's version again for the next candidate.  I worked my way back from the end of Main, as chances are, OLD theorems are not referenced there (a lot).

This is a tedious process, and I won't do many deletions a day, let alone all of them.

git blame is another tool to see  when a line has changed last.  This time I worked the above described way.

wlammen:
I merged the current set.mm into my personal copy, and run rewrap. It caused just one change, which will not cause a lot of trouble.

 $( The binary Goldbach conjecture is valid for all even numbers less than or
-     equal to 4 x 10^18, see result of [OeSilva] p. ?.  Temporarily
-     provided as "axiom".  (Contributed by AV, 3-Aug-2020.) $)
+     equal to 4 x 10^18, see result of [OeSilva] p. ?.  Temporarily provided as
+     "axiom".  (Contributed by AV, 3-Aug-2020.) $)
   ax-bgbltosilva $a |- ( ( N e. Even /\ 4 < N /\ N <_ ( 4 x. ( 10 ^ ; 1 8 ) ) )
                         -> N e. GoldbachEven ) $.

nmegill:
For the ones without an "Obsolete" comment in the right format, we can just take a year-old set.mm and see if the OLD is there, and if so delete it from the current set.mm.  Or if we want to be extra cautious, we can add "Obsolete....as of <today>" so that in a year they will be deleted.

There are about 200 OLDs that are used by other proofs.  I didn't look in detail but I imagine almost all of them are used in proofs of other \*OLDs, in which case deleting them backwards will prevent problems.

If a \*OLD is used by a non-OLD theorem, that is a bad thing and needs to be corrected.  I think these should be relatively rare.

If a non-OLD has a ~ xxxOLD label reference in its comment, that is bad and needs to be corrected.  I would guess that typically it would just be a matter of removing the 3 letters "OLD".  Probably what happened in these cases was that a global substitution to change the label to \*OLD was done blindly.

So there is a lot of work needed to weed out all of the cases where \*OLD has been abused, but it's not urgent and can be done over a long period of time.

nmegill:
> I merged the current set.mm into my personal copy, and run rewrap. It caused just one change, which will not cause a lot of trouble.

I wonder what caused the > 6000 change lines in this PR.

wlammen:
Instead of going back and forth in git, one could have done this just once, saved a copy of set.mm under a different name, and used this all the time.  Damn it. Time to go to bed.

nmegill:
A large number of the \*OLDs are explained by the work of @TjWhale (Jon Pennant) identifying candidates for deletion in the deprecated sections.  See this thread:
https://groups.google.com/g/metamath/c/OdIB-vF-Qo0/m/pB6D9_9xAQAJ
The were marked "Obsolete as of..." around March 2020, so they won't be deleted until next year.  It makes sense that many of these \*OLDs would be used by later \*OLDs, since the extensible structure version cannot be plugged into the old uses (unlike the case of a proof shortening).  When it comes time to delete them, we will just work backwards to prevent missing \*OLD references in proofs.

There are currently a total of 469 \*OLDs, of which 64 don't have "Obsolete..." in their comments.  These 64 will have to be studied individually, and we should either add "Obsolete..." to their comment, or if they aren't really meant to be deleted, then they should be renamed to something else, maybe \*ALT in some cases.  In order of appearance in set.mm, the 64 theorems requiring manual attention are:

iunonOLD dfsup2OLD ltneOLD ledivmul2OLD ltdiv2OLD cmndOLD lidlssOLD
cmetuOLD df-metuOLD ctpsOLD df-topspOLD eltopspOLD tpsexOLD istpsOLD
istps2OLD istps3OLD istps4OLD istps5OLD metuvalOLD metustelOLD
metustssOLD metustrelOLD metusttoOLD metustidOLD metustsymOLD
metustexhalfOLD metustfbasOLD metustOLD cfilucfilOLD metuustOLD
cfilucfil2OLD metuelOLD metustblOLD metutopOLD xmsuspOLD metucnOLD
retpsOLD cfilucfil3OLD cfilucfil4OLD ghgrplem1OLD shmulclOLD rmoxfrdOLD
rmo4fOLD funcnvmptOLD ogrpinvOLD xrge0tmdOLD isrnsigaOLD measdivcstOLD
probfinmeasbOLD axtglowdim2OLD axtgupdim2OLD cgcdOLD df-gcdOLD ee7.2aOLD
bj-nalnaleximiOLD bj-nalnalimiOLD an43OLD ceqsex3OLD ceqsex3vOLD
latmassOLD pwfi2f1oOLD pwfi2enOLD sbc3orgOLD sbcssOLD

nmegill:
Looking at the very first one on the list of 64, we have:
```
MM> sh st iunon*/c
"The indexed union of a set of ordinal numbers ` B ( x ) ` is an ordinal
       number.  (Contributed by NM, 13-Oct-2003.)  (Revised by Mario Carneiro,
       5-Dec-2016.)"
$d x A $.
25290 iunon $p |- ( ( A e. V /\ A. x e. A B e. On ) -> U_ x e. A B e. On ) $=
      ... $.
-------------------------------------------------------------------------------
"The indexed union of a set of ordinal numbers ` B ( x ) ` is an ordinal
       number.  (Contributed by NM, 13-Oct-2003.)  (Revised by Mario Carneiro,
       5-Dec-2016.)  (New usage is discouraged.)
       (Proof modification is discouraged.)"
$d x A $.
25291 iunonOLD.1 $e |- A e. _V $.
25292 iunonOLD.2 $e |- B e. _V $.
25293 iunonOLD $p |- ( A. x e. A B e. On -> U_ x e. A B e. On ) $= ... $.
MM> sh us iunonOLD
Statement "iunonOLD" is directly referenced in the proofs of 4 statements:
  alephon cfsmolem hsmexlem5 inar1
```
Here the OLD and non-OLD are incompatible, and the proofs of alephon etc. will have to be modified to use the non-OLD.  There may be quite a few of these kinds of situations among the 64 \*OLDs.  Most should be straightforward and could even serve as a way for a new user to become familiar with mmj2/metamath.exe and the PR submission process, while helping us out at the same time.

PR #1786, opened 2020-08-25
Move "shorter proof" discussion into conventions

david-a-wheeler:
Move the discussion text about shorter proofs into
"conventions".


david-a-wheeler:
@nmegill @digama0 - This is a very good sign. I created this new pull request, and Travis jobs were immediately kicked off as intended.

Hopefully this means that Travis is working again!

david-a-wheeler:
@nmegill - I don't know if you want to move this "shorter proof" text into the conventions section. I think it'd make sense to do so (since it's really just another convention), but you make the call!

benjub:
The problem is that, as I wrote in https://github.com/metamath/set.mm/pull/1776#issuecomment-678488815, this convention contradicts the choice that was made regarding equsex/equsexALT in that PR (that choice contradicts the sentence "If two proofs have the same number of compressed proof steps, we will typically give preference to the one with the smaller number of different labels").  In this special case, I do not care which is the ALT one (although I do care that both be in set.mm).  However, I think it is important to make the rule precise and unambiguous, which is currently not the case.  I'll try to propose a draft soon.

david-a-wheeler:
I agree it's important to make the rule as an ambiguous and precise as we can. This particular pull request didn't change the text, I was just trying to make it more visible. Now that it is more visible, we can improve it further where everyone can more easily see it.

benjub:
Following my previous comment, I drafted a proposal below.  The precise criteria, and their ordering, can be discussed collegially (maybe on the discussion group?) and ultimately decided by @nmegill.  What is important is that they be precise and unambiguous, which is not currently the case.

Among the criteria below, some are more easily visible on the MPE webpages and others are more easily visible directly on set.mm or via the Metamath program.  Therefore, someone who uses more the MPE website would put emphasis on different criteria than someone using more set.mm.  For instance, I would put the criterion on essential steps before the criterion on total steps, while it currently comes after it and after the criterion on distinct labels.  So I'm open to your proposals, including the proposal to post it on the discussion group.  (I'm also tagging @digama0, @david-a-wheeler, @wlammen). 

Note: after checking dictionary.com, "alternate" and "alternative" seem to be both acceptable, so I used the shorter word, but correct me if I'm wrong (as well as for other English mistakes).

Here is a proposal of what could go in the ~conventions comment in set.mm:


<li><b>Preferred proofs and alternate proofs.</b>
A $p-statement generally has many different proofs.  We generally keep a single "preferred" proof in set.mm, and sometimes also keep alternate proofs, which are labeled by suffixing ALT (or ALT2, ALT3, etc.) to the label of the preferred proof.  Here is how we decide which proof to keep as the preferred one (this is a strict rule) and when to keep alternate proofs (this rule is more loose and open to some subjectivity).  In the following, "axiom" means "non-syntactic axiom which is not a definition."

In determining the preferred proof, each criterion has precedence over the next one (i.e., one looks at the next criterion only if there is a "tie"):
1. The proof which uses the fewest axioms, and in case of a tie, the proof using the earliest axioms (e.g. a proof using ax-mp,1,2,3,gen,6 would be preferred over a proof using ax-mp,1,2,gen,4,5).
2. The proof with the fewest steps (this is the length of the block of capital letters in the compressed proof format).
3. The proof with the fewest number of labels (including syntactic labels).
4. The proof with the fewest essential (i.e., non-syntactic) steps (this is the number of lines in the proof displayed on the MPE webpage).
5. The proof with the fewest number of syntactic labels.
6. The proof which is represented by the shortest string in the compressed format.
7. The proof which comes first alphabetically in the compressed format.

Here is when to keep alternate (ALT) proofs.  A single one of these criteria is sufficient to keep an alternate proof (even if all other criteria fail):
1. A proof avoiding an axiom that the preferred proof uses (even if it uses more axioms in total and is longer).
2. A proof with fewer steps than the preferred proof.
3. A proof with fewer essential steps than the preferred proof.
4. More loosely, a proof which adds insight to the preferred proof, for instance by avoiding certain defined terms and syntax (a proof using only implications and conjunctions, or universal quantifiers, when the preferred proof mixes conjunctions and disjunctions, or universal and existential quantifiers), or by being more "understandable", or by its particular proof structure, etc.
5. Most probably, a proof which is tie with the preferred proof for the first four criteria above (unless it differs anecdotally from the preferred one).

Alternate proofs should have the "(Proof modification is discouraged.)" and "(New usage is discouraged.)" tags, together with an explanation of why they are kept as alternate proofs.  Their statement should be identical to the statement of the preferred proof, including its disjoint variables conditions.  The comment of the preferred proof should link to the alternate ones (e.g., "See also ~~ xxxALT and ~~ xxxALT2 ." ) and the comment of the alternate ones should begin with "Alternate proof of ~~ xxx".

In particular, before "obsoleting" a proof, the contributor should check whether the above applies, in which case the proof should not be obsoleted but kept as an alternate (ALT) proof.  If there is any doubt, advice should be sought from other contributors before obsoleting a proof.

One could trick these rules by adding various lemmas, making the final proof shorter.  Therefore, the previous criteria should be understood "the rest of the database being unchanged, except possibly by reordering statements other than axioms."
</li>

<li><b>Adding new statements.</b>
About any reasonable statement can be added to a contributor's mathbox.
Adding a statement to the main part of set.mm is more restricted.  Here are a precise rule and two less precise rules as to what can be added, while the final decision is up to a maintainer of set.mm (Norman Megill and Mario Carneiro).

1. A statement whose addition permits to shorten enough proofs of other statements so that the total number of steps saved from other proofs is at least equal to the number of characters used to state and prove the added statement (more precisely: the number of characters within the added $d, $e and $p statement(s), plus the total proof length in compressed format).
2. A statement which is interesting enough, in particular a statement which appears in the peer-reviewed published literature.
3. A group of statements which formalize or expand a well-developed area of mathematics.

Before beginning work that could fulfill 2, or even more importantly 3, it is recommended to seek advice on the discussion group to determine which approach is more suitable.

The addition of an $a-statement is rare: it is either the addition of a "commonly accepted" definition (conforming to the soundness rules stated below) and its associated syntactic axiom, or even more rarely the addition of an axiom.

Note that after the addition of a new statement, especially a statement satisfying the first criterion, an "alternate proof" may become the "preferred proof" (see above) and the preferred one should become ALT (or maybe OLD).  Of course, this cannot be done instantly, but these are the rules to which we try to adhere.
</li>

digama0:
While I think it is useful to have a rigorous definition of what criteria take precedence for determining the non-ALT version of the proof, I think by the time you get halfway down that list it should be "this is a judgment call". I have difficulty imagining a scenario where we need a definition to this level of detail.

The rule on axiom usage in particular seems dubious. As long as the axioms of T1 are contained in the axioms of T2, T1 should be preferred, but if the axiom sets are incomparable then it should be a judgment call, which depends on which axiom set fits better into a subset that is preferred for study. I don't think these sets can necessarily fit strictly into a lexicographic ordering.

benjub:
@digama0 I agree with both paragraphs (although your first sentence seems to lack a verb?), and part of me wonders if what I wrote above is a bit overkill.  However, there are also advantages in removing "judgement calls", and since we can render the rules totally objective and unambiguous, maybe the few extra lines required to state them are worth it? Let's hear other opinions.

nmegill:
@benjub I also think this is overkill, but I won't object to including it.  We can always change it later.  Maybe there should be a statement above it that these are suggestions for resolving borderline cases and are subject to judgment calls.

Note that a $p or $a with a \*ALT label is required to have both discouragements by 'verify markup', but \*ALT2, \*ALT3,... are not official, and I currently don't have any plans to change 'verify markup' to recognize these also.  I think it would or should be very rare to have more than one \*ALT.  With too many different versions of a proof, we are getting into the territory of overly specialized variants that most people wouldn't be interested in, and they should probably be treated as experiments that live in a mathbox.  That would also help protect them from casual use by others.  BTW there is nothing wrong with referencing a mathbox theorem in a comment in the main set.mm, and the \*ALT comment could mention the ~\*ALT2, ~\*ALT3,... variants for the interested reader.

There is a danger of having so many conventions and rules that that the user becomes overwhelmed and doesn't read them thoroughly or remember them well.  Like we did for ~conventions-label, maybe we should break out these rules into ~conventions-proof.

@david-a-wheeler there is a reference to id1 in the new text that should be ~ id1.

wlammen:
As far as I am concerned, I can live with the current Acceptable Shorter Proof rules.  This means the equsexALT proof becomes the main (and IMHO only) proof.  It inherits from the previous proof all attributions, extended by some (Proof shortened/revised by BJ) tag.  

If we decide to revise the rules, I am a tiny bit inclined to give the syntactic labels a less important role, because they don't impact the visual layout of a proof on the web page, by which I judged proofs before.  But in the end we talk here about rare corner cases that do not influence the overall impression of how proofs are presented, so I will agree with the above rules as well.

I have no particular idea about when to create, and to what degree, alternate proofs.  From a beginners view, I would say less is more.  Actually, the alternate proof idea hides the necessity of some kind of modularization of the database, where users pick their favourite plug-in.  I understand this means a lot of work, though, and that this is not feasible in short term.

@david-a-wheeler Nice to have the rules in a place where I can find them more easily than this time.

benjub:
Thanks for your comments. I hope to do the PR this weekend.
* I will remove "5. fewest nonsyntactic labels", to simplify a bit (in practice, the first three items will probably break most ties anyway).
* I will add that having more than one ALT proof is very rare. @nmegill, would the form ~ xxxALTALT instead of ~ xxxALT2 allow "verify markup" to check that both discouragement tags are there ?
* @nmegill is right about the danger of overwhelming the newcomer. I'll move the text to ~conventions-proof, and more generally, over time, we'll have to be clearer about which conventions are more important, which are directed to users and which to contributors.
* Like @wlammen, I would prefer to put more emphasis on essential labels/steps, for instance by switching Items 2 and 3. But for the moment, I keep the rules unchanged. I'm open to your proposals.

On a related note: @nmegill, you wrote that there's nothing wrong with referencing a theorem in a mathbox from a comment in the main part. But mathboxes may vary more often, while the main part is designed to be more stable. Isn't it a bit problematic?

nmegill:
> * I will add that having more than one ALT proof is very rare. @nmegill, would the form ~ xxxALTALT instead of ~ xxxALT2 allow "verify markup" to check that both discouragement tags are there ?

All that "verify markup" cares about is that the last 3 letters are "ALT", so "ALTALT" would be checked for discouragement tags.  That doesn't mean that I would like to see ALTALTALT, ALTALTALTALT,... etc. For very rare things you can just maintain the tags by hand, without depending on "verify markup" to enforce them.  There are many other non-ALT situations where we have discouragement tags for various reasons, and we don't depend on "verify markup" to enforce them.  Instead we hope that the user pays some attention to the "discouraged" file.

> * Like @wlammen, I would prefer to put more emphasis on essential labels/steps, for instance by switching Items 2 and 3. But for the moment, I keep the rules unchanged. I'm open to your proposals.

I don't have a preference, so if you two agree then go ahead.

> On a related note: @nmegill, you wrote that there's nothing wrong with referencing a theorem in a mathbox from a comment in the main part. But mathboxes may vary more often, while the main part is designed to be more stable. Isn't it a bit problematic?

After I submitted the comment I thought it might have been a mistake to say "nothing" wrong, which sounds like I am encouraging such things.  I guess I will amend that to say that it's better to avoid it but sometimes it can be useful or helpful or interesting for the reader, in which case it's acceptable.

Two things prompted my original remark.

First, "verify markup" will immediately identify a missing label if someone changed their mathbox, so the mathbox user will have to fix the comment in the main part for the PR to be accepted.  It's not like such errors will accumulate unnoticed and slowly degrade the set.mm quality, which might have happened before "verify markup".

Second, I recently moved the obsolete ~ax-c4 through ~ax-c16 into my mathbox because they don't really belong in the main set.mm and just cause confusion there.  Not only are they referenced frequently in the main set.mm, they are even referenced in mmset.html.  And this hasn't bothered me too much.  I think the references are informative and shed some light how our pred calc axioms evolved from the ones in my 1995 paper that set.mm started off with historically.

PR #1788, opened 2020-08-27
The Laplace Expansion of the matrix determinant

tirix:


tirix:
I believe the Laplace expansion of the determinant is important and reusable enough to be moved to the main part of set.mm, but for now I did it in my mathbox.

If we choose to move it to the main part, then the fact that it uses a different notion of submatrix than the one introduced by @avekens may be a problem. Do we want two different definitions of submatrix, coexisting ?

avekens:
My definition of a submatrix is intentionally very general, related to the definition of a matrix. This means the indices are arbitrary finite sets, not necessarily integer ranges  ( 1 ... N ) ( or ( 0 ..^N )?). The definition of a matrix was provided by Stefan O'Rear already in 2015. This approach has some advantages, but I also had the problem to prove the Laplace expansion in general, based on this definition. Fortunately, only a special case of the Laplace expansion was necessary to prove Cramer's Rule, see ~smadiadetr. 

Working with the new definition of a submatrix requires a matrix with an integer range as index set  ( A = ( ( 1 ... N ) Mat R ) ). I do not know if this will be a problem (I don't think so). 

For me it would be no problem two have two kinds of submatrices, so one can select the variant which is most appropriate for the current need. There should be, however, theorems to transform one representation into the other, and showing the relationship between/equivalence of both representations.

Regards,
Alexander

BTW: bj-disjdif is not from me, it's from BJ (please correct the entry in Recent label changes)

benjub:
> BTW: bj-disjdif is not from me, it's from BJ (please correct the entry in Recent label changes)

Thanks for noticing. I'm alway glad when a theorem of mine finds some use!

tirix:
> For me it would be no problem two have two kinds of submatrices, so one can select the variant which is most appropriate for the current need. There should be, however, theorems to transform one representation into the other, and showing the relationship between/equivalence of both representations.
> 
Ok, I also think if it is properly documented it shall not be too much a source of confusion.
@norm, would you also agree that I move the Laplace expansion of the determinant from this PR to the respective sections of the main part of set.mm?

> integer ranges ( 1 ... N ) ( or ( 0 ..^N )?). 
> 
Indeed that's a good point! I remember there was much discussion about NN vs. NN0. Words are indexed with ` ( 0 ..^ N ) `. Fox matrices, it seems everywhere in Mathematics, indexing starts at one, and some programming languages use zero, while some other programming languages use one (like R). I believe we shall use indexes starting with one for matrices, as that way the proofs will more closely follow literature. I could still change however if there is work associated.
@norm @avekens what do you think? Shall I raise the question on the mail discussion group?

I would like to define a "literal" matrix function which would be used like this to transform words (for any matrix size) :
``` ( litMat ` <" <" A B C "> <" D E F "> "> ) ```
This would allow to easily write down given matrixes if required in proofs. It would also handle the index conversion (from zero-based words to one-based matrices).

> BTW: bj-disjdif is not from me, it's from BJ (please correct the entry in Recent label changes)
>
I'll fix it, thank you for noticing!

nmegill:
> Ok, I also think if it is properly documented it shall not be too much a source of confusion.
> @norm, would you also agree that I move the Laplace expansion of the determinant from this PR to the respective sections of the main part of set.mm?
> 

I will agree to that.

> > integer ranges ( 1 ... N ) ( or ( 0 ..^N )?).
> 
> Indeed that's a good point! I remember there was much discussion about NN vs. NN0. Words are indexed with `( 0 ..^ N )`. Fox matrices, it seems everywhere in Mathematics, indexing starts at one, and some programming languages use zero, while some other programming languages use one (like R). I believe we shall use indexes starting with one for matrices, as that way the proofs will more closely follow literature. I could still change however if there is work associated.
> @norm @avekens what do you think? Shall I raise the question on the mail discussion group?

I personally would probably be in favor starting at one.  I looked at 3 linear algebra books and they all start at 1.    Still, I think it would be a good idea to get input from the group.

PR #1792, opened 2020-08-29
Start of absolute value in iset.mm

jkingdon:
This is a pass through the first part of the absolute value section from set.mm, from absneg to zabscl . The easy theorems are copied from set.mm or given simple proofs. The trickier ones are mentioned in the missing theorems list.

This pull request stops short of a number of theorems in the absolute value section which are particularly relevant to Cauchy sequences and convergence. As of now, I'm still a bit unclear what to do about them. But that is for another day.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4630,6 +4630,13 @@ This is a contradiction so ` -. ( abs `` A ) < A ` , which by
 <TD>Should be provable once we have proved absmul</TD>
 </TR>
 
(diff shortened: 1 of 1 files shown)


jkingdon:
The travis build failed due to "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself." which I take to just be a network problem or something of the sort. I'll try to re-run it (if I can figure out how).

PR #1793, opened 2020-08-29
remove outdated retpsOLD, metucnOLD, xmsuspOLD

wlammen:
All removed OLD theorems are more than 1 year old and unmodified by now, I checked the version of set.mm in develop@{2019-08-27} .

Other unrelated changes are due to rewrap.

@benjub I am not sure why rewrap rearranged bj-2upln1upl.


benjub:
> @benjub I am not sure why rewrap rearranged bj-2upln1upl.

I don't see it but that shoudn't be a problem. It's probably because it used ~bj-disjdif which was relabeled ~disjdif2

tirix:
Yes, there shall be no problem removing those OLD theorems, they have more powerful versions.

PR #1800, opened 2020-09-02
Cleanup OLD theorems related to the support of functions (Part 3)

avekens:
* Cleanup of outdated OLD theorems related to the support of functions, see issue #809 (final part)


wlammen:
Nice move! I already thought about how to get rid of tsmsrelOLD, and found it not that easy to do. I love to see someone more familiar with the topic address this issue.

PR #1804, opened 2020-09-05
rm more outdated istps*OLD theorems

wlammen:


wlammen:
(AV:  TODO-NM - can this definition (and related theorems) be removed?)

This question was part of df-topspOLD, and is now implicitly answered by yesterday's and today's pull request.  If necessary, the old content can be restored from git's history tracking features at any time in future.

PR #1817, opened 2020-09-11
gsumws{3,4} and amgm{2,3,4}d

spolu:
Deduction form of ~ amgm2 (based on ~ amgmlem instead of a manual demonstration) and new deduction form version of amgm for n={3,4} with supporting ~ gsumws{3,4} theorems.

(If there is interest we can easily go all the way to 8 (after that we a different syntax will be needed))


digama0:
The official line is that we are not limited to 8, but rather if you need more than this you should build a binary tree structure using `concat` (so an appropriate lemma supporting proofs of `concat` nodes is also useful).

benjub:
There is a "from from" in a comment.
Did you run the command from https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md ? It looks like the format differs (in addition to the error detected by Travis).
Did you "MM-PA>minimize *" the proofs ?

spolu:
@benjub I haven’t yet. I will do so.

@digama0 gotcha, I think I see what you mean and will add that in a subsequent PR 👍

tirix:
The Bibilographic checker complains about the comment of amgm2d:
`?Warning: The label token "amgmlem." (referenced in comment of statement
"amgm2d") is not a $a or $p statement label.`
Simply inserting a space between "amgmlem" and the dot shall make it happy (right now it's looking for "amgmlem." with the dot included)

spolu:
(updated)

(I'll make a follow-up PR with more variants and a lemma supporting proofs of concat nodes as suggested by Mario.

PR #1819, opened 2020-09-12
rm ref to ltdiv2OLD in bposlem6 and slightly shorten it

wlammen:
The shortening (two proof lines) is a result of aggressive use of 'improve all', so I do not bother tagging it as my own achievement.
@nmegill During the replay of the original proof in Metamath to get to the position where ltdiv2OLD was used, I made the mistake to not assign the name 'p' to a temporarily used setvar, so there was an orphaned $nnn item with no value attached.  This leads to the surprising result that the proof finishes (all proof lines resolved) without the CONGRATULATION cheer. I guessed the possible cause almost immediately, but there is no automatic hint to the reason of the fail, which might confuse beginners.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -9147,7 +9147,6 @@
 "ltbtwnnq" is used by "nqpr".
 "ltbtwnnq" is used by "reclem2pr".
 "ltdiv2OLD" is used by "basellem1".
-"ltdiv2OLD" is used by "bposlem6".
 "ltdiv2OLD" is used by "perfectlem2".
 "ltdiv2OLD" is used by "sin01gt0".
 "ltdiv2OLD" is used by "sincos6thpi".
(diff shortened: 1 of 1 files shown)


nmegill:
> @nmegill During the replay of the original proof in Metamath to get to the position where ltdiv2OLD was used, I made the mistake to not assign the name 'p' to a temporarily used setvar, so there was an orphaned $nnn item with no value attached. This leads to the surprising result that the proof finishes (all proof lines resolved) without the CONGRATULATION cheer. I guessed the possible cause almost immediately, but there is no automatic hint to the reason of the fail, which might confuse beginners.

I'm a little bit confused.  If there was a $nnn item, the proof isn't complete, so there will be no congrat message, and "show new_proof /unknown/all" should show the missing steps.  It is possible that omitting "/all" will not show anything if all the "essential" steps are there.  Is that what you saw?

PR #1821, opened 2020-09-14
replace ax13b with the more general instance it proves

benjub:
I noticed that the proof of ax13b actually proves a more general instance.  Stating the more general instance:
* gives a more general result (hence more reusable, even though in this case it probably has not much use),
* makes the statement and the proof clearer,
* shortens the proof (more than it lengthens the proof of the theorem referencing it: -12 + 3 = -9 bytes).

By the way: I'm in favor, when a statement in set.mm has a true and more general instance, of proving that more general instance in set.mm (either in addition or in replacement, depending on the cases).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -244,6 +244,13 @@
 "4atex2-0aOLDN" is used by "4atex2-0bOLDN".
 "4ipval2" is used by "ip1ilem".
 "4ipval2" is used by "ipasslem10".
+"4sqlem13OLD" is used by "4sqlem14OLD".
+"4sqlem13OLD" is used by "4sqlem17OLD".
+"4sqlem13OLD" is used by "4sqlem18OLD".
+"4sqlem14OLD" is used by "4sqlem17OLD".
+"4sqlem15OLD" is used by "4sqlem16OLD".
+"4sqlem16OLD" is used by "4sqlem17OLD".
+"4sqlem17OLD" is used by "4sqlem18OLD".
 "4syl" is used by "1stmbfm".
 "4syl" is used by "2ndmbfm".
 "4syl" is used by "2reu5".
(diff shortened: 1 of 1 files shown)


nmegill:
I'm aware that ax13b has a more general form.  Its sole purpose was to show an alternate form of ax-13 that might be easier for the reader to understand, which I added after a complaint that ax-13 was rather non-intuitive.  It wasn't meant to be a general-purpose theorem.  I showed the substitution instance with ax-13 on one side of \<-> and the equivalent on the other side, so the reader can see that immediately without having to perform a mental substitution.  Its comment originally said, "Two equivalent ways of expressing ~ ax-13", which is no longer the case for the more general form because neither side holds in isolation.

Although it found an incidental use in a proof shortened by WL, that wasn't its intended purpose, and in any case it doesn't need the more general form for that purpose.  Your comment, "An equivalence used in the proof of ~ ax13 from ~ axc9", doesn't express the original purpose.  (Also, the fact that it is used in the proof of ~ ax13 is already shown in the usage list, so saying that in the comment is redundant.  And if someone finds another shorter ~ ax13 proof that doesn't use ~ ax13b, how would they know to update the ~ ax13b comment?)

So I am a little torn about this update.

avekens:
I would propose to keep the current ax13b, to add Benoît's version $p |- ( ( -. x = y -> ( y = z -> ph ) ) <-> ( -. x = y -> ( -. x = z -> ( y = z -> ph ) ) ) ) $ (named ax13b0, for example) right before ax13b, and to simplify the proof of ax13b using ax13b0. Additionally, ax-13b could be marked with "(New usage is discouraged.). By this, both intentions could be served.

nmegill:
Since we do show the exact substitution instance in the ax-13 comment, I'll accept your PR if you change the ax13b comment to read, "An equivalence used to show two ways of expressing ~ ax-13 .  See the comment for ~ ax-13 ." (and don't mention the use in the ax13 proof since it is already listed on the web page).

> I'm in favor, when a statement in set.mm has a true and more general instance, of proving that more general instance in set.mm (either in addition or in replacement, depending on the cases).

I think we should approach this on a case by case basis as the need arises.  For theorems that already exist, occasionally there may have been a reason for a less general version.

benjub:
Done, although the sentence sounds a bit strange. Should I write:
> An equivalence used to show two ways of expressing ~ ax-13 are equivalent .  See the comment for ~ ax-13 .
or maybe "are actually/indeed equivalent" ?

PR #1823, opened 2020-09-15
Misc

benjub:
Retrying, following merge conflict in #1821
I can reword the comment to:
> An equivalence used to show that two ways of expressing ~ ax-13 are equivalent .  See the comment for ~ ax-13 .

or maybe "are actually/indeed equivalent" ?


nmegill:
> Retrying, following merge conflict in #1821
> I can reword the comment to:
> 
> > An equivalence used to show that two ways of expressing ~ ax-13 are equivalent .  See the comment for ~ ax-13 .
> 
> or maybe "are actually/indeed equivalent" ?

If you feel strongly it should be reworded one of those ways, go ahead, but it seems redundant.  To me, when one says "two ways of expressing ~ ax-13", it automatically implies that they are equivalent, in the sense that in the presence of the other axioms they have the same consequences.   Does anyone else think otherwise?  What would be an example of "two ways of expressing ~ ax-13" that are not equivalent?

benjub:
What was bothering me with the wording is that a statement can **show** two ways of... But I'm fine with the current version.

PR #1828, opened 2020-09-20
converting wl-mo2dnae into a tautology

wlammen:
This is a definition of df-mo without any distinct variable conditions.  One can achieve a similar result for df-eu by replaying most of the proof and some extra effort.  It remains to be seen whether some slick adjustments can reduce this extra effort and still yield the same result.

I personally prefer starting with df-mo, and later arrive at df-eu as a particular case of df-mo in the eu-section, because it is natural to start with general theorems, and specialize them later.  In set.mm the reverse way is chosen, perhaps because mo is less known than eu.


nmegill:
> I personally prefer starting with df-mo, and later arrive at df-eu as a particular case of df-mo in the eu-section, because it is natural to start with general theorems, and specialize them later. In set.mm the reverse way is chosen, perhaps because mo is less known than eu.

Yes, a formal notation for mo is very rare.  Bell & Machover's E* is the only one I have ever seen.  df-eu was added to set.mm very early, and df-mo a couple of years later.  df-rmo was added 23 years after df-reu (2017 vs. 1994).

I agree that mo is more basic, and in principle it would be more logical to start with mo then build up to eu.  The main objection I see is that eu is more common, and many people will encounter eu without ever needing mo.  If we start with mo, someone wanting to learn eu would first have to learn mo, whereas the way it is now they can learn eu immediately i.e. one less definition to learn.

david-a-wheeler:
I think starting with eu is quite reasonable. I think it's important that set.mm *have* mo, but many people have never seen it.

nmegill:
BTW your ~ wl-mo2tf is a nice theorem, and I have no objection to using it to build up to an eu version.  It is for a very specialized purpose that most people won't encounter, and my comments don't apply to it.

benjub:
Indeed, I too think that it would be more logical to define eu in terms of mo ( E! x ph <-> ( E. x ph /\ E* x ph ) ).

sctfn:
If it makes any difference, my Discrete Mathematics course in college
started with mo and built to eu. A bit informally, but that's the order I
was taught (after encountering Metamath!)

david-a-wheeler:
@sctfn - that does make a difference!

Okay, it's clear that either order makes sense.

Norm, as always, your call.

nmegill:
Personally I prefer to start with E! first (as I already did since that's the way it is now) simply because it's usually one less definition to learn, given that E\* is rarely used.  I don't like having to learn unnecessary definitions if I'm already mentally overwhelmed by new material, and I assumed that some readers might feel that way.

The proposed (E. x ph /\\ E\* x ph) replacement for ~df-eu is theorem ~eu5 and is already mentioned in the ~df-eu comment.  If we wish, we could change the comment to emphasize it and show its formula explicitly if we think that will help the reader.  ~eu5 doesn't appear anywhere in the literature as far as I know.

~df-eu, ~eu1, and ~eu2 are the only definitions for E! that that I could find  in the literature, where they are spelled out with explicit formulas.

The only literature reference with E\* that I'm aware of is Bell & Machover.  They define both E! and E\* with explicit formulas and don't show them expressed in terms of each other.  They introduce E! first.

Given all of the above, if several people still feel strongly enough about defining E! as ~eu5, and someone submits a PR to do that, I'll accept it (if somewhat reluctantly).  However, all I've seen in favor is that E\* is supposedly a more basic concept (although in proofs we sometimes have to prove both the "empty" and "exists 1" cases e.g. ~moexex) and that a professor did it informally in a classroom.

PR #1831, opened 2020-09-24
rm dependency on ax-ext in wl-euequ1; sincos6thpi replace OLD theorem

wlammen:


avekens:
I am not happy with adding "Replace OLD theorem.  (Revised by ..." to the comments of theorems. This information does not have any value for the future. An exception may be that if the proof is shortened by replacing the OLD theorem (not only by removing the 3 letters OLD), the tag "(Proof shortened by ..." could be added.

wlammen:
@avekens I do not add this because it is my invention, but because I learned you should add a short reason why you revise a proof.  Don't ask me where this is written.  If I remember right this shall give an orientation what was going on in the past, so you can more easily check out the right version in case you are interested in old states.  If you think,the wording should be chosen otherwise, well, then I can change this of course, but I want to hear other opinions, too.

As for the shortening, this is a result of using 'improve all' frequently throughout replaying most of the proof.  The proof is in total shortened by less than 0.5% measured in proof bytes, and most important, I did not care nor use my brain to get there, so I feel I should NOT add a 'Proof shortened' tag.  The main motivation for working on sincos6thpi was to free it of an OLD theorem.  This is expressed by the tag.

nmegill:
There is no written policy on this.  I may have suggested somewhere that it is useful to have a comment associated with a "Revised by" tag so we know what was done.  "Revised by" is typically used when the actual theorem is changed (eliminating an unnecessary $d or hypothesis, changing "A e. _V ->" to "A e. V ->", etc.) but there's currently no rule against using it for proof changes.

How about the following compromise.

The tag "Replace OLD theorem. (Revised by ..." may occasionally be useful for the short term, if someone is wondering why the proof changed.  But longer term it doesn't add much value.  Perhaps we can have an informal policy of deleting a "Replace OLD theorem. (Revised by ..." tag after a year has elapsed, like we do for *OLDs.

wlammen:
I am fine with the compromise.  A tag/reason should be added, because the modification appears (and should appear) on the most recent changes page.

benjub:
@wlammen I'm not sure if you've already done it, but you can similarly remove dependence on ax-ext and df-cleq from wl-eudf (and wl-eutf) by using equequ2 instead of eqeq2

wlammen:
@benjub Thanks for the hint, I did not notice it crept into wl-eudf as well.

avekens:
I also agree with Norm's proposal to 
> delete a "Replace OLD theorem. (Revised by ..." tag after a year has elapsed, like we do for *OLDs.

Issue #1805, opened 2020-09-05
Definition for infimum


avekens:
Currently, there is only a definition for the supremum in set.mm. The infimum is "simulated" by using the "reversed supremum" sup ( * , * , ``` ' R ) , which is often hard to understand at the first glance. As discussed in the Google group and supported by Norm, see https://groups.google.com/g/metamath/c/p5XIt_HdDik/m/pg84xtI2AAAJ, I want to provide a separate definition and related theorems (mainly the dual forms of the corresponding theorems for supremum) for the infimum (of a subset of a totally ordered set). Afterwards, definitions/theorems using the pattern sup ( * , * , ` ` ` ' R ) should be revised using the new definition.

(Remark: unfortunately, the symbol for the converse, see ~df-cnv, is not (always) displayed properly by GitHub...)

avekens:
With PR #1806, I provide a separate definition and related theorems (mainly the dual forms of the corresponding theorems for supremum) for the infimum (of a subset of a totally ordered set) in my Mathbox. In addition, I also revised some theorems using the pattern sup ( * , * , ` ' R ) using the new definition. These are currently separate versions of the current theorems, marked by the suffix ALTV.

If there are no objections, I will move the separate definition and related theorems for the infimum to main set.mm, and will replace the current theorems using the pattern sup ( * , * , ` ' R ) by the version using the new definition, removing the suffix ALTV. The former versions of these theorems will be kept as OLD theorems.

By the way, I used the symbol "inf", and also "inf" as abbreviation in labels, It seems, however, that also "infm" was used until now. I think using "inf" is more consistent with the conventions used for the supremum ("sup" and not "supr" or "supm"), Therefore, I would propose to use "inf" consequently in the future.

avekens:
With PR #1806, I provide a separate definition and related theorems (mainly the dual forms of the corresponding theorems for supremum) for the infimum (of a subset of a totally ordered set) in my Mathbox. In addition, I also revised some theorems using the pattern sup ( * , * , ` ' R ) using the new definition. These are currently separate versions of the current theorems, marked by the suffix ALTV.

If there are no objections, I will move the separate definition and related theorems for the infimum to main set.mm, and will replace the current theorems using the pattern sup ( * , * , ` ' R ) by the version using the new definition, removing the suffix ALTV. The former versions of these theorems will be kept as OLD theorems.

[...]

PR #1844, opened 2020-10-09
Remove tauOLD from Jim Kingdon's mathbox in set.mm

jkingdon:
If there is something I misunderstood about either *OLD removal in general, or @avekens 's message in https://github.com/metamath/set.mm/issues/1805#issuecomment-704720076 please speak up.

As far as I am concerned the *OLD ones can go away now. And if someone (in their own branch/copy of set.mm) is relying on tau, I don't think the current tau would work any less well than `tauOLD` (since the definition is the only part that changes; the two taus are used the same way and the statement of taupi - the key theorem - is unchanged).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4877,8 +4877,6 @@
 "df-ssp" is used by "sspval".
 "df-st" is used by "isst".
 "df-subgo" is used by "issubgo".
-"df-tauOLD" is used by "taupiOLD".
-"df-tauOLD" is used by "taupilem2OLD".
 "df-trkg2d" is used by "istrkg2d".
 "df-tru" is used by "tru".
 "df-unop" is used by "elunop".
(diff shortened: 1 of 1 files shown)


nmegill:
You're welcome to remove OLDs from your mathbox.

In general, though, their purpose is to track revisions temporarily until we're sure everyone has caught up.  For example, if someone has been doing local development using the older version, proofs may no longer work after merging (e.g. where a symbol was changed as in this case).  Comparing the OLD version shows what needs to be changed.  This could be done with git history, but the presence of the OLD is an immediate indication that something changed, and it can be convenient to have the OLD available when you are in an editor or want to see a 'show statement' version of the change.

We normally delete OLDs after a year.  This time frame was chosen for convenience (it's easy to match "Oct-2020." next October) but it is arbitrary, and in particular cases they could be deleted earlier if they are interfering with work.  Since they will be purged eventually, they don't cause much harm other than perhaps occasional annoyance, and we have been routinely using them to accompany theorem revisions without thinking too much about it.

jkingdon:
Thanks for the general thoughts on OLD. There seem to be multiple reasons why there isn't any need to wait a year in this case.

avekens:
I think the "1 year rule" is only useful for the main part. Mathboxes should not contain OLD theorems/definitions (unless the mathbox owner thinks it is useful), because definitions/theorems of other's mathboxes should not be used. In the case of infimum, I made changes in other's mathboxes, and I thought it could be useful for the mathbox owners to see the changes (and to use them for their ongoing work, as Norm explained) - so I did not delete the OLD versions (I needed them for the transition anyway).

In general, mathbox owners should be allowed to delete OLD versions anytime.

PR #1849, opened 2020-10-18
reduce axiom usage in axc9lem2

wlammen:
It does not help much otherwise, since nfeqf2 depends on ax-10 and ax-12 via nf2, for example.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19285,7 +19285,6 @@ Proof modification of "bj-equs5v" is discouraged (41 steps).
 Proof modification of "bj-equsal" is discouraged (31 steps).
 Proof modification of "bj-equsalhv" is discouraged (10 steps).
 Proof modification of "bj-equsalv" is discouraged (39 steps).
-Proof modification of "bj-equsalvv" is discouraged (38 steps).
 Proof modification of "bj-equsb1v" is discouraged (17 steps).
 Proof modification of "bj-equsexhv" is discouraged (10 steps).
 Proof modification of "bj-equsexv" is discouraged (37 steps).
--- a/discouraged
+++ b/discouraged
@@ -1442,6 +1442,7 @@
 "axc711toc7" is used by "axc711to11".
 "axc9lem1" is used by "ax6e".
 "axc9lem1" is used by "axc9lem2".
+"axc9lem1" is used by "axc9lem2OLD".
 "axc9lem1" is used by "nfeqf2".
 "axc9lem2" is used by "nfeqf2".
 "axicn" is used by "sineq0ALT".
@@ -14398,8 +14399,9 @@ New usage of "axc5sp1" is discouraged (0 uses).
 New usage of "axc711" is discouraged (2 uses).
 New usage of "axc711to11" is discouraged (0 uses).
 New usage of "axc711toc7" is discouraged (1 uses).
-New usage of "axc9lem1" is discouraged (3 uses).
+New usage of "axc9lem1" is discouraged (4 uses).
 New usage of "axc9lem2" is discouraged (1 uses).
+New usage of "axc9lem2OLD" is discouraged (0 uses).
 New usage of "axcc" is discouraged (0 uses).
 New usage of "axcnex" is discouraged (0 uses).
 New usage of "axcnre" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@wlammen Your ~equsalvw is actually ~bj-equsalvv , with identical proof, it seems. I'm not sure how to make it easier for you to locate this kind of lemma that could prove useful. One way would be to temporarily put your mathbox after mine and use the Metamath option /allow_mathboxes. This is described in https://github.com/metamath/set.mm/issues/1741#issuecomment-667733873 (I hope it works; not tested yet).
You can (or I can) also move these lemmas into Main. I hadn't done it since @nmegill was hesitant to do so. But maybe the current situation changes things? I'm open to your suggestions, although I currently have little time for Metamath.

Anyway, it's always nice to see axiom dependency removals !

wlammen:
I looked for possible equsal variants, but obviously overlooked your version. I will update my pull request and attribute it to you

nmegill:
Since equsalvw is used by axc9lem2 (and reduces axiom usage in the latter, which is always nice), I think it can remain in the main set.mm.  Since @benjub's bj-equsalvv was first (I guess great minds think alike), we should acknowledge him as contributor.  I prefer the label "equsalvw" because we usually have used a "vv" ending to indicate a 2-variable version e.g. 19.41vv.

So @wlammen could you do this:  revise the comment in equsalvw to acknowledge @benjub, delete bj-equsalvv, and add an entry "18-Oct-20 bj-equsalvv equsalvw moved from BJ's mathbox to main set.mm" to the list at the top of set.mm.

Thanks,
Norm

benjub:
Thanks @wlammen and @nmegill . The new comment of equsalvv should read:
> Version of ~ equsal with two dv conditions, which does not require ~ ax-7 , ~ ax-10 , ~ ax-12 , ~ ax-13 .

I also suggest the relabelings:
ax13v --> ax13vv
axc9lem1 --> ax13v
axc9lem2 --> ax13ve (since it is the existential version of ax13v, similar to what ax5e is to ax5)
This would add clarity and consistency, I think, since now axc9lem1 and axc9lem2 are more than lemmas for axc9 (and are not referenced directly by axc9).

**Edit**: And it should be "equsalvv" at the top of the set.mm file.

nmegill:
> Thanks @wlammen and @nmegill . The new comment of equsalvv should read:
> 
> > Version of ~ equsal with two dv conditions, which does not require ~ ax-7 , ~ ax-10 , ~ ax-12 , ~ ax-13 .
> 
> I also suggest the relabelings:
> ax13v --> ax13vv
> axc9lem1 --> ax13v
> axc9lem2 --> ax13ve (since it is the existential version of ax13v, similar to what ax5e is to ax5)
> This would add clarity and consistency, I think, since now axc9lem1 and axc9lem2 are more than lemmas for axc9 (and are not referenced directly by axc9).
> 
> **Edit**: And it should be "equsalvv" at the top of the set.mm file.

"show label \*vv" indicates that the suffix "vv" is almost always used to indicate a two-quantifier version of a theorem, so when I see ~equsalvv I think of something like `( A. x A. y ( ( x = z /\ y = z ) -> ph ) <-> ps )`.  In ~equsalvw, the "w" means "weaker" than ~equsalv, which is a convention we've used in  other places such as ~ax10w - ~ax13w.

Similarly, ~ax13vv would suggest a version with `A. x A. w` in place of `A. x`.

There are 2 conventions for double quantification:  a "2" prefix as in ~2ralbidv and a "vv" suffix as in ~19.41vv.  The "vv" is usually used when the label starts with a number, where a "2" prefix would be awkward (219.41v).  I'm willing to revisit our conventions, but until we have a general renaming plan, I think we should try to keep to the present conventions for individual cases.

The use of a "v" suffix to indicate distinct setvars is rare, but it was what I happened to pick for the weaker ~ax6v.  Maybe ~ax6w (for "weaker") would be better.  But then we'd have a clash if we also renamed ~ax13v to ~ax13w, which is already used for the ~ax10w - ~ax13w series.  Besides, ~ax6v and ~ax13v aren't "really" weaker in the presence of the others since we prove ~ax6 and ~ax13 from them.

When ~ax-13 was added, I imitated the ax-6 -> ax6v naming.  I think we shouldn't rename ~ax13v; it is supposed to be the normal starting point as the ~ax-13 comment indicates (just as ~ax6v is supposed to be the normal starting point for ~ax-6), and it deserves a simple name rather than having a possibly misleading "vv" suffix.  If the number of v's indicated the number of distinct setvar conditions, we'd have a lot of renaming to do. :)

As for
> axc9lem1 --> ax13v
> axc9lem2 --> ax13ve

we have sometimes used "-2" for an unspecified variation, and maybe ax13v-2 and ax13v-2e would be better.

However, overall I don't see this as urgent, and I'm going to accept the PR as is for now, and we can continue the discussion later.

PR #1852, opened 2020-10-20
Misc

benjub:
Typos and bookkeeping; this carries out the modifications of the closed #1821.


benjub:
Travis fails:
> The command "scripts/download-metamath" failed and exited with 3 during .

david-a-wheeler:
Everything succeeded except metamath-exe. I clicked on the details:

~~~~
curl --silent -L -o metamath-program.zip 
The command "scripts/download-metamath" failed and exited with 3 during .
~~~~

I don't see any reason that should fail. Odd.

david-a-wheeler:
I can't seem to restart it (!). I suggest making a trivial pointless change & resubmitting. I think it's a transient problem.

nmegill:
Regarding the hyphens added to "object language":

I believe the following 2 cases are incorrect, because "object language" isn't a compound adjective.
...an underlying "object-language." ...
...variables of the object-language...

The next 2 cases are correct because the compound expression "object language" serves as a compound adjective.
...ranging over object-language individual variables...
...meaning that all object-language instances they...

Generally, you need the hyphen only if the two words are functioning together as an adjective before the noun they're describing. 
https://apastyle.apa.org/learn/faqs/when-use-hyphen

There are exceptions where the hyphen is also used for the noun form, but I don't think this is one of them.

benjub:
@nmegill You're completely right, I knew this but overlooked these two cases in the change.
@david-a-wheeler Thanks, hopefully this new change will put things back in order.

Edit: there are (two?) more cases... next commit soon to fix these.

david-a-wheeler:
@benjub - it worked! Travis seems to occasionally fail in mysterious ways :-(.

PR #1853, opened 2020-10-21
Misc

benjub:
* I removed from a comment "See also ~ snex" since ~snex is already linked in that comment.

* Shortened proof of elpr2 simply by changing order of two statements.

* Substantially shortened proof of axnul by using eximii, and also by using the constant "FALSE". This also makes the proof more readable.

There is an html rendering problem in the page for axnulALT, since in set.mm, there is a "trace_back" which is rendered as a subscript. @nmegill : I'm not sure how to fix this... add a space somewhere?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14453,6 +14453,7 @@ New usage of "axmulf" is discouraged (1 uses).
 New usage of "axmulrcl" is discouraged (0 uses).
 New usage of "axnul" is discouraged (0 uses).
 New usage of "axnulALT" is discouraged (0 uses).
+New usage of "axnulOLD" is discouraged (0 uses).
 New usage of "axpjcl" is discouraged (7 uses).
 New usage of "axpjpj" is discouraged (8 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> There is an html rendering problem in the page for axnulALT, since in set.mm, there is a "trace_back" which is rendered as a subscript. @nmegill : I'm not sure how to fix this... add a space somewhere?

It looks like the code didn't anticipate this situation.  I will think about this and modify set.mm and/or metamath.exe when I have a solution.

benjub:
> It looks like the code didn't anticipate this situation. I will think about this and modify set.mm and/or metamath.exe when I have a solution.

Thanks @nmegill . Maybe a solution is to have "escaped characters", like \\_ to represent an underscore (and therefore \\\\ to represent a backslash) ? And maybe this would be an opportunity to add \\^ for the carret and to make ^ output superscripts, paralleling the behaviour of _ for subscripts.

PR #1854, opened 2020-10-23

avekens:
Main set.mm:
* ~id1 -> ~idALT
* theorems ~fi1uzind, ~brfi1indALT, ~opfi1uzind, ~opfi1ind added
* ~usgraexvlem -> ~usgraexmplvtxlem
* ~usgraexmplef extracted from ~usgraexmpl

* Terminology in comments adjusted
* Header comment of "Vertices and edges" enhanced
* Header comment of "Undirected simple graphs - basics" enhanced
* USLGraph -> USPGraph (introduction of "pseudograph")
* ~usgrop, ~usgredgappr, ~usgrpredgav, ~edgassv2 added
* new section "Undirected simple graphs - examples" (some new theorems, some theorems moved from previous section)
* new section "Undirected simple graphs - finite graphs" (including definition for FinUSGraph)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -8247,6 +8247,7 @@
 "hvsubvali" is used by "pjcji".
 "hvsubvali" is used by "pjssmii".
 "hvsubvali" is used by "pjsubii".
+"idALT" is used by "id1".
 "idcnop" is used by "nmcopex".
 "idcnop" is used by "nmcoplb".
 "idhmop" is used by "hmopidmch".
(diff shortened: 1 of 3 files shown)


nmegill:
The problem with ~id1 is that it is referenced externally in a number of places, which is why it hasn't been changed.  mmset.raw.html and mmmusic.html, at least, will have to be updated.  I think there are some slide presentations referencing it as an introductory example.

Here's what I suggest you do, if we want to move forward with this change:
1. Rename ~id1 to ~idALT in mmset.raw.html and mmmusic.html.
2. Add ~id1 to set.mm with the comment "Alias for ~ idALT that may be referenced in some older works."  You can use yourself as the contributor if you wish, I don't care.  The proof for id1 should be one step referencing ~idALT. 
3. Both ~idALT and ~id1 should have new usage & proof mod. discouraged.

benjub:
I like @nmegill's compromise, and hopefully after a transition period, we'll be able to retire ~id1.  For the record, the label ~idALT had already been proposed in #1098 and #1745.

avekens:
I made the changes as proposed by Norm.

In mmmusic.html, however, I did not change the file name id1.mid ...

nmegill:
> In mmmusic.html, however, I did not change the file name id1.mid ...

I'll take care of that.  I might not rename it since the old version of set.mm is needed to reproduce those files, and "id1" provides a good hint as to what the old name was.

Issue #1863, opened 2020-10-31
Proximity of "Bundled" and "Bounded"


jkingdon:
From Gérard Lang:

> There is a rather great proximity between the english words « bundled » and « bounded », so that the current name for the axiom of Bundling in iset.mm is « ax-bnd », when Benoît Jubin’s Mathbox uses many times axioms regarding bounded notiions, whose notation is « ax-bd0,…,ax-bdsb » (in the section 5.3.4), « ax-bdsep » (in the section 5.3.6) or even « ax-bdsetind » (in the section 5.3.8.1).
> In my personal opinion, this could be a source of confusion, so that I suggest that the  notation concerning the axiom of Bundling could be changed from « ax-bnd » to ‘« ax-bndl ».

benjub:
Gérard's proposal (ax-bndl) sounds good to me. It will also distinguish it from the axiom of boundedness (a variant of Collection), since the latter has no "L".

jkingdon:
> Gérard's proposal (ax-bndl) sounds good to me. It will also distinguish it from the axiom of boundedness (a variant of Collection), since the latter has no "L".

I have no objection.

david-a-wheeler:
I like it, making names less confusing is a good thing.

benjub:
Done (see #1864), so the issue can be closed (I do not know how to close it).

david-a-wheeler:
Resolved with #1864.

PR #1862, opened 2020-10-30
id1

benjub:
Most diffs are due to "MM>save proof */compressed" or the script from CONTRIBUTING.md.
The only "manual" changes in nf.mm are the replacements of the six uses of id1 with uses of id.
Also, one forgotten id1 --> ~ idALT in mmset.raw.html

This is a followup to the introduction of idALT in set.mm (in #1854 by @avekens).  I did a google search "site:us.metamath.org id1" and spotted those.  Also "id1" appears in:
* some slides by @nmegill (http://us.metamath.org/downloads/ihp2014mm.pdf).  I leave it to him if he wants to update to idALT.
* the Metamath book.  Maybe @david-a-wheeler can make a global change to idALT ?
* iset.mm and nf.mm.  Maybe use idALT there too ?  And maybe keeping the alias "id1" is unnecessary since there are fewer external links to these databases ?
* an older entry in mmnotes.txt
* the mmbiblio.html of mpe, ile, nfe.  Are these files automatically generated ?
* other occurrences ?

PR #1864, opened 2020-11-01
Id1

benjub:
Followup of #1862 and implementation of #1863

@nmegill Is it worth creating a "nf-discouraged" file ?


nmegill:
> @nmegill Is it worth creating a "nf-discouraged" file ?

I don't think it provides an advantage at this point, since SF is essentially the only contributor (not counting the FOL borrowed from set.mm).  I would wait until there is more activity with it.

PR #1870, opened 2020-11-08
Misc

benjub:
@nmegill I propose to move ~bj-termab to Main (relabeled ~abid1, proved from ~abbi2i, as proposed in the comment of ~bj-termab), for the reasons stated in the comment of ~bj-termab.
Actually, I suspect that ~abid1 was present in set.mm at some point in the past, and was removed for some reason.  I could restore the original attribution if you indicate it to me. I'll also adapt the comment.

### DIFF
--- a/_frege_A.svg
+++ b/_frege_A.svg
@@ -1,0 +1,31 @@
+<?xml version="1.0" standalone="no"?>
+<svg width="80" height="32" version="1.1" xmlns="http://www.w3.org/2000/svg">
+  <desc>Alpha</desc>
+  <!--
+      size = 16 ;
+      tx = 4 = size / 4 ;
+      ty = 4 = size / 4 ;
+      x1 = 8 = size / 2 ;
+      x2 = 32 = x1 + (3/2) size ;
+      x3 = 56 = x2 + (3/2) size ;
+      width = 80 = x3 + max((3/2) size, size + 2 tx ) ;
+      y1 = 16 = size ;
+      height = 32 = y1 + size ;
+  -->
+  <style>
+	.arc { fill: black; stroke-width: 0.75px; stroke: black; stroke-linecap: round; stroke-linejoin: round; }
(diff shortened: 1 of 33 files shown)


nmegill:
> @nmegill I propose to move ~bj-termab to Main (relabeled ~abid1, proved from ~abbi2i, as proposed in the comment of ~bj-termab), for the reasons stated in the comment of ~bj-termab.

> Actually, I suspect that ~abid1 was present in set.mm at some point in the past, and was removed for some reason. I could restore the original attribution if you indicate it to me. I'll also adapt the comment.

There has never been an ~abid1.  The name ~abid2 was chosen (maybe not the best choice) to distinguish it from ~abid.

I'm not convinced there is a need for both ~abid1 and ~abid2.  Almost always they will be part of a transitive inference, and with the many variants ~\*eqtr\* it would be rare that either one would shorten a proof compared to the other.  Perhaps I'm wrong.

We often have a choice between (what I will call) a "definitional" form where the shorter expression is on the lhs, and a "computational" form where the shorter expression is on the rhs.  A classic example is ~df-2 vs. ~1p1e2.  We really don't need ~1p1e2, but because it occurs "naturally" in computations, I eventually (9 years later) felt it useful to have a uniform set of 1-digit operations like ~1p2e3, etc., maybe because I kept wanting to type "1p1e2" then correcting myself.

In most cases, though, we don't have both a definitional and computational form for the same theorem.  A long time ago I favored the definitional form for everything (i.e. shorter expression on the lhs) thinking it would provide a consistent rule for both theorems and definitions.  But over time I started to feel the computational form was almost always more natural.  So I have almost always been using the computational form when it wasn't an actual definition.  An example is ~anidm, where we "compute" ph /\\ ph to result in ph.  (There is also ~pm4.24 only for the purpose of having a complete set of PM theorems; apparently PM favored the definitional form.  Most books, though, show ~anidm in computational form.)  Another example is ~inidm, where we don't have a reversed version and probably never will.

I won't reject your proposal to add the "definitional form" ~abid1, but it seems a little redundant, and I don't fully understand its justification, unless it is meant for an educational reason like demonstrating a class version of ~cvjust.

PR #1873, opened 2020-11-12
shorten 19.29r

wlammen:
shorten 19.29r

I removed alexeq as it was tagged as obsolete.  The theorem name was not suffixed with OLD, as is usually done with obsolete material.  You might want to have a closer look at this particular change.


benjub:
Regarding ~alexeq : I introduced ~alexeqg in #832 and #835 and @nmegill had advised me there to obsolete ~alexeq .  I should have suffixed OLD to its label.  So removing it is ok.  I had added the attribution "(Contributed by NM, 2-Mar-1995.)" to ~alexeqg so that the credit and the date stay, even after ~alexeq is removed.

PR #1876, opened 2020-11-16
add ~ alexbii (biconditional form of ~ aleximi) and use it to shorten…

benjub:
… four proofs


benjub:
Remark: Travis rejected the first commit for reasons unclear to me (it looks like it did not even run)...

PR #1879, opened 2020-11-20
Switch from Travis to GitHub Actions, fixes #1867

### LINKED ISSUE #1867: Consider switching from Travis
We may need to consider switching from Travis. Alternatives include GitHub actions.

Details:
https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

david-a-wheeler:
Add a simple GitHub Actions workflow.
This adds a single verifier as a single job.
The goal is to get something working in a simple way before
making anything more complex.

This uses checkmm and recomples it each time; that's the simple
way to get going.


digama0:
Are all the failures just because you are cancelling the travis version or did this mess travis up in some way?

david-a-wheeler:
Mario - it shouldn't have "messed up" Travis. But Travis is so much slower than GitHub actions that GitHub actions will far outpace it. I am stressing Travis a little; I haven't removed the calls to them yet.

BTW, the Rust version is one of the slowest right now, because we have to wait for it to compile first :-).

david-a-wheeler:
Believe it or not, I think we can use this. Comments welcome.

There are many ways to optimize the jobs, in particular, by adding caches. But this is already so much better than our Travis CI that I think we can use this, and optimize later.

Transitioning from Travis to GitHub Actions was a pain, but it was easier than I expected. The systems are similar enough that it wasn't *too* hard to change. I simply transliterated our old job system without trying to do a complete restructure.

digama0:
https://github.com/actions/cache talks about how to cache depdendencies, which should help speed up the rust build, since we just need to cache the executable.

digama0:
SMM3 cache is fixed. We can probably do the same treatment to other verifiers that compile slowly like metamath.exe, although this seems to have a slightly weird setup.

david-a-wheeler:
Mario: Thanks!!

I think it's ready to merge. We can improve things later (like caching metamath.exe), but this is so much better than the Travis setup that we don't need to wait.

david-a-wheeler:
In this run the verifiers' timing was:

* 47 seconds (checkmm, C++)
* 69 seconds (metamath.exe, C, does extra checks)
* 14 seconds (smm3, Rust *when cached*). Note: it's 1111 seconds without a cache
* 102 seconds (mmj2, Java, does extra checks)
* 89 seconds (mmverify.py, Python, set.mm body)
* 73 seconds (mmverify.py, Python, set.mm mathboxes)
* 16 seconds (mmverify.py, Python, iset.mm)

The Rust one is cached. Without caching Rust is the *slowest* because of compilation time; one sample was 111 seconds - https://github.com/metamath/set.mm/runs/1428848217?check_suite_focus=true

We can shave off 30 seconds in metamath.exe by caching its compiled result, and 7 seconds by checkmm by caching its compiled result.

I still think there's no need to move anything to a "nightly" run:

* These all run in parallel. The longest one (mmj2) is 1 minute 42 second long. I think that's fine to keep running all in parallel, we aren't in *that* much of a rush to merge things. I don't think there's a problem if it takes less than 2 minutes to run all the checks.
* The longest-running one (mmj2) does added checks (for our definition conventions) that most other verifiers don't do. So it's not true that omitting mmj2 will have identical results.
* This doesn't really impact the environment. We're talking seconds, and a lot of this is waiting (e.g., for checkouts & downloads).  It'd be impractical to try to measure the "extra" energy used. Adding the two caches would be a nice touch & reduce CPU time some, and that's easy to do, so we should do that eventually. But I don't see a reason to go beyond that.

david-a-wheeler:
@nmegill - please merge this. It's *way* better than our Travis CI setup.

There are a few speedup opportunities to come (e.g., via caching), but there's need to wait for that.

PR #1880, opened 2020-11-21
Add a GitHub Actions cache for metamathexe

david-a-wheeler:


david-a-wheeler:
@nmegill - Please merge. I tweaked the Metamath CI pipeline so that we won't normally recompile the C metamath implementation (metamathexe).

This version still downloads the metamath source code. However, it records the version number, and only recompiles if the version number is different from the compiled version we've cached.

So this will be faster *except* the first run after an update of metamath-exe (where it will have to do a recompile).

PR #1881, opened 2020-11-21
the Lebesgue outer measure of n-dimensiona half open intervals

glacode:
The Lebesgue outer measure of n-dimensional half open intervals equals their n-dimensional volume

I've applied Gérard Lang's suggestions to fix some comments in my previous commits. Thank you Gérard, for your precious help


glacode:
it looks like Travis CI didn't kick in.

I know David is working on porting to GitHub Actions, is this pull request going to be submitted to Travis, manually?

Glauco

david-a-wheeler:
Travis isn't supposed to be kicked in, instead you should see the GitHub Actions.

Except I don't see the GitHub Actions either :-(.

Question: You're pushing to your own repo, and then making a pull request to metamath, correct?!?

I think the problem is that it currently only triggers on a "push". Maybe we need to say "also on a pull request if it's from an outside source", or something like that. Sounds like I need to add conditions that trigger the CI actions. Does that sound correct?

glacode:
I pushed to my repo and then I submitted a pull request from my web page (https://github.com/glacode/set.mm/pull/new/HOIOUTMEASURE)
to the base repository (metamath/set.mm)

Please, let me know if I can do anything to help with the troubleshooting

Glauco

david-a-wheeler:
(I'm back, I had to drive my father to a scheduled medical appointment.)

I would have expected that the push to your personal repo would have triggered the verifiers. I don't understand why they didn't. I guess it would be good to also run verifiers on a pull_request; that's easy to do, but then we run 2x as many verifiers (one for push & one for pull_request) which is weird & confusing.

That didn't happen. I'm going to peruse the documentation on triggers to see what options we have to trigger things more "the way we expect".

david-a-wheeler:
The "easy solution" in GitHub Actions is to do this:

~~~~
on: [push, pull_request]
~~~~

However, this causes duplicates in some circumstances. E.g., "[If we get a pull request from a forked repo, the CI runs correctly. The problem is, if we create a pull request ourselves, we get duplicated  checks on all items](https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012/7)"

I'm investigating this as a possible solution: https://github.com/marketplace/actions/skip-duplicate-actions

glacode:
I'm available if you need me to perform any action for testing

david-a-wheeler:
Thanks!

I created a separate pull request that I think will resolve the problem here:

https://github.com/metamath/set.mm/pull/1882

Can you create a new branch that merges your change and pull request 1882, then push *that*?

glacode:
Let me try...

PR #1882, opened 2020-11-21
Check pull requests without duplicate checks

david-a-wheeler:
Also check pull requests (as well as pushes).
This is necessary when people create a pull request for a change
they made within their own repo.

The "easy solution" in GitHub Actions is to do this:

~~~~
on: [push, pull_request]
~~~~

However, this causes duplicates in some circumstances. E.g.,
"[If we get a pull request from a forked repo, the CI runs correctly.
The problem is, if we create a pull request ourselves,
we get duplicated  checks on all items"
Source: [dups](https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012/7)

We resolve this by using "skip-duplicate-actions" action from:
https://github.com/marketplace/actions/skip-duplicate-actions

Note - this also adds a "workflow_dispatch" event, so that these
jobs can also be easily triggered at any time directly from the
GitHub API.

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -1,5 +1,5 @@
 name: verifiers
-on: [push]
+on: [push, pull_request, workflow_dispatch]
 
 # Run many verifiers on metamath files.
 # We *primarily* check set.mm and iset.mm, but we also run some checks on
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
You might want to merge #1884 instead, which combines this with a specific theorem change.

PR #1883, opened 2020-11-21
the Lebesgue outer measure of n-dimensiona half open intervals (2nd try for GitHub Actions)

glacode:
The Lebesgue outer measure of n-dimensional half open intervals equals their n-dimensional volume

I've applied Gérard Lang's suggestions to fix some comments in my previous commits. Thank you Gérard, for your precious help


glacode:
It looks like Actions are not running, are they?

glacode:
I see, I had to merge your changes in the PR, a bit beyond my current skills with GitHub,

I have to study...

PR #1886, opened 2020-11-22

avekens:

theorem ~f1ssf1 added

section Subgraphs:
* theorems ~issubgr2, ~subgrpprop, ~subgrpprop2, ~subgrpprop3, ~uhgrissubgr , ~subgrfun, ~subgreldmiedg added
* some proofs shortened
* some theorems renamed
* some comments improved


avekens:
@david-a-wheeler I think it was a good idea to use GitHub Actions instead of Travis: it was fast, the usabilitiy (navigation) is better. A slight defiency is that one has to scroll the list of jobs at the left. Unfortunately (or fortunately?) there were no errors in my PR, so I could not check if handling errors is easy...

PR #1890, opened 2020-11-26
comment edits

benjub:
I realize I should have warned before making this PR, since several parts of the file are affected (including @tirix's mathbox, I hope you're ok with these). Mainly typos and spelling consistency.

I also spotted around 400 comments lacking a final period, which I am willing to take care of if @nmegill agrees.


nmegill:
> I also spotted around 400 comments lacking a final period, which I am willing to take care of if @nmegill agrees.

Sure, thanks.

PR #1891, opened 2020-11-27
Misc

benjub:
As agreed in #1890.


nmegill:
@benjub There was a line too long that caused the checks failure.

@david-a-wheeler In the checks log, I see:
```
Run scripts/verify ... --extra 'write bibliography mmbiblio.html' ...
...
?Sorry, couldn't open the file "mmbiblio.html".
```
The initial mmbiblio.html isn't currently in the repository because it is a template file that is dangerous to edit and isn't meant to be edited.  (We could put it there if necessary.)  However, I don't recall why we decided to regen mmbiblio.html - I thought the 'verify markup \*' was sufficient.  Do you recall why we are doing that?

PR #1893, opened 2020-11-29
shorten cbv3hv

wlammen:


benjub:
I had a look at @wlammen's shortening of ~cbv3hv and it made me realize I could move from my mathbox to Main:
bj-spimv
bj-cbv3v
bj-cbv3hv (with, it seems, a shorter proof than cbv3hv)
bj-cbvalv
which will also remove dependency on ax-13 from cleqh, and will probably propagate to remove other dependencies on ax-13.
May I, @nmegill ?

More generally, there are several results like this in my mathbox (Section 21.13.4.11  Removing dependencies on ax-13 (and ax-11)) which are kind of "brute force", so I think it it not wise to move them as a block. Furthermore, some results by @wlammen superseded them. But some of them may be worth moving to Main if they permit to remove ax-13 dependencies from enough later theorems.

nmegill:
> May I, @nmegill ?

Yes, go ahead.

PR #1895, opened 2020-12-01
Misc

benjub:
As agreed in #1893


benjub:
The last comment edits mention that ax-7 is actually right-Euclideanness of equality. There was also a comment mistakenly describing a statement as transitivity.

PR #1897, opened 2020-12-06
Misc

benjub:
within my mathbox

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19128,7 +19128,7 @@ Proof modification of "bj-axc16b" is discouraged (14 steps).
 Proof modification of "bj-axc16g" is discouraged (30 steps).
 Proof modification of "bj-axc16gb" is discouraged (18 steps).
 Proof modification of "bj-axc16nf" is discouraged (31 steps).
-Proof modification of "bj-axc4" is discouraged (25 steps).
+Proof modification of "bj-axc4" is discouraged (15 steps).
 Proof modification of "bj-axd2d" is discouraged (16 steps).
 Proof modification of "bj-axdd2" is discouraged (28 steps).
 Proof modification of "bj-axext3" is discouraged (57 steps).


benjub:
@nmegill What do you think of bj-ax7v and bj-ax7 ? Is it worth moving to Main, similarly to ax6v and ax6 ? (note that the proof of bj-ax7 does not require auxiliary axioms)
This possibility of weakening ax-7 may have been already noticed by you or other contributors, in which case credit should be given to you/them.

nmegill:
> @nmegill What do you think of bj-ax7v and bj-ax7 ? Is it worth moving to Main, similarly to ax6v and ax6 ?

I think it is worth moving.

> (note that the proof of bj-ax7 does not require auxiliary axioms)

That is a nice, even essential, feature.  This is a little different from ax6v, because we know ax6v is adequate for logical completeness in Tarski's original system, whereas we don't know that an unbundled ax-7 is adequate.  I think it is desirable to keep the logical completeness of Tarski's system intact if we can.

> This possibility of weakening ax-7 may have been already noticed by you or other contributors, in which case credit should be given to you/them.

I have wondered about it but didn't know one way or the other.  In particular, many other proof languages work largely at the object-language level and don't have the concept of bundled variables, at least for bound variables.  I don't know how they would treat the "free" variables in ax-7; one could argue that all free variables are implicitly "bound" because of ax-gen and spi.  So I've wondered whether ax-7,8,9 could be weakened with distinct variables.   @digama0 has had more language translation experience and might have something to say.

In Tarski's original system,  ax-7,8,9 are all sub-schemes of a higher-level scheme, and the higher-level scheme can generate any object-language instance of ax-7,8,9 directly, so in a sense it contains "bundling".  In Monk's system, our scheme ax-7 is shown directly and is bundled, and he exploits the bundling to derive x=y -> y=x.

benjub:
@nmegill I'll have to slightly review the proof of bj-ax7 since it uses an unbundled version of equcomi, which uses equid, which can be proved from bj-ax7v. I'll do this when moving to Main (asap).

PR #1898, opened 2020-12-06

avekens:
* new general theorems added: ~fdisjdmun (formerly ~uhgrunlem), ~elpr2elpr, ~riotaeqimp
* rearrangements of theorems
* new theorems ~uhgredgiedgb, ~upgredg2vtx, ~uspgredg2vtxeu, ~uspgridx2vlem, ~uspgridx2v, ~uspgredgaleord, ~uhgr0vusgr, ~uspgr1v1eop added
* some proofs shortened
* some comments improved

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18770,6 +18770,7 @@ New usage of "usgra2adedgwlkonALT" is discouraged (0 uses).
 New usage of "usgrafiedgALT" is discouraged (0 uses).
 New usage of "usgrafisbaseALT" is discouraged (2 uses).
 New usage of "usgredg2vtxeuALT" is discouraged (0 uses).
+New usage of "usgredgaleordALT" is discouraged (0 uses).
 New usage of "usgresvm1ALT" is discouraged (1 uses).
 New usage of "usgsizedgALT" is discouraged (1 uses).
 New usage of "usgsizedgALT2" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@avekens Funny note regarding elpr2elpr: one also has
|- ( A e. { B , C } -> E. x { B , C } = { A , x } )
Worth formalizing ?

avekens:
> @avekens Funny note regarding elpr2elpr: one also has
> |- ( A e. { B , C } -> E. x { B , C } = { A , x } )
> Worth formalizing ?

Yes, that's really interesting: it does not require to assume that B and C are sets: if both are proper classes , then { B , C } = (/), and therefore A e. { B , C } is always false. If one of them is a proper class, say C, then  { B , C } = { B }. It follows that A = B, and with x = A it follows that { B , C } = { B } = { A } = { A , x }. I would appreciate that this theorem becomes part of main set.mm.

benjub:
Actually, I was not volunteering ;-) Too much case-handling ! So it's up to you (and of course all credit would go to the person who formalizes).

nmegill:
> I would appreciate that this theorem becomes part of main set.mm.

That would be fine.

sctfn:
I just proved it in my branch. I'll make a PR including it when I get home.

avekens:
@sctfn I have a proof, too! But make your PR, and we'll see which proof is shorter.

PR #1900, opened 2020-12-07
move bj-ax7v and bj-ax7 to Main

benjub:
As agreed in #1897 

As noted in the comment of ax7v, doing the same thing with ax-8 and ax-9 is possible (and even easier). Should I do it ? Or note it in the comments of ax-8 and ax-9 ?


nmegill:
I would say do it with ax-8 to show the technique, and make a mention in the comment for ax-9.

PR #1901, opened 2020-12-07
Misc

benjub:
As agreed in #1900 
Now, I feel that I should also do it for ax-9 to restore symmetry... But not tonight, so merge for now !


benjub:
Just found a way to prove that adding DV(x,z) to ax-7 is also possible... Soon in my mathbox.

PR #1902, opened 2020-12-08
comment edit

benjub:


benjub:
@nmegill I proved in my mathbox that ax-7 can be recovered by the conjunction of bj-ax7v1 and bj-ax7v2, which is a strict improvement over the proof from ax7v. Can I move this to main ? (ax-7 would be referenced only by ax7v, which would be referenced only by ax7v1 and ax7v2, from which ax7 would be recovered, see bj-ax7) The same is doable with ax-8 and ax-9.

It is not possible to recover ax7 from its weakening with DV(x,z), or from its weakening with DV(y,z), as I note in the comment of bj-ax7. For the first, take "x_i = x_j" (where x_i, x_j are the variables x,y,z...) to be evaluated to true if and only if "i \neq j", and ignore quantifiers. Then reflexivity evaluates to false.  For the second, let the universe of discourse be \ZZ and let "x=y" be evaluated to true if and only if x \equiv y or x \equiv y-1 (where \equiv is equality in \ZZ). Then, symmetry evaluates to false. (In this second model, the auxiliary axiom ax-12 also evaluates to false so, independence is proved when ax-12 is not included in the axiom system.)

(See also #1900 and #1901.)

nmegill:
It looks like a nice result.  Go ahead and move it to main.

PR #1905, opened 2020-12-12
Misc

benjub:
Move equsexv to Main; this removes dependency on ax-13 from 22 theorems.

@nmegill Can I similarly move bj-equsexhv to Main, and use it in cleljust ?  Apparently, this proof modification of cleljust does not change what mmset.html says about it.  This would remove dependency on ax-13 from cleljust, which I think is important, because cleljust is an important "justification" theorem and it is partially bundled (so the fact that it can be proved without ax-13 is not a priori obvious).


nmegill:
> @nmegill Can I similarly move bj-equsexhv to Main, and use it in cleljust ?

That will be fine.

Issue #1904, opened 2020-12-10
Update the diagrams of the Algebraic / Topological structures


tirix:
I volunteer to update the [diagrams of the Algebraic / Topological structures](http://us.metamath.org/mpeuni/mmtopstr.html) in Metamath.
I think some new structures have been defined and may be added to the diagrams:
- new topological structures: Magma (Mgm), Semigroups (SGrp), maybe the non-unital Ring (Rng)
- new graph definitions "as structures": (UHGraph , USHGraph, UPGraph, UMGraph, USPGraph, USGraph, FinUSGraph, ComplGraph, ComplUSGraph)
- categories: Cat, SetCat, CatCat, others?
- possibly, geometries: TarskiG & co.

Is it worth doing the update? Am I missing any? Any suggestion?

nmegill:
> Is it worth doing the update?

Yes.  Stephen Wolfram was impressed when I pointed him to your page:  "That's nice.  I tried to construct things like this around 25 years ago; this is better."

savask:
I also think that an update is very much welcome, and I have a suggestion: maybe algebraic and topological diagrams could have links to corresponding structures in them? Right now links are provided only in the algebraic-topological hierarchy.

tirix:
Thanks for your (very gratifying) feedbacks!

I've updated the first diagram (algebraic hierarchy) and added the links in PR #1907 .

I've tried to provide links even to the arrows. I'm keeping this issue open for the last 3 bullets.

avekens:
Sorry for my late remarks...

I had already a look at the new diagram provided by PR #1907: it's great! 
* The only major problem with it may be that the non-unital Ring (Rng) is still in my Mathbox only - I would appreciate if it can be moved to main set.mm (there was no consensus for that until now...). 
* the arrow to Neutral Element should come from Mnd (Monoid), not from SGrp (SemiGroup)
* what about SRing (semirings)? A SRing is a CMnd with a Mnd as multiplicative group (see ~issrg), and a Ring is a SRing (~ringsrg)
* A Rng is an abelian Group (Abel) with a SGrp as multiplicative group (see ~isrng)
* the mulGrp of a Ring is a Mnd, not a Grp 

Regarding one of the three remaining bullet points:

[...]

PR #1907, opened 2020-12-13
Updated the Algebraic Topology SVG file with links and recent additions

tirix:
As proposed in #1904


avekens:
see my remarks in #1904

PR #1909, opened 2020-12-13
move bj-equsexhv to Main and use it to remove dependency on ax-13 fro…

benjub:
…m cleljust

as agreed in #1905

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19224,7 +19224,6 @@ Proof modification of "bj-equsal" is discouraged (31 steps).
 Proof modification of "bj-equsalhv" is discouraged (10 steps).
 Proof modification of "bj-equsalv" is discouraged (39 steps).
 Proof modification of "bj-equsb1v" is discouraged (17 steps).
-Proof modification of "bj-equsexhv" is discouraged (10 steps).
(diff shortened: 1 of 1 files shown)


benjub:
@nmegill :
- moving bj-equsexvv to Main removes dependency on ax-10, ax-12 from cleljust, thus giving a proof of this important justification theorem from Tarski's FOL system
- moving bj-chvarv to Main removes dependency on ax-13 from axrep1, axrep2, axrep3, axrep4, axrep5
- moving bj-spimedv, bj-spimev, bj-spimevv to Main removes dependency on ax-13 from axsep, dtru, axc16b, eunex, dtrucor, dvdemo1 (and maybe other later theorems as well), which are important theorems at the very beginning of set theory.

May I do these moves (in another PR) ?

nmegill:
You can move these to main.

While I'll accept these, in general I don't see a big benefit to putting in a lot of effort to remove logic axioms from set theory proofs.

david-a-wheeler:
@nmegill said, "While I'll accept these, in general I don't see a big benefit to putting in a lot of effort to remove logic axioms from set theory proofs."

I think you're right, but it's consistent with our general policy of removing axioms where we can.

PR #1910, opened 2020-12-14
Conventions: Limit proof size (200/500 steps)

david-a-wheeler:
Modify conventions to document a recommended maximum
number of essential steps (200 in general, 500 is special).
It also documents that these aren't hard-and-fast rules.
The idea is that expressly documenting *some* preferred maximum
will encourage people to stay below it unless they have a
real reason to do otherwise.


david-a-wheeler:
This attempts to implement the mailing list discussion. I added a little text based on the mailing list discussions.

PR #1913, opened 2020-12-22
Misc

benjub:
In addition to commit messages: manually shorten 19.8a using ~exlimiiv since, strangely, "minimize *" did not catch it.


nmegill:
> manually shorten 19.8a using ~exlimiiv since, strangely, "minimize *" did not catch it.

That's odd, I'll have to investigate it.  I agree that it seems it should be shortened; maybe there's something subtle I'm missing.

Could you modify this PR to add a 19.8aOLD containing the proof before your shortening?  That will save me from having to locate the old proof when I study this problem (maybe in a couple of weeks because of other commitments).  It's on my todo list already.

benjub:
I just added 19.8aOLD. Note that the same thing happens with wl-19.8a, which is not minimized, and maybe other proofs (those using exlimiv followed by ax-mp).

Issue #1908, opened 2020-12-13
analog - analogue - analogon


avekens:
Currently, there are two spellings of the noun "analog" in set.mm:
* analog (AE): 62 times
* analogue (BE): 42 times
* analogon (AE/BE): 0 times

Since we have the convention that American English (AE) should be used (see section 17.1.1  Conventions in set.mm), the 42 occurences of "analogue" should be replaced by "analog".

Alternatively, all occurences could be replaced by "analogon" (from ancient Greek ἀνάλογον), which I would prefer, but which may be a little bit old-fashioned...

benjub:
I'm all for spelling consistency throughout set.mm and I'm ok with the American English convention.  The problem with "analog", though, is that it can be either a noun or an adjective, whereas "analogue" is a noun and "analogous" is an adjective, so maybe the latter two could be preferred ?

See https://www.merriam-webster.com/dictionary/analogue , especially the paragraph "Did you know?".

digama0:
I often use "analogue" instead of "analog" (even though I generally speak AmE) because the latter has also the confounding meaning "not digital". [Random internet person](https://oilpatchwriting.wordpress.com/2013/01/16/analog-vs-analogue/) agrees:

> “With the word traditionally spelled analogue, American writers tend to drop the silent –ue in some contexts, making analog. The spellings are largely interchangeable, though analog is usually used in relation to electronics, while analogue is often used in the sense something that bears analogy to something else. Outside the U.S., analogue prevails for all senses of the word.

benjub:
So we agree to "analogue" for the noun and "analogous" for the adjective ?

Remark: I had a similar dilemma some time ago with alternate/alternative, since their meanings, in some cases, overlap.  I chose to prefer the shorter "alternate" each time that it is possible (https://www.merriam-webster.com/dictionary/alternate). In particular, most comments of ~xxxALT should begin with "Alternate proof of ~ xxx ."

avekens:
For me, "analogue" for the noun and "analogous" for the adjective would be fine. 
@nmegill what's your opinion?

nmegill:
> For me, "analogue" for the noun and "analogous" for the adjective would be fine.
> @nmegill what's your opinion?

I agree.

avekens:
Done with PR #1914 : "analog" replaced by "analogue" resp. "analogous"

PR #1914, opened 2020-12-23
Analog vs. analogue

avekens:
See #1908: Agreement to use "analogue" for the noun and "analogous" for the adjective:
* "analog" replaced by "analogue" resp. "analogous"
* convention to use "analogue" for the noun and "analogous" for the adjective added

AV's mathbox:
* Some rearrangements and minor changes/corrections in the comments in section "extended nonnegative integers"


wlammen:
In some places you separate the reference marker ~ from the theorem/definition name, as done with ~ df-xr. To my knowledge, the Metamath translation process will not render the name as a hyperlink anymore, and you cannot access the originating theorem entry by just clicking. This is within your mathbox, so maybe, there was good reason to do so. But just in case this was not intended, I notify you hereby.

avekens:
@wlammen I got an email with a comment for this PR (about the usage of ~) - but it does not appear here (anymore).
@nmegill Indeed, it seems that for a tilde ~ followed by a token a link in the HTML output is created no matter if there is a blank between ~ and the token or not. Both variants are used in set.mm (e.g. ~ax6e2nd and ~ ax6e). Which is the preferred way? I think it is the variant with the blank between ~ and the token. Should this become a convention?

nmegill:
> @nmegill Indeed, it seems that for a tilde ~ followed by a token a link in the HTML output is created no matter if there is a blank between ~ and the token or not. Both variants are used in set.mm (e.g. ~ax6e2nd and ~ ax6e). Which is the preferred way? I think it is the variant with the blank between ~ and the token. Should this become a convention?

Even though  "\~ax6e2nd" works, the variant "\~ ax6e" has been our convention which is used in 99+% of cases.  Historically I chose that so a match to " ax6e " can be used to make global label changes in a text editor without having to use a regex.  I suppose "verify markup" should complain when the space is missing to make sure we're consistent; I'll put it on my list.

wlammen:
@avekens I thought for a moment a tilde must precede a theorem name immediately to form a valid link. Before finally posting to this respect, I checked the set.mm file and learned I used the space myself. An incoming message distracted me then, and sometime later I accidentally hit the send button of the prepared post. Although I deleted the message immediately afterwards, it was online visible for a couple of seconds. And that was apparently sufficient to notify you of this event.

So I was simply mistaken, and there is no need to care about the contents.

benjub:
There are a few cases remaining. I fixed the three occurrences of "analog\n" (end of lines) and the single occurrence of "analog," in #2380. There are 298 occurrences of "analog.". Probably, most of them are nouns (so have to be changed to "analogue.", but we have to check whether some of them are adjectives (so have to be changed to "analogous."). I'll fix them when things are quieter (to avoid merge conflicts), if @avekens agrees.

PR #1918, opened 2020-12-25
A few theorems moved to the main part (mostly about countable sets)

glacode:
My next commit contains proofs using some of the theorems moved to
the main part, with this commit.
In my opinion, they could be useful for other proofs, in the future.


nmegill:
I noticed that in some of the theorems imported, the comments use "elementhood" to mean "membership".  "Elementhood" is not a word in Miriam-Webster or Wiktionary, although google shows it is used on some sites discussing set theory, including exactly 2 mentions in the Stanford Encyclopedia of Philosophy (vs. 1200 mentions of "membership", not all of them related to set theory).  In set.mm, "elementhood" occurs 96 times and "membership" 684 times.  Should we use one or the other always for consistency?

benjub:
I agree with @nmegill, and "membership" is much more common.

By the way: one phrase I much dislike is "epsilon relation/predicate" in place of "membership relation/predicate". This term comes from the fact that the symbol used for membership was in older times a lowercase epsilon. But with things like that, we would soon say "the cross operation" for "addition", the "dash operation" for "subtraction", etc. It is like the "R-matrix", named like this because one of the first papers studying the scattering matrix used the letter "R" for it...

nmegill:
> By the way: one phrase I much dislike is "epsilon relation/predicate" in place of "membership relation/predicate".

I agree in principle, but it doesn't seem to be standard practice.

In set.mm, it seems we use "membership relation" for the stylized-epsilon primitive (e.g. eqeltri) (even though "e." is not a formal relation - we should probably call those "membership predicate" or "membership primitive") whereas "epsilon relation" is reserved for the defined symbol "_E" because that's what most literature seems to call it, at least the literature I worked from.

There are many definitions that use the symbol as part of their name, such as "gamma function" instead of "generalized factorial".  There's "lambda calculus" and "beta reduction" and "iota operator" and "Hilbert's epsilon operator".

If we use "membership predicate" for "e." and "membership relation" for "_E", we should at least mention that the literature typically calls it "epsilon relation".  However, I am somewhat hesitant to deviate from what seems to be (for better or worse) standard terminology, and I'd like to hear another opinion.

Curiously, ProofWiki has separate pages for the same thing that apparently don't know about each other:
https://proofwiki.org/wiki/Definition:Membership_Relation
https://proofwiki.org/wiki/Definition:Epsilon_Relation

digama0:
Actually, looking at those two pages, they do express a distinction that I was also going to mention: The two are not interchangeable, with "epsilon relation" referring more specifically to the binary relation that comes with a model of set theory (where the things on each side of the `e.` are of the same kind), while "membership relation" generally relates "members" to "sets", which may or may not be of the same kind. For example we might use it as a relation between sets and classes in ZFC, or in type theory we might relate elements of `A` to elements of `set A` aka `A -> bool`.

In these theories there may not be anything like set theory's "epsilon relation" which is an actual binary relation between things of the same kind (i.e. `A -> A -> bool` for some `A`), where we can ask meaningfully if the relation is reflexive, for example. Model theoretically, we think of a set theory model as a pair `(M, e)` where `e C_ (M X. M)` is the "epsilon relation" on that model, the thing that interprets membership; you would not call the relation `A -> set A -> bool` between sets and members in type theory an "epsilon relation" in this sense.

benjub:
Thanks @nmegill and @digama0 for these interesting remarks. I think in the model (M, e), e could still be called the membership relation of the model. But obviously I won't change standard practice !

Yes, "Gamma function" and "zeta function" are unfortunate. The first could be called "generalized factorial function" as you wrote (noting the "shift-by-one"), and the second the "sum of powers function"... in which case no one would know that we are talking about the zeta function, failing the main goal of denomination ;) At the extreme, there are also the constants, like the "circle constant" pi (initial letter of the Greek "perimetros"), etc.

As for "lambda calculus", I think both the name and the notation make it appear obscure and frightening, even esoteric, alas.
I think "functional calculus" (or even "applicational calculus") and notation `(x \mapsto M)` for `\lambda x.M` would have made a lot for its popularization. And finding meaningful names for alpha conversion and beta reduction.

Grothendieck strongly emphasized the importance of giving meaningful, if possible common names, to newly defined concepts (and successfully practiced what he preached in many instances).

PR #1920, opened 2020-12-27
Misc

benjub:
* move sb56, cleljust, cleljustALT earlier since they do not depend on ax-13 anymore !
* move elopg to Main to remove an extraneous hypothesis to elop
* mathbox: substitution

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16009,6 +16009,7 @@ New usage of "elnlfn" is discouraged (4 uses).
 New usage of "elnlfn2" is discouraged (2 uses).
 New usage of "elnp" is discouraged (5 uses).
 New usage of "elnpi" is discouraged (4 uses).
+New usage of "elopOLD" is discouraged (0 uses).
 New usage of "elpaddatiN" is discouraged (2 uses).
 New usage of "elpaddatriN" is discouraged (0 uses).
 New usage of "elpclN" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
Here my 2 cents on ax-12:
I have a somewhat different view on ax-12 than is described in its comments, where its role in substitution is emphasized. For me, it sheds a light on two distinct properties of the binary relation =. The first is that = is the finest granular expression in logic. If you cannot distinguish two objects x and y by =, i.e. x=y holds, then you cannot distinguish it by means of any other expression ph i.e. the logical value of ph is constant for all x equal to y. This idea is expressed by ax12v, best seen if one puts an extra distinct variable constraint on ph and x. 
Edit: This far reaching aspect of ax-12 is a heavy-weight promise: You must never extend a logic by introducing a new operator (such as "is set variable 'x'"), that is more separating than =, in particular one that allows inspection of the underlying textual representation of a logical expression.
Second, ax-12 has an impact on bundling. This is expressed in axc112. It is based on a second property of = holding in any non-trivial universe: = is not constantly true across all possible object combinations. So, if it appears so, expressed by a for-all quantifier, the possible combinations must somehow be restricted. This opens a door into the meta level of logical expressions. You can now tell whether two set variables are the same after substitution, and if they are, this carries over to qualifiers (edit: must, otherwise quantifiers would be more separating than equality). This second aspect is not covered by bj-ax12.

benjub:
@wlammen That is a nice way of seeing it.  This is the reason why I introduced the weakening bj-ax12: to remove the bundling aspect.  Actually, I can prove that neither sp nor ax-12 nor ax-13 can be recovered from ax-1 to ax-11 and bj-ax12.  I hope to post about this on the disucssion group today or tomorrow.

PR #1923, opened 2020-12-29
Misc

benjub:
@nmegill I moved equsexvw to Main (parallel to equsalvw) in order to remove dependencies from cleljust. Indeed, the latter can be proved from Tarski's system, as shown in ~bj-cleljust.  However, the proof modification of cleljust is discouraged since it serves as an example in setmm.html.  I see a few options (given that I think it's important to have this fundamental theorem proved solely from Tarski's system):
* adapt the wording in setmm.html (I can make a proposal); it may be the best, although a bit harder to follow for a newcomer;
* use cleljustALT for the example (actually, I think @digama0's cleljustALT can be deleted, since the proofs of cleljust and cleljustALT are almost exactly the same, adding no insight);
* maybe the best option: use another, simpler example in setmm.html ?

I signaled in two comments the correspondences to modal axioms D and T.

Mathbox.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19119,6 +19119,9 @@ Proof modification of "bj-aev" is discouraged (39 steps).
 Proof modification of "bj-aevlem1v" is discouraged (42 steps).
 Proof modification of "bj-alcomexcom" is discouraged (45 steps).
 Proof modification of "bj-ax12v" is discouraged (35 steps).
+Proof modification of "bj-ax6e" is discouraged (43 steps).
+Proof modification of "bj-ax6elem1" is discouraged (27 steps).
+Proof modification of "bj-ax6elem2" is discouraged (27 steps).
 Proof modification of "bj-ax8" is discouraged (55 steps).
 Proof modification of "bj-ax9" is discouraged (35 steps).
 Proof modification of "bj-axc10v" is discouraged (37 steps).
@@ -19216,7 +19219,6 @@ Proof modification of "bj-equsal" is discouraged (31 steps).
 Proof modification of "bj-equsalhv" is discouraged (10 steps).
 Proof modification of "bj-equsalv" is discouraged (39 steps).
 Proof modification of "bj-equsb1v" is discouraged (17 steps).
-Proof modification of "bj-equsexvv" is discouraged (36 steps).
 Proof modification of "bj-eunex" is discouraged (53 steps).
 Proof modification of "bj-exlimmpbi" is discouraged (11 steps).
 Proof modification of "bj-exlimmpbir" is discouraged (11 steps).


nmegill:
> * adapt the wording in setmm.html (I can make a proposal); it may be the best, although a bit harder to follow for a newcomer;

(You mean mmset.html.) I chose cleljust to be easy for a newcomer: it is close to axioms (the dv violation is even in axiom ax-5); it is part of pred calc,; it uses only standard definitions that (almost) everyone knows; and the violation involves substituting an expression with 2 setvars into a wff variable, illustrating how dv requirements propagate.

> * use cleljustALT for the example (actually, I think @digama0's cleljustALT can be deleted, since the proofs of cleljust and cleljustALT are almost exactly the same, adding no insight);

The present cleljustALT has the same proof length but uses nfv instead of ax-5, which requires a newcomer to learn df-nf, so it wouldn't be an optimal example.

> * maybe the best option: use another, simpler example in setmm.html ?

For now, the simplest option would be to rename the present cleljustALT to cleljustALT2 and move it to Mario's mathbox (he can delete it if he wants or ask us to), then rename cleljust to cleljustALT in both set.mm and mmset.html.raw, then add your new cleljust.

If you think of a simpler or better example for mmset.html  to illustrate dv violations in , I'm all for that.  But it can wait.

digama0:
I don't remember this particular theorem, but from the looks of it this was when I was updating all uses of the `(ph -> A. x ph)` idiom to replace them with the `F\ x ph` idiom instead. I think that this is the better style generally, but if you are dodging some axiom or something then I don't mind it being changed back. Certainly I have no interest in maintaining the proof as is outside of general maintenance concerns (which suggest to pick one version and stick with it).

I don't think asking users to learn df-nf is a hard sell, considering that we use it widely. No one is going to drop in and just look at this one proof in isolation. Plus using `x e. y -> A. z x e. y` as a coded form of df-nf is not clearer unless you have already come to terms with the way in which metamath is not FOL, at which point df-nf is already natural.

benjub:
@nmegill You're right, I will use the current cleljust, relabeled cleljustALT, as the example for mmset.html. It is indeed simpler for a newcomer, and maybe the simplest in this respect: it is the earliest theorem with no wff metavariable directly referencing ax-5.
I will relabel the current cleljustALT to cleljustALT2, keeping it in Main smaybe adding an explanation. @digama0 : I noticed this parallelism in many theorems with hypothesis ph -> A. x ph versus F/ x ph , and I agree that the latter, that you introduced, is more natural.  For this specific case however, I think it is best to present in mmset.html an example directly referencing ax-5.

PR #1924, opened 2020-12-29
remove dependencies from cleljust and move it earlier (now within Tar…

benjub:
…ski's system); keep cleljustALT and cleljustALT2; adapt the explanation in mmset.raw.html to use cleljustALT

(as agreed in #1923)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15256,6 +15256,7 @@ New usage of "chub2i" is discouraged (24 uses).
 New usage of "chunssji" is discouraged (0 uses).
 New usage of "circgrpOLD" is discouraged (0 uses).
 New usage of "cleljustALT" is discouraged (0 uses).
+New usage of "cleljustALT2" is discouraged (0 uses).
 New usage of "climinfOLD" is discouraged (1 uses).
 New usage of "climinffOLD" is discouraged (0 uses).
 New usage of "clmgmOLD" is discouraged (1 uses).
@@ -19181,7 +19182,6 @@ Proof modification of "bj-ceqsalv" is discouraged (10 steps).
 Proof modification of "bj-chvarv" is discouraged (18 steps).
 Proof modification of "bj-chvarvv" is discouraged (11 steps).
 Proof modification of "bj-clabel" is discouraged (43 steps).
-Proof modification of "bj-cleljust" is discouraged (22 steps).
 Proof modification of "bj-clelsb3" is discouraged (63 steps).
 Proof modification of "bj-cleqhyp" is discouraged (21 steps).
 Proof modification of "bj-cmnssmnd" is discouraged (36 steps).
@@ -19345,8 +19345,8 @@ Proof modification of "cdleme20yOLD" is discouraged (314 steps).
 Proof modification of "ceqsalgALT" is discouraged (58 steps).
 Proof modification of "chordthmALT" is discouraged (440 steps).
 Proof modification of "circgrpOLD" is discouraged (74 steps).
-Proof modification of "cleljust" is discouraged (25 steps).
 Proof modification of "cleljustALT" is discouraged (25 steps).
+Proof modification of "cleljustALT2" is discouraged (25 steps).
 Proof modification of "climinfOLD" is discouraged (850 steps).
 Proof modification of "climinffOLD" is discouraged (219 steps).
 Proof modification of "clmgmOLD" is discouraged (50 steps).
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -3481,7 +3481,7 @@ This substitution transforms the original distinct variable requirement
 into the two distinct variable pairs ` z ` ,` x ` and ` z ` ,` y ` ,
 which will implicitly accompany step 1 (although not
 shown explicitly in step 1 of the proof listing).  Thus step 1 can be
-read in full, "` ( x e. y -> A. z x e. y ) ` where ` z ` ,` x ` are
+read in full, "` ( x e. y -> A. z ` ` x e. y ) ` where ` z ` ,` x ` are
 distinct and ` z ` ,` y ` are distinct."  The proof
 verifier will check that this requirement accompanies the final theorem,
 otherwise it will flag the proof step as invalid.  You can see this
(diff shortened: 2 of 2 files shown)


nmegill:
In mmset.raw.html, I noticed you changed
```
` ( x e. y -> A. z ` ` x e. y ) `
```
to
```
` ( x e. y -> A. z x e. y ) `
```
The purpose of   \`  \`  is to put extra space after the z (you will see this done elsewhere in this file also).  The processing of math symbol markup in comments and by the "markup" command is just a simple symbol-by-symbol replacement without the spacing tweaks performed when processing actual math content ($p, $a, $e statements).  It is normally rare that you need such tweaks in comments, and unlike in $p statements you can control spacing manually by breaking up the math string.  I don't plan to change this behavior, at least not in the near future, as it would require a large change to the code.  Could you put back the  \`  \`   and resubmit?

(I noticed you fixed the "^^^^" under the dv violation printout - thanks.  What happened was that two years ago I changed ax-17  to ax-5 globally and overlooked that.)

nmegill:
Also, can you change
```
Note that the "normal" proof format is used,
```
to
```
Note that the "normal" proof format is used, which you can convert to
with "save proof cleljustALT /normal",
```
Thanks.

benjub:
Done.   I did not know about the spacing issue for ` ( x e. y -> A. z ` ` x e. y ) ` , I thought this was related to an older line break, which is why I removed it at first.

> (I noticed you fixed the "^^^^" under the dv violation printout - thanks. What happened was that two years ago I changed ax-17 to ax-5 globally and overlooked that.)

I actually reran mm194 as indicated, which also explains the change in statement number and the unindent.

benjub:
There is something slightly worrying about the behavior of "write source" in the metamath program: it looks like it removes some comments. Following the instructions in mmset.html about cleljust, here is what I did and what I obtained:

Initial state of set.mm:
```
  ${
    $d x z $.  $d y z $.
    $( When the class variables in definition ~ df-clel are replaced with
```
Manually change to:
```
  ${
    $( $d x z $. $) $d y z $.
    $( When the class variables in definition ~ df-clel are replaced with
```
Run metamath (0.194)
```
MM> r set.mm
MM> wr so set.mm/rew
```
Resulting set.mm:
```
  ${
    $d y z $.
    $( When the class variables in definition ~ df-clel are replaced with
```
The commented out DV condition has disappeared.

david-a-wheeler:
@benjub - I think metamath.exe is only retaining the *last* comment in certain cases. That should be easy to test.

nmegill:
> There is something slightly worrying about the behavior of "write source" in the metamath program: it looks like it removes some comments.

This is definitely a bug.  It seems to happen with 'write source.../rewrap' but not with just 'write source'.  I'll investigate it.

benjub:
I think that "write source .../rewrap" is run (by contributors) before most PRs ?  At least, I run it before most of mine.  I don't know if there is an easy way to check that no comments were lost this way, for instance using some of the many functionalities of git (and how long back we have to go, depending on which version(s) of metamath are affected).

nmegill:
The bug affects the following situations:

(1) If a line starts with a comment and is followed on the _same_ line by a $c, $v, ${, $}, or $d statement, the comment is (incorrectly) deleted.

e.g. `$( abc $) ${ $( def $) $}` incorrectly becomes `${ $( def $) $}`

(2) If a line starts with multiple comments and is followed on the _same_ line by a $c, $v, ${, $}, or $d statement, those comments are (incorrectly) deleted.

e.g. `$( abc $) $( def $) ${ $( ghi $) $}` incorrectly becomes `${ $( ghi $) $}`

(3) If the end of a multi-line comment  is followed on the _same_ line by a $c, $v, ${, $}, or $d statement, the part of the comment on that line is (incorrectly) deleted.  In this case, the comment would not be closed, causing a syntax error on the next read, so the problem normally wouldn't go unnoticed.  No one has reported seeing this, though.

e.g.
```
$( multi
line $) $c ( $.
```
becomes the syntactically incorrect
```
$( multi
$c ( $.
```
I hope to fix the bug in a day or two but this should give an idea of its scope.

nmegill:
To prevent further problems until this bug is fixed, I updated metamath.exe to bypass "/rewrap" temporarily.  The "/rewrap" qualifier is still allowed so that any scripts will still work, but it is now a do-nothing qualifier that prints this message:
```
Metamath - Version 0.195 30-Dec-2020          Type HELP for help, EXIT to exit.
...
*** (30-Dec-2020) /REWRAP and /FORMAT are temporarily bypassed in
*** version 0.195 while a bug is being fixed.
Writing "set.mm"...
191440 source statement(s) were written.
MM>
```
If you want to use /rewrap anyway, you can save version 0.194 (and of course carefully check the differences that result).

----

The '/rewrap' qualifier was added on 12-Jun-2011.  When I did  ` grep '\$) *\$[dcv}{]' ` on all versions of set.mm in my personal archive (snapshots going back to 1998, usually one or more per year but not every github release), there was a single match due to unconventional formatting below theorem ~fintopcomp (renamed to ~fincmp in 2014):
```
set.mm.2010-08-29:    $( [22-Dec-2008] $) $( [22-Dec-2008] $) $}
```
In the next set.mm I kept, 2011-08-25, this theorem had been moved from FL's mathbox to the main set.mm with a correctly dated acknowledgment.

After 2011 it is of course possible that someone added a comment followed on the same line by  $c, $v, ${, $}, or $d that was deleted by '/rewrap' (we'll never know), but that is something that would be unconventional for us to do, so I don't think it is too big a concern.  If someone knows how to do this grep in all past git versions, as @benjub suggests, that could be useful, especially since not everyone did /rewrap before their commits.

david-a-wheeler:
I think `git log -G '\$) *\$[dcv}{]'` is what you want.  I'm running it now.

nmegill:
This is fixed in version 0.196.  What I thought would be a one-line fix turned into a 70-line fix.  But that's OK, not as much New Year's revelry going on this year. :)

Here's how it now handles the commented-out $d in @benjub's original bug report.

First, we comment out the $d on cleljustALT:
Before "/rewrap":
```
     $( $d x z $. $)  $d y z $.
```
After "/rewrap":
```
     $( $d x z $. $)
     $d y z $.
```
Note that it placed the comment above the real $d in order to have the correct indentation for the real $d.  Basically it wants to keep a consistent indentation for the keywords that don't have labels.  Maybe this isn't the behavior everyone expects, but putting the comment on a separate line makes the code simpler and arguably looks better in some cases (e.g. ${ will always be aligned with matching $}).

Next, we uncomment the $d (in the re-wrapped set.mm):
Before "/rewrap":
```
     $d x z $.
     $d y z $.
```
After "/rewrap":
```
     $d x z $.  $d y z $.
```
Note that it reassembled the 2 $d's onto one line, so everything is back to where it started, with 0 differences from the original set.mm.

PR #1936, opened 2021-01-30

avekens:
Main set.mm:
* theorems ~sylbbr, ~sylbb1,  sylbb2  moved from BJ's mathbox, proof of ~ bitri minimized as indicated. @benjub section "Syllogisms" in your mathbox is empty now - please revise it accordingly
* theorem ~neqne moved from GS's mathbox
* theorems ~preqr1g, ~fntpb, ~wrdl1exs1, ~wrdl2exs2, ~wrd3tpop, ~wrdlen3s3 added
* proof of ~wrd2pr2op shortened
* comments about "singleton words" and "doubleton words" adjusted

* theorems ~wrdred1, ~wrdred1hash (about truncated words) added
* theorems ~seqvtx1wlk, ~red1wlklem, ~red1wlk added
* section "Walks for loop-free graphs" and related theorems added
* theorems ~pthontrlon, ~spthonpthon added
* theorems ~uhgr1wlkspth, ~usgr2wlkneq, ~usgr2wlkspth added
* theorems for walks of length 2 added to section "Examples for walks, trails and paths"

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14518,6 +14518,7 @@ New usage of "bezoutlem4OLD" is discouraged (0 uses).
 New usage of "bianirOLD" is discouraged (0 uses).
 New usage of "bitr3" is discouraged (6 uses).
 New usage of "bitr3VD" is discouraged (0 uses).
+New usage of "bitriOLD" is discouraged (0 uses).
 New usage of "bitsfzolemOLD" is discouraged (0 uses).
 New usage of "bj-0" is discouraged (1 uses).
 New usage of "bj-1" is discouraged (0 uses).
@@ -18677,6 +18678,7 @@ New usage of "ubth" is discouraged (1 uses).
 New usage of "ubthlem1" is discouraged (1 uses).
 New usage of "ubthlem2" is discouraged (1 uses).
 New usage of "ubthlem3" is discouraged (1 uses).
+New usage of "umgr2adedgwlkonALT" is discouraged (0 uses).
 New usage of "un0.1" is discouraged (1 uses).
 New usage of "un01" is discouraged (0 uses).
 New usage of "un10" is discouraged (0 uses).
@@ -19093,6 +19095,7 @@ Proof modification of "bezoutlem4OLD" is discouraged (906 steps).
 Proof modification of "bianirOLD" is discouraged (19 steps).
 Proof modification of "bitr3" is discouraged (14 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).
+Proof modification of "bitriOLD" is discouraged (18 steps).
 Proof modification of "bitsfzolemOLD" is discouraged (737 steps).
 Proof modification of "bj-0" is discouraged (5 steps).
 Proof modification of "bj-1" is discouraged (5 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
I am upgrading the us2 server (to the latest version of Devuan).  There may be a day or more delay before I update the Metamath pages, depending on how it goes.

wlammen:
Hi
While I looked at the changes to bitri, I thought, its modification is a bit against the policy I used in my more active phases.

1. I kept attributions to obsoleted proof improvements. This was to honour, and encourage, efforts on improving the database.
2. If, for some reason, I introduced a convenience theorem, like xchbinx, I almost always made sure, it has more than 7 applications. This was a number Norm gave me as the threshold value, beyond that the total number of proof bytes in the database usually start to decrease. Only then I (and Norm, if I remember right) considered a convenience theorem as justified.
3.  If a convenience theorem minimizes a proof by merely replacing a bunch of proof lines verbatim, I did not consider this enough to add an attribution. A simple editing automaton succeeds on this. Please verify, that none of the applications of xchbinx carries in fact such a note, although many were updated along with the new xchbinx.

Maybe the world has changed since then. Please comment in case you think so.

avekens:
Hi Wolf,
here my response  to your remarks:

> 1. I kept attributions to obsoleted proof improvements. 
> This was to honour, and encourage, efforts on improving the database.

I was not sure how to deal with the comment/attributions for ~bitri, because it was already shortened by you before, another shortening was suggested by BJ, and I implemented it. But it was only the result of an automatic minimization, after I moved ~sylbbr to main set.mm, so there should be no attribution for this (and also no OLD version), as if it was performed by a global minimization. Therefore, I would propose to recover the previous comment/attributions, delete the OLD version, and only leave the shortened proof. @nmegill @benjub would this be OK also for you?

> 2. If, for some reason, I introduced a convenience theorem, like xchbinx, I almost always made sure, 
> it has more than 7 applications. This was a number Norm gave me as the threshold value, 
> beyond that the total number of proof bytes in the database usually start to decrease. 
> Only then I (and Norm, if I remember right) considered a convenience theorem as justified.

There is another rule/convention to move a theorem out of a mathbox to main set.mm if it is used in another mathbox. This is the case with ~sylbb1, which I use in my mathbox. Since this theorems belongs to a series of related theorems, I moved them "en bloc", because I think it makes no sense to break up this series.

> 3. If a convenience theorem minimizes a proof by merely replacing a bunch of proof lines verbatim, 
> I did not consider this enough to add an attribution. A simple editing automaton succeeds on this. 
> Please verify, that none of the applications of xchbinx carries in fact such a note, although many were 
> updated along with the new xchbinx.

I think this corresponds to my proposal for item 1.

benjub:
As for proof shortenings, those which can be done automatically by using "minimize_with" can be done "silently", that is, no credit modifications, and no *OLD version.

As for the now empty section in my mathbox: I'll copy its comment to the comment for sylbbr, since I think it can be useful. (And I think moving these theorems to Main is justified, by reasons given in that comment: they have several applications, and they form a coherent set with other syl* theorems.)

As for "singleton word" and "doubleton word", I would prefer "1-letter word" and "2-letter word".

nmegill:
> Therefore, I would propose to recover the previous comment/attributions, delete the OLD version, and only leave the shortened proof. @nmegill @benjub would this be OK also for you?

It's OK with me.

wlammen:
Its ok with me, too.

PR #1943, opened 2021-03-03
The 1-dimensional Lebesgue measure agrees with the Lebesgue measure…

glacode:
…on subsets of Real numbers

- The Lebesgue Measure ` vol ` that Mario defined in 2014 is the "same" as the n-dimensional Lebesgue Measure ` voln ` for n equal to 1 (up to the "canonical" isomorphism)

- Mario's ` vol ` is a measure (with respect to my definition of measure, df-mea)

- a few typos in comments from my previous commit have been fixed (thanks to savask for his precious feedback)

- please, notice I've moved a few small theorems to the main body (Norm, please let me know if you agree with such moves)


nmegill:
> * please, notice I've moved a few small theorems to the main body (Norm, please let me know if you agree with such moves)

Yes, this is fine.

PR #1948, opened 2021-03-14
remove a dependency on ax-12 from two lemmas

wlammen:
I am not sure where this finally will lead. If we are unlucky, that's it, and removing ax-12 from these lemmas cannot be extended to other theorems. But maybe, focusing on equality as a replacement of ph in this Tarski section with implicit substitution can solve some bundling issues in more general theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14358,6 +14358,7 @@ New usage of "axc11nALT" is discouraged (0 uses).
 New usage of "axc11next" is discouraged (0 uses).
 New usage of "axc11nfromc11" is discouraged (0 uses).
 New usage of "axc11nlemALT" is discouraged (2 uses).
+New usage of "axc11nlemOLD" is discouraged (0 uses).
 New usage of "axc16ALT" is discouraged (1 uses).
 New usage of "axc16b" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
Very nice !
Note that the dv conditions for ~speq can be written $d x y z $. (and the dv condition involving ph is probably superfluous since it does not appear in the statement).
In the comment of ~ speq, can you replace "to an equation" by "to an equality", or maybe even "to an equality with a dv condition"?

nmegill:
This is nice.

BTW note that any FOL theorem containing only setvars that are all mutually distinct (and no wff variables) can be proved without ax-10 through ax-13, by invoking ax10w through ax13w.  It might be interesting (and useful) to see if there are any other such theorems that can be reproved and moved up to the ax-4 through ax-9 section.

In this case (if I understand correctly), speqv effectively replaced an internal section of the axc11nlem proof that previously used spsd.  Since axc11nlem does not have a $d between x and y, it's not apparent from the theorem itself that ax-12 could have been eliminated.  As this example illustrates, an analysis of proofs might also be needed to see if other uses of ax-10 through ax-13 could be eliminated.

PR #1949, opened 2021-03-14
Document dis-preference for ax-10..13

david-a-wheeler:
Document that we prefer people to not use ax-10..13
when they don't need to (and how to do that).


benjub:
Note that this is a special case of prefering to use fewer axioms whenever possible, and similarly for the positioning, in general, consider the axioms ax-XXX as "milestones" of set.mm dividing it into intervals, and place a theorem in the earliest possible interval.

david-a-wheeler:
@benjub - I agree. I think this pull request does that; are you proposing a different change?

benjub:
@david-a-wheeler Your text sounds good to me.

PR #1951, opened 2021-03-16
Ifp

benjub:
As agreed on the google group: use ifp in weak deduction theorem, and adapt mmdeduction.raw.html.  There may be some adjustments to make.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4432,6 +4432,7 @@
 "cvpss" is used by "cvnsym".
 "cvpss" is used by "cvntr".
 "cvr2N" is used by "cvrval4N".
+"dedtOLD" is used by "con3OLD".
 "df-0" is used by "ax1ne0".
 "df-0" is used by "axi2m1".
 "df-0" is used by "axpre-mulgt0".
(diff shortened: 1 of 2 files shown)


benjub:
@nmegill I noticed I probably should have used relative links to the page mmdeduction.html

Actually, in set.mm, there are about 30 links of the form ~ us.metamath.org/mpeuni/xxx.html . Should I change them to ~ xxx.html ?

nmegill:
> @nmegill I noticed I probably should have used relative links to the page mmdeduction.html
> 
> Actually, in set.mm, there are about 30 links of the form ~ us.metamath.org/mpeuni/xxx.html . Should I change them to ~ xxx.html ?

Yes, as long as "xxx" begins with "mm" (which is a special prefix not allowed for statement labels).  E.g. `~ http://us.metamath.org/mpeuni/mmcomplex.html` can be changed to `~ mmcomplex.html`
and `~ http://us.metamath.org/mpeuni/mmset.html#axiomnote` can be changed to `~ mmset.html#axiomnote`.

Make sure the link occurs either in a statement description or in the text under a section header.  I think all the ones to change currently all have a preceding ~.  Other links such as those in the big section at the beginning of set.mm should be left as is; e.g. `http://us.metamath.org/mpeuni/conventions.html` on the 5th line of set.mm should be left alone.

Looking at set.mm, I noticed that in the ~ conventions description there is a mistake:  `~ mmcomplex ` should be `~ mmcomplex.html `.  (`~ mmcomplex ` actually usually works since `.html` is most web servers' default file extension, but it isn't recommended.)  Not sure if there are others like these.

(While your example was `~ us.metamath...`  I think you meant `~ http://us.metamath...`)

Thanks!

PR #1954, opened 2021-03-19
prove aev without ax-12

wlammen:
1. Fix a couple of dates in the renaming section.
2. Rename speqv to spaev for consistent naming style.
3. Give a short introduction to the semantics of A. x x = y in spaev
4. Cluster aev related theorems in one place
5. Most important: Reprove aev without ax-12

As always: Look out for spelling errors and poor English

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14179,6 +14179,7 @@ New usage of "aecom-o" is discouraged (4 uses).
 New usage of "aecoms-o" is discouraged (6 uses).
 New usage of "aev-o" is discouraged (1 uses).
 New usage of "aevALT" is discouraged (0 uses).
+New usage of "aevOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
That's very nice!  I never would have guessed that only ax-4 thru 9 would be needed, especially since aev bundles 60 different cases.

wlammen:
The latest modifications to aev root in spimw.  When I saw the Tarski section first in 2017, I thought, what on earth is this good for.  Now I know.  If you look at the credit list of aev, you see how it was gradually stripped off references to axioms.  Although my name appears frequently, I should point out, that it is in fact a collaboration of several persons contributing ideas over the years to the final result.  All of this had not been possible, if I (and others) had no access to this data base.  Nice achievement of something along the lines of open source.

PR #1955, opened 2021-03-19
Misc

benjub:
@nmegill To me, bj-dfnf5, bj-abfalbi, bj-abtrubi, bj-complss, bj-compleq seem important enough to be moved to Main. What do you think?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12386,7 +12386,6 @@
 "retbwax2" is used by "merco1lem2".
 "retbwax2" is used by "merco1lem3".
 "retbwax2" is used by "retbwax3".
-"rexsnsOLD" is used by "r19.12snOLD".
 "rhmsubcALTV" is used by "rhmsubcALTVcat".
 "rhmsubcALTVlem1" is used by "rhmsubcALTV".
 "rhmsubcALTVlem2" is used by "rhmsubcALTVlem3".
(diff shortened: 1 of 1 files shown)


wlammen:
You might want to change this:
"version of ~ aev , so for credits for it see there.  (Contributed by
       Wolf Lammen, 19-Mar-2021.)"

benjub:
@wlammen I don't understand. What do you propose ?

wlammen:
...so for credits for it...
I botched this sentence already, but you did not improve it either.

benjub:
I think I made it correct, even if not elegant!  But I'm not a native speaker. Maybe remove that sentence altogether?

wlammen:
I had used "... so for credits see there ..." , if I were more attentive in the early morning.  Another question is, whether we should point to other theorems for credits at all.  I agree, this could open a can of worms, which might develop into something unpleasent over time.

benjub:
I agree with @wlammen and propose to remove that sentence. Waiting for @nmegill to chime in.
(@wlammen: the wording you proposed is indeed better!)

nmegill:
> @nmegill To me, bj-dfnf5, bj-abfalbi, bj-abtrubi, bj-complss, bj-compleq seem important enough to be moved to Main. What do you think?

I will agree, but it would be nice if they had some applications.  Note that abvor0 (never used) and abf (used 7 times) are special cases of these.  Note that bj-abfalbi is a variant of abn0 (used 36 times).

> I agree with @wlammen and propose to remove that sentence. Waiting for @nmegill to chime in.

I think the sentence can be removed and the credit can remain with @wlammen.

People "borrow" proof techniques and subproofs from other theorems all the time - e.g. when a subproof is discovered to be common to many theorems.  Our practice has been for the "Contributed by" credit to go to the person adding the theorem to set.mm, regardless of where the proof came from.  If the subproof is subsequently used to shorten the parent proof it was borrowed from, anyone tracing back the dates can see that the parent theorem (presumably including the extracted subproof or some form of it) was proved first.

Of course a proof shortened by an extracted subproof shouldn't have "Proof shortened by" since "minimize_with" will pick up the extracted subproof automatically.

I will merge this PR, and the above can be addressed in a separate PR.

PR #1956, opened 2021-03-20
wl-nfaev a specialized version of nfae

wlammen:


wlammen:
mathbox update and correction to text comments, as discussed yesterday

benjub:
The penultimate step of your proof of wl-nfaev could be added as ~hbaev and positioned right after "aev".  Could you add it, @wlammen ?

Also, before looking at theorems involving F/ , I think we should change its definition to df2 as agreed in https://groups.google.com/g/metamath/c/Ovxv2aXJOIM/m/BVU7SvgFBwAJ .  I began in ~df-bj-nf and the following theorems but I still have to do most of the work.  This would remove the use of ax-10 and ax-12 from many theorems (and add the use of these axioms to fewer, less elementary theorems).

wlammen:
I will do that, if it is agreed to be valuable enough to be added to Main, even without an application. Edit: It can be used to simplify ~aev , so its proper position is right before it. I'll create  a pull request as soon as I have time for it.

nmegill:
> The penultimate step of your proof of wl-nfaev could be added as ~hbaev and positioned right after "aev". Could you add it, @wlammen ?

I'll agree to this.

> 
> Also, before looking at theorems involving F/ , I think we should change its definition to df2 as agreed in https://groups.google.com/g/metamath/c/Ovxv2aXJOIM/m/BVU7SvgFBwAJ . I began in ~df-bj-nf and the following theorems but I still have to do most of the work. This would remove the use of ax-10 and ax-12 from many theorems (and add the use of these axioms to fewer, less elementary theorems).

I'll agree that df-nf can be changed accordingly.  I assume it is essentially swapping df-nf and nf2. 
 Some wording in the df-nf description would have to be changed, in particular the referenced to nf2.  Note that the current definition is also stated in mmset.raw.html and should be changed to the new one.

benjub:
Actually, Wolf's proof proves the more general scheme

  A. x x = y -> A. z A. t u = v , DV(x,y)

Maybe one can add this as ~aev2 with a comment like: "this is a generalization of the statement that could be labeled hbaev (see ~ hbae ). It is clear from the proof that one can prove similar statements with an arbitrary number of universal quantifiers in the consequent (the first in the series being ~aevlem2, ~aev, ~aev2)."

As a consequence, maybe relabel aevlem2 to aev0 (because it is more than a lemma), and aevlem1 to aevlem ?

PR #1960, opened 2021-03-22
hbaev

wlammen:
1. add hbaev
2. hbaev could be used in aev, but this increases the proof length
3. upgrade aevlem2 from a lemma to a theorem (I did not add a renaming record in the beginning of set.mm, as it was recently introduced)
4. I still think it is worth to add a comment to ax13v along the lines of wl-ax13v

Roadmap
Prove ( y = t -> A. x ( x = t -> x = y ) ) dv(y, t) (x, t), likely by cases (A. x x = y , A.x x = t are easy, the rest needs ax13v)).

Provide an instance of ax-12 with ph set to x = t. A note here: I do not connect ax-12 with substitution, I rather see it as a manifesto to never introduce a predicate that is finer grained than equality. This is by the way a reason why you must not introduce disjoint(x, ph) as an ordinary logical expression, as this has the potential to cut through the equivalence classes generated by =. I am pretty sure there is a huge amount of literature upon this topic. Unfortunately I have no background in logic, and have to come across this by myself. Would have loved to read this on the Metamath pages instead.

Derive ax6e based on ax13v only.

I work full time in a job, maybe I won't get to this before Easter.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19076,14 +19076,12 @@ Proof modification of "bj-abeq1" is discouraged (36 steps).
 Proof modification of "bj-abeq2" is discouraged (47 steps).
 Proof modification of "bj-abf" is discouraged (13 steps).
 Proof modification of "bj-abfal" is discouraged (54 steps).
-Proof modification of "bj-abfalbi" is discouraged (41 steps).
 Proof modification of "bj-abid2" is discouraged (14 steps).
 Proof modification of "bj-ablsscmn" is discouraged (10 steps).
 Proof modification of "bj-ablsscmnel" is discouraged (5 steps).
 Proof modification of "bj-ablssgrp" is discouraged (10 steps).
 Proof modification of "bj-ablssgrpel" is discouraged (5 steps).
 Proof modification of "bj-abtru" is discouraged (29 steps).
-Proof modification of "bj-abtrubi" is discouraged (45 steps).
 Proof modification of "bj-aecomsv" is discouraged (16 steps).
 Proof modification of "bj-aev" is discouraged (39 steps).
 Proof modification of "bj-aevlem1v" is discouraged (42 steps).
@@ -19157,8 +19155,6 @@ Proof modification of "bj-clelsb3" is discouraged (63 steps).
 Proof modification of "bj-cleqhyp" is discouraged (21 steps).
 Proof modification of "bj-cmnssmnd" is discouraged (36 steps).
 Proof modification of "bj-cmnssmndel" is discouraged (5 steps).
-Proof modification of "bj-compleq" is discouraged (42 steps).
-Proof modification of "bj-complss" is discouraged (33 steps).
 Proof modification of "bj-con4iALT" is discouraged (13 steps).
 Proof modification of "bj-consensusALT" is discouraged (30 steps).
 Proof modification of "bj-csbprc" is discouraged (47 steps).
@@ -19169,7 +19165,6 @@ Proof modification of "bj-df-nul" is discouraged (11 steps).
 Proof modification of "bj-df-v" is discouraged (29 steps).
 Proof modification of "bj-dfclel" is discouraged (11 steps).
 Proof modification of "bj-dfcleq" is discouraged (11 steps).
-Proof modification of "bj-dfnf5" is discouraged (57 steps).
 Proof modification of "bj-disjsn01" is discouraged (18 steps).
 Proof modification of "bj-dral1v" is discouraged (36 steps).
 Proof modification of "bj-drex1v" is discouraged (42 steps).


benjub:
@wlammen Have you seen my comment at #1956?

wlammen:
@benjub That's why I renamed aevlem2 to aeveq.  And yes, one can stack up lots of quantifiers by iteratively applying hbaev and alrimih. I personally see no application of this.

benjub:
I meant that you proved

hbaev ( A. x x = y -> A. z A. x x = y ) DV(x,y)

but you can prove instead (it seems to me) the more general

aev2 ( A. x x = y -> A. z A. t u = v ) DV(x,y)

wlammen:
This is applying hbaev and alrimih to aev.

benjub:
Yes. My point is that hbaev is an instance of aev2, so aev2 is worth adding.  More generally, I would say that if a statement is proved in set.mm and is an instance of a more general statement, then that statement should also be in set.mm (in addition to or in place of it).

wlammen:
hmm.  The idea of hb.. is found all over the place in set.mm. I do not like the idea to compromise this for something I hardly see an application for.  However, feel free to add a pull request of your own to reflect your point.

benjub:
@wlammen Will do. Thanks for the roadmap and good luck.

nmegill:
> 4\. I still think it is worth to add a comment to ax13v along the lines of wl-ax13v

Please go ahead and do that if you wish.

> A note here: I do not connect ax-12 with substitution, I rather see it as a manifesto to never introduce a predicate that is finer grained than equality.

The connection to proper substitution was mentioned in Tarski 1965 p. 62 where (using set.mm's notation) he says: "Let [ y / x ] ph be the formula obtained from the formula ph by proper substitution of the variable y for the variable x. ... Then, assuming that x and y are two distinct variables,  [ y / x ] ph proves to be equivalent to the formula A. x ( x = y -> ph )."

> This is by the way a reason why you must not introduce disjoint(x, ph) as an ordinary logical expression, as this has the potential to cut through the equivalence classes generated by =.

I'm not clear about what you mean by "cut through equivalence classes generated by =".  There are 2 levels of equality involved with set.mm's language:  the equality meaning the same object-language variable (at the $d level,  meaning "not distinct") or the same set of set theory (at the object-language level, the "=" sign in set.mm).

Perhaps I'm repeating what you know, but keep in mind that the variables in set.mm aren't the actual (object) variables of logic, and "disjoint(x, ph)" is a metalogical proviso that isn't part of the object language, which has variables v_0, v_1... .  The object language has no wff variables or $d provisos, and any two object language variables are distinct by definition if they have different names.  When we say "disjoint(x, y)" we mean that metavariables x and y must be instantiated with object-language variables having different names, say v_5 and v_3, whether or not they represent the same set as in v_5 = v_3.

As for introducing disjoint(x, ph) as "an ordinary logical expression" (by which I assume you mean as part of a wff rather than a separate side condition), as you know we can use the distinctor antecedent -. A. x x = y in place of disjoint(x, y).  The problem with distinctors is that you have to introduce them for each dummy variable used in a proof, and they remain as antecedents of the final theorem obtained even if the theorem itself doesn't mention the dummy variables.  There is no way to get rid of distinctors containing dummy variables without using $d's, and there is no limit to how many dummy variables might accumulate (per a theorem of Andreka).

> I am pretty sure there is a huge amount of literature upon this topic.

Actually I don't think there is much, at least not clearly stated.  Textbooks can be a little informal on the matter, casually interchanging object-language variables with (meta)variables ranging over them.  Tarski's 1965 paper is very careful about the distinction; his language is very precise, although you need to read it carefully, keeping in mind that when he says, "Let x [he uses alpha] be a variable", the x itself is a metavariable or placeholder for an actual (object-language) variable from the set v_0, v_1,...  And when he says "Let ph be a formula", again he is just using ph as a placeholder for some object-language wff like "-. A. v_5 v_5 = v_3" constructed from v_0, v_1,... . (Indeed, he uses the symbols "v_0, v_1,..." for object-language variables, p. 64.)

wlammen:
Thanks for your elaborate answer
> > A note here: I do not connect ax-12 with substitution, I rather see it as a manifesto to never introduce a predicate that is finer grained than equality.
> 
> The connection to proper substitution was mentioned in Tarski 1965 p. 62 where (using set.mm's notation) he says: "Let [ y / x ] ph be the formula obtained from the formula ph by proper substitution of the variable y for the variable x. ... Then, assuming that x and y are two distinct variables, [ y / x ] ph proves to be equivalent to the formula A. x ( x = y -> ph )."

Substitution [ x / y ] is in Metamath a slick, but slightly complex procedure, and I wondered from the onset why it has to backed up by a dedicated axiom, what the alternatives were, or the impact of an omission would be. I finally took on the view, that ax-12 enforces equality to be the finest grained predicate possible, which is a bit far away from substitution.  This interpretation is actually quite useful. In the context of the roadmap I gave in my first post here, for example, I envision there could be headroom for elimination since it does not make much sense to compare equality to itself. Or I can judge whether any predicate candidate (by which I in fact mean, extending wff grammar, and axioms describing its behaviour), such as 'is element of' or 'disjoint(x, ph)' would be sound. And, voila, one is, the other not. I need not even think about the introduction of 'disjoint', it would contradict ax-12.
> 
> > This is by the way a reason why you must not introduce disjoint(x, ph) as an ordinary logical expression, as this has the potential to cut through the equivalence classes generated by =.
> 
> I'm not clear about what you mean by "cut through equivalence classes generated by =". There are 2 levels of equality involved with set.mm's language: the equality meaning the same object-language variable (at the $d level, meaning "not distinct") or the same set of set theory (at the object-language level, the "=" sign in set.mm).

It is the latter. From a philosopher's point of view, there might be just one thing like an empty set, and its representation does not matter.  But Metamath is a pure formalism, completely indifferent to the underlying theory. And its 'objects', or discourse of domain, are just pieces of text. Here the empty set is represented by one of { x | F. }, { y | -. y = y } and so on. Equality of sets creates an equivalence relation on well-formed set terms (see https://en.wikipedia.org/wiki/First-order_logic#First-order_logic_without_equality). 'disjoint' could discern { x | F. } and { y | -. y = y }, hence it splits up the equivalence class of the empty set, or is finer granular than equality. In addition, it may also have other disadvantages, that you depict in your answer. For me, it is not a contender for a sound predicate any more. I mentioned it, because I some years ago thought, it would be nice to have it be part of the logical language, instead of living as side conditions. I found the answer, as described here, but it wasn't that obvious to me.

It is far past midnight now. I think I should stop here, and revisit your answer at a time more appropriate.

benjub:
> I finally took on the view, that ax-12 enforces equality to be the finest grained predicate possible, which is a bit far away from substitution.

Not that far away.  This is the principle of "indiscernibility of identicals".  I think that what encodes substitution is not ax-12 but rather bj-ax12.  It is strictly weaker (sp cannot be proved when ax-12 is weakend to bj-ax12).  @wlammen noticed in #1920 that ax-12 also implies axc112, and this is not the case for bj-ax12.  So at the meta level, @wlammen is right to say that ax-12 says something more than the substitution property, namely axc112, which shows that one can substitute "all-equal" variables in a quantifier.

nmegill:
The original version of ax-12 was axc15 (ax-c15) with a distinctor antecedent.  Later I noticed that the distinctor could be dropped if we quantified the first ph with A. y, resulting in the current shorter ax-12.

Originally I toyed with the idea of using ax-12 to derive ax12v (which matches Monk's axiom C8) then derive back ax-12, sort of like we do with ax13v to recover ax-13.  But later I found that we could also use ax-12 to derive axc11, which at that time was a separate axiom; this probably cannot be done with ax12v because, as @wlammen noticed, axc112 exploits the bundling of x and y in ax-12.

One way to look at this is that ax-12 embodies 2 kinds of substitution.  I used to call  ax-c15 "variable substitution" and ax-c11 "quantifier substitution".  The current ax-12 combines them into one so I just call it "substitution".

Another reason we don't have ax12v (or bj-ax12) + ax-c11 in place of ax-12 is that I wanted to have ax-5 as the only axiom involving disjoint variables.  $d's can be confusing to beginners, and (if we disregard our ax6v, ax7v, ax8v, ax9v, and ax13v weakenings, which are not theoretically necessary) it makes ax-5 the only source of $d's.  Rightly or wrongly I thought that a single axiom with a $d would be less intimidating to a beginner.  It does give us a significantly more compact axiom system display on the MPE home page without "where ? and ? are distinct" scattered throughout.

Eventually it would be nice to have a clean derivation of:

ax-12 => ax12v + axc11 => axc15 + axc11 => ax12

(Some parts of this such as ax12 are now in my mathbox derived from the old ax-c\* system.)  This would allow ax-12 to be swapped with ax12v + axc11 (restated as axioms) without much difficulty, in case someone prefers them over the economy of ax-12 in terms of fewer axioms and fewer $d's.  It might be possible to use the simpler axc11n in place of axc11 above but I'm not sure.

BTW the label "axc112" could be confusing - maybe we should call it axc11-2 or axc11r?

nmegill:
I wrote:

> Eventually it would be nice to have a clean derivation of:
> 
> ax-12 => ax12v + axc11 => axc15 + axc11 => ax12

Actually wl-ax12v is even weaker with both $d x y and $d ph y, so I think it can replace ax12v above.

nmegill:
> I wrote:
> 
> > Eventually it would be nice to have a clean derivation of:
> > ax-12 => ax12v + axc11 => axc15 + axc11 => ax12
> 
> Actually wl-ax12v is even weaker with both $d x y and $d ph y, so I think it can replace ax12v above.

I see @wlammen has essentially done the above with his derivation leading to wl-ax12.  It can be moved to main at his convenience, with the "wl-" prefixes removed.  The ax12 in my mathbox can be renamed to ax12o.

wlammen:
Wait, wait. The rederivation of ax-12 is NOT complete. It still uses ax-12. Some references like sp can be resolved using wl-19.8a, but references to axc112 not yet. This was experimental work, unfinished.
OK, with the aid of axc11 it might be possible. This still has to be carefully checked.

benjub:
I'm in favor of moving to Main @wlammen's wl-ax12v, which is Step 4 of the current proof of ax12v. Maybe relabel it ax12vv (since there are two DV conditions), and place it between ax-12 and ax12v (i.e., keep both). And use it whenever possible, for instance in 19.8a, ax12a2, exsb, mo2v, 2eu6, bj-ssbequ1, bj-ssbid1ALT, rexsb.

I'm in favor of renaming axc112 to axc11r.

PR #1961, opened 2021-03-22
Misc

benjub:
@nmegill : Do you agree with the method of having a subsection at the end for the aliases, which is created in PR? Maybe you want a different wording of that section?

"dummylink" has to be changed to "a1ii" in the page milpgame.html.  There is also iset.mm and nf.mm if we want them to follow set.mm (relabel? keep a copy in a new section at the end?).

See https://groups.google.com/g/metamath/c/_qiufu1l-Pw/m/5nwBEx8zBQAJ

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15776,6 +15776,7 @@ New usage of "drngcatALTV" is discouraged (0 uses).
 New usage of "drngoi" is discouraged (2 uses).
 New usage of "dtruALT" is discouraged (0 uses).
 New usage of "dtruALT2" is discouraged (0 uses).
+New usage of "dummylink" is discouraged (0 uses).
 New usage of "dvadiaN" is discouraged (1 uses).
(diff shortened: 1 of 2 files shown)


nmegill:
> @nmegill : Do you agree with the method of having a subsection at the end for the aliases, which is created in PR? Maybe you want a different wording of that section?

It's ok.  Hopefully the entries will become obsolete after some time and can be deleted.  (There are quite a few labels that have changed over time and can always be found in the label change list at the top of set.mm.  I doubt we want to reproduce that section in an alias list, only a few important ones.)

> 
> "dummylink" has to be changed to "a1ii" in the page milpgame.html. 

I'll take care of that.  I also need to change it in eimm.c,

> There is also iset.mm and nf.mm if we want them to follow set.mm (relabel? keep a copy in a new section at the end?).

I think you can just change them to the new labels.  No need to add an alias section since they aren't used in those databases but are just accidental because they started from set.mm.

Issue #1940, opened 2021-02-19
Prove Cauchy completeness of positive reals in iset.mm


jkingdon:
I don't know if other people are interested in this, but I thought I should make an issue to describe what is hanging out on my https://github.com/jkingdon/set.mm/tree/iset-caucvgprpr branch and the progress underway to finish it.

The previously completed work is to prove that a Cauchy sequence of positive rationals has a positive real limit (where positive real means `P.`, the "temporary" positive reals in iset.mm), as seen in http://us.metamath.org/ileuni/caucvgpr.html . The sequence has to have a rate of convergence as described in the hypothesis of that theorem.

The work currently underway is to allow the members of the sequence to be positive reals rather than positive rationals. Here's the current statement of the theorem to be proved:

```
    caucvgprpr.f $e |- ( ph -> F : N. --> P. ) $.
    caucvgprpr.cau $e |- ( ph ->
      A. n e. N. A. k e. N. ( n <N k ->
        ( ( F ` n ) <P ( ( F ` k ) +P.
          <. { l | l <Q ( *Q ` [ <. n , 1o >. ] ~Q ) } ,
            { u | ( *Q ` [ <. n , 1o >. ] ~Q ) <Q u } >.
        ) /\
        ( F ` k ) <P ( ( F ` n ) +P.
          <. { l | l <Q ( *Q ` [ <. n , 1o >. ] ~Q ) } ,
            { u | ( *Q ` [ <. n , 1o >. ] ~Q ) <Q u } >.
        ) ) ) ) $.
    caucvgprpr.bnd $e |- ( ph -> A. m e. N. A <P ( F ` m ) ) $.
    caucvgprpr $p |- ( ph ->
        E. y e. P. A. x e. P. E. j e. N. A. k e. N. ( j <N k ->
        ( ( F ` k ) <P ( y +P. x ) /\
          y <P ( ( F ` k ) +P. x ) ) ) ) $= ? $.
```

This is quite similar to caucvgpr including the lower bound to avoid the case of a sequence of positive reals which "converges" to zero.

[...]

PR #1962, opened 2021-03-23
Misc

benjub:
Sequel of #1961: iset.mm and nf.mm have been modified to follow set.mm (fyi: @jkingdon , @sctfn , @scott-fenton ), namely:
a1ii --> 2a1i
dummylink --> a1ii
~mathbox statement in set.mm and iset.mm changed to |- ph => |- ph (it used to be proved from equid so was showing up in its "show usage")

@nmegill is taking care of milpgame.html and eimm.c ; @digama0 mentioned "dummylink" might still be in mmj2, so this should be changed to a1ii too.

@nmegill : indeed the section for aliases should be limited to cases where many external references may exist (it will most probably stay limited to the two present cases: dummylink and id1).

### DIFF
--- a/nf.mm
+++ b/nf.mm
@@ -45,6 +45,7 @@ $(
 
 $)
 
+
 $(
 #*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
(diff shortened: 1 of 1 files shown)


jkingdon:
@benjub Thanks for the heads-up - the change to iset.mm looks good to me. Doesn't break anything and merges fine with the branch from #1940 .

PR #1963, opened 2021-03-24
move ax12 related stuff mathbox to main, see pull request #1960

wlammen:
1. Replace current ax12 related versions by my mathbox versions
2. renamed axc112 to axc11r
3. renamed NM's ax12 to ax12o

Not finished yet. Theorems like equs5a, equs5e still refer to ax-12 directly. These have to be moved to a different place for accessing ax12 or have to be re-proven with a somewhat more adequate technique. sb4a, sb4e are dependent on these and effected by this change, too.

Request for comments.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -918,7 +918,7 @@
 "atsseq" is used by "atnemeq0".
 "atssma" is used by "atordi".
 "ax-10" is used by "hbn1".
-"ax-13" is used by "(Contributed".
+"ax-13" is used by "ax13v".
 "ax-3" is used by "con4".
 "ax-3" is used by "dfbi1ALT".
 "ax-4" is used by "alim".
--- a/discouraged
+++ b/discouraged
@@ -918,7 +918,7 @@
 "atsseq" is used by "atnemeq0".
 "atssma" is used by "atordi".
 "ax-10" is used by "hbn1".
-"ax-13" is used by "ax13v".
+"ax-13" is used by "(Contributed".
 "ax-3" is used by "con4".
 "ax-3" is used by "dfbi1ALT".
 "ax-4" is used by "alim".
@@ -960,7 +960,7 @@
 "ax-c11n" is used by "axc11-o".
 "ax-c14" is used by "ax12el".
 "ax-c14" is used by "ax5el".
-"ax-c15" is used by "ax12".
+"ax-c15" is used by "ax12o".
 "ax-c16" is used by "ax5el".
 "ax-c16" is used by "ax5eq".
 "ax-c16" is used by "axc11n-16".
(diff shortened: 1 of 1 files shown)


benjub:
You can shorten the proof of ax12v2 by using ax6evr, as you noticed in a previous PR.

(I think it would have been less disruptive to keep ax12v as is and introduce the new statement as ax12vv, but it's your call since you're the one who did the job.)

nmegill:
> (I think it would have been less disruptive to keep ax12v as is and introduce the new statement as ax12vv, but it's your call since you're the one who did the job.)

My opinion is that it was an oversight for us (me) not to have included $d y ph in the first place, and that ax12v is the right name for the one with both $d's.  Similarly, we don't have ax13v and ax13vv with one and two $d's, instead we have just ax13v with both $d's.

Ordinarily we wouldn't include ax12v2 at all (just like we don't have an ax13v2 with one $d), except that ax12v2 matches Axiom C8 of [Monk2] p. 105 exactly, so it is informative for that reason.  (This can be stated in the ax12v2 comment.)

Indeed, it might be argued that the omission of $d y ph in Monk's C8 was an oversight, since the Tarski/Kalish-Montague/Monk trilogy of papers seemed to value the weakest axioms possible (e.g. they didn't object to the bundling in ax-6 when its additional strength was needed, per the footnote on the first page of the Kalish-Montage paper).  I am just conjecturing this, though; maybe I missed some detail in one of Monk's proofs that depends on the absence of $d y ph.

benjub:
@nmegill : Similarly to what I wrote in https://groups.google.com/g/metamath/c/V296RJ0uEbo/m/mTi_Pn_8BwAJ , I think the treatment of DV conditions by Tarski, Kalish--Montague and Monk is less about maximal weakening than about compatibility with a logic with terms.  In the present case, the "y" of ax12v should be considerered a term metavariable, so DV(y,ph) would make no sense for Tarski et al. (see our recent discussion about DV(ph,ps)).

That's why I denoted it by "t" in http://us2.metamath.org/mpeuni/bj-ax12.html and http://us2.metamath.org/mpeuni/bj-ax12ssb.html

nmegill:
> @nmegill : Similarly to what I wrote in https://groups.google.com/g/metamath/c/V296RJ0uEbo/m/mTi_Pn_8BwAJ , I think the treatment of DV conditions by Tarski, Kalish--Montague and Monk is less about maximal weakening than about compatibility with a logic with terms. In the present case, the "y" of ax12v should be considerered a term metavariable, so DV(y,ph) would make no sense for Tarski et al. (see our recent discussion about DV(ph,ps)).

I kind of see your point, and I'll no longer conjecture that it may have been an oversight by Monk.

However, terms have no place in the set theory that set.mm's FOL is used for (class variables serve that role).  For our purposes, I think it is useful to state maximally weak versions (w.r.t DVs) with a "v" suffix, so the user can depend on "v" suffixed axioms being maximally weak (no need to check if a "vv" version  exists).  For one thing, maximally weak versions might be useful when trying to find independence proofs since there would be less bundling to model.

If terms are introduced, that could also make ax7v somewhat odd with DV(term, term) possible, but again we have no need for terms in set.mm so it is not a problem.

PR #1964, opened 2021-03-25
fix typos; add ref to Monk; rename ax12v2 to axc8

wlammen:
based on comments in PR1963

### DIFF
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -5712,7 +5712,7 @@ Section 107 of the United States Copyright Act (Title 17 of the
 &nbsp;</FONT></TD>
 
 <TD NOWRAP ALIGN=CENTER>
-<I><FONT SIZE=-1>This page was last updated on 24-May-2021.</FONT></I>
+<I><FONT SIZE=-1>This page was last updated on 24-Mar-2021.</FONT></I>
 <BR>
 <FONT FACE="ARIAL" SIZE=-2>Your comments are welcome: Norman Megill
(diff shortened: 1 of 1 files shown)


benjub:
What about the proof shortening using ax6evr (see my comment there) ? (maybe it actually does not work, I didn't check)

Nitpick: the line "24-Mar-21 wl-ax12   ax12       moved from WL's mathbox to main set.mm" lacks a space before "moved" to be aligned with the rest.

wlammen:
"What about the proof shortening using ax6evr". Forgot about that. Thanks for the hints. Edit: I don't see (anymore), how ax6evr can be of advantage. Another Edit: I can it now again.

nmegill:
A minor naming issue:  My 1995 paper's numbering was used for the obsolete ax-c* axioms, so "axc8" could be confusing.  Monk's axiom C8 is not the same as axiom C8 in my paper (which corresponds to our present ax-7) and actually does not match any of the paper's axioms.  I would recommend that "axc8" be changed to "monkc8" or similar in your next PR.

wlammen:
If you don't mind, I would revert it to ax12v2, as this is a somewhat speaking name. So far we have no monk* theorems refering to that particular source, so I suggest not to begin with. Edit: in case of no answer I will use monkc8.

nmegill:
> If you don't mind, I would revert it to ax12v2, as this is a somewhat speaking name. So far we have no monk* theorems refering to that particular source, so I suggest not to begin with. Edit: in case of no answer I will use monkc8.

You can change it back to ax12v2.

PR #1965, opened 2021-03-25
New approach for graph theory (continued) - unnecessary assumption G e. W removed

avekens:
Main set.mm:
* ~difex, ~mpbirand, ~ ffund moved from GS's mathbox to main set.mm

* unnecessary assumption G e. W removed
* Section comments added for "Eulerian paths" and "The Konigsberg Bridge problem" adjusted

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -20452,7 +20452,7 @@ Proof modification of "trubifalOLD" is discouraged (15 steps).
 Proof modification of "trunanfalOLD" is discouraged (19 steps).
 Proof modification of "truniALT" is discouraged (183 steps).
 Proof modification of "truniALTVD" is discouraged (220 steps).
-Proof modification of "umgr2adedgwlkonALT" is discouraged (276 steps).
+Proof modification of "umgr2adedgwlkonALT" is discouraged (294 steps).
 Proof modification of "un0.1" is discouraged (21 steps).
 Proof modification of "un01" is discouraged (18 steps).
 Proof modification of "un10" is discouraged (18 steps).
@@ -20464,7 +20464,7 @@ Proof modification of "unipwrVD" is discouraged (41 steps).
 Proof modification of "unisnALT" is discouraged (146 steps).
 Proof modification of "unitgOLD" is discouraged (95 steps).
 Proof modification of "upgr0eopALT" is discouraged (55 steps).
-Proof modification of "upgr1eopALT" is discouraged (134 steps).
+Proof modification of "upgr1eopALT" is discouraged (126 steps).
 Proof modification of "usgedgimpALT" is discouraged (142 steps).
 Proof modification of "usgedgleordALT" is discouraged (127 steps).
 Proof modification of "usgedgvadf1ALT" is discouraged (476 steps).


nmegill:
In ~ difex, our convention has been to state the hypothesis difex.1 as "$e |- A e. _V $." rather than "$e |- A e. V $.".  This is because "$e |- A e. _V $." is almost always satisfied using an existence theorem stating "... e. _V", and a hard-coded _V in the $e saves a couple of syntax building steps that substitute _V into V.  (A very minor savings to be sure, but still a savings.)

Could you change that in your next PR?  Thanks.

avekens:
Hi Norm, I am a little bit confused about your comment: is this really our convention? In section "Conventions" it is stated (as bullet point "Is-a-set."):

> "` A e. _V ` " should be read "Class ` A ` is a set (i.e. exists)." This is a convention based on Definition 2.9 of [Quine] p. 19.  See ~ df-v and ~ isset . However, instead of using ` I e. _V ` in the antecedent of a theorem for some variable ` I ` , we now prefer to use ` I e. V ` (or another variable if ` V ` is not available) to make it more general. That way we can often avoid needing extra uses of ~ elex and ~ syl in the common case where ` I ` is already a member of something."

This is quite contrary to your statement. Or does "hypothesis" vs. "antecedent" makes the difference?

Currently, we have 841 hypotheses with "... e. _V" in set.mm, but only 58 (52 in mathboxes) with "... e. V". So it seems that there is a (hidden) convention.

Maybe the bullet point in section "Convention" should be enhanced accordingly. I can do this when I provide the change of ~ difex (and other cases in my mathbox) in my next PR.

nmegill:
"Hypothesis" vs. "antecedent" makes a difference.

Hypotheses should be "$e |- A e. _V $." rather than "$e |- A e. V $." (as the 841 cases suggest).  ​The ​6 cases not in mathboxes should be changed to _V, and ideally those in mathboxes as well to be consistent in case they are imported.

Antecedents, on the other hand, should be "|- ( A e. V -> ..." rather than "|- ( A e. _V -> ...".  That is because it is very common to have an antecedent requiring membership in something other than _V, and using "V" instead of "_V" to hold the "something" saves an application of elex+syl.

If you look at e.g. ~pwexg, note that the antecedent is "e. V", but the consequent must be the fixed "e. _V".  Since the consequent is what is used to satisfy a hypothesis (after ax-mp or in this case using ~pwex directly), there is rarely any advantage to having "e. V" instead of "e. _V" in a hypotheses.

Feel free to update ~conventions if you think it's unclear.

avekens:
OK, I started already to adapt the 6 cases with hypotheses $e |- A e. V $.". On (hopefully final) question: This convention does not hold for hypotheses of theorems in deduction form, does it? For such theorems, we still will have "$e |- ( ph -> B e. V ) $." rather than "$e |- ( ph -> B e. _V ) $.", won't we? I count 919 such hypotheses in set.mm ...

nmegill:
> OK, I started already to adapt the 6 cases with hypotheses $e |- A e. V $.". On (hopefully final) question: This convention does not hold for hypotheses of theorems in deduction form, does it? For such theorems, we still will have "$e |- ( ph -> B e. V ) $." rather than "$e |- ( ph -> B e. _V ) $.", won't we? I count 919 such hypotheses in set.mm ...

This is a good question, and I don't know the answer.  It may not be so clear-cut (i.e. lead to slightly shorter proofs in most cases) as the non-deduction cases.  Does anyone have a feel for this based on experience?  We should probably leave the deduction cases alone for now and not have a convention for them yet.  Is there a general trend in set.mm for these (i.e. 919 cases vs. ? cases)?

sctfn:
FWIW, I've been using ( ph -> B e. V ) for deduction forms in nf.mm. It's
come in handy where I've got a deduction that something is a natural.

nmegill:
> FWIW, I've been using ( ph -> B e. V ) for deduction forms in nf.mm. It's come in handy where I've got a deduction that something is a natural.

That makes sense, because a deduction is really a closed theorem form in disguise i.e. the antecedent will include B e. V when the hypothesis is eliminated with simpxxx.  And that seems to be the most frequent convention in set.mm:

```
$ grep '\$e |- ( [a-z][a-z] -> [A-Z] e\. _V ' set.mm | wc -l
171
$ grep '\$e |- ( [a-z][a-z] -> [A-Z] e\. V ' set.mm | wc -l
953
$ grep '\$e |- ( [a-z][a-z] -> [A-Z] e\. [A-Z] ' set.mm | wc -l
3498
```
(In the last case, many of the hypotheses are for membership in a defined class variable rather than a set existence hypothesis.)

So overall it seems that using 'e. V' rather than 'e. _V' could be the recommendation, though maybe not necessarily a requirement, for deduction form cases.  I don't think it is worth changing the 'e. _V' deduction cases except as a low priority project and maybe after some analysis to see if there would be a benefit.

benjub:
@avekens I hope you did not modify theorems within mathboxes?  I have a few theorems with hypothesis |- A e. V (and not _V) but for good reasons (http://us2.metamath.org/mpeuni/bj-isseti.html , http://us2.metamath.org/mpeuni/bj-rexcom4b.html).  Thanks

avekens:
@benjub I will only change the 7 theorems in main set.mm which had "$e |- A e. V $." hypotheses in my next PR. Aftereards, I will change the affected theorems in my mathbox only. I will not change theorems in other mathboxes (unless they are moved to main set.mm).

PR #1966, opened 2021-03-26
equs5a/e depend on ax12 instead of ax-12 directly

wlammen:
1. let ax12v and ax11r be the only theorems referring to ax-12 directly (except for OLD/ALT or mathbox theorems)
2. rename axc8 to ax12v2

Point 1 was achieved by moving equs5a/e further down set.mm beyond ax12 (which in turn uses ax13v), and replace ax-12 with the synthesized version ax-12.  This move lets these theorems depend on weaker axioms. Downside: Because of their close similarity to ax-12, and simple proofs based on it, usage of ax12 drags an extra ax-13 into these theorems.


benjub:
> Downside: Because of their close similarity to ax-12, and simple proofs based on it, usage of ax12 drags an extra ax-13 into these theorems.

This is a downside indeed  Maybe add to the comments of both equs5a and equs5e : "Note that ~ ax-13 is used only via ~ ax12 , so this theorem could be proved from Tarski's FOL and ~ ax-10 and ~ ax-12 only."

wlammen:
@benjub Alternatively, one could add an ALT proof to illustrate this fact. In the end I thought these theorems are not important enough to justify a special handling. The problem is general. At any time you may introduce a helper theorem, that is a close variation of a bundled axiom, and its proof will then appear overly complex when derived from the unbundled axioms. In the end, what does the list of axioms a theorem is proven from say?  Is it the minimal list of bundled or unbundled axioms?  On http://us2.metamath.org:88/mpeuni/conventions.html paragraph "Minimizing axioms and the axiom of choice" we read "We prefer proofs that depend on fewer and/or weaker axioms, even if the proofs are longer."  I am unsure what the consequences are in this particular case.

benjub:
As @wlammen suggested, I think it is better to have both versions, that is, the main and the *ALT ones for both equs5a and equs5e. (It is currently only indicated in the comments, see #1967.)

There is a proof from Tarski+ax12v+ax-13 and a proof from Tarski+ax-12. Thanks to @wlammen we know that the former system is more powerful (conjecturally, strictly more powerful) than the latter. Therefore, I think the main proofs should be the ones from Tarski+ax-12 and the ALT proofs should be the new ones (plus, in the axiom usage, the only visible difference is that one proof uses ax-13 and the other not).

I can do the change if @wlammen and @nmegill agree.

wlammen:
I'd rather have it the other way round: the new ones are the regular ones, and the ALT version illustrates how immediate usage of a bundled axiom leads to shorter proofs in corner cases (as is perhaps already done with the axiom of choice). I am in favour of rigorously deriving theorems from unbundled axioms. They more precisely pinpoint individual aspects of logic, and ax-12 is no exception.  ax12v and axc11r represent two very different ideas, and (to me) it is of interest, which one is used where. Any direct use of ax-12 blurs this information.

benjub:
The most important thing is that both be in set.mm. As for which is the main version and which is the ALT, I let @nmegill decide.

I agree with you (@wlammen) that preference should be on unbundled axioms, for the reason you mention. Note however that both proofs use bundled axioms (ax-12 or ax-13). Actually, equs5a and equs5e are not provable from ax-1--11.

As I said, to me, ax-12 should be replaced by {bj-ax12, sp} anyway .  Many theorems currently in set.mm that depend on ax-12 only depend on it via sp (including all theorems not involving equality).

nmegill:
> The most important thing is that both be in set.mm. As for which is the main version and which is the ALT, I let @nmegill decide.

I would lean towards using ALT for direct application of bundled versions, with the main name reserved for those proved from unbundled ones, when we have both cases to display.  So I more or less agree with @wlammen.  In the end, though, it is a trivial label change that we can discuss later.

> 
> I agree with you (@wlammen) that preference should be on unbundled axioms, for the reason you mention. Note however that both proofs use bundled axioms (ax-12 or ax-13). Actually, equs5a and equs5e are not provable from ax-1--11.
> 
> As I said, to me, ax-12 should be replaced by {bj-ax12, sp} anyway . Many theorems currently in set.mm that depend on ax-12 only depend on it via sp (including all theorems not involving equality).

As I've already discussed, I prefer our main ("official") display of axioms on the MPE Home Page to have as few axioms as possible, with the shortest possible axioms and the fewest possible $d's (i.e. only on ax-5).  I recognize that not everyone likes this approach, but that is my personal preference.

But the unbundled versions are important, of course, and I think we should display and discuss them in a separate section on the MPE Home Page.  We may want to include bj-ax12 and sp there.

Specifically, I am thinking of discarding "Appendix 7: Some Predicate Calculus Subsystems" (http://us.metamath.org/mpeuni/mmset.html#subsys) completely.  I don't think it has much relevance anymore since we achieved the current clean separation of Tarski's and auxiliary axioms.  We would add a new appendix (it could be earlier than the discarded Appendix 7, maybe after Appendix 3) with the unbundled versions.

benjub:
> > The most important thing is that both be in set.mm. As for which is the main version and which is the ALT, I let @nmegill decide.
> 
> I would lean towards using ALT for direct application of bundled versions, with the main name reserved for those proved from unbundled ones, when we have both cases to display.

Ok. Note that both proofs use bundled axioms anyway. The theorems equs5a/e themselves are bundled, so there is no way around that.

I'm trying to find the original proofs to restore them as the *ALT versions, but apparently they were not kept as *OLD proofs ? And even on the "us" server (as opposed to "us2") they seem to have been already updated ? I'm not a git expert, and I don't know how to retreive them...

**Edit:** Even on http://us2.metamath.org/old/mpeuni/ , the proofs have already been updated...

> As I've already discussed, I prefer our main ("official") display of axioms on the MPE Home Page to have as few axioms as possible, with the shortest possible axioms and the fewest possible $d's (i.e. only on ax-5). I recognize that not everyone likes this approach, but that is my personal preference.

Yes, both approaches have advantages.  Since all of this exists thanks to you, you of course get the final decision.

wlammen:
@benjub you can (a) use git log to find a hash value ## of the commit just before my first here (that has 5aed334, see below the line 'wlammen added 3 commits 2 days ago' and then use git checkout ## to restore a historical state on your file system; OR (b) you replay the current proofs in metamath.exe replacing ax12 with ax-12; OR (c) you copy the old contents from the 'files changed' button above, that are marked with a red background color.

To cut this short: I copied the old versions here
$( A property related to substitution that unlike ~ equs5 does not require a
     distinctor antecedent.  (Contributed by NM, 2-Feb-2007.) $)
  equs5a $p |- ( E. x ( x = y /\ A. y ph ) -> A. x ( x = y -> ph ) ) $=
    ( weq wal wa wi nfa1 ax-12 imp exlimi ) BCDZACEZFLAGZBEZBNBHLMOABCIJK $.

  $( A property related to substitution that unlike ~ equs5 does not require a
     distinctor antecedent.  (Contributed by NM, 2-Feb-2007.)  (Proof shortened
     by Wolf Lammen, 15-Jan-2018.) $)
  equs5e $p |- ( E. x ( x = y /\ ph ) -> A. x ( x = y -> E. y ph ) ) $=
    ( weq wa wex wi wal nfa1 hbe1 19.23bi ax-12 syl5 imp exlimi ) BCDZAEPACFZGZ
    BHZBRBIPASAQCHZPSATCACJKQBCLMNO $.

PR #1968, opened 2021-03-27
Hypotheses should be "$e |- A e. _V $." rather than "$e |- A e. V $."

avekens:
Revision of the hypotheses of the follwing theorems (in main set.mm), see also discussion in #1965:
* rabex2
* rab2ex
* difexi
* elmptrab2
* mptrabex
* trclubgi
* trclubi

Furthermore, the conventions (bullet point "Is-a-set.") are enhanced accordingly.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12260,6 +12260,7 @@
 "psubspi2N" is used by "pclclN".
 "psubspi2N" is used by "pclfinN".
 "psubspi2N" is used by "pclfinclN".
+"rabex2OLD" is used by "rab2exOLD".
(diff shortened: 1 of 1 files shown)


nmegill:
To see the effect of  "$e |- A e. V $." => "$e |- A e. _V $." I picked a couple of changed proofs at random and used "show proof .../size":
mdegldg:  1459 => 1426 characters
evlslem1:   9261 => 9143 characters
So (unless these were lucky hits or you did something else like "minimize") it seems A e. _V was the right choice.

PR #1971, opened 2021-03-28
shorten hypstkd

wlammen:


benjub:
Maybe you can relabel it to something more consistent ? (sorry, I don't have a good proposal to make...something like mpd2an ?)

nmegill:
> Maybe you can relabel [hypstkd] to something more consistent ? (sorry, I don't have a good proposal to make...something like mpd2an ?)

Since the "3" in our ~mpd3an* refers to the wff ~w3a, the "2" before the "an" doesn't quite fit.  I might suggest ~mpdan3 (we're discharging the 3rd conjunct in the antecedent).   ~mpdan2 would be ok too.

Also, in our ~mp* theorems, the minor premise almost always comes first.  I also suggest swapping the hypothesis order.

These changes can be made in a future PR.

wlammen:
mpidan was the best I could come up with, from mpid and 'an(d)'

PR #1973, opened 2021-03-29
Misc

benjub:
See commit messages.
Restore equs5(a|e)ALT (see #1967).
I shortened the proof of axc11n, avoiding the use of axc11nlem.  Then, I found a more direct and shorter proof of aevlem, using the new aevlem0 instead of axc11nlem, hence deprecating the latter. Since aevlem0 is a variant of axc11nlem, I copied the credits for that one to aevlem0.  Please @nmegill and @wlammen tell me if this is ok.

Also, @wlammen's proof of hbaev actually proves the more general statement bj-hbaev. So I'm in favor of putting that statement as hbaev (saying in the comment that it is a bit more general than what the "hb" series of labels say).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14533,7 +14533,6 @@ New usage of "bitr3VD" is discouraged (0 uses).
 New usage of "bitsfzolemOLD" is discouraged (0 uses).
 New usage of "bj-0" is discouraged (1 uses).
 New usage of "bj-1" is discouraged (0 uses).
-New usage of "bj-aev2ALT" is discouraged (0 uses).
 New usage of "bj-ax12iOLD" is discouraged (0 uses).
 New usage of "bj-ax6e" is discouraged (0 uses).
 New usage of "bj-axc4" is discouraged (0 uses).
@@ -19103,7 +19102,6 @@ Proof modification of "bj-ablssgrpel" is discouraged (5 steps).
 Proof modification of "bj-abtru" is discouraged (29 steps).
 Proof modification of "bj-aecomsv" is discouraged (16 steps).
 Proof modification of "bj-aev" is discouraged (39 steps).
-Proof modification of "bj-aev2ALT" is discouraged (33 steps).
 Proof modification of "bj-aevlemv" is discouraged (42 steps).
 Proof modification of "bj-alcomexcom" is discouraged (45 steps).
 Proof modification of "bj-ax12iOLD" is discouraged (20 steps).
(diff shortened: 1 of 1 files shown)


benjub:
@nmegill I revised morimOLD to make it prove what its comment said, and I updated its comment.  I propose to un-deprecate it, to "morim" or "moa1" (like ala1 and exa1).

nmegill:
> Also, @wlammen's proof of hbaev actually proves the more general statement bj-hbaev. So I'm in favor of putting that statement as hbaev (saying in the comment that it is a bit more general than what the "hb" series of labels say).

We could call bj-hbaev as hbaegv, like we do for axc16g vs. axc16.

> @nmegill I revised morimOLD to make it prove what its comment said, and I updated its comment. I propose to un-deprecate it, to "morim" or "moa1" (like ala1 and exa1).

"moa1" seems like a reasonable name.

BTW I'm wondering if it would also be a good idea to restore the original morimOLD (and remove "OLD"), esp. since you and AV earlier suggested keeping it.  I can't remember why I originally deprecated it; maybe the proof seemed too trivial.  But I've sometimes found the E* quantifier non-intuitive and frustrating to work with, since it can reverse implication order, as moimi demonstrates.  Yet sometimes it gives the appearance of not reversing the implication order, as the original morimOLD demonstrates.  Even if it's not used, it might useful for people browsing mo\* theorems to see this behavior.

benjub:
> BTW I'm wondering if it would also be a good idea to restore the original morimOLD (and remove "OLD"), esp. since you and AV earlier suggested keeping it. I can't remember why I originally deprecated it; maybe the proof seemed too trivial. But I've sometimes found the E* quantifier non-intuitive and frustrating to work with, since it can reverse implication order, as moimi demonstrates. Yet sometimes it gives the appearance of not reversing the implication order, as the original morimOLD demonstrates. Even if it's not used, it might useful for people browsing mo* theorems to see this behavior.

Actually, it is that same morimOLD that I "un-deprecated", and I will label it ~moa1.  It used to be the penultimate step of morimOLD, whose last step was a1d, so only weakening the result. I think ~moa1 is the result we want to display. Similarly to ~ala1 and ~exa1, it is rather trivial, but still interesting enough, in my opinion. Letting @avekens know about this.

Indeed, "at least one" preserves implication, "at most one" reverses implication, therefore both (and their conjunction, "exactly one") preserve the biconditional. It would probably be more intuitive to define E* as the conjunction of E. and E! (instead of the current definition of E* as E. -> E!) but this may wait...

**Edit:** namely, morimOLD before my change was:
> morimOLD $p |- ( E* x ( ph -> ps ) -> ( ph -> E* x ps ) ) $=

and now it is
> (currently morim, soon to be moa1) $p |- ( E* x ( ph -> ps ) -> E* x ps ) $=

The previous statement was weaker and tried to restore an analogy of form in the implications, only producing potential confusion (a bit like the "freshman's dream") --- that second ph could be any ch.

PR #1976, opened 2021-03-30
Misc

benjub:


benjub:
What do you think of bj-aev2 and bj-aevdemo ?

Even though bj-aev2 is not important in itself, hbaev and hbaevg are instances of bj-aev2, and I would propose that if a theorem in set.mm is an instance of another theorem, then that other theorem should also be in set.mm.  Also, this gives a placeholder for its comment, which may be of interest to some readers.

bj-aevdemo illustrates this comment.  I think this "finishes off" this part of the study (mainly due to @wlammen), and clearly shows the power of a distinctor antecedent with a DV condition.

nmegill:
Shouldn't aevlemALT and aevALT have an OLD suffix if they are obsolete?

> What do you think of bj-aev2 and bj-aevdemo ?

I like them, and bj-aevdemo provides a nice example.

> 
> Even though bj-aev2 is not important in itself, hbaev and hbaevg are instances of bj-aev2, and I would propose that if a theorem in set.mm is an instance of another theorem, then that other theorem should also be in set.mm. Also, this gives a placeholder for its comment, which may be of interest to some readers.

It's OK with me in this case, but I think we should handle it on a case-by-case basis and not make it a strict requirement.  It should depend on whether the generalized version is useful, insightful, or helps the reader grasp the concept in some way (like any other theorem).  In a borderline case we could mention the existence of the generalized version in the comment.

For example, sometimes it may be more helpful to the reader for a lemma to focus on the instance that is used by the main proof.  Other times, using the more general theorem as the lemma could provide more insight.  But most of the time it wouldn't be useful to have two versions of the lemma.

In the big picture, I think these situations are relatively rare.  In most cases we automatically prefer the more general version as the only theorem.  I believe the conflict in this case is that the more general version didn't fit the pattern for an hb\* theorem, which could be confusing to the reader who expects that pattern.

> 
> bj-aevdemo illustrates this comment. I think this "finishes off" this part of the study (mainly due to @wlammen), and clearly shows the power of a distinctor antecedent with a DV condition.

benjub:
> Shouldn't aevlemALT and aevALT have an OLD suffix if they are obsolete?

Should I relabel them aevlemOLD2 (since there is already an aevlemOLD) and aevOLD2 (since there is already an aevOLD) ?

**Edit:** or use *ALTOLD ?

I will move bj-aev2 and bj-aevdemo to Main. The former uses bj-alrimih, which is "the general statement that alrimih" proves. Should it replace alrimih ? As for the case you mention, this clashes with label conventions. Do you have a label proposal for bj-alrimih and bj-alrimh ? (and keep alrimih with a 1-step proof from it)

nmegill:
> Should I relabel them aevlemOLD2 (since there is already an aevlemOLD) and aevOLD2 (since there is already an aevOLD) ?

No, just add "OLD" to the existing name:  aevlemALTOLD, aevALTOLD.  "OLD2" isn't a suffix that is recognized by "verify markup".  In "...ALTOLD", the suffix (last 3 characters) is "OLD", and the "ALT" part is ignored for that purpose.

The main reason for adding "OLD" though is it might assist a human or a script looking for expired theorems.  @wlammen does a lot of this by hand; I'm not sure if he restricts his search to *OLD, but it can't hurt.

> 
> **Edit:** or use *ALTOLD ?

Yes. :)

nmegill:
I wrote:

> The main reason for adding "OLD" though is it might assist a human or a <i><B>script</B></i> looking for expired theorems.

BTW I'm not encouraging a script for this.  Each deletion must be carefully reviewed by a human anyway, and it's pretty quick to search manually for expired ones matching 1 year + 1 month ago.  So I don't know if a script would even save time.  Plus there is a certain satisfaction of cleaning out old stuff you wouldn't get from a script. :)

benjub:
@nmegill I will move bj-aev2 and bj-aevdemo to Main. The former uses bj-alrimih, which is "the general statement that alrimih proves". Should it replace alrimih ? As in the case you mention, this clashes with label conventions. Do you have a label proposal for bj-alrimih ? And keep its instance alrimih with a 1-step proof from it or remove it ?

**Edit:** I use many setvars: a, b, c... in aevdemo, and these are introduced via $v only later.  Can I move these $v declarations earlier "en masse" ? or something slicker ?
Maybe: have only local declarations for aevdemo, or maybe keep aevdemo in the "examples" section 17.1.5 around the end of Main.
For aev2, I lack only one setvar, so maybe move one or two setvar declarations (vr, vs) together with the first group (vx).

nmegill:
> @nmegill I will move bj-aev2 and bj-aevdemo to Main. The former uses bj-alrimih, which is "the general statement that alrimih proves". Should it replace alrimih ? 

I don't think it should replace alrimih.  It would be confusing and hard to find for someone looking for a theorem that just adds a universal quantifier to the r.h.s. of an implication (which is its normal use).  We should keep the existing one for that reason.  (A minor point is that proofs using bj-alrimih will be slightly longer because of the extra $f hypothesis, but not enough to worry about.)

> As in the case you mention, this clashes with label conventions. Do you have a label proposal for bj-alrimih ? And keep its instance alrimih with a 1-step proof from it or remove it ?

bj-alrimih could be renamed alrimihg maybe.  If we move it up, alrimih could have a 1-step proof from it.

While I'm not always convinced of the usefulness of these more general versions, I think they are relatively rare, so I'm not too worried about it becoming a trend that would bloat set.mm.

> 
> **Edit:** I use many setvars: a, b, c... in aevdemo, and these are introduced via $v only later. Can I move these $v declarations earlier "en masse" ? or something slicker ?

The order in which $f's are introduced is important since it determines the hypothesis order in proofs.  The extra setvars a,b,c... are used only in set theory and also have class variables declared before them, so in order for existing proofs to work you'd have to also move up those class variables to keep the order the same.  Since aevdemo is the only theorem that would use even close to that many variables, I think it is preferable not to do it just to address that one case.

BTW the way you can change the $f order if you really need to is to save all proofs with "save proof */explicit /packed" (which might take an hour or more to run, not sure), then "write source" and move around the $v and $f statements in an editor, then "save proof * /compressed /fast".  This would usually cause a huge diff in set.mm and potentially have conflicts with any other work in progress unless those contributors do the same, and isn't something we want to do casually.

> Maybe: have only local declarations for aevdemo, or maybe keep aevdemo in the "examples" section 17.1.5 around the end of Main.

I think either one of these would be the best thing to do.  I would lean towards putting it in the "examples" section, which will reduce the clutter of local declarations, and it will still be easy to find since I assume you will link to it in the aev2 comment.

> For aev2, I lack only one setvar, so maybe move one or two setvar declarations (vr, vs) together with the first group (vx).

w is available, so you just need one.  Maybe declare it locally for now, since it's the only proof needing it.  Perhaps we should think about moving up some more setvars if another case arises.

PR #1978, opened 2021-03-31
aev2, aevdemo

benjub:
Sequel of #1976 : @nmegill please check that this is ok. In particular:
* I found the label ~sylg for bj-alrimih (similar to ~mpg)
* aevdemo moved to the later part reserved for examples near the end of Main
* one dummy variable 's' declared locally for aev2 and aev2ALT with label v.vs

Relabel proposal:
notnot1 --> notnot
notnot2 --> notnotr
(since there are already labels notnoti, notnotd, notnotri, notnotrd that follow that convention).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14191,8 +14191,10 @@ New usage of "adjvalval" is discouraged (1 uses).
 New usage of "aecom-o" is discouraged (4 uses).
 New usage of "aecoms-o" is discouraged (6 uses).
 New usage of "aev-o" is discouraged (1 uses).
+New usage of "aev2ALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> Sequel of #1976 : @nmegill please check that this is ok. In particular:
> 
> * I found the label ~sylg for bj-alrimih (similar to ~mpg)
> * aevdemo moved to the later part reserved for examples near the end of Main
> * one dummy variable 's' declared locally for aev2 and aev2ALT with label v.vs

This is fine.

> 
> Relabel proposal:
> notnot1 --> notnot
> notnot2 --> notnotr
> (since there are already labels notnoti, notnotd, notnotri, notnotrd that follow that convention).

I agree, good idea.

PR #1979, opened 2021-04-01
New approach for graph theory (continued) - Friendship graphs

avekens:
Main set.mm:
* ~a1i13 moved from JGH's mathbox to main set.mm
* ~bj-animorl, ~bj-animorr, ~bj-animorlr, ~bj-animorrl renamed and moved from BJ's mathbox to main set.mm (@benjub: en bloc, although only (bj-)animorr is used in AV's Mathbox)

* section "Friendship graphs - basics" added, including the revised definition of friedship graphs and related theorems
* section "The friendship theorem for small graphs" added, using the revised definition of friedship graphs

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -265,15 +265,15 @@
 "4syl" is used by "acndom".
 "4syl" is used by "acunirnmpt2".
 "4syl" is used by "acunirnmpt2f".
-"4syl" is used by "aevlem1".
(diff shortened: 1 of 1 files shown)


benjub:
Just noticed:
 similiar --> similar
if you can fix it. Thanks !

PR #1980, opened 2021-04-01
notnot-->notnotb; notnot2-->notnotr; notnot1-->notnot

benjub:
As agreed in #1978

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -5435,7 +5435,7 @@
 "e11" is used by "elex2VD".
 "e11" is used by "eqsbc3rVD".
 "e11" is used by "hbimpgVD".
-"e11" is used by "notnot2ALTVD".
+"e11" is used by "notnotrALTVD".
 "e11" is used by "onfrALTVD".
 "e11" is used by "orbi1rVD".
 "e11" is used by "ordelordALTVD".
@@ -5530,7 +5530,7 @@
 "e1a" is used by "hbalgVD".
 "e1a" is used by "hbexgVD".
 "e1a" is used by "hbimpgVD".
-"e1a" is used by "notnot2ALTVD".
+"e1a" is used by "notnotrALTVD".
 "e1a" is used by "onfrALTVD".
 "e1a" is used by "onfrALTlem2VD".
 "e1a" is used by "onfrALTlem3VD".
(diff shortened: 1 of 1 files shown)


nmegill:
I got an email from GitHub a few minutes after I merged (IIRC, all checks had passed before I merged).  I'm not sure who else it was sent to, but basically it says:

Subject: [metamath/set.mm] Run failed: verifiers - develop (00f6924)

[metamath/set.mm] verifiers workflow run
verifiers: Some jobs were not successful
        Check for duplicate jobs to avoid duplication
verifiers / Check for duplicate jobs to avoid duplication
Failed in 0 seconds
        annotations for verifiers / Check for duplicate jobs to avoid duplication 1
        Verify using checkmm (a C++ verifier by Eric Schmidt)
verifiers / Verify using checkmm (a C++ verifier by Eric Schmidt)
Skipped
[...rest omitted...]

Does anyone understand what's going on?

david-a-wheeler:
The original commit 69d76aa successfully verified.

It's the merge into our default branch (develop) that failed as commit 00f6924. *Why* it failed is unclear. I'll take a look.

david-a-wheeler:
You can see more details by clicking on the little fail symbol, which should lead you here: https://github.com/metamath/set.mm/runs/2249905376

It failed on the first step, "Check for duplicate jobs to avoid duplication". That should *never* fail, that's just a preliminary step to eliminate duplicate jobs. 

The code that's run is here: https://github.com/metamath/set.mm/blob/develop/.github/workflows/verifiers.yml

Here's the relevant snippet:

~~~~
  skip_dups:
    name: Check for duplicate jobs to avoid duplication
    # continue-on-error: true # Uncomment once integration is finished
    runs-on: ubuntu-latest
    # Map a step output to a job output
    outputs:
      should_skip: ${{ steps.skip_check.outputs.should_skip }}
    steps:
      - id: skip_check
        uses: fkirc/skip-duplicate-actions@master
        with:
          github_token: ${{ github.token }}
          paths_ignore: '["**/README.md", "**/docs/**"]'
~~~~

I think the two most likely causes are:

1. A temporary failure in GitHub's build system. Real systems fail, sadly.
   I think that is the *most* likely reason. The failure happened 2 hours ago, and
   the ["GitHub Status" Twitter account](https://twitter.com/githubstatus?lang=en)
  reported at that time "GitHub Actions is experiencing degraded performance.
  We are still investigating and will provide an update when we have one."
2. If it's a *permanent* failure, the most likely cause is the
  "skip-duplicate-actions" action is failing (e.g., it's removed and/or
  it uses a GitHub API that is no longer supported). I think this is unlikely.

So I think you just need to [forcibly re-run the workflow](https://docs.github.com/en/actions/managing-workflow-runs/re-running-a-workflow) (click on the link to see the instructions). Let me know if that doesn't solve it!

PR #1982, opened 2021-04-04
Misc (only my mathbox)

benjub:
@nmegill : note the proof that bj-minimp is a single axiom for minimal implicational calculus. Maybe it can be moved to Main; as you see fit.


nmegill:
> @nmegill : note the proof that bj-minimp is a single axiom for minimal implicational calculus. Maybe it can be moved to Main; as you see fit.

I'll look at it later today.

nmegill:
> @nmegill : note the proof that bj-minimp is a single axiom for minimal implicational calculus. Maybe it can be moved to Main; as you see fit.

OK.  It can go in the section "Other axiomatizations of classical propositional calculus".  We should change the title to "Other axiomatizations related to classical propositional calculus" (similar to the existing "Other axiomatizations related to classical predicate calculus"), or maybe you have another suggestion.
            ​

PR #1983, opened 2021-04-05
New approach for graph theory (continued) - Friendship graphs (2)

avekens:
Main set.mm:
* ~adant423 moved from GS's mathbox to main set.mm
* ~bnj833 renamed and moved from JB's mathbox to main set.mm
* ~1fpid3 added

* ~xnn0n0n1ge2b added
* ~lfuhgr1v0e, ~vtxdlfuhgr1v, ~vdumgr0, ~fusgrn0degnn0 added
* ~1wlkvtxeledglem, ~wlkOnl1iedg, ~wlkOn2n0 added
* ~conngrv2edg, ~vdn0conngrumgrv2 added
* section "Theorems according to Mertzios and Unger" added, using the revised definition of friedship graphs

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2601,10 +2601,6 @@
 "bnj832" is used by "bnj1523".
 "bnj832" is used by "bnj605".
 "bnj832" is used by "bnj908".
-"bnj833" is used by "bnj1145".
-"bnj833" is used by "bnj1398".
-"bnj833" is used by "bnj1442".
-"bnj833" is used by "bnj570".
 "bnj835" is used by "bnj1175".
 "bnj835" is used by "bnj1219".
 "bnj835" is used by "bnj1280".
@@ -14903,7 +14899,6 @@ New usage of "bnj769" is discouraged (7 uses).
 New usage of "bnj770" is discouraged (9 uses).
 New usage of "bnj771" is discouraged (7 uses).
 New usage of "bnj832" is discouraged (9 uses).
-New usage of "bnj833" is discouraged (4 uses).
 New usage of "bnj835" is discouraged (17 uses).
 New usage of "bnj836" is discouraged (6 uses).
 New usage of "bnj837" is discouraged (5 uses).


benjub:
Would simplbiimp be more useful if its first hypothesis is replaced by an implication ? (so with a label more like simplim ?)
And can you rename variables alphabetically ? Like
ph -> ( ps /\ ch )
ch -> th
=> ph -> th

avekens:
@benjub I do not really understand how you want the first hypothesis to be replaced by an implication. In the current form, the theorem is used 4 times in JB's mathbox, and once in my mathbox (shortening a proof). I do not know if changing the theorem will still result in shorter proofs.
Concerning the label, however, you are right: "im" is used more often for "implication". Therefore, I will change the label to "simplbiim" in my next PR.

benjub:
> @benjub I do not really understand how you want the first hypothesis to be replaced by an implication.

I wrote the statement above.

> In the current form, the theorem is used 4 times in JB's mathbox, and once in my mathbox (shortening a proof). I do not know if changing the theorem will still result in shorter proofs.

Me neither. I leave it up to you to decide which form is best. If you have the courage, you can use "scripts/min.cmd" to see how many theorems it could minimize.

avekens:
OK, I see.
I proved the implication version of the theorem, and tried to minimize the concerned theorems (with the script min.cmd) using the new theorem, but no shorter proofs were found. Looking at ~2pthfrgr, the concerned part of the proof is:

18 eldifsn       $p |- ( b e. ( V \ { a } ) <-> ( b e. V /\ b =/= a ) )
19 necom         $p |- ( b =/= a <-> a =/= b )
20 19 biimpi     $p |- ( b =/= a -> a =/= b )
21 18,20 simplbiimp  $p |- ( b e. ( V \ { a } ) -> a =/= b )

Replacing line 21 with the new version of the theorem results in a longer proof (1 additional step), and running minimize reverts the change, using simplbiimp again. Therefore, I think the current version has its reason for being.

About changing the variable names: they cannot be changed without changing all the concerned proofs. Therefore, I will not perform the change.

benjub:
Thanks @avekens .  If I have time, I'll do the renaming of variables. To do it, first create xxxNEW with the right order of variables (prove it from xxx and make MM-PA>expand xxx), then for all theorems using xxx, do MM-PA>min xxxNEW/may_grow (e.g. use MM>show usage xxx, and make a script), then make a global rename xxxNEW --> xxx in the text editor and remove the older xxx.  Maybe @nmegill has a better method or a script already made?

nmegill:
> Maybe @nmegill has a better method or a script already made?

What you described is pretty much what I do.  I have a simple script that creates a "minimize" script from a "show usage" output that  I think I posted on the Google group a few years ago.   But sometimes I find it easier to just to write a new one than to recall the filenames, usage, limitations, and quirks of some previous quick-and-dirty script I wrote.

PR #1988, opened 2021-04-09
hbnae with dv condition

wlammen:


benjub:
I would move this to Main, with a comment like "Version of ~ hbnae with a DV condition, requiring fewer axioms." and add to the comment of hbnae "See ~ hbnaev for a version with a DV condition, requiring fewer axioms."

wlammen:
I am still struggling with removing dependencies on ax-12 as far as equality is concerned, in particular by using ax-13 instead. This would be much easier if instead of ax13v we had dveeq2 as the root ax-13 instance in place. In this light I consider wl-spae more powerful (for example it allows deriving ax13v from dveeq2 in a simple fashion). Until things sort out I think I keep it in my mathbox for frequent and unbureaucratic modification.
A pleasant intermediate goal would be ax6e proved from FOL and ax-13v alone.

And as always, my time is limited to some spare time, which makes fast progress impossible.

benjub:
> This would be much easier if instead of ax13v we had dveeq2 as the root ax-13 instance in place.

I've been thinking about it, and if we go in that direction, then I think that axc9 would be the best axiom. One recovers trivially ax-13 from it and sp.

axc9, also due to @nmegill, is easier to understand, but I think that Norm puts preference on shorter axioms.

wlammen:
The key observation about ax13v is (in my eyes) that it resembles an instance of ax-5, except that one dv condition d(x, y) is replaced with an inequality between x and y.  So inequality serves as a distinctor available to logical treatment.  I am not sure whether the rest y=z->A.x y=z is just a placeholder for something that needs x and y be distinct.  And as such in principle be replaceable with something else having the same property.
The problem with inequality as a distinctor is that it is difficult to handle in case studies.  -. Ax x = y is a better choice here.
So if you want to emphasize the distinctor role, and nothing more, I'd rather vote for dveeq2 (with an extra dv condition on y, z).

Case studies are (x, t distinct): -. x = y and x = y /\ -. A. x = t and A. x x = t 
vs -. Ax x = y and A. x x= y
For the latter I can come up with some solutions out of the box.

benjub:
> I am not sure whether the rest y=z->A.x y=z is just a placeholder for something that needs x and y be distinct. And whether it is replaceable with something else having the same property.

Interesting remark. I don't have an answer, but (y=z -> A.x y=z ) is arguably among the simplest formulas having this property.

> The problem with inequality as a distinctor is that it is difficult to handle in case studies. -. Ax x = y is a better choice here.

Completely agree. That's why I even prefer axc9.

> So if you want to emphasize the distinctor role, and nothing more, I'd rather vote for dveeq2 (with an extra dv condition on y, z).

Then go all the way and use axc9 !  A property (which can be considered a drawback) of dveeq2 is that if you remove the DV conditions, then it becomes false, contrary to ax-13, ax13v and ax-c9 (since it has no DV conditions, "removing DV conditions" does nothing). I think, and @nmegill would probably agree, that having a system where only one axiom has this property (namely ax-5) is preferable. By the way, there is no axiomatization where no axiom has this property (proof: such an axiomatization would prove ax-5, and therefore it would prove ax-5 without DV condition (simply remove from the proof all DV conditions of the form DV(var,wff)) which is false).

wlammen:
OK, I see why one should prefer axc9 as an axiom and derive dveeq2 as the root instance (d(y, z) included).  I don't know whether I am already done with -. x = y and x = y /\ -. A. x = t and A. x x = t. My problems with handling them stand for nothing. In fact, ax-13 is sufficient, with the aid of other axioms. It looks messy only, if you want to switch easily between ax-6 to ax-13 and ax6v to ax13v. The axiom usage display is annoying then.

PR #1989, opened 2021-04-10
dv version of equveli

wlammen:
Also align appearance of equvin and equviniv, so the former proof becomes shorter.  This has no impact on other theorems, as equvin has no reference so far.


benjub:
Good change on equvin. Maybe it could actually be changed to 
  E. z ( z = x /\ z = y )
simlarly to aev and others where the quantified variable is on the LHS of the equality. Same with ~equvini.

I like your formulation of wl-equveliv (with the above change). Maybe you can do the same change with ~equveli ?

I'm in favor of moving wl-hbnaev (with the comment I proposed in #1988) and wl-equveliv to Main.

wlammen:
There is more headroom for changes like this. If you prefer A. x x = y over A. x y = x (which I do), you might carry the same idea even further. Say that we have some sub-expressions T1 and T2, and T1 is more dependent on x than T2, then we prefer A.x ( T1 op T2 ) over A.x ( T2 op T1 ) with some operation op. This would give set.mm a more uniform look, and eases searching for expressions.
Unfortunately, this looks like a lot of work.

benjub:
> There is more headroom for changes like this. If you prefer A. x x = y over A. x y = x (which I do), you might carry the same idea even further. Say that we have some sub-expressions T1 and T2, and T1 is more dependent on x than T2, then we prefer A.x ( T1 op T2 ) over A.x ( T2 op T1 ) with some operation op. This would give set.mm a more uniform look, and eases searching for expressions.

I agree. Although "more dependent" may not always be that clear-cut.  Another "mild convention" is
> "shorter term" OP "longer term"

or sometimes the opposite, depending on context. And sometimes these conventions would clash anyway.

But we can probably agree to prefer expressions like A. x ( x = y /\ x = z ) or E. x ( x = y \\/ x = z ) over the "reversed" versions. Giving:

  equvin $p |- ( x = y <-> E. z ( z = x /\ z = y ) ) $=
(equivini: idem)

  wl-equveliv $p |- ( x = y <-> A. z ( z = x -> z = y ) ) $=
(equiveli: idem)

And maybe relabel:
equvin --> equvinv
equvini --> equvin
equveli --> equvel
wl-equveliv --> equvelv (and move to Main)

nmegill:
I think I also would prefer A. x x = y over A. x y = x, and I think that's roughly been the informal convention.  It also matches the bound-variable-first order required for e. in extensionality and "A. x ( x e. A ->..." type theorems.

As for x = z /\\ z = y in equvini, that is the order used in Monk's Lemma 15, but I've had mixed feelings about it  (IIRC it has sometimes been awkward in proofs, requiring an extra equcom) and would be happy with  z= x /\\ z = y.  Also, note that wl-equveliv is the same as sbeqalbi (in ATS's mathbox) with the z reversed in the equalities.  The latter uses ax-10,12,13, though, so wl-equveliv is of course preferable.

nmegill:
> But we can probably agree to prefer expressions like A. x ( x = y /\ x = z ) or E. x ( x = y \/ x = z ) over the "reversed" versions.

I agree.

> And maybe relabel:
> equvin --> equvinv
> equvini --> equvin
> equveli --> equvel
> wl-equveliv --> equvelv (and move to Main)

These are ok with me.

PR #1991, opened 2021-04-11
Implement requests from PR 1989

wlammen:

PR #1992, opened 2021-04-11
new section 'minimal implicational calculus'

benjub:
As agreed in @1982

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16101,6 +16101,7 @@ New usage of "equs5aALT" is discouraged (0 uses).
 New usage of "equs5eALT" is discouraged (0 uses).
 New usage of "equsb3ALT" is discouraged (0 uses).
 New usage of "equsexALT" is discouraged (0 uses).
+New usage of "equtr2OLD" is discouraged (0 uses).
 New usage of "equvinivOLD" is discouraged (0 uses).
 New usage of "equvinvOLD" is discouraged (0 uses).
 New usage of "erngbase-rN" is discouraged (4 uses).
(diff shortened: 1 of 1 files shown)


benjub:
Second commit of this PR: deprecate equviniv as proposed by @wlammen in #1991.  Also, remove ax8OLD and ax9OLD since I can confirm that the older proofs do not give any additional insight.

benjub:
Third commit of this PR: add equeucl : it is only used three times, but several statements about transitivity, symmetry, reflexivity are already in set.mm, so Euclideanness (Euclideanity?) should be too, and its uncurried form is also in set.mm, as equtr2.

benjub:
Just noticed that the proof of equeucl could be shortened to
>  equeucl $p |- ( x = z -> ( y = z -> x = y ) ) $= ( weq equequ2 biimprcd ) BCDABDACDBCAEF $.

But it feels backwards to prove the implication from the biconditional. I will probably move equeucl up in order to prove equequ2 from it.  Similarly to equtr / equtrr , I might add equeuclr $p |- ( x = z -> ( y = z -> y = x ) ) . Any thoughts ?

nmegill:
> But it feels backwards to prove the implication from the biconditional. I will probably move equeucl up in order to prove equequ2 from it. 

I'm ok with that.

> Similarly to equtr / equtrr , I might add equeuclr $p |- ( x = z -> ( y = z -> y = x ) ) . Any thoughts ?

equeuclr would save at most a single step (com12) over equeucl, and my initial reaction is that it feels like too trivial a variation.  I don't see that it would "pay for itself" by reducing the set.mm file size (if it does, feel free to add it of course).  I guess we could say the same about equtrr vs. equtr, but at least equtrr is used 13 times and also has a literature reference.  Do you know how often equeuclr would be used?

benjub:
equeuclr would shorten: aevlem0 (-11B), ax13b (-14B), equvini (-15B), sbequi (-15B).

Also, if equeucl is moved before equequ2 in order to prove it, then equequ2 will use equeucl followed by com12, but the penultimate step of equeucl is equeuclr and the ultimate step uses com12. So, if expanded, the proof of equequ2 would have two com12 in a row. Therefore, adding equeuclr is not going out of one's way to add yet another variant, but rather using something that's already a subproof.

It's true that equtr, equtrr, ax7, equeucl, equeuclr are closely related variants. I think that for such short statements which express the most basic properties of equality, it is ok (as opposed to longer statements, where it would be useless to have closely related variants differing only by the order of x and y in "x=y").

nmegill:
Well, go ahead and add equeuclr if you want.  I just want to be cautious about cluttering set.mm with trivial variants of existing theorems without justification.

PR #1995, opened 2021-04-14
Add amgmw2d

DyeKuu:
- Adds Weighted AM-GM for n = 2.


nmegill:
Thanks for your contribution.

It did not pass all checks - could you fix and resubmit?  (It looks like just a trailing space on line number 616807.  We require trailing spaces to be removed because they can cause confusing diffs.)

DyeKuu:
> Thanks for your contribution.
> 
> It did not pass all checks - could you fix and resubmit? (It looks like just a trailing space on line number 616807. We require trailing spaces to be removed because they can cause confusing diffs.)

Thank you for your hint! It is fixed now.

nmegill:
Thanks for the correction.  Your proof can be seen at http://us2.metamath.org/mpeuni/amgmw2d.html.  A couple of suggestions for your next PR:

1. Our practice is to include an English description with each theorem.  I would suggest something like "Weighted arithmetic-geometric mean inequality for \` n = 2 \` (compare ~ amgm2d )."

2. The title of the link above is "Mathbox for Stanislas Polu" because your theorem is in his mathbox.  I would suggest you add your own mathbox.  Since your name is alphabetically last, you would add the following section header at the end of set.mm:
```

$(
                Mathbox for Kunhao Zheng
$)
```
and add your theorems below it.

david-a-wheeler:
@DyeKuu : Welcome to Metamath! I look forward to whatever you do next!

benjub:
Actually, the metamath symbol RR+ means "(strictly) positive reals", and not "nonnegative reals". Does your (@DyeKuu) proof work for nonnegative reals ? At worst, you finish off the proof with a case analysis: "if at least one is zero, then..., else..." Not very important but the statement would feel a bit more natural. (I see that the same applies to @spolu's AMGM proofs.)

And now that we have the unweighted AMGM inequality ~amgm and the weighted AMGM inequality for two numbers, the next goal is obvious: weighted AMGM inequality for any (nonempty) finite set of nonnegative real numbers. Any volunteers ?

**Edit:** I see from @digama0's proof of ~amgm that actually, the (strictly) positive case is proved by ~amgmlem, and ~amgm itself takes care of the case analysis. So maybe that's a bit too much hassle to worry about extending the other proofs from the (strictly) positive case to the nonnegative case.

PR #2003, opened 2021-04-24
Shorten pm3.2an3

KyleWyonch:
Hope I did this right, first commit here, found a proof for pm3.2an3 which reduces size from 76 to 52 bytes.  Has same axiom usage as old proof.


nmegill:
It looks good, but there were 2 small problems with the checks:
1. The date should be "24-Apr-2021" instead of "24-Ap-2021".
2. The "discouraged" file changed and needs to be regenerated; see "Regenerating the discouraged file" at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md

Could you fix these and resubmit?

PR #2004, opened 2021-04-25
Shorten pm3.2an3 Fixed

KyleWyonch:
This one should work, fixed the date and regenerated the discouraged file.


nmegill:
There is still a small problem I overlooked:  line number 21278 ends with a space character.  (We require trailing spaces to be removed because they have caused confusing diffs in the past.)

Note that if you run the "Formatting recommendation prior to submitting a pull request" at  https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md it will identify or fix these problems in advance.

PR #2009, opened 2021-04-27
mathbox: elementwise intersection

benjub:
@nmegill : I propose to replace zfregcl and zfreg by their generalizations where the hypothesis A e. _V is replaced by the antecedent A e. V (and there is no DV condition on V). The new versions are in this PR: bj-zfregcl , bj-zfreg.  I actually need the latter to prove bj-restreg.  The other usages can be adapted, and this even provides (in the cases I looked at) shortenings of the proofs using these. I also propose to deprecate zfreg2 as it is a minor variation of zfreg and has only one usage.

This PR was motivated by the google post https://groups.google.com/g/metamath/c/lkOfogl6cPg/m/o3a1mDifAgAJ by @glacode. 
I propose to have a section dedicated to df-rest earlier in set theory, but the version in my mathbox is not ready (there are many theorems, mostly by @digama0, already in set.mm, that have to be moved, and I have to see if there are some redundancies). I think I will open an issue when this is a bit more polished.

Note to @nmegill  : in the head comment of the section, I had written \` X |\`t A \` and this caused a warning on verify markup, so I replaced the rest symbol with dots. This is probably due to a parsing problem with backticks somewhere.

Below are the warnings.

?Warning: In the comment for statement "bj-rest00", math symbol token "the"
does not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token
"elementwise" does not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token
"intersection" does not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token "on" does
not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token "the"
does not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token "family"
does not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token "by" does
not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token "the"
does not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token "REMARK:"
does not have a LaTeX and/or an HTML definition.
?Warning: In the comment for statement "bj-rest00", math symbol token "many"
does not have a LaTeX and/or an HTML definition.
[...]

PR #2010, opened 2021-04-28
typos

benjub:
Fix two typos from #2009.


benjub:
Generalize zfregcl, zfreg as agreed in #2009. It turns out that it slightly lengthened the proofs using them, but I think this is minor (and already noted in previous discussions about sethood hypothesis versus sethood antecedent with @nmegill and @avekens).

PR #2014, opened 2021-05-01
New theorem iseri; new approach for graph theory (continued)

avekens:
Main set.mm:
* ~iseri and ~iseriALT added, proofs of ~eqer, ~0er, ~ecopover, ~ener, ~gicer, ~phtpcer, ~vitalilem1, ~erclwwlk, ~erclwwlkn shortened
* ~elqsecl added

AV's Mathbox:
* ~clwwlksnbprop, ~clwwlksnwrd added
* proof of ~erclwwlks shortened
* revision of additional theorems for closed walks as words

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13995,6 +13995,7 @@ New usage of "0blo" is discouraged (1 uses).
 New usage of "0cnALT" is discouraged (0 uses).
 New usage of "0cnfn" is discouraged (4 uses).
 New usage of "0cnop" is discouraged (1 uses).
+New usage of "0erOLD" is discouraged (0 uses).
 New usage of "0heALT" is discouraged (0 uses).
 New usage of "0hmop" is discouraged (9 uses).
 New usage of "0idsr" is discouraged (8 uses).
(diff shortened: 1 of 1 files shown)


avekens:
There seems to be something wrong with http://us.metamath.org/mpegif/mmbiblio.html - Benoît's PR failed also because of this reason.

avekens:
http://us.metamath.org/ is not accessible!

benjub:
The error was triggered by 
> wget -N -q http://us.metamath.org/downloads/checkmm.cpp

and
> wget -N -q http://us.metamath.org/mpegif/mmbiblio.html

I have two questions (not related to us.metamath.org being down, but seeing these lines just made me wonder):
* maybe we should replace all "mpegif" with "mpeuni" (probably no big difference, but now mpeuni is the default) ?
* since the PRs are done to us2, maybe one should replace all "us" with "us2" ?

Maybe our continuous-integration wizard @david-a-wheeler can chime in !

nmegill:
It looks like every mirror is down. 

us.metamath.org was temporarily pointed to at.metamath.org, when the real one was shut down due to a false virus positive.  But now at.metamath.org is down.

So I pointed us.metamath.org to the computer at my house for now.  It may take a few hours for the DNS to propagate.

avekens:
OK, the checks are working again now...

david-a-wheeler:
This is not the first time a problem with metamath.org has caused check failures.

Maybe we should set things up so that the checks will download everything from GitHub?

PR #2016, opened 2021-05-03
Misc

benjub:
@nmegill : I added the two utility theorems vpwex and vuniex (to minimize respectively 28 and 50 proofs). Are the labels good to you ?

@nmegill : I noticed that "write source set.mm/rewrap" rewrapped two lines below, namely:
```
    $( The sequence of vertices of a 1-walk is a word over the set of vertices.
       (Contributed by AV, 27-Jan-2021.) $)
    $( The sequence of vertices of a 1-walk is a word over the set of
       vertices.  (Contributed by AV, 27-Jan-2021.) $)

     its edges plus 1.  (Contributed by Alexander van der Vekens, 29-Jun-2018.)
     (Revised by AV, 1-May-2021.) $)
     its edges plus 1.  (Contributed by Alexander van der Vekens,
     29-Jun-2018.)  (Revised by AV, 1-May-2021.) $)
```
even though their length was 79.  Both lines end with a period (or a period followed by a closing parenthesis).  Maybe this is connected to the fact that a double space is automatically added after a period, so the program thinks that the length of the line is 79+1, so rewraps it ?


nmegill:
> @nmegill : I added the two utility theorems vpwex and vuniex (to minimize respectively 28 and 50 proofs). Are the labels good to you ?

I assume that these result in a smaller set.mm file size overall.

As for the labels, a "v" after the label would have been nice, but of course we reserve that for dv versions.  So maybe a "v"  before is OK.  Typically I would search for pwex\* to find existence theorems, but since the "v" version would automatically be brought in by 'minimize', I guess it isn't important to make it easier to find.

Another case in set.mm is ssnid added in 2018, which is the setvar version of snid.  That name is confusing (it makes me think of "subset", although I think DAW meant "setvar") so the name should probably be changed to vsnid for consistency.

Yet another case is elsn vs. elsnc (very old theorems; the "c" version came later and was meant to stand for "class version").  We could change these to velsn and elsn resp. (and take out the "c" from some other elsnc\* labels).

> 
> @nmegill : I noticed that "write source set.mm/rewrap" rewrapped two lines below, namely:
> ...
> even though their length was 79.

I'm not sure what is going on.  I'll investigate it.

benjub:
> I assume that these result in a smaller set.mm file size overall.

Yes, minimizations gained 3 to 8 bytes, say 4 on average, so the respective gains are around 120 bytes for vpwex and 200 bytes for vuniex, which is more than the size needed to write them (for vpwex, maybe not counting the main part of the comment).

> As for the labels, a "v" after the label would have been nice, but of course we reserve that for dv versions. So maybe a "v" before is OK. Typically I would search for pwex* to find existence theorems, but since the "v" version would automatically be brought in by 'minimize', I guess it isn't important to make it easier to find.
> 
> Another case in set.mm is ssnid added in 2018, which is the setvar version of snid. That name is confusing (it makes me think of "subset", although I think DAW meant "setvar") so the name should probably be changed to vsnid for consistency.

Yes, I thought that \*v is reserved for DV versions and we have ~vex, so I used v\*.  I will change ~ssnid to ~vsnid in my next PR for consistency.

> Yet another case is elsn vs. elsnc (very old theorems; the "c" version came later and was meant to stand for "class version"). We could change these to velsn and elsn resp. (and take out the "c" from some other elsnc* labels).

I will do.  Another argument for that change is that ~ elsni is already of the given form (it is not "elsnci").

To sum up:
* ssnid --> vsnid
* elsn --> velsn
* elsncg, elsnc, elsnc2g, elsnc2 --> remove the c

PR #2017, opened 2021-05-03
six relabelings

benjub:
As agreed in #2016


nmegill:
@benjub I noticed that velsn has $d x A which is obviously unnecessary (with a different proof e.g. as an instance of elsn).  Funny this hasn't been noticed after so many years.  I wonder if removing the $d would make it shorten more proofs, if you feel like investigating it.

benjub:
> @benjub I noticed that velsn has $d x A which is obviously unnecessary (with a different proof e.g. as an instance of elsn). Funny this hasn't been noticed after so many years. I wonder if removing the $d would make it shorten more proofs, if you feel like investigating it.

Indeed, I have edited the comment of velsn to point to this fact, with a link to elsn.  I think I will simply move elsng and elsn before velsn, to prove the latter from elsn.  Even if no shortening ensues, this will be clearer. Should I keep the current velsn proof as ALT ? or OLD ?

nmegill:
> Should I keep the current velsn proof as ALT ? or OLD ?

Not ALT, it is of no interest.  If you want to keep it as OLD that is ok.

PR #2018, opened 2021-05-04
remove DV condition from velsn and minimize *elsn* with *elsn*

benjub:
As agreed in #2017:
* move elsng, elsn a bit earlier
* prove velsn from elsn (and remove DV condition)
* minimize proofs using \*elsn\* with \*elsn\*, excluding some extralong proofs in @glacode's mathbox (see log below)

I did these changes "silently", that is, no credits and no OLD version, since this seemed to be ok with @nmegill and the changes are minor.

```
Proof of "fperdvper" decreased from 5881 to 5866 bytes using "nelsn".
Proof of "fvunsn" decreased from 342 to 326 bytes using "nelsn".
Proof of "inelpisys" decreased from 515 to 500 bytes using "elsni".
Proof of "lbsextlem4" decreased from 3270 to 3248 bytes using "nelsn".
Proof of "obslbs" decreased from 1256 to 1236 bytes using "nelsn".
Proof of "qqhval2" decreased from 2231 to 2211 bytes using "nelsn".
Proof of "sge00" decreased from 1080 to 1075 bytes using "elsni".
Proof of "sge0fodjrnlem" decreased from 2769 to 2742 bytes using "elsn".
Proof of "sitgaddlemb" decreased from 1621 to 1610 bytes using "velsn".
Proof of "sniota" decreased from 184 to 177 bytes using "velsn".
Proof of "submateqlem1" decreased from 737 to 720 bytes using "nelsn".
Proof of "submateqlem2" decreased from 423 to 407 bytes using "nelsn".
Proof of "ttukeylem7" decreased from 2247 to 2237 bytes using "velsn".
Proof of "unirnmapsn" decreased from 1192 to 1158 bytes using "elsn".
Proof of "vdgr1b" decreased from 892 to 882 bytes using "nelsn".
Proof of "wallispi" decreased from 3040 to 3020 bytes using "nelsn".
```


benjub:
I was about to edit some comments and wondered:
> How do you call the expression { A } ?

Do you say "a singleton" or "a potential singleton" ? Indeed, in set.mm, it is either a singleton or the empty set (and this may be undecidable in iset.mm, so ping @jkingdon).

And if you specify the element, do you say "the potential singleton on A" ? or "the pontential singleton (potentially) containing A" ?

nmegill:
> I was about to edit some comments and wondered:
> 
> > How do you call the expression { A } ?
> 
> Do you say "a singleton" or "a potential singleton" ? Indeed, in set.mm, it is either a singleton or the empty set (and this may be undecidable in iset.mm, so ping @jkingdon).
> 
> And if you specify the element, do you say "the potential singleton on A" ? or "the pontential singleton (potentially) containing A" ?

I think "potential singleton" is much too pedantic.  We don't say + is "potential addition" even though it evaluates to the empty set for proper class arguments.  Informally, { . } is a "function" of an argument that is meant to be a set, and a proper class is outside of its "domain".  I think we make it clear in df-sn that a proper class argument is not meaningful.

jkingdon:
> Do you say "a singleton" or "a potential singleton" ?

I would agree with Norm, that "potential singleton", at least in the vast majority of contexts, would be confusing rather than enlightening.

As for iset.mm, it looks like the naming and a few other details are a bit behind set.mm, but as far as I can tell from http://us.metamath.org/ileuni/elsnc.html and similar theorems, this is not one of those cases where set.mm uses excluded middle based on "A is a proper class or it is not" (for that, see http://us.metamath.org/mpeuni/snex.html vs http://us.metamath.org/ileuni/snexg.html ).

PR #2020, opened 2021-05-05

avekens:
Main set.mm:
* ~hasheqf1oi, ~hashf1rn revised
* ~focdmex, ~hasheqf1od  added
* ~several proofs shortened

BJ's mathbox:
* proof of ~bj-finsumval0 shortened

AV's mathbox:
* several proofs shortened
* new theorems ~1wlklenvclwlk, ~upgrclwlkcompim
* remaining theorems for closed walks revised

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -6757,6 +6757,7 @@
 "h2hvs" is used by "h2hmetdval".
 "h2hvs" is used by "hhvs".
 "halfnq" is used by "nsmallnq".
+"hasheqf1oiOLD" is used by "hashf1rnOLD".
 "hatomic" is used by "atomli".
 "hatomici" is used by "atcvat4i".
 "hatomici" is used by "atcvatlem".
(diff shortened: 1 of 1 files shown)


benjub:
@avekens I just noticed on http://us2.metamath.org/mpeuni/usgsizedg.html what looks like extra spaces. It looks like GrSize and Edges were defined with an extra space before them, that is, as " GrSize " instead of "GrSize ". As a rule of thumb, a function of one argument should have a trailing space, but no beginning space, and a function of two arguments (which is generally used infixed) should have both.

avekens:
> @avekens I just noticed on http://us2.metamath.org/mpeuni/usgsizedg.html what looks like extra spaces. It looks like GrSize and Edges were defined with an extra space before them, that is, as " GrSize " instead of "GrSize ". As a rule of thumb, a function of one argument should have a trailing space, but no beginning space, and a function of two arguments (which is generally used infixed) should have both.

We have the following "Typography conventions" in set.mm:

- Class symbols for functions (e.g., ` abs ` , ` sin ` ) should usually not have leading or trailing blanks in their HTML/Latex representation.
- This is in contrast to class symbols for operations (e.g., ` gcd ` , ` sadd ` , ` eval ` ), which usually do include leading and trailing blanks in their representation.

According to this convention, "Edges" should be defined as 'Edges' , and "GrSize" as 'GrSize'. I'll change it with my next PR, although these definitions will be removed soon (after I finished the revision of graph theory).

PR #2021, opened 2021-05-05
mathbox: wl-speqv2

wlammen:
Remove some outdated proofs as well.


benjub:
@wlammen wl-speqv2 looks like an instance of spvw ? If it has special interest, put the "Proof modif. discouraged" tag to avoid automatic minimization.

wlammen:
@benjub or 19.9v. I know that. I am currently struggling with converting wl-spae into to x=y -> E.x x=y, which has the potential to remove ax-12 from ax6e.
So wl-speqv2 is just a reminder about a particular technique. I currently have little time to look into this regularly, and I tend to forget results I found a week ago or so.  This theorem is not designed to enter main and will be finally deleted.

benjub:
Then I advise "Proof modif. discouraged" tag to avoid automatic minimization (by any other contributor). Good luck for your next goals.

PR #2022, opened 2021-05-08

avekens:
Main set.mm:
* ~ee02 (renamed ~mpsylsyld) moved from AS's mathbox
*  comments for ~df-f, ~df-f1, ~df-fo, ~df-f1o enhanced, especially introduction of  the terms injection, surjection and bijection, and reference to category theory.

AV's mathbox:
* blanks in html representation of `GrSize` and `Edges` removed
* theorems ~frusgrnn0, ~rusgrnumwrdl2, ~uspgr2wlkeqi, ~0enwwlksnge1, ~1wlkpwwlkf1ouspgr, ~1wlkisowwlkupgr added
* section "Walks in regular graphs" added with revised theorems


benjub:
> " ` F ` is a bijection between ` A ` and ` B ` ". 

I would write: "from A to B" or "from A onto B"

avekens:
> > " `F` is a bijection between `A` and `B` ".
> 
> I would write: "from A to B" or "from A onto B"

I think this can also be written, but this is the usual reading for ordinary functions (df-f) or onto functions (df-fo). Therefore, it needs not  be metioned again in the comment for df-f1o. The reading  "between `A` and `B` " is often used in literature and the web, and it makes clear that the order of the two sets is irrelevant (from A to B is equivalent to from B to A in this case).

benjub:
> the order of the two sets is irrelevant (from A to B is equivalent to from B to A in this case).

Of course the order is relevant ! The exponential function is a bijection from \R onto \R_{>0}, but not the other way around.

avekens:
> > the order of the two sets is irrelevant (from A to B is equivalent to from B to A in this case).
> 
> Of course the order is relevant ! The exponential function is a bijection from \R onto \R_{>0}, but not the other way around.

Of course its not the same function if you reverse the order, but there is a bijection from B to A. And a little bit of order is also indicated by "between A and B" (A is mentioned first). Therefore, I think it is still valid to say "bijection between A and B" (it's not my invention...). It's also not wrong to say "the exponential function is a bijection between \R and \R_{>0}".

benjub:
As you wish. I might be being over-pedantic.

PR #2026, opened 2021-05-12
lemmas on non-convergence and non-differentiability

acipsen:


nmegill:
In the comment for unbdqndv, the backticks (for function value) need to be escaped with double backtick (\`\`) to avoid ambiguity with the symbol string delimiters i.e.
```
` ( ( ( F ` y ) - ( F  ` x ) ) / ( y - x ) ) `
```
should be
```
` ( ( ( F `` y ) - ( F `` x ) ) / ( y - x ) ) `
```
(No error was detected in 'verify markup' since you have an even number of function values, so it thought there are 2 symbol strings.)

Could you fix this and resubmit? Thanks.

acipsen:
Thank you for spotting the typo.  It should be fixed now.

PR #2031, opened 2021-05-27
Shorten proof of disjxiun

jamesjer:
Most of the savings comes from using impbid1 instead of impbid, but I also squeezed a few more bytes out by moving where unneeded hypotheses are dropped.  The original size of the disjxiun proof is 2243 bytes; after this change it is 2164 (and 3752 steps compared to the original 4731).


nmegill:
@jamesjer When a proof is shortened by hand, we typically place a copy of the original proof, with its label suffixed with "OLD", after the shortened proof.  While this is optional, we recommend it because it is useful for people who want to compare it to the old proof (and better appreciate your hard work).  The \*OLD version is kept for about a year then deleted.

A recent example is ~ener and ~enerOLD.  You can see that the comment for ~enerOLD was changed to add `Obsolete proof of ~ ener as of 1-May-2021.` which provides the date that is referenced for deletion after a year; `(Proof modification is discouraged.)` which prevents "minimize" from reducing the proof to 1 step referencing ~ener; and `(New usage is discouraged.)` which prevents proof assistants from using it in other proofs.

jamesjer:
Ah, my apologies.  I'm still learning the conventions surrounding metamath, I'm afraid.  Should I submit another pull request to add an OLD version, or leave things as they are?

nmegill:
@jamesjer Oh, no apology needed - I didn't mean to sound critical. :)  There is no easy way you could have known, and I should have noticed it before merging but overlooked it.  At this point you need to submit another PR to add the OLD version, but it's no big deal.

benjub:
FWIW, I noticed in the previous PR that there was no *OLD version added, so I compared
http://us2.metamath.org/mpeuni/disjxiun.html
with
http://us2.metamath.org/old/mpeuni/disjxiun.html
(in addition to the reduction by 9 essential steps, I appreciate that the longest line in the new proof is shorter).

I think the /old directory hosts the previous version of us2 each time it is updated, so the time that older proofs are kept can fluctuate a lot, so the *OLD method is of course better (but the /old directory was useful to me on a few occasions).

nmegill:
@jamesjer BTW in addition to the reformatting described at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md, you will also have to regenerate the "discouraged" file as described later on that page.  The "...is discouraged.)" comments in the \*OLD will cause that file to change.  The "discouraged" file is used by the CI validations to flag any unintentional uses or proof modifications of discouraged theorems.

PR #2034, opened 2021-06-01
Generalized topology.

arpie-steele:
- Beginning of some theorems to show that specifying a generalized
  closure function is equivalent to specifying a generalized interior
  function which is equivalent to specifying a generalized neighborhood
  (of a point) function, and logical equivalences of conditions on them.
- A demonstration of three ways to say a closure/interior function
  is monotone/isotonic.
- Utility theorems to show that if a binary relation holds and the
  relation is a function or operator value then the base set exists.


benjub:
@arpie-steele I haven't read your PR carefully, but just in case you are not aware of it, there is some work about "Moore collections" in set.mm, around similar questions. (I mention this because I myself missed it some time ago and did some redundant work --- the term "Moore collections" is not very widespread.)

arpie-steele:
@benjub Thank you. I was aware of Moore closure spaces and the work in set.mm (particularly the comment in ~ df-mrc ) but I am also targeting other structures following the introduction of functions and operators in #2027 
I hope to soon have all equivalences for those Kuratowski’s axioms for topology from "Sur l'opération A̅ de l`Analysis Situs." (1922) (Another result from there is in ~ kur14 ) . 
I suspect "Sur la notion de limite topologique d’ensembles" (1958-1959) would also be helpful, but I don't think it is available online, so I am working off of later papers by Bärbel M. R. Stadler and Peter F. Stadler which summarize some of the general theorems.

PR #2035, opened 2021-06-01
Shorten proof of fzo0ssnn0

jamesjer:


benjub:
Could you add the theorem |- ( 0 ... N ) C_ NN0 ?
It may in turn shorten the proof of fzo0ssnn0 (as well as others).

jamesjer:
Sure, I can do that.  Do you mean as part of this pull request, or separately?  I would prefer separately, as it seems only somewhat connected to this pull request.

nmegill:
@jamesjer You can do |- ( 0 ... N ) C_ NN0 in another PR.

@benjub Note that we have fz1ssnn already.  Typically limits would be 1 to N for a 1-based usage and 0 to (N-1) for a 0-based usage, the latter being expressed with 0 ...^ N.  Do you think  ( 0 ... N ) C_ NN0  would have applications?

benjub:
|- ( 0 ... N ) C_ NN0 is an intermediate step in fzo0ssnn0OLD so it would obviously have shortened it, and maybe it even results in a proof shorter than fzo0ssnn0.  I understand your point and I don't know how useful it would be.  It was more a matter of aesthetics on my part, since ( 0 ... N ) seems simpler than ( 0 ..^ N ) or ( 1 ... N ).  But as you said, it has N+1 elements whereas the other two have N elements (which corresponds to the symbol appearing in these expressions), and ..^ is useful when considering unions of contiguous intervals, which probably makes it more useful. So... either way !

jamesjer:
The suggested theorem (fz0ssnn0?) can be used to minimize the following:

 - fzossnn0: 155 to 138 bytes
 - mertenslem1: 7619 to 7616 bytes
 - bpolylem: 2042 to 2039 bytes
 - nn0gsumfz: 678 to 646 bytes
 - gsummptnn0fz: 832 to 819 bytes
 - psrbaglefi: 1478 to 1463 bytes
 - coe1mul2lem2: 773 to 755 bytes
 - pmatcollpw3fi: 457 to 431 bytes
 - plypf1: 7711 to 7708 bytes
 - aannenlem1: 3667 to 3651 bytes
 - iseupa: 2156 to 2147 bytes

That's a total savings of 155 bytes, versus 50 bytes for the proof of the new theorem.

I'm running full minimization on each of the theorems listed above, which has already used `f1oeq3d` and `fz1ssnn` to further shorten a couple of the proofs.  Once that is done I'll make a new pull request.

PR #2036, opened 2021-06-01
Add fz0ssnn0 and minimize 11 theorems

jamesjer:
As noted in https://github.com/metamath/set.mm/pull/2035, this adds a 50 byte proof and reduces 11 other proofs by a total of 155 bytes.  We can also remove 1 dv from `coe1mul2lem2` (due to removing an `ssriv` step).  I ran a full minimization on each of the 11 theorems; a few were further reduced as a result.


benjub:
How did you notice that the dv condition could be removed ? Was it automatically detected ? And how did you notice these 11 theorems ? (typically, I do things like "MM>show usage xx1*,xx2" and use the script "scripts/min.cmd")

jamesjer:
I looked at each of the proofs to see if `ssriv` was used in the subproof to be minimized and, if so, whether that meant a dv condition could be removed.  So it was a manual process.  For the 11 theorems, I have a little patch to metamath that I wrote a few years back that searches for minimization possibilities without actually changing any proofs.  It is probably obsolete due to scripts/min*, but I like being able to see the minimization possibilities without changing anything.  I can share it if it is of interest.

nmegill:
@jamesjer This is good.  BTW you may already know this, but in case you don't, we normally don't put "(Proof shortened by...)" in a proof shortened with the "minimize" command, and there is no need for an OLD version; just save the minimized proof and don't modify the theorem's comment.

jamesjer:
Yes, I saw that mentioned somewhere.  I do have a question, though.  Let's say that a proof can be shortened by a few bytes by making 2 changes, so MINIMIZE_WITH won't find it.  Is that worthy of a "(Proof shortened by ...)" comment and an OLD version?

I mention this because I have found a couple of proofs that can be shortened by a few bytes apiece by replacing `syl2anc` with either `syl2an2` or `syl2an2r`, but that also requires a second change due to the difference between `syl2anc.1` and `syl2an2.1`.  So MINIMIZE_WITH won't make these changes, but they seem so small they hardly seem worth the bother of the comment and the OLD version.

nmegill:
> Yes, I saw that mentioned somewhere. I do have a question, though. Let's say that a proof can be shortened by a few bytes by making 2 changes, so MINIMIZE_WITH won't find it. Is that worthy of a "(Proof shortened by ...)" comment and an OLD version?

In that case it is up to you.  I typically don't add my name to the comment or add a \*OLD if I do a minor shortening, but other people do.

benjub:
> I looked at each of the proofs to see if `ssriv` was used in the subproof to be minimized and, if so, whether that meant a dv condition could be removed. So it was a manual process. For the 11 theorems, I have a little patch to metamath that I wrote a few years back that searches for minimization possibilities without actually changing any proofs.

I'd be glad to see the script. Surely you don't try and minimize all the theorems in the database ? How do you select which ones to try to minimize ? For instance, the script "scripts/min.cmd" by @nmegill permits to search for minimizations of proofs using given theorems (given by pattern-matching) with another given set of theorems (given by pattern-matching). This is explained in the comments of the script.

I agree with @nmegill that for very small proof shortenings, attributions and *OLD are not necessary.  Up to you to decide where to draw the line.

**Edit**: scripts/min.cmd also allows "dry runs" as explained in its comments

jamesjer:
It isn't a script, but rather a patch to metamath itself: http://jamezone.org/pleasure/mathematics/metamath-bulk-min.patch.  As I said, it should probably be considered obsolete, given the existence of the scripts/min* scripts.  It adds a `BULK_MINIMIZE_WITH` command:
```
BULK_MINIMIZE_WITH my_new_theorem [/FROM name] [/TO name] [/TIME] [/OVERRIDE]
```
where:

- `/FROM` names the first theorem to check, defaulting to the one immediately following `my_new_theorem`
- `/TO` names the theorem where checking stops, defaulting to `mathbox`
- `/TIME` tells you how much time you spent on this endeavour
- `/OVERRIDE` indicates that theorems marked with `(Proof modification is discouraged.)` should be checked

PR #2037, opened 2021-06-02
revising proof of mreexexd and updating mathbox

Achaemenid:


benjub:
What is the nature of the revision of mreexexd ? It seems that the statement and comment are unchanged, so is this a proof shortening ? (if this is the case, "Revised by" should be "Proof shortened by")

Achaemenid:
> What is the nature of the revision of mreexexd ? It seems that the statement and comment are unchanged, so is this a proof shortening ? (if this is the case, "Revised by" should be "Proof shortened by")

It is a little shorter, but the main effect was to remove the dependence on ax-ac2 and ax-rep.

benjub:
That's very nice !  I propose you add:

(Contributed by David Moews, 1-May-2017.)  Remove dependencies on ~ ax-rep and ~ ax-c2 .  (Revised by Brendan Leahy, 2-Jun-2021.)

nmegill:
Gérard Lang asked the following question:

Seeing the theorem curunc (32144), with conclusion curry uncurry  F = F, would it be also true under the same hypothesis (that uncurry curry F = F ?
Très amicalement.
Gérard Lang

Achaemenid:
> Gérard Lang asked the following question:
> 
> Seeing the theorem curunc (32144), with conclusion curry uncurry F = F, would it be also true under the same hypothesis (that uncurry curry F = F ?
> Très amicalement.
> Gérard Lang

I didn't do it because it looked a little harder and I wasn't expecting to need it, but I'm pretty sure it'd need a stronger antecedent, the same way curfv has a stronger antecedent than uncov.

PR #2038, opened 2021-06-02
Reduce axiom usage in ax13

wlammen:
1. ax13 does not base on ax-10 and ax-12 any more. In the wake, equvini has lost its dependency on ax-10, too.
2. remove some outdated OLD theorems
3. I did not update the file date, because some other pull requests are in the queue, but 7-May-2021 shows it has not been touched for quite some time.
4. Rename axc9lem1/2 to ax13lem1/2.  ax13lem1 is used in the context of ax13, and both are closer to ax13 than to axc9 (IMO).

Shouldn't have done "Merge remote-tracking branch 'upstream/sctfen' into develop" . It was a bit of a pain to revert this, but finally I am convinced I succeeded.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1415,6 +1415,16 @@
 "ax12indn" is used by "ax12indi".
 "ax12v2-o" is used by "ax12a2-o".
 "ax12v2OLD" is used by "ax12a2OLD".
+"ax13lem1" is used by "ax13".
+"ax13lem1" is used by "ax13lem2".
+"ax13lem1" is used by "ax6e".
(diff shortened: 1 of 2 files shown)


nmegill:
> 1. ax13 does not base on ax-10 and ax-12 any more. In the wake, equvini has lost its dependency on ax-10, too.

Nice!

david-a-wheeler:
> ax13 does not base on ax-10 and ax-12 any more. In the wake, equvini has lost its dependency on ax-10, too.

That's really cool!

PR #2041, opened 2021-06-04
Develop addition theorems.

arpie-steele:
Expository information.
- Kuratowski's Axioms K0', K0, K1, KA equivalent forms listed
- 13 spaces characterized in terms of axioms

References.
- HTML comments have references to books and papers as I try to verify statements

Theorems.
- Relative contraposition law for subsets. This time bidirectional.
- Proof K0 -> K0' ; ( K0' + K1 ) -> K0
- Equivalence theorems for various expressions for K0, KA and K1, as well as the convergents form of K0'


arpie-steele:
@benjub Following up on #2027 and #2034 here is another chunk. This one has two HTML table that explain the equivalences and their relation to the properties of the pseudo-topological spaces and how 13 of those spaces relate to those properties.

In the HTML there are references that I used to name the spaces as the field of math is large and naming things is hard.

benjub:
In the table, what do you mean by "Assuming a prefix of: ∀𝑥 ∈ 𝐵∀𝑠 ∈ 𝒫 𝐵∀𝑡 ∈ 𝒫 𝐵" ?
Do you mean: "In the following formulas, if x, s, or t is not explicitly quantified, then it is implicitly universally quantified and x (resp. s, t) ranges over B (resp. ~P B)" ?  I'm asking this because some of your formulas have "E. t e. ~P B" so you probably don't want to prefix "A. t e. ~P B ". Or maybe simply change that "t" to "u".

I'm not sure rows are well-aligned. Is K1, monotonic, isotonic supposed to be one block ?  Do you plan to complete the table with the axioms you mention below: KB, K2, K3, K4, K5 ?

I haven't checked everything, but I hope that convergence spaces appear somewhere, and that the fact that neighborhoods of a point form a filter appears somewhere in the hierarchy.

arpie-steele:
> In the table, what do you mean by "Assuming a prefix of: ∀𝑥 ∈ 𝐵∀𝑠 ∈ 𝒫 𝐵∀𝑡 ∈ 𝒫 𝐵" ?

> Do you mean: "In the following formulas, if x, s, or t is not explicitly quantified, then it is implicitly universally quantified and x (resp. s, t) ranges over B (resp. ~P B)" ? I'm asking this because some of your formulas have "E. t e. ~P B" so you probably don't want to prefix "A. t e. ~P B ". Or maybe simply change that "t" to "u".

That's it exactly. I already fixed one, and now I fixed the other t -> u case. It will be in the next push.

> I'm not sure rows are well-aligned. Is K1, monotonic, isotonic supposed to be one block ? Do you plan to complete the table with the axioms you mention below: KB, K2, K3, K4, K5 ?

I'm mostly not working with primary literature and all three names have been given to the conditions in that row, all of which are logically equivalent.  The remainder I plan to complete soon (this month).

> I haven't checked everything, but I hope that convergence spaces appear somewhere, and that the fact that neighborhoods of a point form a filter appears somewhere in the hierarchy.

According to my notes: if K0' + K1 + KA + K3 hold then the (pseudo-)neighborhood of each point is a filter, giving the properties of being (K0') a non-empty collection, (K1) upward-closed, (KA) empty set is not a member and (K3) closure under finite intersection. That would include Smyth Spaces and following in the second table. 

From M.B. Smyth, "Semi-metrics, closure spaces and digital topology." Theoretical Computer Science. 151:257-276, 1995.

> **Definition 2.1**. A _neighbourhood space_ is a pair (X, N), where N: X -> ~P ~P X assigns to each point x a (not necessarily proper) filter N_x; of subsets of X (the neighbourhoods of x).

I probably should change the label to "Smyth neighborhood space" for clarity.

benjub:
Thanks @arpie-steele for the precisions (hopefully you'll include the ones you see fit in the comment in set.mm).  The phrase "neighborhood space" is probably enough, since the idea of associating to a point a filter of subspaces called its neighborhoods probably dates back to long before 1995.  Also include (pre)convergence spaces when they show up !

PR #2042, opened 2021-06-05
taking Benoit's suggestion in mreexexd description, also adding unccu…

Achaemenid:
…r at GL's suggestion and citing fin2so


nmegill:
GL sent me the following additional comment.

Dear Norm, 
Coming back over currying, we could consider that, being given a function of two variables F:AxB —> C, we have two « natural » notions of « currying ».
The first one is the current one in set.mm, that could be named « left currying ,» builds a family of functions of one variable indexed by X in A, Fx:B—> C, with Fx<y>=F<x,y>.
The sexcond one, that could be named « right currying », builds a family of functions of one variable indexed by y in B, yF:A—> C, with yF<x>=F<x,y>.
In fact, there is no need to define « rigth curring », because it is possible to define this notion from the current (left-)currying by using the notion of « symetrization »F# of a function of two variabl as follows:
if  F is defined as F:AxB—> C, then the symetrized function of F is the function F# is defined as F#:BxA—>C with F#<y,x> = F<x,y>.
Indeed, we have curry F# = (y e. domdomF#—> (<x,z>/ <y,x>F#z)) = (y e. ranranF —> (<x,z>/ <x,y> Fz)) = right-curry F.

 do not know if such a remark could be an useful comment in the the definition def-cur (7088).

Très amicalement.
Gérard Lang

PR #2044, opened 2021-06-10
New approach for graph theory (continued)

avekens:
By this PR, section "Graph theory (old)" is removed completely from my mathbox. Theorems which were worth to keep are revised and moved to section "Graph theory (revised)", which will replace PART 16 GRAPH THEORY soon.

Here the details of this PR:

Main set.mm:
* theorem ~0in moved from GS's mathbox
* theorems ~resdifcom, ~fundif added
* theorems about/with maximum (~2resupmax, ~lemaxle, ~ssfzunsn) added
* theorems for sSet (~setsdm, ~setsfun, ~setsfun0, setsstruct) added

AV's mathbox:
* theorems ~fundmge2nop0, ~fun2dmnop0 added; ~fundmge2nop0, proofs of ~fun2dmnop0 shortened; ~funvtxdm2val, ~funiedgdm2val, ~funvtxval0, ~funvtxval,  ~funvtxdmge2val, ~funiedgdmge2val, ~funiedgval, ~grstructd, ~grstructeld revised
* ~uhgrstrrepelem, ~uhgrstrrepe (formely in section "Graph theory (old") added
* ~umgr2edg1, ~umgrvad2edg, ~umgr2edgneu  added
* ~uhgr0edgfi, ~isfusgrcl, ~sPthdifv, ~usgr2pthspth, ~usgr2pthlem, ~usgr2pth, ~usgr2pth0  (formely in section "Graph theory (old)") added to section "Graph theory (revised)"
* ~nbgrcl, ~usgr2trlncl, ~usgr2trlspth, ~usgr2trlncrct added
* section "Graph theory (old)" deleted

PR #2045, opened 2021-06-11
Proofs minimized using ~0in

avekens:
As suggested by @benjub (see #2044), I used the script min.cmd to find proofs which can be shortened using ~0in (instead of ~in0), which was moved from GS's mathbox to main set.mm with the previous PR. 15 proofs could be minimized!


benjub:
It's interesting that minimize_with couldn't shorten ~pred0, whereas the shortening is obvious to a human being.  On the other hand, starting from an incomplete proof, with only ~eqtri and ~df-pred filled in, the command MM-PA>improve all, with possibly options "/depth x /y" would have probably completed the proof to give the shorter one, and probably mmj2 as well on the command "ctrl+U".

nmegill:
> It's interesting that minimize_with couldn't shorten ~pred0, whereas the shortening is obvious to a human being.

The reason is that 'minimize_with' can normally work with only one statement at a time.  in0 and incom are part of a 3eqtri chain, and the "output" of that 3eqtri is not a 0in instance, which is what 'min' tries to match.

Note that breaking up 3eqtri will cause 'min' to use 0in:

MM-PA> expand 3eqtri  ( <- expands 3eqtri into 2 eqtri's)
Proof of "pred0" increased from 82 to 85 bytes after expanding "3eqtri".
MM-PA> min 0in  ( <- works now)
Proof of "pred0" decreased from 85 to 69 bytes using "0in".

Of course you need some knowledge of both the proof and how 'min' works to use this trick, so the technique probably isn't generally useful, but I've used it once or twice.  And there's a 50-50 chance the 2 eqtri's will be connected the wrong way i.e. neither of them has a 0in match as its output.

I wonder if doing 'expand 3\*tr\*' routinely before 'min' to break up all 3 way chains might give slightly better results in general ('min' should restore the unused expansions), although I've never tried that.  Maybe try it with and without 'expand' to be sure that 'expand' doesn't make the proof longer in the end, for some reason I haven't thought of.

PR #2047, opened 2021-06-12
proofs minimized using ~0in - 2. step

avekens:
As Norm proposed in #2045, I enhanced the script to minimize all proofs using ~in0 by adding "expand 3`*tr`*" (without blanks, of course - it seems that this editor in github would convert the asterisks otherwise...) as follows:

prove 1pthdlem2
expand 3`*tr`*
min in0/no_new ax-**
save new_proof/compressed
_exit_pa/force

This resulted in longer proofs in many cases, because the expansion was not reverted! Therefore, I modified the script again (global minimization):

prove 1pthdlem2
min *
save new_proof/compressed
_exit_pa/force

This resulted in the minimization of 22 proofs (using ~in0)
4 among them using ~0in instead of ~in0:
* ppi1
* cht1
* ballotlemfp1
* mblfinlem2 (using both 0in and in0!)

The others would have been minimized either anyway (without expansion) or because of the expansion only (which allowed for other theorems to be used for minimization).

It is also interesting that 2 proofs were changed, but had the same length as before (~fresaun, ~ fctop),  and 2 proofs had (a little) longer proofs than before (~smumullem, ~dfpo2). These 4 proofs are not changed by this PR.

Furthermore, 10 proofs where converted in exactly the same proofs after expansion followed by the global minimization.

I wonder, however, if the two steps I have performed can be combined in a single step, maybe as follows:

prove 1pthdlem2
expand 3`*tr`*
min in0,3`*tr`*/no_new ax-**
save new_proof/compressed
_exit_pa/force

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1422,8 +1422,10 @@
 "ax13lem1" is used by "nfeqf2".
 "ax13lem1" is used by "wl-19.2reqv".
 "ax13lem1" is used by "wl-19.8eqv".
+"ax13lem1" is used by "wl-dveeq12".
 "ax13lem2" is used by "nfeqf2".
 "ax13lem2" is used by "wl-19.2reqv".
+"ax13lem2" is used by "wl-dveeq12".
 "ax13lem2" is used by "wl-speqv".
 "ax5el" is used by "dveel2ALT".
 "ax5eq" is used by "dveeq1-o16".
@@ -14336,8 +14338,8 @@ New usage of "ax12vOLD" is discouraged (0 uses).
 New usage of "ax12vOLDOLD" is discouraged (0 uses).
 New usage of "ax13OLD" is discouraged (0 uses).
 New usage of "ax13fromc9" is discouraged (0 uses).
-New usage of "ax13lem1" is discouraged (7 uses).
-New usage of "ax13lem2" is discouraged (3 uses).
+New usage of "ax13lem1" is discouraged (8 uses).
+New usage of "ax13lem2" is discouraged (4 uses).
 New usage of "ax1cn" is discouraged (0 uses).
 New usage of "ax1ne0" is discouraged (0 uses).
 New usage of "ax1rid" is discouraged (0 uses).


avekens:
PS: the global minimization of the 53 involved proofs took several hours on my machine...

avekens:
... something wrong with the verification script?

nmegill:
This is odd.  Something went wrong with the mmj2 build; it looks like a tar file got truncated?  Try it again.  @david-a-wheeler Any idea what went wrong?

BTW to escape "\*" in GitHub markup, use "\\\*".

Regarding 'minimize_with':  In 2019 the behavior was changed to prevent the usage of any new $a's by default.  Search for "10-Aug-2019" on the Most Recent Proofs page.  To allow new df-\* but prohibit any new ax-\*, you would use

"minimize_with \* /allow_new_axioms \* /no_new_axioms_from ax-\*"

or

"minimize_with \* /allow_new_axioms df-\*"

For set.mm these do the same thing.  The former provides for better fine-tuning e.g. if you want to avoid ax-ac2 but allow other new axioms.  See 'help minimize_with' for more details; in particular:  "The use of / NO_NEW_AXIOMS_FROM without / ALLOW_NEW_AXIOMS has no effect."  I know this is confusing, but after discussing it we felt it better to have 'minimize_with' be ultra-conservative, so you have to go out of your way to allow the use any new $a's.

nmegill:
> This resulted in longer proofs in many cases, because the expansion was not reverted!

I would be interested in an example of not reverting to see what's going on.

benjub:
> after discussing it we felt it better to have 'minimize_with' be ultra-conservative, so you have to go out of your way to allow the use any new $a's.

I remember that discussion (on the google group) and I agree it's very important to prevent involuntary addition of new dependencies on $a-statements.

avekens:
OK, now all checks worked.
Here an example Norm requested:
Initially, res0 was:
  res0 $p |- ( A |` (/) ) = (/) $=
    ( c0 cres cvv cxp cin df-res 0xp ineq2i in0 3eqtri ) ABCABDEZFABFBABGLBADHI
    AJK $.
After "expand 3*tr*" and "min in0/no_new ax-**" it was:
 res0 $p |- ( A |` (/) ) = (/) $=
    ( c0 cres cvv cxp cin df-res 0xp ineq2i in0 eqtri ) ABCABDEZFZBABGMABFBLBAD
    HIAJKK $.
After "min *" it was (again):
  res0 $p |- ( A |` (/) ) = (/) $=
    ( c0 cres cvv cxp cin df-res 0xp ineq2i in0 3eqtri ) ABCABDEZFABFBABGLBADHI
    AJK $.

I think thie reason is that "min in0" only tries to minimize using in0, and not 3eqtri.

david-a-wheeler:
The retry worked, so it looks like the problem was a temporary hiccup.

PR #2048, opened 2021-06-15
Mathbox: peirce, curry, roll

benjub:
@nmegill I think that bj-peirceroll is noteworthy enough to be moved to the main part: it shows, in closed form and from ax-mp, ax-1, ax-2 only, an implication among two classical axioms (Peirce and Roll).  It will probably shorten ~ looinv.  Keep the current proof as ~ looinvALT ? Also, add a ~ roll of which ~ looinv is a special instance ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14547,6 +14547,7 @@ New usage of "bj-ceqsalgvALT" is discouraged (0 uses).
 New usage of "bj-con4iALT" is discouraged (0 uses).
 New usage of "bj-consensusALT" is discouraged (0 uses).
 New usage of "bj-csbsnlem" is discouraged (1 uses).
+New usage of "bj-currypeirce" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> @nmegill I think that bj-peirceroll is noteworthy enough to be moved to the main part: it shows, in closed form and from ax-mp, ax-1, ax-2 only, an implication among two classical axioms (Peirce and Roll).

OK.

> It will probably shorten ~ looinv. Keep the current proof as ~ looinvALT ? Also, add a ~ roll of which ~ looinv is a special instance ?

I don't have any particular preference for the current looinv proof.  I added it on a whim after attending a 2004 lecture by Ted Ulrich.  I'll leave it up to you; if you think it is interesting that the current looinv proof illustrates one of the rare uses of peirce, then go ahead and create an ALT, otherwise just shorten the proof.  Since looinv = *2.69, there is also a proof in pmproofs.txt.

PR #2052, opened 2021-06-19
Topological axioms K3 and K13

arpie-steele:
Exposition
- Tables scrubbed of previous claims that K0+KB->KA, K3->KB
- Hidden comment: Stadler paper has errors 
- Hidden comment: counter example to Stadler claim K0, K2, K3, K4 imply K1
- Table entries for equivalent expressions for axioms K3 and K13
- Rebaselined spaces that required K1 and K3 to require K13

Theorems
- ntrkbimka: KB alone is sufficient to imply KA
- ntrk0kbimka: Obsolete proof of Stadler's claim K0, KB imply KA
- clsk3nimkb: An infinite family of counter-examples disprove Stadler's claim K3 implies KB
- ntrk1k3eqk13: K13 is equivalent to asserting both K1 and K3
- ntrclsk3, ntrneik3, ntrneix3: equivalencies to closure being sub-linear
- ntrclsk13, ntrneik13, ntrneix13: equivalencies to closure being finitely additive


arpie-steele:
@benjub
In both "Basic Properties of Closure Spaces" and "Generalized Topological Spaces in Evolutionary Theory and Combinatorial Chemistry" the claim seems to be made in a table that appears in both that K0, K2, K3, K4 imply K1
I think I found a counter-example: ( s e. ~P { 1o , 2o , 3o } |-> if ( s = { 1o } , { 1o , 2o } , s ) )

On page 12 of  "Basic Properties of Closure Spaces", the following three claims are made verbatim:

> (KB) and (K0) implies (KA).
> (K2) implies (KB).
> (K3) implies (KB).

But while pursuing these I have proven while K2 implies KB, KB (alone) implies KA and that K3 does not imply KB. I have updated the tables to reflect these.

benjub:
Thanks @arpie-steele for the update. A few comments:
* can you "unhide" your comments, so that they appear on the webpage ?
* can you give the precise references to the papers, and, when you quote a claim in a paper, to the place within that paper (theorem number or page) ? (sorry if I missed those)
* once you have sorted things out, I advise you to contact the authors (both, even if only one of them is the "corresponding author"), but in a non-metamath language.
* pure form: in your table, can you put the boldface "neighborhoods, interior, ..." as column titles, rather than on the second line of each column ?
* maybe recall above the tables the correspondences `x e. ( I `` A ) <-> A e. ( N `` x )` etc.
* montonic --> monotonic
* I realized that in the line K1, you mean "isotonic, aka monotonic", and that the three lines of formulas are actually not to be understood as the conjunction of three properties, but are equivalent
* maybe uniformize the notation, using the fact that union is akin to addition and intersection to multiplication, so use "sub-additive, super-additive, additive, sub-multiplicative..." (these for the binary union/intersection, and "finitely ..." for the conjunction of nullary and binary union/intersection... maybe?). I think that "monotonic" should actually be "increasing" (or with the English convention, "non-decreasing").
* in the equivalences within each row, sometimes you have the stronger property that the equivalence holds for fixed x and/or s and/or t. You probably used it in your proofs and even detached these as lemmas, and if not you might want to do so. Interestingly, in the three lines for K1, this is true for some but not all lines (and if you insist on (s,t)-wise equivalence, you get another property for neighborhoods, e.g. A e. N(x) /\\ B e. N(x) -> A u. B e. N(x).

benjub:
On second thought, I think that the four formulas stating binary super-additivity and sub-multiplicativity of K and I can be removed from the table, as non-fundamental, and it can be simply noted below the table that increasing-ness implies super-additivity and sub-multiplicativity for arbitrary unions/intersections.

arpie-steele:
@benjub  I am in progress of reorganizing the table and adding lemmas. Here is what I have on the papers with errors.
I will email the authors.

Bärbel M. R. Stadler and  Peter F. Stadler. "Basic Properties of Closure Spaces" (2001).  is a preprint/unpublished technical report dated to 2001 from the time stamp and the file naming convention.
https://www.bioinf.uni-leipzig.de/~studla/Publications/PREPRINTS/01-pfs-007-subl2.pdf
It appears to be one of two studies associated with an early draft of "Generalized Topological Spaces in Evolutionary Theory and Combinatorial Chemistry"

On page 10, Table 1 has some errors. Under "interior" the entry for K2 should read "int(A) ⊆ A" and the entry for K3 should read "int(A) ∩ int(A) ⊆ int(A ∩ B)." The row for K5 is obviously incomplete.

On page 12, Table 2 seems to make the claim that K1 is implies by the other properties, K0, K2, K3, K4 of a Topology.

- Counter example. For the set {a, b, c} a pseudo-closure function which gives cl({a}) = {a, b} and otherwise returns its argument unchanged. 
- Proof: cl(∅) = ∅ [K0] ; Two easy cases show ∀A∈𝒫{a, b, c} A ⊆ cl(A) [K2] ; Since cl({a, b}) = cl({a}) ∪ cl({b}) = {a, b} it is easy to see ∀A,B∈𝒫{a, b, c} cl(A∪B) ⊆ cl({A}) ∪ cl({B}) [K3]; Two easy cases show ∀A∈𝒫{a, b, c} cl(cl(A)) = cl(A) [K4] ; but cl({a}) is not a subset of cl({a, c}) so this pseudo-closure function cannot have the K1 property of isotony

Also on page 12, are the claims "(KB) and (K0) implies (KA)." and "(K3) implies (KB)." which are weaker than needed and wrong, respectively.

KB is sufficient to prove KA.

- Proof: If KB holds, then disjoint subsets have disjoint interiors. But the empty set is disjoint with itself, so the interior of the empty set is the empty set, which is the property KA.
- https://us.metamath.org/mpeuni/ntrkbimka.html

It is not true that K3 implies KB since a pseudo-closure function which gives the relative complement of the subset satisfies K3 but not KB.
- Proof: cl(A ∪ B) = X \ (A ∪ B) = ( X \ A ) ∩ ( X \ B ) ⊆ ( X \ A ) ∪ ( X \ B ) = cl(A) ∪ cl(B)  so [K3] but for non-empty base sets, cl(∅) = ( X \ ∅ ) = X ≠ ∅ which violates property KB.
- https://us.metamath.org/mpeuni/clsk3nimkb.html

Bärbel M. R. Stadler and Peter F. Stadler. "Generalized Topological Spaces in Evolutionary Theory and Combinatorial Chemistry." _J. Chem. Inf. Comput. Sci._, **42**:577–585, 2002. Proceedings MCC 2001, Dubrovnik.
Preprint: http://citeseerx.ist.psu.edu/viewdoc/citations?doi=10.1.1.15.8351
Table 2 on page 3 repeats the error of Table 2 of "Basic Properties of Closure Spaces".
The other problematic claims do not appear here.

PR #2056, opened 2021-06-26
Measurable functions w.r.t. a sigma-algebra

glacode:
This section contains the proofs for most of the theorems in section 121 of [Fremlin1].


nmegill:
A merge conflict occurred w/ previous PR, most likely in the list of moved theorems at the top of set.mm.

PR #2061, opened 2021-07-04
remove dependency on ax-10 in stdpc5

wlammen:
The bi-conditional in 19.21t is asymmetric: One direction is dependent on ax-10, the other not. Be explicit about this, and reduce axiom usage in stdpc5.
I wonder whether the same asymmetry holds in 19.23t. This is harder to detect, because nfnt drags ax-10 into the proof, and must be replaced to achieve a similar result.


benjub:
The axiom dependencies on ax-10 and ax-12 may change (in many cases, one or both removed, and in fewer cases, one or both added) when the change is made in the definition df-nf to adopt nf2 (see https://groups.google.com/g/metamath/c/Ovxv2aXJOIM/m/BVU7SvgFBwAJ and ~wnff and following theorems).

Unfortunately, I haven't had time to make this change yet (and the change has many implications, so I'm not sure yet how to do it). Hopefully the summer vacations will give me some time. I'm open to suggestions on the best way to do this change (probably, first change all concerned theorems between the current df-nf and nf2 (and moving some after nf2 if necessary), and in a second phase, look at later theorems that may benefit from this change to remove axiom dependencies).

wlammen:
nf2 as a definition looks nicer, and replacing the antececedent -. x = y with -. A. x x = y in ax-13 could help prevent ax-12 and ax-10 kicking in very early, I guess. Both axioms are omnipresent now, which can be seen as a drawback, because their individual impact on predicate logic appears somewhat blurred.  I'd further like ax-12 be split into ax12v and axc11, because both parts cover very different aspects.

As for my spare time: I am currently snowed under. This weekend was the first one I could dedicate more time to Metamath again after quite some time. If possible I will help out with the transition, but my time is limited.  I work full time, and, sadly, often overtime, in a complete different field outside of formal logic or mathematics.  So don't expect too much.

benjub:
@wlammen : Actually, I would replace ax-13 with ax-c9 and I would split ax-12 too. I had envisioned replacing ax-12 with {sp, bj-ax12} (bj-ax12 is simply the generalization over x of ax12v2), but @nmegill told me it would probably be incomplete since axc11 would probably not be provable. He proposed {sp, bj-ax12, axc11} (which is similar to your proposal).  There is also the possibility {sp, "A. x ax-12"} but I have to think more.  The main reason I want to have sp as an axiom is because not having it makes ax-12 be required for many elementary theorems where only sp is needed, so it creeps in many places. Like you, I think this blurs the real necessary use of ax-12.

Anyway, this is hypothetical (@nmegill emphasizes the necessity to stay close to Tarski's work, which is a legitimate reason too).  The first step to see clearer should be to change df-nf to nf2, and I'll try to do it soon. If anyone has any advice for a smooth transition, I'm all ears.

PR #2064, opened 2021-07-06
Reduce axiom usage in axc16nf

wlammen:


benjub:
Note that the feasibility of proving axc16nf from ax-1--ax-7 followed from the comment of aev2 (but it's of course nice to have it explicitly).

wlammen:
I am not sure whether I understand you right. axc16nf still needs ax-12, and that is not dispensable. One aspect of ax-12 is that it forces equality to be the finest grained operation in logic. So objects not discernable by = cannot be differed by anything else *logical*. *Textually* one can see a difference between { x | x e. z } and { y | y e. z }, so if your universe consists of all well-formed text chunks describing a set, the predicte 'contains character y' could differ between these, but such a predicate must never be declared within our logic, or a contradiction with ax-12 arises. This is why the distinct variable condition cannot be made an antecedent (context) in a theorem, except for the distinctor expression, which somehow bypasses this limitation.

If we gave up on ax-12, we could establish such a predicate containsY, at the same time making = trivial by forcing equality among all objects, and A. x containsY would not hold.

benjub:
You're right, I commented too fast; the comment of aev2 applies only when there is no formula metavariable.

Indeed axc16 is not provable from ax-1--ax-7, but axc16 (and axc16g) are provable from {ax-1--ax-7, ax12v} (without the full ax-12). This is because only the version of axc11r with a DV condition is needed, and it is provable from ax12v.  I'm going to add this version of axc11r with a DV condition.

PR #2065, opened 2021-07-06
Independence (of K1) day.

arpie-steele:
- Added new table with emphasis on the functions, not the axioms.
- Renamed axiom X0' to KA' because K0 and KA are already dual.
- Since KA implies KA' added it to third table.
- Proved the independence of K1 from K0,K2-K4. Cited preprint that implied otherwise : clsk1ind*
- 3or utilities: or3or and andi3or
- Three cases for when a union is equal to a singleton: uneqsn
- Utilities showing when functions are related, they exist: clsneikex, neicvgmex, neicvgnex,
- Corollary showing when B is inferred to exist, so does the complement of S : ntrclsrcomplex, ntrneircomplex, neicvgrcomplexclsneircomplex
- Utilities to express the value of one function in terms of related function: ntrclsfv, ntrclsfveq1, ntrclsfveq2, ntrclsfveq, ntrneifv3, ntrneifv4, clsneifv3, clsneifv4, neicvgfv
- Removed unnecessary hypothesis from neik0imk0p


benjub:
Nice work @arpie-steele.  I have yet to look at this and your latest comment in PR #2052 more carefully... hopefully soon.

PR #2067, opened 2021-07-06
Shorten proofs of ceqsralt, elrab3t, euind, reu6, reu8, and reuind

jamesjer:
I've been studying the proofs in the section "The universal class" and found a few that could be shortened a bit.  Almost always, I simply reversed the order of two steps and then `MINIMIZE_WITH` did the rest, so I don't think "Proof shortened by" comments are warranted.


benjub:
Did you notice any change in axiom usage ? I see that in a proof (of  ceqsralt?), {vex, ceqsalv} was replaced by {equsalvw}, so maybe we gain something. If it's too hard to compare, then we can wait for the merge and check the old/mpeuni url.

jamesjer:
The output of `show trace_back <label> /axioms` for each is the same before and after, except for reu8, which has these in the before case, but not in the after case: 

- w3a
- df-3an
- df-clab
- cvv
- df-v

benjub:
Thanks @jamesjer. Avoiding df-clab is a real gain (and the others are also appreciable)!

**Edit:** equsalvw can probably be used in other places (typically when equsal, ceqsal(v)... are used) to reduce axiom usage.

PR #2068, opened 2021-07-06
[Mathbox] Young's inequality

DyeKuu:
- adds Young's inequality to KZ's Mathbox


benjub:
Nice !
Side note: the inequality also works for a or b equal to zero (assuming a>=0, b>=0, p>=0 suffices, since the hypothesis 1/p+1/q=1 implies p and q strictly positive). Recall that RR+ denotes in set.mm the (strictly) positive real numbers.

I don't know what @DyeKuu, @nmegill and @digama0 (and everyone) think, but if we begin to have a growing library of analysis inequalities (like Young's and the previous AMGM inequalities) it may be worth the trouble to have them in full generality for future use. If a new symbol is added, I'm a proponent of RR>=0 (displayed \mathbb{R}\_{\geq 0}) and RR>0 (displayed \mathbb{R}\_{>0}) to avoid confusion as to RR+ containing 0 or not.

DyeKuu:
I totally agree with extending these inequalities to their full version (including side cases of 0).
As for adding new definitions like \mathbb{R}\_{\geq 0}, I'm neutral with this proposal. I guess this may take a large amount of work on transforming all the theorem involving (A e. RR /\ 0 <_ A). But given the long-term advantage, I think it worth a shot.

nmegill:
> I don't know what @DyeKuu, @nmegill and @digama0 (and everyone) think, but if we begin to have a growing library of analysis inequalities (like Young's and the previous AMGM inequalities) it may be worth the trouble to have them in full generality for future use. If a new symbol is added,

Before RR+ was added, there were enough theorems with (A e. RR /\\ 0 < A) to make it apparent there would be a net benefit by introducing the RR+ definition and associated utility theorems like ~rpre.  Currently there are over 450 theorems using RR+.

RR\>=0 is used less frequently, but there are over 60 matches for ( A e. RR /\ 0 \<_ A ) so it may be at the point of being useful.  We have to be careful, though, to look at how these theorems are typically used.  If a proof requires unwinding the definition back to RR in order to work with regular RR theorems, the proof may become longer instead of shorter.

> I'm a proponent of RR>=0 (displayed \mathbb{R}_{\geq 0}) and RR>0 (displayed \mathbb{R}_{>0}) to avoid confusion as to RR+ containing 0 or not.

I could go with this, although I don't see how RR+ is ambiguous.  In the context of real numbers, "positive" always means nonzero, right?

DyeKuu:
> I could go with this, although I don't see how RR+ is ambiguous. In the context of real numbers, "positive" always means nonzero, right?

This is a very interesting point of divergence in French Maths education (I think both Benoit and I have gone through this way), where people tend to call "0<_ A" positive and "0 < A" strictly positive.

benjub:
> Before RR+ was added, there were enough theorems with (A e. RR /\ 0 < A) to make it apparent there would be a net benefit by introducing the RR+ definition and associated utility theorems like ~rpre. Currently there are over 450 theorems using RR+.
> 
> RR>=0 is used less frequently, but there are over 60 matches for ( A e. RR /\ 0 <_ A ) so it may be at the point of being useful.

As for the number of uses, note that after NN0 was introduced, several theorems using NN were restated (or at least their NN0 version was added).  In textbooks, I saw many places where theorems were stated in the (strictly) positive case but were true when, says, n=0 (or x=0 for reals).   A similar thing would probably happen for RR>=0.

> We have to be careful, though, to look at how these theorems are typically used. If a proof requires unwinding the definition back to RR in order to work with regular RR theorems, the proof may become longer instead of shorter.

One could say "a new definition typically makes a proof longer".  Consider the theorem "a compact metric space is complete". Its proof is longer than the proof of "a set with a function d such that .... and such that for any open cover one can extract a finite subcover, is such that every sequence such that... converges."  A proof typically begins with unravelling the definitions, then doing the "real proof" and at the end repackaging things into defined terms.  But you know that and I think I understand what you mean: there can be intermediate "helper theorems" which actually shortcut this, so we have to look if the necessary theorems are there.

> I could go with this, although I don't see how RR+ is ambiguous. In the context of real numbers, "positive" always means nonzero, right?

You're right that "positive" always implies nonzero AFAIK (with the remark of @DyeKuu that French speakers should be careful; but Metamath is in English anyway).  I was talking about the ambiguity of the symbol "RR+", which may mean "positive" or "nonnegative".

benjub:
A few links in favor of R_{>0} and R_{\geq 0}:
* https://en.wikipedia.org/wiki/Positive_real_numbers
* https://math.stackexchange.com/questions/27968/how-does-one-denote-the-set-of-all-positive-real-numbers (read the most voted answer)
* https://math.stackexchange.com/questions/1366348/symbol-for-set-of-strictly-positive-real-numbers#1366352

sctfn:
I tend to just use ( 0 [,) +oo ) when needed. It’s a compact expression that doesn’t require a new definition.

PR #2069, opened 2021-07-06
mathbox: bj-axc11rv, axc11n11

benjub:
* bj-axc11rv as announced in https://github.com/metamath/set.mm/pull/2064#issuecomment-874939383
* axc11n11: as given by @nmegill , so "Contributed by NM" but in my mathbox. I hope it's ok; happy to move it to Main or to @nmegill's mathbox.

@wlammen, @nmegill : I propose to move bj-axc11rv to Main and use it to prove axc16g (since we try to prove as many theorems as possible with only ax12v and not ax-12). Is it ok with you ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18992,6 +18992,7 @@ Proof modification of "axac3" is discouraged (74 steps).
 Proof modification of "axc10" is discouraged (37 steps).
 Proof modification of "axc11-o" is discouraged (47 steps).
 Proof modification of "axc11n-16" is discouraged (154 steps).
+Proof modification of "axc11n11" is discouraged (23 steps).
 Proof modification of "axc11nALT" is discouraged (62 steps).
 Proof modification of "axc11nOLD" is discouraged (64 steps).
 Proof modification of "axc11nOLDOLD" is discouraged (63 steps).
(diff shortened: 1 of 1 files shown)


nmegill:
> * bj-axc11rv as announced in [Reduce axiom usage in axc16nf #2064 (comment)](https://github.com/metamath/set.mm/pull/2064#issuecomment-874939383)
> * axc11n11: as given by @nmegill , so "Contributed by NM" but in my mathbox. I hope it's ok; happy to move it to Main or to @nmegill's mathbox.

It's fine in your mathbox.  We intentionally have no rule against that so that people could collaborate on a project before it is moved to main.

> 
> @wlammen, @nmegill : I propose to move bj-axc11rv to Main and use it to prove axc16g (since we try to prove as many theorems as possible with only ax12v and not ax-12). Is it ok with you ?

This is ok with me.

wlammen:
The choice is here to prove axc16g either from axc11r (current version), or alternatively from ax12v (your version). Both theorems are, more or less, direct consequences of ax-12, that usually describe different aspects of that axiom. In case of A. x x = y with distinct x, y both aspects converge, as you show, which is a nice observation. Are these theorems equivalent? axc11r is provable from ax12v, but the reverse is, I guess, not true (because ax-12 commands equality to be the finest grained operation in logic, something I cannot see in axc11r).  So axc11rv should appear in main to clarify this, and all ax16* theorems should base on it, IMO.

benjub:
> axc11r is provable from ax12v

You probably meant ax-12 ?

> So axc11rv should appear in main to clarify this, and all ax16* theorems should base on it, IMO.

Will do.

wlammen:
I meant: axc11r is provable from ax12v under the assumption A. x x = y dv(x, y), but the reverse not.

PR #2070, opened 2021-07-07
axc11rv, axc11v; mathbox: axc11n11r

benjub:
* move axc11rv to Main, as agreed in #2069 ; use it instead of axc11r to prove axc16g (and edit comment of axc16g); move axc11r right before ax-13 to make it clear that it is not used to prove theorems in the ax-12-section. Also add axc11v (note that here, one also avoids use of ax-13 since aecom with a dv condition is an instance of aevlem, so does not require ax-13).
* mathbox: add axc11n11r with motivational comments (might be of interest to @nmegill).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14541,6 +14541,7 @@ New usage of "bitsfzolemOLD" is discouraged (0 uses).
 New usage of "bj-0" is discouraged (1 uses).
 New usage of "bj-1" is discouraged (0 uses).
 New usage of "bj-ax12iOLD" is discouraged (0 uses).
+New usage of "bj-ax12v3ALT" is discouraged (0 uses).
 New usage of "bj-ax6e" is discouraged (0 uses).
 New usage of "bj-axc4" is discouraged (0 uses).
 New usage of "bj-axd2d" is discouraged (0 uses).
@@ -19107,6 +19108,7 @@ Proof modification of "bj-aecomsv" is discouraged (16 steps).
 Proof modification of "bj-alcomexcom" is discouraged (45 steps).
 Proof modification of "bj-ax12iOLD" is discouraged (20 steps).
 Proof modification of "bj-ax12v" is discouraged (35 steps).
+Proof modification of "bj-ax12v3ALT" is discouraged (44 steps).
 Proof modification of "bj-ax6e" is discouraged (43 steps).
 Proof modification of "bj-ax6elem1" is discouraged (27 steps).
 Proof modification of "bj-ax6elem2" is discouraged (23 steps).
@@ -19119,6 +19121,7 @@ Proof modification of "bj-axc11nv" is discouraged (66 steps).
 Proof modification of "bj-axc11v" is discouraged (14 steps).
 Proof modification of "bj-axc15v" is discouraged (103 steps).
 Proof modification of "bj-axc16b" is discouraged (14 steps).
+Proof modification of "bj-axc16g16" is discouraged (25 steps).
 Proof modification of "bj-axc4" is discouraged (15 steps).
 Proof modification of "bj-axd2d" is discouraged (16 steps).
 Proof modification of "bj-axdd2" is discouraged (28 steps).
--- a/discouraged
+++ b/discouraged
@@ -16141,6 +16141,7 @@ New usage of "ex-natded9.20" is discouraged (0 uses).
 New usage of "ex-natded9.20-2" is discouraged (0 uses).
 New usage of "ex-natded9.26" is discouraged (0 uses).
 New usage of "ex-natded9.26-2" is discouraged (0 uses).
+New usage of "exanOLD" is discouraged (0 uses).
 New usage of "exatleN" is discouraged (1 uses).
 New usage of "exbiOLD" is discouraged (0 uses).
 New usage of "exbidhOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@nmegill Your ax13fromc9 (in your mathbox) can be shortened (203->120 bytes) by using ax13b and twice ax-c5, which streamlines it. May I do the change within your mathbox "silently" (i.e. no credit comment) ?

    ( weq wn wal wi ax-c5 con3i ax-c9 syl2im ax13b mpbir ) ABDZEZBCDZPAFZGZGOACD
    ZEZRGGONAFZETSAFZERUANNAHIUBSSAHIBCAJKQABCLM $.
---------The proof of "ax13fromc9" (120 bytes) ends above this line.

 75       con3i.a=ax-c5    $a |- ( A. x x = y -> x = y )
 76     syl2im.1=con3i   $p |- ( -. x = y -> -. A. x x = y )
 89       con3i.a=ax-c5    $a |- ( A. x x = z -> x = z )
 90     syl2im.2=con3i   $p |- ( -. x = z -> -. A. x x = z )
 94     syl2im.3=ax-c9    $a |- ( -. A. x x = y -> ( -. A. x x = z -> ( y = z -> A. x y = z ) ) )
 95   mpbir.min=syl2im $p |- ( -. x = y -> ( -. x = z -> ( y = z -> A. x y = z ) ) )
104   mpbir.maj=ax13b  $p |- ( ( -. x = y -> ( y = z -> A. x y = z ) ) <-> ( -. x = y -> ( -. x = z -> ( y = z -> A. x y = z ) ) ) )
105 ax13fromc9=mpbir $p |- ( -. x = y -> ( y = z -> A. x y = z ) )

nmegill:
Go ahead.  It doesn't matter to me if you want to include a proof shortened acknowledgment - up to you.  BTW this material used to be in main but I moved it back to my mathbox when we did an axiom cleanup in main, but it can be treated as public since I plan no more changes to it myself.  The main thing to keep in mind is to take care that proofs do not make use of the new axioms, except in cases where the new axioms have already been proved from the old, as described in the ax13fromc9 comment.

While you are in there:  I want to change the labels "ax4" to "ax4fromc4",  "ax10" to "ax10fromc7", and "ax12o" to "ax12fromc15", to be consistent with the other ax\*from\* labels in mmset.html Appendix 8 where the main axiom being replaced is stated after the "from".  (We will lose the automatic check that ax-4 = ax4 and ax-10 = ax10 in 'verify markup' but I think consistency and descriptiveness is more useful in these special cases.)  Could you do that, and also update mmset.raw.html with the changes?  Thanks.

benjub:
@nmegill I'm doing it.  How to deal with section titles that are more than 79 characters long ?

> =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
>   Rederive new axioms from old: ax4fromc4, ax10fromc7, ax6fromc10, ax12fromc15, ax13fromc9
> =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

aecom-o claims it does not use ax-5 but it actually does (it is actually identical to axc11nfromc11 but its comment says "unlike axc11nfromc11...). Should I delete it ?

nmegill:
> @nmegill I'm doing it. How to deal with section titles that are more than 79 characters long ?

Right now you can't.  In order to keep the table of contents tidy (so people don't put essays into the titles), I'm not sure we want to allow it, but we can discuss that separately.  If more information needs to be stated, it can go under the header as usual.

> 
> > =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
> > Rederive new axioms from old: ax4fromc4, ax10fromc7, ax6fromc10, ax12fromc15, ax13fromc9
> > =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

I suggest changing it to:  "Rederive new axioms ax-4, ax-10, ax-6, ax-12, ax-13 from old"

> 
> aecom-o claims it does not use ax-5 but it actually does (it is actually identical to axc11nfromc11 but its comment says "unlike axc11nfromc11...). Should I delete it ?

It looks like the ax-5 (old ax-17) dependence happened sometime between 2001 and 2005, when the ax-5-free proof of equid was changed to a shorter one depending on ax-5.  equid1 still has a proof without ax-5 dependence.

More recently, references to ax-7 were changed to ax7, which also depends on ax-5.  So the proof could be recovered, but it's going to make the comment in ax-7 confusing (contrary to "the only theorem referencing ~ ax-7 should be ~ ax7v"), and I don't think it's worth it.  The original proof without ax-5 (a full condensed detachment proof directly from axioms) is in Lemma  L12 the article as stated in the comment.

I think we should delete axc11nfromc11 and change the aecom-o comment to say that an ax-5-free proof is given for Lemma L12 in the article.  Note that several later proofs depend on aecom-o.  If you feel it's more informative to change the label of aecom-o to axc11nfromc11, that's fine too.

Someday someone may want to go through the old axiom section and clean up these odds and ends, but it is only of obscure historical interest and low priority.

PR #2073, opened 2021-07-07
Shorten proofs of eqsbc3r, sbcimdv, sbcrext, rmob, and rmob2

jamesjer:
Continuing my study of proofs, I found a few more that can be shortened.  The shortening of rmob and rmob2 was mostly accomplished by `MINIMIZE_WITH` with only a small nudge from me, so no "Proof shortened by" comments for those two.  Neither the old nor the new proof of eqsbc3r requires that x and A be distinct, so I removed that dv condition.

I checked each old/new proof pair with "show trace_back <label> /axioms", and none of them changed.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16088,6 +16088,7 @@ New usage of "eqeqan12dALT" is discouraged (0 uses).
 New usage of "eqerOLD" is discouraged (0 uses).
 New usage of "eqid1" is discouraged (0 uses).
 New usage of "eqresr" is discouraged (4 uses).
+New usage of "eqsbc3rOLD" is discouraged (0 uses).
 New usage of "eqsbc3rVD" is discouraged (0 uses).
 New usage of "equequ2OLD" is discouraged (0 uses).
 New usage of "equid1" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
Thanks for checking axiom usage. Can you see if the removal of the DV condition propagates to the theorems recursively using it (and if yes, carry it, even in mathboxes) ?

jamesjer:
It does, but all 3 uses of it are by theorems marked "Proof modification discouraged".  Should I remove the DV conditions on those theorems anyway?

nmegill:
> I checked each old/new proof pair with "show trace_back /axioms", and none of them changed.

BTW I typically use 'show trace_back /essential /axioms /match a\*' or 'sh t/e/a/m a\*'. 
 '/essential' ignores syntax axioms, and the (optional) '/match a\*' ignores df-\* (definitions).  You can omit '/match a\*' when you want to know if new definitions were used or old ones dropped.

jamesjer:
Great, thank you for the tip.

nmegill:
> It does, but all 3 uses of it are by theorems marked "Proof modification discouraged". Should I remove the DV conditions on those theorems anyway?

You should go ahead and remove them.  There's hardly ever a reason to have unnecessary DVs (rare exceptions being theorems like ~ax6v that have the reasons clearly stated in their comments).  "Discouraged" doesn't necessarily mean "prohibited" but more "be sure any modification you make doesn't conflict with the author's original intent" e.g. to avoid the use of certain axioms or to demonstrate an alternate proof.  Anyway, removing an unnecessary DV doesn't modify the proof.

benjub:
For fast checking, I type "MM>sh tr xxx /ma ax-*" (taking advantage of the syntax used for axiom labels), but for full information, I'm happy that @jamesjer checks all $a-statements (for instance, in reu8 in #2067 he removed use of df-clab, which is nice to know).

PR #2075, opened 2021-07-08
Lemmas for knoppndv. ( ZZ>= ` 1 ) ==> NN.

acipsen:


benjub:
It looks like N e. RR suffices for the first few lemmas (and maybe most of them?) and also shortens their proofs. Maybe introduce the restriction N e. NN (or NN0) only when/if needed ?

acipsen:
You are, of course, right that N e. RR suffices for some (but not all) of the lemmas. I don't really see these lemmas having any use outside the specific theorem (that there exists continuous nowhere differentiable functions). So, for reasons of readability, I would prefer to keep the "type" of N fixed throughout. Regards, Asger.

PR #2083, opened 2021-07-11
K4 - closure is idempotent - equivalencies

arpie-steele:
Continue developing equivalencies between the analogues of neighborhoods, closures, interiors and convergents.
A topology satisfies all 5 of K0, K1, K2, K3, and K4. 
Still to do:
- K5 - closure commutes with arbitrary union even for infinite sets.
- Prove K13 and B e. Fin implies K5
- KF - The neighborhoods of every point form a filter (probably equivalent to some combination of axioms already known)
- Explore if removing the requirements for a topology to be closed under finite intersection and arbitrary union maps elements of ~P ~P B to ( ~P B ^m ~P B ) in a consistent manner.


benjub:
@arpie-steele : rereading the section comment, I saw a few things:
* first line: "the one-to-one relation": I would even say "the involution"
* there is no "natural isomorphism between K and N"; rather, it is the natural isomorphism PPB^B >-->> PB^PB which permits to transfer back and forth properties of a closure function K \in PB^PB and a neighborhood function N \in PPB^B, which correspond to each other under that isomorphism.
* the term "convergent function" is not very good... I would say that each M(x) is the set of adherents of x, or the adherent system of x (analogous to the term "neighborhood system") and M is an "adherent function". ("convergent" makes us think of sequences, which is more restrictive)
* N(x) is called the neighborhood system of x. For instance, in your table, "neighborhoods are nonempty" should read "neighborhood systems are nonempty".
* For your row "correspondences", I think two lines are enough:
```
S e. N(X)           <-> X e. I(S) <-> ...
-. (B \ S ) e. N(X) <-> ....      <-> ...
```
with the "<->" signs centered in-between columns, like the "=" signs below them (that array is a great idea).
Some columns appear centered and others left-aligned.

PR #2084, opened 2021-07-11
mathbox: typo and bj-genan/bj-genr/bj-genl

benjub:
By the way, @nmegill, @wlammen: have you tried in the past to work with ~ bj-ax12v3 ? It is stronger than ax12v but it does not seem to allow to prove much more, and in particular, not ax-c11(n)...


wlammen:
I think, this sentence in the accompanying text to ax12v needs to be reworded. I was a bit sloppy when stating it:
"If for a moment we assume 𝑦 could be set to 𝑥, then, after elimination of the tautology 𝑥 = 𝑥, immediately we have 𝜑 → ∀𝑥𝜑 for all 𝜑 and 𝑥. Any relevant logic model prohibits this."
In fact, if you allow substitution of x for y, ax-5 is the result, which is NOT prohibited by a relevant logic model. It is an interesting observation, that ax-5 is derivable from ax-12. There seems to be some redundancy in the axiom system, which calls for reduction.
The way I translate ax12v2 into English is "equality must be under all circumstances the finest grained operation in logic" This is actually a commandment restricting the introduction of any future logical operation, in particular text based ones reflecting the notation of a logical expression. For those loving to minimize bundling, ax12v2 is hardly replaceable.
I am not sure what the extra value of bj-ax12v3 is, in particular if you derive it from ax12. I share your opinion, that axc11n is likely not within reach of this theorem.

benjub:
I should have read the comment of ax12v through!

You may rephrase that sentence beginning with "If for a moment" to something like the second sentence in bj-ax12v3.

Or you may also write:
> One can remove either of the two DV conditions, but not both. If you remove DV(x,y), then you get ax12v2 (which is directly provable from this version ax12v and earlier axioms), and if you remove DV(y,ph), then you get bj-ax12v3 (see comment there)". However, if you remove both, then you can substitute y for x and obtain x -> A. x ph without DV condition, which is not vallid.

(pinging @nmegill if he wants to chime in)

nmegill:
> In fact, if you allow substitution of x for y, ax-5 is the result, which is NOT prohibited by a relevant logic model. It is an interesting observation, that ax-5 is derivable from ax-12. There seems to be some redundancy in the axiom system, which calls for reduction.

Yes, ax-5 can be derived from bj-ax12v3, but ax-5 was used to derive bj-ax12v3 from ax-12, so ax-5 doesn't seem to be redundant w.r.t. ax-12.  I agree that in principle we could replace ax-5 + ax-12 with bj-ax12v3 + ax-c11 (and maybe bj-ax12v3 + ax-c11n, although that is less certain).

wlammen:
The idea is here to assume x = x as an axiom, so it need not be derived using ax-5, and then apply it to ax-12 to get ax-5 as a theorem.  I haven't worked the details out, though.

benjub:
> The idea is here to assume x = x as an axiom, so it need not be derived using ax-5, and then apply it to ax-12 to get ax-5 as a theorem. I haven't worked the details out, though.

Indeed, this is the content of bj-ax12v3's comment ("modal logic K" means {ax-mp, ax-gen, ax-1--ax-4}).

nmegill:
> The idea is here to assume x = x as an axiom, so it need not be derived using ax-5, and then apply it to ax-12 to get ax-5 as a theorem. I haven't worked the details out, though.

I must be misunderstanding.  You also need ax-5 to eliminate the A. y quantifier in ax-12.

BTW x = x is derived without ax-5 in equid1 from the old ax-c\* axioms.  I don't know if it can be done with the current axioms.

benjub:
> > The idea is here to assume x = x as an axiom, so it need not be derived using ax-5, and then apply it to ax-12 to get ax-5 as a theorem. I haven't worked the details out, though.
> 
> I must be misunderstanding. You also need ax-5 to eliminate the A. y quantifier in ax-12.

He (and I) meant bj-ax12v3 instead of ax-12.

> BTW x = x is derived without ax-5 in equid1 from the old ax-c* axioms. I don't know if it can be done with the current axioms.

Probably not.

wlammen:
Yes, I referred to ax12v* in my comment. I remember I was about to write "a variant of ax-12", but was distracted then, and later forgot about it. I am sorry for any inconvenience.

PR #2086, opened 2021-07-14
reduce axiom usage in charv

wlammen:
1. Fix an incorrect statement in ax12v;
2. Remove dependency on ax-10 in chvarv
3. delete a couple of outdated OLD theorems

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14018,10 +14018,6 @@ New usage of "19.41rg" is discouraged (3 uses).
 New usage of "19.41rgVD" is discouraged (0 uses).
 New usage of "19.43OLD" is discouraged (0 uses).
 New usage of "19.8aOLD" is discouraged (0 uses).
-New usage of "19.9OLD" is discouraged (0 uses).
-New usage of "19.9dOLD" is discouraged (0 uses).
-New usage of "19.9hOLD" is discouraged (0 uses).
-New usage of "19.9tOLD" is discouraged (0 uses).
 New usage of "1dimN" is discouraged (0 uses).
 New usage of "1div0" is discouraged (0 uses).
 New usage of "1div0apr" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
(This comment is not about this PR specifically but at least has to do with its axiom usage topic...)

@wlammen In 2013, you shortened the proof of con4i by 6 bytes.  Although this was a valid (and clever) shortening based on our rules at that time, the shortened proof uses 2 more axioms.  For that reason I think we should revert it to the original (ax-mp applied to con4=ax-3).  Could you do that sometime at your convenience?

I bring this up because I received an email from someone learning logic, and he was puzzled why our con4i proof required 27 earlier theorems, making it somewhat tedious to study in complete detail back to axioms, when (as he pointed out) we could have just applied ax-mp to ax-3.  So I explained to him our criterion for accepting shorter proofs.

Hopefully this situation is not that common.  In any case, the original proof of con4i used fewer axioms, so that should be sufficient reason for reverting it.

wlammen:
@nmegill I will revert the proof.

benjub:
I would keep @wlammen's version as *ALT since it's clever and also to avoid that in the future someone "refinds" @wlammen's shortening and later it has to be reverted again.

wlammen:
I am a bit opposed to that ALT version, since the shortening is not big, or gives otherwise a better insight into the 'mechanics' of the theorem.
I understand what the student complains about. If I remember right (after a decade), in the whole section beginning with negation, I deferred theorems often as long as possible to have as much knowledge as possible available, before implementing the proof. This led to a sort of 'staircase effect', where each theorem is based on its predecessor, with a long chain down to the axioms.  This has the positive effect, that really important theorems bubble up to the front, whereas a theorem like con4i with little influence on other theorems appears late. This can be seen as unpleasant, because con4i is separated from its brothers con4 and con4d, and proved by a different technique. In a sense, the section so gives the impression of a sloppy systematic to the unaware.
Of course, the systematic is completely defined by the metrics in use, all having their pluses and downsides:
1. Current metrics: Minimize axiom dependencies and then prefer short proofs;
2. Minimize the depth of the proof tree;
3. ...
Maybe metrics 2 is favoured by the student here, because it better supports her/his learning preference. There is nothing against it, except some decision scheme must be in place, and for historical reasons, it is version 1.

nmegill:
I agree with @wlammen that an ALT isn't necessary.  To prevent refinds, move con4i to immediately after con4.

PR #2087, opened 2021-07-15
Shorten proofs of 32 theorems

jamesjer:
I may have gone a bit overboard tonight.  If this is too big to digest in one lump, let me know and I can break it up into smaller pull requests.

Checking with SHOW TRACE_BACK /AXIOMS shows only two differences:
- rab0 gains wo, df-or, wne, df-ne, wral, df-ral, cin, df-in, wss, and df-ss
- funtpg loses w3o and df-3or

I expect that the rab0 change is not a problem, but let me know if it is and I can remove that change.

I found 2 theorems tonight that puzzle me.  First, dfss1 and sseqin2 are identical; in fact, the proof of the latter is simply the former.  Both seem to be used in set.mm.  Should one be removed?

Second, I don't understand rgenz.  It is rgen with an additional hypothesis, but an identical conclusion.  That doesn't seem useful.  It isn't used by any proof in set.mm.  The proof can be can be shortened to the following, which enables removing the dv condition, but I wondered if the theorem shouldn't be removed instead.

```
15     mpancom.1=ne0i    $p |- ( x e. A -> A =/= (/) )
16     mpancom.2=rgenz.1 $e |- ( ( A =/= (/) /\ x e. A ) -> ph )
17   rgen.1=mpancom    $p |- ( x e. A -> ph )
18 rgenz=rgen        $p |- A. x e. A ph
```

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16055,6 +16055,7 @@ New usage of "elreal" is discouraged (7 uses).
 New usage of "elreal2" is discouraged (3 uses).
 New usage of "elringchomALTV" is discouraged (1 uses).
 New usage of "elrngchomALTV" is discouraged (1 uses).
+New usage of "elsnxpOLD" is discouraged (0 uses).
 New usage of "elspancl" is discouraged (1 uses).
 New usage of "elspani" is discouraged (2 uses).
 New usage of "elspansn" is discouraged (6 uses).
@@ -16171,6 +16172,7 @@ New usage of "flddivrng" is discouraged (2 uses).
 New usage of "fldhmsubcALTV" is discouraged (0 uses).
 New usage of "fnexALT" is discouraged (0 uses).
 New usage of "fngid" is discouraged (0 uses).
+New usage of "foco2OLD" is discouraged (0 uses).
 New usage of "fourierdlem31OLD" is discouraged (0 uses).
 New usage of "fourierdlem42OLD" is discouraged (0 uses).
 New usage of "fsumshftdOLD" is discouraged (0 uses).
@@ -16179,6 +16181,7 @@ New usage of "ftalem5OLD" is discouraged (0 uses).
 New usage of "funadj" is discouraged (4 uses).
 New usage of "funcnvadj" is discouraged (1 uses).
 New usage of "funcnvmptOLD" is discouraged (0 uses).
+New usage of "funcnvqpOLD" is discouraged (0 uses).
 New usage of "funcringcsetcALTV" is discouraged (0 uses).
 New usage of "funcringcsetcALTV2" is discouraged (0 uses).
 New usage of "funcringcsetcALTV2lem1" is discouraged (4 uses).
@@ -16200,6 +16203,8 @@ New usage of "funcringcsetclem7ALTV" is discouraged (1 uses).
 New usage of "funcringcsetclem8ALTV" is discouraged (1 uses).
 New usage of "funcringcsetclem9ALTV" is discouraged (1 uses).
 New usage of "funcrngcsetcALT" is discouraged (0 uses).
+New usage of "funprgOLD" is discouraged (0 uses).
+New usage of "funtpgOLD" is discouraged (0 uses).
 New usage of "fvimacnvALT" is discouraged (0 uses).
 New usage of "fzo0ssnn0OLD" is discouraged (0 uses).
 New usage of "gen11" is discouraged (19 uses).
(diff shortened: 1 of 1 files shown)


benjub:
> I expect that the rab0 change is not a problem, but let me know if it is and I can remove that change.

That's probably ok, waiting for @nmegill's approval (the definitions other than df-clab, df-cleq, df-clel are innocuous), even if this puzzles me a bit.  Interestingly, the new dependencies come from abfal, while my version of abfal (bj-abfal) does not have them (except for df-or and df-en).  I'll try to have a closer look.

> I found 2 theorems tonight that puzzle me. First, dfss1 and sseqin2 are identical; in fact, the proof of the latter is simply the former. Both seem to be used in set.mm. Should one be removed?

I would merge them (moving sseqin2's comment to dfss1).

> Second, I don't understand rgenz. It is rgen with an additional hypothesis, but an identical conclusion. That doesn't seem useful. It isn't used by any proof in set.mm.

I would remove it.

nmegill:
> I may have gone a bit overboard tonight. If this is too big to digest in one lump, let me know and I can break it up into smaller pull requests.

It is fine.  Nice work.

> 
> Checking with SHOW TRACE_BACK /AXIOMS shows only two differences:
> 
> * rab0 gains wo, df-or, wne, df-ne, wral, df-ral, cin, df-in, wss, and df-ss
> * funtpg loses w3o and df-3or
> 
> I expect that the rab0 change is not a problem, but let me know if it is and I can remove that change.

I looked at it, and it shouldn't be a problem.

> 
> I found 2 theorems tonight that puzzle me. First, dfss1 and sseqin2 are identical; in fact, the proof of the latter is simply the former. Both seem to be used in set.mm. Should one be removed?

These should be one theorem.  sseqin2 seems a more meaningful name than dfss1.  Both are used heavily, so we can't just delete one without side effects.  We should move the sseqin2 comment to dfss1 as Benoit suggested.  Then "delete" the sseqin2 $p statement, i.e. rename it to  sseqin2OLD and add "Obsolete as of dd-mmm-yyyy. (New usage is discouraged.) (Proof modification discouraged.)" to its comment. Finally, rename all dfss1 to become sseqin2.  (For search and replace, note that every reference to "dfss1" in set.mm is preceded by space and followed by space or newline.)

Also add a line to the top of the label change list at the top of set.mm:
```
dd-mmm-yy dfss1     ---         deleted; use sseqin2
```

I think we should also delete dfss5.  This is a trivial manipulation of dfss1 which is used only 4 times, saving only 2 steps in each case.  If we really needed it frequently, it would be better to add a theorem that swaps equalities in general (something like eqcomd with biconditionals instead of implications) instead of being limited to ( B i^i A ) = A ).

To eliminate dfss5 from a proof, in MM-PA use "expand dfss5" then "save new_proof/compressed".  Once there are no more references ('show usage dfss5' shows no uses), rename dfss5 to dfss5OLD and add the notices to its comment.  Add the following label change list entry:
```
dd-mmm-yy dfss5     ---         deleted
```

There is no theorem sseqin1 because it would be identical to df-ss (whose name we cannot change without violating our convention for definition labels).  I have considered changing df-ss to be what is now called dfss2 in order to be closer to most textbook definitions; if that is done, then what is now df-ss would be called sseqin1.

> 
> Second, I don't understand rgenz. It is rgen with an additional hypothesis, but an identical conclusion. That doesn't seem useful. It isn't used by any proof in set.mm. The proof can be can be shortened to the following, which enables removing the dv condition, but I wondered if the theorem shouldn't be removed instead.

Yes, it should be removed.  (I can't remember what I intended it for, and I don't see any uses for it.)  Add to the list:
```
dd-mmm-yy rgenz     ---         deleted
```



> ```
> 15     mpancom.1=ne0i    $p |- ( x e. A -> A =/= (/) )
> 16     mpancom.2=rgenz.1 $e |- ( ( A =/= (/) /\ x e. A ) -> ph )
> 17   rgen.1=mpancom    $p |- ( x e. A -> ph )
> 18 rgenz=rgen        $p |- A. x e. A ph
> ```

benjub:
> There is no theorem sseqin1 because it would be identical to df-ss (whose name we cannot change without violating our convention for definition labels).

There is the possibility to add it as an alias, like con4 for ax-3, or alim for ax-4, which keeps the best of both worlds.  I would be in favor of it.

> I have considered changing df-ss to be what is now called dfss2 in order to be closer to most textbook definitions; if that is done, then what is now df-ss would be called sseqin1.

The current definitions try to avoid dummy variables (like df-nul, which is similarly not the textbook definition in general). So I think df-ss is fine, given that it mentions dfss2 as more traditional.

PR #2088, opened 2021-07-15
Add eqvf, eq0f, neq0f

benjub:
* Add neq0f and prove neq0 from it. Before this PR, it was proved from n0, itself from n0f, which had it (more or less) as an intermediate proof step, which was backward, and added unnecessary dependencies on df-ne --- this change was triggered by PR #2087, where a dependency on df-ne had strangely appeared).

* Similarly, add eq0f. Before this PR, eq0 was proved from neq0, which was backward. Again, streamlined proofs and fewer df-ne dependencies as a result. This unexpectedly shortened two proofs by around 100 bytes each.

* For consistency, add eqvf, and move eqvf and eqv several sections earlier, where they belong, right after the definition of V.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17449,6 +17449,7 @@ New usage of "mulpqf" is discouraged (5 uses).
 New usage of "mulpqnq" is discouraged (7 uses).
 New usage of "mulresr" is discouraged (4 uses).
 New usage of "mulsrpr" is discouraged (9 uses).
+New usage of "n0fOLD" is discouraged (0 uses).
 New usage of "naecoms-o" is discouraged (1 uses).
 New usage of "natded" is discouraged (0 uses).
 New usage of "nbgraopALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@nmegill 
Similarly, I would reverse the order of rabn0 and rabeq0 : first prove the equality case, and then the inequality case from it. This would give similar proofs, but again reduce usage of df-ne in several later theorems, and in my opinion streamline the proofs (for instance, currently, abeq0 relies on two negations in a row instead of zero). Same number of essential steps in both cases, although three more steps for my proposal. May I carry it out ?
(below, the current proofs are labeled *OLD)
```
rabeq0 $p |- ( { x e. A | ph } = (/) <-> A. x e. A -. ph ) $=
( cv wcel wa cab c0 wceq wn wal crab wral abfal df-rab eqeq1i raln 3bitr4i ) BDCEAFZBGZHISJBKABCLZHIAJBCMSBNUATHABCOPABCQR $.

rabn0 $p |- ( { x e. A | ph } =/= (/) <-> E. x e. A ph ) $=
( crab c0 wne wn wral wrex rabeq0 necon3abii dfrex2 bitr4i ) ABCDZEFAGBCHZGABCIONEABCJKABCLM $.

rabn0OLD $p |- ( { x e. A | ph } =/= (/) <-> E. x e. A ph ) $=
( cv wcel wa cab c0 wne wex crab wrex abn0 df-rab neeq1i df-rex 3bitr4i ) BDCEAFZBGZHIRBJABCKZHIABCLRBMTSHABCNOABCPQ $.

rabeq0OLD $p |- ( { x e. A | ph } = (/) <-> A. x e. A -. ph ) $=
( wn wral wrex crab c0 wceq ralnex rabn0OLD necon1bbii bitr2i ) ADBCEABCFZDABCGZHIABCJNOHABCKLM $.
```

nmegill:
> May I carry it out ?

Yes.

PR #2090, opened 2021-07-16
n0ii, imnang, raln, rabn0, rabeq0

benjub:
* rabn0/rabeq0 as agreed in #2088
* raln was used in my proof of rabn0 and permitted a few other shortenings, so I added it, and similarly for imnang (although I did not try to minimize all proofs with them because of time issues on longer proofs)
* adding n0ii was unexpectedly efficient.  There is always the tension between the two forms " A =/= B" and "-. A = B" and associated utility theorems. For instance, set.mm contains vn0: _V =/= (/) . Probably, adding also vneq0: -. _V = (/) would shorten a few later proofs (which currently use theorems like neeq* and necon*), but I will probably not add it since the addition may not be worth it.
* Maybe my recently added abtru and abfal should be relabeled abv and ab0 ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12280,6 +12280,7 @@
 "psubspi2N" is used by "pclfinN".
 "psubspi2N" is used by "pclfinclN".
 "rabex2OLD" is used by "rab2exOLD".
+"rabn0OLD" is used by "rabeq0OLD".
 "ramcl2lemOLD" is used by "ramcl2OLD".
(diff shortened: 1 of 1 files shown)


nmegill:
> * Maybe my recently added abtru and abfal should be relabeled abv and ab0 ?

Yes, those would be better names.

abfal is essentially a restatement of abn0 (from which it can be proved with a shorter proof).
Is there a reason to have both abfal and abn0?  abn0 is used 36 times, abfal is used 3.

PR #2092, opened 2021-07-17
abtru -> abv ; abfal->ab0 and shorten its proof

benjub:
As agreed, in #2090, relabel abtru -> abv ; abfal->ab0 (and shorten the latter proof).
I'd prefer to leave both ab0 and abn0 (they differ enough to have both IMO).

ab0 could be proved from abn0 in as many essential steps, and one fewer syntactic step, than the current proof, at the cost of using df-ne and related theorems, and seeming a bit backward (remark added in comment).

One can shorten ab0 and three other proofs by using bj-notalbii. Close to but not "paying for itself" in terms of length added/saved.


wlammen:
This comment addressed what seemed to be a rewrap issue, but since it is invalid, I deleted it.

PR #2093, opened 2021-07-17
prove cbvalv without ax-10

wlammen:


benjub:
@wlammen : Maybe the same is doable with cbvexv, cbval2v... ?

wlammen:
@benjub Not the same, but something backed by cbvalv (if I guess right).

PR #2096, opened 2021-07-18
shorten cbvalv

wlammen:
Clean up. The previous version of cbvalv used very elementary steps to point out where each axiom kicks in. This version optimizes on proof length again, as usual.
Prove cbval2v, cbvex2v, cbvaldva, cbvexdva without ax-10.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15006,6 +15006,7 @@ New usage of "cbval2vOLD" is discouraged (0 uses).
 New usage of "cbvaldvaOLD" is discouraged (0 uses).
 New usage of "cbvalvOLD" is discouraged (0 uses).
 New usage of "cbvex2vOLD" is discouraged (0 uses).
+New usage of "cbvexdvaOLD" is discouraged (0 uses).
 New usage of "cbvexsv" is discouraged (2 uses).
 New usage of "cbvexvOLD" is discouraged (0 uses).
 New usage of "ccat2s1fvwALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
Clean up. The previous version used very elementary steps to point out where each axiom gets in. This version minimizes on proof steps again.

benjub:
Is there a reason for the final "a" in cbvaldva and cbvexdva ?

wlammen:
An abbreviation for the and in the hypothesis?

nmegill:
> Is there a reason for the final "a" in cbvaldva and cbvexdva ?

As @wlammen said, it is due to the "and" in the hypothesis.  This is an old convention going back to at least 1997 to distinguish e.g. rexbidva (added in 1997; 326 uses) from the weaker rexbidv (added in 1994; 950 uses), although it has also been used for cases where only the "a" version exists.  For newer, lesser-used deductions (which rarely justify both "a" and non-"a" versions) we would usually omit the "a" (and even the "v" when a non-"v" version would have little use).  Later on we also tend to omit the "d" for new theorems in more advanced set theory that use Mario's deduction method.

nmegill:
In the cases of cbvaldva and cbvexdva, a non-a version would have the awkward hypothesis  |- ( ph -> ( x = y -> ( ps \<-\> ch ) ) ), which is unlikely to be useful.  For theorems of this form, I think we could remove the "a".  Maybe the "a" should be used only with theorems like  rexbidv/rexbidva with non-equivalent weak and strong forms.

PR #2102, opened 2021-07-19
shorten pair imp4a and imp4b

wlammen:
1. imp4a is part of a sequence imp, impd, imp4a, imp5a and so on, with an ever increasing number of antecedents. The proof techniques in imp4aOLD and imp5a rely on the availability of a syllogism pattern syl6, syl8, syl10 and so on. While this is a straightforward and intuitive proof scheme, the imp\*a sequence can be proven recursively independently of the syl\* pattern.  This is demonstrated in this pull request, that, as a side effect, shortens imp4a and imp4b in total (not necessarily individually).
2. @avekens I don't understand your remark in theorem jccil.  Could you please have a look at wl-jccil, and explain, why its proof is inferior to that in the main part? IMHO, wl-jccil appears shorter on the web page, and still has the same number of proof bytes.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16813,6 +16813,8 @@ New usage of "imasvalOLD" is discouraged (2 uses).
 New usage of "imbi12VD" is discouraged (0 uses).
 New usage of "imbi13" is discouraged (2 uses).
 New usage of "imbi13VD" is discouraged (0 uses).
+New usage of "imp4aOLD" is discouraged (0 uses).
+New usage of "imp4bOLD" is discouraged (0 uses).
 New usage of "impelOLD" is discouraged (0 uses).
 New usage of "impexpd" is discouraged (1 uses).
 New usage of "impexpdcom" is discouraged (0 uses).
@@ -18821,8 +18823,6 @@ New usage of "wl-com12" is discouraged (2 uses).
 New usage of "wl-con1i" is discouraged (2 uses).
 New usage of "wl-con4i" is discouraged (1 uses).
 New usage of "wl-embant" is discouraged (0 uses).
-New usage of "wl-embantALT" is discouraged (0 uses).
-New usage of "wl-embantd" is discouraged (0 uses).
 New usage of "wl-id" is discouraged (1 uses).
 New usage of "wl-imim2" is discouraged (1 uses).
 New usage of "wl-imim2i" is discouraged (4 uses).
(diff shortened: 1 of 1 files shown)


benjub:
It looks like ~jccil is 10 bytes long (total number of steps = length of the block of capital letters in compressed format) while ~wl-jccil is 11 bytes long ? Maybe one could add wl-jccil as jccilALT so that the comment of jccil can be more explicit by linking to the ALT version.

@nmegill : I would move @wlammen's ~wl-embant to Main as ~embanti (place it earlier than ~embantd), and also prove the closed form ~embant, for instance from ~imim2 and ~com13 (and maybe prove embanti from it, unless reordering theorems shortens the sum of proof lengths ?) They may not "pay for themselves" by shortening enough proofs (or maybe they will?) but they look natural enough to be included in that first {ax-1,ax-2}-subsection.

wlammen:
Regarding jccil: You're right. It seems, I am unable to count beyond 10, missing enough fingers for that.  OK, @avekens this is enough of an explanation.

avekens:
With show proof */norm, you get the exact length of the proof in bytes (52 vs. 55 bytes):
MM> sh pr wl-jccil/norm
Proof of "wl-jccil":
     wph wch wps wph wps wch jccir.1 jccir.2 syl jccir.1 jca $.
---------The proof of "wl-jccil" (55 bytes) ends above this line.
MM> sh pr jccil/norm
Proof of "jccil":
      wph wps wch wph wps wch jccir.1 jccir.2 jccir ancomd $.
---------The proof of "jccil" (52 bytes) ends above this line.

With show proof */lem/ren, you get the number of (not compressed) steps (4 vs. 5 steps):
MM> sh pr jccil/lem/ren
1 jccir.1        $e |- ( ph -> ps )
2 jccir.2        $e |- ( ps -> ch )
3 1,2 jccir      $p |- ( ph -> ( ps /\ ch ) )
4 3 ancomd       $p |- ( ph -> ( ch /\ ps ) )
MM> sh pr wl-jccil/lem/ren
1 jccir.1        $e |- ( ph -> ps )
2 jccir.2        $e |- ( ps -> ch )
3 1,2 syl        $p |- ( ph -> ch )
4 jccir.1        $e |- ( ph -> ps )
5 3,4 jca        $p |- ( ph -> ( ch /\ ps ) )

However, the compressed proof regarded as "string" is longer for jccil than for wl-jccil:

    jccil $p |- ( ph -> ( ch /\ ps ) ) $=
      ( jccir ancomd ) ABCABCDEFG $. // 27 characters including blanks)

   wl-jccil $p |- ( ph -> ( ch /\ ps ) ) $=
      ( syl jca ) ACBABCDEFDG $. // 23 characters including blanks)

but this depends on the chosen token names.

nmegill:
> @nmegill : I would move @wlammen's ~wl-embant to Main as ~embanti (place it earlier than ~embantd), and also prove the closed form ~embant, for instance from ~imim2 and ~com13 (and maybe prove embanti from it, unless reordering theorems shortens the sum of proof lengths ?) They may not "pay for themselves" by shortening enough proofs (or maybe they will?) but they look natural enough to be included in that first {ax-1,ax-2}-subsection.

OK.

wlammen:
@avekens Thanks for your explanations.  Hopefully they will lead me to more sensible comments.

PR #2104, opened 2021-07-20
shorten pair exp4a and exp4b

wlammen:
The same idea as in PR #2102 .


wlammen:
The database contains outdated OLD theorems like raleqdOLD, tagged as usual with an obsolete comment, but reside in a mathbox. Does the community care about these? Or is it the responsability of a mathbox owner to remove them?

nmegill:
If it is tagged with "obsolete" and a date, the it seems pretty clear the intent was to delete it after a year, so go ahead.  If it has an OLD suffix but no "obsolete" tag in the comment, it would be best to contact the mathbox owner.

PR #2109, opened 2021-07-23
Delete dfss1, dfss5, and rgenz.

jamesjer:
As discussed in https://github.com/metamath/set.mm/pull/2087.  I apologize for taking so long to get to this.


jamesjer:
I forgot to mention that of the 4 proofs that used dfss5 (ordtri2or3, difuncomp, diarnN, and fsovcnvlem), it was possible to merge the eqcom with another step (e.g., syl5eq[r] or syl6eq[r]) so that it doesn't appear in the final proof of the latter 3.

PR #2111, opened 2021-07-23
Shorten 18 proofs

jamesjer:
Proofs with significant changes: elpr2, eqoreldif, disjpr2, prnzg, diftpsn3, prss, eqsn, sneqrg, and preqsn.

Proofs where `MINIMIZE_WITH` just needed a little nudge: rabsnifsb, rabrsn, snssd, prssd, sneqr, preq12b, prel12, elpreqpr, and opid.

Also move prssg so it can be used in the proof of prss, and move sneqrg so it can be used in the proof of sneqr.

`SHOW TRACE_BACK /AXIOMS` shows no difference for any of them.  Note that the dv conditions on `prss` were not needed even before this change, so I removed them.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15688,6 +15688,7 @@ New usage of "dicfnN" is discouraged (1 uses).
 New usage of "dicvalrelN" is discouraged (0 uses).
 New usage of "difexOLD" is discouraged (0 uses).
 New usage of "difidALT" is discouraged (0 uses).
+New usage of "diftpsn3OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
> Note that the dv conditions on `prss` were not needed even before this change, so I removed them.

And same with prssg, apparently, since none of these two proofs use dummy variables.  Also, your new proof of sneqrg does away with dummy variables. Maybe others do as well, I'll have a look when they are on the webpages. Nice work !

PR #2116, opened 2021-07-27
Add eliuni and shorten 10 proofs

jamesjer:
I would like to suggest the addition of eliuni, one direction of eliun without quantifiers.  I have found 10 proofs that can be shortened with it.  In some cases, the obvious proof shortening gave `MINIMIZE_WITH` more to work with, leading to further size reductions.  This commit reduces the size of set.mm by 433 bytes, which I admit is not terribly impressive.  `SHOW TRACE_BACK` shows no changes in any of these theorems.

Investigating uses of eliun led to the discovery of 3 other theorems (fsuppmapnn0fiub, fsumcom2, and fprodcom2) where the use of eliun can be simplified, leading to shorter proofs, but I will submit those as a separate pull request.

I was able to remove a few dv conditions in conjunction with this work, but none of them seemed to allow dv condition removal for any consumers other than decpmataa0.

For the curious, I am working on recursive function theory, and after working on about the 4th proof where I only needed one direction of eliun, I started to wonder if proofs already in set.mm might be able to use such a theorem.  I hope to soon have that work in good enough shape to make creating a mathbox worthwhile.


benjub:
I presume you looked at the biconditional form of your eliuni and it was less useful ?

nmegill:
I agree that we can add eliuni.  It's more than just biimpi applied to eliun but also eliminates a quantifier, saving a number of steps in the cases you mention, and it seems it will eventually "pay for itself" i.e. result in a net reduction of set.mm size.  Currently eliun is used over 200 times, and I wonder how many of those eliuni could be used instead - how many have you looked at?

jamesjer:
Please excuse the delay in replying.  @benjub , I'm not sure what you mean.  Do you mean this?
```
    ( A e. D -> ( E e. C <-> E e. U_ x e. D B ) )
```
(which can be simplified with eleq2d, of course).  If that's what you mean, it isn't true.  Here is an assignment of variables that satisfies eliuni, but for which the above is false.
```
A = 1
B = { y | y = x }
C = { 1 }
D = { 0 , 1 }
E = 0
```

@nmegill I looked at all of them.  Most of the proofs use eliun in biconditional chains.  These 10 were the only ones I could find where there was an obvious way to substitute eliuni instead.

PR #2117, opened 2021-07-31
Fermat numbers

avekens:
Main set.mm:
* ~6p4e10b, ~9p1e10b moved from SP's mathbox - **questions**:
** should the related theorems ~5p5e10b, ~7p3e10b, ~8p2e10b also be moved to main set.mm, although they are not used in other mathboxes?
** Is there something special to consider if these theorem are moved to main set.mm (they were from section "N-Digit Addition Proof Generator"?
* ~muls1d, ~isprm7, ~ltsubnn0 moved from other mathboxes
* new theorems about decimal integers added: ~3decltc , ~decaddm10, ~decsubi, ~decmul1, ~decmulnc, ~11multnc, ~decmul10add, ~9t11e99, ~sq10, ~sq10e99m1, ~ 3dec
* new theorems about the molulo operation and the divides relation added: ~mulp1mod1, ~summodnegmod, ~nndivides, ~3dvdsdec, ~3dvds2dec, ~dvdsprmpweq, ~dvdsprmpweqle
* misc. theorems added: ~muladd11r, ~fzo1to4tp

AV's Mathbox:
* section "Fermat numbers" with a definition of Fermat numbers and related theorems added

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15002,8 +15002,13 @@ New usage of "cba" is discouraged (92 uses).
 New usage of "cbncms" is discouraged (6 uses).
 New usage of "cbv3hvOLD" is discouraged (0 uses).
 New usage of "cbv3hvOLDOLD" is discouraged (0 uses).
+New usage of "cbval2vOLD" is discouraged (0 uses).
+New usage of "cbvaldvaOLD" is discouraged (0 uses).
 New usage of "cbvalvOLD" is discouraged (0 uses).
+New usage of "cbvex2vOLD" is discouraged (0 uses).
+New usage of "cbvexdvaOLD" is discouraged (0 uses).
 New usage of "cbvexsv" is discouraged (2 uses).
+New usage of "cbvexvOLD" is discouraged (0 uses).
 New usage of "ccat2s1fvwALT" is discouraged (0 uses).
 New usage of "cdj1i" is discouraged (0 uses).
 New usage of "cdj3i" is discouraged (0 uses).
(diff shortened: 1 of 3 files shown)


nmegill:
> * **questions**:
>   ** should the related theorems ~5p5e10b, ~7p3e10b, ~8p2e10b also be moved to main set.mm, although they are not used in other mathboxes?

I think that would be ok.

Related question:  do we really need "10" at all, or should it be considered an obsolete relic of pre-Mario days?  Right now, "10" is used by only 45 $p statements, and 6p4e10/6p4e10b etc. are essentially duplicated theorems.  df-dec would become slightly more complex with "( 9 + 1 )" instead of "10", but things like 10p10e20 "( 10 + 10 ) = ; 2 0" seem like an awkward mixture of two notations.  In a few cases "10" makes a theorem slightly nicer to read, but two notations for it is one more thing to confuse newcomers.

digama0:
> Related question: do we really need "10" at all, or should it be considered an obsolete relic of pre-Mario days? 

One issue with eliminating "10" is that it is used in the definition of `; A B`; you wouldn't be able to use `; 1 0` there because it would be circular (and it would also be true in any base, so you wouldn't know what base is being described. As the joke goes, every base is base 10.)

EDIT: Using `9 + 1` there directly would probably be fine. There are a bunch of special cases in the numeric normalizer to deal with 10, so I would not be against eliminating it.

david-a-wheeler:
> Related question: do we really need "10" at all, or should it be considered an obsolete relic of pre-Mario days?

It's quite plausible to divide history into "pre-Mario days" & "post-Mario days" :-) :
https://www.youtube.com/watch?v=LVGSeDjWzUo&t=2m24s

PR #2124, opened 2021-08-04
mmset.raw.html - fix typos, validation errors, rewrite some text

nmegill:


benjub:
Thanks @nmegill and sorry for the HTML errors I apparently introduced.  I thought I had removed them all in #2081. As I wrote there, I had added some missing closing paragraph tags in mmset.raw.html. This caused the linked W3C validator to indicate errors. I did some research and saw that the closing tags were not necessary under certain conditions (though not closing tabs is considered bad practice, and may cause problems with some processing tools or complexify migration to HTML5 or other formats). In any case, I wonder why they caused errors... Maybe @david-a-wheeler or @jkingdon have an idea ?

Also, I had written \` A. x x = y \` at some places but as you explained, spacing considerations impose to write \` A. x \` \` x = y \`

david-a-wheeler:
Closing tags are optional in HTML5. They're required in XHTML, but XHTML is rarely used today. Adding closing tags is "fine but not important".

The W3C validator *should* be fine if there are closing tags. However, you have to make sure you close things consistently. It's also possible that the W3C validator has a bug :-).

jkingdon:
I vote for closing tags. Granted, the question of what the tools and/or standards actually require is rather more complicated than "always close them", but if you leave the tags to be implicitly closed for you, they might not be closed in the way you think which I suppose is probably what led to other validator failures. So the simplest way is "just close them".

Issue #2128, opened 2021-08-05
Duplicate theorems


savask:
In a recent pull request #2087 a pair of duplicate theorems was found, so I decided to run an automated tool to hunt for such examples. Because of the way my program works, the following list is definitely not comprehensive, but I hope it's of some use anyway. Here it is:

- [x] [0mbf](http://us.metamath.org/mpeuni/0mbf.html) in a mathbox of @Achaemenid and [mbf0](http://us.metamath.org/mpeuni/mbf0.html) in a mathbox of @glacode 
- [x] [issmfltle](http://us.metamath.org/mpeuni/issmfltle.html) and [smfpreimale](http://us.metamath.org/mpeuni/smfpreimale.html) both in a mathbox of @glacode
- [x] [mblvon](http://us.metamath.org/mpeuni/mblvon.html) and [vonval2](http://us.metamath.org/mpeuni/vonval2.html) both in a mathbox of @glacode
- [x] [rabeq2i](http://us.metamath.org/mpeuni/rabeq2i.html) in main by @nmegill and [rabid3](http://us.metamath.org/mpeuni/rabid3.html) in a mathbox of @glacode
- [x] [ge0nemnf2](http://us.metamath.org/mpeuni/ge0nemnf2.html) in a mathbox of @glacode and [xrge0neqmnf](http://us.metamath.org/mpeuni/xrge0neqmnf.html) in main by @tirix
- [x] [oddm1d2](http://us.metamath.org/mpeuni/oddm1d2.html) and [zoddb](http://us.metamath.org/mpeuni/zoddb.html) both in main by @avekens
- [x] [snunioo2](http://us.metamath.org/mpeuni/snunioo2.html) in a mathbox by @glacode and [ioounsn](http://us.metamath.org/mpeuni/ioounsn.html) in a mathbox by @TjWhale
- [x] [mptss](http://us.metamath.org/mpeuni/mptss.html) in a mathbox by @glacode and [fmptss](http://us.metamath.org/mpeuni/fmptss.html) in a mathbox by @tirix

In the following cases a newer version was moved to the main part, while the "original" stayed in a mathbox:
- [x] [complss](http://us.metamath.org/mpeuni/complss.html) in main by @benjub and [conss34](http://us.metamath.org/mpeuni/conss34.html) in a mathbox by Andrew Salmon
[...]

PR #2130, opened 2021-08-07
Make conss34 OLD and update the label of complss

savask:
Resolving one of the pairs in #2128 as suggested by @benjub, but instead of deleting conss34, I added OLD to it, and made it discouraged.

Also rewrapped the file, so the comment of clsk1independent changed a bit.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15390,6 +15390,7 @@ New usage of "con4iOLD" is discouraged (0 uses).
 New usage of "con5" is discouraged (1 uses).
 New usage of "con5VD" is discouraged (0 uses).
 New usage of "con5i" is discouraged (2 uses).
+New usage of "conss34OLD" is discouraged (0 uses).
 New usage of "conventions" is discouraged (0 uses).
 New usage of "conventions-label" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> Resolving one of the pairs in #2128 as suggested by @benjub, but instead of deleting conss34, I added OLD to it, and made it discouraged.

Good; this is exactly what I had in mind.  (For duplicate theorems whose proof didn't change, "Obsolete proof of..." could be "Obsolete version of..." if proof didn't change, but it doesn't really matter since it will be deleted in a year.)

> 
> Also rewrapped the file, so the comment of clsk1independent changed a bit.

Also good, but I am curious:  'write source ... /rewrap' isn't supposed to touch this because of the \<HTML\>.  Did you do this by hand?  (I want to make sure there's no bug in /rewrap.)

savask:
@nmegill
>Also good, but I am curious: 'write source ... /rewrap' isn't supposed to touch this because of the \<HTML\>. Did you do this by hand? (I want to make sure there's no bug in /rewrap.)

No, metamath program did that, but I used an old version.

PR #2131, opened 2021-08-07
Fermat numbers (continued)

avekens:
Main set.mm:
* ~dvdsnprmd, ~oddprmdvdsn2pw added

AV's Mathbox:
* Header comment for section "Fermat numbers" added
* Theorems ~pwdif, ~2pwp1prm2pw2pwp1prm and ~ 2pwp1prmfmtno added, showing that every prime number of the form ` ( ( 2 ^ k ) + 1 ) ` must be a Fermat number.


nmegill:
I like your Fermat number work.  I have a couple of comments on some of the labels.

The label ~oddprmdvdsn2pw is pushing the limit of our label conventions.  In particular, "n2pw" seems a little too specific to have as part of the label, and ~oddprmdvds may be adequate.  Labels are supposed to provide a hint and not necessarily a complete description.   From ~conventions-label: "Labels are not a 1-to-1 mapping, because that would create long names that would be difficult to remember and tedious to type. Instead, label names are relatively short while suggesting their purpose."

In your mathbox, ~2pwp1prm2pw2pwp1prm isn't something I think I could easily remember.  We would probably want to change it to something like ~prmfmtnopw2 if it is moved to main.

PR #2147, opened 2021-08-13
Shorten 10 proofs & correct a typo

jamesjer:
Only 2 are worth a "proof shortened by" comment, I think.  `SHOW TRACE_BACK` shows only 1 change: wral and df-ral are added to elxp2.  Notes:

- 0nelelxp stays the same size, 170 bytes, but the number of steps is reduced from 136 to 113.  Is that worth doing?  If not, I can remove that change from the commit.
- The xpss change is just dv condition removal, removal of the curly braces, and reindenting.  I assume that the proof of xpss12 used to be included in the proof.  Like above, if you don't want this done, I can remove it from the commit.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14010,6 +14010,7 @@ New usage of "0lno" is discouraged (3 uses).
 New usage of "0lnop" is discouraged (6 uses).
 New usage of "0lt1sr" is discouraged (2 uses).
 New usage of "0ncn" is discouraged (3 uses).
+New usage of "0nelxpOLD" is discouraged (0 uses).
 New usage of "0ngrp" is discouraged (2 uses).
 New usage of "0nnq" is discouraged (22 uses).
 New usage of "0npi" is discouraged (9 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> * 0nelelxp stays the same size, 170 bytes, but the number of steps is reduced from 136 to 113.  Is that worth doing?  If not, I can remove that change from the commit.

This is fine.  It would be interesting to see if the web page is smaller, even though it's not something we usually check for.

> * The xpss change is just dv condition removal, removal of the curly braces, and reindenting.  I assume that the proof of xpss12 used to be included in the proof.  Like above, if you don't want this done, I can remove it from the commit.

This is fine.

nmegill:
> > * 0nelelxp stays the same size, 170 bytes, but the number of steps is reduced from 136 to 113.  Is that worth doing?  If not, I can remove that change from the commit.
> 
> This is fine. It would be interesting to see if the web page is smaller, even though it's not something we usually check for.

mpeuni/0nelelxp.html decreased from 16374 to 15797 bytes.
mpegif/0nelelxp.html decreased from 23941 to 22403 bytes.
Not significant in the big picture, of course, but something...

benjub:
* The reduction in the number of syntactic steps will generally imply that the proof appearing on the webpage contains fewer symbols, which decreases the size of the webpage, but more importantly makes the proof probably easier to read.
*  Removal of dv conditions, especially useless ones as in xpss, is always good IMHO, even more when it removes a level of curly braces nesting.

PR #2149, opened 2021-08-15
set.mm - eliminate definitional hypotheses on rpnnen1

nmegill:
~rpnnen1 was renamed to ~rpnnen1lem6, and a new ~rpnnen1 was proved with definitional hypotheses eliminated.  This saves having to eliminate them twice in ~reexALT and  ~rpnnen, saving about 400 bytes per proof.


benjub:
The same can be done with the hypothesis of ~rpennen2, used twice (and ~eqid suffices to prove the hypothesis since there is no DV condition). Can you do it @nmegill ?

nmegill:
> The same can be done with the hypothesis of ~rpennen2, used twice (and ~eqid suffices to prove the hypothesis since there is no DV condition). Can you do it @nmegill ?

OK, it is done.

PR #2161, opened 2021-08-25
Finish section "Limits" in iset.mm

jkingdon:
A lot of this is pretty straightforward given the now-familiar differences between iset.mm and set.mm over `seq` and similar topics.

Includes climcvg1n which represents taking previous "a Cauchy sequence converges' results in iset.mm and updating to (a) use the absolute value notation, (b) use the `~~>` notation for convergence, (c) extend from real numbers to complex numbers.

Includes a number of results related to Cauchy sequences (even in set.mm there are a few different expressions and theorems to convert between them - in iset.mm there is the additional complication that "Cauchy sequence" sometimes means the climcvg1n-style concept which includes a rate of convergence, and sometimes just means the set.mm expressions).

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -20,6 +20,7 @@
 "4syl" is used by "fzssp1".
 "4syl" is used by "isose".
 "4syl" is used by "isoselem".
+"4syl" is used by "serif0".
 "4syl" is used by "smoiso".
 "4syl" is used by "tposss".
 "a9evsep" is used by "ax9vsep".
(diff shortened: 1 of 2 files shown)


nmegill:
The iset-discouraged check fails.
-New usage of "4syl" is discouraged (11 uses).
+New usage of "4syl" is discouraged (12 uses).

jkingdon:
> The iset-discouraged check fails.
> -New usage of "4syl" is discouraged (11 uses).
> +New usage of "4syl" is discouraged (12 uses).

Fixed now.

PR #2163, opened 2021-08-25
Shorten 16 proofs

jamesjer:
`SHOW TRACE_BACK` shows no difference for any of them.  I added "proof shortened by" comments to 3.  Twelve others were just minor proof step rearrangements to give MINIMIZE_WITH some traction.  The last one is elsnxp, which I already shortened a month ago.  On looking at it again, I spotted 2 more ways the proof could be shortened.  Since it already has a "proof shortened by" comment from me, I just left it alone.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18235,6 +18235,9 @@ New usage of "renegclALT" is discouraged (0 uses).
 New usage of "renicax" is discouraged (0 uses).
 New usage of "resfunexgALT" is discouraged (0 uses).
 New usage of "resgrprn" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


nmegill:
> The last one is elsnxp, which I already shortened a month ago. On looking at it again, I spotted 2 more ways the proof could be shortened. Since it already has a "proof shortened by" comment from me, I just left it alone.

If you want, you can just update your previous "Proof shortened by" date.  That will make it appear near the top of the Most Recent Proofs page to let people know there was a change.

jamesjer:
The most recent changes are minor enough that I'm okay with leaving it as it is.

PR #2165, opened 2021-08-26
There exists a continuous nowhere differentiable function

acipsen:


benjub:
Congratulations !
You can state the final result as
|- E. f e. ( RR -cn-> RR ) dom ( RR _D f ) = (/)
instead of
|- E. f ( f e. ( RR -cn-> RR ) /\ dom ( RR _D f ) = (/) )

acipsen:
Thanks! Right, that is more elegant.

PR #2170, opened 2021-08-27
Shorten 10 proofs

jamesjer:
While studying proofs, I have come across adjacent syl5eq and syl6eq steps often enough that it stuck in my memory.  I recall being irritated that neither 3eqtr3g nor 3eqtr4g would do, since both required an eqcomi on one branch.  In the absence of 3eqtr1g or 3eqtr2g (I don't understand the numbering system, so I'm not sure which), I decided to try using either 3eqtr3g or 3eqtr4g anyway to see if proofs got any shorter.  Some did, some didn't.  These are the ones that became shorter (by as little as 1 byte!), although I've only gone back over proofs I've already studied.  There may be more opportunities elsewhere in set.mm.

I was able reduce the length of csbprc further.  The other changes just replace syl5eq + syl6eq with 3eqtr3g or 3eqtr4g, then adapt to the "backwards" equality in one branch.  Note that disjssun is the same number of bytes, but fewer steps.

There are no changes in axiom or definition usage.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15269,6 +15269,7 @@ New usage of "csbima12gOLD" is discouraged (2 uses).
 New usage of "csbingOLD" is discouraged (4 uses).
 New usage of "csbingVD" is discouraged (0 uses).
 New usage of "csbopabgALT" is discouraged (1 uses).
+New usage of "csbprcOLD" is discouraged (0 uses).
 New usage of "csbresgOLD" is discouraged (2 uses).
 New usage of "csbresgVD" is discouraged (0 uses).
 New usage of "csbrngOLD" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@jamesjer The situation you describe suggests that there may be a useful lemma combining syl5eq/syl6eq or related 3eqtr*g that would justify itself by the proof shortenings it allows ?

jamesjer:
Sorry to take so long to respond to this @benjub.  I have done some searching and cannot come up with a sufficient number of uses to justify a theorem.  It would save only a couple of bytes per proof, so a lot of proofs would be needed to make it worthwhile.  Plus, in many proofs, there is an opportunity elsewhere in the proof to reverse the order of the terms in an equality without making the proof longer, which means that 3eqtr3g or 3eqtr4g can be used.  I think I will leave things as they are.

david-a-wheeler:
@jamesjer - no problems. Thanks for doing the investigation!

Issue #2167, opened 2021-08-27
Add summation to iset.mm


jkingdon:
The summation syntax - http://us.metamath.org/mpeuni/df-sum.html - in set.mm is kind of an amazing thing. It handles finite sets and infinite sequences which can be indexed not just in the obvious ways (NN, NN0, etc) but also "all even indexes" or other subsets of an upper integer set. Furthermore, it has two halves which can overlap (for example in integer ranges like `3...5`) but which agree where they overlap, making iota still work.

How does all this work in iset.mm?

* The `if(𝑛 ∈ 𝐴, ⦋𝑛 / 𝑘⦌𝐵, 0)` portion of the definition will only work well if `𝑛 ∈ 𝐴` is decidable. I expect various theorems will need a hypothesis like `( ( ph /\ x e. ZZ ) -> DECID x e. A`. Fortunately, this should be satisfied by a lot of the usual suspects like `NN` and other upper integer sets, integer ranges, and maybe others. But it would not be satisfied by pathological subsets of `(ℤ≥‘𝑚)` analogous to the ones in, for example, http://us.metamath.org/ileuni/onsucelsucexmidlem.html .
* A theorem like http://us.metamath.org/mpeuni/fsum.html should presumably not need a new hypothesis since it already has narrowed things down to a case that works (Update 10 Sep 2022: It does, however, need some changes for seq including defining `G` beyond `M`)
* The first disjunct in http://us.metamath.org/mpeuni/sumz.html is an example of the "need a DECID hypothesis" sort. In a nutshell, "summable set" would be expressed a different way whereever it occurs.
* ~~As far as I could tell, there is no need to change `df-sum` itself (beyond the obvious change of adding `CC` as an additional argument to `seq`)~~ (Update: https://us.metamath.org/ileuni/df-isum.html does have some modifications from when this issue was created but perhaps has everything it needs as of 10 Sep 2022).

Because this is a fairly large section and even the parts which don't require a rethink are likely to involve some fiddling (syntax changes and the like), it may be a while if I find there are problems with the approach outlined above, but that is my thinking so far. ~~My work to date is at https://github.com/jkingdon/set.mm/tree/iset-sum~~ _That branch was merged at #2184 but further changes will be needed before summation can be considered "done"_

digama0:
[...]

PR #2173, opened 2021-08-31
Add cardinality to iset.mm

jkingdon:
So I always knew cardinality was going to be tough in iset.mm, because:

1. even in ZF as opposed to ZFC various things are different (as well documented in set.mm). And of course we can't have the full axiom of choice due to http://us.metamath.org/ileuni/acexmid.html
2. there are a bunch of related things iset.mm can't have - for example http://us.metamath.org/mpeuni/sbth.html implies excluded middle (this was proven in 2019)
3. another example of something relevant which iset.mm can't have is http://us.metamath.org/mpeuni/onint.html - this pull request includes a proof (informally given by @digama0 , formalized by me) that this implies excluded middle.

So why bother at all? Well my personal motive is basically just to get some theorems about the sizes of finite sets (because a lot of the work at #2167 is likely to benefit from it), which seems like a more modest goal than to develop cardinality in general very fully.

How is it going? Well, there seem to be a lot of things which don't work and will only work with substantial effort (if at all). But this pull request at least gets things started. And hopefully identifies some areas for future work.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2046,8 +2046,8 @@ fair bit of intuitionizing.</TD>
 
 <TR>
 <TD>onint</TD>
-<TD><I>none</I></TD>
-<TD>Conjectured to not be provable without excluded middle. If you apply onint to a pair you can derive totality of the order.</TD>
+<TD>~ onintexmid</TD>
+<TD>Implies excluded middle as shown in onintexmid.</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2730,6 +2730,13 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>cardf2</TD>
+  <TD>~ cardcl</TD>
+  <TD>It would also be easy to prove ` Fun card ` or
+  ` { x | E. y e. On y ~~ x } C_ dom card ` if there is a need.</TD>
+</TR>
+
+<TR>
 <TD>addcompi</TD>
 <TD>~ addcompig </TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2730,7 +2730,7 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
-  <TD>cardf2</TD>
+  <TD>cardf2 , cardon</TD>
   <TD>~ cardcl</TD>
   <TD>It would also be easy to prove ` Fun card ` or
   ` { x | E. y e. On y ~~ x } C_ dom card ` if there is a need.</TD>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -37662,6 +37709,41 @@ class of all ordinal numbers to be a set (so instead it is a proper
...
+  $}
+
+  ${
+    $d ph y z $.  $d x y z $.
+    onintexmid.onint $e |- ( ( { y , z } C_ On /\ E. x x e. { y , z } ) ->
+      |^| { y , z } e. { y , z } ) $.
+    $( If the intersection (infimum) of an inhabited class of ordinal numbers
+       belongs to the class, excluded middle follows.  The hypothesis is an
+       instance of ` ( ( A C_ On /\ E. x x e. A ) -> |^| A e. A ) ` which we
+       would be able to prove given encluded middle.  (Contributed by Mario
```
bprzybocki, line 37736: "encluded" -> "excluded"

jkingdon: Thanks! I've pushed a fix

iset.mm
```diff
@@ -37178,6 +37202,31 @@ is a way of connecting statements about ordinals (such as trichotomy in
       CVD $.
   $}
 
+  ${
+    $( Verison of ~ 2on with the definition of ` 2o ` expanded and expressed in
```
avekens, line 37206: Verison -> Version

iset.mm
```diff
@@ -59437,6 +59519,119 @@ the cancellation property (fifth hypothesis), show that the relation
...
+    $d x y z A $.
+    $( The value of the cardinal number function with an ordinal number as its
+       argument.  (Contributed by NM, 24-Nov-2003.)  (Revised by Mario
+       Carneiro, 13-Sep-2013.) $)
+    oncardval $p |- ( A e. On -> ( card ` A ) = |^| { x e. On | x ~~ A } ) $=
+      ( vy con0 wcel cv cen wbr wrex ccrd cfv crab cint wceq enrefg breq1 mpdan
+      rspcev cardval3ex syl ) BDEZCFZBGHZCDIZBJKAFBGHADLMNUABBGHZUDBDOUCUECBDUB
+      BBGPRQCABST $.
+
+    $( The cardinal of an ordinal number is less than or equal to the ordinal
```
avekens, line 59607: The cardinal -> The cardinality ?

jkingdon: This language is in set.mm and there is no reason to diverge.

I suspect "the cardinal" more emphasizes the specific set which represents the result of `card` and "cardinality" more represents the concept. But I'd be glad to defer to someone else if they want to come up with some kind of ruling on this terminology.

iset.mm
jkingdon: Thanks! I've pushed up a fix.

iset.mm
avekens: OK, no change necessary.

PR #2182, opened 2021-09-11
Develop an alternative definition of "not free"

wlammen:
1. nf2 is a better definition of 'not free' as it over all reduces axiom usage. Demonstrate this in a couple of instances in my Mathbox.
2. delete some outdated and unused OLD theorems

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -9026,7 +9026,6 @@
 "infmrclOLD" is used by "minvecolem6OLD".
 "infmrclOLD" is used by "pilem2OLD".
 "infmrclOLD" is used by "pilem3OLD".
-"infmrclOLD" is used by "supminfOLD".
 "infmrgelbOLD" is used by "infmrgelbiOLD".
 "infmrgelbOLD" is used by "infmxrreOLD".
 "infmrgelbOLD" is used by "minveclem2OLD".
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
I'm fine with adding it to your mathbox.

Once it's added, do you want to raise the idea of switching to it? If so, that's a big enough change that it's probably worth a discussion on the mailing list. I haven't thought through all the ramification, but in general we do try to minimize axiom use, so it makes sense in that context... but there may be a downside I don't see.

wlammen:
@david-a-wheeler I'd like to see the results first in a browser, before continuing or discussing the results on google groups.

benjub:
> Once it's added, do you want to raise the idea of switching to it? If so, that's a big enough change that it's probably worth a discussion on the mailing list.

Indeed it is worth it: https://groups.google.com/g/metamath/c/Ovxv2aXJOIM/m/9WRk8TgHBwAJ...
See also http://us2.metamath.org/mpeuni/df-bj-nf.html and following theorems. What is nice about mathboxes is that it allows collaborations. Well, if one is willing to collaborate.

**Edit:** Also mentioned here: https://github.com/metamath/set.mm/pull/1375#issuecomment-570196308.  That said, I'm willing to recognize that I took too much time to implement this plan which is two years old, and I'm happy to let @wlammen carry it through.

wlammen:
> I took too much time to implement this plan which is two years old...

@benjub Me too. Actually I think we briefly exchanged notes on this topic, agreeing that df-nf is inferior to nf2. I don't remember the details, though.

wlammen:
I'm done for today.  Nothing more will be added to this pull request.

PR #2192, opened 2021-09-17
wl-nfim

wlammen:
If we know that x is not free in each operand, the same holds for all propositional operations using them. This result is very elementary for a nf2 style definition, and holds even in many logic models not supporting sort of idem-potence of A. x (hba1, see wl-nf2-nf).

Remove some outdated OLD theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -5246,8 +5246,6 @@
 "df-odOLD" is used by "odfvalOLD".
 "df-odzOLD" is used by "odzvalOLD".
 "df-omsOLD" is used by "omsvalOLD".
-"df-ovolOLD" is used by "ovolvalOLD".
-"df-pellfundOLD" is used by "pellfundvalOLD".
 "df-ph" is used by "isphg".
 "df-ph" is used by "phnv".
 "df-pjh" is used by "pjhfval".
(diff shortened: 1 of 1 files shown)


benjub:
@wlammen So that our PRs do not confilct, can you do in your next PR:
* comment of wl-nf2, first line: missing closing parenthesis
* move wl-nf2i, wl-nf2ri, wl-nf2d earlier since they use only prop calc (I put a note in the comment of wl-nfntht2 reading "theorems up to this point use only propositional calculus", and we should do the same for theorems using only ax-1--ax-4, etc)
* remove wl-nftht ? or ALT/OLD, and rename my wl-nftht0 to wl-nftht

benjub:
Note regarding your (@wlammen) latest commits: I think theorems should be ordered by which axioms they require, so this would put wl-nf2-nf and wl-nf-nf2 after "wl-nf?", which use only propositional calculus.

wlammen:
@benjub Yes, I prefer ordering them by axiom usage, too. This means nfa1 and nfe1 have to move further back. Note: I am still at work, the latest pushes were made, when I had to wait for something, and could not proceed with the work earning me money.

wlammen:
Feel free to change things the way you like, its sufficient to leave a note in the pull request. In particular, fix missing parens or misspelling at your leisure.

wlammen:
I put wl-nf-nf2 and wl-nf2-nf in front of wl-nf2, because they are justifications for the changes made in this section. Much like bijust. But I agree, they will finally have to appear after the definition of 'not free' which is introduced too early for the proofs of these explaining theorems.
Edit: fix mispelling

wlammen:
Won't add more in the next few hours

PR #2196, opened 2021-09-18
wl-nfimd

wlammen:


wlammen:
Will add more
Edit Found some reductions of axiom usage, that I will submit

PR #2199, opened 2021-09-19
wl-nfal

wlammen:
1. add wl-nfal to mathbox, needs ax-10 (like nfex, but unlike the original nfal)
2. move csbima12gOLD to Alan Sare's Mathbox. This (very) outdated theorem has its only applications there.

Explanation: Alan Sare passed away several years ago, and his Mathbox is kept in honour of his contributions.  Since the Main part slowly develops away from what his Mathbox is based upon, I thought it is time to cut outdated dependencies and move (for now just) one OLD theorem to its sole user.  Hopefully this is not recognized as a disrespectful attitude.  Please let me know what you think.


nmegill:
What you did is OK.  However, in this case we could just to use csbima12 + a1i in proofs that use csbima12gOLD, so that we can get rid of csbima12gOLD completely.

The problem is a little deeper.  There are many csb\*gOLD and sbc\*gOLD in his mathbox.  I added the OLD suffix to his sbc\*g/csb\*g versions when I did a massive change to remove unnecessary antecedents in Aug. 2018.  My hope was that he would update his "virtual deduction" proofs eventually, using the \*OLD suffixes to guide him, and remove the unnecessary antecedents from his *VD versions.  But of course that didn't happen.

One option is to remove the OLD suffixes from these (thus undoing what I did in 2018) so that his mathbox becomes essentially the same as he left it.  I don't think there would be many (if any) name conflicts since the "g" versions of most sbc/csb are no longer there; if there are conflicts, the ones in his mathbox could have ALT added.  These theorems have "discouraged" tags, so there shouldn't be a problem with accidentally using them.  The purpose of many of these theorems is primarily to illustrate his "virtual deduction" method. 

We could also leave the OLD suffixes in place, since they do suggest that a stronger version (without the antecedent) exists in main.  Or we could change the OLD on those to some other suffix to reduce confusion when looking for obsolete OLDs to delete.  I'll listen to what others think.

Below is his email of 27-Sep-2018 about this, which was actually his last email to me.  I am posting it with permission from his estate.

---------

Hi Norm,

Thanks for giving me a heads up regarding this.

I haven't given much thought to it yet and I haven't yet reviewed the affected theorems.

It seems like my choice would be to either revise the affected VD proofs or lose them as VD proofs.  It may be very easy for
me to make the necessary revisions or maybe not.  If its easy, I'll probably end up revising them.  If not, I'm not sure what I would and might end up procrastinating until the generous window of opportunity has closed.

I have always been impressed by dynamic and flexible nature of set.mm

Alan

-----Original Message-----
From: Norman Megill <nm@alum.mit.edu>
To: alansare3 <alansare3@aol.com>
Sent: Tue, 18 Sep 2018 17:19
Subject: sbc* and csb* updates

Hi Alan,

A few weeks ago I reproved many substitution theorems to eliminate
unnecessary sethood antecedents.  (This was the result of a decision to
relax an earlier prohibition on using df-sbc directly - see comments in
dfsbcq.)  For example, the "A e. V" antecedent on sbcang was eliminated,
and now it is called sbcan.

The old versions still exist with \*OLD suffixes.  While I've eliminated
them everywhere else, I didn't want to touch your \*VD proofs since they
serve a different purpose than just showing a shortest proof.  I will
let you decide what to do with them.

Our policy is to keep \*OLD versions for a year.  So you have 11 months
to decide what to do.  If you don't do anything, then before deleting
the \*OLDs I will run 'minimize \*sbc\*,\*csb\*' on your *VD proofs, or edit
them directly as needed, so that they use the new versions.

BTW your proofs are "locked" with "(Proof modification is discouraged.)"
tags, and by default MM-PA won't let you change them.  You can override
this with the 'set discouragement off' command.

Here are the \*VD theorems that need updating:

Statement "sbcangOLD" is referenced in:
  csbingVD onfrALTlem5VD onfrALTlem4VD csbxpgVD csbunigVD
Statement "sbcorgOLD" is referenced in:
  sbc3orgVD
Statement "sbcalgOLD" is referenced in:
  trsbcVD sbcssgVD
Statement "sbcexgOLD" is referenced in:
  onfrALTlem5VD csbxpgVD csbrngVD csbunigVD
Statement "sbcbiiOLD" is referenced in:
  eqsbc3rVD
Statement "sbcel1gvOLD" is referenced in:
  sbcoreleleqVD onfrALTlem4VD
Statement "sbcel12gOLD" is referenced in:
  csbxpgVD csbrngVD
Statement "sbcel2gOLD" is referenced in:
  sbcssgVD csbingVD csbunigVD
Statement "csbabgOLD" is referenced in:
  csbingVD csbsngVD csbxpgVD csbrngVD csbunigVD csbfv12gALTVD
Statement "csbingOLD" is referenced in:
  onfrALTlem5VD onfrALTlem4VD csbresgVD
Statement "csbunigOLD" is referenced in:
  csbfv12gALTOLD csbfv12gALTVD
Statement "csbxpgOLD" is referenced in:
  csbresgVD
Statement "csbresgOLD" is referenced in:
  csbima12gALTVD
Statement "csbima12gOLD" is referenced in:
  csbfv12gALTVD
Statement "csbrngOLD" is referenced in:
  csbima12gALTVD
Statement "sbc3orgOLD" is referenced in:
  sbcoreleleqVD

Thanks,
Norm

PR #2210, opened 2021-09-23
replace iunonOLD with iunon in one instance

wlammen:


nmegill:
3 out of the 5 verifiers are failing because the proof of inar1 is incomplete.  The ones that pass (but shouldn't) are checkmm and metamath.exe

I downloaded this set.mm version and in metamath.exe, 'verify proof \*' gives "Warning: The following $p statement(s) were not proved:  inar1".  @david-a-wheeler any idea why the script is not detecting this?

nmegill:
When run manually, checkmm.cpp gives "Warning: Proof of theorem inar1 is incomplete".

digama0:
My guess is that these verifiers have explicit detection for incomplete proofs and consider them "acceptable", hence the warning. In strict verification mode they should be rejected. The easy way to do this without changing the verifiers is to grep for these warning messages, but I think this should be considered a bug in the verifier, as these are not valid proofs according to the spec, even if treating them as a softer kind of error makes sense for proof authors.

nmegill:
@digama0 The spec (p. 115) says, "A proof verifier may ignore any proof containing `?` but should warn the user that the proof is incomplete."  For PR submissions, though, we don't want any incomplete proofs, and I think all warnings as well as errors should cause the checks to fail (as they now do with e.g. "verify markup" warnings).

While we can discuss whether the spec should be changed to require that incomplete proofs be flagged as hard errors, my original intent with the spec wording was to allow the verification to continue gracefully to the end and not exit abnormally like it might with a severe syntax error.  It was intended to be friendlier to developers who will typically have several incomplete proofs in progress.  The milder word "warning", which they will see all the time as a result, distinguishes them from errors in the syntax or whatever that they need to fix before continuing.

digama0:
> While we can discuss whether the spec should be changed to require that incomplete proofs be flagged as hard errors, my original intent with the spec wording was to allow the verification to continue gracefully to the end and not exit abnormally like it might with a severe syntax error.

Ideally, this would be handled as an error flag in the verifier, such that hitting an incomplete proof sets the flag without halting the process, and when it gets to the end it reports the error via exit code and/or printing. I like error code reporting in particular because it makes it simple for scripts to boil down all the complexity of the verification into a single bit - is the proof good or not - without having to look for particular verifier-specific keywords.

wlammen:
There are several issues with this pull request:
1. I was unable to figure out how to revoke it.
2. I replayed the proof in metamath twice with the necessary adaptions to the iunon line, but failed at apparently the same location (!) corresponding to the path of lines on the web page for inar1: 270 -> 261 -> 255 -> 254 -> 252 -> 251 -> 249 -> 248 -> 247 -> 246 -> 245 -> 244 -> 242 -> 235 -> 205 -> 194 -> 193 -> 190 -> 189 -> 188 -> 187 -> 49.  I am used to look out for bounded setvars and usually assign them directly.
3. User request: I'd love to be hinted to incomplete proofs, in particular when proof variables become unreferenced, and are still not fully determined.

wlammen:
At the depicted point a variable maps to the final conjunct in elina, and this conjunct is a qualified expression with an undetermined bound variable (A. z e.A ...). This conjunct is irrelavant to the proof, but somehow needs to be mapped to some dummy variable during substitution. The mapping succeeds, apart from the value of the bound variable.
This is hard to see and to fix in metamath, since you now need to supply a complete expression instead of a variable name. :-/

nmegill:
@digama0 
> Ideally, this would be handled as an error flag in the verifier, such that hitting an incomplete proof sets the flag without halting the process, and when it gets to the end it reports the error via exit code and/or printing.

metamath.exe is a combination verifier and development tool set.   Some errors and warnings may later be resolved in the proof assistant.  I suppose we could have the very last "verify proof *' (and specifically that command), immediately followed by "exit", set the final exit error code if there are incomplete or incorrect proofs.  Otherwise we wouldn't set the exit code (except in the case of syntax errors or warnings when reading a database).

I'll think about this, but for the time being, a "Warning" from metamath.exe means the same as "Error" for the purpose of calling the database verified. "Warning" indicates a lesser severity level - meaning we can continue to work with the database - but is still an error that must not be ignored, just like car inspection will not pass if a warning light is turned on.  (Do any cars have error lights?)  Maybe my choice of the word "Warning" rather than a more technical "Error: severity level 2" was misguided, but that's the way it is now.

So for the time being the Github script needs to check for both words in both metamath.exe and checkmm verifications. 

@wlammen 
> User request: I'd love to be hinted to incomplete proofs, in particular when proof variables become unreferenced, and are still not fully determined.

Where would you want this to appear?  In response to a command?

While they don't address your request directly, here are some things I do.

1. "verify proof */syntax" is a quicker way to check for incomplete proofs than waiting for a full verification.
2. In MM-PA. "improve all" for a proof that has no missing essential steps will have the response "The proof is already complete." if the proof is complete, and "No new subproofs were found." if there are missing dummy variable assignments.  To see them in the latter case, use "show new_proof /unknown /all".

wlammen:
@nmegill I'd like to have it displayed while I am entering the proof.  In the above example: After having typed in 'assign last elina' in MM-PA, it could, after printing the list of statements yet to prove, print a warning like 'Incomplete assignment to variable $xyz going out of scope'. I could then simply undo the last operation, assign a proper value to the variable, and reissue the command.  If you defer this warning until the end, it is too late for an easy handling. The proof of inar1 is way too long, for example. It took me already quite some time to pinpoint the exact position of my error, let alone fix it.  The latter may be due to my lack of knowledge of the proof explorer. The usual proofs I did so far didn't exceed some 50 lines or so. In case of an error, I simply replayed the proof from scratch, but this strategy is a pain in the case of inar1.

Thanks for considering my request.

Wolf

nmegill:
> @nmegill I'd like to have it displayed while I am entering the proof. In the above example: After having typed in 'assign last elina' in MM-PA, it could, after printing the list of statements yet to prove, print a warning like 'Incomplete assignment to variable $xyz going out of scope'. I could then simply undo the last operation, assign a proper value to the variable, and reissue the command.

Although I'm not sure what you mean by "going out of scope", the problem was that a work variable (like "$4"), representing a dummy variable in the proof, wasn't assigned. In MM-PA you assign it with a fresh setvar variable e.g. "let variable $4 = z", then you do "improve all" to complete the proof.  I don't see how an undo would be involved.

This can happen in prop calc too.  For example, if you "prove idALT/override", then "delete floating", "initialize all", "unify all", "improve all", there will be no steps shown with "show new_proof /unknown".  But "show new_proof /unknown/all" will show that work variable $2 wasn't assigned.  In this case, since there are no $d's involved, you can assign $2 with any legal wff variable or expression to complete the proof.  In the existing proof we reuse "ph" which results in a slightly shorter proof.

> If you defer this warning until the end, it is too late for an easy handling. The proof of inar1 is way too long, for example. It took me already quite some time to pinpoint the exact position of my error, let alone fix it. The latter may be due to my lack of knowledge of the proof explorer. The usual proofs I did so far didn't exceed some 50 lines or so. In case of an error, I simply replayed the proof from scratch, but this strategy is a pain in the case of inar1.

What I would propose is the following.  Currently there are several commands that automatically perform "show new_proof / unknown" after completion, such as after an "assign", for the user's convenience to see what remains to be proved.  This behavior could be modified as follows.  If all essential steps are assigned but the proof is still incomplete as in this case, instead of printing nothing it as it does now, it would print "Use SHOW NEW_PROOF /UNKNOWN /ALL to see remaining unassigned steps."   (While we could automatically perform this actual command at that point, I think it would be too verbose and potentially confusing to have some unknown step printouts implicitly use /all and others not.)

I can add this to my todo list for the program if it seems like something people want.  I'll try to do it in a future version, but it isn't quite as trivial as it might appear on the surface.  Currently, the automatic "show new_proof /unknown" after "assign" etc. is done by a function that doesn't keep track of whether or not anything was actually printed, so that will have to be enhanced.

digama0:
> I suppose we could have the very last "verify proof *' (and specifically that command), immediately followed by "exit", set the final exit error code if there are incomplete or incorrect proofs.

I would suggest something like `verify proof * /exit` which means "verify and then exit with a relevant error code" for this, rather than adding additional state between commands.

But like you say, for CI it's probably fine to just error on any "Error" or "Warning", since we are also checking non-verification stuff like markup, and the result of `verify proof * /exit` would presumably not include any warnings from previous commands.

nmegill:
Yes, "verify proof \*/exit" is a good idea.  I will add it to metamath.exe. 

The algorithm will set an error indicator if there were syntax warnings/errors during "read" (and clear it with "erase"), then "verify proof \*/exit"  will return the OR of that and any verification warnings/errors.  It will return with exit(0) for no errors and exit(1) for warnings/errors, which I think is standard.

david-a-wheeler:
@nmegill - yes, that's standard. More precisely, an error return of 0 means "no error" aka "true", while an error return of nonzero means "error" aka "false". You can use different nonzero integers for different errors, if you think another process might want to distinguish between various kinds of errors.

nmegill:
> What I would propose is the following. Currently there are several commands that automatically perform "show new_proof / unknown" after completion, such as after an "assign", for the user's convenience to see what remains to be proved. This behavior could be modified as follows. If all essential steps are assigned but the proof is still incomplete as in this case, instead of printing nothing it as it does now, it would print "Use SHOW NEW_PROOF /UNKNOWN /ALL to see remaining unassigned steps."

This will be incorporated in the next metamath.exe version (0.199).  The algorithm is that if there are no unknown essential steps where MM-PA would normally print them, but there are still remaining unknown steps, this message will be printed. E.g.
```
MM> prove a1i
...
MM-PA> show new_proof/all
1   wph=wph   $f wff ph
2     wph=wps   $f wff ps
3     wps=wph   $f wff ph
4   wps=wi    $a wff ( ps -> ph )
5   min=a1i.1 $e |- ph
6     wph=wph   $f wff ph
7     wps=wps   $f wff ps
8   maj=ax-1  $a |- ( ph -> ( ps -> ph ) )
9 a1i=ax-mp $a |- ( ps -> ph )
MM-PA> delete step 3
Step 3 was deleted.
Use SHOW NEW_PROOF / UNKNOWN / ALL to see remaining unassigned steps.
MM-PA> delete step 5
Step 5 was deleted.
5      min=?     $? |- ph
MM-PA>
```

nmegill:
> I would suggest something like `verify proof * /exit` which means "verify and then exit with a relevant error code" for this, rather than adding additional state between commands.

I have decided instead to add `/exit` to `read ... /verify`:

`read set.mm /verify /exit`.

This will print the usual parsing and proof verification errors and warnings.  It will then exit(1) if there were any errors or warnings; otherwise it will exit(0).

Doing it this way will keep the new code in one place instead of having to keep track of a global "parsing error found" variable for use by the proposed `verify proof */exit`.

This will be available in the forthcoming version 0.199.

PR #2217, opened 2021-09-25
Add imaex and shorten 33 proofs

jamesjer:
I would like to suggest the addition of `imaex`.  I kept finding myself wanting it for a set of theories I am working on.  Adding it increases the size of `set.mm` by 241 bytes.  Out of 53 theories in the main part of `set.mm` that currently use `imaexg`, 33 can use `imaex` instead, leading to a reduction in size of 575 bytes.  I also ran a full `MINIMIZE_WITH` on each theorem, leaving `set.mm` 911 bytes shorter than when I started.


avekens:
The hypothesis should read "imaex.1 $e |- A e. _V $." instead of "imaex.1 $e |- A e. V $.", See section "17.1.1  Conventions", bullet point "Is-a-set.":
* For hypotheses ($e statement) of theorems (mostly in inference form), however, ` |- A e. _V ` is used rather than ` |- A e. V ` (e.g. ~ difexi ).

Or are there any of the 33 theorems which cannot be shorted anymore after this change?

nmegill:
@jamesjer - @avekens is right that our convention is `$e |- A e. _V` rather than `$e |- A e. V`.  My intuition has been that this leads to slightly shorter proofs most of the time, since  an extra `$f` hypothesis is avoided, and there is an exact match to `_V` when an earlier `$p |- ... e. _V` is used to satisfy the hypothesis, eliminating a substitution of `_V` for `V`.  The disadvantage is that occasionally we need to use elexi, which increases proof length; so far elexi has been needed 227 times, which I think is a very small percentage of cases.

So it should be changed to `$e |- A e. _V` to conform to the convention.  But not all is lost:  this looks like a good test case to see if this intuition is correct, by comparing the savings of `$e |- A e. _V` vs. `$e |- A e. V`, and I'd be interested to know the result.

jamesjer:
Yes, I missed that part of Conventions.  I've made the requested change.  The proofs shrank by between 1 and 12 bytes apiece.  Nearly all of the changes were 1 or 2 bytes; there were just a handful that were more than that.  On the other hand, the proofs of `cnheiborlem` and `shsval` cannot use the new version of `imaex` without also using `elexi`, but the original proofs with `imaexg` are shorter.  With the new version and 31 proofs minimized, the size of set.mm drops by 952 bytes, 41 bytes more than with the original version and 33 proofs minimized.  I would say your intuition is justified.

jamesjer:
Well, since I botched the `ssenen` proof, make that 958 bytes instead now that the proof is correct.  Sorry about that.

PR #2221, opened 2021-10-01
Add links between 2p2e4 and o2p2e4

david-a-wheeler:
Add relevant hyperlinks between some proof comments.

The proof 2p2e4 is clear, but depends on the construction of
complex numbers, and some people will want to see a proof that
depends on fewer steps. Add links in the comments between 2p2e4 and o2p2e4,
expand df-suc to help explain ordinal numbers, and link to df-suc
from o2p2e4.


jkingdon:
This is fine as far as it goes but the description of ordinals is perhaps a little bit misleading as it might make it sound like limit ordinals don't exist or that ordinals only include natural numbers.

david-a-wheeler:
@jkingdon - I was going for simplicity, and that's always a risk. Can you suggest improved wording?

jkingdon:
> @jkingdon - I was going for simplicity, and that's always a risk. Can you suggest improved wording?

One fix is to be even more minimal. Don't change the df-suc comment and remove the sentence "Ordinal numbers are modeled as Von Neumann ordinals; see ~ df-suc ." Less urgently, maybe remove "(finite integers starting at 0)" (because we are talking about `_om` not `NN0`). Other than that the changes to o2p2e4 and 2p2e4 would work as proposed.

I could imagine trying to introduce ordinals in a sentence or two (say, at df-ord), with something sort of like "0, 1, 2 ... w, w + 1 ... w + w (= 2 x. w) ... 3 x. w ... w x. w ..." but (a) that's a bit tricky, in terms of getting the level of detail right (it would be easy for this to only make sense to people who don't really need it), and (b) it might be hard to figure out where to link it so that people who might benefit from it would see it, and/or that it links out to df-om or whatever is implied by that informal notation.

Another choice would be to try to explain (in the o2p2e4 comment) what 2 and 4 are in terms of the empty set and pairs and singletons but no ordinals bigger than 4. But as I try to come up with wording I'm not sure it is falling into place, so I guess I'd suggest the minimal solution above.

benjub:
I'm pretty sure that the proof of ~2p2e4 does **not** depend on set.mm's construction of the complex numbers: @nmegill took great care of restating as axioms the key results about complex numbers, and discouraging the corresponding justification statements, e.g. ax-resscn/axresscn, ax-1cn/ax1cn, etc.

What you call "ordinal natural number" is usually called "finite ordinal".

nmegill:
> I'm pretty sure that the proof of ~2p2e4 does **not** depend on set.mm's construction of the complex numbers: @nmegill took great care of restating as axioms the key results about complex numbers, and discouraging the corresponding justification statements, e.g. ax-resscn/axresscn, ax-1cn/ax1cn, etc.

That is true, but the question in the mm group post had to do with deriving 2p2e4 from ZFC axioms.  The complex number axioms are in a sense "convenience" axioms which I originally\* added to prevent very long "This theorem depends on definitions:" lists on every web page using complex numbers.  They also to isolate the construction section so that in principle another construction could be plugged in.  To get a full proof of 2p2e4 directly from ZFC axioms, rather than using additional axioms on top of ZFC, we would replace all references to ax-1cn with ax1cn etc.  I did this temporarily in order to get the 2p2e4 statistics on the MPE home page.

(\*Well, historically the complex number axioms came first and I later added the construction. Once the construction was complete, I changed $a ax-1cn to $p ax1cn etc., for several years, but after some discussions that this made the web pages cluttered with stuff most people didn't care to see all the time, I once again isolated the complex number axioms.)

> What you call "ordinal natural number" is usually called "finite ordinal".

Yes.  I used "ordinal natural number" in my post thinking that some people might not know what a "finite ordinal" is, but I agree it is awkward.  Perhaps

"Ordinal natural numbers defined using this successor function and 0 as the empty set are also called von Neumann ordinals; 0 is the empty set {}, 1 is {0, {0}}, 2 is {1, {1}}, and so on."

 could be

"Natural numbers can be defined using the successor function, starting from the empty set to represent 0.  These are called the (finite) von Neumann ordinals:  0 is the empty set {}, 1 is {0, {0}}, 2 is {1, {1}}, and so on."

If @david-a-wheeler and @benjub and @jkingdon agree, David can make that change and I'll merge the PR.

benjub:
I'll let you choose the best wording, but the proposed PR contains
> and thus it [the proof of 2p2e4] depends on our construction of complex numbers

which is not true.  Maybe write:

> This proof is simple, but it depends on many other proof steps because 2 and 4 are defined as complex numbers, which in set.mm's encoding are more complicated objects than finite ordinals. The proof ~ o2p2e4 is similar but proves ` 2 + 2 = 4 ` when 2 and 4 are encoded as  finite ordinals, and thus depends on fewer intermediate steps.

jkingdon:
> Natural numbers can be defined using the successor function, starting from the empty set to represent 0. These are called the (finite) von Neumann ordinals: 0 is the empty set {}, 1 is {0, {0}}, 2 is {1, {1}}, and so on

I'm not sure how much to indulge in word-smithing but I suppose I'd write this as

> Natural numbers can be defined using the successor function, starting from the empty set to represent zero. These are called the (finite) von Neumann ordinals: zero is ` (/) `, ` 1o ` is ` { (/), { (/) } } `, ` 2o ` is ` { 1o, { 1o } } `, and so on.

Having said that, I don't really object to the wording quoted at the start of this comment even if it is less "metamath-like" in terms of being specific about notations which are often overloaded in informal writing.

nmegill:
I'm going to merge this PR so it won't grow stale and risk merge conflicts.  I don't object to the other proposed wordings.   @benjub and/or @jkingdon , feel free to change them in new PRs if you wish.

@jkingdon  I would agree that formal Metamath-specific symbols should normally be used in comments in general.  But \*2p2e4 is a widely viewed special case that even gets occasional mention on reddit (at least twice causing the previous us.metamath.org provider to complain about it using too much of the account's "unlimited" data allowance).  @david-a-wheeler  is using informal symbols to attempt to appeal to a wider audience.  I can see both points of view and don't feel strongly for or against the use of formal symbols in the o2p2e4 comment, but I will accept your change to use them if that is what you prefer (and hopefully at least someone else also prefers them).

jkingdon:
> don't feel strongly for or against the use of formal symbols in the o2p2e4 comment

I have much the same problem; I too can see both sides.

david-a-wheeler:
I'll create another PR that tries to improve things by taking these comments into account.

PR #2227, opened 2021-10-04
Full minimization through "Ordinals"

jamesjer:
I have been wishing for a `scripts/minimize-part`, that would let me do a full minimization of the theorems in part of a metamath file.  I've got a patch that adds 2 more command line parameters to `scripts/list-theorems`, with the following properties:

- If a nonempty 3rd argument is given, all theorems preceding $3 are removed from the output
- If a nonempty 4th argument is given, all theorems following $4 are removed from the output

I've also modified `scripts/minimize-all` to pass $3 and $4 to list-theorems.  With that in place, I can run, for example:

```
scripts/minimize-all '*' set.mm "" onnev
```

which does a full minimization of all theorems in set.mm up to and including `onnev`.  This pull request is for the result of running exactly that command.  As one would expect, no new axioms are used, but the trace back of `sbhypf` does show the addtion of `cab`, `df-clab`, `df-clel`, `cvv`, and `df-v` due to replacing `ceqsexv` with `equsexvw`.

This pull request aside, are you interested in the changes to `scripts/list-theorems` and `scripts/minimize-all`?  The "all" part of the latter name becomes a bit of a misnomer, but perhaps that doesn't matter.  My thought is that this would be useful for doing a full minimization of a mathbox, or other section of a metamath file.


benjub:
When you mention "the addition of cab...", do you mean "the removal" (since currently sbhypf uses those) ?  Actually, one of my  goals with equsexv, equsexvw, equsalvw (and bj-equsalv, which I should move to Main) was this kind of removal.  If you can try and minimize theorems thanks to these, it would be nice (only if it does not take you too much time).

Also, can you upload these new scripts ?

jamesjer:
Oops, I did the diff backwards.  You're right, I should have said "the removal".  Sorry about that.  I will try to do the requested minimizations in the next few days.

I will make a separate pull request for the script changes in a moment.

nmegill:
I'm all in favor of full minimization, but I'm wondering if doing it right now it might interfere with the change of df-nf that is currently in progress. @wlammen could you check that this PR will not conflict with your work?  Thanks.

wlammen:
@nmegill Anything between df-nfold (line 25880) and nf5NEW (line 29170) should be left untouched for a limited time, if it directly or indirectly uses F/. This is not the case, so go ahead.

benjub:
@jamesjer: out of curiosity, what was the approximate runtime for the minimization through onnev ?

jamesjer:
@benjub I apologize.  I overlooked your question.  I'm not sure of the exact running time.  It was long enough that I left for lunch while it was still running, but short enough that it was done when I returned, so less than an hour.  The machine that I ran this on is several years old (a quad-core Intel i7-6700 with hyperthreading), and is slowed somewhat by Linux mitigations for the various Meltdown and Spectre vulnerabilities.

PR #2229, opened 2021-10-05
df-nf change step #3

wlammen:
1. change definition of not-free;
2. delete outdated theorems


benjub:
That's some nice work !  It looks like you added a spurious "D" in "[--samDe--]".

wlammen:
There is still a lot to do: The list of moved theorems needs to be updated. Some theorems affected by the change lost their OLD counterparts, Comments and texts need an update and so on, and so forth. This is something for the weekend.

PR #2230, opened 2021-10-05
Improvements of comments

avekens:


benjub:
There is a comment
> the (complex) left modules over the rational or real or complex numbers are complex vector spaces.

You probably meant to remove the first "complex" ?

avekens:
No, the comment should be correct: in set.mm, the complex left modules/vector spaces CMod resp. CVec are actually left modules/vector spaces over subrings of CC, not only over CC itself. See headrr comment of section "Complex vector spaces".

benjub:
Ah, yes, I remember now...  I should do the changes (which I think were agreed upon but which I did not carry out):
> complex vector space --> subcomplex vector space
> complex left module --> subcomplex module

Since these are non-standard notions, they should not have names which universally mean something else.  Also, CC being commutative, I'll remove the "left".

avekens:
OK, I did not know that there was a discussion about the naming. Please feel free to revise the comments accordingly (also my comments, after this PR was accepted). According to the name "left module": Since CC (or a subring) is only the scalar field/ring, and the scalar product is defined as multiplication with the scalar from the left side, the "left" in the name is still justified and can be "left in the name" ;-).

avekens:
OK, according to Wikipedia (https://en.wikipedia.org/wiki/Module_(mathematics)): "If R is commutative, then left R-modules are the same as right R-modules and are simply called R-modules. ". I think "are the same" must be replaced by "are isomorphic" to be more precise (formally, the scalar multiplications are different functions, so the modules cannot be "the same"). Then it it, however, justified to remove the "left" from the name (Gérard Lang also proposed this in an email to me).

benjub:
You're right that it should be left... technically. But at this point (I mean, for more "advanced" topics further away from set theory), this is more a matter of how things are specifically encoded in set.mm than about the true nature (if there is such a thing) of the mathematical object. So for all purposes, let's say that there are "the same" (or we could have a metamath database for hott and univalent mathematics!).

PR #2231, opened 2021-10-06
janitorial work

wlammen:
1. shorten nfv;
2. consistent naming of theorems;
3. duplicated theorems;
4. update history of theorems (still incomplete);
5. group OLD theorems for a. better comparison b. less disruptive experience for normal users;
6. restore some OLD theorems that went lost, but are suitable for comparison'.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -184,11 +184,15 @@
 "10reOLD" is used by "problem2OLD".
 "10reOLD" is used by "tgblthelfgottOLD".
 "10reOLD" is used by "tgoldbachOLD".
+"19.21OLD" is used by "19.21hOLD".
+"19.21t-1OLD" is used by "19.21tOLD".
+"19.21tOLD" is used by "19.21OLD".
 "19.41rg" is used by "ax6e2nd".
 "19.41rg" is used by "ax6e2ndALT".
 "19.41rg" is used by "ax6e2ndVD".
 "19.9OLD" is used by "19.9hOLD".
 "19.9dOLD" is used by "19.9tOLD".
+"19.9tOLD" is used by "19.21tOLD".
 "19.9tOLD" is used by "19.9OLD".
 "1idpr" is used by "1idsr".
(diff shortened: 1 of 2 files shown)


benjub:
I think wnfOLD can also be removed, since it is identical to wnf (what's important is df-nfOLD).

benjub:
Also, the labels I had given to nftht0, nfntht, nfntht2 are not very good... Maybe @nmegill or @digama0 have better suggestions?

wlammen:
> I think wnfOLD can also be removed, since it is identical to wnf (what's important is df-nfOLD).

I am not sure here.  I think you cannot define the same thing twice.  If I remember right, it was mmj2 that complains about that and rejects the idea.  Growling sth like 'Definitions are not sound'.

In a situation where you compare two directly competing ideas like df-nf vs nf5, I think a full and enforced separation is a good idea.  This leaves open, what to do, if two competing theories share a common base. This leads to the concept of interfaces, something I know from my profession.  Here is headroom for improvement and relaxation of rules.

In the case of not-free could one have allowed a common usage of the symbol? Because you can demonstrate equivalence in BOTH cases, no matter what definition you pick?  Well, actually only, if the axioms up to ax-12 are in full effect. Before you are on unsafe grounds.

wlammen:
I added yet another commit to this pull request, but it wouldn't show up.  Did I surpass some internal counter limit?  My private copy on github shows a total count of 16 commits, but only 15 are transmitted here.  Anyway, I will leave it at that, and create another PR once this one is accepted.

wlammen:
@benjub I am pretty sure I overlooked some details. In the end, it was a massive change, at least compared to my usual contributions.  I work full time in a job that is neither related to Metamath, or mathematics or science in general. Which puts some stress on me in a situation like this, where others possibly wait for me to finish. Feel free to correct anything you spot without bothering to ask for permission.

PR #2234, opened 2021-10-08
janitorial work

wlammen:
restore some old theorems, that got lost.  This finally finishes the transition df-nfOLD -> df-nf


wlammen:
Does someone have problems applying the new semantics of 'not free'? Should something like a readme be released on this topic?

benjub:
Nice work !  As for a possible readme, I think the comment at df-nf is sufficient.  You may post in the thread opened on the google group to inform everyone that you carried the change through, with maybe some information about what you did, what you deleted or deprecated and why (I haven't followed everything).

PR #2236, opened 2021-10-09
shorten nfan

wlammen:


avekens:
Hi Wolf, during the last weeks, you removed a lot of OLD theorems using the old notation for the infimum. Thank you very much for this. There are about 37 theorems left to be deleted (obsolete as of 6-Oct-2020). If you do not mind, I would like to do this, and I want to check if there are still some old notations of the infimum in use afterwards, to finish issue #1805. These changes should not interfere with your current work with "not free".

Alexander

wlammen:
@avekens Please go ahead.  There are still a few inf*OLD theorems left and in use.  Best consult the discouraged file, where they are listed alphabetically. And the change of 'not free' is completed.  No need to hesitate any longer, just add/remove/change theorems at your leisure.

PR #2240, opened 2021-10-11
shorten stdpc5

wlammen:
1. stdpc5 is a specialization of 19.21;
2. nfnf use less symbols in visual display of the proof;
3. axc11v and axc11rv are direct consequences of axc16g, so shorten them all in total by reversing proof dependencies;
4. move a couple of direct consequences of sp into the neighborhood of sp;
5. revert axc16nf to former proof, ax-10 cannot be avoided under current definition of df-nf.


benjub:
Nice. One typo:
> the ~ ax12v --> the result ~ ax12v

or "the theorem", if you prefer.

wlammen:
> Nice. One typo:

Done.

Issue #1810, opened 2020-09-07
Prove convergence of Cauchy sequences of positive reals in iset.mm


jkingdon:
Right now iset.mm has http://us.metamath.org/ileuni/cauappcvgpr.html which says that a Cauchy approximation with values in `Q.` has a limit in `P.`.

The extension to a Cauchy sequence (which is indexed by a natural number, rather than by a rational number indicating the desired closeness of terms) should be straightforward. Here is my current statement of what we are trying to prove:

```
  ${
    $d A j $.
    caucvgpr.f $e |- ( ph -> F : _om --> Q. ) $.
    caucvgpr.cau $e |- ( ph ->
      A. x e. Q. E. j e. _om A. k e. _om ( j <Q k ->
      ( ( F ` j ) <Q ( ( F ` k ) +Q x ) /\
        ( F ` k ) <Q ( ( F ` j ) +Q x ) ) ) ) $.
    caucvgpr.bnd $e |- ( ph -> A. j e. _om A <Q ( F ` j ) ) $.
    $( A Cauchy sequence of postiive fractions converges to a positive real.
       This is basically Corollary 11.2.13 of [HoTT], p.  (varies) (one key
       difference being that this is for positive reals rather than signed
       reals).  We also specify that every term needs to be larger than a
       fraction ` A ` , to avoid the case where we have positive fractions
       which converge to zero (which is not a positive real).  (Contributed by
       Jim Kingdon, 18-Jun-2020.) $)
    caucvgpr $p |- ( ph ->
        E. y e. P. A. x e. Q. E. j e. _om A. k e. _om ( j <Q k ->
        ( <. { l | l <Q ( F ` k ) } , { u | ( F ` k ) <Q u } >.
            <P ( y +P. <. { l | l <Q x } , { u | x <Q u } >. ) /\
          y <P <. { l | l <Q ( ( F ` k ) +Q x ) } ,
            { u | ( ( F ` k ) +Q x ) <Q u } >. ) ) ) $=
      (  ) ? $.
  $}
```

Using the proof of Corollary 11.2.13 of [HoTT] as our guide, there are several steps to proving this:

[...]

PR #2242, opened 2021-10-12
Add floor/ceiling theorems to iset.mm

jkingdon:
At least as a first pass, we just define floor and ceiling for rationals. It would presumably be possible to define them for real numbers which are apart from any integer (such as any number which is irrational in the sense of "apart from any rational number" which is the usual constructive definition of irrational), but this pull request does not do that (it briefly mentions this possibility in a few comments).

The pull request includes:

* A variety of miscellaneous theorems which were recently added to set.mm
* A variety of rational number theorems such as trichotomy and related results. Generally these are similar to theorems we already had for ` ZZ `.
* Adding subfzo0 and a few other theorems right before the floor/ceiling section which aren't especially related except that they are from roughly the same part of the file so I noticed their presence.
* An intuitionizing of most of the theorems from the floor/ceiling section. A few didn't seem worth it, and at the end of the section I omitted a few which we may want to add later, but for the most part everything carries over (with real changed to rational except where the theorem already was for rationals).
* Perhaps the most mathematically interesting part of the pull request is the proof of qbtwnz , which is the rational number analogue to http://us.metamath.org/mpeuni/rebtwnz.html . The proof starts by finding two integers which are less than and greater than the given rational number. Then this range can be shrunk by choosing an integer in between the endpoints of the range and then deciding which half of the range to keep based on rational number trichotomy, and iterating until the range consists of two consecutive integers. This is fairly easy to describe in words, but expressing it as a proof by induction is slightly tricky and the technique here is similar to the one from http://us.metamath.org/ileuni/prarloclem3.html (although fortunately simpler because we don't have quite as many complications as in that proof).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4513,6 +4513,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>ceicl</TD>
+  <TD>~ ceiqcl</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4354,6 +4354,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flcl</TD>
+  <TD>~ flqcl</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4369,6 +4369,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flltp1</TD>
+  <TD>~ flqltp1</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4553,6 +4553,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>ceilidz</TD>
+  <TD>~ ceilqidz</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4446,6 +4446,13 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>ico01fl0</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably could be proved for rationals, but lightly used in
+  set.mm.</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4507,6 +4507,12 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>dfceil2 , ceilval2</TD>
+  <TD><I>none</I></TD>
+  <TD>Unused in set.mm.</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4483,6 +4483,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>fldivle</TD>
+  <TD>~ flqle</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4389,6 +4389,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>fllt</TD>
+  <TD>~ flqlt</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4364,6 +4364,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flle</TD>
+  <TD>~ flqle</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4523,6 +4523,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>ceilge</TD>
+  <TD>~ ceilqge</TD>
+</TR>
+
+<TR>
   <TD>ceige</TD>
   <TD>~ ceiqge</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4518,6 +4518,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>ceilcl</TD>
+  <TD>~ ceilqcl</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4369,7 +4369,7 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
-  <TD>flltp1</TD>
+  <TD>flltp1 , fllep1</TD>
   <TD>~ flqltp1</TD>
 </TR>
 
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4384,6 +4384,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flge</TD>
+  <TD>~ flqge</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4359,6 +4359,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>fllelt</TD>
+  <TD>~ flqlelt</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4425,6 +4425,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flword2</TD>
+  <TD>~ flqword2</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4463,6 +4463,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>refldivcl</TD>
+  <TD>~ flqcl</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4415,6 +4415,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flltnz</TD>
+  <TD>~ flqltnz</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4090,10 +4090,15 @@ in iset.mm</TD>
 </TR>
 
 <TR>
-  <TD>zbtwnre , rebtwnz</TD>
+  <TD>zbtwnre</TD>
   <TD><I>none</I></TD>
   <TD>Proved in terms of supremum theorems and presumably not possible
   without excluded middle.</TD>
+</TR>
+
+<TR>
+  <TD>rebtwnz</TD>
+  <TD>~ qbtwnz</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4568,6 +4568,12 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>quoremz , quoremnn0 , quoremnn0ALT</TD>
+  <TD><I>none</I></TD>
+  <TD>Unused in set.mm.</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4586,6 +4586,12 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>fldiv</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would be provable if real is changed to rational.</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4592,6 +4592,19 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>fznnfl</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would be provable if real is changed to rational.</TD>
+</TR>
+
+<TR>
+  <TD>uzsup , ioopnfsup , icopnfsup , rpsup , resup , xrsup</TD>
+  <TD><I>none</I></TD>
+  <TD>As with most theorems involving supremums, these would
+  likely need significant changes</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4394,6 +4394,12 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flflp1</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on case elimination.</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4354,8 +4354,8 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
-  <TD>flcl , reflcl</TD>
-  <TD>~ flqcl</TD>
+  <TD>flcl , reflcl , flcld</TD>
+  <TD>~ flqcl , ~ flqcld</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4420,6 +4420,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flwordi</TD>
+  <TD>~ flqwordi</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3580,7 +3580,7 @@ favor of theorems in deduction form.</TD>
 
 <TR>
 <TD>lelttric</TD>
-<TD>~ zlelttric </TD>
+<TD>~ zlelttric , ~ qlelttric</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4574,6 +4574,12 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>intfrac2</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would be provable if real is changed to rational.</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4586,7 +4586,7 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
-  <TD>fldiv</TD>
+  <TD>fldiv , fldiv2</TD>
   <TD><I>none</I></TD>
   <TD>Presumably would be provable if real is changed to rational.</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4478,6 +4478,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>flmulnn0</TD>
+  <TD>~ flqmulnn0</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4523,6 +4523,11 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>ceige</TD>
+  <TD>~ ceiqge</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4580,6 +4580,12 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>intfracq</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably provable.</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3597,6 +3597,11 @@ only if they are apart" which relies on excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>leneltd</TD>
+  <TD>~ leltapd</TD>
+</TR>
+
+<TR>
 <TD>dedekind , dedekindle</TD>
 <TD><I>none</I></TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4468,6 +4468,16 @@ proof is not intuitionistic and it is lightly used in set.mm.</TD>
 </TR>
 
 <TR>
+  <TD>fladdz</TD>
+  <TD>~ flqaddz</TD>
+</TR>
+
+<TR>
+  <TD>flzadd</TD>
+  <TD>~ flqzadd</TD>
+</TR>
+
+<TR>
 <TD>df-mod and other modulus theoerms.</TD>
 <TD><I>none</I></TD>
 <TD>As with df-fl it may be feasible to define this for rational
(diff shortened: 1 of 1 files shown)


benjub:
Regarding ` qbtwnz $p |- ( A e. QQ -> E! x e. ZZ ( x <_ A /\ A < ( x + 1 ) ) ) ` and out of curiosity, what can we hope to prove if ` A e. RR ` ? Can we prove the same thing ? Or do we have to either weaken the conclusion or strengthen the hypothesis, e.g.:

* For all A in RR and for all 1 < delta in RR, there exists x in ZZ such that x <= A and A < x + delta ?

* For all A in RR apart from an integer, there exists a unique x in ZZ such that x < A and A < x + 1 ? (yes by your earlier comment)

**Edit:** By the way: having detailed comments for most of your PRs is very valuable !

digama0:
I don't think your first statement is provable, because the function mapping A in RR to x(A) in ZZ is discontinuous. The second statement should be provable, because the hypothesis pokes holes in the domain such that the resulting function can be continuous (locally constant) even though it is not constant.

benjub:
That first statement is equivalent to:
\forall x \in \RR \forall \epsilon > 0 \exists q \in \QQ q < A < q + \epsilon.
So this is not intuitionistically provable ? What about density of QQ in RR ?

jkingdon:
> That first statement is equivalent to: \forall x \in \RR \forall \epsilon > 0 \exists q \in \QQ q < A < q + \epsilon. So this is not intuitionistically provable ? What about density of QQ in RR ?

What you say here sounds like http://us.metamath.org/ileuni/prarloc2.html which is for ` P. ` and ` Q. ` and which I assume would hold for ` RR ` and ` QQ ` (although looking at the comments above about continuity and idly pondering Cauchy versus Dedekind reals - a topic I have little intuition about - I don't know whether that would be "RR as specified by the current complex number axioms" or "RR with an additional axiom which follows from the current construction").

As for whether that indeed implies "For all A in RR and for all 1 < delta in RR, there exists x in ZZ such that x <= A and A < x + delta" I didn't immediately see the proof. But it sounds like you have one in mind.

jkingdon:
> By the way: having detailed comments for most of your PRs is very valuable !

Always glad to see I spurred some thought and/or discussion. Hopefully we won't go so far down a tangent that we make it hard to merge this pull request (or decide not to), which unless I misunderstand is probably not much affected by all this talk of what is possible beyond rationals.

benjub:
I hope the following is intuitionistic:
Let A in RR and delta in RR with delta > 1.
By "localizability of Dedekind cuts" applied to A with epsilon = delta - 1,
there exists q in QQ such that q < A < q + delta - 1.
By "localizability of rationals in intervals of the form [n, n + 1[ with n
in ZZ" applied to that q, there exists n in ZZ such that n <= q < n + 1.
For that n, one has n <= q < A < q + delta - 1 < n + delta.

digama0:
That proof works, given prarloc2, but this theorem is not available for axiomatic complex numbers. However, the archimedean property is an axiom (ax-arch), and I believe it can be used to prove density of QQ.

Given `A e. RR` and `n e. NN`, by the archimedean property applied to `A*n` and `-A*n` we can deduce that there exists `a,b e. ZZ` such that `A*n e. (a, b)`. We can then proceed by induction on `b-a` to reduce the size of the interval: If `b-a <= 2` then we're done, and otherwise let `c = b-2` and `d = b-1` so that `a < c < d < b`; then using axltwlin we have `c < d` so `c < A*n` or `A*n < d`, so either `A*n e. (c, b)` (an interval of size 2) or `A*n e. (a, d)` (an interval of size b-a-1).

So by induction we can prove that `A*n` is strictly located between integers at most 2 apart, that is, there exists an integer `a e. ZZ` such that `|A*n-a| < 1`, and therefore `|A-a/n| < 1/n`. Using the archimedean property one more time, for any `epsilon > 0` there exists an `n` such that `1/n < epsilon`, so `|A-a/n| < epsilon`. Thus QQ is dense.

benjub:
That the archimedean property implies density of QQ is fundamental.  I
hope your proof makes it to iset.mm.

david-a-wheeler:
> By the way: having detailed comments for most of your PRs is very valuable !

I agree. That said, I encourage everyone (me included) to include the important information as comments in the `.mm` file itself, including hyperlinks to related material. That way readers of the "final result" can easily see all that good information, placed where they're most likely to encounter it.

jkingdon:
> > By the way: having detailed comments for most of your PRs is very valuable !
> 
> I agree. That said, I encourage everyone (me included) to include the important information as comments in the `.mm` file itself, including hyperlinks to related material. That way readers of the "final result" can easily see all that good information, placed where they're most likely to encounter it.

A fair point and I think this pull request mostly already does that. For example, "at the end of the section I omitted a few which we may want to add later" matches quite closely new entries in mmil.html in the pull request.

I will include in my next pull request an edit to add "The proof starts by finding two integers which are less than and greater than the given rational number. Then this range can be shrunk by choosing an integer in between the endpoints of the range and then deciding which half of the range to keep based on rational number trichotomy, and iterating until the range consists of two consecutive integers." to the comment of `qbtwnzlemexz` . To be honest, how to summarize proofs is one of the things I struggle with most. On the one hand, the actual metamath proof is the most authoritative description for how a proof goes (and that goes double if a large proof is nicely decomposed into lemmas). On the other hand, metamath is so low-level that this is by necessity a bit drowning in details so I can see a place for some level of informal proof in the comments some of the time. But I think overdoing this can obscure rather than clarify.

Also, I'd encourage others to submit changes to comments and web pages, whether to add content that is in pull requests or for other reasons. This is partly to spread out the work, and also because I'm not necessarily the best judge of what information people would find most helpful.

jkingdon:
> That the archimedean property implies density of QQ is fundamental.

Density like http://us.metamath.org/mpeuni/qbtwnre.html ? That's definitely on the hit list of theorems we want and is, for example, Theorem 11.2.6 of [HoTT] (and shouldn't need excluded middle or anything else which departs from the real numbers as generally understood in this sort of setting).

Personally I don't have an especially strong opinion about whether to use a construction-dependent proof (and thus add a new complex number axiom) or a proof on top of the existing axioms, but the biggest practical problem with the former is that I don't see an especially easy way to map between `Q.` and `QQ`. So partly for that reason and partly just because I suppose I do have a preference for reducing the number of axioms (albeit a weak preference), I'll see if I can formalize the proof above.

benjub:
> Density like http://us.metamath.org/mpeuni/qbtwnre.html ?

That's order-density (and it's good that it's on the list too!).  I was refering to Mario's proof, which is about topological density, but of course they are closely related since the topology of RR is generated by its order.

As for using the axiomatic reals or their construction from IZF, I'd say that the field of reals is a central enough object to first establish its axioms and then prove later results from them, as Mario indicated (if not, then my earlier proof would be enough). An additional argument for the axiomatic way is that in the intuitionisitc case, there seem to be more constructions of the reals, which are subtly inequivalent (as opposed to the classical Dedekind cut = Cauchy completion).

jkingdon:
> > Density like http://us.metamath.org/mpeuni/qbtwnre.html ?
> 
> That's order-density (and it's good that it's on the list too!). I was refering to Mario's proof, which is about topological density, but of course they are closely related since the topology of RR is generated by its order.

Hmm, the first definition I found for topological density was "A set D is dense if and only if there is some point of D in each nonempty open set of X" so I can see how that could apply without an order. In the case of `QQ` and `RR` the definitions look kind of the same to me, but if there is another way of stating density worth adding (other than `qbtwnre`), we can do that.

> As for using the axiomatic reals or their construction from IZF, I'd say that the field of reals is a central enough object to first establish its axioms and then prove later results from them

I'm not opposed to this but I want to be realistic that it would probably take some work by someone other than me to make this a reality. Starting with figuring out what axioms we want (I suppose I'd start with the ones in [Geuvers]) and proceeding to removing the axioms which are now there but which are redundant.

> there seem to be more constructions of the reals, which are subtly inequivalent (as opposed to the classical Dedekind cut = Cauchy completion).

We know we have at least Dedekind reals and Cauchy reals (and that we'd need countable choice to show the two are equivalent). [HoTT] chapter 11 is largely about this topic.

I'd probably also refer to [HoTT] for at least a start on some of the other issues, for example around the impredicativity which tends to go with Dedekind cuts.

benjub:
> > > Density like http://us.metamath.org/mpeuni/qbtwnre.html ?
> > 
> > 
> > That's order-density (and it's good that it's on the list too!). I was refering to Mario's proof, which is about topological density, but of course they are closely related since the topology of RR is generated by its order.
> 
> Hmm, the first definition I found for topological density was "A set D is dense if and only if there is some point of D in each nonempty open set of X" so I can see how that could apply without an order. In the case of `QQ` and `RR` the definitions look kind of the same to me, but if there is another way of stating density worth adding (other than `qbtwnre`), we can do that.

Indeed everything is closely related since the topology of RR is generated by its order, so that order theoretic density is the same thing as topological density, and very easily so. Even if this is a very close and trivial restatement, maybe the formulation `(A e. RR /\ E e. RR) -> (0 < E -> E. q e. QQ |A - q| < E)` is worth adding?  This is topological density phrased not using open subsets but neighborhoods: every point in RR is an accumulation point of QQ, that is, all of its neighborhoods intersect QQ (and the open intervals centered around A form a neighborhood basis of A... here we are using the fact that the topology is generated by the metric).

Also, is my proposal above `(A e. RR /\ D e. RR) -> (1 < D -> E. n e. ZZ (n < A /\ A < n + D)` worth adding to iset.mm ?

> > As for using the axiomatic reals or their construction from IZF, I'd say that the field of reals is a central enough object to first establish its axioms and then prove later results from them
> 
> I'm not opposed to this but I want to be realistic that it would probably take some work by someone other than me to make this a reality. Starting with figuring out what axioms we want (I suppose I'd start with the ones in [Geuvers]) and proceeding to removing the axioms which are now there but which are redundant.

I agree that it's probably more work (and that you are the one doing the work while I'm just waving hands!), and it's better to have the results proven not synthetically but from the IZF construction than to have nothing at all.  Also, finding the best set of axioms for the intuitionistic reals (or, as in set.mm, for the complex and real numbers at once), is probably nontrivial.

> > there seem to be more constructions of the reals, which are subtly inequivalent (as opposed to the classical Dedekind cut = Cauchy completion).
> 
> We know we have at least Dedekind reals and Cauchy reals (and that we'd need countable choice to show the two are equivalent). [HoTT] chapter 11 is largely about this topic.
> 
> I'd probably also refer to [HoTT] for at least a start on some of the other issues, for example around the impredicativity which tends to go with Dedekind cuts.

Thanks for the reference, and indeed adding some comment on the difference between Dedekind and Cauchy reals and about predicativity will be welcomed !

jkingdon:
> maybe the formulation `(A e. RR /\ E e. RR) -> (0 < E -> E. q e. QQ |A - q| < E)` is worth adding?  . . .
> 
> Also, is my proposal above `(A e. RR /\ D e. RR) -> (1 < D -> E. n e. ZZ (n < A /\ A < n + D)` worth adding to iset.mm ?

If you are unsure you can always put them in your mathbox.

But if you think they belong outside a mathbox, fine with me. I mean I could try to mutter things about how fundamental they are or whether they will be used by other theorems or whatever, but I haven't been especially strict with myself on such matters so I won't try to be too strict with anyone else unless it seems like we have problems with clutter or disorganization or whatever (neither of which seem to be especially worrying for the two proposed above).

digama0:
I think a theorem about density of QQ, broadly construed, will certainly be useful in later theorems. Here are some statements that I think will be useful:

* arch: `A e. RR -> E. n e. NN A < n`
* nnrecl: `A e. RR+ -> E. n e. NN 1/n < A`
* btwnz, first part: `A e. RR -> E. n e. ZZ n < A`
* qbtwnre: `A e. RR /\ B e. RR /\ A < B -> E. q e. QQ ( A < q /\ q < B )`
* ``A e. RR /\ B e. RR+ -> E. q e. QQ abs`(A - q) < B``

Most of these already exist in set.mm, and I have the names written here. The one @benjub suggests about approximate floors of course doesn't exist in set.mm, since there is a real floor function, but it might come in useful. The version I used in the proof isn't quite as tight as @benjub's version, but is probably good enough for many uses:

* ``A e. RR -> E. n e. ZZ abs`(A - n) < 1``

benjub:
For the last one, here is a (probably clumsy) proof. Is the label OK with you, so that I can do a PR ?
```
  ${
    $d A x $.  $d B x $.
    $( The rational numbers are dense in ` RR ` : any real number can be
       approximated with arbitrary precision by a rational number.  For order
       theoretic density, see ~ qbtwnre .  (Contributed by BJ, 15-Oct-2021.) $)
    qdenre $p |-
           ( ( A e. RR /\ B e. RR+ ) -> E. x e. QQ ( abs ` ( x - A ) ) < B ) $=
      ( cr wcel crp wa cmin co caddc clt wbr w3a cv cabs cfv wrex simpl adantl
      cq rpre resubcld readdcld ltsubrp ltaddrp lttrd 3jca qbtwnre qre ad2antlr
      wi simpll absdiflt biimprd syl3anc reximdva syl5 mpd ) BDEZCFEZGZBCHIZDEZ
      BCJIZDEZVBVDKLZMZANZBHIOPCKLZATQZVAVCVEVFVABCUSUTRZUTCDEZUSCUAZSZUBZVABCV
      KVNUCZVAVBBVDVOVKVPBCUDBCUEUFUGVGVBVHKLVHVDKLGZATQVAVJAVBVDUHVAVQVIATVAVH
      TEZGVHDEZUSVLVQVIUKVRVSVAVHUISUSUTVRULUTVLUSVRVMUJVSUSVLMVIVQVHBCUMUNUOUP
      UQUR $.
  $}
```

digama0:
Seems fine to me. I'm surprisingly out of practice for metamath naming...

jkingdon:
> I think a theorem about density of QQ, broadly construed, will certainly be useful in later theorems. Here are some statements that I think will be useful: . . .
> Most of these already exist in set.mm, and I have the names written here.
> 

The first four of these (arch , nnrecl , btwnz , qbtwnre) already exist in iset.mm (the last one new as of yesterday, but it is there), with the same names and same statement of the theorem as in set.mm.

> A e. RR /\ B e. RR+ -> E. q e. QQ abs`(A - q) < B

I see @benjub has a proof for this one. Yay!

> A e. RR -> E. n e. ZZ abs`(A - n) < 1

Hmm, this is worth adding if for no other reason than that I'm having trouble doing the conversion from http://us2.metamath.org/ileuni/rebtwn2z.html to a version involving absolute value in my head. I ended up adding a factor of two to the proof of http://us2.metamath.org/ileuni/qbtwnre.html to deal with `< 1` versus `< 2` but now I'm wondering whether it would have also worked to add/subtract one.

jkingdon:
> For the last one, here is a (probably clumsy) proof. Is the label OK with you, so that I can do a PR ?

Excellent. The label looks good - there are two (maybe unwritten?) rules of metamath naming that can help here: (1) a non-systematic name is fine especially if the theorem doesn't obviously fit one of the existing patterns, and (2) we can always change it later (and often do), especially if a better name becomes apparent. I'm not even sure "non-systematic" is even fair in this case - qdenre does seem like a pretty nice way to say "QQ is dense in RR".

As for whether the proof is clumsy, if I had let that stop me we wouldn't have half of iset.mm. For one thing, we can improve/shorten it later if need be. Although I did look it over and (a) I'm guessing it is probably roughly as short as it can be (qbtwnre, absdiflt, and the usual amount of glue) (b) I didn't see any improvements which jumped out at me.

benjub:
A comment/question: have you proved somewhere ` $e |- A. x e. [0, 1] ( x = 0 \/ x =/= 0) $.  $p |- ( ph \/ -. ph ) $= ? $.` ?

Then, I record an obvious proof here:
` $e |- A. x e. [0, 1] E. n e. ZZ ( n <= x /\ x < n + 1 ) $.  $p |- ( ph \/ -. ph ) $= ? $.`
Informal proof: it suffices to prove the "$e" of the above.  Let x e. [0, 1].  Then, one has 1 - x e. [0,1], so by hypothesis E. n e. ZZ ( n <= 1 - x /\ 1 - x < n + 1 ), so in particular n <= 1 - x <= 1 and -1 <= -x < n, so n is 0 or 1.  If n = 0, then x > -n = 0 so x =/= 0. If n = 1, then x <= 1 - n = 0 but since 0 <= x, one has x = 0.

**Edit:** That's really nitpicking, but it would probably be better to prove the following three inferences
` $e |- A. x e. [0, 1) ( x = 0 \/ x =/= 0) $.  $p |- ( ph \/ -. ph ) $= ? $.`
` $e |- A. x e. (0, 1] E. n e. ZZ ( n <= x /\ x < n + 1 ) $.  $p |- A. x e. [0, 1) ( x = 0 \/ x =/= 0) $= ? $.`
` $e |- A. x e. (0, 1] E. n e. ZZ ( n <= x /\ x < n + 1 ) $.  $p |- ( ph \/ -. ph ) $= ? $.`

jkingdon:
> A comment/question: have you proved somewhere ` $e |- A. x e. [0, 1] ( x = 0 \/ x =/= 0) $. $p |- ( ph \/ -. ph ) $= ? $.` ?

I assume by [0, 1] you mean the real closed interval? ( `( 0 [,] 1 )` in iset.mm notation).

As far as I can tell from Theorem 3 on page 4 of http://www1.maths.leeds.ac.uk/~matpsh/DoraisHirstShaferRMTD.pdf , real number trichotomy (well, maybe just dichotomy, I didn't read the paper enough to grasp some of the finer distinctions here) does not imply excluded middle, but instead the Lesser limited principle of omniscience.

It would be nice to add things like this but we're starting from a pretty basic place along the lines of "how do we even state the Lesser limited principle of omniscience" (maybe we want to add a notation sort of like `CHOICE` from http://us.metamath.org/mpeuni/df-ac.html ?).

> Then, I record an obvious proof here: ` $e |- A. x e. [0, 1] E. n e. ZZ ( n <= x /\ x < n + 1 ) $. $p |- ( ph \/ -. ph ) $= ? $.` Informal proof: it suffices to prove the "$e" of the above. Let x e. [0, 1]. Then, one has 1 - x e. [0,1], so by hypothesis E. n e. ZZ ( n <= 1 - x /\ 1 - x < n + 1 ), so in particular n <= 1 - x <= 1 and -1 <= -x < n, so n is 0 or 1. If n = 0, then x > -n = 0 so x =/= 0. If n = 1, then x <= 1 - n = 0 but since 0 <= x, one has x = 0.

Oooh nice. Even if we don't tackle the larger issues (yet) we could add a proof that `A. x e. [0, 1] E. n e. ZZ ( n <= x /\ x < n + 1 )` implies `A. x e. [0, 1] ( x = 0 \/ x =/= 0)`.

As for whether to do this for `RR`, `( 0 [,] 1 )`, `( 0 (,] 1 )`, `( 0 [,) 1 )`, etc, I guess my instincts go to `RR` as being the most fundamental, but I'm not sure I have a strong opinion on that part.

david-a-wheeler:
> I'm surprisingly out of practice for metamath naming...

Good news, we have some of the conventions written down here:
http://us.metamath.org/mpeuni/conventions-label.html
Mario, I think you helped write some of that text :-).

digama0:
> Oooh nice. Even if we don't tackle the larger issues (yet) we could add a proof that `A. x e. [0, 1] E. n e. ZZ ( n <= x /\ x < n + 1 )` implies `A. x e. [0, 1] ( x = 0 \/ x =/= 0)`.
> 
> As for whether to do this for `RR`, `( 0 [,] 1 )`, `( 0 (,] 1 )`, `( 0 [,) 1 )`, etc, I guess my instincts go to `RR` as being the most fundamental, but I'm not sure I have a strong opinion on that part.

I would take the theorem to be about `RR`, saying that equality is decidable: `A e. RR /\ B e. RR -> DECID (A = B)`. The proof can easily be strengthened in this direction: If `D := abs(A-B)` then  `A = B` iff `D / (D + 1) = 0`, and `D/(D+1) e. [0, 1]`.

> As far as I can tell from Theorem 3 on page 4 of http://www1.maths.leeds.ac.uk/~matpsh/DoraisHirstShaferRMTD.pdf , real number trichotomy (well, maybe just dichotomy, I didn't read the paper enough to grasp some of the finer distinctions here) does not imply excluded middle, but instead the Lesser limited principle of omniscience.

I think it should imply the limited principle of omniscience, which I know as "markov's principle": If `A. n e. NN DECID ph(n)`, then `E. n e. NN ph(n) \/ A. n e. NN -. ph(n)`. The idea is to consider the real number `sum_ n e. NN (if ph(n) (2^-n) 0)`. The `if` is well defined because we are assuming `ph` is decidable, and it is clearly a convergent series with an explicit modulus of convergence, so it converges to some real number `x`. If `x = 0`, then `A. n e. NN -. ph(n)`, and if `x # 0` then `E. n e. NN ph(n)`.

Note that I had to upgrade the second conjunct to `x # 0`, but @benjub 's proof yields this stronger form anyway.

benjub:
> > A comment/question: have you proved somewhere ` $e |- A. x e. [0, 1] ( x = 0 \/ x =/= 0) $. $p |- ( ph \/ -. ph ) $= ? $.` ?
> 
> I assume by [0, 1] you mean the real closed interval? ( `( 0 [,] 1 )` in iset.mm notation).
> 
> As far as I can tell from Theorem 3 on page 4 of http://www1.maths.leeds.ac.uk/~matpsh/DoraisHirstShaferRMTD.pdf , real number trichotomy (well, maybe just dichotomy, I didn't read the paper enough to grasp some of the finer distinctions here) does not imply excluded middle, but instead the Lesser limited principle of omniscience.

The condition `A. x e. [0, 1) (x = 0 \/ x =/= 0)` is probably stronger than real number dichotomy (`A. x e. RR (0 <= x \/ x <= 0)`), and it may imply full excluded middle.

An additional result: the weak form `A. x e. (-1, 1) E. n e. ZZ (n <= x /\ x <= n + 1)` (note the non-strictness) implies real number dichotomy.

digama0:
Hm, it occurs to me that there is a missing real number axiom: every real number is a limit of rational numbers. This is clear from the construction, but I don't think it is possible to prove it using only the archimedean property (=> density of QQ and approximate floor) and completeness (which only helps if you already have a sequence). You would need countable choice to construct a sequence from the existentials, because there is no canonical way to pick the sequence of rationals.

Another random observation, while playing with the axioms: the completeness axiom is equivalent to a variation with less quantifier complexity: If `F` is a sequence such that for all `k >= n`, `|F(k) - F(n)| < 1/n`, then there exists `y e. RR` such that `|y - F(n)| <= 1/n` for all `n`. (This has the same hypothesis but a stronger conclusion - the resulting inequality is not asymptotic). Clearly this version implies the original: If `x e. RR+` then by the archimedean property there exists `1/(2n) < x`, so `|y - F(k)| <= |y - F(n)| + |F(k) - F(n)| < x` for `k >= 2n`. Conversely, if `y` is a limit of `F`, then (reasoning by contradiction which works for `<=`), if `|y - F(n)| > 1/n`, then let `a := |y - F(n)| - 1/n > 0`, so that `|y - F(k)| > a` for all large enough `k`, which is a contradiction.

> The condition `A. x e. [0, 1) (x = 0 \/ x =/= 0)` is probably stronger than real number dichotomy (`A. x e. RR (0 <= x \/ x <= 0)`), and it may imply full excluded middle.

Real number dichotomy implies the LLPO: If x_n is a 0/1 sequence with at most one 1, then `sum_ n e. NN (-1)^n (x_n / n)` is a convergent series which converges to `x`, and if `x <= 0` then for all n, `x_(2n) = 0`, and if `x >= 0` then for all n, `x_(2n+1) = 0`.

In a similar way, decidability of real number equality leads to decidability of Pi_1 statements, i.e. `A. n e. NN ph(n)` where `ph` is decidable, but not the LPO or Markov's principle unless you boost the disequality to an apartness.

jkingdon:
> Hm, it occurs to me that there is a missing real number axiom: every real number is a limit of rational numbers. This is clear from the construction

Limit as in limit of a sequence? Why is it clear? Wouldn't you need countable choice in a construction-dependent proof too?

I think your statement is equivalent to "Dedekind reals = Cauchy reals" but it is possible there is something I misunderstand, either about the mathematics or about what axiom you are proposing.

digama0:
> > Hm, it occurs to me that there is a missing real number axiom: every real number is a limit of rational numbers. This is clear from the construction
> 
> Limit as in limit of a sequence? Why is it clear? Wouldn't you need countable choice in a construction-dependent proof too?

For Cauchy reals, it is clear because the reals are a quotient of rational sequences. For Dedekind reals, you are right that you won't get this property, and I forgot that the current implementation uses Dedekind reals. The analogous property for Dedekind reals says that every real number is the supremum of a set of rational numbers.

What are your thoughts regarding exposing real number axioms that witness the difference between Cauchy and Dedekind reals? I suspect that iset.mm may want both eventually, because the relations between them are interesting. Rather than doing the whole construction twice, a lighter weight version would be to carve out the Cauchy reals as a subset of RR (Dedekind reals), defined as the set of reals that are the limit of a sequence of rationals. You can then use the real number theorems to show that this set is also closed under lots of operations and acts like RR.

jkingdon:
> What are your thoughts regarding exposing real number axioms that witness the difference between Cauchy and Dedekind reals?

Sounds cool. We could use ℝ<sub>c</sub> for the Cauchy reals (following the HoTT book's notation).

> Rather than doing the whole construction twice, a lighter weight version would be to carve out the Cauchy reals as a subset of RR (Dedekind reals)

That sounds better than trying to show the Cauchy reals as a separately constructed thing are isomorphic to a certain subset of the Dedekind reals, especially given the lack of "automatically apply an isomorphism" functionality in metamath. And I guess in this scenario we'd just keep ℝ as the notation for Dedekind reals, because although ℝ<sub>d</sub> might be clearer in some contexts, there are plenty of other contexts where it isn't especially important which reals we are talking about, and metamath doesn't especially make it easy to have two notations which mean the same thing.

benjub:
> That sounds better than trying to show the Cauchy reals as a separately constructed thing are isomorphic to a certain subset of the Dedekind reals

That result is really interesting, and it would be great to formalize it.  Of course, you choose your own goals !

If I understand correctly what @digama0 said, the rationals are always dense in RR (Dedekind of Cauchy), and every Cauchy real is the limit of a sequence of rationals, but not every Dedekind real is (because we do not assume countable choice) ?

Is there a set of axioms for the Dedekind reals such that when adding sequential density of the rationals you obtain an independent set of axioms for the Cauchy reals ?

digama0:
> Is there a set of axioms for the Dedekind reals such that when adding sequential density of the rationals you obtain an independent set of axioms for the Cauchy reals ?

I don't think it works like that, the axioms characterizing dedekind reals are different from the axioms characterizing cauchy reals, so you wouldn't expect one to be a subset of the other. The axiom that cauchy reals satisfy and dedekind reals don't is the "smallness axiom" saying that every cauchy real is a limit of rationals, as mentioned; on the other hand, the "largeness axiom" that dedekind reals satisfy and cauchy reals don't is that every dedekind cut of dedekind reals defines a dedekind real (which implies some statements about suprema as well). The "smallness axiom" that even Dedekind reals should enjoy is that every real is a dedekind cut of rationals.

benjub:
Thanks @digama0.  So there are common core axioms for Cauchy and Dedekind reals (probably: archimedean ordered field), then one or several axioms for Cauchy reals, and one or several axioms for Dedekind reals. And using classical logic and countable choice (or some principle of omniscience or Markov), one can show that the resulting sets of axioms are equivalent. And reversely, that they are equivalent implies some non-constructive principle.  Correct ?

As for the axioms:
* Common axioms: the reals form an archimedean ordered field
* Cauchy axioms:
  * smallness: every real is a limit of a sequence of rationals
  * largeness = Cauchy completeness : "Cauchy sequences of reals converge"
* Dedekind axioms:
  * smallness: every real is a supremum of a set of rationals
  * largeness = Dedekind completeness: "every nonempty bounded-above set of reals has a supermum"

The Cauchy and Dedekind smallness axioms make sense since an archimedean ordered field contains a well-determined copy of the rationals.  Reasonable ?

I deviated a bit far from the initial PR, because of my curiosity.  I think I should first have a look at some of the literature.

digama0:
That's roughly correct, but I think there are caveats in both cases for the smallness axiom. For example, I don't think the Cauchy smallness includes sequences without a modulus of convergence, and the Dedekind smallness might require that the set of rationals is located (which requires the upper cut to be provided as well).

benjub:
Maybe the modulus of convergence could be included in the definition of a Cauchy sequence ? As in: a Cauchy sequence is a couple consisting of a sequence u : NN \to RR and a function f : NN \to NN such that...

As for Dedekind smallness, I think I see the problem which might occurbut I'm not sure of a good fix.  The axiom should be as natural as possible and not mention Dedekind cuts, since this would feel a bit circular.

Looks like we (well, you) are quite closed to a satisfying axiomatic definition of both kinds of reals. Something which might be a bit difficult in practice (although straightforward in theory) is to reference to "rationals" in the specific axioms (they are the elements of the prime field, but that might be a bit tricky to encode in iset.mm).

digama0:
> As for Dedekind smallness, I think I see the problem which might occur but I'm not sure of a good fix. The axiom should be as natural as possible and not mention Dedekind cuts, since this would feel a bit circular.

It's not really circular, because we are talking about characterizing dedekind reals here, the set is otherwise axiomatically given. The point is to determine the entire set up to isomorphism with dedekind reals so it's not too much of a surprise that we end up talking about dedekind cuts. Also, this is a dedekind cut of *real* numbers, which does not obviously follow from the construction of dedekind reals as cuts of *rational* numbers.

After looking at some of the literature, I think I was wrong to assert that you need both upper and lower cuts, though. Given a lower cut, you can take the upper cut to be either `RR \ L` or `{x e. RR | E. y < x, y e/ L}` depending on whether you need the upper cut to be open or not. And of course you can always turn an arbitrary nonempty upper bounded set into a lower cut by closing downward.

The only thing that doesn't come for free is locatedness: For all `x < y`, either `x e. L` or `y e/ L`. In terms of the set `S` being supremumed, this says that for all `x < y`, either `x < s` for some `s e. S` or `y > s` for every `s e. S`.

> Something which might be a bit difficult in practice (although straightforward in theory) is to reference to "rationals" in the specific axioms (they are the elements of the prime field, but that might be a bit tricky to encode in iset.mm).

The natural numbers are already being referenced in the statement of ax-arch, so something similar could be done to refer to rational numbers. It is a bit annoying though. It is also possible to make the definition of `NN` and `QQ` *prior* to stating and re-stating the axioms; they have a definition that only makes use of + and * and so on so I don't see any reason this would compromise the API boundary.

We talked about four axioms for largeness and smallness of dedekind and cauchy reals, but they don't all actually have to be axioms, assuming we are working from the current axiom system. Dedekind smallness can also be expressed as an extensionality principle: if two reals have the same upper and lower rational sets, then they are equal. Because of ~axapti, it suffices to prove that if `x < y` then `x` and `y` have different cut sets, and this follows immediately from ~qbtwnre.

Cauchy largeness is already an axiom, but it is not strong enough since we are axiomatizing dedekind reals as `RR`. So Dedekind largeness would have to be added, for nonempty, upper bounded, located sets. Although, it seems difficult to ever prove that a set is located if you don't already know it has a supremum.

benjub:
> Cauchy largeness is already an axiom, but it is not strong enough since we are axiomatizing dedekind reals as `RR`.

I'm not sure I understand your sentence, since we are giving an axiom system for Cauchy reals, and an axiom system for Dedekind reals ?

digama0:
I mean the current axiom system in iset.mm. One of the axioms is cauchy completeness, but the construction is dedekind reals.

benjub:
Sorry, I hadn't seen that (ax-caucvg). So we should add to iset.mm:
```
ax-ddkcomp |- ( ( ( A C_ RR /\ A =/= () ) /\
                   E. x e. RR A. y e. A y < x /\
                   A. x e. RR A. y e. RR ( x < y -> ( E. z e. A x < z \/ A. z e. A z < y ) ) ) ->
                 E. x e. RR ( A. y e. A y <_ x /\ ( ( B e. R /\ A. y e. A y <_ B ) -> x <_ B ) ) ) $.
```
And prove, using IFOL, the core real axioms, and ax-ddkcomp (and no other axioms in IZF): `caufromddkcomp $p |- [the statement ax-caucvg]` using standard methods, probably proving the intermediate result that Dedekind completeness implies the existence of limsup's (which sounds tedious to formalize...).

digama0:
I would imagine that the proof of ax-caucvg from ax-ddkcomp would be more or less the same as the current proof, which proves ax-caucvg for the concrete construction of dedekind reals. The notation of some sets will change a bit, since you can't directly use a pair of sets as a real number, but otherwise it should be a copy paste job. (Or even a move, since we won't need the proof on the construction anymore if we drop ax-caucvg as an axiom in favor of ax-ddkcomp.)

benjub:
You're right, and I should have looked at ileuni/axcaucvg before ! That's an impressive formalization by @jkingdon !
The big proof should actually be axddkcomp, which proves that the reals thus constructed indeed satisfy ax-ddkcomp ?

jkingdon:
> You're right, and I should have looked at ileuni/axcaucvg before ! That's an impressive formalization by @jkingdon !

Thanks! It was a lot of work over about a year (looks like it started with #1777 and got as far as axcaucvg in #2100 - so almost exactly a year measured that way). I do need to share the credit with @digama0 especially at https://github.com/metamath/set.mm/issues/1810 although I'll admit it got a bit lonely when I was buried in details and didn't have any really juicy questions to ask, just a whole pile of symbols and lemmas.

As for whether this could be adapted into something which is in terms of supremums of located bounded sets, maybe? Just converting the proof of http://us.metamath.org/ileuni/cauappcvgpr.html (the first one I wrote) into http://us.metamath.org/ileuni/caucvgpr.html and then into http://us.metamath.org/ileuni/caucvgprpr.html was harder than I thought it would be. But then again, maybe some annoying details would get easier? I'm all for encouraging people to improve on what I have done, and my guess is that there are some circuitous routes in the current proof even if I'm not sure which parts could be streamlined and how.

#2253 tries to expand the comments of these theorems a bit in the sense of explaining what is going on in terms of having `cauappcvgpr`, `caucvgpr`, and `caucvgprpr`.

PR #2243, opened 2021-10-12
nfa1 without ax-12

wlammen:
That's it for today, Won't push anything more in this pull request.

The new definition of 'not free' is more symmetric around the dual expressions 'for all' and 'there exists'.  This should result in the same axiom usage for dual theorems like nfa1/nfe1.

1. Remove dependency on ax-12 from nfa1, nfna1, nfnf1;
2. Remove dependency on ax-10 from axc16nf;
3. shorten hba1;
4. turn df-nf in remarks into hyperlinks;
5. delete 19.41-1.  This partial result was proven without ax-10, but 19.41 has lost its dependency on this axiom anyway. Similar for 19.42-1, but here we have a usage;
6. reorder theorems such that those based on ax-12 only are listed, then those additionally based on ax-10, and finally the rest also using ax-11.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13950,6 +13950,7 @@ New usage of "axc16g-o" is discouraged (1 uses).
 New usage of "axc16gALT" is discouraged (0 uses).
 New usage of "axc16gOLD" is discouraged (0 uses).
 New usage of "axc16nfALT" is discouraged (0 uses).
+New usage of "axc16nfOLD" is discouraged (0 uses).
 New usage of "axc4i-o" is discouraged (5 uses).
 New usage of "axc5" is discouraged (0 uses).
 New usage of "axc5c4c711to11" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -28855,6 +28855,26 @@ scheme is logically redundant (see ~ ax10w ) but is used as an auxiliary
   nfe1 $p |- F/ x E. x ph $=
     ( wex hbe1 nf5i ) ABCBABDE $.
 
+  $( The setvar ` x ` is not free in ` A. x ph ` .  (Contributed by Mario
+     Carneiro, 11-Aug-2016.) df-nf changed.  (Revised by Wolf Lammen,
+     11-Sep-2021.)  Remove dependency on ~ax-12 .  (Revised by Wolf Lammen,
```
david-a-wheeler, line 28860: The phrase:

> df-nf changed

should be:

> ~ df-nf changed.

Also, in:

> ~ax-12

You need a space after "~":

> ~ ax-12

david-a-wheeler:
Hi, I just posted a few small tweaks to the comments so that the hyperlinks will work correctly.

PR #2246, opened 2021-10-15
add iset/qdenre

benjub:
As per https://github.com/metamath/set.mm/pull/2242#issuecomment-944225670
The other changes are due to rewrap with the new mm198.


jkingdon:
Nice.

As for the rewrapping I am seeing similar results with metamath Version 0.198 7-Aug-2021. I have just upgraded to this version so hopefully this will be a one-time problem rather than something which shows up on every pull request.

jkingdon (approved):

PR #2248, opened 2021-10-16
shorten nfald, revise nfex, hbex

wlammen:
1. The combined proofs of nfex and hbex are not shorter than the OLD ones by byte count, but use about 30 fewer symbols;
2. shorten nfald

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -6612,6 +6612,7 @@
 "hbalg" is used by "hbexgVD".
 "hbequid" is used by "equidq".
 "hbequid" is used by "nfequid-o".
+"hbexOLD" is used by "nfexOLD".
 "hbim1OLD" is used by "hbimOLD".
 "hbim1OLD" is used by "nfim1OLD".
 "hbnae-o" is used by "ax12inda2ALT".
@@ -15883,6 +15884,7 @@ New usage of "hbalg" is discouraged (1 uses).
 New usage of "hbalgVD" is discouraged (0 uses).
 New usage of "hbanOLD" is discouraged (0 uses).
 New usage of "hbequid" is discouraged (2 uses).
+New usage of "hbexOLD" is discouraged (1 uses).
 New usage of "hbexg" is discouraged (0 uses).
 New usage of "hbexgVD" is discouraged (0 uses).
 New usage of "hbim1OLD" is discouraged (2 uses).
@@ -16875,6 +16877,7 @@ New usage of "nf3orOLD" is discouraged (0 uses).
 New usage of "nfa1-o" is discouraged (4 uses).
 New usage of "nfa1OLD" is discouraged (0 uses).
 New usage of "nfa1OLDOLD" is discouraged (1 uses).
+New usage of "nfaldOLD" is discouraged (0 uses).
 New usage of "nfan1OLD" is discouraged (1 uses).
 New usage of "nfanOLD" is discouraged (0 uses).
 New usage of "nfanOLDOLD" is discouraged (3 uses).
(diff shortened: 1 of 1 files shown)


benjub:
As for nfex/hbex, the byte counts are the same, as you wrote, so the only reason why your proposed version is shorter is that the label "nfxfr" is two letters shorter than "hbxfrbi".

The old version, which first proves hbex and then nfex from it, rather than the other way around, has the advantage that it first proves the more fundamental hbex (in that it does not contain `F/`).  In particular, hbex currently does not depend on df-nf whereas with your proposal, it will (not very important either, but perhaps a bit more than label lengths).

Maybe you can add your proposed version as *ALT proofs (or yours as the main proofs and the current one as *ALT).  What do you @wlammen and @nmegill think ?  I guess the same thing will happen with hbal/nfal ?

wlammen:
The byte count is the primary criterion:
> As for nfex/hbex, the byte counts are the same

and using less symbols in the proof display is the secondary (or ternary? After line count?) criterion that decides whether one proof is shorter than another.  Since the total symbol count decreases by 30 symbols, this criterion applies.

This question has been asked before (actually more than once), so I know these criteria exist, and I know they are listed somewhere on the Metamath site.  And I have referenced them in discussions like this.  But.. forgive me, I forgot exactly where.  In case you don't believe me, I will undergo the effort and search the whole site.  But only then.

Edit: See here https://groups.google.com/g/metamath/c/O38iPe68W-A/m/qo38XvVIDQAJ
And, with less relevance, here https://groups.google.com/g/metamath/c/CMKRwV0CX3w/m/0-flqYbyAgAJ

hbal/nfal are not affected, because hbal is a direct consequence of ax-11, wheras nfal needs additionally ax-10 and ax-12.
And the old proof of hbex depends on df-nf, although not directly.  Since it depends on ax-10 and ax-12, too, where all sensible definitions of 'not-free' are equivalent, the choice of technique in its proof does not matter.

benjub:
I'm not focusing on exactly which criteria are used.  I'm saying that when two proofs are "almost tie", then maybe both should be kept as *ALT.  If you're confident that the proofs are basically the same and nothing is lost in (deprecating and one year later) deleting one of them, then it's ok.  But if there is any doubt that some insight may be lost in the deletion, then *ALT is here for that. That's all.

wlammen:
IMO there is nothing special about the old version.  Both versions are simple transscriptions of each other, valid under any choice of df-nf. If for some reason the new version must be reverted to the old one again, I would see the proof outline within seconds.  But of course, I am highly specialized in this field, and what is obvious to me, may not be to others.

What is really getting lost, is some know-how of how theorems are built up from the old definition of not-free. (Well, actually not, since you can checkout a version of set.mm as of 10-Sep-2021, and create the web pages from it.)  Yes, that's the way it is.  Each time I check in a new proof overriding an old one, I throw an idea into the bin. This might hurt emotionally, and could be tagged as a valuable loss.  But the alternative, keeping all possible proofs in parallel, looks like a nightmare to me.

benjub:
Agreed. So it's a matter of finding a good balance between not losing important ideas by deleting too many proofs and not letting set.mm be cluttered with too many ALT proofs. That's a judgment call for each case.

nmegill:
I agree with @wlammen that this wouldn't be a good use of the ALT suffix.  I think ALT should be reserved for a significantly different proof that makes a specific point spelled out in the comment of the ALT.  And I think it should be used somewhat sparingly, so that when people see ALT it should suggest to them that there is an interesting difference worth looking at, such as using a different set of axioms to prove the same thing.  Too many ALTs will dilute their significance, causing people to more or less ignore them.

The OLD suffixes stay around for a year, which is hopefully enough time for someone to make an argument that it should have been an ALT.

(Above I'm talking about the main part of set.mm.  People can do whatever they want in their mathboxes as far as their own usage of ALT.)

PR #2256, opened 2021-10-21
A few updates related to ordinals in iset.mm

jkingdon:
This pull request starts out by updating mmil.html to list a number of ordinal-related theorems which are in set.mm and not iset.mm (none of which seem possible and important to prove just now, but some of which could be proved if we want them).

After that I got interested in ordinals which are of the form ` A u. { B } ` , mainly because they show up in the induction step for various proofs using finite set induction ( using http://us.metamath.org/ileuni/findcard2sd.html or the other findcard* theorems).
Although some of what I was hoping for didn't immediately work, I did get two results:

* ((𝐴 ∈ 𝑉 ∧ {𝐴} ∈ On) → 𝐴 = ∅)
* ((𝐵 ∈ 𝑉 ∧ (𝐴 ∪ {𝐵}) ∈ On) → 𝐵 ⊆ 𝐴)
  - _would like to be able to add a `-. B e. A` condition and strengthen the conclusion to A = B but the obvious proof of that goes via http://us.metamath.org/mpeuni/ordtri1.html so this is an unsolved problem at the moment_

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1943,6 +1943,12 @@ a similar result via theorems such as ~ oneluni or ~ ssequn1 .</TD>
 </TR>
 
 <TR>
+  <TD>unizlim</TD>
+  <TD><I>none</I></TD>
+  <TD>The reverse direction is basically ~ uni0 plus ~ limuni</TD>
+</TR>
+
+<TR>
 <TD>iotaex</TD>
 <TD>~iotacl , ~ euiotaex </TD>
 </TR>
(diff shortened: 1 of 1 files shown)


jkingdon:
> the obvious proof of that goes via http://us.metamath.org/mpeuni/ordtri1.html

I take that back. `B` wouldn't be an ordinal here, and `A` hasn't yet been shown to be an ordinal. So it is a different situation than ordtri1 would cover (if we had it).

> so this is an unsolved problem at the moment

This remains true.

Issue #2201, opened 2021-09-20
Transformation of PART 18  COMPLEX TOPOLOGICAL VECTOR SPACES (DEPRECATED)


avekens:
The definitions and theorems in Part 18 COMPLEX TOPOLOGICAL VECTOR SPACES (DEPRECATED) should be transferred to be based on extensible stuctures (if not already available) and integrated into PART 12  BASIC TOPOLOGY. Afterwards, this deprecated part can/should be removed, including the deprecated definitions/theorems of the former Part 18 ADDITIONAL MATERIAL ON GROUPS, RINGS, AND FIELDS (DEPRECATED), which were only kept because they are used in the current PART 18 (see issue #1432).

Tasks:
- [x] Transformation of section 18.2  Complex vector spaces
- [x] Transformation of section 18.3  Normed complex vector spaces I (subsections 18.3.1 and 18.3.2)
- [x] Transformation of section 18.3  Normed complex vector spaces II (subsections 18.3.3 to 18.3.5)
- [ ] Transformation of section 18.4  Operators on complex vector spaces
- [ ] Transformation of section 18.5  Inner product (pre-Hilbert) spaces
- [ ] Transformation of section  18.6  Complex Banach spaces
- [ ] Transformation of section 18.7  Complex Hilbert spaces
- [ ] Deletion of PART 18 (maybe some definitions/theorems will still be used in mathhboxes - these definitions/theorems must be moved into the corresponding mathboxes)

avekens:
With PR #2226, the first task is done. A summary of the results is provided in the attached PDF document: 
[Deprecated_Part_18_Complex_Topological_Vector_Spaces_2.pdf](https://github.com/metamath/set.mm/files/7309581/Deprecated_Part_18_Complex_Topological_Vector_Spaces_2.pdf)
Many theorems had already a corresponding version in main set.mm (first column), the missing theorems were added by me (second column). There are several deprecated theorems which are currently not used (third column): @nmegill @digama0 can these theorems (and the not used theorems in the remaining sections of PART 18) be deleted immediately?

I started already to work on the second task (Transformation of section 18.3 Normed complex vector spaces). It will be finished soon (there is only a little number of theorems about the norm, most of the deprecated theorems are about arbitrary groups and vector spaces...)

benjub:
[...]

PR #2257, opened 2021-10-21
Transformation of deprecated PART 18 - not used theorems removed

avekens:
Theorems of PART 18 COMPLEX TOPOLOGICAL VECTOR SPACES (DEPRECATED), sections 18.1-3, which are not used in proofs are removed, because there are corresponding, not deprecated theorems in main set.mm, see issue #2201.

In addition, some comments are adjusted.


avekens:
Remark that ZZ-modules are complex left modules, but not complex vector spaces added, as proposed by Gérard Lang.

PR #2259, opened 2021-10-22
Addmathbox

emmett-w:
I would like to add my mathbox to the metamath repository. This is my first time publishing my mathbox, so please let me know if there are any conventions I need to follow that I'm not following already.

Thank you!

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4837,6 +4837,11 @@
 "df-rq" is used by "dmrecnq".
 "df-rq" is used by "recmulnq".
 "df-scon" is used by "isscon".
+"df-setrecs" is used by "nfsetrecs".
+"df-setrecs" is used by "setrec1".
+"df-setrecs" is used by "setrec2".
+"df-setrecs" is used by "setrec2fun".
+"df-setrecs" is used by "setrecseq".
 "df-sgrOLD" is used by "issmgrpOLD".
 "df-sgrOLD" is used by "smgrpismgmOLD".
 "df-sh" is used by "issh".
@@ -5952,6 +5957,7 @@
 "elreal2" is used by "ltresr2".
 "elringchomALTV" is used by "ringccatidALTV".
 "elrngchomALTV" is used by "rngccatidALTV".
+"elsetrecslem" is used by "elsetrecs".
 "elspancl" is used by "elspansncl".
 "elspani" is used by "spansni".
 "elspani" is used by "spanuni".
@@ -10940,6 +10946,10 @@
 "onfrALTlem4" is used by "onfrALTlem1VD".
 "onfrALTlem5" is used by "onfrALTlem3".
 "onfrALTlem5" is used by "onfrALTlem3VD".
+"onsetreclem1" is used by "onsetreclem2".
+"onsetreclem1" is used by "onsetreclem3".
+"onsetreclem2" is used by "onsetrec".
+"onsetreclem3" is used by "onsetrec".
 "opelcn" is used by "axicn".
 "opelopabsbALT" is used by "brabsb2".
 "opelopabsbALT" is used by "cnvopab".
@@ -12100,6 +12110,15 @@
 "sbcorgOLD" is used by "sbc3orgVD".
 "sbcrexgOLD" is used by "2sbcrexOLD".
 "sbcrexgOLD" is used by "sbc2rexgOLD".
+"setrec1lem1" is used by "setrec1lem2".
+"setrec1lem1" is used by "setrec1lem4".
+"setrec1lem1" is used by "setrec2fun".
+"setrec1lem2" is used by "setrec1lem3".
+"setrec1lem3" is used by "setrec1".
+"setrec1lem4" is used by "setrec1".
+"setrec2fun" is used by "setrec2".
+"setrec2lem1" is used by "setrec2".
+"setrec2lem2" is used by "setrec2".
 "sh0" is used by "ch0".
 "sh0" is used by "hhssabloilem".
 "sh0" is used by "hhssnv".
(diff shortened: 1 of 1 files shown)


nmegill:
Thank your for your contribution.  There are a few small changes we would like you to make.

We have been placing the mathboxes in alphabetical order by last name, so your mathbox should go before David A. Wheeler's.

Some of your proofs aren't in compressed format.  To make sure all proofs in your mathbox are compressed, you can do
```
MM> save proof nfintd~elpg /compressed
```

You should then run the commands shown at
https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md ,
specifically,
```
./metamath
MM> read set.mm
MM> erase
MM> read set.mm
MM> save proof */compressed/fast
MM> verify markup */file_skip/top_date_skip
MM> verify proof *
MM> write source set.mm
MM> quit
```
This will ensure a standard formatting for set.mm so that future PRs by others won't result in large reformatting changes in your mathbox.  In particular, make sure that the "verify markup" command does not produce any warnings, which would cause the CI to fail.  (Note that the `/fast` qualifier on `save proof *` above does not actually re-compress - which takes a longer time and we assume was done previously - but just reformats existing proofs, whether compressed or not, into compressed style to ensure a standard indentation and wrapping.)

Thanks!

emmett-w:
This seems to have no markup errors, but the verify markup command on my computer aborts even on the set.mm file without my addition.

In the middle of "Checking bibliographic comments...", I get an error message saying "Illegal instruction: 4." I think this is an issue with something else on my computer though and not the set.mm file.

Please let me know if I need to change anything else or if you happen to know how to fix the error. Thank you!

nmegill:
> This seems to have no markup errors, but the verify markup command on my computer aborts even on the set.mm file without my addition.
> 
> In the middle of "Checking bibliographic comments...", I get an error message saying "Illegal instruction: 4." I think this is an issue with something else on my computer though and not the set.mm file.

I haven't seen this error message before.  It isn't an error message produced from within metamath.exe.  I tried this with your set.mm on both the lcc (precompiled) and gcc versions of metamath (version 0.198 7-Aug-2021) and couldn't reproduce it.  Does it seem to depend on previous commands run in the metamath session?  Did you use the precompiled Windows version, or did you use gcc and if so what platform and compilation options?

If you can find a reproducible test case, that would be great, since it sounds like a serious bug if it is not a hardware issue.

> Please let me know if I need to change anything else or if you happen to know how to fix the error. Thank you!

The CI run failed because nfsetrecs has an incomplete proof.  That will need to be fixed (by completing the proof or commenting it out).

david-a-wheeler:
What operating system (OS), OS version, computer, and CPU are you using?

I did a Google search on the error. This appears to be a problem on MacOS at least, where the OS cannot run the binary for some reason. One potential cause is compiling on MacOS with one version, and then trying to run the resulting binary on an older version of the operating system (you may need to rerun the compiler, or add flags, to make that work). For some discussion see:

> https://stackoverflow.com/questions/14268887/what-is-the-illegal-instruction-4-error-and-why-does-mmacosx-version-min-10

nmegill:
set.mm itself has now passed the CI, but there was a mismatch in the "discouraged" file.  You should regenerate it per the instructions at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md

emmett-w:
> set.mm itself has now passed the CI, but there was a mismatch in the "discouraged" file. You should regenerate it per the instructions at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md

I'm using Mac OS, and the instructions at that link didn't change the set.mm file. Is there something I need to do differently on a Mac?



> What operating system (OS), OS version, computer, and CPU are you using?
> 
> I did a Google search on the error. This appears to be a problem on MacOS at least, where the OS cannot run the binary for some reason. One potential cause is compiling on MacOS with one version, and then trying to run the resulting binary on an older version of the operating system (you may need to rerun the compiler, or add flags, to make that work). For some discussion see:
> 
> > https://stackoverflow.com/questions/14268887/what-is-the-illegal-instruction-4-error-and-why-does-mmacosx-version-min-10

That sounds like it could definitely be what's causing the error on my computer. I'm using macOS Big Sur version 11.0.1. The processor is a 1.1 GHz Quad-Core Intel Core i5. Thank you again for your help.

nmegill:
> > set.mm itself has now passed the CI, but there was a mismatch in the "discouraged" file. You should regenerate it per the instructions at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md
> 
> I'm using Mac OS, and the instructions at that link didn't change the set.mm file. Is there something I need to do differently on a Mac?

No, there should be nothing different on a Mac.  The commands under "Regenerating the discouraged file" do not affect set.mm.  They generate a file whose name is "discouraged", that needs to checked in along with set.mm.  The purpose of the file is to allow easier identification of usage or proof modification for statements with "is discouraged" tags in their comments, so the user can make sure that the usage or proof modification was intentional and not accidental.

Normally metamath.exe and mmj2 will prevent the accidental usage or proof modification of "is discouraged" statements unless (in metamath.exe) a `/override` qualifier is present:
```
MM> prove imp4aOLD
>>> ?Error: Modification of this statement's proof is discouraged.
>>> You must use PROVE ... / OVERRIDE to work on it.
MM> prove imp4aOLD /override
You will be working on statement (from "SHOW STATEMENT imp4aOLD"):
1924 imp4.1 $e |- ( ph -> ( ps -> ( ch -> ( th -> ta ) ) ) ) $.
1927 imp4aOLD $p |- ( ph -> ( ps -> ( ( ch /\ th ) -> ta ) ) ) $= ... $.
>>> ?Warning: Modification of this statement's proof is discouraged.
MM-PA>
```

emmett-w:
Got it, thank you.

Issue #2262, opened 2021-10-23
Formatting of section headers


avekens:
I have started the revision of graph theory (see issue #1418), and I wonder if there are any (written) conventions how to format section headers and section header comments. I want to format the section headers for graph theory uniformly and according to existing conventions.

Looking at the first headers (e.g. "Propositional calculus"), there should be an empty line between the header and its comment, the comment is indented by two blanks, and there should be an empty line at the end of the comment:

```
$(
#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
                           Propositional calculus
#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#

  Propositional calculus deals with general truths about well-formed formulas
  (wffs) regardless of how they are constructed.  The simplest propositional
  ...
  for a complete cross-reference from sources used to its formalization in the
  Metamath Proof Explorer.

$)
```

If these are the conventions, we should document them in section 17.1.1 "Conventions" and adjust the headers which do not follow these conventions step by step (I think @benjub is doing this already).

Furthermore, it seems that there should be two empty lines beween the section header and the preceding text:

```
  ${
    idi.1 $e |- ph $.
    $( This inference, which requires no axioms for its proof, is useful as a
       ..
       ~ id .  (Contributed by Alan Sare, 31-Dec-2011.) $)
    idi $p |- ph $=
      (  ) B $.
  $}


$(
#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
                           Propositional calculus
#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#

  Propositional calculus deals with general truths about well-formed formulas
  ...
```
Two questions on that:
* Should this always be the case, even if the preceding text is a section header itself (e.g. between "CLASSICAL FIRST-ORDER LOGIC WITH EQUALITY" and "Pre-logic")? I don't think so.
* Should this be the case for all section levels? I think for the fourth level (-.-.-.-.-.-.-.-.-), this should not be required.

[...]

PR #2263, opened 2021-10-23
conventions: add some formatting rules

benjub:
Addressing Issue #2262.
Before merging, maybe let some time for people to suggest corrections or additions.


benjub:
@nmegill : I just read your comment on the other thread.
* I will remove the exception of "empty sections", so: always two blank lines.
* As for spacing rules for header comments, I'm fine with yours as well, so let's hear some other voices.

**Edit:** And I will see what I can copy from `help write theorem_list`

benjub:
A few corrections to take into account the comments at #2262 from @nmegill and @avekens. Is there a way to ping all metamath contributors ?

benjub:
There is also the case of the Begin/End commands that split the database, e.g.,
```
    EUAABCLMNNOPQ $.

$( End $[ set-pred.mm $] $)


$( Begin $[ set-class.mm $] $)

$(
###############################################################################
```
I think this is good spacing (one could even remove the blank line after the "Begin" line), in which case I should write in the conventions:
> The file set.mm should have a double blank line before each section header, and at no other places (if there is a $( Begin $[...$] $) comment before the section header, then the double blank line should go before that comment).

avekens:
I agree with @benjub that the double blank lines should go before the $( Begin $[...$] $) comment, and that there should be no blank line between the $( Begin $[...$] $) comment and the following header.

nmegill:
I checked the code, and what I described (qualified with "IIRC") was wrong.  The `/rewrap` algorithm for removing blank lines is simply "Don't allow more than 2 consecutive blank lines".  So e.g. 5 blank lines will be changed to 2, but 2 blank lines will never be changed to 1.

I don't currently plan to change this.  (It would actually be somewhat involved since there is currently no header recognition in that part of the code.)  So the blank line convention will remain a "suggestion" and not a requirement.  That will give us flexibility to adjust the double-blank-line convention to whatever we want.

If someone wants to go through set.mm manually to make it conform to what we decide on, and maybe do so every year or so, there shouldn't be too many cases to consider.

> Is there a way to ping all metamath contributors ?

I think most of them follow the Google group.

benjub:
Thanks for this information. My wording "that we try to stick to" makes it clear that these rules are not enforced.  Although this might be too colloquial ? Maybe "that we try to follow" ?
Apart from that, this is ready to merge, at least for me.

avekens:
@benjub Could you enhance the first bullet point as you proposed, please: 
`The file set.mm should have a double blank line before each section header, and at no other places. If there is a $( Begin $[...$] $) comment before the section header, then the double blank line should go before that comment, and there should be no blank line(s) after the "Begin" line and the section header.`

benjub:
@avekens : done, except for the last clause: whether you put zero or one blank line after the "Begin" line is left to one's taste.  I think that's acceptable.

nmegill:
@benjub to fix the CI error, just change "$[...$]" to "$[ ... $]".  If there are spaces, it won't be treated as a bib ref.

benjub:
Sorry for the multiple typos.  Ready to merge for me.

avekens:
Are there really problems with "$( Begin $[ ... $] $)"? In the Metamath book it is written that $( and $) must be surrounded by white space to be regarded as tokens, but we have no whitespace before - "$( - and after - $)" - $( and $). But on "read set.mm". I got the error message "A keyword must be followed by white space." in the second case (with "$( xx $)" and even with "$(xx$)" within a comment). Is there really no way (e.g. escape sequence) to use $( and $) within a comment?

benjub:
> Are there really problems with "$( Begin $[ ... $] $)"? In the Metamath book it is written that $( and $) must be surrounded by white space to be regarded as tokens, but we have no whitespace before - "$( - and after - $)" - $( and $). But on "read set.mm". I got the error message "A keyword must be followed by white space." in the second case (with "$( xx $)" and even with "$(xx$)" within a comment). Is there really no way (e.g. escape sequence) to use $( and $) within a comment?

I think there is a rule of "no nested comments" somewhere, with the suggestion to use @( ... @) instead.

digama0:
No, the metamath specification disallows nested comments to simplify verifiers (and relaxing this rule would break at least the mmamm.m verifier, which uses regexes to strip comments). One common trick is to replace $ with @ inside comments for quick large commenting.

PR #2264, opened 2021-10-24
iset/mathbox: add ax-ddkcomp

benjub:
Add iset/ax-ddkcomp to my mathbox, to keep a record from the discussion at #2242 with @digama0 and @jkingdon.  Hopefully, the plan exposed in the comment is what you had in mind.  I wrote "Contributed by BJ" since it's currently in my mathbox, but I encourage anyone who carries out the plan to move it to Main and update the "contributed by" field.


jkingdon:
> Add iset/ax-ddkcomp to my mathbox, to keep a record from the discussion at #2242 with @digama0 and @jkingdon. Hopefully, the plan exposed in the comment is what you had in mind. I wrote "Contributed by BJ" since it's currently in my mathbox, but I encourage anyone who carries out the plan to move it to Main and update the "contributed by" field.

Agreed, having someone move it if/when they work on it seems sensible. I'm not as worried about the location as about whether there is progress on proving any of this, 

As for the statement of the axiom, wish I could advise more on this. I mean, I can look at pages like https://iep.utm.edu/con-math/ (where they talk about "constructive least upper bound principle") but I haven't really tried to prove any results along these lines.

For one small example, the sources I quickly found which talk about constructive least upper bound principles do talk about "nonempty" when I would assume they mean "inhabited" (I see you currently have `A =/= (/)` in ax-ddkcomp when I would have imagined `E. x x e. A`). But I guess I don't know that for sure. As a formal math project, surely writing metamath proofs is a big part of how we answer such questions. And inhabited versus non-empty is presumably kind of trivial compared with how order-locatedness is expressed and the like.

As for whether to prove http://us.metamath.org/ileuni/ax-caucvg.html in terms of least upper bounds, uh, maybe? The sources I was consulting don't seem to do that (instead they use Dedekind cuts more fully or just assume Cauchy convergence as an axiom without discussing how to prove it) but I'm not sure I object if we can find a proof. I'm also a bit confused, perhaps relatedly, about some of the ideas about one sided cuts versus two sided cuts (some of the sources seem to say that you need two sided cuts if you don't have excluded middle but I don't know whether that is a hard "need" or more of a "we like it better" or anything in between).

Having said all that, thanks for adding this to a mathbox. Even if it is work in progress, always good to have what we have so far expressed in a formal way.

benjub:
You're right, I'll change "non-empty" to "inhabited".  Regarding lower versus two-sided, I thought that in this case, they are equivalent, by @digama's argument of defining the exterior of A, that is, \{ x \in \RR \mid \exists y \in \RR \setminus A y < x \}.

As for the proof of Cauchy completeness from Dedekind completeness, I am stating the obvious sketch for the record: given a real sequence (a_n), define A_n := ]-\infty, a_n[, then B_n := \cup_{k \geq n} A_k and B := \cap_{n \in \NN} B_n.  Then prove that B is located. If it is bounded, then it has a supremum M.  Similarly define m. If (a_n) is Cauchy, then it is bounded so m and M exist, and m=M, and it is the limit. Of course all the difficulty is in the details...

benjub:
> You're right, I'll change "non-empty" to "inhabited".

Done in #2268.

PR #2267, opened 2021-10-25
Reduce axiom usage of ssrel, relopabi, cnv0

aaay-aaay:
This is my first time contributing, please let me know if I've missed anything.

The new proofs I've created of ssrel, relopabi and cnv0 all avoid ax-sep, ax-nul, ax-pr, and (for some reason; this was completely unintentional) ax-9.

The proof of relopabi is quite long, I think someone with more experience with Metamath than me could probably shorten it. For now I've made the previous proof "ALT" rather than "OLD" because it's a lot shorter (74 vs 700 steps, 14 vs 38 essential steps).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14752,6 +14752,7 @@ New usage of "cnnvm" is discouraged (1 uses).
 New usage of "cnnvnm" is discouraged (3 uses).
 New usage of "cnnvs" is discouraged (2 uses).
 New usage of "cnopc" is discouraged (1 uses).
+New usage of "cnv0OLD" is discouraged (0 uses).
 New usage of "cnvadj" is discouraged (3 uses).
 New usage of "cnvbrabra" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
Nice result for a first time commit! The tag would be something like "Reduce axiom usage.  (Revised by KP, 25-Oct-2021.)"

benjub:
> Nice result for a first time commit! The tag would be something like "Reduce axiom usage. (Revised by KP, 25-Oct-2021.)"

And if you want to be more precise: "Remove dependency on ~ ax-sep , ~  ax-nul , ~ ax-pr.  (Revised by KP, 25-Oct-2021.)" And congratulations !

As for removal of the dependency on ax-9, this is actually an artefact of the definition df-cleq. See ~bj-ax9 and ~bj-df-cleq.

benjub:
Add spaces between ax-sep and the comma following it, and similarly between ax-pr and the period following it (sorry, I did not do it myself in my previous comment...).

PR #2271, opened 2021-10-28
set.mm - prove dfcleq with axext3, modify df-cleq comment

nmegill:
See discussion at
https://groups.google.com/g/metamath/c/xdHB0oW0aZ4/m/L1ng44caAQAJ

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18721,6 +18721,7 @@ Proof modification of "decsuccOLD" is discouraged (44 steps).
 Proof modification of "dedtOLD" is discouraged (23 steps).
 Proof modification of "demoivreALT" is discouraged (1087 steps).
 Proof modification of "dfbi1ALT" is discouraged (100 steps).
+Proof modification of "dfcleq" is discouraged (10 steps).
 Proof modification of "dfdecOLD" is discouraged (35 steps).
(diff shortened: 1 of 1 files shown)


avekens:
Since df-cleq is only used by dfcleq (at the moment), and should not be used directly as I understood the discussion in the Google group, why not tag it with  "(New usage is discouraged.)"and add "Use ~ dfcleq instead."?

nmegill:
> Since df-cleq is only used by dfcleq (at the moment), and should not be used directly as I understood the discussion in the Google group, why not tag it with "(New usage is discouraged.)"and add "Use ~ dfcleq instead."?

OK, go ahead and do that if you wish.

nmegill:
> > Since df-cleq is only used by dfcleq (at the moment), and should not be used directly as I understood the discussion in the Google group, why not tag it with "(New usage is discouraged.)"and add "Use ~ dfcleq instead."?
> 
> OK, go ahead and do that if you wish.

Actually. let's wait until we have an agreed-upon version of df-cleq.  After we revert the dfcleq proof hack so that it no longer inserts ax-9 usage, direct use of df-cleq in a theorem will be minimized out anyway, so maybe it's not necessary to discourage its use.

PR #2268, opened 2021-10-25
Misc

benjub:
As for shortenings using syldc (41 proofs for a theorem in minimal implicational calculus!), here is the output:
Proof of "0enwwlksnge1" decreased from 896 to 889 bytes using "syldc".
Proof of "2ndcctbss" decreased from 4336 to 4319 bytes using "syldc".
Proof of "cfsmolem" decreased from 5115 to 5105 bytes using "syldc".
Proof of "clwlkclwwlklem2" decreased from 2864 to 2843 bytes using "syldc".
Proof of "clwlkisclwwlklem1" decreased from 2858 to 2837 bytes using "syldc".
Proof of "clwlksfclwwlk" decreased from 4643 to 4633 bytes using "syldc".
Proof of "conngrv2edg" decreased from 900 to 887 bytes using "syldc".
Proof of "cply1mul" decreased from 2135 to 2124 bytes using "syldc".
Proof of "cusgrafi" decreased from 702 to 694 bytes using "syldc".
Proof of "cusgrasize2inds" decreased from 1213 to 1183 bytes using "syldc".
Proof of "cusgrsize2inds" decreased from 1197 to 1167 bytes using "syldc".
Proof of "dfac21" decreased from 1288 to 1263 bytes using "syldc".
Proof of "dfgrp3lem" decreased from 1824 to 1817 bytes using "syldc".
Proof of "evengpoap3" decreased from 947 to 933 bytes using "syldc".
Proof of "evengpop3" decreased from 896 to 882 bytes using "syldc".
Proof of "fgcfil" decreased from 633 to 626 bytes using "syldc".
Proof of "flnn0div2ge" decreased from 662 to 649 bytes using "syldc".
Proof of "frrlem11" decreased from 1200 to 1185 bytes using "syldc".
Proof of "fzen" decreased from 1594 to 1586 bytes using "syldc".
Proof of "genpnnp" decreased from 1224 to 1216 bytes using "syldc".
Proof of "hashge2el2dif" decreased from 1200 to 1161 bytes using "syldc".
Proof of "inf3lem3" decreased from 715 to 708 bytes using "syldc".
Proof of "initoeu1" decreased from 1067 to 1053 bytes using "syldc".
Proof of "lcmf" decreased from 929 to 922 bytes using "syldc".
Proof of "lindslinindsimp1" decreased from 2991 to 2980 bytes using "syldc".
Proof of "lindslinindsimp2lem5" decreased from 2282 to 2272 bytes using "syldc".
[...]

PR #2277, opened 2021-10-30
automatic minimizations of proofs that 'acquired' dependency on ax-9

benjub:
To prove that the question of df-cleq discussed on the google group has concrete consequences:

I extracted from the previous version of set.mm theorems that depended on df-cleq but not on ax-9 (I did a "diff" of the outputs of "show usage df-cleq/recursive" and "show usage ax-9/recursive). There were 4552 of these. So there are now 4552 html pages (times 2 if we count mpeuni+mpegif) which correctly indicate dependency on ax-9. Correcting the internet one page at a time ;)

Then, I ran the minimization script on that list and could minimize 143 proofs.  To be fair, many minimizations were not connected to ax-9 and would have happened anyway, but in at least (exactly?) five cases, uses of eleq2 were replaced by uses of elequ2 (aceq1, aceq2, axsep2, kmlem4, restlly): tiny steps, but definitely towards simplification. In the case of axsep2, this even removed dependencies on ax-ext, df-cleq and df-clel ! (although this was known, and one can also remove dependencies on ax-12 and ax-13, see ~bj-axsep2).

A few coments:
* @glacode , your dfcleqf appears to be a duplicate of cleqf. I hope the minimization is ok with you (your proof was very similar to that of cleqh, so I think there is no reason to keep it as ALT).
* Many minimizations appear to be of theorems proved using mmj2.  If this is not too hard with your workflow, then try to minimize your new proofs with Metamath's "minimize_with *" commands before doing a PR.
* @avekens : as you noted, eupth2lem1 and eupth2lem2 are duplicates of eupath2lem1 and eupath2lem2, and the proofs are identical.  They got minimized since they had no proof modif discouragement tag.  I hope this is ok (but reverting it is trivial by manual copy-pasting).
* @avekens 's numclwwlkdisj was greatly minimized, I assumed it's ok since there is no proof modif discouragement tag. Same with srhmsubcALTVlem1.

Also pinging @tirix since many of his proofs were minimized, and @nmegill , @digama0 , @wlammen since they were involved in the google group discussion.


nmegill:
@avekens Per your comment https://github.com/metamath/set.mm/pull/2268#issuecomment-951265255 will these interfere with what you are doing?

avekens:
Unfortunately, the proofs of many theorems about graph theory (in main set.mm and my mathbox) I am currently working on are minimized, which will cause merge conflicts when I will provide my changes. I hope they will not be too many and too difficult to solve, therefore this PR can be accepted. But please do not perform any global minimization anymore until I finished the revision  of GraphTheory. 

```
@avekens : as you noted, eupth2lem1 and eupth2lem2 are duplicates of eupath2lem1 and eupath2lem2, and the proofs are identical. They got minimized since they had no proof modif discouragement tag. I hope this is ok (but reverting it is trivial by manual copy-pasting).
```
This will be cleaned up by my revision.
```
@avekens 's numclwwlkdisj was greatly minimized, I assumed it's ok since there is no proof modif discouragement tag. Same with srhmsubcALTVlem1.
```
minimization of numclwwlkdisj is OK. As for srhmsubcALTVlem1, I think I forgot to add "(Proof modification is discouraged.)" - I will revert and correct this later.

benjub:
Unrelated: @nmegill , I noticed that trintss and trint0 are very minor variations of each other (through ancoms), so I propose to keep only the proof of trint0 (shorter and uses only one dummy variable) but use the label "trintss", merge the comments and credits, and deprecate the current trintss to trintssOLD (the proof is not worth being an ALT, in my opinion). Do you agree ?

nmegill:
> Unrelated: @nmegill , I noticed that trintss and trint0 are very minor variations of each other (through ancoms), so I propose to keep only the proof of trint0 (shorter and uses only one dummy variable) but use the label "trintss", merge the comments and credits, and deprecate the current trintss to trintssOLD (the proof is not worth being an ALT, in my opinion). Do you agree ?

I agree.  The description for trint0 should be used since it is concise and has a reference.  Due to the date precedence, I would use for the credit "(Contributed by Scott Fenton, 3-Mar-2011.) (Proof shortened by by Andrew Salmon, 14-Nov-2011.)", even though AS probably proved it independently rather than as a conscious shortening.  And I agree the other proof isn't worth keeping as an ALT.

PR #2278, opened 2021-10-30
Some theorems moved from mathboxes, to the main part of set.mm

glacode:
Some theorems moved from mathboxes, to the main part of set.mm (my upcoming PR will use those theorems)

I've done a distinct PR to let Norm evaluate the theorems moved to the main part.


nmegill:
@glacode It seems a merge conflict arose.

PR #2279, opened 2021-10-30
merge trint0/trintss

benjub:
As agreed in https://github.com/metamath/set.mm/pull/2277#issuecomment-955282870
I did it also for iset.mm, so letting @jkingdon know.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17953,6 +17953,7 @@ New usage of "trclubiOLD" is discouraged (0 uses).
 New usage of "trelded" is discouraged (1 uses).
 New usage of "trintALT" is discouraged (0 uses).
 New usage of "trintALTVD" is discouraged (0 uses).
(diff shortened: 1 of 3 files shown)


nmegill:
The reference is incorrect; it should be Exercise 3 in [TakeutiZaring] p. 44.  However, in their book the exercise does not include the A =/= (/) condition, which is obviously wrong!  Some authors (Suppes) define |^| (/) = (/) as a special case, others leave it undefined (Enderton, who compares it to 1/0), but as their Exercise 2 shows, in T&K's version |^| (/) = _V like ours.

jkingdon:
Thanks for doing this, and for fixing it in both iset.mm and set.mm.

As you might have surmised from the mmill.html entries, I had noticed the similarity when I first intuitionized these, but hadn't gotten around to doing anything about it. The solutions in this pull request look good to me, so thanks to both @benjub and @nmegill .

PR #2280, opened 2021-10-31
The superior limit of a countable set of sigma-measurable functions is sigma-measurable

glacode:
- partially defined functions are allowed

- a number of theorems about ` limsup ` (the superior limit) have been added


savask:
@glacode Great work, as always!

As a small suggestion, perhaps you could run minimize on the new theorems (at least on short ones)? For example, your `ralimda` seems to be identical to [ralimdaa](http://us2.metamath.org/mpeuni/ralimdaa.html) from the main set.mm, I bet minimize would detect some other duplicates.

avekens:
What about ~mptfnd? It's comment contains "(Contributed by NM, 9-Apr-2013.) (Revised by Thierry Arnoux, 10-May-2017.) ", which does not seem correct.

glacode:
I expect to do some cleanup of my mathbox after this PR (I have a list of TODOs; I add ~mptfnd comment checking to the list)

W.r.t. to minimize, I have to learn how to do it and how to insert it into my workflow (I always have a long list of .mmt files that I use to create / append / insert new proofs to set.mm in a single script, just before the PR; but some handwork is required)

PR #2281, opened 2021-10-31
replace "PART 16 GRAPH THEORY" - step 4

avekens:
Task 3 (part 2) of issue #2265: Sections of "Graph theory (revised)" moved from AV's mathbox to main set.mm:

* Trails
* Paths

Additional changes:
* moved from BJ's mathbox: ~sylancl3 (TODO: shorten proofs)
* moved from AV's mathbox: ~rexdifpr, ~pr1eqbg, ~pr1nebg, ~opab0, ~nprrel12, ~brfvopabrbr, ~2f1fvneq, ~prinfzo0, ~elfzo0l, ~elfzr, ~elfzlmr, ~elfz0lmr
* ~fvmptopab1/~fvmptopab2 replaced by fvmptopab
* theorems of the form ( * ` G )= { <. f , p >. | * } and ( F ( * ` G ) P <-> revised by removing not required antecedents  F e. Y , P e. Z and (sometimes even) G e. X.

PR #2284, opened 2021-11-01
Misc

avekens:
* comment of ~nprrel12 corrected as indicated by Gérard Lang
* 9 proofs using "sylancl" shortened by using "sylancl3" as indicated in the comment of ~sylancl3:  ~1stmbfm, ~2ndmbfm, ~dfac14, ~fnwelem, ~tfrlem10, ~gruina, ~rfcnpre1, ~umgra1, ~uslgra1 (@benjub you wrote about 11 proofs to be shortened: ~fourierdlem103 could be shortened, too, but it would take too long on my machine; some of the ~fourier* theorems using ~sylancl I did not check, because the proofs of these theorems are too huge).
* section "Walks, paths and cycles" and subsections lifted up by one level
* @glacode double space before end of comment reduced to one space (and other changes caused by write source set.mm/rewrap), see discussion in PR #2281


benjub:
@avekens Thanks.  I'm not sure which theorems I tried to minimize with sylancl3.  Probably, I used "scripts/min.cmd" on the theorems using theorems used in the proof of sylancl3. Quite often, I remove from 1.tmp the proofs of fourierlem* stoweilem* stirlinglem* (or something like that), since many of @glacode's proofs are very long and take time to minimize.

Issue #2265, opened 2021-10-25
Revision of GRAPH THEORY - Part 2


avekens:
According to issue #1418, the definitions and theorems for Graph Theory should be revised. The revised version is already available in AV's mathbox. The next step, tracked by this issue,  will be to replace "PART 16 GRAPH THEORY" entirely by section "21.34.8 Graph theory (revised)" of AV's mathbox. Some adjustments of names, labels and comments have to be done during this action.

At first, a new "PART 16 GRAPH THEORY (UNDER CONSTRUCTION)" is created before the current part, which becomes "PART 17 GRAPH THEORY (DEPRECATED)". The revised sections will be moved from AV's mathbox into this new part step by step, resolving all broken dependencies. Minor adjustments will be performed. See the following task list. Finally, the deprecated part will be removed.

**IT IS RECOMMENDED NOT TO WORK WITH DEFINITIONS AND THEOREMS FOR GRAPH THEORY UNTIL THIS ISSUE IS DONE!**

- [x] Move sections up to and including "Edges as subsets of vertices of graphs" to main set.mm
- [x] Move sections up to and including "Regular graphs" to main set.mm
- [x] Move sections up to and including "Connected graphs" to main set.mm
- [x] Move sections up to and including "The Königsberg Bridge problem" to main set.mm
- [x] Move sections up to and including "Huneke's Proof of the Friendship Theorem " to main set.mm
- [x] Remove previous/deprecated part GRAPH THEORY
- [x] Adjust of names, labels and comments

avekens:
With PR #2302, the revision of PART 16 GRAPH THEORY is finished. I'll leave this issue open for a while - if there are any comments and suggestions to improve this part, please add them here. I plan to perform some additional finetunings in the next time.

avekens:
No relevant comments/suggestions until now, so this issue can be closed.

PR #2285, opened 2021-11-02
replace "PART 16 GRAPH THEORY" - step 6

avekens:
Task 3 (final part 4) of issue #2265: Sections of "Graph theory (revised)" moved from AV's mathbox to main set.mm:

* Closed walks as words
* Examples for walks, trails and paths
* Connected graphs

Additional changes:
* moved from AV's mathbox: ~fpropnf1


avekens:
Again, the diff used here in github cannot resolve the modifications. Therefore I attach the diff obtained by

git diff 00ec5d9f 112166e --patience > changelog.txt

(SHA of the latest commit in main branch: 00ec5d9f; SHA of my commit 112166e)

here: [changelog.txt](https://github.com/metamath/set.mm/files/7459345/changelog.txt)

@glacode It seems that my latest "save proof */compressed/fast" reformatted the proofs of some of your recently added theorems.

glacode:
> @glacode It seems that my latest "save proof */compressed/fast" reformatted the proofs of some of your recently added theorems.

@avekens do you have any idea why commit #2284 didn't apply "all" the reformatting?

I'm experiencing some "unexpected" behavior with the reformatting script (if I run it twice, sometimes the second time it applies many changes, I guess it shouldn't)

@nmegill @benjub was metamath.exe modified, in that part, with the latest releases?

@david-a-wheeler I'd consider an additional CI control: reformat set.mm and save as set.mm1 ; diff set.mm with set.mm1 and reject the merge request if the diff is not empty (but it may be the topic was already discussed, and rejected, in the forum)

Thanks in advance
Glauco

avekens:
@glacode the explanation is simple:
* for #2284, I executed write source set.mm/rewrap only, because I did not modify any proof
* for #2285, I executed both write source set.mm/rewrap and save proof */compressed/fast
* 
I am working with metamath.exe - Version 0.198 since August 2021, which is the official current version.

david-a-wheeler:
@glacode - it would be easy to add a check that merges are correctly formatted. We've resisted that in the past because some contributors don't understand how to do it.

That said, maybe it's time to reconsider. Maybe post on the Metamath mailing list for discussion? It'd be easy to *implement*.

nmegill:
> @avekens do you have any idea why commit #2284 didn't apply "all" the reformatting?
> 
> I'm experiencing some "unexpected" behavior with the reformatting script (if I run it twice, sometimes the second time it applies many changes, I guess it shouldn't)

This should never happen. I haven't seen this behavior myself.  If you can show a reproducible test case, that would help.

> 
> @nmegill @benjub was metamath.exe modified, in that part, with the latest releases?

The current official version is 0.198 7-Aug-2021.  Make sure that is the only version you are running.  The version is shown as soon as you enter the program.

Version 0.198 fixes a /rewrap bug that existed in prior versions that would sometimes wrap a line prematurely when a double space occurred after a period. As a result of the fix, the output of "write source.../rewrap" in 0.198 will result in many small changes to set.mm compared to the output of the previous versions.

Please check again that all of your metamath copies are at version 0.198.  Could your $PATH be picking up an earlier version in another directory?  Do "metamath" and "./metamath" give different version numbers?

> 
> @david-a-wheeler I'd consider an additional CI control: reformat set.mm and save as set.mm1 ; diff set.mm with set.mm1 and reject the merge request if the diff is not empty (but it may be the topic was already discussed, and rejected, in the forum)

This shouldn't be necessary.  If running the reformatting twice gives different results, that is a serious bug that will need to be fixed.

glacode:
> > @david-a-wheeler I'd consider an additional CI control: reformat set.mm and save as set.mm1 ; diff set.mm with set.mm1 and reject the merge request if the diff is not empty (but it may be the topic was already discussed, and rejected, in the forum)
> 
> This shouldn't be necessary. If running the reformatting twice gives different results, that is a serious bug that will need to be fixed.

Norm, the suggested control is just to ensure that the contributor did a reformatting before the PR (it is not intended to deal with the reformatting twice, giving different results: I see if I can find a reproducible example; I'm on 0.198)

nmegill:
> > > @david-a-wheeler I'd consider an additional CI control: reformat set.mm and save as set.mm1 ; diff set.mm with set.mm1 and reject the merge request if the diff is not empty (but it may be the topic was already discussed, and rejected, in the forum)
> > 
> > 
> > This shouldn't be necessary. If running the reformatting twice gives different results, that is a serious bug that will need to be fixed.
> 
> Norm, the suggested control is just to ensure that the contributor did a reformatting before the PR (it is not intended to deal with the reformatting twice, giving different results: I see if I can find a reproducible example; I'm on 0.198)

OK, this may be a good idea, and also will detect the wrong metamath.exe version if there is a change in the rewrapping algorithm.  There is another rare rewrapping issue (an extra space at beginning of line) that will be fixed in a forthcoming version; it affects only 4 lines in set.mm, but I can see a lot of flip-flopping back and forth on those 4 lines until everyone updates to the new version.

One concern is adding to the run time of the CI.  Unfortunately `/rewrap` is very slow.  I'm hoping to double the speed in a future version, but it will take time to implement.

benjub:
As for the rewrapping: if I recall correctly, there is still an exception that forbids rewrapping of discouragement tags ?  I think this exception can be removed.

As for the run time: the regeneration of the "discouraged" file is also pretty slow.  I think you had ideas to improve it too ?

nmegill:
> As for the rewrapping: if I recall correctly, there is still an exception that forbids rewrapping of discouragement tags ? I think this exception can be removed.

I prevented breaking them across lines so that humans (or a simple script) can search for them more easily, fearing that line breaks might be an inconvenience since  the two "(... is discouraged.)" tags are really built-in fixed "keywords" rather than informal English phrases.  I don't know if anyone has actually used this "feature".  It is easy to remove this exception from the program if that's preferred.  Personally I don't have a preference one way or the other.  We would want to make sure that mmj2 tolerates line breaks in those tags.  It will result in a large number of line changes in set.mm, so if we remove it, it might be best to wait after @glacode's CI suggestion is implemented.

> 
> As for the run time: the regeneration of the "discouraged" file is also pretty slow. I think you had ideas to improve it too ?

That is on my todo list.  I put it at a lower priority since mmj2 already has a much faster algorithm for it.  I can't recall, isn't mmj2 used for that in the CI script?

benjub:
No strong opinion about the rewrapping either.  Generally speaking, fewer exceptions sounds better, but it's true that the tags can be considered as "single keywords".

As for "dicouraged", I do not know how the CI works. As for me, I very rarely use mmj2: I use mostly Metamath, so before each PR, I regenerate the "discourage" file(s) using the recommendation at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md .  But the inconvenience is very minor.

glacode:
I also regenerate the "discourage" file, as suggested in the contributing page, and it is rather slow.

But (locally) the whole /rewrap script doesn't seem slow to me (I'm not sure if the GitHub action will be much slower, due to resource constraints).
@david-a-wheeler , do you expect a noticeable impact on the automatic check performance?

glacode:
As a further note, the slowest check is always the mmj2 job. If the formatting check could be executed "in parallel" it may be it would have no impact at all (don't quote me on that... :-) )

PR #2289, opened 2021-11-02
Add scripts: check-braces and check-dvs.py

jamesjer:
I have described these two scripts on Google Groups.  I thought perhaps they should be named with a check prefix, since they don't modify anything, just check whether useless braces or dv conditions are present.  I also added an MIT license to the top of each, since many of the other scripts carry that license.


jamesjer:
I have been going over the output of check-dvs.py on set.mm and found another bug in the handling of $e statements.  The most recent push fixes that bug.

PR #2291, opened 2021-11-03
Misc

benjub:
In addition to commit messages: I cleaned up my mathbox's "FF/" theorems, now that the definition df-nf is official.  I deleted some of the theorems, and moved others to Main (nfimt, nfbiit, nfnbi) to shorten other proofs.  (letting @wlammen know)


nmegill:
Why is 19.38a repeated as 19.38b?

benjub:
> Why is 19.38a repeated as 19.38b?

19.38a has antecedent F/ x ph while 19.38b has antecedent F/x ps.  I considered grouping both theorems as
```
|- ( ( F/ x ph \/ F/ x ps ) -> ...
```

but this looks less usable.

PR #2294, opened 2021-11-04
Remove dvs for nonexistent variables, part 1

jamesjer:
This commit removes dv conditions identified by scripts/check-dvs.py as referring to variables that are not used in any axiom, hypothesis, theorem, or proof in the same scope as the $d statement.  This covers all such dv conditions in the first two major sections of set.mm:
- CLASSICAL FIRST-ORDER LOGIC WITH EQUALITY; and
- ZF (ZERMELO-FRAENKEL) SET THEORY

Note that one definition is included in this commit: df-recs.  The variables named in the associated $d statement appear to have been moved to df-wrecs instead.  I did a definitionCheck with mmj2, and it reported no problems.


benjub:
That's very nice, especially when for some theorems, no dv's remain and you can remove one nesting level !  What is your roadmap ?  The whole of set.mm and then the other .mm files ?  And then the same process with extraneous braces ?

avekens:
I wonder why there are such a lot of superfluous dv's. I spotted about 3 proof provided by me, and I expected that because sometimes I am not careful enough (although I almost always try to be careful with dv's). But the many others? Is it because of automatic minimization of proofs? Or because of revisions of theorems used in the proofs? Or just copy&paste issues?

benjub:
@avekens : I was thinking of new proofs (automatic or manual) that use fewer dummy variables, but the $d-conditions are not updated.  And maybe, as you suggest, also some savage copypasting.

nmegill:
I'll mention another situation that could in principle be optimized for less file space and in some cases better readability.

There is no reason not to make all dummy variables distinct from each other, even if the proof itself doesn't strictly require it.  For example, if w and v are dummy variables that must be distinct from A but not necessarily from each other, mmj2 tends to  create "$d w A $. $d v A $.".  I have sometimes manually changed this to the shorter "$d w v A $." since we don't care about redundant dv's between dummy variables.

I suppose this could be automated to some extent, although I'm not sure if it is worth the effort.

BTW in this example, all occurrences of v in the proof could be changed to w in the proof so that we could have just "$d w A $.".  However, we probably don't want to do that in general since the different names may enhance readability of the proof.

benjub:
In that case, we lose the information of which dummy variable is really required to be disjoint from which other (dummy or not) variable.  I cannot think, however, of an application where that information would prove important.

For your second point: I think there has been a discussion on that point: either minimizing the number of dummy variables, or on the contrary having different ones, and I tend to agree with you for readability reasons.

nmegill:
> In that case, we lose the information of which dummy variable is really required to be disjoint from which other (dummy or not) variable. I cannot think, however, of an application where that information would prove important.

If by original you mean the minimal number of $d pairs (as opposed to minimal number of compound $d statements), it be recovered using mmj2, so the information isn't lost, just hidden.

jamesjer:
Sorry for the silence @benjub .  I had to take an unexpected trip.  Yes, the roadmap is to work through the rest of set.mm, then do the other .mm files.  Then start over with unnecessary braces.  I'm uncertain whether I should touch other people's mathboxes, though.  What do you all think?  Should I limit my attention to the main part of set.mm?

nmegill:
> I'm uncertain whether I should touch other people's mathboxes, though. What do you all think? Should I limit my attention to the main part of set.mm?

I think it should  be fine to do this.  You may want to include the mathbox owners "@" handle (when known) in the PR comment so they can be alerted.

In the ~mathbox comment, "occasionally other people may make maintenance edits to your mathbox  for things like keeping it synchronized with the rest of set.mm,  reducing proof lengths, moving your theorems to the main part of set.mm  when needed, and fixing typos or other errors."  I would include removing redundant dv's and braces in that list.

If a mathbox owner is obviously in the middle of a project (as indicated by e.g. many recent PRs), as a practical matter you may want to coordinate with them before making changes in order to avoid merge conflicts; just use common sense.

benjub:
I just cleaned my mathbox using your two scripts in #2307.

jamesjer:
Okay, that makes sense.  I think it will be a little while before I get to the mathboxes anyway, and hopefully others will do as @benjub has done in the meantime.

PR #2301, opened 2021-11-06
set.mm - recompress proofs

nmegill:
I did "save proof */compressed" without "/fast" to do a full re-compression, rather than just reformatting into the compressed format.  We don't do this all the time because it takes a long time to run, but it is good to do it occasionally.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18260,7 +18260,7 @@ Proof modification of "alrimdOLD" is discouraged (13 steps).
 Proof modification of "alrimddOLD" is discouraged (21 steps).
 Proof modification of "alrimiOLD" is discouraged (9 steps).
 Proof modification of "altgsumbcALT" is discouraged (1038 steps).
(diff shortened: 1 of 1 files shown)


benjub:
> I did "save proof */compressed" without "/fast" to do a full re-compression

What can it do, concretely, that /fast does not ?  It seemed to me that there was only one way to "compress" a given proof ? Apparently, three proofs (out of ~40000) were "wrongly compressed" and the recompression shortened them greatly ?

nmegill:
> > I did "save proof */compressed" without "/fast" to do a full re-compression
> 
> What can it do, concretely, that /fast does not ? It seemed to me that there was only one way to "compress" a given proof ? Apparently, three proofs (out of ~40000) were "wrongly compressed" and the recompression shortened them greatly ?

See https://groups.google.com/d/msg/metamath/7uJBdCd9tbc/r5iRJifpAgAJ

PR #2303, opened 2021-11-07
Fixing 4 items in #2128 (duplicate therems)

glacode:
- ~ issmfltle has been removed from GS's mathbox (it's a duplicate of ~ smfpreimale , but it has less dv constraints; ~ issmfltle was not referenced)

- ~ vonval2 has been removed from GS's mathbox (references to ~ vonval2 have been changed to ~ mblvon )

- ~ rabid3 has been removed from GS's mathbox (references to ~ rabid3 have been changed to ~ rabeq2i )

- ~ fnvinran has been removed from GS's mathbox (references to ~ fnvinran have been changed to ~ ffvelrnda )

@savask I don't know if I had a way to automatically fix the single items in #2128 (I've read how to mark as fixed the whole issue, but this is not the case)


savask:
> I don't know if I had a way to automatically fix the single items in #2128 (I've read how to mark as fixed the whole issue, but this is not the case)

@glacode I don't think you can, but I'll fill corresponding checkboxes once this pull request gets merged.

PR #2310, opened 2021-11-11
con-->conn for "connected"

benjub:
@digama0 : is there a reason why df-sconn has a discouragement tag ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4830,7 +4830,7 @@
 "df-rngo" is used by "relrngo".
 "df-rq" is used by "dmrecnq".
 "df-rq" is used by "recmulnq".
-"df-scon" is used by "isscon".
+"df-sconn" is used by "issconn".
 "df-setrecs" is used by "nfsetrecs".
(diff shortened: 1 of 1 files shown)


digama0:
@benjub No idea, I don't have any memory of putting that tag on, and there isn't any other definition of simply connected that we are trying to migrate toward. Feel free to remove the tag.

nmegill:
- When the htmldef uses a GIF, it no longer matches the token name for the changed symbols.  I would recommend using the althtmldef information for the htmldef in these cases (i.e. abandon the GIF).


- We have been putting token changes in the label change list, like this example:
```
26-Apr-20 ---       ---         math token "(++)" changed to "/_\"
```

- The token name is misspelled here:
```
latexdef "SConn" as "{\rm SConnn}";
```

nmegill:
(Due to the size of the changes, I'm merging this now to avoid merge conflicts in the queue; you can fix these in a separate PR.)

PR #2311, opened 2021-11-12
Remove dvs for nonexistent variables, part 3

jamesjer:
This commit removes dv conditions identified by scripts/check-dvs.py as referring to variables that are not used in any axiom, hypothesis, theorem, or proof in the same scope as the $d statement. This covers all such dv conditions in the sixth through ninth major sections of set.mm:
- ELEMENTARY NUMBER THEORY
- BASIC STRUCTURES
- BASIC CATEGORY THEORY
- BASIC ORDER THEORY

- df-sad
- df-smu
- df-gcd
- df-0g and df-gsum
- df-ordt and df-xrs
- df-cat through df-comf
- df-nat and df-fuc
- df-evlf through df-diag
- df-lub through df-meet

For each, I did a visual inspection to verify that the variables in question really do not appear. I also checked all of set.mm with mmj2's definitionCheck.


david-a-wheeler:
Thanks for doing this! More than I expected.

Perhaps we should check for non-existent variables on check-in, once all current ones are fixed.

jamesjer:
I think we humans are not very good at noticing such things, so if we wish to keep the databases free of nonexistent variables, enlisting the aid of some software is necessary.  Doing so at check-in time makes sense to me.

nmegill:
> Perhaps we should check for non-existent variables on check-in, once all current ones are fixed.

Of course I am very happy with what @jamesjer is doing.  My concern is that some users may not be able to run the script on platforms like Windows, and in any case it is one more thing they will have to remember to run.

Redundant dv's waste space in set.mm and maybe detract a little from its human readability, so it is certainly good to clean them out from time to time.  But they don't violate the Metamath spec, and the generated web pages are unaffected by them.  Should they be considered severe enough to block CI from passing?

I'm not necessarily saying we shouldn't do it in CI - maybe that is the best approach - but we should also look at how users might be affected by this.

jamesjer:
Good point.  It would perhaps be less disruptive to users' workflows to run the script periodically and clean up anything it finds.  It probably doesn't need to be run terribly often.

PR #2312, opened 2021-11-12
fix 'connected' changes

benjub:
Fix errors noticed in #2310

Note to @jkingdon : added a comment about jaob being an alias of ax-io and discouraging use of the latter.  I don't know if there are other instances in iset.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4830,7 +4830,6 @@
 "df-rngo" is used by "relrngo".
 "df-rq" is used by "dmrecnq".
 "df-rq" is used by "recmulnq".
-"df-sconn" is used by "issconn".
 "df-setrecs" is used by "nfsetrecs".
 "df-setrecs" is used by "setrec1".
 "df-setrecs" is used by "setrec2".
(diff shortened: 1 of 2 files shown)


jkingdon:
> Note to @jkingdon : added a comment about jaob being an alias of ax-io and discouraging use of the latter.

Looks good, thanks for doing this.

> I don't know if there are other instances in iset.

According to the iset-discouraged change which is part of this pull request, ax-io is now just used by jaob .

benjub:
> > I don't know if there are other instances in iset.
> 
> According to the iset-discouraged change which is part of this pull request, ax-io is now just used by jaob .

I meant: are there other axioms that have an alias, like iset/ax-io or set/ax-3, set/ax-10, etc. (I remember I had "fixed" those), in iset.mm ? (That's not very important, and one can deal with them as one runs into them.)

jkingdon:
> I meant: are there other axioms that have an alias, like iset/ax-io or set/ax-3, set/ax-10, etc. (I remember I had "fixed" those), in iset.mm ? (That's not very important, and one can deal with them as one runs into can deal with them as one runs into them.)

Oh I see. I support doing similar things for those (especially if it achieves other goals like making it easier to copy proofs between set.mm and iset.mm), but I too am OK with "can deal with them as one runs into them".

PR #2316, opened 2021-11-13
Remove dvs for nonexistent variables, part 5

jamesjer:
This commit removes dv conditions identified by scripts/check-dvs.py as referring to variables that are not used in any axiom, hypothesis, theorem, or proof in the same scope as the $d statement. This covers all such dv conditions in the tenth major section of set.mm (BASIC REAL AND COMPLEX ANALYSIS) through the remainder of the main part of set.mm; i.e., everything except the mathboxes.

- df-mbf through df-itg
- df-logb
- df-lgam through df-igam
- df-rprm
- df-grpo through df-gdiv
- df-nv through df-ims
- df-dip
- df-aj and df-hmo
- df-ph
- df-hcau
- df-hosum through df-hfmul
- df-nmop through df-spec
- df-at

For each, I did a visual inspection to verify that the variables in question really do not appear. I also checked all of set.mm with mmj2's definitionCheck.


avekens:
Really impressive! I did not expect such a lot of unnecessary dv conditions in part GRAPH THEORY with theorems created/revised by me (fortunately, the definitions were not concerned). As @david-a-wheeler suggested before, we should provie a mechanism to check for/clean up such unnecessary dv conditions more regularly.

jamesjer:
I think the number of unnecessary dv conditions surprised everybody.  We have a script now that we can run every once in awhile to catch new unnecessary dv conditions.  I don't expect new ones to pile up very rapidly, but we'll see. :-)

PR #2320, opened 2021-11-16
Remove dvs for nonexistent variables, part 7

jamesjer:
This commit removes dv conditions identified by scripts/check-dvs.py as referring to variables that are not used in any axiom, hypothesis, theorem, or proof in the same scope as the $d statement. This covers all such dv conditions in iset.mm.

- ax-coll
- ax-nul
- df-iom
- df-irdg
- df-frec
- df-card
- df-r
- ax-caucvg
- df-inn
- df-uz
- df-fz
- df-fzo
- df-bc
- df-shft
- df-clim
- ax-inf2

For each, I did a visual inspection to verify that the variables in question really do not appear. I also checked all of iset.mm with mmj2's definitionCheck.


jamesjer:
The metamath.exe verifier objected to the dv conditions differing between ax-caucvg and axcaucvg, so I reverted that part.

nmegill:
> The metamath.exe verifier objected to the dv conditions differing between ax-caucvg and axcaucvg, so I reverted that part.

@jkingdon any idea why the dv conditions differ?

jkingdon:
> > The metamath.exe verifier objected to the dv conditions differing between ax-caucvg and axcaucvg, so I reverted that part.
> 
> @jkingdon any idea why the dv conditions differ?

I assume the issue is that http://us.metamath.org/ileuni/axcaucvg.html has "Dummy variables 𝑎 𝑙 𝑢 𝑧 𝑏 𝑐" and these variables (because they are dummies) don't need to appear in http://us.metamath.org/ileuni/ax-caucvg.html . Not sure why @jamesjer was not able to fix this up, but I'm guessing that it had to do with needing to be very careful about whether the non-dummy constraints are the same (and, according the error message anyway, even listed in the same order). I've submitted https://github.com/metamath/set.mm/pull/2321 to deal with this.

nmegill:
I'm still unclear about what the problem was.  The 'verify markup' check of ax-caucvg vs. axcaucvg shouldn't care about dvs on dummy vars used in the proof or the order in which the dvs are declared, and it should ignore any redundant dvs on either ax-caucvg or axcaucvg.  If this isn't the case, that's a bug that would need to be fixed.

jamesjer:
All I know is that the verifier emitted an error message stating that the dv conditions on ax-caucvg and axcaucvg have to be the same, and in the same order.  I didn't (and don't) understand why, so I backed off the change rather than inadvertently break something.

nmegill:
OK, I understand it now.  I looked at the code, and it does require that the non-dummy dv's be in the same order for both ax-caucvg and axcaucvg (as the error message indicates).  While in theory this shouldn't be required, it was a compromise to make the code shorter to write and run faster.  Normally there are very few "ax-" vs. "ax" cases in a .mm file, and I guess I felt this compromise was acceptable.  Anyway @jkingdon has fixed it.

nmegill:
To help reduce future confusion in an "ax-" vs "ax" dv mismatch, I'll add a more detailed explanation to the warning message indicating that it is an artificial compromise made to simplify the program.

I looked at changing the code in "verify markup", and indeed it would be somewhat messy to fix this vs. the present method of just comparing two existing lists with a single line of code.  It doesn't seem worth it for such a rare problem, as long as the warning message is clear.  On the plus side, requiring the same dv order for "ax-" and "ax" can make manual comparison easier.

jkingdon:
>  It doesn't seem worth it for such a rare problem, as long as the warning message is clear.

I agree. I thought the existing message was clear but if you want to explain the reasoning further that's okay with me too.

david-a-wheeler:
Although Norm did this for performance, I think it's sensible anyway.

A "proven axiom" is a special thing. It's reasonable to require the proof to match *precisely* the axiom being proved.

PR #2325, opened 2021-11-25
Surreal number work

sctfn:
Begin work on a better way to separate two sets of surreal numbers.


sctfn:
It's failing at the bibliography step. I'm not sure why

nmegill:
For set.mm, you have:
?Warning: Line number 508021 ends with a space character, which is discouraged.

PR #2326, opened 2021-11-26
Misc

benjub:
See commit messages.
* I shortened 0nelrel "silently", that is, no *OLD version kept and no crediting, since the proof modification is very minor. I hope @nmegill and @avekens are OK with that.
* I added 0nelfun since this is the result that is mentioned in the comment of df-struct. I am fine with attributing it to AV.
* The comment of df-top mentioned that the class `Top` is proper, so I added a reference to that result (bj-topnex), and I propose to move it to Main together with related nearby theorems: bj-funtopon, bj-dmtopon, bj-fntopon, bj-toprntopon, which in my opinion are important in explaining the interplay between `Top` and `TopOn` (I would also add a header comment to the "Topology" subsection to explain it).
* What are the uses for strlemor0,1,2,3 ? strlemor0 is described as a utility lemma but has no usages, and neither do the other ones (except among them). Can we deprecate them ? Maybe @digama0 has an idea about them ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12797,6 +12797,9 @@
 "strlem4" is used by "strlem6".
 "strlem5" is used by "strlem6".
 "strlem6" is used by "stri".
+"strlemor1OLD" is used by "strlemor2OLD".
+"strlemor1OLD" is used by "strlemor3OLD".
+"strlemor2OLD" is used by "strlemor3OLD".
 "structgrssvtxlemOLD" is used by "structgrssiedgOLD".
 "structgrssvtxlemOLD" is used by "structgrssvtxOLD".
 "sumdmdi" is used by "dmdbr4ati".
@@ -14755,6 +14758,7 @@ New usage of "cnvbrabra" is discouraged (2 uses).
 New usage of "cnvbracl" is discouraged (2 uses).
 New usage of "cnvbramul" is discouraged (1 uses).
 New usage of "cnvbraval" is discouraged (1 uses).
+New usage of "cnvcnvOLD" is discouraged (0 uses).
 New usage of "cnvssOLD" is discouraged (0 uses).
 New usage of "cnvunop" is discouraged (2 uses).
 New usage of "com3rgbi" is discouraged (1 uses).
@@ -17902,6 +17906,10 @@ New usage of "strlem3a" is discouraged (1 uses).
 New usage of "strlem4" is discouraged (1 uses).
 New usage of "strlem5" is discouraged (1 uses).
 New usage of "strlem6" is discouraged (1 uses).
+New usage of "strlemor0OLD" is discouraged (0 uses).
+New usage of "strlemor1OLD" is discouraged (2 uses).
+New usage of "strlemor2OLD" is discouraged (1 uses).
+New usage of "strlemor3OLD" is discouraged (0 uses).
 New usage of "structgrssiedgOLD" is discouraged (0 uses).
 New usage of "structgrssvtxOLD" is discouraged (0 uses).
 New usage of "structgrssvtxlemOLD" is discouraged (2 uses).
@@ -18664,6 +18672,7 @@ Proof modification of "cnncvsabsnegdemo" is discouraged (74 steps).
 Proof modification of "cnncvsaddassdemo" is discouraged (46 steps).
 Proof modification of "cnncvsmulassdemo" is discouraged (95 steps).
 Proof modification of "cnv0OLD" is discouraged (40 steps).
+Proof modification of "cnvcnvOLD" is discouraged (86 steps).
 Proof modification of "cnvssOLD" is discouraged (62 steps).
 Proof modification of "com3rgbi" is discouraged (35 steps).
 Proof modification of "compneOLD" is discouraged (89 steps).
@@ -19725,6 +19734,10 @@ Proof modification of "ssuniOLD" is discouraged (84 steps).
 Proof modification of "stdpc5OLD" is discouraged (24 steps).
 Proof modification of "stdpc5OLDOLD" is discouraged (20 steps).
 Proof modification of "stdpc5t" is discouraged (23 steps).
+Proof modification of "strlemor0OLD" is discouraged (26 steps).
+Proof modification of "strlemor1OLD" is discouraged (327 steps).
+Proof modification of "strlemor2OLD" is discouraged (62 steps).
+Proof modification of "strlemor3OLD" is discouraged (73 steps).
 Proof modification of "structgrssiedgOLD" is discouraged (62 steps).
 Proof modification of "structgrssvtxOLD" is discouraged (62 steps).
 Proof modification of "structgrssvtxlemOLD" is discouraged (148 steps).


avekens:
>     I shortened 0nelrel "silently", that is, no *OLD version kept and no crediting, since the proof modification is very minor. I hope @nmegill and @avekens are OK with that.
OK for me.

>     I added 0nelfun since this is the result that is mentioned in the comment of df-struct. I am fine with attributing it to AV.
No problem, leave it as it is.

>     The comment of df-top mentioned that the class Top is proper, so I added a reference to that result (bj-topnex), and I propose to move it to Main together with related nearby theorems: bj-funtopon, bj-dmtopon, bj-fntopon, bj-toprntopon, which in my opinion are important in explaining the interplay between Top and TopOn (I would also add a header comment to the "Topology" subsection to explain it).
I would appreciate that.

>     What are the uses for strlemor0,1,2,3 ? strlemor0 is described as a utility lemma but has no usages, and neither do the other ones (except among them). Can we deprecate them ? Maybe @digama0 has an idea about them ?
 ` Fun `' `' (/) ` looks interesting and should be kept, but `dom (/) C_ ( 1 ... 0 ) ` is trivial/strange. But maybe strlemor0 is just the completion of the series strlemorX, so either all of them should be kept as they are, or be removed all together.

digama0:
> * What are the uses for strlemor0,1,2,3 ? strlemor0 is described as a utility lemma but has no usages, and neither do the other ones (except among them). Can we deprecate them ? Maybe @digama0 has an idea about them ?

strlemor0,1,2,3 is the old set of automation lemmas for proving that a collection of ordered pairs is a structure. They all have the form ``( Fun `' `' F /\ F C_ ( 1 ... N ) )``, and so `df-struct` was created to capture this conjunct instead of just carrying it around through the lemmas. They have been superceded by strle{1,2,3,un}, which use df-struct directly. There is no strle0 because df-struct demands `M <_ N`, so `(/) Struct <. 1 , 0 >.` is not a theorem.

They are utility lemmas, which is to say they aren't particularly interesting on their own. ``Fun `' `' (/)`` is not a theorem we need, that's just the composition of `` `' (/) = (/)`` and `Fun (/)`. (The main reason to use strlemor0 is as a hypothesis to strlemor{1,2,3} - it is the base case of the induction.) So since the automation is using new lemmas now they can probably just be deleted.

benjub:
Thanks @digama0.  I will deprecate them, with a comment inspired by your explanation.

benjub:
By the way: is there a reason why the domain of a structure is restricted that much ? Why not define
>  Struct = { f | ( Fun ( f \ { (/) } ) /\ dom f C_ NN ) }

(with possibly the requirement that the domain be finite) ?

digama0:
@benjub (I should write a post about automation lemma design at some point.) What you wrote is what we care about (in fact we don't even care that the domain is anything in particular, the only thing we actually need for ``<. A , B >. e. F -> ( F ` A ) = B`` is the `Fun F` part), but that doesn't make a good induction hypothesis.

With a one-parameter family of domains of the form `(1 ... N)`, we can have lemmas like strlemor3 that let you append triples on the right, resulting in a left-leaning tree of triples. For large n this means that extracting all the components can be done in O(n^2) steps.

The current df-struct uses a two-parameter family of domains `M ... N`, with some extra assertions about the bounds so that we know they are natural numbers. The core automation lemma in the new approach is strleun; notice that it has no superfluous hypotheses, the only thing you need is that the upper bound of left subtree is less than the lower bound of the right subtree. This means that you can prove df-struct for any tree of unions of 1/2/3-ary sets, with any balance distribution. So for large n we are best off with a perfectly balanced tree of unions, for which the proof of all component extraction lemmas can be done in only O(n log n) steps.

benjub:
Thanks @digama0. But what are the uses of theorems like strle* ?

 Actually, is it useful to know that something is a structure (in the set.mm sense) ? Equivalently, are there (projects of) set.mm theorems that have "Struct" in the hypothesis ?  If I compare with the mathematical notion of category: in addition to its independent interest, it is useful to know that something is a category, because then you can apply to it categorical notions like that of isomorphism, subobject, quotient, functor, natural transformation, limit, adjoint... But there is no such thing for structures, it seems ? (I see there are definitions of substructure and product structure, but they seem to be applicable to any class.)

We could just define (with the current coding of "Slots"): a group is a class containing a function on {0, 1} such that... ; a ring is a class containing a function on  {0, 1, 2} such that... ; a topological space is a class containing a function on {0, 9} such that...

Another question: is it useful to define Slot as a class-function, instead of merely a set-function ? And also, could we rename it "Eval", to make it maybe less mysterious by showing it is a familiar notion ? This would give:
> df-eval $a |- Eval = ( x e. _V |-> ( f e. _V |-> ( f ` x ) ) ) $.

avekens:
As for having "Struct" in hypotheses, I proved some theorems for graph theory, see ~basvtxval/edgfiedgval, ~structgrssvtx/structgrssiedg, ~setsvtx/setsiedg, ~structtousgr/structtocusgr, and particularly ~uhgrstrrepe/usgrstrrepe, where a given structur is made a graph by inserting an edge function in the corresponding slot. As written in the comment, however, it would be sufficient to require ` ( ph -> Fun ( G \ { (/) } ) ) ` and ` ( ph -> G e. _V ) ` instead of ` ( ph -> G Struct X ) `, or disregarding the empty set, even ` ( ph -> Fun G ) ` and ` ( ph -> G e. _V ) ` would be sufficient.

Applied to the field of complex numbers `CCfld `, which is an extensible structure, theorem ~cffldtocusgr shows that it can be made a complete simple graph with the set of pairs of complex numbers regarded as edges. This is not a surprising or significant mathematical insight, but it demonstrates the capabilities of the current definitions for graphs applied to extensible structures.

digama0:
> Thanks @digama0. But what are the uses of theorems like strle* ?
> 
> Actually, is it useful to know that something is a structure (in the set.mm sense) ? Equivalently, are there (projects of) set.mm theorems that have "Struct" in the hypothesis ?

The intended use of strle* is for proving theorems like rngbase. Extracting a component of an explicitly given set of ordered pairs is expensive, because you have to ensure that all the other components do not play a role: If the set is `F = { <. 1 , A >. , <. 2 , B >. , <. 1 , C >. }` then we cannot conclude ``(F `` 1) = A`` because `C` gets in the way, and in general for a 10-ary set this is quite onerous; we have to prove that all pairwise components are unequal.

If we know the set is a `Fun`, then this problem goes away; as long as `<. 1 , A >.` is in the set we know none of the other components can interfere. So `df-struct` and `strle*` is used to prove that the set is a `Fun` in linear time to make the component extractions easy.

As originally designed, this is the beginning and end of the purpose of both df-struct and strle*. So no, there is no need to ever have a `Struct` hypothesis, except in the strle* lemmas themselves, and there is also no need to ever have a `Struct` except in the empty context. But df-struct is a regular definition, and if people find a use for it in other contexts I have no problem with that. But it wasn't the original intention.

> If I compare with the mathematical notion of category: in addition to its independent interest, it is useful to know that something is a category, because then you can apply to it categorical notions like that of isomorphism, subobject, quotient, functor, natural transformation, limit, adjoint... But there is no such thing for structures, it seems ? (I see there are definitions of substructure and product structure, but they seem to be applicable to any class.)

That's right. A structure is just a set, so there is no predicate associated to "being a structure", unlike with categories.

> We could just define (with the current coding of "Slots"): a group is a class containing a function on {0, 1} such that... ; a ring is a class containing a function on {0, 1, 2} such that... ; a topological space is a class containing a function on {0, 9} such that...

I'm not sure I understand the suggestion. What do you mean by "containing a function"? If you mean "being a function", i.e. a group is a function on {0, 1} such that ... then it's too strict since we want a ring to be a group. If you mean that a group is a class with a subset that is a function on {0, 1} such that ... then it's not strict enough because e.g. `RR X. RR` is a superset of `( x e. RR |-> (x ^ 2) )` but that doesn't mean that evaluating `(RR X. RR)` at `2` yields `4`. You need to at least know that the set has no extra stuff interfering on the coordinates 0 and 1.

But yes, there are ways to make structures more constrained than mere sets. This is mostly just inertia from Norm's original definition. The most constrained definition that covers all examples I know would be that a structure is a finite set with `Fun F /\ dom F C_ NN`. In order to make explicit sets of ordered pairs satisfy this definition we would have to "clean" them using an operation equivalent to `\ { (/) }`, so that you can still have theorems like rngbase which don't need to assert sethood of irrelevant things.

> Another question: is it useful to define Slot as a class-function, instead of merely a set-function ? And also, could we rename it "Eval", to make it maybe less mysterious by showing it is a familiar notion ? This would give:
> 
> > df-eval $a |- Eval = ( x e. _V |-> ( f e. _V |-> ( f ` x ) ) ) $.

For the first part, no it can't be a set-function because ``( f e. _V |-> ( f ` x ) )`` is a proper class, so ``( Eval ` 1 ) = (/)`` which is not what you want. For the second part (why Slot instead of Eval), the idea is to hide the implementation details of slots and just talk about components of a structure. The fact that that structures are functions(-ish) and slot evaluates the function at some particular natural number is something we want to hide from the API. The only thing you should need to know about structures is that `G` is a set and ``( Base ` G )`` gets the base set of the structure and ``( x ( +g ` G ) y )`` applies the group operation. The way `Base` and `+g` work is not relevant to the definition of a group at all.

benjub:
> I'm not sure I understand the suggestion. What do you mean by "containing a function"? If you mean "being a function", i.e. a group is a function on {0, 1} such that ... then it's too strict since we want a ring to be a group. If you mean that a group is a class with a subset that is a function on {0, 1} such that ... then it's not strict enough because e.g. `RR X. RR` is a superset of `( x e. RR |-> (x ^ 2) )` but that doesn't mean that evaluating `(RR X. RR)` at `2` yields `4`. You need to at least know that the set has no extra stuff interfering on the coordinates 0 and 1.

I meant the second (i.e., having as subset a function on {0, 1}), but indeed as you note this is not good enough, and I actually meant your third option (fix below).

> > Another question: is it useful to define Slot as a class-function, instead of merely a set-function ? And also, could we rename it "Eval", to make it maybe less mysterious by showing it is a familiar notion ? This would give:
> > > df-eval $a |- Eval = ( x e. _V |-> ( f e. _V |-> ( f ` x ) ) ) $.
> 
> For the first part, no it can't be a set-function because `` ( f e. _V |-> ( f ` x ) ) `` is a proper class, so `` ( Eval ` 1 ) = (/) `` which is not what you want. For the second part (why Slot instead of Eval), the idea is to hide the implementation details of slots and just talk about components of a structure. The fact that that structures are functions(-ish) and slot evaluates the function at some particular natural number is something we want to hide from the API. The only thing you should need to know about structures is that `G` is a set and `` ( Base ` G ) `` gets the base set of the structure and `` ( x ( +g ` G ) y ) `` applies the group operation. The way `Base` and `+g` work is not relevant to the definition of a group at all.

Thanks, I had missed the fact the values of genuine functions are sets and cannot be proper classes.  As for the rest, I understood what part of the "API" is "public" and what is "private", but that's no reason to try and make the private part clearer.  But maybe "Slot" is more standard than "Eval" in the CS community, so I'll let you judge what is best.
 
> But yes, there are ways to make structures more constrained than mere sets. This is mostly just inertia from Norm's original definition. The most constrained definition that covers all examples I know would be that a structure is a finite set with `Fun F /\ dom F C_ NN`.

Thanks, this gives back my first proposal above.

The parts of your post I haven't quoted are perfectly clear.  To summarize, here are a few possible definitions for Struct; please correct/complement them if needed.

There are several options to implement "extensible structures".

* Very-lax convention: do not define Struct at all; only use component extractions, e.g.:
```
  |- Magma_lax = { x | ( +g ` x) is a law on ( Base ` x ) }
  |- TopSpace_lax = { x | ( TopSet ` x) is a topology on ( Base ` x ) }
```
* Lax convention: do not define Struct at all; only use component extractions, e.g.:
```
  |- MagmaLike_lax = { x | E! y 1 x y /\ E! z 2 x z }
  |- TopSpaceLike_lax = { x | E! y 2 x y /\ E! z 9 x z }
```
* Stricter conventions, in increasing order of strictness:
```
  |- Struct1 = { f | Fun ( f \ { (/) } ) } $.
  |- Struct2 = { f | ( Fun ( f \ { (/) } ) /\ dom f C_ NN ) } $.
  |- Struct3 = { f | ( Fun ( f \ { (/) } ) /\ dom f C_ NN /\ dom f e. Fin ) } $.
```
* Finally, consider the following two definitions, where Struct_i is to be used as a binary relation:
```
  |- Struct_1 = { <. f , x >. | ( x e. NN /\ Fun ( f \ { (/) } ) /\ dom f C_ ( 1 ... x ) ) }
  |- Struct_2 = { <. f , x >. | ( x e. ( <_ i^i ( NN X. NN ) ) /\ Fun ( f \ { (/) } ) /\ dom f C_ ( ... ` x ) ) } $.
```

With these last two, one obtains the same structures as with Struct3.  More precisely:
`$p  |- Sruct3 = U. { f | E. x f Struct_i x }`
Even though there is no difference in the class of structures thus obtained, the reason to prefer Struct_i over Struct3 is algorithmic: having control on the maximum integer (resp. the minimum and maximum integers) of interest allows to prove more efficiently lemmas to extract components of a structure, and for this, Struct_2 is more efficient than Struct_1, which is why it was chosen for the definition of structures.  [details: write them down or point to Mario's answer above].

For each of the five definitions, one could replace "Fun ( f \ { (/) } )" with "Fun f" to be more stricter.  Adding the possibility that f may include the empty set ensures that some expressions using pairs (which equal the empty set when used with proper classes), are always structures without needing sethood hypotheses.

Of course, choosing NN as the domain of a structure is an arbitrary choice, and we give mnemonics to extract the components: for instance, a topological space as an extensible structure is a class X which is a function on a domain containing {0, 9} (with the possible addition of the empty set), but this is transparent to the "user".  All the user needs to know is that the underlying set of X can be recovered as ( Base \` X ) and its topology as ( TopSet \` X ).

PR #2332, opened 2021-11-29
Misc.

avekens:
Main set.mm:
* enhancement of convention about theorems in "deduction form" according to Google group discussion https://groups.google.com/g/metamath/c/HwYShAtjptA/m/ksfXqm6MBgAJ
* new theorems ~l2p, ~n0lplig, ~pliguhgr (as proposed by Gerard Lang) in section "Planar incidence geometry"

AV's mathbox:
* comments adjusted


wlammen:
I wonder whether "the hypotheses and the conclusion are implications that share the same antecedent." should be changed to 
"... first antecedent".

avekens:
@wlammen Usually, it's only one antecedent ph which is used for the hypotheses and the conclusion. The exception of the usage of more than one antecedent (i.e. a conjunction) is mentioned afterward. Or do yo refer to chains of implications ( ph -> ( ps -> ( th -> ch ) ) )? In the comment for ~wi, it is written: "In the case of ` ( ph -> ( ps -> ch ) ) ` , the middle ` ps ` may be informally called either an antecedent or part of the consequent depending on context." Then it makes sense to talk about the first antecedent. But saying "THE antecedent" implies that there is only one, or only the first one is considered as an/the only antecedent in this context, 

I can change this if there is no objection. @nmegill @david-a-wheeler what do you think? The text in set.mm, however, is currently identical with the text in the Metamath book...

david-a-wheeler:
Instead of changing the definition of "deduction form", why not change the definition of when "d" is added (which I think is really the point)? Basically, you can add "d" if it's in deduction form, zero hypothesis, or the "non" deductive hypothesis declare not-freeness?

wlammen:
@avekens I indeed referred to chains of implication with perhaps more than one antecedent, as in a1dd.  Such examples are rare, and the antecedent marking the context is usually the first one in these cases.  My remark was super picky, and can perhaps be ignored.  It was just in case...

benjub:
The discussion is still ongoing on the discussion group.  Maybe it's a bit early to modify set.mm yet ?

nmegill:
> The discussion is still ongoing on the discussion group. Maybe it's a bit early to modify set.mm yet ?

I'll accept it as is for now (changing "one" to "zero"), clearing up the immediate "d"  label suffix convention issue, and we can revise it as the discussion there evolves.

PR #2333, opened 2021-11-29
editorial fixes, shorten wl-ax8clv2

wlammen:


nmegill:
There is a merge conflict with the previous PR.  (BTW @wlammen are you as the PR author notified of these?  I don't know if I need to keep posting comments informing authors of PR merge conflicts, or whether they are notified automatically.)

wlammen:
@nmegill I am not automatically notified.  I learn about the conflict, when I sync my local copy of set.mm with the most recent updates on github.  If you don't state the reason here, it looks at first glance as if the PR was accidentally overlooked. I am pretty sure, after a few incidences of this kind, I will have learned the lesson, and routinely check for merge conflicts myself.  Still, I think, it is helpful, to inform all about the maintainer's decision (and reason) to reject the PR.  Your choice is involved, you could have preferred my PR over the other.

PR #2336, opened 2021-12-01
use term 'subcomplex' for modules/vector spaces over subrings/subfields of CC

benjub:
Title says it all. I finally did this, which was announced in #947 #2201 #2230.  Pinging @nmegill, @avekens, @digama0 since they were involved in these discussions.

I did not touch the deprecated parts nor the mathboxes.


avekens:
Thank you, Benoît, for the revision. From some comments, however, I think you removed too much - I will make a purpose in my next PR. Additionally, I want two show by a theorem that left and right modules are "the same" if the underlying ring is commutative.

PR #2343, opened 2021-12-02
Modify CI pipeline to check all databases

david-a-wheeler:
This uses C metamath.exe to check all .mm databases.
Currently this fails spectacularly, because the other databases
don't comply with the checks. We could weaken the tests or
fix the databases, but this at least lets us see what that means.

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -106,7 +106,7 @@ jobs:
       - run: metamath 'read iset.mm' 'markup mmil.raw.html mmil.html /ALT /CSS' quit
       - run: scripts/verify --top_date_skip --extra 'write bibliography mmbiblio.html' set.mm
       - run: scripts/verify --top_date_skip iset.mm
-      - run: ls -1 *.mm | sed -e '/^set\.mm$/d' -e '/^iset\.mm$/d' | while read db ; do scripts/verify --top_date_skip "$db"; done
+      - run: ls -1 *.mm | sed -e '/^set\.mm$/d' -e '/^iset\.mm$/d' | while read db; do scripts/verify --top_date_skip --no-verify-markup "$db"; done
       - run: scripts/check-raw-html set.mm mmcomplex.raw.html mmdeduction.raw.html mmnatded.raw.html mmnf.raw.html mmset.raw.html mmzfcnd.raw.html mmfrege.raw.html
       - run: scripts/check-raw-html iset.mm mmil.raw.html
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
@nmegill @digama0 This pull request checks every `.mm` database. It fails spectacularly, because many .`.mm` dataabases don't conform to the checks. Should we weaken the validation test, or fix the databases?

nmegill:
We should be verifying their proofs but NOT doing "verify markup" for every database.  "verify markup" is irrelevant to the correctness of proofs.  The markup is primarily intended for the purposes of display cosmetics (htmldefs etc.) and development assistance (discouragements, mathbox independence).

"verify markup" should not be done for non-displayed database like miu.mm and big-unifier.mm, because they will never have htmldef statements etc. nor should they be required to.

The only displayed databases besides set.mm are iset.mm, nf.mm, hol.mm, and ql.mm.  I think iset.mm passes "verify markup", and we can check it (if it isn't already).  The other 3 fail, and while it would be nice to update them eventually to pass "verify markup", there is no urgency (the web pages currently display fine) and we shouldn't make passing "verify markup" a requirement at this point.  If eventually someone volunteers to update them (it may be a lot of work) then we can add them to the "verify markup" checks.

digama0:
Can you modify the script so that it still runs all the checks even if one of the files fails, rather than serializing them? Right now `big_unifier.mm` fails so `peano.mm` is not checked.

david-a-wheeler:
@nmegill :

> We should be verifying their proofs but NOT doing "verify markup" for every database...
> "verify markup" should not be done for non-displayed database like miu.mm and big-unifier.mm, because they will never have htmldef statements etc. nor should they be required to.
> The only displayed databases besides set.mm are iset.mm, nf.mm, hol.mm, and ql.mm. I think iset.mm passes "verify markup", and we can check it (if it isn't already).

Yes, iset.mm passes "verify markup". I'll have to tweak the verify script to skip verify markup, but that's easily doable.

@digama0 :
> Can you modify the script so that it still runs all the checks even if one of the files fails, rather than serializing them? Right now big_unifier.mm fails so peano.mm is not checked.

That's possible, but not entirely trivial. "Fail immediately" is easy, but to do what you're asking I have to track each result. I think once we stop checking markup this won't be a big deal. So I'd rather not, unless it's really important to you.

david-a-wheeler:
Okay, it now checks all databases' proofs. On set.mm and iset.mm it also checks the markup.

I didn't try to "run to completion no matter what"; once there's a database failure, I don't think we need to go further, and in most cases it won't be necessary or helpful.

PR #2344, opened 2021-12-03
add rspc2gv, nsnlplig, eulplig

benjub:
@avekens : I proved your n0lplig from nsnlplig, and I also used -. A e. B instead of A e/ B since the translation to the latter idioms added two steps at the end of n0lplig and the translation back added two more steps where it was used in pliguhgr.

Maybe you prefer n0lpligOLD to be n0lpligALT instead ?

I also added eulplig, which is the third (out of three) property of a planar incidence geometry (the other two being tncp and l2p). To prove it, I added rspc2gv which might have other uses.

@nmegill : I added isplig2, connected to my latter question on the discussion group (https://groups.google.com/g/metamath/c/8D175dX4cp8/m/H5dVhWIEBwAJ).  As for tncp/tncp2, I will eventually remove one of them (it is the same statement, one proved from isplig, the other from isplig2).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16740,6 +16740,7 @@ New usage of "mulpqnq" is discouraged (7 uses).
 New usage of "mulresr" is discouraged (4 uses).
 New usage of "mulsrpr" is discouraged (9 uses).
 New usage of "n0fOLD" is discouraged (0 uses).
+New usage of "n0lpligOLD" is discouraged (0 uses).
 New usage of "naecoms-o" is discouraged (1 uses).
 New usage of "natded" is discouraged (0 uses).
 New usage of "negexsr" is discouraged (0 uses).
@@ -19398,6 +19399,7 @@ Proof modification of "mulgnnassOLD" is discouraged (439 steps).
 Proof modification of "mulgnnclOLD" is discouraged (39 steps).
 Proof modification of "mulgnndirOLD" is discouraged (440 steps).
 Proof modification of "n0fOLD" is discouraged (51 steps).
+Proof modification of "n0lpligOLD" is discouraged (77 steps).
 Proof modification of "naecoms-o" is discouraged (19 steps).
 Proof modification of "nelsnOLD" is discouraged (46 steps).
 Proof modification of "nexdOLD" is discouraged (9 steps).


nmegill:
I noticed you changed ~l2p so that the line break occurs after the "->" instead of before.  I used to do that in early theorems, then (many years ago) someone wrote to me about this, arguing that connectives should go after the line break; e.g.
```
1 + 2 + 3
    + 1 + 1 + 1 + 1
  = 6 + 4
  = 10
```
is arguably easier for the eye to follow than
```
1 + 2 + 3 +
    1 + 1 + 1 + 1 =
  6 + 4 =
  10
```
and in any case the former is usually used in textbooks.

Eventually he convinced me, and I started using the "line break before connective" method.  I vaguely intended to update all existing theorems but never got around to it.  But most theorems I added since then use "line break before connective".

---

I also noticed you right-aligned some continuation lines like in ~tncp.  While the original could use some more indentation, full right alignment hasn't been our typical practice.  One reason I don't usually right-align is that it wastes bytes in set.mm; maybe that's a silly reason these days (I developed some early parts of set.mm on a laptop with 4MB of memory). But does it really help readability vs. say 4 spaces of additional indentation compared to the line above it?

---

These are trivial issues, and I'm not saying that you should change anything at this point, just that maybe they're worth some discussion.

avekens:
> @avekens : I proved your n0lplig from nsnlplig, and I also used -. A e. B instead of A e/ B since the translation to the latter idioms added two steps at the end of n0lplig and the translation back added two more steps where it was used in pliguhgr.
> 
> Maybe you prefer n0lpligOLD to be n0lpligALT instead ?

I am not happy with this change, because I try to use A e/ B as often as possible instead of -. A e. B, because A e/ B looks nicer on the web pages (textbook style), and is shorter. That proofs become longer is because there are only few (basic) theorems using e/ - when this will change, then also the proofs become shorter. To optimize higher level theorems so that proofs using these theorems become shorter should not be the primary goal (in my opinion).

According to the line breaks and the placement of the connectors, I fully agree with Norm, and I use the style to put the connector at the beginning of a line. I think I made a proposal to fix this as convention some time ago (maybe in the Google group, but I cannot find it anymore). Until there is no agreement about such a convention, the theorems (of others) should not be restuctured. As far as I understand Norm, he regards the "line break before connective" method as de facto standard.

Finally, I also use the "right-aligned" method, but only if there is no better way to arrange the theorem. It should not be a must.

benjub:
@avekens , @nmegill :
Concerning n0lplig, there are two questions:
1) do you prefer that n0lpligOLD be kept as n0lpligALT (the proof is longer, but bypasses nsnlplig) ?
2) do you prefer that I use the idiom A e/ B instead of -. A e. B (note that I did not change the "e/" in n0lpligOLD). I have long balanced between both, and I'm not sure what to think of the multiplication of library lemmas to handle e/ directly (I added some myself in the past).

Concerning line breaks: I sometimes did that to statements that I needed to look at closely, in order to see clearer, but this may very well be because I'm used to it, and be very idiosyncratic.  I'll refrain to do so in the future.  Note, however, that the line breaks are all after "$=", and not before.  Similarly, I have a few times broken the line right after "|-" in order to allow the formula itself to fit on a single line.

@nmegill : which of isplig/isplig2 should be kept ? Both ? You may also want to react to the related discussion https://groups.google.com/g/metamath/c/8D175dX4cp8/m/H5dVhWIEBwAJ

nmegill:
> 1. do you prefer that n0lpligOLD be kept as n0lpligALT (the proof is longer, but bypasses nsnlplig) ?
> 2. do you prefer that I use the idiom A e/ B instead of -. A e. B (note that I did not change the "e/" in n0lpligOLD). I have long balanced between both, and I'm not sure what to think of the multiplication of library lemmas to handle e/ directly (I added some myself in the past).
> 

Personally I would have chosen n0lplig over n0lpligOLD, but if AV prefers the latter, he is the contributor and, given that we don't have a clear convention on this, maybe his preference should take precedence at least until we do.

The bigger issue is what should our general convention be w.r.t. "-. A e. B" vs. "A e/ B".  Here is some background.

In early set.mm development, I placed a lot of emphasis on keeping the set.mm size small with short proofs and not too many seldom used utility theorems (motivated partly by limited CPU resources at the time, but also by a sense that shorter is better).

My original intent when introducing "=/=" and "e/" was to make them available mainly for what you might call "display" purposes for occasional important theorems.  For example, for a long time about the only theorems using "e/" were a couple of equality theorems and sqrt2irr.  I felt sqrt2irr was useful to display with "e/" because it was featured in the 17 Provers of the World book, and the conversion to "e/" was done as the last step.  If sqrt2irr were frequently used, I might have even added a "-. e." version for general use to save the conversion step (if that "paid for itself").  Currently sqrt2irr is used in only one proof, where we convert it back to "-. e." in that proof.

IIRC it was well into the early 2000s before "=/=" became commonly used.  After I did an analysis around that time, I made a massive change to retrofit almost all uses of "-.  A = B" to "A =/= B", and the net savings in proof size were significant, even with the new library theorems for "=/=" that became needed.  The savings mostly resulted from things like neeq12d saving a step over eqeq12d + notbid, more than compensating for the occasional need to convert between "-.  A = B" and "A =/= B".

I looked at doing the same with "e/", and it was not even close, at least at that time.  We would almost always need conversions from "-.  A e. B" to "A e/ B" and vice-versa.  A property of "A =/= B" was that it was frequently a common antecedent in a series of theorems (e.g.  "A =/= 0" in division theorems).  That wasn't often true of "-.  A e. B", which seemed to be more commonly used as a consequent.

I haven't analyzed the current set.mm and don't know the full picture anymore.  I did a rough count of occurrences:

"-. ? e." occurs in 430 theorems
"e/"  occurs in 170 theorems

"-. ? =" occurs in 70 theorems
"=/=" occurs in 3000 theorems

My guess is that most of the theorems with "e/" were added by newer contributors (after 2015 or so) but haven't checked.  Most earlier contributors including myself tended to use "-. e.".

I am willing to go either way, but we should check the impact on proof sizes before doing a massive conversion to "e/".  We should also consider the additional impact of having to add perhaps many new library theorems for "e/".


> 
> @nmegill : which of isplig/isplig2 should be kept ? Both ? You may also want to react to the related discussion https://groups.google.com/g/metamath/c/8D175dX4cp8/m/H5dVhWIEBwAJ

I think I have mostly used the isplig pattern.  We can use any membership relation, not just explicit "e. \_V", to satisfy the antecedent.  Moreover, the consequent "looks" like a normal definition without an awkward "G e. \_V" conjunct that would not be present in ordinary literature.

In the end, I would say that what matters is which is easier to use.  If there is no practical difference - the sethood term can be eliminated in 1 step with ax-mp or mpbiran resp. - I would probably favor isplig, but I can't think of a good objection to either.  Actually, do we even ultimately need the "e. \_V" in the conclusion, or could it be a $e?

benjub:
Thanks @nmegill for the background information.  I suspected indeed that the cases of =/= and e/ would be different. After all, even in plain in English, there is a word for =/= ("different") but not for e/ (same in French).

As for isplig versus isplig2, I can remove isplig2 if you wish, or by default keep both.  Probably, having the sethood hypothesis in an $e-statement would be enough for applications in practice, but I like the idea of having a stronger "characterization among all classes" rather than a weaker "characterization among classes that are sets", even if said characterization reads "you are xxx if and only if you are a set and yyy".

@avekens : I'm ready, if you wish, to
1) change n0lpligOLD to n0lpligALT (since it bypasses nsnlplig and is substantially shorter than the proof obtained from n0lplig by expanding nsnlplig).
2) change n0lplig to use the idiom "e/" instead of "-. e." (note: n0lpligOLD still uses "e/")

avekens:
@benjub sorry for my late response to your comments, but I wanted to cleanup/merge my latest changes first to have a clear picture of the discussed issues. See my remarks below.

> Thanks @nmegill for the background information. I suspected indeed that the cases of =/= and e/ would be different. After all, even in plain in English, there is a word for =/= ("different") but not for e/ (same in French).

Yes,  Norm's background information is very enriching. Is there a better place to preserve it than in a comment of a PR?

> As for isplig versus isplig2, I can remove isplig2 if you wish, or by default keep both. Probably, having the sethood hypothesis in an $e-statement would be enough for applications in practice, but I like the idea of having a stronger "characterization among all classes" rather than a weaker "characterization among classes that are sets", even if said characterization reads "you are xxx if and only if you are a set and yyy".

I'am also often uncertain which of these variants to use. ~ isplig2 is more powerful because it is a biconditional, but ~ isplig is easier to use/understand. Maybe the theorems can be swapped and ~isplig can be proven (with a shorter proof) by using ~isplig2. ~isplig2 can/should be renamed ~ispligb ("b" for "biconditional" - see conventions for Suffixes". The comment of ~isplig can be refined to "The predicate "is a planar incidence geometry" for a set.".

> @avekens : I'm ready, if you wish, to
> 
>     1. change n0lpligOLD to n0lpligALT (since it bypasses nsnlplig and is substantially shorter than the proof obtained from n0lplig by expanding nsnlplig).
> 
>     2. change n0lplig to use the idiom "e/" instead of "-. e." (note: n0lpligOLD still uses "e/")
I think option 1. would be fine/sufficient (at least for the moment, until it becomes clearer when and how to use e/). The comment should be "$( There is no "empty line" in a planar incidence geometry.  Alternate version of ~ n0lplig , using the symbol ` e/ ` (which is more appropriate for display reasons) and not using ~ nsnlplig . ..." Selecting option 2. would result in a conflict with ~ nsnlplig (one would use e/ and the other not).

PR #2345, opened 2021-12-04
Right and left modules

avekens:
follow up PR for #2336 (see also the conversation there)

* ~ovexd moved from GS's mathbox
* ~2ralimi, ~ralcom13, ~ralro3  added
* ~theorem ~rmodislmod (right and left modules are "the same" if the underlying ring is commutative) added
* comment of ~df-clm enhanced
* header comments of sections "Subcomplex vector spaces" and "Normed subcomplex vector spaces" enhanced
* comments of ~recvs, ~qcvs and ~zclmncvs reverted (terms "real left module" and "rational left module" are not defined)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12797,6 +12797,9 @@
 "strlem4" is used by "strlem6".
 "strlem5" is used by "strlem6".
 "strlem6" is used by "stri".
+"strlemor1OLD" is used by "strlemor2OLD".
+"strlemor1OLD" is used by "strlemor3OLD".
+"strlemor2OLD" is used by "strlemor3OLD".
 "structgrssvtxlemOLD" is used by "structgrssiedgOLD".
 "structgrssvtxlemOLD" is used by "structgrssvtxOLD".
 "sumdmdi" is used by "dmdbr4ati".
@@ -13402,9 +13405,7 @@ New usage of "19.23hOLD" is discouraged (0 uses).
 New usage of "19.23tOLD" is discouraged (1 uses).
 New usage of "19.27OLD" is discouraged (0 uses).
 New usage of "19.28OLD" is discouraged (1 uses).
-New usage of "19.29rOLD" is discouraged (0 uses).
 New usage of "19.3OLD" is discouraged (2 uses).
-New usage of "19.40bOLD" is discouraged (0 uses).
 New usage of "19.41rg" is discouraged (3 uses).
 New usage of "19.41rgVD" is discouraged (0 uses).
 New usage of "19.43OLD" is discouraged (0 uses).
@@ -14755,6 +14756,7 @@ New usage of "cnvbrabra" is discouraged (2 uses).
 New usage of "cnvbracl" is discouraged (2 uses).
 New usage of "cnvbramul" is discouraged (1 uses).
 New usage of "cnvbraval" is discouraged (1 uses).
+New usage of "cnvcnvOLD" is discouraged (0 uses).
 New usage of "cnvssOLD" is discouraged (0 uses).
 New usage of "cnvunop" is discouraged (2 uses).
 New usage of "com3rgbi" is discouraged (1 uses).
@@ -15424,7 +15426,6 @@ New usage of "eleigvec" is discouraged (2 uses).
 New usage of "eleigvec2" is discouraged (2 uses).
 New usage of "eleigveccl" is discouraged (3 uses).
 New usage of "eleq2dALT" is discouraged (0 uses).
-New usage of "eleq2dOLD" is discouraged (0 uses).
 New usage of "elex22VD" is discouraged (0 uses).
 New usage of "elex2VD" is discouraged (0 uses).
 New usage of "elghomOLD" is discouraged (5 uses).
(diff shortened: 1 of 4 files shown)


benjub:
~ rmodislmod is a nice addition. It would be nice to prove it from the more general statement:
> If M is a right R-module, then it is a left R^{op}-module.

That is, you should define in an $e-hypothesis a multiplication x2 on F such that ( x x2 y ) = ( y .x. x ).  Actually, the result that F with x2 is still a ring is a separate theorem, and there could be a specific definition for the opposite of a rng/ring (and actually of any magma...).

Also, you denote the module by R and the ring by F, which is a bit puzzling (one would think that R denotes a ring and F a field).

(Of course, these are small criticisms on a very nice achievement.)

avekens:
> ~ rmodislmod is a nice addition. It would be nice to prove it from the more general statement:
> 
> > If M is a right R-module, then it is a left R^{op}-module.
> 
> That is, you should define in an $e-hypothesis a multiplication x2 on F such that ( x x2 y ) = ( y .x. x ). Actually, the result that F with x2 is still a ring is a separate theorem, and there could be a specific definition for the opposite of a rng/ring (and actually of any magma...).

That would be interesting, but I do not know if I have time for that (~ rmodislmod took a lot of time already). Maybe someone else could add such a theorem.

> Also, you denote the module by R and the ring by F, which is a bit puzzling (one would think that R denotes a ring and F a field).

I have chosen R to denote the _R_ight module, fully aware of a possible confusion with a ring. The F for the "scalar component" is also used in the theorems for LMod, usually also being a ring in this context. Unfortunately, a lot of class variables are required, but we have only a limited number of symbols for them...

benjub:
It's the first time I tried the "review" thing, and my review are still "pending" while the PR has been merged, so I'm not sure what to do with them. @avekens : do you want to address them ?

As for ~rmodislmod, I can believe you that it took a long time !  Defining the "opposite" operation and the isomorphism of categories
> right-R-Mod -~-> left-R^{op}-Mod

is indeed an ambitious project !  Should I record it as a github "issue" ? (question to @nmegill as well)

nmegill:
I'm sorry, I missed the review.  I'll try to revert the merge, never done that before...

nmegill:
> I'm sorry, I missed the review. I'll try to revert the merge, never done that before...

(The above comment should have been sent at 6:00pm my time, I was on the wrong screen, sorry.  It was in response to:)

> It's the first time I tried the "review" thing, and my review are still "pending" while the PR has been merged, so I'm not sure what to do with them. @avekens : do you want to address them ?
> 
> As for ~rmodislmod, I can believe you that it took a long time ! Defining the "opposite" operation and the isomorphism of categories
> 
> > right-R-Mod -~-> left-R^{op}-Mod
> 
> is indeed an ambitious project ! Should I record it as a github "issue" ? (question to @nmegill as well)

benjub:
@nmegill I think you do not need to revert: it's only one small thing in a larger PR, and can be dealt with in a forthcoming PR.

avekens:
> It's the first time I tried the "review" thing, and my review are still "pending" while the PR has been merged, so I'm not sure what to do with them. @avekens : do you want to address them ?

I cannot see any review remarks (anymore)... I think the "review" feature is not used consequently for set.mm - and not all users of the metamath github seem to have the same permissions to use (parts of) it. I think I cannot perform reviews on set.mm PR's (I mentioned this already some time ago). @david-a-wheeler @nmegill maybe you can have (again) a look at it and try to clarify who can use this feature and if it should be used for the metamath repositories (and how).
 
> As for ~rmodislmod, I can believe you that it took a long time ! Defining the "opposite" operation and the isomorphism of categories
> 
> > right-R-Mod -~-> left-R^{op}-Mod
> 
> is indeed an ambitious project ! Should I record it as a github "issue" ? (question to @nmegill as well)

Yes, please open an github issue for this project.

benjub:
On this webpage, when I click the tab "Files changed", I can still see my "reviews", with the "pending" tag, but maybe they are visible only to me because I did not "validate" them or something...  Anyway, I'll restate them here, as proposals to @avekens to modify some comments (which either you or I can do after you approve them):

1.
> Since the field of complex numbers is commutative, so are its subrings, left modules over such subrings are the same as right modules, see ~ rmodislmod .  Therefore, we drop the word "left" from "subcomplex left module".

I would split the sentence (which appears to lack a conjunction). For instance:
> Since the field of complex numbers is commutative, so are its subrings.  But over commutative rings, left and right modules coincide (see ~ rmodislmod ).  Therefore, we drop the word "left" from "subcomplex left module".

2.
> The set of real [resp. rational] numbers (as a subring of itself) is a subcomplex vector space.

I think the subject of these two sentences should be a module, not a set.  Therefore, following the wording of df-rgmod, I would propose something like any of:

> The module of rational numbers (as a module over itself) is...
> The module of rational numbers (with scalar ring the rational numbers) is...
> The module of rational numbers (that is, the additive group of rational numbers with scalars the ring of rational numbers) is...
> ...

By the way: I was also happy with my previous wording: you wrote that "real vector space" and "rational vector space" are not defined, but these are very common terms.

avekens:
I'll come back to these review remarks after the changes which still are in the queue (minimizing proofs) are processed.

PR #2351, opened 2021-12-05
Motivate the idea of a class definition separate from sets

wlammen:


nmegill:
> In ZFC, as presented in this document, classes are meant to be just a notational convenience, that can be reduced to pure set theory by means of ~ df-clab (there stated in the eliminable property).  That is, in an expression ` x e. A ` , the class variable ` A ` is implicitely assumed to represent an expression ` { z | ph } ` with some appropriate ` ph ` .  Unfortunately, ` ph ` syntactically covers any well-formed formula (wff), including ` z e. A ` . This choice inevitably breaks the stated property.

I'm not clear on what property this breaks.  The substitution of {z | ph} for A is part of the conversion from class to wff notation, and the substitution of z e. A for ph is part of the reverse process of converting wff notation to class notation.

We have extended the wff syntax to include class expressions, so z e. A can be substituted into any previous theorem with wff variables e.g. ( z e. A -> z e. A ) is an instance of ~id.

Some texts like Takeuti/Zaring introduce class notation by defining a "wff in the wider sense" that includes the primitive wffs plus the new wffs introduced for class notation.  T/Z is reasonably rigorous, and it might be more satisfying to you than my informal discussion here.  If you don't have it, I can send you that part of the text.

I found that after I did a number of conversions from wff notation to class notation and back as exercises, the notation became intuitive and "obvious".  Of course what is intuitive to me may not be to everyone.

> And it potentially carries over to any expression containing class variables.  To fix this, a simple rule could exclude class variables at all in a class defining wff .

I'm not sure what you mean by "a class defining wff".  What would be an example that should be excluded?

> A more elaborate rule could detect, and limit exclusion to proper classes (potentially problematic).

What is an example of something that would be problematic?

> In any case, the verification process should enforce any such rule during replacement, which it currently does not.  The result is that we rely on the awareness of theorem designers to this problem.  It seems, in ZFC proper classes are reduced to a few instances only, so a careful study may reveal that this limited use does not impose logical loop holes.

There is no such thing as a proper class in the native language of ZFC, so I'm not sure what is meant by "reduced to a few instances only".  I found it helpful to convert a few theorems containing `_V` to better understand what the symbol `_V` really means; it does not, of course, correspond to any setvar nor can be directly substituted for one (~vtocl etc. require that a class term exists before it can be substituted for a setvar).

For example, ~vprc, `-. _V e. _V`, translates to `-. { x | x = x } e. { y | y = y }` (by df-v) to `-. E. z ( A. w ( w e. z <-> [ w / x ] x = x ) /\ [ z / y ] y = y )` (after applications of df-clel, df-cleq, df-clab to exhaust all class terms), which simplifies to `-. E. z A. w w e. z` using predicate calculus.  All steps in that process are reversible.  The expressions `-. _V e. _V` and `-. E. z A. w w e. z` are two different ways of expressing the same mathematical fact.

> It must be said, still, this necessary extra knowledge contradicts the general philosophy of Metamath, trying to establish certainty by a machine executable confirmation.

> An extension to ZFC allows classes to exist on their own.

I think you mean that class symbols can be manipulated as stand-alone symbols.  Something like `_V` doesn't "exist" in the ZFC sense of existence; e.g. `E. x x = _V` is of course false.

> Classes are then extensions to sets, also seamlessly extending the idea of elementhood.  In order to move from the general to the specific, sets presuppose classes, so classes should be introduced first.

Class notation is a shorthand for representing the native language in a more compact form.  Anything with class notation can be proved with wff notation, so class notation is redundant in that sense.  It is merely a definitional extension of the native language.  As I've stated elsewhere, it feels "wrong" to me to make them part of the primitive language because they do not add any logical strength nor produce any new mathematics.

>  This is somewhat in opposition to the classic order of introduction of syntactic elements, but has been carried out in the past, for example by the von-Neumann theory of classes.

NBG native variables are not the same as our "class variables" (which might be better called "class term metavariables" if we are working in NBG, or alternately we could call an NBG native variable a "classvar" to avoid confusion).  Their primitive variable type (that can be the first argument of A. and E.) is intended to range over classes (proper and non-proper).  Most NBG presentations also introduce a separate (defined) variable type ranging over sets i.e. a variable that belongs to some class, so that if c and s are classvars and setvars, E. s s = ...  would translate to E. c e. _V c = ....  All theorems using these defined setvars are the same as ZFC theorems using our setvars, but there are a few theorems quantifying over their class variables that can't be expressed in ZFC (and thus NBG "conservatively extends" ZFC).

The definition of class terms and the development of class notation in NBG is almost identical to that in ZFC.  I think if we implemented NBG in Metamath, I think the class term definitions would be the same as df-clab, df-clel, df-cleq.  And we would still have to introduce these definitions; class terms are not part of the primitive language of NBG.

wlammen:
I am currently snowed under, and will only have significant time again for private projects after Christmas.  Sorry to have delayed this response until now.  Will get back to this once I have spare time again.

PR #2356, opened 2021-12-06
Add the rest of the section "The division algorithm" to iset.mm

jkingdon:
Well, that was quick. After spending a week (or whatever it was) on https://github.com/metamath/set.mm/pull/2353 (the first theorem in this section) it turns out the rest of that section intuitionized in less than a day with no significant problems. Either proofs worked with no changes at all, or just with the usual intuitionizing like changing `≠` to `#` or changing real to rational.

I also noticed that I made a mistake back in 2018 by adding [`pm3.37`](http://us.metamath.org/mpeuni/pm3.37.html) only for decidable propositions. Turns out it holds for all propositions so this pull request makes that change.

In very minor changes, minimizes one proof in set.mm and copies one recently added theorem from set.mm to iset.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1113,6 +1113,11 @@ pm2.61dne , pm2.61dane , pm2.61da2ne , pm2.61da3ne , pm2.61iine
 </TR>
 
 <TR>
+  <TD>pm4.14</TD>
+  <TD>~ pm4.14dc , ~ pm3.37</TD>
+</TR>
+
+<TR>
   <TD>pm5.17</TD>
   <TD>~ xorbin</TD>
   <TD>The combination of ~ df-xor and ~ xorbin is the forward direction


david-a-wheeler:
@jkingdon - you're getting amazing at this intuitionizing. Please add any important tips on intuitionizing to the docs if they aren't there already. I think there are things you're doing that I wouldn't guess given only the current docs. I'm hoping to capture in the docs some of the wisdom that you, Mario, & others have learned over time :-).

jkingdon:
> @jkingdon - you're getting amazing at this intuitionizing. Please add any important tips on intuitionizing to the docs if they aren't there already. I think there are things you're doing that I wouldn't guess given only the current docs. I'm hoping to capture in the docs some of the wisdom that you, Mario, & others have learned over time :-).

If you really want the documentation to be good, other people will need to try to use it and submit changes or at least suggestions about what they found to be more or less helpful.

I mean, I have established that it is quite useful to me, which is in the short term I suppose the audience we have (well with the possible exception of @benjub or a handful of others). I couldn't do this work without the "How to intuitionize classical proofs" section of http://us.metamath.org/ileuni/mmil.html which I refer to regularly.

As for the intuitionizing in this pull request,
* There's already a bullet point `With excluded middle not equal (≠) and apart (#) . . .` as well as entries for individual theorems such as `divcan5` etc
* Real versus rational is covered at entries like `flcl` (and many other in the sections for `mod` etc)
* The one about `pm3.37` is an instance of the `If your proof relies on a biconditional  . . . see which direction your proof is using` bullet point.

benjub:
> I mean, I have established that it is quite useful to me, which is in the short term I suppose the audience we have (well with the possible exception of @benjub or a handful of others).

I'm your biggest fan !

PR #2358, opened 2021-12-06
metamath 100: mention CZF proof of induction

benjub:
Finally added mention of the CZF proof of induction, announced in https://groups.google.com/g/metamath/c/OyBTnESDvnM/m/-K14IWBECwAJ

I started from @david-a-wheeler's proposal there, but I only mentioned one variant from IZF and one from CZF (and I let @david-a-wheeler judge whether one variant from ZF is also enough, since once you get to its webpage, there are cross-references among the different variants).  I did not add precisions about the systems IZF and CZF (apart from expanding the acronyms) since I think it is not the appropriate place and any interested viewer can easily find those from a web search.

Maybe @jkingdon can provide us with the date when he added iset.mm/finds

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -497,7 +497,7 @@ The intuitionistic logic explorer (iset.mm) database also includes <a
 href="http://us.metamath.org/ileuni/finds.html">finds</a>
 (adapted from Norman Megill's set.mm proof by Jim Kingdon)
 proved from IZF (intuitionistic Zermelo--Fraenkel) and <a
-href="http://us2.metamath.org/ileuni/bj-findes.html">bj-findes</a>
+href="http://us.metamath.org/ileuni/bj-findes.html">bj-findes</a>
 (by Benoit Jubin, 2019-11-21) proved from CZF (constructive
 Zermelo--Fraenkel).
 </li>
--- a/mm_100.html
+++ b/mm_100.html
@@ -212,10 +212,6 @@ aka database iset.mm, which uses intuitionistic logic instead.
 Please email missing entries to <a
 href="email.html">Norm Megill</a>!
 
-The intuitionistic logic explorer (iset.mm) database also includes <a
-href="http://us.metamath.org/ileuni/finds.html">finds</a> and <a
-href="http://us.metamath.org/ileuni/findes.html">findes</a>.
-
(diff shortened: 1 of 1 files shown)


jkingdon:
> Maybe @jkingdon can provide us with the date when he added iset.mm/finds

Well, not by virtue of being the author - my memory doesn't file away things like that. But digging around in git (starting with `git blame iset.mm` which takes a while - maybe a minute or a few - to run but which is a very useful command to see what lines were last changed when) I did find https://github.com/metamath/set.mm/pull/654 from December 1, 2018.

This sort of illustrates some of the paradoxes of authorship - it would appear that finds itself was a copy-paste from set.mm (although further digging into set.mm as of then could confirm this) and so sorta isn't my work at all. And yes I did a lot to prove theorems leading up to it but so did @digama0 and others. I mean, go ahead and assign authors to the extent you'd like to - it doesn't really bother me - but I will repeat what I've said before - there is no way I could have done the work I've done on my own.

benjub:
Date added in PR #2359.

Indeed, it's always difficult if not impossible to assess the role of each one in any achievements, and not only in Metamath, though the problem is more there acute because of its collaborative nature.  For iset.mm/finds, @nmegill's proof may have been easy to intuitionize, but it builds on a lot of iset theorems, some of them more difficult, so all in all I think @jkingdon should be mentioned there.  As for bj-findis, I sincerely think that the attribution to me is not stolen (not that anyone said it was), and as I announced in my initial post on the discussion group, and as I mentioned in some header comments in iset.mm, this builds on works by many, probably mainly by @nmegill, @jkingdon, @digama0 (for results in set.mm, iset.mm, and discussions on the best way to formalize bounded formulas).

I thought about adding to mm_100.html a preliminary remark that most efforts are collaborative and build onto each other's works, so the personal attribution should not hide that, but then I thought that there are also many mm_100 results which were mostly individual efforts, and the authors may not like such a remark.  The problem is insoluble...

jkingdon:
> The problem is insoluble...

As long as we're aware of that we'll be OK.

david-a-wheeler:
Attribution is hard, and the old statement is true that "no man is an island".

We could still add a statement at the beginning that "Most efforts are collaborative and build on others' works." - it's still true.

PR #2359, opened 2021-12-07
mm_100: add date

benjub:
(from #2358)


benjub:
Shortened nsnlplig by copying the proof of nsnlpligALT, so attributed the shortening to @avekens.

benjub:
I shortened @avekens's pm2.61danel silently since the change is minor.

PR #2363, opened 2021-12-10
Vinogradov trigonometric sums and the circle method

tirix:


jkingdon (approved):
Mathbox changes only.

benjub:
I get a warning
> ?Warning: Bibliography keyword missing in comment for "breprexp".
    (See HELP WRITE BIBLIOGRAPHY for list of keywords.)

I was about to fix it but I'm afraid it will create merge conflicts. Could you do it @tirix ? @david-a-wheeler : any idea why the CI missed it ? Or is this warning due to something else ?

When applying the standard script (at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md) to the version just pulled from "uptream develop", I get this warning, and I get several rewraps. How is it possible ?

tirix:
Indeed the comment's bibliographic reference is not well-formed. I'll fix that, but I have no idea why the CI did not detect it.

jkingdon:
> 



> Indeed the comment's bibliographic reference is not well-formed. I'll fix that, but I have no idea why the CI did not detect it.

That's mysterious to me too. I can see it in CI sometimes (for example https://github.com/jkingdon/set.mm/actions/runs/1597980267 but not others, for example https://github.com/metamath/set.mm/runs/4573373217?check_suite_focus=true  ). @david-a-wheeler do you know what would be different between those?

david-a-wheeler:
Would someone write the ci miss as a separate issue, and then assign to me? I would like to fix it, but it needs to be tracked separately so we won't forget about it.

jkingdon:
> Would someone write the ci miss as a separate issue, and then assign to me? I would like to fix it, but it needs to be tracked separately so we won't forget about it.

Done at #2377

PR #2365, opened 2021-12-10
Remove unnecessary braces

jamesjer:
This PR updates scripts/check-braces to avoid false positives on curly braces in comments.  The other 3 commits remove braces flagged by the script in nf.mm, iset.mm, and set.mm.  I have included mathboxes, but will modify the commit at the request of mathbox owners.

### DIFF
--- a/nf.mm
+++ b/nf.mm
@@ -8105,13 +8105,11 @@ $)
       ( w3a wa pm3.2i df-3an mpbir2an ) ABCGABHCABDEIFABCJK $.
   $}
 
-  ${
-    $( ~ pm3.2 for a triple conjunction.  (Contributed by Alan Sare,
-       24-Oct-2011.) $)
-    pm3.2an3 $p |- ( ph -> ( ps -> ( ch -> ( ph /\ ps /\ ch ) ) ) ) $=
-      ( wa w3a wi pm3.2 ex df-3an bicomi syl8ib ) ABCABDZCDZABCEZABCMFLCGHNMABC
-      IJK $.
-  $}
+  $( ~ pm3.2 for a triple conjunction.  (Contributed by Alan Sare,
+     24-Oct-2011.) $)
+  pm3.2an3 $p |- ( ph -> ( ps -> ( ch -> ( ph /\ ps /\ ch ) ) ) ) $=
+    ( wa w3a wi pm3.2 ex df-3an bicomi syl8ib ) ABCABDZCDZABCEZABCMFLCGHNMABCIJ
+    K $.
 
   ${
     3jca.1 $e |- ( ph -> ps ) $.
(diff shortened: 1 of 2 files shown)


avekens:
Great work, thanks a lot!
Although I think that sometimes it was intended to use the curly braces just to group related theorems (without the need for common hypotheses or distinct variables), such cases were rare, and in my opinion they have no significant meaning.

jamesjer:
If any are intended, I'm happy to remove them from this pull request.  I think this work has relatively low value, so I don't want it to interfere with anything others are trying to do.

david-a-wheeler:
I like this in principle. However, this could be very disruptive in terms of interfering with other merges.

I propose doing this in stages: First merge the script, then have a warning on the mailing list of what's going down. Then merge one database update at a time (start with a small less-busy one). Maybe even merge *parts* of databases.

That way, we can revert some parts without losing the whole thing.

In particular, if we merge the script *first* as its own thing, we can often repair a marge conflict by re-running the script on the "not yet updated" code.

jkingdon:
> I like this in principle. However, this could be very disruptive in terms of interfering with other merges.

I'm fine with breaking this apart into a series of smaller changes. That would potentially solve a number of problems as @david-a-wheeler suggests.

I'm also OK (if others are) with just ripping off the bandaid and merging it as written. It may seem large but it only touches something like 1% of the lines in set.mm (if I am correctly reading the diff display correctly).

jkingdon (approved):

sctfn:
Let's merge the other stuff in the queue first to make sure no conflicts
arise.

david-a-wheeler:
@sctfn :

> Let's merge the other stuff in the queue first to make sure no conflicts arise.

Too late, there's already a conflict :-). Besides, we can't know what conflicts could happen to changes being *currently* developed.

Anyway, the fact that we *already* have a conflict suggests, to me, that doing it incrementally would be better.

Don't get me wrong, I'm fully supportive of this nice cleanup by @jamesjer . But it touches so many things that I think incremental would be better.

tirix:
I also like the idea of this script. I can't think of any actual case where we'd use braces to only _logically_ group theorems. We do have chapters, sections, subsections and sub-sub-sections for that.

I think I mentioned on the mailing list that I believe this could even be part of the CI.

> Too late, there's already a conflict

Since this is "just" a script, I suppose it would be easy to just cancel this PR and relaunch it once things have calmed down and the PR queue is empty?

david-a-wheeler:
> Since this is "just" a script, I suppose it would be easy to just cancel this PR and relaunch it once things have calmed down and the PR queue is empty?

Cancelling is easy. But it's more than the PR queue, it's the queue of things people are working on that *aren't* being posted for a PR.

If we submit *just* the script first, & let that get distributed to people first, then I think it'll be easier to handle.

david-a-wheeler:
Oh - @jamesjer - thanks for this work!!

jamesjer:
Sorry, I've been away.  Everything that was said makes sense.  I'll cancel this pull request and start with just the script, followed by small chunks at a time as suggested.

david-a-wheeler:
Sounds good!

Once this script & all its changes are merged, let's revisit the idea of adding this script to the CI pipeline. Noticing & complaining about unnecessary braces, at least for set.mm and iset.mm, seems reasonable to me.

PR #2366, opened 2021-12-12
Minimize proofs in Main containing idi

wlammen:


benjub:
Did you do a "minimize *" on the output of "show usage idi" ? I think you can do the same with all such theorems, including those in mathboxes.  Proofs are not systematically minimized when there are moved from mathboxes to Main, so this would take care of them preemptively.

wlammen:
No, I limited minimize to Main.  The idea was better not interfere with some work in progress.  Mathboxes have about 15 more containing idi.

david-a-wheeler (approved):
Removing idi makes sense to me, unless they were examples to show the use of idi.

jkingdon (approved):
I'm operating under the assumption that it is always OK to run minimize and I don't see any reason why these should have "proof modification is discouraged" on them.

wlammen:
See the proof of jccil to see a valid reason for a "Proof modification is discouraged" tag

tirix:
I am the contributor for ~ opphllem2 and as far as it is concerned, no objection to remove the `idi`. 
Thanks for noticing and fixing it!

tirix (approved):

PR #2368, opened 2021-12-14
Applications of the circle method

tirix:


tirix:
Sorry the git commit history is messed up, this PR actually only includes the last few commits. 
This can be squashed.

jkingdon:
> Sorry the git commit history is messed up, this PR actually only includes the last few commits. This can be squashed.

Hmm, I didn't look too deeply into what happened, but I did just squash it as suggested.

PR #2374, opened 2021-12-19
infimum for iset.mm

jkingdon:
This is very similar to the supremum section and differs from set.mm in just about the same ways (most notably, the condition on the order is different but would be satisfied by, for example, real numbers in both cases). I don't think there are really any surprises relative the supremum section - just about the same things are present or not present or modified.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3061,6 +3061,11 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>eqinf</TD>
+  <TD>~ eqinfti</TD>
+</TR>
+
+<TR>
 <TD>ax-reg , axreg2 , zfregcl</TD>
 <TD>~ ax-setind </TD>
 <TD>ax-reg implies excluded middle as seen at ~ regexmid</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3150,6 +3150,12 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>infltoreq</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof depends on supgtoreq and fiinfcl</TD>
+</TR>
+
+<TR>
 <TD>ax-reg , axreg2 , zfregcl</TD>
 <TD>~ ax-setind </TD>
 <TD>ax-reg implies excluded middle as seen at ~ regexmid</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3055,6 +3055,12 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>infexd</TD>
+  <TD><I>none</I></TD>
+  <TD>See supexd</TD>
+</TR>
+
+<TR>
 <TD>ax-reg , axreg2 , zfregcl</TD>
 <TD>~ ax-setind </TD>
 <TD>ax-reg implies excluded middle as seen at ~ regexmid</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3109,6 +3109,11 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>infmo</TD>
+  <TD>~ infmoti</TD>
+</TR>
+
+<TR>
 <TD>ax-reg , axreg2 , zfregcl</TD>
 <TD>~ ax-setind </TD>
 <TD>ax-reg implies excluded middle as seen at ~ regexmid</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3167,6 +3167,11 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>infiso</TD>
+  <TD>~ infisoti</TD>
+</TR>
+
+<TR>
 <TD>ax-reg , axreg2 , zfregcl</TD>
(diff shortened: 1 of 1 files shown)


benjub:
Is it really worth having a new token for infimum, where one may simply use the sup for the opposite order ? In terms of length of the expression, basically nothing is changed (only one more character for "opposite"), and this would double the number of library theorems. If I recall correctly, set.mm has only supremum, and not infimum.

**Edit:** I checked: the definition of infimum has actually been added in set.mm too: http://us.metamath.org/mpeuni/df-inf.html (from 2020). Not sure it was the best move, and I let @jkingdon see if he wants to replicate this in iset.mm.

jkingdon:
> the definition of infimum has actually been added in set.mm too: http://us.metamath.org/mpeuni/df-inf.html (from 2020). Not sure it was the best move, and I let @jkingdon see if he wants to replicate this in iset.mm.

I have a pretty strong desire to follow set.mm on matters where following set.mm is feasible (this rule has been written down for a while as "We do try to keep set.mm and iset.mm similar where we can" in a comment near the top of iset.mm). I'm about to formalize least common multiple, based on the set.mm theorems, and I'm not sure everyone appreciates how much of that kind of work is just copy-paste, either without any change at all, or with pretty limited changes.

There are some harder cases (for example, I'm not sure proper subset is doing any good in iset.mm as currently defined, and a few things like `We` or maybe `Or` operate differently enough that I'm not convinced that our current definitions are helping us much). But those are ones which are much more affected by excluded middle than the issue of whether infimum or supremum have one notation or two.

As for why set.mm got the infimum notation, there didn't seem to be much dissent when it was suggested at https://github.com/metamath/set.mm/issues/1805 (I looked over the mailing list archives linked there and I saw a lot more discussion of least common multiple notation than whether the infimum thing is a good idea). I can see arguments both ways so I'm not sure I have a huge amount to say on that topic.

benjub:
Fair enough for following set.mm (maybe we should have given more thought before introducing infimum there).

As for strict subsets, I think that indeed the definition should require that the difference be inhabited (if this definition is needed at all).

I just had a look at http://us2.metamath.org:88/ileuni/df-iso.html which indeed requires something weaker than trichotomy (~issod), and strictly so.  Is this an interesting notion ? I do not know, but it certainly is not as interesting as trichotomy.  An intermediate property of a partial order is
> A. x e. A A. e. A y A. z e. A ( -. x R y \\/ x R z \\ z R y )

which implies the property in df-iso by propcalc, which is implied by trichotomy, and which, importantly, implies
> ( ... -> DECID A R B )

Note that an empty relation satisfies this property but not trichotomy.

avekens:
There was a discussion in the google group https://groups.google.com/g/metamath/c/p5XIt_HdDik/m/pg84xtI2AAAJ , and it seemed that there was a consent to add a definition for the infimum. The main reasons (for me) ware that almost in all comments of theorems using the "infimum" it had to be explained, and that it was very confusing to "reverse" the order in one's head if you want to understand the theorem.

jkingdon:
> I just had a look at http://us2.metamath.org:88/ileuni/df-iso.html which indeed requires something weaker than trichotomy (~issod), and strictly so. Is this an interesting notion ?

Weak linearity is interesting in the sense that `show usage sowlin/recursive` shows 138 usages and `show usage axltwlin/recursive` shows another 1909.

> I do not know, but it certainly is not as interesting as trichotomy.

Trichotomy is indeed interesting and we have quite a few usages of theorems like `ztri3or` and `qtri3or` .

Another interesting property is the one seen at http://us.metamath.org/ileuni/lttri3.html which, like weak linearity but unlike trichotomy, holds for the iset.mm real numbers. This one features in theorems like http://us.metamath.org/ileuni/supclti.html .

Trichotomy, of course, implies the other two (as you point out) but I'm not sure whether weak linearity implies the `lttri3` property or vice versa.

Whether any or all of these notions deserve their own notation, I'm less clear on, much less whether the notation for any of them should share `Or` with set.mm.

jkingdon:
Can I get a review on this? Tests are passing and it should be ready to merge because the math here is just what is in set.mm and already in iset.mm.

iset.mm
```diff
@@ -61068,6 +61078,248 @@ all reals (greatest real number) for which all positive reals are greater.
...
+    infsnti.b $e |- ( ph -> B e. A ) $.
+    $( The infimum of a singleton.  (Contributed by Jim Kingdon,
+       19-Dec-2021.) $)
+    infsnti $p |- ( ph -> inf ( { B } , A , R ) = B ) $=
+      ( csn cinf ccnv csup df-inf cnvti supsnti syl5eq ) AEIZDFJQDFKZLEQDFMABCD
+      ERABCDFGNHOP $.
+  $}
+
+  $( The infimum regarding an empty bas set is always the empty set.
+     (Contributed by AV, 4-Sep-2020.) $)
```
avekens, line 61301: typo bas -> base (I'll correct this in set.mm)

avekens (commented):
Please correct the typo, the rest is OK.

iset.mm
jkingdon: I've put a fix for both set.mm and iset.mm into this pull request (if there is some reason why fixing it for set.mm here is a bad idea, I can change it, but seemed easier to do both at once).

avekens (approved):
It's OK now.

PR #2376, opened 2021-12-20
Remove proper subclass from iset.mm

jkingdon:
This is little used and it seems like every usage in set.mm which
I have tried to intuitionize to date hasn't worked in iset.mm (which is basically why this diff is so small - theorems which include a reference to proper subclass seem not to survive the intuitionizing process so there are few occurrences).

As far as I know @digama0 and @benjub have both expressed support for getting rid of the current definition (and, if we add it back, adding it back with the definition about the set difference being inhabited). Most recently this was discussed at https://github.com/metamath/set.mm/pull/2374#issuecomment-997424408

My thinking is that we don't have an immediate need for the definition about set difference being inhabited (that I'm aware of anyway), and so the best thing to do for now is remove proper subclass entirely, make a note of it in mmil.html, and bring it back in the future if it seems like we are writing out ` E. x x e. ( B \ A ) ` a lot.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1379,18 +1379,28 @@ is double negation elimination.</TD>
 </TR>
 
 <TR>
+  <TD>df-pss and all proper subclass theorems</TD>
+  <TD><I>none</I></TD>
+  <TD>In set.mm, "A is a proper subclass of B" is defined to be
+  ` ( A C_ B /\ A =/= B ) ` and this definition is almost always used in
+  conjunction with excluded middle. A more natural definition
+  might be ` ( A C_ B /\ E. x x e. ( B \ A ) )` , if we need proper subclass
+  at all.</TD>
+</TR>
+
+<TR>
   <TD>nss</TD>
   <TD>~ nssr</TD>
 </TR>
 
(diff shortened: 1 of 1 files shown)


benjub:
I would propose:
> ~ df-ss3 which --> ~ df-ss3 , which
> positive real's lower cut --> lower cut of positive reals
> positive real's upper cut --> upper cut of positive reals

though these can of course be done later.

jkingdon:
> > ~ df-ss3 which --> ~ df-ss3 , which

I pulled over the whole sentence from set.mm which is worded differently (and then trimmed the list of theorems to those which exist in iset.mm).

> > positive real's lower cut --> lower cut of positive reals
> > positive real's upper cut --> upper cut of positive reals

Changed to "The lower cut of a positive real" and "The upper cut of a positive real". Each real is a pair of a lower cut and an upper cut, so "positive reals" is not correct here.

The pull request is now updated with these changes.

benjub:
Ah, so I had misunderstood ! It is actually the lower/upper cut defining a positive real. All is clear, now.

benjub (approved):

PR #2378, opened 2021-12-20
Extend decimal point and Misc.

tirix:
In this PR, I'm pulling the chapter "Decimal Point" from David's Mathbox into mine, and extend it with a few theorems.

Tough, I'd like to propose a slightly different construct for decimal point:
My proposition is to add a marker _after_ the decimal expansion, like in ` ( 3 . 1 4 1 5 9 : : : : ) ` . 
I believe this is a bit more readable than the original proposal which would result in interleaving underscores in the decimal expansion, as in ` ( 3 . _ 1 _ 4 _ 1 _ 5 9 ) ` .

I've kept both constructions (previous one with underscore as `df-dp2` , and new one as `dp-dp3`), and I've also kept the attributions to David.

Apart from a few generic theorems moved from other people's mathboxes to the main part, the other changes are within my mathbox.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -5682,6 +5682,9 @@
 "e3bi" is used by "en3lplem2VD".
 "e3bir" is used by "en3lplem2VD".
 "edgfiedgvalOLD" is used by "structgrssiedgOLD".
+"edgvalOLD" is used by "edg0iedg0OLD".
+"edgvalOLD" is used by "edgiedgbOLD".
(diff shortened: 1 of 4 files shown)


tirix:
There seems to be some issue with discouraged, although I don't get it when I regenerate the discouraged file locally.

tirix:
Same changes are in #2379 , from a clean branch.

PR #2379, opened 2021-12-20
Extend decimal point and Misc.

tirix:
In this PR, I'm pulling the chapter "Decimal Point" from David's Mathbox into mine, and extend it with a few theorems.

Tough, I'd like to propose a slightly different construct for decimal point:
My proposition is to add a marker _after_ the decimal expansion, like in ` ( 3 . 1 4 1 5 9 : : : : ) ` . 
I believe this is a bit more readable than the original proposal which would result in interleaving underscores in the decimal expansion, as in ` ( 3 . _ 1 _ 4 _ 1 _ 5 9 ) ` .

I've kept both constructions (previous one with underscore as `df-dp2` , and new one as `dp-dp3`), and I've also kept the attributions to David.

Apart from a few generic theorems moved from other people's mathboxes to the main part, the other changes are within my mathbox.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,15 +1,66 @@
-If you are new to GitHub and git, the basic instructions for
-contributing changes to set.mm are described in the wiki,
-[Getting started with contributing](https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing).
-(Feel free to enhance the wiki.)
+# Contributing and Acceptance
 
-The rest of this note shows how to achieve the recommended
-formatting for set.mm and iset.mm prior to submitting your changes.
+We're a friendly community, and we would love to have more collaborators!
 
-### Formatting recommendation prior to submitting a pull request
+This document briefly explains
+[how to contribute proposed changes](#how-can-i-contribute)
+("contributions" or "merge requests") to this collection of databases
+We also discuss
+[Formatting recommendation prior to submitting a pull request](#formatting-recommendation-prior-to-submitting-a-pull-request), and for the rare cases
+when it's needed, information on
+[regenerating the `discouraged` file](#regenerating-the-discouraged-file).
+This page also describes
+[how these proposed changes are evaluated](#what-are-the-requirements-for-acceptance-merging)
+to decide if they will be accepted, since presumably you'll want to
+meet those criteria when creating a contribution.
+
+## How can I contribute?
+
+First, take a look at some existing proofs.
+Walking through the
+[Metamath Proof Explorer (MPE, aka set.mm)](http://us.metamath.org/mpeuni/mmset.html) is a good way to start.
+
+To create a proof, you need to pick a tool.
+The Metamath system specifies a common file format, not a single tool
+everyone has to use.
+Many in the community use the [mmj2 tool](http://us.metamath.org#book), and
+many of the rest use the original metamath C program (aka "metamath.exe").
+The page
+[Metamath Proof Assistants](https://github.com/metamath/set.mm/wiki/Metamath-Proof-Assistants)
+provides more information about each tool.
+The video [Introduction to Metamath and mmj2](https://www.youtube.com/watch?v=Rst2hZpWUbU) demonstrates using mmj2.
+The mmj2 tool includes an interactive tutorial, and if you just want
+to watch it, there's a video giving a
+[Walkthrough of the tutorial in mmj2](https://www.youtube.com/watch?v=87mnU1ckbI0).
+
+You should start small and try to reprove what's already
+proven with Metamath before you try to prove something new.
+If you're contributing to set.mm or iset.mm, take a look at its
+[conventions and style](http://us.metamath.org/mpeuni/conventions.html).
+
+At some point you should join the
+[Metamath mailing list](https://groups.google.com/g/metamath).
+If you're seriously working on proofs, we'd love
+to help you get started!
+
+
+For contributing a final proof once you have one,
+you'll basically create a "pull request" on GitHub to request
+merging of some change(s) into the corresponding Git repository.
+**Warning**: Changes are made to the **develop** branch.
+The Wiki page
+[Getting started with contributing](https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing) will walk you the process of contributing
+a final proof once you have one.
+
+**NOTE**: Any proposed contribution must be licensed under the
+[Creative Commons 0 (CC0) License](LICENSE), so any proposed contribution
+is automatically licensed as such.
+
+## Formatting recommendation prior to submitting a pull request
 
 Periodically we rewrap set.mm to help conform to its formatting conventions.
-This may affect your mathbox if you submitted it without rewrapping, possibly
+This may affect your mathbox (your local sandbox)
+if you submitted it without rewrapping, possibly
 causing merge conflicts with your work in progress.
 
 Here is the procedure recommended prior to submitting a pull request:
@@ -55,17 +106,21 @@ indentation.  Finally, 'verify proof' is there because you might as well.
 
 (You can also check definitional soundness with mmj2 and any 'discouraged'
 markup changes before submission if you want, or you can just leave it up to
-Travis to check those.)
+the automated verification checks to check those.)
 
+## Regenerating the `discouraged` file
 
-### Regenerating the `discouraged` file
+In almost all cases, especially when you're starting, you won't need to
+regenerate the `discouraged` file. So if you're just starting out, you can
+probably ignore this section.However, there are rare cases when that
+is needed. This section explains what the discouraged file is all about.
 
 Some statement descriptions in `set.mm` and ` iset.mm` have one or both of the
 markup tags `(New usage is discouraged.)` and `(Proof modification is discouraged.)`.
 In order to monitor accidental violations of these tags in set.mm and iset.mm,
 we store the usage and proof lengths of statements with these tags in a file
 called `discouraged` for set.mm and `iset-discouraged` for iset.mm.
-The Travis check will return an error if this file does not exactly match the
+The automated check will return an error if this file does not exactly match the
 one that it generates for a set.mm pull request.
 
 If you make modifications that affect the `discouraged` file, you should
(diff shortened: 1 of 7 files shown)


tirix:
It looks like the CI does not fail because of any issue with discouraged theorems, but rather because the `discouraged` file could not be built:
```
Run echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
  echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_HOME_10.0.2_x64: /opt/hostedtoolcache/jdk/10.0.2/x64
    JAVA_HOME: /opt/hostedtoolcache/jdk/10.0.2/x64
    JAVA_HOME_10_0_2_X64: /opt/hostedtoolcache/jdk/10.0.2/x64
Checking discouraged file:
diff: discouraged.new: No such file or directory
```
Actually, MMJ2 crashes when generating the file:
```
java.lang.NullPointerException
	at mmj.verify.LRParser.initialize(LRParser.java:175)
	at mmj.verify.LRParser.load(LRParser.java:320)
	at mmj.verify.LRParser.load(LRParser.java:317)
	at mmj.verify.LRParser.parseExpr(LRParser.java:382)
	at mmj.verify.Grammar.grammaticalParseSyntaxExpr(Grammar.java:732)
	at mmj.verify.GrammarAmbiguity.basicAmbiguityEdits(GrammarAmbiguity.java:208)
	at mmj.verify.Grammar.initializeGrammarTables(Grammar.java:1008)
	at mmj.verify.Grammar.initializeGrammar(Grammar.java:704)
	at mmj.util.GrammarBoss.initializeGrammar(GrammarBoss.java:288)
	at mmj.util.GrammarBoss.doParse(GrammarBoss.java:242)
	at mmj.util.Boss.lambda$0(Boss.java:107)
	at mmj.util.Boss.doRunParmCommand(Boss.java:121)
	at mmj.util.BatchFramework.executeRunParmCommand(BatchFramework.java:281)
	at mmj.util.BatchFramework.runIt(BatchFramework.java:223)
	at mmj.util.BatchMMJ2.main(BatchMMJ2.java:53)
```
I regenerated it locally, but with the Metamath program.

benjub:
Wouldn't this make the grammar more complicated and the file harder to parse ?  Are there other cases in set.mm of n-ary postfix notation with n>1 ?  It seems that it would introduce the need of unbounded look-ahead.

tirix:
That might be the reason, but actually all (other) verifiers are fine with it, and I've used "E-Metamath" to develop it, which is based on MMJ2, so I'm puzzled why this would happen for generating "discouraged" and not for the rest. 
I'd like to take time to review and step-through MMJ2 to fully understand the issue, but I did not have the opportunity yet.

digama0:
I'm glad the LRParser caught this. This is indeed not an allowed grammar construction, because it violates condition (ii) of theorem 4 in http://us.metamath.org/downloads/grammar-ambiguity.txt : no proper prefix of a class expression is a class expression. This can lead to ambiguous parses.

The reason it was not caught by other verifiers is that grammar ambiguity is difficult to check and not required by the metamath spec, but it can lead to misleading theorems. Most metamath verifiers / proof assistants either don't do math parsing at all or parse dynamically, which means that unambiguity is only checked if it actually occurs in the statements that appear in the database (and sometimes not even then; many parsers just take some parse at random). LRParser is the only existing metamath parser that can check unambiguity up-front, which is part of why it is in CI.

For a concrete example of an ambiguous expression, consider `; A B C :`, which can be either `cdc cA (cdp3 cB cC)` or `cdp3 (cdc cA cB) cC`. This one is "benign" since the two expressions are provably equal, but `-u A B :` has the same issue and is not benign.

david-a-wheeler:
Yes, I agree with @digama0 - we prevent ambiguous parses, by ensuring that constructs can be unambiguously resolved after a certain number of tokens (I believe the number is 5 in parse system). Which is why I defined `_` in this way. Alternatives are worth considering, but I don't think creating an ambiguous parse is worth it.

david-a-wheeler:
BTW, just to make it clear, you are *absolutely* welcome to move decimal point from my mathbox to yours & extend it. I think it'd be good to have *some* support for the decimal point in the main body; I hear they are used sometimes :-).

tirix:
Thank you Mario for the explanation! 

I would not say that the LRParser caught it, or at least it did not catch it gracefully, since it crashed without clear message: maybe we could improve the error handling / error message? (interestingly, metamath-knife also panicked without any explanation, so there is also room for improvement there!)

I'll see what I can do!

tirix:
Thanks Mario for the additional explanation, especially the example of ambiguous expression: this proves that this construct is ambiguous.

And thanks David for allowing me to experiment with the section you have initiated!

**TL; DR: The last commit reverts to David's original construct and passes all checks. I propose to first move on with it.**

I think there is some value in keeping the decimal expansion uninterrupted, in order to stay near to the usual decimal representation, which is something Norm valued.

I'm taking the opportunity to list my thoughts about different alternatives:
Reminder: With the current solution, two examples of expressions representing decimal numbers would be
` ( 3 . _ 1 _ 4 _ 1 _ 5 9 ) ` and ` ( ; ; 1 0 0 . _ 0 _ 0 1 ) ` .

### Solution 1: A new type code
I believe I have seen somewhere Norm or Mario tried this without success but can't find where anymore. A new type code would allow to define completely new parsing constraint, including a suffix notation. However, it would break the beautifully minimal wff/set/class triplet, and is way too heavy-handed for this simple target.

### Solution 2: Using 3 new constructs
Since we can only group uninterrupted list of tokens left-to right, for decimals, we have to know what place is the next one. One could use pairs to encode that, with the first element of the pair being the current value, and the second being the current place. This would require 3 constructs:

* an initial one to construct the first pair: e.g. ` A . B `
* an iteration step which adds a number to the decimal expansion: ` : A B `
* a final step which unpacks the pair and drops the second element: ` ." A ". `

A resulting decimal expression would be something like ` ." : : : : 3 . 1 4 1 5 9 ". ` or ` ." : : ; ; 1 0 0 . 0 0 1 ". `
Here is what it might look like:
```
$c ." ". . $.
$a class A . B $.
$a class : A B $.
$a class  ." A ". $.
$a |- A . B = <. ( A + ( B / ; 1 0 ) ) , 2 .> $.
$a |- : A B = <. ( ( 1st ` A ) + ( B x. ( ; 1 0 ^ -u ( 2nd ` A ) ) ) ) , ( ( 2nd ` A ) + 1 ) .> $.
$a |- ." A ". = ( 1st ` A ) $.
```
I believe this is too convoluted, and hard to understand.

### Solution 3: Using words
Word constructs have the nice property that the string of tokens is uninterrupted. With this definition, and no new construct:
```
$a |- . = ( a e. NN0 , b e. Word NN0 |-> ( a + sum_ i e. dom b ( ( b ` i ) x. ( ; 1 0 ^ -u ( i + 1 ) ) ) ) $.
```
We could write decimal numbers as ` ( 3 . <" 1 4 1 5 9 "> ) ` or ` ( ; ; 1 0 0 . <" 0 0 1 "> ) `
Words are currently limited to 8 places, but this would cover most usages. The advantage here is that we don't need to define any new syntax construct. Calculus might however be unnecessarily complicated.

### Solution 4: Word-like constructs
One issue with the previous syntax is that the fist ` <" ` token comes in the middle of the decimal expression. One could define a series of constructs of different lengths, just like for words, for decimal expansions of different lengths.
```
$c . ." ". $.
$a class ." A . B ". $.
$a class ." A . B C ". $.
$a class ." A . B C D ". $.
etc.
```
I think the issue here is that when moving to doing calculus, these may then require an onslaught of theorems in order to cover all cases, or conversion theorems to move from one form to another one. But the advantage here is that the final syntax would be very readable: ` ." 3 . 1 4 1 5 9 ". ` which is my graal here, or ` ." ; ; 1 0 0 . 0 0 1 ". `. 

One may then wish to extend this to also get rid of the initial ` ; ` notation, allowing for example to write ` ." 1 0 0 . 0 0 1 ". `. Unfortunately the number of cases then explodes quadratically if we are to allow any number of places before and after the dot.

In this case it is interesting to use this kind of definition:
```
$a |- ." A . B C D ". = ( A + ( ." B . C D ". / ; 1 0 ) ) $. 
```
rather than something like:
```
$a |- ." A . B C D ". = ( ." A . B C ". + ( D / ; ; ; 1 0 0 0 ) ) $. 
```
Because then, all definitions have the same form ` ( A + ( B / ; 1 0 ) ) ` , which can become a common essential hypothesis and lead to common generic theorems for all syntax forms. This sounds like an interesting idea and I might experiment with it in the future, but for the moment I'm rather interested in the application than the actual decimal expansions, and I'd like to move on with the current definition.

digama0:
Personally, I have found decimal numbers to be of very limited use, and I don't think that any of the proposed alternatives are better than David's solution, but I would suggest rather than any of them to simply use rational expressions like `( 2 5 3 ; ; / 3 5 7 ; ; )`, which are much simpler to work with in all respects, and can be used for much more accurate approximations than decimals admit. (More precisely, you generally need numbers with half as many digits to represent a random real to a given precision, compared to decimals; multiplications on such numbers are therefore 4x smaller and faster.)

tirix:
Yes. Here, I specifically wanted to stick as much as possible to the way the theorems were expressed in the original article, and they were using decimal expansions.

You can find these theorems further down in this PR's change set, some of them still commented out.

david-a-wheeler:
@tirix - merge conflicts!

david-a-wheeler:
Oh, and "fration" should be "fraction". That's my fault, but since you're there you can fix it :-).

tirix:
Merge conflict is resolved and typo is fixed.
I think discussions are closed so this would be ready to be merged!

set.mm
```diff
@@ -489156,6 +489830,362 @@ of the indicator function (Contributed by Thierry Arnoux,
...
+      ad2antlr r19.21bi simpr2 simpr3 tgoldbachgtda hgt749d r19.29vva ) AHJZUAJ
+      ZKUCKLMUDUHUIUINNNNUEOUFPZHUGUJZWFUBJZKUCKLUKLUKNNUEOUFPZHUGUJZMMMMUKQQUK
+      ULNNNNNNNUEOCQUMOROIMLUNOIJZUOWJRUPZOCUQOKWMUOWGWNOCUQOKQUMOROURQUSROROCU
+      TWMROROVAKROVFUFPZVBZMDVCVDCVEVGKOVHKVIPUBUASTVJOZWQAWJWQVKZVLZWGWQVKZVLZ
+      WPVLZIBHWJWGCDEACDVKWRWTWPFVMALMVNQUDVNUMOCUFPWRWTWPGVMWRTSWJVOAWTWPWJSTV
+      PVQWTTSWGVOWSWPWGSTVPVSXBWHHUGXAWIWLWOVRVTXBWKHUGXAWIWLWOWAVTXAWIWLWOWBWC
+      AIBUBUAHCDEFGWDWE @.
+  @}
+$)
+
```
avekens, line 490188: That would be really great if you can prove this! It corresponds to ~ax-tgoldbachgt in my mathbox, which I provided as axiom because I coud not prove it. As soon as you have finished your work, we should talk about merging our results, and maybe move them to the main body of set.mm. Of course we have to harmonize/discuss the definition of odd numbers (my class ` Odd ` and your class ` O = { z e. ZZ | -. 2 || z } `, see discussion in #1682).

avekens (approved):

set.mm
```diff
@@ -465654,6 +465751,432 @@ its graph has a given second element (that is, function value).
...
+    $( Equality theorem for the decimal expansion constructor.  (Contributed by
+       David A. Wheeler, 15-May-2015.) $)
+    dp2eq12i $p |- _ A C = _ B D $=
+      ( cdp2 dp2eq1i dp2eq2i eqtri ) ACGBCGBDGABCEHCDBFIJ $.
+  $}
+
+  ${
+    dp20u.1 $e |- A e. NN0 $.
+    $( Add a zero in the tenths (lower) place.  (Contributed by Thierry Arnoux,
+       16-Dec-2021.) $)
```
benjub, line 465822: tenths ?
"Adding a zero as the least significant digit after a decimal separator does not change the value of a number." ?

set.mm
```diff
@@ -465654,6 +465751,432 @@ its graph has a given second element (that is, function value).
...
+    $( Add a zero in the unit places.  (Contributed by Thierry Arnoux,
+       16-Dec-2021.) $)
+    dp20h $p |- _ 0 A = ( A / ; 1 0 ) $=
+      ( cc0 cdp2 c1 cdc cdiv co caddc df-dp2 crp wcel cc ax-mp 10nn0 nn0cni 0re
+      rpcn 10pos gtneii divcli addid2i eqtri ) CADCAECFZGHZIHUECAJUEAUDAKLAMLBA
+      RNUDOPCUDQSTUAUBUC $.
+  $}
+
+  $( Define the closure for the decimal fraction constructor if both values are
+     reals.  (Contributed by David A. Wheeler, 15-May-2015.) $)
```
benjub, line 465840: Define the closure for --> Closure of/for
?

set.mm
```diff
@@ -465654,6 +465751,432 @@ its graph has a given second element (that is, function value).
...
+       ` -u ( ; 3 2 . _ 7 _ 1 8 ) = -u ( ; ; ; ; 3 2 7 1 8 / ; ; ; 1 0 0 0 ) `
+       Unary minus, if applied, should normally be applied in front of the
+       parentheses.
+
+       Metamath intentionally does not have a built-in construct for numbers,
+       so it can show that numbers are something you can build based on set
+       theory.  However, that means that metamath has no built-in way to handle
+       decimal numbers as traditionally written, e.g., "2.54", and its parsing
+       system intentionally does not include the complexities necessary to
+       define such a parsing system.  Here we create a system for modeling
```
benjub, line 465959: "parsing system" twice ?

set.mm
```diff
@@ -465654,6 +465751,432 @@ its graph has a given second element (that is, function value).
...
+
+  Define the decimal point operator and the decimal fraction constructor.
+  This can model traditional decimal point notation, and serve
+  as a convenient way to write some fractional numbers.
+  See ~ df-dp and ~ df-dp2 for more information; ~~? dfpval
+  provides a more convenient way to obtain a value.
+  This is intentionally similar to ~ df-dec .
+
+  TODO: Fix non-existent label dfpval.
+
```
benjub, line 465937: Is there a dfpval to be added, or is it simply dpval ? (in which case, you may remove the TODO and the "~~?" ).
You may also rewrap this to width 79, and have a double blank line before that subsection.

benjub (approved):
I made some comments that you may follow (or not) in a later PR.

set.mm
tirix: Yes, the tenths place! 
See for example [https://en.wikipedia.org/wiki/Numerical_digit#Computation_of_place_values](the Wikipedia for the place value). In french we might say "les unités, les dixièmes, les centièmes"

set.mm
tirix: Agreed! This will be fixed in my next commit.

set.mm
tirix: This is originally for David, but I agree with your comment, and it will be changed it in my next commit.

set.mm
tirix: This is also from @david-a-wheeler . In the line above, there is a reference to a ~ dpfrac theorem which had not been proven yet.  In my next commit, I'm removing this TODO and I refer to ~ dfpval2 and ~ dpfrac1.

david-a-wheeler:
@benjub - feel free to fix it up as you think best. I'm just glad someone else also thinks it has some utility!

PR #2380, opened 2021-12-20
axc16g: use ax12v instead of ax12v2

benjub:
The weaker version ax12v (instead of ax12v2) is sufficient.


wlammen:
benjub is right here.  It is sensible to found proofs on the more specific theorem, if there is a choice.  This will result in some cases in smaller definition and axiom footprint. But even if not, it is a good rule of thumb to only create dependencies on what you cannot avoid.

tirix (approved):

wlammen (approved):

avekens (approved):

PR #2382, opened 2021-12-20
iset: clean OLD theorems

benjub:


jkingdon (approved):
Nice work! I've been hoping someone would get around to this for a while.

jkingdon:
This seems straightforward but if anyone has any issues speak up and we can make a follow up change if need be.

PR #2383, opened 2021-12-20
Graph theory: sum of vertex degrees is twice the number of edges

avekens:
Changes in parts before GRAPH THEORY
* ~bj-eleq1w (renamed ~eleq1w) moved from BJ's mathbox to main: @benjub please check if this is the right place, and if the comment should be adjusted esp. if the tag "(Proof modification is discouraged.)" should be removed.
* ~rabeqi, ~rabeqif, ~ bn21 (renamed ~ssrab3), ~funfnd, ~ elrabd moved from mathboxes to main
* ~elneldisj and ~elnelun generalized
* comment for df-pr enhanced ("proper unordered pairs" explained)
* theorems ~prproe and ~3elpr2eq about proper unordered pairs added
* ~rabasiun removed (duplicate)
* ~hashres, ~hashreshashfun added
*  ~fsummsnunz, ~fsumsplitsnun generalized
* spelling of "power set" (vs. "powerset") harmonized

Changes in part GRAPH THEORY
* ~iedgedg, ~upgrop added
* ~edglnl, ~numedglnl, ~vtxdgop  added
* ~upgrres, ~umgrres, ~usgrres added
* ~vtxdginducedm1, ~vtxdginducedm1fi added
* ~finsumvtxdg2sstep and ~finsumvtxdg2size (main result of this PR, dedicated to Norm, because I think this is an example for his (and my!) motivation for working with Metamath.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2220,11 +2220,6 @@
 "bnj206" is used by "bnj207".
 "bnj207" is used by "bnj600".
 "bnj207" is used by "bnj908".
-"bnj21" is used by "bnj1212".
-"bnj21" is used by "bnj1286".
-"bnj21" is used by "bnj1312".
-"bnj21" is used by "bnj1523".
-"bnj21" is used by "bnj213".
 "bnj213" is used by "bnj1128".
 "bnj213" is used by "bnj1137".
 "bnj213" is used by "bnj1145".
@@ -14201,7 +14196,6 @@ New usage of "bnj170" is discouraged (6 uses).
 New usage of "bnj18eq1" is discouraged (1 uses).
 New usage of "bnj206" is discouraged (2 uses).
 New usage of "bnj207" is discouraged (2 uses).
-New usage of "bnj21" is discouraged (5 uses).
 New usage of "bnj213" is discouraged (8 uses).
 New usage of "bnj216" is discouraged (9 uses).
 New usage of "bnj219" is discouraged (4 uses).
@@ -15454,6 +15448,8 @@ New usage of "elintgOLD" is discouraged (0 uses).
 New usage of "ellnfn" is discouraged (5 uses).
 New usage of "ellnop" is discouraged (9 uses).
 New usage of "elmptrab2OLD" is discouraged (0 uses).
+New usage of "elneldisjOLD" is discouraged (0 uses).
+New usage of "elnelunOLD" is discouraged (0 uses).
 New usage of "elni" is discouraged (7 uses).
 New usage of "elni2" is discouraged (7 uses).
 New usage of "elnlfn" is discouraged (4 uses).
@@ -15606,7 +15602,9 @@ New usage of "fnexALT" is discouraged (0 uses).
 New usage of "foco2OLD" is discouraged (0 uses).
 New usage of "fprodcom2OLD" is discouraged (0 uses).
 New usage of "fsumcom2OLD" is discouraged (0 uses).
+New usage of "fsummsnunzOLD" is discouraged (0 uses).
 New usage of "fsumshftdOLD" is discouraged (0 uses).
+New usage of "fsumsplitsnunOLD" is discouraged (0 uses).
 New usage of "fsuppmapnn0fiubOLD" is discouraged (0 uses).
 New usage of "funadj" is discouraged (4 uses).
 New usage of "funcnvadj" is discouraged (1 uses).
@@ -18521,7 +18519,6 @@ Proof modification of "bj-dvelimdv1" is discouraged (63 steps).
 Proof modification of "bj-dvelimv" is discouraged (28 steps).
 Proof modification of "bj-eeanvw" is discouraged (32 steps).
 Proof modification of "bj-el" is discouraged (48 steps).
-Proof modification of "bj-eleq1w" is discouraged (51 steps).
 Proof modification of "bj-eleq2w" is discouraged (51 steps).
 Proof modification of "bj-elisset" is discouraged (24 steps).
 Proof modification of "bj-elissetv" is discouraged (25 steps).
@@ -18961,6 +18958,7 @@ Proof modification of "el12" is discouraged (19 steps).
 Proof modification of "el123" is discouraged (26 steps).
 Proof modification of "el2122old" is discouraged (25 steps).
 Proof modification of "elALT" is discouraged (27 steps).
+Proof modification of "eleq1w" is discouraged (51 steps).
 Proof modification of "eleq2dALT" is discouraged (62 steps).
 Proof modification of "elex22VD" is discouraged (111 steps).
 Proof modification of "elex2VD" is discouraged (59 steps).
@@ -18976,6 +18974,8 @@ Proof modification of "eliminable3a" is discouraged (7 steps).
 Proof modification of "eliminable3b" is discouraged (8 steps).
 Proof modification of "elintgOLD" is discouraged (47 steps).
 Proof modification of "elmptrab2OLD" is discouraged (73 steps).
+Proof modification of "elneldisjOLD" is discouraged (53 steps).
+Proof modification of "elnelunOLD" is discouraged (53 steps).
 Proof modification of "elopOLD" is discouraged (35 steps).
 Proof modification of "elpr2OLD" is discouraged (59 steps).
 Proof modification of "elpwgded" is discouraged (23 steps).
@@ -19221,7 +19221,9 @@ Proof modification of "frege96" is discouraged (39 steps).
 Proof modification of "frege97" is discouraged (106 steps).
 Proof modification of "frege98" is discouraged (116 steps).
 Proof modification of "fsumcom2OLD" is discouraged (1241 steps).
+Proof modification of "fsummsnunzOLD" is discouraged (91 steps).
 Proof modification of "fsumshftdOLD" is discouraged (217 steps).
+Proof modification of "fsumsplitsnunOLD" is discouraged (239 steps).
 Proof modification of "fsuppmapnn0fiubOLD" is discouraged (421 steps).
 Proof modification of "funcnvmptOLD" is discouraged (291 steps).
 Proof modification of "funcnvqpOLD" is discouraged (249 steps).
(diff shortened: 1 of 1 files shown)


benjub:
@avekens : I would move eleq1w and eleq2w both before the first theorem of the form eleq*. That is, in the sequence:
...
eleq1w
eleq2w
eleq1d
...
As for the comment, keep the first sentence only for the theorem in Main, and remove the discouragement tags. That is, remove "Remark..." which I'll recover and put somewhere else in my mathbox as a TODO.

**Edit:** You have two "TODO-AV" mentioning eleq2w, which you can also carry out.

set.mm
```diff
@@ -94,6 +94,13 @@ proposed  syl       imtri       (analogous to *bitr*, sstri, etc.)
 
 DONE:
 Date      Old       New         Notes
+20-Dec-21 bj-eleq1w eleq1w      moved from BJ's mathbox to main set.mm
+20-Dec-21 rabeqi    [same]      moved from GS's mathbox to main set.mm
+20-Dec-21 rabeqif   [same]      moved from GS's mathbox to main set.mm
```
benjub, line 99: why not "rabeqfi" ?

avekens:
> @avekens : I would move eleq1w and eleq2w both before the first theorem of the form eleq*. That is, in the sequence: ... eleq1w eleq2w eleq1d ... As for the comment, keep the first sentence only for the theorem in Main, and remove the discouragement tags. That is, remove "Remark..." which I'll recover and put somewhere else in my mathbox as a TODO.
> 
> **Edit:** You have two "TODO-AV" mentioning eleq2w, which you can also carry out.

I'll have a look at it tomorrow - many thanks for your remarks so far.

set.mm
```diff
@@ -51613,6 +51695,37 @@ we define it to be empty in this case (see ~ opprc1 , ~ opprc2 , and
     ( vx cvv cpw cv wcel wss ssv selpw mpbir vex 2th eqriv ) ABCZBADZMEZNBEONBF
     NGABHIAJKL $.
 
+  ${
+    $d A v $.  $d B v $.  $d C v $.  $d V v $.
+    $( For an element of a proper unordered pair of elements of a class ` V ` ,
+       there is another (different) element of class ` V ` which is an element
```
benjub, line 51701: "of `V`" or "of the class `V`", but not "of class `V`"

set.mm
```diff
@@ -153948,6 +154055,29 @@ are used instead of sets because their representation is shorter (and more
       DXRYEUVEUVFXPXSXQBCDAUVJUVGYEUVK $.
   $}
 
+  $( The number of elements of a finite function restricted to a subset of its
+     domain.  (Contributed by AV, 15-Dec-2021.) $)
```
benjub: The number of elements of a finite function restricted to a subset of its domain is equal to the number of elements of that subset.

set.mm
```diff
@@ -153948,6 +154055,29 @@ are used instead of sets because their representation is shorter (and more
...
+  hashres $p |- ( ( Fun A /\ A e. Fin /\ B C_ dom A )
+                  -> ( # ` ( A |` B ) ) = ( # ` B ) ) $=
+    ( wfun cfn wcel cdm wss w3a cres chash cfv wceq funres 3ad2ant1 wb 3ad2ant2
+    finresfin hashfun syl mpbid ssdmres biimpi 3ad2ant3 fveq2d eqtrd ) ACZADEZB
+    AFGZHZABIZJKZUJFZJKZBJKUIUJCZUKUMLZUFUGUNUHBAMNUIUJDEZUNUOOUGUFUPUHBAQPUJRS
+    TUIULBJUHUFULBLZUGUHUQBAUAUBUCUDUE $.
+
+  $( The number of elements of a finite function expressed by a restriction.
+     (Contributed by AV, 15-Dec-2021.) $)
+  hashreshashfun $p |- ( ( Fun A /\ A e. Fin /\ B C_ dom A )
```
benjub, line 154069: The number of elements of a finite function is equal to the sum of the number of elements with argument (first component of the ordered pair) in some subset of its domain and of the number of elements with argument not in that subset.

set.mm
```diff
@@ -171874,19 +172004,53 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) ) ~~> x ) \/
   $}
 
   ${
-    $d A k x $.  $d B x $.  $d k x z $.
+    $d A k x $.  $d B x $.  $d Z k x $.
     $( A finite sum with an additional integer summand is an integer.
```
benjub: What about "A finite sum all of whose summands is an integer is itself an integer (case where the summation set is the union of a finite set and a singleton)." ?

set.mm
```diff
@@ -171874,19 +172004,53 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) ) ~~> x ) \/
...
+       17-Dec-2021.) $)
+    fsummsnunz $p |- ( ( A e. Fin /\ A. k e. ( A u. { Z } ) B e. ZZ )
+                       -> sum_ k e. ( A u. { Z } ) B e. ZZ ) $=
+      ( vx cfn wcel cz csn cun wral wa csu cv nfcv nfcsb1v csbeq1a cbvsumi snfi
+      csb a1i unfi syldan wi rspcsbela expcom adantl imp fsumzcl syl5eqel ) AFG
+      ZBHGCADIZJZKZLZUMBCMUMCENZBTZEMHUMBUQCEEBOCUPBPCUPBQRUOUMUQEUKUNULFGZUMFG
+      URUODSUAAULUBUCUOUPUMGZUQHGZUNUSUTUDUKUSUNUTCUPUMBHUEUFUGUHUIUJ $.
+
+    $d V x $.
+    $( Separate out a term in a finite sum by splitting the sum into two parts.
```
benjub, line 172019: Single out...

benjub (approved):
I made a few comments about wording.

set.mm
avekens: I also worried about that, so I took what Glauco used. I think there is no rule in which order to append the suffixes "i" and "f". Using "if" could cause confusion with the conditional operator or "dif", but "fi" is also used as suffix for "finite". Since we have both rabeqf and rabeqi, we can apply the rule in both ways. I found no other example for such a combination...

set.mm
avekens: OK, I'll fix that.

set.mm
avekens: OK, I will change that.

set.mm
avekens: I think the proposed one-to-one translation of the formula is correct, but hard to understand. I remember that in several similar cases provided by me Norm asked me to revise such comments and not to try to exactly describe the formula. Therefore, I would not chnage the comment. Are there other opinions?

set.mm
avekens: "Separate out" is often used (by many contributors) and in similar cases, whereas "single out" is only used once (in a different case) in set.mm. Therefore, I would not change this.

set.mm
avekens: OK, I will adapt the comment correspondingly.

set.mm
tirix: I agree with Alexander (and Norm), especially here where the statement of the theorem is particularly short and easy to understand by itself already.

avekens:
I revised my contribution, and I think only the question whether to use "rabeqfi" or "rabeqif" is open. Any opinions from others?

tirix:
Indeed there are very few such examples ending in "~if" (`sbrbif` (main), `elunif` (GS), `hashunif` (TA)), but there are quite a few "~gf" and "~df", and it seems to me they are more than "~fg" and "~fd".

So both orders are used, but it would seem the "f" is more often placed after the "i", "d" or "g".
I'd suggest to make that a rule, even if in the case of "if" it collides with the "if".

tirix (approved):

benjub:
Just putting my comments here before merging (@avekens you can deal with them later) since I'm not sure I did put them in the right place:

* As you wrote in the first paragraph about Goldbach, stick with the terminology you chose, so avoid "weak ternary" and "strong ternary".
fromGoldbach --> from Goldbach

* I would remove "providing the still missing official acceptance (usually obtained by a publication in a peer-reviewed journal)" since Helfgott's paper has been accepted, it seems (https://webusers.imj-prg.fr/~harald.helfgott/anglais/book.html) and we don't want to put set.mm into this kind of a debate.

* You may add in the bibliography the link for [OeSilva] https://www.ams.org/journals/mcom/2014-83-288/S0025-5718-2013-02787-1/home.html
p. ? --> p. 2057 (Theorem 2.1)
Comment: the added value of Metamath to mathematics would be mainly in the other part of the proof, that is, a result of the form "If n>= 10^20, then..." As you note, the verification for small numbers is interesting for the engineering part, in that it may force us to develop new tools.

set.mm
avekens: As @tirix wrote in his comment, the "f" is more often placed after the "i", "d" or "g" and therefore shoud become a rule/convention. I will add it to our conventions in set.mm in my next PR.

In this special case, rabeqif will not be changed.

avekens:
@benjub I have got your comments (twice ;-) ), and I have put them on my Todo list.
Thank you for merging my PR.

avekens:
@benjub @tirix I think you merged my PRs in different ways: PR #2362 was merged by @tirix as it was done by Norm before, but this PR was merged by @benjub differently (I cannot see that it is merged in GitHub Deskop, for example). Furthermore, I wanted to create a new PR now, but Github still showed the differences of this PR #2383 (in addition to the new changes). It seems to me as if the merging was not complete somehow. Before I create a new PR, could you have a look at it, please?

benjub:
I did a "squash and merge", but I think this shouldn't matter. I do not see a difference between both merges, but I'm away from my main computer for a few days and I haven't used GitHub desktop.  Can you do a kind of "git pull upstream develop" ? (be careful not to lose you work: save it somewhere outside the git directories).

PR #2386, opened 2021-12-21
Add maximum of two real numbers to iset.mm

### LINKED ISSUE #2115: Add maximum of two real numbers to iset.mm

jkingdon:
Based on the discussion at #2115 I think we have a moderate amount of understanding of notation and what we want to prove here.

This includes the theorems from Geuvers about the maximum of two real numbers and I believe it includes all the results discussed in the comments here and in #2115 .

Fixes #2115 .

~~I'm opening this as a draft pull request even though I am just starting to prove/state things, on the theory that it is easier for others to follow along than if it were merely a branch without a draft pull request. Some of the results here will probably be dropped or restated before this is ready to consider merging.~~ Now ready to review


iset.mm
```diff
@@ -99758,6 +99758,64 @@ A Cauchy sequence (as defined here, which has a rate of convergence
...
+     Lemma 3.10 of [Geuvers], p. 10.
+     (Contributed by Jim Kingdon, 21-Dec-2021.) $)
+  maxle1 $p |- ( ( A e. RR /\ B e. RR ) -> A <_ sup ( { A , B } , RR , < ) ) $=
+    ? $.
+
+  $( The maximum of two reals is no smaller than the second real.
+     Lemma 3.10 of [Geuvers], p. 10.
+     (Contributed by Jim Kingdon, 21-Dec-2021.) $)
+  maxle2 $p |- ( ( A e. RR /\ B e. RR ) -> B <_ sup ( { A , B } , RR , < ) ) $=
+    ? $.
```
benjub: maxle2 is a consequence of maxle1 and maxcom (probably shorter proof), so maybe put maxcom at the very top.

iset.mm
```diff
@@ -99758,6 +99758,64 @@ A Cauchy sequence (as defined here, which has a rate of convergence
...
+  maxcom $p |- ( ( A e. RR /\ B e. RR ) ->
+      sup ( { A , B } , RR , < ) = sup ( { B , A } , RR , < ) ) $=
+    ? $.
+
+  $( Equivalence of ` <_ ` and being equal to the maximum of two reals.
+     Lemma 3.12 of [Geuvers], p. 10.
+     (Contributed by Jim Kingdon, 21-Dec-2021.) $)
+  maxleb $p |- ( ( A e. RR /\ B e. RR ) ->
+      ( A <_ B <-> sup ( { A , B } , RR , < ) = B ) ) $=
+    ? $.
```
benjub: Wouldn't you need first that sup is the *least* upper bound ? i.e. (from #2115)
|- ( ( ( A e. RR /\ B e. RR /\ C e. RR ) /\ ( A <_ C /\ B <_ C ) ) -> sup ( { A , B } , RR , < ) <_ C )

iset.mm
```diff
@@ -99758,6 +99758,64 @@ A Cauchy sequence (as defined here, which has a rate of convergence
     LXCSVPWEXESVGURVHVPVRWAVPVQCDZVRNDABVIZVQPQVPVSVTWTXAUMVPXJTVRIJXKVQVJQVPVS
     VTWTXAVPVNTVSIJWOAVJQVPVOTVTIJWPBVJQVKVLVM $.
 
+  $( An upper bound for ` { A , B } ` .
+     (Contributed by Jim Kingdon, 20-Dec-2021.) $)
+  maxabsle $p |- ( ( A e. RR /\ B e. RR ) ->
+      A <_ ( ( ( A + B ) + ( abs ` ( A - B ) ) ) / 2 ) ) $=
+    ? $.
```
benjub: I don't understand the usefulness of that result ?

iset.mm
```diff
@@ -99758,6 +99758,64 @@ A Cauchy sequence (as defined here, which has a rate of convergence
...
+    maxabslemlub.b $e |- ( ph -> B e. RR ) $.
+    maxabslemlub.c $e |- ( ph -> C e. RR ) $.
+    maxabslemlub.clt $e |- ( ph ->
+      C < ( ( ( A + B ) + ( abs ` ( A - B ) ) ) / 2 ) ) $.
+
+    $( Lemma for ~ maxabs . A least upper bound for ` { A , B } ` .
+       (Contributed by Jim Kingdon, 20-Dec-2021.) $)
+    maxabslemlub $p |- ( ph -> ( C < A \/ C < B ) ) $=
+      ? $.
+  $}
```
benjub: I don't understand the need for this lemma ? Once you have dfabsmax, you deduce
a+b+|a-b| = a+b+max(a-b, b-a) = max(2a,2b) = 2max(a,b)
I confess that this looks a bit tedious to formalize, since the proper way is probably to state it as the supremum of a translated subset in an ordered group --- or at least (RR,+,<) --- and then of a positive dilation of a subset of RR...

iset.mm
```diff
@@ -99758,6 +99758,64 @@ A Cauchy sequence (as defined here, which has a rate of convergence
...
+     (Contributed by Jim Kingdon, 21-Dec-2021.) $)
+  maxleb $p |- ( ( A e. RR /\ B e. RR ) ->
+      ( A <_ B <-> sup ( { A , B } , RR , < ) = B ) ) $=
+    ? $.
+
+  $( Absolute value defined in terms of maximum.
+     Definition 3.13 of [Geuvers], p. 11.
+     (Contributed by Jim Kingdon, 21-Dec-2021.) $)
+  dfabsmax $p |- ( A e. RR -> ( abs ` A ) = sup ( { A , -u A } , RR , < ) ) $=
+    ? $.
```
benjub: The inequality >= is an easy consequence of leabs and absneg (by the way, -A <= |A| may be worth stating). The inequality <= is a consequence of 0 <= max(A, -A). For the latter, do you have in mind specific results already formalized for sup and/or Dedekind cuts that migh apply ? **Edit:** or maybe shorter to use ~absle.

benjub:
I got interested so I made a few comments. Sorry if they were obvious to you (ideas are cheap, execution is hard)...

iset.mm
jkingdon: In order to apply theorems like http://us.metamath.org/ileuni/eqsupti.html or http://us.metamath.org/ileuni/supclti.html we're going to use rspcev to get the ∃𝑥 ∈ 𝐴 part with `` x = ( ( ( A + B ) + ( abs ` ( A - B ) ) ) / 2 ) ``.

This theorem (applied twice, once for A and once for B) would turn into ∀𝑦 ∈ 𝐵 ¬ 𝑥𝑅𝑦 (the upper bound part).

maxabslemlub would turn into ∀𝑦 ∈ 𝐴 (𝑦𝑅𝑥 → ∃𝑧 ∈ 𝐵 𝑦𝑅𝑧) (the least upper bound part).

jkingdon:
> I got interested so I made a few comments. Sorry if they were obvious to you (ideas are cheap, execution is hard)...

Most of this is subject to change as execution gets going a bit more.

But even "obvious" stuff like http://us.metamath.org/ileuni/lenlt.html may not be available until we prove ` sup ( { A , B } , RR , < ) e. RR ` and by the time we get that, we're mostly done (assuming it comes from http://us.metamath.org/ileuni/supclti.html and I don't know where else it would come from).

iset.mm
jkingdon: Probably. At least, that's how Geuvers proves it.

benjub:
This made me look at df-sup and I have suggestions/questions regarding modifications, also in set.mm, and I'd like to have your opinions (including @digama0's).  The current form is
> df-sup |- sup ( A , B , R ) = U. { x e. B | ( A. y e. A -. x R y /\ A. y e. B ( y R x -> E. z e. A y R z ) ) }

1. This expression has meaning only when A C_ B (and R C_ ( B X. B ) ) so why not replace both occurrences of A on the RHS with ( A i^i B ) ?  This automatically removes the "garbage" that is outside B, so that we have sup ( A , B , R ) = sup ( ( A i^i B ) , B , R ) and there is no "unwanted behavior".

2. It would probably be clearer to use riota, that is, sup ( A , B , R ) = ( iota_ x e. B ...... ) , which automatically evaluates to (/) when there is no unique element having that property.

3. The formula -. x R y on the RHS looks a bit strange. We want an upper bound, so why not write instead ( y = x \\/ y R x ) ? This is classically equivalent in the cases of interest (orders) and it seems to be closer to what we really want. Intuitionistically, that may have better properties (if I recall correctly, the negation of a strict order may be a non-strict order with some additional property).

4. Would it be useful to define functions -infty_ and +infty_ so that the sup may be assigned to ( -infty_ \` B ) when ( A i^i B ) = (/) and to ( +infty_ \` B ) when ( A i^i B ) is not R-upper-bounded in B ?

jkingdon:
>     1. This expression has meaning only when A C_ B (and R C_ ( B X. B ) ) so why not replace both occurrences of A on the RHS with ( A i^i B ) ? 

I'm having a bit of a negative reaction to this but I'm not sure why. It complicates a definition which is already kind of hard to work with. But perhaps more importantly, I think it might be kind of confusing in the case where the intersection is doing something. You'd appear to be taking the supremum of one set but you'd really be taking the supremum of another, unstated, set.
> 
>     2. It would probably be clearer to use riota, that is, sup ( A , B , R ) = ( iota_ x e. B ...... ) , which automatically evaluates to (/) when there is no unique element having that property.

I'm not sure I immediately grasp all the consequences (for example would more theorems require a hypothesis that the supremum exists?) but I think this might be an improvement, or at least a little clearer.
> 
>     3. The formula -. x R y on the RHS looks a bit strange. We want an upper bound, so why not write instead ( y = x \/ y R x ) ? This is classically equivalent in the cases of interest (orders) and it seems to be closer to what we really want. Intuitionistically, that may have better properties (if I recall correctly, the negation of a strict order may be a non-strict order with some additional property).

No, I'm pretty sure the disjunction doesn't behave well without excluded middle. See for example https://ssl.metamath.org/ileuni/lenlt.html . I mean you can try your proposed definition and see how much you are able to prove and I suspect you'll see why all the constructive literature I was able to find uses lenlt or the like.
> 
>     4. Would it be useful to define functions -infty_ and +infty_ 

You mean like https://ssl.metamath.org/mpeuni/supxr.html and https://ssl.metamath.org/mpeuni/supxrunb1.html ?

digama0:
> 2. It would probably be clearer to use riota, that is, sup ( A , B , R ) = ( iota_ x e. B ...... ) , which automatically evaluates to (/) when there is no unique element having that property.

A long time ago I went through and used `iota` or `iota_` everywhere I could. df-sup was a deliberate omission, because without it one of the "cute" theorems about sup (~dfsup2) was not provable.

> 3. The formula -. x R y on the RHS looks a bit strange. We want an upper bound, so why not write instead ( y = x \\/ y R x ) ? This is classically equivalent in the cases of interest (orders) and it seems to be closer to what we really want. Intuitionistically, that may have better properties (if I recall correctly, the negation of a strict order may be a non-strict order with some additional property).

This will not fare any better. If you need le in this definition you should take it as an argument; and perhaps le alone is sufficient.

benjub:
@digama0 : Do you have opinions on point 1 ? (@jkingdon : the goal is to remove confusion: if A happens to have elements outside B, these elements should not play any role: it is always the supremum of a subset of B, that you want to take.)

Regarding point 2, if I'm correct: when there is no sup, both definitions give (/), and when there is more than one sup (possible only when R is not an order), df-sup will give their union while iota_ will still give (/).

As for point 3, I'll have to think more about it; I was used to using only non-strict orders, and in intuitionistic logic, both points of view are no longer equivalent.

@jkingdon : yes, the addition of a least and a greatest element when there is none for that order generalizes the construction of extended reals. By the way, the functions +-infty should also take the order as an argument to return either the least/greatest element if it exists, and otherwise a new element, and it should also return the new extended order... So it should be a functor from the category of ordered sets to itself... Long-term project !

iset.mm
jkingdon: > The inequality >= is an easy consequence of leabs and absneg . . .
> The inequality <= . . . [follows from] ~absle.

This formalizes nicely. I added you to the Contributed line because the above is a sketch of the proof I ended up formalizing.

benjub:
@jkingdon Nice ! I'm not sure I deserve the credit for saying "this is easy to do" and letting you do the hard work :-| 
I was thinking of proving the "hard" theorem ~maxabs from ~dfabsmax by "elementwise operations". Namely:
> a+b+|a-b| = a+b+max(a-b, b-a) = max(2a,2b) = 2max(a,b) 

where the first equality is given by ~dfabsmax.  From that, you obtain ~maxabs.  This would require other changes since currently, ~dfabsmax relies indirectly on ~maxabs.  Also, the required lemmas, which should be formulated in terms of the supremum of a translated subset in an ordered group --- or at least (RR,+,<) --- and then of a positive dilation of a subset of RR, would probably be tedious to formalize. What do you think ? Maybe I'll try if I find some time.

benjub (approved):

jkingdon:
> Nice ! I'm not sure I deserve the credit for saying "this is easy to do" and letting you do the hard work :-|

Apportioning credit is always hard but you suggested both defining the maximum as ` (x + y + abs(x - y)) / 2 ` and also using ` sup ( { A , B } , RR , < ) ` , neither of which I thought of myself. As for `dfabsmax` it is true that what you said was more "try using these three theorems, and proving <_ and >_ separately" than a full informal proof, but your proof sketch did turn out to be the basis for my proof, so I don't think shared credit is pushing it much (especially since I had no ideas how to approach this one until I read your remarks).

> Also, the required lemmas, which should be formulated in terms of the supremum of a translated subset in an ordered group --- or at least (RR,+,<) --- and then of a positive dilation of a subset of RR, would probably be tedious to formalize. What do you think ? Maybe I'll try if I find some time.

I'm not sure I know enough to have an opinion. While I try to be curious and learn new things, my formal math education doesn't go beyond the undergraduate level and I tend to get lost easily (in this case, on "dilation" and perhaps a few other concepts). I don't know if you have heard this story before, but the whole reason I started on iset.mm was reading the HoTT book, sorta getting it a bit but not that fully, and thinking "well I guess I better learn coq or isabelle or something to be able to try this out hands-on" and in that exploration I discovered iset.mm and well the rest is history. Some day I may make it back to type theory but constructive set theory is pretty rich and certainly fits into my brain more easily.

I guess the part I never understood about the approach you were trying to take is how you were going to prove that the supremum exists. In set.mm you can often just show that the set is nonempty and bounded above. In iset.mm you can't do that, and probably need an expression for what the supremum is, as in `maxabslemval` (from this pull request) or `zsupcllemex` (which perhaps could better be described as an algorithm than an expression, but it does step-by-step show how to build the supremum as part of that proof by induction).

Having said all that, I'd encourage you to give it a try. You'll learn a lot more by writing proofs than by attempting to figure it out via mailing list or github issue discussions.

benjub:
Maybe "dilation" is a bad term, but I was thinking for instance of the set of reals 2A = {2x | x \in A} if A is a set of reals.
I withdraw my claim about "ordered group" and stay within RR. Dedekind reals have the lub property, don't they ? And for Cauchy reals, which do not have the lub property, it is still the case that translates and nonnegative "dilates" of sets of reals with suprema have suprema, and these are the ones we expect (in set.mm, http://us.metamath.org/mpeuni/supadd.html). This is because (RR, <) is homogeneous under translations. Of course, it's RR_ddk (resp. RR_cauchy) under translations and dilations by Dedekind (resp. Cauchy) reals. At least, I hope so.

jkingdon:
> Dedekind reals have the lub property, don't they ?

Not in IZF. See a modified lub property for example at https://plato.stanford.edu/entries/mathematics-constructive/ although I don't have all the right references handy.

benjub:
> > Dedekind reals have the lub property, don't they ?
> 
> Not in IZF. See a modified lub property for example at https://plato.stanford.edu/entries/mathematics-constructive/ although I don't have all the right references handy.

Thanks. At least, the constructive lub property stated there is invariant under translation and positive dilation/scaling/multiple. That property holds for Bishop's construction, which uses Cauchy sequences with preassigned convergence rate, so there may be other subtleties lurking... !

PR #2387, opened 2021-12-21
Make the check-braces script skip comments entirely

jamesjer:
The current version of the script complains about curly braces in comments when the comment markers are not on the same line.


tirix (approved):

tirix:
Looks good to me! Thank you Jerry!

PR #2389, opened 2021-12-22
Add list of other databases, fixes #2384

### LINKED ISSUE #2384: List of Metamath databases outside of the set.mm repository
In a remark in PR #2373 , I suggested we'd list Metamath databases which are not in the set.mm repository (this repository).
Here are a few, in no particular order:

* @digama0 's [Dependent-Type Theory](https://github.com/digama0/dtt.mm/blob/master/dtt.mm) 
* ~@sctfn 's [New Foundations](https://github.com/sctfn/metamath-nf/blob/master/nf.mm)~ - (Actually moved here)
* @tirix 's [Q0 Logic](https://github.com/tirix/q0.mm)
* @pbrosnan [Number Theory Game](https://github.com/pbrosnan/ntg) Based on peano.mm, with additions.
* @geohot 's [TwitchCoq](https://github.com/geohot/twitchcoq/blob/master/metamath/realtwoplustwo.mm) has a few interesting 2+2 proofs
* @ishanpm 's [Linear Logic](https://github.com/ishanpm/metamathlinear/blob/master/linear.mm)
* @jamesjer 's [Unlimited Register Machine and Hilbert Geometry](https://github.com/jamesjer/set.mm/blob/geometry/set.mm) is a fork of set.mm with very interesting work which has not been merged back, IIRC
* @sorear 's [A variation on Peano](https://github.com/sorear/metamath-stuff/blob/master/yapeano.mm) where expressions variables cannot fail to refer
* @rod-lin 's [Matching Logic](https://github.com/rod-lin/kore2mm/blob/master/prelude.mm) - the result of a Kore-to-MM conversion.

There are also several [test databases](https://github.com/david-a-wheeler/metamath-test) collected by @david-a-wheeler .

david-a-wheeler:
@tirix suggested adding a list of other databases in #2384.
This implements that suggestion, using @tirix's list.

I put it in a separate file, not in the README, so that readers
won't accidentally think these are *in* this repository.

### DIFF
--- a/OTHER-DATABASES.md
+++ b/OTHER-DATABASES.md
@@ -1,0 +1,19 @@
+# Other databases
+
+Here is a list of some other Metamath databases that we know of
+but are not included in this repository:
+
+
+* @digama0 's [Dependent-Type Theory](https://github.com/digama0/dtt.mm/blob/master/dtt.mm) 
+* ~@sctfn 's older version of [New Foundations](https://github.com/sctfn/metamath-nf/blob/master/nf.mm) (this has since been moved into this repository)
+* @tirix 's [Q0 Logic](https://github.com/tirix/q0.mm)
+* @pbrosnan [Number Theory Game](https://github.com/pbrosnan/ntg) based on peano.mm, with additions.
+* @geohot 's [TwitchCoq](https://github.com/geohot/twitchcoq/blob/master/metamath/realtwoplustwo.mm) has a few interesting 2+2 proofs
+* @ishanpm 's [Linear Logic](https://github.com/ishanpm/metamathlinear/blob/master/linear.mm)
+* @jamesjer 's [Unlimited Register Machine and Hilbert Geometry](https://github.com/jamesjer/set.mm/blob/geometry/set.mm) is a fork of set.mm with very interesting work which has not been merged back
+* @sorear 's [A variation on Peano](https://github.com/sorear/metamath-stuff/blob/master/yapeano.mm) where expressions variables cannot fail to refer
+* @rod-lin 's [Matching Logic](https://github.com/rod-lin/kore2mm/blob/master/prelude.mm) - the result of a Kore-to-MM conversion.
+* @david-a-wheeler 's collection of [test databases](https://github.com/david-a-wheeler/metamath-test).
+
+*Anyone* can create a Metamath database, but we thought linking to
+some of them might be of use.
--- a/README.md
+++ b/README.md
@@ -76,6 +76,9 @@ in (approximate) decreasing size, are:
 * "[demo0.mm](./demo0.mm)" - a simple formal system used as a demonstration in
   Chapter 2 of the Metamath book.
 
+Feel free to also look at this
+[list of other Metamath databases](./OTHER-DATABASES.md).
+
 ## How are the databases verified?
 
 We work to provide *extremely* high confidence that the


OTHER-DATABASES.md
```diff
@@ -0,0 +1,19 @@
+# Other databases
+
+Here is a list of some other Metamath databases that we know of
+but are not included in this repository:
```
benjub, line 4: Something like
> ...Metamath databases available on GitHub that we know of

?

OTHER-DATABASES.md
```diff
@@ -0,0 +1,19 @@
...
+
+
+* @digama0 's [Dependent-Type Theory](https://github.com/digama0/dtt.mm/blob/master/dtt.mm) 
+* ~@sctfn 's older version of [New Foundations](https://github.com/sctfn/metamath-nf/blob/master/nf.mm) (this has since been moved into this repository)
+* @tirix 's [Q0 Logic](https://github.com/tirix/q0.mm)
+* @pbrosnan [Number Theory Game](https://github.com/pbrosnan/ntg) based on peano.mm, with additions.
+* @geohot 's [TwitchCoq](https://github.com/geohot/twitchcoq/blob/master/metamath/realtwoplustwo.mm) has a few interesting 2+2 proofs
+* @ishanpm 's [Linear Logic](https://github.com/ishanpm/metamathlinear/blob/master/linear.mm)
+* @jamesjer 's [Unlimited Register Machine and Hilbert Geometry](https://github.com/jamesjer/set.mm/blob/geometry/set.mm) is a fork of set.mm with very interesting work which has not been merged back
+* @sorear 's [A variation on Peano](https://github.com/sorear/metamath-stuff/blob/master/yapeano.mm) where expressions variables cannot fail to refer
```
benjub, line 14: "expression variables" (no 's' at 'expression') ? By the way, what do you mean ?  If you mean the same thing as "fail to denote", then this is also the case in set.mm without class variables, since the logic used is classical (as opposed to free).

digama0:
I don't think we need to SHOUT_CASE the filename. README.md is traditional, but for most md files they should just get regular names.

OTHER-DATABASES.md
```diff
@@ -0,0 +1,19 @@
+# Other databases
+
+Here is a list of some other Metamath databases that we know of
+but are not included in this repository:
+
+
+* @digama0 's [Dependent-Type Theory](https://github.com/digama0/dtt.mm/blob/master/dtt.mm) 
+* ~@sctfn 's older version of [New Foundations](https://github.com/sctfn/metamath-nf/blob/master/nf.mm) (this has since been moved into this repository)
+* @tirix 's [Q0 Logic](https://github.com/tirix/q0.mm)
+* @pbrosnan [Number Theory Game](https://github.com/pbrosnan/ntg) based on peano.mm, with additions.
```
benjub, line 10: I know he was not a good James Bond, but doesn't he deserve an apostrophe nonetheless ?

OTHER-DATABASES.md
tirix: To be honest I did not invest the time to understand Stefan's work. This seems to be an experiment, based on set.mm's predicate calculus, with this special axiom which caught my eye:
```
  ${
    $d x A $.
    $( This expresses the fact that in yapeano.mm, unlike in set.mm, expression
       variables cannot fail to refer. $)
    ax-ex $a |- E! x x = A $.
  $}
```
I did not give it further thoughts.

There is also a definition of the "unique individual such that" (`iota`) operator. It does not seem to be very much developed beyond definition and a few basic properties.

We shall maybe rather skip this entry altogether?

digama0: What it means is that there is a sort of expressions for natural numbers, unlike set.mm which has a sort for classes but no sort for set expressions. [peano.mm0](https://github.com/digama0/mm0/blob/master/examples/peano.mm0#L207) also has a sort for natural number expressions, because this is the most natural sort to give term constructors like `a + b` in peano arithmetic.

digama0: Looking at the database though, it doesn't seem to actually have any peano arithmetic in it. Most of it is just FOL which looks copy-pasted from set.mm, except for the last section that defines a sort `exp` and an iota operator but doesn't leave the realm of pure FOL - there are no peano axioms, 0, + and so on. I don't think it should be omitted from the list though - it certainly meets the qualifications, since it is a metamath database which is not in this repo.

benjub: Ok, so it's a kind of FOL with terms. Do you have a proposal for what to write ? E.g.
> /yapeano.mm) whose goal is to treat Peano arithmetic as a first-order theory with terms.

?

jkingdon (approved):

jkingdon:
There are some good ideas in comments on this pull request, but I guess our usual approach is that there's no reason why they can't be addressed in follow up pull requests, so I'll go ahead and merge this because it seems like a good start.

PR #2393, opened 2021-12-22
Fix Benoît's remarks, generalize ~ elfznelfzo and ~ elfznelfzob

tirix:
This fixes Benoît's remarks from #2379 , and revises ~ elfznelfzo and ~ elfznelfzob to make them more general.
Indeed, the original statement were for a set variable ` y `. The revised version uses a class variable ` M ` instead.
This also somehow shortened the proofs.


benjub (approved):

avekens (approved):
Great! ~ elfznelfzo and ~ elfznelfzob where some of my early proven theorems, and of course it is better to use class variables instead of set variables.

PR #2394, opened 2021-12-22
mathbox: elementwise operations (only the definition)

benjub:
mathbox only, except for rewrapping (and five final periods in @tirix's comments).

This is only the definition of "elementwise operations", as proposed in #1682, which could be useful to define some sets of integers, like odd natural numbers for instance.  But this could also be used in any structure, for instance to state that the sum of two convex sets is convex in a real vector space, etc. Might be of interest to @tirix and @avekens in particular.


benjub:
Also, I fixed
> ?Warning: The label token "tgoldbachgtd" (referenced in comment of statement
"hgt750lem") is not a $a or $p statement label.

and again, I'm not sure why it was not caught in previous PRs.

Note to @tirix : there are other missing final periods in comments of some of your commented out theorems.

set.mm
```diff
@@ -528217,7 +528221,34 @@ sethood hypotheses (compare ~ opth ).  (Contributed by BJ, 6-Oct-2018.) $)
...
+       ` A ` and ` B ` are sets of complex numbers, then
+       ` ( A ( elmtwise `` + ) B ) ` is the set of numbers of the form
+       ` ( x + y ) ` with ` x e. A ` and ` y e. B ` .  The set of odd natural
+       numbers is
+       ` ( ( { 2 } ( elmtwise `` x. ) NN0 ) ( elmtwise `` + ) { 1 } ) ` , or
+       less formally ` 2 NN0 + 1 ` .  (Contributed by BJ, 22-Dec-2021.) $)
+    df-elmtwise $a |- elmtwise = ( o e. _V |->
+       ( x e. _V , y e. _V |-> { z | E. u e. x E. v e. y z = ( u o v ) } ) ) $.
+  $}
+
```
avekens, line 528247: This looks interesting, but I see no benefit regarding odd numbers being defined directly as { z e. ZZ | ( ( z + 1 ) / 2 ) e. ZZ } or { z e. ZZ | -. 2 || z }.

avekens (approved):

set.mm
benjub: It makes the definitions more uniform, and is close to the standard symbol 2 NN + 1 (see e.g. Conjecture 1.5 of https://arxiv.org/pdf/1909.03562.pdf).

tirix:
I think defining this operation is interesting. It might be used in the following theorems or definitions of set.mm:

- [~ df-mp](http://us2.metamath.org:88/mpeuni/df-mp.html)
- [~ df-plp](http://us2.metamath.org:88/mpeuni/df-plp.html)
- [~ genpv](http://us2.metamath.org:88/mpeuni/genpv.html)
- [~ supadd](http://us2.metamath.org:88/mpeuni/supadd.html)
- [~ supmul](http://us2.metamath.org:88/mpeuni/supmul.html)
- [~ sqrlem5](http://us2.metamath.org:88/mpeuni/sqrlem5.html)
- [~ lsppr](http://us2.metamath.org:88/mpeuni/lsppr.html)
- [~ df-pstm](http://us2.metamath.org:88/mpeuni/df-pstm.html)

benjub:
Thanks @tirix.  I was thinking about future developments but I had not expected that there were already such results in set.mm.  The search
> MM> SEARCH * "E. ? e. ? E. ? e. ? ? = ("

gave a few candidates. In particular, "elmtwise" is closely related to the range. Something like
> A ( elmtwise \` F ) B = ran ( F |\` ( A X. B ) ) 

under the right sethood and functionhood hypotheses.   Also, ~df-lsm can be roughly rewritten
> LSSum = ( elmtwise o. +g )

sctfn:
It’s actually just what I need for the surreal field operations as well!

PR #2395, opened 2021-12-23
Generalize ~ injresinjlem, ~ftc2re, ~ fdvposlt and co.

tirix:
Similar to ~ elfznelfzo and ~ elfznelfzob in #2393 , this generalizes ~ injresinjlem by using a class instead of a setvar.

Also, following a very good remark from Gérard, this also makes ~ ftc2re, ~ fdvposlt and other similar theorems more general by weakening the hypothesis that the target function is continuous to just being a function. Indeed, since its derivative is continuous, it must itself be continuous.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,15 +1,66 @@
-If you are new to GitHub and git, the basic instructions for
-contributing changes to set.mm are described in the wiki,
-[Getting started with contributing](https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing).
-(Feel free to enhance the wiki.)
+# Contributing and Acceptance
 
-The rest of this note shows how to achieve the recommended
-formatting for set.mm and iset.mm prior to submitting your changes.
+We're a friendly community, and we would love to have more collaborators!
 
-### Formatting recommendation prior to submitting a pull request
+This document briefly explains
+[how to contribute proposed changes](#how-can-i-contribute)
+("contributions" or "merge requests") to this collection of databases
+We also discuss
+[Formatting recommendation prior to submitting a pull request](#formatting-recommendation-prior-to-submitting-a-pull-request), and for the rare cases
+when it's needed, information on
+[regenerating the `discouraged` file](#regenerating-the-discouraged-file).
+This page also describes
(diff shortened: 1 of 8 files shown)


set.mm
```diff
@@ -465091,7 +465091,8 @@ its graph has a given second element (that is, function value).
 
   That decimal point operator is defined in the next section.
   The bulk of these constructions have originally be proposed by David A.
-  Wheeler on 15-May-2015.
+  Wheeler on 12-May-2015, and discussed with Mario Carneiro in this thread:
+  ~ https://groups.google.com/g/metamath/c/2AW7T3d2YiQ/m/iSN7g87t3ikJ .
```
benjub, line 465095: This seems to be the URL to a specific message in the thread. Link to the thread:
https://groups.google.com/g/metamath/c/2AW7T3d2YiQ
(I removed the "/m/...").

benjub (approved):

set.mm
```diff
@@ -145465,36 +145465,36 @@ Finite intervals of nonnegative integers (or "finite sets of sequential
...
+       31-Oct-2017.)  (Proof shortened by AV, 14-Feb-2021.)  (Revised by
+       Thierry Arnoux, 23-Dec-2021.) $)
+    injresinjlem $p |- ( -. Y e. ( 1 ..^ K ) -> ( ( F ` 0 ) =/= ( F ` K )
        -> ( ( F : ( 0 ... K ) --> V /\ K e. NN0 )
          -> ( ( ( F " { 0 , K } ) i^i ( F " ( 1 ..^ K ) ) ) = (/)
-           -> ( ( x e. ( 0 ... K ) /\ y e. ( 0 ... K ) )
-             -> ( ( F ` x ) = ( F ` y ) -> x = y ) ) ) ) ) ) $=
+           -> ( ( X e. ( 0 ... K ) /\ Y e. ( 0 ... K ) )
+             -> ( ( F ` X ) = ( F ` Y ) -> X = Y ) ) ) ) ) ) $=
       ( vz wcel wn cc0 wa wceq cfv wi fveq2 eqcoms notbid 2a1d syl com12 sylbi
```
avekens, line 145477: This is only a lemma, but OK...

avekens (approved):

PR #2403, opened 2021-12-27
edit comment of df-slot; move basfn to Main; typos

benjub:
@digama0 can you check that the comment edit for df-slot is ok ?

I still think "Eval" would be a good name for Slot ! This change would become beneficial if there are uses outside of the extensible structures machinery. Can you think of any? At least, the remark is now in the comment.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16745,6 +16745,7 @@ New usage of "n0fOLD" is discouraged (0 uses).
 New usage of "n0lpligALT" is discouraged (0 uses).
 New usage of "naecoms-o" is discouraged (1 uses).
 New usage of "natded" is discouraged (0 uses).
+New usage of "ndxidOLD" is discouraged (0 uses).
 New usage of "negexsr" is discouraged (0 uses).
 New usage of "nelsnOLD" is discouraged (0 uses).
 New usage of "nexdOLD" is discouraged (0 uses).
@@ -19410,6 +19411,7 @@ Proof modification of "mulgnndirOLD" is discouraged (440 steps).
 Proof modification of "n0fOLD" is discouraged (51 steps).
 Proof modification of "n0lpligALT" is discouraged (74 steps).
 Proof modification of "naecoms-o" is discouraged (19 steps).
+Proof modification of "ndxidOLD" is discouraged (48 steps).
 Proof modification of "nelsnOLD" is discouraged (46 steps).
 Proof modification of "nexdOLD" is discouraged (9 steps).
 Proof modification of "nexdvOLD" is discouraged (18 steps).
--- a/discouraged
+++ b/discouraged
@@ -18553,6 +18553,8 @@ Proof modification of "bj-issetwt" is discouraged (63 steps).
 Proof modification of "bj-mo3OLD" is discouraged (206 steps).
 Proof modification of "bj-mpt2mptALT" is discouraged (130 steps).
 Proof modification of "bj-nalset" is discouraged (100 steps).
+Proof modification of "bj-ndxarg" is discouraged (37 steps).
+Proof modification of "bj-ndxid" is discouraged (28 steps).
 Proof modification of "bj-nfab1" is discouraged (10 steps).
 Proof modification of "bj-nfcjust" is discouraged (29 steps).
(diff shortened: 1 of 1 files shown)


avekens (approved):
Everything is fine!

set.mm
```diff
@@ -199795,31 +199797,39 @@ implemented as a function (a set of ordered pairs)".  The current
...
-       function argument can be any set, although it is meaningful only if it
-       is a member of ` Poset ` ( ~ df-poset ) when used for posets or of
-       ` Grp ` ( ~ df-grp ) when used from groups.  (Contributed by Mario
-       Carneiro, 22-Sep-2015.) $)
+    $( Define the slot extractor for extensible structures.  The class
+       ` Slot A ` is a function whose argument can be any set, although it is
+       meaningful only if that set is a member of an extensible structure (such
+       as a partially ordered set ( ~ df-poset ) or a group ( ~ df-grp )).
+
+       Note that ` Slot A ` is "evaluation at ` A ` ".  The class ` Slot `
```
digama0, line 199815:
```suggestion
       Note that ` Slot A ` is implemented as "evaluation at ` A ` ". That is,
       ` ( Slot A `` S ) ` is defined to be ` ( S `` A ) ` , where ` A ` will
       typically be a small natural number. Each extensible structure ` S ` is a function on
       defined on specific natural number "slots", and this function extracts the value at
       a particular slot.
       
       The special "structure" ` ndx ` , defined as the identity function, can be used to extract
       the number ` A ` from a slot, since ` ( Slot A `` ndx ) = A ` . This is typically used to refer
       to the number of a slot when defining structures without having to expose
       the detail of what that number is.
       
       The class ` Slot `
```

digama0:
As I mentioned before, I don't support renaming `Slot` to `Eval`. In fact, even if we found a separate use for the "evaluation at A" function, I would want a separate definition `df-eval` and either leave `Slot` as is or redefine it as `Slot A = Eval A`. df-slot is strictly an implementation detail of extensible structures, and it should be possible to swap out the encoding of extensible structures with minimal changes beyond foundational definitions like `df-slot` and `df-struct`.

benjub:
@digama0 : done.  Can you have a look at the mathbox theorems added in this PR ? (The comments and labels are in terms of "evaluation" while the math expressions are in terms of Slot, so I'll harmonize that one way or the other.) In particular, bj-sloteq/evaleq seems useful, for instance to deduce ndxid from ndxarg more naturally than with the current proof. And maybe others too ?

digama0:
I think that we should definitely have bj-sloteq, and indeed we should enforce that equality theorems are proven for every term constructor in set.mm. For auxiliary constructions that are only used in very limited ways like df-slot, sometimes we can get away without it, but it makes things harder for automatic proofs that want congruence for arbitrary terms. (This is in particular required for mmj2's definition checker; I would guess that a definition like `Foo = ( x e. _V |-> Slot x )` would be spuriously rejected as a result.)

benjub:
I moved ~bj-evaleq to Main as ~sloteq and used it to shorten ~ndxid. I deduce from your answer that for the moment, I leave ~bj-evalval and ~bj-evalid (which would be relabeled ~slotval and ~slotid) in my mathbox ? If so, you may approve/merge.

(For the record: structures were recently discussed in #2326.)

set.mm
```diff
@@ -199795,31 +199797,62 @@ implemented as a function (a set of ordered pairs)".  The current
...
-       Carneiro, 22-Sep-2015.) $)
+    $( Define the slot extractor for extensible structures.  The class
+       ` Slot A ` is a function whose argument can be any set, although it is
+       meaningful only if that set is a member of an extensible structure (such
+       as a partially ordered set ( ~ df-poset ) or a group ( ~ df-grp )).
+
+       Note that ` Slot A ` is implemented as "evaluation at ` A ` ".  That is,
+       ` ( Slot A `` S ) ` is defined to be ` ( S `` A ) ` , where ` A ` will
+       typically be a small nonzero natural number.  Each extensible structure
+       ` S ` is a function defined on specific natural number, "slots", and
```
digama0, line 199818:
```suggestion
       ` S ` is a function defined on specific natural number "slots", and
```

PR #2404, opened 2021-12-28
Toc

benjub:
Create own sections for each definition by splitting two sections, each into two sections (as indicated in commit messages).

Sometimes, definitions are naturally introduced in a group, but more often, it seems clearer to introduce one at a time.  This also ensures proximity of the $c-statement, the syntax $a-statement, the definition, and the first elementary properties (xxxeq...) for each new token.  I think something similar was also advocated by @digama0.  Since we have one more level of nesting, we also have the option of keeping the common section and subdividing it into subsections when apropriate.

This could also be done (with care because there may be cases when definitions should really be introduced as a group), with the sections "Unordered and ordered pairs", "Indexed union and intersection", "Partial and complete ordering", "Founded and well-ordering relations", "Relations", "Ordinals", "Functions", "Operations".


david-a-wheeler:
> Sometimes, definitions are naturally introduced in a group, but more often, it seems clearer to introduce one at a time. This also ensures proximity of the $c-statement, the syntax $a-statement, the definition, and the first elementary properties (xxxeq...) for each new token. I think something similar was also advocated by @digama0. Since we have one more level of nesting, we also have the option of keeping the common section and subdividing it into subsections when apropriate.

+1 for this specific change. However, I'd like @digama0 at least to also take a look at this, because while it's trivial on its own, it suggests a larger push towards a bigger structure.

Where's it's natural to introduce a definition on its own, giving a single definition this way makes sense to me. As you noted, sometimes definitions are more naturally introduced as a group, so we shouldn't be dogmatic about it.

Presuming this is merged, would you mind trying to express this as an extension of `conventions`?

digama0:
For the specific change, it looks good, but since this also entails some changes to the way things are presented I think they should all be reviewed individually, a mass update commit is not warranted.

PR #2405, opened 2021-12-28
Remove unnecessary braces from nf.mm

jamesjer:
This is the first in a series of brace removal patches.  There is little activity in nf.mm, so I thought it is probably okay to do the entire file at once.


david-a-wheeler:
It's an automated change, so I just spot-checked it, but it looks good to me.

david-a-wheeler (approved):

sctfn (approved):
Looks good on a quick read through

PR #2406, opened 2021-12-29
iset / CZF

benjub:
In preparation to the addition of a paragraph on CZF in http://us.metamath.org/ileuni/mmil.html, I restructured my iset.mm mathbox to shows the CZF part better and added a head comment. (I also sorted the mathboxes alphabetically, as in set.mm).

@digama0 , @jkingdon : can you have a look at the head comment for CZF ?


jkingdon (commented):
My only feedback is that I found the text on "metalogic" and "logic" a bit hard to follow (that is, it didn't really enlighten me on the difference between bounded separation and full separation, which I would state as something along the lines of "bounded separation restricts the formulas which can be used to separate"). But I'd go ahead and merge this now and wordsmith later (if you want to). It clearly is an improvement.

jkingdon (approved):

benjub:
@jkingdon : that's a good point. Here is a rewording, hopefully clearer.

More precisely: that paragraph was not meant to be about separation, but only about the necessity of defining "bounded formulas" within the logic, independently of that. As for bounded separation versus separation, I very briefly explained why the latter is not predicative in the head comment in my mathbox, though it is obviously not very detailed, and that's why I did put some references.

jkingdon:
> @jkingdon : that's a good point. Here is a rewording, hopefully clearer.

That is an improvement.

> the necessity of defining "bounded formulas" within the logic

Our need to define them as we do is a relatively simple issue but perhaps slightly hard to describe because many explanations will be in terms of what metamath does not have rather than what it does have. But we can definitely go with this wording unless/until we think of something clearer.

benjub:
> Our need to define them as we do is a relatively simple issue but perhaps slightly hard to describe because many explanations will be in terms of what metamath does not have rather than what it does have. But we can definitely go with this wording unless/until we think of something clearer.

I think I wrote the explanations where they belong: short introduction to CZF in the head comment of the section introducing CZF, and short explanation of the way we formalized bounded formulas in the head comment of the section introducing bounded formulas.  The paragraph we are discussing is a third head comment and is in Main (the other two are in my mathbox), and I tried to explain why *you* wrote it was not "as easy to formalize", and I provided the analogy with our treatments of proper substitution and free and bound variables.  That's not at all a criticism, that's only to explain why I wrote things that way at these places.  Maybe you had something else in mind when you wrote that comment, and this can be improved in a future PR.

Thanks for the approval.  I'd still like to have @digama0's opinion, but I'm going to merge soon regardless.

jkingdon:
> Maybe you had something else in mind when you wrote that comment, and this can be improved in a future PR.

No suggestions. Sounds like I was being too wordy and should stick with "looks good to me".

benjub:
> No suggestions. Sounds like I was being too wordy and should stick with "looks good to me".

No: it's better to be bold and explicit, and avoid misunderstandings, pease keep on !

Issue #2399, opened 2021-12-25
Prove Bézout's identity in iset.mm


jkingdon:
The set.mm proof (which at a glance would appear to be similar to the one at https://en.wikipedia.org/wiki/B%C3%A9zout's_identity#Proof ) relies on being able to take the infimum of `{𝑧 ∈ ℕ ∣ ∃𝑥 ∈ ℤ ∃𝑦 ∈ ℤ 𝑧 = ((𝐴 · 𝑥) + (𝐵 · 𝑦))}`. Our first impulse might be to formalize this using `zsupcl` (already in iset.mm), or more conveniently a new theorem which combines `zsupcl` with something more or less like `infssuzcl` from set.mm, but only if we could show that `∃𝑥 ∈ ℤ ∃𝑦 ∈ ℤ 𝑧 = ((𝐴 · 𝑥) + (𝐵 · 𝑦))` is decidable. Unless there is some shortcut I'm not seeing, to prove this would require much of the work or perhaps more (that is, decidability of this expression means giving a yes/no answer to whether a number has cofactors, whereas Bézout's identity is saying that the answer for the greatest common divisor is "yes").

The more promising approach would appear to be via https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm#Proof or the "Proof of Bézout's Identity" section of https://brilliant.org/wiki/bezouts-identity/ . Both of these do look a bit daunting in terms of how much there is to formalize, but they at least do provide a way to compute the cofactors which apparently could be turned into a proof by induction in the now-familiar style used in zsupcllemex (currently in iset.mm) and others which prove something of the form `E. ? ??`  via induction.

Any other suggestions or references to consult? I have a number theory textbook sitting at home which I haven't tried yet (but won't be back there for roughly a week, and I don't know whether it would help anyway).

digama0:
I think the extended euclidean algorithm is the way to go. The lean proof is https://github.com/leanprover-community/mathlib/blob/f80c18bfdf620b6fcf462430136c61d0ded07035/src/data/int/gcd.lean#L102 which is completely constructive. Rather than defining the `gcd_a` and `gcd_b` functions here, I would just prove that there exists a value satisfying the properties of `xgcd_aux` there.

jkingdon:
Here's my attempt to write up the proof using a combination of metamath notation and informal conventions which are hopefully clear.

[...]

Issue #2212, opened 2021-09-24
Add order types to iset.mm


jkingdon:
Specifically, as described in #2203 we want to define an equivalent of http://us.metamath.org/mpeuni/dfoi.html which says that `OrdIso` isomorphically maps a set+relation to an ordinal. R and A should satisfy http://us.metamath.org/ileuni/df-se.html and `R We A` as defined in https://github.com/metamath/set.mm/pull/2211 (the most important part of this definition being http://us.metamath.org/ileuni/df-frfor.html ).

Here's what the set.mm definition looks like (slightly trimmed down in inessential ways):

```
    dfoi.1 $e |- C = { w e. A | A. j e. ran h j R w } $.
    dfoi.2 $e |- G = ( h e. _V |-> ( iota_ v e. C A. u e. C -. u R v ) ) $.
    dfoi.3 $e |- F = recs ( G ) $.
    dfoi $p |- OrdIso ( R , A ) = ( F |`
       { x e. On | E. t e. A A. z e. ( F " x ) z R t } ) $=
```

which can be summarized as "Given a partial mapping h from ordinals to a subset of A, let C be the set of upper bounds for that subset. Then choose the least upper bound (which exists due to the set.mm definition of We) and map the next ordinal to that" (I realize "next ordinal" is perhaps not great wording because this works for limit ordinals as well as successors).

The definition we need is something like "Given a partial mapping h from ordinals to a subset of A, ...... ". Oh, fudge. I know it involves some kind of induction over some kind of property or class or something, but I'm having trouble stating it informally, much less the exact definition we'll use.

Notes:

* since we aren't using the set.mm definition anyway, I'm not sure there is much point in keeping the if expression, since `( R We A /\ R Se A )` won't be decidable.

digama0:
This actually looks rather bleak, and probably we need some more assumptions.

* Can we prove extensionality for `R`? i.e. `A. z e. A ( z R x <-> z R y )` implies `x = y`. This is obviously true for ordinals because of extensionality, and it is true for set.mm well orders because of trichotomy, but I don't see any way to derive it from `Fr` directly.
* We will probably also need a way to talk about the equivalent of `C_` for `R`, and there are two candidates - `A. z e. A ( z R x -> z R y )` and `A. z e. A ( y R z -> x R z )`; extensionality is needed to derive useful properties of this relation. This would be used instead of `-. u R v` in the iota.
[...]

Issue #2203, opened 2021-09-20
Show that finite ordinals are natural numbers in iset.mm


jkingdon:
There are a number of theorems, most of which which exist in set.mm, but which are missing from iset.mm, one or more of which is needed to get cardinality of finite sets ~~and http://us.metamath.org/mpeuni/df-hash.html unblocked (and thus allow resuming work on #2167 ):~~ _(Update Sep 2022: we defined https://us.metamath.org/ileuni/df-ihash.html another way and were able to prove most of the set size theorems from set.mm, so the topic of finite ordinals being natural numbers is still unresolved, but not blocking development of summation)._

* http://us.metamath.org/mpeuni/onomeneq.html
* http://us.metamath.org/mpeuni/onfin2.html
* http://us.metamath.org/mpeuni/ficardom.html
* http://us.metamath.org/mpeuni/cardnn.html
* `( ( A e. On /\ B e. _om /\ A ~~ B ) -> A e. _om )`

As far as I have been able to tell most/all of these are sufficient to imply the others, and in fact I've encountered similar difficulties in proving a variety of them.

The basic approaches I have tried are some variation of:

* Use http://us.metamath.org/ileuni/findcard2sd.html or one of the other findcard* theorems on the finite ordinal. Usually the difficulty here is that you can add an element to a finite ordinal (or subtract one in the case of `findcard`), but I haven't found a way to identify which element would result in that also giving us an ordinal. That sort of thing is solved for natural numbers, because we have the successor (or the predecessor) but there is nothing order-preserving about the 1-1 and onto mapping we get from `n ~~ A`. It would help if we could say a finite ordinal is either `(/)` or the successor of some ordinal, but I haven't been able to prove that.
* Use http://us.metamath.org/ileuni/finds.html on the natural number and do something along the lines of showing, one induction step at a time, that the elements of the natural number are found in the finite ordinal. This has a roughly similar problem: we can identify the next natural number to look at, but it isn't clear how to map that over to the finite ordinal and show that the finite ordinal contains each natural number, or can produce an ordinal at each step, or the like.
[...]

PR #2407, opened 2021-12-29
Add strong induction to iset.mm

jkingdon:
* the main theorem here is uzsinds , which is stated as in set.mm. The proof, instead of the set.mm
one which gets fancy with well-ordering, is the textbook one from uzind4
(the one where the textbooks say "strong induction isn't exactly
stronger, in the sense that it follows from weak induction as follows")
* also, add nnct which I noticed now works (and which happens to be next to the strong induction theorems.
* I briefly took a look at uzindi but it was a bit hard to get my head around it and I don't think we need it yet (that is, as far as I know it won't be needed for #2399 ).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -5355,7 +5355,7 @@ middle and iset.mm has few cardinality related theorems.</TD>
 <TR>
 <TD>uzindi</TD>
 <TD><I>none</I></TD>
-<TD>This could presumably be proved, perhaps from ~ uzind4 ,
+<TD>This could presumably be proved, perhaps from ~ uzsinds ,
 but is lightly used in set.mm</TD>
 </TR>
 


benjub:
> * the main theorem here is uzsinds , which is stated as in set.mm. The proof, instead of the set.mm
>       one which gets fancy with well-ordering, is the textbook one from uzind4
>       (the one where the textbooks say "strong induction isn't exactly
>       stronger, in the sense that it follows from weak induction as follows")

Does it mean that that proof should be ported to set.mm, at least as ~uzsindsALT, with a comment like "Alternate proof of ~uzsinds, bypassing general results about well-orderings" ? (I think so.)  Since the proof is genuinely different from Scott's, I think it should be "Contributed by JK" ?

sctfn:
This whole conversation has me thinking more about founded relationships in
iset. Do we want to define R Fr A using something like frmin from set.mm?
That and a proof for ordinals would probably get us away from using
ax-setind too much.

sctfn:
Sorry, frind, not frmin

jkingdon:
> Does it mean that that proof should be ported to set.mm, at least as ~uzsindsALT, with a comment like "Alternate proof of ~uzsinds, bypassing general results about well-orderings" ? (I think so.) Since the proof is genuinely different from Scott's, I think it should be "Contributed by JK" ?

All of that sounds great to me but I haven't always had the time to make it happen when I do see a result for which that would make sense.

In this case it might be as simple as just copy-pasting the proof and tweaking the name/comment. At least, I don't remember using any theorems which aren't also in set.mm.

jkingdon:
> This whole conversation has me thinking more about founded relationships in iset. Do we want to define R Fr A using something like [frind] from set.mm? That and a proof for ordinals would probably get us away from using ax-setind too much.

Isn't the current definition of `Fr` in iset.mm ( http://us.metamath.org/ileuni/df-frind.html ) something like frind?

I have no objection to reducing use of ax-setind even though my own efforts have been focused on full IZF. In the case of ordinals, our assumption has been that the definition of ordinal would need to be modified with an `_E Fr A` conjunct, and existing ordinal theorems would need to use that (possibly easier than it sounds if they already use http://us.metamath.org/ileuni/ordirr.html or the like, or could be changed to do so). "Use ordirr more" and the like would be good first steps (in the sense that the work could be done incrementally, as opposed to changing the definition which would necessitate changing a bunch of things at once, or some footwork about having two notations coexist at least for a while, or something.

sctfn:
OK, sorry, I wasn't up on the current state of iset apparently. That seems
like a reasonable project.

jkingdon:
> OK, sorry, I wasn't up on the current state of iset apparently. That seems like a reasonable project.

No worries, `Fr` was a fairly recent addition. @digama0 talked me into it. I mean, not just talked me into accepting it, that would have been easy, but talked me into doing a lot of the work, which was part of some cardinality work that is a bit stalled out at #2203 and #2212 . But the `Fr` part is working, so that's not what I mean by "stalled out".

benjub:
> > Does it mean that that proof should be ported to set.mm, at least as ~uzsindsALT, with a comment like "Alternate proof of ~uzsinds, bypassing general results about well-orderings" ? (I think so.) Since the proof is genuinely different from Scott's, I think it should be "Contributed by JK" ?
> 
> All of that sounds great to me but I haven't always had the time to make it happen when I do see a result for which that would make sense.

Then maybe you can keep a list of candidates that may be "backported" to set.mm, in order to keep both databases close when possible.

benjub (approved):

jkingdon:
> Then maybe you can

Rats! My devious plan to subtlely suggest that you do it isn't working.

PR #2392, opened 2021-12-22
remove top date and adapt preamble

benjub:
After yet another merge conflict due to the top date, I decided to go on and remove it, since, as mentioned several times, git permits good version control.  I adapted a bit the preamble. What do you think about this proposal ? (most important is to keep the preamble short, in my opinion). The part about the licence and "principal curator" will have to be changed too, but I do not know to what. You may propose the modification here so that I'll do it, or you can do a PR later.

PR #2411, opened 2021-12-30
check-braces: handle multiple comment markers on a line

jamesjer:
A recent commit put two `$)` comment end markers on a single line (currently line 12521 of set.mm and line 128 of iset.mm).  This exposed a logic error in scripts/check-braces, which started reporting no results, due to `c` having the value 1 or 2 after that line until the end of the file.  Update the script to count the number of matches per line and update the variable `c` accordingly.

### DIFF
--- a/scripts/check-braces
+++ b/scripts/check-braces
@@ -40,8 +40,8 @@
 # braces are ignored in comments.
 
 BEGIN { i = 0; c = 0 }
-/\$\(/ { ++c }
-/\$\)/ { --c }
+/\$\(/ { c += gsub(/\$\(/, "") }
+/\$\)/ { c -= gsub(/\$\)/, "") }
 /\$\{/ { if (c == 0) scope[++i] = NR }
 /\$\}/ { if (c == 0 && scope[i] != 0) print(scope[i]); delete scope[i] }
 /\$[cdefv]/ { scope[i] = 0 }


benjub:
I don't remember that line being changed "recently" ?

jamesjer:
You're right.  Those lines have not changed recently.  Hmmm.  Yet the script used to produce results and now does not, and it is definitely due to a line with two comment-ending `$)` markers.  Maybe there also used to be a line with two comment-starting '$(` markers, and that has changed recently?

In any case, the script is incorrect as is.  This commit fixes it to correctly handle multiple comment markers per line.

benjub:
It'd be interesting to know which change caused this.  Is it #2392 ?

jamesjer:
No, weirdly enough it is https://github.com/metamath/set.mm/pull/2387.  I don't understand that.  I submitted that pull request because I found the script was complaining about braces in comments.  I could swear that I tested the changes before submitting that pull request ... but git bisect doesn't lie.  That's the commit that broke the script.

jkingdon (approved):

avekens:
There is another case of two comment end markers in set.mm:

  $( $c UFD $. $)  $( Unique factorization domain. $)

Did you check if this is handled correctly? Or is there no problem because of two full comments? 
Line 12521 is:

$)        $( ... $)

benjub (approved):

jamesjer:
@avekens I apologize for taking so long to reply.  So far, 2022 has been a madhouse at work, leaving me little time for hobbies.

The new version will handle that other case correctly, since it now counts the number of `$(` and `$)` per line.  The old version would have been okay too, actually, since it would have added one and subtracted one for that line.

PR #2413, opened 2021-12-31
Add bezoutlemex to iset.mm

jkingdon:
This is Bezout's identity for nonnegative integers (with some caveats about what we mean by "greatest common divisor").

See #2399 for more discussion. Specifically this is the proof of what is called nn0bezoutlem at https://github.com/metamath/set.mm/issues/2399#issuecomment-1003898164 and covers the proof up to that point including I think all parts of https://github.com/metamath/set.mm/issues/2399#issuecomment-1001332925

Took me a long time to get myself untangled about where to put the various variables, quantifiers, and conditions, but once I got that the proof from #2399 formalized without much trouble.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -49,7 +49,7 @@
 "alrimih" is used by "nfexd".
 "alrimih" is used by "sb4or".
 "alrimih" is used by "sb6rf".
-"alrimih" is used by "sbbid".
+"alrimih" is used by "sbbidh".
 "ax-0id" is used by "addid1".
 "ax-0lt1" is used by "0lt1".
 "ax-10o" is used by "ax10".
@@ -143,6 +143,7 @@
 "reapval" is used by "reaplt".
 "reapval" is used by "reapti".
 "reapval" is used by "recexre".
+"sbbidh" is used by "sbcomxyyz".
 "sbiedh" is used by "sbcomxyyz".
 "sbiedh" is used by "sbied".
 "sbiedh" is used by "sbieh".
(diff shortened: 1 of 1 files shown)


benjub:
I think dashes in labels so far appear only in ax-xxx and df-yyy.  No strong opinion for the moment, but allowing dashes could lead to longer labels, e.g. fin-skfield-is-comm (although it looks like for the moment, you kept it to $e statements).

jkingdon:
> I think dashes in labels so far appear only in ax-xxx and df-yyy. No strong opinion for the moment, but allowing dashes could lead to longer labels, e.g. fin-skfield-is-comm (although it looks like for the moment, you kept it to $e statements).

Hmm, I'm not sure how high to climb on the "set.mm has a readability problem due to extreme abbreviation" soapbox, but in this case, I used a somewhat more verbose name pretty sparingly and to try to solve a pretty specific problem - it is hard to show (even to myself as a I write the proof) what `ph` and `ps` mean. When I'm reading https://github.com/metamath/set.mm/issues/2399#issuecomment-1001332925 I get to look at `is-bezout` and `sub-gcd` and although a single `$e` label won't give that clarity of notation, this proof is hairy enough that I would advocate for cutting ourself a bit more slack than we might in a much simpler proof.

P.S. Generally speaking I'd pick a verbose label over a verbose comment.

benjub:
I too would prefer to shift slightly from terser to more explicit labels, within reason. As for dashes, maybe `isbezout` and `subgcd` are almost as readable ?
Anyway, that's a minor point.

jkingdon:
This is now ready for review.

benjub (approved):
Good for me, although @digama0's review might be more useful since you discussed the proof together (and another opinion on the "dashes in labels" issue would be nice too).

jkingdon:
> Good for me, although @digama0's review might be more useful since you discussed the proof together (and another opinion on the "dashes in labels" issue would be nice too).

I won't delay a merge based on either of those (we can always revise things later, whether style points or other aspects of the proof). I'm following the path we laid out in #2399 pretty closely.

PR #2410, opened 2021-12-30
Added Theorems 12 and 13 of [RosserSchoenfeld] as temporary axioms.

tirix:
Except for the new bibliographic reference to [RosserSchoenfeld], this is purely in my Mathbox.

- added Theorems 12 and 13 of [RosserSchoenfeld] as temporary axioms, and derived theorems for [Helfgott],
- several proofs modified or added within comments, since intermediate steps are still missing,
- one new theorem about addition with decimal point (no carry) : `~ dpadd2`,
- a few supporting theorems, including on the first and second Chebyshev functions (e.g. `~ chtvalz`).
- I think among those, ` ~ uzssico ` is interesting and I'm surprised I did not find it in the main part of set.mm.

Issue #2381, opened 2021-12-20
Better-formatted PRs


benjub:
It has been suggested a few months ago that PRs should be more standardly formatted in order to reduce merge conflicts and diffs that one has to review.  This could be done by requiring contributors to execute the standard scripts (for set.mm and iset.mm) available at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md

The CI could execute the script itself on each PR and merge the resulting files, but in order not to hide changes, it is more transparent that the CI excecute the scripts and reject the PR if the resulting files are different from the submitted ones.

This solution was deemed not immediately necessary, thanks to Norm's diligence in examining and merging PRs.  With Norm's passing, and the queue of pending PRs likely to get longer, **I think it would be a good thing to implement this check**.

A disadvantage is that it asks a bit more from contributors.  I think copy-pasting the lines at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md and executing them in one's CLI isn't too long (but maybe some contributors have different workflows), and in any case way shorter than resolving potential merge conflicts or reviewing uselessly long PRs.

jkingdon:
> the CI excecute the scripts and reject the PR if the resulting files are different from the submitted ones.

That sounds right. I believe `scripts/rewrap` already has the commands which would be needed.

I guess the other choice is that CI could have the ability to push a follow up commit to the pull request (don't see why it would be involved in merging). I haven't seen this done much, so I'm not sure whether this is an obviously bad idea or just a "well it sounds like more work and maybe not needed".

One key advantage of doing this in CI is that it controls for the version of metamath.exe doing the rewrapping. I think it is 0.197 and 0.198 which wrap differently, although I could be misremembering the exact versions.

digama0:
> I guess the other choice is that CI could have the ability to push a follow up commit to the pull request (don't see why it would be involved in merging). I haven't seen this done much, so I'm not sure whether this is an obviously bad idea or just a "well it sounds like more work and maybe not needed".

Mathlib does this, for example with commits like [this one](https://github.com/leanprover-community/mathlib/commit/67cf4061035cbfd0940fbe87ff4a1845238bb588). This is the equivalent of set.mm's discouraged file; whenever the discouraged file changes a bot automatically commits the change. Personally, I would prefer this over other alternatives where possible, but it does require setting up a bot / commit hook.

jkingdon:
[...]

PR #2414, opened 2021-12-31
Add dfgcd3 to set.mm

jkingdon:
This definition for greater common divisor was suggested by @digama0 at https://github.com/metamath/set.mm/issues/2399#issuecomment-1001332925

It is not hugely different from dfgcd2 but has some virtues including being a real definition in the sense of being in the form `( M gcd N ) =` _something_, rather than a more complicated expressing with `( M gcd N )` in the middle.


sctfn (approved):

benjub (approved):

glacode:
After a
git pull upstream develop

and a run of the script for recommended formatting, see here:
https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md#formatting-recommendation-prior-to-submitting-a-pull-request

git diff upstream/develop
gives me some small changes (it shouldn't) referred to a couple of theorems:
Theorem 12. of [RosserSchoenfeld] p. 71
Theorem 13. of [RosserSchoenfeld] p. 71.

if I'm not mistaken, those should have been introduced with
https://github.com/metamath/set.mm/pull/2410

I'd like to understand if my procedure is "wrong" or the reformatting script was not applied in recent PRs

@tirix may it be the reformatting script was not run with PR 2410 ?

@jkingdon was the reformatting script run before this PR?

Please, let me know if I should run the script with my next PR (I would vote for this one) or leave it as it is now.

Glauco

benjub:
Yes, please run the script.  We strongly encourage it although for the moment it is not mandatory, so there may be small diffs like the one you noticed. We would need some kind of automation for the CI to run the script automatically, see #2381.

jkingdon:
> I'd like to understand if my procedure is "wrong" or the reformatting script was not applied in recent PRs

I totally agree with @benjub that automating this would be a good idea. I can try to remember to run `PATH=../metamath-exe:$PATH scripts/rewrap set.mm` (that would be sufficient, right?) but I'm afraid that I'm thinking of a lot of things when I submit a pull request and sometimes I forget even steps which I intend to do.

Your procedure sounds good and submitting those changes with your next pull request sounds good (as long as you do it with metamath.exe 0.198 and not an older version).

> @jkingdon was the reformatting script run before this PR?

What I ran was `PATH=../metamath-exe:$PATH scripts/minimize dfgcd3 set.mm` which as far as I can tell from a bit of experimenting covers some, but not all, of the differences I'm seeing.

PR #2416, opened 2022-01-02
Add rewrap checking to build

jkingdon:
Unless I've misunderstood the discussion to date, we would prefer that
this be checked automatically so that we don't have a lot of "where do
these diffs come from? how are we supposed to prevent or fix them
again?" sorts of issues.

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -120,12 +120,12 @@ jobs:
       # We'd probably rather have a bot push up a commit with the rewrapping
       # when a pull request is submitted, but until we set that up,
       # at least check that the rewrapping has been done.
-      - run: cp set.mm set-unwrapped.mm
-      - run: scripts/rewrap set.mm
-      - run: echo 'Checking for set.mm rewrapping:' && diff -U 0 set-unwrapped.mm set.mm
-      - run: cp iset.mm iset-unwrapped.mm
-      - run: scripts/rewrap iset.mm
-      - run: echo 'Checking for iset.mm rewrapping:' && diff -U 0 iset-unwrapped.mm iset.mm
+      - run: cp set.mm set-wrapped.mm
+      - run: scripts/rewrap set-wrapped.mm
+      - run: echo 'Checking for set.mm rewrapping:' && diff -U 0 set.mm set-wrapped.mm
+      - run: cp iset.mm iset-wrapped.mm
+      - run: scripts/rewrap iset-wrapped.mm
+      - run: echo 'Checking for iset.mm rewrapping:' && diff -U 0 iset.mm iset-wrapped.mm
       # The following checks are arbitrarily included in the metamath job.
       - run: echo 'Looking for tabs (not allowed)...' && ! grep "$(printf '\t')" set.mm
       - run: echo 'Looking for tabs (not allowed)...' && ! grep "$(printf '\t')" iset.mm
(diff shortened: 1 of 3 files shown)


jkingdon:
This is working as intended; here is a screenshot of what the failure looks liike

Screenshot from 2022-01-02 08-58-35

jkingdon:
Running scripts/rewrap locally and pushing up the change caused the check to go green.

benjub:
Yes, I think this is what at least @digama0, you and I agreed on in #2381, so thanks for adding this to verifiers.yml
I think this is not enough though. One has to add one pass of "rewrap" first, either to verifiers.yml or to scripts/rewrap, in order to conform to the script at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md (see the command and the paragraph "The reason for doing /rewrap first...").

jkingdon:
> I think this is not enough though. One has to add one pass of "rewrap" first, either to verifiers.yml or to scripts/rewrap

I think I vote for doing this in `scripts/rewrap`. I would like to have a simple command to run locally which means "make this OK with the github checks" and I suppose that's what `scripts/rewrap` is there for. What do people think?

benjub:
The current scripts/rewrap currently has the command:
> metamath "read ${source}" 'save proof */c/f' \\
  "write source ${source} /rewrap" quit

which I think should be rewritten:
> metamath "read ${source}" "write source ${source} /rewrap" \\
  erase "read ${source}" "save proof */compressed/fast" \\
  "write source ${source}" quit

benjub:
Even thouth it can be done later, can you modify scripts/rewrap (see my proposal above) ?
Second question : it looks like the commands you introduced in verifiers.yml simply echo the difference. Should I understand that a non-empty difference triggers a warning/error ? (and is (i)set-wrapped.mm automatically deleted at the end?)

jkingdon:
> Even thouth it can be done later, can you modify scripts/rewrap (see my proposal above) ?

Done. And I also pushed up a change to update CONTRIBUTING to describe the existence of this automated check.

> Second question : it looks like the commands you introduced in verifiers.yml simply echo the difference.

As with discouraged, diff exits with nonzero exit status and that causes a failure (see screenshot above). Just to double check and demonstrate should I make a draft pull request off this one which demonstrates the failure?

> is (i)set-wrapped.mm automatically deleted at the end?)

Yeah, I don't know all the details in github actions but it would be something along the lines of "the whole virtual machine goes away once the run is done". We don't seem to manually clean up `discouraged.new` (or the files we wget, or otherwise create), so unless a file is going to confuse the code later in the run I don't think it is a problem.

benjub:
Thanks for the commits and the answers. LGTM. So I gess we leave Issue #2381 open with the longer term goal of having an automation that makes these rewraps on the submitted files themselves ?

benjub (approved):

jkingdon:
> Thanks for the commits and the answers. LGTM. So I gess we leave Issue #2381 open with the longer term goal of having an automation that makes these rewraps on the submitted files themselves ?

Maybe it is better to close it and make a new issue if we find ourselves constantly cursing the checks (or don't make an issue if we get used to the checks without much worry)?

But I wasn't sure whether I should make that call unilaterally.

benjub:
I don't know either ! In any case, we can move the discussion there, and if we close the issue, say why and what can be done in the future to automate that task.

PR #2419, opened 2022-01-04
Shorten proof of sqr2irrlem

jvierling:
This PR shortens the proof of `sqr2irrlem`.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17819,6 +17819,7 @@ New usage of "sps-o" is discouraged (7 uses).
 New usage of "sq10OLD" is discouraged (1 uses).
 New usage of "sq10e99m1OLD" is discouraged (1 uses).
 New usage of "sqgt0sr" is discouraged (1 uses).
+New usage of "sqr2irrlemOLD" is discouraged (0 uses).
 New usage of "srhmsubcALTV" is discouraged (4 uses).
 New usage of "srhmsubcALTVlem1" is discouraged (2 uses).
 New usage of "srhmsubcALTVlem2" is discouraged (1 uses).
@@ -19721,6 +19722,7 @@ Proof modification of "spimvALT" is discouraged (9 steps).
 Proof modification of "sps-o" is discouraged (10 steps).
 Proof modification of "sq10OLD" is discouraged (57 steps).
 Proof modification of "sq10e99m1OLD" is discouraged (25 steps).
+Proof modification of "sqr2irrlemOLD" is discouraged (288 steps).
 Proof modification of "ssdisjOLD" is discouraged (49 steps).
 Proof modification of "sseliALT" is discouraged (152 steps).
 Proof modification of "sseqin2OLD" is discouraged (3 steps).


set.mm
```diff
@@ -182295,17 +182295,16 @@ infinite descent (here implemented by strong induction).  This is
        Metamath 100 proof #1.  (Contributed by NM, 20-Aug-2001.)  (Revised by
        Mario Carneiro, 12-Sep-2015.) $)
```
digama0, line 182296:
```suggestion
       Metamath 100 proof #1.  (Contributed by NM, 20-Aug-2001.)  (Proof
       shortened by Jannik Vierling, 4-Jan-2022.) $)
```

benjub:
Welcome and thanks for this first commit.
Can you keep the old version of the proof, in the same scope and right after sqr2irrlem, as
```
$( Obsolete proof of ~ sqr2irrlem as of 4-Jan-2022.  (Contributed by
   NM, 20-Aug-2001.)  (Revised by Mario Carneiro, 12-Sep-2015.)
    sqr2irrlemOLD $p |- ( ph -> ( ( A / 2 ) e. ZZ /\ ( B / 2 ) e. NN ) ) $=
      ( c2 cdiv co cz wcel cn cexp cmul 2cn oveq1d nncnd nnne0d eqtrd syl csqrt
      cfv cc wceq sqrtth ax-mp syl5eqr zcnd sqdivd nnsqcld divcan1d a1i cc0 wne
      sqcld 2ne0 divcan3d eqtr3d eqeltrd nnzd zesq mpbird clt wbr sqvali oveq2i
      divdiv1d 3eqtr4a zsqcl eqeltrrd nnrpd rphalfcld rpgt0d elnnz sylanbrc jca
      wb nnesq ) ABGHIZJKZCGHILKZAVTBGMIZGHIZJKZAWCAWCCGMIZLAGWENIZGHIWCWEAWFWB
      GHAWFWBWEHIZWENIWBAGWGWENAGBCHIZGMIZWGAGGUAUBZGMIZWIGUCKZWKGUDOGUEUFAWJWH
      GMFPUGABCABDUHZACEQACERUISPAWBWEABWMUOZAWEACEUJZQZAWEWORUKSPAWEGWPWLAOULZ
      GUMUNAUPULZUQURZWOUSUTABJKVTWDVQDBVATVBZAWAWEGHIZLKZAXAJKUMXAVCVDXBAVSGMI
      ZXAJAXCWCGHIZXAAWBGGMIZHIWBGGNIZHIXCXDXEXFWBHGOVEVFABGWMWQWRUIAWBGGWNWQWQ
      WRWRVGVHAWCWEGHWSPSAVTXCJKWTVSVITVJAXAAWEAWEWOVKVLVMXAVNVOACLKWAXBVQECVRT
      VBVP $.
$}
```
(and you will need to regenerate the `discouraged` file as indicated in https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md).  This way, the proofs can be compared, and the OLD proof is deleted one year later (unless someone judges that it is interesting to keep both for some reason, in which case it becomes an ALT proof).  Thanks.

set.mm
avekens: Usually, we keep the "Revised" tag, don't we? Therefore, we should have:

... (Contributed by NM, 20-Aug-2001.)  (Revised by Mario Carneiro, 12-Sep-2015.) (Proof shortened by Jannik Vierling, 4-Jan-2022.) $)

digama0: I think we only keep 2 credits, the first one and the latest change. I'm okay with doing it either way.

benjub (approved):

set.mm
benjub: I merged without realizing this suggestion was not added :(  I or @jvierling can add it if you're ok with that.  We'll also add your name to the "contributor list" at the top of the file, so that the credit can also be `(Proof shortened by JV, 4-Jan-2022.)`.

By the way, @digama0 : is there an easy tool to add one contributor to the three-column list (every line changes to keep alphabetical order...)?  If not, maybe we can make it single column ? (This time, it's ok since "V" is near the end of the list.)

avekens: > I think we only keep 2 credits, the first one and the latest change. I'm okay with doing it either way.

In my opinion, we should keep one tag of each kind (Contributed, Revised, Proof shortened). If a theorem is revised, the "Proof shortened" tag can be removed, since the proof would be changed because of the revision anyway.

digama0: @benjub You mean the contributor list with the initials? Pretty sure that was maintained manually. @avekens Let's do that.

jvierling: Since this PR is already merged, I'll open a new pull request (#2421) with the required changes.

benjub: As for the attribution tags: I would vote for keeping, obviously, the earlierst (which is the "contributed by") and also the latest. For the other ones, it should be on a case-by-case analysis, depending on the importance of the shortening/revision (especially when the latter is documented); if there were two important successive revisions, both should be kept.

PR #2420, opened 2022-01-05
Graph theory - minor improvements of Huneke's proof (2)

avekens:
* ~rabidim1 moved from GS's mathbox
* ~reueubd, ~ssdifim, ~ ssdifsym, ~eqeuel, ~eldifeldifsn added
* ~s3eq2, ~s3eqs2s1eq, ~s3eq3seq added
* ~frgreu: proof shortened and moved up
* ~frgrwopreglem2, ~frgrwopreglem5 revised
* proof of ~frgrwopreglem3, ~frgrwopreglem4, ~frgrwopreg, ~frgrwopreg2, ~frgrregorufr0, ~frgr2wwlkeu, ~frgr2wwlk1 shortened


jkingdon (approved):

set.mm
```diff
@@ -47381,6 +47411,16 @@ practical reasons (to avoid having to prove sethood of ` A ` in every use
       BJZACHADCKPQACBOLMN $.
   $}
 
+  ${
+    $d x y A $.
+    $( A condition which implies the existential uniqueness of elementhood.
+       (Contributed by AV, 4-Jan-2022.) $)
+    eqeuel $p |- ( ( A =/= (/) /\ A. x A. y ( ( x e. A /\ y e. A ) -> x = y ) )
+                   -> E! x x e. A ) $=
```
benjub, line 47419: "A condition which implies the existence of a unique element of a class." ?

Is there in set.mm a theorem of this kind for any predicate ? Something like
x = y -> ( ph <-> ps ) (with DVs so that ph is P(x) and ps is P(y))
x = z -> ( ph <-> ch ) (with DVs so that ch is P(z))
 ==>
A. y A. z ( ( ps /\ ch ) -> y = z ) <-> E* x ph

This could then be applied to ph being "x e. A" to prove your theorem.

set.mm
```diff
@@ -46776,6 +46793,19 @@ practical reasons (to avoid having to prove sethood of ` A ` in every use
...
 
+  $( Implication of a class difference with a subclass.  (Contributed by AV,
+     3-Jan-2022.) $)
+  ssdifim $p |- ( ( A C_ V /\ B = ( V \ A ) ) -> A = ( V \ B ) ) $=
+    ( wss cdif wceq dfss4 eqcom sylbb difeq2 eqcomd sylan9eq ) ACDZBCAEZFZACNEZ
+    CBEZMPAFAPFACGPAHIOQPBNCJKL $.
+
+  $( Symmetric class differences for subclasses.  (Contributed by AV,
+     3-Jan-2022.) $)
+  ssdifsym $p |- ( ( A C_ V /\ B C_ V )
```
benjub, line 46804: This looks like ssconb. See if shorter proof.

benjub (approved):

set.mm
avekens: The specialty of this theorem is the condition A =/= (/), which is used instead of E. x ph (with x e. A as ph in this case).

set.mm
avekens: Although these theorems look similar, I do not see any direct relationship (regarding to use the one theorem in the proof of the other) between them. But maybe I should move up ~ssdifim and ~ssdifsym near to ~ssconb.

avekens: PS: not possible to move up ~ssdifim and ~ssdifsym , because they depend on ~dfss4.

benjub: Ah yes, I missed the inclusions instead of equalities

set.mm
benjub: I saw that. So you have a condition for existence and a condition for uniqueness; you deduce from these two a condition for existential uniqueness (namely, the conjunction of said conditions).

avekens: Yes, correct. This corresponds to the nearby Theorems of this kind, e.g., ~n0moeu.

benjub: My point is that a more natural proof of eqeuel would be to deduce it from the theorem I sketched above. So that's a future project.
For now, I proposed above a rewording of the comment.  "elementhood" is synonymous with "membership", so the comment means something different from what is proved.

avekens: OK, I understood your first remark as question, not as proposal for a rewording. And yes, your suggestion is more appropriate. I changed the wording in my latest commit.

benjub (approved):

PR #2422, opened 2022-01-06
typos

benjub:


set.mm
```diff
@@ -530526,7 +530530,7 @@ automatically a set (see ~ bj-mooreset ).  (Contributed by BJ,
```
avekens, line 530530: Why did you change the order of theses definitions? I think it is usual to first declare (all) new constant symbols (within a section), then to define their types and then provide the corresponding definitions. See, for example, section "Relations"

set.mm
```diff
@@ -530574,10 +530578,10 @@ automatically a set (see ~ bj-mooreset ).  (Contributed by BJ,
 
 $)
 
-  $( Symbol for the nonnegative reals. $)
+  $( Token for the nonnegative reals. $)
   $c RR>=0 $.
 
-  $( Symbol for the (strictly) positive reals. $)
+  $( Token for the (strictly) positive reals. $)
   $c RR>0 $.
```
avekens, line 530586: What's wrong with "symbol"? According to the Metamath book (section 4.2.1) "... symbols are [user-defined] tokens ...". Actually, $c defines "(math) symbols" (see section 4.2.3), and "tokens" are more general (" ... tokens, which are strings of non-whitespace printable characters ... separated by white space ...", see section 4.2).

set.mm
benjub: This can be changed, but I thought it was preferred to have "token-syntax-definition" as three consecutive statements for most symbols.  For sets of symbols which only make sense as a whole, one may order them as you wrote.  See #2404 (where the general case is not settled); for the moment it's more on a case-by-case basis.

set.mm
benjub: No strong opinion.  To me, the symbol is the thing that is displayed on the webpage.  In other words, the typesetting definitions at the end of set.mm (by the way, the title of the appendix uses "token"), are of the form:
>htmldef "token" as "unicode-symbol";
  althtmldef "token" as "gif-symbol";
  latexdef "token" as "latex-symbol";

But again, not a golden rule to me.

set.mm
avekens: If the definitions are placed in separate (sub)sections, as done in #2404, then it is OK. But curry_ and uncurry_ belong together, i.e. are a "set of (related) symbols", so they would remain in the same (sub) section.

jkingdon (approved):

PR #2424, opened 2022-01-06
The inferior limit (lim inf)

glacode:
among the new theorems, some that could be of interest:

- "basic" properties: ~ liminfcl , ~ liminflelimsupuz

- relationships between convergence, limsup and liminf: from ~ climliminflimsup to ~ climliminflimsup4

- the inferior limit of a countable set of sigma-measurable functions is sigma-measurable: ~ smfliminf

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18525,7 +18525,6 @@ Proof modification of "bj-elissetv" is discouraged (25 steps).
 Proof modification of "bj-equs45fv" is discouraged (43 steps).
 Proof modification of "bj-equsal" is discouraged (31 steps).
 Proof modification of "bj-equsalhv" is discouraged (10 steps).
-Proof modification of "bj-equsalv" is discouraged (39 steps).
 Proof modification of "bj-equsb1v" is discouraged (17 steps).
 Proof modification of "bj-equsexval" is discouraged (38 steps).
 Proof modification of "bj-eunex" is discouraged (53 steps).
--- a/discouraged
+++ b/discouraged
@@ -17819,6 +17819,7 @@ New usage of "sps-o" is discouraged (7 uses).
 New usage of "sq10OLD" is discouraged (1 uses).
 New usage of "sq10e99m1OLD" is discouraged (1 uses).
 New usage of "sqgt0sr" is discouraged (1 uses).
+New usage of "sqr2irrlemOLD" is discouraged (0 uses).
 New usage of "srhmsubcALTV" is discouraged (4 uses).
 New usage of "srhmsubcALTVlem1" is discouraged (2 uses).
 New usage of "srhmsubcALTVlem2" is discouraged (1 uses).
@@ -19721,6 +19722,7 @@ Proof modification of "spimvALT" is discouraged (9 steps).
 Proof modification of "sps-o" is discouraged (10 steps).
 Proof modification of "sq10OLD" is discouraged (57 steps).
 Proof modification of "sq10e99m1OLD" is discouraged (25 steps).
+Proof modification of "sqr2irrlemOLD" is discouraged (288 steps).
 Proof modification of "ssdisjOLD" is discouraged (49 steps).
 Proof modification of "sseliALT" is discouraged (152 steps).
 Proof modification of "sseqin2OLD" is discouraged (3 steps).


glacode:
Mathbox only (I feel so stupid I cannot find a way to add the nice green label...)

benjub:
> Mathbox only (I feel so stupid I cannot find a way to add the nice green label...)

It's a cranked wheel next to "Labels" on the right panel, but maybe permissions are needed for it to be available.

avekens:
> Mathbox only (I feel so stupid I cannot find a way to add the nice green label...)

I think you need special permissions for setting labels. I also cannot do it ...

benjub:
@glacode Do you want to modify ~climlimsupcex in view of the discussion group thread ?

There are about 2100 lines to review (cursorily, since it's in a mathbox). Can we split ? @avekens ?  @tirix ?  700 lines each ?

avekens (approved):
I had a look at the changes up to line 658261 (no remarks), but stopped at 0cnv because of the google group discussion. Therefore, I give my approval up to line 658261

benjub:
> I had a look at the changes up to line 658261 (no remarks), but stopped at 0cnv because of the google group discussion. Therefore, I give my approval up to line 658261

Thanks. I call dibs on 658261--663996.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
       ALVMEFHIACFBUTGJVDKVEVF $.
   $}
 
+  ${
+    $d k m x y $.
+    $( If (/) is a complex number, then it converges to itself.  (see ~ 0ncn
+       and its comment ; see also the comment in ~ climlimsupcex ) (Contributed
+       by Glauco Siliprandi, 2-Jan-2022.) $)
+    0cnv $p |- ( (/) e. CC -> (/) ~~> (/) ) $=
```
benjub: See discussion group thread. In set.mm, one has `|- -. (/) e. CC`, therefore one has `|- ( (/) e. CC -> ph )`. If your proof exhibits a special feature, then maybe extract that subproof to state precisely that.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+  $}
+
+  ${
+    $d A j k x $.  $d F j k x $.  $d M j $.  $d Z j k $.  $d j k ph x $.
+    climuzlem.1 $e |- ( ph -> M e. ZZ ) $.
+    climuzlem.2 $e |- Z = ( ZZ>= ` M ) $.
+    climuzlem.3 $e |- ( ph -> F : Z --> CC ) $.
+    $( Express the predicate:  The limit of complex number sequence ` F ` is
+       ` A ` , or ` F ` converges to ` A ` .  (Contributed by Glauco
+       Siliprandi, 2-Jan-2022.) $)
```
benjub: Maybe label it climuzv and add to the comment a sentence: "This is the same as ~ climuz with a dv condition on ` k , F ` instead of a non-freeness hypothesis." [if I understood correctly]

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      QZRZSTZSKUCZBKLZDNUGZACNUECSEUFBSUHVIVSUIFGHBCIJEVPVPUJZUKULAVRWDINAVJNUH
+      ZUMZVQWCBKWOJVJVOWCNVPUNWMVKVJUOZSVNWBKWPVMWASWPVLVTEVKVJOPUPUQURUTAWNUSW
+      CUNUHWOSWBKVAVBVCVDVEVFWEWLUIAWDWKIDNVJWFUOZWCWJBKWQSWBWIKWQWAWHSWQVTWGEV
+      JWFOPUPUQURUTVEVGVCVH $.
+  $}
+
+  ${
+    $d A w x y z $.  $d B w x y z $.  $d F x $.
+    $( Ordering property of the superior limit function.  (Contributed by
+       Glauco Siliprandi, 2-Jan-2022.) $)
```
benjub, line 663512: superior --> inferior
The comment is a bit strange (I see it is the same for limsupgord by @digama0). This statement is not about inferior/superior limit: it is simply used in some liminf/limsup results, but it merely says that the sup on a larger set is larger. What about "The infimum of an extended real function is a non-increasing function from uppersets of reals ordered by inclusion to the extended reals." and similarly with limsupgord ? @digama0 ?

It would be nice (future project) to prove these from the general theorem:
(A and B are subsets of an ordered set (R,<) with A C_ B) -> sup (A,R,<) <= sup (B,R,<) [defining <= to be "< or equal"]
and deduce from it
(f: X --> R and (R, <) is an ordered set and A and B are subsets of X with A C_ B) -> sup (F"A,R,<) <= sup (F"B,R,<)
and then our theorems.

More generally: I haven't checked the details of your proofs, but I imagine you took advantage of the fact that liminf for the standard order is limsup for the opposite order, to reuse the theorems for limsup ?

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  $c liminf $.
+
+  $( Extend class notation to include the liminf function. $)
+  clsi $a class liminf $.
+
+  $d x k $.
+  $( Define the inferior limit of an infinite sequence of extended real
```
benjub, line 663485: an infinite sequence --> a sequence ?
(actually, it makes sense for any extended real function defined on a subset of RR which is not upper-bounded) [<-- you can add this to the comment if you want]

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      LVNUQUOURVRPTVMPIZVTVSUSVQPUEWBWGWCVLUTVADVRVMVBVCVD $.
+  $}
+
+  ${
+    $d F k $.
+    limsupvald.1 $e |- ( ph -> F e. V ) $.
+    limsupvald.2 $e |- G = ( k e. RR |-> sup ( ( ( F " ( k [,) +oo ) ) i^i RR*
+       ) , RR* , < ) ) $.
+    $( The superior limit of an infinite sequence ` F ` of extended real
+       numbers, which is the infimum of the set of suprema of all upper
```
benjub, line 663531: an infinite sequence --> a sequence (or as in the comment of df-liminf)
, which is --> is
... of all subsequences of F defined on an upperset (or, if you consider any function F as in the comment of df-liminf, "... of all restrictions of F to an upperset of reals")

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      PQ $.
+  $}
+
+  ${
+    $d A x $.  $d Z x $.
+    limsupresicompt.a $e |- ( ph -> A e. V ) $.
+    limsupresicompt.m $e |- ( ph -> M e. RR ) $.
+    limsupresicompt.z $e |- Z = ( M [,) +oo ) $.
+    $( The superior limit doesn't change when a function is restricted to the
+       upper part of the reals.  (Contributed by Glauco Siliprandi,
```
benjub, line 663545: restricted to an upperset of reals.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+    limsupresicompt $p |- ( ph -> ( limsup ` ( x e. A |-> B ) ) = ( limsup ` (
+       x e. ( A i^i Z ) |-> B ) ) ) $=
+      ( cmpt cres clsp cfv cin cvv mptexd limsupresico wceq resmpt3 a1i fveq2d
+      eqtr3d ) ABCDKZGLZMNUDMNBCGODKZMNAUDEPGIJABCDFHQRAUEUFMUEUFSABCGDTUAUBUC
+      $.
+  $}
+
+  ${
+    limsupcli.1 $e |- F e. V $.
+    $( Closure of the superior limit.  (Contributed by Glauco Siliprandi,
```
benjub, line 663556: A superior limit is an extended real.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+       2-Jan-2022.) $)
+    limsupcli $p |- ( limsup ` F ) e. RR* $=
+      ( wcel clsp cfv cxr limsupcl ax-mp ) ABDAEFGDCABHI $.
+  $}
+
+  ${
+    liminfgf.1 $e |- G = ( k e. RR |-> inf ( ( ( F " ( k [,) +oo ) ) i^i RR* )
+       , RR* , < ) ) $.
+    $( Closure of the inferior limit function.  (Contributed by Glauco
+       Siliprandi, 2-Jan-2022.) $)
```
benjub, line 663566: I don't understand why "closure" is used here, but I see it is in many other places in set.mm so maybe it comes from computer science. Leave it if it makes sense to you. Also, "inferior limit function" may not be the best description, but leave it if it's consistent with your sources.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+  $}
+
+  ${
+    $d F k m x $.  $d M k m x $.  $d Z k m x $.  $d k m ph x $.
+    climlimsup.1 $e |- ( ph -> M e. ZZ ) $.
+    climlimsup.2 $e |- Z = ( ZZ>= ` M ) $.
+    climlimsup.3 $e |- ( ph -> F : Z --> RR ) $.
+    $( A sequence of real numbers converges if and only if it converges to its
+       superior limit.  The first hypothesis is needed (see ~ climlimsupcex for
+       a counterexample) (Contributed by Glauco Siliprandi, 2-Jan-2022.) $)
```
benjub, line 663593: I would remove "The first..." (see discussion group), since when writing ( ZZ>= ` M ) , it is more or less expected that M e. ZZ (with the current definition; one could adopt a definition of ZZ>= making sense for any argument in RR*).

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      ZVMEVRTAJUMACKVQVLGAVHKSZUTLVJVKMAVJVKTVSAVKVJAVJLUNVKVJTABLDVIIUOVJLUPVA
+      UQURUSVBVCVDVEVF $.
+  $}
+
+  ${
+    $d A k x y $.  $d F k x y $.  $d k ph x y $.
+    limsupresxr.1 $e |- ( ph -> F e. V ) $.
+    limsupresxr.2 $e |- ( ph -> Fun F ) $.
+    limsupresxr.3 $e |- A = ( `' F " RR* ) $.
+    $( The superior limit of a function only depends on the preimage of the
```
benjub, line 663680: ...only depends on the restriction of that function to the preimage of the set of extended reals.
(same thing in the next statement)

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      UVDWBUVEAUVFWBYAXSXLVOXTNPZUVBYQYOYPTVDUVEUVSUVFXSWAVSUVHXTYBXMXKXPXQ $.
+  $}
+
+  ${
+    climlimsupcex.1 $e |- -. M e. ZZ $.
+    climlimsupcex.2 $e |- Z = ( ZZ>= ` M ) $.
+    climlimsupcex.3 $e |- F = (/) $.
+    $( Counterexample for ~ climlimsup , showing that the first hypothesis is
+       needed, if the empty set is a complex number (see ~ 0ncn and its
+       comment) (Contributed by Glauco Siliprandi, 2-Jan-2022.) $)
```
benjub, line 663769: Again, see if you want to change this in view of the discussion group thread.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      ( wcel clsi cfv cxr liminfcl syl ) ABCEBFGHEDBCIJ $.
+  $}
+
+  ${
+    $d F k $.  $d Z k y $.  $d k ph y $.
+    liminfresico.1 $e |- ( ph -> M e. RR ) $.
+    liminfresico.2 $e |- Z = ( M [,) +oo ) $.
+    liminfresico.3 $e |- ( ph -> F e. V ) $.
+    $( The inferior limit doesn't change when a function is restricted to the
+       upper part of the reals.  (Contributed by Glauco Siliprandi,
```
benjub, line 663795: Same comment as above.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+    $d F i j l $.  $d X i j k l $.  $d Z i j l $.  $d i l ph $.
+    limsupgt.k $e |- F/_ k F $.
+    limsupgt.m $e |- ( ph -> M e. ZZ ) $.
+    limsupgt.z $e |- Z = ( ZZ>= ` M ) $.
+    limsupgt.f $e |- ( ph -> F : Z --> RR ) $.
+    limsupgt.r $e |- ( ph -> ( limsup ` F ) e. RR ) $.
+    limsupgt.x $e |- ( ph -> X e. RR+ ) $.
+    $( Given a sequence of real numbers, there exists an upper part of the
+       sequence that's appxoximated from below by the superior limit.
+       (Contributed by Glauco Siliprandi, 2-Jan-2022.) $)
```
benjub, line 663986: appxoximated --> approximated (same above)
I would write something like: "For any positive real, the superior limit of F is larger than any of its values at large enough arguments, up to that positive real."

benjub (commented):
Reviewed lines 658261--663996, and my comments are only suggestions, since it's in your mathbox.
Also, when a statement is simply a lemma labeled e.g. xxxlem, the comment may simply say "Lemma for xxx", possibly with a comment about the lemma, but do not repeat the comment of xxx in xxxlem.  I noticed that many of your lemmas can be commented as: "Lemma for xxx : same conclusion as xxx but with dv conditions on... / with the hypothesis that..."

set.mm
glacode: > See discussion group thread.

Which thread? Can you please post a link?

set.mm has

|- -. (/) e. CC

but it is a construction-dependent property, in fact new usage is discouraged.

Here is Norman's comment in ~0ncn

"Note: do not use this after the real number axioms are developed, since it is a construction-dependent property."

Thus, actually we don't have the theorem

|- -. (/) e. CC

On the contrary,

|- ( (/) e. CC -> (/) ~~> (/) )

would be proven with any arbitrary construction of the complex numbers (it is construction independent). That's why I put it in.

set.mm
glacode: I guess, I simply adopted Mario's comment for limsupgf (of course, I changed superior with inferior)

set.mm
benjub: This one: https://groups.google.com/g/metamath/c/BXJq5o5i8a8/m/WowODKouCgAJ

After looking at df-clim, I think I can see what you're doing (I haven't looked at your proof yet, I'll do it when it's merged).  In the expression (/) ~~> (/), the (/) on the left is the empty function, the (/) on the right is a complex number, and you're probably using the fact that a function evaluates outside its domain to the empty set (which is due to the particular encoding used in set.mm), so this is abusing the encoding used in set.mm. It's not meant to be a criticism at all, but I think we agree there is no mathematical content in this theorem ?  Of course feel free to leave it as is in your mathbox.  If you think I'm missing something, I'll look at the proof once on the website.

set.mm
glacode: I believe I have used the same comment as ~ limsupval (for consistency)

To say the truth, I was not so happy about neither the "sequence" nor the "subsequence" terms, because the definition would be "meaningful" for any set: in particular, for functions with dense domain in RR (or defined on all RR).
But the "standard" use will be with sequences, that's why I adapted Norman's comment.

set.mm
glacode: Much better, thank you for pointing it out. I'll fix it.

set.mm
glacode: I simply adopted Norman's comment for ~ limsupcl  (for consistency)

I agree your comment provides more information (but may be "Closure of..." is a "standard" comment in set.mm ? )

set.mm
glacode: > This one: https://groups.google.com/g/metamath/c/BXJq5o5i8a8/m/WowODKouCgAJ
> 
> In the expression (/) ~~> (/), the (/) on the left is the empty function, the (/) on the right is a complex number

Exactly. I was investigating the behavior of ~~> when applied to the empty function.

Thinking about it, I guess I could complete it with another theorem like

|- ( -. (/) e. CC -> -.  (/) e. dom ~~> )

that is, "assuming that the empty set is not a complex number, the empty function doesn't converge".

(but I've not proven it yet, so I'm not sure about it).

I absolutely agree, there's no mathematical content in this theorem, just a couple of assertions to clarify the behavior of our particular definition of convergence when applied to the empty set.
Of course, our definition of convergence agrees with what we find in textbooks for the "standard" case, i.e. sequences, but with metamath we often have to deal with edge cases that textbooks overlook, so it might be useful to write it once and keep if for future reference: if, in the future, someone will have to deal with an edge case and will search for "(/) ~~>" , she will see that it depends on an "undecidable" antecedent and look no further.

glacode: w.r.t. the thread, I asked about

-. +oo e. CC

because, while for -. (/) e. CC it is clear to me Norman's intent to state it as something that we CANNOT rely upon (no textbook will ever "take a position" about the empty set being a complex number or not), it was not so clear to me if math literature had an "official" position about chosing +oo and -oo distinct not only from the reals, but from the complex numbers too.

benjub: Since we don't want the empty function to converge to anything, your theorem shows that we'd better require that -. (/) e. CC, as is rightly the case (or that the value of a function out of its domain be not the empty set). Similarly, we'd better have +oo not in any other set that could come up, to avoid clashes and "accidental" or "junk" theorems (for instance: "plus infinity belongs to the algebraic closure of RR"!!). See for instance https://groups.google.com/g/metamath/c/GQmelsKtaWQ

avekens:
There was a discussion about the term "closure" in this PR (I cannot find it anymore - is it removed, because it is resolved?).

I think the term "closure" is a little misused in set.mm, many times meaning simply "membership". But it in the original sense, it means "if the argument/operands belong to a subset of the domain the function/operation value also belongs to this subset (see Wikipdia "Closure (mathematics)": "a subset of a given set is **closed** under an operation of the larger set if performing that operation on members of the subset always produces a member of that subset. . See, for example ~zaddcl: ( ( M e. ZZ /\ N e. ZZ ) -> ( M + N ) e. ZZ ) with + : ( CC X. CC ) --> CC. 

For ~ceilcl, for example, I think the usage of "closure" is maybe not correct: ( A e. RR -> ( |^ ` A ) e. ZZ ) This theorem simply states that the function value belongs to the range/codomain of the function. In this case, however, there is no theorem |^ : RR --> ZZ, but it could be simply derived from ~ceilcl. Therefore, I think the term "closure" is exactly used also for such cases: "The range of a function/operation is (a subset of) a special class".

benjub:
@avekens wrt "closure": you're right about the correct usage of "closure" (first paragraph). Other example: "a topology is a family of subsets *closed under* unions and finite intersections".  There is also a use in set.mm of "reverse closure" e.g. things like (x+y) e. z -> x e. z.

I agree that "closure" is used a bit too widely in set.mm, and ~ceicl, the usage of "closure" is not correct.  This can be corrected on a case by case basis if you encounter them.

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      AULUKUMPUKUHUIMAUKUIUHUKUCUDUEUFUG $.
+  $}
+
+  ${
+    $d A k y z $.  $d A x y $.  $d F k y z $.  $d k ph y z $.
+    liminfvalxr.1 $e |- F/_ x F $.
+    liminfvalxr.2 $e |- ( ph -> A e. V ) $.
+    liminfvalxr.3 $e |- ( ph -> F : A --> RR* ) $.
+    $( Alternate definition of ` liminf ` when ` F ` is an extended real valued
+       Function.  (Contributed by Glauco Siliprandi, 2-Jan-2022.) $)
```
benjub, line 664055: Function --> function

set.mm
```diff
@@ -662596,6 +663406,1223 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+      BVQVCVKVDVEVFADKVQBGVGVHWDWIVIVKVLVJVMVNVOVP $.
+  $}
+
+  ${
+    $d A x $.
+    liminfvalxrmpt.1 $e |- F/ x ph $.
+    liminfvalxrmpt.2 $e |- ( ph -> A e. V ) $.
+    liminfvalxrmpt.3 $e |- ( ( ph /\ x e. A ) -> B e. RR* ) $.
+    $( Alternate definition of ` liminf ` when ` F ` is an extended real valued
+       Function.  (Contributed by Glauco Siliprandi, 2-Jan-2022.) $)
```
benjub, line 664136: function

set.mm
glacode: I've fixed superior --> inferior (see new commit)
If Mario agrees, we can change both comments as you suggested (I'd prefert to do it after the merge, so that I can dedicate a new pull request just to clean up the mathbox and minimize the risk of conflicts, becasue this pending PR is pretty large)

set.mm
glacode: In my last commit:
- an infinite sequence --> a sequence
- added (actually, it makes sense for any extended real function defined on a subset of RR which is not upper-bounded)
I like the additional comment, I wasn't so happy with "sequence" because the definition is more general than that.

set.mm
glacode: last commit:
- I've replaced an infinite sequence --> a sequence
- , which is --> is
- now it is    ... of the set of suprema of all restrictions of ` F ` to an upperset of reals

set.mm
glacode: Changed as suggested (in the last commit)

set.mm
glacode: New commit, now it is:
The inferior limit doesn't change when a function is restricted to an upperset of reals.

set.mm
glacode: New commit:
changed as suggested

set.mm
```diff
@@ -662642,6 +663452,1224 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
+  $}
+
+  ${
+    $d A x $.  $d M x $.
+    liminfval4.x $e |- F/ x ph $.
+    liminfval4.a $e |- ( ph -> A e. V ) $.
+    liminfval4.m $e |- ( ph -> M e. RR ) $.
+    liminfval4.b $e |- ( ( ph /\ x e. ( A i^i ( M [,) +oo ) ) ) -> B e. RR )
+                  $.
+    $( Alternate definition of ` liminf ` when the given a function is
```
benjub, line 664228: when the given function

benjub (approved):
Approved after a very cursory look at the remaining third of the PR.  I made a few suggestions related to comments.  If you have time and haven't done so, you may do a "MM>minimize_with *" on these proofs.

set.mm
glacode: fixed in e5878437

set.mm
glacode: fixed in e5878437

set.mm
glacode: fixed in e5878437 (thank you for the revision!)

PR #2426, opened 2022-01-09
Minimize

benjub:
Two automatic minimizations.
Theorem feqresmpt2 by @glacode seems to be a weaker version of resmptd. Can I remove it ?


glacode:
it looks like it actually is weaker,

Does it automatically "adjust" the invoking theorem?   (supcnvlimsupmpt)
If that's the case, feel free to remove it.

Is there a page where I can learn how to use minimize in an "effective" way?

avekens (approved):

benjub:
Minimized ~supcnvlimsupmpt and removes ~feqresmpt2.

@glacode : I typed:
>./metamath
r set.mm
pr supcnvlimsupmpt
min *
sa n/C
exi
wr so set.mm/rew
exi

(short for: read set.mm, prove supcnvlimsupmpt, minimize_with \*, save new proof in compressed format, exit, write source and rewrap, exit) The help of the metamath program (accessed via MM> help "command") is pretty well written.

When proving theorems with metamath, I always do a "min \*" at the end. If you use mmj2 then it's less convenient of course.  You can do various sorts of "bulk minimize" (i.e., try and minimize a list of theorems). An example is given in the comment in scripts/min.cmd (to minimize all theorems using a given list of theorems), which you can adapt to "minimize all theorems with labels bj-\*" (this is what I do), or "minimize all theorems in the range thma~thmb", etc.

jkingdon:
> When proving theorems with metamath, I always do a "min *" at the end. If you use mmj2 then it's less convenient of course.

One easy way (to minimize the theorem you just proved anyway) is with scripts/minimize.

For example, I just wrote a proof for `bezout` in mmj2 and after I pasted the proof from mmj2 to iset.mm I ran:

```
scripts/minimize bezout iset.mm
```

This also usually does enough rewrapping to keep the pull request from failing for wrapping reasons.

jkingdon (approved):

glacode:
@jkingdon thanks for the advice.

I believe most mmj2 users write a proof and then move it to set.mm (I guess that's the case, because when there was a bug with loading .mmt files, it went unnoticed for some time; furthermore, I see that's the way @tirix 's Emetamath plugin is working)

My workflow is different: I keep (often more than one hundred) .mmp and .mmt files in a folder, and I move the .mmt's to set.mm when I do a PR.

I could move every single proof back and forth to set.mm just to minimize it, but it would slow down my workflow a bit.

I'm now trying to build a couple of scripts that fits well with my workflow. I'm looking at 1.cmd and min.cmd that @benjub suggested: I'd like to specify a range of theorems and run minimize on those.

I'm going to study metamath.exe commands: thanks in advance, if anybody reading this message can reply with the metamath command to list the labels from **firstlabel** to **lastlabel** possibly in the format that min.cmd requires (for example, "show usage" lists labels in the proper way, for min.cmd)

Glauco

glacode:
show labels ifcli~nel2nelini

writes the list of labels in the range, but the produced log file cannot be given to 1.cmd , as is

show labels ifcli~nel2nelini / LINEAR

also requires some post-processing, I'll work on it. Please let me know if there is something already written (for minimizing ranges of theorems)

Issue #2298, opened 2021-11-05
Prove that the square root of two is irrational in iset.mm


jkingdon:
That is, formalize https://en.wikipedia.org/wiki/Square_root_of_2#Constructive_proof or some other proof that the square root of two is apart from any rational number (as described at http://us.metamath.org/ileuni/sqrt2irr.html this is different from "the square root of two is not rational").

Assuming we stick with that proof, the steps are

* given positive integers `a` and `b` , `( 2 x. ( b ^ 2 ) ) =/= ( a ^ 2 )`
  * use http://us.metamath.org/mpeuni/oddpwdc.html to decompose ` ( 2 x. ( b ^ 2 ) ) ` and ` ( a ^ 2 ) ` each into a power of two times an odd number and apply the logic that the exponents are odd and even respectively
  * ~~To prove oddpwdc, the set.mm proof uses~~ _oddpwdc is now proved at http://us2.metamath.org/ileuni/oddpwdc.html_
    * ~~The ` sup ` notation, including uzsupss , supcl , and supub to express the idea "the largest power of two which divides a natural number". There might be a way to express this via induction or something but supremum as used in the set.mm proof runs immediately afoul of http://us.metamath.org/ileuni/nnregexmid.html .~~ _Induction indeed works as seen at https://github.com/metamath/set.mm/pull/2319_ 
    * ~~`ssfi` to show that "the set of powers of two which divide a natural number" is finite. Although we can't have `ssfi` ( see http://us.metamath.org/ileuni/ssfiexmid.html ), we probably could show this set is finite if we need to.~~ _We didn't end up needing this_
    * ~~A number of theorems using http://us.metamath.org/mpeuni/df-dvds.html notation. Presumably intuitionizable~~ _I assume these are now in iset.mm. At least, most of the theorems in the df-dvds section and the next one are._
    * ~~Some of the usual intuitionizing as described in the Metamath Proof Explorer cross reference at http://us.metamath.org/ileuni/mmil.html#setmm~~ _done_
    * ~~Some other theorems from set.mm which aren't in iset.mm yet: http://us.metamath.org/mpeuni/nexple.html , http://us.metamath.org/mpeuni/f1od2.html , http://us.metamath.org/mpeuni/cnvoprab.html , and http://us.metamath.org/mpeuni/spc2ed.html .~~ _The first one is handled by http://us.metamath.org/ileuni/bernneq3.html and the others are in https://github.com/metamath/set.mm/pull/2319_
[...]

PR #2427, opened 2022-01-10
Finish iset.mm proof of Bézout's identity

### LINKED ISSUE #2399: Prove Bézout's identity in iset.mm

jkingdon:
This proof is via the Extended Euclidean Algorithm and much of it was already merged in #2413 . This pull request completes the job, mainly in terms of showing that the number which is shown to exist there is the greatest common divisor as defined by df-gcd , and a few other odds and ends.

This means that iset.mm goes from one entry to two on the Metamath 100 list :tada: . Furthermore, Bézout's identity is expected to be needed to prove #2298 which is another Metamath 100 theorem.

Fixes #2399

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -422,7 +422,10 @@ href="mpeuni/hashbc.html">hashbc</a>, by Mario Carneiro, 2014-07-13)</li>
 
 <!-- 20th added to list -->
 <li><a name="60">60</a>.  Bezout's Theorem (<a
-href="mpeuni/bezout.html">bezout</a>, by Mario Carneiro, 2014-02-22)</li>
+href="mpeuni/bezout.html">bezout</a>, by Mario Carneiro, 2014-02-22).
+The intuitionistic logic explorer (iset.mm) database also includes <a
+href="http://us.metamath.org/ileuni/bezout.html">bezout</a>
+(by Mario Carneiro and Jim Kingdon, 2022-01-09).</li>
 
 <!-- 67th added to list -->
 <li><a name="61">61</a>.  Theorem of Ceva (<a


digama0 (approved):
:tada:

PR #2428, opened 2022-01-10
STS definition file update

tirix:
This is an update of the MathML STS definition file for set.mm.

### DIFF
--- a/set-mathml.mmts
+++ b/set-mathml.mmts
@@ -1619,7 +1619,7 @@ $s class-n 7 $: 7 $.
 $s class-n 8 $: 8 $.
 $s class-n 9 $: 9 $.
 $s class-n ; M N $: #M##N# $.
-$s class -u N $: -#N# $.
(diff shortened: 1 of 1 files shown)


jkingdon (approved):

tirix:
Commit and PR CI passed with these changes, however the CI on the `develop` branch failed after this was merged.
The failure does not seem to be related with the content of this PR (which actually shall have no impact on CI).
I suppose this is a transient problem and the next commit in `develop` will clear the issue.

jkingdon:
> Commit and PR CI passed with these changes, however the CI on the `develop` branch failed after this was merged. The failure does not seem to be related with the content of this PR (which actually shall have no impact on CI). I suppose this is a transient problem and the next commit in `develop` will clear the issue.

I chose "re-run all jobs" (upper right of the screen which shows the failures) and then it passed. So it indeed was a transient problem.

PR #2430, opened 2022-01-11
Finish Bezout's identity section in iset.mm

jkingdon:
This section intuitionizes without much trouble.

The diff is perhaps slightly hard to read because of the last commit, which switches the order of two sections.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1055,6 +1055,9 @@ pm2.61iii , pm2.61ian , pm2.61dan , pm2.61ddan , pm2.61dda , pm2.61ine , pm2.61n
 pm2.61dne , pm2.61dane , pm2.61da2ne , pm2.61da3ne , pm2.61iine
 </TD>
 <TD><I>none</I></TD>
+<TD>If the proposition being eliminated is decidable
+(for example due to ~ nndceq , ~ zdceq , ~ zdcle , ~ zdclt , ~ eluzdc , or ~ fzdcel ),
+then case elimination will work using theorems such as ~ exmiddc and ~ mpjaodan</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1109,7 +1109,7 @@ pm2.61dne , pm2.61dane , pm2.61da2ne , pm2.61da3ne , pm2.61iine
 
 <TR>
 <TD>ianor</TD>
-<TD>~ pm3.14 </TD>
+<TD>~ pm3.14 , ~ ianordc</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -105767,549 +105767,549 @@ which defines the set we are taking the supremum of must (a) be true at
...
+      YJYKYLYMVMVNVLWMYCYPABCDEFXKHIJKLMNOVOVRXBCRZVGSZXBFRZXBSZTYPYQTAXLEXBXLS
+      ZYSYPUUAYQUUBYRYOVGXBXLCUTVAUUBYTYGXBXLXBXLFUTUUBVJVPVSPVTWAWBVAWCWDWEWFW
+      GWHWIWJWKWL $.
   $}
 
+
 $(
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
-                                 Algorithms
+        Rationality of square root of 2
```
benjub, line 106040: Non-rationality of the square root of 2 (?)

iset.mm
```diff
@@ -105738,6 +105738,18 @@ which defines the set we are taking the supremum of must (a) be true at
       UKAVNNWDRUPRWEVOWFVPVRVSVTWG $.
   $}
 
+  $( Partial converse to ~ bezout .  Existence of a linear combination does not
+     set the GCD, but it does upper bound it.  (Contributed by Stefan O'Rear,
```
benjub, line 105742: The GCD divides any linear combination. (?)

iset.mm
```diff
@@ -105750,6 +105750,20 @@ which defines the set we are taking the supremum of must (a) be true at
     QUSUKUNURAKLZVDVGVHVIUQVLURBKLZUMVLVMGUPABUENZUFUSUKUNPVLVDURACQRTUQUSULUOV
     MVEVGVJVKUQVLVMVNUJUSULUOPVMVEURBDQRTUSVAVCPVDVEGVFURUTVBUGRUH $.
 
+  $( Converse of ~ bezout for the gcd = 1 case, sufficient condition for
```
benjub, line 105753: ... when the GCD equals 1, ...

iset.mm
jkingdon: I'll gladly accept this section header if it leads to people not complaining about the header. The idea is to cover both non-rationality and irrationality in this section, although the latter won't be fully developed until all this number theory stuff is in place.

iset.mm
jkingdon: I'm pretty sure I like "upper bound" better. Granted it isn't an upper bound expressed quite as we usually state one, but it is an upper bound in a certain (multiplicative) sense, and makes for a better parallel to "set the GCD".

Also, this is not a place where iset.mm should diverge from set.mm.

iset.mm
jkingdon: Sounds good to me; I've pushed up a change to incorporate this wording.

jkingdon:
All comments addressed; ready for re-review.

iset.mm
```diff
@@ -105599,6 +105599,31 @@ which defines the set we are taking the supremum of must (a) be true at
       TUMSEASEVBVDLUKUMTATUMAUBQUCUOVBURUQGUNBRNUDUEUFUGUHUJ $.
   $}
 
+  ${
+    $d x y A $.  $d x y B $.  $d x y C $.
+    $( Relationship between the order of an element and that of a multiple.  (a
+       divisibility equivalent).  (Contributed by Stefan O'Rear,
```
benjub, line 105605: Hard to understand...
Maybe:
> An integer divides another integer iff it divides its GCD with that other integer (stated with an additional factor `B` in the multiples). This can be used to compute orders of iterates of elements in groups: if `A` is the order of an element `x` in a group, then the order of `x^C`, say `D`, satisfies `A = ( D x. ( C gcd  A ) )`.

benjub (approved):

benjub:
I reviewed "commit-wise" since the diff was a bit long. If the last one was only moving around, then all is good.

These are only suggestions, feel free to ignore.

From time to time, you bring an enhancement to the set.mm version, either in the proof or the comment. Don't be shy to adopt the better version and bring it back to set.mm !

iset.mm
jkingdon: > Hard to understand... Maybe:
> 
> > An integer divides another integer iff it divides its GCD with that other integer (stated with an additional factor `B` in the multiples). This can be used to compute orders of iterates of elements in groups: if `A` is the order of an element `x` in a group, then the order of `x^C`, say `D`, satisfies `A = ( D x. ( C gcd  A ) )`.

Feel free to make a follow up pull request for this. I'm afraid I don't understand this comment, or the set.mm one, or the one I put in iset.mm (a minor edit of the set.mm one), well enough to be an especially good reviewer, however.

jkingdon:
> I reviewed "commit-wise" since the diff was a bit long. If the last one was only moving around, then all is good.

Yeah, the last one was only moving around, no changes. I was a bit disappointed that the diff algorithm didn't notice that, but I guess if there are enough blank lines and lines with only braces and that sort of thing, it has a tendency to say (unhelpfully), "oh those lines are the same between new and old".

I even thought about making two separate pull requests, but I don't know, there are a lot of way to manage git and github which create one problem and solve another so I try not to obsess too much about finding the perfect strategy.

> These are only suggestions, feel free to ignore.

OK, good to know. And likewise, if you want to propose a further revision to any of them, feel free to make a follow up pull request.

> From time to time, you bring an enhancement to the set.mm version, either in the proof or the comment. Don't be shy to adopt the better version and bring it back to set.mm !

Thanks for the words of encouragement! I'll try not to feel inhibited if I see something and am able to derail/redirect my brain to set.mm long enough to figure out whether it makes sense there too.

Issue #2433, opened 2022-01-13
Update odd/even so we can move theorems from Alexander van der Vekens mathbox


jkingdon:
When last we visited the odd/even saga it was at #1682 and here is my attempt to summarize what we discussed there:
* Standard notation for even and odd, respectively, in set.mm is `2 || A` and `A e. ZZ /\ -. 2 || A`. The former implies `A e. ZZ` by http://us.metamath.org/mpeuni/dvdszrcl.html (and although this discussion isn't primarily about iset.mm, that's also present in iset.mm: http://us.metamath.org/ileuni/dvdszrcl.html )
* Many of the odd/even theorems are in the section http://us.metamath.org/mpeuni/mmtheorems151.html#mm15003s and there is a section header there which summarizes the situation. (That's the section titled "Even and odd numbers" under "Elementary properties of divisibility", starting with http://us.metamath.org/mpeuni/evenelz.html , in case sections have been renumbered by the time you read this).
* Many theorems in Alexander van der Vekens mathbox use classes `Even` and `Odd` and at least some of those we'd like to move to main.

As I understand it the choices on the mathbox are:
1. Add `Even` and `Odd` to main, mostly because `A e. Odd` is shorter than `A e. ZZ /\ -. 2 || A` (if there are contexts where `A e. ZZ` is not already present), or
2. Modify the theorems in the mathbox (especially whichever ones we want to move to main) to use the `2 || A` and `A e. ZZ /\ -. 2 || A` notation.

I'm not sure I have a strong opinion between those two. There was much discussion when this came up before and although I guess I'm sympathetic to the idea of sticking with the `2 || A` notation as a way of reducing the number of notations we introduce, the odd case isn't quite as easy as the even case as described above.

avekens:
Thierry's remarks in the Google group (https://groups.google.com/g/metamath/c/oOHiy3i-cRw/m/exLZuw-mAgAJ):

Indeed this is a good opportunity to raise (again) the question about the definition of odd integers.

I agree with Norm that` 2 || N ` is just as long as ` N e. Even `, and therefore will not bring any database size improvement. Since it also implies ` 2 e. ZZ `, I don't see worth in defining `Even`. For `Odd` this does not hold, one cannot just replace it with ` -. 2 || N ` but also has to add ` N e. ZZ ` to recover a definition of `Odd`.

[...]

PR #2435, opened 2022-01-17
Add mathbox for Peter Mazsa to set.mm

jkingdon:
This is exactly as sent to the mailing list.


jkingdon:
Here is the description as sent to the mailing list:


My purpose in set theory was to reveal the Partition-Equivalence
Theorem with general ` R `. My results are attached as a Mathbox.
Proofs and markup are verified, but I need some theorems of Thierry
Arnoux, Scott Fenton, Frédéric Liné, Rodolfo Medina and Richard
Penner, and my latexdefs, htmldefs and althtmldefs, in the main part
of set.mm. Some theorems are slight modifications of theorems of Mario
Carneiro, Rodolfo Medina and Norman Megill.

Since the attached Mathbox is quite long (600+ necessary and
sufficient definitions and theorems incl. versions), I‘ll now give you
a summary. I’d like to discuss my results here on the email-list.

I had to redefine some basic concepts (e.g. reflexivity, symmetry,
transitivity, equivalence relation and equivalence relation with
domain part, function, disjointness and partition), and suggest a new
one (converse reflexivity), in order to reach the
Partition-Equivalence Theorem.
This motivation is very important here: I’m aware of the discussion
https://github.com/metamath/set.mm/pull/1286 with the participation of
Thierry Arnoux, Benoit Jubin, Normann Megill and David A. Wheeler, and
if you find better definitions for reflexivity etc. than me, I do not
oppose them from the outset. However it would be great if
1. At least reflexivity, symmetry and transitivity had coherent
definitions (compare ~ df-refs and ~ df-syms below, as opposed to ~
issref with ` A ` vs. ~ cnvsym without ` A ` in set.mm),
2. Based on the 2. requirement of the Guidelines at
http://us.metamath.org/mpeuni/mathbox.html all concepts were defined
primarily as 0-ary class constants (like ~ df-refrels ),
3. Moreover, all class constants had a corresponding predicate for
proper classes (like ~ df-refrel for _I in ~ refrelid $p |- RefRel _I
)  so that e.g. in case of reflexive relations, the element of the
class of all reflexive relations ( ~ elrefrels2 ) and the reflexive
relation predicate ( ~ df-refrel ) were the same when ` R ` is a set:
~ elrefrelsrel $p |- ( R e. V -> ( R e. RefRels <-> RefRel R ) ) ,
4. You proved that they reproduce at least ~ mpet $p |- ( MembPart A
<-> MembEr A ) , ~ pet and ~ pets , or their (possible future)
generalization,  ` gpet ` , see below.

After the (new, suggested) definitions you’ll find the (old)
definitions which are presently in set.mm or in a Mathbox (
https://github.com/metamath/set.mm/blob/develop/set.mm ).

_
Classes of all reflexive, converse reflexive, symmetric, transitive
sets (new, suggested definitions, used only once in ~ df-refrels , ~
df-cnvrefrels , ~ df-symrels and ~ df-trrels below):
df-refs $a |- Refs = { x | ( _I i^i ( dom x X. ran x ) ) _S ( x i^i (
dom x X. ran x ) ) }
df-cnvrefs $a |- CnvRefs = { x | ( _I i^i ( dom x X. ran x ) ) `' _S (
x i^i ( dom x X. ran x ) ) } (new concept, we need it below at
functions and disjoints)
df-syms $a |- Syms = { x | `' ( x i^i ( dom x X. ran x ) ) _S ( x i^i
( dom x X. ran x ) ) }
df-trs $a |- Trs = { x | ( ( x i^i ( dom x X. ran x ) ) o. ( x i^i (
dom x X. ran x ) ) ) _S ( x i^i ( dom x X. ran x ) ) }
where ` _S ` is the subsets class, or the class of the subset
relations (new, suggested definition):
df-ssr $a |- _S = { <. x , y >. | x C_ y } , or ` df-ss `, since the
present ~ df-ss definition of the subclass relation ` A C_ B ` should
be relabeled as ` df-sc `.
Scott Fenton's subset class ~ df-sset and  ` _S ` are the same
ssreqsset $p |- _S = SSet  ,
however, the notation ` _S ` and the construction of ~ df-ssr reveals
its similarity to these old definitions in set.mm:
df-id $a |- _I = { <. x , y >. | x = y } and
df-eprel $a |- _E = { <. x , y >. | x e. y } .

Classes of all reflexive, etc. relations (new, suggested):
df-refrels $a |- RefRels = ( Refs i^i Rels )
df-cnvrefrels $a |- CnvRefRels = ( CnvRefs i^i Rels )
df-symrels $a |- SymRels = ( Syms i^i Rels )
df-trrels $a |- TrRels = ( Trs i^i Rels )
where the relations class (new, suggested) is
df-rels $a |- Rels = ~P ( _V X. _V ) .

In predicate form the above definitions give:
dfrefrel2 $p |- ( RefRel R <-> ( ( _I i^i ( dom R X. ran R ) ) C_ R /\ Rel R ) )
! dfrefrel3 $p |- ( RefRel R <-> ( A. x e. dom R A. y e. ran R ( x = y
-> x R y ) /\ Rel R ) )
dfcnvrefrel2 $p |- ( CnvRefRel R <-> ( R C_ ( _I i^i ( dom R X. ran R
) ) /\ Rel R ) )
dfcnvrefrel3 $p |- ( CnvRefRel R <-> ( A. x e. dom R A. y e. ran R ( x
R y -> x = y ) /\ Rel R ) )
dfsymrel2 $p |- ( SymRel R <-> ( `' R C_ R /\ Rel R ) )
dfsymrel3 $p |- ( SymRel R <-> ( A. x A. y ( x R y -> y R x ) /\ Rel R ) )
dftrrel2 $p |- ( TrRel R <-> ( ( R o. R ) C_ R /\ Rel R ) )
dftrrel3 $p |- ( TrRel R <-> ( A. x A. y A. z ( ( x R y /\ y R z ) ->
x R z ) /\ Rel R ) )
Note that while ` Rel R ` cancels restriction of ` R ` , it keeps
restriction of ` _I ` : this is why the very similar definitions ~
df-refs , ~ df-syms and ~ df-trs diverge when we switch from (general)
sets to relations in ~ dfrefrel* , ~ dfsymrel* and ~ dftrrel* .

Reflexivity and symmetry reproduce ` x R x ` for the reflexivity part,
but only together:
refsymrel2 $p |- ( ( RefRel R /\ SymRel R ) <-> ( ( ( _I |` dom R ) C_
R /\ `' R C_ R ) /\ Rel R ) )
! refsymrel3 $p |- ( ( RefRel R /\ SymRel R ) <-> ( ( A. x e. dom R x
R x /\ A. x A. y ( x R y -> y R x ) ) /\ Rel R ) )
Reflexivity and symmetry are often bound together in equivalence
relation. Reflexivity may be perceived primarily as a necessary part
of equivalence. This seems to be the main reason why we thought so far
that reflexivity on its own should be defined like the reflexivity
part of ~ dfeqvrel3 below, i.e., by using the apparently self-evident
` x R x ` form.
Another possible reason is that we are inclined to constrain ourselves
to convenient special cases where the two forms are equivalent, e.g.
the one with square Cartesian product:
idinxpssinxp3 $p |- ( ( _I i^i ( A X. A ) ) C_ ( R i^i ( A X. A ) )
<-> ( _I |` A ) C_ R ) and
idinxpssinxp4 $p |- ( A. x e. A A. y e. A ( x = y -> x R y ) <-> A. x
e. A x R x ) .

Old reflexivity versions in set.mm or in a Mathbox now, with ` x R x `:
df-reflexive $a |- ( R Reflexive A <-> ( R C_ ( A X. A ) /\ A. x e. A x R x ) )
issref $p |- ( ( _I |` A ) C_ R <-> A. x e. A x R x )

Equivalence relations (new, suggested for set.mm, but other math
sources seem to define it in a way which is similar to this one):
df-eqvrels $a |- EqvRels = ( ( RefRels i^i SymRels ) i^i TrRels )
which gives
dfeqvrel2 $p |- ( EqvRel R <-> ( ( ( _I |` dom R ) C_ R /\ `' R C_ R
/\ ( R o. R ) C_ R ) /\ Rel R ) )
dfeqvrel3 $p |- ( EqvRel R <-> ( ( A. x e. dom R x R x /\ A. x A. y (
x R y -> y R x ) /\ A. x A. y A. z ( ( x R y /\ y R z ) -> x R z ) )
/\ Rel R ) )
There is no comparable definition of equivalence relation now in set.mm.

Equivalence relations with domain part (new, suggested):
df-ers $a |- Ers = ( DomainQss |` EqvRels )
where df-dmqss $a |- DomainQss = { <. x , y >. | ( dom x /. x ) = y }
which gives
dferALTV2 $p |- ( R ErALTV A <-> ( EqvRel R /\ ( dom R /. R ) = A ) )
This definition implies that the natural domain of equivalence relations
is not ` R Domain A ` (i.e. ` dom R = A ` cf. ~ brdomaing ), like in
the present ~ df-er in set.mm,
but ` R DomainQss A ` (i.e. ` ( dom R /. R ) = A ` , cf. ~ brdmqss ),
"domain quotients".
The main reason why I think the new ` R DomainQss A ` to be the
natural domain (and not the old ` R Domain A ` ) is a small set of new
theorems, the Membership Partition-Equivalence Theorems (see mpet*
below) and the Partition-Equivalence Theorems (see pet* below).
Incidentally, with the new definitions above, the old ~ prter3 in set.mm
prter3 $p |- ( ( R Er U. A /\ ( U. A /. R ) = ( A \ { (/) } ) ) -> .~ = R )
simplifies to the new
erALTVimlem $p |- ( R e. V -> ( R ErALTV A -> ~ A = R ) )
where the coelements ` ~ A ` are a special case of cosets ` ,~ R ` :
df-coss $a |- ,~ R = { <. x , y >. | E. u ( u R x /\ u R y ) }
df-coels $a |- ~ A = ,~ ( `' _E |` A )
dfcoels $p |- ~ A = { <. x , y >. | E. u e. A ( x e. u /\ y e. u ) }
While I'm sure we need both equivalence relation ~ df-eqvrels and
equivalence relation on domain quotient ~ df-ers , I'm not sure
whether we need a third equivalence relation concept with the present
` dom R = A ` component as well: this needs further investigation.
Using Occam’s razor I suppose that these two concepts ~ df-eqvrels and
~ df-ers are enough and named the predicate version of the one on
domain quotient as the alternate version ~ df-erALTV of the present ~
df-er (cf. "Alternative (ALTV) versions",
http://us.metamath.org/mpeuni/conventions.html ).

The old definition of equivalence relation (with a domain part) in set.mm
df-er $a |- ( R Er A <-> ( Rel R /\ dom R = A /\ ( `' R u. ( R o. R ) ) C_ R ) )
"is not standard" , "somewhat cryptic", has no corresponding costant
0-ary class, does not follow the traditional transparent
reflexive-symmetric-transitive relation way of definition of
equivalence, and fails to deliver the  Membership
Partition-Equivalence Theorems and the Partition-Equivalence Theorems.

With the new definitions above comes the
Main Theorem of Equivalences: any equivalence relation implies
equivalent membership as well
mainer $p |- ( R ErALTV A -> MembEr A )
where
df-member $a |- ( MembEr A <-> ~ A ErALTV A )
dfmember3 $p |- ( MembEr A <-> ( EqvRel ~ A /\ ( U. A /. ~ A ) = A ) )

Now we should define functions and disjoints (new, suggested):
df-funss $a |- Funss = { x | ,~ x e. CnvRefRels } , used only once in
df-funsALTV below
df-funsALTV $a |- FunsALTV = ( Funss i^i Rels )
dffunALTV2 $p |- ( FunALTV F <-> ( ,~ F C_ _I /\ Rel F ) )
The new definition of the function predicate ~ df-funALTV (based on a
more general, converse reflexive, relation) and the old definition of
function in set.mm ~ df-fun , are the same:
funALTVfun $p |- ( FunALTV F <-> Fun F )

df-disjss $a |- Disjss = { x | ,~ `' x e. CnvRefRels } , used only
once in df-disjs below
df-disjs $a |- Disjs = ( Disjss i^i Rels )
We need ` Disjs ` and ` Disj R ` for the definition of ` Parts ` and `
R Part A ` for the Partition-Equivalence Theorems: this need for `
Parts ` as disjoint relations on their domain quotients is the reason
why we should define ` Disjs ` instead of simply using converse
functions, cf. ~ pet2 below.
dfdisjALTV2 $p |- ( Disj R <-> ( ,~ `' R C_ _I /\ Rel R ) ) , (read: `
R ` is a disjoint)
dfdisjALTV4 $p |- ( Disj R <-> ( A. x E* u u R x /\ Rel R ) )
dfdisjALTV5 $p |- ( Disj R <-> ( A. u e. dom R A. v e. dom R ( u = v
\/ ( [ u ] R i^i [ v ] R ) = (/) ) /\ Rel R ) )
The old definition of disjointness in set.mm is
df-disj $a |- ( Disj_ x e. A B <-> A. y E* x e. A y e. B )

The disjoint elementhood predicate (read: the elements of ` A ` are
disjoint) is a special case of disjointness (new, suggested):
df-eldisj $a |- ( ElDisj A <-> Disj ( `' _E |` A ) )
dfeldisj5 $p |- ( ElDisj A <-> A. u e. A A. v e. A ( u = v \/ ( u i^i
v ) = (/) ) )
As of now, disjoint elementhood is defined as "partition" in set.mm :
df-prt $a |- ( Prt A <-> A. u e. A A. v e. A ( u = v \/ ( u i^i v ) =
(/) ) ) (cf. the new partition definitions below)
Disjoint elementhood changes
prter1 $p |- ( Prt A -> .~ Er U. A ) of the present set.mm to
mdisjim $p |- ( ElDisj A -> EqvRel ~ A )
and in general form, the
 "Divide et Aequivalere" Theorem (new): every disjoint relation
generates equivalent cosets by the relation
disjim $p |- ( Disj R -> EqvRel ,~ R )
Moreover, this changes
prter2 $p |- ( Prt A -> ( U. A /. .~ ) = ( A \ { (/) } ) ) of the
present set.mm to
eldisjn0el $p |- ( ElDisj A -> ( -. (/) e. A <-> ( U. A /. ~ A ) = A ) )
and in general form (new):
disjdmqseq $p |- ( Disj R -> ( ( dom R /. R ) = A <-> ( dom ,~ R /. ,~
R ) = A ) )

Partitions (new, suggested) are disjoints on domain quotients (or:
domain quotients restricted to disjoints):
df-parts $a |- Parts = ( DomainQss |` Disjs ) $.
dfpart2 $p |- ( R Part A <-> ( Disj R /\ ( dom R /. R ) = A ) )
The above definition is a more general meaning of partition than we
accustomed to: the conventional meaning of partition (e.g. partition
` A ` of  ` X `, [Halmos] p. 28: "A partition of ` X ` is a disjoint
collection ` A ` of non-empty subsets of ` X ` whose union is ` X ` ",
or Definition 35, [Suppes] p. 83., cf. https://oeis.org/A000110 ) is
what we call here membership partition:
dfmembpart2 $p |- ( MembPart A <-> ( ElDisj A /\ -. (/) e. A ) ) , defined as
df-membpart $a |- ( MembPart A <-> ( `' _E |` A ) Part A )

A (somewhat maybe depressing, new) theorem with the membership
partition concept above is the
Theorem of Fences by Equivalences: all kinds of equivalence relations
imaginable (in addition to the membership equivalence relation cf. ~
mpet ) generate partition of the members:
fences $p |- ( R ErALTV A -> MembPart A )

Now we are ready to prove the (new) Membership Partition-Equivalence Theorems:
mpet3 $p |- ( ( ElDisj A /\ -. (/) e. A ) <-> ( EqvRel ~ A /\ ( U. A
/. ~ A ) = A ) )
mpet2 $p |- ( ( `' _E |` A ) Part A <-> ,~ ( `' _E |` A ) ErALTV A )
mpet $p |- ( MembPart A <-> MembEr A )

and the (new) Partition-Equivalence Theorems, with general ` R `:
pet2 $p |- ( ( Disj ( R (x) ( `' _E |` A ) ) /\ ( dom ( R (x) ( `' _E
|` A ) ) /. ( R (x) ( `' _E |` A ) ) ) = A ) <-> ( EqvRel ,~ ( R (x) (
`' _E |` A ) ) /\ ( dom ,~ ( R (x) ( `' _E |` A ) ) /. ,~ ( R (x) ( `'
_E |` A ) ) ) = A ) )
pet $p |- ( ( R (x) ( `' _E |` A ) ) Part A <-> ,~ ( R (x) ( `' _E |`
A ) ) ErALTV A )
pets $p |- ( ( A e. V /\ R e. W ) ->
  ( ( R (x) ( `' _E |` A ) ) Parts A <-> ,~ ( R (x) ( `' _E |` A ) ) Ers A ) )
where
br1cosstxpcnvepres $p |- ( ( ( B e. V /\ C e. W ) /\ ( D e. X /\ E e. Y ) ) ->
  ( <. B , C >. ,~ ( R (x) ( `' _E |` A ) ) <. D , E >. <->
    E. u e. A ( ( C e. u /\ u R B ) /\ ( E e. u /\ u R D ) ) ) )

This small set of theorems is the main result of my investigations in
set theory. ~ pet is not more general than the conventional Membership
Partition-Equivalence Theorem ~ mpet2 (because you cannot set ` R ` in
~ pet in a way that you would get ~ mpet2 ), i.e., this is not the
hypothetical General Partition-Equivalence Theorem ` gpet |- ( R Part
A <-> ,~ R ErALTV A ) `. But ~ pet has a general part that ~ mpet2
lacks: ` R `, which is well enough for my future application of set
theory, for my purpose outside of set theory. Motto: "Le meglio è
l'inimico del bene.":)

avekens:
What shall we do with this PR? Besides the errors, the main issue is the strong dependency on Scott Fenton's mathbox. I don't think that it would be a good idea to move all these theorems of Scott's mathbox to the main body of set.mm (the few theorems of Thierry's mathbox which are references may be moved to main). It is maybe the same kind of issue as described in #2433, for which a joint mathbox (for Scott and Peter) may be a solution.

digama0:
There are multiple issues with this PR, and it is our responsibility as reviewers to point out issues, not fix them. Unless there are extenuating circumstances, every author is responsible for championing their own contributions through the review process. Mathlib has many abandoned PRs that sit open for a long time because their authors were not interested to finish the job. Sometimes these are closed with a "please-adopt" label if there is no movement for, say, a year. But I want to stress that it is not our job to do an author's work for them.

sctfn:
Personally, I have no issues moving these theorems to main. They're a bit esoteric, but they're perfectly valid.

jkingdon:
> every author is responsible for championing their own contributions through the review process

It is a fair point and I do expect some level of participation from Mázsa Péter (who doesn't seem to have yet made a github account as far as I can tell). However, he has championed this as far as sending it to the mailing list with a fairly detailed explanation, and has also reacted there to the creation of this pull request:

```
Thank you for your very kind mail and help! For now it is optimized
only for ` verify proof * ` and ` verify markup */… ` of Metamath (and
you are right: I need quite a lot of theorems from the Mathboxes of
other community members in the main part of set.mm, these are the
markup failures), I just wanted to have some gut feelings and the
widest possible collective wisdom about the whole enterprise at
first:)
```

Since he is a new contributor, I would like to see us do what we can to meet halfway, explaining how we do things and listening to reactions. In particular, he has asked (in the quote above) for an answer on what to do about the theorems from other mathboxes and he can't solve that on his own without some engagement from the rest of us.

jkingdon:
> these theorems of Scott's mathbox to the main body of set.mm (the few theorems of Thierry's mathbox which are references may be moved to main).

Do we have a list of these theorems? (Either the @sctfn ones which are in his words "a bit esoteric, but  . . . perfectly valid" or the @tirix ones which I guess from the above are smaller in number and less esoteric).

avekens:
> Do we have a list of these theorems? (Either the @sctfn ones which are in his words "a bit esoteric, but . . . perfectly valid" or the @tirix ones which I guess from the above are smaller in number and less esoteric).

Here are the lists of used theorems of other's mathboxes, extracted from the details of the check "Verify using metamath.exe (original C verifier by Norman Megill)":

- "idssxp" , "eqvincg" in the mathbox for Thierry Arnoux.
- "brv", "dfres3" , "ctxp", "df-txp", "txpss3v", "txprel", "csset", "df-sset" in the mathbox for Scott Fenton.

jkingdon:
> Here are the lists of used theorems of other's mathboxes, extracted from the details of the check "Verify using metamath.exe (original C verifier by Norman Megill)":

Ah, thanks. Perhaps this was easier to find in the output than I realized.

>     * "idssxp" , "eqvincg" in the mathbox for Thierry Arnoux.

These strike me as unobjectionable if anyone has a use for them.

>     * "brv", "dfres3" , "ctxp", "df-txp", "txpss3v", "txprel", "csset", "df-sset" in the mathbox for Scott Fenton.

Hmm, I see why "esoteric" could be a fitting description. This introduces two new notations (tail cross and a version of subset which differs from the existing one on proper classes).

I guess I don't have an especially strong reaction one way or the other although perhaps I would if I knew this material better. I also don't have a really good feel for how we prefer to handle topics which seem unlikely to be used widely (if this is indeed one of those). Harmless? Clutter? Best kept to mathboxes (although that sorta breaks down if multiple people want to play with them, as appears to be the case here)?

avekens:
> * "idssxp" , "eqvincg" in the mathbox for Thierry Arnoux.
> * "brv", "dfres3" , "ctxp", "df-txp", "txpss3v", "txprel", "csset", "df-sset" in the mathbox for Scott Fenton.

Having a deeper look at these theorems and definitions, I would propose the following:
* "idssxp" , "eqvincg", "brv", "dfres3" can be moved to main set.mm without further discussion (an experienced contributor should do this, to find the right places for them).
* df-txp (Define the tail cross of two classes) and the related txpss3v, txprel: This may be useful in general - to be discussed, if df-txp (together with all/some related theorems) should be moved to main. In my opinion, nothing speaks against it.
*  df-sset (Define the subset class): Is this really useful? Why is ` C_ ` not sufficient? I would suggest not to move this to main.

tirix:
> * df-sset (Define the subset class): Is this really useful? Why is `C_` not sufficient? I would suggest not to move this to main.

it seems that the only theorem where df-sset is referenced is `ssreqsset`, which simply states that Péter's `_S` is exactly the same as Scott's `SSet`:
```
  $( Subset relation class and Scott Fenton's subset class ~ df-sset are the
     same.  (Contributed by Peter Mazsa, 9-Aug-2021.) $)
  ssreqsset $p |- _S = SSet $=
    ( cssr cvv cxp cep cdif ctxp crn csset dfssr2 df-sset eqtr4i ) ABBCDBDEFGEH
    IJK $.
```
The problem could be solved by commenting out just this short theorem.

If at any time in the future we would move either Scott's `Sset` or Péter's `_S` to main, we probably would only move one of them and replace the other one by that new official one. Though, I agree I don't think we need it now.

We do have `_I` and `_E` in the main part of set.mm, which prove to be useful to write closed forms of complex expressions. Maybe `_S` might also help for that in some cases (I have not checked).

benjub:
@tirix : could you do this in a separate PR :
idssxp : move to Main (right after fnresi?).  Run min.cmd to "minimize_with idssxp" all theorems using fnresi.
equvincg : move right after/before equvinc, equvincf.  Minimize one with the other (find which ordering makes things smaller). Maybe do the same thing with "eqvincg" with regards to eqvinc/eqvincf ?  Run min.cmd to "minimize_with eq\*vin*" all theorems using eq\*vin*. 

I agree with @avekens that brv and dfres3 can also be moved to Main now (i.e., regardless of other changes).

df-txp: I had never heard of the "tail cross"... The "variable free" definitions it permits are interesting, but I'm not sure the range of application justifies its moving to Main.

df-sset: I agree with @tirix.  Cases where we need both versions ("=" and "_I" , "e." and "_E") are rare. Arguably, subsets are the third most important case, and as above this permits to write "variable free" definitions, but I'm not sure its applications justify its moving to main.  In any case, I would change in both (sub)label and token, sset to subset (since "sset" commonly means "simplicial set").

mazsa:
Hi Community, thank you for your efforts again.
I inspected the logs of the not successful checks and I found that all of it may come from 4 sources:
1. These 9 items in 4 Mathboxes: 
(!) Rodolfo Medina`s n0el , eqbrrdv2 , 2r19.29 , jca2 , 
(!) Richard Penner´s dfss6 ,
Scott Fenton´s brv , dfres3 (mentioned above)
Thierry´s idssxp , eqvincg (mentioned above)
I think nobody is against these theorems for the main set.mm.
2. Scott Fenton´s further 4 tail Cartesian product items : ctxp , df-txp , txprel and txpss3v are necessary for my pet* results, and if possible I´d prefer them in the main set.mm.
3. @tirix suggested that we should comment out my ssreqsset and that will solve the problem of these 2 items in Scott Fenton`s Mathbox: csset and df-sset. I find it a good solution.
4. My necessary latexdefs, htmldefs and althtmldefs: they are commented out under “Notations (incl symbol definitions)” in my Mathbox.
After that I expect most checks, if not all, to be successful.
 
The class of subsets df-ssr $a |- _S = { <. x , y >. | x C_ y } is not indispensable but I like it. I mentioned that it explicates the “converse” part in the definition of converse reflexive relation as opposed to the reflexive relation, cf. df-refs vs df-cnvrefs. Moreover, if you define extensional relation as e.g. as dfextrel* |- ( ExtRel R <-> ( A. x A. y ( [ x ] `' R = [ y ] `' R <-> x = y ) /\ Rel R ) ) you’ll find not only that extrelep |- ExtRel _E and extrelid |- ExtRel _I but extrelssr |- ExtRel _S as well. This is outside of my scope (so I did not define it), but it shows that _S may have further interesting properties / use cases. 
 
Naming: I think the most consistent labels would be
present: df-ss $a |- ( A C_ B <-> ( A i^i B ) = A ) suggested: df-sc (subclass)
present: df-ssr $a |- _S = { <. x , y >. | x C_ y } suggested: df-ss (subset)
present: df-eprel $a |- _E = { <. x , y >. | x e. y } suggested: df-ep (epsilon)
present: df-id $a |- _I = { <. x , y >. | x = y } suggested: df-id (identity)
But keeping the old labels is ok as well.
 
As you can see I’m not native at Github:) I`m trying to clone set.mm (it's coming down with 80KiB/s via SSH) in order to create my first pull requests tomorrow.

jkingdon:
> As you can see I’m not native at Github:) I`m trying to clone set.mm (it's coming down with 80KiB/s via SSH) in order to create my first pull requests tomorrow.

Keep asking if there is anything we can help with. That first clone can take a very long time (partly due to the large files we use, I suspect) but after that I found things got faster (I probably also had to run `git gc` although I don't remember for sure whether I did that or whether git did whatever it had to do automatically).

benjub:
This PR is very large, with disproportionately many definitions, and we'll have at least to divide it into several chunks.

>     1. These 9 items in 4 Mathboxes:
>        (!) Rodolfo Medina`s n0el , eqbrrdv2 , 2r19.29 , jca2 ,
>        (!) Richard Penner´s dfss6 ,
>        Scott Fenton´s brv , dfres3 (mentioned above)
>        Thierry´s idssxp , eqvincg (mentioned above)
>        I think nobody is against these theorems for the main set.mm.

No: only brv, dfres3, idssxp, eqvincg have been discussed (and agreed to be moved to Main, with some changes given above by @avekens, @tirix and I).  I haven't looked at the other theorems yet, but for instance eqbrrdv2 doesn't use the conventional labeling and statement for deduction forms.

The standards are not the same for mathbox theorems and theorems in the main part, so it will take some work to check the various theorems that would have to be moved.

New definitions should be added only when they really save a lot of rephrasings.  Although I haven't looked at the PR in detail, the amount of new definitions is suspicious.

Can you tell in plain English what the partition-equivalence theorem is ?  Google gives no result...

>     4. My necessary latexdefs, htmldefs and althtmldefs: they are commented out under “Notations (incl symbol definitions)” in my Mathbox.
>        After that I expect most checks, if not all, to be successful.

There is a place in set.mm to put these $t definitions: between comments of the form:
/* Mathbox of Thierry Arnoux */
....
/* End of Thierry Arnoux's mathbox */
(use alphabetical order of the contributor).


What is a "converse-reflexive" relation ?

> As you can see I’m not native at Github:) I`m trying to clone set.mm (it's coming down with 80KiB/s via SSH) in order to create my first pull requests tomorrow.

Begin with one small PR.

tirix:
> I`m trying to clone set.mm (it's coming down with 80KiB/s via SSH) in order to create my first pull requests tomorrow.

This command:
```
git clone https://github.com/metamath/set.mm.git –-depth 1
```
Performs a "shallow" clone, with only the last status and not the full history of set.mm. It shall greatly speed up the initial clone.

mazsa:
Benoit @benjub of course I'll start with one small PR (just finished the cloning process), I wouldn't dare do anything else:) I'm quite lost in my half-baked mental model of git, fortunately Jim sent me this https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing . I´ll try my best.
 
Math answers:
The (hypothetical) General Partition-Equivalence Theorem ( gpet ) would be the theorem which states the equivalence of 2 items:
1. ` A ` is a partition by ` R ` : ` R Part A `
(or: ` R ` is a disjoint on its domain quotient ` A ` : ` dfpart2 $p |- ( R Part A <-> ( Disj R /\ ( dom R /. R ) = A ) )  )`, 
where ``df-disjALTV $a |- ( Disj R <-> ( CnvRefRel ,~ `' R /\ Rel R ) ) ``
where `` |- ( CnvRefRel ,~ `' R <-> ,~ `' R C_ _I ) )`` (cf. `cnvrefrelcoss2` ) and 
where ``cosscnvssid5 $p |- ( ( ,~ `' R C_ _I /\ Rel R ) <-> ( A. u e. dom R A. v e. dom R ( u = v \/ ( [ u ] R i^i [ v ] R ) = (/) ) /\ Rel R ) )``,
2. ` ,~ R ` is an equivalence relation on its domain quotient ` A ` : `,~ R ErALTV A `,
where `|- ( ,~ R ErALTV A <-> ( EqvRel ,~ R /\ ( dom ,~ R /. ,~ R ) = A ) )` (cf. `dferALTV2`),
where `|- ( EqvRel ,~ R <-> ( RefRel ,~ R /\ SymRel ,~ R /\ TrRel ,~ R ) )` ( cf. `df-eqvrel` ),
where `df-coss $a |- ,~ R = { <. x , y >. | E. u ( u R x /\ u R y ) }` .
 
The Partition-Equivalence Theorem ( `~ pet` ) is the (proved) theorem which states the equivalence of 2 items:
1. ` A ` is a partition by `` ( R (x) ( `' _E |` A ) ) `` : ``( R (x) ( `' _E |` A ) ) Part A ``
2. `` ,~ ( R (x) ( `' _E |` A ) ) `` is an equivalence relation on its domain quotient ` A ` : ``,~ ( R (x) ( `' _E |` A ) )  ErALTV A ``,
``pet $p |- ( ( R (x) ( `' _E |` A ) ) Part A <-> ,~ ( R (x) ( `' _E |` A ) ) ErALTV A )``
or ``pet2 $p |- ( ( Disj ( R (x) ( `' _E |` A ) ) /\ ( dom ( R (x) ( `' _E |` A ) ) /. ( R (x) ( `' _E |` A ) ) ) = A ) <-> ( EqvRel ,~ ( R (x) ( `' _E |` A ) ) /\ ( dom ,~ ( R (x) ( `' _E |` A ) ) /. ,~ ( R (x) ( `' _E |` A ) ) ) = A ) ) ``
where ``br1cosstxpcnvepres $p |- ( ( ( B e. V /\ C e. W ) /\ ( D e. X /\ E e. Y ) ) -> ( <. B , C >. ,~ ( R (x) ( `' _E |` A ) ) <. D , E >. <-> E. u e. A ( ( C e. u /\ u R B ) /\ ( E e. u /\ u R D ) ) ) )``
 
The Membership Partition-Equivalence Theorem ( `~ mpet` ) is the (proved) theorem which states the equivalence of 2 items:
1. ` A ` is a membership partition (or: ` A ` is a partition by `` ( `' _E |` A ) `` )
2. a membership equivalence relation on ` A ` (or: `` ,~ ( `' _E |` A ) `` is an equivalence relation on its domain quotient ` A ` )
``mpet2 $p |- ( ( `' _E |` A ) Part A <-> ,~ ( `' _E |` A ) ErALTV A )``
or with other notation defined in the Mathbox:: 
``mpet3 $p |- ( ( ElDisj A /\ -. (/) e. A ) <-> ( EqvRel ~ A /\ ( U. A /. ~ A ) = A ) )``
``mpet $p |- ( MembPart A <-> MembEr A )``
 
Almost all of the above are existing concepts, with existing references written by mathematicians. For example, the conventional meaning of partition (e.g. partition  ` A ` of  ` X `, [Halmos] p. 28: "A partition of ` X ` is a disjoint collection ` A ` of non-empty subsets of ` X ` whose union is ` X ` ", or Definition 35, [Suppes] p. 83., cf. https://oeis.org/A000110 ) is what we call here membership partition:
`dfmembpart2 $p |- ( MembPart A <-> ( ElDisj A /\ -. (/) e. A ) ) `, defined as 
``df-membpart $a |- ( MembPart A <-> ( `' _E |` A ) Part A ) ``.
I just revealed that it is possible to define them in a general way. The most important theorems ( `pet*` ) are, alas, not general (because you cannot set ` R ` in ` ~ pet ` in a way that you would get `~ mpet2` ), but at least they make it possible to use general ` R ` in them.  E.g. `~ pet ` has a general part that `~ mpet2` lacks: ` R `, which is fortunately good enough for my future use case.
 
Converse reflexivity is the only concept which, as I now, was not revealed by mathematicians yet, or, at least, not under this name (if you search for “converse reflexive” you will find some shoes:). It is in mathematics, well hidden by the conventional definition of equivalence relation, i.e. it was not invented by me, I just helped it to emerge from the cave to the sunshine:) This concept was the main reason why I used ` _S ` and not ` C_ ` in the definitions of the components of equivalence relation. The difference is in the surprising definition of reflexivity. I think the conventional definition of reflexivity with the ` x R x ` part is the result of that that we use to meet with reflexivity bound to symmetry in equivalence relation. I suggest to define reflexivity as 
```
df-refs $a |- Refs = { x | ( _I i^i ( dom x X. ran x ) ) _S ( x i^i ( dom x X. ran x ) ) } ,
df-syms $a |- Syms = { x | `' ( x i^i ( dom x X. ran x ) ) _S ( x i^i ( dom x X. ran x ) ) } ,
df-trs $a |- Trs = { x | ( ( x i^i ( dom x X. ran x ) ) o. ( x i^i ( dom x X. ran x ) ) ) _S ( x i^i ( dom x X. ran x ) ) } ,
```
which, together with
```
df-eqvrels $a |- EqvRels = ( ( RefRels i^i SymRels ) i^i TrRels ) etc.
```
reproduces the conventional  
```
dfeqvrel3 $p |- ( EqvRel R <-> ( ( A. x e. dom R x R x /\ A. x A. y ( x R y -> y R x ) /\ A. x A. y A. z ( ( x R y /\ y R z ) -> x R z ) ) /\ Rel R ) ) 
```
with the usual ` x R x ` part.
At the same time, it reveals a new concept, the converse reflexivity as well, defined as the class of the converses of the reflexive sets: 
```
df-cnvrefs $a |- CnvRefs = { x | ( _I i^i ( dom x X. ran x ) ) `' _S ( x i^i ( dom x X. ran x ) ) } .
```
“Converse reflexivity” can only exist if the “standalone” definition of reflexivity is revealed (as opposed to the ` x R x ` type of definition “mixed” with symmetry).
We use it to define two other conventional concepts, disjoints and functions:
```
df-disjss $a |- Disjss = { x | ,~ `' x e. CnvRefRels } , df-disjs $a |- Disjs = ( Disjss i^i Rels )
df-funss $a |- Funss = { x | ,~ x e. CnvRefRels }, df-funsALTV $a |- FunsALTV = ( Funss i^i Rels )
funALTVfun $p |- ( FunALTV F <-> Fun F )
```

benjub:
I confess I only skimmed through your answer, which is not very easy to parse.  There are really two distinct things: the mathematics on the one hand, and the way they are encoded in set.mm on the other hand.  Before getting lost in the details of the encoding, which may be intricate, can you state the partition equivalence theorem, and converse reflexivity, in plain English, without excessive formalization, even if it's not 100% precise nor rigorous ? Here are examples of what I mean:
* a **partition** of a set X is a disjoint collection of nonempty sets whose union is X
* a binary relation R on X is **reflexive** if xRx for all x in X
* a binary relation R on X is **symmetric** if xRy implies yRx for all x and y in X
* **Intermediate value theorem**: given a continuous function f : [a, b] --> RR, if c is between f(a) and f(b), then there exists x in [a, b] such that f(x)=c
* **Partition equivalence theorem**: .......

mazsa:
@benjub thank you for the opportunity: it is extremely important for my use case to be able to explain them in plain English as well:) First approach:
A relation is reflexive iff: for all elements on its domain and range, if an element of its domain is the same as an element of its range, then there is the relation between them. ( ~ dfrefrel3 ) 
A relation is converse/ly reflexive iff: for all elements on its domain and range, if for an element of its domain and for an element of its range there is the relation between them, then the two elements are the same. ( ~ dfcnvrefrel3 ) 
A class is a partition by R iff: R is a disjoint and the “domain quotient” (dom R /. R ) is the class. ( ~  dfpart2 )
A class is a membership partition iff: the class is a partition by the converse epsilon relation restricted to the class ( ~ df-membpart ), i.e. the elements of the class are disjoint and the empty set is not an element of the class ( ~ dfmembpart2 ).
A relation R is a disjoint iff: for any two elements on the domain of the relation, they are the same or the intersection of the R-coset of the first one and of the R-coset of the second one is empty. ( ~ dfdisjALTV5 , but the main definition is based on the converse reflexivity concept, cf. ~ df-disjALTV )
The elements of a class are disjoint iff: the converse epsilon relation restricted to the class is a disjoint  ( ~ df-eldisj ), i.e., for any two elements of the class they are the same or their intersection is empty. ( ~ dfeldisj5 )
A relation is an equivalence relation iff: it is a reflexive symmetric transitive relation. ( ~ df-eqvrel )
A relation is an equivalence relation on a (domain quotient) class iff: it is an equivalence relation and its domain quotient is the class. ( ~ dferALT2 should be ~ dferALTV2 )
Membership equivalence relation on a class: the coelements class on the class is an equivalence relation on the class. ( ~ df-member )
x and y are cosets by R iff there exists a set u such that both u R x and u R y hold, i.e., both x and y are elements of the R-coset of u .
Coelements on a class: cosets by the converse epsilon relation restricted to the class. ( ~ df-coels )
The (hypothetical) General Partition-Equivalence Theorem: a class is a partition by a relation iff the cosets by the relation is an equivalence relation on the class.
Partition-Equivalence Theorem: a class is a partition on a class by the tail Cartesian product with the converse epsilon relation restricted to the class iff the cosets by that relation is an equivalence relation on the class. ( ~ pet )
Membership Partition-Equivalence Theorem: the membership partition class and the membership equivalence relation on the class are the same. ( ~ mpet )

benjub:
Let's hear some other opinions.  I will be bold in the following, for the sake of clarity, but there is of course no value judgment. I do not believe a theorem about binary relations needs that many nonstandard definitions.  You have to use common notions and definitions, and not reformulate everything in your own idiosyncratic jargon.  I proposed a definition of "reflexive" above but you had to complexify it in a useless way.  Forget about the difference between sets and classes, which is irrelevant here, and about domain/range.  It looks like you can state everything by positing once and for all that you have a binary relation R on a set X, and go from there.  From what I understand, a "converse reflexive relation" is one such that xRy implies x=y.  If I had to give a name, I would choose "discrete relation", but it has no standard name because it amounts to specifying a subset of X.

jkingdon:
> I do not believe a theorem about binary relations needs that many nonstandard definitions.

I think my reaction is that we can be a lot more tolerant of experiments, even if we later decide they are over complications, in mathboxes.

I mean I have a bit of the same reaction in the sense of "oh I'd need to dig into a bunch of things to even understand this". But whether these are interesting theorems or whether this is "writing these is how I learned set theory" or anything in between, I don't want to lose sight of the learning part.

I mean I suppose there might be some level of "experiment with that on your own git fork, not in a mathbox" but I think we mostly reserve that for things that won't pass the checks or won't neatly confine themselves to one mathbox or the like. And the dependence on theorems currently in other mathboxes seems pretty normal here - those pieces seem either uncontroversial or at least things which occurred to at least one other contributor.

benjub:
> I think my reaction is that we can be a lot more tolerant of experiments

I don't think I'm being intolerant: I have not opposed the PR, and I asked for other opinions.  I'll note that for the moment, I seem to be the only one who has tried to understand the math in this PR.

jkingdon:
> I don't think I'm being intolerant: I have not opposed the PR, and I asked for other opinions. I'll note that for the moment, I seem to be the only one who has tried to understand the math in this PR.

I hope I didn't make it sound like it was a bad thing to understand the math!

I mean, sure I'm trying to promote some version of https://en.wikipedia.org/wiki/Wikipedia:Please_do_not_bite_the_newcomers as best as I can (and perhaps I'm overdoing it in terms of worrying about problems which might not even exist, because I don't think anyone has been hostile) but that doesn't mean that we need to change our various checks or policies about mathboxes or anything else that exists for some good reason. And there's nothing wrong with pointing out definitions which can be expressed much more simply with more standard notation and that sort of thing. I mean, the whole reason we do this as a collaborative project is to learn from each other and improve each other's work (at least, that's why *I* want my own work to be critiqued and revised and suggested upon).

avekens:
To solve the deadlock in this PR, I would propose the following, which was almost also suggested by Peter himself already yesterday.

* The following theorems can (and should) be moved to main (@tirix could you do this, also considering @benjub remarks in a previous comment)?
** move Rodolfo Medina`s n0el , eqbrrdv2 (after eqbrrdiv - maybe renamed "eqbrriv2"?) , 2r19.29 , jca2 (after jcad)
** Richard Penner´s dfss6
** Scott Fenton´s brv , dfres3 
** Thierry´s idssxp , eqvincg
* The definition df-txp (and maybe df-sset, if it is not commented out) and related theorems should be duplicated with a hint in the comment that it is a duplicate. A Github issue should be opened for a discussion about moving it to main later.
* The necessary latexdefs, htmldefs and althtmldefs should be moved to the right place, as indicated by @benjub in a previous comment.

I agree with @jkingdon that a mathbox is a place for studies and experiments, and everybody can do anything within her/his mathbox, as long as the general rules are not violated. On the other hand, I also think it is important to review new material in mathboxes and to give hints, ask questions and give one's opinion, as @benjub did - I always appreciate his (and other's) remarks on my contributions, regardless they are in main or in my mathbox.

tirix:
> @tirix could you do this?

Sure, I'll try to do that tomorrow. 
I'll check minimizing; I've not been doing a lot of that so I don't want to make any promise there, but I believe this can always be done separately.

Generally I think there has been much more eclectic, exploratory or experimental material in mathboxes and we shall continue to be open to what gets in there. Understanding Metamath and writing one's own proof is often hard enough, I don't think that as a community, we shall make it even harder to actually commit new contributions!

tirix:
The preparation PR with theorems moved to the main part of set.mm is #2445.
Once #2445 has been approved and merged into `set.mm`, this PR can be rebased.

mazsa:
Thank you all. Then, based on the comment of @avekens point 2. above, if I understand it well, I’ll make a new version with
1. introducing temporary items ` $c (X. $. ` (with temporary html althtml and  latexdef) , ctxpTEMP , df-txpTEMP , txprelTEMP and txpss3vTEMP , with references to @sctfn , then change all occurrences of Sott’s items in my Mathbox to these TEMP items
2. change eqbrrdv2 to  eqrelrdv2 in my Mathbox
3. commenting out my ssreqsset with Scott’s csset and df-sset
4. putting my html althtmls and  latexdefs (incl. of the new ` $c (X. $. ` ) in their place (they should be in the main at the same time when my c and w statements are in my Mathbox).
Then, after the PR of @tirix will be merged, I’ll check this new version the new version above locally with metamath.exe 
and leave a comment here when I´m ready.

benjub:
@mazsa : I suggest you first do the "toy PR" suggested in #2441, for practice.  And then, do smaller PRs to move theorems by chunks.

benjub:
@mazsa : now that #2446 is merged, can you move in this "preparatory theorems" section all your theorems which do not use new definitions ?

mazsa:
Benoit @benjub , I have a question before I prepare for this partial PR: you mentioned in https://github.com/metamath/set.mm/pull/2441#issuecomment-1018596933 that I should prove anbi1ci ”with reference only to theorems in Main (that is, no bian)”. Do you think I should remove all references to bian and biand in my Mathbox?

jkingdon:
Now that @mazsa is on github and has figured out how to submit pull requests for himself (yay :tada: #2446 ), I don't see any particular reason to keep this pull request open. Feel free to move any issues still outstanding to a github issue (or re-open this pull request if you really must, but since a lot of the issues discussed here have been resolved, perhaps a new issue or new pull requests would be clearer).

benjub:
> Benoit @benjub , I have a question before I prepare for this partial PR: you mentioned in [#2441 (comment)](https://github.com/metamath/set.mm/pull/2441#issuecomment-1018596933) that I should prove anbi1ci ”with reference only to theorems in Main (that is, no bian)”. Do you think I should remove all references to bian and biand in my Mathbox?

@mazsa I'm not sure.  If they are useful, keep them, but I feel they do not add much to existing theorems.  Have you tried the command "minimize_with *" on your proofs ?

mazsa:
@benjub ok, I´ll delete them, tomorrow I´ll push the part you suggested to my mathbox (if you find similar theorems later on, please be explicit:). I try this "minimize with" as well.

benjub:
@mazsa Sorry, I was not clear. Maybe try "minimize_with *" on your proofs using bian and biand. If this removes all the uses of bian or biand, then remove it/them.  If not, then keep them.

mazsa:
Benoit @benjub , this is what I did, please review whether this is what you suggested:
 
(anbi1ci, former anbi2ri, with the removed bian:)
MM> prove anbi1ci
(...) The proof you are starting with is already proved.
MM-PA> minimize_with * / include_mathboxes
(..) Proof of “anbi1ci”  decreased from 44 bytes to 30 bytes using “bian”. (...)
 
(anbi1cd, former anbi2rid, with removed biand, using anbi2d, ancom and syl6bb:)
MM> prove anbi1cd
(...) The proof you are starting with is already proved.
MM-PA> minimize_with * / include_mathboxes
(..) Proof of “anbi1cd”  decreased from 47 bytes to 33 bytes using “biand”. (...)
 
My problem is that I don’t know if this could have been in another way: as I interpret this, I removed bian/d and the minimize_with * showed that I removed them. In theory, perhaps it could have been the case that this minimize_with * find another way to minimize anbi1ci and anbi1cd, but it did not. As I interpret what you wrote above, I should keep bian and biand. Is this true? 
 
Another question: should I follow this minimize_with * process with all of my theorems by hand?

benjub:
In that case, yes, keep bian(d).  If you want to save an output of the metamath program, you can use `MM> open log file.txt` and later `MM> close log` (see the metamath help: `MM> help open`).  To learn the commands, I recommend typing `MM> set echo on` at the beginning of each session.  That way, when you type only prefixes of the commands (as I do below), the program displays the whole command.

As for minimizing, when I use the metamath proof assistant, I generally do this:
```
MM> pr xxx
MM-PA> as l yyy       [many times!]
...
...
MM-PA> im al
MM-PA> min *          [if "improve all" succeeded]
MM-PA> sa n/c
MM-PA> exi
MM> wr so set.mm/rew
```
I do the "improve all" when I feel that the proof is almost complete and the program will find it (you can try `im al /1 /d 2` or more... type `help im` for help).  If this does not work, then I continue to "assign" labels to proof steps, and I retry "improve all" when I feel it will be ok.  It may also be useful to do it mid-proof from time to time.  (The command "improve" cannot lead you to a wrong direction or create dv violations: either it finds a correct solution, or it does nothing.)

If you're not using metamath.c but mmj2, then once the proof is finished and written in set.mm, you can type in the shell `$ scripts/minimize xxx [set.mm]` (correct me @jkingdon if I'm wrong).

For global minimizations, there is scipts/minimize-all and scripts/min.cmd to minmize certain lists of theorems (those with given label-match, those within a given range, those using a given theorem... and many other variants) with other lists of theorems.

(Minimizations of theorems in mathboxes are strongly recommended but not enforced.)

More generally, it might be useful to gather somewhere the workflows followed by big contributors like @jkingdon @tirix @avekens @wlammen @digama0 @sctfn @glacode @giomasce @david-a-wheeler @jamesjer and others (sorry to those I forgot!) to share experiences.

benjub:
@mazsa : for your next PR, I would suggest that once you sorted out the minimization questions, you move to your mathbox the theorems of this PR which are between elv and 0qs (that is, in the section "Preparatory theorems") which do not use new definitions or syntax (so in that PR, do not put $c, $v, or $a statements nor the symbol definitions (alt)htmldef).

mazsa:
@benjub ok. I use mmj2, I'll check minimize there and PR those theorems asap. If somebody writes something for new contributors, I have very fresh eyes:) since I didn't read/used anything yet (neither commands of metamath.exe, nor set theory, etc.) and I'm ready to help with this after my theorems are in my mathbox. (For example, based on a conversation with Jim @jkingdon , I will read and try to parse  http://us.metamath.org/ileuni/mmil.html#intuitionize maybe I can reveal if something is not undestandable for greenhorns:).

jkingdon:
> More generally, it might be useful to gather somewhere the workflows followed by big contributors like @jkingdon @tirix @avekens @wlammen @digama0 @sctfn @glacode @giomasce @david-a-wheeler @jamesjer and others (sorry to those I forgot!) to share experiences.

I've made a start at this at https://github.com/metamath/set.mm/wiki/Minimizing but this is on the wiki partly because I wanted to encourage other people to add their thoughts or ways of doing it.

If we come up with something we like we can mention it in CONTRIBUTING.md or whereever we think it makes sense.

PR #2436, opened 2022-01-18
Add least common multiple to iset.mm

jkingdon:
The title is slightly misleading because this pull request is mostly about infimums. But they are motivated for what is needed for least common multiple (as far as lcmval anyway) and the pull request does go ahead and add the least common multiple syntax, definition, and the lcmval theorem.

Since the infimum (and supremum) work in iset.mm is a bit of a work in progress, it is perhaps a bit hard to say in a sentence or two exactly what is here, but I'll try. Sets of integers have the same supremum properties as in set.mm (for example, that the supremum is an element of the set), with two provisos: (1) the supremum needs to exist, which typically would be shown via decidability hypotheses as we do here in lcmval, and (2) only some of this has been proved in iset.mm to date, other parts are for the future as described in mmil.html.

There's one change to set.mm which is removing a comment which is no longer true (and no longer needed).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4490,9 +4490,7 @@ which makes this different from ~ nnregexmid .</TD>
 
 <TR>
 <TD>supminf</TD>
-<TD><I>none</I></TD>
-<TD>Probably could prove something like this, but the conditions
-around the existence of the supremum would need to change.</TD>
+<TD>~ supminfex</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4432,13 +4432,24 @@ a fintely supported function.</TD>
 
 <TR>
 <TD>suprzcl</TD>
-<TD><I>none</I></TD>
+<TD>~ zsssupcl</TD>
+<TD>Presumably it is possble to remove the ` sup ( B , RR , < ) e. ZZ `
+hypotheses (if we can intuitionize suprleub and the other theorems needed
+by the set.mm proof in a way satisfied by the other hypotheses).</TD>
 </TR>
 
 <TR>
 <TD>zriotaneg</TD>
 <TD><I>none</I></TD>
 <TD>Lightly used in set.mm</TD>
+</TR>
+
+<TR>
+  <TD>suprleub</TD>
+  <TD><I>none</I></TD>
+  <TD>Looks like it might be intuitionizable (with suitable changes to
+  the conditions for existence of the supremum), if we can get suplub2
+  (which set.mm uses via suprnub and suprlub).</TD>
 </TR>
 
 <TR>
@@ -4508,7 +4519,8 @@ which makes this different from ~ nnregexmid .</TD>
 <TD>suprzub</TD>
 <TD><I>none</I></TD>
 <TD>Presumably could prove something like this with different conditions
-for the existence of the supremum.</TD>
+for the existence of the supremum (see ~ infssuzledc for something along
+these lines).</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1552,7 +1552,7 @@ csbifgOLD</TD>
 
 <TR>
   <TD>ifcl , ifcld</TD>
-  <TD>~ ifcldcd</TD>
+  <TD>~ ifcldcd , ~ ifcldadc</TD>
 </TR>
 
 <TR>
@@ -1566,7 +1566,7 @@ favor of theorems in deduction form.</TD>
 
 <TR>
   <TD>ifex , ifexg</TD>
-  <TD>~ ifcldcd</TD>
+  <TD>~ ifcldcd , ~ ifcldadc</TD>
 </TR>
 
 <TR>
@@ -6598,6 +6598,13 @@ intuitionistic and it is lightly used in set.mm</TD>
   <TD>eucalg</TD>
   <TD>~ eucialg</TD>
   <TD>Tweaked for the different ` seq ` syntax</TD>
+</TR>
+
+<TR>
+  <TD>df-lcmf and theorems using it</TD>
+  <TD><I>none</I></TD>
+  <TD>Although this could be defined, most of the theorems would
+  need decidability conditions analogous to ~ zsupcl</TD>
 </TR>
 
 </TABLE>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4385,6 +4385,11 @@ favor of theorems in deduction form.</TD>
 </TR>
 
 <TR>
+  <TD>infrenegsup</TD>
+  <TD>~ infrenegsupex</TD>
+</TR>
+
+<TR>
 <TD>crne0</TD>
 <TD>~ crap0 </TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4482,10 +4482,16 @@ which makes this different from ~ nnregexmid .</TD>
 </TR>
 
 <TR>
-  <TD>infssuzle , infssuzcl</TD>
-  <TD><I>none</I></TD>
-  <TD>Would need additional conditions around the existence of
-  the infimum, for example along the lines of ~ zsupcl </TD>
+  <TD>infssuzle</TD>
+  <TD>~ infssuzledc</TD>
+</TR>
+
+<TR>
+  <TD>infssuzcl</TD>
+  <TD>~ infuzre</TD>
+  <TD>It should
+  be possible to show the infimum is an element of ` S ` , but ~ infuzre
+  only shows it is an element of ` RR ` .</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -85328,6 +85349,110 @@ nonnegative integers (cont.)". $)
       WBLYFXSUIXPWMWAXBWAWMXBWCWDWEWFSWGWHWIWJFWK $.
   $}
 
+  ${
+    $d x y z $.  $d y z F $.
+    negiso.1 $e |- F = ( x e. RR |-> -u x ) $.
+    $( Negation is an order anti-isomorphism of the real numbers, which is its
```
benjub, line 85355: ...is an involutive order anti-isomorphism...
?

iset.mm
```diff
@@ -85328,6 +85349,110 @@ nonnegative integers (cont.)". $)
...
+      QVKVRWAWB $.
+  $}
+
+  ${
+    $d A a b c y z $.  $d A a c w y z $.  $d A a w x y z $.  $d a c ph y $.
+    supinfneg.ex $e |- ( ph -> E. x e. RR ( A. y e. A -. x < y
+        /\ A. y e. RR ( y < x -> E. z e. A y < z ) ) ) $.
+    supinfneg.ss $e |- ( ph -> A C_ RR ) $.
+    $( If a set of real numbers has a least upper bound, the set of the
+       negation of those numbers has a greatest lower bound.  For a theorem
```
benjub, line 85399: "then the set of negations" ?  Actually, I find "opposite" more suiting and more common, in these statements.

iset.mm
```diff
@@ -85328,6 +85349,110 @@ nonnegative integers (cont.)". $)
...
+      UVGRYHUVGVKZYOUVKUUAUVNUXEYJUVJCYNUXEYIUVIYHUVGYGMUGUFUEUXEYTUVMCRUXEYPUV
+      LYSYHUVGYGMUDUHUEUIWSXLXHXMWK $.
+  $}
+
+  ${
+    $d A f g w x y z $.  $d f g ph x y z $.
+    supminfex.ex $e |- ( ph -> E. x e. RR ( A. y e. A -. x < y
+        /\ A. y e. RR ( y < x -> E. z e. A y < z ) ) ) $.
+    supminfex.ss $e |- ( ph -> A C_ RR ) $.
+    $( A supremum is the negation of the infimum of that set's image under
```
benjub, line 85441: "of the image of that set" ?

iset.mm
```diff
@@ -85424,7 +85549,9 @@ nonnegative integers (cont.)". $)
   ${
     $d A a b x y z $.
     $( The image under negation of a bounded-above set of reals is bounded
-       below.  (Contributed by Paul Chapman, 21-Mar-2011.) $)
+       below.  For a theorem which is similar but also adds that the bounds
+       need to be the tightest possible, see ~ supinfneg .  (Contributed by
```
benjub, line 85553: For a similar theorem about existence of lower and upper bounds, see ~ supinfneg .
?

iset.mm
```diff
@@ -104463,6 +104596,106 @@ which defines the set we are taking the supremum of must (a) be true at
       VNAVSWGABCDIJKEVMLVRFVCVHVDVIWAVKVMWFWEVEVFWCVGVJ $.
   $}
 
+  ${
+    $d B f g x y z $.  $d f g ph x y z $.
+    zssinfcl.ex $e |- ( ph -> E. x e. RR ( A. y e. B -. y < x /\
+                      A. y e. RR ( x < y -> E. z e. B z < y ) ) ) $.
+    zssinfcl.ss $e |- ( ph -> B C_ ZZ ) $.
+    zssinfcl.zz $e |- ( ph -> inf ( B , RR , < ) e. ZZ ) $.
+    $( The infiumum of a set of integers is an element of the set.
```
benjub, line 104605: infimum

iset.mm
```diff
@@ -104463,6 +104596,106 @@ which defines the set we are taking the supremum of must (a) be true at
...
+      BVWAVVMUUIWMUVGVWANVVRQUVEVVRVAUVFVVTGUVEVVRVBVCYNWPVVOVVTVVMGVVOVVMVVMYJ
+      VDVCUUJUUMUUKZUULUVPUWADQUVOUVTUVLUVNEUVJGVWCUUNUUOUUPUUQUURUQ $.
+
+    $d A j m y $.  $d A m n $.  $d M j m $.  $d M m n $.  $d S a b m x y z $.
+    $d S j m x y z $.  $d S a m w x y z $.  $d a b m n ph $.
+    $d j m ph x y z $.
+    $( The infimum of a subset of an upper set of integers is less than or
+       equal to all members of the subset.  (Contributed by Jim Kingdon,
+       13-Jan-2022.) $)
+    infssuzledc $p |- ( ph -> inf ( S , RR , < ) <_ A ) $=
```
benjub, line 104679: I was wondering why this was not trivial in much greater generality... And this is, seemingly, because df-sup is defined in terms of `-. R` ?  I think I made a proposal a few commits ago about defining it in a more streamlined way... I'll have to get back to it.

iset.mm
```diff
@@ -106184,6 +106417,73 @@ which defines the set we are taking the supremum of must (a) be true at
   $}
 
 
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+                      The least common multiple
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+According to Wikipedia ("Least common multiple", 27-Aug-2020,
```
benjub, line 106424: Formatting: blank lines after =-=-=-= and before $) and indentation of 2.

iset.mm
```diff
@@ -104463,6 +104596,106 @@ which defines the set we are taking the supremum of must (a) be true at
...
+      weq infclti cuz cfv crab elrabi eleq2s eluzelre 3syl inflbti mpd nltled )
+      ADPQUAZCAKLMNOPDQOUBZPRNUBZPRUCONUHVAVBQSTVBVAQSTUCUDAVAVBUEUFZABKLMCDEFG
+      HIJUGZUIACDRZCFUJUKZRZCPRIVGCBEVFULDBECVFUMHUNFCUOUPAVECUTQSTIAKLMNOPDCQV
+      CVDUQURUS $.
+
+    ${
+      $d S f g x y z $.  $d f g ph x y z $.
+      $( The infimum of a subset of an upper set of integers exists.  It should
+         be possible to show the infimum is an element of ` S ` , but this
+         theorem only shows it is an element of ` RR ` .  (Contributed by Jim
```
benjub, line 104690: Hopefully soon !

benjub (approved):
One typo and a few suggestions.

iset.mm
jkingdon: Not a fan of that wording because it is supinfneg is about a least upper bound not just an upper bound (and likewise for greatest lower bound and lower bound).

iset.mm
jkingdon: Maybe I need to see what you are thinking in the form of proofs? This is just a translation of http://us.metamath.org/mpeuni/infssuzle.html with a bit of decidability added. So it isn't really more or less trivial than set.mm except for the decidability (which set.mm gets for free).

iset.mm
jkingdon: In this case I think it probably will be a "soon" (depending a bit on how much I get distracted by going through the set.mm supremum/infimum sections more generally - which is all good stuff even aside from this particular topic).

jkingdon:
I'll make a follow up for the typo and perhaps a few of the other suggestions.

iset.mm
benjub: You're right. I meant: For a similar theorem about existence of infimum and sumremum, see ~ supinfneg .

iset.mm
benjub: I mean changing df-sup, although I cannot retrieve my proposal. Currently, we have
> sup ( A , B , R ) = U. { x e. B | ( A. y e. A -. x R y /\ A. y e. B ( y R x -> E. z e. A y R z ) ) }

and I meant replacing -. x R y with ( y R x \\/ y = x ) , which seems to better translate "upper bound" and avoids negation, so is more faithful in intuitionistic logic.

iset.mm
jkingdon: I'm adding the blank lines. As for the identation, set.mm seems to have either an identation of 0 or 2 in different places. I'm reluctant to just mess with this one and still leave things inconsistent.

iset.mm
jkingdon: I don't think I understand either wording. And yes this is the same in set.mm and iset.mm.

iset.mm
jkingdon: :shrug: "of the image of that set" and "of that set's image" mean the same thing and I'm not sure I have an opinion.

iset.mm
jkingdon: Hmm, http://us.metamath.org/mpeuni/df-neg.html makes it look like we'd say "the set of negatives of those numbers".

I found a few wordings in set.mm including "negatives", "negations", "additive inverses" and (in a different place in a sentence) "unary minus". I didn't notice anything using "opposite" for additive inverse but perhaps that usage is out there too. I didn't research outside set.mm/iset.mm.

iset.mm
benjub: This is discussed in #2262 (but not a priority to me).

iset.mm
benjub: I thought " 's " was used for persons, but obviously I'll trust you on this !

iset.mm
benjub: "Involutive" means "equal to its inverse".
An "anti-endomorphism" of (RR, <) is a function f : RR \to RR such that a<b implies f(b)<f(a). (the "anti" means that we consider the opposite order structure).
So actually, the comment should read "involutive anti-automorphism" ! But it's fine as is.

iset.mm
jkingdon: I wouldn't be surprised if there is a style guide somewhere that says to use 's for persons, and there is (at least for some writers) a difference in formality, but these are pretty much fine points.

iset.mm
jkingdon: > I meant replacing -. x R y with ( y R x \\/ y = x )

I'm a little unsure of what to tell you on this beyond "give it a try and see how disjunction and equality behave when you try to write proofs using them". Maybe try to prove something like maxle1 using this definition. Seems like this could be fruitful whether the result is "you discover something that Mario and I have been missing" or "you better understand what we are talking about" (or some of each, even).

iset.mm
jkingdon: Making progress - #2440 . Although I did get a bit "distracted" by going through one supremum section, I have no regrets because a lot of this will be useful one way or another.

jkingdon: I have a branch which addresses the "Hopefully soon" (in the sense of replacing `infuzre` with a similar theorem whose conclusion is `ph -> inf ( S , RR , < ) e. S`). But I'm waiting on submitting it until #2440 is reviewed because it depends on the contents of that pull request.

jkingdon: Now submitted as #2443

PR #2438, opened 2022-01-19
minimize test

glacode:
A first test to include minimize in my workflow (I didn't check new discouraged, but I hope minimize doesn't use those)

A number of theorems in my mathbox have been minimized (most didn't change)

  liminfltlem
and from ifcli to nel2nelini, i.e.
  ifcli
  resabs2i
  ssdf2
  rabssd
  ssrind
  rexnegd
  rexlimd3
  resabs1i
  nel1nelin
  nel2nelin
  rexlimdva2
  nel1nelini
  nel2nelini


benjub:
So ifcli is a duplicate of keepel ? If yes, you may remove it.
You can PR the mentioned minimizations.
minimize_with does not use discouraged statements, unless you use the option /override.

A global minimization is done more or less regularly (yearly, in the best case).  The latest one gave rise to discussions on the google group, so you may have a look. In particular, do not take too much time to minimize already minimized theorems.

benjub (approved):

avekens (approved):

jkingdon:
> A global minimization is done more or less regularly (yearly, in the best case). The latest one gave rise to discussions on the google group, so you may have a look. In particular, do not take too much time to minimize already minimized theorems.

The last global minimization was in... maybe 2019 I think? And I remember at least some discussion of planning to not do any more of them (although I'm not sure where, so I'm not sure I can find it).

I've been pretty liberally minimizing anything which will minimize, although it is more "if I notice something in the process of my regular work" than anything more systematic than that.

jkingdon (approved):

PR #2439, opened 2022-01-19
my mathbox: elmtwise --> elwise; other: pair-wise --> pairwise, compo…

benjub:
…nent-wise --> componentwise, for consistency


benjub:
Pinging @sctfn for the comment edit. Also, @sctfn , it would be nice to upgrade ~eltrans to the closed form
|- ( A e. V -> ( A e. Trans <-> Tr A ) )
(and maybe also |- ( A e. Trans <-> ( A e. _V /\ Tr A ) ) ), maybe the same with ~ brsset, ~brbigcup, ~fvbigcup, ~elfix (two forms), ellimits (two forms), elfuns (second form since elfunsg is already there), snelsingles...
And similarly maybe upgrade ~btrxp to either closed form or the deduction form
( ph -> X e. V ) & ( ph -> Y e. W ) & ... => ( ph -> ...)
same with ~brsingle, ~brimage, ~brcart, ~brcup, ~brcap, ~brsuccf, ~brfunnfun, ~brrestrict, ~brub, ~brlb.

The comment of your ~altopth says that ~opth requires D to be a set but this is not the case.

sctfn:
Comment edit is fine by me. 

I can do the retrofit. These theorems are mostly from before deduction form was a thing, so brtxp wasn't written with that in mind, and it kind of trickles out from there.

For altopth, that comment is from before we redefined ordered pair to have the if condition. I keep thinking of proposing my version as the official definition, since it involves no Metamath specific operators or dummy variables, but I don't know how big a deal that is.

benjub:
> I keep thinking of proposing my version as the official definition, since it involves no Metamath specific operators or dummy variables, but I don't know how big a deal that is.

Then prepare for some competition: http://us.metamath.org/mpeuni/bj-2uplth.html ;)

But I think this can wait a bit.

avekens (approved):

jkingdon (approved):

PR #2442, opened 2022-01-22
Graph theory - minor improvements of Huneke's proof (4)

avekens:
Main set.mm:
* ~clelsb3f moved from TA's mathbox
* ~reu8nf, ~fvmptd3f added
* ~reuccats1, ~reuccats1v, ~numclwlk2lem2f1o revised
* proofs of ~fvmptdf, ~numclwwlk5 shortened
* ~numclwlk1lem2*: setvar variable changed
* ~numclwwlk3 made independent from local definitions

AV's mathbox:
* header comment of section "Goldbach's conjectures" restructured


set.mm
```diff
@@ -43972,6 +43985,24 @@ practical reasons (to avoid having to prove sethood of ` A ` in every use
...
 
+  ${
+    $d w x y A $.  $d x y z A $.  $d y z ph $.  $d w ph $.  $d z ps $.
+    $d w ps $.  $d y ch $.
+    reu8nf.1 $e |- F/ x ps $.
+    reu8nf.2 $e |- F/ x ch $.
+    reu8nf.3 $e |- ( x = w -> ( ph <-> ch ) ) $.
+    reu8nf.4 $e |- ( w = y -> ( ch <-> ps ) ) $.
+    $( Restricted uniqueness using implicit substitution.  This version of
+       ~ reu8 uses a bound-variable hypothesis for ` x ` and ` ps ` instead of
```
benjub, line 43996: I suggest: ...uses a non-freeness hypothesis for...
Since we generally refer to F/ as the non-freeness predicate, this might make searches easier (and there is always the fact that our definition is not syntactic e.g. x is non-free in our sense in x=x, although it is not bound either).

set.mm
```diff
@@ -69377,22 +69408,34 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
...
-    $d x A $.  $d x D $.  $d x ph $.
+    $d x A $.  $d x D $.
     fvmptdf.1 $e |- ( ph -> A e. D ) $.
     fvmptdf.2 $e |- ( ( ph /\ x = A ) -> B e. V ) $.
     fvmptdf.3 $e |- ( ( ph /\ x = A ) -> ( ( F ` A ) = B -> ps ) ) $.
+    ${
+      fvmptd3f.4 $e |- F/_ x F $.
+      fvmptd3f.5 $e |- F/ x ps $.
+      fvmptd3f.6 $e |- F/ x ph $.
+      $( Alternate deduction version of ~ fvmpt with 3 bound-variable
```
benjub, line 69419: ...three non-freeness hypotheses...

set.mm
```diff
@@ -425279,9 +425340,9 @@ afterwards the last vertex (which is the first vertex) itself ("walking
       IYOUUEUURSUUAABCFGHIJKLMNYJVSYKWD $.
 
     ${
-      $d C w $.
-      numclwwlk.t $e |- T = ( w e. ( X C N )
-            |-> <. ( w substr <. 0 , ( N - 2 ) >. ) , ( w ` ( N - 1 ) ) >. ) $.
+      $d C u $.  $d F u $.  $d V u $.
```
benjub, line 425343: maybe you can move $d F u $. and  $d V u $. a bit down, where they begin to be necessary.
**Edit**: maybe it's the correct place, I do not see the following $e and $p statements to tell for sure.

set.mm
```diff
@@ -425322,96 +425387,97 @@ with the first (and therefore last) vertex to the pair of the shorter
...
 
       $d C a b p x $.  $d F a b p x $.  $d G a b p w x $.  $d N a b p x $.
-      $d T a b p x $.  $d V a b p x $.  $d X a b p x $.  $d a i $.
+      $d T a b p x $.  $d V a b p x $.  $d X a b p x $.  $d a i $.  $d b u $.
       $( ` T ` is an onto function.  (Contributed by Alexander van der Vekens,
          20-Sep-2018.)  (Revised by AV, 29-May-2021.) $)
       numclwlk1lem2fo $p |- ( ( G e. USGraph /\ X e. V
                                    /\ N e. ( ZZ>= ` 3 ) )
                  -> T : ( X C N )
                         -onto-> ( ( X F ( N - 2 ) ) X. ( G NeighbVtx X ) ) ) $=
```
benjub, line 425428: you can probably write the statement on three lines instead of four.

set.mm
```diff
@@ -425735,15 +425807,11 @@ by two edges to a closed walk v(0) ... v(n)=v=v(0) in exactly one way.
...
-       v(n-2) = v(n) (see ~ numclwwlk1 ) and with v(n-2) =/= v(n) ( see
-       ~ numclwwlk2 ): f(n) = kf(n-2) + k^(n-2) - f(n-2) = (k - 1)f(n - 2) +
-       k^(n-2).  (Contributed by Alexander van der Vekens, 26-Aug-2018.)
-       (Revised by AV, 1-Jun-2021.) $)
-    numclwwlk3 $p |- ( ( ( G RegUSGraph K /\ G e. FriendGraph )
+    $d F w $.
+    $( Obsolete version of ~ numclwwlk3 as of 21-Jan-2022.  (Contributed by
+       Alexander van der Vekens, 26-Aug-2018.)  (Revised by AV, 1-Jun-2021.)
+       (Proof modification is discouraged.)  (New usage is discouraged.) $)
+    numclwwlk3OLD $p |- ( ( ( G RegUSGraph K /\ G e. FriendGraph )
```
benjub, line 425814: generally, one puts the OLD after the new statement, but do not bother if this requries additional changes

set.mm
```diff
@@ -715873,13 +715947,15 @@ be used (following the terminology used in [Helfgott] p. 2), because there
   are a strong and a weak version of the ternary Goldbach conjecture.  The term
   _Goldbach partition_ is used for a sum of two resp. three (odd) primes
   resulting in an even resp. odd number without further specialization.
+  <br><br>
```
benjub, line 715950: two blank lines each time ?

benjub (approved):

set.mm
avekens: I copied the text from another comment, but I agree that your suggestion is better. I'll change it here, but it should also be changed in other comments (later).

set.mm
avekens: OK, will be changed.

set.mm
avekens: It is at the correct place.

set.mm
avekens: Good suggestion, I'll change that.

set.mm
avekens: To move the OLD theorem is not possible in this case, because it is inside a block. In about a year, this will be removed anyway...

set.mm
avekens: no, only one blank line: the first <br> performs a "carriage return", the second <br>produced an empty line.

avekens: Curiously, the "< b r > " (without blanks) in my previous comment caused carriage returns there :-).

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
benjub: bound-variable --> non-freeness !

set.mm
benjub: maybe even two lines !
```
      numclwlk1lem2fo $p |- ( ( G e. USGraph /\ X e. V /\ N e. ( ZZ>= ` 3 ) )
       -> T : ( X C N ) -onto-> ( ( X F ( N - 2 ) ) X. ( G NeighbVtx X ) ) ) $=
```

set.mm
avekens: Done

set.mm
avekens: With two lines only the statement is too compressed and is not clearly apparent anymore (squeezed between the comment and the proof...).

PR #2443, opened 2022-01-22
More supremum/infimum theorems and least common multiple

jkingdon:
This finishes up the supremum/infimum work which was started in #2440 and earlier pull requests, at least "finished" in the sense of what is needed for least common multiple (in particular infssuzcldc which strengthens a previous version as discussed at https://github.com/metamath/set.mm/pull/2436#discussion_r787220648 ).

Another improvement is suprzclex which removes an unnecessary hypothesis from its predecessor.

There are a small number of miscellaneous improvements to iset.mm (for example copying theorems from set.mm or adjusting text in mmil.html).

The last part of the pull request is least common multiple. Once the infimum theorems are in place, intuitionizing the least common multiple section is the usual combination of proofs which work without any changes and (because this is integers and just about everything is decidable) proofs which work with straightforward intuitionizing. This pull request goes through lcmdvds which is most of the way through the least common multiple section (well until we get to the least common multiple of a set, which I wasn't thinking of tackling just yet).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4619,10 +4619,7 @@ which makes this different from ~ nnregexmid .</TD>
 
 <TR>
   <TD>infssuzcl</TD>
-  <TD>~ infuzre</TD>
-  <TD>It should
-  be possible to show the infimum is an element of ` S ` , but ~ infuzre
-  only shows it is an element of ` RR ` .</TD>
+  <TD>~ infssuzcldc</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3098,7 +3098,7 @@ this implies excluded middle</TD>
 
 <TR>
   <TD>fisup2g , fisupcl</TD>
-  <TD>~ zsssupcl</TD>
+  <TD><I>none</I></TD>
   <TD>Other variations may be possible, but the set.mm proof will not
   work as-is or with small modifications.</TD>
 </TR>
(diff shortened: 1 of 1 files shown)


avekens (approved):

iset.mm
```diff
@@ -106619,28 +106737,309 @@ According to Wikipedia ("Least common multiple", 27-Aug-2020,
...
+     divisor is the absolute value of the product of the two numbers.  In
+     particular, that absolute value _is_ the least common multiple of two
+     coprime numbers, for which ` ( M gcd N ) = 1 ` .
+
+     Multiple methods exist for proving this, and it is often proven either as
+     a consequence of the fundamental theorem of arithmetic or of
+     B&eacute;zout's identity ~ bezout ; see e.g.
+     ~ https://proofwiki.org/wiki/Product_of_GCD_and_LCM and
+     ~ https://math.stackexchange.com/a/470827 .  This proof uses the latter to
+     first confirm it for positive integer ` M ` and ` N ` (the "Second Proof"
```
benjub, line 106992: integers

benjub (approved):

iset.mm
jkingdon: Sounds like a good change.

I guess the way Norm sometimes handled this was to say "in your next pull request, could you X?". So I've put this change (for both set.mm and iset.mm) on a branch but I guess I'll not put it into a pull request until I'd be making one anyway.

benjub: Agreed. When I approve the PR while making some comments, the comments are only suggestions and/or can wait a future PR.

jkingdon: Sounds good. If the branch is starting to get large, I'll definitely make a pull request, but it shouldn't cause trouble to combine this comment fix with the next little chunk of work.

PR #2445, opened 2022-01-23
Moving some theorems from Mathboxes to the main part of set.mm

tirix:
In order to prepare for PR #2435, this moves the following theorems to the main part of set.mm:
* Rodolfo Medina`s n0el , ~eqbrrdv2 (renamed "eqbrriv2")~ , 2r19.29 , jca2
* Richard Penner´s dfss6
* Scott Fenton´s brv , dfres3 
* Thierry´s idssxp , eqvincg

`eqvincg` has been used to shorten `eqvinc`.
(this PR also includes an update of the STS typesetting file with a definition of the `liminf` syntax)

### DIFF
--- a/set-mathml.mmts
+++ b/set-mathml.mmts
@@ -1329,6 +1329,7 @@ $s class CaraGen $: <mi href="df-caragen.html">CaraGen</mi> $.
 $s class voln* $: <mi href="df-ovoln.html">voln*</mi> $.
 $s class voln $: <mi href="df-voln.html">voln</mi> $.
 $s class SMblFn $: <mi href="df-smblfn.html">SMblFn</mi> $.
+$s class liminf $: <mi href="df-liminf.html">lim inf</mi> $.
 
 
 $( Alexander van der Vekens $)


set.mm
```diff
@@ -41910,13 +41925,20 @@ choice between (what we call) a "definitional form" where the shorter
 
   ${
     $d x A $.  $d x B $.
+    $( A variable introduction law for class equality, deduction version.
+       (Contributed by Thierry Arnoux, 2-Mar-2017.) $)
+    eqvincg $p |- ( A e. V -> ( A = B <-> E. x ( x = A /\ x = B ) ) ) $=
```
avekens: This theorem is not a "deduction version" (of ~eqvinc?), since it is not of deduction form according our conventions (its conclusion does not have the antecedent ` ph -> `), but it is in "closed form" (using suffix "g"). ~eqvinc, however, is the inference version of ~eqvincg (see, for example, the comment of ~difexi).

avekens (changes requested):

set.mm
benjub: Agreed.
> deduction version --> closed form

set.mm
```diff
@@ -39389,6 +39397,13 @@ choice between (what we call) a "definitional form" where the shorter
       ( nfv r19.29af ) ABCDEADHFGI $.
   $}
 
+  $( Double the quantifiers of theorem r19.29.  (Contributed by Rodolfo Medina,
+     25-Sep-2010.) $)
```
benjub: Theorem ~ r19.29 with two quantifiers.

(add the tilde)

set.mm
```diff
@@ -57059,6 +57098,11 @@ This theorem is proved directly from set theory axioms (no set theory
       ALABMNUDUEOUHUDPABQUDUERSTTJACUAT $.
   $}
 
+  $( The binary relationship over ` _V ` always holds.  (Contributed by Scott
+     Fenton, 11-Apr-2012.) $)
+  brv $p |- A _V B $=
```
benjub: Two classes are always in relation by ` _V ` ?
Anyway, it's hard to come up with a meaningful description since this theorem is not very mathematical (since _V is not a binary relation).

set.mm
```diff
@@ -59996,6 +60040,17 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
   $}
 
+  ${
+    $d x y A $.  $d x y B $.  $d ph x $.  $d ph y $.
+    eqbrriv2.1 $e |- ( ( ( Rel A /\ Rel B ) /\ ph )
+                        -> ( x A y <-> x B y ) ) $.
+    $( Other version of ~ eqbrrdiv .  (Contributed by Rodolfo Medina,
+       30-Sep-2010.) $)
+    eqbrriv2 $p |- ( ( ( Rel A /\ Rel B ) /\ ph ) -> A = B ) $=
+      ( wrel wa wceq cv wbr cop wcel df-br 3bitr3g eqrelrdv2 anabss5 ) DGEGHZAD
```
benjub: Since this is in deduction form, I think the label should have a "d" suffix, and the ph in the antecedent of the hypothesis and the conclusion should be the first conjunct (and "deduction form" should appear in the comment).

benjub:
jca2 can also be moved after jcad (the comment looks fine).

set.mm
tirix: The `eqbrriv2` naming was suggested by Alexander so I used it. As you point out this is not a well-formed deduction version, I would have written this theorem like this in deduction form:
```
    eqbrriv2.1 $e |- ( ph -> Rel A ) $.
    eqbrriv2.2 $e |- ( ph -> Rel B ) $.
    eqbrriv2.3 $e |- ( ph -> ( x A y <-> x B y ) ) $.
    $( Other version of ~ eqbrrdiv .  (Contributed by Rodolfo Medina, 30-Sep-2010.) $)
    eqbrriv2 $p |- ( ph -> A = B ) $=
```
However, if I rewrite this theorem as such, then I would also have to change Rodolfo's proof of `prter3` in his own Mathbox. We generally try to avoid that.

set.mm
tirix: I've updated the comment with a bit longer explanation:
```
  $( With our set-theoretic definition, the universal class ` _V ` can be seen
     as a binary relation, which holds without condition for any two classes.
     (Contributed by Scott Fenton, 11-Apr-2012.) $)
```

avekens (approved):

set.mm
tirix: **Edit**
Actually what I'm proposing is exactly the statement of `eqbrrdv`! Then I would prefer _not_ to move `eqbrriv2 ` to the main part of `set.mm` and to ask Peter to rewrite his proofs using `~ eqbrrdv`

avekens: @tirix the already existing ~eqbrrdv is exactly what you propose! Using this theorem, the proof of ~ eqbrriv2 can be shortened:
       ( wrel wa simpll simplr eqbrrdv ) DGZEGZHAHBCDELMAILMAJFK $.
Knowing this, I think we should NOT move eqbrriv2 to main, but Peter should directly use ~eqbrrdv instead of ~eqbrriv2.

set.mm
benjub: This just shows that "our set-theoretic definition" is misleading !  The theorem also uses the fact that a couple with at least one component being a proper class is encoded as the empty set...
Anyway, we're not going to change this here and now.

benjub (approved):

set.mm
tirix: I think out messages crossed!
Obviously we agree, I'll merge this with `eqbrriv2` _not_ moved.

set.mm
tirix: Very right!
I've opened #2447 with a proof that ` _V ` is actually _not_ a relation (`-. Rel _V`). We can discuss there further how to phrase the explanation...

PR #2441, opened 2022-01-21
htmldefs althtmldefs latexdefs for my Mathbox

mazsa:
Inserted htmldefs althtmldefs latexdefs for Mathbox of Peter Mazsa.

PR #2446, opened 2022-01-23
First theorem in my Mathbox

mazsa:
anbi1ci cf. https://github.com/metamath/set.mm/pull/2441#issuecomment-1018596933


tirix (approved):
Looks very good to me: valid, mathbox-only.

PR #2447, opened 2022-01-23
The universal class is not a relation.

tirix:
A possibly interesting proof following a remark from @benjub in PR #2445


benjub:
Nice. I think this ~nrelv can be in Main (since ~rel0 is).

tirix:
Ok, I've moved it just after `~ rel0`. 
Does anyone have any issue with this theorem? @avekens @jkingdon ?

benjub (approved):

set.mm
```diff
@@ -57107,9 +57107,9 @@ This theorem is proved directly from set theory axioms (no set theory
       ALABMNUDUEOUHUDPABQUDUERSTTJACUAT $.
   $}
 
-  $( With our set-theoretic definition, the universal class ` _V ` can be seen
-     as a binary relation, which holds without condition for any two classes.
-     (Contributed by Scott Fenton, 11-Apr-2012.) $)
+  $( Two classes are always in relation by ` _V ` .  This theorem is somewhat
+     misleading, as the universal class ` _V ` is actually not a relation (see
+     ~ nrelv ).  (Contributed by Scott Fenton, 11-Apr-2012.) $)
   brv $p |- A _V B $=
```
avekens, line 57113: Maybe we could/should add "Nevertheless, this theorem is true, because of the general definition ~ df-br of three class symbols in a row, which is not defined for relations only. It simply states that an ordered pair is a set: ` ( A _V B <-> <. A , B >. e. _V ) `.

avekens (changes requested):

jkingdon (approved):

set.mm
benjub: I think it's not necessary to add this. (Of course it's true since it's in a verified mm file.)

avekens: Somebody reading this theorem may ask her-/himself "Why is it true, although ` _V ` is not a relation?" I asked this question myself, too, and it took some minutes until I got the answer. If we write the answer into the comment, is is immediatly available. But maybe it is not worth to dispute over this...

avekens (approved):

set.mm
benjub: As you prefer.

jkingdon (approved):

PR #2448, opened 2022-01-23
add contributor; single-column

benjub:
Added @mazsa as a contributor.

I converted the list to single-column since it is easier to update while keeping alphabetical order.  It ok ? @tirix @avekens @jkingdon @digama0 ?


avekens:
Although the list is very long, it is more practical, so I agree.

avekens (approved):

jkingdon (approved):

digama0 (approved):

PR #2449, opened 2022-01-24
Add coprimality and congruence sections to iset.mm

jkingdon:
These are the sections "Coprimality and Euclid's lemma" and "Cancellability of congruences". Most of the theorems here intuitionize with just the usual tweaks around not equal versus apart and the like.

There are a few small changes to earlier in iset.mm based on recent changes in set.mm.

The pull request minimizes two theorems in set.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4180,6 +4180,16 @@ zero changes to apart from zero.</TD>
 </TR>
 
 <TR>
+  <TD>divmulass</TD>
+  <TD>~ divmulassap</TD>
+</TR>
+
+<TR>
+  <TD>divmulasscom</TD>
+  <TD>~ divmulasscomap</TD>
+</TR>
+
+<TR>
 <TD>divdir , divcan3 , divcan4</TD>
 <TD>~ divdirap , ~ divcanap3 , ~ divcanap4 </TD>
 </TR>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -107154,6 +107200,530 @@ According to Wikipedia ("Least common multiple", 27-Aug-2020,
...
+` A e. ZZ ` and ` B e. ZZ ` are coprime (or relatively prime) if
+` ( A gcd B ) = 1 `. The equivalence of the definitions is shown by
+~ coprmgcdb .  The negation, i.e. two integers are not coprime, can be
+expressed either by ` ( A gcd B ) =/= 1 `, see ~ ncoprmgcdne1b , or
+equivalently by ` 1 < ( A gcd B ) `, see ~ ncoprmgcdgt1b .
+
+A proof of Euclid's lemma based on coprimality is provided in ~ coprmdvds (as
+opposed to Euclid's lemma for primes).
+$)
+
```
avekens, line 107224: Do the conventions about section headers for set.mm also hold for iset.mm? If yes, then they should be applied here:  
      "The header comments should be spaced as those of Part 1, namely, with
       a blank line before and after the comment, and an indentation of two
       spaces."

avekens (approved):

iset.mm
jkingdon: Sure, if set.mm starts following them consistently. This header is copy-pasted from set.mm without re-indenting anything.

The topic is being tracked at #2262 and I'd support having iset.mm adopt whatever set.mm ends up doing.

benjub: Yes, this is also the case with https://github.com/metamath/set.mm/pull/2436#discussion_r787220429

PR #2450, opened 2022-01-28
Remove unnecessary braces from iset.mm and set.mm (part 1).

jamesjer:
Sorry this has been so long in coming.  This PR removes unnecessary braces from the FOL parts of both set.mm and iset.mm.  If it makes sense to everybody, I think I will do small sections of set.mm and iset.mm together like this.  Since theorems are actively being copied from set.mm to iset.mm, this will avoid reintroducing unneeded braces.


avekens (approved):

avekens:
Great, thanks a lot! And yes, please continue the clean up, and it would be very helpful if you do it in small portions as you suggested.

jkingdon:
> If it makes sense to everybody, I think I will do small sections of set.mm and iset.mm together like this. Since theorems are actively being copied from set.mm to iset.mm, this will avoid reintroducing unneeded braces.

I like the idea of doing set.mm and iset.mm together. It has long been my vision that iset.mm and set.mm be the same where there is no reason for them to be different.

PR #2451, opened 2022-01-28
Add prime numbers to iset.mm through dvdsprm

jkingdon:
The definition of prime number (including some of the alternative definitions) is the same as in set.mm. All of the theorems intuitionize (except for prime number induction, which probably will be the next pull request), mostly without much difficulty.

The trickiest part of this pull request is dealing with equinumerosity. Some of the ways that set.mm proves the equinumerosity theorems used by its prime number definition point back to theorems which we tried and failed (so far at least) to prove at #2203 and related areas. Fortunately the set.mm proofs for `pm54.43` and `pr2ne` work with minimal modification and I also found a proof for `en2eqpr`. Taken together these (a) hopefully advance our progress towards more finite set theorems at least a tiny bit, and (b) are what we need for prime numbers.

A few more minor things:

* Mark elirr as discouraged which is part of the "reduce use of ax-setind" project which @sctfn was pondering at https://github.com/metamath/set.mm/pull/2407#issuecomment-1002624704 (and the current state of which is outlined in the text which this pull request adds to the elirr comment)
* Since the iset.m proof of `isprm2lem` ended up being about half the length of the set.mm proof (by avoiding various unnecessary detours), copy it to set.mm. It only relies on theorems which are already in set.mm.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -108,6 +108,22 @@
 "df-pnf" is used by "pnfxr".
 "df-tru" is used by "tru".
 "dvelimALT" is used by "hbsb4".
+"elirr" is used by "elirrv".
+"elirr" is used by "fiunsnnn".
+"elirr" is used by "nndcel".
+"elirr" is used by "nndceq".
+"elirr" is used by "nntri2".
+"elirr" is used by "nntri3".
+"elirr" is used by "onnmin".
+"elirr" is used by "onunsnss".
+"elirr" is used by "ordirr".
+"elirr" is used by "ordsoexmid".
+"elirr" is used by "ordtri2or2exmid".
+"elirr" is used by "phpelm".
+"elirr" is used by "reg3exmidlemwe".
+"elirr" is used by "snon0".
+"elirr" is used by "ssnel".
+"elirr" is used by "sucprcreg".
 "equsalh" is used by "dvelimALT".
 "equsalh" is used by "dvelimfALT2".
 "equsalh" is used by "dvelimfv".
@@ -248,6 +264,7 @@ New usage of "df-tru" is discouraged (1 uses).
 New usage of "dftru2" is discouraged (0 uses).
 New usage of "difidALT" is discouraged (0 uses).
 New usage of "dvelimALT" is discouraged (1 uses).
+New usage of "elirr" is discouraged (16 uses).
 New usage of "equsalh" is discouraged (5 uses).
 New usage of "eu3h" is discouraged (3 uses).
 New usage of "fnexALT" is discouraged (0 uses).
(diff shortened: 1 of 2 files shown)


avekens (approved):

iset.mm
```diff
@@ -37961,8 +37961,19 @@ this axiom (that is, Non-wellfounded Set Theory but in the absence of
...
+       The reason that this theorem is marked as discouraged is a bit subtle.
+       If we wanted to reduce usage of ~ ax-setind , we could redefine
+       ` Ord A ` ( ~ df-iord ) to also require ` _E Fr A ` ( ~ df-frind ) and
+       in that case any theorem related to irreflexivity of ordinals could use
+       ~ ordirr (which under that definition would presumably not need
+       ~ ax-setind to prove it).  But since ordinals have not yet been defined
+       that way, we cannot rely on the "don't add additional axiom use" feature
+       of the minimizer to get theorems to use ~ ordirr .  To encourage
+       ~ ordirr when possible, we mark this theorem as discouraged.
+
```
benjub, line 37974: Can you add to the comment of ordirr something like:
The present proof requires ax-setind.  If in the definition of ordinals df-iord, we also required that membership be well-founded on any ordinal (see df-frind), then we could prove ordirr without ax-setind.

jkingdon: I'd approve a pull request to add that.

benjub: So would I.  I think the addition of this comment to elirr entails the addition of a similar comment to ordirr, and the person who does one does the other.

jkingdon: OK, OK, time for both of us to take a step back and laugh at the absurdity of the situation where we have a wording, we both agree on it, and we're trying to argue about who should put it in a pull request. I'll put the following in my next pull request:

```diff
diff --git a/iset.mm b/iset.mm
index e1d46bc0..2690d6ab 100644
--- a/iset.mm
+++ b/iset.mm
@@ -37983,8 +37983,11 @@ $)
   $}
 
   $( Epsilon irreflexivity of ordinals: no ordinal class is a member of itself.
-     Theorem 2.2(i) of [BellMachover] p. 469, generalized to classes.
-     (Contributed by NM, 2-Jan-1994.) $)
+     Theorem 2.2(i) of [BellMachover] p. 469, generalized to classes.  The
+     present proof requires ~ ax-setind .  If in the definition of ordinals
+     ~ df-iord , we also required that membership be well-founded on any
+     ordinal (see ~ df-frind ), then we could prove ~ ordirr without
+     ~ ax-setind .  (Contributed by NM, 2-Jan-1994.) $)
   ordirr $p |- ( Ord A -> -. A e. A ) $=
     ( wcel wn word elirr a1i ) AABCADAEF $.
 
```

benjub: Thanks.  I think it was indeed up to you, but don't worry, I'm taking that very lightly.

jkingdon: Taking it lightly seems like good advice.

PR #2453, opened 2022-01-28

mazsa:

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16994,6 +16994,7 @@ New usage of "nqex" is discouraged (4 uses).
 New usage of "nqpr" is discouraged (1 uses).
 New usage of "nsmallnq" is discouraged (3 uses).
 New usage of "nsnlpligALT" is discouraged (0 uses).
+New usage of "numclwwlk3OLD" is discouraged (0 uses).
 New usage of "nv0" is discouraged (5 uses).
 New usage of "nv0lid" is discouraged (4 uses).
 New usage of "nv0rid" is discouraged (7 uses).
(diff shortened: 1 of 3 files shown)


mazsa:
One of the checks failed, the log writes some git-specific thing different from what I submitted here:
https://github.com/metamath/set.mm/pull/2453/files

Log:
´Run echo 'Checking for set.mm rewrapping:' && diff -U 0 set.mm set-wrapped.mm
Checking for set.mm rewrapping:
--- set.mm	2022-01-28 09:21:04.928683631 +0000
+++ set-wrapped.mm	2022-01-28 09:21:49.968959034 +0000
@@ -552841 +552841 @@
- ${
+  ${
Error: Process completed with exit code 1.´

I think I caused this because I forgot to fetch some commits from the metamath/set.mm to my mazsa/set.mm before I committed from my computer the 2 parts of the Preparatory theorems, and git didn´t parsed it as different parts, I reversed them ... so I messed everyting up. Do you @jkingdon @tirix @benoit have some ideas how to fix it?

benjub:
This may be due to PR #2450. First do a `git pull upstream develop` while in your branch, and then (from https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md):
```
./metamath 'read set.mm' 'write source set.mm /rewrap' \
   erase 'read set.mm' 'save proof */compressed/fast' \
   'verify markup */file_skip/top_date_skip' 'verify proof *' \
   'write source set.mm' quit
```
You may have to change some paths.  Then, `git commit -am "...."` and `git push origin develop`.

As a side note: I recommend creating on your computer a new branch for each PR (which you can then delete), instead of doing everything on `develop`.
Namely:
```
git pull upstream develop   [while in develop]
git checkout -b newbranch
[modify, git commit -am "..."]
[modify, git commit -am "..."]
...
git push origin newbranch
[and once the PR is accepted: ]
git checkout develop
git branch -D newbranch
```

jkingdon:
I think @benjub answered the git questions so let us know if you have any additional problems with that.

As for the `el*v*` theorems, seems interesting in that I don't recall having seen this particular approach before. I assume you have seen the "Theorem forms" section of http://us.metamath.org/mpeuni/conventions.html (but that is just the background) because you seem to be using the terminology from there. One of the reasons to get this into a mathbox is so that we can see how these play out in a limited setting (I'm not sure I can see the consequences either without a bit more experience with them, but I do like the general idea of using deduction form or closed form rather than inference form for A e. V type conditions).

PR #2454, opened 2022-01-28

mazsa:


benjub:
If you revert the second commit, then the first commit will pass, and we'll probably merge it. There is no problem in doing this incrementally.

If you want to refer in comments to theorems not yet in set.mm, then use ~~? xxx

PR #2457, opened 2022-01-29

mazsa:


jkingdon (approved):

avekens:
@david-a-wheeler Why are most of the checks skipped? As long as these checks are not performed successfully, it makes no sense to review this PR, and it cannot be accepted and merged of course.

mazsa:
@avekens (cc @david-a-wheeler ) because github seems to have a deduplication process; actually, this is of the first of our 9 verifiers and it resulted https://github.com/metamath/set.mm/runs/4988743702?check_suite_focus=true in this: ` Run fkirc/skip-duplicate-actions@master
Skip execution because the exact same files have been successfully checked in https://github.com/metamath/set.mm/actions/runs/1764004854 ` The explanation is that I submitted exactly this file as a PR yesterday before this PR, all 9 checks performed then, but I revoked the submission.

jkingdon:
> github seems to have a deduplication process

Yeah, exactly. If you click through to https://github.com/metamath/set.mm/actions/runs/1764004854 you'll see all the tests ran and passed there.

So we can count this as a successful check.

mazsa:
Jim´s @jkingdon comment belongs here: https://github.com/metamath/set.mm/pull/2453/#issuecomment-1024328717 (because I closed PR 2453 and this 2457 version was merged).

PR #2459, opened 2022-01-30
Add "Coprimality and Euclid's lemma (cont.)" to iset.mm

jkingdon:
This section intuitionizes without much trouble. A few theorems needed some changes. The one which I didn't do yet is `ncoprmlnprm`, because it looked harder than the rest, but I noted that as a possible future project as I don't see any fundamental obstacle to intuitionizing what would basically be the set.mm proof.

Perhaps the largest change beyond the usual intuitionizing is to add expcan. _The following text is struck through because I now don't plan on pursuing any set.mm change, as described in comments below._ ~~The set.mm proof ( http://us.metamath.org/mpeuni/expcan.html )  is in terms of http://us.metamath.org/mpeuni/eqord1.html which at least to my eyes feels like kind of an old school lemma with a lot of hypotheses and a lot of set and class variables and sort of a tangled relationship between eqord1 and the theorem which is using it (for a more extreme example where I simplified this sort of thing in the past see #1364 where Norm said "I like this. Nice work!" but there wasn't discussion beyond my write-up and Norm's reaction much less an easy comparison with whether it is analogous to `eqord1` or a different situation) . The proof of `expcan` here basically calls http://us.metamath.org/mpeuni/letri3.html and http://us.metamath.org/mpeuni/lenltd.html to do the work that `eqord1` had been doing and has a lemma `expcanlem` which it can call twice to handle both directions (which is what `eqord1` is doing with those set variables). The length is somewhere in the same ballpark (maybe a few steps longer although I didn't try to carefully count nor factor in what the result of deleting eqord1 would be, if we were able to do that down the road).~~

~~Anyway, that's a bit long-winded, my real question is whether it is worth putting this proof of expcan into set.mm and marking eqord1 as discouraged in set.mm. It would appear that eqord1 was worked on by Norm and @digama0 (based on a quick glance, not extensive digging into old versions) and that @digama0 and @sorear have written proofs which use eqord1 . There are four usages of eqord1 other than expcan, so the point of marking it as discouraged is that I'm volunteering to do expcan but would leave the others for another time. I'd be glad to put this in pull request form if people think it is at least worth considering.~~

P.S. the second two paragraphs are about whether to make a follow up pull request as described there. I don't think that topic affects whether this pull request can be merged.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4049,8 +4049,11 @@ favor of theorems in deduction form.</TD>
 <TR>
 <TD>ltordlem , ltord1 , leord1 , eqord1 , ltord2 , leord2 , eqord2</TD>
 <TD><I>none</I></TD>
-<TD>These depend on real number trichotomy and are not used until
-later in set.mm.</TD>
+<TD>Although these presumably could be proved using theorems like
+~ letri3 and ~ lenlt , at least for now we have chosen to just invoke
+those other theorems directly (example: ~ expcan and its lemma
+~ expcanlem ) which avoids some extra set variables and produces proofs
+which are almost as short.</TD>
 </TR>
 
 <TR>
@@ -6001,10 +6004,8 @@ dividing by zero.</TD>
 </TR>
 
 <TR>
-<TD>expcan , expcand</TD>
+<TD>expcand</TD>
 <TD><I>none</I></TD>
-<TD>Presumably provable, but the set.mm proof uses eqord1
-and the theorem is lightly used in set.mm</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6004,11 +6004,6 @@ dividing by zero.</TD>
 </TR>
 
 <TR>
-<TD>expcand</TD>
-<TD><I>none</I></TD>
-</TR>
-
-<TR>
 <TD>ltexp2 , leexp2 , leexp2 , ltexp2d , leexp2d</TD>
 <TD><I>none</I></TD>
 <TD>Presumably provable, but the set.mm proof uses ltord1</TD>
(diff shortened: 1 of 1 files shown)


digama0:
For a bit more context on eqord1: It says that a strictly increasing function is injective, that is: `x < y -> f(x) < f(y)` implies `x < y <-> f(x) < f(y)`, `x <= y <-> f(x) <= f(y)`, and `x = y <-> f(x) = f(y)`. It's a pretty useful theorem since you get six order implications from one.

To state this in metamath, we need the expressions `f(x)` and `f(y)`, given by ltord.1, and `f(C)` and `f(D)` in ltord.2 and ltord.3 are used so that the output theorem does not use set variables (typical for most applications). Finally, we need `f : S -> RR` where `S` is some subset of `RR` (usually `RR` itself) in order for the `<` in the domain and codomain to make sense, and that's what ltord.4 and ltord.5 are about.

All in all, I don't think there is anything that needs changing in the theorem statement, except possibly for adding ph to the substitution lemmas (it's not conventional in set.mm but I now regret this), or moving `C e. S /\ D e. S` to additional hypotheses (but for most applications they are fine as antecedents).

jkingdon:
> It's a pretty useful theorem since you get six order implications from one.

I guess I was more happy getting those implications as I did by applying expcanlem twice in the expcan proof in this pull request. But :shrug: it is certainly easier for me to do less work so I'm fine with just leaving set.mm alone. There's nothing much wrong with how it proves expcan.

avekens (approved):

avekens:
I agree with leaving set.mm alone regarding the proof of ~expcan, but I would suggest to enhance the comment of ~eqord1 by the information Mario provided in his comment.

PR #2464, opened 2022-01-31
Add sqne2sq to iset.mm

jkingdon:
This is a key lemma that will be part of the proof of the irrationality of
the square root of two, as described at https://github.com/metamath/set.mm/issues/2298#issue-1046169760 .

Includes lemmas sqpweven and 2sqpwodd concerning exponents being odd
or even.


benjub:
I assume these lemmas are stated in this peculiar form because this works better with the other proofs ? (I mean: the definition of F is probably useful at other places?)
As for the comments, I would write: "The greatest power of two dividing the square of an integer (twice the square of an integer) is an even (an odd) power of two." This is more precise than your comment (20 is divisible by an odd power of 2, and also by an even power of 2).
Apart from that, LGTM.

jkingdon:
> I assume these lemmas are stated in this peculiar form because this works better with the other proofs ? (I mean: the definition of F is probably useful at other places?)

Maybe because I wasn't smart enough to think of a non-peculiar way to state them? I mean, http://us.metamath.org/ileuni/oddpwdc.html is stated the way it is because that's what is in set.mm so perhaps I was following that pattern but I don't really remember what I was thinking several months ago when I wrote these lemmas (before I went on the whole number theory kick which led up to euclemma which enabled me to finish their proofs).

> As for the comments, I would write: "The greatest power of two dividing the square of an integer (twice the square of an integer) is an even (an odd) power of two." This is more precise than your comment (20 is divisible by an odd power of 2, and also by an even power of 2). Apart from that, LGTM.

Ah good catch. I'll fix that in my next pull request.

**Edit** by @benjub : sorry, I accidentally edited your comment.  I hope that I reverted it and that what appears above is @jkingdon's version.  If in doubt, look at the latest version of this comment by @jkingdon.

jkingdon:
Oh, and I corrected the title of this pull request to "sqne2sq". Too late to fix it in the git history, alas.

benjub:
> > I assume these lemmas are stated in this peculiar form because this works better with the other proofs ? (I mean: the definition of F is probably useful at other places?) 

> Maybe because I wasn't smart enough to think of a non-peculiar way to state them?

It's really a genuine question. I would have thought of 
```
|- ( A e. NN -> E! n  e. NN0 ( ( ( 2 ^ ( 2 x. n ) ) || A ^ 2 /\ -. ( 2 ^ ( ( 2 x. n ) + 1 ) ) || A ^ 2 ) )
```
But the function ` F ` looks important, and maybe your version is more readily usable by later theorems. It could also be used to state:
```
${ $e |- Even = ... $. $e |- Squares = ... $.
$p |- ( 2nd ` ( `' F " Squares ) ) = Even $= ? $. $}
```
Rk: wouldn't it be more natural to "swap" the aguments of ` F `, that is, write an integer as n = 2^x . y (instead of 2^y . x) ?

Also: would it be useful to define the "prime decomposition" bijection: something like NN --> ( NN0 ^ (NN) ) (finite lists) or ( NN0 ^ (Primes) ). Or use polynomials in NN0[X] ?  Also asking @tirix , @avekens , @digama0 for their opinions.

jkingdon:
> It's really a genuine question.

If I gave a non-answer it was partly because I'm not sure I have many answers. My immediate need is for http://us2.metamath.org/ileuni/sqne2sq.html and anything about the function ` F ` was - at least for my current purposes - a step towards that goal.

> I would have thought of
> ```
> |- ( A e. NN -> E! n  e. NN0 ( ( ( 2 ^ ( 2 x. n ) ) || A ^ 2 /\ -. ( 2 ^ ( ( 2 x. n ) + 1 ) ) || A ^ 2 ) )
> ```

Yeah, seems like we probably could prove a number of rearrangements or corollaries. Seems likely to come up one way or another if people keep doing stuff with this sort of decomposition.

> But the function `F` looks important, and maybe your version is more readily usable by later theorems. It could also be used to state:
> 
> ```
> ${ $e |- Even = ... $. $e |- Squares = ... $.
> $p |- ( 2nd ` ( `' F " Squares ) ) = Even $= ? $. $}
> ```

Defining these sorts of classes doesn't immediately excite me but maybe I just don't see what you could do that way. It does remind me of the question of whether it is useful to think of "typing assumptions" versus "predicates" to use the language of https://github.com/metamath/set.mm/issues/2433#issuecomment-1013022657 

> Rk: wouldn't it be more natural to "swap" the aguments of `F`, that is, write an integer as n = 2^x . y (instead of 2^y . x) ?

Possibly? At least I always was saying to myself "OK, 1st is the odd number and 2nd is the exponent" but maybe I would have been needing to remind myself of the order if they were swapped too.

> Also: would it be useful to define the "prime decomposition" bijection: something like NN --> ( NN0 ^ (NN) ) (finite lists) or ( NN0 ^ (Primes) ). Or use polynomials in NN0[X] ?

You're definitely getting beyond my knowledge of the mathematics here.

tirix:
As for the prime decomposition theorem, it is [`arith1`](http://us2.metamath.org:88/mpeuni/1arith.html) in set.mm, written as :
```
|- M = ( n e. NN |-> ( p e. Prime |-> ( p pCnt n ) ) )
|- R = { e e. ( NN0 ^m Prime ) | ( `' e " NN ) e. Fin }
=> |- M : NN -1-1-onto-> R
```
(`M` is the unique decomposition of natural numbers into products of primes, and `pCnt` is the prime counting function, returning the highest exponent of a prime number `p` which divides a number `n`)

tirix:
> I assume these lemmas are stated in this peculiar form because this works better with the other proofs ? (I mean: the definition of F is probably useful at other places?)

I'm the one guilty for the formulation of `~ oddpwdc `. I originally wrote it as part of the proof for the Euler partition theorem [`~ eulerpart`](http://metamath.tirix.org:3030/mpests/eulerpart). That proof relies on establishing a bijection between the "odd" and "distinct" partitions, which exist because of the bijection between a natural number and its product of a power of two and an odd integer, stated by `~ oddpwdc`.

> Rk: wouldn't it be more natural to "swap" the arguments of F, that is, write an integer as n = 2^x . y (instead of 2^y . x) ?

I'm probably guilty of this one too, since it is part of the formulation of `oddpwdc`. I distinctly remember thinking about this and choosing this direction over the other, possibly to save one commutation step in the proof.

benjub:
@jkingdon @tirix Thanks for your remarks. Indeed, the formulation with `Squares` is probably not useful: it was only to have a variable-free statement.  I hadn't looked, but it's nice to have ~arith1 and ~df-pc (the p-adic valuation).  So one has something like 
```
|- ( ( A e. NN /\ N e. NN /\ P e. Prime ) -> N || ( P pCnt ( A ^ N ) ) )
```
? And one can deduce @jkingdon's result above in the case N=P=2.

About the order of the arguments of F: I was also thinking of F as the first step (while pCnt is the infinity-th step), while the second step would be to write n = 2^x.3^y.z, the third step n = 2^x.3^y.5^z.t, etc.

**Edit**: maybe the above could be proved from ( P pCnt ( A ^ N ) ) = ( N x. ( P pCnt A ) ) under natural hypotheses, or more generally, morphic properties of ` r e. QQ |-> \mapsto ( P pCnt r ) ` ... and maybe this is already in set.mm !

PR #2465, opened 2022-02-01
Further develop the maximum of two real numbers in iset.mm

jkingdon:
Although the hard work on this was done in December, this is tidying up some loose ends, proving a few more theorems we now can prove, and updating the maximum theorems in mmil.html.

The newly proved theorems are either exactly as stated in set.mm or merely have `if(𝐴 ≤ 𝐵, 𝐵, 𝐴)` changed to `sup({𝐴, 𝐵}, ℝ, < )` (equivalent in set.mm but in iset.mm we need to use the latter).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4747,7 +4747,7 @@ numbers).</TD>
 <TR>
 <TD>xrmax1 , xrmax2 , xrmin1 , xrmin2 , xrmaxeq , xrmineq ,
 xrmaxlt , xrltmin , xrmaxle , xrlemin , max1 , max2 , min1 , min2 ,
-maxle , lemin , maxlt , ltmin , max0sub , ifle</TD>
+lemin , maxlt , ltmin , max0sub , ifle</TD>
 <TD><I>none</I></TD>
 </TR>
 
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -108925,8 +108993,8 @@ infinite descent (here implemented by strong induction).  (Contributed
 
     ${
       $d A x y z $.  $d F a x y z $.  $d J a x y $.
-      $( The square of a natural number is divisible by a power of two to an
-         even exponent.  (Contributed by Jim Kingdon, 17-Nov-2021.) $)
+      $( The greatest power of two dividing the square of an integer is an even
+         power of two (Contributed by Jim Kingdon, 17-Nov-2021.) $)
```
benjub, line 108997: period after "two"

benjub (approved):

benjub:
Can you create a subsection "Maximum...", maybe beginning at ~maxcom ? (and, maybe also a subsection "maximum of integers..." where, if I remember, you already proved some results.) This can wait a later PR. The idea is that if one does a ctrl+F "maximum" on the TOC page http://us2.metamath.org/ileuni/mmtheorems.html, one gets a match.
Can you remind me if the following is intuitionistic : A, B e. RR -> sup ( { A , B } , RR , < ) e. { A, B } ? (it is in ZZ, but I do not remember in RR)

avekens (approved):

jkingdon:
> Can you remind me if the following is intuitionistic : A, B e. RR -> sup ( { A , B } , RR , < ) e. { A, B } ? (it is in ZZ, but I do not remember in RR)

In #2466 I show this is equivalent to real number dichotomy (or whatever name you want to give to `A <_ B \/ B <_ A `. They call this "analytic LLPO" at http://nlab-pages.s3.us-east-2.amazonaws.com/nlab/show/principle+of+omniscience#analytic which was the best reference I could quickly find on the topic.

PR #2467, opened 2022-02-02
iset.mm: The square root of two is irrational

### LINKED ISSUE #2298: Prove that the square root of two is irrational in iset.mm

jkingdon:
There's more on this proof at #2298 but basically (1) what we are proving is ``( sqrt ` 2 ) # ( A / B )`` (where `A` and `B` are natural numbers), and (2) the proof is not especially hard from http://us.metamath.org/ileuni/sqne2sq.html and http://us.metamath.org/ileuni/sqrt11ap.html .

This brings the number of iset.mm entries in mm_100.html to five (until now I was unwilling to give myself credit for http://us.metamath.org/ileuni/sqrt2irr.html for the reasons described there related to the definition of "irrational").

Fixes #2298

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -579,7 +579,10 @@ Glauco Siliprandi, 2017-06-29)</li>
 
 <!-- 4th added to list -->
 <li><a name="91">91</a>.  The Triangle Inequality (<a
-href="mpeuni/abstrii.html">abstrii</a>, by Norman Megill, 1999-10-02)</li>
+href="mpeuni/abstrii.html">abstrii</a>, by Norman Megill, 1999-10-02).
+The intuitionistic logic explorer (iset.mm) database includes this proof
+as <a href="http://us.metamath.org/ileuni/abstrii.html">abstrii</a>
+(added 2021-08-13).</li>
 
 <!-- 47th added to list -->
 <li><a name="93">93</a>.  The Birthday Problem (<a
--- a/mm_100.html
+++ b/mm_100.html
@@ -234,7 +234,11 @@ announcements.</font></p>
 
 <!-- 5th added to list -->
 <li><a name="1">1</a>.  The Irrationality of the Square Root of 2 (<a
-href="mpeuni/sqrt2irr.html">sqrt2irr</a>, by Norman Megill, 2001-08-20)</li>
+href="mpeuni/sqrt2irr.html">sqrt2irr</a>, by Norman Megill, 2001-08-20).
+The intuitionistic logic explorer (iset.mm) database also includes <a
+href="http://us.metamath.org/ileuni/sqrt2irrap.html">sqrt2irrap</a>
+(by Jim Kingdon, 2022-02-01).</li>
+</li>
 
 <!-- 37th added to list -->
 <li><a name="2">2</a>. The Fundamental Theorem of Algebra (<a


avekens (approved):

iset.mm
```diff
@@ -109111,6 +109111,46 @@ infinite descent (here implemented by strong induction).  (Contributed
...
+    XJWGWLGWMWGWLWSUOWGUPWGWMXDUOWGWMXDUQURUSOWGWNPDZGPDZXHXISWGWLTDWMCDXKWGWLW
+    SUTXDWLWMVDQZGTDXLWGVAGVBVCWNGVEQOWGWNVFDZRWNVGKGVFDZRGVGKZXGXHSWGXKXNXMWNV
+    HVIZWGRWNXAXQWGWLWMWTWGWMXDLXBWGWMXDMVJNXOWGVKVLXPWGVMVLWNGVNVOOVPWGWHVQDZW
+    IVQDZWJWKSWGWHPDZXRWEATDWFXTAVRABVDVSWHVTVIXSWGWIWAWBVLWHWIWCQWD $.
+
+  ${
+    $d Q a b $.
+    $( The square root of 2 is irrational.  That is, for any rational number,
+       ` ( sqrt `` 2 ) ` is apart from it.  In the absence of excluded middle,
+       we can distinguish between this and "the square root of 2 is not
```
benjub, line 109136: Maybe:
> In the absence of excluded middle, this is strictly stronger than "the square root..."

benjub:
Congratulations!  I made one suggestion above; it's as you prefer.

This is really "express merging".  Of course, it was approved, so all is ok, and this PR is perfectly fine. Without having hard rules, can we wait a bit more before merging ? Unless of course there are a lot of pending PRs, or large ones which may cause merging conflicts or are necessary to get merged quick.

jkingdon:
> This is really "express merging". Of course, it was approved, so all is ok, and this PR is perfectly fine. Without having hard rules, can we wait a bit more before merging ? 

Uh, maybe? On the one hand (a) we do say 'We generally tend to have a bias towards "merge this now" with separate follow-up fixes for minor issues' in CONTRIBUTING.md and (b) a merge doesn't have to be the end of that content getting reviewed or revised. On the other hand I suppose this wasn't a case where I particularly had other work backed up which would be easier with a merge (well except https://github.com/metamath/set.mm/issues/2468 although that is sorta not really related to whether the mm_100.html changes in this pull request are merged or not).

iset.mm
jkingdon: I really struggle with wording on this sort of thing. On the one hand no theorem is stronger than any other theorem in the sense of http://us.metamath.org/mpeuni/2th.html . On the other hand we need to find the language to talk about what sqrt2irr proves versus what sqrt2irrap proves.

benjub: I think "stronger" is appropriate, here.  It's a common wording, even if of course two truths are equivalent.  You can put quotation marks around it. It says that one implication is more "obvious" than the other.  For a justification, there is probably a weak base theory A (something like IFOL plus some real number axioms) such that over A, sqrt2irrap implies sqrt2irr but not conversely.  Basically, A should be able to prove that apartness implies difference, but should not be so strong that it would prove sqrt2irrap from sqrt2irr.  (If, on the other hand, there is no such theory A, then that is a result in itself!)

PR #2470, opened 2022-02-03
add mmbiblio.html file

digama0:
This file is missing from the list of files needed for web site generation, because it is treated as both an input and an output: the `WRITE BIBLIOGRAPHY` command will perpetuate all the surrounding text and only replace the content inside the `<!-- #START# -->...<!-- #END# -->` markers (as well as the `This page was last updated on` line). The version pushed here is a template that includes everything except the autogenerated content.

### DIFF
--- a/mmbiblio.raw.html
+++ b/mmbiblio.raw.html
@@ -1,0 +1,160 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
+   "http://www.w3.org/TR/html4/loose.dtd">
+<HTML LANG="EN-US">
+<HEAD>
+
+<!-- improve mobile display -->
+<META NAME="viewport" CONTENT="width=device-width, initial-scale=1.0">
+
+<META HTTP-EQUIV="Content-Type"
+ CONTENT="text/html; charset=iso-8859-1">
+<STYLE TYPE="text/css">
+<!--
+.p { font-family: "Arial Narrow";
+     font-size: x-small;
+     color: #FA8072;
+   }
+.r { font-family: "Arial Narrow";
+     font-size: x-small;
+   }
+.i { font-family: "Arial Narrow";
+     font-size: x-small;
+     color: gray;
+   }
+-->
+</STYLE>
+ <TITLE>Bibliographic Cross-Reference - Metamath Proof Explorer</TITLE>
+<LINK REL="shortcut icon" HREF="favicon.ico" TYPE="image/x-icon">
+</HEAD>
+
+<BODY BGCOLOR="#FFFFFF" STYLE="padding: 0px 8px">
+
+<TABLE BORDER=0 CELLSPACING=0 CELLPADDING=0 WIDTH="100%">
+
+  <TR>
+    <TD ALIGN=LEFT VALIGN=TOP><A HREF="mmset.html"><IMG SRC="mm.gif"
+      BORDER=0
+      ALT="Metamath Proof Explorer Home"
+      TITLE="Metamath Proof Explorer Home"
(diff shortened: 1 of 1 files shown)


benjub:
This PR adds mmbiblio.raw.html (not mmbiblio.html). Does the metamath.c command "write bibliography" take as input mmbiblio.raw.html and outputs mmbiblio.html, or does it modify one of them (the latter, as your comment seems to imply) and the transformation from "raw" to "non-raw" is done (apparently, before) somewhere else (by a script for the website generation)?
In any case, the end result is http://us2.metamath.org/mpeuni/mmbiblio.html and the file looks fine.

mmbiblio.raw.html
```diff
@@ -0,0 +1,160 @@
...
+<!-- do not change the next comment -->
+<!-- #START# -->
+<!-- #END# -->
+<!-- Do not change the preceding comment -->
+<!-- End of automatically generated section -->
+
+</TABLE>
+
+<HR NOSHADE SIZE=-1> <CENTER><I>
+This page was last updated on 1-Jan-2000.
```
benjub, line 146: Maybe you can add a comment that the date on that line is automatically updated (or put that comment with the explanation a few lines above), as you did in the explanation of the PR. Is it better to write a "non-date" instead of 1-Jan-2000 (e.g., dd-Mmm-yyyy) ?

digama0: I didn't check the granularity of the replacement pass, but if it's using a regex to find this line it might be problematic if it's not a valid date.

digama0:
So far this file isn't connected to anything. The `WRITE BIBLIOGRAPHY` command overwrites its input, so you have to copy this to `mmbiblio.html` and then run WRITE BIBLIOGRAPHY on it. I did check that doing so produces the same result on this file as it does with the real mmbiblio.html, so the anonymization does not seem to have caused any problems.

david-a-wheeler:
If we change things, we'll probably need to change the script on us2 that grabs the data & places it in the "right place".

mmbiblio.raw.html
benjub: Ok. What about adding the comment as I proposed above ?

jkingdon:
I think maybe we need two different checked in versions of what you are calling mmbiblio.raw.html - one to be generated into http://us.metamath.org/mpeuni/mmbiblio.html and one to be generated into http://us.metamath.org/ileuni/mmbiblio.html . 

Some of the machinery to keep set.mm and iset.mm separate is already there, for example set.mm contains `htmlbibliography "mmset.html";` and iset.mm contains `htmlbibliography "mmil.html";` but I guess we need to invent a new mechanism for whatever pieces had been manual when Norm was doing it.

P.S. Suitable disclaimers about whether I understand any of this - I somehow got a few files downloaded to my local machine, enough to work with my usual workflows, and promptly forgot what little I understood about how we handle the bibliography.

david-a-wheeler:
Either there should be 1 file used by both, or two different files (with clear names) that are separately edited.

digama0:
FYI, the name `mmbiblio.html` is baked into metamath.exe, it is not one of the HTML $t variables.

jkingdon (approved):
So we know there is more to do both in terms of documenting/scripting the process of turning this into something which goes on the website and in handling iset.mm.

I guess I'm putting this down as "approve" because I don't have an especially strong opinion about whether we (1) merge what we have so far and figure we can keep working on it in future, or (2) try to sort out some of these things before merging something which might be confusing if it isn't clear how one is supposed to use this .raw file.

mmbiblio.raw.html
digama0:
```suggestion
<!-- The line below is automatically updated. -->
This page was last updated on 1-Jan-2000.
```

PR #2472, opened 2022-02-04
fix markup

digama0:
This updates set.mm and iset.mm to pass the verify markup checks in david-a-wheeler/metamath-knife#70 .


set.mm
```diff
@@ -102,7 +102,7 @@ been continuously enriched since then (list of contributors below). $)
 
 $( (See 'help verify markup' for help with modularization tags.) $)
 $( Begin $[ set-header.mm $] $)
-$(
+$( !
```
digama0, line 105: The purpose of this is to break up the header comment sequence so that the comment is not interpreted as a regular header comment, because header comments have verified markup and the head comment has a lot of ascii syntax that does not conform to the markup parser's expectations (not least of which is the existence of multiple sub-headers).

Alternatives:
* Use another character like `.`
* Use something like `<not-a-header>`, which I wanted to avoid since readers will be confused by it since this is right at the start of the file
* Use a variation on the usual header decoration like `=~=~=~=~=~=~` so that it is not recognized as a header

set.mm
```diff
@@ -160005,8 +160006,7 @@ at index N (modulo the length of the word) of the original word.
 
   $( The symbol at index ((n-N)-1) of a word of length n (not 0) cyclically
      shifted by N positions is the symbol at index (n-1) of the original word.
-     (Contributed by AV, 23-May-2018.)  (Revised by AV, 21-May-2018.)  (Revised
-     by AV, 30-Oct-2018.) $)
+     (Contributed by AV, 23-May-2018.)  (Revised by AV, 30-Oct-2018.) $)
```
digama0, line 160009: @avekens I took a look around at the theorems here and made my best guess at the true history, although I didn't do a git blame. Do you have any extra insight into how the dates got mixed up here?

digama0: Aha, git blame reveals the truth: the date got mangled in 48c50b251 . Will fix.

set.mm
benjub: or maybe something like:
```
$( (text on this line to avoid comment being interpreted as a header comment)
```
?

benjub:
Minor side remark to @digama0 and @avekens : in both cases of unchronological order, I noticed that there are two or three contributed/revised credits that are close to each other in time and from the same person, and where the nature of the revision is not explained.  In these cases, it may be enough to keep only one "revised" credit.

jkingdon (approved):

set.mm
avekens: I would prefer the variation on the usual header decoration like =~=~=~=~=~=~ so that it is not recognized as a header.

set.mm
```diff
@@ -156082,7 +156082,8 @@ as finite sequences of _symbols_ (or _characters_) being elements of the
   <li>The length of a word is usually represented by class variables ` N `
   or ` L `.</li>
   <li>The number of position to cyclically shift a word is usually represented
-  by class variables ` N ` or ` L `.</li></ul></HTML>
+  by class variables ` N ` or ` L `.</li></ul>
+  </HTML>
```
avekens, line 156086: Excellent!

avekens (approved):

set.mm
digama0: would the variant header decoration be used on all the headers in this comment? It wouldn't match with the rest of the file stylistically, which seems not great either.

avekens: To be consequent - yes. Maybe a different style could be used for the 2nd level, e.g.,

-~-~-~-~-~-~-~-~-~-~-~-~
or
.~.~.~.~.~.~.~.~.~.~.~.~.~.~

jkingdon:
I hope I'm not jumping the gun on merging this but as far as I can tell the vast majority of the changes here are not controversial. Whatever we decide on `$( !` or equivalent we can make a followup pull request for.

PR #2474, opened 2022-02-06
Add finite set theorems such as negfi to iset.mm

jkingdon:
Some of these can be copied from set.mm without changes. Some of them cannot be proven (infiexmid and domfiexmid prove that these two set.mm theorems imply excluded middle). Some need additional conditions (e.g. dmfi, rnfi, cnvfi). Some, such as fimaxre2 or fnfi , can be stated as in set.mm but need some significant intuitionizing in the proof.

One miscellaneous (unrelated) change;
* state ax-0lt1 as an axiom which had been intended all along

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3037,8 +3037,19 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
-  <TD>unfi</TD>
-  <TD><I>none</I></TD>
+  <TD ROWSPAN="3">unfi</TD>
+  <TD>~ unsnfi</TD>
+  <TD>For the union of a set and a singleton whose element is not
+  a member of that set</TD>
+</TR>
+
+<TR>
+  <TD><I>for the union of two disjoint sets</I></TD>
+  <TD>presumably provable</TD>
+</TR>
+
+<TR>
+  <TD><I>for any two finite sets</I></TD>
   <TD>Not provable according to Remark 8.1.17 of [AczelRathjen].</TD>
 </TR>
 
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2970,8 +2970,7 @@ set.mm proof also uses domfi</TD>
 <TR>
 <TD>infi</TD>
 <TD><I>none</I></TD>
-<TD>Presumably the proof of ~ ssfiexmid could be adapted to show
-this implies excluded middle</TD>
+<TD>Implies excluded middle as shown at ~ infiexmid</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4441,18 +4441,11 @@ favor of theorems in deduction form.</TD>
 
 <TR>
   <TD>fimaxre</TD>
-  <TD><I>none</I></TD>
-  <TD>The theorem as stated states there is an upper bound (rather
-  than directly a maximum), so it probably would be provable
+  <TD>~ fimaxre2</TD>
+  <TD>When applied to a pair this could show which of two unequal
+  real numbers is larger, so perhaps not provable for that reason.
   (see ~ fin0 for inhabited versus nonempty).
   </TD>
-</TR>
-
-<TR>
-  <TD>fimaxre2</TD>
-  <TD><I>none</I></TD>
-  <TD>Presumably provable (see ~ fin0or for how to replace the case
-  elimination).</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2957,7 +2957,7 @@ ssfi , since ` { (/) } e. _om `</TD>
 <TR>
 <TD>domfi</TD>
 <TD><I>none</I></TD>
-<TD>The set.mm proof is in terms of ssfi</TD>
+<TD>Implies excluded middle as shown at ~ domfiexmid</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3062,6 +3062,11 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>dmfi</TD>
+  <TD>~ fundmfi</TD>
+</TR>
+
+<TR>
   <TD>abrexfi</TD>
   <TD><I>none</I></TD>
   <TD>At first glance it would appear that the mapping would need to
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3066,6 +3066,13 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>resfnfinfin</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would need to have a ` B C_ A ` condition added,
+  but lightly used in set.mm</TD>
+</TR>
+
+<TR>
   <TD>abrexfi</TD>
   <TD><I>none</I></TD>
   <TD>At first glance it would appear that the mapping would need to
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3073,6 +3073,12 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>residfi</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably provable, but lightly used in set.mm</TD>
+</TR>
+
+<TR>
   <TD>abrexfi</TD>
   <TD><I>none</I></TD>
   <TD>At first glance it would appear that the mapping would need to
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3084,6 +3084,11 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>rnfi</TD>
+  <TD>~ funrnfi</TD>
+</TR>
+
+<TR>
   <TD>abrexfi</TD>
   <TD><I>none</I></TD>
   <TD>At first glance it would appear that the mapping would need to
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3226,7 +3226,7 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
-  <TR>fiinfg , fiinf2g</TR>
+  <TD>fiinfg , fiinf2g</TD>
   <TD><I>none</I></TD>
   <TD>The set.mm proof relies on fiming</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3051,6 +3051,13 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>abrexfi</TD>
+  <TD><I>none</I></TD>
+  <TD>At first glance it would appear that the mapping would need to
+  be one to one or some other condition.</TD>
+</TR>
+
+<TR>
   <TD>dfsup2</TD>
   <TD><I>none</I></TD>
   <TD>The set.mm proof uses excluded middle in several places and the
@@ -4451,8 +4458,7 @@ favor of theorems in deduction form.</TD>
 <TR>
   <TD>fimaxre3</TD>
   <TD><I>none</I></TD>
-  <TD>Presumably provable (the set.mm proof would appear to work once
-  we have fimaxre2.</TD>
+  <TD>The set.mm proof relies on abrexfi .</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3079,6 +3079,11 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>cnvfi</TD>
+  <TD>~ relcnvfi</TD>
+</TR>
+
+<TR>
   <TD>abrexfi</TD>
   <TD><I>none</I></TD>
   <TD>At first glance it would appear that the mapping would need to
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4500,13 +4500,6 @@ favor of theorems in deduction form.</TD>
 </TR>
 
 <TR>
-  <TD>negfi</TD>
-  <TD><I>none</I></TD>
-  <TD>The set.mm proof depends on fundmfibi which seems like it
-  should be provable but which we don't have yet.</TD>
-</TR>
-
-<TR>
   <TD>fiminre</TD>
   <TD><I>none</I></TD>
   <TD>See fimaxre</TD>


iset.mm
```diff
@@ -41635,6 +41635,22 @@ arbitrary sets or real numbers will not be provable (without the law of
       ( cvv wcel cres resexg ax-mp ) ADEABFDECABDGH $.
   $}
 
+  $( Class restriction distributes over intersection.  (Contributed by FL,
+     6-Oct-2008.) $)
+  resindm $p |- ( Rel A -> ( A |` ( B i^i dom A ) ) = ( A |` B ) ) $=
+    ( wrel cres cdm cin resdm resindi incom inres inidm reseq1i 3eqtrri 3eqtr4g
```
benjub, line 41641: The comment does not correspond to the statement.

A precise description would be verbose, so maybe the less precise: "When restricting a relation, intersecting [the restrictor] with the domain of the relation has no effect."

iset.mm
```diff
@@ -41635,6 +41635,22 @@ arbitrary sets or real numbers will not be provable (without the law of
...
+    ineq2d ) ACZABDZAAEZDZFQAFZABRFDQPSAQAGOABRHTAQFAAFZBDQQAIAABJUAABAKLMN $.
+
+  $( Restricting a function to a domain without one element of the domain of
+     the function, and adding a pair of this element and the function value of
+     the element results in the function itself.  (Contributed by AV,
+     2-Dec-2018.) $)
+  resdmdfsn $p |- ( Rel R -> ( R |` ( _V \ { X } ) )
+                             = ( R |` ( dom R \ { X } ) ) ) $=
+    ( wrel cdm csn cdif cres cvv cin indif1 incom eqtri difeq1i reseq2i resindm
```
benjub, line 41650: The comment does not match the statement. Ping also @avekens

iset.mm
```diff
@@ -60958,6 +61008,23 @@ the cancellation property (fifth hypothesis), show that the relation
       UAUBUCUOURVDIVCUDULUQBUNUENUFUGUQAUMUHUIUJRUK $.
   $}
 
+  ${
+    $d A m n $.  $d B m n $.  $d V n $.
+    $( Adding a singleton to a finite set yields a finite set.  (Contributed by
+       Jim Kingdon, 3-Feb-2022.) $)
+    unsnfi $p |- ( ( A e. Fin /\ B e. V /\ -. B e. A )
+        -> ( A u. { B } ) e. Fin ) $=
+      ( vn vm cfn wcel wn cv cen wbr csn cun com wrex isfi wa ad2antrl cin c0
```
benjub, line 61017: Classically, you do not need B e. V nor -. B e. A (which you could join as B e. ( V \ A )). Do you need both intuitionistically ?
Maybe add also to iset: $p |- ( [what is needed] -> { A } e. Fin )

iset.mm
```diff
@@ -60969,6 +61036,73 @@ the cancellation property (fifth hypothesis), show that the relation
...
+    ( cfn wcel wfun cdm cen wbr fundmeng enfii syldan ) ABCADAEZAFGKBCABHKAIJ
+    $.
+
+  $( A function (set) is finite if and only if its domain is finite.
+     (Contributed by AV, 10-Jan-2020.) $)
+  fundmfibi $p |- ( Fun F -> ( F e. Fin <-> dom F e. Fin ) ) $=
+    ( wfun cfn wcel cdm fundmfi ancoms wfn funfn fnfi sylanb impbida ) ABZACDZA
+    EZCDZNMPAFGMAOHPNAIOAJKL $.
+
+  $( If a set is finite, its converse is as well.  (Contributed by Jim Kingdon,
```
benjub, line 61071: If a relation...

iset.mm
```diff
@@ -60969,6 +61036,73 @@ the cancellation property (fifth hypothesis), show that the relation
...
+      JAQVEPYCYAWLBWIVFVGVHYBXSWLXTGWLWIXTVIVJVKVLWJXHQVMVNVPVQVRWAWBRVSVT $.
+  $}
+
+  $( The domain of a finite function is finite.  (Contributed by Jim Kingdon,
+     5-Feb-2022.) $)
+  fundmfi $p |- ( ( A e. Fin /\ Fun A ) -> dom A e. Fin ) $=
+    ( cfn wcel wfun cdm cen wbr fundmeng enfii syldan ) ABCADAEZAFGKBCABHKAIJ
+    $.
+
+  $( A function (set) is finite if and only if its domain is finite.
```
benjub, line 61065: why "(set)" ?

iset.mm
```diff
@@ -60969,6 +61036,73 @@ the cancellation property (fifth hypothesis), show that the relation
...
+    relcnv cnvexg cnven sylancr adantl enfii syl2anc ) ABZACDZEZAFZFZCDUFUGGHZU
+    FCDUEUGACUCUGAIZUDUCUIAJKLUCUDMNUDUHUCUDUFBUFODUHAPACQUFORSTUFUGUAUB $.
+
+  $( The range of a finite relation is finite if its converse is a function.
+     (Contributed by Jim Kingdon, 5-Feb-2022.) $)
+  funrnfi $p |- ( ( Rel A /\ Fun `' A /\ A e. Fin ) -> ran A e. Fin ) $=
+    ( wrel ccnv wfun cfn wcel w3a crn cdm df-rn relcnvfi 3adant2 simp2 syl5eqel
+    fundmfi syl2anc ) ABZACZDZAEFZGZAHRIZEAJUAREFZSUBEFQTUCSAKLQSTMROPN $.
+
+  $( A 1-1 function (class) with a set as domain is finite if and only if its
```
benjub, line 61084: > An injective function whose domain is a set is finite if and only if its range is finite.  See also ~ f1vrnfibi .

? (ping @avekens )

iset.mm
```diff
@@ -60969,6 +61036,73 @@ the cancellation property (fifth hypothesis), show that the relation
...
+    ( wcel wf1 wa cfn crn wrel ccnv wfun f1rel ad2antlr wf1o f1cnv syl simpr wb
+    wceq f1ofun funrnfi syl3anc cdm cen wbr wi f1dm f1f1orn eleq1 f1oeq2 eqcoms
+    anbi12d biimpd expcomd sylc impcom adantr f1oeng enfii syl2anc f1fun mpbird
+    fundmfibi impbida ) ADEZABCFZGZCHEZCIZHEZVHVIGCJZCKZLZVIVKVGVLVFVIABCMNVGVN
+    VFVIVGVJAVMOVNABCPVJAVMUAQNVHVIRCUBUCVHVKGZVICUDZHEZVOVKVPVJUEUFZVQVHVKRVOV
+    PDEZVPVJCOZGZVRVHWAVKVGVFWAVGVPATZAVJCOZVFWAUGABCUHABCUIWBVFWCWAWBVFWCGZWAW
+    DWASAVPAVPTVFVSWCVTAVPDUJAVPVJCUKUMULUNUOUPUQURVPVJDCUSQVPVJUTVAVOCLZVIVQSV
+    GWEVFVKABCVBNCVDQVCVE $.
+
+  $( A 1-1 function (set) is finite if and only if its range is finite.
```
benjub, line 61097: > An injective function which is a set is finite if and only if its range is finite.  See also ~ f1dmvrnfibi .

? (ping @avekens )

benjub (commented):
The first two comments do not match the statements. One question about unsnfi. The other comments are suggestions.

iset.mm
avekens: OK, I will change this in set.mm to "Restricting a relation to its domain without one element of the domain is
     the same as restricting the relation to the universe without this element.".

iset.mm
avekens: ... because A is also a set. But maybe this can be removed.

iset.mm
benjub: I can't see: is there a $e-hypothesis saying that X e. dom R ? If yes, that's fine; if no, just remove "of the domain".

iset.mm
avekens: I will change this to "A one-to-one function whose domain is a set is finite if and only if its range is finite. See also ~ f1vrnfibi ." in set.mm. I think "one-to-one function" is more commonly used in set.mm than "injective function".

iset.mm
benjub: I would remove it, since the antecedent is simply "Fun F".

iset.mm
avekens: I will change this to "A one-to-one function which is a set is finite if and only if its range is finite. See also ~ f1dmvrnfibi ." in set.mm.

avekens (approved):

iset.mm
avekens: I cannot find fundmfi in set.mm anymore. Maybe it was removed in favour of ~fundmfibi.

iset.mm
avekens: OK, "of the domain" can be removed.

iset.mm
jkingdon: > Classically, you do not need B e. V nor -. B e. A . . .
>  Do you need both intuitionistically ? 

Your intuition is perhaps mapping more to "subfinite" than "finite" (and those are equivalent given excluded middle) - there's a lot more on this in section 8.1 of [AczelRathjen] or http://nlab-pages.s3.us-east-2.amazonaws.com/nlab/show/finite%20set for example.

In this example, to say `( A u. { B } ) e. Fin` means you are mapping `( A u. { B } )` to a natural number, and whether that number is the size of ` A ` or the size of ` A ` plus one will tell which case we are dealing with. I think that could be said as `( A u. { B } ) e. Fin -> DECID ( B e. _V /\ B e. A )` or if not something along those lines.

There's at least a bit of a meta-point too - the current proof does use both (`-. B e. A` to apply `unen` and `B e. V` to apply `en2sn`). It is interesting to ponder whether a result can be strengthened but we will sometimes prove a result which is not as strong as what might be possible later (although this is sort of a point for another day if the reasoning above, about why the `B e. _V ` and ` B e. A` conditions are needed, holds up).

> (which you could join as B e. ( V \ A )).

That is what I originally wrote but I was thinking that the point of writing `B e. V` rather than `B e. _V` is that the former is often already present (in the form of `B e. _om` or whatever) and that having to break apart the `( V \ A )` may just end up an inconvenience whenever this is applied. We can always change it later, or have both forms, if it seems like `B e. ( V \ A )` is often a natural way to express this.

> Maybe add also to iset: $p |- ( [what is needed] -> { A } e. Fin )

If I understand the question this is http://us.metamath.org/ileuni/snfig.html . I mean we could also prove ` -. A e. _V -> { A } e. Fin` but that's pretty much just a restatement of http://us.metamath.org/ileuni/0fin.html and http://us.metamath.org/ileuni/snprc.html and it isn't like we could use http://us.metamath.org/mpeuni/pm2.61.html to combine the `A e. _V` and `-. A e. _V` cases so I'm not sure there's any particular reason to add the proper class version too.

iset.mm
jkingdon: Ooh good catch. I tried to update these when I added conditions but I obviously missed this one.

iset.mm
jkingdon: Although it is true that this theorem doesn't apply for a function which is a proper class (because both sides of the biconditional imply the function is a set), sounds like we are agreed that there isn't any particular need to emphasize that and we can drop "(set)" from the comment. (Plus if we did want to emphasize it, I think we'd want a more explicit wording).

As for `fundmfi`, it is added in this pull request and corresponds to http://us.metamath.org/mpeuni/dmfi.html in set.mm (as noted in the mmil.html entry

```
<TR>
  <TD>dmfi</TD>
  <TD>~ fundmfi</TD>
</TR>
```

which is part of this pull request).

iset.mm
jkingdon: Yeah I agree that "distributes over" is pretty far from what is happening here.

I think "When restricting a relation, intersecting with the domain of the relation has no effect." is probably about the right balance of conveying the idea without getting too bogged down. These comments are (generally speaking) designed to be read in conjunction with the formal statement of the theorem rather than being something you'd read instead of the theorem.

I'll make that change in both set.mm and iset.mm.

iset.mm
benjub: > `( A u. { B } ) e. Fin -> DECID ( B e. _V /\ B e. A )` or if not something along those lines.

That would be a nice addition (you probably meant -. B e. A). I do not know if it's true (because each conjunct acts in a differnet way on the size of `( A u. { B } )`). An easier goal might be ` B e. V -> ( ( A u. { B } ) e. Fin -> DECID B e. A )`

iset.mm
benjub: > there isn't any particular need to emphasize that and we can drop "(set)" from the comment. (Plus if we did want to emphasize it, I think we'd want a more explicit wording).

I agree with both remarks.

iset.mm
jkingdon: I'll put this wording ("Restricting a relation to its domain without one element is the same as restricting the relation to the universe without this element.") into this pull request unless there is some reason you'd rather do so yourself.

jkingdon: Hmm, on second thought I might be better off letting you make these changes to set.mm in #2475 and I can grab the wording from that branch for iset.mm. I'll try to do that within the next few hours.

avekens: Yes, I changed it already in set.mm with #2475, so please do not change it (again) to avoid merge conflicts.

jkingdon: Yeah, if we both make the exact same change git will just merge it without a conflict. If I make an almost-same change we'd need to resolve conflicts (which isn't the end of the world - at least if I do the resolution I'm not super scared of that task - but to make it even easier I'll let you make the changes in set.mm and I'll try to carry those over into iset.mm).

iset.mm
jkingdon: I'll use the comment from #2475

iset.mm
jkingdon: I'll use this version of the comment.

jkingdon:
Looked to me like both #2474 and #2475 were ready to merge. If we find any lingering wording fixes or other issues we can make follow up pull requests which is probably easier than trying to keep doing revisions on separate branches.

iset.mm
digama0: I don't think this is true. Counterexample: let `X = { x e. 1o | p }` as usual and then let `A = X` and `B = (/)`. Then `(A u. {B}) = 1o` which is finite but `B e. A` is equivalent to `p`.

I believe it is true if you assume `A` is also finite, because as Jim said you can case on whether `|A| = |(A u. {B})|` or `|A| + 1 = |(A u. {B})|`.

jkingdon: > I don't think this is true. Counterexample: let `X = { x e. 1o | p }` as usual and then let `A = X` and `B = (/)`. Then `(A u. {B}) = 1o` which is finite but `B e. A` is equivalent to `p`.
> 
> I believe it is true if you assume `A` is also finite, because as Jim said you can case on whether `|A| = |(A u. {B})|` or `|A| + 1 = |(A u. {B})|`.

I'll make a follow up pull request for this, but you are right, we weren't being very explicit. I did indeed intend the case where `A e. Fin`.

jkingdon: Follow up pull request is https://github.com/metamath/set.mm/pull/2477

PR #2477, opened 2022-02-07
Add unsnfidcex and unsnfidcel to iset.mm

jkingdon:
This is perhaps a bit of a curiosity but it perhaps sheds at least some
light on why unsnfi is written the way it is. This is based on discussion with @benjub and @digama0 at https://github.com/metamath/set.mm/pull/2474#discussion_r800202578 .

It may be possible to strengthen these results, but after spending most of a day on it and not finding it was something I could just knock out, this is about as far as I'm likely to go for the near future. Perhaps someone else will want to play around with it more.


avekens (approved):

benjub (approved):
Nice results.  The comments sound strange, though (what is "it" in "it would need to be strengthened")?.

jkingdon:
> Nice results. The comments sound strange, though (what is "it" in "it would need to be strengthened")?.

Same as "this" earlier in the sentence.

At least to my eyes the biggest problem was not wording as much as that I was having trouble precisely stating what these do and do not prove. But I think that is part of the scope of "Perhaps someone else will want to play around with it more."

benjub:
Ok... still a bit hard, but I understand the sentence now !  I'll try to see but from Mario, your and my posts in the previous discussion, its looks hard to prove more... Perhaps strengthen `DECID -. B e. _V` to  `DECID B e. _V` ?  I'll  give it a try if I find some time.

jkingdon:
> its looks hard to prove more... Perhaps strengthen `DECID -. B e. _V` to `DECID B e. _V` ? I'll give it a try if I find some time.

Yeah, that's what I started with but didn't see how to prove it. Like we'd need to show that if A and A u. { B } are not equinumerous then { B } is inhabited or something like that.

There's also the question of whether any of this DECID stuff fully shows those arguments are necessary. Because if the theorem held with one antecedent removed, that removed proposition would be true (in the presence of the other antecedents), and a true statement is decidable.

Issue #2469, opened 2022-02-02
Moving 4 txp-related items of Scott Fenton to the main set.mm


mazsa:
Namely ctxp , df-txp , txprel and txpss3v .
Related relevant conversation is https://github.com/metamath/set.mm/pull/2435 , summary in chronological order:
@sctfn is not against of moving my theorems (impl. his 4 theorems above) to the main: "Personally, I have no issues moving these theorems to main. They're a bit esoteric, but they're perfectly valid."
@jkingdon  (on the 4 theorems above) : "I don't have an especially strong reaction one way or the other although perhaps I would if I knew this material better." 
@avekens : "This may be useful in general - to be discussed, if df-txp (together with all/some related theorems) should be moved to main. In my opinion, nothing speaks against it." and "A Github issue should be opened for a discussion about moving it to main later."
@benjub : "df-txp: I had never heard of the "tail cross"... The "variable free" definitions it permits are interesting, but I'm not sure the range of application justifies its moving to Main." 
(+ I found them usable and very useful in 62 of my theorems not yet submitted to Github.)

avekens:
I searched for "tail Cartesian product" and "tail cross" in Google, but I did not find anything except the metamath pages with Scott's definition and theorems. Is there any poublic reference (book, website) available? First I want to see how this construct is used by others before I can decide if it is useful to be provided in main set.mm.

mazsa:
@avekens I don't know the motivations of @sctfn or about prior use, but this is how I used it:

The first part of my main results is the reproduction of the conventional meaning of partition (e.g. partition ` A ` of  ` X `, [Halmos] p. 28: "A partition of ` X ` is a disjoint collection ` A ` of non-empty subsets of ` X ` whose union is ` X ` ", or Definition 35, [Suppes] p. 83., cf. https://oeis.org/A000110 ) :
mpet3 $p |- ( ( ElDisj A /\ -. (/) e. A ) <-> ( EqvRel ~ A /\ ( U. A /. ~ A ) = A ) )
mpet2 $p |- ( ( `' _E |` A ) Part A <-> ,~ ( `' _E |` A ) ErALT A )
mpet $p |- ( MembPart A <-> MembEr A )
where
df-coels $a |- ~ A = ,~ ( `' _E |` A ) and 
brcoels $p |- ( ( B e. V /\ C e. W ) -> ( B ~ A C <-> E. u e. A ( B e. u /\ C e. u ) ) )

[...]

PR #2478, opened 2022-02-07
4 range cross items by Scott Fenton in the Mathbox of Peter Mazsa

mazsa:
Implementing the output of https://github.com/metamath/set.mm/issues/2469 . 
Range cross is almost the same as tail cross, differences: 
name is now range Cartesian product / range cross (vs. tail Cartesian product / tail cross), 
range cross is in my Mathbox (tail cross remains in Scott´s Mathbox),
the labels are ` *rxp* ` (vs ` *txp* ` ) 
reference to Holmes (like in Scott´s http://us2.metamath.org/nfeuni/df-txp.html ), 
` X.. ` in set.mm (vs. ` (x) ` in set.mm ).
If you are dissatisfied with the markings, please tell me now.

### DIFF
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -5566,6 +5566,12 @@ Sol&egrave;r,&quot; <I>Bull.  Am.  Math.  Soc.</I> 32:205-234 (1995)
 http://arxiv.org/abs/math/9504224v1</A> (retrieved 11 Nov 2014).
 </LI>
 <LI>
+<A NAME="Holmes"></A> [Holmes] Holmes, Robert, <I>Elementary Set Theory With a
+Universal Set</I>; available at
+<A HREF="https://randall-holmes.github.io/head.pdf">
+https://randall-holmes.github.io/head.pdf</A> (retrieved 7 Feb 2022).
+</LI>
+<LI>
 <A NAME="Huneke"></A> [Huneke] Huneke, Craig L.,
 &quot;The Friendship Theorem&quot; <I>American Mathematical Monthly</I>
 109:192-194 (2002); available at <A


jkingdon (approved):

avekens (approved):

benjub (commented):
Since it's in your mathbox, I do not oppose merging.

A better symbol for the range cartesian product could be $\times_\mathrm{ran}$, and the token, for instance, Xran.  The use of the tensor product symbol is misleading.

"range cross product" and "range cross class" could be removed.

relationship --> relation (in this case)

Could you add ~ links from your comments to @sctfn's results ?

As for ~ rxpss3v, I wouldn't say "ordered triples", but maybe a more precise description would be too wordy.

mazsa:
Benoit @benjub 
 > "range cross product" and "range cross class" could be removed.

done

>  relationship --> relation (in this case)

done

>  Could you add ~ links from your comments to [@sctfn](https://github.com/sctfn)'s results ?

Done, thank you

>  As for ~ rxpss3v, I wouldn't say "ordered triples", but maybe a more precise description would be too wordy.

Yes, ordered triples are ( ( _V X. _V ) X. _V ) , not ( _V X. ( _V X. _V ) ). However, the latter are ordered and are triples as well. Hm… do you have a suggestion here or should I leave it as it is?
 
>  (...) the token, for instance, Xran.

I changed xran to xrn (XRN) because of our token conventions, otherwise accepted and done.

>  The use of the tensor product symbol is misleading.

Yes, the tensor product symbol is not optimal, see below.

>  Since it's in your mathbox, I do not oppose merging.

I think we are very close now, it seems to me that I understand your approach in broad terms (check below), and it is not incommensurable with mine, I’ll explain below. So my aim is here to persuade you that my point of view is acceptable to you so you could merge it even to the main set.mm as well.

>  A better symbol for the range cartesian product could be $\times_\mathrm{ran}$

This is the only remaining problem. I’ll explain my point of view:
*Suppose* you have a concept  ` X ` (I don’t understand fully yet but) I model it as follows:
` X(R,S,...;M,N,...) ` is the general form of the concept  ` X `
` X(R,S;2,2) ` is pprod
` X(R,S;1,2) ` is txp or now: xrn, and
` X('' _E,'' _E;1,1) ` is xp .
Details (the number of dimensions, etc.) are not important. All I say is that 
just like we simplify the symbol ` X('' _E,'' _E;1,1) ` to   ` X. ` ,
we should simplify the symbol ` X(R,S;1,2) ` to something as well. It should keep the original logic of the concept ` X(R,S,...;M,N,...) ` but *at the same time* it should be as simple as possible  (because it has a central place in my ~ pet theorem and I must be able to explain this concept at *high school* level, without the need to explain the general concept:).
Which is the most simple symbol possible for this  ` X(R,S;1,2) ` without oversimplification? It should refer to  ` X ` and it should contain R and S. 1 and 2 are very special numbers so we may leave them out, maybe we should refer to the fact that 1 and 2 are not the same, i.e., that ` X(R,S;1,2) ` is asymmetric in a way. So the form of this simplified symbol should be this:
 ` ( R 1asymmetriccharacter S ) `
This 1asymmetriccharacter can not be the tensor product symbol: while it refers to ` X ` , it is not asymmetric, and, as you wrote, misleading.
It can not be \times_2 , because what would then be \times_R ?
It can not be \times_R , because R here is not range but something general which is misleading.
And it can not be \times_\mathrm{ran} because it has 4 characters and attracts a high level of abstraction which certainly does not belong to the targeted high school level. They should think on this concept exclusively as the class form of the br version ( x R y /\ x S z ) , and I should help them with the simplest possible form of the class.
I suggest that if you have this  ` X ` in your mind, and you think it is so important that you explicitly define it and show us, then we’ll have to define
df-xpNEW  ` X. ` as ` X('' _E,'' _E;1,1) ` , and
df-xrnNEW  ` ( R 1asymmetriccharacter S ) ` as ` X(R,S;1,2) ` .
Until that day, we have to stick to df-xp and Scott’s txp/xrn definition.
I suggest this 1asymmetriccharacter should be \ltimes :  http://us.metamath.org/symbols/ltimes.png .

benjub (approved):

benjub:
That's the standard symbol for (left) semidirect product.  But well, your mathbox, your choice of symbol (at least, it is asymmetric, which is indeed important).
Approved and ready to merge.

PR #2479, opened 2022-02-07
mathbox cleanup: 4 lemmas renamed according to convention

glacode:
A minimal cleanup:

~ xlimmnflemim1 and ~ xlimmnflemim2 were actually lemmas for xlimmnfv

~ xlimpnflemim1 and ~ xlimpnflemim2 were actually lemmas for xlimpnfv

Thus, they've been renamed to

~ xlimmnfvlem1 , ~ xlimmnfvlem2 , ~ xlimpnfvlem1 , ~ xlimpnfvlem2

according to our convention.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15600,6 +15600,7 @@ New usage of "fldhmsubcALTV" is discouraged (0 uses).
 New usage of "fnexALT" is discouraged (0 uses).
 New usage of "foco2OLD" is discouraged (0 uses).
 New usage of "fprodcom2OLD" is discouraged (0 uses).
+New usage of "frgrwopreglem5ALT" is discouraged (0 uses).
 New usage of "fsumcom2OLD" is discouraged (0 uses).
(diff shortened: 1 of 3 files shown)


avekens (changes requested):
If these theorems are actually lemmas only, this should be reflected in their comments. This means that the comments should start with "Lemma for ~ xlimmnfv : ...". This is used to show or suppress lemmas in theorem lists (see option /SHOW_LEMMAS for WRITE THEOREM_LIST).

benjub:
I agree with @avekens. And they should be labeled and commented as "lemmas" only if they have no interest in themselves (that is, if nothing would be lost if they were simply "inlined" in the proofs of the theorems for which they are lemmas).

jkingdon (approved):

benjub:
Merging. Hopefully @glacode can implement @avekens's remarks in a near-future PR.

PR #2480, opened 2022-02-08
Remove unnecessary braces from iset.mm and set.mm (part 3).

jamesjer:
This covers the "REAL AND COMPLEX NUMBERS" sections of both set.mm and iset.mm.


benjub (approved):
I'm wondering if there is a way to have a smaller diff (after all, the "relevant" diffs are only the removal of whitespace, ${ and $} ). A way would be to have a diff on the set.mm versions obtained after typing "set width 9999" in ./metamath.
But since @jamesjer's script is reliable, that's not a big deal.

avekens (approved):

PR #2462, opened 2022-01-30
Preparatory theorems 2/2

mazsa:
The remaining Preparatory theorems (which don't need new definitions). Changed references to theorems not yet in my Mathbox from ` ~ ` to ` T ` and now to ` T_ ` in order to prevent reference and rewrapping issues revealed by ` verify markup ` .

PR #2483, opened 2022-02-09
remove extraneous hypotheses in br1steq and br2ndeq

### LINKED ISSUE #2481: br1steq br2ndeq: implementing the solution of Alexander by Benoit
Although the contribution date is before the contribution date of the current theorems ~br1steqg and ~br2ndeqg, the contribution itself is later than the current one (first come, first serve). Therefore, Scott Fenton should be kept as contributor. Removing the condition C e. X, however, is a good idea/result, therefore I propose to replace the current theorems by these ones, marking the current theorems as obsolete (using suffix OLD), and tagging them with "(Contributed by Scott Fenton, 2-Jul-2020.)  (Revised by Peter Mazsa, 30-Jan-2022.)"

_Originally posted by @avekens in https://github.com/metamath/set.mm/pull/2462#discussion_r795163713_

benjub:
Remove extraneous hypotheses in br1steq and br2ndeq, and adapt the proofs of the theorems using them.  This shortens slightly some proofs. This is a follow-up to the discusion in #2462.  Actually, @sctfn's original proofs did not make real use of the extraneous hypotheses, so I think the credits should remain unchanged.

@mazsa : it appears that you "rediscovered independently" the proofs (and maybe even before, but what is important is the date when the result was made public).  I know from experience that it can be frustrating sometimes.

@avekens , @mazsa : the hypothesis that C is a set can also be removed from br1steqg/br2ndeqg.  This is obvious since both sides of the biconditional in the consequent imply that C is a set (given the antecedent).  This gives a method to revise the proof, though there may be a shorter way.  Therefore, the person adapting the proof should appear in a `Revised to remove sethood hypothesis on C.  (Revised by XX, ...)` credit.  Who wants to give it a try ?

After that, it could be moved to Main.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14360,6 +14360,8 @@ New usage of "bnnv" is discouraged (7 uses).
 New usage of "bnrel" is discouraged (1 uses).
 New usage of "bnsscmcl" is discouraged (0 uses).
 New usage of "bocardo" is discouraged (0 uses).
+New usage of "br1steqgOLD" is discouraged (0 uses).
+New usage of "br2ndeqgOLD" is discouraged (0 uses).
 New usage of "bra0" is discouraged (1 uses).
 New usage of "bra11" is discouraged (6 uses).
 New usage of "braadd" is discouraged (1 uses).
@@ -18642,6 +18644,8 @@ Proof modification of "bj-zfpow" is discouraged (71 steps).
 Proof modification of "bnj142OLD" is discouraged (51 steps).
 Proof modification of "bnj145OLD" is discouraged (112 steps).
 Proof modification of "bnj538OLD" is discouraged (94 steps).
+Proof modification of "br1steqgOLD" is discouraged (134 steps).
+Proof modification of "br2ndeqgOLD" is discouraged (134 steps).
 Proof modification of "brfi1indALT" is discouraged (789 steps).
 Proof modification of "brfi1indALTOLD" is discouraged (789 steps).
 Proof modification of "brfi1indOLD" is discouraged (48 steps).


avekens (approved):

mazsa:
@benjub @avekens @sctfn I only have the general version (cf. https://github.com/metamath/set.mm/pull/2457 for explanation), if these minimized proofs below are ok with you ( = nobody is against them / nobody will submit a shorter one, say until tomorrow 11 a.m. CET), I'll open a PR with this content (17-Jan-2022 was the date when I published them in the attachment https://groups.google.com/g/metamath/c/EQUGHDUCi2M/ ) :

  $( Uniqueness condition for binary relationship over the ` 1st ` relationship.  Revised to remove sethood hypothesis on ` C ` .  (Contributed by Scott Fenton, 2-Jul-2020.) (Revised by Peter Mazsa, 17-Jan-2022.) $)
  br1steqg $p |- ( ( A e. V /\ B e. W ) -> ( <. A , B >. 1st C <-> C = A ) ) $=
    ( wcel wa cop c1st cfv wceq wbr op1stg eqeq1d cvv wfn wb wfo fo1st fofn
    ax-mp opex fnbrfvb mp2an eqcom 3bitr3g ) ADFBEFGZABHZIJZCKZACKUHCILZCAKUGUI
    ACABDEMNIOPZUHOFUJUKQOOIRULSOOITUAABUBOUHCIUCUDACUEUF $.

  $( Uniqueness condition for binary relationship over the ` 2nd ` relationship.  Revised to remove sethood hypothesis on ` C ` .  (Contributed by Scott Fenton, 2-Jul-2020.) (Revised by Peter Mazsa, 17-Jan-2022.) $)
  br2ndeqg $p |- ( ( A e. V /\ B e. W ) -> ( <. A , B >. 2nd C <-> C = B ) ) $=
    ( wcel wa cop c2nd cfv wceq wbr op2ndg eqeq1d cvv wfn wb wfo fo2nd fofn
    ax-mp opex fnbrfvb mp2an eqcom 3bitr3g ) ADFBEFGZABHZIJZCKZBCKUHCILZCBKUGUI
    BCABDEMNIOPZUHOFUJUKQOOIRULSOOITUAABUBOUHCIUCUDBCUEUF $.

sctfn:
I have no objection to this. Nice work!

benjub:
@mazsa : nice !  When you add them, relabel the existing proofs to br1steqgOLD and br2ndeqgOLD, add to each both discouragement tags, and modify the comment to: "Obsolete proof of ~ br1steqg as of 10-Feb-2022.  (Contr..." Then, you'll have to regenerate the "discouraged" file as indicated in https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md

benjub:
@mazsa : Actually, I'm going to add your versions now, it's going to be easier.  I'll add them in your mathbox, but only for a few days since we will move them to Main very soon.

**Edit**: I'll move the new versions next to the OLD ones in @sctfn 's mathbox since it appears before yours (so you can use it in your mathbox).  This will only stay like this for a few days, to be able to compare them on the webpage.

mazsa:
@benjub perfect, thank you. Please do this "discouraged" labeling etc as well: I was afraid of it anyway:)

benjub:
Ready to merge. @avekens : good for you ?  We'll give ourselves a few days to look at the webpages and see if we move only the *eqg forms of both *eq and *eqg to Main, and where precisely.

benjub:
May I obtain an approval from @avekens or @jkingdon or @sctfn or @tirix ?

sctfn (approved):

avekens (approved):

mazsa:
Benoit @benjub 
1. I think you can remove 
$d A x y z $.  $d B x y z $.  $d C x y z $.
conditions from br1steqg and br2ndeqg (row 510736) as well.
2. If we keep them in a Mathbox, verify markup will fail with
?Warning: The proof of "brxrn" in the Mathbox for Peter Mazsa references "br1steqg" / "br2ndeqg" in the Mathbox for Scott Fenton. 
Should I include in my PR a br1steqgTEMP and br2ndeqgTEMP until they are in their final place in the main set.mm?

benjub:
@mazsa : You're right, the DV conditions are useless since the new proofs have no dummy variables.  I'll remove them when I move the theorems to Main.

Make your changes in your mathbox as if the theorems were already in Main: I'm going to move them very soon (I just want to see the proofs on the website to get a better view), so by the time you do your PR, they will be in Main and "verify markup *" will pass.

mazsa:
@benjub thank you:)

benjub:
The proofs have been updated: http://us2.metamath.org/mpeuni/br1steqg.html

1. It appears that br1steq has a short proof from br1steqg. Same with br2ndeq(g).  @sctfn : can we shorten these proofs in your mathbox, without keeping the old ones ?

2. The proofs of br1steqg and br2ndeqg have a common part that can be proved independently. Actually, both are simple manipulations around
```
     fnbrfvb	|- ( ( F Fn A /\ B e. A ) -> ( ( F ` B ) = C <-> B F C ) )
```
The proof uses a shortcut which I'm not fond of, namely using `1st Fn _V` and `<. A , B >. e. _V` (while I would prefer using `1st Fn ( _V X. _V )` and `<. A , B >. e. ( _V X. _V )`). What do you think, @digama0 ?  I would propose adding a "couple version" of fnbrfvb as
```
     fnbrfvb2	|- ( ( F Fn ( V X. W ) /\ ( A e. V /\ B e. W ) ) ->
                     ( ( F ` <. A , B >. ) = C <-> <. A , B >. F C ) )
```
and prove br1steqg and br2ndeqg (and maybe other theorems) from it. Ping also @avekens.

mazsa:
@benjub I think @sctfn  proved br1steq and br2ndeq with the C e. _V hypothesis, @digama0 had the fnbrfvb solution but did not recognize that C e. _V became superfluous, I just removed C e. _V and wrote it in general form. If you escalate this to a PR, I think Mario should decide whether he approves it.

mazsa:
Benoit @benjib I think this will be a longer decision process: no problem, I'll go on tomorrow with a temporal br1steqT .

benjub:
PR #2500 moves br1steqg and br2ndeqg to Main. I'm still interested in @digama0 and/or @avekens's thoughts on my proposal to prove them from a new
```
fnbrfvb2        |- ( ( F Fn ( V X. W ) /\ ( A e. V /\ B e. W ) ) ->
                     ( ( F ` <. A , B >. ) = C <-> <. A , B >. F C ) )
```
for the reason mentioned above, and in @sctfn's approval to shorten the proofs of br1steq/br2ndeq thanks to br1steqg/br2ndeqg.

sctfn:
You’ve got my approval. Shorten away.

avekens:
> PR #2500 moves br1steqg and br2ndeqg to Main. I'm still interested in @digama0 and/or @avekens's thoughts on my proposal to prove them from a new
> 
> ```
> fnbrfvb2        |- ( ( F Fn ( V X. W ) /\ ( A e. V /\ B e. W ) ) ->
>                      ( ( F ` <. A , B >. ) = C <-> <. A , B >. F C ) )
> ```
> 
> for the reason mentioned above, and in @sctfn's approval to shorten the proofs of br1steq/br2ndeq thanks to br1steqg/br2ndeqg.

I agree with Benoît's suggestion.

mazsa:
@avekens @benoit just about their form, not their content:

fnbrfvb |- ( A e. V -> 
                  ( F Fn V -> 
                    ( ( F ` A ) = B <-> A F B ) ) )

fnbrfvb2 |- ( ( A e. V /\ B e. W )  -> 
                    ( F Fn ( V X. W ) -> 
                      ( ( F ` <. A , B >. ) = C <-> <. A , B >. F C ) ) )

mazsa:
@avekens @benoit @sctfn and perhaps
 
br1steqg $p |- ( ( A e. V /\ B e. W ) -> ( <. A , B >. 1st C <-> A = C ) )

br2ndeqg $p |- ( ( A e. V /\ B e. W ) -> ( <. A , B >. 2nd C <-> B = C ) ) 

by removing eqcom .

benjub:
@mazsa : it's more natural to have "C = A": "C is the image of <A,B> under the map 1st if and only if C is equal to A.

As for antecedents, we generally prefer the uncurried form (that is, ( . /\ . ) -> . over . -> ( . -> . )  ) because it turns out to be more readily applicable.

benjub:
Ah, I was tricked by the encoding of set.mm ! It is not true that ` 1st Fn ( _V X. _V ) ` , but one has ` 1st Fn _V ` ! It seems unnatural and I've already ranted about this kind of things enough... The reason of the set.mm encoding is that it allows for more shortcuts (a thing which I do not like but which is apparently appreciated by people who do more formalizations than me).  I proved fnbrfvb2 locally and will probably push it, but it will not be easily applicable to br1steqg and br2ndeqg, so I think they'll stay unchanged for the time being.

tirix:
Since we have `~1stnpr` and `~2ndnpr`:
```metamath
1stnpr $p |- ( -. A e. ( _V X. _V ) -> ( 1st ` A ) = (/) ) $.
2ndnpr $p |- ( -. A e. ( _V X. _V ) -> ( 2nd ` A ) = (/) ) $.
```
The behaviour seems to be exactly the same as `1st Fn ( _V X. _V)`, so I don't really see why we could not change that definition. Could you point me to that discussion?

benjub:
@tirix I haven't looked into the details, and do not know how much work the change would imply.  Since out-of-domain values are set to (/), the behavior would indeed be the same, but maybe some theorems, like fo1st: `1st : _V -onto-> _V` would need to be changed; I do not know how many of them. Indeed, my preference would be, e.g.,
`df-1st $p |- 1st = ( x e. ( _V X. _V ) |-> U. dom { x } )`
`fo1st $p |- 1st : ( _V X. _V ) -onto-> _V`
etc.
This may also make some proofs longer ? Some theorems have a hypothesis "X e. dom F" and we have existence theorems of the form X e. _V for basically any construction (labeled *ex), so we could apply them with F<-1st. Having to prove X e. (_V X. _V) would typically take a few more steps.

I'm interested in your thoughts, and maybe @digama0 as well.

jkingdon:
> I haven't looked into the details, and do not know how much work the change would imply.

A lot of places would need to be touched (`fo1st` has 33 usages), although I don't have especially good intuition about whether the changes would typically be simple or whether there would be a lot of case elimination on `A e. ( _V X. _V )` vs ` -. A e. ( _V X. _V ) ` which would presumably add steps.

A few other existing theorems don't necessarily tell us whether to change anything but perhaps show how things work currently:

```
f1stres $p |- ( 1st |` ( A X. B ) ) : ( A X. B ) --> A
df1st2 $p |- { <. <. x , y >. , z >. | z = x } = ( 1st |` ( _V X. _V ) )
```

sctfn:
I’ve thought about this for a while, and my preference would be:

1st = ( x e. _V , y e. _V |-> x ) 

That way, we completely avoid depending on the definition of an ordered pair.

jkingdon:
> 1st = ( x e. _V , y e. _V |-> x )

We have this in the form of http://us.metamath.org/mpeuni/df1stres.html which is `` |- ( 1st |` ( A X. B ) ) = ( x e. A , y e. B |-> x ) ``

sctfn:
Sorry, I meant I’d like that as the official definition. The old definition is an artifact of our previous definition of ordered pair and I’ve never been comfortable with its “out of band” behavior

benjub:
It's important to have both forms, but I think the definition should be
```
$a |- 1st = ( x e. ( _V X. _V ) |-> U. dom { x } ) $.
```
The RHS is not very important, since one will prove `( 1st `` <. A , B >. ) = A` under sethood hypotheses very soon after that, and then only use the latter.
The other definition `1st = ( x e. _V , y e. _V |-> x )` can be derived.  I understand that it may make more sense when viewing 1st as "the binary function which returns the first argument", which may be used more in computer science. But if you see 1st as "the projection on the first factor", I prefer the one above, since the "modern", or category theoretic definition of products is: the *product* of two objects A and B is a third object (denoted A X. B) *together with* two morphisms pr_1 : A X. B -> A and pr_2 : A X. B -> B such that...

**Edit**: to be clear: what is important is to have these two forms, the "value" theorem, as well as the functional statements I stated in the above paragraph.  Then, the choice of the "official" definition is secondary.

sctfn:
I guess I'm uncomfortable defining pr_1 in terms of domain, which in a
category setting would be ( 1st " A ).

benjub:
I understand what you mean. dom A = (  1st " A ) is indeed a fundamental
result, but note that the direct image is currently defined in terms of
dom. Anyway, what's important is that we agree, as we do, on the "public
interface" they should have, and then implementation can be changed without
too many repercussions.

Le mer. 16 févr. 2022 à 18:22, Scott Fenton ***@***.***> a
écrit :

> I guess I'm uncomfortable defining pr_1 in terms of domain, which in a
> category setting would be ( 1st " A ).
>

PR #2485, opened 2022-02-10
set.mm: double blank lines before each section

benjub:
(following https://github.com/metamath/set.mm/issues/2262#issuecomment-1034345208)


digama0:
@wlammen While checking this change I noticed that you have a section "Other stuff" which is followed immediately by a comment using markup syntax. Is that supposed to be a header comment, or are they separated on purpose?

@mazsa Ditto for your "Contents" section. Metamath automatically generates a table of contents based on the section headers, you should not create a separate "contents" header.

digama0 (approved):

benjub:
@digama0 : I see that your "$( !" does not appear at the same place in set.mm and iset.mm (the position for iset.mm seems more logical). Is there a reason?

digama0:
It looks like the difference is that in set.mm the head comment has been split, in order to splice in the `$( Begin $[ set-header.mm $] $)` comment. I guess the first part of the set.mm comment didn't trigger anything funny because it's actually reasonably well formed markup, although it doesn't seem to be intentional. Should the `set-header.mm` module comment go before the header? Otherwise the set.mm comment should probably have the `$( !` treatment on both header comments.

benjub:
I'll implement your second proposal. The part before `$( Begin $[ set-header.mm $] $)` is there because it should stay even after a `MM> write source set.mm/split` command.

wlammen:
`While checking this change I noticed that you have a section "Other stuff" which is followed immediately by a comment using markup syntax. Is that supposed to be a header comment, or are they separated on purpose?`
The intention was to describe the purpose of the following theorems.  In that way it is a heading comment.  This is in my Mathbox.  I don't think I wrote it to appear in some index.

mazsa:
@digama (cc @benjub ) yes but you can not see the generated Contents in the set.mm . And I think  http://us.metamath.org/mpeuni/mmtheorems.html#dtl:20.21 is not permanent link. 2 possible solutions: 

1. renaming the link above to a permanent http://us.metamath.org/mpeuni/mmtheorems.html#mathbox-mazsa then 

$(
                            Mathbox for Peter Mazsa
$)


$( Contents: http://us.metamath.org/mpeuni/mmtheorems.html#mathbox-mazsa $)

or
2.

$(
                            Mathbox for Peter Mazsa
$)


$( Contents:
Notations
Preparatory theorems
Range Cartesian product
$)

By default I'll implement the 2nd.

digama0:
@wlammen If you put markup after the second header decoration but before the end of the comment, it will appear in the generated HTML, for example http://us.metamath.org/mpeuni/mmtheorems259.html#mm25867b . This is true for mathbox sections just as well as other sections. If you are writing that for the purpose of readers of your mathbox (which I presume you are), then it would make sense to put that markup inside the header comment instead of in a separate comment, so that people reading from the web page can also see it.

@mazsa The contents for your mathbox are part of the table of contents here: http://us.metamath.org/mpeuni/mmtheorems.html#dtl:20.21 . In set.mm itself the headings are indicated by the giant decorators (which you can also search for in a text editor to jump around). A manual table of contents (and one that does not match the style which is checked by the tooling) is a maintenance hazard, so I would strongly recommend against it, although I won't tell you what to do here.

mazsa:
@digama0 (cc @benjub ) for me this 3. version is ok as well (I'd like to have *any* kind of Contents under my Mathbox header):

$(
########################################
Mathbox for Peter Mazsa
########################################
$)

$( Contents: http://us.metamath.org/mpeuni/mmtheorems.html#dtl:20.21 $)

the problem is that this may be a maintenance hazard as well because as I see its logic, it is not a permanent link, i.e., before my Mathbox entered set.mm this link pointed to Rodolfo Medina's Mathbox, and after adding a new Mathbox for say Joe Maaa his Mathbox will be under that /mmtheorems.html#dtl:20.21 link. If we had permanent links to Mathboxes, this problem would vanish because I could implement my (1st best) solution /mmtheorems.html#mathbox-mazsa above (+ the reference to this hypothetical permanent link would be an optional answer to the "What have you been up to lately?" question:).

PR #2486, opened 2022-02-11
Add minimum of two real numbers to iset.mm

jkingdon:
This adds most of rest of the the set.mm minimum/maximum theorems (stated to use supremum and infimum as with the existing minimum/maximum theorems).

The key lemma for minimum is `minmax` which I sort of thought would be easy from http://us.metamath.org/ileuni/infrenegsupex.html but it turns out there are a bunch of steps involved in applying infrenegsupex - including for one maxleastlt (in this pull request) which basically states that the maxmimum is a least upper bound, in the form needed by the various supremum theorems which have a hypothesis around the existence of the supremum.

I guess the second hardest thing was the reverse direction of http://us.metamath.org/mpeuni/maxlt.html - after trying a few dead ends I found a proof from http://us.metamath.org/ileuni/maxabs.html and http://us.metamath.org/ileuni/abslt.html and a bunch of rearrangement of expressions.

Unrelatedly, there is a wording fix to the comment of fimaxre2 in both set.mm and iset.mm - since the number that this theorem provides is not an element of the set it is more accurate to call it an upper bound not a maximum.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4777,7 +4777,7 @@ numbers).</TD>
 <TR>
 <TD>xrmax1 , xrmax2 , xrmin1 , xrmin2 , xrmaxeq , xrmineq ,
 xrmaxlt , xrltmin , xrmaxle , xrlemin ,
-maxlt , ltmin , max0sub , ifle</TD>
+ltmin , max0sub , ifle</TD>
 <TD><I>none</I></TD>
 </TR>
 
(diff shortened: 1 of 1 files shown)


benjub (approved):

jkingdon:
Looks like I better merge this before conflicts creep in. Hope this isn't too fast for @benjub or others, but if there's anything else to revise we can always make a follow up pull request.

PR #2487, opened 2022-02-11
Indent section titles with two white spaces

benjub:
(following https://github.com/metamath/set.mm/issues/2262#issuecomment-961951990)

The goal is to get closer to a "normal form" for indentations, blank lines, etc., with the associated benefits, like fewer potential merge conflicts.  This should also be performed by "/rewrap" (as well as double blank lines and other head comment indentations if possible, as well as wrapping section titles if longer than 76 characters), as Norm had suggested in the mentioned comment. Is this easily doable, @digama0 ? Should I create an issue in the metamath-exe repo ? 

I plan to do the same (and also add double blank lines as in #2485) for iset.mm, unless @jkingdon objects, or this might cause conflicts with something he's working on.


jkingdon:
> I plan to do the same (and also add double blank lines as in #2485) for iset.mm, unless @jkingdon objects, or this might cause conflicts with something he's working on.

No objection. I can resolve conflicts if any.

jkingdon (approved):

set.mm
```diff
@@ -19,9 +19,9 @@ Dwight Megill (1950--2021).  For more information, visit
    been continuously enriched since then (list of contributors below). $)
 
 
-$(
+$( !
```
tirix, line 22: I'm coming after the battle here, but I think maybe this exclamation mark was meant so that this comment is not considered a heading, and skipped when building the table of content.

digama0: Yes, this is the followup to https://github.com/metamath/set.mm/pull/2485#issuecomment-1035675818

PR #2488, opened 2022-02-11
Remove unnecessary braces from iset.mm and set.mm (part 4).

jamesjer:
This covers the ELEMENTARY NUMBER THEORY portions of iset.mm and set.mm.  After this, only mathboxes are left in iset.mm.  There is more to do in set.mm, however.


benjub:
Did you investigate after my comment in #2480 to produce a smaller diff file ? Namely, do "set width 9999 && write source set.mm/rewrap" in ./metamath on the before/after versions, and do a diff on these (which should be kind of "character-wise" instead of "line-wise" if that makes sense).

iset.mm
```diff
@@ -106549,23 +106546,21 @@ which defines the set we are taking the supremum of must (a) be true at
       HWAXCRZXIXHXLXMVAAKVCVFUONQUQNVGOVHVIVJVKVL $.
   $}
 
-  ${
-    $( Extend ~ gcdmultiple so ` N ` can be an integer.  (Contributed by Scott
-       Fenton, 18-Apr-2014.)  (Revised by Mario Carneiro, 19-Apr-2014.) $)
-    gcdmultiplez $p |- ( ( M e. NN /\ N e. ZZ ) ->
+  $( Extend ~ gcdmultiple so ` N ` can be an integer.  (Contributed by Scott
+     Fenton, 18-Apr-2014.)  (Revised by Mario Carneiro, 19-Apr-2014.) $)
+  gcdmultiplez $p |- ( ( M e. NN /\ N e. ZZ ) ->
```
benjub, line 106551: formula fits on a single line; can you do it ?

set.mm
```diff
@@ -187363,21 +187356,18 @@ obviously not a bijection (by Cantor's theorem ~ canth2 ), and in fact
...
-      nnzd 3eqtr3d mul01 cn0 nn0gcdid0 pm2.61ne ) ACDZBEDZFZAABGHZIHZAJAAKGHZIH
-      ZAJBKBKJZVKVMAVNVJVLAIBKAGLMNVIBKUAZFZAVJOPZIHZAABOPZGHZIHZVKAVIVRWAJVOVI
-      VQVTAIVIVQAOPZVSGHZVTVGARDZBRDVQWCJVHAUBZBUCABUDUEVGWCVTJVHVGWBAVSGVGAAUF
-      VGAAUKZUGUHUIQSMQVPAEDZVJEDZVRVKJVPAVGVHVOUJVAVIWHVOVGWGVHWHAULABUMUNQVJA
-      UOUPVGVHVOWAAJZVHVOFVGVSCDWIBUQAVSURUSUTVBVIVMAKIHZAVGVMWJJZVHVGWDWKWEWDV
-      LKAIAVCMTQVGWJAJZVHVGAVDDWLWFAVETQSVF $.
-  $}
+  $( Extend ~ gcdmultiple so ` N ` can be an integer.  (Contributed by Scott
+     Fenton, 18-Apr-2014.)  (Revised by Mario Carneiro, 19-Apr-2014.) $)
+  gcdmultiplez $p |- ( ( M e. NN /\ N e. ZZ ) -> ( M gcd ( M x. N ) ) = M ) $=
```
benjub, line 187361: Out of curiosity: was the action of writing the formula on a single line automatic ?

set.mm
```diff
@@ -191951,70 +191935,65 @@ prime number there is always a second nonnegative integer (less than the
...
-      UVAUUTUYAYBFBCYCVFSUUCUXTFKPGFUUCUXSKFPABCYDWNYGYEWRVQUUCUVPUXPUVQUUIVRWA
-      UXRFUUFYFXCWFUUGUUHYHYI $.
-  $}
+  $( Lemma for ~ pythagtrip .  Show that ` C ` and ` B ` are relatively prime
+     under some conditions.  (Contributed by Scott Fenton, 8-Apr-2014.)
+     (Revised by Mario Carneiro, 19-Apr-2014.) $)
+  pythagtriplem3 $p |- ( ( ( A e. NN /\ B e. NN /\ C e. NN ) /\
+     ( ( A ^ 2 ) + ( B ^ 2 ) ) = ( C ^ 2 ) /\
+     ( ( A gcd B ) = 1 /\ -. 2 || A ) ) ->
+     ( B gcd C ) = 1 ) $=
```
benjub, line 191944: can you put the last two lines on a single line ?

set.mm
```diff
@@ -191951,70 +191935,65 @@ prime number there is always a second nonnegative integer (less than the
...
+    WEVPWGVPBCVNVMWTVOXAUEVOVMCTEVNCOVEVDZVFRVPWAXFWEVPWGXHVGRXEXFNLUTEVALVBMXG
+    VHVIWGLVJVKSVL $.
+
+  $( Lemma for ~ pythagtrip .  Show that ` C - B ` and ` C + B ` are relatively
+     prime.  (Contributed by Scott Fenton, 12-Apr-2014.)  (Revised by Mario
+     Carneiro, 19-Apr-2014.) $)
+  pythagtriplem4 $p |- ( ( ( A e. NN /\ B e. NN /\ C e. NN ) /\
+     ( ( A ^ 2 ) + ( B ^ 2 ) ) = ( C ^ 2 ) /\
+     ( ( A gcd B ) = 1 /\ -. 2 || A ) ) ->
+     ( ( C - B ) gcd ( C + B ) ) = 1 ) $=
```
benjub, line 191963: same

benjub (approved):
Can you do the small changes I suggested ?  If no time now, that's no big deal, and this should not delay merging.

set.mm
jkingdon: > can you put the last two lines on a single line ?

I wouldn't be a big fan of that because the first three lines form the antecedent and the last line forms the consequent.

I don't have a super strong opinion about having the line breaks follow the parse tree but I think it is generally better to do so.

iset.mm
jkingdon: If someone is going to do this please do it in both iset.mm and set.mm.

jkingdon (approved):

jkingdon:
> Can you do the small changes I suggested ? If no time now, that's no big deal, and this should not delay merging.

I went ahead and merged this, based on several pull requests being in flight or (perhaps) anticipated soon and a desire to avoid conflicts which are, I suppose, more likely for this kind of textual rearrangement.

jamesjer:
@benjub I did see your comment in https://github.com/metamath/set.mm/pull/2480 but I don't think I understand.  A PR has to modify the text as it is.  I could generate such a diff, but I would have to show it to you separately somehow.  Am I misunderstanding what you want?

@jkingdon You don't need to rush to merge anything I submit.  I'm quite comfortable with git and can rebase on top of conflicting commits if necessary.

iset.mm
jamesjer: Since this PR has been merged, I can sneak that into the next one.  @jkingdon it has already been done in set.mm.  I must have overlooked doing the same for iset.mm.  Sorry about that.

set.mm
jamesjer: No, I did that manually.

benjub:
> @benjub I did see your comment in #2480 but I don't think I understand. A PR has to modify the text as it is. I could generate such a diff, but I would have to show it to you separately somehow. Am I misunderstanding what you want?

Yes, that's it: you could attach a diff to the PR, in addition.  It has been done in the past for large PRs with artificially large diffs, but I admit this is probably overkill in the present case, so don't bother: these PRs are actually clean.

iset.mm
jkingdon: Sneaking it into the next one sounds good. Sorry if I was too eager on merging: although we do have a general project norm of "err on the side of merging something which is mostly good and fixing up the fine points later", there isn't anything absolute about this and perhaps I got a little more worried about the various pull requests in flight than I needed to.

PR #2490, opened 2022-02-12
$e |- A e. _V $. generalized to " $p |- ( A e. V -> "

benjub:
Title and commit messages say it all. This could be used straightforwardly (but painstakingly, unless you can automate it) to shorten the proofs of fsn, fsng, fsn2, fsn2g, and maybe some later proofs.

The common pattern is that in the old versions, xxx is proved first, and then xxxg is proved from it in a routine way, using a theorem of the family of vtoclg.  In the new versions, xxxg is proved first, and then xxx is proved from it by using simply ax-mp, since xxx is the inference associated with xxxg.

I did no hard work: the hard work is in proving the g-versions for the more fundamental results. After that, if one has a proof of yyy from xxx, then the proof of yyyg from xxxg is easily deduced from it by using deduction style proofs.

**Edit**: Note for snssg/snss: the new proof of snssg is longer than the older one, but the combined length of the proofs of snssg and snss decreases.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17783,6 +17783,7 @@ New usage of "snelpwrVD" is discouraged (0 uses).
 New usage of "sneqrgOLD" is discouraged (0 uses).
 New usage of "snexALT" is discouraged (1 uses).
 New usage of "snnexOLD" is discouraged (0 uses).
+New usage of "snssgOLD" is discouraged (0 uses).
 New usage of "snssiALT" is discouraged (0 uses).
 New usage of "snssiALTVD" is discouraged (0 uses).
 New usage of "snssl" is discouraged (0 uses).
@@ -19731,6 +19732,7 @@ Proof modification of "sneqrgOLD" is discouraged (47 steps).
 Proof modification of "snex" is discouraged (64 steps).
 Proof modification of "snexALT" is discouraged (48 steps).
 Proof modification of "snnexOLD" is discouraged (119 steps).
+Proof modification of "snssgOLD" is discouraged (38 steps).
 Proof modification of "snssiALT" is discouraged (40 steps).
 Proof modification of "snssiALTVD" is discouraged (64 steps).
 Proof modification of "snssl" is discouraged (18 steps).
--- a/discouraged
+++ b/discouraged
@@ -14756,6 +14756,8 @@ New usage of "cnvbracl" is discouraged (2 uses).
 New usage of "cnvbramul" is discouraged (1 uses).
 New usage of "cnvbraval" is discouraged (1 uses).
 New usage of "cnvcnvOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


avekens:
It is a good idea to prove the *g theorem first and derive the corresponding theorem with a hypothesis from it, at least for new theorems . We should add this to our conventions. But do we really want to revise all the existing theorems accordingly (at least if both variants are already available)? I think there are a lot more of them. In my opinion, we can leave them as they are for historical reasons. By this, we would avoid the question who is the contributor of  the theorem(s).

By the way, a theorem with a hypothesis ` $e |- A e. _V $. ` is in "inference form" according to the general characterization of theorem forms (see Conventions). Hypotheses/antecedents about sethood, however, seem to be treated differently already (see Conventions: When an inference is converted to a theorem by eliminating an "is a set" hypothesis, we sometimes suffix the closed form with "g" (for "more general") as in ~ uniex vs. ~ uniexg . "). I wonder if theorems with hypotheses  ` $e |- A e. _V $. ` should be generally avoided, at least in the future, unless they are uses in many proofs to shorten them), because of our directive ""Deduction form" is the preferred form for theorems", hypotheses should be in the form ` $e |- ( ph -> A e. V ) $. ` .

benjub:
As for crediting: you're right, and I did it that way in clel2g and snssg, relsn2.  With the latest commit, I also wrote a more faithful crediting for relsng and relsnopg.  The idea is that if the new version is only a small variant, then the contributor should be the same as the contributor of the old version.

As for relsn2, I did the same as was done for alexeqg, where alexeqg was deprecated (see #835 and #1873, and the line in set.mm `1-May-19 alexeq    ---         deleted; use alexeqg`).  In the present case, relsn2 has no usage, so I kept the same label and didn't write the change in the preamble. Tell me if you prefer to label it relsn2g and/or put it in the preamble.

For the rest, I think we agree. As follows from the way they are proved, and since we agree on favoring xxxg over xxx, it is even more important to do so for earlier theorems (that's why I went "all the way down to" clel2g --- and by the way, there was already a clel3g from which clel3 was proved).

For later theorems, and when there are more hypotheses/antecedents, the deduction form should indeed be favored.  Around the theorems I modified, there are many results stated in the three forms (closed form, inference, deduction).

avekens (approved):

benjub:
A concrete example, maybe for a future PR: uniprg is currently proved from unipr in a standard way, using equality axioms and vtocl2g. But one can prove uniprg first (and then unipr from it simply by ax-mp) as follows:
one notices that the $e-hypotheses are used in the proof only by clel3, but there is a clel3g in set.mm. Therefore, replacing in the current proof of uniprg the two occurrences of clel3 with clel3g, and a few other theorems with their "deduction versions", one obtains a proof of uniprg.

PR #2491, opened 2022-02-12
Obsolete

benjub:
After looking at other obsolete theorems, I would propose to move `  bnj142OLD $p |- ( F Fn { A } -> ( u e. F -> u = <. A , ( F `` A ) >. ) ) $=` to Main. It is currently Step 8 in the proof of ~fnsnb (as indicated in the comment of bnj142OLD), but I think it deserves to be "detached" from that proof since it is a natural statement and does not require hypotheses, contrary to ~fnsnb (which could have a "generalized" form where the $e is replaced by an antecedent, by the way). Can I add it ? Maybe as fnsnr (in view of fnsn) ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2116,7 +2116,6 @@
 "bnj1423" is used by "bnj1312".
 "bnj1424" is used by "bnj1423".
 "bnj1424" is used by "bnj1452".
-"bnj142OLD" is used by "bnj145OLD".
 "bnj1436" is used by "bnj1296".
 "bnj1436" is used by "bnj1371".
 "bnj1436" is used by "bnj1374".
@@ -4297,7 +4296,6 @@
 "csbabgOLD" is used by "csbrngOLD".
 "csbabgOLD" is used by "csbrngVD".
 "csbabgOLD" is used by "csbsngVD".
-"csbabgOLD" is used by "csbunigOLD".
 "csbabgOLD" is used by "csbunigVD".
 "csbabgOLD" is used by "csbxpgOLD".
 "csbabgOLD" is used by "csbxpgVD".
@@ -4311,9 +4309,7 @@
 "csbingOLD" is used by "onfrALTlem4VD".
 "csbingOLD" is used by "onfrALTlem5VD".
 "csbopabgALT" is used by "csbcnvgALT".
-"csbresgOLD" is used by "csbima12gALTOLD".
 "csbresgOLD" is used by "csbima12gALTVD".
-"csbrngOLD" is used by "csbima12gALTOLD".
 "csbrngOLD" is used by "csbima12gALTVD".
 "csbxpgOLD" is used by "csbresgOLD".
 "csbxpgOLD" is used by "csbresgVD".
@@ -12063,7 +12059,6 @@
 "sbcalgOLD" is used by "sbcssgVD".
 "sbcalgOLD" is used by "trsbcVD".
 "sbcangOLD" is used by "csbingVD".
-"sbcangOLD" is used by "csbunigOLD".
 "sbcangOLD" is used by "csbunigVD".
 "sbcangOLD" is used by "csbxpgOLD".
 "sbcangOLD" is used by "csbxpgVD".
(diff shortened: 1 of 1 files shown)


benjub:
The reformulated comments were due to @avekens

benjub:
Another comment: I would be inclined to remove the already commented out sbcom3OLD by @giomasce . Any objections ? Or do you want to keep it somewhere else for testing purposes ? (I see that @jkingdon implemented some tests recently)
```
    $( $d x y $.  $d x z $. $)
    $( Substituting ` y ` for ` x ` and then ` z ` for ` y ` is equivalent to
       substituting ` z ` for both ` x ` and ` y ` .  (Contributed by Giovanni
       Mascellani, 8-Apr-2018.)  Obsolete version of ~ sbcom3 as of
       16-Sep-2018.  (New usage is discouraged.)
       (Proof modification is discouraged.)  Note: this proof requires many
       hours to 'save proof', 'prove...', and 'minimize_with...'  The
       discouragements will prevent accidental minimization attempts, etc. that
       would interfere with scripts.  For the time being we are keeping this
       proof as a test case, but it will be deleted eventually. -NM $)
    $(
    [Commented out by NM 28-Aug-2019 to avoid slowdown in proof format
    conversions.  Uncomment it if needed for testing.]
    sbcom3OLD $p |- ( [ z / y ] [ y / x ] ph <-> [ z / x ] [ z / y ] ph ) $=
```

avekens:
> After looking at other obsolete theorems, I would propose to move `` bnj142OLD $p |- ( F Fn { A } -> ( u e. F -> u = <. A , ( F `` A ) >. ) ) $= `` to Main. It is currently Step 8 in the proof of ~fnsnb (as indicated in the comment of bnj142OLD), but I think it deserves to be "detached" from that proof since it is a natural statement and does not require hypotheses, contrary to ~fnsnb (which could have a "generalized" form where the $e is replaced by an antecedent, by the way). Can I add it ? Maybe as fnsnr (in view of fnsn) ?

OK, but it should be changed to ( ( F Fn { A } /\ U e. F ) -> U = <. A , ( F `` A ) >. )

avekens:
> Another comment: I would be inclined to remove the already commented out sbcom3OLD by @giomasce . Any objections ? Or do you want to keep it somewhere else for testing purposes ? (I see that @jkingdon implemented some tests recently)

I think we should keep it as test case. If there is a better place for it, it can be moved.

avekens (approved):

giomasce:
> Another comment: I would be inclined to remove the already commented out sbcom3OLD by @giomasce . Any objections ? Or do you want to keep it somewhere else for testing purposes ? (I see that @jkingdon implemented some tests recently)
> 
> ```
>     $( $d x y $.  $d x z $. $)
>     $( Substituting ` y ` for ` x ` and then ` z ` for ` y ` is equivalent to
>        substituting ` z ` for both ` x ` and ` y ` .  (Contributed by Giovanni
>        Mascellani, 8-Apr-2018.)  Obsolete version of ~ sbcom3 as of
>        16-Sep-2018.  (New usage is discouraged.)
>        (Proof modification is discouraged.)  Note: this proof requires many
>        hours to 'save proof', 'prove...', and 'minimize_with...'  The
>        discouragements will prevent accidental minimization attempts, etc. that
>        would interfere with scripts.  For the time being we are keeping this
>        proof as a test case, but it will be deleted eventually. -NM $)
>     $(
>     [Commented out by NM 28-Aug-2019 to avoid slowdown in proof format
>     conversions.  Uncomment it if needed for testing.]
>     sbcom3OLD $p |- ( [ z / y ] [ y / x ] ph <-> [ z / x ] [ z / y ] ph ) $=
> ```

No objection on my side, thanks for cleaning stuff up!

benjub:
Thanks, @giomasce.  So we can remove it from set.mm if/when we find a better place for it, as @avekens wisely suggests. For instance, in a database dedicated to testing, like the ones @david-a-wheeler and Mel O'Cat (https://github.com/david-a-wheeler/metamath-test/issues/2) and @jkingdon (https://github.com/metamath/metamath-exe/tree/master/tests) have developed.
Merging and leaving it as is for now.

jkingdon:
> For instance, in a database dedicated to testing, like the ones @david-a-wheeler and Mel O'Cat ([david-a-wheeler/metamath-test#2](https://github.com/david-a-wheeler/metamath-test/issues/2)) and @jkingdon (https://github.com/metamath/metamath-exe/tree/master/tests) have developed.

So I looked at this a little bit and tried running at least a few of the commands described. Part of the problem here is that "testing" is a pretty general word. The metamath-exe tests run on every pull request, and to be useful for that they should complete within a short amount of time (I think the existing ones are well under a second each) and have output sufficient to tell us whether the code did the right thing (currently we just store a known good output, which must match the output of the test run verbatim).

As for sbcom3OLD, I guess here are the issues:
* If I uncomment it and run `verify proof sbcom3OLD` I get distinct variable constraint failures. Perhaps this is a case of bitrot (like, renaming a theorem it depends on or something)? I don't think it is supposed to be a test case for distinct variable checking.
* According to the comment what is interesting about this is that it "requires many hours to 'save proof', 'prove...', and 'minimize_with...'". What is unstated is whether this is seen as something which could plausibly be improved or needs to be improved. I mean, I assume there will likely always be proofs which can make the minimizer take a long time (like, unless it were to give up after a certain amount of time or a certain search depth or something).

I guess I'd vote for removing it - maybe putting it in a github issue if you think there is some action which could be taken with it - or maybe not even that especially if there is no obvious action. But whatever it is I don't think it is closely related to "testing" in the sense that I've been working on testing.

Perhaps someone has further thoughts on what they mean by "test case" and what sbcom3OLD might be good for, and that help suggest what a good place, if any, for it would be.

Issue #2401, opened 2021-12-26
Definition, syntax and token for "setting a component of an extensible structure" (df-sets)


benjub:
Digging into my emails with Norm, I found this topic I had discussed with him, Gérard and Mario.  I had made the following proposal, which I still support.

The current definition of `sSet`, which denotes "setting components of an extensible structure",
```
csts $a class sSet $.
df-sets $a |- sSet = ( s e. _V , e e. _V |-> ( ( s |` ( _V \ dom { e } ) ) u. { e } ) ) $.
```
First, `sSet` is too reminiscent of "simplicial set" and should be changed.  Adapting a proposal from Mario, I proposed the following definition:
```
$a class [ B / A ]s S $.
$a |- [ B / A ]s S = ( ( S |` ( _V \ ( A ` ndx ) ) u. { <. ( A ` ndx ) , B >. } ) $.
```
This could be displayed as $[ B / A ]_{mathrm{struct}} S$ and is purposefully reminiscent of the syntax for substitution. I think this renders the reading, human parsing and understanding easier.

It is generally better to define new constructions as class constants, e.g. `class sin` versus `class sin A`. I think this is indeed important whenever the defined term is a mathematical object, like `sin`.  But here, I don't see what would be gained by considering "setting a component of an extensible structure" as a mathematical object.

This change of df-sets would entail some local changes, but, at first sight, nothing too large. Strictly speaking, the current df-sets may set several components at once, but I do not know if this feature is used often, and I think `[ . / . ]s .` can be used twice or thrice in a row anyway, and that this may even be clearer.

What do people think ?

avekens:
I like the proposed new notation setting/sustituting a component of a structure, at least from a display point of view.. Since the second argument is (almost) always an ordered pair (is there any reason/application for something else?) it makes sense to move this into the definition, as proposed. Although the comment of df-sets says "Set one or more components of a structure", I think there is no way to set more than one component simultaneously, at least I found no example for it. Therefore, this should be no problem.

What is still to be considered are the cases for A, B or S being proper classes and not sets (as assured by the current definition) - is the definition still meaningful or at least does not produce harmful results?

The advantage of the current definition is that the basic library theorems (for equality, explicit/implicit class substitution, not free variables, etc.) need not be provided because they are already available by the maps-to notation.

avekens:
PS: an example for sustituting 2 components is ~tngval:
[...]

PR #2492, opened 2022-02-13
mathbox: add df-strset

benjub:
Groundwork related to Issue #2401.


sctfn (approved):

set.mm
```diff
@@ -531439,6 +531445,41 @@ automatically a set (see ~ bj-mooreset ).  (Contributed by BJ,
...
+
+  $( Tokens for component-setting in extensible structures $)
+  $c [s ]s $.
+
+  $( Syntax for component-setting in extensible structures. $)
+  cstrset $a class [s B / A ]s S $.
+
+  $( Component-setting in extensible structures.  Define the extensible
+     structure ` [s A / B ]s S ` , which is like the extensible structure ` S `
+     except that the value ` B ` has been put in the slot ` A ` (replacing the
```
avekens, line 531467: ` [s A / B ]s S `  -> ` [s B / A ]s S ` : as for the other substitutions, the substituent comes first.

avekens (changes requested):

set.mm
benjub: Fixed. It's written correctly at the other places.

PR #2494, opened 2022-02-13
Measures are continuous from below: general case

glacode:
Now that ` ~~>* ` is available, we can state continuity from below for both the bounded and the unbounded case (with a single notion of convergence).

( Changed comments for Lemmas, as suggested in https://github.com/metamath/set.mm/pull/2479 )

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14360,6 +14360,8 @@ New usage of "bnnv" is discouraged (7 uses).
 New usage of "bnrel" is discouraged (1 uses).
 New usage of "bnsscmcl" is discouraged (0 uses).
 New usage of "bocardo" is discouraged (0 uses).
+New usage of "br1steqgOLD" is discouraged (0 uses).
+New usage of "br2ndeqgOLD" is discouraged (0 uses).
 New usage of "bra0" is discouraged (1 uses).
 New usage of "bra11" is discouraged (6 uses).
 New usage of "braadd" is discouraged (1 uses).
@@ -14754,6 +14756,8 @@ New usage of "cnvbracl" is discouraged (2 uses).
 New usage of "cnvbramul" is discouraged (1 uses).
 New usage of "cnvbraval" is discouraged (1 uses).
 New usage of "cnvcnvOLD" is discouraged (0 uses).
+New usage of "cnvsnOLD" is discouraged (0 uses).
+New usage of "cnvsngOLD" is discouraged (0 uses).
 New usage of "cnvssOLD" is discouraged (0 uses).
 New usage of "cnvunop" is discouraged (2 uses).
 New usage of "com3rgbi" is discouraged (1 uses).
@@ -17522,6 +17526,9 @@ New usage of "reglogmul" is discouraged (1 uses).
 New usage of "relopabVD" is discouraged (0 uses).
 New usage of "relopabiALT" is discouraged (0 uses).
 New usage of "relrngo" is discouraged (6 uses).
+New usage of "relsn2OLD" is discouraged (0 uses).
+New usage of "relsnOLD" is discouraged (0 uses).
+New usage of "relsnopOLD" is discouraged (0 uses).
 New usage of "renegclALT" is discouraged (0 uses).
 New usage of "renicax" is discouraged (0 uses).
 New usage of "resfunexgALT" is discouraged (0 uses).
@@ -17776,6 +17783,7 @@ New usage of "snelpwrVD" is discouraged (0 uses).
 New usage of "sneqrgOLD" is discouraged (0 uses).
 New usage of "snexALT" is discouraged (1 uses).
 New usage of "snnexOLD" is discouraged (0 uses).
+New usage of "snssgOLD" is discouraged (0 uses).
 New usage of "snssiALT" is discouraged (0 uses).
 New usage of "snssiALTVD" is discouraged (0 uses).
 New usage of "snssl" is discouraged (0 uses).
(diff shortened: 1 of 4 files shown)


glacode:
I forgot to write that's mathbox only

set.mm
```diff
@@ -659696,6 +659696,123 @@ not even needed (it can be any class).  (Contributed by Glauco
...
+      OVGVLPVCVICUJZUKULIUMAVCDESUNLUALZMZNZVCSUOLZCOZVGTUPZQAVIVQMZNZVISUOLZCO
+      ZVLTUPZQBKWDWGBAWCBFWCBUBUCBWFVLTBWECGBWERUDBTRVNUQUFVOVSWDWBWGVOVRWCABKV
+      QUGUIVOWAWFVGVLTVOVTWECVCVISUOVAURVPUSULJUMVB $.
+  $}
+
+  ${
+    $d A x $.
+    $( An extended real is plus infinity iff it's larger than all real numbers.
+       (Contributed by Glauco Siliprandi, 13-Feb-2022.) $)
+    xrpnf $p |- ( A e. RR* -> ( A = +oo <-> A. x e. RR x <_ A ) ) $=
```
benjub, line 659802: this should probably go to Main if it is not there yet

set.mm
```diff
@@ -698706,6 +698823,125 @@ sequence of increasing measurable sets (with uniformly bounded measure)
...
+    meaiunincf.n $e |- ( ph -> N e. ZZ ) $.
+    meaiunincf.z $e |- Z = ( ZZ>= ` N ) $.
+    meaiunincf.e $e |- ( ph -> E : Z --> dom M ) $.
+    meaiunincf.i $e |- ( ( ph /\ n e. Z ) -> ( E ` n ) C_ ( E ` ( n + 1 ) ) )
+                   $.
+    meaiunincf.x $e |- ( ph -> E. x e. RR A. n e. Z ( M ` ( E ` n ) ) <_ x )
+                  $.
+    meaiunincf.s $e |- S = ( n e. Z |-> ( M ` ( E ` n ) ) ) $.
+    $( Measures are continuous from below (bounded case): if ` E ` is a
+       sequence of non-decreasing measurable sets (with bounded measure) then
```
benjub, line 698840: ... is a nondecreasing sequence of measurable sets with bounded measure, then...
(next statement: replace "bounded" by "finite")

benjub (approved):
Approved with suggestions for some comments.
Is it really necessary to prove each time a version with DV conditions and a version with F/ hypotheses ?  The latter probably suffices, for these more advanced results.

glacode:
> Is it really necessary to prove each time a version with DV conditions
> and a version with F/ hypotheses ? The latter probably suffices, for these more advanced results.

The *v version would become a lemma of the second theorem, thus there would not be a real advantage.

And making *v its own theorem gives the advantage to save an hyp, when dj vars are not an issue.

avekens (approved):

benjub:
Can you see, @glacode, what happens if you "inline" the *v version in the general version ? If the resulting proof is not overly complex, then it might be enough to keep that proof.  In Metamatn, type "MM> help expand" to see how this works.  First, "expand xxxv" and then, "minimize */except xxxv".  Maybe @avekens has something to add, from his experience.

And what about my other suggestions ? Do you prefer that we merge now and deal with them later (if at all) ?

avekens:
> Can you see, @glacode, what happens if you "inline" the *v version in the general version ? If the resulting proof is not overly complex, then it might be enough to keep that proof. In Metamatn, type "MM> help expand" to see how this works. First, "expand xxxv" and then, "minimize */except xxxv". Maybe @avekens has something to add, from his experience.

I use the versions with F/ hypotheses only in special cases, and I prefer to use the versions with DV conditions in general. Therefore, I propose to provide the versions with F/ hypotheses only if they are actually necessary  for a proof. There is no problem if they are provided in a mathbox (if not used yet), but can be moved to main if required.

glacode:
I use whatever shortens a proof.
As an example, day in and day out I use rexlimdv (when I can) and rexlimd (when I have to). rexlimdv gives a shorter proof, but rexlimd is more general.

tirix (approved):

tirix (approved):

tirix:
I'm like @avekens , I try to avoid `\F` as much as possible, and use it only if there is no other choice.
I would not systematically add them.

For example, @glacode , do you have in mind a use for `~meaiuninc3` where they will be needed?
Neither `~meaiuninc` nor `~meaiuninc2` have such non-free conditions.

Anyway, no problem to merge.

benjub:
Since it's mathbox-only, there is little risk of merge conflicts. @glacode do you want to take into account my suggestions above for the comments (finite/bounded measure), and/or @tirix's, or do you prefer that we merge now ?

glacode:
I've already applied changes in the branch of my local repo. Checking diffs would be hard.

I guess the best option is to merge a9f5493

glacode:
> For example, @glacode , do you have in mind a use for `~meaiuninc3` where they will be needed?

@tirix this would deserve a long post on the forum. I'll try to be supershort (I'm sure you can fill in the details):

1. the ` F/_ n E ` condition will immediately be used to easily derive the following form of the theorem (that's the "useful" one)
```metamath
|- ( ( ph /\ n e. Z ) -> B e. dom M )
```
because the main use of the theorem will be something like this example
```metamath
|- ( ph -> ( M ` RR ) = ( ~~>* ` ( k e. Z |-> ( -u k (,) k ) ) ) )
```

2.  |-   ` F/ n ph ` condition makes it easy to use the theorem when ` ph ` contains a  ` A. n e. Z B e. dom M ` subformula and / or a `` A. n e. Z ( E ` n ) C_ ( E ` ( n + 1 ) )  `` subformula (and it will very often be the case)

Couldn't someone do a cbv* in the "caller" ? Sure, but with a single use of the theorem, you will have already added the "code" that I've used to write it once and for all.

As a further note, the next generation of Proof Assistants will be Language Servers, and then it will be "easy" to add a Diagnostic with a QuickFix that invokes the F/_ version of the theorem to fix a dj var constraint violation.

In my opinion, it is underestimated the importance to have theorems with the (reasonable) smallest number of dj constraints (for the newcomer, this is the most difficult part of metamath to grasp; mathematicians consider "non-free w.r.t. a variable" as a given).

> Neither `~meaiuninc` nor `~meaiuninc2` have such non-free conditions.

As you can see, ~ meaiunincf  was in fact part of this PR.

BR
Glauco

PR #2504, opened 2022-02-17
Add sumeq2d to iset.mm

jkingdon:
In general, my progress on #2167 has been going extremely slowly. This pull request is the first time in a long time that I've made enough of an advance that it seems like progress.

The theorem is sumeq2d which is an (approximate) intuitionization of http://us.metamath.org/mpeuni/sumeq2ii.html . Unless I missed something about how this works or could easily work, this requires changing the definition of `sum_` in two ways (relative to what is now in iset.mm):

1. Add an if statement to the finite set case so that the function we pass to `seq` is defined for all ` NN `
2. Add a decidability condition to the series case (basically matches the discussion of decidability in #2167 although here we are adding the decidability to the definition itself rather than merely theorems using it).

The new definition requires changing existing sum theorems and there is also one lemma which is used in the sumeq2dc proof.

Mostly unrelatedly, there's a theorem `frecfun` . Every time I look at the `seq` theorems in iset.mm I think there would be a way to change or remove some of the conditions on them. I still think that is possible, but it never ends up being as easy I think it will be, so if I can make a baby step here and there I do it and `frecfun`, although pretty far from enabling changing much with `seq`, was one thing I was able to prove without much fuss.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6700,6 +6700,14 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>df-sum</TD>
+  <TD>~ df-isum</TD>
+  <TD>Although this defintion is intended to function similarly to
+  the set.mm one, a lot of details have to be changed, especially
+  around decidability, to make sum work.</TD>
+</TR>
+
+<TR>
   <TD>sumex</TD>
   <TD><I>none</I></TD>
   <TD>This will need to replaced by suitable closure theorems.</TD>
@@ -6713,10 +6721,15 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>sumeq2ii</TD>
+  <TD>~ sumeq2d</TD>
+</TR>
+
+<TR>
   <TD>df-prod and theorems using it</TD>
   <TD><I>none</I></TD>
   <TD>To define this, will need to tackle all the issues
-  with ~ df-sum plus some more around, for example, not equal to zero
+  with ~ df-isum plus some more around, for example, not equal to zero
   versus apart from zero</TD>
 </TR>
 


iset.mm
```diff
@@ -103467,68 +103480,122 @@ Cauchy sequence (for example, fixing a rate of convergence as in
...
+       + ... = 1.  (Contributed by Jim Kingdon, 15-Feb-2022.) $)
+    df-isum $a |- sum_ k e. A B = ( iota x ( E. m e. ZZ ( A C_ ( ZZ>= ` m ) /\
+      A. j e. ( ZZ>= ` m ) DECID j e. A /\
+      seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
+        ~~> x ) \/
+      E. m e. NN E. f ( f : ( 1 ... m ) -1-1-onto-> A /\
+        x = ( seq 1 ( + ,
+         ( n e. NN |-> if ( n <_ m , [_ ( f ` n ) / k ]_ B , 0 ) ) ,
+         CC ) ` m )
+      ) ) ) $.
```
benjub, line 103498: Maybe fit the statements on fewer lines (while keeping breaking lines, as you do, at meaningful places) ? At least the last two lines can be merged.

iset.mm
```diff
@@ -103467,68 +103480,122 @@ Cauchy sequence (for example, fixing a rate of convergence as in
...
+  $}
+
+  ${
+    $d A x $.  $d M x $.  $d N x $.
+    sumdc.m $e |- ( ph -> M e. ZZ ) $.
+    sumdc.ss $e |- ( ph -> A C_ ( ZZ>= ` M ) ) $.
+    sumdc.dc $e |- ( ph -> A. x e. ( ZZ>= ` M ) DECID x e. A ) $.
+    sumdc.n $e |- ( ph -> N e. ZZ ) $.
+    $( Restating decidability of a summable set indexed by integers.
+       (Contributed by Jim Kingdon, 1-Jan-2022.) $)
```
benjub, line 103564: Even if the current applications of this result are when A is a set of indices, the result itself is about any upperset of integers, so maybe adapt the comment ?
ctually, it might be useful to add to iset.mm a more general result (and maybe also a definition, or else "inline" the definition in the proof): say "A is decidable in B" if ` A. x e. B DECID x e. A `; then, if A is a decidable subclass of B (meaning: it is a subclass of B and it is decidable in B), and B is decidable in C, then A is decidable in C.
The idea of a decidable class (in another) is reminiscent of my definition of bounded class http://us2.metamath.org/ileuni/df-bdc.html

iset.mm
```diff
@@ -103467,68 +103480,122 @@ Cauchy sequence (for example, fixing a rate of convergence as in
...
-     seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
-       ~~> x ) \/
-     E. m e. NN E. f ( f : ( 1 ... m ) -1-1-onto-> A /\
-       x = ( seq 1 ( + , ( n e. NN |-> [_ ( f ` n ) / k ]_ B ) , CC ) ` m )
-     ) ) ) $.
-
+       and summing in the induced order.  In both cases we have an ` if `
+       expression so that we only need ` B ` to be defined where ` k e. A ` .
+       These two methods of summation produce the same result on their common
+       region of definition (i.e. finite subsets of the upper integers).
```
benjub, line 103486: (i.e. finite sets of integers)

iset.mm
```diff
@@ -103467,68 +103480,122 @@ Cauchy sequence (for example, fixing a rate of convergence as in
...
+       Examples: ` sum_ k e. { 1 , 2 , 4 } ` ` k ` means ` 1 + 2 + 4 = 7 ` ,
+       and ` sum_ k e. NN ` ` ( 1 / ( 2 ^ k ) ) = 1 ` means 1/2 + 1/4 + 1/8
+       + ... = 1.  (Contributed by Jim Kingdon, 15-Feb-2022.) $)
+    df-isum $a |- sum_ k e. A B = ( iota x ( E. m e. ZZ ( A C_ ( ZZ>= ` m ) /\
+      A. j e. ( ZZ>= ` m ) DECID j e. A /\
```
benjub, line 103491: Maybe add something like (if I understood correctly): In the inifinite case, we also require that the indexing set be a decidable subset of an upperset of integers (that is, membership of integers in it is decidable).

iset.mm
jkingdon: Often a tradeoff between those two goals. But sure, merging the last two lines looks fine so I revised the pull request to do that.

iset.mm
jkingdon: I changed the comment to "Decidability of a subset of upper integers."

As for related theorems, I don't have a super clear vision of exactly what to prove in this area - I tend to prove them as I need them. Also in this general family are http://us.metamath.org/ileuni/infssuzcldc.html , http://us.metamath.org/ileuni/infssuzledc.html , http://us.metamath.org/ileuni/zsupcl.html , and perhaps others.

iset.mm
jkingdon: I've made this change for both set.mm and iset.mm.

iset.mm
jkingdon: I've added this wording.

benjub:
Thanks @jkingdon

PR #2502, opened 2022-02-15
Range cross theorems + minor changes

mazsa:
Mathbox only.
* range cross theorems
* modified Contents of my Mathbox based on the discussion of https://github.com/metamath/set.mm/pull/2485
* TEST ONLY in the comments section of ~ elv based on the https://groups.google.com/g/metamath/c/kdbrroGzJnc/m/7PCMZGgSAQAJ discussion: @digama how do you like my "Inference forms (with ` |- A e. _V ` hypotheses) (...)" modification? Should I revert them next time or can I keep them?
* renaming ~ bropabid to ~ brabidga (cf. ~ brabga in main set.mm)
* comments of ~ brabidga ~ motr ~ ineleq ~ inecmo ~ inecmo2 ~ inecmo3 ~ ineccnvmo ~ ineccnvmo2 ~ alrmomo ~ alrmomo2 ~ opabssi changed because of the comments of @avekens and @benjub in https://github.com/metamath/set.mm/pull/2479
* changing 1 ~~? to ~ in ~ df-xrn
* changing plural to singular in comments of ~ df-xrn ~ xrnss3v and ~ xrnrel .

PR #2496, opened 2022-02-13
Graph theory - minor improvements of Huneke's proof (8)

avekens:
General:
* Convention about "Organizing proofs" (conventions for lemmas) enhanced

Graph theory:

Neighbors/complete graphs:
* definitions of neighbors, universal vertices and complete graphs moved to the corresponding subsections
* theorems about neighbors revised: ~nbgrcl, ~nbgrnvtx0, ~nbgrel, ~nbgrisvtx, ~nbgrssvtx, ~nbgrnself2, ~nbgrssovtx, ~nbgrssvwo2, ~nbgrsym
* ~usgrnbnself2 marked as obsolete
* proofs of theorems about neighbors shortened: ~nbgr2vtx1edg, ~nbuhgr2vtx1edgblem, ~nbuhgr2vtx1edgb, ~nbgr0vtxlem, ~nbusgredgeu0, ~nbfusgrlevtxm1, ~nbfusgrlevtxm2, ~nb3grprlem2, nb3grprlem2, ~nbupgruvtxres
* proofs of theorems about complete graphs shortened: ~cplgr3v, ~cusgrexi

Closed walks as words:
* header comment of section "Closed walks as words" improved/enhanced
* theorems about closed walks as words revised: ~clwwlksn1
* new theorems about closed walks as words : ~loopclwwlksn1b, ~clwwlksn1loopb, ~0clwlkv

Friendship graphs:
* proofs of theorems about friendship graphs shortened: ~frgrnbnb, ~frgrncvvdeqlem1, ~frgrncvvdeqlem2, ~frgrncvvdeqlem3, ~frgrncvvdeqlem9, ~fusgreghash2wspv, ~numclwwlkovf2num, ~numclwlk1lem2foa, ~numclwlk1lem2fo
* theorems about friendship graphs revised: ~extwwlkfablem2, ~numclwwlkovf2exlem2, ~numclwwlkovf2
* comment for ~numclwwlkovf enhanced
* new theorems about friendship graphs: ~numclwwlkovf1, ~clwwlk1loop, ~clwwlk1nloop, ~clwwlk1, ~numclwwlkle1

PR #2505, opened 2022-02-17
Graph theory - some improvements (9)

avekens:
see discussion in PR #2502:
* ~opidg moved from AV's mathbox,
* proof of ~opid shortened
* ~opideq revised in PM's mathbox
* ~opidORIG deleted in PM's mathbox

misc.:
* ~fvmptrab added

graph theory:
* label fragment "uvtxa" changed to "uvtx"
* theorems for universal vertices revised: ~uvtxval, ~uvtxel, ~isuvtx, ~uvtxel1, ~uvtx01vtx
* proofs of theorems for universal vertices shortened: ~uvtxisvtx, ~vtxnbuvtx, ~uvtx0, ~uvtx2vtx1edg, ~uvtxnbgrb
* definitions ~df-cplgr and ~df-cusgr of complete graphs revised (as proposed by TA and BJ, see also discussion in PR [#2496](https://github.com/metamath/set.mm/issues/2496))
* theorems about complete graphs added/revised: ~prcliscplgr, ~cplgruvtxb, ~iscusgr
* proofs of theorems for complete graphs shortened: ~cplgr1v, ~cusgrexi, ~structtocusgr
* Header comment for section "Walks, paths and cycles" added, as promised in PR [#2496](https://github.com/metamath/set.mm/issues/2496)
* 3 comments revised as suggested by TA in PR #[#2496](https://github.com/metamath/set.mm/issues/2496)


set.mm
```diff
@@ -554884,19 +555102,20 @@ relation with an intersection with the same Cartesian product (cf.
...
+     ` <. B , B >. ` .  (Contributed by Peter Mazsa, 22-Jul-2019.)  (Revised by
+     AV, 16-Feb-2022.) $)
+  opideq $p |- ( A e. V -> ( <. A , A >. = <. B , B >. <-> A = B ) ) $=
+    ( wcel cop wceq wa wb opthg anidms anidm syl6bb ) ACDZAAEBBEFZABFZOGZOMNPHA
+    ABBCCIJOKL $.
+
+  $( Obsolete version of ~ opideq as of 16-Feb-2022.  (Contributed by Peter
+     Mazsa, 22-Jul-2019.)  (Proof shortened by AV, 16-Feb-2022.)
+     (New usage is discouraged.)  (Proof modification is discouraged.) $)
+  opideqOLD $p |- ( ( A e. V /\ B e. W ) ->
```
mazsa: I would prefer to delete this opideqOLD: it seems silly after opideq and nobody needs it (incl. me).

set.mm
```diff
@@ -51758,13 +51759,20 @@ we define it to be empty in this case (see ~ opprc1 , ~ opprc2 , and
     MABQUKVMVAVNVCABCGERABDGERUAUBUKVKABUOMZABUPMZJVGAUOUPBESUKVOVEVPVFABCEUCAC
     DBESUDUJABKEUEUFUGABVIURCDTUHUTVBTUI $.
 
+  $( The ordered pair ` <. A , A >. ` in Kuratowski's representation.  Closed
+     form of ~ opid .  (Contributed by Peter Mazsa, 22-Jul-2019.)  (Avoid
+     depending on this detail.) $)
+  opidg $p |- ( A e. V -> <. A , A >. = { { A } } ) $=
```
mazsa, line 51765: thank you.

set.mm
avekens: Usually, OLD theorem will be deleted after one year. However, I can delete it immediatly. But let's wait for other opinions.

set.mm
```diff
@@ -412515,6 +412599,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
+  * Paths with particular endpoints: ` ( F ( A ( PathsOn `` G ) B ) `
+
+  * Simple paths with particular endpoints: ` ( F ( A ( SPathsOn `` G ) B ) `
+
+  * Walks of a fixed length N as words with particular endpoints:
+    ` ( F ( A ( WWalksNOn `` G ) B ) `
+
+  * Simple paths of a fixed length N as words with particular endpoints:
+    ` ( F ( A ( WSPathsNOn `` G ) B ) `
+
```
tirix: Thank you very much!
Very clear.

tirix (approved):

set.mm
benjub: OK to delete now.

benjub:
Side remark, not on the content: there is often a large list of commits in your PRs, sometimes old (here, there are many which are two months old). Is it because you really work in parallel to the develop branch, or is it because you do not start with a fresh branch each time ? It looks like some of the older commits have already been merged. The question is not only to @avekens but also to some people who know git better than me and understand what's going on (maybe @tirix ?)

avekens:
> Side remark, not on the content: there is often a large list of commits in your PRs, sometimes old (here, there are many which are two months old). Is it because you really work in parallel to the develop branch, or is it because you do not start with a fresh branch each time ? It looks like some of the older commits have already been merged. The question is not only to @avekens but also to some people who know git better than me and understand what's going on (maybe @tirix ?)

I have to work on a fork, because I have not enough permissions to work in the set.mm repository directly. To create a new fork after each PR would cause a lot of (useless) work, therefore I keep the fork and update it with each PR. Doing this, I can handle merge conflicts easily.

Until December last year, this worked without problems. Since December, however, I also observed that the complete history is shown in the PRs. I think this is not harmful, at least only the changes I actually want to be merged are shown in the diff, and merging itself seems to be without problems' too.

Maybe it depends on the way how my PRs are merged (maybe Norm did it in a different way...).

avekens:
I resolved the merge conflicts now, which were caused by Peter's PR. Although it was a small conflict only, it took some time to handle it...

benjub:
> > Side remark, not on the content: there is often a large list of commits in your PRs, sometimes old (here, there are many which are two months old). Is it because you really work in parallel to the develop branch, or is it because you do not start with a fresh branch each time ? It looks like some of the older commits have already been merged. The question is not only to @avekens but also to some people who know git better than me and understand what's going on (maybe @tirix ?)
> 
> I have to work on a fork, because I have not enough permissions to work in the set.mm repository directly. To create a new fork after each PR would cause a lot of (useless) work, therefore I keep the fork and update it with each PR. Doing this, I can handle merge conflicts easily.
> 
> Until December last year, this worked without problems. Since December, however, I also observed that the complete history is shown in the PRs. I think this is not harmful, at least only the changes I actually want to be merged are shown in the diff, and merging itself seems to be without problems' too.
> 
> Maybe it depends on the way how my PRs are merged (maybe Norm did it in a different way...).

Yes, this is probably due on defaulting to "squash and merge".  I do not know what others think ?

When looking at the global diff of the PR, it is indeed the same, but sometimes it is easier to review commits separately.

My workflow is: after my PR has been merged, I almost always delete my corresponding local branch (git branch -D mybranch), and when I want to start a new work, first I synchonize (git checkout develop && git pull upstream develop), then I create a fresh branch (git checkout -b newbranch), then I do a few iterations of (modify set.mm ; git commit -am "this and that"), and finally I do a (git push origin newbranch).

That said, if you are on another system and changing your workflow would take too much time, then stay as you are.

mazsa:
@avekens @benjub @tirix Strangely enough, I'm that guy who knows the answer. I had the same problem (old commits in new PRs) I googled this out https://stackoverflow.com/questions/41283955/github-keeps-saying-this-branch-is-x-commits-ahead-y-commits-behind and that solved it. (This is useful as well https://gist.github.com/lttlrck/9628955 )

mazsa:
I'll try the workflow of @benjub (published in between).

benjub:
For the record: the page https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing mainly written by @david-a-wheeler does a good job at explaining how to set things up (before reading it, I had no experiene of git).
The thing to remember is that it is a "triangular" thing: you push to your github repository ("origin"), and you pull from the official set.mm github repository ("upstream"), which is why your git commands will look like (git pull upstream develop) and (git push origin mybranch).  The third side of the triangle is the actual PR, done on github. Of course, the first time, you have to indicate git what the "upstream" is and what the "origin" is. This is Step 4 in https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing

When looking at instructions on forums, you will generally want to replace "master" with "develop" since this is how we currently work.

set.mm
```diff
@@ -69732,6 +69740,24 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
...
+    $d M y $.  $d N x y $.  $d X x y $.  $d V x $.  $d ps x $.
+    fvmptrab.f $e |- F = ( x e. V |-> { y e. M | ph } ) $.
+    fvmptrab.r $e |- ( x = X -> ( ph <-> ps ) ) $.
+    fvmptrab.s $e |- ( x = X -> M = N ) $.
+    fvmptrab.v $e |- ( X e. V -> N e. _V ) $.
+    fvmptrab.n $e |- ( X e/ V -> N = (/) ) $.
+    $( Implications for the value of a function defined by the maps-to notation
+       with a class abstraction as a result having an element.  (Contributed by
+       Alexander van der Vekens, 15-Jul-2018.) $)
+    fvmptrab $p |- ( F ` X ) = { y e. N | ps } $=
```
benjub, line 69753: I'm not sure it is worth adding this to Main. It relies on the fact that out-of-domain arguments evaluate to the empty set, which relies on set.mm's particular encoding. It looks like it has a single usage, http://us2.metamath.org/mpeuni/uvtxaval.html, where it would shorten steps around http://us2.metamath.org/mpeuni/fvmptd.html ? To me, the current proof looks preferable.

set.mm
```diff
@@ -412491,6 +412575,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
 #*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
+
+  A "walk" in a graph is usually defined for simple graphs, multigraphs or even
+  pseudographs as "alternating sequence of vertices and edges x0 , e1 , x1 ,
+  e2 , ... , e(l) , x(l) where e(i) = x(i-1)x(i), 0<i<=l.", see definition of
+  [Bollobas] p. 4, or "A walk (of length k) in a graph is a non-empty
+  alternating sequence v_0 e_0 v_1 e_1 ... e_k-1 v_k of vertices and edges in G
+  such that e_i = { v_i , v_i+1 } for all i < k.", see definition of [Diestel]
+  p. 10.
+
```
benjub, line 412586: So why not define a walk in G as a function
f : [[0, 2n]] --> ( Edge(G) u. Vertex(G) ) such that for all 0<=k<=n, f(2k) e. Vertex(G) and for all 0<=k<=n-1, f(2k+1) e. Edge(G, f(2k), f(2k+2))
(where the last set is the set of edges in G between two given vertices; I do not know how it is named in set.mm) ?

set.mm
avekens: I think this pattern is used very often in set.mm, especially to avoid the distiction between sets and proper classes. In the case of uvtxval, the left hand side as well as the right hand side relies on this specialty ( { v e. V | .. } = { v e. ( Vtx `G ) | .. } = { v e. (/) | .. } = (/) if -. G e. _V). Would it be OK if I revise fvmptrab so that N is replaced by a function value ( G ` X ) ?

set.mm
avekens: There is no definition for the set of edges between (two) vertices in set.mm, as far as I know. Therefore, f(2k+1) e. Edge(G, f(2k), f(2k+2)) would have to be replaced by ( f(2k+1) e. Edge(G) /\ { f(2k), f(2k+2) } C_ f(2k+1) ). But is it really useful to mix both sequences into one function? I could imagine that working with such a definition, although it might be more elegant, would be very tedious.

Since a lot of theorems (especially the Koenigsbrg bridge theorem and the friendship theorem) are based on the current definition, which was already used/introduced by @digama0, it would be a huge undertaking to change it. Maybe a study of this approach could be started in a mathbox.

set.mm
benjub: Do you mean something like
```
$e ...
$e |- ( X e. dom F <-> Y e. dom G )
$p |- ( F ` X ) = { y e. ( G ` Y ) | ps }
```
?

set.mm
benjub: Ok. Maybe add to the head comment something like:
Therefore, to stay close to the textbook definition, one could define a walk as a function f : [0...2n] --> ( Edge(G) u. Vertex(G) ) such that for all 0<=k<=n, f(2k) e. Vertex(G) and for all 0<=k<=n-1, ( f(2k+1) e. Edge(G) /\ { f(2k), f(2k+2) } C_ f(2k+1) ) (in "pseudocode").  However, this definition would be uneasy to use, therefore, we decide to define walks as pairs of functions f from { 1 , ... , n } and p from { 0 , ... , n }, where f enumerates...

It might indeed be interesting to state the alternate definition in a mathbox and prove equivalence:
```
function f into (Vertex u. Edges)      <->  interlaced functions v and e
f                                      |->  (f o. (x |-> 2x) , f o. (x |-> 2x+1)) 
(x |-> if(x even, v(x/2), e((x-1)/2))) <-|  (v, e)
```

set.mm
```diff
@@ -412491,6 +412575,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
+  p(0) e(f(1)) p(1) e(f(2)) ... p(n-1) e(f(n)) p(n).
+
+  Based on this definition of ` Walks `, the class of all walks, more
+  restrictive constructs are defined:
+
+  * ` Trails `: A "walk is called a _trail_ if all its edges are distinct.",
+    see Definition of [Bollobas] p. 5, i.e., f(i) =/= f(j) if i =/= j.
+
+  * ` Paths `: A _path_ is a walk whose vertices except the first and the last
+    vertex are distinct, i.e., p(i) =/= p(j) if i =/= j, except i = 0 and
```
benjub, line 412603: if i < j, except possibly when (i , j) = (0, n)
OR
if i =/= j, except possibly when (i , j) = (0, n) or (n, 0)
OR
(you may write couples as <. i , j >. instead of (i, j), I don't know what is preferred in comments)

set.mm
avekens: see my latest commit.

set.mm
```diff
@@ -412491,6 +412575,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
+  * ` SPaths ` (simple paths): A _simple path_ "is a walk with distinct
+    vertices.", see Notation of [Bollobas] p. 5, i.e., p(i) =/= p(j)
+    if i =/= j.
+
+  * ` ClWalks ` (closed walks): A walk whose endvertices coincide is called a
+    _closed walk_, i.e.,  p(0) = p(n).
+
+  * ` Circuits `: "A trail whose endvertices coincide (a closed trail) is
+    called a _circuit_." (see Definition of [Bollobas] p. 5), i.e.,
+    f(i) =/= f(j) if i =/= j and p(0) = p(n).
```
benjub, line 412615: Maybe add: equivalently, a circuit is a closed walk with distinct edges.

set.mm
```diff
@@ -412491,6 +412575,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
+
+  * ` Cycles `: A path whose endvertices coincide (a closed path) is called a
+    _cycle-, i.e., i.e. p(i) =/= p(j) if i =/= j, except i = 0 and j = n, and
+    p(0) = p(n).
```
benjub, line 412619: Maybe add: equivalently, a cycle is a closed walk with distinct vertices.

set.mm
```diff
@@ -412491,6 +412575,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
+
+  * ` EulerPaths ` (Eulerian paths): An _Eulerian path_ is "a trail containing
+    all edges" (see definition in [Bollobas] p. 16), i.e.,  f(i) =/= f(j) if
```
benjub, line 412622: double quotes unnecessary

benjub: maybe: all edges of the graph

benjub: so an Eulerian path is a trail, not a path ? Why not "Eulerian trail", then ?

set.mm
```diff
@@ -412491,6 +412575,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
+
+  * Cycles are circuits ( ~ cycliscrct ):
+    ` ( Cycles `` G ) C_ ( Circuits `` G ) `
+
+  * (Non trivial) cycles are not simple paths ( ~ cyclnspth ):
+    ` ( F =/= (/) -> ( F ( Cycles `` G ) P -> -. F ( SPaths `` G ) P ) ) `
+
+  * Eulerian paths are trails ( ~ eupthistrl ):
+    ` ( EulerPaths `` G ) C_ ( Trails `` G ) `
+
```
benjub, line 412661: Is an Eulerian path a path ?

set.mm
```diff
@@ -412491,6 +412575,135 @@ about equation (2) in section I.1 in [Bollobas] p. 4.  (Contributed by
...
+  * Paths with particular endpoints: ` ( F ( A ( PathsOn `` G ) B ) `
+
+  * Simple paths with particular endpoints: ` ( F ( A ( SPathsOn `` G ) B ) `
+
+  * Walks of a fixed length N as words with particular endpoints:
+    ` ( F ( A ( WWalksNOn `` G ) B ) `
+
+  * Simple paths of a fixed length N as words with particular endpoints:
+    ` ( F ( A ( WSPathsNOn `` G ) B ) `
+
```
benjub, line 412706: Thanks @avekens, that's a very useful head comment !

set.mm
avekens: Hm, in textbooks, the vertices and edges in a walk are usually indexed separately, like x0 , e1 , x1 , e2 , ... , e(l) , x(l) where e(i) = x(i-1)x(i), 0<i<=l, see definition of [Bollobas] p. 4, or v0 e0 v1 e1 ... e(k-1) v(k) such that e(i) = { v(i) , v(i+1) } for all i < k, see definition of [Diestel]  p. 10. With your definition, it would be x0 , e1 , x2 , e3 , ... , e(2l-1) , x(2l) where e(2i+1) = x(2i)x(2i+2), 0<i<l.

set.mm
benjub: Your new statement does not depend on the encoding and looks more natural, so I prefer it and I think it's ok to have it in Main (although I do not know if it is widely applicable).  The comment has problems ("result having" and "an function"). Maybe:
> Value of a function defined in maps-to notation with value a particular kind of restricted abstraction (the function ` F ` is the subfunction of ` G ` cut out by ` ph ` ).

Does it hold when _V is replaced by V and you add $p |- X e. V -> ( ... ? (or use deduction form)

benjub: And do you think that the extra generality of my above proposal with `( X e. dom F <-> Y e. dom G )` is useless ? (it may very well be)

set.mm
benjub: Ok. So either add "One could define a walk as a function f : [0...2n] --> ( Edge(G) u. Vertex(G) ) such that...", or leave the comment as it is, as you prefer.

avekens: I will add a hint to this alternate definition in my next PR.

set.mm
avekens: * double quotes indicates that it is a quotation of Bollobas.
* although not mentioned by Bollobas, I will add "[of the graph]" as proposed.
* Although Eulerian paths need not be paths according to our definition, they are often called "Eulerian paths" in literature. @digama0 also used this term for his proof of the Koenigsberg Bridge theorem. A more detailed discussio  is provided in the header comment of section "Eulerian paths".

set.mm
avekens: not necessarily, see above.

set.mm
avekens: Your proposal is more general, but I see no application at the moment. Furthermore, I think it is a little bit different regarding DV conditions, and therefore it is not (directly) applicable for the prove of ~uvtxval.
I'll put it in my mathbox (as theorem ~fvmptrabdm ).

set.mm
benjub: Since this is a red herring, it may be worth repeating here: "Note, however, that an Eulerian path need not be a path."

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: done (based on the first suggestion)

benjub (approved):

set.mm
digama0: > Ok. Maybe add to the head comment something like: Therefore, to stay close to the textbook definition, one could define a walk as a function f : [0...2n] --> ( Edge(G) u. Vertex(G) ) such that for all 0<=k<=n, f(2k) e. Vertex(G) and for all 0<=k<=n-1, ( f(2k+1) e. Edge(G) /\ { f(2k), f(2k+2) } C_ f(2k+1) ) (in "pseudocode").

I would challenge the claim that this is in fact closer to the textbook definition if such a path is notated as `v_0 e_0 v_1 e_1 ...`. In a textbook presentation I can't imagine a single sequence where the type of the object changes on every alternate index being less confusing in any way.

benjub: @digama0 : you're probably right, but... pistol or sword ? Where shall we meet at dawn ? I choose @avekens as my second.

digama0: If we're dueling, then as my evidence I would point to [wikipedia](https://en.wikipedia.org/wiki/Path_(graph_theory)) and [MathWorld](https://mathworld.wolfram.com/Walk.html), which both use two sequences `v_i` and `e_i` (although the wikipedia version differs slightly in taking the v_i sequence as existential). Neither one even mentions the idea of an even/odd sequence such as you have proposed. Do you have a source?

benjub: As I wrote above, you're probably right, but dueling is no fun without a dose of bad faith ! (and irony is harder to convey in words)
(edit: so to be blunt: yes, you're right, and I recognized it in my previous comment; I was just trying to be humorous!)

PR #2506, opened 2022-02-18
Proof for ~opabssi, catching up STS formatting

tirix:
- Proof for Peter's `~opabssi` theorem discussed in #2502 without the `Rel A` hypothesis
- Updates to the MathML/STS formatting.

### DIFF
--- a/set-mathml.mmts
+++ b/set-mathml.mmts
@@ -901,7 +901,7 @@ $s class SubGraph $: <mi href="df-subgr.html">SubGraph</mi> $.
 $s class NeighbVtx $: <mi href="df-nbgr.html">NeighbVtx</mi> $.
 $s class ComplGraph $: <mi href="df-cplgr.html">ComplGraph</mi> $.
 $s class ComplUSGraph $: <mi href="df-cusgr.html">ComplUSGraph</mi> $.
-$s class UnivVtx $: <mi href="df-uvtxa.html">UnivVtx</mi> $.
+$s class UnivVtx $: <mi href="df-uvtx.html">UnivVtx</mi> $.
 $s class .ef $: <mi href="df-edgf.html"> ef </mi> $.
 $s class Vtx $: <mi href="df-vtx.html"> Vtx </mi> $.
 $s class iEdg $: <mi href="df-iedg.html"> iEdg </mi> $.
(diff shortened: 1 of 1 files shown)


benjub:
Can you see if you can also prove the version of opabid2 I suggested at https://github.com/metamath/set.mm/pull/2502#discussion_r809969982 ?

set.mm
```diff
@@ -464135,6 +464135,17 @@ Class abstractions (a.k.a. class builders)
       FUMQABCTUAUBUCUDUE $.
   $}
 
+  ${
+    $d A x y z $.  $d ph z $.
+    opabssi.1 $e |- ( ph -> <. x , y >. e. A ) $.
+    $( Sufficient hypothesis for a collection of ordered pairs to be a subclass
+       of a relation.  (Contributed by Peter Mazsa, 21-Oct-2019.) $)
+    opabssi $p |- { <. x , y >. | ph } C_ A $=
```
mazsa, line 464143: 1. I see, great:) 2. No way. I insist on this addition at least : (Revised by Thierry Arnoux, 18-Feb-2022.)

mazsa: @benjub do you think this "Revised by" is OK? I think @tirix thinks that you was the first who had the idea. But Thierry implemented it.

benjub: I think it's ok as is, but the credit could as well go to @tirix, that's not very important in this case.  The general view is that "ideas are cheap, implementation is hard", so credits should generally go to the person who does the formalization over the one who has the idea of doing it.

mazsa: @benjub I completely agree. @tirix please correct this.

mazsa: @tirix I mean please correct this in a way that you Thierry is the author.

tirix:
> Can you see if you can also prove the version of `~ opabid2` I suggested at [#2502 (comment)](https://github.com/metamath/set.mm/pull/2502#discussion_r809969982) ?

See last commit: is that what you meant?

set.mm
tirix: Ok, since it's my choice I kept the version with contributed/revised. 😁

benjub:
> > Can you see if you can also prove the version of `~ opabid2` I suggested at [#2502 (comment)](https://github.com/metamath/set.mm/pull/2502#discussion_r809969982) ?
> 
> See last commit: is that what you meant?

Yes, that's the statement ! (although I was thinking of proving opabssi from opabid2ss instead of the other way around, but yours probably makes for a shorter total)

benjub (approved):

set.mm
```diff
@@ -464135,6 +464135,26 @@ Class abstractions (a.k.a. class builders)
       FUMQABCTUAUBUCUDUE $.
   $}
 
+  ${
+    $d A x y z $.  $d ph z $.
+    opabssi.1 $e |- ( ph -> <. x , y >. e. A ) $.
+    $( Sufficient hypothesis for a collection of ordered pairs to be a subclass
```
benjub, line 464141: "Sufficient condition", rather ?

avekens (approved):

set.mm
tirix: Indeed! Fixed.

PR #2507, opened 2022-02-18
prove opelres/brres in terms of opelresg/brresg

benjub:
This shortens the total length of the combined proofs and avoids dummy variables. It is in line with a recent PR that favors proving the "g-version" `$p |- ( A e. V ->` first and then the version `$e |- A e. _V $.` from it.

The deprecation of resid carries out one of Norm's TODOs.

I propose to move @mazsa's ssbr to Main, next to the existing ssbri and ssbrd (there are many results in the three forms closed/inference/deduction in the neighborhood of theses results -- for later theorems, and especially when there are several hypotheses/antecedents, the deduction form is preferred). I'll uniformize the comments and see which order makes proofs shorter. Any objections ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14379,6 +14379,7 @@ New usage of "brfi1indALT" is discouraged (0 uses).
 New usage of "brfi1indALTOLD" is discouraged (0 uses).
 New usage of "brfi1indOLD" is discouraged (0 uses).
 New usage of "brfi1uzindOLD" is discouraged (1 uses).
+New usage of "brresOLD" is discouraged (0 uses).
 New usage of "c-bnj14" is discouraged (131 uses).
 New usage of "c-bnj18" is discouraged (58 uses).
 New usage of "cayleyhamiltonALT" is discouraged (0 uses).
@@ -17118,6 +17119,8 @@ New usage of "opelcn" is discouraged (1 uses).
 New usage of "opelopab4" is discouraged (0 uses).
 New usage of "opelopabsbALT" is discouraged (3 uses).
 New usage of "opelreal" is discouraged (8 uses).
+New usage of "opelresOLD" is discouraged (0 uses).
+New usage of "opelresgOLD" is discouraged (0 uses).
 New usage of "opfi1indOLD" is discouraged (0 uses).
 New usage of "opfi1uzindOLD" is discouraged (1 uses).
 New usage of "opidon2OLD" is discouraged (1 uses).
@@ -17529,6 +17532,7 @@ New usage of "relsnopOLD" is discouraged (0 uses).
 New usage of "renegclALT" is discouraged (0 uses).
 New usage of "renicax" is discouraged (0 uses).
 New usage of "resfunexgALT" is discouraged (0 uses).
+New usage of "residOLD" is discouraged (0 uses).
 New usage of "resima2OLD" is discouraged (0 uses).
 New usage of "resiun1OLD" is discouraged (0 uses).
 New usage of "restidsingOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


mazsa:
@benjub please consider using the ALTV versions with the same authors in my Mathbox (opelresALTV, brresALTV, etc., I don't use the -g suffix), I found them better when I used them with constrained quantifications. If you don't agree, or the timing is not perfect here, no problem:  I keep them in my Mathbox and I'll open an issue later.

set.mm
```diff
@@ -61668,50 +61668,77 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
-       17-Apr-2014.)  (Revised by Mario Carneiro, 19-Apr-2014.) $)
-    dfres3 $p |- ( A |` B ) = ( A i^i ( B X. ran A ) ) $=
-      ( vx vy vz cres cvv cxp cin crn df-res cv wcel wa cop wceq wex vex elxp
-      wb biantru opelrn biantrud syl5bbr syl6bi pm5.32d 2exbidv 3bitr4g pm5.32i
-      eleq1 com12 elin bitr4i ineqri eqtri ) ABFABGHZIABAJZHZIZABKCAUPUSCLZAMZU
-      TUPMZNVAUTURMZNUTUSMVAVBVCVAUTDLZELZOZPZVDBMZVEGMZNZNZEQDQVGVHVEUQMZNZNZE
-      QDQVBVCVAVKVNDEVAVGVJVMVGVAVJVMTZVGVAVFAMZVOUTVFAUJVJVHVPVMVIVHERZUAVPVLV
-      HVDVEADRVQUBUCUDUEUKUFUGDEUTBGSDEUTBUQSUHUIUTAURULUMUNUO $.
+    opelresOLD.1 $e |- B e. _V $.
+    $( Old proof of ~ opelresgOLD .  Obsolete as of 18-Feb-2022.  (Contributed
```
avekens, line 61715: Old proof of ~ opelres?

set.mm
```diff
@@ -61668,50 +61668,77 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
+       by NM, 13-Nov-1995.)  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    opelresOLD $p |- ( <. A , B >. e. ( C |` D ) <->
+                    ( <. A , B >. e. C /\ A e. D ) ) $=
+      ( cop cres wcel cvv cxp cin wa df-res eleq2i elin opelxp mpbiran2 anbi2i
+      3bitri ) ABFZCDGZHTCDIJZKZHTCHZTUBHZLUDADHZLUAUCTCDMNTCUBOUEUFUDUEUFBIHEA
+      BDIPQRS $.
+
+    $( Old proof of ~ opelresgOLD .  Obsolete as of 18-Feb-2022.  (Contributed
```
avekens, line 61724: Old proof of ~ brres?

set.mm
```diff
@@ -61668,50 +61668,77 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
+      ( cop cres wcel wa wbr opelres df-br anbi1i 3bitr4i ) ABFZCDGZHOCHZADHZIA
+      BPJABCJZRIABCDEKABPLSQRABCLMN $.
   $}
 
   ${
     $d y A $.  $d y B $.  $d y C $.  $d y D $.
-    $( Ordered pair membership in a restriction.  Exercise 13 of
-       [TakeutiZaring] p. 25.  (Contributed by NM, 14-Oct-2005.) $)
-    opelresg $p |- ( B e. V -> ( <. A , B >. e. ( C |` D ) <->
+    $( Old proof of ~ opelresgOLD .  Obsolete as of 18-Feb-2022.  (Contributed
```
avekens, line 61734: Old proof of ~ opelresg?

avekens (changes requested):

set.mm
benjub: Good catches, thanks.

benjub:
I'll move ssbr in a future PR.
@mazsa : I understand that your ALTV versions are slight modifications (the conjunction in the consequent is commuted) ?
Apart from exceptional cases (earliest theorems, abid1/abid2 --- read comment of abid1), we have only one variant in set.mm. If your variant is a net gain, in that its proof is shorter and it is easier to use later (in this case, this could be because all of its applications need to commute back the conjunction), then it could replace the existing one. Else, it would be better that you use the existing one, so as not to have too many variants when not necessary. But nothing is urgent.

(Note that we try to have consistent conventions for these "variants up to commutation/symmetry", like A=... versus ...=A, etc., that try to reduce the number of "gluing lemmas" (symmetry, commutation,...) needed, but we of course may be missing some.

mazsa:
@benjub yes, after everything in my Mathbox, I'll gather my arguments, open an issue, we decide whether this or that, and I'll implement changes accordingly. Thank you.

PR #2508, opened 2022-02-18
revise eumo

benjub:
Following private communication from Gérard Lang.


jkingdon (approved):

avekens (approved):

benjub:
@jkingdon : I now see that you had already written that same proof in iset.  It was rediscovered independently by Gérard, so I'll leave credits like that.

PR #2509, opened 2022-02-19
move ssbr to Main

benjub:
As agreed in #2507.
I compared the total lengths of the combined proofs when proving ssbr from ssbrd and the other way around, and chose the shortest.


tirix (approved):

digama0:
@benjub I just got 178 emails from github in the past 5 minutes about you pushing to every branch that exists in the repo; did you do something unusual?

benjub:
Oh... did I do something wrong ? I deleted the stale, merged branches which are zero commits ahead, as announced in https://groups.google.com/g/metamath/c/nplf76VwJ_o/
Is it ok ?

digama0:
Ah, okay that makes sense. I wasn't able to find any of the branches mentioned in the emails, of course, so I wasn't sure if that was what was happening. (I have github set up to email me on any change to the repo, so I doubt anyone else got the same deluge.)

PR #2516, opened 2022-02-20
reduce usage of some OLD theorems

benjub:
Some *OLD obsolete theorems are still not removable because they are used by some non-OLD theorems.  We'll have to dig into that, but this PR cleans things a bit.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4293,25 +4293,20 @@
 "crngcALTV" is used by "rngcrescrhmALTV".
 "crngcALTV" is used by "rngcvalALTV".
 "csbabgOLD" is used by "csbingVD".
-"csbabgOLD" is used by "csbrngOLD".
 "csbabgOLD" is used by "csbrngVD".
 "csbabgOLD" is used by "csbsngVD".
 "csbabgOLD" is used by "csbunigVD".
-"csbabgOLD" is used by "csbxpgOLD".
 "csbabgOLD" is used by "csbxpgVD".
 "csbeq2gOLD" is used by "csbima12gALTVD".
 "csbeq2gOLD" is used by "csbresgVD".
 "csbeq2gOLD" is used by "csbrngVD".
 "csbeq2gOLD" is used by "csbsngVD".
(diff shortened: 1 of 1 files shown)


tirix (approved):

tirix:
But that seems to be only old theorems using old theorems?
*(Edit)* Oh, you're right, there are around 50 such dependencies:
```
> grep -E 'OLD\" is used by \"[a-z0-9\._-]+\"' discouraged  | head
"cghomOLD" is used by "ghomco".
"cghomOLD" is used by "ghomdiv".
"cghomOLD" is used by "ghomf".
"cghomOLD" is used by "grpokerinj".
"cghomOLD" is used by "rngogrphom".
"cghomOLD" is used by "rngohom0".
"cghomOLD" is used by "rngohomsub".
"cghomOLD" is used by "rngokerinj".
"clmgmOLD" is used by "exidcl".
"cnaddabloOLD" is used by "cncph".
```

avekens (approved):

avekens:
I think a lot (or most, or maybe all) of these cases come from the deprecated PART 18  COMPLEX TOPOLOGICAL VECTOR SPACES: See issue #2201, which is far from being finished (I started to work on it, but there is still a lot do, and I hoped that somebody else would do it).  I moved already a lot of OLD theorems into mathboxes, in which they still are used (mainly in mathbox of JM, see, for example, ~exidcl using ~clmgmOLD). Maybe JM could clean up his mathbox accordingly (who can contact him?).

benjub:
@tirix : indeed, there are some complications.  There are also many "virtual deduction" theorems (labeled *VD) which still use *OLD theorems but whose proofs were automatically generated, so there are more questions to settle.

The point of this short PR was to quickly clean what I could, in the hope that reducing the numbers of usages/dependencies would clear things up for when the real work is done.

@avekens : I do not know how to contact Jeff Madsen, but it seems that he is not an active contributor anymore ?  Anyway, there is no urgency. If you can make some progress on #2201 at your own pace, that would be nice, since you know better what to do, but again, no urgency.

PR #2511, opened 2022-02-19
Graph theory: alternate definition of a walk

avekens:
Alternate definition of a walk as proposed by BJ (see #2505) described in the header comment of section "Walks, paths and cycles".

This is an PR with a small change only, because I tried to cleanup my fork before. Unfortunately, it seems that it didn't help to reduce the number of indicated commits. I have to investigate/test further...

PR #2418, opened 2022-01-03
Ternary Goldbach Conjecture

tirix:
Alexander called for a proof of the ternary Goldbach conjecture in the header comment of Goldbach conjectures:
> _It would be great if this proof can be formalized with Metamath_

This commit brings in the final step of the proof of the ternary Goldbach conjecture. The main statement is `~ tgoldbachgt` , it is stated using the same language as Alexander's `~ ax-tgoldbachgt`. This proof follows chapter 7 of [Helfgott].

The only ;-) part left to prove are:
- the statement 7.49 at the top of p. 70 of the paper, provided temporarily as an axiom `~ ax-hgt749`
- the theorem 12 of [RosserSchoenfeld], provided temporarily as axiom `~ ax-ros335`
- the theorem 13 of [RosserSchoenfeld], provided temporarily as axiom `~ ax-ros336`

This PR also brings in supporting theorems and lemmas.
All changes are within my mathbox.

PR #2517, opened 2022-02-20
Minor corrections of comments

avekens:
Small PR to clean up my Git repository (a wrong comment was corrected).

@mazsa thank you very much for your hints. I hope that I will provide clean PRs in the following. I was not able to call
git push --force-with-lease origin ... on the command line until now, because I had no access token (used my plain PWD) which were intoduced in 2021. Usually, I am pushing changes via Github Desktop, and for this no access token seems to be required...


benjub:
@tirix : do you know what's going on, relative to @avekens's comment above ? Do you have some experience with GitHub Desktop ? (I do not, I only use git through command line and GitHub through my web browser)

benjub (approved):

tirix:
I'm mostly using GitHub Desktop too, and sometimes using the command line where GitHub Desktop fails me.
Actually I face exactly the same problem. Longer metamath development might take months, and I usually stay on the same branch for that (I usually already have started with the next changes on my branch before the current PR is merged). See for example #2418, which suffers the same disease as Alexander's recent PR's: the commit history of one branch is not reset after it's merged.

Here is what I understand of the problem ([this SO question](https://stackoverflow.com/questions/49200667/github-squash-and-merge-subsequent-pull-request-showing-all-previous-changes?noredirect=1&lq=1) seems a good explanation):
In order to find out which commits are in a PR, GitHub compares the (SHA of the) commits in your branch and in the origin (develop) branch. Originally, with "merge commit", commits were added one by one, so your previous commits on the branch would be the same on the origin branch, GitHub would know they are the same and would not list them.

However now, our merge policy is "squash and merge", so that all commits are squashed into a single one on develop. This means GitHub cannot identify the commits on the branch with the commits on develop, and mistakenly assumes they are all new.

The solution seems to be to _rebase_ your branch after it has been merged, and then force push. I think this is similar to what Peter mentioned in #2511.

avekens:
I will try a rebase after this PR is merged (to start with a "clean" branch with no changes  from me).

avekens:
So why do we have the "squash and merge" policy now, instead of an ordinary "merge"? It seems that it causes overhead for serveral contributors...

benjub:
@tirix: thanks.  That's what I thought: with "squash and merge", there is no way for git to compare the commits on both branches.  So the "squash and merge", which definitely reduces the number of commits on `develop`, has the side effect of having them "pile up" on local branches... I do not know if we should change the merging procedure here... @digama0 ?

tirix:
Alternatively we could go with the flow and use short-lived branches.
This means moving our ongoing work to a new branch after every PR.
We'd probably name each short-lived branch with the name of the most important theorem added or changed...

avekens:
> Alternatively we could go with the flow and use short-lived branches. This means moving our ongoing work to a new branch after every PR. We'd probably name each short-lived branch with the name of the most important theorem added or changed...

I created a new branch, but had the same problem as before (maybe my fork is corruped somehow...). Since working with Github Desktop is not as flexible as using a command line, but much more convenient, for creating a new branch for each (planned) PR we have to switch to the command line (as git-bash), and also pushing may not be possible with Github Desktop anymore. Finally, I almost always do not know which will be the main goal of my next PR, so providing an appropriate name for a branch would not be possible. Therefore, I would prefer not to "squash" in the future (as was done until December 2021).

avekens:
Indepenently from the question how to merge (in the future), could somebody merge this PR as soon as possible, please? Since I do not want to make further changes until this change is merged, I am blocked somehow...

tirix (approved):

digama0:
My suggestion for the "squash and merge" workflow is: when your PR branch is merged, rebase all your WIP material which depends on the PR branch onto `develop`. This rebase should not involve any complicated merging since the new state of master is identical to the state of the PR branch before merging. You should abandon the original PR branch once it is merged; either delete it, or if you would like to keep the name for another PR then just `git reset` it to point to master and forget about the original PR commits (github will hold on to them but you don't need them in your local copy anymore).

benjub:
> My suggestion for the "squash and merge" workflow is: when your PR branch is merged, rebase all your WIP material which depends on the PR branch onto `develop`. This rebase should not involve any complicated merging since the new state of master is identical to the state of the PR branch before merging. You should abandon the original PR branch once it is merged; either delete it, or if you would like to keep the name for another PR then just `git reset` it to point to master and forget about the original PR commits (github will hold on to them but you don't need them in your local copy anymore).

When writing "master", I think you mean `metamath/set.mm:develop` (which can be called as `upstream/develop` once you have set git on your computer with `git remote add upstream https://github.com/metamath/set.mm.git`).

digama0:
yes, I have to find/replace "master" with "develop" in my speech when talking about set.mm. The split never really worked out as intended, I think we should just rename "develop" and remove the long abandoned "master" branch.

benjub:
> yes, I have to find/replace "master" with "develop" in my speech when talking about set.mm. The split never really worked out as intended, I think we should just rename "develop" and remove the long abandoned "master" branch.

This might be a good idea.

david-a-wheeler:
This is an area where people disagree. Mario prefers squash and merge, because then the histories are simpler. I avoid squash and merge, because it removes information that is sometimes helpful or important.

This may warrant a larger email list discussion, I can't do that right now.

david-a-wheeler:
If we rename the "develop" branch, I suggest renaming it to "main". Many are switching to "main" as the main branch. We can then rename the current "master" branch to something like "historical" so people will be unlikely to use it by accident.

PR #2518, opened 2022-02-20
Cosets + minor changes

mazsa:
1. htmldefs, etc.
2. new symbols
3. the final version of the comments for ` ~ el*v* `
4. dfxrn2, but commented out, because one necessary theorem is in the Mathbox of @tirix
5. new section for the cosets-related theorems

PR #2519, opened 2022-02-21
Shorten and move `~cnvoprab` to main

tirix:
This is a proposal to move `~cnvoprab` to main, since it is required by @mazsa's `~dfxrn2` in #2518 .

In the same move, I'm shortening `~cnvoprab` from around 11 lines down to just a bit more than 3 lines (while also replacing the non-free hypothesis with distinct variable requirements).
The OLD version is kept in my mathbox.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14737,6 +14737,7 @@ New usage of "cnvbracl" is discouraged (2 uses).
 New usage of "cnvbramul" is discouraged (1 uses).
 New usage of "cnvbraval" is discouraged (1 uses).
 New usage of "cnvcnvOLD" is discouraged (0 uses).
+New usage of "cnvoprabOLD" is discouraged (0 uses).
 New usage of "cnvsnOLD" is discouraged (0 uses).
 New usage of "cnvsngOLD" is discouraged (0 uses).
 New usage of "cnvssOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


avekens (approved):

benjub (approved):

set.mm
```diff
@@ -465888,37 +465904,45 @@ its graph has a given second element (that is, function value).
     $( Describe an implicit one-to-one onto function of two variables.
        (Contributed by Thierry Arnoux, 17-Aug-2017.) $)
     f1od2 $p |- ( ph -> F : ( A X. B ) -1-1-onto-> D ) $=
-      ( wa wsbc va cxp wfn ccnv wf1o wcel wral ralrimivva fnmpt2 syl cv opelxpi
```
benjub, line 465891: I guess the proof is now much longer because the DV conditions in cnvoprab force you to use dummy variables and "rebuild" formulas like x = A -> ( phi[x] <-> phi[A]) (or x = A -> ( B[x] = B[A]) ) ? Did you `MM-PA>minimize_with *` ?

benjub: Comment suggestion: Sufficient condition for a binary function expressed in maps-to notation to be bijective.

tirix: > I guess the proof is now much longer because the DV conditions in cnvoprab force you to use dummy variables and "rebuild" formulas like x = A -> ( phi[x] <-> phi[A]) (or x = A -> ( B[x] = B[A]) ) ?

Yes, that's correct!

> Did you `MM-PA>minimize_with *` ?

No, I'll try that.

tirix: Minimizing did save a few bytes: _Proof of "f1od2" decreased from 2830 to 2819 bytes using "nfeq2"._

PR #2514, opened 2022-02-19
Indent section titles

benjub:
A few comment edits and rewraps.

Strangely, the script indenting section titles did not catch them all...

PR #2521, opened 2022-02-21
comment edits of seqid3, seqid, seqid2, df-gsum

benjub:
* comment edits of seqid3, seqid, seqid2, df-gsum
* rewraps of some nearby statements (when the math-expression after the typecode '|-' fits on a single line, then write it on a single line)

(this is the first half of the recently closed #2514)

(edit: should be merged before I push the second half, since I changed the capitalization of one section title and the second half will modify indentation of section titles)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15435,7 +15435,6 @@ New usage of "elnlfn" is discouraged (4 uses).
 New usage of "elnlfn2" is discouraged (2 uses).
 New usage of "elnp" is discouraged (5 uses).
 New usage of "elnpi" is discouraged (4 uses).
-New usage of "elopOLD" is discouraged (0 uses).
 New usage of "elpaddatiN" is discouraged (2 uses).
 New usage of "elpaddatriN" is discouraged (0 uses).
 New usage of "elpclN" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -149685,11 +149685,11 @@ seq M ( .+ , F ) ) $=
     seradd.2 $e |- ( ( ph /\ k e. ( M ... N ) ) -> ( F ` k ) e. CC ) $.
     seradd.3 $e |- ( ( ph /\ k e. ( M ... N ) ) -> ( G ` k ) e. CC ) $.
     seradd.4 $e |- ( ( ph /\ k e. ( M ... N ) ) ->
-                     ( H ` k ) = ( ( F ` k ) + ( G ` k ) ) ) $.
+                                     ( H ` k ) = ( ( F ` k ) + ( G ` k ) ) ) $.
     $( The sum of two infinite series.  (Contributed by NM, 17-Mar-2005.)
```
avekens: I think there is no convention/agreement that a (follow up) line should always be right aligned. I would prefer the following line break/indentation (the dots in the second lien are only for displaying purposes in this comment):
```metamath
seradd.4 $e |- ( ( ph /\ k e. ( M ... N ) )
................................-> ( H ` k ) = ( ( F ` k ) + ( G ` k ) ) ) $.
```

avekens (approved):

set.mm
benjub: You're right that there is no agreement on that point, and I might have abused the editing towards my preference.  I'll refrain from doing it.

digama0: I prefer the original indentation (which I think is also mmj2 style). Right alignment is way too much work to keep in sync.

benjub: Ok. Reverted these alignments (the ones where the mmj2-style indentation was followed)

tirix (approved):

PR #2524, opened 2022-02-21
add conventions-contrib

benjub:
(as promised)
@digama0 : there is an implicit feature-wish for metamath-exe's command `write source set.mm/rewrap`.

Probably some more items will have to be moved from ~conventions to ~conventions-contrib, but this can be done progressively.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14754,6 +14754,7 @@ New usage of "con5VD" is discouraged (0 uses).
 New usage of "con5i" is discouraged (2 uses).
 New usage of "conss34OLD" is discouraged (0 uses).
 New usage of "conventions" is discouraged (0 uses).
+New usage of "conventions-contrib" is discouraged (0 uses).
 New usage of "conventions-label" is discouraged (0 uses).
 New usage of "coprmdvdsOLD" is discouraged (0 uses).
 New usage of "counop" is discouraged (0 uses).


jkingdon:
Didn't we view the pseudo-theorem (~ conventions , ~ conventions-contrib , etc ) style of formatting documentation as being not really needed now that we can process html from .raw.html files? (Or put the content a foo.md file in github for things like CONTRIBUTING.md and others in that style).

Having said that, I don't really object if people think it isn't a big deal.

avekens:
> (as promised) @digama0 : there is an implicit feature-wish for metamath-exe's command `write source set.mm/rewrap`.
> 
> Probably some more items will have to be moved from ~conventions to ~conventions-contrib, but this can be done progressively.

Yes, there are other items which would fit into the new section (e.g., "Maximum text line length is 79 characters."). But item "Comment and layout" is duplicated now, This at least should be cleaned up with PR.

avekens:
> Didn't we view the pseudo-theorem (~ conventions , ~ conventions-contrib , etc ) style of formatting documentation as being not really needed now that we can process html from .raw.html files? (Or put the content a foo.md file in github for things like CONTRIBUTING.md and others in that style).
> 
> Having said that, I don't really object if people think it isn't a big deal.

There are pros and cons: Often, it is very helpful to have the conventions right at hand when you are working on set.mm, On the other hand, the focus of set.mm is unclear (is it a library, a knowledge base, a handbook, or a collection of everything relevant?). Maybe, in future, set.mm will be split anyway (there were already a lot of discussions about that), then the conventions can be swapped out, too.

digama0:
I don't think the contrib should contain a literal header decoration. It's too easy to confuse for a real one, at least if this documentation is going in set.mm itself. There are a number of ways to get around this, like saying `repeat "#*#*" for 79 characters`. (Using a separate markdown file sidesteps all these issues.)

I'm not sure what the feature-wish is. Is this supposed to rewrap in some way?

benjub:
I made some edits in view of @avekens and @digama0's remarks.

As for where the conventions should be, what level of overlap is tolerated or beneficial, I admit that I have not thought of it much.

The present PR was mainly to address @avekens's comment about the recent commit which indented section titles, and does not have a wide scope.

@digama0 : the feature-wish is that, similarly to standard comments, the command `MM> write source set.mm/rewrap` would format section headers as indicated in the conventions" (possibly adding/removing blank lines, indenting, hard-wrapping). The goal is to get closer to a "normal form", thus having easier diffs and fewer merge conflicts.  I let you see if this is worth it.

avekens (approved):

jkingdon (approved):

PR #2525, opened 2022-02-22
Add set size function to iset.mm

jkingdon:
The set size function - http://us.metamath.org/mpeuni/df-hash.html has a few issues in intuitionizing.

The dumb one is that we already use ASCII `#`, typeset as `#`, for apartness ( http://us.metamath.org/ileuni/df-ap.html ). At least one source from the literature suggests a solution - to use a sharp sign as the set size function. So this pull request uses ASCII `size`, typeset as `♯`.

But there are also mathematical questions. The first is that set.mm defines set size in terms of cardinality. Cardinality as currently defined in iset.mm is pretty much stalled out, even for finite sets, due to #2203 . And we sort of know from the literature that cardinality won't be of much use beyond finite sets (granted, "much use" is a slightly vague concept). So it seems both practical and fitting to define the set size function independent of cardinality. I don't know if the definition here appears in the literature but it seems to work.

The definition here behaves well for a set which is finite (`A e. Fin`) or infinite (`_om ~<_ A`). We only have `_om ~<_ A -> -. A e. Fin` (included in this pull request as `infnfi`) not the converse and we mostly use `_om ~<_ A` (well except for one place where the stronger `-. A e. Fin` is possible, so why not?). Because of the inability to use case elimination or similar mechanisms to combine the finite and infinite cases, we don't have the ability to say that `♯` is defined on all sets ( http://us.metamath.org/mpeuni/hashf.html and similar).

There are a number of reverse closure theorems (``( ♯ ` A ) e. NN0 -> A e. Fin`` or similar) which are not included. As with other reverse closure theorems I suppose they aren't provable; at least the set.mm proofs don't seem to be adaptable.

Because the set size section is a big one it seemed better to submit what I have so far (which is enough to make me think that this definition is likely to work well and that what remains is likely to build on this rather than need to re-work it).

The pull request has a few smaller things less directly related to set size:
* `fict $p |- ( A e. Fin -> A ~<_ _om ) $= ...`, stated as in set.mm. I thought I'd need this one but didn't end up using it. Seems good to keep around, though, as I suppose we'll probably want it at some point.
* Adding `NN0*` which intuitionizes without trouble.
* `infn0 $p |- ( _om ~<_ A -> A =/= (/) ) $=`, stated as in set.mm (plus a similar version for inhabited rather than nonempty). Again, didn't end up using this but it seems worth adding.
* Mentioning a variety of theorems in mmil.html as being not provable, or presumably provable, or whatever. Can't think of anything to especially call out here - the details are in the pull request.
* Add `sumeq2` which is just a slight step on summation from the work in #2504 
* moving a few theorems to elsewhere in iset.mm
* the one change to set.mm is to change "CC" to "countable choice" in a comment because I'm not sure the former is self explanatory enough.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6341,6 +6341,13 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashkf , hashgval , hashginv</TD>
+  <TD><I>none</I></TD>
+  <TD>Due to the differences between df-hash in set.mm and
+  ~ df-size here, there's no particular need for these as stated</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6430,6 +6430,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hasheqf1od</TD>
+  <TD>~ sizeeqf1od</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6368,6 +6368,14 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashnnn0genn0</TD>
+  <TD><I>none</I></TD>
+  <TD>Not yet known whether this is provable or whether it is the
+  sort of reverse closure theorem that we (at least so far) have
+  been unable to intuitionize.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3103,6 +3103,13 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>fofi</TD>
+  <TD>~ f1ofi</TD>
+  <TD>Presumably precluded by an argument similar to ~ domfiexmid
+  (the set.mm proof relies on domfi).</TD>
+</TR>
+
+<TR>
   <TD>abrexfi</TD>
   <TD><I>none</I></TD>
   <TD>At first glance it would appear that the mapping would need to
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6397,6 +6397,13 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hasheni</TD>
+  <TD>~ sizeen , ~ sizeinf</TD>
+  <TD>It is not clear there would be any way to combine the finite
+  and infinite cases.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3060,6 +3060,13 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>xpfi</TD>
+  <TD><I>none</I></TD>
+  <TD>This is Lemma 8.1.16 of [AczelRathjen] and is therefore presumably
+  provable.</TD>
+</TR>
+
+<TR>
   <TD>prfi</TD>
   <TD><I>none</I></TD>
   <TD>The set.mm proof depends on unfi and it would appear that
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6359,6 +6359,15 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashfxnn0 , hashf , hashxnn0 , hashresfn , dmhashres ,
+  hashnn0pnf</TD>
+  <TD><I>none</I></TD>
+  <TD>Although ~ df-size is defined for finite sets and infinite
+  sets, it is not clear we would be able to show this definition
+  (or another definition) is defined for all sets.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6748,6 +6748,20 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>znnen</TD>
+  <TD><I>none</I></TD>
+  <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
+  The set.mm proof would not work as-is or with small changes, however.</TD>
+</TR>
+
+<TR>
+  <TD>qnnen</TD>
+  <TD><I>none</I></TD>
+  <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
+  The set.mm proof would not work as-is or with small changes, however.</TD>
+</TR>
+
+<TR>
   <TD>dvdsaddre2b</TD>
   <TD><I>none</I></TD>
   <TD>Something along these lines (perhaps with real changed to
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6446,6 +6446,21 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashxrcl</TD>
+  <TD><I>none</I></TD>
+  <TD>It is not clear there would be any way to combine the finite
+  and infinite cases.</TD>
+</TR>
+
+<TR>
+  <TD>hashclb , hashvnfin</TD>
+  <TD><I>none</I></TD>
+  <TD>Not yet known whether this is provable or whether it is the
+  sort of reverse closure theorem that we (at least so far) have
+  been unable to intuitionize.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6423,6 +6423,13 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashf1rn</TD>
+  <TD>~ sizef1rn</TD>
+  <TD>It is not clear there would be any way to combine the finite
+  and infinite cases.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6468,6 +6468,17 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashvnfin</TD>
+  <TD><I>none</I></TD>
+  <TD>This is a form of reverse closure, presumably not provable.</TD>
+</TR>
+
+<TR>
+  <TD>hashnfinnn0</TD>
+  <TD>~ sizeinf</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6404,6 +6404,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hasheqf1o</TD>
+  <TD>~ sizeeqf1o</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3036,6 +3036,13 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>fisupg</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on excluded middle and presumably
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -73673,6 +73720,35 @@ product of extended reals (common case).  (Contributed by David A.
...
   $}
 
+  $( Plus infinity belongs to the set of extended reals.  (Contributed by NM,
+     13-Oct-2005.)  (Proof shortened by Anthony Hart, 29-Aug-2011.) $)
+  pnfxr $p |- +oo e. RR* $=
+    ( cpnf cr cmnf cpr cun cxr ssun2 cc cuni cpw df-pnf cnex uniex pwex eqeltri
+    cvv prid1 sselii df-xr eleqtrri ) ABACDZEZFUAUBAUABGACAHIZJPKUCHLMNOQRST $.
+
+  $( Plus infinity exists (common case).  (Contributed by David A. Wheeler,
+     8-Dec-2018.) $)
```
benjub: What do you mean by "common case" ? That's the only case that I can think of...

iset.mm
```diff
@@ -61210,6 +61249,14 @@ would need to be strengthened (for example, to imply included middle) to
     ( wrel ccnv wfun cfn wcel w3a crn cdm df-rn relcnvfi 3adant2 simp2 syl5eqel
     fundmfi syl2anc ) ABZACZDZAEFZGZAHRIZEAJUAREFZSUBEFQTUCSAKLQSTMROPN $.
 
+  ${
+    $( If a 1-1 and onto function has a finite domain, its range is finite.
```
benjub: "a bijection" ?

iset.mm
```diff
@@ -82997,6 +83073,100 @@ Nonnegative integers (as a subset of complex numbers)
     $.
 
 
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+                              Extended nonnegative integers
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+
+  The function values of the hash (set size) function are either nonnegative
+  integers or positive infinity.  To avoid the need to distinguish between
```
benjub: "plus infinity" is used throughout

iset.mm
```diff
@@ -82997,6 +83073,100 @@ Nonnegative integers (as a subset of complex numbers)
...
+
+  $( The set of extended nonnegative integers. $)
+  cxnn0 $a class NN0* $.
+
+  $( Define the set of extended nonnegative integers that includes positive
+     infinity.  Analogue of the extension of the real numbers ` RR* ` , see
+     ~ df-xr .  (Contributed by AV, 10-Dec-2020.) $)
+  df-xnn0 $a |- NN0* = ( NN0 u. { +oo } ) $.
+
+  $( An extended nonnegative integer is either a standard nonnegative integer
```
benjub: Here and below, I would remove "standard" since it makes us think of nonstandard models, which are not at play here.

iset.mm
```diff
@@ -82997,6 +83073,100 @@ Nonnegative integers (as a subset of complex numbers)
     $.
 
 
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+                              Extended nonnegative integers
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+
+  The function values of the hash (set size) function are either nonnegative
```
benjub: Simply: "the set size function" ?

I do not know how standard it is, but I think calling the set size function the "hash" function because of the symbol generally denoting it, is bad practice (a bit like "the epsilon relation" for membership, or "the vertical bar letter" for I).

iset.mm
```diff
@@ -97573,6 +97724,256 @@ Knuth for iterated exponentiation (_Science_ 194, 1235-1242, 1976) and
...
+       finite sets).
+
+       This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
+       from the literature, but for finite sets should be equivalent to the
+       conventional definition that the size of a finite set is the unique
+       natural number which is equinumerous to the given set.  (Contributed by
+       Jim Kingdon, 19-Feb-2022.) $)
+    df-size $a |- size = ( ( frec (
+        ( x e. ZZ |-> ( x + 1 ) ) , 0 ) u. { <. _om , +oo >. } ) o.
+        ( x e. _V |-> U. { y e. ( _om u. { _om } ) | y ~<_ x } ) ) $.
```
benjub, line 97757: That's smart !

iset.mm
```diff
@@ -97573,6 +97724,256 @@ Knuth for iterated exponentiation (_Science_ 194, 1235-1242, 1976) and
...
+  $}
+
+  ${
+    $d A y z $.
+    $( The size (second half of ~ df-size ) of an infinite set is ` _om ` .
```
benjub, line 97762: "The size (second half of ~ df-size )"
Do you mean: "The ordinal size" ?

iset.mm
```diff
@@ -97573,6 +97724,256 @@ Knuth for iterated exponentiation (_Science_ 194, 1235-1242, 1976) and
...
+    $d A f $.  $d B f $.
+    $( The size of two finite sets is equal if and only if there is a bijection
+       mapping one of the sets onto the other.  (Contributed by Alexander van
+       der Vekens, 17-Dec-2017.) $)
+    sizeeqf1o $p |- ( ( A e. Fin /\ B e. Fin )
+        -> ( ( size ` A ) = ( size ` B ) <-> E. f f : A -1-1-onto-> B ) ) $=
+      ( cfn wcel wa csize cfv wceq cen wbr cv wf1o wex sizeen bren syl6bb ) ADE
+      BDEFAGHBGHIABJKABCLMCNABOABCPQ $.
+
+    $( There is no bijection between a finite set and an infinite set.  By
```
benjub, line 97899: Can you prove "there is no surjection" ?

iset.mm
```diff
@@ -97573,6 +97724,256 @@ Knuth for iterated exponentiation (_Science_ 194, 1235-1242, 1976) and
...
+      RABCUJABDUIUKT $.
+  $}
+
+  ${
+    $d A f $.  $d B f $.  $d F f $.
+    $( The size of a finite set which is a one-to-one function is equal to the
+       size of the function's range.  (Contributed by Jim Kingdon,
+       21-Feb-2022.) $)
+    sizef1rn $p |- ( ( A e. Fin /\ F : A -1-1-> B )
+        -> ( size ` F ) = ( size ` ran F ) ) $=
```
benjub, line 97936: Is this result needed somewhere ? It is very encoding-dependent.

iset.mm
```diff
@@ -97573,6 +97724,256 @@ Knuth for iterated exponentiation (_Science_ 194, 1235-1242, 1976) and
...
+    fiinfnf1o $p |- ( ( A e. Fin /\ -. B e. Fin )
+                      -> -. E. f f : A -1-1-onto-> B ) $=
+      ( cfn wcel cv wf1o wex f1ofi ex exlimdv con3dimp ) ADEZABCFZGZCHBDEZMOPCM
+      OPABNIJKL $.
+  $}
+
+  $( The codomain of an onto function is a set if its domain is a set.
+     (Contributed by AV, 4-May-2021.) $)
+  focdmex $p |- ( ( A e. V /\ F : A -onto-> B ) -> B e. _V ) $=
+    ( wcel wfo wa crn cvv wf fof anim2i ancomd rnexg 3syl wb forn eleq1d adantl
```
benjub, line 97912: should probably go earlier since not about size.

set.mm
```diff
@@ -111235,10 +111235,10 @@ enough that it can be proven using DC (see ~ axcc ).  It is, however,
   ${
     $d A b n y $.  $d b j k m n y $.
     domtriom.1 $e |- A e. _V $.
-    $( Trichotomy of equinumerosity for ` _om ` , proven using CC.
-       Equivalently, all Dedekind-finite sets (as in ~ isfin4-2 ) are finite in
-       the usual sense and conversely.  (Contributed by Mario Carneiro,
-       9-Feb-2013.) $)
+    $( Trichotomy of equinumerosity for ` _om ` , proven using countable
+       choice.  Equivalently, all Dedekind-finite sets (as in ~ isfin4-2 ) are
```
benjub, line 111239: clearer indeed

benjub (approved):

iset.mm
jkingdon: I believe that comment is intended to contrast it with http://us.metamath.org/mpeuni/pnfxr.html . I agree it isn't especially good wording but it is one we use a lot of places for convenience theorems like this.

iset.mm
jkingdon: We generally speaking use the "1-1" and "onto" terminology (with, I suppose, http://us.metamath.org/mpeuni/df-f1o.html as being the defining place) but I guess there's no reason to try to completely avoid "surjection" and "injection" and especially "bijection" :shrug: .

iset.mm
jkingdon: There are 13 usages of "positive infinity" in set.mm. I don't object to changing them to "plus infinity", though, if that's the preference (it seems to be by number, as there are 45 of those).

benjub: yes please, and also negative infinity --> minus infinity (in both set.mm and iset.mm, and not missing the ones with line breaks in the middle would be great)

iset.mm
jkingdon: I like "set size function". I think I've probably seen it in the literature somewhere but maybe just in set.mm. AczelRathjen seems to just say "This justifies that we call this number n the number of elements of A and denote it by ♯(A)."

iset.mm
jkingdon: Forged in the fires of failed cardinality definitions! But yeah, I was pleased that this one appears to work out.

iset.mm
jkingdon: I'm not in love with either one but "the ordinal size" seems to be the better of the two. It is kind of an intermediate lemma for something which doesn't naturally have a name, at least not a super obvious one.

iset.mm
jkingdon: Beats me. But if we don't like it we should remove it from set.mm too.

iset.mm
jkingdon: Dunno. set.mm just seems to have the bijection case as far as I've noticed so far anyway.

iset.mm
jkingdon: I don't object to moving it although I'd move it in both set.mm and iset.mm if so.

iset.mm
benjub: Sorry, I don't know this idiom "beats me". Does it mean that you disagree ? What I mean is that the cardinality or size of a function is not a very natural notion (in set.mm, a function is encoded as a set of couples (argument, value), so it is coincidental that this is equal to the cardinality of the domain).  If you consider that a function is "a rule that associates with each element of a set some element of another set", then speaking of its size has no real meaning.
If you want to keep it, maybe add, as in set.mm, "(Avoid depending on this [encoding?] detail)"

digama0: https://en.wiktionary.org/wiki/beats_me

digama0: It is relevant for some calculations, in particular for proving that functions are finite sets, and the relationship between `F` (as a set of pairs), `dom F` and `ran F`. It's not really encoding dependent; if we had another encoding then this would be a theorem about the function's graph (as a set of pairs).

jkingdon: Thanks for the wiktionary link, Mario. And in this case it means "I don't know" or "I don't know enough to have an opinion" or something like that.

benjub: So what about adding "(Avoid depending on this detail.)" ? (I checked it is the used formulation.)
For the present theorem, the important result is that for injections, size dom F = size ran F.

jkingdon:
I'm inclined to think this should be merged as-is (I did resolve conflicts and adopt the suggestion to say "ordinal size" in two of the comments).

I have a bunch of possible changes I'm thinking about concerning notation (`#` versus `size` and likewise for the typeset versions). And there are a lot of comments here suggesting changes to set.mm (which I'm not really sure what to do with - I'm not volunteering to act on them, at least not all of them).

But I'm not sure leaving this pull request unmerged helps with any of those.

jkingdon:
> have a bunch of possible changes I'm thinking about concerning notation (`#` versus `size` and likewise for the typeset versions)

I've written this up at #2528 - hopefully that does a good job of summing up the issues and why we might want to make or not make those changes.

david-a-wheeler:
Hi! I'm coming late to the party, but I thought I'd explain one item:

>> $( Plus infinity exists (common case).  (Contributed by David A. Wheeler, 8-Dec-2018.) $)

> What do you mean by "common case" ? That's the only case that I can think of...

The "(common case)" annotation is used for theorems that are simple in themselves (and thus easily rederived) but are reused so many times that they've been stated as their own theorems. They generally stem from various analyses I've done looking for common patterns (e.g., if something is re-derived 20 or 30 times, it's time to add it by itself). Adding theorems for common cases slightly simplifies the theorems that formerly re-derived them, and it can sometimes help automation (because it can detect an immediate match instead of having to figure something out).

jkingdon:
> The "(common case)" annotation is used for theorems that are simple in themselves

Maybe we can reword to "convenience theorem" or "convenience theorem version of ~ foo" or something? The concept makes sense to me but I'm not excited about "common case" as a way to describe it.

benjub:
> Maybe we can reword to "convenience theorem" or "convenience theorem version of ~ foo" or something? The concept makes sense to me but I'm not excited about "common case" as a way to describe it.

I commented on it at #2541. I think we can simply remove this kind of comments.  It would only add the burden to judge "how much" each of the 40000 statements is a convenience statement or is interesting in itself, which is a rather subjective thing.  To me, this adds more noise than information.

PR #2526, opened 2022-02-22
Relations and Subsets

mazsa:
I. Define the relations class.  Proper class relations (like ` _I ` , cf.
     ~ reli ) are not elements of it.  The element of this class and the
     relation predicate are the same when ` R ` is a set (cf. ~ elrelsrel ).

     The class of relations is a great tool we can use when we define classes
     of different relations as nullary class constants as required by the 2.
     point in our Guidelines ~ http://us.metamath.org/mpeuni/mathbox.html .
     When we want to define a specific class of relations as a nullary class
     constant, the appropriate method is the following:

     1.  We define the specific nullary class constant for general sets (cf.
     e.g. ~~? df-refs ), then

     2. we get the required class of relations by the intersection of the class
     of general sets above with the class of relations ~ df-rels (cf. ~~?
     df-refrels and the resulting ~~? dfrefrels2 and ~~? df-refrels3 ).

     3.  Finally, in order to be able to work with proper classes (like
     ~ iprc ) as well, we define the predicate of the relation (cf. ~~?
     df-refrel ) so that it is true for the relevant proper classes (cf. ~~?
     refrelid ), and that the element of the class of the required relations
     (e.g. ~~? elrefrels3 ) and this predicate are the same in case of sets
     (cf. ~~? elrefrelsrel ).

II. Define the subsets class or the class of all subset relations.  Similar
       to definitions of epsilon relation ( ~ df-eprel ) and identity relation
       ( ~ df-id ) classes.  The subset relation and subclass relationship
       ( ~ df-ss ) are the same, that is, ` ( A _S B <-> A C_ B ) ` when ` B `
       is a set, cf. ~ brssr .  Subset relation class and Scott Fenton's @sctfn subset
       class ~ df-sset are the same: ` _S = SSet ` (compare ~ dfssr2 with
       ~ df-sset , cf. comment of ~ df-xrn ), the only reason we do not use
       ~ dfssr2 as the base definition of the subsets class is the way we
       defined the epsilon relation and the identity relation classes.

       We need ` _S ` to define the class of reflexive sets so that we can
       define the class of converse reflexive sets with the help of the converse
       of ` _S ` , cf. ~~? df-refrels , ~~? df-refrel , ~~? elrefrelsrel vs.
       ~~? df-cnvrefrels , ~~? df-cnvrefrel , ~~? elcnvrefrelsrel .

       The subsets class ` _S ` has another place in set.mm as well: if we
       define extensional relation based on the common property in ~ extid ,
       ~ extep and ~ extssr , then extrelssr $p |- ExtRel _S is a theorem along
       with extrelep $p |- ExtRel _E and extrelid $p |- ExtRel _I .

       "df-ss" is an alternative tag for the class of subsets ` _S ` since the
       definition of subclass relation ` A C_ B ` , ~ df-ss should be "df-sc":
       please consider renaming it/them for consistency.


jkingdon (approved):

set.mm
```diff
@@ -556554,6 +556571,269 @@ range of a range Cartesian product ( ~ dfcoss4 ), but neither of these
...
+
+       The subsets class ` _S ` has another place in set.mm as well: if we
+       define extensional relation based on the common property in ~ extid ,
+       ~ extep and ~ extssr , then extrelssr $p |- ExtRel _S is a theorem along
+       with extrelep $p |- ExtRel _E and extrelid $p |- ExtRel _I .
+
+       "df-ss" is an alternative tag for the class of subsets ` _S ` since the
+       definition of subclass relation ` A C_ B ` , ~ df-ss should be "df-sc":
+       please consider renaming it/them for consistency.  (Contributed by Peter
+       Mazsa, 25-Jul-2019.) $)
```
tirix: Personally I would agree that our current `df-ss` actually denotes "**s**ub**c**lass" and not "**s**ubt**s**et", so `sc` would be a better fitting name.
However it's maybe not just about `df-ss`, we have a whole lot of theorems where `ss` appears in the name.

I believe this deserves a larger discussion. I'm not sure what is the best way (an email on the list, or maybe a GitHub issue, since it seems most of the community's discussions take place on GitHub recently)

benjub: Not very important in my opinion, but for this topic, the discussion group would probably be a better place.

jkingdon: > However it's maybe not just about `df-ss`, we have a whole lot of theorems where `ss` appears in the name.

As noted at http://us.metamath.org/mpeuni/conventions-label.html we have adopted "ss" as the thing we put in the name. I'd be fine with saying "subclass" more in comments but I'm perhaps a bit more reluctant to change that many names (or, probably worse, change a few and not get around to most of them).

mazsa: I think everybody above thinks 1. that sc would be optimal for C_ and ss for _S  (and el for e. , ep for _E and eq for = , id for _I), so arguing about this is not necessary 2. that smart changing ss to sc would be an enormous and not very satisfying task. Typically this is what we use money for. I think we should have I. a nice-to-have-but-boring tasks and services list (incl. e.g. https://doesmysiteneedhttps.com cf. https://groups.google.com/g/metamath/c/r9b63QxGd5w/ ) II. an explicit way how to donate for these tasks. We have perfectly functioning admins who can decide everything else:)

digama0: I prefer the name `df-ss` for subset for `C_`, despite the fact that it applies to classes. It is much more common to refer to this as the subset relation than subclass, especially since most of the time the arguments are in fact sets (and for pragmatic reasons we don't want to need to use different operations for sets). The way set.mm is set up, lots of things are technically classes despite morally (and provably) being sets, like `1` or `RR`. This has to do with the fact that there are no "set expressions": classes are expressions but there are only set variables in set.mm.

mazsa: @digama0 ok that would mean that we should keep labels (big relief!!) and this can be omitted from the boring-list (https is still on it:) Do you think we should mention this subclass/ss resolution in the comment section of df-ss? After that I would modify the text of the comment of df-ssr in my Mathbox by referring to df-ss.

digama0: Sure, documentation about the naming conventions is always welcome, if you want to add text to the df-ss comment.

mazsa: I'll do it tonight, mostly by copy&paste of your comment above.

avekens: I`m happy that df-ss is not changed. I'll review this PR after the comments are revised.

mazsa: Please review (and feel free to modify) the new "Modification" commit below, esp. the comment of df-ss (which is of course in the main set.mm).

avekens (approved):

benjub:
I'm not sure there is even a need to modify the comment of df-ss (or maybe less verbosely), but I'll let @digama0 tell what he prefers.  I think in his comment, "expression" should be replaced by "term" (i.e., "class" is the type for class terms, while in set.mm there is no type for set terms).

tirix (approved):

set.mm
jkingdon: The df-ss comment there looks OK to me. I think we may want to edit it in the future (for example to make it a bit shorter and to refer to some of our existing pages about classes, sets, terms, and variables), but I wouldn't hold up this pull request on that. I'd merge it now and worry about tweaking it later.

mazsa:
> I think in his comment, "expression" should be replaced by "term" (i.e., "class" is the type for class terms, while in set.mm there is no type for set terms).

Benoit, should I replace this or should we wait for Mario (I'm unable to decide it myself).

benjub:
@mazsa : as you prefer: say it and we can merge, and we can rephrase the comment later. @digama0 : isn't "term" more appropriate than "expression" (see my previous comment) ?

Also, @mazsa : the token C_ was chosen because it looks like \subseteq.  So maybe what you call `_S` can have token [C_]. But again, it's your choice and it can be done later.

tirix:
I guess the choice of `_S` is to follow the same template as `_I` ("set equality") and `_E` ("set membership")?

sctfn:
FWIW, it also matches nf.mm.

digama0:
@benjub They are somewhat interchangeable, but I prefer "expression" to emphasize that we're talking about the syntactic aspect of putting things together. People's views on these words might differ, but I think the metamath book uses the word "expression" for "sequence of constants and variables".

mazsa:
@benjub I'm not able to decide between "term" and "expression", I definitely lack the knowledge about that. What I guess is that if you are right, then we should review e.g. http://us.metamath.org/downloads/grammar-ambiguity.txt with a dozen occurences of "class expression*" as well. So this is well beyond the scope of this PR, but I think if you are sure of your own position, you should communicate with Mario, pehaps in the form of an own PR where you assign the reviewer role to Mario:)

` _S ` : I know about [C.], and [C_] has an appeal (because it reveals its origin C_ in it). So I would not be against the change - if and only if we change _I to [=] and _E to [e.] as well. But without this systematic change I'm in favor of keeping ` _S ` as it is.

(Edit: I wrote and publish this comment parallel with the comment of Mario above.)

benjub:
Thanks. So I guess we can merge as is.
I like your idea (for a potential future PR) to denote specific binary relations as follows:
```
token | unicode
[=]   | [=]
[e.]  | [∈]
[C_]  | [⊆]
[C.]  | [⊊] (http://us2.metamath.org/mpeuni/df-rpss.html)
```

PR #2527, opened 2022-02-22
misc typos

benjub:


set.mm
```diff
@@ -440817,65 +440817,65 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
...
-htmldef "SubGraph" as ' SubGraph ';
- althtmldef "SubGraph" as ' SubGraph ';
- latexdef "SubGraph" as "{\rm SubGraph}";
-htmldef "FinUSGraph" as ' FinUSGraph ';
- althtmldef "FinUSGraph" as ' FinUSGraph ';
- latexdef "FinUSGraph" as "{\rm FinUSGraph}";
-htmldef "NeighbVtx" as ' NeighbVtx ';
- althtmldef "NeighbVtx" as ' NeighbVtx ';
- latexdef "NeighbVtx" as "{\rm NeighbVtx}";
+  althtmldef "Edg" as 'Edg';
```
avekens, line 440838: I agree with removing the blanks for most of the "*Graph" classes, because these are not used with arguments/operands.

For SubGraph and NeighbVtx, however, I disagree: they are almost always used as relations ` S SubGraph G ` resp. operations ` ( G NeighbVtx N ) `. According to the conventions (item "Typography conventions"), the representations should  include leading and trailing blanks. Maybe this convention has to be enhanced for relations (see, for example, the html representation for ~QG.

avekens (changes requested):

set.mm
```diff
@@ -440801,8 +440801,8 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
   latexdef "mmu" as "\mu";
 htmldef "sigma" as
     " <IMG SRC='sigma.gif' WIDTH=10 HEIGHT=19 ALT=' sigma' TITLE='sigma'> ";
-  althtmldef "sigma" as " &sigma; ";
-  latexdef "sigma" as " \sigma ";
+  althtmldef "sigma" as "&sigma;";
+  latexdef "sigma" as "\sigma";
```
avekens: sigma is also an operation - therefore, there should be leading and trailing blanks (seee also html representation of "mod".

avekens (changes requested):

tirix (approved):

avekens (approved):

PR #2530, opened 2022-02-25
iset.mm: add set size theorems from isfinite4im to sizedom

jkingdon:
There isn't anything especially fundamentally new here, but it does add a bunch of set size theorems. Some of the theorems from set.mm in this range (isfinite4 to hashdomi) just won't work, some intuitionize with some changes, and a few need very few changes.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6484,6 +6484,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hasheq0</TD>
+  <TD>~ sizeeq0</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6514,6 +6514,13 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashrabrsn</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would need conditions around the existence of ` A ` and
+  decidability of ` ph ` but unused in set.mm.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6551,6 +6551,14 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashgval2</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably provable, when reestated as
+  ` ( size |`` _om ) = frec ( ( x e. ZZ |-> ( x + 1 ) ) , 0 ) ` ,
+  but lightly used in set.mm.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6528,6 +6528,13 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashrabsn1</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof uses excluded middle and this theorem is
+  unused in set.mm.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6521,6 +6521,13 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashrabsn01</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would need conditions around the existence of ` A ` and
+  decidability of ` ph ` but unused in set.mm.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6489,6 +6489,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashneq0</TD>
+  <TD>~ sizeneq0</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6546,6 +6546,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashgadd</TD>
+  <TD>~ omgadd</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6453,7 +6453,7 @@ than reals.</TD>
 </TR>
 
 <TR>
-  <TD>hashclb , hashvnfin</TD>
+  <TD>hashclb</TD>
   <TD><I>none</I></TD>
   <TD>Not yet known whether this is provable or whether it is the
   sort of reverse closure theorem that we (at least so far) have
@@ -6476,6 +6476,11 @@ than reals.</TD>
 <TR>
   <TD>hashnfinnn0</TD>
   <TD>~ sizeinf</TD>
+</TR>
+
+<TR>
+  <TD>isfinite4</TD>
+  <TD>~ isfinite4im</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6509,6 +6509,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashen1</TD>
+  <TD>~ sizeen1</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6499,6 +6499,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hash0</TD>
+  <TD>~ size0</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6489,7 +6489,7 @@ than reals.</TD>
 </TR>
 
 <TR>
-  <TD>hashneq0</TD>
+  <TD>hashneq0 , hashgt0n0</TD>
   <TD>~ sizeneq0</TD>
 </TR>
 
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6494,6 +6494,11 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashnncl</TD>
+  <TD>~ sizenncl</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6559,6 +6559,12 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashdom</TD>
+  <TD>~ sizedom</TD>
+  <TD>There is an added condition that ` B ` is finite.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6535,6 +6535,12 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashfn</TD>
+  <TD>~ sizefn</TD>
+  <TD>There is an added condition that the domain be finite.</TD>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -93987,6 +93987,22 @@ gives an ordered pair whose first element is the index (translated
...
+  ${
+    $d C x $.  $d ph x $.
+    frec2uzled.1 $e |- ( ph -> C e. ZZ ) $.
+    frec2uzled.2 $e |- G = frec ( ( x e. ZZ |-> ( x + 1 ) ) , C ) $.
+    frec2uzled.a $e |- ( ph -> A e. _om ) $.
+    frec2uzled.b $e |- ( ph -> B e. _om ) $.
+    $( The mapping ` G ` (see ~ frec2uz0d ) preserves order.  (Contributed by
+       Jim Kingdon, 24-Feb-2022.) $)
+    frec2uzled $p |- ( ph -> ( A C_ B <-> ( G ` A ) <_ ( G ` B ) ) ) $=
+      ( wcel wceq wo cfv wbr com wb syl2anc cz frec2uzzd clt wss cle wf1o f1of1
```
benjub, line 93999: I just had a quick look at http://us2.metamath.org/ileuni/frec2uz0d.html and that section.
Is it worth to define, either with an $a-stmt or in $e-stmts,
```
G = ( x e. CC |-> frec ( ( y e. CC |-> ( y + 1 ) ) , x ) )
```
(maybe replacing CC with ZZ or NN0) ?

jkingdon: Hmm, there are a few ways to answer that:

1. What we do now is mostly as the above comment suggests, except that the comment has an extra function, which I think would just need to be unwrapped. Pretty much everywhere this is used, `C` is a fixed value (generally `0`) so needing to apply http://us.metamath.org/ileuni/fvmpt.html (or whatever we use to remove the extra function) seems unnecessary.
2. I'm just following what set.mm does here (in http://us2.metamath.org/mpeuni/om2uz0i.html and following theorems), I didn't think too hard about whether there is a better way. The correspondence between set.mm and iset.mm can be seen by searching for om2uz0i at http://us.metamath.org/ileuni/mmil.html
3. I suppose `CC` would work; I don't really remember why I chose `ZZ` rather than `CC` on on the one hand or ``( ZZ>= ` C )`` on the other. Of course it would be work to change now.

iset.mm
```diff
@@ -98005,6 +98021,113 @@ This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
...
 
+  $( A finite set is equinumerous to the range of integers from one up to the
+     hash value of the set.  (Contributed by Jim Kingdon, 22-Feb-2022.) $)
+  isfinite4im $p |- ( A e. Fin -> ( 1 ... ( size ` A ) ) ~~ A ) $=
+    ( cfn wcel c1 csize cfv cfz co wceq cen wbr cn0 sizecl sizefz1 syl wb cz 1z
+    nn0zd fzfig sylancr sizeen mpancom mpbid ) ABCZDAEFZGHZEFUFIZUGAJKZUEUFLCUH
+    AMZUFNOUGBCZUEUHUIPUEDQCUFQCUKRUEUFUJSDUFTUAUGAUBUCUD $.
+
+  $( Two ways of saying a finite set is empty.  (Contributed by Jim Kingdon,
+     22-Feb-2022.) $)
```
avekens, line 98032: I do not know if there are agreements about or conventions for the contributors of theorems transferred from set.mm to iset.mm? Shouldn't the contributor of the theorem in set.mm be taken (unless there is a significant change of the proof)? For ~size0, for example, the proof seems to be exactly the same.

avekens (approved):

iset.mm
jkingdon: My practice has been to change the contributor  if I change the name and syntax but not to consider whether the proof is similar or not. It is an imperfect rule as you note but it does have the virtue of being easy to follow.

david-a-wheeler: @jkingdon - I think when in doubt we should give credit to previous contributors. In particular, if the proof is similar, also give credit to the original contributor. Having multiple contributors is fine!

jkingdon: We don't usually add a Contributed line for renames, do we? If so, I should just remove myself from most of the theorems in this pull request (which is fine with me) and use the Contributed line from set.mm.

As for the more general point, I'm not too sure and if anyone wants to audit Contributed lines against git logs I'm almost surely quite willing to let someone else try to figure out where to draw the lines. We definitely do have some practices around only adding these comments for the biggest or most defining contributions but I don't know if I always pick exactly the best place to draw the line (the history of contributions is, of course, credited in more detail in git logs).

iset.mm
```diff
@@ -98005,6 +98021,113 @@ This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
...
+  fseq1size $p |- ( ( N e. NN0 /\ F Fn ( 1 ... N ) ) -> ( size ` F ) = N ) $=
+    ( cn0 wcel c1 cfz co wfn csize cfv cfn wceq simpr simpl nn0zd fzfigd sizefn
+    wa 1zzd syl2anc sizefz1 adantr eqtrd ) BCDZAEBFGZHZRZAIJZUEIJZBUGUFUEKDUHUI
+    LUDUFMUGEBUGSUGBUDUFNOPUEAQTUDUIBLUFBUAUBUC $.
+
+  ${
+    $d A n z $.  $d B n $.  $d G n z $.
+    omgadd.1 $e |- G = frec ( ( x e. ZZ |-> ( x + 1 ) ) , 0 ) $.
+    $( Mapping ordinal addition to integer addition.  (Contributed by Jim
+       Kingdon, 24-Feb-2022.) $)
```
benjub, line 98094: Maybe add as a second sentence: "G is a morphism from finite ordinals with ordinal addition to nonnegative integers with integer addition."
The important thing is that "morphism" appears, since a person looking for that kind of result would probably search for the keyword "morphism".

iset.mm
```diff
@@ -98005,6 +98021,113 @@ This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
...
+    AEZAFAGHNOIJAKLM $.
+
+  $( The size of a singleton.  (Contributed by Jim Kingdon, 23-Feb-2022.) $)
+  sizesng $p |- ( A e. V -> ( size ` { A } ) = 1 ) $=
+    ( wcel csn csize cfv c1 cen wbr cz 1z en2sn mpan2 cfn wb snfig ax-mp sizeen
+    wceq sylancl mpbird cfz co cn0 1nn0 sizefz1 fzsn fveq2d syl5reqr syl6eq ) A
+    BCZADZEFZGDZEFZGUKUMUOSZULUNHIZUKGJCZUQKAGBJLMUKULNCUNNCZUPUQOABPURUSKGJPQU
+    LUNRTUAURUOGSKURGGGUBUCZEFZUOGUDCVAGSUEGUFQURUTUNEGUGUHUIQUJ $.
+
+  $( A set with only one element is equinumerous to the ordinal number 1.
```
benjub, line 98069: "A finite set has size 1 if and only if it is equinumerous to the ordinal 1."
(even though you have probably proved somewhere else that having size one is the same as having exactly one element)

iset.mm
```diff
@@ -98005,6 +98021,113 @@ This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
...
+    UNNZQUMUSUPIUOUMUNJAGAOZPRSUAUMUQLZUOUPUQUOTUMUOUNJUTUBQVBAGKZUPTUMUQVCVAUC
+    AGUJUDUEUMUNUFCUOUQUGAUHUNUIUKUL $.
+
+  $( The empty set has size zero.  (Contributed by Jim Kingdon,
+     23-Feb-2022.) $)
+  size0 $p |- ( size ` (/) ) = 0 $=
+    ( c0 csize cfv cc0 wceq eqid cfn wcel wb 0fin sizeeq0 ax-mp mpbir ) ABCDEZA
+    AEZAFAGHNOIJAKLM $.
+
+  $( The size of a singleton.  (Contributed by Jim Kingdon, 23-Feb-2022.) $)
```
benjub, line 98062: "... is 1." ?

iset.mm
```diff
@@ -98005,6 +98021,113 @@ This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
...
+  $( Two ways of saying a finite set is not empty.  Also, "A is inhabited"
+     would be equivalent by ~ fin0 .  (Contributed by Jim Kingdon,
+     23-Feb-2022.) $)
+  sizeneq0 $p |- ( A e. Fin -> ( 0 < ( size ` A ) <-> A =/= (/) ) ) $=
+    ( cfn csize cfv cc0 cap wbr wne clt c0 cz wb sizecl nn0zd 0zd zapne syl2anc
+    wcel cn0 cr cle nn0re nn0ge0 ap0gt0 syl sizeeq0 necon3bid 3bitr3d ) ABRZACD
+    ZEFGZUJEHZEUJIGZAJHUIUJKREKRUKULLUIUJAMZNUIOUJEPQUIUJSRZUKUMLZUNUOUJTREUJUA
+    GUPUJUBUJUCUJUDQUEUIUJEAJAUFUGUH $.
+
+  $( Positive natural closure of the hash function.  (Contributed by Jim
```
benjub, line 98047: "the size function"

iset.mm
benjub: Maybe add also: "A finite set has size zero if and only if it is empty."

iset.mm
```diff
@@ -98005,6 +98021,113 @@ This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
...
+  $( A function on a finite set is equinumerous to its domain.  (Contributed by
+     Jim Kingdon, 24-Feb-2022.) $)
+  sizefn $p |- ( ( F Fn A /\ A e. Fin ) -> ( size ` F ) = ( size ` A ) ) $=
+    ( wfn cfn wcel wa csize cfv wceq cen fndmeng ensymd wb fnfi sizeen sylancom
+    wbr mpbird ) BACZADEZFZBGHAGHIZBAJQZUAABADBKLSTBDEUBUCMABNBAOPR $.
+
+  $( The value of the size function on a finite 1-based sequence.  (Contributed
+     by Jim Kingdon, 24-Feb-2022.) $)
+  fseq1size $p |- ( ( N e. NN0 /\ F Fn ( 1 ... N ) ) -> ( size ` F ) = N ) $=
+    ( cn0 wcel c1 cfz co wfn csize cfv cfn wceq simpr simpl nn0zd fzfigd sizefn
```
benjub, line 98086: As in the last PR, in this and the previous result, the "size of a function" is a strange concept, but Mario said it could be useful sometimes, so, so be it.

iset.mm
```diff
@@ -98005,6 +98021,113 @@ This definition (in terms of ` U. ` and ` ~<_ ` ) is not taken directly
...
+      XDXJWAXQXDUSZXGXATKIZXIWAXQXGXSLXDWAXQUTZXGWTUEZDJXSXTXFYADBWSVAPXTAWTRDX
+      TUAEBWSVBVCVDVFXRXCTKIZWDXBTKIZKIZXSXIXRWDXBTWAXQWDVOGXDXPVGXQWAXBVOGXDXQ
+      XBFSWSDXOUOUNVHXRVIVJXDWAXSYBLXQXAXCTKVKVLXQWAXIYDLXDXQXHYCWDKXQAWSRDXQUA
+      EXQVPVCQVHVMVDVNVQVRVSVT $.
+  $}
+
+  ${
+    $d A m n x $.  $d B m n x $.
+    $( Dominance relation for the size function.  (Contributed by Jim Kingdon,
+       24-Feb-2022.) $)
```
benjub, line 98117: I would add "The size function preserves order, that is, it is a morphism from the finite sets with the dominance relation to the nonnegative integers with the less-than-or-equal-to relation".  But maybe I'm being fussy, plus this comment does not capture the biconditional. So, your choice!

benjub (approved):

iset.mm
benjub: As for attribution/credits, they indeed deal with two separate issues: (1) giving credit to the right person, (2) informing about the history/versions of a statement, without "ethical" concern.  Therefore, things are a bit vague when (1) and (2) lead to different solutions. For the present statement, which is minor, it is not very important in my opinion, and I'll go with what @jkingdon decides.

jkingdon: > Maybe add also: "A finite set has size zero if and only if it is empty."

When the statement of the theorem is as direct as ``A e. Fin -> ( ( size ` A ) = 0 <-> A = (/) )`` , I'm not sure translating that into text (almost word for word) is an improvement. Now, there are plenty of cases in which the comment can omit various technical details or otherwise restate the theorem productively so this isn't one size fits all, but at least for this one I won't propose changing it.

jkingdon: > As for attribution/credits

I've pushed up a change which brings over most of the contributed lines from set.mm. It turned out that even in this section there was more judgement call than I realized, as they run the gamut from "only the name of the theorem and `#` versus `size` is changed" to "the theorem is fairly similar but the set.mm one is for all sets and the iset.mm is only for finite sets" to even bigger differences. So hopefully I didn't do anything too far off base.

iset.mm
jkingdon: I guess I'll go with "the set size function".

iset.mm
jkingdon: I think I tend to like the terser comments in a case like this, where the theorem itself is so clear. Not that I have an especially strong opinion on it.

iset.mm
jkingdon: OK, I'll use that wording. And the same for set.mm without the word "finite".

When I was working on this I noticed that the comment only mentioned one direction of the biconditional, so that seems worth doing something about.

iset.mm
jkingdon: It's perfectly well defined given our definitions, so &lt;shrug&gt; keep it. I haven't looked at its usages more than briefly or formed much of an opinion whether there is a better way to prove the theorems where it is used.

iset.mm
jkingdon: Whether that kind of wording is helpful or off-putting probably depends on how much category theory the reader has seen. And I'm not sure how good morphism would be as a search term - set.mm already has 1367 occurrences.

I guess at least for now I'll stick with the terser wording.

iset.mm
jkingdon: As above I'll stick with the existing (terser) wording (which is taken from set.mm).

PR #2531, opened 2022-02-27
refrel cnvrefrel symrel refsymrel

mazsa:
4 sections: Reflexivity, Converse reflexivity, Symmetry, Reflexivity and symmetry.

A relation is reflexive iff: for all elements on its domain and range, if an element of its domain is the same as an element of its range, then there is the relation between them ( ~ dfrefrel3 ).

This definition is surprising: the conventional definition with the x R x part may be the result of that reflexivity is perceived primarily as a necessary part of equivalence. This seems to be the main reason why we thought so far that reflexivity on its own should be defined like the reflexivity part of ~~? dfeqvrel3 , i.e., by using the apparently self-evident x R x form, cf. section "Reflexivity and symmetry" below.

Another possible reason is that we are inclined to constrain ourselves to convenient special cases where the two forms are equivalent, e.g. the one with square Cartesian product:
idinxpssinxp3 $p |- ( ( I i^i ( A X. A ) ) C_ ( R i^i ( A X. A ) ) <-> ( I |` A ) C_ R ) and
idinxpssinxp4 $p |- ( A. x e. A A. y e. A ( x = y -> x R y ) <-> A. x e. A x R x ) .
Cf. https://github.com/metamath/set.mm/pull/1286

A relation is converse/ly reflexive iff: for all elements on its domain and range, if for an element of its domain and for an element of its range there is the relation between them, then the two elements are the same ( ~ dfcnvrefrel3 ).

We use this new general concept to define two special but conventional concepts, disjoints and functions.

“Converse reflexivity” can only exist if the “standalone” definition of reflexivity is revealed (as opposed to the x R x type of definition which is “mixed” with and obscured by symmetry). I like the "naughtiness" of the concept ("what on earth could be the converse of reflexivity??"), its name is technically correct, and I stick to it:) Of course all proofs are welcome if this concept and a concept X are the same.


jkingdon (approved):

avekens (approved):

set.mm
```diff
@@ -556883,6 +556950,626 @@ to definitions of epsilon relation ( ~ df-eprel ) and identity relation
...
+  df-symrels $a |- SymRels = ( Syms i^i Rels ) $.
+
+  $( Define the symmetric relation predicate.  (Read: ` R ` is a symmetric
+     relation.)  The element of the class of all symmetric relations and the
+     symmetric relation predicate are the same when ` R ` is a set, cf.
+     ~ elsymrelsrel .  Alternate definitions are ~ dfsymrel2 and ~ dfsymrel3 .
+     (Contributed by Peter Mazsa, 16-Jul-2021.) $)
+  df-symrel $a |- ( SymRel R <->
+      ( `' ( R i^i ( dom R X. ran R ) ) C_ ( R i^i ( dom R X. ran R ) ) /\
+        Rel R ) ) $.
```
tirix, line 557310: I believe you don't need both `Rel R` and `( R i^i ( dom R X. ran R ) )`.
Thanks tp `~ relssdmrn`, if you have the first, then the second is simply `R`.

tirix (approved):
Some of these definitions are quite elegant, it's interesting to find those similarities!

mazsa:
> Some of these definitions are quite elegant, it's interesting to find those similarities!

Thank you Thierry:)

set.mm
mazsa: Yes, we have this simple version 
``` dfsymrel2 $p |- ( SymRel R <-> ( `' R C_ R /\ Rel R ) ) ```
as well. The logic is that the base definition df- is always that version which is the most similar to the corresponding other base df- definitions. I.e., df-refrel df-cnvrefrel df-symrel and df-trsrel are the most _similar_ possible, their dfXrel2 and dfXrel3 versons are the most _simple_ possible versions of the similar base versions.

tirix: I see!
I suppose we usually do it the other way around, i.e. use the textbook / simplest possible statements as definition, and then prove the interesting/notable statements as theorems.

mazsa: You are right. However there is a tradeoff here: I think this is the simplest possible system of definitions with consistent numbering. I'll explain.
df-syms is the simplest definition, as I know, used only once, at least used only once by me:
``` df-syms $a |- Syms = { x | `' ( x i^i ( dom x X. ran x ) ) _S ( x i^i ( dom x X. ran x ) ) } ```
df-symrels is the simplest possible definition as well: 
``` df-symrels $a |- SymRels = ( Syms i^i Rels ) ```
Here comes the numbering:
``` dfsymrels2 ```  is ``` $p |- SymRels = { r e. Rels | `' r C_ r } ``` and
``` dfsymrels3 ``` is ``` $p |- SymRels = { r e. Rels | A. x A. y ( x r y -> y r x ) } ```
this logic of numbering of definitions continues long-long after reflexivity, symmetry etc. as well e.g.:
``` df-funsALTV $a |- FunsALTV = ( Funss i^i Rels ) ```
``` dffunsALTV2 $p |- FunsALTV = { f e. Rels | ,~ f C_ _I } ```
``` dffunsALTV3 $p |- FunsALTV = { f e. Rels | A. u A. x A. y ( ( u f x /\ u f y ) -> x = y ) } ```
We should continue this ``` df*2 ``` ``` df*3 ``` logic in case of predicate forms as well:
``` dfsymrel2 $p |- ( SymRel R <-> ( `' R C_ R /\ Rel R ) ) ```
``` dfsymrel3 $p |- ( SymRel R <-> ( A. x A. y ( x R y -> y R x ) /\ Rel R ) ) ```
and e.g.
``` dffunALTV2 $p  |- ( FunALTV F <-> ( ,~ F C_ _I /\ Rel F ) ) ```
``` dffunALTV3 $p |- ( FunALTV F <-> ( A. u A. x A. y ( ( u F x /\ u F y ) -> x = y ) /\ Rel F ) ) ```
Now the question is: what should be the base definition of the predicate form _supposed that ``` df*2 ``` and ``` df*3 ``` are given_?
The answer is that it should be the predicate form of the most simple form of the class of all related specific sets ( = similar to a simplest not-predicate form). In case of functions, both forms are quite simple:
``` df-funss $a |- Funss = { x | ,~ x e. CnvRefRels } ```
``` df-funALTV $a |- ( FunALTV F <-> ( CnvRefRel ,~ F /\ Rel F ) ) ```
in case of symmetry, they seem to be not so simple 
``` df-syms $a |- Syms = { x | `' ( x i^i ( dom x X. ran x ) ) _S ( x i^i ( dom x X. ran x ) ) } ```
``` df-symrel $a |- ( SymRel R <-> ( `' ( R i^i ( dom R X. ran R ) ) C_ ( R i^i ( dom R X. ran R ) ) /\ Rel R ) ) ```
but 1. df-syms is the simplest definition I guess and 2. df-symrel is the simplest possible definition supposed that we want to keep the consistent numbering.

mazsa: @tirix if you have an idea for a better labeling system, I'll consider to use it. But it seems to me that if we insist on consistent labeling&numbering, then either df-symrels or df-symrel can not be the simplest.

PR #2532, opened 2022-02-28
shorten cases2

wlammen:


set.mm
```diff
@@ -20469,10 +20459,43 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
       ZHACIZPDIZHQBAJKAPBLRTSUAABCEMPBDFMNO $.
   $}
 
+  $( Lemma for an alternate version of weak deduction theorem.  (Contributed by
```
jkingdon, line 20462: Is there a way to make this comment a bit more descriptive? Especially if it is now being using for things other than the weak deduction theorem.

(not a blocker for merging this, though).

jkingdon (approved):

avekens (approved):

set.mm
wlammen: There is an article http://us.metamath.org/mpeuni/mmdeduction.html on the Metamath pages that explains what purpose Norm in 1994 might have had in mind when he contributed these theorems.  I am not familiar with the theoretical  backgrounds, although I might have applied some ideas, when investigating how to arrive from a *i theorem at a *d on various occasions.

Maybe @benjub can tell more here, he introduced a bunch of if* theorems.  The dedlem* theorems more or less end up at these.

benjub: I think Wolf's proof cases2OLD should be relabeled cases2ALT (can you do it, @wlammen ?).  This is in case we want to revert to it because we want to modify dedlem[0](a|b) and prove cases2 without them.

As for the deduction theorem, even though it was superseded, as a practical tool, by @digama0's method of "deduction style", it is of course worth keeping in set.mm. I proposed a while ago on the google group, and Norm had agreed, to reformulate it and its lemmas using the conditional operator on propositions df-ifp, which would probably make it a bit clearer.  I guess it's on our todo list (maybe I'll open an issue?).  As @jkingdon notes, the comment should then also be reviewed.  But concerning dedlem0(a|b), they seem really adhoc (compare dfif2), not very natural, and a comment would only paraphrase the statement, its seems.

benjub (approved):
As commented above, I would propose:
```
$( Alternate proof of ~ cases2 , not using ~ dedlema or ~ dedlemb .
   (Contributed by BJ,  6-Apr-2019.)
   (Proof shortened by Wolf Lammen, 2-Jan-2020.)
  cases2ALT $p |- ......
```

set.mm
avekens: > I think Wolf's proof cases2OLD should be relabeled cases2ALT (can you do it, @wlammen ?). This is in case we want to revert to it because we want to modify dedlem[0](a%7Cb) and prove cases2 without them.

I agree.
> 
> As for the deduction theorem, even though it was superseded, as a practical tool, by @digama0's method of "deduction style", it is of course worth keeping in set.mm. I proposed a while ago on the google group, and Norm had agreed, to reformulate it and its lemmas using the conditional operator on propositions df-ifp, which would probably make it a bit clearer. I guess it's on our todo list (maybe I'll open an issue?). 

Yes, please open an isue for that.

> As @jkingdon notes, the comment should then also be reviewed. But concerning dedlem0(a|b), they seem really adhoc (compare dfif2), not very natural, and a comment would only paraphrase the statement, its seems.

I agree: changing the comment would only paraphrase the statement.

wlammen:
con3ALT is a how-to of using the weak deduction theorem.  Maybe that should be mentioned somewhere, too.

set.mm
benjub: Well, I didn't remember but apparently I already did that on 30-Sep-2019 according to the comments of elimh, dedt, con3ALT.  When I have time, I'll see if other results in the vicinity can be streamlined.
Edit: I also modified accordingly http://us2.metamath.org/mpeuni/mmdeduction.html on 16-Mar-2020 with Norm's agreement.

benjub:
> con3ALT is a how-to of using the weak deduction theorem. Maybe that should be mentioned somewhere, too.

This appears in the comment of con3ALT and on http://us2.metamath.org/mpeuni/mmdeduction.html (I remember having modified that page with Norm's agreement).

benjub (approved):

PR #2533, opened 2022-02-28
Graph theory - minor improvements of Huneke's proof (11): definition ClWWalksOnN

avekens:
Changes in main set.mm (Part 16 GRAPH THEORY only):
* header comment of section "  Closed walks as words" revised/enhanced
* definition ClWWalksOnN (closed walks anchored at a fixed vertex and of a fixed length as word) and related theorems added. This definition should replace the local definition F of theorems used for Huneke's proof of the friendship theorem (see ~ numclwwlkovf)
* ~clwwlksnndef and ~clwwlksn0 moved up

Special remarks:

1. The local definition ` *.f $e |-  F = ( v e. V , n e. NN |-> { w e. ( n ClWWalksN G ) | ( w `` 0 ) = v } ) $.` (closed walks anchored at a fixed vertex and of a fixed length as word, see ~ numclwwlkovf, etc.) plays a central role in Huneke's proof of the friendship theorem.  It is used in many theorems, and has a significance by itself, Therefore, I propose to introduce an explicit definition ` ClWWalksOnN ` for it. This PR contains this definition and related theorems - most of them can replace existing theorems using the local definition of F. This definition would also go well with the already existing definitions for (special) walks having fixed endpoints A and B (WalksOn, TrailsOn, WWalksNOn, etc.).
2. Concerning ` ClWWalksN ` (closed walks of a fixed length as word), I think it would be better to define it as function over ` NN0 ` instead of ` NN `, because then ~ clwwlksn0 could be proven directly by the definition ~ df-clwwlks of ` ClWWalks ` and not by using the specialty of set.mm that functions/operations evaluate to the empty set outside their domains (see item "Undefined results and junk theorems" in ~ conventions ). In contrast to ` ClWWalksN `, ` ClWWalksOnN ` is already based on ` NN0 ` instead of ` NN `, with ` ( X ( ClWWalksOnN `` G ) 0 ) = (/) ` (see ~clwwlkson0 ). I want to propose to change the definition of ` ClWWalksN ` accordingly, although a lot of theorems/proofs will be affected (That's why I want to get an prior approval before I start with it). 

@digama0 @benjub @tirix @david-a-wheeler and others: do you agree with my proposals?


benjub:
> * The local definition ``  *.f $e |-  F = ( v e. V , n e. NN |-> { w e. ( n ClWWalksN G ) | ( w `` 0 ) = v } ) $. ``

Why not use NN0 ?

This seems a bit specific to warrant a definition per se, but if it's used often enough in that Huneke proof, maybe it is worth it after all.  I haven't looked carefully enough...

> * Concerning `ClWWalksN` (closed walks of a fixed length as word), I think it would be better to define it as function over `NN0` instead of `NN`, because then ~ clwwlksn0 could be proven directly by the definition ~ df-clwwlks of `ClWWalks` and not by using the specialty of set.mm that functions/operations evaluate to the empty set outside their domains (see item "Undefined results and junk theorems" in ~ conventions ). In contrast to `ClWWalksN`, `ClWWalksOnN` is already based on `NN0` instead of `NN`, with `` ( X ( ClWWalksOnN `` G ) 0 ) = (/) `` (see ~clwwlkson0 ). I want to propose to change the definition of `ClWWalksN` accordingly, although a lot of theorems/proofs will be affected (That's why I want to get an prior approval before I start with it).

I agree.

avekens:
> > * The local definition ` *.f $e |-  F = ( v e. V , n e. NN |-> { w e. ( n ClWWalksN G ) | ( w` 0 ) = v } ) $. ``
> 
> Why not use NN0 ?

That is exactly my intention: ` F = ( v e. V , n e. NN |-> ...` should be replaced by ` ( ClWWalksOnN `` G ) = ( v e. ( Vtx `` G ) , n e. NN0 |-> ... ` .

> 
> This seems a bit specific to warrant a definition per se, but if it's used often enough in that Huneke proof, maybe it is worth it after all. I haven't looked carefully enough...

It is not more specific than the other ...On definitions, which were introduced 5 years ago on proposal of @digama0  - and ClWWalksOnN would be used much often than the others are currently used.

avekens:
If there is no strong objection against my proposals, I would like that this PR is accepted and merged, and I would continue with the realisation of my proposals.

benjub:
No objection to the proposals in this thread, but I do not have time to review the changes unfortunately. I can do it this weekend but I'll be happy if someone can do it before.

set.mm
```diff
@@ -419561,6 +419601,26 @@ the sequence p(0) p(1) ... p(n-1) of the vertices in a closed walk p(0)
       BUCUHCDABOPUFUJDCUGUDCLUEUIBUDCJRSTUA $.
   $}
 
+  ${
+    $d g n w $.
+    $( Conditions for ` ClWWalksN ` not being defined.  (Contributed by
```
benjub, line 419606: "Sufficient condition for..."
?

set.mm
```diff
@@ -421662,6 +421701,321 @@ corresponds to the size of the set of closed walks of length n (in an
       RXIXNVGVNVOVSVPVQVRTVTWAWBWC $.
   $}
 
+  ${
+    $d G g n v w $.
+    $( ` ( ClWWalksOnN `` G ) ` is an operator mapping a vertex ` X ` and a
+       nonnegative integer ` N ` to the set of closed walks on ` X ` of length
+       ` N ` in a graph ` G ` , as words over the set of vertices.
```
benjub, line 421708: X and N are called v and n in the statement.

set.mm
```diff
@@ -419491,6 +419515,22 @@ the sequence p(0) p(1) ... p(n-1) of the vertices in a closed walk p(0)
...
+       corresponds to the sequence v=p(0) p(1) ... p(n-1) of the vertices in a
+       closed walk p(0) e(f(1)) p(1) e(f(2)) ... p(n-1) e(f(n)) p(n)=p(0)=v as
+       defined in ~ df-clwlks .  The set ` ( ( v ( ClWWalksOnN `` g ) n ) `
+       corresponds to the set of "walks from v to v of length n" in a statement
+       of [Huneke] p. 2.  (Contributed by AV, 24-Feb-2022.) $)
+    df-clwwlksonn $a |- ClWWalksOnN = ( g e. _V
+                    |-> ( v e. ( Vtx ` g ) , n e. NN0
+                          |-> { w e. ( n ClWWalksN g ) | ( w ` 0 ) = v } ) ) $.
+  $}
+
```
benjub, line 419533: It looks like this definition is used much later ? If so, maybe put it later ?

benjub (approved):

tirix (approved):

set.mm
tirix: I usually look for definition right at the start of a chapter, just after the new constants are introduced, so maybe putting the definition later would make them a bit harder to find?

tirix:
> If there is no strong objection against my proposals, I would like that this PR is accepted and merged, and I would continue with the realisation of my proposals.

I agree that a domain of definition starting with 0 makes more sense. 
It's also nicer/easier to work with if it homogeneous among other definitions.

It looks like I'm the second approval, so I'll merge this PR.

set.mm
avekens: Only one of the disjuncts in the antecedent would already be sufficient. Therefore I would not add this here. Mario sometimes comments such theorems by "... equals the empty set when not meaningful" (see ~riotaund or ~oi0).

set.mm
avekens: OK, I will replace the variables accordingly.

set.mm
avekens: This is a general question about how sections with more than one definition should be structured. In my opinion, all definitions should be at the beginning, then the theorems should follow, grouped by theitr usage of the definitions (first group: only first defiition is used, second group: the first and the second definition used, etc.). This makes sence at least if the definitions are based on each other (as in the case of closed walks as words).

Maybe some basic theorems, however, should follow immediately after the definitions. If this is preferred, the section "Closed walks as words" could be structured as follows: 

  $c ClWWalks $.
  $c ClWWalksN $.
  $c ClWWalksOnN $.

  cclwwlks $a class ClWWalks $.
  cclwwlksn $a class ClWWalksN $.
  cclwwlksonn $a class ClWWalksOnN $.

    df-clwwlks $a |- ClWWalks = ... $.
    df-clwwlksn $a |- ClWWalksN = ... ClWWalks ... $.
    df-clwwlksonn $a |- ClWWalksOnN = ... ClWWalksN ... $.

    clwwlks $p |- ( ClWWalks ` G ) = ... $=
    isclwwlks $p |- ( W e. ( ClWWalks ` G ) <-> ... $=

    clwwlksn $p |- ( N e. NN -> ( N ClWWalksN G ) = ... $=
    isclwwlksn $p |- ( N e. NN -> ( W e. ( N ClWWalksN G ) <-> ... $=

    clwwlksonnmpt2 $p |- ( ClWWalksOnN ` G ) = ... $=
    clwwlksonn $p |- ( ( X e. V /\ N e. NN0 ) -> ( X ( ClWWalksOnN ` G ) N ) = ... $=
    isclwwlksonn $p |- ( ( X e. V /\ N e. NN0 ) -> ( W e. ( X ( ClWWalksOnN ` G ) N ) <-> ... $=

    [theorems using ClWWalks only]

    [theorems using ClWWalks and/or ClWWalksN]

    [theorems using ClWWalks and/or ClWWalksN and/or ClWWalksOnN]

If acceptable, I would move up the theorems ~ clwwlksonnmpt2 , ~clwwlksonn and ~ isclwwlksonn.

set.mm
avekens: ... in my next PR!

set.mm
avekens: Since @tirix merged this PR already (thank you), I could move up the theorems ~ clwwlksonnmpt2 , ~clwwlksonn and ~ isclwwlksonn in my next PR. Maybe we should discuss my proposal for structuring sections with more than one definition separately (shall I open an issue for it?) and add it as convention if it is accepted.

benjub: Then I suggest you have three sections (either add two sections at the same level, or subdivide the section into three subsections), titled e.g. "Closed walks", "Closed walks based at a vertex", "Closed walks based at a vertex of a given length".  It is easier to be introduced to a new notion one at a time.

avekens: > Then I suggest you have three sections (either add two sections at the same level, or subdivide the section into three subsections), titled e.g. "Closed walks", "Closed walks based at a vertex", "Closed walks based at a vertex of a given length". It is easier to be introduced to a new notion one at a time.

So you propose that we should have at most one definition in a section in general? With our four header levels thisould maybe be possible, but I can remember that Norm was not a friend of such an approach. If we want to follow this suggestion, we have to start with section "Logical disjunction and conjunction", which should be split into two sections "Logical disjunction" and "Logical conjunction" and so forth. We should have a broader discussion on this...

benjub: This would have my preference in most cases, but this does not have to be a rule.  My comment above was about one specific case. As for the broader discussion... maybe, but I do not have the time to launch it now. So do as you prefer.

PR #2534, opened 2022-03-02
Add set size theorems to iset.mm from sizeun to pr0size2ex

jkingdon:
The intuitionizing here is pretty similar to previous `size` theorems with needed changes mostly falling into existing categories like not being able to combine the finite case and in the infinite case, not having reverse closure, etc. The going is slower than, say number theory, but I suppose we've known that about finite sets for a while.

A few specific notes:
* `unfidisj` has been listed as "presumably provable" in mmil.html for a while. Now it is proven.
* The `sizeun` (`hashun`) proof is not super difficult but did take some effort and is a chance to dredge up some of those finite set theorems I've been working on in recent months.
* Perhaps surprisingly, `hashun2` doesn't seem to fall as readily as `hashun`. The pull request adds a few comments to the web page about topics like http://us.metamath.org/mpeuni/fodomfi.html but many of these are left for another day.
* `1domsn` is a cute little result because it is the sort of thing which set.mm might tend to prove by case elimination but which at least for this theorem can be proved without it.
* `ominf` is, as noted, not really the way we'd say "_om is infinite" but prove it anyway because we might as well prove something which is in set.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6687,6 +6687,14 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashgt0elex , hashgt0elexb</TD>
+  <TD>~ sizeneq0</TD>
+  <TD>See ~ fin0 for inhabited versus non-empty. It isn't clear
+  it would be possible to also include the infinite case as
+  hashgt0elex does.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6597,6 +6597,12 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashun3</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on various theorems we do not have</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2922,13 +2922,6 @@ and thus prove fisseneq or something which easily implies it.</TD>
 </TR>
 
 <TR>
-<TD>ominf</TD>
-<TD><I>none</I></TD>
-<TD>Although this theorem presumably could be proved, it would
-probably need a very different proof than the set.mm one</TD>
-</TR>
-
-<TR>
 <TD>isinf</TD>
 <TD><I>none</I></TD>
 <TD>The set.mm proof uses the converse of ~ ssdif0im</TD>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -61226,6 +61246,23 @@ would need to be strengthened (for example, to imply included middle) to
       WMVPRXEVQVRVSVTVTWJWAWB $.
   $}
 
+  ${
+    $d A w y z $.  $d B w y z $.
+    $( The union of two disjoint finite sets is finite.  (Contributed by Jim
+       Kingdon, 25-Feb-2022.) $)
+    unfidisj $p |- ( ( A e. Fin /\ B e. Fin /\ ( A i^i B ) = (/) )
+        -> ( A u. B ) e. Fin ) $=
```
avekens, line 61254: Why is ( A i^i B ) = (/) necessary? Is ( ( A e. Fin /\ B e. Fin ) -> ( A u. B ) e. Fin ) not provable (see ~unfi in set.mm)?

avekens (changes requested):

iset.mm
jkingdon: unfi is not provable. As we say in http://us.metamath.org/ileuni/mmil.html , "Not provable according to Remark 8.1.17 of [AczelRathjen]"

Although we don't have a specific proof that unfi implies excluded middle, I suppose it does and that a proof along the lines of http://us.metamath.org/ileuni/ssfiexmid.html or http://us.metamath.org/ileuni/domfiexmid.html would work (although I'm not completely sure especially about what the proof would be).

avekens: OK, thanks a lot for the explanation.

avekens (approved):

iset.mm
digama0: Let `X = { x e. 1o | ph }` be my favorite ordinal. Then `{ 1o } e. Fin` and `{ X } e. Fin` but if `( { 1o } u. { X } ) e. Fin` then either this set has cardinality 1 and `X = 1o` (so `ph`), or the set has cardinality 2 and `X =/= 1o` (hence `-. ph`, since `ph` implies `X = 1o`).

digama0: A stronger version of the theorem that may be easier to apply in practice is
`( ( A e. Fin /\ B e. Fin /\ ( A i^i B ) e. Fin ) -> ( A u. B ) e. Fin )`. I think that's even a biconditional:
`( ( A e. Fin /\ B e. Fin ) -> ( ( A u. B ) e. Fin <-> ( A i^i B ) e. Fin ) )`.
Another variation that looks helpful in proving this is
`( ( A e. Fin /\ B e. Fin /\ B C_ A ) -> A = ( B u. ( A \ B ) ) )` (and `A\B` is also finite)

jkingdon: Ooh, nice, I'll make a separate pull request for "unfi implies excluded middle" (which formalizes pretty much exactly as stated above) and probably for the others too.

jkingdon: This seems to be going well but since it probably won't be done today I'll put the progress so far in a draft pull request at #2536

PR #2539, opened 2022-03-07
iset.mm: intuitionize sumrb and a few more theorems

jkingdon:
Most of these intuitionize relatively straightforwardly.

* includes several `seq` related theorems, intuitionized much in the style of the existing seq theorems
* includes some minor additions to mmil.html, around CZF and the Theorem Sampler
* the one change to set.mm is shortening the proof of fz1f1o by using isfinite4

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7132,6 +7132,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>fsumcvg</TD>
+  <TD>~ fisumcvg</TD>
+</TR>
+
+<TR>
   <TD>df-prod and theorems using it</TD>
   <TD><I>none</I></TD>
   <TD>To define this, will need to tackle all the issues
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -817,11 +817,25 @@ variables</A></LI>
 <B>Real and complex numbers</B>
 
 <MENU>
+<LI> <A HREF="arch.html">Archimedean property of real numbers</A></LI>
+
 <LI>Properties of apartness:
 <A HREF="apirr.html">1</A>
 <A HREF="apsym.html">2</A>
 <A HREF="apcotr.html">3</A>
 <A HREF="apti.html">4</A></LI>
+
+<LI> <A HREF="sqrt2irrap.html">The square root of 2 is irrational</A> (a
+different statement than <A HREF="sqrt2irr.html">The square root of 2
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -171599,11 +171599,10 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) ) ~~> x ) \/
...
                    E. f f : ( 1 ... ( # ` A ) ) -1-1-onto-> A ) ) ) $=
-      ( cfn wcel chash cfv cc0 wceq cn wo c0 c1 cfz co cv wf1o wex sylib mpbid
-      wa cn0 hashcl orcomd hasheq0 cen wbr hashfz1 syl wb fzfi hashen mpan bren
-      elnn0 biantrud orbi12d ) ACDZAEFZGHZURIDZJAKHZUTLURMNZABOPBQZTZJUQUTUSUQU
-      RUADZUTUSJAUBZURUNRUCUQUSVAUTVDACUDUQVCUTUQVBAUEUFZVCUQVBEFURHZVGUQVEVHVF
-      URUGUHVBCDUQVHVGUILURUJVBAUKULSVBABUMRUOUPS $.
+      ( cfn wcel chash cfv cc0 wceq cn wo c0 c1 cfz co cv wf1o wex cn0 hashcl
+      wa elnn0 sylib orcomd hasheq0 cen wbr isfinite4 bren sylbb biantrud mpbid
+      orbi12d ) ACDZAEFZGHZUNIDZJAKHZUPLUNMNZABOPBQZTZJUMUPUOUMUNRDUPUOJASUNUAU
+      BUCUMUOUQUPUTACUDUMUSUPUMURAUEUFUSAUGURABUHUIUJULUK $.
```
tirix, line 171605: Feel free to add the "Shortened by" contribution tag!

tirix (approved):
Looks good!

set.mm
jkingdon: I realize it is an option, but don't put it in for my sake.

I don't know if the level of effort is a huge factor, but in this case all that happened was that the minimizer was capable of using `isfinite4im` (from iset.mm) but not `isfinite4` (from set.mm) because it would have needed to insert a `biimpi` or something of the sort. So what I did was only one step removed from running the minimizer, which by longstanding practice does not result in a "Shortened by" tag.

tirix: I see. Anyway, good catch!

PR #2541, opened 2022-03-10
Document "(common case)" in conventions

david-a-wheeler:
Many theorems are marked "(common case)" but that notation
was never explained anywhere.

This commit documents what "(common case)" means in the
"conventions" section. While we're there, this also provides
a mnemonic for proper substitution notation (taken straight from
the Metamath book).


benjub:
I'd rather remove the phrase "common case" altogether. That a statement is used often can be checked either by "MM> show usage" or on the webpage of each statement ("This theorem is referenced by:...").  The fact that a statement is used often is a reason to add it to set.mm, but once it is there, it is not necessary to specify it (even if only to indicate that it is "not interesting by itself but only here because used often").

jkingdon:
I'd agree with 'remove the phrase "common case" altogether.' I'm not sure there is any especially clear distinction between theorems which have this and theorems which don't, and all of the comments I can remember would read just fine with "common case" omitted.

avekens:
Before I read David's explanation, I had a completely different understanding what "common case" should mean (kind of "general version of ..." or "mostly used version of ..."), but if I understand it correctly, it should mean "variant of ... which is often used in proofs". If I have a look at ~pnfnemnf (+oo =/= -oo) and ~mnfnepnf (-oo =/= +oo ), for example, why should the second be a "commen case" and the first not?

Because such an attribution is more confusing than helpful, I would agree with BJ and JK to remove it from the comments (and not to add it to the conventions). Maybe it can be replaced by "Variant of ~... ." (for example, in the comment of ~mnfnepnf: "Variant of ~pnfnemnf .") to give a hint to the "main" theorem.

david-a-wheeler:
The original reason to add "(common case)" text was because it served more as a hint to Norm: "don't remove this even though it's focused and the proof is short". Norm tended to prefer general theorems with long proofs, but these "common case" theorems tended to be specific & short.

I think today it's unlikely we'll remove a widely-used theorem just because its proof is short. So I'm okay with removing all the "(common case") notations as an alternative.

However, I think we need to *either*:
1. document "(common case)" (as is done in this pull request), or
2. remove the phrase "(common case)" in the database (I could do that instead).

david-a-wheeler:
> if I understand it correctly, it should mean "variant of ... which is often used in proofs".

The "common case" theorems aren't necessarily a variant of anything. I have a simple script that looks for intermediate steps that are re-proved in many different theorems. If some intermediate step is proven 20 or 40 times, it makes sense to prove it directly once, then reuse the results everywhere. That's a useful thing to know when arguing for its addition. That said, once it's added it's not clear we need to document that within set.mm itself. So we should document that or remove the marking.

avekens:
> > if I understand it correctly, it should mean "variant of ... which is often used in proofs".
> 
> The "common case" theorems aren't necessarily a variant of anything.

OK, not necessarily, but often or at leat sometimes. Then this could be documented in the comment (as in my example).

> I have a simple script that looks for intermediate steps that are re-proved in many different theorems. If some intermediate step is proven 20 or 40 times, it makes sense to prove it directly once, then reuse the results everywhere. That's a useful thing to know when arguing for its addition. 

That's very interesting. When/how do you run this script? How does it work? Does it detect only simple steps consisting of 2 or three lines? Can you make it available for everybody? Can the minimize script be executed afterwards to minimize the proofs found by your script?

tirix:
I would also agree to go with "option 2", removing the "(common case)" tags.

avekens:
I also vote vor option 2.

david-a-wheeler:
Looks like there's a general consensus for removing the marking "(common case)", instead of explaining it. So I'm going to close this issue, and submit a separate issue to remove the marking.

david-a-wheeler:
Let's remove them instead, see: #2544

david-a-wheeler:
> That's very interesting. When/how do you run this script? How does it work? Does it detect only simple steps consisting of 2 or three lines? Can you make it available for everybody? Can the minimize script be executed afterwards to minimize the proofs found by your script?

The script is here in set.mm:
scripts/find-repeats

david-a-wheeler:
> That's very interesting. When/how do you run this script?

Just every once in a great while.

> How does it work?

It prints every line of every proof, counts how often each one occurs, & shows them sorted by how often it occurs.

> Does it detect only simple steps consisting of 2 or three lines?

No. However, it becomes increasingly unlikely for something to repeatedly get proved if it has many steps.

> Can you make it available for everybody?

Already is, it's `scripts/find-repeats`

> Can the minimize script be executed afterwards to minimize the proofs found by your script?

Absolutely!! Once you create a separate proof, you can minimize the result.

PR #2551, opened 2022-03-19
Add the ability to recurse up to a certain point to iset.mm

jkingdon:
Before I say what this is, let me give the motivation.

A number of existing theorems require that a function (which we will use for recursion) has to be defined "everywhere" (for varying definitions of "everywhere" and "defined"). For example, in http://us.metamath.org/ileuni/iseqfveq.html there is the condition `(𝐹‘𝑥) ∈ 𝑆` (for all x above a starting integer) whereas the set.mm counterpart, http://us.metamath.org/mpeuni/seqfveq.html , has no constraints on ` F ` outside `(𝑀...𝑁)`. Is this a problem or just something we can figure out how to live with? To be honest, I'm not completely sure, and every time I try to get my head around the full picture (from `recs` to `frec` to `seq` to `sum`), I easily get lost and have trouble understanding what changing a condition on `recs` theorem means for `sum` and everything in between.

One theorem which is pretty complicated, apparently needed for summation, and which appears to be affected by this issue is http://us.metamath.org/mpeuni/seqf1o.html . There is currently no version of this in iset.mm and although I'm not sure I can quite see fully how it could intuitionize and what the obstacles will be, I'm guessing it is harder with http://us.metamath.org/ileuni/iseqfveq.html (and others) in their current form.

Anyway, that's the motivation. What this pull request does is take http://us.metamath.org/ileuni/tfri1d.html and relaxes the `(𝐺‘𝑥) ∈ V` requirement. The result is:
```
21600 tfr1on.f $e |- F = recs ( G ) $.
21601 tfr1on.g $e |- ( ph -> Fun G ) $.
21602 tfr1on.x $e |- ( ph -> X e. On ) $.
21603 tfr1on.ex $e |- ( ( ph /\ x e. X /\ f Fn x ) -> ( G ` f ) e. _V ) $.
21751 tfr1on.u $e |- ( ( ph /\ x e. U. X ) -> suc x e. X ) $.
21752 tfr1on.yx $e |- ( ph -> Y e. X ) $.
21753 tfr1on $p |- ( ph -> Y C_ dom F ) $= ... $.
```

where we are defining ``( G ` f )`` up to an ordinal `X`. Whether a stronger result than `Y C_ dom F` (where `Y` is any element of `X`) would have been possible is not clear to me, but this is what seemed possible based on adapting the tfri1d proof without large changes.

The condition that the converse of http://us.metamath.org/ileuni/sucunielr.html needs to hold surprised me a bit in writing the proof (for tfri1d we don't need this as such because http://us.metamath.org/ileuni/suceloni.html suffices). But it shouldn't be a problem for the `frec` or `seq` related cases because we are able to prove `B e. _om -> ( A e. U. B <-> suc A e. B )` (included in this pull request). The pull request includes a few notes in mmil.html which relate to the converse of sucunielr.

The addition of `tfr0dm` is not because I'm sure we'll have a use for it, but because it is easy to break out this part of the `tfr0` proof.

The change to `tfrfun` is just to move it earlier in the file so we can use it in the added proofs.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2563,14 +2563,17 @@ middle as shown at ~ 0elsucexmid .</TD>
 
 <TR>
 <TD>limsuc</TD>
-<TD><I>none yet</I></TD>
-<TD>Conjectured to be provable.</TD>
+<TD><I>none</I></TD>
+<TD>This would be trivial if dflim4 were the definition of a limit ordinal.
+With ~ dflim2 as the definition, limsuc might need ordunisuc2 (which we
+believe is not provable, see its entry in this list).</TD>
 </TR>
 
 <TR>
 <TD>limsssuc</TD>
-<TD><I>none yet</I></TD>
-<TD>Conjectured to be provable.</TD>
+<TD><I>none</I></TD>
+<TD>Conjectured to be provable (is this also based on dflim4 being
+the definition of limit ordinal or is it unrelated?).</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2474,7 +2474,7 @@ middle as seen at ~ ordpwsucexmid .</TD>
 
 <TR>
 <TD>ordsucuniel</TD>
-<TD>~ sucunielr</TD>
+<TD>~ sucunielr , ~ nnsucuniel</TD>
 <TD>Full ordsucuniel implies excluded middle, as shown at ~ ordsucunielexmid .</TD>
 </TR>
 


digama0 (approved):
What is the reason that http://us.metamath.org/ileuni/frecuzrdg0.html uses a function defined on `` ( ZZ>=` C ) X. S `` instead of ``_V X. _V`` like set.mm? I imagine it has something to do with the general requirement for function values to be in domain but it seems like it would be fine as long as you never attempt to evaluate the function whenever it is not in range. That is, you can get away with a much weaker requirement than assuming (x F y) is closed in S, you can just say that it works when evaluated once, or it is a set or something along those lines.

jkingdon:
> What is the reason that http://us.metamath.org/ileuni/frecuzrdg0.html uses a function defined on `` ( ZZ>=` C ) X. S `` instead of `_V X. _V` like set.mm? I imagine it has something to do with the general requirement for function values to be in domain

I'm sure it was something like that but let me see if I can look around and find a more specific answer. Yeah, here it is: http://us.metamath.org/ileuni/frecuzrdgrom.html uses http://us.metamath.org/ileuni/mpt2exga.html to show that the function is a set and not a proper class, which it in turn uses to satisfy a `(∀𝑧(𝐹‘𝑧) ∈ V` hypothesis via http://us.metamath.org/ileuni/fvexg.html . That wouldn't work if we replaced either `` ( ZZ>=` C ) `` or ` S ` with a proper class (although perhaps it would be nice if ` S ` could be a proper class because I think it might simplify some of the `seq` theorems and maybe even let us use the set.mm two-argument version of `seq` rather than the three-argument version which specifies a value of ` S ` every time we use `seq` ).

> but it seems like it would be fine as long as you never attempt to evaluate the function whenever it is not in range. That is, you can get away with a much weaker requirement than assuming (x F y) is closed in S, you can just say that it works when evaluated once, or it is a set or something along those lines.

Well, ` F ` is (at least at http://us.metamath.org/ileuni/iseq1.html ) `(𝑧 ∈ (ℤ≥‘𝑀), 𝑤 ∈ 𝑆 ↦ (𝑤 + (𝐹‘(𝑧 + 1))))` so I guess it is currently a set but wouldn't be if we switched to ` _V X. _V ` .

Thanks for taking a look at some of this, and do keep the ideas coming. Even in set.mm it isn't quite as simple as either "it is all about _V and things being defined everywhere" or "it is all about closure theorems" ( http://us.metamath.org/mpeuni/seqcl.html is widely used to give an example of the latter). And little of it intuitionizes easily. So I very much appreciate your past contributions to this area and would welcome future ones as you get ideas and/or have time to look at it.

digama0:
> > What is the reason that http://us.metamath.org/ileuni/frecuzrdg0.html uses a function defined on `` ( ZZ>=` C ) X. S `` instead of `_V X. _V` like set.mm? I imagine it has something to do with the general requirement for function values to be in domain
> 
> I'm sure it was something like that but let me see if I can look around and find a more specific answer. Yeah, here it is: http://us.metamath.org/ileuni/frecuzrdgrom.html uses http://us.metamath.org/ileuni/mpt2exga.html to show that the function is a set and not a proper class, which it in turn uses to satisfy a `(∀𝑧(𝐹‘𝑧) ∈ V` hypothesis via http://us.metamath.org/ileuni/fvexg.html . That wouldn't work if we replaced either `` ( ZZ>=` C ) `` or `S` with a proper class (although perhaps it would be nice if `S` could be a proper class because I think it might simplify some of the `seq` theorems and maybe even let us use the set.mm two-argument version of `seq` rather than the three-argument version which specifies a value of `S` every time we use `seq` ).

Hm, this is probably not an easy suggestion to act on, but I wonder whether we can cut down all these theorems to have minimal hypotheses. For example, rather than even proving that `frec Fn om` we prove that it is defined and unique on 0..3 if ``A, F(A), F(F(A)), F(F(F(A)))`` are all sets. That applies to `recs` too, you can talk about it being partially defined on an initial ordinal if the given function evaluates successfully at all the previous values.

For `iseqfveq` you have an even harder problem: even if the values are not sets, the value of `frec(F, A)(3)` is fully determined by the values of `A, F(A), F(F(A)), F(F(F(A)))`. This is tricky since we don't even know that we are evaluating the second stage from the results of the first stage (since `F(A)` is not necessarily a set), so it might evaluate `F` somewhere else which is bad. Most likely this would require breaking down the function value operation and considering everything relationally, and it's not clear to me whether this actually holds.

jkingdon:
> Hm, this is probably not an easy suggestion to act on, but I wonder whether we can cut down all these theorems to have minimal hypotheses. For example, rather than even proving that `frec Fn om` we prove that it is defined and unique on 0..3 if `A, F(A), F(F(A)), F(F(F(A)))` are all sets.

I guess this is sort of the thought process that led me thinking in the direction of ``∀ x ∈ S ( F ` x ) ∈ S`` style closure (to replace the existing `∀𝑧(𝐹‘𝑧) ∈ _V`). Although of course even that still isn't minimal because although it does restrict where we have to be able to apply ` F `, it doesn't restrict it all the way to "only apply it three times".

> That applies to `recs` too, you can talk about it being partially defined on an initial ordinal if the given function evaluates successfully at all the previous values.

That's expressed in [tfr1on](http://us2.metamath.org/ileuni/tfr1on.html) as a ` ((𝜑 ∧ 𝑥 ∈ 𝑋 ∧ 𝑓 Fn 𝑥) → (𝐺‘𝑓) ∈ V)` hypothesis.

> For `iseqfveq` you have an even harder problem: even if the values are not sets, the value of `frec(F, A)(3)` is fully determined by the values of `A, F(A), F(F(A)), F(F(F(A)))`. This is tricky since we don't even know that we are evaluating the second stage from the results of the first stage (since `F(A)` is not necessarily a set), so it might evaluate `F` somewhere else which is bad. Most likely this would require breaking down the function value operation and considering everything relationally, and it's not clear to me whether this actually holds.

I suspect you are seeing farther ahead than me but I suppose for the moment I'll see what I can do with `frec` (even if I prove a few results which might prove not to be any more useful than what we have now), and later ponder what they might mean for `seq`.

PR #2554, opened 2022-03-22
Remove dv conditions on unused variables (iset.mm and set.mm)

jamesjer:
DV conditions referring to variables that do not appear in the theorem or proof have crept into iset.mm and set.mm again.  This PR removes them all.  Some of the changes are in close proximity to changes made by https://github.com/metamath/set.mm/pull/2553, but I think they shouldn't conflict.  If they do, I am happy to fix up the conflicts in this PR.


tirix (approved):

tirix:
Ideally, I think we shall find a way to run this script (together with rewrap, and the one removing the unnecessary brackets) automatically after each PR.

We could even write a new script detecting which new theorems were introduced, and running a "minimize with" using those.

avekens (approved):
Great work again!

avekens:
> Ideally, I think we shall find a way to run this script (together with rewrap, and the one removing the unnecessary brackets) automatically after each PR.

This is a very good idea. But maybe it would be better to have a tool to be used before a PR is created (like rewrap).
 
> We could even write a new script detecting which new theorems were introduced, and running a "minimize with" using those.

This could take a while, depending on the proofs (numer and size) to be minimized.

benjub:
> We could even write a new script detecting which new theorems were introduced, and running a "minimize with" using those.

More important would be to `MM-PA> minimize_with *` the new or modified proofs of the PR. As @avekens notices, a global minimization using the new theorems would take a while.

benjub (approved):

jkingdon:
> Ideally, I think we shall find a way to run this script (together with rewrap, and the one removing the unnecessary brackets) automatically after each PR.

Yeah, having it built in would be ideal.

If I remember I'll run it before submitting a pull request. I tend to introduce these especially when I'm working on a proof with a lot of variables (due to some aspects of my habits and some aspects of how mmj2 works).

PR #2557, opened 2022-03-26
Remove unnecessary braces from iset.mm and set.mm (part 8).

jamesjer:
This is the last unnecessary brace removal patch.  After this, there are none left in any mm file.  With the exception of `tfr1onlembacc` in iset.mm, all of these braces are in mathboxes.  I'm still a little nervous about touching other people's mathboxes, to be honest, so I don't know if this should be merged or not.  Nevertheless, here it is for the community's consideration.


avekens (approved):
The changes in my mathbox are OK. I n general, I think it should be no problem to modify mathboxes as long as these are commonly agreed, as in this case.

jkingdon (approved):

jkingdon:
> I'm still a little nervous about touching other people's mathboxes, to be honest

In case it makes you feel any better, in https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md we write "Mathboxes should normally only be changed by the owner of the mathbox (unless it's a general improvement that applies across the database)." and this is exactly the sort of thing we had in mind by "general improvement that applies across the database".

PR #2562, opened 2022-03-30
Reduce cbvex4v dv conditions

jamesjer:
The cbvex4v theorem currently has an asymmetry.  Looking at pairs of corresponding variables across the bijection, we see:
- the first pair, x and v, have to be distinct
- the second pair, y and u, have to be distinct
- the third pair, z and f, do _not_ have to be distinct
- the fourth pair, w and g, do _not_ have to be distinct

The proof does not require this asymmetry.  Corresponding pairs of variables do not have to be distinct, similar to cbvex2v.  This change does not allow simplifying any current users of cbvex4v in iset.mm or set.mm, but it does slightly simplify the dv conditions for a proof I am working on.


benjub:
Looks like the line
```
$d g z $.  $d u v w z $.  $d u w x z $.  $d v w y z $.  $d w x y z $.
```
can be simplified to
```
$d g z $.  $d v u z w $.  $d x y z w $.  $d u x $.  $d v y $.
```
(given the preceding line of dv conditions).

Also, the variables u and v could be swapped in the statement (to respect alphabetical order), but this is annoying to do (you should prove cbvex4vNEW with the new variables, then `minimize_with cbvex4vNEW/may_grow` on theorems using cbvex4v, then delete cbvex4v, and finally rename cbvex4vNEW to cbvex4v). Maybe not worth it. Up to you.

[Above, you meant "biconditional" instead of "bijection".]

tirix (approved):

david-a-wheeler (approved):

avekens (approved):
If I see things correctly, x and v as well as y and u need not be distinct?

benjub:
Then I guess the proposed simplification is for a future PR ? (ping @jamesjer for when you have time)

PR #2563, opened 2022-04-01
shorten annotanannot

wlammen:
... and remove some outdated OLD theorems

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -340,8 +340,6 @@
 "4syl" is used by "acunirnmpt2".
 "4syl" is used by "acunirnmpt2f".
 "4syl" is used by "aevlem".
-"4syl" is used by "aevlemALTOLD".
-"4syl" is used by "aevlemOLD".
 "4syl" is used by "afv0nbfvbi".
 "4syl" is used by "aomclem6".
 "4syl" is used by "axdc3lem2".
@@ -1513,10 +1511,7 @@
 "axacndlem4" is used by "axacndlem5".
 "axacndlem5" is used by "axacnd".
 "axaddf" is used by "axaddcl".
-"axc11nlemALT" is used by "aevlemALTOLD".
 "axc11nlemALT" is used by "axc11nALT".
-"axc11nlemOLD2" is used by "aevlemOLD".
-"axc11nlemOLD2" is used by "axc11nOLDOLD".
 "axc11rvOLD" is used by "axc16gOLD".
 "axc16ALT" is used by "axc16gALT".
 "axc16g-o" is used by "ax12inda2".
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -31823,10 +31802,9 @@ requires either a dv condition ( ~ sb56 ) or a non-freeness hypothesis
 
   ${
     $d w x $.  $d w y $.
-    $( Alternate version of ~ axc11nlemOLD2 used in an older proof.
-       (Contributed by NM, 8-Jul-2016.)  (Proof shortened by Wolf Lammen,
-       17-Feb-2018.)  (Proof modification is discouraged.)
-       (New usage is discouraged.) $)
+    $( Used in an older proof.  (Contributed by NM, 8-Jul-2016.)  (Proof
+       shortened by Wolf Lammen, 17-Feb-2018.)
+       (Proof modification is discouraged.)  (New usage is discouraged.) $)
```
tirix: So you've removed `~aevlemALTOLD` which was the only proof using `~axc11nlemALT`. 
Do we even need to keep it?

tirix (approved):

set.mm
wlammen: See discouraged file:
New usage of "axc11nlemALT" is discouraged (1 uses).

tirix: Ok, so one usage remains, thanks!

set.mm
```diff
@@ -31823,10 +31802,9 @@ requires either a dv condition ( ~ sb56 ) or a non-freeness hypothesis
 
   ${
     $d w x $.  $d w y $.
-    $( Alternate version of ~ axc11nlemOLD2 used in an older proof.
-       (Contributed by NM, 8-Jul-2016.)  (Proof shortened by Wolf Lammen,
-       17-Feb-2018.)  (Proof modification is discouraged.)
-       (New usage is discouraged.) $)
+    $( Used in an older proof.  (Contributed by NM, 8-Jul-2016.)  (Proof
```
benjub, line 31805: I propose:
```
$( Lemma for ~ axc11nALT .  (Contributed ...
```

wlammen: axc11nlemALT is a weak replacement for aev in the proof of axc11n.  aev is provable from surprisingly few axioms.  Since axc11nlemALT is a specialization of aev, it is in principle not based on more axioms.  But its proof uses more, which is justified only, if it is structurally based on theorems you likely accept as more primitive, or that are more versatile, if you modify the axioms system slightly.  Whatever it is, it should somehow be expressed in axc11nALT, otherwise tirix question (shall we get rid of this?) will silently or openly crop up again.  That said, I will modify the text of axc11nlemALT as suggested on the next occasion.

benjub: @wlammen @tirix : Indeed, after looking at it, I think both axc11nlemALT and axc11nALT can be removed (the proof of axc11nALT is basically the same as that of axc11n once the use of axc11nlemALT is replaced by aev).

wlammen: axc11nALT is not exactly the topic of this PR.  I routinely checked the file for OLD theorems, and, so to say, by accident stumbled over axc11nlemALT, because a theorem mentioned in the introducing text had disappeared.  I would like to leave it here as is and suggest to open a different pull request, where the removal of axc11n*ALT is addressed.  I think that would be the right place to discuss this in greater detail.

benjub: Ok. Current PR can be merged now.

avekens (approved):

PR #2566, opened 2022-04-05
Fix the distinct variable requirements for `~df-liminf`

tirix:
Just before the definition of `~df-liminf`, a distinct variable definition `$d x k $.` was at the top scope.
This impacted the rest of `set.mm`, i.e. all the following mathboxes. 
This adds a scope to limit the distinct variable definition.

@glacode Warning, this is in your mathbox. We normally refrain from modifying other people's mathboxes, but this seemed to me to obviously be a mistake.


digama0:
We should have a CI check for this, although it's not clear exactly where it should live since it is set.mm specific.

tirix:
We could certainly have a check which we would only run on `set.mm` and `iset.mm`.
(just added a second case, with `~df-xlim`. This seems to be the last.)

tirix:
FYI for those curious, I found this while playing with some annotation which adds DV indications to metamath files in VS Code, it looks like this:
image
The `(k,x)` annotation popped up way too often...

I wanted to try this out following [a question from Andrew Lubrino on the mailing list](https://groups.google.com/g/metamath/c/kc70ViZIAlE/m/AApb80mpAAAJ). This version of `ax-4` would answer his question about free variables:
image

I'll probably keep this feature, but leave it off by default as it might bring too much/misleading information, but it could be useful and interesting for some.

avekens (approved):

wlammen:
The correct solution is to limit effects to each mathbox, so mistakes by others do not affect your own ideas.

benjub:
How would we check this in CI ? Check for overly long $d-scopes ? Or only check that definitions have their own $d's ? (but the mistake could also occur for $p-statements)

In my own contributions, I have always been weary of shared $d-statements, and often had $d-statements specific to each $a or $p-statement, even if this meant a slightly longer file. In many other cases, however, they are fully justified for series of theorems in the same family (e.g., theorems sharing the same dummy variables, or the same quantified setvar variables when most other variables are class variables).

digama0:
No $d or $e at global scope. That's not strictly an error in the metamath spec (and I recall one case where Norm used it to demonstrate the effect of "unbundling" all of set.mm, so it's not totally useless). but it's definitely a global decision regarding set.mm structure and I think we don't want it.

benjub:
> No $d or $e at global scope. That's not strictly an error in the metamath spec (and I recall one case where Norm used it to demonstrate the effect of "unbundling" all of set.mm, so it's not totally useless). but it's definitely a global decision regarding set.mm structure and I think we don't want it.

By "we don't want it", you mean "we dont want $d or $e at global scope", correct ?  If you think the CI test would be beneficial, then so be it.  Nevertheless, in general, I would avoid rules refering to the "global scope", since this does not play well with file inclusions (something in the global scope of the included file would not be at global scope if its inclusion command is not, as currently allowed by metamath.c - v0.198), and also because I think that if `<db>` is a valid metamath database, then so should `${ <db> $}` be.

digama0:
There is some stuff in smetamath / metamath-knife about controlling database inclusion statements. I thought we got it added to the spec that file inclusion statements can only happen at global scope, which would solve that issue. (EDIT: [Yes](https://github.com/david-a-wheeler/metamath-knife/pull/84), mm-knife rejects inclusion statements not at global scope.) IIRC metamath.exe is doing something fairly low level for file inclusion statements, kind of C preprocessor style, which means that it allows file inclusion statements in weird places like between tokens and such.

There are already similar rules about global scope: $c statements can only appear at global scope. File inclusions wouldn't really affect this, both because we're talking about set.mm which doesn't use file inclusion, and also because I'm envisioning this as a check inside a metamath-aware verifier, which would therefore follow file inclusions correctly and know exactly what scope things are at.

benjub:
@digama0 : indeed, there is a discrepancy between the metamath book and metamath.c - v0.198 that I signaled in https://github.com/metamath/metamath-book/issues/232

If you think additional checks are beneficial in the case of set.mm and iset.mm, then so be it. But I reiterate that as for the metamath language rules, I think there should be no mentions at all of "outer scope", which, as I currently see things, uselessly complexify matters. As for $c-statements, there should be no restriction (when a $c-statement is encountered, then the constants are added to a global variable, regardless of where the $c-statement is). Even simpler would be to remove/ignore all $c-statements, and simply add a constant to the global variable when a new symbol is encountered which has not been defined as a variable. 

Similarly, all $v-statements could be removed in favor of fully local $f-statements, as you proposed in https://github.com/metamath/metamath-exe/issues/5. I had forgotten about that, but I recently arrived at the same conclusion after modifying mmverify.py (see https://github.com/david-a-wheeler/mmverify.py/pull/7) and writing another parser/verifier in another language.

glacode:
> Just before the definition of `~df-liminf`, a distinct variable definition `$d x k $.` was at the top scope. This impacted the rest of `set.mm`, i.e. all the following mathboxes. This adds a scope to limit the distinct variable definition.
> 
> @glacode Warning, this is in your mathbox. We normally refrain from modifying other people's mathboxes, but this seemed to me to obviously be a mistake.

Thank you, @tirix , for fixing this.

Were you looking for something "unexpected" in the outermost scope?

set.mm
```diff
@@ -670161,10 +670161,12 @@ is real and it is less than or equal to its inferior limit (in such a
...
-  $d x k $.
-  $( Define the convergence relation for extended real sequences.  (Contributed
-     by Glauco Siliprandi, 5-Feb-2022.) $)
-  df-xlim $a |- ~~>* = ( ~~>t ` ( ordTop ` <_ ) ) $.
+  ${
+    $d x k $.
+    $( Define the convergence relation for extended real sequences.
+       (Contributed by Glauco Siliprandi, 5-Feb-2022.) $)
+    df-xlim $a |- ~~>* = ( ~~>t ` ( ordTop ` <_ ) ) $.
+  $}
```
glacode: There are only constants in this definition, @tirix what about simply removing

 $d x k $.

?  (and no extra scope needed)

Glauco

tirix: You're totally right, thank you! I fixed that.

tirix:
> Were you looking for something "unexpected" in the outermost scope?

Actually, like I wrote above, I was following up on [a question from Andrew Lubrino on the mailing list](https://groups.google.com/g/metamath/c/kc70ViZIAlE/m/AApb80mpAAAJ). He was expecting to see something like `A. x ph`**`(x)`** to emphasise that x is free in ph, so I updated the VS Code extension to explicitly show free variables, as in:
image
(You can see the editor has added the `(x)` after _`ph`_ and _`ps`_)

When I first tried that feature out, `x` and `k` appeared as bound variables in expressions where they shall not have, so I went on the search where they were coming from and found this.

jkingdon (approved):

PR #2567, opened 2022-04-05
shorten ad*antr

wlammen:

PR #2568, opened 2022-04-06
shorten ad*antr

wlammen:


wlammen:
This is a copy of PR #2567, closed by accident, approved by @tirix

tirix:
@wlammen I've reopened #2567 , I suggest we merge the original one (with description and - even if so short - discussion)

PR #2570, opened 2022-04-07
more recursion closure theorems

jkingdon:
There probably isn't a really clear place to break this into separate pull requests, but here is the latest round of removing `S e. _V` conditions from various theorems, namely `frecuzrdgrom`, `frecuzrdglem`, `frecuzrdgfn`, `iseqcl`, and `frecuzrdgcl`.

Also add `relsng` and `relsnopg`, stated as in set.mm. This is not because I see an immediate need for them, just that I noticed that they weren't in iset.mm and easily could be.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -140,6 +140,10 @@
 "frecsucOLD" is used by "frec2uzsucd".
 "frecsucOLD" is used by "frecclOLD".
 "frecsucOLD" is used by "frecuzrdgsuc".
+"frecuzrdgfn" is used by "frecuzrdg0".
+"frecuzrdgfn" is used by "frecuzrdgcl".
+"frecuzrdgfn" is used by "frecuzrdgsuc".
+"frecuzrdgfn" is used by "iseqfn".
 "frecuzrdglemOLD" is used by "frecuzrdgcl".
 "frecuzrdglemOLD" is used by "frecuzrdgfn".
 "frecuzrdglemOLD" is used by "frecuzrdgsuc".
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -95522,18 +95595,32 @@ the previous value and the value of the input sequence (second operand).
...
+  ${
+    $d .+ w x y z $.  $d F w x y z $.  $d M w x y z $.  $d N x y $.
+    $d S w x y z $.  $d ph w x y z $.
+    iseqclOLD.1 $e |- ( ph -> N e. ( ZZ>= ` M ) ) $.
+    iseqclOLD.ex $e |- ( ph -> S e. V ) $.
+    iseqclOLD.2 $e |- ( ( ph /\ x e. ( ZZ>= ` M ) ) -> ( F ` x ) e. S ) $.
+    iseqclOLD.3 $e |- ( ( ph /\ ( x e. S /\ y e. S ) ) -> ( x .+ y ) e. S ) $.
+    $( Closure properties of the recursive sequence builder.  New proofs should
+       use ~ iseqcl instead (which is the same with one fewer hypotheses).
+       (Contributed by Jim Kingdon, 1-Jun-2020.)  (New usage is discouraged.)
```
benjub, line 95620: property (?)
hypothesis

benjub (approved):

iset.mm
jkingdon: I fixed this (two places) and made a few other small comment edits. Will merge as soon as the tests pass.

PR #2573, opened 2022-04-09
shorten 3com23

wlammen:
It is more efficient to prove the pair 3comr and 3com23 from 3comr than to base it on 3com23, as was done before.  The proof of 3comr changes, but is neither shorter nor longer, but 3com23 becomes 5 proof bytes/1 proof line shorter.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13452,6 +13452,9 @@ New usage of "3anidm12p1" is discouraged (0 uses).
 New usage of "3anidm12p2" is discouraged (0 uses).
 New usage of "3anorOLD" is discouraged (0 uses).
 New usage of "3atnelvolN" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


jkingdon (approved):

avekens (approved):
Hm, the old proof of 3comr has the charm to use only one other theorem 3com1 (it is used twice). But for the sake of uniformity, the new proof can be used.

wlammen:
@avekens If you look at commit [ef39944](https://github.com/metamath/set.mm/pull/2573/commits/ef39944bc53043366186ae5a3fd54f860ad17146) you will find that 3coml was reproved using only 3comr. In the end I reverted this detail for the sake of minimizing changes and not overwriting contributions of others without a necessity.

PR #2574, opened 2022-04-10
Remove S e. V hypothesis in iseq1 and iseqp1 in iset.mm

jkingdon:
Showing this can be done for `iseq1` and `iseqp1`  pretty much convinces me that this can be done pretty much everywhere which depends on the theorems in this pull request and nearby ones. I don't know whether I'll do them all myself unless I have a more concrete need than I have at the moment.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -134,21 +134,22 @@
 "eu3h" is used by "mo2r".
 "frec2uzrdgOLD" is used by "frecuzrdgfn".
 "frec2uzrdgOLD" is used by "frecuzrdglemOLD".
-"frec2uzrdgOLD" is used by "frecuzrdgsuc".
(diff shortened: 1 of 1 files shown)


tirix:
So this is work in progress, and you'll now replace uses of `iseq1OLD` by `iseq1` everywhere it occurs, correct?

jkingdon:
> So this is work in progress, and you'll now replace uses of `iseq1OLD` by `iseq1` everywhere it occurs, correct?

Well that's the idea but given that it isn't just this theorem but dozens like it, I'm not sure I'd predict an especially short value of "now". 

Help is certainly welcome if someone wants to speed up this process. My next project is probably #2575 which I'm not sure needs the iseq1OLD elimination stuff.

tirix (approved):

PR #2576, opened 2022-04-11
AV's mathbox: class of inhabited sets/well-defined functions

avekens:
In 2017, shortly after I started with Metamath, I was surprised that there is no direct definition of undefined function values, but only an implicit indication (the function value for arguments outside the domain of the function is the empty set, which can be a valid value also for arguments of the domain). Therefore, I started a study to define undefined function values as being the universe and to look at the implications such a definition would have, see section "Alternative definitions of function and operation values" in my mathbox. 

In the meantime, I appreciate that the function value is guaranteed to be a set (see ~fvex), so my former approach would not be practical. Therefore, I looked for a different approach, and I came up to define the class of inhabited sets (or restricted universal class, because this class is the universe without the empty set), which play a special role in intuitionistic logic (see definition and related theorems in iset.mm). By claiming (by definition) that the range of a function must be an inhabited set, it can be decided immediately if a function value is defined (if it is not the empty set) or if it is not defined (if it is the empty set).  

This PR provides the definitions and some basic theorems. Some of them are already interesting. I plan, however, to look for more advanced examples where my new approach would provide some benefits.

Again, everything is in my mathbox (and may remain there forever, as will be the case with section "Alternative definitions of function and operation values").

Details of changes:

main set.mm:
* some theorems moved from other mathboxes

AV's mathbox:
* definition and related theorems for the restricted universal class/class of inhabited sets
* definition and related theorems for well-defined functions


set.mm
```diff
@@ -716530,6 +716542,1358 @@ contain the empty set ( ` -. (/) e. S ` ), no additional assumption is
...
+      DUBZAIZVABIZVACIZUTVBFZVCJVEVDVCUNVDVCJUMUSVBCBVAKUCLVDUDZUTVBVCVFVBUTVCV
+      BVFUTVCJZVBVFFVAUQIZVGVAACMVHUSVCUOUSVHVCUSVHVAURIZVCUQURVAUEZVABCNUFLROP
+      QSTVDUTVCFZVBJVKVDVBUMVDVBJUNUSVCCAVAKUGLVFUTVCVBVFVCUTVBVCVFUTVBJZVCVFFV
+      IVLVABCMVIUSVBUOUSVIVBUSVIVHVBVJVAACNUHLROPQSTUIUJUKUL $.
+  $}
+
+  $( The difference between a class and the universe is empty.  (Contributed by
+     AV, 30-Mar-2022.) $)
+  difv0 $p |- ( A \ _V ) = (/) $=
+    ( cvv wss cdif c0 wceq ssv ssdif0 mpbi ) ABCABDEFAGABHI $.
```
tirix, line 716584: This looks like it's general and interesting enough to be moved to the main part  -  let's see.

tirix (approved):

tirix:
So examples of functions, which would _not_ be well-defined functions, would be `(,)` (real open range), other range functions (half-open and closed), `...` (integer closed range), as these functions all result in the empty set if the second argument is less than the first.
Same for ``( int ` J )`` and ``( cls ` J )``, the functions mapping sets to their interior, resp. closure relative to a topology `J`, which have empty values for singletons, resp. the empty set itself.

avekens:
> So examples of functions, which would _not_ be well-defined functions, would be `(,)` (real open range), other range functions (half-open and closed), `...` (integer closed range), as these functions all result in the empty set if the second argument is less than the first. 

That's a good point, I'll investigate this further. But in principle you should be right:  ( M ... N ) should not be defined for N < M, and actually ( M ... N ) = (/) according to ~fzn in this case. 


> Same for `` ( int ` J ) `` and `` ( cls ` J ) ``, the functions mapping sets to their interior, resp. closure relative to a topology `J`, which have empty values for singletons, resp. the empty set itself.

I must have a closer look at these examples, if it is justified to call these functions "undefined" for the given cases.

tirix:
Other examples:
- In metric spaces, the `ball` function, which maps real numbers to open balls of that radius, is the empty set for radius 0,
- In graph theory, I think it's perfectly valid to have a graph without edges. So ``( iEdg ` G ) = (/)`` does not mean that `G` is outside the domain of the function / not a graph.

I think my point is not that your definition/proposal is not correct, but rather that it will not apply to this class of functions/operations, where the empty set happens to be a valid evaluation for some value which is actually in the domain.

jkingdon:
The bit about "functions which never return the empty set are special" is present in some parts of set.mm and I think maybe iset.mm. http://us.metamath.org/ileuni/relelfvdm.html and http://us.metamath.org/mpeuni/elfvdm.html spring to mind. But there are also things like http://us.metamath.org/mpeuni/ndmovord.html and probably others like it.

As for this pull request, seems perfectly plausible for a mathbox, as a way to explore the topic.

avekens:
@tirix @jkingdon thank you for your remarks. They showed that my approach is too restrictive, because it does not work (at least not with much overhead which would nullify its advantages) for functions with values which are intentionally sets. And because of our convention to define special kinds or families of sets by mappings, for example the set of edges, a large number of "functions" would not be well defined. And in these cases, it would still not be possible to distinguish if ` ( F `` X ) = (/) ` means "the (indexed) set is empty" or "` F ` is not defined at ` X `".

Therefore, I retract my PR.

tirix:
Just to be clear, my point was not that you should retract the PR (I've actually approved it!), just that this could not completely replace the existing definition and be universally used. Of course it's your choice to do so if you so wish, though!

avekens:
> Just to be clear, my point was not that you should retract the PR (I've actually approved it!), just that this could not completely replace the existing definition and be universally used. Of course it's your choice to do so if you so wish, though!

No problem, I was happy that your comments helped me to recognize that my approach is too restrictive, and I see no meaning/usage of my definitions, at least in the current form. I'll keep the changes in my feature branch...

PR #2580, opened 2022-04-14
Remove iseqclOLD , iseq1OLD , and iseqp1OLD from iset.mm

jkingdon:
So at #2574 I said I might not get to this for a while. But at least part of it I got inspired about (as suggested there, there is plenty left even after this). The fact that my progress on iseqf1o is going very slowly has nothing to do with it :grin: .

In addition to iseqclOLD , iseq1OLD , iseqp1OLD and their usages, there are a few fairly similar things:

* remove a hypothesis from http://us.metamath.org/ileuni/rdgon.html by using http://us.metamath.org/ileuni/tfrcl.html
* add iseqfcl which replaces iseqf and iseqfn

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -188,7 +188,6 @@
 "iseqp1OLD" is used by "facp1".
 "iseqp1OLD" is used by "ialgrp1".
 "iseqp1OLD" is used by "iseqhomo".
-"iseqp1OLD" is used by "iseqid2".
 "iseqp1OLD" is used by "iseqz".
 "iseqp1OLD" is used by "resqrexlemfp1".
 "mo3h" is used by "mo2dc".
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -95653,14 +95654,12 @@ the previous value and the value of the input sequence (second operand).
...
+       instead.  (Contributed by Jim Kingdon, 23-Jul-2021.)
+       (New usage is discouraged.) $)
     iseqf $p |- ( ph -> seq M ( .+ , F , S ) : Z --> S ) $=
-      ( vk cfv wf cv wcel wa cuz cseq wral eleq2i sylan2br iseqfn simpr adantlr
-      wfn adantr co iseqclOLD ralrimiva ffnfv sylanbrc feq2i sylibr ) AGUAPZEDE
-      FGUBZQZIEUSQAUSURUIORZUSPESZOURUCUTABCDEFGHLKBRZURSZAVCISVCFPESZIURVCJUDM
-      UEZNUFAVBOURAVAURSZTBCDEFGVAHAVGUGAEHSVGKUJAVDVEVGVFUHAVCESCRZESTVCVHDUKE
-      SVGNUHULUMOUREUSUNUOIUREUSJUPUQ $.
+      ( iseqfcl ) ABCDEFGIJLMNO $.
   $}
```
tirix, line 95663: Why not completely remove this theorem in favor of ~iseqfcl, since it is only an alias now?

tirix (approved):

iset.mm
jkingdon: > Why not completely remove this theorem in favor of ~iseqfcl, since it is only an alias now?

We'll get there. But `iseqf` has a `𝑆 ∈ 𝑉` hypothesis missing in `iseqfcl`, so all the usages will need to be updated for that.

tirix: Oh, sure, I thought this already covered all usages.

avekens (approved):

PR #2583, opened 2022-04-16
Remove dependencies on ax-6

CatsAreFluffy:
`19.23v`, `19.36iv`, `19.37iv`, and `19.41v` had unnecessary dependencies on `ax-6`, so I removed those dependencies and moved theorems in the `ax-6` section that no longer depended on `ax-6` to the `ax-5` section. This lengthens the proofs of `19.36iv` and `19.37iv`, since the closed forms of those theorems are biconditionals whose other directions require ax-6, but I don't think that issue applies to the other theorems and their `ax-6`-free consequences.


benjub:
Congratulations @CatsAreFluffy !
We sort the list of contributors at the beginning of set.mm by alphabetical order of the last name.
Also, it might be worth adding the closed forms of 19.36iv and 19.37iv as well.

tirix:
Welcome @CatsAreFluffy !
Our syntax checker is rather picky, you'll have to insert a space after `~ ax-6` in the comments for `~ 19.13iv`,  `~ 19.36iv`, and `~ 19.37iv`.

tirix (approved):

CatsAreFluffy:
I made the suggested changes. Are `19.36imv`/`19.37imv` good names for the closed forms?

jkingdon:
I clicked the "Approve and run" button for "1 workflow awaiting approval. First-time contributors need a maintainer to approve running workflows." This just means we are willing to run GitHub Actions, which will run the usual checks, for pull requests from @CatsAreFluffy 

The checks passed (as seen where it says "All checks have passed" on this page), so that looks good.

tirix:
> Are `19.36imv`/`19.37imv` good names for the closed forms?

The convention is that closed forms do not have the "i", "g" or "d" suffix, so the closed form of `19.36iv` would simply be `19.36v`. It already exists, so you can also supersede the current version by changing it into "OLD", just like you did for `19.23v`.

(See "[Conventions](http://us2.metamath.org:88/mpeuni/conventions.html)", last section of "Theorem forms".)
@benjub is that what you meant?

CatsAreFluffy:
`19.36imv` and `19.37imv` are only the forward directions of `19.36v`/`19.37v`, and those theorems are used in situations where both directions of the biconditionals are necessary. I got the `im` suffix from `excomim`; another convention that seems to be used is a `-1` suffix, so I could change them to use that if you think that's a good idea.

jkingdon:
> `19.36imv` and `19.37imv` are only the forward directions of `19.36v`/`19.37v`, and those theorems are used in situations where both directions of the biconditionals are necessary. I got the `im` suffix from `excomim`; another convention that seems to be used is a `-1` suffix, so I could change them to use that if you think that's a good idea.

I'd go for `im` as that is a convention for "the forward direction of a biconditional" (mostly concretely codified at http://us.metamath.org/ileuni/conventions.html although it is found in set.mm too as you noticed).

`-1` doesn't mean anything specific, it is just a way of saying "the name I wanted was taken, this is the next available name".

CatsAreFluffy:
I'll keep the names as they are then.

benjub (approved):

tirix:
Excellent, then I'm merging this PR.
Thanks a lot, @CatsAreFluffy !

Issue #2575, opened 2022-04-10
Intuitionize seqf1o


jkingdon:
That is, set.mm has http://us.metamath.org/mpeuni/seqf1o.html and iset.mm does not have anything particularly close.

There's some discussion of potentially related theorems at https://github.com/metamath/set.mm/pull/2574 and https://github.com/metamath/set.mm/pull/2551 (disclaimer: although the topics mentioned there are interesting ones, some of them might be red herrings with respect to `seqf1o` - so don't take that discussion as necessarily a blocker).

It is likely that some of the occurences in seq1fo of `( M ... N )` can be changed to ``( ZZ>= ` M )`` without making life more difficult for http://us.metamath.org/mpeuni/summolem3.html (this is the usage of `seqf1o` which we're concerned about currently)

jkingdon:
I'm not sure I see all the pieces, but my current thinking is that we'll want a version of theorems such as http://us.metamath.org/ileuni/iseqfveq.html which change `𝑥 ∈ (ℤ≥‘𝑀)` in the closure hypothesis to `x e. M ... N` (likewise for `x e. M ... ( N + 1 )` in theorems such as http://us.metamath.org/ileuni/iseqp1.html ).

Most of the reason that existing theorems need to look at the values of `F` "too far out" (that is, beyond values which should matter) is that they they use closure hypotheses when they are proving that something like `seq` ends up with a function. So if we can show we end up with a function anyway, we should be good. I'm thinking of something along the lines of http://us.metamath.org/ileuni/frecuzrdgtcl.html but with a weaker conclusion (`Fun T` or something similar) and which would omit the `((𝜑 ∧ (𝑥 ∈ (ℤ≥‘𝐶) ∧ 𝑦 ∈ 𝑆)) → (𝑥𝐹𝑦) ∈ 𝑆)` hypothesis (or replace it with something weaker).

[...]

PR #2584, opened 2022-04-17
iset.mm: a theorem about functions which are not defined everywhere, plus remove more sethood hypotheses in seq related theorems

jkingdon:
The part about this which is new is proving `Fun ( { <. A , B >. } i^i ( _V X. _V ) )` which by itself might not seem interesting, but which is intended to be combined with things like http://us.metamath.org/ileuni/funun.html , http://us.metamath.org/ileuni/brinxp.html , and http://us.metamath.org/ileuni/funbrfv2b.html to make some progress on #2575 (and in particular being able to relax ``x e. ( ZZ>= ` M )`` to `x e. ( M ... ( N + 1 ) )` in contexts like the `iseqp1.pl` hypothesis of http://us.metamath.org/ileuni/iseqp1.html

The part which is familiar is getting rid of more of the `S e. V` conditions in `seq` related theorems, as seen in #2580 and earlier pull requests.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -132,17 +132,13 @@
 "eu3h" is used by "2eu4".
 "eu3h" is used by "eu3".
 "eu3h" is used by "mo2r".
-"frec2uzrdgOLD" is used by "frecuzrdgfn".
 "frec2uzrdgOLD" is used by "frecuzrdglemOLD".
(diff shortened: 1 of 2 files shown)


jkingdon:
Closing in favor of #2588

PR #2587, opened 2022-04-26
Update the "other databases" list with an OpenAI collection

tirix:
It seems actually just be a collection a statements to train AI, and requires `set.mm`, so it is not a database per se, but I thought it might be worth mentioning.


jkingdon (approved):
Seems worth a link for sure.

avekens (approved):

PR #2588, opened 2022-05-01
Remove old seq/frec related theorems in iset.mm and pave the way for removing third argument to seq

jkingdon:
1. Recent pull requests have discouraged some theorems (in favor of replacements with fewer constraints), and this pull request removes some of the old versions.

2. The pull request includes a variety of theorems which lead up to a version of http://us.metamath.org/ileuni/iseqss.html which removes the `𝑇 ∈ 𝑉` and `(𝑥 ∈ 𝑇 ∧ 𝑦 ∈ 𝑇) → (𝑥 + 𝑦) ∈ 𝑇` hypotheses. Since this allows `T` to be `_V` it means that we'll be able to (eventually) make iset.mm more similar to set.mm, by removing the third argument to `seq` (and even before we do that, the `iseqss` replacement should make it easier to work with `seq`).

Because #2584 is not merged yet and this pull request depends on changes made there, this pull request also contains all the changes from that pull request. If #2584 is merged first, github should update automatically (and if not I'll be glad to update as needed).

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -132,7 +132,6 @@
 "eu3h" is used by "2eu4".
 "eu3h" is used by "eu3".
 "eu3h" is used by "mo2r".
-"frecsucOLD" is used by "frec2uzsucd".
 "hbs1" is used by "eu1".
 "hbs1" is used by "hbab1".
 "hbs1" is used by "mopick".
@@ -268,7 +267,7 @@ New usage of "elirr" is discouraged (16 uses).
 New usage of "equsalh" is discouraged (5 uses).
 New usage of "eu3h" is discouraged (3 uses).
 New usage of "fnexALT" is discouraged (0 uses).
-New usage of "frecsucOLD" is discouraged (1 uses).
+New usage of "frecsucOLD" is discouraged (0 uses).
 New usage of "hbs1" is discouraged (5 uses).
 New usage of "idALT" is discouraged (0 uses).
 New usage of "mathbox" is discouraged (0 uses).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -132,17 +132,13 @@
 "eu3h" is used by "2eu4".
 "eu3h" is used by "eu3".
 "eu3h" is used by "mo2r".
-"frec2uzrdgOLD" is used by "frecuzrdgfn".
 "frec2uzrdgOLD" is used by "frecuzrdglemOLD".
 "frecclOLD" is used by "frecuzrdgrrnOLD".
 "frecsucOLD" is used by "frec2uzrdgOLD".
 "frecsucOLD" is used by "frec2uzsucd".
 "frecsucOLD" is used by "frecclOLD".
-"frecuzrdglemOLD" is used by "frecuzrdgfn".
-"frecuzrdgrom" is used by "frecuzrdgfn".
 "frecuzrdgrom" is used by "frecuzrdglemOLD".
 "frecuzrdgrrnOLD" is used by "frec2uzrdgOLD".
-"frecuzrdgrrnOLD" is used by "frecuzrdgfn".
 "hbs1" is used by "eu1".
 "hbs1" is used by "hbab1".
 "hbs1" is used by "mopick".
@@ -282,13 +278,12 @@ New usage of "elirr" is discouraged (16 uses).
 New usage of "equsalh" is discouraged (5 uses).
 New usage of "eu3h" is discouraged (3 uses).
 New usage of "fnexALT" is discouraged (0 uses).
-New usage of "frec2uzrdgOLD" is discouraged (2 uses).
+New usage of "frec2uzrdgOLD" is discouraged (1 uses).
 New usage of "frecclOLD" is discouraged (1 uses).
 New usage of "frecsucOLD" is discouraged (3 uses).
-New usage of "frecuzrdgfn" is discouraged (0 uses).
-New usage of "frecuzrdglemOLD" is discouraged (1 uses).
-New usage of "frecuzrdgrom" is discouraged (2 uses).
-New usage of "frecuzrdgrrnOLD" is discouraged (2 uses).
+New usage of "frecuzrdglemOLD" is discouraged (0 uses).
+New usage of "frecuzrdgrom" is discouraged (1 uses).
+New usage of "frecuzrdgrrnOLD" is discouraged (1 uses).
 New usage of "hbs1" is discouraged (5 uses).
 New usage of "idALT" is discouraged (0 uses).
 New usage of "iseqf" is discouraged (4 uses).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -132,7 +132,6 @@
 "eu3h" is used by "2eu4".
 "eu3h" is used by "eu3".
 "eu3h" is used by "mo2r".
-"frecclOLD" is used by "frecuzrdgrrnOLD".
 "frecsucOLD" is used by "frec2uzsucd".
 "frecsucOLD" is used by "frecclOLD".
 "hbs1" is used by "eu1".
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -95229,6 +94988,217 @@ gives an ordered pair whose first element is the index (translated
...
+        UUJUVBYRLVRRUVNXNVMXOXTXPXQ $.
+    $}
+
+    ${
+      $d C v x y $.  $d C x y $.  $d F x y $.  $d R v $.  $d R x y $.
+      $d G x y $.  $d S x y $.  $d T x y $.  $d ph v x y $.  $d ph x y $.
+      frecuzrdgdom.2 $e |- G = frec ( ( x e. ZZ |-> ( x + 1 ) ) , C ) $.
+      $( The domain of the result of the recursive definition generator on
+         upper integers.  (Contributed by Jim Kingdon, 24-Apr-2022.) $)
+      frecuzrdgdom $p |- ( ph -> dom ran R = ( ZZ>= ` C ) ) $=
```
avekens, line 95077: This theorem does not depend on G!? Why is frecuzrdgdom.2 needed?

iset.mm
```diff
@@ -95229,6 +94988,217 @@ gives an ordered pair whose first element is the index (translated
...
+    $}
+
+    ${
+      $d C x y z $.  $d F x y z $.  $d G w x y z $.  $d R v w z $.
+      $d R w x y z $.  $d S x y z $.  $d T x y z $.  $d ph v w z $.
+      $d ph w x y z $.
+      frecuzrdgfun.g $e |- G = frec ( ( x e. ZZ |-> ( x + 1 ) ) , C ) $.
+      $( The recursive definition generator on upper integers produces a a
+         function.  (Contributed by Jim Kingdon, 24-Apr-2022.) $)
+      frecuzrdgfun $p |- ( ph -> Fun ran R ) $=
```
avekens, line 95099: This theorem does not depend on G!? Why is frecuzrdgfun.g needed?

iset.mm
```diff
@@ -95229,6 +94988,217 @@ gives an ordered pair whose first element is the index (translated
...
+      $d B x y z $.  $d C x y z $.  $d F x y z $.  $d G x y $.  $d R x y $.
+      $d S x y z $.  $d T x y z $.  $d ph x y z $.
+      frecuzrdgsuct.g $e |- G = frec ( ( x e. ZZ |-> ( x + 1 ) ) , C ) $.
+      frecuzrdgsuct.ran $e |- ( ph -> P = ran R ) $.
+      $( Successor value of a recursive definition generator on upper integers.
+         See comment in ~ frec2uz0d for the description of ` G ` as the mapping
+         from ` _om ` to ` ( ZZ>= `` C ) ` .  (Contributed by Jim Kingdon,
+         29-Apr-2022.) $)
+      frecuzrdgsuct $p |- ( ( ph /\ B e. ( ZZ>= ` C ) ) ->
+          ( P ` ( B + 1 ) ) = ( B F ( P ` B ) ) ) $=
```
avekens, line 95166: This theorem does not depend on G!? Why is frecuzrdgsuct.g needed?

avekens (changes requested):

avekens:
> Because #2584 is not merged yet and this pull request depends on changes made there, this pull request also contains all the changes from that pull request. If #2584 is merged first, github should update automatically (and if not I'll be glad to update as needed).

... This means if this PR #2588 is accepted and merged, PR #2584 is obsolete and can be deleted?

jkingdon:
> This means if this PR #2588 is accepted and merged, PR #2584 is obsolete and can be deleted?

That's right. github is fairly flexible about whether to do it that way or in a different order, but I have gone ahead and closed #2584 figuring that is the easiest way to proceed.

iset.mm
jkingdon: It is used in the proof. Although it would, of course, be possible to use `frec ( ( x e. ZZ |-> ( x + 1 ) ) , C )` throughout the proof, I think this sort of "local definition" usually makes for a more readable situation (especially in a case like this where something like a third or a half of the steps have a `G` in them).

Did you have another idea of how to handle this situation? I could imagine doing things with naming or comments or something, but perhaps there is a possibility I hadn't thought of.

iset.mm
jkingdon: As with the previous comment, a lot of the steps in this proof have a `G` in them, so this is to make the proof more readable.

iset.mm
jkingdon: Pretty much every step in the second half of this proof has a `G` in it, so it is for readability of the proof as described above.

iset.mm
benjub: It can be left as is, or be a lemma followed by the theorem, proved from that lemma and eqid, as in http://us2.metamath.org/mpeuni/bj-pwcfsdom.html
I would favor the latter, but there are many such "proof-facilitating hypotheses" in set.mm and apparently it is not cared much about leaving these hypotheses.

avekens: OK, I see. Are there really many such cases in set.mm? Either I did not use such theorems before, or I did not notice that. For myself, I often use local definitions as "proof-facilitating hypotheses", but always provide at least one theorem/lemma which contains this defintion explicitly. Therefore, I would prefer Benoît's suggestion.

jkingdon: There are certainly examples in this work of "lemma followed by the theorem" style, such as  http://us.metamath.org/ileuni/freccllem.html / http://us.metamath.org/ileuni/freccl.html or http://us.metamath.org/ileuni/frecsuclem.html / http://us.metamath.org/ileuni/frecsuc.html . Now basically `frecuzrdgdom` and `frecuzrdgfun` are lemmas for `frecuzrdgtclt` , and `frecuzrdgsuct` is a lemma for `iseqp1t` so I don't think this is fundamentally different. If they are lemmas should they have `lem` in the name? Well, maybe, but the names here are already getting long and I'm not sure I have especially brilliant ideas about how to name any of this. Some of the considerations are (1) set.mm has some of the same naming questions (for example, whether `uzrdg` is a good fragment of a name for the "pre-seq" theorems as now used in both set.mm and iset.mm), (2) there are several generations of these theorems, some of which may go away at some time and some of which might be worth keeping side by side (and no, I don't have super firm opinions about which are which), (3) there are more changes coming up mostly likely, at least if I (or someone else) takes this work further, (4) at least some of these "lemmas" are used in more than one (but less than, say, five) theorems which I guess argues for some kind of name not including `lem`.

jkingdon: > Therefore, I would prefer Benoît's suggestion.

Give me a day or so and I'll update this pull request to do that.

jkingdon: done

jkingdon:
I have revised the pull request to address the comments. Ready for re-review, @avekens and @benjub

avekens (approved):

iset.mm
```diff
@@ -44466,6 +44466,21 @@ We use their notation ("onto" under the arrow).  (Contributed by NM,
       ( cvv wcel cop csn wfun funsng mp2an ) AEFBEFABGHICDABEEJK $.
   $}
 
+  ${
+    $d A x y z $.  $d B x y z $.
+
+    $( A function based on the singleton of an ordered pair.  Unlike ~ funsng ,
+       this holds even if ` A ` or ` B ` is a proper class.  (Contributed by
+       Jim Kingdon, 17-Apr-2022.) $)
+    funinsn $p |- Fun ( { <. A , B >. } i^i ( _V X. _V ) ) $=
```
benjub, line 44475: It might be as easy to prove
```
funinsn $p |- Fun ( { <. A , B >. } i^i ( V X. W ) ) $=
```
?

benjub (approved):

benjub:
Removing that third argument of `seq` would be nice indeed !

iset.mm
jkingdon: Are you thinking of the `𝐴 ∈ 𝑉 ∧ 𝐵 ∈ 𝑊` case? Because then there is no need to intersect with anything: http://us.metamath.org/ileuni/funsng.html

jkingdon:
> Removing that third argument of `seq` would be nice indeed !

Yeah, I'm not sure whether this gets us all the way there (for all the cases we care about) but it does at least get closer (either to that, or other things we might want to do with `seq`).

iset.mm
benjub: No, actually I was thinking of what I wrote ;)
But this is probably not intuitionistic, since the proof I was thinking of is a case analysis: classically, the intersection is either empty (so is a function), or is the singleton `{ <. A , B >. }`, but this happens only if A \in V and B \in W, so we can use funsng.

jkingdon: > actually I was thinking of what I wrote ;)

Where `V` and `W` are arbitrary classes (not otherwise mentioned in the theorem)? I suppose that follows from funinsn from this pull request and http://us.metamath.org/ileuni/funin.html .

I agree the obvious proofs here are via case elimination. That's why I thought it was cool I found a proof of funinsn without it.

benjub: (proved in #2590)

PR #2590, opened 2022-05-05
imisc

benjub:
(imisc is to misc what iset is to set)

Generalize ~funinsn as hinted in #2588. I did it "silently" (i.e., no credit or date tag) since this is basically @jkingdon's proof. Notice that the new ~funinsn is a strict generalization of the old one, in the sense that the old one is a substitution instance of the new one (equivalently, has a 1-essential-step-proof from it), and I think it is generally good to prove the most general instance of a particular theorem (from a given set of axioms).

I also removed some dv's indicated by @jamesjer's script.  I also noticed (but left) some extraneous dv's  (e.g. in frecuzrdgfunlem), which a more precise script may detect in the future (this is doable, since for instance mmj2 finds the necessary dv's to add to each proof).


avekens (approved):
OK for me, if @jkingdon accepts this.

jkingdon (approved):
Nice to see those classes, and glad to see it was pretty simple to add them.

As for the distinct variables, I've been running the `check-dvs.py` script some, but apparently I missed a few. Thanks for noticing and fixing them!

jkingdon:
> OK for me, if @jkingdon accepts this.

Feel free to approve stuff without me, in case I'm busy or traveling or anything else.

But in this case, I do approve it so I guess the issue doesn't come up.

PR #2592, opened 2022-05-06
Remove S e. V from various seq theorems.

jkingdon:
Remove S e. V hypothesis from iseqfveq2 , iseqfeq2 , iseqfveq , iseqfeq ,
iseqid3s , iseqm1 and iseqid .

Update proofs for iseqz , bcn2 , serif0 , iser0 and isumrblem accordingly.


benjub (approved):

benjub:
@jkingdon @tirix: help ! The CI fails because of mmverify.py which I just changed at https://github.com/david-a-wheeler/mmverify.py. The CI complains:
```
Traceback (most recent call last):
  File "./mmverify.faster.py", line 42, in <module>
    Stmt = list[Symbol]
TypeError: 'type' object is not subscriptable
```
but mmverify.py does run on my computer. This might be due to the Python version used by the CI. How do we know that version ? And how to change it if possible ?
I think this might be a version problem because the syntax `list[Symbol]` was introduced with https://peps.python.org/pep-0585/ which was implemented in Python 3.9.

**Edit**: probably, verifiers.yml has to be completed, see for instance https://github.com/actions/setup-python. Something like replacing
```
- uses: actions/checkout@v2
```
with
```
- uses: actions/checkout@v3
- uses: actions/setup-python@v3
  with:
    python-version: '>=3.9' 
```
?

jkingdon:
> Python 3.9

You pretty much answered your own question, and your suggested syntax for getting that version of python worked - see #2593

tirix (approved):

PR #2593, opened 2022-05-07
Update to python 3.9

jkingdon:
With a recent change in mmverify.py a newer python is needed.

This is a change suggested by Benoit Jubin @benjub .

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -197,7 +197,11 @@ jobs:
     needs: skip_dups
     if: ${{ needs.skip_dups.outputs.should_skip != 'true' }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+      # See https://github.com/actions/setup-python
+      - uses: actions/setup-python@v3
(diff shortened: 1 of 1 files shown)


tirix (approved):
Perfect, thanks Jim and Benoît!

benjub:
Thanks @jkingdon !
I see that the CI test to check mm.set/Main using mmverify.py took 19s where it used to take 69s, so this is about the four fold speedup I was seeing with the new mmverify.py. Probably this allows removing the splitting in two jobs:
```
      - run: ./mmverify.faster.py --stop-label mathbox < set.mm
      - run: ./mmverify.faster.py --begin-label mathbox < set.mm
```
and have only one job (whose runtime will still be less than other jobs, like the one using mmj2):
```
      - run: ./mmverify.faster.py set.mm
```
I'll propose a PR next week.

jkingdon:
> Probably this allows removing the splitting in two jobs

Yeah, as much as I'd usually lean towards many small jobs, this seems like a case where combining these two makes sense now that it is fast enough.

Issue #2595, opened 2022-05-09
Prove qnnen in iset.mm


jkingdon:
This is stated in http://us.metamath.org/mpeuni/qnnen.html but the iset.mm proof likely will need to be different in several ways:

* We will presumably prove http://us.metamath.org/mpeuni/xpnnen.html (well, or more exactly we'll show a one to one and onto mapping between `NN0 X. NN0` and `NN0`, which is close enough) by means of the Cantor diagonal function, and explicitly inverting it as shown in https://en.wikipedia.org/wiki/Pairing_function#Inverting_the_Cantor_pairing_function .

  * To do it this way will presumably require a number of http://us.metamath.org/ileuni/df-fl.html related theorems beyond the rationals. I think it is likely true that the floor function from inverting the pairing function is always either irrational (and thus apart from any integer, and thus we can show it has a floor), or rational (in which case all the existing floor related theorems in iset.mm work), although I might have to spend a bit more time with https://en.wikipedia.org/wiki/Quadratic_irrational_number before I understand all the issues here.

  * There are a number of steps involved in showing that this closed form involving floor inverts the pairing function, but hopefully these are straightforward :crossed_fingers: 

* Once we have `ZZ X. NN ~~ NN` there is still another step, which is that there are multiple elements of `ZZ X. NN` which correspond to a given rational number. Is is true that `_om ~<_ A /\ A ~<_ _om -> A ~~ _om` ? We know that the full http://us.metamath.org/mpeuni/sbth.html implies excluded middle but some special cases - so far http://us.metamath.org/ileuni/fisbth.html - are provable.

  * Presumably there is some other way to prove this part if we don't get `_om ~<_ A /\ A ~<_ _om -> A ~~ _om`

tirix:
Worst case, an obvious way would be to construct a bijection between NN and QQ, which I believe is purely computational and should be feasible in iset.mm. 
[Math Stack Exchange](https://math.stackexchange.com/questions/7643/produce-an-explicit-bijection-between-rationals-and-naturals) / [Math Overflow](https://mathoverflow.net/questions/200656/is-there-a-natural-bijection-from-mathbbn-to-mathbbq) have examples with e.g. continued fraction expansions.

[...]

PR #2597, opened 2022-05-12
Equinumerosity in iset.mm

jkingdon:
Add a number of equinumerosity theorems, in particular `( NN X. NN ) ~~ NN ` using an approach suggested by @tirix in #2595  .

Also there are a few theorems about the floor of an irrational number, although we'd need to go further with this if we wanted to, for example, formalize the Cantor pairing function using floor.


iset.mm
```diff
@@ -111968,6 +112091,60 @@ infinite descent (here implemented by strong induction).  (Contributed
...
+    $d x y z $.
+    $( The Cartesian product of the set of positive integers with itself is
+       equinumerous to the set of positive integers.  (Contributed by NM,
+       1-Aug-2004.)  (Revised by Mario Carneiro, 9-Mar-2013.) $)
+    xpnnen $p |- ( NN X. NN ) ~~ NN $=
+      ( vz vx vy cn c2 cv cdvds wbr wn crab cn0 cxp cexp co cmul cmpt2 wf1o cen
+      ccnv eqid oddpwdc f1ocnv nnex f1oen mp2b oddennn nn0ennn mp2an entr2i
+      xpen ) DEAFGHIADJZKLZDDLZULDBCUKKECFMNBFONPZQDULUNSZQDULRHBCAUNUKUKTUNTUA
+      ULDUNUBDULUOUCUDUEUKDRHKDRHULUMRHAUFUGUKDKDUJUHUI $.
+  $}
```
tirix: I'd say your are the prime contributor for this version, since your proof is different from Mario's ?

iset.mm
```diff
@@ -111968,6 +112091,60 @@ infinite descent (here implemented by strong induction).  (Contributed
...
+
+  $( The Cartesian product of omega (the set of ordinal natural numbers) with
+     itself is equinumerous to omega.  Exercise 1 of [Enderton] p. 133.
+     (Contributed by NM, 23-Jul-2004.) $)
+  xpomen $p |- ( _om X. _om ) ~~ _om $=
+    ( cn cxp com cen wbr nnenom xpen mp2an xpnnen entri entr3i ) AABZCCBZCACDEZ
+    NLMDEFFACACGHLACIFJK $.
+
+
```
tirix: Same here, this seems to differ enough from Norm's original proof. Up to you.

tirix (approved):

iset.mm
jkingdon: Good catch. Assuming that the 2013 revision was to the proof rather than the statement of the theorem, it is a clear-cut case for removing `Revised by Mario Carneiro, 9-Mar-2013`. So I've pushed up a change to do that here and for the similar situation in `xpen`.

As for the `Contributed by` line itself, my usual practice has been to keep it if the statement of the theorem didn't change, or replace it with myself and the current date if the statement of the theorem changed. Although sometimes people have found this rule too simplistic, it is at least an easy one to apply.

My bigger dilemma here was more whether to credit you (@tirix ) since you provided the key idea of using `oddpwdc`. I'll be glad to say in a github comment for sure that that suggestion was very helpful and that I was headed down a rather more difficult path without it, but as for whether/how to reflect that in a `Contributed` line? I guess I rarely have a very clear idea of where to draw any of the lines there.

tirix: No problem, my reward is to know that I helped a bit here, and that `oddpwdc` is a useful theorem after all!

david-a-wheeler:
> My bigger dilemma here was more whether to credit you (@tirix ) since you provided the key idea of using oddpwdc.

Use your judgment, but I think in case of doubt it's best to give credit somehow. Most results are from the work of many people, so it is unsurprising when it's noted.

iset.mm
jkingdon: > No problem, my reward is to know that I helped a bit here, and that `oddpwdc` is a useful theorem after all!

Well you are listed as the contributor of `oddpwdc` so perhaps that is the right way to approach this.

jkingdon:
> > My bigger dilemma here was more whether to credit you (@tirix ) since you provided the key idea of using oddpwdc.
> 
> Use your judgment, but I think in case of doubt it's best to give credit somehow. Most results are from the work of many people, so it is unsurprising when it's noted.

In this case I think it is relatively clear that having the credit be on `oddpwdc` is the right place.

It is a fair point though especially for first time contributors or those who don't have a lot of credits. We should figure out some way in those sorts of situations.

PR #2600, opened 2022-05-17
replace ad5ant1345 with adantl3r

wlammen:
Theorem ad5ant1345 has currently (at least) two duplicates, namely adantl3r and adantlllr.  This PR reduces the usage to that of ad5ant1345.
I preferred ad5ant1345 over adantl3r, because it uses the greek letters in the usual order ph, ps, ch ..., which helps locating it in set.mm by pure text search for the formula.  The second duplicate adantlllr is an alias of adantl3r in a mathbox (so it seems to me)  introduced after a large proof was created unaware of adantl3r, later avoiding a time-consuming rewrite of the proof after adantl3r was detected.  This interpretation underlines the necessity to suppress proliferation of duplicates, and adhere to standard use patterns in formula.
Two proofs ([climxlim2lem](http://us2.metamath.org/mpeuni/climxlim2lem.html)  [hspmbllem2](http://us2.metamath.org/mpeuni/hspmbllem2.html) used two variants in parallel!

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -596,6 +596,7 @@
 "abshicom" is used by "bcs3".
 "ac2" is used by "ac3".
 "ac3" is used by "axac2".
+"adantl3rOLD" is used by "ad5ant1345OLD".
 "addassnq" is used by "addasspr".
 "addassnq" is used by "ltaddnq".
 "addassnq" is used by "ltexprlem7".
@@ -13568,6 +13569,7 @@ New usage of "ad4ant23OLD" is discouraged (0 uses).
 New usage of "ad4ant24OLD" is discouraged (0 uses).
 New usage of "ad4antlrOLD" is discouraged (0 uses).
 New usage of "ad4antrOLD" is discouraged (0 uses).
+New usage of "ad5ant1345OLD" is discouraged (0 uses).
 New usage of "ad5ant13OLD" is discouraged (0 uses).
 New usage of "ad5ant14OLD" is discouraged (0 uses).
 New usage of "ad5ant15OLD" is discouraged (0 uses).
@@ -13588,6 +13590,7 @@ New usage of "ad8antrOLD" is discouraged (0 uses).
 New usage of "ad9antlrOLD" is discouraged (0 uses).
 New usage of "ad9antrOLD" is discouraged (0 uses).
 New usage of "adant423OLD" is discouraged (0 uses).
+New usage of "adantl3rOLD" is discouraged (1 uses).
 New usage of "addassnq" is discouraged (4 uses).
 New usage of "addasspi" is discouraged (1 uses).
 New usage of "addasspr" is discouraged (17 uses).
@@ -18352,6 +18355,7 @@ Proof modification of "ad4ant23OLD" is discouraged (25 steps).
 Proof modification of "ad4ant24OLD" is discouraged (18 steps).
 Proof modification of "ad4antlrOLD" is discouraged (17 steps).
 Proof modification of "ad4antrOLD" is discouraged (17 steps).
+Proof modification of "ad5ant1345OLD" is discouraged (8 steps).
 Proof modification of "ad5ant13OLD" is discouraged (49 steps).
 Proof modification of "ad5ant14OLD" is discouraged (57 steps).
 Proof modification of "ad5ant15OLD" is discouraged (66 steps).
@@ -18372,6 +18376,7 @@ Proof modification of "ad8antrOLD" is discouraged (29 steps).
 Proof modification of "ad9antlrOLD" is discouraged (32 steps).
 Proof modification of "ad9antrOLD" is discouraged (32 steps).
 Proof modification of "adant423OLD" is discouraged (15 steps).
+Proof modification of "adantl3rOLD" is discouraged (27 steps).
 Proof modification of "addltmulALT" is discouraged (497 steps).
 Proof modification of "addmodlteqALT" is discouraged (237 steps).
 Proof modification of "aecom-o" is discouraged (28 steps).
--- a/discouraged
+++ b/discouraged
@@ -596,7 +596,6 @@
 "abshicom" is used by "bcs3".
 "ac2" is used by "ac3".
 "ac3" is used by "axac2".
-"adantl3rOLD" is used by "ad5ant1345OLD".
 "addassnq" is used by "addasspr".
 "addassnq" is used by "ltaddnq".
 "addassnq" is used by "ltexprlem7".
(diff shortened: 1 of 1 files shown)


tirix:
I agree with the argument of using Alan Sare's version, which is predating mine, and the choice of variables.

However how about the naming choice? `adantl3r` is shorter, and is but one in a series of theorems with a pattern of names:
```
adantlr  $p |- ( ( ( ph /\ th ) /\ ps ) -> ch ) $=
adantllr $p |- ( ( ( ( ph /\ ta ) /\ ps ) /\ ch ) -> th ) $=
adantl3r $p |- ( ( ( ( ( ph /\ si ) /\ rh ) /\ mu ) /\ la ) -> ka ) $=
adantl4r $p |- ( ( ( ( ( ( ph /\ ze ) /\ si ) /\ rh ) /\ mu ) /\ la ) -> ka ) $=
adantl5r $p |- ( ( ( ( ( ( ( ph /\ et ) /\ ze ) /\ si ) /\ rh ) /\ mu ) /\ la ) -> ka ) $=
adantl6r $p |- ( ( ( ( ( ( ( ( ph /\ ta ) /\ et ) /\ ze ) /\ si ) /\ rh ) /\ mu ) /\ la ) -> ka ) $=
```

(there is a similar pattern in naming `ad4antlr`, `ad5antlr`...)

wlammen:
Changing the name of the theorem is in fact a matter of pure text replacement.  The same does not hold for the naming of the variables.  This modifies the proofs in a subtle manner not seen on the essential proof step level, and that made this pull request in a sense hard for me, since I had to replay the proofs step by step.  At least this is the level of my knowledge.

I anticipated that there would be some discussion on the name, but since an update is easy, I am open to its outcome.  Still, I feel the ad5ant1345 scheme is a bit more appealing to me, since I can tell at a glance, where the additions are made.

benjub:
> Changing the name of the theorem is in fact a matter of pure text replacement. The same does not hold for the naming of the variables. This modifies the proofs in a subtle manner not seen on the essential proof step level

Side remark: I faced that difficulty several times. Say you want to rename some variables in the assertion xxx:
1. Add a theorem xxxNEW with the new variable names.
2. Do `MM-PA> minimize_with xxxNEW / may_grow` on all theorems using xxx (one can write easy script like `scripts/min.cmd` by using the output of `MM> show usage xxx`). After that, no proofs should be using xxx.
3. Remove the theorem xxx.
4. Do a global text replacement of xxxNEW with xxx.

Another possibility would be to temporarily save the proofs of the theorems using xxx in the "normal" (or "explicit"?) format, with `MM-PA> save proof yyy /normal`, and then do a global text replacement.

wlammen:
Thanks @benjub, this will be of great help next time.

digama0:
My preference is also for the adantl3r name. ad5ant1345 has too much non-orthogonal naming in it (the 5 has to match the latter numbers, the numbers have to be in order...), and it also does not indicate how the conjuncts are associated. I much prefer a naming system that derives from "this theorem is the same as `adantl(adantl(adantl(adantr(h)))`" using similar conventions to the `cadaddr`-style lisp functions, and the `l3r` is just run length encoding the name to avoid it getting too long.

wlammen:
I don't quite understand how you arrive at one of the formulas (except the first) tirix gave above by applying adantl multiple times and adantr once.  Anyway, no matter which name you prefer, it will rip a hole in either the series ad5ant? or adantl?r.  So it boils down to the question which hurts less for the moment. Mid term we need a consistent (at least) and simple to understand (at best) theorem naming scheme for adding antecedents in various locations.  This goal is beyond the scope of this PR and I suggest to cover it in a separate issue.  If the majority goes for adantl3r as a quick and dirty solution, because it suits current habits best, I am fine with that, and will adapt the PR accordingly, say in the next 24 hours.  To still give a limited time frame for expressing opinions.

tirix:
FYI, I did not show the `$e` hypotheses in my previous posts, just the final statements. Here you have the first few:
```
adant2.1 $e |- ( (   ph         /\ ps ) -> ch ) $.
adantlr $p  |- ( ( ( ph /\ th ) /\ ps ) -> ch ) $=
```
```
adantl2.1 $e |- ( ( (   ph         /\ ps ) /\ ch ) -> th ) $.
adantllr $p  |- ( ( ( ( ph /\ ta ) /\ ps ) /\ ch ) -> th ) $=
```
```
adantl3r.1 $e |- ( ( ( (   ph         /\ ps ) /\ ch ) /\ th ) -> ta ) $.
ad5antl3r $p  |- ( ( ( ( ( ph /\ et ) /\ ps ) /\ ch ) /\ th ) -> ta ) $=
```

wlammen:
So the factor 3 applies to the 'r', not the 'l'.  Yes that makes sense.

But compare to the version adantlllr, where this is spelled out.  Ok it is as it is.  I suggest to address this confusion outside of this PR and for now just vote which name should survive short term.

wlammen:
replaced theorem name ad5ant1345 with adantl3r

tirix (approved):
Looks good to me, thanks for noticing the duplicate, and doing the change!
We can further discuss naming, and choice of variables in another PR/issue.

wlammen:
Some proofs suffered from growing when I used the 'improve xxx' command while replaying.  I used @benjub 's method to be closer to the original.  Affected are dfcgra2, omssubadd and circlemeth.

All of these now use the same numbers of essential proof lines than before.  I created html pages for all proofs.  I don't know of one that uses more lines than before.

avekens (approved):

avekens:
> > Changing the name of the theorem is in fact a matter of pure text replacement. The same does not hold for the naming of the variables. This modifies the proofs in a subtle manner not seen on the essential proof step level
> 
> Side remark: I faced that difficulty several times. Say you want to rename some variables in the assertion xxx:
> 
>     1. Add a theorem xxxNEW with the new variable names.
> 
>     2. Do `MM-PA> minimize_with xxxNEW / may_grow` on all theorems using xxx (one can write easy script like `scripts/min.cmd` by using the output of `MM> show usage xxx`). After that, no proofs should be using xxx.
> 
>     3. Remove the theorem xxx.
> 
>     4. Do a global text replacement of xxxNEW with xxx.
> 
> 
> Another possibility would be to temporarily save the proofs of the theorems using xxx in the "normal" (or "explicit"?) format, with `MM-PA> save proof yyy /normal`, and then do a global text replacement.

This hint would be an ideal item of a "Tips and Tricks" page I proposed some time ago. Is there any plan/idea when and how to provide such a page?

wlammen:
It would be even better to have it as command in Metamath available, now that at least two persons used it.

But I support a Tips and Tricks page, too.

PR #2601, opened 2022-05-21
Use bicom1 and syl in bicomd

icecream17:
I don't know if this counts as an improvement, but if there was a theoretical tree starting from `bicomd`, using `bicom1 syl` instead of `bicom sylib` is simpler.


wlammen:
Here us.metamath.org/mpeuni/conventions.html are some metrics listed that I used in the past to decide whether an improvement is sufficient for a change.

Occasionally I submit a change not covered by that paper.  My PR #2600 is such an example: Removing duplicates is not listed as a goal, so I pointed to some effects I considered negative enough to suggest a change.  I'd rather see that amended to the paper, though.

avekens:
Maybe it is a common pattern to use ~sylibr to derive a deduction form from a closed form of an equivalence, see ~con2bid, ~inegd, ~nfd, ~nfxfrd, etc. Hence, the original proof should be kept. I admit, however, that the proposed proof is a little bit more elegant.  Therefore, I would propose to add the new proof as an altenative (as theorem ~bicomdALT).

wlammen:
There are more examples like this one in the database (orcomd, ancomd), so this opens the door for many more such alternatives.  The tree height of the proof depends on whether bicom1 or bicom is more primitive.  This is not always obvious (see orass vs pm2.31), and might change.

For me a proof design is a creative act of the author that I want to keep unless I have a good reason for a change.  We are here in the range of personal preferences (elegant).  I wonder whether we should open this can of worms.

tirix:
What about this pull request? It's been sitting idle for a week @avekens @wlammen.
I must admit I don't see what's wrong. In terms of compressed proof length, it's shorter.

wlammen:
Since the names of the theorems are arbitrary, it is the block of compressed proof steps (here for both ABCECBEDBCFG) that counts.  This is not a shortening, it is a different proof, having the same length by all measures accepted so far.
 
The current hierarchy of criteria used in the shortening process, are given in [conventions](http://us.metamath.org/mpeuni1/conventions.html) in the paragraphs "Acceptable shorter proofs" and "Minimizing axioms and the axiom of choice".  The new proof does not match any of them.

One can of course devise a new metric in favour of one of them (for example: expands to a smaller number of steps comprising of axioms only; the height (or its average) of the proof tree from conclusion to the axioms, and more).  The problem: all of them express just a random preference.  This holds for the current proof, too.  It was the one its author once preferred for whatever reason.  So far this wins in all conflicts.

benjub:
No strong opinion. The proposed proof is in some sense more economical than the current one which first proves a biconditional (bicom is proved from bicom1) and then uses only one direction of the biconditional (sylb uses biimpi). This probably shows in `MM> show trace bicomd/essential` versus the same command with the suggested proof (see http://us2.metamath.org/mpeuni/mmset.html#trivia).

On the other hand, it's a tie in terms of both number of essential steps and total number of steps.  There are probably dozens of similar cases (whether producing/consuming a biconditional, or being content with one direction). Maybe as @avekens suggests, @icecream17 can propose his proof as an *ALT (with a comment like this one to explain it), but afther this, we should not waste too much energy in going back and forth on these basic lemmas.

PR #2605, opened 2022-05-29
Add Euler's phi (totient) function to iset.mm

### LINKED ISSUE #2528: How should we use # in iset.mm?
Here's a summary of the status quo (after #2525 ) with respect to usage of the ` # ` symbol or variations thereof:

| file    | abbreviation | ASCII | typeset | meaning | example
|--------|------------------|---------|------------|-------------|-----
| set.mm  | hash         | #     | #       | set size function | [hashen](http://us.metamath.org/mpeuni/hashen.html)
| iset.mm | ap           | #     | #       | apartness relation | [apcotr](http://us.metamath.org/ileuni/apcotr.html)
| iset.mm | size         | size  | ♯   | set size function | [sizeen](http://us.metamath.org/ileuni/sizeen.html)

I'm thinking of maybe changing this to:

| file    | abbreviation | ASCII | typeset | meaning
|--------|------------------|---------|------------|-------------
| set.mm  | hash         | #     | #       | set size function
| iset.mm | ap           | =//=  | #       | apartness relation
| iset.mm | hash         | #     | ♯   | set size function

Perhaps my strongest motive is to reduce the need to change "#" to "size" when copy-pasting expressions from set.mm to iset.mm. Also changing the abbreviation used in names would reduce needless theorem renames when they are driven just by "hash" versus "size".

The suggestion of `=//=` is one that @digama0 made at https://github.com/metamath/set.mm/pull/1437#issuecomment-578433332 . Although apartness is common enough (it appears in pretty much any theorem involving division or multiplicative cancellation of real or complex numbers) that I'm now beginning to wonder whether the above is misguided and we should stick with the "status quo" table above. Hence me putting this forth in an issue for community discussion.

jkingdon:
This pull request contains:

1. copy a few small theorems from set.mm which work as-is
2. Add ssfirab which is a version of http://us.metamath.org/mpeuni/ssfi.html which we can prove because it has an additional decidability condition. (Full ssfi is not possible, see http://us.metamath.org/ileuni/ssfiexmid.html ).
3. Rename theorems concerning the set size function to have `hash` in their names (as in set.mm) rather than `size` (which is maybe better than `hash` although there doesn't seem to be a strong consensus to that effect and changing both set.mm and iset.mm would be a significant effort, so better to just adopt the set.mm naming for iset.mm). Fixes #2528 . (see that issue for more discussion of the naming).
4. Add the first part of the "Euler's theorem" section, from `df-phi` through `hashdvds`. This section does intuitionize well (in the sense that every theorem in that range is proved), although many of the proofs require various kinds of adjustments.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6725,11 +6725,6 @@ than reals.</TD>
 </TR>
 
 <TR>
-  <TD>hashnncl</TD>
-  <TD>~ sizenncl</TD>
-</TR>
-
-<TR>
   <TD>hash0</TD>
   <TD>~ size0</TD>
 </TR>
@@ -6841,7 +6836,7 @@ than reals.</TD>
 
 <TR>
   <TD>hashgt0 , hashge1</TD>
-  <TD>~ sizenncl</TD>
+  <TD>~ hashnncl</TD>
   <TD>It is not clear there would be any way to combine the finite
   and infinite cases.</TD>
 </TR>
@@ -6853,7 +6848,7 @@ than reals.</TD>
 
 <TR>
   <TD>hashnn0n0nn</TD>
-  <TD>~ sizenncl</TD>
+  <TD>~ hashnncl</TD>
   <TD>To the extent this is reverse closure, we probably can't prove
   it. For inhabited versus non-empty, see ~ fin0</TD>
 </TR>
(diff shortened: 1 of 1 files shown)


tirix (approved):
Nice! I agree that "hash" is not a very good name, but I think it's better to align with `set.mm`.

jkingdon:
> Nice! I agree that "hash" is not a very good name, but I think it's better to align with `set.mm`.

We can change the naming some day if we want but I'd rather do that only if we change it in both set.mm and iset.mm. But that's for the future, the changes here seem best for now.

PR #2606, opened 2022-05-29
mostly shortenings

wlammen:
a. The shortenings in this pull request are mostly trivial.  The core idea behind the majority of them is to defer adding conjuncts to the latest possible, so they don't litter intermediate formulas unnecessarily.  Besides these there are a few shortenings with a bit more background, like 3ancoma or 3simpb.
b. OLD theorems very often had the description "Obsolete proof of ...". The web page [conventions](http://us2.metamath.org/mpeuni/conventions.html) suggests instead "Obsolete version of...".  I replaced the descriptions in the main part to arrive at a consistent style.
c. Shortenings required 3ad* theorems be clustered in a single location, so I "defragmented" set.mm in this respect a bit.
d. Removed an outdated OLD theorem.


benjub:
I would generally write "obsolete proof" if the statement is identical and "obsolete version" if the statement is different (generally because it removes a hypothesis or antecedent, or is more general in some way, or because a modified statement increases consistency with theorems in the same family or permits theorems using it to have shorter proofs). Since OLD theorems are to be ultimately deleted, this is not too important.

As for shortenings, I understand how a. reduces the total number of steps, and it is indeed a good thing. But it may be that this reduction overbalances an increase in the number of essential steps. I know that the "official rules" for shorter proofs look first at total number of steps, but I think that if there is a proof with fewer essential steps, it should at least appear as an ALT. (I haven't checked if this happens in this PR.)

wlammen:
> I haven't checked if this happens in this PR

It does not.

The wording "obsolete proof" is semantically a tiny bit more to the point in practically all instances where I replaced it.  So I deliberately kept it in 19.43OLD.  Still, I wonder whether we should sacrifice a convention for this small advantage.  One should also bear in mind that this comment appears in OLD theorems only, bound for deletion after a year.

I mostly made the change to allow for automatic removal of OLD theorems by a script in future, where it could ease phrase matching.

wlammen:
I am a bit at a loss to understand what is going on here.  Are the shortenings questioned?  Or the rules in Conventions and Styles, and how (whether?) they should apply?

To get on, I am going to close this PR now.

PR #2607, opened 2022-06-01
Shorten proof of ssdifsn

jamesjer:


wlammen (approved):

wlammen:
The only downside I noticed, is that the new proof uses df-nul, whereas the old one avoids it.  I don't consider this important here.

wlammen:
... and the OLD version has the description "Obsolete proof ..." instead of "Obsolete version..." as formally required by [Conventions and Style](http://us2.metamath.org:88/mpeuni/conventions.html), paragraph "Old (OLD) versions or proofs".
This is already part of my PR #2606, and can be decided there.

jamesjer:
Okay, I guess I was just copying what I saw elsewhere in set.mm instead of actually looking at Conventions and Style.  Shame on me.  I'm happy to change this PR if the consensus is to do so.  Meanwhile, I realized the proof of ssdifsn can be shortened still further.  The new version adds both df-ne and df-nul, but is shorter yet by another 27 bytes and 22 steps.

wlammen (approved):
Approve again

tirix (approved):

PR #2608, opened 2022-06-04
First pass at the rest of "Euler's theorem" section in iset.mm

jkingdon:
Although some of the most important theorems aren't here yet, we can prove some of the theorems in this section and add things I noticed along the way. The main thing holding the missing theorems back seems to be http://us.metamath.org/mpeuni/f1finf1o.html which seems like it should be provable but I'm not sure how.

Contains:
* as stated above, some of the theorems from the Euler's Theorem section
* prove a modified version of http://us.metamath.org/mpeuni/prfi.html
* prove a number of set size related theorems (and add others to the missing theorems list in mmil.html)
* prove http://us.metamath.org/mpeuni/xpfi.html - finite sets have advanced to the point where this is now straightforward intuitionizing
* add a version of http://us.metamath.org/mpeuni/ifbothda.html with a decidability condition added

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3186,13 +3186,6 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
-  <TD>xpfi</TD>
-  <TD><I>none</I></TD>
-  <TD>This is Lemma 8.1.16 of [AczelRathjen] and is therefore presumably
-  provable.</TD>
-</TR>
-
-<TR>
   <TD ROWSPAN="2">prfi</TD>
   <TD>~ prfidisj</TD>
   <TD>for two unequal sets</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6928,6 +6928,14 @@ than reals.</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


avekens (approved):

jkingdon:
> The main thing holding the missing theorems back seems to be http://us.metamath.org/mpeuni/f1finf1o.html 

Good news! I found a proof of this which is now on the https://github.com/jkingdon/set.mm/tree/nelelne branch, and at least so far it seems to unblock the rest of the theorems from the "Euler's theorem" section.

I'll make this a pull request soon, probably when that section is done.

jkingdon:
> Good news! I found a proof of this which is now on the https://github.com/jkingdon/set.mm/tree/nelelne branch

Oh and in light of that branch being there no one can accuse me of saying "I have a truly marvelous demonstration of this proposition that this margin is too narrow to contain". I mean, it is indeed marvelous (or at least marvelous enough in the sense that I've been searching for this for many months and what ended up working was not brute force or piling on more and more steps, but taking a different approach - of induction on A rather than on B). But it does fit in the github "margin".

PR #2615, opened 2022-06-14
shorten simpl, simpr

wlammen:
while investigating whether to begin 3adant* theorems with a 3simp* or adant* theorem I found these shortenings.   The small visual impact is in the order of replacing an idd line with an id

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17851,11 +17851,13 @@ New usage of "simp-8rOLD" is discouraged (0 uses).
 New usage of "simp-9lOLD" is discouraged (0 uses).
 New usage of "simp-9rOLD" is discouraged (0 uses).
 New usage of "simpl2imOLD" is discouraged (0 uses).
+New usage of "simplOLD" is discouraged (0 uses).
 New usage of "simplbi2VD" is discouraged (0 uses).
 New usage of "simplbi2comtVD" is discouraged (0 uses).
 New usage of "simplbiimOLD" is discouraged (0 uses).
 New usage of "simplllOLD" is discouraged (0 uses).
 New usage of "simpllrOLD" is discouraged (0 uses).
+New usage of "simprOLD" is discouraged (0 uses).
 New usage of "sineq0ALT" is discouraged (0 uses).
 New usage of "smcn" is discouraged (3 uses).
 New usage of "smcnlem" is discouraged (1 uses).
--- a/discouraged
+++ b/discouraged
@@ -19883,11 +19883,13 @@ Proof modification of "simp-8rOLD" is discouraged (27 steps).
 Proof modification of "simp-9lOLD" is discouraged (30 steps).
 Proof modification of "simp-9rOLD" is discouraged (30 steps).
 Proof modification of "simpl2imOLD" is discouraged (12 steps).
+Proof modification of "simplOLD" is discouraged (7 steps).
 Proof modification of "simplbi2VD" is discouraged (24 steps).
 Proof modification of "simplbi2comtVD" is discouraged (42 steps).
 Proof modification of "simplbiimOLD" is discouraged (12 steps).
 Proof modification of "simplllOLD" is discouraged (10 steps).
 Proof modification of "simpllrOLD" is discouraged (10 steps).
+Proof modification of "simprOLD" is discouraged (7 steps).
 Proof modification of "sineq0ALT" is discouraged (986 steps).
 Proof modification of "smgrpassOLD" is discouraged (54 steps).
 Proof modification of "smgrpismgmOLD" is discouraged (22 steps).


set.mm
```diff
@@ -16854,10 +16854,32 @@ inference eliminates the need for a syllogism ( ~ syl ) -type inference
...
+  $}
+
+  ${
+    adantl.1 $e |- ( ph -> ps ) $.
+    $( Inference adding a conjunct to the left of an antecedent.  (Contributed
+       by NM, 30-Aug-1993.)  (Proof shortened by Wolf Lammen, 23-Nov-2012.) $)
+    adantl $p |- ( ( ch /\ ph ) -> ps ) $=
+      ( adantr ancoms ) ACBABCDEF $.
+  $}
+
```
jamesjer, line 16872: I know that adantl is just being moved, not changed, but I want to point out that proving adantl with impcom and a1d shaves 3 bytes off the proof.

wlammen: It does not change the size of the compressed proof, which stays unchanged at 9 bytes.  What becomes shorter: The names of the used theorems happen to be shorter, but that is a complete random effect, as the names are somewhat arbitrary.  I can see some advantage in your suggestion, as the proof (1) parallels that of adantr; (2) is based on earlier theorems.  The latter is dependent on what axiomization was picked, though.  In a propositional logic based on meredith, for example, this judgement may change dramatically.

I am very much used to Norms ideas of a "short proof".  It appears to me that he viewed the theorems in a layered fashion.  There are the axioms at the lowest level.  Built upon them are direct consequences.  Further development leads to more complicated deductions in higher levels.  To find a better proof you may pick any out of the previous levels, and he was indifferent about which one you came up with.  He allowed me to even move theorems across levels if that resulted in shorter compressed proofs in total.  That way 'useful' theorems tend to bubble up to earlier positions, as is seen with adantl in this PR.  I never heard of it, but I wouldn't be surprised if discussions about what makes up a good proof were around before my first contributions.  But somehow he settled on this and wrote a few rules down on his Contribution page.  I always stuck to them, and I will do so in this PR.

That said, if I had the opportunity to do fundamental changes, I would rearrange a couple of things.  But in the end, I fear, I'd just express my personal preferences without a sustained advantage over Norm's or anyone else's likings.

wlammen: It does not change the size of the compressed proof, which stays unchanged at 9 bytes.  What becomes shorter: The names of the used theorems happen to be shorter, but that is a completely random effect, as the names are somewhat arbitrary.  I can see some advantage in your suggestion, as the proof (1) parallels that of adantl; (2) is based on earlier theorems.  The latter is dependent on what axiomization was picked, though.  In a propositional logic based on meredith, for example, this judgement may change dramatically.

I am very much used to Norms ideas of a "short proof".  It appears to me that he viewed the theorems in a layered fashion.  There are the axioms at the lowest level.  Built upon them are direct consequences of them.  Further development leads to more complicated deductions in higher levels.  To find a better proof you may pick any out of the previous levels, and he was indifferent about which one you came up with.  He allowed me to even move theorems across levels if that resulted in shorter compressed proofs in total.  That way 'useful' theorems tend to bubble up to the beginning of set.mm, as is seen with adantl in this PR.  I never heard of it, but I wouldn't be surprised if discussions about what makes up a good proof were around before my first contributions.  But somehow he settled on this and wrote a few rules down on his Contribution page.  I always stuck to them, and I will do so in this PR.

That said, if I had the opportunity to do fundamental changes, I would rearrange a couple of things.  But in the end, I fear, I'd just express my personal preferences without a sustained advantage over Norm's or anyone else's likings.

avekens: I tried another proof using a1i and imp, which would be even closer to the proof of adantr, but it is 1 byte longer:
`( wi a1i imp ) CABABECDFG $.`
I do not exactly know why "wi" must be used here (required by a1i) and not for the other proofs...

In principle, I share Wolf's (and maybe Norm's, as described by Wolf) view on how to shorten and rearrange proofs. Maybe we have to be careful not to end up in endless rearrangements of proofs, depending on personal preferences. But maybe the rule "result in shorter compressed proofs **in total**" prevents this. This is a general statement only, in this concrete case I have no objections.

wlammen: in total meant: You may result in a longer proof of one (or more) theorem, if that is more than compensated for by shortenings of others (measured in compressed proof bytes).  This often is accompanied by a change in the hierarchy of  theorems.  The longer one tends to move towards the end, or becomes dependent on the shortened ones.  PR #2609 is an example for this.

jkingdon: > In principle, I share Wolf's (and maybe Norm's, as described by Wolf) view on how to shorten and rearrange proofs. Maybe we have to be careful not to end up in endless rearrangements of proofs, depending on personal preferences. But maybe the rule "result in shorter compressed proofs **in total**" prevents this.

Although I'm sure there are bigger shortenings to be had later in set.mm rather than in small rearrangements of propositional logic, I too support our current rules (which I think are codified in the "Acceptable shorter proofs" section of http://us.metamath.org/mpeuni/conventions.html ). Having some objective rules, even if they are a bit arbitrary, would perhaps be our best chance to avoid the "endless rearrangement" problem. Saying "don't rearrange without a compelling reason" could I suppose cut down rearrangements but that would also impact rearrangements that we do want and only shifts the problem to figuring out how good the reason has to be, to be compelling.

jkingdon (approved):

avekens (approved):

set.mm
wlammen: > I do not exactly know why "wi" must be used here (required by a1i) and not for the other proofs...

The wi type instruction appears when you have to build up a formula during replacement.  In your first step the standalone ph (in a1i) is replaced with the hypothesis of the theorem ( ph -> ps ).  That's where the wi is used.  By contrast, adantr does not need this step, since its hypothesis matches exactly in structure the hypothesis of the theorem.

I do not work as a mathematician, actually have no education in that respect, other than what is taught at school.  My contributions are privately motivated.  Occasionally I use a short break at work to answer to posts here.  Maybe I should stop doing so, because they are not always exactly to the point then.

I would rewrite some part of the second paragraph of my first post here.  For the sake of clarity I present the idea here:
I think Norm's shortening rules are a brutal manifestation of what is elsewhere known as the 'economic principle'.  Norm looked at the intrinsic complexity of a proof. The less proof lines it needs, and the fewer replacements take place, the better (more economic) the proof is.  Both aspects are combined in the compressed proof length.  It might not be the theoretical best measure, but it is IMO a really good match.  On the fringes it may produce some disputable results (as in the a1i case) but over all it is convincing.

My profession makes me work occasionally as a software architect.  Being able to produce compact interfaces, or even minimize them, is an asset in this field, and has measurable trade-offs.  So I do like shortenings (actually complexity reductions), and in my eyes they are of value.  Here we are again in the field of personal preferences, and there's 'nough said before.

All my last PRs, namely this one, #2613, #2609, #2606, #2602 were checked against the rules on the Conventions page, and I expected them to be waived through, to be honest.  Instead they take 4 days to complete on the average.  That is a bit discouraging, in particular if you see others passing by.  Was i punished for what I wrote in #2601?

benjub: Regarding your last paragraph: to me, the delay before either merging or rejecting is not an indication of the value of the PR. It is mainly an indication of the spare time the reviewers have (they also have works and do this in their spare time, at least for me). Also note that none of the PRs you mention were rejected: that's the only thing that matters. (As for the one you closed, I would have merged it, actually. I guess this was a busy week for me on the job side.)

jkingdon: > Instead they take 4 days to complete on the average

Can't speak for anyone else, but as for me (1) I do try to get to reviews, including yours, and (2) I am still in the process of learning about our shortening rules, so hopefully once I figure it out I'll be a bit more able to wave things through.

As for this pull request, it does now have approvals - are you able to merge it? If not, I think we just need to ask David or Mario to enable that permission for you (I would support doing so).

wlammen: I am not able to merge PRs, not even my own, whether they are approved or not.

Thanks for your support, though.

jkingdon: > I am not able to merge PRs, not even my own, whether they are approved or not.
> 
> Thanks for your support, though.

Hmm, that's part of what is creating the delays. I've made an inquiry to David and Mario so hopefully we can get to the bottom of this part of the issue.

PR #2617, opened 2022-06-16
A few theorems about finite sets, _om, and the like

jkingdon:
That `x e. Fin \/ _om ~<_ x` implies excluded middle came to me in a dream last night. (Well, OK, dream might be a bit of an exaggeration but I was thinking about it as I lay in bed, so close enough).

There are a few fixes to comments.

There is a version of http://us.metamath.org/mpeuni/isinf.html (but for our usual definition of infinite).

There is another special case of http://us.metamath.org/mpeuni/undif.html , this one where the containing set is `_om`.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3731,7 +3731,9 @@ this implies excluded middle</TD>
   <TD>~ infnfi</TD>
   <TD>Defining "A is infinite" as ` _om ~<_ A ` follows definition
   8.1.4 of [AczelRathjen], p. 71. It can presumably not be shown to
-  be equivalent to ` -. A e. Fin ` in the absence of excluded middle.</TD>
+  be equivalent to ` -. A e. Fin ` in the absence of excluded middle
+  (see ~ inffiexmid which isn't exactly about ` -. A e. Fin <->
+  _om ~<_ A` but which is close).</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -58497,6 +58497,20 @@ ordinals implies excluded middle ( ~ ordsucunielexmid ).  (Contributed
       MUNUOUP $.
   $}
 
+  ${
+    $d A x $.  $d B x $.
+    $( If we remove a single element from ` _om ` then put it back in, we end
+       up with ` _om ` back.  This strengthens ~ difsnss from subset to
+       equality but the proof relies on equality being decidable.  (Contributed
+       by Jim Kingdon, 11-Jun-2022.) $)
+    omdifsnid $p |- ( B e. _om -> ( ( _om \ { B } ) u. { B } ) = _om ) $=
```
digama0, line 58506: This generalizes, to `( ( A \ B ) u. B ) = A <-> A. x e. A DECID x e. B`

jkingdon: Hmm, interesting.I too noticed that it could be more general although I don't remember exactly which generalization I was thinking of (probably something similar to the above but not identical).

It is fair to say that I have `_om` on the brain which is partly why I didn't go right to the generalizations. The reason for this fixation is the question posed at https://github.com/metamath/set.mm/issues/2595 - if a set is countable and infinite can we conclude it is countably infinite? (that is, `_om ~<_ A /\ A ~<_ _om -> A ~~ _om` ).

jkingdon: The above-suggested generalization and one other (which I think is the one I had in mind before) is now an issue at https://github.com/metamath/set.mm/issues/2618

avekens (approved):

PR #2623, opened 2022-06-16
Interesting factorisation

Dante-Broggi:
In iset.mm, unlike in set.mm ~ ax-in2 can be non-trivially factored into what is here ~ ax-in2a and ~ ax-in2b .
In set.mm ~ ax-in2b is trivialized by general double negation elimination, so this factorization is trivial.
```
ax-in2  $a |- ( -. ph -> ( ph -> ps ) ) $.
ax-in2a $a |- ( -. ph -> ( ph -> -. ps ) ) $.
ax-in2b $a |- ( -. ph -> ( ph -> ( -. -. ps <-> ps ) ) ) $.
```

Although intuitionistic logic generally avoids things which can imply general double negation elimination, iset.mm still allows a contradiction to imply it.
I don't see why it shouldn't avoid such, however.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -228,6 +228,8 @@ New usage of "axcnre" is discouraged (0 uses).
 New usage of "axdistr" is discouraged (0 uses).
 New usage of "axi2m1" is discouraged (0 uses).
 New usage of "axicn" is discouraged (0 uses).
+New usage of "axin2a" is discouraged (0 uses).
+New usage of "axin2b" is discouraged (0 uses).
 New usage of "axltirr" is discouraged (1 uses).
 New usage of "axlttrn" is discouraged (1 uses).
 New usage of "axmulass" is discouraged (2 uses).


jkingdon:
First of all, thanks for the contribution! Always good to have submissions, and let us know if you have any questions about how this project operates.

I think I get the point of this pull request - that there are multiple ways to axiomize intuitionistic propositional logic, and that we want to show that several of them are equivalent. From that point of view, I do appreciate the proofs that the different axioms imply each other, although I'm a bit fuzzy on how we usually do that without getting ourselves confused about what axioms we are using. Browsing through set.mm I think I'm seeing that these kinds of explorations only use `$p` not `$a` statements, but I'm open to hearing other people's ideas about what we have done or what we should do.

Our main axiomization is currently described at http://us.metamath.org/ileuni/mmil.html#axioms and although I'm not sure why @digama0 chose that exact set, I'm operating under the assumption that the current discussion is about showing there are other possibilities rather than get into a (unresolvable?) debate about which of several equivalent axiomizations is best.

@benjub any thoughts?

digama0:
I'm not sure what the point of this factorization is. Both factors are more complicated than the original axiom.

It is true in general that (assuming EM for simplicity) we can take any axiom `A` and "factorize" it into `B -> A` and `-. B -> A` and it may well be that for an appropriate choice of `B` both axioms are strictly weaker than the original axiom `A` but together they imply it. Unless there is a reason to study the subsystem consisting of only one of these two axioms, I don't see why we would want to do this.

The current system of axioms was chosen to be as short as possible without using "weird" axioms that are less obviously true. (That is, it should consist of things that are already theorems in set.mm.) This was because it was a delta from set.mm which has a very short axiom system and I did not want to diverge from that any more than necessary.

benjub:
I agree with @jkingdon and @digama0. It's sometimes useful to find equivalent axiomatizations, and I generally favor having more (independent) axioms to allow modularity and studies of some subsystems, but this is only useful if the subsystems have some importance or interest of their own.
In the present case, I do not see a particular interest in dividing ax-in2 into two parts. Note that you could have chosen the simpler and weaker
```
ax-in2a $a |- ( -. ph -> ( ph -> -. -. ps ) ) $.
ax-in2b $a |- ( -. ph -> ( ph -> ( -. -. ps -> ps ) ) ) $.
```
but this does not seem to yield interesting subsystems. Axiom ax-in2 (ex contradictione quodlibet, or principle of explosion), is pretty "monolithic" (contrary to, say, ax-3 of set.mm, which bundles the principle of explosion, contraposition, and non-intuitionism).

jkingdon:
In case you aren't familiar with mathboxes, @Dante-Broggi , they are described at http://us.metamath.org/ileuni/mathbox.html and based on the above comments it would seem like this is a case where putting this in a mathbox might be appropriate if you are still interested in doing something.

I agree with the comments above that this factorisation doesn't have obvious opportunities for subsystems (well at least not opportunities we can see so far) or conciseness, but I also want to encourage you to continue to contribute to iset.mm (either on this topic, or something else). So let us know if there is anything we can do to help.

Dante-Broggi:
Perhaps I will move this to a mathbox.

The reason I consider this a factorization is that in many programming languages ~ ax-in2a can be implemented simply, 
similar to (I think) all iset.mm axioms except the original ~ ax-in2 , and is still useful for some of the same functions, but (I think) implementing ~ ax-in2b , without implicitly assuming all of ~ ax-in2 , requires multithreading:

```rust
pub fn ax_in2a<T, U>(t: T, nt: impl FnOnce(T) -> !) -> impl FnOnce(U) -> ! {
    move |_u| nt(t)
}

// after conjoining nested premises
pub fn ax_in2b<T: Send + Clone + 'static, U: Send + 'static>(t: T, nt: impl Fn(T) -> ! + Send + 'static, mut nnu: impl FnMut(Box<dyn Fn(U) -> ! + Send + 'static>) -> ! + Send + 'static) -> U {
    let (send, recv) = std::sync::mpsc::channel();
    std::thread::spawn(move || {
        nnu(Box::new(move |u| { send.send(u).unwrap(); nt(t.clone()) }))
    });
    recv.recv().unwrap()
}
```

jkingdon:
> Perhaps I will move this to a mathbox.

Makes sense to me. There's isn't much of a high bar for that (other than the obvious ones like passing the automated checks - as described in the mathbox page I already linked to and the mathbox text at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md ), so this should go right in with little or none of the "why would you factor it that way?" questions.

> The reason I consider this a factorization is that in many programming languages ~ ax-in2a can be implemented simply, similar to (I think) all iset.mm axioms except the original ~ ax-in2 

Oh cool. This is a bit over my head (well, a lot over my head) in terms of the theory of programming languages, so I can't really peer-review whether what you say is true, but it does sound like an interesting motivation.

Assuming you do turn this into a mathbox submission, I'd encourage you to put some or all of this into a comment (or a reference to papers which cover the topic, if any).

digama0:
@Dante-Broggi Regarding the curry-howard correspondence to turn the problem into a statement about types in a programming language, there are no problems with `ax-in2` as is: ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9198bdeae842f4a045222e087a9ac8d))
```rust
pub fn ax_in2<T, U>(t: T, nt: fn(T) -> !) -> U {
     nt(t)
}
```
Note that this is taking advantage of the fact that `!` coerces to `U` in rust, but you can also write it as `match nt(t) {}` if you find that to be cheating. (I also changed `impl FnOnce` in the signature to `fn` because using `!` in the former case is unstable.

benjub:
I'm closing this PR, but do know that we welcome new contributions in mathboxes, if you wish. Just insert one for you in alphabetical order, see http://us2.metamath.org/mpeuni/mathbox.html

Issue #756, opened 2019-02-17
Rename numbered axioms in iset.mm


jkingdon:
Currently the predicate logic axioms in iset.mm have names which are based on the old axiom numbering in set.mm.

The reasons that set.mm uses numbers, rather than names, are described at #703 and in particular https://groups.google.com/d/msg/metamath/G1xyJb6RjfI/JSNXPIkVAwAJ 

As for iset.mm, there is some level of agreement that names may be better than numbers here, because:

    * there are more axioms

    * I'm not sure numbering is as well established in the constructive literature, although the only one I could find quickly which uses either number or names is https://en.wikipedia.org/wiki/Intuitionistic_logic#Hilbert-style_calculus (which uses, basically, names). Might be a little bit hard to find a lot of precedents, especially if we are looking for literature on Hilbert-style systems (as opposed to natural deduction rules).

    * numbers would presumably need to be prefixed with "i" which isn't really a problem, it just might make it slightly more awkward to use the numbers in other theorem names (things similar to the pre-rename `a9ev`).

    * one of the rationales for set.mm was "I think the axioms are going to be stable at this point, so this should be a one-time change". I have no such confidence for iset.mm. The axioms have already changed a few times and that has been with relatively few people working on iset.mm.

Since this is a question of aesthetics and clarity, there isn't a huge cost to worrying about it later. But especially if we see names rather than numbers as being the likely end state, there also isn't much cost to doing this "prematurely", while the axioms still might be evolving.

jkingdon:
I'd be OK with closing this issue as I'm not sure there is a large degree of consensus that anything needs to done soon (or interest in working on it), and if someone does pick it up they can look at a closed issue just as easily as an open one. For what it is worth, I haven't really thought much about axioms in a while; I've been busy with real number construction and other things which are several steps removed from the axioms themselves.

I guess I'll close this in a few days if no one has any further comments.

digama0:
I think names are generally better than numbers because the numbers are fragile to reordering. I guess they were inherited from set.mm and things have since changed on the set.mm side? In that case either renumbering to match or just giving them names would make sense to me.

jkingdon:
> I think names are generally better than numbers because the numbers are fragile to reordering. I guess they were inherited from set.mm and things have since changed on the set.mm side? In that case either renumbering to match or just giving them names would make sense to me.

Yeah, they were inherited and switched around a bit and then set.mm renumbered.

I would prefer names as well.

benjub:
[...]

PR #2624, opened 2022-06-17
Add dfdif3 to iset.mm and set.mm

### LINKED ISSUE #2619: Consider alternate forms of relative set complements in iset.mm
Currently iset.mm copies the definition of relative set complement from set.mm: `(𝐴 ∖ 𝐵) = {𝑥 ∣ (𝑥 ∈ 𝐴 ∧ ¬ 𝑥 ∈ 𝐵)}` ( http://us.metamath.org/ileuni/df-dif.html ). Although this is widely used enough in iset.mm that presumably we will keep it, it does fail to have many of the properties that it has in set.mm and perhaps the situation is at least a little bit analogous to http://us.metamath.org/mpeuni/df-pss.html and http://us.metamath.org/mpeuni/df-sdom.html

There is another definition given by page 10 of [1] which is `X \ Y = { α ∈ X | ∀β ∈ Y α =/= β }` . At a minimum, I suppose we should mention this in the df-dif comment but perhaps there is a source we can cite which goes into more detail (I didn't quickly find one), or theorems/definitions we can add.

[1] "Infinite Omniscient Sets in
Constructive Mathematics" by Ceel Pierik, https://www.cs.ru.nl/bachelors-theses/2020/Ceel_Pierik___4806182___Infinite_Omniscient_Sets_in_Constructive_Mathematics.pdf

jkingdon:
To be honest I don't know whether there is a reason for this definition
(perhaps in another system other than IZF?) rather than the more
straightforward df-dif , but at least for us it is equivalent.

Fixes #2619

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7748,6 +7748,12 @@ Theory,</I> McGraw-Hill, Inc. (1969) [QA248.M745].</LI>
 Predicate Logic with Identity,&quot; <I>Archiv f&uuml;r Mathematische Logik
 und Grundlagenforschung,</I> 7:103-121 (1965) [QA.A673].</LI>
 
+<LI><A NAME="Pierik"></A> [Pierik], Pierik, Ceel, "Infinite Omniscient Sets in
+Constructive Mathematics", Bachelor thesis Computing Science, Radboud University,
+June 25, 2020,
+<A HREF="https://www.cs.ru.nl/bachelors-theses/2020/Ceel_Pierik___4806182___Infinite_Omniscient_Sets_in_Constructive_Mathematics.pdf"
+>https://www.cs.ru.nl/bachelors-theses/2020/Ceel_Pierik___4806182___Infinite_Omniscient_Sets_in_Constructive_Mathematics.pdf</A></LI>
+
 <LI><A NAME="PradicBrown2021"></A> [PradicBrown2021] Pradic, Pierre, and
 Brown, Chad E. (December 8, 2021), &quot;Cantor-Bernstein implies
 Excluded Middle&quot;,


iset.mm
```diff
@@ -27243,6 +27243,24 @@ practical reasons (to avoid having to prove sethood of ` A ` in every use
...
+      ssriv exbii mpbir df-ral con2b df-ne imbi2i notbid pm5.74i 3bitr4i exintr
+      eleq1 albii bitri sylbi mpi 19.41v sylib simprd a1i dfdif2 sseqtr4i eqssi
+      ss2rabi ) CDFZAGZBGZHZBDIZACJZEVOVTEGZVOKZWACKZWAVQHZBDIZWAVTKWBWCWADKZLZ
+      WACDMZNWBWCWGWEWHWGWDBDVQDKZWGWDWIWFWAVQVQDWAOUAUBUCUDVSWEAWACAEPVRWDBDVP
+      WAVQUEUFUGUHUKVTVPDKZLZACJVOVSWKACVSWKQVPCKVSABPZBRZWKVSWLWKSBRZWMWKSVSWM
+      WNWMBAPZBRBAUIWLWOBABUJULUMVSWLWKQZBTZWMWNQVSWIVRQZBTWQVRBDUNWRWPBWIWLLZQ
+      WLWILZQWRWPWIWLUOVRWSWIVPVQUPUQWLWKWTWLWJWIVPVQDVBURUSUTVCVDWLWKBVAVEVFWL
+      WKBVGVHVIVJVNACDVKVLVM $.
+  $}
+
```
avekens: Shoud this theorem be added also to set.mm, shoudn't it?

jkingdon: Since I don't really know the reason for writing this definition this way (like, is it a thing in multiple sources or just an oddity of this one paper?), I have trouble answering this question in terms of whether this is an important result and if so where. But if the thought process is "sure, why not?" then it becomes easier.

I've pushed a change to this pull request to add it to set.mm. The iset.mm proof works with only one change related to The Great Axiom Renumbering.

jkingdon: And when I decided to minimize it that revealed a few more (unnecessary) differences between set.mm and iset.mm, so I fixed those up, which is the latest change I've pushed to this pull request.

Hopefully this isn't scope-creeping too badly.

avekens (approved):

wlammen:
> A commuted form of ~ a9ev .

Shouldn't this be ~ a6ev? Or maybe the theorem name ax6evr should change? There is not much in iset.mm with an ax6 prefix.  ax6b, but that has an alias hbn1.

jkingdon:
> Shouldn't this be ~ a6ev? Or maybe the theorem name ax6evr should change?

The situation is explained in a comment added in this pull request and there is also an issue #756 on the topic.

None of the possible solutions are perfect or easy but I think this is mostly for another day.

PR #2626, opened 2022-06-17
 Ordered pair definition of Felix Hausdorff and misc.

avekens:
General:
* proof of ~suppfnss shortened by using (the recently added theorem) ~rabssrabd.
Remark 1: the minimize script (executed for ` *rab* `) did not find any shorter proofs, but the proof of ~suppfnss could have been minimizes manually.
Remark 2: ~suppfnss could be shortened further if hypothesis 2 in rabssrabd is changed to ` ( ( ph /\ x e. A ) -> ( ps -> ch ) ) ` - but this would make the proofs of ~clwlknon2num and ~numclwlk1lem2 longer.

Singletons:
* comment of ~df-sn improved.
* alternate definition of singleton ~dfsn2ALT added, based on the (alternate!) definition of unordered pair.
Remark: The present alternate definition of singleton ~dfsn2 is actually not an alternate definition, because it is based on the definition ~df-pr of unordered pairs which itself uses the present definition ~df-sn of a singleton! Therefore, ~dfsn2 should be renamed (maybe ~pridsn). 

Unordered pairs:
* ~prel12 and ~prel12g revised: the hypotheses ` C e. _V ` and ` D e. _V ` are not needed.
* ~preqsn and ~preqsnd revised: the hypothesis ` C e. _V ` is not needed
=> proofs of ~hash2prde, ~opeqsn, ~propeqop, ~propssopi, ~relop, ~symg2bas shortened
* new theorems ~prnesn, ~prneprprc, ~preq12nebg added

Ordered pairs:
* ~opthprneg added
* Justifications of the ordered pair definition of Felix Hausdorff: 2 variants ~opthhausdorff and ~opthhausdorff0
* comment of ~df-op enhanced

Axioms of Regularity and Choice:
* theorems ~elneq, ~elnotel, ~elnel added
* ~preleq revised (most of the hypotheses are not needed), see also ~preleqg and ~preleqALT
* proof of ~opthreg shortened
* "left-to-right" implication extracted from ~dfac2 (as ~dfac2b)
Remark: I think this was a former situation, because many comments refer to this. Since both directions are so different (especially regarding the dependency on axioms), I think it is clearer to have a theorem for each direction and then to combine them to ~dfac by applying ~impbii.
The comments are adjusted accordingly (this was still a TODO in the comment of ~dfac2).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -11650,6 +11650,10 @@
 "prcdnq" is used by "psslinpr".
 "prcdnq" is used by "reclem2pr".
 "prcdnq" is used by "suplem1pr".
+"prel12OLD" is used by "dfac2OLD".
+"prel12OLD" is used by "prel12gOLD".
+"preleqOLD" is used by "dfac2OLD".
+"preleqOLD" is used by "opthregOLD".
 "prlem934" is used by "ltaddpr".
 "prlem934" is used by "ltexprlem7".
 "prlem934" is used by "prlem936".
@@ -15091,6 +15095,7 @@ New usage of "df-vhc3" is discouraged (1 uses).
 New usage of "df-vs" is discouraged (1 uses).
 New usage of "df-wrecs" is discouraged (11 uses).
 New usage of "df0op2" is discouraged (3 uses).
+New usage of "dfac2OLD" is discouraged (0 uses).
 New usage of "dfadj2" is discouraged (7 uses).
 New usage of "dfbi1ALT" is discouraged (0 uses).
 New usage of "dfbi3OLD" is discouraged (0 uses).
@@ -15104,6 +15109,7 @@ New usage of "dfiota4OLD" is discouraged (0 uses).
 New usage of "dfnfc2OLD" is discouraged (0 uses).
 New usage of "dfpjop" is discouraged (4 uses).
 New usage of "dfpleOLD" is discouraged (2 uses).
+New usage of "dfsn2ALT" is discouraged (0 uses).
 New usage of "dfss1OLD" is discouraged (0 uses).
 New usage of "dfss5OLD" is discouraged (0 uses).
 New usage of "dftru2" is discouraged (0 uses).
@@ -17202,6 +17208,7 @@ New usage of "opsqrlem4" is discouraged (2 uses).
 New usage of "opsqrlem5" is discouraged (1 uses).
 New usage of "opsqrlem6" is discouraged (0 uses).
 New usage of "opsrbaslemOLD" is discouraged (0 uses).
+New usage of "opthregOLD" is discouraged (0 uses).
 New usage of "orbi1r" is discouraged (0 uses).
 New usage of "orbi1rVD" is discouraged (0 uses).
 New usage of "ordelinelOLD" is discouraged (0 uses).
@@ -17489,7 +17496,13 @@ New usage of "poml4N" is discouraged (3 uses).
 New usage of "poml5N" is discouraged (1 uses).
 New usage of "poml6N" is discouraged (1 uses).
 New usage of "prcdnq" is discouraged (14 uses).
+New usage of "prel12OLD" is discouraged (2 uses).
+New usage of "prel12gOLD" is discouraged (0 uses).
+New usage of "preleqALT" is discouraged (0 uses).
+New usage of "preleqOLD" is discouraged (2 uses).
 New usage of "preqsnOLD" is discouraged (0 uses).
+New usage of "preqsnOLDOLD" is discouraged (0 uses).
+New usage of "preqsndOLD" is discouraged (0 uses).
 New usage of "prlem934" is discouraged (3 uses).
 New usage of "prlem936" is discouraged (1 uses).
 New usage of "prmgaplcm" is discouraged (0 uses).
@@ -18031,6 +18044,7 @@ New usage of "superpos" is discouraged (1 uses).
 New usage of "supexpr" is discouraged (1 uses).
 New usage of "suplem1pr" is discouraged (1 uses).
 New usage of "suplem2pr" is discouraged (1 uses).
+New usage of "suppfnssOLD" is discouraged (0 uses).
 New usage of "supsr" is discouraged (1 uses).
 New usage of "supsrlem" is discouraged (1 uses).
 New usage of "syl2an2rOLD" is discouraged (0 uses).
@@ -18900,6 +18914,7 @@ Proof modification of "decsubiOLD" is discouraged (76 steps).
 Proof modification of "decsucOLD" is discouraged (41 steps).
 Proof modification of "decsuccOLD" is discouraged (44 steps).
 Proof modification of "demoivreALT" is discouraged (1087 steps).
+Proof modification of "dfac2OLD" is discouraged (822 steps).
 Proof modification of "dfbi1ALT" is discouraged (100 steps).
 Proof modification of "dfbi3OLD" is discouraged (45 steps).
 Proof modification of "dfcleq" is discouraged (10 steps).
@@ -18908,6 +18923,7 @@ Proof modification of "dfdp2OLD" is discouraged (37 steps).
 Proof modification of "dfiota4OLD" is discouraged (40 steps).
 Proof modification of "dfnfc2OLD" is discouraged (116 steps).
 Proof modification of "dfpleOLD" is discouraged (44 steps).
+Proof modification of "dfsn2ALT" is discouraged (30 steps).
 Proof modification of "dfss1OLD" is discouraged (24 steps).
 Proof modification of "dfss5OLD" is discouraged (18 steps).
 Proof modification of "dfvd1imp" is discouraged (10 steps).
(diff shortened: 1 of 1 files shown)


avekens:
I think at least @digama0 and @benjub should have a look at this PR, but there is no hurry to merge it - I can continue with other stuff on a separate Git branch.

set.mm
```diff
@@ -50034,14 +50033,22 @@ we define it to be empty in this case (see ~ opprc1 , ~ opprc2 , and
...
-       p. 281.
+       ~ opthprc .  Nearly at the same time as Norbert Wiener, Felix Hausdorff
+       proposed the following definition in "Grundz&uuml;ge der Mengenlehre"
+       ("Basics of Set Theory"), p. 32, in 1914: ` <. A , B >. ` _5
+       ` = { { A , O } , { B , T } } ` .  Hausdorff used ` 1 ` and ` 2 `
+       instead of ` O ` and ` T ` , but actually any two different fixed sets
+       will do (e.g., ` O = (/) ` and ` T = { (/) } ` , see ~ 0nep0 ).
+       Furthermore, Hausdorff demanded that ` O ` and ` T ` are both different
+       from ` A ` as well as ` B ` , which is actually not necessary (at least
+       not in full extend), see ~ opthhausdorff0 and ~ opthhausdorff .  If we
```
benjub, line 50044: extend --> extent
(but maybe better: "at least not for most applications")

OK. I haven't checked the original sources, but it sounds indeed like one of the historically proposed alternatives for the ordered pair definition (among yet others). On the other hand, set.mm is not meant as a history reference, so this one is fine, but no need to put too many of these in my opinion.

set.mm
```diff
@@ -98936,6 +99138,16 @@ is that it denies the existence of a set containing itself ( ~ elirrv ).
...
     ACDZBCDZEZABDZBADZEZFZCGHRUBCICABJKUARSPTQABLBALMNO $.
 
+  $( A class cannot be an element of one of its elements.  (Contributed by AV,
+     14-Jun-2022.) $)
+  elnotel $p |- ( A e. B -> -. B e. A ) $=
+    ( wcel en2lp imnani ) ABCBACABDE $.
+
+  $( A class cannot be an element of one of its elements.  (Contributed by AV,
+     14-Jun-2022.) $)
+  elnel $p |- ( A e. B -> B e/ A ) $=
```
benjub: necessary ? elnotel probably suffices

set.mm
```diff
@@ -98897,6 +99094,11 @@ is that it denies the existence of a set containing itself ( ~ elirrv ).
       CZODZEPBFZQCZDPBAAQAGZROSQAQASHZTIJBKLOMN $.
   $}
 
+  $( A class is not equal to any of its elements.  (Contributed by AV,
+     14-Jun-2022.) $)
+  elneq $p |- ( A e. B -> A =/= B ) $=
```
benjub: do we have in set.mm `|- -. (A e. B /\ A = B )` ?
If not, it would be nice to add it (mabye with the variable-free version `( _E i^i _I ) = (/)`).

set.mm
```diff
@@ -106076,59 +106340,118 @@ is defined as the Axiom of Choice (first form) of [Enderton] p. 49.  The
...
+       AC, but the Axiom of Regularity is used (by applying ~ dfac2b ).
+       (Contributed by NM, 5-Apr-2004.)  (Revised by Mario Carneiro,
+       26-Jun-2015.)  (Revised by AV, 16-Jun-2022.) $)
     dfac2 $p |- ( CHOICE <-> A. x E. y A. z e. x ( z =/= (/) ->
         E! w e. z E. v e. y ( z e. v /\ w e. v ) ) ) $=
+      ( wac cv c0 wne wel wa wrex wreu wi wral wex wal dfac2b dfac2a impbii ) F
+      CGZHICEJDEJKEBGLDUAMNCAGOBPAQABCDERABCDEST $.
+
+    $( Obsolete proof of ~ dfac2 as of 16-Jun-2022.  Axiom of Choice (first
+       form) of [Enderton] p. 49 implies of our Axiom of Choice (in the form of
```
benjub, line 106406: implies of --> implies

benjub:
Just a few remarks. That's globally fine. Indeed an opinion from @digama0 would be nice.

set.mm
avekens: OK, typo will be corrected.

Of course not *all* historic variants should be discussed (and proven) in set.mm, but the important or interesing ones should be.

set.mm
avekens: This is the old philosophical discussion: ~elnotel is useful for proofs, but ~elnel has a better looking representation.

set.mm
avekens: Not as far as I know. I agree, it would be nice to have this theorem in set.mm, analogously to ~en2lp. Also the variable-free version looks interesting (is there any analog for ~en2lp)? I try to add both of them...

set.mm
avekens: This typo (and it's not only a typo, but also not semantically correct) was contained in the original comment. I kept it as "proof" that some revision was necessary here...
It will vanish together with the whole theorem in about a year ;-).

set.mm
avekens: The variable-free version of ~en2lp is ` ``' _E =/= _E ` (The membership (epsilon) relation is asymmetric.), or with the definition in Peter's mathbox: ` -. SymRel _E `. I'll add these theorems, too.

benjub: Rather ```(`' _E i^i E ) = (/) ```

set.mm
benjub: An important variant, which also works for proper classes, is defined in my mathbox (section "Tuples of classes").

sctfn:
The definition in nf.mm also carries through in ZF (without infinity!)

set.mm
avekens: It's a shame that I do not know all the treasures which are buried in the mathboxes... Yes, I agree that this a more important variant (especially the variant ~bj-2uplth of ~opth is very interesting), and I would have proposed to move it to main, too, unless there were a lot of new definitions involved (for my Hausdorff variant, no new definition is needed - its just a justification).

avekens:
> The definition in nf.mm also carries through in ZF (without infinity!)
@sctfn you mean ~df-op, the definition of an ordered pair according to Rossner, in nf.mm (in contrast to ~df-opk, the Kuratowski definition of an ordered pair)? There is a hint to it in the comment of df-op in set.mm ("a common alternate definition in New Foundations is the definition from [Rosser] p. 281."), but I do not see how this can be (easily) expressed in the terms of set.mm.

set.mm
avekens: Typo fixed.

set.mm
```diff
@@ -558947,6 +559324,18 @@ We use this concept to define functions ( ~~? df-funsALTV , ~~?
...
+  $( The identity relation is symmetric.  (Contributed by AV, 19-Jun-2022.) $)
+  idsym $p |- SymRel _I $=
+    ( cid wsymrel ccnv wceq wrel cnvi reli dfsymrel4 mpbir2an ) ABACADAEFGAHI
+    $.
+
+  $( The membership (epsilon) relation is not symmetric.  (Contributed by AV,
+     18-Jun-2022.) $)
+  epnsym $p |- -. SymRel _E $=
+    ( cep wsymrel ccnv wceq wrel wa epasym neii intnanr dfsymrel4 mtbir ) ABACZ
+    ADZAEZFMNLAGHIAJK $.
```
avekens, line 559336: @mazsa I hope you don't mind that I have put these two theorems into your mathbox.

set.mm
avekens: Done.

set.mm
mazsa: @avekens of course not: feel free to use my Mathbox any time. (I'll continue to submit my theorems in July. This year:)

benjub:
After a quick look, it looks like the definitions that @sctfn formalized in nf.mm and that I formalized in set.mm/mathbox both come from Quine's idea to "disjointify" both components. A nice review is in the introduction of the paper I mentioned in the section comment "Tuples of classes", written by McCarty and Scott (I have the pdf if you want), and there is also the survey by Kanamori, which is available on his website.

set.mm
```diff
@@ -99138,6 +99172,25 @@ is that it denies the existence of a set containing itself ( ~ elirrv ).
...
+    cnvepnep $p |- ( `' _E i^i _E ) = (/) $=
+      ( vy vx cep ccnv cin wel wa copab c0 df-eprel cnveqi cnvopab eqtri inopab
+      ineq12i wceq wn wal cv en2lp gen2 opab0 mpbir ) CDZCEZABFZBAFZGZBAHZIUEUF
+      BAHZUGBAHZEUIUDUJCUKUDUFABHZDUJCULABJKUFABLMBAJOUFUGBANMUIIPUHQZARBRUMBAA
+      SBSTUAUHBAUBUCM $.
+
+    $( The membership (epsilon) relation is asymmetric.  (Contributed by AV,
+       18-Jun-2022.) $)
+    epasym $p |- `' _E =/= _E $=
+      ( cep ccnv cin c0 wceq wa wb wne cnvepnep disjeq0 wi epn0 eqneqall adantl
```
benjub, line 99189: "asymmetric" is the property proved in ~cnvepnep.

~epasym simply is "not symmetric" (which is not a very important property)

avekens: Comments and labels adjusted.

benjub (approved):

benjub:
I'll wait a day or two in case @digama0 has a comment, and if nothing new comes, I'll merge this weekend.

PR #2627, opened 2022-06-18
Various undif related additions in iset.mm

### LINKED ISSUE #2618: More undif theorems for iset.mm
I'll probably get these done pretty soon, but might as well make this an issue just to describe what is in progress.

* full undif implies excluded middle
* `( ( A \ B ) u. B ) = A <-> A. x e. A DECID x e. B` as suggested at https://github.com/metamath/set.mm/pull/2617#discussion_r898785273
* If `A` has decidable equality and `B` is finite, `( ( A \ B ) u. B ) = A` (the proof of `undifom` in https://github.com/metamath/set.mm/pull/2617 should prove this with few changes)

jkingdon:
The summary is in #2618 and the even shorter summary is that this adds a few theorems to iset.mm related to http://us.metamath.org/mpeuni/undif.html (which is not provable in iset.mm, as shown here).

Fixes #2618 .

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1488,7 +1488,7 @@ is double negation elimination.</TD>
 </TR>
 
 <TR>
-  <TD ROWSPAN="4">undif</TD>
+  <TD ROWSPAN="5">undif</TD>
   <TD>~ undifss</TD>
   <TD>subset rather than equality, for any sets</TD>
 </TR>
@@ -1501,6 +1501,11 @@ is double negation elimination.</TD>
 <TR>
   <TD>~ undifdc</TD>
   <TD>where the container has decidable equality and the subset is finite</TD>
+</TR>
+
+<TR>
+  <TD>~ undifdcss</TD>
+  <TD>when membership in the subset is decidable</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1488,7 +1488,7 @@ is double negation elimination.</TD>
 </TR>
 
 <TR>
-  <TD ROWSPAN="5">undif</TD>
+  <TD ROWSPAN="4">undif</TD>
   <TD>~ undifss</TD>
   <TD>subset rather than equality, for any sets</TD>
 </TR>
@@ -1501,11 +1501,6 @@ is double negation elimination.</TD>
 <TR>
   <TD>~ undifdc</TD>
   <TD>where the container has decidable equality and the subset is finite</TD>
-</TR>
-
-<TR>
-  <TD>~ undifom</TD>
-  <TD>where the container is ` _om ` and the subset is finite</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


avekens (approved):

iset.mm
```diff
@@ -34948,6 +34948,21 @@ than an implication (but the two are equivalent by ~ bm1.3ii ).
       ONFOAHOMBIJKL $.
   $}
 
+  ${
+    $d ph x y z $.
+    undifexmid.1 $e |- ( x C_ y <-> ( x u. ( y \ x ) ) = y ) $.
+    $( If the union of complementary parts always produces the whole, excluded
+       middle follows.  (Contributed by Jim Kingdon, 16-Jun-2022.) $)
+    undifexmid $p |- ( ph \/ -. ph ) $=
```
benjub, line 34956: The statement doesn't match the comment. Are you using the biconditional in the proof, or only the right implication ? 

**Edit:** the left implication is trivial, so you only need the right implication. Can you modify to
```
undifexmid.1 $e |- ( x C_ y -> ( x u. ( y \ x ) ) = y ) $.
```
?
Or maybe
```
undifexmid.1 $e |- y C_ ( x u. ( y \ x ) ) $.
```
?

iset.mm
```diff
@@ -58537,34 +58552,34 @@ ordinals implies excluded middle ( ~ ordsucunielexmid ).  (Contributed
     NUKBRZUMULSUJUQUIBTUAUQUMULBAUBUCUDUNULSUKABUEUFUGUH $.
 
   ${
-    $d A x $.  $d B x $.
+    $d A x y z $.  $d B x y z $.
+    $( If we remove a single element from a set with decidable equality then
+       put it back in, we end up with the original natural number.  This
```
benjub, line 58557: ...with the original set

iset.mm
jkingdon: Seems to me that the hypothesis should exactly be http://us.metamath.org/mpeuni/undif.html (with class changed to set). If so, it is the comment which needs adjusting (the word "always" in particular but maybe other bits).

jkingdon: How does

```
Union of complementary parts producing the whole and excluded middle.
Although special cases such as ~ undifss and ~ undifdcss are provable,
the full statement implies excluded middle as shown here.
```

look?

iset.mm
jkingdon: Ooh good catch (copy paste error), thanks.

I've pushed up this change.

iset.mm
benjub: I understand your argument.
Note that my proposed hyothesis can be read: "every subset is complemented", which also has its charm.
As you prefer.
And after that, you can merge.

jkingdon: I'll merge this for now. If we find we miss the version phrased in terms of `y C_ ( x u. ( y \ x ) )` we can always revisit (maybe by then we'll have `EXMID` if that comes to pass).

Issue #1887, opened 2020-11-22
Organize the metamath/set.mm folder ?


benjub:
With the recent addition of about thirty files in the repository, it's getting a bit long. Is it possible to organize it, for instance by putting the files *frege\*.svg (or all the *.svg files) in a subdirectory ?

avekens:
I would appreciate that. With all these *.svg-files, the folder is really confusing, and these files seem to be more prominent than they actually are.

nmegill:
I'm fine with creating a directory called "svg" and moving \*.svg to it.  Either of you (or anyone else) may go ahead and create a PR for that.   (The .svg creators @arpie-steele and @tirix should be made aware of this change so they know where their files went, but no action is required on their part.)

BTW the set.mm directory structure on GitHub is unrelated to the directory structure of metamath.org, so this has no effect on the web site.

david-a-wheeler:
> On Dec 9, 2020, at 9:21 AM, nmegill <notifications@github.com> wrote:
> 
> I'm fine with creating a directory called "svg" and moving *.svg to it. Either of you (or anyone else) may go ahead and create a PR for that. (The .svg creators @arpie-steele <https://github.com/arpie-steele> and @tirix <https://github.com/tirix> should be made aware of this change so they know where their files went, but no action is required on their part.)

.svg is a file extension/file format, not a *purpose*. I suggest that instead of “svg/“ using a directory name that expresses its purpose, e.g., “images/“.

--- David A. Wheeler

digama0:
I would suggest `images/svg/` then, since we also have a lot of gif images in the metamath web site (which may or may not live here, but just in case we decide to move them here).

nmegill:
I would favor David's images/ with no subdirectories.  The .svg files are easily identified with 'ls \*.svg'.  Having to cd to another lower level is extra step after you go into images and determine that images/ contains svg/.  (Of course there's 'find' or 'ls -R' on Linux but I tend to start with just a simple 'ls'.)

As for the GIF math symbol images, my long-term goal is to make mpegif obsolete.  I understand that people mainly use it for ASCII copy/paste, and if Igor's pages end up supporting that, we can get rid of mpegif.  Currently I don't have plans to put them on GitHub.

If we start with just images/ and later determine images/svg/ would make more sense, we can always create it and move the .svg's there.  If we start with images/svg, human nature is such that the svg/ subdirectory will probably never go away. :)

[...]

PR #2628, opened 2022-06-18
Changes file

jkingdon:
This moves about 10,000 lines out of set.mm and hopefully is an easier place to manage this content. My goal was to make an incremental step which addresses issues such as having to scroll down a lot when looking at set.mm from the top of the file, rather than get too deeply into things like what changes should be listed here versus just in the git history.

The division into two commits is to make it easier to review - one commit just moves a lot of lines, the other has the other changes (updating text and the like).

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -3,7 +3,10 @@
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 
 This is part of an ongoing project to improve naming consistency.  If you have
-suggestions for better names, let me know.
+suggestions for better names, let us know. It lists changes in set.mm (for
+iset.mm, see the git history, although many of the proposed or historical
+changes described here also apply to iset.mm, or should be adopted in iset.mm
+when someone gets the chance).
 
 To update your mathbox, you can make global substitutions into
 your local version by processing the ones WITHOUT "Notes" in REVERSE order.
@@ -19,7 +22,8 @@ proposed  syl5*     *trid       (syl5bi -> biimtrid; syl5eqel -> eqeltrid;etc.)
 proposed  syl6*     *trdi
 proposed  *mpt*     *mpf* or *mptf* (maps-to for functions)
 proposed  *mpt2*    *mpo* or *mpto* (maps-to for operations)
-(Please send any comments on these proposals to Norm Megill.)
+(Please send any comments on these proposals to the mailing list or
+make a github issue.)
 
 DONE:
 Date      Old       New         Notes
@@ -12411,3 +12415,5 @@ Date      Old       New         Notes
  8-Feb-04 zfnull    0ex
  8-Feb-04 limelon   [--same--]  changed first -> to /\ in antecedent
 
+For older changes, the git history has some old archives checked in
+going back to 1998.
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -1,0 +1,12413 @@
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  Recent label changes
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+
+This is part of an ongoing project to improve naming consistency.  If you have
+suggestions for better names, let me know.
+
+To update your mathbox, you can make global substitutions into
+your local version by processing the ones WITHOUT "Notes" in REVERSE order.
+The ones WITH "Notes" may have to be processed manually.
+
+PROPOSED:
+Date      Old       New         Notes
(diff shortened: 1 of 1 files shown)


benjub:
So it settles one of the heated discussions of the mailing list. I think it's a good change.
I would not capitalize "Changes.txt" (you would also need to change the reference to it in set.mm)

jkingdon:
> So it settles one of the heated discussions of the mailing list. I think it's a good change.

Nice to hear at least one voice that this is a good step forward.

> I would not capitalize "Changes.txt"

Tradition is for things like this to be in all caps so they sort first in a case-sensitive sort (`README`, `CONTRIBUTING`, etc). An initial capital is sort of a compromise between the "we like lowercase, all caps feels like shouting" sentiment and this tradition, but I don't have a super-strong opinion about the name.

benjub:
Then maybe all caps ? The mix seems strange, and one is likely to misremember or miss the capital C... (no strong opinion either; if that naming scheme is also well spread, then so be it)

jkingdon:
> Then maybe all caps ? The mix seems strange, and one is likely to misremember or miss the capital C... (no strong opinion either; if that naming scheme is also well spread, then so be it)

It is always hard to know which precedents "count" - I mean there is `Makefile`, `Gemfile`, and `Rakefile`, etc, or the capitalized directories (`Documents`, `Downloads`, etc) of Macintosh, some linux distributions, and probably others. But in terms of changes specifically, an internet search for "changes.txt" shows a lot of `changes.txt` and `CHANGES.txt` and some `Changes.txt` (for example http://www.qhull.org/ ). So if we do choose the latter we won't be the first.

avekens:
This issue was discussed in https://groups.google.com/g/metamath/c/1Lh763NZpDA/m/cQw17LLWBwAJ, and it contains some of the very last comments of Norm. Although he was not against extracting the list from set.mm in general, he gave same arguments to keep it, at least the ones of recent date. The main argument was (and still is) maintainability - how is it ensured that the list is maintained thoroughly (in either way)? And is the maintenance effort justified by the use of such a list? Should the file Changes.txt maintained in the same way as the list currently maintained in set.mm? Which changes must and which changes can be documented and how? Is there a way to automatically check this file against the actual changes in set.mm (as it is done with the discouraged file)? 

From my point of view, there is no need for such a list - there are sufficent ways to track the history within Git. Therefore I would the extraction of this list from set.mm. Some other main contributors, however, should agree with it, too.

jkingdon:
> Should the file Changes.txt maintained in the same way as the list currently maintained in set.mm?

My intended answer in the context of this pull request is "yes, same list with same rules". There are a lot of issues raised in that mailing list thread and many of them are good points, but my idea was that this pull request could be a small step which doesn't prejudice any of the larger issues around this list.

I mean, at a minimum we should change the "contact Norm" language, and if people really think a super-minimal fix is all we want to do now, fine, but I guess the pull request here was my best guess about something which was big enough to make things better, but small enough to avoid us getting lost in too many choices and not enough consensus. My reading of the mailing list thread is that there isn't complete unanimity but that there seems to be at least a moderate amount of sentiment in favor of "keep the list but in its own file".

david-a-wheeler:
I think moving this information into a separate file is a good idea. Should it be called "changes-set.txt" or similar (since it's really only the changes in set.mm)?  I'm not sure it needs to be initial-upper-case, it might be better as a lowercase filename.

jkingdon:
> I think moving this information into a separate file is a good idea. Should it be called "changes-set.txt" or similar (since it's really only the changes in set.mm)? I'm not sure it needs to be initial-upper-case, it might be better as a lowercase filename.

Seems like lowercase is the popular option so far. So I guess I'll propose `changes-set.txt` unless there are other suggestions.

benjub:
Another possibility is to have (without blocking this PR, so can be done later):
```
...
databases
 ├── set
 │  ├─ set.mm
 │  ├─ discouraged
 │  └─ changes.txt
 ├── iset
 │  ├─ iset.mm
 │  ├─ discouraged
 │  └─ changes.txt
 ├─
...
```
and more generally the organization discussed in Issue #1887.

jkingdon:
I've changed the name to `changes-set.txt` and based on comments in this pull request we have:

* @benjub -  "I think it's a good change."
* @david-a-wheeler - "moving this information into a separate file is a good idea."
* @avekens - "there is no need for such a list - there are sufficent ways to track the history within Git. Therefore I would [verb seems to be missing here] the extraction of this list from set.mm."

Anyone else want to weigh in?

wlammen:
Has anyone ever used it (except updating it)? Certainly not me.

I think the answer to this question can hint towards how to maintain the list in future.  Since this PR is reversible, I don't hereby object to a  merge.

Maybe a "Recent changes" with a timeout of, say 12 months, for individual entries, is a decent replacement.  The 12 months are in accordance with the timeout for OLD theorems.  This has the charm that the change list can be automatically updated by a script, say the one that generates discouraged.  Again, this is a suggestion that can be introduced later.

avekens:
>     * @avekens - "there is no need for such a list - there are sufficent ways to track the history within Git. Therefore I would [verb seems to be missing here] the extraction of this list from set.mm."

The missing verb is "accept".

avekens (approved):

jkingdon:
> The missing verb is "accept".

I like that verb 😉 . Since there seems to be little dissent, I'll merge this now and we can leave further fine tuning or changes for future pull requests.

Issue #2620, opened 2022-06-16
define notation for omniscience in iset.mm


jkingdon:
As suggested at https://github.com/metamath/set.mm/issues/2595#issuecomment-1157367557 we should define

```
A e. Omni <-> A. f : A --> 2o ( E. x e. A ( f ` x ) = (/) \/ A. x e. A ( f ` x ) = 1o )
```

This matches the definition of '''omniscient set''' at Definition 3.1 of [Pierik], p. 14.

We can also prove:
* `A e. Fin -> A e. Omni`
* `A. x x e. Omni` implies excluded middle (see "If you take it to be the subsingleton corresponding to a given truth value and apply this principle to the constantly true proposition, then EM follows" at https://ncatlab.org/nlab/show/principle+of+omniscience )
* Perhaps prove in set.mm that ``A. z A. f : z --> 2o ( E. x e. z ( f ` x ) = (/) \/ A. x e. z ( f ` x ) = 1o )`` (no need for a specific notation for `Omni` here, I suppose)

Mention in a comment that `_om e. Omni` is the Limited Principle of Omniscience (LPO).

[Pierik] "Infinite Omniscient Sets in
Constructive Mathematics" by Ceel Pierik, https://www.cs.ru.nl/bachelors-theses/2020/Ceel_Pierik___4806182___Infinite_Omniscient_Sets_in_Constructive_Mathematics.pdf

More background reading: https://ncatlab.org/nlab/show/principle+of+omniscience

digama0:
By the way, I think it would also be useful to have a definition for EM. We can't define it in pure propositional logic because we need to quantify over propositions, but we can once we get to set theory, by quantifying over subsets of `1o` . set.mm has a definition for `CHOICE` and I think it will make it easier to do similar things like proving implications between weak EM principles just like set.mm does for choice principles.

The purpose of the definitions is not really to make the derivations shorter, although I can imagine something like that for `Omni` since there are a lot of properties that converge on this notion; rather it is intended to give a vocabulary for things that look unmotivated otherwise.

benjub:
Something like
```
df-em $a |- ( EM <-> A . x ( x C_ 1o -> ( x = (/) \/ x = 1o ) ) ) $. 
```
?

digama0:
Yep. That could also be written as `A. x ( x C_ 1o -> DECID (/) e. x )`, or even just `A. x DECID (/) e. x` or `A. x A. y DECID x e. y`.

[...]

Issue #739, opened 2019-01-27
Prove ordinal trichotomy in iset.mm given a suitable decidability hypothesis


jkingdon:
In http://us.metamath.org/ileuni/ordtriexmid.html we show that `∀x ∈ On ∀y ∈ On (x ∈ y ∨ x = y ∨ y ∈ x)` implies Excluded Middle.

We should be able to go the other direction too. That is, we'd adopt a suitable formalization of the Law of the Excluded Middle (We can't just put "`ph \/ -. ph` for an arbitrary proposition" as a hypothesis, but we can adopt the case we need, perhaps in the form `DECID _something_`), and then show ordinal trichotomy follows.

This would be an example of following the program of Anderson, Belnap, and Dunn when they suggest "adding the appropriate excluded middles as premisses" ( https://github.com/metamath/set.mm/pull/737#issuecomment-457891897 ).

As for how this proof would go, proving it the way that set.mm does might be the hard way (given how many theorems and definitions build up to produce ordinal trichotomy in set.mm), but I haven't thought too hard about the easiest way to write this proof (including what lemmas could be proved without excluded middle).

jkingdon:
So I did a little reading, and don't see a lot of easy shortcuts (in which we prove, or already have proved, a lot of the supporting theorems in IZF and then just add a bit of EXMID at the end).

The set.mm proof is via https://us.metamath.org/mpeuni/tz7.7.html and https://us.metamath.org/mpeuni/tz7.5.html and depends on `We` being fairly well developed. I assume that https://us.metamath.org/ileuni/df-wetr.html is not sufficient for this purpose (in fact, based on discussions at #2212 especially the "rather bleak" comment, I'm not sure we have yet established that it is sufficient for any purpose). https://us.metamath.org/mpeuni/df-we.html (and some other definitions of well order) seem to start with trichotomy so I think I'd need to dig into this further to understand what is going on here and what is shown to be well ordered how.

The proof at https://proofwiki.org/wiki/Ordinal_Membership_is_Trichotomy is also based on well orderings.

The proof at https://builds.openlogicproject.org/content/set-theory/ordinals/basic.pdf relies on https://builds.openlogicproject.org/content/set-theory/ordinals/basic.pdf#thm.0.3 which has a "as α is an ordinal, it has some ∈-least element which is ..." step.

[...]

PR #2629, opened 2022-06-19
EXMID notation for iset.mm

jkingdon:
Similar to http://us.metamath.org/mpeuni/df-ac.html in set.mm this will make it easier to express results like http://us.metamath.org/ileuni/ordtriexmid.html and future ones like #739 .

Most of the existing discussion was in comments at #2620 (per our longstanding habit of digressing on issues rather than making new ones).

The results are:
* definition of EXMID as `A. x ( x C_ { (/) } -> DECID (/) e. x`
* `EXMID -> DECID ph`
* `EXMID <-> A . x ( x C_ 1o -> ( x = (/) \/ x = 1o ) )`
* `EXMID <-> A. x DECID (/) e. x`
* `EXMID <-> A. x A. y DECID x e. y`
* showing that undif is equivalent to excluded middle. Intended as a proof of concept that with these definitions we can show that statements imply excluded middle. Turned out to be more of a first step than a definitive demonstration of the general case, but seems like a good place to pause for now and submit this pull request.

Probably best to wait for convenience theorems until a bit later: 
* `EXMID <-> A. x ( x C_ 1o -> ( (/) e. x \/ -. (/) e. x ) )`
* `EXMID <-> A. x ( (/) e. x \/ -. (/) e. x )`


jkingdon:
This pull request now includes the theorems we talked about before and is ready for review.

avekens:
I am missing some examples for applications/usage of this definition. How, for example, could the new theorem ~undifexmid provided with #2627 could be expressed by using EXMID? Or did I misunderstand something completely?

jkingdon:
> I am missing some examples for applications/usage of this definition. How, for example, could the new theorem ~undifexmid provided with #2627 could be expressed by using EXMID? Or did I misunderstand something completely?

You lucked out - that is the one which is a proof of concept in this pull request:

```
    exmidundif $p |- ( EXMID
        <-> A. x A. y ( x C_ y <-> ( x u. ( y \ x ) ) = y ) ) $=
```

In the direction of the undif proposition implying `EXMID` I'm not sure it has a huge amount of difference from undifexmid . Either way you can use it in another proof in order to prove excluded middle via this theorem, both show that a given proposition implies excluded middle, etc.

But in the direction of EXMID implying the undif proposition there's no good way to do that in the undifexmid style because you can't write "ph \\/ -. ph for an arbitrary proposition ph". I mean, you can put in the specific proposition and try to explain it in a comment, but that's sort of clunky and makes it hard to see the difference between the statement being proved and how it is being proved. As long ago as #739 we had a desire to do this, but hadn't figured out the mechanism. I'm pretty sure `EXMID` solves this problem.

I think that distinction between something implying EXMID versus EXMID implying something also applies to more complicated formulas where EXMID is combined with other formulas such as `_om e. Omni` (from #2620 ) but I might be going too deep into the future which isn't needed to understand the present.

avekens:
> > I am missing some examples for applications/usage of this definition. How, for example, could the new theorem ~undifexmid provided with #2627 could be expressed by using EXMID? Or did I misunderstand something completely?
> 
> You lucked out - that is the one which is a proof of concept in this pull request:
> 
> ```
>     exmidundif $p |- ( EXMID
>         <-> A. x A. y ( x C_ y <-> ( x u. ( y \ x ) ) = y ) ) $=
> ```
> 
> In the direction of the undif proposition implying `EXMID` I'm not sure it has a huge amount of difference from undifexmid . Either way you can use it in another proof in order to prove excluded middle via this theorem, both show that a given proposition implies excluded middle, etc.

OK, great! Could this be explained, however, somehow in the comments of ~exmidundif and/or ~undifexmid, referencing each other? Or does ~undifexmid becomes obsolete? Is (or can be)  ~undifexmid part of the proof of ~exmidundif ?

jkingdon:
> Could this be explained, however, somehow in the comments of ~exmidundif and/or ~undifexmid, referencing each other?

I pushed up a paragraph in `df-exmid` which I suppose is the place for it as it would be true of any such pair and is a generic issue not specific to `undif`.

> Or does ~undifexmid becomes obsolete? Is (or can be) ~undifexmid part of the proof of ~exmidundif ?

undifexmid could be proved in terms of exmidundif or omitted but I'm not sure we want to. The proof of undifexmid and exmidundif read pretty differently and aren't especially similar in terms of how they work. For what it is worth, undifexmid style proofs are mostly what I have seen in places like https://plato.stanford.edu/entries/set-theory-constructive/incompatible-set-axs.html or the [Bauer] reference. The new paragraph I pushed up in the df-exmid comment briefly touches on this.

Perhaps this will become clearer (at least to me) as we prove more such theorems and I figure out how to prove `EXMID` from various propositions.

avekens (approved):

iset.mm
```diff
@@ -34966,6 +34966,126 @@ than an implication (but the two are equivalent by ~ bm1.3ii ).
...
+       largely because it hides statements of the form "for any proposition" in
+       a system which can only quantify over sets, not propositions.
+
+       To see how this compares with other ways of expressing excluded middle,
+       compare ~ undifexmid with ~ exmidundif .  The former may be more
+       recognizable as excluded middle because it is in terms of propositions,
+       and the proof may be easier to follow for much the same reason (it just
+       has to show ` ph ` and ` -. ph ` in the the relevant parts of the
+       proof).  The latter, however, has the key advantage of being able to
+       prove both directions of the biconditional.  To show that excluded
```
benjub, line 34992: To state that excluded...

iset.mm
```diff
@@ -34966,6 +34966,126 @@ than an implication (but the two are equivalent by ~ bm1.3ii ).
...
+       arbitrary sets.  (Contributed by Jim Kingdon, 18-Jun-2022.) $)
+    exmidel $p |- ( EXMID <-> A. x A. y DECID x e. y ) $=
+      ( wem cv wcel wdc wal exmidexmid alrimivv c0 wceq eleq1 dcbid albidv spcv
+      0ex exmid0el sylibr impbii ) CADZBDZEZFZBGZAGZCUCABUBHIUEJUAEZFZBGZCUDUHA
+      JPTJKZUCUGBUIUBUFTJUALMNOBQRS $.
+  $}
+
+  ${
+    $d x y z $.
+    $( Union of complementary parts producing the whole and excluded middle.
```
benjub, line 35067: I would write the first sentence as:
> Excluded middle is equivalent to every subset having a complement (i.e., the union of a subset and its relative complement is the whole set).

benjub (approved):
That makes for a nice section.

Thanks for the comment about bounded separation <-> bounded excluded middle. This equivalence spreads to later results, but the proofs may blur this fact if they use ax-pow.

benjub:
@jkingdon By the way: you may have seen the section in my mathbox "Dedidability of classes". Maybe it can be of some help ([undifdcss](http://us2.metamath.org/ileuni/undifdcss.html) says that a subclass is decidable if and only if it is complemented), although I would agree that a new definition is not worth it at this time.

iset.mm
jkingdon: Ah, OK, that's a bit better. Changed.

iset.mm
jkingdon: I reworded this as two sentences rather than a sentence with a parenthetical, because I thought there was a tool which grabbed the first sentence. But now I'm not so sure. WRITE THEOREM_LIST seems to take the whole description, so maybe I am imagining it.

I guess I'll leave it as two sentences (hope it doesn't seem confusing that way).

jkingdon:
> Thanks for the comment about bounded separation <-> bounded excluded middle. This equivalence spreads to later results, but the proofs may blur this fact if they use ax-pow.

Hmm, interesting point about the axioms interacting. But I guess we don't need to get into that level of detail in this comment (in practice, I suppose people will look at CZF as a whole versus IZF as a whole more so than individual axioms).

jkingdon:
> By the way: you may have seen the section in my mathbox "Dedidability of classes"

It does seem like a concept that comes up from time to time

> I would agree that a new definition is not worth it at this time.

Yeah. I'm not sure whether there is a circumstance where making this a definition would be a good idea, but we don't seem to be there yet.

iset.mm
benjub: Thanks. Just for the record and to clarify why I insisted on the phrase "every subset has a complement": this is a recurring theme: "every subspace of a fin. dim. vector space has a supplement", "every closed subspace of a Hilbert space has a (closed) supplement", the Zassenhaus theorem for "group complements", etc.

benjub (approved):

iset.mm
jkingdon: I'm not familiar with those examples but I'll take your word for it.

PR #2630, opened 2022-06-20
shorten pair 3imp, 3impa

wlammen:
The theorems following df-3an are all bi-conditional centred.  Starting with 3simpa, the df-3an expression is first used in implications.  3simpa is a good starting point to arrive at implication centred theorems, since it has a very short proof.  But is it the best one?  I doubt it.  My last PR showed, that building up antecedents using adant* like theorems lead to slightly better results.

So I looked how this idea can carry over from the simple /\ connective to 3an.  Here I focus on 3impa, since it allows me to modify 3an expressions in place in antecedent position.  I base its proof directly on df-3an, so it can be moved even before 3simpa.  Unfortunately, the proof size of 3impa increases in this step.  Fortunately, this is more than compensated for in its first application in 3imp.  The total number of compressed proof bytes shrinks by 5.  And in total a proof line is spared.  I expect further shortening effects in the range from 3simpa to 3imp, but the details remain to be seen right now.

This PR obeys the rules for acceptable shorter proofs as displayed on the Conventions page.

By the way, meanwhile I was assigned maintainer status on this site.  If no one objects, I will automatically merge this PR after 24 hours.  Approvals are welcome, though.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13498,6 +13498,8 @@ New usage of "3dvdsOLD" is discouraged (0 uses).
 New usage of "3dvdsdecOLD" is discouraged (0 uses).
 New usage of "3ianorOLD" is discouraged (0 uses).
 New usage of "3imp3i2anOLD" is discouraged (0 uses).
+New usage of "3impOLD" is discouraged (0 uses).
+New usage of "3impaOLD" is discouraged (0 uses).
 New usage of "3impdirp1" is discouraged (0 uses).
 New usage of "3impexpVD" is discouraged (0 uses).
 New usage of "3impexpbicomVD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub (approved):

benjub:
> By the way, meanwhile I was assigned maintainer status on this site. If no one objects, I will automatically merge this PR after 24 hours. Approvals are welcome, though.

Even for maintainers, we ask approval from another person.

wlammen:
OK, will do in future.  I wonder how I can be of use in my new role besides merging my own PRs.  I see some difficulties beyond set theory, because I have limited knowledge there.  Let me know what you think.

jkingdon:
> OK, will do in future. I wonder how I can be of use in my new role besides merging my own PRs. I see some difficulties beyond set theory, because I have limited knowledge there. Let me know what you think.

We try to keep our procedures relatively simple. If you focus on things you know or want to learn about, it is pretty much the familiar steps - submit pull requests, review pull requests, merge approved pull requests, comment on issues. Does that answer the question?

jkingdon (approved):

wlammen:
> Does that answer the question?

If I focus on things I know that is in fact no problem.

PR #2631, opened 2022-06-21
shortenings based on the new proof of 3impa

wlammen:
There are a couple of low hanging fruits that can be collected by using 3impa as a starting point of theorems containing triple conjunction in antecedent position.  So far 3simpa served this role, but 3impa and related theorems are more versatile.
That's why we move 3impa and friends across 200 theorems into an earlier position.

This PR contains 7 shortenings, 6 are a consequence of having 3imp* available.  One, 3impia, is a rather technical shortening.  The current measure for short proofs favours specialized theorems with better matching hypotheses over more general ones.  This has no real visual impact, and is noticeable only for insiders.  The other shortenings are visible on the HTML pages.  In total 27 proof bytes are saved, and the essential proof lines decrease by 1.

This PR complies to the acceptable shorter proofs paragraph on the Conventions page.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13478,6 +13478,9 @@ New usage of "2uasban" is discouraged (0 uses).
 New usage of "2uasbanh" is discouraged (1 uses).
 New usage of "2uasbanhVD" is discouraged (0 uses).
 New usage of "2zrngALT" is discouraged (0 uses).
+New usage of "3adant1OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


avekens (approved):

wlammen:
Usually I consider a PR as finished, once it was approved.  I'm going to revert the latest commit that was almost simultaneously applied with an approval.  Need some spare time to do this, though.

PR #2634, opened 2022-06-23
shortenings based on moving adant* theorems closer to df-3an

wlammen:
All modified proofs are shortenings, their compressed proof size shrinks in any instance.  The 35 shortenings made a reordering possible, where similar theorems are packed together.  In total 257 compressed proof bytes and 27 essential proof lines are saved.

1. move more adant* theorems closer to df-3an.  These theorems help writing shorter proofs, so keeping them in an early position is of advantage.  Some are shortened in this process.
2. In the wake, a lot of simp* based on triple conjunction theorems are shortened and clustered.  In many cases the proof idea 'add antecedents at the latest possible moment' was applied.
3. 3imp21 is subject to a very technical correction.

Is this change too big, and you want it rather split up into smaller PRs? Let me know what you think.


benjub:
I approve the changes as described in your comment, but I have to admit that I did not review carefully the code changes. It is indeed a bit hard to see the global gains.  I'm not sure if the "diff visualization tool" of github is a bit lazy, or if there is a way to separate commits so as to reduce diffs and make them more easily checkable, e.g. some commits with only line changes, and others with only modifications (which is generally not possible since line changes imply that proofs cannot refer to some assertions anymore, and it's the whole point of these moves...), or at least more commits with fewer modifications in each.

Also please mention when a modification increases the number of essential steps of a proof.  This does not invalidate the modification, since as you write both the global number of essential steps and each number of total steps may decrease, but it's good to know.

To conclude: maybe wait a bit for another review, and if it does not come and if you're sure of yourself, go ahead and merge.

wlammen:
If you look at the changes in github, the default is a unified view where old and new text appears interleaved in red and green.  But you can change the view by clicking on the cogwheel in the menu line, and pick a side by side view instead, which, I think, is better for this PR, because you can see the bulk of the theorems in direct neighbourhood to the corresponding OLD theorem.

This PR does not increase the compressed proof size of a single theorem.

35 shortenings in a single PR is indeed a lot. That's why I asked.

I don't know whether I understand the "global gains" part correctly.  When I encounter a particular problematic detail in a proof, I occasionally look up a theorem covering a similar problem, and try to adapt its proof to my needs.  So if the proof is sloppy, there is a chance this propagates elsewhere.  The bulk of the corrected theorems miss the principle "add an antecedent at the latest possible moment", and hence are overly complex, although admittedly not to a catastrophic degree.  Another global effect is the clustering of related theorems in a single location.  Following a common order principle may help to find the theorem better you look for.

For those curious, here's the list with saved proof bytes / essential lines, by theorem.
3imp21           1/0 
3adant1l         6/1
3adant1r        6/1
ad5ant14        3/0
ad5ant24        3/0
simpl1             2/0
simpl2              2/0
simpl3              2/0
simpr1          2/0
simpr2          2/0
simpr3          2/0
simpll1         2/0
simpll2         2/0
simpll3         2/0
simplr1         2/0
simplr2         2/0
simplr3         2/0
simprl1         2/0
simprl2         2/0
simprl3         2/0
simprr1         2/0
simprr2         2/0
simprr3         2/0
simpl1l         2/0
simpl1r         2/0
simpl2l         2/0
simpl2r         2/0
simpl3l         2/0
simpl3r         2/0
ad5ant123   39/4
ad5ant124   35/5
ad5ant125   9/1
ad5ant134   41/6
ad5ant135   51/7
ad5ant145   15/2

avekens (approved):
Reviewing the changes, I have the same problems as @benjub, and the split view does not really help a lot. I trust, however, what @wlammen has written in the comments, and quickly stepping through the changes I did not find anything suspicious. Therefore, I approve this PR.

benjub (approved):
Thanks @wlammen for these details.

wlammen:
> ... the split view does not really help a lot

The split view shows the updated file state on the right  pane, and there you have new and old proofs in succession for most of the theorems.  OK, for perfect scrutiny you still have to verify that the old versions agree in both panes, but ... I think to some extent one can trust the established developers here.  Random checks are okay in my eyes.

PR #2385, opened 2021-12-21
Mathbox and (somewhat large) automatic minimization

benjub:
* within my mathbox: minor comment edit
* automatic minimization of the theorems using eleq(1|2) with eleq(1|2)w. It turns out that there were respectively ~340 and 30 of them. Here is the log for the second (I kept the log for the first but it's a bit large to copy it here).
```
Proof of "1stcfb" decreased from 4009 to 4001 bytes using "eleq2w".
Proof of "cantnflem1c" decreased from 879 to 878 bytes using "eleq2w".
Proof of "conncompconn" decreased from 401 to 400 bytes using "eleq2w".
Proof of "cvmcov" decreased from 582 to 581 bytes using "eleq2w".
Proof of "cvmseu" decreased from 662 to 661 bytes using "eleq2w".
Proof of "ddemeas" decreased from 3087 to 3080 bytes using "eleq2w".
    (Uncompressed steps decreased from 1632 to 1630).
Proof of "elcls3" decreased from 894 to 891 bytes using "eleq2w".
Proof of "elintab" decreased from 208 to 207 bytes using "eleq2w".
Proof of "elintabg" decreased from 109 to 107 bytes using "eleq2w".
Proof of "eluniab" decreased from 206 to 205 bytes using "eleq2w".
Proof of "elunif" decreased from 188 to 187 bytes using "eleq2w".
Proof of "flftg" decreased from 777 to 776 bytes using "eleq2w".
Proof of "fnchoice" decreased from 3190 to 3189 bytes using "eleq2w".
Proof of "fneint" decreased from 353 to 349 bytes using "eleq2w".
Proof of "hilbert1.2" decreased from 459 to 456 bytes using "eleq2w".
Proof of "isf32lem2" decreased from 2019 to 2018 bytes using "eleq2w".
Proof of "lssacs" decreased from 895 to 894 bytes using "eleq2w".
Proof of "nbusgrf1o" decreased from 100 to 97 bytes using "eleq2w".
Proof of "nsgacs" decreased from 759 to 758 bytes using "eleq2w".
Proof of "r0cld" decreased from 767 to 764 bytes using "eleq2w".
Proof of "sadcp1" decreased from 1246 to 1245 bytes using "eleq2w".
Proof of "sdrgacs" decreased from 1181 to 1180 bytes using "eleq2w".
Proof of "subgacs" decreased from 652 to 651 bytes using "eleq2w".
Proof of "tcrank" decreased from 2842 to 2840 bytes using "eleq2w".
[...]

PR #2635, opened 2022-06-23
Remove "$d ph x" from rspcdva

jamesjer:
I noticed that the proof of `rspcda` doesn't use `rspcda.1`.  That means it can be removed, which makes it the same theorem as `rspcdva`, but without `$d ph x`.  Therefore, only one of the 2 theorems is needed, and it is not necessary for users of the theorem to prove that `x` is either not free in `ph` or distinct from `ph`.  I chose to keep the `rspcdva` name, because `rspcda` is only used 2 times in `iset.mm` and 0 times in `set.mm`, but `rspcdva` is used multiple times in both.  If the name should be `rspcda`, I can do that rename as a follow-up commit.

Notes:
- I did not add a comment claiming to have shortened the proof of `rspcdva`, because I didn't.  This is Thierry Arnoux's proof (of `rspcda`).
- I am not suggesting that it is a good idea to use `rspcdva` when `x` occurs free in `ph`.  Proving `rspcdva.2` may be difficult in that case.  But that doesn't mean that users of `rspcdva` are _required_ to prove that `x` does not occur free in `ph`.  Indeed, it is already possible to use the combination of `sylc` and `rspcv` shown in the proof of `rspcda` to avoid that (and many proofs do just that; see the next bullet).
- I ran the minimize script with `rspcdva` as the argument on both `iset.mm` and `set.mm`.  The result is quite a large change: iset.mm is reduced by 996 bytes and set.mm by 7,732 bytes, spread over 34 theorems in iset.mm and 319 in set.mm.  I'm not sure what to do with the result.  It's going to make a rather large PR.  Should I break it up into smaller pieces?
- Some `$d` removals are possible as a result of this change, but when I started on them, they began to snowball.  I stopped working on that to focus on this PR.  I can try to figure out the full set that can be removed if there is interest.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+23-Jun-22 rspcda    ---         deleted; use rspcdva instead
 17-May-22 ad5ant1345  adantl3r  eliminate duplicated theorem
 17-May-22 adantlllr  adantl3r  eliminate duplicated theorem
 21-Apr-22 fz1ssfz0  [same]      moved from GS's mathbox to main set.mm


benjub:
Thanks for this neat PR.

As for minimizations, if the PR is the exact result of the minmize script, then even if it's quite large, a single PR makes sense (we'll trust you that you didn't intersperse other changes in between). So one for set.mm and one for iset.mm is fine. The minimize script also writes a log file; can you extract the lines containing "creased" and post the result as a comment to the PR (as in #2385)?

As for the $d removals, it would be nice to propagate them indeed, but maybe it's a bit too time consuming. It is theoretically possible to recompute the minimal sets of DVs for all theorems, since e.g. mmj2 finds them automatically. One day...

**Edit**: turns out I had to close #2385... I guess it's on my TODO list to resubmit it, proceeding that time as I wrote.

benjub (approved):

jamesjer:
Hmm, did I do something wrong?  Or maybe I'm looking in the wrong place?  I ran `scripts/minimize-all rspcdva set.mm "" "mathbox"` (and ditto for iset.mm).  I see files names 0.tmp and 1.tmp, but neither has any lines containing "creased".  For example, nvocnv was minimized, but 1.tmp just says:
```
prove nvocnv
min rspcdva/allow */no_new ax-*
save new_proof/compressed
_exit_pa
```
I don't see any other file that could be a log file.  I'm running git HEAD of metamath-exe, if that matters.

jamesjer:
This PR now conflicts with the commit that moved changes to changes-set.txt, so I will force update in a moment.

jkingdon:
Thanks for noticing this and looking into it. I was writing a proof just the other day, saw a somewhat odd distinct variable violation when using rspcdva, noticed it didn't happen with rspcv, and switched to the latter without really thinking about it. Glad to see you took a deeper look.

As for how to submit the followup minimizations, I'm fine with two big pull requests (for iset.mm and set.mm) as @benjub suggests, or if that seems to have problems with merge conflicts or github limitations or anything, smaller ones are OK with me too. But I suppose I'd start with the large ones unless it seems unworkable.

As for the name, I'd keep the `v` because of the `ch x` distinct variable constraint (compare with http://us.metamath.org/ileuni/rspcv.html versus http://us.metamath.org/ileuni/rspc.html where it is a distinct variable constraint with the result of the substitution which controls the `v`).

jkingdon (approved):

avekens:
~rspcda is/was not used within set.mm, so it could have been deleted (maybe it was used sometime, but was removed explicitly or by a minimization meanwhile). There could be cases for which ~rspcda is useful (if ` ph ` does contain ` x `, but not as free variable): otherwise, `x `has to be substituted (maybe by ` y `) within ` ph ` so that ~rspcdva can be applied. But if @tirix has no objection, ~rspcda can be deleted as proposed by this PR.

About minimization: Which theorem should be replaced by which theorem to obtain a minimization? I tried it with set.mm: Since ~rspcda is not used, it cannot be replaced. If the usage of ~rspcdva should be replaced by ~rspcda, the minimize script does not report any decrease (~rspcdva is used in 36 theorems). Therefore, I don't understand the meaning of

> I ran the minimize script with rspcdva as the argument on both iset.mm and set.mm. The result is quite a large change: iset.mm is reduced by 996 bytes and set.mm by 7,732 bytes, spread over 34 theorems in iset.mm and 319 in set.mm.

@jamesjer You can obtain the theorems which will get a shorter proof by opening a log  (log.txt) before submitting the generated script 1.tmp:

MM> open log log.txt  
MM> submit 1.tmp
MM> close log

avekens:
Ah, I see that scripts/minimize-all rspcdva set.mm "" "mathbox" was used! This script may be modified to get a log with the results (as I described above). Furthermore, it does not check proofs containing ~rspcdva to be replaced by ~rspcda, but the proofs of all theorems if they can be shortened using ~rspcdva. I wonder, nevertheless, why this should result in the mentioned shortenings. I tried it manually for ~nvocnv, and there was no shorter proof found:

```
MM> pr nvocnv
You will be working on statement (from "SHOW STATEMENT nvocnv"):
$d x A $.  $d x F $.
21959 nvocnv $p |- ( ( F : A --> A /\ A. x e. A ( F ` ( F ` x ) ) = x ) -> `' F
      = F ) $= ... $.
MM-PA> min rspcdva/V/i
No shorter proof was found.
```

avekens (approved):

avekens:
Since  ` $d ph x $. ` is removed from ~rspcdva , ~rspcda is actually not needed anymore. Therefore, I approved this PR.

avekens:
Now I got it (sorry for the confusion I may have caused): the proof of ~nvocnv (and maybe many others) can be minimized **after** ` $d ph x $. ` is removed from ~rspcdva:  

```
MM> pr nvocnv
You will be working on statement (from "SHOW STATEMENT nvocnv"):
$d x A $.  $d x F $.
21950 nvocnv $p |- ( ( F : A --> A /\ A. x e. A ( F ` ( F ` x ) ) = x ) -> `' F
      = F ) $= ... $.
MM-PA> min */V/i
Proof of "nvocnv" decreased from 661 to 640 bytes using "rspcdva".
    Proof of "nvocnv" decreased from 1282 to 1230 steps using "rspcdva".
Scanning backward through statements...
Proof of "nvocnv" decreased from 661 to 640 bytes using "rspcdva".
    Proof of "nvocnv" decreased from 1282 to 1230 steps using "rspcdva".
Forward vs. backward: 640 vs. 640 bytes; 1230 vs. 1230 steps
The forward scan results were used.
```

avekens:
Hopefully my last comment on this PR ;-) :
As @jamesjer proposed, 

> the name should be rspcda, ... that rename [can be done] as a follow-up commit.

Maybe together with the minimizations. This is because the suffix/label fragment "v" should be used for

> a theorem eliminating a hypothesis such as ` F/ x ph ` in ~ 19.21 via the use of disjoint variable conditions combined with ~ nfv . 

according to our conventions.

tirix:
This looks good! Thank you Jerry!

I think it could be done in 3 PRs: 
- first this PR,
- then one PR with the minimize,
- and a third PR with removal of unnecessary `$d` distinct variables.

Sorry I've been quite busy recently and did not contribute much...

tirix (approved):

benjub:
> MM> open log log.txt
> MM> submit 1.tmp
> MM> close log

As @avekens says above. Then, you probably know ten ways to select the lines of log.txt containing "decreased". One of them is
> MM> tools
> TOOLS> match log.txt "decreased" Y

jamesjer:
Sorry to be silent.  $DAYJOB has been a beast today.  I probably won't have time to touch this again until tomorrow.  I'll get the requested log information and we can go from there.

Issue #2625, opened 2022-06-17
Add disjoint union to set.mm and iset.mm


jkingdon:
As suggested in https://github.com/metamath/set.mm/issues/2595#issuecomment-1158537525 we should (presumably) add this.

The basic idea can be seen at [df-cda](http://us.metamath.org/mpeuni/df-cda.html) but the idea here is to give it some notation which makes it more useful for purposes other than cardinality.

We need three things:

* a disjoint union syntax. I guess this can be an operator in the sense of http://us.metamath.org/mpeuni/df-ov.html if we want to confine ourselves to sets. I see that the above comment refers to sets but is this usually defined on classes in general or just on sets? Suppose for the sake of argument that the notation is `( A +u B )`
* `inl` which has the property ``x e. A -> ( inl ` x ) e. ( A +u B )``
* `inr` which has the property ``x e. B -> ( inr ` x ) e. ( A +u B )``
* A theorem which lets us go the other way: ``x e. ( A +u B ) -> ( E. y e. A x = (inl ` y ) \/ E. y e. B x = ( inr ` y ) )``

Open questions:
1. sets versus proper classes as described above
2. Does defining `inl` and `inr` as functions work well? (I would imagine so by analogy with `1st`, `ZZ>=`, and others, but I haven't tried implementing anything yet)
3. Notation of the disjoint union syntax. Is this usually a plus sign as in PradicBrown2021 (from the iset.mm bibliography)? If so I suppose it should be plus sign with a subscript, which is where ASCII `+u` and typeset +<sub>u</sub> came from.

digama0:
Common notations for disjoint union are plus, circled plus, and square cup. The definition works on proper classes as well, and the inl and inr functions can indeed be class functions which do not depend on A or B:

```
inl = ( x e. _V |-> <. (/) , x >. )
inr = ( x e. _V |-> <. 1o , x >. )
```
PS: I think df-cda is more useful with the discriminant being the first component rather than the second. [xpsc](http://us.metamath.org/mpeuni/xpsc.html) is an example of a theorem (collection) that actually makes use of this ordering since it can be used as a two-point function.

[...]

PR #2636, opened 2022-06-24
Add disjoint union of two classes to iset.mm

jkingdon:
It seemed good to submit this for iset.mm so I can get feedback and then I'll make another pull request for set.mm.

Most of the background is at #2625 but to summarize:
* `( A |_| B )` is the disjoint union of A and B
* `inl` and `inr` map from elements of A and B to elements of the disjoint union
* the usual equality and not-free theorems
* ``C e. A -> ( inl ` C ) e. ( A |_| B )`` and likewise for `inr`
* ``( ( inl " A ) i^i ( inr " B ) ) = (/)``
* ``( ( inl " A ) u. ( inr " B ) ) = ( A |_| B )``
* ``C e. ( A |_| B ) -> ( E. x e. A C = ( inl ` x )
        \/ E. x e. B C = ( inr ` x ) )``

Open questions (for follow up pull requests, no need to block this one on these):

* What is the universal property as described in https://github.com/metamath/set.mm/issues/2625#issuecomment-1158868564 ? Trying to look this up got me up to my shoulders in category theory and I was pretty lost about how any of this translates to set.mm or iset.mm.
* Can this go directly into set.mm (I suppose in the same part of the file as in iset.mm, right before cardinality), or is it supposed to go into a mathbox first? It would seem to be relatively "done" and intended to be built upon, but I don't contribute enough to set.mm to have a feel for how strong the "mathbox first" urge is.


digama0:
The simplest counterpart of the universal property is an operation I called `case` in peano.mm1 which merges two functions on A and B respectively to make a function on A+B.

``case F G = { <. ( inl ` x ) , y >. | <. x , y >. e. F } u. { <. ( inr ` x ) , y >. | <. x , y >. e. G }``

benjub:
Consider the problem: "given two sets A and B, I want to find a set X such
that for any set C, the data of two functions `F : A -> C` and `G : B -> C`
amounts to the data of one function `H : X -> C`".  If X is a solution to
this problem, then taking H to be the identity of X, it gives me two
functions for the corresponding F and G, which we call inl and inr.
Therefore, you can check that what I'm really looking for is a triple
consisting in a set, denote it by `A |_| B`, and two functions, `inl : A ->
A |_| B` and `inr : B -> A |_| B`, such that
```
ph -> F : A -> C
ph -> G : B -> C
=> ph -> E! h ( h : A |_| B -> C /\ ( ( h o. inl ) = F /\ ( h o. inr ) = G )
```
Indeed, once I have such a triple with that property, check that the
"amounts to" above holds: given F and G, take H given by the universal
property, and given H, take `F := ( H o. inl )` and `G := ( H o. inr )`.
These universal properties give "existence up to unique isomorphism making
the obvious diagram commute" (if there are two solutions X and Y, just
apply the universal property to X with C:=Y and to Y with C:=X: you find
two morphisms which have to be inverse to each other).

The function `h` is the one given by Mario above, and the properties you
just proved for the coproduct are exactly the ones needed to prove its
universal property.  For the product of two sets, this is exactly the same
thing with all arrows reversed (this is the "dual" construction); maybe
@avekens already formalized the associated universal property in set.mm ?

It can go directly to the main part of set.mm, in its own subsection
"Disjoint union of two classes" as early as its dependencies allow
(ideally, right after the cartesian product of two classes).

avekens:
According to the discussion about the "universal property": there are several universal properties in set.mm (e.g., ~frmdup3 for free monoids, ~frlmup4 for free modules, ~upxp for the Cartesian product - maybe this is the one @benjub is thinking of). Unfortunately, all these theorems are not provided/proven by me...

benjub:
> ~upxp for the Cartesian product

Bingo !

> Unfortunately, all these theorems are not provided/proven by me...

You proved some theorems about initial and final objects, no ? These are universal properties too !

benjub (approved):

avekens:
I have an idea how to prove @benjub`s universal property theorem
ph -> F : A -> C
ph -> G : B -> C
=> ph -> E! h ( h : A |_| B -> C /\ ( ( h o. inl ) = F /\ ( h o. inr ) = G )

Instead of going into details "on paper", I would like to have the basic definitions and theorems available in (i)set.mm. Therefore, I propose to accept and merge this PR first. The universal property theorem can be added later in a separate PR.

avekens (approved):

jkingdon:
> I have an idea how to prove @benjub`s universal property theorem

I'm glad to hear that, because I didn't get very far on this. In case it helps, here's the statement of the theorem from this github issue with a little bit of the metamath syntax added:

```
    updju.a $e |- ( ph -> F : A --> C ) $.
    updju.b $e |- ( ph -> G : B --> C ) $.
    updju $p |- ( ph -> E! h ( h : ( A |_| B ) --> C
        /\ ( h o. inl ) = F /\ ( h o. inr ) = G ) ) $=
      ? $.
```

jkingdon:
> It can go directly to the main part of set.mm, in its own subsection "Disjoint union of two classes" as early as its dependencies allow

Submitted as #2642 .

As far as I noticed the limiting dependency was `1o`. Also, putting this right before cardinality seems logical to me in the sense that http://us.metamath.org/mpeuni/df-cda.html is a kindred definition.

digama0:
I think the talk about universal properties is hurting more than it helps. The basic theorems you want here are:

```
|- ( inl |` A ) : A --> ( A |_| B )
|- ( inr |` B ) : B --> ( A |_| B )
|- ( ( F : A --> C /\ G : B --> C ) -> case F G : ( A |_| B ) --> C )
|- ( X e. A -> ( case F G ` ( inl ` X ) ) = ( F ` X ) )
|- ( Y e. B -> ( case F G ` ( inr ` Y ) ) = ( G ` Y ) )
|- ( X e. ( A |_| B ) -> ( E. x e. A x = ( inr ` x ) \/ E. y e. B x = ( inr ` y ) ) )
|- ( F : ( A |_| B ) --> C -> F = case ( F o. inl ) ( F o. inr ) )
```
From these (especially the last one) the universal property follows (although you have to assume `A` and `B` are sets for that existential uniqueness to be inhabited).

benjub:
I'm not sure how it hurts. Universal properties are a pretty central concept.  I understand your preference: instead of proving an abstract unique existence result, you *construct* a map (`case F G`), prove that it satisfies the constraints, and prove that all other maps satisfying these are equal to it. This is more concrete, more constructive, and provides a "witness" of the proof of unique existence.

On the other hand, this requires a new definition (`case`) which does not appear in (m)any math textbooks, and the result appearing in most textbooks is the one about existential uniqueness. The metamath lines you wrote are the canvas of its proof, or maybe intermediate lemmas. On the other hand, I am aware of the new importance, for instance in HOTT, given to a "witness" of a proof, so... not sure where to draw the line.

PS: Going with the above properties, are
```
|- ( F : A --> C -> ( case F G o. ( inl |` A ) ) = F )
|- ( G : B --> C -> ( case F G o. ( inr |` B ) ) = G )
```
**Edit**: sorry, you actually wrote them above (these are just the variable-free versions, but still noteworthy of their own)

sctfn:
Unless I’m missing something, case is just a Skolem function. They’re pretty standard in FOL courses/textbooks. Basically the statement is that FOL with existential quantifiers is equiprovable to FOL with function symbols.

benjub:
> Unless I’m missing something, case is just a Skolem function.

It is. It is even the dual of the "range cartesian product" you introduced and we discussed a while back. But that does not make `case F G` widely used. I see how it can be widely used in @digama0's developments connected with the x64 architecture, but I'm not sure for set.mm (of course someone could develop a corpus which uses it intensely, but it seems we are not there).

digama0:
Any time you have discriminated unions or "sum types" `case` comes up. Data structures are all about sums and products, and it's one of the primitive inductive type constructors. It is essential for doing anything "syntactic" like reasoning about formulas, because you have to encode the formulas using sums and products.

If `|_|` is to be defined at all, then I would expect `inl` `inr` and `case` to go along with it, these are just the basic introduction and elimination rules for the type. We have managed to get quite far without it but I do think it is hindering things like free algebras, proof theory and other areas that require more than just product types.

> I'm not sure how it hurts. Universal properties are a pretty central concept. I understand your preference: instead of proving an abstract unique existence result, you construct a map (case F G), prove that it satisfies the constraints, and prove that all other maps satisfying these are equal to it. This is more concrete, more constructive, and provides a "witness" of the proof of unique existence.

In practice, `case` is simply much more usable than the abstract existence version, because you can't compose that a bunch of times when crafting a definition. The abstract perspective can be helpful to identify what theorems are interesting, but it is also more limiting - for example `case` works on proper class functions while the existence statement doesn't. It also requires everyone else to speak category theory, which is unnecessary in this instance.

benjub:
> Any time you have discriminated unions or "sum types" `case` comes up. Data structures are all about sums and products, and it's one of the primitive inductive type constructors. It is essential for doing anything "syntactic" like reasoning about formulas, because you have to encode the formulas using sums and products.

That's why I wrote above that I understand its need and use in your formalization of computer architectures. For the moment, set.mm, which is getting more representative of undergrad maths, with around 40,000 theorems, does not seem to need it much.

I am not opposed to its introduction as a new definition, it's just that the attitude towards new definitions was more like "once there is a sizable corpus of theorems using it (e.g. in a mathbox), one might introduce a definition", rather than "make the definition in case it is used later".

If you think time has come, then go for it (but maybe with an infix symbol, like `F |_|dom G` with symbol $F \sqcup_\mathrm{dom} G$ similar to the $F \times_\mathrm{ran} G$ for the dual construction "range product").

> It also requires everyone else to speak category theory

Is this a pro or a con ? Evangelization of the masses !

digama0:
> > Any time you have discriminated unions or "sum types" `case` comes up. Data structures are all about sums and products, and it's one of the primitive inductive type constructors. It is essential for doing anything "syntactic" like reasoning about formulas, because you have to encode the formulas using sums and products.
> 
> That's why I wrote above that I understand its need and use in your formalization of computer architectures. For the moment, set.mm, which is getting more representative of undergrad maths, with around 40,000 theorems, does not seem to need it much.

My point was more than just that. I assert that *not having the definition is already holding us back* in areas where syntactic reasoning is necessary, such as in free algebras, proof theory, some kinds of combinatorics and graph theory, computability theory, and of course other more concrete kinds of computer science. My argument has nothing to do with x86, although that's certainly a very syntactic kind of reasoning (in fact, it comes up not in the x86 instructions themselves encoded as bits but rather their formal model as a sum type).

Most other proof assistants have some kind of helper infrastructure for constructing enumerated types with or without associated data, but it's difficult to miss something you've never had and it is hard to say just how many areas have simply been unexplored because metamath is not very well suited to this kind of reasoning. Looking at the peano.mm1 formalization is probably telling for the kind of things that are missing for more computer science oriented applications.

avekens:
Instead of using ` case `, I tried to use the following function ` H ` as ` h` in the universal property theorem:

```
    updjud.f $e |- ( ph -> F : A --> C ) $.
    updjud.g $e |- ( ph -> G : B --> C ) $.
    updjudhf.h $e |- H = ( x e. ( A |_| B ) |-> if ( ( 2nd ` x ) e. A ,
                           ( F ` ( 2nd ` x ) ) , ( G ` ( 2nd ` x ) ) ) ) $.
    updjudhf $p |- ( ph -> H : ( A |_| B ) --> C ) $=
```

However, I am not able to prove ~updjudhf in iset.mm, because it requires

`( ( ph /\ x e. ( A |_| B ) ) -> DECID ( 2nd `` x ) e. A )`

Since I am not so familiar with iset.mm, I do not know if this holds and how this can be proven. In set.mm, the proof of ~updjudhf would be straightforward...

digama0:
You don't want to case on ``( 2nd ` x ) e. A``, because not only is that not decidable, it's also just wrong for the case where `A = B` or they are otherwise overlapping. You should instead be asking whether ``( 1st ` x ) = (/)`` or ``( 1st ` x ) = 1o``, and this is decidable.

benjub:
Maybe you can alternately define
```
|- H = ( x e. ( A |_| B ) |-> if ( x e. ( inl " A ) , ( F ` ( 2nd ` x ) ) , ( G ` ( 2nd ` x ) ) ) )
```
and use http://us2.metamath.org/ileuni/djur.html and http://us2.metamath.org/ileuni/djuin.html to get decidability (this is the same, but probably makes for a little shorter proof thanks to these theorems).

digama0:
I don't think there is any need to involve `A` there. It could be `x e. ran inl` instead, which is equivalent to ``( 1st ` x ) = (/)``.

Well, abstracting away this if statement and the argument for decidability is exactly the point of the `case` function.

benjub:
> I don't think there is any need to involve `A` there. It could be `x e. ran inl` instead, which is equivalent to `` ( 1st ` x ) = (/) ``.

My goal was to be able to use djur more directly, but maybe your proposal leads to a shorter proof... not sure.

> Well, abstracting away this if statement and the argument for decidability is exactly the point of the `case` function.

It's common that we have "local definitions" as $e-hypotheses within a chain of lemmas. If you think `case` is beneficial enough to justify a new definition, then ok (maybe with the infix notation I proposed above). I don't know about other proof checkers or proof assistants, but if as you write this is a very common device (including for those not based on type theory, but on set theory?), then it may indeed be time to introduce it. My fear is mainly to diverge too much from "textbook math" and go closer to computer science, but since we are midway...

avekens:
Thanks a lot, @digama0 and @benjub for your hints. With these, I can proceed now...

PR #2641, opened 2022-06-25
shortenings based on syl3an*

wlammen:
Theorem syl3an and friends can be used for shortenings.  They depend neither on any 3simp* or 3adant* theorems, so they can be moved to an early position.  Some infrastructure like 3jca needs to be moved as well.
Here are the details (saved compressed proof bytes / essential proof steps):
3imp3i2an   6/1; Could have been the result of a minimize so comments are not changed
3adant1l    2/0; A minor change improving on a very recent commit, so no commit update provided
3adant1r    2/0; A minor change improving on a very recent commit, so no commit update provided
3adant2l    7/1
3adant2r    7/1
3adant3l    7/1
3adant3r    7/1
In total 38 bytes and 5 essential proof lines are saved.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13486,7 +13486,11 @@ New usage of "3adant1OLD" is discouraged (0 uses).
 New usage of "3adant1lOLD" is discouraged (0 uses).
 New usage of "3adant1rOLD" is discouraged (0 uses).
 New usage of "3adant2OLD" is discouraged (0 uses).
+New usage of "3adant2lOLD" is discouraged (0 uses).
+New usage of "3adant2rOLD" is discouraged (0 uses).
 New usage of "3adant3OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
This pull request is the same as https://github.com/metamath/set.mm/pull/2639, but has merged the latest changes by others, nothing else.
@benjub approved #2639, so I think this carries over to this PR.

benjub:
It's true that github makes it hard to review changes with global moves, even with the "split view" (one would need to be able to scrool and two columns independently). The real diffs are actually much smaller than they appear...

benjub (approved):

wlammen:
> one would need to be able to scroll and two columns independently

I suggest to open two browser instances showing the same diff.  Adjusting the browser size such that you have the old pane on the left, and the new pane on the right half of the screen without any overlapping of the instances may help.  Not perfect, but not really bad either.

PR #2642, opened 2022-06-25
Add Disjoint union to set.mm

jkingdon:
This section is copied from iset.mm. Two of the proofs needed small
changes for things like differences in fvex and opex but other than that
it is copied without change.


benjub (approved):

jkingdon:
Since this is essentially unchanged from the iset.mm version which was reviewed at #2636 I suppose there is little need to delay merging it.

If anyone has any suggested changes, feel free to speak up because follow up pull requests are always an option.

PR #2652, opened 2022-06-28
Universal property of the disjoint union

avekens:
As discussed in PR #2636, the universal property of the disjoint union is provided and proven in iset.mm: ~updjud. 

In contrast to the proposal in PR #2636, this theorem holds for the disjoint union of sets only, not for classes in general (see hypotheses updjud.a and updjud.b). Furthermore, the very generally defined functions ` inl ` and ` inr ` had to be restricted to the corresponding (base) sets of the disjoint union.

To prove this theorem, several auxiliary theorems have been provided. I think the theorems ~inlresf and ~inrresf could be improved to show that the corresponding restricted funtions are one-to-one, as their names indicate (left/right injection), but this was not necessary for the proof of the universal property.

If this PR is accepted and merged, I will copy these theorems to set.mm (I expect only some minor modifications of the proofs) and provide them in a separate PR.


iset.mm
```diff
@@ -63863,6 +63893,54 @@ all reals (greatest real number) for which all positive reals are greater.
       KVBFVKVDTVEWHXFAWKCXDWGXEDWCWKPVFRVHTVLVM $.
   $}
 
+  ${
+    $d A x y $.  $d B x y $.
+    $( A disjoint union is a subset of a Cartesian product.  (Contributed by
+       AV, 25-Jun-2022.) $)
+    djuss $p |- ( A |_| B ) C_ ( { (/) , 1o } X. ( A u. B ) ) $=
+      ( vx vy c0 c1o cv wcel cinl cfv wceq wrex cinr wa cop cvv cmpt a1i adantr
```
benjub, line 63901: Remark: djuex is also a consequence of this, since the RHS is a set. I don't know if this provides a shorter proof.

iset.mm
```diff
@@ -63863,6 +63893,54 @@ all reals (greatest real number) for which all positive reals are greater.
...
+      AVBVCWCWEDBVRBHZWCNZVQFVROZVPWPVQWBWQWOWCTWOWBWQKWCWOCVRFVQOZWQPMPMCPWRQK
+      WOCVDRWJWRWQKWOVQVRFUOUIVRBUJWQPHWOFVRFVEVFVGZWKUMRUNSUPWPFVNHZWMNWQVPHWP
+      WMWTWOWMWCVRBAVHSEFWSVIUSFVRVNVOUTVAVBVCVJVKVL $.
+  $}
+
+  $( The first component of an element of a disjoint union is either ` (/) ` or
+     ` 1o ` .  (Contributed by AV, 26-Jun-2022.) $)
+  eldju1st $p |- ( X e. ( A |_| B )
+                   -> ( ( 1st ` X ) = (/) \/ ( 1st ` X ) = 1o ) ) $=
+    ( cdju c0 c1o cpr cun cxp wss wcel c1st cfv wceq wo djuss ssel2 xp1st elpri
```
benjub, line 63917: Using djur or djuun doesn't make the proof shorter ? Or conversely does eldju1st (and eldju2ndl/r) allow shortenings of these ?

benjub:
Thanks. Both the additional staple theorems and the universal property are nice.
The `case` function proposed by @digama0 appears in a $e-hypothesis (`updjudhf.h` defining `H`) used by the three characterization theorems, so if/when this is introduced as a genuine definition, updates will be easy.

Indeed, it's be nice to strengthen your two theorems to prove that ``(inl |` A)`` and ``(inl |` B)`` are injective. Isn't the work already done in djulf1o ? (either "pass along" injectivity in each line or your proof, or simply append at the end of your proof injectivity, as a restriction of an injective function, which might be somewhere in set.mm ?)

Tell me if you prefer to merge this first.

benjub (approved):

iset.mm
avekens: Yes, the proof of ~djuex becomes shorter if it is using ~djuss:

``djuex2 $p |- ( ( A e. V /\ B e. W ) -> ( A |_| B ) e. _V ) $=
    ( wcel wa cdju c0 c1o cpr cun cxp cvv con0 0ex 1on prexg mp2an unexg xpexg
    sylancr wss djuss a1i ssexd ) ACEBDEFZABGZHIJZABKZLZMUFUHMEZUIMEUJMEHMEINEU
    KOPHIMNQRABCDSUHUIMMTUAUGUJUBUFABUCUDUE $.``

But then this theorem has to be moved after ~djuss, and depends indirectly on the definitions of ` inl ` and ` inr ` (whereas the current proof doesn't). Therefore, I do not know which is the best choice...

iset.mm
avekens: I tried to prove ~eldju1st with ~djur, but the proof becomes longer. I do not expect many gains playing around with the other combinations, therefore I would not work on this further.

avekens:
> Indeed, it's be nice to strengthen your two theorems to prove that `` (inl |` A) `` and `` (inl |` B) `` are injective. Isn't the work already done in djulf1o ? (either "pass along" injectivity in each line or your proof, or simply append at the end of your proof injectivity, as a restriction of an injective function, which might be somewhere in set.mm ?)
> 
> Tell me if you prefer to merge this first.

Unfortunately, there was no adequate theorem to prove the injectivity of the functions using djulf1o/djurf1o. But I can add one:
``f1resf1 $p |- ( ( F : A -1-1-> B /\ C C_ A /\ ( F |` C ) : C --> D ) -> ( F |` C ) : C -1-1-> D )``
By this, it is easy to prove that `` (inl |` A) `` and `` (inl |` B) `` are injective. I would like to do this within this PR.

benjub:
> f1resf1 $p |- ( ( F : A -1-1-> B /\ C C_ A /\ ( F |` C ) : C --> D ) -> ( F |` C ) : C -1-1-> D )

Don't we have
```
$p |- ( F : A -1-1-> B -> ( F |` C ) : ( A i^i C ) -1-1-> B ) $= ? $.
```
?  The question of either assuming ` C C_ A` or else not assuming it and using `( A i^i C )` is not very important. The important part is in not requiring anything of ``( F |` C )``.

iset.mm
```diff
@@ -63850,6 +63854,14 @@ all reals (greatest real number) for which all positive reals are greater.
       JZLZCAMTNMOZEPEQVBVGCARUMUAMVHEUBCAVAEUCUDVCAJZVEVFVIVCMVDVCMJZVICUEZSVDC
       MTVCUFZUGZIZMEVMCUHUIVLMJZVNMUJCMCMVLMUKVOVJTVCUNVKULSUOUPUQABVCURUSUT $.
 
+    $( The left injection restricted to the first class of a disjoint union is
+       an injective function from the first class into the disjoint union.
+       (Contributed by AV, 28-Jun-2022.) $)
```
benjub, line 63859: Maybe I missed in in the previous commit, but since B is arbitrary, maybe:
>The left injection restricted to a class is an injective function into the disjoint union of that class (on the left) with any other class.

And same with inrresf1 (and same with the versions without injectivity).

iset.mm
benjub: Given that it's sizably shorter, I would go for this second proof. Or if you prefer, keep the first proof and simply mention in its comment "Note that a shorter proof is possible from ~ djuss and ~  ssexd ."

avekens: I would like to keep the first proof and to provide the second proof as ALT version. Iwould propose, however, to provide this ALT version in set.mm only, not in iset.mm (I do not know exactly how to handle discouraged theorems in iset.mm). @jkingdon what do you think?

avekens:
> > f1resf1 $p |- ( ( F : A -1-1-> B /\ C C_ A /\ ( F |` C ) : C --> D ) -> ( F |` C ) : C -1-1-> D )
> 
> Don't we have
> 
> ```
> $p |- ( F : A -1-1-> B -> ( F |` C ) : ( A i^i C ) -1-1-> B ) $= ? $.
> ```
> 
> ? The question of either assuming ` C C_ A` or else not assuming it and using `( A i^i C )` is not very important. The important part is in not requiring anything of `` ( F |` C ) ``.

We have 
``  f1ssres $p |- ( ( F : A -1-1-> B /\ C C_ A ) -> ( F |` C ) : C -1-1-> B ) $=``
but this theorem as well as your proposal requires  `B = D` in my version. This is, however, needed to prove that ``(inl |` A)`` and ``(inl |` B)`` are injective: here we have `B = ( { (/) } X. _V )` and `D = ( A |_| B )`. Notice that ~f1resf1 does even not require `D C_ B`!

iset.mm
avekens: Although yor proposal is correct and more precise, I think my description is also correct and a little bit simpler: it should be clear by the theorem itself that B is arbitrary and that A is the left side of the disjoint union. Therefore, I would like to keep the current comment. Maybe I can replace "first/second class of the disjoint union" by "left/right class of the disjoint union".

benjub: Ok, as you prefer.

benjub:
> We have `` f1ssres $p |- ( ( F : A -1-1-> B /\ C C_ A ) -> ( F |` C ) : C -1-1-> B ) $= `` but this theorem as well as your proposal requires `B = D` in my version. This is, however, needed to prove that `` (inl |` A) `` and `` (inl |` B) `` are injective: here we have `B = ( { (/) } X. _V )` and `D = ( A |_| B )`.

Ah, so it was indeed already in set.mm !  I was surprised it was not. And I see that your theorem uses it, so all is fine. I'll have a look at the proof, but note that there are two kinds of results: one (f1ssres) dealing only with injectivity of a restriction, and the other only with the range of a restriction (where injectivity plays no role).

>  Notice that ~f1resf1 does even not require `D C_ B`

This is not surprising: in expressions like `X : C --> D`, `D` is only an upper bound for the range of `X`.

benjub (approved):

iset.mm
avekens: comments adjusted

benjub (approved):

iset.mm
jkingdon: > Iwould propose, however, to provide this ALT version in set.mm only, not in iset.mm (I do not know exactly how to handle discouraged theorems in iset.mm

The answer to the question about `discouraged` is that there is an `iset-discouraged` which is managed just like its set.mm counterpart.

I guess I'd vote for just having the ALT version in set.mm. In most cases there isn't any reason to have ALT versions in both set.mm and iset.mm.

Issue #2621, opened 2022-06-16
Define N_infinity in iset.mm


jkingdon:
At the moment iset.mm has http://us.metamath.org/ileuni/df-xnn0.html but this is not the most useful definition in the absence of excluded middle.

We would like to define it with a definition like the one in Section 3.1 of [Pierik], p. 15.

Chances are #2622 is a prerequisite.

There's also the question of notation: because this set includes zero, perhaps we want to keep the ℕ<sub>0</sub><super>*</super> notation  rather than trying to subscript ℕ with both zero and `∞`? It doesn't seem especially appealing to use ℕ<sub>∞</sub> for a something which includes zero. More than notation but perhaps related is the question of whether ℕ<sub>0</sub> is a subset of N_infinity or whether its elements can just be mapped to elements of N_infinity.

We can prove:
* N_infinity = ( NN0 u. { ∞ } ) <-> _om e. Omni (Proposition 3.6.1 of [Pierik], p. 15)
* N_infinity e. Omni (Theorem 3.4 of [Pierik], p. 15, where Omni is defined at https://github.com/metamath/set.mm/issues/2620 )

[Pierik] "Infinite Omniscient Sets in
Constructive Mathematics" by Ceel Pierik, https://www.cs.ru.nl/bachelors-theses/2020/Ceel_Pierik___4806182___Infinite_Omniscient_Sets_in_Constructive_Mathematics.pdf

More background reading: https://ncatlab.org/nlab/show/extended+natural+number

benjub:
To keep compatibility with set.mm, I think we should keep df-xnno with its current definition, and *in addition* add Escardo's one-point compactification. Maybe use $\mathbb{N}_\infty$ for the latter (two subscripts would be a bit heavy). State in its comment that it's the more useful notion in the constructive context. Then we have a theorem `$e |- I = ... $.  $p |- I : xNN0 -1-1-> NNinfty` and you can prove that I being bijective is equivalent to the limited principle of omniscience in the form `omega e. Omni`.

I think it would be too cumbersome to adapt either set to have a genuine inclusion. We have `NN0 C_ xNN0` and the above natural  injection `I`, and we should probably be content with that.

jkingdon:
I've been looking at the Pierik in more detail, and a few observations:

[...]

PR #2654, opened 2022-06-30
Add omniscience syntax to iset.mm

### LINKED ISSUE #2620: define notation for omniscience in iset.mm

jkingdon:
Most of the details are at #2620 and this pull request follows that issue quite closely except that it uses the `EXMID` syntax for ` EXMID -> A. x x e. Omni `.

Fixes #2620

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1369,7 +1369,7 @@ is double negation elimination.</TD>
 
 <TR>
 <TD>dfrex2</TD>
-<TD>~ rexalim </TD>
+<TD>~ rexalim , ~ dfrex2dc</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -64121,6 +64129,112 @@ all reals (greatest real number) for which all positive reals are greater.
...
+       Omniscience (LPO).  (Contributed by Jim Kingdon, 28-Jun-2022.) $)
+    df-omni $a |- Omni = { y | A. f ( f : y --> 2o
+        -> ( E. x e. y ( f ` x ) = (/) \/ A. x e. y ( f ` x ) = 1o ) ) } $.
+  $}
+
+  ${
+    $d A f x y $.
+    $( Expansion of ~ df-omni in terms of classes.  (Contributed by Jim
+       Kingdon, 28-Jun-2022.) $)
+    dfomnicl $p |- ( A e. V -> ( A e. Omni <-> A. f ( f : A --> 2o
```
benjub, line 64160: Isn't our convention to use "isomni", with comment "The predicate of being omniscient." or something like that ?

Side remark: the statement you wrote is the most useful statement, but it is a characterization for sets, not for classes. The characterization for classes is
> |- ( A e. Omni <-> ( A e. _V /\ A. f ... ) )

Sometimes, both statements are in set.mm, but generally the one you wrote is more useful.

benjub (commented):
Another comment: the article by Escardo has the advantage of being the original source (if not for the definition of omniscient, at least for some results about this notion) and of being published. I'm not advocating to always quote the original source, of course ((i)set.mm is not an encyclopedia, and textbooks/monographs are often better choices for references), but here I'm not sure referring to a master thesis is a better choice. (I'm not sure it is not either, depending on whether you want to formalize large chunks of it, including original results. Up to you).

iset.mm
```diff
@@ -64121,6 +64129,112 @@ all reals (greatest real number) for which all positive reals are greater.
...
+      WTYPLWPHWSWEOWFWGWHAYPHJZXDAYRUUAYTWRHHQWIWGWKWLWMWNWOT $.
+  $}
+
+  ${
+    $d f x y $.
+    $( Given excluded middle, every set is omniscient.  Remark following
+       Definition 3.1 of [Pierik], p. 14.  Given ~ omniexmid it presumably
+       would be possible to strengthen this to a biconditional.  (Contributed
+       by Jim Kingdon, 29-Jun-2022.) $)
+    exmidomni $p |- ( EXMID -> A. x x e. Omni ) $=
```
benjub, line 64226: > |- ( EXMID -> _V = Omni )

has the advantage of being variable-free and of looking like the axiom of constructibility V = L.

iset.mm
```diff
@@ -64121,6 +64129,112 @@ all reals (greatest real number) for which all positive reals are greater.
...
+      MUWAYBUUDNUWBSWQWRWSWTWLYLXMXQXAUVKXMXPXBXQKXMYKXCXDXEXFVHXGXQRIXRUUAUIXM
+      XPUVTXOUVIXHXIFXQERUKULWAVHXJ $.
+  $}
+
+  ${
+    $d f x y $.  $d ph x y z $.
+    omniexmid.1 $e |- x e. Omni $.
+    $( If all sets are omniscient, excluded middle follows.  (Contributed by
+       Jim Kingdon, 29-Jun-2022.) $)
+    omniexmid $p |- ( ph \/ -. ph ) $=
```
benjub, line 64203: Why not use `DECID ph` ?
And actually,
> |- ( _V = Omni -> DECID ph )

iset.mm
benjub: And for the converse: let x \subseteq 1o and choose the function f which is constant equal to 0. By hypothesis, x is omniscient so either there is a y in x with f(y)=0, and in particular x is inhabited so x = 1o, or else for all y in x, f(y)=1, that is, for all y in x FALSE, that is, x is empty. Therefore, EXMID holds. Correct ?

iset.mm
avekens: I agree with @benjub that the name of this theorem should be changed.

iset.mm
jkingdon: As for `_V = Omni` I don't know, it doesn't read as well to me but maybe I'm just not used to it or something.

As for the converse, thanks for that, I've formalized that proof. That seems like a mostly general procedure that probably could work for a lot of the existing *exmid theorems.

iset.mm
jkingdon: I guess there is a bit of a stylistic question, at least in my mind, about whether to use definitions like `DECID` everywhere we can, or just where it is hard not to. This thing about `ph \/ -. ph` as the conclusion of an *exmid theorem is what we have been doing, for whatever it is worth.

It is a moot point, however, if we use `EXMID` instead (as this pull request now does). So maybe there is no need to worry about it much any more.

iset.mm
jkingdon: I was hunting around for existing conventions (because I had forgot about the `isfoo` thing) and didn't find it. So thanks for this.

As for class versus set, I suppose that comment could have been worded as "class variable" rather than "class" but I'll just go with "The predicate of being omniscient." which avoids the issue.

jkingdon:
> the article by Escardo has the advantage of being the original source (if not for the definition of omniscient, at least for some results about this notion) and of being published.

Hmm, there seems to be a copy at https://www.cs.bham.ac.uk/~mhe/papers/omniscient-journal-revised.pdf so my first concern - paywalls - is perhaps not a big issue.

I do think the Pierik is a lot more friendly to someone who doesn't know all about intuitionistic/constructive topics. So I think it is probably the better of the two for our purposes. On the topic of whether I, or anyone else, is likely to formalize more of the results in either paper, I suppose it depends on whether we get around to #2621 (which I guess is in both papers). At least, that strikes me as a more likely target than the results in Pierik on squashed sums and strong extensionality.

As for seeking out the original source, I remember that at a young age I went to the university library to find a copy of Cantor's original work (I forget which one, probably whichever one they had) on set theory. I suppose it was an interesting enough exercise, although I think the main thing I learned is that there is this German way of writing letters which I hadn't seen before, rather than being able to grasp any of the math.

jkingdon:
@benjub @avekens I have responded to all comments. Ready for re-review.

iset.mm
benjub: So maybe a lemma like
```
|- ( EXMID <-> A. x ( x C_ 1o -> ( E. y y e. x \/ x = (/) ) )
```
will turn out to be useful ? (works for any set, not necessarily included in 1o).  Just an idea for the future.

benjub: So maybe a lemma like
```
|- ( EXMID <-> A. x ( x C_ 1o -> ( E. y y e. x \/ x = (/) ) )
```
will turn out to be useful ? (works for any set, not necessarily included in 1o).  Just an idea for the future.

benjub (approved):
Thanks for the answers. All good to me.

Issue #664, opened 2018-12-04
Prove moabex in iset.mm


jkingdon:
http://us.metamath.org/mpeuni/moabex.html is a theorem in set.mm which we do not have (yet at least) in iset.mm but which seems potentially provable.

The weaker version http://us.metamath.org/mpeuni/euabex.html is easy to prove:

```
${
  $d y x $.  $d y ph $.
  $( The abstraction of a wff with existential uniqueness exists.
     (Contributed by NM, 25-Nov-1994.) $)
  euabex $p |- ( E! x ph -> { x | ph } e. _V ) $=
    ( vy weu cab csn wceq wex cvv wcel euabsn2 vex snexg ax-mp mpbiri
    cv eleq1 exlimiv sylbi ) ABDABEZCPZFZGZCHTIJZABCKUCUDCUCUDUBIJZUA
    IJUECLUAMNTUBIQORS $.
$}
```

As for how to prove `moabex` the following are things we can show:

1. From `mo3`, `elsn` and logic:
```
|- ( E* x ph
   <-> A. x
         ( ph
          -> A. y
                ( [ y / x ] ph -> y e. { x } ) ) )
```

2. `{ x | ph } = { y | [ y / x ] ph }` follows from `cbvab`

3. We do have some potentially helpful theorems like `abss` and `ss2ab`.

I can't really tell whether those are on the right track or not. At times it seems like `moabex` should just be a few easy steps away, at other times it feels like I'm running right into "there are two cases: the set has one element or it has none" or other territory which won't work.

As for why it would be useful, `moabex` would lead to `snex` and other good stuff.

jkingdon:
Oh let me not be coy, I'm asking @digama0 if he has a proof of this theorem or a proof that it can't be done.

digama0:
I don't think it is provable. I know it feels very plausible, but the right way to think about it is that, relative to the rank hierarchy, an assertion that a set exists is giving a constructive bound on the level V_alpha in which all of the set's members can be found. In the case of a predicate that has at most one satisfying instance, we have no such bound. If we know additionally that there is some element in the set, then this element provides the bound, but without it we have nothing. You could write something like `( E* x ph /\ E. y ( A. x ph -> x e. y ) ) -> { x | ph } e. _V`, but that's trivial from the second conjunct; the `mo` pulls no weight.

digama0:
[...]

PR #2659, opened 2022-07-03
Show that fodomr is equivalent to excluded middle (over IZF)

jkingdon:
This is still a draft pull request because it relies on one needed theorem which isn't proved yet:

```
  djudom $p |- ( ( A ~<_ B /\ C ~<_ D )
      -> ( A |_| C ) ~<_ ( B |_| D ) ) $=
```

At first I thought this would be an easy consequence of http://us.metamath.org/mpeuni/undom.html and maybe it would be (although I'm not completely sure about the "easy" part), but iset.mm doesn't even have undom.

Any ideas about how to prove this? I'm assuming it can be done, as I suppose the disjoint unions would preclude most of the difficult situations which can happen without excluded middle.

As for the rest of the branch, it contains:

* ~~A proof of `EXMID <-> ~P 1o ~~ 2o`. This was mentioned in passing in one of the sources, and seemed worth proving especially because it was easy~~ merged
* ~~A few more notes for the missing theorems list in mmil.html~~ merged
* ~~A proof of `-. E. x x e. A -> A = (/)` which I was kind of surprised we didn't have already~~ merged as https://us.metamath.org/ileuni/notm0.html
* ~~Copying 1oex from set.mm~~ merged
* ~~A proof of `( 1o |_| 1o ) ~~ 2o` which is essentially the same as http://us.metamath.org/mpeuni/pm110.643.html~~ merged
* ~~``C e. V -> ( C e. A <-> ( inl ` C ) e. ( A |_| B ) )`` which turns out to be needed because the proof we are trying to formalize has a lot of steps about whether ``( inl ` (/) )`` is an element of a certain disjoint union.~~ merged
* ~~``( inl ` A ) =/= ( inr ` B )`` which is similar in spirit to http://us.metamath.org/ileuni/djuin.html but for individual elements, not entire images~~ merged
* The proof that excluded middle implies http://us.metamath.org/mpeuni/fodomr.html . This turned out to be relatively easy to prove by taking the set.mm proof and applying `EXMID` to the steps that needed excluded middle.
* The proof that fodomr implies excluded middle. This follows the proof from [PradicBrown2021] fairly closely.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -272,6 +272,7 @@ New usage of "dvelimALT" is discouraged (1 uses).
 New usage of "elirr" is discouraged (16 uses).
 New usage of "equsalh" is discouraged (5 uses).
 New usage of "eu3h" is discouraged (3 uses).
+New usage of "exmidfodomrlemrALT" is discouraged (0 uses).
 New usage of "fnexALT" is discouraged (0 uses).
 New usage of "foelrnOLD" is discouraged (0 uses).
 New usage of "hbs1" is discouraged (5 uses).
@@ -400,6 +401,7 @@ Proof modification of "difidALT" is discouraged (20 steps).
 Proof modification of "djulclALT" is discouraged (53 steps).
 Proof modification of "djurclALT" is discouraged (53 steps).
 Proof modification of "dvelimALT" is discouraged (158 steps).
+Proof modification of "exmidfodomrlemrALT" is discouraged (714 steps).
 Proof modification of "findset" is discouraged (96 steps).
 Proof modification of "fnexALT" is discouraged (111 steps).
 Proof modification of "foelrnOLD" is discouraged (58 steps).
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3131,8 +3131,7 @@ is Theorem 1 of [PradicBrown2021], p. 1.</TD>
 <TR>
   <TD>fodomr</TD>
   <TD><I>none</I></TD>
-  <TD>That fodomr implies excluded middle is Proposition 1.2
-  of [PradicBrown2021], p. 2</TD>
+  <TD>Equivalent to excluded middle per ~ exmidfodomr</TD>
 </TR>
 
 <TR>


benjub:
> djudom ``$p |- ( ( A ~<_ B /\ C ~<_ D ) -> ( A |_| C ) ~<_ ( B |_| D ) ) $=``

The antecedent gives injections f : A \to B and g : C \to D. From them, you have functions f1 := inl \circ f and g1 := inr \circ g, which are injective as compositions of injections. Now, the function $f1 \sqcup_{\mathrm{dom}} g1$ given by the universal property of the disjoint union, which @digama0 proposed to introduce as a definition, is injective because f1 and g1 are injective and their images are disjoint.

Namely, prove a lemma:
> if f : A \to X and g : B \to X are injective and their images are disjoint, then $f \sqcup_{\mathrm{dom}} g : A \sqcup B \to X$ is injective.

 Maybe this is an example of application of that function that is sufficient to justify introducing it as a definition ?

iset.mm
```diff
@@ -63822,6 +63846,21 @@ all reals (greatest real number) for which all positive reals are greater.
       KUM $.
   $}
 
+  ${
+    $d C x $.
+    $( Left biconditional closure of disjoint union.  (Contributed by Jim
+       Kingdon, 2-Jul-2022.) $)
+    djulclb $p |- ( C e. V -> ( C e. A <-> ( inl ` C ) e. ( A |_| B ) ) ) $=
```
benjub: Do we need the sethood antecedent ?
For the right implication, obviously not.
For the left implication, if C were a proper class, then ``(inl ` C)`` would be empty, and the empty set does not belong to a disjoint union, since `( A |_| B ) C_ ( _V X. _V )` (and even `C_ ( 2o X. ( A u. B ) )`, which I think @avekens proved). However, this proof looks non-intuitionistic.

iset.mm
```diff
@@ -62575,6 +62585,20 @@ the cancellation property (fifth hypothesis), show that the relation
       CVTYIXPABCWAWBWCWDVR $.
   $}
 
+  $( Excluded middle is equivalent to the power set of a singleton having two
```
benjub: "the power set of `1o`"

iset.mm
jkingdon: > Do we need the sethood antecedent ? For the right implication, obviously not. For the left implication, if C were a proper class, then `` (inl ` C) `` would be empty, and the empty set does not belong to a disjoint union, since `( A |_| B ) C_ ( _V X. _V )` (and even `C_ ( 2o X. ( A u. B ) )`, which I think @avekens proved). However, this proof looks non-intuitionistic.

Most of the reverse closure theorems use excluded middle in their proofs and if there is a way to prove them without it, we haven't found it yet. I don't know if we can make an argument analogous to https://github.com/metamath/set.mm/issues/664#issuecomment-444338796 which might convince ourselves whether it it is even possible in principle.

If any of this gets borrowed into set.mm the situation would be different there, but at least for now I'm only looking at iset.mm.

iset.mm
jkingdon: I suspect it doesn't matter whether you are looking at `1o` or a different singleton, but I'll take your suggested wording. There are cases where the comment should describe the result in terms of what it means rather than just repeating the formal theorem directly, but to say "a singleton" is, I suppose, a bit too far beyond what we actually prove.

jkingdon:
> The antecedent gives injections f : A \to B and g : C \to D. From them, you have functions f1 := inl \circ f and g1 := inr \circ g, which are injective as compositions of injections. Now, the function . . .

Yeah, that's the proof I was starting in on, and the difficulties you mention are more or less the ones I was hitting too.

You might be right about this being a proof which would be a whole lot easier to write given a definition for the function produced by the universal property. I suspect that without that definition the proof would need to repeat `ℎ:(𝐴 ⊔ 𝐵)⟶𝐶 ∧ (ℎ ∘ (inl ↾ 𝐴)) = 𝐹 ∧ (ℎ ∘ (inr ↾ 𝐵)) = 𝐺` a lot, which is possible but which is a fairly long expression.

iset.mm
avekens: It is easy to show that the empty set cannot be an element of a disjoint union, but to use this for the left implication, we would have to use pm2.61ddc (to get a contradiction if ``C e/ _V`` => ``( inl ` C ) = (/)`` => `` ( ( inl ` C ) e/ ( A |_| B )`` ). But this would require to show ``DECID C e. _V`` (which is not possible in iset.mm?).

avekens:
What about the failed checks?

iset.mm
benjub: It is a bit far indeed. When formalizing mathematics, the risk of errors moves from the proof itself (which is certain if we trust the kernel), to the interpretation of the formal statement. This is not purely theoretical, but is a real concern (I read a particularly acute version of this stance in a survey by Wiedijk). That's why I'm often careful about comments being precise enough.

benjub:
> Yeah, that's the proof I was starting in on, and the difficulties you mention are more or less the ones I was hitting too.

I didn't mention difficulties. Maybe I missed some ? I'm not saying formalizing it is not a hassle, but tell me if you see some difficulties I overlooked. And if you think that instead of saying "this is easy", I should just do it, then, fair enough, tell me and give me a few more days.

> the proof would need to repeat ℎ:(𝐴 ⊔ 𝐵)⟶𝐶 ∧ (ℎ ∘ (inl ↾ 𝐴)) = 𝐹 ∧ (ℎ ∘ (inr ↾ 𝐵)) = 𝐺 a lot, which is possible but which is a fairly long expression.

Probably three $e-hypotheses in deduction forms in the various lemmas ? Or maybe just add it as a definition (credits to @digama0) ? I can try either in the next few days if you want.

benjub:
> What about the failed checks?

@jkingdon posted it as a draft (see top-left gray icon) to allow preliminary comments, so I guess he'll do a global rewrap and regenerate the discouraged files at the end, when ready to be a genuine PR.

iset.mm
benjub: @avekens : yes, that is the proof I envisioned above, and it is probably non-intuitionistic. Even if we knew it is non-intuitionistic, this would not imply that no intuitionistic proof exists, of course. The comment referenced by @jkingdon above and the following discussion let me think that it is not provable, but I have no precise argument.

As for `DECID C e. _V` being provable, it would imply that `{ C } e. _V` is provable: either C is a set so {C} is a set, or it is not, so {C} is empty, so is a set. And the latter is thought to be not provable (see that same referenced discussion).

digama0: A simple proof that `DECID C e. _V` is unprovable is that `{ x | ph } e. _V <-> -. ph` (where `ph` does not depend on `x`), so it implies EM for negative propositions.

benjub: So ("forall C" DECID C e. _V) is equivalent to weak excluded middle. I guess we could define
```
$a |- ( WEXMID <-> A. x ( x C_ 1o -> DECID x = (/) ) $.
```
? And then, `( WEXMID -> DECID -. ph )` and `( WEXMID -> DECID A e. _V )` ?

digama0: I don't think it's an equivalence, or at least I don't see an obvious way to prove that `C e. _V` is a negative proposition so that weak EM will decide it.

benjub: Oh yes, you would need to find ph such that C = { x | ph } **where x does not occur in ph**, so `x e. C` will not do...

jkingdon:
> I didn't mention difficulties. Maybe I missed some ? I'm not saying formalizing it is not a hassle, but tell me if you see some difficulties I overlooked.

Fair enough, we'll rebrand them as hassles rather than difficulties (I didn't see anything which appeared to be insurmountable).

> > the proof would need to repeat ℎ:(𝐴 ⊔ 𝐵)⟶𝐶 ∧ (ℎ ∘ (inl ↾ 𝐴)) = 𝐹 ∧ (ℎ ∘ (inr ↾ 𝐵)) = 𝐺 a lot, which is possible but which is a fairly long expression.
> 
> Probably three $e-hypotheses in deduction forms in the various lemmas ? Or maybe just add it as a definition (credits to @digama0) ? I can try either in the next few days if you want.

Yeah, the three $e-hypotheses should work if the definition is still controversial. I don't object to the definition either, so your choice I guess unless there are other people to convince.

I'll accept your offer to formalize this. Just make a new pull request and then I'll update this one once it is merged. And next few days is no problem, I'm not piling up a bunch of work behind this.

iset.mm
jkingdon: I have put the above result about `C e. _V` into pull request form at https://github.com/metamath/set.mm/pull/2660

iset.mm
```diff
@@ -64371,6 +64371,12 @@ all reals (greatest real number) for which all positive reals are greater.
       GYQVEUUDYJYSFYGYQYIVHVFUUDYMUUAGYGYQYLVHVFVIYBYCYOEYDYE $.
   $}
 
+  $( Dominance law for disjoint union.  (Contributed by Jim Kingdon,
+     1-Jul-2022.) $)
+  djudom $p |- ( ( A ~<_ B /\ C ~<_ D )
+      -> ( A |_| C ) ~<_ ( B |_| D ) ) $=
+    (  ) ? $.
+
```
benjub: I'm getting close to the crucial lemma for this (injectivity of the unique H : A |_| B -> X when both F : A -> X and G : B -> X are injective and their ranges are disjoint)...

jkingdon: Ooh, thanks for the status report.

After a (less than a day) break that I took for myself, I'm back on the case of what to do about restricted `inl` from the https://github.com/metamath/set.mm/pull/2669#discussion_r916286944 thread. Hopefully this isn't hugely hairy to figure out, but the exmidfodomr proof does rely on the theorems being discussed over there.

iset.mm
```diff
@@ -65455,6 +65455,104 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
...
   $}
 
+  ${
+    $d A x $.  $d B x $.  $d ph x $.
+    exmidfodomrlemeldju.a $e |- ( ph -> A C_ 1o ) $.
+    exmidfodomrlemeldju.el $e |- ( ph -> B e. ( A |_| 1o ) ) $.
+    $( Lemma for ~ exmidfodomr .  A variant of ~ djur .  (Contributed by Jim
+       Kingdon, 2-Jul-2022.) $)
+    exmidfodomrlemeldju $p |- ( ph
+        -> ( B = ( inl ` (/) ) \/ B = ( inr ` (/) ) ) ) $=
```
benjub: Again, I would write
```
-> ( B = ( ( inl |` A ) ` (/) ) \/ B = ( ( inr |` 1o ) ` (/) ) )
```
since I think the needed theorems with the restrictions are all available, and a consistent use of the restricted injections is possible. But this is only a lemma, and maybe the proofs of the lemmas up to the main theorem would be more cumbersome (even though clearer, to me) ?

jkingdon: Already on this branch as `exmidfodomrlemeldjuN` (excerpt below).

I had in mind keeping both versions around (because I agree with "cumbersome" on the one side but also want to find a good way to deal with the other being "clearer" for those who find that). However, I'm not sure I had great ideas about naming. The `N` convention was intended to be temporary. For the case of `exmidfodomrlemrN` it could be an `ALT` because it proves the same theorem as `exmidfodomrlemr` but with a different proof. However, as far as I know we don't use the `ALT` convention where the theorem being proved is different. So maybe we include `res` in the names of the versions with the restrictions? I mean, sometimes in this sort of situation we'd put one of the choices in a mathbox but to do that we'd need to decide which one goes in main.

```
  ${
    $d A x $.  $d B x $.  $d ph x $.
    exmidfodomrlemeldjuN.a $e |- ( ph -> A C_ 1o ) $.
    exmidfodomrlemeldjuN.el $e |- ( ph -> B e. ( A |_| 1o ) ) $.
    $( Lemma for ~ exmidfodomr .  A variant of ~ eldju .  (Contributed by Jim
       Kingdon, 9-Jul-2022.) $)
    exmidfodomrlemeldjuN $p |- ( ph
        -> ( ( (/) e. A /\ B = ( ( inl |` A ) ` (/) ) )
          \/ B = ( ( inr |` 1o ) ` (/) ) ) ) $=
```

benjub: Ah, missed it ! Great minds think alike ! (Now I understand why the cumbersome `(/) e. A` is needed, simply because `(/)` was chosen to represent an undefined value.)

Don't worry about the label.  I chose to suffix an "r" for "restriction", for lack of a better option, but probably we won't keep both in Main, and we can change that later.

jkingdon: > Now I understand why the cumbersome `(/) e. A` is needed, simply because `(/)` was chosen to represent an undefined value.

I don't think it has anything to do with undefined values here. If we were doing the proof with `{ 0 , 1 }` rather than `{ (/) , 1o }` it would be `0` not `(/)`. I described this at https://github.com/metamath/set.mm/pull/2669#discussion_r917289132 and it is why avoiding `djulclb` in the `exmidfodomrlemrN` proof is not just a matter of slotting in a slightly different variant of `djulclb` - you need to carry around `(/) e. A` to implement the suggestion of "this is an argument for proving C e. A first" (in the words of https://github.com/metamath/set.mm/pull/2669#discussion_r917302632 )

benjub: Ah yes, you're right, I remember.

jkingdon:
@benjub Are you still working on proving 

```
djudom $p |- ( ( A ~<_ B /\ C ~<_ D )
      -> ( A |_| C ) ~<_ ( B |_| D ) ) $=
```

? If you have given up I can take it on.

benjub:
> @benjub Are you still working on proving
> 
> ```
> djudom $p |- ( ( A ~<_ B /\ C ~<_ D )
>       -> ( A |_| C ) ~<_ ( B |_| D ) ) $=
> ```
> 
> ? If you have given up I can take it on.

Sorry, I should have been clearer that I stopped at casef1 (the "case" function from the disjoint union is injective). From there, the proof is not too long; you already know it but I'm writing it here for the record: the two antecedents give f and g injective, and then one considers
```
case((inl |` B) circ f, (inr |` D) circ g)
```
which is injective as a "case" of two compositions of injective functions.  I haven't done anything with ~<_ so I feared it would take me an exagerate amount of time.

jkingdon:
> From there, the proof is not too long . . . I'm writing it here for the record

Thanks, that formalized without too much trouble.

This pull request is now passing the automated checks and is ready for review (and merging if approved).

iset.mm
```diff
@@ -65414,6 +65414,27 @@ all reals (greatest real number) for which all positive reals are greater.
       PLZJVRVTXEBCVCVEVKVPVFVGQR $.
   $}
 
+  ${
+    $d A f g h $.  $d B f g h $.  $d C f g h $.  $d D f g h $.
+    $( Dominance law for disjoint union.  (Contributed by BJ and Jim Kingdon,
+       25-Jul-2022.) $)
```
benjub, line 65420: The key step is casef1, which I contributed, but the proof of djudom is really "contributed by JK". Or you can leave it as is. As you prefer.

iset.mm
```diff
@@ -65764,6 +65785,167 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
...
+      wo cinl cinr sselda el1o eqeltrrd biantrurd bitrd rexlimdva imp orcd olcd
+      cv cdju eldju mpjaodan ) ACFULZUABGZHZIZFBJZKBLZCKUQHZIZMZCKUBNGZHZIZTCUP
+      VEHZIZFNJZAUTMVDVGAUTVDAUSVDFBAUPBLZMZUSVDVLUSVCVDVLURVBCVLUPKUQVLUPNLZUP
+      KIZABNUPDUCUPUDZOZPQVLVAVCVLUPKBVPAVKRUEUFUGSUHUIUJAVJMVGVDAVJVGAVIVGFNAV
+      MMZVIVGVQVHVFCVQUPKVEVQVMVNAVMRVOOPQSUHUIUKACBNUMLUTVJTEFBNCUNOUO $.
+  $}
+
+  ${
+    $d f g x z $.  $d f g y z $.  $d g u x $.  $d g v x $.  $d w z $.
+    $( Excluded middle implies the existence of a function onto an inhabited
```
benjub, line 65822: an --> any ?

Plus, I'm not sure if it's me or if the grammar of these sentences is weird. A correct formulation would be much heavier, so I guess this is ok.

benjub (approved):

iset.mm
jkingdon: I wasn't basing it on casef1 because that has its own contributed line. I was basing it on you having written the informal proof in a comment here.

I guess I'll go with the simpler Contributed line, but still you should know that your comment, even though you thought you were just saying an obvious thing which I already knew, was very helpful. Without it I would have had to go back and read old messages, remember what `case` has to do with `djudom`, etc.

iset.mm
jkingdon: The grammar of these sentences is definitely weird. I tried again starting from the set.mm comment for fodomr and came up with:

```
Excluded middle is equivalent to the existence of 
a mapping from any set onto any inhabited set that it
dominates
```

which only changes a few words but which I think is better.

benjub: Perfect !

benjub (approved):

avekens (approved):

PR #2660, opened 2022-07-04
Two theorems about set existence in iset.mm

jkingdon:
The first is showing that decidability of the existence of an arbitrary class (as often used in set.mm to combine set and proper class cases) implies decidability of an arbitrary negated proposition, as discussed in https://github.com/metamath/set.mm/pull/2659#discussion_r912535856 and the rest of that thread.

The second is `-. -. { A } e. _V ` as proposed at https://github.com/metamath/set.mm/issues/664#issuecomment-1173175167 . Along with this I also put a bunch of text from https://github.com/metamath/set.mm/issues/664#issuecomment-444341096 into mmil.html because that seems as good as we're likely to get for the near term in terms of showing that we won't be able to prove `{ A } e. _V ` without conditions.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1779,7 +1779,28 @@ set exists or it does not"</TD>
 <TR>
 <TD>snex</TD>
 <TD>~ snexg , ~ snex </TD>
-<TD>The iset.mm version of ~ snex has an additional hypothesis</TD>
+<TD><P>The iset.mm version of ~ snex has an additional hypothesis.</P>
+
+<P>We conjecture that the set.mm snex ( ` { A } e. _V ` with
+no condition on whether ` A ` is a set) is not provable.</P>
+
+<P>The axioms of set theory allow us to construct rank levels from others, but
+not to construct something from nothing (except the 0 rank and stuff you can
+bootstrap from there). A class provides no "data" about where it lives,
+because it is spread out over the whole universe - you only get the ability
+to ask yes-no(-maybe) questions about set membership in the class. An
+assertion that a class exists is a piece of data that gives you a bound
+on the rank of this set, which can be used to build other existing things
+like unions and powersets of this class and so on. Anything with unbounded
+rank cannot be proven to exist.</P>
+
+<P>So snex, which starts from an arbitrary class and produces evidence
+that { A } exists, cannot be provable, because the bound here can't
+depend on A and cannot be upper bounded by anything independent of
+A either - there are singletons at every rank.</P>
+
+<P>However, we can refute a singleton being a proper class - see
+~ notnotsnex .</P></TD>
 </TR>
 
 <TR>


avekens (approved):

iset.mm
```diff
@@ -38820,6 +38826,22 @@ class of all ordinal numbers to be a set (so instead it is a proper
...
+  ${
+    $d ph x y $.
+    dcextest.ex $e |- DECID { x | ph } e. _V $.
+    $( If the proposition ` A e. _V ` is decidable for any class, any negated
+       proposition is decidable.  Here we set ` A ` to be ` { x | ph } ` (since
+       we have no way to express "for any class A", we have to write the
+       theorem with a particular ` A ` ).  (Contributed by Jim Kingdon,
+       3-Jul-2022.) $)
+    dcextest $p |- ( -. ph \/ -. -. ph ) $=
+      ( vy cab cvv wcel wn wo wdc exmiddc ax-mp elirr weq df-v wb equid pm5.1im
```
avekens, line 38838: why is ~elirr used here, although its usage is discouraged?

jkingdon: It is only discouraged for ordinals, as described at http://us.metamath.org/ileuni/elirr.html we'd like to use http://us.metamath.org/ileuni/ordirr.html then.

benjub: I don't understand why you don't use "`DECID -. ph`" for the conclusion.
For the comment, I would write:
> If it is decidable whether `{ x | ph }` is a set, then `-. ph` is decidable (where `x` does not occur in `ph`). From this fact, we can deduce (outside the formal system, since we cannot quantify over classes) that if it is decidable whether any class is a set, then "weak excluded middle" (that is, any negated proposition `-. ph` is decidable) holds.

digama0: @jkingdon It is discouraged everywhere, because it depends on the foundation axiom and you can usually prove equivalent results without it. How is it being used in this proof?

jkingdon: > @jkingdon It is discouraged everywhere, because it depends on the foundation axiom and you can usually prove equivalent results without it.

LOL, I won't belabor the point of who put in the discouraged tag and based on what reasoning, but let's talk about how foundation is being used here rather than procedure.

> How is it being used in this proof?

Same way it is being used in http://us.metamath.org/ileuni/ruv.html and http://us.metamath.org/mpeuni/ruv.html

I didn't see a way to adapt the http://us.metamath.org/ileuni/ru.html proof or another one I thought of to this situation, but I suppose this probably can be proved without foundation. Maybe I could get a bit of help from people who want to reduce the use of foundation (either in terms of this particular proof, or the 14 existing usages)?

digama0: Oops, I didn't realize it had an actual discouraged tag. I meant it in the colloquial sense. It's possible that the tag was added to avoid introducing axioms? But we already have the `/no_new_axioms` flag for that. :shrug: 

The way you can avoid it here is by using vprc instead, which is proved not by showing that the russell class is *equal* to V but rather that it is a *subset* of V and hence if V is a set then the russell class is too. That is, under the assumption `ph`, `{ x | ph } = _V e/ _V`, so by contrapositive `{ x | ph } e. _V` implies `-. ph`.

By the way, on the subject of snex and constant classes, one other counterintuitive fact I noticed is that `{ { x | ph } } e. _V`, because it is unconditionally a subset of `1o`, because any element of the singleton `y = { x | ph }` implies that `-. ph` and hence `y = (/)`.

jkingdon: > It's possible that the tag was added to avoid introducing axioms? But we already have the `/no_new_axioms` flag for that. shrug

`/no_new_axioms` won't (currently) help (in fact, I added the discouraged tag a while ago when I noticed the minimizer removing an `ordirr` which I had carefully inserted), To explain I'll just quote the comment of http://us.metamath.org/ileuni/elirr.html :

> The reason that this theorem is marked as discouraged is a bit subtle. If we wanted to reduce usage of [ax-setind](http://us.metamath.org/ileuni/ax-setind.html) 4309, we could redefine Ord 𝐴 ([df-iord](http://us.metamath.org/ileuni/df-iord.html) 4150) to also require E Fr 𝐴 ([df-frind](http://us.metamath.org/ileuni/df-frind.html) 4116) and in that case any theorem related to irreflexivity of ordinals could use [ordirr](http://us.metamath.org/ileuni/ordirr.html) 4314 (which under that definition would presumably not need [ax-setind](http://us.metamath.org/ileuni/ax-setind.html) 4309 to prove it). But since ordinals have not yet been defined that way, we cannot rely on the "don't add additional axiom use" feature of the minimizer to get theorems to use [ordirr](http://us.metamath.org/ileuni/ordirr.html) 4314. To encourage [ordirr](http://us.metamath.org/ileuni/ordirr.html) 4314 when possible, we mark this theorem as discouraged.

_going back to the message I'm responding to:_
> The way you can avoid it here is by using vprc instead

Ah, cool. That's easier than I had feared. Done.

> By the way, on the subject of snex and constant classes, one other counterintuitive fact I noticed is that `{ { x | ph } } e. _V`, because it is unconditionally a subset of `1o`, because any element of the singleton `y = { x | ph }` implies that `-. ph` and hence `y = (/)`.

This is an odd little theorem, and I'll agree with counterintuitive. Which I guess is an argument for formalizing this proof. Done.

jkingdon: > I don't understand why you don't use "`DECID -. ph`" for the conclusion.

Well, I could offer an impassioned defence about how overusing definitions increases the cognitive load especially for people new to this material, who are least able to bear it. However, I think that might fall kind of flat in a case where the hypothesis of the very same theorem uses `DECID` so, OK, there seems to be little reason to avoid it here. Changed.

> For the comment, I would write:
> 
> > If it is decidable whether `{ x | ph }` is a set, then `-. ph` is decidable (where `x` does not occur in `ph`). From this fact, we can deduce (outside the formal system, since we cannot quantify over classes) that if it is decidable whether any class is a set, then "weak excluded middle" (that is, any negated proposition `-. ph` is decidable) holds.

I'm not in love with either that wording or the one I had, but I guess this looks good (for example, defines "weak excluded middle" rather than assuming people will know/remember what that is), so I'l use it. Wording adopted.

jkingdon:
@benjub @digama0 Ready for re-review.

iset.mm
benjub: > overusing definitions increases the cognitive load especially for people new to this material, who are least able to bear it

I think this is an important concern. But to me this is more of an argument to introduce a definition only if it's "really" needed, but once it is introduced, we might as well use it without being coy (there is even an argument to say that if we use it only in certain cases, then it will remain obscure, whereas if it's systematically used, the reader will become familiar with it).

> I'm not in love with either that wording or the one I had, but I guess this looks good (for example, defines "weak excluded middle" rather than assuming people will know/remember what that is), so I'l use it. Wording adopted.

Sometimes my English grammar is a bit rusty, so please correct it when necessary. As for the comment itself, I prefer to first state precisely what the formal theorem proves, and then digressing about what metatheorem we can deduce, rather than the opposite ("we want to prove that, but we can't within the formal system, so we prove that instead").  This of course is only valid for "small units" (in an article, or on the home page for ileuni, it is generally best to begin with an informal explanation).

iset.mm
```diff
@@ -38820,6 +38820,24 @@ class of all ordinal numbers to be a set (so instead it is a proper
...
   $}
 
+  ${
+    $d ph x y $.
+    dcextest.ex $e |- DECID { x | ph } e. _V $.
+    $( If it is decidable whether ` { x | ph } ` is a set, then ` -. ph ` is
+       decidable (where ` x ` does not occur in ` ph ` ).  From this fact, we
+       can deduce (outside the formal system, since we cannot quantify over
+       classes) that if it is decidable whether any class is a set, then "weak
+       excluded middle" (that is, any negated proposition -. ph is decidable)
```
benjub, line 38830: backquotes around -. ph

**edit:** this comment was meant to go after the line "(that is, any negated proposition -. ph is decidable)"

iset.mm
```diff
@@ -63027,6 +63027,20 @@ would need to be strengthened (for example, to imply included middle) to
     ( wcel c1o cen wbr csn wceq en1eqsn ex ensn1g breq1 syl5ibrcom impbid ) ABC
     ZBDEFZBAGZHZOPRABIJOPRQDEFABKBQDELMN $.
 
+  ${
+    $d ph x y $.
+    $( A case where the antecedent of ~ snexg is not needed.  The class
+       ` { x | ph } ` is from ~ dcextest .  (Contributed by Mario Carneiro and
+       Jim Kingdon, 4-Jul-2022.) $)
```
benjub, line 63034: >  The class ` { x | ph } ` is from ~ dcextest .

What does it mean ?

benjub (approved):
Two comments (a typo and a genuine question), and ready to merge after that.

iset.mm
jkingdon: > backquotes around -. ph

Added.

iset.mm
jkingdon: I'm not sure I have an answer for that.

For example, in the case of ` { x e. 1o | ph } ` I think our "official" name for it is "Mario's favorite ordinal". But what does it mean? I guess I sort of think of it as "a class which is in a superposition of (/) and 1o" but I'm not sure I'd use that kind of language in trying to describe it to someone else.

benjub: What I mean is that I cannot make sense of the sentence "The class { x | ph } is from ~ dcextest ." in the comment. Maybe a coment could be:
> The singleton formed on the class ` { x | ph } ` where ` x ` does not occur in ` ph ` is a set.  This is in contrast to the general situation of ~  snexg .

(or for the second sentence, use your version).

iset.mm
benjub: > But to me this is more of an argument to introduce a definition only if it's "really" needed,

To give an example, the definition of "testable" is not really needed, since `TEST ph` is simply defined as `DECID -. ph` (so even when you substitute a huge expression for `ph`), there is no gain in having it (which is why it was recently removed).  Not unrelatedly, the meaning of "testable" is certainly much less known than that of "decidable". (I just realized that the "test" in the label "dcextest" was referring to testability of ph).

PR #2662, opened 2022-07-04
add free_var annotation to wssb

digama0:
This came up when testing the HOL translation of the latest set.mm. The MM -> MM0 translator has to make heuristic inferences about whether set variables in a notation are binders or free variables, and this applies the same fix to `df-ssb` as is used in `df-sb` so that it will correctly infer that `t` is free.


jkingdon (approved):
This looks correct by analogy to the similar annotation on http://us.metamath.org/mpeuni/wsb.html

Incidentally, where do we document `free_var` and `free_var_in`? I can sort of see what is going on by looking at the examples and what few comments are there, but am not sure I'm drawing the correct inferences.

digama0:
There isn't any real documentation for it. We should definitely start a page on $j extensions and documentation for it.

The simplest way to explain the effect of this command is to point at the result of the MM0 translation. Without the annotation, it gets translated like this:
```
term wssb {t x2: setvar} (ph: wff t x2): wff;
```
and with it the result is:
```
term wssb (t: setvar) {x2: setvar} (ph: wff x2): wff;
```
When translated to HOL, the first one has a type of
```
term wssb: (setvar -> setvar -> wff) -> wff;
```
and the second one is
```
term wssb: setvar -> (setvar -> wff) -> wff;
```
In other words, the `t` is treated as binding occurrences of `t` in `ph` in the first case and as a free variable in the second case. The first approach with the binder is explicitly disprovable with respect to the definitional axiom df-ssb, since t is a free variable on the right. Since the translation does not look ahead to this definitional axiom, it can't tell the difference between wssb, which binds `x` but not `t`, and copab, which binds both `x` and `y` in `ph`.

jkingdon:
> There isn't any real documentation for it. We should definitely start a page on $j extensions and documentation for it.

I've tried to get the ball rolling a bit at https://github.com/metamath/set.mm/issues/2664 but I sort of ran out of time at least for the moment, so there is plenty more to do on that issue.

benjub (approved):

PR #2663, opened 2022-07-04
Disjoint union: staple lemmas

benjub:
Add a few lemmas to shorten existing theorems:
* general functions: add f1ff1, shorten f1resf1
* add resflem, which is used later
* add 1oex, which is used often
* disjoint unions: use lemmas since many proofs are very similar for the left and right injections, so the exposition is both clearer and shorter. Note: even though it does not reduce size, I would also like to move to Main ~ djucllem, which follows the same idea. I think this would make the exposition clearer and remove this feeling of proving things twice. @jkingdon @avekens ?

As for credits, I credited the lemmas to me. Now I'm thinking that this may be pushing it a bit too much, since I drew inspiration from your proofs. Tell me what you think and I'll be happy to make changes. Also, since the results are recent, I did the shortenings "silently".

There are other shortenings that can be done, in particular to djuun, where the inclusion `C_` is a quick consequence of in(l|r)resf. I'll do it soon.

(All of this is in iset. We'll have to update set.mm then...)

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -397,7 +397,7 @@ Proof modification of "bj-zfpair2" is discouraged (80 steps).
 Proof modification of "brab2ga" is discouraged (79 steps).
 Proof modification of "difidALT" is discouraged (20 steps).
 Proof modification of "djulclALT" is discouraged (51 steps).
-Proof modification of "djurclALT" is discouraged (54 steps).
+Proof modification of "djurclALT" is discouraged (51 steps).
 Proof modification of "dvelimALT" is discouraged (158 steps).
 Proof modification of "findset" is discouraged (96 steps).
 Proof modification of "fnexALT" is discouraged (111 steps).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -266,6 +266,8 @@ New usage of "df-pnf" is discouraged (3 uses).
 New usage of "df-tru" is discouraged (1 uses).
 New usage of "dftru2" is discouraged (0 uses).
 New usage of "difidALT" is discouraged (0 uses).
+New usage of "djulclALT" is discouraged (0 uses).
+New usage of "djurclALT" is discouraged (0 uses).
 New usage of "dvelimALT" is discouraged (1 uses).
 New usage of "elirr" is discouraged (16 uses).
 New usage of "equsalh" is discouraged (5 uses).
@@ -394,6 +396,8 @@ Proof modification of "bj-vprc" is discouraged (53 steps).
 Proof modification of "bj-zfpair2" is discouraged (80 steps).
 Proof modification of "brab2ga" is discouraged (79 steps).
 Proof modification of "difidALT" is discouraged (20 steps).
+Proof modification of "djulclALT" is discouraged (51 steps).
+Proof modification of "djurclALT" is discouraged (54 steps).
 Proof modification of "dvelimALT" is discouraged (158 steps).
 Proof modification of "findset" is discouraged (96 steps).
 Proof modification of "fnexALT" is discouraged (111 steps).


iset.mm
```diff
@@ -48367,6 +48373,22 @@ We use their notation ("onto" under the arrow).  (Contributed by NM,
...
   $}
 
+  ${
+    $d A x $.  $d ph x $.  $d F x $.  $d Y x $.
+    resflem.1 $e |- ( ph -> F : V --> X ) $.
+    resflem.2 $e |- ( ph -> A C_ V ) $.
+    resflem.3 $e |- ( ( ph /\ x e. A ) -> ( F ` x ) e. Y ) $.
+    $( A lemma to bound the range of a restriction.  The conclusion would also
+       hold with ` ( X i^i Y ) ` in place of ` Y ` but in applications, it is
+       generally easier to prove, if needed, the third hypothesis with
```
jkingdon, line 48383: I'm not following this wording. Are the two sets in `` ` ( X i^i Y ) ` in place of ` Y ` `` reversed?

Another option would be to replace the entire text from "in applications" to "in place of" with something shorter like "this form is likely to be more convenient".

Your choice.

iset.mm
```diff
@@ -57966,6 +57988,10 @@ defined for all sets (being defined for all ordinals might be enough if
   1on $p |- 1o e. On $=
     ( c1o c0 csuc con0 df-1o 0elon onsuci eqeltri ) ABCDEBFGH $.
 
+  $( Ordinal 1 is a set.  (Contributed by BJ, 4-Jul-2022.) $)
```
jkingdon, line 57991: A good change - I'll update #2659 accordingly assuming this is merged first.

iset.mm
```diff
@@ -63822,36 +63848,46 @@ all reals (greatest real number) for which all positive reals are greater.
...
-      bicomd f1o2d trud ) CDEZCFZGHIABCVFDAQZJZBQZKLZGAMVGCNZVHVFNZIDVENZVKVLDO
-      NVMRDOPUADVGVECUBUCSVIVFNZVJCNIVIVECUDSVNVGVJTZVIVHTZUJIVKVNVPVOVNVHVITVH
-      DVJJZTZVPVOVNVIVQVHVNVIVIUELZVJJVQVIVECUFVNVSDVJVNVSVENVSDTVIVECUGVSDUHUI
-      UKULUMVHVIUNVRDDTVODUODVGDVJDORUPAUQURUSUTVBVAVCVD $.
+    $d X x y $.
+    djuf1olem.1 $e |- X e. _V $.
+    djuf1olem.2 $e |- F = ( x e. _V |-> <. X , x >. ) $.
+    $( Lemma for ~ djulf1o and ~ djurf1o .  (Contributed by BJ and Jim Kingdon,
+       4-Jul-2022.) $)
+    djuf1olem $p |- F : _V -1-1-onto-> ( { X } X. _V ) $=
```
jkingdon, line 63863: As stated, the advantages of this are to remove the feeling of proving things twice.

The disadvantage is that reading the proof means following one more layer of indirection.

I guess if we went by the "rule of three" from https://wou.edu/las/cs/csclasses/cs161/Lectures/rulesofthumb.html we'd say this is only two, not three, occurences.

But this sort of thing should be (in my opinion) a bit case by case and in this situation I don't really have a strong opinion one way or the other.

jkingdon (approved):

jkingdon:
> As for credits, I credited the lemmas to me. Now I'm thinking that this may be pushing it a bit too much, since I drew inspiration from your proofs. Tell me what you think and I'll be happy to make changes. Also, since the results are recent, I did the shortenings "silently".

The credits are fine with me as they are. I guess from my point of view trying to assess whether a new proof borrows enough from an existing one to note it in the Contributed lines hits a "shoulders of giants" problem pretty fast, in which I never quite know whether to list a long line of people back to Aristotle, or a pruned version, or what.

I do try to be a bit more verbose in the git logs because that seems like it is easier to put nuance there ("based on a proof X, with Y and Z changed") and that perhaps it is the place to focus on historical information anyway.

iset.mm
```diff
@@ -46093,6 +46093,12 @@ We use their notation ("onto" under the arrow).  (Contributed by NM,
     wfn ) ABDEZDFCGZHZACDIZDJKZACDEUBDASZUAUCTUEUAABDLMTUANACDOPTUDUATABDIUDABD
     QRMACDQP $.
 
+  $( If a function is one-to-one from A to B and is also a function from A to
+     C, then it is a one-to-one function from A to C. (Contributed by BJ,
+     4-Jul-2022.) $)
+  f1ff1 $p |- ( ( F : A -1-1-> B /\ F : A --> C ) -> F : A -1-1-> C ) $=
+    ( wf wf1 crn wss frn f1ssr sylan2 ) ACDEABDFDGCHACDFACDIABCDJK $.
+
   $( A function that is one-to-one is also one-to-one on some aubset of its
```
avekens, line 46102: Typo aubset -> subset - or even better -> subclass?

iset.mm
avekens: I also had problems to understand the meaning of the comment first. Suggestion: "The conclusion would also hold with ` ( X i^i Y ) ` in place of ` Y `. If needed, however, ` Y ` can be replaced generally by ` ( X i^i Y ) ` (also in the third hypothesis).

iset.mm
avekens: This is already available in set.mm!

iset.mm
avekens: I like these lemmas which extract the essentials from the theorems/proofs for inl/inr. Not only they shorten the proofs in sum, and provide the essentials in a more explicit way, they may also be used in other contexts.

avekens (changes requested):

avekens:
> > As for credits, I credited the lemmas to me. Now I'm thinking that this may be pushing it a bit too much, since I drew inspiration from your proofs. Tell me what you think and I'll be happy to make changes. Also, since the results are recent, I did the shortenings "silently".
> 
> The credits are fine with me as they are. I guess from my point of view trying to assess whether a new proof borrows enough from an existing one to note it in the Contributed lines hits a "shoulders of giants" problem pretty fast, in which I never quite know whether to list a long line of people back to Aristotle, or a pruned version, or what.

I also have no problems with the credits.

> I do try to be a bit more verbose in the git logs because that seems like it is easier to put nuance there ("based on a proof X, with Y and Z changed") and that perhaps it is the place to focus on historical information anyway.

That's a different point: how many commits should be performed (which can be commented separately)? One for each theorem? I used to commit my changes belonging to one issue/topic as a whole block...

benjub:
> I do try to be a bit more verbose in the git logs because that seems like it is easier to put nuance there ("based on a proof X, with Y and Z changed") and that perhaps it is the place to focus on historical information anyway.

So I guess you often look at git logs ? How do you do it ? Things like `$ git log -5` ? Or are there better ways ?

iset.mm
benjub: Reworded following more or less @avekens. I hope it's clearer.

iset.mm
benjub: Thanks for the link: I read with interest all rules. It looks like the rule is "if you write it at least thrice, then write a lemma/subprogram". But the converse is not stated ! After all, sometimes, even a part of a proof or code that is not repeated deserves its own lemma, for reasons other than repetition, like @avekens explained.

So I guess I will keep the lemmas already in Main, which offer a global shortening. However, I will not move djucllem to Main, since it does not offer a global shortening. (I might keep it in my mathbox, and probably delete it in a while).

iset.mm
benjub: Great, let me credit the same person in iset.mm ... wait... nevermind ! (I had forgotten about it)

iset.mm
benjub: done

benjub:
> That's a different point: how many commits should be performed (which can be commented separately)? One for each theorem? I used to commit my changes belonging to one issue/topic as a whole block...

Do as you prefer, but I think having several smaller commits, even within a single PR (grouped in meaningful ways), can make it easier to review (the reviewers can choose to see the global changes of the PR, or the individual changes), permit to revert individual commits, in case only some changes are accepted, which also makes error recoveries easier. Note that if you committed and then make a small change and you want to add to that last commit, you can still do `$ git commit -am --amend "new message"`.

iset.mm
jkingdon: > It looks like the rule is "if you write it at least thrice, then write a lemma/subprogram". But the converse is not stated !

Well https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming) states the converse but to treat these as rigid rules kind of misses the point (in my opinion). As important as "don't remove duplication if there are only two copies" is a sense for why you might not (for example, you might generalize in a less useful way if you don't have enough sense for how it might be used).

jkingdon:
> So I guess you often look at git logs ? How do you do it ? Things like `$ git log -5` ? Or are there better ways ?

In the case of metamath, I guess the most common is "git log" return followed by slash followed by the name of a theorem and return. Then "n" until I've had my fill.

In some of the smaller files, `git blame mmil.raw.html` (from the command line, or https://github.com/metamath/set.mm/blame/develop/mmil.raw.html for the github view) can also be a nice way to find a log entry). That makes me curious about the larger files. On my machine, I can get a git blame for iset.mm in  about a minute, and for set.mm in 16 minutes (there is a nice progress display that it updates as it runs). So this isn't quite as hopeless as I had assumed, although it is still more useful for the small files.

Let me not overstate this. I don't look at them super often. But from time to time, yes.

jkingdon (approved):

avekens (approved):

PR #2665, opened 2022-07-05
shorten nfimt

wlammen:
Use 19.35 / 19.38 earlier.  Use nfrd instead of df-nf.  This saves 7 proof bytes and a couple of symbols on the display. Sorry, @jkingdon, I saw your approval too late. We were almost simultaneous (3 min gap)

Using sylan9 instead of syl9 directly leads to nfimt2.  This has the advantage, that in the final step nfd instead of the couple df-nf along with syl6ibr can be used.  nfimt is then proved from nfimt2 by ex. In total the couple nfimt and nfimt2 shrinks by more bytes and a proof line.  Disadvantage: df-an creeps in.  Some may consider this less beautiful.  Maybe one should rather start with nfimd.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16883,6 +16883,7 @@ New usage of "nfim1OLD" is discouraged (1 uses).
 New usage of "nfimOLD" is discouraged (1 uses).
 New usage of "nfimdOLD" is discouraged (3 uses).
 New usage of "nfimdOLDOLD" is discouraged (0 uses).
+New usage of "nfimtOLD" is discouraged (0 uses).
 New usage of "nfnOLD" is discouraged (2 uses).
 New usage of "nfnanOLD" is discouraged (0 uses).
 New usage of "nfndOLD" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


jkingdon (approved):

jkingdon (approved):

jkingdon:
> I saw your approval too late. We were almost simultaneous (3 min gap)

No worries. Still looks good with the recent changes.

avekens (approved):

benjub:
As for the current proofs of nfimd and nfim, currently using nfimt2: did you check whether using nfimt shorten either ?

I indeed like the implication-only (i.e., curried) version, but the uncurried version works better with the deduction-style technique of proof. Do as you prefer. If you think nfimt is useless, then maybe it should be removed (e.g. placed in my mathbox before its only usage, and I'll delete both theorems later), and relabel nfimt2 to nfimt.

wlammen:
I'll look into this later again

jkingdon:
> If you think nfimt is useless, then maybe it should be removed (e.g. placed in my mathbox before its only usage, and I'll delete both theorems later), and relabel nfimt2

Makes sense to me, just in the sense that we usually go with `( ph /\ ps ) -> ch ` rather than `ph -> ( ps -> ch )`. I don't start with a big a priori preference for one or the other, but we have a lot of convenience theorems for the former, and doesn't seem to be like we need both in cases like nfimt.

PR #2667, opened 2022-07-06
Djuun

benjub:
Mainly:
* big shortening of djuun
* restructuring, with three subsections
* theorems with label ending by 'N' are meant to ultimately replace the similar labels without an 'N' (this entails minimal modifications of a few proofs). This globally reduces proof size (not by a small margin), and it states most theorems abount injections using e.g. ``( inl |` A )`` instead of ``inl``. As I explain in a head comment, `inl` has to be thought as a "template" for injections, and the meaningful objects are actually the functions ``( inl |` A ) : A --> (A |_| B )``

I would like to merge this now to ease further modifications. However, if you think it's not mature enough to be merged, then that's fine and I'll remove all the xxxN/xxx before merging.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -396,8 +396,8 @@ Proof modification of "bj-vprc" is discouraged (53 steps).
 Proof modification of "bj-zfpair2" is discouraged (80 steps).
 Proof modification of "brab2ga" is discouraged (79 steps).
 Proof modification of "difidALT" is discouraged (20 steps).
-Proof modification of "djulclALT" is discouraged (51 steps).
-Proof modification of "djurclALT" is discouraged (51 steps).
+Proof modification of "djulclALT" is discouraged (53 steps).
+Proof modification of "djurclALT" is discouraged (53 steps).
 Proof modification of "dvelimALT" is discouraged (158 steps).
 Proof modification of "findset" is discouraged (96 steps).
 Proof modification of "fnexALT" is discouraged (111 steps).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -273,6 +273,7 @@ New usage of "elirr" is discouraged (16 uses).
 New usage of "equsalh" is discouraged (5 uses).
 New usage of "eu3h" is discouraged (3 uses).
 New usage of "fnexALT" is discouraged (0 uses).
+New usage of "foelrnOLD" is discouraged (0 uses).
 New usage of "hbs1" is discouraged (5 uses).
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -63872,6 +63865,60 @@ all reals (greatest real number) for which all positive reals are greater.
...
+
+  $( Right injection of a disjoint union.  (Contributed by Mario Carneiro,
+     21-Jun-2022.) $)
+  df-inr $a |- inr = ( x e. _V |-> <. 1o , x >. ) $.
+
+  ${
+    $d C x $.
+    $( Left closure of disjoint union.  (Contributed by Jim Kingdon,
+       21-Jun-2022.)  (Revised by BJ, 6-Jul-2022.) $)
+    djulclN $p |- ( C e. A -> ( ( inl |` A ) ` C ) e. ( A |_| B ) ) $=
```
jkingdon, line 63902: This (and the other results involving restriction) make sense to me, I had been thinking of maybe doing the same myself.

jkingdon (approved):

jkingdon:
> I would like to merge this now to ease further modifications. However, if you think it's not mature enough to be merged, then that's fine and I'll remove all the xxxN/xxx before merging.

Whichever will get djudom done faster &lt;wink&gt;.

More seriously, if anyone else wants to weigh in, go for it, but as far as I'm concerned you can merge this now if you'd like.

benjub:
A few more shortenings in the same vein, including some large ones (the proof of djuunN is now much shorter than that of djuun, does not use djur, and is setvar-free!).
I propose to replace djur with eldju, too (which is a biconditional and involves the restricted injections). And even that should be less used than djuunN. For those results, setvar-free proofs are possible (reflected in the fact that these constructions work in category theory, where objects need not have elements).

I'll do the removals in a separate PR.

Side remark about djuinN/djuunN versus the old ones: since ``( inl |` A )`` is the more natural object (rather than `inl`), it makes more sense (and is faster to prove) the versions with ``ran ( inl |` A )`` rather than `( inl " A )`. (Sorry for having used the latter when informally stating the main results expected from such a section, in the initial Issue.)

jkingdon:
> I propose to replace djur with eldju, too (which is a biconditional and involves the restricted injections)

Should be fine, but maybe hold up on removing theorems until #2659 can be completed and merged? One of the proofs there uses djur and although revising it to use eldju (or djuun if feasible) is certainly an option at some point, my instinct would be to get some version of it working before we go too wild.

PR #2669, opened 2022-07-07
Some theorems for iset.mm related to disjoint union and related topics

jkingdon:
This is most of what is in #2659 but which is ready to go now (without djudom).

* A proof of `EXMID <-> ~P 1o ~~ 2o`. This was mentioned in passing in one of the sources, and seemed worth proving especially because it was easy
* A few more notes for the missing theorems list in mmil.html
* A proof of `-. E. x x e. A -> A = (/)` which I was kind of surprised we didn't have already
* A proof of `( 1o |_| 1o ) ~~ 2o` which is essentially the same as http://us.metamath.org/mpeuni/pm110.643.html
* ``C e. V -> ( C e. A <-> ( inl ` C ) e. ( A |_| B ) )`` which turns out to be needed because the proof we are trying to formalize has a lot of steps about whether ``( inl ` (/) )`` is an element of a certain disjoint union.
* ``( inl ` A ) =/= ( inr ` B )`` which is similar in spirit to http://us.metamath.org/ileuni/djuin.html but for individual elements, not entire images

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1664,6 +1664,17 @@ favor of theorems in deduction form.</TD>
 </TR>
 
 <TR>
+  <TD>sssn</TD>
+  <TD>~ sssnr , ~ sssnm</TD>
+</TR>
+
+<TR>
+  <TD>pwsn</TD>
+  <TD>~ pwsnss</TD>
+  <TD>Also see ~ exmidpw</TD>
+</TR>
+
+<TR>
 <TD>iundif2</TD>
 <TD>~ iundif2ss </TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1980,6 +1980,11 @@ there is less need for this convenience theorem.</TD>
 </TR>
 
 <TR>
+<TD>dmxp</TD>
+<TD>~ dmxpm </TD>
+</TR>
+
+<TR>
 <TD>dmxpid</TD>
 <TD>~ dmxpm </TD>
 </TR>
@@ -1987,6 +1992,11 @@ there is less need for this convenience theorem.</TD>
 <TR>
 <TD>relimasn</TD>
 <TD>~ imasng </TD>
+</TR>
+
+<TR>
+  <TD>rnxp</TD>
+  <TD>~ rnxpm</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1444,7 +1444,7 @@ is double negation elimination.</TD>
 
 <TR>
 <TD>n0</TD>
-<TD>~ n0r , ~ fin0 , ~ fin0or</TD>
+<TD>~ n0r , ~ notm0 , ~ fin0 , ~ fin0or</TD>
 </TR>
 
 <TR>


avekens (approved):

avekens:
When/how will (most of) these theorems go to set.mm?

benjub:
> When/how will (most of) these theorems go to set.mm?

As for the "when", I'd prefer that things stabilize in iset.mm before porting them to set.mm, since this will give the opportunity to evaluate the changes by comparing the webpages.

As for the "how", I'm afraid things are not easily automatizable with our current design choices... There could be ways to ease this.

iset.mm
```diff
@@ -28761,6 +28761,11 @@ subset operation on the right hand side could be an equality (that is,
     eqv $p |- ( A = _V <-> A. x x e. A ) $=
       ( cvv wceq cv wcel wb wal dfcleq vex tbt albii bitr4i ) BCDAEZBFZNCFZGZAH
       OAHABCIOQAPOAJKLM $.
+
+    $( A class which is not inhabited is empty.  (Contributed by Jim Kingdon,
+       1-Jul-2022.) $)
+    nm0 $p |- ( -. E. x x e. A -> A = (/) ) $=
```
benjub, line 28767: maybe state it as a biconditional ?

iset.mm
```diff
@@ -64617,6 +64617,16 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
     BRVETUACZATUBZEFVCVNVOVEUCUIVEADVPEVCVDUDUEUFVMATBUGUHVGUJVHABUKULUMBAVHUNU
     OUPUTUQVCVKBURVDBAUSVAVB $.
 
+  $( Disjoint union version of one plus one equals two.  (Contributed by Jim
+     Kingdon, 1-Jul-2022.) $)
+  dju1p1e2 $p |- ( 1o |_| 1o ) ~~ 2o $=
```
benjub: Nice theorem. Plus, the funny face that it represents in ascii art proves that " |_| " was the best choice for disjoint union.

iset.mm
```diff
@@ -64015,7 +64015,22 @@ all reals (greatest real number) for which all positive reals are greater.
   $}
 
   ${
-    $d A x $.
+    $d A x $.  $d C x $.
+    $( Left biconditional closure of disjoint union.  (Contributed by Jim
+       Kingdon, 2-Jul-2022.) $)
+    djulclb $p |- ( C e. V -> ( C e. A <-> ( inl ` C ) e. ( A |_| B ) ) ) $=
```
benjub: May I suggest
```
$p |- ( C e. V -> ( C e. A <-> ( ( inl |` A ) ` C ) e. ( A |_| B ) ) ) $=
```
It's very easy to recover the other version by using ``( F |` _V ) = F`` (**edit:** or rather ``( C e. A -> ( ( F |` A ) ` C ) = ( F ` C )`` which is in iset.mm with the correct hypotheses on F being a function with domain containing A), but for the other direction, you'd have to reprove the result.

iset.mm
```diff
@@ -64252,6 +64252,15 @@ all reals (greatest real number) for which all positive reals are greater.
       UDUMUGUJBMSHBQAUETUF $.
   $}
 
+  ${
+    $( Left and right injection never produce equal values.  (Contributed by
+       Jim Kingdon, 2-Jul-2022.) $)
+    djune $p |- ( ( A e. V /\ B e. W ) -> ( inl ` A ) =/= ( inr ` B ) ) $=
```
benjub: Not sure how you want to use it, but similarly, I would propose
```
$p |- ( ( A e. V /\ B e. W ) -> ( ( inl |` V ) ` A ) =/= ( ( inr |` W ) ` B ) ) $=
```

jkingdon:
> > When/how will (most of) these theorems go to set.mm?
> 
> As for the "when", I'd prefer that things stabilize in iset.mm before porting them to set.mm, since this will give the opportunity to evaluate the changes by comparing the webpages.

That seems like a good plan, I think. There seem to be quite a few things we are still working out.

> As for the "how", I'm afraid things are not easily automatizable with our current design choices... There could be ways to ease this.

I do aim for some ability to copy-paste, but that will never be 100% (for example, if an iset.mm proof uses DECID or a set.mm proof uses excluded middle, and various other situations).

iset.mm
jkingdon: Part of the problem is that it takes a bit of effort to see how this is used, because the usage is off in #2659 which is waiting for djudom but let me take a look.

So `djuclb` is instantiated as `⊢ (∅ ∈ V → (∅ ∈ 𝑢 ↔ (inl‘∅) ∈ (𝑢 ⊔ 1𝑜)))`. And I don't see how to make that work with ``( ( inl |` u ) ` (/) )`` because the whole point of this part of the proof is that we have `(inl‘∅) ∈ (𝑢 ⊔ 1𝑜)` and are trying to go from there to `∅ ∈ 𝑢`.

This isn't just something I thought up, I'm formalizing the proof of Proposition 1.2 of PradicBrown2021 and that proof is full of `inl(0)` and `inr(0)`.

I probably was too unclear with my earlier comment in favor of the restrictions on `inl` and `inr` - I was mostly talking about http://us.metamath.org/ileuni/djulf1o.html in particular and didn't think very hard beyond that. More widely, maybe it is good, maybe it is bad, I'm not sure I know enough to say.

iset.mm
jkingdon: I suppose that would work with some extra steps (with V and W set to `_V` for the case I care about), but I'm not sure I see the point, especially when V and W are just arbitrary classes, not tethered to any disjoint union.

To explain how it is being used, it is part of the formalization of:

> Suppose g(0) = inr(0) and g(1) = inr(0). We prove ¬p holds. To this end, assume p holds.
Hence 0 ∈ A. Since g is surjective there must be some i ∈ 2 such that g(i) = inl(0),
contradicting our assumption.

from the PradicBrown2021 proof.

The proof in #2659 formalizes this by showing that ``( inl ` (/) ) = ( inr ` (/) )`` (because they are both equal to g(i) as described in the quote), and then ``( inl ` (/) ) =/= ( inr ` (/) )`` by djune which is the contradiction described in the quote.

iset.mm
jkingdon: Seems like a good idea. I'll have to change the proof which depends on it (in #2659 ) but that won't be hard.

jkingdon: I updated this pull request with making this biconditional (and renaming it to `notm0` to avoid a naming clash with set.mm).

I also copied http://us.metamath.org/mpeuni/nel0.html from set.mm because it is a similar theorem which iset.mm doesn't have yet but where the set.mm proof works unmodified.

I should probably stop now before I make this pull request even bigger than it is.

iset.mm
benjub: > I'm formalizing the proof of Proposition 1.2 of PradicBrown2021 and that proof is full of `inl(0)` and `inr(0)`.

It looks to me like the functions used in that proof are (using their notation) `` ( inl |` A ) `` and `` ( inr |` 1o ) ``.

iset.mm
benjub: As above: they actually use the functions `` ( inl |` A ) `` and `` ( inr |` 1o ) ``. Every left-injection is from some X to some X |_| Y, so is of the form ``( inl |` X )``. Actually, `inl` is a "function template".
(**Edit:** from time to time (although probably rarely, only when you need to go under the hood), it might be easier to use `inl` directly, but since ``inl = ( inl |` _V )``, you can prove these easily from the theorems which use the restricted versions.)

iset.mm
jkingdon: > > I'm formalizing the proof of Proposition 1.2 of PradicBrown2021 and that proof is full of `inl(0)` and `inr(0)`.
> 
> It looks to me like the functions used in that proof are (using their notation) `` ( inl |` A ) `` and `` ( inr |` 1o ) ``.

Hmm, I see your point. They do define `inl` (in their notation) as `inl : A → A + B` which seems like there is a different `inl` for each disjoint union in play (which in informal notation can just be glossed over, but for us corresponds to, presumably, restriction).

I'll try formalizing with the restricted `inl` because I guess it probably is feasible (I'll have to try it to convince myself for sure).

Now, even if both are possible, I'm not sure I'm sold on which is better. I want to see what the proofs from #2659 look like with the extra notation. (and even more so if we try to do something difficult, like the proof that http://us.metamath.org/mpeuni/sbth.html implies excluded middle - since all the current work is just the warm-up exercise for that rather longer proof). For the moment I'll keep both `djuclb` and `djuclbN` for the two flavors. But we'll have to figure out whether we keep both (with this naming convention or a better one), consign some to mathboxes, or something else. I think I'm leaning towards some flavor of having both around until we have more experience with this stuff (in the context of set.mm and/or iset.mm - experience from elsewhere is sorta relevant but probably not completely the same), but that's a pretty tentative conclusion.

jkingdon: I now believe that ``$p |- ( C e. V -> ( C e. A <-> ( ( inl |` A ) ` C ) e. ( A |_| B ) ) ) $=`` is not provable (or at least, not using any technique we have figured out yet).

The situation is that you have `C e. V` and ``( ( inl |` A ) ` C ) e. ( A |_| B )`` and you need to prove `C e. A`. If `C e. A` were decidable this would be easy, you can show that `-. C e. A` is a contradiction (because a function outside its domain is `(/)` and `(/)` is not an element of any disjoint union). But unless we have some form of reverse closure, I don't see any way to get anything useful out of ``( ( inl |` A ) ` C ) e. ( A |_| B )``.

jkingdon:
I'd like to get the non-controversial parts of this merged so they don't get tangled up in a bunch of unrelated issues.

The two choices I thought of are:

1. Remove `djulclb` and `djune` from this pull request and move them back to #2659 . That's where they are being used which perhaps slightly better clarifies the tradeoffs in how they are stated.
2. Merged this pull request as-is because after all these theorems do hold and we can always supplement or replace them later.

What do you prefer, @benjub and @avekens ?

avekens:
> What do you prefer, @benjub and @avekens ?

For me it is OK to merge this pull request now.

benjub:
Sure, merge, and we can make other changes later.

iset.mm
benjub: (Obvious preamble: this question is independent from the above question about restriction.)
These questions about "reverse closure" do not appear very natural to me anyway. To me, the mere writing of "``( ( inl |` A )  ` C )``" should be "allowed" only once it has been proved that `C e. A` (possibly, both in a context `ph`).
As a consequence, I wonder if one would need this kind of result for any applications ?

jkingdon: Maybe reverse closure isn't the right name for the situation with disjoint union, given that disjoint union carries with it more information, or at least more explicit information, than you have with something like http://us.metamath.org/mpeuni/ndmovrcl.html .

I'm in the process of writing the #2659 proof both ways so we can compare. But basically the situation is either that the proof can carry around (directly, or via a context  and hypotheses) either ``( inl ` C ) e. ( A |_| B )`` or ``A e. C /\ ( ( inl |` A ) ` C ) e. ( A |_| B )``.

The statement that "the mere writing of ``( ( inl |` A ) ` C )`` should be 'allowed' only once it has been proved that C e. A" is perhaps an argument for writing ``( inl ` C )`` instead. But perhaps that has other downsides (which I'll admit I don't really see yet).

benjub: > The statement that "the mere writing of `` ( ( inl |` A ) ` C ) `` should be 'allowed' only once it has been proved that C e. A" is perhaps an argument for writing `` ( inl ` C ) `` instead.

To me, this is an argument for proving `C e. A` first !  But I admit that I may be a bit of an extremist on that, and a while back, @digama0 remarked that from time to time (especially for the most basic results), we should unapologetically take advantage of some "reverse closure" or related results.

jkingdon: > To me, this is an argument for proving `C e. A` first !

A lot of times, I would tend to see this as both appealing and practical. But I'm not so sure when I see how it is interacting with http://us.metamath.org/ileuni/eldju.html in the context of the #2659 proof. Anyway, the #2659 branch now has both versions.

Issue #2648, opened 2022-06-26
Theorems which don't use some of their hypotheses


savask:
A recent pull request #2635 mentioned a theorem `rspcda` which didn't use one of its essential hypotheses. I ran a program to find other examples of such theorems in set.mm, but since there are hundreds of occurrences, I narrowed down the search with the following rules:

* A theorem must be in the main part of set.mm,
* It is not discouraged,
* A theorem and all its essential hypotheses are enclosed in a block, and this block contains one theorem only.

There are 53 such theorems, here's the list grouped by contributors:

**AV**: ~~[chmaidscmat](http://us.metamath.org/mpeuni/chmaidscmat.html), [clwwlkvbij](http://us.metamath.org/mpeuni/clwwlkvbij.html), [cramerimplem1](http://us.metamath.org/mpeuni/cramerimplem1.html), [distspace](http://us.metamath.org/mpeuni/distspace.html), [edgnbusgreu](http://us.metamath.org/mpeuni/edgnbusgreu.html), [fnmpt2ovd](http://us.metamath.org/mpeuni/fnmpt2ovd.html), [gsummptnn0fz](http://us.metamath.org/mpeuni/gsummptnn0fz.html), [inveq](http://us.metamath.org/mpeuni/inveq.html), [matmulcell](http://us.metamath.org/mpeuni/matmulcell.html), [mdet0](http://us.metamath.org/mpeuni/mdet0.html), [minmar1marrep](http://us.metamath.org/mpeuni/minmar1marrep.html), [ressval3d](http://us.metamath.org/mpeuni/ressval3d.html), [uhgrvtxedgiedgb](http://us.metamath.org/mpeuni/uhgrvtxedgiedgb.html), [uhgrwkspth](http://us.metamath.org/mpeuni/uhgrwkspth.html), [ushgredgedgloop](http://us.metamath.org/mpeuni/ushgredgedgloop.html), [zrhcofipsgn](http://us.metamath.org/mpeuni/zrhcofipsgn.html)~~
**JH**: ~~[2ndcctbss](http://us.metamath.org/mpeuni/2ndcctbss.html)~~
[...]

PR #2672, opened 2022-07-07
Superfluous hypotheses removed

avekens:
* ~ssrind moved from mathbox -> Proof of 43 theorems can be decreased using "ssrind" (planned for the next PR)
* chmaidscmat, clwwlkvbij, cramerimplem1, distspace, edgnbusgreu, fnmpt2ovd, gsummptnn0fz, inveq, matmulcell, mdet0, minmar1marrep, ressval3d, uhgrvtxedgiedgb, uhgrwkspth, ushgredgedgloop, zrhcofipsgn revised by removing superfluous hypotheses according to the list of @savask , see issue #2648.
* ~estrres, ~zrhcopsgndif (renamed ~copsgndif), revised by removing superfluous hypotheses
* comments for ~djuss and ~symg2bas adjusted according to remarks of GL
* ~djuunxp added as proposed by GL
* some dates for contributions corrected

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 5-Jul-22 zrhcopsgndif copsgndif
+ 5-Jul-22 ssrind    [same]      moved from GS's mathbox to main set.mm
+ 3-Jul-22 zrhcofipsgn cofipsgn
  1-Jul-22 bj-1ex    1oex        moved from BJ's mathbox to main set.mm
 23-Jun-22 rspcda    ---         deleted; use rspcdva instead
 17-May-22 ad5ant1345  adantl3r  eliminate duplicated theorem
--- a/discouraged
+++ b/discouraged
@@ -6431,6 +6431,7 @@
 "grpplusgx" is used by "cnaddablx".
 "grpplusgx" is used by "isgrpix".
 "grpplusgx" is used by "zaddablx".
+"gsummptnn0fzOLD" is used by "gsummptnn0fzvOLD".
 "gt-lth" is used by "ex-gt".
 "gt0srpr" is used by "mulgt0sr".
 "gt0srpr" is used by "recexsrlem".
@@ -11832,6 +11833,7 @@
 "relrngo" is used by "rngoablo2".
 "relrngo" is used by "rngoi".
 "relrngo" is used by "rngosn3".
+"ressval3dOLD" is used by "estrresOLD".
 "retbwax2" is used by "merco1lem2".
 "retbwax2" is used by "merco1lem3".
 "retbwax2" is used by "retbwax3".
@@ -13373,6 +13375,7 @@
 "zfinf" is used by "axinfndlem1".
 "zfpair" is used by "axpr".
 "zrdivrng" is used by "dvrunz".
+"zrhcofipsgnOLD" is used by "zrhcopsgndifOLD".
 New usage of "0.999...OLD" is discouraged (0 uses).
 New usage of "00sr" is discouraged (4 uses).
 New usage of "0bdop" is discouraged (2 uses).
@@ -14782,6 +14785,7 @@ New usage of "clwwlknunOLD" is discouraged (0 uses).
 New usage of "clwwlknwwlknclOLD" is discouraged (0 uses).
 New usage of "clwwlknwwlksnOLD" is discouraged (0 uses).
 New usage of "clwwlkvbijOLD" is discouraged (0 uses).
+New usage of "clwwlkvbijOLDOLD" is discouraged (0 uses).
 New usage of "cm0" is discouraged (1 uses).
 New usage of "cm2j" is discouraged (1 uses).
 New usage of "cm2ji" is discouraged (1 uses).
@@ -14885,6 +14889,7 @@ New usage of "coprmdvdsOLD" is discouraged (0 uses).
 New usage of "counop" is discouraged (0 uses).
 New usage of "cplgruvtxbOLD" is discouraged (0 uses).
 New usage of "cqpOLD" is discouraged (0 uses).
+New usage of "cramerimplem1OLD" is discouraged (0 uses).
 New usage of "crhmsubcALTV" is discouraged (1 uses).
 New usage of "cringcALTV" is discouraged (9 uses).
 New usage of "cringcatALTV" is discouraged (0 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -220226,8 +220331,8 @@ operation is a permutation group (group consisting of permutations), see
         UOUL $.
 
       $( The symmetric group on a pair is the symmetric group S_2 consisting of
-         the identity and the transposition.  This theorem is also valid if the
-         elements are identical: then it collapses to theorem ~ symg1bas .
+         the identity and the transposition.  In the case that both elements
+         are identical, S_2 collapses to S_1, see theorem ~ symg1bas .
```
benjub: "S_2 collapses to S_1" is incorrect. It's no different from, say, "3 collapses to 6".

set.mm
```diff
@@ -300838,22 +301071,19 @@ unit group (that is, the nonzero numbers) to the field.  (Contributed
...
-      ( cpsmet cfv wcel w3a cxp cxr wf co cc0 wceq wa cle wbr psmetf psmet0 jca
-      3ad2ant1 3adant3 psmetge0 psmetsym jca32 ) CEHIJZAEJZBEJZKZEELMCNZAACOPQZ
-      RPABCOZSTUOBACOQULUMUNUIUJUMUKCEUAUDUIUJUNUKACEUBUEUCABCEUFABCEUGUH $.
-  $}
+  $( A set ` X ` together with a (distance) function ` D ` which is a
+     pseudometric is a _distance space_ (according to E. Deza, M.M. Deza:
+     "Dictionary of Distances", Elsevier, 2006), i.e. a (base) set ` X `
+     equipped with a _distance_ ` D ` , which is a mapping of two elements of
+     the base set to the (extended) reals and which is non-negative, symmetric
+     and equal to 0 if the two elements are equal.  (Contributed by AV,
```
benjub: That's an extended pseudometric space. Why invent new names ? "Distance space" has a much more ambiguous meaning. This removes the need for the reference.

benjub:
With regards to your question on a previous PR, in this example, there could have been one commit for each bullet point.

set.mm
avekens: OK, I will revert this change (proposed by Gérard...).

set.mm
avekens: I did not invent this name, but took it from the referenced book. If this book is not serious, I will remove the reference, otherwise I would like to keep it. Is there a reference where the term "extended pseudometric space" is defined? This could be added here.

avekens:
> With regards to your question on a previous PR, in this example, there could have been one commit for each bullet point.

This would impede my personal workflow, because I work in a separate working directory (using some temporary files, and to avoid unintended changes in "official" files). To synchronize the changes with the git directory more often would cause some overhead and wold increase the risk of "distroying" something.

set.mm
avekens: > > With regards to your question on a previous PR, in this example, there could have been one commit for each bullet point.
> 
> This would impede my personal workflow, because I work in a separate working directory (using some temporary files, and to avoid unintended changes in "official" files). To synchronize the changes with the git directory more often would cause some overhead and wold increase the risk of "distroying" something.

Furthermore, I did not perform the changes in a sequence: the main task here was the removal of the superfluous hypotheses, the other changes are made "en passant".

set.mm
avekens: Comment revised.

jkingdon (approved):

benjub:
> > With regards to your question on a previous PR, in this example, there could have been one commit for each bullet point.
> 
> This would impede my personal workflow, because I work in a separate working directory (using some temporary files, and to avoid unintended changes in "official" files). To synchronize the changes with the git directory more often would cause some overhead and wold increase the risk of "distroying" something.

Then do what works best for you.  I'm pretty new to git myself.  I remember having had a few losses, and then working like you in a separate directory to make sure I didn't lose anything.  Then, I learned a bit of git and now I managed to have a leaner workflow and also be more confident that I do not lose changes (which means: I can recover the ones I may accidentally lose).  So, there is a learning curve, but maybe it is worth it. (Again: that's only an opinion, not a recommendation.)

set.mm
benjub: In short: I'm going to open an Issue (see below), but for now, leave it like that.

As for "extended pseudometric": I don't know for the combination, but:
- "extended metric" is standard for "metric with values in the (nonnegative) extended reals",
- "pseudometric" is standard for "nonseparated metric".

For these, see https://en.wikipedia.org/wiki/Metric_(mathematics).

From these two, "extended pseudometric" is self-explanatory.  Bourbaki calls that notion "écart", I don't know how it is translated in English (in common language, "écart" means "spacing" or "gap"), but I think a new word is not needed.

Note: "quasimetric" is standard (maybe a bit less than the other two above) for "nonsymmetric metric".  So one could introduce a notion of "extended quasipseudometric".  Even though this may look like "centipede mathematics" (*), this is actually a natural notion, since it corresponds to $([0,+\infty], \leq, +)$-enriched categories, as noted by Lawvere.  Hence, the "crucial" property of metrics is the triangle inequality, more than the other axioms.

As for the seriousness of the reference, it would be out of place for me to judge it.  What I can say is that in many specialized handbooks or dictionaries, one is more likely to find specialized terms used only within a smaller community, that may have diverged overtime. (And indeed no one is immune to this kind of bias; @digama0 recently remarked here that I may bend too much toward category theory language, which is less universal than I thought/hoped.)

I just had a look at df-psmet and nearby definitions df-xmet and df-met... We should rewrite these definitions, with each in terms of the "previous" (like magma > semigroup > monoid > group). I'm going to open an Issue.

(*) "centipede mathematics" (I think the term was coined by John Baez) is the tendency to study generalizations of usual notions without any motivation, hence removing axioms one by one, like the legs of a centipede, and seeing what properties remain. A symptom is the multiplication of terms like pseudo-smurf, quasi-, almost-, near-, semihemi-... Sometimes it is justified, but the justification should not be forgotten.

benjub (approved):

set.mm
avekens: Yes, it is a good idea to discuss this separately, because it is off topic here (this PR is about removal of hypotheses).

avekens:
> > > With regards to your question on a previous PR, in this example, there could have been one commit for each bullet point.
> > 
> > 
> > This would impede my personal workflow, because I work in a separate working directory (using some temporary files, and to avoid unintended changes in "official" files). To synchronize the changes with the git directory more often would cause some overhead and wold increase the risk of "distroying" something.
> 
> Then do what works best for you. I'm pretty new to git myself. I remember having had a few losses, and then working like you in a separate directory to make sure I didn't lose anything. Then, I learned a bit of git and now I managed to have a leaner workflow and also be more confident that I do not lose changes (which means: I can recover the ones I may accidentally lose). So, there is a learning curve, but maybe it is worth it. (Again: that's only an opinion, not a recommendation.)

@benjub thank you for your hints. I will try to follow them as far as appropriate for me with my next PRs.

PR #2675, opened 2022-07-08
spimvALT is the better alternative

wlammen:
The comments to spimv and spimvALT say, that spimv uses less axioms than spimvALT, and therefore is the preferred version.  But both depend on the same set of axioms, so this is simply not true.
If anything speaks for spimv, then that it does not depend on df-nf.  I think, this reason is not good enough for preference.  Otherwise expanding all definitions in a proof would always yield better proofs beating the short ones.  We cannot be interested in such a misleading metric.
In this PR spimvALT is made the main version, whereas spimv is degraded to an OLD version.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 8-Jul-22 spimvALT  spimv       shorter, and based on the same set of axioms
  6-Jul-22 nfimt     bj-nfimt    moved to mathbox on request of its owner
  6-Jul-22 nfimt2    nfimt       nfimt was more difficult to prove and has no application
  1-Jul-22 bj-1ex    1oex        moved from BJ's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -18006,7 +18006,7 @@ New usage of "speccl" is discouraged (0 uses).
 New usage of "specval" is discouraged (1 uses).
 New usage of "speimfwALT" is discouraged (0 uses).
 New usage of "spfwOLD" is discouraged (0 uses).
-New usage of "spimvOLD" is discouraged (0 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -19926,17 +19926,17 @@ theorem as an axiom of set theory (Axiom 0 of [Kunen] p. 10).  In the
...
 
-    $( Alternate proof of ~ spimv .  Shorter but requires more axioms.
-       (Contributed by NM, 31-Jul-1993.)  (Proof modification is discouraged.)
+    $( Obsolete version of ~ spimv as of 8-Jul-2022.  (Contributed by NM,
+       31-Jul-1993.)  Removed dependency on ~ ax-10 .  (Revised by BJ,
+       29-Nov-2020.)  (Proof modification is discouraged.)
        (New usage is discouraged.) $)
-    spimvALT $p |- ( A. x ph -> ps ) $=
-      ( nfv spim ) ABCDBCFEG $.
+    spimvOLD $p |- ( A. x ph -> ps ) $=
```
benjub, line 19938: Can you keep it as ALT, with the comment:
> Alternate proof of  ~ spimv .  Note that it requires only ~ ax-1 -- ~ ax-5 together with ~ ax6e (so that it requires only ~ ax-1 -- ~ ax-6 if we allow use of ~ ax-6 ).

Thanks.

wlammen: spim (and likely some other theorems) can likewise be proven from ax-6 directly, with a smaller footprint on axioms.

benjub: The current proof of spimv does not show that it can be proved from ax-1--ax-6. If and when it becomes clear, then maybe we can name the theorem spimvOLD, but until then, I'd like to keep it as spimvALT.

avekens (approved):

benjub (approved):

PR #2680, opened 2022-07-09
Add set exponentiation (mapping operation) to iset.mm

jkingdon:
This section intuitionizes with very few problems. Many theorems copy over without any change at all, and the others (except one) only need minor changes to the proof. The one which is not intuitionized is http://us.metamath.org/mpeuni/ralxpmap.html which I suppose is probably not provable in iset.mm (but there is no need to worry about this one any time soon).

The changes outside the section are:
* Add abn0m , straightforwardly analogous to theorems we already have
* Add f0bi , f0dom0 , and f0rn0 (copied from set.mm)

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1027,7 +1027,7 @@ with the closest replacements.</P>
 </TR>
 
 <TR>
-<TD>ax-3 , con4d , con34b , necon4bd</TD>
+<TD>ax-3 , con4d , con34b , necon4d , necon4bd</TD>
 <TD>~ con3 </TD>
 <TD>The form of contraposition which removes negation does not hold
 in intutionistic logic.</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1458,11 +1458,6 @@ is double negation elimination.</TD>
 </TR>
 
 <TR>
-<TD>rabn0</TD>
-<TD>~ rabn0m , ~ rabn0r </TD>
-</TR>
-
-<TR>
 <TD>ssdif0</TD>
 <TD>~ ssdif0im </TD>
 </TR>
@@ -1470,6 +1465,16 @@ is double negation elimination.</TD>
 <TR>
 <TD>inssdif0</TD>
 <TD>~ inssdif0im </TD>
+</TR>
+
+<TR>
+  <TD>abn0</TD>
+  <TD>~ abn0r</TD>
+</TR>
+
+<TR>
+<TD>rabn0</TD>
+<TD>~ rabn0m , ~ rabn0r </TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2303,6 +2303,14 @@ and is evaluated at a set</TD>
 </TR>
 
 <TR>
+  <TD>fnsnsplit</TD>
+  <TD><I>none</I></TD>
+  <TD>Although it should be possible to prove subset rather than
+  equality (using difsnss instead of difsnid ), this theorem is
+  lightly used in set.mm.</TD>
+</TR>
+
+<TR>
 <TD>funiunfv</TD>
 <TD>~ fniunfv , ~ funiunfvdm </TD>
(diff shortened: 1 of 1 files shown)


avekens (approved):

iset.mm
```diff
@@ -28802,6 +28802,15 @@ subset operation on the right hand side could be an equality (that is,
     ( wex cv cab wcel c0 wne abid exbii nfab1 n0rf sylbir ) ABCBDABEZFZBCNGHOAB
     ABIJBNABKLM $.
 
+  ${
+    $d ph y $.  $d x y $.
```
benjub, line 28806: Probably DV(x, y) is not needed. (Though this should not block merging.)

jkingdon: Ooh good catch. For a while I had been running `scripts/check-dvs.py` pretty often, but I sort of got out of the habit. I'll probably do a round of that and sneak the results into this pull request (unless something gets complicated in which case I'll make a follow up).

jkingdon: OK, nothing got especially complicated as such but these unneeded distinct variables are on several branches (including some now on develop) so I'll just merge this pull request and put the unneeded distinct variable removal somewhere else.

jkingdon: Oh, wait, I didn't understand this at first.

The current proof does require `$d x y $.`. But the comment is whether we should replace it with a proof that doesn't require that. I'm not sure I have a really strong opinion about that, although I think maybe the current ethos is that bundling is tolerated more so than encouraged?

Don't know if the precedent of http://us.metamath.org/ileuni/rabn0m.html tell us much but that has had the `$d x y $.` all along (not that either rabn0m or abn0m are especially important given that they are both almost unused).

benjub: Yes, I was meaning that the statement does not require it (not the proof!). At least this is true in set.mm, since the case x==y is also provable, the bundled scheme is provable by Megill's completeness theorem. I think this is probably also true in iset.mm.
As for: "is it worth the trouble trying to prove it ?" : your call (if this adds axiom requirements or complicates the proof too much, then both should be kept anyway).
As for usage: note that when x==y, then phi is "constant", so this would not be too useful anyway !

jkingdon: As for Megill's completeness theorem (theorem 9.7 of "A Finitely Axiomatized Formalization of
Predicate Calculus with Equality"), we do seem to have proved http://us.metamath.org/ileuni/ax16.html (C16′ in the paper) so I guess maybe. But given how much case elimination there is on distinctors in set.mm, I'd take a closer look before concluding any of that paper applies to iset.mm.

I spent a lot of time (well, probably just months but it seemed like a lot when I was doing it) trying to avoid/reduce distinct variables in iset.mm only to find once I got to set theory that most of the distinct variable constraints I was worried about at were just non-issues when it came time to use the theorems.

I don't have any particular way of producing a proof of `abn0m` without the `x y` constraint, so also on the dimension of effort needed I'd say it isn't worth it at least for me right now.

If someone wants to spend more time on this sort of thing (whether bundling in particular or predicate logic axiomization in iset.mm in general), go for it. You don't have to worry much about duplicating things I'm doing, because anything I'm doing in this area is incidental and driven by what we need later in iset.mm.

digama0: > As for Megill's completeness theorem (theorem 9.7 of "A Finitely Axiomatized Formalization of Predicate Calculus with Equality"), we do seem to have proved http://us.metamath.org/ileuni/ax16.html (C16′ in the paper) so I guess maybe. But given how much case elimination there is on distinctors in set.mm, I'd take a closer look before concluding any of that paper applies to iset.mm.

I recall one of the early iset.mm discussions was about the admissibility of LEM for distinctors; I think there should be an axiom about it ([ax-bndl](http://us.metamath.org/ileuni/ax-bndl.html)). Certainly it is admissible in conventional FOL.

jkingdon: Oh, thanks. I had forgotten about ax-bndl . I suppose that makes it more likely that paper would be applicable.

PR #2681, opened 2022-07-10
Definition of the "domain-disjoint-union"; its function-ness; its injectivity

benjub:
Since the universal property works on the category of sets and relations (and not only on the category of sets and functions), I rewrote the definition to work with relations (By the way: @avekens's theorem proving the universal property can thus be upgraded.)
@jkingdon @digama0 : what say you ?

(I'll prove the equivalents in maps-to notation tomorrow)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,8 +27,12 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 8-Jul-22 spimvALT  spimv       shorter, and based on the same set of axioms
  6-Jul-22 nfimt     bj-nfimt    moved to mathbox on request of its owner
  6-Jul-22 nfimt2    nfimt       nfimt was more difficult to prove and has no application
+ 5-Jul-22 zrhcopsgndif copsgndif
+ 5-Jul-22 ssrind    [same]      moved from GS's mathbox to main set.mm
+ 3-Jul-22 zrhcofipsgn cofipsgn
  1-Jul-22 bj-1ex    1oex        moved from BJ's mathbox to main set.mm
 23-Jun-22 rspcda    ---         deleted; use rspcdva instead
 17-May-22 ad5ant1345  adantl3r  eliminate duplicated theorem
--- a/discouraged
+++ b/discouraged
@@ -13431,6 +13431,8 @@ New usage of "19.27OLD" is discouraged (0 uses).
 New usage of "19.28OLD" is discouraged (1 uses).
 New usage of "19.36ivOLD" is discouraged (0 uses).
 New usage of "19.37ivOLD" is discouraged (0 uses).
+New usage of "19.38aOLD" is discouraged (0 uses).
+New usage of "19.38bOLD" is discouraged (0 uses).
 New usage of "19.3OLD" is discouraged (2 uses).
 New usage of "19.41rg" is discouraged (3 uses).
 New usage of "19.41rgVD" is discouraged (0 uses).
@@ -16901,6 +16903,7 @@ New usage of "nfimdOLD2OLD" is discouraged (0 uses).
 New usage of "nfimdOLDOLD" is discouraged (0 uses).
 New usage of "nfnOLD" is discouraged (2 uses).
 New usage of "nfnanOLD" is discouraged (0 uses).
+New usage of "nfnbiOLD" is discouraged (0 uses).
 New usage of "nfndOLD" is discouraged (1 uses).
 New usage of "nfnf1OLD" is discouraged (0 uses).
 New usage of "nfnf1OLDOLD" is discouraged (2 uses).
@@ -18428,6 +18431,8 @@ Proof modification of "19.27OLD" is discouraged (28 steps).
 Proof modification of "19.28OLD" is discouraged (28 steps).
 Proof modification of "19.36ivOLD" is discouraged (16 steps).
 Proof modification of "19.37ivOLD" is discouraged (16 steps).
+Proof modification of "19.38aOLD" is discouraged (48 steps).
+Proof modification of "19.38bOLD" is discouraged (48 steps).
 Proof modification of "19.3OLD" is discouraged (12 steps).
 Proof modification of "19.41rg" is discouraged (58 steps).
 Proof modification of "19.41rgVD" is discouraged (127 steps).
@@ -19768,6 +19773,7 @@ Proof modification of "nfimdOLD2OLD" is discouraged (19 steps).
 Proof modification of "nfimdOLDOLD" is discouraged (74 steps).
 Proof modification of "nfnOLD" is discouraged (12 steps).
 Proof modification of "nfnanOLD" is discouraged (21 steps).
+Proof modification of "nfnbiOLD" is discouraged (46 steps).
 Proof modification of "nfndOLD" is discouraged (13 steps).
 Proof modification of "nfnf1OLD" is discouraged (18 steps).
 Proof modification of "nfnf1OLDOLD" is discouraged (18 steps).
--- a/discouraged
+++ b/discouraged
@@ -1479,10 +1479,12 @@
 "ax13lem1" is used by "ax13lem2".
 "ax13lem1" is used by "ax6e".
 "ax13lem1" is used by "nfeqf2".
+"ax13lem1" is used by "nfeqf2OLD".
 "ax13lem1" is used by "wl-19.2reqv".
 "ax13lem1" is used by "wl-19.8eqv".
 "ax13lem1" is used by "wl-dveeq12".
 "ax13lem2" is used by "nfeqf2".
+"ax13lem2" is used by "nfeqf2OLD".
 "ax13lem2" is used by "wl-19.2reqv".
 "ax13lem2" is used by "wl-dveeq12".
 "ax13lem2" is used by "wl-speqv".
@@ -6431,6 +6433,7 @@
 "grpplusgx" is used by "cnaddablx".
 "grpplusgx" is used by "isgrpix".
 "grpplusgx" is used by "zaddablx".
+"gsummptnn0fzOLD" is used by "gsummptnn0fzvOLD".
 "gt-lth" is used by "ex-gt".
 "gt0srpr" is used by "mulgt0sr".
 "gt0srpr" is used by "recexsrlem".
@@ -11832,6 +11835,7 @@
 "relrngo" is used by "rngoablo2".
 "relrngo" is used by "rngoi".
 "relrngo" is used by "rngosn3".
+"ressval3dOLD" is used by "estrresOLD".
 "retbwax2" is used by "merco1lem2".
 "retbwax2" is used by "merco1lem3".
 "retbwax2" is used by "retbwax3".
@@ -13373,6 +13377,7 @@
 "zfinf" is used by "axinfndlem1".
 "zfpair" is used by "axpr".
 "zrdivrng" is used by "dvrunz".
+"zrhcofipsgnOLD" is used by "zrhcopsgndifOLD".
 New usage of "0.999...OLD" is discouraged (0 uses).
 New usage of "00sr" is discouraged (4 uses).
 New usage of "0bdop" is discouraged (2 uses).
@@ -13433,6 +13438,7 @@ New usage of "19.41vOLD" is discouraged (0 uses).
 New usage of "19.43OLD" is discouraged (0 uses).
 New usage of "19.9OLD" is discouraged (2 uses).
 New usage of "19.9dOLD" is discouraged (1 uses).
+New usage of "19.9dOLDOLD" is discouraged (0 uses).
 New usage of "19.9hOLD" is discouraged (0 uses).
 New usage of "19.9tOLD" is discouraged (2 uses).
 New usage of "1dimN" is discouraged (0 uses).
@@ -13845,7 +13851,6 @@ New usage of "ax1" is discouraged (0 uses).
 New usage of "ax10fromc7" is discouraged (3 uses).
 New usage of "ax10w" is discouraged (0 uses).
 New usage of "ax11w" is discouraged (0 uses).
-New usage of "ax12OLD" is discouraged (0 uses).
 New usage of "ax12a2-o" is discouraged (0 uses).
 New usage of "ax12el" is discouraged (0 uses).
 New usage of "ax12eq" is discouraged (0 uses).
@@ -13862,8 +13867,8 @@ New usage of "ax12vALT" is discouraged (0 uses).
 New usage of "ax13ALT" is discouraged (0 uses).
 New usage of "ax13dgen4OLD" is discouraged (0 uses).
 New usage of "ax13fromc9" is discouraged (0 uses).
-New usage of "ax13lem1" is discouraged (7 uses).
-New usage of "ax13lem2" is discouraged (4 uses).
+New usage of "ax13lem1" is discouraged (8 uses).
+New usage of "ax13lem2" is discouraged (5 uses).
 New usage of "ax1cn" is discouraged (0 uses).
 New usage of "ax1ne0" is discouraged (0 uses).
 New usage of "ax1rid" is discouraged (0 uses).
@@ -13898,7 +13903,6 @@ New usage of "axaddf" is discouraged (1 uses).
 New usage of "axaddrcl" is discouraged (0 uses).
 New usage of "axc11-o" is discouraged (0 uses).
 New usage of "axc11n-16" is discouraged (0 uses).
-New usage of "axc11nOLD" is discouraged (0 uses).
 New usage of "axc11next" is discouraged (0 uses).
 New usage of "axc11nfromc11" is discouraged (0 uses).
 New usage of "axc11rvOLD" is discouraged (1 uses).
@@ -13927,6 +13931,7 @@ New usage of "axc5sp1" is discouraged (0 uses).
 New usage of "axc711" is discouraged (2 uses).
 New usage of "axc711to11" is discouraged (0 uses).
 New usage of "axc711toc7" is discouraged (1 uses).
+New usage of "axc7eOLD" is discouraged (0 uses).
 New usage of "axcc" is discouraged (0 uses).
 New usage of "axcnex" is discouraged (0 uses).
 New usage of "axcnre" is discouraged (0 uses).
@@ -14782,6 +14787,7 @@ New usage of "clwwlknunOLD" is discouraged (0 uses).
 New usage of "clwwlknwwlknclOLD" is discouraged (0 uses).
 New usage of "clwwlknwwlksnOLD" is discouraged (0 uses).
 New usage of "clwwlkvbijOLD" is discouraged (0 uses).
+New usage of "clwwlkvbijOLDOLD" is discouraged (0 uses).
 New usage of "cm0" is discouraged (1 uses).
 New usage of "cm2j" is discouraged (1 uses).
 New usage of "cm2ji" is discouraged (1 uses).
@@ -14885,6 +14891,7 @@ New usage of "coprmdvdsOLD" is discouraged (0 uses).
 New usage of "counop" is discouraged (0 uses).
 New usage of "cplgruvtxbOLD" is discouraged (0 uses).
 New usage of "cqpOLD" is discouraged (0 uses).
+New usage of "cramerimplem1OLD" is discouraged (0 uses).
 New usage of "crhmsubcALTV" is discouraged (1 uses).
 New usage of "cringcALTV" is discouraged (9 uses).
 New usage of "cringcatALTV" is discouraged (0 uses).
@@ -15427,6 +15434,7 @@ New usage of "edg0iedg0OLD" is discouraged (0 uses).
 New usage of "edgfiedgvalOLD" is discouraged (1 uses).
 New usage of "edgiedgbOLD" is discouraged (0 uses).
 New usage of "edginwlkOLD" is discouraged (0 uses).
+New usage of "edgnbusgreuOLD" is discouraged (0 uses).
 New usage of "edgvalOLD" is discouraged (3 uses).
 New usage of "ee001" is discouraged (0 uses).
 New usage of "ee002" is discouraged (0 uses).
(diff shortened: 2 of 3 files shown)


digama0:
Not a fan of the name or symbol. `case` is better, or something along the same lines. I have no strong feelings about it being an infix vs prefix operator, but I think infix is a bit weird with the words also in there.

You also shouldn't need `dom R` and `dom S` in the definition; I think ``( ( R o. `' inl ) u. ( S o. `' inr ) )`` alone should work.

benjub:
> Not a fan of the name or symbol. `case` is better, or something along the same lines. I have no strong feelings about it being an infix vs prefix operator, but I think infix is a bit weird with the words also in there.

I haven't found a good name nor token/symbol/label... What I would like is:
* infix
* reminiscent of the disjoint union
* that will be similar to the dual notion "unique function into the direct product" (in @scott-fenton @sctfn 's mathbox)

I thought of something like `( F |__| G )` (two underscores) and using something like a boldface or underlined $\sqcup$ for the symbol, but this may look too similar to the disjoint union symbol, so people would think it denotes the disjoint union on functions: $F \sqcup G \colon A \sqcup B \to X \sqcup Y$. What do you think ?

There are already symbols with a single letter as index in set.mm, like $+_g$
 so maybe
$( F \sqcup_d G ) \colon A \sqcup B \to X$ and
$( F \times_r G ) \colon X \to  A \times B$
where the $d$ and $r$ indices show that the operation is done at the domain and range respectively ?

As for the name, I'm open to suggestions... The phrase "the unique function from the disjoint union" (dually, "the unique function into the direct product") is descriptive but a bit long. I'm not very happy with "the domain-disjoint-union" either, but haven't found anything else.

The abbreviation used in labels would probably be a consequence of the choice made for the name.

> You also shouldn't need `dom R` and `dom S` in the definition; I think `` ( ( R o. `' inl ) u. ( S o. `' inr ) ) `` alone should work.

Indeed it is not needed: since you post-compose with `R`, resp. `S`, the extra stuff is thrown away thanks to our encoding, but I find it cleaner to make sure that two relations are composable before composing them (that is, write `( R o. S )` only once you know that $S \subseteq A \times B$ and $R \subseteq B \times C$, to which the current df-djudom at least brings us closer), especially since here it comes at almost no extra cost, and the utility theorems to handle it are there.

digama0:
I don't think overloading the disjoint union symbol is a good idea either, even if that notation is sometimes used in mathematics. They are fairly different operations, and as you say disjoint union of functions (in the sense of a union that is disjoint) is also a common operation to perform on functions so I think it will be more confusing. We don't use `a X. b` or a variant to indicate the pairing operator either.

If you really want to use an operator symbol, I would suggest `+`, but I don't think this improves the situation as much as just using a prefix operator and not getting tricky about this.

By the way, what you write as $\times_r$ already exists in set.mm, it is spelled `oF _I`.

> As for the name, I'm open to suggestions... The phrase "the unique function from the disjoint union" (dually, "the unique function into the direct product") is descriptive but a bit long. I'm not very happy with "the domain-disjoint-union" either, but haven't found anything else.

I've already made several indications of the direction I would like the name to go. This is definition by cases in its purest form. The name `case` is used in haskell for destructing inductive types, and in metamath I would expect all inductive types to essentially be built out of recursive application of `X.` and `|_|` so this is essentially the only `case` operation we will ever need.

> Indeed it is not needed: since you post-compose with `R`, resp. `S`, the extra stuff is thrown away thanks to our encoding, but I find it cleaner to make sure that two relations are composable before composing them (that is, write `( R o. S )` only once you know that S⊆A×B and R⊆B×C, to which the current df-djudom at least brings us closer), especially since here it comes at almost no extra cost, and the utility theorems to handle it are there.

I don't think this is necessary. It's not just encoding; this is exactly how you would write what I originally wrote as `{ <inl x, y> | <x, y> e. R }` since we're just applying a function to the first component, and it doesn't matter what the domain of `inl` is there, you always get the expected relation (they are always "composable"). You would usually not write ``((F |` ran G) o. G)`` for the composition of `F` and `G` either, even though the domain of `F` is generally a proper superset of the range of `G`. This is just the same thing but for right-composition by a function. (Note that ``( F o. `' inl )`` is equal to `` `' ( inl o. `' F )``, which makes it look more like a left composition.)

jkingdon:
I'm a bit surprised there isn't more of a standard name or notation for this, but apparently there isn't. So I'm not sure I have a lot to add on that.

I would favor removing the unnecessary restriction. I guess I just don't see what it gets us.

benjub:
> what you write as $\times_r$ already exists in set.mm, it is spelled oF _I.

Thanks. It looks like it works for functions only, though (but generalizing to relations is straightforward).

> I've already made several indications of the direction I would like the name to go. This is definition by cases in its purest form. The name case is used in haskell for destructing inductive types, and in metamath I would expect all inductive types to essentially be built out of recursive application of X. and |_| so this is essentially the only case operation we will ever need.

I don't know Haskell, but I agree that it corresponds to a definition by cases and to "destructing" inductive sum types, and for instance in OCaml it is `match x with inl(a) -> F(a) | inr(b) -> G(b)`.  The thing is that this will make (i)set.mm bend towards computer science and further from mathematics. I'm not saying it is a good or bad thing, or even an important shift. A symptome is that we begin to justify additions by "Haskell has this" rather than "Lang's algebra, or Bourbaki, etc., has that". Formalization is of course at the crossroad, and you have of course a broad view on it. One common obstacle to wide adoption of formalization by mathematicians is the distance to mathematical practice (it is not only my personal opinion, but is written in several surveys, e.g. the QED manifesto), and here we go a little bit further away from mathematical practice: certainly seeing $F \sqcup_d G$ in a math book would be less surprising than seeing `case(F, G)` for the object under consideration). But again, practice changes under the influence of computer science and formalization, so... : 

I'll change to what you propose, that is:
* token: `case ( F , G )`;
* unicode symbol: case(F, G) (that is: no pretty-printing);
* abbreviation within labels (as in df-xxx):  ??? ;
* name: "The relation case(F, G) is the relation given by a definition by cases from the relations F and G, with domain the disjoint union of the domains of F and of G." ???

**IF** all things pondered, including the above paragraph and the direction that (i)set.mm may take, you still think it is the right direction to go. Maybe we can also seek for other opinions ? (on mathoverflow ? among your colleagues ?)

> >Indeed it is not needed: since you post-compose with R, resp. S, the extra stuff is thrown away thanks to our encoding, but I find it cleaner to make sure that two relations are composable before composing them (that is, write ( R o. S ) only once you know that S⊆A×B and R⊆B×C, to which the current df-djudom at least brings us closer), especially since here it comes at almost no extra cost, and the utility theorems to handle it are there.

> I don't think this is necessary. It's not just encoding; this is exactly how you would write what I originally wrote as { <inl x, y> | <x, y> e. R } since we're just applying a function to the first component, and it doesn't matter what the domain of inl is there, you always get the expected relation (they are always "composable"). You would usually not write ``((F |` ran G) o. G)`` for the composition of F and G either, even though the domain of F is generally a proper superset of the range of G.

We do not have a notion of codomain in the current set.mm encoding, unfortunately. It is not a useless notion. For instance, "the exponential function from \R to \R" is a different object from "the exponential function from \R to \R_{>0}" (for instance, one is surjective and the other is not). In a "usual" situation, before writing  `( F o. G )`, I would have proved already that `F C_ ( B X. C )` and  `G C_ ( A X. B )` for some A, B, C. I'm trying to use consistently ``( inl |` A )`` instead of the bare `inl`, since there will be cases when the former is needed, and going from the former to the latter is much easier than the other way around.

I would like to keep the restriction in the definition at least for now, and be able to see how it plays out on the generated webpages.

> (Note that ``( F o. `' inl )`` is equal to `` `' ( inl o. `' F )``, which makes it look more like a left composition.)

Yes, this fact is repeatedly used in the proofs in this PR, but in the case where F is a function, we have a composition of functions on the left expression, which looks more natural to me (this fact might be reflected by a slightly shorter proof of djufun).

digama0:
> I don't know Haskell, but I agree that it corresponds to a definition by cases and to "destructing" inductive sum types, and for instance in OCaml it is `match x with inl(a) -> F(a) | inr(b) -> G(b)`. The thing is that this will make (i)set.mm bend towards computer science and further from mathematics.

This is only true to the extent that the definition is actually used. If it's not useful it won't be used, but if it is used it should be used with the accepted name.

> I'm not saying it is a good or bad thing, or even an important shift. A symptome is that we begin to justify additions by "Haskell has this" rather than "Lang's algebra, or Bourbaki, etc., has that". 

I think this is reasonable given that this is a much more important concept for CS than it is for math. The area of math that uses this operation the most is type theory and this is heavily influenced by CS concepts.

> Formalization is of course at the crossroad, and you have of course a broad view on it. One common obstacle to wide adoption of formalization by mathematicians is the distance to mathematical practice (it is not only my personal opinion, but is written in several surveys, e.g. the QED manifesto), and here we go a little bit further away from mathematical practice: certainly seeing F⊔dG in a math book would be less surprising than seeing `case(F, G)` for the object under consideration). But again, practice changes under the influence of computer science and formalization, so...

Indeed, I suspect that mathematicians are also being influenced by these notions from computer science, it's not just a one way path.

> 
> I'll change to what you propose, that is:
> 
>     * token: `case ( F , G )`;
> 
>     * unicode symbol: case(F, G) (that is: no pretty-printing);
> 
>     * abbreviation within labels (as in df-xxx):  ??? ;

I think df-case would make sense here.

>     * name: "The relation case(F, G) is the relation given by a definition by cases from the relations F and G, with domain the disjoint union of the domains of F and of G." ???

I think the description should emphasize the function case, although the definition can be set up so as to work on relations too.
 
> **IF** all things pondered, including the above paragraph and the direction that (i)set.mm may take, you still think it is the right direction to go. Maybe we can also seek for other opinions ? (on mathoverflow ? among your colleagues ?)

I'm currently at an ICERM meeting for mathematicians to learn about lean (I'm one of the instructors). I will try to take some straw polls about this matter.

> > > Indeed it is not needed: since you post-compose with R, resp. S, the extra stuff is thrown away thanks to our encoding, but I find it cleaner to make sure that two relations are composable before composing them (that is, write ( R o. S ) only once you know that S⊆A×B and R⊆B×C, to which the current df-djudom at least brings us closer), especially since here it comes at almost no extra cost, and the utility theorems to handle it are there.
> 
> > I don't think this is necessary. It's not just encoding; this is exactly how you would write what I originally wrote as { <inl x, y> | <x, y> e. R } since we're just applying a function to the first component, and it doesn't matter what the domain of inl is there, you always get the expected relation (they are always "composable"). You would usually not write `` ((F |` ran G) o. G) `` for the composition of F and G either, even though the domain of F is generally a proper superset of the range of G.
> 
> We do not have a notion of codomain in the current set.mm encoding, unfortunately. It is not a useless notion. For instance, "the exponential function from \R to \R" is a different object from "the exponential function from \R to \R_{>0}" (for instance, one is surjective and the other is not). In a "usual" situation, before writing `( F o. G )`, I would have proved already that `F C_ ( B X. C )` and `G C_ ( A X. B )` for some A, B, C. I'm trying to use consistently `` ( inl |` A ) `` instead of the bare `inl`, since there will be cases when the former is needed, and going from the former to the latter is much easier than the other way around.

Well, there is such an A,B,C in this case: inl is a function from V to V, and similarly for F. The only caveat is that F isn't necessarily a relation, but the effect of operations like `o.` on non-relations is fairly tame so it suffices to consider `` `' `' F`` which is a relation. There is nothing special about `dom F` here: if you want to think about `case` as an operation on relations, then `F` could have a "domain" which is a superset of `dom F` (i.e. `F C_ (A X. B)` for some `A` which is not `dom F`), in which case you will want the expression to be ``( F o. `' ( inl |` A ) )`` rather than ``( F o. `' ( inl |` dom F ) )`` anyway.

> 
> I would like to keep the restriction in the definition at least for now, and be able to see how it plays out on the generated webpages.
> 
> > (Note that `` ( F o. `' inl ) `` is equal to ``  `' ( inl o. `' F ) ``, which makes it look more like a left composition.)
> 
> Yes, this fact is repeatedly used in the proofs in this PR, but in the case where F is a function, we have a composition of functions on the left expression, which looks more natural to me (this fact might be reflected by a slightly shorter proof of djufun).

Another point I would make in favor of the other definition is that set.mm has historically preferred the smallest possible definition for things, including using "tricky" definitions or point-free definitions if it yields a shorter term. This is also helpful if definitions are ever unfolded for technical / metatheory reasons. Adding a completely redundant piece of syntax is unnecessary given that you can easily prove it as a theorem if you prefer to use it.

benjub:
> There is nothing special about `dom F` here: if you want to think about `case` as an operation on relations, then `F` could have a "domain" which is a superset of `dom F` (i.e. `F C_ (A X. B)` for some `A` which is not `dom F`), in which case you will want the expression to be `` ( F o. `' ( inl |` A ) ) `` rather than `` ( F o. `' ( inl |` dom F ) ) `` anyway.

Ah, yes !! I was thinking of the case of functions, where the domain is equal to the "corange", but of course this is not true for relations anymore. So I'll adopt your version.

Since the weekend is over, this may take a few more days... Sorry @jkingdon !

jkingdon:
> Adding a completely redundant piece of syntax is unnecessary given that you can easily prove it as a theorem if you prefer to use it.

I think this raises a relevant point - often when several definitions are possible we pick one possibility as the definition (the one which is shortest and/or has fewest dummy variables, typically) but then prove the other candidates as theorems. So in that sense we can have both. Although maybe it is a moot point given the above discussion about whether `dom F` is special.

> Since the weekend is over, this may take a few more days... Sorry @jkingdon !

LOL you have correctly detected my ..... excitement in advance we'll call it .... about getting `( A ~<_ B /\ C ~<_ D ) -> ( A |_| C ) ~<_ ( B |_| D )` proved.

But no worries, as long as we are making progress rather than spinning our wheels I'm not too worried. There's also nothing wrong with spending a bit of time planning out what we want to do, as long as we don't get stuck and let that lead to paralysis or a discussion which is too far removing from proving things. And that goes a bit more for matters of syntax and definitions. Although those things can be changed (and have been in the history of set.mm), it can be a bit harder to change them later than, say, a theorem name or various other details.

iset.mm
```diff
@@ -64371,6 +64378,85 @@ all reals (greatest real number) for which all positive reals are greater.
...
+      JZVPDVEVEVIUEZVFNZWBAVEVIUFWDVEWCVFOWBVEWCVFPQRBVGSUGACJVKJZVQEVIWCVJNZWE
+      AVEVIUHWFVIWCVJOWEVIWCVJPQRCVKSUGAVTLTWAAVFUIZVJUIZKZVTLVRWGTVSWHTVTWITVR
+      VGFWGBVGUJVFUKUMVSVKFWHCVKUJVJUKUMVRWGVSWHULUNWILMAVEVIUOUPUQVTURUSVHVLVD
+      UTVOVMBCVAVBVC $.
+  $}
+
+  $( The domain of the "domain-disjoint-union" is the disjoint union of the
+     domains.  Remark: its range is the (standard) union of the ranges.
+     (Contributed by BJ, 10-Jul-2022.) $)
+  djudom $p |- dom ( F |_|d G ) = ( dom F |_| dom G ) $=
```
jkingdon: Maybe this wants a name like `casedm` or something? (for one thing, `dm` rather than `dom` is the usual abbreviation for domain in theorem names).

I was thinking that `djudom` would naturally be the name for `( A ~<_ B /\ C ~<_ D ) -> ( A |_| C ) ~<_ ( B |_| D )` by analogy to http://us.metamath.org/mpeuni/undom.html

benjub:
Done !
Had to prove some basic results; for instance:
```
cocnvres $p |- ( S o. `' R ) = ( ( S |` dom R ) o. `' ( R |` dom S ) ) $=
cocnvss $p |- ( S o. `' R ) C_ ( ran ( R |` dom S ) X. ran ( S |` dom R ) ) $=
```
Most (all?) proofs use no dummy variables, which is more elegant.
I kept the older version to compare some proofs on the webpages, but will remove them afterwards. I think the graphical representation is an aesthetic loss, but I guess I'll live with that.

PS regarding some wording choices: there is a strange "composand", which I chose in analogy to, say, "summand" or "multiplicand" or "integrand" (I think "component" was not appropriate there). There is also the noun "composed", understood as the result of a composition (in analogy to, say, "sum" as the result of a summation).

**Edit**: I'll probably move df-case and friends at the beginning of the subsubsection "Universal property of the disjoint union" since the proof of the universal property will be adapted using `case`.

jkingdon (approved):

iset.mm
```diff
@@ -43783,6 +43804,33 @@ arbitrary sets or real numbers will not be provable (without the law of
     ( ccom cdm crn cxp wrel wss relco relssdmrn ax-mp dmcoss rncoss mp2an sstri
     xpss12 ) ABCZQDZQEZFZBDZAEZFZQGQTHABIQJKRUAHSUBHTUCHABLABMRUASUBPNO $.
 
+  ${
+    cossxp2.r $e |- ( ph -> R C_ ( A X. B ) ) $.
+    cossxp2.s $e |- ( ph -> S C_ ( B X. C ) ) $.
+    $( The composed of two relations is a relation, with bounds on its domain
+       and codomain.  (Contributed by BJ, 10-Jul-2022.) $)
```
avekens, line 43811: composed -> composition

iset.mm
```diff
@@ -63996,31 +64051,29 @@ all reals (greatest real number) for which all positive reals are greater.
...
+       would be more usable.  (Contributed by BJ and Jim Kingdon,
        4-Jul-2022.) $)
-    djuf1olem $p |- F : _V -1-1-onto-> ( { X } X. _V ) $=
-      ( vy cvv csn cxp wf1o wtru cv cop c2nd cfv wcel snid opelxpi adantl wceq
-      mpan xp2nd wb c1st 1st2nd2 xp1st elsni syl opeq1d eqtrd eqeq2d eqcom eqid
-      vex opth mpbiran 3bitr3g bicomd ad2antll f1o2d trud ) GCHZGIZBJKAFGVCCALZ
-      MZFLZNOZBEVDGPZVEVCPZKCVBPVHVICDQCVDVBGRUASVFVCPZVGGPKVFVBGUBSVJVDVGTZVFV
-      ETZUCKVHVJVLVKVJVEVFTVECVGMZTZVLVKVJVFVMVEVJVFVFUDOZVGMVMVFVBGUEVJVOCVGVJ
-      VOVBPVOCTVFVBGUFVOCUGUHUIUJUKVEVFULVNCCTVKCUMCVDCVGDAUNUOUPUQURUSUTVA $.
+    djuf1olemr $p |- ( F |` A ) : A -1-1-onto-> ( { X } X. A ) $=
```
avekens, line 64074: What does the suffix "r" mean?

avekens (changes requested):
only two minor review remarks

iset.mm
benjub: "restriction". I'm open to suggestions.

iset.mm
benjub: I do not know... See the "PS regarding wording choices" above. What does a native speaker, like @jkingdon, think ? Is it too weird ?

iset.mm
jkingdon: Since this is just the name of a lemma, I'm fine with anything. I mean, if you want it to be descriptive, you'd say `res` but I don't really expect that of lemmas (in fact the number one reason I usually name them instead of numbering them is that I often add/remove them while writing a proof and don't feel like renumbering each time).

iset.mm
jkingdon: Definitely should be "composition" here. This is probably normal English usage but it certainly is standard mathematical usage. See for example "the composition of g and f" at https://en.wikipedia.org/wiki/Function_composition .

jkingdon:
> there is a strange "composand"

It is a newly coined word (as far as I know) but I suppose there might be a case to be made for it. Perhaps it is better to change "Restricting the composands of the [composition] of a relation and an inverse relation." to "Restricting a relation and an inverse relation when they are composed together".

digama0:
I think "component" is a more normal noun form of "composition". (EDIT: I see this was already mentioned above.)

benjub:
Followed your rewordings. Ready to merge.

jkingdon (approved):

benjub:
> I'm currently at an ICERM meeting for mathematicians to learn about lean (I'm one of the instructors). I will try to take some straw polls about this matter.

@digama0 : did you get to do some ?

As for the function into a direct product, which you wrote can already be obtained in set.mm as ` oF _I ` : isn't the latter a bit backwards, and shouldn't one instead first define $\times_r$, and then define `( F oF R G )` as $R \circ (F \times_r G)$ ?

As for the symbol, we lose the duality
$( F \sqcup_d G ) \colon A \sqcup B \to X$
$( F \times_r G ) \colon X \to  A \times B$
so how to call it ? `esac` ("case" in reverse) ?

digama0:
> > I'm currently at an ICERM meeting for mathematicians to learn about lean (I'm one of the instructors). I will try to take some straw polls about this matter.
> 
> @digama0 : did you get to do some ?

I didn't find a good time and place for it, sorry. I could ask on the lean zulip but I'm fairly sure everyone will find it to be an obvious thing.

> As for the function into a direct product, which you wrote can already be obtained in set.mm as `oF _I` : isn't the latter a bit backwards, and shouldn't one instead first define ×r, and then define `( F oF R G )` as R∘(F×rG) ?

I suppose you could do that, but $×_r$ is less useful than `oF` for a general binop (if you look in set.mm you will overwhelmingly find it applied to operations other than `_I`), and if we are going to have only one definition then `oF` is clearly preferable. Given `oF`, the case for $×_r$ is much weakened.

Another thing to keep in mind is that sometimes these aren't actually operations in the strict sense of functions `F : (A X. B) --> C` and are instead just something for which `( x F y )` evaluates usefully (this is often the case for structure definitions). For such things the composition operator `o.` sometimes doesn't work as well as you would like.

> As for the symbol, we lose the duality (F⊔dG):A⊔B→X (F×rG):X→A×B so how to call it ? `esac` ("case" in reverse) ?

Please no. This kind of duality is only useful when doing literal category theory. In any other context it's just tremendously confusing to emphasize categorical duality.

benjub:
Ok, no need to ask further. I guess if you ask among computer scientists, the outcome will be biased anyway (I realize there would be the opposite bias with mathematicians, obviously, but at least if even among mathematicians, "`case`" was preferred, we would have been certain, with no regret).

Ok with the rest. Although it looks like `oF(F, G)` does not work when F and G are not functions (as opposed to `case(F, G)`) ? Just a question out of curiosity: you write "if you look in set.mm you will overwhelmingly find oF applied to operations other than _I". Is it not the case with `case` (once there are applications, of course) ? (i.e., you will have a function H producing an alternative, and then you apply `case(F, G)` to its result) ?

PR #2682, opened 2022-07-10
shorten 19.26-3an, nfimd

wlammen:
1. delete an outdated OLD theorem
2. micro-optimization: If you have a sequence of 4 bitri steps in succession, you usually split it up in a 3bitri and a bitri step.  There is a choice where to place the bitri step in the sequence.  The proof size in bytes will stay the same.  A typical choice is at the beginning or end, but our shortening rules favour the one with the least complex intermediate expression.  This is now effectuated in this PR in theorem 19.26-3an.  Note that the order of the steps in the proof changes, but they are otherwise unchanged, except for the bitri step.  Here the proof display appears different: `(∀𝑥((𝜑 ∧ 𝜓) ∧ 𝜒) ↔ ((∀𝑥𝜑 ∧ ∀𝑥𝜓) ∧ ∀𝑥𝜒))` vs. `(∀𝑥(𝜑 ∧ 𝜓 ∧ 𝜒) ↔ (∀𝑥(𝜑 ∧ 𝜓) ∧ ∀𝑥𝜒))`.  A small improvement saving a quantifier and a pair of braces.  I did not add a shortening tag for this minor correction.
3. Further shorten nfimd by 7 bytes / 2 essential steps (no extra OLD version created after only 4 days)

I wonder whether I should create a pull request for a barely visible effect like step 2 alone.  I often spot these "shortcomings" at once, because I did so many shortenings in the past.  But this puts extra load on git administration and peer review.  It is perhaps better to only pursue this idea if you change a proof anyway.  Or one should not create an OLD version in this special case. Let me know what you think.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13427,6 +13427,7 @@ New usage of "19.23OLD" is discouraged (2 uses).
 New usage of "19.23hOLD" is discouraged (0 uses).
 New usage of "19.23tOLD" is discouraged (1 uses).
 New usage of "19.23vOLD" is discouraged (0 uses).
+New usage of "19.26-3anOLD" is discouraged (0 uses).
 New usage of "19.27OLD" is discouraged (0 uses).
 New usage of "19.28OLD" is discouraged (1 uses).
 New usage of "19.36ivOLD" is discouraged (0 uses).
@@ -18426,6 +18427,7 @@ Proof modification of "19.23OLD" is discouraged (20 steps).
 Proof modification of "19.23hOLD" is discouraged (8 steps).
 Proof modification of "19.23tOLD" is discouraged (52 steps).
 Proof modification of "19.23vOLD" is discouraged (44 steps).
+Proof modification of "19.26-3anOLD" is discouraged (61 steps).
 Proof modification of "19.27OLD" is discouraged (28 steps).
 Proof modification of "19.28OLD" is discouraged (28 steps).
 Proof modification of "19.36ivOLD" is discouraged (16 steps).
--- a/discouraged
+++ b/discouraged
@@ -14889,7 +14889,6 @@ New usage of "conss34OLD" is discouraged (0 uses).
 New usage of "conventions" is discouraged (0 uses).
 New usage of "conventions-contrib" is discouraged (0 uses).
 New usage of "conventions-label" is discouraged (0 uses).
-New usage of "coprmdvdsOLD" is discouraged (0 uses).
 New usage of "counop" is discouraged (0 uses).
 New usage of "cplgruvtxbOLD" is discouraged (0 uses).
 New usage of "cqpOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub (approved):

benjub:
Good to me. I admit I didn't look at the proof modifications in 19.26-3an in non-compressed format, so I am not sure of the details of the change, only that it is very minor.

I would like **others' opinions**, but as for me, I agree that you can do the "very minor" modifications "silently", as I do sometimes, especially if they follow a very recent earlier modification. (By "silently", I mean: no need for OLD version (even less an OLDOLD version), and not necessarily indication/credit, but for that it's your choice.)

avekens:
>     2. micro-optimization: If you have a sequence of 4 bitri steps in succession, you usually split it up in a 3bitri and a bitri step.  There is a choice where to place the bitri step in the sequence.  The proof size in bytes will stay the same.  A typical choice is at the beginning or end, but our shortening rules favour the one with the least complex intermediate expression.  This is now effectuated in this PR in theorem 19.26-3an.  Note that the order of the steps in the proof changes, but they are otherwise unchanged, except for the bitri step.  Here the proof display appears different: `(∀𝑥((𝜑 ∧ 𝜓) ∧ 𝜒) ↔ ((∀𝑥𝜑 ∧ ∀𝑥𝜓) ∧ ∀𝑥𝜒))` vs. `(∀𝑥(𝜑 ∧ 𝜓 ∧ 𝜒) ↔ (∀𝑥(𝜑 ∧ 𝜓) ∧ ∀𝑥𝜒))`.  A small improvement saving a quantifier and a pair of braces.  I did not add a shortening tag for this minor correction.

I have compared the proofs in ASCII, and I counted 708 characters in the new proof, and 717 in the old one (overhead, like spaces, were identical in both cases). Therefore, it is correct to say that the new proof is shorter (regarding this new metric  "shortening rules favour the one with the least complex intermediate expression" where was/is this documented/stated?). 

> I wonder whether I should create a pull request for a barely visible effect like step 2 alone. I often spot these "shortcomings" at once, because I did so many shortenings in the past. But this puts extra load on git administration and peer review. It is perhaps better to only pursue this idea if you change a proof anyway. Or one should not create an OLD version in this special case. Let me know what you think.

Although I agree that short proof texts should be preferred (if the other criteria are identically fulfilled), I think it is difficult to achive this in general (unless we have a tool which provides the standardized length of a proof text, and this is only the simplest way to obtain a measure for the "complexity of intermediate expressions"). We must avoid that proofs are modified too often (rearranging the order of proof steps back and forth). Therefore, I would propose to perform such rearrangements only if the proof is modified anyway. In this case, we need not discuss if OLD versions should be created, "shortened by" tags shoud be added or separate PRs should be opened fur such rearrangements only.

wlammen:
@avekens I wonder whether one can interpret http://us2.metamath.org/mpeuni/conventions.html, Acceptable shorter proofs

> (2) reduce the size of the HTML file generated with SHOW STATEMENT xx / HTML

in that way, since fewer symbols result in a shorter page.

avekens (approved):
I think for this PR we can keep 19.26-3anOLD, to have this case as an example. In future, however, we should not have an OLD version only for such a rearrangement, as I described above.

avekens:
> @avekens I wonder whether one can interpret http://us2.metamath.org/mpeuni/conventions.html, Acceptable shorter proofs
> 
> > (2) reduce the size of the HTML file generated with SHOW STATEMENT xx / HTML
> 
> in that way, since fewer symbols result in a shorter page.

OK, but I think this is even harder to be checked than the ASCII-version. And without tool support, it is not possible (with acceptable effort) for a reviewer to check if such a modification of a proof is justified (see also Benoît's remark).

PR #2690, opened 2022-07-16
Superfluous hypotheses removed (3)

avekens:
~opseqsn revised not only by removing superfluous hypotheses according to the list of @savask , see issue #2648, but also by renaming it ~opseqsni and by providing its closed form ~opeqsng. "(Avoid depending on this detail.)" is added to its comment, because this theorem depends on the Kuratowski definition of ordered pairs used in set.mm .

Furthermore, theorems depending on this theorem were revised. Their comments were also enhanced by "(Avoid depending on this detail.)":

* ~snopeqop: two superfluous hypotheses are removed
* ~propeqop, ~iunopeqop, ~funopsn, ~funsneqopb: proof adjusted
* ~snopeqopsnid: new theorem, replacing ~funsneqopsn
* ~funsneqopsn, ~funsneqop marked as obsolete
* ~vtxvalsnop, ~iedgvalsnop: proof shortened
* header comment for section "Vertices and indexed edges" enhanced with a hint to the specific (Kuratowski) definition of ordered pairs in set.mm

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -6129,6 +6129,7 @@
 "funcringcsetclem9ALTV" is used by "funcringcsetcALTV".
 "funiedgdm2valOLD" is used by "funiedgvalOLD".
 "funiedgdmge2valOLD" is used by "edgfiedgvalOLD".
+"funsneqopsnOLD" is used by "funsneqopOLD".
 "funvtxdm2valOLD" is used by "funvtxval0OLD".
 "funvtxdm2valOLD" is used by "funvtxvalOLD".
 "funvtxdmge2valOLD" is used by "basvtxvalOLD".
@@ -10983,6 +10984,7 @@
 "opelreal" is used by "axrnegex".
 "opelreal" is used by "axrrecex".
 "opelreal" is used by "ltresr".
+"opeqsnOLD" is used by "snopeqopOLD".
 "opidon2OLD" is used by "exidreslem".
 "opidonOLD" is used by "opidon2OLD".
 "opidonOLD" is used by "rngopidOLD".
@@ -12547,6 +12549,7 @@
 "snatpsubN" is used by "pclfinclN".
 "snatpsubN" is used by "pointpsubN".
 "snexALT" is used by "p0exALT".
+"snopeqopOLD" is used by "funsneqopsnOLD".
 "spancl" is used by "elspancl".
 "spancl" is used by "shatomistici".
 "spancl" is used by "shsupcl".
@@ -15750,6 +15753,8 @@ New usage of "funiedgdm2valOLD" is discouraged (1 uses).
 New usage of "funiedgdmge2valOLD" is discouraged (1 uses).
 New usage of "funiedgvalOLD" is discouraged (0 uses).
 New usage of "funprgOLD" is discouraged (0 uses).
+New usage of "funsneqopOLD" is discouraged (0 uses).
+New usage of "funsneqopsnOLD" is discouraged (1 uses).
 New usage of "funtpgOLD" is discouraged (0 uses).
 New usage of "funvtxdm2valOLD" is discouraged (2 uses).
 New usage of "funvtxdmge2valOLD" is discouraged (1 uses).
@@ -17259,6 +17264,7 @@ New usage of "opelopabsbALT" is discouraged (3 uses).
 New usage of "opelreal" is discouraged (8 uses).
 New usage of "opelresOLD" is discouraged (0 uses).
 New usage of "opelresgOLD" is discouraged (0 uses).
+New usage of "opeqsnOLD" is discouraged (1 uses).
 New usage of "opidon2OLD" is discouraged (1 uses).
 New usage of "opidonOLD" is discouraged (2 uses).
 New usage of "opnmblALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
Why relabel it opeqsni ? It seemed to me that our conventions are xxxg / xxx.

set.mm
```diff
@@ -45957,61 +45982,78 @@ This theorem is proved directly from set theory axioms (no set theory
...
+      ( cvv wcel cop csn wceq wb wa a1i wn wi c0 eqeq2d syl6bi eleq1 w3a ancoms
+      eqcom opeqsng syl5bb opeqsni anbi2d 3anan12 bicomi 3bitrd opprc2 wne opex
+      snnz eqneqall mpi adantr notbid eqcoms pm2.21 impd com12 impbid pm2.61ian
+      3ad2ant2 opprc1 snex pm2.24i 3ad2ant3 pm2.61i ) CGHZABIZJZCDIZKZABKZCDKZC
+      AJZKZUAZLZDGHZVKWAWBVKMZVOVQVLCJKZMZVQVPVSMZMZVTVOVNVMKZWCWEVMVNUCVKWBWHW
+      ELCDVLGGUDUBUEWCWDWFVQWDWFLWCABCEFUFNUGWGVTLWCVTWGVPVQVSUHUINUJWBOZVKMZVO
+      VTWIVOVTPVKWIVOVMQKZVTWIVNQVMCDUKRWKVMQULVTVLABUMUNVTVMQUOUPZSUQVTWJVOVQV
+      PWJVOPVSVQWIVKVOVQWIVKOZVKVOPWIWMLDCDCKWBVKDCGTURUSVKVOUTSVAVEVBVCVDWMVOV
+      TWMVOWKVTWMVNQVMCDVFRWLSVTWMVOVSVPWMVOPVQVSWMVRGHZOVOVSVKWNCVRGTURWNVOAVG
+      VHSVIVBVCVJ $.
```
benjub: I'm surprised that the proof is that much longer ? I guess I'll wait for the webpages to have a look.

set.mm
```diff
@@ -393194,6 +393251,9 @@ Besides the usual way to represent graphs without edges (consisting of
   Other classes should not be used to represent graphs, because there could be
   a degenerated behavior of the vertex set and (indexed) edge functions, see
```
benjub, line 393252: degenerated --> degenerate (there are other instances below)

benjub (approved):

benjub:
Thanks for this work. A few remarks above.

I don't know the details of most of your proofs in graph theory, but I'm a bit surprised by the number of theorems in set.mm using the specifics of the Kuratowski definition of couples. Outside from the most basic results in set theory proper, this shouldn't be needed. In a word, the tag "avoid depending on this detail" should not only be added, but also followed. (Again, I haven't looked at the details of most proofs, so this is more of a general remark.)

set.mm
avekens: The cases ` C e/ _V ` and ` D e/ _V ` must be handled separately which makes the proof longer. Then we can assume ` C e. _V ` and ` D e. _V ` for the rest of the proof, and therefore the corresponding two hypotheses can be removed. This is also the reason why ~opeqsng ist needed.

set.mm
avekens: I think both forms exist and are possible, but "degenerate" is maybe more appropriate (it is an inherent property of these cases, and not by an action applied to it). I will fix this.

avekens:
> Why relabel it opeqsni ? It seemed to me that our conventions are xxxg / xxx.

Our conventions say 

> "When an inference is converted to a theorem by eliminating an "is a set" hypothesis, we sometimes suffix the closed form with "g" (for "more general") as in ~ uniex vs. ~ uniexg ."

An inference usually has suffix i (Conventions: "the associated inference is often labeled by adding the suffix "i""), but I think it has historic reasons that inferences with sethood hypotheses have no suffix "i" - as indicated by the example ~ uniex.

avekens:
> I don't know the details of most of your proofs in graph theory, but I'm a bit surprised by the number of theorems in set.mm using the specifics of the Kuratowski definition of couples. Outside from the most basic results in set theory proper, this shouldn't be needed. In a word, the tag "avoid depending on this detail" should not only be added, but also followed. (Again, I haven't looked at the details of most proofs, so this is more of a general remark.)

I wanted to show the degenerate cases for the two-layered definition of graphs (as pair as well as extensible structure). These cases should never be used (therefore also marked with "avoid depending on this detail"). The theorems in the chain of theorems which ends in these two theorems looked interesting in general, therefore I provided them earlier.

set.mm
avekens: Done

benjub: > it is an inherent property of these cases, and not [the result of] an action applied to it

exactly

benjub:
>>Why relabel it opeqsni ? It seemed to me that our conventions are xxxg / xxx.
>
>Our conventions say
>>"When an inference is converted to a theorem by eliminating an "is a set" hypothesis, we sometimes suffix the closed form with "g" (for "more general") as in ~ uniex vs. ~ uniexg ."
>
>An inference usually has suffix i (Conventions: "the associated inference is often labeled by adding the suffix "i""), but I think it has historic reasons that inferences with sethood hypotheses have no suffix "i" - as indicated by the example ~ uniex.

To me, it's either xxxg/xxx or xxx/xxxi, not both. So from the current situation, you could remove the final letter from either ~opeqsng or ~opeqsni (the one you prefer, but I would go for xxxg/xxx, which seems more in line with existing practice: there are many more examples in addition to uniex(g)).  Indeed, it is strange that the "main", unsuffixed, label ~opeqsn is not used.

avekens:
> To me, it's either xxxg/xxx or xxx/xxxi, not both. So from the current situation, you could remove the final letter from either ~opeqsng or ~opeqsni (the one you prefer, but I would go for xxxg/xxx, which seems more in line with existing practice: there are many more examples in addition to uniex(g)). Indeed, it is strange that the "main", unsuffixed, label ~opeqsn is not used.

OK, I renamed opeqsni -> opeqsn

PR #2691, opened 2022-07-16
moved cbv3v2 to main

wlammen:
It seems that I (re-)invented bj-cbv3v2 independently, but at a later date.  Since my version has a shorter proof, I keep it, but attribute it to BJ.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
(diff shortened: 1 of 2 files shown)


benjub:
I would suggest you move both to the main part (as cbv3v2 and cbv3v3) and remove bj-cbv3v2.

wlammen:
Only one of them cbv3v2 is (likely) provable without ax-11, I accidentally pushed a trial version.  This is now corrected.

benjub (approved):

avekens (approved):

PR #2692, opened 2022-07-17
Define NN+oo in iset.mm; also some results for set exponentiation

jkingdon:
The main content here is defining `NN+oo` as discussed in #2621 and proving some basic theorems about it. There are also some miscellaneous results especially about set exponentiation.

In more detail:
* isomnimap , a variant of http://us.metamath.org/ileuni/isomni.html stated in terms of set exponentiation
* enomni which was my way of convincing myself it was OK to not obsess too much about `_om e. Omni` versus `NN0 e. Omni` , because this theorem implies they are equivalent
* Definition of `NN+oo` as the subset of `2o ^m _om` which consists of descending sequences. I don't remember seeing any alternative definitions out there, and once we've decided on that it isn't hard to express in iset.mm notation.
* Various theorems which map http://us.metamath.org/ileuni/df-xnn0.html to `NN+oo`. I won't list all of them here but they lead up to `I : NN0* --> NN+oo` for an `I` which is shown in the hypotheses. This is most of what we are hoping to do with `I` with the most obvious next step being that we don't yet show that it is one to one (probably not hard, but this pull request is already getting big so I want to get something sent in now).

Results which are stated as in set.mm (and in most cases a proof similar to the set.mm one):
* `mapsnen`, `map1`, `mapen`, `0domg`, `mapdom1` (slightly modified and renamed to `mapdom1g`), `dom0`, `0dom`

Add to the missing theorems list some set.mm theorems which (I think) can't be proved as-is in iset.mm:
* `pw2f1o` , `pw2eng` , `pw2en`

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3074,6 +3074,11 @@ is Theorem 1 of [PradicBrown2021], p. 1.</TD>
 </TR>
 
 <TR>
+  <TD>mapdom1</TD>
+  <TD>~ mapdom1g</TD>
+</TR>
+
+<TR>
 <TD>2pwuninel</TD>
 <TD>~ 2pwuninelg</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3044,6 +3044,16 @@ other theorems we do not have</TD>
 </TR>
 
 <TR>
+  <TD>pw2f1o , pw2eng , pw2en</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would require some kind of decidability
+  hypothesis. Discussions of this sort tend to get into how
+  many truth values there are and sets like ` { s | s C_ 1o } `
+  (relatively undeveloped so far except for a few results like
+  ~ exmid01 and ~ uni0b ).</TD>
+</TR>
+
+<TR>
 <TD>enfixsn</TD>
 <TD><I>none</I></TD>
 <TD>The set.mm proof relies on difsnen</TD>


iset.mm
```diff
@@ -61943,8 +61943,40 @@ the first case of his notation (simple exponentiation) and subscript it
...
+      VKVLVM $.
+  $}
+
   ${
     $d x y A $.  $d x y B $.  $d x y C $.  $d x y D $.  $d x y V $.
+    $( Set exponentiation: ordinal 1 to any set is equinumerous to ordinal 1.
+       Exercise 4.42(b) of [Mendelson] p. 255.  (Contributed by NM,
+       17-Dec-2003.) $)
+    map1 $p |- ( A e. V -> ( 1o ^m A ) ~~ 1o ) $=
+      ( vx vy wcel c1o cmap co c0 cxp cvv 1oex a1i cv 0ex p0ex wceq wa syl6rbb
```
avekens, line 61972: If ~mapsnen was provided in closed form ( ` ( ( A e. V /\ B e. W ) -> ( A ^m { B } ) ~~ A ) ` , ~map1 would follow immediatly applying ~1oex.

iset.mm
```diff
@@ -97612,6 +97825,32 @@ equinumerous to omega (the set of finite ordinal numbers).  (Contributed
...
+    fxnn0nninf $p |- I : NN0* --> NN+oo $=
+      ( cxnn0 xnninf wf cn0 cpnf csn cun com c1o wceq wcel ccnv ccom cxp cop wa
+      cin c0 fnn0nninf wf1o pnfex omex 1oex snex xpex f1osn f1of ax-mp infnninf
+      wss snssi mp2an pm3.2i cv wn disj nn0nepnf neneqd elsni nsyl mprgbir fun2
+      fss feq1i mpbir df-xnn0 feq2i ) JKFLMNOZPZKFLZVSVRKDEUAUBZNQROZUCZUDOZPZL
+      ZMKVTLZVQKWCLZUEMVQUFUGSZWEWFWGABCDEGHUHVQWBOZWCLZWIKUSZWGVQWIWCUIWJNWBUJ
+      QWAUKRULUMUNUOVQWIWCUPUQWBKTWKURWBKUTUQVQWIKWCVLVAVBWHAVCZVQTZVDAMAMVQVEW
+      LMTZWLNSWMWNWLNWLVFVGWLNVHVIVJMVQKVTWCVKVAVRKFWDIVMVNJVRKFVOVPVN $.
+  $}
+
```
avekens, line 97853: I still wonder how the elements of NN+oo look like. Maybe an example will help, e.g., what **is** `` ( I ` 1 ) ``.
But what is most interesting: what **is** `` ( I ` 0 ) `` (the element of NN+oo corresponding to 0). If NN+oo has such a "0", the it would be better to name it NN0+oo (regardless how it is represented in HTML), wouldn't it?

iset.mm
avekens: GS provides the deduction form ~mapsnend of this theorem in his mathbox (in set.mm). This should be moved to main, and ~mapsnen and ~map1 can be derived from it afterwards (and maybe other proofs using ~mapsnen could be shortened). I'll do that in set.mm at least with my next PR.

iset.mm
```diff
@@ -62258,6 +62290,30 @@ the first case of his notation (simple exponentiation) and subscript it
...
   $}
 
+  $( Any set dominates the empty set.  (Contributed by NM, 26-Oct-2003.)
+     (Revised by Mario Carneiro, 26-Apr-2015.) $)
+  0domg $p |- ( A e. V -> (/) ~<_ A ) $=
+    ( wcel c0 wss cdom wbr 0ss ssdomg mpi ) ABCDAEDAFGAHDABIJ $.
+
+  ${
+    $d A x $.
+    $( A set dominated by the empty set is empty.  (Contributed by NM,
```
benjub, line 62300: A class is dominated by the empty set if and only if it is empty.

iset.mm
```diff
@@ -62258,6 +62290,30 @@ the first case of his notation (simple exponentiation) and subscript it
...
+      wceq cvv wcel 0ex domrefg ax-mp breq1 mpbiri impbii ) ACDEZACQZUFACBFZGZU
+      GBACBHUIUGUFUIUHCQZUGUIACUHIUJUGJACUHKAUHLMNOPUGUFCCDEZCRSUKTCRUAUBACCDUC
+      UDUE $.
+  $}
+
+  ${
+    0sdom.1 $e |- A e. _V $.
+    $( Any set dominates the empty set.  (Contributed by NM, 26-Oct-2003.)
+       (Revised by Mario Carneiro, 26-Apr-2015.) $)
+    0dom $p |- (/) ~<_ A $=
```
benjub, line 62313: I would move that right after ~0domg (i.e., swap positions of ~0dom and ~dom0).

iset.mm
```diff
@@ -86379,15 +86532,35 @@ finite and infinite sets (and therefore if the set size is a nonnegative
 $)
 
   $c NN0* $. $( The set of extended nonnegative integers $)
+  $c NN+oo $. $( Set of descending sequences in ` 2o ^m _om ` $)
```
benjub, line 86535: descending --> non-increasing

But wouldn't it be more natural to take non-decreasing instead ? Then, zero maps to the zero sequence. I don't remember what Escardo does. (**edit:** forget about that: with that option, 0 and 1 would correspond to sequences which are different at every but one point, and the natutal "closest to infinity" would be 1, which is not very desirable either.)

iset.mm
```diff
@@ -86379,15 +86532,35 @@ finite and infinite sets (and therefore if the set size is a nonnegative
...
-     ~ df-xr .  (Contributed by AV, 10-Dec-2020.) $)
+     ~ df-xr .  If we assumed excluded middle, this would be essentially the
+     same as ` NN+oo ` as defined at ~ df-nninf but in its absence the
+     relationship between the two is more complicated.  (Contributed by AV,
+     10-Dec-2020.) $)
   df-xnn0 $a |- NN0* = ( NN0 u. { +oo } ) $.
 
+  ${
+    $d f i $.
+    $( Define the set of descending sequences in ` 2o ^m _om ` .  Definition in
```
benjub, line 86552: non-increasing

iset.mm
```diff
@@ -86457,6 +86630,46 @@ finite and infinite sets (and therefore if the set size is a nonnegative
     ( cxnn0 wcel cn0 wn cpnf wceq wo wi elxnn0 pm2.53 sylbi imp ) ABCZADCZEZAFG
     ZNOQHPQIAJOQKLM $.
 
+  ${
+    $d f i $.
+    $( The point at infinity in ` NN+oo ` .  (Contributed by Jim Kingdon,
```
benjub, line 86635: Maybe for clarity, and @avekens's remark shows it is necessary:
> The point at infinity in NN+oo is the constant sequence equal to 1o.

iset.mm
```diff
@@ -86457,6 +86630,46 @@ finite and infinite sets (and therefore if the set size is a nonnegative
...
+      2onn omex peano2 eqtr4d eqimss sseq12d ralbidv df-nninf elrab2 mpbir2an
+      rgen ) BCDEZFGVAHBIUAZGZAQZJZVAKZVDVAKZLZABMZVCBHVAUBBCHCCJHCNUCUDUEUFHBV
+      AHBUJUGUKUHUIVHABVDBGZVFVGRVHVJVFCVGVJVEBGVFCRVDULBCVENOSBCVDNOUMVFVGUNSU
+      TVEPQZKZVDVKKZLZABMVIPVAVBFVKVARZVNVHABVOVLVFVMVGVEVKVATVDVKVATUOUPPAUQUR
+      US $.
+  $}
+
+  ${
+    $d N f i j $.
+    $( Elements of ` NN+oo ` corresponding to natural numbers.  (Contributed by
```
benjub, line 86649: Add a second sentence for clarity:
> In other words, the element in NN+oo corresponding to the natural number i is the sequence which "becomes zero" at index i (which explains that the sequence which "never becomes zero", that is, the constant sequence equal to 1, corresponds to infinity, see ~ infnninf ).

iset.mm
benjub: Maybe add to the comment:
> One checks (see also comments of ~infnninf and ~nnnninf) that the function I maps the natural number i to the sequence which "becomes zero" at index i, and maps infinity to the sequence which "never becomes zero", that is, the constant sequence equal to 1.

One would need to prove, with same $e-hypotheses:
```
inftonninf $p |- ( N e. _om -> ( I ` N ) = ( n e. _om |-> if ( n e. N , 1o , (/) ) ) ) $=
nntonninf $p |- ( I ` +oo ) = ( _om X. { 1o } ) $=
```
(but I  haven't looked at the detail of your proofs... This looks redundant with infnninf and nnnninf, but maybe this appears in some subproofs too.

benjub: As for the token, see discussion at #2621. Simply stated,NN0+oo would be a bit heavy. (The original sin is using NN0 for the natural numbers in my opinion.)

iset.mm
jkingdon: Why would `map1` follow from `mapsnen` ? The exponent in `map1` doesn't have to be a singleton.

iset.mm
jkingdon: Interesting. I'm not sure whether "descending" has a standard meaning here (in a context like databases or spreadsheets, descending sort order doesn't imply there are no equal elements). But sure, non-increasing sounds good.

iset.mm
jkingdon: Will change it here too.

iset.mm
jkingdon: Yeah, figuring out a way to mention "the constant sequence equal to 1o" does sound good to me. You have to be kind of deep into set.mm/iset.mm conventions to see `( A X. { B } )` and immediately think "constant function".

I went with

```
    $( The point at infinity in ` NN+oo ` (the constant sequence equal to
       ` 1o ` ).  (Contributed by Jim Kingdon, 14-Jul-2022.) $)
```

iset.mm
jkingdon: Agreed this is worth stating. Does the following look clear?

```
    $( Elements of ` NN+oo ` corresponding to natural numbers.  The natural
       number ` N ` corresponds to a sequence of ` N ` ones followed by zeroes.
       Contrast to a sequence which is all ones as seen at ~ infnninf .
       (Contributed by Jim Kingdon, 14-Jul-2022.) $)
```

iset.mm
avekens: Ah, sorry, I always confuse the two operands for the ^m operation...
However, ~map1 can be generalized to a theorem similar to ~mapsnen:

`snmapen $p |- ( ( A e. V /\ B e. W ) -> ( { A } ^m B ) ~~ { A } ) $=`

The proof of this theorem is not much longer than the proof of ~map1, therefore I will provide this with my next PR (for set.mm).

iset.mm
jkingdon: I've added comments to infnninf and nnnninf to clarify what elements of `NN+oo` look like. Let us know @avekens if the revised wording there makes it clearer.

It is a bit unclear at least in my mind what we'll state in terms of `I` explicitly and what we'll state in other ways (for example `NN0* ~<_ NN+oo` or `_om e. Omni <-> NN0* ~~ NN+oo`). But that's mostly for future pull requests.

As for the notation, I agree with @benjub that trying to put `0` into the name, either in the source or in the typeset version, is not a good idea. I guess what made me come around to this is partly that `NN+oo` is, while related to natural numbers (as seen in terminology like "one-point compactification of the natural numbers"), its own thing (defined as a sequence).

iset.mm
jkingdon: Where feasible I tend to keep the same order as set.mm. It makes it easier to do things like figure out what has been intuitionized and what has not.

iset.mm
jkingdon: Norm has a way of writing the comments to call out the part he thinks is most interesting (especially, as here, when the precise statement of the theorem is also quite easy to read). I'm not saying we should never revise his comments but I'm not sure this is a case where the revised version is obviously better (at least in my mind).

iset.mm
benjub: I like both theorems proposed by @avekens, except I would write the second as
```
snmapen $p |- ( ( A e. V /\ B e. W ) -> ( { A } ^m B ) ~~ 1o ) $=
```
since in the world of equinumerosity, the standard representant for "one" is 1o.

iset.mm
benjub: Of course, I meant that I would swap positions in both ! But I understand that it's a bit of a hassle. We have to find a better way of keeping set.mm and iset.mm in sync.

iset.mm
benjub: I liked the phrase "becomes zero at index N" (or maybe "switcthes to zero at index N"), but I'll also go with your wording. Then you can prove $\sum_{i \in \mathbb{N}} I(n)_i = n$, and say in a comment that the inverse of I (corestricted to the image of xNN0) is given by summation.

iset.mm
benjub: The statements `NN0* ~<_ NN+oo` and `_om e. Omni <-> NN0* ~~ NN+oo` should be proved as corollaries of
```
I : NN0* -1-1-->  NN+oo
_om e. Omni <-> I : NN0* -1-1-onto-> NN+oo
```
since the latter give more information. This is similar to the fact that knowing that two things are isomorphic is nice, but it's nicer to know that a given map is an isomorphism. You can see this in the current trends in (homotopy) type theory.

As an example, you may know Whitehead's theorem: if a map between CW complexes induces isomorphisms on all homotopy groups, then that map is a homotopy equivalence. However, two CW complexes with isomorphic homotopy groups need not be homotopy equivalent.

avekens: The new comments of infnninf and nnnninf are very helpful. Two add two additional theorems like the following would make things even clearer:

``0nninf $p |- ( I ` 0 ) = ( x e. _om |-> (/) )``
``1nninf $p |- ( I ` 1 ) = ( x e. _om |-> if ( x = (/) , 1o , (/) ) )``

As for the naming, I am afraid that `NN+oo` could be confused with `NN* := ( NN u. { +oo } )` (not defined yet) - that was at least my first expectation.

jkingdon: I have added ``( I ` 0 ) = ( x e. _om |-> (/) )`` and ``( I ` 1 ) = ( x e. _om |-> if ( x = (/) , 1o , (/) ) )`` . Hopefully it isn't too much scope creep to add it to this pull request (if there is some controversial thing there, I suppose I should make it a separate branch so we can get this one merged).

avekens (approved):

jkingdon:
I'm merging this partly because it seems maybe unnecessary to have both this and #2694 outstanding at the same time. As usual, if there are remaining things we want to change, we can make a follow up.

iset.mm
benjub: Fine for this one, but I really think the comment should describe the formal statement and the formal statement should formalize the comment. The "most interesting part" is not so objective. There is a discrepancy between being so cautious as to formally proving $\phi \to \phi$ on one side, and being lax with comments on the other side. The problem is not for any one statement; the problem is that these approximations might add up, and at the end of a chain of statements, we don't know what the formal version is proving (without a lot of backtracking work).

iset.mm
benjub: These examples may be useful because in theses two cases (especially for 0), the formula simplifies, but more useful would be the above two:
```
inftonninf $p |- ( N e. _om -> ( I ` N ) = ( n e. _om |-> if ( n e. N , 1o , (/) ) ) ) $=
nntonninf $p |- ( I ` +oo ) = ( _om X. { 1o } ) $=
```
(for the second, either using direct product or function notation as you did for 0nninf).

jkingdon: > ``$p |- ( I ` +oo ) = ( _om X. { 1o } ) $=``

Already present as
```
    inftonninf $p |- ( I ` +oo ) = ( x e. _om |-> 1o ) $=
```

> ``$p |- ( N e. _om -> ( I ` N ) = ( n e. _om |-> if ( n e. N , 1o , (/) ) ) ) $=``

I have no objection, so go for it.

benjub: In the comment of `df-nninf $a |- NN+oo = { f e. ( 2o ^m _om ) | ...`, you write:
> This definition would function much the same whether we used ` _om ` or ` NN0 ` , but the former allows us to take advantage of ` 2o = { (/) , 1o } ` ( ~ df2o3 ) so we adopt it.

With all the proofs using the injection ` I ` which is defined using the frightening `frec`, is this still true that `_om` is more convenient ? Sorry I didn't catch this earlier, but given the proximity of NN+oo with NN0, it seems strange to resort to using `( 2o ^m _om )` instead of `( { 0 , 1 } ^m NN0 )` ?

benjub: Another belated comment:
```
nnnninf $p |- ( N e. _om -> ( i e. _om |-> if ( i e. N , 1o , (/) ) ) e. NN+oo ) $=
```
This is also true if `N = _om`. I don't know how you write it, especially for the intuitionistic version:
```
|- ( N e. succ _om ->
|- ( N e. ( _om u. { _om } )->
|- ( ( N e. _om \/ N = _om ) ->
```
?

jkingdon: > With all the proofs using the injection `I` which is defined using the frightening `frec`,

I guess I'm just so used to converting between `_om` and `NN0` that I don't find the particular `frec((𝑥 ∈ ℤ ↦ (𝑥 + 1)), 0)` idiom as frightening as `frec` in general. Cumbersome or clutter, yes. But it is well behaved in light of https://us.metamath.org/ileuni/frec2uzf1od.html and other similar theorems.

> is this still true that `_om` is more convenient ? Sorry I didn't catch this earlier, but given the proximity of NN+oo with NN0, it seems strange to resort to using `( 2o ^m _om )` instead of `( { 0 , 1 } ^m NN0 )` ?

I guess I'm just so used to the literature which pretty much universally uses notation like 2<sup>ℕ</sup> and maybe I'm taking that too literally. I could probably be talked into `{ 0 , 1 }` (especially if I don' t have to do all the work of switching).

jkingdon: > I don't know how you write it,

In https://us.metamath.org/ileuni/df-ihash.html we use `( _om u. { _om } )` but all three of the ones you write should be equivalent even intutionistically.

Of course this trick of using `e.` to mean both "element of the the natural numbers" or "natural number less than a given natural number" only works if we use `_om` rather than `NN0` (which I'm not sure is a good reason to use `_om` but is worth pointing out).

benjub: As for the extension of nnnninf, I added it as a remark in the third commit of #2705. I wrote that it generalizes to "N e. succ _om", even if "succ _om" is not very natural in this context.  If using NN0 in place of _om, this would read "N e. NN0x".

PR #2693, opened 2022-07-18
First cut at documentation page for $j annotations

### LINKED ISSUE #2664: Document $j annotations
As described in https://github.com/metamath/set.mm/pull/2662#issuecomment-1173993333 we don't have documentation for the `$j` annotations and we should. I suppose this should be a new `annotate.html` file we check in, and then we'll need to update the website-updating script to copy it over? I suppose maybe at top level like http://us.metamath.org/mm_100.html rather than inside a directory like `mpeuni` or `ileuni`? Anyway, in order to start with the content, I'll put the proposed text here:

> <p>As noted in section 4.4.3, "Additional Information Comment ($j)", of the metamath specification, we have a way to include information in metamath proof files which are not needed for verifying, but which are useful for other tools like proof assistants, definition checkers, conversion to and from other proof formats, and the like. Although the metamath specification has some examples, it seems better to document the full list here because they aren't part of the specification in the sense of being needed for verification.</p>
> 
> <code>$( $j justification ’JUSTIFICATION’ for ’DEFINITION’; $)</code>
> <p>State that the theorem called JUSTIFICATION is a justification theorem for the definition DEFINITION, as described in the section "Appendix 4: A Note on Definitions" of mmset.html
> 
> <code>$j syntax ’TYPECODE’;</code>
> <p>State that TYPECODE is a type code. That is, it is defined by $a but its purpose is to introduce a syntax.
> 
> <code>$j syntax ’SYNTAX’ for 'TYPECODE';</code>
> <p>State that SYNTAX is defined by $a and represents an expression of type code TYPECODE.
> 
> <code>$j unambiguous ’klr 5’;</code>
> <p>State that the grammar is unambiguous, verifiable using the KLR parser, with compositing depth 5.</p>
> 
> <code>$j primitive ’PRIMITIVE1’ ’PRIMITIVE2’;</code>
> <p>State that PRIMITIVE1 and PRIMITIVE2 are the names of $a statements which represent syntax which is only defined implicitly (by axioms), not by a definition.</p>
> 
> <code>$j equality ’SYNTAX’ from ’REFL’ ’SYM’ ’TRAN’;</code>
> Register an equivalence relation. SYNTAX is the $a statement for the syntax, and REFL, SYM, and TRAN are the names of theorems (in inference form) establishing reflexivity, symmetry, and transitivity respectively.
> 
> <code>$j definition 'DEFTHM' for 'SYNTAX';</code>
> Declare a theorem to be a definition. DEFTHM is a $p statement which represents an alternative definition for the syntax represented by the $a statement SYNTAX. (As of July 2022 `$j definition` is only used once in set.mm and I'm not sure what tools, if any, do anything with it).
> 
> <code>$j congruence 'THEOREM';</code>
>State that THEOREM is a congruence law. The conclusion should be an equivalence of two terms, and the hypotheses should state the equivalence (that is, via a syntax which has a $j equality declaration) of each of the subterms which make up those terms.
>
> <code>$j bound 'TYPECODE';</code>
> State that TYPECODE is used for binding variables.
> 
> <code>$j free_var 'SYNTAX' with 'VARIABLE';</code>
> SYNTAX should be the name of a $a statement which defines a syntax containing the variable 
> VARIABLE. State that VARIABLE is free in that syntax.
> 
> <code>$j free_var_in 'SYNTAX' with 'TERM' 'VARIABLE';</code>
> Where SYNTAX is a $a statement which defines a syntax containing both TERM and VARIABLE,
> say that VARIABLE is not bound in TERM in that syntax. This is similar to a $d statement between
> TERM and VARIABLE.
> 
> <code>$j restatement 'AXIOM' of 'THEOREM'; </code>
> This states that a $a statement AXIOM is just a restated version of a $p statement THEOREM.
> This might be desirable if you want to track AXIOM's usage or use other tools that operate on
> axioms, while also making it clear that it is isn't a true axiom, in the sense that it can be proved
> from other axioms as shown at THEOREM.
>
> <code>$j garden_path SEQUENCE-OF-SYMBOLS... ;</code>
> Give a hint to parsers that SEQUENCE-OF-SYMBOLS might introduce more than one possible
> syntax. Some parsers may be able to deduce this from the syntaxes themselves, but
> <code>garden_path</code> is there to aid parsers which need some help.
> 
> <code>type_conversions</code>
> Part of the <code>garden_path</code> mechanism.
>
> <code>natded_init</code>, <code>natded_assume</code>, <code>natded_weak</code>, <code>natded_cut</code>, <code>natded_true</code>, <code>natded_imp</code>, <code>natded_and</code>, <code>natded_or</code>, and <code>natded_not</code>
> See comments in set.mm for more on these. Together they state that a number of specified theorems obey certain rules related to nature deduction.
>
> <code>condequality</code> , <code>condcongruence</code>, and <code>notfree</code>
> These are part of an experiment which involves a <code>CondEq</code> syntax and a metatheorem which can be proved by a system which uses these tags to know what to prove.

jkingdon:
This is text by @jkingdon and @digama0 from a github issue #2664 
The formatting is pretty rough and the content might be too.
But it is at least a start. There is almost zero intentional deviation from the text in that github issue, so any differences are probably mistakes.

I don't think the script which would make this live on the website is yet checked in to the https://github.com/metamath/metamath-website-scripts repository. But I suppose it will be in due course (cc @david-a-wheeler ).

Fixes #2664

### DIFF
--- a/annotate.html
+++ b/annotate.html
@@ -1,0 +1,411 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
+   "http://www.w3.org/TR/html4/loose.dtd">
+<html>
+  <head>
+<META NAME="viewport" CONTENT="width=device-width, initial-scale=1.0">
+    <title>Metamath $j annotations</title>
+  </head>
+  <body>
+<TABLE BORDER=0 WIDTH="100%"><TR>
+<!-- ...TD style="line-height:100%"... -->
+
+<TD NOWRAP ALIGN=LEFT VALIGN=MIDDLE WIDTH="25%"> <FONT SIZE=-2
+FACE=arial><A HREF="../index.html"><IMG SRC="mm.gif" BORDER=0
+ALT="Home" HEIGHT=32 WIDTH=32 ALIGN=LEFT><IMG SRC="spacer.gif" BORDER=0
+ALT="" HEIGHT=32 WIDTH=1 ALIGN=LEFT>Metamath<BR>Home</A></FONT></TD>
+
+<TD NOWRAP ALIGN=CENTER VALIGN=MIDDLE><FONT SIZE="+3"
+COLOR="#006633"><B>Metamath $j annotations</B></FONT></TD>
+
+<TD WIDTH="25%"> &nbsp; </TD>
+
+</TR></TABLE>
+
+<HR NOSHADE SIZE=1>
+<p>As noted in section 4.4.3, "Additional Information Comment ($j)",
+of the metamath specification, we have a way to include information in
+metamath proof files which are not needed for verifying, but which are
+useful for other tools like proof assistants, definition checkers,
+conversion to and from other proof formats, and the like. Although the
+metamath specification has some examples, it seems better to document
+the full list here because they aren't part of the specification in
+the sense of being needed for verification.</p>
+
+<p>Many of these are experimental and may be used by one or a few
+tools.  Or in many cases, are intended for future use by tools but
+the tools haven't been written yet.  Be prepared to look at the
+documentation (or source code) for a particular tool before you make
+too many assumptions about what these comments do, or whether they do
+anything at all. But some of the reasons for documenting them
+in a centralized way are: (1) to register
+names so that different tools don't use the same name in different
+ways, and (2) as a pointer to where more information might be found about
+each comment.</p>
+
+<h2>Syntax and definitions</h2>
+<code>$j syntax 'TYPECODE' ...;</code>
+
+<p>State that TYPECODE is a typecode. That is, it is a constant symbol
+which appears in the leftmost position in statements. This form is
+specifically for variable typecodes, which means they have one or
+more $f statements and are usually omitted from HTML display. Syntax
+axioms (which have this typecode) should not have any repeated
+variables or $d or $e hypotheses.</p>
+
+<p>This command is repeatable by passing a list of typecodes.</p>
+
+<code>$j syntax 'SYNTAX' as 'TYPECODE';</code>
+
+<p>State that SYNTAX is a logical typecode, that is, a constant symbol
+which appears in the leftmost position in statements. This form is
+for typecodes that are not directly used for variables but can appear
+in provable statements, and any statement starting with `SYNTAX` should
+be followed by a grammatically correct expression which parses as
+an expression with typecode `TYPECODE`.</p>
+
+<code>$j primitive 'PRIMITIVE' ...;</code>
+
+<p>State that PRIMITIVE is the name of a $a statement which represents
+syntax which is only defined implicitly (by axioms), not by a
+definition.</p>
+
+<p>This command is repeatable by passing a list of constructors.</p>
+
+<code>$j definition 'DEFTHM' for 'SYNTAX';</code>
+
+<p>Declare a theorem to be a definition. `DEFTHM` is a $p statement which
+represents an alternative definition for the syntax represented by
+the $a statement `SYNTAX`. The definition should have a top level
+equality declared by the `equality` command with the definition on
+the right hand side. (This annotation is only needed when the
+definition does not immediately follow the syntax itself, which
+in set.mm only occurs for df-bi. For most definitions we can
+automatically infer the requisite structure.)</p>
+
+<code>$j justification 'JUSTIFICATION' for 'DEFINITION';</code>
+
+<p>State that the theorem called JUSTIFICATION is a justification
+theorem for the definition DEFINITION, as described in the section
+"Appendix 4: A Note on Definitions" of mmset.html</p>
+
+<code>$j restatement 'AXIOM' of 'THEOREM'; </code>
+
+<p>This is used immediately after an axiom which exactly matches a
+previous theorem statement. This is used when the full set of axioms
+is redundant but there are many overlapping subsystems, so some axioms
+imply the others. Theorem displays can make use of this information
+to decrease the number of listed axioms. For example, if a theorem
+depends on `A, B, C` but `C` can be proven from `A + B`, a theorem
+display might say that the theorem only depends on `A, B` even
+though it makes use of a theorem that depends on `A, C` which
+cannot be minimized in this way.</p>
+
+<code>$j type_conversions;</code>
+
+<p>Adds all "type conversions" to the grammatical parser. These are
+syntax axioms like <code>class x</code> (where <code>x</code> is a
+variable of type <code>setvar</code>) that introduce no syntax but
+allow conversion from one typecode to another.</p>
+
+<code>$j unambiguous 'klr 5';</code>
+
+<p>State that the grammar is unambiguous, verifiable using the KLR
+parser, with compositing depth 5.</p>
+
+<code>$j garden_path 'TOKEN' 'TOKEN' ... =&gt; 'TOKEN' ...;</code>
+
+<p>This is used to help the metamath-knife parser parse expressions
+which can be parsed only after some lookahead. The syntax gives an
+initial token sequence (possibly containing variables representing
+previously parsed subexpressions), and gives a replacement parse
+to consider instead. For example,
+<code>garden_path ( x e. A   =&gt;   ( ph ;</code>
+asserts that after parsing an initial sequence <code>( x e. A</code>,
+it should consider revising the parse to <code>( ph</code> (that is,
+reducing <code>x e. A</code> to a wff) if the next token does not
+continue constructing a syntax form. (In this case, it could be the
+start of <code>( x e. A |- ph )</code> or the <code>x e. A</code>
+may need to be interpreted as a wff if it
+is <code> ( x e. A /\ ph ) </code>.)</p>
+
+<h2>Equality and congruence</h2>
+
+<code>$j equality 'SYNTAX' from 'REFL' 'SYM' 'TRAN';</code>
+
+<p>Register an equivalence relation. SYNTAX is the $a statement
+for the syntax, and REFL, SYM, and TRAN are the names of theorems
+(in inference form) establishing reflexivity, symmetry, and
+transitivity respectively.</p>
+
+<code>$j congruence 'THEOREM' ...;</code>
+
+<p>State that `THEOREM` is a congurence lemma. This should have the
+form</p>
+
+<blockquote>
+  <code>|- a1 = b1 & ... & |- an = bn ==&gt; |- F(a1,...,an) = F(b1,...,bn)</code>
+</blockquote>
+
+<p>where each `=` is the appropriate `equality` for the type and `F` is
+the constructor to prove the congruence about. This is only required
+for `primitive` constructors; for definitions these theorems can be inferred.</p>
+
+<p>This command is repeatable by passing a list of theorems.</p>
+
+<code>$j condequality 'CONDEQ' from 'THEOREM';</code>
+
+<p>This labels a declaration `CONDEQ` as a "conditional equality".
+This is a technical definition used in the proof that equality is a
+congruence; it should be a ternary constructor
+<code>CondEq(x, y, ph)</code> which
+expresses ` ( x = y -> ph )`. The THEOREM should be a proof of
+<code>ph |- CondEq(x, y, ph)</code>, which is one of the structural
+rules required for it.</p>
+
+<code>$j condcongruence 'THEOREM' ...;</code>
+
+<p>The THEOREM should be a proof of
+<code>CondEq(x, y, a1 = b1), ..., CondEq(x, y, an = bn)
+|- CondEq(x, y, F(a1, ..., an) = F(b1, ..., bn))</code> for a
+primitive constructor <code>F</code>. There should be one for every
+primitive constructor. For setvar arguments to binding constructors,
+there should be two versions: one has the form
+<code>CondEq(x, y, A = B) |- CondEq(x, y, F(z, A) = F(z, B))</code>
+where <code>x,z</code> and <code>y,z</code> are distinct, and the
+other should have the form
+<code>CondEq(x, y, A = B) |- CondEq(x, y, F(x, A) = F(y, B))</code>
+where <code>x,B</code> and <code>y,A</code> are distinct. This is
+used in the proof that equality is a congruence.</p>
+
+<p>This command is repeatable by passing a list of theorems.</p>
+
+<h2>Bound and free variables</h2>
+
+<code>$j bound 'TYPECODE' ...;</code>
+
+<p>State that `TYPECODE` is a typecode which permits binding
+constructors, like the <code>x</code> in <code>A. x ph</code>
+or <code>{ x | ph }</code>. These
+typecodes are required to not have any constructors.</p>
+
+<p>This command is repeatable by passing a list of typecodes.</p>
+
+<code>$j free_var 'SYNTAX' with 'BVAR' ...;</code>
+
+<p>SYNTAX should be the name of a $a statement which defines a syntax containing the variable 
+> VARIABLE. State that VARIABLE is free in that syntax.
+
+<p>Declares that variables `BVAR` are not binders in syntax `SYNTAX`,
+and instead represent free variables. This is needed when `SYNTAX`
+is a syntax constructor containing both a variable of `bound` typecode
+and another variable, in which case the default assumption is that
+the syntax is a binder like <code>A. x ph</code>. If the definition
+of syntax <code>foo x ph</code> is instead <code>x = 0 /\ ph</code>
+we want to indicate that <code>x</code> is not binding occurrences
+in <code>ph</code> and <code>x</code> is a free variable in the
+result.</p>
+
+<code>$j free_var_in 'SYNTAX' with 'VAR' 'BVAR';</code>
+
+<p>Declares that occurrences of variable `BVAR` are not bound in
+`VAR` in syntax `SYNTAX`, and instead represent free variables.
+This is needed when `SYNTAX` is a syntax constructor containing both
+a variable of `bound` typecode and another variable, in which case
+the default assumption is that the syntax is a binder like
+<code>A. x ph</code>. If the definition of syntax <code>foo x ph</code>
+is instead <code>x = 0 /\ ph</code> we want to indicate that
+<code>x</code> is not binding occurrences in <code>ph</code>
+and <code>x</code> is a free variable in the result.</p>
+
+<p>This is a more specific version of `free_var` which points to a
+specific variable `VAR` which `BVAR` does not bind, while still
+allowing it to bind occurrences in other variables in the syntax.
+For example if <code>foo A x ph</code> has definition
+<code>A = 0 /\ A. x ph</code> we want to indicate that occurrences
+of <code>x</code> in <code>A</code> are not bound, even though
+<code>x</code> is a binding variable in the notation.</p>
+
+<code>$j notfree 'NF' from 'THEOREM';</code>
+
+<p>This declares <code>NF</code> as a "not-free" operator. The 
+THEOREM should be a proof of 
+<code>NF(x, A), CondEq(x, y, A = B) |- A = B</code> where `=` is the
+appropriate `equality` for the typecode. It makes use of a
+`condequality` to justify the definition.</p>
+
+<h2>Natural deduction based automation</h2>
+
+<code>$j natded_init 'IMP', 'AND', 'TRUE';</code>
+
+<p>Sets up the basic structure for natural deduction based automation.
+Expects arguments indicating the names of implication (the turnstile
+symbol in natural deduction), conjunction (which is used to build
+up the context of a natural deduction sequent), and true (the empty
+context). This command must precede any other `natded_*` commands.</p>
+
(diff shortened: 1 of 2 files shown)


jkingdon:
Any reactions? If no one has suggestions, perhaps we are better off merging this now and worrying about formatting (including whether HTML or another format is best), web site generation, etc, when we get to it.

I think I've incorporated all the suggested text from #2664 and at-mentioned the people who participated there (except perhaps @benjub ).

annotate.html
```diff
@@ -0,0 +1,370 @@
...
+conversion to and from other proof formats, and the like. Although the
+metamath specification has some examples, it seems better to document
+the full list here because they aren't part of the specification in
+the sense of being needed for verification.</p>
+
+<h2>Syntax and definitions</h2>
+<code>$j syntax ’TYPECODE’ ...;</code>
+
+<p>State that TYPECODE is a typecode. That is, it is a constant symbol
+which appears in the left most position in statements. This form is
```
benjub: "leftmost" with no space

**edit**: a few other instances below

annotate.html
```diff
@@ -0,0 +1,370 @@
...
+specifically for variable typecodes, which means they have one or
+more $f statements and are usually omitted from HTML display. Syntax
+axioms (which have this typecode) should not have any repeated
+variables or $d or $e hypotheses.</p>
+
+<p>This command is repeatable by passing a list of typecodes.</p>
+
+<code>$j syntax ’SYNTAX’ for 'TYPECODE';</code>
```
benjub: These are not the same quotes... Is that on purpose ?

**edit:** same thing below. Since this is code, this should be the basic apostrophe everywhere, it seems.

benjub: Isn't it "as" instead of "for" ?

annotate.html
```diff
@@ -0,0 +1,370 @@
...
+<code>$j syntax ’TYPECODE’ ...;</code>
+
+<p>State that TYPECODE is a typecode. That is, it is a constant symbol
+which appears in the left most position in statements. This form is
+specifically for variable typecodes, which means they have one or
+more $f statements and are usually omitted from HTML display. Syntax
+axioms (which have this typecode) should not have any repeated
+variables or $d or $e hypotheses.</p>
+
+<p>This command is repeatable by passing a list of typecodes.</p>
```
benjub: I had trouble understanding "repeatable". Maybe clearer is:
> One can declare multiple typecodes with a single command by `$j syntax 'TYPECODE1' 'TYPECODE2' ...;`

?

**edit**: I checked set.mm and apparently I misunderstood... The command `syntax` as a whole is repeated, but wihtin a single $j-statement...

benjub:
I began to check, but I think I do not know $j-statements well enough to be useful, here...

annotate.html
jkingdon: Good catch, I might have mangled these copying from github (or something). If you look at the `$j` statements in *.mm they contain the https://www.fileformat.info/info/unicode/char/27/index.htm character so that is what we should use here.

jkingdon: Must be `as` because set.mm contains `syntax '|-' as 'wff'` and no line which would match the `for` version.

annotate.html
jkingdon: I think you had it right with the "multiple typecodes with a single command". I don't think set.mm uses this for `$j syntax` but there is an example for `$j natded_assume` which uses the same "repeatable" language in this pull request.

I'm inclined to leave this as-is for now. I don't know if there is something more self explanatory than  "repeatable" but if so I'm not immediately thinking of the best wording.

jkingdon:
> I began to check, but I think I do not know $j-statements well enough to be useful, here...

Your comments looked good to me.

I don't really know the $j additional comments either, I'm just collecting text from @digama0 and/or trying to describe what I see in set.mm.

I embarked on this pull request not because I thought I'd have the knowledge or time to write a great guide to $j additional comments, just that we'd be better off with something as a starting point. So I guess in my mind, that's the threshold. I'm not sure whether the relatively light responses to this pull request/issue means people don't see the need for this documentation, would prefer to give it some time as a pull request for a while with no urgency to merging it soon, are worried about checking in something which might be incomplete or even wrong, plan on commenting but haven't gotten around to it yet, or what.

annotate.html
digama0: Note that the wording was chosen so that it could be reused almost verbatim in the other commands, to help indicate that it's a common pattern. (Alternatively / additionally, the word "repeatable" could be an internal link to some point before or after the big list which explains the pattern in more detail. As @jkingdon says, it's saying that `cmd "foo" "bar" "baz";` has the same meaning as `cmd "foo"; cmd "bar"; cmd "baz";`.)

savask:
>I'm not sure whether the relatively light responses to this pull request/issue means people don't see the need for this documentation ...

For me personally $j statements seems like something external to metamath. As far as I understand, some $j annotations are used by mmj2, some by MM0, etc., so instead of trying to explain all the relevant semantics on the matamath pages, it would seem more natural to simply give links to respective mmj2 or MM0 documentation pages.

digama0:
I consider them more as "extension points" for metamath, since metamath itself is fairly set in stone by design. One reason to have a centralized account of extensions is so that different extensions don't reuse names, but also because set.mm has most of them already, so they aren't entirely external to the project (they are maintained in-tree).

Also, pointing to mmj2 or MM0 only works if these places have documentation for the relevant commands, and I can tell you already that they don't. 

Also, $j commands are not actually tool-specific, even though they may be primarily motivated by a specific tool that wants to implement support for such-and-such. They should always have tool-independent semantics, and having a page document those semantics without specific reference to the tool is one way to ensure this property.

savask:
>One reason to have a centralized account of extensions is so that different extensions don't reuse names ...

That makes sense.

>Also, pointing to mmj2 or MM0 only works if these places have documentation for the relevant commands, and I can tell you already that they don't.

Maybe they should, after all this pull request is doing exactly that but at metamath web pages.

>Also, $j commands are not actually tool-specific, even though they may be primarily motivated by a specific tool that wants to implement support for such-and-such.

What is the real situation though? As I understand, some $j comments are used for mmj2 parsing and most other commands are for MM0 translations. Even though in theory $j annotations could be used by other programs, it would be great at least to mention proof assistants which use specific $j comments.

digama0:
> > Also, pointing to mmj2 or MM0 only works if these places have documentation for the relevant commands, and I can tell you already that they don't.
> 
> Maybe they should, after all this pull request is doing exactly that but at metamath web pages.

Yes, because it's a metamath feature not a MM0 feature.

> > Also, $j commands are not actually tool-specific, even though they may be primarily motivated by a specific tool that wants to implement support for such-and-such.
> 
> What is the real situation though? As I understand, some $j comments are used for mmj2 parsing and most other commands are for MM0 translations. Even though in theory $j annotations could be used by other programs, it would be great at least to mention proof assistants which use specific $j comments.

FYI, mmj2 doesn't have a $j parser at all. To my knowledge, $j parsing is only supported in `metamath-knife`, `mm-scala`, and `mm0-hs from-mm`. If you like we can mention the state of tool support for various commands on this page but I think *that* is external data (and it is likely to change as people create more "smart" verifiers).

These commands are used less than you might think. Many of them were added in advance of tool development; for example mmj2 has an ad hoc definition checker and later $j commands were added to encode all the ad hoc stuff so that the next definition checker doesn't have to be so ad hoc. (This is the stuff that you classified as "for MM0 translations".) Nevertheless, mmj2 *is still an ad hoc definition checker*. It still uses hardcoded stuff just like it used to, so it doesn't "need" the $j commands. But that's not so friendly to non-set.mm databases.

Making a good proof assistant for metamath (i.e. higher level reasoning tools, like you would find in any other proof assistant) requires some knowledge of the syntax, but this is fundamentally at odds with the "logical framework" aspects of metamath: if you don't even know what implication is how are you supposed to use a SAT solver to automate your proof away? Most of the $j commands are targeted at this kind of problem - determining what is what in a generic way so that you don't have to have [`String proofTypecode = "|-"; // TODO: generalize`](https://github.com/digama0/mmj2/blob/b959d6b82d104cd6c481bf48dc283d20e6125587/src/mmj/setmm/SetMMConstants.java#L54-L69) hardcoded in the proof assistant.

savask:
>Yes, because it's a metamath feature not a MM0 feature.

I understand the idea of reusing $j statements in different proof assistants, but calling it a "metamath feature" is in my opinion an overstatement, given that the only programs effectively using $j statements seem to be MM0 tools (that may change in the future, I guess).

>FYI, mmj2 doesn't have a $j parser at all. ...

Oh well. I was looking in the section "Additional information comment ($j)" 4.4.3 of the metamath book which said

>The mmj2 verifier supports many commands in additional information comments.

I guess it meant something else.

digama0:
Here are the commands actually used by the $j parsers that exist:

* [mm-scala](https://github.com/digama0/mm-scala/blob/a7b2f9a4aa34f64e93853fd671568e2b96f43a65/src/main/scala/org/metamath/scala/Parser.scala#L178-L218): `syntax`, `grammar`, `unambiguous`, `equality`, `primitive`, `congruence`, `definition`, `justification`
* metamath-knife: [`syntax`](https://github.com/david-a-wheeler/metamath-knife/blob/071292ef3a10350d0dae9e409b4921f91a08ce95/src/grammar.rs#L419-L431), [`garden_path`](https://github.com/david-a-wheeler/metamath-knife/blob/071292ef3a10350d0dae9e409b4921f91a08ce95/src/grammar.rs#L1129-L1140), [`type_conversions`](https://github.com/david-a-wheeler/metamath-knife/blob/071292ef3a10350d0dae9e409b4921f91a08ce95/src/grammar.rs#L1142-L1155)
* mm0-hs from-mm: [`syntax`](https://github.com/digama0/mm0/blob/5ea40f6f30b467537ca48d961eef0c485ac9a3b9/mm0-hs/src/MM0/FromMM/Parser.hs#L363-L370), [`bound`](https://github.com/digama0/mm0/blob/5ea40f6f30b467537ca48d961eef0c485ac9a3b9/mm0-hs/src/MM0/FromMM/Parser.hs#L371-L376), [`equality`](https://github.com/digama0/mm0/blob/5ea40f6f30b467537ca48d961eef0c485ac9a3b9/mm0-hs/src/MM0/FromMM/Parser.hs#L384-L396), [`free_var`](https://github.com/digama0/mm0/blob/5ea40f6f30b467537ca48d961eef0c485ac9a3b9/mm0-hs/src/MM0/FromMM/Parser.hs#L496-L499), [`free_var_in`](https://github.com/digama0/mm0/blob/5ea40f6f30b467537ca48d961eef0c485ac9a3b9/mm0-hs/src/MM0/FromMM/Parser.hs#L513-L518), [`restatement`](https://github.com/digama0/mm0/blob/5ea40f6f30b467537ca48d961eef0c485ac9a3b9/mm0-hs/src/MM0/FromMM/Parser.hs#L524-L539)
  * Additionally parsed without doing anything with the result: `primitive`, `justification`, `congruence`, `condequality`, `condcongruence`, `notfree`, `natded_init`, `natded_assume`, `natded_weak`, `natded_cut`, `natded_true`, `natded_imp`, `natded_and`, `natded_or`, `natded_not`

digama0:
> > Yes, because it's a metamath feature not a MM0 feature.
> 
> I understand the idea of reusing $j statements in different proof assistants, but calling it a "metamath feature" is in my opinion an overstatement, given that the only programs effectively using $j statements seem to be MM0 tools (that may change in the future, I guess).

I'm not sure how else to put this, but I disagree. The development of MM0 was in fact an outgrowth of the question "what does a generalized definition checker for metamath look like?", and these commands were added to fix ambiguities in metamath itself and declare intent for things that otherwise had to be inferred from context or hardcoded. (A lot of things *still are* inferred from context, to minimize the need for $j annotations.) They are in no way MM0 specific or indeed even expressed in MM0 language; they instead talk about things like "is this a binder" which is relevant to determining whether a definition is consistent with the rest of set.mm.

(In case you think it is odd that something like "is this a binder" would be relevant to definitional consistency, consider that this is an okay definition:
```
$d ph y $. $a |- good ph = { y | ph } $.
```
but this is not:
```
$d x y $. $a |- bad x = ( y u. x ) $.
```
even though they have essentially the same syntactic structure.)

> > FYI, mmj2 doesn't have a $j parser at all. ...
> 
> Oh well. I was looking in the section "Additional information comment ($j)" 4.4.3 of the metamath book which said
> 
> > The mmj2 verifier supports many commands in additional information comments.
> 
> I guess it meant something else.

I think Norm was under the mistaken impression that this had been implemented rather than just planned. Alas, time is finite and I have mostly poured my refactoring time into better and newer tools like mm-knife for this kind of thing.

savask:
>I'm not sure how else to put this, but I disagree. ...

Okay, I'm convinced now.

>I think Norm was under the mistaken impression that this had been implemented rather than just planned.

Maybe it should be added to the errata of the metamath book then.

jkingdon:
> I consider them more as "extension points" for metamath, since metamath itself is fairly set in stone by design. One reason to have a centralized account of extensions is so that different extensions don't reuse names, but also because set.mm has most of them already, so they aren't entirely external to the project (they are maintained in-tree).

I've added the following text to the pull request:

```
Many of these are experimental and may be used by one or a few
tools.  Or in many cases, are intended for future use by tools but
the tools haven't been written yet.  Be prepared to look at the
documentation (or source code) for a particular tool before you make
too many assumptions about what these comments do, or whether they do
anything at all. But some of the reasons for documenting them
in a centralized way are: (1) to register
names so that different tools don't use the same name in different
ways, and (2) as a pointer to where more information might be found about
each comment.
```

which is intended to summarize the balance people have been discussing in this issue, about whether these are official and centralized, or what. Feel free to propose changes to this wording, but perhaps this strikes the right balance in terms of what we do or do not expect out of centralized documentation of $j comments.

jkingdon:
> Here are the commands actually used by the $j parsers that exist

I have put this text (lightly edited) into the pull request.

Although such a list may be hard to keep completely up to date, it is better to have it checked in as part of annotate.html here, where we have a fighting chance of updating it as things change, than leaving it to lore, old github comments, information scattered among many repositories, etc. This also is a version of @savask 's suggestion of "give links to respective mmj2 or MM0 documentation pages." (and I'll add "and/or source code" because in some cases that might be the only documentation for better or worse).

annotate.html
```diff
@@ -0,0 +1,411 @@
...
+conversion to and from other proof formats, and the like. Although the
+metamath specification has some examples, it seems better to document
+the full list here because they aren't part of the specification in
+the sense of being needed for verification.</p>
+
+<p>Many of these are experimental and may be used by one or a few
+tools.  Or in many cases, are intended for future use by tools but
+the tools haven't been written yet.  Be prepared to look at the
+documentation (or source code) for a particular tool before you make
+too many assumptions about what these comments do, or whether they do
```
digama0:
```suggestion
too many assumptions about what these commands do, or whether they do
```
We should try to be consistent with naming: these are "$j commands" or just "commands" rather than comments.

digama0:
> Be prepared to look at the documentation (or source code) for a particular tool before you make too many assumptions about what these comments do, or whether they do anything at all.

I think this leaves the wrong impression, by suggesting that the source code of parsers is the source of truth for the semantics of these commands. I would rather have this page (or maybe a more spec-language version of it) be the source of truth, and the parsers are just implementing that. In particular, it's no good referring people to tools that haven't been written yet; it would probably be better to refer people to the oral history instead (aka "ask on the Metamath group") if this documentation is incomplete.

annotate.html
savask: Note that the metamath book calls them "$j comments" and things inside the $j are referred to as "commands", for instance, it writes:

>Within an additional information comment is a sequence of one or more commands of the form `command arg arg ... ;`

digama0: I think "$j comment" is the descriptor for a comment unit like `$( $j ... $)`, while ($j) command is the term for a command like `syntax "foo";` that can appear in a $j comment. (The $j- prefix is only needed to distinguish between $j commands and $t commands or other kinds of commands not related to these special comment sections.)

jkingdon: Sounds right. In the words of the metamath book, "Within an additional information comment is a sequence of one or more commands of the form [goes on to give details]".

I took a pass to speak more consistently of `$j commands`, to mostly drop the word "comments" and to completely drop the word "annotations" (which I sort of like but not enough to propose displacing existing terminology).

jkingdon:
> would rather have this page (or maybe a more spec-language version of it) be the source of truth, and the parsers are just implementing that. In particular, it's no good referring people to tools that haven't been written yet; it would probably be better to refer people to the oral history instead (aka "ask on the Metamath group") if this documentation is incomplete.

We are walking at least two tightropes here:

* Centralized versus decentralized. Is $j an extension mechanism where we expect people to try different things with relatively little coordination with other tools at least initially (for example, as they try to write things like Metamath Zero and find information missing)? Or more of a specify-then-implement thing which flows out from set.mm?
* Prescriptive versus descriptive. Are we trying to describe what tools do today, or what they might ideally do in some ideal world? (with mmj2 being perhaps the most obvious answer of a discrepancy between what we have and what we might imagine having if people spent more time taking it in a particular direction).

A few words at the top of the file can't fully answer these kinds of questions (and I, personally, am pretty much in the nuanced middle region on both topics), so part of me wonders whether to say anything at all. But I suppose it is worth making at least some attempt to write down what people are thinking, so here's the wording I just added to this pull request:

> Many of these are experimental and may be lightly used in practice.
We would like for this page to be the source of truth which
describes the rules that the tools should implement. But don't
make too many assumptions, including about whether the $j commands
do anything at all in a particular situation. Contact us (for example
via a github issue or the mailing list) if this documentation
is incomplete.

How does that look? Does it perhaps give us at least a tiny bit of clarity about what this document is and is not (at least, as much as we need for now)?

digama0:
> 1. Centralized versus decentralized. Is $j an extension mechanism where we expect people to try different things with relatively little coordination with other tools at least initially (for example, as they try to write things like Metamath Zero and find information missing)? Or more of a specify-then-implement thing which flows out from set.mm?

Generally, a $j command is only really useful if it is to be integrated into set.mm itself. You can always make your own list on the side of data associated to labels in set.mm, but it runs the risk of going out of date when things get renamed or new things are added. So if you want these things to last in the long term that means coming up with a $j command with at least a rough spec and PR-ing it to set.mm, at which point you would also add it to the docs page.

This process means that it is necessarily a bit centralized. Of course you could have your own .mm file with your own $j commands but at that point it's a separate system and you have total control to do whatever you like; the "list on the side" approach doesn't have any desync issues in this case so it's probably simpler.

Historically, Norm has been fairly uninterested in the nature of these $j commands and permissive in adding them, which perhaps has lead to a dearth of documentation and specification for them. I would like them to go through a more thorough review in the future, but since I've been the one proposing most of them I haven't made much of a fuss about it.

> 2. Prescriptive versus descriptive. Are we trying to describe what tools do today, or what they might ideally do in some ideal world? (with mmj2 being perhaps the most obvious answer of a discrepancy between what we have and what we might imagine having if people spent more time taking it in a particular direction).

The state of tooling around each command is an interesting side-note, but I don't think it's the most important thing by a long shot. One thing that is useful to glean from the data is which commands are the most "popular" / important, like `syntax`, and which support only niche features like `natded_*` or `condequality`. But I think in the ideal state all .mm files would have all applicable $j commands applied to them: they each represent information that cannot be reliably derived from the file in another way.

In short, this is a prescriptive list. It's not describing what the tools do with the information but rather what the information actually is and how to apply it to an arbitrary .mm file.

jkingdon:
@digama0 Do you have a proposed wording? Or a suggestion to remove that paragraph? I'd prefer to get this pull request done and merged.

digama0:
No, your wording is fine. That was just context and an attempt to answer your questions.

j-commands.html
```diff
@@ -0,0 +1,408 @@
...
+<p>This command is repeatable by passing a list of theorems, or it
+can be split into multiple commands; the `'NOT' 'FALSE' with` part
+must be repeated in this case. Multiple `NOT` and `FALSE` constants
+can also be introduced by using multiple commands, and a negation
+can associate to multiple falsity constants and vice versa.
+The theorems provided should not mention any other constructors
+than the specified `NOT` and `FALSE` however (and
+`IMP`, `AND`, `TRUE` from `natded_init`).</p>
+
+<h2>Usage of these commands</h2>
```
digama0:
```suggestion
<h2>Usage of these commands in existing verifiers</h2>
```

j-commands.html
```diff
@@ -0,0 +1,408 @@
...
+`simpld` and `simprd` are recommended.)</p>
+
+<p>This command is repeatable by passing a list of theorems,
+or it can be split into multiple commands; the `'AND' with` part
+must be repeated in this case. Multiple `AND` constants can also be
+introduced by using multiple commands.</p>
+
+<code>$j natded_or 'OR' with 'THEOREM' ...;</code>
+
+<p>Registers `THEOREM` as a natural deduction rule pertaining to
```
digama0:
```suggestion
<p>Registers <code>THEOREM</code> as a natural deduction rule pertaining to
```
Backticks were used in the original post for code spans. A search/replace for " \`" and "\` " / "\`," should get most of them.

jkingdon: Done.

PR #2694, opened 2022-07-18
Superfluous hypotheses removed (4)+(5); deduction versions for inferences; generalization of map1

avekens:
~lssneln0 revised by removing a superfluous hypothesis according to the list of [@savask](https://github.com/savask), see issue [#2648](https://github.com/metamath/set.mm/issues/2648):

* Theorems depending on this theorem (mostly in Norm's Mathbox) were revised. There is one theorem (hmapcl) which is commented out and using ~lssneln0. I left ~lssneln0OLD within its proof, but I think it is very difficult to keep such theorems/proofs maintained in the long run.

~mapsnd and ~mapsnend moved from GS's mathbox to main set.mm
* Comments revised: the Contributor of a new(er) theorems in deduction form should be the contributor of the already existing inference form (especially if the old proof is only slightly modified). The author of the new theorem should be mentioned in the "revised by" tag (see also the comment for ~snmapen, where I declared Norm to be the contributor, because I used the main ideas of the proof of Norm's proof of ~map1. By shortening the prove of ~map1, Norm's work vanishes here, so it should continue to live in ~snmapen). Maybe this topic should/must be discussed further elsewhere...
* proofs of ~mapsn and ~mapsnen shortended using ~mapsnd and ~mapsnend
* proofs shortened using ~mapsnend instead of ~mapsnen: ~map2xp, ~mapdom3, ~ackbij1lem5, ~pwxpndom2, ~hashmap: the former versions of these proofs transformed the class variable in the antecedent into a setvar variable (by applying ~vtoclg), so that the hypotheses of the theorems  in inference form were satisfied. This transformation is not necessary anymore using the deduction forms of the theorems.

generalizations ~snmapen and ~snmapen1 of ~map1 added (see also discussion in #2692) 
* proof of ~map1 shortened

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+17-Jul-22 mapsnend  [same]      moved from GS's mathbox to main set.mm
+17-Jul-22 mapsnd    [same]      moved from GS's mathbox to main set.mm
 16-Jul-22 bj-cbv3v2 cbv3v2      moved from mathbox to main
 12-Jul-22 equviniva equvinva    correct a typo
 11-Jul-22 prodge02   ---        deleted; use prodge0ld instead
--- a/discouraged
+++ b/discouraged
@@ -16604,6 +16604,7 @@ New usage of "lshpnel2N" is discouraged (0 uses).
 New usage of "lshpset2N" is discouraged (1 uses).
 New usage of "lspfixedOLD" is discouraged (0 uses).
 New usage of "lspsncv0OLD" is discouraged (0 uses).
+New usage of "lssneln0OLD" is discouraged (0 uses).
 New usage of "lssssrOLD" is discouraged (0 uses).
 New usage of "ltaddnq" is discouraged (7 uses).
 New usage of "ltaddpr" is discouraged (5 uses).
@@ -19669,6 +19670,7 @@ Proof modification of "lcmftp" is discouraged (1057 steps).
 Proof modification of "lediv2aALT" is discouraged (264 steps).
 Proof modification of "lspfixedOLD" is discouraged (1103 steps).
 Proof modification of "lspsncv0OLD" is discouraged (149 steps).
+Proof modification of "lssneln0OLD" is discouraged (64 steps).
 Proof modification of "lssssrOLD" is discouraged (119 steps).
 Proof modification of "ltrnmwOLD" is discouraged (292 steps).
 Proof modification of "luk-1" is discouraged (73 steps).
--- a/discouraged
+++ b/discouraged
@@ -18331,6 +18331,7 @@ New usage of "w-bnj17" is discouraged (103 uses).
 New usage of "w-bnj19" is discouraged (8 uses).
 New usage of "watfvalN" is discouraged (1 uses).
 New usage of "watvalN" is discouraged (1 uses).
+New usage of "wfrlem4OLD" is discouraged (0 uses).
 New usage of "wl-a1d" is discouraged (1 uses).
 New usage of "wl-a1i" is discouraged (2 uses).
 New usage of "wl-ax1" is discouraged (3 uses).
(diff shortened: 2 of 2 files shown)


benjub:
Thanks for the careful work.

Avoiding use of vtoclg, which is a "hard" theorem (even though I removed its dependency on ax-11 and ax-13 recently), is a nice bonus.

Looking at the statement of lssneln0 with the extra hypothesis removed, `V` is not related to the rest, so the natural conclusion is `|- ( ph -> -. X = .0. )` and the hypothesis where `V` occurs can also be removed.  The current proof confirms it: the two-but-last step is precisely `|- ( ph -> -. X = .0. )`.  I looked at a few usages, and when llsneln0 is used, it is often followed by the theorem `( A e. V \ { (/) } -> -. A = (/) )`, which confirms that the above statement is more natural.

Unfortunately, the idiom `e. V \ { (/) }` propagates in a few other results, which may make some proofs uselessly longer.

Could you uncomment that theorem in Norm's mathbox (hmapcl) ? (**edit:** I just had a look and it's actually not an isolated commented theorem... we'll have to look at that another day, though the more we wait, the more proofs may become broken)

As for credits, I agree: if a new theorem is a slight variation and relies heavily on an existing theorem (as in xxxd, xxxi, xxxg or xxxt if xxx already exists), then the indicated contributor should be the contributor of the original theorem, especially if the original theorem may be removed in favor of that new variation. The new contributor should be indicated in the "Revised by" tag, ideally explaining the nature of the revision. This could go in ~conventions or ~conventions-contrib for instance.

jkingdon (approved):

jkingdon (approved):

avekens:
> Looking at the statement of lssneln0 with the extra hypothesis removed, `V` is not related to the rest, so the natural conclusion is `|- ( ph -> -. X = .0. )` and the hypothesis where `V` occurs can also be removed. The current proof confirms it: the two-but-last step is precisely `|- ( ph -> -. X = .0. )`. I looked at a few usages, and when llsneln0 is used, it is often followed by the theorem `( A e. V \ { (/) } -> -. A = (/) )`, which confirms that the above statement is more natural.
> 
> Unfortunately, the idiom `e. V \ { (/) }` propagates in a few other results, which may make some proofs uselessly longer.

According to these remarks, I added a new theorem ~ lssvneln0 (part of ~lssneln0). With this theorem, the proofs of ~lssneln0 (of course), ~lvecindp, ~lshpne0 were shortened. The proofs of ~lspexchn1 and ~baerlem5amN  could not be shortened. I stopped the investigation here, because the remaining theorems are in Norm MAthbox. Therefore, I hope the current state is acceptable.

> Could you uncomment that theorem in Norm's mathbox (hmapcl) ? (**edit:** I just had a look and it's actually not an isolated commented theorem... we'll have to look at that another day, though the more we wait, the more proofs may become broken)

I tried already, but there were too many dependencies on other deactivated theorems. I will try again...

avekens:
> > Could you uncomment that theorem in Norm's mathbox (hmapcl) ? (**edit:** I just had a look and it's actually not an isolated commented theorem... we'll have to look at that another day, though the more we wait, the more proofs may become broken)
> 
> I tried already, but there were too many dependencies on other deactivated theorems. I will try again...

I am afraid that the proofs of these theorems cannot be used anymore: A lot of definitions seem to be missing (all labels with the suffix SBC, for example). To recover these proofs, an older version of set.mm has to be used. To update this would be a lot of work. @benjub Maybe you want to create an issue to clarify how to proceed with these theorems.

benjub (approved):

benjub:
> I am afraid that the proofs of these theorems cannot be used anymore: A lot of definitions seem to be missing (all labels with the suffix SBC, for example). To recover these proofs, an older version of set.mm has to be used. To update this would be a lot of work. @benjub Maybe you want to create an issue to clarify how to proceed with these theorems.

This looks like the only big commented chunk in Norm's mathbox, together with the one beginning with the section title "Projective geometries based on Hilbert lattices".  Note sure what to do about these, not even sure enough to open a new Issue. As @avekens says, it would be a lot of work to "recover" them.  Is there enough motivation ?  Do @jkingdon or @digama0 have thoughts on this ?

jkingdon:
> it would be a lot of work to "recover" them. Is there enough motivation ? Do @jkingdon or @digama0 have thoughts on this ?

Of course they can be recovered with enough work (going back to an old version of set.mm is a good suggestion in terms of things like figuring out what references are to theorems which are now renamed or revised). Is it worth it? I'm not especially able to help on that, as when I look at " Closure of map from vectors to functionals in the closed kernel dual space" I'd have to research the meaning of almost every word in that sentence (with the possible exception of "vector" - even "closed" is pretty context-dependent). I also didn't try to research why they were commented out.

I don't think there is any particular rush if no one has the motivation and time right now to either rescue them or dig into what they are for and whether there is some other path.

PR #2695, opened 2022-07-18
fix typo + add mathbox + utility theorems

icecream17:
I started trying to write proofs about combinatorial games, but little did I know that there are not only one but two other descriptions! <sup>https://us.metamath.org/mpeuni/df-pg.html and https://us.metamath.org/mpeuni/df-no.html</sup>

After some simplifications my definition ended up being... probably exactly equivalent to `df-pg`:

```
gGen = rec ( ( x e. _V |-> ( ~P x X. ~P x ) ) , (/) )
Fpg = U. ( gGen " _om )
```

Not sure what to do about that.

---
In the meantime here's some utility theorems.

Also fixed typo in `CONTRIBUTING.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,7 +117,7 @@ the automated verification checks to check those.)
 
 In almost all cases, especially when you're starting, you won't need to
 regenerate the `discouraged` file. So if you're just starting out, you can
-probably ignore this section.However, there are rare cases when that
+probably ignore this section. However, there are rare cases when that
 is needed. This section explains what the discouraged file is all about.
 
 Some statement descriptions in `set.mm` and ` iset.mm` have one or both of the


set.mm
```diff
@@ -611012,6 +611013,84 @@ fixed reference functional determined by this vector (corresponding to
...
+
+  ${
+    opelxpis.1 $e |- A e. C $.
+    opelxpis.2 $e |- B e. D $.
+    $( Ordered pair membership in a Cartesian product (implication).
+       (Contributed by Steven Nguyen, 17-Jul-2022.) $)
+    opelxpis $p |- <. A , B >. e. ( C X. D ) $=
+      ( wcel cop cxp opelxpi mp2an ) ACGBDGABHCDIGEFABCDJK $.
+  $}
+
```
avekens, line 611090: Unfortunately, the naming conventions do not hold for the opexp family: ~opelxp should be called ~opelxpb (because it is a logical equivalence), then ~opelxpi could be called ~opelxp (theorem in closed form) and finaly ~opelxpis could be called ~opelxpi  (it is in inference form). But ~opelxp and ~opelxpi are used very often, so they should not be changed. For the moment, however, I woud propse to rename ~ opelxpis to ~opelxpii: the first "i" for implication, the second for "inference". (BTW: what does the suffix "s" mean in this case?).

set.mm
```diff
@@ -611012,6 +611013,84 @@ fixed reference functional determined by this vector (corresponding to
...
+    wceq c0 intnanrd 3syl wb xp11 mtbird neqne syl df-pss sylanbrc ) ABEZCDEZFZ
+    ACGZBDGZHZUMUNIZUMUNEUJABHCDHUOUKABJCDJABCDKLULUNUMSZMZUPULUQBASZDCSZFZULUJ
+    BAIZVAMUJUKNUJABABOPVBUSUTBAQUAUBUJBTIDTIUQVAUCUKABRCDRBDACUDLUEURUNUMUNUMU
+    FPUGUMUNUHUI $.
+
+  ${
+    elpw2i.1 $e |- B e. _V $.
+    $( Membership in a power set, biimplication.  (Contributed by Steven
+       Nguyen, 17-Jul-2022.) $)
+    elpw2i $p |- ( A C_ B <-> A e. ~P B ) $=
```
avekens, line 611077: I would call this theorem ~elpwbi ("b" for biconditional, "i" for "infernece because of the hypothesis).

avekens (changes requested):
The utility theorems look very nice, and there are good chances that they are also used by others (in this case they will be moved to the main body of set.mm). I only request two changes of labels, otherwise this PR is fine.

benjub:
> Not sure what to do about that.

Maybe, once you've sorted things out, ask these two contributors (creating an issue here and pinging them), and see if their work is ongoing and if they want to join forces or coordinate.

It is best to avoid having closely related but different definitions in set.mm for basically the same concept.

set.mm
benjub: Yes. And "biconditional" instead of "biimplication" in the comment.

set.mm
icecream17: "s" meant syllogism (changed in commit 2)

benjub (approved):

PR #2697, opened 2022-07-20
revert chvarv to OLD version

wlammen:
1. revert chvarv to its OLD version.  The comment says the OLD version was dependent on ax-10, but somehow it lost this dependency, maybe because chvar also lost this dependency later.  Anyway, the reason for the revision is not valid (any longer), so discard the revised version in favor of the shorter original version.
2. Move dvelimdh out of a cluster of cbvar* theorems.
3. Remove some outdated OLD theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14479,13 +14479,7 @@ New usage of "cases2ALT" is discouraged (0 uses).
 New usage of "cayleyhamiltonALT" is discouraged (0 uses).
 New usage of "cba" is discouraged (85 uses).
 New usage of "cbncms" is discouraged (5 uses).
-New usage of "cbval2vOLD" is discouraged (0 uses).
-New usage of "cbvaldvaOLD" is discouraged (0 uses).
-New usage of "cbvalvOLD" is discouraged (0 uses).
-New usage of "cbvex2vOLD" is discouraged (0 uses).
-New usage of "cbvexdvaOLD" is discouraged (0 uses).
 New usage of "cbvexsv" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -19139,6 +19092,23 @@ This axiom scheme is logically redundant (see ~ ax12w ) but is used as an
       ACJBDJSUARTCDADEKBCFKCDLABGMNOACPBDPQ $.
   $}
 
+  ${
+    $d x z $.  $d y z $.
+    dvelimhw.1 $e |- ( ph -> A. x ph ) $.
+    dvelimhw.2 $e |- ( ps -> A. z ps ) $.
+    dvelimhw.3 $e |- ( z = y -> ( ph <-> ps ) ) $.
+    $( dvelimhw.4 $e |- ( -. A. x x = y -> ( z = y -> A. x z = y ) ) $. $)
```
benjub, line 19100: I think you can remove that commented line (probably a remnant of a variable renaming).

benjub (approved):

set.mm
wlammen: done

benjub:
> * revert chvarv to its OLD version.  The comment says the OLD version was dependent on ax-10, but somehow it lost this dependency, maybe because chvar also lost this dependency later.  Anyway, the reason for the revision is not valid (any longer), so discard the revised version in favor of the shorter original version.

Good catch! The removal of the ax-10 dependency may come from nfv and the change in df-nf.

benjub (approved):

PR #2698, opened 2022-07-20
Misc

benjub:
Mostly janitorial work:
* some typos;
* carry out some TODOs: removed hdmap1neglem1N, hdmap1p6N which were to be deleted, and adapted comments so as not to lose information;
* carry out the TODO of ifeq123d (i.e. adapt proofs using it and then remove it); there are two others like that, of a theorem identical to one in Main; I think in the future, this kind of TODOs could be asked before merging (especially since many fourirerdlem* are barely manageable due to their length);
* iset.mm: add "TODO: xnn0tonninf".

I don't know the author of
> $( TODO: After introducing a definition of "linear combination", it should be
shown that these definitions are actually equivalent. $)

which I kept, but I think it should be removed. There is already a definition of "span" so I think a definition of "linear combination" is not necessary. I also think only one of http://us2.metamath.org/mpeuni/df-lindf.html and http://us2.metamath.org/mpeuni/df-linds.html is necessary.

@avekens : if you have a proposal for a better label for brfi1indlem, I'll make the change in this PR.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,12 +27,13 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+20-Jul-22 brfi1indlem hashdifsnp1
 17-Jul-22 mapsnend  [same]      moved from GS's mathbox to main set.mm
 17-Jul-22 mapsnd    [same]      moved from GS's mathbox to main set.mm
 16-Jul-22 bj-cbv3v2 cbv3v2      moved from mathbox to main
 12-Jul-22 equviniva equvinva    correct a typo
-11-Jul-22 prodge02   ---        deleted; use prodge0ld instead
-11-Jul-22 prodge0i   ---        deleted; use prodge0rd instead
+11-Jul-22 prodge02  ---         deleted; use prodge0ld instead
+11-Jul-22 prodge0i  ---         deleted; use prodge0rd instead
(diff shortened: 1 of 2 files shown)


set.mm
```diff
@@ -144860,9 +144859,10 @@ Proper unordered pairs and triples (sets of size 2 and 3)
...
 $)
 
-  $( TODO-AV1: no lemma, but self-reliant theorem!  Lemma for ~ brfi1ind :  The
-     size of a set is the size of this set with one element removed, increased
-     by 1.  (Contributed by Alexander van der Vekens, 7-Jan-2018.) $)
+  $( The size of a finite set minus one of its elements is equal to the size of
+     that set minus one.  TODO-AV1: find a better label, since not only a lemma
+     but interesting in its own right.  (Contributed by Alexander van der
+     Vekens, 7-Jan-2018.) $)
   brfi1indlem $p |- ( ( V e. W /\ N e. V /\ Y e. NN0 )
```
avekens, line 144866: maybe hashdifsnp1? hashdifsn already exists ( `` ... -> ( # ` ( A \ { B } ) ) = ( ( # ` A ) - 1 )``), and "p1" stands for "plus 1" (i.e., for ` ( Y + 1 ) `).

benjub: And it has the same length, so no rewrap needed !

PR #2699, opened 2022-07-21
remove ax-12 from nfeqf2

wlammen:
1. Remove a dependency on ax-12 from nfeqf2.  In the wake nfeqf1 is affected as well.  Maybe preferred usage of these two can strip this dependency from other theorems as well.  This proof was essentially established in my mathbox as wl-dveeq12, and its potential got unnoticed for a year.  I did not bother creating an OLDOLD version, as I shortened this theorem only a fortnight ago.
2. micro-optimization: Avoid temporary results like -. A. x -. ph <-> -. A. x -. ps with lots of NOT by a smarter arrangement of proof steps.  This does not affect the proof length, but has a (small) impact on the visual complexity of the proof on the web site.  As agreed some time ago, this is neither accompanied by a shortening tag, nor by an OLD version.


set.mm
```diff
@@ -20160,8 +20160,9 @@ theorem as an axiom of set theory (Axiom 0 of [Kunen] p. 10).  In the
        (Contributed by NM, 14-Sep-2003.)  (Revised by Mario Carneiro,
        6-Oct-2016.)  (Proof shortened by Wolf Lammen, 16-Jun-2019.) $)
     cbvex2 $p |- ( E. x E. y ph <-> E. z E. w ps ) $=
-      ( wn wal wex nfn weq wa notbid cbval2 2exnaln notbii 3bitr4i ) ALZDMCMZLB
-      LZFMEMZLADNCNBFNENUDUFUCUECDEFAEGOAFHOBCIOBDJOCEPDFPQABKRSUAACDTBEFTUB $.
+      ( wex wn wal nfn weq wa notbid cbval2 2nexaln 3bitr4i con4bii ) ADLCLZBFL
+      ELZAMZDNCNBMZFNENUCMUDMUEUFCDEFAEGOAFHOBCIOBDJOCEPDFPQABKRSACDTBEFTUAUB
+      $.
```
benjub, line 20165: Ah, yes !  I see how in these four proofs, use of alnex (which is df-ex followed by con2bii) together with con4bi*, instead of df-ex, makes things more streamlined.

set.mm
```diff
@@ -20290,11 +20291,12 @@ requires either a dv condition ( ~ sb56 ) or a non-freeness hypothesis
...
-       Wolf Lammen, 9-Jun-2019.)  (Proof shortened by Wolf Lammen,
-       9-Jul-2022.) $)
+       Wolf Lammen, 9-Jun-2019.)  Remove dependency on ~ ax-12 .  (Revised by
+       Wolf Lammen, 20-Jul-2022.) $)
     nfeqf2 $p |- ( -. A. x x = y -> F/ x z = y ) $=
-      ( weq wal wn wex wnf exnal nfnf1 ax13lem2 ax13lem1 syld nfd exlimi sylbir
-      ) ABDZAEFQFZAGCBDZAHZQAIRTASAJRSARSAGSSAEABCKABCLMNOP $.
+      ( weq wal wn wex wnf exnal hbe1 ax13lem2 ax13lem1 syldc com12 hbe1a syl56
+      aleximi nfd sylbir ) ABDZAEFTFZAGZCBDZAHTAIUBUCAUCAGZUDAEZUBUCAEZAGZUFUCA
+      JUEUBUGUDUAUFAUAUDUCUFABCKABCLMQNUCAOPRS $.
```
benjub, line 20299: That's nice !
I just noticed that your hbe1a is identical to http://us2.metamath.org/mpeuni/bj-modal5e.html. No need to modify the credits. Proofs look the same size ? I'll uniformize the comments.

avekens (approved):

benjub (approved):

PR #2700, opened 2022-07-21
rm extra hyp: itgulm2, xrhmeo, gsummptun

benjub:
Remove extra hypotheses (see #2648) from itgulm2, xrhmeo, gsummptun.


savask:
Perhaps not very important, but the proof of `itgulm2` can be minimized with `mptex2`.

avekens (approved):

PR #2701, opened 2022-07-21
rename two files in /people and /scripts to README.md

benjub:
Rename the "description" files in /peope and /scripts to "README.md" in order to be more in line with github standard practices, and to be able to read them directly on the website. Blank lines are there to ensure line breaks in the markdown-formatted output.

### DIFF
--- a/people/000-README
+++ b/people/000-README
@@ -1,1 +1,0 @@
-Please modify the "urls" file to record the source of a new avatar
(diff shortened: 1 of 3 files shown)


jkingdon (approved):
Part of me wants to resist this kind of forced uniformity, but :shrug: I don't think there is a huge amount of character or charm in the old filenames here.

benjub:
> Part of me wants to resist this kind of forced uniformity, but shrug I don't think there is a huge amount of character or charm in the old filenames here.

The thing is that the content of the file can now be seen directly on https://github.com/metamath/set.mm/tree/develop/scripts
I will also try to revive #1887...

PR #2702, opened 2022-07-22
Conventions and misc.

avekens:
Conventions:
* Theorem forms: special case for theorems in inference form
* Characters to be used for labels: only lower case chacters allowed (with exceptions) -see my remark in PR #2681 and Norm`s remark in 2014 (Google group https://groups.google.com/g/metamath/c/Z_ERHq4knJg/m/cbUarfKEQqoJ)
* label fragments revised/enhanced
* bullet point "Contributors:" added: Who should be the "contributor" of a theorem? As proposed by BJ, see discussion in PR #2694.
* bullet point "Usage of Parentheticals." added: How to use "Revised by" and "Proof shortened by" (this was discussed some time ago - I don't know when/where anymore...).

There were some discussions about our conventions in the last weeks, therefore I wanted to take up their results and fix them in set.mm. Of course they still can be discussed. 
@benjub , @jkingdon, @david-a-wheeler , @digama0 , @tirix , @wlammen , please have a closer look at them.

Miscellaneous:
* TODO in ~om0x added


set.mm
```diff
@@ -420312,6 +420323,35 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
+       theorem and added it to set.mm. She/he is mentioned in the "Contributed
+       by" parenthetical within the comment of the theorem, see Metamath Book,
+       p. 142.
+       An exception should be made if a theorem is an (essential) extract or a
+       variant of an already existing theorem.  An example of the latter is
+       the provision of a theorem in deduction form as variant of a theorem in
+       inference form, e.g., ~ mapsn and ~ mapsnd .  In these cases, the
+       contributor of the former theorem (whose proof can often be shortened by
+       the new theorem) should also be the contributor of the new theorem.  The
+       one who provided the new theorem should be mentioned in hte "Revised by"
```
wlammen, line 420337: typo hte -> the

set.mm
```diff
@@ -420312,6 +420323,35 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
+       parenthetical (see Metamath Book, p. 143) within the comment of the new
+       theorem.
+       </li>
+
+       <li><b>Usage of Parentheticals.</b>
```
wlammen, line 420342: This means you give up on the idea of a proof / theorem history overview.  I never knew how important this was to others, that's why I always added a new credits tag, even if my name was already there.

set.mm
```diff
@@ -420312,6 +420323,35 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
        </p>
 
        <ul>
+       <li><b>Contributors.</b>
+       The (original) contributor of a theorem is the first one who proved the
```
wlammen, line 420327: Well, you don't prove an axiom or definition, but it still carries a contributed tag.

set.mm
jkingdon: I've never been a fan of trying to put an extensive history here. It is both overcomplete, in the sense that it gives a lot of detail which won't be relevant to most readers, and undercomplete, in the sense that a string of parentheticals gives little sense for what changed each time. I would say that a tool like git is better suited to keeping track of all the ways that something changed.

jkingdon:
Looks OK to me. Hopefully we won't treat any of this as being so rigid that we spend all our time trying to figure out and apply the rules, but it seems like a plausible set of practices.

set.mm
```diff
@@ -420312,6 +420323,35 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
+       theorem and added it to set.mm. She/he is mentioned in the "Contributed
+       by" parenthetical within the comment of the theorem, see Metamath Book,
+       p. 142.
+       An exception should be made if a theorem is an (essential) extract or a
```
wlammen, line 420331: I am not sure whether this is too general.  I would like to hear how a case like [wl-2sb6d](http://us2.metamath.org/mpeuni/wl-2sb6d.html) is handled.  This further leads to the question how extensions of theorems in mathboxes should look like.  Note that they are often moved unchanged to main in case of need.  I think, the mover is responsible for all aspects of his action, but I like to see this confirmed somewhere.

wlammen:
> ... we spend all our time trying to figure out and apply the rules ...

This is effectively avoided by unambiguous statements.

set.mm
avekens: Fixed.

set.mm
avekens: Maybe "The (original) contributor of an assertion (theorem, definition or axiom) is the first one who proved/provided the assertion ..." would be better?

wlammen: I think that is exact enough

benjub:
I think that's good (with @avekens's proposed modification above). It is hard to find a balance between overly general rules, which may be too verbose if every case has to be precisely handled, and rules which are too vague.
It looks like we are all reasonable here, so we know when to make an exception when needed, and the current proposal seems sufficient.

set.mm
benjub: For instance in wl-2sb6d, the differences in the proof look substantial enough that you remain the contributor. If it were to *replace* 2sb6 by Norm, then there would be no credit to Norm left, so you could write in the comment of 2sb6d something like "replaces an earlier version contributed by NM on ..."

wlammen: Your reaction to this real-world example points out there is somewhere a zone of uncertainty, where decisions flip and people may not always agree.  Okay.  Its fine to take a risk, as long as you are aware of it.

set.mm
avekens: Sentence revised accordingly.

avekens:
Many thanks to all who commented on this PR (and gave some good hints for improvements). I think that we all agree that such conventions are helpful and at least not wrong. They should serve as guidelines for contributors as well as for reviewers, and should avoid unneccessary discussions. Exceptions, however, can alwaly be discussed and made, if there is a consent. And the conventions can even be changed/adapted in the future if necessary or helpful. They are not  set in stone...

PR #2704, opened 2022-07-23
Mathbox: rm wl-sb5nae, wl-dveeq12

wlammen:
1. micro optimization: drex2 looks smaller on the web page, but needs the same amount of proof bytes.  No OLD theorem, no credits.
3. mathbox: eliminate wl-sb5nae, a duplicate of bj-sb3b, the same proof, but a later contribution date.  Follow the "prior art" principle.
4. remove wl-dveeq12: its proof is found in that of nfeq2f

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1482,11 +1482,9 @@
 "ax13lem1" is used by "nfeqf2OLD".
 "ax13lem1" is used by "wl-19.2reqv".
 "ax13lem1" is used by "wl-19.8eqv".
-"ax13lem1" is used by "wl-dveeq12".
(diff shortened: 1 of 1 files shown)


wlammen:
@benjub: what about hbe1a?

benjub:
@wlammen : hbe1a: your proof is actually shorter by one byte (compressed format) than mine (bj-modal5e), so leave hbe1a that way and I will remove bj-modal5e.  (this duplication was discovered in #2699)

I will move bj-sb3b to Main, since it appeared in two mathboxes.

benjub (approved):

PR #2705, opened 2022-07-23
add elovolmlem and automatic minimization of 11 proofs, each by about…

benjub:
Add elovolmlem, corresponding to the line
> 35	0	["reex","reex","xpex","inex2","nnex","elmap"]

in @savask's list at https://gist.github.com/savask/dce5b48b7317e8144635fda09e983fae (see thread https://groups.google.com/g/metamath/c/aXvD9rs36Ps). Automatic minimization of 11 proofs, each by about 40 bytes.

@savask : is there a way to update that list ? or maybe turn it into an Issue here, with checkboxes ?


benjub:
Second commit: remove a duplicate from the list at #2128.

avekens (commented):
I think ~elovolmlem as lemma is OK here (maybe as example), but we should not continue to look for such technical lemmata to save some bytes in a very bounded context. The list of @savask should serve (at least in my opinion) as indication which general purose theorems can be derived from it, to gain benefit from them also in the future.

benjub:
> I think ~elovolmlem as lemma is OK here (maybe as example), but we should not continue to look for such technical lemmata to save some bytes in a very bounded context. The list of @savask should serve (at least in my opinion) as indication which general purose theorems can be derived from it, to gain benefit from them also in the future.

I agree that ~elovolmlem is only a lemma, which is why I chose that label and associated comment. Since it is applied 11 times and saves more bytes than the total size of its statement+proof+comment, it is justified, as other cases with similar properties.  I agree that not all theorems in the list should be added blindly.

PS: I chose that particular one in the list because it was the longest theorem with zero essential hypotheses.  I remember I saw somewhere that you (@avekens) were going to take care of the other hypotheses-free theorems ? (you shouldn't feel obliged if that's too much work, you can say which ones you plan to do and I can choose a few others) (**edit**: I see I missed Line 107: 30	0	["elequ2","anbi1d","exbidv","bibi2d","albidv","imbi2d","exbidv"] ; I'll do it)

PR #2707, opened 2022-07-24
Intuitionize the rest of the section The # (set size) function in iset.mm

jkingdon:
In some cases the theorems intuitionize without much trouble. In some cases they probably won't. In other cases, there are things we may end up doing in the future but this is pretty much a first pass through this section.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7117,6 +7117,13 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>hashres , hashreshashfun</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably would need to add ` B e. Fin ` or similar
+  conditions.</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7112,7 +7112,8 @@ than reals.</TD>
 <TR>
   <TD>hashfun</TD>
   <TD><I>none</I></TD>
-  <TD>The set.mm proof will not work as-is</TD>
+  <TD>Presumably provable but the set.mm proof relies on excluded
+  middle and undif.</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3128,6 +3128,12 @@ is Theorem 1 of [PradicBrown2021], p. 1.</TD>
 </TR>
 
 <TR>
+  <TD>pwen</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on pw2eng</TD>
+</TR>
+
+<TR>
   <TD>php</TD>
   <TD>~ phpm</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3134,6 +3134,12 @@ is Theorem 1 of [PradicBrown2021], p. 1.</TD>
 </TR>
 
 <TR>
+  <TD>limenpsi</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on sbth</TD>
+</TR>
+
+<TR>
   <TD>php</TD>
   <TD>~ phpm</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7212,6 +7212,12 @@ than reals.</TD>
 </TR>
 
 <TR>
+  <TD>fz1iso</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof depends on OrdIso</TD>
+</TR>
+
+<TR>
   <TD>seqshft</TD>
   <TD><I>none currently</I></TD>
   <TD>need to figure out how to adjust this for the differences
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7129,6 +7129,13 @@ than reals.</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


avekens (approved):

iset.mm
```diff
@@ -63759,6 +63914,13 @@ would need to be strengthened (for example, to imply included middle) to
     ( wfun cfn wcel cdm fundmfi ancoms wfn funfn fnfi sylanb impbida ) ABZACDZA
     EZCDZNMPAFGMAOHPNAIOAJKL $.
 
+  $( The restriction of a function by a finite subset of its domain is finite.
```
benjub, line 63917: by --> to

By the way: are these theorems about the "size of a function" really useful ? The "size of a function" is not really a mathematical concept (and in our encoding, it is simply the size of its domain, isn't it?).

jkingdon: > By the way: are these theorems about the "size of a function" really useful ? The "size of a function" is not really a mathematical concept (and in our encoding, it is simply the size of its domain, isn't it?).

You've commented on this before, and I don't really have an opinion. I'm just intuitionizing what is in set.mm.

benjub: You're also propagating the above mistake: by --> to.

I understand it's a hassle. Not sure how to better integrate set.mm and iset.mm and avoid double maintenance costs... 

Well, at least for the propositional calculus part, I have an idea using T. Thacher Robinson's axiomatization. Maybe I can propose a proof of concept soon.

avekens: > > By the way: are these theorems about the "size of a function" really useful ? The "size of a function" is not really a mathematical concept (and in our encoding, it is simply the size of its domain, isn't it?).
> 
> You've commented on this before, and I don't really have an opinion. I'm just intuitionizing what is in set.mm.

Most of these theorems are contributed by me, and I need(ed) them for graph theory: The general definition of walks (and other concepts based on it, like trails and paths) uses functions to represent the vertices and edges of a walk. The size of such a function represents the length of the walk, therefore it is used very often. To replace this by looking at the domains of these functions would enlarge the proofs and make them less intuitive.

jkingdon: > Not sure how to better integrate set.mm and iset.mm and avoid double maintenance costs...

If this is the paramount concern, the obvious approach would be similar to how set.mm handles the axiom of choice. You'd develop everything as in iset.mm and then add in `EXMID` and/or `ax-3` only where needed.

However, I probably agree with Norm (I could in theory dig up the old mailing list post, although if memory serves it didn't go into much detail), that the natural order/way in which you treat different topics is different. And it would be too much conceptual overhead in set.mm to add a lot of stuff about decidability conditions, apartness, weak linearity, etc.

If this was done with `ax-3` and organized as set.mm now is - with one big file - it could also make it a bit hard to keep track of where ax-3 is being used (I realize there are tools to show axiom usage, but at least for me, while useful, they don't provide something quite like being able to say "all of iset.mm doesn't use ax-3".

At least for me, the current organization isn't especially onerous. Perhaps that's just because I'm used to it, though.

> Well, at least for the propositional calculus part, I have an idea using T. Thacher Robinson's axiomatization. Maybe I can propose a proof of concept soon.

Sounds like cool work. And I should probably wait to see what the proof of concept looks like before I form a lot of conclusions.

benjub: Halving maintenance costs is indeed a concern. The FOL part of set.mm continues to undergo modifications at a fast pace, and even those modifications that could be transferred to iset.mm are not always, in spite of your diligence.

You (@jkingdon) summarized indeed a few key concerns above. My idea is to have set.mm and iset.mm share some core and diverge at some point. As currently written, the core could be all assertions relying on ax-mp, ax-1, ax-2 only, that is:
* core.mm: contains what corresponds roughly to the current first ~110 assertions of set.mm (already ~10 of them missing from iset.mm, and comments are probably not in sync either)
* set.mm: begins with `$[ core.mm $]`
* iset.mm: begins with `$[ core.mm $]`

A bit more ambitious would be to have all intuitionistic propositional calculus as core (~1360 intuitionistic assertions), and then diverge. Etc.

As you know, what should be added to iset.mm to obtain set.mm is not `ax-3` since it would create a (complete but) redundant axiomatization, but `peirce`. See https://arxiv.org/abs/2109.14745v3 and I can also send you Robinson's article if it's behind a paywall.

Proof of concept: I could work on it in August (as an mm file on my personal repo, maybe with generated webpages hosted somewhere).

benjub: > Most of these theorems are contributed by me, and I need(ed) them for graph theory: The general definition of walks (and other concepts based on it, like trails and paths) uses functions to represent the vertices and edges of a walk. The size of such a function represents the length of the walk, therefore it is used very often. To replace this by looking at the domains of these functions would enlarge the proofs and make them less intuitive.

Thanks @avekens for this remark. This gives a satisfactory motivation for these theorems. It may be a good idea (in the long run) to define a function `WalkLength : Walks --> NN0` ? I know that in a way it may sound contradictory to the recent discussion on "(/) as set versus (/) as ordinal". But to me, this is not exactly the same, since ordinals are very classically and standardly defined as sets, whereas walks may have different encodings. I haven't thought much about it. Just an idea like that.

jkingdon: > even those modifications that could be transferred to iset.mm are not always, in spite of your diligence

Diligence is perhaps overstating my intentions a bit - I don't anticipate they will ever be fully in sync, nor is that particularly my goal. I usually copy something over when I find myself missing it.

> * core.mm: contains what corresponds roughly to the current first ~110 assertions of set.m

Yeah, this seems like a low hanging fruit.

Things get hairy pretty quickly after that, but that's not a reason to avoid doing it for these first 110 or so theorems.

PR #2708, opened 2022-07-24
new theorems fvoveq1/fvoveq1d created from @savak's list

avekens:
After ~fvoveq1d was available, 160 proofs were shortened by 2358 bytes.
After ~fvoveq1 was additionally available (this corresponds to no. 23 in @savask `s list), additionally 342 proofs were shortened by 4666 bytes (curiously, now we have 168 proofs using  ~fvoveq1d!).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1482,11 +1482,9 @@
 "ax13lem1" is used by "nfeqf2OLD".
 "ax13lem1" is used by "wl-19.2reqv".
 "ax13lem1" is used by "wl-19.8eqv".
-"ax13lem1" is used by "wl-dveeq12".
 "ax13lem2" is used by "nfeqf2".
 "ax13lem2" is used by "nfeqf2OLD".
 "ax13lem2" is used by "wl-19.2reqv".
-"ax13lem2" is used by "wl-dveeq12".
 "ax13lem2" is used by "wl-speqv".
 "ax5el" is used by "dveel2ALT".
 "ax5eq" is used by "dveeq1-o16".
@@ -13872,8 +13870,8 @@ New usage of "ax12vALT" is discouraged (0 uses).
 New usage of "ax13ALT" is discouraged (0 uses).
 New usage of "ax13dgen4OLD" is discouraged (0 uses).
 New usage of "ax13fromc9" is discouraged (0 uses).
-New usage of "ax13lem1" is discouraged (8 uses).
-New usage of "ax13lem2" is discouraged (5 uses).
+New usage of "ax13lem1" is discouraged (7 uses).
+New usage of "ax13lem2" is discouraged (4 uses).
 New usage of "ax1cn" is discouraged (0 uses).
 New usage of "ax1ne0" is discouraged (0 uses).
 New usage of "ax1rid" is discouraged (0 uses).
@@ -15143,8 +15141,6 @@ New usage of "dfnfc2OLD" is discouraged (0 uses).
 New usage of "dfpjop" is discouraged (4 uses).
 New usage of "dfpleOLD" is discouraged (2 uses).
 New usage of "dfsn2ALT" is discouraged (0 uses).
-New usage of "dfss1OLD" is discouraged (0 uses).
-New usage of "dfss5OLD" is discouraged (0 uses).
 New usage of "dftru2" is discouraged (0 uses).
 New usage of "dfvd1imp" is discouraged (1 uses).
 New usage of "dfvd1impr" is discouraged (1 uses).
@@ -15201,7 +15197,6 @@ New usage of "dicelvalN" is discouraged (1 uses).
 New usage of "dicfnN" is discouraged (1 uses).
 New usage of "dicvalrelN" is discouraged (0 uses).
 New usage of "difidALT" is discouraged (0 uses).
-New usage of "diftpsn3OLD" is discouraged (0 uses).
 New usage of "dih0bN" is discouraged (0 uses).
 New usage of "dih0vbN" is discouraged (0 uses).
 New usage of "dih2dimbALTN" is discouraged (0 uses).
@@ -15260,7 +15255,6 @@ New usage of "dipfval" is discouraged (3 uses).
 New usage of "diporthcom" is discouraged (1 uses).
 New usage of "dipsubdi" is discouraged (2 uses).
 New usage of "dipsubdir" is discouraged (2 uses).
-New usage of "disjpr2OLD" is discouraged (0 uses).
 New usage of "distrlem1pr" is discouraged (1 uses).
 New usage of "distrlem4pr" is discouraged (1 uses).
 New usage of "distrlem5pr" is discouraged (1 uses).
@@ -15573,7 +15567,6 @@ New usage of "elpjhmop" is discouraged (3 uses).
 New usage of "elpjidm" is discouraged (2 uses).
 New usage of "elpjrn" is discouraged (0 uses).
 New usage of "elpqn" is discouraged (24 uses).
-New usage of "elpr2OLD" is discouraged (0 uses).
 New usage of "elprnq" is discouraged (22 uses).
 New usage of "elpwgded" is discouraged (2 uses).
 New usage of "elpwgdedVD" is discouraged (1 uses).
@@ -15611,11 +15604,9 @@ New usage of "enrex" is discouraged (9 uses).
 New usage of "eqeq1dALT" is discouraged (0 uses).
 New usage of "eqeqan12dALT" is discouraged (0 uses).
 New usage of "eqid1" is discouraged (0 uses).
-New usage of "eqoreldifOLD" is discouraged (0 uses).
 New usage of "eqrdOLD" is discouraged (0 uses).
 New usage of "eqresr" is discouraged (4 uses).
 New usage of "eqsbc3rVD" is discouraged (0 uses).
-New usage of "eqsnOLD" is discouraged (0 uses).
 New usage of "equcomi1" is discouraged (1 uses).
 New usage of "equid1" is discouraged (1 uses).
 New usage of "equid1ALT" is discouraged (0 uses).
@@ -17541,7 +17532,6 @@ New usage of "prel12gOLD" is discouraged (0 uses).
 New usage of "preleqALT" is discouraged (0 uses).
 New usage of "preleqOLD" is discouraged (2 uses).
 New usage of "preqsnOLD" is discouraged (0 uses).
-New usage of "preqsnOLDOLD" is discouraged (0 uses).
 New usage of "preqsndOLD" is discouraged (0 uses).
 New usage of "prlem934" is discouraged (3 uses).
 New usage of "prlem936" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


avekens:
Since the amount of changes is too huge to be displayed by Github, here are are the new theorems:

```
  ${
    fvoveq1d.1 $e |- ( ph -> A = B ) $.
    $( Equality deduction for nested function and operation value.
       (Contributed by AV, 23-Jul-2022.) $)
    fvoveq1d $p |- ( ph -> ( F ` ( A O C ) ) = ( F ` ( B O C ) ) ) $=
      ( co oveq1d fveq2d ) ABDFHCDFHEABCDFGIJ $.
  $}

  $( Equality theorem for nested function and operation value.  Closed form of
     ~ fvoveq1d .  (Contributed by AV, 23-Jul-2022.) $)
  fvoveq1 $p |- ( A = B -> ( F ` ( A O C ) ) = ( F ` ( B O C ) ) ) $=
    ( wceq id fvoveq1d ) ABFZABCDEIGH $.
```

The rest are modified (=shortened) proofs.

avekens:
Unfortunately, the proofs of theorems ~ovolunlem1a and ~ovolunlem1 were also modified by @benjub before, causing conflicts. So I have to merge the changes (i.e., I will take Benoît`s proof and minimize it again with my new theorems). 
We should avoid simultaneous mass minimizations in the future...

benjub:
That's unfortunate.  My PR was merged before I had any idea you were working on this one, so I'm not sure how to proceed in the future ? Giving a heads-up that one is going to do a maybe-massive minimization ? Maybe one full day in advance ?

As a side note: I would do a separate commit containing exactly the addition of the two theorems that you quoted above. And then a commit with only automatic modifications. In the present case, when seeing the merge conflict, you could have reverted only that second commit, then merged, and later done the automatic minimization (hopefully you kept the log file so as to try and minimize only those proofs you know get minimized).

avekens:
Solving the conflicts was not complicated, only the repeated minimization took a long time (because the proofs of the concerned theorems are large).

benjub (approved):
Since I cannot see the changes here (I would have to pull your branch to my computer and check the diff locally), I'm reduced to blindly approve it. After merging, I'll do a cursory check on my computer.

benjub:
> Solving the conflicts was not complicated, only the repeated minimization took a long time (because the proofs of the concerned theorems are large).

Are the problematic theorems fourierdlem* and stoweilem* ? They caused problems to me too. Maybe we can put a "proof modif discouraged" to them, waiting for a better solution, like splitting them ?

wlammen:
a diff file created locally has more than 33.000 lines.  This is impossible to check in full manually.
I investigated some samples and found that in all cases the proof size has shrunk (albeit down to 2 characters only in faclbnd4lem2).  It seems the minimize program works as expected.  I then in particular looked into the conflicting proofs, but the proofs shrunk there, too.  Finally I skimmed through the diff file to see whether something was unintentionally added or deleted apart from the two mentioned theorems (was there a sole green or red line range?), but found no sign of that.

OK. The PR is merged now, so my approval is not really needed any more.

benjub:
I merged, and then did `$ git diff HEAD^1 set.mm > diff.txt`. This seems fine.

wlammen:
Creating a local diff is not that difficult.  Proceed as follows (time needed < 5 min, excluding the check of the diff). 
```
# assume your working directory with a decent git history and github conf file is 'set.mm'
# this clone keeps your current work safe
cp -rf set.mm scratch
cd scratch
# just in case you currently work in a dedicated branch
git checkout develop
# find a commit that is in sync with one of the develop branch on github.  Going back a few days before
# is usually sufficient.  The gist is to reset to a state not affected by your current work in progress.
git log
git reset --hard <hash of commit>
# Update to the git repo state again
git fetch upstream
# the following should result in a fast forward.  This effectively clones the github repo
git merge upstream/develop
# now load the pull request to investigate. gh is a tool distributed by github
gh pr checkout <pull request id without hash>
git diff develop > diff-file
# now open the diff-file in your favourite editor.
...
# clean up
cd ..
rm -rf scratch
```

avekens:
> > Solving the conflicts was not complicated, only the repeated minimization took a long time (because the proofs of the concerned theorems are large).
> 
> Are the problematic theorems fourierdlem* and stoweilem* ? They caused problems to me too. Maybe we can put a "proof modif discouraged" to them, waiting for a better solution, like splitting them ?

No, I excluded ~fourier*, ~stow* and ~footex alread before from my minimizations (I forgot to mention this starting this PR). I minimized the confliction theoems ~ovolunlem1a and ~ovolunlem1 only, which took some time (several minutes on my computer), but not as much as the excluded theorems would need. I think it is a good idea to add "proof modification discouraged" to the theorems ~fourier*, ~stow* , or at least ~fourierdlem* and ~stoweilem* , but this should be done by @glacode, because they are in his mathbox.

glacode:
Hi,

if I understand correctly, the problem is that the minimization process doesn't terminate in a reasonable amount of time, when a proof is too long. In my opinion, the natural solution would be to add a parameter (to the minimize command) to skip long proofs.

For long proofs, modification should be encouraged, thus a (proof modification discouraged) clause would be difficult to understand (and contrary to intuition).

Anyway, in my opinion, the automatic solution (minimize /skiplongproof) should be preferred over manual modification to (which?) proofs.

Of course, the real implementation would be something like

minimize_with /shortonly nnnnnn

where nnnnnn is the max number of proof steps (proofs with more steps, will be skipped)

I'm using all my spare time to work on a metamath related project, and I would love to stay focused on it, but if we want to put together a community effort to add this switch to the minimize command, I can try to contribute.
Unfortunately, my C skills are rusty, and I don't have a working C development environment, already set up; but here

https://github.com/metamath/metamath-exe/blob/adb9bec92afeb0dbc65fee0aa7c029bcf8c4fb8f/src/metamath.c#L5222

there seems to be the relevant piece of code.

A few lines above, there is this snippet

          if (forbidMatchList[0]) { /* User provided a /FORBID list */
            /* First, we check to make sure we're not trying a statement
               in the forbidMatchList directly (traceProof() won't find
               this) */
            if (matchesList(g_Statement[k].labelName, forbidMatchList, '*', '?'))
              continue;
          }
          
that skips the minimization for a /FORBID label. We could add a similar if statement, just below, to skip the minimization if g_Statement[k] proof is longer than a given threshold.

BR Glauco

benjub:
Do @wlammen and @digama0 think this is doable/desirable in metamath.c ?

As for the proofs themselves: I guess you haven't handwritten them. Are they completely computer generated ? Or maybe by chunks ?  I (and probably others) would be interested to know a bit more.

savask:
Since I've recently looked into relevant parts of metamath-exe code, maybe I could also comment.

>A few lines above, there is this snippet ...

Here `k` is the number of the theorem which is used for minimization, i.e. in our case it will be `fvoveq1` and `fvoveq1d`. The proof length of `k` does not affect the minimization time at all.

The `minimize_with` command tries to minimize the currently loaded proof, so the typical use case is something like
```
prove fourierdlem102
minimize */allow */no_new ax-*
save new /compressed
quit
```
I assume Alexander used some sort of script (like `scripts/minimize-all`) to minimize all the relevant theorems. If one wants to skip the long theorems, one could do that by modifying the `scripts/list-theorems` script.

I would also like to mention that the main reason why metamath-exe takes so long to minimize those proofs is because the proof assistant mode can't work with compressed proofs directly and the `prove` command has to fully expand the compressed proof (for example, on my computer `prove fourierdlem102` even without minimization takes quite some time). Mmj2 can load such proofs pretty quickly, so this gives hope that minimization of large statements can also be performed rather swiftly.

wlammen:
I am not really familiar with this part of code.

A year ago or so I started documenting the code using doxygen style commenting, and a documentation can readily be generated with ./build.sh -d.  So far it covers just the infrastructure level.
The reason was to not let the knowledge get lost after NM passed away.
If possible I suggest to add documentation explaining the structure of the code, so others have an easier access.

benjub:
Can @glacode or @savask open an Issue on the other repo (https://github.com/metamath/metamath-exe) so that this idea doesn't get lost ? I'm sorry but my C skills will not help... Thanks.

glacode:
@savask , you are right. If I'm not mistaken, the statement to be minimized is

g_proveStatement 

and here 

https://github.com/metamath/metamath-exe/blob/adb9bec92afeb0dbc65fee0aa7c029bcf8c4fb8f/src/metamath.c#L5145

origCompressedLength is computed (see the statement following the one linked above)

Thus, it looks like a 

if ( origCompressedLength > THRESHOLD )
  <skip minimize>

may do the job. If that's the case, the "difficult" part should just be to add / store the new parameter(s).

As a final note, it looks like the MINIMIZE_WITH segment of code is all contained in the if statement that starts here

https://github.com/metamath/metamath-exe/blob/adb9bec92afeb0dbc65fee0aa7c029bcf8c4fb8f/src/metamath.c#L5051

BR
Glauco

glacode:
Open issue 

https://github.com/metamath/metamath-exe/issues/95

glacode:
> As for the proofs themselves: I guess you haven't handwritten them. Are they completely computer generated ? Or maybe by chunks ? I (and probably others) would be interested to know a bit more.

handwritten (mmj2 is fast...)

glacode:
The solution suggested here

https://github.com/metamath/metamath-exe/issues/95#issuecomment-1199200287

is to filter upfront, using a script that builds the list of theorems to be minimized.

Here's an example of a script snippet that lists theorems in order of proof size

https://gist.github.com/savask/e44ee6086a1321b2bbddbce027975bae#file-preparejobs-sh-L40-L44

david-a-wheeler:
Historically we would ask people to minimize any *new* proofs they added before admitting them. Tools like `scripts/minimize-all` can minimize a range of theorems.

We'd then do a mass minimization maybe once a year or so. The `scripts/jobs` script is one of the scripts we'd use to do this. There was a social announcement to pause, then everyone who had a lot of computing power available would run scripts to try to minimize things. Norm would combine those results into a newly-minimized set.mm, and after that, things went back to normal. I don't know if we should do the same process now, we have more participants & set.mm is larger now. But I thought it'd be good document "how it was done".

benjub:
@avekens : can you edit your comment at https://gist.github.com/savask/dce5b48b7317e8144635fda09e983fae to add to your list Line 23 there ?

As for the labels fvoveq1/fvoveq1d, I think the number 2 would be more appropriate (i.e., label them fvoveq2/fvoveq2d) since these are equality theorems for the **2nd** argument in the expression ``( F ` ( A O C ) )``.

Can you have a look, when you have time, at the other lines using `*veq*` in @savask's list with at most one essential hypothesis (I think there are five of them) ?

savask:
@benjub 
>can you edit your comment at https://gist.github.com/savask/dce5b48b7317e8144635fda09e983fae to add to your list Line 23 there ?

I suggest to refrain from adding anything to the gist, since now we are using issue https://github.com/metamath/set.mm/issues/2710 to track the progress.

I will modify my comment at the gist to point at this issue and will delete your and @avekens comments to clean things up, if you're okay with that.

benjub:
> I suggest to refrain from adding anything to the gist, since now we are using issue #2710 to track the progress.

You're right, I forgot about that too ! Yes, edit that gist page as you proposed.

PR #2709, opened 2022-07-24
Duplicate ge0nemnf2 / xrge0neqmnf

benjub:
Deal with Item http://us.metamath.org/mpeuni/ge0nemnf2.html versus http://us.metamath.org/mpeuni/xrge0neqmnf.html in #2128. See commit messages.

Ping @tirix and @glacode in case they want to review their versions (not a necessary condition for merging).

DO NOT MERGE before approval by @avekens (in case this overlaps the minimization he's working on). (**edit:** his PR is merged now, so approval by anyone would suffice.)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18361,6 +18361,7 @@ New usage of "wwlksext2clwwlkOLD" is discouraged (0 uses).
 New usage of "wwlksnwwlksnonOLD" is discouraged (1 uses).
 New usage of "xihopellsmN" is discouraged (0 uses).
 New usage of "xpexgALT" is discouraged (0 uses).
+New usage of "xrge0neqmnfOLD" is discouraged (0 uses).
 New usage of "xrge0tmdOLD" is discouraged (0 uses).
 New usage of "zaddablx" is discouraged (0 uses).
 New usage of "zexALT" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@savask : I would say the latest commit "detailed the TODO of tgiooss and discouraged new usage" allows us to check the corresponding box in #2128. From that point, I'd say it's up to the mathbox owner to deal with it (fourierdlem* are just too huge for me to take the time and do a `MM-PA> minimize_with` on them).

tirix (approved):
Looks good to me! Thank you Benoît!

savask:
>I would say the latest commit "detailed the TODO of tgiooss and discouraged new usage" allows us to check the corresponding box in https://github.com/metamath/set.mm/issues/2128.

I'd rather have things checked _only_ if the duplication was completely resolved, i.e. one of the theorems is not used anymore and is removed or marked as OLD. It seems that `rerest` and `tgiooss` have exactly the same formulation, so instead of using minimize_with you can just use replace-with functionality of your text editor.

benjub:
> I'd rather have things checked _only_ if the duplication was completely resolved, i.e. one of the theorems is not used anymore and is removed or marked as OLD. It seems that `rerest` and `tgiooss` have exactly the same formulation, so instead of using minimize_with you can just use replace-with functionality of your text editor.

Unfortunately, they differ by a variable renaming, so I'd have to convert to normal format, then do the change, then convert back to compressed format. It's faster than minimize_with, but with such huge proofs, even format conversion is long.  I will not check the corresponding box in your Issue if you think it should stay unchecked.

PR #2715, opened 2022-07-26
Add memorial at start of Norm's mathbox

jkingdon:
So I know we have already memorialized Norm various places such as the dedication at the bottom of the README.md (and may continue to add more).

For me, though, my first introduction to how we handle this was the memorial that Norm wrote for Alan Sare at the start of Alan Sare's mathbox. There's something about how it captures both the contributions to metamath and Alan's enthusiasm for formal proofs.

Perhaps it is too tall an order to try to capture Norm's contributions, his dedication, or his joy in what metamath has been able to do. But that hardly seems to be a reason not to try.

Perhaps someone has some suggestions about what we might put here (such as a different or additional quote, or something else I didn't think to include).


wlammen:
Nice idea  and a good start!

wlammen (approved):

avekens (approved):
Great! I have nothing to be added (at least for the moment).

benjub:
Thanks for this addition.

Minor comments: I think "Metamath" should be capitalized everywhere ? (there are a few instances of "metamath" in your text).  The set.mm style uses double space after periods.

Did you have other candidates for the quote ?

Paragraph "key insights": I would rephrase (and then you would have to rephrase my rephrasing if my grammar is incorrect) to something like:
> One of his key insights was that formal proofs can exist not only to be verified by computers, but also to be read by humans.  Both the specification of the proof format (which stores full proofs, as opposed to the proof templates used by most proof assistants) and the generated web display of Metamath proofs, one of its distinctive features, contribute to this double objective.

To me, there were two other key insights. @digama0 may wish to comment on that too.

The foundational theory behind the Metamath specification involves two key insights.  First, it is possible to develop a proof verifier and proof assistant on a very weak metalogic (the notions of free and bound variables and of proper substitution are not required by the metalogic, but can be recovered/emulated within the logic), making the system all the more trustworthy/robust. This is based on work by Tarski (...article...), but its adaptation(or better word?) to computer proofs was new and unique to Metamath.  Second, and it seems to be new, it was realized by Norm that since many theories (like FOL or ZFC) are not finitely axiomatized but have finitely many axiom schemes, it is these axiom schemes that should play the main role, and not the axioms, and all proofs should be among schemes, rather than at the object-level as traditionally done. This is the origin of the "meta" in "Metamath."  Details can be found in his article "A Finitely Axiomatized Formalization of Predicate Calculus with Equality," Notre Dame Journal of Formal Logic, 36:435-453, 1995 (preprint: https://us.metamath.org/downloads/finiteaxiom.pdf) and Appendix C of....

jkingdon:
> I think "Metamath" should be capitalized everywhere ? (there are a few instances of "metamath" in your text). The set.mm style uses double space after periods.

Both changed.

> Did you have other candidates for the quote ?

The only other thing I even thought of was https://groups.google.com/g/metamath/c/mnkBQV1_cXc/m/zz4C8_2KAAAJ but I don't see anything there that gets into larger issues of what metamath is or why it is cool, just pretty detailed discussions of certain choices. The paragraph I picked has always been one of my favorites so unless someone else has a similar favorite it strikes me as a good choice.

> Paragraph "key insights": I would rephrase

That looks like a good change, adopted. We don't need to get too deep into all the design choices of metamath, but this is a place where can be a bit more technical than in a memorial for a non-mathematical audience.

> weak metalogic . . . axiom schemes that should play the main role, and not the axioms

Elsewhere we go into the details including literature citations and more lengthy explanations but I did add the following:

```
  On a technical note, Metamath innovated both by using a very simple
  substitution rule (and then using that to build more complicated
  notions like free and bound variables) and also by taking the axiom
  schemas found in many theories and taking them to the next level - by
  making all axioms, theorems and proofs operate in terms of schemas.
```

I did want to emphasize the wider aspects of the work - such as trying to make metamath accessible to the largest feasible audience, and cultivating a community of contributors - but like I said before, this does seem to be a place where we can also include a few technical details (while avoiding the temptation to make it a treatise on everything that is distinctive about metamath or the parts of set.mm that Norm wrote).

benjub:
Thanks, that's good. I'll be away from my main computer for a fortnight, without much access to github. That paragraph you added is nice, though to me, this is not merely "technical". In retrospect, this is really what I see as the main innovation (together with the fact that this "method" suffices, which is summed up in what we can call "Megill's completeness theorem").

jkingdon:
> this is not merely "technical"

Ooh good catch. I was aiming for a different meaning of the word "technical" which was just "about math rather than say people or teaching". But I agree it didn't read that way (especially to mathematicians who love phrases like "technical details") and I've pushed up a change to just drop the words "On a technical note" because the paragraph doesn't really those introductory words anyway.

wlammen:
> I'll be away from my main computer for a fortnight, without much access to github.

@benjub Have a nice holiday!

PR #2716, opened 2022-07-26
move jarli from WL mathbox

wlammen:
1. move jarli from my mathbox to main, as requested in #2714
2. update changes-set.txt with recent moves
3. delete some outdated OLD theorems
4. shorten notnotr with jarli.  I do not add a credit here, because the shortening is a consequence of the introduction of a new theorem.

It is fairly seldom that you want a complex intermediate result be simplified.  In general short proofs function the other way round: you construct a complex final result from basic expressions.  So I expect theorems like jarl* or jarr* to be applied just on a few occasions.  One is pm2.86i.  I replaced its proof (by me anyway) to underline that pm2.86i is essentially an instance of jarri.  The proof length stays the same.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main
+26-Jul-22 wl-jarli  jarli       moved from WL's mathbox to main
+25-Jul-22 bj-sb3b   sb3b        moved from BJ's mathbox to main
 20-Jul-22 brfi1indlem hashdifsnp1
 17-Jul-22 mapsnend  [same]      moved from GS's mathbox to main set.mm
 17-Jul-22 mapsnd    [same]      moved from GS's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -15137,7 +15137,6 @@ New usage of "dfdp2OLD" is discouraged (1 uses).
 New usage of "dfhnorm2" is discouraged (3 uses).
 New usage of "dfiop2" is discouraged (3 uses).
 New usage of "dfiota4OLD" is discouraged (0 uses).
-New usage of "dfnfc2OLD" is discouraged (0 uses).
 New usage of "dfpjop" is discouraged (4 uses).
 New usage of "dfpleOLD" is discouraged (2 uses).
 New usage of "dfsn2ALT" is discouraged (0 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -1551,7 +1551,7 @@ between the hypotheses and the assertion (conclusion), abbreviating the
     $( Inference associated with ~ pm2.86 .  (Contributed by NM, 5-Aug-1993.)
        (Proof shortened by Wolf Lammen, 3-Apr-2013.) $)
```
benjub, line 1552: Could you add to the comment of pm2.86i:
> An equally short proof is possible from ~ ax-1 and ~ syl11 .

so that we do not lose this information?

set.mm
```diff
@@ -1682,6 +1682,21 @@ giving a shorter proof but depending on more axioms (namely, ~ ax-1 and
       ( wn wi pm2.18 syl ) ABDBEBCBFG $.
   $}
 
+  $( Elimination of a nested antecedent as a partial converse of ~ ja (the
+     other partial converse being ~ jarri ).  (Contributed by Wolf Lammen,
```
benjub, line 1686: Maybe we can remove the remark about partial converse, which is more appropriate in the comment of jarli ?

benjub (commented):
Just two small comments.

Thanks for the update of changes-set.txt

As for the move of jarli to Main, this was not "requested" in #2714 and I was wondering about it, but I think it's a good move.

set.mm
wlammen: Both proofs represent the sequence ax-1, syl, com12.  The old one pairs the two last, the new one the first two steps.  So which one is to be preferred? Asking me, the former looks like a technical shortcut, the latter has, from a pedagogical point of view, slightly more relevance.  I can create an OLD version of course, but is it really worth it?

set.mm
wlammen: done

set.mm
benjub: It's not worth creating an OLD or ALT version. I thought one sentence in the comment was appropriate. But if you think it's not, then as you prefer.

set.mm
benjub: I would have removed "as a partial converse of ~ ja" since the partial converse of ~ ja is actually jarli, as indicated in the comment of jarli. But if you want to keep it, it's as you prefer too.

benjub:
Again two comments, but they are non-blocking and if you think it's ok, you can merge.

avekens (approved):

set.mm
wlammen: also done

set.mm
wlammen: This opens a can of worms: Keep a myriad of variants in the theorem comment, for what reason?  To not get any information lost?  If the idea is not valuable enough to make it to an ALT version, I'd rather see it discarded, as was the case with numerous alternative proof ideas I threw away in a shortening process.  I am not really convinced.  A remedy to that I occasionally encounter, is: use the git history commands.

benjub: We do not have to be systematic in everything we do. I wouldn't have added this comment if I'd discovered this alternate proof. It is the action of removing it which made me think it could be kept in a comment at a very low cost. But that's not very important indeed.

I merged the PR anyway.

PR #2719, opened 2022-07-28
sb6 without ax-13

wlammen:
1. sb6 is provable without ax-13.  I noticed this, and replayed its proof using dv variants of some used theorems.  I then found bj-sb6 in BJ's mathbox.  He essentially did the same, although he broke the proof down using auxiliary theorems bj-sb2v and bj-sb4v.  I kept my proof, but attributed the revision to him.  He was the first to see this.  I wonder why his version did not make it to main over the years?  It is clearly an improvement.  In the wake sb5 looses its dependency on ax-13 as well, maybe more theorems.
2. base nfs1v on sb6, and shorten hbs1 using nfs1v.  In total 2 proof bytes are saved, and both theorems loose their dependency on ax-13, thanks to sb6.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15867,6 +15867,7 @@ New usage of "hbnae-o" is discouraged (3 uses).
 New usage of "hbntOLD" is discouraged (0 uses).
 New usage of "hbntal" is discouraged (3 uses).
 New usage of "hbra2VD" is discouraged (0 uses).
+New usage of "hbs1OLD" is discouraged (0 uses).
 New usage of "hcau" is discouraged (4 uses).
 New usage of "hcaucvg" is discouraged (1 uses).
 New usage of "hcauseq" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


avekens (approved):

benjub:
Thanks @wlammen.  I think the lemmas bj-sb2v/sb4v coud also be moved to Main, but I'll do it when I come back.
In that section in my mathbox ("Removing dependencies on ax-13"), there are around 30 theorems which remove dependency on ax-13, and I did plan to move them to Main. Norm was wary of a mass move, so I was planning on moving them bit by bit, but didn't get around to do it yet. I'll try to do it when I come back too. If you plan on working on it before, feel free to move them yourself.

david-a-wheeler:
@wlammen / @benjub - thanks for all this! Removing the use of axioms is generally a good idea, so I'm glad to see this. I expect Norm just wanted to "do no harm" by moving them bit by bit, instead of all at once.

PR #2721, opened 2022-07-29
shorten pair expd and expdcom

wlammen:
1. The proof steps of expd contain expdcom as a subproof.  Reorder the theorems and base expd on expdcom to save in total 5 proof bytes / 1 essential proof step.
2. I accidentally visited a theorem which I had shortened twice.  It was recently decided to let just the latest label survive in the credits section.  I am not going to clean up that issue systematically, but randomly, as I notice it in the course of other work.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15675,6 +15675,8 @@ New usage of "exlimexi" is discouraged (2 uses).
 New usage of "exlimiOLD" is discouraged (1 uses).
 New usage of "exlimihOLD" is discouraged (0 uses).
 New usage of "expcomdg" is discouraged (0 uses).
+New usage of "expdOLD" is discouraged (0 uses).
+New usage of "expdcomOLD" is discouraged (0 uses).
 New usage of "extwwlkfablem1OLD" is discouraged (0 uses).
 New usage of "f1cofveqaeqALT" is discouraged (0 uses).
 New usage of "f1oweALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -4300,10 +4300,25 @@ inference eliminates the need for a syllogism ( ~ syl ) -type inference
       ( ex com12 ) ABCABCDEF $.
   $}
 
+  ${
+    expdcom.1 $e |- ( ph -> ( ( ps /\ ch ) -> th ) ) $.
+    $( Commuted form of ~ expd .  (Contributed by Alan Sare, 18-Mar-2012.)
+       Shorten ~ expd .  (Revised by Wolf Lammen, 28-Jul-2022.) $)
```
benjub, line 4306: That line seems redundant since the mention of your shortening already appears in the comment of ~expd.

benjub (approved):

set.mm
wlammen: I added it to indicate the date and incentive of the latest proof change.  I am not in particular keen on letting my name appear there.  Something like 'Last modified on 28-Jul-2022.' would be just fine, and help anyone interested in the evolution of a theorem proof.  But that has to be agreed upon.

I did that already in the past, see imp4b.

I removed my name, so we can see how well this is received by the audience.

benjub: Ah, it was for the date, I see. Do as you see fit, and you can merge.

PR #2722, opened 2022-07-30
Add fodjuomni to iset.mm

jkingdon:
This is Lemma 3.2 from [PradicBrown2021] and in their words/notation it is:

```
Suppose that we have an omniscient set O and some sets A and B. If there exists
a surjection f : O → A + B, then either A is inhabited or it is empty.
```

It is convenient to break this into a handful of lemmas but none of them are especially difficult.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1606,7 +1606,7 @@ is double negation elimination.</TD>
 </TR>
 
 <TR>
-<TD>elimif , ifid , eqif , ifval , elif ,
+<TD>elimif , ifid , ifval , elif ,
 ifel , ifeqor , 2if2 , ifcomnan , csbif ,
 csbifgOLD</TD>
 <TD><I>none</I></TD>
@@ -1620,6 +1620,11 @@ csbifgOLD</TD>
 <TR>
   <TD>ifboth</TD>
   <TD>~ ifbothdc</TD>
+</TR>
+
+<TR>
+  <TD>eqif</TD>
+  <TD>~ eqifdc</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -65603,6 +65603,112 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
       OXGWPVOSWQWTXASWRBWSXBXC $.
   $}
 
+  ${
+    $d A z $.  $d B w z $.  $d F w z $.  $d O w z $.  $d X w z $.  $d ph w z $.
+    fodjuomnilemdc.fo $e |- ( ph -> F : O -onto-> ( A |_| B ) ) $.
+    $( Lemma for ~ fodjuomni .  Decidability of a condition we use in various
+       lemmas.  (Contributed by Jim Kingdon, 27-Jul-2022.) $)
+    fodjuomnilemdc $p |- ( ( ph /\ X e. O )
+        -> DECID E. z e. A ( F ` X ) = ( inl ` z ) ) $=
```
benjub, line 65612: In this statement and the next one, I would use ``( ( inl |` A ) ` z )`` (since it's within a `E. z e. A`, it's equivalent). The reason is that  ``( inl |` A )`` is more tame than the raw `inl`, especially when `A` is a set. Up to you to see if it's worth it.

iset.mm
```diff
@@ -65603,6 +65603,112 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
...
+          HVJZVQKTLWOVQOPZGQZRSZOIUAAVQVDVPWSMOIVNWQSVOWRRVNWQGVEUJVFVGVHVIAVSW
+          BAVSVBCDMEFGHIAWMVSJTAWPVSKTLAVSVDVKVIVLVM $.
+      $}
+    $}
+
+    ${
+      $d A a b x y z $.  $d B z $.  $d F a b y z $.  $d O a y z $.
+      $d ph y z $.
+      $( A condition which ensures ` A ` is either inhabited or empty.  Lemma
+         3.2 of [PradicBrown2021], p. 4.  (Contributed by Jim Kingdon,
```
benjub, line 65701: Can you phrase that condition ? E.g.: If..., then A is either inhabited of empty. (**edit**: it's basically what you wrote in your first comment above on the github PR page; I think it has its place as a comment in iset.mm)

iset.mm
jkingdon: I'm not sure whether to (again) try to describe my thoughts on this, or whether to just take "up to you" literally and say "I choose not to put in those restrictions".

iset.mm
jkingdon: I didn't write that, it is a quote from the paper.

My philosophy is that the formal statement of the theorem should (most of the time) be how we state what is being proved. Especially when proposed wording doesn't end up being any more concise or otherwise different from the formal notation.

iset.mm
benjub: I thought you were still ambivalent, but if that's settled, then ok. (As you noticed, I'm often bold in my comments in order to avoid ambiguity, and apart from the occasional obvious jokes, you are encouraged to take my comments literally.)

iset.mm
benjub: If you wrote the statement in plain English above (the quotation from the paper), it is because it helps.  So we both think it is helpful. Putting it in the comment in iset.mm would therefore be helpful.

benjub (approved):

PR #2723, opened 2022-07-30
restructure and/or section

wlammen:
The section where OR and AND is introduced always struck me as untidy and difficult to navigate.  Similar theorems lie scattered around.  This pull request attempts to improve on this.
AND appears to be the more important of both operations wrt building basic blocks, so its theorems are introduced first.  This is contrary to the current order, starting with OR.  So the first part is AND-only, followed by an OR-only section.  The first mixed-op theorems focus on completing the OR part, adding orim12d, that would otherwise be too separated from orim12i.  This is followed by higher level AND/OR-only theorems, which appear to me less fundamental, and more on the application side.  Then the connection between both operations is developed (oran etc.).  At the end are the introduction of NOR, XOR related theorems, and assorted results.
Let me know what you think.

This is a massive PR.  If we agree on how this section should be restructured, it is perhaps better to break this PR up into multiple smaller ones.


benjub:
* I agree with the general restructuring.
* Conjunction should indeed come before disjunction. Another reason is that intuitionistically, conjunction behaves the same as classically, but disjunction is a "harder" connective (the intuitionistic disjunction is strictly stronger than the classical one).
* I would even move statements as follows: wa, df-an, [and-only theorems], wo, df-or, [or-only theorems], etc.
* As you saw, some theorems still have to be mixed up (there will remain some and-only and some or-only theorems which use some and-or theorems so need to be placed after them) (*). A way to have a neat structuring and organization of the results of propositional calculus is to use an axiomatization with the separation property, like the one I advocated in https://github.com/metamath/set.mm/issues/756#issuecomment-877045990, where I had already mentioned the messy section 1.2.6 "Conjunction and disjunction".
* Let's wait for other opinions, but if others agree, then it's true that smaller PRs would help.

(*): of course, since conjunction and disjunction are introduced by definitions and definitions are eliminable, one could theoretically prove and-ony and or-only theorems successively, but that would be at the cost of longer proofs.

jkingdon:
Seems logical to me.

david-a-wheeler:
> Seems logical to me.

*rim shot sound*

I'm coming at this late, but I agree, this restructuring makes sense. Thanks!

PR #2724, opened 2022-07-31
restructure AND: importation/exportation

wlammen:
As discussed in #2723: Break a huge PR up into smaller parts. Step 1:
Importation and Exportation lies at the heart of AND.  An early introduction makes these features available to all other theorems.
1. Put all imp/exp theorems in one place at the beginning.
2. Separate the definition df-an and df-or from each other to underline that they can be developed individually to a great extent.

If this PR gets accepted, more work in this direction will follow.
Follow ups:
- develop ancom, simple anass early, so that they are generally available, and variants differing in that respect can be kept together.
- develop implication/conjunction relations
- develop elementary adant*, simp* theorems
- develop elementary biconditional theorems, so that theorems differing in biconditional/implication can be kept together.
- develop jca based theorems
- develop iba based theorems
- develop ancl based theorems
- develop basic sylan based theorems
- develop anim* based theorems
- fully develop adant*, simp*, most of sylan theorems
- develop mpan based theorems
- pm4.71/pm5.32 and friends
- develop anbi* theorems
- fully develop associative laws
- develop absorption rules
- assorted results with little strategic value
- swap and/or sections


set.mm
```diff
@@ -3860,12 +3860,9 @@ Here we define disjunction (logical 'or') ` \/ ` ( ~ df-or ) and conjunction
 
   $( Declare connectives for disjunction ('or') and conjunction ('and'). $)
   $c \/ $. $( Vee (read:  'or') $)
-  $c /\ $. $( Wedge (read:  'and') $)
```
benjub, line 3863: I thought we agreed that conjunction was to be introduced before disjunction ?

benjub (approved):
See one comment above. You can merge this PR.

I was thinking of a series of successive PRs such that after the n^th PR, the theorems are in the final order up to the a(n)^th statement, with `a` an increasing sequence.

I think there is one level of structuring left available (the one with -.-.-.-), so you may introduce three of these: 1. Conjunction, 2. Disjunction, 3. Mixed theorems (in that third section, there will also be some and-only and or-only theorems whose proofs rely on mixed theorems).

set.mm
```diff
@@ -3860,12 +3860,9 @@ Here we define disjunction (logical 'or') ` \/ ` ( ~ df-or ) and conjunction
 
   $( Declare connectives for disjunction ('or') and conjunction ('and'). $)
```
benjub, line 3861: you can remove "and conjunction..."

set.mm
wlammen: I aim at being less disruptive, so people still find most of the stuff most of the time.  In the end disjunction will follow conjunction.  In private I have a version where (almost) all steps are done.  But doing all in one step is difficult to review.  This PR aims at contracting scattered importation/exportation theorems into a single location, so you can easily browse your options using the 'nearby theorems' overview.  Nothing more.  No proofs are changed, just the order of theorems.

set.mm
wlammen: done

wlammen:
> I was thinking of a series of successive PRs such that after the n^th PR, the theorems are in the final order up to the a(n)^th statement, with a an increasing sequence.

Its about restructuring, starting with the most basic level, that is still a complete topic (importation/exportation).  Following PRs will successively build more and more sophisticated layers.

> I think there is one level of structuring left available (the one with -.-.-.-), so you may introduce three of these: 1. Conjunction, 2. Disjunction, 3. Mixed theorems (in that third section, there will also be some and-only and or-only theorems whose proofs rely on mixed theorems).

Lets wait with the finish until the task is finished.

Issue #2118, opened 2021-07-31
Replacing "10" by "; 1 0"


avekens:
In PR #2117, @nmegill raised the following question: 
> do we really need "10" at all, or should it be considered an obsolete relic of pre-Mario days? Right now, "10" is used by only 45 $p statements, and 6p4e10/6p4e10b etc. are essentially duplicated theorems. df-dec would become slightly more complex with "( 9 + 1 )" instead of "10", but things like 10p10e20 "( 10 + 10 ) = ; 2 0" seem like an awkward mixture of two notations. In a few cases "10" makes a theorem slightly nicer to read, but two notations for it is one more thing to confuse newcomers.

@digama0 replied:
> One issue with eliminating "10" is that it is used in the definition of ; A B; you wouldn't be able to use ; 1 0 there because it would be circular (and it would also be true in any base, so you wouldn't know what base is being described. As the joke goes, every base is base 10.)
> EDIT: Using 9 + 1 there directly would probably be fine. There are a bunch of special cases in the numeric normalizer to deal with 10, so I would not be against eliminating it.

I agree with Norm and I think it would be more consistent to remove the definition of "10" (df-10) and to rigorously use "; 1 0" instead. In addition to replacing "10" in the 45 $p statements Norm mentioned, a lot of work would be necessary to revise the proofs which are using these statements. Even the one theorem ~1lt10 is used in 42 proofs.

The following tasks have to be done to replace "10" by "; 1 0" (number of concerned theorems based on the results of @nmegill , see below):

- [x] Replace df-dec (renaming it dfdecOLD) by dfdec2
- [x] Replace 10 by ; 1 0 in all theorems using dfdecOLD (about 50), being replaced by dfdec10 or directly by (the new) df-dec
- [x] Mark df-10 (renaming it df-10OLD) as deprecated
- [x] Replace 10 by ; 1 0 in all theorems (about 90), marking the former versions as deprecated
- [x] Replace all occurrences of deprecated theorems by the corresponding not deprecated theorem (in about 70 theorems). Maybe some proofs can be additionally improved.
- [x] Delete dfdecOLD, df-10OLD and all related deprecated theorems

nmegill:
Usually a quick way to get an approximate usage for a symbol is to use the syntax constructor:

```
MM> show statement c10
36630 c10 $a class 10 $.
MM> show usage c10
Statement "c10" is directly referenced in the proofs of 140 statements:
[...]

PR #2726, opened 2022-08-02
restructure AND: rel. to bicond & impl, basic adant*, simpl*

wlammen:
1. The relation between AND and implication is expressed by variations of df-an.  These theorems are not used in the development of AND-only.  In fact the only applications of df-an occur in ex and imp, otherwise AND-only is fully self-contained.  So there is some freedom where to put these theorems.  Since they are variants of df-an, I left them in its vicinity.  If one later wants to prop up these theorems (using the commutative law e.g.), moving to a different location is an option.
2. Shorten imp5a
3. Rearrange the import/export section.  Theorems that indicate from hypotheses and conclusion, that both importation/exportation is needed are moved more to the end.  Proofs before may use both imp/ex as long as the effects annihilate, though.
4. The relation between bi-conditional and AND: start with a rewrite of df-bi and give basic support.  It is intended that from here on theorems differing just in the operation type (->, <->) in many instances can be kept together, because there is a common infrastructure.
5. The first layer of simp* and adant* theorems.  Higher levels use sylan* support that needs to be developed in parallel.
6. Delete some outdated OLD theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -151,7 +151,6 @@
 "10nn0OLD" is used by "tgoldbachOLD".
 "10nnOLD" is used by "10nn0OLD".
 "10nnOLD" is used by "1t10e1p1e11OLD".
-"10nnOLD" is used by "3decOLD".
 "10nnOLD" is used by "3dvdsOLD".
 "10nnOLD" is used by "9t11e99OLD".
 "10nnOLD" is used by "bgoldbachltOLD".
@@ -164,7 +163,6 @@
 "10nnOLD" is used by "otpsstrOLD".
 "10nnOLD" is used by "pleidOLD".
 "10nnOLD" is used by "plendxOLD".
-"10nnOLD" is used by "sq10OLD".
 "10nnOLD" is used by "tgblthelfgottOLD".
 "10nnOLD" is used by "tgoldbachltOLD".
 "10posOLD" is used by "0.999...OLD".
@@ -303,7 +301,6 @@
 "2uasbanh" is used by "2uasban".
 "3atnelvolN" is used by "2atnelvolN".
 "3atnelvolN" is used by "islvol2aN".
-"3decOLD" is used by "3dvds2decOLD".
 "3lcm2e6woprm" is used by "lcmf2a3a4e12".
 "3lt10OLD" is used by "2lt10OLD".
 "3noncolr1N" is used by "lplnribN".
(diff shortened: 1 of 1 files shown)


benjub:
I think it's ok. I haven't checked carefully, though.
If proof shortenings and OLD deletions were in a separate commit, that would help (if one commit is only moving theorems around, then the indication on the github page "629 lines added, 629 lines deleted", like in the previous PR, is already reassuring).

**edit**: so, to be clear, either someone (@jkingdon ? @avekens ?) has time to review, or else @wlammen is sure that everything is ok and he can merge.  Sorry but I don't have much time this week (hiking takes time!).

wlammen:
OK, next time there will be smaller commits.
ad 1. This is about iman and imnan.  Both are not needed in the AND-only section (actually this section would be provable from imp and ex alone, if they replaced df-an).  The question is now: where shall we put these theorems?  For one they are on the application side (seen from AND), so they should appear late.  On the other hand, where does a less initiated user look for them? I decided to have them right at the beginning.  For me they are an answer to how AND is embedded in implication.
ad 2. compare imp5a and imp5aOLD for verification. The number of proof bytes decreased from 19 to 17.
ad 3. This is fine tuning.  How can I present the whole set of imp exp theorems, so that an infrequent visitor somehow knows where to look?  I decided to have plain importation and exportation, where a chain of implications is fully or partially transformed to/from ANDed antecedents.  In addition, these plain transformations have traditionally been accompanied by theorems where, an imported version is transformed to a different imported one.  These theorems are more on the application level, and have less principle value.  So let them appear after the pure ones.
ad 4.  dfbi2 is a version of df-bi, that uses AND for a less complex expression.  I tried to move it to an early position, so theorems employing <-> have easy access to them.  The bi-conditional is actually not a necessary concept to develop AND (apart from df-an, of course).  I see theorems using <-> as convenience extensions.  And it is convenient to have them adjacent to corresponding -> like theorems.
ad 5.  simp* and adantr* are wide-spread in short proofs.  The upcoming theorems are no exceptions.  So introduce them as early as possible.
ad 6. All deleted OLD theorems are located far from the AND section.  They appear later in the database.  For verification read the comments attached to the OLD theorems.
>  wlammen is sure that everything is ok 

Well, I am sure I can come up with a fairly decent solution.  But some decisions are a compromise.  A trade-off between easy random usage using 'nearby theorems' and a conceptional thorough idea, that pleases from a theoretical point of view.  And then I must remind you that I have no education in logic or math, so some of my ideas may not stand up to expert review.

avekens (approved):
Unfortunately, I did not fully get a feeling about the restructurings. I will have a look at them after this PR is merged and the theorems are shown on the web page.

As for the deletion of the OLD theorem: these are mostly about issue #2118, which I planned for September. It does not matter, however, that some of these OLD proofs are removed earlier.

benjub:
Thanks Wolf. I'd say you can merge and follow your judgement for the points
you raised.

Le mar. 2 août 2022 à 20:24, Wolf Lammen ***@***.***> a
écrit :

wlammen:
Thanks for your support.

PR #2727, opened 2022-08-03
restructure AND: jca & iba

wlammen:
Restructure AND: jca centred contributions.

1. collect a couple of convenience theorems around importation, that we need
2. jca (and similar) is a generalization of pm3.2.  Put it closest to this theorem
3. iba is a convenient result based on jca and simp* that is used frequently.  Let it come next.
4. anc* theorems, a loose form of iba
5. Update comments to df-an and df-or.

Many thanks to Gérard Lang who pointed out that df-an, being the first definition now, needs an update of its comment.

The individual commits are better suited for review than the 'Files changed' view


benjub (approved):

benjub (approved):

benjub (approved):

benjub (approved):

benjub (approved):

benjub (approved):

benjub (approved):

benjub:
Well, I wanted to approve commits independently, but this appears as several approvals of the whole PR !  Anyway: ready to merge (and much easier to review indeed).

avekens (changes requested):
As the definitions are not at the beginning of the section, I would propose to split the section and provide two (3rd level) headers:
* Logical conjunction
* Logical disjunction

Or to keep the current 3rd level header, and add two 4th level headers for "and" and "or".

PR #2728, opened 2022-08-03
restructured AND: imdistan* & sylan* & anim*

wlammen:
New to restructured AND:
1. add imdistan* theorems
2. add first layer of sylan* theorems.  Higher level ones need more to be proved efficiently.
3. add anim* like theorems

I recommend to look at the individual commits for review instead of the 'Files changed' view.


benjub (approved):

avekens (changes requested):
Since I was too late for the last PR, I repeat my remarks here again:

As the definitions are not at the beginning of the section anymore, I would propose to split the section and provide two (3rd level) headers:

* Logical conjunction
* Logical disjunction

Or to keep the current 3rd level header, and add two 4th level headers for "and" and "or".

wlammen:
Thanks for your suggestion.  I think there will be at least a third section with mixed operations.  As far as I am concerned, I like to leave it for now as is, until things have matured enough to see what the titles are, and where to draw the lines.  It is on my TODO list.   But of course feel free to implement your own suggestions.  I like to see this as a collaboration.

benjub:
Indeed, I proposed three fourth-level sections at https://github.com/metamath/set.mm/pull/2724#pullrequestreview-1057396698, but Wolf prefers to defer this to the end of the restructuring, and I'm fine with that.

avekens:
> Indeed, I proposed three fourth-level sections at [#2724 (review)](https://github.com/metamath/set.mm/pull/2724#pullrequestreview-1057396698), but Wolf prefers to defer this to the end of the restructuring, and I'm fine with that.

OK, then this should be done at the end...

avekens (approved):

PR #2732, opened 2022-08-04
restructure AND: anbi & anabsan & anass*

wlammen:
restructure AND:
1. add anbi*
2. add anabsan*
3. complete associative laws
4. add anandi*

By and large, here we are.  This is the infrastructure presented by a sort of stand-alone AND to other parts of logic and even a couple of AND based theorems remaining.  Let me know where you see headroom for improvement.


avekens (approved):
OK, so it's time now to adjust the section headers (as discussed before)?

wlammen:
You still need some work on OR-only and mixed AND / OR as well, to find the right place for all headers.  But of course you can prepare the headers in advance and leave the sections (but this one) empty for now.  If this ends this discussion I will go this way. Next PR.

PR #2733, opened 2022-08-05
headers; more sylan theorems in the AND only section

wlammen:


wlammen:
Introduces headers.  Check the comment to the introduction of conjunction for correctness.
syl6an and variants are added to the second sylan layer.
It seems having syllogism at the end of the AND only section is more promising.  Part of this PR reverts commit [876d6b0](https://github.com/metamath/set.mm/pull/2732/commits/876d6b020c5382aefff5411bd33b73aa94764d0d) again.  Unfortunately this messes up the diff view a bit.  Therefore I won't add more to this PR.

avekens (approved):

PR #2734, opened 2022-08-05
Graph theory - walks as word (17)

avekens:
General (main set.mm):
* new theorems ~rabrabi, ~dfss7, ~f1oresf1orab added

Graph theory - walks as word:
* ~wlkpwwlkf1ouspgr renamed ~wlkswwlksf1o
* ~wlkisowwlkupgr replaced by ~wlkswwlksen
* theorems ~wlknwwlksnfun, ~wlknwwlksninj, ~wlknwwlksnsur, ~wlknwwlksnbij2 obsolete; ~wlknwwlksnbij, ~wlknwwlksnen revised
* theorems ~wlkwwlkfun, ~wlkwwlkinj, ~wlkwwlksur, ~wlkwwlkbij, ~wlkwwlkbij2 obsolete
* ~wlksnwwlknvbij revised
* proof of ~rusgrnumwlkg shortened
* ~numclwlk3lem3 renamed numclwwlk3lem1 and moved
* comments revised, using term "double loop"
* some labels revised
* ~numclwwlk1lem2 revised

AV's mathbox:
* new theoems ~f1oresf1o, ~f1oresf1o1 added

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13314,6 +13314,16 @@
 "wl-syl5" is used by "wl-pm2.04".
 "wl-syl6" is used by "wl-ax3".
 "wl-syl6" is used by "wl-pm2.27".
+"wlknwwlksnbijOLD" is used by "wlksnwwlknvbijOLD".
+"wlknwwlksnfunOLD" is used by "wlknwwlksninjOLD".
+"wlknwwlksnfunOLD" is used by "wlknwwlksnsurOLD".
+"wlknwwlksninjOLD" is used by "wlknwwlksnbijOLD".
+"wlknwwlksnsurOLD" is used by "wlknwwlksnbijOLD".
+"wlkwwlkbijOLD" is used by "wlkwwlkbij2OLD".
+"wlkwwlkfunOLD" is used by "wlkwwlkinjOLD".
+"wlkwwlkfunOLD" is used by "wlkwwlksurOLD".
+"wlkwwlkinjOLD" is used by "wlkwwlkbijOLD".
+"wlkwwlksurOLD" is used by "wlkwwlkbijOLD".
 "wnfOLD" is used by "19.21OLD".
 "wnfOLD" is used by "19.21t-1OLD".
 "wnfOLD" is used by "19.21tOLD".
@@ -17097,6 +17107,7 @@ New usage of "nsnlpligALT" is discouraged (0 uses).
 New usage of "numclwlk2lem2f1oOLD" is discouraged (1 uses).
 New usage of "numclwlk2lem2fOLD" is discouraged (1 uses).
 New usage of "numclwlk2lem2fvOLD" is discouraged (1 uses).
+New usage of "numclwwlk1lem2OLD" is discouraged (0 uses).
 New usage of "numclwwlk2OLD" is discouraged (0 uses).
 New usage of "numclwwlk2lem1OLD" is discouraged (1 uses).
 New usage of "numclwwlk2lem1lemOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
The three theorems you're requesting to add to the main part have peculiar statements and the first is used only once and the other two are not used at all (except the second for the third). Is there a particular reason you want to add them ?

avekens:
> The three theorems you're requesting to add to the main part have peculiar statements and the first is used only once and the other two are not used at all (except the second for the third). Is there a particular reason you want to add them ?

The first one, which is used in a proof, minimized this proof (of a theorem in main), so I added it. ~f1oresf1orab, I guess, was "minimized away". I can move this theorem and dfss7 to my mathbox, but I find them intersting by themselves.

avekens:
OK, ~dfss7 and ~f1oresf1orab are in my mathbox now.

benjub:
@avekens : thanks, I think it's better for the moment. As for the first one, since it is used, you can leave it in Main, though I think that if in a proof you have an expression containing a substitution instance of `{ ? e. { ? |` or  `{ ? e. { ? .e ? |` then there is probably a shorter or more direct proof.

I'm going to review this PR today.

benjub (approved):

avekens:
With my latest commit, the proof of clwwlknonclwlknonf1o is shortened using rabrabi (to be honest, this was possible only by adapting the lemma ~clwwlknonclwlknonf1olem).

set.mm
```diff
@@ -417031,45 +417031,46 @@ to v(n-2)=v.  This theorem holds even for k=0, but not for n=2, since
...
-      GRUSFVAUCUTTPPGRZMZAVBSVCJUSVIVAAVBVIVAUKUSUTVBLMVAVHULUMUNUOHVCFVEUPUQKU
-      R $.
+      ( wcel wa c1st cfv chash wceq c2nd cc0 csubstr cuspgr cn cclwlks crab cop
+      cv cmpt cres wss ssrab3 a1i weq fveq2 fveq2d eqeq1d fveq1d anbi12d elrab2
+      co simpl simplbiim adantl ssrabdv resmptd opeq2d oveq12d cbvmptv syl6reqr
+      syl6eq ) CUALDUBLMZAHUFZNOZPOZDQZHCUCOZUDZAUFZROZSVQNOZPOZUEZTUSZUGFUHZHF
+      VKROZSVMUEZTUSZUGZBVJWCAFWBUGWGVJAVPFWBVJVNHVOFFVOUIVJVTDQZSVROZGQZMZAVOF
+      JUJUKVKFLZVNVJWLVKVOLVNSWDOZGQZMZVNWKWOAVKVOFAHULZWHVNWJWNWPVTVMDWPVSVLPV
+      QVKNUMUNZUOWPWIWMGWPSVRWDVQVKRUMZUPUOUQJURVNWNUTVAVBVCVDAHFWBWFWPVRWDWAWE
+      TWRWPVTVMSWQVEVFVGVIKVH $.
```
benjub, line 417047: Why did you lengthen the proof ?

avekens: as explained above: to shorten the proof of clwwlknonclwlknonf1o

benjub: But the statement looks unchanged ?

avekens: The variables changed ( w <-> c )! And this requires some conversion steps (~cbvmptv etc.).

benjub: Ah, I see... This does not appear in the diff, which only shows the $p-statement, so I thought it was simple alpha-renaming, but there are $e-hypotheses (which do not appear in the diff) which contain `c` and `w` variables and are unchanged, so this is not simple alpha-renaming.

benjub (approved):
Approved. The occurrence of nested restricted class abstractions (which are difficult to parse for a human) might be a sign that the proofs can be shortened/streamlined.  But I wouldn't have the courage of trying some "proof-engineering", so I wouldn't blame anyone for that !

PR #2735, opened 2022-08-06
move a couple of AND based theorems to the end of AND only

wlammen:
A couple of theorems are AND based, but are not used for the basic infrastructure.  Create an 'application' level at the end of the AND only section.


avekens (changes requested):
~annotanannotOLD should be moved together with ~annotanannot. For me, however, it would be fine to delete ~annotanannotOLD already now.

wlammen:
> ~annotanannotOLD should be moved together with ~annotanannot.

This is not possible.  Its proof uses OR.

> it would be fine to delete ~annotanannotOLD already now.

This opens up the question, whether a 1 year grace period for OLD proofs is an overkill.  Who really looks at these after, say a few weeks, any more?  A reason for keeping them so long could be to detect changes that do not pay in the long run.  In such a case a permanent reminder might be of some value.

Opinions?

avekens:
> > ~annotanannotOLD should be moved together with ~annotanannot.
> 
> This is not possible. Its proof uses OR.

Ah, I see - then leave it as it is, or delete it.

 
> > it would be fine to delete ~annotanannotOLD already now.
> 
> This opens up the question, whether a 1 year grace period for OLD proofs is an overkill. Who really looks at these after, say a few weeks, any more? A reason for keeping them so long could be to detect changes that do not pay in the long run. In such a case a permanent reminder might be of some value.
> 
> Opinions?

So you want to distinguish between long term OLDs (having a special remark in their comments, but deleted after 1 year) and short term OLDs (no special remark, as today, deleted after, say, 3 months)?

I think 3 months should be the minimum, because some people are not looking at the changes weekly (as we do...).

avekens (approved):

wlammen:
@avekens Oh, I just interpreted your comment. Because discarding ~annotanannotOLD, created on 1-Apr-2022, was fine, the grace period should be shorter in your opinion, I concluded.  There are obvious downsides to OLD theorems:  they litter the database, they distract from concepts otherwise presented concisely.  For the purpose of discussion, I then tried to figure out a reason to keep them for a year, and came up with the 'probation' idea.  In the end, I do not know, why a year is picked, and cannot reason in particular for or against it.  So lets hear what others say.  That is the idea behind the 'Opinions?'

benjub:
I think it's good to have a default period of one year for *OLD theorems. Also, when some proof modifications are clearly minimal and one is sure that no information is lost from the old proof, then it may not be necessary to keep an *OLD version. So: no clear rule here, but contributors can use their judgment (and when in doubt, keep an *OLD version).

sctfn:
Long term (when we have the webpage dynamically generated), could we work on a version that lets us go through the git history on the web? That’d get rid of the whole need for OLD versions, as well as removing the artificial one year boundary.

PR #2742, opened 2022-08-08
Use cbvalvw/cbvexvw when sufficient

benjub:
This follows #2741. This removes dependencies on ax-11,12,13 from a few theorems, like nfcjust, zfpow.


set.mm
```diff
@@ -549389,12 +549389,12 @@ This axiom is _logically_ redundant in the (logically complete)
        27-Jul-2011.)  (Proof modification is discouraged.)
        (New usage is discouraged.) $)
     axc11n-16 $p |- ( A. x x = z -> A. z z = x ) $=
-      ( weq wal ax-c16 alrimiv axc4i-o equequ1 cbvalv a1i imbi12d albidv biimpi
-      vw wi wb wex nfa1-o 19.23 alcoms albii ax6ev pm2.27 ax-mp alimi spv sps-o
-      equequ2 syl sylbi 3syl ) ABCZADZANCZUNADZOZNDZADZBNCZUSBDZOZNDZBDZBACZBDU
-      LUQAUMUPNUNABEFGURVCUQVBABULUPVANULUNUSUOUTABNHZUOUTPULUNUSABVEIJKLIMVBVD
-      BVAVDNBVABDZNDUSBQZUTOZNDZVDVFVHNUSUTBUSBRSUAVIUTNDVDVHUTNVGVHUTOBNUBVGUT
-      UCUDUEUSVDBNUSNDVDBUSVDNANABUHUFUGTUIUJTGUK $.
+      ( vw weq wal wi ax-c16 alrimiv axc4i-o equequ1 cbvalvw a1i imbi12d albidv
```
wlammen: This theorem (axc11n-16, there are more) is
1. in a mathbox not your own;
2. modifications are discouraged.

I don't see much harm.  Still I'd like to know why either condition does not prevent you from updating.

benjub: Good catch... I just did a global "search and replace" and a "verify proof *". I'm going to look into these.

benjub:
As for the latest commit: I checked the references books and made comment citations more faithful. (Not related to the previous commits of this PR, but I hope that the separate commit is enough to make it reviewable.)

set.mm
benjub: So, I looked at them. The "discouraged" tags are there to preserve features of the proofs that are not affected by the change cbv(al|ex)v-->cbv(al|ex)vw (e.g., not using the axiom of choice, or some of the auxiliary schemes of predicate calculus ax-10,11,12,13, etc.).  Furthermore, that change is very minor (which is why I did not add credits nor OLD versions), so I think it is ok to make this modification in the whole file, including mathboxes, as indicated in http://us2.metamath.org/mpeuni/mathbox.html (see sentence with "still part of shared body..."), especially considering the eventuality that some are moved to Main (we would have to keep track of them and do the change cbv(al|ex)v-->cbv(al|ex)vw then).

benjub:
Drawing @tirix's attention on the small edit (in the last commit) of the comment of neiptopreu.

set.mm
wlammen: > see sentence with "still part of shared body..."

The same page: 
`mathbox...is maintained by its contributor independently from the main part of set.mm`

Ok, I am not going to be knit-picky here.  So I will accept this.  I already verified this won't do harm.  Still, I think this should be evaluated in advance.

wlammen (approved):

PR #2745, opened 2022-08-08
Proofs of ~clwwlknonclwlknonf1o, ~clwwlknonclwlknonen, ~dlwwlknondlwlknonf1o shortened

avekens:
By generalizing their proofs (resulting in theorem ~f1ossf1o), the proofs of ~clwwlknonclwlknonf1o and ~dlwwlknondlwlknonf1o were shortened, especially by making the corresponding lemmas obsolete. As a side effect, the proof of ~clwwlknonclwlknonen was shortened, too.

Follow up PR of #2734: this makes the discussion about lengthening the proof of clwwlknonclwlknonf1olem obsolete.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4120,6 +4120,8 @@
 "clwwlknOLD" is used by "isclwwlknOLD".
 "clwwlknclwwlkdifsOLD" is used by "clwwlknclwwlkdifnumOLD".
 "clwwlknonOLD" is used by "isclwwlknonOLD".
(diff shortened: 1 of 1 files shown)


benjub (approved):
So rabrabi is not used anymore ? I would put it back in your mathbox, though if you think it is applicable for other theorems, you can leave it in Main.

avekens:
Although ~rabrabi is not used for ~clwwlknonclwlknonf1o anymore, it is still used for ~wlksnwwlknvbij. Therefore, it must remain in Main.

PR #2752, opened 2022-08-09
more mixed op results

wlammen:


set.mm
```diff
@@ -7643,7 +7652,7 @@ inference eliminates the need for a syllogism ( ~ syl ) -type inference
   1. Mixed Op means in this section, that both df-or and df-an is referenced in
   a theorem and its proof.  The first part is mostly populated with theorems
   that are in principle derivable from df-or only.  They have only ` \/ ` in
-  both the hypotheses and conclusion, but their proof benefit from shortcuts
+  both hypotheses and conclusion, but their proof benefit from shortcuts
```
benjub: Add an "s" to either "proof" or "benefit" (both spellings are used and both are ambiguous...).

benjub (approved):

wlammen:
This now finishes the restructuring.  In a few cases (like a2and) I'm not sure whether theorems rather belong to the 'Miscellaneous theorems of propositional calculus' section.  Ok, this can be decided later.  For now the work is done 
@benjub I still need your final approval

benjub (approved):

benjub:
@wlammen : as for ~a2and, it's AND-only, so it should go/remain before ~df-or.

wlammen:
This distinction is not really uphold throughout set.mm, see for example intnan.  Some theorems seem to have nothing in common with others and appear isolated, and a2and is one of them for me.  But the gory details can be left out for now.
I will revisit my decisions, but only after some time has passed.

benjub:
You're right that there will remain inconsistencies at some places. That said, if you can move intnan(r)(d) before wo/df-or that would be better. Actually, all theorems in that subsection not using df-or (resp. not using df-an) should be moved in the previous "AND-only" (resp. "OR-only") subsection. (**edit**: if you've had enough, I can do it)

PR #2754, opened 2022-08-09
prove in that order: vpwex > pwexg > pwex

benjub:
The following three successive theorems are currently proved in that order:
* http://us2.metamath.org/mpeuni/pwex.html
* http://us2.metamath.org/mpeuni/vpwex.html from pwex
* http://us2.metamath.org/mpeuni/pwexg.html from vpwex

But vpwex is actlually the penultimate proof step of pwex. Therefore, Prove vpwex first, then pwexg from it (proof unchanged), and finally pwex from pwexg.  This shortens the total number of proof steps (both essential and total).

As for the comments: I did check the referenced book, and their statement is actually vpwex, which is the main theorem, so I gathered information there, keeping the credit to the original contributors.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15623,6 +15623,7 @@ New usage of "equsb3ALT" is discouraged (0 uses).
 New usage of "equsexALT" is discouraged (0 uses).
 New usage of "equvelvOLD" is discouraged (0 uses).
 New usage of "equvinvOLD" is discouraged (0 uses).
+New usage of "eqvOLD" is discouraged (0 uses).
 New usage of "erngbase-rN" is discouraged (4 uses).
 New usage of "erngdv-rN" is discouraged (0 uses).
 New usage of "erngdvlem1-rN" is discouraged (3 uses).
@@ -19284,6 +19285,7 @@ Proof modification of "equsb3ALT" is discouraged (44 steps).
 Proof modification of "equsexALT" is discouraged (37 steps).
 Proof modification of "equvelvOLD" is discouraged (37 steps).
 Proof modification of "equvinvOLD" is discouraged (47 steps).
+Proof modification of "eqvOLD" is discouraged (6 steps).
 Proof modification of "estrresOLD" is discouraged (427 steps).
 Proof modification of "euexALT" is discouraged (32 steps).
 Proof modification of "eujustALT" is discouraged (188 steps).


benjub:
Second commit: for later theorems, there is no need to go out of our way to avoid dependency on some logical axioms, but in the present case, we are at the very early stages of set theory (still basic theorems to help use the language of classes), and the avoided axioms are the auxiliary schemes, so avoiding them is still worth the work at this stage.

benjub:
Third and last commit (ignoring the rewrap's): vnex was proved from vprc but was actually a proof step in it, so prove vprc from vnex instead, shortening combined proof length (both essential and total steps).

wlammen (approved):

wlammen:
I have no access to TakeutiZaring, so I cannot check whether its removal from pwexg is justified.  The technical changes are ok, though.  That's what the approval covers.

set.mm
```diff
@@ -44095,24 +44104,23 @@ holding in an empty domain (see Axiom A5 and Rule R2 of [LeBlanc]
...
+     depend on the Axiom of Regularity.  (Contributed by NM, 23-Aug-1993.) $)
+  vprc $p |- -. _V e. _V $=
+    ( vx cvv wcel cv wceq wex vnex isset mtbir ) BBCADBEAFAGABHI $.
+
   $( The universal class doesn't belong to any class.  (Contributed by FL,
      31-Dec-2006.) $)
   nvel $p |- -. _V e. A $=
     ( cvv wcel vprc elex mto ) BACBBCDBAEF $.
 
-  $( The universal class does not exist.  (Contributed by NM, 4-Jul-2005.) $)
```
avekens, line 44112: Although this is the original comment: this is a little bit confusing. Of course does the universal class "exist" (as a class). We should write "The universal class is not a set", shouldn't we?

avekens: Sorry, my comment is at the wrong place... It belongs to ~vnex.

avekens (changes requested):

set.mm
benjub: I changed it to "does not exist as a set" (since the statement reads "E. x x = ...").

benjub: @avekens : can you approve that new commit ?

avekens (approved):

PR #2755, opened 2022-08-10
Group miscellaneous theorems at the end of AND-only

wlammen:
1. At the end of the AND-only section there were lots of theorems that are not important for the foundations of conjunction or disjunction.  Sort them by topic, inside each topic by complexity, so that browsing is improved.
2. Move the XOR stuff to the Miscellaneous section.


set.mm
```diff
@@ -8056,6 +8067,31 @@ an illustration of the conservativity of definitions (definitions do not
...
+  including proofs by cases and other theorems with negated wffs.
+
+$)
+
+$( Theorems relating XOR to conjunction or dsjunction. $)
+
+  $( Theorem *5.16 of [WhiteheadRussell] p. 124.  (Contributed by NM,
+     3-Jan-2005.)  (Proof shortened by Wolf Lammen, 17-Oct-2013.) $)
+  pm5.16 $p |- -. ( ( ph <-> ps ) /\ ( ph <-> -. ps ) ) $=
+    ( wb wn wi wa pm5.18 biimpi imnan mpbi ) ABCZABDCZDZEKLFDKMABGHKLIJ $.
```
benjub, line 8093: this does not use df-or

wlammen: This is a compromise.  The XOR related theorems were all in one place, and I want to keep it like that.  The other option is to put this theorem to the NAND based ones, but that is not my preference, others might think different.

benjub: This should not block the merge, so you can merge now.

benjub (approved):

PR #2758, opened 2022-08-11
Show that NN+oo is omniscient

jkingdon:


wlammen:
This obviously only affects your mathbox, so whatever the contents is, I don't see a reason to object to the additions.

wlammen (approved):

PR #2759, opened 2022-08-11
move intnan* to AND only

wlammen:
1. intnan is moved close to simpri, a theorem it has close connections with.
2. We can use intnan to move bianf near to biantru in the AND-only section.  This mirrors the fact that biort and biorf are also close neighbours.
3. delete an outdated OLD theorem.

Sorry @avekens , the version this morning was premature.


avekens (approved):

avekens:
No problem, still approved.

PR #2761, opened 2022-08-12
Schroeder-Bernstein implies excluded middle

jkingdon:


wlammen (approved):

benjub:
In this and your previous PR: I think some statements would gain clarity by making a more systematic use of the natural injection I: \overline\N \to \N_\infty. Also, if I remember your previous PR, one statement basically proves I \circ succ = S \circ I on \N, that is, S is the successor on \N (of course, that identity morally holds on \overline\N, but for that one should redefine succ in this particular context to fix +\infty... probably not worth the trouble).

digama0:
Nice! If you are looking for more ideas, I would suggest also doing the forward direction, LEM implies schroeder-bernstein. As stated this is just the thing in set.mm, but you can also make a more "local" version that assumes only the decidability of some set with the necessary properties to distinguish the f and g^-1 sets in the mapping.

benjub:
By the way, @digama0 or @jkingdon : do you have (access to) a pdf of the cited article
> G. C.L. Banaschewski, B.; Brümmer. Thoughts on the Cantor-Bernstein Theorem. Quaes-
tiones Mathematicae, 9, 01 1986. doi:10.1080/16073606.1986.9632106.

(if yes, you know my email)

jkingdon:
> Nice! If you are looking for more ideas, I would suggest also doing the forward direction, LEM implies schroeder-bernstein. As stated this is just the thing in set.mm, but you can also make a more "local" version that assumes only the decidability of some set with the necessary properties to distinguish the f and g^-1 sets in the mapping.

Yeah, I was thinking of proving the forward direction before I made a bit more of a splash and added it to mm_100 and so on. I may want to enlist your help then, to bring the news to non-metamath-specific mailing lists/forums which might be interested.

As for how to prove the forward direction, I was just thinking of sticking in a decidability condition (`A. x STAB x e. A ` or whatever is needed in each spot) whereever set.mm uses excluded middle, mainly because that is less work than trying to rearrange the proof. But that is a starting point as much as the final word (for example, the set.mm proof avoids the axiom of infinity and I don't know whether that property would carry over without additional background work on theorems which rely on infinity and don't need to).

jkingdon:
> In this and your previous PR: I think some statements would gain clarity by making a more systematic use of the natural injection I: \overline\N \to \N_\infty. Also, if I remember your previous PR, one statement basically proves I \circ succ = S \circ I on \N, that is, S is the successor on \N (of course, that identity morally holds on \overline\N, but for that one should redefine succ in this particular context to fix +\infty... probably not worth the trouble).

If we had more agreement on these topics I suppose I wouldn't have put this into a mathbox.

But I don't know, perhaps there is more merit in things like the explicit `I` than I have seen to date, so I'm not trying to shut down trying different approaches and seeing where they lead. Which I guess is why I thought of mathboxes.

jkingdon:
> By the way, @digama0 or @jkingdon : do you have (access to) a pdf of the cited article

I get these offers from time to time (not just from you) and although I try to be practical I do have a preference for sources which anyone can readily access (I really love the way the HoTT book, for example, not only makes itself available freely but made it easy for me to contribute two small fixes - speaking of which I need to send in my errata to the PradicBrown2021 paper if I can figure out where and whether they are already in an errata file).

I guess the high minded way of saying this is that I do view metamath as an education project as much as it is a research project and I want to reduce friction in spreading mathematical ideas.

benjub:
> > By the way, @digama0 or @jkingdon : do you have (access to) a pdf of the cited article
> 
> I get these offers from time to time (not just from you) and although I try to be practical I do have a preference for sources which anyone can readily access (I really love the way the HoTT book, for example, not only makes itself available freely but made it easy for me to contribute two small fixes - speaking of which I need to send in my errata to the PradicBrown2021 paper if I can figure out where and whether they are already in an errata file).
> 
> I guess the high minded way of saying this is that I do view metamath as an education project as much as it is a research project and I want to reduce friction in spreading mathematical ideas.

Although I agree with the above, this seems unrelated to my initial question, which is not an offer but a request: I cannot get hold of a copy of that Banaschewski and Brümmer article, so if you or @digama0 (or anyone reading this) happen to have (access to) an electronic copy of it, I would be grateful if you could email it to me.

I share your frustration with researchers who do not make their research freely available, but here you cannot be too critical of Banaschewski and Brümmer because in 1986 they did not have the same options to disseminate their research. (But if your crisiticsm was directed towards Quaestiones Mathematicae for not releasing older issues like many journals do, then I share your criticism, which is why I'm not ashamed of asking a pdf copy, even though this may infringe some copyrights.)

jkingdon:
> not an offer but a request: I cannot get hold of a copy of that Banaschewski and Brümmer article

Ah, OK, I clearly wasn't reading very carefully.

No, I don't have a copy nor do I think I've ever seen it.

jkingdon:
> I may want to enlist your help then, to bring the news to non-metamath-specific mailing lists/forums which might be interested.

Maybe this should be qualified by "if they'd be interested" - Andrej Bauer has since pointed me to the formalization in https://www.cs.bham.ac.uk/%7Emhe/TypeTopology/CantorSchroederBernstein.CSB.html#6327 and although I assumed metamath probably wasn't the first, I didn't until now know of others.

As for my errata on the paper, I have sent an email to Pierre Pradic so I feel like I've done my duty on that.

PR #2762, opened 2022-08-13
revise pm3.22 so that it is not separated from ancom anymore

wlammen:
Replace the proof of pm3.22 with one of equal length, so it can better group with other ancom* theorems.

Since I replaced my own proof I neither added a label to the theorem, nor created an OLD version.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12069,14 +12069,6 @@
 "rngosn6" is used by "dvrunz".
 "rngoueqz" is used by "dvrunz".
 "rngoueqz" is used by "isdmn3".
-"rpnnen1lem1OLD" is used by "rpnnen1OLD".
-"rpnnen1lem1OLD" is used by "rpnnen1lem3OLD".
-"rpnnen1lem1OLD" is used by "rpnnen1lem4OLD".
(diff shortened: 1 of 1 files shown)


avekens (approved):

benjub:
I don't understand the need for the revision of the proof of pm3.22 since it seems that the current proofs allow to move it next to ancom ?
(But the PR looks fine)

wlammen:
Previously pm3.22 used pm3.21, which means both theorems and their immediate consequences were somehow intermingled.  Now both can be separated, and that is in fact done.  Actually, one can move 3.21 before jcad, which it has close connections with, but dfbi needs it, too, so it stays roughly where it was.  I deliberately moved dfbi to an early location, because of its greater theoretical value (as I see it).  This is one of the decisions I made based on my preferences.

All in all the rework is done, as fas as I am concerned, I think. So the titles and sections can now be developed, and of course, you may correct whatever you think needs a revision.

I removed some outdated OLD theorems, without noticing that meanwhile somebody has approved the PR.

benjub (approved):

PR #2764, opened 2022-08-13
shorten elsymdifxor

benjub:
Since the change is minor, I did not keep an OLD version.


wlammen (approved):

set.mm
```diff
@@ -35383,13 +35383,26 @@ technically classes despite morally (and provably) being sets, like ` 1 `
 -.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
 $)
 
+  ${
+    $d x A $.  $d x B $.
+    $( Two ways to express symmetric difference ( ~ df-symdif ).  (Contributed
+       by NM, 17-Aug-2004.)  (Proof shortened by Andrew Salmon,
+       26-Jun-2011.) $)
+    symdif2 $p |- ( ( A \ B ) u. ( B \ A ) ) =
```
avekens, line 35391: Why not using the definition df-symdif: ( A /_\ B ) = { x | -. ( x e. A <-> x e. B ) } ?

avekens (changes requested):

set.mm
benjub: I did not modify that proof.  Since it is not relying on df-symdif, I moved it before it.
You're right, there's a shorter proof from df-symdif, elsymdif, abbi2i. You may prove it that way if you want. And/or you may add the theorem you mention ( A /_\ B ) = { x | -. ( x e. A <-> x e. B ) } after elsymdif and mention in its comment that it permits to prove symdif2 straightfowardly.
I'm merging this and you may do it in another PR.

PR #25, opened 2016-06-25
Modify comments

david-a-wheeler:

PR #2765, opened 2022-08-13
Schroeder-Bernstein is equivalent to excluded middle

jkingdon:
As discussed in https://github.com/metamath/set.mm/pull/2761#issuecomment-1213007326 this proves both directions of `EXMID` being equivalent to the Schroeder-Bernstein Theorem. This is a whole lot closer to "just the thing in set.mm" than "assumes only the decidability of some set with the necessary properties " (in the words of that comment). One piece of good news: the proof preserves the set.mm property of not relying on the axiom of infinity (I didn't do any extra work to make this so, it just seems like all the theorems that we rely on already are set.mm-like in this sense).

There isn't a whole lot here other than bringing over https://us.metamath.org/mpeuni/sbth.html adding `EXMID` conditions where needed, but there are analogues to https://us.metamath.org/mpeuni/iman.html and https://us.metamath.org/mpeuni/dfss4.html which add conditions to make them provable in iset.mm. The other big result along these lines which we use is already in iset.mm: https://us.metamath.org/ileuni/undifdcss.html

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -322,7 +322,11 @@ href="mpeuni/pythagtrip.html">pythagtrip</a>, by Scott Fenton, 2014-04-19)</li>
 <!-- 3rd added to list -->
 <li><a name="25">25</a>.  Schroeder-Bernstein Theorem (<a
 href="mpeuni/sbth.html">sbth</a>, by Norman Megill, 1998-06-08 revised
-2007-01-20)</li>
+2007-01-20). The intuitionistic logic explorer (iset.mm) database shows
+that Schroeder-Bernstein is equivalent to excluded middle, resolving
+this problem in the negative, at <a
+href="http://us.metamath.org/ileuni/exmidsbth.html">exmidsbth</a>
+(by Jim Kingdon, 2022-08-13).</li>
 
 <!-- 45th added to list -->
 <li><a name="26">26</a>.  Leibniz' Series for Pi (<a
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1117,7 +1117,7 @@ then case elimination will work using theorems such as ~ exmiddc and ~ mpjaodan<
 
 <TR>
 <TD>iman</TD>
-<TD>~ imanim </TD>
+<TD>~ imanim , ~ imanst</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1434,7 +1434,7 @@ is double negation elimination.</TD>
 
 <TR>
   <TD>dfss4</TD>
-  <TD>~ ssddif</TD>
+  <TD>~ ssddif , ~ dfss4st</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3129,8 +3129,8 @@ other theorems we do not have</TD>
 <TR>
 <TD>sbth and its lemmas, sbthb , sbthcl</TD>
 <TD>~ fisbth</TD>
-<TD>That the Schroeder-Bernstein Theorem implies excluded middle
-is Theorem 1 of [PradicBrown2021], p. 1.</TD>
+<TD>The Schroeder-Bernstein Theorem is equivalent to excluded middle
+by ~ exmidsbth </TD>
 </TR>
 
 <TR>


benjub (approved):

iset.mm
```diff
@@ -64043,6 +64073,177 @@ would need to be strengthened (for example, to imply included middle) to
...
+        NVJKVRVHUDBUEZVOKVGVIUQZVJVPVRWETVJVHCLZWIVJWFWKWIPVJWFVQWGRCBVHUKQZTZA
+        BCDEFGHIJUOUPGBSURVJVLMZVLNZCOZVMVJKWBVHMZVHNCOZPZWIPVQWNWJVJWAWBWCTVJW
+        SWIVJWKWSVJWKWIWLRVHCSQZWMUSWHABCDEFGHIJUTVAVJKVTWRWIPZVQWPWJVJVSVTWDTV
+        JWRWIVJWQWRWTTWMUSWHKVTPXAVQPPWOGUDCGVBABCDEFGHIJVCVDVAVLCSURBCGVEUR $.
+
+      ${
+        $d f g A $.  $d f g B $.
+        sbthlem.4 $e |- B e. _V $.
+        $( Lemma for ~ isbth .  (Contributed by NM, 28-Mar-1998.) $)
+        sbthlemi10 $p |- ( ( EXMID /\ ( A ~<_ B /\ B ~<_ A ) ) -> A ~~ B ) $=
```
avekens, line 64205: Why are some lemmas called sbthlem* and some sbthlemi* ? What is the meaning of "i"? Can it be omitted?

avekens: Ah, I see: "i" means "intuitionistic version of" (additionally requiring EXMID as antecedent compared with the non-intuitionistic version). Maybe this should be explained in the first concerned lemma.

iset.mm
```diff
@@ -120972,6 +121173,17 @@ C_ if ( A. k e. suc J ( Q ` ( i e. _om
       IEGEHZIJZKUJLZUFMZNZODEGUHUNUCUJJZUDUKUGUMKUCUJIPUOUEULUFUCUJQRSTUAUB $.
   $}
 
+  ${
+    $d x y $.
+    $( The Schroeder-Bernstein Theorem is equivalent to excluded middle.  This
+       is Metamath 100 proof #25.  (Contributed by Jim Kingdon,
+       13-Aug-2022.) $)
```
avekens, line 121180: This is confusing: Metamath 100 proof #25 is not about the equivalence of the Schroeder-Bernstein Theorem and excluded middle, but "only" about the Schroeder-Bernstein Theorem, which would be ~isbth. Therefore, "This is Metamath 100 proof #25. " should be moved to the comment of ~isbth.

avekens (changes requested):

iset.mm
jkingdon: The rule is "Theorems which are the same as in set.mm should be named the same (that is,
where the statement of the theorem is the same; the proof can differ without
a new name being called for).  Theorems which are different should be named
differently" (quote from a comment near the top of iset.mm). As for the `i` this convention is used here and there starting with https://us.metamath.org/ileuni/ax-i9.html or so although I try to name things another way if I can think of a better way in a particular situation.

iset.mm
jkingdon: I'm not willing to give credit for isbth - that would reduce iset.mm to just ZFC with odd notation. No, in order to count as a solution we need to either prove Schoeder-Bernstein from the IZF axioms (example: https://us.metamath.org/ileuni/bezout.html , another Metamath 100 problem, is stated as in set.mm but needs a very different proof) or show that it is not possible. Of course, for Schoeder-Bernstein it is the latter and for this reason the https://us.metamath.org/ileuni/exmidsbthr.html direction is the more interesting of the two.

A more subtle example is "the square root of two is irrational" where constructive mathematicians have a definition of "irrational" which I think is how we should interpret this problem, but Schroeder-Bernstein doesn't have that sort of issue.

As for where to say this, maybe in mm_100.html? I'll push up a change to make this say "The intuitionistic logic explorer (iset.mm) database shows that Schroeder-Bernstein is equivalent to excluded middle, resolving
this problem in the negative, at exmidsbth"

iset.mm
avekens: Can you add the usage of an "i" to provide a different name as described in "1. Naming" in that section? For example:

> Theorems which are different should be named differently (although if additional hypotheses are added in iset.mm
> the name need not be changed). **Often, an "i" is added to achieve this (see, for example, ~sbthlemi10 versus sbthlem10 in set.mm).**

jkingdon: I'll make this a separate pull request.

benjub (approved):

iset.mm
jkingdon: Submitted as https://github.com/metamath/set.mm/pull/2769

Issue #2710, opened 2022-07-24
Extract common proof patterns into lemmata


savask:
Some statements are often reproved in set.mm over and over. The following list was identified with computer search:

https://gist.github.com/savask/dce5b48b7317e8144635fda09e983fae

The first column is 'score' which correlates with the potential amount of saved bytes, second is the number of essential hypotheses, and the third is the list of essential steps of the proof, see https://groups.google.com/g/metamath/c/aXvD9rs36Ps for details.

Per suggestion of @benjub this issue will be used to record which theorems were incorporated into set.mm and which were deemed unworthy. By line number in the gist:
* 1 as `fvexi` (#2766; was already in set.mm/Main, so only carry out `minimize_with fvexi`); lowers the scores of Lines 11, 25, 28, 33, 61
* 5 as `hllatd` (#2770)
* 8 as `ffnd` (#2771; was already in set.mm/Main, so only carry out `minimize_with ffnd`)
* 9 as `frnd` (#2773; was already in set.mm/mathbox, so move to Main and carry out `minimize_with frnd`)
* 10 as `fdm` (#2774; was already in set.mm/mathbox, so move to Main and carry out `minimize_with fdmd`)
* 11 score lowered by Line 1
* 15 as `ne0d` (#2775; was already in set.mm/mathbox, so move to Main and carry out `minimize_with ne0d`)
* 22 as `fmpttd` (#2776)
* 23 as `fvoveq1` (#2708, which also adds `fvoveq1d`)
* 25 score lowered by Line 1
* 27 as `brralrspcev` (#2782)
* 28 as `mptfvmpt` (#2784)
* 30 as `fssdm` (#2784, which also adds `fssdmd`)
* 32 score lowered by Line 42
* 33 score lowered by Line 1
* 34 discarded (see https://groups.google.com/g/metamath/c/aXvD9rs36Ps/m/QxTG7qDQBAAJ)
* 39 as `eqelssd` (#2784)
* 40 as `rspceaimv` (#2794)
* 42 as `toponrestid` with a slight modification (#2668); lowers scores of Lines 32, 66, 82, 88
* 47 as `elicc01` (#2782)
* 48 discarded (similar to Line 34)
* 54 as `animpimp2impd` (#2784)
* 55 discarded as too specialized
* 58 as `ovanraleqv` (#2768)
* 61 score lowered by Line 1
* 62 as `reximssdv` (#2784)
[...]

PR #2766, opened 2022-08-13
minimize_with fvexi: 574 proofs minimized

benjub:
Use scripts/min.cmd to "minimize_with fvexi". This corresponds to the first line in @savask's list at #2710. The theorem fvexi had been moved to main but the corresponding minimization hadn't been done.

The diff is huge, but nothing more has been done: only "minimize_with fvexi" on theorems using both fvex and eqeltri.

(**edit**: regenerate discouraged file since in the process, three discouraged uses were removed)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -9924,7 +9924,6 @@
 "nmcvfval" is used by "nvop".
 "nmcvfval" is used by "nvop2".
 "nmcvfval" is used by "phop".
-"nmcvfval" is used by "phpar".
 "nmfn0" is used by "branmfn".
 "nmfn0" is used by "nmbdfnlb".
 "nmfnlb" is used by "nmbdfnlbi".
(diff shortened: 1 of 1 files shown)


wlammen (approved):

wlammen:
I did a random check and every of the 50 cases I investigated was in fact a shortening (roughly 7 proof bytes per case), caused by the introduction of fvexi.

Then I looked into phpar, the theorem responsible for the changes in discouraged, but it was easy to see why fvexi avoided the use of the discouraged-use theorems.

This PR modifies theorems in mathboxes, but I think that is OK for such an automated process.

Finally: In the mentioned list https://gist.github.com/savask/dce5b48b7317e8144635fda09e983fae entry 11 and 25 seem to be affected by this PR.  Anything to be done here, @benjub ?

avekens:
I do not know if the 574 proofs overlap with the proofs I minimized (about 40). Shall I wait with my PR until this PR is merged? There may be merge conflicts anyway...

benjub:
> Finally: In the mentioned list https://gist.github.com/savask/dce5b48b7317e8144635fda09e983fae entry 11 and 25 seem to be affected by this PR. Anything to be done here, @benjub ?

Thanks. This will probably lower the scores of Lines 11 and 25 and make them less useful. I'll mention it there.

PR #2767, opened 2022-08-13
Move NN+oo earlier in iset.mm

### LINKED ISSUE #2713: Should NN_infinity be defined in terms of NN0 or _om ?
There has been a little discussion of this but I wanted to raise it to
a larger audience because (a) the best answer is not completely clear
to me, and (b) unlike definitions where we can prove two definitions
are equivalent, this one affects what the elements of ℕ<sub>∞</sub> are,
so alternative definitions end up being isomorphic rather than
equivalent and expressions like x ∈ ℕ<sub>∞</sub> behave differently
depending on what definition we choose.

I'll start with the non-metamath literature (as best as I can summarize
it).  Every source I can remember assumes that 2 = { 0 , 1 } and that
2<sup>ℕ</sup> is the set of sequences whose values are zero or one.
The question is how to best express that in iset.mm.

The current definition of ℕ<sub>∞</sub> is:

```
df-nninf $a |- NN+oo = { f e. ( 2o ^m _om ) |
  A. i e. _om ( f ` suc i ) C_ ( f ` i ) } $.
```

This is in some sense the most literal translation of the informal
2<sup>ℕ</sup> notation - we have https://us.metamath.org/ileuni/df2o3.html
which fits the 2 = { 0 , 1 } idea.

However, in some ways including theorems which connect this with
https://us.metamath.org/ileuni/df-xnn0.html this means a number of
conversions between _om and NN0 , for example at
https://us.metamath.org/ileuni/fxnn0nninf.html . So the question
is whether we'd prefer this definition:

```
df-nninfNN $a |- NN+oo = { f e. ( { 0 , 1 } ^m NN0 ) |
  A. i e. NN0 ( f ` ( i + 1 ) ) <_ ( f ` i ) } $.
```

There is at least some precedent for this - for example
https://us.metamath.org/mpeuni/eulerpartgbij.html .

I guess in general my assumption is we are better off working with NN0
once we get to the point of iset.mm where we can. And that the only
real downside in this case is that `{ 0 , 1 }` is more characters than
` 2o `. But are there implications I'm not seeing?

jkingdon:
Because it is in terms of 2o , ^m , _om , etc, this is a more logical
place for it.

Fixes #2713


iset.mm
```diff
@@ -65740,6 +65757,50 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
...
+      TVEPQZKZVDVKKZLZABMVIPVAVBFVKVARZVNVHABVOVLVFVMVGVEVKVATVDVKVATUOUPPAUQUR
+      US $.
+  $}
+
+  ${
+    $d N f i j $.
+    $( Elements of ` NN+oo ` corresponding to natural numbers.  The natural
+       number ` N ` corresponds to a sequence of ` N ` ones followed by zeroes.
+       Contrast to a sequence which is all ones as seen at ~ infnninf .
+       Remark/TODO: the theorem sill holds if ` N = _om ` , that is, the
```
avekens: sill -> still

avekens (approved):

iset.mm
jkingdon: Fixed

PR #2768, opened 2022-08-14
new theorems for minimizations

avekens:
* new theorem ~ovanraleqv created from @savask's list, No. 58, see issue #2710: 21 proofs (many of them in mathboxes) are shortened, saving 774 bytes.
* new theorem ~fv0p1e1 created from @savask's list, No. 63, see issue #2710: 20 proofs are shortened, saving 637 bytes.
* new theorem ~2fveq3 added: more than 300 proofs can be shortened (will be done in a separate PR)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -9924,7 +9924,6 @@
 "nmcvfval" is used by "nvop".
 "nmcvfval" is used by "nvop2".
 "nmcvfval" is used by "phop".
-"nmcvfval" is used by "phpar".
 "nmfn0" is used by "branmfn".
 "nmfn0" is used by "nmbdfnlb".
 "nmfnlb" is used by "nmbdfnlbi".
@@ -12069,14 +12068,6 @@
 "rngosn6" is used by "dvrunz".
 "rngoueqz" is used by "dvrunz".
 "rngoueqz" is used by "isdmn3".
-"rpnnen1lem1OLD" is used by "rpnnen1OLD".
-"rpnnen1lem1OLD" is used by "rpnnen1lem3OLD".
-"rpnnen1lem1OLD" is used by "rpnnen1lem4OLD".
-"rpnnen1lem1OLD" is used by "rpnnen1lem5OLD".
-"rpnnen1lem3OLD" is used by "rpnnen1lem4OLD".
-"rpnnen1lem3OLD" is used by "rpnnen1lem5OLD".
-"rpnnen1lem4OLD" is used by "rpnnen1lem5OLD".
-"rpnnen1lem5OLD" is used by "rpnnen1OLD".
 "rspsbc2" is used by "tratrb".
 "rspsbc2" is used by "tratrbVD".
 "sbc2rexgOLD" is used by "sbc4rexgOLD".
@@ -12529,7 +12520,6 @@
 "smfval" is used by "nvtri".
 "smfval" is used by "nvvop".
 "smfval" is used by "phop".
-"smfval" is used by "phpar".
 "smgrpassOLD" is used by "ismndo1".
 "smgrpismgmOLD" is used by "mndoismgmOLD".
 "smgrpmgm" is used by "ismndo1".
@@ -12933,7 +12923,6 @@
 "vafval" is used by "nvsz".
 "vafval" is used by "nvvop".
 "vafval" is used by "phop".
-"vafval" is used by "phpar".
 "vc0" is used by "nv0".
 "vc0" is used by "vcm".
 "vc0" is used by "vcz".
@@ -13396,7 +13385,6 @@ New usage of "0lno" is discouraged (3 uses).
 New usage of "0lnop" is discouraged (6 uses).
 New usage of "0lt1sr" is discouraged (2 uses).
 New usage of "0ncn" is discouraged (3 uses).
-New usage of "0nelxpOLD" is discouraged (0 uses).
 New usage of "0ngrp" is discouraged (2 uses).
 New usage of "0nnq" is discouraged (22 uses).
 New usage of "0npi" is discouraged (9 uses).
@@ -15590,7 +15578,6 @@ New usage of "elunirnALT" is discouraged (0 uses).
 New usage of "elunop" is discouraged (7 uses).
 New usage of "elunop2" is discouraged (0 uses).
 New usage of "elwwlks2ons3OLD" is discouraged (0 uses).
-New usage of "elxp2OLD" is discouraged (0 uses).
 New usage of "en3lpVD" is discouraged (0 uses).
 New usage of "en3lplem1VD" is discouraged (1 uses).
 New usage of "en3lplem2VD" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


savask:
Could you, please, remove "Proposed by @SAVASK" from theorem comments? My set.mm database initials are SS, but in any case I don't want any credit to me for these theorems.

set.mm
```diff
@@ -62783,17 +62788,33 @@ ordered pairs (for use in defining operations).  This is a special case
 
   ${
     fvoveq1d.1 $e |- ( ph -> A = B ) $.
-    $( Equality deduction for nested function and operation value.
+    $( Equality deduction for nested function and operation value.  Technical
```
wlammen, line 62791: It is an accepted idea to introduce an infrastructure theorem for the sole purpose of extracting repeated proof steps.  I did it myself on various occasions (see sylnbi and many many more).  This is covered by conventions.html:

> Usually we will also automatically accept a _new_ theorem that is used to shorten multiple proofs, if the total size of set.mm (including the comment of the new theorem, not including the acknowledgment) decreases as a result.

As a rule of thumb, Norm accepted one, if it was referenced at least seven times.  fvop1e1 is applied 20 times.  To best match the above rule, I would get rid of the second sentence (Technical...).  It wasn't considered necessary in the past, and I am not fond of starting this habit.

set.mm
```diff
@@ -62783,17 +62788,33 @@ ordered pairs (for use in defining operations).  This is a special case
...
+       theorem to be used to reduce the size of a significant number of proofs.
        (Contributed by AV, 23-Jul-2022.) $)
     fvoveq1d $p |- ( ph -> ( F ` ( A O C ) ) = ( F ` ( B O C ) ) ) $=
       ( co oveq1d fveq2d ) ABDFHCDFHEABCDFGIJ $.
   $}
 
   $( Equality theorem for nested function and operation value.  Closed form of
-     ~ fvoveq1d .  (Contributed by AV, 23-Jul-2022.) $)
+     ~ fvoveq1d .  Technical theorem to be used to reduce the size of a
+     significant number of proofs.  (Proposed by @SAVASK, 8-Jul-2022)
```
wlammen, line 62800: The Proposed label is not exactly a novelty (see elpreqprb), but still rare.  I occasionally attributed a proof idea to someone else in the comment (as in aevlem), if its complexity/ingenuity called for it.  But in this particular case not even the proof is meant...  Will this really create extra motivation for similar suggestions?

We agreed that the number of labels is in excess, and I changed my habits accordingly.  This seems to be contrary to this idea.  What do you (others) think?

set.mm
```diff
@@ -62783,17 +62783,34 @@ ordered pairs (for use in defining operations).  This is a special case
 
   ${
     fvoveq1d.1 $e |- ( ph -> A = B ) $.
-    $( Equality deduction for nested function and operation value.
+    $( Equality deduction for nested function and operation value.  Technical
+       theorem to be used to reduce the size of a significant amount of proofs.
        (Contributed by AV, 23-Jul-2022.) $)
     fvoveq1d $p |- ( ph -> ( F ` ( A O C ) ) = ( F ` ( B O C ) ) ) $=
```
benjub, line 62789: As mentioned in the PR that introduced it, I think this should be relabeled fvoveq2d (and same with fvoveq2) since it is the equality theorem when the **second** variable in the expression ``( F ` ( A O C ) )`` is substituted for an equal one.
A global text search&replace would do it (with a space preceding the word, i.e., " fvoveq1d" --> " fvoveq2d").

benjub (approved):
Approved, with the comments from @savask and me above.

set.mm
benjub: I agree with @wlammen.

set.mm
benjub: I think it's nice and "honest" to mention *significant* help from other people in a formalization, like @avekens, @jkingdon and you (@wlammen) did by mentioning me on a few occasions.  (I think @jkingdon and @avekens have sometimes been "too" nice by mentioning me when I just said "you could prove that" without giving details on the proof !).  As for the way to mention it, I have no real opinion and I'm fine with the various existing ones.

What do you mean by "the number of labels is in excess" ?

set.mm
avekens: I think this is a matter of opinion: it is the second variable in the whole expression, but it is the first variable in the operation (you could read "fvoveq1d" as "fv(oveq1)d". What do others think about this?

set.mm
avekens: I would agree for theorems like ~fvoveq1 (or ~fvoveq2, see above) and ~2fveq3 because they are self-explanatory. But what about 

```
ovanraleqv.1 $e |- ( B = X -> ( ph <-> ps ) ) $.
ovanraleqv $p |- ( B = X -> ( A. x e. V ( ph /\ ( A .x. B ) = C ) <-> A. x e. V ( ps /\ ( A .x. X ) = C ) ) )
```
and 
```
fv0p1e1 $p |- ( N = 0 -> ( F ` ( N + 1 ) ) = ( F ` 1 ) )?
```
Nobody who would stumble over these theorems could understand their meanings. They are really technical only. Things may be different for first-order logic  where a lot of varaitions of the same "idea" are provided as theorems (like ~sylnbi, which I would see in the same group of infrastructure theorems as ~2fveq3). Mathematically, theorems like ~ovanraleqv and ~fv0p1e1 do not make any sense.

As a compromise, we may make such theorems lemmas (~ovanraleqv and ~fv0p1e1 -> ~ovanraleqvlem and ~fv0p1e1lem), but this would still require an explanation in the comment, and it increases the (textual) size of the proofs unnecessarily.

set.mm
avekens: Since the Proposal labels are (and will be) rare, I do not see the danger/risk to clutter the comments with too many parentheticals of this kind.
I would leave it to the contributor to decide if he/she wants to mention another one who was involved (and to decide if the help was significant or not). Sometimes it is helpful just to raise a vague idea to formulate a theorem (and then to prove it). 

To provide this information in a semi-formal way (as parentheticals) has the advantage that it is available in a uniform way (e.g. also to seach for it).

By the was, I often got (and get) hints by Gérard Lang for new/additianal proofs, and I really would like to document in set.mm that this was his idea.

avekens: PS: If someone explicitly does not want to be mentioned (like @savask in this case), this should be respected of course.

avekens:
> Could you, please, remove "Proposed by @savask" from theorem comments? My set.mm database initials are SS, but in any case I don't want any credit to me for these theorems.

No problem, I will remove these parentheticals from the comments.

avekens:
As I feared, I have to solve the merge conflicts now.

avekens:
> As I feared, I have to solve the merge conflicts now.

Fortunately, only 2 theorems/proofs are affected.

set.mm
avekens: I removed "Technical theorem to be used to reduce the size of a significant number of proofs." from the comments except for ~ovanraleqv and ~fv0p1e1.

set.mm
avekens: The current Proposal labels are removed now from this PR (as @savask asked for). Therefore, the discussion about this issue should be continued elsewhere (if necessary).

set.mm
benjub: My justification is: the fact that they are technical lemmas is obvious from their statement, and the fact that they are used in many theorems appears on the webpage ("this theorem is referenced by"). So the comment "Technical theorem..." is not needed. It does not hurt either, so no strong opinion.

set.mm
benjub: Ok. Both are justifiable, and that's not that important. I'm fine with both.

benjub (approved):

PR #2769, opened 2022-08-14
Tweaks to iset.mm naming rules

jkingdon:
1. Move them from the header of iset.mm to the conventions section, so
   they go alongside the other naming rules.
2. Re-word the parenthetical about exceptions to the rule about
   different theorems being named differently. The old text was
   "although if additional hypotheses are added in iset.mm the name
   need not be changed" and this is an awkward way to try to make a general
   rule to explain fvex and opex (the compatibility issues with added
   hypotheses are if anything worse than for a change in the conclusion
   because the error message is confusing). It is better to just be honest
   that sometimes we make an exception, so the new wording is "we do have a
   small number of intentional exceptions to this rule but on the whole
   it serves us well"
3. Add paragraph about the convention of adding "i" to an iset.mm name.


avekens (approved):

iset.mm
```diff
@@ -116375,6 +116354,33 @@ Metamath Proof Explorer (aka "set.mm").  This list of conventions is
...
+       in the theorems. For example, if a theorem in set.mm is an equality
+       and the iset.mm analogue is a subset, add "ss" to the iset.mm name.
+       If need be, add "i" to the iset.mm name (usually as a prefix to
+       some portion of the name).</p>
+
+       <p>As with set.mm, we welcome suggestions for better names (such
+       as names which are more consistent with naming conventions).</p>
+
+       <p>We do try to keep set.mm and iset.mm similar where we can. For
+       example, if a theorem exists both places but the name in set.mm
```
wlammen, line 116377: typo
...exists in both places...

wlammen (changes requested):

iset.mm
jkingdon: I think either way is grammatical but :shrug: I guess "in both places" is a little better. Fixed.

wlammen (approved):

PR #2770, opened 2022-08-14
Add hllatd and shorten 357 proofs

benjub:
This corresponds to Line 5 of #2710.  First commit adds ~hllatd; second commit makes the automatic `MM> minimize_with hllatd` on theorems using ~hllat, using `scripts/min.cmd`.

Eventhough it is in Norm's mathbox, I decided to add it since it minimizes 357 proofs (each from 4 to around 10 bytes), so it largely pays for itself.  Since the added theorem ~hllatd is simply the deduction form of the existing theorem ~hllat, I kept the same contribution/date tags.  (In case one is interested in the precise history of the file set.mm, git permits to get it.)


wlammen (approved):
Skimmed through the file.  Affected lines are from NM's mathbox only.  All investigated theorems were shortenings.

savask:
> Since the added theorem ~hllatd is simply the deduction form of the existing theorem ~hllat, I kept the same contribution/date tags. (In case one is interested in the precise history of the file set.mm, git permits to get it.)

I'm not sure what are the conventions on this, but Norm didn't contribute `hllatd` and it was definitely not added on 20-Oct-2011. I understand omitting overly detailed information and relying more on git history, but the current tags of `hllatd` are incorrect.

avekens:
> > Since the added theorem ~hllatd is simply the deduction form of the existing theorem ~hllat, I kept the same contribution/date tags. (In case one is interested in the precise history of the file set.mm, git permits to get it.)
> 
> I'm not sure what are the conventions on this, but Norm didn't contribute `hllatd` and it was definitely not added on 20-Oct-2011. I understand omitting overly detailed information and relying more on git history, but the current tags of `hllatd` are incorrect.

I would also propose that @benjub is the contributor, and the date should be 14-Aug-2022, since ~hllat is still existing. If ~hllat had been deleted/replaced, then the original contributor/date should have been kept, together with a "revised by" tag.

avekens (approved):

PR #2771, opened 2022-08-14
minimize_with ffnd; shortens 536 proofs

benjub:
This corresponds to Line 8 of #2710. Only automatic use of `scripts/min.cmd`


wlammen:
I skimmed through set.mm.  Each investigated case was a shortening.  Mathboxes are affected.

wlammen (approved):
Random checks.  Each investigated case was a shortening. Mathboxes are affected.

avekens:
"because the diff is too large", nothin can be seen here at github. @benjub  ould you please add at least the new theorem here (within a comment)?.

wlammen:
The diff begins here:

```
--- a/set.mm
+++ b/set.mm
@@ -56241,7 +56241,7 @@ $)
   $( A Cartesian product with a singleton is a constant function.  (Contributed
      by NM, 24-Jul-2014.) $)
   fnconstg $p |- ( B e. V -> ( A X. { B } ) Fn A ) $=
-    ( wcel csn cxp wf wfn fconstg ffn syl ) BCDABEZALFZGMAHABCIALMJK $.
+    ( wcel csn cxp fconstg ffnd ) BCDABEZAIFABCGH $.
```

You can see that ffn was replaced with ffnd, a theorem already available.  The rest of the diff looks alike (as far as I have checked it).

One way to handle these big changes is to download the raw file and compare it to a local copy of the develop branch on your computer. The download is available via click on the 'File changed' tab, then clicking on the far right '...' near the checkbox 'Viewed', selecting 'View file', further clicking on 'View raw', and then either copying the browser contents to an editor, or saving it to  your 'Downloads' folder.  The final step depends on your software, of course.

avekens:
> You can see that ffn was replaced with ffnd, a theorem already available. The rest of the diff looks alike (as far as I have checked it).

Ah, ~ffnd was already available - this was the information I missed.
> 
> One way to handle these big changes is to download the raw file and compare it to a local copy of the develop branch on your computer. The download is available via click on the 'File changed' tab, then clicking on the far right '...' near the checkbox 'Viewed', selecting 'View file', further clicking on 'View raw', and then either copying the browser contents to an editor, or saving it to your 'Downloads' folder. The final step depends on your software, of course.

Thanks for the hint. I think it is sufficient that only one reviewer checks this way if only automatic changes were performed.

avekens (approved):

PR #2772, opened 2022-08-15
Propositional calculus TOC

### LINKED ISSUE #2751: Sectioning of the propositional calculus part of set.mm
After @wlammen's huge work of refactoring the somewhat messy section on conjunction and disjunction (thanks!), I propose to modify the TOC from
```
      ​​*1.2  Propositional calculus
            1.2.1  Recursively define primitive wffs for propositional calculus   wn 3
            ​*1.2.2  The axioms of propositional calculus   ax-mp 5
            ​*1.2.3  Logical implication   mp2 9
            ​*1.2.4  Logical negation   con4 113
            ​*1.2.5  Logical equivalence   wb 196
            ​*1.2.6  Logical conjunction and disjunction   wa 382
                  *1.2.6.1  Logical conjunction   wa 382
                  ​*1.2.6.2  Logical disjunction   wo 826
                  1.2.6.3  Assorted results not assigned yet   animorl 918
            ​*1.2.7  Miscellaneous theorems of propositional calculus   pm5.62 995
            ​*1.2.8  The conditional operator for propositions   wif 1049
            [...]
```
to
```
      ​​*1.2  Propositional calculus
            1.2.1  Recursively define primitive wffs for propositional calculus   wn 3
            ​*1.2.2  The axioms of propositional calculus   ax-mp 5
            ​*1.2.3  Implication   mp2 9
            ​*1.2.4  Negation   con4 113
            ​*1.2.5  Equivalence   wb 196
            ​*1.2.6  Conjunction   wa 382
            *1.2.7  Disjunction   wo 826
            *1.2.8  Mixed connectives   animorl 918
            ​*1.2.9  The conditional operator for propositions   wif 1049
            [...]
```
and then we can adjust the various section comments.  With the current axiomatization, I guess we cannot do much better organized without going out of our way.  What do you think ?

Actually, I would even begin with `1.2.1  Implication` (beginning with wi, ax-1, ax-2) and `1.2.2  Negation` (beginning with wn, ax-3), etc.  I once suggested it to Norm who was not opposed in principle, but I didn't get to propose anything in time.  Opinions ?

benjub:
Reorganization of the propositional calculus part of the TOC, following #2751.

I placed a "note on definitions" in the head comment of the section that contains the first definition, so as to be able to refer to it later and not have to repeat things.


wlammen (approved):
I like it.

set.mm
```diff
@@ -3849,34 +3860,34 @@ the conclusion alone (see ~ pm5.1im and ~ pm5.21im ).  (Contributed by
...
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
-  Logical conjunction and disjunction
+  Logical conjunction
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 
-  Here we define disjunction (logical 'or') ` \/ ` ( ~ df-or ) and conjunction
-  (logical 'and') ` /\ ` ( ~ df-an ). We also define various rules for
-  simplifying and applying them, e.g., ~ olc , ~ orc , and ~ orcom .
+  This section defines conjunction of two formulas, denoted by infix `/\` and
+  read "and."  It is defined in terms of implication and negation, which is
```
avekens, line 3867: "and".

set.mm
```diff
@@ -7096,31 +7107,33 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
...
--.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
   Logical disjunction
--.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 
-  We continue with logical disjunction.  This particular section is dedicated
-  to theorems proved without ~ df-an .
+  This section defines disjunction of two formulas, denoted by infix `\/` and
+  read "or."  It is defined in terms of implication and negation, which is
```
avekens, line 7117: "or".

set.mm
```diff
@@ -7096,31 +7107,33 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
...
-  ~ anor .  And if we consistently replace ` /\ ` occurrences even in their
-  proofs, then we can reconstruct all the results using ` \/ ` instead of
-  ` /\ `.  That is why ` /\ ` and ` \/ ` are known as dual operations.  Either
-  is capable of developing all of the logic the other can represent.  In
-  practise ` /\ ` is often preferred, its big advantage being grouping
-  antecedents in a convenient way.
+  As mentioned in the "note on definitions" in the section comment for logical
+  equivalence, all theorems in this and the previous section can be stated in
+  terms of implication and negation only.  Additionally, in classical logic
+  (but not in intuitionistic logic: see ise.mm), it is also possible to
```
avekens, line 7126: ise.mm -> iset.mm

set.mm
```diff
@@ -7096,31 +7107,33 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
...
+  translate conjunction into disjunction and conversely via the De Morgan law
+  ~ anor : conjunction and disjunction are dual connectives.  Either is
+  sufficient to develop all propositional calculus of the logic (together with
+  implication and negation).  In practice, conjunction is more efficient, its
+  big advantage being the possibility to use it to group antecedents in a
+  convenient way, using ~ imp and ~ ex as noted in the previous section.
+
+  An illustration of the conservativity of ~ df-an is given by ~ orim12dALT ,
```
avekens, line 7134: df-an -> df-or?

avekens (changes requested):
only minor changes requested.

set.mm
wlammen: I was about to write a similar comment, but looked the issue up in the internet first.  It seems the American grammar puts the period within the quotes, see https://www.grammarly.com/blog/quotation-marks/ or https://www.thesaurus.com/e/grammar/does-punctuation-go-inside-or-outside-quotation-marks/.  Maybe one should not use quotation marks here, but simple apostrophes, since it is not really a quote.
```
Quotation marks are for when you want to use someone else’s words in your writing. Let’s say you
want to write about something you heard your friend say. You could do it like this:

John said, “I really hate when it’s hot outside.”
```

set.mm
benjub: fixed

set.mm
benjub: It is df-an: orim12dALT is an alternate proof of orim12d not using df-an, illustrating conservativity of that definition.

set.mm
benjub: I also prefer the semantically better `"and".` but unfortunately, American usage puts the period within the quote. As for single/double quote, I think this is byzantine, so I'll let it that way, but you can change them later.

A quick search through set.mm shows that neither of these rules is applied consistently.

wlammen: One could actually conceive a case where it is misleading:

The ASCII translation for the operator "for all" is "A."

avekens: Well, '."' is used many times in set.mm, so it seems to be OK.

set.mm
avekens: OK, I see...

set.mm
wlammen: Just in case this particular thread of discussion blocks the merge, the result is good enough for me.  I don't oppose to merging it as is.

benjub:
Thank you @avekens and @wlammen. I was waiting a bit before merging, in case others wanted to react, but I guess it will still be time after the merge, so I'm merging now.

set.mm
jkingdon: There's more than anyone probably wants to read about this topic at https://en.wikipedia.org/wiki/Wikipedia:Logical_quotation_on_Wikipedia . Personally, I'd vote for the logical quotation style but I'm not sure how deep we need to get into trying to standardize this sort of thing.

benjub: @jkingdon: it would certainly be nice to standardize to the logical quotation style. Not volunteering for a global update though... I'm not sure how hard it would be (e.g., does a search for `".` and `",` yield some false positives?) I'll use logical style from now on.

avekens: I would also prefer the "logical quotation style" (especially because metamath is about logic ;-) ). Then the sentence in the comment discussed above would be changed to:

> This section defines conjunction of two formulas, denoted by infix `/\` and read "and".

I would propose to add this to our conventions, adding the remark that this style should be used from now on, and existing violations should be corrected on occasion. Or is someone volunteering for a global update? @david-a-wheeler @wlammen would you agree?

david-a-wheeler: I also prefer the logical quotation style over typographical quotation style. In technical works the typographical quotation style often leads to misleading statements and errors, so I think many mathematical and computer science books are using logical style. Wikipedia requires it. [The Jargon File](http://www.catb.org/~esr/jargon/html/writing-style.html) talks about it. They're sometimes called "British" vs. "American" style, but that's a little misleading; publications on both sides of the pond use both styles. See: https://en.wikipedia.org/wiki/Wikipedia:Logical_quotation_on_Wikipedia

I suggest adding it to the set of conventions for *future* changes. We can then slowly transform as we go, and commit a mass-change of the rest at some time in the future.

david-a-wheeler: I created an issue that proposes using logical quotation here: https://github.com/metamath/set.mm/issues/2786

benjub:
Closes #2751.

PR #2773, opened 2022-08-15
Minimize using frnd

benjub:
This corresponds to Line 9 in #2710.
First commit moves frnd from mathbox to Main.
Second commit: only automatic minimization.
Third commit: only automatic minimization.


wlammen:
It is not always easy to see, that the automatic process runs flawlessly.  dffi3 is such an example.  It uses frn in two locations and only one is optimized.  The resulting diff disturbs my 'automatic' verification process.

The second commit modifies mathboxes.  Each investigated case was a shortening.
The third commit modifies mathboxes.  Each investigated case was a shortening

avekens (approved):
I only looked at commit [fc3fefb](https://github.com/metamath/set.mm/pull/2773/commits/fc3fefb476d830c0f39037cd9b170e8d60df9b47)

benjub:
When I write "only automatic minimization", I do the following:
```
$ metamath
MM> read set.mm
MM> open log 1.tmp
MM> show usage frn
MM> close log
MM> submit "scripts/min.cmd"
MM> tools
TOOLS> substitute 1.tmp xxx frnd a ""
TOOLS> exit
MM> open log log.txt
MM> submit 1.tmp
MM> close log
$ git commit -am "'minimize_with frnd' on theorems using frn; 372 proofs shortened"
```

benjub:
> dffi3 is such an example. It uses frn in two locations and only one is optimized.

dffi3 uses frn in three places, but only one is followed by syl, which is the one which is minimized. The other two are followed by ax-mp (so the proof could be shortened by a new "frni")

wlammen (approved):

PR #2774, opened 2022-08-15
Move fdmd to Main, minimize

benjub:
This corresponds to Line 10 in #2710.
First and third commits:only automatic minimization.


wlammen (approved):
As usual: random checks.  One investigated case shows that the minimization process is not perfect: in ~mndpsuppss a possible shortening of a ~fdm and ~3syl was not caught.  Mathboxes are affected.  The file size shrinks by 47 lines.

avekens (approved):

PR #2775, opened 2022-08-15
Move ne0d to Main; minimize 249 proofs

benjub:
This corresponds to Line 15 in #2710.
The first three commits are from the previous PR... I should rebase when both are merged. Then, one commit moves ne0d to Main and the next one is only automatic minimization.


avekens (approved):
Only commit [3bde9be](https://github.com/metamath/set.mm/pull/2775/commits/3bde9befd07fa142a20e81254be45f38c3ab5b3b) checked.

wlammen:
I did not check this PR, because I wanted to have the rebase done first.  It is not impossible (actually I know exactly what to do), but cumbersome to emulate the rebase locally on my computer first, and then checkout this PR to reduce the diff to the rebased version only.

benjub:
@wlammen : sorry, I took @avekens's approval as an agreement to merge. Is an action required on my part ?

wlammen:
No, if you strongly believe that the automatic procedure needs no review, it simply does what is expected. Yes if you think, one should always verify to some extent, nothing weird happened.  My checks are never thorough, I am not going to look at 500+ modifications to catch any regression in any niche of the software.  The NASA wouldn't let you go with the first option, and my statistic review style is certainly viewed inappropriate, if human life depends on it.
What are the worst implications, if no review occurs? Techno-religious people may think, everything went according to plan, no need to bother further.  But if in fact something really went wrong, the damage is limited.  There is still the general verification process, that catches errors, so only optimization might not be carried out properly.  This might be viewed as an acceptable risk by a majority.
If considered necessary, I can do a review in retrospect.

benjub:
Thanks for the info @wlammen. In the present case, the unchecked commit is really exactly the output of `scripts/min.cmd` applied as written, and as you said the CI tests would catch real problems. So no further reviewed needed in the present case.

PR #2776, opened 2022-08-16
Add fmptd and minimize

benjub:
This corresponds to Line 22 of #2710.
Third commit is only the automatic minimization.


avekens (approved):
only the first two commits checked.

wlammen (approved):
Again I investigated a small portion of the modifications only.  In each case the optimization reduced the number of proof bytes, in total the file shrinks by 60 lines.  Mathboxes are affected.  In particular, some theorem comments were modified.  To the better, I'd say, wrt grammar or wording. Still I don't know whether consent of the mathbox owner can automatically be assumed here.

PR #2778, opened 2022-08-16
remove duplicate snunioo2 in favor of ioounsn

benjub:
Remove duplicate snunioo2 in favor of ioounsn; move it to Main with the needed xrltled, xrleidd; minimize_with these.
A duplicate from the list at #2128.
I should have done the minimization in a separate commit... I didn't expect it to be that large. Fortunately, it's still visible on github.


wlammen (approved):
The minimizations account for a reduction of at least 19 lines.  So they are ok I think.  Mathboxes are affected, even not counting the moved theorems.

benjub:
This one has merge conflicts, so it'll be faster to redo it than to resolve conflicts.

PR #2780, opened 2022-08-18
new theorem for minimizations, cleanup symdif

avekens:
Cleanup symdif (see also PR #2764) - 1. commit:
* ~symdif2 revised (and renamed ~dfsymdif4)
* ~difsymdifssd added
* usage of ~symdif2 replaced (in ~mbfeqalem)

Contribution for #2710, item no. 90 - 1. commit:
* ~imbrov2fvoveq added

Contribution for #2710, item no. 90 - 2. commit:
* ~eel1111 (renamed ~syl1111anc) moved from AS's mathbox
* 17 proofs (10 in mathboxes) shortened using ~imbrov2fvoveq (saving 806 bytes). Some of them further shortened, e.g., using ~syl1111anc.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+18-Aug-22 eel1111   syl1111anc  moved from AS's mathbox to main
 26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main
 26-Jul-22 wl-jarli  jarli       moved from WL's mathbox to main
 25-Jul-22 bj-sb3b   sb3b        moved from BJ's mathbox to main
--- a/discouraged
+++ b/discouraged
@@ -18114,6 +18114,7 @@ New usage of "syl5imp" is discouraged (0 uses).
 New usage of "syl5impVD" is discouraged (0 uses).
 New usage of "syld3an1OLD" is discouraged (0 uses).
 New usage of "syld3an2OLD" is discouraged (0 uses).
+New usage of "symdif2OLD" is discouraged (0 uses).
 New usage of "tb-ax1" is discouraged (3 uses).
 New usage of "tb-ax2" is discouraged (1 uses).
 New usage of "tb-ax3" is discouraged (2 uses).
@@ -20113,6 +20114,7 @@ Proof modification of "syl5imp" is discouraged (23 steps).
 Proof modification of "syl5impVD" is discouraged (57 steps).
 Proof modification of "syld3an1OLD" is discouraged (23 steps).
 Proof modification of "syld3an2OLD" is discouraged (23 steps).
+Proof modification of "symdif2OLD" is discouraged (59 steps).
 Proof modification of "tb-ax1" is discouraged (4 steps).
 Proof modification of "tb-ax2" is discouraged (3 steps).
 Proof modification of "tb-ax3" is discouraged (3 steps).


set.mm
```diff
@@ -325401,29 +325437,29 @@ or are almost disjoint (the interiors are disjoint).  (Contributed by
...
-        VHVITUURUUSFPNZUUOFRUVBUUQAUUSUVCUVAKVGBDFPXRXRVHVITVJVKUUPUULBMUUJUUPM
-        WCBUUCUUFBDEYNVLBDFYNVLVNUUMYNRUUNUUCUUOUUFUUMYNXKVOUUMYNXRVOVPVQVRVSVT
-        WAWBWDUUAXKDWEZYOUUESAUVDYTADPXKABDEPJWFZWGWHDYNXLXKWIWJUUAXRDWEZYPUUHS
-        AUVFYTADPXRABDFPKWFZWGWHDYNXLXRWIWJWKWNWLWOWMZWPZGWQAYICUSCPUSZCUTQVARZ
-        YMUVIGHYICWRWSYIWTTZXMYIXAXBXCAYHXOAYFPUSYFUTQVARZYHAYFCPAYIYFCUVHXDZGW
-        QAYFCUSUVJUVKUVMUVNGHYFCWRWSYFWTTZWHYEYFXETXFAXTOZXMYKYIUMZXNXMXSUNUVPY
-        KXNNYLUVQXNNUVPYKXSYFULZXNYKYEUVRXMXSUOXSXMUPUQXTXTYHUVRXNNAXTURUVOXSYF
-        XAXBXCAYLXTUVLWHYKYIXETXFXGXHADPXKXIYBXQSUVELDXKXJWJADPXRXIYCYASUVGLDXR
-        XJWJWK $.
+        ( vy vz wcel wa cdif cr cfv wceq wb cmpt ccnv cv cima cvol cdm cioo crn
```
wlammen, line 325440: The new proof of mbfeqalem is longer than the old one: OLD 15 lines + 27 Bytes vs NEW 15 lines + 44 Bytes.  Why should we prefer the longer one?

wlammen (changes requested):

set.mm
avekens: This change was not in the scope of a minimization, but during the cleanup of symdif.

wlammen: So the new symdif in this particular case produces a longer proof, and this cannot be avoided?

avekens: Maybe it could, but this would mean to restructure the proof of mbfeqalem completely. In this case, the OLD theorem symdif2OLD had to be replaced, and this caused the (slight) enlargement of the proof. 
As already discussed before, to shorten the proofs of high level theorems should not be the primary goal, but clarity and using definitions consequently are also very important (which is the objective here).

wlammen: So far I am not able to do a side-by-side comparison.  I have to finish my work first, before further looking into this.

avekens: All minimizations are in the second commit [77b1980](https://github.com/metamath/set.mm/pull/2780/commits/77b1980a8ffa8ccef08b0f6c173698cfc8328239) only (if that helps...).

wlammen: I now had a closer look into the proof of mbfeqalem.  In order to avoid a longer proof, you need a companion theorem to difsymdifssd:  (𝜑 → (𝐴 △ 𝐵) ⊆ 𝐶) => (𝜑 → (𝐵 ∖ 𝐴) ⊆ 𝐶).  The current set of theorems are biased towards 𝐴 ∖ 𝐵.  If you need the reverse difference, as in mbfeqalem, then the proofs become more complicated.  This is the background for the observed regression.

I leave it open to further discussion whether one should lift this bias.  For me this issue is resolved.

wlammen (approved):

set.mm
benjub: How could we keep @wlammen's information ? By putting it as a TODO in the comment of mbfeqalem ?

benjub:
As for
```
difsymdifssd.1 $e |- ( ph -> ( A /_\ B ) C_ C ) $.
difsymdifssd $p |- ( ph -> ( A \ B ) C_ C ) $=
```
maybe the following theorem, which is more direct, would be sufficient:
```
symdifssdiff $p |- ( A /_\ B ) C_ ( A \ B ) $=
```
(**edit:** reverse inclusion, obviously! Thanks @wlammen. How embarassing!)
since then one can use it thanks to
```
syl5ss $e |- A C_ B $e |- ( ph -> B C_ C ) $p |- ( ph -> A C_ C )
```

wlammen:
symdifssdiff (should be difsymdifss) is the other way round: ((𝐴 ∖ 𝐵) ⊆ (𝐴 △ 𝐵)).

Following the idea of building up some primitive infrastructure around symmetric difference step by step, IMO, this is a sensible move, even if there are not many applications yet.

avekens:
I shortened the proof of ~mbfeqalem by extracting lemma ~mbfeqalemlem (net size reduced from 1269 bytes to 491+666 = 1157 bytes). Regarding symdif, the need for using ~symdifcom was avoided by this "trick". I hope this satisfies the required criteria now.

Additionally, ~difsymssdif as proposed by BJ is added (but not used yet).

avekens:
> 
> Additionally, ~difsymssdif as proposed by BJ is added (but not used yet).

OK, ~difsymssdif was used to shorten the proof of ~difsymdifssd at least ...

wlammen:
Sorry, @avekens.  There is still a problem with the theorem names.  difsymssdif is the newly added theorem, and its name is not inline with the name of the deduction form difsymdifssd.

avekens:
> Sorry, @avekens. There is still a problem with the theorem names. difsymssdif is the newly added theorem, and its name is not inline with the name of the deduction form difsymdifssd.

What would be your suggestion? I would propose the following:
* difsssymdif $p |- ( A \ B ) C_ ( A /_\ B ) (currently difsymssdif): the DIFference is a SubSet of the SYMmetric DIFference
* difsymdifssd.1 $e |- ( ph -> ( A /_\ B ) C_ C ) $. difsymdifssd $p |- ( ph -> ( A \ B ) C_ C ) (no renaming): a DIFference is a subset of a class if the SYMmetric DIFference is a SubSet of this class.

The latter theorem is not the deduction form of the first one, therefore appending a "d" would not be appropriate.

wlammen:
Looks good from my POV

benjub:
> * difsssymdif $p |- ( A \ B ) C_ ( A /_\ B ) (currently difsymssdif): the DIFference is a SubSet of the SYMmetric DIFference
> 
> * difsymdifssd.1 $e |- ( ph -> ( A /_\ B ) C_ C ) $. difsymdifssd $p |- ( ph -> ( A \ B ) C_ C ) (no renaming): a DIFference is a subset of a class if the SYMmetric DIFference is a SubSet of this class.

Yes for the first one (difsssymdif). The sense of my proposal above where I mentioned syl5ss was to actually replace the deduction by difsssymdif, not having both. But both can stay. For the latter, note the similar example `inss1/ssinss1`, so maybe a label like `difsymssdifssd` (to mean: if difsym is a subset (of C) then dif is a subset (of C), deduction form). I think "difsym" should come before "sym" in the label, since it is a hypothesis.

wlammen:
I already had a look at your extraction of mbfeqalemlem.  This is nice.  If only one could single out each direction of the impbida into a common subproof...  But I see the difficulties, so I will accept the current solution.

avekens:
Labels renamed as discussed.

> I already had a look at your extraction of mbfeqalemlem. This is nice. If only one could single out each direction of the impbida into a common subproof... But I see the difficulties, so I will accept the current solution.

I did exactly this, but having two small lemmata for a lemma seemed overblown for me, therefore I discarded it.

set.mm
```diff
@@ -325398,32 +325439,40 @@ or are almost disjoint (the interiors are disjoint).  (Contributed by
     ${
       mbfeqalem.4 $e |- ( ( ph /\ x e. B ) -> C e. RR ) $.
       mbfeqalem.5 $e |- ( ( ph /\ x e. B ) -> D e. RR ) $.
-      $( Lemma for ~ mbfeqa .  (Contributed by Mario Carneiro, 2-Sep-2014.) $)
+      $( Lemma for ~ mbfeqalem .  (Contributed by Mario Carneiro, 2-Sep-2014.)
+         (Revised by AV, 19-Aug-2022.) $)
+      mbfeqalemlem $p |- ( ph -> ( ( `' ( x e. B |-> C ) " y )
```
benjub, line 325444: Side note: labels for lemmas generally follow the naming xxxlem1, xxxlem2... instead of xxxlemlem..., even if a lemma is referencing another. But this is very minor, and does not even need to be changed IMO.
So: ready to merge.

benjub (approved):

PR #2783, opened 2022-08-21
Update PradicBrown paper in mmil.raw.html

jkingdon:
Update references in iset.mm comments.

I checked the 2022 version of the paper and all the theorem numbering
and page numbers are still correct. The 2021 and 2022 versions are very
similar except for fixing three small errors (see the Acknowledgements
section of the latter where it refers to "mistaken type
annotations").

In mmil.html, refer to exmidfodomr one place which had not been updated
for exmidfodomr having been proved.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3930,9 +3930,8 @@ this implies excluded middle</TD>
 <TR>
   <TD>fodomb</TD>
   <TD><I>none</I></TD>
-  <TD>That the reverse direction implies excluded middle is
-  Proposition 1.2 of [PradicBrown2021], p. 2. The forward direction
-  is presumably also not provable.</TD>
+  <TD>That the reverse direction is equivalent to excluded middle is
+  ~ exmidfodomr . The forward direction is presumably also not provable.</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


avekens (approved):

mmil.raw.html
```diff
@@ -3930,9 +3930,8 @@
 <TR>
   <TD>fodomb</TD>
   <TD><I>none</I></TD>
-  <TD>That the reverse direction implies excluded middle is
-  Proposition 1.2 of [PradicBrown2021], p. 2. The forward direction
-  is presumably also not provable.</TD>
+  <TD>That the reverse direction is equivalent to excluded middle is
+  ~ exmidfodomr . The forward direction is presumably also not provable.</TD>
```
benjub, line 3934: "Presumably" ? Unless I'm missing something, this is one of the formulations of the axiom of choice ?

jkingdon: I guess so, although even in set.mm we just have a bit of text in https://us.metamath.org/mpeuni/fodom.html , not a proof that (anything which includes the token `-onto->`) implies `CHOICE`.

benjub: I was not precise enough above: a formulation of Choice is that every surjection splits, so the statement `fodom` may not be strong enough since the injection witnessing B ~< A (given a surjection F : A -->> B) is not required to be a section of F. In other words, the statement
```
A. x A. y A. f ( f : x -->> y -> E. g ( f o. g ) = ( Id |` y ) )
```
is equivalent to choice. Do you know if that equivalence is formalized in iset.mm or set.mm ? (I have not found it)

As for the mere `fodom` being equivalent to Choice, I would think so but am not sure.

PR #2784, opened 2022-08-21
new theorems for minimizations (no. 28, 30, 39, 54, 62, 77)

avekens:
New common theorems to shorten proofs, see issue #2710:
* new theorem **~mptfvmpt** created from @savask's list, no. 28: although its score may be lowered by line 1, it is still useful (49 proofs shortened, saving 777 bytes in total).
* new theorems **~fssdmd** and **~fssdm** (the latter created from @savask's list, no. 30): ~fssdmd is our standard deduction form, and is used at least to shorten the proof of ~fssdm (12 proofs shortened using ~fssdmd, 86 proofs shortened using ~fssdm, saving 1930 bytes in total).
* new theorem **~eqelssd** created from @savask's list, no. 39 (66 proofs shortened, saving 1756 bytes in total).
* new theorem **~animpimp2impd** created from @savask's list, no. 54 (15 proofs shortened, saving 854 bytes in total).
* new theorem **~reximssdv** created from @savask's list, no. 62 (19 proofs shortened, saving 899 bytes in total).
* new theorem **~rnfvprc** created from @savask's list, no. 77 (11 proofs shortened, saving 575 bytes in total).

The new theorems are provided in the first commit, the shortenings in separate subsequent commits.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -3021,6 +3021,7 @@
 "cba" is used by "hlcompl".
 "cba" is used by "hlex".
 "cba" is used by "htth".
+"cba" is used by "htthlem".
 "cba" is used by "imsmet".
 "cba" is used by "imsmetlem".
 "cba" is used by "isblo2".
@@ -7046,7 +7047,6 @@
 "hlex" is used by "axhilex-zf".
 "hlex" is used by "h2hcau".
 "hlex" is used by "h2hlm".
-"hlex" is used by "htthlem".
 "hlim0" is used by "hsn0elch".
 "hlimadd" is used by "chscllem4".
 "hlimcaui" is used by "chscllem2".
@@ -14455,7 +14455,7 @@ New usage of "c-bnj14" is discouraged (131 uses).
 New usage of "c-bnj18" is discouraged (58 uses).
 New usage of "cases2ALT" is discouraged (0 uses).
 New usage of "cayleyhamiltonALT" is discouraged (0 uses).
-New usage of "cba" is discouraged (85 uses).
+New usage of "cba" is discouraged (86 uses).
 New usage of "cbncms" is discouraged (5 uses).
 New usage of "cbvexsv" is discouraged (2 uses).
 New usage of "ccat2s1fvwALT" is discouraged (0 uses).
@@ -15989,7 +15989,7 @@ New usage of "hlcom" is discouraged (1 uses).
 New usage of "hlcompl" is discouraged (1 uses).
 New usage of "hldi" is discouraged (1 uses).
 New usage of "hldir" is discouraged (1 uses).
-New usage of "hlex" is discouraged (4 uses).
+New usage of "hlex" is discouraged (3 uses).
 New usage of "hlexch4N" is discouraged (0 uses).
 New usage of "hlhils1N" is discouraged (0 uses).
 New usage of "hlim0" is discouraged (1 uses).


set.mm
```diff
@@ -27523,6 +27532,19 @@ choice between (what we call) a "definitional form" where the shorter
       ( wrex cv wcel expr reximdva mpd ) ABDEHCDEHGABCDEADIEJBCFKLM $.
   $}
 
+  ${
+    $d x ph $.
+    reximssdv.1 $e |- ( ph -> E. x e. B ps ) $.
+    reximssdv.2 $e |- ( ( ph /\ ( x e. B /\ ps ) ) -> x e. A ) $.
+    reximssdv.3 $e |- ( ( ph /\ ( x e. B /\ ps ) ) -> ch ) $.
+    $( Derivation of a restricted existential qualification over a subset
```
benjub, line 27540: qual --> quant

set.mm
```diff
@@ -33738,6 +33760,16 @@ technically classes despite morally (and provably) being sets, like ` 1 `
...
   $}
 
+  ${
+    $d A x $.  $d B x $.  $d ph x $.
+    eqelssd.1 $e |- ( ph -> A C_ B ) $.
+    eqelssd.2 $e |- ( ( ph /\ x e. B ) -> x e. A ) $.
+    $( Equality deduction from subclass relationship and membership.
+       (Contributed by AV, 21-Aug-2022.) $)
+    eqelssd $p |- ( ph -> A = B ) $=
+      ( cv wcel ex ssrdv eqssd ) ACDEABDCABGZDHLCHFIJK $.
```
benjub, line 33770: When I see this statement and am told that it shortens many proofs, I'm not sure what to make of it.  This could be a sign that said proofs are not very clear: probably more readable proofs would have proved ph -> A C_ B and ph -> B C_ A and from these two subproofs, ph -> A = B. But if we shorten them using eqelssd, the deficiency of these proofs will be harder to notice... I had made such a remark on the discussion group, and I realize that I am not offering a solution here... What do others think (@digama0, @avekens, @savask, @jkingdon) ?

set.mm
```diff
@@ -56004,6 +56036,24 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
       ( wf wss fss syl2anc ) ABCEHCDIBDEHFGBCDEJK $.
   $}
 
+  ${
+    fssdmd.f $e |- ( ph -> F : A --> B ) $.
+    fssdmd.d $e |- ( ph -> D C_ dom F ) $.
+    $( Restricting the domain of a mapping, deduction form.  (Contributed by
```
benjub, line 56042: There is no restriction here (and in the next theorem). Maybe "Expressing that a class is a subclass of the domain of a function expressed in maps-to notation ?"

set.mm
```diff
@@ -57021,6 +57071,15 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
       CIFABJKLCABMCABNO $.
   $}
 
+  ${
+    rnfvprc.y $e |- Y = ( F ` X ) $.
+    $( The range of a function value for a proper class is empty.
```
benjub, line 57076: "of a function value at"
or maybe "of the value of a function at"

set.mm
```diff
@@ -60519,6 +60578,17 @@ in the range of the function (the implication "to the right" is always
   fex $p |- ( ( F : A --> B /\ A e. C ) -> F e. _V ) $=
     ( wf wfn wcel cvv ffn fnex sylan ) ABDEDAFACGDHGABDIACDJK $.
 
+  ${
+    $d A y $.  $d V x y $.  $d W y $.  $d Y y $.
+    mptfvmpt.y $e |- ( y = Y -> M = ( x e. V |-> A ) ) $.
+    mptfvmpt.g $e |- G = ( y e. W |-> M ) $.
+    mptfvmpt.v $e |- V = ( F ` X ) $.
+    $( A mapping as a the function value of a mapping.  (Contributed by AV,
```
benjub, line 60586: Maybe: "Value of a function-valued function" ?
(in any case: you have written "a the")

set.mm
wlammen: 1. eqelssd is a valid shortening according to the rules on the conventions page.  The new theorem adds 9 extra lines to set.mm, by contrast the shortenings in the main section only save already 13 lines.  The advantage is much greater, if one takes mathboxes into account.  So from a formal standpoint one can hardly argue against it.  Actually, eqssd is part of it's proof, it seems that the target expression is frequently constructed from ( ( ph /\ x e. B ) -> x e. A ) as a starting point.  
2. Your discomfort seems to be based on a sort of pedagogical aspect.  Some applications may hide, for example, symmetric aspects of the proof, that were otherwise better visible.  If one was to present the proof to an audience and pinpoint the essential steps, one might get lost when the proof is built from lots of adhoc theorems, not clearly focussing on a single thought.  I think this is a valid observation, and some visiting Metamath for the first time may even be repelled.  To some extent the concept of the Metamath language reminds me of Assembler in my profession: Extremely fast results on the machine, hard to grasp for a human being.  I like to use the term 'frogs-eye' perspective for this effect
3. If you really want to address the 'machine-focussed' style, a first step in my eyes is breaking long proofs down into lemmas, each focussing on a single aspect of the proof.  Here you have full control over the layout of a proof, and a wise choice could express the 'birds-eye view' properly. This could be a starting point for your intention.
4. When looking into how programming languages have evolved, I'd say, you need on top of the simple replacement scheme, that Metamath offers, an extension that allows you to express concepts.   But off the top of my head, I can't come up with something useful.  If this hadn't been solved already in some other proof program, a meta-Metamath language will take time and experience to develop.
5. First step in such a situation is (my experience) analysis of the current situation using statistics for example.  Even if the current rules are sometimes to the detriment of human readers, this might still be a neglectable portion of the whole database.

jkingdon: Since `( ph /\ x e. B ) -> x e. A` is just an `ex` and an `ssrdv` away from `B C_ A`, at least to my eyes `eqelssd` is a perfectly valid way of expressing the idea that showing the subclass relationship in both directions implies equality. In fact, it might be more clear than something with an explicit `B C_ A` if it avoids a few steps which are just housekeeping rather than statements about the classes in question.

So I don't see a conflict in this case between "shorter" and "more readable".

avekens: There are 66 proofs using this pattern (wherefore ~eqelssd is used for them now). Are all of these proofs badly designed? Maybe two or three of them should be analyzed in more detail, to find out why this pattern is used (and if it can be avoided). Without concrete examples, a judgement about ~eqelssd is difficult.

set.mm
avekens: Comments changed accordingly.

set.mm
avekens: Comment changed accordingly.

set.mm
avekens: Comment corrected/modified.

set.mm
avekens: I had a look at ~dvaddf which is proofed using eqelssd now:
```
67 37,39,41,42,44,45,46,56,65,66 dvaddbr  $p |- ( ( ph /\ x e. X )
                                          -> x ( S _D ( F oF + G ) ) ( ( ( S _D F ) ` x ) + ( ( S _D G ) ` x ) ) )
68 reldv         $p |- Rel ( S _D ( F oF + G ) )
69 68 releldmi   $p |- ( x ( S _D ( F oF + G ) ) ( ( ( S _D F ) ` x ) + ( ( S_D G ) ` x ) ) -> x e. dom ( S _D ( F oF + G ) ) )
70 67,69 syl     $p |- ( ( ph /\ x e. X ) -> x e. dom ( S _D ( F oF + G ) ) )
71 35,70 eqelssd  $p |- ( ph -> dom ( S _D ( F oF + G ) ) = X )
```

Here, it is ~releldmi which requires this pattern. and I see no easy/more elegant way to prove that ` X C_ dom ( S _D ( F oF + G ) ) ` in this case.

benjub: Ok. This is more of an open question about a general phenomenon where shortening with an ad-hoc lemma may sometimes hide more conceptual and larger shortenings.
In any case, this is non-blocking for a merge.

benjub: I see: it may happen that one inclusion is more naturally proved as A C_ B while the reverse inclusion is more naturally proved using a dummy variable.

benjub (approved):

PR #2787, opened 2022-08-22
proofs shortened by ~2fveq3

avekens:
355 proofs shortened by ~2fveq3, saving 5585 bytes.
Changes made by minimization script only.


benjub:
Note: ~2fveq3 was added in #2768 (and minimization was promised there).

benjub (approved):
Everything looks good after some random checks I did locally.

PR #2790, opened 2022-08-22
add bijust0

benjub:
Since the current proof ~bijust actually proves a more general instance, first state that instance (as ~bijust0) and then deduce ~bijust from it.  The compressed proof of the general instance is shorter (since fewer substitutions) and it is more readable (especially so in the present case: see the current http://us2.metamath.org/mpeuni/bijust.html).


avekens (approved):

wlammen:
I, too, think readability is improved.

PR #2791, opened 2022-08-23
Misc

benjub:
Everything in this PR is very minor. Only the first commit has some changes outside comments: using spimvw when possible reduces axiom usage.
(remark: the changes to "logical quotation" style are nothing like systematic: I only changed the ones I ran into.)


set.mm
```diff
@@ -610,7 +610,7 @@ that the result (using the rules described in ~ wi and ~ wn ) always
     $( Rule of Modus Ponens.  The postulated inference rule of propositional
        calculus.  See e.g.  Rule 1 of [Hamilton] p. 73.  The rule says, "if
        ` ph ` is true, and ` ph ` implies ` ps ` , then ` ps ` must also be
-       true."  This rule is sometimes called "detachment," since it detaches
+       true".  This rule is sometimes called "detachment," since it detaches
```
wlammen, line 613: Suggest to change a comment consistently to logical style.  Here 'detachment', 'modus ponendo ponens'  need transition, too.

wlammen (changes requested):

set.mm
benjub: You're right. Fixed. I took the opportunity to add one more commit to standardize more spellings. That's the last in this PR.

set.mm
```diff
@@ -652471,7 +652472,7 @@ not even needed (it can be any class).  (Contributed by Glauco
 
   ${
     $d A x y z $.  $d B x y z $.
-    $( An left open right closed interval doesn't contain its left endpoint.
+    $( An left-open right-closed interval does not contain its left endpoint.
```
wlammen, line 652475: Typo: A left-...

wlammen (changes requested):

wlammen:
The inclusion and omission of 'closed' in this PR preceding interval, sphere etc. seems to be correctly chosen from my POV.  But I am not an expert here.

wlammen (approved):
The question with a couple of 'an' is whether 'a' or 'any' is the best correction.  This is less of a problem in Main than in the mathboxes where people might have their own ideas.
Anyway, there are now fewer typos than before.

benjub:
@wlammen : thanks for the careful review, and sorry for the additional commit... one typo fix calls for another...
The "an + consonant" typos should be fixed now. (I do the changes after temporarily rewrapping with line width 9999 so that the search function of my text editor is not fooled by line wraps.)
I'm going to take a second look at the an-->a I just made to see if "any" was meant. (**edit:** checked: "any" was not meant in any of these cases, so all is good.)

wlammen:
My approval carries over to the latest commit.

PR #2793, opened 2022-08-24
Cross-link MPE/ILE Schroeder-Bernstein Theorem

david-a-wheeler:


benjub:
`$ metamath "r set.mm" "w s set.mm/r" "q"` should fix things.

benjub (approved):

david-a-wheeler:
[benjub](https://github.com/benjub) - ah yes. I even wrote the `scripts/rewrap` script.

I've rewrapped set.mm and iset.mm.

david-a-wheeler:
I've rewrapped, it passes all checks, it's been approved, it's easily reverted, it only affects comment text, and I think this is a pretty noncontroversial change anyway.

So while I usually wait longer, I'm going to just merge this now so I don't forget :-).

jkingdon:
I don't object, but I will offer one caveat. When we make links within iset.mm or within set.mm, we have tools to detect dead links (so we will remember to update when we rename theorems). Links like the ones here don't have that, so I have somewhat sparingly been using links between the two.

Having said that, I don't want to say we never can do so (none of the links in mm_100 have this kind of checking, for example). Just that I don't do it quite as freely as within set.mm or within iset mm

benjub:
@jkingdon : That's true indeed. Maybe we should think of a way of having such checks among a set of "official databases". The problem may come up more often if/when other databases are added and/or we want to increase modularity (e.g., a "core" database which serves as the base for both set.mm and iset.mm should be able to point to "later" theorems in these two databases).

Issue #2743, opened 2022-08-08
Finite intersections


benjub:
While editing the comment of http://us2.metamath.org/mpeuni/neiptopreu.html, I noticed that http://us2.metamath.org/mpeuni/df-fi.html defines nonempty finite intersections, whereas its comment, and the section title, say "Finite intersections".  (Side note: I think this is sloppy and wouldn't give a good image of our work if some outsider landed on that page and saw that.)  The quick fix is to modify the comment and section title, but I would suggest instead to include the empty intersection.

Of course, the abstract empty intersection is the universal class, which is a proper class so cannot be collected in a class.  But in the context of df-fi, we typically deal with intersections of families of subsets, so it is natural to make the definition
```
df-fi $a |- fi = ( x e. _V |-> { z | E. y e. ( ~P x i^i Fin ) z = ( |^| y i^i U. x ) } ) $.
```
This would entail modifying some proofs using it, so it would require some work, but it is probably worth it (the case of the empty intersection would no longer have to be treated separately).

Incidentally, this is what I had done in http://us2.metamath.org/mpeuni/bj-mooreset.html and http://us2.metamath.org/mpeuni/df-bj-moore.html (for arbitrary intersections) before noticing the issue of df-fi.

By the way: the theorem bj-mooreset, which is right before the definition df-bj-moore, serves as a kind of justification for it: it says that all classes sastisfying some predicate are actually sets, so can be collected in a class; in other words, the definition of the class does not "miss" any class satisfying the given predicate.  I don't know if there are other occurrences of this in set.mm.

**Edti**: I will do the minimal change of adding "nonempty" to the comment of df-fi, while we look at the possible implications of a definition change.  I haven't noticed problematic implications for now but the range of uses is vast...

digama0:
I disagree that the comment is imprecise. The definition clearly has no special case for the empty intersection, and it is indeed the collection of all finite intersections of the input set. The empty intersection doesn't exist so it is not included in that quantifier.

If you want to make a definition for the class of finite intersections intersected with some set, you are welcome to but that would be a different definition.

benjub:
> it is indeed the collection of all finite intersections of the input set.

This is at best misleading. set.mm talks of the empty intersection (e.g., ~int0) even if it is a proper class so is a "non-exsitent object" to use terms from free logic.

[...]

PR #2795, opened 2022-08-25
revise comment of cfi and df-fi

benjub:
Minimal comment edit of df-fi and cfi to clarify the case of finite intersections.
Resolves Issue #2743 in a minimal fashion.


set.mm
```diff
@@ -84691,16 +84691,20 @@ of Infinity (see comments in ~ fin2inf ).  (Contributed by NM,
...
 
   ${
     $d x y z $.
-    $( Function whose value is the class of all the finite intersections of the
-       elements of ` x ` .  (Contributed by FL, 27-Apr-2008.) $)
-    df-fi $a |- fi = ( x e. _V |-> { z |
-                  E. y e. ( ~P x i^i Fin ) z = |^| y } ) $.
+    $( Function whose value is the class of finite intersections of the
+       elements of the argument.  Note that the empty intersection being the
+       universal class, hence a proper class, it cannot be an element of that
```
wlammen, line 84702: typo: ...class, cannot...

wlammen (dismissed):

set.mm
benjub: I don't see a typo: I would repeat the subject since there is no comma:
> Jack being tall, he is not short.
> Jack, being tall, is not short.

We may have to ask a native speaker, e.g. @jkingdon

wlammen (commented):
Don't want to insist on my request.

avekens (approved):

PR #2799, opened 2022-08-26
#2128 one instance of equvelv

wlammen:
This PR emulates a move of equvelv from Andrew Salmon's mathbox to Main about 2 years ago.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15243,7 +15243,6 @@ New usage of "distrnq" is discouraged (7 uses).
 New usage of "distrpi" is discouraged (5 uses).
 New usage of "distrpr" is discouraged (7 uses).
 New usage of "distrsr" is discouraged (3 uses).
-New usage of "divalgmodOLD" is discouraged (0 uses).
 New usage of "djaclN" is discouraged (0 uses).
 New usage of "djaffvalN" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


avekens (approved):

avekens:
Since I have the permissions now, I want to merge this (relatively small) PR. I wonder, however, which is the merge policy we agreed on (see discussion in #2517): "Squash and merge", or "Rebase and merge". As I understand correctly, Mario was in favour of the first, but David of the second. @wlammen @benjub how do you merge? Can we see after a PR is merged which option was used?

wlammen:
Squash and merge.  The individual commits are eliminated (2 in this case), and just 1 entry remains.  In git log the commit messages still appear with an * in front of them.  I don't see a reason why someone would checkout a state in between my 2 commits.

avekens:
> Squash and merge. The individual commits are eliminated (2 in this case), and just 1 entry remains. In git log the commit messages appear with an * in front of them.

Thanks a lot, Wolf. Now I try to "squash and merge"...

wlammen:
Just to add this aspect: I think (but have not confirmed) that a squashed merge is more space efficient, something that becomes more and more important over time.  Downloading the whole repository takes already quite some time.

PR #2800, opened 2022-08-26
Alternate definition of iota

avekens:
As proposed by BJ via email on 13-Aug-2022, an alternate definition of iota (~df-aiota) is introduced (in AV's mathbox) which is the universe V if it is not defined (in contrast to the current definition ~df-iota, which is the empty set in this case). By this definition, the alternate definition of a function value (~df-afv) can be adapted accordingly (in a next PR).

Changes in detail, main set.mm:
* ~elv moved from PM's mathbox (TODO: check if additional proofs can be shortened)
* ~dfiota2lem extracted from proof of ~dfiota2
* new symbol "iota'"

Changes in detail, AV's mathbox:
* new section "General auxiliary theorems (1)" with auxiliary theorems created to be used for the alternate iota (some theorems of section "General auxiliary theorems (2)" are also moved here)
* new auxiliary theorems ~al0ssb, ~eusnsn, ~absnsb, ~euabsn22
* new section "Definite description binder (inverted iota) - extension" with two examples for the current iota.
* new section "Alternative for Russell's definition of a description binder" with the definition of the alternate iota and basic theorems for it.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,7 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
-26-Aug-22 elv       syl1111anc  moved from PM's mathbox to main
+26-Aug-22 elv                   moved from PM's mathbox to main
 18-Aug-22 eel1111   syl1111anc  moved from AS's mathbox to main
 26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main
 26-Jul-22 wl-jarli  jarli       moved from WL's mathbox to main
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+26-Aug-22 elv       syl1111anc  moved from PM's mathbox to main
 18-Aug-22 eel1111   syl1111anc  moved from AS's mathbox to main
 26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main
 26-Jul-22 wl-jarli  jarli       moved from WL's mathbox to main


avekens:
Remark: 301 proofs can be shortened using ~elv!

set.mm
```diff
@@ -54198,13 +54208,20 @@ Definite description binder (inverted iota)
...
 
   ${
-    $d y x $.  $d y ph $.
+    $d x y $.
+    $( Lemma for ~ dfiota2 , ~ dfaiota2 and ~ aiotaval .  Formerly part of
+       proof of ~ dfiota2 .  (Contributed by Andrew Salmon, 30-Jun-2011.)
+       (Revised by AV, 24-Aug-2022.) $)
+    dfiota2lem $p |- ( { x | ph } = { y } <-> A. x ( ph <-> x = y ) ) $=
+      ( cab cv csn wceq weq wb wal df-sn eqeq2i abbi bitr4i ) ABDZCEZFZGOBCHZBD
+      ZGARIBJQSOBPKLARBMN $.
```
benjub, line 54217: I would replace `y` with `A`. I think this is more than a lemma. It could be moved earlier with a comment like
> Condition for a class abstraction to be a singleton.

as the label you see fit.

set.mm
```diff
@@ -29221,6 +29221,16 @@ choice between (what we call) a "definitional form" where the shorter
...
 
+  ${
+    elv.1 $e |- ( x e. _V -> ph ) $.
+    $( New way to shorten some proofs.  Inference forms (with ` |- A e. _V `
+       hypotheses) of the general theorems (proving ` |- ( A e. V -> `
+       assertions) may be superfluous.  (Contributed by Peter Mazsa,
+       13-Oct-2018.) $)
+    elv $p |- ph $=
+      ( cv cvv wcel vex ax-mp ) BDEFABGCH $.
+  $}
```
benjub, line 29232: Not sure it shortens sufficiently many proofs to be moved to Main... But I guess you haven't tried the scripts/min.cmd yet?
If it is to stay, then the comment should be adapted. Maybe:
> Technical lemma used to shorten proofs.  If a proposition is implied by ` x e. _V ` (which is true, see ~ vex ), then it is true.

set.mm
```diff
@@ -704635,6 +704646,264 @@ exists a proof for (a and b).  (Contributed by Jarvin Udandy,
...
+
+  ${
+    $d X y $.
+    $( The empty set is the one and only set which is a subset of every set.
+       (Contributed by AV, 24-Aug-2022.) $)
+    al0ssb $p |- ( A. y X C_ y <-> X = (/) ) $=
+      ( cv wss wal c0 wceq 0ex sseq2 ss0b syl6bb 0ss ax-gen sseq1 albidv mpbiri
+      spcv impbii ) BACZDZAEZBFGZTUBAFHSFGTBFDUBSFBIBJKQUBUAFSDZAEUCASLMUBTUCAB
+      FSNOPR $.
+  $}
```
benjub, line 704669: It looks like this could be moved to Main (now or in a later PR).

set.mm
```diff
@@ -704635,6 +704646,264 @@ exists a proof for (a and b).  (Contributed by Jarvin Udandy,
...
+    eusnsn $p |- E! x { x } = { y } $=
+      ( vz csn wceq weu weq wal wex equequ2 bibi2d albidv cvv sneqbg elv ax-gen
+      cv wb spei df-eu mpbir ) AQZDBQZDEZAFUDACGZRZAHZCIUGUDABGZRZAHCBCBGZUFUIA
+      UJUEUHUDCBAJKLUIAUIAUBUCMNOPSUDACTUA $.
+  $}
+
+  ${
+    $d x y $.
+    $( If the class abstraction ` { x | ph } = { y } ` of a wwf ` ph ` is a
+       singleton, the wwf is true for the singleton element ` y ` .
```
benjub, line 704709: wwf --> wff (twice).  And I would write: either
> the class abstraction { x | ph } associated with the wff ph

or
> the extension { x | ph } of the wff ph

set.mm
```diff
@@ -704635,6 +704646,264 @@ exists a proof for (a and b).  (Contributed by Jarvin Udandy,
...
+    absnsb $p |- ( { x | ph } = { y } -> [ y / x ] ph ) $=
+      ( cab cv csn wceq weq wi wal wsb wcel nfab1 nfcv cleqf abid velsn bibi12i
+      wb sylbi biimpr alimi sb2 syl ) ABDZCEZFZGZBCHZAIZBJZABCKUHBEZUELZULUGLZS
+      ZBJUKBUEUGABMBUGNOUOUJBUOAUISUJUMAUNUIABPBUFQRAUIUATUBTABCUCUD $.
+  $}
+
+  ${
+    $d x y z $.  $d ph y z $.
+    $( Another way to express existential uniqueness of a wff ` ph ` : its
+       class abstraction ` { x | ph } = { y } ` is a singleton.  Variant of
```
benjub, line 704720: idem:
> its extension { x | ph }

or
> its associated class abstraction { x | ph }

set.mm
avekens: OK, I will try - but it is a little bit more complicted, because ` A ` might be a proper class...

set.mm
avekens: I primarily moved it because I use it in my mathbox. But as I wrote in my comment, 301 proofs can be shortened using ~elv! I expect a saving of about 3000 bytes. But I will change the comment as you propose.

set.mm
avekens: Do you think we can shorten sufficiently many proofs by it, or only because it is interesting?

set.mm
avekens: OK

set.mm
avekens: OK

david-a-wheeler:
Maybe it's just me, but this alternate definition of iota seems really weird to me, and I missed the email proposal.

Is there any math literature we can point to recommending this or discussing this kind of thing?

avekens:
> Maybe it's just me, but this alternate definition of iota seems really weird to me, and I missed the email proposal.

This was a personal email to me, and I think there was not much more information about this alternate definition than I mention in the comment od df-aiota. 

> Is there any math literature we can point to recommending this or discussing this kind of thing?

Why do you think it is weird, more weird than the original definition df-iota? It is nearly the same, only the "undefined" case is different. And to use the empty set for "undefined" is a specialty of set.mm, which you won't find in literature, will you?

set.mm
avekens: Done, it was easier than I expected (no case discrimination necessary)

set.mm
avekens: Comment changed, minimization, as announced, in a next PR.

set.mm

set.mm
avekens: Done

set.mm
benjub: Only on its own interest. But this can be done (if you wish so) in a later PR.

benjub (approved):

benjub:
> > Is there any math literature we can point to recommending this or discussing this kind of thing?
> 
> Why do you think it is weird, more weird than the original definition df-iota? It is nearly the same, only the "undefined" case is different. And to use the empty set for "undefined" is a specialty of set.mm, which you won't find in literature, will you?

I agree wit @avekens. Both df-iota and df-aiota have the same level of "weirdness", but it is necessary if you think about it for a while. Note that when there is a unique $x$ (call it x_0) such that $\varphi(x)$ holds, then the definiens (RHS) is $\bigcup${x_0} (resp. $\bigcap${x_0}), which is simply $x_0$ as wanted.

set.mm
avekens: OK, I have moved it to main.

benjub (approved):

wlammen:
This is beyond what I am familiar with, so you hopefully don't expect an opinion from me.

benjub:
> This is beyond what I am familiar with, so you hopefully don't expect an opinion from me.

Ok, merging. I was waiting in case @david-a-wheeler had something to add but I hope that Alexander's answer and mine did explain the matter.

As for your familiarity with set theory, I hope you're working on it !  Part 2 (ZF) could benefit a lot from the kind of work you did in Part 1 (FOL).

david-a-wheeler:
This is merged and I'm *okay* with it being merged as an alternative definition.

However, I worry about making other things equivalent to the universe when they don't need to be. I'm not sure I can put my concern into words, and they may be silly, but here goes.

As I'm sure you know, in ZFC, the universe `_V` is the class of all sets and is *not* itself a set. So:

* How `_V` is handled varies between axiomatic systems (e.g., New Foundations). The empty set seems more "stable" (okay, Aristotelian logic doesn't accept it, but I don't think we worry about that case). I'd prefer to use similar or the same definitions across axiomatic systems when that's reasonable to do.
* I also worry that making other things equivalent `_V` gets dangerously close to some sort of paradox. I don't think it *does*, but I conservatively avoid such constructs where it's reasonable to do so. If nothing else, avoiding them means they're more portable across axiomatic systems & philosophically easier to justify.

These thoughts are probably unclear, and they certainly aren't well-stated. They may even be absurd... but I thought I'd try to capture my thinking anyway.

benjub:
@david-a-wheeler : that's precisely because _V "does not exist" and is out of the domain of quantification (and, in pure ZFC, out of the domain of discourse), that it is a good candidate for an undefined value. As for the risks of clashes, they look greater when making an undefined value equal to an actual set. With mmj's definition checker, there is no risk of paradox with the formal parts of course, but the risks, as always with formal provers, lie in the translation back and forth between formal and natural language (i.e., "what does this statement really prove", for instance, with our current conventions, we could prove "the logarithm of zero is the only set which contains every set", since both are encoded as the empty set... even more: both exist !).

david-a-wheeler:
Ooooooh. Okay. Now I see. That's kind of neat.

There *is* a weirdness with this definition, though; any set would be a member of the result if df-iota isn't unique.

It might be better if the result was a proper class when there wasn't a single result. Then no set would be either equal to *or* a subset of the resulting class.

david-a-wheeler:
Er, sorry, I don't mean proper class. `_V` is already a proper class. Hmmm... what *do* I mean?

I guess I want a class that contains only-non-sets. That way, neither equality not is-member-of will "work" on sets with the result.

My head hurts. I don't even know if we can express such a construct in set.mm, never mind if it'd be worth doing. I'm honestly leaning on my Computer Science background, where it's sometimes useful to be able to return a value that distinguishes between "normal" values and "error" values (e.g., Haskell's Maybe).

benjub:
> I guess I want a class that contains only-non-sets.

The only class that contains only nonsets is the empty set, since in our set theory there are no urelements and a class can only contain sets.

Encoding undefined values is a recurring issue in formal systems based on set theory.  Haskell is based on type theory, where there are no such problems.

david-a-wheeler:
That makes sense, and it looks like there is no better answer.

Thanks for the reply!

avekens:
Several weeks ago, I made an attempt to use the empty set as representative of an "unfdefined" value, but this was too restrictive (the empty set is often a valid function value). 

If the range of a function is known (as in ` F : A --> B ` ) then we can use the already existing definition ` Undef ` which is related to a given set (which could be the range B) and which results in a value not contained in the given set B: `` -. ( Undef ` B ) e. B ``, see ~undefnel).  But this does not hold for prope classes B, and cannot be applied to functions without an explicitly given range, e. g. ` Fun F ` and ` F Fn A `.  Maybe it would be interesting how far we can get with defining `` ( F ` X ) =  ( Undef ` B ) `` instead of `` ( F ` X ) =  (/) `` if ` F : A --> B ` and ` -. X e. A `. Maybe I will make some experiments with this in the future...

david-a-wheeler:
@avekens - that sounds interesting! I'm curious to see how that works out. Good luck.

PR #2802, opened 2022-08-26
deprecate issref in favor of idref

benjub:
Deprecate issref in favor of idref; duplicate from #2128.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16379,6 +16379,7 @@ New usage of "issh" is discouraged (3 uses).
 New usage of "issh2" is discouraged (10 uses).
 New usage of "issh3" is discouraged (1 uses).
 New usage of "issmgrpOLD" is discouraged (3 uses).
+New usage of "issref" is discouraged (0 uses).
 New usage of "isssp" is discouraged (8 uses).
 New usage of "isst" is discouraged (4 uses).
 New usage of "issubgoilem" is discouraged (0 uses).
@@ -19522,7 +19523,6 @@ Proof modification of "idiVD" is discouraged (6 steps).
 Proof modification of "idn1" is discouraged (5 steps).
 Proof modification of "idn2" is discouraged (7 steps).
 Proof modification of "idn3" is discouraged (15 steps).
-Proof modification of "idref" is discouraged (92 steps).
 Proof modification of "iedgvalOLD" is discouraged (77 steps).
 Proof modification of "iidn3" is discouraged (8 steps).
 Proof modification of "iin1" is discouraged (1 steps).
@@ -19558,6 +19558,7 @@ Proof modification of "isosctrlem1ALT" is discouraged (363 steps).
 Proof modification of "isrnsigaOLD" is discouraged (202 steps).
 Proof modification of "issgrpALT" is discouraged (53 steps).
 Proof modification of "issmgrpOLD" is discouraged (85 steps).
+Proof modification of "issref" is discouraged (347 steps).
 Proof modification of "istrkg2d" is discouraged (439 steps).
 Proof modification of "isuvtxaOLD" is discouraged (154 steps).
 Proof modification of "isvcOLD" is discouraged (170 steps).
--- a/discouraged
+++ b/discouraged
@@ -16379,7 +16379,7 @@ New usage of "issh" is discouraged (3 uses).
 New usage of "issh2" is discouraged (10 uses).
 New usage of "issh3" is discouraged (1 uses).
 New usage of "issmgrpOLD" is discouraged (3 uses).
-New usage of "issref" is discouraged (0 uses).
+New usage of "issrefOLD" is discouraged (0 uses).
 New usage of "isssp" is discouraged (8 uses).
 New usage of "isst" is discouraged (4 uses).
 New usage of "issubgoilem" is discouraged (0 uses).
@@ -19558,7 +19558,7 @@ Proof modification of "isosctrlem1ALT" is discouraged (363 steps).
 Proof modification of "isrnsigaOLD" is discouraged (202 steps).
 Proof modification of "issgrpALT" is discouraged (53 steps).
 Proof modification of "issmgrpOLD" is discouraged (85 steps).
-Proof modification of "issref" is discouraged (347 steps).
+Proof modification of "issrefOLD" is discouraged (347 steps).
 Proof modification of "istrkg2d" is discouraged (439 steps).
 Proof modification of "isuvtxaOLD" is discouraged (154 steps).
 Proof modification of "isvcOLD" is discouraged (170 steps).
--- a/discouraged
+++ b/discouraged
@@ -8193,6 +8193,11 @@
 "idn3" is used by "tratrbVD".
 "idn3" is used by "trintALTVD".
 "idn3" is used by "truniALTVD".
+"idrefALT" is used by "elrefsymrels3".
+"idrefALT" is used by "idinxpssinxp2".
+"idrefALT" is used by "idinxpssinxp3".
+"idrefALT" is used by "refsymrels3".
+"idrefALT" is used by "symrefref3".
 "idrval" is used by "cmpidelt".
 "idrval" is used by "iorlid".
 "idunop" is used by "idlnop".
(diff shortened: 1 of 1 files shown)


wlammen (approved):

avekens:
Unfortunately, I missed the discussion about this issue in #2128. I think it is quite interesting that this theorem ~issref can be proven without all the definitions and theorems about functions, and it "fits" into the context: the following theorems are about symmetry (~cnvsym), antisymmetry (~intasym), etc. Therefore, I would like to keep this theorem and call it idrefALT (with a reference to ~idref of course). 

Furthermore, I would like to move up ~idref several lines, immediately after ~fnasrn which is used in the proof. I see no reason for the current location (the theorems before and after it are completely different), and the theorems following ~fnasrn are also of different kind. Then, this theorem is closer to ~issref (resp. ~isrefALT).

avekens (changes requested):
see my previous comment

avekens:
BTW: Does anybody has contact with Frédéric Liné? He is the contributor of the discussed theorems, and he contributed a lot to set.mm several years ago. Unfortunately, he quit the metamath community a few years ago (I think it was in anger), but he was in contact via email with Norm. His last sign of life is in the google group: https://groups.google.com/g/metamath/c/_kYIflam93o/m/SYjl3MvRAgAJ.

wlammen:
It is always possible to strip all definitions but those required by used axioms from a theorem, and so place it anywhere after the last referenced axiom.  Doing so comes with a price: The proofs get longer than necessary, and are usually more difficult to understand.  That's what definitions are for: keep the proofs and theorems concise.

Good reasons for not using available definitions could be, for example:
1. The theorem is of use to justify a definition, or simplify proofs closely related to that definition.
2. Dragging in a theory in some unrelated topic disturbs a suggesting layering or even creates a cyclic reference forcing a somewhat interleaved presentation of concepts.

Is this the case here?  In what way?

If you point to ~intirr, I think ( 𝑅 ⊆ I ) ↔ ∀𝑥 𝑥𝑅𝑥 is the correct companion, not ~issref.  The restriction needs the cross product, and is itself a higher level construct, seen from the point of reflexivity alone.

A related problem: It is known that some theorems about reals have surprisingly short proofs in the complex plane.  Should we not make use of these shortcuts, because we use a concept extending reals, much as the function concept is used here?

benjub:
@wlammen summed up pretty well the situation with definitions. So this is not an argument to keep issref.

@avekens mentions the related theorems cnvsym, intasym... They may have some shorter proofs, similarly as idref versus issref ? To me, their existence is not an argument fo  keep issref. So, if @avekens insists, I'll relabel it idrefALT, but given the other comments on the related Issue and here, I'm inclined to keep it as issrefOLD.

As @wlammen noticed, these theorems are not relative to a given class like idref is. The relative versions (e.g., "is symmetric on A" versus "is symmetric"), are typically more useful, so these would be nice additions (if they are not already in set.mm).

Moved idref up as @avekens suggested, and edited its comment to say it's relative to a class.

avekens:
Although all given arguments are convincing, I still cannot accept to delete ~issref completely. As I read _".. both in main by FL. I propose to deprecate ~issref (which is only used by mathbox theorems) ..."_ in the discussion of issue #2128, I thought it was FL's mathbox. But it is PM's mathbox which is concerned (FL's mathbox was deleted on his request...)! And I think the whole mathbox does not depend on the definitions of functions (Fun, Fn, F : A --> B, etc.). Currently, the material of his mathbox could be inserted after section "Relations" in main set.mm, which would not be possible anymore if ~issref was replaced by ~idref. So at least, we should ask @mazsa if he would agree with this replacement. Otherwise, ~issref should be replaced by ~idrefALT in his mathbox. If necessary, ~idrefALT could be moved into his mathbox, but I would prefer to leave ~idrefALT at its/~issref's current place.

avekens:
> Moved idref up as @avekens suggested, and edited its comment to say it's relative to a class.

Thanks a lot (Thumb up!)

mazsa:
@avekens thank you for the mention. I completely agree with your suggestion to leave ~issref (perhaps renamed preferably to ~idref) at its current place: it seems to me a more conservative and bottom up approch to speak about reflexivity under Relations (not under Functions) even if its proof is longer. In addition to this general argument, I have a special reason as well: some of my theorems use it before my Functions, so if we think there is a chance to upgade them into the main part of set.mm, I´ll need it in Relations.

benjub:
Ok, so we keep both proofs. There may indeed be some interest in avoiding the function-based definitions, especially since the proof of idref relies on the particular encoding of functions (but this is true at some degree of most definitions) via ~fnasrn. On the other hand, issref was way too long. I added a much shortened proof (even when inlining ~elrid, which is interesting in itself, it is much shorter than the current one).

mazsa:
Thank you @benjub

avekens (approved):
Everything is fine now.

avekens:
I found an example where an original theorem/proof is lost because of a minimization (in a similar way as discussed in this PR):

* ~pwuninel2 - comment: "Direct proof of ~ pwuninel avoiding functions and thus several ZF axioms."
* but the proof of ~pwuninel was shortened using ~pwuninel2 meanwhile, and therefore does not depend on functions and thus several ZF axioms (anymore)! So either the original proof should be reconstructed (and provided as ALT), or the comment of ~pwuninel2 has to be corrected...

benjub:
> I found an example where an original theorem/proof is lost because of a minimization (in a similar way as discussed in this PR):
> 
>     * ~pwuninel2 - comment: "Direct proof of ~ pwuninel avoiding functions and thus several ZF axioms."
> 
>     * but the proof of ~pwuninel was shortened using ~pwuninel2 meanwhile, and therefore does not depend on functions and thus several ZF axioms (anymore)! So either the original proof should be reconstructed (and provided as ALT), or the comment of ~pwuninel2 has to be corrected...

Not sure what this has to do with this PR. pwuninel and pwuninel2 are different statements, so the comment of the pwuninel2 should be changed, and its label too. Maybe label it ~pwuninellem and comment "Lemma for ~pwuninel. Note that it uses fewer axioms than ~pwuninel." Also, in the comment of pwuninel, I would change all occurrences of "set" to "class".

**edit**: another possibility is to simply inline the proof of pwuninel2 and remove it.

mazsa:
For the same reason, I would need to change the proof of ~idssxp to something more basic like
```
50::idrefALT       |- ( ( _I |` A ) C_ ( A X. A ) <-> A. x e. A x ( A X. A ) x )
51::brxp           |- ( x ( A X. A ) x <-> ( x e. A /\ x e. A ) )
52::pm4.24         |- ( x e. A <-> ( x e. A /\ x e. A ) )
53:51,52:bitr4i    |- ( x ( A X. A ) x <-> x e. A )
54:53:ralbii       |- ( A. x e. A x ( A X. A ) x <-> A. x e. A x e. A )
55::ralel          |- A. x e. A x e. A
56:55,54:mpbir     |- A. x e. A x ( A X. A ) x
qed:56,50:mpbir    |- ( _I |` A ) C_ ( A X. A )
$d A x
$=  ( vx cid cres cxp wss cv wbr wral wcel ralel pm4.24 bitr4i ralbii
    wa brxp mpbir idrefALT ) CADAAEZFBGZTSHZBAIZUBTAJZBAIBAKUAUCBAUAU
    CUCOUCTTAAPUCLMNQBASRQ $.
```
(this proof is not longer than the present one) and to put ~idssxp directly after ~idrefALT . Or do you rather prefer to create there a new ~idssxpALT ? @avekens @benjub @tirix @wlammen

benjub:
As a first step, put it as idssxpALT in your mathbox, so that it is easier for others to compare the proofs. Then, we can see which proof we keep, or if we keep both, and where.

benjub:
@mazsa : better: use `elrid` to prove
```
$( Equivalent expression for the restricted identity relation.  (Contributed by PM, 8-Sep-2022.) $)
ridxp $p |- ( _I |` A ) = ( _I i^i ( A X. A ) ) $= ? $. 
```
and prove `idssxp` from `ridxp`. Put both `ridxp` and `idssxp` right after `elrid` and relabel the current `idssxp` to `idssxpOLD`.

**edit**: you'll probably need to add `elid $p |- ( A e. _I <-> E. x A = <. x , x >. ) $= ? $.` (which is an application of elrid with X=_V).

PR #2805, opened 2022-08-31
Intuitionize seqf1o

### LINKED ISSUE #2575: Intuitionize seqf1o

jkingdon:
The statement of the theorem iseqf1o is only slightly different from https://us.metamath.org/mpeuni/seqf1o.html and differs mostly in the usual ways related to where functions need to be defined. The proof is much longer and split across a large number of lemmas. Probably there is a shorter proof. But this one does work and I'm not sure small tweaks to the set.mm proof would suffice given differences around where functions are defined and how `seq` works and the like.

Also in this pull request:
* `isummolem3` which had been waiting on an intuitionized seqf1o .
* Add to the missing theorems list: `fofinf1o`
* ~~`f1ofni` which is a variation of https://us.metamath.org/mpeuni/f1oi.html~~ _Removed this: not used and to be seen whether we'd rather express this idea another way_
* `fzodcel` which is similar to https://us.metamath.org/ileuni/fzdcel.html but for `..^` rather than `...`
* continuing the process of removing `S e. V` hypotheses from `seq` theorems: `iseqshft2` , `iseqsplit` (and according changes to where they are used)
* copy `nnsplit` from set.mm
* the one change to set.mm is merely to run the minimzer on one theorem

Fixes #2575

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6363,10 +6363,11 @@ or ~ ssfiexmid </TD>
 
 <TR>
   <TD>seqf1o</TD>
-  <TD><I>none</I></TD>
-  <TD>May be adaptable with sufficient attention to issues such as
-  whether the operation ` .+ ` is closed with respect to ` C ` as
-  well as ` S ` , and perhaps other issues.</TD>
+  <TD>~ iseqf1o</TD>
+  <TD>The functions ` G ` and ` H ` need to be defined on ` ( ZZ>= `` M ) `
+  not merely ` ( M ... N ) ` as in seqf1o . Also, a single set ` S `
+  takes the place of ` C ` and ` S ` because that is sufficient
+  flexibility at least for now.</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3414,6 +3414,12 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>fofinf1o</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof uses excluded middle in several places.</TD>
+</TR>
+
+<TR>
   <TD>dmfi</TD>
   <TD>~ fundmfi</TD>
 </TR>


iset.mm
```diff
@@ -49989,6 +49989,21 @@ is the image of an element of its domain (see ~ foima ).  (Contributed
       $.
   $}
 
+  ${
+    $d A x y $.  $d F x y $.  $d ph y $.
+    f1ofni.f $e |- ( ph -> ( F |` A ) Fn A ) $.
+    f1ofni.i $e |- ( ph -> A. x e. A ( F ` x ) = x ) $.
+    $( A function which maps sets to themselves is one-to-one and onto.
+       (Contributed by Jim Kingdon, 25-Aug-2022.) $)
+    f1ofni $p |- ( ph -> ( F |` A ) : A -1-1-onto-> A ) $=
```
benjub, line 49998: It looks like the statement is trying hard to hide the fact that F restricted to A is the identity.  It would be more natural to have the conclusion ``( F |` A ) = ( _I |` A ) ``.  If you need the precise statement of your conclusion, there is probably a theorem somewhere `` ( _I |` A ) : A -1-1-oneto-> A`` from which you conclude what you need.

benjub:
As a rule for breaking lines within formulas: if you can avoid one line break without creating another one, then do it (i.e., if two consecutive lines fit on a single line, use only one line).

iset.mm
```diff
@@ -99583,6 +99612,769 @@ seq K ( .+ , G , S ) ) $=
...
+          |-> if ( x <_ N , ( G ` ( f ` x ) ) , ( G ` M ) ) ) $.
+
+        ${
+          $d .+ x y z $.  $d G f x $.  $d J f v x $.  $d J u v x $.
+          $d J f x y z $.  $d K u v x $.  $d K x y z $.  $d M f x $.
+          $d M u x $.  $d N f x $.  $d N u x $.  $d P v x $.  $d P x y z $.
+          $d Q f v x $.  $d Q f x y z $.  $d S x y z $.  $d ph u v x $.
+          $d ph x y z $.
+          $( Lemma for ~ iseqf1o . ` Q ` gives the same sum as ` J ` in the
+             range ` ( K ... ( ` ' J ` K ) ) ` .  (Contributed by Jim Kingdon,
```
benjub, line 100017: Not sure if there should be a space between backtick and single quote.

iset.mm
```diff
@@ -99583,6 +99612,769 @@ seq K ( .+ , G , S ) ) $=
...
+            UVOXPUUHYKVEXNXGAUUHMUKZLUKNLUKZUUMUURAVUKNLAUVMUVGVUKNWCUDUCUVFUVF
+            NMYSVPXNABUUHFIMJLNOPUCUDUVNUBUHXSAUURNHUKZLUKZVULABNFIHJLNOPUCUYFU
+            CUBUHXSZAVUMNLAVUMNUWIUSZNNWCZNNULUMUNMUKZWDZNMUKZWDVUSNAENHMNOPUCU
+            DUCUGXEAVUPVUSVUTAUUHUVEUSZVUPAUVBUVJNUUHUTVAVVAUVIUVOABMNOPTUCUDUE
+            YINUUHVRVSNUUHYLVEXFAVUQNVURANYMXFXOZXNXGYNYOABCGIUUJNUUHUVIAVUGUVJ
+            UUOUUHUTVAZUUHVUBUSANUVIWJUVOAUVPVVCUVQAUVBUVJUVPVVCVOUVIUVONUUHYPV
+            PVQUUOUUHVRVSZVUAQYQAUUTUURUUQGUNUUSABCDGIUUNNUUHQSVVDUVIVUIYRABCUU
+            RUUQIGRAUURVUNIVUOAUYKVUNIUSBUYHVUMUYGVUMWCUYJVUNIUYGVUMLUUAUUBAUYK
+            BUYHUBUUCAVUMNUYHVVBUYSYTUUDYTABCGIUUNUUOUUHVVDVUJQUUEUUFXGYN $.
+        $}
```
benjub, line 100071: Reviewed up to here.

iset.mm
jkingdon: > It looks like the statement is trying hard to hide the fact that F restricted to A is the identity. It would be more natural to have the conclusion `` ( F |` A ) = ( _I |` A ) ``. If you need the precise statement of your conclusion, there is probably a theorem somewhere `` ( _I |` A ) : A -1-1-oneto-> A `` from which you conclude what you need.

I'll just delete this theorem. I didn't end up using it.

iset.mm
jkingdon: The backtick inside backticks should be quoted. That is the original problem here which probably is what led to that extraneous space (maybe from a rewrap).

jkingdon:
> As a rule for breaking lines within formulas: if you can avoid one line break without creating another one, then do it (i.e., if two consecutive lines fit on a single line, use only one line).

Is this in reference to a particular line? I mostly do what you are suggesting, but Norm always used to say things like "how would you prefer to break the lines?" so I guess I was supposed to apply some level of eyeball test.

iset.mm
jkingdon: Now fixed. As the metamath book says "Two consecutive grave accents ‘‘ are treated as a single actual grave
accent (both inside and outside of math mode) and will not cause the output
processor to enter or exit math mode." so the grave accents need to be doubled both in the converse and the function application.

iset.mm
```diff
@@ -94723,6 +94723,16 @@ Finite intervals of nonnegative integers (or "finite sets of sequential
     ZKIZUHEFZLZDAKIZUKLBUGMUFQNUFUHUGCUGULMUFUHBUGAOQPDUHRSUFUJUMUKUFUIADKUFATC
     UIAMAUAAUBSUCUDUE $.
 
+  $( Express the set of positive integers as the disjoint union of the first
+     ` N ` values and the rest.  (Contributed by Glauco Siliprandi,
+     21-Nov-2020.) $)
+  nnsplit $p |- ( N e. NN -> NN = ( ( 1 ... N ) u. ( ZZ>= ` ( N + 1 ) ) ) )
+          $=
```
benjub, line 94730: fits on single line

iset.mm
```diff
@@ -99583,6 +99597,769 @@ seq K ( .+ , G , S ) ) $=
...
+      AZWFAXGXLYBXIAWMXLGXMOZBDVPVDVQVRAWJWKLWFUEAWKWKBCWRIVOVQAXGXLWEPLZWFWAXI
+      YCAWEWKLYDAWKWKDWDAWQWKWKWDUIWKWKWDUHHWKWKCVSWKWKWDUJWBGVOWEEFSOBDWEVTWCV
+      R $.
+
+    ${
+      $d A u $.  $d J u $.  $d K u $.  $d M u $.  $d N u $.
+      iseqf1olemqval.q $e |- Q = ( u e. ( M ... N ) |->
+        if ( u e. ( K ... ( `' J ` K ) ) ,
+          if ( u = K , K , ( J ` ( u - 1 ) ) ) ,
+          ( J ` u ) ) ) $.
```
benjub, line 99671: fits on single line

iset.mm
```diff
@@ -99583,6 +99597,769 @@ seq K ( .+ , G , S ) ) $=
...
+      $( Lemma for ~ iseqf1o .  Value of the function ` Q ` .  (Contributed by
+         Jim Kingdon, 28-Aug-2022.) $)
+      iseqf1olemqval $p |- ( ph -> ( Q ` A )
+          = if ( A e. ( K ... ( `' J ` K ) ) ,
+          if ( A = K , K , ( J ` ( A - 1 ) ) ) ,
+          ( J ` A ) ) ) $=
```
benjub, line 99677: fits on single line

iset.mm
```diff
@@ -99583,6 +99597,769 @@ seq K ( .+ , G , S ) ) $=
...
+        VBCEUIUJLUKUL $.
+    $}
+
+    ${
+      iseqf1olemnab.b $e |- ( ph -> B e. ( M ... N ) ) $.
+      iseqf1olemnab.eq $e |- ( ph -> ( Q ` A ) = ( Q ` B ) ) $.
+      iseqf1olemnab.q $e |- Q = ( u e. ( M ... N ) |->
+        if ( u e. ( K ... ( `' J ` K ) ) ,
+          if ( u = K , K , ( J ` ( u - 1 ) ) ) ,
+          ( J ` u ) ) ) $.
```
benjub, line 99691: fits on single line

iset.mm
```diff
@@ -99583,6 +99597,769 @@ seq K ( .+ , G , S ) ) $=
...
+
+  ${
+    iseqf1olemqf.k $e |- ( ph -> K e. ( M ... N ) ) $.
+    iseqf1olemqf.j $e |- ( ph
+      -> J : ( M ... N ) -1-1-onto-> ( M ... N ) ) $.
+    iseqf1olemqf.q $e |- Q = ( u e. ( M ... N ) |->
+      if ( u e. ( K ... ( `' J ` K ) ) ,
+        if ( u = K , K , ( J ` ( u - 1 ) ) ) ,
+        ( J ` u ) ) ) $.
+
```
benjub, line 99792: fits on single line

iset.mm
```diff
@@ -99583,6 +99597,769 @@ seq K ( .+ , G , S ) ) $=
...
+    ${
+      $d J u x $.  $d K u x $.  $d M u x $.  $d N u $.  $d Q x $.  $d ph u x $.
+      $( Lemma for ~ iseqf1o .  Domain and codomain of ` Q ` .  (Contributed by
+         Jim Kingdon, 26-Aug-2022.) $)
+      iseqf1olemqf $p |- ( ph
+          -> Q : ( M ... N ) --> ( M ... N ) ) $=
```
benjub, line 99798: fits on single line

iset.mm
```diff
@@ -99583,6 +99597,769 @@ seq K ( .+ , G , S ) ) $=
...
+      $d N u x $.  $d N v w x $.  $d Q v w x $.  $d ph u x $.  $d ph v w x $.
+      $d u v w $.
+      $( Lemma for ~ iseqf1o . ` Q ` is a permutation of ` ( M ... N ) ` .
+         ` Q ` is formed from the constant portion of ` J ` , followed by the
+         single element ` K ` (at position ` K ` ), followed by the rest of J
+         (with the ` K ` deleted and the elements before ` K ` moved one
+         position later to fill the gap).  (Contributed by Jim Kingdon,
+         21-Aug-2022.) $)
+      iseqf1olemqf1o $p |- ( ph
+          -> Q : ( M ... N ) -1-1-onto-> ( M ... N ) ) $=
```
benjub, line 99838: fits on single line

etc.

iset.mm
```diff
@@ -109407,6 +110183,76 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
         TVNQWCVOPUPAVOURUSUTVAWDGUGQZVLUHQWFVMUEAWGVOKUIRSFHUJDVLGUHUKUMVBAWGVS
         VIVOVCKLGHVEVFVG $.
     $}
+
+    ${
+      isummolem3.5 $e |- ( ph -> ( M e. NN /\ N e. NN ) ) $.
+      isummolem3.6 $e |- ( ph -> f : ( 1 ... M ) -1-1-onto-> A ) $.
+      isummolem3.7 $e |- ( ph -> K : ( 1 ... N ) -1-1-onto-> A ) $.
+      $( Lemma for summation.  (Contributed by Jim Kingdon, 15-Aug-2022.) $)
```
benjub, line 110191: Lemma for ~isummo ? Same for next theorem.
(We have in our conventions that xxxlem should have a comment beginning with "Lemma for ~xxx.")

benjub (approved):
With a few non-blocking comments above.

iset.mm
jkingdon: git has a few ways to do this, including `git commit --amend` and `git rebase -i`.

Github shows this with an "Outdated" marker next to the comments which refer to a previous version of the pull request.

As for what the preferred practice is, everyone has a somewhat different opinion on this but most of the choices have some downside.

benjub: Ok, thanks. I often use `git commit --amend -am "message"` before submitting in order not to have too many commits in a given PR. I didn't know one could do it after `git push origin mybranch` and doing a PR.
I still do not master `git rebase`...

iset.mm
jkingdon: Will change.

iset.mm
jkingdon: Oh another piece I forgot to mention is `git push -f` (or `git push --force-with-lease` which is pretty much the same). That is perhaps why you couldn't amend and then change a pull request.

Force push (and rebase) is a bit of a sharp edged tool so I won't object if you choose not to use them.

iset.mm
jkingdon: The newline here is trying to make it easier to see which expression is part of the inner `if` and which expression is part of the outer `if`. I'm willing to combine those lines if you think we should, but at least to me this feels a bit different to some of these other cases.

iset.mm
jkingdon: Will remove.

iset.mm
jkingdon: Will remove.

As for the "etc" I combined a few more - hopefully I got enough of them to deal with the worst of this.

iset.mm
jkingdon: Sure, but there isn't (yet) a theorem called isummo so the validation would fail if we worded it that way.

jkingdon:
> With a few non-blocking comments above.

Thanks for the clarification on "non-blocking". I've made changes in response to many of these, and will merge now. As always, we can always tweak things later if we want.

iset.mm
benjub: I think that if a line break is needed because a formula doesn't fit on a line, then the line break should occur at a semantically meaningful place, but if no line break is needed, then there is no reason to add one.

PR #2807, opened 2022-09-01
Add ssidd

benjub:
From the second list at #2710.
Second commit is only automatic application of `scripts/min.cmd`.


wlammen (approved):
random checks were ok.

PR #2810, opened 2022-09-02
wl-euequ1f does not need ax-11 

wlammen:
Mathbox: (micro-)optimize wl-exeq, saves a proof byte and a couple of symbols on the web page.
Prove wl-euequ1f without ax-11.

Delete outdated OLD theorems: One should pay attention to ~rmspecsqrtnqOLD.  This theorem is part of @sorear 's mathbox, and I feel a bit uneasy about messing with its contents.  On the other hand, this OLD theorem tends to pop up in lists over and over again, so if you want to keep OLD theorems in your mathbox for an extended period, I suggest to change its format slightly, so it does not match the regular pattern.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -9640,8 +9640,6 @@
 "mulerpq" is used by "mulassnq".
 "mulerpq" is used by "recmulnq".
 "mulerpqlem" is used by "mulerpq".
-"mulgnnclOLD" is used by "mulgnnassOLD".
-"mulgnndirOLD" is used by "mulgnnassOLD".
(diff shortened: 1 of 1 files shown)


benjub:
I think you did well to delete it. One solution for contributors is to write, say, "Obsolete as of 31-Dec-2029." in their mathboxes.  But then: if it's "OLD", why want to keep it, and if you want to keep it, why mark it as OLD instead of, say, ALT ?

benjub (approved):

PR #2812, opened 2022-09-03
Alternate definition of function value

avekens:
* Alternate definition of function value (1): As proposed by BJ via email on 13-Aug-2022, an alternate definition of iota was introduced (in AV's mathbox), see PR #2800) which is the universe V if it is not defined (in contrast to the current definition, which is the empty set in this case). By this definition, the alternate definition of a function value was adapted accordingly.
* Alternate definition of function value (2): A second alternative of a definition is provided. If the function is not defined, the function value is not in the range of the function (inspired by the definition of `Undef`). This approach was already discussed in PR #2800.

Changes in detail, main set.mm:
* ~moimd, ~csbvargi, ~pwexd moved from mathboxes (TODO: check if additional proofs can be shortened)
* new symbol "'''''"

Changes in detail, AV's mathbox:
* section "General auxiliary theorems (1)" with new auxiliary theorems, and some theorems of other sections moved here.
* new theorems ~fundmdfat, ~dfatprc, ~dfatelrn added for "defined at"
* ~ df-afv and ~ dfafv2 revised, using the alternate iota
* Section "Alternative definitions of function values (2)" added with a second alternate definition of function value (~df-afv2) and related theorems.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 3-Sep-22 pwexd                 moved from GS's mathbox to main
+ 3-Sep-22 moimd                 moved from TA's mathbox to main
+ 3-Sep-22 csbvargi              moved from GM's mathbox to main
 26-Aug-22 elv                   moved from PM's mathbox to main
 18-Aug-22 eel1111   syl1111anc  moved from AS's mathbox to main
 26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 4-Sep-22 funresfunco           moved from AV's mathbox to main
  3-Sep-22 pwexd                 moved from GS's mathbox to main
  3-Sep-22 moimd                 moved from TA's mathbox to main
  3-Sep-22 csbvargi              moved from GM's mathbox to main


wlammen (approved):
The 3 additions to Main are in sensible locations.  The rest seem to be modifications of your Mathbox, which I can easily accept.

jkingdon (approved):

set.mm
```diff
@@ -704453,17 +704504,110 @@ is a wff not satisfied by a (unique) value ` x ` (there is no set, and
...
+     ~ pm2.21i ).  (Contributed by Alexander van der Vekens, 26-May-2017.) $)
+  fveqvfvv $p |- ( ( F ` A ) = _V -> ( F ` A ) = B ) $=
+    ( cfv wceq cvv wcel wi fvex eleq1a ax-mp vprc pm2.21i syl eqcoms ) ACDZBEZF
+    PFPEZFFGZQPFGRSHACIPFFJKSQLMNO $.
+
+  $( Composition of two functions, generalization of ~ funco .  (Contributed by
+     Alexander van der Vekens, 25-Jul-2017.) $)
+  funresfunco $p |- ( ( Fun ( F |` ran G ) /\ Fun G ) -> Fun ( F o. G ) ) $=
+    ( crn cres wfun ccom funco wss wceq ssid cores ax-mp eqcomi funeqi sylibr
+    wa ) ABCZDZEBEPRBFZEABFZERBGTSSTQQHSTIQJABQKLMNO $.
```
benjub, line 704525: I'm surprised it didn't exist yet. Worth moving to Main in my opinion.

set.mm
```diff
@@ -704453,17 +704504,110 @@ is a wff not satisfied by a (unique) value ` x ` (there is no set, and
...
+
+  $( Composition of two functions, similar to ~ fnco .  (Contributed by
+     Alexander van der Vekens, 25-Jul-2017.) $)
+  fnresfnco $p |- ( ( ( F |` ran G ) Fn ran G /\ G Fn B )
+                    -> ( F o. G ) Fn B ) $=
```
benjub, line 704530: Can be strengthened to:
```
fnresfnco $p |- ( ( Fun ( F |` ran G ) /\ G Fn B ) -> ( F o. G ) Fn B ) $=
```
from which the above is directly proved.

benjub:
Thanks @avekens for this PR.  I made two comments above.

The alternate definition using `Undef` looks less interesting: you define the value of a function at a point (in ~df-afv2) in terms of its range, but the range is by definition the set of values, so this is circular.

set.mm
avekens: Is this really true? Let `(dom F i^i ran G ) = (/) `, then `` ( F |` ran G ) = (/) ``, so that ``Fun ( F |` ran G )`` is true (~fun0). But 
` ( F o. G ) = (/) `in this case (~coeq0) and therefore ` B = (/) ` (~fn0). Therefore, this theorem holds only in the (degenerate) case `B = (/) ` ( G Fn B => G Fn (/) => G = (/) => ( F o. G ) = ( F o. (/) ) = (/) => ( F o. G ) Fn B because of (/) Fn (/) ) or if `ran G C_ dom F `).

benjub: Ah, ok. Forgot the definition of Fun.

avekens:
> The alternate definition using `Undef` looks less interesting: you define the value of a function at a point (in ~df-afv2) in terms of its range, but the range is by definition the set of values, so this is circular.

I don't think that the definition is circular, because the range is defined independently from the function values: `ran A` is just the class of all second components of pairs contained in `A`. To mark function values as undefined, we just need to pick a set/class that is guaranteed to be not in the range, and this can be done by `~P U. ran F` without knowing any detail about `F` or `ran F`. 

For example, let `F = { <. 1 , 2 >. , < 1 , 3 >. , <.  2 , 4 >. }`.  Then F is not a function (because ``( F ` 1 )`` is not defined) , but ``( F |` { 2 } ) = { <.  2 , 4 >. }`` is a function (so ``( F ` 2 )`` is defined with ``( F ` 2 ) = 4``). Furthermore, `ran F = { 2 , 3 , 4 }`  , so we can "pick" 0 to represent "undefined" in this case (notice that we have no universal "undefined", it's always in the context of a given function/class only). Then we would have ``( F ` 1 ) = 0``, ``( F ` 2 ) = 4``, ``( F ` 3 ) = 0`` with the second alternate definition, or ``( F ` 1 ) = (/)``, ``( F ` 2 ) = 4``, ``( F ` 3 ) = (/)`` with the current definition of a function value (or ``( F ` 1 ) = _V``, ``( F ` 2 ) = 4``, ``( F ` 3 ) = _V`` with the first alternate definition).

avekens:
BTW: With `F = { <. 1 , (/) >. , < 1 , 1o . >. , <. 2 , 2o >. }`, we would have `ran F = { (/), 1o, 2o }`, and therefore `~P U. ran F = ~P U. { (/), 1o, 2o } = ~P ( ((/) u. 1o ) u. 2o) = ~P ( ( (/) u. { (/) } ) u. { (/) , { (/) }}) = ~P { (/) , { (/) } = ( { (/) , { (/) } } u. { { { (/) } } , { (/) , { (/) } } } )` (see ~pwpwpw0), which seems not to be an ordinal number. This would be the "undefined" value in this case.

set.mm
avekens: Theorem moved to main.

benjub:
As you wrote above, the range is the class of values of a function (in the encoding used by set.mm: "the class of all second components of pairs contained in the function"), and if you want to know `~P U. ran F`, then you need to know `ran F`. Therefore, this alternate definition of function values is circular (not in a strict sense of course, since you have a two-step process: you have a preliminary notion of function value, from which you define the range, from which you define the final notion of function value).

**Edit**: the ultimate and most natural "not in the range, whatever the range is" is the universal class.

benjub (approved):

PR #2813, opened 2022-09-03
Add isummolem2a to iset.mm

jkingdon:
The latest summation theorem intuitionizes in a relatively straightforward way, although it is a long proof with a number of things which need to be changed.

Includes:

* copying 2fveq3 , fveqeq2d and fveqeq2 from set.mm to iset.mm
* one more step in getting rid of `S e. V` hypotheses from `seq` theorems
* intuitionizing iseqcoll
* adding leisorel to iset.mm. Stated as in set.mm but with a new proof.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7234,10 +7234,19 @@ than reals.</TD>
   to ` _om ~<_ A ` but the set.mm proof does not work as is.</TD>
 </TR>
 
+
 <TR>
-  <TD>seqcoll , seqcoll2</TD>
+  <TD>seqcoll</TD>
+  <TD>~ iseqcoll</TD>
+  <TD>The functions ` F ` and ` H ` need to be defined on
+  ` ZZ>= `` M ` not just a subset thereof.</TD>
+</TR>
+
+<TR>
+  <TD>seqcoll2</TD>
   <TD><I>none</I></TD>
-  <TD>Presumably will require significant modifications.</TD>
+  <TD>Presumably can be done with modifications similar
+  to ~ iseqcoll .</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7216,7 +7216,7 @@ than reals.</TD>
 </TR>
 
 <TR>
-  <TD>leiso , leisorel</TD>
+  <TD>leiso</TD>
   <TD><I>none</I></TD>
   <TD>The set.mm proof uses isocnv3 .</TD>
 </TR>


iset.mm
```diff
@@ -110246,6 +110385,89 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
...
+      isummolem2a.g $e |- G = ( n e. NN |-> if ( n <_ ( # ` A ) ,
+        [_ ( f ` n ) / k ]_ B , 0 ) ) $.
+      isummolem2a.h $e |- H = ( n e. NN |-> if ( n <_ N ,
+        [_ ( K ` n ) / k ]_ B , 0 ) ) $.
+      summolem2.5 $e |- ( ph -> N e. NN ) $.
+      summolem2.6 $e |- ( ph -> M e. ZZ ) $.
+      summolem2.7 $e |- ( ph -> A C_ ( ZZ>= ` M ) ) $.
+      summolem2.8 $e |- ( ph -> f : ( 1 ... N ) -1-1-onto-> A ) $.
+      summolem2.9 $e |- ( ph -> K Isom < , < ( ( 1 ... ( # ` A ) ) , A ) ) $.
+      $( Lemma for summation .  (Contributed by Mario Carneiro, 3-Apr-2014.) $)
```
benjub, line 110403: extra space before period

benjub:
I see that you added to iset.mm three theorems recently added to set.mm. But there are many additions to set.mm all the time. Is there a way for you to notice which additions are interesting to make ?
(I'm asking because I still have in the back of my head questions about easier synchronization of both databases.)

iset.mm
jkingdon: Updated.

jkingdon:
> I see that you added to iset.mm three theorems recently added to set.mm. But there are many additions to set.mm all the time. Is there a way for you to notice which additions are interesting to make ?

Almost always because I want to use it in iset.mm. The most common scenario is that I'm copying or intuitionizing a proof which uses a theorem which isn't yet in iset.mm, but sometimes I do a search in set.mm or something like that.

Less often I'll go through a whole section theorem-by-theorem (either to see what changed, or more likely because it never was fully intuitionized in the first place). Even that often has some kind of focus (e.g. "if I do a bunch of number theory, I'll be able to formalize the proof that the square root of 2 is irrational" or "I'll need a bunch of finite set theorems for any of this summation stuff to work").

Sometimes proofs can be copied verbatim (and the only task is finding the analogous location in iset.mm). Sometimes the statement of the theorem is the same but the proof needs a small or large change. And sometimes the theorem needs to be changed as well.

> (I'm asking because I still have in the back of my head questions about easier synchronization of both databases.)

I'm unlikely to complain if someone gets inspired to spend some time on this.

iset.mm
```diff
@@ -110246,6 +110385,89 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
...
+        [_ ( K ` n ) / k ]_ B , 0 ) ) $.
+      summolem2.5 $e |- ( ph -> N e. NN ) $.
+      summolem2.6 $e |- ( ph -> M e. ZZ ) $.
+      summolem2.7 $e |- ( ph -> A C_ ( ZZ>= ` M ) ) $.
+      summolem2.8 $e |- ( ph -> f : ( 1 ... N ) -1-1-onto-> A ) $.
+      summolem2.9 $e |- ( ph -> K Isom < , < ( ( 1 ... ( # ` A ) ) , A ) ) $.
+      $( Lemma for summation.  (Contributed by Mario Carneiro, 3-Apr-2014.) $)
+      isummolem2a $p |- ( ph -> seq M ( + , F , CC )
+          ~~> ( seq 1 ( + , G , CC ) ` N ) ) $=
+        ( vm vx caddc cc cseq cfv c1 cli cuz cfz co wf1o wf clt wiso chash wceq
```
avekens, line 110406: This not the original ~summolem2a of set.mm, so "(Revised by Jim Kingdon, 3-Sep-2022.) $)" should be added, shoudn't it?

jkingdon: Don't know if we want that every time there is some intuitionizing, but sure, this one seems major enough.

Updated.

avekens (approved):

PR #2814, opened 2022-09-03
shorten nftht

wlammen:


benjub (approved):
Express approval !
You may even delete the OLD version already (the shortening would probably have been found by minimize_with).

wlammen:
I keep the OLD version along with the other two, since there was no automatic proving involved, namely I did not use minimize_with.

avekens (approved):

PR #2817, opened 2022-09-07
preset-->proset

### LINKED ISSUE #1289: Should "preset" be "proset"?
@benjub recommended in pull request https://github.com/metamath/set.mm/pull/1286 the following (which I'm moving to its own issue):

> I advise to change
everywhere (i.e. in comments, labels, symbols) "preset" to "proset".
Indeed, "proset" can only mean "preordered set", whereas "preset" can mean
several things (including "set without an equality relation"). Plus,
"proset" is apparently more widely used.  If you don't like "proset"
(personally, I'm not fond of it and I think it is dangerously close to
"poset"), then only use the non-abbreviated "preordered set" in comments
(and still "proset" in labels and symbols).

benjub:
Fixes #1289


wlammen (approved):
This is not my domain of expertise.  But the change is backed by references to Wikipedia and other sites in the referenced #1289.  Sufficiently, from my point of view.  So I am willing to believe the pull request is fit for inclusion.

avekens (approved):
The arguments given in #1289 are convincing, therefore I accept this PR.

PR #2820, opened 2022-09-07
Trujust

benjub:
I think I might move bj-bijust00 to Main (in another PR) since it is the most general theorem of which a theorem already in Main (namely, bijust) is an instance. Is there opposition ? What label would you suggest ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18147,6 +18147,7 @@ New usage of "trsbcVD" is discouraged (0 uses).
 New usage of "trsspwALT" is discouraged (0 uses).
 New usage of "trsspwALT2" is discouraged (0 uses).
 New usage of "trsspwALT3" is discouraged (0 uses).
+New usage of "trujustOLD" is discouraged (0 uses).
 New usage of "truniALT" is discouraged (0 uses).
 New usage of "truniALTVD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -12580,10 +12587,18 @@ When the universe is infinite (as with set theory), such a
        These will be redeclared globally in the predicate calculus section. $)
     vx.tru $f setvar x $.
     vy.tru $f setvar y $.
-    $( Soundness justification theorem for ~ df-tru .  (Contributed by Mario
-       Carneiro, 17-Nov-2013.)  (Revised by NM, 11-Jul-2019.) $)
+    $( Soundness justification theorem for ~ df-tru .  Instance of
+       ~ monothetic .  (Contributed by Mario Carneiro, 17-Nov-2013.)  (Revised
+       by NM, 11-Jul-2019.)  (Proof shortened by BJ, 7-Sep-2022.) $)
```
wlammen, line 12592: After adding a new theorem to Main, I used to update theorems profiting from that addition silently.  Was this just my habit?

benjub: You're right, I'll remove that "shortened" tag.

wlammen (approved):
ok

PR #2821, opened 2022-09-07
sectioning

benjub:
Have T. and F. introduced each in their own subsection (this does not add a nesting level); subsection the section about truth tables and order subsections in the order of definition of the respective connectives.

This follows, on a much smaller scale, the separation of the definitions of conjunction and disjunction in their own sections.


david-a-wheeler (approved):
Makes sense to me!

PR #2824, opened 2022-09-08
discourage dfid3 in favor of df-id

benjub:
Replace usages of dfid3 by df-id.
Write in comment of dfid3 that df-id should be preferred when sufficient.
For that reason, add discouragement tag to dfid3, so that it is not used unintentionally instead of df-id. I wondered if this was a bit too strict, but after all, a discouragement is not an interdiction, so I added the tag. Letting @digama0 know.

Recall: `df-id $a |- _I = { <. x , y >. | x = y }` with DV condition and dfid3 without DV condition.

By the way: dfid4 is `_I = { <. x , x >. | x = x }`. Replace it with `_I = { <. x , x >. | T. }` ? Discourage it ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4926,6 +4926,7 @@
 "dfhnorm2" is used by "hilnormi".
 "dfhnorm2" is used by "normf".
 "dfhnorm2" is used by "normval".
+"dfid3" is used by "dfid2".
 "dfiop2" is used by "hoico1".
 "dfiop2" is used by "hoico2".
 "dfiop2" is used by "hoif".
@@ -15120,6 +15121,7 @@ New usage of "dfcnqs" is discouraged (4 uses).
 New usage of "dfdecOLD" is discouraged (35 uses).
 New usage of "dfdp2OLD" is discouraged (1 uses).
 New usage of "dfhnorm2" is discouraged (3 uses).
+New usage of "dfid3" is discouraged (1 uses).
 New usage of "dfiop2" is discouraged (3 uses).
 New usage of "dfiota4OLD" is discouraged (0 uses).
 New usage of "dfpjop" is discouraged (4 uses).
--- a/discouraged
+++ b/discouraged
@@ -4926,6 +4926,7 @@
 "dfhnorm2" is used by "hilnormi".
 "dfhnorm2" is used by "normf".
 "dfhnorm2" is used by "normval".
+"dfid2" is used by "fsplit".
 "dfid3" is used by "dfid2".
 "dfiop2" is used by "hoico1".
 "dfiop2" is used by "hoico2".
@@ -15121,6 +15122,7 @@ New usage of "dfcnqs" is discouraged (4 uses).
 New usage of "dfdecOLD" is discouraged (35 uses).
 New usage of "dfdp2OLD" is discouraged (1 uses).
 New usage of "dfhnorm2" is discouraged (3 uses).
+New usage of "dfid2" is discouraged (1 uses).
 New usage of "dfid3" is discouraged (1 uses).
 New usage of "dfiop2" is discouraged (3 uses).
 New usage of "dfiota4OLD" is discouraged (0 uses).


wlammen:
I am not sure whether I understand this.  We often have v-Versions of theorems, for example mo2 exists as mo2v in a variant, or perhaps more to the point here: df-sb vs. sb6, or the other way round: ax6v vs. ax6.  In the case of df-id, I think, everything is perfect: You have a definition with dv conditions and afterwards show in a theorem that 'bundling' is possible.  Why should clients not use the bundled version, if proven?
It is perhaps a sort of shortening, if you can prove a theorem from a v-version instead of a bundled one.  Or is there more to it?

avekens:
> By the way: dfid4 is `_I = { <. x , x >. | x = x }`. Replace it with `_I = { <. x , x >. | T. }` ? Discourage it ?

I think you mean dfid2? The numbering of the dfid labels is not consistent (anymore) and may be changed (swap dfid2 and dfid3?). I would keep this theorem as it is (not discourage it - it is used in ~fsplit)

benjub:
@wlammen: I was on the verge of not adding the discouragement tag, and I'm still not sure. As I wrote, it's in order to avoid unintentional used of dfid3 when df-id is sufficient, but not to forbid its use when it is needed. If you insist, I can remove it, but I'd like to hear @digama0's opinion on that. But at least we should keep the comment "use df-id when sufficient". As you note, this is not a shortening, but a proof using df-id instead of dfid3 when sufficient is more "economical". The situation is like ax6v/ax6 you mentioned (reading it, I see that there is no discouragement tag to ax6 but the comment says "use ax6v when sufficient"), and the reason is the same: in order to prove the bundled version, you need the auxiliary axioms ax-10 to ax-13 (and in particular ax-13) and ax-ext (and df-clab, df-cleq, df-clel), so a proof using df-id is more economical.

@avekens: you're right, it's dfid2. There is also a sense of economy in using `T.` instead of `x = x` (and the expression `{ <. x , x >. | T. }` is an instance of `{ <. x , y >. | ph }` with DV(x,ph), DV(y,ph), which may simplify some future usages). As for ~split, I would modify the step ~equid with the more elementary ~tru. I would not add a discouragement tag: this would be too strict indeed.

digama0:
Note that dfid3 and dfid2 have been identified as problematic lemmas for the MM0 exporter, because it generates an [unsound definitional axiom](https://github.com/digama0/mm0/blob/6391620f1e0c4878586b95e9badc127bd63258eb/mm0-lean/mm0/set/basic.lean#L393-L396) for df-opab in the bundled case. I am currently basically ignoring this problem, but that only works because dfid3 is basically unused, and thus I am interested in keeping it that way. (There are no examples of bundling of a syntax constructor which has multiple binding arguments in set.mm except for dfid2. There are very few multiple binders in set.mm to begin with: the main ones are df-opab, df-mpt2, and df-oprab; and they are never used bundled except in counterexample-flavored lemmas like dfid2.)

benjub:
@digama0 : so to be explicit, you agree with putting a discouragement tag to both ?

benjub:
@digama0 :
* what phrasing would you put in the comment of dfid2, dfid3 to explain why they are discouraged ?
* should we do the same with df-opab, df-oprab, df-mpt2: for instance, discourage usage and restate them with DV conditions ?

digama0:
The axioms themselves can't have DV conditions, that will trip the definition checker, but I guess you mean to make all uses go through a restatement with a DV condition, which I suppose is okay.

As for why they are discouraged, it is partially due to laziness on my part in the implementation: I did not account for the fact that not only *theorems* can be bundled but also *definitions*: normally each definition turns into a HOL-like constructor function with a type like `copab: (setvar -> setvar -> wff) -> class` (for df-opab), but with this rendering it is not the case that `opab (λ x y, ph(x,y))` corresponds to `{ <. x , y >. | ph }` in the bundled case, because in the HOL rendering `opab (λ x x, ph(x))` is simple name shadowing (so for example `opab (λ x x, T.)` is still `_V X. _V`) while in metamath the action of bundling can do arbitrary things to the RHS expression (in this case making it partially degenerate to `_I`), so you just need an entirely separate definition `copab_b: (setvar -> wff) -> class` which is used whenever metamath would use `{ <. x , x >. | ph }`.

That doesn't really answer your question about what to put in the comment though; it's just a lot of work for a really rare case.

jkingdon:
> Why should clients not use the bundled version, if proven?

I don't spend enough time in the logic parts of either set.mm or iset.mm to feel like I have a super-informed opinion, but I guess my feeling is that bundling has created a lot of hassles for little benefit. As mathematicians we have this impulse of "what's the most general version?" and "how can we figure out how to avoid adding conditions". While this is a perfectly sensible attitude in most situations, I'm not sold on applying this thinking to distinct variable constraints. This is why the change, in ghilbert and Metamath Zero, to make distinctness the default and non-distinctness to need a declaration, is potentially more consequential than it appears. While the exact same rules can be expressed each way, it sets expectations a bit differently in terms of whether distinct variable constraints are to be removed where possible, or only when there is a concrete need.

Even if we agree to this it is perhaps hard to sum up in a sentence to put in a comment, so if we do make `dfid3` discouraged I might just say "This theorem is discouraged in favor of ~ df-id because it is generally easier to think about the case where variables are distinct, unless there is a concrete reason to allow them to be non-distinct" or something relatively generic like that.

As for the analogy to other situations, I think I'm probably saying that we could have just used https://us.metamath.org/mpeuni/sb6.html as the definition of substitution (well, or some variant on that theme, I don't know if that's literally how it would work given the definition checker's rules) and saved ourselves a lot of trouble without causing any significant problems. But I haven't investigated that in any detail (especially not recently) and it is perhaps better to leave that for another day.

avekens:
Is there still some ongoing activity with this PR? I want to finally review it after everything is finished...

wlammen:
Thanks, @jkingdon for your explanations and your point of view.  I read it with great interest.  But I think this is not the place to continue the discussion.

benjub:
I discouraged both dfid3 and dfid2, and added a short comment to explain why.  @digama0's explanation above concerns the definition df-id and why the DV conditions in the RHS among the binding variables are important (roughly: bundling among binding variables does not translate straightforwardly to the formalism of lambda-calculus). @jkingdon's explanation, while true, is I think more subjective (since it's about ease of thinking), and does not catch the particularity of multiple binders (for instance, the situation is not the same with, e.g., `ax6e |- E. x x = y` or `sb1 |- ( [ y / x ] ph -> E. x ( x = y /\ ph ) )` (edit), which are bundled but do not exhibit bundling among binding variables).

Ready for review.

On a separate note: this is not at all to discourage producing bundled theorems. I would say: prove general statements, but use only what you really need (without going out of your way doing so). It's a bit the same as "be conservative in what you produce and liberal in what you accept".

digama0:
It's worth mentioning that *writing* bundled theorems is not a problem for the MM0 exporter, as long as you don't *use* them in a degenerated form, since by default it will only translate the unbundled version of a theorem and lazily create the bundled ones.

discouraged
```diff
@@ -4926,6 +4926,8 @@
 "dfhnorm2" is used by "hilnormi".
 "dfhnorm2" is used by "normf".
 "dfhnorm2" is used by "normval".
+"dfid2" is used by "fsplit".
```
avekens, line 4929: I think ~dfid2 should be replaced by ~df-id  in the proof of ~fsplit, making it 1 byte longer...

avekens (changes requested):
~dfid2 should be replaced by ~df-id in the proof of ~fsplit.

discouraged
benjub: It looks like the modification of ~fsplit to use df-id instead of dfid2 would not be too complicated, but I don't see such a modification as "1 byte longer" ? Maybe you can do it in a following PR ?
At least, I don't see this as merge-blocking.

avekens (approved):

discouraged
benjub: (I did `MM-PA> EXPAND dfid2`)

avekens: This is what I did:

MM> pr fsplit
You will be working on statement (from "SHOW STATEMENT fsplit"):
25840 fsplit $p |- `' ( 1st |` _I ) = ( x e. _V |-> <. x , x >. ) $= ... $.
MM-PA> sh n/lem
 88 vex          $p |- x e. _V
 90 vex          $p |- y e. _V
 91 88,90 brcnv  $p |- ( x `' ( 1st |` _I ) y <-> y ( 1st |` _I ) x )
126 vex          $p |- x e. _V
127 126 brres    $p |- ( y ( 1st |` _I ) x <-> ( y 1st x /\ y e. _I ) )
242 19.42v       $p |- ( E. z ( ( 1st ` y ) = x /\ y = <. z , z >. ) <-> ( (
                                      1st ` y ) = x /\ E. z y = <. z , z >. ) )
317 vex          $p |- z e. _V
319 vex          $p |- z e. _V
320 317,319 op1std  $p |- ( y = <. z , z >. -> ( 1st ` y ) = z )
321 320 eqeq1d   $p |- ( y = <. z , z >. -> ( ( 1st ` y ) = x <-> z = x ) )
322 321 pm5.32ri  $p |- ( ( ( 1st ` y ) = x /\ y = <. z , z >. ) <-> ( z = x /\
                                                            y = <. z , z >. ) )
323 322 exbii    $p |- ( E. z ( ( 1st ` y ) = x /\ y = <. z , z >. ) <-> E. z (
                                                   z = x /\ y = <. z , z >. ) )
379 fo1st        $p |- 1st : _V -onto-> _V
383 fofn         $p |- ( 1st : _V -onto-> _V -> 1st Fn _V )
384 379,383 ax-mp  $a |- 1st Fn _V
386 vex          $p |- y e. _V
393 fnbrfvb      $p |- ( ( 1st Fn _V /\ y e. _V ) -> ( ( 1st ` y ) = x <-> y
                                                                      1st x ) )
394 384,386,393 mp2an  $p |- ( ( 1st ` y ) = x <-> y 1st x )
432 dfid2        $p |- _I = { <. z , z >. | z = z }
Press <return> for more, Q <return> to quit, S <return> to scroll to end... q
MM-PA> del step 432
A 2-step subproof at step 432 was deleted.  Steps 432:689 are now 431:688.
431                      eleq1i.1=?        $? |- _I = { <. z , z >. | z = z }
MM-PA> ass 431 df-id
MM-PA> improve all
A proof of length 1 was found for step 432.
A proof of length 1 was found for step 431.
Steps 431 and above have been renumbered.
CONGRATULATIONS!  The proof is complete.  Use SAVE NEW_PROOF to save it.
Note:  The Proof Assistant does not detect $d violations.  After saving
the proof, you should verify it with VERIFY PROOF.
MM-PA> sh n/c
Proof of "fsplit":
      ( vy vz cv c1st cid cres ccnv wbr copab cop wceq cvv cmpt vex wcel wa wex
      exbii bitri brcnv brres cfv 19.42v op1std eqeq1d pm5.32ri wfn wb wfo fofn
      fo1st ax-mp fnbrfvb mp2an df-id eleq2i nfe1 elopab biantru 3bitr4i bitr2i
      19.9 equid anbi12i 3bitr3ri id opeq12d eqeq2d ceqsexv opabbii wrel relcnv
      dfrel4v mpbi mptv 3eqtr4i ) ADZBDZEFGZHZIZABJZVSVRVRKZLZABJWAAMWDNWBWEABW
      BVSVRVTIZWEVRVSVTAOZBOZUAWFVSVREIZVSFPZQZWEVSVREFWGUBWKCDZVRLZVSWLWLKZLZQ
      ZCRZWEVSEUCZVRLZWOQZCRWSWOCRZQWQWKWSWOCUDWTWPCWOWSWMWOWRWLVRWLWLVSCOZXBUE
      UFUGSWSWIXAWJEMUHZVSMPWSWIUIMMEUJXCULMMEUKUMWHMVSVREUNUOWJVSWLWLLZCCJZPZX
      AFXEVSCCUPUQWOXDQZCRZCRXHXFXAXHCXGCURVCXDCCVSUSWOXGCXDWOCVDUTSVAVBVEVFWOW
      ECVRWGWMWNWDVSWMWLVRWLVRWMVGZXIVHVIVJTTTVKWAVLWAWCLVTVMABWAVNVOABWDVPVQ
      $.
---------The proof of "fsplit" (732 bytes) ends above this line.
MM-PA>

jkingdon (approved):

jkingdon:
> @jkingdon's explanation, while true, is I think more subjective

Fair enough. If we can state this in terms of use of the axiom ax-13 (or the like) I can see the appeal.

discouraged
benjub: @avekens : your proof has to have a DV condition violation. Did you do `MM> VERIFY PROOF fsplit` ?

avekens: @benjub Yes, you are right, I get:

```
There is a disjoint variable ($d) violation at proof step 160.  Assertion
"df-id" requires that variables "x" and "y" be disjoint.  But "x" was
substituted with "z" and "y" was substituted with "z".  These substitutions
have variable "z" in common.
```

I thought I had checked this, but maybe I mixed up something.

As a consequence: we should not change the proof of ~fsplit, and take it as an example for the use of ~dfid2 (when it is not easily replaceable by ~df-id).

benjub: @avekens Ok.  Then I'm going to merge.

PR #2827, opened 2022-09-11
overdue OLD theorems deleted

avekens:
Theorems which were marked as "obsolete" (suffix OLD) in the context of issue #2118 are deleted

AV's Mathbox:
* comments adjusted
* theorems ~afv2orxorb , ~afv2fv0xorb added

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -129,59 +129,6 @@
 "0vfval" is used by "nvrinv".
 "0vfval" is used by "nvsz".
 "0vfval" is used by "nvzcl".
-"10nn0OLD" is used by "dec0hOLD".
-"10nn0OLD" is used by "dec0uOLD".
-"10nn0OLD" is used by "decaddOLD".
-"10nn0OLD" is used by "decaddcOLD".
-"10nn0OLD" is used by "decclOLD".
-"10nn0OLD" is used by "decma2cOLD".
-"10nn0OLD" is used by "decmaOLD".
-"10nn0OLD" is used by "decmacOLD".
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -40170,7 +40170,7 @@ will do (e.g., ` O = (/) ` and ` T = { (/) } ` , see ~ 0nep0 ).
       AVBUQUODCIZJZVAVBUNUPVHUOCDTUAVAVIVBVAVIDBIZVHJVBVAUOVJVHADBNOBCDFGPQRUGS
       UHUIUTUQVCABCDUJVCUOVHUPABDCUJDCTUKULUM $.
 
-    $( Obsolete as of 9-Sep-2021.  Use ~ prel12g instead.  (Contributed by NM,
+    $( Obsolete as of 16-Jun-2022.  Use ~ prel12g instead.  (Contributed by NM,
```
wlammen, line 40173: prel12OLD is referenced in dfac2OLD, so it cannot be deleted before that OLD theorem.

wlammen (approved):

set.mm
avekens: Yes, therefore I increased the due date so that it can be removed together with the OLD theorems using it.

wlammen: My comment was meant to help understanding why the deletion time was postponed.  This cannot be understood by looking just at the piece of text presented here.  In fact, it is used in prel12gOLD, too, but that goes out of scope earlier.

PR #2829, opened 2022-09-12
Minor changes related to df-id

benjub:
* move residOLD earlier
* add ~elid (using residOLD)
* mathbox: add ~bj-e…lid4, sectioning

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+13-Sep-22 epelc     epeli
 13-Sep-22 a1tru     trud
 13-Sep-22 trud      mptru
  8-Sep-22 el2v2     elvd        moved from PM's mathbox to main
--- a/discouraged
+++ b/discouraged
@@ -17359,7 +17359,7 @@ New usage of "relsnopOLD" is discouraged (0 uses).
 New usage of "renegclALT" is discouraged (0 uses).
 New usage of "renicax" is discouraged (0 uses).
 New usage of "resfunexgALT" is discouraged (0 uses).
-New usage of "residOLD" is discouraged (0 uses).
+New usage of "residOLD" is discouraged (1 uses).
 New usage of "ressval3dOLD" is discouraged (1 uses).
 New usage of "retbwax1" is discouraged (0 uses).
 New usage of "retbwax2" is discouraged (3 uses).
--- a/discouraged
+++ b/discouraged
@@ -11655,6 +11655,7 @@
 "relrngo" is used by "rngoablo2".
 "relrngo" is used by "rngoi".
 "relrngo" is used by "rngosn3".
+"residOLD" is used by "elid".
 "ressval3dOLD" is used by "estrresOLD".
 "retbwax2" is used by "merco1lem2".
 "retbwax2" is used by "merco1lem3".


benjub:
Proposal: currently, ~epel is
```
epel |- ( x _E y <-> x e. y )
```
I propose to revise it to
```
epel |- ( A _E x <-> A e. x )
```
whose proof has the same essential steps (there is just one floating hypothesis ~cv to say that x is a class, instead for two). Then, theorems using it will need on more ~cv (if it is not already among the floating hypotheses of that proof). I think the more general form of ~epel is more important than these minor changes.

set.mm
```diff
@@ -521689,20 +521699,21 @@ automatically a set (see ~ bj-mooreset ).  (Contributed by BJ,
...
 
 $(
 -.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
-  Diagonal in a Cartesian square
+  Identity relation (complements)
 -.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
 
-  Complements on the idendity relation and definition of the diagonal in the
-  Cartesian square of a set.
+  Complements on the idendity relation.
```
avekens, line 521716: idendity -> identity

set.mm
```diff
@@ -50971,6 +50977,16 @@ the restriction (of the relation) to the singleton containing this
       CTUDUEUFVCUTVADLDAUGUTVADUHUKTUIUJ $.
   $}
 
+  ${
+    $d x A $.
+    $( Characterization of the elements of the identity relation.  (Contributed
+       by BJ, 28-Aug-2022.) $)
+    elid $p |- ( A e. _I <-> E. x A = <. x , x >. ) $=
+      ( cid wcel cvv cres cv cop wceq wrex wex wrel reli residOLD eqcomi eleq2i
```
avekens, line 50985: Why cannot ~dfrel3 be used here? If ~residOLD is really needed, then ~residOLD cannot be obsolete

avekens (changes requested):
see inline remarks

avekens:
> Proposal: currently, ~epel is
> 
> ```
> epel |- ( x _E y <-> x e. y )
> ```
> 
> I propose to revise it to
> 
> ```
> epel |- ( A _E x <-> A e. x )
> ```
> 
> whose proof has the same essential steps (there is just one floating hypothesis ~cv to say that x is a class, instead for two). Then, theorems using it will need on more ~cv (if it is not already among the floating hypotheses of that proof). I think the more general form of ~epel is more important than these minor changes.

There are currently 36 proofs using ~epel. Is there any of them which can profit from the change? If yes, I would agree with the change, otherwise not (at least until it can actually used with benefit - this could be added to the comment maybe).

set.mm
benjub: ~dfrel3 is after ~elid so cannot be used at this stage. I think it is possible to move theorems around to have dfrel3 before elid and therefore use it, but that subsection is kind of a mess, because many definitions are introduced "en masse". Actually, I decided to first try to reorder a bit that subsection (in a following PR) and then resume work related to the identity relation. I'll put a TODO in ~elid to tell to move reorder statements and use dfrel3.

benjub:
> There are currently 36 proofs using ~epel. Is there any of them which can profit from the change? If yes, I would agree with the change, otherwise not (at least until it can actually used with benefit - this could be added to the comment maybe).

A quick check gives me this (where `epev $p |- A _E x <-> A e. x )`):
```
Proof of "oismo" decreased from 1205 to 1192 bytes using "epev".
Proof of "ordiso2" decreased from 2520 to 2517 bytes using "epev".
```
The steps it saves are the essential steps ~vex (twice) while the steps it adds in other theorems are syntactic steps (~cv, 36 times), so there is a gain on the webpages, for instance. In terms of bytes, it gains 13+3=16 bytes and loses 36 bytes, but to me that's less important than having a more general statement.

avekens:
Since changing the proof of ~oismo provides my requested application, I agree with this change now.

set.mm
avekens: That's fine, thanks.

benjub:
Rk: as for the "Merge branch 'develop' into ridxp" commit, I did (while checked out in branch `ridxp`):
```(ridxp) $ git merge --strategy-option theirs develop```
that is, I took `develop` to override `ridxp` when conflicts happened.

avekens (approved):

PR #2832, opened 2022-09-16
Misc

benjub:
Minor edits, but many of them. I suggest reviewing commits independently. See commit messages. Explanations on demand.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12701,6 +12701,7 @@
 "ubthlem3" is used by "ubth".
 "un0.1" is used by "sspwimpVD".
 "un2122" is used by "suctrALT3".
+"unisnOLD" is used by "unisngOLD".
 "unop" is used by "cnvunop".
 "unop" is used by "counop".
 "unop" is used by "unopadj".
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -47135,10 +47135,10 @@ The membership (or epsilon) relation
 
   ${
     $d A x y $.  $d B x y $.
-    $( The epsilon relation and membership are the same when the "containing"
-       class is a set.  General version of ~ epel and closed form of ~ epeli .
-       (Contributed by Scott Fenton, 27-Mar-2011.)  (Revised by Mario Carneiro,
-       28-Apr-2015.) $)
+    $( The epsilon relation and the membership relation are the same when the
```
jkingdon, line 47138: I don't think we want to use the word "relation" for `e.` because it isn't a relation in the sense of https://us.metamath.org/mpeuni/df-rel.html

In the comment for `$c e.` we use the terminology "membership predicate" which sounds better.

jkingdon (changes requested):
Looks good except for one small wording thing.

set.mm
benjub: I aligned the wording with the other two statements (epeli/epel/epelg). But I'm ok to change all three, since indeed it's a predicate. Are you proposing:
```
The membership relation and the membership predicate are the same when...
```
? Or maybe something other than "are the same" ? Maybe "agree" since a relation and a predicate cannot be the same since they are of a different nature. So maybe best is:
```
The membership relation and the membership predicate agree when...
```
?

jkingdon: > I aligned the wording with the other two statements (epeli/epel/epelg).

Ah OK.

> But I'm ok to change all three, since indeed it's a predicate. Are you proposing:
> ```
> The membership relation and the membership predicate agree when...
> ```
> 
> ?

That seems like the best choice.

jkingdon (approved):

jkingdon:
I approve assuming that it will be updated with the "The membership relation and the membership predicate agree " wording.

PR #2836, opened 2022-09-17
Axextmo

benjub:
Two small changes for more explicit labels. See commit messages.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+17-Sep-22 zfnuleu   ---         deleted; use nulmo instead
 17-Sep-22 bm2.5ii   uniordint
 17-Sep-22 bm1.1     ---         deleted; use axextmo instead
 13-Sep-22 epelc     epeli
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+17-Sep-22 bm2.5ii   uniordint
 17-Sep-22 bm1.1     ---         deleted; use axextmo instead
 13-Sep-22 epelc     epeli
 13-Sep-22 a1tru     trud
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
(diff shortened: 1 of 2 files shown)


set.mm
```diff
@@ -23862,11 +23862,22 @@ number of additional axioms (mainly to replace definitions like ~ df-or and
...
     $d x y z $.  $d ph z $.
-    bm1.1.1 $e |- F/ x ph $.
-    $( Any set defined by a property is the only set defined by that property.
-       Theorem 1.1 of [BellMachover] p. 462.  (Contributed by NM, 30-Jun-1994.)
-       (Proof shortened by Wolf Lammen, 13-Nov-2019.) $)
-    bm1.1 $p |- ( E. x A. y ( y e. x <-> ph ) ->
+    axextmo.1 $e |- F/ x ph $.
+    $( There exists at most one set with prescribed elements.  Theorem 1.1 of
+       [BellMachover] p. 462.  (Contributed by NM, 30-Jun-1994.)  (Proof
+       shortened by Wolf Lammen, 13-Nov-2019.)  Use the uniqueness quantifier.
```
wlammen, line 23868: According to Wikipedia https://en.wikipedia.org/wiki/Uniqueness_quantification E* is not a uniqueness quantifier.  A linked PDF in this article denotes "at most" as a "numerical quantifier.

wlammen (changes requested):

set.mm
benjub: I thought "uniqueness quantifier" is self-explanatory. I can rename it "at-most-one quantifier" if you prefer. (but not "numerical quantifier" or "counting quantifier")

wlammen: This is about having a consistent naming convention.  Almost everyone understands what is meant here.  So: Will we relax the idea of a uniqueness quantifier such that it covers E*?  And if it is viewed as a rare but valid exemption here, what are the limitations?  My personal preferences do not really count.

benjub: You're right, consistency first. Currently, set.mm is not consistent on that, and "uniqueness quantifier" is used sometimes for `E*` (rmoeq1f, cbvrmov) and sometimes for `E.`. I'm opening an issue, but maybe we can merge this now, since it will be caught if we do a Ctrl+F on `uniqueness quantif` (on the unwrapped version of set.mm) in a future consistent renaming.

benjub: Ok: used "at-most-one quantifier" in latest commit.

wlammen (approved):

Issue #2786, opened 2022-08-22
Proposal: Switch to logical quotation system by convention


david-a-wheeler:
I propose that we document in our conventions that we use the *logical* quoting system, not the *typesetters'* quoting system, within future text and that we slowly switch to it. In the *logical* system, quotations only include punctuation if they are in the original quote; in the *typesetters'* system, commas and periods (but typically not other punctuation) are moved from the outside to the inside of a quote if they appear after the quote, even if the quote did not include the punctuation. Sometimes these are called "British" and "American" quoting, but those names are misleading, since both systems are used in formal publications on both sides of the pond.

This proposal was briefly discussed in [PR #2772](https://github.com/metamath/set.mm/pull/2772).

Currently we inconsistently use both, depending on the committer, and that inconsistency can be confusing. So I think it's worth picking one and moving towards it.

Wikipedia (among other organizations) uses the logical quoting system, and there's a good article on [Why Wikipedia uses logical quotation](https://en.wikipedia.org/wiki/Wikipedia:Logical_quotation_on_Wikipedia). One of their arguments is especially relevant here: "Logical quotation is precise; typesetters' often falsifies quoted material". Given that we are trying to be precise, I think that *alone* justifies using it.

It will take time to make things consistent, and there's no reason to drop everything. I suggest starting by documenting in our conventions that all *new* text use this convention, and that we encourage people to adjust the text as they adjust other things. We could eventually do a mass fix of the remainder after a while; at that point it would be easier to do.

wlammen:
I am in favour of this change.

avekens:
I fully support David's proposal.

benjub:
In the section `17.1.1  Conventions`, we currently have:
* ~conventions
* ~conventions-label
* ~conventions-contrib

I propose to add ~conventions-nonmath to group the non-mathematical conventions, typically typographical conventions used in comments (e.g., language and spelling, quotation style, spacing, capitalization...) and add a mini-TOC of these labels in the section head comment (like I began to do last time when I added ~conventions-label and ~conventions-contrib.

What do you think ? Recommendations on the order of these four labels and the labels themselves ? (my first idea was "conventions-typography", but there are also mathematical typography conventions, and the relevant grouping in my opinion is "non-mathematical things in their own group")

david-a-wheeler:
@benjub - splitting it up makes sense, but "-nonmath" seems, well, not right.

[...]

PR #2840, opened 2022-09-20
~conventions in set.mm

benjub:
* relabel `conventions-contrib` --> `conventions-comments`
* relabel `conventions-label` --> `conventions-labels`
* harmonize `conventions*` statements and add mini-TOC
* move the already written conventions related to comments to `~conventions-comments`

Apart from moving around and some minor adjustments, the additions are:
* add convention on "logical quotation style" as discussed in #2786; add that the default quotation mark is double quote (since single quote is also used as apostrophe), and single quotes are used for nested quotations;
* since the Metamath program uses "double spaces after periods", I documented it with a short justification;
* added a convention about compound words to conform to what I checked was the most common usage in set.mm.

Potential future PRs:
* be clearer about what the Metamath program "rewrap" command does and what it does not among the indicated conventions;
* add a `~conventions-proofs` to group the conventions already written in `~conventions` that are about proofs.

Writing this, I realized that having the Metamath program rewrap section comments like usual comments could be prioritized in the development of metamath.exe.

Opinions welcome on any of this.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14658,8 +14658,8 @@ New usage of "con5" is discouraged (1 uses).
 New usage of "con5VD" is discouraged (0 uses).
 New usage of "con5i" is discouraged (2 uses).
 New usage of "conventions" is discouraged (0 uses).
-New usage of "conventions-contrib" is discouraged (0 uses).
-New usage of "conventions-label" is discouraged (0 uses).
+New usage of "conventions-comments" is discouraged (0 uses).
+New usage of "conventions-labels" is discouraged (0 uses).
 New usage of "counop" is discouraged (0 uses).
 New usage of "cplgruvtxbOLD" is discouraged (0 uses).
 New usage of "cqpOLD" is discouraged (0 uses).


set.mm
```diff
@@ -419171,116 +419164,181 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
+       single quote also serves as apostrophe), and the single quote in the
+       case of a nested quotation.
+       </p>
        </li>
 
-       <li><b>Comments and layout.</b>
+       <li><b>Sectioning and section headers.</b>
+       <p>
+       The database set.mm has a sectioning system with four levels of titles,
+       signaled by "decoration lines" which are 79-character long repetitions
```
wlammen, line 419236: Maybe 80 characters, too.  This would also be a better match for separations using alternative characters =-=-.

set.mm
```diff
@@ -419171,116 +419164,181 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
+       <li>a line with "@(" (with the "@" replaced by "$");</li>
+       <li>a decoration line;</li>
+       <li>section title indented with two spaces;</li>
+       <li>a (matching) decoration line;</li>
+       <li>[blank line; header comment indented with two spaces;
+            blank line;]</li>
+       <li>a line with "@)" (with the "@" replaced by "$");</li>
+       <li>one blank line.</li>
+       </ul>
+       As everywhere else, lines are hard-wrapped to be 79-character long.
```
wlammen, line 419254: 80 characters

set.mm
```diff
@@ -419171,116 +419164,181 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
-       Header comments are not rewrapped by the Metamath program [as of
-       24-Oct-2021], but similar spacing and wrapping should be used as for
-       other comments: double spaces after a period ending a sentence, line
-       wrapping with line width of 79, and no trailing spaces at the end of
-       lines.
+       As of 20-Sep-2022, section comments are not rewrapped by the Metamath
+       program "MM> WRITE SOURCE set.mm / REWRAP" command, though this is
+       expected in a future version.  Similar spacing and wrapping should be
+       used as for other comments: double spaces after a period ending a
+       sentence, line wrapping with line width of 79, and no trailing spaces
```
wlammen, line 419285: Again 80

set.mm
```diff
@@ -419171,116 +419164,181 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
+       </p>
        </li>
 
-       <li><b>Format of section headers.</b>
-       The database set.mm has a sectioning system with four levels of titles,
-       indicated by "decoration lines" which are 79-character long repetitions
-       of ####, #*#*, =-=-, and -.-. (in descending order of sectioning level).
+       <li><b>Usage of parentheticals.</b>
+       <p>
+       Usually, the comment of a theorem should contain at most one of the
```
wlammen, line 419314: A revision tag should have a reason added to it, like "Avoid ax-10."  That is also a good reason to not collapse several revisions from the same person into a single one.

wlammen (changes requested):

set.mm
```diff
@@ -419171,116 +419164,181 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
...
-       parenthetical (see Metamath Book, p. 143) within the comment of the new
-       theorem.
+       <li><b>Input format.</b>
+       <p>
+       The input format is ASCII.  Tab characters are not allowed.  If
+       non-ASCII characters have to be displayed in comments, use embedded
+       mathematical symbols when they have been defined (e.g.,
+       <pre>` -> `</pre> for ` -> ` ) or HTML entities (e.g., "&amp;eacute;"
+       for "&eacute;").  Default indentation is by two spaces.  Lines are
+       hard-wrapped to be at most 79-character long (this can be achieved,
```
wlammen, line 419194: 80 characters, if the new line counts as well.  See https://stackoverflow.com/questions/32189491/is-the-new-line-character-part-of-the-same-line-as-the-previous-character

benjub: I had left the "79" that was there, but indeed it is 80, so changed it in (all?) 5 places.

set.mm
benjub: Indeed. Added. There are obviously a few conventions yet to be documented.

wlammen:
There is perhaps a difference between the technical line length (that includes the control character 'new line'), and the perceived line length in a book, or text in your editor, so either of them may be picked.  I think, one should be clear what exactly is meant.  I'd like to see a "including/excluding the new line character" at the first mention of the number.

benjub:
> There is perhaps a difference between the technical line length (that includes the control character 'new line'), and the perceived line length in a book, or text in your editor, so either of them may be picked. I think, one should be clear what exactly is meant. I'd like to see a "including/excluding the new line character" at the first mention of the number.

It seems to me it is "excluding the new line character" since some lines in set.mm have indeed 80 characters excluding LF. Do you confirm ?
On Windows, I think lines end with LF,CR ? So should I write "80 characters (excluding the new line and carriage return characters)" ?

wlammen:
After I appended a character to line 10 of set.mm increasing its length to 80 characters (new line excluded), I got the following output:
```
MM> verify markup */file_skip/top_date_skip
?Warning: Line number 10 has 80 characters (should have 79 or less):
    $( New users may want to read http://us.metamath.org/mpeuni/conventions....
```
So the intended, but not enforced by Metamath alone, maximum line length is 79 characters, excluding the new line character.
As for the line endings, Metamath replaces any Windows/Mac line endings with a simple LF during a file read already.  This is automatically normalized.

I suggest to use 79 as the maximum line length, and add an 'excluding the new line character' at its first appearance.

wlammen (approved):

PR #2841, opened 2022-09-21
[mathbox] Various arithmetic theorems + revise elpwbi

icecream17:
Various arithmetic theorems spurred on by trying to multiply faster (or in this case of `1t1e1ALT`, investigating `ax-mulcom`).
In practice I've only found `sqn5ii` (and I guess `sqn5i`) to help. Naming was hard as always.

(Also just noticed that `elpwbi` is just `elpw2i` commuted.)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13270,6 +13270,7 @@ New usage of "1pi" is discouraged (11 uses).
 New usage of "1pr" is discouraged (16 uses).
 New usage of "1psubclN" is discouraged (0 uses).
 New usage of "1sr" is discouraged (8 uses).
+New usage of "1t1e1ALT" is discouraged (0 uses).
 New usage of "2atm2atN" is discouraged (0 uses).
 New usage of "2atnelvolN" is discouraged (0 uses).
 New usage of "2bornot2b" is discouraged (0 uses).
@@ -18151,6 +18152,7 @@ Proof modification of "19.9hOLD" is discouraged (7 steps).
 Proof modification of "19.9tOLD" is discouraged (17 steps).
 Proof modification of "1div0apr" is discouraged (77 steps).
 Proof modification of "1p1e2apr1" is discouraged (7 steps).
+Proof modification of "1t1e1ALT" is discouraged (13 steps).
 Proof modification of "2bornot2b" is discouraged (26 steps).
 Proof modification of "2pm13.193" is discouraged (91 steps).
 Proof modification of "2pm13.193VD" is discouraged (223 steps).


set.mm
```diff
@@ -609582,6 +609582,66 @@ fixed reference functional determined by this vector (corresponding to
...
+      KJZDDKJZIJLADKJZKJZIJZBBKJACKJZUKIJZADEFMUIBUIBKGGUAUNUJUMIJZUKIJUPUJUKUM
+      AAEENDDFFNLULOADEFNNUBUQUOUKIUOAADDIJZIJZKJUJAURKJZIJUQCUSAKBDIJUIDIJCUSB
+      UIDIUIBGPQHADDEFFUCRSAAUREEDDFFUDUEUTUMUJIUTALDKJZKJUMURVAAKVAURDFUFPSALD
+      EOFUGTSUHQTR $.
+  $}
+
+  ${
+    sqn5i.1 $e |- A e. NN0 $.
+    $( The square of a number ending in 5.  This shortcut only works because 5
+       is half of 10. (Contributed by Steven Nguyen, 16-Sep-2022.) $)
```
icecream17, line 609625:
```suggestion
       is half of 10.  (Contributed by Steven Nguyen, 16-Sep-2022.) $)
```
(easier to commit a review than to go back and rewrap and commit)

set.mm
```diff
@@ -609582,6 +609582,66 @@ fixed reference functional determined by this vector (corresponding to
...
+      ( c5 cdc cmul co cc0 caddc c2 0nn0 deccl nn0cni 5nn0 eqid addid2i decaddi
+      5cn 2nn0 cn0 eqtri c1 5p5e10 decaddci2 diffsqi 5t5e25 peano2nn0 nn0mulcli
+      wcel ax-mp decmulnc mul01i deceq2i 2cn mul02i oveq1i decma ) ACDZUQEFAGDZ
+      AUAHFZGDZEFCCEFZHFAUSEFZIDZCDURUQUTCURAGBJKLQAGCURCBJMURNZCQOZPACUSUQCBMM
+      UQNUSNUBUCUDAGICUTVCCURVABJRMVDUEUSGASUHUSSUHBAUFUIZJKZVBGIAUTEFZIAUSBVFU
+      GJRVHVBAGEFZDVBGDUSGABVFJUJVIGVBAABLUKULTIUMOPGUTEFZCHFGCHFCVJGCHUTUTVGLU
+      NUOVETUPT $.
+
+    sqn5ii.2 $e |- ( A + 1 ) = B $.
+    sqn5ii.3 $e |- ( A x. B ) = C $.
```
benjub, line 609636: Your mathbox your choice, but I would suggest replacing these two hypothses with
```
$e |- B = ( ( A x. ( A + 1 ) )  $.
```
(in local definitions, it's usual to put the definiendum on the LHS)

benjub (approved):

set.mm
```diff
@@ -609582,6 +609582,66 @@ fixed reference functional determined by this vector (corresponding to
...
+      AGFBBGFEFZABGFZQEFZEFPHQGFZEFABABCDCDIRSPESRQABCDJKLMN $.
+  $}
+
+  ${
+    diffsqi.a $e |- A e. CC $.
+    diffsqi.n $e |- N e. CC $.
+    diffsqi.b $e |- ( A + N ) = B $.
+    diffsqi.c $e |- ( B + N ) = C $.
+    $( Difference of squares.  (Contributed by Steven Nguyen, 20-Sep-2022.) $)
+    diffsqi $p |- ( B x. B ) = ( ( A x. C ) + ( N x. N ) ) $=
```
avekens, line 609613: Where is the "difference"? ( ( B x. B ) - ( N x. N ) ) = ( A x. C )?

B^2 = A^2 + 2AN + N^2 
=> B^2 - N^2 = A^2 + 2AN = A ( A + 2N). Why not choosing C = ( A + ( 2 x. N ) ) directly? See also comment of @benjub about ~sqn5ii

icecream17: Really I created the theorem first and then the name/comments. I'm not sure how to describe this theorem

set.mm
icecream17: The applications I had in mind were multiplying numbers ending in 5. Would the current statements be slower than the combined hypothesis?

set.mm
benjub: "Value of the square of the midterm of a 3-term arithmetic progression." ?

benjub (approved):

PR #2842, opened 2022-09-21
mmdeduction: fix broken links

### LINKED ISSUE #2801: Broken links in mmdeduction.html
Links to theorems `hbneg`, `hbnegd` and `abidhb` in the final paragraph of https://us.metamath.org/mpeuni/mmdeduction.html are broken. Apparently they were renamed with an `ENF` suffix in bc8e4ad5798dfbcf87a2fe3fb631511294d55910 and later removed.

I don't know what that suffix was meant to indicate, but I would suggest to either restore the theorems if they are useful examples, or remove the "Another Special Case of the Deduction Theorem" paragraph if they are not.

benjub:
mmdeduction:
* fix broken links;
* stand-alone-->standalone;
* logical quotation style

Fixes #2801.

### DIFF
--- a/mmdeduction.raw.html
+++ b/mmdeduction.raw.html
@@ -162,7 +162,8 @@ to rely on higher-level metatheorems derived outside of that axiom system.
 From a practical point of view, this minimalist philosophy also makes
 automated proof verification extremely simple:  the only metalogical rule
 needed to construct a proof from the axioms is the substitution of an
-expression for a variable.
+expression for a variable (as well as checking the non-violation of disjoint
+variable conditions).
 </p>
 -->
 
--- a/mmdeduction.raw.html
+++ b/mmdeduction.raw.html
@@ -80,7 +80,7 @@ Metalogic, and Metametalogic</A>
 <LI> <A HREF="#set">The Weak Deduction Theorem in Set Theory</A></LI>
 <LI> <A HREF="#setex">Set Theory Example</A></LI>
 <LI> <A HREF="#proof">Informal Proof of the Weak Deduction Theorem</A></LI>
-<LI> <A HREF="#special">Another Special Case of the Deduction Theorem</A>
+<LI> <A HREF="#special">Another Special Case of the Weak Deduction Theorem</A>
 <!--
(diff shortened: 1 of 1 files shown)


avekens (approved):

mmdeduction.raw.html
```diff
@@ -152,17 +152,18 @@
...
+<p>
+Why does Metamath need these complicated-looking steps to convert an inference
+(deduction) into a closed theorem, when other systems of logic typically have
+a (metalogical) rule that allows to do it in one step?  The answer is that
+these steps are part of the price for Metamath's philosophical goal to show
+all proofs <I>directly</I> from the axiom system in its database and not have
+to rely on higher-level metatheorems derived outside of that axiom system.
+From a practical point of view, this minimalist philosophy also makes
+automated proof verification extremely simple:  the only metalogical rule
+needed to construct a proof from the axioms is the substitution of an
```
wlammen, line 164: Well, not exactly.  It needs to decide whether a set variable is part of an expression, too.  I haven't delved too deep into this, but I think, this process is not completely covered by substitution alone.

benjub: @wlammen : Maybe I'm misunderstanding your remark, but for proof verification alone, there is no need to check the "kind" of a variable (i.e., in the case of (i)set.mm, setvar variable, formula variable, class variable). This is thanks to syntactic proof steps, which require to construct things like `class { x e. A | 2 <= /\ x < 4 }` when that expression is to be substituted for a class.

I did not modify this text, which was already here. Now that I'm reading it, I think I'll add as a metalogical rule: "check the non-violation of DV conditions".

wlammen: Even in the simplest case $d x y one may first substitute the variable mapped to x to that for y, but then you still need to compare the resulting expressions for equality - a different textual process.  But please, I just commented what came to my mind.  This need neither be correct nor necessary in this context.  This is not a blocking review remark.

benjub: @wlammen : ok, thanks. I think what I wrote is a faithful description (I modified the python verifier to make it faster and I coded an OCaml verifier, so I have a clear idea of what a verifier has to do; but you're right that at one point, the verifier has to check that the resulting expressions match).

wlammen (approved):

PR #2848, opened 2022-09-23
collapse two adjacent shortened-by-WL tags into one

wlammen:
This conforms to recent rule changes.

nfOLD needs an extra grace period, because it is still referenced by theorems marked a bit later as OLD.

In addition I removed an outdated OLD theorem.  There are more old theorems in the section Complex Vector Spaces bound for deletion, judging by the Obsolete... description alone, but they are still heavily referenced in that section.  It seems all of the section is superseded by a more general concept in section Subcomplex Vector Spaces.  I'd like to see a comment to that respect added there.  It could have saved me some time removing OLD theorems, only to learn the hard way that the section in total must be deleted, not individual theorems, despite of their description.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15610,7 +15610,6 @@ New usage of "h2hva" is discouraged (9 uses).
 New usage of "h2hvs" is discouraged (2 uses).
 New usage of "halfnq" is discouraged (1 uses).
 New usage of "hashfOLD" is discouraged (0 uses).
-New usage of "hashprgOLD" is discouraged (0 uses).
 New usage of "hatomic" is discouraged (1 uses).
 New usage of "hatomici" is discouraged (6 uses).
 New usage of "hatomistici" is discouraged (1 uses).
@@ -19222,7 +19221,6 @@ Proof modification of "gsummptnn0fzvOLD" is discouraged (17 steps).
 Proof modification of "gtinfOLD" is discouraged (249 steps).
 Proof modification of "hashfOLD" is discouraged (95 steps).
 Proof modification of "hashge3el3dif" is discouraged (229 steps).
-Proof modification of "hashprgOLD" is discouraged (172 steps).
 Proof modification of "hb3anOLD" is discouraged (23 steps).
 Proof modification of "hba1-o" is discouraged (34 steps).
 Proof modification of "hba1wOLD" is discouraged (62 steps).


benjub (approved):

benjub:
> In addition I removed an outdated OLD theorem. There are more old theorems in the section Complex Vector Spaces bound for deletion, judging by the Obsolete... description alone, but they are still heavily referenced in that section. It seems all of the section is superseded by a more general concept in section Subcomplex Vector Spaces. I'd like to see a comment to that respect added there. It could have saved me some time removing OLD theorems, only to learn the hard way that the section in total must be deleted, not individual theorems, despite of their description.

Probably related to #2201. I do not know how @avekens wants to handle this.

avekens:
> > In addition I removed an outdated OLD theorem. There are more old theorems in the section Complex Vector Spaces bound for deletion, judging by the Obsolete... description alone, but they are still heavily referenced in that section. It seems all of the section is superseded by a more general concept in section Subcomplex Vector Spaces. I'd like to see a comment to that respect added there. It could have saved me some time removing OLD theorems, only to learn the hard way that the section in total must be deleted, not individual theorems, despite of their description.
> 
> Probably related to #2201. I do not know how @avekens wants to handle this.

I also do not not how to proceed with these deprecated sections: I started to examine/clean them as far as possible, but starting with  "18.4  Operators on complex vector spaces" it would be an immense effort to transfer this material to main set.mm (I do not have the time and ambition to do that).

The only thing I can do is to clean up part 18 a little further (as promised in #2201), and maybe to remove (or at least modify) the "Obsolete .. as of" tags (maybe just simply remove the dates) so that they will not be considered during the normal cleanup procedures.

jkingdon:
> I also do not not how to proceed with these deprecated sections: I started to examine/clean them as far as possible, but starting with "18.4 Operators on complex vector spaces" it would be an immense effort to transfer this material to main set.mm (I do not have the time and ambition to do that).
> 
> The only thing I can do is to clean up part 18 a little further (as promised in #2201), and maybe to remove (or at least modify) the "Obsolete .. as of" tags (maybe just simply remove the dates) so that they will not be considered during the normal cleanup procedures.

Wow. That does look like a big job. Thanks for all the work you did for the first 3 bullet points, and that doesn't mean you have to take on the other 5 just because you did those.

We shouldn't give ourselves a hard time (collectivelhy) if the time and ambition isn't there yet. So I support adjusting the comments as you propose (to paraphrase, making it is clear that this is a "when someone gets around to it" rather than a "this theorem should already be unused, just giving people a bit of time in case they are used to the old one or have an unmerged branch referring to it").

PR #2849, opened 2022-09-23
3 lemmas minimized

mazsa:
Cf. https://github.com/metamath/set.mm/pull/2831#issuecomment-1250894822


benjub (approved):

benjub:
Maybe I merged too fast: if the proof shortenings are the result of automatic minimization, then there is no need for a "proof shortened" tag. Was it the case for these three lemmas ?

mazsa:
It was far from automatic: I worked on this for hours. However maybe someone who is a more skilled metamath program user could make it in an automated way, I don´t know. Please feel free to remove the tag if you can reproduce it automatically (and please @benjub send me a link about how to do this:)

benjub:
@mazsa : I didn't mean to belittle the shortenings: it was a genuine question since I didn't look at the details. I'll look at the proofs on the webpages. I know that this can take a lot of efforts, so thanks for your work !

mazsa:
@benjub I tried not to be ironic, but, after rereading my comment, perhaps I was not neutral in fact. Yes, fpwwe2lem8 was especially nasty: I found the place of syl2an2r only by writing out all possible cases, and it was a lot in that case. I thought maybe we have a tool for this... but now I think Metamath is not an automated prover but a proof assistant, and it helps us a lot without this feature as well.

benjub:
@mazsa : Indeed, I think the only minimization/shortening tools we have are very basic, like minimize_with and maybe some things in mmj2.

jkingdon:
> @mazsa : Indeed, I think the only minimization/shortening tools we have are very basic, like minimize_with and maybe some things in mmj2.

Yeah, when people speak of "minimize" without elaboration they are talking about `minimize_with` (likely invoked as in `scripts/minimize` or another script in that directory or some slight variation on that).

Thanks for the efforts in finding shortenings: there are many benefits to shorter proofs. Even if we had more automated tools I suspect there would still be things they can't do. And although it doesn't apply to this pull request, running the automated minimizer is also appreciated, even if it doesn't generate a "Shortened by" comment!

PR #2850, opened 2022-09-23
Comment edits

benjub:
Comment edits: standardize some spellings; port recent changes to iset.mm.

A few precisions and links added to some comments

as for the removed OLD theorems: even if they were not one year old, I had contributed them and checked that the OLD proof didn't have anything interesting compared to the current ones.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16660,8 +16660,6 @@ New usage of "nfnfcALT" is discouraged (0 uses).
 New usage of "nfntOLD" is discouraged (3 uses).
 New usage of "nfntOLDOLD" is discouraged (0 uses).
 New usage of "nfnthOLD" is discouraged (0 uses).
-New usage of "nfntht2OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -469439,7 +469422,7 @@ embedding at each step ( ` ZZ ` , ` QQ ` and ` RR ` ).  It would be
     esumiun.0 $e |- ( ph -> A e. V ) $.
     esumiun.1 $e |- ( ( ph /\ j e. A ) -> B e. W ) $.
     esumiun.2 $e |- ( ( ( ph /\ j e. A ) /\ k e. B ) -> C e. ( 0 [,] +oo ) ) $.
-    $( Sum over a non necessarily disjoint indexed union.  The inegality is
+    $( Sum over a nonnecessarily disjoint indexed union.  The inegality is
```
wlammen, line 469425: I'd rather use "possibly" here.  But its mathbox material, so ok.

wlammen (approved):

PR #2852, opened 2022-09-24
Remove dependencies on axioms (nnmulcl)

icecream17:
Uses ~ ax-1rid instead of ~ mulid1

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16797,6 +16797,7 @@ New usage of "nn0indALT" is discouraged (3 uses).
 New usage of "nn0pnfge0OLD" is discouraged (0 uses).
 New usage of "nnexALT" is discouraged (3 uses).
 New usage of "nnindALT" is discouraged (0 uses).
+New usage of "nnmulclOLD" is discouraged (0 uses).
 New usage of "nonbooli" is discouraged (0 uses).
 New usage of "norm-i" is discouraged (13 uses).
 New usage of "norm-i-i" is discouraged (2 uses).
@@ -19488,6 +19489,7 @@ Proof modification of "nn0indALT" is discouraged (15 steps).
 Proof modification of "nn0pnfge0OLD" is discouraged (26 steps).
 Proof modification of "nnexALT" is discouraged (34 steps).
 Proof modification of "nnindALT" is discouraged (15 steps).
+Proof modification of "nnmulclOLD" is discouraged (239 steps).
 Proof modification of "normlem7tALT" is discouraged (177 steps).
 Proof modification of "notnotrALT" is discouraged (12 steps).
 Proof modification of "notnotrALT2" is discouraged (2 steps).


set.mm
```diff
@@ -121476,8 +121476,23 @@ subset of complex numbers ( ~ nnsscn ), in contrast to the more elementary
...
+      nncn nnre ax-1rid biimprd mpcom nnaddcl ancoms ax-1cn adddi mp3an3 syl2an
+      syl adantr oveq2d eqtrd syl5ibr exp4b pm2.43b a2d nnind impcom ) BEFAEFZA
+      BGHZEFZVAACIZGHZEFZJVAAKGHZEFZJVAADIZGHZEFZJVAAVIKLHZGHZEFZJVAVCJCDBVDKMZ
+      VFVHVAVOVEVGEVDKAGNOPVDVIMZVFVKVAVPVEVJEVDVIAGNOPVDVLMZVFVNVAVQVEVMEVDVLA
+      GNOPVDBMZVFVCVAVRVEVBEVDBAGNOPAQFZVAVHAUAZVSVHVAVSVGAEAUBZOUCUDVIEFZVAVKV
+      NWBVAVKVNJVAWBVAVKVNVAVKRVNVAWBRZVJALHZEFZVKVAWEVJAUEUFWCVMWDEWCVMVJVGLHZ
+      WDVAASFZVISFZVMWFMZWBATVITWGWHKSFWIUGAVIKUHUIUJWCVGAVJLVAVGAMZWBVAVSWJVTW
+      AUKULUMUNOUOUPUQURUSUT $.
+
+    $( Obsolote proof of ~ nnmulcl as of 24-Sep-2022.  Closure of
```
wlammen, line 121492: 1. Typo Obsolete
2. Prefer Obsolete version of ...

set.mm
```diff
@@ -121476,8 +121476,23 @@ subset of complex numbers ( ~ nnsscn ), in contrast to the more elementary
       PVIVHVAEUIARFZUQRFZVHVALZVGASUQSVJVKKRFVLTAUQKUAUBUCNUDUEUFUGUH $.
 
     $( Closure of multiplication of positive integers.  (Contributed by NM,
-       12-Jan-1997.) $)
+       12-Jan-1997.)  Reduce dependencies on axioms.  (Revised by Steven
```
benjub, line 121479: Can you be more explicit ? I.e.:
```
...).  Remove dependency on ~ ax-A , ~ ax-B , ~ ax-C .  (Revised by...
```

benjub: And, while on this topic: in ~1t1e1ALT, can you write "Alternate proof" instead of "Alternate version" (since this is the same statement), and can you write:
```
using a different set of axioms (add ~ ax-mulrcl , ~ ax-i2m1 , ~ ax-1ne0 , ~ ax-rrecex and
remove ~ ax-resscn , ~ ax-mulcom , ~ ax-mulass , ~ ax-distr ).
```
This may help other people studying the axiomatizations of $\mathbb{C}$.

wlammen: ax-mulcom and ax-mulass are not referenced in the new proof

wlammen (changes requested):

icecream17:
I think I'm missing something about "Obsolete version" and "Obsolete proof". Is the guideline at https://us.metamath.org/mpeuni/conventions.html incomplete?

Edit: I think what I did was read the convention then looked at the example https://us.metamath.org/mpeuni/19.43OLD.html, and somehow used `proof` in `nnmulcl` and `version` in `1t1e1ALT`.

Edit 2: I think the closest meaning is that "alternate version" is when a proof changes slightly while "alternate proof" is when a statement is proved in a major different way.

Edit 3: I'm reconfused now because after rechecking I see there's also "version ≝ the statement being proved".

benjub:
> I think I'm missing something about "Obsolete version" and "Obsolete proof". Is the guideline at https://us.metamath.org/mpeuni/conventions.html incomplete?

I think it's not that important since they will be removed, but if it's the same statement, then it should be "obsolete proof" and if it's a different statement, then it should be "obsolete version", like for ALT statements (where it is more important since they will stay).

As for `~conventions`, it is certainly very incomplete on many things, and we add things as they show up. I'll add the above convention.

benjub (approved):

wlammen:
"Obsolete version of..." is a fixed idiom and the conventions page you mentioned explicitly requires this wording.  Preventing variations of this text eases maintenance, since the file is regularly scanned for this expression to weed out outdated OLD proofs.  Maybe someone will write a script in the future to automate this janitorial task.  This will be more likely if it does not require artificial intelligence.  The current state of set.mm does not provide this grade of uniformity, but this will change within a year, if we now keep an eye on it.

wlammen (approved):

PR #2854, opened 2022-09-24
shorten moeq+eueq; add mosneq

benjub:
* shorten moeq+eueq by swapping them ; similar trick can probably be done with other pairs eu*/mo*.

* add mosneq after seeing https://us.metamath.org/mpeuni/euabsneu.html  By the way, @avekens : here is a shorter proof:
```
Proof of "euabsneu":
      ( cab cv csn wceq wex wmo weu mosneq eqcom mobii mpbi biantru euabsn2 eu5
      wa 3bitr4i ) ABDZCEFZGZCHZUCUBCIZRABJUBCJUDUCUATGZCIUDCTKUEUBCUATLMNOABCP
      UBCQS $.
---------The proof of "euabsneu" (152 bytes) ends above this line.
```
(**edit**: feel free to use it without credit, since you would have found it anyway)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -5888,6 +5888,7 @@
 "erngset-rN" is used by "erngbase-rN".
 "erngset-rN" is used by "erngfmul-rN".
 "erngset-rN" is used by "erngfplus-rN".
+"eueqOLD" is used by "moeqOLD".
 "exatleN" is used by "cdlema2N".
 "exidu1" is used by "cmpidelt".
 "exidu1" is used by "exidresid".
@@ -15397,6 +15398,7 @@ New usage of "erngplus2-rN" is discouraged (0 uses).
 New usage of "erngring-rN" is discouraged (0 uses).
 New usage of "erngset-rN" is discouraged (3 uses).
 New usage of "estrresOLD" is discouraged (0 uses).
+New usage of "eueqOLD" is discouraged (1 uses).
 New usage of "euexALT" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
 New usage of "eumoOLD" is discouraged (0 uses).
@@ -16562,6 +16564,7 @@ New usage of "mndoisexid" is discouraged (2 uses).
 New usage of "mndoismgmOLD" is discouraged (3 uses).
 New usage of "mndoissmgrpOLD" is discouraged (1 uses).
 New usage of "mndomgmid" is discouraged (3 uses).
+New usage of "moeqOLD" is discouraged (0 uses).
 New usage of "mptssALT" is discouraged (0 uses).
 New usage of "mpv" is discouraged (1 uses).
 New usage of "mulassnq" is discouraged (10 uses).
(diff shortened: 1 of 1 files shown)


benjub:
typo: "midterm"-->"middle term" (my bad, sorry @icecream17 !)

wlammen (approved):

avekens:
> ```
> Proof of "euabsneu":
> ---------Clip out the proof below this line to put it in the source file:
>       ( cab cv csn wceq wex wmo weu mosneq eqcom mobii mpbi biantru euabsn2 eu5
>       wa 3bitr4i ) ABDZCEFZGZCHZUCUBCIZRABJUBCJUDUCUATGZCIUDCTKUEUBCUATLMNOABCP
>       UBCQS $.
> ---------The proof of "euabsneu" (152 bytes) ends above this line.
> ```
> 
> (**edit**: feel free to use it without credit, since you would have found it anyway)

Great! Please replace the current proof by the shorter one (maybe already with this PR), and of course you can add a "Proof shortened by" tag (to keep an OLD version is not necessary).

avekens (approved):

set.mm
```diff
@@ -417512,33 +417512,33 @@ or axiom of dependent choice ( ~ ax-dc ) can be used instead.
        and moving them up to the ~ ax-4 through ~ ax-9 section.
        </li>
 
-       <li><b>Alternative (ALT) proofs.</b>
-       If a different proof is significantly shorter or clearer but
-       uses more or stronger axioms, we prefer to make that proof an
-       "alternative" proof (marked with an ALT label suffix), even if
-       this alternative proof was formalized first.
+       <li><b>Alternate (ALT) proofs.</b>
```
wlammen, line 417515: I am not sure whether alternate is a good replacement for alternative.  I thought 'alternate' means 'repeatedly changing' as in alternate current or alternate number sequence (1, -1, 1, -1...).  

This is from the Oxford dictionary:
1. every other; every second.
"she was asked to attend on alternate days"
2. North American
another term for alternative.
"a novel set in an alternate universe"

I think here is support from native English speakers called for.  I personally tend towards 'alternative'.

benjub: I see (for the adjectives):
* https://www.dictionary.com/browse/alternate : constituting an alternative (e.g., "an alternate route")
* https://www.dictionary.com/browse/alternative : there is not quite such meaning

and when interchangeable, I tend to prefer the shorter alternative (<-- this is the noun "alternative" !).

wlammen: I entered 'alternate vs alternative' into Google and the first hit is
https://www.merriam-webster.com/words-at-play/differences-between-alternate-and-alternative
which elaborates on the question at hand.  There seems to be no distinct favorite:
``
Usage commentators, however, have often emphasized a distinction: one should use alternate to mean "occurring or succeeding by turns" and alternative to mean "offering or expressing a choice." Yet a look at the way these words are used makes it clear that things aren't quite so simple.
``
IMO the version of the original author wins in such cases, otherwise I imagine a future of frequently alternating versions here and elsewhere for whatever liking is dominating at a particular moment.

My English skills are not sufficient to decide on this.  So I am going to accept any solution.

avekens (approved):

PR #2858, opened 2022-09-26
add proof about signum function

SimonDold:


benjub (dismissed):
Thanks for the contribution !

I think that `~sgn0iff0` is sufficient and there is no need for `sgnz`. So you can "inline" the proof of the latter in the former:
```
MM> prove sgn0iff0
MM-PA> expand sgnz
MM-PA> save new / c
MM-PA> exit
MM-PA> write source set.mm / r
```
and then remove ~sgnz.

As for the comment, every $p-statement is a "proof that", so there is no need to repeat this. Since the statement already has math symbols, I would prefer plain English for the comments, if not too heavy. Here:
```
An extended real is zero if and only if its signum is zero.
```
Actually, it seems more logical to have the reversed biconditional:  ``( A e. RR* -> ( A = 0 <-> ( sgn ` A ) = 0 ) )``.

Can you add your name in the list at the beginning of the file (alphabetical order of last name) ?

Label: this seems good to me; maybe "iff" is abbreviated differently ? @avekens ?

SimonDold:
Thank you for the feedback @benjub !

I changed it according to your suggestions beside the direction of the biconditional. I think deducing the parameter A from the result sgn(A) is the less trivial direction and the reason I created sgnz in the first place.

benjub:
You can simply add two spaces at the beginning of that line:
```
-$( The signum of an extended real is zero if and only if that extended real
+  $( The signum of an extended real is zero if and only if that extended real
```
Also, "o" comes before "r" in the alphabet !

I still think a reversed biconditional makes more sense, see e.g. `MM> sh st cnveqb,compleq,negeq0,cos11,chcon3i,opcon3b,rmyeq,fsneq` but I won't block because of that. So I'll remove my blocking and let someone else approve.
(Also: the fact that your comment is a bit more complex than the one I suggested is an indication of the naturality of the statement I proposed.)

set.mm
```diff
@@ -152852,6 +152853,17 @@ a function (ordinarily on a subset of ` CC ` ) and produces a new
     ( cmnf cxr wcel cc0 clt wbr csgn cfv c1 cneg wceq mnfxr mnflt0 sgnn mp2an )
     ABCADEFAGHIJKLMANO $.
 
+$( The signum of an extended real is zero if and only if that extended real
+     equals zero.  (Contributed by Simon Dold, 26-Sep-2022.) $)
+  sgn0iff0 $p |- ( A e. RR* -> ( ( sgn ` A ) = 0 <-> A = 0 ) ) $=
```
avekens, line 152858: Sorry, but this theorem is already existing in set.mm: see ~ sgn0bi, contributed by TA in 2018 (in his mathbox). Furthermore, TA's proof is shorter, so his one would be preferred.

Regarding the naming: We usually use the suffix "bi"  or "b" for biconditionals, not the label fragment "iff".

avekens (changes requested):
PR rejected because theorem is already existing.

Furthermore, such theorems should be provided in a mathbox first, before they are moved to main set.mm.

benjub:
Sorry @SimonDold and @avekens for the extra work this caused. I was indeed surprised it was not in Main but I did not check further. Are you sure Thierry's proof is shorter though ? Note that it uses ~sgnd3a.

@SimonDold : you could create a mathbox and add your proof, maybe as `~sd-sgn0bi` with a link to ~sgn0bi, so that we have both proofs on webpages and we can compare. Hopefully, you'll make other contributions and your mathbox will grow.

glacode:
Welcome, @SimonDold 

I agree, it would be nice if you could add your own new mathbox. It will also be easier for you to contribute: here

https://groups.google.com/g/metamath/c/2YeOxZHF3TQ

Norman wrote 'After the main part, each individual user has a "mathbox" with his or her own theorems. In general, users can add or delete whatever they want in their mathboxes as well as make any revisions they want.'

and in the following message, Mario defines it as "a sort of scratchpad for your work".

Other contributors will then be able to take advantage of your work and move to the main part the most useful theorems.

Glauco

benjub:
Mathbox policies are centralized in https://us.metamath.org/mpeuni/mathbox.html

PR #2857, opened 2022-09-26
reduce axiom dependencies in mobidv, eubidv

benjub:
* reduce axiom dependencies in mobidv, eubidv (remove ax-6, ax-7, ax-12); this propagates to several theorems
* relabel eueq(u)1-->eueq(u)i (since they are the inferences associated with eueq(u))
* minor comment edits

PR #2862, opened 2022-09-27
Misc

benjub:
Minor and non-mathematical. See commit messages.
First commit according to https://github.com/metamath/set.mm/pull/2857#issuecomment-1259619943
Third commit: see also https://github.com/metamath/metamath-exe/issues/99

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2258,7 +2258,6 @@
 "bnj523" is used by "bnj934".
 "bnj525" is used by "bnj1040".
 "bnj525" is used by "bnj523".
-"bnj525" is used by "bnj538OLD".
 "bnj525" is used by "bnj539".
 "bnj525" is used by "bnj540".
 "bnj525" is used by "bnj91".
@@ -14106,14 +14105,13 @@ New usage of "bnj519" is discouraged (2 uses).
 New usage of "bnj521" is discouraged (2 uses).
 New usage of "bnj523" is discouraged (3 uses).
 New usage of "bnj524" is discouraged (0 uses).
-New usage of "bnj525" is discouraged (8 uses).
+New usage of "bnj525" is discouraged (7 uses).
 New usage of "bnj526" is discouraged (1 uses).
 New usage of "bnj528" is discouraged (2 uses).
 New usage of "bnj529" is discouraged (3 uses).
 New usage of "bnj534" is discouraged (2 uses).
 New usage of "bnj535" is discouraged (1 uses).
 New usage of "bnj538" is discouraged (3 uses).
-New usage of "bnj538OLD" is discouraged (0 uses).
 New usage of "bnj539" is discouraged (4 uses).
 New usage of "bnj540" is discouraged (2 uses).
 New usage of "bnj543" is discouraged (1 uses).
@@ -16383,7 +16381,6 @@ New usage of "ltrniotacnvN" is discouraged (0 uses).
 New usage of "ltrniotafvawN" is discouraged (0 uses).
 New usage of "ltrniotaidvalN" is discouraged (1 uses).
 New usage of "ltrniotavalbN" is discouraged (0 uses).
-New usage of "ltrnmwOLD" is discouraged (0 uses).
 New usage of "ltrnq" is discouraged (3 uses).
 New usage of "ltsonq" is discouraged (10 uses).
 New usage of "ltsopi" is discouraged (4 uses).
@@ -17011,14 +17008,11 @@ New usage of "opsqrlem6" is discouraged (0 uses).
 New usage of "opthregOLD" is discouraged (0 uses).
 New usage of "orbi1r" is discouraged (0 uses).
 New usage of "orbi1rVD" is discouraged (0 uses).
-New usage of "ordelinelOLD" is discouraged (0 uses).
 New usage of "ordelordALT" is discouraged (0 uses).
 New usage of "ordelordALTVD" is discouraged (0 uses).
-New usage of "ordnbtwnOLD" is discouraged (0 uses).
 New usage of "ordpinq" is discouraged (6 uses).
 New usage of "ordpipq" is discouraged (5 uses).
 New usage of "ordtr3OLD" is discouraged (0 uses).
-New usage of "ordtri3OLD" is discouraged (0 uses).
 New usage of "orim12dALT" is discouraged (0 uses).
 New usage of "orthcom" is discouraged (6 uses).
 New usage of "orthin" is discouraged (2 uses).
@@ -17857,7 +17851,6 @@ New usage of "suctrALT2VD" is discouraged (0 uses).
 New usage of "suctrALT3" is discouraged (0 uses).
 New usage of "suctrALTcf" is discouraged (0 uses).
 New usage of "suctrALTcfVD" is discouraged (0 uses).
-New usage of "suctrOLD" is discouraged (0 uses).
 New usage of "sumdmdi" is discouraged (2 uses).
 New usage of "sumdmdii" is discouraged (2 uses).
 New usage of "sumdmdlem" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -39903,13 +39903,13 @@ will do (e.g., ` O = (/) ` and ` T = { (/) } ` , see ~ 0nep0 ).
...
-    raltpd $p |- ( ph -> ( A. x e. { A , B , C } ps <-> ( ch /\ th /\ ta ) ) )
-      $=
-      ( wa wb ctp wral w3a an3andi a1i wcel cv expcom pm5.32d raltpg syl3anc c0
-      wceq wne tpnzd r19.28zv syl 3bitr2d bianabs bicomd ) ABFGHIUAZUBZCDEUCZAA
-      VCSZVBAVDVBAVDACSZADSZAESZUCZABSZFVAUBZAVBSZVDVHTAACDEUDUEAGJUFHKUFILUFVJ
-      VHTPQRVIVEVFVGFGHIJKLFUGZGUMZABCAVMBCTMUHUIVLHUMZABDAVNBDTNUHUIVLIUMZABEA
-      VOBETOUHUIUJUKAVAULUNVJVKTAGHIJPUOABFVAUPUQURUSUTUS $.
+    raltpd $p
+            |- ( ph -> ( A. x e. { A , B , C } ps <-> ( ch /\ th /\ ta ) ) ) $=
+  ( wa wb ctp wral w3a an3andi a1i wcel cv expcom pm5.32d raltpg syl3anc c0 wne
```
wlammen, line 39908: Why has the indentation changed here?  This seems to be triggered by $= in the last position of a line.

set.mm
```diff
@@ -42649,16 +42649,15 @@ same distinct variable group (meaning ` A ` cannot depend on ` x ` ) and
...
-      nfcv nfin ssrexf 3imtr4g ssrdv iuneq2 iun0 syl6eq syl5sseq ss0 syl uneq1d
-      cv iunxun inundif nfth nfdif id eqidd iuneq12df eqtr3i uncom un0 3eqtr3rd
-      nfun eqtri ) CGHADIZABDJZCKZABDLZCKZMZGVKMZABCKZVKVGVIGVKVGVIGNVIGHVGADCK
-      ZVIGVHDNZVIVONBDOVPFVIVOVPFUMZCPZAVHQVRADQVQVIPVQVOPVRAVHDABDABUAZEUBZEUC
-      AVQVHCRAVQDCRUDUESVGVOADGKGADCGUFADUGUHUIVIUJUKULVLVNHVGAVHVJMZCKZVLVNAVH
-      VJCUNWABHZWBVNHBDUOZWCAWABCCWCAWDUPAVHVJVTABDVSEUQVEVSWCURWCCUSUTSVATVMVK
-      HVGVMVKGMVKGVKVBVKVCVFTVD $.
+    iunxdif3 $p
+               |- ( A. x e. E B = (/) -> U_ x e. ( A \ E ) B = U_ x e. A B ) $=
+  ( vy c0 wceq wral cin ciun cdif cun wss inss2 wcel wrex eliun ax-mp a1i ssrdv
```
wlammen, line 42654: Unexpected indentation change

set.mm
```diff
@@ -653849,23 +653719,22 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
-      SZXMCSZDTSZAXNXOXGAJCXMPUSUTXHXOUOZXPBIVAZXGXPAXOXRXGAXOUOZITBIDUDZVBZXRX
-      SFIVCVDSZGTVCVDSZXTXFSZYAAYBXOMUQYCXSGLVEVROXTFGITVFVGBITDXTXTVHVIVJUTAXG
-      XOVKXPBIVLVMZVNEHDVOZXHHCSZXDTSZXHHCJXLVPYGXHYGUOZYHXHYGVQXQXPWIYIYHWIEHC
-      EHVSZYIYHEXHYGEXJYGEUKZULEXDTXKVTWAXMHWBZXQYIXPYHYLXOYGXHXMHCWCZWDYLDXDTY
-      FWEWFYEWGWJWKWHWLABXCXDUFFGGGIMRAYGBIXDUDZXFSZAHCJQVPAYGYOXSYDWIZAYGUOZYO
-      WIEHCYJYQYOEAYGEKYKULEYNXFEBIXDEIVSXKWMEXFVSWNWAYLXSYQYDYOYLXOYGAYMWDYLXT
-      YNXFYLBIDXDYFWOWEWFYPOWPWGWQWKUFGGWTUGGUHUGSAGLXAVRWRWS $.
+    fprodcnlem $p
+       |- ( ph -> ( x e. X |-> prod_ k e. ( Z u. { W } ) B ) e. ( J Cn K ) ) $=
+  ( wcel cc csn cun cprod cmpt csb cmul co ccn cv wa cdif nfv nfcsb1v cfn ssfid
```
wlammen, line 653724: Unexpected indentation change

set.mm
```diff
@@ -654568,40 +654437,38 @@ closed under the multiplication ( ' X ' ) of a finite number of
...
-      WCAUWBUVCUVPABEVAZWEAUUSUVCUVPUUTWEUVQUVAUVRVABCUVAWKWNWLWMWOWPUVAYQUUGWQ
-      WRWSIUULUUHWTWRUUHUULUUCUUOUUOOXDWNABUULUUMAQCBUVOAVSMUUTEABEXAGXBAUUNUUL
-      UUCRZUUMUULTUUQUWDAQCXCMUULUUCXEVCXFXGABCBEFEABEXHABCEFGXIXJZXKAUUNYINLZY
-      QYILZUUFUUKTUUQYOUWGABCXOMYQUUCUUDNYIVFUUDOXLWNXFAYQUUEYSAUUDYRPAYRUUDAUW
-      FYRUUDTYOYIUUCYKYPUUCOXMXNXPXQXRXSAYQUUAYSUUBAYRYNPAYIYMYKUGAYMYIAYLYILYM
-      YITABYIUWEXTYLYIYAYBXPYEXQAUUAYLYJUFZYQYJYLYFAUWBUUSBCVNSUWHYQTUWCUUTGBCY
-      CWNYDYGXSYH $.
+    limciccioolb $p
+               |- ( ph -> ( ( F |` ( A (,) B ) ) limCC A ) = ( F limCC A ) ) $=
+  ( vx co cfv wss a1i cr eqid cnt cmnf wcel wbr wceq cicc cioo ccnfld ctopn csn
```
wlammen, line 654449: Unexpected indentation change

wlammen (changes requested):

set.mm
benjub: Like for the following cases, this is indeed a bug of the rewrapping caused by `$p` ending a line, see https://github.com/metamath/metamath-exe/issues/93
There is a suggestion to fix it by @savask, maybe you can check if it is correct. When this is fixed that the new version of metamath is used by the CI, then proofs should be rewrapped as usual. I think this is not a problem, and when it is possible to have the whole statement `|- ... $=` on a single line, it is more readable to do so.

avekens: > Like for the following cases, this is indeed a bug of the rewrapping caused by `$p` ending a line, see [metamath/metamath-exe#93](https://github.com/metamath/metamath-exe/issues/93) There is a suggestion to fix it by @savask, maybe you can check if it is correct. When this is fixed that the new version of metamath is used by the CI, then proofs should be rewrapped as usual. I think this is not a problem, and when it is possible to have the whole statement `|- ... $=` on a single line, it is more readable to do so.

I do not agree that a whole (long) statement on a signle line is generally more readable. In the current example,

```
raltpd $p |- ( ph -> ( A. x e. { A , B , C } ps 
                       <-> ( ch /\ th /\ ta ) ) ) $=
```

would also be possible. At least, we should not force to squeeze whole statements into a single line.

benjub: I think when a statement fits on a single line, then it should be on a single line. Else, this opens the door to even more subjective decisions to make. But anyway, I can change this one if you want. Ultimately, one should have a program able to parse a statement as a tree (like mmj2 or metamath-knife, but unlike metamath.c) and wrap it according to some uniform rules.

@wlammen, @avekens : what are the requested changes in the present PR ?

avekens: > I think when a statement fits on a single line, then it should be on a single line. 

I would agree that if the statement including the label and the terminating $= fits on a single line, then it should be on a single line. This does not hold for ~raltpd, so we have a line break anyway. Therefore, we should split the statement in a sensible way, and not after the $p (as demanded also by @wlammen ).

benjub:
I basically reverted all the statement rewraps since I do not want to argue further.

Note however that the current conventions (https://us.metamath.org/mpeuni/conventions-comments.html) say "As for line wrapping in statements, we try to break lines before the most important token." (this was not written by me: you can recognize @david-a-wheeler's style). And note that the root token of "`|- ...`" is "`|-`", so the "current state" would impose to break the lines in consideration after "`$p`".  Anyway, I think we should eventually have an automatic line wrapping, even for statements, in order to have a normal form and not go back and forth with a lot of special cases.

avekens (approved):

wlammen (approved):

PR #2863, opened 2022-09-29
Add iset.mm summation through fsumcl2lem

jkingdon:
Some of the intuitionizing here is significant, in other cases much smaller, but on the whole it went fairly smoothly.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7782,6 +7782,14 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>seqabs</TD>
+  <TD><I>none</I></TD>
+  <TD>Although something along these lines could be proved,
+  finite sums are more naturally expressed with ` sum_ ` rather
+  than ` seq ` especially in iset.mm (see fsumabs ).</TD>
+</TR>
+
+<TR>
   <TD>df-prod and theorems using it</TD>
   <TD><I>none</I></TD>
   <TD>To define this, we will need to tackle all the issues analogous
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6295,11 +6295,10 @@ or ~ ssfiexmid </TD>
 
 <TR>
   <TD>seqcl</TD>
-  <TD>~ iseqcl</TD>
-  <TD>iseqcl requires that ` F ` be defined on ` ( ZZ>= `` M ) `
-  not merely ` ( M ... N ) ` as in seqcl . This is not a problem
-  when used on infinite sequences, but perhaps this requirement
-  could be relaxed if there is a need.</TD>
+  <TD>~ iseqcl , ~ iseqclt</TD>
+  <TD>~ iseqcl requires that ` F ` be defined on ` ( ZZ>= `` M ) `
+  not merely ` ( M ... N ) ` . This requirement is
+  relaxed somewhat in ~ iseqclt .</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7689,7 +7689,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 <TR>
   <TD>sumex</TD>
   <TD><I>none</I></TD>
-  <TD>This will need to replaced by suitable closure theorems.</TD>
+  <TD>This will need to be replaced by suitable closure theorems.</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


mmil.raw.html
```diff
@@ -7764,8 +7764,8 @@
 <TR>
   <TD>df-prod and theorems using it</TD>
   <TD><I>none</I></TD>
-  <TD>To define this, will need to tackle all the issues
-  with ~ df-isum plus some more around, for example, not equal to zero
+  <TD>To define this, will need to tackle all the issues analogous
+  to ~ df-isum plus some more around, for example, not equal to zero
```
benjub, line 7802: I know I'm a bit formal, but: "we will" / "one will" ?
A few lines above there is "to replaced" --> "to be replaced"

iset.mm
```diff
@@ -109039,6 +109039,15 @@ A Cauchy sequence (as defined here, which has a rate of convergence
     KZULUKKUGUJUIALZUIBLZKZUMUGUICDUJUPMABNUIABCSOUGUKUNULUOUFUEUKUNMUFUEEUKBAF
     CGHZALUNBAPUQUIABAQRTUAABPUBUCUKULUDT $.
 
+  $( The maximum of two integers is an integer.  (Contributed by Jim Kingdon,
+     27-Sep-2022.) $)
+  zmaxcl $p |- ( ( A e. ZZ /\ B e. ZZ )
+      -> sup ( { A , B } , RR , < ) e. ZZ ) $=
```
benjub, line 109045: worth stating indeed

mmil.raw.html
```diff
@@ -6295,11 +6295,10 @@
...
 <TR>
   <TD>seqcl</TD>
-  <TD>~ iseqcl</TD>
-  <TD>iseqcl requires that ` F ` be defined on ` ( ZZ>= `` M ) `
-  not merely ` ( M ... N ) ` as in seqcl . This is not a problem
-  when used on infinite sequences, but perhaps this requirement
-  could be relaxed if there is a need.</TD>
+  <TD>~ iseqcl , ~ iseqclt</TD>
+  <TD>~ iseqcl requires that ` F ` be defined on ` ( ZZ>= `` M ) `
+  not merely ` ( M ... N ) ` as in seqcl . This requirement is
```
benjub, line 6300: I guess there should be no space in `seqcl.` ? When you don't put tildes it's because it's a result of set.mm, is that it ? (same with `fsumabs )`). One way to disambiguate would be to write, e.g., `set.mm/seqcl`.

iset.mm
```diff
@@ -111628,6 +111628,52 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
...
+    $d A a f k m $.  $d A a f k x y $.  $d B a f m $.  $d B a f x y $.
+    $d S f k x y $.  $d a f k m ph $.  $d ph x y $.
+    fsumcllem.1 $e |- ( ph -> S C_ CC ) $.
+    fsumcllem.2 $e |- ( ( ph /\ ( x e. S /\ y e. S ) ) -> ( x + y ) e. S ) $.
+    fsumcllem.3 $e |- ( ph -> A e. Fin ) $.
+    fsumcllem.4 $e |- ( ( ph /\ k e. A ) -> B e. S ) $.
+    ${
+      fsumcl2lem.5 $e |- ( ph -> A =/= (/) ) $.
+      $( - Lemma for finite sum closures.  (The "-" before "Lemma" forces the
+         math content to be displayed in the Statement List - NM 11-Feb-2008.)
```
benjub, line 111641: Why does this "- Lemma" have a special treatment ? Is it necessary to create a third category of "half lemma" ?

benjub (approved):
Approved, once you've at least considered the comments.

mmil.raw.html
jkingdon: Those seem like improvements, will do.

mmil.raw.html
jkingdon: The layout of the table is mainly what says what is from set.mm and what is from iset.mm (left column set.mm, middle column iset.mm).

I guess I'll go with just removing "as in seqcl" as I'm not sure much is gained by repeating the name of the theorem.

iset.mm
jkingdon: I don't know (this was already there in set.mm) but I suppose because it is used in multiple theorems. Probably we should just call it a non-lemma.

Probably for another day, though, I'm a bit reluctant to try to come up with the right wording in this pull request.

jkingdon:
> Approved, once you've at least considered the comments.

Fair enough, will plan on merging this soon, probably once the tests pass.

PR #2864, opened 2022-09-29
Transformation of PART 18 cont.

avekens:
See also issue #2201:

* new theorem ~phlssphl
* new theorem ~cphsscph replaces ~sspphOLD (must be kept because it is still used)
* new theorem ~cphssphl replaces ~ssphlOLD (can be deleted because it is not used)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -2970,7 +2970,7 @@
 "cba" is used by "sspmlem".
 "cba" is used by "sspmval".
 "cba" is used by "sspn".
-"cba" is used by "sspph".
+"cba" is used by "sspphOLD".
 "cba" is used by "ssps".
 "cba" is used by "sspz".
 "cba" is used by "ubth".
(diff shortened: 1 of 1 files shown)


avekens:
Sorry, I think this PR is mixed up with my last PR - please look at the second commit for the latest changes.

benjub:
So sspphOLD is only used by hhssph, which is only used by hhsshl. Therefore, after revising the latter two to make use of sspph, we can put an "Obsolete as of 29-Sep-2022" in sspphOLD.  Is that correct ?  Do you have the energy to do that ?

avekens:
> So sspphOLD is only used by hhssph, which is only used by hhsshl. Therefore, after revising the latter two to make use of sspph, we can put an "Obsolete as of 29-Sep-2022" in sspphOLD. Is that correct ? Do you have the energy to do that ?

I had the same idea, and I tried it, but I was not successful. It was more complicated than it seemed to be. ~hhsshl does not only depend on ~hhssph, but also on ~hhssbn, whose analogon has also to be proven...

avekens:
> > So sspphOLD is only used by hhssph, which is only used by hhsshl. Therefore, after revising the latter two to make use of sspph, we can put an "Obsolete as of 29-Sep-2022" in sspphOLD. Is that correct ? Do you have the energy to do that ?
> 
> I had the same idea, and I tried it, but I was not successful. It was more complicated than it seemed to be. ~hhsshl does not only depend on ~hhssph, but also on ~hhssbn, whose analogon has also to be proven...

I think the main difficulty was the very different definitions of a "closed subspace": `CH`/`SH` vs. `CSubSp`.

benjub (approved):

PR #2865, opened 2022-09-30
Symbol CSubSp-->ClSubSp

benjub:
Change symbol for closed subspaces: CSubSp --> ClSubSp because "Cl" is standard for "closed" and "C" alone can be thought of as complete, compact, complex... (I ran into that after @avekens's recent work.)

I also clarified some comments of related definitions. Note that the "closed subspace" function currently works only for pre-Hilbert spaces and uses the property that they are precisely the subsets equal to their double orthogonal. This should be a theorem, not the definition, and the definition should be the standard one (and valid for all topological vector spaces (or topological modules if they are in set.mm, I don't remember). I'm going to open an Issue.

I also clarified the definition of "orthocomplement", which is defined for all subsets but deserves its name only when applied to closed subspaces.  So a more correct name would be "orthogonal".


set.mm
```diff
@@ -259089,34 +259089,40 @@ inner product (as a function) on the parent space.  (Contributed by NM,
...
-  ccss $a class CSubSp $.
+  ccss $a class ClSubSp $.
 
   $( Extend class notation with the Hilbert lattice. $)
   cthl $a class toHL $.
 
   ${
     $d h s x y $.
-    $( Define orthocomplement of a subspace.  (Contributed by NM,
+    $( Define the orthocomplement function in a given pre-Hilbert space: it
```
avekens, line 259106: It is a little bit confusing that there is no pre-Hilbert space in the definition, also it is mentioned in the comment. Maybe we should write "Define the orthocomplement function for a set `h` which usually is a pre-Hilbert space".

We often have very general definitions, and make them "meaningful" in theorems by providing appropriate antecedents.

avekens:
> Change symbol for closed subspaces: CSubSp --> ClSubSp because "Cl" is standard for "closed" and "C" alone can be thought of as complete, compact, complex... (I ran into that after @avekens's recent work.)

Very good!

set.mm
benjub: I made the change (although I'm not sure it adds much information or is better). As you say the definition is too general. A better definition would be
```
df-ocv $a |- ocv = ( h e. Prehil |-> ...
```

benjub: I just ran into https://us.metamath.org/mpeuni/ocvval.html and https://us.metamath.org/mpeuni/elocv.html, etc. whose comments say "pre-Hilbert space" whereas their statements are also not "guarded", that is, have antecedent/hypothesis of the form `e. V` instead of `e. PreHil`.

avekens: I think for consistency, the comments for these should be changed, too. And also for df-css and df-thl.
In general, I think there was a general agreement to keep definitions as general as possible. Maybe @digama0 or @david-a-wheeler can explain this further.

avekens:
> I also clarified some comments of related definitions. Note that the "closed subspace" function currently works only for pre-Hilbert spaces and uses the property that they are precisely the subsets equal to their double orthogonal. This should be a theorem, not the definition, and the definition should be the standard one (and valid for all topological vector spaces (or topological modules if they are in set.mm, I don't remember). I'm going to open an Issue.

I would really appreciate to have a definition which can be used for arbitrary topological spaces (df-topsp).

set.mm
benjub: I think there is really no need for that (actually, I think I should go back to my first version for df-ocv. There is really nothing gained, nor any information added, by commenting (for https://us.metamath.org/mpeuni/df-ocv.html)
```
Define the orthocomplement function in a given set (which usually is a pre-Hilbert space)
```
instead of
```
Define the orthocomplement function in a given pre-Hilbert space
```
when one can see that the statement has the form `df-ocv $a |- ocv = ( h e. _V |-> ...`.

(The definition is a bit more general than for pre-Hilbert spaces of course: it could be pre-Hilbert module or any extensible structure that has defined the slots used in the definition.)

benjub: To be honest, I think it's not that a big deal. Can we merge this now @avekens, and if you want, change the comments in a future PR and/or if we get an opinion from @digama0 ?

avekens (approved):
OK, the open issues about the definitions and their comments can be discussed further together with the issue about closed subspaces for arbitrary topological spaces.

PR #2867, opened 2022-10-01
Uniqueness and unique existence (2)

benjub:
Continue reordering after #2866.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,21 +27,24 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
-26-Sep-22 euequ1    euequi
-26-Sep-22 eueq1     eueqi
+ 1-Oct-22 eu5       dfeu        this is now a rederivation
+ 1-Oct-22 mo2v      dfmo        this is now a rederivation
+ 1-Oct-22 mo2       mof         "f-version" of df-mo
+26-Sep-22 euequ1    euequ
+26-Sep-22 eueq1     eueqi       inference associated with eueq
 17-Sep-22 zfnuleu   ---         deleted; use nulmo instead
-17-Sep-22 bm2.5ii   uniordint
+17-Sep-22 bm2.5ii   uniordint   more informative label
 17-Sep-22 bm1.1     ---         deleted; use axextmo instead
-13-Sep-22 epelc     epeli
-13-Sep-22 a1tru     trud
-13-Sep-22 trud      mptru
+13-Sep-22 epelc     epeli       inference associated with epelg
+13-Sep-22 a1tru     trud        deduction form of tru
+13-Sep-22 trud      mptru       modus ponens when minor is tru
  8-Sep-22 el2v2     elvd        moved from PM's mathbox to main
  8-Sep-22 exlimimdd             moved from ML's mathbox to main
  4-Sep-22 funresfunco           moved from AV's mathbox to main
  3-Sep-22 pwexd                 moved from GS's mathbox to main
  3-Sep-22 moimd                 moved from TA's mathbox to main
  3-Sep-22 csbvargi              moved from GM's mathbox to main
- 1-Sep-22 keepel    ifcli
+ 1-Sep-22 keepel    ifcli       inference associated with ifcl
 26-Aug-22 elv                   moved from PM's mathbox to main
 18-Aug-22 eel1111   syl1111anc  moved from AS's mathbox to main
 26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main
--- a/discouraged
+++ b/discouraged
@@ -5887,6 +5887,7 @@
 "erngset-rN" is used by "erngbase-rN".
 "erngset-rN" is used by "erngfmul-rN".
 "erngset-rN" is used by "erngfplus-rN".
+"eubidvOLDOLD" is used by "mobidvOLDOLD".
 "eueqOLD" is used by "moeqOLD".
 "exatleN" is used by "cdlema2N".
 "exidu1" is used by "cmpidelt".
@@ -15398,6 +15399,7 @@ New usage of "erngset-rN" is discouraged (3 uses).
 New usage of "estrresOLD" is discouraged (0 uses).
 New usage of "eu5" is discouraged (0 uses).
 New usage of "eubidvOLD" is discouraged (0 uses).
+New usage of "eubidvOLDOLD" is discouraged (1 uses).
 New usage of "eueqOLD" is discouraged (1 uses).
 New usage of "euexALT" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
@@ -16565,6 +16567,7 @@ New usage of "mndoissmgrpOLD" is discouraged (1 uses).
 New usage of "mndomgmid" is discouraged (3 uses).
 New usage of "mo2v" is discouraged (0 uses).
 New usage of "mobidvOLD" is discouraged (0 uses).
+New usage of "mobidvOLDOLD" is discouraged (0 uses).
 New usage of "moeqOLD" is discouraged (0 uses).
 New usage of "mptssALT" is discouraged (0 uses).
 New usage of "mpv" is discouraged (1 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -22442,40 +22506,32 @@ of the unique existential quantifier (note that ` y ` and ` z ` need not
...
-    $( Bound-variable hypothesis builder for uniqueness.  (Contributed by NM,
-       9-Jul-1994.)  (Revised by Mario Carneiro, 7-Oct-2016.) $)
-    nfeu1 $p |- F/ x E! x ph $=
+    $( Alternate proof of ~ nfeu1 .  Shorter but using ~ eu6 and a dummy
+       variable.  Note that ~ nfeu1 has no usage before ~ nfeu1ALT so that the
+       latter could be used, but ~ nfeu1 's proof is more systematic.
+       (Contributed by NM, 9-Jul-1994.)  (Revised by Mario Carneiro,
+       7-Oct-2016.)  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    nfeu1ALT $p |- F/ x E! x ph $=
```
wlammen, line 22526: In what way is this shorter?

set.mm
```diff
@@ -189193,7 +189241,7 @@ implemented as a function (a set of ordered pairs)".  The current
        sets: without this, ~ ipsstr and theorems like it will have many sethood
        assumptions, and may not even be usable in the empty context.  Instead,
        the sethood assumption is deferred until it is actually needed, e.g.,
-       ~ ipsbase , which requires that the base set is a set but not any of the
+       ~ ipsbase , which requires that the base set be a set but not any of the
```
wlammen, line 189244: ?

set.mm
benjub:
```
---------The proof of "nfeu1" (74 bytes) ends above this line.
---------The proof of "nfeu1ALT" (73 bytes) ends above this line.
```

set.mm
benjub: That was the `minor comment edits: subjunctive` of the commit message. I do not know German, but in English, see Paragraph 4 of https://www.grammar.com/situations-requiring-the-subjunctive-mood/

set.mm
wlammen: It seems nfeu1ALT is shorter, because it has 4 lines, and nfeu1 has 5.  This makes nfeu1ALT be the preferred candidate.

benjub: Yes, I wrote in its comment that it is shorter, and I also wrote why nfeu1 is preferred (more systematic, no dummy variable, and nfeu1ALT uses eu6 while nefeu1 is proved directly from the definition):
```
sh pr nfeu1
14   nfbii.1=df-eu $a |- ( E! x ph <-> ( E. x ph /\ E* x ph ) )
20     nfan.1=nfe1   $p |- F/ x E. x ph
23     nfan.2=nfmo1  $p |- F/ x E* x ph
24   nfxfr.2=nfan  $p |- F/ x ( E. x ph /\ E* x ph )
25 nfeu1=nfxfr   $p |- F/ x E! x ph
sh pr nfeu1ALT
17   nfbii.1=eu6    $p |- ( E! x ph <-> E. y A. x ( ph <-> x = y ) )
23     nfex.1=nfa1    $p |- F/ x A. x ( ph <-> x = y )
24   nfxfr.2=nfex   $p |- F/ x E. y A. x ( ph <-> x = y )
25 nfeu1ALT=nfxfr $p |- F/ x E! x ph
```

wlammen: From conventions:

If two proofs have the same number of compressed proof steps, we will typically give preference to the one with the smaller number of different labels

which in our case is nfeu1ALT.  It is of no importance whether it declares a dummy variable.  Shorter proofs also make better use of other theorems, so the use of eu6 is no downside either.  And the idea of systematic is disputable, too.  And is no criterium either.

set.mm
jkingdon: My native speaker intuition doesn't have a big problem with either one but I guess prefers `is`.

If I'm correctly reading that article it is saying to use `is` here, as `be` is archaic:

```
In today’s writing, we [rarely](https://www.definitions.net/definition/rarely) see the use of the main verb be to show the [subjunctive](https://www.definitions.net/definition/subjunctive) mood of am, is, are. [These](https://www.definitions.net/definition/These) uses are [archaic](https://www.definitions.net/definition/archaic) or form [traditional](https://www.definitions.net/definition/traditional) idioms:

    be that as it may …. If God be for us, who can be [against](https://www.definitions.net/definition/against) us? far be it from me ….
```

wlammen: I'd preferred the old version, but as always, my English is not good enough to decide anything.

set.mm
jkingdon: I suppose we should use the shorter version although I'm having trouble forming opinions either way.

benjub: I know it does not follow the convention, and that is the reason I wrote these explanations in the comment of nfeu1ALT. As for "systematic", it is precise here in the sense that nfeu1 follows the standard algorithm to prove nonfreeness in a defined expression (just regard the definiens as a tree whose nodes are formulas and prove by tree-induction nonfreeness at each node, from the leaves up to the root). Let's see if we can get an opinion from the originator of the nonfreeness predicate, @digama0. If we cannot (or obviously if he agrees with you), then I'll swap nfeu1 and nfeu1ALT.

set.mm
benjub: @jkingdon : you're quoting Paragraph 2 ("Situations contrary to fact") whereas I mentioned Paragraph 4 ("Suggestion and necessity"). In Paragraph 4, the example is
> This emergency requires that they be there.

(Note that in Paragraph 2, the preferred form is "were", e.g., "If I were you", "If god were for us", and not "am"/"is".)

Given that set.mm is a written reference work, I think it should't be too colloquial.  But if you think I'm really nitpicking then I'll change it back.  (I tried to make that correction in set.mm silently because I'm not a grammar nazi in the sense that I do not correct people for making similar kinds of mistakes in French, written or spoken; I try to keep being liberal in what I accept and conservative in what I produce!)

set.mm
wlammen: I don't know what to say here.  Maybe this: First, I think, you need to provide some material why the conventions fail - not in your case alone, but on a broader scale.  The material should convince the community that the conventions need an update.  Once they allow your move, then it will be fine with me.

benjub: I did provide a reason why this should be an exception, both in the initial comment of ~nfeu1ALT, and then with further precisions in the above comment.  This did not convince you, so I just swapped nfeu1 and nfeu1ALT in the latest commit, with appropriate comments.  I hope that's ok now.

Note that I had to move nfeu1(ALT) after eu6. That is, we first prove an equivalent definition (eu6) and from it we prove nonfreeness.  In the future, it'd be nice to automate the statement and proof of the "staple theorems" associated with each new definition (e.g., df-xxx is followed by the equality theorems like xxxeq1, xxxeq2, xxxbi1... the nonfreeness theorems when they apply, nfxxx1/nfxxx, the existence theorem when it applies, xxxeq, etc.).  I think that @digama0 implemented some of this in mmj2 (and probably plans it for MM0 ?). With this automation, the resulting proof of nfeu1 would be the one I proposed.

jkingdon (approved):

set.mm
jkingdon: > I just swapped nfeu1 and nfeu1ALT in the latest commit, with appropriate comments. I hope that's ok now.

Thanks. Like I said before, I don't really have an opinion one way or the other, but this seems like the expedient thing at least for now.

> In the future, it'd be nice to automate the statement and proof of the "staple theorems" associated with each new definition (e.g., df-xxx is followed by the equality theorems like xxxeq1, xxxeq2, xxxbi1... the nonfreeness theorems when they apply, nfxxx1/nfxxx, the existence theorem when it applies, xxxeq, etc.). I think that @digama0 implemented some of this in mmj2 (and probably plans it for MM0 ?). With this automation, the resulting proof of nfeu1 would be the one I proposed.

Oooh, interesting. This is getting my mind running about all the ways definitions could (hypothetically) be better (especially in a next generation system like ghilbert, MM0, or whatever it ends up being). But I suppose coming up with ideas has been the easy part of those efforts, but writing code and dealing with the annoying details has been the hard part. I guess I'll go back to dreaming at least for now.

set.mm
jkingdon: I'm fine with going with your preference here. As far as I can tell either one is grammatical.

set.mm
digama0: As a matter of fact, MM0 does generate equality theorems automatically for every definition, and it could generate nf theorems as well but it wasn't too important since I don't have many fancy binders that would need it.

wlammen (approved):

PR #2870, opened 2022-10-03
Transitive classes

benjub:
Commit messages say it all.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17909,6 +17909,7 @@ New usage of "tratrbVD" is discouraged (0 uses).
 New usage of "trelded" is discouraged (1 uses).
 New usage of "trintALT" is discouraged (0 uses).
 New usage of "trintALTVD" is discouraged (0 uses).
+New usage of "trintOLD" is discouraged (0 uses).
 New usage of "trintssOLD" is discouraged (0 uses).
 New usage of "trlcoabs2N" is discouraged (1 uses).
 New usage of "trnfsetN" is discouraged (1 uses).
@@ -19852,6 +19853,7 @@ Proof modification of "trcleq2lemRP" is discouraged (32 steps).
 Proof modification of "trelded" is discouraged (27 steps).
 Proof modification of "trintALT" is discouraged (166 steps).
 Proof modification of "trintALTVD" is discouraged (211 steps).
+Proof modification of "trintOLD" is discouraged (129 steps).
 Proof modification of "trintssOLD" is discouraged (67 steps).
 Proof modification of "trsbc" is discouraged (203 steps).
 Proof modification of "trsbcVD" is discouraged (398 steps).


set.mm
```diff
@@ -206874,14 +206907,14 @@ that F(A') is isomorphic to B'.".  Therefore, the category of sets and
       UIUQDVACSUJUKUSERZVRWAVHWCVPVTVQWCVOVIVBVSUSEDCTUSEVACSUKULUMVAFRZWAVMVHW
       DVTVKVQVLWDVSVJVIVAFECTUMVAFDCTUJUMUNUO $.
 
-    $( Less-or-equal is reflexive in a proset.  (Contributed by Stefan O'Rear,
-       1-Feb-2015.) $)
+    $( Less-than-or-equal-to is reflexive in a proset.  (Contributed by Stefan
```
avekens, line 206910: Why not "Less than or equal to" as above?

set.mm
```diff
@@ -707210,8 +707244,8 @@ Ordering on reals (cont.) - extension
 -.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
 $)
 
-  $( Addition and subtraction on one side of 'less or equal'.  (Contributed by
-     Alexander van der Vekens, 18-Mar-2018.) $)
+  $( Addition and subtraction on one side of "less-than-or-equal-to".
```
avekens, line 707247: Either "less than or equal to" or less-than-or-equal-to.

avekens (changes requested):
Only a mior remark: Either "less than or equal to" or less-than-or-equal-to should be used.

set.mm
benjub: Done.

set.mm
benjub: I don't know ! I found both ways in set.mm and wasn't sure which was best and if I was a bit too "dictatorial" in enforcing one or the other. What's important is that I fixed the few "less or equal to..." to "less than or equal to". If we want to enforce a consistent spelling, we can do it in a future PR.

avekens (approved):
Less-than-or-equal-to should be changed to "Less than or equal to" in a leter PR.

benjub:
> Less-than-or-equal-to should be changed to "Less than or equal to" in a leter PR.

Ok, I'm doing it now.

benjub:
Ok, I did it in the latest (hopefully last) commit. Though I'm not convinced by it. Now, there are still many occurrences of `less-than` that we should also change to `"less than"`, and there is also the `at-most-one quantifier`... I think we can leave it that way. At least we reduced these expressions to only two versions, so it's a bit easier to search.

avekens (approved):

PR #2871, opened 2022-10-04
Begin removing unneeded argument to seq in iset.mm

jkingdon:
Ever since https://us.metamath.org/ileuni/iseqsst.html was proved, it seemed likely that there was no need for the last (fourth, if you count M as well as .+ and F) argument to seq in https://us.metamath.org/ileuni/df-iseq.html . But it was only in the last day or two that I took a serious look at what would be involved. One of the key insights is that the three argument and four argument syntaxes can coexist, at least for a transitional period, with no more difficulty than we have from having both https://us.metamath.org/ileuni/df-pr.html and https://us.metamath.org/ileuni/df-tp.html

This pull request adds the following definition:

```
  df-seq3 $a |- seq M ( .+ , F ) = seq M ( .+ , F , _V ) $.
```

and starts the process of using this instead of the four argument version of seq.

There are a large number of affected theorems, so this will not be done all at once. But this pull request introduces the new definition and makes a dent on introducing theorems which use it and removing theorems which become no longer needed.

The pull request is probably easiest to read commit-by-commit. In particular, a lot of the affected lines only change `cseq` to `cseq4`, which is in its own commit. For the same reason, I would propose that whoever merges this uses "rebase and merge" rather than "squash and merge".

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -141,6 +141,10 @@
 "iseqss" is used by "iserile".
 "iseqss" is used by "serige0".
 "iseqss" is used by "serile".
+"iser0" is used by "iser0f".
+"iser0" is used by "isumz".
+"iser0" is used by "ser0f".
+"iser0f" is used by "iserclim0".
 "mo3h" is used by "mo2dc".
 "mo3h" is used by "mo3".
 "mo3h" is used by "mo4f".
@@ -278,6 +282,8 @@ New usage of "foelrnOLD" is discouraged (0 uses).
 New usage of "hbs1" is discouraged (5 uses).
 New usage of "idALT" is discouraged (0 uses).
 New usage of "iseqss" is discouraged (4 uses).
+New usage of "iser0" is discouraged (3 uses).
+New usage of "iser0f" is discouraged (1 uses).
 New usage of "mathbox" is discouraged (0 uses).
 New usage of "mo3h" is discouraged (7 uses).
 New usage of "nn0ge2m1nnALT" is discouraged (0 uses).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -137,10 +137,6 @@
 "hbs1" is used by "mopick".
 "hbs1" is used by "nfs1v".
 "hbs1" is used by "sb9v".
-"iseqss" is used by "climserile".
-"iseqss" is used by "iserile".
-"iseqss" is used by "serige0".
-"iseqss" is used by "serile".
 "iser0" is used by "iser0f".
 "iser0" is used by "isumz".
 "iser0" is used by "ser0f".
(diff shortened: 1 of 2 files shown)


benjub:
That's nice !  And thanks for doing that carefully and step by step.  So the new definition is closer to set.mm's but not quite the same yet (cf. `ran frec`) ?  Maybe you can put a short explanation of the difference in mmil.html ?  I saw that this PR modifies it, but did not see it (doing a ctrl+F on mmil.html for "seq " also shows that some lines should still be changed, it seems).  Obviously, I did not take time to search carefully, so please disregard if it's already there.  I'll try to review tomorrow.

jkingdon:
No modifications to https://us.metamath.org/ileuni/df-iseq.html here (that would be far less step-by-step given how many things depend on it). The change is to add `df-seq3` in addition (defined in terms of df-iseq at least for the short to medium term).

As for the differences between https://us.metamath.org/ileuni/df-iseq.html and https://us.metamath.org/mpeuni/df-seq.html there's only so much mileage to be gained by unifying them a lot more than they are already. At some point (either the definition itself or some of the early theorems you prove) you hit aspects of recursion and closure theorems thereof where set.mm is doing tricks which just won't work in iset.mm.

Last time I tackled recursion it sure was frustrating with tons of dead ends and things that I really didn't know how to approach. I did eventually make a lot of progress with things like https://us.metamath.org/ileuni/tfrcl.html which in fact is what enables the present pull request, but .... can't speak for anyone else but I've had my fill of that kind of work for a while. Perhaps someone will want to jump in with fresh enthusiasm and/or new approaches.

iset.mm
```diff
@@ -98980,9 +98980,12 @@ equinumerous to omega (the set of finite ordinal numbers).  (Contributed
 $)
 
   $c seq $.
-  $( Extend class notation with recursive sequence builder. $)
+  $( Extend class notation with four argument recursive sequence builder. $)
```
benjub, line 98983: four-argument

benjub (approved):

iset.mm
jkingdon: This could turn into one of those epic style guide things, but I will add the hyphen here and hopefully save us that trouble.

iset.mm
```diff
@@ -99018,6 +99021,12 @@ the previous value and the value of the input sequence (second operand).
        <. M , ( F ` M ) >. ) $.
   $}
 
+  $( Define a three-argument version of ` seq ` .  By theorems such as
+     ~ iseqsst , it should be capable of doing pretty much everything that the
+     four-argument version can, and may eventually replace the three-argument
```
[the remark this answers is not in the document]
jkingdon: Good catch, changed.

jkingdon (commented):
Planning on merging once tests pass.

iset.mm
benjub: Is that a thousand-year-old debate ? ;) (in the present case, this was more about consistency, and triggered by the other comment)

jkingdon: > Is that a thousand-year-old debate ? ;) (in the present case, this was more about consistency, and triggered by the other comment)

LOL, clever use of hyphens in "thousand-year-old"

PR #2872, opened 2022-10-04
reduce axioms on complex membership

icecream17:
Moves ~ 2nn ~ 3nn et al around the ~ 2cn proofs. They were previously in different sections.
Originally I thought that ~ 2nn + ~ nncni would be used for ~ 2cn et al
but proving directly with ~ eqeltri and ~ addcli used less axioms

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13287,6 +13287,7 @@ New usage of "1t1e1ALT" is discouraged (0 uses).
 New usage of "2atm2atN" is discouraged (0 uses).
 New usage of "2atnelvolN" is discouraged (0 uses).
 New usage of "2bornot2b" is discouraged (0 uses).
+New usage of "2cnOLD" is discouraged (0 uses).
 New usage of "2llnjN" is discouraged (1 uses).
 New usage of "2llnjaN" is discouraged (1 uses).
 New usage of "2llnm2N" is discouraged (1 uses).
@@ -13331,6 +13332,7 @@ New usage of "3anidm12p2" is discouraged (0 uses).
 New usage of "3anorOLD" is discouraged (0 uses).
 New usage of "3anrotOLD" is discouraged (0 uses).
 New usage of "3atnelvolN" is discouraged (2 uses).
+New usage of "3cnOLD" is discouraged (0 uses).
 New usage of "3com12OLD" is discouraged (0 uses).
 New usage of "3com13OLD" is discouraged (0 uses).
 New usage of "3com23OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen (approved):
The reductions are massive: 3cn (old) is based on following additional axioms: ax-10 ax-11 ax-12 ax-13 ax-resscn ax-icn ax-addrcl ax-mulcl ax-mulrcl ax-i2m1 ax-1ne0 ax-rrecex ax-cnre.

The same for 4cn to 9cn.

Similar for nnsscn nncni and 2cn, although to a lesser degree.

benjub:
Thanks @icecream17 for these reductions, and thanks @wlammen for the detailed review. Maybe it's worth indicating these removed dependencies explicitly in the (or at least some) comments ?

It may be worth keeping at least some OLD proofs as ALT since some are much shorter ? E.g., keep 2cnOLD as 2cnALT and write in its comment that the same can be done with 3cn, etc.

If I understand correctly, the reordering is so that we have 1nn, 1re, 1cn, 2nn, 2re, 2cn... instead of 1nn, 2nn, 3nn..., 1re, 2re, 3re..., 1cn, 2cn, 3cn... ? I don't know which is best.

icecream17:
I'm not sure which of those above orderings are better either.

Though the actual old ordering was:

#### 5.4.3 Decimal representation of numbers

...
2re
2cn
3re
3cn
...

#### 5.4.4 Some properties of specific numbers

...
2nn
3nn
4nn
...

benjub:
Thanks @icecream17 for the precision. What about my other two remarks ? Let's hear @wlammen's opinions too. If you two think it's not necessary, then so be it.

wlammen:
I already approved this pull request.  Any modification to it like keeping ALT theorems are non-blocking, and can IMO be added in a separate contribution.  Given the really massive reduction in axiom usage, I suspect that not only the 10 and odd theorems here allow for such improvements, but that the whole environment they are embedded in after some thorough inspection will not look the same again.  So @icecream17 's changes here are perhaps the first promising steps in a long row.

icecream17:
I'll do these changes in a separate pr.

wlammen:
merge based on the last comments of @icecream17 and @benjub

benjub:
> I'll do these changes in a separate pr.

Yes please @icecream17. After seeing the webpages, I think they are indeed necessary (2cnALT and nnsscnALT).

icecream17:
@benjub What would you suggest for the comments of 2cnALT and nnsscnALT? I don't think nnsscnALT is necessary since I think their axiom saves as equivalent (use axioms of complex numberhood directly instead of proving realness then using ax-resscn)

<sup>(The new pr is taking awhile because of in real life things. I also have ideas for 0re nn0sscn nn0cni 10re <s>and ltxrlt</s>)</sup>

benjub:
2cnALT:
```
$( Alternate proof of ~ 2cn .  Shorter but uses more axioms.  Similar proofs
   are possible for ~ 3cn , ... , ~ 9cn .  (Contrib...
```

nnsscnOLD: ok, keep it as OLD, but write for nnsscn:
```
$( The positive integers form a subset of the complex numbers.  Remark: this
   could also be proven from ~ nnssre and ~ ax-resscn at the cost of using
   more axioms.  (Contrib...
```

benjub:
@icecream17: the first step of the proof https://us.metamath.org/mpeuni/0re.html is ax-1cn but the comment says "the first step could also be ax-1cn".  Should that remark be removed ?

icecream17:
Oops, that's a typo. I meant ax-icn

Edit: #2892

Issue #2861, opened 2022-09-27
Finer sectioning of set.mm


benjub:
(The "finer" of the title has to be read as the opposite of "coarser".)  There are a few sections in set.mm where several definitions are introduced "en masse", and then their properties proven altogether.  This entangles results and makes it harder to organize them, which makes searches more difficult in huge sections (this is not an abstract concern, actually on several occasions it was painful to retrieve some results and see what depends on what, especially in the sections related to binary relations and functions).  Unless there is opposition, I will try and untangle them, and organize them into subsections (slowly, one section at a time, and giving a heads up here before each PR).  A fourth level of nested sectioning was introduced a few years ago, and this will allow this finer sectioning.  In some cases, additional nesting will not be necessary, and it will be more natural to "flatten" the TOC.  I did this for a few sections in the past, and recently a huge work was done by @wlammen to organize conjunction and disjunction into separate sections (see https://us.metamath.org/mpeuni/mmtheorems.html#dtl:1.2) and one can see the benefits, or at least the absence of disruption if your work is on more advanced parts.

I will begin small: I plan to organize the section introducing the unique existential quantifier and the at-most-one quantifier into two subsections.  One caveat: I plan to introduce first the at-most-one quantifier, and then the unique existential quantifier, defined as the conjunction of existence and uniqueness (that is, make https://us.metamath.org/mpeuni/mo2v.html the definition of the at-most-one quantifier and https://us.metamath.org/mpeuni/eu5.html the definition of the unique existential quantifier), and make the current definitions theorems.  This is more natural than the current definition of at-most-one as "existence implies unique existence" (just think of your usual ways of proving unique existence and of proving uniqueness).  This will probably go with a slight shortening of the combined proof length.

Any thoughts?

(**edit:** corrected the worst typo ever... thanks @david-a-wheeler !)

david-a-wheeler:
I like this idea in general. Thank you for being willing to do it.

There may be a few cases where it's easier to introduce a pair of ideas, or a small set. So I suggest starting with areas where they are *obviously* entangled in an unnecessary way. As always, please do the untangling in *separate* pull requests, so if there's an agreement on one and not the other, we can immediately accept the one and focus discussion on the other.

> Unless there is opposition, I will try and entangle them

I presume you mean "untangle" here :-).

No opposition here.

wlammen:
[...]

PR #2796, opened 2022-08-25
remove duplicate snunioo2 in favor of ioounsn

benjub:
This is basically #2778, which was closed because of merge conflicts.
A duplicate from the list at #2128.

PR #2873, opened 2022-10-04
brres and related theorems

mazsa:
Finishing the project which was initiated in PR 2831 along the logic of the conversation
https://github.com/metamath/set.mm/pull/2831#pullrequestreview-1109220740
with revision suggested by @benjub

Revision of ~ brinxp is ready by the PR 2831 (and now I filled in missing lines in changes-set.txt),
revision of ~ brres is the topic of this PR.

Original ~ brresALTV and ~ opelresALTV was renamed to ~ brres and ~ opelres,
original ~ brres and ~ opelres was renamed to ~ brresi and ~ opelresi
(and original ~ opelresi was renamed to ~ opelidres).

Some (cca. 1/3 of the 100+) related theorems was minimized, at least with ~ ancom ~ ancoms ~ ancomd ~ biancom or similar.
We could shorten ~ relexpindlem further with ~ syl2an2r but I did not find its place.
In one case, ~ dvres, I minimally changed, and shortened, its lemma, ~ dvreslem .
In one case, ~ setrec2lem2, I left a theorem in its proof, ~ tz6.12-1 , unchanged, which caused 1 step longer proof with ~ ancoms.
It was new for me that minimize_with * /include_mathboxes does not find all options:
in case of ~ relexpindlem, I shortened ~ jca and ~ mpd176 with ~ mp2and, but this was not suggested by the program.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,14 +27,36 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 4-Oct-22 brresi    brresi2
+ 4-Oct-22 opelresi  opelidres
+ 4-Oct-22 brresg    brresgOLD2
+ 4-Oct-22 opelresg  opelresgOLD2
+ 4-Oct-22 brres     brresOLD2
+ 4-Oct-22 brres     brresi
+ 4-Oct-22 opelres   opelresOLD2
+ 4-Oct-22 opelres   opelresi
+ 4-Oct-22 brresALTV brres       moved from PM's mathbox to main set.mm
+ 4-Oct-22 opelresALTV opelres   moved from PM's mathbox to main set.mm
+ 4-Oct-22 inxpssres [same]      moved from PM's mathbox to main set.mm
+ 4-Oct-22 biancomd  [same]      moved from PM's mathbox to main set.mm
+ 4-Oct-22 biancomi  biancom     moved from PM's mathbox to main set.mm
+ 4-Oct-22 dvreslem  [same]      commutation in theorem
  1-Oct-22 eu5       dfeu        this is now a rederivation
  1-Oct-22 mo2v      dfmo        this is now a rederivation
  1-Oct-22 mo2       mof         "f-version" of df-mo
 26-Sep-22 euequ1    euequ
 26-Sep-22 eueq1     eueqi       inference associated with eueq
+18-Sep-22 brinxp2ALTV brinxp2   moved from PM's mathbox to main set.mm
+18-Sep-22 brinxp2   brinxp2OLD
 17-Sep-22 zfnuleu   ---         deleted; use nulmo instead
 17-Sep-22 bm2.5ii   uniordint   more informative label
 17-Sep-22 bm1.1     ---         deleted; use axextmo instead
+14-Sep-22 idssxp    [same]      proved by pre-function definitions
+14-Sep-22 idssxp    idssxpOLD 
+14-Sep-22 idinxpres [same]      moved from PM's mathbox to main set.mm
+14-Sep-22 relinxp   [same]      moved from PM's mathbox to main set.mm
+14-Sep-22 opelinxp  [same]      moved from PM's mathbox to main set.mm
+14-Sep-22 ralanid   [same]      moved from PM's mathbox to main set.mm
 13-Sep-22 epelc     epeli       inference associated with epelg
 13-Sep-22 a1tru     trud        deduction form of tru
 13-Sep-22 trud      mptru       modus ponens when minor is tru
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -29,15 +29,13 @@ DONE:
 Date      Old       New         Notes
  4-Oct-22 brresi    brresi2
  4-Oct-22 opelresi  opelidres
- 4-Oct-22 brresg    brresgOLD2
- 4-Oct-22 opelresg  opelresgOLD2
- 4-Oct-22 brres     brresOLD2
  4-Oct-22 brres     brresi
- 4-Oct-22 opelres   opelresOLD2
  4-Oct-22 opelres   opelresi
  4-Oct-22 brresALTV brres       moved from PM's mathbox to main set.mm
  4-Oct-22 opelresALTV opelres   moved from PM's mathbox to main set.mm
  4-Oct-22 inxpssres [same]      moved from PM's mathbox to main set.mm
+ 4-Oct-22 anbi1cd   [same]      moved from PM's mathbox to main set.mm
+ 4-Oct-22 anbi1ci   [same]      moved from PM's mathbox to main set.mm
  4-Oct-22 biancomd  [same]      moved from PM's mathbox to main set.mm
  4-Oct-22 biancomi  biancom     moved from PM's mathbox to main set.mm
  4-Oct-22 dvreslem  [same]      commutation in theorem
@@ -47,12 +45,9 @@ Date      Old       New         Notes
 26-Sep-22 euequ1    euequ
 26-Sep-22 eueq1     eueqi       inference associated with eueq
 18-Sep-22 brinxp2ALTV brinxp2   moved from PM's mathbox to main set.mm
-18-Sep-22 brinxp2   brinxp2OLD
 17-Sep-22 zfnuleu   ---         deleted; use nulmo instead
 17-Sep-22 bm2.5ii   uniordint   more informative label
 17-Sep-22 bm1.1     ---         deleted; use axextmo instead
-14-Sep-22 idssxp    [same]      proved by pre-function definitions
-14-Sep-22 idssxp    idssxpOLD 
 14-Sep-22 idinxpres [same]      moved from PM's mathbox to main set.mm
 14-Sep-22 relinxp   [same]      moved from PM's mathbox to main set.mm
 14-Sep-22 opelinxp  [same]      moved from PM's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -2703,6 +2703,7 @@
 "braval" is used by "kbass2".
 "braval" is used by "kbass6".
 "braval" is used by "rnbra".
+"brresOLD2" is used by "funressnvmoOLD".
 "brresgOLD2" is used by "brresOLD2".
 "c-bnj14" is used by "bnj1000".
 "c-bnj14" is used by "bnj1006".
@@ -14261,7 +14262,7 @@ New usage of "brecop2OLD" is discouraged (0 uses).
 New usage of "brfi1indALT" is discouraged (0 uses).
 New usage of "brinxp2OLD" is discouraged (0 uses).
 New usage of "brresOLD" is discouraged (0 uses).
-New usage of "brresOLD2" is discouraged (0 uses).
+New usage of "brresOLD2" is discouraged (1 uses).
 New usage of "brresgOLD2" is discouraged (1 uses).
 New usage of "c-bnj14" is discouraged (131 uses).
 New usage of "c-bnj18" is discouraged (58 uses).
@@ -15501,6 +15502,7 @@ New usage of "funcrngcsetcALT" is discouraged (0 uses).
 New usage of "funiedgdm2valOLD" is discouraged (1 uses).
 New usage of "funiedgdmge2valOLD" is discouraged (1 uses).
 New usage of "funiedgvalOLD" is discouraged (0 uses).
+New usage of "funressnvmoOLD" is discouraged (0 uses).
 New usage of "funsneqopOLD" is discouraged (0 uses).
 New usage of "funsneqopsnOLD" is discouraged (1 uses).
 New usage of "funvtxdm2valOLD" is discouraged (2 uses).
@@ -19208,6 +19210,7 @@ Proof modification of "funcrngcsetcALT" is discouraged (765 steps).
 Proof modification of "funiedgdm2valOLD" is discouraged (68 steps).
 Proof modification of "funiedgdmge2valOLD" is discouraged (56 steps).
 Proof modification of "funiedgvalOLD" is discouraged (57 steps).
+Proof modification of "funressnvmoOLD" is discouraged (87 steps).
 Proof modification of "funsneqopOLD" is discouraged (25 steps).
 Proof modification of "funsneqopsnOLD" is discouraged (74 steps).
 Proof modification of "funvtxdm2valOLD" is discouraged (68 steps).
--- a/discouraged
+++ b/discouraged
@@ -2703,6 +2703,7 @@
 "braval" is used by "kbass2".
 "braval" is used by "kbass6".
 "braval" is used by "rnbra".
+"brresgOLD2" is used by "brresOLD2".
 "c-bnj14" is used by "bnj1000".
 "c-bnj14" is used by "bnj1006".
 "c-bnj14" is used by "bnj1020".
@@ -10820,6 +10821,12 @@
 "opelreal" is used by "axrnegex".
 "opelreal" is used by "axrrecex".
 "opelreal" is used by "ltresr".
+"opelresOLD2" is used by "brresOLD".
+"opelresOLD2" is used by "elresOLD".
+"opelresOLD2" is used by "opelresgOLD".
+"opelresgOLD2" is used by "brresgOLD2".
+"opelresgOLD2" is used by "idrefOLD".
+"opelresgOLD2" is used by "opelresOLD2".
 "opeqsnOLD" is used by "snopeqopOLD".
 "opidon2OLD" is used by "exidreslem".
 "opidonOLD" is used by "opidon2OLD".
@@ -14254,6 +14261,8 @@ New usage of "brecop2OLD" is discouraged (0 uses).
 New usage of "brfi1indALT" is discouraged (0 uses).
 New usage of "brinxp2OLD" is discouraged (0 uses).
 New usage of "brresOLD" is discouraged (0 uses).
+New usage of "brresOLD2" is discouraged (0 uses).
+New usage of "brresgOLD2" is discouraged (1 uses).
 New usage of "c-bnj14" is discouraged (131 uses).
 New usage of "c-bnj18" is discouraged (58 uses).
 New usage of "calemesOLD" is discouraged (0 uses).
@@ -17001,7 +17010,9 @@ New usage of "opelopab4" is discouraged (0 uses).
 New usage of "opelopabsbALT" is discouraged (3 uses).
 New usage of "opelreal" is discouraged (8 uses).
 New usage of "opelresOLD" is discouraged (0 uses).
+New usage of "opelresOLD2" is discouraged (3 uses).
 New usage of "opelresgOLD" is discouraged (0 uses).
+New usage of "opelresgOLD2" is discouraged (3 uses).
 New usage of "opeqsnOLD" is discouraged (1 uses).
 New usage of "opidon2OLD" is discouraged (1 uses).
 New usage of "opidonOLD" is discouraged (2 uses).
@@ -18622,6 +18633,8 @@ Proof modification of "brfi1indALT" is discouraged (789 steps).
 Proof modification of "brfvidRP" is discouraged (92 steps).
 Proof modification of "brinxp2OLD" is discouraged (58 steps).
 Proof modification of "brresOLD" is discouraged (43 steps).
+Proof modification of "brresOLD2" is discouraged (26 steps).
+Proof modification of "brresgOLD2" is discouraged (46 steps).
 Proof modification of "calemesOLD" is discouraged (26 steps).
 Proof modification of "calemosOLD" is discouraged (31 steps).
 Proof modification of "camestresOLD" is discouraged (23 steps).
@@ -19529,7 +19542,9 @@ Proof modification of "onfrALTlem5VD" is discouraged (320 steps).
 Proof modification of "opelopab4" is discouraged (69 steps).
 Proof modification of "opelopabsbALT" is discouraged (106 steps).
 Proof modification of "opelresOLD" is discouraged (54 steps).
+Proof modification of "opelresOLD2" is discouraged (26 steps).
 Proof modification of "opelresgOLD" is discouraged (58 steps).
+Proof modification of "opelresgOLD2" is discouraged (85 steps).
 Proof modification of "opeqsnOLD" is discouraged (124 steps).
 Proof modification of "opidon2OLD" is discouraged (80 steps).
 Proof modification of "opidonOLD" is discouraged (198 steps).


set.mm
```diff
@@ -50673,6 +50733,14 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
     ( wss ccnv cdm crn cnvss dmss syl df-rn 3sstr4g ) ABCZADZEZBDZEZAFBFLMOCNPC
     ABGMOHIAJBJK $.
 
+  ${
+    rnssi.1 $e |- A C_ B $.
+    $( Subclass inference for range.  (Contributed by Peter Mazsa,
+       24-Sep-2022.) $)
+    rnssi $p |- ran A C_ ran B $=
+      ( wss crn rnss ax-mp ) ABDAEBEDCABFG $.
+  $}
```
benjub, line 50742: Used only once ? You should try to `scripts/min.cmd` to `minimize_with rnssi` theorems currently using `rnss`, and if (and only if) it decreases proofs by a total of more bytes than it takes to state and prove rnssi, then add it (and for ease of reviewing, in its own PR or at least its own commit).

set.mm
```diff
@@ -49040,6 +49094,12 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
     ( cxp cin wbr wcel wa cop brinxp2 df-br anbi2i 3bitr3i ) CDEABFGZHCAIDBIJZC
     DEHZJCDKZPIQSEIZJABCDELCDPMRTQCDEMNO $.
 
+  $( Intersection with a Cartesian product is a subclass of restriction.
+     (Contributed by Peter Mazsa, 19-Jul-2019.) $)
+  inxpssres $p  |- ( R i^i ( A X. B ) ) C_ ( R |` A ) $=
+    ( cxp cin cvv cres wss ssid ssv xpss12 mp2an sslin ax-mp df-res sseqtr4i )
+    CABDZEZCAFDZEZCAGQSHZRTHAAHBFHUAAIBJAABFKLQSCMNCAOP $.
```
benjub, line 49101: Nice theorem. Place it after xpss ?
(Well, actually that subsection is huge and ordering it is too difficult. That behemoth was my main target in #2861 but this will take time...)

benjub:
Thaks @mazsa. The main commit (d642c7c) is pretty large, and the changes are quite important. This may take some time to review.  Separate commits might have made it easier. (But since I've also been guilty of that, I cannot be too demanding...)

@wlammen could give good advice for the propcalc and fol parts of the PR, and @avekens about the label changes (and of course, both for the rest of the PR too).
At first glance, I see that some proofs have been lengthened and not shortened ? Maybe this was in order to avoid some definitions ? In that case, both proofs should be kept, using ALT (and only when avoiding a given definition has a special interest, since in general we do not shy away from using definitions: they are made for it). Generally, the shorter proof should be the main one, and the longer proof should be the ALT one (according to me, there could be exceptions to that rule if they are justified and explained in the comment... @wlammen is stricter !).
I'll try to have a closer look tomorrow or in the following days.

set.mm
mazsa: I like this policy: do not introduce an alternate form to the default (general) form of a theorem until you prove that its net effect to the size of the database is not positive. Based on this maybe we can even delete some inference forms and reduce the size of the database as well! I think @wlammen would like this (in the context of https://github.com/metamath/set.mm/pull/2867#pullrequestreview-1127557358 ), couldn't we make this an explicit convention?
I'll have time to experiment with this at the weekend.

mazsa:
You are right: some is longer, or, more precisely, some have longer compressed proofs. This was surprising, because, as I wrote, with the exception of ~ setrec2lem2 which was 1 ~ ancoms longer, *all proofs have less or equal steps*.

The reason of this, I think in all cases, that I prefer the default (or general) form (e.g. the new ~ brres ) over the inference form (so much that I do not use inference forms in my already big Mathbox at all: ~ elv ~ el2v and ~ el3v and the default form with $p |- ( A e. V -> ... etc. was enough for me, because I created short theorems and more complicated theorems was built from them, and for short theorems the default general form with ~ elv has the same number of steps than the inference form with ~ vex ). And, in all cases of this PR when I found an inference form AND it did not imply more proof steps, I exchanged it for the general form.

To my surprise, same number of steps increased the size of the compressed proofs: ~ elv in compressed proofs comes with friends:), for example in ~ fvres , it comes with "wa wb cvv", wich is not the case with "vex" in case of proofs using the inference form.

Compare:
in the present database, with vex:
```metamath
( vx wcel cv cres wbr cio cfv vex brres rbaib iotabidv df-fv 3eqtr4g ) AB
      EZADFZCBGZHZDIARCHZDIASJACJQTUADTUAQARCBDKLMNDASODACOP $.
```
my new suggestion with elv (and wa wb cvv, longer):
```metamath
( vx wcel cv cres wbr cio cfv wa wb cvv brres baib iotabidv df-fv 3eqtr4g
      elv ) ABEZADFZCBGZHZDIAUACHZDIAUBJACJTUCUDDUCTUDUCTUDKLDBAUACMNSOPDAUBQDA
      CQR $.
```
potential new with vex (short again)
```metamath
( vx wcel cres wbr cio cfv vex brresi baib iotabidv df-fv 3eqtr4g cv ) AB
      EZADPZCBFZGZDHARCGZDHASIACIQTUADTQUABARCDJKLMDASNDACNO $.
```

@benjub do you know what is the reason of this lengthening of compressed form in case of default form + ~ elv? (I dont´t understand the way of compression.)

I think by default we should use the general form if it does not imply more steps, even if it increases the size of compressed proofs. And only when the sum of the increases of different proofs are above the threshold should we allow to use alternate form of the general theorem. If we accept the Net Nonpositive Database Effect policy above then I'll happily use inference forms as well.

set.mm
benjub: Right after ~xpss is an appropriate location.

set.mm
benjub: I don't know whether it is written in ~conventions (if not, I'll add it), but it has been a rule of thumb for quite some time, initiated by Norm, who used the phrase "[the lemma] pays for itself".  I would consider it as a sufficient condition for moving a lemma to Main (but not a necessary condition: a statement can be interesting in itself, without being widely used). Also, I would be more reluctant to remove a theorem from Main, even if I would not have introduced it in the first place (that is: "now that it's here, let it be").

benjub:
Conflicts are minimal, and only in changes-set.txt:
```
<<<<<<< mazsa-brres
 4-Oct-22 brresi    brresi2
 4-Oct-22 opelresi  opelidres
 4-Oct-22 brresg    brresgOLD2
 4-Oct-22 opelresg  opelresgOLD2
 4-Oct-22 brres     brresOLD2
 4-Oct-22 brres     brresi
 4-Oct-22 opelres   opelresOLD2
 4-Oct-22 opelres   opelresi
 4-Oct-22 brresALTV brres       moved from PM's mathbox to main set.mm
 4-Oct-22 opelresALTV opelres   moved from PM's mathbox to main set.mm
 4-Oct-22 inxpssres [same]      moved from PM's mathbox to main set.mm
 4-Oct-22 biancomd  [same]      moved from PM's mathbox to main set.mm
 4-Oct-22 biancomi  biancom     moved from PM's mathbox to main set.mm
 4-Oct-22 dvreslem  [same]      commutation in theorem
=======
 3-Oct-22 brrelexi  brrelex1i
 3-Oct-22 brrelex   brrelex1
>>>>>>> develop
```

set.mm
```diff
@@ -5631,6 +5662,15 @@ use disjunction (although this is not required since definitions are
   anbi2 $p |- ( ( ph <-> ps ) -> ( ( ch /\ ph ) <-> ( ch /\ ps ) ) ) $=
     ( wb id anbi2d ) ABDZABCGEF $.
 
+  ${
+    anbi1cd.1 $e |- ( ph -> ( ps <-> ch ) ) $.
+    $( Deduction adding a left conjuct to the left side and a right conjunct to
```
benjub, line 5667: Introduce a proposition as left conjunct on the left-hand side and right conjunct on the right-hand side of an equivalence. Deduction form.

set.mm
```diff
@@ -6167,6 +6212,15 @@ use disjunction (although this is not required since definitions are
       ( wa biantrurd bitr4d ) ABCDGDFACDEHI $.
   $}
 
+  ${
+    mpbiran2d.1 $e |- ( ph -> th ) $.
+    mpbiran2d.2 $e |- ( ph -> ( ps <-> ( ch /\ th ) ) ) $.
+    $( Detach truth from conjunction in biconditional.  (Contributed by Peter
```
benjub, line 6218: Add: `Deduction form.`

benjub:
I'm a bit overwhelmed. So, the main changes are that you changed opelres/brres to commute the conjunction in the consequent ? And the rest is only the consequential adaptation of the proofs using them ? Why did you do that commutation ? I do not see an obvious winner between both forms, and it looks like half the usages will use it directly and the other half will need to commute the consequent anyway. (We may be able to have partial conventions regarding that, in order to globally shorten proofs, but I think this would need automation and we are not there yet.)

You ask above about compressed size. This is related to the total number of steps. To see a complete proof, do `MM>show proof <label-match>/all`. You will see that one can have proofs with fewer essential (i.e., non-syntactic) steps but more steps in total. E.g., if you apply ax-mp with $\varphi$ replaced by a huge expression, then there will be many syntactic steps to "construct" that formula (i.e., to show that the expression is indeed a (well-formed) formula).

As for inference form/closed form/general form, there are some rules of thumb in ~conventions. I think they should not be taken too strictly and one should have a pragmatic view, and if there is a doubt, have several forms.

avekens:
I share the concerns of @benjub that some (or many?) proofs become longer, without understanding why (and for which benefit).
Because of the huge amount of changes, I could not follow the intention of most of the changes: there are new theorems, revised theorems, and moved theorems. Therefore, I can neither reject nor approve this PR.

set.mm
jkingdon: The wording in conventions is "Usually we will also automatically accept a _new_ theorem that is used to shorten multiple proofs, if the total size of set.mm (including the comment of the new theorem, not including the acknowledgment) decreases as a result. "

mazsa:
I did some corrections and changes, the results may demonstrate that this new form of ~ brres may be desirable:

1. All of the the 100+ proofs, with one ~ setrec2lem2 exception, have *less or equal* steps than the old proofs, and cca. 1/3 of theorems have strictly less steps.

2. @avekens now all of the the theorems have *shorter or same size* compressed proofs (except, of the non discouraged proofs, some 1 characters, mostly where I corrected the labels of the inference form of the default theorem, from the present brres/opelres to brresi/opelresi: do you think I should rename these inference forms of the default theorem something shorter? and additionally ~ resco ) than the old theorems. It seems we should always consider to use of the inference form of a theorem if we already have it --- even if we would prefer using the default general form.

3. @benjub new proofs has *more than an one order of magnitude less* ancom-related steps than the old ones (deleted 20+ and added 1 \*ancom\* ).

And about the reason of this:

1. We should ask @digama0 why he used ~ ancom in the proof of ~ dfres2
```metamath
dfres2 $p |- ( R |` A ) = { <. x , y >. | ( x e. A /\ x R y ) }
```
only to get similar result to new suggested form of ~ brres
```metamath
brres $p |- ( C e. V -> ( B ( R |` A ) C <-> ( B e. A /\ B R C ) ) )
```
from the old form ~ brresOLD2  ? It was not necessary for him and yet he used it. But why then?

2. We should think about why Norman *unavoidably* used ~ ancom in the proof of
```metamath
dfima2 $p |- ( A " B ) = { y | E. x e. B x A y }
```
How could we sistematically avoid using ~ ancom in this E. x x ( x A y /\ x e. B ) situation (which have many instances in my Mathbox)?

3. Btw, if we refused this PR I had to litter my mathbox with a *lot* of ~ ancom , ~ df-3an and similar absolutely unnecessary and easily avoidable steps due to the old ' E. x x ( x R y /\ x e. A ) ' and ' E. x E. y ( x e. A /\ x e. B /\ x R y ) ' situations (which I avoided until now only by using the new brres and brtxp as brresALTV and brtxpALV ),

4. because we can accept brinxp2
```metamath
brinxp2 $p |- ( C ( R i^i ( A X. B ) ) D <-> ( ( C e. A /\ D e. B ) /\ C R D ) )
```
iff we accept this new suggested form of ~brres
```metamath
brres $p |- ( C e. V -> ( B ( R |` A ) C <-> ( B e. A /\ B R C ) ) )
```
(if we want to keep consistency in set.mm). If we refused this PR somebody should change ~ brinxp and all related theorems as well (to what?).

5. But even if we return to the old ~ brxp2 (now
```metamath
brinxp2OLD $p |- ( A ( R i^i ( C X. D ) ) B <-> ( A e. C /\ B e. D /\ A R B ) )
```
it has the same structure ( ' A e. C /\ B e. D ' in front and ' A R B ' behind ) than the new suggested ~ brres . Or you want to solve this inconsistency by defining a brand new ~ brinxp similar to the old ~ brres , with ' A e. C /\ B e. D ' behind?

6. If we refused this PR should we swap ' x e. A ' and ' x R y ' in
```metamath
inxprnres $p |- ( R i^i ( A X. ran ( R |` A ) ) ) = { <. x , y >. | ( x e. A /\ x R y ) }
```
to be similar to ~ brresOLD2 old or should we keep to be similar to ~ dfres2 ?
etc. etc. And

7. I would not rush with this PR if @benjub would not suggested the revision during the conversation https://github.com/metamath/set.mm/pull/2831#pullrequestreview-1109220740 (I would kept ~ brresALTV and ~ brtxpALTV ). I put a lot of work into what we both thought was rational. Should we revise our former opinion?

The *only* reason why we did not do yet what I suggest here is that nobody was motivated enough to change an accidentally determined, clearly inconsistent (cf. ~ brresOLD2 vs ~ brinxp2OLD ) notation which implied a slight but global change of 100+ theorems, and everybody used local \*ancom\* -s instead. Please do not condemn me because I do not save the world with this PR: I promise later I will:) Yes, it is nothing more that what you see now: a single commutation and the consistent implementation of its preconditions and implications throughout the whole database. Yes, it is a banal, but at the same time rational and laboriuos, PR. It's a kind of social service just like yours when you organize the set.mm database or when you substantively examine these not too exciting changes (for which I thank you in advance). And, as a matter of fact, in the middle term, until we automate it, this is the only way how you get globally rational, however extremely boring:), changes.

benjub:
I think this PR is globally for the better, and we acknowledge the effort, of course. By the way: no one rejected this PR. It's just that it's large and a lot of effort to review (and people are always more willing to put effort in their own stuff rather than reviewing someone else's --- as for me, I'd like to be working right now on a few theorems I've been meaning to add for a while...).

So, what do we do (@avekens, @wlammen, @jkingdon, @mazsa and I) ? Do we ask @mazsa to resubmit as several independent PRs ? But this may take you too much work, I'm not sure how much. Or does someone take on him to carefully review, say over the weekend ? Or do we share the work, and how ?

By the way, @mazsa : unless a proof modification is automatic or very minimal, keep an OLD version for comparison purposes.

jkingdon:
Seems to me we can just accept this. I glanced over the whole thing (perhaps not in as much detail as a BJ review, but still) and between that and the comments here I see some neutral to positive reactions and relatively minor concerns. Can't really speak for anyone else, but I do see some benefits, not a lot of risks (even if some theorems become slightly worse, it doesn't seem to be by much), and I'm also (at least in this case) willing to defer to someone who has stronger opinions about ancom than I do.

Now, if @avekens and @benjub would prefer to ask to have it broken down into smaller changes, I'd probably go along with that too. But I'm not sure I noticed things which would strongly push me in that direction.

mazsa:
@benjub, cc @avekens @jkingdon (thank you!), I wrote about refusing PR only because I think that it would be completely legitimate and sometimes you may practice it. In some cases we should *not* accept PR-s, no matter how much work is involved (this is a common practice in open source communities: admins decide on strategy). As I mentioned in my comment https://github.com/metamath/set.mm/pull/2831#discussion_r973330144 , I think this PR is a superior solution but definitely not mission critical to me, a kind of *de gustibus* thing. I'm partly an economist, I can live with sunk costs ( https://en.wikipedia.org/wiki/Sunk_cost#Fallacy_effect ). Of course, this understanding attitude of mine is not generally valid: e.g., if you would deny that my ~ uniqsALTV is the true form of ~ uniqs , I would be totally upset and would fight to first blood:), but in case of this PR, I can reprove my theorems to be compatible with ~ brresOLD2 , if needed. So please feel free to decide based only your preferences aversions available informtion and mathematical intuitions (something which I almost completely lack:). My only suggestion as a submitter is that you should decide based on the desirability of the definition and not on the basis of the additional work required for the examination of it.

benjub:
Ok. I'm going to review the PR more carefully this weekend. I think the propcalc part is ok but I'd be interested in @wlammen's opinion. As for relabelings, I think that's good too but maybe @avekens has some remarks?

set.mm
```diff
@@ -51040,64 +51108,98 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
 
-  $( Ordered pair membership in a restriction.  Exercise 13 of [TakeutiZaring]
-     p. 25.  (Contributed by NM, 14-Oct-2005.)  (Revised by BJ,
-     18-Feb-2022.) $)
-  opelresg $p |- ( B e. V -> ( <. A , B >. e. ( C |` D ) <->
-                    ( <. A , B >. e. C /\ A e. D ) ) ) $=
+  $( Ordered pair elementhood in a restriction.  Exercise 13 of [TakeutiZaring]
+     p. 25.  (Contributed by NM, 13-Nov-1995.)  (Revised by BJ, 18-Feb-2022.)
+     (Revised by Peter Mazsa, 24-Sep-2022.) $)
+  opelres $p |- ( C e. V ->
```
avekens, line 51114: Why is the suffix "g" removed? It was correct according to our conventions: 

>       When  an inference is converted to a theorem by eliminating an "is a set"
>        hypothesis, we sometimes suffix the closed form with "g" (for "more
>        general") as in ~ uniex vs. ~ uniexg .  In this case, the inference
>        often has no suffix "i".

According to this rule, also the former ~opelres had a correct label, although it is an inference.

set.mm
```diff
@@ -51040,64 +51108,98 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
+      ( <. B , C >. e. ( R |` A ) <-> ( B e. A /\ <. B , C >. e. R ) ) ) $=
```
avekens, line 51115: What was the reason why the class variables were exchanged? Maybe I missed this part in the (long) discussion...

changes-set.txt
```diff
@@ -27,14 +27,36 @@ make a github issue.)
...
  1-Oct-22 mo2       mof         "f-version" of df-mo
 26-Sep-22 euequ1    euequ
 26-Sep-22 eueq1     eueqi       inference associated with eueq
+18-Sep-22 brinxp2ALTV brinxp2   moved from PM's mathbox to main set.mm
+18-Sep-22 brinxp2   brinxp2OLD
 17-Sep-22 zfnuleu   ---         deleted; use nulmo instead
 17-Sep-22 bm2.5ii   uniordint   more informative label
 17-Sep-22 bm1.1     ---         deleted; use axextmo instead
+14-Sep-22 idssxp    [same]      proved by pre-function definitions
+14-Sep-22 idssxp    idssxpOLD
```
avekens, line 55: idssxp is not changed in this PR!? Than this should be removed (in general, only changing a proof should not documented in this file...).

set.mm
```diff
@@ -58090,8 +58189,9 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
     $d x F $.  $d x A $.  $d x B $.
     $( The value of a restricted function.  (Contributed by NM, 2-Aug-1994.) $)
     fvres $p |- ( A e. B -> ( ( F |` B ) ` A ) = ( F ` A ) ) $=
-      ( vx wcel cv cres wbr cio cfv vex brres rbaib iotabidv df-fv 3eqtr4g ) AB
-      EZADFZCBGZHZDIARCHZDIASJACJQTUADTUAQARCBDKLMNDASODACOP $.
+      ( vx wcel cv cres wbr cio cfv wa wb cvv brres baib iotabidv df-fv 3eqtr4g
+      elv ) ABEZADFZCBGZHZDIAUACHZDIAUBJACJTUCUDDUCTUDUCTUDKLDBAUACMNSOPDAUBQDA
+      CQR $.
```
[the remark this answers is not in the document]
avekens: OK now!

changes-set.txt
```diff
@@ -27,14 +27,36 @@ make a github issue.)
...
+ 4-Oct-22 opelresg  opelresgOLD2
+ 4-Oct-22 brres     brresOLD2
+ 4-Oct-22 brres     brresi
+ 4-Oct-22 opelres   opelresOLD2
+ 4-Oct-22 opelres   opelresi
+ 4-Oct-22 brresALTV brres       moved from PM's mathbox to main set.mm
+ 4-Oct-22 opelresALTV opelres   moved from PM's mathbox to main set.mm
+ 4-Oct-22 inxpssres [same]      moved from PM's mathbox to main set.mm
+ 4-Oct-22 biancomd  [same]      moved from PM's mathbox to main set.mm
+ 4-Oct-22 biancomi  biancom     moved from PM's mathbox to main set.mm
```
avekens, line 42: anbi1ci and anbi1cd are also moved from PM's mathbox to main set.mm!?

avekens (changes requested):
- The following theorems still have (significantly) longer proofs: ~cnvresima, ~resco, ~funressnvmo. This should be analysed/explained.
- changes-set.txt must be revised (see my inline comments)
- my inline comments in set.mm should be clarified/considered.

set.mm
benjub: It's true that one is the inference associated with the other (except for `V` replaced by `_V`), so this would give labels xxx/xxxi, but in this special case, as @avekens points out, it is our convention to use xxxg/xxx. Maybe we should change that convention, but for the moment it is the one we used. (There are also cases where we use xxxt/xxx... maybe one day we should be a bit more consistent, but another day).

set.mm
benjub: @avekens : the main change is that the conjunction in the consequent is commuted. The reason is that most usages use that form, so this saves commutation theorems like ~ancom more often than it introduces them. (The change of variable names is not very important, but having `R` for what is intended to be a relation seems to be for the better.)

set.mm
```diff
@@ -332247,42 +332342,43 @@ or are almost disjoint (the interiors are disjoint).  (Contributed by
       dvres.b $e |- ( ph -> B C_ S ) $.
       dvres.y $e |- ( ph -> y e. CC ) $.
       $( Lemma for ~ dvres .  (Contributed by Mario Carneiro, 8-Aug-2014.)
-         (Revised by Mario Carneiro, 28-Dec-2016.) $)
+         (Revised by Mario Carneiro, 28-Dec-2016.)  (Revised and shortened by
+         Peter Mazsa, 2-Oct-2022.) $)
```
benjub, line 332346: In order to have standard tags, write:
```
         (Revised by Mario Carneiro, 28-Dec-2016.)  Commute the consequent and
         shorten proof.  (Revised by Peter Mazsa, 2-Oct-2022.) $)
```

set.mm
```diff
@@ -51040,64 +51108,98 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
   $}
 
-  $( Ordered pair membership in a restriction.  Exercise 13 of [TakeutiZaring]
-     p. 25.  (Contributed by NM, 14-Oct-2005.)  (Revised by BJ,
-     18-Feb-2022.) $)
-  opelresg $p |- ( B e. V -> ( <. A , B >. e. ( C |` D ) <->
-                    ( <. A , B >. e. C /\ A e. D ) ) ) $=
+  $( Ordered pair elementhood in a restriction.  Exercise 13 of [TakeutiZaring]
+     p. 25.  (Contributed by NM, 13-Nov-1995.)  (Revised by BJ, 18-Feb-2022.)
+     (Revised by Peter Mazsa, 24-Sep-2022.) $)
```
benjub, line 51113: Write:
```
     p. 25.  (Contributed by NM, 13-Nov-1995.)  (Revised by BJ, 18-Feb-2022.)
     Commute the consequent.  (Revised by Peter Mazsa, 24-Sep-2022.) $)
```

set.mm
```diff
@@ -51040,64 +51108,98 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
+  opelres $p |- ( C e. V ->
+      ( <. B , C >. e. ( R |` A ) <-> ( B e. A /\ <. B , C >. e. R ) ) ) $=
+    ( cop cres wcel cvv cxp wa df-res elin2 elex biantrud opelxp anbi1cd syl5bb
+    syl6rbbr ) BCFZDAGZHTDHZTAIJZHZKCEHZBAHZUBKTDUCUADALMUEUDUFUBUEUFUFCIHZKUDU
+    EUGUFCENOBCAIPSQR $.
+
+  $( Binary relation on a restriction.  (Contributed by Mario Carneiro,
+     4-Nov-2015.)  (Revised by Peter Mazsa, 24-Sep-2022.) $)
```
benjub, line 51121: Write:
```
  $( Binary relation on a restriction.  (Contributed by Mario Carneiro,
     4-Nov-2015.)  Commute the consequent.  (Revised by Peter Mazsa,
     24-Sep-2022.) $)
```

benjub:
Thanks @avekens for the review. Good for me too. @mazsa : can you make the changes requested by @avekens and me (and no other changes) ? Please make the relabeling requested by @avekens in its own commit, and the modifications to `changes-set.txt` also in their own commit, and finally the comment changes I proposed in their own commit (but all in this PR of course).

set.mm
mazsa: Simple method, thank you the idea @avekens . I checked cca. 20% of possible cases and found useful in 5 of them: ~ ssrnres ~ fssres ~ smores ~ brdom4 ~ smobeth . So it seems that we need this inference. Based on that I mark this resolved.

mazsa:
> The following theorems still have (significantly) longer proofs: ~cnvresima, ~resco, ~funressnvmo. This should be analysed/explained.

@avekens In case of ~ cnvresima and ~ resco both have the same number of steps than their old versions. Their compressed versions are longer for the same reason. If you open the proof of any of them, you'll see that the old version uses ~ anbi1i two times which means that in compressed proof there is only one ~ anbi1i . Its new version uses first the same ~ anbi1i but at second time it uses ~ anbi2ci , which means that in the compressed form there are both ~ anbi1i and ~ anbi2ci . This is the reason why their new compressed proofs are 1 label longer while they have the same number of steps.
~ funressnvmo : I'll detail it after based on next commit of this PR tomorrow (edit: today. Sunday).

mazsa:
@avekens @benjub thank you for you efforts. And thank you for your suggestions.
Alexander, I dare to write you because of your wording "my inline comments in set.mm should be clarified/considered". I considered one of your suggestions, I found it that *you are right*, and I decided that *I will do it without hesitation* ...if you do not like my following suggestion:
We should change the conventions that way that in case of new/changing theorems we should regard the general form to be the default form, without the suffix \*g .
Please feel free to say no (because of the traditions, because of the existing conventions, anything), or to say not now, and I'll begin my work at once.

set.mm
benjub: Let's say it's ok here.
If you want to do this properly in the future, you have to check that it shortens proofs by more bytes than the number of bytes (characters) necessary to state and prove the result, as I recalled above, and as was also recalled by @jkingdon (e.g., do as in https://github.com/metamath/set.mm/pull/2796#discussion_r955343777 where you see the output of `scripts/min.cmd`)

avekens:
> ... following suggestion: We should change the conventions that way that in case of new/changing theorems we should regard the general form to be the default form, without the suffix *g . 

I think this is a good suggestion, so not further changes are needed regarding this issue.

benjub:
> I think this is a good suggestion, so not further changes are needed regarding this issue.

There are still the three explicit comment suggestions I made above, and the merge conflicts to resolve. PRs are queueing and `changes-set.txt` causes merge conflicts, so let's do these last changes, and no other changes.  We'll adapt the conventions later.

set.mm
mazsa: (Resolved below https://github.com/metamath/set.mm/pull/2873#issuecomment-1272455776 )

mazsa:
@avekens This new version of ~ funressnvmo has the same number of steps and the
same number of labels (deleted ~ id ~ sylanblrc , added  ~ mpbiran ~
biimpri).
Since these are not trivial changes or shortenings I created an OLD
version as well.
@benjub I´m on it

set.mm
mazsa: Done ( https://github.com/metamath/set.mm/pull/2873/commits/9d55c34a5c968159491c2b0a38aaa05f3c3bd727 )

set.mm
mazsa: done ( https://github.com/metamath/set.mm/pull/2873/commits/9d55c34a5c968159491c2b0a38aaa05f3c3bd727 )

set.mm
mazsa: done ( https://github.com/metamath/set.mm/pull/2873/commits/c72c6adbeef3a854ff3937ff8578338780a8ed2b )

changes-set.txt
```diff
@@ -27,14 +27,36 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 4-Oct-22 brresi    brresi2
+ 4-Oct-22 opelresi  opelidres
+ 4-Oct-22 brresg    brresgOLD2
```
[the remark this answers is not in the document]
mazsa: done ( https://github.com/metamath/set.mm/pull/2873/commits/1961e2a3ae7faae6628f84557a3e4cf029cdc556 )

changes-set.txt
mazsa: done ( https://github.com/metamath/set.mm/pull/2873/commits/1961e2a3ae7faae6628f84557a3e4cf029cdc556 )

changes-set.txt
mazsa: done ( https://github.com/metamath/set.mm/pull/2873/commits/1961e2a3ae7faae6628f84557a3e4cf029cdc556 )

mazsa:
@avekens @benjub @jkingdon Alexander, Benoit, Jim thank you for all again! (I think it´s ready to merge).

avekens (approved):
Since my issues are solved, I approve this PR. However, @benjub and @jkingdon should adpprove this PR (after remainig revisions are done), too.

jkingdon (approved):
I haven't reviewed this in nearly as much detail as @benjub or @avekens so I'm not sure how much my approval counts for, but I glanced over it and it looked good.

jkingdon:
I resolved the conflicts in changes-set.txt using the github web interface. Hope I'm not stepping on any toes by doing this but it looked straightforward.

benjub:
I cannot see
https://github.com/metamath/set.mm/pull/2873/commits/a62102f26484fe6ee325dc007eb529a1baae5eda
https://github.com/metamath/set.mm/pull/2873/commits/9d55c34a5c968159491c2b0a38aaa05f3c3bd727
https://github.com/metamath/set.mm/pull/2873/commits/c72c6adbeef3a854ff3937ff8578338780a8ed2b
"This page is taking too long to load."

PR #2874, opened 2022-10-04
Minor

benjub:
Subdivided into commits; commit messages say it all.
Hopefully this will not conflict with @mazsa's PR (at first sight, it won't, except maybe for changes-set.txt)


wlammen (approved):
Apart from the renamings and reorderings the addition brrelex12i is justified by the shortenings possible.

PR #2875, opened 2022-10-05
Summation theorems from isumclim to isumadd

jkingdon:
These intuitionize without a lot of trouble (some need no changes, others need the usual things about seq and summation theorems, apart versus not equal, etc).

Includes additional changes from four-argument `seq` to three-argument `seq` including adding a conversion theorem `iseqseq3` and using it to save some steps most places which had been using `iseqsst` directly. This is mostly driven by the usages of `seq` in the section being intuitionized.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7719,7 +7719,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <TR>
   <TD>fsumser</TD>
-  <TD>~ fisumser</TD>
+  <TD>~ fsum3ser</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7737,6 +7737,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>isumdivc</TD>
+  <TD>~ isumdivapc</TD>
+  <TD>Changes not equal to apart</TD>
+</TR>
+
+<TR>
   <TD>seqabs</TD>
   <TD><I>none</I></TD>
   <TD>Although something along these lines could be proved,
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7683,11 +7683,6 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
-  <TD>isum</TD>
-  <TD>~ iisum</TD>
-</TR>
-
-<TR>
   <TD>fsum</TD>
   <TD>~ fisum</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7724,9 +7724,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <TR>
   <TD>fsumser</TD>
-  <TD><I>none</I></TD>
-  <TD>Presumably provable once modified in the usual ways which
-  arise when using ` seq ` for finite sums.</TD>
+  <TD>~ fisumser</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6387,7 +6387,7 @@ or ~ ssfiexmid </TD>
 
 <TR>
   <TD>seradd</TD>
-  <TD>~ iseradd</TD>
+  <TD>~ ser3add</TD>
   <TD>iseradd requires that ` F ` , ` G ` , and ` H ` be defined on
   ` ( ZZ>= `` M ) ` not merely ` ( M ... N ) ` as in seradd . This
   is not a problem when used on infinite sequences, but perhaps this


jkingdon:
~~Github is having an outage with webhooks according to https://www.githubstatus.com/ . I don't know for sure whether that is why the automated checks aren't running on this pull request, but that would be my guess. We can wait for that to resolve and it might fix itself or we might be able to try the "close and reopen" trick again once github is better.~~ Checks have now run.

iset.mm
```diff
@@ -111475,6 +111475,23 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
     $}
   $}
 
+  ${
+    $d F k v $.  $d M k v $.  $d Z k $.  $d k ph v $.
+    zsum.1 $e |- Z = ( ZZ>= ` M ) $.
+    zsum.2 $e |- ( ph -> M e. ZZ ) $.
+    isum.3 $e |- ( ( ph /\ k e. Z ) -> ( F ` k ) = B ) $.
+    isum.4 $e |- ( ( ph /\ k e. Z ) -> B e. CC ) $.
+    $( Series sum with an upper integer index set (i.e. an infinite series).
```
benjub, line 111484: Shouldn't it be something like "The sum of a series is equal to its limit. " ?

iset.mm
```diff
@@ -111891,6 +111891,26 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
       CXJMPUAOZMPSXJYCLUGMPAXJYCWDVOWEVH $.
   $}
 
+  ${
+    $d k v F $.  $d k v M $.  $d k v N $.  $d k v ph $.
+    fsum3ser.1 $e |- ( ( ph /\ k e. ( ZZ>= ` M ) ) -> ( F ` k ) = A ) $.
+    fsum3ser.2 $e |- ( ph -> N e. ( ZZ>= ` M ) ) $.
+    fsum3ser.3 $e |- ( ( ph /\ k e. ( ZZ>= ` M ) ) -> A e. CC ) $.
+    $( A finite sum expressed in terms of a partial sum of an infinite series.
+       The recursive definition of follows as ~ fsum1 and ~ fsump1 , which
```
benjub, line 111900: of ?

iset.mm
```diff
@@ -112364,6 +112364,34 @@ seq m ( + , ( n e. ZZ |-> if ( n e. A , [_ n / k ]_ B , 0 ) ) , CC )
...
+    $d j m x A $.  $d j k m x M $.  $d j k m x ph $.  $d j k m x Z $.
+    $d j x F $.
+    isumclim3.1 $e |- Z = ( ZZ>= ` M ) $.
+    isumclim3.2 $e |- ( ph -> M e. ZZ ) $.
+    isumclim3.3 $e |- ( ph -> F e. dom ~~> ) $.
+    isumclim3.4 $e |- ( ( ph /\ k e. Z ) -> A e. CC ) $.
+    isumclim3.5 $e |- ( ( ph /\ j e. Z ) ->
+                        ( F ` j ) = sum_ k e. ( M ... j ) A ) $.
+    $( The sequence of partial finite sums of a converging infinite series
+       converge to the infinite sum of the series.  Note that ` j ` must not
```
benjub, line 112377: converges

benjub (approved):
A few minor comments.

iset.mm
avekens: "of ...?" also in the comment of the next theorem. It was/is already unclear in the comment of the original theorem ~fsumser in set.mm.

avekens (approved):
Also only minor remarks.

iset.mm
```diff
@@ -99786,6 +99786,18 @@ seq K ( .+ , G , S ) ) $=
       TJOKMASTPQR $.
   $}
 
+  ${
+    $d F x y $.  $d M x y $.  $d ph x y $.
+    iseqseq3.m $e |- ( ph -> M e. ZZ ) $.
+    iseqseq3.f $e |- ( ( ph /\ x e. ( ZZ>= ` M ) ) -> ( F ` x ) e. CC ) $.
+    $( Equality using four argument or three argument sequence builders.
```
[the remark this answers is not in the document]
benjub: @avekens : this is `seq M ( +. , F , S )` versus `seq M ( +. , F )`
@jkingdon : maybe write: "four-argument or three-argument sequence builders ( ~ df-... or ~ df-... )."

iset.mm
jkingdon: It's the limit of the partial sums, but usually I don't try to wordsmith stuff like this when I'm just copying it from set.mm to iset.mm.

iset.mm
jkingdon: Here I'll just delete the "of" in all three places.

iset.mm
jkingdon: Fixed in set.mm and iset.mm.

iset.mm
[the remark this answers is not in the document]
jkingdon: I'm usually a fan of having the comment restate or explain the symbols, rather than just repeat what is in the formal statement of the theorem. But this is a case where I'm leaning towards just saying

```
Equality of ` seq M ( + , F , CC ) ` and ` seq M ( + , F ) `
```

rather than getting into counting arguments or pointing to definitions.

The good news is that this whole distinction should go away once we switch everything (or almost everything) to ` seq M ( + , F ) ` .

PR #2880, opened 2022-10-08
iset.mm summation from sumsplit to fsum2d

jkingdon:
The parts of this which directly affect summation are fairly small:
* sumsplit is intuitionized with two additional decidability hypotheses
* fsump1i , copied without change
* fsum2d . Stated as in set.mm.  The proof is also similar to the set.mm one but needs a little intuitionizing and a fairly large development of finiteness for indexed unions and disjoint collections.

The various helper theorems are:
* dcun . I was a little surprised we didn't have more theorems like this already, but apparently we haven't needed them.
* Intuitionizing more of the `Disj_` section: `disjnim` (replaces `disjor`), `disjnims` (replaces `disjors`), `disji2`, `disjiun`. Plus `disjxp1` and `disjsnxp` (taken from a set.mm mathbox but needed for other proofs here).
* Various copies from set.mm: `iunxsngf`, `elv`, `elvd`, `clelsb3f` , `rmo3f` , `rmo4f`, `mptru`
* Finiteness of indexed unions: `iunfidisj` (conceptually similar to existing things like https://us.metamath.org/ileuni/unfidisj.html but the details depend on a lot of the disjoint collection theorems in this pull request)
* A few additional tweaks to the missing theorems list in mmil.html

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1781,7 +1781,7 @@ favor of theorems in deduction form.</TD>
 </TR>
 
 <TR>
-  <TD>disji2 , disji</TD>
+  <TD>disji</TD>
   <TD><I>none</I></TD>
 </TR>
 
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1766,6 +1766,13 @@ favor of theorems in deduction form.</TD>
 </TR>
 
 <TR>
+  <TD>disjss3</TD>
+  <TD><I>none</I></TD>
+  <TD>Might need to be restated or have decidability conditions
+  added</TD>
+</TR>
+
+<TR>
 <TD>trintss</TD>
 <TD>~ trintssm </TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1786,12 +1786,6 @@ favor of theorems in deduction form.</TD>
 </TR>
 
 <TR>
-  <TD>disjiun</TD>
-  <TD><I>none</I></TD>
-  <TD>The set.mm proof relies on the converse of ~ rmo2i</TD>
-</TR>
-
-<TR>
   <TD>disjprg , disjxiun , disjxun</TD>
   <TD><I>none</I></TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3526,6 +3526,23 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD ROWSPAN="2">iunfi</TD>
+  <TD>~ iunfidisj</TD>
+  <TD>for a disjoint collection</TD>
+</TR>
+
+<TR>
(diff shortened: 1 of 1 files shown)


avekens (approved):

benjub:
I didn't check if/how you did it, but if theorems in set.mm mathboxes are used in iset.mm/main, then I think they should also be moved in set.mm from mathboxes to Main.

avekens:
> I didn't check if/how you did it, but if theorems in set.mm mathboxes are used in iset.mm/main, then I think they should also be moved in set.mm from mathboxes to Main.

Which theorem(s) is/are from a mathbox? How did you find it/them? 
We could decise in this PR if they shuld be moved to main or not, but in general, it seems to be reasonable to have them in main  of set.mm if they are in main of iset.mm.

jkingdon:
> Which theorem(s) is/are from a mathbox? How did you find it/them? We could decise in this PR if they shuld be moved to main or not, but in general, it seems to be reasonable to have them in main of set.mm if they are in main of iset.mm.

In this pull request I think it was https://us.metamath.org/mpeuni/iunxsngf.html , https://us.metamath.org/mpeuni/rmo3f.html , and https://us.metamath.org/mpeuni/rmo4f.html . There have been others in the past (no, I didn't keep a list) and will be more in the future.

Usually I find them with the metamath-exe "search" command. Or maybe something like `show label rmo*` (because in this case I tried to use https://us.metamath.org/ileuni/rmo4.html , got a distinct variable violation, and knew I needed a not-free version. Pretty much all of the ways I search would hit mathboxes and main equally.

I don't object to moving them in set.mm if they end up in iset.mm main but I'm a little less sure I want to add another thing to my todo list every time this situation happens. I mean, I can try to make a principled statement about how whether they have any usages in set.mm and whether they have any usages in iset.mm are two separate things and maybe I believe that. But there's also a question of who would do the work (including where to put them in set.mm, whether the name is good, etc).

benjub:
Thanks @jkingdon for #2889.  It would be nice to keep that policy of moving set.mm theorems from mathboxes to Main as soon as they are in iset.mm/Main.  It indeed is some work, so for the days when you're too lazy, simply keeping a list would already be great (no need to go to the hassle of having a shared updated list: you could simply keep a list for yourself, and when you have a bunch, open an issue with the list of labels in plain text, and some good samaritan will take it).

jkingdon:
> keeping a list  . . . open an issue

Not a bad suggestion. I'll keep this in mind as an option.

And while we are proposing tasks for our tireless good samaritan I also wrote up https://github.com/metamath/set.mm/issues/2891 . Don't know if you have been thinking in similar terms as you've been pondering keeping set.mm and iset.mm in sync, but that proposed script seems like a potentially attainable step (at least for someone who does more with software than I have done on this project).

PR #2881, opened 2022-10-08
reduce axioms in number elementhood proofs

icecream17:
rename 2cnOLD -> 2cnALT, as noted in https://github.com/metamath/set.mm/pull/2872
and also reorder rpssre

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 8-Oct-22 2cnOLD    2cnALT
  3-Oct-22 brrelexi  brrelex1i
  3-Oct-22 brrelex   brrelex1
  1-Oct-22 eu5       dfeu        this is now a rederivation
--- a/discouraged
+++ b/discouraged
@@ -13287,7 +13287,7 @@ New usage of "1t1e1ALT" is discouraged (0 uses).
 New usage of "2atm2atN" is discouraged (0 uses).
 New usage of "2atnelvolN" is discouraged (0 uses).
 New usage of "2bornot2b" is discouraged (0 uses).
-New usage of "2cnOLD" is discouraged (0 uses).
+New usage of "2cnALT" is discouraged (0 uses).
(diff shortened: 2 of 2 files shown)


benjub:
To me, mergeable after that typo is fixed.

wlammen (changes requested):
I think you should add rpreOLD and rpssreOLD.  Otherwise well done. Reduction in axiom usage confirmed for all but rpre and rpssre.  The latter are an obvious shortening.

benjub:
> I think you should add rpreOLD and rpssreOLD. Otherwise well done. Reduction in axiom usage confirmed for all but rpre and rpssre. The latter are an obvious shortening.

To me, the swap rpre <-> rpssre is straightforward enough that there is no need to keep OLD versions, but if @wlammen thinks there is, then let it be so.

changes-set.txt
```diff
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 8-Oct-22 2cnOLD    2cnALT
```
benjub, line 30: Not needed since we have 2cn with identical statement anyway.

wlammen:
> if @wlammen thinks there is, then let it be so.

Why only do I feel being criticized here?

I don't know why my personal opinion plays any role here.  I almost never look at OLD versions, and I am often annoyed because they litter pages like 'Nearby theorems'.  That's why I am often keen to remove outdated ones early.

But it is an established habit to create these, for whatever reason, and there might be people around regarding them as important.  
So I let general ideas prevail.

avekens:
I also think rpreOLD and rpssreOLD should be added. Afterwards, i will accept this PR.

benjub:
> Why only do I feel being criticized here?

That was really not my intent. Quite the opposite: it's because I know you have experience with this that I say "if @wlammen  thinks this, then let it be so". If that came from someone totally new to set.mm, I would be more likely to go over their opinion.

icecream17:
this took a while, sorry


I made a new pr with all the commented updates:
https://github.com/metamath/set.mm/pull/2888

PR #2883, opened 2022-10-09
Shorten 19.23t

benjub:


benjub:
@wlammen : do you think that https://us.metamath.org/mpeuni/19.42-1.html still has some interest ? It has the same dependencies as https://us.metamath.org/mpeuni/19.42.html and its only usage https://us.metamath.org/mpeuni/bnj596.html can use 19.42/sylib instead of 19.42-1/syl.

wlammen:
No

jkingdon:
> do you think that https://us.metamath.org/mpeuni/19.42-1.html still has some interest ?

Even in iset.mm we get both directions of the biconditional: https://us.metamath.org/ileuni/19.42.html

So I'd agree with getting rid of `19.42-1`.

set.mm
```diff
@@ -18693,8 +18698,18 @@ Converse of the inference rule of (universal) generalization ~ ax-gen .
 
   $( Closed form of Theorem 1977.23 of [Margaris] p. 90.  See ~ 19.23 .
```
avekens, line 18699: Theorem 1977.23? I think it must be 19.23...

set.mm
```diff
@@ -18693,8 +18698,18 @@ Converse of the inference rule of (universal) generalization ~ ax-gen .
...
+     (Proof shortened by BJ, 8-Oct-2022.) $)
   19.23t $p |- ( F/ x ps -> ( A. x ( ph -> ps ) <-> ( E. x ph -> ps ) ) ) $=
+    ( wnf wex wal wi 19.38b 19.3t imbi2d bitr3d ) BCDZACEZBCFZGABGCFMBGABCHLNBM
+    BCIJK $.
+
+  $( Obsolete proof of 19.23t as of 8-Oct-2022.  (Contributed by NM,
+     7-Nov-2005.)  (Proof shortened by Wolf Lammen, 13-Aug-2020.) ~ df-nf
+     changed.  (Revised by Wolf Lammen, 11-Sep-2021.)
+     (Proof modification is discouraged.)  (New usage is discouraged.) $)
+  19.23tOLDOLD $p |-
```
avekens, line 18711: The current ~19.23tOLD can be deleted (due 6-Oct-2021), so this OLDOLD version can be called ~19.23tOLD.

avekens (changes requested):

set.mm
benjub: I swapped labels 19.23tOLD <-> 19.23tOLDOLD. That other is associated with all the `nfOLD` theorems which have to be removed, but let's do that separately.

set.mm
benjub: fixed

avekens (approved):

PR #2884, opened 2022-10-10
Conventions-labels: default form for new/revised/upgraded theorems

mazsa:
A short follow up PR based on https://github.com/metamath/set.mm/pull/2873#issuecomment-1272455776 of @avekens . It seems that @benjub wouldn't object either https://github.com/metamath/set.mm/pull/2873#discussion_r990653917 .


set.mm
```diff
@@ -419279,6 +419279,16 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
        </li>
        </ul>
 
+       <li><b>Default form.</b>
```
avekens, line 419282: I would propose not to create a new list item for this, but to append this paragraph at the end of list item "Theorem forms.", which ends with

> When an inference is converted to a theorem by eliminating an "is a set"
>  hypothesis, we sometimes suffix the closed form with "g" (for "more
>  general") as in ~ uniex vs. ~ uniexg .  In this case, the inference
>  often has no suffix "i".

avekens (changes requested):

set.mm
mazsa: @avekens this way? https://github.com/metamath/set.mm/pull/2884/commits/2173e350a8fd5fbc96b2cc1d28efab6fe097e7e8

avekens (approved):

benjub:
I thought we were removing a special case, so I was expecting a welcomed shortening of the conventions, but I see it's an addition... At least it describes the situation faithfully, so let's merge it, and maybe we'll try to shorten the whole thing in the future.

PR #2885, opened 2022-10-10
Minor

benjub:
Edit to conform to logical quotation style, as adopted recently.
On the way, I reworded a few comments.

(edit: sorry for the multiple commits... do not review by commit, but review the global change)


set.mm
```diff
@@ -281840,7 +281846,7 @@ F C_ ( CC X. X ) ) $=
     $d o x y z A $.  $d o x z B $.  $d a j m n o x y z J $.  $d j o x y z X $.
     $d m n x y P $.  $d m n y Q $.
     ist0.1 $e |- X = U. J $.
-    $( The predicate "is a T_0 space."  Every pair of distinct points is
+    $( The predicate "is a T_0 topology".  Every pair of distinct points is
```
avekens, line 281849: ? T_0 is a (Kolmogorov) space!

benjub: Yes, set.mm is not very consistent here.  https://us.metamath.org/mpeuni/df-haus.html defines a class of topologies. That's probably because a lot of topology was added to set.mm before the introduction of extensible structures. I wrote the section comment of "12.1.1  Topological spaces" to try and clarify things. Eventually, I guess everything should be done within the framework of extensible structures.

For the present case: even if this is a bit abusive, do you prefer I retain "space", to be consistent within that subsection ?

benjub: done

avekens (approved):

PR #2886, opened 2022-10-10
om0

benjub:
Carry out one of Norm's todo's.
Some other minor changes.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16999,7 +16999,6 @@ New usage of "ocval" is discouraged (4 uses).
 New usage of "ogrpinvOLD" is discouraged (0 uses).
 New usage of "oldmm3N" is discouraged (2 uses).
 New usage of "olposN" is discouraged (0 uses).
-New usage of "om0x" is discouraged (0 uses).
 New usage of "omlfh1N" is discouraged (2 uses).
 New usage of "omlfh3N" is discouraged (0 uses).
 New usage of "omllaw2N" is discouraged (3 uses).
--- a/discouraged
+++ b/discouraged
@@ -16999,6 +16999,7 @@ New usage of "ocval" is discouraged (4 uses).
 New usage of "ogrpinvOLD" is discouraged (0 uses).
 New usage of "oldmm3N" is discouraged (2 uses).
 New usage of "olposN" is discouraged (0 uses).
+New usage of "om0x" is discouraged (0 uses).
 New usage of "omlfh1N" is discouraged (2 uses).
 New usage of "omlfh3N" is discouraged (0 uses).
 New usage of "omllaw2N" is discouraged (3 uses).


set.mm
```diff
@@ -75934,14 +75944,12 @@ currently used conventions for such cases (see ~ cbvmpt2x , ~ ovmpt2x and
       UIHQBEARTUFUGUHEUAUBUC $.
   $}
 
-  $( Ordinal multiplication with zero.  Definition 8.15 of [TakeutiZaring]
-     p. 62.  Unlike ~ om0 , this version works whether or not ` A ` is an
```
avekens, line 75938: "Definition 8.15 of [TakeutiZaring] p. 62." should be kept, shouldn't it?

avekens: It seems that the proofs which used ~om0x became longer. after it was replaced by om0. This seems to be a little bit contradictory to our efforts to shorten proofs. With the same argument, we could forbid to use inference forms of theorems, because all of them can be replaced by theorems in deduction form, which is our preferred style.

Furthermore, we have a lot of additional theorems which are not conventional in the sense that they do not respect the domain of a function. How should we handle these cases?

avekens: > "Definition 8.15 of [TakeutiZaring] p. 62." should be kept, shouldn't it?

I just detected that "Definition 8.15 of [TakeutiZaring] p. 62." is also contained in the comment for ~om0. Which of these theorems is closer to the definition of [TakeutiZaring]?

benjub: Definition 8.15 page 62 of TZ is the definition of ordinal addition. The base case reads: $\alpha \cdot 0 \coloneqq 0$ without any further precision. One has to go back to the first occurrence of $\alpha$, which is Definition 7.16 page 39 and the remarks preceding and following it, which say that any $\alpha, \beta, \dots$ appearing in a formula are ordinals and that one shoud mentally translate, e.g., $\varphi(\alpha)$ to $(Ord(x) \to \varphi(x))$. Therefore, the closer transcription of Definition 8.15(a) is our ~om0.

benjub: Yes, since ~om0 is less general than ~om0x, proofs were a bit lengthened. I carried out that TODO by Norm and I agree with it.  It's a bit like ``sinneg $p |- ( A e. CC -> ( sin ` -u A ) = -u ( sin ` A ) ) $=``, which would also hold without antecedent (then, everything is out of domain so evaluates to the empty set). And actually, ~om0x is a bit worse, since the empty set is also an ordinal number, so the result is not "obviously garbish" if A is not an ordinal, as it would be for ~sinneg if A was not a complex number.

There are many places where the out-of-domain behavior of our encoding is used, but since ~om0 is pretty fundamental, I agree with the TODO that it seems better to use only ~om0 and avoid ~om0x (but keep it to illustrate the phenomenon).  I can maybe also add a link to ~om0x from the comment of ~om0 ?

benjub: And as for your latter question, "how should we handle these cases": I used to want to be super-strict and forbid all uses of such theorems (e.g., I would discourage use of ``df-fv |- ( F ` A ) = ( iota x A F x )`` and only use it to prove, with ~a1i, ``dffv |- ( ( Fun F /\ A e. dom F ) -> ( F ` A ) = ( iota x A F x ) )`` and then only use the latter) but as @digama0 remarked, that would be too much trouble when actually proving theorems, compared to the little "purity" gain, and he wrote that a "gentlemen's agreement" should be reached.  For instance, we have a few "`(Avoid depending on this detail.)`" tags for couples (ordered pairs), which are less strict than usage discouragement tags, but hint at the matter.

avekens: OK, so we should keep this in mind for future cases of this kind.

avekens (approved):

Issue #1648, opened 2020-05-10
New definition for "prefixes of words"


avekens:
Motivated by the introduction of the new symbol ++ for the concatenation of words, and the discussion about the indices for words, starting either at 0 or 1 (see google group https://groups.google.com/d/topic/metamath/UwTUuNPgaB0/discussion), I looked at the theorems about words and could improve/shorten the proofs of a lot of them. Meanwhile, I observed that many theorems about subwords are actually about prefixes, i.e. subwords starting at the beginning of a word:  ` ( W substr <. 0 , L >. ) `.

I think this gives reason to provide a special definition/symbol for prefixes of words. You can find a proposal in my mathbox:

df-pref $a |- prefix = ( s e. _V , l e. NN0 |-> ( s substr <. 0 , l >. ) ) $.

By this definition, a prefix (of length ` L `) can be expressed as ` ( W prefix L ) ` instead of ` ( W substr <. 0 , L >. ) `, saving 10 characters in length (including blanks). Besides being shorter, such a definition has the advantage that it hides the "implementation detail" that the indexes of a word start with 0. Furthermore, having separate theorems for prefixes can shorten many proofs.

Because of these advantages, I like to ask for moving this definition and the related theorems into the main body of set.mm, section "5.7 Words over a set". After that, many existing theorems about prefixes could be revised accordingly (I identified already 30 of them, see theorems in my mathbox having "Could replace ~ ..." in their comments).

Regarding the names, labels and symbols, however, these could be discussed (see also the header comment of section "Prefixes of a word" in my mathbox): Would "pfx" be a better label (fragment) than "pref"? In general, is "prefix" an adequate name/symbol for subwords starting at the beginning of a word? Maybe "leading characters" or "leading symbols" or even "leading sequence of characters/symbols" would be more precise, but clumsier. A handy alternate symbol, however, could be "lead", the corresponding label (fragment) "ld" or even "lead" itself".

Searching for "prefix abbreviation" in the web, I found https://www.allacronyms.com/prefix/abbreviated, proposing "pfx" as abbreviation. Regarding the meaning of "prefix", I think it is different in automata theory/formal languages compared with linguistics, see Wikipedia "Substring" (https://en.wikipedia.org/wiki/Substring#Prefix), or https://math.stackexchange.com/questions/2190559/is-there-standard-terminology-notation-for-the-prefix-of-a-word).

[...]

PR #2890, opened 2022-10-14
New definition for "prefixes of words"

avekens:
As discussed in #1648, the new definition for "prefixes of words" df-pfx is moved from my mathbox to main set.mm.

Until now:
* df-pfx and related theorems are moved from AV's mathbox to main
* suffix OLD is appended to theorems which are replaced by theorems using the prefix operator.
* theorems are rearranged, mainly to have OLD theorems near to the current ones
* revision of comments:
** "Could replace.." replaced by original contribution tag.
** OLD theorems marked as obsolete
* OLD theorems are replaced in the proofs of the following theorems: ~psgnunilem5 ~wrdind ~wrd2ind ~splval2 ~cshwidxmod ~wrdeqs1cat ~swrdccat3b ~swrdccat ~ccatopth ~splid ~cshw0 ~iwrdsplit (as a side effect, this saved 324 bytes in total)

TODO:
* revision of df-splice 
* replace remaining OLD theorems used in proofs (mainly of theorems of Graph Theory) - I assume that this will shorten again the proofs, hopefully significantly.
* rewrap, especially the proofs of the OLD theorems (I didn't it yet intentionally, to make reviews easier).

This is an intermediate PR not intended to be merged immediately. The TODOs listed above should be done first, but I want to publish the material already now to get an early feedback (is somebody is interested).

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,13 +27,15 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
-13-Oct-22 rmo4f     [same]      moved from TA's mathbox to main
-13-Oct-22 rmo3f     [same]      moved from TA's mathbox to main
-13-Oct-22 iunxsngf  [same]      moved from TA's mathbox to main
- 9-Oct-22 ndisj     [same]      moved from RP's mathbox to main
- 9-Oct-22 19.9alt   19.3t       moved from NM's mathbox to main
+13-Oct-22 rmo4f     [same]      moved from TA's mathbox to main set.mm
+13-Oct-22 rmo3f     [same]      moved from TA's mathbox to main set.mm
+13-Oct-22 iunxsngf  [same]      moved from TA's mathbox to main set.mm
+12-Oct-22 df-pfx    [same]      definition and related theorems
+                                moved from AV's mathbox to main set.mm
+ 9-Oct-22 ndisj     [same]      moved from RP's mathbox to main set.mm
+ 9-Oct-22 19.9alt   19.3t       moved from NM's mathbox to main set.mm
  7-Oct-22 exmoeu2   exmoeub     biconditional form of exmoeu
- 7-Oct-22 eubi      [same]      moved from ATS's mathbox to main
+ 7-Oct-22 eubi      [same]      moved from ATS's mathbox to main set.mm
  6-Oct-22 hhsshl    csschl      hhsshl was in deprecated Part 19
  6-Oct-22 hhssbn    cssbn       hhssbn was in deprecated Part 19
  6-Oct-22 ssphl     cphssphl    ssphl was in deprecated Part 19
@@ -68,18 +70,18 @@ Date      Old       New         Notes
 13-Sep-22 epelc     epeli       inference associated with epelg
 13-Sep-22 a1tru     trud        deduction form of tru
 13-Sep-22 trud      mptru       modus ponens when minor is tru
- 8-Sep-22 el2v2     elvd        moved from PM's mathbox to main
- 8-Sep-22 exlimimdd [same]      moved from ML's mathbox to main
- 4-Sep-22 funresfunco [same]    moved from AV's mathbox to main
- 3-Sep-22 pwexd     [same]      moved from GS's mathbox to main
- 3-Sep-22 moimd     [same]      moved from TA's mathbox to main
- 3-Sep-22 csbvargi  [same]      moved from GM's mathbox to main
+ 8-Sep-22 el2v2     elvd        moved from PM's mathbox to main set.mm
+ 8-Sep-22 exlimimdd [same]      moved from ML's mathbox to main set.mm
+ 4-Sep-22 funresfunco [same]    moved from AV's mathbox to main set.mm
+ 3-Sep-22 pwexd     [same]      moved from GS's mathbox to main set.mm
+ 3-Sep-22 moimd     [same]      moved from TA's mathbox to main set.mm
+ 3-Sep-22 csbvargi  [same]      moved from GM's mathbox to main set.mm
  1-Sep-22 keepel    ifcli       inference associated with ifcl
-26-Aug-22 elv       [same]      moved from PM's mathbox to main
-18-Aug-22 eel1111   syl1111anc  moved from AS's mathbox to main
-26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main
-26-Jul-22 wl-jarli  jarli       moved from WL's mathbox to main
-25-Jul-22 bj-sb3b   sb3b        moved from BJ's mathbox to main
+26-Aug-22 elv       [same]      moved from PM's mathbox to main set.mm
+18-Aug-22 eel1111   syl1111anc  moved from AS's mathbox to main set.mm
+26-Jul-22 wl-jarri  jarri       moved from WL's mathbox to main set.mm
+26-Jul-22 wl-jarli  jarli       moved from WL's mathbox to main set.mm
+25-Jul-22 bj-sb3b   sb3b        moved from BJ's mathbox to main set.mm
 20-Jul-22 brfi1indlem hashdifsnp1
 17-Jul-22 mapsnend  [same]      moved from GS's mathbox to main set.mm
 17-Jul-22 mapsnd    [same]      moved from GS's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -239,6 +239,11 @@
 "2polvalN" is used by "sspmaplubN".
 "2sb5nd" is used by "2uasbanh".
 "2sb5nd" is used by "2uasbanhVD".
+"2swrd1eqwrdeqOLD" is used by "clwwlkf1".
+"2swrd1eqwrdeqOLD" is used by "wwlksnextinj".
+"2swrd2eqwrdeqOLD" is used by "numclwwlk1lem2f1".
+"2swrdeqwrdeqOLD" is used by "2swrd1eqwrdeqOLD".
+"2swrdeqwrdeqOLD" is used by "2swrd2eqwrdeqOLD".
 "2uasbanh" is used by "2uasban".
 "3atnelvolN" is used by "2atnelvolN".
 "3atnelvolN" is used by "islvol2aN".
@@ -653,6 +658,8 @@
 "adderpq" is used by "distrnq".
 "adderpq" is used by "ltexnq".
 "adderpqlem" is used by "adderpq".
+"addlenrevswrdOLD" is used by "addlenswrdOLD".
+"addlenrevswrdOLD" is used by "lenrevcctswrdOLD".
 "addnqf" is used by "addassnq".
 "addnqf" is used by "addcomnq".
 "addnqf" is used by "adderpq".
@@ -2987,6 +2994,9 @@
 "cbncms" is used by "ubthlem2".
 "cbvexsv" is used by "onfrALTlem1".
 "cbvexsv" is used by "onfrALTlem1VD".
+"ccats1swrdeqOLD" is used by "ccats1swrdeqbiOLD".
+"ccats1swrdeqOLD" is used by "ccats1swrdeqrexOLD".
+"ccats1swrdeqrexOLD" is used by "reuccats1lemOLD".
 "ccatws1lenOLD" is used by "ccatw2s1lenOLD".
 "ccatws1lenOLD" is used by "ccatws1lenrevOLD".
 "ccatws1lenOLD" is used by "ccatws1n0OLD".
@@ -4230,6 +4240,13 @@
 "csbresgOLD" is used by "csbima12gALTVD".
 "csbrngOLD" is used by "csbima12gALTVD".
 "csbxpgOLD" is used by "csbresgVD".
+"cshwordOLD" is used by "cshw0".
+"cshwordOLD" is used by "cshw0OLD".
+"cshwordOLD" is used by "cshwcl".
+"cshwordOLD" is used by "cshwidxmodOLD".
+"cshwordOLD" is used by "cshwlen".
+"cshwordOLD" is used by "cshwmodn".
+"cshwordOLD" is used by "repswcshw".
 "cvati" is used by "cvbr4i".
 "cvbr" is used by "cvbr2".
 "cvbr" is used by "cvcon3".
@@ -5004,6 +5021,8 @@
 "dipsubdi" is used by "siilem1".
 "dipsubdir" is used by "dipsubdi".
 "dipsubdir" is used by "siilem1".
+"disjxwrdOLD" is used by "disjxwwlkn".
+"disjxwrdOLD" is used by "disjxwwlksn".
 "distrlem1pr" is used by "distrpr".
 "distrlem4pr" is used by "distrlem5pr".
 "distrlem5pr" is used by "distrpr".
@@ -11681,6 +11700,10 @@
 "retbwax2" is used by "merco1lem2".
 "retbwax2" is used by "merco1lem3".
 "retbwax2" is used by "retbwax3".
+"reuccats1OLD" is used by "numclwlk2lem2f1o".
+"reuccats1OLD" is used by "numclwlk2lem2f1oOLD".
+"reuccats1OLD" is used by "reuccats1vOLD".
+"reuccats1lemOLD" is used by "reuccats1OLD".
 "rhmsubcALTV" is used by "rhmsubcALTVcat".
 "rhmsubcALTVlem1" is used by "rhmsubcALTV".
 "rhmsubcALTVlem2" is used by "rhmsubcALTVlem3".
@@ -12660,6 +12683,130 @@
 "suplem2pr" is used by "supexpr".
 "supsr" is used by "axpre-sup".
 "supsrlem" is used by "supsr".
+"swrd0fOLD" is used by "swrdidOLD".
+"swrd0fOLD" is used by "swrdn0OLD".
+"swrd0fv0OLD" is used by "2clwwlklem".
+"swrd0fv0OLD" is used by "clwlksfclwwlkOLD".
+"swrd0fv0OLD" is used by "clwwlkf".
+"swrd0fv0OLD" is used by "clwwlkinwwlk".
+"swrd0fv0OLD" is used by "clwwlknonclwlknonf1o".
+"swrd0fv0OLD" is used by "clwwlkvbij".
+"swrd0fv0OLD" is used by "clwwlkvbijOLD".
+"swrd0fv0OLD" is used by "clwwlkvbijOLDOLD".
+"swrd0fv0OLD" is used by "numclwlk2lem2f".
+"swrd0fv0OLD" is used by "numclwlk2lem2fOLD".
+"swrd0fv0OLD" is used by "wwlksnextproplem1".
+"swrd0fv0OLD" is used by "wwlksnredwwlkn0".
+"swrd0fvOLD" is used by "clwlksfclwwlkOLD".
+"swrd0fvOLD" is used by "clwwlkf".
+"swrd0fvOLD" is used by "clwwlkinwwlk".
+"swrd0fvOLD" is used by "cshwidxmodOLD".
+"swrd0fvOLD" is used by "dlwwlknonclwlknonf1olem1".
+"swrd0fvOLD" is used by "swrd0fv0OLD".
+"swrd0fvOLD" is used by "swrd0fvlswOLD".
+"swrd0fvOLD" is used by "swrdeqOLD".
+"swrd0fvOLD" is used by "swrdtrcfvOLD".
+"swrd0fvOLD" is used by "wwlksm1edg".
+"swrd0fvOLD" is used by "wwlksnred".
+"swrd0fvOLD" is used by "wwlksubclwwlk".
+"swrd0fvlswOLD" is used by "clwlksfclwwlkOLD".
+"swrd0fvlswOLD" is used by "clwwlkf".
+"swrd0fvlswOLD" is used by "clwwlkinwwlk".
+"swrd0fvlswOLD" is used by "numclwlk2lem2f".
+"swrd0fvlswOLD" is used by "numclwlk2lem2fOLD".
+"swrd0fvlswOLD" is used by "swrdtrcfvlOLD".
+"swrd0fvlswOLD" is used by "wwlksnextproplem2".
+"swrd0fvlswOLD" is used by "wwlksnredwwlkn".
+"swrd0len0OLD" is used by "wwlksnred".
+"swrd0lenOLD" is used by "addlenrevswrdOLD".
+"swrd0lenOLD" is used by "clwlkclwwlk".
+"swrd0lenOLD" is used by "clwlknf1oclwwlknlem1".
+"swrd0lenOLD" is used by "clwlksfclwwlk2sswdOLD".
+"swrd0lenOLD" is used by "clwwlkf".
+"swrd0lenOLD" is used by "clwwlkinwwlk".
+"swrd0lenOLD" is used by "cshwidxmodOLD".
+"swrd0lenOLD" is used by "efgcpbllemb".
+"swrd0lenOLD" is used by "efgredlem".
+"swrd0lenOLD" is used by "efgredlemc".
+"swrd0lenOLD" is used by "efgredlemd".
+"swrd0lenOLD" is used by "efgredleme".
+"swrd0lenOLD" is used by "efgsres".
+"swrd0lenOLD" is used by "signstfveq0".
+"swrd0lenOLD" is used by "swrd0fvlswOLD".
+"swrd0lenOLD" is used by "swrd0len0OLD".
+"swrd0lenOLD" is used by "swrdccatin12OLD".
+"swrd0lenOLD" is used by "swrdeqOLD".
+"swrd0lenOLD" is used by "wlkreslem0".
+"swrd0lenOLD" is used by "wrd2indOLD".
+"swrd0lenOLD" is used by "wrdindOLD".
+"swrd0lenOLD" is used by "wwlksm1edg".
+"swrd0lenOLD" is used by "wwlksnextproplem3".
+"swrd0lenOLD" is used by "wwlksnred".
+"swrd0lenOLD" is used by "wwlksubclwwlk".
+"swrd0swrd0OLD" is used by "swrd0swrdidOLD".
+"swrd0valOLD" is used by "efgredlem".
+"swrd0valOLD" is used by "efgredlemd".
+"swrd0valOLD" is used by "efgsres".
+"swrd0valOLD" is used by "iwrdsplitOLD".
+"swrd0valOLD" is used by "psgnunilem5OLD".
+"swrd0valOLD" is used by "signsvtn0".
+"swrd0valOLD" is used by "swrd0lenOLD".
+"swrd0valOLD" is used by "swrdccat1OLD".
+"swrd0valOLD" is used by "wlkreslem0".
+"swrd0valOLD" is used by "wrdsplex".
+"swrd0valOLD" is used by "wwlksm1edg".
+"swrdccat1OLD" is used by "ccatopthOLD".
+"swrdccat1OLD" is used by "clwwlkfo".
+"swrdccat1OLD" is used by "reuccats1OLD".
+"swrdccat1OLD" is used by "wwlksnextbi".
+"swrdccat1OLD" is used by "wwlksnextsur".
+"swrdccat3OLD" is used by "swrdccat3aOLD".
+"swrdccat3OLD" is used by "swrdccat3bOLD".
+"swrdccat3OLD" is used by "swrdccatOLD".
+"swrdccat3aOLD" is used by "swrdccatidOLD".
+"swrdccatidOLD" is used by "ccats1swrdeqbiOLD".
+"swrdccatidOLD" is used by "clwlkclwwlk2".
+"swrdccatidOLD" is used by "clwlkclwwlkfo".
+"swrdccatidOLD" is used by "clwlksfoclwwlkOLD".
+"swrdccatidOLD" is used by "numclwwlk1lem2fo".
+"swrdccatidOLD" is used by "numclwwlk1lem2foalem".
+"swrdccatin12OLD" is used by "swrdccat3OLD".
+"swrdccatin12OLD" is used by "swrdccatin12dOLD".
+"swrdccatin12lem2OLD" is used by "swrdccatin12OLD".
+"swrdccatin12lem2bOLD" is used by "swrdccatin12lem2OLD".
+"swrdccatwrdOLD" is used by "ccats1swrdeqOLD".
+"swrdccatwrdOLD" is used by "iwrdsplitOLD".
+"swrdccatwrdOLD" is used by "psgnunilem5OLD".
+"swrdccatwrdOLD" is used by "signstfveq0".
+"swrdccatwrdOLD" is used by "signsvtn0".
+"swrdccatwrdOLD" is used by "wrd2indOLD".
+"swrdccatwrdOLD" is used by "wrdindOLD".
+"swrdccatwrdOLD" is used by "wwlksnextwrd".
+"swrdeqOLD" is used by "clwlkclwwlkf1lem2".
+"swrdeqOLD" is used by "clwlksf1clwwlklemOLD".
+"swrdidOLD" is used by "cshw0".
+"swrdidOLD" is used by "cshw0OLD".
+"swrdidOLD" is used by "efgcpbllemb".
+"swrdidOLD" is used by "efgredlemc".
+"swrdidOLD" is used by "efgredleme".
+"swrdidOLD" is used by "frgpuplem".
+"swrdidOLD" is used by "splidOLD".
+"swrdidOLD" is used by "swrdccat3aOLD".
+"swrdidOLD" is used by "swrdccat3bOLD".
+"swrdidOLD" is used by "swrdccatidOLD".
+"swrdidOLD" is used by "swrdccatwrdOLD".
+"swrdidOLD" is used by "wrdcctswrdOLD".
+"swrdidOLD" is used by "wrdeqs1catOLD".
+"swrdidOLD" is used by "wrdsplex".
+"swrdn0OLD" is used by "clwlkclwwlk".
+"swrdn0OLD" is used by "clwlksfclwwlkOLD".
+"swrdn0OLD" is used by "clwwlkinwwlk".
+"swrdn0OLD" is used by "wwlksnred".
+"swrdswrd0OLD" is used by "swrd0swrd0OLD".
+"swrdtrcfv0OLD" is used by "clwlkclwwlk".
+"swrdtrcfvOLD" is used by "clwlkclwwlk".
+"swrdtrcfvOLD" is used by "swrdtrcfv0OLD".
+"swrdtrcfvlOLD" is used by "clwlkclwwlk".
 "tb-ax1" is used by "re1ax2".
 "tb-ax1" is used by "re1ax2lem".
 "tb-ax1" is used by "tbsyl".
@@ -13184,6 +13331,8 @@
 "wnfOLD" is used by "nfxfrOLD".
 "wnfOLD" is used by "nfxfrdOLD".
 "wnfOLD" is used by "stdpc5OLDOLD".
+"wrdcctswrdOLD" is used by "2clwwlk2clwwlk".
+"wrdcctswrdOLD" is used by "lencctswrdOLD".
 "wspthnonOLDOLD" is used by "wspthsnwspthsnonOLD".
 "wvd2" is used by "dfvd2".
 "wvd2" is used by "dfvd2i".
@@ -13320,6 +13469,9 @@ New usage of "2sb5nd" is discouraged (2 uses).
 New usage of "2sb5ndALT" is discouraged (0 uses).
 New usage of "2sb5ndVD" is discouraged (0 uses).
 New usage of "2sbcrexOLD" is discouraged (0 uses).
+New usage of "2swrd1eqwrdeqOLD" is discouraged (2 uses).
+New usage of "2swrd2eqwrdeqOLD" is discouraged (1 uses).
+New usage of "2swrdeqwrdeqOLD" is discouraged (2 uses).
 New usage of "2uasban" is discouraged (0 uses).
 New usage of "2uasbanh" is discouraged (1 uses).
 New usage of "2uasbanhVD" is discouraged (0 uses).
@@ -13481,6 +13633,8 @@ New usage of "addcomsr" is discouraged (8 uses).
 New usage of "adderpq" is discouraged (3 uses).
 New usage of "adderpqlem" is discouraged (1 uses).
 New usage of "addgt0sr" is discouraged (0 uses).
+New usage of "addlenrevswrdOLD" is discouraged (2 uses).
+New usage of "addlenswrdOLD" is discouraged (0 uses).
 New usage of "addltmulALT" is discouraged (0 uses).
 New usage of "addmodlteqALT" is discouraged (0 uses).
 New usage of "addnidpi" is discouraged (0 uses).
@@ -14286,6 +14440,10 @@ New usage of "cba" is discouraged (86 uses).
 New usage of "cbncms" is discouraged (5 uses).
 New usage of "cbvexsv" is discouraged (2 uses).
 New usage of "ccat2s1fvwALT" is discouraged (0 uses).
+New usage of "ccatopthOLD" is discouraged (0 uses).
+New usage of "ccats1swrdeqOLD" is discouraged (2 uses).
+New usage of "ccats1swrdeqbiOLD" is discouraged (0 uses).
+New usage of "ccats1swrdeqrexOLD" is discouraged (1 uses).
 New usage of "ccatw2s1lenOLD" is discouraged (0 uses).
 New usage of "ccatws1lenOLD" is discouraged (4 uses).
 New usage of "ccatws1lenrevOLD" is discouraged (0 uses).
@@ -14718,6 +14876,8 @@ New usage of "csbsngVD" is discouraged (0 uses).
 New usage of "csbunigVD" is discouraged (0 uses).
 New usage of "csbxpgOLD" is discouraged (1 uses).
 New usage of "csbxpgVD" is discouraged (0 uses).
+New usage of "cshwidxmodOLD" is discouraged (0 uses).
+New usage of "cshwordOLD" is discouraged (7 uses).
 New usage of "csmdsymi" is discouraged (0 uses).
 New usage of "cvati" is discouraged (1 uses).
 New usage of "cvbr" is discouraged (4 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -147746,6 +147746,473 @@ of an open range of nonnegative integers (of length equal to the length of
...
+$(
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+  Prefixes of a word
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+
+  In ~ https://www.allacronyms.com/prefix/abbreviated , "pfx" is proposed as
+  abbreviation for "prefix".  Regarding the meaning of "prefix", it is
+  different in computer science (automata theory/formal languages) compared
+  with linguistics: in linguistics, a prefix has a meaning (see Wikipedia
+  "Prefix" ~ https://en.wikipedia.org/wiki/Prefix ), whereas in computer
```
benjub, line 147758: No need for these five lines, IMO.

set.mm
```diff
@@ -147746,6 +147746,473 @@ of an open range of nonnegative integers (of length equal to the length of
...
+
+  $c prefix $.
+
+  $( Syntax for the prefix operator. $)
+  cpfx $a class prefix $.
+
+  ${
+    $d l s $.
+    $( Define an operation which extracts prefixes of words, i.e. subwords
+       starting at the beginning of a word.  Definition in section 9.1 of
```
benjub, line 147775: There is no definition number ?
Capitalize "Section"

set.mm
```diff
@@ -147746,6 +147746,473 @@ of an open range of nonnegative integers (of length equal to the length of
...
+    $d L l s $.  $d S l s $.  $d V l s $.
+    $( Value of a prefix.  (Contributed by AV, 2-May-2020.) $)
+    pfxval $p |- ( ( S e. V /\ L e. NN0 )
+                    -> ( S prefix L ) = ( S substr <. 0 , L >. ) ) $=
+      ( vs vl wcel cn0 wa cvv cv cc0 cop csubstr co cpfx wceq df-pfx a1i adantl
+      cmpt2 simpl opeq2 oveq12d elex adantr simpr ovexd ovmpt2d ) ACFZBGFZHZDEA
+      BIGDJZKEJZLZMNZAKBLZMNZOIODEIGUOTPUKDEQRULAPZUMBPZHZUOUQPUKUTULAUNUPMURUS
+      UAUSUNUPPURUMBKUBSUCSUIAIFUJACUDUEUIUJUFUKAUPMUGUH $.
+
+    $( A zero length prefix.  (Contributed by AV, 2-May-2020.) $)
```
benjub, line 147792: The zero length prefix is the empty set.

benjub:
I began, but it's rather long. Is it better to review commits independently or grouped ?

avekens:
> I began, but it's rather long. Is it better to review commits independently or grouped ?

Until now, most of the intellectual work was already done in my mathbox. The current two commits are almost only editorial work.

In the first commit, df-pfx and related theorems are moved from my mathbox to main and suffix OLD is appended to theorems which are replaced by theorems using the prefix operator. These are mainly theorems containing the pattern ( * substr <. 0 , * >. ).

In the second commit, the theorems are rearranged a little bit, and the comments are revised. Only the replacenment of OLD theorems in the proofs of the theorems mentioned in the description of this PR is of substantial interest: ~psgnunilem5 ~wrdind ~wrd2ind ~splval2 ~cshwidxmod ~wrdeqs1cat ~swrdccat3b ~swrdccat ~ccatopth ~splid ~cshw0 ~iwrdsplit : but except checking that most of their proofs are shortened nothing meaningful can be done by a review.

Therefore, I think it makes no great difference if the first commit or the whole PR is reviewed.

Maybe my next commit (revision of the splice operator) will be more interesting...

set.mm
avekens: > There is no definition number ? 

Unfortunately, there is no definition number.

>Capitalize "Section"

OK

set.mm
avekens: Which five lines exactly? The section comment consists of 10 lines...

benjub: The five quoted lines. That is, you can begin at "A prefix is an arbitrary..."

benjub:
Ok, so as I understand, there will be a next commit, and also somes checks did not pass. So I'll wait for your next commit(s). To indicate explicitly that more is coming, you can "convert to draft" in the upper-right corner, and then convert back when it's ready for review.

avekens:
> Ok, so as I understand, there will be a next commit, and also somes checks did not pass. So I'll wait for your next commit(s). To indicate explicitly that more is coming, you can "convert to draft" in the upper-right corner, and then convert back when it's ready for review.

Sorry, the intention of this (draft) PR was to get an early feadback, not a thorough review. 

I will finish the next step (revision of definition of `splice`) soon, provide the changes in a separate commit, and then clean up the material for a proper PR. The remaining tasks - replace remaining OLD theorems used in proofs (mainly of theorems of Graph Theory) - will be done in a separate PR.

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: OK. I have restructured this text a little bit, because I would like to keep this background information.

avekens:
The following proofs were shortened using the prefix operator, many of them significantly (and only three proofs became larger). It is expected that many additional proofs (which still used the OLD theorems) can be shortened, too.

- splval (687 bytes -> 657 bytes) -30
- splcl (not used) (694 bytes -> 671 bytes) -23
- splid (not used) (569 bytes -> 450 bytes) -119
- spllen (not used) (996 bytes -> 940 bytes) -56
- splfv1 (not used) (1058 bytes -> 850 bytes) -208
- splfv2a (not used) (1142 bytes -> 1064 bytes) -78
- splval2 (not used) (1510 bytes -> 1447 bytes) -63
- -----
- psgnunilem5 (3324 bytes-> 3092 bytes) shortened! -232
- wrdind (1982 bytes -> 1957 bytes) shortened! -25
- wrd2ind (5134 bytes -> 5081 bytes) shortened! -53
- cshwidxmod (3640 bytes -> 3594 bytes) -46
- wrdeqs1cat (625 bytes -> 387 bytes) -238
- swrdccat3b (841 bytes -> 825 bytes) -16
- swrdccat (2499 bytes -> 2880 bytes) + 381
- ccatopth (651 bytes -> 626 bytes) -25
- cshw0 (582 bytes -> 626 bytes) +44
- iwrdsplit (908 bytes -> 839 bytes) -69
- -----
- gsumspl ( (721 bytes -> 723 bytes) +2
- efgredleme (7237 bytes -> 7066 bytes) -271
- efgredlemd (4579 bytes -> 4489 bytes) -90
- efgredlemc (5900 bytes -> 5813 bytes) -87
- efgcpbllemb (5393 bytes -> 5317 bytes) -76
- frgpuplem (4565 bytes -> 4463 bytes) -102

In total, 1480 bytes were saved.

avekens:
This PR is teady for review. As I mentioned in #1648, the remaining OLD theorems used in proofs will be replaced after this PR is accepted and merged.

benjub:
> This PR is teady for review. As I mentioned in #1648, the remaining OLD theorems used in proofs will be replaced after this PR is accepted and merged.

I will. Hopefully tomorrow.

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+
+  A prefix is an arbitrary substring/subword starting at the
+  beginning of a string/word (see Wikipedia "Substring"
+  ~ https://en.wikipedia.org/wiki/Substring#Prefix or
+  ~ https://math.stackexchange.com/questions/2190559/
+  is-there-standard-terminology-notation-for-the-prefix-of-a-word ).
+
+  This is the meaning of "prefix" in computer science (automata theory or
+  formal languages) , whereas in linguistics, a prefix has a meaning (see
+  Wikipedia "Prefix" ~ https://en.wikipedia.org/wiki/Prefix ).
```
benjub, line 147164: If you want. But that's the same with every word: "ball", "union"... do not have the same meaning in everyday language.

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+
+  $( Syntax for the prefix operator. $)
+  cpfx $a class prefix $.
+
+  ${
+    $d l s $.
+    $( Define an operation which extracts prefixes of words, i.e. subwords
+       starting at the beginning of a word.  Definition in Section 9.1 of
+       [AhoHopUll] p. 318.  "pfx" is used as label fragment (as propose in
+       ~ https://www.allacronyms.com/prefix/abbreviated .  (Contributed by AV,
```
benjub, line 147178: "proposeD"
But I think you can remove that parenthesis anyway (there is no closing parenthesis, by the way). There is no need to justify our label choices by websites, especially nonmathematical ones.

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+       by Stefan O'Rear, 24-Aug-2015.)  (Revised by AV, 2-May-2020.) $)
+    pfxres $p |- ( ( S e. Word A /\ L e. ( 0 ... ( # ` S ) ) )
+                    -> ( S prefix L ) = ( S |` ( 0 ..^ L ) ) ) $=
+      ( vx cword wcel cc0 chash cfv co wa cpfx cfzo cv cmpt cres pfxmpt wf wrdf
+      cfz adantr cuz wss elfzuz3 adantl fzoss2 syl feqresmpt eqtr4d ) BAEFZCGBH
+      IZTJFZKZBCLJDGCMJZDNBIOBUNPDABCQUMDGUKMJZAUNBUJUOABRULABSUAUMUKCUBIFZUNUO
+      UCULUPUJCGUKUDUECGUKUFUGUHUI $.
+
+    $d V x $.  $d W x $.
+    $( A prefix of a word is a function from a half-open range of nonnegative
```
benjub, line 147244: Is "half-open" the terminology used in set.mm ? (because for integer intervals, there is no notion of open or closed, but maybe it was decided in the past to mean "upper bound excluded" ?) So do as you see fit.

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+                 -> ( W prefix L ) =/= (/) ) $=
+    ( cword wcel cn chash cfv cle wbr w3a cpfx co c0 wne cc0 cfzo 3ad2ant2 wceq
+    cn0 lbfzo0 ne0i sylbir wf wb cfz simp1 nnnn0 lencl 3ad2ant1 simp3 syl3anbrc
+    elfz2nn0 pfxf syl2anc f0dom0 bicomd syl necon3bid mpbird ) CBDEZAFEZACGHZIJ
+    ZKZCALMZNOPAQMZNOZVBVAVHVDVBPVGEVHAUAVGPUBUCRVEVFNVGNVEVGBVFUDZVFNSZVGNSZUE
+    VEVAAPVCUFMEZVIVAVBVDUGVEATEZVCTEZVDVLVBVAVMVDAUHRVAVBVNVDBCUIUJVAVBVDUKAVC
+    UMULABCUNUOVIVKVJVFVGBUPUQURUSUT $.
+
+  $( The value of the prefix extractor is the empty set (undefined) if the
+     argument is not within the range of the word.  (Contributed by AV,
```
benjub, line 147316: Maybe I am not using the correct words, but I'd prefer something like:
> The value of a prefix operation when the integer argument is larger than the word length is the empty set. (This is due to our definition of function values for out-of-domain arguments, see ~iotanul ).

It was probably discussed already, but: have you thought about the definition which would give the whole word in that case (and more generally, similarly for substrings) ?

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+    ( cword wcel cn0 chash cfv clt wbr w3a cpfx co cc0 cop csubstr c0 pfxval cz
+    wceq 3adant3 cle w3o simp1 0zd nn0z 3ad2ant2 3jca 3mix3 3ad2ant3 sylc eqtrd
+    swrdnd ) CBDZEZAFEZCGHAIJZKZCALMZCNAOPMZQUOUPUSUTTUQCAUNRUAURUONSEZASEZKNNI
+    JZANUBJZUQUCZUTQTURUOVAVBUOUPUQUDURUEUPUOVBUQAUFUGUHUQUOVEUPUQVCVDUIUJNABCU
+    MUKUL $.
+
+  $( Length of a prefix of a word reduced by a single symbol.  (Contributed by
+     AV, 4-Aug-2018.)  (Revised by AV, 3-May-2020.) $)
+  pfxlen0 $p |- ( ( W e. Word V /\ L e. NN0 /\ ( # ` W ) = ( L + 1 ) )
+                  -> ( # ` ( W prefix L ) ) = L ) $=
```
benjub, line 147329: isn't it enough to suppose `L <= length(W)` ?

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+     (Contributed by AV, 21-Oct-2018.)  (Revised by AV, 3-May-2020.) $)
+  addlenpfx $p |- ( ( W e. Word V /\ M e. ( 0 ... ( # ` W ) ) )
+               -> ( ( # ` ( W prefix M ) )
+                  + ( # ` ( W substr <. M , ( # ` W ) >. ) ) ) = ( # ` W ) ) $=
+    ( cword wcel cc0 chash cfv cfz co wa cpfx cop csubstr caddc pfxlen swrdrlen
+    cmin cc nn0cnd oveq12d wceq elfznn0 lencl pncan3 syl2anr eqtrd ) CBDEZAFCGH
+    ZIJEZKZCALJGHZCAUIMNJGHZOJAUIARJZOJZUIUKULAUMUNOBCAPABCQUAUJASEUISEUOUIUBUH
+    UJAAUIUCTUHUIBCUDTAUIUEUFUG $.
+
+  $( The first symbol in a prefix of a word.  (Contributed by Alexander van der
```
benjub, line 147356: Maybe
> The first symbol of a prefix is the first symbol of the word.

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+     Alexander van der Vekens, 16-Jun-2018.)  (Revised by AV, 3-May-2020.) $)
+  pfxtrcfv0 $p |- ( ( W e. Word V /\ 2 <_ ( # ` W ) )
+                     -> ( ( W prefix ( ( # ` W ) - 1 ) ) ` 0 ) = ( W ` 0 ) ) $=
+    ( cword wcel c2 chash cfv cle wbr wa c0 wne cc0 c1 cmin co cfzo cpfx wceq
+    cz simpl wrdlenge2n0 cn cuz 2z a1i lencl nn0zd adantr simpr eluz2 syl3anbrc
+    uz2m1nn syl lbfzo0 sylibr pfxtrcfv syl3anc ) BACDZEBFGZHIZJZUSBKLMMUTNOPZQP
+    DZMBVCRPGMBGSUSVAUAABUBVBVCUCDZVDVBUTEUDGDZVEVBETDZUTTDZVAVFVGVBUEUFUSVHVAU
+    SUTABUGUHUIUSVAUJEUTUKULUTUMUNVCUOUPMABUQUR $.
+
+  $( The last symbol in a (not empty) prefix of a word.  (Contributed by
```
benjub, line 147385: in a nonempty prefix

set.mm
```diff
@@ -147154,6 +147148,405 @@ of an open range of nonnegative integers (of length equal to the length of
...
+    cn pfxfv syl3anc 3eqtrd ) CBDZEZAFCGHZIJZEZKZCALJZMHZVAGHZFNJVAHZAFNJZVAHZV
+    ECHZUTVAUOEZVBVDOUPVHUSBCAPQVAUOUASUTVCAFVANUSUPATUQIJZEZVCAOURVIAUQUBUCZBC
+    AUDUEUFUTUPVJVETAUGJEZVFVGOUPUSUHUSVJUPVKRUSVLUPUSAUKEVLAUQUIAUJSRVEABCULUM
+    UN $.
+
+  ${
+    $d M i $.  $d N i $.  $d U i $.  $d V i $.  $d W i $.
+    $( The prefixes of two words are equal iff they have the same length and
+       the same symbols at each position.  (Contributed by Alexander van der
+       Vekens, 7-Aug-2018.)  (Revised by AV, 4-May-2020.) $)
```
benjub, line 147400: That's true for any word. Is it useful to have it for prefixes ?

benjub:
I made a few comments. So it's kind of reviewed up to the latest comment I wrote (Line 147400). I have to say the diff is huge and I do not have much to say in that topic. By the way, why are all these things about words in the part "Real and complex numbers"? Can someone review a bit more ? Maybe someone who took part in #1648 can chime in ?

set.mm
avekens: I added it (and would like to keep it) mainly because the Wikipedia page "prefix" is about the linguistic meaning...

set.mm
avekens: Yes, this is the offical wording used in set.mm, see "5.5.7  Half-open integer ranges".

set.mm
avekens: > isn't it enough to suppose `L <= length(W)` ?

This is covered by ~ pfxlen.

But you are right, ~pfxlen0 makes no sense, and its predecessor swrd0len0(OLD) is only used once. Therefor I will delete ~pfxlen0.

set.mm
avekens: This theorem (or its predecessor ~ swrdeqOLD) is used 3 times, therefore I would like to keep it.

set.mm
avekens: Done

set.mm
avekens: Done

set.mm

set.mm
avekens: Comment revised.

> It was probably discussed already, but: have you thought about the definition which would give the whole word in that case (and more generally, similarly for substrings) ?

No, I did not consider this, and I don't know if this was discussed by others. I cannot see any advantage of this at the moment, and the definition(s) would become wore complicated.

set.mm
avekens: > "proposeD" But I think you can remove that parenthesis anyway (there is no closing parenthesis, by the way). 

Done

>There is no need to justify our label choices by websites, especially nonmathematical ones.

I added this because there was a long discussion about the label in #1648.

avekens:
> I made a few comments. So it's kind of reviewed up to the latest comment I wrote (Line 147400). 

@benjub thank you very much for your review. I considered/commented your remarks as far as possible.

>I have to say the diff is huge and I do not have much to say in that topic. 

A review is meaningful only up to line 148000 - the rest is almost only technical.

> By the way, why are all these things about words in the part "Real and complex numbers"? 

I do not know why it was placed here (in 2014?), but since the main prerequisites are the integers and the size function, which are also topics of this part and definied immediatly before words, I think it is a good place. If this should be discussed further, this should be done in another place (open an issue for that?). 

> Can someone review a bit more ? Maybe someone who took part in #1648 can chime in ?

@savask @david-a-wheeler do you want to have a (quick) look at this PR?

savask:
I looked at the changes. I don't have any serious remarks, besides perhaps the comment for `df-pfx`:
```
    $( Define an operation which extracts prefixes of words, i.e. subwords
       starting at the beginning of a word.  Definition in Section 9.1 of
       [AhoHopUll] p. 318.  "pfx" is used as label fragment as proposed in
       ~ https://www.allacronyms.com/prefix/abbreviated .  (Contributed by AV,
       2-May-2020.) $)
```
In my opinion, the reference to [AhoHopUll] is not required as the definition of a prefix is intuitive enough and, moreover, the definition was given at the start of the section `Prefixes of a word`. If you want to keep the reference it is much better to put it at the beginning of a section (replacing the stackexchange link, for example). I also suggest removing the reference to allacronyms.com completely, as it's not a credible source and we don't usually express any justification for using a specific abbreviation in set.mm.

On the other hand it might be beneficial to hint the way this function works, for instance, like this:
```
Define an operation which extracts prefixes of words, i.e. subwords
starting at the beginning of a word. In other words, ` ( S prefix L ) ` is
a prefix of ` S ` of length ` L `.  (Contributed by AV, 2-May-2020.) 
```

On a more general note, I can't get rid of the feeling that it's a huge bulk of technical work without any immediate merit. Sure some theorems are shorter, but it would be great to see a nontrivial result involving prefixes (something like the "commuting strings theorem", i.e. `xy = yx` implies `x = z^n` and `y = z^m` for some string `z`. The proof uses prefixes, as far as I remember).

set.mm
benjub: Ok. Then, "has a DIFFERENT meaning"

benjub:
> ` ( S prefix L ) ` is a prefix of ` S ` of length ` L `.

Maybe use the definite article, and add a provision: ` ( S prefix L ) ` is the prefix of ` S ` of length ` L ` (when the natural `L` is not larger than the length of the word `S`).

set.mm
```diff
@@ -147154,6 +147148,398 @@ of an open range of nonnegative integers (of length equal to the length of
...
+    WDUDNDZWGWQRWFWDUEDZXBWBWEXCWCWBWECULUFZXCCWAUGBCUHUMUIWDUJSAWJBCUKUNWFWIWP
+    XAWFWIPZWOWTWKXEWOWRVBZWSVBZKZWTXEWMXFWNXGXEWBXDPZWMXFKWFXIWIWBWEXIWCWBWEXD
+    WBWEXDCWAUOUPUQUIURBCUSSXEWNXGKZAWHLMNZWHUCZONZXGKZWIWFXNWIWFWBWCEWHHIZJZXN
+    WIWEXOWBWCWDWHEHVCUTXPWCAULUFZPZXNWCXOXRWBWCXOXQWCXOXQAWAUOUPUQVABAUSSVDVEW
+    IXJXNRWFWIWNXMXGWIWLXLAOWIWJXKWDWHWDWHLMVFWIVGVHVIVJVKVNVLXEWRTDWSTDXHWTRXE
+    CQVMAQVOTWRWSVPVQVRVSVTVR $.
+
+  ${
+    $d N y $.  $d V x $.  $d x y $.
+    $( Sets of words are disjoint if each set contains extensions of distinct
```
benjub, line 147474: contain ONLY

set.mm
```diff
@@ -147154,6 +147148,398 @@ of an open range of nonnegative integers (of length equal to the length of
...
+       words of a fixed length.  (Contributed by AV, 29-Jul-2018.)  (Revised by
+       AV, 6-May-2020.) $)
+    disjwrdpfx $p |- Disj_ y e. W { x e. Word V | ( x prefix N ) = y } $=
+      ( cword cv cpfx co invdisjrab ) ABEDFAGCHIJ $.
+  $}
+
+  ${
+    $d x A $.  $d x S $.  $d x Y $.  $d x Z $.
+    $( Joining a prefix with an adjacent subword makes a longer prefix.
```
benjub, line 147483: Joining --> Concatenating
(I haven't checked other occurrences, but let's be consistent with "concatenate", "concatenation".)

set.mm
```diff
@@ -147154,6 +147148,398 @@ of an open range of nonnegative integers (of length equal to the length of
...
+      YRVLRZUVCUVKUVJEZUVKCCNHZYLJHZEZUVCUUNUUAOUVPUVCUUAUUNXOUUAUUNUURWGWHUUEC
+      DCWIWJXOUVMUVPWKUUNXOUVJUVOUVKXOFUVNYLJXKXJFUVNMXNXKUVNFXKCXKCUUBPWLWMVTW
+      NWBRWPWOABCDUVDWQWRUVCUVFUUEBUVCUVFUVKCLHUUEUVCUVDUVKCLUVLWNUVCUUECUUNUUE
+      VEEXOUUNUUEUUECDWSPWTXOYSUUNUUCRXAQXBXCXDWRUUGXJXNUUFUUJUUIMXOXJUUFYNRXJX
+      KXNUUFXGUUQUUEDABXFTXHXE $.
+  $}
+
+  ${
+    $d k B $.  $d k S $.  $d k T $.
+    $( Recover the left half of a concatenated word.  (Contributed by Mario
```
benjub, line 147517: Maybe:
The left "concatenand" of a concatenated word is the prefix of that word of its own length.

set.mm
```diff
@@ -147154,6 +147148,398 @@ of an open range of nonnegative integers (of length equal to the length of
...
+      cuz syl2an fzoss2 fnssres wrdfn cv fvres adantl ccatval1 3expa eqfnfvd )
+      BAUAZEZCVMEZFZBCUBSZBGHZUCSZVQIVRJSZUDZBVPVQVMEVRIVQGHZKSZEVSWALABCUEVPVR
+      IVRCGHZUFSZKSZWCVPVRMEZWDMEZFVRIVRKSEZVRWFEVNWGVOWHABNZACNZUGVNWIVOVNWGWI
+      WJVRUKUHOVRWDVRUIUJVPWBWEIKABCULUMUNAVQVRUOPVPDVTWABVPVQIWEJSZQVTWLUPZWAV
+      TQBCAUQVPWEVRVBHZEZWMVNVRWNEZWHWOVOVNVRUREWPVNVRWJUSVRUTRWKWDVRVRVAVCVRIW
+      EVDRWLVTVQVEPVNBVTQVOABVFOVPDVGZVTEZFWQWAHZWQVQHZWQBHZWRWSWTLVPWQVTVQVHVI
+      VNVOWRWTXALABCWQVJVKTVLT $.
+  $}
+
+  $( A prefix of length 1.  (Contributed by AV, 15-May-2020.) $)
```
benjub, line 147533: Not sure I'm using the correct words, but maybe:
The prefix of length one of a given nonempty word expressed as a single character word.

set.mm
```diff
@@ -147273,9 +147660,24 @@ of an open range of nonnegative integers (of length equal to the length of
     LUSUQVOURBHCUCRVAUDSUEUTVMVNHABCDEUFUGUHVHVJVFENVHVIBVEUTVIBPZVCUSUQVPURUSB
     QGZVPBCUIVQBBUJUKSRULUMUNUOUP $.
 
-  $( A subword of a prefix.  (Contributed by Alexander van der Vekens,
-     6-Apr-2018.) $)
-  swrdswrd0 $p |- ( ( W e. Word V /\ N e. ( 0 ... ( # ` W ) ) )
+  $( A prefix of a subword.  (Contributed by AV, 2-Apr-2018.)  (Revised by AV,
```
benjub, line 147663: A prefix of a subword is a subword.

set.mm
```diff
@@ -147288,24 +147690,66 @@ of an open range of nonnegative integers (of length equal to the length of
     MFZWKCGVGTWLWECWLCCUNULUOOPZRSVIVLWHBVICWEAHWMRSUPUQABGCDEURUSVRWAVOELVRVSA
     VTBVRAVNAUTFZVIVKWNVMVKAAGCTVAQPVBVRBVNBUTFZVIVMWOVKVMBBACTVAPPVBVCRVDVE $.
 
-  $( A prefix of a prefix.  (Contributed by Alexander van der Vekens,
-     7-Apr-2018.) $)
-  swrd0swrd0 $p |- ( ( W e. Word V /\ N e. ( 0 ... ( # ` W ) )
+  $( A subword of a prefix.  (Contributed by Alexander van der Vekens,
```
benjub, line 147693: A subword of a prefix is a subword.

set.mm
```diff
@@ -147288,24 +147690,66 @@ of an open range of nonnegative integers (of length equal to the length of
...
-    ISGUJFZUKSDGBMNKGAMZNKDUNNKOUGUIUKPULUMUKUIUGUMUKUIBTFUMBUHQBRUAUBUGUIUKUCU
-    DGABCDUEUF $.
+    wa cn0 syl 3ad2ant2 simp3 jca swrdswrd0OLD sylc ) DCEFZBGDHIZJKFZAGBJKZFZLZ
+    UGUISGUJFZUKSDGBMNKGAMZNKDUNNKOUGUIUKPULUMUKUIUGUMUKUIBTFUMBUHQBRUAUBUGUIUK
+    UCUDGABCDUEUF $.
 
-  $( A prefix of a prefix with the same length is the prefix.  (Contributed by
-     AV, 5-Apr-2018.)  (Proof shortened by AV, 14-Oct-2018.) $)
-  swrd0swrdid $p |- ( ( W e. Word V /\ N e. ( 0 ... ( # ` W ) ) )
+  $( A prefix of a prefix.  (Contributed by Alexander van der Vekens,
```
benjub, line 147721: A prefix of a prefix is a prefix.

set.mm
```diff
@@ -147288,24 +147690,66 @@ of an open range of nonnegative integers (of length equal to the length of
...
+  swrd0swrdidOLD $p |- ( ( W e. Word V /\ N e. ( 0 ... ( # ` W ) ) )
                         -> ( ( W substr <. 0 , N >. ) substr <. 0 , N >. )
                            = ( W substr <. 0 , N >. ) ) $=
     ( cword wcel cc0 chash cfv cfz co cop csubstr wceq cn0 elfznn0 nn0fz0 sylib
-    adantl swrd0swrd0 mpd3an3 ) CBDEZAFCGHZIJEZAFAIJEZCFAKZLJZUELJUFMUCUDUAUCAN
-    EUDAUBOAPQRAABCST $.
+    adantl swrd0swrd0OLD mpd3an3 ) CBDEZAFCGHZIJEZAFAIJEZCFAKZLJZUELJUFMUCUDUAU
+    CANEUDAUBOAPQRAABCST $.
+
+  $( A prefix of a prefix with the same length is the prefix.  (Contributed by
```
benjub, line 147746: is the ORIGINAL prefix. In other words, the operation "prefix of length N" is idempotent.

set.mm
```diff
@@ -147314,83 +147758,180 @@ of an open range of nonnegative integers (of length equal to the length of
...
+    nn0fz0 3jca ccatswrd swrdidOLD 3eqtrd ) BACZDZBUAUBZUCZBEBUDFZGUEHZIJHZBUFF
+    ZKZLHWFBWEWDIZJHZLHZBEWDIJHZBWCWHWJWFLWCBWEWEGUGHZIZJHZWEBFZKZWJWHWAWBWEEWD
+    UHHDZWOWQQWCWDUIDZWRABUJZWDUKMAWEBULNWCWIWNBJWCWDWMWEWCWSWDWMQWTWSWMWDWSWDG
+    WDUMWSUNUOUPMVEOWCWGWPWAWGWPQWBBVTUQPURUSOWAWBEEWERHDZWEEWDRHZDZWDXBDZUTZWK
+    WLQWCWSXEWTWSXAXCXDWSWESDXAWDVAWEVBMWSGXBDZWDGWDRHDZXCWSGSDZWDSDZGWDVCVDXFX
+    HWSVFVGWDVHZWDVIGWDVJVKWSXGWDVLTWDGWDVMVNWSXIXDXJXIXDWDVOTMVPMABEWEWDVQNWAW
+    LBQWBABVRPVS $.
+
+  $( The last symbol of a word concatenated with the word with the last symbol
+     removed having results in the word itself.  (Contributed by Alexander van
```
benjub, line 147852: delete "having"

set.mm
```diff
@@ -147314,83 +147758,180 @@ of an open range of nonnegative integers (of length equal to the length of
...
+    breq2 3ad2ant2 mpbid swrdccatwrdOLD syl2anc eqtrd adantr ex ) CBUAZDZAVMDZA
+    EFZCEFZGUBHZIZUCZCAJVQKZLHZIZACAULFUDZMHZIVTWCUEWEWBWDMHZAWCWEWFIVTCWBWDMNU
+    FVTWFAIWCVTWFAJVPGQHZKZLHZWDMHZAVTWBWIWDMVTWAWHALVTVQWGJVTWGVRGQHZVQVSVNWGW
+    KIVOVPVRGQNOVNVOWKVQIZVSVNVQUGDZVQUHDWLBCUIZVQUJVQUKUMPRUNUOUPVTVOAUQURZWJA
+    IVNVOVSUSVTJVPSTZWOVTWPJVRSTZVNVOWQVSVNWMWQWNVQUTVAPVSVNWPWQVBVOVPVRJSVEOVC
+    VOVNWPWOVBVSAVMVDVFVGBAVHVIVJVKRVL $.
+
+  ${
+    $d U s $.  $d V s $.  $d W s $.
+    $( There exists a symbol such that its concatenation with the prefix
```
benjub, line 147887: "its concatenation AFTER" is clearer: on first reading, I just thougth "how can this be true?" since I assumed "EARLIER" was meant.
The statement looks overly complex, but maybe it's used as a lemma in this form later.

set.mm
```diff
@@ -148416,6 +149653,19 @@ Splicing words (substring replacement)
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 $)
 
+  $c reverse $.
+
+  $( Syntax for the word reverse operator. $)
```
benjub, line 149658: I think a better name for this operation is "word reversal"

set.mm
```diff
@@ -148416,6 +149653,19 @@ Splicing words (substring replacement)
...
+  creverse $a class reverse $.
+
+  ${
+    $d s x $.
+    $( Define an operation which reverses the order of symbols in a word.
+       (Contributed by Stefan O'Rear, 26-Aug-2015.) $)
```
benjub, line 149664: Maybe add some synonyms, e.g., `This operation is also known as "word reversal" and "word mirroring".` This way, it will show up when someone searches for these expressions.

set.mm
```diff
@@ -147154,6 +147148,398 @@ of an open range of nonnegative integers (of length equal to the length of
...
+    $( Closure of the prefix extractor.  (Contributed by AV, 2-May-2020.) $)
+    pfxcl $p |- ( S e. Word A -> ( S prefix L ) e. Word A ) $=
+      ( vx vs vl cword wcel cpfx co c0 eleq1 wne cn0 cv wex n0 cc0 cop csubstr
+      wi cvv df-pfx elmpt2cl2 idi exlimiv sylbi wa pfxval swrdcl adantr eqeltrd
+      sylan2 wrd0 a1i pm2.61ne ) BAGZHZBCIJZUQHZKUQHZUSKUSKUQLUSKMZURCNHZUTVBDO
+      ZUSHZDPVCDUSQVEVCDVEVCUAEFUBNEORFOSTJBCIVDEFUCUDUEUFUGURVCUHUSBRCSTJZUQBC
+      UQUIURVFUQHVCABRCUJUKULUMVAURAUNUOUP $.
+
+    $d A x $.
+    $( Value of the prefix extractor as mapping.  (Contributed by AV,
```
benjub, line 147222: I think it should be "as a mapping".

set.mm
```diff
@@ -148551,6 +149801,20 @@ Splicing words (substring replacement)
...
+  $( Extend class notation with words consisting of one repeated symbol. $)
+  creps $a class repeatS $.
+
+  ${
+    $d n s x $.
+    $( Definition to construct a word consisting of one repeated symbol, often
+       called "repeated symbol word" for short in the following.  (Contributed
+       by Alexander van der Vekens, 4-Nov-2018.) $)
+    df-reps $a |- repeatS
+                 = ( s e. _V , n e. NN0 |-> ( x e. ( 0 ..^ n ) |-> s ) ) $.
```
benjub, line 149815: It looks more and more specialized. I hope there are (future?) applications to justify these definitions.

set.mm
```diff
@@ -150680,6 +152142,35 @@ the symbol at any position is repeated at multiples of L (modulo the
       BCWCVOVPVJVPWCQWCVPRSTVPVLHZWBVMCWDVJVPVJVLWDVJRWDQSTVAVB $.
   $}
 
+  ${
+    $d W i $.
+    $( A prefix of length 2.  (Contributed by AV, 15-May-2020.) $)
```
benjub, line 152147: 2 --> two

set.mm
```diff
@@ -150680,6 +152142,35 @@ the symbol at any position is repeated at multiples of L (modulo the
...
+      KUUAYQUUCYIEYJWBYSYOCYPUUBYSYPUUJUUBYIEITWCWDWEWFWGWIWAWHYHXSXNDZYBXNDYCY
+      RVOXOUUSXRABEWJPYHXTYAAXRXOIIXPTJZDZXTADXRXPUSDZUVAXRYEYFXQUQZUVBYGUVCYFK
+      XPUBHZMUVBUVCYFUVDYEYFXQWKYFXQUVDYEYFXQUVDYFUUPXQUVDVKYFKWLDZEWLDZXPWLDZU
+      QZUUPXQMZUVDWMYFUVEUVFUVGYFWNUVFYFWOOXPWPVCZKEXPWQQXBWRWSWTXPXAXCXDZXPVAX
+      CIABXMXEXRXOKUUTDZYAADXRUUOUVBKXPVIHZUVLUUOXRVJOUVKXRUVCUVMYGYEYFXQUVMYEY
+      FMZUUPXQUVMVKUVNUVHUVIUVMWMYFUVHYEUVJWIKEXPXFQXBXGXDKXPVLUJKABXMXEXHXSCAY
+      BXIXJXKXL $.
+  $}
+
   $( A word of length 3 represented as triple of ordered pairs.  (Contributed
```
benjub, line 152174: 3 --> three
(I may be a bit rigid... but to me, that's a bit like writing in all caps.)

set.mm
```diff
@@ -150740,10 +152231,41 @@ the symbol at any position is repeated at multiples of L (modulo the
 
   $( Two words of length at least 2 are equal if and only if they have the same
```
benjub, line 152232: 2 --> two

benjub:
I added a few more comments, and I reviewed the whole thing (although not in depth). I'm not formally approving since there are a few comments by @savask and I above that require changes. But after they are resolved, it'll be ok for me.

set.mm
avekens: The whole section header comment is removed now. Some of its contents is moved to the comment of df-pfx.

set.mm
avekens: The whole section header comment is removed now.

avekens:
Thank you, @savask , for your remarks. I considered them in my latest commit.

> I looked at the changes. I don't have any serious remarks, besides perhaps the comment for `df-pfx`:
> 
> ```
>     $( Define an operation which extracts prefixes of words, i.e. subwords
>        starting at the beginning of a word.  Definition in Section 9.1 of
>        [AhoHopUll] p. 318.  "pfx" is used as label fragment as proposed in
>        ~ https://www.allacronyms.com/prefix/abbreviated .  (Contributed by AV,
>        2-May-2020.) $)
> ```
> 
> In my opinion, the reference to [AhoHopUll] is not required as the definition of a prefix is intuitive enough and, moreover, the definition was given at the start of the section `Prefixes of a word`. If you want to keep the reference it is much better to put it at the beginning of a section (replacing the stackexchange link, for example). I also suggest removing the reference to allacronyms.com completely, as it's not a credible source and we don't usually express any justification for using a specific abbreviation in set.mm.

The whole section header comment is removed now (as for the other subsections in this section) - the link to Wikipedia is moved to the comment for the definition ~df-pfx. 

To have the reference in the comment of the definition is a common practise in metamath, see, for example, df-iun: "Definition indexed union in [Stoll] p. 45.". I think also obvious/intuitive definitions should be referenced if possible.

The hint regarding the label fragment is also removed. All these information are still available in issue #1648.

> On the other hand it might be beneficial to hint the way this function works, for instance, like this:
> 
> ```
> Define an operation which extracts prefixes of words, i.e. subwords
> starting at the beginning of a word. In other words, ` ( S prefix L ) ` is
> a prefix of ` S ` of length ` L `.  (Contributed by AV, 2-May-2020.) 
> ```

I added this hint to the comment.

> On a more general note, I can't get rid of the feeling that it's a huge bulk of technical work without any immediate merit. Sure some theorems are shorter, but it would be great to see a nontrivial result involving prefixes (something like the "commuting strings theorem", i.e. `xy = yx` implies `x = z^n` and `y = z^m` for some string `z`. The proof uses prefixes, as far as I remember).

Everybody is invited to prove and contribute this theorem...

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: Done (The prefix of length one of a nonempty word expressed as a singleton word.)

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: Done

set.mm
avekens: I did not create this definition, I just moved it to the corresponding subsection. Since this does not belong to the topic of this PR, I do not want to change this (without agreement of the contributor).

set.mm
avekens: Done.

set.mm
avekens: Done

set.mm
avekens: This definition is not new, also only moved to the corresponding subsection. It is used to prove the friendship theorem.

set.mm
```diff
@@ -149815,6 +151265,18 @@ the symbol at any position is repeated at multiples of L (modulo the
       KYNWMXJEYNWLXIJKYNXIWLWJWPXIWLSXRWDUTVAVBVCWHVGVEVHVIWEBXBDEXSVNTWFWI $.
   $}
 
+  $( Mapping of words commutes with the prefix operation.  (Contributed by AV,
```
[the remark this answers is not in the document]
avekens: "mapping of words" is often used, I added the suggested explanation at its first occurence.

set.mm
avekens: 2 replaced by two.

set.mm

avekens:
I think all review remarks are considered now (as far as possible), so that this PR should be accepted and merged afterwards.

Thanks again, @benjub and @savask, for your valuable remarks.

set.mm
avekens: The comment was written by Mario, I  do not want to change it without his acceptance. In Addition, I think the word "concatenand" is not commonly used.

set.mm
avekens: Done

set.mm
avekens: Done

benjub:
I'm getting the unicorn page more and more often lately (regardless of the PR, and even for small commits). Is it the same with you ?
When I can see the new commits, I'll probably approve.

avekens:
> I'm getting the unicorn page more and more often lately (regardless of the PR, and even for small commits). Is it the same with you ? When I can see the new commits, I'll probably approve.

What is the "unicorn page"? I have (and had) no problems with GitHub, except that the large complete diff is not shown (message "Oops, something went wrong.  Retry").

savask:
@benjub
>I'm getting the unicorn page more and more often lately (regardless of the PR, and even for small commits). Is it the same with you ?

I also haven't been able to check the new diff or even the raw set.mm file from the pull request page, but the new set.mm version is available in the relevant branch of avekens' fork.

benjub (approved):
Approved.  Thanks @avekens for the careful work.

benjub:
> What is the "unicorn page"? I have (and had) no problems with GitHub, except that the large complete diff is not shown (message "Oops, something went wrong. Retry").

 Right now, trying to access `https://github.com/metamath/set.mm/pull/2890/commits/4cc64b319ad5586d008a5a2f06bb52932b00ee77` I get what I call the "unicorn page" since there is a picture of an angry unicorn with the text:
```
This page is taking too long to load.

Sorry about that. Please try refreshing and contact us if the problem persists.
```
And at other times, I also get the message  "Oops, something went wrong. Retry".

But I checked on your branch as @savask suggested, so that's ok.

PR #2895, opened 2022-10-15
Banach and Hilbert spaces

benjub:
Everyone, and especially @avekens : this is a heads up that I'm going to split the subsection "Banach & Hilbert".

In this first commit, I saw that some labels do not respect the convention(?) cxxx/df-xxx for the syntax/definition labels
```
cxmt/df-xmet
cme/df-met
cxme/df-xms
cmt/df-ms
ctmt/df-tms
cca/df-cau
cms/df-cmet
chs/df-hil
```
I propose to relabel to:
```
cxmet/df-xmet
cmet/df-met
cxms/df-xms
cms/df-ms
ctms/df-tms
ccau/df-cau
ccmet/df-cmet
chil/df-hil
```


avekens (approved):

benjub:
For the latest commit: let's pretend that the revision and the shortening were just one thing.

benjub:
Good with you, @avekens ? (since there are global changes, I'm a bit worried of merge conflicts, so the sooner we merge, the better)

avekens:
What is the meaning of the label change "chil-->chba"? There are a lot of changes in the commit.

benjub:
> What is the meaning of the label change "chil-->chba"? There are a lot of changes in the commit.

What used to be `chil` was the syntax axiom for the definition `df-hba`, and I needed to "free" that label in order to give that label to the syntax axiom for the definition `df-hil` (which was `chs`).

benjub:
> There are a lot of changes in the commit.

`chba` is the first label of `PART 19  COMPLEX HILBERT SPACE EXPLORER (DEPRECATED)` so it is basically used everywhere in that part (its last usage is in `cdj3i`, which is the last statement of that part).

avekens (approved):
Because there are so many changes, I could not check everything in detail. But what I checked is OK.

benjub:
> Because there are so many changes, I could not check everything in detail. But what I checked is OK.

Thanks. The changes are large, but each commit is the result of an automatic "search and replace", and I was careful (case sensitive, keyword preceded and followed by a space).

PR #2903, opened 2022-10-16
prove euf without ax-13

wlammen:


benjub:
I realize that I had missed equvelv. I think sbceqal1 does not add anything to equvelv and could be removed. What do you think ? eu6 can be proved (and shortened) by using equvelv directly.

wlammen:
sbeqal1 does not add anything of value, has just one user, and can therefore be deleted.  eu6 has more potential for shortenings, more than I already did, and using equvelv does not hurt.  But I've got no time any more today.

benjub:
Actually, equvelv could probably be stated in commuted form:
```
equvelv $p |- ( A. z ( z = x -> z = y ) <-> x = y ) $=
```
This would remove ~bicomi from its proof and from its only usage in eu6.
Can you do that, and remove sbeqal1 since we both agree ? (there is a version kept as as-sbeqal1, so no need for *OLD versions)

benjub:
And yes, eu6 could probably be shortened (probably results involving substitution could help, but it may be nice to keep a version avoiding df-sb... there could be ALT versions if that case happens).

wlammen:
I've got no time any more.

benjub:
Other TODOs (for you or me):
* edit comment of exmoeuOLD as usual
* the changes in this PR remove dependency on ax-11, ax-13 from some OLD theorems; check that the corresponding "NEW" theorems also do not have these dependencies, and if they do, either remove the dependencies as well, or revert to the OLD version.

**edit:** Sorry, didn't see your "no time" above. I'm merging, and doing the above TODOs.
**edit:** Done in #2904, including the check that OLD versions do not have lighter axiom usage than "NEW" versions.

benjub (approved):

PR #2904, opened 2022-10-16
Uniqueness 

benjub:
Minor edits following #2903.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17286,7 +17286,6 @@ New usage of "sbcoreleleq" is discouraged (2 uses).
 New usage of "sbcoreleleqVD" is discouraged (0 uses).
 New usage of "sbcrexgOLD" is discouraged (2 uses).
 New usage of "sbcssgVD" is discouraged (0 uses).
-New usage of "sbeqal1OLD" is discouraged (0 uses).
 New usage of "sbequ8ALT" is discouraged (0 uses).
 New usage of "sbtT" is discouraged (0 uses).
 New usage of "scmateALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen (approved):

set.mm
```diff
@@ -17242,26 +17242,19 @@ it were an axiom (as it is in the case of Tarski).
       ( weq wex wa ax6evr equtr ancrd eximdv mpi ) ABDZBCDZCEACDZMFZCECBGLMOCLM
       NABCHIJK $.
 
-    $( A specialized version of ~ equvel with distinct variable restrictions
-       and fewer axiom usage.  (Contributed by Andrew Salmon, 2-Jun-2011.)
+    $( A biconditional form of ~ equvel with DV conditions and proved from
```
wlammen, line 17245: the dv shorthand is lowercase elsewhere.  Is it explained somewhere, so people new to Metamath know how to interpret it?

wlammen (changes requested):

set.mm
avekens: > the dv shorthand is lowercase elsewhere. Is it explained somewhere, so people new to Metamath know how to interpret it?

That's partially true: "dv condition" appears 174 times, "DV condition" 30 times. I would propose to unify it to "DV condition".

Our conventions contain the list item:

<li>No two variables occurring in the LHS may share a
       disjoint variable (DV) condition.
       </li>

(This should read "distinct variable (DV) condition", shouldn't it? (see https://us.metamath.org/mpeuni/mmset.html#distinct). But we have (at least) 11 "disjoint variable condition" in set.mm (and 106 "distinct variable condition").

wlammen: Disjoint is my favorite here, because of $d x ph $.  What about ~ DV and a fake theorem page explaining it?

avekens: The error message of the metamath program also says "disjoint variable requirement". I think there was already a discussion about these terms, but I do not know where...

wlammen: I think this is a non-blocking issue

wlammen (approved):

set.mm
avekens: OK, I will create an Github issue for the alignment of the terms disjoint/distinct variables.

avekens (approved):

PR #2908, opened 2022-10-19
Alternative denial

benjub:
Commit messages say it all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ in (approximate) decreasing size, are:
   uses intuitionistic set theory.
   In particular, it does not presume that the law of excluded middle is
   necessarily true in all cases.
-  [[Generated display](http://us.metamath.org/mpeuni/mmil.html)]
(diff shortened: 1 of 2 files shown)


benjub:
Converted to draft because I realized @wlammen has similar theorems in his mathbox. I'll merge them to night.

benjub:
Latest commit: use shortenings from WL's mathbox (and attribute ~nanimn to him). @wlammen: your ~wl-nanbi2 was identical to ~wl-nanbi1, so I removed it.

By the way: ~nanass can probably be shortened.

wlammen:
wl-nanbi2: I guess I wanted to add the new operand to the left, but somehow forgot to finish it.  The proof is easy, use nancom.

wlammen (approved):

benjub:
> wl-nanbi2: I guess I wanted to add the new operand to the left, but somehow forgot to finish it. The proof is easy, use nancom.

Yes, that's the original proof of nanbi2 (by AH and SF) so I left it that way.

PR #2909, opened 2022-10-21
A set is finite if and only if it has decidable equality and is finitely enumerable

jkingdon:
This formalizes a proof from AczelRathjen which they say uses a similar construction to how they prove Corollary 8.1.13 as discussed in https://github.com/metamath/set.mm/issues/2595#issuecomment-1242394670

The forward direction - fidcenumlemim - is mostly already proved.

The reverse direction is fidcenumlemr .  This is adapted from the AczelRathjen proof but because we have a variety of theorems regarding finite sets to build on, we don't need to explicitly construct bijections with natural numbers.

Includes lemmas fidcenumlemrk and fidcenumlemrks also based on AczelRathjen .


iset.mm
```diff
@@ -64682,6 +64682,113 @@ would need to be strengthened (for example, to imply included middle) to
...
+    $}
+  $}
+
+  ${
+    $d A f n x y $.
+    $( A set is finite if and only if it has decidable equality and is finitely
+       enumerable.  Proposition 8.1.11 of [AczelRathjen], p. 72.  (Contributed
+       by Jim Kingdon, 19-Oct-2022.) $)
+    fidcenum $p |- ( A e. Fin <-> ( A. x e. A A. y e. A DECID x = y
+        /\ E. n e. _om E. f f : n -onto-> A ) ) $=
```
benjub, line 64785: So I conclude that in iset.mm conventions, an enumeration allows repetitions ?  Is this standard ?
I haven't checked, but maybe it's safer to mention it at the first occurrence of "enumeration" or "enumerable" in iset.mm
Disregard if it's already the case.

benjub (approved):

iset.mm
jkingdon: > So I conclude that in iset.mm conventions, an enumeration allows repetitions ?  Is this standard ?
> I haven't checked, but maybe it's safer to mention it at the first occurrence of "enumeration" or "enumerable" in iset.mm
> Disregard if it's already the case.

I don't think we have enough theorems of this sort to have a convention really. "Finitely enumerable" is as defined in AczelRathjen, maybe I should cite their definition explicitly because we have been bitten a bit about not noticing two (inequivalent, I think) definitions for "countable" as mentioned at https://github.com/metamath/set.mm/issues/2595#issuecomment-1287083072

If by "enumeration without repetitions" you mean a bijection, the word for having one from a natural number is "finite" (df-fin and df-en which I think matches other sources).

benjub: So your reasoning is that if one reads "finitely enumerable", it has to mean something else than "finite", else we would use that latter expression. Therefore, it can only mean that repetitions are allowed (i.e., surjection from a finite ordinal, instead of the stronger requirement of a bijection). I agree, but this may not be totally obvious on first reading, so if this can be said at the point where it's first introduced in iset.mm, it would be nice.

jkingdon: > maybe I should cite their definition explicitly

I have updated this pull request with that change.

benjub (approved):

PR #2914, opened 2022-10-23
Intuitionize two sections related to sequences and series

jkingdon:
This is the sections "Miscellaneous converging and diverging sequences" and "Arithmetic series".

Some of the theorems intuitionize without much trouble. Others are just noted in the missing theorems list.

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -377,7 +377,10 @@ by Stefan O'Rear, 2014-11-22)</li>
 <!-- 24th added to list -->
 <li><a name="42">42</a>.  Sum of the Reciprocals of the Triangular Numbers (<a
 href="mpeuni/trirecip.html">trirecip</a>, by Scott Fenton,
-2014-05-02)</li>
+2014-05-02).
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -115023,6 +115023,24 @@ Infinite sums (cont.)
       UUASUVTXGXNYTFEVBWEVCCUSUTYFYDKUVJYPUGWNUVRXEXHXFXIXJXK $.
   $}
 
+  ${
+    $d k n $.
+    $( The sum of the reciprocals of the triangle numbers converge to two.
```
benjub, line 115028: The sum ... is

benjub (approved):

iset.mm
jkingdon: To be honest I don't really know how one is supposed to say this in the case of an infinite sum. Maybe "the sum ... is" works for infinite but I'd most naturally reach for that in the finite case.

benjub: There is that, but there is also the grammatical problem, so maybe "the partial sumS ... converge" or at least "the sumS ... converge".

PR #2915, opened 2022-10-23
shorten nanass

wlammen:
6 lines shorter, and mostly long lines killed.


benjub (approved):
Note: ~nanass was recently moved from RP's mathbox in #2908.

PR #2916, opened 2022-10-23
remove some OLD theorems

benjub:
Remove mobiOLD, exsbOLD, eufOLD, mofOLD even if they are less than one-year old, since the proof changes are minor, so no need to clutter the database with the OLD versions.  On the other hand, I kept nexmoOLD since the change is a bit less minor.

In the cases of exsbOLD, eufOLD, mofOLD, the change is actually very minor: it's simply that the DV conditions allow to use cbvexv1 in place of cbvex (both are identical except for DV conditions). By the way, @wlammen (or anyone else): did you check other usages of cbvex to see if they can be replaced with cbvexvw/cbvexv/cbvexv1 (and similarly with cbval and variants) ? It's a bit tedious. Maybe the simplest thing to do is to globally change cbvex to cbvexv1, then `MM> verify proof *` and see where there are DV violations, and revert the changes where there are. We should do this eventually. Maybe I should have done it more systematically than I did when I added cbvalv1 and cbvexv1 but I was a bit lazy. In my opinion, these changes can be done silently (i.e., no new tags, no OLD versions).

By the way: @wlammen: have you had a look at eu6 ? My proof is probably inefficient and shortenable.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15218,7 +15218,6 @@ New usage of "eubidvOLDOLD" is discouraged (1 uses).
 New usage of "eueqOLD" is discouraged (1 uses).
 New usage of "euexALTOLD" is discouraged (0 uses).
 New usage of "euexOLD" is discouraged (0 uses).
-New usage of "eufOLD" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
 New usage of "eumoOLD" is discouraged (0 uses).
 New usage of "ex-gt" is discouraged (0 uses).
@@ -15253,7 +15252,6 @@ New usage of "exmoeuOLD" is discouraged (0 uses).
 New usage of "expcomdg" is discouraged (0 uses).
 New usage of "expdOLD" is discouraged (0 uses).
 New usage of "expdcomOLD" is discouraged (0 uses).
-New usage of "exsbOLD" is discouraged (0 uses).
 New usage of "extwwlkfablem1OLD" is discouraged (0 uses).
 New usage of "f1cofveqaeqALT" is discouraged (0 uses).
 New usage of "f1oweALT" is discouraged (0 uses).
@@ -16370,14 +16368,12 @@ New usage of "mndoismgmOLD" is discouraged (3 uses).
 New usage of "mndoissmgrpOLD" is discouraged (1 uses).
 New usage of "mndomgmid" is discouraged (3 uses).
 New usage of "moabsOLD" is discouraged (0 uses).
-New usage of "mobiOLD" is discouraged (0 uses).
 New usage of "mobidOLD" is discouraged (0 uses).
 New usage of "mobidvALT" is discouraged (0 uses).
 New usage of "mobidvOLD" is discouraged (0 uses).
 New usage of "mobidvOLDOLD" is discouraged (0 uses).
 New usage of "moeqOLD" is discouraged (0 uses).
 New usage of "moeuOLD" is discouraged (0 uses).
-New usage of "mofOLD" is discouraged (0 uses).
 New usage of "mptssALT" is discouraged (0 uses).
 New usage of "mpv" is discouraged (1 uses).
 New usage of "mulassnq" is discouraged (10 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
mobi is a simple shortening. The others are not shortenings, but a revision with the aim to reduce axiom usage.  The result looks obvious, but it is not easily seen (at least for me).  I don't remember exactly, but from the moment of suspicion, there might be headroom of optimization (which takes some experience in this field), down the road to success took me at least an hour in the first case.  Another example: I investigated mof further yesterday, because I thought the -> direction is possibly dependent on even fewer axioms.  This time it didn't pan out, but I was again a minimum of 1 hour busy investigating it.  I am not of the opinion this is just minor work.

The idea to tentatively replace cbvex with cbvexv1, and then let the verifier check the result ... well, go ahead.  In general, you also need to look at the rest of the proof, to see where ax-13 gets dragged in, and whether that part can be adapted as well.  I don't know of a mechanic scheme to deal with that.

Then the question of the value of OLD theorems.  One idea behind keeping them, is (possibly) that a comparison between OLD and NEW version gives an inspiration for similar shortenings, I don't know, maybe learn how to improve should you happen to be the original author.  If the OLDs are seen to be of value at least in some instances, I am opposed to introduce a sort of ranking by which some are sorted out.  This creates an easy source of contention, one example already appearing on the horizon in the first paragraph here.

Finally, I had looked at eu6, but saw no obvious improvement.  So this will take time, which I unfortunately have not in abundance.  I'll look at it closer in near future, promised.

benjub:
> mobi is a simple shortening. The others are not shortenings, but a revision with the aim to reduce axiom usage. The result looks obvious, but it is not easily seen (at least for me). I don't remember exactly, but from the moment of suspicion, there might be headroom of optimization (which takes some experience in this field), down the road to success took me at least an hour in the first case. Another example: I investigated mof further yesterday, because I thought the -> direction is possibly dependent on even fewer axioms. This time it didn't pan out, but I was again a minimum of 1 hour busy investigating it. I am not of the opinion this is just minor work.

You are right, and maybe I was not clear. It's not because the resulting change is minor that the work involved in finding it was minor. Very generally, and outside metamath, there are many proofs that are very simple to understand and look obvious once we know them, but which were hard to find. What I wrote above is about the resulting change, and not the work involved in finding it.

> Then the question of the value of OLD theorems. One idea behind keeping them, is (possibly) that a comparison between OLD and NEW version gives an inspiration for similar shortenings, I don't know, maybe learn how to improve should you happen to be the original author. If the OLDs are seen to be of value at least in some instances, I am opposed to introduce a sort of ranking by which some are sorted out. This creates an easy source of contention, one example already appearing on the horizon in the first paragraph here.

Yes, keeping OLDs should be the rule, and if there is any doubt the OLD proof could be useful, as you say, for comparisons or additional insight, then indeed keep it for one year. Here, I think we are beyond any doubt (especially for exsbOLD, eufOLD, mofOLD). Don't you agree ? (it's a real question: it was not clear to me if your lack of approval means rejection; not a big deal, this PR took me under five minutes and I can close it)

wlammen:
I'd like to comment on the `clutter the database` part, too.  Maybe one can reduce the time-to-live of OLD theorems to say 6 months?  I don't know how often people revisit the Metamath web page.  If most do within half of a year, one can reduce this time, to alleviate the littering effect.

In total I do not approve of this PR, out of the reasons noted above, but perhaps the conversation here is valuable.  I don't like the idea that just because of my opinion, this is dismissed in total.

benjub:
As for the duration, I would prefer to keep one year, and not reduce it to six months. First, it's easier to compute the date, and second, people (at least, I) can happen to take a few months off, and six months could still feel a bit short. In any case, as @wlammen wrote above, the worst solution would be to make different categories, with the "less important ones" staying six months, and some others one year. Some trivial shortenings (or shortenings in mathboxes) can be done silently, but that's it.

As for this PR, that's really not a big deal, and I'll probably close it tomorrow, without any regret.

avekens:
For this PR, I do not have any clear opinion: I am not agaist deleting these OLD theorems.
In general, I think it was discussed already to shorten the time or to categorize OLD theorems (I cannot remember where and when..), without a result. Therefore, I would suggest not to change the procedure, at least not now.

jkingdon:
I guess we should stick with one year. At least, it is simpler than trying to assess when to do something different.

PR #2918, opened 2022-10-25
First six theorems from Geometric series

jkingdon:
There is a fair bit to do here in terms of not equal versus apart and avoiding case elimination.

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -464,7 +464,10 @@ by Saveliy Skresanov, 2017-01-01)</li>
 
 <!-- 10th added to list -->
 <li><a name="66">66</a>.  Sum of a Geometric Series (<a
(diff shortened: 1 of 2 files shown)


avekens (approved):

avekens:
It's great that we have another Metamath 100 theorem in iset.mm now.

jkingdon:
> It's great that we have another Metamath 100 theorem in iset.mm now.

They are coming thick and fast. I sort of knew a lot of them were waiting for summation but I didn't anticipate how many I'd get just by intuitionizing the next few sections after summation without really paying much attention to mm_100.

The streak might be getting ready to slow down, though, as I don't think the ratio test (used to define trigonometry in set.mm) is going to be as easy as the recent stuff.

PR #2920, opened 2022-10-29
Transitivity and equivalence

mazsa:
Mathbox only.

Nothing surprising: fits into the logic of https://github.com/metamath/set.mm/pull/2531 on reflexivity and symmetry.

@digama0 Mario: I wonder whether the way of reference to your name here is OK with you.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -8006,6 +8006,7 @@
 "idn3" is used by "tratrbVD".
 "idn3" is used by "trintALTVD".
 "idn3" is used by "truniALTVD".
+"idrefALT" is used by "dfeqvrels3".
 "idrefALT" is used by "elrefsymrels3".
 "idrefALT" is used by "idinxpssinxp2".
 "idrefALT" is used by "idinxpssinxp3".
@@ -16004,7 +16005,7 @@ New usage of "idlnop" is discouraged (5 uses).
 New usage of "idn1" is discouraged (76 uses).
 New usage of "idn2" is discouraged (39 uses).
 New usage of "idn3" is discouraged (12 uses).
-New usage of "idrefALT" is discouraged (5 uses).
+New usage of "idrefALT" is discouraged (6 uses).
 New usage of "idrefOLD" is discouraged (0 uses).
 New usage of "idrval" is discouraged (2 uses).
 New usage of "idssxpOLD" is discouraged (0 uses).


set.mm
```diff
@@ -545650,6 +545679,28 @@ A collection of theorems for commuting equalities (or
      (Read: ` R ` is a symmetric relation.) $)
   wsymrel $a wff SymRel R $.
 
+  $( Extend the definition of a class to include the transitivity class (but
+     cf. the transitive class defined in ~ df-tr ). $)
+  ctrs $a class Trs $.
+  $( Extend the definition of a class to include the transitive relations
```
benjub, line 545685: blank line (idem below)

set.mm
```diff
@@ -548230,9 +548281,9 @@ to definitions of epsilon relation ( ~ df-eprel ) and identity relation
      converse reflexivity ( ~ df-cnvrefs ), cf. the comment of ~ df-ssr .  The
      elements of this class are not necessarily relations (vs. ~ df-refrels ).
 
-     Note the similarity of the definitions ~ df-refs , ~ df-syms and ~~?
-     df-trs , cf. the comments of ~ dfrefrels2 .  (Contributed by Peter Mazsa,
-     19-Jul-2019.) $)
+     Note the similarity of the definitions ~ df-refs , ~ df-syms and
+     ~ df-trs , cf. the comments of ~ dfrefrels2 .  (Contributed by Peter
```
benjub, line 548285: cf. comments (idem below)

set.mm
```diff
@@ -548850,6 +548900,515 @@ class of symmetric relations as well (like the elements of the class of
     ( crefrels csymrels cin wcel wrefrel wsymrel elin elrefrelsrel elsymrelsrel
     wa anbi12d syl5bb ) ACDEFACFZADFZLABFZAGZAHZLACDIQORPSABJABKMN $.
 
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  Transitivity
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  $( Define the class of all transitive sets (vs. the transitive class defined
```
benjub, line 548909: versus

set.mm
```diff
@@ -548850,6 +548900,515 @@ class of symmetric relations as well (like the elements of the class of
...
+
+     Note the similarity of the definitions of ~ df-refs , ~ df-syms and
+     ~ df-trs .  (Contributed by Peter Mazsa, 17-Jul-2021.) $)
+  df-trs $a |- Trs = { x |
+      ( ( x i^i ( dom x X. ran x ) ) o. ( x i^i ( dom x X. ran x ) ) ) _S
+        ( x i^i ( dom x X. ran x ) ) } $.
+
+  $( Define the class of transitive relations.  The element of the class of
+     transitive relations and the transitive relation predicate are the same
+     when ` R ` is a set, cf. ~ eltrrelsrel .  Alternate definitions are
```
benjub, line 548920: For sets, being an element of the class of.... is equivalent to satisfying the ... predicate.
(idem below)

digama0:
> @digama0 Mario: I wonder whether the way of reference to your name here is OK with you.

Assuming those theorems are variations/modifications on an existing theorem, sure you can use the original contributor if you like, although I certainly wouldn't have noticed if you claimed them for yourself!

set.mm
```diff
@@ -548850,6 +548900,515 @@ class of symmetric relations as well (like the elements of the class of
...
+      ( ( ( R i^i ( dom R X. ran R ) ) o. ( R i^i ( dom R X. ran R ) ) ) C_
+          ( R i^i ( dom R X. ran R ) ) /\ Rel R ) ) $.
+
+  $( Alternate definition of the class of transitive relations.
+
+     I'd prefer to define the class of transitive relations by using the
+     definition of composition by [Suppes] p. 63. df-coSUP
+     ` ( A o. B ) = { <. x , y >. | E. u ( x A u /\ u B y ) } ` as opposed to
+     the present definition of composition ~ df-co
+     ` ( A o. B ) = { <. x , y >. | E. u ( x B u /\ u A y ) } ` because the
```
benjub, line 548943: Eternal debate !

set.mm
```diff
@@ -548850,6 +548900,515 @@ class of symmetric relations as well (like the elements of the class of
...
+     Suppes definition keeps the order of ` A ` , ` B ` , ` C ` , ` R ` ,
+     ` S ` , ` T ` by default in trsinxpSUP ( ( ( R i^i ( A X. B ) ) o.  ( S
+     i^i ( B X. C ) ) ) C_ ( T i^i ( A X. C ) ) <-> A. x e.  A A. y e.  B A. z
+     e.  C ( ( x R y /\ y S z ) -> x T z ) ) while the present definition of
+     composition disarranges them: trsinxp ( ( ( S i^i ( B X. C ) ) o.  ( R i^i
+     ( A X. B ) ) ) C_ ( T i^i ( A X. C ) ) <-> A. x e.  A A. y e.  B A. z e.
+     C ( ( x R y /\ y S z ) -> x T z ) ) This is not mission critical to me,
```
benjub, line 548950: Period before "This". Maybe you want to add some backticks to have the comment formatted.

set.mm
```diff
@@ -548850,6 +548900,515 @@ class of symmetric relations as well (like the elements of the class of
...
+     the implication of the Suppes definition is just more aesthetic, at least
+     in the above case.
+
+     If we swap to the Suppes definition of class composition, I would define
+     the present class of all transitive sets as df-trsSUP and I would consider
+     to switch the definition of the class of cosets by ` R ` from the present
+     ~ df-coss to a df-cossSUP. But perhaps there is a mathematical reason to
+     keep the present definition of composition.  (Contributed by Peter Mazsa,
```
benjub, line 548958: One has (f o g)(x) = f(g(x)). If you swap composition, then to be fully consistent you should also write function application in a postfix way. Some authors do that, e.g., Jacobson, Lectures in Abstract Algebra.
Changing the definition in set.mm would be too much work for unclear benefit.

set.mm
```diff
@@ -548850,6 +548900,515 @@ class of symmetric relations as well (like the elements of the class of
...
+    ( wcel ccom wss crels wa ctrrels wtrrel elrelsrel anbi2d eltrrels2 dftrrel2
+    wrel 3bitr4g ) ABCZAADAEZAFCZGQANZGAHCAIPRSQABJKALAMO $.
+
+  $( Equality theorem for transitive relation.  (Contributed by Peter Mazsa,
+     15-Apr-2019.)  (Revised by Peter Mazsa, 23-Sep-2021.) $)
+  trreleq $p |- ( R = S -> ( TrRel R <-> TrRel S ) ) $=
+    ( wceq ccom wrel wa wtrrel coideq id sseq12d releq anbi12d dftrrel2 3bitr4g
+    wss ) ABCZAADZAOZAEZFBBDZBOZBEZFAGBGPRUASUBPQTABABHPIJABKLAMBMN $.
+
+$(
```
benjub, line 549028: double blank lines between sections

set.mm
```diff
@@ -548850,6 +548900,515 @@ class of symmetric relations as well (like the elements of the class of
...
+  $}
+
+  $( The element of the class of transitive relations and the transitive
+     relation predicate are the same when ` R ` is a set.  (Contributed by
+     Peter Mazsa, 22-Aug-2021.) $)
+  eltrrelsrel $p |- ( R e. V -> ( R e. TrRels <-> TrRel R ) ) $=
+    ( wcel ccom wss crels wa ctrrels wtrrel elrelsrel anbi2d eltrrels2 dftrrel2
+    wrel 3bitr4g ) ABCZAADAEZAFCZGQANZGAHCAIPRSQABJKALAMO $.
+
+  $( Equality theorem for transitive relation.  (Contributed by Peter Mazsa,
```
benjub, line 549022: ...for the transitive relation predicate.

benjub:
I just made a few comments above. I did not review carefully since this is mathbox-only (tag added). It is mergeable to me. Maybe we can wait a bit to get comments from other contributors ?

As a friendly general advice, from someone who fell prey to similar dispositions (and who still struggles sometimes against them): we arrive after a few thousand years of mathematics, and sometimes, not all definitions are precisely suited to our tastes. But it would be a waste of time that everyone redefines everything with small variations compared to the commonly used definitions. Mathematics is a common endeavour, so I would encourage you to use existing definitions and prove results using them, rather than redefining things. An occasional redefinition might be necessary, or course, and even beneficial, but this is more the exception than the rule (and knowing when it is beneficial is a difficult problem). As a rule of thumb, one should aim for a high ratio (interesting theorems)/(definitions) in their work. (Actually, I'm writing this advice for myself as well...)

set.mm
avekens: I agree with @benjub not to change the current definition of composition.

avekens (approved):
OK since everything is in a mathbox.

set.mm
mazsa: done

set.mm
mazsa: done

set.mm
mazsa: done

set.mm
mazsa: done

set.mm
mazsa: Period: done. Backticks: they would ruin format with long rows.

set.mm
mazsa: done

set.mm
mazsa: done

mazsa:
@digama Thank you. Then I'll leave it this way, you as the original Contributor.

@benjub @avekens You are completely right. I try to pick my fights carefully: this is why I did not try to publish my experiments with df-coSUP, not even in my Mathbox, I only mention this possibility for further reference (or should I delete it?). My rule of thumb is the following: if even I don't think I see the whole picture (my default status in math:) then I avoid making suggestions.

I have only 2 Metamath-projects related to conventions, and one of them is Mathbox-only:
1. I'll continue to put into my Mathbox my results which I (almost completely) published on the Metamath mail-list a couple of months ago (except my next PR:). The community will either accept them as a whole in the long term and upgrade them all to the main set.mm, or they won't stand the test of time and will fade into oblivion. (Of course I secretly hope that my results, due to the superpower of Metamath, will become a part of the millennia-long history of mathematics, although in a less important corner of it for sure:)
2. My project concerning the main set.mm is only about notation: I want to implement the much more transparent notation for _I _E and _S along the line of [C.] of @sorear Stefan O'Rear (see ~ df-rpss ) which is suggested by Benoit in https://github.com/metamath/set.mm/pull/2526#issuecomment-1050649051 . Alexander: do you agree it would be a good idea?

avekens:
>     2. My project concerning the main set.mm is only about notation: I want to implement the much more transparent notation for _I _E and _S along the line of [C.] of Stefan O'Rear (see ~ df-rpss ) which is suggested by Benoit in [Relations and Subsets #2526 (comment)](https://github.com/metamath/set.mm/pull/2526#issuecomment-1050649051) . Alexander: do you agree it would be a good idea?

I think there is no urgent need to change the symbols, but I am also not against it. So if nobody objects, the symbols can be changed as proposed by @benjub in #2526. 
By the way, `_I` is used about 1000 times in set.mm, `_E` about 200 times.

benjub:
> I only mention this possibility for further reference (or should I delete it?)

No, it's fine as a comment.

> (except my next PR:).

Remember that multiple shorter PRs are appreciated.

> I want to implement the much more transparent notation for _I _E and _S along the line of [C.] of Stefan O'Rear (see ~ df-rpss ) which is suggested by Benoit in https://github.com/metamath/set.mm/pull/2526#issuecomment-1050649051

To clarify what I meant: yes to change both tokens and symbols as follows (each in a separate PR, and there is no urgency as @avekens wrote):
```
token | symbol
[e.]  | [∈]
[C_]  | [⊆]
[C.]  | [⊊] (https://us.metamath.org/mpeuni/df-rpss.html)
```
As for the identity relation (currently `_I` rendered as upright capital I), then it's true that when regarded as a relation, the clearest token/symbol would be `[=]`, but when viewed as a function, it would be `Id` rendered as upright Id.  So I'm not sure which to choose yet.

benjub (approved):

avekens:
> As for the identity relation (currently `_I` rendered as upright capital I), then it's true that when regarded as a relation, the clearest token/symbol would be `[=]`, but when viewed as a function, it would be `Id` rendered as upright Id. So I'm not sure which to choose yet.

Yes, that could be a problem. There are currently 54 occurences of ``( _I ` * )`` in set.mm, and the theorems would actually look strange if the propsed symbol is used, e.g. ``ids1 $p |- <" A "> = <" ( [=] ` A ) "> $=``. ``Id`` would be a better choice here.

PR #2921, opened 2022-10-29
Misc

benjub:
For the record: a field is "quadratically closed" if it is closed under square roots. Some uses in set.mm were incorrect. An ordered field closed under square roots of positive numbers is sometimes called "Euclidean", but here in our cases, it is sufficient to simply write out the explicit property, without using a not-so-common terminology and burdening the reader's mind.


set.mm
```diff
@@ -261589,21 +261593,14 @@ According to Wikipedia ("Linear algebra", 03-Mar-2019,
...
-  matrices, determinants, vector spaces, and linear transformations."  However,
-  dealing with modules (over rings) instead of vector spaces (over fields)
-  allows for a more general approach. Therefore, "vectors" are regarded as
-  members (elements of the base set) of a (free) module over a ring (see
-  ~ df-frlm ) in the following.  By this, the number of entries in a vector is
-  determined by the size of the index set of the direct sum building the free
-  module the vector is belonging to. Since every vector space is isomorphic to
-  a free module (see ~ lvecisfrlm ), the theorems stated for free modules are
-  also valid for vector spaces.
-
```
avekens, line 261601: Why did you remove
```
Therefore, "vectors" are regarded as members (elements of the base set) of a (free) module over a ring
(see ~ df-frlm ) in the following.  By this, the number of entries in a vector is determined by the size
of the index set of the direct sum building the free module the vector is belonging to. 
Since every vector space is isomorphic to a free module (see ~ lvecisfrlm ), the theorems stated for
free modules are also valid for vector spaces.
```
I think this are valuable information to understand the following theorems for somebody who knows about vector spaces, but is not familiar with (left) modules.

set.mm
```diff
@@ -261972,23 +261969,22 @@ consisting of the same ring regarded as a (left) module over itself, see
...
-       ` i ` -many copies of the ring with componentwise addition and
-       multiplication.  If ` i ` is infinite, the allowed vectors are
-       restricted to those with finitely many nonzero coordinates; this ensures
-       that the resulting module is actually spanned by its unit vectors.
-       (Contributed by Stefan O'Rear, 1-Feb-2015.) $)
+    $( Define the function associating with a ring and a set the direct sum
+       indexed by that set of copies of that ring regarded as a left module
+       over itself.  Recall from ~ df-dsmm that an element of a direct sum has
+       finitely many nonzero coordinates.  (Contributed by Stefan O'Rear,
+       1-Feb-2015.) $)
```
avekens, line 261976: From my point of view, to much information is removed by the new comment. The old and new version of the comment should be merged somehow.

avekens (changes requested):
Please have a look at my review remarks.

avekens:
> For the record: a field is "quadratically closed" if it is closed under square roots. Some uses in set.mm were incorrect. An ordered field closed under square roots of positive numbers is sometimes called "Euclidean", but here in our cases, it is sufficient to simply write out the explicit property, without using a not-so-common terminology and burdening the reader's mind.

So you replaced ``quadratically closed subfield of ` CCfld ` `` by `` subfield of ` CCfld ` closed under square roots of nonnegative reals``.  This seems to be reflected in the definition df-cph of `CPreHil` by intersecting `k` with the nonnegatrove reals:

```
... ( sqrt " ( k i^i ( 0 [,) +oo ) ) ) C_ k ...
```

Therefore, the suggested changes seems to be adequate.

set.mm
```diff
@@ -246825,19 +246828,20 @@ U C_ ( N ` { X } ) ) -> ( U = ( N ` { X } ) \/ U = { .0. } ) ) $=
...
   ${
     $d w s $.
-    $( Given any subring of a ring, we can construct a left-algebra by
-       regarding the elements of the subring as scalars and the ring itself as
-       a set of vectors.  (Contributed by Mario Carneiro, 27-Nov-2014.)
-       (Revised by Thierry Arnoux, 16-Jun-2019.) $)
+    $( Given any subring of a ring, we can consider the full ring as a left
+       algebra over the subring (which forms the ring of scalars).  It has an
+       inner product which is simply the ring product.  (Contributed by Mario
+       Carneiro, 27-Nov-2014.)  (Revised by Thierry Arnoux, 16-Jun-2019.) $)
```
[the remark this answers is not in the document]
benjub: Revised accordingly, as well as the following definition: a first introductory sentence giving the gist of it, and then a precise description.

set.mm
benjub: I added a paragraph for these readers. What I removed gives a misleading intuition. First, it focuses on free modules, and then, it gives the idea that there is always a notion of "dimension" for modules, which is not the case.

set.mm
benjub: What was remove gives a wrong intuition. It talks about "the i-dimensional", but if $i$ is replaced by $\pi$, then one obtains the direct sum of uncountably many copies of the ring (simply because in set.mm, $\pi$ is encoded as some uncountable set). Then, there is no notion of dimension for general modules. I kept the last sentence recalling that elements of a direct sum have finitely many nonzero summands.

benjub:
By the way, @avekens : I changed both the token and the symbol toCHil-->toCPreHil. Do you think we should change in labels tch-->tcph ? (there are around twenty impacted labels)

avekens (approved):

avekens:
> By the way, @avekens : I changed both the token and the symbol toCHil-->toCPreHil. Do you think we should change in labels tch-->tcph ? (there are around twenty impacted labels)

Yes, it would be consistent with our convention to change the label, and I would endorse that. If you can do it with this PR - fine. Otherwise, it can be done later.

avekens (approved):
great, thanks!

Issue #2905, opened 2022-10-17
Terms "disjoint/distinct variables"


avekens:
Currently, two terms are used for the same concept: "disjoint/distinct variables"

On the Metamath Proof Explorer Home Page (see https://us.metamath.org/mpeuni/mmset.html#distinct), "distinct" is used ("distinct variables", "distinct variable condition"), but we have (at least) 11 "disjoint variable condition" in set.mm (and 106 "distinct variable condition"). There is, however, also the following hint: 

> ... This is why a distinct variable pair is officially called a "disjoint variable pair" in the Metamath specification.

@wlammen Proposed to use "disjoint", because of `$d x ph $.`. The error message of the metamath program also says "disjoint variable requirement".... 

I think there was already a discussion about these terms, but I do not know where and if there was an agreement which term to be used.

Finally, the term(s) "disjoint/distinct variables" are abbreviated as "DV" and "dv" in set.mm:
* "dv condition" 174 times,
* "DV condition" 30 times.
I would propose to unify it to "DV condition". This is also according to our conventions: "No two variables occurring in the LHS may share a disjoint variable (DV) condition."

(See also discussion in #2904).

benjub:
I agree with @avekens and @wlammen to capitalize "DV condition" since it's an acronym. The "D" is for "disjoint"; I think "distinct" can be misleading (even if @digama0 uses it in https://arxiv.org/abs/1601.07699 and https://us.metamath.org/downloads/metamath.pdf uses both rather interchangeably). We can also add a paragraph in ~conventions, and link to it from ~ ax-5, which is the first statement with a DV condition.
Of course I'm a bit partial since I wrote that latest example that @avekens mentioned above, and I used "DV condition", meaning "disjoint", systematically in https://arxiv.org/abs/2202.10383

benjub:
Just as a heads up to @avekens and @wlammen to avoid duplicate work: I will do that.

benjub:
I did this in PR #2923, as promised. @jkingdon proposes there to spell out "DV".  I can do that, through a simple search and replace. I thought "DV" was ubiquitous enough to deserve an abbreviation, but indeed it might be a bit obscure to someone landing on a given page.  So no strong opinion.  What do you think, @avekens @wlammen ?

avekens:
[...]

PR #2923, opened 2022-10-31
Consistent naming "DV condition"

benjub:
This follows #2905 (with a few unrelated minor fixes). This is an improvement towards consistency, but not the final step: in a coming PR, there needs to be some better explanation of "DV condition" both where it's first introduced (ax-5) and in ~conventions, with links to the home page where it is explained in more detail.


jkingdon:
I'd advocate spelling DV out. Are you just worried about whether to call it "distinct variable" or "disjoint variable"?

set.mm
```diff
@@ -727059,7 +727059,7 @@ last digit of the integer part (for ` b = ` 10 the first digit before
        different problems can be shown to be equivalent.  Thus, Conway
        preferred defining surreal numbers as equivalence classes of surreal
        number forms, rather than sign-expansions.  Although sign-expansions are
-       easier to implement in ZF set theory, Conway argued that "formalisation
+       easier to implement in ZF set theory, Conway argued that "formalization
```
avekens, line 727062: Is this a citation? Then it should not be modified.

avekens (approved):
I would vote for using the abbreviation DV (and replacing the abbreviation dv by it), but there are different opinions, see #2905.

set.mm
benjub: He uses "s" so I reverted to that. I also wrote the reference in full (full title of the book, and John "Horton" Conway, so as not to confuse him with the functional analyst).

avekens (approved):
So in principle, the main change in this PR is that DV/dv is spelled out as "disjoint variable". It's OK for the moment, but I want to wait for the dollow up PR if this is the final status of this topic ("disjoint variable" could be easily replaced by "distinct variable" or "DV" if this was necessary).

avekens:
I would like that @jkingdon and @wlammen will approve this PR, too.

jkingdon:
> I would like that @jkingdon and @wlammen will approve this PR, too.

Not sure I'm really qualified. I don't understand wording like "additional disjoint variable condition on ` y , z ` " (I'm not sure I even have the right terminology to formulate why this seems confusing to me, but what is disjoint from what? and does disjointness as a concept make sense for setvars?)

Maybe we should try to write the style guide first and then apply it? Or accept a bit more anarchy?

wlammen:
For me x and y are distinct variables, too.  I think in situations like these, replacing disjoint with distinct should be an option, but in general the variable checking procedure, without specifying setvars, should be a 'disjoint variable condition'.
For now, I am satisfied.  The PR is an improvement in my eyes, and fine-tuning can be done in another one.

wlammen (approved):

digama0:
My recommendation is to use "distinct variables" when referring specifically to setvars `x, y` and "disjoint variables" in all other cases (either `x, ph` or when speaking in general).

avekens:
> My recommendation is to use "distinct variables" when referring specifically to setvars `x, y` and "disjoint variables" in all other cases (either `x, ph` or when speaking in general).

That's what I pointed to in #2905. Maybe this can be subject of a fine tuning in the follow up PR.

benjub:
> My recommendation is to use "distinct variables" when referring specifically to setvars `x, y` and "disjoint variables" in all other cases (either `x, ph` or when speaking in general).

@digama0 : can I still merge the present PR, and maybe deal with your proposal later ? n There are a few other modifications in this PR and I fear merge conflicts because of its global nature; moreover, the change you propose would be more complex than a simple search-and-replace.

And regarding distinct/disjoint: I think that the same term (hence "disjoint") should be used in all cases to emphasize unity of the concept. I also think that using "distinct" could be misleading, and give the reader a false sense of understanding. Specifically, what is the best term for the following definition:
> We say that the setvars `x, y` are **xxxxx variables** if they can only be substituted by distinct setvars.

Using the same word "distinct" can be misleading, which is why I think "disjoint" is better.

digama0:
> > My recommendation is to use "distinct variables" when referring specifically to setvars `x, y` and "disjoint variables" in all other cases (either `x, ph` or when speaking in general).
> 
> @digama0 : can I still merge the present PR, and maybe deal with your proposal later ? n There are a few other modifications in this PR and I fear merge conflicts because of its global nature; moreover, the change you propose would be more complex than a simple search-and-replace.

I'm not too fussed about the timing of things. Do whatever is easiest for you.

> And regarding distinct/disjoint: I think that the same term (hence "disjoint") should be used in all cases to emphasize unity of the concept. I also think that using "distinct" could be misleading, and give the reader a false sense of understanding. Specifically, what is the best term for the following definition:
> 
> > We say that the setvars `x, y` are **xxxxx variables** if they can only be substituted by distinct setvars.
> 
> Using the same word "distinct" can be misleading, which is why I think "disjoint" is better.

I disagree that using "disjoint" uniformly is less confusing, because variables are not sets and calling them disjoint is confusing. I would rephrase the example to:

> When setvars `x, y` have a disjoint variable condition (a `$d` directive), we say they are *distinct* because they can only be substituted by distinct object-level variables. (The terms "disjoint variable" and "distinct variable" are used interchangeably on setvars.)

This makes it clearer that this is not so much a definition of what "disjoint variable" means for setvars as a specialization of the pre-existing notion to setvars.

benjub:
> I'm not too fussed about the timing of things. Do whatever is easiest for you.

Thanks. I will therefore merge this one now.

> I disagree that using "disjoint" uniformly is less confusing, because variables are not sets and calling them disjoint is confusing. I would rephrase the example to:
> 
> > When setvars `x, y` have a disjoint variable condition (a `$d` directive), we say they are _distinct_ because they can only be substituted by distinct object-level variables. (The terms "disjoint variable" and "distinct variable" are used interchangeably on setvars.)
> 
> This makes it clearer that this is not so much a definition of what "disjoint variable" means for setvars as a specialization of the pre-existing notion to setvars.

Then we disagree !  On the other hand, your proposed explanation above is clear enough, given the space constraint, and you probably had to explain this concept to more outsiders at talks or conferences, so you have a better idea of what conveys the message best in one sentence.  I hadn't thought of the confusion that may arise in readers' mind, who could think that "disjoint" is interpreted in the object language, i.e., `( x i^i y ) = (/)`, which of course has nothing to do with DV(x,y).

PR #2924, opened 2022-11-01
Redundancy

mazsa:
Completely experimental. I truly need your advice here: I'm sure I'm onto something but not sure this PR has a place even in my Mathbox *in its present form*. I'll mention everybody who had an opinion on the reflexivity topic: @david-a-wheeler , @benjub , @tirix ( in https://github.com/metamath/set.mm/pull/1286 ) + @jkingdon , @avekens (in https://github.com/metamath/set.mm/pull/2531 ).

The problem is the following: based on
```metamath
refrelred3 $p |- red ( ( A. x e. dom R x R x /\ Rel R ) , RefRel R , EqvRel R )
```
, the definiens of the "naive" version of the definition of reflexive relation, ` ( A. x e. dom R x R x /\ Rel R ) ` , is redundant with respect to the "real" or "general" reflexive relation (see ~ dfrefrel3 ) in equivalence relation. So far so good: *in case of definitions*,
```metamath
df-redp $a |- ( red ( ph , ps , ch ) <-> ( ( ph -> ps ) /\ ( ( ph /\ ch ) <-> ( ps /\ ch ) ) ) )
```
works as a redundancy operation. The question is whether in general, i.e. including cases which do not refer to definitions like ` RefRel R ` or ` EqvRel R ` above, can I name this ~ df-redp operator as "redundancy" operator for propositions? Or is this a more general operator which in case of definitions works as a redundancy operator? If the latter, then what would be its appropriate name?

Thank you for your help.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -8010,6 +8010,8 @@
 "idrefALT" is used by "elrefsymrels3".
 "idrefALT" is used by "idinxpssinxp2".
 "idrefALT" is used by "idinxpssinxp3".
+"idrefALT" is used by "refrelred3".
+"idrefALT" is used by "refrelsred3".
 "idrefALT" is used by "refsymrels3".
 "idrefALT" is used by "symrefref3".
 "idrval" is used by "cmpidelt".
@@ -16005,7 +16007,7 @@ New usage of "idlnop" is discouraged (5 uses).
 New usage of "idn1" is discouraged (76 uses).
 New usage of "idn2" is discouraged (39 uses).
 New usage of "idn3" is discouraged (12 uses).
-New usage of "idrefALT" is discouraged (6 uses).
+New usage of "idrefALT" is discouraged (8 uses).
 New usage of "idrefOLD" is discouraged (0 uses).
 New usage of "idrval" is discouraged (2 uses).
 New usage of "idssxpOLD" is discouraged (0 uses).


benjub:
If it is experimental, then it should be mathbox-only.  Once you move these two theorems in Main to your mathbox, I don't have any reason to reject the PR. That being said, I do not think these new definitions have an interest.

mazsa:
@benjub corrected: now Mathbox-only.

benjub (approved):

set.mm
```diff
@@ -544674,14 +544683,6 @@ the next (since the empty set has a finite subcover, the
...
   $}
 
-  ${
-    selconj.1 $e |- ( ph <-> ( ps /\ ch ) ) $.
-    $( An inference for selecting one of a list of conjuncts.  (Contributed by
-       Giovanni Mascellani, 23-May-2019.) $)
-    selconj $p |- ( ( et /\ ph ) <-> ( ps /\ ( et /\ ch ) ) ) $=
-      ( wa anbi2i an12 bitr4i ) DAFDBCFZFBDCFFAJDEGBDCHI $.
-  $}
-
```
avekens, line 544684: A theorem must not be (re)moved from a mathbox to another mathbox, at least not without permission of the owner. The genereal rule is: if a theorem in a mathbox is used within another mathbox, the theorem should be moved to main! See item 5. in "Mathbox guidelines".

avekens (changes requested):

set.mm
benjub: After looking at its statement, I would actually "inline it" in the proofs using it, since it is too specific.  There is the command `MM-PA> EXPAND selconj` (followed by `MM-PA> MINIMIZE_WITH *`) for that.

mazsa: @avekens of course I mentioned @giomasce in my commit. I chose this temporary option because I did not want to put it directly into Main: @benjub is right, this is a too specific theorem. Now I put it back to its original place and solved the proof without it.

avekens (approved):

benjub (approved):

PR #2927, opened 2022-11-05
In iset.mm, add seq3f1o and related theorems (converting df-iseq syntax to df-seq3 syntax)

jkingdon:
As can be seen by the remaining entries in `iset-discouraged` it will take a while to convert everything over. But the good news is that it is all going smoothly - https://us.metamath.org/ileuni/seqf.html , https://us.metamath.org/ileuni/seq3-1.html , and https://us.metamath.org/ileuni/seq3p1.html provide the basic theorems needed and I'm making progress on converting over everything which depends on them.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6351,7 +6351,7 @@ or ~ ssfiexmid </TD>
 
 <TR>
 <TD>seqfeq2</TD>
-<TD>~ iseqfeq2</TD>
+<TD>~ seq3feq2</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6396,9 +6396,9 @@ or ~ ssfiexmid </TD>
 
 <TR>
   <TD>seq1p</TD>
-  <TD>~ iseq1p</TD>
-  <TD>iseq1p requires that ` F ` be defined on ` ( ZZ>= `` M ) `
-  not merely ` ( M ... N ) ` as in seq1p . This is not a problem
+  <TD>~ seq3-1p</TD>
+  <TD>Requires that ` F ` be defined on ` ( ZZ>= `` M ) `
+  not merely ` ( M ... N ) ` . This is not a problem
   when used on infinite sequences, but perhaps this requirement
   could be relaxed if there is a need.</TD>
 </TR>
@@ -6432,9 +6432,9 @@ or ~ ssfiexmid </TD>
 
 <TR>
   <TD>seqf1o</TD>
-  <TD>~ iseqf1o</TD>
+  <TD>~ seq3f1o</TD>
   <TD>The functions ` G ` and ` H ` need to be defined on ` ( ZZ>= `` M ) `
-  not merely ` ( M ... N ) ` as in seqf1o . Also, a single set ` S `
+  not merely ` ( M ... N ) ` . Also, a single set ` S `
   takes the place of ` C ` and ` S ` because that is sufficient
   flexibility at least for now.</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6356,7 +6356,7 @@ or ~ ssfiexmid </TD>
 
 <TR>
 <TD>seqfveq</TD>
-<TD>~ iseqfveq</TD>
+<TD>~ seq3fveq</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6346,7 +6346,7 @@ or ~ ssfiexmid </TD>
 
 <TR>
 <TD>seqfveq2</TD>
-<TD>~ iseqfveq2</TD>
+<TD>~ seq3fveq2</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6398,7 +6398,7 @@ or ~ ssfiexmid </TD>
   <TD>seq1p</TD>
   <TD>~ seq3-1p</TD>
   <TD>Requires that ` F ` be defined on ` ( ZZ>= `` M ) `
-  not merely ` ( M ... N ) ` . This is not a problem
+  not merely ` ( M ... N ) ` .  This is not a problem
   when used on infinite sequences, but perhaps this requirement
   could be relaxed if there is a need.</TD>
 </TR>
@@ -6407,7 +6407,7 @@ or ~ ssfiexmid </TD>
   <TD>seqcaopr3</TD>
   <TD>~ iseqcaopr3</TD>
   <TD>iseqcaopr3 requires that ` F ` , ` G ` , and ` H ` be defined on
-  ` ( ZZ>= `` M ) ` not merely ` ( M ... N ) ` as in seqcaopr3 . This
+  ` ( ZZ>= `` M ) ` not merely ` ( M ... N ) ` as in seqcaopr3 .  This
   is not a problem when used on infinite sequences, but perhaps this
   requirement could be relaxed if there is a need.</TD>
 </TR>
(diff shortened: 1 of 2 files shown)


wlammen:
I feel a bit at a loss what to do here.  I am not an expert of iset.mm, so I feel a bit reluctant to review here.  Any suggestions how to proceed?

jkingdon:
> I feel a bit at a loss what to do here. I am not an expert of iset.mm, so I feel a bit reluctant to review here. Any suggestions how to proceed?

Thanks for taking a look!

I can start by explaining in more detail.

The syntax of `seq` in set.mm is `seq𝑀( + , 𝐹)` ( https://us.metamath.org/mpeuni/df-seq.html ) but many seq related theorems won't work as-is without excluded middle. Because we are talking about a large set of connected theorems I don't know whether there is one place to point to but a lot of it comes down to whether `seq𝑀( + , 𝐹)` is a function and how we can proceed in the absence of things like https://us.metamath.org/mpeuni/fvex.html .

When we first added `seq` to iset.mm, the way to make it work with the recursion theorems we had at the time seemed to be to add an additional argument for the set which the terms of the series in `seq` are contained in (very often `CC` when this is applied). So that is https://us.metamath.org/ileuni/df-iseq.html and the syntax is `seq𝑀( + , 𝐹, 𝑆)`.

Fast forward a few years and there has been a bunch of work including https://us.metamath.org/ileuni/tfrcl.html which enables being able to revert the syntax to the set.mm syntax - https://us.metamath.org/ileuni/df-seq3.html - and prove either the set.mm theorem exactly, as in https://us.metamath.org/ileuni/seqf.html , or a version with more hypotheses, as in https://us.metamath.org/ileuni/seq3-1.html (which can be contrasted with https://us.metamath.org/mpeuni/seq1.html or https://us.metamath.org/ileuni/iseq1.html ).

Sorry if that is too verbose but I guess the quick summary is that this pull request is just like https://github.com/metamath/set.mm/pull/2902 and other recent pull requests which switch from the `seq𝑀( + , 𝐹, 𝑆)` syntax to the `seq𝑀( + , 𝐹)` syntax. There isn't any significant remaining obstacle to converting everything, it will just require a lot of pull requests similar to this one.

benjub:
Side remark, looking at pages like https://us.metamath.org/ileuni/mmtheorems99.html#iseqex : for the sake of concision, I would generally use fewer line breaks, and put the "Use xxx instead." right before the discouragement tag.  E.g.,
```
Existence of the sequence builder operation.  (Contributed by Jim Kingdon,
20-Aug-2021.)  Use seqex instead (together with iseqsst or iseqseq3 if need
be).  (New usage is discouraged.)
```
instead of
```
Existence of the sequence builder operation.

New proofs should use seqex (together with iseqsst or iseqseq3 if need be).

(Contributed by Jim Kingdon, 20-Aug-2021.)  (New usage is discouraged.)
```

iset.mm
```diff
@@ -100691,15 +100691,15 @@ seq K ( .+ , G ) ) $=
...
+    seq3-1p.4 $e |- ( ph -> M e. ZZ ) $.
+    seq3-1p.5 $e |- ( ( ph /\ x e. ( ZZ>= ` M ) ) -> ( F ` x ) e. S ) $.
     $( Removing the first term from a sequence.  (Contributed by Jim Kingdon,
        16-Aug-2021.) $)
-    iseq1p $p |- ( ph -> ( seq M ( .+ , F , S ) ` N ) =
-        ( ( F ` M ) .+ ( seq ( M + 1 ) ( .+ , F , S ) ` N ) ) ) $=
-      ( cseq4 cfv c1 caddc co wcel cz cuz uzid syl iseqsplit iseq1 oveq1d eqtrd
-      ) AIEFGHOZPHUIPZIEFGHQRSOPZESHGPZUKESABCDEFGHHIJKLAHUATHHUBPTMHUCUDNUEAUJ
-      ULUKEABCEFGHMNJUFUGUH $.
+    seq3-1p $p |- ( ph -> ( seq M ( .+ , F ) ` N ) =
```
benjub, line 100698: Is that a temporary label ? Not sure about our conventions regarding dashes in labels, except `ax-` and `df-`. As for me, I would prefer to avoid them.

mmil.raw.html
```diff
@@ -6396,9 +6396,9 @@
 
 <TR>
   <TD>seq1p</TD>
-  <TD>~ iseq1p</TD>
-  <TD>iseq1p requires that ` F ` be defined on ` ( ZZ>= `` M ) `
-  not merely ` ( M ... N ) ` as in seq1p . This is not a problem
+  <TD>~ seq3-1p</TD>
+  <TD>Requires that ` F ` be defined on ` ( ZZ>= `` M ) `
+  not merely ` ( M ... N ) ` . This is not a problem
```
benjub, line 6401: double space

mmil.raw.html
```diff
@@ -6432,9 +6432,9 @@
 
 <TR>
   <TD>seqf1o</TD>
-  <TD>~ iseqf1o</TD>
+  <TD>~ seq3f1o</TD>
   <TD>The functions ` G ` and ` H ` need to be defined on ` ( ZZ>= `` M ) `
-  not merely ` ( M ... N ) ` as in seqf1o . Also, a single set ` S `
+  not merely ` ( M ... N ) ` . Also, a single set ` S `
```
benjub, line 6437: double space

benjub:
Reviewed the whole thing. Some non-blocking comments.  Removing that extraneous argument of seq is indeed for the better.

iset.mm
jkingdon: There's a decent amount of precedent for hyphens especially in many cases between two digits. This is partly written down at https://us.metamath.org/mpeuni/conventions-labels.html but I'm basing it at least as much on what I see in set.mm and iset.mm theorems that I didn't name.

As for whether the `seq3` and `sum3` convention should persist after we are done converting `seq` syntax, I'm not really sure. If someone has a better idea I'd probably be open to it, but using `iseq` and `isum` (prefixed `i` being our fallback for when set.mm and iset.mm differ) is kinda not great when `isum` is used in set.mm for infinite sum (as opposed to `fsum` for finite sum). I'm not really eager to double down on the `iisum` names I have used occasionally for "intuitionistic infinite sum".

benjub: Ok.  I guess `seq3` will become `seq` when the transition is over ?

jkingdon: In some cases, like `seqf` or `seqeq1`, it already is `seq`.

But the hard part is something like https://us.metamath.org/mpeuni/seq1.html . Assuming we follow "Theorems which are different should be named differently" from https://us.metamath.org/ileuni/conventions.html it can't be `seq1` . Right now it is https://us.metamath.org/ileuni/seq3-1.html which is perhaps an odd convention but is at least documented in the table at the end of https://us.metamath.org/ileuni/conventions.html . Maybe it is OK unless we think of something better?

jkingdon:
> Reviewed the whole thing. Some non-blocking comments. Removing that extraneous argument of seq is indeed for the better.

I guess I'll take this as an approval unless you want to take another look at anything.

jkingdon:
> I would generally use fewer line breaks, and put the "Use xxx instead." right before the discouragement tag

Yeah, maybe that's better. I haven't been using "(together with iseqsst or iseqseq3 if need be)" lately. As for where to put the suggestion of what to use, there are cases where I sort of want to say a little more about what to use instead and why, but I suppose that doesn't really apply so much to all these `seq` syntax things.

iset.mm
benjub: I thought you were adding an `i` in these cases, like `iseq1` ?

jkingdon: > I thought you were adding an `i` in these cases, like `iseq1` ?

Perhaps but that's what I was discussing when I said:

> using iseq and isum (prefixed i being our fallback for when set.mm and iset.mm differ) is kinda not great when isum is used in set.mm for infinite sum (as opposed to fsum for finite sum). I'm not really eager to double down on the iisum names I have used occasionally for "intuitionistic infinite sum".

PR #2931, opened 2022-11-20
generalized ~ ru

icecream17:
Adds the generalized version of ~ ru (taken from [~ pwnss](https://us.metamath.org/mpeuni/pwnss.html))

Edit: ~ rug --> ~ rru, ~ rru will not use ax-13.

<s>~ rug uses the same axioms as ~ ru (and 2-4 more definitions).

By itself this pr adds 3 lines of proof overall, but</s> I plan to use ~ rug to prove [~ pnfnre](https://us.metamath.org/mpeuni/pnfnre.html) and [~ mnfnre](https://us.metamath.org/mpeuni/mnfnre.html) using less axioms.

[~ df-pnf](https://us.metamath.org/mpeuni/df-pnf.html) and [~ df-mnf](https://us.metamath.org/mpeuni/df-mnf.html) would be changed to `+inf = { x e. CC | x e/ x }` and `-inf = { x e. ( CC u. +inf ) | x e/ x }` respectively (as noted in their comments the only requirements are that the infinities be distinct and non-real).

(Wait, non-real. Sounds like `+inf = { x e. RR | x e/ x }` would also be a valid definition? This would remove [~ ax-resscn](https://us.metamath.org/mpeuni/ax-resscn.html)...)

---

Outdated, ru will stay the same, no new proof

> I feel like [~ ru](https://us.metamath.org/mpeuni/ru.html) has been referenced enough that the proof in 1994 is historically significant, even though it's strictly longer and would be deleted under conventions (similar to idALT). I'm unsure about its future so it's ALTV.
>
> So this pr is a draft. If ~ ruALTV is kept ( ~ ruALTV --> ~ ruALT, ~ ruALT --> ~ ruALTALT ) then I think it would be mentioned in > ~ ru like: `For the original proof see ~ ruALT`?
>
> Here's the proof in ~ ruALTV since you can't see it completely in Github's file preview
> ```
>       wal id notbid mtbir syl5bb bibi12d spv mto abeq2 nex isset nelir ) ACZUID
>       ZAEZFUKFGBCZUKHZBIUMBUMABJZUJKZAQZUPBBJZUQLZKZUQMUOUSABUIULHZUNUQUJURABUL
>       NUJAAJZLUTURUIUIOUTVAUQUTUIULUIULUTRZVBPSUAUBUCUDUJAULUETUFBUKUGTUH $.
> ```


set.mm
```diff
@@ -44776,13 +44850,10 @@ holding in an empty domain (see Axiom A5 and Rule R2 of [LeBlanc]
   ${
     $d A x y $.  $d V x y $.
     $( The power set of a set is never a subset.  (Contributed by Stefan
-       O'Rear, 22-Feb-2015.) $)
+       O'Rear, 22-Feb-2015.)  (Revised by Steven Nguyen, 20-Nov-2022.) $)
```
icecream17, line 44853:
```suggestion
       O'Rear, 22-Feb-2015.) $)
```

benjub:
Have you had a look at bj-ru0 / bj-ru1 / bj-ru ?

icecream17:
no
wow that's an even better result
edit: wait it's not generalized, reopening

icecream17:
looks like in ~ bj-ru, ~ bj-abeq2 and ~ bj-spvv are used to reduce ax-13 (seems like ~ abeq2 now doesn't use ax-13 anymore by some upstream improvement?)

So now ~ rug in fact _would_ use an extra axiom...

benjub:
> looks like in ~ bj-ru, ~ bj-abeq2 and ~ bj-spvv are used to reduce ax-13 (seems like ~ abeq2 now doesn't use ax-13 anymore by some upstream improvement?)

Yes, I should update my mathbox accordingly. As for ~rug, I think it's a nice addition (I proposed it in https://groups.google.com/g/metamath/c/GQmelsKtaWQ/m/G1PqaOXJFQAJ but never got to formalize it). Looks like you can remove that extra axiom, can't you ?

Anyway, it's a good idea to put it first in your mathbox, so that people can compare the webpages, which are easier to read.

digama0:
I am against changing the proof of `ru`. It should be marked `(Proof modification is discouraged.)` if it isn't already. The intention is for it to be (roughly) a direct from axioms proof (modulo some basic propositional logic), so you can survey the complete proof and understand it. Deferring to other proofs can make the overall proof longer, and also harder to read in one sitting.

Assuming that we are going to keep the theorem labeled `ruALTV` here, I don't think there is any need for the new proof of `ru`. People can just reference `rug` directly if they need it - it's certainly more useful.

benjub:
By the way: what you labeled ~rug should probably be ~rru since it is a "restricted" version (same labeling scheme as in df-rab and related theorems, e.g., abid/rabid).
I hope to see it soon in your mathbox.

icecream17:
Oops, somehow did something weird involving git

PR #2932, opened 2022-11-21
rru (generalized ~ ru try 2)

icecream17:
I somehow deleted the entire commit history of my repo when pushing <sup>(something something github codespaces only loads latest commit and that was a commit i wanted to revert so technically there was no history when I pushed and it overwrote?)</sup> and now I can't reopen #2931
image

So here's another pr.

I managed to prove ~ rru without ax-13 through the use of `expand`, `minimize`, and `delete` and reproving a few steps. This resulted in a much larger proof (replacing 2 steps in a 12 step proof and ended up with 53 steps) so I wouldn't be surprised if there was a better one.

For convenience, here are some screenshots previewing the html content
image
image
Somehow `improve all /depth 2` found that `CondEq` was fastest (ax-8 happens to be "not used") which is amazing. The `CondEq` section happens to be the section just before `ru`.
image
Didn't really see anything I could use so I just 3tr4i'd into primitives. This 28th step is the same as the 10th step in [~ pwnss](https://us.metamath.org/mpeuni/pwnss.html)
image
Steps 29-51 are the remains of `expand`. `minimize_with` found `vtoclg1f`. And so removing ax-13 from ~ elrab2 was actually much easier than the first part (~ cbvrabv). The last two steps are as in the original ~ pwnss proof.

(Expand chain: elrab2 elrab elrabf elrabgf, vtoclgf replaced with vtoclg1f. Maybe if this were ported to main, then ~ elrab would be changed to not use ~ elrabf thus avoiding dependency on ax-13. Edit: The new ~ elrab could use ~ elabg if ~ elabg was changed to not use ~ elabgf)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ We're a friendly community, and we would love to have more collaborators!
 
 This document briefly explains
 [how to contribute proposed changes](#how-can-i-contribute)
-("contributions" or "merge requests") to this collection of databases
+("contributions" or "merge requests") to this collection of databases.
 We also discuss
 [Formatting recommendation prior to submitting a pull request](#formatting-recommendation-prior-to-submitting-a-pull-request), and for the rare cases
 when it's needed, information on
@@ -120,7 +120,7 @@ regenerate the `discouraged` file. So if you're just starting out, you can
 probably ignore this section. However, there are rare cases when that
 is needed. This section explains what the discouraged file is all about.
 
-Some statement descriptions in `set.mm` and ` iset.mm` have one or both of the
+Some statement descriptions in `set.mm` and `iset.mm` have one or both of the
 markup tags `(New usage is discouraged.)` and `(Proof modification is discouraged.)`.
 In order to monitor accidental violations of these tags in set.mm and iset.mm,
 we store the usage and proof lengths of statements with these tags in a file
--- a/discouraged
+++ b/discouraged
@@ -19595,6 +19595,7 @@ Proof modification of "rpreOLD" is discouraged (21 steps).
 Proof modification of "rpssreOLD" is discouraged (7 steps).
 Proof modification of "rspsbc2" is discouraged (65 steps).
 Proof modification of "rspsbc2VD" is discouraged (90 steps).
+Proof modification of "ru" is discouraged (88 steps).
 Proof modification of "ruALT" is discouraged (29 steps).
 Proof modification of "rucALT" is discouraged (25 steps).
 Proof modification of "rusbcALT" is discouraged (77 steps).
--- a/discouraged
+++ b/discouraged
@@ -15256,6 +15256,7 @@ New usage of "el123" is discouraged (1 uses).
 New usage of "el2122old" is discouraged (1 uses).
 New usage of "elALT" is discouraged (0 uses).
 New usage of "ela" is discouraged (3 uses).
+New usage of "elabgOLD" is discouraged (0 uses).
 New usage of "elat2" is discouraged (4 uses).
 New usage of "elatcv0" is discouraged (0 uses).
 New usage of "elbdop" is discouraged (4 uses).
@@ -15306,6 +15307,7 @@ New usage of "elpqn" is discouraged (24 uses).
 New usage of "elprnq" is discouraged (22 uses).
 New usage of "elpwgded" is discouraged (2 uses).
 New usage of "elpwgdedVD" is discouraged (1 uses).
+New usage of "elrabOLD" is discouraged (0 uses).
 New usage of "elreal" is discouraged (7 uses).
 New usage of "elreal2" is discouraged (3 uses).
 New usage of "elresOLD" is discouraged (0 uses).
@@ -18899,6 +18901,7 @@ Proof modification of "el12" is discouraged (19 steps).
 Proof modification of "el123" is discouraged (26 steps).
 Proof modification of "el2122old" is discouraged (25 steps).
 Proof modification of "elALT" is discouraged (27 steps).
+Proof modification of "elabgOLD" is discouraged (13 steps).
 Proof modification of "eleq2dALT" is discouraged (62 steps).
 Proof modification of "elex22VD" is discouraged (111 steps).
 Proof modification of "elex2VD" is discouraged (59 steps).
@@ -18916,6 +18919,7 @@ Proof modification of "elneldisjOLD" is discouraged (53 steps).
 Proof modification of "elnelunOLD" is discouraged (53 steps).
 Proof modification of "elpwgded" is discouraged (23 steps).
 Proof modification of "elpwgdedVD" is discouraged (23 steps).
+Proof modification of "elrabOLD" is discouraged (16 steps).
 Proof modification of "elrefsymrels3" is discouraged (65 steps).
 Proof modification of "elresOLD" is discouraged (171 steps).
 Proof modification of "elridOLD" is discouraged (107 steps).


set.mm
```diff
@@ -611927,6 +611927,35 @@ fixed reference functional determined by this vector (corresponding to
...
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  Russell's paradox
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  ${
+    $d x y z A $.
+    $( Russell's Paradox, generalized to elementhood in all classes.  See ~ ru
+       for the non-generalized version.
```
benjub, line 611939: I would propose:
> Relativized version of Russell's paradox ~ ru (which corresponds to ` A = _V ` ).

set.mm
```diff
@@ -32039,7 +32039,7 @@ his New Foundations set theory (axiom system NF of [Quine] p. 331).  In
        ~ elirrv (derived from the Axiom of Regularity), so for us the Russell
        class equals the universe ` _V ` (theorem ~ ruv ).  See ~ ruALT for an
        alternate proof of ~ ru derived from that fact.  (Contributed by NM,
-       7-Aug-1994.) $)
+       7-Aug-1994.)  (Proof modification is discouraged.) $)
```
benjub, line 32042: @digama0 : do we really want to add the discouragement tag ? (if yes, we should provide a short reason, to write just before the tag)

benjub:
Since it's an elementary theorem, I would use only primitive symbols, i.e., `-. e.` instead of `e/` (especially since you're currently mixing both).

Yes, there are some possible shortenings, and I'm surprised `minimize_with` didn't catch anything. For instance, there are `CondEq` theorems, which should not be used (e.g., Lines 5 and 12 are proved directly via elequ1/elequ2)

@digama0: what's the status of the CondEq section ? Can we move it to your mathbox ?


> Steps 29-51 are the remains of expand. minimize_with found vtoclg1f. And so removing ax-13 from ~ elrab2 was actually much easier than the first part (~ cbvrabv). The last two steps are as in the original ~ pwnss proof.

Yes, I introduced `vtoclg1f` with that purpose, so it's nice to see another use for it.

benjub:
> (ax-8 happens to be "not used")

Actually, everything using df-clel (like eleq1/2) "morally" uses ax-8, as explained in https://us.metamath.org/mpeuni/bj-df-clel.html We should probably adopt that definition. I'll propose it in a PR when I have time.

set.mm
icecream17: I know it's probably simply "relative to all classes"... but I've never heard "relative" used in that way and I can't find anything except https://us.metamath.org/mpeuni/bj-disj2r.html

Would "restricted" be better here?

benjub: I would use "relative". First, "restricted" could be understood as "special case", which is not the case here. Here, `|- -. { x e. A | -. x e. x } e. A ` is Russell's result *relative to A*. The relative point of view is useful for instance when doing stuff "relative to a universe" in set theory, or in a somewhat different way https://en.wikipedia.org/wiki/Grothendieck%27s_relative_point_of_view

By the way: I would prefer the formulation I wrote above, with `-.  X e.` instead of a mix of `-. X e.` and `X e/`.

jkingdon:
I removed the `mathbox only` label because of the change to make proof modification of ru  discouraged.

benjub:
Good for me, except for possibly adding the discouragement tag to ~ru. So I'll let @jkingdon or @digama0 see what they prefer. If the discouragement tag is kept, then there should be a short explanation why proof modification is discouraged, right before the tag.

david-a-wheeler:
Here's what @digama0 said about making `ru` proof modification discouraged:

> I am against changing the proof of ru. It should be marked (Proof modification is discouraged.) if it isn't already. The intention is for it to be (roughly) a direct from axioms proof (modulo some basic propositional logic), so you can survey the complete proof and understand it. Deferring to other proofs can make the overall proof longer, and also harder to read in one sitting.
> Assuming that we are going to keep the theorem labeled ruALTV here, I don't think there is any need for the new proof of ru. People can just reference rug directly if they need it - it's certainly more useful.

That makes sense to me. If we do that, then I suggest:

1. The proof of `ru` should say something like, "This proof is intentionally developed as being a proof direct from axioms (modulo some basic propositional logic). For a simpler proof, or a proof you want to build on further, use ~ rru instead. .... (Proof modification is discouraged.) (New usage is discouraged.)".
2. The proof of `rru` should say something like, "This proof intentionally minimizes the number of axioms depended on, but is intentionally simplified by building on other theorems. For a proof direct from axioms (modulo some basic propositional logic), see ~ ru ."

Note the cross-reference between them, so from one it's easy to see the other (and why there is both).

benjub:
Given the previous comment, let's keep the discouragement tag (even if I think it is not necessary). I'd prefer to merge as is, and we can add a **short** reason for that tag later.

benjub (approved):

jkingdon:
I assume the changes to `elabg` and `elrab` in https://github.com/metamath/set.mm/pull/2932/commits/98feb9883445d56da91d4f542140ea10ce9bf4ed are fine. At least, they seem to match what was discussed in comments here on github.

benjub:
> @digama0: what's the status of the CondEq section ? Can we move it to your mathbox ?

@digama0 (and @jkingdon for iset.mm): can I move that experimental section to your mathbox ?

digama0:
It's intended to support metamath metatheory, but this kind of method is maybe too invasive. It's kind of pointless if it moves to my mathbox since it has to come before the definitions it is justifying, but there has been little motion on this front except that I discovered that it is easier to just start from scratch and build a system with a better metatheory (MM0) to support definitions. You may as well just delete it if you want to remove it from main.

jkingdon:
> It's intended to support metamath metatheory, but this kind of method is maybe too invasive

Regardless of what happens with set.mm, this sounds like an argument for deleting it from iset.mm. Unless shown otherwise my assumption is always that the metatheory would be different there.

digama0:
No, at least as far as this is concerned the way definitions are handled in iset.mm vs set.mm is basically identical. You would need a much more different setup like ql.mm or miu.mm for definitions to behave (relevantly) different.

benjub:
Thanks @digama0. I understand the intent of that section and I think we should keep it. In order not to break the flow of the standard presentation of set theory, I wanted to put it in a mathbox, or maybe at a later point in Main. The fact that it actually comes before set theory definitions is not that important in my opinion since it is more an explanation than something later theorems formally depend on (and actually, it also justifies definitions like df-sb, which comes before set theory anyway).

If you insist, then maybe it can be moved to 2.1.4 (instead of the current 2.1.7) to be a bit more consistent. Also, I'm not sure introducing the new syntax `CondEq` gains much. What about stating the theorems simply like e.g., `|- ( x = y ->  ( ph <-> ps ) )  =>  |- ( x = y ->  ( A. x ph <-> A. y ps ) )` without introducing new syntax ?

Tell me what you think and I'll see if I can do the changes.

digama0:
> Also, I'm not sure introducing the new syntax `CondEq` gains much. What about stating the theorems simply like e.g., `|- ( x = y -> ( ph <-> ps ) ) => |- ( x = y -> ( A. x ph <-> A. y ps ) )` without introducing new syntax ?

That was done to overcome a technical restriction in the automation, which is that the parsed formula expression should have three easily identified arguments, for `x`, `y` and `ph`. Without  `CondEq` it is more difficult to recognize a term like `(wi (wceq (cv X) (cv Y)) PH)` as having three arguments, esp. when doing this generically where we don't want to hard code names like `wi`, and moreover to recognize that we should use e.g. `eqeq12` (which does not have any `X` and `Y` terms in the LHS) with the particular substitution `ph = (wceq (cv X) (cv Y))` to derive the substitution proof.

Basically, proof automation likes to have carefully curated lemmas, stated in a way that makes them easy to stick together without intermediate steps, rather than a more "human" approach making use of more general purpose lemmas with more complicated substitutions in them. (By the way, it's no coincidence that these lemmas came up in an `IMPROVE` search - the proofs really are shorter, at least if you also count the substitution construction steps in addition to regular proof steps.)

jkingdon:
> at least as far as this is concerned the way definitions are handled in iset.mm vs set.mm is basically identical

Fair enough.

Y'all can go back to figuring this out.

benjub:
Ok. As you wrote in the head comment of that section, the expression `CondEq(x = y -> ph)` is a bit deceptive. Wouldn't it be more transparent to write `CondEq(x , y , ph)` ?

**edit:** another question: is the term "conditional equality" appropriate ? This machinery to prove the metatheorem has to do with what we call in set.mm "implicit substitution". Maybe `Subst` is better than `CondEq`?

Also, IIUC, the reason to have a three-arguments `CondEq` is to treat simultaneously two cases:
```
Subst_prop(x, y, ph, ps) defined as (x = y -> (ph <-> ps))
Subst_class(x, y, A, B) defined as (x = y -> (A = B))
```
?

PR #2933, opened 2022-11-24
Convert more seq theorems in iset.mm to the preferred syntax

jkingdon:
Includes a theorem which will later become the definition of `seq` once we finish all this.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -167,7 +167,6 @@
 "iseq1" is used by "iseqz".
 "iseq1" is used by "sumsnf".
 "iseq1t" is used by "iseqsst".
-"iseq1t" is used by "seq3-1".
 "iseqcl" is used by "fisum".
 "iseqcl" is used by "ialgrp1".
 "iseqcl" is used by "ibcval5".
@@ -407,7 +406,7 @@ New usage of "hbs1" is discouraged (5 uses).
 New usage of "idALT" is discouraged (0 uses).
 New usage of "iisermulc2" is discouraged (2 uses).
 New usage of "iseq1" is discouraged (17 uses).
-New usage of "iseq1t" is discouraged (2 uses).
+New usage of "iseq1t" is discouraged (1 uses).
 New usage of "iseqcl" is discouraged (13 uses).
 New usage of "iseqclt" is discouraged (2 uses).
 New usage of "iseqdistr" is discouraged (1 uses).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -249,7 +249,6 @@
 "iseqp1" is used by "iseqz".
 "iseqp1" is used by "isermono".
 "iseqp1t" is used by "iseqsst".
-"iseqp1t" is used by "seq3p1".
 "iseqshft2" is used by "seq3shft2".
 "iseqsplit" is used by "ibcval5".
 "iseqval" is used by "iseq1".
@@ -423,7 +422,7 @@ New usage of "iseqhomo" is discouraged (1 uses).
 New usage of "iseqm1" is discouraged (5 uses).
 New usage of "iseqoveq" is discouraged (0 uses).
 New usage of "iseqp1" is discouraged (16 uses).
-New usage of "iseqp1t" is discouraged (2 uses).
+New usage of "iseqp1t" is discouraged (1 uses).
 New usage of "iseqshft2" is discouraged (1 uses).
 New usage of "iseqsplit" is discouraged (1 uses).
 New usage of "iseqval" is discouraged (4 uses).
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -99507,6 +99503,47 @@ the previous value and the value of the input sequence (second operand).
      version entirely.  (Contributed by Jim Kingdon, 3-Oct-2022.) $)
   df-seq3 $a |- seq M ( .+ , F ) = seq M ( .+ , F , _V ) $.
 
+  ${
+    $d .+ x y $.  $d F x y $.  $d M x y $.
+    $( Define a general-purpose operation that builds a recursive sequence
+       (i.e. a function on the positive integers ` NN ` or some other upper
+       integer set) whose value at an index is a function of its previous value
```
benjub, line 99510: Looking at the statement, I see no reason to mention `NN` ? So:
> (i.e., a function on some upper set of integers)

iset.mm
```diff
@@ -99507,6 +99503,47 @@ the previous value and the value of the input sequence (second operand).
...
+       and the value of an input sequence at that index.  This definition is
+       complicated, but fortunately it is not intended to be used directly.
+       Instead, the only purpose of this definition is to provide us with an
+       object that has the properties expressed by ~ seqf , ~ seq3-1 and
+       ~ seq3p1 .  Typically, those are the main theorems that would be used in
+       practice.
+
+       The first operand in the parentheses is the operation that is applied to
+       the previous value and the value of the input sequence (second operand).
+       The operand to the left of the parenthesis is the integer to start from.
```
benjub, line 99520: Another attempt:
> The first argument of `seq` is the integer to start from.  The third argument is the input sequence.  The second argument is the binary operation used to combine the recursively defined values of the resulting sequence with the input sequence.  For instance, `seq M ( + , F )` is the sequence of partial sums of the sequence `F` (starting at the integer `M`) and `seq M ( x. , F )` is the sequence of its partial products.

iset.mm
jkingdon: I have mixed feelings about this. On the one hand, the definition is indeed not specific to `NN` at all. On the other hand, "upper integer set" is metamath-specific terminology (as far as I could tell - I tried `"upper integer set"` in two search engines, one of which had no results and one of which just had metamath on the first page) and just seems a bit awkward to me.

Given that this wording has been at https://us.metamath.org/mpeuni/df-seq.html for a long time I'm a bit reluctant to try to tackle this as part of this pull request. But I don't know, maybe we can change ``a function on the positive integers ` NN ` or some other upper integer set`` to ``a function on an upper integer set such as ` NN ` or ` NN0 ` `` without opening too many cans of worms.

benjub: That's why I wrote "upper set of integers" above, which is standard.
Yes, something like your last proposal sounds better (with "upper set of integers").

iset.mm
```diff
@@ -99854,6 +99854,84 @@ the previous value and the value of the input sequence (second operand).
...
+          <. M , ( F ` M ) >. ) $.
+    seq3val.f $e |- ( ( ph /\ x e. ( ZZ>= ` M ) ) -> ( F ` x ) e. S ) $.
+    seq3val.pl $e |- ( ( ph /\ ( x e. S /\ y e. S ) ) -> ( x .+ y ) e. S ) $.
+    $( Value of the sequence builder function.
+
+       This helps us expand the definition although there should be little need
+       for it once we have proved ~ seqf , ~ seq3-1 and ~ seq3p1 , as further
+       development can be done in terms of those.
+
+       (Contributed by Mario Carneiro, 24-Jun-2013.)  (Revised by Jim Kingdon,
```
benjub, line 99879: Probably no line breaks needed.  I say this considering the pages mmtheoremsXXX.html, where they tend to add up.  Of course, that's a judgment call.

benjub: Probably "us" is not needed either.

benjub (approved):
See non-blocking comments above.

iset.mm
jkingdon: "upper set of integers" is no better than "upper integer set" in the searches I did.

Anyway, I'll go with the above.

iset.mm
jkingdon: I'm going to leave this for someone else and another pull request. Just numbering the arguments to `seq` is tricky.

iset.mm
jkingdon: Changed.

jkingdon:
> See non-blocking comments above.

I'll plan on merging this once the tests pass.

iset.mm
benjub: https://en.wikipedia.org/wiki/Upper_set

jkingdon: Yeah, I can see where we got it from and I'm not sure there is any better terminology out there. I'm just advocating being a bit cautious, in the sense that I don't think most authors would talk about upper sets in the context of how to index a sequence.

benjub: This strikes me as particularly bad faith. The only way one can pretend to misunderstand "upper set of integers" is by saying that the order relation on integers hasn't been specified, which is technically true, but come on.

PR #2935, opened 2022-11-26
Misc

benjub:
See commit messages. In particular:
* move sbimd, sbbid before ax-13, following @wlammen's recent PR #2934
* move sb4v to Main following #2719

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18511,7 +18511,6 @@ Proof modification of "bj-ru" is discouraged (22 steps).
 Proof modification of "bj-ru0" is discouraged (50 steps).
 Proof modification of "bj-ru1" is discouraged (27 steps).
 Proof modification of "bj-sb3v" is discouraged (25 steps).
-Proof modification of "bj-sb4v" is discouraged (25 steps).
 Proof modification of "bj-sb56" is discouraged (50 steps).
 Proof modification of "bj-sbab" is discouraged (17 steps).
 Proof modification of "bj-sbceqgALT" is discouraged (143 steps).


set.mm
```diff
@@ -18869,6 +18869,26 @@ Converse of the inference rule of (universal) generalization ~ ax-gen .
...
+    sbimd $p |- ( ph -> ( [ y / x ] ps -> [ y / x ] ch ) ) $=
+      ( weq wi wa wex wsb imim2d anim2d eximd anim12d df-sb 3imtr4g ) ADEHZBIZS
+      BJZDKZJSCIZSCJZDKZJBDELCDELATUCUBUEABCSGMAUAUDDFABCSGNOPBDEQCDEQR $.
+  $}
+
+  ${
+    sbbid.1 $e |- F/ x ph $.
+    sbbid.2 $e |- ( ph -> ( ps <-> ch ) ) $.
+    $( Deduction substituting both sides of a biconditional.  (Contributed by
+       NM, 30-Jun-1993.)  Remove dependency on ~ ax-10 and ~ ax-13 .  (Revised
```
wlammen, line 18886: What is the purpose of expanding the revision reason to this detail?  The proof became longer, but that is justified because its axiom footprint is smaller.  Is there a need to express more?

set.mm
```diff
@@ -20885,25 +20913,9 @@ primitive connectives (no defined terms) on the right-hand side.
...
-    $( Deduction substituting both sides of a biconditional.  (Contributed by
-       NM, 30-Jun-1993.)  Reduce axiom usage.  (Revised by Wolf Lammen,
-       24-Nov-2022.) $)
-    sbbid $p |- ( ph -> ( [ y / x ] ps <-> [ y / x ] ch ) ) $=
-      ( wsb biimpd sbimd biimprd impbid ) ABDEHCDEHABCDEFABCGIJACBDEFABCGKJL $.
-
-    $( Obsolete proof of ~ sbbid as of 24-Nov_2022.  (Contributed by NM,
+    sbbidOLD.1 $e |- F/ x ph $.
+    sbbidOLD.2 $e |- ( ph -> ( ps <-> ch ) ) $.
+    $( Obsolete proof of ~ sbbid as of 24-Nov-2022.  (Contributed by NM,
```
wlammen, line 20918: The conventions page requires "Obsolete version..." as a marker for OLD theorems.  Recently I approved a change where a redundant antecedent was removed, so "Obsolete proof.." is too narrow as a general expression.

This was my fault, and it was on my TODO list.  But since you modify the lines anyway, could you do me a favor and correct this glitch?

set.mm
benjub: I think this information can be useful, especially since we're trying to chase dependencies on ax-13. For instance, one may want to locate proofs where one succesfully removed it, to take inspiration. Plus, it's not very long to state it (if it's only one or two axioms, it's not much longer than "reduce axiom usage").

set.mm
benjub: I can. But I would rather modify the conventions, with "obsolete proof" when the statement is identical and "obsolete version otherwise", similar to "alternate proof"/"alternate version". The current comments need not be changed since they will be removed within a year.
To locate them for removal, I generally search the regexp "as of ..-Oct-2021" so this wouldn't change.

What do you think ?

wlammen: If you think of this phrase as a marker, a fixed one is preferable over a variety, I think.  In the end this helps automate the removal of outdated OLD theorems  -  some time in the future.  Searching for the date will succeed most often, but if the range crosses a change of a month, or even more, this method has some drawbacks.

wlammen (approved):

set.mm
wlammen: thanks

PR #2938, opened 2022-11-28
Ratio test for iset.mm

jkingdon:
This is based on cvgrat in set.mm but adds a 0 < A condition (the set.mm proof goes through quite a few contortions to remove this condition but the theorem isn't really very interesting in the A <_ 0 case so at least for now we don't bother).

The proof in set.mm is based on various theorems we have there but do not have in iset.mm (and I suspect cannot have, as they seem to be missing hypotheses about the rate of convergence). So the proof in this pull request is fairly directly from https://us.metamath.org/ileuni/climcvg1n.html

Add cvgrat to mmil.html

Includes cvgratz which is the theorem without having W and Z be different.

Includes cvgratnn which is the theorem where M = 1 . This is not just a stylistic change; proving convergence via climcvg1n is sensitive to how we index the sequence.

Includes lemmas cvgratnnlemnexp and cvgratnnlemmn which are based on a portion of the set.mm proof of cvgrat although changed around in various ways.

Includes lemmas cvgratnnlembern , cvgratnnlemseq , cvgratnnlemabsle , cvgratnnlemsumlt , cvgratnnlemfm , and cvgratnnlemrate

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7947,6 +7947,14 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>cvgrat</TD>
+  <TD>~ cvgratgt0</TD>
+  <TD>Adds a ` 0 < A ` condition which presumably is omitted
+  from the set.mm theorem only for convenience (the theorem isn't
+  interesting unless it holds).</TD>
+</TR>
+
+<TR>
   <TD>df-prod and theorems using it</TD>
   <TD><I>none</I></TD>
   <TD>To define this, we will need to tackle all the issues analogous


iset.mm
```diff
@@ -115790,6 +115790,432 @@ Infinite sums (cont.)
...
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  ${
+    $d A m p $.  $d m ph $.
+    cvgratnnlembern.3 $e |- ( ph -> A e. RR ) $.
+    cvgratnnlembern.4 $e |- ( ph -> A < 1 ) $.
+    cvgratnnlembern.gt0 $e |- ( ph -> 0 < A ) $.
+    $( Lemma for ~ cvgratnn .  The rate at which a constant less than one
+       raised to a power shinks as the exponent increases.  The answer is in
```
avekens, line 115805: shinks?

benjub: Probably "shrinks" was meant, although I don't see why not "decreases". In any case, this is not what the statement is about.
Since p can be made explicit, why not prove the statement with an explicit expression instead of `E. p` ?

jkingdon: "decreases" does seem better but I'm not sure I quite follow "this is not what the statement is about". Is the statement inaccurate or is there just a better way to express the idea?

> Since p can be made explicit, why not prove the statement with an explicit expression instead of E. p ?

Hmm, if memory serves at the time I formulated this I was thinking that `p` would somehow be based on `arch` which isn't readily amenable to being replaced by ceiling in iset.mm.

But between then and when I finished it looks like `p` got a nice closed form, namely `(1 / ((1 / 𝐴) − 1))`. I guess I should probably go back and re-do this. That'll affect this lemma and cvgratnnlemfm and cvgratnnlemrate, so it is probably a few days work. Unless people are eager to merge something now (and submit an issue for the cleanup), I suppose I'll leave this pull request open until I can revise it.

benjub: I haven't looked carefully, but the statement is not a claim about something being decreasing.

From your comment, I thought that you had an expression for $p$, so I think keeping the explicit result is better. I'm willing to wait !

jkingdon: > I haven't looked carefully, but the statement is not a claim about something being decreasing.

Since you are waiting for the other thing, you have time to either look more carefully or figure we won't worry about it. The concept we are trying to talk about is what happens to `A ^ m` as `m` gets larger (in a setting where we care about how fast that happens, or modulus of convergence if we want to talk about it that way).

benjub: You're studying f(m) := A^m.
The statement says that for m large enough, one has f(m) < p/m.
The current comment is hard to parse, but seems to say that f(m) is decreasing.
These are two different things.

Maybe you can just write:
> Upper bound for a geometric progression of positive ratio less than one.

jkingdon: I could elaborate more on what that comment was trying to get at but it is a bit irrelevant because I agree that "Upper bound for a geometric progression of positive ratio less than one." more clearly expresses what needs to be said.

jkingdon:
> Since p can be made explicit, why not prove the statement with an explicit expression instead of E. p ?

I've pushed up the version with the explicit expression.

Ready for re-review.

avekens (approved):

benjub:
> > Since p can be made explicit, why not prove the statement with an explicit expression instead of E. p ?
> 
> I've pushed up the version with the explicit expression.

Much better indeed ! (though `( A / ( 1 - A ) )` looks more readable !)

PR #2939, opened 2022-11-30
Fixed minor latex issue

GinoGiotto:
For some Tex programs like TeXstudio, the <code>\\^</code> command gives the error message: "Please use <code>\mathaccent</code> for accents in mathmode" in the log.

So I found out it's sufficient to replace <code>\\^</code> with<code>\hat </code> in the <code>$t</code> comment of set.mm and this error message will not appear anymore.


set.mm
```diff
@@ -431068,7 +431068,7 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
   latexdef "..." as "\ldots";
 htmldef "..^" as "..^";
   althtmldef "..^" as "..^";
-  latexdef "..^" as "..\^";
+  latexdef "..^" as "..\hat";
```
benjub, line 431071: Good catch. For spacing reasons, I would prefer:
```
  latexdef "..^" as " \,..\hat{}\, ";
```

GinoGiotto: Done

GinoGiotto: Although I just noticed that with your edit the <code>^</code> accent doesn't go over the next letter anymore.

GinoGiotto: For example If the symbol is followed by N with <code>..\hat N</code> it shows:
 hat1

 but with <code> \\,..\hat{}\\, N</code> it shows:
 hat2

I tested metamath.exe with a proof of set.mm and this is how it came out in the generated Tex file.

Is this how it is supposed to look?
https://github.com/metamath/set.mm/pull/2939#discussion_r1035440207

benjub:
Yes, that's the intent. For instance, consider what would happen with `A
..^ ( B + C )`.

In any case, this is not "pretty-printing", so it's better to have a TeX
version that looks like the unicode version and does not depend on the
surrounding symbols.

GinoGiotto:
> Yes, that's the intent. For instance, consider what would happen with `A ..^ ( B + C )`.

Got it

> In any case, this is not "pretty-printing", so it's better to have a TeX version that looks like the unicode version and does not depend on the surrounding symbols. Message ID: ***@***.***>

What Tex version  do you use? I think I'm going to switch too since I encountered many problems with mine.

benjub:
I meant "TeX version" informally, as opposed to the "unicode version" and "gif version".
For my writings, I use LaTeX (actually, pdflatex). For the distributions, I use TeX Live since I'm on linux, but there is MacTeX for Mac and MikTeX for Windows. As for the choice of the editor, it's more a question of taste. Don't use anything too fancy in my opinion. I'm using Texmaker, and before that I used Texworks.

benjub:
I meant "TeX version" informally, as opposed to the "unicode version" and "gif version".
For my writings, I use LaTeX (actually, pdflatex). For the distributions, I use TeX Live since I'm on linux, but there is MacTeX for Mac and MikTeX for Windows. As for the choice of the editor, it's more a question of taste. Don't use anything too fancy in my opinion. I'm using Texmaker, and before that I used Texworks.

benjub (approved):

GinoGiotto:
Thanks for answering, I'll go check them out

PR #2940, opened 2022-11-30
misc axiom reductions + typo fixes

icecream17:
Axiom reductions:

- elab through elabg
- nnssz through nnssre and elz
  - removes (resscn mulcom addass mulass distr 1rid rnegex pre-lttri pre-lttrn pre-ltadd pre-mulgt0)
- vtocl2g through vtoclg

Typos:

- abid1
- df-phi

Revise comment in sb6 to be more clear about which theorem is being specified.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15241,6 +15241,7 @@ New usage of "el123" is discouraged (1 uses).
 New usage of "el2122old" is discouraged (1 uses).
 New usage of "elALT" is discouraged (0 uses).
 New usage of "ela" is discouraged (3 uses).
+New usage of "elabOLD" is discouraged (0 uses).
 New usage of "elabgOLD" is discouraged (0 uses).
 New usage of "elat2" is discouraged (4 uses).
 New usage of "elatcv0" is discouraged (0 uses).
@@ -16728,6 +16729,7 @@ New usage of "nnexALT" is discouraged (3 uses).
 New usage of "nnindALT" is discouraged (0 uses).
 New usage of "nnmulclOLD" is discouraged (0 uses).
 New usage of "nnsscnOLD" is discouraged (0 uses).
+New usage of "nnsszOLD" is discouraged (0 uses).
 New usage of "nonbooli" is discouraged (0 uses).
 New usage of "norm-i" is discouraged (13 uses).
 New usage of "norm-i-i" is discouraged (2 uses).
@@ -17977,6 +17979,7 @@ New usage of "vk15.4j" is discouraged (0 uses).
 New usage of "vk15.4jVD" is discouraged (0 uses).
 New usage of "vmcn" is discouraged (1 uses).
 New usage of "vsfval" is discouraged (4 uses).
+New usage of "vtocl2gOLD" is discouraged (0 uses).
 New usage of "vtoclALT" is discouraged (0 uses).
 New usage of "vtxdusgr0edgnelALT" is discouraged (0 uses).
 New usage of "w-bnj13" is discouraged (5 uses).
@@ -18871,6 +18874,7 @@ Proof modification of "el12" is discouraged (19 steps).
 Proof modification of "el123" is discouraged (26 steps).
 Proof modification of "el2122old" is discouraged (25 steps).
 Proof modification of "elALT" is discouraged (27 steps).
+Proof modification of "elabOLD" is discouraged (10 steps).
 Proof modification of "elabgOLD" is discouraged (13 steps).
 Proof modification of "eleq2dALT" is discouraged (62 steps).
 Proof modification of "elex22VD" is discouraged (111 steps).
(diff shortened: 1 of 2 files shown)


wlammen (approved):

set.mm
```diff
@@ -18955,7 +18955,7 @@ Converse of the inference rule of (universal) generalization ~ ax-gen .
     $d x y $.
     $( Equivalence for substitution.  Compare Theorem 6.2 of [Quine] p. 40.
        Also proved as Lemmas 16 and 17 of [Tarski] p. 70.  The implication "to
-       the left" is ~ sb2v and also holds without disjoint variable condition
+       the left", ~ sb2v , also holds without a disjoint variable condition
```
benjub, line 18958: I'm not sure it's an improvement, but let it be.

set.mm
```diff
@@ -123600,8 +123616,16 @@ distinguish between finite and infinite sets (and therefore if the set size
     wf mp2b eqeltri ) ABCCDZEZFGHHDZHBPBITFJKUAHBLBSCCMMNOQR $.
 
   $( Positive integers are a subset of integers.  (Contributed by NM,
-     9-Jan-2002.) $)
+     9-Jan-2002.)  Reduce dependencies on axioms.  (Revised by Steven Nguyen,
+     29-Nov-2022.) $)
   nnssz $p |- NN C_ ZZ $=
+    ( vx cn cz cv wcel cr cc0 wceq cneg w3o nnre 3mix2 elz sylanbrc ssriv ) ABC
+    ADZBEZPFEPGHZQPIBEZJPCEPKQRSLPMNO $.
```
benjub, line 123623: In this and related cases, it might be globally shorter to prove the NN0 version and then the NN version from it. Have you had / can you have a look at it in a future PR ?

set.mm
```diff
@@ -181253,7 +181277,7 @@ reduced fraction representation (no common factors, denominator
     df-odz $a |- odZ = ( n e. NN |-> ( x e. { x e. ZZ | ( x gcd n ) = 1 } |->
                    inf ( { m e. NN | n || ( ( x ^ m ) - 1 ) } , RR , < ) ) ) $.
 
-    $( Define the Euler phi function (also called _ Euler totient function_),
+    $( Define the Euler phi function (also called _Euler totient function_),
```
benjub, line 181280: Here and in iset.mm, I would write `"Euler totient function"`, that is, double quotes instead of formatting, which should be used sparingly.

set.mm
```diff
@@ -30137,8 +30137,16 @@ choice between (what we call) a "definitional form" where the shorter
     vtocl2g.2 $e |- ( y = B -> ( ps <-> ch ) ) $.
     vtocl2g.3 $e |- ph $.
     $( Implicit substitution of 2 classes for 2 setvar variables.  (Contributed
-       by NM, 25-Apr-1995.) $)
+       by NM, 25-Apr-1995.)  Remove dependency on ~ ax-10 , ~ ax-11 , and
+       ~ ax-13 .  (Revised by Steven Nguyen, 29-Nov-2022.) $)
```
benjub, line 30141: Thanks for doing that. I was not very careful when I removed these dependencies from vtoclg. There are certainly similar cases that could benefit from it. Anyway, I should move bj-elissetv to Main so that we can see a bit clearer in the various dependencies.

set.mm
benjub: Ok, I see. I guess ideally one would have
```
df-nn0 $a |- NN0 = ( rec ( ( x e. _V |-> ( x + 1 ) ) , 0 ) " om ) $.
df-nn $a |- NN = ( NN0 \ { 0 } ) $.
df-zz $a |- ZZ = { n e. RR | ( n e. NN0 \/ -u n e. NN0 ) } $.
```
but that would be a big piece of work.

benjub:
I'm trying to review the changes, but I'm getting from github the very informative message "Oops, something went wrong." (complete with the onomatopoeia). This used to happen from time to time for larger changes, but now this happens more and more often even with small changes... Not sure what can be done about it. Is there a known degradation of github services to non-paying projects ?

set.mm
avekens: Are ther rules/conventions/practices when to use formatting (bold letters in this case) and when double quotes? Who decided to use formatting sparingly? In my opinion, new terms can be shown in bold the first time they are mentioned.

icecream17:
Looking at the latest commit seems to work
https://github.com/metamath/set.mm/pull/2940/commits/255ef884b8fd783fdb3fb42059acd700006d38f6

set.mm
benjub: I don't think there are written conventions about that, but I think the closest we stay to plain text, the less the maintenance effort, which is especially important since we do not have much workforce available (for instance: what happens with bold or italic if the rendering engine has to be changed for whatever reason).

benjub (approved):

Issue #2917, opened 2022-10-23
Change definitions of the "shift" function and of the "subword" function


benjub:
I propose to (eventually) amend the definitions df-shift and df-substr as follows:
```
${
$d f x y $.
$( Define the translate of a function (or relation) on complex numbers by a
   given number.  (The name "shift" is historical: its first applications were
   for sequences and translates by integers.) $)
   df-shift $a |- shift =
                 ( f e. _V , x e. CC |- ( f o. ( y e. CC |-> ( y - x ) ) ) ) $.
$}

${
$d w b $.
$( Define the function that gives a subword of a word between the two given
   indices.  (The symbol "substr" is for "substring", since both the terms
   "(sub)word" and "(sub)string" are widely used.) $)
   df-substr $a |- substr = ( w e. _V , b e. ( NN0 X. NN0 ) |->
                                ( ( w |` ( ..^ ` b ) ) shift ( 1st ` b ) ) ) $.
$}
```
(I thought of this after reviewing #2890.)

df-shift: simpler than the current version (formula is less complex, has one fewer dummy variable and no df-opab), and also clearer, since the value is explicitly a composed relation (`o.`). As a corollary, this would simplify some existing theorems (e.g., proving that the shift function is indeed a function, etc.).  Unless I'm overlooking some special case, the value is precisely the same.

df-substr: simpler than the current version (formula is less complex, has one fewer dummy variable and, no df-if, df-dom, df-subset, etc.), and also clearer: it is explicitly defined for indices in `NN0`. The value changes in one way: if one asks for a longer subword, then the returned value is simply truncated (whereas currently, the returned value is the empty set). To me, this change is for the better, and would probably allow to remove some hypotheses in later theorems (and I seem to recall that that's the behavior of most common programming languages? --- at least the "safe" ones, not C).

Cost of the changes: For df-shift, since the value is not changed, the modifications would be rather localized, and not too large.  For df-substr, on the other hand, this would need more work, since the value changes, and many theorems use that definition.  The value of df-prefix, for instance, would also change.  That's why I wrote "eventually" at the beginning.  But probably, most of the affected theorems have an $e-hypothesis or an antecedent stating that the subword/prefix is not going beyond the word, so they would remain true, and at worst would have superfluous hypotheses.

avekens:
[...]

PR #2941, opened 2022-11-30
New definition for "prefixes of words" - additional TODOs done

avekens:
New theorems for the prefix operation:
* pfxnd0
* pfxwrdsymb

Revison of theorems using the prefix operation:
* wlkreslem
* wlkres
* trlreslem
* trlres
* eupthres
* eucrct2eupth1
* eucrct2eupth

Obsolete lemmas:
* wlkreslem0

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -10872,6 +10872,9 @@
 "pexmidlem6N" is used by "pexmidlem8N".
 "pexmidlem7N" is used by "pexmidlem8N".
 "pexmidlem8N" is used by "pexmidALTN".
+"pfxnndmnd" is used by "pfxnd0".
+"pfxnndmnd" is used by "pfxval0".
+"pfxval0" is used by "pfxwrdsymb".
 "phnv" is used by "ajval".
 "phnv" is used by "dipassr".
 "phnv" is used by "dipdi".
@@ -12695,6 +12698,7 @@
 "swrdn0OLD" is used by "clwlksfclwwlkOLD".
 "swrdn0OLD" is used by "clwwlkinwwlkOLD".
 "swrdn0OLD" is used by "wwlksnredOLD".
+"swrdnznd" is used by "swrdnnn0nd".
 "swrdswrd0OLD" is used by "swrd0swrd0OLD".
 "swrdtrcfv0OLD" is used by "clwlkclwwlkOLD".
 "swrdtrcfvOLD" is used by "clwlkclwwlkOLD".
@@ -15534,6 +15538,7 @@ New usage of "fh2" is discouraged (3 uses).
 New usage of "fh2i" is discouraged (1 uses).
 New usage of "fh3i" is discouraged (1 uses).
 New usage of "fh4i" is discouraged (0 uses).
+New usage of "fimadmfoALT" is discouraged (0 uses).
 New usage of "fldcALTV" is discouraged (0 uses).
 New usage of "fldcatALTV" is discouraged (1 uses).
 New usage of "flddivrng" is discouraged (2 uses).
@@ -17128,6 +17133,8 @@ New usage of "pexmidlem5N" is discouraged (1 uses).
 New usage of "pexmidlem6N" is discouraged (1 uses).
 New usage of "pexmidlem7N" is discouraged (1 uses).
 New usage of "pexmidlem8N" is discouraged (1 uses).
+New usage of "pfxnndmnd" is discouraged (2 uses).
+New usage of "pfxval0" is discouraged (1 uses).
 New usage of "phnv" is discouraged (19 uses).
 New usage of "phnvi" is discouraged (22 uses).
 New usage of "phoeqi" is discouraged (1 uses).
@@ -17946,6 +17953,7 @@ New usage of "swrdccatwrdOLD" is discouraged (8 uses).
 New usage of "swrdeqOLD" is discouraged (2 uses).
 New usage of "swrdidOLD" is discouraged (9 uses).
 New usage of "swrdn0OLD" is discouraged (4 uses).
+New usage of "swrdnznd" is discouraged (1 uses).
 New usage of "swrdswrd0OLD" is discouraged (1 uses).
 New usage of "swrdtrcfv0OLD" is discouraged (1 uses).
 New usage of "swrdtrcfvOLD" is discouraged (2 uses).
@@ -19156,6 +19164,7 @@ Proof modification of "falimfal" is discouraged (6 steps).
 Proof modification of "falimtru" is discouraged (6 steps).
 Proof modification of "fesapoOLD" is discouraged (28 steps).
 Proof modification of "festinoALT" is discouraged (24 steps).
+Proof modification of "fimadmfoALT" is discouraged (108 steps).
 Proof modification of "fnexALT" is discouraged (111 steps).
 Proof modification of "fnmpt2ovdOLD" is discouraged (219 steps).
 Proof modification of "fnotovbOLD" is discouraged (79 steps).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -147402,6 +147402,42 @@ computer programs (as last() or lastChar()), the terminology used for
...
     MUKUL $.
 
+  ${
+    $d l s $.
+    $( The value of a prefix operation for a length argument not in the range
+       of the word length is the empty set.  (This is due to our definition of
+       function values for out-of-domain arguments, see ~ ndmfv ).
+       (Contributed by AV, 30-Nov-2022.) $)
+    pfxnd0 $p |- ( ( W e. Word V /\ L e/ ( 0 ... ( # ` W ) ) )
+                   -> ( W prefix L ) = (/) ) $=
```
benjub, line 147412: For information regarding this fact, see #2917.

set.mm
```diff
@@ -404277,34 +404375,34 @@ According to Wikipedia ("Path (graph theory)",
...
     trlres.i $e |- I = ( iEdg ` G ) $.
     trlres.d $e |- ( ph -> F ( Trails ` G ) P ) $.
     trlres.n $e |- ( ph -> N e. ( 0 ..^ ( # ` F ) ) ) $.
-    trlres.h $e |- H = ( F |` ( 0 ..^ N ) ) $.
+    trlres.h $e |- H = ( F prefix N ) $.
     $( Lemma for ~ trlres .  Formerly part of proof of ~ eupthres .
        (Contributed by Mario Carneiro, 12-Mar-2015.)  (Revised by Mario
-       Carneiro, 3-May-2015.)  (Revised by AV, 6-Mar-2021.) $)
+       Carneiro, 3-May-2015.)  (Revised by AV, 6-Mar-2021.)  (Revised by AV,
+       30-Nov-2022.) $)
```
benjub, line 404386: Maybe two "revised" tags in a row are unnecessary, especially if the revisions are not explained.

set.mm
```diff
@@ -147402,6 +147402,42 @@ computer programs (as last() or lastChar()), the terminology used for
...
+      ZQXDWOWQWMUOWOXDXEWGWOWKWGWNWKBCUPZUQRWMWQXDWMXDWQWMWKWGWMCSFZWLTKZWKWMXG
+      KZWMQXHWMXIURXGWLUSUTDEDVHGEVHVAVBHICASJDEVCVDVEVFWMKZWQXDXJWLWQXDPWLVIWL
+      WGWQWKWGWLWQWKPWGWLTWQWHAVJLZWKWGWHVKFZAVKFXKWQNWLWGWNXLXFWHVLVEAVLWHAVMV
+      NWGWLXKWKABCVOVPVQVRVSVTWAWBWCWDWEWERWFWA $.
+  $}
+
+  $( A prefix is a word over the symbols it consists of.  (Contributed by AV,
+     30-Nov-2022.) $)
+  pfxwrdsymb $p |- ( S e. Word A
+                     -> ( S prefix L ) e. Word ( S " ( 0 ..^ L ) ) ) $=
```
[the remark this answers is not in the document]
avekens: Unfortunately, there are no such theorems. You think of something like:
* ( S e. Word A -> ( S substring <. M , N >. ) e. Word ( S " ( M ..^ N ) ) )
* ( S e. Word A -> S e. Word ( S " ( 0 ..^ ( # ` S ) ) ) )

If you want, I can provide/proof these two theorems (~pfxwrdsymb would be a special case of the first one)

set.mm
avekens: OK, I will add an explanation...

set.mm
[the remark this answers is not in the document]
benjub: Yes, it would seem natural that if the version for prefixes is there, it should be proved as a special case of the version of subwords. And the version for words should also be there. So it would be nice in a future PR.

benjub (approved):

avekens:
I am already about to revise set.mm according to @benjub`s remarks.

avekens:
* ~wrdsymb and ~swrdwrdsymb (and auxiliary theorems) added as proposed by BJ
* proof of ~pfxwrdsymb shortened using ~swrdwrdsymb

set.mm
```diff
@@ -57019,6 +57019,14 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
     ( wfo wa wceq wfn fofn fndmu syl2an crn forn sylan9req jca ) ABEFZCDEFZGACH
     ZBDHQEAIECISRABEJCDEJACEKLQRBEMDABENCDENOP $.
 
+  $( A function is a function onto the image of its domain.  (Contributed by
+     AV, 1-Dec-2022.) $)
+  fimadmfo $p |- ( F : A --> B -> F : A -onto-> ( F " A ) ) $=
```
benjub, line 57024: Do we have in set.mm
```
$p |- ( F : A --> B -> F : A -onto-> ran F ) $=
```
? If not, it's worth adding in a future PR.

set.mm
```diff
@@ -146805,6 +146836,48 @@ computer programs (as last() or lastChar()), the terminology used for
       NABCDVQVRWAWBVR $.
   $}
 
+  $( The value of a subword operation for arguments not being nonnegative
+     integers is the empty set.  (Contributed by AV, 2-Dec-2022.) $)
+  swrdnnn0nd $p |- ( ( S e. Word V /\ -. ( F e. NN0 /\ L e. NN0 ) )
+                     -> ( S substr <. F , L >. ) = (/) ) $=
```
benjub, line 146842: Is `S e. Word V` needed ? (same with next theorem)

set.mm
```diff
@@ -147407,36 +147521,28 @@ computer programs (as last() or lastChar()), the terminology used for
     $( The value of a prefix operation for a length argument not in the range
        of the word length is the empty set.  (This is due to our definition of
        function values for out-of-domain arguments, see ~ ndmfv ).
-       (Contributed by AV, 30-Nov-2022.) $)
+       (Contributed by AV, 30-Nov-2022.)  (Proof shortened by AV,
+       3-Dec-2022.) $)
```
benjub, line 147525: I think the "shortened" tag is not needed here considering that it is the same person and three days apart. Similarly at other places, try to reduce tag inflation.

set.mm
avekens: We have `dffn4 $p |- ( F Fn A <-> F : A -onto-> ran F )`. Your theorem follows from this together with ~ffn.

set.mm

set.mm
avekens: In principle, ` S e. Word V ` is not needed (and actually I tried to eliminate it), but then some more auxiliary theorems  would be necessary, which I did not want to provide for these degenerated cases (which holds only because of our definition of function values for out-of-domain arguments).

avekens: I forgot to save my previous comment (saved it now), but I found out that `S e. Word V` is actually needed:
Let S : ( -u 5 ..^ -u 1 ) --> B, then `( S substr <. F , L >. ) =/= (/)` with, for example,  `F = -u 5 e/ NN0` and  `L = -u 3 e/ NN0`. This is because substr is defined for `( F e. ZZ /\ L e. ZZ )`.

set.mm
benjub: I mark as "resolved" since this was only a FYI anyway.

set.mm
benjub: Thanks. You be the judge as to adding it in a future PR.

set.mm
benjub: Thanks. This is a hint that in the future, the definition of `substr` should be stricter so as not to allow the above phenomenon. I mark it as "resolved" here since further discussion can be written at #2917.

benjub:
Ready to merge to me. @avekens, I'll let you do it.

PR #2942, opened 2022-12-04
cbvrabvw (from cbvabvw) + c0exALT

icecream17:
(copied from commit)
use cbvrabvw in rru

The proof for rru is now exactly the same as the subproof in ~ pwnss except that ~ cbvrabv is replaced by ~ cbvrabvw . (and ~ df-nel isn't used)

Interestingly, ~ cbvrabv has the extra (unused) `$d x y $.` (as part of `$d x y A $.`) which makes it exactly equivalent to ~ cbvrabvw . ~ cbvabv doesn't have `$d x y $.` so I think this equivalence is just accidental.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14286,6 +14286,7 @@ New usage of "brresOLD2" is discouraged (1 uses).
 New usage of "brresgOLD2" is discouraged (1 uses).
 New usage of "c-bnj14" is discouraged (131 uses).
 New usage of "c-bnj18" is discouraged (58 uses).
+New usage of "c0exALT" is discouraged (0 uses).
 New usage of "calemesOLD" is discouraged (0 uses).
 New usage of "calemosOLD" is discouraged (0 uses).
 New usage of "camestresOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -613523,6 +613538,14 @@ fixed reference functional determined by this vector (corresponding to
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 $)
 
+  $( Alternate proof of ~ c0ex using more set theory axioms but less complex
```
benjub, line 613541: less --> fewer

benjub:
Good catch. Can you change `$d x y A $.` to `$d x A $.  $d y A $.` in `cbvrabv` in this PR ?  I think this can be done without revision tag.

benjub (approved):

icecream17:
<s>one more commit</s> done

benjub (approved):

avekens (approved):

PR #2943, opened 2022-12-05
Isomorphic graphs

avekens:
Main set.mm:
*  ~f1oeq2d, ~dmexd moved from GS's mathbox to main set.mm

AV's Mathbox:
*  Definition `IsomGr` of the "is isomorphic" relation for graphs and related basic theorems.
*  Specialisation for hypergraphs, simple hypergraphs and simple pseudographs (the latter corresponds to the definition in [Bollobas] p. 3).
* The "is isomorphic" relation for graphs is an equivalence relation (~isomgrref, ~isomgrsym, ~isomgrtr) 
* isomorphic graphs with different representations (~strisomgrop, ~ushrisomgr)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 5-Dec-22 f1oeq2d   [same]      moved from GS's mathbox to main set.mm
+ 5-Dec-22 dmexd     [same]      moved from GS's mathbox to main set.mm
 13-Oct-22 rmo4f     [same]      moved from TA's mathbox to main set.mm
 13-Oct-22 rmo3f     [same]      moved from TA's mathbox to main set.mm
 13-Oct-22 iunxsngf  [same]      moved from TA's mathbox to main set.mm


avekens:
Remark: 
* 34 proofs can be shortened using ~f1oeq2d
* 27 proofs can be shortened using ~dmexd

(will be done in a subsequent PR).

set.mm
```diff
@@ -717097,6 +717100,491 @@ Graph theory (extension)
...
+
+  $( The "is isomorphic" relation is reflexive for hypergraphs.  (Contributed
+     by AV, 11-Nov-2022.) $)
+  isomgrref $p |- ( G e. UHGraph -> G IsomGr G ) $=
+    ( cuhgr wcel cvtx cfv wceq ciedg wa cisomgr wbr id eqid pm3.2i a1i syl21anc
+    isomgreqve ) ABCZQQADEZRFZAGEZTFZHZAAIJQKZUCUBQSUARLTLMNAABPO $.
+
+  ${
+    $d A e f g h i j x $.  $d B e f g h i j x $.  $d Y f g i j $.
+    $( The "is isomorphic" relation is symmetic for hypergraphs.  (Contributed
```
benjub, line 717446: symmetric
(same below)

set.mm
```diff
@@ -717097,6 +717100,491 @@ Graph theory (extension)
...
+      WKXOUWJUUJYIRUVQUWDTUVNUWKYJYIUUKUUJYGYIVGZUVIWIHIUUDYHXNWJWKWLWMWNWOUWAU
+      VOUWIUWAUVORUWHUWEUVOUWAUWHUVOUVSUWFUVTUWGXOUWJUUIYIRUVSUWFTUVNUWKYJYIUUK
+      UUIUWMUVHWIHIUUCYHXNWJWKXOUWJUUJYGRUVTUWGTUVNUWKYJYGUUKUUJUWLUVIWIHIUUDYF
+      XNWJWKWLWMWPWOWRUUCUUDYFYHPWQUAWQJWQKWQWDWSWTVAXAXBUJVCUQUQXCXDVBXEXFUQXF
+      XGXHXEXIWOXJXIXGWO $.
+
+    ${
+      $d F e x $.  $d X e x $.
+      isomuspgrlem2.g $e |- G = ( x e. E |-> ( F " x ) ) $.
+      $( Lemma 1 for ~ isomuspgrlem2 .  (Contributed by AV, 29-Nov-2022.) $)
```
benjub, line 717305: Numbering of lemmas does not agree it comments and in labels (same below).

set.mm
```diff
@@ -717097,6 +717100,491 @@ Graph theory (extension)
...
 $)
 
+$(
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+  Isomorphic graphs
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  $( Introduce new constant symbols. $)
+  $c IsomGr $. $( Predicate "is isomorphic" for graphs $)
```
benjub, line 717110: Here and below, this is "the graph isomorphy relation", or "the isomorphy relation for graphs".

set.mm
```diff
@@ -717097,6 +717100,491 @@ Graph theory (extension)
...
+
+  ${
+    $d f g i x y $.
+    $( Define the "is isomorphic" relation for graphs.  (Contributed by AV,
+       11-Nov-2022.) $)
+    df-isomgr $a |- IsomGr = { <. x , y >. |
+                          E. f ( f : ( Vtx ` x ) -1-1-onto-> ( Vtx ` y )
+            /\ E. g ( g : dom ( iEdg ` x ) -1-1-onto-> dom ( iEdg ` y )
+                      /\ A. i e. dom ( iEdg ` x ) ( f " ( ( iEdg ` x ) ` i ) )
+                                        = ( ( iEdg ` y ) ` ( g ` i ) ) ) ) } $.
```
benjub, line 717124: A general comment: more important than graph isomorphy is the notion of graph isomorphism.  Indeed, in one direction you can define graph isomorphy as the existence of a graph isomorphism between the two graphs, but the extra information of knowing **how** two things are isomorphic rather than merely that they **are**, can be useful. This is a very general pattern.

set.mm
avekens: Why not? There are 5 lemmas for lemma isomuspgrlem2, named isomuspgrlem2a (lemma 1 ), isomuspgrlem2b (lemma 2), etc.

set.mm
avekens: So you propose to define graph isomorphisms as

```
df-grisom $a |- GrIsom = ( x e. _V , y e. _V |-> { <. f , g >. | ( f : ( Vtx ` x ) -1-1-onto-> ( Vtx ` y )
            /\ ( g : dom ( iEdg ` x ) -1-1-onto-> dom ( iEdg ` y )
                  /\ A. i e. dom ( iEdg ` x ) ( f " ( ( iEdg ` x ) ` i ) ) = ( ( iEdg ` y ) ` ( g ` i ) ) ) ) } )
```

Then `A IsomGr B <-> ( A GrIsom B ) =/= (/)` (By the way, is there at most one graph isomorphism between two graphs?).
I can add this alternate definition/concept to the header comment of the section, and it can be considered if/before this section is moved to main.

benjub: I haven't checked the precise definition, since I do not know how the various kinds of graphs are encoded in set.mm, but that's the idea. For graph isomorphy, I would use a definition like `A IsomGr B <-> E. f f e. ( A GrIsom B )`, that is, inhabited set versus nonempty set, to be compatible with intuitionistic logic (for when @jkingdon reaches that part of the database!), and as a theorem the version with "nonempty".

Actually, graph isomorphy will be a secondary notion, present only in "final" theorems, but probably most intermediate theorems will use explicitly in their hypotheses or conclusion things like `F e. ( A GrIsom B)`.

Best would probably be to define a category of graphs (there are multiple ones, I haven't looked at the details), where graph morphisms are couples consisting in a function on vertices and a function on edges with required compatibilities, like you wrote. And then, a graph isomorphism is defined as an isomorphism in the category of graphs (something like ``$a |- GraphIsom = ( Isom ` GraphCat ) $.``). I don't remember how much category theory is present in set.mm to make this approach doable/useful, maybe @digama0 knows. With that, you'll be able to use general category theory theorems (provided they are in set.mm). E.g., you'll have that graph isomorphy is an equivalence relation for free.

To your last question: yes, a graph can have multiple automorphisms (take a complete graph and consider vertex permutations).

set.mm
avekens: Typo fixed

set.mm
avekens: I put most of this material into the section header comment, and also provided a definition for `GrIsom` as described above. This should be sufficient for the moment.

set.mm
```diff
@@ -717119,6 +717125,529 @@ Graph theory (extension)
...
+
+  ${
+    $d f g i x y $.
+    $( Define the isomorphy relation for graphs.  (Contributed by AV,
+       11-Nov-2022.) $)
+    df-isomgr $a |- IsomGr = { <. x , y >. |
+                          E. f ( f : ( Vtx ` x ) -1-1-onto-> ( Vtx ` y )
+            /\ E. g ( g : dom ( iEdg ` x ) -1-1-onto-> dom ( iEdg ` y )
+                      /\ A. i e. dom ( iEdg ` x ) ( f " ( ( iEdg ` x ) ` i ) )
+                                        = ( ( iEdg ` y ) ` ( g ` i ) ) ) ) } $.
```
benjub, line 717188: Now that you have df-grisom, there is no reason not to define
```
$a |- ( A IsomGr B <-> E. f f e. ( A GrIsom B ) ) $.
```
(?)

set.mm
benjub: The numbering of lemmas is often "flattened", that is, they are numbered from lem1 to lemX, even if, say, Lemmas 1 and 2 are only used in Lemma 3. At least, that's what I say in the main part.

set.mm
avekens: Yes, this would be possible, but I would have to change/add some theorems, for which I do not want to spend time at the moment. Maybe in a later PR (at least when the material goes to main, if ever...).

benjub (approved):

PR #2944, opened 2022-12-08
shorten pnfxr and sq10 + remove extra hypothesis in decmul1

icecream17:
pnfxr (from pnfex) is just a proof extraction

In trying to update all theorems to use the new decmul1 I ended up shortening sq10 (no tag because it was through `min *`), but I eventually figured it out.

<sup>

there's probably a better way but note to self:
1. Add decmul1NEW

2. Use `show usage decmul1` and `submit` some commands to minimize all theorems using decmul1NEW

3. Rename decmul1 to decmul1OLD and replace all decmul1NEW with decmul1

4. Rewrap proofs

</sup>

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 7-Dec-22 decmul1   [same]      revised - remove unneeded hypothesis
 13-Oct-22 rmo4f     [same]      moved from TA's mathbox to main set.mm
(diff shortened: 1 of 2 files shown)


avekens (approved):

benjub:
> there's probably a better way but note to self:
>     1. Add decmul1NEW
>     2. Use `show usage decmul1` and `submit` some commands to minimize all theorems using decmul1NEW
>     3. Rename decmul1 to decmul1OLD and replace all decmul1NEW with decmul1
>     4. Rewrap proofs

That's usually how we do it, at least I do not see a faster way.

PR #2945, opened 2022-12-09
Add Mertens theorem to iset.mm

jkingdon:
The theorem is the same as mertens in set.mm with one additional
hypothesis. The proof is similar to the set.mm one with the usual
intuitionizing various places, plus recasting all the uses of supremum
to instead just refer to an upper bound (because that is all we are using).

Because we don't (yet at least) have abscvgcvg or anything
  else relating the convergence of a sequence's absolute values
  to the convergence of the sequence itself, we add the condition
  that both the sequence ` F ` and the sequence of its absolute
  values converges (that is, ` seq 0 ( + , F ) e. dom ~~> ` is
  an additional hypothesis beyond what set.mm has).

Also includes more seq syntax updates and copying a few theorems from set.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7861,10 +7861,8 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <TR>
   <TD>isumless</TD>
-  <TD><I>none</I></TD>
-  <TD>Should be provable with the addition of a decidability
-  condition such as the one found in ~ isumss2 and
-  ~ fisumcvg3 .</TD>
+  <TD>~ isumlessdc</TD>
+  <TD>Adds a decidability condition on the index set for the sum</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7724,7 +7724,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <TR>
   <TD>fsumcvg2</TD>
-  <TD>~ fisumcvg2</TD>
+  <TD>~ fsum3cvg2</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7627,12 +7627,6 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
-  <TD>serf0</TD>
-  <TD>~ serif0</TD>
-  <TD>The only difference is the syntax of ` seq ` .</TD>
-</TR>
-
-<TR>
   <TD>iseralt</TD>
   <TD><I>none</I></TD>
   <TD>The set.mm proof relies on caurcvg2 which does not specify
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7734,7 +7734,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <TR>
   <TD>fsumcvg3</TD>
-  <TD>~ fisumcvg3</TD>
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -116197,6 +116276,361 @@ Infinite sums (cont.)
...
+    $d G a k n $.  $d G k n z $.  $d S a k n $.  $d S k n z $.  $d X a n $.
+    $d X n z $.  $d a k n ph $.
+    mertenslemub.gb $e |- ( ( ph /\ k e. NN0 ) -> ( G ` k ) = B ) $.
+    mertenslemub.b $e |- ( ( ph /\ k e. NN0 ) -> B e. CC ) $.
+    mertenslemub.cvg $e |- ( ph -> seq 0 ( + , G ) e. dom ~~> ) $.
+    mertenslemub.t $e |- T = { z | E. n e. ( 0 ... ( S - 1 ) ) z =
+                        ( abs ` sum_ k e. ( ZZ>= ` ( n + 1 ) ) ( G ` k ) ) } $.
+    mertenslemub.elt $e |- ( ph -> X e. T ) $.
+    mertenslemub.s $e |- ( ph -> S e. NN ) $.
+    $( Lemma for ~ mertensabs .  An upper bound for ` T ` .  (Contributed by
```
avekens, line 116295: "upper bound fo ` X `"?

mmil.raw.html
```diff
@@ -7946,6 +7946,17 @@
...
 </TR>
 
+<TR>
+  <TD>mertens</TD>
+  <TD>~ mertensabs</TD>
+  <TD>Because we don't (yet at least) have abscvgcvg or anything
+  else relating the convergence of a sequence's absolute values
+  to the convergence of the sequence itself, we add the condition
+  that both the sequence ` F ` and the sequence of its absolute
+  values converges (that is, ` seq 0 ( + , F ) e. dom ~~> ` is
```
benjub, line 7956: converges --> converge

iset.mm
```diff
@@ -115222,6 +115222,34 @@ Infinite sums (cont.)
...
+    isumless.1 $e |- Z = ( ZZ>= ` M ) $.
+    isumless.2 $e |- ( ph -> M e. ZZ ) $.
+    isumless.3 $e |- ( ph -> A e. Fin ) $.
+    isumless.4 $e |- ( ph -> A C_ Z ) $.
+    isumless.5 $e |- ( ( ph /\ k e. Z ) -> ( F ` k ) = B ) $.
+    isumless.dc $e |- ( ph -> A. k e. Z DECID k e. A ) $.
+    isumless.6 $e |- ( ( ph /\ k e. Z ) -> B e. RR ) $.
+    isumless.7 $e |- ( ( ph /\ k e. Z ) -> 0 <_ B ) $.
+    isumless.8 $e |- ( ph -> seq M ( + , F ) e. dom ~~> ) $.
+    $( A finite sum of nonnegative numbers is less or equal to its limit.
```
benjub, line 115236: less than or equal to

iset.mm
jkingdon: Since `X` is an arbitrary element of `T` (and the bound doesn't depend on `X`), it is an upper bound for `T`.

mmil.raw.html
jkingdon: Changed.

iset.mm
jkingdon: Changed in both set.mm and iset.mm.

avekens (approved):

PR #2946, opened 2022-12-12
Add complex exponentiation to iset.mm

jkingdon:
This is the section "The exponential, sine, and cosine functions" from https://us.metamath.org/mpeuni/ce.html to https://us.metamath.org/mpeuni/efcan.html .

The only theorem which doesn't intuitionize involves product notation (which we don't yet have in iset.mm) and the only one which needs to be stated differently is eftval . Most of the proofs need a little bit of intuitionizing but there weren't any big difficulties. In particular, the invocations of the ratio test and Mertens theorem both worked as expected.

This portion is a self contained chunk because the next theorem is se.tmm is https://us.metamath.org/mpeuni/efne0.html and we'll want to figure out how to prove a version of that for apartness. But that's for another day.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8364,6 +8364,12 @@ Excluded Middle&quot;,
 <A HREF="https://arxiv.org/abs/1904.09193">https://arxiv.org/abs/1904.09193</A>
 </LI>
 
+<LI>
+<A NAME="Rudin"></A> [Rudin] Rudin, Walter, <I>Principles of Mathematical
+Analysis,</I> McGraw-Hill, New York, second edition (1964) [QA300.R916
+1964].
+</LI>
+
 <LI><A NAME="Quine"></A> [Quine] Quine, Willard van Orman, <I>Set Theory
 and Its Logic,</I> Harvard University Press, Cambridge, Massachusetts,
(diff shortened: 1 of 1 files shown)


avekens (approved):

benjub:
I hadn't realized you had gone as far as trigonometry.  Trigonometry in intuitionistic logic, that's quite an achievement !

jkingdon:
> I hadn't realized you had gone as far as trigonometry. Trigonometry in intuitionistic logic, that's quite an achievement !

To be nit-picky about it, this is just `_e` to a complex power generically. It will be a few pull requests from now until we build on that to decompose it into sine and cosine. But it seems to be straightforward from here to there even though it is a long section.

As for this as a direction, it has been on my mind at least since I proved completeness of the real numbers ( https://us.metamath.org/ileuni/ax-caucvg.html ). I've been pretty obsessed with the ratio test because I knew we'd need it (and I could say the same about Mertens Theorem although it was much more recently that I learned we'd need that).

Totally agreed that it is super cool to be proving this stuff rather than just vaguely handwaving in the direction of Bishop (which seems to be what a lot of constructive texts do)! :tada: :rocket:

PR #2947, opened 2022-12-14
1p2e3, nnaddcom, decpmul

icecream17:
The new version of 1p2e3 saves: "ax-8" ax-sep ax-nul ax-pow ax-pr ax-un ax-resscn ax-icn ax-addcl ax-addrcl ax-mulcl ax-mulrcl ax-mulcom and 12 more complex number axioms

I didn't add (Proof shortened by...) on rru because it's more like I accidentally reproved Stefan O'Rear proof in a longer way and then later matched the proof 1-to-1

nnaddcom is one of the results in my research about ~ ax-mulcom.
In a similar proof by induction, nnmulcom could be proved without ~ ax-mulcom, though just using mulcom is better.

~ sqdeccom12lem was surprisingly hard to prove
It's interesting enough that I didn't add any tags

~ cdecadd is added and used in ~ sqdeccom12lem; there's no guarantee of natural numberness because of the subtraction. ~ cdecadd is unoptimized right now, it's mostly ~ decadd expanded

Add the partial products multiplication algorithm. I named it `decpmul` based on the SSE2 instruction name `pmuludq` found at https://en.wikipedia.org/wiki/Grid_method_multiplication

Wikipedia says the partial products algorithm is slower but since standard long multiplication in Metamath is done digit by digit, it could be faster. Based on ~ 235t711 it doesn't seem like one algorithm is obviously better.

The shortening of sqmid3api makes me think sqsumi isn't useful when there's already muladdi...

Finally, fix typo; ` | `` ` should be ` |`` `

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13318,11 +13318,13 @@ New usage of "1ne0sr" is discouraged (1 uses).
 New usage of "1nq" is discouraged (9 uses).
 New usage of "1nqenq" is discouraged (1 uses).
 New usage of "1p1e2apr1" is discouraged (0 uses).
+New usage of "1p2e3OLD" is discouraged (0 uses).
 New usage of "1pi" is discouraged (11 uses).
 New usage of "1pr" is discouraged (16 uses).
 New usage of "1psubclN" is discouraged (0 uses).
 New usage of "1sr" is discouraged (8 uses).
 New usage of "1t1e1ALT" is discouraged (0 uses).
+New usage of "235t711" is discouraged (0 uses).
 New usage of "2atm2atN" is discouraged (0 uses).
 New usage of "2atnelvolN" is discouraged (0 uses).
 New usage of "2bornot2b" is discouraged (0 uses).
@@ -15493,6 +15495,7 @@ New usage of "eufOLD" is discouraged (0 uses).
 New usage of "eujustALT" is discouraged (0 uses).
 New usage of "eumoOLD" is discouraged (0 uses).
 New usage of "eupthresOLD" is discouraged (1 uses).
+New usage of "ex-decpmul" is discouraged (0 uses).
 New usage of "ex-gt" is discouraged (0 uses).
 New usage of "ex-gte" is discouraged (0 uses).
 New usage of "ex-natded5.13" is discouraged (0 uses).
@@ -18276,6 +18279,7 @@ Proof modification of "19.43OLD" is discouraged (72 steps).
 Proof modification of "19.9dOLDOLD" is discouraged (24 steps).
 Proof modification of "1div0apr" is discouraged (77 steps).
 Proof modification of "1p1e2apr1" is discouraged (7 steps).
+Proof modification of "1p2e3OLD" is discouraged (7 steps).
 Proof modification of "1t1e1ALT" is discouraged (13 steps).
 Proof modification of "2bornot2b" is discouraged (26 steps).
 Proof modification of "2clwwlk2clwwlkOLD" is discouraged (782 steps).
@@ -19134,6 +19138,7 @@ Proof modification of "eufOLD" is discouraged (62 steps).
 Proof modification of "eujustALT" is discouraged (188 steps).
 Proof modification of "eumoOLD" is discouraged (18 steps).
 Proof modification of "eupthresOLD" is discouraged (142 steps).
+Proof modification of "ex-decpmul" is discouraged (261 steps).
 Proof modification of "ex-natded5.13" is discouraged (67 steps).
 Proof modification of "ex-natded5.13-2" is discouraged (21 steps).
 Proof modification of "ex-natded5.2" is discouraged (18 steps).
--- a/discouraged
+++ b/discouraged
@@ -13318,7 +13318,7 @@ New usage of "1ne0sr" is discouraged (1 uses).
 New usage of "1nq" is discouraged (9 uses).
 New usage of "1nqenq" is discouraged (1 uses).
 New usage of "1p1e2apr1" is discouraged (0 uses).
-New usage of "1p2e3OLD" is discouraged (0 uses).
+New usage of "1p2e3ALT" is discouraged (0 uses).
 New usage of "1pi" is discouraged (11 uses).
 New usage of "1pr" is discouraged (16 uses).
 New usage of "1psubclN" is discouraged (0 uses).
@@ -18279,7 +18279,7 @@ Proof modification of "19.43OLD" is discouraged (72 steps).
 Proof modification of "19.9dOLDOLD" is discouraged (24 steps).
 Proof modification of "1div0apr" is discouraged (77 steps).
 Proof modification of "1p1e2apr1" is discouraged (7 steps).
-Proof modification of "1p2e3OLD" is discouraged (7 steps).
+Proof modification of "1p2e3ALT" is discouraged (7 steps).
 Proof modification of "1t1e1ALT" is discouraged (13 steps).
 Proof modification of "2bornot2b" is discouraged (26 steps).
 Proof modification of "2clwwlk2clwwlkOLD" is discouraged (782 steps).


set.mm
```diff
@@ -122142,8 +122142,16 @@ ordinal natural numbers (finite integers starting at 0), so that proof
...
+     dependencies on axioms.  (Revised by Steven Nguyen, 12-Dec-2022.) $)
   1p2e3 $p |- ( 1 + 2 ) = 3 $=
+    ( c1 c2 caddc co c3 oveq2i ax-1cn addassi 1p1e2 oveq1i 2p1e3 eqtri 3eqtr2i
+    df-2 ) ABCDAAACDZCDOACDZEBOACNFAAAGGGHPBACDEOBACIJKLM $.
+
+  $( Obsolete version of ~ 1p2e3 as of 12-Dec-2022. 1 + 2 = 3.  (Contributed by
+     David A. Wheeler, 8-Dec-2018.)  (New usage is discouraged.)
+     (Proof modification is discouraged.) $)
+  1p2e3OLD $p |- ( 1 + 2 ) = 3 $=
     ( c2 c1 c3 2cn ax-1cn 2p1e3 addcomli ) ABCDEFG $.
```
benjub, line 122155: That axiom dependency reduction looks great.  Once it is on the website, I'll have a closer look.  Maybe, given that the OLD proof is much shorter, we will keep it as ALT ?

set.mm
```diff
@@ -613950,6 +613958,31 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+    nnadd1com $p |- ( A e. NN -> ( A + 1 ) = ( 1 + A ) ) $=
+      ( vx vy cv c1 caddc co wceq oveq1 oveq2 eqeq12d eqid cn wcel 1cnd addassd
+      weq nncn sylan9eqr ex nnind ) BDZEFGZEUBFGZHEEFGZUEHCDZEFGZEUFFGZHZUGEFGZ
+      EUGFGZHZAEFGZEAFGZHBCAUBEHUCUEUDUEUBEEFIUBEEFJKBCQUCUGUDUHUBUFEFIUBUFEFJK
+      UBUGHUCUJUDUKUBUGEFIUBUGEFJKUBAHUCUMUDUNUBAEFIUBAEFJKUELUFMNZUIULUIUOUJUH
+      EFGUKUGUHEFIUOEUFEUOOZUFRUPPSTUA $.
+
+    $( Addition is commutative for natural numbers.  Uses fewer axioms than
+       ~ addcom .  (Contributed by Steven Nguyen, 9-Dec-2022.) $)
+    nnaddcom $p |- ( ( A e. NN /\ B e. NN ) -> ( A + B ) = ( B + A ) ) $=
```
benjub, line 613974: Can you do these for NN0, or again is there a problem related to the current definition of NN0 ? If so, we may want to revise the definitions, and define NN0 directly and then NN from it.

set.mm
```diff
@@ -613966,14 +613999,14 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
     sqmid3api.b $e |- ( A + N ) = B $.
     sqmid3api.c $e |- ( B + N ) = C $.
     $( Value of the square of the middle term of a 3-term arithmetic
-       progression.  (Contributed by Steven Nguyen, 20-Sep-2022.) $)
+       progression.  (Contributed by Steven Nguyen, 20-Sep-2022.)  (Proof
+       shortened by Steven Nguyen, 10-Dec-2022.) $)
```
benjub, line 614003: Since it is in a mathbox and with the same contributor, the second tag is not necessary. Up to you.

set.mm
```diff
@@ -613998,6 +614031,150 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
   $}
 
+  ${
+    carrydec.a $e |- A e. NN0 $.
+    carrydec.b $e |- B e. NN0 $.
+    carrydec.c $e |- C e. NN0 $.
+    carrydec.1 $e |- ( A + B ) = D $.
+    $( Carry a digit, removing a decimal constructor.  (Contributed by Steven
+       Nguyen, 10-Dec-2022.) $)
+    carrydec $p |- ; A ; B C = ; D C $=
```
benjub, line 614041: This and the following: probably the deduction versions would be more usable, i.e.,
```
$e |- ( ph -> A e. NN0 ) $.
...
$p |- ( ph -> ; A ; B C = ; D C ) $=
```

set.mm
```diff
@@ -613998,6 +614031,150 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+      ( cdc c1 cc0 cmul co caddc dfdec10 0nn0 dec0u eqid nn0cni addid2i decadd
+      eqtri ) ABCIZIJKIALMZUCNMDCIAUCOAKBCDCUDUCEPFGAEQUCRHCCGSTUAUB $.
+  $}
+
+  ${
+    cdeccl.a $e |- A e. CC $.
+    cdeccl.b $e |- B e. CC $.
+    $( Closure for a numeral.  (Contributed by Steven Nguyen, 11-Dec-2022.) $)
+    cdeccl $p |- ; A B e. CC $=
```
benjub, line 614050: Not sure it is very useful, since `; A B` was devised to make sense when A and B are "digits". Maybe @digama0 can tell.

set.mm
benjub: By the way: I'm not sure that `; A ; B C = ; D C` is sensical, and even if that result is used later, I do not know if we want this kind of intermediate lemma in set.mm.  I haven't looked at that part of set.mm in detail, so maybe @digama0 can chime in.

benjub:
Side remark: in cbvabvw and cbvrabvw, you write `which removes ~  ax-13`. Can you write `which removes dependency on ~ ax-11 and ~ ax-13` ?

set.mm
icecream17: yeah using 0 would basically be a longer version of nncn and addcom as far as I can tell

---
pretty much all theorems involving addition and just the digits 1 through 9 are really just ax-1cn, ax-addcl, and ax-addass:

4p5e9 ( 4 + 5 ) = 9
expands to
( ( ( ( 1 + 1 ) + 1 ) + 1 ) + ( ( ( ( 1 + 1 ) + 1 ) + 1 ) + 1 ) ) =
( ( ( ( ( ( ( ( 1 + 1 ) + 1 ) + 1 ) + 1 ) + 1 ) + 1 ) + 1 ) + 1 )

With df-dec and/or multiplication, mulid1 and ax-distr is needed
;11 = ( ( ( 9 + 1 ) x. 1 ) + 1 ) which by mulid1 eventually becomes ( 9 + 2 ) [9p2e11 6p5e11 7p5e12 et al will be improved in the future like this]

;21 is the same thing but ax-distr to expand the multiplication

( 7 x. 3 ) expands with ax-distr into ( ( 7 x. 1 ) + ( ( 7 x. 1 ) + ( 7 x. 1 ) ) )

But the mere addid1 uses every complex number axiom except cnex and supremum...

benjub: Ok. So changing the definition of NN0 wouldn't change that, correct ? Maybe you can add some of this in a comment, if you find a good phrasing and location to put it.

set.mm
avekens: Although allowed by definition, I also would like to avoid such constructs (; should be used only as prefix). I did not find any such construct in set.mm yet.

set.mm
avekens: I agree with @benjub that this theorem may not be useful.

icecream17:
My proof of sqdeccom12 is currently very inefficient so for now that may be in a separate pr

benjub (approved):

set.mm
avekens: I agree with @benjub that the old proof should be kept as ALT theorem (OLD theorems will be removed after one year...).

avekens (changes requested):
See my inline comment, and the failed check must be fixed.

avekens:
The failing check seems to be a technical issue (@wlammen it also occurs in PR #2952). @david-a-wheeler could you have a look at it, please?

avekens (approved):

PR #2948, opened 2022-12-16
More complex exponentiation for iset.mm, from efap0 through efgt1p2

jkingdon:
At the end of my previous pull request, I was worried about `efap0` (the apartness analogue to https://us.metamath.org/mpeuni/efne0.html ). I won't even try to describe how unnecessarily elaborate my intended proof of `efap0` was, but in the process of starting with the first steps mmj2 set up me with the fact that https://us.metamath.org/ileuni/mulap0bad.html doesn't require that ` B ` be apart from zero, and from there, a few little revisions and `efap0` was done.

From this https://us.metamath.org/mpeuni/efgt0.html and a bunch of similar results follow without much fuss.

The rest of the theorems in this pull request are relatively straightforward. None of them need to be stated differently than set.mm.  Only a few of the proofs worked with no changes at all, but with the usual changes around `seq` and apartness and differences in https://us.metamath.org/mpeuni/eftval.html , and so on, it all worked out.

This is a long section, so this pull request is part two of n.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7799,13 +7799,7 @@ intuitionistic and it is lightly used in set.mm</TD>
   <TD>seqabs</TD>
   <TD>~ fsumabs</TD>
   <TD>Finite sums are more naturally expressed with ` sum_ ` rather
-  than ` seq ` especially in iset.mm.</TD>
-</TR>
-
-<TR>
-  <TD>iserabs</TD>
-  <TD><I>none</I></TD>
-  <TD>Presumably provable but the set.mm proof is via seqabs .</TD>
+  than ` seq ` especially in iset.mm.  Use ~ fsum3ser as needed.</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -117094,6 +117094,12 @@ Infinite sums (cont.)
     efcl syl eqeq12d syl5ib necon3d mpi ) ABCZDEFAGHZEFIUDUEEDEUEEJUEAKZGHZLMZE
     UGLMZJUDDEJUEEUGLNUDUHDUIEAOUDUGUDUFBCUGBCAPUFRSQTUAUBUC $.
 
+  $( Exponent of a negative number.  (Contributed by Mario Carneiro,
```
benjub, line 117097: Exponential

iset.mm
```diff
@@ -117163,6 +117163,19 @@ Infinite sums (cont.)
     ALZMNUQHUSOJZDEZVAURUQUSPCHRCVDVAUAUQUSVBUBUCUSHUDUEUQVCADUQAHAUFZUQUGHGSTU
     QUHUKUIUJULUMUQAPCURGSTVEAUNUOUP $.
 
+  $( The exponential function of a real number is a positive real.
```
benjub, line 117166: The exponential of a...

iset.mm
```diff
@@ -117398,6 +117398,32 @@ Infinite sums (cont.)
       WDIXBKWEMWATTOSWGWHNRQZMJZIJWJVPWCXEWIWGIXDUBWHMWBTTOS $.
   $}
 
+  ${
+    $d A k n y $.
+    $( The exponential function of a positive real number is greater than the
```
benjub, line 117403: The exponential of... the **sum of** the first three...

iset.mm
jkingdon: Changed in both set.mm and iset.mm

iset.mm
jkingdon: Changed in both set.mm and iset.mm

iset.mm
jkingdon: Changed in both set.mm and iset.mm

benjub (approved):

jkingdon:
The tests are failing apparently because http://us.metamath.org/downloads/symbols.tar.bz2 is returning zero bytes, not because of anything on this branch.

jkingdon:
> The tests are failing apparently because http://us.metamath.org/downloads/symbols.tar.bz2 is returning zero bytes, not because of anything on this branch.

Reported at https://groups.google.com/g/metamath/c/cs_pmXWwetU

iset.mm
```diff
@@ -117077,6 +117105,325 @@ Infinite sums (cont.)
...
+  ${
+    $( The exponential function's value is always apart from zero.
+       (Contributed by Jim Kingdon, 12-Dec-2022.) $)
+    efap0 $p |- ( A e. CC -> ( exp ` A ) =//= 0 ) $=
+      ( cc wcel ce cfv cneg efcl negcl syl cmul co cc0 cap efcan 1ap0 mulap0bad
+      c1 syl6eqbr ) ABCZADEZAFZDEZAGSUABCUBBCAHUAGISTUBJKQLMANORP $.
+  $}
+
+  $( The exponential function never vanishes.  Corollary 15-4.3 of [Gleason]
+     p. 309.  (Contributed by NM, 13-Jan-2006.)  The same result also holds
```
avekens, line 117117: Should the text after "Contributed by" be placed before it, shouldn't it?

iset.mm
avekens: But shouldn't "the exponential function of" be replaced by "the exponential of" also in the following theorems (~efgt0 etc.)?

avekens (commented):
The problem with the failing check should be solved already.

jkingdon:
> The problem with the failing check should be solved already.

Why do you think that? If I tell it to rebuild the check it still fails. And if I visit http://us.metamath.org/downloads/symbols.tar.bz2 in a browser I get a zero byte download.

avekens:
> > The problem with the failing check should be solved already.
> 
> Why do you think that? If I tell it to rebuild the check it still fails. And if I visit http://us.metamath.org/downloads/symbols.tar.bz2 in a browser I get a zero byte download.

PR #2947 and #2952 had the same problem, but there it vanished somehow. Since the check is performed again here, and still fails, then there is still an issue with it.

iset.mm
benjub: Probably "MM> set width 9999" followed by a search for "exponential function of" will allow to detect most cases.

benjub:
It looks like the bug comes from the directory `https://us.metamath.org/downloads/` for which I get a "403 Forbidden" error when I try to access it. Maybe @david-a-wheeler knows what's happening. Indeed, it appeared already in a few PRs, then disappeared and now appears again, so something is not stable...

By the way: why download metamath from the website instead of the github repository directly ?

**Edit:** for that last question: apparently, the only thing being downloaded from the website is symbols.tar.bz2, which is probably the list of gifs. I already wanted to deprecate gif's, and this appears to be one more reason. Still, the initial question about `https://us.metamath.org/downloads/` should be solved.

david-a-wheeler:
> It looks like the bug comes from the directory https://us.metamath.org/downloads/ for which I get a "403 Forbidden" error when I try to access it. Maybe @david-a-wheeler knows what's happening...

The webserver we're using is currently configured to *not* provide a list of all items in a directory. When you ask it do so, you'll get a 403. It used to do something different, but it's not clear why anyone would need the webserver to provide a list, and I'm generally trying to be restrictive where I can to enhance security.

You *can* request specific files in the directory, e.g., this works:

> https://us.metamath.org/downloads/metamath.pdf

I think this would be easy to change, but do we *want* to change it?

Also:

> By the way: why download metamath from the website instead of the github repository directly ?

Historial reasons. Originally Norm served everything from the site; the use of GitHub came later.

benjub:
> I think this would be easy to change, but do we _want_ to change it?

Thanks for the explanations. The real problem comes from `http://us.metamath.org/downloads/symbols.tar.bz2` which appears to be empty or corrupted, or at least, that's what @jkingdon and I infer from the CI bug (https://github.com/metamath/set.mm/actions/runs/3736342673/jobs/6343906038).

For the time being, we cannot merge PRs because of this.

**edit:** FYI, excerpt from https://github.com/metamath/set.mm/actions/runs/3736342673/jobs/6343906038 :
```
wget -N -q http://us.metamath.org/downloads/symbols.tar.bz2
+ wget -N -q http://us.metamath.org/downloads/symbols.tar.bz2

# Untar the symbols in the current directory so we can use them
# (e.g., to test if they exist).
tar --strip-components=1 -jxf symbols.tar.bz2
+ tar --strip-components=1 -jxf symbols.tar.bz2

bzip2: Compressed file ends unexpectedly;
	perhaps it is corrupted?  *Possible* reason follows.
bzip2: Inappropriate ioctl for device
	Input file = (stdin), output file = (stdout)

It is possible that the compressed file(s) have become corrupted.
You can use the -tvv option to test integrity of such files.

You can use the `bzip2recover' program to attempt to recover
data from undamaged sections of corrupted files.

tar: Child returned status 2
tar: Error is not recoverable: exiting now
```

benjub:
That error is still here, and affects all PRs (now also #2953, #2954). It was also signaled at https://groups.google.com/g/metamath/c/cs_pmXWwetU

Since we do not have access rights to the metamath website, we should really avoid that the CI process depends on the website.  I'll see what can be done (for instance, dropping tests on the GIFs, and actually removing GIFs altogether), but I'm really not well versed in these matters...

digama0:
I took a look at the server logs, and it appears the issue is that the server ran out of space while running `install.sh`. `symbols.tar.bz2` is not the only affected file, the following are all zero-sized:
```
ileuni.tar.gz
metamathsite.tar.bz2
metamathsite.tar.gz
mmsolitaire.tar.gz
quantum-logic.tar.gz
symbols.tar.bz2
symbols.tar.gz
```
I ran bzip myself on the server so hopefully the file should be okay now, but that is at best a short term solution since it will most likely fail again tomorrow.

digama0:
I freed up some space on the server, it looks like the logging was not size-capped. The full metamath site is about 10GB and there are two copies of it on the server, the live one and the one used during generation; before cleanup we were at 23/25GB used so hopefully the 2GB freed should be enough to give the script some head-room. It's not immediately clear to me whether there is something we should be doing differently about the website generation, but for sure if we drop mpegif that's about 4.3GB removed from the site (mpeuni is 2.3GB and most other things are much smaller).

benjub:
> I freed up some space on the server, it looks like the logging was not size-capped. The full metamath site is about 10GB and there are two copies of it on the server, the live one and the one used during generation; before cleanup we were at 23/25GB used so hopefully the 2GB freed should be enough to give the script some head-room. It's not immediately clear to me whether there is something we should be doing differently about the website generation, but for sure if we drop mpegif that's about 4.3GB removed from the site (mpeuni is 2.3GB and most other things are much smaller).

Thanks. I thought there was a way to rerun the CI tests on the PR page, but I can't find it.

To me, the ideal would be to completely drop GIF and replace it with a plain/ASCII text version. In the function in metamath.c that generates the GIF pages, there is probably a line of code saying "if you see that token, then go search for its GIF definition and display it", which could be changed to "if you see that token, then display it".

Also, the CI tests should not rely on the metamath website, as the present incident shows (since, if I'm correct, only two persons have access rights to the website, and they are very busy persons). Is this (the download of symbols.tar.bz2) the only place where it depends on it?

digama0:
I agree it would be great if we could get all CI dependencies to be hosted by github. it's currently being used for the symbols check, which is still relatively important since it prevents people from writing references to GIF symbols that don't exist when creating new tokens. Of course if we dropped the GIF site than we wouldn't need that, nor the `htmldef` commands in the $t comment. But we can get the list of gifs by another means, by downloading them from the metamath-site repo instead of from us.metamath.org itself.

digama0:
I think for the present we should just ignore metamath.exe CI until the server issue is fixed. The daily build runs about 13 hours from now, so let's just see if it's fixed tomorrow.

benjub:
> I agree it would be great if we could get all CI dependencies to be hosted by github. it's currently being used for the symbols check, which is still relatively important since it prevents people from writing references to GIF symbols that don't exist when creating new tokens. Of course if we dropped the GIF site than we wouldn't need that, nor the `htmldef` commands in the $t comment. But we can get the list of gifs by another means, by downloading them from the metamath-site repo instead of from us.metamath.org itself.

I located two other dependencies in https://github.com/metamath/set.mm/blob/develop/.github/workflows/verifiers.yml:
```
      - run: wget -N -q http://us.metamath.org/ocat/mmj2/mmj2jar.zip
[...]
      - run: wget -N -q http://us.metamath.org/downloads/checkmm.cpp
```
which should be hosted in the metamath-site repo too. And I think that's it.

digama0:
mmj2jar.zip should be downloadable from the [mmj2 repo](https://github.com/digama0/mmj2/tree/master/mmj2jar) by just checking it out, and checkmm.cpp is a file [on the metamath-site repo](https://github.com/metamath/metamath-website-seed/blob/main/downloads/checkmm.cpp)

jkingdon:
> I think for the present we should just ignore metamath.exe CI until the server issue is fixed. The daily build runs about 13 hours from now, so let's just see if it's fixed tomorrow.

I can wait to merge anything until then.

Sounds like we expect them to pass after that daily website build.

jkingdon:
> Thanks. I thought there was a way to rerun the CI tests on the PR page, but I can't find it.

You have to click through to the actions page - in this case https://github.com/metamath/set.mm/actions/runs/3736342673/jobs/6407477916 - and there is a "re-run jobs" towards the top of the page there.

benjub:
So in https://github.com/metamath/set.mm/blob/develop/.github/workflows/verifiers.yml we should change
```
      - run: wget -N -q http://us.metamath.org/downloads/checkmm.cpp
```
to
```
      - run: wget -N https://github.com/metamath/metamath-website-seed/blob/main/downloads/checkmm.cpp
```
or is there something special given that both verifiers.yml and checkmm.cpp are in `https://github.com/metamath/` ?

And we should change
```
      - run: wget -N -q http://us.metamath.org/ocat/mmj2/mmj2jar.zip
      - run: unzip -q mmj2jar.zip
```
to
```
      - run: wget -N https://github.com/digama0/mmj2/tree/master/mmj2jar
```
?
(I removed the `-q` options since the outputs may turn out to be useful.)

jkingdon:
> we should change  . . .

Both of those changes look good to me.

david-a-wheeler:
BTW, this is a good direction. Historically the website was hand-managed by Norm, and he liked to download from the website. I've flipped the script, so that the website is now automatically generated, so we should only be depending on version controlled stuff (e.g., GitHub) and not the website to generate the website.

@digama0 : What did you do to modify the logs to limit their size? They are *supposed* to be limited already. We need to put that change into the website generation script, so that when we erase & recreate the website (which is a possibility I want to retain) the change is there. Basically, we need to modify https://github.com/metamath/metamath-website-scripts to do this on a permanent basis.

PR #2949, opened 2022-12-16
Shortened Proofs

avekens:
61 proofs shortened using ~f1oeq2d and ~dmexd moved to main on 5-Dec-22 (see PR #2943).


wlammen (approved):
* All proofs became shorter, measured in proof bytes;
* The deduction versions f1oeq2d and dmesd should not introduce extra axioms in proofs (the diff between the two corresponding theorems covers only ax-mp - ax-3, and these are used anyway);
* This shortening covers mathboxes.  In theory someone could have deliberately avoided one of the deduction versions for whatever reason, but that is very unlikely in my eyes.

avekens:
PS: all proofs were shortened by the minimize script

>     * This shortening covers mathboxes.  In theory someone could have deliberately avoided one of the deduction versions for whatever reason, but that is very unlikely in my eyes.

In this case, the mathbox owner must add the tag "(Proof modification is discouraged.)". Otherwise, the proof would be shortened in a global minimization run (as formerly performed regularily) anyway.

benjub:
Thanks @wlammen for the careful review.

Automatic minimizations can be carried out in mathboxes (and I actually favor including mathboxes), as implied in https://us.metamath.org/mpeuni/mathbox.html : "other people may make maintenance edits to your mathbox for things like keeping it synchronized with the rest of set.mm, reducing proof lengths..."

PR #2950, opened 2022-12-16
shorten nfeqf2

wlammen:
* shorten nfeqf2
* delete outdated OLD theorems
* move nfeqf family before the cbval* set, where it can be used to eliminate dv conditions, if wanted.
* cluster hbal* family

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1372,10 +1372,12 @@
 "ax13lem1" is used by "ax6e".
 "ax13lem1" is used by "nfeqf2".
 "ax13lem1" is used by "nfeqf2OLD".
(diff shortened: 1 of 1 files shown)


wlammen:
The error log entry

`bzip2: Compressed file ends unexpectedly;
	perhaps it is corrupted?  *Possible* reason follows.
bzip2: Inappropriate ioctl for device
	Input file = (stdin), output file = (stdout)`

points to a  technical problem in the verification process not related to the modified files.  I'll trigger a recheck by adding an empty line and deleting it again.

I don't have a problem on my side.  Rewrap runs as expected and metamath shows no flaw whatsoever.

PR #2954, opened 2022-12-23
"Minus times minus is plus"

avekens:
Recently, I stumbled over the citation "Minus times minus is plus, The reason for this we need not discuss." in the Metamath book (p. 5). I wondered if this statement is proved in set.mm, and found that it isn't. Therefore I formalized it and proved it, see ~nnmtmip and ~rpmtmip. Having these theorems, this statement actually "needs not be discussed" anymore. In the comments of these two theorems, I also explained "The reason for this".

Although the proofs of the two theorems are short and straightforward, I think it is more than appropriate to have them explicitly in set.mm, just to avoid the arguments "that's trivial" or even "that easily follows from ~mul2neg" which are criticised in the Metamath book (and by Norm himself, which was one of the starting points to "invent" Metamath). As a consequence, each theorem/matemathical statement  made in the Metamath book should be proved in set.mm.

Are there additional such theorems/matemathical statements in the Metamath book? Maybe the statement in the footnote 27 of page 17: "There exist irrational numbers $a$ and $b$ such that $a^b$ is rational."? I did not find an appropriate theorem in set.mm yet. Maybe I will prove it and provide it with this PR (as I may have time until the verification process is fixed...).


set.mm
```diff
@@ -121414,6 +121414,20 @@ subset of complex numbers ( ~ nnsscn ), in contrast to the more elementary
       ( cn wcel cmul co nnmulcl mp2an ) AEFBEFABGHEFCDABIJ $.
   $}
 
+  $( "Minus times minus is plus, The reason for this we need not discuss."  (W.
+     H. Auden, as quoted in M. Guillen "Bridges to Infinity", p. 64, see also
+     Metamath Book, p. 5) This statement, formalized to "The product of two
+     negative integers is a positive integer", is proved by the following
+     theorem, therefore it actually needs not be discussed anymore.  "The
```
benjub, line 121421: needs --> need

Also: to nitpick, what you proved is "the product of the opposites of two positive integers is a positive integer". A formalization more faithful to the first sentence would be
```
|- ( ( -u A e. NN /\ -u B e. NN ) -> ( A x. B ) e. NN )
```
But that's not really a problem and I'm fine with your statement (though maybe you could add the above remark in the comment).

avekens: > needs --> need

OK, is already corrected.

> Also: to nitpick, what you proved is "the product of the opposites of two positive integers is a positive integer". A formalization more faithful to the first sentence would be
> 
> ```
> |- ( ( -u A e. NN /\ -u B e. NN ) -> ( A x. B ) e. NN )
> ```
> 
> But that's not really a problem and I'm fine with your statement (though maybe you could add the above remark in the comment).

The opposite of a positive integer is a negative integer! Furthermore, the rule "minus times minus is plus" is often presented with explicit integers in the form "-3 x -4 = 12" (especially in school), and that fits to ~ nnmtmip. And finally, since there is no reverse closure of -u, we cannot derive A e. ZZ from -u A e. NN, and therefore the proposed theorem cannot be proven.
To be precise, one would have to define the set of negative integers ZZ- = { z e. ZZ | z < 0 }. Then 

|- ( ( A e. ZZ- /\ B e. ZZ- ) -> ( A x. B ) e. NN )

would be the appropriate theorem.

To make thinks not too complicated, I would like to leave the theorem as it is.

avekens:
As mentioned in the latest commit, the new theorem ~sqrt2irr0 (`( sqrt `` 2 ) e. ( RR \ QQ )`) seems to be a more appropriate formalization of "The square root of 2 is an irrational number. " than the already existing theorem ~sqrt2irr, which only says that "The square root of 2 is not a rational number. ": `( sqrt `` 2 ) e/ QQ`. I would have renamed it (maybe ~sqrt2nq), but this is a Metamath 100 theorem, which is widely referenced. Furthermore, if I rename ~sqrt2irr, and change the label of ~sqrt2irr0 to ~sqrt2irr, the credit for the 1st Metamath 100 theorem would go to me, which is not justified (I just combined two existing theorems).

set.mm
benjub: You're right, the statement should be
```
|- ( ( ( A e. ZZ /\ -u A e. NN ) /\ ( B e. ZZ /\ -u B e. NN ) ) -> ( A x. B ) e. NN )
```
And indeed, it assumes that "a negative integer is an integer whose opposite is a positive integer".
As for deciding which is the better definition:
```
A negative integer is the opposite of a positive integer.
A negative integer is an integer whose opposite is a positive integer.
```
maybe the first version is better indeed !

set.mm
```diff
@@ -353426,6 +353431,36 @@ the complex square root function (the branch cut is in the same place for
...
+    wne syl6eq adantr oveq1d 3eqtr4a a1d cexp sqrtcl simpl simpr sqr00d necon3d
+    id ex imp cz 2z a1i cxpexpzd sqrtth eqtrd expcom pm2.61ine ) ABCZADEZFGHZAI
+    ZJAKAKIZVDVAVEKFGHZKVCAFBCFKRLVFKIMFNOVEVBKFGVEVBKDEKAKDPQSUAVEUJUBUCVAAKRZ
+    VDVAVGLZVCVBFUDHZAVHVBFVAVBBCVGAUETVAVGVBKRVAVBKAKVAVBKIZVEVAVJLAVAVJUFVAVJ
+    UGUHUKUIULFUMCVHUNUOUPVAVIAIVGAUQTURUSUT $.
+
+  ${
+    $d a b $.
+    $( There exist irrational numbers ` a ` and ` b ` such that ` ( a ^ b ) `
+       is rational.  Statement in the Metamath book, footnote 27 on page 17.
```
benjub, line 353448: Can you add the Section (which is typically more stable than the page):
```
...footnote 27 on page 17 (Section 1.1.5).
```
and similarly for your other theorems.

benjub:
> As mentioned in the latest commit, the new theorem ~sqrt2irr0 (`` ( sqrt `` 2 ) e. ( RR \ QQ ) ``) seems to be a more appropriate formalization of "The square root of 2 is an irrational number. " than the already existing theorem ~sqrt2irr, which only says that "The square root of 2 is not a rational number. ": `` ( sqrt `` 2 ) e/ QQ ``. I would have renamed it (maybe ~sqrt2nq), but this is a Metamath 100 theorem, which is widely referenced. Furthermore, if I rename ~sqrt2irr, and change the label of ~sqrt2irr0 to ~sqrt2irr, the credit for the 1st Metamath 100 theorem would go to me, which is not justified (I just combined two existing theorems).

I would have only one theorem, labeled ~sqrt2irr, but with the statement you propose (with the few steps added at the end of the proof as you propose, removing the nelir/df-nel pair). To do it, `MM-PA> expand sqrt2irr` in your proof.  Then, at the end of the comment of ~sqrt2irr, you may write:
```
Change ` e/ QQ ` to ` e. ( RR \ QQ ) ` .  (Revised by AV, xxxx).
```
**edit:** since this is a "100 challenge" theorem, ping @digama0, and since the same things may be done in iset.mm, ping @jkingdon

digama0:
I think we should not touch the statement nor proof of sqrt2irr, it is of historical significance and even appears verbatim in "the seventeen provers of the world".

avekens:
> I think we should not touch the statement nor proof of sqrt2irr, it is of historical significance and even appears verbatim in "the seventeen provers of the world".

I had the same concerns, therefore I will not change this theorem (including its comment).

By the way, the only proof (besides the proof of ~sqrt2irr0) in which ~sqrt2irr is used is the proof of ~ nthruc, which can be shortened using ~sqrt2irr0.

avekens:
There is still an (other?) error in the verification process...

digama0:
I think for the present we should just ignore metamath.exe CI until the server issue is fixed. (see #2948)

benjub (approved):

set.mm
jkingdon: I don't think the current tools let you word it quite like that, but including the section does make sense to me.

jkingdon:
The situation in iset.mm is quite different, in ways that aren't affected by the existence of `sqrt2irr0`. In iset.mm, the usual way to represent "the square root of two is irrational" is https://us.metamath.org/ileuni/sqrt2irrap.html which is to be read in conjunction with https://us.metamath.org/ileuni/sqrt2re.html

benjub:
> The situation in iset.mm is quite different, in ways that aren't affected by the existence of `sqrt2irr0`. In iset.mm, the usual way to represent "the square root of two is irrational" is https://us.metamath.org/ileuni/sqrt2irrap.html which is to be read in conjunction with https://us.metamath.org/ileuni/sqrt2re.html

You can still pair up sqrtirr an sqrt2re to get sqrt2irr0, as @avekens did in set.mm. But if I understand you, what you mean is that this is not so useful since this will not subsume sqrt2irrap, which is stronger. I think it could still be added, with the mention that sqrt2irrap is stronger.

avekens:
> The situation in iset.mm is quite different, in ways that aren't affected by the existence of `sqrt2irr0`. In iset.mm, the usual way to represent "the square root of two is irrational" is https://us.metamath.org/ileuni/sqrt2irrap.html which is to be read in conjunction with https://us.metamath.org/ileuni/sqrt2re.html

@jkingdon More interesting would be if ~2irrexpq (see Metamath book, section 1.1.5, footnote 27 on page 17) can be proven in iset.mm. To use the proof from set.mm, it must be shown that (sqrt 2) ^ (sqrt  2) e. QQ is decidable. Or the fact that (sqrt 2) ^ (sqrt  2) e. ( RR \ QQ  ) can be used directly (which means that (sqrt 2) ^ (sqrt  2) e. QQ is decidable?), which follows from the Gelfond-Schneider Theorem. Unfortunately, this theorem is not available in set.mm or iset.mm. Is there a simpler way to prove ~2irrexpq in iset.mm? Does the Gelfond-Schneider Theorem holds for intuitionistic logic at all?

jkingdon:
> @jkingdon More interesting would be if ~2irrexpq (see Metamath book, section 1.1.5, footnote 27 on page 17) can be proven in iset.mm.

I've added this as https://github.com/metamath/set.mm/issues/2958 including a cited source which discusses this theorem. That source mentions the idea of showing that `(sqrt 2) ^ (sqrt 2)` is irrational, but it has a simpler proof which I suppose is the way to go.

PR #2957, opened 2022-12-24
Update verifiers.yml

benjub:
Update to the latest versions of `actions`.  See:
https://github.com/actions/checkout
https://github.com/actions/cache
https://github.com/actions/setup-java
https://github.com/actions/setup-python
https://github.com/hecrj/setup-rust-action

The `distribution` field for java became mandatory, see https://github.com/actions/setup-java#supported-distributions

This commit aims to remove some warnings in https://github.com/metamath/set.mm/actions/runs/3736342673, see https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -176,7 +176,7 @@ jobs:
           java-version: '10' # The JDK version to make available on the path.
           # java-package: jdk # (jdk, jre, jdk+fx, or jre+fx) - defaults to jdk
           # architecture: x64 # (x86, x64, armv7, aarch64, or ppc64le) - default value derived from the runner machine
-      - run: wget -N https://github.com/digama0/mmj2/tree/master/mmj2jar
+      - run: wget -N https://github.com/digama0/mmj2/tree/master/mmj2jar/mmj2.jar
       # Do extra checking and generate the '*discouraged.new' file
       # (the 'printf' lets us insert multiple lines).
       # We use setparser to check that definitions don't create
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -61,7 +61,7 @@ jobs:
       - run: ls -l
       - run: g++ --version
       # Get and compile checkmm, a C++ verifier by Eric Schmidt
-      - run: wget -N https://github.com/metamath/metamath-website-seed/blob/main/downloads/checkmm.cpp
+      - run: wget -N https://github.com/metamath/metamath-website-seed/raw/main/downloads/checkmm.cpp
       - run: g++ -O2 -o checkmm checkmm.cpp
       # Use it to verify set.mm and iset.mm
       - run: ./checkmm set.mm
@@ -176,7 +176,7 @@ jobs:
           java-version: '10' # The JDK version to make available on the path.
           # java-package: jdk # (jdk, jre, jdk+fx, or jre+fx) - defaults to jdk
           # architecture: x64 # (x86, x64, armv7, aarch64, or ppc64le) - default value derived from the runner machine
-      - run: wget -N https://github.com/digama0/mmj2/tree/master/mmj2jar/mmj2.jar
+      - run: wget -N https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar
       # Do extra checking and generate the '*discouraged.new' file
       # (the 'printf' lets us insert multiple lines).
       # We use setparser to check that definitions don't create
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -170,10 +170,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       # See: https://github.com/actions/setup-java
-      - uses: actions/setup-java@v3
+      - uses: actions/setup-java@v2
         with:
           distribution: 'zulu' # required field
-          java-version: '11' # The JDK version to make available on the path.
+          java-version: '10' # The JDK version to make available on the path.
           # java-package: jdk # (jdk, jre, jdk+fx, or jre+fx) - defaults to jdk
           # architecture: x64 # (x86, x64, armv7, aarch64, or ppc64le) - default value derived from the runner machine
       - run: wget -N https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -170,13 +170,17 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       # See: https://github.com/actions/setup-java
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v3
         with:
-          # distribution: 'temurin' # or 'zulu' ; required field if using actions/setup-java@v3
-          java-version: '10' # The JDK version to make available on the path.
+          distribution: 'zulu' # ('temurin', 'zulu', ...) - required field
+          java-version: '11' # The JDK version to make available on the path.
           # java-package: jdk # (jdk, jre, jdk+fx, or jre+fx) - defaults to jdk
           # architecture: x64 # (x86, x64, armv7, aarch64, or ppc64le) - default value derived from the runner machine
       - run: wget -N https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar
+      - run: mkdir macros
+      - run: wget -O macros/definitionCheck.js https://github.com/digama0/mmj2/raw/master/mmj2jar/macros/definitionCheck.js
+      - run: wget -O macros/init.js https://github.com/digama0/mmj2/raw/master/mmj2jar/macros/init.js
+      - run: wget -O macros/showDiscouraged.js https://github.com/digama0/mmj2/raw/master/mmj2jar/macros/showDiscouraged.js
       # Do extra checking and generate the '*discouraged.new' file
       # (the 'printf' lets us insert multiple lines).
       # We use setparser to check that definitions don't create
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -61,7 +61,7 @@ jobs:
       - run: ls -l
       - run: g++ --version
       # Get and compile checkmm, a C++ verifier by Eric Schmidt
-      - run: wget -N -q http://us.metamath.org/downloads/checkmm.cpp
+      - run: wget -N https://github.com/metamath/metamath-website-seed/tree/main/downloads/checkmm.cpp
       - run: g++ -O2 -o checkmm checkmm.cpp
       # Use it to verify set.mm and iset.mm
       - run: ./checkmm set.mm
@@ -176,8 +176,7 @@ jobs:
           java-version: '10' # The JDK version to make available on the path.
           # java-package: jdk # (jdk, jre, jdk+fx, or jre+fx) - defaults to jdk
           # architecture: x64 # (x86, x64, armv7, aarch64, or ppc64le) - default value derived from the runner machine
-      - run: wget -N -q http://us.metamath.org/ocat/mmj2/mmj2jar.zip
-      - run: unzip -q mmj2jar.zip
+      - run: wget -N https://github.com/digama0/mmj2/tree/master/mmj2jar
       # Do extra checking and generate the '*discouraged.new' file
       # (the 'printf' lets us insert multiple lines).
       # We use setparser to check that definitions don't create
@@ -199,7 +198,7 @@ jobs:
       - uses: actions/setup-python@v3
         with:
           python-version: '>=3.9'
-      - run: wget -N -q https://raw.githubusercontent.com/david-a-wheeler/mmverify.py/master/mmverify.py
+      - run: wget -N https://raw.githubusercontent.com/david-a-wheeler/mmverify.py/master/mmverify.py
       # We can speed up the script by deleting all logging calls, since we do
       # not need them.
       - run: sed -E '/^ *vprint\(/d' mmverify.py > mmverify.faster.py
@@ -218,7 +217,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '>=3.9'
-      - run: wget -N -q https://raw.githubusercontent.com/david-a-wheeler/mmverify.py/master/mmverify.py
+      - run: wget -N https://raw.githubusercontent.com/david-a-wheeler/mmverify.py/master/mmverify.py
       # We can speed up the script by deleting all logging calls, since we do
       # not need them.
       - run: sed -E '/^ *vprint\(/d' mmverify.py > mmverify.faster.py
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -172,10 +172,10 @@ jobs:
       # See: https://github.com/actions/setup-java
       - uses: actions/setup-java@v3
         with:
-          distribution: 'zulu'
+          distribution: 'zulu' # required field
           java-version: '10' # The JDK version to make available on the path.
-          # java-package: jdk # (jre, jdk, or jdk+fx) - defaults to jdk
-          # architecture: x64 # (x64 or x86) - defaults to x64
+          # java-package: jdk # (jdk, jre, jdk+fx, or jre+fx) - defaults to jdk
+          # architecture: x64 # (x86, x64, armv7, aarch64, or ppc64le) - default value derived from the runner machine
       - run: wget -N -q http://us.metamath.org/ocat/mmj2/mmj2jar.zip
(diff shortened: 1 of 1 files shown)


benjub:
The CI error seems to come from the fact that version 10 of java is not found in the distribution `temurin`. Maybe it is in the distribution`zulu` ? (see https://github.com/actions/setup-java#supported-distributions)
Output of the test at https://github.com/metamath/set.mm/actions/runs/3771520708/jobs/6411959866 :
```
Run actions/setup-java@v3
  with:
    distribution: temurin
    java-version: 10
    java-package: jdk
    check-latest: false
    server-id: github
    server-username: GITHUB_ACTOR
    server-password: GITHUB_TOKEN
    overwrite-settings: true
    job-status: success
    token: ***
Installed distributions
  Trying to resolve the latest version from remote
  Error: Could not find satisfied version for SemVer '10'. 
  Available versions: 19.0.1+10, 19.0.0+36, 18.0.2+101, 18.0.2+9, 18.0.1+10, 18.0.0+36, 17.0.5+8, 17.0.4+101,
 17.0.4+8, 17.0.3+7, 17.0.2+8, 17.0.1+12, 17.0.0+35, 16.0.2+7, 11.0.17+8, 11.0.16+101, 11.0.16+8, 11.0.15+10,
 11.0.14+101, 11.0.14+9, 11.0.13+8, 11.0.12+7, 8.0.352+8, 8.0.345+1, 8.0.342+7, 8.0.332+9, 8.0.322+6,
 8.0.312+7, 8.0.302+8
```

benjub:
Regarding the CI fail: maybe I am not downloading from the correct URL... I am not sure which to choose:
```
      - run: wget -N https://github.com/digama0/mmj2/tree/master/mmj2jar
```
why not `https://github.com/digama0/mmj2/mmj2jar` ?  And
```
      - run: wget -N https://github.com/metamath/metamath-website-seed/tree/main/downloads/checkmm.cpp
```
why not `https://github.com/metamath/metamath-website-seed/downloads/checkmm.cpp`?

Same for
```
      - run: wget -N https://raw.githubusercontent.com/david-a-wheeler/mmverify.py/master/mmverify.py
```
why not `https://github.com/david-a-wheeler/mmverify.py/mmverify.py`?

jkingdon:
> Regarding the CI fail

Well, there is one in https://github.com/metamath/set.mm/actions/runs/3771843743/jobs/6412454173 which I guess has to do with g++ version or something sorta like that (but I assume you aren't talking about that one). The one in https://github.com/metamath/set.mm/actions/runs/3771843743/jobs/6412454060 I guess is the one you are asking about.

> : maybe I am not downloading from the correct URL... I am not sure which to choose:
> 
> ```
>       - run: wget -N https://github.com/digama0/mmj2/tree/master/mmj2jar
> ```
> 
> why not `https://github.com/digama0/mmj2/mmj2jar` ?

I don't think the latter exists. If you are asking why github designed the URLs the way they did, *shrug* but we can't change that.

As for the problem `Unable to access jarfile mmj2.jar`, I think the old script, when it ran `unzip -q mmj2jar.zip`, created the files in the current directory and the replacement, which just fetches a directory, creates them in a subdirectory `mmj2jar`. It would appear that we could just fetch `https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar` rather than the whole directory, as I didn't notice anything using any of the other files.

benjub:
Changed `tree` to `blob` in
```
      - run: wget -N https://github.com/metamath/metamath-website-seed/tree/main/downloads/checkmm.cpp
```
but it does not seem to work better. All these URLs look a bit mysterious to me.

So for the last two errors, probably coming from a wrong URL (I don't understand the directory structure of github, with the `tree`, `blob`, `raw`, or `master` in the path), I can only ask for help from @david-a-wheeler or @digama0.

After that, the only thing still downloaded from https://us.metamath.org will be the GIF symbols, for which we can find a workaround (and the best option is probably to drop GIF support altogether and replace it with an ASCII version).

jkingdon:
> All these URLs look a bit mysterious to me.
> . . . I don't understand the directory structure of github, with the `tree`, `blob`, `raw`, or `master` in the path),

I don't really "understand" it, in the sense of trying to construct URLs myself. I click around github in a web browser and look for links like "download" or "raw" (that is, I want just the contents of the file, not an HTML page which talks about the file - some of your attempts failed because you were getting HTML instead of the contents of the file). In firefox a right click on a link shows a menu and "copy link" is one of the choices (and other web browsers should be similar).

Anyway, my results are at https://github.com/jkingdon/set.mm/actions/runs/3788570201/jobs/6441487702 . By using the URL `https://github.com/metamath/metamath-website-seed/raw/main/downloads/checkmm.cpp`  I got the checkmm build to pass and by using `https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar` I got a real error rather than a problem in downloading:

> java.lang.UnsupportedClassVersionError: mmj/util/BatchFramework has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 54.0

Maybe I can kick this one back to you - do you know what version of java we should be using? 54 versus 55 looks like they correspond to java 10 and java 11 according to https://stackoverflow.com/questions/9170832/list-of-java-class-file-format-major-version-numbers

> I can only ask for help from @david-a-wheeler or @digama0.

If they want to help, that's great, although I'm not sure it has to be them in particular for most of this.

> After that, the only thing still downloaded from https://us.metamath.org will be the GIF symbols, for which we can find a workaround (and the best option is probably to drop GIF support altogether and replace it with an ASCII version).

Probably (or if there is a way to avoid generating two copies of the site that would save a bunch of disk space). But that can be for a future pull request.

benjub:
Thanks @jkingdon I changed the URLs as you indicated (I hope they are the URLs that stay in sync with the most "up-to-date" versions).
As for the java version, verifiers.yml specifies `java-version: '10'` since mmj2 uses deprecated functions and has to be compiled with older versions.

**edit:** the error seems to be not related to the java version, since it outputs:
```
Run echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
  echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Zulu_jdk/10.0.2-13/x64
    JAVA_HOME_10_X64: /opt/hostedtoolcache/Java_Zulu_jdk/10.0.2-13/x64
Checking discouraged file:
diff: discouraged.new: No such file or directory
```

benjub:
I tried reverting to actions/setup-java@v1 (or setup-java@v2) and I tried using java versions 10 and 11. Nothing works.  I think that the problem is that the version of mmj2.jar downloaded from github was compiled using version 11, and for some reason other parts of mmj2 compile with version 10 but not version 11 ??

jkingdon:
> Thanks @jkingdon I changed the URLs as you indicated (I hope they are the URLs that stay in sync with the most "up-to-date" versions).

Should be. The `main` or `master` in the URL is the name of the branch (I'm pretty sure) so it would be whatever is on that branch as of when the job runs (this does mean that the tests can go red without anything changing in the set.mm repository, but that's probably a pretty good choice of tradeoffs for our situation).

> **edit:** the error seems to be not related to the java version, since it outputs:
> 
> ```
> Run echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
>   echo 'Checking discouraged file:' && diff -U 0 discouraged discouraged.new
>   shell: /usr/bin/bash -e {0}
>   env:
>     JAVA_HOME: /opt/hostedtoolcache/Java_Zulu_jdk/10.0.2-13/x64
>     JAVA_HOME_10_X64: /opt/hostedtoolcache/Java_Zulu_jdk/10.0.2-13/x64
> Checking discouraged file:
> diff: discouraged.new: No such file or directory
> Error: Process completed with exit code 2.
> ```

Ah, you hit one of the funkiest things about how errors are reported. It turns out that `discouraged.new` being missing is not the root cause, but is a result of mmj2 not successfully running a few lines up. If you go to (for example) https://github.com/metamath/set.mm/actions/runs/3789173550/jobs/6442685604 and click on either of the two lines which start `Run scripts/verify-mmj2` you will see the java error which I posted a few comments ago. (And this is indeed the step which is supposed to write `discouraged.new`.)

jkingdon:
> I tried reverting to actions/setup-java@v1 (or @v2) and I tried using java versions 10 and 11. Nothing works. I think that the problem is that the version of mmj2.jar downloaded from github was compiled using version 11, and for some reason other parts of mmj2 compile with version 10 but not version 11 ??

It looks like the jarfile that we are getting from github does indeed require java 11 but the jarfile we had been getting from metamath.org only requires java 8 or later.

As for how I know this, hang onto your hat as this gets a bit deep into aspects of java that we don't usually interact with directly. Here's what I ran from the command line locally:

```
wget -N https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar
jar -x --file mmj2.jar
find mmj -print | xargs file
```

(see many lines containing `compiled Java class data, version 55.0 (Java SE 11)`)

```
mkdir web
cd web
wget -N http://us.metamath.org/ocat/mmj2/mmj2jar.zip
unzip mmj2jar.zip
jar -x --file mmj2.jar
find mmj -print | xargs file
```

(see only lines containing `compiled Java class data, version 52.0 (Java 1.8)`)

benjub:
But when downloading `https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar`and running java version 11, we get  https://github.com/metamath/set.mm/actions/runs/3789122649/jobs/6442583803 (three commits above)... Actually, it says ` java.io.FileNotFoundException: definitionCheck.js` but the file is at `https://github.com/digama0/mmj2/raw/master/mmj2jar/macros/definitionCheck.js`, so we should download not only the jar file but also other files... Downloading the directory, you wrote above, creates a subdirectory, so either we can ask wget` to download without creating a subdirectory, or we adapt the commands following it to the new subdirectory (or we download only the necessary files, i.e., mmj2.jar, definitionCheck.js, etc... but we need an exhaustive list).

jkingdon:
> But when downloading `https://github.com/digama0/mmj2/raw/master/mmj2jar/mmj2.jar`and running java version 11, we get https://github.com/metamath/set.mm/actions/runs/3789122649/jobs/6442583803 (three commits above)... Actually, it says ` java.io.FileNotFoundException: definitionCheck.js` but the file is at `https://github.com/digama0/mmj2/raw/master/mmj2jar/macros/definitionCheck.js`, so we should download not only the jar file but also other files... Downloading the directory, you wrote above, creates a subdirectory, so either we can ask wget` to download without creating a subdirectory, or we adapt the commands following it to the new subdirectory (or we download only the necessary files, i.e., mmj2.jar, definitionCheck.js, etc... but we need an exhaustive list).

Yup, that's right. I'm working on trying out `download only the necessary files, i.e., mmj2.jar, definitionCheck.js, etc... but we need an exhaustive list` but I'm not wedded to that if you think one of the other choices above is a better way to fix it.

benjub:
> Yup, that's right. I'm working on trying out `download only the necessary files, i.e., mmj2.jar, definitionCheck.js, etc... but we need an exhaustive list` but I'm not wedded to that if you think one of the other choices above is a better way to fix it.

I guess that depends on the number of necessary files ! Please try and see which are needed. Unfortunately, I am not on my main computer (Christmas time...) and I cannot test different solutions.
(I looked at wget options, e.g., https://stackoverflow.com/questions/5043239/how-do-i-mirror-a-directory-with-wget-without-creating-parent-directories and https://stackoverflow.com/questions/11507198/wget-download-a-sub-directory, but the options `-nd` and `--cut-dirs`, which seem to remove the nesting, seem to completely flatten the subdirectories as well.)

digama0:
I think it suffices to have the jar file, the macros folder, and the RunParms.txt file (well, this one you probably want to write on the spot in any case since it determines what mmj2 will actually check / do).

jkingdon:
> I guess that depends on the number of necessary files ! Please try and see which are needed. Unfortunately, I am not on my main computer (Christmas time...) and I cannot test different solutions.

Turns out the magic number was three. I got a successful build at https://github.com/jkingdon/set.mm/actions/runs/3790021667/jobs/6444306204

> I looked at wget options

I suspect what we really should be doing here is getting the files using `git` rather than `wget`. But if we are willing to go with my working wget based approach from https://github.com/jkingdon/set.mm/commit/11afe7e0ebd0faf57352de83226b9093a55f3ba8 (that's the revision of the successful build above) then we can worry about whether to switch to something fancier later.

jkingdon:
> I think it suffices to have the jar file, the macros folder, and the RunParms.txt file (well, this one you probably want to write on the spot in any case since it determines what mmj2 will actually check / do).

Yup! This is already going via https://github.com/metamath/set.mm/blob/develop/scripts/verify-mmj2 which writes its own runparms file. And my successful run was indeed the jar file and those files from the macros folder which the particular usage here seemed to need.

benjub:
Thanks @jkingdon and @digama0.  Indeed, downloading with git may be more stable and future-compatible, but I prefer that someone more qualified than me would do it.
This PR looks ready to merge.

jkingdon:
> This PR looks ready to merge.

Yay! Thanks for all your work on it.

benjub:
To sum up:
* what we did in this PR (thanks to @jkingdon and @digama0):
  * remove deprecation warnings from the CI by updating `actions/*@v?` to the latest versions, and update from java version 10 to 11 (a LTS version), specifying the java distribution `'zulu'`;
  * reduce dependence on https://us.metamath.org by downloading checkmm.ccp and mmj2.jar (together with the three needed `.js` macro files) from github;
* what is left to do:
  * remove remaining dependencies on https://us.metamath.org ; apparently, there are only two, both in @david-a-wheeler's `scripts/download-metamath`, namely, `wget -N -q http://us.metamath.org/mpegif/mmbiblio.html`and `wget -N -q http://us.metamath.org/downloads/symbols.tar.bz2`, which could be stored somewhere in the github repository `metamath/website`.  I'll let @david-a-wheeler see what can be done;
  * maybe do all the downloads using `git` rather than `wget`, if this is more stable or future-compatible (I am not sure).

david-a-wheeler:
Switching to downloading from GitHub is a sensible move.

Downloading using git in this case is probably a disadvantage. Git is great for downloading history, but if you don't need any history, wget or curl do the job in a more straightforward way.

benjub:
> Switching to downloading from GitHub is a sensible move.

Since you have the access rights to the github repository `metamath/website`, can you host/store a copy of these two files there ? (then, there is of course the question of keeping the files synchronized when they get modified, and you probably know better than me the preferred way to handle that; maybe putting biblio.html in `metamath/set.mm` is better, or even using the already present `mmbiblio.raw.html`...).

jkingdon:
> Git is great for downloading history, but if you don't need any history, wget or curl do the job in a more straightforward way

wget seems to be straightforward for downloading one file at a time (which apparently is all we need for now), so it should be fine. If we want to get a whole directory tree without having the script have a list of files, I didn't find a straightforward way to do it with wget but maybe someone will come up with something I missed. That is, if there's even a reason to bother.

digama0:
it's fine in this case to explicitly name the macros that are used - you don't even need everything in the folder like I said, since macros call each other explicitly by name so I think you only need init.js, prep.js, and definitionCheck.js if you want to run the definition checker, showDiscouraged.js if you want to regenerate the discouraged file, and possibly transformations.js although that one is mainly for proof assistantry. Most of the others are experimental or for testing purposes.

david-a-wheeler:
I've created a new Metamath repository *specifically* for symbols, and created a new pull request to bring the symbols in from there:

https://github.com/metamath/set.mm/pull/2964

We can separately drop using GIFs/PNGs, but I think putting the symbols in a separate repo is the right first step no matter what.

That will leave <http://us.metamath.org/mpegif/mmbiblio.html> as the only cyclic dependency (one we need to break).

benjub:
> That will leave http://us.metamath.org/mpegif/mmbiblio.html as the only cyclic dependency (one we need to break).

Is there a difference between mpegif/mmbiblio.html and mpeuni/mmbiblio.html ? Can we use https://github.com/metamath/set.mm/blob/develop/mmbiblio.raw.html ?

david-a-wheeler:
> Is there a difference between mpegif/mmbiblio.html and mpeuni/mmbiblio.html ? Can we use https://github.com/metamath/set.mm/blob/develop/mmbiblio.raw.html ?

We can probably use it, but I haven't checked. I suspect it'd be easier to do this a step at a time. Let's first get "symbols" under control, then deal with the rest.

digama0:
mmbiblio.raw.html was definitely intended (and tested) to function as a way to break the dependency of mmbiblio.html on itself. But it hasn't been hooked up to the website script yet.

PR #2959, opened 2022-12-27
More intuitionizing of complex exponentiation and trigonometry from efgt1p to sinmul

jkingdon:
This section is mostly quite straightforward - proofs need either no change, or need to change not equal to apart (especially in the context of division).

A few theorems (especially `tanadd`) needed to have apartness a bit better developed to easily prove the apartness results they need. This is what led to `subap0`, `apdivmuld`, 

Includes proof of `eqord1` and `eqord2` (based on existing theorems). This proof doesn't carry over in an obvious way to `ltord1` , `leord1` , `ltord2` , or `leord2`. Therefore, this pull request does not prove `eflt` or `efle`.

There are a set of theorems about hyperbolic trigonometry: sinhval , coshval , resinhcl , rpcoshcl , recoshcl ,retanhcl , tanhlt1 , and tanhbnd which are not proved here. Seems like they'd probably be easy when someone does want to take them up.

Add a note to iset.mm about the relationship between equality and apartness when intuitionizing a proof which starts with equality and uses that to derive not equal.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8001,6 +8001,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>retancl</TD>
+  <TD>~ retanclap</TD>
+</TR>
+
+<TR>
   <TD>qnnen</TD>
   <TD><I>none</I></TD>
   <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7996,6 +7996,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>tanval3</TD>
+  <TD>~ tanval3ap</TD>
+</TR>
+
+<TR>
   <TD>qnnen</TD>
   <TD><I>none</I></TD>
   <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7975,6 +7975,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>efle</TD>
+  <TD>~ efler</TD>
+  <TD>The set.mm proof of the converse relies on eflt</TD>
+</TR>
+
+<TR>
   <TD>qnnen</TD>
   <TD><I>none</I></TD>
   <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -995,6 +995,16 @@ When working with real and complex numbers, apartness is almost always what
 you want. See ~ df-ap for more on apartness.
 </LI>
 
+<LI>
+Given a theorem of the form ` X = Y <-> Z = W ` we can derive
+` X =/= Y <-> Z =/= W ` but in many contexts what we really want is
+` X =//= Y <-> Z =//= W ` .  See if the version with apartness already
+exists and if not consider adding it (building from basic apartness
+theorems like ~ apadd1 and ~ apmul1 for example).  Once you have proved
+the version using apartness you can use it to prove the version with
+equality if you don't already have it, using ~ notbid and ~ apti .
+</LI>
+
 <LI>Exclusive or ( ` ph \/_ ps ` ) is equivalent to ` ph <-> -. ps `
 given excluded middle but we just have one direction ( ~ xorbin ).
 Consider intuitionizing ` ph <-> -. ps ` as ` ph \/_ ps `
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8011,6 +8011,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>sinhval , coshval , resinhcl , rpcoshcl , recoshcl ,
+  retanhcl , tanhlt1 , tanhbnd</TD>
+  <TD><I>none yet</I></TD>
+  <TD>should be provable</TD>
+</TR>
+
+<TR>
   <TD>qnnen</TD>
   <TD><I>none</I></TD>
   <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8006,6 +8006,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>tanneg</TD>
+  <TD>~ tannegap</TD>
+</TR>
+
+<TR>
   <TD>qnnen</TD>
   <TD><I>none</I></TD>
   <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7981,6 +7981,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>tanval</TD>
+  <TD>~ tanvalap</TD>
+</TR>
+
+<TR>
   <TD>qnnen</TD>
   <TD><I>none</I></TD>
   <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7986,6 +7986,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>tancl</TD>
+  <TD>~ tanclap</TD>
+</TR>
+
+<TR>
   <TD>qnnen</TD>
   <TD><I>none</I></TD>
   <TD>Corollary 8.1.23 of [AczelRathjen] and thus presumably provable.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7986,8 +7986,8 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
-  <TD>tancl</TD>
-  <TD>~ tanclap</TD>
+  <TD>tancl , tancld</TD>
+  <TD>~ tanclap , ~ tanclapd</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8001,8 +8001,8 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
-  <TD>retancl</TD>
-  <TD>~ retanclap</TD>
+  <TD>retancl , retancld</TD>
+  <TD>~ retanclap , ~ retanclapd</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7991,6 +7991,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -81880,6 +81880,53 @@ Ordering on reals (cont.)
...
+        ( clt wbr wi wceq cv wral wcel wa ralrimivva breq1 breq1d imbi12d breq2
+        eqeq1 eqeq1d chvarv breq2d rspc2v mpan9 ) ABUAZCUAZQRZDEQRZSZCHUBBHUBFH
+        UCGHUCUDFGQRZIJQRZSZAUTBCHHPUEUTVCFUQQRZIEQRZSBCFGHHUPFTZURVDUSVEUPFUQQ
+        UFVFDIEQLUGUHUQGTZVDVAVEVBUQGFQUIVGEJIQUPGTZDJTZSVGEJTZSBCUPUQTZVHVGVIV
+        JUPUQGUJVKDEJKUKUHMULUMUHUNUO $.
+
+      $( Infer an ordering relation from a proof in only one direction.
+         (Contributed by Mario Carneiro, 14-Jun-2014.)  (Revised by Jim
+         Kingdon, 20-Dec-2022.) $)
+      eqord1 $p |- ( ( ph /\ ( C e. S /\ D e. S ) ) -> ( C = D <-> M = N ) ) $=
```
benjub, line 81905: I don't understand the comment. As I understand the statement, you have a real function on S, which is $x \mapsto A(x)$, which is strictly increasing (by hypotheses 1 and 6). Its value at C (resp. D) is M (resp. N) by hypotheses 1 and 2 (resp. 1 and 3). So the theorem is:
> a real strictly increasing function on a subset of $\R$ is injective

(?)

iset.mm
```diff
@@ -81880,6 +81880,53 @@ Ordering on reals (cont.)
...
+        HMUEUIULUMVPIJUNVLZUOZJIUNVLZUOZRFGUNVLZUOZGFUNVLZUOZRVRVQVPWGWKWIWMVPW
+        JWFABCDEFGHIJKLMNOPUPUQAVOVNWIWMURAVOVNRRWLWHABCDEGFHJIKMLNOPUPUQUSUTVP
+        IJAVNITQZVOADTQZBHVBZVNWNAWOBHOVAZWOWNBFHWADITLVCVDVEVFAVOJTQZVNAWPVOWR
+        WQWOWRBGHWEDJTMVCVDVEVGVHVPFGVPHTFNWCVIVPHTGNAVNVOVJVIVHVKVM $.
+    $}
+
+    ltord2.6 $e |- ( ( ph /\ ( x e. S /\ y e. S ) ) -> ( x < y -> B < A ) ) $.
+    $( Infer an ordering relation from a proof in only one direction.
+       (Contributed by Mario Carneiro, 14-Jun-2014.) $)
+    eqord2 $p |- ( ( ph /\ ( C e. S /\ D e. S ) ) -> ( C = D <-> M = N ) ) $=
```
benjub, line 81920: Same as above with "strictly decreasing" ?

iset.mm
```diff
@@ -117423,8 +117486,619 @@ Infinite sums (cont.)
       SIFYGYROUGABWSSYBVSVTYMYRYPSMRZAYQSYPMWBQYMYSASMRAYMYPASMAWCWDAWETWFTWHVB
       TWNYMXGXDOYOYMXGXCGKLZMRZXDYMGIFZXGUUAOWGABWSGYBVSVTYTGXCMWIQWJWHVBTWNABW
       SGYBWNWKUUBWNWGVDWLWM $.
+
+    $d A n $.
+    $( The exponential function of a positive real number is greater than 1
```
benjub, line 117491: just do a global search "exponential function of"

iset.mm
```diff
@@ -117423,8 +117486,619 @@ Infinite sums (cont.)
...
+       30-Apr-2014.) $)
+    efgt1 $p |- ( A e. RR+ -> 1 < ( exp ` A ) ) $=
+      ( crp wcel c1 caddc co ce cfv 1red cr 1re readdcl sylancr reefcld clt wbr
+      rpre ltaddrp mpan efgt1p lttrd ) ABCZDDAEFZAGHUBIUBDJCZAJCUCJCKAQZDALMUBA
+      UENUDUBDUCOPKDARSATUA $.
   $}
 
+  ${
+    $d x y A $.  $d x y B $.
+    $( The exponential function on the reals is strictly monotonic.
```
benjub, line 117518: strictly increasing

iset.mm
```diff
@@ -117423,8 +117486,619 @@ Infinite sums (cont.)
...
+                 ( A < B -> ( exp ` A ) < ( exp ` B ) ) ) $=
+      ( cr wcel clt wbr ce cfv w3a cmin co cmul c1 crp simp2 cc0 syl reefcld cc
+      recnd simp1 resubcld biimp3a elrpd efgt1 wb efgt0 ltmulgt11 syl3anc mpbid
+      posdif caddc wceq efadd syl2anc pncan3d fveq2d eqtr3d breqtrd 3expia ) AC
+      DZBCDZABEFZAGHZBGHZEFVAVBVCIZVDVDBAJKZGHZLKZVEEVFMVHEFZVDVIEFZVFVGNDVJVFV
+      GVFBAVAVBVCOZVAVBVCUAZUBZVAVBVCPVGEFABUKUCUDVGUEQVFVDCDVHCDPVDEFZVJVKUFVF
+      AVMRVFVGVNRVFVAVOVMAUGQVDVHUHUIUJVFAVGULKZGHZVIVEVFASDVGSDVQVIUMVFAVMTZVF
+      VGVNTAVGUNUOVFVPBGVFABVRVFBVLTUPUQURUSUT $.
+
+    $( The exponential function on the reals is strictly monotonic.
```
benjub, line 117531: is increasing
(or, depending on (i)set.mm's conventions, "nondecreasing")

iset.mm
```diff
@@ -117423,8 +117486,619 @@ Infinite sums (cont.)
...
+  $( Domain and codomain of the sine function.  (Contributed by Paul Chapman,
+     22-Oct-2007.)  (Revised by Mario Carneiro, 30-Apr-2014.) $)
+  sinf $p |- sin : CC --> CC $=
+    ( vx cc ci cv cmul co ce cfv cneg cmin c2 cdiv csin df-sin wcel ax-icn mpan
+    mulcl efcl syl negicn subcld cc0 cap 2mulicn 2muliap0 divclap mp3an23 fmpti
+    wbr ) ABBCADZEFZGHZCIZUKEFZGHZJFZKCEFZLFZMANUKBOZUQBOZUSBOZUTUMUPUTULBOZUMB
+    OCBOUTVCPCUKRQULSTUTUOBOZUPBOUNBOUTVDUAUNUKRQUOSTUBVAURBOURUCUDUJVBUEUFUQUR
+    UGUHTUI $.
+
+  $( Domain and codomain of the sine function.  (Contributed by Paul Chapman,
```
benjub, line 117620: cosine

iset.mm
```diff
@@ -117423,8 +117486,619 @@ Infinite sums (cont.)
...
+    ZVAVAMHZJFZKLFZANHZVAUAHZUSVEVHKLUSVDVGVAJUSUTMHZGHZVDVGUSVLVCGUSVLDMHZAMHZ
+    EFZVCUSDRCZARCZVLVPOPAUBZDAUCQUSVPVBVOEFVCVNVBVOEUDUEUSVOAVBEAUFSUGUHUIUSUT
+    RCZVMVGOUSVQVRVTPVSDAUJQZUTUKTULSUMUSVRVJVFOVSAUNTUSVTVARCVKVIOWAUTUOVAUPUQ
+    UR $.
+
+  ${
+    $d A k n $.  $d F k $.
+    efi4p.1 $e |- F = ( n e. NN0 |-> ( ( ( _i x. A ) ^ n ) / ( ! ` n ) ) ) $.
+    $( Separate out the first four terms of the infinite series expansion of
+       the exponential function of an imaginary number.  (Contributed by Paul
```
benjub, line 117747: remove "of an imaginary number" (not needed, and the statement is made for an arbitrary complex number anyway)

iset.mm
jkingdon: I think that's a better way to state it, but perhaps @digama0 who wrote the comment would see something I'm not seeing (either about whether that is accurate or about whether it emphasizes the wrong thing).

I'm not sure I'd design any of this exactly how `eqord1` does it (with so many hypotheses and classes), but then again I'm not sure what my alternative would be and anyway I'm usually in the business of intuitionizing things like this rather than revamping them.

iset.mm
jkingdon: Changed in both set.mm and iset.mm.

benjub:
If I click on "Compare" next to "jkingdon force-pushed the `efgt1p` branch from ... to ...", the webpage does not display the changes (it displays "empty file", even though it says "Showing 4 changed files with 37 additions and 23 deletions").  I'm not sure if it's a problem from github or from my old unsupported browser or something else. I'll be back to my main computer tomorrow and maybe I'll understand better.

iset.mm
jkingdon: Changed in both set.mm and iset.mm.

iset.mm
jkingdon: Fixed in set.mm and iset.mm.  Funny that this one went unfixed for (apparently 15) years.

iset.mm
jkingdon: ``( exp ` ( _i x. A ) )`` is what is being expanded in `efi4p`, so it is pure imaginary rather than arbitrary complex. The arbitrary complex one is `ef4p`.

To my ears "separate out" is a natural way to express the idea of listing the first few terms explicitly, then a summation expression for the rest.

benjub: Yes, it's `( _i x. A )`, but the line reads ``efi4p $p |- ( A e. CC -> ( exp ` ( _i x. A ) ) =`` so it's any complex number.

iset.mm
jkingdon: I'm still not sure I'm sold on the idea that this is my job. But I did try making a pass through these (it wasn't always clear to me what to change), in set.mm and iset.mm.

benjub: Well, it's no one's job, but we can try and fix the typos we run into.  Thanks for that pass.

iset.mm
jkingdon: Hmm, that's a bit of a tough one. Although `A` can be any complex number, the normal use case would be where it is real, and in fact both usages currently in set.mm supply `A` which is always real. I guess removing `of an imaginary number` is fine.  I've pushed up a change to do that in set.mm and iset.mm.

iset.mm
benjub: Looks like this (and the similar next) are the only two remaining points. Since you and I agree, if there is no objection from @digama0, I think you can adopt the wording I proposed.

jkingdon: I've pushed up a change to use what is basically BJ's wording, for both set.mm and iset.mm. If we want to revise further we can always do so later.

iset.mm
jkingdon: I've pushed up a change to use what is basically BJ's wording, for both set.mm and iset.mm. If we want to revise further we can always do so later.

benjub (approved):

jkingdon:
> If I click on "Compare" next to "jkingdon force-pushed the `efgt1p` branch from ... to ...", the webpage does not display the changes (it displays "empty file", even though it says "Showing 4 changed files with 37 additions and 23 deletions"). I'm not sure if it's a problem from github or from my old unsupported browser or something else. I'll be back to my main computer tomorrow and maybe I'll understand better.

I was seeing the same (at least for that particular button).

I'm not sure exactly what github does but I do know it can get a bit odd when branches have been rebased since the pull request was opened (as this one has here).

PR #2960, opened 2022-12-27
remove ax-2 from notnotri

icecream17:
not not phi
not phi --> not psi
psi --> phi

And "psi" is replaced with "not not phi"


set.mm
```diff
@@ -1720,7 +1720,16 @@ In classical logic (our logic) this is always true.  In intuitionistic
 
   ${
     notnotri.1 $e |- -. -. ph $.
-    $( Inference associated with ~ notnotr .
+    $( Inference associated with ~ notnotr .  For a shorter proof using
+       ~ ax-2 , see ~ notnotriALT .
+
```
benjub, line 1725: no blank line

benjub (approved):

PR #2961, opened 2022-12-28
shorten eu6

wlammen:
* shorten eu6
* delete outdated OLD theorem

@avekens Do you still need the prototype sumvtxdg2size in line 401972?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15467,6 +15467,7 @@ New usage of "erngplus2-rN" is discouraged (0 uses).
 New usage of "erngring-rN" is discouraged (0 uses).
 New usage of "erngset-rN" is discouraged (3 uses).
 New usage of "estrresOLD" is discouraged (0 uses).
+New usage of "eu6OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
I cannot see the extent of the shortening on github, but I'm looking forward to seeing it on the webpages !

benjub (approved):

wlammen:
@benjub I hope you are not disappointed, because the proof structure is basically the same.  3 modifications were made:
* bitri* were packed into 3bitri* on two occasions;
* A smarter use of dfbi2 replaces an ancom with long sub-expressions with a shorter bicom;
* An equcoms replaces a bicomd.

avekens:
> @avekens Do you still need the prototype sumvtxdg2size in line 401972?

I think ~sumvtxdg2size makes no sense, at least not in the current form (if the graph is not finite or does not have finite size, 
``sum_ v e. V ( D ` v ) = +oo = ( 2 x. ( # ` I ) )`` must be shown, which is not possible with the definitions for `x.` and `sum_`). 

I will delete this prototype with my next PR.

Issue #2955, opened 2022-12-23
Shorter Proofs for Theorem brcnvg and Theorem opelcnvg


walkingsophie:
Currently the proof for brcnvg uses opelcnvg, meanwhile opelcnvg already has the proof for brcnvg in it. 
Changing the proof of brcnvg to be the first 4 steps of opelcnvg, and making opelcnvg use brcnvg would make the proofs overall shorter.

![image](https://user-images.githubusercontent.com/43860774/209333631-b5e7bb56-00d3-4ec3-a870-828ada80975e.png)
![image](https://user-images.githubusercontent.com/43860774/209333713-ce4d3bf3-e3c5-4295-85e2-37479275beb9.png)

benjub:
Good catch !  Can you do it ?

walkingsophie:
Unfortunately I'm having issues with mmj2 detecting the correct version of Java, so I can't do it now.

A similar thing also happens with theorem ordon and epweon. Perhaps someone could write a script to find all instances of a theorem being proven in the proof of another theorem.

benjub:
* mmj2: yes, that has been a hassle for me too. You have to downgrade to java version 11 to have it compile (it does not compile with later versions). For shorter proofs (say below 15 essential steps), I prefer to use metamath.c.
* I'll do the changes for the theorems you mention.
* @david-a-wheeler has a script that outputs all the intermediate steps proved in an entire database. He used it to look at the largest occurrences to deduce some useful lemmas to extract and prove. But one may also look in that list formulas which appear both as an intermediate step and as a final step. I'm not sure his script is online. Also, I seem to remember it was checking only exact equality of formulas, not up to variable renaming, let alone instances.
* Another possibility would be to add an option to the metamath.c command `MM-PA> MINIMIZE_WITH *`: currently, it tries to minimize the current proof using all previous theorems; it could be changed to try and minimize using all theorems which do not rely on the current theorem (to avoid circular proofs), an information which is given by `MM> SHOW TRACE_BACK <label-match>`.

PR #2962, opened 2022-12-29
Two minor shortenings

benjub:
* prove opelcnvg from brcnvg (which was an intermediate step in opelcnvg)

* prove ordon from epweon (which was an intermediate step in ordon)

as noticed in #2955

**edit**: the shortenings are so minor there is no need IMO for revision tags.


set.mm
```diff
@@ -67281,22 +67283,21 @@ associative structure (such as a group).  (Contributed by NM,
...
-      UJPUKPUPUQQUJUKOZUMUKUJOZRUOUJUKUAULURUMUMUNUSBUJSUMUHAUKSUBUCUDUEABCFUFT
-      CUGT $.
-  $}
-
-  $( The epsilon relation well-orders the class of ordinal numbers.
-     Proposition 4.8(g) of [Mendelson] p. 244.  (Contributed by NM,
-     1-Nov-2003.) $)
-  epweon $p |- _E We On $=
-    ( con0 word cep wwe ordon ordwe ax-mp ) ABACDEAFG $.
+    $( The membership relation well-orders the class of ordinal numbers.
```
avekens, line 67286: Is it really wise to replace "epsilon relation" by "membership relation" if `_E` is involved? Both terms are synonymous, and "membership relation" is also used if ony `e.` is involved.

benjub: To me, it is. The term "membership relation" is self-descriptive, while "epsilon relation" is slightly obscure and outdated.
I modified a while ago the comment of https://us.metamath.org/mpeuni/df-eprel.html to mention both terms with a short explanation for the use of the term "epsilon relation" (the use of the term "membership relation", on the other hand, does not require any explanation since, again, it is self-descriptive). To me, that comment could be the sole place where the term "epsilon relation" appears. I also modified around the same time the comments of epelg/epeli/epel to use the precise terms "membership relation" and "membership predicate".

jkingdon: I like "membership relation" and "membership predicate".

set.mm
```diff
@@ -67281,22 +67283,21 @@ associative structure (such as a group).  (Contributed by NM,
...
+       1-Nov-2003.) $)
+    epweon $p |- _E We On $=
+      ( vx vy con0 cep wwe wfr cv wbr weq w3o wral onfr wcel eloni wa ordtri3or
+      word wel epel biid 3orbi123i sylibr syl2an rgen2a dfwe2 mpbir2an ) CDECDF
+      AGZBGZDHZABIZUHUGDHZJZBCKACKLULABCUGCMUGQZUHQZULUHCMUGNUHNUMUNOABRZUJBARZ
+      JULUGUHPUIUOUJUJUKUPBUGSUJTAUHSUAUBUCUDABCDUEUF $.
+  $}
+
+  $( The class of all ordinal numbers is ordinal.  Proposition 7.12 of
+     [TakeutiZaring] p. 38, but without using the Axiom of Regularity.
```
jkingdon, line 67297: I confirmed that the non-use of the axiom of regularity is indeed maintained.

jkingdon (approved):

avekens (approved):

Issue #2958, opened 2022-12-26
Prove 2irrexpq in iset.mm


jkingdon:
This is a famous theorem which can be proved using excluded middle, for example the proof at https://us.metamath.org/mpeuni/2irrexpq.html

It is Theorem 1.2 of [[Bauer],](https://us.metamath.org/ileuni/mmil.html#Bauer) p. 483. The "Classical proof" given there is the one in set.mm

The "Constructive proof" in Bauer is "(√ 2 )<sup>log<sub>2</sub>9</sup> = √(2<sup>log<sub>2</sub>9</sup>) = √9 = 3"  plus "Of course, irrationality of log<sub>2</sub>9 needs to be established, but the proof is no harder than the one of irrationality of √2." (_update 30 Dec 2022: added missing radical symbol_)

Note that if we take `2irrexpq` as stated in set.mm and copy that to iset.mm, the resulting statement would require `a` and `b` to not be rational. We'd presumably also want to strengthen that condition and require that `a` and `b` are irrational, in the sense of apart from any rational number, as in https://us.metamath.org/ileuni/sqrt2irrap.html .

Of course we'd need https://us.metamath.org/mpeuni/df-logb.html and the related theorems (or at least a version for real arguments). Likewise to even state the desired result we need https://us.metamath.org/mpeuni/df-cxp.html (where again, we only need this to work for real arguments).

benjub:
You can probably skip the definition of `logb`.  The proof that log(3)/log(2) is irrational looks accessible since you have already proved Euclid's lemma. Good luck !

avekens:
> You can probably skip the definition of `logb`. The proof that log(3)/log(2) is irrational looks accessible since you have already proved Euclid's lemma. Good luck !

I think there is a "root" symbol missing in the constructive proof of Bauer. It should read: `(√2)^log_2(9) = √(2^log_2(9)) = √9 = 3` (unfortunatley, I do not know how to obtain sub-/superscripts in this editor).

How should the proof work with `log_2(3)`? We would have `(√2)^log_2(3) = √(2^log_2(3)) = √3` which is not rational.

benjub:
If (log 3)/(log 2) is irrational, then so is (log 9)/(log 2). What I mean is that as a standalone result, irrationality of (log 3)/(log 2) seems neater.

jkingdon:
> I think there is a \"root\" symbol missing in the constructive proof of Bauer.

[...]

PR #2963, opened 2022-12-30
Exponentiation with two irrationals

avekens:
* auxiliary theorems: ~elpq, ~cxpcom, ~logbgt0b, ~logbgcd1irr, ~2logb9irr, ~sqrt2cxp2logb9e3
* Alternate proof of ~ 2irrexpq, usable in intuitionistic logic (see also issue #2958) (many thanks to JK for his hint)
* comment of ~2irrexpq enhanced
* minor corrections in comments
* prototype theorem ~sumvtxdg2size removed (see PR #2961)
* ~expnngt1, ~expnngt1b added to AV's mathbox

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13327,6 +13327,7 @@ New usage of "2bornot2b" is discouraged (0 uses).
 New usage of "2clwwlk2clwwlkOLD" is discouraged (0 uses).
 New usage of "2clwwlklemOLD" is discouraged (2 uses).
 New usage of "2cnALT" is discouraged (0 uses).
+New usage of "2irrexpqALT" is discouraged (0 uses).
 New usage of "2llnjN" is discouraged (1 uses).
 New usage of "2llnjaN" is discouraged (1 uses).
 New usage of "2llnm2N" is discouraged (1 uses).
@@ -18266,6 +18267,7 @@ Proof modification of "2bornot2b" is discouraged (26 steps).
 Proof modification of "2clwwlk2clwwlkOLD" is discouraged (782 steps).
 Proof modification of "2clwwlklemOLD" is discouraged (90 steps).
 Proof modification of "2cnALT" is discouraged (3 steps).
+Proof modification of "2irrexpqALT" is discouraged (90 steps).
 Proof modification of "2pm13.193" is discouraged (91 steps).
 Proof modification of "2pm13.193VD" is discouraged (223 steps).
 Proof modification of "2sb5nd" is discouraged (125 steps).
--- a/discouraged
+++ b/discouraged
@@ -13335,6 +13335,7 @@ New usage of "2llnm3N" is discouraged (0 uses).
 New usage of "2llnma2rN" is discouraged (1 uses).
 New usage of "2llnne2N" is discouraged (1 uses).
 New usage of "2llnneN" is discouraged (0 uses).
+New usage of "2logb9irrALT" is discouraged (0 uses).
 New usage of "2lplnm2N" is discouraged (0 uses).
 New usage of "2lplnmN" is discouraged (0 uses).
 New usage of "2pm13.193" is discouraged (3 uses).
@@ -18268,6 +18269,7 @@ Proof modification of "2clwwlk2clwwlkOLD" is discouraged (782 steps).
 Proof modification of "2clwwlklemOLD" is discouraged (90 steps).
 Proof modification of "2cnALT" is discouraged (3 steps).
 Proof modification of "2irrexpqALT" is discouraged (90 steps).
+Proof modification of "2logb9irrALT" is discouraged (141 steps).
 Proof modification of "2pm13.193" is discouraged (91 steps).
 Proof modification of "2pm13.193VD" is discouraged (223 steps).
 Proof modification of "2sb5nd" is discouraged (125 steps).
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -5228,6 +5228,12 @@ John Wiley &amp; Sons Inc. (1967) [QA300.A645 1967].</LI>
 Projective Geometry,</I> Academic Press, New York (1952)
 [QA471.B34].
 </LI>
+
+<LI><A NAME="Bauer"></A> [Bauer] Bauer, Andrej, &quot;Five stages of
+accepting constructive mathematics,&quot; <I>Bulletin (New Series) of
+the American Mathematical Society</I>, 54:481-498 (2017),
+DOI: <A HREF="http://dx.doi.org/10.1090/bull/1556">10.1090/bull/1556</A> .</LI>
+
 <LI>
 <A NAME="BellMachover"></A> [BellMachover] Bell, J. L., and M.
 Machover, <I>A Course in Mathematical Logic,</I> North-Holland,


set.mm
```diff
@@ -708474,6 +708599,38 @@ but which is a consequence of that a contradiction implies anything (see
   $}
 
 
+$(
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+    Integer powers - extension
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  $( If an integer exponentiation with a positive mantissa is greater than 1,
```
benjub, line 708608: An integer power with...
(same below)

To me, these two are worth going to Main.

set.mm
```diff
@@ -127214,6 +127214,24 @@ Rational numbers (as a subset of complex numbers)
     LVOVHVRQVKVHVOVRVHVDVEVRVFBUCRUDUEVPVQADVKVDVEVQAUFZVDVKBSFZVEVSBUKVKASFVTV
     EVSAULABUGUHUIRUJUMUNUOUPUQURUSUTVFCDMVA $.
 
+  ${
+    $d A x y z $.
+    $( A positive rational is the quotient of two positive integers.
+       (Contributed by AV, 29-Dec-2022.) $)
+    elpq $p |- ( ( A e. QQ /\ 0 < A )
+                 -> E. x e. NN E. y e. NN A = ( x / y ) ) $=
```
benjub, line 127222: Maybe prove instead the biconditional
```
 elpq $p |- ( A e. QQ -> ( 0 < A <-> E. ... ) ) $=
```
?
Maybe move to Main ?

set.mm
```diff
@@ -353435,9 +353453,18 @@ the complex square root function (the branch cut is in the same place for
...
        is rational.  Statement in the Metamath book, section 1.1.5, footnote 27
-       on page 17.  (Contributed by AV, 23-Dec-2022.) $)
+       on page 17, and the "classical proof" for theorem 1.2 of [Bauer],
+       p. 483.  This proof is not acceptable in intuitionistic logic, since it
+       is based on the law of excluded middle:  Either
+       ` ( ( sqrt `` 2 ) ^c ( sqrt `` 2 ) ) ` , is rational, then
+       ` ( sqrt `` 2 ) ` are appropriate irrational numbers (see ~ sqrt2irr ),
+       otherwise ` ( ( sqrt `` 2 ) ^c ( sqrt `` 2 ) ) ` and ` ( sqrt `` 2 ) `
+       are appropriate irrational numbers, since
+       ` ( ( ( sqrt `` 2 ) ^c ( sqrt `` 2 ) ) ^c ( sqrt `` 2 ) ) = 2 ` .  For
```
benjub, line 353465: I would write:
> either s2 ^ s2 is rational, in which case s2, being irrational, can be chosen for both a and b, or s2 ^ s2 is irrational, in which case s2 ^ s2 can be chosen for a and s2 for b, since (s2 ^ s2) ^ s2 = 2 is rational.

set.mm
```diff
@@ -353500,6 +353527,15 @@ the complex square root function (the branch cut is in the same place for
       DNMOEFGBCDPQ $.
   $}
 
+  $( Commutative law for complex exponentiation of real numbers.  (Contributed
```
benjub, line 353530: real exponentiation

set.mm
```diff
@@ -354553,6 +354589,103 @@ be obtained in Metamath (without the need for a new definition) by the curry
       DWDOAWHDWEWFWG $.
   $}
 
+  $( The general logarithm of a positive real number with respect to a real
```
benjub, line 354592: The logarithm of a positive real number to a real base...

set.mm
```diff
@@ -354553,6 +354589,103 @@ be obtained in Metamath (without the need for a new definition) by the curry
...
+     base greater than 1 is positive iff the number is greater than 1.
+     (Contributed by AV, 29-Dec-2022.) $)
+  logbgt0b $p |- ( ( A e. RR+ /\ ( B e. RR+ /\ 1 < B ) )
+                   -> ( 0 < ( B logb A ) <-> 1 < A ) ) $=
```
benjub, line 354596: I don't know how your proof goes precisely, but if you proved the intermediate result
> |- ( 0 < A -> 0 <= (A-1) log A )

then it might be worth having it as a standalone.

benjub:
As I wrote at #2958, I'd rather have stated irrationality of (log 3)/(log 2) rather than (log 9)/(log 2). (and I'm not convinced by the usefulness of having logb rather than writing quotients explicitly).

set.mm
avekens: OK, according to Wikipedia "Exponentiation" or Bronshtein "Handbook of Mathematics": "Exponentiation" is the operation of raising a "base" x to the power of the "exponent" y, resulting in the "power" x^y, also called "x to the power of y". In German: Das "Potenzieren" ist die Operation, eine "Basis" x mit einem "Exponenten" (oder auch einer "Hochzahl") y zu potenzieren. Das Ergebnis ist die "Potenz" (oder auch der "Potenzwert") x^y, auch mit "x hoch y" bezeichnet. 

Unfortunately, the "exponent" is sometimes also called "power" (see also Bronshtein "Handbook of Mathematics"!).

Therefore, "integer exponentiation" is the operation of raising a "base" x to the power of an integer y, resulting in the "integer power" x^y.

By the way, the term "mantissa" is used in the context of logarithms, isn't it? In the context of exponentiations, "base" should be used, shoudn't it?

As a consequence, "integer exponentiation" must be replaced by "integer power" in many places in set.mm, e.g., ~expeq0 ("Positive integer exponentiation is 0 iff its mantissa is 0" => "A positive integer power is 0 iff its base is 0."). Furthermore, statements like "raised to a negative integer power" are also not precise according this terminology (it shoud read "raised to the power of a negative integer").

If this is the agreed terminology, it should be explained at the beginning of section" Integer powers".

set.mm
avekens: I can (and will) prove 
```
elpqb $p |- ( ( A e. QQ /\ 0 < A ) <-> E. x e. NN E. y e. NN A = ( x / y ) ) $=
```

set.mm
avekens: No I used ~loggt0b to show that `( 0 < ( log `` A ) <-> 1 < A )` and then I divided both sides of the greater than relation on the left hand side by `( log `` B ) )`, which is greater than 0 because of `1 < B`.

set.mm
benjub: Yes, these are the meanings of base / exponent / power / exponentiation I learned in middle school (in French) and they are pretty standard. I didn't know the term "mantissa".

I think there is no urging need to make global changes, though. The terms power/exponent are sometimes confused with each other, but context allows the reader to know which is meant.

You may move your two results to Main in my opinion.

set.mm
benjub: You're right, that's better (prove it in place of elpq).

set.mm
jkingdon: One quirk which may affect the terminology is that we have three notations https://us.metamath.org/mpeuni/df-exp.html and https://us.metamath.org/mpeuni/df-ef.html and https://us.metamath.org/mpeuni/df-cxp.html which sometimes give the same values ( https://us.metamath.org/mpeuni/efzval.html and https://us.metamath.org/mpeuni/ecxp.html ) but in general are basically different things. The need to make this distinction is mostly metamath specific, although other texts may have something similar until they sweep it all under the rug with "since blah blah blah we can identify (notation one) with (notation two)").

As for mantissa, the usages in set.mm which I noticed seemed to be misusing the term, at least in the sense that https://en.wikipedia.org/wiki/Common_logarithm#Mantissa_and_characteristic and https://mathworld.wolfram.com/Mantissa.html and other web pages I found seemed to agree on what it is, and it doesn't seem to be the base.

Having said all that, I agree there isn't an urgent need to make any global changes.

set.mm
avekens: Done

set.mm

set.mm

set.mm

set.mm
avekens: Terminology adjusted (and documented).
2 theorems moved to main.

set.mm
```diff
@@ -127214,6 +127214,34 @@ Rational numbers (as a subset of complex numbers)
     LVOVHVRQVKVHVOVRVHVDVEVRVFBUCRUDUEVPVQADVKVDVEVQAUFZVDVKBSFZVEVSBUKVKASFVTV
     EVSAULABUGUHUIRUJUMUNUOUPUQURUSUTVFCDMVA $.
 
+  ${
+    $d A x y z $.
+    $( A positive rational is the quotient of two positive integers.
+       (Contributed by AV, 29-Dec-2022.) $)
+    elpq $p |- ( ( A e. QQ /\ 0 < A )
+                 -> E. x e. NN E. y e. NN A = ( x / y ) ) $=
```
benjub, line 127222: You can inline it in the proof of elpqb since it is subsumed by it.
You probably know that but just in case: do `MM> PROVE elpqb` and then `MM-PA> EXPAND elpq`, save the proof , write to source, and then remove elpq.

set.mm
```diff
@@ -139476,7 +139491,11 @@ notation is used by Donald Knuth for iterated exponentiation (_Science_
        4-Jun-2014:  The definition was extended to include negative integer
        exponents.  For example, ` ( -u 3 ^ -u 2 ) = ( 1 / 9 ) ` ( ~ ex-exp ).
        The case ` x = 0 , y < 0 ` gives the value ` ( 1 / 0 ) ` , so we will
-       avoid this case in our theorems.  (Contributed by Raph Levien,
+       avoid this case in our theorems.
+
+       For a definition of exponentiation including complex exponents see
+       ~ df-cxp (complex power function).  Both definitions are equivalent for
```
benjub, line 139497: (complex exponentiation function)

avekens: I used "complex exponentiation" in my latest commit, because `^c` is not a function, but an operation.

set.mm
avekens: I want to keep ~elpq as a separate theorem, because it is the more difficult part of the bijection (takes about 2/3 of the inlined proof). I could have made it a lemma (for ~elpqb ), but I think that one of the directions of a bijection deserves to be a theorem. 

Once there was a discussion that long proofs are difficult to understand (and to be minimized automatically), and somebody suggested to provide additional means to structure a proof. David and I argued that this is not needed, because interesting parts can (and should) be singled out as separate theorems, or as lemmas at least.

benjub: I think the combined proof would not be very long either, and the last step being biimpi, it is easy to see the two parts on the webpage. To me it is superfluous to have both a biconditional statement and one of its implications (with some exceptions, for instance for the simplest propcalc theorems, or when one direction requires fewer axioms), and there are many library theorems that accept biconditionals when only one implication is needed (mpbi, sylbi...).

I do not want to appear dictatorial, so if you do not agree, let's hear from a third person (and since @jkingdon was around...).

jkingdon: For me it is mostly a question of how long the proofs are (with a certain bias towards separating out theorems which can be readily stated separately, such as the two directions of a biconditional).

In this case, the one direction is 35 steps and the other 17 so I'd separate them. I might call them `elpq` (for the biconditional) and `elpqlem` (or something else starting with `elpqlem`) for the direction which is split out, but that is a minor non-blocking point and a few comments up @avekens expresses a preference not to do this, so I'd go with the author's preference on a thing like that.

jkingdon (approved):
Nice work! We've been talking about some of these issues and it is great to have many of the ideas in proof form.

set.mm
avekens: Since @jkingdon is at least not against a separate theorem elpq, I think we can keep it (and merge this PR).

Another (minor) reason to keep it is the comment "A positive rational is the quotient of two positive integers." is easier to understand (for non-formalists) than "A class is a positive rational...".

PR #2964, opened 2022-12-30
Get symbols from GitHub instead

david-a-wheeler:

### DIFF
--- a/scripts/download-metamath
+++ b/scripts/download-metamath
@@ -13,6 +13,9 @@ set -v -e
 # We determine the "latest" version from the number itself, not the date,
 # since a older commit might be tagged later on
 latest_zipball () {
+  echo "Debug info for latest_zipball $1" >&2
+  curl --silent "https://api.github.com/repos/$1/tags" >&2
+  echo "End of debug output from latest_version" >&2
   curl --silent "https://api.github.com/repos/$1/tags" | \
     grep zipball_url | grep -o 'https:[^"]*' | sort -n | tail -n 1
 }
@@ -25,7 +28,7 @@ latest_zipball () {
 set -x
 latest_version="$(latest_zipball metamath/metamath-exe)"
 echo "metamath-exe latest_version=$latest_version"
-curl --silent -L -o metamath-program.zip "$latest_version"
+curl -L -o metamath-program.zip "$latest_version"
 
 # Extract into a different directory to prevent overwriting .mm's
 # This will be updated, but we need a starting point:
--- a/scripts/download-metamath
+++ b/scripts/download-metamath
@@ -43,8 +43,9 @@ rm -f symbols.tar.bz2 symbols.tgz
 # get the symbols from our version controlled sources.
 # wget -N -q http://us.metamath.org/downloads/symbols.tar.bz2
 
-wget -N -q -o symbols.tgz https://github.com/metamath/symbols/tarball/main
+wget -N -O symbols.tgz https://github.com/metamath/symbols/tarball/main
 
 # Untar the symbols in the current directory so we can use them
 # (e.g., to test if they exist).
-tar --strip-components=1 -xzf symbols.tgz
+# Change the level of --strip-components to "1" to extract them into symbols/
+tar --strip-components=2 -xzf symbols.tgz
--- a/scripts/download-metamath
+++ b/scripts/download-metamath
@@ -8,27 +8,38 @@
 # Note that set.mm includes mmset.html and mmhil.html
 
 set -v -e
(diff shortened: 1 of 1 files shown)


digama0 (approved):
LGTM. We also need to update the website nightly build to download the symbols from the new repo.

scripts/download-metamath
```diff
@@ -36,9 +36,15 @@ cd ..
...
-rm -f symbols.tar.bz2
-wget -N -q http://us.metamath.org/downloads/symbols.tar.bz2
+rm -f symbols.tar.bz2 symbols.tgz
+
+# Get symbols file. We used to get it from the Metamath website, but now the
+# website is generated from version controlled files, so we need to
+# get the symbols from our version controlled sources.
+# wget -N -q http://us.metamath.org/downloads/symbols.tar.bz2
+
+wget -N -q -o symbols.tgz https://github.com/metamath/symbols/tarball/main
```
benjub, line 46: Why `-q` ? It can be informative to see the detailed output of the download. I do not know if there is a speed difference, but if there is, it is probably minor.

david-a-wheeler: Only because it was -q before. Happy to remove it, though hopefully it won't matter :-).

david-a-wheeler:
digama0 : Agreed. The website daily process that needs updating is in repo `metamath-website-scripts` file `install.sh`. But one step at a time, let's get this working first :-).

david-a-wheeler:
Huh, now I'm stuck. The script works fine locally, but not as a GitHub Action. I have no idea why. If anyone knows, please speak up :-).

benjub:
maybe you can add a similar
> echo "latest_zipball=$latest_zipball"

to get some info, and remove `--silent` from `curl --silent "https://api.github.com/repos/$1/tags"` ? Maybe it comes from the `grep`s on the following line, which may be different on your computer ?

david-a-wheeler:
Good news, I now know what the problem is - an API rate limit!!

~~~~
+ curl --silent https://api.github.com/repos/metamath/metamath-exe/tags
{"message":"API rate limit exceeded for 172.177.114.129. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
+ echo End of debug output from latest_version
End of debug output from latest_version
~~~~

Bad news: I don't know how to solve that. Hmm. Guru meditation.

jkingdon:
> Bad news: I don't know how to solve that. Hmm. Guru meditation.

Yeah, I'm not sure either. My first impulse was just "oh we need to authenticate" but the more I read https://docs.github.com/en/rest/overview/other-authentication-methods?apiVersion=2022-11-28#basic-authentication and the pages linked there the more I wasn't sure whether there was a way to create an access token with a small amount of access that we'd be willing to check it into a public repository. I mean, I'm not even sure you are allowed to do that because it would appear to kind of defeat the point of making you authenticate.

And then I was browsing https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28 and .....

Maybe there is a simpler solution to all this (for example: do we need to be using api.github.com or is there another way?)

digama0:
It is possible to use secrets in github CI actions without exposing them to the public repository. I don't quite follow how it works but it involves passing API keys in environment variables. See https://docs.github.com/en/actions/security-guides/encrypted-secrets

david-a-wheeler:
The only reason I used an apo was to get the current release of metamath. I suspect there is an easier way to do that now, let me investigate that first. I'd rather replace something complicated with something simple.

david-a-wheeler:
Okay, GitHub doesn't provide a way to do *exactly* what I was looking for, but you can easily just download a tarball or zip ball of the "current release an arbitrary branch" (like the master branch), and I think that's fine for our purposes. Downloading the file this way 100% eliminates the use of the GitHub API at all, and thus we don't need to worry about being rate-limited.

jkingdon (approved):
Nice work. I wonder whether rate limits explain some of the mysterious failures we've gotten every once in a while (which go away on rebuild).

digama0:
I think the rate limit on unauthorized requests is something like 10 requests per hour, so it's probably fine to have it in CI and your getting blocked might just have been a fluke due to all the requests for testing purposes.

david-a-wheeler:
> Nice work. I wonder whether rate limits explain some of the mysterious failures we've gotten every once in a while (which go away on rebuild).

Possibly! I guess we may see once this is merged. Intermittent problems are always the worst to try to fix. :-)

PR #2965, opened 2022-12-31
A few typos/edits

benjub:
A few typos/edits from a backlog.


set.mm
```diff
@@ -60102,8 +60105,8 @@ in the range of the function (the implication "to the right" is always
       H $.
   $}
 
-  $( The Cartesian product of two singletons.  (Contributed by Mario Carneiro,
-     30-Apr-2015.) $)
+  $( The Cartesian product of two singletons is the singleton consisting in the
+     associated ordered pair.  (Contributed by Mario Carneiro, 30-Apr-2015.) $)
```
avekens, line 60109: consisting in -> consisting of

set.mm
```diff
@@ -170889,23 +170892,23 @@ seq m ( x. , G ) ~~> z ) ) $.
...
     ( cc wcel cneg caddc co ce cfv cmul wceq negcl efadd mpdan cc0 negid fveq2d
     c1 ef0 syl6eq eqtr3d ) ABCZAADZEFZGHZAGHUBGHIFZQUAUBBCUDUEJAKAUBLMUAUDNGHQU
     AUCNGAOPRST $.
 
-  $( The exponential function never vanishes.  Corollary 15-4.3 of [Gleason]
-     p. 309.  (Contributed by NM, 13-Jan-2006.)  (Revised by Mario Carneiro,
-     29-Apr-2014.) $)
+  $( The exponential of a complex number is nonzero.  Corollary 15-4.3 of
+     [Gleason] p. 309.  (Contributed by NM, 13-Jan-2006.)  (Revised by Mario
+     Carneiro, 29-Apr-2014.) $)
```
avekens, line 170904: I would like to have the explanation/introduction of the term/noun "exponential" in the comment of df-ef. Maybe: The value `( e `` x )` of the exponential function at a complex number ` x ` is called "exponential (at ` x ` )".

set.mm
```diff
@@ -279710,7 +279713,7 @@ we show (in ~ tgcl ) that ` ( topGen `` B ) ` is indeed a topology (on
   ${
     $d g j v x J $.  $d g v N $.  $d g P $.  $d g v x S $.  $d g j v x X $.
     neifval.1 $e |- X = U. J $.
-    $( The neighborhood function on the subsets of a topology's base set.
+    $( Value of the neighborhood function on the subsets of the base set.
        (Contributed by NM, 11-Feb-2007.)  (Revised by Mario Carneiro,
```
avekens, line 279717: Why is "topology" removed from the comment? What is "the base set" then? I would like to keep "of a topology's base set".

set.mm
```diff
@@ -15296,7 +15296,7 @@ only postulated (that is, axiomatic) rule of inference of predicate
   altru $p |- A. x T. $=
     ( wtru tru ax-gen ) BACD $.
 
-  $( For all sets, ` F. ` is not true.  (Contributed by Anthony Hart,
+  $( For all sets, ` -. F. ` is true.  (Contributed by Anthony Hart,
```
avekens, line 15299: I would prefer not to change the comment: Both versions are equivalent, but while the new one is more formal, the former one is more intuitive/better understandable.

set.mm
benjub: Both are correct, but since a singleton has by definition a single element, that element really determines it, and this is what "consist in" emphasizes.

set.mm
benjub: Ok, done for that one, but I won't add it for "sine of A" and others.

set.mm
benjub: It's obvious, but ok, I will leave it. By the way, I don't know a single math book using that terminology, or defining a topology before "its base set".

I'll write "the base set of a topology" since apostrophes are for persons (that's what was bugging me in the comments and the main reason I removed it in the first place).

set.mm
benjub: What you say is correct, and I wouldn't change the comment for a more advanced theorem.  Here, it's one of the more elementary results, so I prefer the comment to be more faithful to the formal statement (and of course it's not that faithful, but there has to be a limit before completely paraphrasing the statement).

Also: I haven't completely given up on the idea of a more modular axiomatization (https://github.com/metamath/set.mm/issues/756#issuecomment-877045990) which has paraconsistent logic (that is, without "ex falso quodlibet") as one of its subsystems, where the above statement corresponds to the comment I proposed.

set.mm
jkingdon: "consist in" wasn't familiar to me, but the dictionary (at least https://www.merriam-webster.com/dictionary/consist%20in ) agrees with @benjub - "consist in" refers to "essential or main part" and "consist of" is merely "made up of".

iset.mm
```diff
@@ -55483,8 +55483,8 @@ associative structure (such as a group).  (Contributed by NM,
 
   ${
     $d A x y $.  $d B x y $.  $d F x y $.
-    $( The ` 2nd ` (second member of an ordered pair) function restricted to a
-       function ` F ` is a function of ` F ` onto the range of ` F ` .
+    $( The ` 2nd ` (second component of an ordered pair) function restricted to
```
jkingdon, line 55486: Although using "member" for ordered pairs is a common usage, I agree that "component" is better. So this change looks good to me.

set.mm
```diff
@@ -172276,8 +172282,8 @@ seq m ( x. , G ) ~~> z ) ) $.
     HMWBVMUNRUOQVFVKVFVKVFVJWEUPNUQVFVJWDVFSVJURTZSVJUSTZVFVIPCWFWAVIUTRVFSPCVJ
     PCWFWGVAVBWDSVJVCOVDVEQ $.
 
-  $( A number is real iff its imaginary exponential has absolute value one.
-     (Contributed by NM, 21-Aug-2008.) $)
+  $( A complex number is real iff the exponential of its imaginary part has
```
jkingdon, line 172285: Hmm, ` _i x. A ` is not the imaginary part of ` A ` . (that would be https://us.metamath.org/mpeuni/df-im.html ).

A verbose wording would be "the exponential of the number times `_i`" and I suppose that's probably better than "its imaginary exponential" (although I'd also be OK with keeping the latter even though it is not really self explanatory).

jkingdon (commented):
Thanks for taking this on! These look like good edits and I'm especially glad to see that you found more "exponential function" wording improvements than I did when I tried to get those.

One (presumably blocking, unless you disagree) comment but that should be easy.

set.mm
jkingdon: I tried to research "apostrophes are for persons" but the best reference materials (e.g. https://www.cambridge.org/core/books/cambridge-grammar-of-the-english-language/A78402ABF5176AD283494180BCA2046F ) are behind a paywall and the various pages I did find said something like "a thing" (or something more general than persons) and proceed to give examples that are exclusively persons or animals, so that's a bit inconclusive. Anyway, I don't object to "the base set of a topology" so we can go with that.

benjub: Genitives: you're right, I should have added animals. I learned in school that you can use it when it answers the question "whose is it" and we can answer "it's his/hers/theirs". I found the page: https://www.colanguage.com/genitive-english for what it's worth. But as you said, usage changes and apostrophes are used in other cases as well.  English does not have a "spec" like ISO C !
Side remark: I may sound pedantic, but actually I am more comfortable with strict grammar rules precisely because I am less comfortable with the English language and its evolving usage. Stricter rules make it easier (same as in math, where a beginning mathematician would be puzzled by abuses of language that mathematicians make every day).

iset.mm
benjub: Yes, especially since in set.mm they are encoded as `{ { x } , { x , y } }` (Kuratowski pair), which has one or two members, but there is no such thing as a "first" or "second" one.

set.mm
benjub: You're absolutely right, I overlooked it. I chose `the exponential of its product with _i has ...`.

On the other hand, the function $t \mapsto e^{it}$ appears often enough that it may be called "imaginary exponential" ? I don't know how standard that is.

benjub:
CI problem ! Help script wizard @david-a-wheeler !!! :)

benjub:
> Thanks for taking this on! These look like good edits and I'm especially glad to see that you found more "exponential function" wording improvements than I did when I tried to get those.

Thanks. I accumulate a few typos over time and then do them in batches once in a while. The key is to `MM> SET WIDTH 9999` `MM> WRITE SOURCE (i)set.mm /REWRAP` to make the `ctrl+F` find all instances.
> 
> One (presumably blocking, unless you disagree) comment but that should be easy.

You're right. I hope the new phrasing is good with you.

david-a-wheeler:
> CI problem ! Help script wizard @david-a-wheeler !!! :)

I'll try. My wizard hat doesn't fit well :-). Good news, the latest version seems to have passed.

Even better news, I *may* have recently fixed this. GitHub seems to now be imposing very strict rate limits on... GitHub. It appears that was causing mysterious intermittent problems, ones I don't think we had before. Now that I know about this issue, this newly-merged pull request eliminates our API request that was rate-limited:

https://github.com/metamath/set.mm/pull/2964

So hopefully that merged PR will eliminate the problem in the future.

set.mm
avekens: OK

set.mm
avekens: OK, again something learned.

avekens (approved):

set.mm
jkingdon: If it is standard I didn't find it in a (admittedly cursory) internet search.

jkingdon (approved):
Fixed wording looks good!

PR #2969, opened 2023-01-02
shorten eu6 again

wlammen:
*  reorder theorems based on df-eu directly before eu6. In particular eu3v is moved, too.
*  use eu3v, albiim, alanimi in eu6. This reduces the line count to 27.
*  move euf next to eu6


benjub:
But do we allow use of eu3v in eu6 ?  IIRC when the redefinition was done, we wanted to prove eu6 from df-eu only (with earlier theorems of course) ? And we wanted to have eu6 as early as possible since it was the old definition.

That being said, I'm ok to use eu3v in eu6. This is more natural and we should not be captive of the historical definition. But I would put eu6 right after eu3v, so that it comes early enough (and, as you did, euf right after it).

**edit:** How many essential steps now ?...

wlammen:
> And we wanted to have eu6 as early as possible since it was the old definition.

I personally see its former role rather noted in the comment of df-eu or eu6, than breaking the common ordering scheme. Simple theorems precede complex ones, as usage pattern suggests.  In the end it was the need of eu3v that triggered my moves.  On the other hand, the historic situation will soon be forgotten, people new to Metamath won't recognize the reason behind it anyway.
I don't remember the discussion any more.  If there is a promise to keep df-eu and eu6 together, I will move eu6, euf after eu3v.  It is not that important for me.

benjub:
So the best is probably to simply move eu3v right before eu6.

benjub:
Sorry, my latest message crossed your latest commit...

wlammen:
done for today

benjub (approved):

benjub:
@wlammen good for me if you want to merge as is

PR #2972, opened 2023-01-03
reduce axiom footprint of 2mo2

wlammen:
remove dependency on ax-6, ax-7, ax-10 and ax-12.  Propagates to 2eu4 as well.
eujust is an explanatory theorem for eu6 that should come before it (which was the case 2 days before)


benjub:
Looks like exdistrv might replace eeanv in other places too. Did you try some ?
Can you put cross links in the comments of both ? E.g. 
> For a version requiring fewer axioms but with a DV on ` x , y ` , see ~ exdistrv .
> For a version without DV on ` x , y ` but requiring more axioms, see ~ eeanv .

benjub (approved):

wlammen:
It was late in the night, I did not look further into this.

PR #2973, opened 2023-01-04
shorten cbvmo

wlammen:
... and remove an outdated OLD theorem.


benjub:
Can you prove cbveu from cbvmo and cbvex ? If it's shorter than the existing proof, then replace it. If it's longer, then put it as ALT since it is the "canonical way" to prove a cbvCONSTRUCTOR from more basic cbvCONSTRUCTOR's.

benjub (approved):
Good for me. You may merge now or also do the cbveu proof in that same PR.

avekens (approved):

PR #2974, opened 2023-01-04
Finish intuitionizing section "The exponential, sine, and cosine functions" (from cosmul to demoivreALT)

jkingdon:
Most of this section intuitionizes without change or just with easy changes to use apartness when dividing.

The case of `sinltx` is an interesting one which presumably can be intuitionized somehow, but for now just note it in mmil.html.

The only changes to set.mm are to minimize a few theorems there.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -390,6 +390,7 @@ New usage of "bj-nn0sucALT" is discouraged (0 uses).
 New usage of "bj-omex2" is discouraged (0 uses).
 New usage of "bj-omssonALT" is discouraged (0 uses).
 New usage of "bocardo" is discouraged (0 uses).
+New usage of "demoivreALT" is discouraged (0 uses).
 New usage of "df-div" is discouraged (2 uses).
 New usage of "df-ilim" is discouraged (1 uses).
 New usage of "df-inn" is discouraged (1 uses).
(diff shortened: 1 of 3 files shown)


avekens (approved):

iset.mm
```diff
@@ -118099,6 +118104,495 @@ Infinite sums (cont.)
...
+    oveq1d ) ABCZADEZFGHZIVBJHZJHZVBVBKHZIJHZFALHZDEZFVBLHZIJHUTVBBCZVJVDVFQZUT
+    VAAMZNZVMVJIBCZVJVKOVBIVBPUARUTVHVBAUBEZFGHZJHZVDUTAAKHZDEZVAVALHZVOVOLHZJH
+    ZVHVQUTVSWBQAAUCUDUTVGVRDAUEUFUTVBVTVPWAJUTVAVLSUTVOAUHZSUGTUTVCVPVBJUTVCVP
+    QZVBVPKHZIQZUTVPVBKHWEIUTVPVBUTVOWCNZVMUIAUJUKUTVJVPBCZWDWFULZVMWGVNVJWHWIO
+    IVBVPUMUNRUOUPUQUTVIVEIJUTVBVMURUST $.
+
+  $( Double-angle formula for cosine in terms of sine.  (Contributed by NM,
+     12-Sep-2008.) $)
+  cos2tsin $p |- ( A e. CC -> ( cos ` ( 2 x. A ) ) =
+                           ( 1 - ( 2 x. ( ( sin ` A ) ^ 2 ) ) ) ) $=
```
benjub, line 118185: IMO cos(2A) = (cos A)^2 -  (sin A)^2 is worth having in (i)set.mm and it could probably be used to prove both cos2t and cos2tsin and globally shorten the proofs. Just browsing nearby theorems on sums and products of sines and cosines, there might be some other shortening opportunities. Any takers ? @avekens maybe ?

PR #2977, opened 2023-01-07
exdistrv

benjub:
I added exdistrv a while ago to have a version of eeanv using fewer axioms. The idea, as with other lemmas I added overtime, is that these are used mainly with dummy variables, which can always be assumed disjoint, so that the versions of the lemmas with DV conditions are as useful as the ones without them. This is what happened for instance in #2972.

To avoid having to do them one at a time, I simply replaced all uses of eeanv with exdistrv and ran the verifier to see where I should revert to eeanv because of DV violations. To my surprise, no violation occurred in the ~60 replacements. There are probably some axiom usage reductions (when the axioms at stake (ax-5,6,7,10,11,12) are not used in other parts of the proof). Since the changes are almost automatic, I didn't add credit tags.

Also, I added 4exdistrv, which is to ee4anv what exdistrv is to eeanv (with interestingly two quantifiers swapped to avoid use of alcom).

Same thing could be done with the other lemmas I added around the same time: equsalv(w), equsexv(w) (and one could also check for the older lemmas cbvalv*, cbvexv*).

**edit:** done in #3004

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13338,7 +13338,6 @@ New usage of "2llnneN" is discouraged (0 uses).
 New usage of "2logb9irrALT" is discouraged (0 uses).
 New usage of "2lplnm2N" is discouraged (0 uses).
 New usage of "2lplnmN" is discouraged (0 uses).
-New usage of "2mo2OLD" is discouraged (0 uses).
 New usage of "2pm13.193" is discouraged (3 uses).
 New usage of "2pm13.193VD" is discouraged (0 uses).
 New usage of "2pmaplubN" is discouraged (1 uses).
@@ -18273,7 +18272,6 @@ Proof modification of "2clwwlklemOLD" is discouraged (90 steps).
 Proof modification of "2cnALT" is discouraged (3 steps).
 Proof modification of "2irrexpqALT" is discouraged (90 steps).
 Proof modification of "2logb9irrALT" is discouraged (141 steps).
-Proof modification of "2mo2OLD" is discouraged (146 steps).
 Proof modification of "2pm13.193" is discouraged (91 steps).
 Proof modification of "2pm13.193VD" is discouraged (223 steps).
 Proof modification of "2sb5nd" is discouraged (125 steps).


benjub:
CI bug (ping @david-a-wheeler)
See https://github.com/benjub/set.mm/actions/runs/3863021064/jobs/6584950169
```
# Compile metamath, a C verifier (and more) of metamath files by Norm Megill.
# We could just run "gcc *.c -O2 -o metamath", but this will do some
# auto-detection to try to find the best available options.
autoreconf -i
configure.ac:41: installing 'config/compile'
configure.ac:62: installing 'config/config.guess'
configure.ac:62: installing 'config/config.sub'
configure.ac:18: installing 'config/install-sh'
configure.ac:18: installing 'config/missing'
configure.ac:33: error: required file 'src/Makefile.in' not found
Makefile.am: installing 'config/depcomp'
autoreconf: error: automake failed with exit status: 1
Error: Process completed with exit code 1.
```

**edit:** now it works (apparently it was spontaneously rerun ?).  It may still be worth investigating.

jkingdon:
> **edit:** now it works (apparently it was spontaneously rerun ?). It may still be worth investigating.

I made an issue at https://github.com/metamath/set.mm/issues/2979 as I have also been seeing this failure on one of my branches.

jkingdon (approved):
I saw something similar when I was intuitionizing predicate logic: that in some cases I could add a distinct variable constraint and all usages of the theorem would be fine with that, and that adding it made the proof significantly easier.

Examples include [nfraldxy](https://us.metamath.org/ileuni/nfraldxy.html) (compared with [nfrald](https://us.metamath.org/mpeuni/nfrald.html) in set.mm), or maybe [sbco2yz](https://us.metamath.org/ileuni/sbco2yz.html), although in that case we remove the distinct variable constraint in [sbco2](https://us.metamath.org/ileuni/sbco2.html) so I don't know whether all usages would satisfy the constraint or not.

set.mm
```diff
@@ -22712,11 +22729,11 @@ of the unique existential quantifier (note that ` y ` and ` z ` need not
        (New usage is discouraged.) $)
     2mo2OLD $p |- ( ( E* x E. y ph /\ E* y E. x ph ) <->
                           E. z E. w A. x A. y ( ph -> ( x = z /\ y = w ) ) ) $=
-      ( wex weq wi wal wmo eeanv jcab 2albii 19.26-2 19.23v albii alcom anbi12i
-      wa df-mo bitri 3bitri 2exbii 3bitr4ri ) ACFZBDGZHZBIZABFZCEGZHZCIZSZEFDFU
-      HDFZULEFZSAUFUJSHZCIBIZEFDFUEBJZUICJZSUHULDEKUQUMDEUQAUFHZAUJHZSZCIBIUTCI
-      ZBIZVACIBIZSUMUPVBBCAUFUJLMUTVABCNVDUHVEULVCUGBAUFCOPVEVABIZCIULVABCQVFUK
-      CAUJBOPUARUBUCURUNUSUOUEBDTUICETRUD $.
+      ( wex weq wi wal wa wmo exdistrv jcab 2albii 19.26-2 19.23v albii anbi12i
```
wlammen, line 22732: The diff between 2mo2 and 2mo2OLD (created in #2972)  is that one uses exdistrv and the other eeanv.  So either delete 2mo2OLD, or keep it using eeanv

wlammen (changes requested):

set.mm
benjub: Good catch. I removed it. Since there were many usages, I did a global "search and replace", so some "discouraged" proofs may have been affected, but the reason (if any) for their discouragement was not related to the use of eeanv/exdistrv.

wlammen (changes requested):
wfrlem4OLD is another theorem changed, although it has a "modification discouraged" tag attached.  I am not a friend of disrespecting this tag for whatever reasons, as this might stir up discussions whether some future "minor" change should also have a special allowance to break this rule again (and again...). In the end the rule may deteriorate to a badly followed recommendation.
Here the extra work is not that big.  It took me 5 minutes to search for eeanv in my set.mm, and find the above exception.  For ease of correction I give its old proof here:

`      ( vb vc wcel wa cfv wceq wral wss cv cdm cin cres wfn cpred wfrlem2 funfn
      wfun sylib fnresin1 syl adantr inss1 sseli w3a wex wi wfrlem1 abeq2i fndm
      raleqdv biimpar rsp 3adant2 exlimiv sylbi imp adantlr fvres adantl resres
      syl5 predss sseqin2 mpbi 3an6 2exbii eeanv bitri ssinss1 ad2antrr syl5com
      nfra1 nfan inss2 anim12d biimpi syl6com ralrimi ad2ant2l jca wb ineqan12d
      ssin sseq1 sseq2 raleqbi1dv anbi12d imbi2d mpbiri 3adant3 exlimivv sylbir
      syl2anb simpr preddowncl sylc syl5eq reseq2d fveq2d 3eqtr4d ralrimiva ) G
      UAZDOZHUAZDOZPZXNXNUBZXPUBZUCZUDZYAUEZJUAZYBQZYBYAEYDUFZUDZIQZRZJYASXOYCX
      QXOXNXSUEZYCXOXNUIYJABCDEFGILUGXNUHUJXSXTXNUKULUMXRYIJYAXRYDYAOZPZYDXNQZX
      NCEYDUFZUDZIQZYEYHXOYKYMYPRZXQXOYKYQYKYDXSOZXOYQYAXSYDXSXTUNUOXOXNMUAZUEZ
      YSCTZYNYSTZJYSSZPZYQJYSSZUPZMUQZYRYQURZUUGGDABMJCDEFGILUSUTZUUFUUHMYTUUEU
      UHUUDYTUUEPYQJXSSZUUHYTUUJUUEYTYQJXSYSYSXNVAZVBVCYQJXSVDULVEVFVGVMVHVIYKY
      EYMRXRYDYAXNVJVKYLYGYOIYLYGXNYAYFUCZUDYOXNYAYFVLYLUULYNXNYLUULYFYNYFYATUU
      LYFRYAEYDVNYFYAVOVPYLYACTZYNYATZJYASZPZYKYFYNRXRUUPYKXOUUGXPNUAZUEZUUQCTZ
      YNUUQTZJUUQSZPZYDXPQXPYNUDIQRJUUQSZUPZNUQZUUPXQUUIUVEHDABNJCDEFHILUSUTUUG
      UVEPZYTUURPZUUDUVBPZUUEUVCPZUPZNUQMUQZUUPUVKUUFUVDPZNUQMUQUVFUVJUVLMNYTUU
      RUUDUVBUUEUVCVQVRUUFUVDMNVSVTUVJUUPMNUVGUVHUUPUVIUVGUVHUUPUVGUVHUUPURZUVH
      YSUUQUCZCTZYNUVNTZJUVNSZPZURZUVHUVOUVQUUAUVOUUCUVBYSUUQCWAWBUUCUVAUVQUUAU
      USUUCUVAPZUVPJUVNUUCUVAJUUBJYSWDUUTJUUQWDWEYDUVNOZUVTUUBUUTPZUVPUWAUUCUUB
      UVAUUTUWAYDYSOUUCUUBUVNYSYDYSUUQUNUOUUBJYSVDWCUWAYDUUQOUVAUUTUVNUUQYDYSUU
      QWFUOUUTJUUQVDWCWGUWBUVPYNYSUUQWOWHWIWJWKWLUVGYAUVNRZUVMUVSWMYTUURXSYSXTU
      UQUUKUUQXPVAWNUWCUUPUVRUVHUWCUUMUVOUUOUVQYAUVNCWPUUNUVPJYAUVNYAUVNYNWQWRW
      SWTULXAVHXBXCXDXEUMXRYKXFJCYAEYDXGXHXIXJXIXKXLXMWL $.
`

benjub:
> wfrlem4OLD is another theorem changed, although it has a "modification discouraged" tag attached. I am not a friend of disrespecting this tag for whatever reasons, as this might stir up discussions whether some future "minor" change should also have a special allowance to break this rule again (and again...). In the end the rule may deteriorate to a badly followed recommendation. Here the extra work is not that big. It took me 5 minutes to search for eeanv in my set.mm, and find the above exception. For ease of correction I give its old proof here:

I restored the proof since you gave it. Note that "discouraged" does not mean "forbidden", and here we're dealing with an OLD proof which will be removed soon anyway.

The reason for the discouragement tags is to avoid involuntary modifications, not to forbid them.

More generally, proofs are discouraged for a reason, typically because we want to retain this or that property of the proof (e.g., not depending on the axiom of choice, or having a certain structure). A modification of the proof which retains that property can still be performed.

wlammen (approved):

PR #2978, opened 2023-01-07
remove options 'quiet' and 'silent' from all scripts

benjub:
remove options 'quiet' and 'silent' from all scripts; this eases troubleshooting


benjub:
Rk: as in #2977, the CI process failed at first, and then apparently was rerun and succeeded...

jkingdon (approved):

Issue #49, opened 2016-07-03
Add support for Euclidean Geometry


david-a-wheeler:
I think it'd be possible to select some well-known set of "axioms" defining 2D/3D Euclidean Geometry, e.g., those by [Hilbert](https://en.wikipedia.org/wiki/Hilbert's_axioms), [Tarski](https://en.wikipedia.org/wiki/Tarski's_axioms), or [Birkhoff](https://en.wikipedia.org/wiki/Birkhoff's_axioms).  Then you can build up to usual Side-Angle-Side (SAS), Angle-Side-Angle (SAS), Side-Side-Side (SSS), and CPCTC... and then many other proofs would become much easier (because they can be expressed as traditional Euclidean geometry proofs).  E.g., the isosceles triangle theorem would be trivial to prove with these tools.  It seems to me it'd be possible to connect existing axioms to Euclidean geometry, just as numbers are "created" from set and logic axioms.  I'm sure there are implications I'm not thinking of, but I thought I'd record the idea here.

sctfn:
I've got Tarski's axioms in my mathbox already except for the upper
dimensional axiom (it needs full linear algebra for the N-dimensional
version).

sorear:
Yeah, Tarski's axioms look like the best bet for me too.  Is upper dimension 2 doable with complex numbers?

sctfn:
Should be - I can do it in any concrete dimension w/o LA, but I need
something like inverse matrices for the general case.

david-a-wheeler:
@sctfn : Very interesting - I didn't realize you did that!  I was thinking that it'd be easier to prove things like the isosceles triangle theorem if I had Euclidean geometry.  However, I was thinking about using SSS and CPCTC, e.g., "side-side-side proves that these two triangles are congruent, and so by CPCTC these two corresponding angles are congruent."  I don't see anything in the Euclidean construction that defines SSS and CPCTC... am I missing them?

sctfn:
They're not there at the moment - I haven't even defined angle yet! I
burned out a bit back since the only textbook using Tarski's Axioms is in
Germany, which I don't speak ☺️. I'll get back to it sometime soon, now
that my life is quieter.
On Jul 3, 2016 7:33 PM, "David A. Wheeler" <notifications@github.com> wrote:

@sctfn <https://github.com/sctfn> : Very interesting - I didn't realize you
did that! I was thinking that it'd be easier to prove things like the
[...]

PR #2981, opened 2023-01-07
Document theorems implementing CPCTC.

david-a-wheeler:
Many geometry textbooks refer to the theorem congruent triangles are congruent (CPCTC).  This commit adds notes to various comments to specifically indicate the various theorems that together constitute CPCTC.

In some sense this completes #49, which identifies CPCTC as an example of what it means to add "support for Euclidean Geometry".


jkingdon (approved):

set.mm
```diff
@@ -382312,20 +382312,23 @@ proved using the inner form (see theorem 9.6 of [Schwabhauser] p. 69
       tgbtwnxfr.e $e |- ( ph -> E e. P ) $.
       tgbtwnxfr.f $e |- ( ph -> F e. P ) $.
       tgbtwnxfr.2 $e |- ( ph -> <" A B C "> .~ <" D E F "> ) $.
-      $( Deduce segment congruence from a triangle congruence.  (Contributed by
-         Thierry Arnoux, 27-Apr-2019.) $)
+      $( Deduce segment congruence from a triangle congruence.  This is a
+         segment portion of congruent triangles are congruent (CPCTC).
```
benjub, line 382316: The second sentence seems to lack some words or have an extra verb ?
What does CPCTC stand for ?

david-a-wheeler: You're right. Sorry about that. I will tweak it later.

david-a-wheeler: I've tweaked it, hopefully that resolves the problem!

david-a-wheeler:
Thanks to @benjub for noting the poor wording, I think it's fixed now.

In the process I've realized that there are two variants of CPCTC not directly proved, namely, the other two corresponding angles of congruent triangles. We probably should add those theorems, but that's a different problem than just documenting common alternate names for theorems :-).

benjub:
Thanks.  Now I can make sense of the comments.  Basically, I understand the first sentence of each comment, and I can agree that the second sentence is a rephrasing of the first.  I had never heard of "CPCTC" but if that's a staple of the US curriculum and you think adding it may help some people understand (or more likely, find) it better, then go for it.

david-a-wheeler:
Your guess is exactly correct, CPCTC is a staple of many US curricula, which is why I want to make it clear where it's proved. Use a search engine on "geometry CPCTC" and you'll see it all over. I want to try to connect our proofs to terms widely used by others, so that we can connect our rigorous results with what others already know.

jkingdon:
> Your guess is exactly correct, CPCTC is a staple of many US curricula, which is why I want to make it clear where it's proved. Use a search engine on "geometry CPCTC" and you'll see it all over.

I tried this and I found a whole lot more hits for "geometry SAS" or "geometry SSS". At least from memory, my (US) schooling included those terms but not CPCTC (at least not as an acronym).

>  I want to try to connect our proofs to terms widely used by others, so that we can connect our rigorous results with what others already know.

Totally agreed with this as a goal, although figuring out what terminology people know it by can sometimes be a bit tricky. Still, we should try as best we can.

david-a-wheeler:
> I tried this and I found a whole lot more hits for "geometry SAS" or "geometry SSS". At least from memory, my (US) schooling included those terms but not CPCTC (at least not as an acronym).

SAS & SSS go the other way.

CPCTC starts with 2 congruent triangles, and lets you prove statements about segments and angles.  SAS & SSS start with segments (and an angle for SAS), and let you prove a statement that certain triangles are congruent.

benjub (approved):

set.mm
```diff
@@ -382312,19 +382312,23 @@ proved using the inner form (see theorem 9.6 of [Schwabhauser] p. 69
       tgbtwnxfr.e $e |- ( ph -> E e. P ) $.
       tgbtwnxfr.f $e |- ( ph -> F e. P ) $.
       tgbtwnxfr.2 $e |- ( ph -> <" A B C "> .~ <" D E F "> ) $.
-      $( Deduce segment congruence from a triangle congruence.  (Contributed by
+      $( Deduce segment congruence from a triangle congruence.  This is a
+         portion of the theorem that congruent parts of congruent triangles are
+         congruent (CPCTC), focusing on a specific segment.  (Contributed by
```
avekens, line 382317: CPCTC stands for "corresponding parts of congruent triangles are congruent" (see https://www.cuemath.com/geometry/cpctc/), not "congruent parts of ...".

avekens (changes requested):
see inline comment

set.mm
david-a-wheeler: You're absolutely correct, nice catch. I've fixed the change. I'll separately propose changes to other sections based on #2994, #2984, and #3014.

david-a-wheeler:
I think this PR and https://github.com/metamath/set.mm/pull/3028 will clarify how we handle geometric congruence. My thanks to everyone who pitched in with relevant proofs!!

david-a-wheeler:
@avekens - I think it's resolved, do you agree?

avekens (approved):

PR #2982, opened 2023-01-08
Fixup build-metamath script

david-a-wheeler:
We're downloading the Metamath files from GitHub, so change the build-metamath script to match. In particular, we used to remove the directory structure of the Metamath files, but that's counterproductive. Instead, work with its structure.

### DIFF
--- a/scripts/build-metamath
+++ b/scripts/build-metamath
@@ -13,11 +13,14 @@ fail () {
 
 test -f metamath-program.zip || fail 'Cannot find metamath-program.zip'
 
-# Unzip the .zip file; put all its contents into directory "./metamath"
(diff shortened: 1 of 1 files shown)


jkingdon (approved):
I don't fully understand how this script works, so I guess I'll take @david-a-wheeler 's word for it that this will work better.

david-a-wheeler:
Basically, the code as distributed by GitHub uses a different directory structure, so the programs to go compile and record it have to be adjusted to work with the new structure.

david-a-wheeler:
I've gotten an approval, the tests pass, and I doubt anyone else wants to muck with the code, so I'll merge :-).

PR #2985, opened 2023-01-08
real subtraction

icecream17:
<details>
<summary>the real subtraction section exists to save some axioms when proving a few theorems</summary>

there are several identity theorems; it would be nice to use 0 as the additive identity, but the only way to prove ` 0 -R 0 = 0 ` that I know of is through [` 0 + 0 = 0 `](https://us.metamath.org/mpeuni/00id.html) which already uses all the axioms in rsubidaddid1 and more

there's readdcan (C + A) = (C + B) <-> A = B which removes a shared left addend, but eliminating a right addend uses (addcom <-- addcan2 <-- addid1 <-- 00id). Though if you can prove that the right identity is the same for all reals, then that right additive identity is any left additive identity:

R = right additive identity
L = left additive identity (proven exists)

L + L = L (left identity)
L + R = L (right identity)
L = R (readdcan)

</details>

---

save axioms on 0cnALT; 0cnALT is now (the 0re proof with icn) + recni, saving {8 13 sep nul pow pr un 1cn addcl mulcl mulrcl mulcom addass mulass distr i2m1 1ne0 1rid rrecex pre-lttri pre-lttrn pre-ltadd}

(0cnALT used 1cn? must be an upstream change idk what happened)

0cnALT2 is 0re + recni, it's used in rsubeulem1

afaict there's no difference between ALTALT and ALT2
The only use of ALTALT seems to be bj-rabtrALTALT
...it has the same proof as bj-rabtr. the comments don't mention it so I assume it's unintentional, so I deleted it

So now ALTALT has no uses.

---

on the mpegif display of df-rsub, a space after the subscript double-stroked r looked "too wide" so I eventually settled on using a hair space (&#x200A;)

no space, hair space, thin space, space
image
image
image
image

---

rnegmod was named based on negeu, I would initially guess rsubmod (and subeu) though. the construction of rnegmod was easier to prove than using `E*` (2 minutes later: wait there's rmo4 !?)

rsubeu corresponds to negeu, rnegeu corresponds to negeu with B = 0

I assume the "n" in npncan et al is something like negation but I don't see it explained anywhere (not in conventions-labels either). So based on that, "rn" is real negation or something.

---

based on conventions-comments (section contributors), (contributed by + revised by) is used in rru and a few theorems in the real subtraction section
this feels similar to some copyright rabbit holes so I'm following without thought to avoid nerdtrap though there's more than one contributor for some theorems so I give a link as well

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17329,6 +17329,7 @@ New usage of "pmod2iN" is discouraged (0 uses).
 New usage of "pmodN" is discouraged (0 uses).
 New usage of "pmodl42N" is discouraged (1 uses).
 New usage of "pn0sr" is discouraged (4 uses).
+New usage of "pncan3OLD" is discouraged (0 uses).
 New usage of "pnfexOLD" is discouraged (0 uses).
 New usage of "pnonsingN" is discouraged (3 uses).
 New usage of "pointpsubN" is discouraged (0 uses).
@@ -19673,6 +19674,7 @@ Proof modification of "pm2.21ddALT" is discouraged (10 steps).
 Proof modification of "pm2.43bgbi" is discouraged (16 steps).
 Proof modification of "pm2.43cbi" is discouraged (34 steps).
 Proof modification of "pm3.2an3OLD" is discouraged (19 steps).
+Proof modification of "pncan3OLD" is discouraged (49 steps).
 Proof modification of "pnfexOLD" is discouraged (4 steps).
 Proof modification of "prel12OLD" is discouraged (191 steps).
 Proof modification of "prel12gOLD" is discouraged (329 steps).
--- a/discouraged
+++ b/discouraged
@@ -5,6 +5,7 @@
 "0bdop" is used by "adjeq0".
 "0bdop" is used by "nmbdoplb".
 "0blo" is used by "nmblolbi".
+"0cnALT2" is used by "rsubeulem1".
 "0cnfn" is used by "nmcfnex".
 "0cnfn" is used by "nmcfnlb".
 "0cnfn" is used by "riesz1".
@@ -13260,6 +13261,8 @@ New usage of "00sr" is discouraged (4 uses).
 New usage of "0bdop" is discouraged (2 uses).
 New usage of "0blo" is discouraged (1 uses).
 New usage of "0cnALT" is discouraged (0 uses).
+New usage of "0cnALT2" is discouraged (1 uses).
+New usage of "0cnALTOLD" is discouraged (0 uses).
 New usage of "0cnfn" is discouraged (4 uses).
 New usage of "0cnop" is discouraged (1 uses).
 New usage of "0csh0OLD" is discouraged (0 uses).
@@ -13875,7 +13878,6 @@ New usage of "bj-nuliotaALT" is discouraged (0 uses).
 New usage of "bj-peirce" is discouraged (0 uses).
 New usage of "bj-peircecurry" is discouraged (0 uses).
 New usage of "bj-rabtrALT" is discouraged (0 uses).
-New usage of "bj-rabtrALTALT" is discouraged (0 uses).
 New usage of "bj-rabtrAUTO" is discouraged (0 uses).
 New usage of "bj-sbceqgALT" is discouraged (0 uses).
 New usage of "bj-sbeqALT" is discouraged (0 uses).
@@ -18242,7 +18244,9 @@ New usage of "znnenlemOLD" is discouraged (0 uses).
 New usage of "zrdivrng" is discouraged (1 uses).
 New usage of "zrhcofipsgnOLD" is discouraged (1 uses).
 New usage of "zrhcopsgndifOLD" is discouraged (0 uses).
-Proof modification of "0cnALT" is discouraged (49 steps).
+Proof modification of "0cnALT" is discouraged (82 steps).
+Proof modification of "0cnALT2" is discouraged (3 steps).
+Proof modification of "0cnALTOLD" is discouraged (49 steps).
 Proof modification of "0csh0OLD" is discouraged (195 steps).
 Proof modification of "0heALT" is discouraged (25 steps).
 Proof modification of "0reALT" is discouraged (15 steps).
@@ -18659,7 +18663,6 @@ Proof modification of "bj-peirce" is discouraged (25 steps).
 Proof modification of "bj-peircecurry" is discouraged (58 steps).
 Proof modification of "bj-rababwv" is discouraged (34 steps).
 Proof modification of "bj-rabtrALT" is discouraged (36 steps).
-Proof modification of "bj-rabtrALTALT" is discouraged (33 steps).
 Proof modification of "bj-rabtrAUTO" is discouraged (33 steps).
 Proof modification of "bj-ralcom4" is discouraged (63 steps).
 Proof modification of "bj-ralvw" is discouraged (34 steps).
--- a/discouraged
+++ b/discouraged
@@ -5,7 +5,7 @@
 "0bdop" is used by "adjeq0".
 "0bdop" is used by "nmbdoplb".
 "0blo" is used by "nmblolbi".
-"0cnALT2" is used by "rsubeulem1".
+"0cnALT3" is used by "resubeulem1".
 "0cnfn" is used by "nmcfnex".
 "0cnfn" is used by "nmcfnlb".
 "0cnfn" is used by "riesz1".
@@ -13261,8 +13261,8 @@ New usage of "00sr" is discouraged (4 uses).
 New usage of "0bdop" is discouraged (2 uses).
 New usage of "0blo" is discouraged (1 uses).
 New usage of "0cnALT" is discouraged (0 uses).
-New usage of "0cnALT2" is discouraged (1 uses).
-New usage of "0cnALTOLD" is discouraged (0 uses).
+New usage of "0cnALT2" is discouraged (0 uses).
+New usage of "0cnALT3" is discouraged (1 uses).
 New usage of "0cnfn" is discouraged (4 uses).
 New usage of "0cnop" is discouraged (1 uses).
 New usage of "0csh0OLD" is discouraged (0 uses).
@@ -18246,8 +18246,8 @@ New usage of "zrdivrng" is discouraged (1 uses).
 New usage of "zrhcofipsgnOLD" is discouraged (1 uses).
 New usage of "zrhcopsgndifOLD" is discouraged (0 uses).
 Proof modification of "0cnALT" is discouraged (82 steps).
-Proof modification of "0cnALT2" is discouraged (3 steps).
-Proof modification of "0cnALTOLD" is discouraged (49 steps).
+Proof modification of "0cnALT2" is discouraged (49 steps).
+Proof modification of "0cnALT3" is discouraged (3 steps).
 Proof modification of "0csh0OLD" is discouraged (195 steps).
 Proof modification of "0heALT" is discouraged (25 steps).
 Proof modification of "0reALT" is discouraged (15 steps).


digama0:
The n/p naming in `npncan` is an abbreviated naming system for "negative" and "positive" specifically within the context of that set of theorems. For theorems about real versions of those theorems, I would just prefix `re` as in `readdcan`, so `renpncan`.

set.mm
```diff
@@ -114701,10 +114701,25 @@ it could represent the (meaningless) operation of
...
+
+  $( Obsolete version of ~ 0cnALT as of 7-Jan-2022.  Alternate proof of ~ 0cn
+     which does not reference ~ ax-1cn .  (Contributed by NM, 19-Feb-2005.)
+     (Revised by Mario Carneiro, 27-May-2016.)
      (Proof modification is discouraged.)  (New usage is discouraged.) $)
-  0cnALT $p |- 0 e. CC $=
+  0cnALTOLD $p |- 0 e. CC $=
     ( vx ci cv caddc co cc0 wceq wrex wcel ax-icn cnegex ax-mp addcl mpan eleq1
     cc syl5ibcom rexlimiv ) BACZDEZFGZAPHZFPIZBPIZUBJABKLUAUCAPSPIZTPIZUAUCUDUE
     UFJBSMNTFPOQRL $.
```
benjub, line 114725: This may reduce axiom dependencies (which ones?) but this lengthens the proof. Keep as 0cnALT2 ?

set.mm
```diff
@@ -523993,12 +524024,6 @@ which requires fewer axioms (i.e., removes dependency on ~ ax-6 ,
       ( wtru crab wceq cv wcel wb nfrab1 nfcv cleqf tru rabid mpbiran2 mpgbir )
       CABDZBEAFZPGZQBGZHAAPBCABIABJKRSCLCABMNO $.
 
-    $( Alternate proof of ~ bj-rabtr .  (Contributed by BJ, 22-Apr-2019.)
-       (Proof modification is discouraged.)  (New usage is discouraged.) $)
-    bj-rabtrALTALT $p |- { x e. A | T. } = A $=
-      ( wtru crab ssrab2 wss wral ssid tru rgenw ssrab mpbir2an eqssi ) CABDZBC
-      ABEBNFBBFCABGBHCABIJCABBKLM $.
-
```
benjub, line 524001: Thanks. bj-rabtr may have changed to align with bj-rabtrALTALT when the definition of TRUE was changed.

set.mm
```diff
@@ -614237,6 +614271,250 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  $c -R $.
+
+  $( Real number subtraction. $)
+  crsub $a class -R $.
+
+  $( Based on ~ conventions-comments , I add a "(Contributed by...)" and a
+     "(Revised by...)" tag every time the proofs are equivalent.  I also add a
```
benjub, line 614286: what are "equivalent proofs" ?

set.mm
```diff
@@ -614237,6 +614271,250 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+    df-rsub $a |- -R = ( x e. RR , y e. RR |->
+                      ( iota_ z e. RR ( y + z ) = x ) ) $.
+  $}
+
+  ${
+    $d x y A $.  $d x y B $.  $d ph x y $.
+    rnegmodv.a $e |- ( ph -> A e. RR ) $.
+    rnegmodv.b $e |- ( ph -> B e. RR ) $.
+    rnegmodv.1 $e |- ( ph -> E. y e. RR ( B + y ) = A ) $.
+    $( There is at most one real negative, deductive form.  (Contributed by
```
benjub, line 614308: `E!` is "exactly one" ; `E*` is "at most one"

benjub:
I looked at the changes outside of your mathbox, so to me there is no opposition to merge, but I don't have time to review. Maybe @avekens ?

set.mm
icecream17: yeah I originally tried `E*` but then `E. => E!` was easier. Would `rnegeudv` still be the label?

set.mm
icecream17: it saves {8 13 sep nul pow pr un 1cn addcl mulcl mulrcl mulcom addass mulass distr i2m1 1ne0 1rid rrecex pre-lttri pre-lttrn pre-ltadd}

sure, although [currently 0cnALT] uses ax-1cn somehow so the comment isn't accurate anymore, how do I describe the new 0cnALT2?

set.mm
benjub: yes rnegeudv

benjub:
Side remark for the future: a commit message describes the changes that were made, so the one above should not be "pncan3OLD is discouraged", but something like "discourage pncan3OLD".

avekens:
I don't know if I will have time for a review in the next days. Maybe at the weekend.

set.mm
```diff
@@ -614237,6 +614271,250 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+
+  $( Real number subtraction. $)
+  crsub $a class -R $.
+
+  $( Based on ~ conventions-comments , I add a "(Contributed by...)" and a
+     "(Revised by...)" tag every time the proofs are equivalent.  I also add a
+     "Based on..." sentence so that any other contributors are credited.  In
+     these theorems the only difference is the use of ` -R ` instead of ` - ` .
+     All theorems are named and constructed based on the equivalent complex
+     number subtraction theorems. $)
```
avekens, line 614290: I think in the following cases, the original contributot need not be mentioned. It is sufficient to reference the original theorem.

avekens: The whole comment can be removd.

set.mm
avekens: The comments must be adapted, too. And the following theorem(s) are also affected.

set.mm
```diff
@@ -614237,6 +614271,250 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+      SUHUGOUIUNDESUSDBATUNURDECULCJUMUQAULCBLUAQUBUCUDUEUF $.
+  $}
+
+  ${
+    rsubaddd.1 $e |- ( ph -> A e. RR ) $.
+    rsubaddd.2 $e |- ( ph -> B e. RR ) $.
+    rsubaddd.3 $e |- ( ph -> C e. RR ) $.
+    $( Relationship between subtraction and addition.  (Contributed by Mario
+       Carneiro, 27-May-2016.)  (Revised by Steven Nguyen, 8-Jan-2023.) $)
+    rsubaddd $p |- ( ph -> ( ( A -R B ) = C <-> ( B + C ) = A ) ) $=
```
avekens, line 614434: "Based on ~ subaddd. " should be added, " (Contributed by Mario Carneiro, 27-May-2016.)  (Revised by Steven Nguyen, 8-Jan-2023.) " can be replaced by  (Contributed by Steven Nguyen, 8-Jan-2023.)

set.mm
```diff
@@ -614237,6 +614271,250 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+  repncan3 $p |- ( ( A e. RR /\ B e. RR ) -> ( A + ( B -R A ) ) = B ) $=
+    ( cr wcel crsub co caddc wceq rsubcl w3a eqid rsubadd mpbii mpd3an3 ancoms
+    ) BCDZACDZABAEFZGFBHZPQRCDZSBAIPQTJRRHSRKBARLMNO $.
+
+  ${
+    mvrlladdd.1 $e |- ( ph -> A e. RR ) $.
+    mvrlladdd.2 $e |- ( ph -> B e. RR ) $.
+    mvrlladdd.3 $e |- ( ph -> ( A + B ) = C ) $.
+    $( Move real LHS left addition into RHS (real subtraction).  (Contributed
+       by Steven Nguyen, 8-Jan-2023.) $)
```
avekens, line 614467: The comment and the label are a little bit misleading. Maybe "Move LHS of a sum into RHS of a (real) difference."; label "reladdrsub".

set.mm
```diff
@@ -614237,6 +614271,250 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  Real subtraction
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  $c -R $.
+
+  $( Real number subtraction. $)
+  crsub $a class -R $.
```
avekens, line 614283: As @digama0 sugested, the prefix "re" should be used for this real number theorems and definitions. This also conserns the label above (crsunb -> cresub), the following definition (df-rsub -> df-resub) and most of the related theorems (rsubval -> resubval).

avekens (changes requested):
see inline comments

set.mm
```diff
@@ -614237,6 +614271,250 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+  ${
+    mvrlladdi.1 $e |- A e. RR $.
+    mvrlladdi.2 $e |- B e. RR $.
+    mvrlladdi.3 $e |- ( A + B ) = C $.
+    $( Move real LHS left addition into RHS (real subtraction).  (Contributed
+       by Steven Nguyen, 8-Jan-2023.) $)
+    mvrlladdi $p |- B = ( C -R A ) $=
+      ( crsub co wceq wtru cr wcel a1i caddc mvrlladdd mptru ) BCAGHIJABCAKLJDM
+      BKLJEMABNHCIJFMOP $.
+  $}
```
benjub, line 614483: Unless you have something specific in mind, I would say it is probably useless to have the inference ~mvrlladdi when you already have the deduction ~mvrlladdd.

avekens:
I want to review the latest changes, but it is not possible to see the modifications via "Files changed" here in Github (Error message "Oops, something went wrong. Retry" or sometimes the "unicorn"). Is this because there are also problems with the metamath server (calling https://us.metamath.org/mpeuni/mmset.html results in "404 Not Found"). @david-a-wheeler is there a known issue?

digama0:
I don't think these are related.

benjub:
@avekens : same here, but this is a problem with GitHub, not metamath, so unfortunately there's not much we can do.

set.mm
avekens: OK now

set.mm
avekens: OK now

set.mm
```diff
@@ -614040,6 +614066,14 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
   c0exALT $p |- 0 e. _V $=
     ( cc0 ci cmul co c1 caddc ax-i2m1 eqcomi ovexi ) ABBCDZEFJEFDAGHI $.
 
+  $( Alternate proof of ~ 0cn using ~ ax-resscn , ~ ax-addrcl , ~ ax-rnegex ,
+     ~ ax-cnre instead of ~ ax-icn , ~ ax-addcl , ~ ax-mulcl , ~ ax-i2m1 .
+     Version of ~ 0cnALT using ~ ax-1cn instead of ~ ax-icn .  (Contributed by
+     Steven Nguyen, 7-Jan-2022.)  (Proof modification is discouraged.)
+     (New usage is discouraged.) $)
+  0cnALT2 $p |- 0 e. CC $=
```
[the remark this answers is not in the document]
avekens: OK now

set.mm
avekens: There are still some theorem labels prefixed with "r" instead of "re": ~rnegcl, ~rsubcl

set.mm
avekens: OK now

set.mm
avekens: OK now

avekens:
Since I still cannot see the modifications here on Github, I cloned Steven's fork locally and compared it there. Most of my issues are solved, only two labels should still be changed (see above). After this is done, I will accept this PR.

set.mm
icecream17: renegcl and resubcl already exist, the name conflicts

benjub: I cannot see the changes because github is still down, but since your novelty is to use your flavor of "real negation/substraction" instead of negation/substraction, maybe `rernegcl / rersubcl`.

avekens: OK now

avekens (approved):
All previous review remarks are incorporated now.

benjub:
I'm going to merge, but one question to @icecream17 : it may be the case that github's problem to display the changes is linked to the way you push your branches (e.g., something that makes github think the change is much larger than it really is). I might very well be wrong, since I don't know git/github enough. Can you briefly tell us your workflow ?

icecream17:
On my desktop computer I would just have a local clone and then `git push`, though recently I've used my school chromebook <sup>which of course has a few unremovable extensions that record what you do</sup>.
The process is still the same (git clone commit push), just with an online editor+terminal instead of a local one.

<details><summary>I wrote this before I saw the "briefly"</summary>

I use github codespaces/gitpod depending on whether I've used up the free hours per month, an example github codespace looks like:

<details>
<summary>screenshots</summary>

1. Open a new or preexisting "codespace"
image
2. Do changes as usual
image
3. `git commit` and `git push` as usual

gitpod works the same way

</details>

On desktop I also have access to mmj2 but that shouldn't affect anything.

<details>
<summary>I've actually contacted github support before but the response was a vague "try again without extensions + supported browser" which didn't work.</summary>

image
</details>

(To be fair the error is truly unhelpful, in other more actionable problems Github Support is pretty good. I suspect what's happening is that some code has a time limit. Pure guess but in that case, hard problem.)

It's definitely possible that committing on chromebook is causing these problems, so I'll try committing on desktop in the future. (On desktop committing at https://github.com/metamath/set.mm/pull/2841 (see hidden review) was already failing, but I'm hoping that's a separate issue.)

</details>

avekens:
After I had cloned Steve's fork into my local Git environment, there was no problem to work with it: The commit history was clean and showed no abnormalities. The modifications compared with the latest version of the develop branch in set.mm also had a manageable size, therefore I cannot understand the current problems with the Github Webpage (there are other issues which I observe, e.g. there are popups when hovering over a linked name which remain open...). @david-a-wheeler do you have an idea?

benjub:
Thanks @icecream17 and @avekens. Indeed @david-a-wheeler may have an idea.
At least for now, @icecream17 do not waste your time trying to change things, we don't even know if it's connected to your workflow (it's clearly a problem with github, but it looked like it happened more often with your PR's, so maybe github doesn't understand something connected to your workflow). First, let's try and track more precisely when it happens in the future, see if it happens with specific contributors or something else (I'm not even sure it is deterministic for now!).

benjub:
@icecream17 : I just had a look at https://us.metamath.org/mpeuni/mmrecent.html and spotted that in https://us.metamath.org/mpeuni/readdid2addid1d.html the comment is wrong: your hypotheses do not mean that A is a right identity: it is much weaker since it means that adding A to the right fixes *some* element, not *all* elements.
By the way: the fact that a right identity is a left identity in groups is probably already in set.mm ? If not, it could be nice to add it.

avekens:
> @icecream17 : I just had a look at https://us.metamath.org/mpeuni/mmrecent.html and spotted that in https://us.metamath.org/mpeuni/readdid2addid1d.html the comment is wrong: your hypotheses do not mean that A is a right identity: it is much weaker since it means that adding A to the right fixes _some_ element, not _all_ elements. 

Yes, the comment should be adjusted.

> By the way: the fact that a right identity is a left identity in groups is probably already in set.mm ? If not, it could be nice to add it.

If a group was defined as a semigroups with a left identity and a left inverse for each element (see ~dfgrp2), then there would be a unique identity which by definition is also a right identity (see ~ grplid and ~grprid). Since groups are defined as monoids with inverses, and a monoid is defined as having a left and right inverse in set.mm, there is no need for a special theorem showing that a a right identity is a left identity.

tirix:
> > By the way: the fact that a right identity is a left identity in groups is probably already in set.mm ? If not, it could be nice to add it.
> 
> If a group was defined as a semigroups with a left identity and a left inverse for each element (see ~dfgrp2), then there would be a unique identity which by definition is also a right identity (see ~ grplid and ~grprid). Since groups are defined as monoids with inverses, and a monoid is defined as having a left and right inverse in set.mm, there is no need for a special theorem showing that a a right identity is a left identity.

Actually, that is probably [grpridd](https://us.metamath.org/mpeuni/grpridd.html)

icecream17:
So you mean A only needs to be a right identity for one real number? Then this would be revised to "If a number is a right identity for some number, it is a left identity for all numbers"?

I think https://us.metamath.org/mpeuni/grpideu.html proves indirectly what you've stated. The requirements for monoid are even stronger (there is an element which is a left and right identity for all elements.) Edit: Didn't see above 2 comments

Magma (closure) subclass Semigroup (associative) subclass Monoid (left and right identity element)

Currently I've used the closure, add assoc, 0re, and readdcan axioms, which is enough to prove Semigroupness, but so far hasn't been enough to show that there's a unique left and right identity element (Monoidness).

I think it's impossible with the current axioms I'm using actually. Consider an "addition" operator that just outputs its second argument. That operator has closure, is associative, and readdcan is valid. But there is no one real right additive identity.

Also https://us.metamath.org/mpeuni/renegeudv.html doesn't use its first hypothesis. I think removing it is valid since it's implied in (B + y) = A. This saves some axioms in https://us.metamath.org/mpeuni/renegeu.html.

benjub:
> Actually, that is probably [grpridd](https://us.metamath.org/mpeuni/grpridd.html)

Yes, thanks @avekens and @tirix, the two classical results are ~grprinvd and ~grpridd.

benjub:
> So you mean A only needs to be a right identity for one real number? Then this would be revised to "If a number is a right identity for some number, it is a left identity for all numbers"?

Yes, except that you don't say "identity for some element" since "identity" means "for all elements". That's why I wrote the way I did above.

Here, your proof uses the cancellation law, so it's kind of "cheating" ! If you want to study axiomatics, I think you're better off looking at general theorems about monoids and (semi)groups. It looks like there are already many basic results, as @avekens and @tirix wrote above. Also have a look at the wikipedia pages (though my advice is to not waste too much time on that, and study more mainstream math instead; it's too easy to get lost in semigroups, quasigroups and loops...).

One thing I would like to have in set.mm is the Eckmann-Hilton argument. See (without looking at the proof !) https://en.wikipedia.org/wiki/Eckmann%E2%80%93Hilton_argument
(@tirix @avekens do you know if it's already proved ?  I haven't found it after a quick search)

If you decide to prove it, then the first step would be to propose the most natural and general statement, so that we can agree on it before you try to prove it.

> Also https://us.metamath.org/mpeuni/renegeudv.html doesn't use its first hypothesis. I think removing it is valid since it's implied in (B + y) = A. This saves some axioms in https://us.metamath.org/mpeuni/renegeu.html.

Good catch !  Yes, it's legit to remove it, for the reason you said. You can also remove the DV condition DV(x, y), but that is accessory.

PR #2987, opened 2023-01-10
Add continuous functions to iset.mm from df-cncf to cncfmptc

jkingdon:
This set of theorems intuitionizes without much difficulty. Many proofs need no changes, and only one statement needs to be stated differently (to change not equal to apart).

As for why these particular theorems, I suppose it doesn't need justification, but I'm working towards things like the monotonic intermediate value theorem (Theorem 5.5 of [Bauer]). At least in set.mm the intermediate value theorem is used to show that pi exists - https://us.metamath.org/mpeuni/pilem3.html - and it would seem that the monotonic intermediate value theorem would suffice for this purpose.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8246,6 +8246,11 @@ intuitionistic and it is lightly used in set.mm</TD>
   <TD>May be provable once summation is better developed</TD>
 </TR>
 
+<TR>
+  <TD>divccncf</TD>
+  <TD>~ divccncfap</TD>
+</TR>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT


benjub:
@jkingdon : thanks for having taken the habit of backporting your improvements to set.mm

avekens (approved):

iset.mm
```diff
@@ -125588,6 +125608,328 @@ reduced fraction representation (no common factors, denominator
...
+  Metric spaces
+#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#*#
+$)
+
+
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  Topological definitions using the reals
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
```
benjub, line 125629: @jkingdon : The nesting order should be swapped, here: "Metric spaces" is a particular kind of "Topological definition using the reals", which I would rename "Real numbers in topology", since, like most sections, it is not only about definitions.

iset.mm
```diff
@@ -125588,6 +125608,328 @@ reduced fraction representation (no common factors, denominator
...
+
+    $( Imaginary part is continuous.  (Contributed by Paul Chapman,
+       21-Oct-2007.)  (Revised by Mario Carneiro, 28-Apr-2014.) $)
+    imcncf $p |- Im e. ( CC -cn-> RR ) $=
+      ( vw vx vz vy cim cc cr ccncf co wcel wf cv cmin cabs cfv clt wbr crp wss
+      wral wi wrex imf imcn2 rgen2 wa wb ssid ax-resscn elcncf2 mp2an mpbir2an
+      ) EFGHIJZFGEKZALZBLZMINOCLPQUOEOUPEOMINODLPQUAAFTCRUBZDRTBFTZUCUQBDFRDCAU
+      PUDUEFFSGFSUMUNURUFUGFUHUIBDCAFGEUJUKUL $.
+
+    $( Complex conjugate is continuous.  (Contributed by Paul Chapman,
```
benjub, line 125858: Complex conjugation

iset.mm
```diff
@@ -125588,6 +125608,328 @@ reduced fraction representation (no common factors, denominator
...
+      adantl bitr4d ) CIJZDABKLMZNZDACKLMZACDUAZEOZFOZPLQRGOUBUDVBDRVCDRPLQRHOZ
+      UBUDUCEASGTUEZHTSFASZNZVAUSAIJZUQUTVGUFURVHUQABDUGUOUQURUHFHGEACDUIUJUSVF
+      VAURVFUQURVEFHATURVCAMVDTMVEGEABVCVDDUKULUMUOUNUP $.
+
+    $( The set of continuous functions is expanded when the range is expanded.
+       (Contributed by Mario Carneiro, 30-Aug-2014.) $)
+    cncfss $p |- ( ( B C_ C /\ C C_ CC ) -> ( A -cn-> B ) C_ ( A -cn-> C ) ) $=
+      ( vf wss cc wa ccncf co cv wcel cncff adantl simpll fssd cncffvrn adantll
+      wf wb mpbird ex ssrdv ) BCEZCFEZGZDABHIZACHIZUEDJZUFKZUHUGKZUEUIGZUJACUHR
+      ZUKABCUHUIABUHRUEABUHLMUCUDUINOUDUIUJULSUCABCUHPQTUAUB $.
```
benjub, line 125807: Probably shorter to prove cncfss first and pretty directly from the definition of -cn->

Then, a probably more useful version of cncffvrn is
```
$p |- ( ( C C_ CC /\ ( F e. ( A -cn-> B ) /\ ( F " A ) C_ C ) ) -> F e. ( A -cn-> C ) ) $=
```

More generally, I'm not convinced by this -cn-> definition. It would probably be more useful to define -Top-> such that `( A -Top-> B )` is the set of continuous maps from the topological space A to the topological space B, in the frameworl of extensible structures. This would avoid reproving general truths about continuous functions. I know it's not about iset.mm, so feel free to open an issue, or I'll open one myself if I find the motivation for it.

jkingdon: > I know it's not about iset.mm

Yeah, I'm not qualified to address pretty much all these comments. I really don't know this section of set.mm at all. Sometimes I learn a bit when I intuitionize a section but even that can be a very mechanical process which doesn't require a lot of engaging with deeper mathematical issues until/unless the proofs hit something which uses excluded middle in a way which is not easily replaced.

Sure, this is partly an issue of what I am or am not interested in, but it also is a pretty large amount of material which I don't know well enough to even have an ability to say what kinds of rearrangements would be good or bad.

PR #2989, opened 2023-01-14
reduce dependencies on axioms in euorv

wlammen:
1. eubidvOLD was not marked properly with an "Obsolete..." comment.  Did some research and found that it should swap names with eubidvOLDOLD, and that the date of obsoletion was 26-Sep-2022.
2. remove dependencies on ax-6, ax-7 and ax-12 in euorv.
3. As noted by @benjub: the same is possible with euanv and reubidva.


benjub:
Can you try a similar reduction with euanv ?
Also: reubidva

wlammen:
@benjub Done

benjub (approved):

PR #2991, opened 2023-01-15
Add Top (class of topologies) to iset.mm

jkingdon:
This section intuitionizes OK but a number of theorems don't work (at least not as-is with the set.mm proof).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3490,6 +3490,11 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>fiint</TD>
+  <TD>~ fiintim</TD>
+</TR>
+
+<TR>
   <TD>fodomfi</TD>
   <TD><I>none</I></TD>
   <TD>Might be provable, for example via ~ ac6sfi or induction directly.
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8280,6 +8280,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>rintopn</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on riinopn</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3481,7 +3481,8 @@ this implies excluded middle</TD>
   <TD><I>in general</I></TD>
   <TD>The set.mm proof depends on unfi and it would appear that
   mapping ` { A , B } ` to a natural number would decide whether
-  ` A ` and ` B ` are equal and thus imply excluded middle.</TD>
+  ` A ` and ` B ` are equal and thus imply any set has decidable
+  equality.</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8267,6 +8267,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>iinopn</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on abrexfi</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8273,6 +8273,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>riinopn</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on iinopn (the other issues
+  in the set.mm proof could apparently be handled by ~ fin0or ).</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3578,6 +3578,15 @@ this implies excluded middle</TD>
 </TR>
 
 <TR>
+  <TD>df-fi and all theorems using the fi (class of finite
+  intersections) notation</TD>
+  <TD><I>none</I></TD>
+  <TD>Although some of these theorems could be intuitionized if
+  there is a need, others rely on excluded middle in apparently
+  essential ways.</TD>
+</TR>
+
+<TR>
   <TD>dfsup2</TD>
   <TD><I>none</I></TD>
   <TD>The set.mm proof uses excluded middle in several places and the
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8253,6 +8253,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>istop2g</TD>
+  <TD>~ istopfin</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>


iset.mm
```diff
@@ -64477,6 +64485,72 @@ would need to be strengthened (for example, to imply included middle) to
       DBCABDLTACDMTUKUPDULUMUPUQUNSUO $.
   $}
 
+  ${
+    $d x y z w v f A $.
+    $( Two finite intersection properties.  We show that pairwise intersections
+       of elements of a set being contained in that set implies that
+       intersections of finite subsets of that set are contained in the set.
+       The converse would appear to require an additional condition, such as
+       ` x ` and ` y ` not being equal, or ` A ` having decidable equality.
```
benjub, line 64494: > If a class is closed under pairwise intersections, then it is closed under nonempty finite intersections.

As for the converse "failing": I think it's because one should be careful in defining pairwise intersection: we probably have
```
|- ( A. x e. A A. y e. A ( x =/= y -> ( x i^i y ) e. A ) <->
      A. x ( ( x C_ A /\ x =/= (/) /\ x e. Fin ) -> |^| x e. A ) ) $=
```
intuitionistically.

iset.mm
```diff
@@ -125615,6 +125689,168 @@ reduced fraction representation (no common factors, denominator
...
+       220--241.
+
+       (Contributed by NM, 3-Mar-2006.)  (Revised by BJ, 20-Oct-2018.) $)
+    df-top $a |- Top = { x | ( A. y e. ~P x U. y e. x /\
+                             A. y e. x A. z e. x ( y i^i z ) e. x ) } $.
+  $}
+
+  ${
+    $d x y z J $.  $d x A $.
+    $( Express the predicate " ` J ` is a topology."  See ~ istopfin for
```
benjub, line 125748: ".

iset.mm
```diff
@@ -125615,6 +125689,168 @@ reduced fraction representation (no common factors, denominator
...
+       of its open sets is due to Aleksandrov.  For the convoluted history of
+       the definitions of these notions, see
+
+       Gregory H. Moore, _The emergence of open sets, closed sets, and limit
+       points in analysis and topology_, Historia Mathematica 35 (2008)
+       220--241.
+
+       (Contributed by NM, 3-Mar-2006.)  (Revised by BJ, 20-Oct-2018.) $)
+    df-top $a |- Top = { x | ( A. y e. ~P x U. y e. x /\
+                             A. y e. x A. z e. x ( y i^i z ) e. x ) } $.
```
benjub, line 125743: Should the definition be changed to
```
  df-top $a |- Top = { x | ( A. y e. ~P x U. y e. x /\ A. y e. x A. z e. x ( -. y = z -> ( y i^i z ) e. x ) ) } $.
```
I haven't thought nor read anything about that.  Has someone seen intuitionistic treatments of topology ? @digama0 ?

iset.mm
```diff
@@ -125615,6 +125689,168 @@ reduced fraction representation (no common factors, denominator
...
+      FDVJUSBVFDVFDURRSSUAEABUBUCULUQVEVAUQUMUPFZUOTZAOULVEUOAUPUDULVOVDAULVNVC
+      UOUMDCUEUFUGUHUIUJ $.
+
+    $( Express the predicate " ` J ` is a topology" using nonempty finite
+       intersections instead of binary intersections as in ~ istopg .  It is
+       not clear we can prove the converse without adding additional
+       conditions.  (Contributed by NM, 19-Jul-2006.)  (Revised by Jim Kingdon,
+       14-Jan-2023.) $)
+    istopfin $p |- ( J e. Top -> ( A. x ( x C_ J -> U. x e. J )
+         /\ A. x ( ( x C_ J /\ x =/= (/) /\ x e. Fin ) -> |^| x e. J ) ) ) $=
```
benjub, line 125774: Depending on the answer to the above comment, we'll know if this is an equivalence or if the second "characterization" is actually too lax for being a definition of "topology".

iset.mm
```diff
@@ -125615,6 +125689,168 @@ reduced fraction representation (no common factors, denominator
...
+  $}
+
+  $( The union of two open sets is open.  (Contributed by Jeff Madsen,
+     2-Sep-2009.) $)
+  unopn $p |- ( ( J e. Top /\ A e. J /\ B e. J ) -> ( A u. B ) e. J ) $=
+    ( ctop wcel w3a cpr cuni cun wceq uniprg 3adant1 wa wss prssi uniopn sylan2
+    3impb eqeltrrd ) CDEZACEZBCEZFABGZHZABIZCUAUBUDUEJTABCCKLTUAUBUDCEZUAUBMTUC
+    CNUFABCOUCCPQRS $.
+
+  $( The empty set is an open subset of a topology.  (Contributed by Stefan
```
benjub, line 125828: of **any** topology

It might be worth also proving the variable-free form `|- I^I Top = { (/) } `.

iset.mm
jkingdon: > > If a class is closed under pairwise intersections, then it is closed under nonempty finite intersections.

Ooh nice wording. I'll use this.

> As for the converse "failing": I think it's because one should be careful in defining pairwise intersection: we probably have

Exactly, I think the biconditional restated as you suggest would be provable.

We don't seem to need it yet, but perhaps some day.

iset.mm
jkingdon: > Should the definition be changed to
> 
> ```
>   df-top $a |- Top = { x | ( A. y e. ~P x U. y e. x /\
>                        A. y e. x A. z e. x ( -. y = z -> ( y i^i z ) e. x ) ) } $.
> ```
> 
> I haven't thought nor read anything about that. Has someone seen intuitionistic treatments of topology ? @digama0 ?

I looked a bit at Franka Waaldijk, modern intuitionistic topology, 1996, cited at the bottom of https://ncatlab.org/nlab/show/topology , but if it had answers to this sort of question I didn't find them. It seemed to be about more advanced topics.

My hunch is that we don't need to change the definition but I suspect it will become clear one way or the other as we try to prove more theorems. To be honest I'm still a bit lost in all the topology related definitions and notations in set.mm and I don't really know how many of them I need for my monotonic intermediate value theorem which is my focus at the moment.

iset.mm
jkingdon: > of **any** topology

I guess that is a bit better, although "a topology" isn't wrong. Changed in set.mm and iset.mm.

> It might be worth also proving the variable-free form `|- |^| Top = { (/) } `.

If you do decide to prove this I'd start out by putting it in your mathbox. At least to my eyes it is harder to read and I'm not sure where we'd use it.

iset.mm
jkingdon: Yeah, like I said the theorems in this pull request are only using the one direction.

iset.mm
jkingdon: Changed in both set.mm and iset.mm.

iset.mm
benjub: "a topology" in the sentence could be understood as "some topology" (at least when I translate it to French, maybe that's not the case in English), which is why "any" is clearer.

The statement I propose also says that the empty set is the only class with that property. I may prove it in set.mm if I find time.

iset.mm
benjub: Of course, because you may have a definition of topology which is too restrictive, making some theorems easier to prove (even if in practice, the difference might occur only for more advanced theorems).

iset.mm
benjub: Thanks for the pointer, I'll have a look. I would say: maybe do not go too fast before we are settled on the definition, since this may cause some rework. Indeed, I tend to believe that the definition I proposed is more in line with intuitionism (though I really haven't thought enough about it).

benjub: On page 50 of http://www.fwaaldijk.nl/modern%20intuitionistic%20topology.pdf the intersection axiom (O_3) is stated as in classical maths, but attention is restricted to topological spaces satisfying an axiom (O_4) which has to do with decidability of apartness.

benjub: I had a look at Bauer's "Five stages of accepting constructive mathematics". So between the two proposals above, I now think it is better to keep the classical one. Indeed, his definition (when converting from locales) is the same as in the classical case.

By the way, it's funny that you could prove stability under finite intersections which are nonempty, and not only inhabited. I'll have a look at the proof.

Is it interesting to see which results remain true if we adopt the more lax definition of a topology ?

Bauer defines topological spaces using locales. Maybe it would be better to do that in iset.mm ? (this can be done later, though)

iset.mm
benjub: Ohh, I had missed your comment in iset.mm saying "such as x and y not being equal". Sorry !

benjub (approved):

iset.mm
jkingdon: > By the way, it's funny that you could prove stability under finite intersections which are nonempty, and not only inhabited. I'll have a look at the proof.

If you know the set is finite the distinction between nonempty and inhabited goes away by https://us.metamath.org/ileuni/fin0or.html

jkingdon:
Sounds like if we find further things, for example from the locale investigation, we can address it in follow up pull requests.

PR #2992, opened 2023-01-17
Add TopOn to iset.mm

jkingdon:
This section intuitionizes without much fuss. A few theorems don't make it for various reasons, but most do.

Also included is a copying over a few theorems from other sections of set.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8298,6 +8298,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>topsn</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on pwsn</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8291,6 +8291,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>toprntopon</TD>
+  <TD><I>none</I></TD>
+  <TD>Presumably could be proved but the set.mm proof does not
+  work as it is.</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8286,6 +8286,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>toponsspwpw</TD>
+  <TD>~ toponsspwpwg</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>


avekens (approved):

iset.mm
```diff
@@ -33291,6 +33296,33 @@ same disjoint variable group (meaning ` A ` cannot depend on ` x ` ) and
       ( cpw wss cuni cv wral sspwuni unissb bitri ) BCDEBFCEAGCEABHBCIABCJK $.
   $}
 
+  $( Characterization of the elements of a double power class: they are exactly
+     the sets whose union is included in that class.  (Contributed by BJ,
+     29-Apr-2021.) $)
+  elpwpw $p |- ( A e. ~P ~P B <-> ( A e. _V /\ U. A C_ B ) ) $=
+    ( cpw wcel cvv wss wa cuni elpwb sspwuni anbi2i bitri ) ABCZCDAEDZAMFZGNAHB
+    FZGAMIOPNABJKL $.
```
benjub, line 33304: We should add  ` |- ( A e. ~P ~P B <-> U. A e. ~P B ) ` (from sspwuni and uniexb), and see if it's shorter to prove it from elpwpw or to prove elpwpw from it (in both set/iset).
You do it @jkingdon ? or @avekens ? Or else I do it in a few days ?

iset.mm
```diff
@@ -125850,6 +125890,157 @@ _Introduction to General Topology_ (1983), p. 114) and it is convenient
...
+      BAJK $.
+  $}
+
+  ${
+    $d B b j $.  $d J b j $.
+    $( Property of being a topology with a given base set.  (Contributed by
+       Stefan O'Rear, 31-Jan-2015.)  (Revised by Mario Carneiro,
+       13-Aug-2015.) $)
+    istopon $p |- ( J e. ( TopOn ` B ) <-> ( J e. Top /\ B = U. J ) ) $=
+      ( vj vb ctopon cfv wcel cvv ctop cuni wceq wa cdm wrel wfun funtopon crab
```
benjub, line 125926: nice to see that funtopon is useful here.

iset.mm
```diff
@@ -125850,6 +125890,157 @@ _Introduction to General Topology_ (1983), p. 114) and it is convenient
...
+  topontopon $p |- ( J e. ( TopOn ` X ) -> J e. ( TopOn ` U. J ) ) $=
+    ( ctopon cfv wcel ctop cuni topontop toptopon2 sylib ) ABCDEAFEAAGCDEBAHAIJ
+    $.
+
+  ${
+    $d A x y $.
+    $( The set of topologies on a set is included in the double power set of
+       that set.  (Contributed by BJ, 29-Apr-2021.)  (Revised by Jim Kingdon,
+       16-Jan-2023.) $)
+    toponsspwpwg $p |- ( A e. V -> ( TopOn ` A ) C_ ~P ~P A ) $=
```
benjub, line 125984: in both set/iset, the conclusion can be strengthened to ``( TopOn ` A ) e. ~P ~P ~P A ``. Is it useful ?

benjub, line 125984: in both set/iset, the conclusion can be strengthened to ``( TopOn ` A ) e. ~P ~P ~P A ``. Is it useful ?

iset.mm
jkingdon: I don't know, I'm not sure I'm really sold on the value of any of these theorems but I copied them over because they were in set.mm.

jkingdon: I think that most of them would need a re-think if we wanted to do this in CZF (although I guess @benjub might know more about CZF than I do, I'm mostly just seeing a power set).

iset.mm
jkingdon: Since `toponsspwpw` is currently unused in set.mm, I suppose I'd think in terms of removing the whole thing (or moving it to a mathbox).

benjub: Having an explicit upper bound on such a common construct as TopOn is important. Even if it's probably not that useful in CZF as you noted since powerset is not predicative, it is useful for instance to bound the set theoretic rank, or to tell that we stay within a given universe, or to bound the cardinality (finite or infinite).

iset.mm
benjub: Regarding CZF: these results probably remain true in CZF since they do not use that the powerclass of a set is a set.

benjub: Added the above proposed theorem (in both set/iset) in #3002.

iset.mm
benjub: As for the theorem I proposed above, let's add it only once a need emerges.

PR #2993, opened 2023-01-18
Scalar product in Euclidean spaces

tirix:
As mentioned in the mailing list by @avekens .


tirix:
(Sorry, several old rebases are polluting the history; only the last commit, "Scalar product in Euclidean Spaces", has actual new content)

avekens (approved):
Great! Thanks a lot, Thierry.

set.mm
```diff
@@ -320962,6 +320962,24 @@ need not be complete (for instance if the given set is infinite
...
+    ${
+      rrxvsca.r $e |- .xb = ( .s ` H ) $.
+      rrxvsca.i $e |- ( ph -> I e. V ) $.
+      rrxvsca.j $e |- ( ph -> J e. I ) $.
+      rrxvsca.a $e |- ( ph -> A e. RR ) $.
+      rrxvsca.x $e |- ( ph -> X e. ( Base ` H ) ) $.
+      $( The scalar product over generalized Euclidean spaces is the component
+         wise real number multiplication.  (Contributed by Thierry Arnoux,
+         18-Jan-2023.) $)
+      rrxvsca $p |- ( ph -> ( ( A .xb X ) ` J ) = ( A x. ( X ` J ) ) ) $=
```
benjub, line 320975: That was seriously hard to understand. There is no scalar product here. "Generalized Euclidean space" is a red herring. It would help to have a sentence like:
```
In ` ( RR^ `` I ) ` , the ` J ` th component of a scalar multiple is the (same)
multiple of the ` J ` th component.
```
Add or replace, as you prefer.
(ping @tirix, @avekens)

tirix: You are right I should obviously change "scalar product" into "[scalar multiplication](https://en.wikipedia.org/wiki/Scalar_multiplication)".

"Generalized Euclidean space" is what the definition [`df-rrx`](https://us.metamath.org/mpeuni/df-rrx.html) names that construct. Other nearby theorems also use that name, I just copied from them. Maybe we could open a different issue if this needs renaming?

benjub: Maybe open an issue.
After looking at df-rrx I think the best name for `` ( RR^ ` I ) `` is simply "`` ( RR^ ` I ) ``". It is a specific space so name it by its name.

avekens: > Maybe open an issue. After looking at df-rrx I think the best name for `` ( RR^ ` I ) `` is simply "`` ( RR^ ` I ) ``". It is a specific space so name it by its name.

I think "Generalized Euclidean space" is an adequate pronounceable name for this class. `` ( RR^ ` I ) `` is not a name, in my opinion. How would you pronounce it ("arar caret at i"?).

benjub: It seems I'll have to write an issue.

icecream17:
looks like the hypothesis rrxbase.b isn't used?
https://us.metamath.org/mpeuni/rrxvsca.html

(edit: Also in rrxcph)

tirix:
> looks like the hypothesis rrxbase.b isn't used?

Indeed! @avekens, I assume you want to use this theorem. 
Would you rather keep it as is or replace the last hypothesis 
```
      rrxvsca.x $e |- ( ph -> X e. ( Base ` H ) ) $.
```
with
```
      rrxvsca.x $e |- ( ph -> X e. B ) $.
```

benjub:
> with
> 
> ```
>       rrxvsca.x $e |- ( ph -> X e. B ) $.
> ```

that second solution looks more consistent

benjub:
By the way: the fact that there were two requests for change made after merging may have something to do with the fact that this PR was merged within roughly one hour of its opening.

avekens:
> > with
> > ```
> >       rrxvsca.x $e |- ( ph -> X e. B ) $.
> > ```
> 
> that second solution looks more consistent

I agree with @benjub. @tirix  could you modify the theorem accordingly, please?

tirix:
> By the way: the fact that there were two requests for change made after merging may have something to do with the fact that this PR was merged within roughly one hour of its opening.

Yes, sorry about that. In my head it was just a quick example following Alexander's mail, and I also wanted to make it quickly available to him. I'll be more patient in the future! ;-)

Issue #2984, opened 2023-01-08
tgsss1, tgsss2, and tgsss3 are not SSS as intended


david-a-wheeler:
Theorems `tgsss1`, `tgsss2`, and `tgsss3` are true, but I don't think they prove what was intended.

The comment currently describing theorem `tgsss1` says it is "Third congruence theorem:  SSS (Side-Side-Side):  If the three pairs of sides of two triangles are equal in length, then the triangles are congruent.  Theorem 11.51 of [Schwabhauser] p. 109.". However, `tgsss1` does *NOT* prove that given congruent segments you end up with ABC is-congruent-as-a-triangle with DEF. I would expect SSS to show that 2 triangles are congruent, and `tgsss1` doesn't do that. It instead proves that ABC is-congruent-as-an-angle with DEF. That is, `tgsss1` proves `<" A B C "> ( cgrA ` G ) <" D E F ">`, but I think it's supposed to instead prove `<" A B C "> ( cgrG ` G ) <" D E F ">` (note the use of `cgrG` instead of `cgrA`).

By using the Google lens app on Schwabhauser, I've become convinced that Schwabhauser also means to use `cgrG` and not `cgrA` for theorem 11.51.

I think we should keep these Metamath theorems (renaming them & changing their comment to match what they prove), but I think we separately need the intended SSS theorems from 11.51 that prove that certain *triangles* are congruent.

Am I misunderstanding something? @tirix @jkingdon @digama0 ?

tirix:
Thanks for reviewing @david-a-wheeler ! That's very precious. There are indeed two kinds of congruence, and what's important here is actually the notation, more than the explanations in german:

- In chapter 4.4, p35, Schwabhäuser first introduces (length) congruence of a set of points, and in particular for triangles, and chooses to write it e.g. `(abc) ≡ (def)` (note the parens). I've used `cgrG` to denote this in set.mm.
<img width="980" alt="image" src="https://user-images.githubusercontent.com/5831830/211225671-b9a16881-1372-4660-9302-973f93b16768.png">

- In chapter 11.2, p95, Schwabhäuser then introduces angle congruence for triangles, and chooses to write it `abc ≡ def` _without parens_ . There is actually a special mention in the german text underlining that the lack of parens makes the difference with length congruence introduced in chapter 4.4. I've used `cgrA` to denote this in set.mm (where the "A" stands for angle).
[...]

PR #2994, opened 2023-01-18
Equivalent definition of triangle congruence

tirix:
As proposed in issue #2984


avekens:
Great, but I would like to have an application of this new theorem, maybe by combining it with ~tgsss1, ~tgsss2, ~tgsss3 and  ~trgcgrg to obtain ~tgsss (in the same block as and after ~tgsss3), as I proposed in #2984:

```
( ph -> <" A B C "> ( cgrG ` G ) <" D E F "> ) $=
```

or even better, using the class .~ representation of ``( cgrG ` G )`` as in ~ trgcgrg (``$e |- .~ = ( cgrG ` G ) $.``):

```
( ph -> <" A B C "> .~ <" D E F "> ) $=
```

Could you add this, please?

tirix:
I started to write that theorem, and later realised that this is exactly the statement of [`trgcgr`](https://us.metamath.org/mpeuni/trgcgr.html). We could have a single-step proof, but it maybe makes more sense to update the theorem's comments accordingly, possibly with a link?

tirix:
If you agree, I'll propose to complete the comment for `dfcgrg2` as follows:

> Congruence for two triangles can also be defined as congruence of each segment and each angle (6 parts).  This is often the actual textbook definition of triangle congruence, see for example ~ https://en.wikipedia.org/wiki/Congruence_(geometry)#Congruence_of_triangles
>
> With this definition, the "SSS" congruence theorem has an additional part, namely, that triangle congruence implies congruence of the length of each sides. Because our development of elementary geometry strives to closely follow Schwabhaeuser's, our original definition of shape congruence, [`df-cgrg`](https://us.metamath.org/mpeuni/df-cgrg.html) already covers that part: see [`trgcgr`](https://us.metamath.org/mpeuni/trgcgr.html)

@david-a-wheeler your opinion is also welcome here!

avekens:
> I started to write that theorem, and later realised that this is exactly the statement of [`trgcgr`](https://us.metamath.org/mpeuni/trgcgr.html). We could have a single-step proof, but it maybe makes more sense to update the theorem's comments accordingly, possibly with a link?

It's not exactly the same, ~trgcgr does not require A =/= B, B =/=C, C =/= A. And using ~tgsss3 ~tgsss1 ~tgsss2 ~dfcgrg2 will result in another proof (definitely longer than 1 step).

You are rigth that such a ~tgsss does not add more mathematical value to set.mm, it would simply be a demonstration how a definition in the meaning of theorem ~dfcgrg2 could be used to prove the  SSS congruence theorem.

avekens:
> If you agree, I'll propose to complete the comment for dfcgrg2 as follows:

Yes, the comment should be enhanced, but more aligned to the CPCTC comments proposed in #2981.

tirix:
> Yes, the comment should be enhanced, but more aligned to the CPCTC comments proposed in #2981.

Do you have any specific wording in mind? I would suggest, _in addition_ to the text above:

> This theorem is also named "CPCTC", which stands for "Corresponding Parts of Congruent Triangles are Congruent".
> [https://en.wikipedia.org/wiki/Congruence_(geometry)#CPCTC](https://en.wikipedia.org/wiki/Congruence_(geometry)#CPCTC)

avekens:
> > Yes, the comment should be enhanced, but more aligned to the CPCTC comments proposed in #2981.
> 
> Do you have any specific wording in mind? I would suggest, _in addition_ to the text above:
> 
> > This theorem is also named "CPCTC", which stands for "Corresponding Parts of Congruent Triangles are Congruent".
> > https://en.wikipedia.org/wiki/Congruence_(geometry)#CPCTC

This is  a very good suggestion. After this is added, I think I can approve this PR. @david-a-wheeler what's your oppinion?

set.mm
```diff
@@ -389540,7 +389540,15 @@ then any point of the half line ( ` R ` ` A ` ) also lies opposite to
        segment and each angle (6 parts).  This is often the actual textbook
        definition of triangle congruence, see for example
 ~ https://en.wikipedia.org/wiki/Congruence_(geometry)#Congruence_of_triangles
-       (Contributed by Thierry Arnoux, 18-Jan-2023.) $)
+       With this definition, the "SSS" congruence theorem has an additional
+       part, namely, that triangle congruence implies congruence of the length
+       of each sides.  Because our development of elementary geometry strives to
```
benjub, line 389545: congruence of length --> do you mean "equality of lengths"

And by the way: isn't equality of lengths precisely the definition of congruence ?

benjub: To clarify:
* Definition: two sets of points A and B are congruent iff there exists a bijection f: A >->> B such that for all x, y in A, one has dist(x, y) = dist(f(x), f(y)).
* In particular, two segments are congruent iff their lengths are equal.

avekens (approved):
For me, this theorem and its comment is Ok now. If there is an additional issue with the concept of "congruence", its should be discussed in another place.

benjub:
> For me, this theorem and its comment is Ok now. If there is an additional issue with the concept of "congruence", its should be discussed in another place.

At least "congruence of length" --> "equality of lengths" (or "congruence of sides")

set.mm
```diff
@@ -389523,6 +389523,48 @@ then any point of the half line ( ` R ` ` A ` ) also lies opposite to
...
+    dfcgrg2.1 $e |- ( ph -> A =/= B ) $.
+    dfcgrg2.2 $e |- ( ph -> B =/= C ) $.
+    dfcgrg2.3 $e |- ( ph -> C =/= A ) $.
+    $( Congruence for two triangles can also be defined as congruence of each
+       segment and each angle (6 parts).  This is often the actual textbook
+       definition of triangle congruence, see for example
+~ https://en.wikipedia.org/wiki/Congruence_(geometry)#Congruence_of_triangles
+       With this definition, the "SSS" congruence theorem has an additional
+       part, namely, that triangle congruence implies congruence of the length
+       of each sides.  Because our development of elementary geometry strives
```
benjub, line 389545: > triangle congruence implies congruence of sides

OR
> triangle congruence implies equality of the lengths of sides

(using "each" correctly would require to write things like "each pair of corresponding sides", which is a bit heavy)

benjub (changes requested):

set.mm
tirix: Updated!

set.mm
```diff
@@ -389523,6 +389523,48 @@ then any point of the half line ( ` R ` ` A ` ) also lies opposite to
...
+    dfcgrg2.1 $e |- ( ph -> A =/= B ) $.
+    dfcgrg2.2 $e |- ( ph -> B =/= C ) $.
+    dfcgrg2.3 $e |- ( ph -> C =/= A ) $.
+    $( Congruence for two triangles can also be defined as congruence of sides
+       and angles (6 parts).  This is often the actual textbook definition of
+       triangle congruence, see for example
+~ https://en.wikipedia.org/wiki/Congruence_(geometry)#Congruence_of_triangles
+       With this definition, the "SSS" congruence theorem has an additional
+       part, namely, that triangle congruence implies congruence of the length
+       of sides.  Because our development of elementary geometry strives to
```
benjub, line 389545: In my previous comment, there was the problem with "each", that you fixed, but the main point was that you can write "congruence of the sides" or "equality of the lengths of sides", but not "congruence of the lengths of sides" (since congruence of segments means equality of their lengths).

Maybe you can write:
> namely, that triangle congruence implies congruence of the sides (which means equality of their lengths).

or something like that.

Anyway, you get the point and you'll find the better phrasing (since I haven't read the whole previous discussion in the issue).

benjub (changes requested):

benjub (approved):

benjub (approved):

PR #2997, opened 2023-01-19
prove mo4f without ax-13

wlammen:
Proving mo4f without ax-13 needs reproving other theorems without ax-13 as well - all over the place.  Mostly this can be done by adding a DV(x,y) condition on an equation x = y.  So we have a trade off between this condition and usage of ax-13.  This was observed by BJ before and was to some extent exemplified in his mathbox. Now let's go and walk further down this path until mo4f is reached.
Removing the dependency on ax-13 in mo4f has a cascading effect on some other theorems as well.
If the addition of such a large amount of extra theorems is not welcome, I can add them as lemma directly before mo4f and equsb3, from where they can be moved later with an appropriate name into a final position, if desired.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,7 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
-19-Jan-23 bj-stdpc4v stdpcv     moved from BJ's mathbox to main set.mm
+19-Jan-23 bj-stdpc4v stdpc4v    moved from BJ's mathbox to main set.mm
 19-Jan-23 bj-sbftv  sbftv       moved from BJ's mathbox to main set.mm
 19-Jan-23 bj-sbfv   sbfv        moved from BJ's mathbox to main set.mm
 19-Jan-23 bj-equsb1v equsb1v    moved from BJ's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,10 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+19-Jan-23 bj-stdpc4v stdpcv     moved from BJ's mathbox to main set.mm
+19-Jan-23 bj-sbftv  sbftv       moved from BJ's mathbox to main set.mm
+19-Jan-23 bj-sbfv   sbfv        moved from BJ's mathbox to main set.mm
+19-Jan-23 bj-equsb1v equsb1v    moved from BJ's mathbox to main set.mm
  7-Dec-22 decmul1   [same]      revised - remove unneeded hypothesis
  5-Dec-22 f1oeq2d   [same]      moved from GS's mathbox to main set.mm
  5-Dec-22 dmexd     [same]      moved from GS's mathbox to main set.mm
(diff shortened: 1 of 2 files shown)


benjub (commented):
Looks good to me. A few non-blocking remarks:
* thanks @wlammen for the rigorous work, and for having carefully looked at my mathbox;
* you generally write "A specialization of ... with a distinct ..., but no dependency on... " while I write "Version of ... with a disjoint ..., which does not require ...". That's mostly a matter of style:
  * "version of..." indicates that the only difference between the two statements is the addition of the DV condition, while "specialization of..." can convey the mistaken idea that there is (1) a specialization (for instance, an instantiation) and (2) the addition of a DV condition;
  * distinct/disjoint variables : I prefer "disjoint" but it was discussed recently that in the case of setvar variables, both are acceptable;
  * but no dependency on / which does not require: ok, that's purely English style and a matter of taste.
* you're asking about the addition of a large amount of new theorems. It's true that I generally look at the ratio (number of theorems where axiom dependency is reduced)/(number of new theorems), and this ratio seems a bit low here. But you write that there is a "cascading effect", so maybe this ratio is higher than it seems. I think this addition is welcomed anyway.
* Finally, I would like to adopt the definition https://us.metamath.org/mpeuni/df-ssb.html also in order to reduce axiom dependencies. This may have some impact on these theorems. I'm going to open an issue to discuss it, hopefully today.

wlammen:
mo4f is provable without ax-13, so the question is how to accomplish it.  The minimum impact is to introduce lots of lemma mo4flemxy that follow just one purpose: do it, and nothing more.  The way I presented it here, goes a bit further:  it assumes the lemma have more value than just proving mo4f, and so I introduced them as theorems. But: they are still not in the best shape and position.  One shortcoming is the comment you already mentioned.  Another issue is a possible vulnerability to 'minimize', which made you add a (Proof modification is discouraged.) to those I took from your mathbox.  And then there is the question of consistent naming and completeness.  Is it ok to have a sbanv but no sborv? Why is mo4f named different to mo3v, although they conceptually meet each other? Are the comments sufficient and consistent?  Shouldn't many of them be moved immediately after sb5 instead of lingering close to the variants without dv condition?

My suggestion is to accept this PR as a first step and answer the questions in later work done by me or others.  An incremental approach that does not claim perfection, but outlines where to go next.

As for other theorems losing their dependency on ax-13, here is an incomplete list: mo4 axextmo eu4 moeq rmo4 mosneq nulmo euind eueq eueqi eueq2 eueq3 moeq3 reuhypd reuhyp reueq rmoeq sndisj 0disj disjxsn... I stop here, because I think the list is already long enough.

benjub:
The way you presented it is IMO the better way. Most of the added theorems are meant to be used in proofs where the setvar's are dummy variables, so are disjoint from any other anyway, so the added DV conditions are natural.

As for the discouragement tags in my mathbox, they were there either because I am paranoid, or because they dated back to when automatic minimizations weren't looking at axiom usage (a feature that I asked Norm to make the default a few years ago).

Indeed, there can be further refinements, but this PR can be merged as is as I wrote above. Maybe wait until the end of the weekend in case there are some reactions ?

Thanks for the beginning of the list of theorems having their dependencies reduced. It is indeed a long one.

benjub (approved):

avekens (approved):

benjub:
Remark regarding stdpc4v: since "stdpcX" theorems are meant to be as close as possible to the axioms used in standard treatments (https://us.metamath.org/mpeuni/mmset.html#traditional) and these treatments use individual variables (which are necessarily "disjoint"), there is a case to make that stdpc4v is actually the "correct" version of https://us.metamath.org/mpeuni/stdpc4v.html.

PR #3004, opened 2023-01-23
usage: cbval-->cbvalv1 / cbvex--> cbvexv1 where possible

benjub:
Replace some usages with the weaker versions having a DV condition. This probably reduces axiom usage in some places.

With this PR, I think that now the weaker variants of equsal*, equsex*, cbval*, cbvex* are used when possible. This was suggested at the end of the comment https://github.com/metamath/set.mm/pull/2977#issue-1523972545


wlammen (approved):
Mathboxes are affected, but I don't think in a negative way.

PR #3005, opened 2023-01-24
reorder some ax-12 based theorems before those also in need of ax-10

wlammen:
title says it all


benjub (approved):
The diffs make it a bit hard to check that no DV conditions were added when moving scopes around. If you checked that on your side, then that's good with me.

wlammen:
No dv conditions were added

wlammen:
Used a backup to restore the PR after an accidental operation in github.

PR #3008, opened 2023-01-27
add extensible structures to iset.mm

jkingdon:
Add `Struct` , `ndx` , `sSet` , `Slot` , `Base` , and ``|`s`` to iset.mm

There is a surprisingly large amount of excluded middle usage in this section given that it basically is just maps from integers to sets. So a fair number of theorems are merely noted as not yet added, or are added with changes such as new hypotheses.

There are a few very small fixes to set.mm and the usual updates to earlier parts of iset.mm/mmil.html for theorems used in this section.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -309,6 +309,9 @@
 "sbieh" is used by "sbie".
 "spimh" is used by "spim".
 "spimth" is used by "equveli".
+"strnfvn" is used by "baseval".
+"strnfvn" is used by "ndxarg".
+"strnfvn" is used by "strn0".
 New usage of "0cnALT" is discouraged (0 uses).
 New usage of "0reALT" is discouraged (0 uses).
 New usage of "19.21h" is discouraged (3 uses).
@@ -379,6 +382,7 @@ New usage of "axpre-mulgt0" is discouraged (0 uses).
 New usage of "axprecex" is discouraged (2 uses).
 New usage of "axresscn" is discouraged (4 uses).
 New usage of "axrnegex" is discouraged (0 uses).
+New usage of "baseval" is discouraged (0 uses).
 New usage of "bdcnulALT" is discouraged (0 uses).
 New usage of "bdnthALT" is discouraged (0 uses).
 New usage of "bdsbcALT" is discouraged (0 uses).
@@ -470,7 +474,7 @@ New usage of "spimth" is discouraged (1 uses).
 New usage of "stoic2a" is discouraged (0 uses).
 New usage of "stoic2b" is discouraged (0 uses).
 New usage of "strcollnfALT" is discouraged (0 uses).
-New usage of "strfvnn" is discouraged (0 uses).
+New usage of "strnfvn" is discouraged (3 uses).
 New usage of "tfri1dALT" is discouraged (0 uses).
 New usage of "trintssmOLD" is discouraged (0 uses).
 New usage of "uzind4ALT" is discouraged (0 uses).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -382,6 +382,7 @@ New usage of "axpre-mulgt0" is discouraged (0 uses).
 New usage of "axprecex" is discouraged (2 uses).
 New usage of "axresscn" is discouraged (4 uses).
 New usage of "axrnegex" is discouraged (0 uses).
+New usage of "basendx" is discouraged (0 uses).
 New usage of "baseval" is discouraged (0 uses).
 New usage of "bdcnulALT" is discouraged (0 uses).
 New usage of "bdnthALT" is discouraged (0 uses).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -470,6 +470,7 @@ New usage of "spimth" is discouraged (1 uses).
 New usage of "stoic2a" is discouraged (0 uses).
 New usage of "stoic2b" is discouraged (0 uses).
 New usage of "strcollnfALT" is discouraged (0 uses).
+New usage of "strfvnn" is discouraged (0 uses).
 New usage of "tfri1dALT" is discouraged (0 uses).
 New usage of "trintssmOLD" is discouraged (0 uses).
 New usage of "uzind4ALT" is discouraged (0 uses).
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2432,10 +2432,7 @@ and is evaluated at a set</TD>
 
 <TR>
   <TD>fnsnsplit</TD>
-  <TD><I>none</I></TD>
-  <TD>Although it should be possible to prove subset rather than
-  equality (using difsnss instead of difsnid ), this theorem is
-  lightly used in set.mm.</TD>
+  <TD>~ fnsnsplitss</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8269,6 +8269,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>isstruct</TD>
+  <TD>~ isstructim , ~ isstructr</TD>
+  <TD>The difference is the addition of the ` F e. V ` condition
+  for the reverse direction.</TD>
+</TR>
+
+<TR>
   <TD>istop2g</TD>
   <TD>~ istopfin</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8533,6 +8533,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>ressabs</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on ressress but at first glance this would
+  appear to be feasible given the ` B C_ A ` condition.</TD>
+</TR>
+
+<TR>
   <TD>istop2g</TD>
   <TD>~ istopfin</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8313,6 +8313,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>setsval</TD>
+  <TD>~ setsvala</TD>
+</TR>
+
+<TR>
   <TD>istop2g</TD>
   <TD>~ istopfin</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8383,6 +8383,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>setsres</TD>
+  <TD>~ setsresg</TD>
+</TR>
+
+<TR>
   <TD>istop2g</TD>
   <TD>~ istopfin</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8398,6 +8398,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>strfv2d</TD>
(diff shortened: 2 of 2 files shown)


david-a-wheeler:
That *is* surprising, my hat's off to the effort.

Given how fundamentally simple extensible structures are *supposed* to be, is there a change to set.mm we could do that would keep the same functionality & interface, yet reduce the problem/complications? Mapping integers to sets should be simple I'd think. I don't think set.mm and iset.mm must always be the same, but if there are easy ways to simplify compatibility, we should think about it.

jkingdon:
> That _is_ surprising, my hat's off to the effort.
> 
> Given how fundamentally simple extensible structures are _supposed_ to be, is there a change to set.mm we could do that would keep the same functionality & interface, yet reduce the problem/complications? Mapping integers to sets should be simple I'd think. I don't think set.mm and iset.mm must always be the same, but if there are easy ways to simplify compatibility, we should think about it.

At the moment it seems like intuitionizing these is probably in the minor annoyance category (like adding a bunch of ``( E ` ndx ) e. NN`` hypotheses for example - this is a lot like fvex in the sense that the convenience which set.mm gets out of this is, I suppose, worth divergence from iset.mm). But there are some cases where I'm not really sure yet (basically, the ones listed in mmil.html in this pull request which have "none" in the iset.mm column). As I (or anyone else who wants to help) gets more into the theorems which use those, it should become clear what is needed.

I'm trying to stay open minded about extensible structures. They seem cumbersome, but I haven't gotten enough into the places where they are used to even have ideas about whether there are alternate solutions.

Don't take statements like "just maps from integers to sets" too literally - for one thing in the current design neither "map" is true (see "such classes cannot be functions" in the https://us.metamath.org/mpeuni/df-struct.html comment) nor is "sets" (see "avoid sethood hypotheses" later in the same comment). For another thing there are a lot of notations, a lot of theorems, and these get used in a lot of places - so it is hard to sum up in a sentence what all this stuff really does.

avekens (approved):
Really impressive! I wonder, however, if and how the modifications will affect further definitions, theorems and proofs based on extensible structure. Let's see...

jkingdon:
> Really impressive! I wonder, however, if and how the modifications will affect further definitions, theorems and proofs based on extensible structure. Let's see...

Some of it won't be a problem (for example requiring slot numbers to be elements of ` NN ` , since we don't use anything else for a slot number). Other parts I'm not so sure about but as you say we'll see.

digama0:
I have a good idea of how the framework is supposed to work and what it is used for, so if you have any concrete change proposals or a summary of what you had to intuitionize I can work out what consequences that would have and whether it is a good idea to backport the change.

jkingdon:
> I have a good idea of how the framework is supposed to work and what it is used for, so if you have any concrete change proposals or a summary of what you had to intuitionize I can work out what consequences that would have and whether it is a good idea to backport the change.

I have little or no idea of whether it could be better and how (I mean, I could imagine a world in which fields don't get group theorems for free - unless they apply some explicit steps - but even if we wanted to go that far I don't have much perspective about where we'd add explicit steps and what they'd look like). I'm assuming we have little appetite for a rethink on that scale, anyway.

More tactically I *can* summarize what I had to intuitionize (or was not able to easily intuitionize):

* many places need conditions that an index is a natural number (``(Base ` ndx ) e. NN`` or the like)
* the `F e. _V` condition in https://us.metamath.org/ileuni/isstruct2r.html might be a hassle or might be relatively innocuous. I think it has been the latter so far.
* many places need sethood conditions (example: `A` in https://us.metamath.org/ileuni/setsvala.html , `A` and `B` in https://us.metamath.org/ileuni/setsresg.html , `W` in https://us.metamath.org/ileuni/setsnidn.html , and others, I don't know whether I found the most important ones to list here)
* the `if` statement in https://us.metamath.org/ileuni/df-ress.html is a bunch of trouble, because a lot of theorems (apparently) will need to have a subset condition added (or split into one theorem with it and one without it). There's no way to get decidability of that condition for free as far as I noticed. See list of affected theorems at https://github.com/metamath/set.mm/blob/af25345e0a4753912a5a9ec91ee0d50bfb240563/mmil.raw.html#L8498 and following lines.
* https://us.metamath.org/ileuni/setsssvald.html is subset rather than equality. I don't have much understanding of the implications of this.
* The whole "protect with identity" concept at https://us.metamath.org/mpeuni/strfvi.html seems not to work (how committed is set.mm to this technique? I think the theorems which use it can be redone - as I did for summation - but I also wasn't convinced that it has as many benefits as, say, fvex or snex)

digama0:
> * many places need conditions that an index is a natural number (`` (Base `` ndx ) e. NN `` or the like)

Ideally this would be bundled into some assumption along the lines `Base is a slot`. The assumption itself should not be problematic. One option would be to use the theorem ``E = Slot ( E ` ndx )`` for this; that is, just have this theorem say ``E = Slot ( E ` ndx ) /\ ( E ` ndx ) e. NN``, and use this combination as a hypothesis of e.g. setsnidn and a conclusion of ndxid and baseid.

> * the `F e. _V` condition in https://us.metamath.org/ileuni/isstruct2r.html might be a hassle or might be relatively innocuous. I think it has been the latter so far.

It should be fine; it should be implied by `F Struct X` so as long as it doesn't add hypotheses to `strleun` and related theorems there are no problems here.

> * many places need sethood conditions (example: `A` in https://us.metamath.org/ileuni/setsvala.html , `A` and `B` in https://us.metamath.org/ileuni/setsresg.html , `W` in https://us.metamath.org/ileuni/setsnidn.html , and others, I don't know whether I found the most important ones to list here)

These all seem fine, they will be fixed by the slot theorem modification mentioned above.

> * the `if` statement in https://us.metamath.org/ileuni/df-ress.html is a bunch of trouble, because a lot of theorems (apparently) will need to have a subset condition added (or split into one theorem with it and one without it). There's no way to get decidability of that condition for free as far as I noticed. See list of affected theorems at https://github.com/metamath/set.mm/blob/af25345e0a4753912a5a9ec91ee0d50bfb240563/mmil.raw.html#L8498 and following lines.

I think the most likely way to intuitionize this is to remove the if-statement and just use the non-trivial case.  It won't be the same structure, but it will be equal enough for most purposes (there are theorems about equal-enough structures which are usable in this context).

> * https://us.metamath.org/ileuni/setsssvald.html is subset rather than equality. I don't have much understanding of the implications of this.

This seems to be not the right theorem, a subset is not very helpful in this context. I think it should instead be an equality under some reasonable assumptions.

Looking at the proof, the issue seems to be that `S` needs to have decidable equality of its domain to prove that removing the slot and putting it back yields the same thing. This is the case here because the domain is a subset of the natural numbers, but we need an assumption to that effect, e.g. `S Struct X`.

> * The whole "protect with identity" concept at https://us.metamath.org/mpeuni/strfvi.html seems not to work (how committed is set.mm to this technique? I think the theorems which use it can be redone - as I did for summation - but I also wasn't convinced that it has as many benefits as, say, fvex or snex)

This is going to be an issue. My recommendation is to consistently use a `ph -> ` before hypotheses like ``B = (Base`S)``; this was already a bit dicey in some set.mm theorems but you are going to have a lot more problems with this because of theorems like [scaffval](https://us.metamath.org/mpeuni/scaffval.html) or [mgpplusg](https://us.metamath.org/mpeuni/mgpplusg.html) that are rife with excluded middle reasoning. Under some hypotheses it is not hard to do the simple rewrites there but it is unlikely that you will be able to do it in the empty context without that id trick.

jkingdon:
> One option would be to use the theorem `` E = Slot ( E ` ndx ) `` for this; that is, just have this theorem say `` E = Slot ( E ` ndx ) /\ ( E ` ndx ) e. NN ``, and use this combination as a hypothesis of e.g. setsnidn and a conclusion of ndxid and baseid.

This is roughly what I have been doing so far but given how this is used I like the idea of using a conjunction as described above. It'll take me a while to implement it all but I think a number of theorems will be read more nicely.

> > * the `F e. _V` condition in https://us.metamath.org/ileuni/isstruct2r.html might be a hassle or might be relatively innocuous. I think it has been the latter so far.
> 
> It should be fine; it should be implied by `F Struct X`

Other direction. This is when you are proving `F Struct X` not when you already have it. It is an annoying little nit because `F e. _V` is almost implied by the other conditions but if there is a way to apply https://us.metamath.org/ileuni/funex.html I didn't quite see it (due to the "function with maybe one more element" nature of `F` and that sort of thing).

> so as long as it doesn't add hypotheses to `strleun` and related theorems there are no problems here.

This should be fine. I have a proof of strleun (off on the https://github.com/jkingdon/set.mm/tree/slot branch which I'll eventually pull request but probably after the items described in this comment) as well as strleun in deduction form (which doesn't change the conditions, just puts them behind `ph ->`).

> > * the `if` statement in https://us.metamath.org/ileuni/df-ress.html is a bunch of trouble . . .
> 
> I think the most likely way to intuitionize this is to remove the if-statement and just use the non-trivial case. It won't be the same structure, but it will be equal enough for most purposes (there are theorems about equal-enough structures which are usable in this context).

Ah, so we'd do something like:

```
df-iress $a |- |`s = ( w e. _V , x e. _V |->
    ( w sSet <. ( Base ` ndx ) , ( x i^i ( Base ` w ) ) >. ) ) $.
```

That seems logical (and I guess it would appear to give the same result in the interesting cases - although I guess as I keep working through these I'll get to the equal-enough stuff).

> > * https://us.metamath.org/ileuni/setsssvald.html is subset rather than equality. I don't have much understanding of the implications of this.
> 
> This seems to be not the right theorem, a subset is not very helpful in this context. I think it should instead be an equality under some reasonable assumptions.
> 
> Looking at the proof, the issue seems to be that `S` needs to have decidable equality of its domain to prove that removing the slot and putting it back yields the same thing. This is the case here because the domain is a subset of the natural numbers, but we need an assumption to that effect, e.g. `S Struct X`.

Ah, OK. This is only used by https://us.metamath.org/mpeuni/estrres.html where `C` (which corresponds to `S` in https://us.metamath.org/mpeuni/setsidvald.html ) is indeed a structure. I'll take another look at this one; I'm guessing it gets easy with `S Struct X`.

> > * The whole "protect with identity" concept at https://us.metamath.org/mpeuni/strfvi.html seems not to work (how committed is set.mm to this technique? I think the theorems which use it can be redone - as I did for summation - but I also wasn't convinced that it has as many benefits as, say, fvex or snex)
> 
> This is going to be an issue. My recommendation is to consistently use a `ph -> ` before hypotheses like `` B = (Base`S) ``; this was already a bit dicey in some set.mm theorems but you are going to have a lot more problems with this because of theorems like [scaffval](https://us.metamath.org/mpeuni/scaffval.html) or [mgpplusg](https://us.metamath.org/mpeuni/mgpplusg.html) that are rife with excluded middle reasoning. Under some hypotheses it is not hard to do the simple rewrites there but it is unlikely that you will be able to do it in the empty context without that id trick.

A lot of stuff has to get put in deduction form anyway for set existence reasons more broadly, so those empty contexts are a lot more rare in iset.mm than set.mm. Thanks for the tip that this is likely one of those times.

avekens:
Instead of discussing this topic further in this closed PR, I would suggest to open a Github issue for this. The latestest comment (containing all relevant information) could be copied into this issue...

jkingdon:
> Instead of discussing this topic further in this closed PR, I would suggest to open a Github issue for this. The latestest comment (containing all relevant information) could be copied into this issue...

I'll make a pull request in a few days

jkingdon:
> Instead of discussing this topic further in this closed PR, I would suggest to open a Github issue for this. The latestest comment (containing all relevant information) could be copied into this issue...

Submitted as https://github.com/metamath/set.mm/pull/3021 and https://github.com/metamath/set.mm/issues/3022 . If there are any other issues which need further attention, create and issue or let me know and I'll create an issue.

Issue #3003, opened 2023-01-22
Changing the definition of proper substitution


benjub:
I would like to change the definition of proper substitution https://us.metamath.org/mpeuni/df-sb.html to the definition https://us.metamath.org/mpeuni/df-ssb.html.  Not right now, since there is still some work to do in my mathbox, but I'm posting now to give a heads up and hear opinions.  The main motivation is to allow a finer study of dependencies on the object-redundant axioms ax-10, ..., ax-13.

Of course, these two definitions are equivalent over {ax-mp, ax-gen, ax-1, ..., ax-13}, but they are not equivalent over {ax-mp, ax-gen, ax-1, ..., ax-7} (of course, axioms ax-8, ax-9 do not play any role here).  If one is in {ax-mp, ax-gen, ax-1, ..., ax-7}, then df-ssb is closer to our conception of proper substitution than df-sb, as show the more elementary theorems following df-ssb in my mathbox, e.g. https://us.metamath.org/mpeuni/bj-ssbim.html, https://us.metamath.org/mpeuni/bj-alsb.html, https://us.metamath.org/mpeuni/bj-ssbequ.html among many others (whereas their counterparts using df-sb require extra axioms among ax-10, ..., ax-13).  There are further comments on the df-ssb page.

This definition change is in the same spirit as the definition change for the nonfreeness predicate https://us.metamath.org/mpeuni/df-nf.html. It accompanies the ongoing effort to reduce axiom dependencies, with many contributors: lately, mainly Wolf and I, but this was already the case with Norm and others since long before I even knew Metamath.  I found that post on the discussion group about df-nf: https://groups.google.com/g/metamath/c/Ovxv2aXJOIM and among the latest PRs related to this are @wlammen's #2997 and #2989 (see also the discussions at #2977 and #2719).  There are also many theorems avoiding ax-13 in my mathbox.

digama0:
This is already the definition I use in MM0: https://github.com/digama0/mm0/blob/45e06c7f95293e974bebd66350c9c45ca304a9a3/examples/peano.mm0#L107-L113 . (This is necessary since you can't play the same tricks with bundling which presumably correspond to using additional axioms in set.mm.) So you won't hear any complaints from me.

david-a-wheeler:
@benjub - that plan makes sense. I would ask that you prove that the new definition is equivalent to the old one for {ax-mp, ax-gen, ax-1, ..., ax-13}. I suspect you were already going to do that :-).

[...]

PR #3009, opened 2023-01-28
reorder more theorems in the ax-12 section

wlammen:


benjub (approved):

benjub:
I guess you're doing this in anticipation of the change #3003, so thanks for this. That definition change may affect some axiom usages, so hopefully there won't be too much need to do some re-work.

By the way: https://us.metamath.org/mpeuni/exsb.html does not require ax-13, so can you move it to the previous section, where you see fit ? (**edit:** probably in another PR to ease reviewing)

wlammen:
you can see the effects easier if the theorems are ordered by axiom usage, that's right.

PR #3011, opened 2023-01-28
generalize sb8vv

wlammen:
In sb8vv the condition d(ph, y) can be replaced with F/ y ph, which makes it more flexible, and will later allow proving 2exsb without ax-13, and (hopefully) fewer dv conditions.  Let's see.


benjub (commented):
I think this can simply replace sb8vv, and similarly for sb8evv being replaced by a new sb8ev. Look at mo3 to see if it can be proved from it without using ax-13 (hence it would subsume mo3v).

benjub:
Thanks. And it looks like it works for mo3 too !

wlammen:
At least one can single out one direction in mo3 and mo3v into a lemma, since they share the proof steps.

But I will finish for today.

benjub (approved):

PR #3012, opened 2023-01-28
add nfmodv and remove ax-13 from mo3

benjub:
Following @wlammen's yesterday's #3011.


set.mm
```diff
@@ -22520,40 +22532,26 @@ of the unique existential quantifier (note that ` y ` and ` z ` need not
...
   ${
     $d x y z $.  $d ph z $.
-    mo3.1 $e |- F/ y ph $.
+    mo3.nf $e |- F/ y ph $.
     $( Alternate definition of the at-most-one quantifier.  Definition of
        [BellMachover] p. 460, except that definition has the side condition
        that ` y ` not occur in ` ph ` in place of our hypothesis.  (Contributed
-       by NM, 8-Mar-1995.)  (Proof shortened by Wolf Lammen, 18-Aug-2019.) $)
+       by NM, 8-Mar-1995.)  (Proof shortened by Wolf Lammen, 18-Aug-2019.)
+       Remove dependency on ~ ax-13 .  (Revised by WL, 29-Jan-2023.) $)
```
wlammen, line 22540: Ok, but this is a collaboration, so, please, `Revised by WL and BJ`.

wlammen (changes requested):

set.mm
benjub: Thanks for your intellectual honesty.
1. To me, my contribution was in adding ~nfmodv (in which I wrote "Contributed by BJ"), and you did most of ~mo3, so the revision could be attributed to you.
2. I noticed you rarely use your initials, so maybe you prefer your name spelled out ?
3. If we keep both names, then I advocate alphabetical order of the last name.

So what do you prefer:
- Revised by Wolf Lammen,
- Revised by BJ and WL,

or something else ?

wlammen: Revised by BJ and WL

If you somehow want to appreciate `and you did most of ~mo3`, you might reverse the order of names.

benjub: Done. I prefer alphabetical order in general to prevent from asking each time whose contribution is larger. (ping @avekens since we also have "co-authorships"). This is what is done for mathematical papers, probably for that reason.

wlammen (approved):

PR #3014, opened 2023-01-29
Justification for using equality of distance in the geometry axioms

tirix:
Following a [comment](https://github.com/metamath/set.mm/issues/2984#issuecomment-1375840081) from @jkingdon in #2984, this shows that given a distance function, one can define a congruence relation between pairs of points, where distances are equal exactly when the corresponding segments are congruents, and vice-versa, from the congruence to the distance-like function.

This might serve as a justification for using distance functions instead of congruence relations to express congruence.

For reference, the [original proposal](https://github.com/metamath/set.mm/issues/49#issuecomment-230382130) from @digama0 mentionned:

> ``D = ( dist ` X )`` is the "measure" of distance between points. Unlike the usage in metric spaces, we do not require that it take values in `RR`, but instead let it take values in any set whatsoever, with only equality being meaningful. Thus the primitive expression is `( x D y ) = ( z D w )`, which substitutes for `<. x , y >. Cgr <. z , w >`. This transformation is valid for any congruence relation, because for any equivalence relation on `( P X. P )` we can take `D` to be the map from `<. x , y >.` to its equivalence class. This way, we don't need to make a new slot for a four-place relation, particularly since in all cases of interest which are metric spaces, the congruence relation _is_ just `( x D y ) = ( z D w )`.

The theorems introduced in this PR are meant to prove just that.


avekens (approved):
very good job!

david-a-wheeler:
This is great! Thanks!

I think the *geometry* section should, in the comments, point back to these theorems as justification for what's been done. So let's merge this, then add the relevant comment text (as a separate PR) to refer to it.

jkingdon:
Nice work. I did not notice that there are no constraints on the range of the distance function (so this isn't really like a metric space, at least not in terms of how it is defined). So I'd encourage saying that (pasting those parts of  [the proposal](https://github.com/metamath/set.mm/issues/49#issuecomment-230382130) quoted above, or some wording along those lines) in suitable places in set.mm comments.

jkingdon (approved):

PR #3016, opened 2023-01-29
decpmul carry fix, 0re vs elre0re

icecream17:
decpmul doesn't carry the ones place multiplication,
the fix makes ex-decpmul longer

rename renegeudv to renegeulemv and remove extra hypothesis
elre0re saves 1cn and cnre over 0re, use in real subtraction
theorems are longer but some get shortened

add readdsub and renpncan3, resubidaddid1lem is now unused


wlammen (approved):
This is all within your mathbox.  Without looking closer into what this is about, I don't see a reason why we should not let you proceed as you like.

PR #3017, opened 2023-01-30
add forgotten mo3OLD and remove ax-13 from sbnf2

wlammen:


wlammen:
* remove ax-13 from sbnf2 and move it to the ax-12 section
* add forgotten mo3OLD

tirix (approved):

benjub:
@wlammen : It looks like there was a change between the approval and the merge ? @david-a-wheeler : shouldn't this be disallowed by github with our parameterization ?

(in the present case, the change is fine, but as a policy, I think we should forbid it, that is, a new change should "invalidate" the approval)

benjub:
re-ping @david-a-wheeler : isn't there a way to paramaterize our github repo so that a new commit invalidates previous approvals ? (see my post just above for the reasons)

PR #3018, opened 2023-01-30
rename equsb3lem et al to equsb3v et al, rmv ax-mulgt0 from 0nnn

icecream17:
also reduces ax-13 from the renamed elsb3v and elsb4v

(I wonder how many of these improvements to set.mm transfer to iset.mm)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+30-Jan-23 elsb4lem  elsb4v
+30-Jan-23 elsb3lem  elsb3v
+30-Jan-23 equsb3lem equsb3v     equsb3lem was a duplicate of equsb3v
 19-Jan-23 bj-stdpc4v stdpc4v    moved from BJ's mathbox to main set.mm
 19-Jan-23 bj-sbftv  sbftv       moved from BJ's mathbox to main set.mm
 19-Jan-23 bj-sbfv   sbfv        moved from BJ's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -15409,9 +15409,7 @@ New usage of "elridOLD" is discouraged (0 uses).
 New usage of "elringchomALTV" is discouraged (1 uses).
 New usage of "elrngchomALTV" is discouraged (1 uses).
 New usage of "elsb3OLD" is discouraged (0 uses).
-New usage of "elsb3vOLD" is discouraged (0 uses).
 New usage of "elsb4OLD" is discouraged (0 uses).
-New usage of "elsb4vOLD" is discouraged (0 uses).
 New usage of "elsetrecslem" is discouraged (1 uses).
 New usage of "elspancl" is discouraged (1 uses).
 New usage of "elspani" is discouraged (2 uses).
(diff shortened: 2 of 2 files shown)


benjub:
For the first renamings: if it's good with @wlammen then it's good with me.

The rest is good with me too, but one remark:
For nnne0/nnne0ALT: it is true that the main proof should be the one requiring fewer axioms and the ALT proof should be the shorter one. I am not sure if we are not extending this a bit too far, here. Axioms for complex numbers do not have exactly the same status as the FOL+ZFC+TarskiGrothendieck axioms, since it was chosen to "artificially" add them back after having constructed the number systems in ZF. (A pretty unique situation: for instance we haven't done so for Tarski's geometry (developed mainly by @tirix), but we could have). No strong opinion in the present case, but maybe a policy is needed in the future ? @digama0 ?

set.mm
```diff
@@ -21469,9 +21467,16 @@ proposition with a distinct variable (closed form of ~ nfsb4 ).
...
+    $( Specialization of ~ elsb4 with an extra distinct variable condition, but
+       no dependency on ~ ax-13 .  Adapted from ~ equsb3v .  (Contributed by
+       Wolf Lammen, 28-Jul-2022.)  (Revised by Steven Nguyen, 23-Jan-2023.) $)
+    elsb4v $p |- ( [ x / y ] z e. y <-> z e. x ) $=
+      ( wel nfv elequ2 sbiev ) CBDCADZBAHBEBACFG $.
+
+    $( Obsolete version of ~ elsb4v as of 23-Jan-2023.  (Contributed by Wolf
+       Lammen, 28-Jul-2022.)  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    elsb4vOLD $p |- ( [ x / y ] z e. y <-> z e. x ) $=
```
benjub, line 21479: To check with @wlammen but to me, since in elsb3v, elsb4v the only change was to replace sbie with sbiev, the OLD versions need not be kept (and when I do this kind of cleaning, I do not even put a "revised" tag, but you can do it if you want)).

wlammen (approved):

set.mm
wlammen: `If a proof, definition, axiom, or theorem is going to be removed, we often stage that change by first renaming its label with an OLD suffix`  From the conventions page.  The `often` allows you to not keep the OLD version, but if you like, you can do so.  I do not keep the OLD version in case of simple minimization.  sbiev was introduced to remove ax-13 from equsb3v, so this was a different case.

icecream17:
Yeah I thought nnne0/nnne0ALT was a little much too. Few if any downstream theorems would be helped by this. Maybe in the future, if such a situation comes up again, the comment of nnne0 could be changed to add the note.

tirix:
To rebound on this:

> It was chosen to "artificially" add complex numbers axioms back [...] we haven't done so for Tarski's geometry

Unlike complex numbers, Tarski's geometry is not introduced as a construction, but more as a definition, just like groups are not constructed but defined in set.mm. In some cases people talk about the "[group axioms](https://math.stackexchange.com/questions/2510774/why-are-the-group-axioms-called-axioms)" for the conditions that a structure must fulfil to be called a group, but those are not actual axioms of the set theory. In that sense I think there is no need to re-state axioms there.

That's appropriate because just like there are many examples of groups, there are many possible geometries (different dimensions, curvatures, euclidean or not, etc.)

wlammen (approved):

wlammen:
I verified that ax-pre-mulgt0 was removed from both nnne0 and 0nnn, and no other axiom was added instead.  Now the only open question is, whether the ALT versions should be kept or not.  If nobody continues the discussion within the next few hours, I will merge this PR.  Discussion still can take place here, and in case the majority later votes for removing them, this can easily be done in a separate task.

tirix:
It looks like the ALT version of `nnne0` is way shorter and probably easier to understand, I would vote for keeping it for those reasons.

PR #3021, opened 2023-02-01
iset.mm extensible structures: slots and cleanups

jkingdon:
This includes:

* Add the first few theorems from the section "Slot definitions" in set.mm. Most notably, `strleun` but a few others as well.
* Change the hypotheses of many of the structure theorems to be ``E = Slot ( E ` ndx ) /\ ( E ` ndx ) e. NN`` and provide a few theorems to help satisfy that hypothesis
* A version of https://us.metamath.org/mpeuni/setsidvald.html with a decidability condition
* Copying theorems from set.mm to iset.mm and fixing one comment in both set.mm and iset.mm

Many of these items were discussed in https://github.com/metamath/set.mm/pull/3008#issuecomment-1407805505

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -315,7 +315,7 @@
 "spimth" is used by "equveli".
 "strnfvn" is used by "baseval".
 "strnfvn" is used by "ndxarg".
-"strnfvn" is used by "strn0".
+"strnfvn" is used by "strsl0".
 New usage of "0cnALT" is discouraged (0 uses).
 New usage of "0reALT" is discouraged (0 uses).
 New usage of "19.21h" is discouraged (3 uses).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -94,6 +94,9 @@
 "axresscn" is used by "peano5nnnn".
 "axresscn" is used by "recriota".
 "axresscn" is used by "rereceu".
+"basendx" is used by "1strstrg".
+"basendx" is used by "2strbasg".
+"basendx" is used by "2strstrg".
 "bj-axemptylem" is used by "bj-axempty".
 "bj-axemptylem" is used by "bj-axempty2".
 "df-div" is used by "divfnzn".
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -126196,6 +126239,15 @@ class of all (base sets of) groups is proper.  (Contributed by Mario
     ndxid $p |- E = Slot ( E ` ndx ) $=
       ( cnx cfv wceq cslot ndxarg eqcomi sloteq syl5eq ax-mp ) BEAFZGZANHZGNBAB
       CDIJOABHPCBNKLM $.
+
+    $( A structure component extractor is defined by its own index.  That the
+       index is a natural number will also be needed in quite a few contexts so
+       it is included in the conclusion of this theorem which can be used as a
+       hypothesis of theorems like ~ strslfv .  (Contributed by Jim Kingdon,
+       29-Jan-2023.) $)
+    ndxslid $p |- ( E = Slot ( E ` ndx ) /\ ( E ` ndx ) e. NN ) $=
```
avekens, line 126248: Would it be useful in set.mm, too?

avekens (approved):

david-a-wheeler:
I think requiring a slot to be a member of `NN` is reasonable, and it'd be nice to avoid *unnecessary* differences between set.mm and iset.mm like this. So I think we should entertain the idea. However, I think Mario's input is key here, there may be something I don't see here.

david-a-wheeler:
@digama0 - can you please chime in, *especially* on the idea of limiting slot indexes to `NN`? I'd like to prevent completely unnecessary differences between set.mm and iset.mm, but perhaps there's an unfortunate limitation this implies that I don't see.

tirix:
@david-a-wheeler I think this was actually suggested by @digama0 in [this comment](https://github.com/metamath/set.mm/pull/3008#issuecomment-1407714158).

david-a-wheeler:
@tirix - great! I guess it's not hard to get Mario on board, since he suggested it. If it's not too hard, we might want to move that to set.mm, to eliminate unnecessary differences between set.mm and iset.mm.

iset.mm
jkingdon: You've done a lot of the recent work in set.mm so you might have as many thoughts on that as I, but I would think that it could be.  Presumably it would be useful if done in conjunction with rearranging some of the theorems which would use it, that is those that have separate ``E = Slot ( E ` ndx )`` and ``( E ` ndx ) e. NN`` hypotheses (or variations on that which amount to the same thing).

PR #3023, opened 2023-02-01
add links to some *v theorems

wlammen:
and: Use mo3OLD in mo4fOLD to better reflect the initial version of mo4f

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -9336,6 +9336,7 @@
 "mndomgmid" is used by "isdrngo2".
 "mndomgmid" is used by "ismndo2".
 "mndomgmid" is used by "rngoidmlem".
+"mo3OLD" is used by "mo4fOLD".
 "mpv" is used by "mulcompr".
 "mulassnq" is used by "1idpr".
 "mulassnq" is used by "addclprlem2".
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -21845,8 +21850,9 @@ proposition with a distinct variable (closed form of ~ nfsb4 ).
 
   ${
     $d z w ph $.
-    $( An equivalent expression for double existence.  (Contributed by Wolf
-       Lammen, 2-Nov-2019.) $)
+    $( An equivalent expression for double existence. for a version requiring
```
benjub, line 21853: f -> F

set.mm
```diff
@@ -21440,9 +21444,10 @@ proposition with a distinct variable (closed form of ~ nfsb4 ).
 
   ${
     $d y w z $.  $d x w $.
-    $( Substitution applied to an atomic wff.  (Contributed by Raph Levien and
-       FL, 4-Dec-2005.)  Remove dependency on ~ ax-11 .  (Revised by Wolf
-       Lammen, 21-Sep-2018.) $)
+    $( Substitution applied to an atomic wff.  For a version requiring disjoint
```
benjub, line 21447: to an atomic wff -> to an equality
(an atomic formula can also be of the form `x e. y`)
It can even be "Substitution in an equality."

benjub (approved):

PR #3028, opened 2023-02-04
Improve geometry documentation on congruence, fixes #2984

### LINKED ISSUE #2984: tgsss1, tgsss2, and tgsss3 are not SSS as intended

david-a-wheeler:
Improve documentation about our approach to congruence in the geometry section.

* We encode congruence using a "distance" function. Make it clear that this encoding is justified by a formal proof, per tgjustc1 and tgjustc2 , which were added via <https://github.com/metamath/set.mm/pull/3014>
* We encode shape congruence as (`cgrG`) as all corresponding segments between all corresponding points are congruent. Make that clear, explain why that's done including its justification in `dfcgrg2` This builds on <https://github.com/metamath/set.mm/pull/2994>.


set.mm
```diff
@@ -381010,7 +381010,15 @@ We intentionally choose to represent congruence (without loss of generality)
   as it is more convenient.  It is always possible to define ` dist ` for any
   particular geometry to produce equal results when conguence is desired, and
   in many cases there is an obvious interpretation of "distance" between
-  two points that can be useful in other situations.  A similar representation
+  two points that can be useful in other situations. Encoding congruence as
+  a distance function makes it easier to use these theorems in cases where
```
benjub, line 381014: Encoding congruence as distance equality makes...
or
Encoding congruence as an equality of distances makes...

set.mm
```diff
@@ -382517,10 +382525,20 @@ proved using the inner form (see theorem 9.6 of [Schwabhauser] p. 69
 
   ${
     $d a b g i j $.
-    $( Define the relation congruence bewteen shapes.  Definition 4.4 of
-       [Schwabhauser] p. 35.  Ideally, we would define this for functions of
-       any set, but we will used words (functions over ` NN ` ) in most cases.
-       (Contributed by Thierry Arnoux, 3-Apr-2019.) $)
+    $( Define the relation congruence between shapes.  Definition 4.4 of
```
benjub, line 382528: the relation OF congruence

set.mm
```diff
@@ -382517,10 +382525,20 @@ proved using the inner form (see theorem 9.6 of [Schwabhauser] p. 69
...
+       [Schwabhauser] p. 35.  Two shapes are congruent if all corresponding
+       segments between all corresponding points are congruent.  Ideally, we
```
benjub, line 382530: I saw in nearby comments the terms "shape" and "sequence". Say somewhere "shapes are finite sequences of points", or maybe simply remove the word "shape" everywhere and only use the term "finite sequence of points" (rather than "word"). @tirix will see what is best.

benjub: Also: maybe write "...if the segments formed on all pairs of corresponding points..." ?

set.mm
```diff
@@ -382517,10 +382525,20 @@ proved using the inner form (see theorem 9.6 of [Schwabhauser] p. 69
...
+       would define this for functions of any set, but we will use words
+       (functions over ` NN ` ) in most cases.  Many books define triangle
+       congruence as requiring both the congruence of all corresponding
+       segments and of all corresponding angles.  However, ~ dfcgrg2 proves
+       that these two definitions are equivalent, so we'll use this definition
+       from [Schwabhauser] p. 35 instead because it's simpler.  Once two
+       triangles are proven congruent, you can use various theorems to prove
+       that corresponding parts of congruent triangles are congruent (CPCTC),
+       for example, see ~ cgr3simp1 , ~ cgr3simp2 , ~ cgr3simp3 , ~ cgrcgra ,
+       permutation laws such as ~ cgr3swap12 , and ~ dfcgrg2 .  (Contributed by
```
benjub, line 382540: I think this is not the place to talk about triangles and even less angles. This definition is about the congruence of general shapes. The two primitive notions in Tarski geometry are betweenness and congruence of segments. Congruence of angles is only a derived notion defined later in terms of it, so it makes no sense to define angle congruence using angles.

set.mm
```diff
@@ -382517,10 +382525,20 @@ proved using the inner form (see theorem 9.6 of [Schwabhauser] p. 69
...
   ${
     $d a b g i j $.
-    $( Define the relation congruence bewteen shapes.  Definition 4.4 of
-       [Schwabhauser] p. 35.  Ideally, we would define this for functions of
-       any set, but we will used words (functions over ` NN ` ) in most cases.
-       (Contributed by Thierry Arnoux, 3-Apr-2019.) $)
+    $( Define the relation congruence between shapes.  Definition 4.4 of
+       [Schwabhauser] p. 35.  Two shapes are congruent if all corresponding
+       segments between all corresponding points are congruent.  Ideally, we
+       would define this for functions of any set, but we will use words
```
benjub, line 382531: It looks like the current definition sees shapes as partial functions on RR.

set.mm
icecream17: congruence relation?

benjub: Probably both are correct, but the former has the advantage of having "congruence between shapes" as a substring.

set.mm
david-a-wheeler: I disagree. Many people are going to expect "shape congruence" to include angle congruence. This definition does *not* include the expected angle congruence, so we need to explain why not. A short justification, with links to relevant proofs, provides that justification.

benjub: Are you sure you're not confusing with Hilbert geometry ?  Hilbert geometry contains angle congruence as a primitive notion, but Tarski geometry does not. These "many books" and "many people" you're talking about are probably developing geometry in other systems than the one used in set.mm. If that is indeed what you mean, then one could write something like:
> Other axiomatic systems of geometry, like Hilbert's system, also have a primitive notion of angle congruence, and therefore, congruence of triangles (which are simply "shapes" with three points) is sometimes defined in these systems using both (segment) congruence and angle congruence.  In Tarski's system, angle congruence is a derived notion, defined in df-.... and theorems ... prove that congruent triangles have congruent angles [and also all the SAS and CPTC stuff].

set.mm
avekens: The section header says "series of points"... I think the term "shape" is more intuitive, so we should keep it, but we should define it as "sequence of points" (see the comment for ~iscgrg), as @benjub proposed

set.mm
avekens: Maybe the additional part of the comment (modified as @benjub proposed) should be moved to the section header.

set.mm
david-a-wheeler: I agree. I slightly prefer the second one, so I'll use that. New version coming soon :-).

set.mm
david-a-wheeler: Sounds good. I'll use "the relation OF congruence between shapes".

set.mm
david-a-wheeler: Good point (hah!). I'll add this text here: **A "shape" is a finite sequence of points.**

set.mm
tirix: The reason for using functions on `RR` was that it could be used to define congruence of more complex shapes, like maybe segments and circles (provided they have the same parametrization. In fact until now, the definition has only been used on triangles represented by their 3 vertices.

benjub: So why not take any common domain ? Or in the other direction (probably to be preferred), why not really take finite sequences i.e. `( 0 ..^ n )` for some `n e. _om` ? That way, you have a theory independent from RR, which is in the spirit of Tarski's FOL axiomatization of geometry (I think models are powers of real closed fields, not necessarily RR ?).

Other question: can you point me to the theorem that says that if the triples (abc) and (def) are congruent and if Babc then Bdef ? That theorem is the crucial one, it even justifies the phrase "congruent" applied to shapes. And from that, you of course obtain angle congruence.

(The idea is that "congruence" of two things means that all primitive notions agree on these two things. And obviously, then all derived notions also agree.)

david-a-wheeler: Can you suggest a better wording?

set.mm
david-a-wheeler: @benjub - I like the idea of your text explaining that the system we're using doesn't have angle congruence as a primitive notion. I'll take your text & modify it to try to make it "flow" in this discussion about definitions.

Even with that statement, though, we need to justify why we (and Tarski) are willing to define congruence of shapes without including congruence of angles. Many typical geometry materials, including those used worldwide in high school, include angles as a primitive fo defining the congruence of triangles. Oh, and don't say high school geometry is irrelevant; Tarski often taught high school geometry In the 1920s and 30s, and this led to his interest in geometry that we're formalizing. This looks like a big deviation, and we can better than an appeal to authority.

As evidence, here are some of sample materials I've found that *expressly* define congruence of triangles as requiring congruence of BOTH their corresponding angles AND their corresponding segments:

* "Two triangles are congruent if a one-to-one correspondence can be set up between their vertices so that corresponding sides are congruent and corresponding angles are congruent." - Euclidean and Non-Euclidean Geometries: Development and History by Marvin Jay Greenberg, 1974, page 69
* "The (example) two triangles are congruent because they are geometrically equivalent. All corresponding sides have equal lengths and all corresponding angles have equal measures." - Advanced Mathematics (Second Edition) by Saxon, Page 67.
* "Two triangles are congruent if their corresponding sides are equal in length, and their corresponding angles are equal in measure." - Wikipedia, [Congruence(geometry)](https://en.wikipedia.org/wiki/Congruence_(geometry)#:~:text=Congruence%20of%20triangles,-See%20also%3A%20Solution&text=Two%20triangles%20are%20congruent%20if,angles%20are%20equal%20in%20measure.)

It's *okay* that we don't include angles in the definition, there are even good reasons for it. We just need a brief comment explaining why why don't. Saying "that's what Tarski did" isn't enough, there's a good reason why Tarski did this that isn't merely an appeal to authority.

A brief note in the header as *well* would make sense, but people who surf this material though hyperlinks may end up on this definition of shape congruence without any context. Let's give them that context, and frankly, the details about shape congruence belong here, in the definition shape congruence.

set.mm
benjub: @david-a-wheeler : if the question is to me: no, I cannot ;)  I think the decision is in @tirix's hands and then he/we'll adapt the text to the definition he chooses.
In any case, I do not know why "ideally" is used: if it were ideal, then it is very easy to make the change. On the other hand, I have not practiced Tarski's geometry enough to know what is gained in considering infinite (and even non-discrete) shapes. Considering infinite shapes removes the feel of simplicity of Tarski's geometry.

david-a-wheeler:
I've posted an update that I *think* addresses all comments here but one. I think it's important to clarify that this definition of shape congruence is different than some others', and provide the rationale.

The exception: "It looks like the current definition sees shapes as partial functions on RR." - do you have any suggested wording?

set.mm
benjub: Thanks for the specific citations in these two books. I'll try to have a look because for now I do not know their axiom systems nor how they define "congruent angles".  By the way, I note that Greenberg indeed uses unordered shapes (i.e. finite sets as opposed to finite sequences), as I proposed in https://github.com/metamath/set.mm/pull/2994#discussion_r1083284350 before I saw that Schwabhauser uses sequences/tuples. I think the unordered shapes are more natural (a triangle does not care about the order in which you enumerate its vertices). But maybe that's too much worl for @tirix to modify it for a too little gain.

The "good reason why Tarski did it" is that angle congruence can be defined in terms of congruence and betweenness alone. So these two primitive notions suffice. This is connected to my question to @tirix in https://github.com/metamath/set.mm/pull/3028#discussion_r1096788692

As for the specific text, I'll let you @david-a-wheeler see, because it looks like you now have all the parts.

david-a-wheeler: I think the ordered sequence representation (words) chosen by @tirix is a good one. If triangle ABC is congruent to DEF, that does *NOT* mean that triangle BAC is necessarily congruent. So while in some sense the "order doesn't matter", when *referring* to a particular shape for making congruence statements, you typically *do* identify the shape by listing a specific order. I view this as a "shorthand for clearly identifying a specific shape".

set.mm
david-a-wheeler: The "Ideally..." text was in the original, I didn't write that (though it makes sense to me). If we don't have any specific changes to that sentence we want to make, I propose we just keep it as-is until we do :-).

tirix: Let me paste Schwabäuser's definition below:
image
Which I would translate to:
> To describe more conveniently several congruences, we introduce the congruence of "geometrical figures" (made of finitely many points)
> 
> Definition: The n-tuples of points $< a_1,...,a_n >$ and $< b_1,...,b_n >$ are said to be congruent when: ... 

Word like `<" A B C ">` are the convenient way in Metamath to express tuples, so if we prefer to restrict ourselves strictly to Schwabhäuser's definition, the domain to be used instead of `RR` would be `NN0`, to match with the domain of words (see [df-word](https://us.metamath.org/mpeuni/df-word.html)). It would be much more convenient that `_om`.

tirix:
> Other question: can you point me to the theorem that says that if the triples (abc) and (def) are congruent and if Babc then Bdef ? That theorem is the crucial one, it even justifies the phrase "congruent" applied to shapes. And from that, you of course obtain angle congruence.

That would be [tgbtwnxfr](https://us.metamath.org/mpeuni/tgbtwnxfr.html), _Theorem 4.6 of [[Schwabhauser](https://us.metamath.org/mpeuni/mmset.html#Schwabhauser)] p. 36._

set.mm
david-a-wheeler: @tirix has responded why, and I think it's sensible.

@benjub : You said, "I think the decision is in @tirix's hands and then he/we'll adapt the text to the definition he chooses." Since this is the definition he chose, and has reaffirmed it here, I *think* this comment is resolved, so I'm resolving it. If I'm wrong please let us know :-).

set.mm
benjub: That's why the definition that Greenberg and I (in https://github.com/metamath/set.mm/pull/2994#discussion_r1083284350) proposed are formulated using a bijection. A triangle is defined by a set of three points, not a sequence of three points.

set.mm
benjub: Yes, I meant NN0 sorry. So the definition should be
```
... { <. a , b >. |
E. n e. NN0 ( ( a : ( 0 ..^ n ) --> ( Base ` g ) /\ b : ( 0 ..^ n ) ->  ( Base ` g ) ) /\ ...
```

benjub:
> > Other question: can you point me to the theorem that says that if the triples (abc) and (def) are congruent and if Babc then Bdef ? That theorem is the crucial one, it even justifies the phrase "congruent" applied to shapes. And from that, you of course obtain angle congruence.
> 
> That would be [tgbtwnxfr](https://us.metamath.org/mpeuni/tgbtwnxfr.html), _Theorem 4.6 of [[Schwabhauser](https://us.metamath.org/mpeuni/mmset.html#Schwabhauser)] p. 36._

Thanks @tirix. That theorem, via its corollary:

If $(a_0, \dots, a_{n-1}) \equiv (b_0, \dots, b_{n-1})$ and $0 \leq i, j, k < n$, then $B a_i a_j a_k$ iff $B b_i b_j b_k$.

is fundamental. If it were not true, then it should be added to the definition of congruence of shapes. I don't know how you emphasized it in comments, but it is really important.

(The same thing could be said of equality: If $(a_0, \dots, a_{n-1}) \equiv (b_0, \dots, b_{n-1})$ and $0 \leq i, j < n$, then $a_i = a_j$ iff $b_i = b_j$. This is true by one of the first axioms of Tarski geometry, but again if there was no such axiom, then that property should be added to the definition of congruence.)

set.mm
david-a-wheeler: > I'll try to have a look because for now I do not know their axiom systems nor how they define "congruent angles". 

The *usual* definition is, I believe, "congruent angles have equal measures".

In any case, I hope I've convinced you that there are *other* systems that regularly refer to "congruent angles" when they define "congruent shapes/triangles". All I want to do is provide a brief justification for the decision to *not* do that here, along with links to relevant proofs. It's not a change in the definition, but an explanation to help the reader understand why this specific definition is used.

david-a-wheeler:
I *think* I've addressed all comments/issues. If not, or if there are new issues, please raise them!

benjub (approved):
Good for me, though it might be wise to wait for @tirix's approval.

set.mm
```diff
@@ -382517,9 +382532,29 @@ proved using the inner form (see theorem 9.6 of [Schwabhauser] p. 69
...
+       that these two definitions are equivalent, and using the Tarski
+       definition instead (given in [Schwabhauser] p. 35) is simpler.  Once two
+       triangles are proven congruent as defined here, you can use various
+       theorems to prove that corresponding parts of congruent triangles are
+       congruent (CPCTC).  For example, see ~ cgr3simp1 , ~ cgr3simp2 ,
+       ~ cgr3simp3 , ~ cgrcgra , and permutation laws such as ~ cgr3swap12 and
+       ~ dfcgrg2 .
+
+       Ideally, we would define this for functions of any set, but we will use
+       words (functions over ` NN ` ) in most cases.
```
tirix, line 382556: Actually, words are defined as functions over `NN0`, not over `NN`.

set.mm
benjub: Then what is the "measure" of an angle ! This is actually trickier than one may think. (Actually, the "measure" of an angle is often defined to be its equivalence class for the angle congruence relation, so it's the other way around; it's similar to a "distance" being defined as an equivalence class for the segment congruence relation.)

I know there are other systems that have angle congruence as a primitive notion (I mentioned Hilbert's as such a system above); my point is that they are *other* systems. I agree with the explanation that you now wrote (my only problem with the first version was that the distinction between the different systems was not clear).

set.mm
benjub: well... functions over `( 0 ..^ n )` for some n \in NN0 !

tirix: Indeed! Maybe just refer to ~ df-word ?
Something like 
> but we will use words (see ~ df-word) in most cases.

david-a-wheeler: I like that, it's both clear and accurate (which is often hard to do). I'll post an update with that.

set.mm
david-a-wheeler: @benjub :
> I agree with the explanation that you now wrote (my only problem with the first version was that the distinction between the different systems was not clear).

Excellent! Good to hear, I think the results are better than my first try for sure. Thank you! Given what you just said, I'm marking this as "resolved".

david-a-wheeler:
I think we're done!! Any more comments? Otherwise I think it's ready to merge.

avekens (approved):

set.mm
avekens: > Let me paste Schwabäuser's definition below: image Which I would translate to:
> 
> > To describe more conveniently several congruences, we introduce the congruence of "geometrical figures" (made of finitely many points)
> > Definition: The n-tuples of points <a1,...,an> and <b1,...,bn> are said to be congruent when: ...
> 

The German "geometrische Figuren" could also be translated to "geometric shapes", which fits to our chosen terminology.
Furthermore, "Streckenkongruenz" does not only mean "congruence ", but "congruence of distances" - also perfectly adopted in our concepts/comments. So everything is fine now!

tirix (approved):
Looks good to me! Thank you David!

david-a-wheeler:
Great!  And @tirix - thanks to *you*. You did a lot of amazing things, I'm just trying to help people better understand all this great work.

david-a-wheeler:
@tirix @benjub - thank you both!!

PR #3030, opened 2023-02-07
More iset.mm structures, from 2strstr1 to lmodvsca

jkingdon:
Because this is a long section, I'll break it apart into multiple pull requests.

Although most theorems need to be changed (mostly for set existence conditions), there is nothing particularly difficult here.

I needed tpid1g and tpid2g (which are just like the existing https://us.metamath.org/mpeuni/tpid3g.html but for the first and second elements of the ordered triple). Since they were in a mathbox in set.mm, I moved them to main in both set.mm and iset.mm.

There's a new theorem `opelstrsl` which is like https://us.metamath.org/mpeuni/opelstrbas.html which we've had for over a year, but the new one is for any slot.

There are a few tweaks to set.mm to fix places where it said "function" and meant "structure".

The new theorems use the ``E = Slot ( E ` ndx ) /\ ( E ` ndx ) e. NN`` idiom that we've been talking about, where appropriate.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 5-Feb-23 srngfn    srngstr
 30-Jan-23 elsb4lem  elsb4v
 30-Jan-23 elsb3lem  elsb3v
 30-Jan-23 equsb3lem equsb3v     equsb3lem was a duplicate of equsb3v
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -99,6 +99,7 @@
 "basendx" is used by "2stropg".
 "basendx" is used by "2strstr1g".
 "basendx" is used by "2strstrg".
+"basendx" is used by "lmodstrd".
 "basendx" is used by "rngstrg".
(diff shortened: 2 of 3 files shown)


iset.mm
```diff
@@ -126955,6 +126982,283 @@ could not be used in an extensible structure (slots must be positive
...
+     14-Aug-2015.) $)
+  mulrndx $p |- ( .r ` ndx ) = 3 $=
+    ( cmulr c3 df-mulr 3nn ndxarg ) ABCDE $.
+
+  $( Utility theorem: index-independent form of ~ df-mulr .  (Contributed by
+     Mario Carneiro, 8-Jun-2013.) $)
+  mulrid $p |- .r = Slot ( .r ` ndx ) $=
+    ( cmulr c3 df-mulr 3nn ndxid ) ABCDE $.
+
+  $( Slot property of ` +g ` .  (Contributed by Jim Kingdon, 3-Feb-2023.) $)
```
avekens, line 127055: Slot property of ` .r ` .

jkingdon: Fixed.

avekens (approved):

PR #3031, opened 2023-02-08
reduce axioms in moanimv

wlammen:
ax-7 and ax-12 are removed.

This time I factored out common proof lines in moanimv and moanim, considerably reducing the number of proof lines in total, but of course, at the expense of an extra lemma.  Let me know what you think, and how credits to the factored out proof should be handled

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16642,6 +16642,7 @@ New usage of "mndomgmid" is discouraged (3 uses).
 New usage of "mo3OLD" is discouraged (1 uses).
 New usage of "mo4fOLD" is discouraged (0 uses).
 New usage of "moabsOLD" is discouraged (0 uses).
+New usage of "moanimvOLD" is discouraged (0 uses).
 New usage of "mobiOLD" is discouraged (0 uses).
 New usage of "mobidOLD" is discouraged (0 uses).
 New usage of "mobidvALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


avekens (approved):

avekens:
The lemma is fine for me. The credit is also OK, but "(Contributed by NM, 23-Mar-1995.)  (Revised by Wolf Lammen, 8-Feb-2023.)" would also be fine.

wlammen:
I checked the state of 23-Dec-2018 out from git history, and the following was NM 's version before I rewrote the proof on 23/24 Dec 2018 completely.  Based on this I will leave the tag as is.  My question was:  Is there a procedure at hand how to credit a proof extracted from other sources, in particular if the author cannot be determined.
```
${
    $d x y $.  $d y ph $.  $d y ps $.
    moanim.1 $e |- F/ x ph $.
    $( Introduction of a conjunct into "at most one" quantifier.  (Contributed
       by NM, 3-Dec-2001.) $)
    moanim $p |- ( E* x ( ph /\ ps ) <-> ( ph -> E* x ps ) ) $=
      ( vy wa weq wi wal wex wmo impexp albii 19.21 bitri nfv mo2 imbi2i 19.37v
      exbii bitr4i 3bitr4i ) ABFZCEGZHZCIZEJABUDHZCIZHZEJZUCCKABCKZHZUFUIEUFAUG
      HZCIUIUEUMCABUDLMAUGCDNOTUCCEUCEPQULAUHEJZHUJUKUNABCEBEPQRAUHESUAUB $.
$}
```

avekens:
> I checked the state of 23-Dec-2018 out from git history, and the following was NM 's version before I rewrote the proof on 23/24 Dec 2018 completely. Based on this I will leave the tag as is. My question was: Is there a procedure at hand how to credit a proof extracted from other sources, in particular if the author cannot be determined.

I do not know if there is already a strict rule or convention how to handle such a case, but I would suggest the following rule of thump:

- if there is only a "Contributed by" tag, but no "Revised by" or "Proof shortened tag", then the original contributer should be the contributor of the theorem representing an extracted proof (if it is not modified significantly)
- otherwise, the one who extracted the proof should be the contributor of the corresponding theorem.

PR #3032, opened 2023-02-09
minimize with nexmo, factor out proof skeleton in sb8eu*

wlammen:
1. nexmo was introduced half a year ago and wasn't available when the shortened proofs were written.  Since this is a typical minimization process, no tags were updated.
2. sb8eu and sb8euv share most of the proof.  Factor out common proof lines.
3. Use the same tagging strategy in moanimlem.

br1steqgOLD and br2steqgOLD are  two theorems in SF's mathbox, created on 9-Feb-2022.  Are these covered by the 1-year rule for automatic deletion, too?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14275,8 +14275,6 @@ New usage of "bnj999" is discouraged (1 uses).
 New usage of "bnnv" is discouraged (7 uses).
 New usage of "bnrel" is discouraged (1 uses).
 New usage of "bnsscmcl" is discouraged (0 uses).
-New usage of "br1steqgOLD" is discouraged (0 uses).
-New usage of "br2ndeqgOLD" is discouraged (0 uses).
 New usage of "bra0" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


benjub:
> br1steqgOLD and br2steqgOLD are two theorems in SF's mathbox, created on 9-Feb-2022. Are these covered by the 1-year rule for automatic deletion, too?

Yes, since they have current (improved) versions.

benjub:
Which theorems did you try and minimize_with `nexmo` ?  If you haven't done so, you could try `scripts/min.cmd` on theorems using `df-mo` recursively (`MM> sh us df-mo/REC`).

wlammen:
I did not use minimize_with, but looked into the theorems referencing exmo and identified those that allowed a shortening by nexmo.

wlammen:
For tagging sb8eulem see also discussion in #3031

benjub (approved):

Issue #2995, opened 2023-01-19
Forget the name "generalized Euclidean space"


benjub:
I was reminded of the name "generalized Euclidean space" in https://github.com/metamath/set.mm/pull/2993#discussion_r1080671402. The definition is at https://us.metamath.org/mpeuni/df-rrx.html

To me the best name to call ``( RR^ ` I )`` is simply "``( RR^ ` I )``" (and if you have to pronounce it: "R to the I"). First, "generalized euclidean space" is nonstandard.  I never heard of it with that meaning (which is no proof it's nonstandard of course, but an indication).  It is probably used in multiple places, and my guess is that it is given various different meanings, but generally not the present one, for a reason I now explain.

The space ``( RR^ ` I )`` is a specific and concrete one (and this is especially true in set.mm where everything is encoded in set theory: this is a specifc set of specific sets of... down to \varnothing). On the other hand, definitions of Euclidean spaces, Banach spaces (or really mostly anything: abelian group, polish space, distributive lattice, affine scheme, Cauchy sequence...) are about objects satisfying some properties, rather than specific objects. That is, these definitions generally begin like "a ...... space is a (real) vector space such that.....".

As an illustration of this, you can point at a specific nonzero element of RR^I (provided that you can point at a specific element of I), like (1, 0, 0, ...), but you cannot point at a specific nonzero vector of B if I just tell you that B is a real Banach space. Still in other words (but this is linked to the previous illustration), anyone reasonable would hope that a space which is isomorphic (in the appropriate category, here the category of real prehilbertian spaces with obvious morphisms) to a "generalized Euclidean space" is also a "generalized Euclidean space". But a real prehilbertian space isomorphic to ``( RR^ ` I )`` need not be an ``( RR^ ` J )`` (for a rather silly reason: just define your favorite "reals" to be of the form (0, x) where x is a real, and transport all the structures to that new set).

Lastly, of course there are names for specific famous objects, like *the* Cantor set or *the* Klein group (or bottle) or *the* Hopf fibration, etc.  We cannot here christen some specific objects with new names like was done for these very standard ones.

tirix:
I agree with Benoît, and as a matter of fact Wikipedia also makes a clear distinction:
- [_Real Coordinate Space_](https://en.wikipedia.org/wiki/Real_coordinate_space) denotes the specific construct, with ``( RR^ ` n )`` as a notation
[...]

PR #3034, opened 2023-02-11
Lines and spheres in generalized real Euclidean spaces

avekens:
New definition and related basic theorems for lines and spheres in generalized real Euclidean spaces are provided, with theorem ~ itsclinecirc0 (the intersection points of a line through two points and a circle around the origin) as main result.

The names/symbols/labels should be changed in the context of issue #2995 in the future.

Details:

main set.mm:
* ~rrxmetfi, ~rrxdsfi, ~rrxbasefi, ~addgtge0d, ~addeq0, ~assraddsubd, ~xpexd, ~fvmptd3, ~eqeqan1d, ~orcomdd  moved from mathboxes
* theorem ~fz12pr added
* theorems ~frlmvplusgvalc, ~frlmplusgvalb, ~frlmvscavalb, ~frlmvplusgscavalb (coordinate-wise operations in free left modules) added
* theorems for generalized Euclidean spaces (~rrxplusgvscavalb, ~rrxsca, ~rrxdsfival, ~ehleudis, ~ehleudisval, ~ehl1eudis, ~ehl1eudisval, ~ehl2eudis, ~ehl2eudisval) added
* theorems for affine combinations (~affineequiv3, ~affineequiv4, ~affineequivne) added

AV's mathbox:
* auxiliary theorems ~fv1prop, ~fv2prop, ~mapfvd, ~submuladdmuld, ~affinecomb1, ~affinecomb2, ~affineid,  ~resum2sqcl, ~resum2sqgt0 added
* definition and related basic theorems for lines and spheres in generalized real Euclidean spaces
* theorems for the intersection points of a line through two points and a circle around the origin.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,11 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+16-Feb-23 subdivcomb2 [same]    moved from SF's mathbox to main set.mm
+16-Feb-23 negelrpd  [same]      moved from GS's mathbox to main set.mm
+16-Feb-23 1xr       [same]      moved from GS's mathbox to main set.mm
+16-Feb-23 rexlimdva2 [same]     moved from GS's mathbox to main set.mm
+16-Feb-23 mvrladdd  [same]      moved from DAW's mathbox to main set.mm
 10-Feb-23 rrxmetfi  [same]      moved from GS's mathbox to main set.mm
 10-Feb-23 rrxdsfi   [same]      moved from GS's mathbox to main set.mm
 10-Feb-23 rrxbasefi [same]      moved from GS's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,10 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+16-Feb-23 bj-ldiv   ldiv        moved from BJ's mathbox to main set.mm
+16-Feb-23 bj-rdiv   rdiv        moved from BJ's mathbox to main set.mm
+16-Feb-23 bj-mdiv   mdiv        moved from BJ's mathbox to main set.mm
+16-Feb-23 bj-lineq  lineq       moved from BJ's mathbox to main set.mm
 16-Feb-23 subdivcomb2 [same]    moved from SF's mathbox to main set.mm
 16-Feb-23 negelrpd  [same]      moved from GS's mathbox to main set.mm
 16-Feb-23 1xr       [same]      moved from GS's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,16 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+10-Feb-23 rrxmetfi  [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 rrxdsfi   [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 rrxbasefi [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 addgtge0d [same]      moved from AI's mathbox to main set.mm
+10-Feb-23 addeq0    [same]      moved from TA's mathbox to main set.mm
+10-Feb-23 assraddsubd [same]    moved from DAW's mathbox to main set.mm
+10-Feb-23 xpexd     [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 fvmptd3   [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 eqeqan1d  [same]      moved from PM's mathbox to main set.mm
+10-Feb-23 orcomdd   [same]      moved from GM's mathbox to main set.mm
  5-Feb-23 srngfn    srngstr
 30-Jan-23 elsb4lem  elsb4v
 30-Jan-23 elsb3lem  elsb3v
--- a/discouraged
+++ b/discouraged
@@ -15437,6 +15437,7 @@ New usage of "enrer" is discouraged (8 uses).
 New usage of "enrex" is discouraged (9 uses).
 New usage of "eqeq1dALT" is discouraged (0 uses).
 New usage of "eqeqan12dALT" is discouraged (0 uses).
+New usage of "eqeqan1dOLD" is discouraged (0 uses).
 New usage of "eqid1" is discouraged (0 uses).
 New usage of "eqresr" is discouraged (4 uses).
 New usage of "eqsbc3rVD" is discouraged (0 uses).
@@ -19091,6 +19092,7 @@ Proof modification of "en3lplem1VD" is discouraged (95 steps).
 Proof modification of "en3lplem2VD" is discouraged (267 steps).
 Proof modification of "eqeq1dALT" is discouraged (62 steps).
 Proof modification of "eqeqan12dALT" is discouraged (23 steps).
+Proof modification of "eqeqan1dOLD" is discouraged (21 steps).
 Proof modification of "eqid1" is discouraged (9 steps).
 Proof modification of "eqsbc3rVD" is discouraged (131 steps).
 Proof modification of "equcomi1" is discouraged (16 steps).
--- a/discouraged
+++ b/discouraged
@@ -15438,6 +15438,7 @@ New usage of "enrex" is discouraged (9 uses).
 New usage of "eqeq1dALT" is discouraged (0 uses).
 New usage of "eqeqan12dALT" is discouraged (0 uses).
 New usage of "eqeqan1dOLD" is discouraged (0 uses).
+New usage of "eqeqan1dOLDOLD" is discouraged (0 uses).
 New usage of "eqid1" is discouraged (0 uses).
 New usage of "eqresr" is discouraged (4 uses).
 New usage of "eqsbc3rVD" is discouraged (0 uses).
@@ -17080,6 +17081,7 @@ New usage of "opsqrlem6" is discouraged (0 uses).
 New usage of "opthregOLD" is discouraged (0 uses).
 New usage of "orbi1r" is discouraged (0 uses).
 New usage of "orbi1rVD" is discouraged (0 uses).
+New usage of "orcomdd" is discouraged (0 uses).
 New usage of "ordelordALT" is discouraged (0 uses).
 New usage of "ordelordALTVD" is discouraged (0 uses).
 New usage of "ordpinq" is discouraged (6 uses).
@@ -19092,7 +19094,8 @@ Proof modification of "en3lplem1VD" is discouraged (95 steps).
 Proof modification of "en3lplem2VD" is discouraged (267 steps).
 Proof modification of "eqeq1dALT" is discouraged (62 steps).
 Proof modification of "eqeqan12dALT" is discouraged (23 steps).
-Proof modification of "eqeqan1dOLD" is discouraged (21 steps).
+Proof modification of "eqeqan1dOLD" is discouraged (12 steps).
+Proof modification of "eqeqan1dOLDOLD" is discouraged (21 steps).
 Proof modification of "eqid1" is discouraged (9 steps).
 Proof modification of "eqsbc3rVD" is discouraged (131 steps).
 Proof modification of "equcomi1" is discouraged (16 steps).
@@ -19671,6 +19674,7 @@ Proof modification of "opnmblALT" is discouraged (332 steps).
 Proof modification of "opthregOLD" is discouraged (112 steps).
 Proof modification of "orbi1r" is discouraged (9 steps).
 Proof modification of "orbi1rVD" is discouraged (101 steps).
+Proof modification of "orcomdd" is discouraged (13 steps).
 Proof modification of "ordelordALT" is discouraged (128 steps).
 Proof modification of "ordelordALTVD" is discouraged (202 steps).
 Proof modification of "orim12dALT" is discouraged (34 steps).


set.mm
```diff
@@ -321668,6 +321874,173 @@ need not be complete (for instance if the given set is infinite
...
   $}
 
+  ${
+    $d B f $.  $d X f $.  $d f ph $.
+    rrxbasefi.x $e |- ( ph -> X e. Fin ) $.
+    rrxbasefi.h $e |- H = ( RR^ ` X ) $.
+    rrxbasefi.b $e |- B = ( Base ` H ) $.
+    $( The base of the generalized real Euclidean space, when the dimension of
+       the space is finite.  This justifies the use of ` ( RR ^m X ) ` for the
+       development of the Lebeasgue measure theory for n-dimensional Real
```
tirix, line 321884: Lebeasgue -> Lebesgue

set.mm
```diff
@@ -730651,6 +730957,1157 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+    $d p t w x y $.
+    $( Definition of lines passing through two different points in a left
+       module (or any extended structure having a base set, an addition, and a
+       scalar multiplication).  (Contributed by AV, 14-Jan-2023.) $)
+    df-line $a |- LineM = ( w e. _V |-> ( x e. ( Base ` w ) ,
+                                         y e. ( ( Base ` w ) \ { x } )
+                    |-> { p e. ( Base ` w ) | E. t e. ( Base ` ( Scalar ` w ) )
+                          p = ( ( ( ( 1r ` ( Scalar ` w ) )
+                                    ( -g ` ( Scalar ` w ) ) t ) ( .s ` w ) x )
+                                ( +g ` w ) ( t ( .s ` w ) y ) ) } ) ) $.
```
tirix, line 731122: This definition is very similar to the `LineG` slot of [``( toTG ` w)``](https://us.metamath.org/mpeuni/ttgval.html). Is the only difference the fact that it uses a generic scalar, while `toTG` assume that the scalars are reals?

It appears that just a few theorems down the line, you are switching back to real scalars only.

set.mm
```diff
@@ -730651,6 +730957,1157 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  $}
+
+  ${
+    $d p r w x y $.
+    $( Definition of spheres for given centers and radii in a metric space (or
+       more general, in a distance space, see ~ distspace , or eben in any
```
tirix, line 731128: more general > more generally
eben > even

set.mm
```diff
@@ -7594,6 +7594,14 @@ sufficient to develop all propositional calculus of the logic (together with
       ( wo pm1.4 syl ) BAEABECBAFDG $.
   $}
 
+  ${
+    orcomdd.1 $e |- ( ph -> ( ps -> ( ch \/ th ) ) ) $.
+    $( Commutativity of logic disjunction, in double deduction form.
+       (Contributed by Giovanni Mascellani, 19-Mar-2018.) $)
+    orcomdd $p |- ( ph -> ( ps -> ( th \/ ch ) ) ) $=
+      ( wo pm1.4 syl6 ) ABCDFDCFECDGH $.
+  $}
```
benjub, line 7603: This is simply orcomd (up to writing "ph /\ ps") and not much used, so I would not add it.

set.mm
```diff
@@ -24853,6 +24861,19 @@ replaced with setvar variables (see ~ cleljust ), so we don't include
       ( wceq wb eqeqan12d ancoms ) ABCEIDFIJABCDEFGHKL $.
   $}
 
+  ${
+    eqeqan1d.1 $e |- ( ph -> A = B ) $.
+    $( Implication of introducing a new equality.  (Contributed by Peter Mazsa,
+       17-Apr-2019.)  (Proof shortened by AV, 10-Feb-2023.) $)
+    eqeqan1d $p |- ( ( ph /\ C = D ) -> ( A = C <-> B = D ) ) $=
+      ( wceq id eqeqan12d ) ADEGZBCDEFJHI $.
```
benjub, line 24869: It has only two usages, so probably does not pay for itself.
eqeqan12d looks sufficient. Actually, what could be useful is an version of eqeqan12d with the same antecedent.

set.mm
```diff
@@ -730651,6 +730957,1157 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+        FBFXJIKVOUPXFYGSNYMBFGIKVOVKVLZVNZVPUUEYEYIUUGUUIUUEYDUUFVQUUEYHUUHVQVR
+        VSVTXGUUDXFYMXGXOUUDXTXOUUBUUCCWACWBVSVGVEYJYKWCWDYNYOXLCYNXLYOYNYMXFPX
+        LYORXHXFYMXQWEXKEXJGBEFWFQZTWGQZJUUKUKTWNOZWFQZUUJTWONUUKUUMRWHUUKTUUKU
+        QWIWJUULFWFUULXPFWKIWLWMXDWLBSFWPOSXPWPOKFXPSWPIWQXDYAWRVGWSWTXAXBXCXE
+        $.
+    $}
+
+    ${
+      $d .0. p $.
+      2sphere0.0 $e |- .0. = ( I X. { 0 } ) $.
```
tirix, line 731506: I stopped on this formulation for a second, it is not very usual.
I was about to suggest `<" 0 0 ">`, but then I realized that indices do not match (0-based vs. 1-based). 
Maybe a comment would help?
Like:
> `( I X. { 0 } )` represents the constant function equal to `0` for both indice values `1` and `2`, i.e. the origin.

set.mm
```diff
@@ -116418,6 +116471,13 @@ used to convert hypothesis of the inference (deduction) form of this
     wceq ) AEFBEFGZCEFDEFGZGABHIZJZCDHIZJZRZUAUCRZBAHIZDCHIZRSUAEFUCEFUEUFKTABL
     CDLUAUCMNSTUBUGUDUHABOCDOPQ $.
 
+  $( Two complex numbers which add up to zero are each other's opposites.
```
benjub, line 116474: Two complex numbers add up to zero iff they are each other's opposites.

set.mm
```diff
@@ -133784,6 +133854,12 @@ subset of a (possibly extended) finite sequence of integers.  (Contributed
       PVBUMTUN $.
   $}
 
+  $( An integer range between 1 and 2 is a pair.  (Contributed by AV,
+     11-Jan-2023.) $)
+  fz12pr $p |- ( 1 ... 2 ) = { 1 , 2 } $=
+    ( c1 c2 cfz co caddc cpr df-2 oveq2i cz wcel wceq ax-mp 1p1e2 preq2i 3eqtri
+    1z fzpr ) ABCDAAAEDZCDZARFZABFBRACGHAIJSTKPAQLRBAMNO $.
```
benjub, line 133861: For this kind of pairs (and in the applications below), I would use 0 and 1 instead of 1 and 2.

tirix (approved):
Wow! Very impressive results!
What is your project here? What will come next?

set.mm
```diff
@@ -262824,6 +262922,76 @@ the ring of the coefficients is not a division ring (not a field in the
...
   $}
 
+  ${
+    $d I i $.  $d X i $.  $d Z i $.  $d ph i $.
+    frlmplusgvalb.f $e |- F = ( R freeLMod I ) $.
+    frlmplusgvalb.b $e |- B = ( Base ` F ) $.
+    frlmplusgvalb.i $e |- ( ph -> I e. W ) $.
+    frlmplusgvalb.x $e |- ( ph -> X e. B ) $.
+    frlmplusgvalb.z $e |- ( ph -> Z e. B ) $.
+    frlmvscavalb.r $e |- ( ph -> R e. Ring ) $.
```
benjub, line 262932: label typo ?

set.mm
```diff
@@ -262804,6 +262880,28 @@ the ring of the coefficients is not a division ring (not a field in the
...
+    frlmvplusgvalc.b $e |- B = ( Base ` F ) $.
+    frlmvplusgvalc.r $e |- ( ph -> R e. V ) $.
+    frlmvplusgvalc.i $e |- ( ph -> I e. W ) $.
+    frlmvplusgvalc.x $e |- ( ph -> X e. B ) $.
+    frlmvplusgvalc.y $e |- ( ph -> Y e. B ) $.
+    frlmvplusgvalc.j $e |- ( ph -> J e. I ) $.
+    frlmvplusgvalc.a $e |- .+ = ( +g ` R ) $.
+    frlmvplusgvalc.p $e |- .+b = ( +g ` F ) $.
+    $( Addition in a free module at a coordinate.  (Contributed by AV,
+       16-Jan-2023.) $)
```
benjub, line 262894: Coordinates of a sum with respect to a basis in a module.

set.mm
```diff
@@ -355904,6 +356277,35 @@ line segment AB (the hypotenuse), and ` O ` is the signed right angle
       KZEDBMKZJKZOUJUIMKZUJUKMKZOCBMKZUHUJJKZOABCDEFGHIPAUJUIUKADBHFQZADCHGQAEU
       JIUPRSAULUNUMUOADBCHFGTAUOLUJJKZUKMKUMALEUJAUAIUPUBAUQUJUKMAUJUPUCUDUEUFU
       G $.
+
+    $( Equivalence between two ways of expressing ` A ` as an affine
+       combination of ` B ` and ` C ` .  (Contributed by AV, 22-Jan-2023.) $)
+    affineequiv3 $p |- ( ph -> ( A = ( ( ( 1 - D ) x. B ) + ( D x. C ) )
+                                 <-> ( A - B ) = ( D x. ( C - B ) ) ) ) $=
```
benjub, line 356284: I remember I did some things like that in my mathbox. Have you had a look to see what's reusable?

set.mm
```diff
@@ -730651,6 +730957,1157 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  csph $a class Sphere $.
+
+  ${
+    $d p t w x y $.
+    $( Definition of lines passing through two different points in a left
+       module (or any extended structure having a base set, an addition, and a
+       scalar multiplication).  (Contributed by AV, 14-Jan-2023.) $)
+    df-line $a |- LineM = ( w e. _V |-> ( x e. ( Base ` w ) ,
+                                         y e. ( ( Base ` w ) \ { x } )
+                    |-> { p e. ( Base ` w ) | E. t e. ( Base ` ( Scalar ` w ) )
```
benjub, line 731119: It does not seem good to have the set for the second argument depending on the value of the first. I would use
```
x e. Base w , y e. Base w |-> if ( x = y , ... , ... )
```
where the ... is some default value, probably `Base w` so that the value of the function is always: the set of z such that x,y,z are collinear.

set.mm
```diff
@@ -730651,6 +730957,1157 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  $}
+
+  ${
+    $d p r w x y $.
+    $( Definition of spheres for given centers and radii in a metric space (or
+       more general, in a distance space, see ~ distspace , or eben in any
+       extended structure having a base set and a distance function into the
+       real numbers.  (Contributed by AV, 14-Jan-2023.) $)
+    df-sph $a |- Sphere = ( w e. _V |-> ( x e. ( Base ` w ) ,
+                                          r e. ( 0 [,] +oo )
```
benjub, line 731132: I would use `r e. RR*` since this could make some future uses easier.

benjub (dismissed):
There are several things in this PR, including several things that are proposed to be added directly to the main part without first maturing in a mathbox. So I think we could allow some time to resolve some issues here before merging.

set.mm
tirix: For reference, I could see this used at least in [`df-subma`](https://us.metamath.org/mpeuni/df-subma.html) and [`df-nbgr`](https://us.metamath.org/mpeuni/df-nbgr.html) (both by @avekens!)

I assume you would put the empty set `(/)` in the "else" part of the "if". That would make the definition specific to our encoding of "out-of-bounds behaviour". The proposed solution does not have that drawback and is shorter, IMO.

tirix: Oops, I missed the second part of your comment!

avekens: I think it was @digama0 who once proposed the style having a difference with a singleton. Furthermore, there are more examples of this kind (e.g. df-trkg by @tirix ;-) ). Finally, we have ~cbvmpt2x which allows the second class to be dependent on the first set variable.

Using this style, the if statement can be avoided, giving the same result (if x = y, then  ` ( x ( LineM ``  w ) y = (/)`in both cases.

avekens:
@benjub @tirix tanks a lot for your comments. Since there are many remarks, it will take a while until I can consider/comment on them.

set.mm
avekens: This is a general issue: I just moved this theorem from a mathbox to main, so it is not added (does not increase the size of set.mm). There is a rule that "If you need a theorem from another mathbox, typically it is moved to the main part of set.mm.".  I obeyed this rule for a long time, and thought at least Norm did support/encourage this. This is, in my opinion, why the option /i (or /INCLUDE_MATHBOXES) is available for the minimize command. Since my mathbox is nearly at the end of set.mm, I often detect cases where my proofs can be shortened by using theorems of other's mathboxes, using this option. Then I move these to main according to the above mentioned rule. I do not want to spend a lot of time to investigate if these theorems pay off in the usual way how new theorems are rated. Only in actually obscure cases I revert such minimizations.

If this proceding is not acceptable (anymore), I will not use option /I in the future anymore, resulting in no or at least less theorems being moved from mathboxes to main. @david-a-wheeler @glacode @giomasce what do you think?

set.mm
avekens: see my comment above

set.mm
avekens: This theorem is used for the case ``( EEhil ` 2 ) = ( RR^ ` ( 1 ... 2 ) ) = ( RR^ ` { 1 , 2 })``. In contrast, ``( EEhil ` 0 ) = ( RR^ ` (/) )`` should be a space of dimension 0, consisting of one (neutral) element only, at least according to the comment for ~df-ehl ("The case ` n = 0 ` corresponds to a space of dimension 0, that is, limited to a neutral element.") This is, however, not proven anywhere, is it? Therefore, the range cannot be changed without any major impacts related to the definition of ~df-ehl.

set.mm
benjub: As I wrote above, choose a definition which is uniform in all cases, including when x=y. I gave one possiblity above: the set of z's such that x,y,z are collinear.

Actually, it would be better to define the "affine hull" of a family of points in all generality. Also, you defined it in terms of a particular parameterization (that is, the set of `(1-t)X+tY` for real t), which should be a property, not a definition, because it is not symmetric in x and y and has a different form for more points. The affine hull of a subset X of an R-affine space A is the set of (finite) linear combinations of elements of X such that the coefficients sum to $1_R$.

I have only one theorem on barycentric coordinates https://us.metamath.org/mpeuni/bj-bary1.html in the section "Affine geometry" but I think it gives the idea.

Then, if you consider spheres (so you leave affine geometry) in Euclidean spaces, you may generalize your results to the intersection of a max-dimensional sphere and an affine subspace. This is roughly the result that the metric induced on an affine subspace by a Euclidean metric is Euclidean.

set.mm
glacode: I agree, and that's my approach. When mmj2 grabs (by automation or search) a theorem from another mathbox, I first prepare a PR for the moveToMainPart (of all the needed theorems), and then a second PR for mathbox only stuff.

I guess that yamma (or other tools), at some point, could have a $allowmathboxes directive and/or a diagnostic warning/error if the diagnostic is missing, and you're using a theorem from another mathbox.
But I don't see this being a priority.

set.mm
avekens: You are right, there is a relatinship between both concepts. The main difference is that LineG is a slot only (which may contain anything), whereas LineM defines lines themselves. Another important difference is that LineM is based on properties of left modules, whereas LineG is based on betweenness. Nevertheless, looking at theorem ~elntg, something like `` ( LineG ` ( EEG ` N ) ) = ( LineM ` ( EEhil ` N ) ) `` or `` ( LineG ` ( toTG ` ( EEhil ` N ) ) ) = ( LineM ` ( EEhil ` N ) )`` should be provable.
Then the definition of LineM could have been (or is equivalet to) `` LineM = ( w e. _V |-> ( LineG ` ( toTG ` w ) )``.

set.mm
avekens: Is there actually a meaningful application for a sphere with negativ radius? If not, is it really easier to cope with `` ( M ( Sphere ` E ) R ) = (/) `` if ` R e. RR* ` and ` R < 0 ` instead of assuming ` 0 <_ R ` explicitly?

set.mm
avekens: I actually had a look at ~bj-lineq, and there are good chances that it could be used within the proof of ~itsclc0qlem1. But ~bj-lineq depends on ~bj-rdiv which depends on bj-ldiv which is in a block together with ~bj-mdiv, so it would be necessary to move these theorems, too. Then it would be adequate to move all theorems of this section ""Complex numbers (supplements)" to main, isn't it? Where would you propose to place these theorems?

~bj-lineq, however, is currently only used in the proof of ~bj-lineqi, which itself is not used. Therefore, the only usage of it would be in the proof of ~itsclc0qlem1.

Then I had a look at ~bj-bary1, which is based on ~bj-subcom and ~bj-rdiv at the end. I wonder if it can be used to shorten the proof of ~affinecomb1 (instead of using ~affineequiv4 and ~affineequivne which are based on ~affineequiv3). I'll have a look...

set.mm
avekens: Fixed

set.mm
avekens: I added a proof ~ehl0 of this statement in the comment of ~df-ehl, which should make clear why the range starts with 1 instead of 0.

set.mm
avekens: Fixed

set.mm
avekens: Fixed

set.mm
avekens: I proved ~rrx0 (saying that `( I X. { 0 } )` is the zero ("origin") in a generalized real Euclidean space) and referenced it in the comment of ~ 2sphere0. I hope this is sufficient.

set.mm
avekens: Fixed

set.mm
avekens: There are 4 (additional) proofs which could be shortened using ~orcomdd: ~3orel2, ~elicoelioo, ~nneo, ~swoer.
Is this a sufficient argument to move/keep ~orcomdd in main?

benjub: I do not know yet. Maybe @digama0 can chime in.  There is a convention that when proving in "deduction style", one uncurries the antecedents, that is, one writes one antecedent as a big conjunction instead of keeping nested implications. That way, a proof part in deduction style typically uses things like https://us.metamath.org/mpeuni/imp.html at the beginning, then uses a bunch of (single) deduction style lemmas, and then if necessary uses https://us.metamath.org/mpeuni/ex.html to go back. Of course, if one adds a whole library of double deduction lemmas, then this will shorten the proof, but at the cost of doubling the size of the library.  And this can be done in a "slippery slope" way: one adds a first double deduction lemma, which will make more proofs shortenable by adding a second double deduction lemma, etc.

Therefore, double deductions should be added only in the most elementary parts of set.mm (typically, formulas with only `->`, `-.`, `<->`).  It's not clear where to draw the line, and it's true that commutation of disjunction is pretty basic too. But the sections with conjunction and disjunction are already very large, so I'm bit skeptical.

set.mm
benjub: To explain my above proposal: it mentions a basis, which the original comment fails to do, and "at a coordinate" has little sense.

Side remark: freeness, for a module, is a property, not the data of a particular basis.

set.mm
avekens: There are 5 (additional) proofs which could be shortened using ~eqeqan1d: ~chcoeffeq, ~gaorb, ~isdmn, ~isrgr, ~sylow1.
Is this a sufficient argument to move/keep ~eqeqan1d in main?

set.mm
benjub: I haven't looked precisely, but I would be in favor to move that block, next to the similar theorems in Main. I think they have a natural enough statement and fall into a pattern. Maybe bj-lineqi can stay in the mathbox indeed.

What would be very nice is if you could prove the version ~bary2 in two dimensions. The proof would be much longer. I do not know if at that point it will be shorter to use Carmer's formulas or not. Having ~bary2 would suffice already to talk about triangles in a satisfactory way. (My work on barycentric coordinates actually started with a discussion about defining the interior of a triangle.)

But I also have a more general question: why not do all of this (including all the small lemmas to help computations in CC) for general rings (commutative rings, fields when needed) and modules/vector spaces instead ? This will require more groundwork but would certainly be nicer

set.mm
benjub: Indeed: a distance has values in $[0,\infty]$ so why define a sphere of negative radius anymore than of radius equal to any random set ? The reason is that you can obtain a negative number by subtracting two positive ones, so defining "spheres" of negative radius (as the empty set, obviously) might allow one to dispense of special degenerate cases in proofs. This is only a matter of convenience.

I do not have a specific example right now, but for instance in RR^3, the cone $\{ (x, y, z) \in \mathbb{R}^3 \mid \sqrt{x^2 + y ^2} = z \}$ intersects each plane z=cte in a "sphere" (here a circle) of radius z, so the formal version of that would not have a special case for z<0.

Also, "undefined values" are currently set to (/) so it's true that it wouldn't change the value here, but explicitly defining it as (/) is cleaner since it does not depend on an outside convention.

set.mm
benjub: Maybe. I will not oppose the move to Main, but I would like someone else (@tirix ?) to give their insight. Also, I think the label is not consistent with other theorems `eqeqan*` but you and Thierry know these labeling conventions better, so please just check that it's ok.

set.mm
digama0: I think it is fine to justify these kind of theorems on a strict "does it decrease character count" basis. That is enough to block any slippery slopes.

set.mm
tirix: I kind of agree with @benjub 's position [above](https://github.com/metamath/set.mm/pull/3034/files#r1103772266), which is especially valid in the case above.

Norm was saying that one could build a huge number of new first order logic theorems using two-steps proofs from the existing ones, but that he considered first order theory in `set.mm` as complete, and was reluctant to introduce such new theorems (I did not make the effort to find the exact reference and his exact words, but I clearly remember that). 

I think this would have applied here, and for `~orcomdd`.

As for naming, I think this is a kind of hybrid case, maybe a new naming convention would be useful?

set.mm
tirix: Yes, that's exactly the spirit of my remark. I find it particularly satisfying to build such bridges between different parts of `set.mm`, and to reach an economy of symbols and definitions.

I think we can commit like this, and maybe simplify, or add these "bridge" proofs later?

set.mm
tirix: Yes, perfect, thank you!

set.mm
benjub: @digama0 : what about the "proof style" I described above ?  Also, I am not sure that the "does it decrease character count" is enough to block slippery slopes, since the addition of one such theorem may make new theorems fulfill the criterion, etc.

As for proof style, this is already illustrated in the current proof of orcomdd:
```
 9   syl6.1=orcomdd.1 $e |- ( ph -> ( ps -> ( ch \/ th ) ) )
12   syl6.2=pm1.4     $p |- ( ( ch \/ th ) -> ( th \/ ch ) )
13 orcomdd=syl6     $p |- ( ph -> ( ps -> ( th \/ ch ) ) )
```
which is indeed shorter than the more systematic proof
```
16       imp.1=orcomdd.1 $e |- ( ph -> ( ps -> ( ch \/ th ) ) )
17     orcomd.1=imp    $p |- ( ( ph /\ ps ) -> ( ch \/ th ) )
18   ex.1=orcomd     $p |- ( ( ph /\ ps ) -> ( th \/ ch ) )
19 orcomdd=ex      $p |- ( ph -> ( ps -> ( th \/ ch ) ) )
```
For easy proofs, shorter is better since anyway both are readable, but for longer proofs, maybe a more systematic method is preferable over a highly optimized but less uniform hence harder to grasp proof. This is analogous to computer programs.

set.mm
avekens: I think the label is OK, because the "2" in ~eqeqan12d is eliminated by merging the 2. hypothesis into the 2. part of the conjunction.

avekens:
> What is your project here? What will come next?

My goal is to prove that every line through a point within a circle (and an abitrary other point) has exactly two intersection points with the circle. These intersection points are determined by ~itsclinecirc0. This could be the basis for the  Beltrami-Klein model for a hyperbolic geometry, but I do not know how far I will dive into this topic afterwards.

set.mm
jkingdon: I too like the style where all the conditions are an antecedent joined by conjunctions. Not sure how hard a line I'd want to take on this, however. This is far from our first double deduction theorem,[1] which doesn't automatically say it is fine to add more, but does say we are partway down the slope whether it is slippery or otherwise.

Anyone have any appetite for putting a discouraged tag on `orcomdd` and having the comment suggest `orcomd` (plus `imp` and `ex` where needed) instead? Although part of me says that sort of thing should be decided by the existing rules around what makes proofs shorter, rather than by trying to form opinions about whether double deduction is good or bad.

[1] https://us.metamath.org/mpeuni/syl6.html instead of https://us.metamath.org/mpeuni/syl.html , https://us.metamath.org/mpeuni/syl6mpi.html instead of https://us.metamath.org/mpeuni/mpisyl.html , probably a fair number of others I'm not finding just now.

benjub: > [1] https://us.metamath.org/mpeuni/syl6.html instead of https://us.metamath.org/mpeuni/syl.html , https://us.metamath.org/mpeuni/syl6mpi.html instead of https://us.metamath.org/mpeuni/mpisyl.html , probably a fair number of others I'm not finding just now.

Note that these are in implicational calculus, and fall into the category I mentioned above (that is: at most `->`, `<->`, `-.`).

set.mm
jkingdon: > Norm was saying that one could build a huge number of new first order logic theorems using two-steps proofs from the existing ones, but [probably we shouldn't]

When you say it that way it sounds persuasive. I wish I had a better feel for the best place to draw the line. We can always fall back on rules like whether it decreases the total size of set.mm which is perhaps not a bad rule (even though it typically takes some effort to measure it in practice).

set.mm
avekens: I think the proposal of @jkingdon is a good compromise for now: I will putting a discouraged tag on orcomdd (name it orcomddOLD) and having the comment suggest orcomd. Fuethermore, I will replace all occurences of orcomdd in proofs (only three) by orcomd. If nobody objects/intervenes within one year, this theorem would be deleted.

In the future, I will be more careful moving theorems from mathboxes to main...

set.mm
avekens: Same as for ~orcomdd, I will mark this theorem as deprecated.

set.mm
avekens: I have problems with ~sbcom3OLD, which uses ~orcomdd: It is commented out (by Norm) with the following comment:

> Note: this proof requires many
>        hours to "MM> PROVE sbcom3OLD" and "MM-PA> MINIMIZE_WITH *" and
>        "MM-PA> SAVE NEW_PROOF".  The discouragements will prevent accidental
>        minimization attempts and other operations that would interfere with
>        scripts.  For the time being, we are keeping this proof as a test case,
>        but it will be deleted eventually. -NM $)

After activating this theorem, I was not able to open it neither with metamath.exe nor with mmj2. Therefore, I cannot replace ~orcomdd. Do we really need this theorem? If yes, we have to keep ~orcomdd (in GM's mathbox).

jkingdon: > After activating this theorem, I was not able to open it neither with metamath.exe nor with mmj2. Therefore, I cannot replace ~orcomdd. Do we really need this theorem?

I don't know, I'm not sure what it is a test case for. Like is the idea that maybe a future tool could open the proof more quickly than "many hours"?

benjub: That infamous theorems has been causing problems already. Since it is superseded by sbcom3 in every aspect, I would remove it.
If one needs "stress tests" for verifiers, there are other places for that.

benjub: Sorry, I have little time this week. Feel free to resolve this one.

set.mm
benjub: Good for me.

set.mm
avekens: With my latest commit I provide a theorem showing ``( LineG ` ( EEG ` N ) ) = ( LineM ` ( EEhil ` N ) )``. It required a lot of effort to provide its proof, but the simplicty of the result was worth it.

set.mm
benjub: frlmvscaval: Coordinates of a scalar multiple with respect to a basis in a module.

Maybe not perfect, but "at a coordinate" just makes no sense.

set.mm
avekens: > I haven't looked precisely, but I would be in favor to move that block, next to the similar theorems in Main. I think they have a natural enough statement and fall into a pattern. Maybe bj-lineqi can stay in the mathbox indeed.

I'll have alook at this next...
> 
> What would be very nice is if you could prove the version ~bary2 in two dimensions. The proof would be much longer. I do not know if at that point it will be shorter to use Carmer's formulas or not. Having ~bary2 would suffice already to talk about triangles in a satisfactory way. (My work on barycentric coordinates actually started with a discussion about defining the interior of a triangle.)

I think this goes beyond the scope of this PR

> But I also have a more general question: why not do all of this (including all the small lemmas to help computations in CC) for general rings (commutative rings, fields when needed) and modules/vector spaces instead ? This will require more groundwork but would certainly be nicer

This is also a general issue which cannot/should not be solved in this PR.

set.mm
avekens: Because of the accordance with the definition of LineG I would leave it as it is.

set.mm
avekens: As long as negative radii  are not really required I would leave this as it is.

set.mm
benjub: Your mathbox, your choice. Out of curiosity: where is the definition of lineG "df-lineg" gives nothing.

set.mm
benjub: Your mathbox, your choice.

set.mm
benjub: Indeed.  I should open an issue. Non-blocking and you can resolve.

benjub (commented):
Good for me.
**edit:** @avekens : I hope there is no hard feeling here. I may have been a bit annoying, but this was only to be sure that the changes in Main were good.

set.mm
avekens: Comments revised according to BJ's proposals.

set.mm
avekens: Required theorems moved to main and proof of ~itsclc0lem1 shortened by ~(bj-)lineq.

set.mm
avekens: The definition of LineG (as a slot index) is in df-lng.

avekens:
Many thanks to everybody who reviewed this PR, especially @benjub and @tirix. I think it is in a good state to be merged now. @tirix , @benjub could you (re)approve this PR, please? @benjub it should be sufficient that you have a short look at my latest commit.

set.mm
benjub: df-lng is not very enlightening...
It might be https://us.metamath.org/mpeuni/ttgval.html ? So for x=y it corresponds to the first proposal I made above (though now I think my second proposal using the affine hull generalizes more naturally to multiple points).

benjub (approved):

set.mm
avekens: @benjub you are right, the relationship between LineM and LineG is not very clear. Therefore, I added two references to important (defining) theorems in the comment of ~eenglngeehlnm.

set.mm
tirix: Yes, it was actually much harder than I thought! Thanks for this very interesting addition.

tirix (approved):

set.mm
```diff
@@ -730632,6 +731072,1417 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+      MWWAWVRWXKWXMWUQSWVRWXJWVSUXKOWVRLWUQWWNWWMYHTTWVRWVSUXKWUQWVRWUQLWWMWWNW
+      RZWXBWWMWWOYFXKYIXLWVRWXEWWCWWAPWVRUXMWUQWWTWWMWWOXMTWVRWWCWWAWVRWWBUXMWV
+      RWUQWWMWWOWSWWTYBWVRWVTUXKWVRWVSWUQWXNWWMWWOYEWXBYBUWJYIYJWOXJXEYKYLYMYQU
+      WKYNYOYMUWLYRUWMUWN $.
+
+    $d N i k l m n p t v w x y z $.
+    $( The line definition in the Tarski structure for the Euclidean geometry
+       (see ~ elntg ) corresponds to the definition of lines passing through
+       two different points in a left module (see ~ rrxlines ).  (Contributed
+       by AV, 16-Feb-2023.) $)
```
benjub, line 731598: But elntg is not a definition ?
Ok, I searched and finally found it: https://us.metamath.org/mpeuni/df-trkg.html
@tirix : https://us.metamath.org/mpeuni/elntg.html says that it's a definition but it's not. It looks more like our *val theorems it seems. Also, the considered space is not made clear in the comment. So reword the comment to something like
> Expression of the line formed on two distinct points in the Tarski geometry formed on the real coordinate space of given dimension.

tirix: Yes, some comments here:
The concept of "line" is not an elementary concept in Tarski's development; only betweenness and congruence are.

However, it is elementary in other developments of elementary geometry, like in Hilbert's axioms of geometry, or in incidence geometry. This is why a slot was added for Line, `LineG`. 

For Tarski's geometries, the way to define the `LineG` slot is part of its definition, and it is defined in terms of betweenness.

benjub: Thanks for the explanation why it got a "Slot" (although I'm not sure it was worth it). Can you change the comment to something like I proposed above?
You can also add your comment above somewhere, but maybe it's already somewhere in set.mm: I only rapidly skimmed through it.

tirix: It would be great if @avekens can do the change in this PR, however since `~elntg` is not changed by this PR, we might also consider it a different topic.

I'm putting this, the other point above about `~elntg` and other points left open in issue #3043 so they are not forgotten.

benjub: Thanks @tirix for the new Issue !

PR #3038, opened 2023-02-12
CCbar

benjub:
Main part:
* extract a proof from @digama0's axcnex (and add comment about an alternate proof); I extracted it because I need it in my mathbox.
* move df-tau from @jkingdon's mathbox since I use it in mine

Mathbox: some groundwork for extended and projective real and complex numbers.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+12-Feb-23 df-tau                moved from JK's mathbox to main set.mm
  5-Feb-23 srngfn    srngstr
 30-Jan-23 elsb4lem  elsb4v
 30-Jan-23 elsb3lem  elsb3v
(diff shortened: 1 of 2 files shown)


benjub:
@david-a-wheeler : is this due to your recent changes ?
```
Connecting to us.metamath.org (us.metamath.org)|173.255.232.114|:80... failed: Connection refused.
Error: Process completed with exit code 4.
```
**edit:** reran and it's ok. Still, a bit puzzling.

jkingdon:
> * move df-tau from @jkingdon's mathbox since I use it in mine

If our principle is to give people choice in mathboxes, seems like we should put the definition, and whatever theorems would otherwise be duplicated, into main.

david-a-wheeler:
Yes, my fault.  I fixed it soon after. Just rerun.

david-a-wheeler:
> If our principle is to give people choice in mathboxes, seems like we should put the definition, and whatever theorems would otherwise be duplicated, into main.

In general, once more than one person uses something, we move it into main. Does anyone object in this case?

avekens (changes requested):
I had a closer look at the changes in main (everything is fine) and only skimmed through the changes in the mathbox (formally OK).

The only thing to be done is to add df-tau (now in main already) to the list in changes-set.txt.

avekens (approved):

Issue #3037, opened 2023-02-12
\rm deprecated in LaTeX


benjub:
Since \rm is deprecated in LaTeX (it has been for 28 years), we should replace it overtime with \mathrm{...} (since symbol definitions assume we are in math mode) or at least use the latter for new symbols. Currently, set.mm has 828 occurrences of `\rm`.

(Remark: also deprecated in LaTeX are: \rm, \it, \bf, \tt, \sc, \sl, \sf and also `$$ ... $$` for displayed equations.)

tirix:
Can we do a global replace instead of replacing "over time"?
Also, are there also occurrences of the other deprecated LaTeX commands?
Also, pinging @GinoGiotto because he was interested in generating LaTeX from Metamath.

benjub:
We can probably take care of most cases automatically by replacing appropriate regexes, but that's not that easy since we need to replace `\rm arbitraryexpression` with `\mathrm{arbitraryexpression}` where the arbitrary expression stops at the following space or quote (**edit:** or backslash, or anything else ending a LaTeX token...), but it can be a single or double quote, maybe there are extra quotes within the main quote, etc.

As for the other deprecated commands, I found in set.mm `\sf` once and `\tt` twice. But there are other databases (ping @jkingdon) and some deprecated commands may be found in the part of `metamath.c` that generates LaTeX code (I haven't checked).

jkingdon:
> But there are other databases

A glance through iset.mm shows mostly markup copy-pasted from set.mm.  The goal is to keep these identical so it should be feasible to make the edits in set.mm and then copy-paste the syntaxes which are found in both over to iset.mm. Or any other means of achieving (roughly) that result.

david-a-wheeler:
I have good news, I know regular expressions <https://xkcd.com/208/>.

I made a first stab here:
https://github.com/metamath/set.mm/pull/3039

GinoGiotto:
> Can we do a global replace instead of replacing "over time"? Also, are there also occurrences of the other deprecated LaTeX commands? Also, pinging @GinoGiotto because he was interested in generating LaTeX from Metamath.

[...]

PR #3039, opened 2023-02-13
Update LaTeX, \rm -> \mathrm

david-a-wheeler:
Per #3037, update the LateX to change from the deprecated \rm to the official \mathrm{...}.

I started this using this regular expression:
sed -E -i '' -e '/latexdef/s/\\rm (\\?[A-Za-z0-9\-^\.\[\]\;]+) *\}/\\mathrm{\1}}/g' set.mm iset.mm

.. and then fixed others by hand. I fixed all the entries in set.mm, and many in iset.mm. If this approach seems reasonable we can fix the rest of iset.mm.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,25 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+16-Feb-23 bj-ldiv   ldiv        moved from BJ's mathbox to main set.mm
+16-Feb-23 bj-rdiv   rdiv        moved from BJ's mathbox to main set.mm
+16-Feb-23 bj-mdiv   mdiv        moved from BJ's mathbox to main set.mm
+16-Feb-23 bj-lineq  lineq       moved from BJ's mathbox to main set.mm
+16-Feb-23 subdivcomb2 [same]    moved from SF's mathbox to main set.mm
+16-Feb-23 negelrpd  [same]      moved from GS's mathbox to main set.mm
+16-Feb-23 1xr       [same]      moved from GS's mathbox to main set.mm
+16-Feb-23 rexlimdva2 [same]     moved from GS's mathbox to main set.mm
+16-Feb-23 mvrladdd  [same]      moved from DAW's mathbox to main set.mm
+12-Feb-23 df-tau                moved from JK's mathbox to main set.mm
+10-Feb-23 rrxmetfi  [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 rrxdsfi   [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 rrxbasefi [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 addgtge0d [same]      moved from AI's mathbox to main set.mm
+10-Feb-23 addeq0    [same]      moved from TA's mathbox to main set.mm
+10-Feb-23 assraddsubd [same]    moved from DAW's mathbox to main set.mm
+10-Feb-23 xpexd     [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 fvmptd3   [same]      moved from GS's mathbox to main set.mm
+10-Feb-23 eqeqan1d  [same]      moved from PM's mathbox to main set.mm
  5-Feb-23 srngfn    srngstr
 30-Jan-23 elsb4lem  elsb4v
 30-Jan-23 elsb3lem  elsb3v
--- a/discouraged
+++ b/discouraged
@@ -50,6 +50,7 @@
 "0ncn" is used by "axaddf".
 "0ncn" is used by "axmulf".
 "0ncn" is used by "bj-inftyexpidisj".
+"0ncn" is used by "bj-inftyexpitaudisj".
 "0ngrp" is used by "vsfval".
 "0ngrp" is used by "zrdivrng".
 "0nnnALT" is used by "nnne0ALT".
@@ -4350,6 +4351,7 @@
 "df-bj-inftyexpi" is used by "bj-inftyexpidisj".
 "df-bj-inftyexpi" is used by "bj-inftyexpiinv".
 "df-bj-inftyexpi" is used by "bj-minftyccb".
+"df-bj-inftyexpitau" is used by "bj-inftyexpitaudisj".
 "df-bj-pr1" is used by "bj-pr1eq".
 "df-bj-pr1" is used by "bj-pr1ex".
 "df-bj-pr1" is used by "bj-pr1un".
@@ -4418,6 +4420,7 @@
 "df-c" is used by "axcnre".
 "df-c" is used by "axmulf".
 "df-c" is used by "axresscn".
+"df-c" is used by "bj-inftyexpitaudisj".
 "df-c" is used by "dfcnqs".
 "df-c" is used by "mulcnsr".
 "df-c" is used by "opelcn".
@@ -4642,7 +4645,6 @@
 "df-nr" is used by "addclsr".
 "df-nr" is used by "addcomsr".
 "df-nr" is used by "addsrpr".
-"df-nr" is used by "axcnex".
 "df-nr" is used by "distrsr".
 "df-nr" is used by "ltasr".
 "df-nr" is used by "ltsosr".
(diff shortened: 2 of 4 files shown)


david-a-wheeler:
I think I got all the set.mm ones, and I also fixed many of the iset.mm ones.

This is *really* fiddly & easy to get wrong, and I don't think we have a test that the LateX generated is okay. If so, anyone have a way to test?

david-a-wheeler:
Remember: the good news is that "I know regular expressions" https://xkcd.com/208/.

iset.mm
```diff
@@ -128314,7 +128314,7 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
     "<IMG SRC='_wff.gif' WIDTH=24 HEIGHT=19 ALT=' wff' TITLE='wff'> ";
   althtmldef "wff" as
     '<SPAN CLASS=typecode STYLE="color:gray">wff </SPAN>'; /* was #00CC00 */
-  latexdef "wff" as "{\rm wff}";
+  latexdef "wff" as "{\mathrm{wff}}";
```
jkingdon, line 128317: I don't think we want the outer curly braces. At least we don't have them in existing entries like:

```
  latexdef "recs" as "\mathrm{recs}";
```

jkingdon:
> This is _really_ fiddly & easy to get wrong, and I don't think we have a test that the LateX generated is okay. If so, anyone have a way to test?

I suppose at some point there is no substitute for looking at the output.

Within metamath I ran

```
open tex foo.tex
show statement eirrap/tex
close tex
```

then from the command line I ran

```
sudo apt install texlive-latex-base
pdflatex foo.tex
```

which I think is close but I apparently need another package or two, because there were error messages like

```
! LaTeX Error: File `pdftexcmds.sty' not found.
```

Probably someone here just knows what I need to install?

This is Pop!OS but I think that is close enough to Ubuntu to figure it is Ubuntu for these purposes.

david-a-wheeler:
A partial answer: https://tex.stackexchange.com/questions/518757/module-pdftexcmds-not-found-after-update

My recommendation is to just install a massive "complete" TeX install. Trying to install just "what you need" is a pain.

jkingdon:
> My recommendation is to just install a massive "complete" TeX install.

That was one of the things I tried, but maybe I was too tired last night to pull it off. Seems like I'm now having better results from `sudo apt-get install texlive-full`

You are right about "massive" though.

I suppose https://help.ubuntu.com/community/LaTeX is probably the place with all the answers.

As for what happens when I install it and run `pdflatex foo.tex`, I get:

```
! Undefined control sequence.
<argument> ...thbb {Q} \rightarrow {\rm e} \apart 
                                                  Q )\label {eq:eirrap}\tag ...
```

which perhaps means this setup doesn't have `\rm`  at all? The more I read things like https://chemistry.meta.stackexchange.com/questions/4474/using-mathrm-versus-rm-mathjax-index-formatting the more I start to remember why I haven't learned (or retained!) that much TeX/LaTeX.

GinoGiotto:
> I think I got all the set.mm ones, and I also fixed many of the iset.mm ones.
> 
> This is _really_ fiddly & easy to get wrong, and I don't think we have a test that the LateX generated is okay. If so, anyone have a way to test?

Not sure if this is what you are looking for, but it happens that I coded a program that collects all latex definitions and produces [THIS](https://github.com/metamath/set.mm/files/10775510/list.txt) file, which is just a list of all latex definitions from set.mm in the same output format of a proof generated by metamath.exe (each numbered line shows a latex translation). To check if a translation doesn't work, I can just open the file with Texmaker or TeXstudio and find it (note: the list here is from an outdated version of set.mm, if this is useful I can set things up to rerun it).

david-a-wheeler:
I accidentally fixed the *develop* branch one line per the discussion above, so it's a fix, but not a *problem* :-).

david-a-wheeler:
Okay to merge?

david-a-wheeler:
Hmm, I'm getting the same error. I don't normally use TeX from metamath.

Here's what I see:

~~~~
! Undefined control sequence.
<argument> ...thbb {Q} \rightarrow {\rm e} \apart 
                                                  Q )\label {eq:eirrap}\tag ...
~~~~

I was hoping a quick search-and-replace would resolve the problem; I wasn't intending to do a deep dive into TeX.

benjub (approved):

benjub:
> ! Undefined control sequence.
> <argument> ...thbb {Q} \rightarrow {\rm e} \apart 
>                                                   Q )\label {eq:eirrap}\tag ...

can you give more context ?

In the example, there is an extraneous space after `\label`.

benjub:
By the way @jkingdon : I don't know where you saw `\apart` but it might be in a pretty exotic package ? It is not even in the comprehensive latex symbol list. (Also, I wondered where that `eirrap` was coming from since it is not in set.mm.)

One possibility: `\neq\!\!\!\neq`.

david-a-wheeler:
Okay @benjub wins the "all eyes are shallow" award; `\apart` is **exactly** the problem

I did a quick tweak to the generated .tex file:

~~~~
diff -u foo.tex foo-mod.tex 
--- foo.tex	2023-02-19 14:46:30.000000000 -0500
+++ foo-mod.tex	2023-02-19 16:41:34.000000000 -0500
@@ -20,7 +20,7 @@
        this and saying that ${\rm e}$ is not rational, which is {\tt eirr}.
        (Contributed by Jim Kingdon, 6-Jan-2023.)
 \begin{align}
-\vdash ( Q \in \mathbb{Q} \rightarrow {\rm e} \apart Q
+\vdash ( Q \in \mathbb{Q} \rightarrow {\rm e} \neq\!\!\!\neq
     )\label{eq:eirrap}\tag{eirrap}
 \end{align}
 \end{theorem}
~~~~

I then ran `pdflatex foo-mod.tex`, and everything just worked.

There are no uses of `\apart` in `set.mm`, which is probably why I've never noticed it.
The guilty statements in `iset.mm` appear to be:

~~~~
htmldef "=//=" as ' # '; 
  althtmldef "=//=" as ' # ';
  latexdef "=//=" as "\apart";
htmldef "#RR" as ' #<SUB>&#8477;</SUB> ';
  althtmldef "#RR" as ' #<SUB>&#8477;</SUB> '; 
  latexdef "#RR" as "\apart_\mathbb{R}";
~~~~

david-a-wheeler:
I ran a new test, this time on set.mm using this branch:
~~~~sh
rm -f foo2.tex
metamath <<END
read set.mm
open tex foo2.tex
show statement nfiota1/tex
close tex
END
pdflatex foo2.tex
~~~~

It all just worked, and the generated PDF worked. In addition, "show proof" also worked with TeX (with the caveat that running it in isolation causes a lot of undefined references, which is fair).

So while this pull request doesn't solve everything, I think it suggests it's a workable direction.

david-a-wheeler:
I *literally* replaced `\rm {...}` with `\mathrm{...}`. So `{\rm {...}}` became `{mathrm{...}}`. That's not *wrong*, but it does have unneeded curly braces. So maybe we should replace that specific case of extra braces to remove them. But we can do that later, the point is to remove the `\rm` since it's deprecated.

benjub:
> I _literally_ replaced `\rm {...}` with `\mathrm{...}`. So `{\rm {...}}` became `{mathrm{...}}`. That's not _wrong_, but it does have unneeded curly braces. So maybe we should replace that specific case of extra braces to remove them. But we can do that later, the point is to remove the `\rm` since it's deprecated.

Yes, this was noticed last week by @jkingdon and his post has two "+1"s. Removing the extra braces looks easier than the replacement you already made above.

david-a-wheeler:
@benjub - okay, I've now tweaked this change to remove the unnecessary curly braces around `\mathrm{...}`.

Ready to merge? This doesn't solve all problems, but hopefully it helps address the issue identified at the top. I only did this for set.mm; now that we know some things, we can make some more fixes in other databases, and make the similar changes to remove the other deprecated items.

tirix (approved):

jkingdon (approved):

jkingdon:
> `\apart` is **exactly** the problem

I have moved this issue to https://github.com/metamath/set.mm/issues/3050 - I should have known that trying to do something (generate LaTeX from metamath) which I'd never done before could hit all manner of snags. Definitely agree that shouldn't a blocker to getting rid of `\rm`. (And thanks to multiple people for all the sleuthing on this).

david-a-wheeler:
This pull request has to keep getting fixed from merge conflicts, because of its many little changes. I've re-fixed the merge conflicts, and since it's been approved other than that, I'll merge as soon as it passes the verification checks. Otherwise it'll be stuck in "always has merge conflicts" limboland.

PR #3048, opened 2023-02-19
shorten eubid

wlammen:
1. shorten eubid

@benjub The obsolete theorem residOLD could not be removed, since it has a use in elid.  There is a TODO where you wrote you want to make residOLD redundant, but this is still open.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -11610,7 +11610,6 @@
 "relrngo" is used by "rngoablo2".
 "relrngo" is used by "rngoi".
 "relrngo" is used by "rngosn3".
-"residOLD" is used by "elid".
 "ressval3dOLD" is used by "estrresOLD".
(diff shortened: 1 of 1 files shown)


tirix (approved):

avekens (approved):

benjub:
Thanks for noticing @wlammen. That section is just huge, we'll have to structure it one day with subsections, and introduce definitions successibely (like was done for the sections and/or and mo/eu). Currently, no fewer than eight definitions are introduced as a bloc.

For now, can you simply move ~elid right after ~dfrel3 ?

wlammen:
elid moved after dfrel3, TODO in elid adjusted, residOLD deleted, as @benjub suggested.

benjub (approved):

PR #3052, opened 2023-02-20
Fix a few straggling \rm -> \mathrm{..} in LaTeX

### LINKED ISSUE #3037: \rm deprecated in LaTeX

david-a-wheeler:
fixes #3037

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+21-Feb-23 moimd     moimdv      align with mobidv, eubidv 
 16-Feb-23 bj-ldiv   ldiv        moved from BJ's mathbox to main set.mm
 16-Feb-23 bj-rdiv   rdiv        moved from BJ's mathbox to main set.mm
 16-Feb-23 bj-mdiv   mdiv        moved from BJ's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -7890,7 +7890,6 @@
 "idi" is used by "imo72b2lem0".
 "idi" is used by "limsupvaluz2".
 "idi" is used by "madjusmdetlem2".
-"idi" is used by "ovncvr2".
 "idi" is used by "rngcifuestrc".
 "idi" is used by "sge0f1o".
 "idi" is used by "smfinfmpt".
@@ -15563,6 +15562,10 @@ New usage of "funressnvmoOLD" is discouraged (0 uses).
 New usage of "funsneqopOLD" is discouraged (0 uses).
 New usage of "funsneqopsnOLD" is discouraged (1 uses).
 New usage of "fvimacnvALT" is discouraged (0 uses).
+New usage of "fvsnOLD" is discouraged (0 uses).
+New usage of "fvsngOLD" is discouraged (0 uses).
+New usage of "fvsnun1OLD" is discouraged (0 uses).
+New usage of "fvsnun2OLD" is discouraged (0 uses).
 New usage of "gen11" is discouraged (19 uses).
 New usage of "gen11nv" is discouraged (5 uses).
 New usage of "gen12" is discouraged (7 uses).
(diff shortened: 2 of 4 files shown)


set.mm
```diff
@@ -437504,10 +437504,10 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 /* Elementary  geometry (extension)  */
 htmldef "LineM" as 'LineM';
  althtmldef "LineM" as 'Line<sub>M</sub>';
- latexdef "LineM" as "{\rm Line}_M";
+ latexdef "LineM" as "{\mathrm{Line}_M}";
```
tirix, line 437507: The outside brackets can be removed here, I guess.

tirix (approved):

set.mm
benjub: By the way: anyone knows why the "suggestion" feature is unavailable ? It would be pretty handy here: we could even apply the suggestion and merge without waiting.

david-a-wheeler: @benjub - you have write access to this repo, so you *should* be able to the suggestion feature as far as I know.

benjub:
```suggestion
 latexdef "LineM" as "\mathrm{Line}_M";
```

benjub: Just made the suggestion after several different trials (the button was inactive when in "Review" mode for a reason I cannot pin down).

iset.mm
```diff
@@ -129928,27 +129928,27 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
   althtmldef "-1-1->" as
     '&ndash;<FONT SIZE=-2 FACE=sans-serif>1-1</FONT>&rarr;';
   latexdef "-1-1->" as
-    "\raisebox{.5ex}{${\textstyle{\:}_{\mbox{\footnotesize\rm 1" +
-    "\tt -\rm 1}}}\atop{\textstyle{" +
-    "\longrightarrow}\atop{\textstyle{}^{\mbox{\footnotesize\rm {\ }}}}}$}";
+    "\raisebox{.5ex}{${\textstyle{\:}_{\mbox{\footnotesize\mathrm{1}" +
+    "\tt -\mathrm{1}}}}\atop{\textstyle{" +
+    "\longrightarrow}\atop{\textstyle{}^{\mbox{\footnotesize\mathrm{\ }}}}}$}";
```
benjub, line 129933: > \mathrm{\ }

not sure about that but this overly complicated latexdef will hopefully soon go away, see #3054

benjub (approved):

set.mm
benjub, line 437507:
```suggestion
 latexdef "LineM" as "\mathrm{Line}_M";
```

benjub: I cannot apply the suggestion... "This diff has recently been updated. Refresh and try again." but refreshing has no effect...

benjub:
@david-a-wheeler can you apply the suggestion so that we can merge ?

set.mm
tirix: I can see the button "commit suggestion". I tried several times but each time I get "_This diff has recently been updated - Refresh and try again_"

david-a-wheeler: Okay, will do!

iset.mm
david-a-wheeler: Well, `\mathrm{\ }`  does have an effect - it creates horizontal space like a Roman-style space. If we don't need it, let's remove it, but I presumed it was put in for a reason. Let's make removing that a separate proposal (as discussed).

david-a-wheeler:
Changed per discussion above, and it has 2 approvers. Presuming it passes the verifiers, I plan to merge it very soon.

PR #3053, opened 2023-02-20
[DRAFT] Remove space characters from TeX

david-a-wheeler:


david-a-wheeler:
Hmm, this doesn't work:

~~~~
rm -f foo2.tex
metamath <<END
read set.mm
open tex foo2.tex
show proof foeq1 /tex
close tex
END

pdflatex foo2.tex && open foo2.pdf
~~~~

I'll need to test the original (need to do that later).

set.mm
```diff
@@ -434160,8 +434159,7 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 */
 htmldef "^" as "<IMG SRC='uparrow.gif' WIDTH=7 HEIGHT=19 ALT=' ^' TITLE='^'>";
   althtmldef "^" as '&uarr;';
-  latexdef "^" as "\uparrow"; /*
-  latexdef "^" as "\widehat{\ }"; */
+  latexdef "^" as "\uparrow";
```
benjub, line 434162: same in `iset.mm`

benjub: ping @david-a-wheeler, and after we can merge

david-a-wheeler:
Oy! I think we have a bigger problem!!

`mathrm{...}` is only allowed in TeX *math mode*. `\rm` works in either mode. Generated proofs aren't always in math mode, so none of our so-called fixes work for all use cases. This is a problem with the current situation, not just this PR.

In particular, this appears to generate a TeX file that can't be processed by TeX:

`show proof foeq1 /tex`

I don't really want to delve into TeX details. Can anyone suggest how to fix this? I guess a solution is ensuring that metamath-exe always uses math mode in these circumstances.

benjub:
LaTeX has the command `\ensuremath` if really needed, but I think it'd be better that in `.mm` databases, it is assumed that the latexdef's are written for math mode (and I thought it was the case...). I'll try your example and see.

benjub:
As I answered you in https://github.com/metamath/set.mm/issues/3037#issuecomment-1438871312 it is the command `\mbox` that triggers text mode.

david-a-wheeler:
Thanks! Perhaps it's just in my specific setup (though I doubt it). Here's how to reproduce.

In my current version of `develop` commit `37c880e1a9729f788e85ff91321a8d4e2dbbe6eb` (not even this PR), running this command:

~~~~
rm -f foo2.tex
metamath <<END
read set.mm
open tex foo2.tex
show proof foeq1 /tex
close tex
END
~~~~

Produces this error:

~~~~
(/usr/local/texlive/2022/texmf-dist/tex/latex/amsfonts/umsa.fd)
(/usr/local/texlive/2022/texmf-dist/tex/latex/amsfonts/umsb.fd)

LaTeX Warning: Reference `eq:fneq1' on page 1 undefined on input line 47.


LaTeX Warning: Reference `eq:rneq' on page 1 undefined on input line 47.


LaTeX Warning: Reference `eq:eqeq1d' on page 1 undefined on input line 47.


LaTeX Warning: Reference `eq:anbi12d' on page 1 undefined on input line 47.


! LaTeX Error: \mathrm allowed only in math mode.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.47 \end{align}
~~~~

benjub:
The statement https://us.metamath.org/mpeuni/foeq1.html involves `"-onto->"` whose latexdef is
```
  latexdef "-onto->" as
    "\raisebox{.5ex}{${\textstyle{\:}_{\mbox{\footnotesize\mathrm{\ }}}}" +
    "\atop{\textstyle{" +
    "\longrightarrow}\atop{\textstyle{}^" +
    "{\mbox{\footnotesize\mathrm{onto}}}}}$}";
```
the `\mbox`'s on the first and last lines trigger text mode and contain a `\mathrm`, so LaTeX complains. If you use there `\textrm` it should be ok. That said, this latexdef is overly complicated (and has several other problems, like the deprecated `\atop`), hence my Issue #3054.

set.mm
david-a-wheeler: Okay, I'll make that change as well!

david-a-wheeler:
@benjub - Ah! Thank you! I didn't notice that this was covered elsewhere.

We're consistently using `\mathrm` in all places, so I think we should be always switching to math mode for this.

So I fixed the merge conflicts. I suggest we merge this as it is, and then we can directly fix the TeX so it consistently uses math mode (and never assumes anything else else). Then we don't have to deal with merge conflicts.

benjub (approved):

PR #3055, opened 2023-02-21
Fix latex rendering for apartness in iset.mm

### LINKED ISSUE #3050: Fix apartness LaTeX in iset.mm
Right now iset.mm has

```
htmldef "=//=" as ' # '; 
  althtmldef "=//=" as ' # ';
  latexdef "=//=" as "\apart";
htmldef "#RR" as ' #<SUB>&#8477;</SUB> ';
  althtmldef "#RR" as ' #<SUB>&#8477;</SUB> '; 
  latexdef "#RR" as "\apart_\mathbb{R}";
```

which doesn't work as described in https://github.com/metamath/set.mm/pull/3039#issuecomment-1436105224 . Probably never did work given that generating LaTeX is (we think) a little used feature.

As for the history, I think I just grabbed `\apart` from https://github.com/HoTT/book and didn't really think too much about where it was defined, which apparently is https://github.com/HoTT/book/blob/95ed534fa8ea321b4beacd2ce9184877e95463a8/macros.tex#L615

So maybe we can just change `\apart` to `\mathrel{\#}` both places it appears in iset.mm? (Disclaimer: haven't tried it yet, and also willing to listen to people who know more LaTeX than me).

jkingdon:
Instead of referencing `\apart` (which we never define), use `\neq\!\!\!\neq` . This is more similar to a not equal sign (with an extra line) than the way # looks in most fonts.

Here's what it looks like:

image

Fixes #3050


benjub:
Thanks. When zooming, it looks like the superimposition makes parts of the horizontal bars bolder.  But "\joinrel" ("\mathrel") do not seem to change anything. Note that `\!` corresponds to `-3mu` so the first two are identical in
```
\[
\neq\!\!\!\neq
\neq\mathrel{\mkern -9mu}\neq
\neq\mathrel{\mkern -10mu}\neq
\]
```
Maybe the slightly smaller spacing in the third looks a bit better, but all are good to me.

jkingdon:
I don't think I have a lot of preferences, so sure `\neq\mathrel{\mkern -10mu}\neq` will do. I've updated the pull request with that.

Here's a screenshot of what is looks like:

image

benjub (approved):

PR #3059, opened 2023-02-22
Add ruc to mmil.html

jkingdon:
What I mean by "assuming it holds up" is that the authors are still writing the paper (about half done according to what Bauer posted to Mastodon) and it hasn't been peer reviewed or stood the test of time.

But with suitable caveats it seems appropriate to post what we know about whether https://us.metamath.org/mpeuni/ruc.html will be provable in iset.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8256,6 +8256,16 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>ruc</TD>
+  <TD><I>none</I></TD>
+  <TD>Apparently not provable without countable choice, assuming
+  the following result holds up:
+  Andrej Bauer (12-May-2022), <a
+  href="https://researchseminars.org/talk/ToposInstituteColloquium/57/"
+  >"The countable reals"</a>, Topos Institute Colloquium</TD>
+</TR>
+
+<TR>
   <TD>dvdsaddre2b</TD>
   <TD><I>none</I></TD>
   <TD>Something along these lines (perhaps with real changed to


jkingdon:
Hmm, I guess there is something about the A tag which confuses the bibliographic software?

jkingdon:
> Hmm, I guess there is something about the A tag which confuses the bibliographic software?

On second thought, I guess not related to anything in this pull request? Should be taken care of by #3060 .

tirix:
> What I mean by "assuming it holds up" is that the authors are still writing the paper (about half done according to what Bauer posted to Mastodon) and it hasn't been peer reviewed or stood the test of time.

Wow, you're at the edge of math research!

jkingdon:
> Wow, you're at the edge of math research!

That's one fun thing about iset.mm.  Although it is a bit off the path of most constructive mathematics (since most researchers use lean or agda or something else as a tool, and some flavor of type theory as a foundation), even a concrete task like intuitionizing set.mm isn't all THAT far from the edge.  Another example was https://us.metamath.org/ileuni/exmidsbth.html which is only the second time this theorem has been formalized (to my knowledge). In the process of formalizing it I submitted some corrections to the key paper (which the authors accepted).

benjub:
> Another example was https://us.metamath.org/ileuni/exmidsbth.html which is only the second time this theorem has been formalized (to my knowledge). In the process of formalizing it I submitted some corrections to the key paper (which the authors accepted).

I don't see any mention of that key paper in the comment of https://us.metamath.org/ileuni/exmidsbth.html ? It would be worth adding a link or a reference.

By the way, I would replace in that page "which accepts excluded middle as an axiom" by "in which excluded middle holds". Indeed, excluded middle is not an axiom in the current version of set.mm.

jkingdon:
> I don't see any mention of that key paper

The mention is at https://us.metamath.org/ileuni/exmidsbthr.html .

> By the way, I would replace in that page "which accepts excluded middle as an axiom" by "in which excluded middle holds". Indeed, excluded middle is not an axiom in the current version of set.mm.

Hmm, OK, I guess I buy that. Given that `ax-3` and `exmid` are equivalent (over iset.mm axioms), this is a matter of style as much as a mathematical distinction, but it doesn't seem cumbersome to say "in which excluded middle holds".

benjub:
> > I don't see any mention of that key paper
> 
> The mention is at https://us.metamath.org/ileuni/exmidsbthr.html .

Oh, I see. But you linked to https://us.metamath.org/ileuni/exmidsbth.html above. To avoid the confusion of not knowing which one to quote or mention, I think it would better to have a single entrypoint: either make ~exmidsbthr a mere lemma, or even better, inline it in the proof of exmidsbth.

jkingdon:
Uh, to bring the topic back to `ruc`, is it OK to merge this pull request?

benjub (approved):

PR #3061, opened 2023-02-23
Revise comment for exmidsbth

jkingdon:
Be more explicit about the forward and reverse directions and say a bit about what they mean.

This my attempt to take https://github.com/metamath/set.mm/pull/3059#issuecomment-1440063909 into account (including the fact that it is harder than perhaps it should be to realize it might be worth clicking through to https://us.metamath.org/ileuni/exmidsbthr.html ).

To me the comment here feels a bit wordy but I guess unless someone has concrete suggestions for fixing that (or something else) we should just merge it. We can always revise it more later if we want.


iset.mm
```diff
@@ -133932,11 +133932,17 @@ C_ if ( A. k e. suc J ( Q ` ( i e. _om
...
-       uses classical logic, which accepts excluded middle as an axiom.  Thus
-       the Schroeder-Bernstein Theorem is true in MPE; see
-       ~ https://us.metamath.org/mpeuni/sbth.html .  (Contributed by Jim
-       Kingdon, 13-Aug-2022.) $)
+       is Metamath 100 proof #25.  The forward direction ( ~ isbth ) is the
+       proof of the Schroeder-Bernstein Theorem from the Metamath Proof
+       Explorer database (in which excluded middle holds), but adapted to use
+       ` EXMID ` as an antecedent rather than being unconditionally true.  The
+       Metamath Proof Explorer proof can be found at
+       ~ https://us.metamath.org/mpeuni/sbth.html .
```
icecream17, line 133940:
```suggestion
       Explorer database (in which excluded middle holds), but adapted to use
       ` EXMID ` as an antecedent rather than being unconditionally true, as in
       the non-intuitionist proof
       ~ https://us.metamath.org/mpeuni/sbth.html .
```

jkingdon: Suggestion accepted.

benjub: I would have written "classical proof", but this is fine by me.

benjub (approved):

PR #3064, opened 2023-02-25
put fvsnun1/2 in deduction form

benjub:


jkingdon (approved):

wlammen:
The number of proof bytes in total INCREASES from 930 to 986 bytes.  So what is the reason to introduce the deduction form here?

wlammen (commented):
Proofs in total become more complex

benjub:
The reason is to adopt the more consistent way to state theorems for more advanced parts of set.mm, which is the deduction form. It also yields a more general theorem, which I will need in my mathbox.

**edit:** see the paragraph beginning with ""Deduction form" is the preferred form for theorems" in https://us.metamath.org/mpeuni/conventions.html

wlammen:
I also prefer the deduction form over the immediate form, as it is more versatile.  But before converting immediate to deduction form, with longer proofs as a consequence, I'd rather see a theorem ph => ( T. -> ph ) , and reverse, as companions to trut, first, to ease transition.

benjub:
> I'd rather see a theorem ph => ( T. -> ph ) , and reverse, as companions to trut, first, to ease transition.

I'm not sure I understand. These theorems are a1i and mptru.

wlammen:
You're right, I did not see mptru, and missed to recognize a1i as an easy means to add T. as an antecedent.  I just wondered what made the proofs longer (measured in proof bytes) than I had expected.

benjub:
Actually, if you look at the usages of mptru, many are due to the translation back from deduction style near the end of the proof. This, plus the uses of a1i at the begining of each proof branch (the "leaves"), plus probably syntactic construction of longer formulas (that is, formulas with the antecedent T. added), makes for the longer proofs.

wlammen:
OK, then nothing is missing to support deduction style.  So when should we keep (maybe in addition) the immediate form then?  How many theorems are needed, whose proofs look nicer without the extra antecedents?

wlammen (approved):
The 3 applications of fvsnun1 / 2 do not justify keeping the immediate form.

benjub:
Yes, in this case the number of usages does not justify keeping the inference form. I couldn't give a more general rule for now.

wlammen:
Now that I had a look at the proof of fvsnun2 as a HTML page, I wonder why the prerequisite fvsnun2.3 wasn't added the antecedent ph as well.  Line 3 of the proof is easily created using reseq1d, so the proof hadn't increased by line number here.

benjub:
That $e-hypothesis does not have antecedents because it acts as a "local definition" and typical usages do not require antecedents for these. This was discussed in the mailing list a while ago but I cannot retrieve it. Maybe @avekens and @tirix remember it better.

**edit:** actually, I may be in a situation where the addition of the antecedent would make using the theorem easier...

jkingdon:
> The 3 applications of fvsnun1 / 2 do not justify keeping the immediate form.

That was my thought process too, that having only three usages make this one a candidate for not having multiple forms.

avekens:
> Now that I had a look at the proof of fvsnun2 as a HTML page, I wonder why the prerequisite fvsnun2.3 wasn't added the antecedent ph as well. 

> That $e-hypothesis does not have antecedents because it acts as a "local definition" and typical usages do not require antecedents for these. This was discussed in the mailing list a while ago but I cannot retrieve it. Maybe @avekens and @tirix remember it better.

According to the Metamath book (page 97), a hypothesis of a theorem in deduction style "is either (1) an implication ... [usually  `ph ->`] or a definition ...". fvsnun2.3 is such a definition.

PR #3066, opened 2023-02-26
conventions: add latex info

benjub:
Add some latex info to ~conventions (ping @GinoGiotto)


set.mm
```diff
@@ -231,13 +231,14 @@ surrounded by whitespace (space, tab, CR, LF, or FF).
       file or has been already been inserted, it will not be inserted again.
 
 Inside of comments, it is recommended that labels be preceded with a tilde (~)
-and math symbol tokens be enclosed in grave accents (` `).  This way the LaTeX
+and math symbol tokens be enclosed in backticks (` `).  These tildes, tokens,
```
wlammen, line 234: The official name of U+0060 in the Unicode charts is **grave accent** (https://unicode.org/charts/PDF/U0000.pdf).  So I wonder whether replacing it with backtick is a good idea.  At least I would introduce it with its official name, then continuing with sth like 'or backtick, as we think is more common, and used in this document'

wlammen (changes requested):

set.mm
benjub: I followed https://en.wikipedia.org/wiki/Backtick, which also mentions "grave accent" and others but uses "backtick" as the main word. To me, "grave accent" is, like other accents, a diacritical mark https://en.wikipedia.org/wiki/Grave_accent : it is the same symbol, but it takes that name when combined with a letter, as in àèù (to mention only the ones on my keyboard), like the German umlaut.

wlammen: Wikipedia has this article https://en.wikipedia.org/wiki/Grave_accent, too.  The accent put on letters has its own code U+0300 and is officially called combining grave accent.

david-a-wheeler: The official name is "grave accent", but I think we should say "backtick", because @benjub is right, a lot of people will be confused if we use the official name.

wlammen: Is there really someone who got it wrong in the past? Ok then, why not : ... enclosed in backticks (U+0060).

set.mm
```diff
@@ -422480,6 +422481,16 @@ Class symbols for functions (e.g., ` abs ` , ` sin ` )
        ` ( V iEdg E ) ` for the edges of a graph ` G = <. V , E >. ` .
        </li>
 
+       <li><b>LaTeX definitions.</b>
+       Each token has a "LaTeX definition" which is used by metamath.exe to
+       output tex files.  When writing LaTeX definitions, contributors should
+       favor simplicity over perfection of the display, and should only use
+       core LaTeX symbols or symbols from the AMS packages (amssymb, amsmath)
+       and the mathtools package (exceptions to this should be discussed).  A
+       useful resource is <a href = https://www.ctan.org/pkg/comprehensive>
```
david-a-wheeler, line 422490: Please don't use:

~~~~
 <a href = https://www.ctan.org/pkg/comprehensive>
~~~~

In particular, no spaces around the "=", and enclose the value in double-quotes, like thisl

~~~~
<a href="https://www.ctan.org/pkg/comprehensive">
~~~~

I believe the [space-around-equals will *work* in typical browsers](https://www.w3schools.com/html/html5_syntax.asp) but it's not considered good practice, and almost certainly some processors won't handle it correctly.

david-a-wheeler:
I'm fine with "backtick", but I've suggested a change to the HTML above so it uses a more typical format (and avoids potential problems). With that change to the HTML I think it's ready to go.

set.mm
wlammen: I still think that official and scientific documentation should adhere to (thus support) standards where possible.  Sth along the lines of my second comment here would meet my expectations.  But if a majority thinks you can do completely without it, I will no longer object.

benjub: No strong opinion. Anyone ? @avekens ? @tirix ?

david-a-wheeler: I'd say whatever consensus is. However, I suggest including both terms, one as the "main term" and the other as an "alternate term" just to make sure people will know what's meant.

E.g.:

> and math symbol tokens be enclosed in grave accents aka backticks (` `).  This way the LaTeX ...

set.mm
```diff
@@ -422480,6 +422481,16 @@ Class symbols for functions (e.g., ` abs ` , ` sin ` )
        ` ( V iEdg E ) ` for the edges of a graph ` G = <. V , E >. ` .
        </li>
 
+       <li><b>LaTeX definitions.</b>
+       Each token has a "LaTeX definition" which is used by metamath.exe to
+       output tex files.  When writing LaTeX definitions, contributors should
+       favor simplicity over perfection of the display, and should only use
+       core LaTeX symbols or symbols from the AMS packages (amssymb, amsmath)
```
GinoGiotto, line 422488: Some latex definitions that I collected in set.mm require an additional package (and indeed produce errors). For example <code>\mathscr{S}_\mathscr{H}</code> needs the <code>mathrsfs</code> package to work.

I was thinking about doing a pull request with the package added, but after reading this I think I should make an issue request.

set.mm
wlammen: For me, David's suggestion (also known as) is still preferable.  But since the character is directly shown afterwards, I will go with that solution.  A similar discussion like this one is found here https://stackoverflow.com/questions/7857278/what-is-the-meaning-of-grave-accent-aka-backtick-quoted-characters-in-mysql.

set.mm
benjub: Added to the generated preambles: https://github.com/metamath/metamath-exe/pull/130/commits/f5b8c11c10ee090c178183b36eb2a2888ecd4cc4

set.mm
benjub: I wrote both

set.mm
benjub: And addded it in these conventions too.

wlammen (approved):

GinoGiotto (approved):

set.mm
tirix: Including both names is probably the best.
I prefer "aka" to "or", because "or" might let someone believe there are two alternatives, actually it's only different names for one and the same character.

set.mm
```diff
@@ -231,11 +231,12 @@ surrounded by whitespace (space, tab, CR, LF, or FF).
       file or has been already been inserted, it will not be inserted again.
 
 Inside of comments, it is recommended that labels be preceded with a tilde (~)
-and math symbol tokens be enclosed in grave accents (` `).  This way the LaTeX
-and HTML rendition of comments will be accurate, and (future) tools to globally
-change labels and math symbols will also change them in comments.  Note that
-inside of grave accents a pair of grave accents is interpreted as a single
-grave accent.  A special comment containing $ t (with no space after the dollar
+and math symbol tokens be enclosed in backticks, or grave accents (` `).  These
+tildes, tokens, math symbols and backticks should be surrounded by spaces.
```
avekens, line 235: Until now, it is not forbidden/discouraged to surround backticks by spaces in comments: Only after the opening and before the closing backtick a space is needed. There is no need to have a space before the opening and after the closing backtick, see, for example, ``wff variable (usually ` ph `) or ...`` or ``...including that wff variable (` ph `).``

This topic was already discussed elsewhere...

set.mm
```diff
@@ -422469,7 +422470,7 @@ that confuses text editors (such as emacs).
        <li><b>Typography conventions.</b>
        Class symbols for functions (e.g., ` abs ` , ` sin ` )
        should usually not have leading or trailing blanks in their
-       HTML/Latex representation.
+       HTML representation.
```
avekens, line 422473: Why are these typography conventions not valid for Latex representations? If not one-to-one, then there should be corresponding conventions for Latex.

avekens:
Unfortunately, this PR was closed/merged very quickly, so that there was not much time to respond... @tirix and I have still remarks (see above). How should they be handled?

wlammen:
`How should they be handled?`

I opened a new PR #3070 addressing one of the issues.

david-a-wheeler:
> How should they be handled?

In general, open a new PR (or a new issue if it's not clear what should be done & you want a general discussion). That way they won't be lost.

set.mm
benjub: Spaces have no effect in LaTeX's math mode (that is, `$a a$` and `$aa$` both give $aa$).

set.mm
benjub: Not forbidden, but the rewrapping automatically adds spaces there, so better ask contributors to put spaces, to avoid CI test fails.

avekens: > Not forbidden, but the rewrapping automatically adds spaces there, so better ask contributors to put spaces, to avoid CI test fails.

So for comments which are not rewrapped (section header comment, comments in HTML) spaces can be omitted.

set.mm
avekens: > Spaces have no effect in LaTeX's math mode (that is, `$a a$` and `$aa$` both give aa).

OK, but if spaces are required to improve the readability in LaTeX, they must be inserted/enforced explicitly (e.g., by \quad, \, etc.). See, for example the LaTeX representation of the operator `gcd`: latexdef "gcd" as "\,{\rm gcd}\,";
By this, it is ensured that the operation `( a gcd b )` is not displayed as "(agcdb)" , but as "(a gcd b)". Therefore, there must be a convention for LaTex  corresponding to the convention for HTML.

set.mm
benjub: Yes, a loophole in the "rewrap" command allows contributors to use non-consistent spacing. Not sure this is worth mentioning in that non-displayed comment.

avekens: > Yes, a loophole in the "rewrap" command allows contributors to use non-consistent spacing. 

Yes , the usage of spacing is not consistent in this case, but as far as I remember there was no agreement how it should be made consistenmt (to enforce a space before the opening and after the closing backtick or not in general).

> Not sure this is worth mentioning in that non-displayed comment.

OK, then I see no need to have changed this text at all.

But let's stop the discussion now. We can leave the text now as it is.

set.mm
benjub: > OK, but if spaces are required to improve the readability in LaTeX, they must be inserted/enforced explicitly (e.g., by \quad, \, etc.).

If that's for readibility of the generated latex code, then it's the job of the code generator to separate tokens with spaces, not for the definition of each token to contain spaces.

Enforcing spaces explicitly like you propose is bad practice (spaces don't scale well, for one).

For instance, `\,{\rm gcd}\,` is an example of what not to do. Since gcd is used in set.mm as a binary operation, it should be `\mathbin{\operatorname{gcd}}`.

Issue #3067, opened 2023-02-26
Fixing some LaTeX definitions


benjub:
This is an update on #3037. Some LaTeX definitions should be fixed in all the databases in this repository with LaTeX definitions. These are: `set.mm`, `iset.mm`, `hol.mm`, `ql.mm`, `nf.mm`.

The commands on the LHS below are deprecated, and here are some proposed replacements:
1. [x] `{\rm xxx}` --> `\mathrm{xxx}` (see #3054 for the only remaining cases)
2. [x] `{\cal xxx}` --> `\mathcal{xxx}`
3. [x] `{\sf xxx}` --> `\mathsf{xxx}`
4. [x] `{\Bbb xxx}` --> `\mathbb{xxx}`
5. [x] `{\bb xxx}` --> `\mathrm{xxx}`
6. [x] `\tt` --> depends; often `\mathtt{...}`
7. [x] `\it`, `\bf`, `\sc`, `\sl`: check that they do not occur in any of the databases (if they do, there is `\mathsc{...}`, etc.)
8. [x] `\atop` --> depends: \frac \genfrac \overset \stackrel...

(note that the expressions on the LHS above are not always surrounded by braces).

9. [ ] many `\mathrm{...}`  should actually be `\operatorname{...}`
10. [x] `\mbox{...}` should generally be `\text{...}`

See @GinoGiotto's comment https://github.com/metamath/set.mm/issues/3067#issuecomment-1450875967 for an advancement step.

GinoGiotto:
I believe most of these checks can be solved if the following few latexdefs are fixed:

<pre> latexdef "-1-1->" as
    "\raisebox{.5ex}{${\textstyle{\:}_{\mbox{\footnotesize\rm 1" +
    "\tt -\rm 1}}}\atop{\textstyle{" +
    "\longrightarrow}\atop{\textstyle{}^{\mbox{\footnotesize\rm {\ }}}}}$}";</pre>

<pre>latexdef "-onto->" as
    "\raisebox{.5ex}{${\textstyle{\:}_{\mbox{\footnotesize\mathrm{\:}}}}" +
    "\atop{\textstyle{" +
    "\longrightarrow}\atop{\textstyle{}^" +
    "{\mbox{\footnotesize\mathrm{onto}}}}}$}";</pre>

<pre>latexdef "-1-1-onto->" as
    "\raisebox{.5ex}{${\textstyle{\:}_{\mbox{\footnotesize\mathrm{1}" +
[...]

PR #3071, opened 2023-02-27
Replacing latexdefs

GinoGiotto:
I made some progress towards #3067, this pull request doesn't address all issues. I'll write a comment about what is left and what I think is already done.


GinoGiotto:
I noticed that latex editors don't add spacing for adjacent latexdefs to <code>\text{...}</code> . I propose to add some kind of spacing inside, like <code>\text{\\,text\\,}</code> or simply <code>\text{ text }</code>.

iset.mm
```diff
@@ -129413,7 +129413,7 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
   latexdef "F/_" as "\underline{\Finv}";
 htmldef "CondEq" as "CondEq";
   althtmldef "CondEq" as "CondEq";
-  latexdef "CondEq" as "\mbox{CondEq}";
+  latexdef "CondEq" as "\text{CondEq}";
```
benjub, line 129416:
```suggestion
  latexdef "CondEq" as "\operatorname{CondEq}";
```

iset.mm
```diff
@@ -129560,7 +129560,7 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
   althtmldef ".0b" as
     ' <SPAN CLASS=symvar STYLE="border-bottom:1px dotted;color:#C3C">' +
     '&#x1D7CE</SPAN> ';
-  latexdef ".0b" as "\mbox{\boldmath$0$}";
+  latexdef ".0b" as "\text{\boldmath$0$}";
```
benjub, line 129563:
```suggestion
  latexdef ".0b" as "\pmb{0}";
```

nf.mm
```diff
@@ -63228,32 +63228,32 @@ the first case of his notation (simple exponentiation) and subscript it
...
 htmldef "2nd" as
     "<IMG SRC='_2nd.gif' WIDTH=21 HEIGHT=19 TITLE='2nd' ALIGN=TOP>";
   althtmldef "2nd" as '2<SUP>nd</SUP> ';
-  latexdef "2nd" as "2^{\rm nd}";
+  latexdef "2nd" as "2^\mathrm{nd}";
 htmldef "Swap" as
     "<FONT FACE=sans-serif> Swap </FONT>";
   althtmldef "Swap" as '<FONT FACE=sans-serif> Swap </FONT>';
-  latexdef "Swap" as "{\rm Swap}";
+  latexdef "Swap" as "\mathrm{Swap}";
```
benjub, line 63248: can you remove the blank lines?

benjub:
I made a few comments. I saw some other "clumsy" latexdefs (among the unchanged ones), but work has to be done twice: in both set.mm and iset.mm...
This will be a large task, and tex ouputting is not used often (at least, not by me), so I'm not sure it's worth doing too much, especially as long as we do not have a way to share latexdefs between multiple databases.

GinoGiotto:
@benjub btw now I'm also unable to use the "commit suggestion" button, it gives me the same error you mentioned. I don't know why.

GinoGiotto:
> I noticed that latex editors don't add spacing for adjacent latexdefs to `\text{...}` . I propose to add some kind of spacing inside, like `\text{\,text\,}` or simply `\text{ text }`.

@benjub what's your opinion about this? What spacing would you suggest?

benjub:
> I noticed that latex editors don't add spacing for adjacent latexdefs to \text{...} . I propose to add some kind of spacing inside, like \text{\,text\,} or simply \text{ text }.
> @benjub what's your opinion about this? What spacing would you suggest?

Between both, the latter, although probably in most cases, something else is meant, e.g., `\mathbin{text}` or `\mathrel{text}`.

This PR looks good to me.

GinoGiotto:
> Between both, the latter, although probably in most cases, something else is meant, e.g., `\mathbin{text}` or `\mathrel{text}`.

@benjub I'm changing <code>\text{text}</code> into <code>\text{ text }</code>, until further analysis I think spacing is needed because those latexdefs together appear like a single word.

GinoGiotto:
Ok, unless there are further commit suggestions, I've finished here

benjub (approved):
LGTM as an intermediate step.

benjub:
If no one objects (or merges by then), I'll merge tonight UTC. Thanks @GinoGiotto for all this work.

PR #3072, opened 2023-02-28
prove euequ without ax-10 .. ax-12

wlammen:
removes dependency on ax-10, ax-11 and ax-12
I did not find outdated OLD theorems as of 28-Feb-2023.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15471,6 +15471,7 @@ New usage of "eubidvOLDOLD" is discouraged (0 uses).
 New usage of "eucrct2eupth1OLD" is discouraged (1 uses).
 New usage of "eucrct2eupthOLD" is discouraged (0 uses).
 New usage of "eueqOLD" is discouraged (1 uses).
+New usage of "euequOLD" is discouraged (0 uses).
 New usage of "euexALTOLD" is discouraged (0 uses).
 New usage of "euexOLD" is discouraged (0 uses).
 New usage of "eufOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
Oh, I see, that's nice to use both ax6ev and ax6evr.

By the way: can you prove "eu3" where phi can depend on both x and y ?

benjub (approved):

wlammen:
I don't know what "eu3" should look like.  The equivalent of mo3 is eu2.  Or should it end in ... <-> x = y?

benjub:
@wlammen: Is the statement of https://us.metamath.org/mpeuni/eu3v.html without the DV on ph, y true ? I would think so, since the comment for the revision says it was added. I think it would be nice to add it / add it back. In the current state, it feels like it is missing.

wlammen:
@benjub git checkout 95e76336a7b435353542bfd1d40040c5d5bcd97d restores the state of set.mm on 29-May-2019.  We see this theorem there

`  ${
    $d x y $.
    eu3OLD.1 $e |- F/ y ph $.
    $( Obsolete theorem as of 29-May-2018.  Superseded by ~ eu3v that better
       fits common usage pattern.  (Contributed by NM, 8-Jul-1994.)
       (Proof modification is discouraged.) $)
    eu3OLD $p |- ( E! x ph <->
                ( E. x ph /\ E. y A. x ( ph -> x = y ) ) ) $=
      ( weu wex wsb wa weq wi wal eu2 mo anbi2i bitr4i ) ABEABFZAABCGHBCIZJCKBK
      ZHPAQJBKCFZHABCDLSRPABCDMNO $.
  $} `

If you replace in the current definition of df-eu the term E* x ph with what mof yields, then you arrive at this theorem.  This is a trivial transformation, and I wonder whether this adds something really of value.

benjub:
Thanks, I see. Pretty straightforward indeed. Still, I think it'd be nice to add it (as eu3f ?) with cross-links in comments. This would prevent the risk of confusion seen in my previous comment (namely, you can remove the DV but you need the nonfreeness hypothesis).

PR #3074, opened 2023-03-01
Fixed malfunctioning latexdefs

GinoGiotto:
These latexdefs are malfunctioning for various reasons, specifically:

1. The <code>&</code> is the alignment character in latex, since in (i)set.mm it is indicated as <code>"and-also"</code> I guess it was meant to be <code>\\&</code>. 

2. The <code>\\^</code> was fixed for set.mm only in https://github.com/metamath/set.mm/pull/2939#issue-1468878153, here I replaced it for iset.mm too.

3.  It looks like <code>\Omicron</code> doesn't exist in latex (at least for standard packages), so it generates an error. I replaced with <code>O</code> according to https://www.overleaf.com/learn/latex/List_of_Greek_letters_and_math_symbols.

4. The <code>#</code> in <code>\\mathrm{#}_b</code> generates the error message <code>you can't use 'macro parameter character #' in math mode</code>, I guess it was meant to be <code>\\mathrm{\\#}_b</code>.

5. This <code>\\]_b</code> generates various error messages, I replaced it with <code>]_b</code>.

6. This <code>|-</code> has an empty latex translation in ql.mm, I replaced it with <code>\vdash</code> according to other databases.

7. For the remaining latexdefs in this PR I fixed some missing and unmatched brackets.

### DIFF
--- a/ql.mm
+++ b/ql.mm
@@ -10867,7 +10867,7 @@ latexdef "1" as "1";
 latexdef "'" as "'";
 latexdef "wff" as "\mathrm{wff}";
 latexdef "term" as "\mathrm{term}";
-latexdef "|-" as "";
+latexdef "|-" as "\vdash";
 latexdef "C" as "C";
 latexdef "," as ",";
 latexdef "=<" as "\le ";


iset.mm
```diff
@@ -129225,7 +129225,7 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
 htmldef "&" as
     " <IMG SRC='amp.gif' WIDTH=12 HEIGHT=19 ALT='&amp;'> ";
   althtmldef "&" as " &amp; ";
-  latexdef "&" as " & ";
+  latexdef "&" as " \& ";
```
benjub, line 129228:
```suggestion
  latexdef "&" as "\mathrel{\&}";
```

iset.mm
```diff
@@ -130381,7 +130381,7 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
   latexdef "..." as "\ldots";
 htmldef "..^" as "..^";
   althtmldef "..^" as "..^";
-  latexdef "..^" as "..\^";
+  latexdef "..^" as "\,..\hat{}\,";
```
benjub, line 130384:
```suggestion
  latexdef "..^" as "\mathbin{..\hat{}}";
```

benjub: Sorry, I meant: `"\mathbin{..\hat{\;}}";`

iset.mm
```diff
@@ -130628,16 +130628,16 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
 htmldef "DECID_in" as " <SMALL><U>DECID</U><SUB>in</SUB></SMALL> ";
   althtmldef "DECID_in" as " <SMALL><U>DECID</U><SUB>in</SUB></SMALL> ";
   latexdef "DECID_in" as
-           " \normalfont\textsc{\underline{decid}$_\mathrm{in}$}} ";
+           " \normalfont{\textsc{\underline{decid}$_\mathrm{in}$}} ";
```
benjub, line 130631:
```suggestion
           "\normalfont{\textsc{\underline{decid}}_\mathrm{in}}";
```

iset.mm
```diff
@@ -130628,16 +130628,16 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
...
 htmldef "Delta0" as "&Delta;<sub>0</sub>";
   althtmldef "Delta0" as "&Delta;<sub>0</sub>";
   latexdef "Delta0" as "\Delta_0 ";
 htmldef "Bdd" as "<SMALL>BOUNDED</SMALL> ";
   althtmldef "Bdd" as "<SMALL>BOUNDED</SMALL> ";
-  latexdef "Bdd" as "\normalfont\textsc{bounded}} ";
+  latexdef "Bdd" as "\normalfont{\textsc{bounded}}";
```
benjub, line 130637: not necessary but ok (as for the next one)

set.mm
```diff
@@ -432863,7 +432863,7 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 htmldef "&" as
     " <IMG SRC='amp.gif' WIDTH=12 HEIGHT=19 ALT='&amp;'> ";
   althtmldef "&" as " &amp; ";
-  latexdef "&" as " & ";
+  latexdef "&" as " \& ";
```
benjub, line 432866: see iset

set.mm
```diff
@@ -437541,10 +437541,10 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
  latexdef "/_f" as "\mathrm{/}_f";
 htmldef "_O" as '_O';
  althtmldef "_O" as "&#x39F;";
- latexdef "_O" as "\Omicron";
+ latexdef "_O" as "O";
```
benjub, line 437544: Why ? Is this for asymptotic comparisons, e.g. f(x) = \Omicron (x) ?

set.mm
```diff
@@ -437676,7 +437676,7 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 htmldef "->.." as " <IMG SRC='__tobold.gif' " +
     "WIDTH=15 HEIGHT=19 ALT=' -&gt;..' TITLE='-&gt;..'> ";
   althtmldef "->.." as " <B>&rarr;</B> ";
-  latexdef "->.." as "\text{\boldmath $\rightarrow$}}";
+  latexdef "->.." as "\text{\boldmath $\rightarrow$}";
```
benjub, line 437679:
```suggestion
  latexdef "->.." as "\pmb{\rightarrow}";
```

set.mm
GinoGiotto: Yes, in set.mm it's described as <code>the class of the "big-O" function</code>, but the reason I replaced it is because <code>\\Omicron</code> generates the error <code>undefined control sequence. \end{align}</code> in latex (see point 3 of the pull request description).

iset.mm
GinoGiotto: you mean the curly brackets or <code>\\normalfont</code>?

david-a-wheeler:
@GinoGiotto - thanks! Most people just look at the generated HTML (e.g., for the proofs), and we never set up tests for the LaTeX. As a result the TeX capabilties have quietly bitrotted, even though it's *mostly* there.

If we can get the TeX back into shape, I think we should add tests to check generation of PDFs from LaTeX in a number of statements. Taht would reduce the likelihood of such problems happening again. But we have to get things back into shape first :-).

set.mm
benjub: Ok. Then "\mathrm{O}".

iset.mm
benjub: No, actually you're right: your change is good and necessary.

Things like `$\frac1{\sqrt2}\mathrm a b$` work as expected, but it's unnecessary risky. (The way LaTeX works is that it knows that \frac takes two arguments, so it takes the next two tokens available.) Sometimes, I use things like `\frac12`, but that's it. When in doubt, put braces anyway.

benjub (approved):

GinoGiotto:
> If we can get the TeX back into shape, I think we should add tests to check generation of PDFs from LaTeX in a number of statements. Taht would reduce the likelihood of such problems happening again. But we have to get things back into shape first :-).

I would suggest to follow a path like this https://github.com/metamath/set.mm/pull/3039#issuecomment-1435811305, the advantage of this method is that the generated TeX file can only compile if no LaTeX substitution generates errors, this way you can find relatively quickly malfunctioning latexdefs and have some confidence that nothing is missed.

Issue #3054, opened 2023-02-20
Tokens and symbols for injections, surjections, and bijections


benjub:
Upon looking at #3053, I recalled how nightmarish the tokens and symbols for injection, surjection, and bijection are. Here is my suggestion:
* Tokens: `>-->`, `-->>`, `>-->>` (recall that the token for functions is `-->`).
* Latex: look at the code below and at its result in pdf:
```
\begin{gather*}
F \colon A \longrightarrow B\\
F \colon A \lhook\joinrel\longrightarrow B\\
F \colon A \longrightarrow\!\!\!\!\rightarrow B\\
F \colon A \lhook\joinrel\longrightarrow\!\!\!\!\rightarrow B
\end{gather*}
``` 
and the unicode versions of the above.

**edit:** here is a screenshot:

![functions](https://user-images.githubusercontent.com/41090811/220202933-6bd02209-ac12-4fe3-b4ad-d943f9f828c1.png)

benjub:
Here is a more detailed account:

![latex](https://user-images.githubusercontent.com/41090811/222970062-6d432b33-5052-4c14-9126-f9dc3879ac1a.png)

jkingdon:
I'm not a fan of this change.

Now, perhaps I'm biased because I learned this with the 1-1 and onto terminology rather than the injection and surjection terminology,[1] but I never can remember which kind of arrow corresponds with which concept - I always have to look it up.

I suppose it is a case of "The challenge of varying mathematical conventions" from https://us.metamath.org/mpeuni/conventions.html which does illustrate the problem but doesn't necessarily tell us how to resolve it. I can make an elaborate plea that the 1-1 and onto notation (with the text over the arrows) is far easier for beginners but even an assertion like that is hard to prove or disprove.

[1] it is long enough ago that I'm not sure I can find the name of the set theory text, but it wasn't metamath, it was an intro set theory textbook.

jkingdon:
According to https://us.metamath.org/mpeuni/df-f1o.html the notation with the 1-1 and onto above/below the arrow is from [[TakeutiZaring](https://us.metamath.org/mpeuni/mmset.html#TakeutiZaring)]

digama0:
> I'm not a fan of this change.
> 
> Now, perhaps I'm biased because I learned this with the 1-1 and onto terminology rather than the injection and surjection terminology,[1] but I never can remember which kind of arrow corresponds with which concept - I always have to look it up.

[...]

PR #3075, opened 2023-03-02
fix tex of injections, surjections and bijections

GinoGiotto:
These latex definitions contain deprecated latex https://github.com/metamath/set.mm/issues/3067 and produce errors https://github.com/metamath/set.mm/pull/3053#issuecomment-1445420003, I fixed them following this suggestion https://github.com/metamath/set.mm/issues/3067#issuecomment-1450904602

### DIFF
--- a/nf.mm
+++ b/nf.mm
@@ -63195,27 +63195,19 @@ htmldef "-1-1->" as
     "<IMG SRC='onetoone.gif' WIDTH=23 HEIGHT=19 TITLE='-1-1-&gt;' ALIGN=TOP>";
   althtmldef "-1-1->" as
     '&ndash;<FONT SIZE=-2 FACE=sans-serif>1-1</FONT>&rarr;';
-  latexdef "-1-1->" as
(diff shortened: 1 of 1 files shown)


benjub:
As I wrote several times recently, there is no need for that since I'm going to implement #3054 soon which will subsume it.

GinoGiotto:
Not even as a temporary solution? This would quickly eliminate all the exceptions in https://github.com/metamath/set.mm/issues/3067, making checks 1,6,8,10 fully complete as well as erasing all errors tex editors generate when compiling them. As stated in https://github.com/metamath/set.mm/issues/3067#issuecomment-1451077064 this can be updated with the improved version of https://github.com/metamath/set.mm/issues/3054 any time, in the meanwhile we have the guarantee that tex proofs will not malfunction.

benjub:
I'm not against it either. It's correct LaTeX, even if I know of no textbook using that notation.

GinoGiotto:
I agree about notaton. I don't know either why in metamath it was implemented in that way. I simply observed that this solution would fix all bugs and deprecations while preserving the same TeX output as before, so it looked good enough for me.

jkingdon (approved):

PR #3076, opened 2023-03-02
prove euae without ax-10 .. ax-13

wlammen:
1. remove dependencies on ax-10, ax-11, ax-12 and ax-13 from euae;
2. move hbnaev from my mathbox to main.  It is used in euae;
3. One direction of exsb and eu6 need fewer axioms than the reverse one.  Show this in exsb-1 and eu6-1.
4. Reword a comment to better explain how eu3v replaced eu3.
5. Remove outdated OLD theorems.

The new proof of euae is surprisingly long.  Maybe someone sees a way to shorten it.

@benjub bj-extru could be used in the proof of euae.  There is a TODO in its description that might be honored within this context.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 2-Mar-23 wl-hbnaev hbnaev      moved from WL's mathbox to main set.mm
 21-Feb-23 moimd     moimdv      align with mobidv, eubidv 
 16-Feb-23 bj-ldiv   ldiv        moved from BJ's mathbox to main set.mm
 16-Feb-23 bj-rdiv   rdiv        moved from BJ's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -4035,7 +4035,6 @@
 "clwwlkf1OLD" is used by "clwwlkf1oOLD".
 "clwwlkf1oOLD" is used by "clwwlkenOLD".
 "clwwlkf1oOLD" is used by "clwwlkvbijOLD".
-"clwwlkf1oOLD" is used by "clwwlkvbijOLDOLD".
 "clwwlkfOLD" is used by "clwwlkf1OLD".
 "clwwlkfOLD" is used by "clwwlkfoOLD".
 "clwwlkfoOLD" is used by "clwwlkf1oOLD".
@@ -12586,7 +12585,6 @@
 "swrd0fv0OLD" is used by "clwwlkinwwlkOLD".
 "swrd0fv0OLD" is used by "clwwlknonclwlknonf1oOLD".
 "swrd0fv0OLD" is used by "clwwlkvbijOLD".
-"swrd0fv0OLD" is used by "clwwlkvbijOLDOLD".
 "swrd0fv0OLD" is used by "numclwlk2lem2fOLD".
 "swrd0fv0OLD" is used by "numclwlk2lem2fOLDOLD".
 "swrd0fv0OLD" is used by "wwlksnextproplem1OLD".
@@ -14613,7 +14611,7 @@ New usage of "clwlkssizeeqOLD" is discouraged (0 uses).
 New usage of "clwlkssizeeqOLDOLD" is discouraged (0 uses).
 New usage of "clwwlkenOLD" is discouraged (0 uses).
 New usage of "clwwlkf1OLD" is discouraged (1 uses).
-New usage of "clwwlkf1oOLD" is discouraged (3 uses).
+New usage of "clwwlkf1oOLD" is discouraged (2 uses).
 New usage of "clwwlkfOLD" is discouraged (2 uses).
 New usage of "clwwlkfoOLD" is discouraged (1 uses).
 New usage of "clwwlkfvOLD" is discouraged (2 uses).
@@ -14625,14 +14623,11 @@ New usage of "clwwlknclwwlkdifsOLD" is discouraged (1 uses).
 New usage of "clwwlknonOLD" is discouraged (1 uses).
 New usage of "clwwlknonclwlknonenOLD" is discouraged (0 uses).
 New usage of "clwwlknonclwlknonf1oOLD" is discouraged (2 uses).
-New usage of "clwwlknondisjOLD" is discouraged (0 uses).
 New usage of "clwwlknonelOLD" is discouraged (0 uses).
 New usage of "clwwlknonwwlknonbOLD" is discouraged (0 uses).
-New usage of "clwwlknunOLD" is discouraged (0 uses).
 New usage of "clwwlknwwlknclOLD" is discouraged (0 uses).
 New usage of "clwwlknwwlksnOLD" is discouraged (0 uses).
 New usage of "clwwlkvbijOLD" is discouraged (0 uses).
-New usage of "clwwlkvbijOLDOLD" is discouraged (0 uses).
 New usage of "clwwnonrepclwwnonOLD" is discouraged (1 uses).
 New usage of "clwwnrepclwwnOLD" is discouraged (2 uses).
 New usage of "cm0" is discouraged (1 uses).
@@ -17915,7 +17910,7 @@ New usage of "suppfnssOLD" is discouraged (0 uses).
 New usage of "supsr" is discouraged (1 uses).
 New usage of "supsrlem" is discouraged (1 uses).
 New usage of "swrd0fOLD" is discouraged (2 uses).
-New usage of "swrd0fv0OLD" is discouraged (11 uses).
+New usage of "swrd0fv0OLD" is discouraged (10 uses).
 New usage of "swrd0fvOLD" is discouraged (11 uses).
 New usage of "swrd0fvlswOLD" is discouraged (8 uses).
 New usage of "swrd0len0OLD" is discouraged (1 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -22293,8 +22293,8 @@ the definition (which used to be ~ eu6 , while this definition was then
   ${
     $d x y $.  $d ph y $.
     $( An alternate way to express existential uniqueness.  (Contributed by NM,
-       8-Jul-1994.)  Add a disjoint variable condition on ` ph , y ` .
-       (Revised by Wolf Lammen, 29-May-2019.) $)
+       8-Jul-1994.)  Convert a non-free premise to a disjoint variable
+       condition on ` ph , y ` .  (Revised by Wolf Lammen, 29-May-2019.) $)
```
benjub, line 22297: I would prefer:
> Replace a nonfreeness hypothesis with a disjoint variable condition on ` ph , y ` to reduce axiom usage.

Ideed: 1) hypothesis is our preferred term for $e-statements  2) the hypothesis is not "nonfree", which makes no sense: it is a hypothesis of "nonfreeness" 3) we explain why we did that: "to reduce axiom usage".

Rk: I would still like the version with the nonfreeness hypothesis !

set.mm
```diff
@@ -17556,6 +17556,15 @@ variables and no wff metavariables (see ~ ax12wdemo for an example of
       ( hbaevg ) ABCBAD $.
   $}
 
+  ${
+    $d x y $.  $d u t z $.
+    $( Any variable is free in ` -. A. x x = y ` , if ` x ` and ` y ` are
+       distinct.  (Contributed by Wolf Lammen, 9-Apr-2021.) $)
```
benjub, line 17562: Is the version without DV(x,y) in set.mm ?  It would be worth adding it, with cross-links between the two.

set.mm
```diff
@@ -530290,6 +530299,14 @@ be tempted to generalize this (or a similar) theorem by weakening the
       ( weq wi wal ax12v2 sp com12 impbid ) BCDZAKAEZBFZABCGMKALBHIJ $.
   $}
 
+  ${
+    $d x y $.  $d y ph $.
+    $( One direction of the equivalence in ~ exsb is based on fewer axioms.
+       (Contributed by Wolf Lammen, 2-Mar-2023.) $)
+    wl-lem2-exsb $p |- ( E. y A. x ( x = y -> ph ) -> E. x ph ) $=
+      ( weq wi wal wex ax6ev exim mpi exlimiv ) BCDZAEBFZABGZCMLBGNBCHLABIJK $.
+  $}
```
benjub, line 530308: Worth moving to Main. What label ? exsbi ?

set.mm
```diff
@@ -530363,6 +530380,20 @@ be tempted to generalize this (or a similar) theorem by weakening the
...
+       2-Mar-2023.) $)
+    wl-eu6-1 $p |- ( E. y A. x ( ph <-> x = y ) -> E! x ph ) $=
+      ( vz weq wb wal wex wi wa weu 19.42v pm3.33 alanimi equvelv sylib pm4.71i
+      albiim biancomi exbii equequ2 imbi2d albidv anbi2d syl5bb pm5.32ri bitr4i
+      ax6evr biantru 3bitr4ri exdistrv bitri wl-lem2-exsb anim1i sylbi sylibr
+      eu3v ) ABCEZFBGZCHZABHZABDEZIZBGZDHZJZABKUTURAIZBGZCHZVEJZVFUTVHVDJZDHZCH
+      VJUSVLCUSCDEZJZDHUSVMDHZJVLUSUSVMDLVKVNDVKVKVMJVNVKVMVKURVBIZBGVMVGVCVPBU
+      RAVBMNCDBOPQVMUSVKUSVHAURIZBGZJVMVKUSVHVRAURBRSVMVRVDVHVMVQVCBVMURVBACDBU
+      AUBUCUDUEUFUGTVOUSDCUHUIUJTVHVDCDUKULVIVAVEABCUMUNUOABDUQUP $.
+  $}
```
benjub, line 530395: Worth moving to Main.
Rk: for this and eu6, wouldn't it be more consistent to write `x = y  <-> ph` ?

benjub: Also: I guess you have the same theorem as exsb with the reverse implication and E* ?

set.mm
wlammen: comment is updated

set.mm
wlammen: it is (hbnae).  links added

set.mm
wlammen: I'd rather leave ph <-> x = y for consistency with df-mo.
Moving wl-eu6-1 to main requires a rewrite, since eu6 and this theorem have 80% of the proof in common, so a lemma is called for.  I don't have the time to do that now.

set.mm
wlammen: I chose exsb-1

set.mm
wlammen: move done

wlammen: > Also: I guess you have the same theorem as exsb with the reverse implication and E* ?

The definition of E* contains everything necessary.  Nothing to prove.

benjub:
> @benjub bj-extru could be used in the proof of euae. There is a TODO in its description that might be honored within this context.

Thanks @wlammen for that remark. I'll apply the TODO this weekend.

As for the choices of labels exsb-1 and eu6-1, it seems to me that we try to avoid dashes in labels; but you're right that exsbi wasn't good since it's in closed form. Maybe @avekens has a proposal for these ?

set.mm
```diff
@@ -22321,19 +22341,34 @@ of the unique existential quantifier (note that ` y ` and ` z ` need not
...
-      CVGVCVFAVBHZBIZJVJVGVCVFVOAVBBUBUCVJVOUSVFVJVNURBVJVBUQACDBUDUEUFUGUHUIUJ
-      SVLVCDCUKULTSUPVIUTABCUMUNTUO $.
+      ( vz wex weq wi wal wa weu wb exsb anbi1i eu3v eu6lem 3bitr4i ) ABEZABDFG
+      BHDEZIBCFZAGBHCEZRIABJASKBHCEQTRABCLMABDNABCDOP $.
+
+    $( One direction of ~ eu6 needs fewer axioms.  (Contributed by Wolf Lammen,
+       2-Mar-2023.) $)
+    eu6-1 $p |- ( E. y A. x ( ph <-> x = y ) -> E! x ph ) $=
+      ( vz weq wb wal wex wi wa weu eu6lem exsb-1 anim1i sylbi eu3v sylibr ) AB
+      CEZFBGCHZABHZABDEIBGDHZJZABKSRAIBGCHZUAJUBABCDLUCTUAABCMNOABDPQ $.
```
benjub, line 22371: couldn't you prove eu6 from eu6-1, and inline/remove eu6lem ?

wlammen: You could ask the same question for exsb.  For this to have any effect, the proof should be based on sth like cb3v2.  Unfortunately you need the role of ph and ps reversed there, which finally drags in all of ax-10, ax-11 and ax-12 again.  I gave up after a few tries, and since the net effect of separating the directions is just extra complexity, so I abandoned that thought.

avekens:
> As for the choices of labels exsb-1 and eu6-1, it seems to me that we try to avoid dashes in labels; but you're right that exsbi wasn't good since it's in closed form. Maybe @avekens has a proposal for these ?

Maybe use ~exsim instead of ~exsb-1, and ~eu6im instead of ~eu6-1 ("im" for implication).

benjub:
> Maybe use ~exsim instead of ~exsb-1, and ~eu6im instead of ~eu6-1 ("im" for implication).

I guess you mean "~exsbim" ?  I like your proposals.

avekens:
> > Maybe use ~exsim instead of ~exsb-1, and ~eu6im instead of ~eu6-1 ("im" for implication).
> 
> I guess you mean "~exsbim" ? I like your proposals.

Hm, I thougth that "b" is the abbreviation for "biconditional" (see conventions), and since the discussed theorem is one direction of the biconditional, I suggested to replace "b" by "im". Then I thought that "sb" may stand for "substitution", but ~exsb has no substitution...

benjub:
> Hm, I thougth that "b" is the abbreviation for "biconditional" (see conventions), and since the discussed theorem is one direction of the biconditional, I suggested to replace "b" by "im". Then I thought that "sb" may stand for "substitution", but ~exsb has no substitution...

There is substitution in disguise: the theorem says: `E. y [y/x] ph -> E. x ph` but maybe another label should be found indeed.

avekens:
> > Hm, I thougth that "b" is the abbreviation for "biconditional" (see conventions), and since the discussed theorem is one direction of the biconditional, I suggested to replace "b" by "im". Then I thought that "sb" may stand for "substitution", but ~exsb has no substitution...
> 
> There is substitution in disguise: the theorem says: `E. y [y/x] ph -> E. x ph` but maybe another label should be found indeed.

Ah, I see:  By ~sb6  `( [ y / x ] ph <-> A. x ( x = y -> ph )` , both versions of the theorem are equivalent, therefore the suffix "sb" is justified. Then, ~exsbim should actually be used instead of ~exsb-1.

set.mm
benjub: > By the way, what does rk stand for? Roundhouse kick?

:laughing: I meant "remark", but I have no idea how much it's used; maybe "rmk" or "rem" is better, but then "remark" is not much longer anyway so I could type it in full !

set.mm
```diff
@@ -20534,9 +20552,10 @@ requires either a disjoint variable condition ( ~ sb56 ) or a non-freeness
   nfae $p |- F/ z A. x x = y $=
     ( weq wal hbae nf5i ) ABDAECABCFG $.
 
-  $( All variables are effectively bound in a distinct variable specifier.
-     Lemma L19 in [Megill] p. 446 (p. 14 of the preprint).  (Contributed by NM,
-     13-May-1993.) $)
+  $( All variables are effectively bound in a distinct variable specifier.  A
+     version with a distinct variable condition based on fewer axioms is
+     ~ hbaev .  Lemma L19 in [Megill] p. 446 (p. 14 of the preprint).
```
benjub, line 20557:
```suggestion
     ~ hbnaev .  Lemma L19 in [Megill] p. 446 (p. 14 of the preprint).
```

avekens (approved):

PR #3077, opened 2023-03-04
adding lindsadd to mathbox

Achaemenid:


jkingdon (approved):

benjub:
It looks like this PR was merged 17 minutes after being opened ??

The comment does not match the statement. The statement is closer to: "In a vector space, the union of an independent set and a vector not in its span is an independent set."

jkingdon:
> It looks like this PR was merged 17 minutes after being opened ??

To be honest, I misread 17 minutes as being 17 hours. Although that it is mathbox only was the bigger factor.

> The comment does not match the statement. The statement is closer to: "In a vector space, the union of an independent set and a vector not in its span is an independent set."

@Achaemenid maybe you can make a follow up pull request if you think it is warranted?

PR #3080, opened 2023-03-04
biadan

benjub:
The existing 2-$e-hypothesis ~biadan2 can be promoted to a 1-$e-hypothesis equivalence  (this is ~biadani), which can in turn be promoted to a closed-form equivalence (this is ~ biadan). Hence the relabeling biadan2-->biadanii. (For some reason, there was no statement labeled biadan, so I conveniently used this label).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13836,8 +13836,8 @@ New usage of "bdophmi" is discouraged (2 uses).
 New usage of "bdophsi" is discouraged (3 uses).
 New usage of "bdopln" is discouraged (9 uses).
 New usage of "bdopssadj" is discouraged (4 uses).
-New usage of "biadaniALT" is discouraged (0 uses).
 New usage of "biadan2OLD" is discouraged (0 uses).
+New usage of "biadaniALT" is discouraged (0 uses).
 New usage of "bitr3VD" is discouraged (0 uses).
 New usage of "bj-0" is discouraged (1 uses).
 New usage of "bj-1" is discouraged (0 uses).
@@ -18472,8 +18472,8 @@ Proof modification of "bamalipOLD" is discouraged (26 steps).
 Proof modification of "barbariALT" is discouraged (22 steps).
 Proof modification of "barocoALT" is discouraged (24 steps).
 Proof modification of "bcsiALT" is discouraged (444 steps).
-Proof modification of "biadaniALT" is discouraged (28 steps).
 Proof modification of "biadan2OLD" is discouraged (17 steps).
+Proof modification of "biadaniALT" is discouraged (28 steps).
 Proof modification of "bitr3VD" is discouraged (36 steps).
 Proof modification of "bj-0" is discouraged (5 steps).
 Proof modification of "bj-1" is discouraged (5 steps).
--- a/discouraged
+++ b/discouraged
@@ -13836,6 +13836,7 @@ New usage of "bdophmi" is discouraged (2 uses).
 New usage of "bdophsi" is discouraged (3 uses).
 New usage of "bdopln" is discouraged (9 uses).
 New usage of "bdopssadj" is discouraged (4 uses).
+New usage of "biadaniALT" is discouraged (0 uses).
 New usage of "bitr3VD" is discouraged (0 uses).
 New usage of "bj-0" is discouraged (1 uses).
 New usage of "bj-1" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -7233,12 +7233,41 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
...
-    biadan2.2 $e |- ( ps -> ( ph <-> ch ) ) $.
-    $( Add a conjunction to an equivalence.  (Contributed by Jeff Madsen,
-       20-Jun-2011.) $)
-    biadan2 $p |- ( ph <-> ( ps /\ ch ) ) $=
+  $( An implication is equivalent to the equivalence of some implied
+     equivalence and some other equivalence.  A utility lemma as illustrated in
+     ~ biadanii and ~ elelb .  (Contributed by BJ, 4-Mar-2023.) $)
+  biadan $p |- ( ( ph -> ps ) <->
+                   ( ( ps -> ( ph <-> ch ) ) <-> ( ph <-> ( ps /\ ch ) ) ) ) $=
+    ( wi wa pm5.32 bicom bitri biass mpbir pm4.71 biancomi bitr4i bibi2i mpbi
```
wlammen, line 7241: Suggest to use pm4.71r and remove the line with biancomi

wlammen (approved):

set.mm
```diff
@@ -524358,6 +524388,26 @@ to being a member of a class abstraction (that every class is equal to a
...
+
+  $( Equivalence between two common ways to characterize elements of a class
+     ` B ` : the LHS says that sets are elements of ` B ` if and only if they
+     satisfy ` ph ` while the RHS says that classes are elements of ` B ` if
+     and only if they are sets and satisfy ` ph ` .  Therefore, the LHS is a
+     characterization among sets while the RHS is a characterization among
+     classes.  Note that the LHS is often formulated using a class variable
+     instead of the universe ` _V ` while this is not possible for the RHS
+     (apart from using ` B ` itself, which would not be very useful).
+     (Contributed by NM, 26-Feb-2023.) $)
```
avekens, line 524405: Oops, the contribution date cannot be correct...

set.mm
```diff
@@ -7233,12 +7233,41 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
...
-  ${
-    biadan2.1 $e |- ( ph -> ps ) $.
-    biadan2.2 $e |- ( ps -> ( ph <-> ch ) ) $.
-    $( Add a conjunction to an equivalence.  (Contributed by Jeff Madsen,
-       20-Jun-2011.) $)
-    biadan2 $p |- ( ph <-> ( ps /\ ch ) ) $=
+  $( An implication is equivalent to the equivalence of some implied
+     equivalence and some other equivalence.  A utility lemma as illustrated in
+     ~ biadanii and ~ elelb .  (Contributed by BJ, 4-Mar-2023.) $)
+  biadan $p |- ( ( ph -> ps ) <->
```
avekens, line 7239: The label does not fit to the description anymore. Formerly it was

> Add ("ad") a conjunction ("an") to an equivalence ("bi").

set.mm
```diff
@@ -7233,12 +7233,41 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
...
+    $( Inference associated with ~ biadani .  Add a conjunction to an
+       equivalence.  (Contributed by Jeff Madsen, 20-Jun-2011.)  (Proof
+       shortened by BJ, 4-Mar-2023.) $)
+    biadanii $p |- ( ph <-> ( ps /\ ch ) ) $=
+      ( wb wi wa biadani mpbi ) BACFGABCHFEABCDIJ $.
+
+    $( Obsolete proof of ~ biadanii as of 4-Mar-2023.  (Contributed by Jeff
+       Madsen, 20-Jun-2011.)  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    biadaniiOLD $p |- ( ph <-> ( ps /\ ch ) ) $=
```
avekens, line 7270: Only formalism: since the old theorem was ~biadan2, this deprecated theorem should be called "biadan2OLD".

set.mm
benjub: bad copy-paste... will fix.

set.mm
benjub: Do you have a proposal ? (not too long)

set.mm
benjub: done.

set.mm
avekens: Maybe to enhance thew comment would be the simplest way to solve this issue:

> An implication is equivalent to the equivalence of some implied equivalence and an equivalence with a conjunction added.

benjub: The label should depend only on the statement, so a change in the comment should not affect it. I had begun writing a more precise comment, but it was basically paraphrasing the statement, adding nothing to it, so I settled to a shorter one. I'll see what I can do.

benjub: Done:
An implication is equivalent to the equivalence of some implied equivalence and some other equivalence involving a conjunction.

avekens (approved):

PR #3082, opened 2023-03-04
remove ax-10, ax-11 from exists2

wlammen:
.. and shorten it, to boot.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15503,6 +15503,7 @@ New usage of "exidu1" is discouraged (3 uses).
 New usage of "exinst" is discouraged (1 uses).
 New usage of "exinst01" is discouraged (1 uses).
 New usage of "exinst11" is discouraged (1 uses).
+New usage of "exists2OLD" is discouraged (0 uses).
 New usage of "exlimexi" is discouraged (2 uses).
 New usage of "exmoOLD" is discouraged (0 uses).
 New usage of "exmoeuOLD" is discouraged (0 uses).
@@ -19120,6 +19121,7 @@ Proof modification of "exbiriVD" is discouraged (70 steps).
 Proof modification of "exinst" is discouraged (12 steps).
 Proof modification of "exinst01" is discouraged (16 steps).
 Proof modification of "exinst11" is discouraged (21 steps).
+Proof modification of "exists2OLD" is discouraged (61 steps).
 Proof modification of "exlimexi" is discouraged (15 steps).
 Proof modification of "exmoOLD" is discouraged (22 steps).
 Proof modification of "exmoeuOLD" is discouraged (39 steps).


benjub:
Looking at the theorem, I realize that the comment does not match the statement. A stronger statement, which is closer to the comment, is
```
|- ( ( E. x ph /\ E. x -. ph ) -> -. E* x T. )
```
could you prove it @wlammen ?

And I woud replace euae by "moae", which is used to prove the theorem above.

wlammen:
@benjub Not really necessary, since ax-6 (or exgen) guarantees at least one thing exists, so the focus here is on the reduction to exactly one object.  Maybe a hint in the comment is sufficient here?
The dissection ` E! x x = x <-> (E. x x = x /\ E* x x = x ) ` will do most of the trick, though.  moae is more or less line 14 in euae, so it is already a subproof of it.

benjub:
moae is the more fundamental theorem, and is true on the empty domain. Then, euae is just a straightforward consequence of that plus ax-6. So moae should be in set.mm. In my opinion, it should replace euae. If you prefer, both can coexist.

wlammen:
No, moae would need ax-6, so it is equivalent to euae.  At this point, the empty domain is no option any more.

benjub:
> No, moae would need ax-6, so it is equivalent to euae. At this point, the empty domain is no option any more.

"No" what ?  I wrote that moae is true on the empty domain and this is indeed the case (even if in our axiomatization, ax-6 is needed).

wlammen:
If you neither have ax-6 nor ax-7 (in form of ax7 or equid) available, the operation = is more or less an arbitrary operation on whatever objects are in your universe.  In such a situation the semantics of ` E* x ` or ` E! x ` is undetermined.  It may not relate to a count of objects any more.  In the context of exists2, in my eyes, such generalization is pointless.  I'd rather refrain from extending tentatively in these directions.  Whether you replace aev with an ` E* x T ` version is a different question that need not be dealt with here.

I understand why you arrived at your suggestion, so a clarifying word might be appropriate in some comment.

avekens (approved):
The current modifications in thois PR are fine, the rest of the discussion may result in another PR.

benjub:
> The current modifications in thois PR are fine, the rest of the discussion may result in another PR.

Yes @wlammen's changes are for the better, so this can be merged. Simply, the comment or statement of exists2 could have been fixed, since they currently don't match.

benjub:
> If you neither have ax-6 nor ax-7 (in form of ax7 or equcomiv) available, the operation = is more or less an arbitrary operation on whatever objects are in your universe.

Yes, this is why I revised ~euae in 2022 to use `T.` (instead of, I suppose, `x = x`). And this is why I proposed in my first post above to use it too in ~ exists2. (This is also why I proposed https://us.metamath.org/mpeuni/bj-df-v.html a few years ago.)

> In such a situation the interpretation of E* x or E ! x or even the expression T in its current definition is completely undetermined.

The semantics of `T.` is completely characterized by ~tru.

More generally, the semantics of all these connectives and quantifiers is just the standard semantics of FOL, which is a given.  As axioms are added, we can prove more and more things, which are of course in line with the semantics. In other words, as axioms are added, we keep soundness, and we progress toward completeness.

In the set.mm axiomatization, equality is used in an essential way to do away with the notion of bound variable (that's Tarski's discovery), so the related axioms for equality show up, but this is not a reason not to state and prove the more fundamental results.  In the present case, `moae |- ( E* x T. <-> A. y y = z ) , DV(y,z)` is universally valid, including on the empty domain, while `euae |- ( E! x T. <-> A. x x = y ) , DV(x,y)` is not valid on the empty domain.

I'll add ~moae in a future PR.

wlammen:
1. 
> The semantics of T. is completely characterized by ~tru.

Yes, and I corrected my post independently and prior to your answer here.
2. My time contingent is limited, as I write and organize software in my profession, which occupies me most of the time.  I am no logician, or mathematician with an education.  Both limit my horizon, which is strictly the contents of set.mm.  So for me the definition of E* in set.mm is what my postings are about, and I simply cannot argue about anything outside of it.

PR #3084, opened 2023-03-05
dffltz (3rspcedvd, negn0nposznnd, oexpreposd) [mathbox]

icecream17:
One equivalence for Fermat's Last Theorem. This took a long time so  that's why I haven't proved more equivalences yet.

Although this particular "definition" (and future "definitions") uses  quantifiers, I would expect practical theorems to be more like

```
flt3d.1 $e |- ( ph -> A e. NN )
flt3d.2 $e |- ( ph -> B e. NN )
flt3d.3 $e |- ( ph -> C e. NN )
flt3d $p |- ( ph -> ( ( A ^ 3 ) + ( B ^ 3 ) ) =/= ( C ^ 3 ) )
```


david-a-wheeler:
Overall looks nice, it's a small step but you can get somewhere with small steps. A few comments:

* I don't understand why the "Relative version of Russell's paradox ~ ru" was modified (removing z).
* In the long term we might want to move some of the utility functions up into the main body, but that's a separate discussion.
* You need to rewrap set.mm. The scripts/rewrap script will do that, or you can do it manually as documented in our conventions. Easily fixed via automation. That's why one of the verifiers is failing.

icecream17:
Oh, for ~ ru, that happened to be an extra dv condition, so it was removed (same for renegadd). Rewrapped now

set.mm
```diff
@@ -615339,6 +615436,30 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
     VEWMTWGTWDEEAVSWEWFWONNJWETWFTVRGAWDERQBWNVDKVREGWDEWNNNWDTVFSVGSAEBHCVHUJV
     IVJVKIVLUPUQUNOURVM $.
 
+  ${
+    oexpreposd.n $e |- ( ph -> N e. RR ) $.
+    oexpreposd.m $e |- ( ph -> M e. NN ) $.
+    oexpreposd.1 $e |- ( ph -> -. ( M / 2 ) e. NN ) $.
+    $( Lemma for ~ dffltz .  (Contributed by Steven Nguyen, 4-Mar-2023.) $)
+    oexpreposd $p |- ( ph -> ( 0 < N <-> 0 < ( N ^ M ) ) ) $=
```
benjub, line 615444: I haven't looked at how this is used, but it seems more natural to state
```
oexpreposd.m $e |- ( ph -> M e. NN0 ) $.
oexpreposd $p |- ( ph -> ( 0 < N <-> 0 < ( N ^ ( ( 2 x. M ) + 1 ) ) ) ) $=
```

set.mm
```diff
@@ -615019,6 +615019,92 @@ fixed reference functional determined by this vector (corresponding to
...
+      ( wrex cv wceq wa 2rexbidv rexbidv rspcedvd ) ABHLTGLTCHLTZGLTFILMAFUAIUB
+      UCBCGHLLPUDAUGDHLTGJLNAGUAJUBUCCDHLQUEADEHKLORSUFUFUF $.
+  $}
+
+  $( The proper subset of a set is also a set.  (Contributed by Steven Nguyen,
+     17-Jul-2022.) $)
+  pssexg $p |- ( ( A C. B /\ B e. C ) -> A e. _V ) $=
+    ( wpss wss wcel cvv pssss ssexg sylan ) ABDABEBCFAGFABHABCIJ $.
+
+  $( A proper superset is nonempty.  (Contributed by Steven Nguyen,
```
benjub, line 615069:
```suggestion
  $( A proper superclass is nonempty.  (Contributed by Steven Nguyen,
```

set.mm
icecream17: It's used in dffltz, here's the (limitedly useful) generate webpage -> print but save as pdf visual https://github.com/icecream17/Stuff/blob/master/dffltz.pdf

Basically the way I proved this is to set `x = if ( ( n / 2 ) e. NN , ( abs ` a ) , if ( 0 < a , if ( 0 < b , a , if ( 0 < c , -u b , a ) ) , if ( 0 < b , if ( 0 < c , -u a , b ) , -u a ) ) ), y = ..., z = ...`. and then splitting out the if statements using theorems like `ph /\ ps -> ch` and `ph /\ -. ps -> ch` and collapsing the if statements per branch using iftrue and iffalse.

It was convenient to have the ( n / 2 ) e. NN for a ^ ( 2 * ( n / 2 ) ) purposes... but I did notice how many theorems were of a different format and how oexpreposd is way longer than it should be. And while checking I just noticed nneo (!) which at least replaces that last hypothesis with `( ( M + 1 ) / 2 ) e. NN`. In the case of the current x = ... y = ... z = ... in dffltz though, this would just add an extra conversion step (the first condition is `( n / 2 ) e. NN` which is very similar to df-even).

As a sidenote, both of these are apparently not the current even/odd convention...
https://github.com/metamath/set.mm/issues/2433

set.mm
icecream17: https://us.metamath.org/mpeuni/df-ss.html says subset is used instead of subclass, so it kinda applies to supersets too.

ctrl+f finds 63 "superset" and 6 "superclass" (and 1378 "subset" and 374 "subclass")

benjub: As you prefer.

avekens:
Proof of ~dffltz :
I think there are essential 4 combinations (` ( a ^ n ) ` and ` ( b ^ n ) ` can be exchanged because the addition is commutative), so maybe the proof can be shortened using this fact (perhaps by providing 2 lemmata for these cases).

icecream17:
Not sure how to incorporate that yet, but I think it definitely could shorten the proof especially since:

```
a + b = c
a + -b = c
a + -b = -c  --> -a + b = c (commuted + var change: a + -b = c)
-a + -b = -c --> a + b = c
```

In the bottom two cases, there are two or more negatives. So negating all variables reduces the cases to 2.

david-a-wheeler:
All checks pass, & this is in a mathbox. I think we can bring it in. It'd be nice to shorten the proof as discussed above, and feel free to do so, but that can also be done as a separate request (we *often* shorten proofs later).

david-a-wheeler (approved):

PR #3086, opened 2023-03-06
hbnaevg, GL motivated a change in comment in wl-moae

wlammen:


wlammen:
Moved exsbim up to the ax-6 section, close to the structurally related equs4v

set.mm
```diff
@@ -529865,10 +529874,13 @@ extends this idea to multiple additions (TODO).
...
        equivalently, "exactly one thing exists" .  The equivalence results from
        the presence of ~ ax-6 in the proof, that ensures "at least one thing
-       exists".  For other equivalences see ~ wl-euae and ~ exists1 .
-       (Contributed by NM, 5-Apr-2004.)  State the theorem using truth constant
-       ` T. ` .  (Revised by BJ, 7-Oct-2022.)  Reduce axiom dependencies, and
-       use ` E* ` .  (Revised by Wolf Lammen, 5-Mar-2023.) $)
+       exists".  For other equivalences see ~ wl-euae and ~ exists1 .  Gerard
+       Lang pointed out, that ` E. y A. x x = y ` with disjoint ` x ` and ` y `
+       ( ~ df-mo , ~ trut ) also belongs to the expressions meaning "exactly
+       one thing exists" in the presence of ~ ax-6 .  (Contributed by NM,
```
benjub, line 529880:
```suggestion
       exists".  For other equivalences see ~ wl-euae and ~ exists1 .  Gerard
       Lang pointed out that ` E. y A. x x = y ` with disjoint ` x ` and ` y `
       ( ~ df-mo , ~ trut ) also means "exactly one thing exists".
       (Contributed by NM,
```
When "meaning" is involved, we are on the "semantics" side, which is given independently from the axioms.

On the other hand, the proof of `( E! x T. <-> E. y A. z z = y ) , DV(y,z)` probably requires ax-6 indeed. Do you plan to prove it ? It would be nice; if you don't try, I'll try it later. (maybe it requires some auxiliary axioms, in which case only prove it with DV(x,y,z), which does not require auxiliary axioms)

wlammen: wl-moae, wl-euae and df-mo manifest mutual equivalences among E* T., E! x T., A. x x = y and E. y A. x x = y d(x,y).  This is a matter of bitri applications.  For me this is easy to see, don't know how others perceive the situation.

benjub (approved):

PR #3087, opened 2023-03-07
shorten biadan, mathbox update

wlammen:
.. and remove outdated OLD theorems

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4039,7 +4039,6 @@
 "clwwlkfvOLD" is used by "clwwlkfoOLD".
 "clwwlkinwwlkOLD" is used by "clwwnrepclwwnOLD".
 "clwwlknOLD" is used by "isclwwlknOLD".
-"clwwlknclwwlkdifsOLD" is used by "clwwlknclwwlkdifnumOLD".
 "clwwlknonOLD" is used by "isclwwlknonOLD".
 "clwwlknonclwlknonf1oOLD" is used by "clwwlknonclwlknonenOLD".
 "clwwlknonclwlknonf1oOLD" is used by "dlwwlknondlwlknonf1oOLD".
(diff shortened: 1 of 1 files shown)


benjub:
Proposal:
```
  ${
    $d y z $.  $d t x $.
    wl-moae $p |- ( E* x T. <-> A. y y = z ) $=
      ( vt wtru wmo weq wal wl-motae wi wex hbaevg 19.2d eximi syl df-mo sylibr
      ala1 impbii ) EAFZBCGBHZBBCAIUAEADGZJAHZDKZTUAUBAHZDKUDUAUEDBCDDALMUEUCDU
      BEARNOEADPQS $.
  $}
```
Contributed by BJ and WL ? Or Contrib by WL, revised by BJ ?

wlammen:
Among the equivalent expressions (ax-6 assumed) ` E* x T. <-> E! x T. <-> A. x x = y  d(x,y) `  the ae expression is the working horse.  It allows for easiest manipulation.  We know by aev aev2 spaev that it is free of any setvar, and that we can add or drop quantors at will, without effectively changing it.  One can still round out the theory a bit wrt to ` E. ` that is used in df-mo.  The E* or E! expressions are perhaps more intuitive, so I understand the idea of focusing on them, after getting the basics done with ae.  In euae and friends the equivalence of  `E! x T.` to ae is expressed.  I favor E! over E*, because it is a bit more honest given the role of ax-6 in Metamath, and it visually restricts itself to one case, the one element universe.

The empty domain is a different case.  I am not much bothered here, as we strive for set theory, where things must exist in any meaningful context.  But I receive mails in private expressing an interest in that very case.  Maybe, it is best singled out using a `-. E. x x = y` antecedent.

If you want to move theorems from my mathbox to main, I suggest to use the common procedure and add your tag, if modified.  As for moae, I suggest to copy the subproof in euae (lines 3 - 13), and leave it at that.  Once the equivalence between E* and ae is established, all of aev, ... automatically transfers.  Creating theorems to this end is a matter of convenience, not necessity.

Once I know you won't use `E* x T.` theorems in my mathbox, I am going to delete them

benjub:
I regret that we are not exactly on the same page regarding these matters, but I'll go from there.

benjub (approved):

PR #3088, opened 2023-03-08
Intuitionize "Examples of topologies" section

jkingdon:
I didn't realize how little of this section would make it.  We get the discrete topology, the singleton of the empty set, and the excluded point topology, and that's it.

None of the indiscrete topology, the finite complement topology, the countable complement topology, or the particular point topology intuitionize in any kind of easy way. I think it is possible that some or all of them cannot be done, although my (somewhat limited) efforts to show that them being a topology implies excluded middle didn't succeed either. My efforts to find references on constructive topology did find references to topics like the compactness of `[0, 1]` which is sort of related but does not directly answer this question.

Anyway, this pull request intuitionizes the theorems which readily can be and notes the others in mmil.html so it is the right course of action for now.

There is some miscellaneous copying of theorems from set.mm and adding notes to mmil.html concerning theorems which are not in iset.mm but that's all routine stuff.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9181,6 +9181,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>cctop</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on theorems we don't have
+  including con1d , rexnal , and difindi .</TD>
+</TR>
+
+<TR>
   <TD>divccncf</TD>
   <TD>~ divccncfap</TD>
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9099,6 +9099,18 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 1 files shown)


digama0:
I think it is possible to do the indiscrete topology at least, by defining it to be the collection of all sets such that the membership relation does not depend on `x`, i.e. `{ s | A. x e. s s = X }`. (I'm guessing that you started from a definition `{ (/) , X }` and this isn't going to work intuitionistically.)

jkingdon:
> I think it is possible to do the indiscrete topology at least, by defining it to be the collection of all sets such that the membership relation does not depend on `x`, i.e. `{ s | A. x e. s s = X }`. (I'm guessing that you started from a definition `{ (/) , X }` and this isn't going to work intuitionistically.)

Thanks! I've moved this to https://github.com/metamath/set.mm/issues/3089 . It isn't particularly on the critical path to very many things, but would be nice to have.

jkingdon:
Is this pull request ready to merge? None of the topics in #3089 should affect whether what is here is mergeable.

avekens (approved):

PR #3092, opened 2023-03-12
some basic results about the empty domain

wlammen:
sparked off by recent mails and discussions.


jkingdon (approved):
Interesting experimentation. In particular, I'm curious about whether any of this is different in iset.mm, since this pull request seems to rely on excluded middle. (My hunch is that excluded middle is largely unrelated to whether there is at least one object, but I haven't done much trying to prove things without https://us.metamath.org/mpeuni/ax-6.html / https://us.metamath.org/ileuni/ax-i9.html and seeing where that leads).

wlammen:
The excluded middle only forces us to assign a truth value to an equation (in set.mm), if we decide to use equality at all.  We can characterize some empty domains by simply postulating the axiom -. E. x T.  But there is still a grey zone in between this axiom and ax-6.  One can use equality to find models, where neither axiom holds.  One example is: Assign true to all equations containing the set variable x on the left side, all others are false.  In this model E. x x = y holds as long as x is not replaced with another set variable, at which point the truth value of the expression flips.  In this model -. E. x T does not hold, since E. x x = y, but for set variables distinct from x it does.  A simple interpretation of this model is, that x is a special variable allowed to contain another type of object than other variables.  So this type of object exists, but other types not.

This all distracts us from set theory where such considerations don't play a role.

PR #3094, opened 2023-03-12
Intuitionize section "Closure and interior"

jkingdon:
The interior theorems intuitionize fairly easily. There isn't much need to do things like "a certain point is an element of a certain open set or it isn't" or the like.

The closed set and closure theorems mostly do not intuitionize, because the definition and theorems are so heavily based on complements.  It isn't clear that what is here will be especially useful, but might as well just put in what intuitionizes, note those that do not, and figure we can revisit at some point when trying to intuitionize theorems currently based on closed sets.

Includes some routine copying of theorems from set.mm to iset.mm.

Includes shortening two set.mm proofs because the intuitionized versions ended up being shorter than what is now in set.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9214,6 +9214,18 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <TR>
+  <TD>iincld</TD>
+  <TD><I>none</I></TD>
+  <TD>The set.mm proof relies on opncld</TD>
+</TR>
+
+<TR>
+  <TD>intcld</TD>
+  <TD><I>none</I></TD>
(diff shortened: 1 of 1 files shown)


avekens (approved):
I did only a formal check, and trust that the unchanged theorems are taken from set.mm without modifications.

jkingdon:
> I did only a formal check, and trust that the unchanged theorems are taken from set.mm without modifications.

I think I'm pretty good at making sure that if the git message says it is unmodified, it really is. So probably it should be fine on this point.

jkingdon:
Latest force push is just a rebase. I plan on merging this if the tests pass.

PR #3098, opened 2023-03-16
shorten spsbimv

wlammen:
.. and remove outdated OLD theorems.  I did not add a "Proof shortened" tag according to our convention not to stack up these tags when the same person modifies a proof in succession.


jkingdon (approved):

jkingdon:
> I did not add a "Proof shortened" tag according to our convention not to stack up these tags when the same person modifies a proof in succession.

Especially when it is only two months apart, I see no need for two tags.

Issue #3091, opened 2023-03-11
Malfunctioning Latexdefs left to fix


GinoGiotto:
Checks 2,3,4,5,6,7,8,10,11,12,14 are fixed in #3099
Checks 1, 9, 13 are fixed in https://github.com/metamath/metamath-exe/pull/133

This is a list of all faulty latex substitutions that I found + some minor questions I have. Hopefully no more errors will be produced by any latexdef after these are fixed:

**In set.mm:**
- [x] Token: <code>[. </code> Tex: <code>\underaccent{\dot}{[}</code>
Token: <code>]. </code> Tex: <code>\underaccent{\dot}{]}</code>
These need <code>\usepackage{accents}</code> to work

- [x] Token: <code>RPrime</code> Tex: <code>\mathsc{RPrime}</code>
Token: <code>UFD</code> Tex: <code>\mathsc{UFD}</code>
It seems that <code>\mathsc</code> doesn't exist, maybe it was meant to be <code>\textsc</code> or <code>\mathscr</code>?

- [x] Token: <code>_Ind</code> Tex: <code>\mathsfbf{1}</code>
Gives the error <code>undefined control sequence. \end{align}</code>, I don't know how to fix this one

- [x] Token: <code>log_</code> Tex: <code>log_</code> Already addressed [here](https://github.com/metamath/set.mm/issues/3079#issuecomment-1454942780)

- [x] Token : <code>(|</code> Tex: <code>\lParen</code>
 Token : <code>|)</code> Tex: <code>\rParen</code>
They where probably meant to be  <code>\lparen</code> and <code>\rparen</code> which are commands of the <code>mathtools</code> package

- [x] Token: <code>iota</code> Tex: <code>\mathrm{\rotatebox[origin=C]{180}{$\iota$}}</code>
Token: <code>iota_</code> Tex: <code>\underline{\mathrm{\rotatebox[origin=C]{180}{$\iota$}}}</code>
Token: <code>iota'</code> Tex: <code>\mathrm{\rotatebox[origin=C]{180}{$\iota$}}\textquoteright</code>
These latexdefs contain <code>\rotatebox</code> and <code>\iota</code> which should be replaced https://github.com/metamath/metamath-exe/pull/131

- [x] Token: <code>Xs_</code> Tex: <code>\text{\large\boldmath$\times$}\_s</code>
Token: <code>X_</code> Tex: <code>\text{\large\boldmath$\times$}</code>
[...]

PR #3099, opened 2023-03-16
Fixed Latexdefs

GinoGiotto:
1. Fixed checks 2,3,4,5,6,7,8,10,11,12,14 of  https://github.com/metamath/set.mm/issues/3091#issue-1620156800 following https://github.com/metamath/set.mm/issues/3091#issuecomment-1465044258
2. Removed unnecessary brackets and spaces.
3. Made some progress towards check 9 of  https://github.com/metamath/set.mm/issues/3067#issue-1600099492 using https://github.com/metamath/set.mm/issues/3067#issuecomment-1448082039 as a reference. This one is quite complicated since it would be necessary to check latex definitions one by one. Since I'm a beginner in Metamath notation and this task is time-consuming I'm not planning to make more progress on this in the future.
4. Applied https://github.com/metamath/set.mm/issues/3091#issuecomment-1471422634
5. This <code>latexdef "seqstr" as "\mathrm{sgn}_str";</code> defined as <code>Sequences defined by strong recursion.</code> seems a mistake  so I replaced it with <code>\mathrm{seq}_str</code>
6. Edited <code>\Sigma</code> to <code>\sum</code> for summation symbols.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -7890,7 +7890,6 @@
 "idi" is used by "smfinfmpt".
 "idi" is used by "smfsupmpt".
 "idi" is used by "ssmapsn".
-"idi" is used by "wspthnonOLD".
 "idiALT" is used by "ax6e2nd".
 "idiALT" is used by "ax6e2ndALT".
 "idiALT" is used by "ax6e2ndeqALT".
@@ -8535,7 +8534,6 @@
 "isvciOLD" is used by "hilvc".
 "isvclem" is used by "isvcOLD".
 "iswspthsnonOLD" is used by "wpthswwlks2onOLD".
-"iswspthsnonOLD" is used by "wspthnonOLDOLD".
 "iswwlksnonOLD" is used by "wpthswwlks2onOLD".
 "iswwlksnonOLD" is used by "wwlknonOLD".
 "iunconnlem2" is used by "iunconnALT".
@@ -13184,7 +13182,6 @@
 "wlkwwlksurOLD" is used by "wlkwwlkbijOLD".
 "wrdcctswrdOLD" is used by "2clwwlk2clwwlkOLD".
 "wrdcctswrdOLD" is used by "lencctswrdOLD".
-"wspthnonOLDOLD" is used by "wspthsnwspthsnonOLD".
 "wvd2" is used by "dfvd2".
 "wvd2" is used by "dfvd2i".
 "wvd2" is used by "dfvd2imp".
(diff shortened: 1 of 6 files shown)


set.mm
```diff
@@ -435480,14 +435479,14 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 */
 htmldef "CMod" as "CMod";
   althtmldef "CMod" as "&#8450;Mod";
-  latexdef "CMod" as "{\mathbb{C}\mathrm{Mod}}";
```
benjub: Here and below: the braces are indeed useless, but they show in a tex file that it comes from a single token, so could be useful to some scripts reading the TeX.

set.mm
```diff
@@ -438886,11 +438884,11 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 htmldef "_Se" as
     " <IMG SRC='_cse.gif' WIDTH=15 HEIGHT=19 ALT=' Se' TITLE='Se'> ";
   althtmldef "_Se" as " Se ";
-  latexdef "_Se" as "\mathrm{\;Se\;}";
+  latexdef "_Se" as "\mathrm{Se}";
```
benjub: aren't these binary relations ?
in which case, \mathrel{\operatorname{Se}}

set.mm
```diff
@@ -439054,7 +439052,7 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
   latexdef "FinSum" as " \mathrm{FinSum} ";
 htmldef "RRVec" as "RR-Vec";
   althtmldef "RRVec" as "&#8477;-Vec";
-  latexdef "RRVec" as "{\mathbb{R}-\mathrm{Vec}}";
+  latexdef "RRVec" as "\mathbb{R}-\mathrm{Vec}";
```
benjub: Please leave the braces in this one.

benjub:
Thanks for the big work. Check the boxes in the other issues if you're positive they're done. A few comments from me. If someone else approves, that's good with me.

GinoGiotto:
@benjub What does it mean the label "Outdated" in your comments? Like the requested changes are outdated?

wlammen:
It is added by Github and says the sources changed after the review was made.  So the review issue might be outdated.

set.mm
GinoGiotto: Ok, so I guess I put them back? Btw `latex_collector.py` writes every LaTeX definition separately in each line, so it doesn't need braces to notice that it comes from a single token, also there are no spaces in each of those substitutions in this pr, which I guess is good enough confermation that they come from a single definition since I doubt a script wouldn't separate them with a whitespace or something else.

david-a-wheeler:
Wonderful! Once these (and any other needed fixes) take place, we can add TeX checks to the CI pipeline, to ensure that future problems are likely to be detected immediately. I created draft #3100 to start (it doesn't work yet, and of course it won't *pass*, but I want to get that process started).

set.mm
benjub: It would be easier if you gave the label of the syntax definition.
If you're talking of https://us.metamath.org/mpeuni/wse.html and https://us.metamath.org/mpeuni/wfr.html, both are binary relations, so use `\mathrel`, since `R Se A` (and the other one), and not classes.

set.mm
benjub: > I doubt a script wouldn't separate them with a whitespace or something else.

You cannot make such an assumption.

GinoGiotto:
@benjub ready to merge?

benjub (approved):

PR #3101, opened 2023-03-18
Add theorems for countable sets to iset.mm

jkingdon:
As seen in the cited works and the theorems here, the usual definition of countable, when working without excluded middle, has to do with an mapping from natural numbers onto that set. So prove a few theorems using this definition, most notably that finite sets are countable.

Along the way:

* move nnpredcl form iset.mm mathbox to iset.mm main
* remove brrelexi in favor of brrelex1i (this change has already been made in set.mm, just moving it over to iset.mm)
* fix a longstanding todo for ctex in set.mm and iset.mm
* update comments in iset.mm to use countable only in the sense described here. Use "A is dominated by _om" for `A ~<_ _om`
* add a theorem that ``case ( F , G ) ` ( inl ` A ) ) = ( F ` A )`` because we needed it in one of the countable set proofs

Although this pull request was inspired by the work on Andrej Bauer and Douglas Bridges - https://elk.zone/sfba.social/@andrejbauer@mathstodon.xyz/110005160302117417 - about whether subsets of countable sets are countable, it doesn't formalize the results there, just some of the more elementary theorems which might lead up to those kinds of results.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9328,6 +9328,13 @@ accepting constructive mathematics,&quot; <I>Bulletin (New Series) of
 the American Mathematical Society</I>, 54:481-498 (2017),
 DOI: <A HREF="http://dx.doi.org/10.1090/bull/1556">10.1090/bull/1556</A> .</LI>
 
+<li><a name="BauerSwan"></a>[BauerSwan] Bauer, Andrej and Swan, Andrew,
+"Every Metric Space is Separable in Function Realizability",
+<i>Logical Methods in Computer Science</i>,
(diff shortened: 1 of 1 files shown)


avekens (approved):
Fine for me, but maybe @tirix should have a look at it (many of his theorems are mentioned).

jkingdon:
> Fine for me, but maybe @tirix should have a look at it (many of his theorems are mentioned).

Seeing no further comments, I suppose we can merge it. If someone has anything to say later, we can always make desired changes then.

PR #3102, opened 2023-03-19
Mathbox

benjub:


wlammen:
You usually add a bj- prefix to theorem names in your mathbox.  This time some are not according to this rule.  Is this intentional?

benjub:
> You usually add a bj- prefix to theorem names in your mathbox. This time some are not according to this rule. Is this intentional?

Thanks, I'm fixing it.

set.mm
```diff
@@ -527093,6 +527118,180 @@ element of the complex projective line (Riemann sphere) to its inverse.
...
+      bj-fununsn2 $p |- ( ph -> ( F ` B ) = C ) $=
+        ( cfv cop csn cun uncom syl6eq bj-funun wcel wceq fvsng syl2anc eqtrd )
+        ABDLBBCMNZLZCABDUDEADEUDOUDEOHEUDPQIRABFSCGSUECTJKBCFGUAUBUC $.
+    $}
+  $}
+
+  $( An element of a difference with a singleton is not equal to the element of
+     that singleton.  Note that ` ( -. A e. { C } -> -. A = C ) ` need not hold
+     if ` A ` is a proper class.  (Contributed by BJ, 18-Mar-2023.) $)
+  eldifsnneq $p |- ( A e. ( B \ { C } ) -> -. A = C ) $=
```
wlammen, line 527218: Still a theorem not following your naming style

benjub: Oh, yes. I plan to move it to Main soon anyway, so it can stay like that.

avekens: This is, in principle, the same as ~eldifsni. I think we do not need both in main.

david-a-wheeler (approved):

david-a-wheeler:
+1. As long as they stay in *mathboxes* and the changes still pass all checks, I'm generally fine with additions.

set.mm
benjub: Oh, I see, the MM-PA didn't catch it when I tried to prove the other theorem, so I added it without looking further.

PR #3103, opened 2023-03-19
shorten 2sb8ev

wlammen:
Not adding a "Proof shortened" tag to avoid littering the file with such tags when people work on the same theorem in succession has a visible effect in this case:  The updated proof won't appear on https://us.metamath.org/mpeuni/mmrecent.html .  Is this a problem?


benjub (approved):

benjub:
Not a problem in this case in my opinion.

PR #3106, opened 2023-03-22
Add Markov's Principle to iset.mm

jkingdon:
As defined here, the class `Markov` is very similar to the class `Omni` (being just an invocation of `pm2.53` away). and just as `_om e. Omni` is the Limited Principle of Omniscience (LPO), `_om e. Markov` is Markov's Principle (MP).

Nothing especially ambitious here in terms of proving theorems, for example from [BauerSwan], concerning countability, Markov's Princple, and excluded middle, but I figured I should send what I have so far.

Also moves two theorems from a mathbox so we can use them in the Markov's Principle section.


iset.mm
```diff
@@ -67733,6 +67743,96 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
...
+$)
+
+  $c Markov $.  $( Class of all Markov sets. $)
+
+  $( Extend class definition to include the class of Markov sets. $)
+  cmarkov $a class Markov $.
+
+  ${
+    $d f x y $.
+    $( An Markov set is one where if a predicate (here represented by a
```
wlammen, line 67759: typo: A Markov set ...

jkingdon: Fixed.

avekens:
I miss a proof for Markov's Principle `_om e. Markov`. Can a corresponding theorem be added?

jkingdon:
> I miss a proof for Markov's Principle `_om e. Markov`. Can a corresponding theorem be added?

It isn't provable from the axioms of IZF.

If you mean a proof from LPO, that is here as an instance of omnimkv . I pushed up a change to say so more explicitly in the comment.

If you mean a proof from excluded middle, that is also something we basically have.  I pushed up a change to say so more explicitly:

```
  $( Excluded middle implies the Lesser Principle of Omniscience (LPO).
     (Contributed by Jim Kingdon, 29-Mar-2023.) $)
  exmidlpo $p |- ( EXMID -> _om e. Omni ) $=
    ( vx wem cv comni wcel wal com exmidomni omex eleq1 spcv sylbi ) BACZDEZA
    FGDEZAHNOAGIMGDJKL $.
```

iset.mm
```diff
@@ -67755,6 +67771,96 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
...
+      GZHBIJZEZAGUNKZLMABNOUQPMABQRZRZCSZURCUONULUMBHUNUAZURRZCSUTABCDUBULUSVBC
+      ULUPVAURHTEULUPVAUCUDHBUNTDUEUFUGUHUIURCUOUJUK $.
+  $}
+
+  ${
+    $d A f x $.
+    $( An omniscient set is Markov.  In particular, the case where ` A ` is
+       ` _om ` means that the Lesser Principle of Omniscience (LPO) implies
+       Markov's Princple (MP).  (Contributed by Jim Kingdon, 18-Mar-2023.) $)
+    omnimkv $p |- ( A e. Omni -> A e. Markov ) $=
```
benjub, line 67829: "principle" (twice)
and I would find `$p |- ( Omni C_ Markov ) $=` cleaner.

jkingdon: > "principle" (twice)

Fixed, thanks.

avekens:
So with ~exmidlpo and ~omnimkv we can immediately prove

```
  $( Excluded middle implies Markov's Principle (MP).
     (Contributed by Jim Kingdon, 29-Mar-2023.) $)
  exmidmp $p |- ( EXMID -> _om e. Markov ) $=
```

I think it's worth to add this theorem.

jkingdon:
> I think it's worth to add this theorem.

Feel free to do this (I guess the mechanics would be to merge this pull request and then make a follow up one).

If you don't get around to it, I can do this but maybe not until next week as I might be too busy this weekend at a crossword puzzle tournament.

jkingdon:
> exmidmp $p |- ( EXMID -> _om e. Markov )

Added.

Ready to merge?

avekens (approved):

PR #3097, opened 2023-03-15
Theorems about pairs

avekens:
Main:
* moved from mathbox: ~mapfvd
* new theorems: ~ss2ralv, ~ss2rexv, ~po2ne, ~infsupprpr

**AV's mathbox: new section "Proper (unordered) pairs"** 

Main theorems: 
* ~prproropen: The set of proper pairs and the set of ordered ordered pairs, i.e., ordered pairs in which the first component is less than the second component, are equinumerous.
* ~rrx2xpreen: The set of points in the two dimensional Euclidean plain and the set of ordered pairs of real numbers (the cartesian product of the real numbers) are equinumerous.
* ~ rrx2plordso: The lexicographical ordering for points in the two dimensional Euclidean plain is a strict complete ordering.
* Furthermore:
* ~rrx2pxel, ~rrx2pyel, ~prelrrx moved from other section

PR #3108, opened 2023-03-23
Shorten zq and leibpilem1; add nn0onn

icecream17:
Two theorems I noticed could be shortened, before starting onto the generalize zsqrtelqelz and subtheorems

As a sidenote, dffltz can be improved by eldifsnneq (very recent) and ss2ralv (still in #3097). But I wonder if a temporary definition like ``F = ( n e. ( ZZ>= ` 3 ) , s e. ~P CC |-> { <. <. x , y >. , z >. | ( ( x ^ n ) + ( y ^ n ) ) = ( z ^ n ) } )`` would be worth it.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16260,6 +16260,7 @@ New usage of "ledi" is discouraged (2 uses).
 New usage of "ledii" is discouraged (2 uses).
 New usage of "lediri" is discouraged (1 uses).
 New usage of "lediv2aALT" is discouraged (0 uses).
+New usage of "leibpilem1OLD" is discouraged (0 uses).
 New usage of "lejdii" is discouraged (1 uses).
 New usage of "lejdiri" is discouraged (1 uses).
 New usage of "lencctswrdOLD" is discouraged (0 uses).
@@ -19395,6 +19396,7 @@ Proof modification of "joincomALT" is discouraged (83 steps).
 Proof modification of "latmassOLD" is discouraged (309 steps).
 Proof modification of "lcmftp" is discouraged (1057 steps).
 Proof modification of "lediv2aALT" is discouraged (264 steps).
+Proof modification of "leibpilem1OLD" is discouraged (254 steps).
 Proof modification of "lencctswrdOLD" is discouraged (34 steps).
 Proof modification of "lenrevcctswrdOLD" is discouraged (77 steps).
 Proof modification of "lspfixedOLD" is discouraged (1103 steps).
--- a/discouraged
+++ b/discouraged
@@ -18194,6 +18194,7 @@ New usage of "zfnuleuOLD" is discouraged (0 uses).
 New usage of "zfpair" is discouraged (1 uses).
 New usage of "zfregs2VD" is discouraged (0 uses).
 New usage of "znnenlemOLD" is discouraged (0 uses).
+New usage of "zqOLD" is discouraged (0 uses).
 New usage of "zrdivrng" is discouraged (1 uses).
 New usage of "zrhcofipsgnOLD" is discouraged (1 uses).
 New usage of "zrhcopsgndifOLD" is discouraged (0 uses).
@@ -20130,5 +20131,6 @@ Proof modification of "zfcndun" is discouraged (72 steps).
 Proof modification of "zfnuleuOLD" is discouraged (59 steps).
 Proof modification of "zfregs2VD" is discouraged (128 steps).
 Proof modification of "znnenlemOLD" is discouraged (201 steps).
+Proof modification of "zqOLD" is discouraged (91 steps).
 Proof modification of "zrhcofipsgnOLD" is discouraged (95 steps).
 Proof modification of "zrhcopsgndifOLD" is discouraged (156 steps).


wlammen (approved):

set.mm
```diff
@@ -359471,10 +359479,20 @@ already know is total (except at ` 0 ` ).  There are branch points at
       $.
   $}
 
+  $( Lemma for ~ leibpi .  (Contributed by Mario Carneiro, 7-Apr-2015.)  (Proof
+     shortened by Steven Nguyen, 23-Mar-2023.) $)
+  leibpilem1 $p |- ( ( N e. NN0 /\ ( -. N = 0 /\ -. 2 || N ) ) ->
+    ( N e. NN /\ ( ( N - 1 ) / 2 ) e. NN0 ) ) $=
```
benjub, line 359485: The proof as well as the usages seem to show that there is not much gain in having these two theorems bundled like that into one. Also, `-. N = 0` is implied. I think two theorems (and not merely lemmas) would be justified:
```
|- ( ( N e. NN0 /\ -. 2 || N ) -> N e. NN ) $=
|- ( ( N e. NN0 /\ -. 2 || N ) -> ( N - 1 ) / 2 ) e. NN0 ) $=
```
and maybe also
```
|- ( ( N e. NN0 /\ -. 2 || N ) -> ( N + 1 ) / 2 ) e. NN ) $=
```

wlammen:
Is there a reason to keep this PR on hold any longer?  The two modified theorems are replacements for the better, both with shortened proofs, and none based on more axioms than the original.  While there might be headroom for further improvements or considerations in the big picture, I don't see why this cannot be postponed until this is merged first.

icecream17:
(I was actually implementing benjub's first two theorems, although the point still stands)

Edit: Actually, given that the second theorem is just nn0oddm1d2 + biimpa, I'm only implementing nn0onn (this commit was made yesterday so it appears above wlammen's comment but it was actually published a few minutes after this comment)

benjub:
Thanks. Now that we have both ~nn0onn and ~nn0oddm1d2, and looking at the two usages of ~leibpilem1, I think that ~leibpilem1 can be removed (i.e., in its two usages, replace it with ~nn0onn or ~nn0oddm1d2 with possibly some glue).

set.mm
```diff
@@ -175578,6 +175586,48 @@ equal to the half of the predecessor of the odd number (which is an even
       EFGZWFWKUDVTWBYGWAWCABUIVQCWEUCVRVS $.
   $}
 
+  $( 0 is even.  (Contributed by AV, 11-Feb-2020.)  (Revised by AV,
```
benjub, line 175589: for consistency with the following comments:
>  $( 2 divides 0.  That means 0 is even.

set.mm
```diff
@@ -175578,6 +175586,48 @@ equal to the half of the predecessor of the odd number (which is an even
...
+    wi 2z mp2an mto ) ABCDZABEDZBAFDUCGHBAIJKLAMNBONUBUCRSPABQTUA $.
+
+  $( 2 does not divide -1.  That means -1 is odd.  (Contributed by AV,
+     15-Aug-2021.) $)
+  n2dvdsm1 $p |- -. 2 || -u 1 $=
+    ( c2 c1 cneg cdvds wbr wn caddc co z0even ax-1cn 1pneg1e0 addcomli breqtrri
+    cc0 neg1cn cz wcel wb neg1z oddp1even ax-mp mpbir ) ABCZDEFZAUCBGHZDEZANUED
+    IBUCNJOKLMUCPQUDUFRSUCTUAUB $.
+
+  $( 2 is even.  (Contributed by AV, 12-Feb-2020.)  (Revised by AV,
```
benjub, line 175607: > $( 2 divides 2. That means 2 is even.

set.mm
```diff
@@ -175578,6 +175586,48 @@ equal to the half of the predecessor of the odd number (which is an even
...
+     23-Jun-2021.) $)
+  z2even $p |- 2 || 2 $=
+    ( c2 cz wcel cdvds wbr 2z iddvds ax-mp ) ABCAADEFAGH $.
+
+  $( 2 does not divide 3, i.e. 3 is an odd number.  (Contributed by AV,
```
benjub, line 175612:
```suggestion
  $( 2 does not divide 3. That means 3 is odd.  (Contributed by AV,
```

set.mm
```diff
@@ -175578,6 +175586,48 @@ equal to the half of the predecessor of the odd number (which is an even
...
+     28-Feb-2021.) $)
+  n2dvds3 $p |- -. 2 || 3 $=
+    ( c2 c3 cdvds wbr wn c1 cmin co cz 2z iddvds ax-mp 3m1e2 breqtrri oddm1even
+    wcel wb 3z mpbir ) ABCDEZABFGHZCDZAAUACAIPAACDJAKLMNBIPTUBQRBOLS $.
+
+  $( 4 is an even number.  (Contributed by AV, 23-Jul-2020.)  (Revised by AV,
```
benjub, line 175618: $( 2 divides 4. That means 4 is even.

set.mm
```diff
@@ -175578,6 +175586,48 @@ equal to the half of the predecessor of the odd number (which is an even
...
+     4-Jul-2021.) $)
+  z4even $p |- 2 || 4 $=
+    ( c2 cmul co c4 cdvds cz wcel wbr 2z dvdsmul1 mp2an 2t2e4 breqtri ) AAABCZD
+    EAFGZOANEHIIAAJKLM $.
+
+  $( An integer which is divisible by 4 is an even integer.  (Contributed by
```
benjub, line 175624: > $( An integer which is divisible by 4 is divisible by 2, that is, is even.

icecream17:
Would there be leibpilem2OLD and leibpiOLD?

This could be more justified if I rewrite leibpi* to not include the redundant `k = 0`.
Additionally, both times leibpilem1 is used, it's really used to prove 
`(𝑘 ∈ NN0 → if((𝑘 = 0 ∨ 2 ∥ 𝑘), 0, ((-1↑((𝑘 − 1) / 2)) / 𝑘)) ∈ ℂ)`, which suggests a completely separate lemma, although some of the intermediate steps are used for other closures... But this pr is already getting much bigger than expected.

avekens (approved):
For me, this PR is fine. The revision of the comments as proposed by @benjub would be OK for me, too.

benjub:
> For me, this PR is fine. The revision of the comments as proposed by @benjub would be OK for me, too.

Fine for me. I made some suggestions but nothing is blocking.

set.mm
icecream17: Alright these suggestions should be applied now

jkingdon:
There seems to be a consensus this is ready to merge. Feel free to make follow up pull requests if there are other things people want to change.

PR #3110, opened 2023-03-26
shorten axc15

wlammen:
... and remove outdated OLD theorems.


benjub (approved):

benjub:
@wlammen : I think you could move wl-mps, wl-syls1, wl-syls2 to Main, as early as possible, and do a `minimize_with wl-mps,wl-syls1,wl-syls2` on the output of, say, `show usage a2i,imim1i` (using `scripts/min.cmd`). Can you do that?

**edit:** wl-embant too, so minimization also on theorems using imim2i. The comment on intuitionistic logic is unnecessary since all these theorems will be in the {ax-1,ax-2}-section.

wlammen:
I can do that, but not today.

PR #3113, opened 2023-04-02
Intersection points of a line and a circle (2)

avekens:
The main result of this PR is ~itscnhlinecirc02p, showing that a nonhorizontal line passing through a point `X` within a circle around the origin (and an arbitray but different point `Y`) intersects the circle at exactly two different points. The special case for a horizontal line should be provable with no problems (but maybe with a lot of additional effort). 

My goal ist still to prove something like
``` ... -> E! s e. ~P P ( ( # ` s ) = 2 /\ A. p e. s p e. ( .0. S R ) /\ p e. ( X L Y ) ) ) ) ```
or with my definition of (proper) pairs:
``` ... -> E! s e. ( Pairs ` P ) A. p e. s p e. ( .0. S R ) /\ p e. ( X L Y ) ) ) ```
but this seems to be not easy because of the nested restricted uniqueness.

Minor changes in main body of set.mm:
* ~2reuswap2, ~reuxfr3d, ~reuxfr4d, ~pm5.31r moved from mathboxes
* ~mul4r, ~addmulsub, ~subaddmulsub, ~mulsubaddmulsub, ~sqn0rp, ~cnsqrt00 added

AV's mathbox:
* ~reuf1odnf, ~reuf1od, ~euoreqb added in section "Restricted uniqueness and "at most one" quantification"
* ~prproropreud added in section "Proper (unordered) pairs"
* Section "Additional theorems for double restricted existential uniqueness" added
* Section "Imaginary and complex number properties - extension" added
* Section "Solutions of quadratic equations" added
* Section "Real euclidean space of dimension 2" added
* ~rrx2pxel, ~rrx2pyel, ~prelrrx2, ~rrx2xpref1o, ~rrx2xpreen, ~rrx2plord, ~rrx2plord1, ~rrx2plord2, ~rrx2plordisom, ~rrx2plordso  moved to section "Real euclidean space of dimension 2"
* ~rrx2linest2, ~elrrx2linest2, ~itsclquadb, ~itsclquadeu, ~2itscp, ~itscnhlinecirc02p added in section "Spheres and lines in real Euclidean spaces"

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,10 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 2-Apr-23 reuxfr4d  [same]      moved from TA's mathbox to main set.mm
+ 2-Apr-23 reuxfr3d  [same]      moved from TA's mathbox to main set.mm
+ 2-Apr-23 2reuswap2 [same]      moved from TA's mathbox to main set.mm
+ 2-Apr-23 pm5.31r   [same]      moved from RM's mathbox to main set.mm
 19-Mar-23 soasym    [same]      moved from SF's mathbox to main set.mm
 19-Mar-23 fvmptd2   [same]      moved from GS's mathbox to main set.mm
 15-Mar-23 mapfvd    [same]      moved from AV's mathbox to main set.mm

### CODE REVIEW

**Comment by avekens** [comment]
Curiously, the command "improve all" used the theorem axicn (which is tagged by "(New usage is discouraged.)") to complete the proof instead of ax-icn. I had to replace it manually...

**Comment by david-a-wheeler** [comment]
> Curiously, the command "improve all" used the theorem axicn (which is tagged by "(New usage is discouraged.)") to complete the proof instead of ax-icn. I had to replace it manually..

As you know, it shouldn't do that. I don't have time to track down why it did that (and fix it), but hopefully someone else can...!

File: set.mm [hunk 1]
```diff
@@ -7314,6 +7314,10 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
   pm5.31 $p |- ( ( ch /\ ( ph -> ps ) ) -> ( ph -> ( ps /\ ch ) ) ) $=
     ( wi wa pm3.21 imim2d imp ) CABDABCEZDCBIACBFGH $.
 
+  $( Variant of ~ pm5.31 .  (Contributed by Rodolfo Medina, 15-Oct-2010.) $)
+  pm5.31r $p |- ( ( ch /\ ( ph -> ps ) ) -> ( ph -> ( ch /\ ps ) ) ) $=
```
jkingdon, line 7318: Shouldn't this be called `pm5.31com`? I expected `pm5.31r` to turn a biconditional into an implication, but that isn't what this is.

set.mm
```diff
@@ -114740,6 +114791,16 @@ this axiom (with the defined operation in place of ` x. ` ) follows as a
     IZUKBDHIHIZUDUEUJUMJZUFUBUCUEUPUBUCUEKUIULDHABCLMNOUDUHEFZUGUJUNJZABPUQUEUF
     URUHCDQRSUBUEUCUFUMUOJZUBUEGUKEFZUCUFGUSACPUTUCUFUSUKBDQRSTUA $.
 
+  $( Rearrangement of 4 factors: swap the right factors in the factors of a
+     product of two products.  (Contributed by AV, 4-Mar-2023.) $)
+  mul4r $p |- ( ( ( A e. CC /\ B e. CC ) /\ ( C e. CC /\ D e. CC ) ) ->
```
jkingdon, line 114796: I suppose the ideal naming convention would somehow have a `23` in it for what is now `mul4` and a `24` for this one. Like `4mul23` and `4mul24` perhaps? Possibly more than you want to tackle right now, though (especially since at least to me it isn't clear what the best choice is).

jkingdon (approved):
I had a few comments, but none are blocking.

set.mm
avekens: It's difficult (I just took this from RM's mathbox without changing the label). "r" usually means "right", maybe because "ps" is the rightmost wff in the antecedent? Furthermore, there is no label following the pattern pm*.*com in set.mm yet. I would like to hear (at least) a third opinion about this.

wlammen:
Examples where com is used: impcomd, ancomd  Examples where l, r suffix is used jccir, jccil, ancld, ancrd. The latter matches the case here best, so i am in favor of an l or r suffix.

set.mm
jkingdon: Maybe just keep the label from the mathbox. I see the analogy to `jccir`, `jccil`, `ancld`, and `ancrd` (although the left variant here isn't named with an "l"), or even more closely to `pm4.71r` (where "r" apparently stands for "reversed").
I don't exactly retract the idea that an analogy to `impcom`  would also make sense, but it doesn't seem as compelling as when I first wrote this.

set.mm
avekens: > I suppose the ideal naming convention would somehow have a `23` in it for what is now `mul4` and a `24` for this one. Like `4mul23` and `4mul24` perhaps? Possibly more than you want to tackle right now, though (especially since at least to me it isn't clear what the best choice is).

Yes, you are right: "r" means "reversed" in this case, and this suffix is also mentioned in the conventions:

> r : theorem with something reversed (e.g., a biconditional)

set.mm
avekens: > ... or even more closely to `pm4.71r` (where "r" apparently stands for "reversed"). ...

Yes, you are right: "r" means "reversed" in this case, and this suffix is also mentioned in the conventions:

    r : theorem with something reversed (e.g., a biconditional)

avekens:
If there are no objections, I will merge this PR tomorrow.

PR #3114, opened 2023-04-04
Add sup3exmid to iset.mm

jkingdon:
This shows that the least upper bound property for real numbers implies excluded middle.

Although it was inspired by https://mathstodon.xyz/@andrejbauer/110123070704027302 it doesn't follow that proof especially closely.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -5421,9 +5421,7 @@ favor of theorems in deduction form.</TD>
   <TD>sup2 , sup3 , sup3ii</TD>
   <TD><I>none</I></TD>
   <TD>We won't be able to have the least upper bound property for all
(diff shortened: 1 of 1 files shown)


avekens (approved):
Formally, everything is OK. I cannot, however, rate the content.

jkingdon:
> Formally, everything is OK. I cannot, however, rate the content.

Sounds good enough for me.  If anyone has any other thoughts, we can always make a follow up pull request.

I'd be glad to answer questions about the math if anyone wants to dig into it further. For example, there is a version of this proof at https://math.andrej.com/2019/09/09/on-complete-ordered-fields/ (look for "Theorem: If there exists a non-trivial complete order then excluded middle holds.").

PR #3115, opened 2023-04-04
Add fodjumkv to iset.mm

jkingdon:
This is a result which might be useful for countable set theorems (although I don't know for sure, just a hunch and noticing that this was easy to prove).

fodjumkv is similar to fodjuomni but only needs Markov's Principle, and states that a certain set is inhabited if it is nonempty rather than saying that set is empty or inhabited.

Rename fodjuomnilemf to fodjuf and change hypothesis from O e. Omni to O e. V .

Rename fodjuomnilemm to fodjum and remove unused hypothesis O e. Omni .

Rename fodjuomnilem0 to fodju0 and remove unused hypothesis O e. Omni .

Includes lemma fodjumkvlemres which is similar to fodjuomnilemres but has different logic for the difference between Omni and Markov .


avekens (approved):
Formally, everything is OK. I cannot, however, rate the content.

jkingdon:
> Formally, everything is OK. I cannot, however, rate the content.

I'll plan on merging it relatively soon then.

I'm not completely sure I can rate the content either (I don't know whether this theorem appears in the literature for example), but the worst case scenario is that we don't ever end up using this theorem, so I'm not too worried about having it around.

PR #3116, opened 2023-04-06
rtprmirr [mathbox]

icecream17:
Many generalized versions of theorems; notably, zrtelqelz generalizes zsqrtelqelz

resubidaddid1 saves steps and symbols on the html page but not any compressed bytes


avekens (approved):
Great results! I wonder why they were not contained/needed in set.mm until now.

PR #3117, opened 2023-04-07
Intuitionize section "Limits and continuity in topological spaces" up to cnprest2

jkingdon:
This is pretty much the usual intuitionizing. Some theorems regarding closed sets didn't make it (but closed sets in general are for another day).

Most of the rest are here, either as-is or with explicit conditions that certain sets are topologies (that is, I didn't see any way to prove https://us.metamath.org/mpeuni/cnptop1.html and similar theorems, so I added a `𝐽 ∈ Top` condition to the theorems which would have used it).

Also includes:
* Add `biadani` and `biadanii` which are (basically) as in set.mm.
* Don't require the two variables in `r19.2m` to be the same. Although this is a form of bundling which I'm not sure we love in general, in this case it seems like a significant convenience which avoids changes of bound variables where this theorem is used.
* Add `mptrel` and `mptrcl` (stated as in set.mm with different but easy proofs).
* Add `ffund` (copied without change from set.mm)
* Add `elfvmptrab1` (stated as in set.mm)
* Add `elfvmptrab` (copied without change from set.mm)
* Add a few theorems to the missing theorems list in mmil.html: `nsyl4`, `biadan`, `relmptopab`, `ovmptss`, `relmpt2opab`, 
* Minimize `cnpimaex` and `cnpnei` in set.mm

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -292,6 +292,17 @@
 "peano1nnnn" is used by "nnindnn".
 "peano2nnnn" is used by "nnindnn".
 "peano5nnnn" is used by "nnindnn".
+"r19.2mOLD" is used by "climuni".
+"r19.2mOLD" is used by "cnviinm".
+"r19.2mOLD" is used by "eusvobj2".
+"r19.2mOLD" is used by "iinerm".
+"r19.2mOLD" is used by "iinexgm".
+"r19.2mOLD" is used by "intssunim".
+"r19.2mOLD" is used by "r19.2uz".
+"r19.2mOLD" is used by "rexfiuz".
+"r19.2mOLD" is used by "riinm".
+"r19.2mOLD" is used by "trintssmOLD".
+"r19.2mOLD" is used by "xpiindim".
 "rdgivallem" is used by "rdgival".
 "reapti" is used by "apreap".
 "reapti" is used by "apti".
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -29490,13 +29507,28 @@ intersection or the difference (that is, this theorem would be equality
...
     $( Theorem 19.2 of [Margaris] p. 89 with restricted quantifiers (compare
        ~ 19.2 ).  The restricted version is valid only when the domain of
-       quantification is inhabited.  (Contributed by Jim Kingdon,
-       5-Aug-2018.) $)
-    r19.2m $p |- ( ( E. x x e. A /\ A. x e. A ph ) -> E. x e. A ph ) $=
+       quantification is inhabited.  (Contributed by Jim Kingdon, 5-Aug-2018.)
+       New proofs should use ~ r19.2m which is the same but which allows ` x `
+       and ` y ` to be either the same variable or different ones.
+       (Proof modification is discouraged.)  (New usage is discouraged.) $)
+    r19.2mOLD $p |- ( ( E. x x e. A /\ A. x e. A ph ) -> E. x e. A ph ) $=
```
avekens, line 29514: This may be called r19.2mALT - OLD theorems will/should be removed after one year...

jkingdon: Well it is an OLD in the sense that the plan is to remove it (yes, in favor of the new r19.2m). I don't think ALT would make sense because isn't that for things which have some reason to continue to exist?

As far as I can tell the best plan is to stick with the OLD nomenclature but follow more of the conventions at https://us.metamath.org/mpeuni/conventions.html . I've pushed up a change to do that. I suppose if we are really committed to exactly a year (rather than at least a year) we can always make sure to switch over the usages before then (it is only 11).

avekens: I do not now the procedure for iset.mm, but for set.mm, several people check if there are obsolete theorems older than 1 year and delete them. This causes trouble if they are still in use. therefore, their usage should be eliminated as soon as possible.

avekens (approved):
OK so far, the usage of r19.2mOLD should be eliminated as soon as possible.

iset.mm
jkingdon: The most obvious difference in iset.mm is that we've been deleting theorems immediately if we don't want them and there are no usages. I suppose if there are more people working on it that might change some day, but at least speaking for myself my preference is to continue with the immediate deletions unless people report this is causing problems.

PR #3119, opened 2023-04-10
Use df-seq3 syntax more in iset.mm (seqfeq3 and others)

jkingdon:
By now this is fairly routine stuff. In some cases the needed theorems already have a version which uses df-seq3 syntax, in others we need to add it. It is a big job but I'm tackling it one piece at a time.

There are also a few hints about what will be involved in changing the syntax in https://us.metamath.org/ileuni/df-isum.html . We can prove an alternate definition (see `dfsumdc` in the pull request) but this isn't quite in the form of a definition because it has a `A. k e. A B e. CC` antecedent. At some point we'll need to figure out how to swap this definition out, but that will be easier once more theorems are converted over, and doesn't really affect how we convert them over.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -7994,16 +7994,16 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <TR>
   <TD>df-sum</TD>
-  <TD>~ df-isum</TD>
-  <TD>Although this defintion is intended to function similarly to
-  the set.mm one, a lot of details have to be changed, especially
-  around decidability, to make sum work.</TD>
+  <TD>~ df-isum , ~ dfsumdc</TD>
+  <TD>The iset.mm definition/theorem adds a decidability condition
+  and an ` if ` expression (which is to deal with differences in
+  using ` seq ` for finite sums).  It does function similarity to
+  the set.mm definiton of ` sum_ ` .</TD>
 </TR>
 
 <TR>
   <TD>sumex</TD>
-  <TD><I>none</I></TD>
-  <TD>This will need to be replaced by suitable closure theorems.</TD>
+  <TD>~ fsumcl , ~ isumcl</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6824,7 +6824,7 @@ or ~ ssfiexmid </TD>
 
 <TR>
   <TD>seqid</TD>
-  <TD>~ iseqid</TD>
+  <TD>~ seq3id</TD>
 </TR>
 
 <TR>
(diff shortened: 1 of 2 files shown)


mmil.raw.html
```diff
@@ -8032,7 +8027,7 @@
 
 <TR>
   <TD>sumrblem</TD>
-  <TD>~ isumrblem</TD>
+  <TD>~ sumrbdclem</TD>
```
wlammen, line 8030: Why is this theorem mentioned here, but not isummolem2, for example?

wlammen:
I did my best to do a review here, so that this PR is not blocked any longer.  For most of it, the theorems are understandable to me, as they cover elementary analysis.  But I am not familiar with some expressions like DECID, so I cannot go beyond a formal check here.

I found a suspicious omission in mmil.raw.html, but nothing more.  So if it is explained, I will approve this PR.

mmil.raw.html
jkingdon: This hasn't been especially systematic. There are places where lemmas match quite closely between set.mm and iset.mm and I guess sometimes they have been listed. In general, though, we don't list them.

If the lemma seems like it could be used outside the proof of a single main theorem, that would be a reason to list it, but that doesn't seem to be true of sumrblem so I've pushed up a change to remove it from this list.

jkingdon:
> I did my best to do a review here, so that this PR is not blocked any longer.

Thank you!

> For most of it, the theorems are understandable to me, as they cover elementary analysis. But I am not familiar with some expressions like DECID, so I cannot go beyond a formal check here.

At a surface level this is quite simple - `DECID ph` (often stated in words as "ph is decidable") is just short for `ph \/ -. ph` (so in set.mm it would hold for any proposition).  As for what propositions are decidable in iset.mm, some widely used ones are equality of integers or rational numbers and a few other integer-related propositions like https://us.metamath.org/ileuni/zdclt.html . But most notably not equality of real numbers.

The way `DECID` shows up in sums is that the definition of summation in iset.mm (and set.mm) allows us to sum a series which is indexed not just by natural numbers or a finite set, but certain subsets of integers. In practice these will be well behaved subsets such as the natural numbers themselves, but the `DECID` condition is part of how we formalize the concept of "well behaved".

Sorry, that's probably more detail than you needed, but I'd be glad to explain this stuff as desired. Part of my hope in working on iset.mm is that mathematics without excluded middle can become more accessible!

> I found a suspicious omission in mmil.raw.html, but nothing more. So if it is explained, I will approve this PR.

I am perhaps biased in the case of my own pull requests but that makes sense to me. We should give reviewers a chance to weigh in but given some level of plausibility I think we can, generally speaking, be willing to treat silence as consent. If people find issues later, we can always revise then.

wlammen (approved):

wlammen:
`Sorry, that's probably more detail than you needed, but I'd be glad to explain this stuff as desired.`

Nice to learn from someone who loves the ideas he publishes.

Issue #3089, opened 2023-03-09
Define the indiscrete topology in iset.mm


jkingdon:
The definition in set.mm - in https://us.metamath.org/mpeuni/indistopon.html and related theorems - is {∅, 𝐴} and that is not going to work intuitionistically.

We plan on defining it to be the collection of all sets such that the membership relation does not depend on `x`, i.e. `{ s | A. x e. s s = X }`

The key theorems are the analogues to https://us.metamath.org/mpeuni/indistopon.html , https://us.metamath.org/mpeuni/indistop.html , https://us.metamath.org/mpeuni/indisuni.html , https://us.metamath.org/mpeuni/indistps.html , https://us.metamath.org/mpeuni/indistps2.html . There are a few more which depend on those, but that's the start.

digama0:
I think you can also do similar things for the other topologies mentioned in #3088:

* finite complement topology: `{ s | A. x e. s E. t e. Fin A = ( s u. t ) }`
* countable complement topology: `{ s | A. x e. s E. t ( t ~<_ om /\ A = ( s u. t ) ) }`
* particular point topology: `{ s e. ~P A | A. x e. s P e. s }`

david-a-wheeler:
I'll ask my usual question: Should we modify the definition in set.mm so it would work in both set.mm and iset.mm? I suspect the answer may be "no" in this case, but I feel I should ask.

benjub:
> ...the collection of all sets such that the membership relation does not depend on `x`, i.e. `{ s | A. x e. s s = X }`

This sentence is not very clear and does not match the statement. I think a more transparent expression for the indiscrete topology on `A` (derived from an explicited version of the above sentence) is
```
{ x e. ~P A | F/ y e. A y e. x } 
```
where the restricted nonfreeness quantifier has the natural definition `( F/ x e. A ph <-> ( E. x e. A ph -> A. x e. A ph ) )`. This expression is indeed equivalent to the one given by @digama0, and this should be a theorem, I think.

digama0:
I think it's a bit much to add a restricted not-free quantifier just for that. FTR I got that definition by starting from `A. x A. y ( x e. s -> y e. s )` and then progressively simplifying it, since we generally prefer short definitions and unpack them in subsequent theorems.

benjub:
[...]

PR #3121, opened 2023-04-11
In iset.mm, finish section "Limits and continuity in topological spaces"

jkingdon:
Also includes parts of section "Continuous function-builders". Topics which are (a little or a lot) more difficult we just note briefly or skip entirely, for example compactness and of course the indiscrete topology already noted at #3089 .

Apparently metric spaces will be next after this, at least I didn't see anything else which seemed likely to be essential.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9524,6 +9524,12 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>The set.mm proof relies on restcldr</td>
 </tr>
 
+<tr>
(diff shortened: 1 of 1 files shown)


wlammen (approved):
This is (more or less) a copy of theorems in the field of general topology and sequences from set.mm.  So I don't see why it cannot be accepted.

avekens (approved):

jkingdon:
The most recent push is just a rebase.

PR #3123, opened 2023-04-16
xrhaus moved from TA's mathbox to main

glacode:
A couple of theorems in my next PR use xrhaus

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+16-Apr-23 xrhaus    [same]      moved from TA's mathbox to main set.mm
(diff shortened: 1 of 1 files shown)


avekens:
Is it possible to shorten the proof of ~xrge0tsms using ~ xrhaus? This would additionally justify having  ~ xrhaus in main.

jkingdon (approved):

jkingdon:
> Is it possible to shorten the proof of ~xrge0tsms using ~ xrhaus? This would additionally justify having ~ xrhaus in main.

That would make it a bit clearer that it belongs in main, yes.

At least for me, I also don't object to moving this theorem (that is, merging this pull request) based on a statement that it will be used in the future, although I suppose if we take that too far we might end up regretting it.

glacode:
A PR with just "move to main", allows for the complex PR to be mathbox only

PR #3124, opened 2023-04-18
expand extended real addition and negation in iset.mm

jkingdon:
All the addition and negation theorems in this section can be proved.  Most of them require changes to the proof, often to use https://us.metamath.org/ileuni/elxr.html in place of excluded middle (or other changes somewhat along those lines).

Extended real multiplication is another story, that seems a lot more difficult. But that can be for later (for example if metric spaces turn out to need it).

Includes the usual updates to iset.mm to bring in some theorems from set.mm.

Includes some updates to the missing theorems list to describe theorems which had only been mentioned in git logs until now.

In set.mm, minimizes two theorems and fixes a comment.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -5817,7 +5817,7 @@ with minor changes.</TD>
 
 <TR>
 <TD>
-xaddmnf2 , pnfaddmnf , mnfaddpnf , rexadd , rexsub , xaddnemnf ,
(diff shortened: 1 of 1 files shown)


wlammen (approved):
I find the splitting into lots of commits very helpful for review.  I actually walk through the list commit by commit.

PR #3125, opened 2023-04-19
Shorten oaabs2

GinoGiotto:
I found an 8 bytes shorter proof for `oaabs2` expressed in the compressed format. This is the first time I put my hands on the theorems themselves so I'm not sure I addressed all conventions correctly, let me know in the comments.


GinoGiotto:
Hmm I'm trying to understand what went wrong, but I'm not sure I understand what I'm supposed to do

GinoGiotto:
Ok, I fixed one check

GinoGiotto:
Nice, I fixed all checks

wlammen:
a sylanbrc is replaced with a sylanblrc, saving a proof line.  The list of axioms has not changed, so this is a valid shortening.

wlammen (approved):

PR #3126, opened 2023-04-20
Intuitionize section "Pseudometric spaces"

jkingdon:
All of the theorems in this section can be proved as stated in set.mm, although many of the proofs need some intuitionizing.

Includes proofs of `dmxpid`, `xpid11` and `xpider` as stated in set.mm (this is exciting because it reduces the need to intuitionize proofs which use these). Removes `dmxpinm`, `xpid11m` and `xpiderm` as these are unused and are just versions of those theorems with additional conditions (which turn out to be unneeded).

Includes `xleaddadd` which is a way to intuitionize one of the pseudometric theorems without having to introduce extended real multiplication (which is a pretty big can of worms).

Includes copying over a few other theorems from set.mm to iset.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2671,7 +2671,7 @@ and is evaluated at a set</TD>
 
 <TR>
 <TD>elfvex</TD>
-<TD>~ relelfvdm </TD>
+<TD>~ relelfvdm , ~ mptrcl</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2371,11 +2371,6 @@ there is less need for this convenience theorem.</TD>
 </TR>
 
 <TR>
-<TD>dmxpid</TD>
-<TD>~ dmxpm </TD>
-</TR>
-
-<TR>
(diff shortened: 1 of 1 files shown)


wlammen (approved):
The only thing that puzzled me was psmettri and psmettri2.  One seems a trivial modification of the other, so I wonder whether one of these two would be sufficient.  But since this is a matter of taste, and convenience theorems are found all over the place, this is non-blocking.

jkingdon:
> The only thing that puzzled me was psmettri and psmettri2. One seems a trivial modification of the other, so I wonder whether one of these two would be sufficient.

Interesting observation. I usually don't pay close attention to such matters (that is, I'm more focused on intuitionizing than revisiting set.mm's choices).

As for the merits of this one in particular, even in a small amount of working on metric spaces, I seemed to end up applying theorems like https://us.metamath.org/mpeuni/psmetsym.html and https://us.metamath.org/mpeuni/addcom.html a lot. So maybe the convenience theorems are justified? Not that I have super strong opinions if someone else wants to take a closer look.

PR #3127, opened 2023-04-21
Shortened proofs

GinoGiotto:
I made a script that runs the proof minimizer through the `set.mm` database. Specifically it runs `MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*` since this is the form Metamath suggests to avoid using new axioms.

This is the result of the analysis of the first 1000 theorems that don't have a `proof modification is discouraged` label in their comment.

I also reverted the OLD theorem I added in my [previous pull request](https://github.com/metamath/set.mm/pull/3125) since I read [here](https://us.metamath.org/mpeuni/conventions.html) that automatic minimizations don't need to keep the OLD version of the theorem nor they need to add credit for the shortening.

I plan to make more of these if they are welcome, divided into smaller pull requests to make them easier for review.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19830,7 +19830,7 @@ Proof modification of "tbwsyl" is discouraged (20 steps).
 Proof modification of "tfr1ALT" is discouraged (19 steps).
 Proof modification of "tfr2ALT" is discouraged (52 steps).
 Proof modification of "tfr3ALT" is discouraged (84 steps).
-Proof modification of "oaabs2OLD" is discouraged (1117 steps).
+Proof modification of "oaabs2OLD" is discouraged (1366 steps).
 Proof modification of "topnfbey" is discouraged (75 steps).
 Proof modification of "toycom" is discouraged (142 steps).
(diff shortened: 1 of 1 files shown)


icecream17 (approved):

jkingdon (approved):

jkingdon:
> `MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*`

Sounds right. That's what https://github.com/metamath/set.mm/blob/develop/scripts/minimize does and probably also what we have in other scripts and documentation.

> I plan to make more of these if they are welcome, divided into smaller pull requests to make them easier for review.

Sounds good.  The only reason we haven't (recently) been doing this on all of set.mm in a systematic way is that we found that it took a lot of compute time to do it. But we are in favor of running the minimizer as feasible (and as far as I know we have fixed issues around accidentally introducing axioms and other undesirable changes).

PR #3130, opened 2023-04-23
shorten 2eu1 2eu3 axbnd revise axi12

wlammen:
2eu1: This is a tiny improvement, saving a proof byte, and a few symbols in the proof display by simplifying a complex expression a bit earlier.
2eu3: This simple improvement at least saves a proof line.
axi12 can be proven without ax-11.
axbnd is shortened by proving it from axi12.
Also delete an outdated OLD proof.  Another one cannot be deleted before May 1, because it is still referenced.


jkingdon:
The tests are not running due to

```
Invalid workflow file

The workflow is not valid. .github/workflows/verifiers.yml: (Line: 147, Col: 14, Idx: 7453) - (Line: 147, Col: 15, Idx: 7454): While parsing a tag, did not find expected tag URI.
```

See #3134 for more details.

jkingdon (approved):

PR #3131, opened 2023-04-23
A sequence of extended reals converges if and only if its inferior limit and its superior limit are equal.

glacode:
The main result is ~ xlimliminflimsup .

The theory of liminf has been further developed.


glacode:
this is mathbox only

wlammen (approved):

PR #3132, opened 2023-04-23
Shortened proofs 2

GinoGiotto:
Continuation of https://github.com/metamath/set.mm/pull/3127. Proof minimizer ran between the 1000th and the 2000th theorems that don't have a `proof modification is discouraged` label in their comment.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19830,7 +19830,7 @@ Proof modification of "tbwsyl" is discouraged (20 steps).
 Proof modification of "tfr1ALT" is discouraged (19 steps).
 Proof modification of "tfr2ALT" is discouraged (52 steps).
 Proof modification of "tfr3ALT" is discouraged (84 steps).
-Proof modification of "oaabs2OLD" is discouraged (1117 steps).
+Proof modification of "oaabs2OLD" is discouraged (1366 steps).
 Proof modification of "topnfbey" is discouraged (75 steps).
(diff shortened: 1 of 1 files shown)


GinoGiotto:
I just noticed that the theorem `exintr` and `alexan` are exactly the same, maybe one of them could be removed?

wlammen:
I suggest to keep exintr, as this was introduced earlier, but with the shorter proof of alexan.

GinoGiotto:
Ok, so no further edits from this pull request

wlammen:
Except correcting the alexan/exintr duplicate I see no other issue with this pull request.

wlammen (approved):

set.mm
```diff
@@ -15665,7 +15660,7 @@ only postulated (that is, axiomatic) rule of inference of predicate
   $( Introduce a conjunct in the scope of an existential quantifier.
      (Contributed by NM, 11-Aug-1993.) $)
```
benjub, line 15661:
```suggestion
     (Contributed by NM, 11-Aug-1993.)  (Proof shortened by BJ,
     16-Sep-2022.) $)
```
Try not to remove credits, please.

GinoGiotto: I forgot, thanks

icecream17 (approved):

PR #3133, opened 2023-04-24
Removed unused OLD theorems

GinoGiotto:
In [conventions](https://us.metamath.org/mpeuni/conventions.html) I read that OLD statements should be kept for one year, I checked the set.mm database and I found some unused ones that are a few years old, so I removed them. There are also some OLD statements that don't have a date in their comment, I didn't touch those, but since the meaning of OLD is "something to be removed", I wonder how we could determine the right time for their deletion.


wlammen (commented):
Except for dfiota4, all other OLD theorems are within mathboxes (roughly from line 420000 on). It is primarily the responsibility of their owners to keep their backyard tidy, I think.  In the end, a mathbox owner can update a proof without further notice, and need not create an OLD version. So why is it there? By deleting we run a risk of messing with some unknown ideas.  It is still an open question on how to deal with apparently abandoned mathboxes (not touched for a decade or so).  To my knowledge no general strategy is applied there.

As for OLD theorems missing a proper 'Obsolete as of ...' description, a `git checkout 90828b4a3d00987c471b4cf0a16461cbe1248d62` will take you back a year in history, and a comparison of the then-state with the current version of set.mm should reveal whether suspicious candidates are in both versions, and, therefore, are outdated.

GinoGiotto:
I reverted the removal of OLD statements in mathboxes and added the removal of nn0pnfge0OLD since it's also more than one year old and it does not belong in any mathbox.

As for the OLD theorems missing a proper 'Obsolete as of ...' description I found that these had their date removed afterwards, but they are not unused, so they would require proof revision before removal:

`vciOLD` Obsolete as of 21-Sep-2021 6 uses

`vcidOLD` Obsolete as of 21-Sep-2021 4 uses

`vc2OLD` Obsolete as of 21-Sep-2021 2 uses

`isvcOLD` Obsolete as of 4-Oct-2021 1 use

`isvciOLD` Obsolete as of 4-Oct-2021 3 uses

`cnaddabloOLD` Obsolete as of 23-Jan-2020 5 uses

`cnidOLD` Obsolete as of 23-Jan-2020 1 use

`cncvcOLD` Obsolete as of 20-Sep-2021 1 use

All of them belong in the main section of set.mm

GinoGiotto:
Hmm the verifiers did not start, it gives the error `The workflow is not valid. .github/workflows/verifiers.yml: (Line: 147, Col: 14, Idx: 7453) - (Line: 147, Col: 15, Idx: 7454): While parsing a tag, did not find expected tag URI.`

Maybe it's better if I close this pull request and open another one?

Edit: Nevermind I just noticed https://github.com/metamath/set.mm/pull/3134

wlammen:
Thanks for doing this janitorial work.

wlammen:
The end of main is marked by
`$( End $[ set-main.mm $] $)`
and this is found approx. in line 425445 (my version of set.mm, slightly modified due to open pull requests).  vciOLD and friends belong to section set-deprec afterwards containing COMPLEX TOPOLOGICAL VECTOR SPACES (DEPRECATED).
So don't sweat it.  I guess this will be removed in one sweep sometime in the future, once a replacement using extensible structures is fully available, see the comment to the section.

GinoGiotto:
I missed that, thanks. In that case I don't think there are any expired OLD theorems left in main, since the ones that I found were either in a deprecated section or in some mathboxes ( `numclwwlkovgOLD` which is obsolete as of 20-Apr-2022 is used by a theorem that expires in May 1, so I agree with you to wait until that time). As for undated theorems in main there is only `19.43OLD`, but it has a clear "do not delete" comment over it.

However I found some unused expired theorems in nf.mm, are those to be removed?

wlammen:
I'm not in charge of nf.mm, which develops mathematics from a different set of axioms called Quines new foundations.  Before doing some maintenance there make sure you have contacted a developer in this field.  In general, the question is raised again when to assume a project/mathbox is abandoned, and under what extra conditions it is finally eligible for removal from the repositories.  To make this clear: my copy of nf.mm is a version from 22-Feb-2021, which I consider to be too recent to throw it away.  Still, the question is open, and maybe I will start an issue discussing this topic there.

GinoGiotto:
I checked. Most of them date back to 2017 and the remaining ones are from 2016 and 2018. There are also a few without a date description, but I with a quick commit check it seems most of them were introduced in 2018. In any case opening an issue is probably the best course.

As for this pull request I think I finished, since I can't really find anything else to add.

GinoGiotto:
I replaced `cspliceold` with `cspliceOLD` and labeled as discouraged in accordance with conventions to avoid it being missed when checking OLD statements for expiration date.

The statement was introduced in commit https://github.com/metamath/set.mm/commit/34a2d24cab43a060d88d0a460f06f2316e2be133. I think this naming wasn't intentional since I don't read any mention of it deserving any special treatment and I don't find any reason why the user wouldn't just use `csplice` instead.

jkingdon:
> I replaced `cspliceold` with `cspliceOLD` and labeled as discouraged in accordance with conventions to avoid it being missed when checking OLD statements for expiration date.
> 
> The statement was introduced in commit [34a2d24](https://github.com/metamath/set.mm/commit/34a2d24cab43a060d88d0a460f06f2316e2be133). I think this naming wasn't intentional since I don't read any mention of it deserving any special treatment and I don't find any reason why the user wouldn't just use `csplice` instead.

Perhaps it was intentional because it is syntax (`$a class`) rather than a `$a |-` axiom? But that's not a reason to keep the  `cspliceold` name if using the OLD convention works. And OLD does seem to work based on this pull request passing all the checks and not requiring anything odd.

wlammen (approved):
I think the renaming of cspliceold is in accordance with its users like gsumsplOLD.

GinoGiotto:
> Perhaps it was intentional because it is syntax (`$a class`) rather than a `$a |-` axiom?

In the set.mm database there are only 5 occurrences of OLD statements with `$a class`, they all follow the standard conventions except for the one I changed, so I think it's just a mistake.

PR #3134, opened 2023-04-24
Revert "Check set.mm LaTeX in CI pipeline"

jkingdon:
Reverts metamath/set.mm#3100

This is causing test failures with:

```
Invalid workflow file: .github/workflows/verifiers.yml#L1
The workflow is not valid. .github/workflows/verifiers.yml: (Line: 147, Col: 14, Idx: 7453) - (Line: 147, Col: 15, Idx: 7454): While parsing a tag, did not find expected tag URI.
```

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -75,17 +75,6 @@ jobs:
     if: ${{ needs.skip_dups.outputs.should_skip != 'true' }}
     steps:
       - uses: actions/checkout@v3
-      # Install LaTeX so we can test generated LaTeX. We use extended packages:
-      # texlive-extra-utils provides pdflatex
-      # texlive-fonts-extra provides phonetic.sty
-      # texlive-science provides accents.sty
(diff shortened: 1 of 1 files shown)


jkingdon:
Can I get an approval on this pull request? This issue is preventing any other pull requests from being merged.

I don't think we need to do any detailed analysis prior to merging this - we can do that after we get the tests passing again.

david-a-wheeler:
+1 !

david-a-wheeler:
Wait just a second, I may be able to make this a quick fix.

david-a-wheeler:
Here's an attempt to fix it instead: https://github.com/metamath/set.mm/pull/3138

david-a-wheeler:
Let's instead merge this: https://github.com/metamath/set.mm/pull/3138

jkingdon:
> Here's an attempt to fix it instead: #3138

Merged! And thanks for approving the revert as well (since "revert first, then think about the fix" is generally a good mindset even if sometimes you get away with bypassing it).

PR #3141, opened 2023-04-26
shorten nulmo

wlammen:
nulmo: becomes one proof line shorter.

This shows automatic minimization does not catch everything.

This time the CI command
`sudo apt-get install texlive-latex-extra texlive-extra-utils texlive-fonts-extra texlive-science`
took almost 25 min to complete.  This is annoying, because a quick check helps detecting simple errors like rewrap forgotten.  This command at least downloads 0.5 GB of data ( the tex fonts). So (wild guess from me!) maybe a repeated download from the same IP address makes the server slow down the request.  As a sort of penalty, so to say.


icecream17 (approved):
(approve)

btw, pi! (#3141)

PR #3142, opened 2023-04-27
Shortened proofs 5

GinoGiotto:
Continuation of https://github.com/metamath/set.mm/pull/3127. Proof minimizer ran between the 4000th and the 5000th theorems that don't have a `proof modification is discouraged label` in their comment.


GinoGiotto:
Looks like `mpteq12d` is identical to `mpteq12df`, so probably we can remove one of them.

set.mm
```diff
@@ -44315,9 +44314,7 @@ proper class (see for example ~ iprc ).  (Contributed by NM,
     $( An equality theorem for the maps-to notation.  (Contributed by Thierry
        Arnoux, 30-May-2020.) $)
     mpteq12df $p |- ( ph -> ( x e. A |-> B ) = ( x e. C |-> D ) ) $=
```
wlammen, line 44316: I am inclined to keep the name mpteq12df, but, of course with the proof of mpteq12d.  The f suffix indicates the use of a not-free hypothesis, and is in this case closer to mpteq12f than to mpteq12.
This means the proof and credits are taken from mpteq12d, the name changes to mpteq12df, and mpteq12d is deleted.

wlammen (changes requested):
It becomes increasingly difficult to monitor whether changes affect the used axioms list.  Is there a guarantee the minimize command never introduces a new axiom?  This PR seems to be clean, though.
One such example is cnvcnv from #3143 where an inss2 mpbir pair is substituted with a relinxp.  Unfortunately, the latter introduces ax-13.  This does not hurt in this particular case, because ax-13 is referenced elsewhere, but a sort of uneasiness creeps in.

GinoGiotto:
>It becomes increasingly difficult to monitor whether changes affect the used axioms list. Is there a guarantee the minimize command never introduces a new axiom? 

From me I can only garantuee that I use and will only use the command `MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*` which is the form Metamath recommends to avoid this event.

>One such example is cnvcnv from #3143 where an inss2 mpbir pair is substituted with a relinxp. Unfortunately, the latter introduces ax-13. This does not hurt in this particular case, because ax-13 is referenced elsewhere, but a sort of uneasiness creeps in.

This should be nominal. Writing `help minimize` in metamath.exe shows a link to  https://groups.google.com/g/metamath/c/f-L91-1jI24/m/3KJnGa8qCgAJ where Norman Megill explains the details of the command:
> /no_new_axioms_from a1i" (when accompanied by "/allow_new_axioms *" in the new program) means that statement a1i, and all statements depending on a1i such as syl, will be avoided whenever the original proof is not dependent on a1i.  But if the original proof does depend on a1i, then a1i and any of its dependencies may be used to reduce the proof size.  (Normally, the argument "a1i" would be an axiom or axioms rather than a theorem.)

Writing `show trace_back cnvcnv /axioms` in metamath.exe indeed shows that cnvcnv already referenced ax-13, so this is an expected behaviour from one of its designers.

In general I can't provide more guarantees if it happens that the minimize command has a bug
(I'm not one of the designers of metamath, maybe someone who worked on it might be more confident). The good news is that this is a good method to check if minimize does behave as expected.

Usually I would recommend to check anyway, even if this would mean to wait a while before merging.

set.mm
GinoGiotto: Actually I noticed the two proofs are identical too, should I switch credits anyway?

wlammen: Because the proofs are identical, the older credits prevail.

GinoGiotto: Done

GinoGiotto:
The lines I edited have 80 characters which is over the limit, I thought rewrapping would solve this issue but looks like it doesn't.

GinoGiotto:
Ok, I just discovered that `save proof <label-match> /compressed` takes care of that.

wlammen (approved):

avekens:
I think it would be helpful to reference the initial pull request #3127 of this series...

GinoGiotto:
> I think it would be helpful to reference the initial pull request #3127 of this series...

Done

set.mm
```diff
@@ -46129,13 +46114,7 @@ This theorem is proved directly from set theory axioms (no set theory
        16-Jan-2012.)  (Revised by NM, 16-Jun-2017.) $)
     reuxfr2d $p |- ( ph
```
avekens, line 46115: This theorem seems to be obsolete now (it's identical with ~reuxfr3d). However, this version is older than ~reuxfr3d, therefore ~reuxfr2d  should be kept, and ~reuxfr3d should be removed. Since the proof of ~reuxfr3d is shorter, this proof should be kept.
In summary:
* Replace the proof of ~reuxfr2d by the proof of ~reuxfr3d and add "(Proof shortened by Thierry Arnoux, 8-Oct-2017.)" to the comment of ~reuxfr2d 
* Replace all usages of ~reuxfr3d  by ~reuxfr2d
* Remove ~reuxfr3d

avekens (changes requested):
Please clean up ~reuxfr3d/~euxfr2d (see inline comment)

set.mm
GinoGiotto: reuxfr3d is referenced by a theorem that appears earlier than reuxfr2d, is it ok if I move reuxfr2d in the same spot where reuxfr3d was placed?

wlammen: yes

GinoGiotto: `reuxfr3d` is not exactly identical to `reuxfr2d` and the proofs are not interchangeable, it's possible to find a shorter proof automatically for `reuxfr2d` (even shorter than the one of `reuxfr3d` by one byte), but `reuxfr3d` is used by `reuxfr4d` and a simple substitution (or authomation tricks) don't work for it. This actually needs manual proof revision, and I think this would deserve its own pull request...

icecream17: it seems like reuxfr3d is a slightly generalized version of reuxfr2d, with y e. C instead of y e. B (C can be substituted with B)
i don't really understand the naming system... so far it seems like "2" just means "2nd reuxfr theorem"
so instead maybe rename reuxfr3d to reuxfr2d, deleting reuxfr2d (subtly different)

wlammen:
The main objective of this series #3127 (see there) seems to be applying a minimizing script, mostly in a mechanical way.  I suggest to not overload it with optimization on the side, except perhaps for very obvious ones like removing duplicates.  When spotting a deficiency, I recommend simply leaving a non-blocking note in the comments, so this can be addressed in the wake, if necessary by others. This also speeds up merges as we need not check for possible conflicts.

avekens:
> The main objective of this series #3127 (see there) seems to be applying a minimizing script, mostly in a mechanical way. I suggest to not overload it with optimization on the side, except perhaps for very obvious ones like removing duplicates. When spotting a deficiency, I recommend simply leaving a non-blocking note in the comments, so this can be addressed in the wake, if necessary by others. This also speeds up merges as we need not check for possible conflicts.

OK, I just created issue #3148 to clean up/unify ~reuxfr3d/~reuxfr2d as indicated above. Therefore, I will accept this PR so that it can be merged and issue #3148 can be solved afterwards.

avekens (approved):

PR #3144, opened 2023-04-27
revise dedt (and elimh)

icecream17:
> Created while thinking about (https://gist.github.com/icecream17/c85a1288ccebd09e21c7609943a9a8d6#ideas), (though I know using deduction theorems are in favor compared to dedt)

----
it is unintuitive for substitution hypotheses to not correspond like so: `A B -> ( Aph <-> Bph )`

(currently it is `A B -> ( Bph <-> Aph )`)

i commuted the consequent since either way saves steps in con3ALT and commuting the antecedent would lengthen dedt and elimh due to ifptru and ifpfal

though, commuting the antecedent would be consistent with dedth and elimhyp

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+27-Apr-23 dedt      [same]      commuted consequent
+27-Apr-23 elimh     [same]      commuted consequent
 16-Apr-23 xrhaus    [same]      moved from TA's mathbox to main set.mm
  7-Apr-23 rexsngf   [same]      moved from GS's mathbox to main set.mm
  2-Apr-23 reuxfr4d  [same]      moved from TA's mathbox to main set.mm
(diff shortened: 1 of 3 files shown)


wlammen:
This is a bit difficult, since it seems more a matter of taste than anything else.  Does it mean if the hypotheses dedt.1 and dedt.2 were swapped, then this theorem had appeared more intuitive to you?

Here we have a situation, where both the original author and a revisor either did not care much about the order, or even were more in favour of the current one.

I can easily imagine lots of requests to invert, reorder something (theorems, hypotheses, antecedents and so on) for "better understanding/systematics/logic".  One problem immediately following up here is dedth.  If you want dedt modified, dedth should follow suit.

While I see some advantage in having a common style (like listing short hypotheses first, for example) I fear the fall out of additional revisions, and possible contention within the audience.  I'd rather see the rules defined first, before applying them.

I like to hear what other developers say.

icecream17:
dedt.1 is `( if-... <-> ph ) -> ( th <-> ta )`, where if-... corresponds to ta, and ph corresponds to th
dedth.1 is `A = if... -> ( ps <-> ch )`, where A corresponds to ps, and if... corresponds to ch

So what this pr does is commute the consequent in dedt.1
(Commuting the antecedent in dedt would be consistent with dedth but it made dedt longer, so I admittedly arbitrarily decided to commute the consequent instead)

Either way this change would generally make proofs shorter, although currently the only theorem using dedt is con3ALT

wlammen:
I see.  Well if applications become shorter, then there is a technical reason for this change, and I am more in favor of it.

digama0:
I'm in favor. Substitution hypotheses are supposed to be provable without commuting anything, so `A = B -> ( ph[B] <-> ph[A] )` is backwards. Regarding `A = B -> ( ph[A] <-> ph[B] )` vs `B = A -> ( ph[B] <-> ph[A] )`, the one on the left is supposed to be the "variable" and the one on the right the "substitution", and in this case I think the "variable" is the if-expression and there are two "substitutions" for the left and right hand sides of the if respectively. So I agree with the proposed ordering.

david-a-wheeler:
I agree with @digama0 , for the reasons he gave. It's a matter of taste in some sense, but there are good arguments. This would require someone care enough to do the work... which I think will reduce the likelihood of excess work because someone wanted to "make a change for the sake of making a change" :-).

wlammen (approved):
Since all comments see an advantage in this change, I'll give my ACK here, hopefully not creating a lead case for all kinds of non-technical improvements.

Issue #3148, opened 2023-04-29
clean-up/unification of ~reuxfr3d/~reuxfr2d


avekens:
After PR #3142 is merged, ~reuxfr3d/~reuxfr2d should be cleaned up/unified as mentioned in the review remarks, because the proof of ~reuxfr2d consists essentially only of ~reuxfr3d. So ~reuxfr2d and ~reuxfr3d seem to be a kind of duplicates.

benjub:
Closing as fixed by #3149.

PR #2930, opened 2022-11-18
Shorten some Word theorems

jamesjer:
I've been reading through the Word section and found a few proofs that can be shortened.  The changes to iswrddm0, wrdnval, hashwrdn, and s2dm are too minor to justify a "shortened by" comment.

Regarding the comment on revrev, "reversion" is the act of reverting, rather than the act of reversing.

One hypothesis can be removed from wrdnfi.  It makes the proof a lot longer, but enables substantial reductions in size of the proofs of wwlksnfi and clwwlknfi.  I wasn't sure whether I should add "proof shortened by" or "revised by" comments for those two theorems.  I went with the latter, since the proof shortenings are a due to the removed wrdnfi hypothesis, rather than a direct shortening of the existing proofs.  Let me know if that is wrong.

PR #3149, opened 2023-04-29
prove pm13.183 eqsbc3 sbcel1v rmo3 without ax-13

wlammen:
Four theorems do not need ax-13.

Elimination of reuxfr2d as requested in #3148.  Since there is still a theorem reuxfr2 (but no reuxfr3), I did not rename reuxfr3d -> reuxfr2d.

Start copying results from the long standing PR #2930 of Jerry James to main, one by one, at least those passing the review.  I hope JJ is still healthy, but I fear something bad might have happened.
1.  wrdexg
2. wrdv

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15367,6 +15367,7 @@ New usage of "eqeqan1dOLDOLD" is discouraged (0 uses).
 New usage of "eqid1" is discouraged (0 uses).
 New usage of "eqresr" is discouraged (4 uses).
 New usage of "eqsb3lemOLD" is discouraged (0 uses).
+New usage of "eqsbc3OLD" is discouraged (0 uses).
 New usage of "eqsbc3rVD" is discouraged (0 uses).
 New usage of "equcomi1" is discouraged (1 uses).
 New usage of "equid1" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
eliminated reuxfr2d according to issue #3148.  Since there is still a theorem reuxfr2 (but no reuxfr3), I did not rename reuxfr3d -> reuxfr2d.

avekens:
> Elimination of reuxfr2d as requested in #3148. Since there is still a theorem reuxfr2 (but no reuxfr3), I did not rename reuxfr3d -> reuxfr2d.

I do not anderstand why reuxfr3d cannot should not be renamed. The current reuxfr3d is the deductioin version of the current reuxfr2, so its correct name (according to the naming conventions) should be reuxfr2d.

wlammen:
In my eyes reuxfr2 and reuxfr3d differ, besides one having a context 𝜑, the other not. 
reuxfr3d uses 3 classes A, B and C, like in its second hypothesis
((𝜑 ∧ 𝑥 ∈ 𝐵) → ∃*𝑦 ∈ 𝐶 𝑥 = 𝐴)

In reuxfr2 (and the former reuxfr2d) only classes A and B (which is identical to C) are used, and the same hypothesis reads
(𝑥 ∈ 𝐵 → ∃*𝑦 ∈ 𝐵 𝑥 = 𝐴)
One can rewrite reuxfr2 such that it better matches reuxfr3d, but I did not have the time for checking the consequences.

avekens:
OK, I see: reuxfr2d is a special case of reuxfr3d (with C replaced by B), so reuxfr3d is a more general theorem. And the proposal to generalize reuxfr2 accordingly should be checked (in the scope of issue #3148). However, since reuxfr2 is not used in any proof, it could also be deleted.

wlammen:
In this PR I will only do the bare minimum requested by #3148, since more than this topic is covered.  I don't want to wait for an approval, because just this detail is not exactly everybody's preferred solution covering all possible side effects.  If you don't like having a gap in the numeration, you need to rename reuxfr4d as well.  So if you want to follow more advanced concepts, please use your own PR.  If necessary I will revert the commits with respect to this issue, but I think, the basic work done here is a good starting point for everything else.

avekens (approved):
OK, but issue #3148 should remain open until reuxfr2 is generalized/deleted.

PR #3154, opened 2023-05-01
introduce projective space definitions

### LINKED ISSUE #3035: Defining projective space / projective plane
## Projective space background

https://en.wikipedia.org/wiki/Projective_space#Definition
> Given a vector space _V_ over a field _K_, the projective space **P**(_V_) is the quotient set of V \ { 0 } by the equivalence _x_ ~ _y_ iff there is a nonzero element _λ_ of _K_ such that _x_ = _λy_.

It's not obvious what properties of a vector space are needed. [benjub mentioned that _K_ need not be commutative (i.e. only need be a division ring)](https://github.com/metamath/set.mm/issues/787#issuecomment-1410974603).

I think the domain of **P** would be _left vector spaces_ then? https://us.metamath.org/mpeuni/df-lvec.html

> An especially common case would be when _V_ = _K_<sup>_n_+1</sup> (i.e. the set of ordered _n_+1–element–tuples of _K_, i.e. singletons pairs triples etc.)
> In such cases, the projective space could be denoted **P**<sub>_n_</sub>(_K_), or **PG**(_n_, _K_).

Would use (n **PG** K) for metamath.

> If _K_ is a finite field with _q_ elements, **P**<sub>_n_</sub>(_K_) is often denoted **PG**(_n_, _q_).

I'm ignoring this but it's unfortunate that the **PG** notation can be ambiguous. I would guess this notation exists because there's an isomorphism between finite projective spaces.

## Projective plane background

> If _n_ is one or two, a projective space of dimension _n_ is called a **projective line** or **projective plane**, respectively.

There are several formulations like the one above, but the defining properties are:
> https://en.wikipedia.org/wiki/Projective_plane#Definition
> (https://ericmoorhouse.org/handouts/projective_planes.pdf)

1. ∀ two distinct points ∃ a unique line containing both
2. ∀ two distinct lines ∃ a unique intersection (point)
3. ∃ four points of which no three are collinear

Note that since projective planes have points (Base) and lines ([LineG](https://us.metamath.org/mpeuni/df-lng.html)) there would have to be some conversion function (say, `toPrPl`).

(Using Word, i.e. 0 --> points, 1 --> lines, 2 --> planes, etc., seems nice for the conversion. Points, lines, planes, etc. are automatically defined for projective spaces, I think: https://en.wikipedia.org/wiki/PG(3,2), though I don't know how. Smth about subspaces, linear independency, dimensions? There's also the definition at https://en.wikipedia.org/wiki/Projective_space#Synthetic_geometry and I don't see how these separate definitions of projective space are the same.)

## Original purpose of definitions

All 3 of the elliptic curve sources I'm using describe elliptic curves using projective _planes_ in some way or another:
> Elliptic Curves, Second Edition by Dale Husemoller
> Rational Points on Elliptic Curves by Joseph H. Silverman
> https://dms.umontreal.ca/~mlalin/WIM2012-ellipticcurves.pdf

I haven't really heard about projective spaces until benjub's comment.
Originally I was just going to define projective plane, construct the real projective plane, etc. But sometimes the rational projective plane is used. And it turns out ``(toPrPl ` (2 PG ℝfld))`` happens to be the real projective plane. So I think defining at least **PG** would be beneficial. Don't know about **P** though.

## resources

note to self: here are some resources

https://webspace.maths.qmul.ac.uk/p.j.cameron/pps/pps1.pdf
https://www.cis.upenn.edu/~jean/gma-v2-chap5.pdf

icecream17:
the definitions are not used yet, these theorems just prove their soundness

the definitions are definitely up to revision so this is like a "draft" but otherwise this is good to go <sup>(hopefully)</sup>

tweak real subtraction symbol
move qseq12 to main set.mm
reduce axioms and dvs in raleqbidv and rexeqbidv

resolves #3035

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+30-Apr-23 qseq12    [same]      moved from PM's mathbox to main set.mm
 27-Apr-23 dedt      [same]      commuted consequent
 27-Apr-23 elimh     [same]      commuted consequent
 16-Apr-23 xrhaus    [same]      moved from TA's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -17331,6 +17331,7 @@ New usage of "quoremnn0ALT" is discouraged (0 uses).
 New usage of "r1omALT" is discouraged (0 uses).
 New usage of "r1pwALT" is discouraged (0 uses).
 New usage of "raleleqALT" is discouraged (0 uses).
+New usage of "raleqbidvOLD" is discouraged (0 uses).
 New usage of "ralsngOLD" is discouraged (0 uses).
 New usage of "ralxfrALT" is discouraged (0 uses).
 New usage of "rb-ax1" is discouraged (5 uses).
@@ -17396,6 +17397,7 @@ New usage of "reuccats1lemOLD" is discouraged (1 uses).
 New usage of "reuccats1vOLD" is discouraged (0 uses).
 New usage of "rexbidvALT" is discouraged (0 uses).
 New usage of "rexbidvaALT" is discouraged (0 uses).
+New usage of "rexeqbidvOLD" is discouraged (0 uses).
 New usage of "rexsngOLD" is discouraged (0 uses).
 New usage of "rhmsubcALTV" is discouraged (1 uses).
 New usage of "rhmsubcALTVcat" is discouraged (0 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -29246,17 +29246,38 @@ choice between (what we call) a "definitional form" where the shorter
...
+    $d x A $.  $d x B $.  $d x ph $.
+    raleqbidvOLD.1 $e |- ( ph -> A = B ) $.
+    raleqbidvOLD.2 $e |- ( ph -> ( ps <-> ch ) ) $.
+    $( Obsolete version of ~ raleqbidv as of 30-Apr-2023.  Equality deduction
+       for restricted universal quantifier.  (Contributed by NM, 6-Nov-2007.)
+       (Proof modification is discouraged.)  (New usage is discouraged.) $)
+    raleqbidvOLD $p |- ( ph -> ( A. x e. A ps <-> A. x e. B ch ) ) $=
+      ( wral raleqdv ralbidv bitrd ) ABDEIBDFICDFIABDEFGJABCDFHKL $.
+
+    $( Obsolete version of ~ raleqbidv as of 30-Apr-2023.  Equality deduction
```
icecream17, line 29277:
```suggestion
    $( Obsolete version of ~ rexeqbidv as of 30-Apr-2023.  Equality deduction
```

tirix (approved):

Issue #3150, opened 2023-04-29
Quick question


GinoGiotto:
Lately you probably noticed that I'm shortening proofs with the minimizer and you also probably noticed that I'm separating the job into multiple pull requests to make the review process humanly feasible.

My question is: would you like a big pull request nicely separated into multiple commits? Each of them would correspond to an analysed section of set.mm, and they would be nicely ordered without intermediate commits (such as fixes or mistake corrections). For me this would make the job quicker.

As usual the commits would only contain proofs automatically minimized with the command `MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*` and nothing else. Let me know if you like this proposal.

david-a-wheeler:
Overall, yes. In general it's best to divide things into multiple commits. That makes it easier to review, for one thing. Also, if someone changes something else, that may create a merge conflict - merge conflicts can be easier to deal with when the changes are smaller. 

Thanks for being willing to work on this!

GinoGiotto:
Thanks for the answer! But maybe I did not express myself properly.

I meant doing fewer pull requests, but with many commits inside them. Each pull request would be very big (much bigger than the ones I did recently), but since they would contain many commits they would be still reviewable.

The reason I ask this is because the estimated amount of pull requests at the current rate would be around 40, and it's a bit time consuming, but with the proposed method I would reduce them to only 2 or 3.

david-a-wheeler:
I'd much rather have many separate pull requests, each of which is smaller. Smaller changes are easier to process. A PR with many changes increases the likelihood of a merge conflict. Also, even if there's a merge conflict, with smaller changesets we can easily merge the PRs that *don't* conflict.

PRs don't cost anything. You can have 40 PRs instead of 2, that sounds great & I'd prefer it.

Anyone else?

wlammen:
Apart from technical problems like merge conflicts, I'd like to add: I do not work as a professional for Metamath, which means I need to find spare time to do a review, for instance.  It is far easier for me to set 15 min aside for a small task than several hours for a big one.  In addition a big task may cause my attention to decrease over time, so the scrutiny will suffer in the end.

I am much in favor of smaller PRs.

GinoGiotto:
[...]

PR #3155, opened 2023-05-02
Shortened proofs 7

GinoGiotto:
Continuation of https://github.com/metamath/set.mm/pull/3127. Proof minimizer ran between the 6000th and the 7000th theorems that don't have a `proof modification is discouraged` label in their comment.

No new axiom dependencies are introduced. For evidence I recommend to scan my [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all) branch instead of this pull request, so that you'll have it already checked for all commits up to **shortened proofs 22**  which will be the 36th pull request of this series. Axiom usage is recommended to be compared to my [develop](https://github.com/GinoGiotto/set.mm/tree/develop) branch, to be sure that any change is caused only by my minimizations.

 This series will finish at **shortened proofs 25.5** which will correspond to the end of main (so deprecated sections and mathboxes are excluded). My computer will end the job in about 3-5 days from now.

At the moment, axiom dependencies can be scanned using [metamath.exe](https://github.com/metamath/set.mm/issues/3150#issuecomment-1529081181), but it's quite slow, in the future I recommend using [metamath-knife](https://github.com/david-a-wheeler/metamath-knife) which has in preparation [a much faster tool](https://github.com/david-a-wheeler/metamath-knife/pull/107) which will be available soon.

I can also upload in my fork the result of the axiom scan if you consider it as valid evidence, although I think your own scan of [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all) is probably more valuable. In general any suggestion on how to present the evidence is welcome.

For more info you may be insterested in consulting the discussion here https://github.com/metamath/set.mm/issues/3150


david-a-wheeler:
Thanks so much for doing this! +1.

david-a-wheeler (approved):

david-a-wheeler:
I didn't review the changes in detail, but the approach seems sound.

wlammen:
I did a manual review, and it took me more than 2 hrs to roughly assess the changes in each theorem.  Often they are about using a deduction version vs. its tautological form.  In many cases the overall proof logic is just a bit streamlined.  And in a few instances replacements were not so obvious, but finally turned out to be possible given the axiom list of the original theorem.

I am not going to do that again.  I simply cannot afford to spend all my spare time verifying these commits.  Unless the PR will get smaller, say covering some 20 odd theorems, a manual control and visual check nothing bad had happened will not occur on my side again.  If nobody else chimes in, we are reduced to trust GinoGiotto he did not make a mistake.

wlammen (approved):

david-a-wheeler:
[wlammen](https://github.com/wlammen) said:

> I did a manual review, and it took me more than 2 hrs to roughly assess the changes in each theorem. ...

First, thank you SO MUCH for checking in detail, that's greatly appreciated.

Humans are notoriously hard to predict :-), but I find that the best predictor is past behavior. So if @GinoGiotto did fine on past simplifications, the odds are good for future ones.

I think it'd be wise to double-check that no new axioms were added. This takes about an hour of computer time (1/2 an hour each to compute before & after), but almost no human time, and it's human time that's expensive. Think someone should do that? I'm happy to accept @GinoGiotto 's word that he did that (if he said he did). I could do it myself, if no one else wants to do that.

My understanding is that the assumptions & final goals aren't being changed by these shortenings, it's just shortening the proofs themselves (without adding new axioms). The proofs are checked mechanically (by multiple verifiers), so as long as results still pass the verifiers, I think it's fine to accept them.

GinoGiotto:
> I'm happy to accept @GinoGiotto 's word that he did that (if he said he did). I could do it myself, if no one else wants to do that.

Yes, I checked axiom dependencies from **shortened proofs 7** up to **shortened proofs 20.5**, but I still haven't checked between **shortened proofs 21** and **shortened proofs 22**. I plan to scan between **shortened proofs 21** and **shortened proofs 25.5** all in one go after my computer will end the job (btw my computer needed 2 hours instead of 30 min to make the scan, I'm not sure if that's because I'm running the minimizer at the same time or if I just have a slower computer).

The only differences that I found were in 5 theorems and they were all reductions in axiom dependencies, so the proof minimizer could also be used as an axiom usage minimizer (although a very inefficient one since it only found 5 theorems out of 14500).

You can check the differences in this commit I just made to show this https://github.com/metamath/set.mm/commit/b0445ac07c011c2e33354f6e76576c8857510096.

If you scan [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all) by yourself the result might differ from mine because I still haven't checked between **shortened proofs 21** and **shortened proofs 22**, but since I'm scanning the set.mm database from top to bottom, your result should (at least) be identical to mine from the first theorem of main up to `ussval`, which is the last theorem scanned in [min20.5](https://github.com/metamath/set.mm/commit/67ec0698eea3347c57b0f360d1d51282376a4061).

avekens (approved):
Just had a quick look - no conspicuousness detected.

avekens:
In my opinion, @GinoGiotto can continue his work. It's almost the same as done before (global minimizations coordinated by Norm), and the only "risk" is to introduce new axiom dependencies, as discussed before. I think the measures to minimize this risk are sufficient, and the dependencies from axioms becomes less important when the theorems become more advanced. Maybe it becomes interesting again after the provision of the axioms for real and complex numbers. but starting with part 7 at the latest, dependencies on axioms needs not be checked anymore.

PR #3156, opened 2023-05-03
misc (axioms: abbii dfnul2 noel) (shorten: n2dvds1, n2dvds3)

icecream17:
some maintenance items on my metamath todo list

om0x isn't used by any proofs so new usage can now be discouraged

---

Sidenote:
regarding sbtv, maybe it's worth it to prove `( ps -> ph ) => ( ps -> [ x / y ] ph )`? There would be `$d y ps $.` (along with `$d x y $.`) so it would be called sbtvdv or something...

```
hyp1
    ps -> ph
a1d
    ps -> ( y = x -> ph )
ancld
    ps -> ( y = x -> ( y = x /\ ph ) )
eximdv
    ps -> ( E. y y = x -> E. y ( y = x /\ ph ) )
ax6ev
    E. y y = x
mpi
    ps -> E. y ( y = x /\ ph ) )
df-sb
    [ x / y ] ph <-> ( ( y = x -> ph ) /\ E. y ( y = x /\ ph ) )
sylanbrc
    ps -> [ x / y ] ph
```

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16606,6 +16606,8 @@ New usage of "mulpqnq" is discouraged (7 uses).
 New usage of "mulresr" is discouraged (4 uses).
 New usage of "mulsrpr" is discouraged (9 uses).
 New usage of "n0lpligALT" is discouraged (0 uses).
+New usage of "n2dvds1OLD" is discouraged (0 uses).
+New usage of "n2dvds3OLD" is discouraged (0 uses).
 New usage of "naecoms-o" is discouraged (1 uses).
 New usage of "nanassOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen (approved):

wlammen:
Regarding sbtv: I wonder what value substitution has in a universe where ax-7 does not hold.  The properties of equality play an important role in theorems like those based on sbequ.  I sometimes think, df-sb should be defined after ax-7

wlammen:
Accidently hit the wrong button

GinoGiotto:
A shortening for `n2dvds3` was found by the minimizer too, but yours is shorter.

You have defeated the engine, well done :-)

PR #3157, opened 2023-05-03
Shortened proofs 8

GinoGiotto:
Continuation of https://github.com/metamath/set.mm/pull/3127. Proof minimizer executed between the 7000th and the 8000th theorems that don't have a `proof modification is discouraged` label in their comment.

No new axiom dependencies are introduced. Scan my [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all) branch and compare it to my [develop](https://github.com/GinoGiotto/set.mm/tree/develop) branch for evidence. 

During the night I checked axiom usage for the commits left in [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all)  that weren't scanned (between [shortened proofs 21](https://github.com/metamath/set.mm/commit/44f29746007b29a2ad26d5f46125d9e96e933112)  and [shortened proofs 22](https://github.com/metamath/set.mm/commit/4a5c995d5111bc81ae2dec5ad9a969db2bb5f404)), I didn't find any difference, so now I can say that the comparison of axiom usage between [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all) and [develop](https://github.com/GinoGiotto/set.mm/tree/develop) should be exactly identical to https://github.com/metamath/set.mm/commit/eb061483b6fb87eab898bb65317ff63217a12949. Some time in the future this information will get outdated since I'll add the final commits in [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all), but I'll tell you when that will happen. In the meantime you can benefit from a clear exhibit (which is this commit https://github.com/metamath/set.mm/commit/eb061483b6fb87eab898bb65317ff63217a12949) that you can compare with your results.

Instead of 1000, next commits will be splitted in chunks of 500 examined theorems to limit the size of future PRs.

I also checked for duplicates and didn't find any. There is a big shortening for `wemapso2lem`, you may be curious to look at it if you find it interesting.

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -144,7 +144,7 @@ jobs:
       - run: python3 scripts/latex_collector.py set.mm list-set.tex
       - run: pdflatex -halt-on-error list-set.tex 2>&1 | tee ,latex-output
       # Complain if the pdflatex output says something is invalid
-      - run: ! grep ' invalid ' ,latex-output
+      - run: "! grep ' invalid ' ,latex-output"
 
   ###########################################################
   smetamath-rs:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,9 +40,7 @@ If you're contributing to set.mm or iset.mm, take a look at its
 
 At some point you should join the
 [Metamath mailing list](https://groups.google.com/g/metamath).
-If you're seriously working on proofs, we'd love
-to help you get started!
-
+If you're seriously working on proofs, we'd love to help you get started!
 
 For contributing a final proof once you have one,
 you'll basically create a "pull request" on GitHub to request
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+30-Apr-23 qseq12    [same]      moved from PM's mathbox to main set.mm
+27-Apr-23 dedt      [same]      commuted consequent
+27-Apr-23 elimh     [same]      commuted consequent
 16-Apr-23 xrhaus    [same]      moved from TA's mathbox to main set.mm
  7-Apr-23 rexsngf   [same]      moved from GS's mathbox to main set.mm
  2-Apr-23 reuxfr4d  [same]      moved from TA's mathbox to main set.mm
@@ -37,7 +40,7 @@ Date      Old       New         Notes
 19-Mar-23 fvmptd2   [same]      moved from GS's mathbox to main set.mm
 15-Mar-23 mapfvd    [same]      moved from AV's mathbox to main set.mm
  2-Mar-23 wl-hbnaev hbnaev      moved from WL's mathbox to main set.mm
-21-Feb-23 moimd     moimdv      align with mobidv, eubidv 
+21-Feb-23 moimd     moimdv      align with mobidv, eubidv
 16-Feb-23 bj-ldiv   ldiv        moved from BJ's mathbox to main set.mm
 16-Feb-23 bj-rdiv   rdiv        moved from BJ's mathbox to main set.mm
 16-Feb-23 bj-mdiv   mdiv        moved from BJ's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -4231,6 +4231,15 @@
 "crngcALTV" is used by "rngcvalALTV".
 "csbopabgALT" is used by "csbcnvgALT".
 "cshnzOLD" is used by "0csh0OLD".
+"cspliceOLD" is used by "gsumsplOLD".
+"cspliceOLD" is used by "splclOLD".
+"cspliceOLD" is used by "splfv1OLD".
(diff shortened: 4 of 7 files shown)


set.mm
```diff
@@ -76238,14 +76236,14 @@ currently used conventions for such cases (see ~ cbvmpt2x , ~ ovmpt2x and
...
-      ( vx vy c1o cv comu co wceq c0 oveq2 id eqeq12d con0 wcel 1on wa coa mpan
-      ciun cvv csuc ax-mp omsuc oveq1 sylan9eq oa1suc adantr eqtrd ex wral wlim
-      om0 cuni iuneq2 uniiun syl6eqr vex omlim limuni syl5ibr tfinds ) DBEZFGZV
-      BHZDIFGZIHZDCEZFGZVGHZDVGUAZFGZVJHZDAFGZAHBCAVBIHZVCVEVBIVBIDFJVNKLVBVGHZ
-      VCVHVBVGVBVGDFJVOKLVBVJHZVCVKVBVJVBVJDFJVPKLVBAHZVCVMVBAVBADFJVQKLDMNZVFO
-      DULUBVGMNZVIVLVSVIPVKVGDQGZVJVSVIVKVHDQGZVTVRVSVKWAHODVGUCRVHVGDQUDUEVSVT
-      VJHVIVGUFUGUHUIVICVBUJZVDVBUKZCVBVHSZVBUMZHWBWDCVBVGSWECVBVHVGUNCVBUOUPWC
-      VCWDVBWEVBTNZWCVCWDHZBUQVRWFWCPWGOCDVBTURRRVBUSLUTVA $.
+      ( vx vy c1o cv comu co wceq c0 oveq2 id eqeq12d con0 wcel wa coa 1on mpan
+      ciun cvv csuc om0x omsuc oveq1 sylan9eq oa1suc adantr ex wral wlim iuneq2
```
icecream17, line 76240: om0x is discouraged?

GinoGiotto: Good catch, I didn't think about this circumstance. What happened is that it didn't have a proper discouraged label when I forked the repository, but it was edited afterwards.

I'll give a look at all pull requests after my fork and see all theorems that were discouraged in the meantime (they are not that many so I don't think that will take long), then I'll remove all the occurrences of those theorems that appear in this commits. It's a bit annoying, but not terrible. 

Definitely when a discouraged label will be added in the future I would very much appreciate being pinged.

I'll get back with more info, I need some time for this.

GinoGiotto: Now that I think about it, shouldn't the CI pipeline fail? In discouraged there is `New usage of "om0x" is discouraged (0 uses).` But that would be incorrect with this commit, so why are all checks passing without any issues?

GinoGiotto: I think I understand why it didn't fail, the discouraged label was added very recently in this commit https://github.com/metamath/set.mm/pull/3156/files. I made the pull request before that commit was merged, so the pipeline was executed when `om0x` wasn't still discouraged and so every check passed. If I trigger the CI pipeline again, it should fail then.

GinoGiotto:
Mh I thought closing and opening the PR would trigger the pipeline, but it just skips the jobs.

set.mm
GinoGiotto: Ok, I checked back through PRs and all discouraged additions that I found (besides `om0x`) were theorems that were added after my fork (so they couldn't be used by my minimization process). So it seems to me that `om0x` is the only oddity here.

I would still like to trigger the pipeline to have evidence of that tho. How can I do that?

wlammen: Change to tab 'Checks', and use the button 'Re-run jobs'

GinoGiotto: I don't understand, the correct label should be `New usage of "om0x" is discouraged (4 uses).` since in this pull request `om0x` was used in 4 thorems for minimization. But in the currently discouraged file in [metamath develop](https://github.com/metamath/set.mm) it's `New usage of "om0x" is discouraged (0 uses).` and I didn't touch that file in this PR. Shouldn't the pipeline complain about that?

wlammen: It may help to merge the 'develop' branch into your branch in this PR.  This will update the discouraged file to the latest state, including an entry for the om0x theorem, and a list of all allowed users (should be empty). My copy contains a line

New usage of "om0x" is discouraged (0 uses).

If your copy of discouraged in your branch has this line, too, then the CI process reads it, and my assumption is that it will fail.  I don't know much about the details of its implementation, but I am under the impression that it only checks that a branch is inherently consistent, and does not take into account the diff to 'develop'.

GinoGiotto: Ok, that seems to be the correct diagnosis

GinoGiotto:
I executed the minimizer again on the 4 theorems that were previously shortened using `om0x`, no improvements were found, so I committed the old proofs.

I also searched for other usages of `om0x` in all my minimization commits. I only found one occurrence, so luckily this is an easy fix.

To prevent this from happening again, I guess always updating the branch is the way to go.

icecream17 (approved):

avekens (approved):

PR #3161, opened 2023-05-06
more df-iseq to df-seq3 in iset.mm

jkingdon:
This is particularly https://us.metamath.org/ileuni/isummo.html and related theorems. Nothing other than the `seq` syntax changes.

As usual, this is one part of an ongoing series of pull requests and unless others chip in (or even if they do) it won't be done all at once. But we are moving in the direction of greater similarity with set.mm and it is pretty clear how we can get there.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -176,7 +176,6 @@
 "iseqcl" is used by "iseqp1".
 "iseqcoll" is used by "isummolem2a".
 "iseqeq1" is used by "iseqid".
-"iseqeq1" is used by "isummolem2".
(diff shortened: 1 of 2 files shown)


wlammen (approved):

jkingdon:
I don't know whether to merge this now or give it another day or so but if someone does want to merge it I recommend "rebase and merge" rather than "squash and merge". Just because the smaller commits might be informative, they are broken into meaningful commits in this case, and this branch doesn't seem to have some of the issues which prevent easily using rebase and merge.

avekens (approved):

PR #3162, opened 2023-05-06
prove abbi2i without ax-11

wlammen:
1. prove abbi2i, abbi2dv without ax-11, base abbi2i on abbi2dv to avoid duplicated proof lines
2. abbid does not need ax-10
3. abbidv does not need ax-12.  Uses new auxiliary theorems sbbidv, sbimdv and abbilem.
4. shorten abbii, clelsb3f
5. Replace s2dm with JJ's shorter proof from #2930.  Since this shortening is rather trivial, and JJ did not add a "Proof shortened" tag in his PR, I followed suit, and neither created an OLD version.
6. Replace my ad-hoc proof of wwlksnfi with JJ's even shorter one.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13397,6 +13397,7 @@ New usage of "abbi2iOLD" is discouraged (0 uses).
 New usage of "abbidOLD" is discouraged (0 uses).
 New usage of "abbidvOLD" is discouraged (0 uses).
 New usage of "abbiiOLD" is discouraged (0 uses).
+New usage of "abbiiOLDOLD" is discouraged (0 uses).
 New usage of "ablo32" is discouraged (4 uses).
 New usage of "ablo4" is discouraged (3 uses).
 New usage of "ablocom" is discouraged (6 uses).
@@ -18260,7 +18261,8 @@ Proof modification of "abbi2dvOLD" is discouraged (24 steps).
 Proof modification of "abbi2iOLD" is discouraged (18 steps).
 Proof modification of "abbidOLD" is discouraged (24 steps).
 Proof modification of "abbidvOLD" is discouraged (9 steps).
-Proof modification of "abbiiOLD" is discouraged (17 steps).
+Proof modification of "abbiiOLD" is discouraged (38 steps).
+Proof modification of "abbiiOLDOLD" is discouraged (17 steps).
 Proof modification of "abscncfALT" is discouraged (71 steps).
 Proof modification of "ackm" is discouraged (71 steps).
 Proof modification of "ad5ant123OLD" is discouraged (45 steps).
--- a/discouraged
+++ b/discouraged
@@ -13394,6 +13394,7 @@ New usage of "9cnOLD" is discouraged (0 uses).
 New usage of "a1ii" is discouraged (0 uses).
 New usage of "abbi2iOLD" is discouraged (0 uses).
 New usage of "abbidOLD" is discouraged (0 uses).
+New usage of "abbidvOLD" is discouraged (0 uses).
 New usage of "abbiiOLD" is discouraged (0 uses).
 New usage of "ablo32" is discouraged (4 uses).
 New usage of "ablo4" is discouraged (3 uses).
@@ -18256,6 +18257,7 @@ Proof modification of "9cnOLD" is discouraged (3 steps).
 Proof modification of "a1ii" is discouraged (1 steps).
 Proof modification of "abbi2iOLD" is discouraged (18 steps).
 Proof modification of "abbidOLD" is discouraged (24 steps).
+Proof modification of "abbidvOLD" is discouraged (9 steps).
 Proof modification of "abbiiOLD" is discouraged (17 steps).
 Proof modification of "abscncfALT" is discouraged (71 steps).
 Proof modification of "ackm" is discouraged (71 steps).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -25711,11 +25732,23 @@ replaced with setvar variables (see ~ cleljust ), so we don't include
...
+      BEIBGCHZBEIPCHAEBJQABEDKEBCLMN $.
+  $}
+
+  ${
+    $d x A $.
+    abbi2iOLD.1 $e |- ( x e. A <-> ph ) $.
+    $( Obsolete version of ~ abbi2i as of 6-May-2023.  (Contributed by NM,
+       26-May-1993.)  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    abbi2iOLD $p |- A = { x | ph } $=
```
avekens, line 25751: I would like to keep this theorem as ~abbi2iALT, because it is simpler than the new ~abbi2i (no auxiliary variables, shorter proof), usage of the central theorem ~abeq2 instead of the definition ~df-clab).

avekens: Since @benjub also experimented with the affected theorems (see ~bj-abbi2i etc.) in his mathbox, I would like to hear his opinion on the current changes.

wlammen: The problem with the tautological version abbi, abeq2 etc are that one direction of the bi-conditional needs fewer axioms than the other way round.  In the deduction, immediate version of these theorems this can be made visible, since they always express one direction only, by nature.  One way to express this idea explicitly is to start with the deduction versions, and synthesize the tautological form from them.

wlammen: I'd like to focus on the term 'central theorem' a bit.  I think it is driven by the desire to solve a problem once and for all.  If you have a general theorem that covers all possible variants, then this is a particular insight into a topic, that is very application friendly.  You only need to plug in all the required provisions, and here you are, even a specialized version is ready at hand.  This is not the way Metamath works in general.  It has a bias towards analysis and let you detect on what foundations your ideas are finally built upon.

To some extent this is not followed to the extreme.  For example once the real/complex numbers are fully developed from their axioms, Norm considered this a finished state, and would allow me to derive a shortening even using more complex number axioms than the original.  One can likewise think of set theory.  Once you arrive at this level, 
predicate logic could be thought of as done with.  You don't care any longer about minimizing axiom footage up to ax-13 any more.

I am not sure whether this goes along with other foundations of predicate logic as in iset.mm.  And we are here at the very basics of set theory that is still in its very early stage.

benjub: I recently haven't had much time nor motivation to work a lot on Metamath, but here are a few remarks. I worked on several ideas regarding axiom use in my mathbox, some of which made it to the main part of set.mm, and some others not (yet).

Removal of some object-redundant axioms (ax-10 -- ax-13): mostly by using lemmas with additional DV conditions.  Some theorems in my mathbox have the comment "remove ax-13 from ..." because they were proved when the corresponding theorem in the main part was using ax-13, but they weren't moved to the main part at the time for some reason. Since then, the corresponding theorem in the main part has seen ax-13 removed (sometimes through similar devices as I used), so now the comment is obsolete.  I'll review them if I find time. Do not hesitate to move some to the main part if you see fit.

Another device to remove these axioms is the improvement of definitions. An example was the change of df-nf that I proposed on the discussion group and that Wolf implemented and used for many further axiom usage reductions. Another example is the change of df-sb that I recently proposed (#3003), as Wolf mentions. I think this should be implemented too, even if I recognize that it is a lot of work that I do not have time to do in full at the moment. In particular, the current definition df-sb blurs some axiom usages, and since that change impacts axiom usage in many theorems relying on df-sb, on which there has been much activity recently, I think it should be done sooner rather than later (without that change, some of the current work on reducing axiom usage might be moot, or even would increase axiom usage with the new definition -- theorems in the family of abeq2, abbi2i are among those, so it's hard to currently tell which proofs are better).

Finally, a third line of work in my mathbox regards use of ax-ext (and df-cleq/df-clel), and I think parts of it should be moved to the main part when I find time.

avekens: Is this inline comment shifted somehow (by additional commits)? It should be applied to ~abbi2dv now....
Nevertheless, I am not so familiar with the approaches to reduce the usage of axioms, and therefore I cannot judge the current changes entirely. Since I have no concrete objections, I will accept this PR.

avekens (approved):

set.mm
```diff
@@ -25753,42 +25801,68 @@ equal class abstractions (inference form).  (Contributed by NM,
...
   $}
 
   ${
-    $d x ph $.
+    $d x y ph $.  $d y ps $.  $d y ch $.
     abbidv.1 $e |- ( ph -> ( ps <-> ch ) ) $.
     $( Equivalent wff's yield equal class abstractions (deduction form).
-       (Contributed by NM, 10-Aug-1993.) $)
+       (Contributed by NM, 10-Aug-1993.)  Avoid ~ ax-12 , based on an idea of
+       Steven Nguyen.  (Revised by Wolf Lammen, 6-May-2023.) $)
```
icecream17, line 25818:
```suggestion
       (Contributed by NM, 10-Aug-1993.)  Avoid ~ ax-10 and ~ ax-12 , based on
       an idea of Steven Nguyen.  (Revised by Wolf Lammen, 6-May-2023.) $)
```

based on abbii, it seems ~ ax-10 is also saved

wlammen: All three ax-10, ax-11 and ax-12 are finally saved, but two of them are inherited from abbid, that was rewritten, too.  The modification to abbidv addresses the final ax-12 only.  I do not list cascading effects explicitly.  Otherwise I would have to update many more comments, not only in the theorems listed here.

I put your name here to let your contribution appear less trivial than the the proof of abbii now expresses.  An astute reader following the dates in the tags will see that your modification was first, and so the proof idea was moved around.

Let me know if my addition is not sufficient.

icecream17 (approved):

PR #3166, opened 2023-05-09
rmv ax-5 from moimi + rename prospective space definitions

### LINKED ISSUE #3035: Defining projective space / projective plane

icecream17:
The ax-5 to ax-gen ("quantified" to inference form) seems to be a sort of pattern for saving axioms.
Renames PrjSp1 --> PrjSp
Renames PrjSp --> PrjSpn

Resolves https://github.com/metamath/set.mm/issues/3035 (again)

Another option is to use $\mathbb{P}_\mathrm{n}$ for df-prjspn. (This is possible since unlike df-prjsp, the symbol does not conflict with df-prm.) The only reason I propose this is that it looks nicer; though it is inconsistent and allayed somewhat by using the ``( PrjSpn ` <. n , K >. )`` notation instead.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16580,6 +16580,7 @@ New usage of "mobidvOLDOLD" is discouraged (0 uses).
 New usage of "moeqOLD" is discouraged (0 uses).
 New usage of "moeuOLD" is discouraged (0 uses).
 New usage of "mofOLD" is discouraged (0 uses).
+New usage of "moimiOLD" is discouraged (0 uses).
 New usage of "mptssALT" is discouraged (0 uses).
 New usage of "mpv" is discouraged (1 uses).
 New usage of "mulassnq" is discouraged (10 uses).
@@ -19425,6 +19426,7 @@ Proof modification of "mobidvOLDOLD" is discouraged (46 steps).
 Proof modification of "moeqOLD" is discouraged (29 steps).
 Proof modification of "moeuOLD" is discouraged (52 steps).
 Proof modification of "mofOLD" is discouraged (62 steps).
+Proof modification of "moimiOLD" is discouraged (17 steps).
 Proof modification of "mptssALT" is discouraged (57 steps).
 Proof modification of "n0lpligALT" is discouraged (74 steps).
 Proof modification of "n2dvds1OLD" is discouraged (37 steps).


set.mm
```diff
@@ -21848,10 +21848,22 @@ proposition with a distinct variable (closed form of ~ nfsb4 ).
...
   ${
+    $d x y $.  $d y ph $.  $d y ps $.
     moimi.1 $e |- ( ph -> ps ) $.
     $( The at-most-one quantifier reverses implication.  (Contributed by NM,
-       15-Feb-2006.) $)
+       15-Feb-2006.)  Remove use of ~ ax-5 .  (Revised by Steven Nguyen,
+       9-May-2023.) $)
     moimi $p |- ( E* x ps -> E* x ph ) $=
+      ( vy weq wi wal wex wmo imim1i alimi eximi df-mo 3imtr4i ) BCEFZGZCHZEIAP
+      GZCHZEIBCJACJRTEQSCABPDKLMBCENACENO $.
```
benjub, line 21858: I think the change is good, but this is an instance where this is not too important since ~mojust already uses ax-1--ax-7.

icecream17:
the verify iset.mm check failed but that's probably a testing bug since i didn't change anything in iset.mm

wlammen:
I needed to rerun the checks on my PR as well.  The second attempt was successful.

I also had difficulties adding or editing comments yesterday.  Maybe something is broken on GitHub?

benjub:
> I needed to rerun the checks on my PR as well. The second attempt was successful.
> 
> I also had difficulties adding or editing comments yesterday. Maybe something is broken on GitHub?

Not sure what happened, and I couldn't see the logs. I just reran and it worked, so I'll merge now.

PR #3168, opened 2023-05-10
Intersection points of a line and a circle (3)

avekens:
As announced in PR #3113, I simplified the theorem ~itscnhlinecirc02p, showing that a nonhorizontal line passing through a point X within a circle around the origin (and an arbitray but different point Y) intersects the circle at exactly two different points by using the definition of the set of proper ordered pairs, see ~ innhlinecirc02p: 

```( ( .0. S R ) i^i ( X L Y ) ) e. ( PrPairs ` P )```

This is equivalent with

```E. a e. P E. b e. P ( a =/= b /\ ( ( .0. S R ) i^i ( X L Y ) ) = { a , b } )```

It is still open to prove the special case for a horizontal line, which I will try next.

Changes in details:

Main set.mm:
* ~eel0001 (renamed ~mpsyl4anc), ~fprb and ~el2v moved from mathboxes

AV's mathbox:
* definition for set of proper (unordered) pairs and rearrangement of sections about unordered pairs
* general auxiliary theorems ~addsubeq0, ~resum2sqrp
* auxiliary theorems for real Euclidean space of dimension 2: ~prelrrx2b, ~ehl2eudis0lt, ~itsclc0lem*
* theorems for  intersection points of a line and a circle: ~itsclc0lem5r, ~itsclc0blem5, ~itsclc0b, ~itsclinecirc0b, ~itsclinecirc0in, ~innhlinecirc02plem, ~innhlinecirc02p

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 9-May-23 eel0001   mpsyl4anc   moved from AS's mathbox to main set.mm
+ 9-May-23 fprb      [same]      moved from SF's mathbox to main set.mm
+ 9-May-23 el2v      [same]      moved from PM's mathbox to main set.mm
  7-Apr-23 rexsngf   [same]      moved from GS's mathbox to main set.mm
  2-Apr-23 reuxfr4d  [same]      moved from TA's mathbox to main set.mm
  2-Apr-23 reuxfr3d  [same]      moved from TA's mathbox to main set.mm

### CODE REVIEW

File: set.mm [hunk 1]
```diff
@@ -7202,6 +7202,17 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
       ( wi exp41 syl3c mpd ) AEFJABCDEFLGHIBCDEFKMNO $.
   $}
 
+  ${
```
wlammen, line 7205: naming conventions:  We have some similar theorems syl1111anc, syl112anc and more in main.  The digits reflect how antecedents are grouped.  22 for example should be attached to ( ( ph /\ ps ) /\ ( ch /\ th ) ).  I wonder whether we should add a 1111 sequence to the name here.

wlammen (commented):
No other objections besides the naming style of one theorem

set.mm
avekens: So you propose to change mpsyl4anc to mpsyl1111anc? This would be quite long, but would be OK for me.
Or should it be called mpsyl211anc (which would be a little bit shorter), because the first conjuncts are directly grouped?

syl1111anc seems to be the only label using the sequence 1111...

digama0: I prefer sticking to `mpsyl4anc`. The special case of left-leaning and-trees is already specialized in simp-Nl and adNantr, and it makes sense to specialize to a linear number of lemmas rather than the exponential combinations with the sylNNNanc series.

wlammen: I am not a friend of this NNNN style either, and I don't expect many more theorems of this kind to follow (if not we simply rename them later).  That's why I used "I wonder...".  Still, the naming scheme is inconsistent, and I wanted to point that out.  I am happy if you are...

wlammen (approved):

set.mm
avekens: OK, then we should leave it as it is.

PR #3173, opened 2023-05-13
prove abeq2f without ax-13

wlammen:
1. Introduce nfcriv, a version of nfcri directly derived from nfrc, with an additional disjoint variable condition, but without the need of ax-13.  nfcri is the more versatile result, to be used in all general contexts.  But on some occasions, we can provide more DV conditions, and do not rely on ax-13 then.  This shortens several proofs that otherwise would make a laborious use of nfcr.
2. move nfrd close to other nfr* theorems.
3. nfnfcALT has no shorter proof than nfnfc any more, so delete it.
4. abeq2f is proven without ax-13 (and shortened to boot).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13398,6 +13398,7 @@ New usage of "abbidOLD" is discouraged (0 uses).
 New usage of "abbidvOLD" is discouraged (0 uses).
 New usage of "abbiiOLD" is discouraged (0 uses).
 New usage of "abbiiOLDOLD" is discouraged (0 uses).
+New usage of "abeq2fOLD" is discouraged (0 uses).
 New usage of "ablo32" is discouraged (4 uses).
 New usage of "ablo4" is discouraged (3 uses).
 New usage of "ablocom" is discouraged (6 uses).
@@ -16648,7 +16649,6 @@ New usage of "nfimdOLDOLD" is discouraged (0 uses).
 New usage of "nfmo1OLD" is discouraged (0 uses).
 New usage of "nfmod2OLD" is discouraged (0 uses).
 New usage of "nfnbiOLD" is discouraged (0 uses).
-New usage of "nfnfcALT" is discouraged (0 uses).
 New usage of "nfopdALT" is discouraged (0 uses).
 New usage of "nfs1vOLD" is discouraged (0 uses).
 New usage of "nfunidALT" is discouraged (0 uses).
@@ -18277,6 +18277,7 @@ Proof modification of "abbidOLD" is discouraged (24 steps).
 Proof modification of "abbidvOLD" is discouraged (9 steps).
 Proof modification of "abbiiOLD" is discouraged (38 steps).
 Proof modification of "abbiiOLDOLD" is discouraged (17 steps).
+Proof modification of "abeq2fOLD" is discouraged (46 steps).
 Proof modification of "abscncfALT" is discouraged (71 steps).
 Proof modification of "ackm" is discouraged (71 steps).
 Proof modification of "ad5ant123OLD" is discouraged (45 steps).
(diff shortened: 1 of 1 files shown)


jkingdon (approved):

jkingdon:
You are getting closer to https://us.metamath.org/mpeuni/nfral.html and https://us.metamath.org/ileuni/nfralxy.html which is on my mind from time to time as it is an area of divergence between set.mm and iset.mm and is related to distinct variable constraints. That's for another day, of course.

wlammen:
There is some generalization in df-ral that I find noteworthy.  Maybe one should add this to the comment of the definition.
Intuitively I assume x and A be disjoint in df-ral (in set.mm).  But A can in fact be a collection of sets A(x), and for each instance of x A may vary.  So what does x e. A mean then?  I find it difficult to come up with, for example, a sequence A(n) of sets of natural numbers, where n e. A(n) is really relevant.  Something that springs to my mind is: let A(n) be the set of integer solutions x^n - n * x + sin (pi*x).  Here the set of n with n e. A(n) is a subset of NN, and A. n e. A(n) ph is in fact a restrictive expression running over all integer n for which the above question holds.  The question is: what guarantees that the filtering expression n e. A(n) is "decidable"?  In set.mm hblem plays a role in this question.

Adding a dv condition on x and A in df-ral is certainly something to consider.

Just momentary thoughts that came to my mind while stopping at nfrald.

jkingdon:
> Adding a dv condition on x and A in df-ral is certainly something to consider.

I'm not enough of an expert to say whether it should be on the definition but it is on many theorems involving ∀ such as https://us.metamath.org/mpeuni/reupick2.html , https://us.metamath.org/mpeuni/caucvg.html (between `j` and `k`), etc.

In browsing I was a little surprised to see that constraint missing in a decent number of places where I expected it. I'd be a bit surprised if we ever use the capability of having `A` contain an `x` but it does seem to be there more than I realize. Like you it doesn't match my intuition very well.

digama0:
This kind of thing actually gave me a lot of trouble when writing the MM -> MM0 translator, and when writing the definition checker. I think that the `A` in `A. x e. A ph` should not depend on `x`. I don't mean that there should be a disjointness relation (DV conditions are allowed on term constructors in grammatical databases anyway), but rather that any occurrence of `x` in `A` should be interpreted as referring to an `x` in the outer scope. This is necessary for making the translated HOL style type of wral `class -> (set -> wff) -> wff` instead of `(set -> class) -> (set -> wff) -> wff`.

I didn't do this for df-ral (it currently still uses the "naive" definition that captures `x` occurrences in `A`), but I did apply such a transformation to df-sum, which is carefully stated with a second dummy variable so that it does not accidentally capture `x`. For df-ixp it went the other way: it was originally a mixture of capturing and non-capturing behaviors due to the two occurrences of `A` in

```
 ⊢ X𝑥 ∈ 𝐴 𝐵 = {𝑓 ∣ (𝑓 Fn 𝐴 ∧ ∀𝑥 ∈ 𝐴 (𝑓‘𝑥) ∈ 𝐵)}
```
so I refactored it to
```
 ⊢ X𝑥 ∈ 𝐴 𝐵 = {𝑓 ∣ (𝑓 Fn {𝑥 ∣ 𝑥 ∈ 𝐴} ∧ ∀𝑥 ∈ 𝐴 (𝑓‘𝑥) ∈ 𝐵)}
```
so that it is at least consistent about capturing `x` in `A`.

I would love to clean all this up, it doesn't affect much since as long as you can assume `$d x A` all the variations are equivalent. But as a general rule, I would say that the "morally correct" behavior is for bounded quantifiers and typing conditions to not have the variable being defined already in the context.

jkingdon:
> I would love to clean all this up, it doesn't affect much since as long as you can assume `$d x A` all the variations are equivalent.

You mean like this? https://github.com/metamath/set.mm/pull/3178

PR #3178, opened 2023-05-15
Add mathbox A. x e. A ph definition.

jkingdon:
This is an experiment to deal with the condundrum about what to do if A contains x in A. x e. A ph .  Isn't ready to move out of the mathbox until we better understand the choices here.

More discussion at https://github.com/metamath/set.mm/pull/3173#issuecomment-1547084388


jkingdon:
I'm not sure what would be needed to get this past mmj2 (or even whether doing this as a mathbox experiment is worth the trouble we might be seeing there), but hopefully this pull request as it exists is enough to get an idea for what is being explored here.

Here's the axiom usage from dfralfromxn :

```
show trace_back dfralfromxn/axioms
Statement "dfralfromxn" assumes the following axioms ($a statements):
  wn wi ax-mp ax-1 ax-2 ax-3 wb df-bi wa df-an wo df-or wal cv wceq wex df-ex
  wnf df-nf ax-gen ax-4 ax-5 wsb df-sb ax-6 ax-7 wcel ax-10 ax-11 ax-12 ax-13
  df-clel wral df-ralxn
```

tirix:
I think because you name your definition with "`df-`", a definitional check is launched, which checks that the definendum is not already used. And of course `A.` has been used before, so it fails.

One way to get around this would be to simply name your definition `dfralnx`, or to define an alternative symbol, like `A_` for example.

jkingdon:
> define an alternative symbol, like `A_`

Unless I missed something about how the various checks interact, this seemed like the only solution which might be capable of being merged. Which I guess we can do if we really want this in a mathbox, but at least for now I'll just leave this as a closed draft pull request to illustrate the idea of how we might define restricted quantifiers to avoid some of the oddities which can happen if ` A ` contains the variable ` x `.

digama0:
Historically, we have used the (rather blunt) instrument of putting `NEW` on the declaration and the constant name, hence `df-ralNEW $a |- ( A.NEW x e. A <-> ... )`.

tirix:
There are several other examples of explorations of alternative definitions in set.mm (e.g. alternative definitions of pairs).
In my opinion, this one is also worth it - either it will be the first step of the transformation, or it can serve as a reference.

PR #3180, opened 2023-05-16
Shortened proofs 12

GinoGiotto:
Continuation of https://github.com/metamath/set.mm/pull/3127. Proof minimizer executed between the 11501th and the 12000th theorems that don't have a `proof modification is discouraged` label in their comment.

Turns out I was optimistic about the amount of time it would take for my computer to finish the job, but it's finally done. I updated the [min-all](https://github.com/GinoGiotto/set.mm/tree/min-all) branch up to the final commit (which is [min25.5](https://github.com/metamath/set.mm/commit/53da2bcd4a8b105982256779ff1a22bf7a764fe0)). A few more theorems had their axiom usage reduced, so I updated the file listing axiom usage too (already linked above). I think this is good evidence that the minimizer is reliable, as none of the 25,475 analyzed theorems exhibited a single occurrence of accidental axiom additions.

Minimizing proofs in a systematic way takes a really long time, requiring weeks of computing power active 24/7, even during the night. I think we would benefit greatly if PR https://github.com/metamath/metamath-exe/pull/94 would be merged soon, as it claims to halve the time required to do these minimizations. I encourage the community to direct its attention towards it, it already has one approval and from reading comments it doesn't seems there are reasons to keep it pending further @digama0.

I have no idea how Norm used to do this flawlessly. The minimizer is incredibly fast for the first thousands theorems and then becomes exponentially slower towards the end of main. Theorem `footex` literally crashed metamath, so I wasn't even convinced to continue since there could have been other ones that could have behaved the same way. I decided to give the minimizer a second chance, I skipped `footex` and luckely no problems happened afterwards. Anyway I'll open an issue about that event.

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -144,7 +144,7 @@ jobs:
       - run: python3 scripts/latex_collector.py set.mm list-set.tex
       - run: pdflatex -halt-on-error list-set.tex 2>&1 | tee ,latex-output
       # Complain if the pdflatex output says something is invalid
-      - run: ! grep ' invalid ' ,latex-output
+      - run: "! grep ' invalid ' ,latex-output"
 
   ###########################################################
   smetamath-rs:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,9 +40,7 @@ If you're contributing to set.mm or iset.mm, take a look at its
 
 At some point you should join the
 [Metamath mailing list](https://groups.google.com/g/metamath).
-If you're seriously working on proofs, we'd love
-to help you get started!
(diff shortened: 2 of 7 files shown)


avekens (approved):

avekens:
From time to time, I also run the minimization script, over the whole set.mm, but only for special minimization focussing on replacing certain theorems (not global ones like @GinoGiotto  does). I never experienced problems with ~footex (maybe it was never affected). I had (runtime) problems with the ~fourier*, ~stirl* and ~stow* theorems, therefore I always exclude them.

A good idea would be to shorten these (sometimes very long) proofs by extracting adequate lemmata...

PR #3186, opened 2023-05-18
seq syntax changes in iset.mm - isum etc

jkingdon:
This is quite similar to previous pull requests in this series - use the theorems which use the preferred seq syntax and remove theorems which are no longer used because of those changes.


wlammen (approved):

david-a-wheeler:
+1

PR #3192, opened 2023-05-19
justification for df-ral without dv condition

wlammen:
1. shorten ralbii2 (reverted since the shortened proof introduces ax-5)
2. explain the consequences of x and A not being disjoint in df-ral.  This is a fall out from a recent discussion in #3173

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17340,6 +17340,7 @@ New usage of "quoremnn0ALT" is discouraged (0 uses).
 New usage of "r19.12OLD" is discouraged (0 uses).
 New usage of "r1omALT" is discouraged (0 uses).
 New usage of "r1pwALT" is discouraged (0 uses).
+New usage of "ralbii2OLD" is discouraged (0 uses).
 New usage of "raleleqALT" is discouraged (0 uses).
 New usage of "raleqOLD" is discouraged (0 uses).
 New usage of "raleqbi1dvOLD" is discouraged (0 uses).
@@ -19614,6 +19615,7 @@ Proof modification of "quoremnn0ALT" is discouraged (360 steps).
 Proof modification of "r19.12OLD" is discouraged (61 steps).
 Proof modification of "r1omALT" is discouraged (13 steps).
 Proof modification of "r1pwALT" is discouraged (151 steps).
+Proof modification of "ralbii2OLD" is discouraged (37 steps).
 Proof modification of "raleleqALT" is discouraged (26 steps).
 Proof modification of "raleqOLD" is discouraged (11 steps).
 Proof modification of "raleqbi1dvOLD" is discouraged (28 steps).
--- a/discouraged
+++ b/discouraged
@@ -17340,7 +17340,6 @@ New usage of "quoremnn0ALT" is discouraged (0 uses).
 New usage of "r19.12OLD" is discouraged (0 uses).
 New usage of "r1omALT" is discouraged (0 uses).
 New usage of "r1pwALT" is discouraged (0 uses).
-New usage of "ralbii2OLD" is discouraged (0 uses).
 New usage of "raleleqALT" is discouraged (0 uses).
 New usage of "raleqOLD" is discouraged (0 uses).
 New usage of "raleqbi1dvOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -27969,6 +27998,16 @@ choice between (what we call) a "definitional form" where the shorter
...
+  ${
+    $d x ph $.
+    ralbidv2.1 $e |- ( ph -> ( ( x e. A -> ps ) <-> ( x e. B -> ch ) ) ) $.
+    $( Formula-building rule for restricted universal quantifier (deduction
+       form).  (Contributed by NM, 6-Apr-1997.) $)
+    ralbidv2 $p |- ( ph -> ( A. x e. A ps <-> A. x e. B ch ) ) $=
+      ( cv wcel wi wal wral albidv df-ral 3bitr4g ) ADHZEIBJZDKPFICJZDKBDELCDFL
+      AQRDGMBDENCDFNO $.
+  $}
+
```
tirix, line 28010: It looks like you still have not moved back `ralbidv2` to its original location.

tirix (commented):
Clever!

So it looks like you don't need a new definition of `df-ral` after all, right?

wlammen:
If you really want the exceptional case (x and A not disjoint) be covered, well, then one can reason along these lines and keep df-ral as is.  I personally have not come across a convincing application of this special case, needed so badly that it should be directly included in the definition.  So there is just the technical simplification of avoiding a dv condition left.

One can simply add a dfralv where the dv condition is added to df-ral, and see how far this will take, and whether there are essential branching points from df-ral anywhere.  I don't know.  From a more abstract point of view dfralv seems to be preferable, as the discussion in #3173 indicates.  I am not an expert in this area, though, so I cannot really say what is best.

tirix (approved):

tirix:
Maybe @digama0's MM -> MM0 translator is the compelling reason for that definition?
That's up to him to tell.

digama0:
The fact that metamath doesn't really declare the types of its variables (specifically as regards binding structure) leads to this kind of issue, where the translator has to infer the types, and depending on how you look at it it is either inferring incorrectly that `A` is a function of `x`, or it is actually reflecting a real behavior that proofs are allowed to depend on (although it would be very weird for them to). This is the kind of issue that often happens in weakly typed languages, where it's never clear whether a peculiar behavior is by design or not.

I would just like to clarify this matter, and as I see it, if one actually considers the desired binding structure if one simply asks which it should be:

1. `A. x e. A ph`
2. `A. x e. A ph(x)`
3. `A. x e. A(x) ph`
4. `A. x e. A(x) ph(x)`

the only reasonable answer is (2), and that (4) is the answer in metamath today is only an accident of the definition. The MM -> MM0 translation is only relevant to the extent that it forces one to explicitly consider and answer this question for every binder in set.mm.

To motivate (2) a bit more, let's consider the `{ x e. A | ph }` binder instead. Here we would like to be able to say that if `A` is a set then `{ x e. A | ph }` is a set, but this isn't really true if `A` depends on `x` - we instead need to say that `{ x | x e. A }` is a set. And there are some basic counterexamples too, `{ x e. { x } | T. }` is not a set even though `{ x }` is. This is just obviously a bad design to me. One of the things you want from bounded quantifiers is for them to be *bounded*, not just be an extra syntax for predicates. I certainly understand why behavior (4) is naturally arrived at given that the types are not explicit but the definitions are, and `A. x ( x e. A -> ph )` is the shortest definition you could have which generally has the right behavior, especially if you are not considering the x,A not disjoint case. But it's just wrong when A depends on x.

wlammen:
I just played around with direct usages of df-ral in theorems, to see how difficult a transition to an improved definition would be.  df-alsc is a definition in a mathbox based on restricted quantifiers and imitating their current limitation.  ssrabf  is a theorem that needs the disjointness be replaced with F/_ x A.  I presume there are plenty of obstacles in the way of such an adventure.

digama0:
With an improved df-ral, ssrabf can be stated without the F/_ x A assumption, and df-alsc could be fixed by only changing `E. x x e. A`  to `E. y y e. A`. I'm not sure how many other theorems directly interact with `df-ral` without the x,A assumption.

wlammen:
I am in favor of a cleaner definition.  I just said a transition is not going to be easy.

digama0:
That's probably true.

PR #3193, opened 2023-05-19
use risset in clel5, cleljust2, repnpcan

icecream17:
use risset to save axioms in clel5

In df-clel, the notation `A e. B` has already been seen before in two contexts: `x e. y` and df-clab `x e. { y | ph }`. cleljust covers the first case, so here cleljust2 covers the second case.

Sidenote: Using df-clab with benjub's definition of substitution, one can prove `x e. { y | T. }`, which when combined with df-clel, proves ax-6 without ax-6. Luckily, bj-dfclel already happens to depend on ax-6 (as well as the intended ax-8)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14526,6 +14526,7 @@ New usage of "chub1i" is discouraged (18 uses).
 New usage of "chub2" is discouraged (14 uses).
 New usage of "chub2i" is discouraged (24 uses).
 New usage of "chunssji" is discouraged (0 uses).
+New usage of "clel5OLD" is discouraged (0 uses).
 New usage of "cleljustALT" is discouraged (0 uses).
 New usage of "cleljustALT2" is discouraged (0 uses).
 New usage of "clelsb3fOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
> use risset to save axioms in clel5
> 
> In df-clel, the notation `A e. B` has already been seen before in two contexts: `x e. y` and df-clab `x e. { y | ph }`. cleljust covers the first case, so here cleljust2 covers the second case.
> 
> Sidenote: Using df-clab with benjub's definition of substitution, one can prove `x e. { y | T. }`, which when combined with df-clel, proves ax-6 without ax-6. Luckily, bj-dfclel already happens to depend on ax-6 (as well as the intended ax-8)

This is not exactly "luck": I gave some thought about these definitions before putting them in my mathbox.

As for your cleljust2: it should depend only on core axioms, and it depends on ax-10 ax-12 ax-13 only because of the equality theorem for df-sb. An equality theorem should rely in my opinion only on core axioms, like ~ bj-ssbequ. This is an argument in favor of df-ssb over df-sb.

wlammen (approved):

wlammen:
I suggest to move cleljust2 right before df-clel and reference it there, as was done with cleljust.

avekens (approved):

tirix (approved):

tirix:
> I suggest to move cleljust2 right before df-clel and reference it there, as was done with cleljust.

That remark makes a lot of sense but on the other hand this PR has been approved and waiting for 3 days.
@icecream17 would you rather have me merge it as-is ?

icecream17:
Let's merge as is; cleljust is currently placed before ax-9 instead of right before df-clel, so I'll wait for the substitution axioms to be reduced before moving cleljust2

PR #3195, opened 2023-05-20
Shortened proofs 14

GinoGiotto:
Continuation of https://github.com/metamath/set.mm/pull/3127. Proof minimizer executed between the 13501th and the 14000th theorems that don't have a `proof modification is discouraged` label in their comment.

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -144,7 +144,7 @@ jobs:
       - run: python3 scripts/latex_collector.py set.mm list-set.tex
       - run: pdflatex -halt-on-error list-set.tex 2>&1 | tee ,latex-output
       # Complain if the pdflatex output says something is invalid
-      - run: ! grep ' invalid ' ,latex-output
+      - run: "! grep ' invalid ' ,latex-output"
 
   ###########################################################
   smetamath-rs:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,9 +40,7 @@ If you're contributing to set.mm or iset.mm, take a look at its
 
 At some point you should join the
 [Metamath mailing list](https://groups.google.com/g/metamath).
-If you're seriously working on proofs, we'd love
-to help you get started!
-
+If you're seriously working on proofs, we'd love to help you get started!
 
 For contributing a final proof once you have one,
 you'll basically create a "pull request" on GitHub to request
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,16 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+18-May-23 fndmd     [same]      moved from GS's mathbox to main set.mm
+18-May-23 elpreimad [same]      moved from GS's mathbox to main set.mm
+18-May-23 fnfvimad  [same]      moved from GS's mathbox to main set.mm
+16-May-23 subdivcomb1 [same]    moved from SF's mathbox to main set.mm
+ 9-May-23 eel0001   mpsyl4anc   moved from AS's mathbox to main set.mm
+ 9-May-23 fprb      [same]      moved from SF's mathbox to main set.mm
+ 9-May-23 el2v      [same]      moved from PM's mathbox to main set.mm
+30-Apr-23 qseq12    [same]      moved from PM's mathbox to main set.mm
+27-Apr-23 dedt      [same]      commuted consequent
+27-Apr-23 elimh     [same]      commuted consequent
 16-Apr-23 xrhaus    [same]      moved from TA's mathbox to main set.mm
  7-Apr-23 rexsngf   [same]      moved from GS's mathbox to main set.mm
  2-Apr-23 reuxfr4d  [same]      moved from TA's mathbox to main set.mm
@@ -37,7 +47,7 @@ Date      Old       New         Notes
 19-Mar-23 fvmptd2   [same]      moved from GS's mathbox to main set.mm
 15-Mar-23 mapfvd    [same]      moved from AV's mathbox to main set.mm
  2-Mar-23 wl-hbnaev hbnaev      moved from WL's mathbox to main set.mm
-21-Feb-23 moimd     moimdv      align with mobidv, eubidv 
+21-Feb-23 moimd     moimdv      align with mobidv, eubidv
 16-Feb-23 bj-ldiv   ldiv        moved from BJ's mathbox to main set.mm
 16-Feb-23 bj-rdiv   rdiv        moved from BJ's mathbox to main set.mm
 16-Feb-23 bj-mdiv   mdiv        moved from BJ's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -4231,6 +4231,15 @@
 "crngcALTV" is used by "rngcvalALTV".
 "csbopabgALT" is used by "csbcnvgALT".
 "cshnzOLD" is used by "0csh0OLD".
+"cspliceOLD" is used by "gsumsplOLD".
+"cspliceOLD" is used by "splclOLD".
+"cspliceOLD" is used by "splfv1OLD".
+"cspliceOLD" is used by "splfv2aOLD".
+"cspliceOLD" is used by "splidOLD".
+"cspliceOLD" is used by "spllenOLD".
+"cspliceOLD" is used by "splval2OLD".
+"cspliceOLD" is used by "splvalOLD".
+"cspliceOLD" is used by "splvalpfxOLD".
 "cvati" is used by "cvbr4i".
 "cvbr" is used by "cvbr2".
 "cvbr" is used by "cvcon3".
@@ -4278,6 +4287,7 @@
 "cvpss" is used by "cvnsym".
 "cvpss" is used by "cvntr".
 "cvr2N" is used by "cvrval4N".
+"dedtOLD" is used by "con3ALTOLD".
 "df-0" is used by "ax1ne0".
 "df-0" is used by "axi2m1".
 "df-0" is used by "axpre-mulgt0".
@@ -5656,6 +5666,7 @@
 "elhmop" is used by "idhmop".
 "elhmop" is used by "lnophmi".
 "elhmop" is used by "pjhmopi".
+"elimhOLD" is used by "con3ALTOLD".
 "elimnv" is used by "elimph".
 "elimnvu" is used by "ajfun".
 "elimnvu" is used by "imsmet".
@@ -5906,6 +5917,8 @@
 "extwwlkfabOLD" is used by "extwwlkfabelOLD".
 "extwwlkfabelOLD" is used by "numclwwlk1lem2fOLD".
 "extwwlkfabelOLD" is used by "numclwwlk1lem2foaOLD".
+"f1rhm0to0OLD" is used by "kerf1hrmOLD".
+"f1rhm0to0OLD" is used by "rim0to0OLD".
 "fh1" is used by "chirredlem3".
 "fh1" is used by "cm2j".
 "fh1" is used by "fh1i".
@@ -10149,11 +10162,8 @@
 "nsmallnq" is used by "nqpr".
 "nsmallnq" is used by "reclem2pr".
 "numclwlk2lem2f1oOLD" is used by "numclwwlk2lem3OLD".
-"numclwlk2lem2f1oOLDOLD" is used by "numclwwlk2lem3OLDOLD".
 "numclwlk2lem2fOLD" is used by "numclwlk2lem2f1oOLD".
-"numclwlk2lem2fOLDOLD" is used by "numclwlk2lem2f1oOLDOLD".
 "numclwlk2lem2fvOLD" is used by "numclwlk2lem2f1oOLD".
-"numclwlk2lem2fvOLDOLD" is used by "numclwlk2lem2f1oOLDOLD".
 "numclwwlk1lem2f1OLD" is used by "numclwwlk1lem2f1oOLD".
 "numclwwlk1lem2f1oOLD" is used by "numclwwlk1lem2OLD".
 "numclwwlk1lem2f1oOLD" is used by "numclwwlk1lem2OLDOLD".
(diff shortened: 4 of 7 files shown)


GinoGiotto:
Theorem `sumeq2sdv` is identical to `sumeq2ad`, so one of them can be removed.

wlammen:
sumeq2sdv is older and has more applications, but has the longer proof.

So I suggest to proceed as follows:
1. Add a tag (Proof shortened by Glauco Siliprandi, 5-Apr-2020.) to sumeq2sdv.
2. Replace the proof of sumeq2sdv with the shorter one `( wceq ralrimivw sumeq2d ) ABCDEACDGEBFHI $.`
3. Delete theorem sumeq2ad.
4. Do a text search for sumeq2ad and replace each occurrence with sumeq2sdv.
5. Rewrap set.mm.

GinoGiotto:
That's curious. So my understanding is that `/rewrap` works for comments, but not for proofs. To rewrap proofs the command  `SAVE PROOF <label> /COMPRESSED` is needed. Since `sumeq2ad` was used by quite a few theorems, I simply did `SAVE PROOF * /COMPRESSED` to rewrap all proofs in set.mm, but changes show that there are a few rewrapped proofs that don't have a `sumeq2ad` label in it, so they weren't edited by me. So it looks like there were some proofs in set.mm that were not properly rewrapped and went overlooked until now (and not detected by the CI checks).

avekens:
The preferred way to adjust proofs in compressed format (after renaming labels, as in the current case) is "`save proof */compressed/fast`" (or, in short, "`sa p */c/f`").
The option /fast means "The proof is merely reformatted and not recompressed."
If the option "/f" had been used here, maybe some of the modified proofs would not have been changed. I do not know if the compressed format is unique, so that it can be checked by the CI checks (and I assume this is not checked). 

I do not know if this PR can be merged in the current state, but if the proofs are still correct (verified by the CI checks), there should be no problems.

GinoGiotto:
Should I reformat them again? Indeed I didn't use the `/FAST` option, I only ran `SAVE PROOF * /COMPRESSED`. However, upon reviewing modifications, I noticed that this recompression provided some benefits:

This is the list of all theorems in the last commit a7fd4070d8ace98ddbdd43a0404b73b87e9317e2 that don't contain `sumeq2ad`, but were still affected by  `SAVE PROOF * /COMPRESSED`:
* `kerf1ghm` -> shortened by 60 bytes
* `f1ghm0to0` -> shortened by 15 bytes
* `sacgr` -> shortened by 7 bytes
* `dimcl` -> same length, but 1 less step
* `0ellsp` -> same length, but 1 less step
* `lmhmlvec2` -> same length, but 2 less steps
* `kerlmhm` -> same length, but 1 less step

Overall, this recompression either resulted in a reduced number of steps or to a further shortening in the compressed format, which appear like additional improvements to me. Is there a specific reason to discard them?

icecream17 (approved):
Like above, I don't think there are problems; approve

sidenote:
It seems like not using the `FAST` option is another way to possibly auto-shorten proofs, which is pretty interesting...

There are several ways to represent the same proof in a compressed format, which involve when to use `Z` when using `Z` doesn't lose or save bytes, how to order the proof labels (which determine which label corresponds to which code, the codes go A through T, then UA to UT, VA to VT, etc. YA to YT, UUA to UUT, etc. with base 5-base 20 numbering.)

I've wondered if there were any improvements that could be made in some edge cases. For example, auto-shortening where the bytes are the same but there are less web page tokens. Some complexity in the existing code is: maybe UA is used more often than T, maybe reordering the labels minimizes how early the rewrapping is. I know for certain the latter is optimized.

avekens:
> Should I reformat them again? Indeed I didn't use the `/FAST` option, I only ran `SAVE PROOF * /COMPRESSED`. However, upon reviewing modifications, I noticed that this recompression provided some benefits:
> 
> This is the list of all theorems in the last commit [a7fd407](https://github.com/metamath/set.mm/commit/a7fd4070d8ace98ddbdd43a0404b73b87e9317e2) that don't contain `sumeq2ad`, but were still affected by `SAVE PROOF * /COMPRESSED`:
> 
>     * `kerf1ghm` -> shortened by 60 bytes
> 
>     * `f1ghm0to0` -> shortened by 15 bytes
> 
>     * `sacgr` -> shortened by 7 bytes
> 
>     * `dimcl` -> same length, but 1 less step
> 
>     * `0ellsp` -> same length, but 1 less step
> 
>     * `lmhmlvec2` -> same length, but 2 less steps
> 
>     * `kerlmhm` -> same length, but 1 less step
> 
> 
> Overall, this recompression either resulted in a reduced number of steps or to a further shortening in the compressed format, which appear like additional improvements to me. Is there a specific reason to discard them?

These theorems were added/modified by @tirix recently (in May 2023).  I think he did not reformat the proofs accordingly (and nobody else after his commit(s)). Therefore, I think the changes made by this PR are acceptable.

avekens (approved):

tirix:
> These theorems were added/modified by @tirix recently (in May 2023).  I think he did not reformat the proofs accordingly (and nobody else after his commit(s)). Therefore, I think the changes made by this PR are acceptable.

These proofs were generated by Yamma, and I did not run 'save proof /compressed' on all of them, but mostly only 'save proof /fast /compressed'. I assume Metamath.exe's compression algorithm is better than Yamma's in that case?

In any case, you can go ahead with the PR.

digama0:
> I do not know if the compressed format is unique, so that it can be checked by the CI checks (and I assume this is not checked).

It is unique (or more precisely, there is a specific algorithm that is implemented by at least metamath-exe, metamath-knife and mmj2 which produces a unique output from a given input step-list-with-backreferences), but I think metamath-exe doesn't always do the full deduplication of the step list (which is also unique from a given input proof) and the `/fast` modifier controls whether it does deduplication.

digama0:
> I've wondered if there were any improvements that could be made in some edge cases. For example, auto-shortening where the bytes are the same but there are less web page tokens. Some complexity in the existing code is: maybe UA is used more often than T, maybe reordering the labels minimizes how early the rewrapping is. I know for certain the latter is optimized.

As you say, the latter is optimized as part of the standardized algorithm that is used for compressed proof formatting. The former is optimized in the sense that Norm tried out several strategies for the base of the step numbering strategy (e.g. to use 20+5 vs 19+6 base encoding for the step numbers) and found that 20+5 was the best on an early version, and last time we checked 19+6 was slightly better on the then-current set.mm, but overall, having a standardized algorithm which is reproducible and minimizes superfluous diffs is more valuable than having a superoptimized result here. I would prefer not to superoptimize the encoding on an individual proof if it will be destroyed by the next `sa p */c` because then we will effectively no longer be able to use this to auto-format proofs.

digama0:
> > These theorems were added/modified by @tirix recently (in May 2023).  I think he did not reformat the proofs accordingly (and nobody else after his commit(s)). Therefore, I think the changes made by this PR are acceptable.
> 
> These proofs were generated by Yamma, and I did not run 'save proof /compressed' on all of them, but mostly only 'save proof /fast /compressed'. I assume Metamath.exe's compression algorithm is better than Yamma's in that case?

In that case, it sounds like the issue is not the compression algorithm but rather the *deduplication* algorithm. There is a very simple deduplication rule I highly encourage you to implement: whenever a statement is proved more than once, refer to the original statement instead of the second copy. You can either take the first one, or be clever about which proof to keep, but in either case the result will have the property that no statement is proved twice in the same proof, and `save proof /compressed` and `save proof /compressed /fast` will have the same result.

tirix:
I've opened [an issue in Yamma's repository](https://github.com/glacode/yamma/issues/7) to track the issue with its compression algorithm.

I will try to run `save proof /compressed` on new proofs to avoid the discrepancy in the future when using Yamma until that issue is resolved.

In the mean time I think we can merge this PR (which is actually not related).

glacode:
> > I do not know if the compressed format is unique, so that it can be checked by the CI checks (and I assume this is not checked).
> 
> It is unique (or more precisely, there is a specific algorithm that is implemented by at least metamath-exe, metamath-knife >and mmj2 which produces a unique output from a given input step-list-with-backreferences), but I think metamath-exe >doesn't always do the full deduplication of the step list (which is also unique from a given input proof) and the `/fast` modifier >controls whether it does deduplication.

Hi everybody,

as I've already written here

https://github.com/metamath/set.mm/issues/2710#issuecomment-1214358302

the compressed proof (built by mmj2 and metamath.exe) is not unique. You can read an example at the linked issue.

Yamma, Metamath and mmj2 produce the same non-compressed proof, but different compressed proofs.

Mmj2 even produces different compressed proofs at the bottom of the .mmp file and in the .mmt file.

Metamath and mmj2 try to format nicely the labels in the proof, but they use different knapsack algorithms and this leads to different sequence of labels and then to different compression strings.

Yamma leaves the order (FIFO) of the uncompressed proof (when the labels' sequence are the same, the compressed strings are the same for both Yamma, mmj2 and Metamath, for what I've seen).

glacode:
> I've opened [an issue in Yamma's repository](https://github.com/glacode/yamma/issues/7) to track the issue with its compression algorithm.
> 
> I will try to run `save proof /compressed` on new proofs to avoid the discrepancy in the future when using Yamma until that issue is resolved.
> 
> In the mean time I think we can merge this PR (which is actually not related).

Hi Thierry,

are you using the "standard formatting" script recommended here?

https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md#formatting-recommendation-prior-to-submitting-a-pull-request

Doesn't it fix the format?

tirix:
> are you using the "standard formatting" script recommended here?

Previously, when using MMJ2 (through my Eclipse plugin), I did not need to use that script. I copy/pasted directly the compressed proof from the MMP into set.mm and it "just worked", with correct indentation and everything.

As-is, that script recommendation will not work because it is the fast method which does not recompute the compressed proof, just reformats it (`SAVE PROOF * /COMPRESSED /FAST`).

As I mentioned, I can just use `SAVE PROOF * /COMPRESSED` (without the `/FAST`), I just need to run it on each of my proofs one by one. It's a bit tedious but it works, and I think the benefits of using Yamma still outweigh that minor issue.

Note also that the proofs generated by Yamma are valid and pass all the checkers - questions will arise if someone runs `SAVE PROOF * /COMPRESSED` and gets a different output.

digama0:
@glacode 

> Hi everybody,
> 
> as I've already written here
> 
> [#2710 (comment)](https://github.com/metamath/set.mm/issues/2710#issuecomment-1214358302)
> 
> the compressed proof (built by mmj2 and metamath.exe) is not unique. You can read an example at the linked issue.

The formatting on that post is broken, so it is difficult to tell whether indentation is the source of the problem. In any case, discrepancies in the algorithm should be treated as bugs and reported to the relevant projects.

EDIT: Here's my reconstruction of your post with fixed formatting:
> ```
> MM> show proof cbviota /compressed
> Proof of "cbviota":
> ---------Clip out the proof below this line to put it in the source file:
>       ( vw vz weq wb wal cab cuni cio wsb nfv nfbi equequ1 bibi12d sbequ12 nfsb
>       nfs1v cbval sbequ sbie syl6bb bitri abbii unieqi dfiota2 3eqtr4i ) ACHJZK
>       ZCLZHMZNBDHJZKZDLZHMZNACOBDOUPUTUOUSHUOACIPZIHJZKZILUSUNVCCIUNIQVAVBCACIU
>       CVBCQRCIJAVAUMVBACIUACIHSTUDVCURIDVAVBDACIDFUBVBDQRURIQIDJZVABVBUQVDVAACD
>       PBAIDCUEABCDGEUFUGIDHSTUDUHUIUJACHUKBDHUKUL $.
> ---------The proof of "cbviota" (336 bytes) ends above this line.
> ```
> the mmj2 proof in the GUI
> ```
> $= ( vw vz cv wceq wb wal cab cuni cio wsb nfv nfbi equequ1 syl6bb
>    nfs1v sbequ12 bibi12d cbval nfsb sbequ sbie bitri abbii dfiota2
>    unieqi 3eqtr4i ) ACJZHJZKZLZCMZHNZOBDJZUOKZLZDMZHNZOACPBDPUSVDU
>    RVCHURACIQZIJZUOKZLZIMVCUQVHCIUQIRVEVGCACIUBVGCRSUNVFKAVEUPVGAC
>    IUCCIHTUDUEVHVBIDVEVGDACIDFUFVGDRSVBIRVFUTKZVEBVGVAVIVEACDQBAID
>    CUGABCDGEUHUAIDHTUDUEUIUJULACHUKBDHUKUM $.
> $)
> ```
> the proof in the .mmt file
> ```
>   ${
>     $d ph w z $. $d ps w z $. $d w x z $. $d w y z $. 
>     cbviota.1 $e           |- ( x = y -> ( ph <-> ps ) ) $.
>     cbviota.2 $e          |- F/ y ph $.
>     cbviota.3 $e           |- F/ x ps $.
>     $( Change bound variables in a description binder. (Contributed by Andrew 
>        Salmon, 1-Aug-2011.) $)
>     cbviota $p       
>                    |- ( iota x ph ) = ( iota y ps ) $=
>       ( vw vz cv wceq wb wal cab cuni cio wsb nfv nfbi equequ1 nfs1v sbequ12
>       nfsb bibi12d cbval sbequ sbie syl6bb bitri abbii unieqi dfiota2 3eqtr4i )
>       ACJZHJZKZLZCMZHNZOBDJZUOKZLZDMZHNZOACPBDPUSVDURVCHURACIQZIJZUOKZLZIMVCUQV
>       HCIUQIRVEVGCACIUAVGCRSUNVFKAVEUPVGACIUBCIHTUDUEVHVBIDVEVGDACIDFUCVGDRSVBI
>       RVFUTKZVEBVGVAVIVEACDQBAIDCUFABCDGEUGUHIDHTUDUEUIUJUKACHULBDHULUM $.
> 
>   $}
> ```

Note that the width of the proof is different: the metamath proof and the mmt proof are 79 chars wide (minus a 6 space indent), while the mmj2 GUI proof is 66 chars (minus a 3 space indent). The indent of a proof in mmj2 is determined from the input file, and the indentation of the right margin (79 chars for set.mm) is determined by a configuration flag. I expect this is the source of your issue, not an algorithm inconsistency.

EDIT 2: Since metamath.exe proof has the indentation parameters as the mmt proof, one would expect them to come up with the same proof, but they still differ and this is because they are actually different proofs: you can see that the metamath proof has a `weq` reference while mmj2 uses `cv wceq` instead. mmj2 does not use syntax theorems ever, because it has a parser and turns the parse trees directly into syntax proofs. metamath.exe uses syntax theorems sometimes just as a result of its heuristic syntax derivation algorithm and sometimes because of MINIMIZE.

avekens:
I think this is not the right place to continue the discussion about compressed proofs. Maybe it can be transferred to and continued at another place. Tomorrow at the latest, I will merge (and thereby close) this PR, because it is accepted by two reviewers and no open issue is reported.

glacode:
> > are you using the "standard formatting" script recommended here?
> 
> Previously, when using MMJ2 (through my Eclipse plugin), I did not need to use that script. I copy/pasted directly the compressed proof from the MMP into set.mm and it "just worked", with correct indentation and everything.

The problem could be that a subsequent PR (from someone else) would run the recommended script, and it should produce many diffs to your mathbox, also (I'm wondering why it didn't happen in the past).

Are you aware of this (old) thread?

Mandatory "standard" formatting prior to submitting a pull request?
https://groups.google.com/g/metamath/c/zZMeeKEC1Dg/m/Xh683K0eCAAJ

avekens:
I think this is not the right place to continue the discussion about compressed proofs. Maybe it can be transferred to and continued at another place. Tomorrow at the latest, I will merge (and thereby close) this PR, because it is accepted by two reviewers and no open issue is reported.

tirix:
> Are you aware of this (old) thread?
> 
> Mandatory "standard" formatting prior to submitting a pull request? https://groups.google.com/g/metamath/c/zZMeeKEC1Dg/m/Xh683K0eCAAJ

Yes, but that thread is about a CI action. 
The action is in place, and my PR have always passed all CI, including the one where I submitted `f1ghm0to0` and the other theorems listed above.

There are two flavours of SAVE PROOF, with or without FAST. Here from MM HELP:
```
    / FAST - The proof is merely reformatted and not recompressed.
```
The recommended formatting is to run the `/FAST` option, so even when following that rule, the non-standard proof is not recompressed.
But here GinoGiotto ran the command without `/FAST`, so when he did, those proofs were actually recompressed.

avekens:
If there are still any problems with the compression, they should be discussed/solved with PR #3202.

PR #3185, opened 2023-05-17
Fundamental theorem of linear algebra

tirix:
In my mathbox:
- Definition of the dimension of a vector space, `~ df-dim`. I chose `dim` as a symbol as it is standard, it was yet unused, and I think vector space dimension is a notion important enough to have such a simple symbol;
- Main theorem is `~ dimkerim`, the sum of the dimensions of the kernel and of the image of a vector space homomorphism is the dimension of the space.
- Misc. supporting theorems

In main:
- 3 theorems moved from @glacode 's mathbox to main part,
- strengthened `~ f1rhm0to0`, `~ rim0to0` and `~kerf1rhm`: those theorems originally applied to rings, this replaces them with the same theorems on groups.

In Brendan Leahy's mathbox:
- revised the comment for `~lindsadd` as per @benjub's [comment](https://github.com/metamath/set.mm/pull/3077#issuecomment-1454692411) in #3077

PR #3196, opened 2023-05-20
More theorems about the dimension of vector spaces

tirix:
This completes #3185 with more theorems about the dimension of vector spaces.
Especially:
- the dimension of a subspace is less than the dimension of the main space (`~lssdimle`)
- the dimension of a subspace plus the dimension of its left coset equals the dimension of the main vector space (`~qusdimsum`)


set.mm
```diff
@@ -467236,6 +467245,280 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+      ABCVMOZSWRWTWJWEWSXAWJWETWRWSWKNXKWJWEVOABXBXLAXDXFXCHIXEBCXERVAVCAXDCVPS
+      XLXBUOHCVQCVRVSVTWRWSWTXAWKWJCEBGWJRWKRWAVEWBAUAPZWLSZWRESZWSESZWFZTJEWJW
+      LDUGOZWMWRWNBXMCDWSGWJRWLRWMRAXDXQHVIAXFXQIVIAXNXOXPWCFXRRWNRAXNXOXPWDAXN
+      XOXPWGWHHWI $.
+
+    ${
+      $d V k u v x y z $.  $d F k u v y z $.  $d G k u v x y z $.
+      $d M k u v x y z $.  $d N k u v x y z $.  $d ph k u v x y z $.
+      quslmhm.f $e |- F = ( x e. V |-> [ x ] ( M ~QG G ) ) $.
+      $( If ` G ` is a normal subspace in ` M ` , then the "natural map" from
```
savask, line 467503: What is a "normal subspace"? There's a similar notion from topology, but it's not relevant.
Did you mean a "submodule" (that is required in order to have a factor-module correctly defined)?

Also the last sentence probably should be "... is a left module homomorphism from **M** to M / G ."

set.mm
```diff
@@ -467676,6 +468012,32 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
       YKZVWPUYHVVJXUSUYDUYIUYMVUGEUYKUBVWTUYERYTYLYMUYFUYG $.
   $}
 
+  ${
+    $d U x $.  $d W x $.  $d Y x $.
+    qusdimsum.x $e |- X = ( W |`s U ) $.
+    qusdimsum.y $e |- Y = ( W /s ( W ~QG U ) ) $.
+    $( Let ` W ` be a vector space, and let ` X ` be a subspace.  Then the
+       dimension of ` W ` is the sum of the dimension of ` X ` and the
+       dimension of the coset of ` X ` .  First part of theorem 5.3 in [Lang]
```
savask, line 468021: Not a "coset", but a "factor space" (or a "quotient space" if you want to be consistent with your earlier comment, but Lang uses "factor space").

savask (commented):
Great work! I think I spotted one typo and one misuse of the terminology.

My traditional remark is "please minimize". For example, the proof of `qusker` can be reduced from 1233 bytes to 1122 (other theorems will also benefit from minimization).

set.mm
tirix: You're right, thanks! I've fixed that in the last commit.

savask:
@tirix I think you added some .mmt files to the pull request.

tirix:
> @tirix I think you added some .mmt files to the pull request.

Oh thanks, these will be part of the next PR!

PR #3197, opened 2023-05-21
Finish intuitionizing "Open sets of a metric space"

jkingdon:
This includes the rest of the standard bounded metric theorems (all of which intuitionize with the obvious change to the notation for minimum).

The theorems in the rest of the section either can be proved as stated in set.mm (with relatively small amounts of inuitionizing needed), or are not proved and noted in mmil.html.

Includes one additional minimum theorem (`xrminltinf`, an obvious consequence of theorems we have).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9940,6 +9940,13 @@ intuitionistic and it is lightly used in set.mm</TD>
   in iset.mm</a></td>
 </tr>
 
+<tr>
+  <td>prdsxms , prdsms</td>
+  <td><i>none</i></td>
+  <td>This would need more extensive development of theorems related
+  to the Xs_ syntax (not just ~ df-prds itself).</td>
+</tr>
+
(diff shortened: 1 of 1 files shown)


mmil.raw.html
```diff
@@ -9893,6 +9922,44 @@
...
+  <td>Awaits revision of ~ df-ress as described at <a
+  href="https://github.com/metamath/set.mm/issues/3022">Clean
+  up multifunction restriction operator for extensible structures
+  in iset.mm</a></td>
+</tr>
+
+<tr>
+  <td>prdsxms , prdsms</td>
+  <td><i>none</i></td>
+  <td>This would need more extensive development of theorems realted
```
avekens, line 9952: realted -> related?

jkingdon: Thanks, fixed.

tirix (approved):

PR #3201, opened 2023-05-22
Vector dimensions of the matrix space and of the euclidean space

tirix:
- Moved `~frlmlvec` to main set.mm
- More theorems about vector space dimension in my mathbox, including the dimension of the matrix space (NxN), and of the euclidean space (N)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+22-May-23 frlmlvec  [same]      moved from SN's mathbox to main set.mm
 20-May-23 dfss7     [same]      moved from AV's mathbox to main set.mm
 18-May-23 fndmd     [same]      moved from GS's mathbox to main set.mm
 18-May-23 elpreimad [same]      moved from GS's mathbox to main set.mm


set.mm
```diff
@@ -467694,6 +467784,45 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
       CADGSNHOSRDEFSNPQ $.
   $}
 
+  ${
+    $d X x $.  $d Y x $.
+    lbslsat.v $e |- V = ( Base ` W ) $.
+    lbslsat.n $e |- N = ( LSpan ` W ) $.
+    lbslsat.z $e |- .0. = ( 0g ` W ) $.
+    lbslsat.y $e |- Y = ( W |`s ( N ` { X } ) ) $.
+    $( The basis of an atom.  (Contributed by Thierry Arnoux, 20-May-2023.) $)
```
savask, line 467793: I've never seen the word "atom" being used in this setting in the literature. Does it just mean "a line"? I don't feel strongly about this, but maybe you could write more explicitly like "A nonzero vector ` X ` is a basis of a line spanned by ` X `."

savask (commented):
Cool stuff! I have purely cosmetic remarks.

One of them is about the usage of the term "atom" to mean 1-dimensional subspaces. I don't think it's a widespread term, but now I see there are some theorems about "atoms" in Norman's mathbox, in particular, he has theorem [lsatlspsn2](https://us.metamath.org/mpeuni/lsatlspsn2.html) which looks very similar to your `lbslsat`. What is the relation between these two?

set.mm
tirix: Yes, I was referring to `LSAtom` as defined in Norm's Mathbox, without wanting to pull that definition out. That's exactly what theorem ~[lsatlspsn2](https://us.metamath.org/mpeuni/lsatlspsn2.html) is. 
"Line" and your proposal is probably better, but I could refer to Norm's definition anyway.

tirix:
> One of them is about the usage of the term "atom" to mean 1-dimensional subspaces. I don't think it's a widespread term, but now I see there are some theorems about "atoms" in Norman's mathbox, in particular, he has theorem [lsatlspsn2](https://us.metamath.org/mpeuni/lsatlspsn2.html) which looks very similar to your `lbslsat`. What is the relation between these two?

Actually `~lsatlspsn2` is purely expanding the definition of "Atoms", I did not find any theorem in Norm's mathbox which would match `lbslsat`.

avekens (approved):
great!

PR #3203, opened 2023-05-23
Intuitionize section "Continuity in metric spaces"

jkingdon:
This is a short section which mostly intuitionizes except for theorems related to concepts we don't yet have in iset.mm.

The change to set.mm is to remove some TODO comments which are no longer applicable.


tirix (approved):

avekens (approved):
OK, only the comments "$( TODO - delete this if not used? $)" are deleted in set.mm, because the corresponding theorems ARE used...

jkingdon:
> OK, only the comments "$( TODO - delete this if not used? $)" are deleted in set.mm, because the corresponding theorems ARE used...

If anyone has some more context maybe we can write a better comment or something but the direct answer to the comments which had been there is, yes, "the corresponding theorems are used".

Issue #3198, opened 2023-05-21
redefine df-ral and friends


wlammen:
The current definition df-ral `(∀𝑥 ∈ 𝐴 𝜑 ↔ ∀𝑥(𝑥 ∈ 𝐴 → 𝜑))`, sloppily phrased as `For all x in class A holds 𝜑`, is perceived as problematic, when x and A are not disjoint.  In this case the filtering is done by querying whether x belongs to A(x), itself an element of a collection of classes A varying with x.  This is rarely needed, if at all, and the comments to theorems using df-ral show, that their authors often are unaware of this exception.

I propose to change the definition to  `(∀𝑥 ∈ 𝐴 𝜑 ↔ (Ⅎ𝑥𝐴 ∧ ∀𝑥(𝑥 ∈ 𝐴 → 𝜑)))`.  From this definition a theorem `(∀𝑥 ∈ 𝐴 𝜑 ↔ ∀𝑥(𝑥 ∈ 𝐴 → 𝜑))` with disjoint x, A follows using only axioms up to ax-5.

The change is too massive to be carried out in a single pull request.  So for some time we need both definitions (that extends to other quantifiers like df-rex, df-rmo, df-reu) in parallel.  Any comments, suggestions, objections?  See also #3192, where this proposal was discussed.

digama0:
I have an alternative proposal:
```
( A. x e. A ph <-> A. y ( y e. A -> [ y / x ] ph ) )
```
(You are also welcome to rephrase this to something provably equivalent; I don't keep up with subsystems of the MM axioms.)

I think that your definition will have more side effects than mine, because yours has an extra conjunct and isn't just a forall anymore. You can't easily generalize that to other binders like df-rex (would it also be a conjunction, or would it be a disjunction now?) or df-rab (it's not even a proposition in this case) or df-sum (at this point there really is no place to put a side condition).

wlammen:
Nice trick! Now x may appear in A without any effect.  One should still assume x y, ph y, A y disjoint.

 I'd replace `[ y / x ] ph` with `A x ( x = y -> ph )` to be independent of the flavour of df-sb.

wlammen:
[...]

PR #3204, opened 2023-05-24
prepare change of df-ral

wlammen:
... and remove outdated and unused OLD theorems

See #3198


set.mm
```diff
@@ -531784,10 +531788,71 @@ sides of the biconditional correlate (they are the same), if they exist at
...
+
+  ${
+    $d y ph $.  $d x y A $.
+    $( Restricted quantification allows for a simpler expression, if we can
+       assume ` x ` and ` A ` are disjoint.  (Contributed by Wolf Lammen,
+       23-May-2023.) $)
+    wl-nfcralv $p |- ( A. x wl-in A ph <-> A. x ( x e. A -> ph ) ) $=
+      ( vy wwl-in cv wcel weq wal df-wl-ral 19.21v albii wl-nfcrallem 3bitr2i
+      wi ) ABCEDFCGZBDHAOZBIOZDIPQOBIZDIBFCGAOBIABDCJSRDPQBKLABDCMN $.
+  $}
```
tirix: Maybe if this theorem was named for example `dfralv`, this would strongly hint that this is a more common definition.
Also in the comment, you could write e.g. "Alternate definition of the restricted universal quantification, assuming `x` does not appear in `A`.".
And to cap it all, I would point to this theorem from the actual (new) definition `df-ral`.

wlammen: `And to cap it all, I would point to this theorem from the actual (new) definition df-ral.`
Then I would lose the 'mathbox only' property.

Lets start in my mathbox first and develop some fundamental theorems there before they are moved to main in a bunch.

tirix: You can do that all in your mathbox, sorry I should have used the theorem names with `wl-` to be more clear. I was suggesting to have:
- `wl-nfcralv` named `wl-dfralv`,
- the comment of `wl-dfralv` clearly stating it's an alternate definition,
- `wl-dfralv` referenced prominently from `df-wl-ral`.

tirix: Actually you have done the first two updates in your last commit, sorry this has crossed my comment.

What do you think about the third one, to make a reference to `wl-dfralv` early in the comment of `df-wl-ral`?
For example:
_Define an improved restricted universal quantification.  See `~ wl-dfralv` for a version which requires `A` and `x` to be disjoint, allowing a shorter statement. This version does not interpret elementhood verbatim in ` A. x wl-in A ph ` . [...]_

tirix: Crossed again! All good now :-)

tirix (approved):

PR #3206, opened 2023-05-24
redefine restricted existential quantification

wlammen:
#3198 continued in my mathbox.  Now the restricted existential quantifier is introduced.  Both the definitions of df-wl-ral and df-wl-rex have been designed to arrive at alternative definitions with minimal axiom footprint.


set.mm
```diff
@@ -531681,10 +531688,81 @@ sides of the biconditional correlate (they are the same), if they exist at
...
 
+  ${
+    $d x y $.  $d A y $.  $d ph y $.
+    $( An alternate definition of restricted universal quantification
+       ( ~ df-wl-ral ) using substitution.  (Contributed by Wolf Lammen,
+       25-May-2023.) $)
+    wl-dfral2 $p |- ( A. x wl-in A ph <-> A. y ( y e. A -> [ y / x ] ph ) ) $=
+      ( wl-ral cv wcel weq wi wal wsb df-wl-ral sb6 imbi2i albii bitr4i ) ABDEC
+      FDGZBCHAIBJZIZCJQABCKZIZCJABCDLUASCTRQABCMNOP $.
+  $}
```
tirix, line 531703: There is already a `dfral2` in the main part of set.mm. it might conflict if we want to merge those changes into the main part.

tirix: Same remark for `~dfrex2`

tirix (approved):

avekens (approved):

PR #3207, opened 2023-05-24
Remove axiom of choice dependency from `~dimval`

tirix:
Gérard Lang noticed that one could remove the dependency on the axiom of choice from `~dimval`.
I only managed to remove it in the case of vector spaces of finite dimension.


wlammen (approved):
Mathbox only: No objections whatsoever.

avekens (approved):

PR #3213, opened 2023-05-28
alternate definitions of rmo

wlammen:


set.mm
```diff
@@ -531482,12 +531495,132 @@ expression can be simplified (see ~ wl-dfrexf , wl-dfrexv ).
...
+      ( wl-rex wl-rmo wa cv wcel wex wmo wl-reu weu wl-dfrexv wl-dfrmov anbi12i
+      df-wl-reu df-eu 3bitr4i ) ABCDZABCEZFBGCHAFZBIZUABJZFABCKUABLSUBTUCABCMAB
+      CNOABCPUABQR $.
+  $}
+
+  ${
+    $d y x $.  $d y ph $.  $d y A $.
+    $( Restricted existential uniqueness ( ~ df-wl-reu ) allows a
+       simplification, if we can assume all occurrences of ` x ` in ` A ` are
+       bounded.  (Contributed by Wolf Lammen, 28-May-2023.) $)
```
avekens, line 531592: if ~wl-dfreuf was placed before ~wl-dfreuv, it could be used in the (afterwards much shorter) proof of ~wl-dfreuv. This would also hold for similar cases.

wlammen: The proof would become shorter, but also depend on more axioms (ax-10 and ax-12 in the case of wl-dfreuv).  I am possibly in a better position to see this, because I produce the HTML pages of these theorems to get a visual impression of the proof.

avekens (approved):

set.mm
avekens: I had a look at similar cases, and I found out that there may be many of the reverse cases. For example ~ralcom is proven by using ~ralcomf: 
`ralcom $p |- ( A. x e. A A. y e. B ph <-> A. y e. B A. x e. A ph ) $= ( nfcv ralcomf ) ABCDECDFBEFG $.`
But this requires the following axioms:

```
MM> sh tr ralcom/ax
Statement "ralcom" assumes the following axioms ($a statements):
  wn wi ax-mp ax-1 ax-2 ax-3 wb df-bi wa df-an wo df-or wal cv wceq wtru df-tru
  wex df-ex wnf df-nf ax-gen ax-4 ax-5 wsb df-sb ax-6 ax-7 wcel ax-10 ax-11
  ax-12 ax-13 df-clel wnfc df-nfc wral df-ral
```

If ~ralcom was proven directly (mainly using ~r2al and ~alcom), the proof would become longer, but would require fewer axioms:
```
ralcom $p |- ( A. x e. A A. y e. B ph <-> A. y e. B A. x e. A ph ) $=
      ( cv wcel wa wi wal wral ancomst 2albii alcom bitri r2al 3bitr4i ) BFDGZC
      FEGZHAIZCJBJZSRHAIZBJCJZACEKBDKABDKCEKUAUBCJBJUCTUBBCRSALMUBBCNOABCDEPACB
      EDPQ $.
```

```
MM> sh tr ralcom/ax
Statement "ralcom" assumes the following axioms ($a statements):
  wn wi ax-mp ax-1 ax-2 ax-3 wb df-bi wa df-an wal cv wex df-ex ax-gen ax-4 ax-5 wcel ax-11 wral df-ral
```

This means, that ax-10, ax-12 and ax-13 are not required for ~ralcom.

I don't know how often such cases occur in set.mm, and maybe at some point in set.mm (maybe after ZFC?)  shorter proofs become more preferred than proofs with fewer axioms (we had this discussion already , hadn't we?). Maybe we should open an issue for that...

wlammen: I added a lot of revisions where dependencies on axioms were removed, mostly in predicate logic.   There is still an ongoing effort to further cut down on axiom usage, not only by me.  It is (currently) a stated goal on the https://us.metamath.org/mpeuni/conventions.html page (paragraph Minimizing axioms...).  In my experience, after finishing predicate logic, and starting ZFC, there is barely any opportunity to still remove a dependency on the propositional axioms up to ax-3.  Even very simple consequences of freshly introduced definitions use all of these axioms, and I do not look out for minimizing in this regard any more.  From the point of Restricted quantifiers is Propositional logic finished, and saving one of its axioms here and there bears hardly any extra insight.  But this does not hold for Predicate Logic.

As soon as I am done with restricted class abstraction in my mathbox I will point out in issue #3198 that the new series of definitions are cleaning up a corner case with a semantic difficulty, but introduce ax-11 on a broader scale.  Maybe this will spark off a discussion you want to see.

PR #3214, opened 2023-05-28
revise df-prjsp

icecream17:
Restricts the domain of the equivalence relation to not include 0, to make theorems using the equivalence operator nicer

----
sidenote: I'm stumped on trying to prove something like:
```
  ${
    $d ph y $.  $d t y $.  $d x y $.
    $( Equivalence between ~ df-ssb and ~ df-sb . $)
    sb11 $p |- ( A. y ( y = t -> A. x ( x = y -> ph ) ) <->
                 ( ( x = t -> ph ) /\ E. x ( x = t /\ ph ) ) ) $=
      ? $.
  $}
```

<s>The reverse direction becomes `( x = t -> ph ) /\ y = t /\ E. x ( x = t /\ ph ) -> A. x ( x = y -> ph )` after a simple alrimiv which seems so possible but I can't find a solution??? `( x = y -> ph )` is easily provable but there's no easy way to add the quantifier.</s> edit: invalid alrimiv because of $d v condition


benjub:
> sidenote: I'm stumped on trying to prove something like:
> 
> ```
>   ${
>     $d ph y $.  $d t y $.  $d x y $.
>     $( Equivalence between ~ df-ssb and ~ df-sb . $)
>     sb11 $p |- ( A. y ( y = t -> A. x ( x = y -> ph ) ) <->
>                  ( ( x = t -> ph ) /\ E. x ( x = t /\ ph ) ) ) $=
>       ? $.
>   $}
> ```

This is basically https://us.metamath.org/mpeuni/bj-dfssb2.html (its RHS is df-sb by https://us.metamath.org/mpeuni/dfsb7.html).

tirix (approved):
Looks good!

set.mm
```diff
@@ -617513,6 +617517,26 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+
+    prjspnf1om1.1 $e |- F = ??? $.
+    @( A bijection between an n-dimensional projective space and its
+       (n-1)-dimensional affine and projective spaces.  (Contributed by Steven
+       Nguyen, ??-??-2023.) @)
+    prjspnf1om1 @p |- ( ( N e. NN /\ K e. DivRing /\ ?? ) ->
+      F : ( N PrjSpn K ) -1-1-onto-> ( ( K ^m ( 0 ..^ ( N - 1 ) ) ) u.
+                                       ( ( N - 1 ) PrjSpn K ) ) ) @=
+      ? $.
+  $)
```
tirix, line 617538: `~f1oun` seems to be the best top-level theorem to start with.
Then you can probably explicitly construct the mappings.

wlammen:
@icecream17 left a comment in #3219 , so I think he saw the remarks here, too, and it is safe to merge this pull request.

PR #3220, opened 2023-06-01
prjsprellsp; revise df-prjspn

icecream17:
turns out that the dimension of a projective space is the dimension of  the affine space minus one

sidenote: i don't exactly see how the oft stated online equivalence in (the theoretical) prjspnf1om1 can be proven since there hasn't been an established "1" vector that all vectors are a multiple of.

sidenote edit: It can't be proven generally, only for specific cases. (Counterexample: real projective plane)
It is provable when scalar multiplication = scalar product

(note to self: df-psr)


wlammen (approved):
I am no expert in projective spaces, but since it is in your mathbox only...

avekens (approved):

avekens:
PR approved and label "mathbox only" added.

tirix:
Nice! I suppose the next step might be to use the `dim` function I defined recently in my mathbox?

PR #3227, opened 2023-06-04
misc shortenings, mteqand, resubdi, fltne

icecream17:
eldifsnneq was originally used in prjspreln0 but then I used mteqand

I shortened eldifsnneq using eldifsn but then the auto minimizer found  eldifsni and the proof is obliterated. The minimizer doesn't find  eldifsni by itself though.

fiminre was originally changed to reduce axioms but I put proof  shortened because the proof is really just the parallel of fimaxre

dffltz was auto minimized using ss2ralv

fltne was made because I thought I found a simple proof of FLT but of  course it didn't work out.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 1-Jun-23 eldifsnneq [same]     moved from BJ's mathbox to main set.mm
 22-May-23 frlmlvec  [same]      moved from SN's mathbox to main set.mm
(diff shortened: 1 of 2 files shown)


tirix:
> fltne was made because I thought I found a simple proof of FLT but of course it didn't work out

Too bad, your proof would probably have fit in the margin! ;-)

wlammen (approved):

tirix (approved):

tirix:
Perhaps the FLT proof should go into the "April's Fool" section? :-)

PR #3229, opened 2023-06-06
Shortened proofs 19

GinoGiotto:
Continuation of https://github.com/metamath/set.mm/pull/3127. Proof minimizer executed between the 18501th and the 19000th theorems that don't have a `proof modification is discouraged` label in their comment.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -1495,7 +1495,6 @@
 "basendx" is used by "indistpsx".
 "basendx" is used by "ipostr".
 "basendx" is used by "lmodstr".
-"basendx" is used by "matbas".
 "basendx" is used by "mgpress".
 "basendx" is used by "odubas".
 "basendx" is used by "oppcbas".
@@ -13756,7 +13755,7 @@ New usage of "bafval" is discouraged (38 uses).
 New usage of "bamalipOLD" is discouraged (0 uses).
 New usage of "barbariALT" is discouraged (0 uses).
 New usage of "barocoALT" is discouraged (0 uses).
-New usage of "basendx" is discouraged (25 uses).
+New usage of "basendx" is discouraged (24 uses).
 New usage of "baseval" is discouraged (0 uses).
 New usage of "bcs" is discouraged (6 uses).
 New usage of "bcs2" is discouraged (2 uses).


GinoGiotto:
Ok, this seems easy to diagnose. Theorem `matbas` used to depend on theorem `basendx`, which is discouraged. The minimizer dropped `basendx` usage and replaced it with theorems that are not discouraged. So this is a somewhat involuntary improvement.

<pre>$( Index value of the base set extractor.  (Normally it is preferred to work
     with ` ( Base `` ndx ) ` rather than the hard-coded ` 1 ` in order to make
     structure theorems portable.  This is an example of how to obtain it when
     needed.)  (New usage is discouraged.)  (Contributed by Mario Carneiro,
     2-Aug-2013.) $)
  basendx $p |- ( Base ` ndx ) = 1 $=
    ( cbs c1 df-base 1nn ndxarg ) ABCDE $.</pre>

I just need to update the discouraged file and that's it, I don't think there are more actions to make.

set.mm
```diff
@@ -265783,18 +265769,16 @@ as element of (the base set of) ` ( ( 1 ... n ) Mat R ) `.
     $( The matrix ring has the same base set as its underlying group.
        (Contributed by Stefan O'Rear, 4-Sep-2015.) $)
     matbas $p |- ( ( N e. Fin /\ R e. V ) -> ( Base ` G ) = ( Base ` A ) ) $=
-      ( cfn wcel wa cbs cfv cnx cmulr cotp cmmul co cop c1 c3 csts eqid basendx
-      matval fveq2d baseid 1re ltneii mulrndx neeqtrri eqnetri setsnid syl6reqr
-      1lt3 ) DHIBEIJZAKLCMNLZBDDDOPQZRUAQZKLCKLUOAURKABUQCDEFGUQUBUDUEUQUPKCUFM
-      KLSUPUCSTUPSTUGUNUHUIUJUKULUM $.
+      ( cfn wcel wa cbs cfv cnx cmulr cotp cmmul co cop csts eqid matval fveq2d
+      baseid basendxnmulrndx setsnid syl6reqr ) DHIBEIJZAKLCMNLZBDDDOPQZRSQZKLC
```
avekens, line 265773: Yes, basendx is replaced by basendxnmulrndx, which is not discouraged (and also not based on discouraged theorems). Therefore, the replacement/shortening is OK.

avekens (approved):

avekens:
@tirix @digama0 are there any objections concerning the shortening of the proof of ~matbas? If not, I will merge this PR today or tomorrow latest.

set.mm
tirix: Looks fine to me!

One naive related question would be, since `basendx` is discouraged, shouldn't also `df-base` be discouraged?
Indeed what we want to discourage here is the direct use of the hidden slot definition.

tirix (approved):

set.mm
avekens: > One naive related question would be, since `basendx` is discouraged, shouldn't also `df-base` be discouraged? Indeed what we want to discourage here is the direct use of the hidden slot definition.

Yes, this is also my opinion. But there are some more theorems which directly depend on the specific slot definition, and it is maybe difficult to draw the line between discouraged and not discourade theorems (should ~basendxnmulrndx also be discouraged?). But this is off topic here...

david-a-wheeler: @tirix - Can you (or someone else) create a *new* issue proposing that df-base be discouraged, so we can track that as its own issue?

Issue #3223, opened 2023-06-02
Define derivative in iset.mm


jkingdon:
The basic idea from set.mm does not have obvious problems: `( S _D F )` is a function whose value at `x` is the limit of ``( ( f ` z ) - ( f ` x ) ) / ( z - x )`` as z approaches `x`, and the domain of `( S _D F )` is the interior points of `dom F` where that limit exists (see for example [dvbssntr](https://us.metamath.org/mpeuni/dvbssntr.html) and [dvres](https://us.metamath.org/mpeuni/dvres.html)).

However, the details in set.mm rely a lot on using decidability of real number equality in places like the if statement in https://us.metamath.org/mpeuni/df-limc.html and `dom 𝑓 ∖ {𝑥}` in https://us.metamath.org/mpeuni/df-dv.html

It seems like this should be intuitionizable, because the kind of limit here doesn't feel super different from the process at https://us.metamath.org/ileuni/df-cnp.html and related theorems but I'm not sure I quite see how this is supposed to work, or have found any relevant literature references.

digama0:
I'm pretty sure the if statement in df-limc is eliminable. Rather than reusing the "continuous at a point" predicate, you can just use one of the direct definitions like ellimc2. You will have to grapple with the "neighborhood minus a point" filter for both df-limc and df-dv, but there are other ways to do that and at least insofar as division by zero in df-dv is the reason for the exclusion, you might consider using "apart from zero" instead. Do you have apartness spaces?

jkingdon:
> I'm pretty sure the if statement in df-limc is eliminable. Rather than reusing the "continuous at a point" predicate, you can just use one of the direct definitions like ellimc2.

I guess I'll try https://us.metamath.org/mpeuni/ellimc3.html because it would appear the only change needed is to change negated equality to apartness one place.

> You will have to grapple with the "neighborhood minus a point" filter for both df-limc and df-dv, but there are other ways to do that and at least insofar as division by zero in df-dv is the reason for the exclusion, you might consider using "apart from zero" instead.

Well, anything involving division is going to have to have an apart from zero in there somewhere.

> Do you have apartness spaces?

[...]

PR #3233, opened 2023-06-07
define limCC in iset.mm

jkingdon:
This is the first step of the work outlined in #3223 .  This might not be quite far enough to gain a lot of confidence that the definition here (based on https://us.metamath.org/mpeuni/ellimc3.html and using apart in place of not equals) will do everything we want it to do, but it does seem to work fine so far.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10025,6 +10025,24 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>cnfldtopn</td>
+  <td><i>none</i></td>
+  <td>if we use ` ( MetOpen `` ( abs o. - ) ) ` as our notation
+  for the topology of the complex numbers, this theorem is not
+  needed</td>
+</tr>
+
+<tr>
+  <td>cnfldtopon</td>
+  <td>~ cntoptopon</td>
+</tr>
+
+<tr>
+  <td>cnfldtop</td>
+  <td>~ cntoptop</td>
+</tr>
+
+<tr>
   <td>qdensere2</td>
   <td><i>none</i></td>
   <td>the set.mm proof depends on qdensere</td>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10105,6 +10105,12 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>rely on decidability of real number equality</td>
 </tr>
 
+<tr>
+  <td>ellimc3</td>
+  <td>~ ellimc3ap</td>
+  <td>changes not equal to apart</td>
+</tr>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10099,6 +10099,12 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>~ df-limced is adapted from ellimc3 in set.mm</td>
 </tr>
 
+<tr>
+  <td>limcvallem , limcfval , ellimc</td>
+  <td><i>none</i></td>
+  <td>rely on decidability of real number equality</td>
+</tr>
+
 </TABLE>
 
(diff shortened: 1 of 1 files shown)


wlammen:
I looked through this and did not find anything to complain about.  I know, more or less, what limits of functions are, but I feel a bit unfamiliar with the notations presented here, so I should not do the approval, I think.

iset.mm
```diff
@@ -136047,6 +136061,163 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
+
+  $( The limit operator. $)
+  climc $a class limCC $.
+
+  ${
+    $d d e f x y z $.
+    $( Define the set of limits of a complex function at a point.  Under normal
+       circumstances, this will be a singleton or empty, depending on whether
+       the limit exists.  (Contributed by Mario Carneiro, 24-Dec-2016.) $)
+    df-limced $a |- limCC = ( f e. ( CC ^pm CC ) , x e. CC |->
```
avekens, line 136101: This definition is very different from the one in set.mm, therefore "(Revised by Jim Kingdon, 3-Jun-2023.) " should be added.

iset.mm
```diff
@@ -136047,6 +136061,163 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
+
+  ${
+    $d A d e u x z $.  $d B d e u x z $.  $d F d e u z $.  $d d e ph u z $.
+    limccl.f $e |- ( ph -> F : A --> CC ) $.
+    limcdifap.a $e |- ( ph -> A C_ CC ) $.
+    $( It suffices to consider functions which are not defined at ` B ` to
+       define the limit of a function.  In particular, the value of the
+       original function ` F ` at ` B ` does not affect the limit of ` F ` .
+       (Contributed by Mario Carneiro, 25-Dec-2016.) $)
+    limcdifap $p |- ( ph ->
```
avekens, line 136184: Shpould also be marked with"Revised by...".

jkingdon:
> I looked through this and did not find anything to complain about. I know, more or less, what limits of functions are, but I feel a bit unfamiliar with the notations presented here, so I should not do the approval, I think.

Thanks for taking a look and reporting what you were able to find.

iset.mm
jkingdon: Agreed, it is pretty much a different definition. Should it just be credited to me, or take the credit from ellimc3 (which I guess in this case is mostly the same thing since df-limc is from the same author in the same year), also with a Revised line?

iset.mm
jkingdon: Because of changing not equal to apart, or because of the ` A C_ CC ` hypothesis? A lot of these attributions seem pretty wordy to me, especially if the "short description of the revision" mentioned in https://us.metamath.org/mpeuni/conventions-comments.html is taken to be something we always do.

I think there is some desire to keep these brief (as opposed to the full history that something went through) but I never know how far to take that.

iset.mm
avekens: I think adding "(Revised by Jim Kingdon, 3-Jun-2023.)" would be sufficient.

iset.mm
avekens: Also here, adding "(Revised by Jim Kingdon, 3-Jun-2023.)" would be sufficient.

iset.mm
jkingdon: Done.

iset.mm
jkingdon: Done. I wonder if we should make the conventions describe the short description as more of an option than something that looks like a requirement, though.

avekens (approved):

iset.mm
digama0: I don't think explaining revisions should be necessary, it's documentation not a changelog. The docs should reflect the current version of the definition and should be revised if they are out of date, but otherwise the acknowledgement of the change itself should be good enough.

jkingdon: I have made https://github.com/metamath/set.mm/pull/3240 where we can discuss what we want to say in conventions.

Issue #3250, opened 2023-06-12
Proposal: be able to declare forbidden axioms for a proof


jkingdon:
Many theorems have comments saying that they are proved without certain axioms, for example https://us.metamath.org/mpeuni/sbth.html says that it avoids https://us.metamath.org/mpeuni/ax-inf.html and https://us.metamath.org/mpeuni/ax-ac.html , and https://us.metamath.org/mpeuni/equsb1v.html says that it avoids https://us.metamath.org/mpeuni/ax-13.html

However, we have no way to check that such claims remain true, despite this being easy to automate in principle. In the past we have had multiple problems where axiom usage gets accidentally re-introduced and although our mechanisms to prevent this have gotten better, they aren't perfect and in some cases are more manual than needed.

Proposal:

* Add a `$j usage` directive, for example for these examples it would be:
```
$j usage 'sbth' avoids 'ax-inf' 'ax-ac';
$j usage 'equsb1v' avoids 'ax-13';
```

* Although this would probably generally be used for axioms, I had in mind saying that any usage of the specified statement, when traced back recursively, would be an error. This would be well defined for theorems as well as axioms. Although I don't have a concrete example of using it for theorems I am thinking of something like "this is an elementary proof which does not require Bézout's identity" (or something sorta like that).

* We would need a tool to check this. I assume it would look a bit like https://github.com/david-a-wheeler/metamath-knife/pull/107 (with various changes to parse the $j comments and so on, as well as presumably being a crate on top of the metamath-knife library crate since we are talking about breaking apart metamath-knife like that).

* We'd want the set.mm automated tests to download this tool and run it, which should be fairly similar to the existing checks.

Open questions:

* I'm not attached to this syntax, perhaps people have ideas about syntax
* Of course someone has to write the checker. I might be volunteering especially if the crate separation in metamath-knife is done first, but if someone else gets to it before me that's great too.

digama0:
SGTM :+1:

tirix:
This could also be useful to confirm that certain proofs do not require the axiom of choice, which we also have later on.
I don't expect the implementation to be very complicated, since we already have several bits and pieces for that.

The syntax seems sensible, so I think we can already add the `$j usage` tag in the database, and implement the check later.

jkingdon:
> I don't expect the implementation to be very complicated, since we already have several bits and pieces for that.

Perhaps someone who knows metamath-knife (if so, that wouldn't be me) will take it on.

[...]

PR #3251, opened 2023-06-13
Add $j usage

jkingdon:
This is intended to be an automatically enforced restriction on what statements can be used in a proof.  Document it in j-commands.html. Add a number of examples in iset.mm and set.mm.

Although the number of additions to set.mm may seem large, it is really just a sampling. I did not try to go through the whole file and there are definitely more places it can be added in the future.

See #3250 for more discussion.

### DIFF
--- a/mm-j-commands.html
+++ b/mm-j-commands.html
@@ -97,6 +97,14 @@ display might say that the theorem only depends on <code>A, B</code> even
 though it makes use of a theorem that depends on <code>A, C</code> which
 cannot be minimized in this way.</p>
 
+<code>$j usage 'THEOREM' avoids 'STATEMENT' ...;</code>
+
+<p>The proof of THEOREM should not use any listed STATEMENT,
+directly or indirectly.  Although STATEMENT will often be an axiom
+(for example the axiom of choice or the axiom of infinity), it can
+also be a theorem which this proof should avoid.  Tools are encouraged
+to check for violations and flag them as errors.</p>
+
 <code>$j type_conversions;</code>
 
 <p>Adds all "type conversions" to the grammatical parser. These are


wlammen (approved):
I will go through all the predicate logic theorems and supply missing axioms.  In many cases this was solved before by placing a theorem appropriately.  If it is listed before ax-13 for example, it cannot depend on it

set.mm
```diff
@@ -82806,6 +82898,7 @@ equivalent to the law of the excluded middle (LEM), and in ILE the LEM
       UAMZVKUBMZWBSZTZSZVJWBTZRZIZFUGZUBUAWEWLUHZUIWDUJVJWMTUIUKZCULWKEMZVJRZWD
       VKWEWOSZTZSZVJWOTZRZIFEWBWONZWCWPWJXAWBWOVJUMXBWHWSWIWTXBWGWRWDXBWFWQVKWB
       WOWEUNUOUPWBWOVJUQURQUSWNUTDULVDVAVBVC $.
+    $( $j usage 'sbth' avoids 'ax-inf' 'ax-inf2' 'ax-ac'; $)
```
tirix, line 82901: Interestingly, we have two axioms of choice: [`ax-ac`](https://us.metamath.org/mpeuni/ax-ac.html) and [`ax-ac2`](https://us.metamath.org/mpeuni/ax-ac2.html).
I suppose we need to exclude both everywhere?

jkingdon: > Interestingly, we have two axioms of choice: [`ax-ac`](https://us.metamath.org/mpeuni/ax-ac.html) and [`ax-ac2`](https://us.metamath.org/mpeuni/ax-ac2.html). I suppose we need to exclude both everywhere?

Good catch. It seems better to list them both, rather than leave it to convention in terms of "oh this is the one we use" versus "this is the one which is just hanging out there". I'll update this pull request to do that, probably later today.

jkingdon:
> I will go through all the predicate logic theorems and supply missing axioms.

Thanks! I got partway through the section but didn't make it to the end. And I'm sure you'd have a clearer idea than I would about what is worth calling out this way (my rule of thumb was that if axiom usage seemed important enough to mention it in the comments, then it probably deserves a `$j usage` directive).

> In many cases this was solved before by placing a theorem appropriately. If it is listed before ax-13 for example, it cannot depend on it

Yeah, I noticed that was true of many of the theorems in which case the new mechanism doesn't really provide a new check so much as document why things are ordered the way they are.

jkingdon:
Although this has been only open a day, I guess between the activity here and at #3250 it seems pretty well received. So I guess I'll merge it tomorrow. Not especially hard to change it later if people want to tweak this.

PR #3252, opened 2023-06-13
Move j-commands.html to mm-j-commands.html

digama0:
This is why it's not showing up on the website; since it is in the same folder as the theorem HTML pages it has to start with mm* to avoid clashing. I added an extra dash to avoid an implication of connection to mmj2 (which is not coincidental, but still not entirely helpful).


tirix (approved):

jkingdon:
This seems like a good idea (well, given our current organization anyway) but don't we need to add it to here: https://github.com/metamath/metamath-website-scripts/blob/59a1a55cc8163427279da7c3e0fb09d6f0470188/regenerate-website.sh#L98 if we want it to show up on the website?

digama0:
Oh, maybe so. I saw `mm*.html` in `install.sh` so it's possible both are necessary.

jkingdon:
> Oh, maybe so. I saw `mm*.html` in `install.sh` so it's possible both are necessary.

That would be my guess (I see `mm*.html` in `install.sh` too) but these scripts are complicated enough that I think I'd need to study them a lot more to have much confidence in that.

digama0:
It seems this still didn't stick, https://us.metamath.org/mpeuni/mm-j-commands.html is a 404.

david-a-wheeler:
Hmm, there seems to be a more general problem with generating the website. I would track it down now, but my dog died today (10yo, had for 9yrs) and I'm just not feeling like doing anything today.

PR #3165, opened 2023-05-09
cbvabv reduce axiom footprint

wlammen:
1. Remove ax-11 and ax-13 from cbvabv.  This was possible only after a disjoint variables condition was added on x and y.  This had minimal effects, since only one usage did not provide this condition, and this case is now based on cbvab.  On the other side, several theorems have their axiom list reduced.
2. drnfc1 does not need ax-11,  shorten proof by 1 byte.
3. shorten proof of drnfc2 by one byte.
4. nfabd and nfcvf do not need ax-ext, nfabd can also be proven without ax-9.
5. shorten nfabd2.
6. prove cleqf without ax-13.

I'd like to close #2930 since all undisputed shortenings are moved to develop.  Your opinions?

PR #3253, opened 2023-06-14
move section russel's paradox to main

icecream17:
cbvabvw is equivalent to cbvabv now (see #3165)
so likewise, change cbvrabv to match cbvrabvw and remove cbvrabvw

move rru to main
now that cbvrabvw --> cbvrabv, the only change is avoiding df-nel, which happened 3 days later
use rru in pwnss which is basically like minimize so no tag added

Note that for #3251 , the avoidance merely explains a change, a future revision can certainly add df-nel if there is a good reason.

---
rru could also be used for df-undef. it would save many axioms (compare undefval and ssrab2+ssex) and it could be used in df-pnf and df-mnf

such future work would then need to fix:
+ ~ ndfatafv2undef depends on the previous construction. presumably this would require changing df-afv2. makes me wonder whether having "Undef" be a prefix instead of a function would be better
+ ~ undefne0 would be invalid, but luckily no theorems use it

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+14-Jun-23 rru       [same]      moved from SN's mathbox to main set.mm
 10-Jun-23 sbtv      [same]      moved fron SN's mathbox to main set.mm
  8-Jun-23 rnmptc    [same]      moved from GS's mathbox to main set.mm
  8-Jun-23 eqneltri  [same]      moved from GS's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -14213,6 +14213,7 @@ New usage of "cbvabvOLD" is discouraged (0 uses).
 New usage of "cbveuALT" is discouraged (0 uses).
 New usage of "cbvexsv" is discouraged (2 uses).
 New usage of "cbvmoOLD" is discouraged (0 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -29915,11 +29915,25 @@ choice between (what we call) a "definitional form" where the shorter
...
-    $d x A $.  $d y A $.  $d y ph $.  $d x ps $.
+    $d x y A $.  $d y ph $.  $d x ps $.
     cbvrabv.1 $e |- ( x = y -> ( ph <-> ps ) ) $.
     $( Rule to change the bound variable in a restricted class abstraction,
-       using implicit substitution.  (Contributed by NM, 26-May-1999.) $)
+       using implicit substitution.  (Contributed by NM, 26-May-1999.)  Require
+       ` x ` , ` y ` be disjoint to avoid ~ ax-11 and ~ ax-13 .  (Revised by
+       Steven Nguyen, 4-Dec-2022.) $)
     cbvrabv $p |- { x e. A | ph } = { y e. A | ps } $=
+      ( cv wcel wa cab crab weq eleq1w anbi12d cbvabv df-rab 3eqtr4i ) CGEHZAIZ
```
wlammen, line 29925: From about 200 applications of this theorem not a single one complains about the extra dv condition.  Interesting

wlammen (approved):

tirix (approved):

avekens (approved):

PR #3260, opened 2023-06-18
Improvement of ~bhmafibid* and ~2sq*

avekens:
Generalization of Brahmagupta-Fibonacci identities and variants of "all primes 4n+1 are the sum of two squares" (as proposed by Gérard Lang)

* ~ssrmo(f), ~rmoimi, ~2reu2rex moved from mathboxes to main
* ~zle0orge1, ~2mulprm added to main
* Brahmagupta-Fibonacci identities for complex numbers: ~bhmafibid1cn, ~ bhmafibid2cn
* All primes of the form ` 4 k + 1 ` are sums of squares of two positive integers: ~2sqnn
* The decomposition of a prime of the form ` 4 k + 1 ` as a sum of squares of two (different) positive integers is unique: ~2sqreult, ~2sqreultb, ~2sqreunn, ~2sqreunnlt, ~2sqreunnltb

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,20 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+25-Jun-23 2reu4     [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2reu4a    [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2reu2     [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2reu1     [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2rexreu   [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2rmoswap  [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2reurmo   [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2reurex   [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 reuan     [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 rmoanim   [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 reuimrmo  [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 2reu5a    [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 raaan2    [same]      moved from AV's mathbox to main set.mm
+25-Jun-23 csbconstgi [same]     moved from GM's mathbox to main set.mm
 18-Jun-23 ssrmo     ssrmof      moved from TA's mathbox to main set.mm
 18-Jun-23 rmoimi    [same]      moved from AV's mathbox to main set.mm
 18-Jun-23 2reu2rex  [same]      moved from AV's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+18-Jun-23 ssrmo     ssrmof      moved from TA's mathbox to main set.mm
+18-Jun-23 rmoimi    [same]      moved from AV's mathbox to main set.mm
+18-Jun-23 2reu2rex  [same]      moved from AV's mathbox to main set.mm
 14-Jun-23 rru       [same]      moved from SN's mathbox to main set.mm
 10-Jun-23 sbtv      [same]      moved fron SN's mathbox to main set.mm
  8-Jun-23 rnmptc    [same]      moved from GS's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -18333,6 +18333,7 @@ Proof modification of "bamalipOLD" is discouraged (26 steps).
 Proof modification of "barbariALT" is discouraged (22 steps).
 Proof modification of "barocoALT" is discouraged (24 steps).
 Proof modification of "bcsiALT" is discouraged (444 steps).
+Proof modification of "bhmafibid1" is discouraged (429 steps).
 Proof modification of "biadan2OLD" is discouraged (17 steps).
 Proof modification of "biadanOLD" is discouraged (121 steps).
 Proof modification of "biadaniALT" is discouraged (28 steps).


set.mm
```diff
@@ -124783,6 +124802,14 @@ distinguish between finite and infinite sets (and therefore if the set size
     UCUHUJUCUJUEUGUDLZLZUHUCUIUKUEUCUKUGUFEFZLUIUCUDUMUGUCAANORUFPQRUHUEUGUDIUL
     UDUEUGSUEUGUDUATQUBT $.
 
+  $( An integer is either less than or equal to ` 0 ` or greater than or equal
```
wlammen, line 124805: Alternative comment: There is no integer in the open unit interval.

set.mm
```diff
@@ -373209,6 +373328,139 @@ minus one divided by eight (` ( 2 /L N ) ` = -1^(((N^2)-1)/8) ).
...
+      YSYQJZYOYGYEYTXLYEXPUIEZYTYEXMXMUNUJUUAYQYSXPUKULUOUMUPUQURYFYRYNSZYOYGXI
+      XKYEUUBYRXKYEXIYNXKYEXIYNSSSAYQAYQJZXKXIYEYNUUCXKXIYEYNSZUUCXKXIKYQFGHZIJ
+      ZYQDEZKZUUDUUCXKUUFXIUUGUUCXJUUEIAYQFGUFURAYQDUSUTYEUUHYNYEUUGUUFYNYEUUGX
+      PIJZUUFYNSYEXPVFEZUUGUUIRYEXMVFEMVAEUUJXMVBVCXMMVGVDXPVEUOUUIUUFMIJZYNUUI
+      UUEMIUUIUUEMFGHZMUUIYQMFGUUIYQMITHMXPIMTVHVIVJVKMVTEFVLEZVMMUBLMFUCLUULMJ
+      VNFPEUUMVOFVPVQVRWAMFVSWBVJURIMUEZUUKYNSWCUUKIMJZUUNYNMIWDUUOUUNYNYNIMWEW
+      FWGVQWHWHWIWFWHWJWLWKWMWNUPWOWPYKYJXOWQWRWSYHYIYMRXSYHYBYLXOYFXMVTEZXNVTE
+      YBYLRYGYEUUPXLXMWTUMXNWTXMXNXCXAXBUHXDXEXFXGXGXH $.
+
+    $( There exists a unique decomposition of an odd prime as a sum of squares
```
wlammen, line 373450: Why is 2 explicitly ruled out here?  It seems that the condition 4k + 1 is sufficient.

avekens:
@wlammen You asked for ~2sqreunnltb "Why is 2 explicitly ruled out here? It seems that the condition 4k + 1 is sufficient.", but then you deleted this comment? But you are absolutely right, 2 needs not be excluded here! I will revise my PR accordingly...

wlammen:
I did not know whether this was meant as a lemma for later proofs, and was hesitant to interfere with your ideas.  But you are right, I should have left the comment in place.  The simple answer could have been: "I need it exactly this way in later applications."

tirix (approved):
Nice!

set.mm
```diff
@@ -373209,6 +373360,149 @@ minus one divided by eight (` ( 2 /L N ) ` = -1^(((N^2)-1)/8) ).
...
+      ZWTWIXJMADWRVGEZWTADWRVHXKWQXCVFZWJADLVIXKWQXLWJPXKWQMZXCWJXMAWNHZCSQZBSQ
+      ZXCWJVJZWQXPXKNSVKZWQXOBNQZXPVLWQXBXNCNQZBNQXSXDXAXTBNWPXNCNWOXNWMWOWNAWO
+      VMVNRTTXTXOBNXRXTXOPVLXNCNSVOVPTVRXOBNSVOVQRXMWIXPXQUJXKWIWQADWRVSUTBCAVT
+      WAWBWCOWDWEWFWGWH $.
+
+    $( There exists a unique decomposition of a prime of the form ` 4 k + 1 `
+       as a sum of squares of two positive integers.  See ~ 2sqnn for the
+       existence of such a decomposition.  (Contributed by AV, 11-Jun-2023.) $)
+    2sqreunn $p |- ( ( P e. Prime /\ ( P mod 4 ) = 1 )
+                     -> E! a e. NN E! b e. NN ( a <_ b
```
wlammen, line 373414: The double E! operator is difficult, since there might be another a  such that even two or more b exist with the given property, see the comment to df-eu.  I think you meant something closer to the antecedent in 2exeu.  This holds likewise for 2sqreult, 2sqreunnlt, 2sqreunnltb.

wlammen (changes requested):

set.mm
avekens: @wlammen You are absolutely right: The comment does not reflect the theorem, and the pattern "`E! x e. A E! y e. B ph`" cannot be used to express "there exists exactly one ` x ` and one ` y ` such that ` ph ` " (as mentioned in the comment of df-eu). Furthermore, there is no real usage of such a pattern (as well as "E! x E! y ph") in set.mm: only theorems about existential uniqueness itself contain it. Therefore, a different approach must be followed here (as you proposed already).

To use ordered pairs (` E! p e. ( NN X. NN )` ) is possible in this case, because we can assume that it is unique ( by claiming ``( 1st p )` <_ (2nd ` p )``). Alternatively, (proper) unordered pairs can be used: `` E! p e ~P NN ( ( # ` p ) = 2 /\ ... )``, or, using the definition of proper pairs in my mathbox:  ``E! p e. ( PrPairs ` NN ) ... ``. I used this already showing that a line passing through a point within a circle around the origin intersects the circle at exactly two different points (see ~inlinecirc02preu.

I will revise the corresponding theorems accordingly (although, the pattern "` E! x e. A E! y e. B ph `" looks nicer), prefering the ordered pair variant.

tirix: Good catch!

If we need this a lot, maybe a new syntax like `E!! x e. A , y e. B ph` could be of use, which could be defined as in `~2exeu`.
Otherwise, from a pair point of view, something like `E! p e. ( NN X. NN ) ( p = <. a , b >. /\ ph )` would be more readable, especially if `a` and `b` repeat a lot in `ph`.

wlammen: A more general approach to uniqueness takes tuples from a multi-dimensional space (with possibly differently typed dimensions, like one being the real line, the  second a finite set, and so on), and claims there is just one with a given property.  Perhaps you can take this into account when deciding on the type of a pair.

avekens:
Before I am going to revise ~2sqreunn, ~2sqreult, ~2sqreunnlt, ~2sqreunnltb to hold for unique pairs of integers, I made a study to illustrate the  strange behavior of double E! operator with a simple, but real numeric example. For details see comments for the new theorems ~addsq2reu, ~addsqrexnreu, ~addsqnreup.

set.mm
```diff
@@ -38055,6 +38055,35 @@ among classes ( ~ eq0 , as opposed to the weaker uniqueness among sets,
       UNABUJTVBUNSSHSVBULSUMSABCTABDTMSUEUFUGUH $.
   $}
 
+  ${
+    $d A x y $.  $d B x y $.  $d X x y $.  $d ph y $.  $d ch x $.  $d ps x y $.
```
wlammen, line 38059: Even if it is not necessary for the proof, I'd mark the dummy variable y as disjoint from ch.

set.mm
```diff
@@ -38036,6 +38055,35 @@ among classes ( ~ eq0 , as opposed to the weaker uniqueness among sets,
...
+    2nreu.a $e |- ( x = A -> ( ph <-> ps ) ) $.
+    2nreu.b $e |- ( x = B -> ( ph <-> ch ) ) $.
+    $( If there are two different sets fulfilling a wwf (by implicit
+       substitution), then there is no unique set fulfilling the wwf.
```
icecream17, line 38063: wwf --> wff

avekens:
The theorems 2sqreu* with conclusions of the form `-> E! x e. A E! y e. A ph` are revised now to have the proper conclusions of the form  `-> ( E! x e. A E. y e. A ph /\ E! y e. A E. x e. A ph )`, actually meaning " "there exists exactly one x and one y such that ph ".
I could reuse the former theorems as lemmata, by showing  `A. a e. NN0 E* b e. NN0 ( ps /\ ( ( a ^ 2 ) + ( b ^ 2 ) ) = P ) ` (see ~ 2sqreulem4) and combining it with the lemmata by ~2reu1 to obtain the final theorems. 
I could use some theorems I added to my mathbox a long time ago to show this.

Unfortunately, it is not so easy to prove the "ordered pair" variants of the theorems, which I first intended. Although it was easy to show that "there exists exactly one x and one y such that ph " follows from the "ordered pair" variant (see ~ opreu2reurex1), the other direction is not easily shown (or even is maybe not true in general? Then the "ordered pair" variant would be a stronger result).  

Althougth the revised theorems are correct now, they do not look "nice" (the inner wff of the quantifiers must be duplicated - here an explicit definition as proposed by @tirix would help), and I will try to convert them in the "ordered pair" form somehow. This should be done, however, in a sepatae PR, so I would propose to merge this PR before I continue.

set.mm
```diff
@@ -373161,14 +373619,169 @@ minus one divided by eight (` ( 2 /L N ) ` = -1^(((N^2)-1)/8) ).
...
+      cn negeu sq1 addcomd adantl mpbird neg1sqe1 jca oveq1 reubidv 2nreu mpsyl
+      id ) DEFZDUAZEFZDVFUBZUCAEFZBGZDHIJZKJZALZBEMZVJVFHIJZKJZALZBEMZSVJCGZHIJ
+      ZKJZALZBEMZCEMUDVEVGVHUEUFDULFVHUGDUHUIUJVIVNVRVIVNDVJKJZALZBEMZVEVIWFVIN
+      BDAUMUKZVIVMWEBEVIVJEFZSZVLWDAWIVLVJDKJZWDWIVKDVJKVKDLWIUNOPWHWJWDLVIWHVJ
+      DWHVDWHNUOUPZQTRUQVIVRWFWGVIVQWEBEWIVPWDAWIVPWJWDWIVODVJKVODLWIUROPWKQTRU
+      QUSWCVNVRCDVFEVSDLZWBVMBEWLWAVLAWLVTVKVJKVSDHIUTPTVAVSVFLZWBVQBEWMWAVPAWM
+      VTVOVJKVSVFHIUTPTVAVBVC $.
+
+    $( For each complex number, there exists a complex number to which the
+       square of more than one (ore no) other complex numbers can be added to
```
wlammen, line 373697: typo ore->or

set.mm
```diff
@@ -373161,14 +373619,169 @@ minus one divided by eight (` ( 2 /L N ) ` = -1^(((N^2)-1)/8) ).
...
+       complex numbers can be added to yield the given complex number (see
+       ~ addsqrexnreu ).  For example, ` <. 1 , ( sqrt `` ( C - 1 ) ) >. ` and
+       ` <. 1 , -u ( sqrt `` ( C - 1 ) ) >. ` are two different decompositions
+       of ` C ` (if ` C =/= 1 ` ).  Therefore, there is no unique decomposition
+       of any complex number as a sum of a complex number and a square of a
+       complex number, as generally proved by this theorem.
+
+       As a consequence, a theorem must claim the existence of a unique pair of
+       sets to express "There are unique ` a ` and ` b ` so that .."  (more
+       formally ` E! p e. ( A X. A ) ph ` with ` p = <. a , b >. ` ), or by
```
wlammen, line 373744: Shouldn't it be A X. B?

wlammen (changes requested):

set.mm
```diff
@@ -373630,50 +373916,159 @@ ordered sets (like ` RR ` or ` Prime ` ).  Alternatively, (proper)
...
+
+  ${
+    $d P a b c $.
+    2sqreu.1 $e |- ( ph <-> ( a <_ b /\ ( ( a ^ 2 ) + ( b ^ 2 ) ) = P ) ) $.
+    $( There exists a unique decomposition of a prime of the form ` 4 k + 1 `
+       as a sum of squares of two nonnegative integers.  See ~ 2sqnn0 for the
+       existence of such a decomposition.  (Contributed by AV, 4-Jun-2023.)
+       (Revised by AV, 25-Jun-2023.) $)
+    2sqreu $p |- ( ( P e. Prime /\ ( P mod 4 ) = 1 )
+           -> ( E! a e. NN0 E. b e. NN0 ph /\ E! b e. NN0 E. a e. NN0 ph ) ) $=
```
icecream17, line 374007: perhaps if the hypothesis didn't require `a <_ b` then `E! a e. NN0 E. b e. NN0 ph` would suffice, since addition is commutative so you could do

```
2sqreu
  ( ( P e. Prime /\ ( P mod 4 ) = 1 ) -> E! a e. NN0 E. b e. NN0 ph )
2sqreu
  ( ( P e. Prime /\ ( P mod 4 ) = 1 ) -> E! b e. NN0 E. a e. NN0 ph )
```

set.mm
```diff
@@ -38036,6 +38171,35 @@ among classes ( ~ eq0 , as opposed to the weaker uniqueness among sets,
       UNABUJTVBUNSSHSVBULSUMSABCTABDTMSUEUFUGUH $.
   $}
 
+  ${
+    $d A x y $.  $d B x y $.  $d X x y $.  $d ph y $.  $d ch x $.  $d ps x y $.
```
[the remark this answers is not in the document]
avekens: OK, if that's the reason, I can/will change this.

set.mm
```diff
@@ -38586,6 +38750,64 @@ disjoint classes (see ~ disjdif ).  Part of proof of Corollary 6K of
...
+      ( c0 wceq wb wa wn wo wral rzal adantr wne df-ne sylbir dfbi3 adantl nfcv
+      pm5.1 syl12anc r19.28z ralbidv nfral r19.27z sylan9bbr jaoi sylbi ) EIJZF
+      IJZKUMUNLZUMMZUNMZLZNABLDFOZCEOZACEOZBDFOZLZKZUMUNUAUOVDURUOUTVAVBVDUMUTU
+      NUSCEPQUMVAUNACEPQUNVBUMBDFPUBUTVCUDUEUQUTAVBLZCEOZUPVCUQFIRZUTVFKFISVGUS
+      VECEABDFGUFUGTUPEIRVFVCKEISAVBCEBCDFCFUCHUHUITUJUKUL $.
+  $}
+
+  ${
+    $d z w ph $.  $d w x y z A $.  $d w x y z B $.
+    $( Definition of double restricted existential uniqueness ("exactly one
```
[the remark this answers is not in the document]
avekens: OK, I will change the label accordingly.

set.mm
avekens: Yes, the idiom "( E! a e. NN0 E. b e. NN0 ph /\ E! b e. NN0 E. a e. NN0 ph )" represents "there are unique a and b ...", and that is what is shown here.

wlammen (approved):

icecream17 (approved):
no reason to not approve, lgtm

(still I wonder whether there's a better way to prove double uniqueness, like maybe `E! x ( x = <. y , z >. /\ ph )`)

wlammen:
In my introductory book to logic a whole chapter is dedicated to this topic.  This is less simple as it appears on the outside.  The tuple form has some nasty dependencies on multiple cross products of spaces, and needs knowledge about isomorphism of such products when shuffling the dimensions, or using subproducts.  In the end you don't want to see your theorems depend on the particular choice of how your space of objects is designed ( R ^ 3 isomorph R ^ 2 X. R isomorph R X. R ^ 2, as far as uniqueness is concerned, at least).

PR #3261, opened 2023-06-19
Intuitionize the next part of the limCC section

jkingdon:
This is basically a first pass - some things are not intuitionized that eventually we will need to. But some things are intuitionized here.

This covers `cnplimc` through `limccnp`.

There are also a few theorems from earlier sections, mostly copied directly from set.mm but in any case basically the same sort of thing that we already have in set.mm and existing iset.mm theorems.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10167,6 +10167,13 @@ intuitionistic and it is lightly used in set.mm</TD>
   working within ` C ` ).</td>
 </tr>
 
+<tr>
+  <td>cnplimc</td>
+  <td>~ cnplimcim</td>
+  <td>The converse is conjectured to also be provable, but
+  would require a more involved proof.</td>
+</tr>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10107,6 +10107,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </TR>
 
 <tr>
+  <td>cncfcn</td>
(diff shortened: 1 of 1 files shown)


tirix (approved):
LGTM

iset.mm
```diff
@@ -136322,6 +136322,21 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
   $}
 
+  ${
+    $d x A $.  $d x B $.  $d x F $.
+    cnlimci.f $e |- ( ph -> F e. ( A -cn-> D ) ) $.
+    cnlimci.c $e |- ( ph -> B e. A ) $.
+    $( If ` F ` is a continuous function, then the limit of the function at any
+       point equals its value.  (Contributed by Mario Carneiro,
+       28-Dec-2016.) $)
+    cnlimci $p |- ( ph -> ( F ` B ) e. ( F limCC B ) ) $=
```
wlammen, line 136332: Shouldn't this be named according to other deduction style theorems, ending with a d?

jkingdon: If there is a corresponding inference or closed form theorem, sure. But there isn't a rule that all deduction style theorems end in `d`, all implication style theorems end in `i`, all closed form theorem end in `t`, all theorems with a not-free hypothesis end in `f`, etc.

I don't know if you are looking a the `i` at the end, but the `i` suffix here means "implication" ("implication (suffix)" entry at https://us.metamath.org/mpeuni/conventions-labels.html ). https://us.metamath.org/mpeuni/cnlimc.html is a biconditional and https://us.metamath.org/mpeuni/cnlimci.html is one direction of that biconditional.

PR #3264, opened 2023-06-21
Swap idi and a1ii within set.mm, iset.mm, and nf.mm

david-a-wheeler:
Theorem `idi` is conceptually more primitive than `a1ii`, so it should be first.  Theorem `idi` is literally just a copy, while `a1ii` is a selection of a copy.

It doesn't *matter* from the point-of-view of truth, but when you're explaining each assertion in sequence, it's easier to explain this pair of theorems if `idi` comes first.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -193,6 +193,7 @@ New usage of "19.41h" is discouraged (5 uses).
 New usage of "19.42h" is discouraged (1 uses).
 New usage of "19.9hd" is discouraged (2 uses).
 New usage of "4syl" is discouraged (15 uses).
+New usage of "a1ii" is discouraged (0 uses).
 New usage of "a9evsep" is discouraged (1 uses).
 New usage of "addmodlteqALT" is discouraged (0 uses).
 New usage of "alrimih" is discouraged (25 uses).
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
@digama0 @jkingdon -

I know it's weird, but in set.mm I'd like to swap `idi` and `a1ii`. I saw no reason for iset.mm and set.mm to be different in this case, so with this commit they both would have them in the same order.

It doesn't *matter* from the point-of-view of truth, but when you're explaining each assertion in sequence, it's easier to explain this pair of theorems if `idi` comes first.

So I want to swap them to make the database easier to explain from the beginninng. Theorem `idi` is conceptually more primitive than `a1ii`, so it should be first. Theorem `idi` is literally just a copy, while `a1ii` is a selection of a copy.

Seem reasonable? As far as I can tell *all* provers & proof assistants refer to theorems by name, not number, so this shouldn't affect anything.

david-a-wheeler:
@tirix - any thoughts?

GinoGiotto:
`idi`  and `a1ii` are present in `nf.mm` as well.

david-a-wheeler:
@GinoGiotto - great point. I'll update this to add nf.mm.

david-a-wheeler:
@GinoGiotto - all okay now?

GinoGiotto (approved):

GinoGiotto:
For me it's all good. In general I don't feel competent enough to do formal reviews, so I would recommend waiting for a second opinion.

tirix (approved):

tirix:
> @tirix - any thoughts?

I agree it's better to have a consistent order between the different databases for `a1ii` and `idi`. In any case none of those should appear in a final proof, so this has no impact on the rest of the databases.
You have my go!

wlammen (approved):

david-a-wheeler:
Thank you everyone!

jkingdon:
> I saw no reason for iset.mm and set.mm to be different in this case, so with this commit they both would have them in the same order.

Yup, good thinking.

There are cases where it is hard to maintain the same order, but generally we try to keep the sections and the order the same as we can.

PR #3265, opened 2023-06-23
Point out a special interpretation of df-ral

wlammen:
1. df-ral allows an interpretation that a hasty viewer might oversee.  Given how the misinterpretation of a double unique quantifier was clearified, I think a note should be added to df-ral as well.  I do not update df-rex etc. to first see how this idea is received.
2. Remove stuff from my mathbox, that I am not going to support further
3. delete outdated OLD theorems


avekens:
> 1. df-ral allows an interpretation that a hasty viewer might oversee.  Given how the misinterpretation of a double unique quantifier was clearified, I think a note should be added to df-ral as well.  I do not update df-rex etc. to first see how this idea is received.

I think it is a good idea to add such an explanation. We should, however, also add some examples. It can be easily shown that `A. x e. x _V e. x` ( A replaced by x) or `A. x e. ~P x x e. _V` (A replaced by `~P x `). But are there meaningful examples, showing a real application of having A depending on x? Otherwise, why don't we enforce A and x to be disjoint by definition? Maybe there was a discussion about this somewhere else...

wlammen:
For easy reading, this is the comment of df-ral:

>   Define restricted universal quantification.  Special case of Definition
     4.15(3) of [TakeutiZaring] p. 22.

     Note:  This notation is most often used to express that ` ph ` holds for
     all elements of a given class ` A ` .  For this ` F/_ x A ` is required,
     though, for example asserted when ` x ` and ` A ` are disjoint.  Should
     instead ` A ` depend on ` x ` , you rather focus on those ` x ` that
     happen to be contained in the corresponding ` A ( x ) ` .  This is rarely
     meant, so be careful to either keep ` A ` independent of ` x ` , or adjust
     your comments to include this special case.

> ...why don't we enforce A and x to be disjoint by definition?

You may still see applications with bounded x like A. x x e. { x | ph }, where 
disjointness is too limiting.

> Maybe there was a discussion about this somewhere else...

df-wl-ral, that is going to be removed from my Mathbox in this PR,  uses `A. y ( y e. A -> A. x ( x = y -> ph ) ) ` , same as `A. y ( y e. A -> [ y / x ] ph ` as a definition of df-ral instead.

See also discussions in #3173 , #3178, #3192, #3198

> But are there meaningful examples...

I looked for one, but did not find an instance.  This does not mean it is never used that way, the usage of restricted quantifiers is too widespread for a quick check.

While it is ok to reopen the discussion here, for the moment, I only want to update the comment.

avekens:
> > ...why don't we enforce A and x to be disjoint by definition?
> 
> You may still see applications with bounded x like A. x x e. { x | ph }, where disjointness is too limiting.

OK, then it could be enforced that A does not contain x as free variable...

> > Maybe there was a discussion about this somewhere else...
> See also discussions in #3173 , #3178, #3192, #3198

Thanks, I will have a look there...

> > But are there meaningful examples...
> 
> I looked for one, but did not find an instance. This does not mean it is never used that way, the usage of restricted quantifiers is too widespread for a quick check.
> 
> While it is ok to reopen the discussion here, for the moment, I only want to update the comment.

I agree, no discussion here (again). But examples would be helpful to understand the comment. Maybe one of my proposed examples could be added (to show at least that there are valid, i.e., provable, theorems with x free in A).

wlammen:
I added an example for the exotic case now.

>   
     Note:  This notation is most often used to express that ` ph ` holds for
     all elements of a given class ` A ` .  For this reading ` F/_ x A ` is
     required, though, for example, asserted when ` x ` and ` A ` are disjoint.

     Should instead ` A ` depend on ` x ` , you rather focus on those ` x `
     that happen to be contained in the corresponding ` A ( x ) ` .  This
     hardly used interpretation could still occur naturally.  For example, look
     at the solutions ` A ( n ) ` in ` n ` to ` n ^ n - n ` ! - 7 = ` p x. n `
     for any ` p e. NN ` , and wonder when ` n ` itself belongs to it.  All
     such ` n ` would be selected by ` A. n e. A ` ... .

     So be careful to either keep ` A ` independent of ` x ` , or adjust your
     comments to include such exotic cases.

avekens (approved):
Although I would have liked a real metamath theorem as example, the sketched example should be enough for the moment. 
I do not understand, however, why you remove the material from your mathbox. Although I am not very deep in this matter, your work looks interesting. It may serve as inspirations for others.

Maybe we should hear one or two other opinions about this.

wlammen:
The example looks in fact very construed (in fact, it is!), so I won't object if something more down to the core replaces it.  Since this would be a simple comment upgrade, we can leave this for the future as a TODO.

PR #3269, opened 2023-06-28
Minimized footex

GinoGiotto:
I managed to minimize [footex](https://us.metamath.org/mpeuni/footex.html), the only theorem that I had to skip in my recent [global minimization process](https://github.com/metamath/set.mm/pull/3127), due to the excessively long time it demanded.

[Footex](https://us.metamath.org/mpeuni/footex.html) is a nasty theorem, even basic commands like `prove footex` or `show proof footex /compressed` are time-challenging. I believe it needs revision because I find it to be impractical to work with (ping @tirix since he contributed it).

The first time I attempted to minimize it, I used metamath-exe `version 0.198`, and as stated in #3180, the program crashed. This time I used `version 0.199.pre` and successfully got this PR as a result. I didn't replicate these observations, so I'm not sure if it's actually the new version that made the difference.

A note that I find interesting: `footex` has neither the longest proof nor highest number of steps in main, despite this it takes days to minimize, which is by far the longest duration I encountered in my global minimization activity. For comparison, the most challenging theorems need at most some hours to minimize, not even close to the time required by `footex`.

Since the proof of `footex` isn't that long, it seems to me that proof length isn't the only factor that matters to determine the execution time to shorten it. So I believe that the addition of a `/SKIPLONGPROOF` option proposed in https://github.com/metamath/metamath-exe/issues/95 might not be the most effective solution for avoiding such significant time requirements, and I agree with the recent decision to close it.

I still wonder tho, why is `footex` so bad? Perhaps it's related to the presence of numerous distinct variable conditions? I really don't know, but I believe it would be interesting to get a better undertanding of it in order to avoid such time-draining theorems in the future.


wlammen (approved):

icecream17:
perhaps it's the `ad*ant*r`/`simp*r` chains
it reminds me of the comment of 4syl

tirix:
Thank you for dedicating time and computing power for this!

I don't know exactly how the minimizer works so these are wild guesses, but it could be the length (and complexity) of the steps statements. It might also be the high number of hypotheses used in Geometry, but then other geometry and algebra theorems should have the same issue.

david-a-wheeler (approved):

david-a-wheeler:
@GinoGiotto - I applaud your perseverence!

I have no idea why it's so hard to minimize this case.

tirix (approved):

digama0:
@tirix One thing that I would recommend for this proof, just as a general readability thing which may help with the minimize (based only on pure speculation) is to split the core of the proof out into a lemma, basically once all the assumptions have been established and you are working within some local context. See for example [pnt3](https://us.metamath.org/mpeuni/pnt3.html), which calls [pntleml](https://us.metamath.org/mpeuni/pntleml.html) after adding a bunch of hypotheses, which calls [pntlemp](https://us.metamath.org/mpeuni/pntlemp.html) after adding hypotheses, which calls [pntleme](https://us.metamath.org/mpeuni/pntleme.html) after adding even more hypotheses. Doing this keeps the statements somewhat readable by hiding the context behind the variable `ph`, and it is possible that the large amount of internal structure in all the proof steps is related to why minimize has a hard time with this theorem.

GinoGiotto:
I opened an issue here https://github.com/metamath/set.mm/issues/3271 to discuss this more thoroughly.

PR #3272, opened 2023-06-29
shorten ralanid r19.29r

wlammen:
1. shorten proof of r19.29r by 1 byte. This also reduces the symbol count in the proof display
2. shorten r19.29vva further.  No OLD theorem created, since it is a follow-up of the last shortening less than 24 h ago.
3. shorten ralanid.  To this end some elementary restricted quantifier theorems had to move closer to the definition df-ral.
4. remove an outdated OLD theorem

Currently the following command from the metamath.exe check

> sudo apt-get install texlive-latex-extra texlive-extra-utils texlive-fonts-extra texlive-science

fails.  This is not related to the changes in this pull request, but rather a problem on the checking system.  The tests should be rerun again a few hours later.

Seems the problem with TEX repositories were temporary and are gone now.

=========================================================================

I am restructuring the section restricted quantifier, so you can better see dependencies, and spot possible shortenings.  As far as I can see, there are 4 theorems using df-ral directly, without interference from the other three.  These are alral, raln, ral2imi and ralimi2, expressing primitive properties of the term ` A.x ( x e. A -> ph) ` (Note this is NOT always expressing an iteration over the elements of a given set A).  Other theorems using restricted quantifiers can be based on these four, so they seem less basic.

dfrex2 could be the definition of df-rex.  This would let df-rex appear as a convenience modification of df-ral, instead of some new definition that later turns out to be tightly related to df-ral.  raln is here the casting theorem between the old and new form.

What do you think? Let me know.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17005,7 +17005,6 @@ New usage of "phop" is discouraged (1 uses).
 New usage of "phpar" is discouraged (2 uses).
 New usage of "phpar2" is discouraged (3 uses).
 New usage of "phrel" is discouraged (1 uses).
-New usage of "pilem3OLD" is discouraged (0 uses).
 New usage of "pinn" is discouraged (17 uses).
 New usage of "pinq" is discouraged (3 uses).
 New usage of "pion" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


tirix (approved):

tirix:
> dfrex2 could be the definition of df-rex. This would let df-rex appear as a convenience modification of df-ral, instead of some new definition that later turns out to be tightly related to df-ral. raln is here the casting theorem between the old and new form.

This is probably something we should first run through the mailing list. I see the advantage, namely, that `rex` would directly be based on `ral`, and would directly benefit of any improvement/evolution, but on the other hand we want to stay close to textbook definitions.

PR #3273, opened 2023-06-29
Intuitionize derivative theorems from df-dv to recnprss

jkingdon:
This portion intuitionizes with various obvious changes (not equal versus apart and others).

It is a good stopping point because after this is https://us.metamath.org/mpeuni/dvfg.html and the set.mm proof of that relies on a bunch of things we don't have yet, which are ultimately defined in set.mm via closed sets. I don't yet have much of a plan for that but whatever I find, it isn't happening right away.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10241,6 +10241,13 @@ intuitionistic and it is lightly used in set.mm</TD>
   topology on the complex numbers</td>
 </tr>
 
+<tr>
+  <td>eldv</td>
+  <td>~ eldvap</td>
+  <td>changes not equal to apart and changes the notation for the
+  topology on the complex numbers</td>
+</tr>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10135,6 +10135,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>reldv</td>
+  <td>~ reldvg</td>
+</tr>
+
+<tr>
   <td>limcvallem , limcfval , ellimc</td>
   <td><i>none</i></td>
   <td>rely on decidability of real number equality</td>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10228,6 +10228,12 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>The set.mm proof relies on limciun .</td>
 </tr>
 
+<tr>
+  <td>dvlem</td>
+  <td>~ dvlemap</td>
+  <td>not equal is changed to apart</td>
+</tr>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT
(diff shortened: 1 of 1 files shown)


wlammen (approved):
I am not always familiar with the details of the notations, but all in all it looks good to me.

Do you have in mind that a derivative exists outside of CC as well, like in differential geometry?

jkingdon:
> Do you have in mind that a derivative exists outside of CC as well, like in differential geometry?

I suppose so, but:

1. I think some multivariate definitions can be defined in terms of the real derivative of one variable (I don't seem to be finding quick answers either in my memory or in the first things I found in internet searches, though),

and perhaps more importantly,

2. I don't notice much development of this in set.mm, at least the parts I looked at.

To a large extent my own focus is to be able to intuitionize https://us.metamath.org/mpeuni/pire.html which at least following the development in set.mm involves the intermediate value theorem and various results involving continuity (basically, that a differentiable function is continuous and the exponential function is differentiable).

iset.mm
```diff
@@ -136192,6 +136204,25 @@ S C_ ( P ( ball ` D ) T ) ) $=
           -> ( abs ` ( ( f ` z ) - y ) ) < e ) ) ) } ) $.
   $}
 
+  ${
+    $d s f x z w $.
+    $( Define the derivative operator on functions on the reals.  This acts on
```
avekens, line 136209: This definition is not only "on the reals".

iset.mm
```diff
@@ -136192,6 +136204,25 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
+       these points.  The set ` s ` here is the ambient topological space under
+       which we are evaluating the continuity of the difference quotient.
+       Although the definition is valid for any subset of ` CC ` and is
+       well-behaved when ` s ` contains no isolated points, we will restrict
+       our attention to the cases ` s = RR ` or ` s = CC ` for the majority of
+       the development, these corresponding respectively to real and complex
+       differentiation.  (Contributed by Mario Carneiro, 7-Aug-2014.)  (Revised
+       by Jim Kingdon, 25-Jun-2023.) $)
+    df-dvap $a |- _D = ( s e. ~P CC , f e. ( CC ^pm s ) |->
+      U_ x e. ( ( int ` ( ( MetOpen ` ( abs o. - ) ) |`t s ) ) ` dom f )
```
avekens, line 136221: Why is ``( ( MetOpen ` ( abs o. - ) ) |`t s )`` used here instead of ``( ( TopOpen ` CCfld ) |`t s )``?

jkingdon: We haven't defined `CCfld` in iset.mm yet. Presumably this will happen some day, but a constructive field is not the same as a classical field and I guess I'm kind of undecided/unclear about the whole thing. It seems odd to have to tackle all the field slots if we are only using the metric. I'm also not sure about how to define everything, especially the order (`le`) slot.

The two notations are equivalent in set.mm as seen in https://us.metamath.org/mpeuni/cnfldtopn.html

iset.mm
jkingdon: Fixed in set.mm and iset.mm.

avekens (approved):

PR #3275, opened 2023-07-01
Split `~ footex`

### LINKED ISSUE #3271: Footex needs revision
I would like [footex](https://us.metamath.org/mpeuni/footex.html) to be revisioned or splitted into less time-draining theorems. Most of the reasoning is already explained in #3269. Here I provide a little bit of data.

I made a script that executes the commands:

<pre>
SHOW PROOF &ltlabel&gt /COMPRESSED
SHOW ELAPSED_TIME
</pre>

For every theorem in main. From this I extracted the time-spans between two subsequent `SHOW ELAPSED_TIME` commands to get the duration of the corresponding `show proof` command.

This is the resulting graph. The x-axis represents the Nth theorem in main, while the y-axis shows the elapsed time (in seconds) from the previous `SHOW ELAPSED_TIME` command:

linear

This doesn't show much... let's make it logarithmic instead:

log

That's better. In this one the y-axis represents the elapsed time expressed in centiseconds (so in this graph 100 means 1 second).

That big spike that you see towards the end is indeed [footex](https://us.metamath.org/mpeuni/footex.html). That means that I waited 3028.67 seconds before getting an answer from it, which is around 50 minutes. The big spike of footex appears for other commands as well such as `prove <label>` or for the minimize command (I chose `show proof` because it's the quickest to work with, but my guess it's that the elapsed times for these commands are correlated).

Thanks to my minimization in #3269 footex's proof decreased from 2178 to 2161 steps (in the compressed format, with wff and repeated steps included), and as I already mentioned I think they are not enough to explain its time-draining property. For the record the highest amount of steps in main belongs to theorem [2503lem2](https://us.metamath.org/mpeuni/2503lem2.html), which has 5279 of them (the website shows a smaller count because it doesn't include wff and repeated steps), but it required only 0.51 seconds to output `show proof 2503lem2 /compressed`.

So the properties that make some theorems more time-challenging than others are an open question to me, I'm looking forward to hear your thoughts about it.

tirix:
Splits [`~ footex` ](https://us.metamath.org/mpeuni/footex.html) into 3 parts. As mentioned in #3269, metamath-exe crashes when attempting to minimize `~ footex`, although it is not a particularly long proof.

I've kept the original version of `~ footex` as `~ footexALT`, so that it can be used for better understanding the issue, experiments, and possibly stress tests. See #3271 for a discussion and investigation on `~ footex`.

Fixes #3271

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15412,6 +15412,7 @@ New usage of "flddivrng" is discouraged (2 uses).
 New usage of "fldhmsubcALTV" is discouraged (0 uses).
 New usage of "fnexALT" is discouraged (0 uses).
 New usage of "fnmpt2ovdOLD" is discouraged (0 uses).
+New usage of "footexALT" is discouraged (0 uses).
 New usage of "fresisonOLD" is discouraged (0 uses).
 New usage of "frgrwopreglem5ALT" is discouraged (0 uses).
 New usage of "funadj" is discouraged (4 uses).
@@ -18897,6 +18898,7 @@ Proof modification of "fimaxreOLD" is discouraged (172 steps).
 Proof modification of "fiminreOLD" is discouraged (312 steps).
 Proof modification of "fnexALT" is discouraged (111 steps).
 Proof modification of "fnmpt2ovdOLD" is discouraged (219 steps).
+Proof modification of "footexALT" is discouraged (2161 steps).
 Proof modification of "frege10" is discouraged (27 steps).
 Proof modification of "frege100" is discouraged (31 steps).
 Proof modification of "frege101" is discouraged (49 steps).


wlammen (approved):

GinoGiotto:
I conducted some tests on the new theorems, targeting the commands that would typically cause metamath-exe to get stuck. These are the time reports:

`show proof footexlem1 /compressed` -> t = 0.14 s
`show proof footexlem2 /compressed`-> t = 0.22 s
`show proof footex /compressed` -> t = 0.94 s

`show proof footexlem1 /normal `-> t = 0.83 s
`show proof footexlem2 /normal `-> t = 1.41 s
`show proof footex /normal` -> t = 5.89 s

`save proof footexlem1 /compressed ` -> t = 0.04 s
`save proof footexlem2 /compressed ` -> t = 0.04 s
`save proof footex /compressed ` -> t = 0.67 s

`save proof footexlem1 /normal ` -> t = 0.04 s
`save proof footexlem2 /normal ` -> t = 0.06 s
`save proof footex /normal` -> t = 0.63 s

`prove footexlem1` -> t = 0.12 s
`MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*` -> t = 77.47 s
`save new_proof /normal` -> t = 0.01 s
`save new_proof /compressed` -> t = 0.01 s

`prove footexlem2` -> t = 0.11 s
`MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*` -> t = 132.39 s
`save new_proof /normal` -> t = 0.03 s
`save new_proof /compressed` -> t = 0.02 s

`prove footex` -> t = 0.33 s
`MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*` -> t = 592.35 s
`save new_proof /normal`-> t = 0.43 s
`save new_proof /compressed` -> t = 0.49 s

It's quite evident that the revision in this PR is a substantial improvement over the old state.

Only note: `~ footexALT` will be annoying when executing `save proof * /compressed` which as shown [here](https://github.com/metamath/set.mm/pull/3195#issuecomment-1556217537) is worth doing once in a while. I think in the long term `~ footexALT` should be deleted.

tirix:
Too bad I did not see your issues with ~footex earlier @GinoGiotto, if I had done that change earlier it would have spared you quite some computing power and time.

I think ideally we should try to reproduce that case and keep it in a separated  test/stress database, but I currently don't see how without pulling a substantial part of set.mm, or better identifying what made that proof so special.

GinoGiotto:
>Too bad I did not see your issues with ~footex earlier @GinoGiotto, if I had done that change earlier it would have spared you quite some computing power and time.

My initial intention with the second attempt was to produce a log file and report it as an issue (basically I wanted to see if metamath-exe writes something before crashing), but since it proceeded without interruptions I simply committed the shortened proof instead.

> I think ideally we should try to reproduce that case and keep it in a separated test/stress database, but I currently don't see how without pulling a substantial part of set.mm, or better identifying what made that proof so special.

Temporarily it might be enough a simple TODO note, like "Delete and extract a test case once its time-draining nature has been examined".

set.mm
```diff
@@ -386751,6 +386752,130 @@ are congruent (CPCTC), focusing on a specific segment.  (Contributed
...
+          UVEVOZWMVOWDUUEXMYPPNHLVEZVFNHVTYPPOHLVEUWAYPEJKLOHPQSTYSUUAYTUUGYPHO
+          YPHCHOEJKMQRSYSYTUUCYTUUAAUVLUVJVTYNAUVJUVLUOXHZWDAUURYNUUSWDWMVOAPHO
+          LVEVFYNAEJKLHOPQSTUAUEUIUJAHCHOEJKMQRSUAUEUCUEUIUWBUUSWMUQVQWDVRYPONH
+          LUUEWTXQXRXSXLXTACOVHYNUVCWDACOGLVEVFYOYCYNAEJKLOGCQTSUAUIUGUCACGOEJK
+          MQRSUAUCUGUIUVFXBYAWDXSXLAUVIYNUVMWDYDYPOGAOGVHYNUVTWDYBYEYFAONCEYQJK
+          LMQRSTYRUAUIUHUCAONCWPUUDVFUVROUUJMVEZVTAUVRUVQUWCUVSADUUJOMVDXCXFAON
+          CEYQJKLMQRSTYRUAUIUHUCXNXGXLYG $.
+      $}
+
+      $( Lemma for ~ foot : existence part.  (Contributed by Thierry Arnoux,
```
avekens, line 386846: I would not describe this as a lemma (and its label has, and justifiably so, no suffix "lem"), but with a full text description:
"From a point ` C ` outside of a line ` A ` , there exists a point ` x ` on ` A ` such that ` ( C L x ) ` is perpendicular to ` A ` .
This point is unique, see ~ foot . "

avekens:
> Only note: `~ footexALT` will be annoying when executing `save proof * /compressed` which as shown [here](https://github.com/metamath/set.mm/pull/3195#issuecomment-1556217537) is worth doing once in a while. I think in the long term `~ footexALT` should be deleted.

~ footexALT should be kept until the "problem" with metamath.exe is solved. It has a "(Proof modification is discouraged.)" tag, so a global minimization will not touch it, and I do not see a need to perform a global `save proof * /compressed` (`save proof * /compressed/fast` seems to be not affected).

avekens (approved):

PR #3276, opened 2023-07-01
move bj-spvv to main (as spvv)

icecream17:
spvw already exists, it's a version of sp

(add sp : spvw; spv : spvv notes)

auto use spvv in place of spv and bj-spvv
in both cases they were noted in bj's mathbox, so add tags to clarify origin

update bj-ru0 now that + spvv and - ax-13

sidenote: nalset is quite related to ru, through ruv vprc

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 1-Jul-23 bj-nalset nalset      moved from BJ's mathbox to main set.mm
+ 1-Jul-23 bj-spvv   spvv        moved from BJ's mathbox to main set.mm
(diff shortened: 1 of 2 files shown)


wlammen:
The important observation here is that ru, nalset need not depend on ax-13.

Is this a good reason enough to override the instruction of not changing the proof?   In my eyes yes.

wlammen (changes requested):
`$( $j usage 'ru' avoids 'ax-reg'; $)` should include ax-13.

wlammen (approved):

wlammen:
Still a merge conflict to resolve.

tirix:
@benjub for information.

benjub:
> @benjub for information.

LGTM. Thanks @tirix, @icecream17 and @wlammen

tirix (approved):

PR #3277, opened 2023-07-03
Improvement of ~2sqreu*

avekens:
Follow up of PR #3260

After proving ~reu3op, it was easy to prove that the representations ` E! p e. ( A X. B ) ph ` with ` p = <. a , b >. ` and ` ( E! x e. A E. y e. B ph /\ E! y e. B E. x e. A ph ) ` for "there are unique ` a ` and ` b ` so that .." are equivalent, see ~opreu2reurex. By this theorem, I could provide variants of the ~2sqreu* theorems with quantification over orderes pairs easily.

For one theorem I provide two variants:
* ~2sqreuopnnltb ``( P e. Prime -> ( ( P mod 4 ) = 1
                          <-> E! p e. ( NN X. NN ) ( ( 1st ` p ) < ( 2nd ` p ) /\ ( ( ( 1st ` p ) ^ 2 ) + ( ( 2nd ` p ) ^ 2 ) ) = P ) ) )`` 
* ~2sqreuopb: `( P e. Prime -> ( ( P mod 4 ) = 1
                          <-> E! p e. ( NN X. NN ) E. a E. b ( p = <. a , b >.  /\ ( a < b /\ ( ( a ^ 2 ) + ( b ^ 2 ) ) = P ) ) ) )`

Although the second variant is a little bit shorter, it has two additional quantifiers, which I do not like much. Therefore, I have no preferences which of these variants to use (in additional cases - for the current case, both variants should remain available for demonstration purposes).

Details:

Main set.mm:
* theorems ~reu3op, ~opreuopreu added
* theorem ~opreu2reurex1  replaced by ~opreu2reurex
* theorems ~addsq2nreurex and ~addsqn2reurex2 added
* comment of ~addsqnreup revised
* variants of ~2sqreu* using quantification over ordered pairs added

AV's mathbox:
* examples for restricted universal quantification in which the restricting class depends on the bound variable (~ralndv1, ~ralndv2)


set.mm
```diff
@@ -70904,6 +70916,23 @@ Power Set ( ~ ax-pow ).  (Contributed by Mario Carneiro, 20-May-2013.)
       KZRUQBCURHHUIUQVEUJUKULUMT $.
   $}
 
+  ${
+    $d A a b p $.  $d B a b p $.  $d ph a b $.
+    opreuopreu.a $e |- ( ( a = ( 1st ` p ) /\ b = ( 2nd ` p ) )
+                         -> ( ps <-> ph ) ) $.
+    $( There is a a unique ordered pair fulfilling a wff iff its components
```
wlammen, line 70923: "... a a ..." -> "... a ..."

set.mm
```diff
@@ -712466,6 +712642,19 @@ Restricted quantification (extension)
...
+     AV, 24-Jun-2023.) $)
+  ralndv1 $p |- A. x e. x _V e. x $=
+    ( cvv cv wcel wel elirrv pm2.21i rgen ) BACZDZAIAAEJAFGH $.
+
+  $( Second example for a theorem about a restricted universal quantification
+     in which the restricting class depends on the bound variable: all subsets
+     of a set are sets.  (Contributed by AV, 24-Jun-2023.) $)
+  ralndv2 $p |- A. x e. ~P x x e. _V $=
+    ( cv cvv wcel cpw vex rgenw ) ABZCDAHEAFG $.
+
```
wlammen, line 712657: Nice to have such examples at hand for reasoning about defining restricted quantification as ' A. x ( x e. A -> ph ) '

wlammen (changes requested):

set.mm
avekens: fixed

wlammen (approved):
A non-blocking remark: Consider using variables ph, ps, ch, th ... in this order.

avekens:
Merged, so that we can go ahead with #3280.

PR #3278, opened 2023-07-04
Add a few more $j commands

jkingdon:
These are $j usage commands ~~mostly for the axiom of regularity~~ for the peano axioms in iset.mm. ~~This is in set.mm and iset.mm.~~


iset.mm
```diff
@@ -35021,6 +35021,7 @@ conditioning it (with ` x e. z ` ) so that it asserts the existence of a
      depend on the Axiom of Regularity.  (Contributed by NM, 23-Aug-1993.) $)
   vprc $p |- -. _V e. _V $=
     ( vx cvv wcel cv wceq wex vnex isset mtbir ) BBCADBEAFAGABHI $.
+  $( $j usage 'vprc' avoids 'ax-setind'; $)
```
wlammen, line 35024: ax-setind (#4381) is located behind vprc (#3992) so this is somewhat questionable.  Why adding the $j in such a case?

iset.mm
```diff
@@ -38439,6 +38440,7 @@ rather than an implication (but the two are equivalent by ~ bm1.3ii ),
   ordon $p |- Ord On $=
     ( vx con0 word wtr cv wral tron wcel df-on abeq2i ordtr sylbi rgen mpbir2an
     dford3 ) BCBDAEZDZABFGQABPBHPCZQRABAIJPKLMABON $.
+  $( $j usage 'ordon' avoids 'ax-setind'; $)
```
wlammen, line 38443: The same for ordon, it is located before ax-setind

iset.mm
```diff
@@ -39807,6 +39813,7 @@ The natural numbers (i.e. finite ordinals)
       ZATVNWMWOWSAWOWPWQVLWPWQHWOWNWQVJVIUEUFUGWNVJVLUKUHUIVMAIPWLWSAWKWPWHWRVI
       IBOVKIBOUJULUMWHAVPPQUNWCWJCVPIBUOVBVSVPUPVTWGWBWIVSVPDRWAWHAVSVPVSVPVKRU
       QURUSQVPWDUTVAVCVRIIMZVRKVQWTVRIVDVEIIBVFVGQ $.
+    $( $j usage 'peano3' avoids 'ax-setind'; $)
```
wlammen, line 39816: peano3 -> peano5

set.mm
```diff
@@ -45473,6 +45473,7 @@ holding in an empty domain (see Axiom A5 and Rule R2 of [LeBlanc]
      depend on the Axiom of Regularity.  (Contributed by NM, 23-Aug-1993.) $)
   vprc $p |- -. _V e. _V $=
     ( vx cvv wcel cv wceq wex vnex isset mtbir ) BBCADBEAFAGABHI $.
+  $( $j usage 'vprc' avoids 'ax-reg'; $)
```
wlammen, line 45476: ax-reg is introduced much later

set.mm
```diff
@@ -74552,6 +74553,7 @@ currently used conventions for such cases (see ~ cbvmpt2x , ~ ovmpt2x and
       GZXIIZYRXKIZXLFSYPYRXGOCLNZYSYPYQLIZYRYROUUAYOUUBXOXJWCWDYRVDCYQLXGYRYRXF
       YQBPWEWFYRTIYSUUAWHYQBVPCLXGYRXHTYKWIWBWGXEYOYTWSXDYTAXJLWTXJOZXBYRXCXKUU
       CXAYQBWTXJULUMWTXJBPUNWJWKYRXIXKWLWMWNWOWPXDALWQWG $.
+    $( $j usage 'onnseq' avoids 'ax-reg'; $)
```
wlammen, line 74556: onnseq is introduced before ax-reg

wlammen (changes requested):
One error, two questionable additions in iset.mm.
Similar observation in set.mm

iset.mm
jkingdon: I kind of lost track about whether we made any decision about this.

The reason to add it is that it expresses an intention that might help in the future as people re-order theorems.

The reason not to add it is that it is, as you point out, redundant with the position in the file.

I guess I'll remove it at least for now. I don't have an especially strong opinion.

iset.mm
jkingdon: Thanks for noticing the copy paste error.

Fixed.

iset.mm
```diff
@@ -39754,6 +39754,7 @@ The natural numbers (i.e. finite ordinals)
       cint sbimi clelsb4 sylib sylbi mpgbir dfom3 eleqtrri ) DDAEZFZBEGUFFBUFHZ
       IZAJZRZKDUKFCEZUJFZDULFZLCCDUJMNUMUIACOZUNUICAPUOUGACOUNUIUGACUGUHQSCADTU
       AUBUCABUDUE $.
+    $( $j usage 'peano1' avoids 'ax-iinf' 'ax-inf2' 'ax-setind'; $)
```
wlammen, line 39757: ax-inf2 is not even in main.  Why do we want to have it mentioned here?

jkingdon: Ah, I didn't notice that was from a mathbox and is different than `ax-inf2` in set.mm. Probably we should rename it to avoid that confusion between the iset.mm one and the set.mm.

(And I agree there is no need for either it or https://us.metamath.org/ileuni/ax-infvn.html to be mentioned in the $j usage commands in main).

jkingdon: ax-inf2 removed.

wlammen (approved):

PR #3279, opened 2023-07-04
Add metamath-knife to automated checks

### LINKED ISSUE #3250: Proposal: be able to declare forbidden axioms for a proof

jkingdon:
This is mostly for the usage check, but while we're here run a verify too.

The result is that metamath-knife runs and gives the same result I'm seeing locally.  Which I think means that https://github.com/david-a-wheeler/metamath-knife/pull/118 has a bug, or exposes a bug which was there, or something. Once we can figure out what is going on there and fix it, we should be able to get passing tests on this pull request.

Fixes #3250.

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -178,6 +178,42 @@ jobs:
           ~/.cargo/bin/smetamath --verify --split --jobs 4 --timing ./iset.mm 2>&1 | tee smm.log && [ `egrep '(:Error:|:Warning:)' < smm.log; echo $?` -eq 1 ]
 
   ###########################################################
+  metamath-knife:
+    name: Verify using metamath-knife
+    runs-on: ubuntu-latest
+    needs: skip_dups
+    if: ${{ needs.skip_dups.outputs.should_skip != 'true' }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Cache metamath-knife
+        id: cache-metamath-knife
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-knife
+        with:
+          path: ~/.cargo/bin/cache-knife
+          key: ${{ runner.os }}-build-${{ env.cache-name }}
+      - name: Install rust
+        if: ${{ !steps.cache-metamath-knife.outputs.cache-hit }}
+        uses: hecrj/setup-rust-action@v1
+        with:
+          rust-version: stable
+      - run: git clone --depth 1 https://github.com/david-a-wheeler/metamath-knife.git
+      - run: cd metamath-knife && cargo build --release
+      - name: Verify set.mm
+        run: |
+          metamath-knife/target/release/metamath-knife --verify ./set.mm
+      - name: Verify iset.mm
+        run: |
+          metamath-knife/target/release/metamath-knife --verify ./iset.mm
+      - name: Check axiom usage for set.mm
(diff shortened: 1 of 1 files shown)


tirix:
That's interesting!
I've created [a PR](https://github.com/david-a-wheeler/metamath-knife/pull/120) to fix this specific issue, but this will not help in the case of definitional checks.

tirix:
The re-run passed with the new version of metamath-knife, and I've logged [an issue](https://github.com/david-a-wheeler/metamath-knife/issues/121) in metamath-knife to track the strange behaviour.

jkingdon:
> I've logged [an issue](https://github.com/david-a-wheeler/metamath-knife/issues/121) in metamath-knife to track the strange behaviour.

I saw different strange behavior with iset.mm but apparently the same fix dealt with it (with respect to `--verify-usage` anyway: as you point out, the problem may still be there for the definitional checks).

tirix (approved):
Three remarks:
- this check covers the same thing, with the same implementation, as smm3 (smetamath-rs), so the smm3 check could/should be removed.
- metamath-knife includes a `discouraged` file generation which is waaay faster than metamath-exe's. It would ease github's server loads and accelerate our CI to move it.
- metamath-knife has also `-t` and `-m` options, for typesetting and markup checks. They are also fast, and cover a bit more than other checkers. You could add it to the `-v` of this PR.

jkingdon:
I guess I'll merge this now and leave the door open for future pull requests for those other issues. They all sound reasonable to me (not that I really know enough about metamath-knife to know whether any of them would be controversial or present any difficulties). But I don't see any particular need to do everything at once.

PR #3280, opened 2023-07-04
Double restricted existential uniqueness

tirix:
- equivalence between double restricted existential uniqueness and ordered pair uniqueness
- definition for a syntax for double restricted existential uniqueness quantifier

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 4-Jul-23 rspcdf    [same]      moved from EW's mathbox to main set.mm
 25-Jun-23 2reu4     [same]      moved from AV's mathbox to main set.mm
 25-Jun-23 2reu4a    [same]      moved from AV's mathbox to main set.mm
 25-Jun-23 2reu2     [same]      moved from AV's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -16929,6 +16929,7 @@ New usage of "opeqsnOLD" is discouraged (1 uses).
 New usage of "opidon2OLD" is discouraged (1 uses).
 New usage of "opidonOLD" is discouraged (2 uses).
 New usage of "opnmblALT" is discouraged (0 uses).
+New usage of "opreu2reuALT" is discouraged (0 uses).
 New usage of "opsqrlem1" is discouraged (0 uses).
 New usage of "opsqrlem2" is discouraged (1 uses).
 New usage of "opsqrlem3" is discouraged (2 uses).
@@ -19383,6 +19384,7 @@ Proof modification of "opeqsnOLD" is discouraged (124 steps).
 Proof modification of "opidon2OLD" is discouraged (80 steps).
 Proof modification of "opidonOLD" is discouraged (198 steps).
 Proof modification of "opnmblALT" is discouraged (332 steps).
+Proof modification of "opreu2reuALT" is discouraged (913 steps).
 Proof modification of "orbi1r" is discouraged (9 steps).
 Proof modification of "orbi1rVD" is discouraged (101 steps).
 Proof modification of "orcomdd" is discouraged (13 steps).


tirix:
EDIT: Ok, it looks like this PR conflicts, or at least partially addresses the same objectives as #3277, which I noticed only after launching this PR.
@avekens feel free to have a look and comment on what seems interesting to keep or drop.

set.mm
```diff
@@ -457457,6 +457498,109 @@ and the expression ( x e. A /\ x e. B ) ` .
...
+  Double restricted existential uniqueness quantification
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  ${
+    $d a b p $.  $d a b ph $.  $d b x $.
+    opreu2reub.a $e |- ( p = <. a , b >. -> ( ph <-> ch ) ) $.
+    $( Lemma for ~ opreu2reub and other similar theorems.  (Contributed by
+       Thierry Arnoux, 4-Jul-2023.) $)
+    2reulem $p |- ( p = <. x , y >. -> ( ph <-> [. y / b ]. [. x / a ]. ch ) )
```
avekens, line 457512: This theorem is interesting by itself, and I would call it ~opsbc2ie.

set.mm
```diff
@@ -457457,6 +457498,109 @@ and the expression ( x e. A /\ x e. B ) ` .
...
+      OHQVMVOFUORUCVJVNUNFUOVLSZKVEFUOUNVLOUDVJVRVDUNVJVRFUOVKSZFUOVCSZJVDFUOVK
+      VCOUAVJVSUOVTVCFUOOUEFUOVCOUBUFUGUHTVJVPAFUOLZUSMVGABFUOOUIVJWAAUSAFUOOUJ
+      UKTULUMQVEVGGUPRUCVBVFUNGUPVDSZKURGUPUNVDOUDVBWBUQUNVBWBGUPUOSZGUPVCSZJUQ
+      GUPUOVCOUAVBWCUOWDUPGUPUOOUBGUPOUEUFUGUHTVBVHAGUPLZUTMVAAUSGUPOUIVBWEAUTA
+      GUPOUJUKTULUM $.
+
+    $d A a b p x y $.  $d B a b p x y $.  $d ph x y $.  $d ch p x y $.
+    $( Version of ~ opreu2reu with one less hypothesis.  (Contributed by
+       Thierry Arnoux, 4-Jul-2023.) $)
+    opreu2reus $p |- ( E! p e. ( A X. B ) ph -> E! a e. A E! b e. B ch ) $=
```
avekens, line 457527: With my recent PR #3277, I revised ~opreu2reu already to have only one hypothesis. Therefore, this theorem is not needed anymore.

set.mm
```diff
@@ -457457,6 +457498,109 @@ and the expression ( x e. A /\ x e. B ) ` .
...
+      ( vx vy cv wsbc 2reulem opreu2reu ) ABFIKLGJKLBIJCDEFGHABIJEFGHMN $.
+
+    $( Version of ~ opreu2reurex1 with one less hypothesis.  (Contributed by
+       Thierry Arnoux, 4-Jul-2023.) $)
+    opreu2reurex1s $p |- ( E! p e. ( A X. B ) ph
```
avekens, line 457532: With my recent PR #3277, ~ opreu2reurex1 became obsolete (would be simpoly is one direction of ~opreu2reurex). Therefore, this theorem is not needed anymore (except for proving ~opreu2reub).

set.mm
```diff
@@ -457457,6 +457498,109 @@ and the expression ( x e. A /\ x e. B ) ` .
...
+      -> ( E! a e. A E. b e. B ch /\ E! b e. B E. a e. A ch ) ) $=
+      ( vx vy cv wsbc 2reulem opreu2reurex1 ) ABFIKLGJKLBIJCDEFGHABIJEFGHMN $.
+
+    $d a b ch p x y $.
+    $( Correspondence between uniqueness of ordered pairs and double restricted
+       existential uniqueness quantification.  (Contributed by Thierry Arnoux,
+       4-Jul-2023.) $)
+    opreu2reub $p |- ( E! p e. ( A X. B ) ph
+                  <-> ( E! a e. A E. b e. B ch /\ E! b e. B E. a e. A ch ) ) $=
+      ( vx vy wreu wrex wa cv wceq wi wral wcel nfv nfan cxp opreu2reurex1s cop
```
avekens, line 457542: This theorem is identical with my theorem ~opreu2reurex which I provided with my recent PR #3277. My proof is much shorter, using ~reu3op and ~2reu4. Maybe the proof of ~opreu2reub is interesting (based on much simpler theorems), so it may be kept as ALT version.

avekens (changes requested):
There are many interesting theorems/definition in this PR, and most of them are in TA's mathbox. Therefore, I propose to remove only ~opreu2reus and ~opreu2reurex1s.

It would be nice to keep ~opreu2reub with an alternate proof for ~opreu2reurex,
but its proof uses ~opreu2reurex1s which uses ~opreu2reurex1 which will be
removed with my recent PR #3277 (could be replaced by ~opreu2reurex, but then
the alternate proof of ~opreu2reurex uses ~opreu2reurex...). A soulution  could be to
keep the old ~opreu2reurex1 in the mathbox.

set.mm
tirix: Removed.

set.mm
tirix: I renamed this ALT for reference, and only kept the direction which was new, so it does not refer to ~ opreu2reurex1.

tirix:
> There are many interesting theorems/definition in this PR, and most of them are in TA's mathbox. Therefore, I propose to remove only ~opreu2reus and ~opreu2reurex1s.
> 
Yes, I removed both.

> It would be nice to keep ~opreu2reub with an alternate proof for ~opreu2reurex, but its proof uses ~opreu2reurex1s which uses ~opreu2reurex1 which will be removed with my recent PR #3277 (could be replaced by ~opreu2reurex, but then the alternate proof of ~opreu2reurex uses ~opreu2reurex...). A soulution could be to keep the old ~opreu2reurex1 in the mathbox.
>
I kept it as an ALT theorem, and only for the direction which was new in the proof, so that there is no reference to the biconditional.

avekens (approved):
Everything is fine now!

PR #3281, opened 2023-07-04
shortening proof of lindsadd

Achaemenid:


wlammen (approved):

avekens (approved):

avekens:
Is the shortening technical only, or is a different approach used? In the second case, the old proof should be kept (lindsaddALT) and a "Proof shortened by" tag should be added.

wlammen:
> In the second case, the old proof should be kept (lindsaddALT) and a "Proof shortened by" tag should be added.

Since the theorem is in a mathbox, where people develop their ideas, at your option, you can bypass rules established for the main part.

I am assuming you are the owner of the mathbox, Brendan Leahy.

Achaemenid:
I'm not sure what qualifies as "purely technical" - I realized that I'd reinvented the wheel w.r.t. ringelnzr.  That's the only change I made.

PR #3282, opened 2023-07-06
Intuitionize dvfg and limcmo

jkingdon:
The statement of `dvfg`  only needs minor changes. The set.mm proof, however, is in terms of https://us.metamath.org/mpeuni/perfdvf.html which doesn't intuitionize in any easy way.

The intuitionized `limcmo` might not be the final word (in the sense that variations may be possible and useful in the future), but it is sufficient for now.  Both the statement of the theorem and the proof are fairly different from what is in set.mm.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10167,6 +10167,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>limcmo</td>
+  <td>~ limcimo</td>
+</tr>
+
+<tr>
   <td>limcmpt , limcmpt2</td>
   <td><i>none</i></td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10264,6 +10264,11 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>~ dvbsssg</td>
 </tr>
 
+<tr>
+  <td>dvfg</td>
+  <td>~ dvfgg</td>
+</tr>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT


iset.mm
```diff
@@ -136318,6 +136318,118 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
+  ${
+    limcflf.f $e |- ( ph -> F : A --> CC ) $.
+    limcflf.a $e |- ( ph -> A C_ CC ) $.
+    limcimo.b $e |- ( ph -> B e. CC ) $.
+    limcimo.bc $e |- ( ph -> B e. C ) $.
+    limcimo.bs $e |- ( ph -> B e. S ) $.
+    limcimo.c $e |- ( ph -> C e. ( K |`t S ) ) $.
+    limcimo.s $e |- ( ph -> S e. { RR , CC } ) $.
+    limcimo.ca $e |- ( ph -> { q e. C | q =//= B } C_ A ) $.
+    limcflfcntop.k $e |- K = ( MetOpen ` ( abs o. - ) ) $.
```
wlammen, line 136330: I wonder whether you should define K as ` K = ( ( MetOpen `(abs o. -) ) |` t S ), since you want to look at the limit points of A either in RR or CC.  This would make C a redundant helper class, and bring you more in line with the original.

wlammen (approved):
Given this word from you
`The intuitionized limcmo might not be the final word`
I think the version presented here is good enough.

iset.mm
jkingdon: I do appreciate these kinds of suggestions. There are a number of related theorems which are not intuitionizing easily. They include all the limit point theorems, perfect set theorems, and many of the places which do things like "look at what F does except at point B".

As for ` K `, I suppose in the next version of this I might not give a name to ` K ` at all and just write `` ( ( MetOpen ` (abs o. -) ) |` t S )`` itself, because it is lightly used in this proof.

I'm not sure I understand what would replace ` C ` though. ` C ` needs to contain ` B ` but ` A ` doesn't have to. Furthermore, ` C ` is open. I wonder if we should remove "limit point" from the comment for limcimo. That was copied from set.mm and I'm not sure how it applies here.

jkingdon:
> Given this word from you `The intuitionized limcmo might not be the final word` I think the version presented here is good enough.

The good news is that this pull request is able to prove a version of `dvfg` with only a simple and presumably unproblematic change to how the theorem is stated.

As for how it gets there and what we'll do in the future, yeah I could see either me or someone else working on that. Either just to make it nicer or for concrete reasons like being able to prove other theorems. At the moment I'm working on some of the `seq` theorems which are not yet intuitionized.

I made a small wording tweak to the comment.

PR #3285, opened 2023-07-09
iset: shorten djuin,djuun,djur and remove all dummy variables

benjub:
iset:
* shorten djuin,djuun,djur and remove all dummy variables
* minor comment edits

The versions of the statements I proposed last year (using systematic restrictions to relevant sets of the left and right injections instead of having unrestricted class versions), e.g., ~djuinr, ~djuunr, ~eldju versus ~djuin, ~djuun, ~djur respectively, have shorter proofs which do not use dummy variables. This is not accidental. In this PR, I prove the latter directly from the former, greatly reducing proof lengths and totally removing use of dummy variables.


wlammen:
Nice to see you on board again.

benjub:
> Nice to see you on board again.

Thank you Wolf.

avekens (approved):

PR #3288, opened 2023-07-11
Fix installing texlive in metamath-exe

david-a-wheeler:
The metamath-exe branch does a number of TeX checks, and that requires installing those packages. Currently installing the relevant packages to do the checks is failing. This change attempts to fix that.


david-a-wheeler:
I normally wait for a second person to approve, but this doesn't change any mathematics, it's purely to fix a CI/CD bug, and this modified version passes. So I'm going to merge it now.

PR #3292, opened 2023-07-14
iset.mm: Shroeder-Bernstein fares little better if one set is _om

jkingdon:
This shows that the special case of Shroeder-Bertstein where one set is _om, plus the Limited Principle of Omniscience (LPO) implies excluded middle.

It is a formalization of the github comment https://github.com/metamath/set.mm/issues/2595#issuecomment-1157389136 by @digama0 

* includes xp01disjl which isn't needed yet but which is aimed eventually replacing df-cda with df-dju
* adds a few listings to mmil.html missing theorems list
* includes a variety of miscellaneous disjoint union theorems: endjusym , inl11 , caseinr , djudomr , djudoml
* includes some miscellaneous natural number and equinumerosity theorems: omp1eom , nnsucpred

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4584,6 +4584,17 @@ this implies excluded middle</TD>
   developed and to a certain extent cannot be.</TD>
 </TR>
 
+<tr>
+  <td>uncdadom</td>
+  <td><i>none</i></td>
+  <td>The set.mm proof relies on undom</td>
+</tr>
+
+<tr>
+  <td>pm110.643</td>
+  <td>~ dju1p1e2</td>
+</tr>
+
 <TR>
   <TD>fodom , fodomnum</TD>
   <TD><I>none</I></TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4595,6 +4595,11 @@ this implies excluded middle</TD>
   <td>~ dju1p1e2</td>
 </tr>
 
+<tr>
+  <td>cdadom3</td>
+  <td>~ djudoml</td>
+</tr>
+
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -67402,6 +67432,36 @@ readily usable (e.g., by ~ djudom and ~ djufun ) while the simpler
       AIWJTVGYSWKYADWTWQWLWMAXCBDAYCBSQXCBPUMYSSYABIWNWMWOWP $.
   $}
 
+  ${
+    caseinr.f $e |- ( ph -> Fun F ) $.
+    caseinr.g $e |- ( ph -> G Fn B ) $.
+    caseinr.a $e |- ( ph -> A e. B ) $.
+    $( Applying the "case" construction to a right injection.  (Contributed by
+       Jim Kingdon, 12-Jul-2023.) $)
+    caseinr $p |- ( ph -> ( case ( F , G ) ` ( inr ` A ) ) = ( G ` A ) ) $=
```
benjub, line 67441: As usual, my inclination would have been to state it with `` ( ( inr |` B ) ` A ) `` since A is assumed to be in B and G is a function on  B, but I haven't checked how the proof and typical usages would be affected. Up to you.

avekens (approved):

iset.mm
jkingdon: There are actually some parts of this pull request where I do use some of the theorems which restrict `inr` and `inl`. So at least in that sense our practice is more similar than it might appear.

But you are also right that I don't do this everywhere. And "too cumbersome" is a decent summary of the reason.

PR #3295, opened 2023-07-15
sbtOLD --> sbtALT; shorten epelg; mathbox: bj-elbr, bj-elep and bj-epelg

benjub:
Title says it all.
I changed sbtOLD to sbtALT since the proof is much shorter than that of sbt and the additional axioms used (ax-4 and ax-5) are part of the core axioms, so I think it is worth keeping it. It illustrates the standard proof of theorems like ` |- A.x PH -> PS` from inferences like `|- PH => |- PS` when generalization is done only over `x`.

I would like to move the theorems I introduced in this PR from my mathbox to main: these are natural results (didn't try to minimize with them yet), the result bj-elep is stronger than epelg and permits a short proof of it (bj-epelg). @avekens do you agree to change 0nelrel and 0nelfun as I propose in my mathbox ? For 0nelrel, the usages would be more direct, see bj-0nelopab, bj-elbr.

BTW: 0neqopab is probably a typo for 0nelopab.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15353,6 +15353,7 @@ New usage of "enrbreq" is discouraged (3 uses).
 New usage of "enreceq" is discouraged (9 uses).
 New usage of "enrer" is discouraged (8 uses).
 New usage of "enrex" is discouraged (9 uses).
+New usage of "epelgOLD" is discouraged (0 uses).
 New usage of "eqeq1dALT" is discouraged (0 uses).
 New usage of "eqeqan12dALT" is discouraged (0 uses).
 New usage of "eqeqan1dOLD" is discouraged (0 uses).
@@ -17567,7 +17568,7 @@ New usage of "sbnOLD" is discouraged (0 uses).
 New usage of "sbnf2OLD" is discouraged (0 uses).
 New usage of "sbnv" is discouraged (1 uses).
 New usage of "sbrimALT" is discouraged (1 uses).
-New usage of "sbtOLD" is discouraged (0 uses).
+New usage of "sbtALT" is discouraged (0 uses).
 New usage of "sbtT" is discouraged (0 uses).
 New usage of "sbtv" is discouraged (0 uses).
 New usage of "scmateALT" is discouraged (0 uses).
@@ -18902,6 +18903,7 @@ Proof modification of "elunirnALT" is discouraged (38 steps).
 Proof modification of "en3lpVD" is discouraged (147 steps).
 Proof modification of "en3lplem1VD" is discouraged (95 steps).
 Proof modification of "en3lplem2VD" is discouraged (267 steps).
+Proof modification of "epelgOLD" is discouraged (136 steps).
 Proof modification of "eqeq1dALT" is discouraged (62 steps).
 Proof modification of "eqeqan12dALT" is discouraged (23 steps).
 Proof modification of "eqeqan1dOLD" is discouraged (12 steps).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -20984,23 +20981,20 @@ requires either a disjoint variable condition ( ~ sb5 ) or a non-freeness
 
   ${
     $d x y $.
-    $( Version of ~ equsb1 with a disjoint variable condition, which neither
-       requires ~ ax-12 nor ~ ax-13 .  (Contributed by BJ, 11-Sep-2019.)
-       Remove dependencies on axioms.  (Revised by Wolf Lammen, 30-May-2023.)
-       (Proof shortened by Steven Nguyen, 19-Jun-2023.)
+    $( Obsolete version of ~ equsb1v as of 19-Jun-2023.  (Contributed by BJ,
```
wlammen, line 20984: I checked out set.mm as of 10-Jul-23, and equsb1v looked the same there as the OLD version here.  I think it became obsolete on 11-Jul.

wlammen (changes requested):

set.mm
benjub: Thanks. This is corrected.

wlammen: rewrapping is needed.

benjub: Rewrapped. I changed the order of two paragraphs in the comment of df-sb in order to have the two paragraphs about notation next to each other.

wlammen (approved):

icecream17 (approved):

avekens:
> I would like to move the theorems I introduced in this PR from my mathbox to main: these are natural results (didn't try to minimize with them yet), the result bj-elep is stronger than epelg and permits a short proof of it (bj-epelg). @avekens do you agree to change 0nelrel and 0nelfun as I propose in my mathbox ? For 0nelrel, the usages would be more direct, see bj-0nelopab, bj-elbr.

This concerns the old discussion if or how to use the symbol ∉ (`x e/ A`) or to replace it by `-. x e. A`.  (Rel 𝑅 → ∅ ∉ 𝑅) just looks nicer and more natural in the context of set theory, therefore I would like to keep it.

> BTW: 0neqopab is probably a typo for 0nelopab.

You are right! If you want, you can rename it 0nelopab.

benjub:
> This concerns the old discussion if or how to use the symbol ∉ (`x e/ A`) or to replace it by `-. x e. A`. (Rel 𝑅 → ∅ ∉ 𝑅) just looks nicer and more natural in the context of set theory, therefore I would like to keep it.

Ok. I need to use the statement `(Rel R -> -. (/) e. R )`, which is at Step 5 of your proof https://us.metamath.org/mpeuni/0nelrel.html. Can I extract it as a separate theorem, and then prove 0nelrel from it ? As for labels, both should be "0nelrel"... Maybe I could label `(Rel R -> -. (/) e. R )` as 0nelrel0 to keep the label 0nelrel unchanged for `(Rel R -> (/) e/ R )`, but I'm open to suggestions.

avekens:
> > This concerns the old discussion if or how to use the symbol ∉ (`x e/ A`) or to replace it by `-. x e. A`. (Rel 𝑅 → ∅ ∉ 𝑅) just looks nicer and more natural in the context of set theory, therefore I would like to keep it.
> 
> Ok. I need to use the statement `(Rel R -> -. (/) e. R )`, which is at Step 5 of your proof https://us.metamath.org/mpeuni/0nelrel.html. Can I extract it as a separate theorem, and then prove 0nelrel from it ? As for labels, both should be "0nelrel"... Maybe I could label `(Rel R -> -. (/) e. R )` as 0nelrel0 to keep the label 0nelrel unchanged for `(Rel R -> (/) e/ R )`, but I'm open to suggestions.

OK, that's a good suggestion, I fully agree.

benjub:
> OK, that's a good suggestion, I fully agree.

Thanks. I'm going to do that in a future MR, to ease reviews.

PR #3298, opened 2023-07-16
mathbox:  let equsalvw explain the basics of substitution

wlammen:
Start explaining df-sb.  Maybe for beginners the definition is too complex to grasp at first (or even second) sight.  Note that some have difficulties in understanding df-bi, so we might expect a similar reaction when confronted with df-sb.  I start here with equsalvw, stating some basic facts about substitution.  Later  comment extensions in sb6 or df-sb may reference this text.

For better discussion and reference I copied the added comment here.  Note that I have no education in mathematics, so a professional might spot deficiencies there.

wl-equsalvw is meant as a replacement for equsalvw in the end, if we can agree on some text.

-----------------------------------------------

       This theorem lays the foundation to a transformation of expressions
       called substitution of set variables in a wff.  Only in this particular
       context we additionally assume ` ph ` and ` y ` disjoint, stated here as
       ` ph ( x ) ` .  Similarly the disjointness of ` ps ` and ` x ` is
       expressed by ` ps ( y ) ` .  Both ` ph ` and ` ps ` may still depend on
       other set variables, but that is irrelevant here.

       We want to transform ` ph ( x ) ` into ` ps ( y ) ` such that ` ps `
       depends on ` y ` the same way as ` ph ` depends on ` x ` .  This
       dependency is expressed in our hypothesis (called implicit
       substitution): ` ( x = y -> ( ph <-> ps ) ) ` .  For primitive enough
       ` ph ` a sort of textual substitution of ` x ` by ` y ` is sufficient
       for such transformation.  But note: ` ph ` must not contain wff
       variables, and the substitution is no proper textual substitution
       either.  We still need grammar information to not accidently replace the
       x in a token 'x.' denoting multiplication, but only catch set variables
       ` x ` .  Our current stage of development allows only equations and
       quantifiers make up such primitives.  Thanks to ~ equequ1 and ~ cbvalvw
       we can then prove in a mechanical way that in fact the implicit
       substitution holds for each instance.

       If ` ph ` contains wff variables we cannot use textual transformation
       any longer, since we don't know how to replace ` y ` for ` x ` in
       placeholders of unknown structure.  Our theorem now states, that the
       generic expression ` A. x ( x = y -> ph ) ` formally behaves as if such
       a substitution was possible and made.


icecream17 (approved):

benjub:
You are right that ~equsalvw deserves an expanded comment. I would like to propose a few paragraphs in the next few days if you do not mind waiting.

Theorem ~equsalvw is fundamental in that it makes the axiomatization of first-order logic possible in a Metamath system. That is, it permits to dispense with the notions of free and bound variables and proper substitution, and use only the much simpler notions of occurring variable and plain substitution. To do that, one needs to be able to express proper substitution in these simpler terms, and that is precisely what ~equsalvw does. This was first noticed in [Tar65], see last paragraph of page 62 and Lemmas 16 and 17. This is also crucial in the proof of Megill's completeness theorem [Meg95, Thm. 9.7], see third point of Remark 9.1.

In https://arxiv.org/abs/2202.10383, I expressed it as:
"""
Remark 0.1. A fundamental insight of Tarski, which makes these systems with weaker metalogic work, is that if x and y are disjoint (that is, the variables substituted for x and for y should be distinct), then the formula ∀x(x ≡ y → ϕ), which we denote by [y/x]ϕ, is equivalent to the result of the proper substitution of y for x in ϕ.
"""

That the hypothesis of ~equsalvw can be interpreted as "psi is the result of properly substituting y for x in phi" is proved for any concrete formula by induction on its height, see Equation (14) in https://arxiv.org/abs/2202.10383:
"""
T ⊢ ( x ≡ y → (Φ ↔ Φx←y), {{y, m} | m ∈ {x} ∪ OC(Φ)}).    (14)

This is easily proved by induction on the height of Φ. The base cases correspond to Φ atomic, that is, of the form φi, or xi ≡ xj , or similarly with a nonlogical predicate. These cases are respectively trivial and proved from EQ and from the predicate axiom schemes for the nonlogical predicates. As for the induction steps, the propositional calculus cases (that is, Φ is an implication or a negation) pose no difficulty. If Φ is of the form ∀zΨ with x ≠ z, then the result follows easily from ALLdistr. Finally, suppose that Φ is of the form ∀xΨ. One has to prove (x ≡ y → (∀xΨ ↔ ∀yΨx←y), {{y, m} | m ∈ OC(Ψ) ∪ {x}}) knowing by induction hypothesis that (x ≡ y → (Ψ ↔ Ψx←y), {{y, m} | m ∈ OC(Ψ) ∪ {x}}). This is a consequence of cbvalvw in set.mm.
"""
(for proper formatting with correctly displayed exponents and indices, see the article).

[Meg95] Norman D. Megill. “A Finitely Axiomatized Formalization of Predicate Calculus with Equality”. Notre Dame Journal of Formal Logic 25 (1995), pp. 435–453. url: https://us.metamath.org/downloads/finiteaxiom.pdf.

[Tar65] Alfred Tarski. “A simplified formalization of predicate logic with identity”. Archiv für Mathematische Logik und Grundlagenforschung 7 (1965), pp. 61–79.

I can email you [Tar65] if you want. The other two articles are freely available.

wlammen:
The only book covering foundational axiomatic systems I ever read was https://en.wikipedia.org/wiki/G%C3%B6del,_Escher,_Bach.  I think I more or less understand what Tarski's work is about.  It is mentioned all over the place in the chapters about predicate logic.  Finally I skimmed through your work.  I am not sure whether ALLEq needs to have the antecedent changed to ∀𝑦 𝑦 = 𝑥.  If I remember right you otherwise run into difficulty proving aecom.

(supplement: I just looked into finiteaxiom.pdf.  ALLEq corresponds to B11 in System S2, or axc11.  set.mm starts out from axc11r instead.  Is this an equivalent choice?  I faintly remember I struggled with this question some years ago, and in the end concluded the difference matters.  I was then unable to prove aecom from axc11 (and the other axioms).  But I don't remember the details any more.)

`.. if you do not mind waiting.`

It certainly is not that pressing.  I might merge this PR, though, to get my fork of set.mm free for other work.  This should not stop you from changing the comment of equsalvw.

Side note: After the revision of df-sb ax-12 is referenced through ax12v and axc11r only (hooray!).   Should we discourage a new usage then?

wlammen:
`Is axc11 an equivalent choice in comparison to axc11r?`

Certainly not, since axc11r is easily derived from axc11, but the other way round is somewhat tricky and requires more axioms.

`Should we discourage new usage of ax-12 then?`
It's already discouraged.

PR #3301, opened 2023-07-18
ErALTV: equivalence relation on its natural domain

mazsa:
Mathbox only.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -8033,8 +8033,8 @@
 "idrefALT" is used by "elrefsymrels3".
 "idrefALT" is used by "idinxpssinxp2".
 "idrefALT" is used by "idinxpssinxp3".
-"idrefALT" is used by "refrelred3".
-"idrefALT" is used by "refrelsred3".
+"idrefALT" is used by "refrelredund3".
+"idrefALT" is used by "refrelsredund3".
 "idrefALT" is used by "refsymrels3".
 "idrefALT" is used by "symrefref3".
 "idrval" is used by "cmpidelt".


wlammen:
Is there a reason why the checks are mostly skipped?

mazsa:
> Is there a reason why the checks are mostly skipped?

Yes: because 
1. your #3298 PR was merged between my third and fourth commit above and I updated my mazsa-ErALTV branch with it 
2. the verifiers automatically checked the product of this and all was green 
3. because the merge of #3298 was now in my commit history above, I decided to rebase my branch based on metamath/develop in order to remove the merge from my commit history. 
4. Rebase + force push result in exactly the same file (but removed merge from history), and this is why verifiers did not run again:
" Run fkirc/skip-duplicate-actions@master
Skip execution because the exact same files have been successfully checked in run https://github.com/metamath/set.mm/actions/runs/5587202955 "

set.mm
```diff
@@ -437141,9 +437141,12 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 htmldef "EqvRel" as ' EqvRel ';
   althtmldef "EqvRel" as ' EqvRel ';
   latexdef "EqvRel" as "\mathrm{EqvRel}";
-htmldef "ElEqvRel" as ' ElEqvRel ';
-  althtmldef "ElEqvRel" as ' ElEqvRel ';
-  latexdef "ElEqvRel" as "\mathrm{ElEqvRel}";
+htmldef "CoElEqvRels" as ' CoElEqvRels ';
```
wlammen, line 437144: This affects the global token space.  In future someone might want to reserve the same tokens for a different purpose.  In order to allow parallel development not interfering each other, I suggest to add a prefix pm- to the tokens, like I did with wl-in:
`htmldef "wl-in" as " : ";
  althtmldef "wl-in" as " : ";
  latexdef "wl-in" as "\text{ : }";`

set.mm
```diff
@@ -550065,13 +550086,23 @@ A collection of theorems for commuting equalities (or
   $c EqvRels $. $( The class of equivalence relations $)
   $c EqvRel $. $( Equivalence relation predicate $)
 
-  $c ElEqvRel $. $( Elementhood equivalence relation predicate
-                    (used only for a placeholder definition) $)
+  $c CoElEqvRels $. $( The class of coelement equivalence relations $)
+  $c CoElEqvRel $. $( Coelement equivalence relation predicate $)
 
   $c Reds $. $( The class of redundants $)
```
wlammen, line 550092: If possible, use a similar technique here, and localize constants introduced for your private usage at the moment.

set.mm
mazsa: "ElEqvRel" was my token as well, this is just a renaming in my Mathbox. 

I think you are completely right (i.e., we need a solution which makes possible parallel development), especially in cases when we already have a definition for the same concept in set.mm , i.e., when you explicitly want to change a definition of an existing concept. However, this is not the case at "EqvRel" or "CoElEqvRel": e.g. the concept of my definition
```metamath
df-eqvrel $a |- ( EqvRel R <-> ( RefRel R /\ SymRel R /\ TrRel R ) ) $.
```
, however conventional outside of set.mm, is not in the main set.mm , so we do not need a superfluously personalized token for this to avoid possible future cases. If, in the future, somebody finds a better definition for this concept, s/he may use our [convention](https://us.metamath.org/mpeuni/conventions.html) for his/her case:  "Alternative (ALTV) versions. If a theorem or definition is an alternate/variant of an existing theorem or definition, its label should have the same name with suffix ALTV. Such alternates are often temporary only, until it is decided which alternative should be used in the future." 

Our convention above is only about the ending of the theorem label, but I use it in stead of your "pm-*" suggestion in the token as well, like "*ALTV". For example, in case of the "equivalence relation with domain" concept, I think we should retire our present 
```metamath
df-er $a |- ( R Er A <-> ( Rel R /\ dom R = A /\ ( `' R u. ( R o. R ) ) C_ R ) ) $.
```
and use
```metamath
  dferALTV2 $p |- ( R ErALTV A <-> ( EqvRel R /\ ( dom R /. R ) = A ) ) $=
```
. If we are recommended to use "ALTV" in the label, I think we should try to use "ALTV" in the token as well and avoid our conventional "df-erALTV" label with an apparently different kind of personalized "R pm-Er A" token. I think the definition of equivalence relation in set.mm is a not so crowded space which would need more personalized tokens than the ALTV-solution I use.

set.mm
mazsa: (same as in https://github.com/metamath/set.mm/pull/3301#discussion_r1267942266 )

wlammen:
I am more bothered about the short tokens like 'red'.  You used it for redundancy, but it may in future be a suggesting term elsewhere with the meaning of 'reduction', for example.  Reserving this short token for exclusive use in your mathbox, while, to avoid a conflict, another longer token sees widespread usage in main, would be an unfortunate result.  In my profession as a software developer, name clashing and global name space pollution are well-known and well-addressed problems, with solutions provided by virtually all program languages for decades now.

Since we cannot easily change the metamath language, we need the cooperation and awareness of all contributors instead.  Prefixing all globally visible tokens with your two-character shorthand like pm- is a viable solution here.

So far name clashing was perhaps no issue in Metamath, and it might turn out difficult to find a negative example at all.  Maybe Metamath never scales to the point, where this effect really hurts.  Its perhaps just my IT-professional attitude that lets me cringe here.  What do the others think?

If there are no objections voiced in this thread, I will accept the traditional way, approve and merge this PR by tomorrow.  Of course, another maintainer can step in and finish it instead.

mazsa:
1. I think you are completely right with red: I’m not sure about it anyway (I met with its formula in such a case, I don’t remember where, which was unrelated to redundancy: perhaps redundancy is a special case of something else and this may never upgrade to the main). I’ll modify it in a commit later today.

2. In one of next PRs, perhaps in the next, I’ll define disjointness with a label df-disjALTV but with a token Disj (not DisjALTV) because a) I’m quite sure that this will be our definition of disjointness in the long run b) I think that the final token should be Disj (not Disj_ which is in the main now) and c) these two, simple or underlined https://us.metamath.org/mpeuni/df-disj.html , are quite different anyway. Perhaps I should use the DisjALTV token: now I’m going to use the Disj token version as planned, but I’ll mention you @wlammen in the PR to decide whether I can keep Disj as a token or I have to change it to DisjALTV .

3. Do you think we can place eqbrrdv2 from Rodolfo Medina’s Mathbox to the main (and delete eqbrrdv2TEMP)?

wlammen:
`Do you think we can place eqbrrdv2 to main?` 
I am not an authoritative instance to decide this, in particular, since I haven't got so far in set.mm.  My impression is: The closely related eqrelrdv2 is already in main.  You need a simple transition between df-br and df-op for this theorem, and 
eqbrrdv2 allows it.  Sadly, eqbrrdiv is not general enough for this task.

I have seen people move theorems from mathboxes to main to fulfil their particular need before.  So, go ahead.  At least I won't complain.

Alternatively, one could turn the proof of eqrelrdv into a full deduction version eqrelrdvd, i.e. use hypotheses ph -> Rel A & ph -> Rel B , and prove eqrelrdv from it with a shortened proof.  This way is perhaps more to our usual conventions.

wlammen (approved):
as announced.  There were no further objections regarding global name space pollution raised, so I approve it as following the tradition of other contributions.

PR #3302, opened 2023-07-18
add wl-axc11rALT

wlammen:
1. mathbox: show that axc11r is easily derived from axc11.  The reverse direction is way more complicated, see main.
2. delete outdated OLD theorems

Clarifies a conjecture made in #3298 .

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17551,6 +17551,7 @@ New usage of "sbi2ALT" is discouraged (1 uses).
 New usage of "sbi2v" is discouraged (0 uses).
 New usage of "sbieALT" is discouraged (1 uses).
 New usage of "sbiedALT" is discouraged (1 uses).
+New usage of "sbievOLD" is discouraged (0 uses).
 New usage of "sbimALT" is discouraged (3 uses).
 New usage of "sbimdOLD" is discouraged (0 uses).
 New usage of "sbimdvOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -530899,6 +530841,21 @@ extends this idea to multiple additions (TODO).
...
+    $( Proving ~ axc11 from ~ axc11r (in the end from ~ ax-12 ) is somewhat
+       tricky and requires ~ ax-10 and ~ ax-13 .  The reverse direction is
+       rather simple as demonstrated here.  We provide the two instances of
+       ~ axc11 used in the proof here as hypotheses.  Note that besides these
+       references to ~ ax-12 no other axiom beyond ~ ax-7 is used.
+       (Contributed by Wolf Lammen, 18-Jul-2023.)  (New usage is discouraged.)
+       (Proof modification is discouraged.) $)
+    wl-axc11rALT $p |- ( A. y y = x -> ( A. x ph -> A. y ph ) ) $=
+      ( weq wal wi pm2.43i equcomi alimi 3syl ) CBFZCGZMBGZBCFZBGABGACGHNODIMPB
+      CBJKEL $.
```
benjub, line 530856: I think this could go to Main, somewhere between ax-7 and ax-8, with label ~axc11rc11 (since I think it is the usual labeling convention: when proving axiom A from axiom B, label it axAB). Then, the third sentence of the comment becomes unnecessary, and I would reverse the beginning of the comment: in the first sentence, describe what you prove ("Prove ~axc11 from ~axc11r."), and then, comment that the reverse direction is harder. Probably, then, no need of any discouragement tag.

PR #3311, opened 2023-07-23
Twin concepts: function and disjoint

mazsa:
Both redefined via a more general concept in order to reveal their common origin. Mathbox only.

Disjointness, defined this way (as opposed to the present df-disj $a |- ( Disj_ x e. A B <-> A. y E* x e. A y e. B ) https://us.metamath.org/mpeuni/df-disj.html way), will be necessary to define partition in one of my next PRs as the general
dfpart2 $p |- ( R Part A <-> ( Disj R /\ ( dom R /. R ) = A ) ) , 
(cf.   dferALTV2 $p |- ( R ErALTV A <-> ( EqvRel R /\ ( dom R /. R ) = A ) )  ),
of which membership partition, which does not contain the null set, is a special case:
dfmembpart2 $p |- ( MembPart A <-> ( ElDisj A /\ -. (/) e. A ) ) .
The latter is the conventional definition of (membership) partition, at least outside of Metamath:)

@wlammen : ALTVs in the labels as requested by our Conventions. In case of function I applied ALTV in the token as well, in case of disjoint I did not as described in https://github.com/metamath/set.mm/pull/3301#issuecomment-1643758334 .


wlammen:
I am a bit at a loss what the definitions mean.  I have the idea that a function is a relation, but I don't know what Disjoints are.  Since it all affects your mathbox only, well, go ahead.  The difficulty is the global name space pollution, but I won't decide this differently than last time.

wlammen (approved):

mazsa:
> I am a bit at a loss what the definitions mean. I have the idea that a function is a relation, but I don't know what Disjoints are. 

It’s a way of generalization of the original “two sets are said to be disjoint sets if they have no element in common. Equivalently, two disjoint sets are sets whose intersection is the empty set” idea
```metamath
disj $p |- ( ( A i^i B ) = (/) <-> A. x e. A -. x e. B ) 
```

-> first, for a class:
Class A is disjoint when its elements are disjoint which is 
```metamath
dfeldisj5 $p |- ( ElDisj A <-> A. u e. A A. v e. A ( u = v \/ ( u i^i v ) = (/) ) )
```

-> then, instead of the elementhood relation restricted to the class above, 
(where 
```metamath
df-eldisj $a |- ( ElDisj A <-> Disj ( `' _E |` A ) ) 
```
)
disjointness with respect to a general relation:
```metamath
dfdisjALTV5 $p |- ( Disj R <-> ( A. u e. dom R A. v e. dom R ( u = v \/ ( [ u ] R i^i [ v ] R ) = (/) ) /\ Rel R ) ) 
```

That it is worthwhile to generalize the definition in this way is shown by the concept of partition, where the original idea is that “The elements of P are said to be pairwise disjoint (... and) P does not contain the empty set", i.e.,
```metamath
dfmembpart2 $p |- ( MembPart A <-> ( ElDisj A /\ -. (/) e. A ) )
```
(Notice that 
```metamath
n0el3 $p |- ( -. (/) e. A <-> ( dom ( `' _E |` A ) /. ( `' _E |` A ) ) = A ) 
```
)
see its generalization ~ dfpart2 above, etc.

tirix:
As far as notations are concerned, we already have `Disj_` in the main part of set.mm, for "indexed" disjointness.

I think it would make sense to keep the parallel with e.g. `U_` for indexed union and `U.` for set union, and have `Disj.` for set disjointness, in place of your `ElDisj`, so as to keep a coherent notation system.

david-a-wheeler:
I'm all for consistent notation when we can get it.

Also, @mazsa et al, can I convince you to add some of this wonderful discussions into a comment somewhere (usually as part of the first definition where it matters, with other definitions linking to it)? That way future readers are more likely to be able to find and read the answer :-).

mazsa:
@tirix if you regard only
( A u. B ) , U_ x e. A B , U. A 
and
( A i^i B ) = (/) , Disj_ x e. A B , Disj. A 
concepts, you are right.

The problem is that we have another story on top on that as well:
we should name the generalization of this Disj. A / ElDisj A a kind of general disjointness Disj as well
since we have a major theorem
```metamath
pet2 $p |- ( ( Disj ( R |X. ( `' _E |` A ) ) /\ ( dom ( R |X. ( `' _E |` A ) ) /. ( R |X. ( `' _E |` A ) ) ) = A ) <->
              ( EqvRel ,~ ( R |X. ( `' _E |` A ) ) /\ ( dom ,~ ( R |X. ( `' _E |` A ) ) /. ,~ ( R |X. ( `' _E |` A ) ) ) = A ) )
```
similar to
```metamath
mpet3 $p |- ( ( ElDisj A /\ -. (/) e. A ) <-> ( CoElEqvRel A /\ ( U. A /. ~ A ) = A ) )
```

So, what would be the name of this general disjointness other than
( A i^i B ) = (/) , ElDisj A , Disj R ?

In case of union we do not have this problem because
- we do not have such a major theorem for the similar general case of U. A , which is { y | E. x x R y } ,
- moreover, { y | E. x x R y } has another, very conventional use case which is independent of of U. A , 
so we can call it range:
```metamath
dfrn2 $p |- ran R = { y | E. x x R y } 
```

I.e., while we clearly do not have a ( A u. B ) , U_ x e. A B , U. A -> ran R problem,
we clearly have a problem with ( A i^i B ) = (/) , Disj_ x e. A B , Disj. A -> ??? ,
and I think the solution of this should be
( A i^i B ) = (/) , ?Disj_ x e. A B , ElDisj A <- Disj R .

@david-a-wheeler thank you for your idea: I'll do that in one of my next PRs.

PR #3312, opened 2023-07-23
use sbievw

icecream17:
remove $d conditions (see commit)
the only notable effect of the first commit that I found was that sbievw does not need `$d y ph $.`

sbco2vvOLD becomes the new proof of sbco2vv

shorten equsb1v

use sbievw

---
I tried to prove dfsb1 without ax-13 but it doesn't work out
It would look something like:
image
image

Despite the comment of the (deleted version of) sbco2vvOLD, sbequALT does use ax-13.
sbequvv(ALT) doesn't but it requires `$d x y $.`

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15372,6 +15372,7 @@ New usage of "equs5eALT" is discouraged (0 uses).
 New usage of "equsb1ALT" is discouraged (1 uses).
 New usage of "equsb1vOLD" is discouraged (0 uses).
 New usage of "equsb1vOLDOLD" is discouraged (0 uses).
+New usage of "equsb1vOLDOLDOLD" is discouraged (0 uses).
 New usage of "equsb3ALT" is discouraged (0 uses).
 New usage of "equsb3vOLD" is discouraged (0 uses).
 New usage of "equsexALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


wlammen:
I'd be surprised if dfsb1 was provable without ax-13.  See equs4 and how it depends on ax6e, a simple theorem using ax-13.

benjub:
@icecream17 can you rewrap ? I'd like to have it merged soon since I have some work which might conflict with your changes.

set.mm
```diff
@@ -16868,14 +16868,44 @@ either all bound ( ` x , y ` ) or all free ( ` t ` ).  Also note that
...
+
   ${
     $d x y $.
     $( Version of ~ equsb1 with a disjoint variable condition, which neither
        requires ~ ax-12 nor ~ ax-13 .  (Contributed by BJ, 11-Sep-2019.)
        Remove dependencies on axioms.  (Revised by Wolf Lammen, 30-May-2023.)
        (Proof shortened by Steven Nguyen, 19-Jun-2023.)  Revise ~ df-sb .
-       (Revised by Steven Nguyen, 11-Jul-2023.) $)
+       (Revised by Steven Nguyen, 11-Jul-2023.)  (Proof shortened by Steven
+       Nguyen, 22-Jul-2023.) $)
```
benjub, line 16897: There are three tags very close in time. Maybe you could have only two (adapt as you see fit).

benjub (approved):
One non-blocking comment. Thanks for this work, and the DV cleaning (I overlooked the removal from sbievw; it is important).

PR #3313, opened 2023-07-23
equsv

benjub:
* I added equsv since I think it is an important step of equsalvw, and so I can refer to it from the comment of ~sbv.

* In my mathbox for now, I added a few theorems including bj-sbievw, which is the closed form of sbievw. There is no hurry, but I think it could be moved to Main.  Opinions ? @wlammen ?

* mndind : I just saw a TODO, so I did it.

* mathbox cleanup

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+22-Jul-23 0neqopab  0nelopab
 16-Jul-23 qseq2i    [same]      moved from PM's mathbox to main set.mm
 16-Jul-23 qseq2d    [same]      moved from PM's mathbox to main set.mm
 14-Jul-23 bj-ssb0   sbv         moved from BJ's mathbox to main set.mm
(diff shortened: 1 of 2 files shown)


wlammen:
> Opinions?

I don't see an application now, but if you are in need you're glad it's there.  Having a closed form, or a deduction version always increases versatility.

set.mm
```diff
@@ -522483,10 +522496,11 @@ Utility lemmas or strengthenings of theorems in the main part (biconditional
...
-    $( Substitution in an equality, disjoint variables case.  Uses only
-       ax-1--6.  It might be shorter to prove the result about composition of
-       two substitutions and prove ~ bj-ssbeq first with a DV on x,t, and then
-       in the general case.  (Contributed by BJ, 22-Dec-2020.) $)
+    $( Substitution in an equality, disjoint variables case.  Uses only ~ ax-1
+       through ~ ax-6 .  It might be shorter to prove the result about
+       composition of two substitutions and prove ~ bj-ssbeq first with a DV
+       condition on ` x , t ` , and then in the general case.  (Contributed by
+       BJ, 22-Dec-2020.)  (Proof modification is discouraged.) $)
     bj-ssbeq $p |- ( [ t / x ] y = z <-> y = z ) $=
```
wlammen, line 522504: looks like an instance of sbv

wlammen (approved):

set.mm
benjub: Thanks. I'll remove it in my next MR.

PR #3316, opened 2023-07-25
The multiplicativity formula for degrees of field extensions

tirix:
Building up on my recent work on the dimension of vector spaces, this adds to set.mm a proof for [the multiplicativity formula for degrees of field extensions](https://en.wikipedia.org/wiki/Degree_of_a_field_extension#The_multiplicativity_formula_for_degrees).
A few theorems from other mathboxes are moved to the main part of set.mm.

The usual notation for the degree of a field extension is `[E:F]`, but I have not introduced a new notation (yet). I'm thinking about using `( E [:] F )`, similar to the open/closed interval `( E [,] F )`.
Comment on whether this would be a useful definition are welcome - I'd add this first in my mathbox anyway.

Sorry for the big amount of changes, I did not see any remarkable by-products.


tirix:
Maybe the field extension notations could be introduced in two steps: First, define a relation `/FExt` where the _wff_ `E /FExt F` would stand for "E is a field extension of F":
```
/FExt = { <. e , f >. | E. u ( f = ( e |`s u ) /\ u e. ( SubRing ` e ) /\ f e. DivRing ) }
```
Then, use this to restrict the domain of the "degree of the field extension" operation to only pairs `E/F` of structures, where `E` is an extension of `F`:
```
[:] = ( e e. _V , f e. ( /FExt " { e } ) |-> ( dim ` ( ( subringAlg ` e ) ` ( Base ` f ) ) ) )
```

changes-set.txt
```diff
@@ -27,6 +27,11 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+25-Jul-23 mpanrd    [same]      moved from ML's mathbox to main set.mm
```
wlammen, line 30: typo: The name is mpnanrd

wlammen:
This seems to be mostly affecting your mathbox.  Right?

set.mm
```diff
@@ -467526,6 +468136,94 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  ${
+    drgext.b $e |- B = ( ( subringAlg ` E ) ` U ) $.
+    drgext.f $e |- F = ( E |`s U ) $.
+    drgext.1 $e |- ( ph -> E e. DivRing ) $.
+    drgext.2 $e |- ( ph -> F e. DivRing ) $.
+    drgext.3 $e |- ( ph -> U e. ( SubRing ` E ) ) $.
+    $( The additive neutral element of a division ring extension.  (Contributed
```
avekens, line 468152: Are the hypotheses drgext.f and drgext.2 used? I see no `F` in the following theorems.

avekens (commented):
Great result! I have only one remark (see inline comment).

avekens:
> Maybe the field extension notations could be introduced in two steps: First, define a relation `/FExt` where the _wff_ `E /FExt F` would stand for "E is a field extension of F":
> 
> ```
> /FExt = { <. e , f >. | E. u ( f = ( e |`s u ) /\ u e. ( SubRing ` e ) /\ f e. DivRing ) }
> ```
> 
> Then, use this to restrict the domain of the "degree of the field extension" operation to only pairs `E/F` of structures, where `E` is an extension of `F`:
> 
> ```
> [:] = ( e e. _V , f e. ( /FExt " { e } ) |-> ( dim ` ( ( subringAlg ` e ) ` ( Base ` f ) ) ) )
> ```

These definitions would be fine for me, but they should remain in the mathbox until there is a broader use of them.

tirix:
> These definitions would be fine for me, but they should remain in the mathbox until there is a broader use of them.

Yes, that's what I plan to do. 
I think it's fair to ask though, because once I use them they will be reserved globally.

set.mm
tirix: They are not used for the first 3 theorems `~drgext0g`, `~drgextvsca` and `~drgext0gsca`, but the fact that `F` is a division ring is used the the later theorems. I'll move those two hypotheses later so that they only appear when needed.

avekens: OK,  `F` seems to be an abbreviation for ``( E |`s U )``  used within the proofs, but still does not appear in the theorems themselves. So the hypothesis ``drgext.3 $e |- ( ph ->  ( E |`s U )  e. DivRing ) $.`` should be sufficient.

I do not know if there is any (unwritten) convention to define symbols in hypotheses only if they are used in the theorems (and not in proofs only), and I am also not aware of any similar case in set.mm.

Nevertheless, this is OK for now.

avekens (approved):

set.mm
tirix: I also wanted to follow the same model as for the main theorem, `fedgmul`. 
We'll see if I maybe reword those theorems using the proposed new syntax (maybe a "Division Ring Extension syntax").

PR #3317, opened 2023-07-25
equsb3v, elsb3v, elsb4v -> lemma, fix regression of sb4b

wlammen:
The *v variants from the title have no advantage over the general versions, since they depend on the same set of axioms.  So turn them into lemmas, as intermediate steps to the final general versions.

sb4b and sb4 now depend on the same axioms (without ax-11) as before the introduction of the new df-sb.  This has a cascading effect on other theorems like sbcom3, for example.  Don't keep OLD versions, the current versions can be seen as initial proofs after transition to the new df-sb.

In the context of the former df-sb theorem sb2 was provable without ax-10.  We cannot achieve this here, because the old proof is a close consequence of the old df-sb.  So we have a kind of regression here.  Fortunately, almost all users of sb2 (except equsb1) employ ax-10 anyway, so this does not hurt much.  This version at least does not depend on ax-11, unlike the overwritten version.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17487,6 +17487,7 @@ New usage of "sb2ALT" is discouraged (7 uses).
 New usage of "sb2vALT" is discouraged (1 uses).
 New usage of "sb2vOLD" is discouraged (0 uses).
 New usage of "sb4ALT" is discouraged (4 uses).
+New usage of "sb4OLD" is discouraged (0 uses).
 New usage of "sb4aALT" is discouraged (1 uses).
 New usage of "sb4bOLD" is discouraged (0 uses).
 New usage of "sb4vALT" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


icecream17 (approved):
may also optionally remove cvjustOLD

benjub:
@wlammen : 
* could you remove cvjustOLD as @icecream17 noted, since it is now identical to cvjust ?
* I agree with you that equsb3ALT could be removed.
* could you swap elsb4lem and elsb3 (so that each lemma comes right before the associated result) ?
* I think sbco2vvOLD could be removed now without waiting one year.

icecream17:
i think applying sbievw twice should be extracted into a theorem, so that all the lem versions become unnecessary

benjub:
> i think applying sbievw twice should be extracted into a theorem, so that all the lem versions become unnecessary

Yes, I think that's a better option indeed !  Can you do it ?

PR #3319, opened 2023-07-27
Provide a way to show that a set is countably infinite in iset.mm

jkingdon:
This is the sort of thing which is often done with Shroeder-Bernstein if you have that (but we don't, of course).

The important theorem here is:

```
    $( A condition for a set being countably infinite.  Corollary 8.1.13 of
       [AczelRathjen], p. 73.  (Contributed by Jim Kingdon, 27-Oct-2022.) $)
    ennnfone $p |- ( A ~~ NN <-> ( A. x e. A A. y e. A DECID x = y
      /\ E. f ( f : NN0 -onto-> A
        /\ A. n e. NN0 E. k e. NN0 A. j e. ( 0 ... n )
        ( f ` k ) =/= ( f ` j ) ) ) ) $=
```

The intended application is #2595

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2553,8 +2553,8 @@ is biconditional, but just for natural numbers.</TD>
 
 <TR>
 <TD>ordtr2 , ontr2</TD>
-<TD><I>none</I></TD>
-<TD>Implies excluded middle as shown at ~ ontr2exmid</TD>
+<TD>~ nntr2</TD>
+<TD>ontr2 implies excluded middle as shown at ~ ontr2exmid</TD>
 </TR>
 
 <TR>
@@ -2576,7 +2576,7 @@ implies excluded middle by ~ ordsoexmid .</TD>
 
 <TR>
 <TD>ordsssuc</TD>
-<TD>~ trsucss </TD>
+<TD>~ trsucss , ~ nnsssuc</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -60031,6 +60031,23 @@ ordinals implies excluded middle ( ~ ordsucunielexmid ).  (Contributed
     ABFZABDZABGZHZUKULBADZIUOABJUKUPUOUKUOUPUKUMUNUPKUOUPHABLUMUNUPMNOQPUKUMULU
     NUKBRZUMULSUJUQUIBTUAUQUMULBAUBUCUDUNULSUKABUEUFUGUH $.
 
+  $( A subset of a natural number belongs to its successor.  (Contributed by
```
benjub, line 60034: {0, 2} is a subset of the natural number 3 but does not belong to its successor.

The $p-statement states that a natural number is a subset of another natural number if and only if it belongs to its successor.

jkingdon: Agreed, we'll need to fix https://us.metamath.org/mpeuni/ordsssuc.html too which has the same wording problem. (I'll see if I can think of something a bit more concise which is correct).

jkingdon: I've pushed up a wording fix.

benjub: Note that both $p-statements are biconditionals. I will abstain and let someone else review.

jkingdon: Yeah, I noticed the biconditional and I also had a bunch more thoughts including about things like what the antecedent of "it" is in the various wordings, and what to do when the formal theorem itself is more self explanatory than would be the case in other situations.

The pull request as it stands mostly sticks with Norm's wording but with one tweak to deal with the issue of what kind of subsets are considered.

Your suggestion of letting other people review seems like a good one - perhaps they will have ideas about how to word this and/or whether this pull request is a good venue for getting into it.

icecream17: given that `e.` is like `<`, this says something like `A C_ B <-> A <_ B`
So a (perhaps uninituitive) comment could be: `` Equivalent versions of ` <_ ` for ordinals. ``
(I feel like such a comment would probably need some more explanation following it)

Otherwise I don't think the comment could really be shortened

avekens: I like @benjub `s suggestion:  "A natural number is a subset of another natural number if and only if it belongs to its successor." It is not much longer than the current comment, but expresses the important fact that both directions hold.

iset.mm
```diff
@@ -127782,6 +127864,740 @@ reduced fraction representation (no common factors, denominator
...
+      SYKUVHUUPVLXGXHXIRXFYTUUQUVCPDUUPUUPXJQHUUCXJXKXLYIUUPMZYJUUQYSUVCKCYIUUP
+      XMUWMYRUVBGKUWMYOUUTFEKYQUWMYLUURYNUUSYKYIUUPXNYMYIUUPXNXOXQXHXRXSRXTTYAV
+      S $.
+  $}
+
+  ${
+    $d A f j n x y $.  $d f j k n $.
+    $( A condition for a set being countably infinite.  Corollary 8.1.13 of
+       [AczelRathjen], p. 73.  (Contributed by Jim Kingdon, 27-Oct-2022.) $)
+    ennnfone $p |- ( A ~~ NN <-> ( A. x e. A A. y e. A DECID x = y
```
avekens, line 128591: Is there a short and concise description of the condition? What is the different/advantage compared with f being a bijection (see ~bren)? Why is the condition about NN0, but the LHS contains NN?

iset.mm
jkingdon: We have two votes for that wording, so that's good enough for me.

iset.mm
jkingdon: > Is there a short and concise description of the condition?

I've pushed up a stab at it. "A is countable and infinite" gets at the intuition but I can't imagine those exact words would ever get by you and @benjub (and not even me without some kind of qualification, which is why I expanded significantly on that thought in the comment I just tried to write).

> What is the different/advantage compared with f being a bijection (see ~bren)?

The best way to answer that will be to start using this theorem. I was looking at the way that AczelRathjen does this to prove `ZZ ~~ NN` and `QQ ~~ NN` and to be honest there were some steps in their proofs of those two things which seemed to me to be missing (or at least, not spelled out).

I might give a try at https://us.metamath.org/mpeuni/unben.html - that looks easier than `QQ ~~ NN`, and at least at first glance would appear to be a case where the theorem in this pull request might be easier than building the bijection directly (the set.mm proof of unben uses excluded middle - via https://us.metamath.org/mpeuni/unbnn.html ).

> Why is the condition about NN0, but the LHS contains NN?

LOL, I thought you'd ask about ` _om ` versus ` NN0 ` rather than ` NN0 ` versus ` NN ` .

Using ` NN ` in `A ~~ NN` is our convention in theorems like https://us.metamath.org/mpeuni/prminf.html , https://us.metamath.org/mpeuni/qnnen.html , and https://us.metamath.org/mpeuni/unben.html .

On the right side, I sort of think the best style might be ` _om ` as that is what we use in theorems like https://us.metamath.org/ileuni/finct.html and https://us.metamath.org/ileuni/df-omni.html . But some level of conversion between ` NN0 ` and ` _om ` is inevitable in theorems like this (`QQ` is expressed in terms of `ZZ` rather than `_om`, `seq` is expressed in terms of `ZZ`, etc). To be honest I stated `ennnfone` this way while we were still having a bunch of discussions about which one to use in some of these theorems and never really thought to revisit because I was very focused on how to prove it (a topic which I got stuck on multiple times before eventually finishing the proof). I guess I don't see any big need to change anything now, and indeed I sort of suspect time may produce greater clarity on what the most convenient way(s) to express it is.

avekens: Thanks a lot for these explanations, its much clearer to me now (although I am very curious about the application of this theorem, as you indicated).

avekens (approved):

iset.mm
jkingdon: Yeah, I'm curious too. It would be a whole lot easier if we could prove something like
```
( E. f f : _om -onto-> A ) /\ _om ~<_ A )
  -> A ~~ NN
```
but my (relatively cursory) attempts at proving that as a corollary of `ennnfone` didn't seem to go anywhere.

So I may take a break from this topic unless I get an idea or decide to dive back in looking for one.

PR #3323, opened 2023-07-28
mathbox minor edits

benjub:


set.mm
```diff
@@ -439147,6 +439147,18 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 htmldef "RRhat" as "RRhat";
   althtmldef "RRhat" as "&#8477;&#x302;";
   latexdef "RRhat" as "\hat{\mathbb{R}}";
+htmldef "NNbar" as "NNbar";
```
wlammen, line 439150: Is this fixed to this set https://en.wikipedia.org/wiki/Extended_natural_numbers?

set.mm
```diff
@@ -439147,6 +439147,18 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
...
+  althtmldef "NNbar" as "&#8469;&#x305;";
+  latexdef "NNbar" as "\bar{\mathbb{N}}";
+htmldef "ZZbar" as "ZZbar";
+  althtmldef "ZZbar" as "&#8484;&#x305;";
+  latexdef "ZZbar" as "\bar{\mathbb{Z}}";
+htmldef "ZZhat" as "ZZhat";
+  althtmldef "ZZhat" as "&#8484;&#x302;";
+  latexdef "ZZhat" as "\hat{\mathbb{Z}}";
+htmldef "||C" as " ||C ";
+  althtmldef "||C" as " &#8741;<sub>&#8450;</sub> ";
```
wlammen, line 439160: What does the C stand for?

benjub: I needed to distinguish it from the already existing token `||` so I chose "C" for "complex".

set.mm
benjub: Yes, it's only a set; I should add its definition as ` |- NNbar = ( NN0 u. pinfty ) `.

set.mm
benjub: (added a reference in the comment of df-bj-divc)

set.mm
benjub: (added the three definitions of NNbar, ZZbar, ZZhat)

wlammen (approved):

avekens:
Having only the definitions, but no (at least basic) theorems using them makes it difficult to to understand the intention behind them. Furthermore, the title of this PR is a little bit misleading: I think introducing new definitions are not minor changes...

benjub:
> Having only the definitions, but no (at least basic) theorems using them makes it difficult to to understand the intention behind them. Furthermore, the title of this PR is a little bit misleading: I think introducing new definitions are not minor changes...

You're right. I wrote the definitions to remove them from my backlog, but I should soon prove some basic facts.

icecream17:
I know this is very late but df-pnf and df-mnf might be changed to some other arbitrary value that is not a complex (or really, real) number. (compare axioms of rru and pwuninel)

This would change a lot of the definitions in this pr (maybe you meant `u. { +inf }` instead of `u. +inf `?)

benjub:
> I know this is very late but df-pnf and df-mnf might be changed to some other arbitrary value that is not a complex (or really, real) number. (compare axioms of rru and pwuninel)
> 
> This would change a lot of the definitions in this pr (maybe you meant `u. { +inf }` instead of `u. +inf `?)

Indeed: singletons everywhere, thanks ! I defined them not completely arbitrarily: they are both on the "circle at infinity" in the extended complex numbers (you can see near the beginning of that section that I am in the process of changing the paramaterization of that circle).

tirix:
Note @digama0 @benjub :
metamath-knife's markup checker throws the errors below following this PR.

This is due to the `` ` `` character in the commented-out formula, assumed to be an opening math string. 
Within math strings, it expects all tokens to be math tokens.

In #3338 I'm doubling those in order to avoid the issue. 

```
$ metamath-knife -m set.mm
warning: Undeclared token '$='
      --> set.mm:529051:29
       |
529051 |      $p |- ( {R ` 0R ) = 0R $= ? $.
       |                             -- This token was not declared in any $v or $c statement
       |
warning: Undeclared token '?'
      --> set.mm:529051:32
       |
529051 |      $p |- ( {R ` 0R ) = 0R $= ? $.
       |                                - This token was not declared in any $v or $c statement
       |
warning: Undeclared token '$.'
      --> set.mm:529051:34
       |
529051 |      $p |- ( {R ` 0R ) = 0R $= ? $.
       |                                  -- This token was not declared in any $v or $c statement
       |
warning: Undeclared token '$p'
      --> set.mm:529053:7
       |
529053 |       $p |- ( {R ` 1/2 ) = 1/2 $= ? $.
       |       -- This token was not declared in any $v or $c statement
       |
warning: Undeclared token '$.'
      --> set.mm:529055:65
       |
529055 |      df-minfty $a |- minfty = ( inftyexpitau ` <. 1/2 , 0R >. ) $. $)
       |                                                                 -- This token was not declared in any $v or $c statement
       |
error: Unclosed math
      --> set.mm:529055:68
       |
529055 |      df-minfty $a |- minfty = ( inftyexpitau ` <. 1/2 , 0R >. ) $. $)
       |                                                                    ^^ A math string must be closed with `
       |
      ::: set.mm:529055:46
       |
529055 |      df-minfty $a |- minfty = ( inftyexpitau ` <. 1/2 , 0R >. ) $. $)
       |                                              - note: Math string started here
       |
6 diagnostics issued.
```

benjub:
The changes in set.mm are fine by me.  I think the formatting checks of metamath.exe and metamath-knife should not diverge. For instance, I used the standard way to rewrap/check/etc. before an MR, that is, the one indicated at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md so either the metamath-knife check or the metamath.exe check should be changed to have the same rule (probably the metamath.exe format check should add that check).

tirix:
> Either the metamath-knife check or the metamath.exe check should be changed to have the same rule.

Yes, sounds like common sense! 

Or we add the `-m` option to the `metamath-knife` CI run, so that this check is covered in the `set.mm`  CI (that's a fraction of a second).

benjub:
> Or we add the `-m` option to the `metamath-knife` CI run, so that this check is covered in the `set.mm` CI (that's a fraction of a second).

But then we'll be in a situation where a submitter may submit an MR that he has formatted in the recommended way at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md, has not encountered any error or warning in the process, but once he submits the MR, the CI fails.

digama0:
I think it is still better to have more checks in CI. I'm not sure if we are ready to suggest that users use metamath-knife instead of metamath-exe for preparing PRs, but eventually I would like it to be usable for that purpose.

> The changes in set.mm are fine by me. I think the formatting checks of metamath.exe and metamath-knife should not diverge.

They do diverge, but mostly only in pathological cases. The metamath-exe implementation is bonkers, involving layers of regexes, and has many strange failure modes. Metamath-knife will instead report an error on problem inputs that would cause metamath-exe to produce broken behavior. (In fact, I believe the example in this issue is one of these problem inputs.)

benjub:
> I think it is still better to have more checks in CI. I'm not sure if we are ready to suggest that users use metamath-knife instead of metamath-exe for preparing PRs, but eventually I would like it to be usable for that purpose.

I agree. But I would rather have first a simple way to format `.mm` files for preparing PRs, documented on `CONTRIBUTING.md`, and only after that put CI checks.

digama0:
I think the metamath-exe formatter is just fine, I don't think there are any issues with it (although I haven't yet tried to reimplement it so maybe I will find something if I do). So what you said is not inconsistent with having an extra CI check that metamath-exe can't do. There is already one CI check that metamath-exe can't do, namely the definition check from mmj2.

PR #3324, opened 2023-07-28
revert sbequ8ALT to be the normal version

wlammen:
sbequ8 is a theorem closely connected to the old df-sb.  It used only axioms up to ax-4.  We cannot even come close with our new definition of substitution.  This is one price we have to pay.

Fortunately, sbequ8 has no applications, so it hurts less.

Revert sbequ8ALT to the normal version, as this version at least does not depend on ax-11.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17526,7 +17526,6 @@ New usage of "sbequ1ALT" is discouraged (4 uses).
 New usage of "sbequ1OLD" is discouraged (0 uses).
 New usage of "sbequ2ALT" is discouraged (4 uses).
 New usage of "sbequ2OLD" is discouraged (0 uses).
-New usage of "sbequ8ALT" is discouraged (0 uses).
 New usage of "sbequALT" is discouraged (1 uses).
 New usage of "sbequOLD" is discouraged (0 uses).
 New usage of "sbequiALT" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -21411,12 +21394,23 @@ proposition with a distinct variable (closed form of ~ nfsb4 ).
...
+  ${
+    $d x y $.
+    $( Obsolete version of ~ equsb1v as of 11-Jul-2023.  (Contributed by BJ,
+       11-Sep-2019.)  Remove dependencies on axioms.  (Revised by Wolf Lammen,
+       30-May-2023.)  (Proof shortened by Steven Nguyen, 19-Jun-2023.)
+       (Proof modification is discouraged.)  (New usage is discouraged.) $)
+    equsb1vOLDOLD $p |- [ y / x ] x = y $=
+      ( weq wsb wi id sbt sbequ8 mpbir ) ABCZABDJJEZABDKABJFGJABHI $.
+  $}
+
```
benjub, line 21413: Could be removed if you agree.

benjub (approved):

avekens (approved):

icecream17 (approved):

set.mm
wlammen: next pull trquest

PR #3328, opened 2023-07-29
Theorems about (double) restricted existential uniqueness and unodered pairs

avekens:
* ~spc2ed , spc2d moved from TA's mathbox to main

AV's mathbox:
* ~2reureurex, ~2reureurex, ~sq2reunnltb, ~addsqnot2reu  added in section "Double restricted existential uniqueness quantification syntax" in TA's mathbox, providing some applications of the recently added definition of the double restricted existential uniqueness quantifier.
* ~oppr, ~opprb, ~ poprelb, ~2exopprim, ~reuopreuprim added in section "Set of proper unordered pairs" in AV's mathbox.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+29-Jul-23 spc2d     [same]      moved from TA's mathbox to main set.mm
+29-Jul-23 spc2ed    [same]      moved from TA's mathbox to main set.mm
  4-Jul-23 rspcdf    [same]      moved from EW's mathbox to main set.mm
  1-Jul-23 bj-nalset nalset      moved from BJ's mathbox to main set.mm
  1-Jul-23 bj-spvv   spvv        moved from BJ's mathbox to main set.mm


avekens:
@tirix I hope you don't mind that I added some theorems to your mathbox.

tirix:
> @tirix I hope you don't mind that I added some theorems to your mathbox.

Not the least, and I'm sure GitHub will handle merging changes gracefully.

tirix (approved):
Mathbox(es) only, and some theorems look like they could be useful enough to be moved to main some day.

set.mm
```diff
@@ -458067,6 +458068,20 @@ and the expression ( x e. A /\ x e. B ) ` .
     ( w2reu wrex wreu df-2reu simplbi reurex syl ) ABCDEFZACEGZBDHZNBDGMOABDGCE
     HABCDEIJNBDKL $.
 
+  $( Double restricted existential uniqueness implies restricted existential
+     uniqueness with restricted existence.  (Contributed by AV, 5-Jul-2023.) $)
+  2reureurex $p |- ( E! x e. A , y e. B ph -> E! x e. A E. y e. B ph ) $=
```
benjub, line 458073: I didn't know this construct. I find it hard to read. @tirix : why not define it like ~copab with respect to ~cab ?
i.e.,
```
df-opeu $a |- ( E! <. x , y >. ph <-> E. z A. x A. y ( ph <-> <. x , y >. = z ) ) $.
```
and then define the restricted version as usual:
```
df-opreu $a |- ( E! <. x , y >. e. A ph <-> E! <. x , y >. ( <. x , y >. e. A /\ ph ) ) $.
```
Then, you can write expressions like `E! <. x , y >. e. ( A X. B ) ph` which I find more readable than `E! x e. A , y e. B ph`. This is also more versatile since you may sometimes want to write expressions like `E! <. x , y >. e. A ph` without writing `A` explicitly as a product.

tirix: > I didn't know this construct.

This is purely in my mathbox, as an idea/proposal/suggestion, so it's a bit hidden.
I like the fact that the syntax you suggest shows clearly we're talking about a _pair_. @avekens what's your opinion?

avekens: Although the current definition lokks nicer some how, the suggusten of @benjub has sevral practical advantages, therefore I would propose to change the definitionaccordingly.

BTW: in df-opeu, there must be an E! instead of E. at the RHS,

benjub: > BTW: in df-opeu, there must be an E! instead of E. at the RHS,

No, I think `E.` is fine (even though it is also true with `E!`).

mazsa: > ```
> df-opreu $a |- ( E! <. x , y >. e. A ph <-> E! <. x , y >. ( <. x , y >. e. A /\ ph ) ) $.
> ```
[...]
> BTW: in df-opeu, there must be an E! instead of E. at the RHS,
[...]
> No, I think E. is fine (even though it is also true with E!).

Do you think is it possible to define this consistently, for all these cases?
|- ( A. <. x , y >. ph <-> [...]
|- ( A! <. x , y >. ph <-> [...] (in @david-a-wheeler 's Mathbox)
|- ( E. <. x , y >. ph <-> [...]
|- ( E* <. x , y >. ph <-> [...]
|- ( E! <. x , y >. ph <-> [...]
and
|- ( A. <. x , y >. e. A ph <-> [...]
|- ( A! <. x , y >. e. A ph <-> [...]
|- ( E. <. x , y >. e. A ph <-> [...]
|- ( E* <. x , y >. e. A ph <-> [...]
|- ( E! <. x , y >. e. A ph <-> [...]

mazsa: Of course not in fact (because in a lot of cases it is equivalent with the duplication) but in theory.

benjub: Yes, you can do more or less the same thing for other quantifiers. But note that "for all x and for all y ..." means the same thing as "for all couples <. x , y >. ...", and similarly for "there exists", so there is limited interest to add gratuitously a new piece of notation. That could be done for `E*` but it's probably not worth introducing a new piece of notation either, with too few interesting applications.

mazsa: (off topic: @david-a-wheeler I have some old theorems about your allsome quantifier: do you think I should commit them to your mathbox? https://github.com/mazsa/set.mm/tree/mazsa-alsc )

avekens: > @mazsa : I'd love to see some theorems about the allsome quantifier! If you have proven theorems, please propose adding them somewhere. Once you make the proposal we can quibble about "where they go":
> 
>     1. We could put them in your mathbox, we'll just need to make sure your mathbox is after mine. That's probably the easiest.

That won't work, because the CI checks will complain (verify markup)

> 
>     2. Put them in my mathbox, with credit to you. I haven't seen them yet, and I'd want to review them before putting them in my own mathbox, but I'll probably +1 them :-).

I would prefer this variant.

> 
>     3. We generally "move things into main" if multiple people use it, but we're also generally conservative about adding new definitions. So I think it's better to start with putting them in a mathbox, and then discussing other options separately.

We should wait until there are more applications.

PR #3332, opened 2023-07-30
Misc

benjub:
Better reviewed by commit. Commit messages are explanatory.

For the fifth commit, I hope the deprecations are ok with @wlammen and @icecream17 . Note that I had to modify (almost automatically: `MM-PA> expand sb2v` && `MM-PA> min */except sb2v`) a theorem in @wlammen's mathbox and one in @avekens's.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+29-Jul-23 19.8ad    [same]      moved from DAW's mathbox to main set.mm
 29-Jul-23 spc2d     [same]      moved from TA's mathbox to main set.mm
 29-Jul-23 spc2ed    [same]      moved from TA's mathbox to main set.mm
 25-Jul-23 mpnanrd   [same]      moved from ML's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -11821,7 +11821,7 @@
 "rspsbc2" is used by "tratrbVD".
 "sb1ALT" is used by "sb4ALT".
 "sb1ALT" is used by "sb4aALT".
-"sb1ALT" is used by "sb4vALT".
+"sb1ALT" is used by "sb4vOLDALT".
 "sb1ALT" is used by "spsbeALT".
 "sb2ALT" is used by "dfsb2ALT".
 "sb2ALT" is used by "equsb1ALT".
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -17744,6 +17748,14 @@ This axiom scheme is logically redundant (see ~ ax12w ) but is used as an
       CNANAFBGNBEOABCHBCINABJKCBLM $.
   $}
 
+  ${
```
wlammen, line 17751: No enty in changes-set.txt?

benjub: Done.

wlammen (approved):

avekens (commented):
Since I am not deep enough in this topic, I cannot approve that the obsolete theorems are really obsolete and can/should be replaced by the indicated ones. Maybe @wlammen can approve this.

Under this assumption, everything is fine, including the change within my mathbox.

wlammen:
Usually, when you have an OLD theorem, you present a new one replacing it.  Here the OLD theorems like stdpc4v became simply specializations of others (stdpc4) without any advantage at all, like reduced axiom usage.  Their existence became pointless, meaningful only in the presence of the old df-sb.  What should a visitor of Metamath learn from it?  At best, if the comment states it this way, that a redefinition can render specialized theorems obsolete.  If it is really of importance, s/he can have a look at the ALT section, where the OLD theorems are still kept in an appropriate context.

benjub:
> Usually, when you have an OLD theorem, you present a new one replacing it. Here the OLD theorems like stdpc4v became simply specializations of others (stdpc4) without any advantage at all, like reduced axiom usage. Their existence became pointless, meaningful only in the presence of the old df-sb. What should a visitor of Metamath learn from it? At best, if the comment states it this way, that a redefinition can render specialized theorems obsolete. If it is really of importance, s/he can have a look at the ALT section, where the OLD theorems are still kept in an appropriate context.

On a related note: as you wrote, it is currently the case that some theorems with a "obsolete as of" comment do not have an "OLD" suffix to their label, which can be confusing to newcomers. Maybe we should simply add an "OLD" suffix to every such theorem ? What do you think ?

wlammen:
I agree

benjub:
> I agree

Done. Can you check ? And I'll stop here for this MR.

wlammen (approved):

PR #3337, opened 2023-08-01
Two theorems involving apartness

### LINKED ISSUE #3333: Two theorems involving apartness worth proving in set.mm
The math overflow comment at https://mathoverflow.net/a/404116/489586 starts with three statements which it would be desirable to have in iset.mm.

* The first is https://us.metamath.org/ileuni/mulap0.html which we already have
* The second is, in our notation, `( ( A e. CC /\ B e. CC ) /\ ( A =//= B /\ ( A x. B ) = 0 ) ) -> ( A = 0 \/ B = 0 )` . As far as I noticed, iset.mm does not have this or anything especially close
* The third is ``( ( A e. CC /\ B e. CC ) -> ( ( A x. B ) = 0 <-> inf ( { ( abs ` A ) , ( abs ` B ) } , RR , < ) = 0 ) )`` . Here too I'm not aware of anything we already have which is very close.

jkingdon:
Well, three theorems if you count apcon4bid which is just a convenience theorem on top of tightness of apartness.

These are being introduced to contrast with https://us.metamath.org/mpeuni/mul0or.html

Fixes #3333

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1450,7 +1450,7 @@ equivalent (in the absence of excluded middle).</TD>
 
 <tr>
   <td>necon4bid</td>
-  <td>~ necon4biddc</td>
+  <td>~ necon4biddc , ~ apcon4bid</td>
 </tr>
 
 <tr>
@@ -5303,8 +5303,9 @@ zero changes to apart from zero.</TD>
 
 <TR>
 <TD>mul0or , mul0ori , mul0ord</TD>
-<TD><I>none</I></TD>
-<TD>Remark 2.19 of [Geuvers] says that this does not hold in
+<TD>~ mul0eqap</TD>
+<TD>Remark 2.19 of [Geuvers] says that
+` ( A x. B ) = 0 -> ( A = 0 \/ B = 0 ) ` does not hold in
 general and has a counterexample.</TD>
 </TR>
 


iset.mm
```diff
@@ -84758,6 +84758,19 @@ would be equivalent to negated equality and this would follow readily (for
...
 
+  ${
+    apcon4bid.a $e |- ( ph -> A e. CC ) $.
+    apcon4bid.b $e |- ( ph -> B e. CC ) $.
+    apcon4bid.c $e |- ( ph -> C e. CC ) $.
+    apcon4bid.d $e |- ( ph -> D e. CC ) $.
+    apcon4bid.1 $e |- ( ph -> ( A =//= B <-> C =//= D ) ) $.
+    $( Contrapositive law deduction for apartness.  (Contributed by Jim
+       Kingdon, 31-Jul-2023.) $)
+    apcon4bid $p |- ( ph -> ( A = B <-> C = D ) ) $=
```
benjub, line 84769: I guess this can be used to shorten ~abs00 ?

benjub (approved):
I just had a look at this region of iset.mm. From the fundamental ~apti, you know that equality is equivalent to a negative proposition, so in particular
```
( A e. CC /\ B e. CC ) -> STAB A = B
```
which could be compared with ~zdceq

jkingdon:
> I just had a look at this region of iset.mm. From the fundamental ~apti, you know that equality is equivalent to a negative proposition, so in particular
> 
> ```
> ( A e. CC /\ B e. CC ) -> STAB A = B
> ```
> 
> which could be compared with ~zdceq

Interesting observation.

I didn't know this, probably because the literature talks about decidable equality, apartness, and tight apartness. I'm not sure I've seen `STAB` mentioned outside a propositional logic page or two, but sure let's add this.

I'll include this in my follow up as well.

PR #3334, opened 2023-07-31
Interchangeable setvar variables

avekens:
main:
* new theorems ~2bbii, ~equsb3r, ~2sbbid, ~2sbiev added
* ~sbcgfi , ~csbgfi, ~sbceqi  moved from GM's mathbox
* typeset for new symbol "<>" added

AV's mathbox:
* new section "Interchangeable setvar variables" with definition and basic theorems for the property of setvar variables being interchangeable within a wff. Theorems ~ichexmpl* are examples illustrating the application of the new definition.

Background information:
* during my studies about unique ordered/unordered pairs fulfilling a wff I had the idea of interchangeable variables, which I made concrete with this PR. I hope I can use it for further investigations in this field.
* I had to consider two cases in the definition combined in a disjunction. Maybe somebody else has an idea how this could be avoided.
* Transforming the explicit substitutions into implicit substitutions, I got a somehow strange theorem ~ichcircsh2d with an even strager application ~ichexmpl2a: ` $p |- ( ( x e. CC /\ y e. CC ) -> [ a <> b ] ( ( a ^ 2 ) + ( b ^ 2 ) ) = ( c ^ 2 ) ) $=`. I had to assume that the temporary variables have to be complex numbers to use the commutativity of the addition, not the variables used in the expression. Is this just a useless provable theorem, or is this a hint to an imperfect definition? The similar theorem ~ichexmpl2: ` $p |- [ a <> b ] ( ( a e. CC /\ b e. CC /\ c e. CC ) -> ( ( a ^ 2 ) + ( b ^ 2 ) ) = ( c ^ 2 ) ) $=`looks more sensible.

PR #3339, opened 2023-08-01
Draft: add two proofs of FALSE

benjub:
Do not merge. Just checking if it passes CI.

First proof uses yesterday's definition from #3334 and second proof uses simplest possible definition.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18970,6 +18970,8 @@ Proof modification of "f1rhm0to0ALT" is discouraged (161 steps).
 Proof modification of "f1rhm0to0OLD" is discouraged (125 steps).
 Proof modification of "falimfal" is discouraged (6 steps).
 Proof modification of "falimtru" is discouraged (6 steps).
+Proof modification of "false" is discouraged (302 steps).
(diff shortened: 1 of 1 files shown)


benjub:
Ok. So mmj2 detects a problem with df-fake:
```
I-PA-0205 Axiom df-fake has failed the definitional soundness check. Non-set dummy variable found, and no justification theorem is available.
```
but not with df-ich. Note that the version of mmj2 I use, which is the one proposed for download on the metamath website, does not detect it, so that bugfix might be recent.

benjub:
The third definition is detected too:
```
 I-PA-0206 Axiom df-fake has failed the definitional soundness check. Dummy variables [x] are possibly free in the definiendum, and no justification is available.
```
I'll stop here the experimentations (I cannot experiment locally since my version of mmj2 seems to be different: it does not detect any unsoundness). The first definition is still not detected.

digama0:
you should report this on mmj2 repo so I don't forget

benjub:
Reported as https://github.com/digama0/mmj2/issues/70 so closing this PR.

PR #3341, opened 2023-08-02
Impsingle

lesyna:
This pull request has issues that I could not resolve on my own. The theorems have a lot of variables in them, and mmj2 was able to use symbols in Mathboxes JBN and JU to handle them. But I couldn't get Metamath to allow verification because I was relying on the contents of Mathboxes. The other issue has to do with "discouraged" usages. I appreciate advice about how to resolve this.


tirix:
Hi @lesyna !
"Discouraged" should not be an issue, you can follow [the steps described here](https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing#step-5f-if-you-insist-on-doing-something-discouraged-rare) to regenerate the `discouraged` file or use the command `metamath-knife set.mm -D discouraged` if you have metamath-knife (much faster).

To summarize your main problem (also possibly for other reviewers) : `set.mm` has 12 symbols for wff: `ph`, `ps`, `ch`, `th`, `ta`, `et`, `ze`, `si`, `rh`, `mu`, `la`, `ka`, but it looks like you need up to 15 wff variables or more.
You are right now using additional symbols in Jarvin Udandy's Mathbox, which is why the checker complains.

I see Łukasiewicz's axiomatization does not have equivalence, one cannot have "aliases" for formulas, and I imagine this is what leads to a larger number of _wff_ variables.

Solutions are either:
1. we define additional wff variables in the main part of `set.mm` (up to the 24 letters of the greek alphabet),
2. you declare the necessary additional wff in your own Mathbox. (for example, `wlph $f wff lph$.` instead of `wjph $f wff jph$.`),
3. you find a way to cut your theorems into smaller theorems and/or lemmata, each using fewer variables.

For 1., this needs to be discussed and agreed in the broader community. I think originally Norm also saw limiting the number of variables as a way to limit the complexity of the theorems, so not everyone might agree,

Solution 2. is simpler in terms of formalization work, but a bit more complex technically, since you'll have to also define markup for the new symbols, etc. but it's basically just copy/paste and some renaming.

Solution 3. implies some new formalization work. Some of your theorems are named `step18` or `step25`, so I assume you have grouped several steps together, maybe each step taken separately uses fewer variables? 

Maybe overall solution 2. would be the simplest to execute?

set.mm
```diff
@@ -742983,22 +742983,202 @@ to Davis and Putnam (1960).  (Contributed by David A. Wheeler,
...
+
+  $( The shortest single axiom for implicational calculus, due to Lukasiewicz.
+     It is now known to be the _unique_ shortest axiom.  The axiom is proved
+     here starting from { ~ ax-1 , ~ ax-2 and ~ peirce }.  (Contributed by
+     Larry Lesyna and Jeffrey P. Machado, 18-Jul-2023.) $)
   impsingle $p |- ( ( ( ph -> ps ) -> ch )
      -> ( ( ch -> ph ) -> ( th -> ph ) ) ) $=
-    ( wi imim1 peirce a1d syl6 ) ABEZCECAEJAEZDAEJCAFKADABGHI $.
+    ( wi peirce a1d imim2 ax-mp imim1 ) CAEZABEZAEZEZKDAEZEZEZLCEZPEZMOEQMADABF
+    GMOKHIRNEQSELCAJRNPJII $.
```
benjub, line 743018: The existing proof is shorter and is also based on {ax-1, ax-2, peirce}, so no need to modify it. Actually, you should put both discouragement tags.

benjub:
Following on @tirix's remarks: his solution 3. might be better if you find more steps which can naturally serve as intermediate steps. Good candidates are the shorter formulas.

Note that wff dummy variables here could always be chosen among the existing ones, but this would be a bad idea since it hampers readability  (see ~ id and its comment)

lesyna:
Thank you for the comments. I am going to try to implement solution 3, as my goal is to get this section out of my Mathbox and into the main section.  For my investigations that may stay in my Mathbox I now understand how use more wff variables than 12 should I need them.

Larry

lesyna:
Thank you for this shorter proof!

set.mm
lesyna: Please tell me how to properly acknowledge this contribution.

lesyna:
I made a new commit with new theorems all having 12 wff variables or less. I tried the command about "discourages" that tirix suggested, but the commit did not pass verification about "discouraged". I'd like to do a pull request with the latest commit but don't know how to do so.

benjub:
The commands at https://github.com/metamath/set.mm/blob/develop/CONTRIBUTING.md should take care of everything: there is one for formatting `set.mm` and another one for regenerating the `discouraged` file.

lesyna:
All checks are now passing. I used the terminal command from the document that was just sent, which might have been slightly different. Thanks. Now I just need to initiate a  pull request with the latest commit. I am at https://github.com/metamath/set.mm/pull/3341 as I write this. I don't know where to navigate to on the site to do the pull request. Thanks for the advice.

lesyna:
> Following on @tirix's remarks: his solution 3. might be better if you find more steps which can naturally serve as intermediate steps. Good candidates are the shorter formulas.
> 

You were right about the shorter formulas being good candidates. |- ( ph -> ph ) was a real variable hog. :)

benjub:
> All checks are now passing. I used the terminal command from the document that was just sent, which might have been slightly different. Thanks. Now I just need to initiate a pull request with the latest commit. I am at #3341 as I write this. I don't know where to navigate to on the site to do the pull request. Thanks for the advice.

Since this PR #3341 is still open and requests to merge your branch `lesyna:impsingle`, pushing to that branch is fine: your changes appear here.

set.mm
```diff
@@ -742983,23 +742983,196 @@ to Davis and Putnam (1960).  (Contributed by David A. Wheeler,
...
+  $( Derivation of ~ ax-2 from the Tarski-Bernays axiom set { ~ ax-1 ,
+     ~ imim1 , ~ peirce }.  Note that no inference rule other than ~ ax-mp is
+     used in this proof.  This proof establishes a circle of equivalence:  From
+     { ~ impsingle }, { ~ ax-1 , ~ imim1 , ~ peirce } was proved.  From {
+     ~ ax-1 , ~ imim1 , ~ peirce }, { ~ ax-1 , ~ ax-2 and ~ peirce } was
+     proved.  From { ~ ax-1 , ~ ax-2 and ~ peirce }, { ~ impsingle } was
+     proved.  Therefore, the theorems that can be proved by selecting any one
+     of these three distinct axiom sets must be equivalent.  Prover9 and mmj2
+     were used to help constuct this proof.  (Contributed by Larry Lesyna and
+     Jeffrey P. Machado, 1-Aug-2023.)  (New usage is discouraged.) $)
```
benjub, line 743160: It would be safer to add "proof modif discouraged" since an automatic minimization could simply rewrite this proof to use ~ax-2 itself.

benjub (approved):
I'm inclined to merge since it's mathbox-only. It may need a few rewordings before going to the main part, but I'd like to see the proof webpages first.

set.mm
benjub: This was an automatic minimization (done with the "minimize_with *" command of metamath.exe) so it does not need acknowledgment tags. (The new proof is actually pretty close to yours: it only uses the lemma ~syl6 to shorten it somewhat.)

set.mm
lesyna: done

jkingdon (approved):

PR #3343, opened 2023-08-03
Add theorems about simple groups

CatsAreFluffy:
I proved a bunch of theorems about simple groups, including the classification of abelian simple groups ~ ablsimpgprmd. I added them in my mathbox, but I expect a lot of them are appropriate for the main part of set.mm. If anyone wants me to move them to the mainspace, I'm fine with that.


tirix (approved):
Great additions!

set.mm
```diff
@@ -643200,6 +643206,781 @@ base set if and only if the neighborhoods (convergents) of every point
...
+    2nsgsimpgd.1 $e |- B = ( Base ` G ) $.
+    2nsgsimpgd.2 $e |- .0. = ( 0g ` G ) $.
+    2nsgsimpgd.3 $e |- ( ph -> G e. Grp ) $.
+    2nsgsimpgd.4 $e |- ( ph -> -. { .0. } = B ) $.
+    2nsgsimpgd.5 $e |- ( ( ph /\ x e. ( NrmSGrp ` G ) ) ->
+      ( x = { .0. } \/ x = B ) ) $.
+    $( If any normal subgroup of a nontrivial group is either the trivial
+       subgroup or the whole group, the group is simple.  (Contributed by Rohan
+       Ridenour, 3-Aug-2023.) $)
+    2nsgsimpgd $p |- ( ph -> G e. SimpGrp ) $=
```
avekens, line 643679: It would be nice to combine ~simpgnsgd and ~ 2nsgsimpgd to, e.g., `` ( ph -> ( G e. SimpGrp ) <-> ( NrmSGrp ` G ) = { { .0. } , B } ) )`` to prove the statement given in the definition: "A simple group is a group with exactly two normal subgroups." or "A nontrivial group is simple iff it has exactly two normal subgroups".  This theorem could/should be references in the comment of the definition.

set.mm
```diff
@@ -643200,6 +643206,781 @@ base set if and only if the neighborhoods (convergents) of every point
...
+    1gensubg2d.4 $e |- ( ph -> G e. Grp ) $.
+    1gensubg2d.5 $e |- ( ph -> A e. B ) $.
+    $( The set of multiples of an element of a group is a subgroup.  Fewer
+       distinct variable conditions than ~ 1gensubgd .  (Contributed by Rohan
+       Ridenour, 3-Aug-2023.) $)
+    1gensubg2d $p |- ( ph -> ran F e. ( SubGrp ` G ) ) $=
+      ( vm cz cv co cmpt oveq1 cbvmptv eqtri 1gensubgd ) ABCDMFGHIFENEOZBDPZQMN
+      MOZBDPZQJEMNUCUEUBUDBDRSTKLUA $.
+  $}
+
```
avekens, line 643736: The previous theorems are general theorems about groups and therefore should be moved to section  "Groups".

set.mm
```diff
@@ -643200,6 +643206,781 @@ base set if and only if the neighborhoods (convergents) of every point
...
+  ${
+    $d ph y $.  $d ps y $.  $d ch x $.  $d th y $.  $d x A $.  $d y A $.
+    rexlimddvcbv.1 $e |- ( ph -> E. x e. A th ) $.
+    rexlimddvcbv.2 $e |- ( ( ph /\ ( y e. A /\ ch ) ) -> ps ) $.
+    rexlimddvcbv.3 $e |- ( x = y -> ( th <-> ch ) ) $.
+    $( Unpack a restricted existential assumption while changing the variable
+       with implicit substitution.  Similar to ~ rexlimdvaacbv .  The
+       equivalent of theis theorem without the bound variable change is
```
avekens, line 643257: typo: "theis"

set.mm
```diff
@@ -643200,6 +643206,781 @@ base set if and only if the neighborhoods (convergents) of every point
...
+  $}
+
+  ${
+    simpgnsg2d.1 $e |- B = ( Base ` G ) $.
+    simpgnsg2d.2 $e |- .0. = ( 0g ` G ) $.
+    simpgnsg2d.3 $e |- ( ph -> G e. SimpGrp ) $.
+    simpgnsg2d.4 $e |- ( ph -> A e. ( NrmSGrp ` G ) ) $.
+    $( A normal subgroup of a simple group is either the whole group or the
+       trivial subgroup.  (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
+    simpgnsg2d $p |- ( ph -> ( A = { .0. } \/ A = B ) ) $=
```
avekens, line 643663: Maybe call this theorem ~simpgnsgeld, because it referse to the elements of the normal subgroup.

set.mm
```diff
@@ -643200,6 +643206,781 @@ base set if and only if the neighborhoods (convergents) of every point
...
+
+  ${
+    $d ph n $.  $d A n $.
+    1gengend.1 $e |- B = ( Base ` G ) $.
+    1gengend.2 $e |- .x. = ( .g ` G ) $.
+    1gengend.3 $e |- F = ( n e. ZZ |-> ( n .x. A ) ) $.
+    1gengend.4 $e |- ( ph -> A e. B ) $.
+    $( The set of multiples of an element of a group includes that element.
+       (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
+    1gengend $p |- ( ph -> A e. ran F ) $=
```
avekens, line 643702: What is the meaning of the label? "gen" = "generate", "generator"? This should be reflected in the comment somehow.

avekens (approved):
Great result, only minor, nonblockig remarks (see inline comments).
These theorems/definition should remain in the mathbox as long as there is no application/usage by another contributor (according to our rules) Furthermore, sometimes there are several variants of the same theorem which may not all be worth to be moved to main.

set.mm
CatsAreFluffy: ~ 1gengend (now ~ cycsubggend) depends on ~ rr-rspce. The obvious unprefixed name for that theorem is "rspce", but unfortunately ~ rspce already exists and is somewhat different. I'm not sure how to resolve the naming conflict, since ~ rr-rspce is more basic than most of the rspce-type theorems so it seems to be that it should have the name "rspce", but that raises the question of what the theorem now named ~ rspce should be named. (~ elrnmptd also already exists in Glauco Silprandi's mathbox, but ~ rr-elrnmptd is a trivial consequence of that and ~ rr-rspce, so the obvious solution is to move ~ elrnmptd to after ~ elrnmpt and delete ~ rr-elrnmptd.)

set.mm
CatsAreFluffy: Fixed.

set.mm
CatsAreFluffy: Done.

set.mm
CatsAreFluffy: Yes. I've adjusted the comment to reflect this (and also changed the labels to match existing conventions for cyclic subgroups).

set.mm
CatsAreFluffy: Added as ~ simpgnsgbid.

set.mm
avekens: OK, then this can be adjusted later.

avekens:
Line number 643490 became too long (by label change), so that one check failed. Please fix this.

CatsAreFluffy:
Done.

set.mm
benjub: I don't understand @avekens's comment: this theorem is about the normal subgroups themselves, not their elements.

CatsAreFluffy: I agree. Changed to ~ simpgnsgeqd.

set.mm
```diff
@@ -643200,6 +643206,784 @@ base set if and only if the neighborhoods (convergents) of every point
...
+    ablsimpgd.2 $e |- ( ph -> G e. Abel ) $.
+    $( An abelian group is simple if and only if its order is prime.
+       (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
+    ablsimpgd $p |- ( ph -> ( G e. SimpGrp <-> ( # ` B ) e. Prime ) ) $=
+      ( csimpg wcel chash cprime wa cabl adantr simpr ablsimpgprmd cgrp ablgrpd
+      cfv prmgrpsimpgd impbida ) ACFGZBHQIGZATJBCDACKGTELATMNAUAJBCDACOGUAACEPL
+      AUAMRS $.
+  $}
+
+$( (End of Rohan Ridenour's mathbox.) $)
```
benjub, line 643984: The final result would probably be something like
```
|- ( G e. ( AbelGrp i^i SimpGrp ) <-> ( E. p e. Prime G GrpIso ( Grp ` ( Z/nZ ` p ) ) ) )
```
Where:
* `A GrpIso B` means that the groups A and B are isomorphic
* `Z/nZ : ZZ --> Ring` is the function that gives the ring Z/nZ.
*  `Grp` is the function giving the underlying group of a richer structure (like a ring)

(I do not know if these definitions are already in set.mm)

Any plans of proving this ?

CatsAreFluffy: `GrpIso` and `Z/nZ` are defined in set.mm at ~ df-gim and ~ df-zn respectively, and rings are groups (~ ringgrp) so there's no need for a conversion function between them. The only significant theorem I'd need to prove to prove this is that abelianness respects group isomorphism, which seems like it shouldn't be too hard. I'd probably do it in another PR though.

set.mm
avekens: The elements of the set of normal subgroups are the normal subgroups -> "eq" is OK.

benjub (approved):
As for the theorems rr-rspce to rr-elrnmpt2d, they are variants of existing theorems, and they might form a useful set of lemmas when using a given proof strategy. The choice of a library of lemmas impacts the choice of a proving strategy (e.g., at a simpler stage, deduction type versus closed forms, or also explicit versus implicit substitution, or manipulating biconditionals as long as possible before using ~biimpi versus first using ~biimpi, etc.). I do not know if we want to generally stick to a single proof strategy, or permit various strategies, at the cost of a flourishing of very similar versions of intermediate lemmas. @digama0 could give interesting comments on proof strategies.

CatsAreFluffy:
Merge?

avekens:
Merged, since there are three approvals and no requests for changes.

PR #3345, opened 2023-08-04
Interchangeable setvar variables (2)

avekens:
Follow up of PR #3334 (review remarks already considered)

main:
* new theorems ~2bbii, ~equsb3r, ~2sbbid, ~2sbiev added *~sbcgfi , ~csbgfi, ~sbceqi moved from GM's mathbox
* typeset for new symbol "<>" added

AV's mathbox:
* new section "Interchangeable setvar variables" with definition and basic theorems for the property of setvar variables being interchangeable within a wff. Theorems ~ichexmpl* are examples illustrating the application of the new definition.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+31-Jul-23 sbcgfi    [same]      moved from GM's mathbox to main set.mm
+31-Jul-23 csbgfi    [same]      moved from GM's mathbox to main set.mm
+31-Jul-23 sbceqi    [same]      moved from GM's mathbox to main set.mm
 29-Jul-23 19.8ad    [same]      moved from DAW's mathbox to main set.mm
 29-Jul-23 pnfnre2   [same]      moved from GS's mathbox to main set.mm
 29-Jul-23 spc2d     [same]      moved from TA's mathbox to main set.mm


avekens:
In #3334 @benjub proposed:

> As for the definition: I think the disjunct x = y should not be there. You probably meant A. x x = y (which is equivalent to A. x A. y x = y, so x and y are bound like in the LHS) but this is not needed. It looks like
> 
> `$a |- ( [ x <> y ] ph <->  A. a A. b ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) ) $.`
> 
> or
> 
> `$a |- (  [ x <> y ] ph <->  A. a ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) ) $.`
> 
> should work?

I tried it already without disjunct x = y, but then I could not proof ~ichid and ~eqich, which I think should be valid theoems.
But your 2nd variant looks interesting. I'll have a try with this (therefore I set this PR to "draft".

avekens:
I experimented with both variants of the definition proposed by @benjub, both have advantages and disadvantages. Therefore, it would be very helpful to show that the varaints are equivalent:

`( A. a ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) <-> A. a A. b ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) ) `

Although it seems to be obvious, I was not able to prove it yet. The main problems are the distinct variables restrictions resp. nonfreeness requirements (e.g. of ~sbied/sbiedv). Does anybody has an idea/hint? The implication from the left to the right would be sufficient for the moment...

benjub:
> `( A. a ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) <-> A. a A. b ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) )`

Indeed you need the disjunct `A. x x = y` on the RHS since when x and y are the same variable, the result of `[ a / x ] [ b / y ] ph` has only `b`'s while the result of the other one has only `a`'s). Also, it will probably be easier if you choose a third fresh variable `c` on the LHS.

Probably you'll have to prove the two parts
* `A. x x = y -> ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph )`: in that case `[ y / x ]` is the identity, so you are left with `[ x / a ] [ a / y ] ph` which cancel out to `ph`
* `-. A. x x = y -> ( A. c ( ph <-> [ x / c ] [ y / x ] [ c / y ] ph ) <-> A. a A. b ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) )`. With the DV conditions, the simpler definition sb6 can be used, and look at the `sb*com*` theorems too. Look at the result of `SEARCH * "-. A. ? ? = ? -> $* [ ? / ? ] ??" / JOIN` to get potentially useful theorems.

By the way, and independently of the above: one should have `F/ a [ x / a ] [ y / x ] [ a / y ] ph` when `a` is fresh.

wlammen:
see #3347

avekens:
I'll wait with this PR until #3347 is finished/merged.

avekens:
In #3347 @wlammen said that `(  [ x <> y ] ph <-> ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) ) ` (without `A. a`) would be sufficient (compared with ` (  [ x <> y ] ph <->  A. a ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) ) $.` as proposed by @benjub), since `a` is not free in `( ph <-> [ x / a ] [ y / x ] [ a / y ] ph )`. Therefore, I will take this as definition, OK?

avekens:
With the help of (the new) theorem 

    $d a x $.
    sbbib.a $e |- F/ a ph $.
    sbbib.x $e |- F/ x ps $.
    $( Reversal of substitution. (Contributed by AV, 6-Aug-2023.) $)
    sbbib $p |- ( A. a ( [ a / x ] ph <-> ps )
              <-> A. x ( ph <-> [ x / a ] ps ) ) $=
      ( wsb wb wal nfs1v nfbi nfal sbid2 spsbbi syl5bbr alrimi syl6bb impbii )
      ACDGZBHZDIZABDCGZHZCIZUAUCCTCDSBCACDJFKLASDCGUAUBADCEMSBDCNOPUDTDUCDCAUBD
      EBDCJKLUDSUBCDGBAUBCDNBCDFMQPR $.

I could easily prove

     $d a b ph $.  $d x y z ph $.  $d a b x y $.
    dfich2ai $p |- ( A. x A. y ( [ x / z ] [ y / x ] [ z / y ] ph <-> ph )
                 -> ( [ a / x ] [ b / y ] ph  <-> [ b / x ] [ a / y ] ph ) ) $=
      ( wsb wb wal sbco4 bicomi bibi1i 2albii nfs1v nfv sbbib albii alcom sylbi
      nfsb bitri 2sp ) ACDGBCGDBGZAHZCIBIACFGZBEGZFBGZECGZAHZCIZBIZUFACEGZBFGHZ
      UDUIBCUCUHAUHUCABCDFEJKLMUKUGULHZEIZBIZUMUJUOBUGAECUFFBCUEBECACFNTTAEOPQU
      PUMFIZEIZUMUPUNBIZEIURUNBERUSUQEUFULFBUEBENULFOPQUAUMEFUBSSS $.

And together with @wlammen's wl-ich that both definitions are equivalent:

    dfich2 $p |- ( A. x A. y [ x <> y ] ph <-> A. a A. b
                   ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) ) $=
      ( vz wich wal wsb wb df-ich 2albii nfv dfich2ai alrimi nfbi nfal dfich2bi
      nfs1v nfsb impbii bitri ) ABCGZCHBHACFIBCIFBIAJZCHZBHZACEIZBDIZACDIZBEIZJ
      ZEHZDHZUCUDBCABCFKLUFUMUFULDUFDMUFUKEUFEMABCFDENOOUMUEBULBDUKBEUHUJBUGBDS
      UIBESPQQUMUDCULCDUKCEUHUJCUGBDCACESTUIBECACDSTPQQABCFDEROOUAUB $.

What bothers me a little bit are the universal quantifiers `A. x A. y` - are they really necessary?

wlammen:
Why?  The RHS does not depend on x, y, so must the LHS.

benjub:
x and y are bound in `[ x <> y ] ph` so indeed you can remove these quantifiers, but I'm surprised you needed them in the first place. Theorem dfich2ai looks strange: it looks like there are too many DV conditions ~and the antecedent looks too strong~, so I'm surprised it is useful to prove dfich2. But I haven't looked at your proofs, so I'll wait for the webpages.

**edit:** just to be sure: for the definition, you finally chose
```
$a |- ( [ x <> y ] ph <-> A. x A. y ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) ) $.
```
Correct ?

avekens:
> x and y are bound in `[ x <> y ] ph` so indeed you can remove these quantifiers, but I'm surprised you needed them in the first place. Theorem dfich2ai looks strange: it looks like there are too many DV conditions ~and the antecedent looks too strong~, so I'm surprised it is useful to prove dfich2. But I haven't looked at your proofs, so I'll wait for the webpages.
> 
> **edit:** just to be sure: for the definition, you finally chose
> 
> ```
> $a |- ( [ x <> y ] ph <-> A. x A. y ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) ) $.
> ```
> 
> Correct ?

No I have chosen df-ich $a |- ( [ x <> y ] ph <-> ( [ x / a ] [ y / x ] [ a / y ] ph <-> ph ) ) without quantifiers (analogous to your proposal `(  [ x <> y ] ph <->  A. a ( ph <-> [ x / a ] [ y / x ] [ a / y ] ph ) )`, but without `A. a`, because `a` is not free. But if it is a requirement for definitions not to have free variables (at the RHS) (see also https://github.com/digama0/mmj2/issues/70 - but where is it written that this is a requirement?), then I have to switch to your variant. And then the ` A. x A. y`will disappear from dfich2 (because they are moved into the definition).

avekens:
I had a look at the Metamath book, section 4.5.3 "Adding Constraints on Definitions": The section ends with "6. Every dummy variable must be bound", but `x` and `y` are no dummy variables (like `a `, or `a` and `b` in my first version) ...

benjub:
Defining without the quantifiers `A. x A. y` wouldn't be unsound, but it wouldn't be the definition you probably mean. I think that the intended meaning for `[ x <> y ] ph` is the one I wrote above, namely
```
$a |- ( [ x <> y ] ph <-> A. x A. y ( ph <-> [ x / z ] [ y / x ] [ z / y ] ph ) ) $.
```
with the quantifiers, and not `( ph <-> [ x / z ] [ y / x ] [ z / y ] ph )`.

wlammen:
If [ x <> y ] ph (swapping substitution) wasn't thought of as holding everywhere, but instead was true for some (x, y) while it did not hold for other instances, this would be along the lines of normal substitution, and justified its similar notation. We don't expect [ x / y ] ph be true all the time, too.  So adding the quantifiers, makes [ x <> y ] be a property of ph, rather than an operation applied to ph, creating a modified version of it.  This would distinguish it from substitution then.  What are the primary goals for your definition?

avekens:
> If [ x <> y ] ph (swapping substitution) wasn't thought of as holding everywhere, but instead was true for some (x, y) while it did not hold for other instances, this would be along the lines of normal substitution, and justified its similar notation. We don't expect [ x / y ] ph be true all the time, too. So adding the quantifiers, makes [ x <> y ] be a property of ph, rather than an operation applied to ph, creating a modified version of it. This would distinguish it from substitution then. What are the primary goals for your definition?

My intention of the definition is to define a property ("interchangeability") and not an operation ("to interchange the variables"). Then, as you said, the quantifiers must be added, I think.

set.mm
```diff
@@ -719932,6 +719985,169 @@ valid cases ( ` (/) ` is the last symbol) and invalid cases ( ` (/) `
...
+    ichcircshi $p |- [ x <> y ] ph $=
+      ( wich wsb wb wal weq bicomd equcoms sbievw 2sbbii sbbii 3bitri df-ich
+      gen2 mpbir ) ADEJAEFKZDEKFDKZALZEMDMUFDEUECDEKZFDKBFDKAUDCFDEDACEFACLFEFE
+      NCAIOPQRUGBFDCBDECBLEDEDNBCHOPQSBAFDBALDFDFNABGOPQTUBADEFUAUC $.
+  $}
+
+  ${
+    $d a b ph $.  $d z ph $.  $d a b x y $.  $d x y z $.
+    $( Definition of the propery of a wff ` ph ` that the setvar variables
+       ` x ` and ` y ` are interchangeable implies alternate definition.
```
wlammen, line 720038: implies an alternate definition

set.mm
```diff
@@ -719932,6 +719985,169 @@ valid cases ( ` (/) ` is the last symbol) and invalid cases ( ` (/) `
...
+                 -> ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) ) $=
+      ( wsb wb wal sbco4 bicomi bibi1i 2albii nfs1v nfv sbbib albii alcom sylbi
+      nfsb bitri 2sp ) ACDGBCGDBGZAHZCIBIACFGZBEGZFBGZECGZAHZCIZBIZUFACEGZBFGHZ
+      UDUIBCUCUHAUHUCABCDFEJKLMUKUGULHZEIZBIZUMUJUOBUGAECUFFBCUEBECACFNTTAEOPQU
+      PUMFIZEIZUMUPUNBIZEIURUNBERUSUQEUFULFBUEBENULFOPQUAUMEFUBSSS $.
+
+    $d a b z $.
+    $( Alternate definition of the propery of a wff ` ph ` that the setvar
+       variables ` x ` and ` y ` are interchangeable implies definition (see
+       also ~ wl-ich ).  (Contributed by AV, 6-Aug-2023.) $)
```
wlammen, line 720050: wl-ich is going to be deleted sooner or later.  It served as an ad hoc solution to your public request, but is of no interest on its own, I think.

changes-set.txt
```diff
@@ -27,6 +27,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+31-Jul-23 sbcgfi    [same]      moved from GM's mathbox to main set.mm
```
wlammen, line 30: Why did you move them to main?

avekens: I had used them in my mathbox in my first commit, but it seems that the corresponding proofs are removed/have changed, so that they are not used anymore.

set.mm
avekens: You can remove the reference from the comment as soon as you delete the theorem. I would like, however to use your proof, because its shorter. Shall I take/move it from your mathbox (together with the required lemmata?) Of course you will still be the contributor.

set.mm
avekens: OK,  I will fix it with my next commit.

set.mm
wlammen: ok

wlammen (approved):

tirix:
> My intention of the definition is to define a property ("interchangeability") and not an operation ("to interchange the variables").

Then as @wlammen mentioned, maybe we could find a different syntax than the square brackets?
They are used in `[ x / y ] ph` as an operation to transform the formula.

tirix:
This PR was approved yesterday, can we merge it?

avekens:
> > My intention of the definition is to define a property ("interchangeability") and not an operation ("to interchange the variables").
> 
> Then as @wlammen mentioned, maybe we could find a different syntax than the square brackets? They are used in `[ x / y ] ph` as an operation to transform the formula.

If there is any suggestion, I can consider it. But I have no better idea for an alternate syntax.

avekens:
I just adjusted the comments as proposed by @wlammen . Therefore, this PR can be merged now.

wlammen (approved):

PR #3347, opened 2023-08-05
relate expressions describing interchangeability

wlammen:
1. Solves request in #3345 .  One direction.  wl-ich. (Mathbox only)
2. Interchangeability in a non trivial context requires x, y be distinct (mathbox)
3. Delete a duplicate of sb4b from my mathbox
4. Prove sbelx without ax-13
5. Remove outdated OLD theorems

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -10172,7 +10172,6 @@
 "numclwlk2lem2fvOLD" is used by "numclwlk2lem2f1oOLD".
 "numclwwlk1lem2f1OLD" is used by "numclwwlk1lem2f1oOLD".
 "numclwwlk1lem2f1oOLD" is used by "numclwwlk1lem2OLD".
-"numclwwlk1lem2f1oOLD" is used by "numclwwlk1lem2OLDOLD".
 "numclwwlk1lem2fOLD" is used by "numclwwlk1lem2f1OLD".
 "numclwwlk1lem2fOLD" is used by "numclwwlk1lem2foOLD".
 "numclwwlk1lem2foOLD" is used by "numclwwlk1lem2f1oOLD".
@@ -13224,21 +13223,11 @@
 "wl-mps" is used by "wl-syls1".
 "wl-sbcom2d-lem1" is used by "wl-sbcom2d".
 "wl-sbcom2d-lem2" is used by "wl-sbcom2d".
-"wlknwwlksnbijOLD" is used by "wlksnwwlknvbijOLD".
-"wlknwwlksnfunOLD" is used by "wlknwwlksninjOLD".
-"wlknwwlksnfunOLD" is used by "wlknwwlksnsurOLD".
-"wlknwwlksninjOLD" is used by "wlknwwlksnbijOLD".
-"wlknwwlksnsurOLD" is used by "wlknwwlksnbijOLD".
 "wlkresOLD" is used by "eupthresOLD".
 "wlkresOLD" is used by "trlresOLD".
 "wlkreslem0OLD" is used by "trlreslemOLD".
 "wlkreslem0OLD" is used by "wlkresOLD".
 "wlkreslemOLD" is used by "wlkresOLD".
-"wlkwwlkbijOLD" is used by "wlkwwlkbij2OLD".
-"wlkwwlkfunOLD" is used by "wlkwwlkinjOLD".
-"wlkwwlkfunOLD" is used by "wlkwwlksurOLD".
-"wlkwwlkinjOLD" is used by "wlkwwlkbijOLD".
-"wlkwwlksurOLD" is used by "wlkwwlkbijOLD".
 "wrdcctswrdOLD" is used by "2clwwlk2clwwlkOLD".
 "wrdcctswrdOLD" is used by "lencctswrdOLD".
 "wrdnfiOLD" is used by "clwwlknfiOLD".
@@ -16062,7 +16051,6 @@ New usage of "imaelshi" is discouraged (1 uses).
 New usage of "imbi12VD" is discouraged (0 uses).
 New usage of "imbi13" is discouraged (2 uses).
 New usage of "imbi13VD" is discouraged (0 uses).
-New usage of "imp5aOLD" is discouraged (0 uses).
 New usage of "impcomdOLD" is discouraged (0 uses).
 New usage of "impexpd" is discouraged (1 uses).
 New usage of "impexpdcom" is discouraged (0 uses).
@@ -16871,9 +16859,8 @@ New usage of "numclwlk2lem2f1oOLD" is discouraged (1 uses).
 New usage of "numclwlk2lem2fOLD" is discouraged (1 uses).
 New usage of "numclwlk2lem2fvOLD" is discouraged (1 uses).
 New usage of "numclwwlk1lem2OLD" is discouraged (0 uses).
-New usage of "numclwwlk1lem2OLDOLD" is discouraged (0 uses).
 New usage of "numclwwlk1lem2f1OLD" is discouraged (1 uses).
-New usage of "numclwwlk1lem2f1oOLD" is discouraged (2 uses).
+New usage of "numclwwlk1lem2f1oOLD" is discouraged (1 uses).
 New usage of "numclwwlk1lem2fOLD" is discouraged (2 uses).
 New usage of "numclwwlk1lem2foOLD" is discouraged (1 uses).
 New usage of "numclwwlk1lem2foaOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -532760,6 +532528,73 @@ dependency is expressed in our hypothesis (called implicit
       CDBJABCDKL $.
   $}
 
+  ${
+    $d t x $.  $d x ph $.
+    $( Reverting substitution yields the original expression.  (Contributed by
+       Wolf Lammen, 5-Aug-2023.) $)
+    wl-ich-lem1 $p |- ( [ t / x ] [ x / t ] ph <-> ph ) $=
```
avekens, line 532535: Why not use sbid2v? It does not require t and x to be distinct.

set.mm
```diff
@@ -532760,6 +532528,73 @@ dependency is expressed in our hypothesis (called implicit
...
+  ${
+    $d a x $.  $d a b y $.  $d a b ph $.
+    $( Reverting two substitutions in succession.  (Contributed by Wolf Lammen,
+       5-Aug-2023.) $)
+    wl-ich-lem2 $p |- ( [ y / b ] [ x / a ] [ a / x ] [ b / y ] ph <-> ph ) $=
+      ( wsb wl-ich-lem1 sbbii bitri ) ACEFZBDFDBFZECFJECFAKJECJDBGHAECGI $.
+  $}
+
+  $( ~ spsbim applied twice.  (Contributed by Wolf Lammen, 5-Aug-2023.) $)
+  wl-ich-lem3 $p |- ( A. a A. b ( ph <-> ps ) ->
```
avekens, line 532548: It is ~spsbbi applied twice, isn't it? Why not use ~spsbbi (twice) within the proof?

set.mm
wlammen: Because I did not notice it, when looking for variants of sbco2

set.mm
wlammen: You're right.

set.mm
```diff
@@ -532760,6 +532528,73 @@ dependency is expressed in our hypothesis (called implicit
...
+
+  ${
+    $d a b x $.  $d a b y $.  $d a b ph $.
+    $( Lemma.  To be meaningful, interchangeability should have ` x ` free in
+       ` ph ` (and for symmetry reasons ` y ` , too).  This requires ` x ` and
+       ` y ` be distinct, and the universe to have more than one element.
+       (Contributed by Wolf Lammen, 5-Aug-2023.) $)
+    wl-ich-lem6 $p |- ( ( A. x x = y /\
+          A. a A. b ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) )
+          -> F/ x ph ) $=
```
avekens, line 532591: According to @benjub's remark, the other variant to define interchangeability is `( A. x x = y \/ A. a A. b ( [ a / x ] [ b / y ] ph <-> [ b / x ] [ a / y ] ph ) )` (disjunction instead of conjunction). Then, however, `F/ x ph` won't follow necessarily.

set.mm
```diff
@@ -532760,6 +532528,73 @@ dependency is expressed in our hypothesis (called implicit
...
+        ( ph <-> [ y / b ] [ x / a ] [ b / x ] [ a / y ] ph ) ) $=
+      ( wsb wb wal wl-ich-lem2 wl-ich-lem3 syl5bbr ) AACEFBDFZDBFECFLACDFBEFZGE
+      HDHMDBFECFABCDEILMBCDEJK $.
+  $}
+
+  ${
+    $d a b x $.  $d a b y $.  $d a b ph $.
+    $( Infer one criterium for interchangeability from another.  (Contributed
+       by Wolf Lammen, 5-Aug-2023.) $)
+    wl-ich $p |- ( A. a A. b ( [ a / x ] [ b / y ] ph
```
avekens, line 532567: That's a great result, but I wonder if `a` in the conclusion can be changed to `z` (or any other variable not used in the antecedent), as also proposed by @benjub .

set.mm
wlammen: This is a bit stricter, since it rules out cases, instead of allowing exceptional ones, as the disjunction does.

set.mm
wlammen: Of course.  I did not read @benjub s comments before, but did an analysis on my own.

avekens:
Thank you very much, @wlammen , for these very helpful theorems. It would be great if we can also proof the other direction to show that both variants equivalently define interchangeability. Do you plan to investigate in this further? If not, I will stop my investigations until this PR is merged, and restart them based on your theorems. If we get a common result, we could move the correspondig theorems to main then (to avoid duplicates in our mathboxes).

wlammen:
I'll have a look at the other direction tomorrow.

set.mm
```diff
@@ -18340,6 +18331,14 @@ Converse of the inference rule of (universal) generalization ~ ax-gen .
   sbid $p |- ( [ x / x ] ph <-> ph ) $=
     ( weq wsb wb equid sbequ12r ax-mp ) BBCABBDAEBFABBGH $.
 
+  ${
+    $d t x $.  $d x ph $.
+    $( Reverting substitution yields the original expression.  (Contributed by
+       Wolf Lammen, 5-Aug-2023.) $)
+    sbid2vv $p |- ( [ t / x ] [ x / t ] ph <-> ph ) $=
```
benjub, line 18338: I think it should be labeled sbid2vw (like equsalvw, etc.). We should probably review these labelings in the future, but let's be consistent and for the moment stick with the existing pattern. I would also add a link to sbid2v in the comment, e.g., the usual: "Version of ~sbid2v with a DV condition requiring fewer axioms."

avekens:
I also was succeessful with proving the one direction of the equivalence of the definitions of interchangeability:

    $d a b z ph $.  $d a b x y z $.
    dfich2i $p |- ( A. a A. b ( [ a / x ] [ b / y ] ph
                                <-> [ b / x ] [ a / y ] ph ) 
                   -> A. z ( [ x / z ] [ y / x ] [ z / y ] ph <-> ph ) ) $=
      ( wsb wb wal nfa1 nfa2 2sp sbcom2 syl6bb sbbid sbco4 sbid2v bitri 3bitr3g
      alrimiv ) ACFGBEGZACEGBFGZHZFIZEIZACDGBCGDBGZAHDUEUAFBGZECGABFGZFBGZCEGZE
      CGZUFAUEUGUJECUDEJUEUGUHCEGZFBGUJUEUAULFBUCFEKUEUAUBULUCEFLACEBFMNOUHCEFB
      MNOABCDFEPUKUIAUIECQAFBQRST $.

It is similar to wl-ich (both proofs are based on sbco4), and I will compare it for details after this PR is merged.

wlammen:
@avekens The quantifier A. z is not necessary, since z is not free in [ x / z ] [ y / x ] [ z / y ] ph <-> ph

set.mm
```diff
@@ -532760,6 +532537,78 @@ dependency is expressed in our hypothesis (called implicit
...
+      ( wsb wb wal wl-2spsbbi sbid2vw sbbii bitri sbco4 3bitr3g ) ACEFZBDFZACDF
+      ZBEFZGEHDHPDBFZECFZRDBFECFAQBCFDBFPRBCDEITOECFASOECODBJKAECJLABCDDEMN $.
+  $}
+
+  ${
+    $d b x $.
+    $( Lemma.  Simplify substitution expression when two particular variables
+       are the same.  (Contributed by Wolf Lammen, 5-Aug-2023.) $)
+    wl-ich-lem1 $p |- ( A. x x = y ->
+                         ( [ a / x ] [ b / y ] ph <-> [ b / x ] ph ) ) $=
```
benjub, line 532562: `[ a / x ] [ b / y ] ph <-> [ b / y ] ph` seems more natural: "When substituting successively for two always equal variables, the second substitution has no effect." I think this result is interesting in its own, and not only as a lemma.

Also: the use of ~sbbid in your proof makes me think that the proof would be more direct with antecedent `A. y y = x`, or, renaming variables:
```
|- ( A. x x = y -> ( [ b / y ] [ a / x ] ph <-> [ a / x ] ph ) )
```

set.mm
```diff
@@ -532760,6 +532537,78 @@ dependency is expressed in our hypothesis (called implicit
...
+      ( weq wal wsb wb wl-ich-lem2 wl-ich-lem3 ) BCFBGACEHBDHACDHBEHIABDEABCDEJ
+      K $.
+
+    $( Lemma.  To be meaningful, interchangeability should have ` y ` free in
+       ` ph ` .  This requires ` x ` and ` y ` be distinct, and the universe to
+       have more than one element.  Companion theorem is ~ wl-ich-lem5 .
+       (Contributed by Wolf Lammen, 6-Aug-2023.) $)
+    wl-ich-lem5 $p |- ( A. a A. b ( [ a / x ] [ b / y ] ph
+                                    <-> [ b / x ] [ a / y ] ph )
+                      -> ( -. F/ y ph -> -. A. x x = y ) ) $=
```
benjub, line 532607: For wl-ich-lem3/4/5, the positive forms (e.g., `( A. x x = y -> F/ y ph )`) seem more natural.

wlammen: I deliberately chose this form, because you usually (want to) have x free in ph, when addressing interchangeability.

set.mm
wlammen: I am pretty sure there is headroom for optimization.  I simply have not unlimited time.  If this is moved to main, anyone is free to improve on these theorems

avekens (approved):
I think this PR is ready to be merged now. Further improvements/additions can be performed in separate PRs.

set.mm
wlammen: Did not have the time to look at this detail.

set.mm
wlammen: There is an extension under way.

PR #3348, opened 2023-08-06
Add sbthac to set.mm

### LINKED ISSUE #3325: Can we confirm or refute "The [Schroeder-Bernstein] theorem can be proved more easily if we allow AC"?
The genesis of this issue is https://mathstodon.xyz/@andrejbauer/110792965005628884 which says "I don't know, the [proof of Schroeder-Bernstein] with excluded middle isn't particularly involved. I am not even aware of one that uses choice directly."

The context, for metamath, is the statement "The theorem can be proved more easily if we allow AC" at https://us.metamath.org/mpeuni/sbth.html 

I attempted to do a literature search of sorts and I don't think any of the proofs I found use choice, but I'm not sure I know enough to be sure (either where to look or how to detect choice):

* [Cantor-Bernstein-Schröder Theorem](https://proofwiki.org/wiki/Cantor-Bernstein-Schr%C3%B6der_Theorem) in proofwiki (6 proofs)
* [Knaster–Tarski theorem](https://en.wikipedia.org/wiki/Knaster%E2%80%93Tarski_theorem) in Wikipedia ("The Knaster–Tarski theorem can be used to give a simple proof of the Cantor–Bernstein–Schroeder theorem.")
* [Schröder–Bernstein theorem](https://en.wikipedia.org/wiki/Schr%C3%B6der%E2%80%93Bernstein_theorem) in Wikipedia

Norm must have had something in mind when he wrote that (assuming it was Norm and not someone who emailed him the text), but I'm not sure it will be easy to figure out what. Does anyone have a copy of Suppes (which Norm cited), and if so does it discuss choice? Of course "more easily" and "quite difficult" are hard to nail down, and this might just mean relative to other theorems which existed in set.mm in 1998 but even so are there even any proofs which use choice?

Interestingly, the comment about the axiom of choice was not there in 1998 when `sbth` was first added (see https://github.com/metamath/set.mm/blob/db5c4a1d938a13be1685998fb5e6cc088fb87810/set.mm ) but instead was added in e89b9b5780698853be9241a6de13b6ed60087cfe (that is to say, some time between 2001 and 2004).

jkingdon:
This is a version of the usual proof of the Schroeder-Berstein theorem in terms of the axiom of choice.

Adjust the comment of sbth to refer to this.

Fixes #3325

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17599,6 +17599,7 @@ New usage of "sbnvOLD" is discouraged (1 uses).
 New usage of "sbrimALT" is discouraged (1 uses).
 New usage of "sbtALT" is discouraged (0 uses).
 New usage of "sbtT" is discouraged (0 uses).
+New usage of "sbthac" is discouraged (0 uses).
 New usage of "sbtvOLD" is discouraged (0 uses).
 New usage of "scmateALT" is discouraged (0 uses).
 New usage of "seq1hcau" is discouraged (0 uses).
@@ -19720,6 +19721,7 @@ Proof modification of "sbrimALT" is discouraged (41 steps).
 Proof modification of "sbsbc" is discouraged (21 steps).
 Proof modification of "sbtALT" is discouraged (12 steps).
 Proof modification of "sbtT" is discouraged (7 steps).
+Proof modification of "sbthac" is discouraged (92 steps).
 Proof modification of "sbtvOLD" is discouraged (31 steps).
 Proof modification of "scmateALT" is discouraged (236 steps).
 Proof modification of "sgrpplusgaopALT" is discouraged (82 steps).


benjub:
This looks a bit like a proof of the Schroeder--Bernstein theorem from the Schroeder--Bernstein theorem. I do not know that field well, so don't take anything for granted. Since the Schroeder--Bernstein theorem states one of the properties of a linear ordering, it may be clearer to say that a simpler special case suffices? E.g.,

The axiom of choice permits to deduce the Schroeder--Bernstein theorem from its special case when at least one set is an ordinal, and that special case is easier to prove, using ordinal trichotomy.

At first, I was going to write "special case when both sets are cardinals (which holds basically by definition)" but then, the proof in the second paragraph of @digama0's comment https://github.com/metamath/set.mm/issues/3325#issuecomment-1657119359 uses ordinal collapse, hence my formulation above. Is this interpretation correct?

A formalization of Mario's second paragraph proof that `A ~<_ B -> A C_ B` for cardinals (actually, ordinals) from only elementary facts would be nice too.

jkingdon:
> The axiom of choice permits to deduce the Schroeder--Bernstein theorem from its special case when at least one set is an ordinal

I guess, but given the large number of intermediate theorems between sbth and sbthac , I'm not sure sbthac can be taken as a demonstration that the special case is sufficient.

If we want to get deeper into the history of this, it would require more reading of Cantor and Dedekind and trying to figure out what they proved, from what (and even what they just assumed without proving or at least not writing down the proof). If we want to get deeper into what sbthac really relies on (or could rely on with the kind of restructuring of cardinal theorems discussed in #3325 comments), well that is the part which seems like a lot of work.

As for what to do about this?
* option 1: Some version of "handwave over the circular reasoning issue" (for example, what this pull request as currently written does)
* option 2: get rid of sbthac and merely reword the sbth comment. For example, change "The theorem can be proved more easily if we allow AC." to "The theorem can also be proved from the axiom of choice and the linear order of the cardinal numbers, but our development does not provide the linear order of cardinal numbers until much later and in ways that depend on Schroeder-Bernstein".

Although there would also appear to be an option of developing more theorems, such as alternate proofs of  `A ~<_ B -> A C_ B` and whatever else would be needed, and I'm not exactly opposed if someone wants to take that on, at least to my thinking it would be nice to pick either option1 or option 2 as a short term improvement before anything like that happens.

benjub:
I think it would be nice to formalize Mario's proof sketch, but since time is of the essence, I agree for a short term solution, and option 1 is better since it gives a "placeholder" for such a theorem. So now, it remains to add to the comment of this PR an argument that there is no circularity, even if it's only handwaving. My proposal is the paragraph I emphasized in my previous comment. Is that a correct way to explain that non-circularity ? Or should we add to it / replace it by Mario's second pagraph, more or less verbatim ? At least there should be *some* explanation. (The sentence "proving Schroeder-Bernstein via linear ordering of the cardinal numbers" does not suffice as an explanation of the non-circularity.)

jkingdon:
> So now, it remains to add to the comment of this PR an argument that there is no circularity, even if it's only handwaving.

I'm open to some level of elaboration but am trying to understand the suggestion. I guess the idea is to say "this proof is circular" but then outline how a non-circular proof might proceed?

I guess the desire is to include some or all of:

```
under AC, every set is equinumerous to an ordinal (its cardinality), and ordinals are totally ordered more or less by definition. There are still a few things to show after you reduce to cardinals, in particular that A ~<_ B implies A C_ B for cardinals, but they are a lot easier than the usual S-B proof.

If A ~<_ B and WLOG assuming B C_ A by ordinal trichotomy, then A is equinumerous to a subset C C_ B. By ordinal collapse there is an order isomorphism from C to an ordinal D, which implies D C_ A. But A is a cardinal and D ~~ A, so D = C = B = A, hence A C_ B.
```

but I was having trouble editing that into text to put into the comment for `sbthac`.

I don't know, all this is pushing me towards option 2, or maybe not even changing anything at all. Citing informal proofs which have been published elsewhere is one thing, but trying to describe them ourselves seems..... like not the strength of the project.

digama0:
I will note that, as a proof sketch, even I am not sure if it can be completed non-circularly without actually trying it. In a new context, who knows whether an unforeseen issue will crop up. So I would be inclined not to make any such claim, or at least qualify it carefully, if we aren't actually planning to work out the details at this point.

benjub:
@digama0 Woud this adaptation of you proof work for the comment:
```
The proof stems from the observation that by the axiom of choice, every set is equinumerous to a
cardinal (its cardinality), and the class of ordinals is totally ordered by inclusion ( ~ordtri2or2 ),
therefore so is its subclass of cardinals. To turn that observation into a proof, we have to show that
 for cardinals, the two order relations from the previous sentence, ` ~<_ ` and ` C_` , agree, that is,
that ` A ~<_ B ` implies ` A C_ B ` for cardinals (the converse being obvious).  This fact is more
elementary than the Schroeder--Bernstein theorem.  Here is a proof sketch: Let ` A ~<_ B ` be cardinals
 and suppose ` B C_ A ` (by ordinal trichotomy  ~ ordtri2or2 , if not, then one has `A C_ B` and there
is nothing to prove). Then, ` A ` is equinumerous to a subset ` C C_ B ` (the image of an injection
witnessing ` A ~<_ B ` ).  By the axiom of choice again, ` C ` can be well-ordered so is order-isomorphic
to an ordinal ` D ` .  Since ` D ` is an ordinal of cardinality ` A `, one has ` A C_ D ` .  Since ` D `
is an ordinal which is order-isomorphic to a subset of the ordinal ` B ` , one has ` D C_ B `.  So by
transitivity of inclusion, ` A C_ B ` .
```
?

digama0:
> By the axiom of choice again, ` C ` can be well-ordered so is order-isomorphic to an ordinal ` D ` .

You don't want to invoke the axiom of choice here, because what we actually want is the canonical well ordering on C given by epsilon (it is a collection of ordinals so it is well ordered already). That way, when we apply the ordinal collapse theorem (https://us.metamath.org/mpeuni/ordtype.html) to get `D` , we can prove by transfinite induction that each element of `C` is greater or equal to its image in `D` and hence `D C_ B`.

benjub:
> > By the axiom of choice again, `C` can be well-ordered so is order-isomorphic to an ordinal `D` .
> 
> You don't want to invoke the axiom of choice here, because what we actually want is the canonical well ordering on C given by epsilon (it is a collection of ordinals so it is well ordered already). That way, when we apply the ordinal collapse theorem (https://us.metamath.org/mpeuni/ordtype.html) to get `D` , we can prove by transfinite induction that each element of `C` is greater or equal to its image in `D` and hence `D C_ B`.

Of course ! I'm writing that "there exists" an order-isomorphism and the following sentence I'm assuming it is the canonical one ! Thanks for catching that. The remaining of the proof is then clear.

Since https://us.metamath.org/mpeuni/ordtype.html is pretty general and stated for all set-like good orders, I was hoping that the special "archetypal" case we need, the membership relation, could be proved earlier and more elementarily, but a quick look at the proof seems to indicate that there would be no simplification since this would be basically the same proof, unfortunately. Correct ? Nevertheless, since that proof is mainly based on the same transfinite induction you mentioned above, there seems to be nothing inherently Schroeder--Bernstein-like in it... Does this count as "handwaving that there is no circular argument" ?

**edit:** if I interpret correctly the ouput
```
!SHOW TRACE_BACK ordtype / TO sbth
The proof of statement "ordtype" traces back to "sbth" via:
  (None)
```
this confirms that ~ordtype does not depend on ~sbth. (And ~ordtri2or2 is proved before ~sbth so does not depend on it either.)

jkingdon:
I'm probably going to close this pull request. Sounds like I should let those who are more actively working on this open a pull request with what, if anything, should be merged.

benjub:
> I'm probably going to close this pull request. Sounds like I should let those who are more actively working on this open a pull request with what, if anything, should be merged.

I think you should keep it open. It's just a matter of knowing which few lines we add to the comment before merging, and it looks like we are converging to a suitable candidate.

jkingdon:
> > I'm probably going to close this pull request. Sounds like I should let those who are more actively working on this open a pull request with what, if anything, should be merged.
> 
> I think you should keep it open. It's just a matter of knowing which few lines we add to the comment before merging, and it looks like we are converging to a suitable candidate.

Sorry, I'm not really following most of this which is why I'm bowing out.

Feel free to open a new one, though.

PR #3353, opened 2023-08-06
Intuitionize tX theorems in the Continuous function-builders section

jkingdon:
These intuitionized very easily.  Many required no changes at all, and the others were easy (for example, the proof had been using a theorem which is a biconditional in set.mm, but only needing the direction which is also present in iset.mm as an implication).


wlammen (approved):
I checked that all the theorems are exactly the same than in set.mm.  The automatic proof checking ensures there is no violation of rules.  Whether this is a sensible movement of theorems to iset.mm is beyond my scope.  I trust you here

jkingdon:
> I checked that all the theorems are exactly the same than in set.mm. The automatic proof checking ensures there is no violation of rules. Whether this is a sensible movement of theorems to iset.mm is beyond my scope. I trust you here

"the same theorems as in set.mm and passes the checks" is pretty much sufficient for a set.mm->iset.mm move to be sensible. The slightly trickier cases come in when it is necessary to change the theorems but that wasn't true in this case.

Issue #98, opened 2017-01-07
Change mpt2 to mpo


digama0:
Does anyone object to changing the label fragment for two-argument mapping from `mpt2` to `mpo`? The former often conflicts with numbered theorems about single argument mapping. It has also been mentioned as a proposal at the top of set.mm for many years, where it is also accompanied by a proposal to change `mpt` to `mpf`. How do people feel about these changes?

Current status (August 2023, by @jkingdon ): `mpt2` to `mpo` rename is underway. Remaining tasks:

* [x] finish renaming theorems in set.mm
* [x] finish renaming theorems in iset.mm (there's at least one which isn't also in set.mm)
* [x] update comment for section "Special maps-to operations" above opeliunxp2f - both set.mm and iset.mm
* [ ] update proposed renames entry in changes-set.txt
* [ ] update references in missing theorems list in mmil.html
* [ ] update reference in mmnatded.raw.html
* [ ] update references in mmnotes.txt

david-a-wheeler:
No objection, makes sense to me.  However, please include in the df-... comment a note explaining what the abbreviation stands for (in general).  We have several cases where the abbreviation is a mystery for native English speakers, and I expect it's worse for non-native speakers.

nmegill:
On 1/7/17 3:20 PM, Mario Carneiro wrote:
> Does anyone object to changing the label fragment for two-argument
> mapping from |mpt2| to |mpo|? The former often conflicts with numbered
> theorems about single argument mapping. It has also been mentioned as a
> proposal at the top of set.mm for many years, where it is also
> accompanied by a proposal to change |mpt| to |mpf|. How do people feel
> about these changes?

I don't have any objection, but I was the one who proposed it.:)  Of 
course we would document all renames in the change list at the top of 
set.mm, as usual.

Norm

jkingdon:
I'm all for changing `mpt2` to `mpo` and would be in favor of doing this change in iset.mm as well (keeping track of `mpt` versus `mpt2` has been mildly confusing when intuitionizing the `mpt2` theorems).

nmegill:
The current proposals at the top of set.mm are:

proposed  *mpt*     *mpf* or *mptf* (maps-to for functions)
proposed  *mpt2*    *mpo* or *mpto* (maps-to for operations

The 4-letter prefix alternatives are ok but may be slightly longer than necessary.  mpto* conflicts with modus ponendo tollens, mpto1 and mpto2, but those can be renamed.

[...]

PR #3354, opened 2023-08-06
Rename df-mpt2 to df-mpo

jkingdon:
This is in both set.mm and iset.mm and is accompanied by a plan to rename all the other theorems which contain this abbreviation.

For the background discussion see #98 but my shorthand summary of why to change it somehow is that "2" does not convey anything meaningful, but "o" stands for "operation" and thus makes sense here.

I'm not sure there is a particularly strong consensus in favor of any particular replacement name, but here is my thinking:

1. **four character versus three character abbreviations** Although we do have some abbreviations which are four characters long (or even longer), generally speaking it seems to me we have been leaning towards two and three character abbreviations. For example in #2625 `djun` was proposed as the abbreviation for disjoint union but we chose `dju` instead. Because these abbreviations (especially one like `mpt2` or whatever replaces it) get combined into longer names a lot (examples include `elmpt2cl2` and `mptmpt2opabovd`), my take on it is that a longer abbreviation both makes for a long name, but more importantly makes it even harder than it already is to figure out how a name is built from abbreviations. There are a few words explicitly about longer versus shorter abbreviations at https://github.com/metamath/set.mm/issues/98#issuecomment-459412686 .
2. **whether to keep mpt or rename that too** I guess for me there's nothing especially wrong with `mpt` and the problem with `mpt2` is (a) the `2` is not mnemonic, (b) the `2` can be confused with `2` being used other ways, (c) having `mpt` as a prefix of `mpt2` causes a cognitive dissonance in realizing which is meant. A secondary consideration is the amount of work required to rename something as widely used as this, but if we prefer `mpf` we can put it in as a proposed rename even if we might not get to it as soon as renaming `mpt2`.
3. **mpt as prefix of mpt2** Perhaps this warrants its own item, as there have been some comments that having one abbreviation be a prefix of the other is harmless or even desirable. While it is true that the two are related (in a way we can formalize in theorems like https://us.metamath.org/mpeuni/mpt2mpt.html ), they are two different syntaxes and my experience with the `mpt2` theorems (especially at first, maybe even now) is that I don't know whether to parse it as two abbreviations - `mpt` and `2` - or as one. Perhaps the best example of the difficulty is the `cnmpt*` theorems (although I will confess I'm a little bit worried that if I look too long at how those ones are named, and why, it will be like staring at the sun and I'll end up going blind). This same issue may apply if the choice is `mpt`/`mpto`. It is less of an issue for `mptf`/`mpto` and even less for `mpf`/`mpo` or `mpt`/`mpo`.
4. **Modus ponendo tollens** Can't speak for everyone, but at least for me this has never caused confusion. Our abbreviations are sometimes ambiguous so I'd ask whether it  has been a problem in practice and in this case I'd say no.

Oh, and as for _how_ to rename it, I'm sort of assuming that our best bet is a (relatively) small pull request which expresses the intention and begins the renames, followed by a series of pull requests which rename several theorems at a time. The goal would be to get it over with fairly quickly but trying to do everything in one giant pull request I think would probably be difficult in a variety of ways.

I would propose to keep this pull request open for discussion for at least a few days. Given that we haven't really had any recent discussion of this topic we need to dust off the various considerations, let people consider it again, etc. But hopefully we can reach some level of (enough) consensus, since I don't think I've seen anyone say that the best outcome is to keep `mpt2`.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -20,13 +20,13 @@ proposed  sylbid    biimtrd     etc.
 proposed  sylbird   biimtrrd    etc.
 proposed  syl5*     *trid       (syl5bi -> biimtrid; syl5eqel -> eqeltrid;etc.)
 proposed  syl6*     *trdi
-proposed  *mpt*     *mpf* or *mptf* (maps-to for functions)
-proposed  *mpt2*    *mpo* or *mpto* (maps-to for operations)
+proposed  *mpt2*    *mpo*       (renaming process underway)
 (Please send any comments on these proposals to the mailing list or
 make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 6-Aug-23 df-mpt2   df-mpo
 31-Jul-23 sbcgfi    [same]      moved from GM's mathbox to main set.mm
 31-Jul-23 csbgfi    [same]      moved from GM's mathbox to main set.mm
 31-Jul-23 sbceqi    [same]      moved from GM's mathbox to main set.mm


digama0:
I'm :+1: on the rename, I agree with your points overall.

avekens:
It seems that @benjub and I were favoring `mpto` in #98. Meanwhile, however, I do not have this strong opinion anymore, so `mto` would also be fine.

For the following renamings, please consider Norm's advice in 2017 (change list is in a separate file now):

> Of course we would document all renames in the change list at the top of set.mm, as usual.

jkingdon:
> For the following renamings, please consider Norm's advice in 2017 . . . we would document all renames in the change list

Yes, that is the plan.  This pull request adds `6-Aug-23 df-mpt2 df-mpo` to `changes-set.txt` and the intention is to add a similar line for each theorem as it is renamed in future pull requests.

david-a-wheeler:
+1 on the rename. I agree that "2" doesn't convey any meaning.

Thanks for posting the heads-up to the mailing list, as that will help people who may need to make some tweaks to "proofs in progress".

benjub:
no objection

jkingdon:
Thanks for all the input, everyone.

I have rebased against the latest develop and plan to merge once the tests pass.

PR #3189, opened 2023-05-18
Reduced axiom usage

GinoGiotto:
This is the PR that I anticipated in https://github.com/metamath/set.mm/issues/3183 + some extra:

* Edited proof of `nfccdeq` -> drops `ax-11`

* Edited proof of `dfss2f` -> drops `ax-13` and automatically drops `ax-13` for these 264 theorems too: https://github.com/metamath/set.mm/commit/70740cdfa56528de67ae2cc7bedeb7c7500ae1b5

* Edited proof of `iunxsngf` -> drops `ax-13`

* Edited proof of `issmo` -> drops `ax-13`

Also I show that `eqsb3v` is identical to `eqsb3lem`, so one of them can be removed.

Issue #3183, opened 2023-05-17
Remove depedencies from ax-13


GinoGiotto:
With the `/MAY_GROW` option is possible to use the minimizer as a sort of axiom usage minimizer. For most axioms this tool provides little success, but I found a significant difference for `ax-13`. With a scan of 5000 theorems in main I found 267 that can be proven without it, also without adding any disjoint variable conditions. The only tradeoff is that some proof are a bit longer, but it's usually a few bytes.

The question is: Are you interested in those proofs?

jkingdon:
I think "We prefer proofs that depend on fewer and/or weaker axioms, even if the proofs are longer." from https://us.metamath.org/mpeuni/conventions.html is pretty clear on this point. I suppose if it is more than a few bytes there might sometimes be a reason for an ALT proof to show both ways but we mostly haven't done that (for example https://us.metamath.org/mpeuni/sbth.html mentions the Axiom of Infinity or the Axiom of Choice as not being used but does not show alternative proofs which use one or both of those axioms).

david-a-wheeler:
Short answer: Yes.

Longer answer: As @jkingdon noted, if the new proofs are significantly longer, we should keep the older proofs as ALT forms. But only if they're a lot longer.

I'm surprised no one has mentioned this before. We in general try to reduce axiom usage. Good idea!

GinoGiotto:
I looked into this more clearly and I think I understand what happened:

* The minimizer edited around 450 proofs out of the 5000 scanned.
* Among these 450 proofs there are 3 that accidentally dropped `ax-13`.
* It happens that one of those 3, which is called `dfss2f`, is used by many following theorems, so all other 264 ones dropped `ax-13` without having their proof edited!

In the end it's much simpler than I thought, it's sufficient to make a PR for 3 theorems only to automatically reduce axiom usage of the others. I guess that's the reason why nobody mentioned it: diminishing axiom usage with this method looks like a rare occurrence, I was just lucky enough to spot a very "influential" theorem. Anyway it was an interesting test to make.

david-a-wheeler:
One of the reasons we prefer the "main" proof to have fewer axioms is to cause later proofs to possibly use fewer axioms. I'm surprised by the *extent* in this case - but also delighted :-).

benjub:
[...]

PR #3355, opened 2023-08-06
Reduced axiom usage

GinoGiotto:
These few proof edits reduce a variety of axiom dependencies, specifically:

* Dropped `ax-13` for 548 theorems.
* Dropped `ax-10` for 12 theorems.
* Dropped `ax-12` for 8 theorems.
* Dropped `ax-7` for 6 theorems.
* Dropped `ax-6` for 6 theorems.
* Dropped `ax-11` for 2 theorems.

The details can be consulted here: https://github.com/metamath/set.mm/commit/37e43a46c76438ae17f687fa854e38d7d5467d2e  (you can verify it with `metamath-knife <database-name> --axiom-use <output-file>`).

The technique I used in this PR is somewhat more convoluted than the one presented at https://github.com/metamath/set.mm/issues/3183, but it seems more fruitful. With a selective scan of only 500 theorems I got over twice the amount of dependency reductions compared to my previous most rewarding PR https://github.com/metamath/set.mm/pull/3189 (which was the result of a scan of 5000 theorems). However, going forward, there might be some diminishing returns, so I'm not entirely sure about the consistency of this new approach.

One hard task was to mimic what I would get if an `/IF_DROP` option would be present https://github.com/metamath/metamath-exe/issues/155. This automation is useful for me because it spares the human time I would spend to filter out unwanted changes. Currently my hacky solution is quite slow, so if you find this result meaningful, then you might be interested in that proposal (I didn't see much discussion there, so maybe this PR could draw some interest towards it).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -8116,6 +8116,30 @@
 "imbi13" is used by "trsbc".
 "imbi13" is used by "trsbcVD".
 "impexpd" is used by "impexpdcom".
+"impsingle" is used by "impsingle-step15".
+"impsingle" is used by "impsingle-step18".
+"impsingle" is used by "impsingle-step20".
+"impsingle" is used by "impsingle-step22".
+"impsingle" is used by "impsingle-step4".
+"impsingle" is used by "impsingle-step8".
+"impsingle-step15" is used by "impsingle-step18".
+"impsingle-step15" is used by "impsingle-step21".
+"impsingle-step15" is used by "impsingle-step25".
+"impsingle-step18" is used by "impsingle-step19".
+"impsingle-step19" is used by "impsingle-step20".
+"impsingle-step20" is used by "impsingle-step21".
+"impsingle-step20" is used by "impsingle-step25".
+"impsingle-step21" is used by "impsingle-imim1".
+"impsingle-step22" is used by "impsingle-peirce".
+"impsingle-step22" is used by "impsingle-step25".
+"impsingle-step25" is used by "impsingle-imim1".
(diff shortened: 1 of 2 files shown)


set.mm
```diff
@@ -21576,7 +21576,8 @@ proposition with a distinct variable (closed form of ~ nfsb4 ).
     $( Equivalence for substitution.  (Contributed by NM, 2-Jun-1993.)  (Proof
        shortened by Wolf Lammen, 23-Sep-2018.) $)
     sb6a $p |- ( [ y / x ] ph <-> A. x ( x = y -> [ x / y ] ph ) ) $=
-      ( wsb weq wi wal sbco sb6 bitr3i ) ABCDACBDZBCDBCEKFBGABCHKBCIJ $.
+      ( wsb weq wi wal sbcom3vv sb6 sbid sbbii 3bitr3ri ) ACBDZBCDACCDZBCDBCEMF
+      BGABCDACBCHMBCINABCACJKL $.
```
benjub, line 21580: It may be more versatile to prove a "sbcov" from "sbcom3vv" in the same way that sbco is prove from sbcom3.

Then, sb6a should be moved to before ax-13.

GinoGiotto: I committed the suggestion, is the new placement good?

david-a-wheeler:
@GinoGiotto - any approach would have diminishing returns as you keep going, so don't worry about that. I'm impressed with these results!

wlammen (approved):

benjub (approved):

Issue #3344, opened 2023-08-03
Add metamath to lets-prove-leftpad


jkingdon:
At https://github.com/hwayne/lets-prove-leftpad there is a collection of proofs for different proof systems which includes:

1. A definition of a `leftpad` function
2. Proof of three theorems about that function

I didn't take a super close look at how to do this in metamath but it seems like words over a set (df-word) would have most of what we need.

I guess the leftpad definition could be a hypothesis ("local definition") if we don't want to make it global.

tirix:
Yes, words over a set is what we should use.

It seems that repository has a directory for each system, so maybe we could write a mini database importing `set.mm`, defining the `leftpad` function, and then proving the 3 theorems, which we could put there.

jkingdon:
> It seems that repository has a directory for each system, so maybe we could write a mini database importing `set.mm`, defining the `leftpad` function, and then proving the 3 theorems, which we could put there.

I was thinking of adding the 3 theorems to set.mm and copy-paste them to the lets-prove-leftpad repo. But maybe a .mm file which imports set.mm is better.

tirix:
My work-in-progress is here:
[https://github.com/tirix/lets-prove-leftpad/tree/metamath/metamath](https://github.com/tirix/lets-prove-leftpad/tree/metamath/metamath)

It seems that the last pull request has been on hold since January 2023, I'm unsure if any new one will get merged.

jkingdon:
> My work-in-progress is here: https://github.com/tirix/lets-prove-leftpad/tree/metamath/metamath
> 
> It seems that the last pull request has been on hold since January 2023, I'm unsure if any new one will get merged.

I suppose the next step (especially since you have done so much of the work already) is to get it in pull request form even if we aren't really sure what would happen after that? The explanation of the proofs could perhaps be a little less detailed if that made it any easier (although I don't have a strong opinion on that).

There were several pull requests merged in 2022 for whatever that is worth.

tirix:
I've added more to the readme file:
[https://github.com/tirix/lets-prove-leftpad/tree/metamath/metamath](https://github.com/tirix/lets-prove-leftpad/tree/metamath/metamath)

> explanation of the proofs could perhaps be a little less detailed

I've pushed the detailed explanations to the very end of the file, and kept only the general principles at the top.

> the next step is to get it in pull request form

Yes. I'd love to get a review of that submission by @david-a-wheeler too!

jkingdon:
[...]

PR #3357, opened 2023-08-08
Leftpad

tirix:
As brought to attention by @jkingdon in #3344, it would be nice to add metamath to [this list of formalizations of leftpad](https://github.com/tirix/lets-prove-leftpad).

This PR implements `leftpad` as specified and proves simple theorems about it, including:
- `lpadmax` : the length of the output is `max(n, len(str))` -- expressed using an `if`
- `lpadleft` : the prefix of the output is padding characters and nothing but padding characters (I'm not using the `prefix function` thoguh)
- `lpadright` : the suffix of the output is the original string.

These changes are not meant to be merged, but instead to be added to the `lets-prove-leftpad` repository a separated MM file, importing `set.mm`. Together with that file, we would add a README file with:
- a general introduction of Metamath and set.mm
- a description of the chosen definition and how it "works"
- explanations about each theorem and its proof.

Ideally, we would also link to the theorem's web pages where people can read the detailed steps.


tirix:
Here are explanations, to be put together with the raw file:

**Definition of `leftpad`**
Character strings in `set.mm` are formalized as functions from integer intervals starting at zero to the character set.
In order to formalize the repetition of `n` time a character `C`, a constant function from the interval `0 ... (n - 1)` is used.

Here the number of repetitions should be the target length `L`, minus the length of the original word `W`:
```
( L - ( # ` W ) )
```

A left-close, right-open integer interval is expressed in `set.mm` using the `..^` operator, therefore this interval is written as:
```
( 0 ..^ ( L - ( # ` W ) ) )
```

A function is formalized as a set of pairs "<argument, value>".  When the function is constant, this is in fact the cross-product of the domain and a singleton containing the unique value:
```
( ( 0 ..^ ( L - ( # ` W ) ) ) X. { C } )
```

Finally, the repetition of the constant is concatenated with the original word ` W `. The symbol for the concatenation operation is `++`:
```
( ( ( 0 ..^ ( L - ( # ` W ) ) ) X. { C } ) ++ W )
```

The `leftpad` function takes 3 arguments: a character, a length, and a word. There is no formal definition in `set.mm` for functions of 3 arguments, but we can use currying and functions of 2 arguments.
A first function takes a length `L` and builds the left-padded string. The domain of that function is `NN0`, the set of non-negative integers. We're using lowercase letters here for free variables. Metamath limits itself to ASCII characters, so `∈` is written as `e.`.  This first function is written:
```
( l e. NN0 |-> ( ( ( 0 ..^ ( l - ( # ` W ) ) ) X. { C } ) ++ W ) )
```

The complete function takes the two other arguments, the character `C` and the length `L`, and returns the first function. In order to keep the definition generic, we don't constrain `C` and `L` other than requiring that they are sets, i.e. elements of the universal class `_V`. This is the complete expression you'll find in the definition of `leftpad`:
```
( c e. _V , w e. _V |-> ( l e. NN0 |-> ( ( ( 0 ..^ ( l - ( # ` w ) ) ) X. { c } ) ++ w ) ) )
```

**Theorem `lpadval`**
In the theorems to be proved, we're not interested in the function itself, but rather its value. Our first theorem therefore simply extracts the value of the `leftpad` function for given arguments, by applying two evaluations, and the required explicit substitutions.

In order to be able to evaluate the function, the arguments have to be sets. We further restrict them to the domains we will actually use. The argument `W` shall be a word over a given set `S`. This is expressed as ` W e. Word S`. The repeated character should itself be an element of `S`: `C e. S`.

It is convenient in many cases to prove theorems in the deductive form, i.e. when both its statement and its hypotheses are derived from a common condition `ph`, so they are written:
```
h1::lpadval.1   |- ( ph -> L e. NN0 )
h2::lpadval.2   |- ( ph -> W e. Word S )
h3::lpadval.3   |- ( ph -> C e. S )

lpadval |- ( ph -> ( ( C leftpad W ) ` L ) = ( ( ( 0 ..^ ( L - ( # ` W ) ) ) X. { C } ) ++ W ) )
```

TO BE CONTINUED...

jkingdon:
Looks good to me.

If we want to have these proofs up on the web, were you thinking of merging it into set.mm or did you have some other plan? Seems to me merging it into set.mm (in addition to the lets-prove-leftpad repository) seems fine. (This is for the "A link to an online demo of your leftpad" part of the [lets-prove-leftpad contribution guidelines](https://github.com/hwayne/lets-prove-leftpad/blob/master/CONTRIBUTING.md)).

As for the explanations, the text above sounds like a good start for the "A description of your proof. " section. We could perhaps elaborate a bit on "What interesting language properties or verification techniques does it showcase?" which I suppose would be currying or words or deduction form or if-expressions although I don't know whether we are better off trying to make them sound interesting or just say what they are.

"The name of your tool, and a link to learn more about it." - hopefully self explanatory

"A description of the language. What does it look like? How does it work? What makes it different or special?" - we can presumably pull a paragraph or two off the metamath website for this.

"Anything you want to plug, like your website or twitter or CV. You proved Leftpad! Brag a bit!" Feel free to include this for yourself, but if you want it for the project, I don't know, maybe something about how we encourage new contributors or something about the Metamath 100 or something.

jkingdon (approved):

tirix:
I agree with all that.

And indeed in hindsight, having this chapter in `set.mm` will also help generate the pages automatically and allow to link them, so this seems like the best solution.

jkingdon (approved):

PR #3359, opened 2023-08-10
Reduced ax-13 usage

GinoGiotto:
Removed `ax-13` from the dependencies of `reusv1` and `reusv2lem1`


wlammen (approved):

benjub:
Good catch ! ~nfmodv was added for that kind of things; have you had a look at other theorems using nfmo(d(2)) ?

GinoGiotto:
> Good catch ! ~nfmodv was added for that kind of things; have you had a look at other theorems using nfmo(d(2)) ?

@benjub I didn't look at those specifically, the automated approach I'm using has a broad spectrum (I simply listed all theorems that do depend on ax-13 and scanned the first 500), so they might have fallen in some of my findings, but I have to take a look. If you think they are promising then I might try to make a focused scan for them.

In general anything that might optimize the process is welcome, I still struggle to get consistent results, so that I can make it sistematic. Are theorems avoiding ax-13 structured in some way? For example all theorems that do depend on ax-13 could be easily classified by how many dependency "layers" they need to reach ax-13, but I'm not sure if it's possible to do this kind of reasoning for the ones avoiding it. I only noticed some naming patterns like the *v at the end of the label, but that's basically it.

benjub:
> @benjub I didn't look at those specifically, the automated approach I'm using has a broad spectrum (I simply listed all theorems that do depend on ax-13 and scanned the first 500), so they might have fallen in some of my findings, but I have to take a look. If you think they are promising then I might try to make a focused scan for them.

No need for extra work on these, I think.

> In general anything that might optimize the process is welcome, I still struggle to get consistent results, so that I can make it sistematic. Are theorems avoiding ax-13 structured in some way? For example all theorems that do depend on ax-13 could be easily classified by how many dependency "layers" they need to reach ax-13, but I'm not sure if it's possible to do this kind of reasoning for the ones avoiding it. I only noticed some naming patterns like the *v at the end of the label, but that's basically it.

There is nothing systematic, but some ideas in the google group: https://groups.google.com/g/metamath/c/1wi1s6qBYqY/m/FPkPsd5oAwAJ and in the section of my mathbox titled "20.14.4.11  Removing dependencies on ax-13 (and ax-11)". I should review that section, since probably some theorems are not any longer improvements on their Main section counterparts, given the recent changes.

PR #3361, opened 2023-08-10
Rename cmpt2 to cmpo

jkingdon:
This is a syntax definition and as such is invisible in many contexts. However, it still seems desirable to have it use the new naming convention.

This was done by:
* rename the definition
* search and replace on the usages
* run scripts/rewrap (many of the lines changes are as a result of this step)

This is part of an ongoing rename, see #3354 for more details.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
I agree! Consistency is a good thing. Thanks for fixing this.

icecream17 (approved):

PR #3363, opened 2023-08-11
A set is countably infinite iff it has decidable equality, is countable, and is infinite

jkingdon:
At the time I added https://us.metamath.org/ileuni/ennnfone.html I wasn't sure whether it could be stated more nicely or whether there was something more subtle about how it was arranged in the textbook I got it from.

Turns out it can be stated nicely, in a direct translation to symbols of "A set is countably infinite iff it has decidable equality, is countable, and is infinite" or

```
    ctinf $p |- ( A ~~ NN <-> ( A. x e. A A. y e. A DECID x = y
        /\ E. f f : _om -onto-> A
        /\ _om ~<_ A ) ) $=
```

The key insight of the proof is that any function from the natural numbers onto `A` will satisfy the conditions of https://us.metamath.org/ileuni/ennnfone.html (under the condition that `_om ~<_ A`) - there is no need to construct a different function.

The proof is not especially difficult but has various fiddly details. Some of these fill in gaps in iset.mm, such as an intuitionized version of https://us.metamath.org/mpeuni/difinf.html (the need for decidable equality here surprised me, but I'm not sure it can be avoided). Some of them are just conversions between `NN0` and `_om` and the like.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4034,6 +4034,11 @@ this implies excluded middle</TD>
   <TD>Implies excluded middle as shown at ~ unfiexmid .</TD>
 </TR>
 
+<tr>
+  <td>difinf</td>
+  <td>~ difinfinf</td>
+</tr>
+
 <TR>
   <TD ROWSPAN="2">prfi</TD>
   <TD>~ prfidisj</TD>


wlammen:
I have only limited knowledge about this topic, in particular under the restrictions of iset.mm.  The results presented here are all automatically checked, so any objection from my side can only go into the direction of whether the theorems and lemmas could be proven with fewer or less strict restrictions (like dv conditions, hypotheses).  I don't know where the journey will go from here, but I feel one should either present reasons against this material, or allow it, tentatively at least.

wlammen (approved):

jkingdon:
> I have only limited knowledge about this topic

Thanks for taking a look.

Given that it has been open for four days and the other considerations you mention, I think we can take silence as consent. If people come up with issues later we can always revise it further.

As for where it is leading, it lets us solve a Metamath 100 problem as described at https://github.com/metamath/set.mm/issues/2595#issuecomment-1676202604 . I'll send up a follow up pull request for the remaining piece shortly.

jkingdon:
> I'll send up a follow up pull request for the remaining piece shortly.

Now at https://github.com/metamath/set.mm/pull/3386

PR #3364, opened 2023-08-11
Prove axpr without ax-ext

CatsAreFluffy:
I kept the old proof as axprALT, since the new proof is a lot weirder, and I moved it and zfpair to the previous section since they aren't required for the new ~ axpr.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13271,7 +13271,7 @@
 "zfac" is used by "axacndlem4".
 "zfinf" is used by "axinf2".
 "zfinf" is used by "axinfndlem1".
-"zfpair" is used by "axpr".
+"zfpair" is used by "axprALT".
 "zrdivrng" is used by "dvrunz".
 New usage of "00sr" is discouraged (4 uses).
 New usage of "0bdop" is discouraged (2 uses).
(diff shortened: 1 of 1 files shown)


jkingdon (approved):
Interesting. I've never learned a lot about what people mean by non-extensional foundations (more than just avoiding one axiom, I imagine) but that is what I think of when I see this.

benjub:
> Interesting. I've never learned a lot about what people mean by non-extensional foundations (more than just avoiding one axiom, I imagine) but that is what I think of when I see this.

I think that if you have a model of a given set theory without extension (e.g. ZF minus extension), then quotienting by the relation "having the same elements" generally gives a model of the extensional counterpart. One should of course check that the other axioms are still satisfied, i.e. that that quotienting preserves their truth...

I think there is a more interesting opposition extensional/intensional in type theory, of which I know little, but you may have seen that dichotomy when studying things in connection with your work on iset.mm ?

As for avoiding ax-ext in set.mm, there is a section in my mathbox (20.14.5.2  Classes without extensionality) for some explanations and results.

jkingdon:
> I think there is a more interesting opposition extensional/intensional in type theory, of which I know little, but you may have seen that dichotomy when studying things in connection with your work on iset.mm ?

I've seen references, for example references to "intensional type theory, rather than extensional type theory" in the HoTT book. But not enough to really know what it means. Perhaps not closely related to set.mm minus ax-ext, at least I get the sense that the word "extentionality" means a lot of different things in different contexts.

PR #3370, opened 2023-08-12
add aesb2

benjub:
@wlammen : here is the simpler version of wl-ich-lem1 I mentioned in https://github.com/metamath/set.mm/pull/3347#discussion_r1285196705. I think that result is interesting in itself so deserves to be in Main. As for contributions, I put the two of us on equal footing (the order is alphabetic), but if you prefer, I can put two tags: "Contr. by WL" and "Revised by BJ".

Maybe you also have a proposal for a better label?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17550,7 +17550,6 @@ New usage of "sbcoreleleq" is discouraged (2 uses).
 New usage of "sbcoreleleqVD" is discouraged (0 uses).
 New usage of "sbcrexgOLD" is discouraged (2 uses).
 New usage of "sbcssgVD" is discouraged (0 uses).
-New usage of "sbelxOLD" is discouraged (0 uses).
 New usage of "sbequ12ALT" is discouraged (2 uses).
 New usage of "sbequ1ALT" is discouraged (4 uses).
 New usage of "sbequ1OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -21197,6 +21197,17 @@ primitive connectives (no defined terms) on the right-hand side.
     dfsb1 ) BCEZBFZBDEZAGZUBAHZBIZHCDEZAGZUFAHZCIZHABDJACDJUAUCUGUEUIUAUBUFATUB
     UFKBBCDLMZNUDUHBCUAUBUFAUJOPQABDSACDSR $.
 
+  ${
+    $d v y $.
+    $( In the case of two successive substitutions for two always equal
+       variables, the second substitution has no effect.  (Contributed by BJ
+       and WL, 9-Aug-2023.) $)
+    aesb2 $p |-
```
wlammen, line 21205: The ae standing for "all equal" is often added as a suffix, as in cbvaev, hbae for example,  The rest is fine from my POV.

benjub: Relabeled to sb2ae.

benjub:
I also removed sbnf2OLD before the 1-year period since the proof is nearly identical to that of sbnf2 (just replace uses of sb8(e) with sb8(e)v and that's it):
https://us.metamath.org/mpeuni/sbnf2OLD.html
versus
https://us.metamath.org/mpeuni/sbnf2.html

benjub:
Regenerate `discouraged` and remove https://us.metamath.org/mpeuni/sbelxOLD.html for the same reason: nearly identical to sbelx.

wlammen (approved):

PR #3372, opened 2023-08-12
Move Subsection "Alternate definition of substitution"

benjub:
It looks like Subsection "Alternate definition of substitution" was created in the middle of an existing subsection, isolating the theorems coming after it. This MR moves it at the end of the subsection in the middle of which it was.

Two commits in order to ease side-by-side checking in case there is any doubt (and the message of the second commit should read "2/2" instead of "1/2"; this will be editable when squashing and merging).


icecream17 (approved):
I wonder if the section should be removed. dfsb7ALT was originally used to prove dfsb1, but now dfsb1 has been revised and the alternate theorems are unused

benjub:
Yes, I checked that removing the section as a bloc does not cause any proof breaking. Therefore, the section is well isolated from the rest.

I thought this was meant to be a temporary section anyway, that you created as a backlog of proofs to progressively adapt to the new definition ?

PR #3374, opened 2023-08-12
Rename mpt2eq123dva , mpt2eq123i and mpt2eq123dv

jkingdon:
Rename:

* mpt2eq123dva to mpoeq123dva
* mpt2eq123i to mpoeq123i
* mpt2eq123dv to mpoeq123dv

This is in both set.mm and iset.mm

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+12-Aug-23 mpt2eq123i mpoeq123i
+12-Aug-23 mpt2eq123dv mpoeq123dv
 12-Aug-23 mpt2eq123dva mpoeq123dva
 11-Aug-23 mpt2eq12  mpoeq12
 11-Aug-23 mpt2eq123 mpoeq123
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+12-Aug-23 mpt2eq123dva mpoeq123dva
 11-Aug-23 mpt2eq12  mpoeq12
 11-Aug-23 mpt2eq123 mpoeq123
 10-Aug-23 cmpt2     cmpo


wlammen (approved):

benjub:
@jkingdon : did you "squash and merge" ? or just merged ? Maybe it's me misusing git: I'm only using the basic commands like `git log -n`, but that gives an unnecessary long output.

jkingdon:
> @jkingdon : did you "squash and merge" ? or just merged ? Maybe it's me misusing git: I'm only using the basic commands like `git log -n`, but that gives an unnecessary long output.

As with many things in git, there are tradeoffs.

The advantage of fine grained commits is that each one can tell a small self contained story of what was changed. For example, if dozens of theorems were added but they are (basically) unrelated to each other, it might be easier to understand each addition if you don't also have to look at changes in the same commit which do not affect the theorem you are looking at. Or in the example of renames, having one commit for the rename and one for the rewrap may make it easier to see that it was a rename and not a bunch of other changes to each proof.

The advantage of coarse grained commits is that, as you suggest, it provides a grouping which may make it easier to see at a higher level how many broad categories of things have changed (that is, shorter git log output).

There are lots more thoughts in places like https://github.blog/2022-06-30-write-better-commits-build-better-projects/ (and a lot of other similar articles across the internet).

If the commits are not meaningful (that is, if they just reflect the steps taken to get something working, rather than a logical breakdown into different logical changes), general thinking is to squash them. If they are meaningful, I tend to prefer to keep them separate on the grounds that it may benefit future uses of git bisect or browsing the history.

I don't think we've had a huge amount of discussion on this in the context of set.mm except that @digama0 felt strongly that we should use either "Rebase and merge" or "Squash and merge" (rather than "Create a merge commit"), and someone (I think it was @wlammen ) said they found that the fine grained commits in my pull requests to be quite helpful during review (which, granted, is a somewhat different context than reading the changes after they have been merged, although people may wish to review commits after they are merged too).

PR #3375, opened 2023-08-12
Add note about df-clel , df-cleq and df-clab

jkingdon:
This is in the conventions where we describe reducing use of axioms.

There's a whole lot more background at https://github.com/metamath/set.mm/pull/3199#issuecomment-1676047526 and in some of the pages a click or two away from the text I'm adding, but the short summary is that whatever we want to do about these definitions/axioms in the future, we might as well describe what is going on now.


set.mm
```diff
@@ -422317,7 +422317,13 @@ or axiom of dependent choice ( ~ ax-dc ) can be used instead.
        and has no wff variables, can be proved *without* using
        ~ ax-10 through ~ ax-13 , by invoking ~ ax10w through ~ ax13w .
        We encourage proving theorems *without* ~ ax-10 through ~ ax-13
-       and moving them up to the ~ ax-4 through ~ ax-9 section.
+       and moving them up to the ~ ax-4 through ~ ax-9 section.</p>
+       <p>For these purposes, ~ df-clel , ~ df-cleq ,
+       and ~ df-clab should be considered axioms rather than definitions.
```
icecream17, line 422322: df-clab is actually conservative

proof: wcel is only used by wel at the time of df-clab. df-clab is not ambiguous with wel, so wcel can be considered to have a piecewise definition with wel and df-clab (at the point of df-clab)

benjub (commented):
This is indeed the kind of text we would have to write if we keep df-cleq and df-clel unchanged. Introducing particular cases adds complexity, so hopefully the changes mentioned in https://github.com/metamath/set.mm/pull/3199#issuecomment-1676095106 will make this not needed.

set.mm
jkingdon: OK, I wasn't sure whether to include `df-clab` just because the definition checker can't check it, or omit it because I had no knowledge of an interaction with axiom usage.

I've updated this pull request to omit it.

wlammen (commented):
I think a rewrap should be applied to improve the alignment of the text.

jkingdon:
> I think a rewrap should be applied to improve the alignment of the text.

Ah. I ran the rewrap script but forgot that it doesn't touch anything inside HTML tags.

I have done a manual rewrap.

wlammen:
Could you then push the result?

jkingdon:
> Could you then push the result?

I pushed up everything I did.

The text I added is pretty well wrapped, isn't it?

In general, I think we should mostly limit our concern for style and formatting to what we can check automatically.

wlammen (approved):

GinoGiotto:
Feel free to close https://github.com/metamath/set.mm/pull/3199 after this is merged.

tirix (approved):

jkingdon:
Whether this text ends up being here for a short time or for longer depends on whether we make other changes to df-clel and df-cleq, but that doesn't seem like a reason not to merge it.

PR #3378, opened 2023-08-13
Add shorter ax-groth equivalent

CatsAreFluffy:
My new ax-groth equivalent is 40 symbols shorter than ~ grothprim (or 106 if you expand the defined symbols in ~ grothprim, since my new equivalent uses only primitive symbols).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17498,7 +17498,6 @@ New usage of "rngosn6" is discouraged (1 uses).
 New usage of "rngoueqz" is discouraged (2 uses).
 New usage of "rpreOLD" is discouraged (0 uses).
 New usage of "rpssreOLD" is discouraged (0 uses).
-New usage of "rr-cp" is discouraged (0 uses).
 New usage of "rspsbc2" is discouraged (2 uses).
 New usage of "rspsbc2VD" is discouraged (0 uses).
 New usage of "ruALT" is discouraged (0 uses).
@@ -19604,7 +19603,6 @@ Proof modification of "rpnnen1lem5" is discouraged (1023 steps).
 Proof modification of "rpnnen1lem6" is discouraged (128 steps).
 Proof modification of "rpreOLD" is discouraged (21 steps).
 Proof modification of "rpssreOLD" is discouraged (7 steps).
-Proof modification of "rr-cp" is discouraged (97 steps).
 Proof modification of "rspsbc2" is discouraged (65 steps).
 Proof modification of "rspsbc2VD" is discouraged (90 steps).
 Proof modification of "ru" is discouraged (88 steps).
--- a/discouraged
+++ b/discouraged
@@ -17498,6 +17498,7 @@ New usage of "rngosn6" is discouraged (1 uses).
 New usage of "rngoueqz" is discouraged (2 uses).
 New usage of "rpreOLD" is discouraged (0 uses).
 New usage of "rpssreOLD" is discouraged (0 uses).
+New usage of "rr-cp" is discouraged (0 uses).
 New usage of "rspsbc2" is discouraged (2 uses).
 New usage of "rspsbc2VD" is discouraged (0 uses).
 New usage of "ruALT" is discouraged (0 uses).
@@ -19603,6 +19604,7 @@ Proof modification of "rpnnen1lem5" is discouraged (1023 steps).
 Proof modification of "rpnnen1lem6" is discouraged (128 steps).
 Proof modification of "rpreOLD" is discouraged (21 steps).
 Proof modification of "rpssreOLD" is discouraged (7 steps).
+Proof modification of "rr-cp" is discouraged (97 steps).
 Proof modification of "rspsbc2" is discouraged (65 steps).
 Proof modification of "rspsbc2VD" is discouraged (90 steps).
 Proof modification of "ru" is discouraged (88 steps).


set.mm
```diff
@@ -643498,13 +643507,43 @@ base set if and only if the neighborhoods (convergents) of every point
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 $)
 
+  ${
+    elnelneqd.1 $e |- ( ph -> C e. A ) $.
+    elnelneqd.2 $e |- ( ph -> -. C e. B ) $.
+    $( Two sets are not equal if an element of one is not an element of the
+       other.  (Contributed by Rohan Ridenour, 11-Aug-2023.) $)
```
benjub, line 643514:
```suggestion
    $( Two classes are not equal if there is an element of one which is not an element of the
       other.  (Contributed by Rohan Ridenour, 11-Aug-2023.) $)
```

set.mm
```diff
@@ -643498,13 +643507,43 @@ base set if and only if the neighborhoods (convergents) of every point
...
+    elnelneqd $p |- ( ph -> -. A = B ) $=
+      ( wceq wcel wa adantr simpr eleqtrd mtand ) ABCGZDCHFANIDBCADBHNEJANKLM
+      $.
+  $}
+
+  ${
+    elnelneq2d.1 $e |- ( ph -> A e. C ) $.
+    elnelneq2d.2 $e |- ( ph -> -. B e. C ) $.
+    $( Two sets are not equal if one is an element of a class but not the
+       other.  (Contributed by Rohan Ridenour, 12-Aug-2023.) $)
```
benjub, line 643524:
```suggestion
    $( Two classes are not equal if one but not the other is an element of a given
       class.  (Contributed by Rohan Ridenour, 12-Aug-2023.) $)
```

set.mm
```diff
@@ -643671,6 +643722,980 @@ base set if and only if the neighborhoods (convergents) of every point
...
+       be deduced, from which equivalence with Grothendieck universes
+       ( ~ grumnueq ) can be deduced.  (Contributed by Rohan Ridenour,
+       11-Aug-2023.) $)
+    df-mnu $a
+      |- MinUni = { y | A. z e. y A. f ( ~P z C_ y /\ E. w e. y
+        ( ~P z C_ w /\ A. i e. z
+          ( E. v e. y ( i e. v /\ v e. f )
+            -> E. u e. f ( i e. u /\ U. u C_ w ) )
+        )
+      ) } $.
```
benjub, line 644131:
```suggestion
            -> E. u e. f ( i e. u /\ U. u C_ w ) ) ) ) } $.
```
We usually use fewer lines. Same below. (most editors do bracket matching, so that shouldn't be a problem)

set.mm
```diff
@@ -643671,6 +643722,980 @@ base set if and only if the neighborhoods (convergents) of every point
...
+  CWIXDDWBWFWFWHXCWFVEIJWGVQVFVGVHVKVKVKTVLVGUIUTVMUTUIVL $.
+  $}
+
+  ${
+    $d u o $.  $d z v t $.  $d w o s $.  $d y z w v u f i $.
+    $( Express "every set is contained in a Grothendieck universe" using only
+       primitives.  The right side uses only 123 symbols (not counting the
+       outer universal quantifier, which would be unnecessary if it were an
+       axiom), which is significantly less than the previous record of 163
+       esablished by ~ grothprim (which uses some defined symbols, and requires
```
benjub, line 644679:
```suggestion
       established by ~ grothprim (which uses some defined symbols, and requires
```

set.mm
```diff
@@ -643671,6 +643722,980 @@ base set if and only if the neighborhoods (convergents) of every point
...
+       229 symbols if expanded to primitives).  (Contributed by Rohan Ridenour,
+       13-Aug-2023.) $)
+    rr-grothprim $p |- ( A. x E. y e. Univ x e. y <->
+      A. x -. A. y ( x e. y -> -. A. z ( z e. y
+        -> A. f -. A. w ( w e. y -> -. A. v -.
+          ( ( A. t ( t e. v -> t e. z ) -> -. ( v e. y -> -. v e. w ) )
+            -> -. A. i ( i e. z -> ( v e. y -> ( i e. v -> ( v e. f ->
+              -. A. u ( u e. f -> ( i e. u ->
+                -. A. o ( o e. u -> A. s ( s e. o -> s e. w ) ) ) ) ) ) ) ) ) )
+      ) ) ) $=
```
benjub, line 644689: Would the RHS be nicer if you allowed E., \/, /\ , C_ as well as restricted quantifiers ?

benjub:
Nice result !

As for MinUni, Coll, Scott, I do not know if it is worth introducing new definitions (maybe it is, I really haven't looked into it). Some contributors use "local definitions", even when they span quite a few results (I think @tirix and @avekens did that even for larger work, like Goldberg's conjecture).

set.mm
tirix: I guess that would look like this, simply applying `~ ismnu`:
```
A. x E. y ( x e. y /\ A. z e. y A. f ( ~P z C_ y /\ E. w e. y
        ( ~P z C_ w /\ A. i e. z
          ( E. v e. y ( i e. v /\ v e. f )
            -> E. u e. f ( i e. u /\ U. u C_ w ) ) ) ) ) ) )
```
(worth adding as a theorem, in my opinion)

tirix:
> As for MinUni, Coll, Scott, I do not know if it is worth introducing new definitions

I would in any case suggest to merge this PR as-is, since it is Mathbox only. For moving to main, here are my thoughts:

For `MiniUni`, since it is actually the same as `Univ`, maybe we could/should avoid the new definition?

For `Scott`, interestingly this is already in `set.mm`, see e.g. [scottex](https://us.metamath.org/mpeuni/scottex.html). So it seems Norm preferred not to introduce a new definition. Rohan is aware of these theorems and used them.

set.mm
benjub: `A. z e. y A. f ( ~P z C_ y /\ ...`
-->
`A. z e. y ( ~P z C_ y /\ A. f ...`
would probably be a bit simpler.

benjub:
> For `MiniUni`, since it is actually the same as `Univ`, maybe we could/should avoid the new definition?

In that case, there is really no need for another token.

set.mm
CatsAreFluffy: Added as ~ rr-groth .

CatsAreFluffy:
Okay, I removed `MinUni`.

tirix:
> Okay, I removed `MinUni`.

Oh, nice, a lot of work went into this! Thank you so much for taking these comments into account.

What about Scott's trick? I see it is also used in [`~karden`](https://us.metamath.org/mpeuni/karden.html), and [`~hta`](https://us.metamath.org/mpeuni/hta.html), but with a different relation. If we choose to define it, could we define Scott's trick in a more general way, like:
```
( Scott x ph ) = { x | A. y ( [. y / x ]. ph -> ( rank ` x ) C_ ( rank ` y ) ) ) }
```
This is also the expression used in [`~scottexs`](https://us.metamath.org/mpeuni/scottexs.html).
So we could have `( Scott x e. A )` or `( Scott x ~ A )`, or more generally `( Scott x ph )`.

Just to state it again: I think this contribution is fine as it is since it's a Mathbox, for the main part of set.mm we try to put more thought into it.

Those additional changes could also come in in a different PR.

CatsAreFluffy:
> If we choose to define it, could we define Scott's trick in a more general way, like:

This can be achieved with the current definition as `Scott { x | ph }`, which could be supported by a theorem like `( x = y -> ( ph <-> ps ) )`=>``Scott { x | ph } = { x | ( ph /\ A. y ( ps -> ( rank ` x ) C_ ( rank ` y ) ) ) }`` (from which the explicit substitution version can be easily derived). Given this, I think the current definition is better since it has fewer arguments and no binding. (Also, your proposed definition does not require `x` to satisfy `ph`.)

CatsAreFluffy:
Added as ~ scottab. I also added `( x = y -> ( ph <-> ps ) )`=>`( y e. Scott { x | ph } -> ps )` as ~ elscottab.

tirix:
> This can be achieved with the current definition as `Scott { x | ph }`

Oh indeed, this works well, and you're right it's better to avoid the binding.

CatsAreFluffy:
I realized that ~ scottab had too many DV constraints to instantiate it with explicit substitution, so I added ~ scottabf which would allow that and proved the explicit substitution version as ~ scottabes.

set.mm
```diff
@@ -644104,182 +644101,206 @@ base set if and only if the neighborhoods (convergents) of every point
...
   ${
-    $d y z f w v i u $.
-    $( Define the class of minimal universes.  These are the sets which satisfy
-       the predicate on ` y ` in ~ rr-grothprim , except for the ` x e. y `
-       clause.
+    $d z w v U f i k m n q p l $.  $d z w u U f i k m n r p l $.
+    ismnu.1 $e |- M = { k | A. l e. k A. m ( ~P l C_ k /\ E. n e. k
+      ( ~P l C_ n /\ A. p e. l ( E. q e. k ( p e. q /\ q e. m )
+        -> E. r e. m ( p e. r /\ U. r C_ n ) ) ) ) } $.
+    $( Properties of a minimal universe.  These are the sets which satisfy the
```
benjub, line 644109: If my understanding is correct, then I think it would be clearer to replace the beginning of the comment with:

The hypothesis of this theorem defines a class ` M ` of sets that we temporarily call "minimal universes", and which will turn out in ~ grumnueq to be exactly Grothendicek universes. Minimal universes are sets that satisfy...

set.mm
```diff
@@ -644661,29 +644740,97 @@ collection and union ( ~ mnuop3d ), from which closure under pairing
...
+      UNKCUOLMUQVCCURUPURFZUQRZVABUPNEZVBVFUTVGOZRZUNVGUTVIUNVGFZUQVEUQVHPVIUPV
+      DFZVJUQSVEVKUQVHVEUPUAFVKUPUBUPUCTUDUNUPAUEUFTUGVFVHUHUJVFUPVEUQUIUKULUMM
+      $.
+  $}
+
+  ${
+    $d x y $.  $d y z w v f i k m n q p l $.  $d y z w u f i k m n r p l $.
+    $( An equivalent of ~ ax-groth using only simple defined symbols.
+       (Contributed by Rohan Ridenour, 13-Aug-2023.) $)
+    rr-groth $p |- E. y ( x e. y /\ A. z e. y A. f ( ~P z C_ y /\ E. w e. y
```
benjub, line 644806: If you're not too tired with this MR, I would suggest
`A. f ( ~P z C_ y /\`
-->
`( ~P z C_ y /\ A. f`

benjub: (basically, choose the smaller scope for quantifiers; same below if applicable)

benjub:
I would call that operation simply the "Scott operation" (if that name is not already taken), and mention when important that this is the operation used in the "Scott trick".

CatsAreFluffy:
Okay, done.

benjub (approved):

PR #3381, opened 2023-08-13
Negative trichotomy property for real numbers

jkingdon:
It is well known that in iset.mm we cannot prove real number trichotomy, ` A < B \/ A = B \/ B < A ` .  Does that mean there is a pair of real numbers where none of those hold (that is, where we can refute each of those three relationships)?  Actually, no, as shown here.  This is another example of distinguishing between being unable to prove something, or being able to refute it.


wlammen (approved):

tirix (approved):
Nice!

jkingdon:
> Nice!

I got onto it because I was thinking of the same question for ordinals - which I submitted as https://mathoverflow.net/questions/452678/does-negative-trichotomy-hold-for-constructive-ordinals . But the real number case turned out to be easy.

PR #3383, opened 2023-08-14
sban, sbal, nfsbv

icecream17:
Note: Lots more commentary than necessary

sban: One direction of dfsb7 and therefore sbn is possible. While investigating this I found that sban was possible with fewer axioms too !
+ I say proof shortened for convenience, and also it's just 19.26 but with df-sb
+ df-sb has equivalents for ax-gen (sbt) ax-4 (sbi1) ax-5 (sbv) and ax-6 (use nsb/"sbn1") so theoretically all theorems before ax-7 (ax-10?) "correspond" to some theorem with df-sb
++ For example albiim would become `[yx](ph<>ps) <> ([yx](ph->ps) /\ [yx](ps->ph))`
+++ pretty interesting that sbbi also exists
+ I was not able to prove either direction of sbor, that was pretty unintuitive

sbal: fewer axioms, basically only mp-5 and 11
sbal1: obsolete
+ i did not investigate consequences

nfsbv: remove $d condition
+ this also would've worked under the previous definition so "revise df-sb" doesn't work
+ currently it doesn't seem very useful, at most it would eliminate a $d condition in sbco2v if there was something like sbiev2 (a la sbievw2)
++ it turns out the dv script only checks for variables that don't appear which makes such analysis harder

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -11872,6 +11872,7 @@
 "sb6ALT" is used by "sb5ALT2".
 "sb6fALT" is used by "sb5fALT".
 "sb7fALT" is used by "dfsb7ALT".
+"sbal1" is used by "sbalOLD".
 "sbanALT" is used by "sbbiALT".
 "sbanvOLD" is used by "sbbivOLD".
 "sbbiALT" is used by "sblbisALT".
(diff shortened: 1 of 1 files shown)


wlammen (approved):

benjub:
@icecream17 for the next MRs, can you try to be a bit more descriptive in the commit message and a bit more careful in the first comment of the MR ? (Several sentences above do not make sense.) Thanks.

(The results themselves are nice.)

PR #3389, opened 2023-08-15
Revise df-cleq and df-clel

benjub:
As discussed in #3199 (and several years ago in https://groups.google.com/g/metamath/c/xdHB0oW0aZ4/).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13673,6 +13673,7 @@ New usage of "ax6fromc10" is discouraged (1 uses).
 New usage of "ax6vsep" is discouraged (0 uses).
 New usage of "ax7v" is discouraged (2 uses).
 New usage of "ax8v" is discouraged (2 uses).
+New usage of "ax9ALT" is discouraged (0 uses).
 New usage of "ax9v" is discouraged (2 uses).
 New usage of "axac2" is discouraged (0 uses).
 New usage of "axacnd" is discouraged (2 uses).
@@ -18286,6 +18287,7 @@ Proof modification of "ax6e2ndeqALT" is discouraged (342 steps).
 Proof modification of "ax6e2ndeqVD" is discouraged (340 steps).
 Proof modification of "ax6fromc10" is discouraged (24 steps).
 Proof modification of "ax6vsep" is discouraged (73 steps).
+Proof modification of "ax9ALT" is discouraged (68 steps).
 Proof modification of "axac" is discouraged (55 steps).
 Proof modification of "axac2" is discouraged (108 steps).
 Proof modification of "axac3" is discouraged (74 steps).
@@ -18370,9 +18372,7 @@ Proof modification of "bj-ax12v3ALT" is discouraged (44 steps).
 Proof modification of "bj-ax6e" is discouraged (43 steps).
 Proof modification of "bj-ax6elem1" is discouraged (27 steps).
 Proof modification of "bj-ax6elem2" is discouraged (23 steps).
-Proof modification of "bj-ax8" is discouraged (55 steps).
-Proof modification of "bj-ax9" is discouraged (35 steps).
-Proof modification of "bj-ax9-2" is discouraged (74 steps).
+Proof modification of "bj-ax9" is discouraged (29 steps).
 Proof modification of "bj-axc10" is discouraged (30 steps).
 Proof modification of "bj-axc10v" is discouraged (37 steps).
 Proof modification of "bj-axc11nv" is discouraged (5 steps).
@@ -18437,7 +18437,6 @@ Proof modification of "bj-chvarvv" is discouraged (11 steps).
 Proof modification of "bj-clabel" is discouraged (43 steps).
 Proof modification of "bj-cleljustab" is discouraged (93 steps).
 Proof modification of "bj-cleljusti" is discouraged (20 steps).
-Proof modification of "bj-cleqhyp" is discouraged (21 steps).
 Proof modification of "bj-cmnssmnd" is discouraged (36 steps).
 Proof modification of "bj-cmnssmndel" is discouraged (5 steps).
 Proof modification of "bj-consensusALT" is discouraged (30 steps).
@@ -18445,12 +18444,8 @@ Proof modification of "bj-csbprc" is discouraged (47 steps).
 Proof modification of "bj-currypeirce" is discouraged (51 steps).
 Proof modification of "bj-denotes" is discouraged (76 steps).
 Proof modification of "bj-denotesv" is discouraged (15 steps).
-Proof modification of "bj-df-clel" is discouraged (4 steps).
-Proof modification of "bj-df-cleq" is discouraged (4 steps).
 Proof modification of "bj-df-nul" is discouraged (11 steps).
 Proof modification of "bj-df-v" is discouraged (29 steps).
-Proof modification of "bj-dfclel" is discouraged (27 steps).
-Proof modification of "bj-dfcleq" is discouraged (27 steps).
 Proof modification of "bj-dfnnf2" is discouraged (34 steps).
 Proof modification of "bj-disjsn01" is discouraged (18 steps).
 Proof modification of "bj-dral1v" is discouraged (36 steps).
@@ -18684,7 +18679,7 @@ Proof modification of "decmul1OLD" is discouraged (119 steps).
 Proof modification of "dedtOLD" is discouraged (19 steps).
 Proof modification of "demoivreALT" is discouraged (1087 steps).
 Proof modification of "dfbi1ALT" is discouraged (100 steps).
-Proof modification of "dfcleq" is discouraged (10 steps).
+Proof modification of "dfclel" is discouraged (17 steps).
 Proof modification of "dfeu" is discouraged (35 steps).
 Proof modification of "dfeuOLD" is discouraged (33 steps).
 Proof modification of "dfiun2gOLD" is discouraged (134 steps).
--- a/discouraged
+++ b/discouraged
@@ -18435,7 +18435,6 @@ Proof modification of "bj-ceqsalv" is discouraged (10 steps).
 Proof modification of "bj-chvarv" is discouraged (18 steps).
 Proof modification of "bj-chvarvv" is discouraged (11 steps).
 Proof modification of "bj-clabel" is discouraged (43 steps).
-Proof modification of "bj-cleljustab" is discouraged (93 steps).
 Proof modification of "bj-cleljusti" is discouraged (20 steps).
 Proof modification of "bj-cmnssmnd" is discouraged (36 steps).
 Proof modification of "bj-cmnssmndel" is discouraged (5 steps).
--- a/discouraged
+++ b/discouraged
@@ -13809,6 +13809,7 @@ New usage of "bj-0" is discouraged (1 uses).
 New usage of "bj-1" is discouraged (0 uses).
 New usage of "bj-ax12v3ALT" is discouraged (0 uses).
 New usage of "bj-ax6e" is discouraged (0 uses).
+New usage of "bj-ax9" is discouraged (0 uses).
 New usage of "bj-axc4" is discouraged (0 uses).
 New usage of "bj-axd2d" is discouraged (0 uses).
 New usage of "bj-axdd2" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub:
@digama0 and @tirix : can you see where the mmj2 error comes from ? It looks like it is about grammar conflicts. Did I introduce some with df-bj-mpt2 ?

On my local install (mmj2 v2.5.3 as of 23-Sep-2019.) there is no error, and I cannot install later versions (see https://github.com/digama0/mmj2/issues/68 and https://github.com/digama0/mmj2/issues/69).

tirix:
Yes, the problem is likely with your new syntax:
```
$( Syntax for maps-to notation for functions with two arguments. $)
  cmpt2-bj $a class ( <. x , y >. e. A |-> B ) $.
```
There is a `$j` statement saying `unambiguous 'klr 5';`. 
This means the grammar is unambiguous at depth 5. You probably made the grammar ambiguous, and maybe increasing that number would work - but I'm unsure which other consequences it might have.

A statement starting with `( <. x , y >. e. A ` might not distinguish between for example `( <. x , y >. e. A /\ x < B )` and the new `( <. x , y >. e. A |-> B )` except with a larger depth.

benjub:
> There is a `$j` statement saying `unambiguous 'klr 5';`. This means the grammar is unambiguous at depth 5. You probably made the grammar ambiguous, and maybe increasing that number would work - but I'm unsure which other consequences it might have.

Thanks for the explanation. Is this an LR(5) parser or an LALR(5) parser ?  In any case, I think that increasing the look-ahead does not degrade complexity by a lot ?  It has an effect only when looking ahead by 6 is actually needed, and during the initial phase before the parsing itself, IIRC.  I'm inclined to change it to `unambiguous 'klr 6';`.

benjub:
I removed that definition and will investigate in a future MR.  This is totally unrelated to the main object of this MR anyway.

tirix:
At least it was worth trying increasing that number.

For reference, here are a few links:
- [a post describing the mechanism](https://groups.google.com/g/metamath/c/yNdxi71vN68/m/hWCJXsztEgAJ),
- [the source code of the MMJ2 LR Parser](https://github.com/digama0/mmj2/blob/master/src/main/java/mmj/verify/LRParser.java)
- [documentation about grammar ambiguity](https://raw.githubusercontent.com/metamath/metamath-website-seed/main/downloads/grammar-ambiguity.txt)

Maybe this `$j unambiguous 'klr 5';`.  KLR(5) information is actually not used by the parser, and changing that value has no effect? I could not find the corresponding information in the MMJ2 source (neither a constant set to 5, nor reading the `klr 5` `$j` command)

avekens (approved):
For me everything looks fine, but I am not so deeply involved in the main topic of this PR, i.e., the change of the very basic definitions df-cleq and df-clel. As long as everything else still works, I trust in this changes. But maybe at least another one should approve these very basic changes.

icecream17 (approved):
I wonder if df-clel should have bj-cleljustab (shorter proof cleljust2) as a hypothesis too

benjub:
> I wonder if df-clel should have bj-cleljustab (shorter proof cleljust2) as a hypothesis too

Ideally, one should do that, but note that there is a formula variable in bj-cleljustab and we can only require a specific instance in an $e-hypothesis, so we can't really require the full scheme bj-cleljustab. (For instance, to prove that $e-hypothesis, you have the liberty of deciding to prove only the instance where `ph` is `T.`.)

Actually, the case of df-clel looks strictly harder than the case of df-cleq, and I do not see how to prove the same conservativity result as for df-clab and df-cleq... Maybe one has to think of a weaker conservativity statement ?

icecream17:
Consider an alternate axiom system where an instance of bj-cleljustab is provable without e.g. ax-12, but the scheme bj-cleljustab proves ax-12 (and the hypotheses are provable without ax-12)

Therefore this is non-conservative in an alternate axiom system.

(However, bj-cleljustab uses less axioms than the hypotheses in this axiom system, so it is conservative in that way. (Note that all finite expressions with classes expand to setvars and df-clab))

GinoGiotto:
Happy to see that this issue is going to be resolved, making axiom minimization clear and easier to do. Hopefully, this PR will settle the debate about these definitions.

jkingdon:
> As discussed in #3199 (and several years ago in https://groups.google.com/g/metamath/c/xdHB0oW0aZ4/).

Thanks for linking to that 2021 mailing list thread.

Although many topics are brought up there, I guess the crux of it is the following comment from Norm:

> Once we get to interesting results in set theory, all of the FOL= axioms will pretty much be used anyway.  Effort spent to eliminate the use of this or that FOL= axiom in the set theory development doesn't seem productive to me, and it may even be confusing to the reader to see longer proofs that may result just to avoid using a certain FOL= axiom in a set theory theorem.

My reaction to that is that if this is our position, we need to get a lot more specific in terms of how we talk about axiom usage. Somewhat in documentation, but in tools too.

I'm kind of inclined to think the current pull request is a step forward even if I'm not sure it solves everything. But I'm not sure I see any of this clearly enough to have a super clear idea of what is best (which is as rigorous as we justifiably want, but also maintains the approachability of metamath by people who don't know this math already).

benjub:
To answer a part of @jkingdon's post: this MR does not solve everything, far from it !  The more "urgent" and pratical need that emerged in #3199 was to make sure that every theorem depending on df-cleq (resp. df-clel) depends also on ax-9 (res. ax-8).

In the latest (and last in this MR, hopefully), I added two theorems proved from df-clab only. Actually, I moved one from my mathbox. The recent change in the definition df-sb enabled to use sbequ (which before used auxiliary axioms) to shorten the proof sizably, as noticed by @icecream17 . @icecream17 : is the way I credited the theorems fine with you ?

I'm working on doing that too for df-cleq and df-clel but I'd like to to it in a future MR and merge this one now.

> Although many topics are brought up there, I guess the crux of it is the following comment from Norm:
>>    Once we get to interesting results in set theory, all of the FOL= axioms will pretty much be used anyway. Effort spent to eliminate the use of this or that FOL= axiom in the set theory development doesn't seem productive to me, and it may even be confusing to the reader to see longer proofs that may result just to avoid using a certain FOL= axiom in a set theory theorem.

quotes the guy who routinely removes ax-3 from the most advanced theorems ;-) More seriously, the goal of this MR is not to remove axiom dependencies from some proofs but to make more transparent the relative dependencies among them. I agree with the rest of your comment.

set.mm
```diff
@@ -24886,20 +24932,25 @@ of logic but rather presupposes the Axiom of Extensionality (see theorem
...
-       theorem ~ bj-ax9 .  We may revisit this in the future.  (Revised by NM,
-       28-Oct-2021.)  (Proof modification is discouraged.) $)
+    $d x y z t u v A $.  $d x y z t u v B $.
+    $( The defining characterization of class equality.  It is proved, over
+       Tarski's FOL, from the axiom of (set) extensionality ( ~ ax-ext ) and
+       the definition of class equality ( ~ df-cleq ).  Its forward implication
+       is called "class extensionality".  Remark: the proof uses ~ axext4 to
+       prove also the hypothesis of ~ df-cleq that is a degenerate instance,
+       but it could be proved also from minimal propositional calculus and {
+       ~ ax-gen , ~ ax-4 }.  (Contributed by NM, 15-Sep-1993.)  (Revised by BJ,
```
icecream17, line 24948: I don't think this is possible: the left hand side is `t = t` (equid) and the right hand side is ~ id + ~ ax-gen

set.mm
```diff
@@ -25515,21 +25553,88 @@ replaced with setvar variables (see ~ cleljust ), so we don't include
...
+       'ax-ext' 'df-clab' 'df-cleq'; $)
+  $}
+
   ${
     $d z x $.  $d z y $.  $d z A $.
     $( Weaker version of ~ eleq1 (but more general than ~ elequ1 ) not
-       depending on ~ ax-ext nor ~ df-cleq .  (Contributed by BJ,
-       24-Jun-2019.) $)
+       depending on ~ ax-ext nor ~ df-cleq .  Note that this proof does not
+       depend on ~ ax-8 (while ~ eleq2w depends on ~ ax-9 ), so this also gives
```
icecream17, line 25617: "if dfclel didn't use ax-8" is probably needed before "this proof does not depend on ax-8"

set.mm
```diff
@@ -25515,21 +25553,88 @@ replaced with setvar variables (see ~ cleljust ), so we don't include
...
+       a proof of ~ ax-8 from Tarski's FOL and ~ dfclel (simply consider an
+       instance where ` A ` is replaced by a setvar and deduce the forward
+       implication by ~ biimpd ), which shows that ~ dfclel is too powerful to
+       be used a definition instead of ~ df-clel . taking an instance of the
+       forward implication where a setvar is substituted for ` A ` (Contributed
```
icecream17, line 25622: The last sentence seems like an incomplete thought (and has typos)

icecream17 (commented):
(approve when this is done, or may happen in a later pr)

set.mm
```diff
@@ -24945,8 +24945,8 @@ the definition of class equality ( ~ df-cleq ).  Its forward implication
        is called "class extensionality".  Remark: the proof uses ~ axext4 to
        prove also the hypothesis of ~ df-cleq that is a degenerate instance,
        but it could be proved also from minimal propositional calculus and {
-       ~ ax-gen , ~ ax-4 }.  (Contributed by NM, 15-Sep-1993.)  (Revised by BJ,
-       24-Jun-2019.) $)
+       ~ ax-gen , ~ ax-4 , ~ equid }.  (Contributed by NM, 15-Sep-1993.)
```
icecream17, line 24948: Oh that's simpler than I thought, I guess ax-4 would be removed, then too
(because id ax-gen equid)

benjub: of course ! I shouldn't do flash edits at midnight !

icecream17 (approved):

set.mm
```diff
@@ -46248,7 +46353,7 @@ conditioning it (with ` x e. z ` ) so that it asserts the existence of a
     $( Alternate proof of ~ axnul , proved from propositional calculus,
        ~ ax-gen , ~ ax-4 , ~ sp , and ~ ax-rep .  To check this, replace ~ sp
        with the obsolete axiom ~ ax-c5 in the proof of ~ axnulALT and type the
-       Metamath program "MM> SHOW TRACE_BACK axnulALT / AXIOMS" command.
+       Metamath program "MM> SHOW TRACE axnulALT / AXIOMS" command.
```
avekens, line 46356: The command is "show trace_back", but "show trace" works as abbreviation.

avekens (approved):

mazsa:
@digama0 wouldn't you like to weigh in on Benoit's request?

set.mm
benjub: Indeed. I would prefer to keep the full command, but on the webpage it appears as in index like TRACE<sub>BACK </sub> which is why I made that modification.

benjub:
> @digama0 wouldn't you like to weigh in on Benoit's request?

I think I will merge this MR since I fear conflicts. We can discuss further in later MRs on the topic. I'll probably try to write up Mario's proof of conservativity in https://us.metamath.org/mpeuni/mmnotes.txt soon.  This is basically Levy's proof of conservativity, with Mario's additions to adapt it to the case of set.mm. The precise statement will probably be weaker (i.e. "object level" and not "scheme level").

set.mm
avekens: You are right, and there are several MINIMIZE_WITH in set.mm which are displayed in the same ugly way. In 2019, I proposed that two consecutive underscores `__` should always be treated as literal `_`   (see google group https://groups.google.com/g/metamath/c/r3pwUvhI3NQ/m/fcEmU5OkAwAJ ). I do not know, however, if this was ever implemented. Currently, there aren't any two consecutive underscores `__`  in comments of set.mm.

mazsa:
> > Once we get to interesting results in set theory, all of the FOL= axioms will pretty much be used anyway.  Effort spent to eliminate the use of this or that FOL= axiom in the set theory development doesn't seem productive to me, and it may even be confusing to the reader to see longer proofs that may result just to avoid using a certain FOL= axiom in a set theory theorem.
> 
> My reaction to that is that if this is our position, we need to get a lot more specific in terms of how we talk about axiom usage. Somewhat in documentation, but in tools too.
> 
> I'm kind of inclined to think the current pull request is a step forward even if I'm not sure it solves everything. But I'm not sure I see any of this clearly enough to have a super clear idea of what is best (which is as rigorous as we justifiably want, but also maintains the approachability of metamath by people who don't know this math already).

It seems to me that what we may need are the actually valid usage equivalence classes of axioms and explicit preferences among them, see also https://us.metamath.org/mpeuni/mmset.html#subsys . + a description of how to check them like in case of proof lenght minimize_with * /include_mathboxes

mazsa:
> + a description of how to check them like in case of proof lenght minimize_with * /include_mathboxes

https://github.com/metamath/set.mm/pull/3399#issuecomment-1684917969

PR #3391, opened 2023-08-17
Remove ax-rep from some theorems

CatsAreFluffy:
I removed ax-rep from grpinvfval, grpsubfval, mulgfval, and odfval (and theorems whose only reference to ax-rep went through those). I added some theorems related to functions and sequences to help with that. I also added a DV condition to odcl and other theorems in its block, which I didn't think needed a (Revised by ...) tag since the change only affected a dummy variable, and I added a section header for relation and function theorems using ax-un since it seemed weird that there wasn't one.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15549,6 +15549,7 @@ New usage of "grothomex" is discouraged (0 uses).
 New usage of "grothpw" is discouraged (0 uses).
 New usage of "grothpwex" is discouraged (1 uses).
 New usage of "grpbasex" is discouraged (3 uses).
+New usage of "grpinvfvalOLD" is discouraged (0 uses).
 New usage of "grpo2inv" is discouraged (4 uses).
 New usage of "grpoass" is discouraged (16 uses).
 New usage of "grpocl" is discouraged (12 uses).
@@ -15595,6 +15596,7 @@ New usage of "grporn" is discouraged (11 uses).
 New usage of "grporndm" is discouraged (4 uses).
 New usage of "grposnOLD" is discouraged (1 uses).
 New usage of "grpplusgx" is discouraged (3 uses).
+New usage of "grpsubfvalOLD" is discouraged (0 uses).
 New usage of "gsumsplOLD" is discouraged (0 uses).
 New usage of "gt-lt" is discouraged (0 uses).
 New usage of "gt-lth" is discouraged (1 uses).
@@ -16614,6 +16616,7 @@ New usage of "mulcompr" is discouraged (6 uses).
 New usage of "mulcomsr" is discouraged (5 uses).
 New usage of "mulerpq" is discouraged (3 uses).
 New usage of "mulerpqlem" is discouraged (1 uses).
+New usage of "mulgfvalOLD" is discouraged (0 uses).
 New usage of "mulgt0sr" is discouraged (2 uses).
 New usage of "mulidnq" is discouraged (11 uses).
 New usage of "mulidpi" is discouraged (5 uses).
@@ -16958,6 +16961,7 @@ New usage of "ocorth" is discouraged (3 uses).
 New usage of "ocsh" is discouraged (6 uses).
 New usage of "ocss" is discouraged (11 uses).
 New usage of "ocval" is discouraged (4 uses).
+New usage of "odfvalOLD" is discouraged (0 uses).
 New usage of "ogrpinvOLD" is discouraged (0 uses).
 New usage of "oldmm3N" is discouraged (2 uses).
 New usage of "olposN" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -70681,6 +70681,13 @@ Finite induction (for finite ordinals)
       FUGRRDUHUI $.
   $}
 
+
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  Relations and functions (cont.)
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
```
avekens, line 70689: That is a very good idea! It was confusing that theorems like dmexg etc. were contained in a section called "Finite induction (for finite ordinals)".

avekens:
> I removed ax-rep from grpinvfval, grpsubfval, mulgfval, and odfval (and theorems whose only reference to ax-rep went through those). 

I don't know if it makes sence on this level of higher mathematics to reduce the number of axioms, making the proofs much longer. Therefore, I would vote for keeping the old proofs.

> I added some theorems related to functions and sequences to help with that. 

I like these *exw theorems, these can be added.

> I also added a DV condition to odcl and other theorems in its block, which I didn't think needed a (Revised by ...) tag since the change only affected a dummy variable, 

Why is an additional dummy variable necessary?

> and I added a section header for relation and function theorems using ax-un since it seemed weird that there wasn't one.

That's great.

benjub:
I agree with @avekens : keep the proofs using ax-rep as xxxALT with comment "Shorter proof of ~ xxx using ~ ax-rep ." and add to the comment of xxx : "For a shorter proof using ~ ax-rep , see ~ xxxALT .".

@avekens : ax-rep is in some sense the "strongest" axiom of ZF, so it is worth the effort IMO to remove it when possible. Actually, Zermelo set theory (called "Z", so without the "F" of "Fraenkel") is very close to ZF minus replacement.

set.mm
```diff
@@ -71054,6 +71061,28 @@ Axiom of Replacement (unlike ~ resfunexg ).  (Contributed by NM,
...
+
+  ${
+    $d x A $.  $d x C $.
+    mptexw.1 $e |- A e. _V $.
+    mptexw.2 $e |- C e. _V $.
+    mptexw.3 $e |- A. x e. A B e. C $.
+    $( Weak version of ~ mptex that holds without ~ ax-rep .  If the domain and
+       codomain of a function given by maps-to notation are sets, the function
+       is a set.  (Contributed by Rohan Ridenour, 13-Aug-2023.) $)
+    mptexw $p |- ( x e. A |-> B ) e. _V $=
```
benjub, line 71080: @avekens : wouldn't we prefer to have a weak version of mptexg instead ?

CatsAreFluffy:
> Why is an additional dummy variable necessary?

I meant that a new DV condition between existing variables (`x` and `w`), one of which is a dummy variable, needed to be added. There aren't any new dummy variables, and the proof of ~ odval is unchanged.

> I agree with @avekens : keep the proofs using ax-rep as xxxALT [...]

Done.

set.mm
avekens: Yes, a weak version mptexgw of mptexg would be preferred, and mptexw could be derived from it easily (in the same way as mptex is derived from mptexg).

wlammen:
A meanwhile renaming of theorems clashes with this PR.
"Proof of theorem grpsubfval refers to mpt20 which is not an active statement"
See changes-set.txt: "17-Aug-23 mpt20     mpo0"

wlammen:
Delete theorem mpt20 after this PR got merged.

jkingdon:
> A meanwhile renaming of theorems clashes with this PR. "Proof of theorem grpsubfval refers to mpt20 which is not an active statement" See changes-set.txt: "17-Aug-23 mpt20 mpo0"

Thanks for your efforts to handle this situation.

I'm somewhat inclined to pause on the mpt2 renames long enough for some of the existing pull requests to work through the queue (I guess a day or a few), but if people would prefer that I continue ripping off the mpt2 bandaid at (roughly) the current rate, I can do that too.

CatsAreFluffy:
Merge?

set.mm
benjub: I'm going to merge. Theorem mptexw could be replaced with mptexgw in a later PR (but you don't have to).

set.mm
```diff
@@ -140327,6 +140374,27 @@ and complete (infinite) series.  (Contributed by NM, 18-Apr-2005.)
       DVQPPWAWEWBVRWDANVSDRVTWDVRAVSDLBMVEUIVRVQWDAVFWBTVQWDAVGVIVJVK $.
   $}
 
+  ${
+    $d x .+ $.  $d x F $.  $d x y M $.  $d y z .+ $.  $d y z F $.  $d y z M $.
+    seqexw.1 $e |- .+ e. _V $.
+    seqexw.2 $e |- M e. ZZ $.
+    $( Weak version of ~ seqex that holds without ~ ax-rep .  A sequence
+       builder exists when its binary operation input exists and its starting
+       value is an integer.  (Contributed by Rohan Ridenour, 14-Aug-2023.) $)
```
benjub, line 140383:
```suggestion
       index is an integer.  (Contributed by Rohan Ridenour, 14-Aug-2023.) $)
```

CatsAreFluffy: Done.

wlammen:
Reverted temporary fix in #3408

PR #3392, opened 2023-08-17
Fix spacing around Scott and Coll

CatsAreFluffy:
Fixes missing spacing around the `Scott` and `Coll` operators added in #3378, cf [~ scottss](https://us.metamath.org/mpeuni/scottss.html) and [~ df-coll](https://us.metamath.org/mpeuni/df-coll.html).


tirix (approved):

avekens (approved):

benjub:
BTW: I do not know if it is essential that Scott and Coll be defined for all classes, or if it would be sufficient to define them only for sets. In the latter, case, we prefer in set.mm to not introduce new syntax, but define functions like `Scott = ( x e. _V -> ... )` and `Coll = ( f e. _V , a e. _V -> ... )`
Of course it can stay like that in your mathbox.

benjub (approved):

CatsAreFluffy:
The mainspace theorems involving Scott's trick, which I expect would be revised to use ~ df-scott if that was mainspaced, depend on its applicability to classes (eg ~ scottex is trivial if you already know `A` is a set). As such, a set-only version of df-scott wouldn't be useful in the mainspace.

avekens:
> BTW: I do not know if it is essential that Scott and Coll be defined for all classes, or if it would be sufficient to define them only for sets. In the latter, case, we prefer in set.mm to not introduce new syntax, but define functions like `Scott = ( x e. _V -> ... )` and `Coll = ( f e. _V , a e. _V -> ... )` 

That is exactly what I also was about to ask, but @CatsAreFluffy answered this question already...

Issue #3396, opened 2023-08-18
Defining countability in terms of a surjection from an element of suc _om


jkingdon:
The pull request #3397  contains the following definition of countability:

```
    $( A set is countable iff there is a surjection from a decidable subset of
       the natural numbers onto it.  Based on a definition proposed by David
       theHigherGeometer oneline.  The decidability condition is needed as
       shown at ~ ctssdc .  (Contributed by Jim Kingdon, 15-Aug-2023.) $)
    ctssdc $p |- ( E. s ( s C_ _om /\ E. f f : s -onto-> A
        /\ A. n e. _om DECID n e. s )
        <-> E. f f : _om -onto-> ( A |_| 1o ) ) $=
```

This issue proposes a variation of this which is similar in spirit:

```
    $( A set is countable iff there is a surjection from a natural number, or
       the set of natural numbers, onto it.  (Contributed by Jim Kingdon,
       15-Aug-2023.) $)
    ctomp1 $p |- ( E. s e. ( _om u. { _om } ) E. f f : s -onto-> A
        <-> E. f f : _om -onto-> ( A |_| 1o ) ) $=
      (  ) ? $.
```

Presumably this is a relatively easily corollary. I suppose you'd prove it by first proving

```
ssom $p |- ( ( S C_ _om /\ A. n e. _om DECID n e. S )
  -> E. n e. ( _om u. { _om } ) n ~~ S ) $=
```

which would be potentially applicable elsewhere, I suppose.

jkingdon:
One comparison is that the proposed theorem would appear to be quite similar to the way we express "finitely enumberable" at https://us.metamath.org/ileuni/fidcenum.html

Issue #3394, opened 2023-08-18
is the union of countably many countable sets countable?


jkingdon:
Andrej Bauer calls this the "litmus test" for a definition of countability ( https://mathstodon.xyz/@andrejbauer/110906869535920346 ). Is this provable given the ∃𝑓 𝑓:ω–onto→(𝐴 ⊔ 1o) definition that we use in most of our countability theorems in iset.mm? Does it require countable choice?

I'm not sure where I'd go to research this further, though.

jkingdon:
I have the sketch of what appears to be a proof of this, but I've hit a snag and I'm not sure whether it is just that I haven't fully gotten the hang of nested quantifiers ("for every x in A there exists a f such that....") or whether I'm attempting to use what amounts to the axiom of choice, or what.

My statement of what we want to prove is:

```
  ${
    ctiunct.a $e |- ( ph -> E. f f : _om -onto-> ( A |_| 1o ) ) $.
    ctiunct.b $e |- ( ( ph /\ x e. A )
      -> E. g g : _om -onto-> ( B |_| 1o ) ) $.
    $( The union of countably many countable sets is countable.

       The proof proceeds by mapping a natural number to a pair of natural
       numbers (by ~ xpomen ) and using the first number to map to an element
       ` x ` of ` A ` and the second number to map to an element of B(x) .  In
       this way we are able to map to every element of ` U_ x e. A B ` .
       Although it would be possible to work directly with countability
       expressed as ` E. f f : _om -onto-> ( A |_| 1o ) ` , it is easier to
       perform this proof by transforming to functions from subsets of the
       natural numbers via ~ ctssdc .

       (Contributed by Jim Kingdon, 24-Oct-2023.) $)
    ctiunct $p |- ( ph -> E. h h : _om -onto-> ( U_ x e. A B |_| 1o ) ) $=
      ? $.
  $}
```

The nested quantifier there is "for every x in A there exists a g".

The key lemma (most of which I"ve proved) is:

```
$d A n x $.  $d B n $.  $d n F x $.  $d z F x $.  $d n G $.  $d n J x $.  $d z
  J $.  $d z S $.  $d z T $.  $d U n $.  $d ph n x $.
ctiunct.som $e |- ( ph -> S C_ _om ) $.
ctiunct.sdc $e |- ( ph -> A. n e. _om DECID n e. S ) $.
ctiunct.f $e |- ( ph -> F : S -onto-> A ) $.
[...]

PR #3397, opened 2023-08-18
Add an equivalent definition of countability to iset.mm 

jkingdon:
This pull request proves:

```
    $( A set is countable iff there is a surjection from a decidable subset of
       the natural numbers onto it.  Based on a definition proposed by David
       theHigherGeometer oneline.  The decidability condition is needed as
       shown at ~ ctssdc .  (Contributed by Jim Kingdon, 15-Aug-2023.) $)
    ctssdc $p |- ( E. s ( s C_ _om /\ E. f f : s -onto-> A
        /\ A. n e. _om DECID n e. s )
        <-> E. f f : _om -onto-> ( A |_| 1o ) ) $=
```

The motivation is a discussion thread starting with https://mathstodon.xyz/@highergeometer/110892635391287895 especially the part about "I probably should have done the even better definition which is that there is a surjection from a [decidable] subset of NN, which means you don't need to worry about the empty set, and also the finite case is more cleanly captured"


jkingdon:
I updated the comment based on https://mathstodon.xyz/@highergeometer/110913478065960269 . There probably isn't much value to a vague citation like that anyway (and a specific URL is likely to bit-rot at least on a scale of years).

benjub (approved):

benjub:
> I updated the comment based on https://mathstodon.xyz/@highergeometer/110913478065960269 . There probably isn't much value to a vague citation like that anyway (and a specific URL is likely to bit-rot at least on a scale of years).

That's an interesting discussion. Maybe you can find a way to write the gist of it somewhere in a theorem comment or section comment in iset.mm or in index.html ? (or maybe it's already there, I haven't checked).

Having these results formalized in metamath is very nice.

From the picture, I guess that you are "soaproot" there. Do you know who "Higher Geometer" is ? And more importantly, do you know who ℵ_ω / ax_groth is ? I think it is problematic that that user uses the Metamath logo. Since his description only says "check out https://us.metamath.org/", one could think this is the official account of Metamath.

DavidMichaelRoberts:
@benjub I'm HigherGeometer :-)

iset.mm
```diff
@@ -67983,6 +67983,135 @@ property of disjoint unions (see ~ updjud in the case of functions).
...
+      N $.
+  $}
+
+  ${
+    $d A f g m s $.  $d A f g x $.  $d A f m s y z $.  $d f m n s y z $.
+    $d A n $.
+    $( A set is countable iff there is a surjection from a decidable subset of
+       the natural numbers onto it.  The decidability condition is needed as
+       shown at ~ ctssdc .  (Contributed by Jim Kingdon, 15-Aug-2023.) $)
+    ctssdc $p |- ( E. s ( s C_ _om /\ E. f f : s -onto-> A
```
icecream17, line 68082: the theorem is quoting itself? + maybe you meant ctssexmid below

(approve when done)

jkingdon:
> That's an interesting discussion. Maybe you can find a way to write the gist of it somewhere in a theorem comment or section comment in iset.mm or in index.html ? (or maybe it's already there, I haven't checked).

Well, the current pull request is for some of it and https://github.com/metamath/set.mm/issues/3396 and https://github.com/metamath/set.mm/issues/3394 are for others. As a formal math project, I'd have us focus on what we prove rather than trying to recreate wikipedia or ncatlab.

> From the picture, I guess that you are "soaproot" there.

Correct. 

> And more importantly, do you know who ℵ_ω / ax_groth is ? I think it is problematic that that user uses the Metamath logo. Since his description only says "check out https://us.metamath.org/", one could think this is the official account of Metamath.

Based on a quick browse, appears to be a student who is a fan of metamath. I guess to me it doesn't look much like an official account, so I'm not especially concerned, but whatever happens I hope we try to encourage their enthusiasm. Who knows, maybe they'll end up being more involved with the project at some point.

iset.mm
jkingdon: Yes, it is supposed to be ctssexmid

Thanks for noticing this.

PR #3398, opened 2023-08-19
typos, sb-direction, misc grp theory

icecream17:
mostly mathbox

I prove some grp theorems, which would theoretically be helpful to even further simplify 0prjspn

---

More on the todo:
"An elliptic curve is a smooth, projective, algebraic curve of genus one"
Similar to the bijection between 3-dimensional affine space and 2-dimensional projective space, [there are polynomials with two variables and homogeneous polynomials with three variables](https://en.wikipedia.org/wiki/Algebraic_curve).

(So in general more theorems about multivariable polynomials need to be proven)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+18-Aug-23 eqsnd     [same]      moved from TA's mathbox to main set.mm
 17-Aug-23 mpt2mpt   mpompt
 17-Aug-23 mpt2mptx  mpomptx
 17-Aug-23 mpt2v     mpov
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,7 +26,6 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
-18-Aug-23 eqsnd     [same]      moved from TA's mathbox to main set.mm
 17-Aug-23 mpt2mpt   mpompt
 17-Aug-23 mpt2mptx  mpomptx
 17-Aug-23 mpt2v     mpov


set.mm
```diff
@@ -619955,6 +619979,73 @@ fixed reference functional determined by this vector (corresponding to
       ( clvec wcel clmod crg lveclmod lmodring syl ) BDEBFEAGEBHABCIJ $.
   $}
 
+  $( The property of modules to be vector spaces is invariant.  (Contributed by
+     Steven Nguyen, 15-Aug-2023.) $)
```
benjub, line 619983:
```suggestion
  $( The property for modules to be vector spaces is invariant under module isomorphism.  (Contributed by
     Steven Nguyen, 15-Aug-2023.) $)
```

set.mm
```diff
@@ -619955,6 +619979,73 @@ fixed reference functional determined by this vector (corresponding to
...
+    KAQFBQFUCUDUGUFUIUCUDUGABCLABCMNUCUEUHJUCUHUEABCUEUHUEOZUHOZPRSUAUEAUJTUHBU
+    KTUB $.
+
+  ${
+    $d I x y $.  $d K u x y $.  $d F x y $.  $d W t u x y $.  $d I t u $.
+    frlmsnic.w $e |- W = ( K freeLMod { I } ) $.
+    frlmsnic.1 $e |- F = ( x e. ( Base ` W ) |-> ( x ` I ) ) $.
+    $( A module whose vectors map a single index to a value is isomorphic to
+       the module whose vectors are a value.  (Contributed by Steven Nguyen,
+       18-Aug-2023.) $)
```
benjub, line 619996: The result is more like:
Given a free module with a singleton as base, the function that maps each vector to its coordinate in that base is a module isomorphism from that module to its ring of scalars seen as a module.

set.mm
```diff
@@ -621195,26 +621286,25 @@ with respect to a linear (1-dimensional) relation.  (Contributed by BJ
   ${
     $d W a b l x y $.  $d K a b l x y $.  $d B a b x y $.
     0prjspn.w $e |- W = ( K freeLMod ( 0 ... 0 ) ) $.
-    0prjspn.0 $e |- .0. = ( 0g ` W ) $.
-    0prjspn.b $e |- B = ( ( Base ` W ) \ { .0. } ) $.
+    0prjspn.b $e |- B = ( ( Base ` W ) \ { ( 0g ` W ) } ) $.
```
benjub, line 621289: Just a remark: I have experimented from time to time with "inlining local definitions" (like `.0.` here). The statement looks better, but the usability will sometimes lessen a bit. There are tradeoffs.

set.mm
```diff
@@ -619955,6 +619979,73 @@ fixed reference functional determined by this vector (corresponding to
...
+      UUDUVPUXNKZKZUVNUXMQZUWCUVOQZUYCUYDUYEUYCUYDKZUVOCUXMLZUWCUYDUVOUYGQUYCCU
+      VNUXMSVHUYFUUCUWCJIUYGUWCQUUBUUCUYBUYDYBHYCCUWCJJYDXSYEYFUYCUYDUYEUVNCUVO
+      XEZUPZQZUYCUUCUVNUUSYGZUYJUUBUUCUYBXJUYCUUSUVJUVNUUDUVPUVQUXNUVSWMYHUUCUY
+      KUYJCUVNYIYMYJUYEUXMUYIUVNUYEUXLUYHUWCUVOCYKYLYNYOYPYQUUDUVJUUGUUFBUVJUUG
+      QUUDUVKRYRYSUUFUUGEUUEBUUNUUGOYTUUA $.
+  $}
+
+  $( TODO: It would be nice to have a function converting free modules to
+     normed modules, a la ~ rlmnlm .  The norm of an n-dimensional vector would
+     be the square root of each of its coordinates squared.
```
avekens, line 620042: I think such a function exists already, it is `toCPreHil`. Applied to the free module over the real numbers (regarded as field) results in (generalized) Eucliedean spaces, as defined in ~df-rrx

benjub:
Side remark: the title says "sb-direction" but there is no such theorem ? More generally, the initial post of each PR, and even more importantly its title, will stay in the git history, so it is important that it describes, even roughly, the contents of the PR. (And, yes, I have not always been a model in this regard!).

jkingdon:
There are merge conflicts because `changes-set.txt` was modified both here and in a mpt2 rename pull request.

I'm somewhat inclined to pause on the mpt2 renames long enough for some of the existing pull requests to work through the queue, but if people would prefer that I continue ripping off the mpt2 bandaid at (roughly) the current rate, I can do that too.

icecream17:
I'll remove the move to main since it didn't end up being used

jkingdon:
> I'll remove the move to main since it didn't end up being used

LOL,  I guess in this case that solution works.

icecream17:
> Side remark: the title says "sb-direction" but there is no such theorem ? More generally, the initial post of each PR, and even more importantly its title, will stay in the git history, so it is important that it describes, even roughly, the contents of the PR. (And, yes, I have not always been a model in this regard!).

This refers to sbn1 and sbor2

set.mm
```diff
@@ -620078,6 +620162,20 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
       ULUMUPUQBCDJQUN $.
   $}
 
+  ${
+    $d x y z A $.
+    $( A natural number that is not 1 is a successor of some other natural
+       number.  (Contributed by Steven Nguyen, 19-Aug-2023.) $)
```
benjub, line 620168:
```suggestion
    $( A nonzero natural number that is not 1 is a successor of some other
       nonzero natural number.  (Contributed by Steven Nguyen, 19-Aug-2023.) $)
```
Sorry, I missed that on first reading. You can probably apply the suggestion directly here.

benjub (approved):

set.mm
icecream17: I put positive integer a la nnaddcl

benjub: Well, it's less unambiguous than "nonzero natural", but as you prefer.

jkingdon (approved):

PR #123, opened 2017-07-07
Ltrr discouraged

david-a-wheeler:

PR #3399, opened 2023-08-19
Reduced axiom usage

GinoGiotto:
Proof edits of  ~ sbelx, ~ cleqf, ~ reeanv, ~ rabeqif, ~ rabeq, ~ rabeqi, ~ rabrabi, ~ cgsex4g, ~ ceqsex2v,  ~ vtocl3, ~ vtoclga, ~ vtocl2ga, ~ spcimdv, ~ spc3egv reduce the following axiom dependencies:

* Dropped `ax-10` for 93 theorems.
* Dropped `ax-11` for 47 theorems.
* Dropped `ax-12` for 10 theorems.
* Dropped `ax-7` for 3 theorems.
* Dropped `ax-6` for 3 theorems.

Details here: e9e5b3e

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18990,7 +18990,7 @@ Proof modification of "exlimexi" is discouraged (15 steps).
 Proof modification of "exmoOLD" is discouraged (22 steps).
 Proof modification of "exmoeuOLD" is discouraged (39 steps).
 Proof modification of "exsbOLD" is discouraged (32 steps).
-Proof modification of "extwwlkfabOLD" is discouraged (381 steps).
+Proof modification of "extwwlkfabOLD" is discouraged (374 steps).
 Proof modification of "extwwlkfabelOLD" is discouraged (139 steps).
 Proof modification of "f1cofveqaeqALT" is discouraged (124 steps).
 Proof modification of "f1oweALT" is discouraged (805 steps).


benjub:
That's very nice! Can you move sbelx earlier ?

The $a-statements of set.mm determine a sequence of intervals, and each theorem (in the Main part) should be in the earliest possible interval. (There are a lot of theorems to move... but this can be done progressively.)

benjub (approved):
Ah, sorry, you couldn't move it because it uses ax-12. My bad.

benjub:
By the way: https://github.com/metamath/set.mm/commit/e9e5b3ea18d70aad432e625ff1d358825167120d is very useful to review MRs. How do you generate it ?

GinoGiotto:
> By the way: [e9e5b3e](https://github.com/metamath/set.mm/commit/e9e5b3ea18d70aad432e625ff1d358825167120d) is very useful to review MRs. How do you generate it ?

I'm using the new tool made by @digama0 in https://github.com/metamath/metamath-knife/pull/107 that lists axiom dependencies of every theorem in a database. It's much better than metamath-exe because of its speed, but it could be even better if it would have options to list theorem dependencies as well (like metamath-exe does with `SHOW TRACE_BACK * /MATCH <matched-labels>`, but with the speed advantage of metamath-knife). This would be useful for example to replace dependencies from `nfmo` to `nfmodv`, as I found an automated procedure that seems promising. The only issue is that I need to list all theorems that depend on `nfmo` first, and for that I have to use metamath-exe which is slow.

The command is: `metamath-knife <database-name> --axiom-use <output-file>`. I execute it for the version of set.mm before my change and after the change, so that I can compare them.

set.mm
```diff
@@ -30533,22 +30537,24 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
     $( Equality theorem for restricted class abstractions.  Inference form of
        ~ rabeqf .  (Contributed by Glauco Siliprandi, 26-Jun-2021.) $)
     rabeqif $p |- { x e. A | ph } = { x e. B | ph } $=
-      ( wceq crab rabeqf ax-mp ) CDHABCIABDIHGABCDEFJK $.
+      ( wceq crab cv wcel wa cab nfth eleq2 anbi1d abbid df-rab 3eqtr4g ax-mp )
```
wlammen, line 30540: It seems that hypotheses rabeqf.1 and rabeqf.2 ar not referenced in  the proof.  Get rid of them.

set.mm
```diff
@@ -30533,22 +30537,24 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
...
   ${
     $d x A $.  $d x B $.
     $( Equality theorem for restricted class abstractions.  (Contributed by NM,
        15-Oct-2003.) $)
     rabeq $p |- ( A = B -> { x e. A | ph } = { x e. B | ph } ) $=
-      ( nfcv rabeqf ) ABCDBCEBDEF $.
+      ( wceq cv wcel eleq2 anbi1d rabbidva2 ) CDEZAABCDKBFZCGLDGACDLHIJ $.
 
     ${
       rabeqi.1 $e |- A = B $.
```
wlammen, line 30553: This is identical to rabeqif if the two nf-hypotheses are dropped.  You don't need the two dv conditions either.  Drop this theorem.

set.mm
GinoGiotto: I don't understand, if I write:

<pre>MM> sh pr rabeqif /lem
13 rabeqif.3     $e |- A = B
18 rabeqf.1      $e |- F/_ x A
19 rabeqf.2      $e |- F/_ x B
20 18,19 rabeqf  $p |- ( A = B -> { x e. A | ph } = { x e. B | ph } )
21 13,20 ax-mp   $a |- { x e. A | ph } = { x e. B | ph }</pre>

They do appear in the proof, also when I hide them from `rabeqif` the metamath-exe verifier complains.

set.mm
avekens: The same would hold for rabeqi then (which would be identical with rabeqif which could also be dropped then).

set.mm
wlammen: The reason why they appear in the old proof, is to show F/ x A = B .  But since A = B is a hypothesis, your approach via nfth effectively arrives at the same result, without the hypotheses in question.  They can be dropped from the new (but not the old) proof.

set.mm
```diff
@@ -30532,14 +30532,16 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
     rabeqf $p |- ( A = B -> { x e. A | ph } = { x e. B | ph } ) $=
       ( wceq cv wcel wa cab crab nfeq eleq2 anbi1d abbid df-rab 3eqtr4g ) CDGZB
       HZCIZAJZBKTDIZAJZBKABCLABDLSUBUDBBCDEFMSUAUCACDTNOPABCQABDQR $.
+  $}
 
+  ${
     rabeqif.3 $e |- A = B $.
```
benjub, line 30538: So I guess it's now
rabeqif.3 --> rabeqi.1
and rabeqif --> rabeqi for the conclusion.

set.mm
GinoGiotto: You're right, I applied the suggestions in commits c2597c9 and 5cc41dd.

Indeed revising the proofs of `extwwlkfab` `extwwlkfabOLD` `smfliminf` wasn't easy at all, but I managed to do it in the end.

wlammen (approved):

benjub (approved):

set.mm
```diff
@@ -423711,7 +423723,7 @@ version of it and the OLD version is obsolete (and will be removed
        variable replaced by a disjoint variable condition "v" with a
        conjunction as antecedent "a").  As a general rule, the suffixes for
        the theorem forms ("i", "d" or "g") should be the first of multiple
-       suffixes, as for example in ~ vtocldf or ~ rabeqif .
+       suffixes, as for example in ~ vtocldf or ~ rabeqi .
```
avekens, line 423726: This is not an example for this case anymore!

avekens (changes requested):
I am a little bit sceptical about these changes, making proofs much longer, without indications in the comments. At least, a comment like "Avoid ~ ax-..." together with a "Revised by" tag should be added. 
Furthermore, I do not know if a "Proof shortened" tag should be kept (as in ~ cleqf) if a proof becomes longer now.

set.mm
GinoGiotto: I added the commit suggestions in a bit of a hurry, I'll take some time to double check again.

wlammen:
In many instances the proofs became longer, because a call to a theorem was expanded inline.  Doing so then allowed replacements that lead to smaller axiom footprints.  That is, the prior state was unfortunate, because it split a sequence of proof lines expressing an idiom (so to say) among a caller and callee.  This effectively prevented the recognition and proper replacement of the idiom before.  Is it desirable to keep this situation, for the sake of shorter proofs?  In my eyes not, because the line (interface) drawn between calling and called theorem, was imperfect.  So the first step is to destroy the unsound structure.  Then you rearrange the fragments into better designed theorems and lemmas again.  If I had spotted the weak points (which I did not, despite my experience, see sbelx), then I possibly would have considered a better theorem design as part of the process.  Can we always expect this behavior?

The results presented here were found by some brute-force scanning algorithm, using a huge amount of CPU time (I hopefully remember this right).  If this is true, then in fact you cannot attribute the improvements to a human being.  That's perhaps why the tags are missing.

benjub:
I tend to agree with @wlammen : the proofs are probably longer because the existing intermediate lemmas were adapted to the previous proofs, and with analogous lemmas adapted to the new proofs, the proof lengths would probably be comparable.

However, we should be careful not to lose potential insight which is present in the shorter proofs, so we should keep them as *ALT if there is a doubt, and we shouldn't rush in any case. I agree with @avekens that a revision tag and brief reasons could be put when one judges it necessary (this is voluntarily vague). Even if the revision is, at the end, the output of an algorithm, this is also the result of @GinoGiotto's work.

avekens:
> In many instances the proofs became longer, because a call to a theorem was expanded inline. Doing so then allowed replacements that lead to smaller axiom footprints. That is, the prior state was unfortunate, because it split a sequence of proof lines expressing an idiom (so to say) among a caller and callee. This effectively prevented the recognition and proper replacement of the idiom before. Is it desirable to keep this situation, for the sake of shorter proofs? In my eyes not, because the line (interface) drawn between calling and called theorem, was imperfect. So the first step is to destroy the unsound structure. Then you rearrange the fragments into better designed theorems and lemmas again. If I had spotted the weak points (which I did not, despite my experience, see sbelx), then I possibly would have considered a better theorem design as part of the process. Can we always expect this behavior?
> 
It would be great to extract parts of the new proofs as lemmata. They should be analogs to the previous ones, I guess. But this can be done ina separate PR.

avekens:
> I tend to agree with @wlammen : the proofs are probably longer because the existing intermediate lemmas were adapted to the previous proofs, and with analogous lemmas adapted to the new proofs, the proof lengths would probably be comparable.

That is a very good argument to provide such lemmata. 

> However, we should be careful not to lose potential insight which is present in the shorter proofs, so we should keep them as *ALT if there is a doubt, and we shouldn't rush in any case.

Good idea: as long as the lemmata mentioned before are not available, the theorems with the former proofs should be kept as *ALT theorems.

> I agree with @avekens that a revision tag and brief reasons could be put when one judges it necessary (this is voluntarily vague). Even if the revision is, at the end, the output of an algorithm, this is also the result of @GinoGiotto's work.

There are not so many theorems, and the rabeq* theorems should be tagged by "Revised by" anyway (because of the changes in the dv conditions), so that the effort should be manageable.

wlammen:
cleqf/cleqh: OLD: Why was ax-10 unnecessarily referenced? cleqf transforms its hypotheses to those of cleqh using nfcriv + nf5ri, and applies this theorem afterwards.  And what does cleqh do? It immediately applies nf5i to the hypotheses, partially reverting the transformation.   NEW: Unfortunately, you cannot simply revert the roles of both theorems, since one depends on df-nfc, the other not, and this definition lies in between.  Instead, factor out the common part by writing a lemma 
Ⅎ𝑥 𝑦 ∈ 𝐴   &&   Ⅎ𝑥 𝑦 ∈ 𝐴   ==>   (𝐴 = 𝐵 ↔ ∀𝑥(𝑥 ∈ 𝐴 ↔ 𝑥 ∈ 𝐵))
The proof is that of cleqh, omitting line 4 and 6.  Then use this lemma in both theorems.  No need for ALT versions.

benjub:
Or just remove cleqh (well, rename it cleqhOLD) and keep only cleqf. The hypotheses are from the older idiom used to express nonfreeness, which has been replaced by the use of F/ and F/_ as in cleqf.

wlammen:
`Or just remove cleqh`
Not that easy, since cleqh has applications abeq2 and abbi, not relying on df-nfc.  This at least triggers cascading effects, beyond this technical correction.

benjub:
> > Or just remove cleqh
>
> Not that easy, since cleqh has applications abeq2 and abbi, not relying on df-nfc. This at least triggers cascading effects, beyond this technical correction.

You're right, it would go beyond this PR. We can do it later (e.g., the problem is that abeq2 uses hbab1 to prove a hypothesis of cleqh, but one has the corresponding lemma nfsab1 with the new idiom, etc.).

GinoGiotto:
@benjub The verifiers are failing after the merge of https://github.com/metamath/set.mm/pull/3391
https://github.com/metamath/set.mm/actions

wlammen:
The HTTP return code 429 https://dataloft.ch/wiki/429-http-statuscodes-too-many-requests/ simply states you exceeded your administrative limits, and too often requested https://github.com/metamath/set.mm from the verifying server.  Bad, but viewing the details of commit e09596f on https://github.com/metamath/set.mm/pull/3391 show all checks passed.  So for the moment I think, let's just ignore it.

wlammen:
reeanv: OLD It is a suspicious move to transform all distinct variable conditions to non-free hypotheses using nfv, and then forward the context to the more general theorem, in our case reean.  I often observed that ignoring dv condtions this way drags in additional axioms.  And reeanv is no exemption.  NEW: The observation here is that exdistrv is a valid replacement of eean used in the proof of reean on line 9, reducing the axiom footprint.  The simple solution is to replay the proof of reean in the context of reeanv, which is what you did.  A more elaborate version factors out common proof lines in a lemma
(∃𝑥∃𝑦((𝑥 ∈ 𝐴 ∧ 𝜑) ∧ (𝑦 ∈ 𝐵 ∧ 𝜓)) ↔ (∃𝑥(𝑥 ∈ 𝐴 ∧ 𝜑) ∧ ∃𝑦(𝑦 ∈ 𝐵 ∧ 𝜓)))
  ==> (∃𝑥 ∈ 𝐴 ∃𝑦 ∈ 𝐵 (𝜑 ∧ 𝜓) ↔ (∃𝑥 ∈ 𝐴 𝜑 ∧ ∃𝑦 ∈ 𝐵 𝜓))
instead, proven from lines 1, 2, 10 - 15 of reean.  Using it both in reeanv and reean leads to an overall shortening effect, measured in proof bytes.  No need for ALT, in any case BTW.

wlammen:
You need to have a space between a reference like ~ ax-10 and surrounding text.  So "Avoid ~ ax-10, ~ ax-11."  is bad. Correct "Avoid ~ ax-10 , ~ ax-11 ."

GinoGiotto:
I added the tags as requested and fixed the incorrect edited text. As already mentioned these PRs are the result of an automated script that I made with the purpose of spotting as much axiom minimizations as possible. At the moment I didn't let my computer spend too much time into it, since I'm still in an experimental phase to check what works best. The problem of adding tags and ALT proofs for every single theorem is that it might get unpractical very quickly in case I would find something that works well. So by default I don't add them, I prefer to wait for requests to specific theorems for specific reasons.

tirix:
> it could be even better if it would have options to list theorem dependencies as well

I fear dumping a file with all dependencies would result in something huge. Could we e.g. limit the output to e.g. just the list of theorems (directly or indirectly) depending on a given theorem?
What would be the most useful format for you?

avekens (approved):
Although the previous proofs are not kept in *ALT theorems (to remind of extracting corresponding lemmata, as discussed), I accept this  PR, because the comments are enhanced accordingly now.

For the future, I am a little bit afraid that automatisms will cause a lot of changes which nobody can control anymore. Therefore, changes should be perfomed step by step as done in this PR.

GinoGiotto:
> I fear dumping a file with all dependencies would result in something huge. Could we e.g. limit the output to e.g. just the list of theorems (directly or indirectly) depending on a given theorem? What would be the most useful format for you?

@tirix Yes, even listing all theorems that directly or indirectly depend on a single given theorem would be good enough. Maybe if it's possible it could be added the equivalent of the comma feature in metamath-exe ( like <code> SHOW TRACE_BACK * /MATCH nfmo,nfmod,nfmod2</code>), this way I could know the references of some given theorems without the "risk of explosion", because we wouldn't allow wildcards to be matched.

tirix:
> > I fear dumping a file with all dependencies would result in something huge. Could we e.g. limit the output to e.g. just the list of theorems (directly or indirectly) depending on a given theorem? What would be the most useful format for you?
> 
> @tirix Yes, even listing all theorems that directly or indirectly depend on a single given theorem would be good enough. Maybe if it's possible it could be added the equivalent of the comma feature in metamath-exe ( like ` SHOW TRACE_BACK * /MATCH nfmo,nfmod,nfmod2`), this way I could know the references of some given theorems without the "risk of explosion", because we wouldn't allow wildcards to be matched.

I've added the option in [PR #123](https://github.com/metamath/metamath-knife/pull/123). 
If you want to try it out before it is merged/released, you can check out the pull request's branch, and launch the new option using e.g.:
```
cargo run -- ../set.mm/set.mm --stmt-use nfmo,nfmod,nfmod2 
```
The result is simply printed into the standard output, in the exact same format as axiom-use, here is an extract:
```
reuf1od: nfmo
euoreqb:
2reu3: nfmod2 nfmo
2reu7: nfmod2
2reu8: nfmod2 nfmo
2reu8i:
2reuimp0:
```
And of course it's very fast.

PR #3403, opened 2023-08-19
Replaced dependencies from nfmo to nfmodv

GinoGiotto:
As asked in https://github.com/metamath/set.mm/pull/3359#issuecomment-1673797171 I made a small scan to find replacements of ~nfmo using ~nfmodv. The advantage is that ~nfmodv doesn't use `ax-13`, so as a result this PR drops `ax-13` for 10 theorems cc0fe66.


jkingdon:
Is this a case where we want to keep the old proofs as ALT theorems? The replacement proofs are a lot longer.

benjub:
> Is this a case where we want to keep the old proofs as ALT theorems? The replacement proofs are a lot longer.

Indeed. I'm not sure how to handle that.

* mosub: it may be better if possible to keep that proof except for a new `moexexvw` instead of moexexv.
* 2reuswap : maybe same as above with a new `2euswapv`.
* 2rmoswap : idem ?

GinoGiotto:
> mosub: it may be better if possible to keep that proof except for a new `moexexvw` instead of moexexv.

I played with the proof assistant a bit and I wasn't able to find a theorem that matched exactly `moexexv`, but I was able to prove this:

<pre>
  ${
    $d x y $.  $d ph y $.
    moexexvw $p |- ( E* x ph -> ( A. x E* y ps -> E* y E. x ( ph /\ ps ) ) ) $=
      ( wmo wex wal wa wi nfmo1 nfa1 wnf wtru nftru nfe1 a1i nfmodv nfim mopick
      mptru ex com23 alrimdv moim spsd syl6 exlimd wn exsimpl nexmo nsyl4 con1i
      exlimiv a1d pm2.61d1 ) ACEZACFZBDEZCGZABHZCFZDEZIZUPAVCCACJUSVBCURCKVBCLM
      VACDDNVACLMUTCOPQTRUPAVABIZDGZVCUPAVDDUPVAABUPVAABIABCSUAUBUCVEURVBCVABDU
      DUEUFUGUQUHVBUSVBUQVADFUQVBVAUQDABCUIUMVADUJUKULUNUO $.
  $}</pre>

With this theorem the new proof of `mosub` would become:

  <pre>
  ${
    $d x y A $.
    mosub.1 $e |- E* x ph $.
    $( "At most one" remains true after substitution.  (Contributed by NM,
       9-Mar-1995.) $)
    mosub $p |- E* x E. y ( y = A /\ ph ) $=
      ( cv wceq wmo wal wa wex moeq ax-gen moexexvw mp2 ) CFDGZCHABHZCIPAJCKBHC
      DLQCEMPACBNO $.
  $}</pre>

Usually I'm not very practical with creating theorems by hand, is this good to shorten the proof?

benjub:
@GinoGiotto : yes, that's exactly what I meant, with the tiny difference that we prefer in set.mm the "uncurried form" `( ph /\ ps ) -> ch` over the curried form `ph -> ( ps -> ch )`. You go from one to the other using ~imp and ~ex. Can you do it ?
What you can do is: rename your theorem moexexvwTEMP, then write the theorem
```
  ${
    $d x y $.  $d ph y $.
    moexexvw $p |- ( ( E* x ph -> A. x E* y ps ) -> E* y E. x ( ph /\ ps ) ) $=
      ? $.
  $}
```
Then in metamath.exe, prove it: something like:
```
MM> PROVE moexexvw
MM-PA> as l imp
MM-PA> as l moexexvwTEMP
MM-PA> im al
MM-PA> expand moexexvwTEMP
MM-PA> save new /C
MM-PA> exit
MM> prove mosub
MM-PA> min moexexvw /MAY
MM-PA> min moexexvw /EXCEPT moexexvwTEMP
...
```
(or if the latter does not work, DELETE STEP [step of moexexvw], etc.). Finally, remove moexexvwTEMP.

icecream17:
(benjub probably meant / typo) `    moexexvw $p |- ( ( E* x ph /\ A. x E* y ps ) -> E* y E. x ( ph /\ ps ) ) $=`\*

(incidentally I have never heard this convention and it would probably need to be added to ~ conventions)

david-a-wheeler:
jkingdon said:

> Is this a case where we want to keep the old proofs as ALT theorems? The replacement proofs are a lot longer.

If the old proofs are significantly smaller then yes, we often keep the shorter ones as ALT proofs.

GinoGiotto:
I added the lemmas to shorten the proofs, so ALT versions are not needed anymore.

jkingdon (approved):

jkingdon:
> I added the lemmas to shorten the proofs, so ALT versions are not needed anymore.

Looks good to me.

Hope I didn't jump the gun by merging but I wanted to reduce merge conflicts given all the things in flight and we can always make a followup if there are more things to change.

jkingdon:
> (incidentally I have never heard this convention and it would probably need to be added to ~ conventions)

I do tend to prefer the `( ph /\ ps ) -> ch` form but we have plenty of both so at least speaking for myself I'd be reluctant to make this a requirement.

benjub:
See https://groups.google.com/g/metamath/c/g7fFUjNalRE/

PR #3404, opened 2023-08-19
Revise comments of df-clab, eleq1ab, and cleljustab

benjub:
I tried to restructure and clarify the comment of df-clab, to clearly explain some subtleties (in the limited space that a comment should fit into), while keeping in some form or another everything from the previous version.

The change is very localized, so there is no risk of merge conflict, or if any, very easily resolved, so there is no hurry to merge.


set.mm
```diff
@@ -24785,52 +24785,79 @@ range over class builders (implicitly in the case of defined class terms
...
-     For a general discussion of the theory of classes, see
-     ~ mmset.html#class .  (Contributed by NM, 26-May-1993.) $)
+  $( Define class abstractions, that is, classes of the form ` { y | ph } ` ,
+     which is read "the class of sets ` y ` such that ` ph ( y ) ` ".
+
+     A few remarks are in order:
+
+     1.  The axiomatic statement ~ df-clab does not define the class
+     abstraction ` { y | ph } ` itself, that is, it does not have the form
+     ` |- ( { y | ph } = ... ) ` that a standard definition should have (for a
```
mazsa, line 24795: ` |- { y | ph } = ... `

set.mm
```diff
@@ -24785,72 +24785,107 @@ range over class builders (implicitly in the case of defined class terms
...
+     systems that can prove ~ ax-ext (and the necessary first-order logic).
+
+     2.  As in all definitions, the LHS (definiendum) has no disjoint variable
+     conditions.  In particular, the setvar variables ` x ` and ` y ` need not
+     be distinct, and the formula ` ph ` may depend on both ` x ` and ` y ` .
+     This is necessary, as with all definitions, since if there was for
+     instance a disjoint variable condition on ` x , y ` , then one could not
+     do anything with expressions like ` x e. { x | ph } ` which are sometimes
+     useful to shorten proofs (because of ~ abid ).  Most often, however, ` x `
+     is fresh in ` x e. { y | ph } ` and ` y ` is free in ` ph ` .
```
jkingdon, line 24813: I see the wikipedia definition of "fresh" but I don't think it is clear what it means here. Perhaps ``Most often, however, ` x ` does not occur in ` { y | ph } ` and ` y ` is free in ` ph ` ``

jkingdon:
I had one small comment but as for the larger changes, I think this is an improvement. Seems to be clear about things like axiom use and what we mean by "conservative".

set.mm
```diff
@@ -24785,72 +24785,107 @@ range over class builders (implicitly in the case of defined class terms
...
+
+     3.  Remark 1 stresses that ~ df-clab does not have the standard form of a
+     definition for a class, but one could be lead to think it has the standard
```
david-a-wheeler, line 24816: "could be lead" should be "could be led"

set.mm
benjub: done

set.mm
benjub: done

mazsa (approved):

jkingdon (approved):

mazsa:
@benjub I think this is too important to just be here. Don't you think we should refresh https://us.metamath.org/mpeuni/mmset.html#class and 
https://us.metamath.org/mpeuni/mmset.html#definitions as well?

benjub:
> @benjub I think this is too important to just be here. Don't you think we should refresh https://us.metamath.org/mpeuni/mmset.html#class and https://us.metamath.org/mpeuni/mmset.html#definitions as well?

Yes, we should modify these pages ultimately, but I think, only once the whole thing is clarified (for the moment, we have only addressed df-clab, which is the simplest of the three), and once we have written a real proof of conservativity from Mario's proof sketch. These are under way.

PR #3405, opened 2023-08-19
prove abbi1

benjub:
The forward implication of abbi, from fewer axioms.

I was looking at where it could be used in place of abbi, and I realized this is actually very close to abbilem, but maybe a bit better. I propose to replace abbilem with abbi1, with common attribution to @wlammen @icecream17 and me ?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13421,10 +13421,6 @@ New usage of "9cnOLD" is discouraged (0 uses).
 New usage of "a1ii" is discouraged (0 uses).
 New usage of "abbi2dvOLD" is discouraged (0 uses).
 New usage of "abbi2iOLD" is discouraged (0 uses).
-New usage of "abbidOLD" is discouraged (0 uses).
-New usage of "abbidvOLD" is discouraged (0 uses).
-New usage of "abbiiOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


icecream17:
shorter proof (basically just spsbbi + abbilem)
```mm
${
$d ph y $.
$d ps y $.
$d x y $.
abbi1 $p |- ( A. x ( ph <-> ps ) -> { x | ph } = { x | ps } ) $=
( wb wal vy cab wsb cv wcel spsbbi df-clab 3bitr4g eqrdv )
ABDCEZFACGZBCGZOACFHBCFHFIZPJRQJABCFKAFCLBFCLMN $.
$}
```

image

as for replacing abbilem, that is ok

benjub:
I also removed some *OLD proofs before the one-year period, since I checked that they really are subsumed by the new proof (i.e., no insight would be gained by studying them). Please double-check it.

wlammen (approved):
The most difficult to review change here is the omission of the OLD versions.  Our conventions page says
"..we often stage that change by first renaming its label with an OLD suffix ..."  The key word here is **often**.  It suggests that exemptions are possible (if sufficiently justified, I'd like to add).  So its ok with me.

icecream17 (approved):

PR #3406, opened 2023-08-20
Start the process of moving set.mm from +c to |_|

jkingdon:
Right now set.mm has two disjoint union definitions which are almost identical, https://us.metamath.org/mpeuni/df-dju.html and https://us.metamath.org/mpeuni/df-cda.html

The former is better because:

1. better name and notation (given that this isn't just defined on cardinals it is odd to call it `+c`)
2. works for proper classes (not sure we use this much, but why not?)
3. the order is switched which means that the disjoint union is a function from 2o in case we want to use it that way. That is, analogues to https://us.metamath.org/mpeuni/xpsc.html and its usages (when we get to that point) will not need the converse operation.

It is quite possible that set.mm will never make as much use of disjoint unions as iset.mm (in the latter disjoint union is well behaved in a way that regular union and set difference are not). The scope of this effort, as I see it, is to replace what is already in set.mm under the `+c` notation with `|_|`.

As for the current pull request, in addition to the obvious things like updating comments it starts the process of going through the `+c` theorems and making sure each one has an analogue for `|_|` (adding it to set.mm if needed).

Oh and as for some of the larger context, #3223 at least following the development in set.mm, goes through https://us.metamath.org/mpeuni/df-xps.html which uses the https://us.metamath.org/mpeuni/xpsc.html style pair function mentioned above. It would be odd to intuitionize `+c` for this purpose when there is no reason to have both `+c` and `|_|`.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4646,7 +4646,7 @@ this implies excluded middle</TD>
 </tr>
 
 <tr>
-  <td>uncdadom</td>
+  <td>uncdadom , undjudom</td>
   <td><i>none</i></td>
   <td>The set.mm proof relies on undom</td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4624,6 +4624,28 @@ this implies excluded middle</TD>
 </TR>
 
 <tr>
+  <td>df-cda</td>
+  <td>~ df-dju</td>
+  <td>df-cda and df-dju are essentially the same. Current
+  thinking is that we want to remove df-cda and the +c syntax
+  from set.mm in favor of df-dju and use ` |_| ` in much the same
+  way that cardinal multiplication uses the df-xp ( ` X. ` )
+  syntax rather than having a notation specific to cardinals.</td>
+</tr>
+
+<tr>
+  <td>cdafn</td>
+  <td>~ djuex</td>
+  <td>Since ` |_| ` is defined on proper classes, the similarity
+  here is somewhat approximate.</td>
+</tr>
+
+<tr>
+  <td>cdaval</td>
+  <td>~ df-dju</td>
+</tr>
+
+<tr>
   <td>uncdadom</td>
   <td><i>none</i></td>
   <td>The set.mm proof relies on undom</td>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9401,7 +9401,8 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>df-xps and all theorems mentioning the binary product on a structure
   (syntax Xs.)</td>
   <td><i>none</i></td>
-  <td>presumably could be added in some form</td>
+  <td>The set.mm definition depends on df-cda ( +c ), df-imas ( "s ),
+  and df-prds ( Xs_ )</td>
 </tr>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4646,6 +4646,12 @@ this implies excluded middle</TD>
 </tr>
 
 <tr>
+  <td>cdadju</td>
+  <td><i>none</i></td>
+  <td>not needed as iset.mm does not have df-cda</td>
+</tr>
+
+<tr>
   <td>uncdadom , undjudom</td>
   <td><i>none</i></td>
   <td>The set.mm proof relies on undom</td>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4662,6 +4662,11 @@ this implies excluded middle</TD>
 </tr>
 
 <tr>
+  <td>cdaenun</td>
+  <td>~ djuenun</td>
+</tr>
+
+<tr>
   <td>pm110.643</td>
   <td>~ dju1p1e2</td>
 </tr>


set.mm
```diff
@@ -98571,6 +98582,15 @@ our Axiom of Choice (in the form of ~ ac2 ).  The proof does not depend
     UOPURANOZUSBNOZPURUSFGHZUPPVBUPUNVCUOVDUNGQEVCUEAGCQRSUOLTEVDUFBLDTRSUGUPVE
     ABUHUIURAUSBUJUKULUM $.
 
+  $( Equinumerosity of a disjoint union and a regular union of two disjoint
```
benjub, line 98585:
```suggestion
  $( Equinumerosity of a disjoint union and a union of two disjoint
```

iset.mm
```diff
@@ -69051,6 +69031,35 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 $)
 
+  $( Equinumerosity of a disjoint union and a regular union of two disjoint
```
benjub, line 69034:
```suggestion
  $( Equinumerosity of a disjoint union and a union of two disjoint
```

benjub:
I think that's a good move. One side remark below.

> the order is switched which means that the disjoint union is a function from 2o in case we want to use it that way. That is, analogues to https://us.metamath.org/mpeuni/xpsc.html and its usages (when we get to that point) will not need the converse operation.

I see this more as a weakness since this permits to prove "junk theorems" (unavoidable in set theories). Hopefully, that weakness won't be abused too much.

avekens:
> It is quite possible that set.mm will never make as much use of disjoint unions as iset.mm (in the latter disjoint union is well behaved in a way that regular union and set difference are not). The scope of this effort, as I see it, is to replace what is already in set.mm under the `+c` notation with `|_|`.
> 
> As for the current pull request, in addition to the obvious things like updating comments it starts the process of going through the `+c` theorems and making sure each one has an analogue for `|_|` (adding it to set.mm if needed).

It's not clear to me if you want to replace `+c` by `|_|` in set.mm (that means `+c`  will be removed at the end), or if you want to duplicate theorems using `+c` (that means adding corresponding theorems using `|_|`, and  `+c`  will be kept at the end).

Since there are about 100 theorems using `+c` (and only 26 using `|_|`), both approaches seem to cause a lot of effort, and I do not see the adbvantages, at least not in set.mm (both definitions have their own, different intentions). 

Finally, I miss a theorem showing the relationship between these defintions (especially to show that they are similar or could be replaced by each other), e.g.  `( ( A e. V /\ B e. W ) -> ( A +c B ) ~~ ( A |_| B ) )`.

benjub:
I think the end goal is to remove `+c` since we want to be frugal on definitions, so we want to avoid two very similar definitions.

jkingdon:
> I think that's a good move. One side remark below.
> 
> > the order is switched which means that the disjoint union is a function from 2o in case we want to use it that way. That is, analogues to https://us.metamath.org/mpeuni/xpsc.html and its usages (when we get to that point) will not need the converse operation.
> 
> I see this more as a weakness since this permits to prove "junk theorems" (unavoidable in set theories). Hopefully, that weakness won't be abused too much.

I suspect one person's junk theorem is another's concise representation.

I haven't really looked enough at the existing usages of the `` `' ( { A } +c { B } ) `` idiom to have much of an opinion about whether I think it is good, bad or indifferent, or what the alternative would be. But for this project I had in mind a direct one to one replacement of `` `' ( { A } +c { B } ) `` with ` ( { A } |_| { B } ) ` .  For what it is worth, we chose the definition of df-dju specifically to allow this usage: https://github.com/metamath/set.mm/issues/2625#issuecomment-1158632646

set.mm
jkingdon: OK, I guess plain `union` is better. For one thing, "regular" here wasn't supposed to be a reference to any form of regularity, it is just a non-technical word.

Perhaps the language here would be less awkward if there was a name for disjoint union which didn't have "union" in the name, but given how well-established disjoint union is as a term, that seems unlikely.

Will accept this suggestion.

jkingdon: Hmm, I see I wasn't the only one to use "regular union" in this context: https://us.metamath.org/mpeuni/konigth.html does too. I guess I'll change that one as well.

iset.mm
jkingdon: done

jkingdon:
> > As for the current pull request, in addition to the obvious things like updating comments it starts the process of going through the `+c` theorems and making sure each one has an analogue for `|_|` (adding it to set.mm if needed).
> 
> It's not clear to me if you want to replace `+c` by `|_|` in set.mm (that means `+c` will be removed at the end), or if you want to duplicate theorems using `+c` (that means adding corresponding theorems using `|_|`, and `+c` will be kept at the end).
> 
> Since there are about 100 theorems using `+c` (and only 26 using `|_|`), both approaches seem to cause a lot of effort, and I do not see the adbvantages, at least not in set.mm (both definitions have their own, different intentions).

The goal is to remove `+c` at the end and the duplicate theorems are just an intermediate step. As for whether there are different intentions, I'd refer you to the comment of https://us.metamath.org/mpeuni/cdaval.html (as it is now, not as this pull request would change it) where it describes `+c` as a disjoint union and contrasts with a different approach (not chosen by set.mm) in which there would be a cardinal addition operator just defined on cardinals.

> 
> Finally, I miss a theorem showing the relationship between these defintions (especially to show that they are similar or could be replaced by each other), e.g. `( ( A e. V /\ B e. W ) -> ( A +c B ) ~~ ( A |_| B ) )`.

To be honest I hadn't quite gotten as far as figuring out *how* to make this transition happen, but probably the theorem you suggest is a good idea, so I'll add that to this pull request, probably later today.

set.mm
benjub: Yes, "disjoint union" is an example of a "red herring". (At first reading, I genuinely wondered what you meant by "regular union"... for at least five seconds!). And indeed, "regular" is one of the most overloaded terms in mathematics: for basically any noun, there is a slightly different "regular" notion.

If you're looking for another term, you can use "coproduct" !

benjub:
> I suspect one person's junk theorem is another's concise representation.

... and everyone's road to chaos !

> 
> I haven't really looked enough at the existing usages of the `` `' ( { A } +c { B } ) `` idiom to have much of an opinion about whether I think it is good, bad or indifferent, or what the alternative would be. But for this project I had in mind a direct one to one replacement of `` `' ( { A } +c { B } ) `` with `( { A } |_| { B } )` . For what it is worth, we chose the definition of df-dju specifically to allow this usage: [#2625 (comment)](https://github.com/metamath/set.mm/issues/2625#issuecomment-1158632646)

I agree, and there are times where the shortcut is so tempting it wouldn't be worth to avoid it.  Indeed, the theorems you mention already use that shortcut. My remark was more about not abusing this kind of shortcuts in general (else we'll end up proving that the value of the sine function at $\mathbb{R}$ is equal to 3 times the cosine function -- since both are the empty set under the current conventions; here it's obviously nonsensical, but there are more subtle cases).

jkingdon:
> > Finally, I miss a theorem showing the relationship between these defintions (especially to show that they are similar or could be replaced by each other), e.g. `( ( A e. V /\ B e. W ) -> ( A +c B ) ~~ ( A |_| B ) )`.
> 
> To be honest I hadn't quite gotten as far as figuring out _how_ to make this transition happen, but probably the theorem you suggest is a good idea, so I'll add that to this pull request, probably later today.

Added.

jkingdon:
> My remark was more about not abusing this kind of shortcuts in general (else we'll end up proving that the value of the sine function at ℝ is equal to 3 times the cosine function -- since both are the empty set under the current conventions; here it's obviously nonsensical, but there are more subtle cases).

LOL, cute example and I agree with the basic idea that we should be wary of being too clever. (And also, I think, about "there are times where the shortcut is so tempting it wouldn't be worth to avoid it").

set.mm
avekens: Yes, maybe "coproduct" would be a better name, see also Wikipedia https://en.wikipedia.org/wiki/Coproduct

benjub: Well, the proposal was a bit of a joke, and tradition will make us stick to "disjoint union".  What we can do is add to the comment of ~df-dju :
```
The disjoint union, together with the left and right injections ~ df-inl and ~ df-inr , is
"the" coproduct in the category of sets and functions.
```
The main benefit of adding this sentence is that if someone searches for "coproduct" in the website, they will get this positive hit (and also, it links to the two definitions of the injections).

benjub: (and we can even link to @avekens's proof that it is indeed the coproduct ~updjud !)

jkingdon: I suppose but this sentence may baffle as many people as it helps. It refers to a lot of concepts which aren't treated in set.mm at all, or if they are in a very different section.

Even for people searching for "coproduct" I'm not sure they'd particularly be well served by landing on df-dju (since that's just one kind of coproduct).

tirix:
Generally we hide those tricks, by only exposing the properties, and marking the constructions with `(Avoid depending on this detail.)`, c.f. the construction of the complex numbers for example.

jkingdon:
> Generally we hide those tricks, by only exposing the properties, and marking the constructions with `(Avoid depending on this detail.)`, c.f. the construction of the complex numbers for example.

Sometimes we say that, sometimes we don't. Anyway, I think there seems to be agreement that the level of trickery here is planned to be the same as the amount of trickery with `+c`, so I don't think this affects this pull requests or the future ones I have in mind related to switching to `|_|`.

set.mm
avekens: We have references to category theory in earlier places, see, for example, ~df-f1 :

>  Injections are precisely the monomorphisms in the category ` SetCat ` of sets and set functions, see ~ setcmon .

digama0: it would be even better if we prove it's a coproduct, although I think the category theory development doesn't get far enough to define coproducts

digama0: I mean to package that theorem into a definition which uses the word "coproduct" in the name. Otherwise it's difficult to look at that big logical formula and see "coproduct" from that.

jkingdon:
Is this ready to merge? There are a variety of comments which went in various directions, but I believe either (a) the pull request has been updated accordingly, (b) there is a response here explaining why it is the way it is, or (c) there is not enough agreement on a topic to fold it into this pull request itself.

avekens:
> Is this ready to merge? There are a variety of comments which went in various directions, but I believe either (a) the pull request has been updated accordingly, (b) there is a response here explaining why it is the way it is, or (c) there is not enough agreement on a topic to fold it into this pull request itself.

It's not clear to me if the comment about category theory (proposed by @benjub) should be added or not (maybe with a "Remark:" put in front):

> The disjoint union, together with the left and right injections ~ df-inl and ~ df-inr , is
"the" coproduct in the category of sets and functions.

In my opinion, it should be added (even if the corrsponding definition is not available in set.mm yet), because it gives a valuable hint for everybody interested in category theory.

jkingdon:
Sounds like this isn't ready for main.

I'll get a mathbox pull request ready, probably in another day or two.

benjub:
As for the remark concerning "coproduct", it's as you want. I think this MR could be merged, and if needed one can do other changes later.

jkingdon:
> As for the remark concerning "coproduct", it's as you want. I think this MR could be merged, and if needed one can do other changes later.

I'll take that as an approval. Merging.

I'm also willing to put this work in a mathbox if people would prefer it. Of course that creates work down the line if we decide to move it back to main, but for the moment it is a bunch of new theorems more so than modifying existing ones.

PR #3407, opened 2023-08-20
Fixes to metamath-knife job in the automated checks

jkingdon:
This is two changes:

Also run `--verify-markup` which seems like a good idea and also caught some formatting problems in iset.mm

Fixes the caching. As far as I could tell reading the output in github this worked, the `Post Cache metamath-knife` output reads:

```
Post job cleanup.
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/set.mm/set.mm --files-from manifest.txt --use-compress-program zstdmt
Cache Size: ~8 MB (8264376 B)
Cache saved successfully
Cache saved with key: Linux-build-cache-knife
```

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -193,15 +193,20 @@ jobs:
         env:
           cache-name: cache-knife
         with:
-          path: ~/.cargo/bin/cache-knife
+          path: metamath-knife/target/release/metamath-knife
           key: ${{ runner.os }}-build-${{ env.cache-name }}
       - name: Install rust
         if: ${{ !steps.cache-metamath-knife.outputs.cache-hit }}
         uses: hecrj/setup-rust-action@v1
         with:
           rust-version: stable
-      - run: git clone --depth 1 https://github.com/david-a-wheeler/metamath-knife.git
-      - run: cd metamath-knife && cargo build --release
+      - name: Install metamath-knife
+        if: ${{ !steps.cache-metamath-knife.outputs.cache-hit }}
+        run: |
+          rm -rf metamath-knife &&
+          git clone --depth 1 https://github.com/metamath/metamath-knife.git &&
+          cd metamath-knife &&
+          cargo build --release
       - name: Verify set.mm
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/verifiers.yml
```diff
@@ -193,21 +193,25 @@ jobs:
...
         if: ${{ !steps.cache-metamath-knife.outputs.cache-hit }}
         uses: hecrj/setup-rust-action@v1
         with:
           rust-version: stable
-      - run: git clone --depth 1 https://github.com/david-a-wheeler/metamath-knife.git
-      - run: cd metamath-knife && cargo build --release
+      - name: Install metamath-knife
+        if: ${{ !steps.cache-knife.outputs.cache-hit }}
+        run: |
+          git clone --depth 1 https://github.com/david-a-wheeler/metamath-knife.git &&
```
tirix, line 206: The repository has moved recently, now it is:
```
https://github.com/metamath/metamath-knife.git
```

.github/workflows/verifiers.yml
```diff
@@ -193,21 +193,25 @@ jobs:
...
+          cd metamath-knife &&
+          cargo build --release
       - name: Verify set.mm
         run: |
-          metamath-knife/target/release/metamath-knife --verify ./set.mm
+          metamath-knife/target/release/metamath-knife --verify --verify-markup ./set.mm
```
tirix, line 211: Since you're at it, maybe you could add `--parse-typesetting` too?

.github/workflows/verifiers.yml
jkingdon: Ah, nice. I was hoping we'd get around to that move.

.github/workflows/verifiers.yml
jkingdon: Added.

.github/workflows/verifiers.yml
jkingdon: Changed.

tirix (approved):
Thank you Jim!

PR #3409, opened 2023-08-21
Centers and the Prime Field, degree of C/R

tirix:
This moves the section about sub division rings from SO's mathbox to main.
Most of the new theorems were suggested by Gérard Lang.

The main results are:
- the smallest sub division ring of a division ring is a field, ~ primefld
- the degree of the field extension [ℂ:ℝ] = 2

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,21 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+22-Aug-23 df-srdg   [same]      moved from SO's mathbox to main set.mm
+22-Aug-23 issrdg    [same]      moved from SO's mathbox to main set.mm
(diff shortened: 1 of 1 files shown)


jkingdon (approved):

avekens:
The theorems were in @sorear `s mathbox, but some of them were contributed by @digama0 . Since three people are/were woking on this topic, I would propose to move section "Subfields" to main then.

david-a-wheeler:
[avekens](https://github.com/avekens) makes a fair point - when this many people are collaborating on this material, that's a good indication that it should go into the main body.

tirix:
Indeed, and sub-division rings are important enough to deserve a section in main.
I've moved that section in my last commit, with theorems from all 3 contributors.

jkingdon (approved):
I'm going to go ahead and merge this since there seems to be a fair bit of agreement on this material including putting division rings in main.

PR #3411, opened 2023-08-21
factor out common steps in reean(v)

wlammen:
1. Clean up a single spot (reeanv) that was affected by #3399 and described there in a comment.
2. Delete an outdated OLD theorem.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17902,7 +17902,6 @@ New usage of "swrdtrcfvOLD" is discouraged (2 uses).
 New usage of "swrdtrcfvlOLD" is discouraged (1 uses).
 New usage of "syl5imp" is discouraged (0 uses).
 New usage of "syl5impVD" is discouraged (0 uses).
-New usage of "symdif2OLD" is discouraged (0 uses).
 New usage of "symdifassOLD" is discouraged (0 uses).
 New usage of "tarski-bernays-ax2" is discouraged (0 uses).
 New usage of "tb-ax1" is discouraged (3 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -29664,6 +29664,19 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
     ( cv wcel wi wral pm2.04 ralimi2 impbii ) BEZDFZAGZBCHLCFZAGZBDHNPBCDOMAIJP
     NBDCMOAIJK $.
 
+  ${
+    $d y A $.  $d x B $.  $d x y $.
+    reeanlem.1 $e |- ( E. x E. y ( ( x e. A /\ ph ) /\ ( y e. B /\ ps ) )
+                    <-> ( E. x ( x e. A /\ ph ) /\ E. y ( y e. B /\ ps ) ) ) $.
+    $( Lemma factoring out common proof steps of ~ reeanv and ~ reean .
+       (Contributed by Wolf Lammen, 20-Aug-2023.) $)
```
tirix, line 29672: Maybe keep Norm's attribution and add your "proof modified by"?

tirix (approved):

set.mm
wlammen: I thought about this, because I am certainly not the 'inventor' of this code, but just extracted portions of an existing proof.  The original proof idea was designed by NM and AS, latest change dating back in 2011, but I don't know who did what, and cannot use git to find out.  So attributing NM alone could be unfair to AS.

This is a general problem in case you refactor a given proof base.  Adding all possible contributors could turn out to be a nightmare, if the base is was designed by, say 10 or more individuals.  Doing the detective work to identify each contribution is not in support of refactoring at all, and might severely hamper development.  So what shall we do?  I am curious to here opinions on this topic.

tirix: That's fine with me! It's not always easy to manage attributions.

avekens (approved):

PR #3414, opened 2023-08-22
Drop ax-10 and ax-11 from rmoanim

GinoGiotto:


avekens:
I would like to keep the old, much shorter proof in a rmoanimALT theorem. Then I would accept this PR.

icecream17 (approved):

avekens (approved):

PR #3415, opened 2023-08-22
Reduced axiom usage

GinoGiotto:
* Drop ax-10 and ax-13 from `sbco4lem`
* Drop ax-10 and ax-13 from `sbco4`
* Drop ax-13 from `sbbib`
* Drop ax-10 and ax-11 from `csbgfi`

Full axiom dependencies here: 1252b00


icecream17 (approved):
nfal alrimi --> hbal alrimih seems to be pretty common

I think https://github.com/metamath/set.mm/issues/2710 is potentially more useful/general, as it provides an easy way to find such sequences. Another example is replacing "bicomi exbii bitri" with "exbii bitr4i"

GinoGiotto:
> I think #2710 is potentially more useful/general, as it provides an easy way to find such sequences.

That list seems very interesting indeed, is it available a script to regenerate an updated one?

avekens (approved):
proofs become shorter or not much longer.

benjub:
Shouldn't some theorems be moved earlier to follow the rule that a theorem should be in the earliest possible $a-interval (except possibly for specific lemmas) ?

benjub (approved):

PR #3418, opened 2023-08-23
Moved theorems in ax-12 interval

GinoGiotto:
Some theorems that are currently in the ax-13 section use axioms up to ax-12, so I moved them in the ax-12 interval.


avekens:
To move these theorems into section ax-12 is a very good idea, but it should be discussed if they are placed at the right positions within this section.

GinoGiotto:
I was wondering that. My simplistic approach has been to place them in the earliest possible position to maximize the amount of following theorems they could be used for. But I noticed set.mm has some local naming patterns, so maybe there are some other criteria to consider.

icecream17 (commented):
not sure about drsb2 but all the other theorems seem to be in the right place, with sbf2 right after sbf and sbim right after sbi2

wlammen:
I think drsb2 is misplaced.  It belongs to the part starting with axc16g. I'd put drsb2 right after axc11rv.

icecream17 (approved):

avekens (approved):

PR #3425, opened 2023-08-24
+ fltnlta

icecream17:
expmord, rpexpmord is placed around ltexp2 (note how ltexp1d uses rpexpmord) however expmordi depends on reexpcld and expp1d which moves things around

fltnlta does not seem to make any progress towards flt, it is merely an intersting theorem

this proof requires ( C - B ) to be at least 1; my aforementioned "false proof of flt" was when I previously thought fltnlta could be generalized to the rationals

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -55,6 +55,11 @@ Date      Old       New         Notes
 22-Aug-23 cntrss    [same]      moved from TA's mathbox to main set.mm
 22-Aug-23 iinssiun  [same]      moved from TA's mathbox to main set.mm
 22-Aug-23 iindif1   [same]      moved from TA's mathbox to main set.mm
+22-Aug-23 mvlladdd  [same]      moved from DAW's mathbox to main set.mm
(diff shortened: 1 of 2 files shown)


benjub:
As written earlier, the initial comment and the title of a PR should describe the changes made to the files.

avekens (approved):
OK, now I understand (after analysing the dependencies of the affected theorems): 
* you created some new theorems in your mathbox
* for the proofs of these theorems some theorems of other's mathboxes are required
* you moved the theorems of other's mathboxes to main, placed them in appropriate places
* the moved theorems require theorems which are placed after them now, so you had to move up some (actually many) theorems.

The result seems to be OK then.

jkingdon (approved):

jkingdon:
> fltnlta does not seem to make any progress towards flt, it is merely an intersting theorem

You should just prove Fermat's Last Theorem. Then `fltnlta` will be an easy consequence.

(Uh, that's a joke.  I mean, for now.)

PR #3427, opened 2023-08-24
More +c to |_| from cda1en to cdaassen

jkingdon:
This is much the same as #3406 for the next bunch of theorems.

As far as I noticed (I might have missed some), the `+c` theorems here are still used, so I need to continue this process until I get to their usages before I can actually remove much of anything.

One exception is `pm110.643` which does not have any usages (ironically, in light of the comment "The above proposition is occasionally useful") so I just kept the `|_|` version.  I merely changed the notation, the name of the theorem, and updated the proof in a very direct fashion (that is, the new proof is the old proof step-by-step with the only changes being the ones needed to change to `|_|`). I kept the existing comment with its historical notes and the like.  The same applies to `pm110.643ALT` which has a similar set of minimal changes.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -60,6 +60,8 @@ Date      Old       New         Notes
 21-Aug-23 elrnmpt2  elrnmpo
 21-Aug-23 elrnmpt2g elrnmpog
 21-Aug-23 reldmmpt2 reldmmpo
+21-Aug-23 pm110.643ALT dju1p1e2ALT Changes from +c notation to |_|
+21-Aug-23 pm110.643 dju1p1e2    Changes from +c notation to |_|
 20-Aug-23 rnmpt2    rnmpo
 20-Aug-23 mpt22eqb  mpo2eqb
 20-Aug-23 mpt2fun   mpofun
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+24-Aug-23 cdaassen  djuassen    Changes from +c notation to |_|
 23-Aug-23 ovmpt2    ovmpo
 23-Aug-23 ovmpt2g   ovmpog
 23-Aug-23 ovmpt2dv2 ovmpodv2
--- a/discouraged
+++ b/discouraged
@@ -15033,6 +15033,7 @@ New usage of "djajN" is discouraged (0 uses).
 New usage of "djavalN" is discouraged (2 uses).
 New usage of "djhljjN" is discouraged (0 uses).
 New usage of "djhunssN" is discouraged (0 uses).
+New usage of "dju1p1e2ALT" is discouraged (0 uses).
 New usage of "djuexALT" is discouraged (0 uses).
 New usage of "dlwwlknonclwlknonf1olem1OLD" is discouraged (1 uses).
 New usage of "dlwwlknondlwlknonenOLD" is discouraged (0 uses).
@@ -17244,7 +17245,6 @@ New usage of "pl42lem4N" is discouraged (1 uses).
 New usage of "plpv" is discouraged (1 uses).
 New usage of "pm10.252" is discouraged (0 uses).
 New usage of "pm11.07" is discouraged (0 uses).
-New usage of "pm110.643ALT" is discouraged (0 uses).
 New usage of "pm13.183OLD" is discouraged (0 uses).
 New usage of "pm13.18OLD" is discouraged (0 uses).
 New usage of "pm2.21ddALT" is discouraged (0 uses).
@@ -18710,6 +18710,8 @@ Proof modification of "disamisOLD" is discouraged (19 steps).
 Proof modification of "disjxwrdOLD" is discouraged (13 steps).
 Proof modification of "disjxwwlknOLD" is discouraged (130 steps).
 Proof modification of "disjxwwlksnOLD" is discouraged (80 steps).
+Proof modification of "dju1p1e2" is discouraged (72 steps).
+Proof modification of "dju1p1e2ALT" is discouraged (34 steps).
 Proof modification of "djuexALT" is discouraged (51 steps).
 Proof modification of "dlwwlknonclwlknonf1olem1OLD" is discouraged (253 steps).
 Proof modification of "dlwwlknondlwlknonenOLD" is discouraged (116 steps).
@@ -19501,8 +19503,6 @@ Proof modification of "orim12dALT" is discouraged (34 steps).
 Proof modification of "p0exALT" is discouraged (2 steps).
 Proof modification of "perfectALTV" is discouraged (528 steps).
 Proof modification of "perpdragALT" is discouraged (241 steps).
-Proof modification of "pm110.643" is discouraged (72 steps).
-Proof modification of "pm110.643ALT" is discouraged (35 steps).
 Proof modification of "pm13.183OLD" is discouraged (111 steps).
 Proof modification of "pm13.18OLD" is discouraged (28 steps).
 Proof modification of "pm2.21ddALT" is discouraged (10 steps).
--- a/discouraged
+++ b/discouraged
@@ -2976,6 +2976,57 @@
 "ccshOLD" is used by "0csh0OLD".
 "ccshOLD" is used by "cshfnOLD".
 "ccshOLD" is used by "cshnzOLD".
+"cda0en" is used by "cdalepw".
+"cda1dif" is used by "canthp1".
+"cda1en" is used by "pwsdompw".
+"cdacomen" is used by "alephadd".
+"cdacomen" is used by "cdadom2".
+"cdacomen" is used by "cdalepw".
+"cdacomen" is used by "gchdomtri".
+"cdacomen" is used by "gchhar".
+"cdacomen" is used by "gchpwdom".
+"cdacomen" is used by "infcda".
+"cdacomen" is used by "pwxpndom".
+"cdaen" is used by "ackbij1lem5".
+"cdaen" is used by "ackbij1lem9".
+"cdaen" is used by "cardacda".
+"cdaen" is used by "cdaenun".
+"cdaen" is used by "gchhar".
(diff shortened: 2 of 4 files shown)


avekens:
> As far as I noticed (I might have missed some), the `+c` theorems here are still used, so I need to continue this process until I get to their usages before I can actually remove much of anything.

Maybe we should make the corresponding `+c` theorems OLD theorems so that it becomes clear that they will be removed soon. In this case, however, we do not need to wait one year until they can be removed - they can be removed as soon as they are not used anymore.

jkingdon:
> Maybe we should make the corresponding `+c` theorems OLD theorems so that it becomes clear that they will be removed soon. In this case, however, we do not need to wait one year until they can be removed - they can be removed as soon as they are not used anymore.

I've been thinking of marking as discouraged. For example, `cda1en` would get a comment "Use ~ dju1en instead. (New usage is discouraged.)" I was maybe going to do that a bit later in the process but I suppose I could start now, so I guess I'll add that to this pull request unless it seems better to make it a new one.

The comment about OLD naming is interesting. I have been naming all the new theorems with `dju` rather than `cda` which does gives a sense of which notation a particular theorems is for, and does mean there are fewer naming conflicts which would mean that a theorem needs to become OLD to free up its name for something else. I at least toyed with the idea of using some kind of name which says "sure this is a  disjoint union but we are using it for cardinal addition here" but existing practice seems to go against that in theorems like https://us.metamath.org/mpeuni/xp1en.html or https://us.metamath.org/mpeuni/omxpen.html , where we use the `xp` naming whether the theorem is about Cartesian product used for cardinal multiplication, or Cartesian product used some other way.

As for whether to do the thing about waiting a year, I guess that was/is mostly about people who might have theorems in their own branches/forks/copies? And maybe about people who want to compare the old and new copies on the website without having to check out an old version in git and generate web pages? I'm probably not the best judge of whether it would be useful because I haven't really been looking at those OLD-with-a-time-limit theorems myself. I'll accept the notion that it isn't necessary here unless someone wants to ask for this convention to be followed.

avekens:
> > Maybe we should make the corresponding `+c` theorems OLD theorems so that it becomes clear that they will be removed soon. In this case, however, we do not need to wait one year until they can be removed - they can be removed as soon as they are not used anymore.
> 
> I've been thinking of marking as discouraged. For example, `cda1en` would get a comment "Use ~ dju1en instead. (New usage is discouraged.)" I was maybe going to do that a bit later in the process but I suppose I could start now, so I guess I'll add that to this pull request unless it seems better to make it a new one.
> 
> The comment about OLD naming is interesting. I have been naming all the new theorems with `dju` rather than `cda` which does gives a sense of which notation a particular theorems is for, and does mean there are fewer naming conflicts which would mean that a theorem needs to become OLD to free up its name for something else. I at least toyed with the idea of using some kind of name which says "sure this is a disjoint union but we are using it for cardinal addition here" but existing practice seems to go against that in theorems like https://us.metamath.org/mpeuni/xp1en.html or https://us.metamath.org/mpeuni/omxpen.html , where we use the `xp` naming whether the theorem is about Cartesian product used for cardinal multiplication, or Cartesian product used some other way.

Our conventions say "If a proof, definition, axiom, or theorem is going to be removed, we often stage that change by first renaming its label with an OLD suffix (to make it clear that it is going to be removed)". It is not necessary that there is still a theorem with the same label without the suffix OLD, so we could use it also here for marking the theorems to be removed.

> As for whether to do the thing about waiting a year, I guess that was/is mostly about people who might have theorems in their own branches/forks/copies? And maybe about people who want to compare the old and new copies on the website without having to check out an old version in git and generate web pages? I'm probably not the best judge of whether it would be useful because I haven't really been looking at those OLD-with-a-time-limit theorems myself. I'll accept the notion that it isn't necessary here unless someone wants to ask for this convention to be followed.

Our conventions say "This typically happens when a shorter proof to an existing theorem is found: the existing theorem is kept as an *OLD statement for one year." That means the 1-year-waiting-time is mainly supposed to be respected if proofs have changed. This is not the case here.

benjub:
I'm fine with either choice proposed above.

A side remark for *OLD statements: if the statement of `xxxOLD` is identical to the statement of `xxx` (including the precise label before "OLD") and all DV conditions of `xxx` are also in `xxxOLD`, then any proof using that theorem would still work verbatim, so there is less of a need to wait for one year, especially if two different contributors have checked that nothing and no insight would be gained from keeping/reviving the OLD proof. We have made a few exceptions like that recently for readability reasons, with @wlammen, @GinoGiotto and @icecream17.

jkingdon:
> I'm fine with either choice proposed above.

OK, I've pushed up the discouragement tags. As far as I can tell this should be sufficient to let people know these theorems are not preferred any more.

As for the wider discussions, we discussed a number of semi-related topics or options (myself included) but I guess what I just pushed up should be workable at least for now.

set.mm
```diff
@@ -98664,12 +98664,13 @@ our Axiom of Choice (in the form of ~ ac2 ).  The proof does not depend
...
        [Mendelson] p. 258.  See ~ cdaval for its value and a description.
-       (Contributed by NM, 24-Sep-2004.) $)
+       (Contributed by NM, 24-Sep-2004.)  Use ~ df-dju instead.
+       (New usage is discouraged.) $)
     df-cda $a |- +c = ( x e. _V , y e. _V |->
                    ( ( x X. { (/) } ) u. ( y X. { 1o } ) ) ) $.
 
     $( Cardinal number addition is a function.  (Contributed by Mario Carneiro,
-       28-Apr-2015.) $)
+       28-Apr-2015.)  Use ~ djuex instead.  (New usage is discouraged.) $)
```
avekens, line 98673: THis is misleading: |_| is not a function, and djuex is not corresponding to cdafn (there is no corresponding theorem). Proof of theorems using cdafn cannot be transformed straightforward, see ~cda1dif.

set.mm
```diff
@@ -98749,7 +98753,7 @@ our Axiom of Choice (in the form of ~ ac2 ).  The proof does not depend
 
   $( Cardinal addition of equinumerous sets.  Exercise 4.56(b) of [Mendelson]
      p. 258.  (Contributed by NM, 28-Sep-2004.)  (Revised by Mario Carneiro,
-     29-Apr-2015.) $)
+     29-Apr-2015.)  Use ~ djuen instead.  (New usage is discouraged.) $)
```
avekens, line 98756: If ~cdaen will be removed soon, the comment should be aligned with the comment of ~djuen. Especially the original contributor should be kept (at least if the proofs are still similar). Proposal for a comment of ~djuen:

`$( Disjoint unions of equinumerous sets are equinumerous.  This corresponds to the cardinal addition of equinumerous sets.  Exercise 4.56(b) of [Mendelson] p. 258.  (Contributed by NM, 28-Sep-2004.)  (Revised by Mario Carneiro, 29-Apr-2015.) (Revised by Jim Kingdon, 30-Jul-2023.) $)`

avekens: This also holds for the other theorems, of course.

avekens: See ~dju1dif  how it should be (its comment corresponds to my proposal).

changes-set.txt
```diff
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+24-Aug-23 cdaassen  djuassen    Changes from +c notation to |_|
```
avekens, line 29: What about the other theorems, e.g. cdacomen - djucomen etc.?

set.mm
jkingdon: The connection is approximate at best, so let's pick a better wording. How about:

```
    $( Cardinal number addition is a function.  New proofs are encouraged to
       use the ` |_| ` notation instead of ` +c ` but because the former is not
       a function, the replacement will vary.  (Contributed by Mario Carneiro,
       28-Apr-2015.)  (New usage is discouraged.) $)
```

changes-set.txt
jkingdon: I don't know, would you put them in the "PROPOSED" section?

The ones that are listed in the pull request now are the ones which are removed, not the ones marked as discouraged.

set.mm
benjub: 1. I like the pattern: `Use ~ xxx instead.  (New usage is discouraged.) $)` at the end of a comment. It is uniform so makes it easier to search, and it doesn't imply that one should simply replace a label with another one or that no further adapting is necessary.
2. I wouldn't take too much time enhancing comments of statements that are going to be deprecated.

set.mm
jkingdon: As for the comments in general, this runs right into the "it depends" nature of our conventions.  There is no convention to preserve existing `Revised` lines (in general). There is a convention to (generally speaking, not always) avoid more than one `Revised` line. If it were just up to me, in the vast majority of cases I'd list the first to contribute in the `Contributed by` comment and last to contribute in a single `Revised`, and that's it. But I realize there is a decent amount of sentiment for applying more subtlety than that.

I don't much care whether there is a `Revised` line for my changes, and given the mostly mechanical nature of most of them I'm not sure there's an especially strong objective case for it.

Now, if there is no existing `Revised` line the concerns about clutter are presumably less (although we might just be delaying the question until the next time someone revises it, so I'd be fine with not adding a `Revised` line there either).

As for djuen in particular, that looks like an oversight. I looked over the pull request and this seems to affect `djuen` and `endjudisj` so I pushed up changes to the comments for those two.

set.mm
jkingdon: Heh, definitely agree with point 2 (especially in this case, where there are only 5 usages and at the current rate I'll be the one to change them all).

As for point 1, I'd already changed it before I saw this comment. Not sure I have a hugely strong opinion which I guess gets me back to point 2.

changes-set.txt
avekens: OK, if you add them when they are removed.

set.mm
avekens: For me, it's Ok now.

avekens (approved):

PR #3431, opened 2023-08-25
Minimized theorems previously depending on df-clel, but not ax-8

### LINKED ISSUE #3402: Minimize proofs depending on df-cleq (resp. df-clel) but until #3389 not on ax-9 (resp. ax-8)
After the revision of df-cleq and df-clel in #3389, every theorem depending on df-cleq (resp. df-clel) now also depends on ax-9 (resp. ax-8).

Consider a theorem which depended, before #3389, on df-cleq but not on ax-9, and suppose it has a shorter proof using ax-9. That proof could not be found by `MM-PA> MINIMIZE_WITH *`, since that command does not consider statements adding axiom dependencies.

Therefore, one should do a global `MM-PA> MINIMIZE_WITH *` on every theorem which before #3389 depended on df-cleq but not on ax-9, and on every theorem which before #3389 depended on df-clel but not on ax-8.

Anyone up for it, just assign yourself to this issue (or comment here if you don't have the rights to assign).

GinoGiotto:
Fixes #3402

The set.mm version preceding #3389 contained 7922 theorems dependent on `df-clel` but not `ax-8`. This PR is the result of a minimization scan of only those theorems. To my surprise this activity only took one night, which is relatively short.

The command executed is the standard `MINIMIZE_WITH * /ALLOW_NEW_AXIOMS * /NO_NEW_AXIOMS_FROM ax-*`.

In the whole set.mm only `bj-ax9` and `bj-ax9-2` used to depend on `df-cleq` but not `ax-9`, so there is nothing to minimize there.

Some theorems in the set.mm version preceding #3389 used to have different names than now, so after minimizing I collected the ones generating the error `?No $p statement label matches <label>` and rescanned them with their actual name.

Some thoughts by looking at the changes:
* I suspect `wl-dfclab` was meant to have a "proof modification is discouraged" tag (maybe `bj-epelg` as well?).
* Theorems `abeq12` `brabg2a` `iunxsngf2` `rabeqd` `fnfvima2` seem to be redundant.

The full comparison in axiom usage can be consulted here: 3d9d2a0.

NOTE: The reason commits skip from [min1000](https://github.com/metamath/set.mm/pull/3431/commits/de85cb62e2ff8c3ca91d6f11998e1544dde1b4b4) to [min3000](https://github.com/metamath/set.mm/pull/3431/commits/a6232ac647440b60f6317c52e886c0a20ae580b7) is because min2000 didn't show any result.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,11 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+27-Aug-23 fnfvima2  ---         deleted; duplicate of fnfvimad
+27-Aug-23 rabeqd    ---         deleted; duplicate of rabeqdv
+27-Aug-23 iunxsngf2 ---         deleted; duplicate of iunxsngf
+27-Aug-23 brabg2a   ---         deleted; duplicate of brab2a
+27-Aug-23 abeq12    ---         deleted; duplicate of abbi1
 23-Aug-23 ovmpt2a   ovmpoa
 23-Aug-23 ovmpt2ga  ovmpoga
 23-Aug-23 ovmpt2x   ovmpox
--- a/discouraged
+++ b/discouraged
@@ -1883,7 +1883,6 @@
 "bnj1294" is used by "bnj1296".
 "bnj1294" is used by "bnj1379".
 "bnj1294" is used by "bnj1421".
-"bnj1294" is used by "bnj1450".
 "bnj1294" is used by "bnj1489".
 "bnj1294" is used by "bnj1501".
 "bnj1294" is used by "bnj1523".
@@ -13954,7 +13953,7 @@ New usage of "bnj1280" is discouraged (1 uses).
 New usage of "bnj1286" is discouraged (1 uses).
 New usage of "bnj1292" is discouraged (4 uses).
 New usage of "bnj1293" is discouraged (4 uses).
-New usage of "bnj1294" is discouraged (10 uses).
+New usage of "bnj1294" is discouraged (9 uses).
 New usage of "bnj1296" is discouraged (1 uses).
 New usage of "bnj1299" is discouraged (3 uses).
 New usage of "bnj130" is discouraged (1 uses).
--- a/discouraged
+++ b/discouraged
@@ -18458,6 +18458,7 @@ Proof modification of "bj-eeanvw" is discouraged (32 steps).
 Proof modification of "bj-el" is discouraged (48 steps).
 Proof modification of "bj-elisset" is discouraged (24 steps).
 Proof modification of "bj-elissetv" is discouraged (25 steps).
+Proof modification of "bj-epelg" is discouraged (33 steps).
 Proof modification of "bj-equs45fv" is discouraged (43 steps).
 Proof modification of "bj-equsal" is discouraged (31 steps).
 Proof modification of "bj-equsalhv" is discouraged (10 steps).
@@ -18715,6 +18716,7 @@ Proof modification of "dmtrclfvRP" is discouraged (46 steps).
 Proof modification of "dral1ALT" is discouraged (34 steps).
 Proof modification of "dral2-o" is discouraged (14 steps).
 Proof modification of "drnfc1OLD" is discouraged (52 steps).
+Proof modification of "drnfc2" is discouraged (52 steps).
 Proof modification of "dtruALT" is discouraged (79 steps).
 Proof modification of "dtruALT2" is discouraged (80 steps).
 Proof modification of "dummylink" is discouraged (1 steps).
@@ -19917,6 +19919,7 @@ Proof modification of "vtoclALT" is discouraged (11 steps).
 Proof modification of "vtxdusgr0edgnelALT" is discouraged (94 steps).
 Proof modification of "wfrlem4OLD" is discouraged (543 steps).
 Proof modification of "wl-cases2-dnf" is discouraged (85 steps).
+Proof modification of "wl-dfclab" is discouraged (73 steps).
 Proof modification of "wl-embant" is discouraged (12 steps).
 Proof modification of "wl-equsal" is discouraged (32 steps).
 Proof modification of "wl-impchain-a1-1" is discouraged (4 steps).


set.mm
```diff
@@ -534405,10 +534370,7 @@ sides of the biconditional correlate (they are the same), if they exist at
     $( Rederive ~ df-clab from ~ wl-clabv .  (Contributed by Wolf Lammen,
        31-May-2023.) $)
     wl-dfclab $p |- ( x e. { y | ph } <-> [ x / y ] ph ) $=
```
avekens, line 534372: @wlammen This theorem can be removed!?

set.mm
```diff
@@ -552405,7 +552367,7 @@ A collection of theorems for commuting equalities (or
   $( Equality deduction for class abstraction.  (Contributed by Giovanni
      Mascellani, 10-Apr-2018.) $)
   abeq12 $p |- ( A. x ( ph <-> ps ) -> { x | ph } = { x | ps } ) $=
```
avekens, line 552369: This trheorem can be removed!?

set.mm
```diff
@@ -620362,11 +620313,7 @@ fixed reference functional determined by this vector (corresponding to
     $( Generalized version of ~ brabga when the abstraction is in the style of
        ~ opelopab2a .  (Contributed by Steven Nguyen, 29-Apr-2023.) $)
     brabg2a $p |- ( A R B <-> ( ( A e. C /\ B e. D ) /\ ps ) ) $=
```
avekens, line 620315: This theorem can be removed!?

set.mm
```diff
@@ -659240,8 +659175,7 @@ unification theorem (e.g., the sub-theorem whose assertion is step 5
     $( A singleton index picks out an instance of an indexed union's argument.
        (Contributed by Glauco Siliprandi, 17-Aug-2020.) $)
     iunxsngf2 $p |- ( A e. V -> U_ x e. { A } B = C ) $=
```
avekens, line 659177: This theorem can be removed!?

set.mm
```diff
@@ -659770,7 +659702,7 @@ unification theorem (e.g., the sub-theorem whose assertion is step 5
     $( Equality theorem for restricted class abstractions.  (Contributed by
        Glauco Siliprandi, 26-Jun-2021.) $)
     rabeqd $p |- ( ph -> { x e. A | ch } = { x e. B | ch } ) $=
```
avekens, line 659704: This theorem can be removed!?

set.mm
```diff
@@ -662217,9 +662122,7 @@ not even needed (it can be any class).  (Contributed by Glauco
     $( Given an element of the preimage, its function value is in the image.
        (Contributed by Glauco Siliprandi, 2-Jan-2022.) $)
     fnfvima2 $p |- ( ph -> ( F ` B ) e. ( F " C ) ) $=
```
avekens, line 662124: This theorem can be removed!?

avekens (changes requested):
There seems to be some redundant theorems now (minimization resulted in one-line-proofs). These should be consolidated (one of them to be removed, the contribution information must be aligned).

avekens:
> Some thoughts by looking at the changes:
> 
>     * I suspect `wl-dfclab` was meant to have a "proof modification is discouraged" tag (maybe `bj-epelg` as well?).
> 
>     * Theorems `abeq12` `brabg2a` `iunxsngf2` `rabeqd` `fnfvima2` seem to be redundant.
> 
OK, then my review remarks are already known. About wl-dfclab, @wlammen should decide, the other 5 theorems should be consolidated.

GinoGiotto:
@benjub what do I do with `bj-epelg`?

benjub:
> @benjub what do I do with `bj-epelg`?

It should have the "proof discouraged" tag, but if it's easier for you, don't worry, leave it like that and I'll restore the proof later, with the tag.

Thanks a lot for this PR and the very rigorous methods you applied.

As for the absence of theorems requiring df-cleq but not ax-9, now I remember that it's because at some point, Norm proved the hypothesis of dfcleq using axext3 and not ax-ext in order to achieve that result, because of bj-ax9. I don't know if there was a global minimization since then, but in any case it's not too important, and it can wait a future global minimization.

set.mm
benjub: Probably not; the old proof should be restored with a proof discouragement tag.

set.mm
benjub: Probably yes, given the current proof.

set.mm
benjub: Steven said yes.

set.mm
benjub: Probably yes, given the current proof (unless @glacode says differently).

set.mm
```diff
@@ -638026,9 +637972,7 @@ adopting f( ` ph ` ) as ` if- ( ph , ps , ch ) ` would result in the
   $( The identity relation is hereditary in any class.  (Contributed by RP,
      28-Mar-2020.) $)
   idhe $p |- _I hereditary A $=
-    ( cid whe cres cxp wss cdm crn relres relssdmrn ax-mp dmresi eqimssi rnresi
-    wrel xpss12 mp2an sstri dfhe2 mpbir ) ABCBADZAAEZFUAUAGZUAHZEZUBUAOUAUEFBAI
-    UAJKUCAFUDAFUEUBFUCAALMUDAANMUCAUDAPQRABST $.
+    ( cid whe cres cxp wss idssxp dfhe2 mpbir ) ABCBADAAEFAGABHI $.
```
benjub, line 637975: It's ok to me, unless @arpie-steele says otherwise.

set.mm
```diff
@@ -63368,10 +63368,10 @@ from the cartesian product of two singletons onto a singleton (case where
...
        Siliprandi, 23-Oct-2021.) $)
     fnfvimad $p |- ( ph -> ( F ` B ) e. ( F " C ) ) $=
-      ( cin cima cfv wss inss2 imass2 ax-mp a1i wfun cdm wa wcel fnfun sseqtr4d
-      wfn syl inss1 fndmd jca elind funfvima2 sylc sseldd ) AEBDIZJZEDJZCEKZUMU
-      NLZAULDLUPBDMULDENOPAEQZULERZLZSCULTUOUMTAUQUSAEBUCUQFBEUAUDAULBURULBLABD
-      UEPABEFUFUBUGABDCGHUHULCEUIUJUK $.
+      ( cin cima cfv wss inss2 imass2 ax-mp wfun cdm wa wcel wfn fnfun syl sylc
+      inss1 fndmd syl5sseqr jca elind funfvima2 sseldi ) AEBDIZJZEDJZCEKZUKDLUL
+      UMLBDMUKDENOAEPZUKEQZLZRCUKSUNULSAUOUQAEBTUOFBEUAUBABUKUPBDUDABEFUEUFUGAB
+      DCGHUHUKCEUIUCUJ $.
```
benjub, line 63374: The proof of this statement given at fnfvima2 is apparently shorter. It should probably be put here, with a further `MINIMIZE_WITH *` just in case.

set.mm
benjub: Yes, but the current proof should actually be kept as the main one for fnfvimad (same contributor @glacode).

benjub:
I looked at some of the more significant shortenings to see if the non-minimized proofs had insight that would be lost in the minimization, but all seem good to me (and some minimizations would have occurred even before the change).

set.mm
benjub: Yes (same proof and same contributor). Since it has several usages, they should be adapted. @GinoGiotto : do you know how to do it easily ? (if you use metamath.exe, you can `minimize_with rabeqdv/may_grow` on all theorems using it, etc.; or you can save the proofs in the "normal" format, do the label change, and save the proofs back in the "compressed" format, I guess).

icecream17:
feel free to delegate brabg2a to me

In fact, to avoid merge conflicts, perhaps this pr could be merged and the duplicate theorems fixed in another pr

GinoGiotto:
> In fact, to avoid merge conflicts, perhaps this pr could be merged and the duplicate theorems fixed in another pr

I fear merge conflicts too, perhaps that's a good idea, either way I promise I will commit the non-self-assigned ones tomorrow or earlier (I can't work on it right now).

benjub:
Ok. So if we merge now to avoid merge conflicts, there remain the following TODOs:
- restore proof and add "proof disc" tag to: wl-dfclab and bj-epelg
- remove the theorems: abeq12, brabg2a, iunxsngf2, rabeqd
- fnfvima2/fnfvimad: the shorter proof and remove the other; do a second `MINIMIZE_WITH *`; for the label, I would choose "fnfvmiad"
- revert the "shortening" of drnfc2 and understand why the minimization made that change

set.mm
```diff
@@ -26678,8 +26678,8 @@ choice between (what we call) a "definitional form" where the shorter
     $( Formula-building lemma for use with the Distinctor Reduction Theorem.
        (Contributed by Mario Carneiro, 8-Oct-2016.) $)
     drnfc2 $p |- ( A. x x = y -> ( F/_ z A <-> F/_ z B ) ) $=
-      ( vw weq wal cv wcel wnf wnfc eleq2d drnf2 albidv df-nfc 3bitr4g ) ABHAIZ
-      GJZDKZCLZGITEKZCLZGICDMCEMSUBUDGUAUCABCSDETFNOPCGDQCGEQR $.
+      ( vw weq wal cv wcel wnf wnfc eleq2d drnf2 dral2 df-nfc 3bitr4g ) ABHAIZG
+      JZDKZCLZGITEKZCLZGICDMCEMUBUDABGUAUCABCSDETFNOPCGDQCGEQR $.
```
benjub, line 26682: Keep the old proof here. It is actually strange that "minimize_with" produced this result... Anyone understands why ?

set.mm
glacode: commit d6a37f4b30c1b0e4db3b0d382a10c8578cf2e2b0
moved iunxsngf from @tirix  mathbox to main, on Fri Oct 14 2022, I'm not entirely sure how to reconcile this with the iunxsngf's comment (Contributed by Mario Carneiro, 25-Jun-2016.)

In any case, I'm okay with its removal. Kindly ensure to update the proofs that make reference to iunxsngf  (iunxsnf , if a minimize didn't already change such reference)

set.mm
benjub: Answer:
```
Proof of "drnfc2" stayed at 129 bytes using "dral2".
    (Uncompressed steps decreased from 78 to 75).
```
I would still keep albidv, so maybe add a proof discouragement tag ?

benjub (approved):

GinoGiotto:
I think I did everything TODO, as suggested I reminimized fnfvimad generating an even shorter proof. 

> * fnfvima2/fnfvimad: the shorter proof and remove the other; do a second `MINIMIZE_WITH *`; for the label, I would choose "fnfvmiad"

I suppose this is a typo, I kept the label "fnfvimad".

set.mm
GinoGiotto: Here the minimizer is referring to the total amount of bytes (labels inlcuded), but maybe for minimizations we are interested in the size of the upper-case section only. I'm not sure whether there was an agreement in the past about which counting matters most.

set.mm
GinoGiotto: Not necessary to save proofs in the normal format. I noticed replacing labels directly in the compressed proof works as well if the replaced statemetent is exactly identical to the old one (and when I finished I do `SAVE PROOF * /COMPRESSED /FAST`).

avekens (changes requested):
Great job! Only one minor thing:
It should be documented in changes-set.txt that the theorems abeq12 brabg2a iunxsngf2 rabeqd fnfvima2 were removed as duplicates, e.g.:

20-Mar-21 biimpa21  ---         deleted; duplicate of biimpac

GinoGiotto:
The additions to the `changes-set.txt` file generated a merge conflict, but I think I solved it.

avekens (approved):

set.mm
benjub: Ah, it's true. I usually do the above trick when the theorems differ only by variable renamings, in which case you need to to the change on the uncompressed proof.

set.mm
benjub: Yes, in case of tie, it's the block of capital letters. Since this may go against the action of `MINIMIZE_WITH` as in the present case, this is not very satisfying and the situation could be improved. But here, using albidv is definitely more "elementary" and natural than using ~dral2.

set.mm
tirix: > I'm not entirely sure how to reconcile this with the iunxsngf's comment (Contributed by Mario Carneiro, 25-Jun-2016.) 

Mario is the original author of the proof, I only revised it to use the `F/_ x C` explicit "not free in" hypothesis instead of the distinct variable statement, and added the **`f`** suffix. Therefore I kept Mario's attribution and only added my name as "Revised by", even though I originally only included the new theorem in my mathbox. That's why it's a theorem authored by Mario, but pulled from my Mathbox.

PR #3435, opened 2023-08-26
prove initial df-mhp theorems

icecream17:
and prove suppofssd (needed in mhpaddcl)
and minimize 0prjspn


set.mm
```diff
@@ -256002,6 +256024,118 @@ Additional definitions for (multivariate) polynomials
...
+    mhpfval.h $e |- H = ( I mHomP R ) $.
+    mhpfval.p $e |- P = ( I mPoly R ) $.
+    mhpfval.b $e |- B = ( Base ` P ) $.
+    mhpfval.0 $e |- .0. = ( 0g ` R ) $.
+    mhpfval.d $e |- D = { h e. ( NN0 ^m I ) | ( `' h " NN ) e. Fin } $.
+    mhpfval.i $e |- ( ph -> I e. V ) $.
+    mhpfval.r $e |- ( ph -> R e. W ) $.
+    $( Value of the "homogeneous polynomial" function.  (Contributed by Steven
+       Nguyen, 25-Aug-2023.) $)
+    mhpfval $p |- ( ph -> H = ( n e. NN0 |->
```
avekens, line 256039: According to our naming convention there should be a suffix d (mhpfval -> mhpfvald, also for the following theorems), because they are in deduction form.

icecream17: I added the suffixes but note that this was based on theorems like mvrfval and psraddcl

benjub: Actually, we add the suffix "d" when both versions are in set.mm, but for more advanced theorems, deduction form is the default, and generally only, form, so the "d" is not necessary (I would actually remove it).

tirix (approved):

tirix:
I should have noted and mentioned that when the symbol `mHomP `was introduced for homogeneous polynoms, but I only notice it now. We use `Hom` commonly for all kinds of Homomorphisms (Monoid, Group, Ring, etc. as well as Category Theory's Hom sets). 

Maybe we could find a different abbreviation which would not overlap with `Hom`?

set.mm
avekens: I must admit that @benjub is right, suffix "d" is not required in this case (the Methamath book is more explicit than the comment in set.mm for this convention), so it can be removed again. Sorry, @icecream17, for the inconvenience.

avekens:
> I should have noted and mentioned that when the symbol `mHomP `was introduced for homogeneous polynoms, but I only notice it now. We use `Hom` commonly for all kinds of Homomorphisms (Monoid, Group, Ring, etc. as well as Category Theory's Hom sets).
> 
> Maybe we could find a different abbreviation which would not overlap with `Hom`?

I see no problems with mHomP (and have no idea how to call it instead), because Hom used as suffix means homomorphism. Furthermore, mHomP was introduce by @digama0 in 2015.

avekens (approved):

tirix:
> have no idea how to call it instead

For example `HmgP` would still resemble "HoMoGeneous Polynom" and not any Homeomorphism.

> Hom used as suffix means homomorphism. Furthermore, mHomP was introduce by @digama0 in 2015.

Actually I should rather have made that an issue. I have already approved this PR, it's good to go for me. I'll merge.

PR #3434, opened 2023-08-26
assorted mathbox fixes

wlammen:
1. delete wl-nfnbi.  It is already part of main
2. add a no modification tag to wl-dfclab, as requested here #3431
3. Add a proof that someone thinking 9 + 10 = 21 is in fact a fool, see #3433

PR #3433, opened 2023-08-26
Proof that 9 + 10 is not equal to 21

BTernaryTau:
> 9 + 10 is not equal to 21. This disproves a popular meme which asserts that 9 + 10 does equal 21. See https://www.quora.com/Can-someone-try-to-prove-to-me-that-9+10-21 for attempts to prove that 9 + 10 = 21, and see https://tinyurl.com/9p10e21 for the history of the 9 + 10 = 21 meme.

PR #3439, opened 2023-08-27
Shorten 19.42vvv

wlammen:
1. Finish #3434 , obsolete #3433 
2. shorten 19.42vvv
3. delete outdated OLD theorem

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -10802,7 +10802,6 @@
 "opelreal" is used by "ltresr".
 "opelresOLD2" is used by "elresOLD".
 "opelresgOLD2" is used by "brresgOLD2".
-"opelresgOLD2" is used by "idrefOLD".
 "opelresgOLD2" is used by "opelresOLD2".
 "opidon2OLD" is used by "exidreslem".
 "opidonOLD" is used by "opidon2OLD".
@@ -13380,6 +13379,7 @@ New usage of "19.23tOLD" is discouraged (0 uses).
 New usage of "19.41rg" is discouraged (3 uses).
 New usage of "19.41rgVD" is discouraged (0 uses).
 New usage of "19.42-1OLD" is discouraged (0 uses).
+New usage of "19.42vvvOLD" is discouraged (0 uses).
 New usage of "19.43OLD" is discouraged (0 uses).
 New usage of "1dimN" is discouraged (0 uses).
 New usage of "1div0" is discouraged (0 uses).
@@ -13483,6 +13483,7 @@ New usage of "6cnOLD" is discouraged (0 uses).
 New usage of "7cnOLD" is discouraged (0 uses).
 New usage of "8cnOLD" is discouraged (0 uses).
 New usage of "9cnOLD" is discouraged (0 uses).
+New usage of "9p10ne21stupid" is discouraged (0 uses).
 New usage of "a1ii" is discouraged (0 uses).
 New usage of "abbi2dvOLD" is discouraged (0 uses).
 New usage of "abbi2iOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


discouraged
```diff
@@ -18308,6 +18308,7 @@ Proof modification of "6cnOLD" is discouraged (3 steps).
 Proof modification of "7cnOLD" is discouraged (3 steps).
 Proof modification of "8cnOLD" is discouraged (3 steps).
 Proof modification of "9cnOLD" is discouraged (3 steps).
+Proof modification of "9p10ne21stupid" is discouraged (33 steps).
```
benjub, line 18311:
```suggestion
Proof modification of "9p10ne21fool" is discouraged (33 steps).
```

set.mm
```diff
@@ -426627,6 +426616,30 @@ This law is thought to have originated with Aristotle (_Metaphysics_,
...
+      UCUFUEGHIUCJKUFUHUFUDUCLZMZNPUDUGUCUDUIQUJUIRSTUAUB $.
+
+    $( 9 + 10 equals 21.  This astonishing thesis lives as a meme on the
+       internet, and may be believed by quite some people.  At least repeated
+       requests to falsify it are a permanent part of the story.  Prof.  Loof
+       Lirpa did not rest until he finally came up with a computer verifiable
+       mathematical proof, that only a fool can think so.  (Contributed by
+       Prof.  Loof Lirpa, 26-Aug-2023.)  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    9p10ne21stupid $p |- ( ( 9 + ; 1 0 ) = ; 2 1 -> F (/) ( 0 x. 1 ) ) $=
```
benjub, line 426638:
```suggestion
    9p10ne21fool $p |- ( ( 9 + ; 1 0 ) = ; 2 1 -> F (/) ( 0 x. 1 ) ) $=
```
I'd rather not have derogatory terms in set.mm. ("fool" can be alllowed because of the "April Fools" phrase)

benjub (approved):

wlammen:
I think one should stop for a moment at #3433 . What is different there compared to all other PRs? We dedicate most (99.99%) of our efforts to develop the scientific nature of the database, and it is our goal to keep it that way.  Only very few entries like avril1 were formerly added to allow for a niche of humorous additions.  Those were to my knowledge based on an intrinsic motivation:  Authors were already part of the development team, and had fun to explore how the syntax of Metamath could be exploited to express something unrelated to math.

This contribution is different.  Here we react to an outside event.  And we are utilized to contribute to this event instead to our domain of interest.  As long as this happens seldom enough (once in a few years) this is likely not important enough to really care about.  But one should be aware of the situation in my eyes.

tirix (approved):

jkingdon:
> And we are utilized to contribute to this event instead to our domain of interest

I guess my domain of interest includes at least a little bit how people are engaging with math.

Something like tau = 2 x pi is a bit more serious (in the sense that the factor of two does have implications for math education) but it is also something with little relevance one way or the other to the research mathematician (who has learned to deal with the factor of two). It does, however, have a fair bit of Internet notoriety. I'd argue we should have it, partly because people are talking about it.

Will our database be filled with silly theorems like 9 + 10 ≠ 21? Well, if it is I suppose we could put them in their own section or own file or own database or some suitable organization, but I would be a bit surprised if that issue came up on a significantly larger scale than what we have seen so far. There are only so many of these circulating much less those noticed by someone who gets around to formalizing them for metamath. So at least from my point of view I'm not sure I see any particular need to worry too much about restrictions or limits on how many to have.

tirix:
It would have been nice if @BTernaryTau could have been mentioned as a GitHub author of this commit. (I think the tools allow that) This way she could have been listed in the list of contributors to the repository. Even for a simple proof, that's still an achievement!

jkingdon:
> It would have been nice if @BTernaryTau could have been mentioned as a GitHub author of this commit. (I think the tools allow that) This way he could have been listed in the list of contributors to the repository. Even for a simple proof, that's still an achievement!

You mean like this? https://github.com/metamath/set.mm/pull/3440

Note the co-authored-by line - that's the usual way to get github to notice multiple people on a commit.

tirix:
> You mean like this? #3440

Yes, exactly! Thank you.

PR #3445, opened 2023-08-28
fix download-metamath script for github changes

### LINKED ISSUE #3444: CI can't download symbols.tar.gz
Something appears to have changed which is breaking the master build: https://github.com/metamath/set.mm/actions/runs/6005609380/job/16288830886

I believe the issue is https://github.com/orgs/community/discussions/65227 .

digama0:
fixes #3444

### DIFF
--- a/scripts/download-metamath
+++ b/scripts/download-metamath
@@ -56,8 +56,9 @@ rm -f symbols.tar.bz2 symbols.tgz
 # website is generated from version controlled files, so we need to
 # get the symbols from our version controlled sources.
 # wget -N http://us.metamath.org/downloads/symbols.tar.bz2
+# (Github seems to require TLSv1_2 now, see #3444)
 
-wget -N -O symbols.tgz https://github.com/metamath/symbols/tarball/main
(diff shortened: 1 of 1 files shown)


benjub (approved):

benjub:
I think it's good, although I'm not sure if TLS 1.2 is "secure enough" or should be replaced with TLS 1.3.

jkingdon (approved):

jkingdon:
> I think it's good, although I'm not sure if TLS 1.2 is "secure enough" or should be replaced with TLS 1.3.

I guess `TLSv1_3` would make sense although `TLSv1_2` is clearly an improvement over the default.

I'm looking at https://github.com/orgs/community/discussions/65227#discussioncomment-6846944

jkingdon:
Seemed worth merging now to get things passing again.

david-a-wheeler:
I'm hoping in the long-term we can find a better solution - maybe changing the header to say "I'm not wget". But working is better than not-working.

PR #3446, opened 2023-08-30
Drop ax-13 from kmlem4

GinoGiotto:
This revision drops ax-13 from `kmlem5` and `kmlem11` as well.


tirix (approved):
LGTM

set.mm
```diff
@@ -98481,15 +98481,15 @@ our Axiom of Choice (in the form of ~ ac2 ).  The proof does not depend
...
-      elequ1 anbi12d elequ2 notbid imbi12d eldif simpr eluni notbii alnex con2b
-      spv wex imnan eldifsn necom anbi2i bitri imbi1i 3bitr3i albii sylib sylbi
-      3bitr2i syl11 ralrimiv disj sylibr ) CAFZBGZCGZHZIZDCFZJZDVKAGZVKKLZMZLZN
-      VTVLOPQVNVPDVTEAFZVKEGZHZIZDEFZJZRZESZVNVPDGZVTTZWGVNVPRECWBVLQZWDVNWFVPW
-      KWAVJWCVMECAUBWBVLVKUAUCWKWEVOECDUDUEUFUMWJDBFZWIVSTZJZIZWHWIVKVSUGWOWNWH
-      WLWNUHWNWEWBVRTZIZEUNZJWQJZESWHWMWREWIVRUIUJWQEUKWSWGEWEWPJRWPWFRWSWGWEWP
-      ULWEWPUOWPWDWFWPWAWBVKHZIWDWBVQVKUPWTWCWAWBVKUQURUSUTVAVBVEVCVDVFVGDVTVLV
-      HVI $.
+      ( vy vv wel cv wne wa wn csn cdif cuni wral cin c0 wceq wi wal wcel eleq1
+      weq neeq2 anbi12d eleq2 notbid imbi12d eldif simpr wex eluni notbii alnex
```
benjub, line 98485: ahhh, I don't know what to do. elequ1/2 are more elementary than eleq1/2, so it would be more natural to use them, but it looks like the minimizer does not like them.

wlammen:
The kmlem* series are all lemmas or subproofs of dfackm.  They have no reference outside this theorem.  Removing ax-13 from a lemma will only help, if you succeed in removing them from all in the end.  Otherwise you just end up having  a longer proof without a benefit.

EDIT: The recent change makes this proof simply shorter, so this is ok from my side as well.

set.mm
GinoGiotto: Now the upper case section is shorter than the original by one byte

GinoGiotto:
> The kmlem* series are all lemmas or subproofs of dfackm. They have no reference outside this theorem. Removing ax-13 from a lemma will only help, if you succeed in removing them from all in the end. Otherwise you just end up having a longer proof without a benefit.

If this aligns with the general sentiment, shouldn't this be updated in conventions?

icecream17 (approved):

set.mm
benjub: It is strange the minimize_with didn't catch that shortening...

GinoGiotto: No no, the minimizer caught it, I just forgot it.

PR #3448, opened 2023-08-31
Add trilpo to iset.mm

### LINKED ISSUE #3423: Show that real number trichotomy implies LPO
According to https://ncatlab.org/nlab/show/principle+of+omniscience#analytic , real number trichotomy implies the Limited Principle of Omniscience, or in our notation:

```
$( Real number trichotomy implies the Limited Principle of Omniscience (LPO).
   We expect that we'd need some form of countable choice to prove the
   converse.  (Contributed by Jim Kingdon, 23-Aug-2023.) $)
trilpo $p |- ( A. x e. RR A. y e. RR ( x < y \/ x = y \/ y < x )
    -> _om e. Omni ) $=
  (  ) ? $.
```

What we need to prove (per https://us.metamath.org/ileuni/isomnimap.html ) is that given an infinite sequence F of zeroes and ones, either the sequence contains a zero or it is all ones.

Here's one sketch of a proof: Construct a real number A whose representation in base two consists of a zero, a decimal point, and then the numbers of the sequence. Compare with with one using trichotomy. Clearly we can rule out 1 < A . If A = 1 then it must be `0.1111111...` . If A < 1 then it must contain a zero in some position.

Our number would be ``sum_ i e. NN ( ( 1 / ( 2 ^ i ) ) x. ( F ` i ) )`` . See https://us.metamath.org/ileuni/geoihalfsum.html which shows that if the F's are all one, then the sum is one. But do we have a way to prove the reverse: if the sum is one, then the F's are all one?

Likewise, it feels like if A < 1 we can know how far out in the sequence to go, based on the distance between A and one, to get a zero. Chances are we'll need to compute that position in some fairly direct way to be able to prove it exists in the sense of `∃`.

jkingdon:
This shows that real number trichotomy implies the Limited Principle of Omniscience (LPO).

Includes isomninn which just restates omniscience in terms of 0 and 1 rather than (/) and 1o .  Includes lemma isomninnlem .

Includes cvgcmp2n which is a convergence test for infinite series similar to cvgcmp from set.mm.  Includes lemma cvgcmp2nlemabs .

Includes lemmas trilpolemclim and trilpolemcl which show that using our sequence from isomninn as binary positional digits does indeed produce a real number.

Includes lemmas trilpolemgt1 , trilpolemeq1 , and trilpolemlt1 which handle the three cases from trichotomy.

Includes lemma trilpolemres which assembles those cases into (basically) the result.

Includes lemma trilpolemisumle

Fixes #3423

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -8577,9 +8577,8 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <TR>
   <TD>cvgcmp</TD>
-  <TD><I>none</I></TD>
-  <TD>Apparently would need an additional condition on the rate of
-  convergence.  The set.mm proof relies on caurcvg2 which does not
+  <TD>~ cvgcmp2n</TD>
+  <TD>The set.mm proof of cvgcmp relies on caurcvg2 which does not
   specify a rate of convergence.</TD>
 </TR>
 


iset.mm
```diff
@@ -145070,6 +145070,402 @@ Limited Principle of Omniscience (LPO) implies excluded middle, so we
...
+       (LPO).  We expect that we'd need some form of countable choice to prove
+       the converse.
+
+       Here's the outline of the proof.  Given an infinite sequence F of zeroes
+       and ones, we need to show the sequence contains a zero or it is all
+       ones.
+
+       Construct a real number A whose representation in base two consists of a
+       zero, a decimal point, and then the numbers of the sequence.  Compare
+       with with one using trichotomy.  The three cases from trichotomy are
```
tirix, line 145450: I guess you mean compare "it" with one.

tirix:
Nice build-up!

iset.mm
jkingdon: I made that change, thank you.

tirix (approved):

jkingdon:
I guess there's no reason to delay merging. But if people still have any comments, there can always be a follow up pull request.

PR #3454, opened 2023-09-03
Some theorems about ordered/unordered pairs and interchangeable setvar variables

avekens:
main:
* ~jcn moved from GS's mathbox to main

AV's mathbox:
* ordered pairs: ~oppr and ~opprb moved up
* ordered pairs:~or2expropbilem1, ~or2expropbilem2, ~or2expropbi added
* interchangeable setvar variables: ~nfich1, ~nfich2, ~ichcom added
* interchangeable setvar variables and ordered pairs: ~ich2exprop, ~ichnreuop, ~ichreuopeq  added

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 3-Sep-23 jcn       [same]      moved from GS's mathbox to main set.mm
 31-Aug-23 mpt2ex    mpoex
 31-Aug-23 mpt2exga  mpoexga
 31-Aug-23 mpt2exg   mpoexg


set.mm
```diff
@@ -3512,6 +3512,15 @@ intermediate steps that are essentially incomprehensible to humans (other
       ( wb a1i pm5.74i ) CABABECDFG $.
   $}
 
+  ${
+    jcn.1 $e |- ( ph -> ps ) $.
+    jcn.2 $e |- ( ph -> -. ch ) $.
+    $( Inference joining the consequents of two premises.  (Contributed by
+       Glauco Siliprandi, 11-Dec-2019.) $)
+    jcn $p |- ( ph -> -. ( ps -> ch ) ) $=
+      ( wn wi jc notnotb imbi2i sylnibr ) ABCFZFZGBCGABLDEHCMBCIJK $.
```
benjub, line 3521: This looks like the deduction associated with mth8, but mth8 too has a peculiar label.

So maybe ~notim and ~notimd ? With a comment like "Sufficient condition for an implication to be false." and "Sufficient condition for an implication to be false. Deduction form."

(Rk: jc and jca should also be labeled *d)

avekens: "jc" stands for "joining the consequents", and I think ~jcn fits to the already existing theorems  ~jc  (adding a negation to ~jc), ~jca and many other theorems prefixed with jc.

~jca cannot be renamed ~jcad, because ~jcad is already existing.

benjub: jcad is a double deduction, so could be labeled jcadd. Independently of the precise labels, which could change, I think the convention of having *d for deductions should prevail, given its predominant use in deduction-style proofs.

This can be done later and doesn't block this PR.

avekens: I think the revision of the jc* theorems is a separate topic and should be done in another PR.

digama0: > I think the revision of the jc* theorems is a separate topic and should be done in another PR.

Relevant: while I mainly borrowed the metamath naming conventions for MM0 the intro/elim rules for and/or is one case where I diverged:

* and intro: MM: `jca`, MM0: `ian`
* and elim: MM: `simpl` / `simpr`, MM0: `anl` / `anr`
* or intro: MM: `olc` / `orc`, MM0: `orl` / `orr`
* or elim: MM: `jao`, MM0: `eor`

These also contain i/d versions as appropriate.

tirix (approved):

set.mm
benjub: It's nice to have mnemonics based on "and intro", "or elim", etc., which are the most standard names for these theorems/axioms. I would favor adopting this labeling scheme.

PR #3455, opened 2023-09-03
shorten dfsb7

wlammen:


benjub (approved):
I think both the OLD and the OLDOLD versions can be removed.

PR #3463, opened 2023-09-05
sbco2v: x and y need not be distinct

wlammen:


avekens (approved):

benjub:
Does it make it more usable, for instance, are there theorems using sbco2 which could now use sbco2v instead ?

wlammen:
This improvement has no immediate effect, to my knowledge

Issue #3461, opened 2023-09-05
Sourcing mmbiblio.html


tirix:
The GitHub verifier pulls `mmbiblio.html` from metamath.org, this led to CI failures in #3459 and #3460 - see [comment](https://github.com/metamath/set.mm/pull/3459#issuecomment-1705983800).

Should we replace it with downloading this file directly from GitHub?
[https://raw.githubusercontent.com/metamath/metamath-website-seed/main/mpegif/mmbiblio.html](https://raw.githubusercontent.com/metamath/metamath-website-seed/main/mpegif/mmbiblio.html)

PR #3464, opened 2023-09-05
Split metamathexe and tex jobs

jkingdon:
The metamath.exe job has acquired a lot of semi-related functionality which is OK up to a point but the script is getting a bit hard to read and perhaps more importantly, the tex job (in particular, installing the tex packages) is a lot slower than the other checks.  So by splitting it out we can speed up the whole run a bit, and get faster feedback if there is an error.

Although I noticed this when looking at #3461 this isn't anything which is down or broken, just maybe a way to make it a bit better.


tirix (approved):
I agree and I support this change!

PR #3460, opened 2023-09-05
Reduced ax-13 usage with bj-spime(d)v

GinoGiotto:
* Moved `bj-spimedv` from @benjub's mathbox to main.
* Renamed `bj-spimedv` into `spimedv`.
* Scanned 100 theorems to find applications for `spimedv` (I noticed big scans aren't worth it for axiom minimizations).
* Found a proof of `dtru` which drops ax-13 thanks to `spimedv`.

The axiom usage reduction of `dtru` automatically drops ax-13 for other 65 theorems: https://github.com/metamath/set.mm/commit/095ad0870d571010339ba67cd5f16d93551bff88

UPDATE: Theorem `bj-spimev` -> `spimefv` was used instead to shorten the proof of `dtru`.

PR #3466, opened 2023-09-06
Reduced ax-13 usage with bj-speiv

GinoGiotto:
* Moved `bj-speiv` from @benjub's mathbox to main.
* Renamed `bj-speiv` into `speiv`.
* Scanned a few theorems in main to find applications for `speiv`.
* Found a proof of `elirrv` which doesn't depend on ax-13 thanks to `speiv`.

 A total of 12 theorems dropped ax-13 with the revision in this PR: https://github.com/metamath/set.mm/commit/b1302dc84cadb1a9e0c848e697f14015002c79d5

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 6-Sep-23 bj-speiv  speiv       moved from BJ's mathbox to main set.mm
  4-Sep-23 mpt2xopynvov0 mpoxopynvov0
  4-Sep-23 mpt2xopxprcov0 mpoxopxprcov0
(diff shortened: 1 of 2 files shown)


wlammen:
In fact, spei is the only theorem in the current proof of elirrv based on ax-13, so replacing it removes the dependency on ax-13.  What I miss here is a revision tag indicating the removal of ax-13 from the proof, or, alternatively, adding a
`$( $j usage 'elirrv' avoids 'ax-13'; $)` ,
so this axiom is not accidentally added in a future modification.

wlammen (approved):

GinoGiotto:
I solved the conflict of the files `changes-set.txt` and `discouraged` generated by my previous PR #3460 that got merged a few minutes ago. Luckily this was an easy fix.

PR #3468, opened 2023-09-07
Provide |_| versions of +c theorems from cdadom3 to pwcdaidm

jkingdon:
The most striking thing here is that the converse of https://us.metamath.org/mpeuni/djuex.html can be proved in both set.mm and iset.mm (this affects +c versus |_| because the latter is defined on proper classes).

Other than that, pretty similar to other recent pull requests.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -11565,6 +11565,11 @@
 "psubspi2N" is used by "pclclN".
 "psubspi2N" is used by "pclfinN".
 "psubspi2N" is used by "pclfinclN".
+"pwcda1" is used by "cdalepw".
+"pwcda1" is used by "gchcdaidm".
+"pwcda1" is used by "gchpwdom".
+"pwcda1" is used by "pwcdaidm".
+"pwcda1" is used by "pwsdompw".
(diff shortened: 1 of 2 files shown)


tirix (approved):

avekens (approved):

benjub:
> The most striking thing here is that the converse of https://us.metamath.org/mpeuni/djuex.html can be proved in both set.mm and iset.mm

Indeed, this is a nice property, and it's one of the reasons why I proposed that couples (ordered pairs) use a similar definition, see https://us.metamath.org/mpeuni/bj-2uplex.html It predates the inclusion of disjoint unions in set.mm, but it is almost the same.

jkingdon:
> it's one of the reasons why I proposed that couples (ordered pairs) use a similar definition

I suspect the Kuratowski ordered pair is pretty well entrenched in terms of being our main definition but reading up on this was kind of interesting. Including the terminology "ordered pair" which is even more entrenched but which does have the oddity that "ordered" doesn't refer to a partial order or anything of the sort.

PR #3470, opened 2023-09-07
add theorems regarding df-frlm and words, and df-ich

icecream17:
The theorems regarding df-frlm et al and its interaction with words will be needed for theorems about prjspn

@avekens ichn and ichal are in your mathbox
I'm pretty interested in this new syntax actually, should it be moved to main?

Incidentally, [dfich2ai](https://us.metamath.org/mpeuni/dfich2ai.html) proves `⊢ (∀𝑥∀𝑦([𝑥 / 𝑧][𝑦 / 𝑥][𝑧 / 𝑦]𝜑 ↔ 𝜑) → ([𝑎 / 𝑥][𝑏 / 𝑦]𝜑 ↔ [𝑏 / 𝑥][𝑎 / 𝑦]𝜑))` but really it proves dfich2 + 2sp

image


icecream17:
hmm, 503 connection error?

tirix (approved):

tirix:
> hmm, 503 connection error?

I've re-run it and it passed, probably some spurious problem.

jkingdon:
> ichn and ichal are in your mathbox
> I'm pretty interested in this new syntax actually, should it be moved to main?

Clearly the concept is a familiar one in contexts like "without loss of generality, suppose it is x which (has some property) instead of y".

I'm maybe not completely sold on having a notation for it, I guess because it is unfamiliar and maybe a hurdle to the reader. But that's without having a very clear idea of how it would be used and what the alternative would be. Even if this is an issue, it might be an argument for using it sparingly rather than not having it at all, however.

jkingdon (approved):
Won't claim to be an expert on these parts of set.mm but it all seems sensible.

icecream17:
technically, everything except shortening vtoclg1f is in my or AV's mathbox

on interchangability, there isn't really a need to move to main right now, just something to keep in mind about mathbox independence. there are several related concepts, like https://us.metamath.org/mpeuni/wlogle.html or https://us.metamath.org/mpeuni/ersym.html

jkingdon:
> on interchangability, there isn't really a need to move to main right now, just something to keep in mind about mathbox independence. there are several related concepts, like https://us.metamath.org/mpeuni/wlogle.html or https://us.metamath.org/mpeuni/ersym.html

Yeah, fair, the mathbox independence thing might force our hands a bit.

For things like ersym , https://us.metamath.org/mpeuni/ensymd.html , and some others like those, it isn't obvious that new notation helps.

But https://us.metamath.org/mpeuni/wlogle.html is a bit of a handful so I totally can see the appeal of trying to see whether an alternative would be better.

jkingdon:
Most of the above discussion is about what we might do in the future, no reason to delay a merge here.

PR #3440, opened 2023-08-27
Add BTernaryTau to contributors list

jkingdon:
This is intended to fulfill the suggestion at https://github.com/metamath/set.mm/pull/3439#issuecomment-1694728667 about noting this contribution

PR #3472, opened 2023-09-09
Expand mathbox guidelines

### LINKED ISSUE #3136: Abandoned work
Slowly, but inevitably, we will see more and more work in our databases like set.mm, that is not maintained any more, like mathboxes.  The reason why maintenance has stopped could be: lost interest in mathematics, or Metamath in particular, basic changes in life consuming time (like children), or, horrible dictu, simply death.  Currently, the amount of abandoned work has not accumulated to a degree that calls for immediate action, in my eyes.  Nevertheless I am curious how you think about this topic.  Should we in preparation for upcoming years develop rules by which we formally decide when to remove something apparently abandoned from the repositories?  Or are ad hoc decisions sufficient, in case something becomes urgent enough?  Looking forward to your opinions.

jkingdon:
The new text in set.mm is particularly around abandoned mathboxes, but the idea of moving important theorems to main (eventually) is a bit more general than that.

In the case of iset.mm, I think what had been there was just an old copy of the set.mm mathbox guidelines.  It seems better to just have a few introductory sentences and refer people to the set.mm guidelines, rather than try to keep two copies in sync.

The intention is just to write down what we already said in #3136 which did not appear to be controversial.

Fixes #3136


set.mm
```diff
@@ -459220,8 +459220,21 @@ SUPPLEMENTARY MATERIAL (USER'S MATHBOXES)
        5.  Mathboxes must be independent from one another (checked by "verify
        markup *").  If you need a theorem from another mathbox, typically it is
        moved to the main part of set.mm.  New users should consult with more
-       experienced users before doing this.  (Contributed by NM, 20-Feb-2007.)
-       (Proof modification is discouraged.)  (New usage is discouraged.) $)
+       experienced users before doing this.
+
+       6.  If a contributor is no longer active, we will continue the usual
+       maintenance edits.  As time goes on, often theorems will be moved to
```
GinoGiotto, line 459226: Are removals of expired OLD proofs part of the usual maintenance edits? Or should we treat them differently if in a mathbox?

jkingdon: That would be like minimization - the same in or out of a mathbox.

GinoGiotto: I agree, I'll make a PR for their removal soon (unless someone else will do it before me).

wlammen (approved):
I think it is an improvement over the current situation.  There is nothing said about #3440 , though.  How can people mark stuff not quite matching the general spirit of this database as junk/fun/experimental etc., and how do we want to handle such contributions over time, especially when a mathbox appears to be abandoned.

set.mm
GinoGiotto: @tirix There are some OLD proofs in your mathbox missing an 'Obsolete as of ...' description, what do we do with those?

My proposal with OLD proofs would be to avoid exceptions: if they can be proven to be older than one year with the github history then even the ones missing a date should be removed as well. I think we should aim to be consistent with the meaning of OLD as "to be removed" (as the conventions page indicates), so if an author adds the OLD suffix to a label then he is sending the message that it can be deleted. If the intention is to keep an alternative version of a theorem, the ALT suffix is available for that purpose.

jkingdon: If they are missing an 'Obsolete as of' description, perhaps they should be renamed to something other than xxxOLD instead of removed. But I didn't look through them, so I don't have a theorem-by-theorem opinion about what to do about each one.

tirix: > @tirix There are some OLD proofs in your mathbox missing an 'Obsolete as of ...' description, what do we do with those?

Yes, I agree there should not be exceptions. It looks like I have taken the habit to append "OLD" to theorem for which I had a better version. But in many cases the formulas are different. Let's review those:
- `ogrpinvOLD` : this is superseded by `ogrpinv0le`, which is more powerful (equivalence, instead of just one direction). So it can be deleted.
- `xrge0tmdOLD`: proof has been shortened, but the proof method has changed. Maybe the old proof should be kept as 'ALT'.
- `isrnsigaOLD` : proof has been shortened, can be deleted.
- `measdivcstOLD` : this is a different formulation as `measdivcst`, not a drop-in replacement, and it is actually still used in `probmeasb`, so first the proof of `probmeasb` should be adapted to the new notation, then this version can be deleted. Or I just keep it with a different name.
- `probfinmeasbOLD` : same as `measdivcstOLD`: this is a different formulation as `measdivcst`, not a drop-in replacement.
- `axtglowdim2OLD` and `axtgupdim2OLD` : They are using the deprecated `TarskiG2D` notation which I chose to keep in my Mathbox. They have been superseded by the corresponding theorems with the `TarskiGDim>=` notation. I guess I was unsure of the future of the `TarskiGDim>=` notation. Maybe these should be renamed something different than with the `OLD` suffix.

I should do these housekeeping changes in my Mathbox, I'm not sure when I can come to it.
Are those inconvenient because of a tool or a plan to enforce specifics for the `OLD` theorems?

GinoGiotto: > I should do these housekeeping changes in my Mathbox, I'm not sure when I can come to it.

Thank you for your exaustive answer. If it's ok for you, I can do these maintenance edits following your instructions.

> Are those inconvenient because of a tool or a plan to enforce specifics for the OLD theorems?

I'm very much in favor of the idea of developing a tool that, given a database, can check all OLD statement dates and automatically generate a copy of the database without those OLD theorems. While I'm not planning on working on it anytime soon, I intend to enforce the updated conventions from this PR, so that there won't be any obstacles in case this tool will be implemented. Besides that, I simply believe that keeping the database clean helps to avoid messiness and prevents unnecessary memory usage.


I'm a bit busy these days, so I'll make the housekeeping PR next week or two.

tirix: > If it's ok for you, I can do these maintenance edits

Sure, thanks! And I would get to review them anyway.

PR #3473, opened 2023-09-10
Switch from +c to |_| from djulepw  to pwdjudom

jkingdon:
This is pretty similar to other pull requests in this series.

I guess the main thing which this one has is a number of cases where the use of `+c` can be removed, because we have enough of the `|_|` theorems and we're starting to get to theorems which use the disjoint union within the proof but not in the result.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 8-Sep-23 cdanum    djunum      Changes from +c notation to |_|
  7-Sep-23 sprmpt2d  sprmpod
  7-Sep-23 brovmpt2ex brovmpoex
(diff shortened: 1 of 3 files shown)


wlammen (approved):
All revised theorems seem now be free of +c.  I tried my best to check this.

jkingdon:
> All revised theorems seem now be free of +c. I tried my best to check this.

The discouragement tags should also help track this, but thanks for looking as well.

wlammen:
Couldn't it sneak in, as part of a used theorems proof not exposing this operation on the outside (hypothesis / conclusion)?

jkingdon:
> Couldn't it sneak in, as part of a used theorems proof not exposing this operation on the outside (hypothesis / conclusion)? 

I guess so (depending on various assumptions about whether I miss a theorem to mark and how we look at the diff to the discouraged file).

PR #3474, opened 2023-09-10
Add exnalimn

benjub:
* Add "exnalimn" (and minimize with it); this is the general instance that equs3OLD proves.
* Shorten abbii
* mathbox

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -75,19 +75,6 @@ jobs:
     if: ${{ needs.skip_dups.outputs.should_skip != 'true' }}
     steps:
       - uses: actions/checkout@v3
-      # Update first to counter problems installing texlive
-      - run: sudo apt-get update --fix-missing
-      # Install LaTeX so we can test generated LaTeX. We use extended packages:
-      # texlive-extra-utils provides pdflatex
-      # texlive-fonts-extra provides phonetic.sty
-      # texlive-science provides accents.sty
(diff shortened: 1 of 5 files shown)


icecream17 (approved):

set.mm
```diff
@@ -14853,6 +14853,12 @@ only postulated (that is, axiomatic) rule of inference of predicate
   alinexa $p |- ( A. x ( ph -> -. ps ) <-> -. E. x ( ph /\ ps ) ) $=
     ( wn wi wal wa wex imnang alnex bitri ) ABDECFABGZDCFLCHDABCILCJK $.
 
+  $( Existential quantification of a conjunction expressed with only primitive
+     symbols ( ` -> ` , ` -. ` , ` A. ` ).  (Contributed by NM, 10-May-1993.)
```
wlammen, line 14857: This theorem appears to be used by just 1 theorem, and it is a quite trivial modification of alinexa.  NM in the past rejected sort of convenience theorems by me, if they had not at least 7 applications.  This was about the average number of shortenings required to compensate for a new addition.  Why should we disregard this idea here?

benjub: This is the general instance of https://us.metamath.org/mpeuni/equs3OLD.html, which Norm added in 1993 (I kept the same date), and which I deprecated a few weeks ago. I should have added its general instance in the same PR as the deprecation but I postponed and then forgot. In addition to being its general instance, it is also elementary and natural, as its comment "Existential quantification of a conjunction expressed with only primitive symbols" expresses, so we are not in the situation of a lemma or convenience theorem, where indeed the number of usages matters.

wlammen (approved):

PR #3475, opened 2023-09-10
prove cbv2 without ax-10

wlammen:
1.  cbv2 can be proven without ax-10
3. remove a duplicated tag from equs3OLD

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14363,6 +14363,7 @@ New usage of "cases2ALT" is discouraged (0 uses).
 New usage of "cayleyhamiltonALT" is discouraged (0 uses).
 New usage of "cba" is discouraged (86 uses).
 New usage of "cbncms" is discouraged (5 uses).
+New usage of "cbv2OLD" is discouraged (0 uses).
 New usage of "cbvabvOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -20169,9 +20169,17 @@ theorem as an axiom of set theory (Axiom 0 of [Kunen] p. 10).  In the
     cbv2.5 $e |- ( ph -> ( x = y -> ( ps <-> ch ) ) ) $.
     $( Rule used to change bound variables, using implicit substitution.
        (Contributed by NM, 5-Aug-1993.)  (Revised by Mario Carneiro,
-       3-Oct-2016.)  Format hypotheses to common style.  (Revised by Wolf
-       Lammen, 13-May-2018.) $)
+       3-Oct-2016.)  Format hypotheses to common style, avoid ax-10.  (Revised
```
icecream17, line 20172: I guess one could link ax-10 (`ax-10.` --> ` ~ ax-10 .`)

icecream17 (approved):

set.mm
wlammen: done

benjub (approved):

PR #3477, opened 2023-09-11
Founded partial recursion

sctfn:
This PR finishes up a long term project of mine- recursion on a general founded predicate. It’s a standard result in set theory textbooks, but we’ve never had it. 

While I was at it, I also proved a variant I’ve never seen in a textbook- you don’t need Infinity if you have transitivity. 

Given these two, should we plan on a move to main soon?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -318,7 +318,6 @@
 "4syl" is used by "fin23lem22".
 "4syl" is used by "fmfnfmlem3".
 "4syl" is used by "fnwelem".
-"4syl" is used by "frrlem5".
 "4syl" is used by "fsumparts".
 "4syl" is used by "fta1".
 "4syl" is used by "fta1lem".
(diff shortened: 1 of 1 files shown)


tirix:
Welcome back! 
Looks great at first sight, I'll try to take more time to review in detail.

jkingdon (approved):
Putting this in a mathbox certainly looks good.

I don't really know enough to say one way or another about whether it belongs in main.

set.mm
```diff
@@ -512104,7 +512129,7 @@ Real and complex numbers (cont.)
 
 $(
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
-  Founded Recursion
+  Founded Parial Recursion
```
icecream17, line 512132:
```suggestion
  Founded Partial Recursion
```

icecream17, line 512132:
```suggestion
  Founded Partial Recursion
```

icecream17 (approved):
lgtm (oops accidentally deleted my previous comment) (other than typo)

set.mm
tirix: Fixed.

tirix (approved):

PR #3478, opened 2023-09-11
Add pwtrufal to iset.mm

jkingdon:
This is a way of stating that there are no truth values other than true and false.


benjub (approved):

digama0:
Actually, the proof I gave proves something slightly stronger than `-. -. ( x = (/) \/ x = { (/) } )`: it is most directly a proof of `x =/= { (/) } -> x = (/)`, from which you can get the double-negated or statement by propositional logic. Up to you whether you consider the stronger statement worth stating.

jkingdon:
> Actually, the proof I gave proves something slightly stronger than `-. -. ( x = (/) \/ x = { (/) } )`: it is most directly a proof of `x =/= { (/) } -> x = (/)`, from which you can get the double-negated or statement by propositional logic. Up to you whether you consider the stronger statement worth stating.

It seems a little harder to motivate/describe the stronger version but maybe I'll add it anyway.

I'm actively working on the follow up pull request (which was part of my reason for merging this one so quickly) so I can include tweaks there.

PR #3480, opened 2023-09-13
add solution to exercises regarding truth values in iset.mm

### LINKED ISSUE #3456: iset.mm problems with the power set of a singleton
The challenge is the two problems in https://mathstodon.xyz/@MartinEscardo/111002413035009813

jkingdon:
pwntru is as mentioned in https://github.com/metamath/set.mm/pull/3478#issuecomment-1714650955 and the other two are stated and discussed at #3456 

Fixes #3456


tirix (approved):

iset.mm
```diff
@@ -144493,6 +144493,72 @@ truth value (as seen in theorems like ~ exmidexmid ), then this theorem
...
+    $d A x $.
+    $( A slight strengthening of ~ pwtrufal .  (Contributed by Mario Carneiro
+       and Jim Kingdon, 12-Sep-2023.) $)
+    pwntru $p |- ( ( A C_ { (/) } /\ A =/= { (/) } ) -> A = (/) ) $=
+      ( vx c0 csn wss wne wa cv wcel wex simpr neneqd simpll simpl sselda elsni
+      wn wceq syl eqeltrrd snssd eqssd ex exlimdv mtod notm0 sylib ) ACDZEZAUHF
+      ZGZBHZAIZBJZQACRUKUNAUHRZUKAUHUIUJKLUKUMUOBUKUMUOUKUMGZAUHUIUJUMMUPCAUPUL
+      CAUPULUHIULCRUKAUHULUIUJNOULCPSUKUMKTUAUBUCUDUEBAUFUG $.
+  $}
+
```
benjub, line 144506: @jkingdon : since it is a strengthening of pwtrufal, then you can prove pwtrufal from it. E.g.,
```
      ( c0 csn wss wceq wn wa wo wne simpr neqned pwntru sylan2 simprl pm2.65da
      ioran sylnibr ) ABCZDZABEZFZAREZFZGZTUBHFSUDTUDSARITUDARUAUCJKALMSUAUCNOT
      UBPQ $.
```
(no need of a new tag IMO: date is close enough)
By the way: I think that both proofs would be shorter and simpler to follow if stating this theorem with `-. A = { (/) }` instead of `=/=`.

jkingdon: > since it is a strengthening of pwtrufal, then you can prove pwtrufal from it.

Of course we could. I chose this order because it seemed more appealing to have the interesting theorem followed by the "oh, by the way" one.

> both proofs would be shorter and simpler to follow if stating this theorem with `-. A = { (/) }` instead of `=/=`.

We have a relatively strong tendency to use `=/=`. There was at least one message from Norm about this. Personally I sort of like the parallel with apartness theorems, although I'm not sure how strongly I'd push that argument.

This proof would probably be better if it used `=/=` throughout the proof more, but I didn't spend much time looking around for where this might be possible or whether it would, well I guess shorten the proof is our usual yardstick.

benjub: > > since it is a strengthening of pwtrufal, then you can prove pwtrufal from it.

> Of course we could. I chose this order because it seemed more appealing to have the interesting theorem followed by the "oh, by the way" one.

This goes in the opposite direction of the proof factorization and proof shortening efforts made by most contributors for as long as I've been here, and encouraged in ~conventions.

PR #3482, opened 2023-09-13
Propose new names for Continuous function-builders

jkingdon:
A lot of these names use a single character "1" for single argument (`mpt`), "2" for two argument (`mpo`), and "k" for curried (here proposed to be `mpk`).

This is perhaps the most challenging part of the `mpt2` to `mpo` renames and to come up with good names involves browsing through the whole "Continuous function-builders" section (rather than just looking at labels containing `mpt2`). Therefore, I'm listing the proposed old and new names and will do the renames later if this is accepted.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -21,6 +21,28 @@ proposed  sylbird   biimtrrd    etc.
 proposed  syl5*     *trid       (syl5bi -> biimtrid; syl5eqel -> eqeltrid;etc.)
 proposed  syl6*     *trdi
 proposed  *mpt2*    *mpo*       (renaming process underway)
+proposed  cnmpt2k   cnmpompk
+proposed  cnmptk2   cnmpkmpo
+proposed  cnmptk1p  cnmpkmptp
+proposed  cnmptkk   cnmpkmpk
+proposed  cnmpt1k   cnmptmpk
+proposed  cnmptk1   cnmpkmpt
+proposed  cnmptcom  cnmpocom
+proposed  cnmpt2res cnmpores
+proposed  cnmpt1res cnmptres
+proposed  cnmpt22f  cnmpompof
+proposed  cnmpt22   cnmpompo
+proposed  cnmpt2t   cnmpot
+proposed  cnmpt21f  cnmpomptf
+proposed  cnmpt21   cnmpompt
+proposed  cnmpt2c   cnmpoc
+proposed  cnmpt2nd  cnmpo2nd
+proposed  cnmpt1st  cnmpo1st
+proposed  cnmpt12   cnmptmpo
+proposed  cnmpt12f  cnmptmpof
+proposed  cnmpt1t   cnmptt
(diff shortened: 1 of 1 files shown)


digama0:
My counter-proposal would be to not rename anything and just leave them with the existing convention. They aren't actually using `mpt` and `mpt2`, rather it is a series of theorems starting `cnmpt` and using a domain-specific indexing convention which uses sequences of `1`,`2`,`k` to identify argument shapes. I don't think that the `mpt2` -> `mpo` change needs to affect this section at all.

changes-set.txt
```diff
@@ -21,6 +21,28 @@ proposed  sylbird   biimtrrd    etc.
...
+proposed  cnmpt21f  cnmpomptf
+proposed  cnmpt21   cnmpompt
+proposed  cnmpt2c   cnmpoc
+proposed  cnmpt2nd  cnmpo2nd
+proposed  cnmpt1st  cnmpo1st
+proposed  cnmpt12   cnmptmpo
+proposed  cnmpt12f  cnmptmpof
+proposed  cnmpt1t   cnmptt
+proposed  cnmpt11f  cnmptmptf
+proposed  cnmpt11   cnmptmpt
```
icecream17, line 45:
```suggestion
proposed  cnmpt2k   cnmpok
proposed  cnmptk2   cnmpko
proposed  cnmptk1p  cnmpktp
proposed  cnmptkk   cnmpkk
proposed  cnmpt1k   cnmptk
proposed  cnmptk1   cnmpkt
proposed  cnmptcom  cnmpocom
proposed  cnmpt2res cnmpores
proposed  cnmpt1res cnmptres
proposed  cnmpt22f  cnmpoof
proposed  cnmpt22   cnmpoo
proposed  cnmpt2t   cnmpoot
proposed  cnmpt21f  cnmpotf
proposed  cnmpt21   cnmpot
proposed  cnmpt2c   cnmpoc
proposed  cnmpt2nd  cnmpo2nd
proposed  cnmpt1st  cnmpo1st
proposed  cnmpt12   cnmpto
proposed  cnmpt12f  cnmptof
proposed  cnmpt1t   cnmpttt
proposed  cnmpt11f  cnmpttf
proposed  cnmpt11   cnmptt
```
another potential idea that preserves the merging of the "prefix" mp
it's not as preserving though, since cnmpt11 collides with cnmpt1t (and cnmpt21 with cnmpt2t)

approve either way, I'm not an expert in this section

icecream17 (approved):

jkingdon:
> My counter-proposal would be to not rename anything and just leave them with the existing convention.

Sounds like this is the way to go.

It isn't an instance of `mpt2` so it is separate from that set of renames, and it isn't clear that there is an obviously better choice.

Unless there are objections, I'll close this pull request, tomorrow I guess.

jkingdon:
Not doing this, at least for now, maybe never.

PR #3489, opened 2023-09-18
Freshman's dream

tirix:
A proof of the ["freshman's dream"](https://en.wikipedia.org/wiki/Freshman%27s_dream) theorem.


icecream17 (approved):

avekens (approved):
nice result, great work!

PR #3490, opened 2023-09-19
rename mpt2->mpo from cnmpt2pc to mpt2cti

jkingdon:
This gets us partway into the mathboxes.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+18-Sep-23 clwwlknonmpt2 clwwlknonmpo
 18-Sep-23 cnmpt2pc  cnmpopc
 17-Sep-23 cdaun     endjudisj   Changes from +c notation to |_|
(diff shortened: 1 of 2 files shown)


wlammen (commented):
The changes are ok and consistent with previous ones.  Given the consent of the affected Mathbox owner this is an approval.

jkingdon:
> The changes are ok and consistent with previous ones. Given the consent of the affected Mathbox owner this is an approval.

Not sure that's needed for this kind of maintenance edit but in this case we can just ask @tirix whether there's any issue.

avekens (approved):
I think @tirix does not object, but he should have a chance to give his explicit approval.

tirix (approved):
No objection, we should keep the same conventions throughout, and Git does a good job at managing merges.

jkingdon:
> I think @tirix does not object, but he should have a chance to give his explicit approval.

Sounds good. I'll give people the chance, and if we don't hear from them (e.g. they aren't active)we'll worry about that then. For this reason, I'll break the remaining pull requests on mathbox boundaries.

PR #3494, opened 2023-09-20
Permutation cycles, and the alternating group

tirix:
Includes a definition for a builder for permutation cycles, as well as a few theorems about the alternating group.
Ultimately I'd like to prove that [the alternating group is generated by 3-cycles](https://en.wikipedia.org/wiki/Alternating_group#Generators_and_relations).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13507,6 +13507,7 @@ New usage of "al2imVD" is discouraged (0 uses).
 New usage of "alephf1ALT" is discouraged (0 uses).
 New usage of "alexsubALT" is discouraged (0 uses).
 New usage of "alrim3con13v" is discouraged (1 uses).
+New usage of "altgnsgALT" is discouraged (0 uses).
 New usage of "altgsumbcALT" is discouraged (0 uses).
 New usage of "amgmlemALT" is discouraged (0 uses).
 New usage of "anabss7p1" is discouraged (0 uses).
@@ -18239,6 +18240,7 @@ Proof modification of "al2imVD" is discouraged (48 steps).
 Proof modification of "alephf1ALT" is discouraged (47 steps).
 Proof modification of "alexsubALT" is discouraged (869 steps).
 Proof modification of "alrim3con13v" is discouraged (74 steps).
+Proof modification of "altgnsgALT" is discouraged (374 steps).
 Proof modification of "altgsumbcALT" is discouraged (1038 steps).
 Proof modification of "amgmlemALT" is discouraged (1054 steps).
 Proof modification of "anabss7p1" is discouraged (5 steps).


set.mm
```diff
@@ -466289,6 +466442,482 @@ real number multiplication operation (this has to be defined in the main
...
+  ctocyc $a class toCyc $.
+
+  ${
+    $d d w u x $.
+    $( Define a convenience permutation cycle builder.  Given a list of
+       elements to be cycled, in the form of a word, this function produces the
+       corresponding permutation cycle.  See definition in [Lang] p. 30.
+       (Contributed by Thierry Arnoux, 19-Sep-2023.) $)
+    df-tocyc $a |- toCyc = ( d e. _V |-> ( w e.
+      { u e. Word d | u : dom u -1-1-> d } |-> ( ( _I |` ( d \ ran w ) )
```
avekens, line 466463: Why is it important/necessary that the word is injective, i.e. must consist of different symbols?

set.mm
```diff
@@ -466289,6 +466442,482 @@ real number multiplication operation (this has to be defined in the main
...
+  $( Extend class notation with the permutation cycle builder. $)
+  ctocyc $a class toCyc $.
+
+  ${
+    $d d w u x $.
+    $( Define a convenience permutation cycle builder.  Given a list of
+       elements to be cycled, in the form of a word, this function produces the
+       corresponding permutation cycle.  See definition in [Lang] p. 30.
+       (Contributed by Thierry Arnoux, 19-Sep-2023.) $)
+    df-tocyc $a |- toCyc = ( d e. _V |-> ( w e.
```
avekens, line 466462: Is this definition really necessary? Why not use ` ( W cyclShift 1 ) ` (see ~cshfn)?
Why is it important/useful to keep the elements outside the word (which are not shifted), expressed by  ``( ( _I |` ( D \ ran W ) )``?

set.mm
```diff
@@ -465114,6 +465132,141 @@ its graph has a given second element (that is, function value).
...
+    wrdsplex $p |- ( ( W e. Word S /\ N e. ( 0 ... ( # ` W ) ) )
+      -> E. v e. Word S W = ( ( W |` ( 0 ..^ N ) ) ++ v ) ) $=
+      ( cword wcel chash cfv cop csubstr co cc0 cfz cfzo cres cconcat wceq wrex
+      cv cpfx swrdcl wa cuz simpr elfzuz2 eluzfz2 ccatpfx mpd3an3 pfxres oveq1d
+      3syl pfxid adantr 3eqtr3rd oveq2 rspceeqv syl2an2r ) DBEZFZDCDGHZIJKZURFC
+      LUTMKZFZDDLCNKOZVAPKZQDVDASZPKZQAURRBDCUTUAUSVCUBZDCTKZVAPKZDUTTKZVEDUSVC
+      UTVBFZVJVKQVHVCUTLUCHFVLUSVCUDCLUTUELUTUFUKBDCUTUGUHVHVIVDVAPBDCUIUJUSVKD
+      QVCBDULUMUNAVAURVGVEDVFVAVDPUOUPUQ $.
+
+    $( Obsolete proof of ~ wrdsplex as of 12-Oct-2022.  (Contributed by Thierry
```
avekens, line 465172: I think this OLD theorem can be deleted now.

set.mm
avekens: There isn't any application of this definition and its related basic theorems yet, so I cannot assess the need/usefulness of this definition.

tirix: Indeed, I oversaw `cyclShift` and I might have used it in the definition, however `( W cyclShift 1 )` lacks some properties needed for [cyclic permutations](https://en.wikipedia.org/wiki/Cyclic_permutation):
- as you point out, a permutation cycle only acts on _some_ of the elements of a possibly infinite set, while `cyclShift` shifts the _entire_ word, which has to be finite.
- a permutation is a function with identical domain and range, whereas the domain for words are always integer indices.

A version using `cyclShift` would look like this:
```
df-tocyc |- toCyc = ( d e. _V -> ( w e.
      { u e. Word d | u : dom u -1-1-> d } |-> ( ( _I |` ( d \ ran w ) )
        u. ( ( w cyclShift ( ( # ` w ) - 1 ) ) o. `' w ) ) } ) )
```
I might try to prove the equivalence.

set.mm
tirix: Yes!

set.mm
tirix: > There isn't any application of this definition

Hopefully with some patience the applications will come!

Like noted above I'd like to prove that [the alternating group is generated by 3-cycles](https://en.wikipedia.org/wiki/Alternating_group#Generators_and_relations), and there are more interesting theorems down the line. For example, the proof that alternating groups are simple for n>4 also uses 3-cycles. 

After @CatsAreFluffy's `ablsimpgd`, we almost have a proof that cyclic groups of prime order are simple.
Alternating groups would be the second [category of finite simple groups](https://en.wikipedia.org/wiki/List_of_finite_simple_groups) to the list... one step nearer to the famed monster group at the very bottom.

avekens: I am still not convinced that a new, separate definition is needed for this purpose. According to our conventions, we should introduce "New definitions infrequent." If special versions are needed (as the 1-1 property of the word function, or the range of the word function being embedded in a larger set), this can/should be expressed in the antecedents, e.g. `( W e. Word D /\ W : dom W -1-1-> D ) -> ( ( W cyclShift 1 ) ... `. That ` ran W C_ D ` and ` ( D \ ran W ) ` is not affected is clear anyway.

As long as this definition is in a mathbox only, I can wait for its application to see if it has an advantage big enough to justify a separate definition. For the moment, I would appreciate to have the definition based on cyclShift as proposed (the shorter version).

tirix: The definition using `cyclShift` is definitely better and there are several theorems which can be reused, for example [`~ cshf1`](https://us.metamath.org/mpeuni/cshf1.html) is exactly what I need, so I'll try to do the switch. It's still an important development so I'd prefer to do it in a different PR.

I believe I'm going to the right direction with this new definition but it might be interesting to hear other voices.

This will certainly stay in my mathbox for very long and I have no issue with that.

avekens: > The definition using `cyclShift` is definitely better and there are several theorems which can be reused, for example [`~ cshf1`](https://us.metamath.org/mpeuni/cshf1.html) is exactly what I need, so I'll try to do the switch. It's still an important development so I'd prefer to do it in a different PR.

Under this assumption I can approve this PR.

avekens (approved):

PR #3498, opened 2023-09-21
Simple groups of prime order or the cyclic groups

tirix:
This is a very simple new theorem based on @CatsAreFluffy's work on simple groups, which I've mentioned in [a comment](https://github.com/metamath/set.mm/pull/3494#discussion_r1332193406) in #3494.

I believe this work belongs to the main part of set.mm. Since this is a single theorem, I've only been adding it to @CatsAreFluffy's mathbox for now.

What is everyone's opinion about moving the "simple group" section to main?


icecream17 (approved):
The simple group sections can be moved to main (in general group theory somehow connects everything together)

Note that https://us.metamath.org/mpeuni/ablsimpgprmd.html has a typo in the contributor name (edit: fixed in this pr)

set.mm
```diff
@@ -647997,6 +647997,18 @@ elements and the subgroup containing only the identity ( ~ simpgnsgbid ).
       AUAMRS $.
   $}
 
+  ${
+    prmsimpcyc.1 $e |- B = ( Base ` G ) $.
+    $( The simple groups of prime order are exactly the cyclic groups.  This is
```
benjub, line 648002: False: the cyclic group ZZ/6ZZ is not simple (nor of prime order) !  I think you meant: a group of prime order is cyclic if and only if it is simple. (The statement is slighly more general because of the extensible structure machinery, but that is unimportant.)

benjub:
> What is everyone's opinion about moving the "simple group" section to main?

I would say yes. As always, a bloc-move to Main requires special scrutiny, since the threshold to make it into a mathbox is lower than that to make it into Main.

set.mm
tirix: Thanks, fixed!

avekens (approved):

avekens:
> > What is everyone's opinion about moving the "simple group" section to main?
> 
> I would say yes. As always, a bloc-move to Main requires special scrutiny, since the threshold to make it into a mathbox is lower than that to make it into Main.

I also endorse to move the "simple group" section to main. It should be done in a separate PR, which can be reviewed carefully as proposed by @benjub before it is merged.

set.mm
savask: The statement is very weird, a group of prime order is always simple, so instead of
``|- ( ( # ` B ) e. Prime -> ( G e. SimpGrp <-> G e. CycGrp ) )``
one can simply state
``|- ( ( # ` B ) e. Prime -> G e. CycGrp )``

Am I missing anything obvious?

tirix: > one can simply state
> ``|- ( ( # ` B ) e. Prime -> G e. CycGrp )``

That's already [`prmcyg`](https://us.metamath.org/mpeuni/prmcyg.html) actually. 
Granted, this is a very small addition to @CatsAreFluffy's work, only adding `prmcyg`... just a pretext to move the "Simple Group" section to main. :-)

savask: > That's already [prmcyg](https://us.metamath.org/mpeuni/prmcyg.html) actually.

Oh, indeed. In that case I think your theorem is completely useless, sorry! I can't imagine where it can be used, since we have [prmgrpsimpgd](https://us.metamath.org/mpeuni/prmgrpsimpgd.html) and [prmcyg](https://us.metamath.org/mpeuni/prmcyg.html).

I like the idea of moving the section about simple groups into main, but maybe `prmsimpcyc` can stay in your mathbox then.

tirix: > I think your theorem is completely useless, sorry! I can't imagine where it can be used, since we have [prmgrpsimpgd](https://us.metamath.org/mpeuni/prmgrpsimpgd.html) and [prmcyg](https://us.metamath.org/mpeuni/prmcyg.html).

Aha! Well, that's all it does : combine [prmgrpsimpgd](https://us.metamath.org/mpeuni/prmgrpsimpgd.html) and [prmcyg](https://us.metamath.org/mpeuni/prmcyg.html).

I can keep it in my mathbox, no problem.

Maybe I will merge this PR and launch another with all these changes ("Simple Groups" to main, `prmsimpcyc` to my mathbox)

Issue #3485, opened 2023-09-16
Definitions of uniqueness and existential uniqueness


benjub:
As the one-year period for removal of obsolete (*.OLD) statements is approaching for many statements about unique existence and uniqueness, I think we should postpone there deletion since I'm having terrible doubts on whether the definition change I proposed last year (effective 30-Sep-2022) was a good idea. (I still think the splitting into two subsections is better and make things clearer, so that it was a good decision.)

As for the definitions themselves, it seemed more natural to define unique existence as the conjunction of two more elementary notions, existence and uniqueness. That is,
```
df-mo $a |- ( E* x ph <-> E. y A. x ( ph -> x = y ) ) $.
df-eu $a |- ( E! x ph <-> ( E. x ph /\ E* x ph ) ) $.
```
(this is the current situation), compared to the older definitions, which are kept as
```
eu6 $p |- ( E! x ph <-> E. y A. x ( ph <-> x = y ) ) $=
moeu $p |- ( E* x ph <-> ( E. x ph -> E! x ph ) ) $=
```

One problem is that `df-mo` is not correct on empty domains, since there, it is false (since it begins with an existential quantifier), but uniqueness should be considered to tautologically hold on the empty domain.  I wrote a definition which would work,
```
df-bj-mo $a |- ( E** x ph <-> A. z E. y A. x ( ph -> x = y ) ) $.
```
preceded in my mathbox by justification theorems on a very weak axiomatic. But having an extra quantifier may be a bit cumbersome, and I'm not sure yet it would be good to adopt it.  One would need to check the respective strengths of these definitions on that weak axiomatic (basically, no existence axiom, so ax-1--ax-5 with `$a |- A. x E. y y = x $.` (**edit:** and the fact that equality is an equivalence relation, or at least symmetric, as in `bj-cbval`)), and also compare it to the more intuitive
```
$a |- ( E*** x ph <-> A. y A. z ( ( [ y / x ] ph /\ [ z / x ] ph ) -> y = z ) ) $.
```

An associated problem is that the current definitions that I proposed bring maybe more auxiliary axioms (ax-10--ax-13) than necessary.

I apologize for not looking into all the possible consequences of this definition change before proposing it.  Fortunately, all theorems needed to go back and forth between the two definition pairs, and to recover them from each other, are still there, so reverting would be rather easy. I should thank Gérard Lang whose questions on the empty domain problem, which looked small at first, made me realize there might be a bigger problem with the new definitions.

[...]

PR #3500, opened 2023-09-21
Misc (dfich2ai is dfich2, reduce axioms: nfsab1, ineq1, dfmo)

icecream17:
Several misc changes, see each commit

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15410,6 +15410,7 @@ New usage of "erngring-rN" is discouraged (0 uses).
 New usage of "erngset-rN" is discouraged (3 uses).
 New usage of "eu1OLD" is discouraged (0 uses).
 New usage of "eu6OLD" is discouraged (0 uses).
+New usage of "eu6OLDOLD" is discouraged (0 uses).
 New usage of "euaeOLD" is discouraged (0 uses).
 New usage of "euanvOLD" is discouraged (0 uses).
 New usage of "eubiOLD" is discouraged (0 uses).
@@ -18675,7 +18676,7 @@ Proof modification of "dfich2OLD" is discouraged (134 steps).
 Proof modification of "dfich2ai" is discouraged (140 steps).
 Proof modification of "dfich2bi" is discouraged (121 steps).
 Proof modification of "dfiun2gOLD" is discouraged (134 steps).
-Proof modification of "dfmo" is discouraged (138 steps).
+Proof modification of "dfmo" is discouraged (44 steps).
 Proof modification of "dfnul2OLD" is discouraged (44 steps).
 Proof modification of "dfsb2ALT" is discouraged (79 steps).
 Proof modification of "dfsb3ALT" is discouraged (39 steps).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -37443,8 +37450,15 @@ technically classes despite morally (and provably) being sets, like ` 1 `
   ${
     $d x A $.  $d x B $.  $d x C $.
     $( Equality theorem for intersection of two classes.  (Contributed by NM,
-       14-Dec-1993.) $)
+       14-Dec-1993.)  (Proof shortened by SN, 20-Sep-2023.) $)
```
wlammen: This also removes ax-10, ax-11, ax-12

set.mm
```diff
@@ -22680,23 +22693,37 @@ of the unique existential quantifier (note that ` y ` and ` z ` need not
        using the at-most-one quantifier.  (Contributed by NM, 12-Aug-1993.)
        This used to be the definition of the unique existential quantifier,
        while ~ df-eu was then proved as ~ dfeu .  (Revised by BJ, 30-Sep-2022.)
-       (Proof shortened by Wolf Lammen, 3-Jan-2023.) $)
+       (Proof shortened by Wolf Lammen, 3-Jan-2023.)  Remove use of ~ ax-11 .
```
wlammen, line 22696: Wasn't exsb provable without ax-11 as well? See #3485 .

wlammen (approved):
The annotations are non-blocking.  I think @avekens  does not object to your modifications in his mathbox.

set.mm
icecream17: Hmm, I didn't prove that exsb didn't need ax-11, but rather that exsb and dfmolem are similar (but dfmolem doesn't need ax-11)

wlammen: Ok then.  I remembered it differently

set.mm
wlammen: See comment to the second commit.

set.mm
```diff
@@ -621580,6 +621617,15 @@ fixed reference functional determined by this vector (corresponding to
       ( wi wo jao syl6c ) ABCDHEDHCEIDHFGCDEJK $.
   $}
 
+  ${
+    $d x y $.  $d x ph $.
+    spimevw.1 $e |- ( x = y -> ( ph -> ps ) ) $.
+    $( Version of ~ spimefv with ` x ` , ` ph ` disjoint, using only Tarski's
+       FOL axiom schemes.  (Contributed by SN, 21-Sep-2023.) $)
+    spimevw $p |- ( ph -> E. x ps ) $=
+      ( ax-5 spimeh ) ABCDACFEG $.
```
benjub, line 621626: That's bj-spimevw. No worries for the tag, but can you add to the comment
> Existential introduction, using implicit substitution. This is to spimeh what spimvw is to spimw.

and then deleted the one in my mathbox.

set.mm
```diff
@@ -621662,6 +621718,15 @@ fixed reference functional determined by this vector (corresponding to
     sn-el $p |- E. y E. x x e. y $=
       ( vw vz wel wi wal wex ax-pow weq ax6ev ax9v1 alrimiv eximii exim mpi ) C
       AECDEFZCGZABEZFAGZSAHZBDBACITRAHUAADJZRAADKUBQCADCLMNRSAOPN $.
+
+    $( ~ dtru without ~ ax-8 or ~ ax-12 .  (Contributed by SN, 21-Sep-2023.) $)
+    sn-dtru $p |- -. A. x x = y $=
```
benjub, line 621723: Better than bj-dtru !

discouraged
avekens, line 1: I have split the proof in two parts, because one direction was proven by @wlammen (~dfich2bi , see  also ~ wl-ich ).  If the theorems are merged, then there is no relation to these theorems (wl-ich*) in WL's mathbox anymore. @wlammen, could yo add on occasion a hint in the comments of your theorem(s) (at least for ~wl-ich) to ~dfich2 then?

wlammen: Thanks @avekens .  I'll remove the wl-ich* altogether.  They were meant to give a head start.  If they are not needed any longer, fine, no problem.

avekens: > Thanks @avekens . I'll remove the wl-ich* altogether. They were meant to give a head start. If they are not needed any longer, fine, no problem.

Great, under this assumption I can approve this PR.

avekens (approved):

PR #3502, opened 2023-09-22
delete wl-ich*

wlammen:
1. as announced in #3500 all wl-ich* theorems and their references are removed.  These theorems were introduced to initiate interchangeability a bit, but seem to be not up to the current development.
2. Fix a link in sb7f and sb7h.  The old definition of substitution dfsb1 is actually meant in the comment.


avekens (approved):
I find the lemmata quite interesting. I think I will transform them using the definition of interchangeability, rename them and put them into my mathbox. The contributer will still be @wlammen of course.

tirix (approved):

PR #3503, opened 2023-09-22
Prove exdifsn

BTernaryTau:
> There exists an element in a class excluding a singleton if and only if there exists an element in the original class not equal to the singleton element.

Similar to eldifsn and rexdifsn.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -4748,6 +4748,8 @@
 "dfhnorm2" is used by "hilnormi".
 "dfhnorm2" is used by "normf".
 "dfhnorm2" is used by "normval".
+"dfich2ai" is used by "dfich2OLD".
+"dfich2bi" is used by "dfich2OLD".
 "dfid2" is used by "fsplit".
 "dfid3" is used by "dfid2".
 "dfiop2" is used by "hoico1".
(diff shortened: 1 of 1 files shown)


avekens:
Thank you @BTernaryTau , that you are still willing to contribute to Metamath. According our conventions, such a theorem (which is not used in proofs of other theorems), it should be placed within a Mathbox. There was a Mathbox for you some time ago, but unfortunately it was removed. 

It would be great if you can create it again (just add the following comment just before "Mathbox for Mario Carneiro", because the mathboxes are ordered alphabetically) and move your theorem into it:

```
$(
  Mathbox for BTernaryTau
$)
```

jkingdon:
Not sure I can give a hard and fast rule for when something should go into a mathbox (until there is a reason to move it out) and when something can be added to main initially, but I guess this one feels kinda like a mathbox one. Perhaps one way to describe it is that putting something into a mathbox is always a safe course of action (much less need to ponder/debate whether it is useful) and if someone wants to use a theorem in a mathbox, they'll move it out of the mathbox then.

BTernaryTau:
So for future reference, would it be fair to say that whether or not a theorem that isn't used in a proof belongs in a mathbox, depends on whether or not it's significant enough to stand on its own (where significance is the subjective part that precludes a hard rule)?

avekens (approved):

PR #3512, opened 2023-09-24
Rename mpt2bi123f to mpobi123f in set.mm

jkingdon:
Will seek approval from @giomasce as we are doing for the mathbox portion of this series of edits. The background is #3354 in case you haven't been following this.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+23-Sep-23 mpt2bi123f mpobi123f
 23-Sep-23 wl-naev   naev        moved from WL's mathbox to main set.mm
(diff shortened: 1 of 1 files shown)


benjub (approved):

giomasce:
Approved, no problem. In general I'm very flexible on theorem names (and, accordingly, theorems in my mathbox are easily named somewhat randomly). I'm not sure if there is any formal step I should follow for approving (I see no "Approve" button in the GH web interface, as I am used with GitLab).

benjub:
> I'm not sure if there is any formal step I should follow for approving (I see no "Approve" button in the GH web interface, as I am used with GitLab).

In the tab "Files changed", there is a "Review changes" button in the upper-right corner. (But in the present case, your approval above is enough.)

PR #3518, opened 2023-09-25
Rewrite the permutation cycle definition with cyclic shifts on words

tirix:
As discussed in #3494, this changes the definition of the permutation cycle to use cyclic shifts on words, `cyclShift` instead of a maps-to notation.

Except for one theorem `0un` moved from Glauco's mathbox to main, all changes are in my mathbox.

Only a few new utility theorem are added compared to the previous version. There is no additional theory on the cycles and permutation, so there is no new justification for this new definition (yet!).

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+24-Sep-23 0un       [same]      moved from GS's mathbox to main set.mm
 24-Sep-23 elrnmpt2id elrnmpoid
 24-Sep-23 cbvmpt21  cbvmpo1
 24-Sep-23 cbvmpt22  cbvmpo2
--- a/discouraged
+++ b/discouraged
@@ -12695,7 +12695,6 @@
 "swrd0valOLD" is used by "swrd0lenOLD".
 "swrd0valOLD" is used by "swrdccat1OLD".
 "swrd0valOLD" is used by "wlkreslem0OLDOLD".
-"swrd0valOLD" is used by "wrdsplexOLD".
 "swrd0valOLD" is used by "wwlksm1edgOLD".
 "swrdccat1OLD" is used by "ccatopthOLD".
 "swrdccat1OLD" is used by "clwwlkfoOLD".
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -466638,202 +466789,147 @@ real number multiplication operation (this has to be defined in the main
...
+        wa 3syl incom disjdif eqtr3i f1oun syl22anc tocycfv uncom frnd syl5reqr
+        wss undif cvv wb fvex elsymgbas2 ax-mp sylibr ) ACCFBMZNZXFDUAMZOZAXGCF
+        UBZUCZXJPZXLUHXKUDZFQUEUIZFUFZUGZPZNZAXKXKXMNZXJXJXPNZXKXJUJZUKRZYBXRXS
+        AXKULUMAFUNZXJXNNZXJYCXONZXTAYDUOFUPMUQUIZXNUBZXNNZAXNYFVDXNUFURZYHAYFC
+        XNAFCUSOZQUTOZYFCXNSIAVAZCQFVBTVCAYJXOURZYKYIIAYCCFSZYMAYCCFVEZYNYMWGJY
+        CCFVFVGVHYLYJYMYKVIXNXNRYIXNVJCQFXNVKVPVLYFXNVMVNAYCYFXJYGXNXNAXNVOAYFC
+        FAYJYFCFSICFVQVRZVSAYGXJAYJYKYGXJRIYLQCFVTTWAWBWCAYOYCXJFNYEJYCCFWDYCXJ
+        FWEWHXJYCXJXNXOWFTYBAXJXKUJYAUKXJXKWIXJCWJWKUMZYQXKXKXJXJXMXPWLWMACXLCX
+        LXFXQABCEFGHIJWNAXLXJXKPZCXJXKWOAXJCWRYRCRAYFCFYPWPXJCWSVGWQZYSWBWCXFWT
+        OXIXGXAFBXBCXHXFDWTKXHVJXCXDXE $.
```
avekens, line 466931: Was this extreme shortening of th proof because of the changed definition/usage of cyclShift ?This would be another justification for the new definition.

avekens (approved):

set.mm
tirix: I would say it's because of the several utility theorems I proved (everything in `prop` and `snop`, `pfx1s2`, `cshwrnid` and `cshw1s2`), so it was rather broken down into smaller pieces. Which is also good, because:
- more, shorter theorems are easier to read,
- those shorter theorems are possibly reusable.

Also, the definition itself had one less bound variable, which I prefer.
So still overall it's clearly a plus.

icecream17 (approved):

PR #3522, opened 2023-09-26
Disable metamath-knife caching

jkingdon:
The problem is that the metamath-knife cache key does not include anything about the git revision of metamath-knife.  Until we can fix that, this cache will cause different test runs to get different versions of metamath-knife.

I noticed this in investigating #3521

### DIFF
--- a/.github/workflows/verifiers.yml
+++ b/.github/workflows/verifiers.yml
@@ -236,7 +236,9 @@ jobs:
         with:
           rust-version: stable
       - name: Install metamath-knife
-        if: ${{ !steps.cache-metamath-knife.outputs.cache-hit }}
+        # Turning off this caching until we can include the git
+        # revision of metamath-knife in the cache key
+        #if: ${{ !steps.cache-metamath-knife.outputs.cache-hit }}
         run: |
           rm -rf metamath-knife &&
           git clone --depth 1 https://github.com/metamath/metamath-knife.git &&


jkingdon:
Although turning off the cache was certainly helpful in debugging why we were or were not seeing metamath-knife related failures, I'm beginning to wonder whether the predictability (about what metamath-knife version we get) is worth the 2 minutes per build.

I still do like the idea of putting git metamath-knife git revision in the cache key, but that isn't this pull request.

digama0:
What *is* in the cache key? Is it just cached forever?

jkingdon:
> What _is_ in the cache key? Is it just cached forever?

At the moment, yes forever (well or however long github actions keeps anything, I'm not sure how predictable that part is). Which wasn't a big deal for smetamath-rs which has been using this caching code for a long time.

digama0:
I think a better solution to this issue (which we may do as part of the install.sh rewrite metamath/metamath-website-scripts#16) is to have set.mm CI download metamath-knife from us.metamath.org, and for the website to recompile metamath-knife daily along with the website build. That way set.mm gets a 1-day-old version of metamath-knife (among other dependencies), which seems sensible as a cache frequency.

digama0:
Note: metamath-knife caching can be managed via https://github.com/metamath/set.mm/actions/caches (not sure what permissions are required to delete caches there, I would assume you can). I think that is sufficient as a solution for now, just make sure to delete the `linux-build-cache-knife` cache if you are debugging an issue that depends on the version of metamath-knife.

jkingdon:
I guess putting a binary on the web site might work, if there is binary compatibility between where we build the website and github actions, but I don't know, just keeping the caching the way it is and managing it at https://github.com/metamath/set.mm/actions/caches might be a good enough answer (I do have the ability to delete entries there, so I suppose it is anyone with write access).

digama0:
I think we want to serve compiled tools from the metamath website anyway, just like we do for metamath.exe and mmj2. It's just a question of whether we want to also use those builds in CI. (And yes, they will be binary compatible, pretty much all linux-ish systems in the past 20 years use the same executable and OS interface, namely ELF / POSIX.)

PR #3524, opened 2023-09-27
move mmrecent.html from seed repo

digama0:
This already seems to be the intent given https://github.com/metamath/metamath-website-scripts/blob/6e59a5da3dde1df34ccbd62373f572bac07819d3/regenerate-website.sh#L104 : this line is currently failing because the file is not present in set.mm repo. It still won't be present after this PR because I have put a `.raw.html` in the name to make it clear that this is only a prototype file which lacks the automatically generated content. (`mmbiblio.raw.html` is also failing for the same reason.)

We could leave this in the metamath-website-seed repo (indeed, all of the html files could go in there), but this repo is higher traffic, so it makes more sense to put "recent news" here. Plus, maybe this will encourage us to update it, it has been left unchanged since Norm's passing and I think we can find some events of note in the past two years to put in there.


benjub:
Furthermore, `mmrecent*` files are associated with a given database, which is one more reason to move them here. If there were a general, "metamath-wide" `mmrecent`, this would probably go to the website repo.

digama0:
Keep in mind that the -seed repo already has a significant number of files "associated with a given database" (see [mpegif](https://github.com/metamath/metamath-website-seed/tree/rm_symbols/mpegif), [qlegif](https://github.com/metamath/metamath-website-seed/tree/rm_symbols/qlegif), [ilegif](https://github.com/metamath/metamath-website-seed/tree/rm_symbols/ilegif); these are *after* all of my simplifications and deletions, and the main file `mmset.html` / `mmil.html` is missing because it's in the set.mm repo instead). Not to mention that the top level site layout itself needs to link to them. The website is very far from being database-agnostic, you have to copy and paste things in 5 different places if you want to add a new database. As I mentioned on the group, I think we should move everything to the website except the .mm files themselves.

benjub:
> [mpegif](https://github.com/metamath/metamath-website-seed/tree/rm_symbols/mpegif), [qlegif](https://github.com/metamath/metamath-website-seed/tree/rm_symbols/qlegif), [ilegif](https://github.com/metamath/metamath-website-seed/tree/rm_symbols/ilegif)

At least, these will disappear when we drop GIF support.

As for the rest, I haven't looked enough at the architecture and website generation to have an informed opinion.

digama0:
No, they are the same for GIF and unicode, it's a copy of the directory so both versions have the same files. The only difference between GIF and unicode directories is in the files that are *not* in the seed version of the directory.

benjub:
> No, they are the same for GIF and unicode, it's a copy of the directory so both versions have the same files. The only difference between GIF and unicode directories is in the files that are _not_ in the seed version of the directory.

Ha !  So not only is the architecture complex, but it also uses obfuscation in file naming !  I'll wait for your simplifications to take effect to try and understand it.

jkingdon:
This is a bit tricky to review since there are several changes here, but here's what I see:

1. removing index.html. This is a good idea per https://github.com/metamath/metamath-website-seed/pull/11#issuecomment-1639025814
2. removing mm-mailing-list.html.  Seems like a good idea for the same reason, and hasn't gotten any recent edits.
3. removing mmmusic.html. I guess the scripts will use https://github.com/metamath/metamath-website-seed/blob/main/mpegif/mmmusic.html
4. adding mmrecent.raw.html. I guess this is good and a followup task can be to remove https://github.com/metamath/metamath-website-seed/blob/main/mpegif/mmrecent.html and update the scripts to know about the `raw`
5. renaming mmrecent_IL.html to mmrecent_IL.raw.html. OK, with similar comments about https://github.com/metamath/metamath-website-seed/blob/main/ilegif/mmrecent_IL.html
6. removing mmtopstr.html. I guess the scripts will use https://github.com/metamath/metamath-website-seed/blob/main/mpegif/mmtopstr.html
7. removing symbols.html.  This is supposed to end up at https://us.metamath.org/symbols/symbols.html . I guess the scripts will use https://github.com/metamath/metamath-website-seed/blob/main/symbols/symbols.html
8. adding `mv mmbiblio.raw.html mmbiblio.html || true` to verifiers.yml. I guess this is fine.

jkingdon (approved):

benjub:
> 8\. adding `mv mmbiblio.raw.html mmbiblio.html || true` to verifiers.yml. I guess this is fine.

The `|| true` amounts to ignoring a potential error of `mv` ? Shouldn't we want, on the contrary, to stop on error, for instance if there is no `mmbiblio.raw.html` ? (actually, we would also need `set -e` I guess ?)

digama0:
If no one objects I will merge all the PRs in this series in about 24 hours, unless there are additional concerns brought up (please speak up if you think something is out of place).

digama0:
> > 8. adding `mv mmbiblio.raw.html mmbiblio.html || true` to verifiers.yml. I guess this is fine.
> 
> The `|| true` amounts to ignoring a potential error of `mv` ? Shouldn't we want, on the contrary, to stop on error, for instance if there is no `mmbiblio.raw.html` ? (actually, we would also need `set -e` I guess ?)

The reason for the `|| true` is so that the command is idempotent.

PR #3525, opened 2023-09-27
The alternating group is generated by 3-cycles

tirix:
Follow-up for #3494 and #3518.

The main result is that the alternating group is generated by 3-cycles. This is proved using induction on words of even length (special theorem for that, I'm unsure whether it will have other uses).
Four theorems moved from GS's mathbox to the main part of set.mm.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,10 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+24-Sep-23 nelpr1    [same]      moved from GS's mathbox to main set.mm
+24-Sep-23 nelpr2    [same]      moved from GS's mathbox to main set.mm
+24-Sep-23 breldmd   [same]      moved from GS's mathbox to main set.mm
+24-Sep-23 fvelimad  [same]      moved from GS's mathbox to main set.mm
 24-Sep-23 0un       [same]      moved from GS's mathbox to main set.mm
 24-Sep-23 ovmpt2x2  ovmpox2
 24-Sep-23 ovmpt2rdx ovmpordx


set.mm
```diff
@@ -460080,6 +460128,40 @@ Class abstractions (a.k.a. class builders)
...
+      sylancl eqeltrd exlimddv ) BAEZCBFZBGHIZUAZBCJZUBZDVCZJZKZWDLZAFDWBWDMHIW
+      GDUCWBWDWDNOZMHWBWIWDWBWDSWBBSFZWDSFWAVSWJVTWAWJGSFZGUDFZWKPGUEUFBGUGUHQZ
+      BWCUIUJUKULWBWITZMTZKWIMKWBBNOZWNWOWBWJVTWPWNKWMVSVTWAUMBCUNUOWAVSWPWOKVT
+      WAWPGWOWAWLWPGKPBGUPUQURRQUSWIMUTVAVBDWDVDVAWBWGVEZWHWEAWQWHWFLWEWQWDWFWB
+      WGVFZVGWEDVHVIRWQWFAEWEWFFWEAFWQWFBAWQWFWDBWRWQBWCVJVKVSVTWAWGVLVMDVNWFAW
+      EVOVPVQVR $.
+
+    $( The other element of a pair is not the known element.  (Contributed by
+       Thierry Arnoux, 26-Aug-2017.) $)
+    unidifsneq $p |- ( ( X e. P /\ P ~~ 2o ) -> U. ( P \ { X } ) =/= X ) $=
```
icecream17, line 460150: maybe
```suggestion
    unidifsnne $p |- ( ( X e. P /\ P ~~ 2o ) -> U. ( P \ { X } ) =/= X ) $=
```

icecream17 (approved):

set.mm
tirix: Yes indeed! Change done.

savask:
A cool result! Just a few comments:

1. One in principle can remove `3 <_ N` from the assumptions of `cyc3genpm` since Alt(1) and Alt(2) are trivial and hence generated by an empty set.
2. I think alternating groups deserve their own notation.
3. I also think "a subgroup generated by a set" deserves its own notation as well.

tirix:
> 1. One in principle can remove `3 <_ N` from the assumptions of `cyc3genpm` since Alt(1) and Alt(2) are trivial and hence generated by an empty set.

Yes! I might try to remove that condition.

> 2. I think alternating groups deserve their own notation.

There is already `pmEven` which I'm using. `` ( pmEven ` D ) `` is the set of even permutation on the set `D`, i.e. _Alt(n)_ if `D` has _n_ elements. Note that this is just a set, it does not have the group structure, but it is easy to inherit from. 
The naming is not great, though. I would also prefer something that would sound more like "alternating group".

> 3. I also think "a subgroup generated by a set" deserves its own notation as well.

Yes, possibly. For the notation, I've got inspiration from [`psgnfitr`](https://us.metamath.org/mpeuni/psgnfitr.html): no special notation.
I think those are the only two examples for now?

Next up is conjugates, I have [`sylow2b`](https://us.metamath.org/mpeuni/sylow2b.html) as a model for the statement, which also does not have a special notation.

savask:
>  I think those are the only two examples for now?

I guess there are also theorems about cyclic groups like [ablsimpg1gend](https://us.metamath.org/mpeuni/ablsimpg1gend.html), but these don't need words to work with generators.

set.mm
```diff
@@ -460080,6 +460128,40 @@ Class abstractions (a.k.a. class builders)
       EBCRUIUOUEUBUFMUG $.
   $}
 
+  ${
+    $d A x $.  $d P x $.  $d X x $.
+    $( The other element of a pair is an element of the pair.  (Contributed by
+       Thierry Arnoux, 26-Aug-2017.) $)
+    unidifsnel $p |- ( ( P C_ A /\ X e. P /\ P ~~ 2o )
+                  -> U. ( P \ { X } ) e. A ) $=
```
avekens, line 460136: Interesting: I was not aware of the idiom ` U. ( P \ { X } ) ` for "the other element" of a pair ` P ` until now. Unfortunately, this is nowhere explicitly explained in set.mm, only implicitly by ~en2eleq.

avekens (approved):

set.mm
tirix: This is thanks to Stefan O'Rear I believe. Very clever!
I've made a new version which I think is slightly simpler, so possibly more reusable.

tirix:
> I might try to remove the N≥3 condition.

Done in the last commit!

Issue #3492, opened 2023-09-19
How to replace +c in xps?


jkingdon:
The project to replace `+c` with `|_|` is mostly done but the remaining piece is that https://us.metamath.org/mpeuni/df-xps.html and related theorems contain `` `' ( { x } +c { y } )``. This is equal to ` { <. (/) , x >. , <. 1o , y >. } ` by https://us.metamath.org/mpeuni/xpscg.html which makes me ask myself: why not just replace it with ` { <. (/) , x >. , <. 1o , y >. } ` rather than ` ( { x } |_| { y } )` ? The former is almost as short and more directly expresses the idea that this is a function which maps ` (/) ` to `x` and `1o` to `y`. And there seemed to be a bit of discomfort with using disjoint union to make a function in this way in comments like https://github.com/metamath/set.mm/pull/3406#issuecomment-1685315093 (and some others on that pull request).

Since there are quite a few df-xps related theorems I thought it best to ask in an issue before I start in on a big way. But as a proof of concept I did prove the following replacement for https://us.metamath.org/mpeuni/xpsfrnel2.html :

```
  ${
    $d A k $.  $d B k $.  $d S k $.  $d T k $.
    xpsfrnel3 $p |- ( { <. (/) , A >. , <. 1o , B >. }
        e. X_ k e. 2o if ( k = (/) , S , T )
        <-> ( A e. S /\ B e. T ) ) $=
      ( c0 cop c1o cpr c2o wceq wcel wa cvv csn cdm 0ex sylbi 3syl 1oex cv cixp
      cif cfv wfn xpsfrnel simp2bi wb ixpfn prid1 df2o3 eleqtrri fndm syl5eleqr
      cun df-pr dmeqi dmun eqtri syl6eleq elun wbr wex eldm df-br opex elsn vex
      wo opth 3bitri simprbi syl6eqelr exlimiv dmsnopss sseli elsni 1n0 pm2.21i
      neii eqcoms jaoi syl wne necomi fvpr1g mp3an13 eleq1d mpbid simp3bi prid2
      fvpr2g jca wf a1i simpl simpr fprg syl221anc sylibr ffnd mpbird syl3anbrc
      feq2i impbii ) FAGZHBGZIZEJEUAZFKCDUCZUBLZACLZBDLZMZXKXLXMXKFXHUDZCLZXLXK
[...]

PR #3532, opened 2023-09-29
convert xps theorems from +c through xpsbas

jkingdon:
As with most of the `+c` pull requests this one proceeds by proving a parallel set of theorems to the existing ones, which will replace the originals a few steps down the road. This pull request gets as far as `xpsbas` which is stated as before but with a proof which does not involve `+c`.

More background at #3492

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+26-Sep-23 xpsfrn    xpsfrncda
 26-Sep-23 xpsff1o   xpsff1ocda
 26-Sep-23 xpsfval   xpsfvalcda
 24-Sep-23 nelpr1    [same]      moved from GS's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+26-Sep-23 xpsff1o   xpsff1ocda
+26-Sep-23 xpsfval   xpsfvalcda
 24-Sep-23 nelpr1    [same]      moved from GS's mathbox to main set.mm
 24-Sep-23 nelpr2    [same]      moved from GS's mathbox to main set.mm
 24-Sep-23 breldmd   [same]      moved from GS's mathbox to main set.mm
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -197624,6 +197650,44 @@ topology is based on the order and not the extended metric (which would
...
+  $( The value of a function with a domain of (at most) two elements.
+     (Contributed by Jim Kingdon, 25-Sep-2023.) $)
+  fvprif $p |- ( ( A e. V /\ B e. W /\ C e. 2o )
+      -> ( { <. (/) , A >. , <. 1o , B >. } ` C ) = if ( C = (/) , A , B ) ) $=
+    ( wcel c2o w3a c0 wceq cop c1o cpr cfv cif wa adantr simpr fveq2d 3eqtr4d
+    fvpr0o 3ad2ant1 iftrued fvpr1o 3ad2ant2 1n0 eqeq1d mtbiri iffalsed wo elpri
+    neii df2o3 eleq2s 3ad2ant3 mpjaodan ) ADFZBEFZCGFZHZCIJZCIAKLBKMZNZVAABOZJC
+    LJZUTVAPZIVBNZAVCVDUTVGAJZVAUQURVHUSABDUAUBQVFCIVBUTVARZSVFVAABVIUCTUTVEPZL
+    VBNZBVCVDUTVKBJZVEURUQVLUSABEUDUEQVJCLVBUTVERZSVJVAABVJVALIJLIUFULVJCLIVMUG
+    UHUITUSUQVAVEUJZURVNCILMGCILUKUMUNUOUP $.
```
tirix, line 197674: This looks a bit like `mptprop` (in my mathbox)!

In general, it would be nice if we could get a more general version of this theorem and maybe also `fnpr2ob`, using the more general ` { <. A , B >. , <. C , D >. } ` and `{ A , C }` where you have `2o`.
Then most of these follow with a simple substitution.

tirix (approved):

set.mm
jkingdon: I got partway down the path of doing this when I was making this branch, and I don't fully remember why I abandoned that. I might have been thinking of iset.mm (which would need decidable equality on V in mptprop) but of course iset.mm doesn't have to do exactly the same thing as set.mm, so that's not a definitive reason.

jkingdon:
I just resolved merge conflicts on this pull request. I was thinking of leaving it open for another half day or day, but I might merge it once the automated checks pass to avoid further merge conflicts.

PR #3534, opened 2023-09-29
Prove and minimize with syl2anc2

BTernaryTau:
> Double syllogism inference combined with contraction.


avekens (approved):
I think the minimizations were performed using the Minimize-Script scripts\min.cmd, right?

The huge number of shortened theorems justifies to add syl2anc2 to the main body of set.mm.

BTernaryTau:
Yep, I just used `scripts/minimize-all syl2anc2` to perform the setup and run it for me.

jkingdon (approved):

PR #3536, opened 2023-09-30
fix garden_path directive for metamath-knife build

digama0:
fixes metamath/metamath-knife#135. It turns out that the cause is not a bug in the parser but rather in the additional "garden_path" hints provided to get the parser to accept the grammar. (IMO the fact that we need these hints in the first place is an issue to be fixed in metamath-knife, but for now let's just use this to get things green again.)


jkingdon (approved):
Not sure I know enough about `garden_path` to give a detailed reaction but this seems plausible and it does make https://github.com/metamath/metamath-knife/pull/135 pass, so looks good to me.

set.mm
```diff
@@ -459446,8 +459446,7 @@ and the expression ( x e. A /\ x e. B ) ` .
      LALR parsers would normally detect them as shift-reduce or
      reduce-reduce conflicts, but the current implementation of metamath-knife
      cannot, and therefore requires those hints. $)
-  $( $j garden_path ( x   =>   ( ph ;
-        garden_path ( A   =>   ( ph ;
+  $( $j garden_path ( A   =>   ( ph ;
```
benjub, line 459449: I'm not sure I understand, but doesn't this contradict the rule https://us.metamath.org/mpeuni/co.html ?

digama0: It's not a promise that we will take the path, it's saying that we might need to do a reduction at that point and yet still change our mind later. Without this directive (IIUC) it will stop at the first "reduce", meaning that anything which requires more to match can paint us into a corner. I don't think these directives can be used to make an incorrect automaton, but it causes the generator to do more work, and the resulting automaton can match more expressions. (I have seen garden_path directives cause construction to fail though, with an ambiguity error.)

tirix (approved):

PR #3537, opened 2023-09-30
Fix markup issues and warnings in hol.mm

digama0:
Three warnings remain:
```
warning: Token contains '?'
    --> ../mm/hol.mm:1017:6
     |
1017 |   $c ? $.    $( There exists term $)
     |      - Used '?' character here
     |
     = note: The '?' character is discouraged in tokens because it is
             sometimes used as a math token search wildcard.
warning: Token contains '?'
    --> ../mm/hol.mm:1019:6
     |
1019 |   $c ?! $.   $( There exists unique term $)
     |      -- Used '?' character here
     |
     = note: The '?' character is discouraged in tokens because it is
             sometimes used as a math token search wildcard.
warning: Token contains '@'
    --> ../mm/hol.mm:1700:6
     |
1700 |   $c @ $.     $( Indefinite descriptor $)
     |      - Used '@' character here
     |
     = note: The '@' character is discouraged in tokens because it is
             traditionally used to replace '$' in commented out database source code.
```
I don't intend to fix these though, the characters chosen are deliberate. So we will need some way to silence the warning.

### DIFF
--- a/hol.mm
+++ b/hol.mm
@@ -43,6 +43,8 @@ $)
   $v x y z f g p q $.  $( Bound variables $)
   $v A B C F R S T $.  $( Term variables $)
 
+  $( $j syntax 'var' 'type' 'term'; bound 'var'; $)
+
   $( Let variable ` al ` be a type. $)
   hal $f type al $.
   $( Let variable ` be ` be a type. $)
@@ -107,6 +109,8 @@ $)
 
   $c wff $.  $( Not used; for mmj2 compatibility $)
 
+  $( $j syntax 'wff'; syntax '|-' as 'wff'; $)
+
   $( Internal axiom for mmj2 use. $)
   wffMMJ2 $a wff A |= B $.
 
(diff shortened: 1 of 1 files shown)


hol.mm
```diff
@@ -228,118 +230,150 @@
...
-       see ~ ax-cb1 .) $)
-    wctr $a |- T : bool $.
+       see ~ ax-cb1 .)  Prefer ~ wctr .  (New usage is discouraged.)
+       (Contributed by Mario Carneiro, 8-Oct-2014.) $)
+    ax-wctr $a |- T : bool $.
+
+    $( Reverse closure for the type of a context.  (This axiom is unnecessary;
+       see ~ ax-cb1 .)  (Contributed by Mario Carneiro, 8-Oct-2014.) $)
+    wctl $p |- S : bool $=
+      ( ax-wctl ) ABCD $.
```
benjub, line 257: Out of curiosity: why is it necessary to duplicate the $a-statement with an identical $p-statement ?

benjub (approved):

hol.mm
digama0: metamath-knife complains that an axiom is not being called `ax-*` or `df-*`. In this database there really is another category, for typing judgments (which are normally excluded from consideration because they are syntax axioms but here have to be provable theorems because they are not context-free). But axiom/definition display is important for the website among other things, so I suppose it's fine to just name them as directed.

jkingdon:
> So we will need some way to silence the warning.

Yeah, `$j allow-unusual-token-name` or something perhaps.

benjub:
> Yeah, `$j allow-unusual-token-name` or something perhaps.

Yes, or something more specific: either "character '?' is allowed in labels", or the even more specific "label ?blah is allowed"

digama0:
I was thinking of something a bit more targeted: `$j allow contains_qmark;` immediately before the $c directive, repeated before the other one, and then `$j allow contains_at;` before the last one. These are "lint warning suppressors", they apply exactly once to the statement or block following them. The `allow` is its own keyword to handle all lint suppressors of this form. (This is modeled after Rust's `#[allow(clippy::too_many_arguments)]` lint supression mechanism.)

PR #3539, opened 2023-09-30
Conventions for suffix "t"

avekens:
Since it was discussed again when and how to use the suffix "t" (see PR #3529), I would like to enhance the conventions accordingly (as I already proposed in the Google group some time ago, see https://groups.google.com/g/metamath/c/4TSHENMYqtw/m/56-5mDWLAgAJ).  

Since the suffix (or more general label fragment) "t" is also used as abbreviation for true/truth sometimes, I also enhanced the table of label fragments accordingly.


set.mm
```diff
@@ -424241,18 +424241,21 @@ allows us to easily use the theorem in places where (in traditional
        Deductions have a label suffix of "d", especially if there are other
        forms of the same theorem (e.g., ~ pm2.43d ).  The labels for inferences
        usually have the suffix "i" (e.g., ~ pm2.43i ).  The labels of theorems
-       in "closed form" would have no special suffix (e.g., ~ pm2.43 ).  When
-       an inference is converted to a theorem by eliminating an "is a set"
+       in "closed form" would have no special suffix (e.g., ~ pm2.43 ) or, if
+       the standard label for the closed form is already in use, suffix "t"
```
benjub, line 424245:
```suggestion
       the non-suffixed label is already used, then we add the suffix "t" for the closed form
```

set.mm
```diff
@@ -424241,18 +424241,21 @@ allows us to easily use the theorem in places where (in traditional
...
+       (for "theorem" or "tautology", e.g., ~ ancomst or ~ nfimt ).  When an
+       inference is converted to a theorem by eliminating an "is a set"
        hypothesis, we sometimes suffix the closed form with "g" (for "more
```
benjub, line 424248: I think it's hard to understand for someone who hasn't seen an example of that. So maybe:
> When a theorem has a hypothesis of the form `A e. _V` and we prove a more general version where that hypothesis is replaced with an antecedent of the form `( A e. V -> ...`, then we suffix a "d" to the label of the original theorem, as in ~uniexg generalizing ~ uniex .

set.mm
```diff
@@ -425067,6 +425070,7 @@ the theorem forms ("i", "d" or "g") should be the first of multiple
        <li> r : theorem with something reversed (e.g., a biconditional)</li>
        <li> s : inference that manipulates an antecedent ("s" refers to an
        application of ~ syl that is eliminated)</li>
+       <li> t : theorem in closed form not having an "is a set" antecedent</li>
```
benjub, line 425073:
```suggestion
       <li> t : theorem in closed form</li>
```

set.mm
```diff
@@ -424241,18 +424241,21 @@ allows us to easily use the theorem in places where (in traditional
...
        When submitting a new theorem, a revision of a theorem, or an upgrade
        of a theorem from a Mathbox to the Main database, please use the
        general form to be the default form of the theorem, without the suffix
-       g . For example, "brresg" lost its suffix g when it was revised for some
-       other reason, and now it is ~ brres . Its inference form which was the
-       original "brres", now is ~ brresi .
+       "g" . For example, "brresg" lost its suffix "g" when it was revised for
+       some other reason, and now it is ~ brres .  Its inference form which was
+       the original "brres", now is ~ brresi .  The same holds for the suffix
+       "t".
```
benjub, line 424258: A bit anecdotical to me, but that's fine.

set.mm
jkingdon: seems like an improvement in wording

set.mm
```diff
@@ -424242,12 +424242,13 @@ allows us to easily use the theorem in places where (in traditional
...
        in "closed form" would have no special suffix (e.g., ~ pm2.43 ) or, if
-       the standard label for the closed form is already in use, suffix "t"
-       (for "theorem" or "tautology", e.g., ~ ancomst or ~ nfimt ).  When an
-       inference is converted to a theorem by eliminating an "is a set"
-       hypothesis, we sometimes suffix the closed form with "g" (for "more
-       general") as in ~ uniex versus ~ uniexg .  In this case, the inference
-       often has no suffix "i".
+       the non-suffixed label is already used, then we add the suffix "t" (for
+       "theorem" or "tautology", e.g., ~ ancomst or ~ nfimt ).  When an
+       inference with an "is a set" hypothesis (e.g., ` A e. _V `) is converted
```
benjub, line 424247: I was going to say: the theorem is not "converted", we just add another one; but then, I remembered that sometimes we replaced the old one with the new one, so we can speak of a "conversion" indeed.

benjub (approved):

PR #3540, opened 2023-09-30
Fix markup issues and warnings in nf.mm

digama0:
Mostly just rewrapping and adding `New usage is discouraged.` on OLD and ALT theorems. I also had to add the garden path directive from set.mm, I just put it next to the typesetting section since @tirix doesn't have a mathbox here.


jkingdon (approved):

benjub (approved):
LGTM, but we should probably wait for @scott-fenton or @sctfn's approval.

tirix (approved):

sctfn:
LGTM

PR #3542, opened 2023-09-30
Fixed OLD proofs in @tirix mathbox

GinoGiotto:
This is the PR that I anticipated in https://github.com/metamath/set.mm/pull/3472#discussion_r1325983063, following the instructions provided by @tirix in https://github.com/metamath/set.mm/pull/3472#discussion_r1324208286. Theorems  `measdivcstOLD`, `probfinmeasbOLD`, `axtglowdim2OLD`, `axtgupdim2OLD` are alternative formulations of pre-existing theorems, therefore, according to [conventions](https://us.metamath.org/mpeuni/conventions.html), I renamed them with the ALTV suffix and placed them (when possible) below their corresponding original ones. Discouraged labels are no longer a requirement with the new suffix, so I removed `(Proof modification is discouraged.)` since I don't see a default reason for them to be excluded by an eventual global minimization, but about this @tirix should have the final word.

This is part of a plan of fixing OLD proofs that don't match the [conventions](https://us.metamath.org/mpeuni/conventions.html), perhaps I'll open an issue to address each one appropriately (sorry tirix for being late compared to what I had promised).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -6039,7 +6039,6 @@
 "golem1" is used by "golem2".
 "golem2" is used by "goeqi".
 "grothac" is used by "axgroth3".
-"grothpwex" is used by "isrnsigaOLD".
 "grpbasex" is used by "cnaddablx".
 "grpbasex" is used by "isgrpix".
 "grpbasex" is used by "zaddablx".
@@ -8542,8 +8541,8 @@
 "isst" is used by "sticl".
 "isst" is used by "stj".
 "isst" is used by "strlem3a".
-"istrkg2d" is used by "axtglowdim2OLD".
-"istrkg2d" is used by "axtgupdim2OLD".
+"istrkg2d" is used by "axtglowdim2ALTV".
+"istrkg2d" is used by "axtgupdim2ALTV".
 "isvcOLD" is used by "isvciOLD".
 "isvciOLD" is used by "cncvcOLD".
 "isvciOLD" is used by "hhssnv".
@@ -9182,8 +9181,8 @@
 "mdsymlem6" is used by "mdsymlem7".
 "mdsymlem7" is used by "mdsymlem8".
 "mdsymlem8" is used by "mdsymi".
-"measdivcstOLD" is used by "probfinmeasbOLD".
-"measdivcstOLD" is used by "probmeasb".
+"measdivcstALTV" is used by "probfinmeasbALTV".
+"measdivcstALTV" is used by "probmeasb".
 "meetcomALT" is used by "meetcom".
 "merco1" is used by "merco1lem1".
 "merco1" is used by "merco1lem10".
@@ -13771,8 +13770,8 @@ New usage of "axresscn" is discouraged (2 uses).
 New usage of "axrnegex" is discouraged (0 uses).
 New usage of "axrrecex" is discouraged (0 uses).
 New usage of "axsep" is discouraged (0 uses).
-New usage of "axtglowdim2OLD" is discouraged (0 uses).
-New usage of "axtgupdim2OLD" is discouraged (0 uses).
+New usage of "axtglowdim2ALTV" is discouraged (0 uses).
+New usage of "axtgupdim2ALTV" is discouraged (0 uses).
 New usage of "baerlem5abmN" is discouraged (0 uses).
 New usage of "baerlem5amN" is discouraged (0 uses).
 New usage of "baerlem5bmN" is discouraged (0 uses).
@@ -15544,7 +15543,7 @@ New usage of "golem2" is discouraged (1 uses).
 New usage of "grothac" is discouraged (1 uses).
 New usage of "grothomex" is discouraged (0 uses).
 New usage of "grothpw" is discouraged (0 uses).
-New usage of "grothpwex" is discouraged (1 uses).
+New usage of "grothpwex" is discouraged (0 uses).
 New usage of "grpbasex" is discouraged (3 uses).
 New usage of "grpinvfvalALT" is discouraged (0 uses).
 New usage of "grpo2inv" is discouraged (4 uses).
@@ -16175,7 +16174,6 @@ New usage of "ispsubcl2N" is discouraged (0 uses).
 New usage of "ispsubclN" is discouraged (10 uses).
 New usage of "isrngo" is discouraged (2 uses).
 New usage of "isrngod" is discouraged (1 uses).
-New usage of "isrnsigaOLD" is discouraged (0 uses).
 New usage of "issetiOLD" is discouraged (0 uses).
 New usage of "issgrpALT" is discouraged (1 uses).
 New usage of "issh" is discouraged (3 uses).
@@ -16510,7 +16508,7 @@ New usage of "mdsymlem5" is discouraged (1 uses).
 New usage of "mdsymlem6" is discouraged (1 uses).
 New usage of "mdsymlem7" is discouraged (1 uses).
 New usage of "mdsymlem8" is discouraged (1 uses).
-New usage of "measdivcstOLD" is discouraged (2 uses).
+New usage of "measdivcstALTV" is discouraged (2 uses).
 New usage of "meetcomALT" is discouraged (1 uses).
 New usage of "merco1" is discouraged (16 uses).
 New usage of "merco1lem1" is discouraged (2 uses).
@@ -16959,7 +16957,6 @@ New usage of "ocsh" is discouraged (6 uses).
 New usage of "ocss" is discouraged (11 uses).
 New usage of "ocval" is discouraged (4 uses).
 New usage of "odfvalALT" is discouraged (0 uses).
-New usage of "ogrpinvOLD" is discouraged (0 uses).
 New usage of "oldmm3N" is discouraged (2 uses).
 New usage of "olposN" is discouraged (0 uses).
 New usage of "om0x" is discouraged (0 uses).
@@ -17295,7 +17292,7 @@ New usage of "prmgapprmo" is discouraged (0 uses).
 New usage of "prn0" is discouraged (8 uses).
 New usage of "prnmadd" is discouraged (2 uses).
 New usage of "prnmax" is discouraged (7 uses).
-New usage of "probfinmeasbOLD" is discouraged (0 uses).
+New usage of "probfinmeasbALTV" is discouraged (0 uses).
 New usage of "prpssnq" is discouraged (8 uses).
 New usage of "prub" is discouraged (6 uses).
 New usage of "psgnunilem5OLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


discouraged
```diff
@@ -9182,8 +9181,8 @@
 "mdsymlem6" is used by "mdsymlem7".
 "mdsymlem7" is used by "mdsymlem8".
 "mdsymlem8" is used by "mdsymi".
-"measdivcstOLD" is used by "probfinmeasbOLD".
-"measdivcstOLD" is used by "probmeasb".
+"measdivcstALTV" is used by "probfinmeasbALTV".
+"measdivcstALTV" is used by "probmeasb".
```
avekens, line 9185: Neither an OLD nor an ALTV theorem should be used by a non-OLD/ALTV theorem. Can the proof of ~probmeasb be revised accordingly (using ~ measdivcst instead of measdivcstALTV)?

GinoGiotto: > Neither an OLD nor an ALTV theorem should be used by a non-OLD/ALTV theorem.

I have some reservations about the truth of this statement in the general case. Notice how OLD proofs always require a `(New usage is discouraged.)` label, but that's not the case for ALTV proofs. Therefore, accoring to [conventions](https://us.metamath.org/mpeuni/conventions.html), some ALTV theorems can be picked up through a minimization process by non-OLD/ALTV theorems (if you don't want that, you would advocate for `minimize * /allow * /no ax-* /forbid *ALTV`, but at the moment that's not how default minimizations are indicated to be done).

So the question is whether the use of ~ measdivcst is preferred over measdivcstALTV in this particular case, but not because of a general rule.

digama0: ALT does require `(New usage is discouraged.)`; I'm not sure what ALTV is supposed to indicate.

GinoGiotto: > ALT does require `(New usage is discouraged.)`; I'm not sure what ALTV is supposed to indicate.

Yes ALT does, but ALTV does not, I read it from the conventions [here](https://us.metamath.org/mpeuni/conventions.html)

GinoGiotto: ALT versions are supposed to be used when different proofs of the same theorem are provided, while ALTV means that the statement itself differs from the original. I think in the database some theorems who were supposed to be ALTV were labeled as ALT instead (example sbequ1 and sbequ1ALT) and this created confusion.

benjub: When was that decided ? That seems rather peculiar to me. The database set.mm is full of "alternative versions", and there are no special labellings for them (in particular, no capitalization). Do you have examples ?

avekens:
I habe no Access to my Computer, so only in short: this Suffix was introduced After a Diskussion with Norm. I can explain ist in  more Detail when I am back home… 

Von meinem iPhone gesendet

discouraged
GinoGiotto: > When was that decided ? That seems rather peculiar to me.

Everything I'm saying comes from the conventions, here is a copy-paste of them:

>Alternate (ALT) proofs. If a different proof is shorter or clearer but uses more or stronger axioms, we make that proof an "alternate" proof (marked with an ALT label suffix), even if this alternate proof was formalized first. We then make the proof that requires fewer axioms the main proof. Alternate proofs can also occur in other cases when an alternate proof gives some particular insight. Their comment should begin with "Alternate proof of ~ xxx " followed by a description of the specificity of that alternate proof. There can be multiple alternates. Alternate (*ALT) theorems should have "(Proof modification is discouraged.) (New usage is discouraged.)" in their comment and should follow the main statement, so that people reading the text in order will see the main statement first. The alternate and main statement comments should use hyperlinks to refer to each other.

>Alternative (ALTV) versions. If a theorem or definition is an alternate/variant of an existing theorem or definition, its label should have the same name with suffix ALTV. Such alternates are often temporary only, until it is decided which alternative should be used in the future. Alternate (*ALTV) theorems or definitions are usually contained in mathboxes. Their comments need not to contain "(Proof modification is discouraged.) (New usage is discouraged.)". Their comment should begin with "Alternate version of ~ xxx " followed by a description of the specificity of the difference from the main variant. Alternate statements should generally follow the main statement.

>Old (OLD) versions or proofs. If a proof, definition, axiom, or theorem is going to be removed, we often stage that change by first renaming its label with an OLD suffix (to make it clear that it is going to be removed). Old (*OLD) statements should have "(Proof modification is discouraged.) (New usage is discouraged.)" and "Obsolete version of ~ xxx as of dd-Mmm-yyyy." (not enclosed in parentheses) in the comment. An old statement should follow the main statement, so that people reading the text in order will see the main statement first. This typically happens when a shorter proof to an existing theorem is found: the existing theorem is kept as an *OLD statement for one year. When a proof is shortened automatically (using the Metamath program "MM-PA> MINIMIZE__WITH *" command), then it is not necessary to keep the old proof, nor to add credit for the shortening.

It clearly states that ALT suffixes are for alternative proofs, while ALTV suffixes are for variants of the original theorem.

> The database set.mm is full of "alternative versions", and there are no special labellings for them (in particular, no capitalization). Do you have examples ?

I'm not sure I understand, you mean examples of already existing ALTV theorems? If yes than there are several of them, such as `uniqs` in main and `uniqsALTV` is in Peter Mazsa's mathbox (all ALTV versions are in mathboxes as far as I know).

benjub: Ok. I didn't remember that. Since *ALTV is a mathbox-only suffix, this should appear in the conventions. Also, it seems that this MR has *ALTV statements in the main part of set.mm, which should not be the case.

GinoGiotto: > Also, it seems that this MR has *ALTV statements in the main part of set.mm, which should not be the case.

Not true, the title of this PR clearly states that I only touched @tirix mathbox, and that's what I did.

benjub: If it's mathbox only, then fair enough.

tirix (approved):

tirix:
Provided the extract of the "conventions" provided by @GinoGiotto, `ALTV` seems to be the correct suffix.
Thanks a lot for helping me maintain my mathbox @GinoGiotto !

Ultimately I think we should get the best out of the Measure Theory from Glauco's and my Mathbox and move it to main, but that's a bigger task.

discouraged
avekens: > > Neither an OLD nor an ALTV theorem should be used by a non-OLD/ALTV theorem.
> 
> I have some reservations about the truth of this statement in the general case. Notice how OLD proofs always require a `(New usage is discouraged.)` label, but that's not the case for ALTV proofs. Therefore, accoring to [conventions](https://us.metamath.org/mpeuni/conventions.html), some ALTV theorems can be picked up through a minimization process by non-OLD/ALTV theorems (if you don't want that, you would advocate for `minimize * /allow * /no ax-* /forbid *ALTV`, but at the moment that's not how default minimizations are indicated to be done).
> 
> So the question is whether the use of ~ measdivcst is preferred over measdivcstALTV in this particular case, but not because of a general rule.

OK, it's not in the conventions, ut thee is no case in set.mm where an ALTV theorem is used in a proof of a non-OLD/ALTV theorem. So no change required now, but it must be replaced in the proof of ~probmeasb before ~probmeasb is moved to Main at the latest.

avekens (approved):

GinoGiotto:
> there is no case in set.mm where an ALTV theorem is used in a proof of a non-OLD/ALTV theorem.

In main this is true, because there are currently no ALTV versions there (note that the conventions don't say it's forbidden for them to be in main, they only say that ALTV versions are usually contained in mathboxes).

However there are some examples of ALTV versions in mathboxes used by non-OLD/ALTV theorems. Theorem [exan3](https://us.metamath.org/mpeuni/exan3.html) uses `elecALTV`, theorem [inxpex](https://us.metamath.org/mpeuni/inxpex.html) uses `inex2ALTV`, theorem [eqvrelqsel](https://us.metamath.org/mpeuni/eqvrelqsel.html) uses `elecALTV` just to name a few.

I think the root of the problem here is that in set.mm there is currently a sort of double standards situation, which I think is the cause of the doubts in this PR. Here is a quick look from me:

In main:

* OLD proofs follow the conventions except for [19.43OLD](https://us.metamath.org/mpeuni/19.43OLD.html).
* ALT proof are sometimes used for alternative versions and not for alternative proofs as the conventions would require.
* No presence of ALTV versions.

In mathboxes:

Here the current situation is messier, to the point that the boundaries between these suffixes are not clear anymore.

* Many OLD proofs are alternative versions without a proper expiration date or they are expired years ago, and there is no clear planning of deleting/renaming them. Some of them may be more appropriately renamed as ALTV.

* ALT is sometimes used for different formulations as well, and some statements are also not capitalized (example [sbgoldbalt](https://us.metamath.org/mpeuni/sbgoldbalt.html) whose comment is `An alternate (related to the original) formulation of the binary
       Goldbach conjecture` or [cdmatalt](https://us.metamath.org/mpeuni/cdmatalt.html))
* ALTV is the advised suffix for alternative formulations, but some statements don't reference their original version.

Perhaps this matter would be more properly addressed through an issue. I'm not particularly attached to these specific suffix definitions, but what I do value is consistency. Labeling convention simplifies the search function which has a non-trivial impact in big databases. The overall collection would look more organized if these suffixes would stay the same for both main and mathboxes (and even across different databases). Also it would make maintenance easier, in a scenario where all OLD proofs follow the conventions there would be no need to ask the original author to delete them or not, because the author itself would have already declared its intention by using the suffix.

benjub:
> * ALTV is the advised suffix for alternative formulations, but some statements don't reference their original version.

Is it ? I think *ALTV was mainly allowed for one contributor (@avekens) and then used by a second one (@mazsa) in their mathboxes. I think this should not spread, and especially not go to the main section. The main reason is that being a variant of another statement is a vage notion. Also, two variants may differ in many different ways, and that should be reflected in the label: do they differ by a commutation, by the fact of being (un)curried, by the fact of having an antecedent instead of a hypothesis, a biconditional instead of an implication... Here are a very few examples of variant statements in Main:
* con3dimp / con3d
* anim12d1 / anim12d
* anbi2ci / anbi2i
* pm5.31r / pm 5.31
* al1imi / alanimi
* ax6ev / ax6evr
* abid1 / abid2

In contrast, the only two special suffixes with capital letters in Main, OLD and ALT (with variants OLDOLD, etc.) are precisely defined and have specific purposes.

`*ALTV` is fine in mathboxes and, as said, they exist only as long as it has been decided which one of the variants will be used, or if both will be kept (in which case, more meaningful labels should be found).

avekens:
I agree with both of you, @benjub and @GinoGiotto : If ALTV is used, it should be used according to the conventions, but it should be used only in Mathboxes, and only for technical reasons. (BTW ~[sbgoldbalt](https://us.metamath.org/mpeuni/sbgoldbalt.html) is not an technical aternative, but an alternative used in literature).

Anyway, I think there is no need for any further discussion (as long as ALTV is used in Mathboxes only, which is in the responsibility of the Mathbox owner), and I will merge this PR soon

PR #3545, opened 2023-10-01
Prove hashf1dmcdm

BTernaryTau:
hashfundm
> The size of a set function is equal to the size of its domain.

hashf1dmrn
> The size of the domain of a one-to-one set function is equal to the size of its range.

hashf1dmcdm
> The size of the domain of a one-to-one set function is less than or equal to the size of its codomain, if it exists.


set.mm
```diff
@@ -501732,6 +501732,34 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
...
+    $.
+
+  $( The size of the domain of a one-to-one set function is equal to the size
+     of its range.  (Contributed by BTernaryTau, 1-Oct-2023.) $)
+  hashf1dmrn $p |- ( ( F e. V /\ F : A -1-1-> B ) ->
+          ( # ` A ) = ( # ` ran F ) ) $=
+    ( wcel wf1 wa chash cfv cdm crn wfun wceq f1fun hashfundm sylan2 cvv adantl
+    f1dm dmexg adantr eqeltrrd hashf1rn sylancom fveq2d 3eqtr3rd ) CDEZABCFZGZC
+    HIZCJZHIZCKHIZAHIUHUGCLUJULMABCNCDOPUGUHAQEUJUMMUIUKAQUHUKAMUGABCSRZUGUKQEU
+    HCDTUAUBABCQUCUDUIUKAHUNUEUF $.
```
tirix, line 501752: The theorem [`~hashf1rn`](https://us.metamath.org/mpeuni/hashf1rn.html) is very similar.
Did you notice it?

set.mm
```diff
@@ -501732,6 +501732,34 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
...
 
+  $( The size of a set function is equal to the size of its domain.
+     (Contributed by BTernaryTau, 30-Sep-2023.) $)
+  hashfundm $p |- ( ( F e. V /\ Fun F ) -> ( # ` F ) = ( # ` dom F ) ) $=
+    ( cfn wcel wfun wa chash cfv cdm wceq wi hashfun biimpd adantld wn w3a cpnf
+    hashinf 3adant2 cvv wb fundmfibi notbid adantl dmexg sylan ex adantr sylbid
+    3impia eqtr4d 3comr 3expib pm2.61i ) ACDZABDZAEZFZAGHZAIZGHZJZKUOUQVBUPUOUQ
+    VBALMNUOOZUPUQVBUPUQVCVBUPUQVCPUSQVAUPVCUSQJUQABRSUPUQVCVAQJZURVCUTCDZOZVDU
+    QVCVFUAUPUQUOVEAUBUCUDUPVFVDKUQUPVFVDUPUTTDVFVDABUEUTTRUFUGUHUIUJUKULUMUN
+    $.
```
tirix, line 501743: Nice, I think this is a clear improvement compared to [`~hashfun`](https://us.metamath.org/mpeuni/hashfun.html) which only covers the finite case.

set.mm
BTernaryTau: Yep, I used it to prove this one.

set.mm
BTernaryTau: Thanks!

set.mm
jkingdon: It seems quite plausible to cast this in terms of the size of the domain and range, since the size of a function, while well defined in set.mm, is arguably less natural.

jkingdon (approved):

set.mm
benjub: > It seems quite plausible to cast this in terms of the size of the domain and range, since the size of a function, while well defined in set.mm, is arguably less natural.

Agreed. I think that hashfundm is more like a lemma used to prove hashf1dmrn, since indeed "the size of a function" does not make sense (even though, with set.mm's definition of functions, it is the size of its domain).

PR #3547, opened 2023-10-03
Replace +c with a pair of ordered pairs from xpsadd to xpsms

jkingdon:
The background is at #3492 

This was straightforward to implement as the needed helper theorems were already proved in #3532 

The stopping point is that this pull request goes as far as needed to move everything from https://us.metamath.org/mpeuni/xpslem.html to https://us.metamath.org/mpeuni/xpsrnbas.html .  After this, there's probably just one more pull request of roughly this size needed to finish the rest of #3492 and change the `xps` definition from https://us.metamath.org/mpeuni/df-xps.html to https://us.metamath.org/mpeuni/dfxpspr.html

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 2-Oct-23 xpsfrncda xpsfrn      changes from +c to a pair of ordered pairs
  2-Oct-23 xpsff1o2cda xpsff1o2  changes from +c to a pair of ordered pairs
  2-Oct-23 xpsvalcda xpsval      changes from +c to a pair of ordered pairs
  2-Oct-23 xpslem    xpsrnbas    changes from +c to a pair of ordered pairs
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -26,6 +26,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 2-Oct-23 xpsff1ocda xpsff1o    changes from +c to a pair of ordered pairs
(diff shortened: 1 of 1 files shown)


changes-set.txt
```diff
@@ -26,6 +26,18 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 2-Oct-23 xpsfeqcda xpsfeq      changes from +c to a pair of ordered pairs
+ 2-Oct-23 xpscfv    fvprif      changes from +c to a pair of ordered pairs
```
wlammen, line 30: Why do the comments of the two functions differ in detail, and why is fvprif valid for a domain of just one element?

wlammen (changes requested):

changes-set.txt
jkingdon: The comment of `fvprif` was incorrect - there are always two elements there because of the 𝐴 ∈ 𝑉 ∧ 𝐵 ∈ 𝑊 condition. I've pushed up a fix to use the comment which had been on `xpsfeq`

wlammen (approved):

benjub:
Thanks @jkingdon for the implementation and @wlammen for the review ! Big efforts went into both.

jkingdon:
> Thanks @jkingdon for the implementation and @wlammen for the review ! Big efforts went into both.

Aw, shucks. Straightforward implementation, but a lot of proof lines to touch. We'll be glad to have it when it is eventually done, as these proofs will be easier to read and work with.

Agreed about the review: even if I just see the few lines which were commented on, I can tell from the comments that the review looked over a lot of things, and in cases like this found things worth improving.

PR #3551, opened 2023-10-03
Reduce ax-13 usage with bj-chvarv

GinoGiotto:
* Moved `bj-chvarv` from @benjub's mathbox to main.

* Renamed `bj-chvarv` into `chvarfv` (~ chvarv already exists).

* Removed discouragement tag from `chvarfv`.

* Use `chvarfv` to drop ax-13 from ~ axrep1, ~ axsep2, ~ isso2i.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18416,7 +18416,6 @@ Proof modification of "bj-ceqsalt0" is discouraged (116 steps).
 Proof modification of "bj-ceqsalt1" is discouraged (118 steps).
 Proof modification of "bj-ceqsaltv" is discouraged (45 steps).
 Proof modification of "bj-ceqsalv" is discouraged (10 steps).
-Proof modification of "bj-chvarv" is discouraged (18 steps).
 Proof modification of "bj-chvarvv" is discouraged (11 steps).
 Proof modification of "bj-clabel" is discouraged (43 steps).
 Proof modification of "bj-cleljusti" is discouraged (20 steps).


set.mm
```diff
@@ -529337,12 +529336,12 @@ at least two objects (see ~ bj-dtru ).  (Contributed by BJ,
     bj-axrep2 $p |- E. x ( E. y A. z ( ph -> z = y ) ->
```
wlammen, line 529336: What is the reason this does not replace axrep2?

GinoGiotto: Good catch. Actually `bj-axrep1` `bj-axrep2` `bj-axrep3` `bj-axrep4` `bj-axrep5` `bj-axsep` drop ax-13 without adding any $d condition, so they can simply replace the current `axrep1` `axrep2` `axrep3` `axrep4` `axrep5` `axsep` in main. I think I'm going take them, I only have to keep an eye on proof minimizations and appropriate credits.

wlammen: Maybe ax-sep was seen as the first in the order, and the rest follows up.   This is just a wild guess, though.

icecream17: I will also mention [sn-axsep](https://us.metamath.org/mpeuni/sn-axsep.html), incidentally I think it has an extra dv condition on `ph z` (so sn-axsep is actually axsep2)

benjub: Thanks @wlammen and @GinoGiotto. The main reason these were in my mathbox is that I had done then in a batch and Norm wisely insisted that they be moved to Main only carefully and progressively, to allow careful strutiny, not in a big batch. I have since been busy with other things and didn't carry out the move. So thank you Gino for doing it.

Maybe it's better to merge this one as is, and do what you propose in another MR, once we see the resulting website ?

benjub (approved):

set.mm
GinoGiotto: > Maybe it's better to merge this one as is, and do what you propose in another MR, once we see the resulting website ?

Yes, maybe it's a little easier to do two separate PRs so that I put things better in order in my mind.

icecream17: maybe call sn-axrep, axrep1

avekens (approved):

PR #3552, opened 2023-10-03
Prove and minimize with syladantr, syladantl

BTernaryTau:
> Syllogism inference combined with an addition to the antecedent.

After minimizing, syladantr is used 758 times and syladantl is used 397 times.

As a general note, I've been finding these minimization opportunities by looking at the most commonly used theorems (as per the description of syl) and checking which pairs of these seem most common.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -13339,6 +13339,7 @@ New usage of "2clwwlklemOLD" is discouraged (2 uses).
 New usage of "2cnALT" is discouraged (0 uses).
 New usage of "2eu1OLD" is discouraged (0 uses).
 New usage of "2eu3OLD" is discouraged (0 uses).
+New usage of "2eu5OLD" is discouraged (0 uses).
 New usage of "2irrexpqALT" is discouraged (0 uses).
 New usage of "2llnjN" is discouraged (1 uses).
 New usage of "2llnjaN" is discouraged (1 uses).
@@ -18172,6 +18173,7 @@ Proof modification of "2clwwlklemOLD" is discouraged (90 steps).
 Proof modification of "2cnALT" is discouraged (3 steps).
 Proof modification of "2eu1OLD" is discouraged (83 steps).
 Proof modification of "2eu3OLD" is discouraged (134 steps).
+Proof modification of "2eu5OLD" is discouraged (81 steps).
 Proof modification of "2irrexpqALT" is discouraged (90 steps).
 Proof modification of "2logb9irrALT" is discouraged (141 steps).
 Proof modification of "2pm13.193" is discouraged (91 steps).


wlammen:
The two theorems syl and adantr have nothing in common.  They are almost unrelated, and just happen to appear frequently in succession in the wild.  The opportunities found by you are probably not exhaustive - breaking up 3syl and 4syl into isolated syl and then looking for adjacent adantr/adantl may reveal more occasions for such shortenings.

The proofs become shorter, but I wonder whether readability increases.  Combining two orthogonal primitive ideas into a single operation is somewhat difficult, because distinctive concepts/subproofs are less easy identifiable.  Though, it has been done in the past (syl6mpi), so lets see how others see it.

The number of changes are very massive.  This has two consequences:  Other PRs  still open, or in preparation may suffer from this because of merge conflicts.  And second I cannot review the contribution because GitHub won't let me see the differences.  I am much in favor of changing over to a new style only in smaller portions, say section-wise.

BTernaryTau:
> The opportunities found by you are probably not exhaustive - breaking up 3syl and 4syl into isolated syl and then looking for adjacent adantr/adantl may reveal more occasions for such shortenings.

I see how that'd be useful in the 4syl case, but it's unclear to me why we'd want to replace 3syl and adantr with syl and syladantr. Can you elaborate?

> Combining two orthogonal primitive ideas into a single operation is somewhat difficult, because distinctive concepts/subproofs are less easy identifiable.

I guess I haven't really been taking this into consideration because at least for me, the way metamath reorders proof steps already makes subproofs too difficult to identify (and in fact this is my biggest gripe with metamath). If that's not the case for others, then I'll have to keep that in mind going forward.

> And second I cannot review the contribution because GitHub won't let me see the differences.

The workaround I use to at least see the changes is to use GitHub Desktop to fetch the pull request and then view the commits from there. I'll admit this isn't great though.

> I am much in favor of changing over to a new style only in smaller portions, say section-wise.

How many sections are you thinking of for this?

wlammen:
I thought along the lines of the sections introduced by
lines.  This roughly limits the number of changes per PR to, say 70 or so.

Important are the number of replacements per PR, and that should be in the order of not overwhelming those checking the results.  PR #3547 is already big, and it takes me more than an hour to verify various aspects.

But before starting to rework your contribution lets see how others think about the idea (emphasizing shortening to the point that maybe subproof boundaries are not sufficiently honored any longer) behind this PR.

BTernaryTau:
Given the concerns about subproof boundaries, I don't think this is a change worth spending ~20 PRs on (or twice that if syladantr and syladantl are separated too). I think I'll just close this, but I'll leave it open for a day to see if there are any objections to me doing that.

benjub:
A few quick remarks:
* As for proof shortenings versus readability: it's true that they can go in opposite directions at times, and there have been many debates here along the years. It seems that most often, we favored shortenings. I can't see the changes either (see third point), but since the combined steps are both in propcalc, this should not hamper readability too much, but I'd be interested to know by how much they shorten a typical proof.
* As for automatically finding often-used subproof patterns, there was a similar issue by @savask (https://github.com/metamath/set.mm/issues/2710); you can read some of the interrogations we had at the time, and maybe he can share his experience here.
* It's true that this MR is too large, but I would like that you do first a MR with only the two new theorems. I think that's what @avekens often does with big minimizations: first the new additions (the human part), and then the automatic minimizations. After we see the changes, maybe, you'll then minimize either by parts, or maybe minimize all of the main section (i.e. exclude mathboxes) with one of your two theorems, and then with the other, and then similarly with mathboxes (I suspect that many of the lines that change in this PR are from some huge proofs in Glauco's mathbox).

jkingdon:
> * As for proof shortenings versus readability: it's true that they can go in opposite directions at times, and there have been many debates here along the years. It seems that most often, we favored shortenings.

I agree with both parts of this:

(1) that one can often shorten proofs by making a big(er) theorem which takes more hypotheses or variables, but which is hard to keep in your head when you are looking at it or the proof which uses it (one example I was able to find is https://us.metamath.org/mpeuni/prlem2.html but I think there might be some others a bit like that or even more so).

but

(2) we have in many cases chosen to add such theorems as long as they do indeed make proofs shorter.

I guess one way around the readability problem, at least some of the time, is if the theorems fit into well characterizable families like `simp*` , `syl*anc`, `ad*ant*`, etc. Even if a theorem like that is kinda long in terms of the number of variables, you know what it is up to if you have seen enough similar theorems to recognize the pattern.

Summary: I do not object to "I think I'll just close this" but will acknowledge this is either in a grey area or pretty close to one.

wlammen:
Let me explain in more detail why I am a bit suspicious about this move:

A syllogism like syl ends with a conclusion, which finalizes a thought, so to say.  Adding an antecedent is in preparation for the next move and belongs to the next step, so to say.  It is natural that on the boundary of proof ideas syllogisms meet the introductions to the next syllogisms.  Using a shortcut bridging these phases serves only the length of a proof in bytes.  Assume you now systematically look out for merging any syllogism (not just syl, bitri for example) with follow up instructions belonging to the next step, whenever you find 7 instances that match -- well, I am not sure we are on a safe path there.

benjub:
@BTernaryTau : can you add syladantr, syladantl to your mathbox, with your remark "is used 758 times and syladantl is used 397 times" in their comments ?

PR #3553, opened 2023-10-03
dfmoeu and dfeumo

wlammen:
1. Move the former dfmolem to the beginning of the uniqueness section, and rename it to dfmoeu.  The purpose of this theorem is to show that df-eu is not an extension of df-mo, but an alternative definition that could have been (and was in the past) picked as the first of the two.  The independence of the two concepts on one hand and how they are mutually related is given in dfmoeu (former dfmolem) and the new dfeumo for the reverse direction.  Putting them right in the beginning of the uniqueness section means that no result developed from df-mo and df-eu may later interfere with them.  They will stay fundamental for this section.  As a consequence the discouraged proof modification holding for dfmo makes now more sense for me, since a key dependence on dfmoeu (dfmolem) will not introduce other unwanted dependencies due to proof changes.


set.mm
```diff
@@ -22283,6 +22283,32 @@ The definition of substitution ( ~ df-sb ) used to be ~ dfsb1 .  These two
      at most one ` x ` such that ` ph ` "). $)
   wmo $a wff E* x ph $.
 
+  ${
+    $d x y $.  $d y ph $.
+    $( An elementary proof of ~ moeu in disguise, connecting an expression
+       characterizing uniqueness ( ~ df-mo ) to that of existential uniqueness
+       ( ~ eu6 ).  No particular order of definition is required, as one can be
+       derived form the other.  This is shown here and in ~ dfeumo .
```
icecream17, line 22291:
```suggestion
       derived from the other.  This is shown here and in ~ dfeumo .
```

set.mm
```diff
@@ -22283,6 +22283,32 @@ The definition of substitution ( ~ df-sb ) used to be ~ dfsb1 .  These two
...
+    dfmoeu $p |- ( ( E. x ph -> E. y A. x ( ph <-> x = y ) ) <->
+                      E. y A. x ( ph -> x = y ) ) $=
+      ( wex weq wb wal wi wn alnex pm2.21 alimi sylbir 19.8ad biimp eximi nfia1
+      ja wa com12 ax12v embantd spsd ancld albiim syl6ibr exlimi eximdv impbii
+      id ) ABDZABCEZFZBGZCDZHAULHZBGZCDZUKUOURUKIZUQCUSAIZBGUQABJUTUPBAULKLMNUN
+      UQCUMUPBAULOLPRUKURUOUKUQUNCAUQUNHBUPUMBQAUQUQULAHBGZSUNAUQVAAUPVABAAULVA
+      AUJULAVAABCUATUBUCUDAULBUEUFUGUHTUI $.
+
+    $( An elementary proof showing the reverse direction of ~ dfmoeu .  Here
+       the charactarizing expression of existential uniqueness ( ~ eu6 ) can be
```
icecream17, line 22302:
```suggestion
       the characterizing expression of existential uniqueness ( ~ eu6 ) can be
```

icecream17 (approved):
approve (edited)

benjub (approved):
That's a nice addition. Actually, it's even possible to put them earlier. At least, before `$c E* $. wmo $a wff E* x ph $.`. Then, either frankly earlier, before the section "Uniqueness and unique existence", or else, since they are "in diguise", as you write, about uniqueness and unique existence, between the section (`#*#*#*`) and the subsection (`=-=-=-`).

benjub (approved):

PR #3556, opened 2023-10-04
Expand proposed changes to *syl5* theorems

jkingdon:
Apparently I'm a glutton for punishment with the renames and haven't had enough yet.

The intention here is to start with Norm's brief suggestion regarding syl5 and make it more specific.  As can be seen from the "compare to" lines in the proposal, many of the new names are close parallels to existing names, so in that sense this is just the next step in a renaming project which was begun years ago. Also includes renames to bitr3i , bitr3d , eqtr3i , eqtr3d , eqsstr3i , and eqsstr3d as it seemed awkward to not rename these at the same time.

Although these renames seem a bit scary because almost all of these theorems are widely used, the number of theorems being renamed is significantly less than for the `mpt2` to `mpo` renames which were recently completed.

If this pull request is accepted my plan would be to proceed with a large number of small(sih) pull requests, perhaps one for each theorem being renamed. I would rename each theorem in iset.mm and set.mm at the same time (I think all of these are already in iset.mm although I plan to compare more carefully as I go).

There are a few references in web pages to update (mmset.raw.html) but not many.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -18,8 +18,42 @@ proposed  syl       imtri       (analogous to *bitr*, sstri, etc.)
 proposed  sylib     imbitri     etc.
 proposed  sylbid    biimtrd     etc.
 proposed  sylbird   biimtrrd    etc.
-proposed  syl5*     *trid       (syl5bi -> biimtrid; syl5eqel -> eqeltrid;etc.)
-proposed  syl6*     *trdi
+proposed  syl5com   imtridcom
+proposed  syl5      imtrid
+proposed  syl56     imtridi
+proposed  syl5d     imtridd
+proposed  syl5bi    biimtrid
+proposed  syl5bir   biimtrrid
+proposed  syl5ib    imbitrid
+proposed  syl5ibcom imbitridcom
+proposed  syl5ibr   imbitrrid
+proposed  syl5ibrcom imbitrridcom
+proposed  syl5bb    bitrid      compare to bitri or bitrd
+proposed  syl5rbb   bitridcom
+proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
+proposed  syl5rbbr  bitr3idcom
+proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
+proposed  syl5req   eqtridcom
+proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
+proposed  syl5reqr  eqtr3idcom
+proposed  syl5eqel  eqeltrid    compare to eqeltri or eqeltrd
+proposed  syl5eqelr eqeltrrid   compare to eqeltrri or eqeltrrd
+proposed  syl5eleq  eleqtrid    compare to eleqtri or eleqtrd
+proposed  syl5eleqr eleqtrrid   compare to eleqtrri or eleqtrrd
+proposed  syl5eqner eqnetrrid   compare to eqnetrri or eqnetrrd
+proposed  syl5ss    sstrid      compare to sstri or sstrd
+proposed  syl5eqss  eqsstrid    compare to eqsstri or eqsstrd
+proposed  syl5eqssr eqsstrrid   compare to eqsstr3i or eqsstr3d
+proposed  eqsstr3i  eqsstrri
+proposed  eqsstr3d  eqsstrrd
+proposed  syl5sseq  sseqtrid    compare to sseqtri or sseqtrd
+proposed  syl5sseqr sseqtrrid
+proposed  syl5eqbr  eqbrtrid    compare to eqbrtri or eqbrtrd
+proposed  syl5eqbrr eqbrtrrid   compare to eqbrtrri or eqbrtrrd
+proposed  syl5breq  breqtrid    compare to breqtri or breqtrd
+proposed  syl5breqr breqtrrid   compare to breqtrri or breqtrrd
+proposed  wl-luk-syl5 wl-luk-imtrid
+proposed  syl6*     *trdi       analogous to *trid theorems
 (Please send any comments on these proposals to the mailing list or
 make a github issue.)
 


icecream17:
I'm not sure that the proposed changes are better. What does `i` and `d` stand for?

(EDIT: mpid is deduction of mpi, mpdi is a1i and mpdd; but this doesn't help because neither syl5 or syl6 is really a deduction)

<del>(EDIT2: I think it goes like this: naming the longest hypothesis `d` and the smaller hypothesis `i`, we have `mptr` (i.e. syl) with `i` `d` or `d` `i`)</del> (EDIT3: correct but see jkingdon's later comment)

icecream17:
bitr3i, bitr3d, eqtr3i, etc are part of a chain: bitri, bitr2i, bitr3i, bitr4i and changing just one of the chain is inconsistent

jkingdon:
> bitr3i, bitr3d, eqtr3i, etc are part of a chain: bitri, bitr2i, bitr3i, bitr4i and changing just one of the chain is inconsistent

Hmm, this is a bigger problem than I saw at first. In a case like eqeltri we have the base theorem here:

```
show statement eqeltri
8527 eqeltri.1 $e |- A = B $.
8528 eqeltri.2 $e |- B e. C $.
8529 eqeltri $p |- A e. C $= ... $.
```

and then we add an `r` to flip the equality:

```
show statement eqeltrri
8532 eqeltrri.1 $e |- A = B $.
8533 eqeltrri.2 $e |- A e. C $.
8534 eqeltrri $p |- B e. C $= ... $.
```

But what if the base theorem has two formulas which are symmetrical? Such as:

```
show statement bitri
989 bitri.1 $e |- ( ph <-> ps ) $.
990 bitri.2 $e |- ( ps <-> ch ) $.
991 bitri $p |- ( ph <-> ch ) $= ... $.
```

then we can either flip the first one:

```
show statement bitr3i
999 bitr3i.1 $e |- ( ps <-> ph ) $.
1000 bitr3i.2 $e |- ( ps <-> ch ) $.
1001 bitr3i $p |- ( ph <-> ch ) $= ... $.
```

or the second one:

```
show statement bitr4i
1004 bitr4i.1 $e |- ( ph <-> ps ) $.
1005 bitr4i.2 $e |- ( ch <-> ps ) $.
1006 bitr4i $p |- ( ph <-> ch ) $= ... $.
```

This is not amenable to adding an `r`.  I think the best solution here will be to bring back the `3` and `4`. For example,
`syl5eqr` would be renamed to `eqtr3id` and `syl6eqr` would be renamed to `eqtr4di`.

jkingdon:
> I'm not sure that the proposed changes are better. What does `i` and `d` stand for?

Let me try to explain in more detail, since Norm just left us `ongoing project to improve naming consistency` and `proposed syl5* *trid (syl5bi -> biimtrid; syl5eqel -> eqeltrid;etc.)` and it is up to us to decipher that (or come up with something even better).

We are used to a family of transitivity theorems with the abbreviation of the first hypothesis, the abbreviation of the second hypothesis, `tr` to show it is one of these transitivity theorems, and then `i` or `d`. For example,

```
eqbrtr.1 $e |- A = B $.
eqbrtr.2 $e |- B R C $.
eqbrtri $p |- A R C $= ... $.
```

and

```
eqbrtrd.1 $e |- ( ph -> A = B ) $.
eqbrtrd.2 $e |- ( ph -> B R C ) $.
eqbrtrd $p |- ( ph -> A R C ) $= ... $.
```

Now, what do we call it if one of the hypotheses is in deduction form and the other hypothesis is in inference form? For example:

```
name-me.1 $e |- A = B $.
name-me.2 $e |- ( ph -> B R C ) $.
name-me $p |- ( ph -> A R C ) $= ... $.
```

or

```
name-me-too.1 $e |- ( ph -> A = B ) $.
name-me-too.2 $e |- B R C $.
name-me-too $p |- ( ph -> A R C ) $= ... $.
```

Norm's proposal is that `i` followed by `d`, or `d` followed by `i` refers to the two hypotheses. So that `name-me` would be called `eqbrtrid` and `name-me-too` would be called `eqbrtrdi`.

> 
> (EDIT: mpid is deduction of mpi, mpdi is a1i and mpdd; but this doesn't help because neither syl5 or syl6 is really a deduction)

In this example, `mpid` means to first apply the `i` as in `mpi`, then apply `d` to that. But none of the repeated `i`/`d` proposed as `syl5` and `syl6` replacements work that way; they apply to respective hypotheses as described above.

Perhaps by starting with `syl5` I started with the hardest to understand ones. We could instead start with `syl` which Norm proposes would be `imtri` (pretty obvious analogue to `bitri`), `sylib` which Norm proposes would be `imbitri` (again a pretty straightforward application of the pattern), and two more examples he gives at the top of `changes-set.txt`. I don't have an especially strong opinion about which one to start with, so speak up if people would prefer a pull request fleshing out renames for `syl` and related theorems, before thinking too hard about the `syl5` ones.

icecream17:
I see. On principle this change makes sense even though I am quite fond of `syl`

I guess this change is more complicated because it's `syl5` --> `sylid` --> `imtrid` and `syl6` --> `syldi` --> `imtrdi` (i.e. two changes at once)

jkingdon:
> I think the best solution here will be to bring back the `3` and `4`. For example,
> `syl5eqr` would be renamed to `eqtr3id`

I have updated the pull request with this change.

jkingdon:
> I see. On principle this change makes sense even though I am quite fond of `syl`

That was very much my first reaction but the names built on `tr` have been growing on me. I guess partly because I've been using the ones which are already named that way and getting to know the naming scheme.

We can mull it over and see what others think. We aren't _forced_ to rename although I guess my take is if we do find names which we are pretty sure are better, we should accept the short term dislocation and get used to them.

> I guess this change is more complicated because it's `syl5` --> `sylid` --> `imtrid` and `syl6` --> `syldi` --> `imtrdi` (i.e. two changes at once)

I hadn't really thought of that, but I guess that's a fair way to describe it.

jkingdon:
Also see mailing list message at https://groups.google.com/g/metamath/c/TwiY90i95zc (subject line "naming of syl theorems" date 7 Oct 2023).

benjub:
It's a bit strange to call "transitivity" a statement that asserts substitutability of equal variables in a given function or predicate, or equivalently, that equality is a congruence relation for the considered function or predicate (e.g., eqeltri above). I know that it's the current situation and I do not propose to change these labels.

One relabeling which I think would make labels more logical is the number prefix in 3syl or 3bitr* (and others) which should be 2syl or 2bitr* since they amount to applying syl or bitr* twice. With this new labeling, applying mxxx then nxxx amounts to applying (m+n)xxx. This is the classical problem of "posts and intervals".

jkingdon:
> It's a bit strange to call "transitivity" a statement that asserts substitutability of equal variables in a given function or predicate

Hmm, so the concern is about using the `tr` naming convention (I don't much care whether it stands for transitivity or is just a more arbitrary collection of two letters by now) both for things which are symmetrical, like `eq`, `bi`, `ne`, or `en` and for things which are not, like `im`, `el`, `ss`, `lt`, `le`? I guess it is more food for thought than anything else, given that the suggestion is not to change this usage at this time.

> One relabeling which I think would make labels more logical is the number prefix in 3syl or 3bitr* (and others) which should be 2syl or 2bitr* since they amount to applying syl or bitr* twice. With this new labeling, applying mxxx then nxxx amounts to applying (m+n)xxx. This is the classical problem of "posts and intervals".

I have thoughts on this (well at least the `3syl` one) but I'm not sure whether to get into it now or start with the `syl5` renames in this pull request.

What the heck, I might as well write down my idea given the disclaimer that I haven't spent as much time with it as I have with the ones now named with `syl5`. If we have:

`imtri` (renamed from `syl`):

```
imtri.1 $e |- ( ph -> ps ) $.
imtri.2 $e |- ( ps -> ch ) $.
imtri $p |- ( ph -> ch ) $= ... $.
```

`imtrid` (renamed from `syl5`):

```
imtrid.1 $e |- ( ph -> ps ) $.
imtrid.2 $e |- ( ch -> ( ps -> th ) ) $.
imtrid $p |- ( ch -> ( ph -> th ) ) $= ... $.
````

`imtrdi` (renamed from `syl6`):

```
imtrdi.1 $e |- ( ph -> ( ps -> ch ) ) $.
imtrdi.2 $e |- ( ch -> th ) $.
imtrdi $p |- ( ph -> ( ps -> th ) ) $= ... $.
```

and then we want a name for the thing currently called `3syl`:

```
name-me.1 $e |- ( ph -> ps ) $.
name-me.2 $e |- ( ps -> ch ) $.
name-me.3 $e |- ( ch -> th ) $.
name-me $p |- ( ph -> th ) $= ... $.
```

then what about `imtriii`? It would also allow things like `imtridi`:

```
imtridi.1 $e |- ( ph -> ps ) $.
imtridi.2 $e |- ( et -> ( ps -> ch ) ) $.
imtridi.3 $e |- ( ch -> th ) $.
imtridi $p |- ( et -> ( ph -> th ) )  $= ... $.
```

It isn't a perfect pattern - when we would have two `i` or two `d` they get collapsed into a single one (`bitri` not `bitrii` and `bitrd` not `bitrdd`), but when there are three then they don't collapse (because we need a way to distinguish `imtri` from `imtriii`. Also, I didn't start looking through whether this would provide a way to rename anything other than `3syl` (for what it is worth, the current pull request, or Norm's suggestions in `changes-set.txt`, don't go beyond theorems whose names now contain `syl`, so from that point of view whether to do anything to `3bitri` and the like is perhaps better considered later).

Anyway, I might be rambling a bit far afield from the topic of whether to rename the theorems which now have `syl5` in their names. But I thought I'd put it here in case it helps anyone think through any of these names.

icecream17:
`imtriii` seems inconsistent with redivcli and other theorems that legitimately have 3 hypotheses in inference form. logically there would also be `ddd`. For theorems with even more inference or deductive hypotheses this seems like a lot of `i` or `d`.

Incidentally, syl5 and syl6 kinda fit into a pattern with syl7 and syl8. No idea about syl9 and syl10 though. The latter might be ad hoc like sb1 sb2 sb3 sb4(b) sb5 sb6 dfsb7

jkingdon:
Thanks for all the thoughts. I added the rest of the proposed renames at #3569 and also some thoughts about what wouldn't be part of this set of renames there, so that pull request replaces this one.

PR #3557, opened 2023-10-04
Moved theorems in earlier sections

GinoGiotto:
I have relocated the remaining theorems from the ax-13 section that do not rely on ax-13. After reviewing the list generated by `metamath-knife set.mm --axiom-use`, I believe there are no more theorems in `Axiom scheme ax-13 (Quantified Equality)` that can currently be moved to earlier sections.

I only left out [pm11.07](https://us.metamath.org/mpeuni/pm11.07.html), since it doesn't use any axioms and I don't know how to handle it.


icecream17:
I guess pm11.07 is supposed to be right after sbcom4, and both of these would fit right after sbv (to match the axiom usage better)

spsbimOLD is a copy of spsbim and can be removed

and then move sbal, sbalv after alrot4 to preserve the chain
image

icecream17 (approved):

avekens (approved):

PR #3559, opened 2023-10-05
Prove zltp1ne and variants

BTernaryTau:
Ordering relation for ZZ, NN, and NN0:

> ( ( A + 1 ) < B <-> ( A < B /\ B =/= ( A + 1 ) ) )

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,8 +23,11 @@ Walking through the
 To create a proof, you need to pick a tool.
 The Metamath system specifies a common file format, not a single tool
 everyone has to use.
-Many in the community use the [mmj2 tool](http://us.metamath.org#book), and
+Many in the community use the [mmj2 tool](http://us.metamath.org#mmj2), and
 many of the rest use the original metamath C program (aka "metamath.exe").
+A third option is the [metamath-lamp tool](https://lamp-guide.metamath.org/),
+which can be accessed directly from your web browser at
+<https://expln.github.io/lamp/latest/index.html>.
 The page
 [Metamath Proof Assistants](https://github.com/metamath/set.mm/wiki/Metamath-Proof-Assistants)
 provides more information about each tool.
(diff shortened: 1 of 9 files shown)


set.mm
```diff
@@ -501626,6 +501626,27 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
...
+  $( Integer ordering relation.  (Contributed by BTernaryTau, 24-Sep-2023.) $)
+  zltp1ne $p |- ( ( A e. ZZ /\ B e. ZZ ) ->
+        ( ( A + 1 ) < B <-> ( A < B /\ B =/= ( A + 1 ) ) ) ) $=
+    ( cz wcel wa c1 caddc co clt wbr cle wne cr zre peano2re ltlen sylan syl2an
+    wb zltp1le anbi1d bitr4d ) ACDZBCDZEZAFGHZBIJZUFBKJZBUFLZEZABIJZUIEUCAMDZBM
+    DZUGUJSZUDANBNULUFMDUMUNAOUFBPQRUEUKUHUIABTUAUB $.
+
+  $( Positive integer ordering relation.  (Contributed by BTernaryTau,
+     24-Sep-2023.) $)
+  nnltp1ne $p |- ( ( A e. NN /\ B e. NN ) ->
```
jkingdon, line 501638: Not a blocker (especially for a mathbox) but unless the version for `NN` or `NN0` is going to be widely used, I would tend to supply just the version for `ZZ` and then use `nn0zd` and similar theorems where it is applied.

jkingdon (approved):

icecream17 (approved):

set.mm
BTernaryTau: That's fair. IIRC I was trying to follow the example of similar groups of theorems like *ltp1le and *ltlem1 which exist for all 3.

jkingdon: The closest thing I can give you to a consistent guideline is what I said before about what is widely used versus what is only used a few places. But I'm not completely sure how much we follow that.

PR #3561, opened 2023-10-05
Prove usgrcyclgt2v

BTernaryTau:
usgrgt2cycl

> A non-trivial cycle in a simple graph has a length greater than 2.

f1resfz0f1d

> If a function with a sequence of nonnegative integers (starting at 0) as its domain is one-to-one when 0 is removed, and if the range of that restriction does not contain the function's value at the removed integer, then the function is itself one-to-one.

0nn0m1nnn0

> A number is zero if and only if it's a nonnegative integer that becomes negative after subtracting 1.

pthhashvtx

> A graph containing a path has at least as many vertices as there are edges in the path.

usgrcyclgt2v

> A simple graph with a non-trivial cycle must have at least 3 vertices.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,8 +23,11 @@ Walking through the
 To create a proof, you need to pick a tool.
 The Metamath system specifies a common file format, not a single tool
 everyone has to use.
-Many in the community use the [mmj2 tool](http://us.metamath.org#book), and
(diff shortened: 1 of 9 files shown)


avekens (approved):
Nice results for Graph Theory. I'll have a closer look at the proofs as soon as these theorems are available on the web site.

PR #3562, opened 2023-10-06
Reduce ax-13 usage with bj-cbv1v

GinoGiotto:
* Moved `bj-cbv1v` from @benjub's mathbox to main.

* Renamed `bj-cbv1v` into `cbv1v`.

* Removed discouragement tag from `cbv1v`.

* Use `cbv1v` to drop ax-13 from ~ vtoclgft (which automatically drops ax-13 for ~ vtocldf, ~ vtocld and ~ bj-vtoclgfALT as well).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18383,7 +18383,6 @@ Proof modification of "bj-bijust0ALT" is discouraged (6 steps).
 Proof modification of "bj-biorfi" is discouraged (12 steps).
 Proof modification of "bj-brrelex12ALT" is discouraged (66 steps).
 Proof modification of "bj-cbv1hv" is discouraged (62 steps).
-Proof modification of "bj-cbv1v" is discouraged (66 steps).
 Proof modification of "bj-cbv2hv" is discouraged (67 steps).
 Proof modification of "bj-cbv2v" is discouraged (47 steps).
 Proof modification of "bj-cbv3hv2" is discouraged (10 steps).
--- a/discouraged
+++ b/discouraged
@@ -18044,6 +18044,7 @@ New usage of "vsfval" is discouraged (4 uses).
 New usage of "vtocl2OLD" is discouraged (0 uses).
 New usage of "vtocl2gOLD" is discouraged (0 uses).
 New usage of "vtoclALT" is discouraged (0 uses).
+New usage of "vtoclgftOLD" is discouraged (0 uses).
 New usage of "vtxdusgr0edgnelALT" is discouraged (0 uses).
 New usage of "w-bnj13" is discouraged (5 uses).
 New usage of "w-bnj15" is discouraged (92 uses).
@@ -19896,6 +19897,7 @@ Proof modification of "vk15.4jVD" is discouraged (268 steps).
 Proof modification of "vtocl2OLD" is discouraged (84 steps).
 Proof modification of "vtocl2gOLD" is discouraged (28 steps).
 Proof modification of "vtoclALT" is discouraged (11 steps).
+Proof modification of "vtoclgftOLD" is discouraged (142 steps).
 Proof modification of "vtxdusgr0edgnelALT" is discouraged (94 steps).
 Proof modification of "wfrlem4OLD" is discouraged (543 steps).
 Proof modification of "wl-cases2-dnf" is discouraged (85 steps).


GinoGiotto:
If the new proof of ~ vtoclgft is too long I can add some lemmas with $d conditions to bring it back to its original size.

icecream17 (approved):
approve

(Now that I've reread conventions-labels, conventions-labels says that v means removing a not-free hypothesis, and no hypothesis is removed here. But adding v to add a dv condition is also quite common a la equsb1v and equsb1. So either conventions-labels is outdated or all these labels are inconsistent and should use "w" instead. Something for another pr)

GinoGiotto:
>(Now that I've reread conventions-labels, conventions-labels says that v means removing a not-free hypothesis, and no hypothesis is removed here. But adding v to add a dv condition is also quite common a la equsb1v and equsb1. So either conventions-labels is outdated or all these labels are inconsistent and should use "w" instead. Something for another pr)

I fully agree with everything said here.

Here is the relevant snippet from convention-labels:

> Suffixes. Suffixes are used to indicate the form of a theorem (see above). Additionally, we sometimes suffix with "v" the label of a theorem eliminating a hypothesis such as Ⅎ𝑥𝜑 in [19.21](https://us.metamath.org/mpeuni/19.21.html) 2137 via the use of disjoint variable conditions combined with [nfv](https://us.metamath.org/mpeuni/nfv.html) 1874. If two (or three) such hypotheses are eliminated, the suffix "vv" resp. "vvv" is used, e.g. [exlimivv](https://us.metamath.org/mpeuni/exlimivv.html) 1892.

This actually confused me while I was making https://github.com/metamath/set.mm/pull/3551. Theorem [chvarv](https://us.metamath.org/mpeuni/chvarv.html) replaces a not-free hypothesis from [chvar](https://us.metamath.org/mpeuni/chvar.html) and so it correctly adds a `v` suffix as the conventions would suggest. But [bj-chvarv](https://us.metamath.org/mpeuni/bj-chvarv.html) is just [chvar](https://us.metamath.org/mpeuni/chvar.html) with an additional (weakening) $d condition. So since "chvarv" was already taken, I wasn't really sure how to name [bj-chvarv](https://us.metamath.org/mpeuni/bj-chvarv.html). I chose `chvarfv` because it has an additional not-free hypothesis compared to [chvarv](https://us.metamath.org/mpeuni/chvarv.html), but that looks odd because that same not-free hypotesis is already present in [chvar](https://us.metamath.org/mpeuni/chvar.html). So maybe [bj-chvarv](https://us.metamath.org/mpeuni/bj-chvarv.html) would be more appropriately renamed as `chvarw`.

benjub:
If a theorem is obtained by adding a DV condition to Theorem ~xxx, then its label should be ~xxxv, at least when there is no label-collision. If indeed often happens that adding a DV condition allows to remove a nonfreeness hypothesis (but that is secondary: the general rule is: if you add a D**V** condition, then you suffix add ~a **d*~ a **v* to the label). I'm going to make this more precise in said convention.

In the present case, there is a label collision, so we have to find another one. You proposal ~chvarfv seems ok.

set.mm
```diff
@@ -31614,12 +31628,15 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
...
-      ABMVCABUJUKULTUOUTVMVKLUSVCBCUMUPSTUQUR $.
+      ( vz wa cv wceq wb wi wal wex wn weq biimp cbv1v alnex syl5ib imp wnf w3a
+      wnfc wcel elisset nfv nfnfc1 nfcvd nfeqd nfnd nfvd eqeq1 a1i notbi syl6ib
+      id syl6 equcomi biimpr syl56 impbid 3bitr3g con4bid ad2antrr 3impia com23
+      imim2i alanimi 19.23t adantl 3adant3 mpd ) CDUCZBCUAZGZCHZDIZABJZKZCLACLG
+      ZDEUDZUBVQCMZBVOVTWAWBVMWAWBKVNVTWAFHZDIZFMZVMWBFDEUEVMWEWBVMWDNZFLZVQNZC
+      LZWENWBNVMWGWIVMWFWHFCVMFUFZCDUGZVMWDCVMCWCDVMCWCUHVMUPUIUJZVMWHFUKZVMFCO
+      ZWFWHJZWFWHKVMWNWDVQJZWOWNWPKVMWCVPDULUMWDVQUNUOZWFWHPUQQVMWHWFCFWKWJWMWL
+      CFOWNVMWOWHWFKCFURWQWFWHUSUTQVAWDFRVQCRVBVCSVDVEVOVTWBBKZWAVOVTWRVTVQBKZC
+      LZVOWRVSAWSCVSAWSVSVQABVRABKVQABPVGVFTVHVNWTWRJVMVQBCVIVJSTVKVL $.
```
benjub, line 31639: Can you keep the old proof as *OLD ?

benjub (approved):

tirix:
> if you add a D**V** condition, then you suffix a *_d_ to the label

You mean then you suffix a *_v_ to the label, right?

PR #3568, opened 2023-10-11
Prove spthcycl

BTernaryTau:
> A walk is a trivial path if and only if it is both a simple path and a cycle.


tirix (approved):
Looks good!
I would be good to wait for @avekens's approval too, though.

avekens (approved):

icecream17 (approved):

PR #3569, opened 2023-10-11
propose renames for some syl theorems

jkingdon:
Thanks for everyone who offered thoughts at https://github.com/metamath/set.mm/pull/3556 . I'm not sure I got a really clear sense there about how much support there is for the idea but I guess it is at least worth another round of refining what we might change.

This is a set of renames proposed by Norm some years ago and is to extend the pattern we are already using for `bitri`, `eqbrtri`, `eleqtrrd`, and many others. These renames seem scary because most of the theorems to be renamed are used a lot and for many of us the names are familiar, but the potential benefit is more natural names.

On looking at the topic, including the discussion so far, my thinking is to not take on too much.  For example, this pull request does not propose any renames for `3syl` , `syl7` , `syl9` , `nsyl` , and various others. Our goal here is "an ongoing project to improve naming consistency" (in the words of the preamble to changes-set.txt) and in cases where the new names aren't really jumping out at us, I'd say we can just keep existing names unless/until we think of names which seem to be making more sense than what we have.

The current pull request fleshes out all the proposed renames sketched out by Norm, which cover basic uses of `syl` (`syl` itself, `sylib`, etc), most theorems with `syl5` in the name, and most theorems with `syl6` in the name.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -15,6 +15,8 @@ The ones WITH "Notes" may have to be processed manually.
 PROPOSED:
 Date      Old       New         Notes
 proposed  syl       imtri       (analogous to *bitr*, sstri, etc.)
+                                there is less agreemnt on renaming syl
+                                than others here
 proposed  syld      imtrd
 proposed  syldc     imtrdcom
 proposed  syldd     imtrdd
@@ -31,7 +33,7 @@ proposed  sylbid    biimtrd
 proposed  sylibrd   imbitrrd
 proposed  sylbird   biimtrrd
 proposed  syl5com   imtridcom
-proposed  syl5      imtrid
+proposed  syl5      imtrid      alternate proposal: sylid
 proposed  syl56     imtridi
 proposed  syl5d     imtridd
 proposed  syl5bi    biimtrid
@@ -65,7 +67,7 @@ proposed  syl5eqbrr eqbrtrrid   compare to eqbrtrri or eqbrtrrd
 proposed  syl5breq  breqtrid    compare to breqtri or breqtrd
 proposed  syl5breqr breqtrrid   compare to breqtrri or breqtrrd
 proposed  wl-luk-syl5 wl-luk-imtrid
-proposed  syl6      imtrdi
+proposed  syl6      imtrdi      alternate proposal: syldi
 proposed  syl6com   imtrdicom
 proposed  syl6d     imtrdid
 proposed  syl6ib    imbitrdi
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -15,9 +15,21 @@ The ones WITH "Notes" may have to be processed manually.
 PROPOSED:
 Date      Old       New         Notes
 proposed  syl       imtri       (analogous to *bitr*, sstri, etc.)
-proposed  sylib     imbitri     etc.
-proposed  sylbid    biimtrd     etc.
-proposed  sylbird   biimtrrd    etc.
+proposed  syld      imtrd
+proposed  syldc     imtrdcom
+proposed  syldd     imtrdd
+proposed  sylbi     biimtri
+proposed  sylib     imbitri
+proposed  sylbb     bitriim
+proposed  sylibr    imbitrri
+proposed  sylbir    biimtrri
+proposed  sylbbr    bitrriim
+proposed  sylbb1    bitr3iim
+proposed  sylbb2    bitr4iim
+proposed  sylibd    imbitrd
+proposed  sylbid    biimtrd
+proposed  sylibrd   imbitrrd
+proposed  sylbird   biimtrrd
 proposed  syl5com   imtridcom
 proposed  syl5      imtrid
 proposed  syl56     imtridi
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -65,7 +65,35 @@ proposed  syl5eqbrr eqbrtrrid   compare to eqbrtrri or eqbrtrrd
 proposed  syl5breq  breqtrid    compare to breqtri or breqtrd
 proposed  syl5breqr breqtrrid   compare to breqtrri or breqtrrd
 proposed  wl-luk-syl5 wl-luk-imtrid
-proposed  syl6*     *trdi       analogous to *trid theorems
+proposed  syl6      imtrdi
+proposed  syl6com   imtrdicom
+proposed  syl6d     imtrdid
+proposed  syl6ib    imbitrdi
+proposed  syl6ibr   imbitrrdi
+proposed  syl6bi    biimtrdi
+proposed  syl6bir   biimtrrdi
+proposed  syl6bb    bitrdi      compare to bitri or bitrd
+proposed  syl6rbb   bitrdicom
+proposed  syl6bbr   bitr4di     compare to bitr4i or bitr4d
+proposed  syl6rbbr  bitr4dicom
+proposed  syl6eq    eqtrdi      compare to eqtri or eqtrd
+proposed  syl6req   eqtrdicom
+proposed  syl6eqr   eqtr4di     compare to eqtr4i or eqtr4d
+proposed  syl6reqr  eqtr4dicom
+proposed  syl6eqel  eqeltrdi    compare to eqeltri or eqeltrd
+proposed  syl6eqelr eqeltrrdi   compare to eqeltrri or eqeltrrd
+proposed  syl6eleq  eleqtrdi    compare to eleqtri or eleqtrd
+proposed  syl6eleqr eleqtrrdi   compare to eleqtrri or eleqtrrd
+proposed  syl6ss    sstrdi      compare to sstri or sstrd
+proposed  syl6sseq  sseqtrdi    compare to sseqtri or sseqtrd
+proposed  syl6sseqr sseqtrrdi
+proposed  syl6eqss  eqsstrdi    compare to eqsstri or eqsstrd
+proposed  syl6eqssr eqsstrrdi
+proposed  syl6eqbr  eqbrtrdi    compare to eqbrtri or eqbrtrd
+proposed  syl6eqbrr eqbrtrrdi   compare to eqbrtrri or eqbrtrrd
+proposed  syl6breq  breqtrdi    compare to breqtri or breqtrd
+proposed  syl6breqr breqtrrdi   compare to breqtrri or breqtrrd
+proposed  wl-luk-syl6 wl-luk-imtrdi
 (Please send any comments on these proposals to the mailing list or
 make a github issue.)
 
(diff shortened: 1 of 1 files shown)


tirix:
Those theorem names are much easier to understand and remember, so I believe it is a step in the right direction overall.
For `syl` I think it would be worth nothing that it used to be called `syl` which stood for "syllogism" if we change it.

IMO the fact the `syl` name is very short is good for two reasons:
- it's quick to type, which is important if you type it often (and that's my case)
- it keeps the database (somewhat) shorter. `imtri` adds 2 bytes per usage, and there are more than 2000 of them.

Maybe we could consider those changes, except for `syl`?

benjub:
Actually, "syl" classically denotes the closed form `|- (A -> B) -> (B -> C) -> A -> C`. So, not having a "syl" in the database that labels another statement is an additional advantage of using "imtri".

changes-set.txt
```diff
@@ -15,11 +15,85 @@ The ones WITH "Notes" may have to be processed manually.
...
+proposed  eqsstr3i  eqsstrri
+proposed  eqsstr3d  eqsstrrd
+proposed  syl5sseq  sseqtrid    compare to sseqtri or sseqtrd
+proposed  syl5sseqr sseqtrrid
+proposed  syl5eqbr  eqbrtrid    compare to eqbrtri or eqbrtrd
+proposed  syl5eqbrr eqbrtrrid   compare to eqbrtrri or eqbrtrrd
+proposed  syl5breq  breqtrid    compare to breqtri or breqtrd
+proposed  syl5breqr breqtrrid   compare to breqtrri or breqtrrd
+proposed  wl-luk-syl5 wl-luk-imtrid
+proposed  syl6      imtrdi
```
benjub, line 68: I haven't followed closely, and this line is the first I picked at random, so that may have already been answered: this theorem is obtained from imtrd by strengthening (in the propcal sense, that is, using a1i) a hypothesis, but there are multiple hypotheses that could be strengthened too. There are multiple examples in set.mm where a number suffix is used, which would give here `imtrd2`. I haven't looked at the implications, that's just a remark.

icecream17: https://github.com/metamath/set.mm/pull/3556#issuecomment-1747781528

yeah here in imtrid (syl5) and imtrdi (syl6) the order of i and d refer to the order of which hypothesis is in inference vs deductive form

jkingdon:
> For `syl` I think it would be worth nothing that it used to be called `syl`

That's a good point. It is mentioned a number of places in the metamath book so this seems like a case where we'd want to note the rename in the comment as we did for https://us.metamath.org/mpeuni/trud.html . I did a search in the book for `syl` and I didn't see anything else obviously needing this treatment (there is some `show trace_back` output but I don't think there would be any expectation that would exact match a current set.mm).

changes-set.txt
benjub: Ok. We know that there is no ideal labeling, so maybe the one you propose is ok.

tirix:
> Actually, "syl" classically denotes the closed form `|- (A -> B) -> (B -> C) -> A -> C`. So, not having a "syl" in the database that labels another statement is an additional advantage of using "imtri".

It looks like that closed form would be [`pm3.33`](https://us.metamath.org/mpeuni/pm3.33.html).

benjub:
> > Actually, "syl" classically denotes the closed form `|- (A -> B) -> (B -> C) -> A -> C`. So, not having a "syl" in the database that labels another statement is an additional advantage of using "imtri".
> 
> It looks like that closed form would be [`pm3.33`](https://us.metamath.org/mpeuni/pm3.33.html).

It is actually https://us.metamath.org/mpeuni/imim1.html (pm3.33 is the uncurried form). It is also called "suffixing" and is the principal type of the combinator B'.

jkingdon:
Sounds like I should merge this pull request. At least that's the way I'm reading comments like "I believe it is a step in the right direction overall" and "On principle this change makes sense" but feel free to speak up if this seems like a bad idea or if there is some third option between "merge it" and "abandon it".

digama0:
Merging this is still a proposal though, right? It presumably doesn't matter much, it's just a way to keep track of proposed renames.

For myself, I have no issues with the majority of the renames. Especially for the more tenuously named things like `syl5eqssr` I think the new names are an improvement. I am less sure about the shorter ones though, I think that `syl` is a very useful and productive label segment and I don't think of these as transitivity lemmas so much as variations on `mp`. They are a bit in between, which is why I think having a dedicated label for them is useful. The main ones that I think deserve syl in the name are `syl`, `syl5` and `syl6`, and possibly a few others obviously derived from these. (I could do without the numbers though, what even happened to 1,2,3,4?)

jkingdon:
> Merging this is still a proposal though, right? It presumably doesn't matter much, it's just a way to keep track of proposed renames.

Treating it that way certainly occurred to me, although perhaps I'd like some vague idea of whether I should put in effort to starting doing some of the renames (ideas like "start with the more tenuously named ones" might be specific enough on that point).

> For myself, I have no issues with the majority of the renames. Especially for the more tenuously named things like `syl5eqssr` I think the new names are an improvement.

I'll take this (and other similar comments) as encouragement. Or mild encouragement anyway.

> I am less sure about the shorter ones though, I think that `syl` is a very useful and productive label segment and I don't think of these as transitivity lemmas so much as variations on `mp`. They are a bit in between, which is why I think having a dedicated label for them is useful. The main ones that I think deserve syl in the name are `syl`, `syl5` and `syl6`, and possibly a few others obviously derived from these. (I could do without the numbers though, what even happened to 1,2,3,4?)

I don't think I have a super strong opinion about `syl` itself although I'm a bit unclear about whether we are attached to it because it is kind of evocative and connects us to thousands of years of logic terminology and all that, or just because we are used to it. Maybe I'm not shocked by `imtri` just because I am so used to `bitri`, `entr`, `sstri`, etc.

For what it is worth, this proposal leaves plenty of names which contain `syl` in them (mostly because they don't obviously fit the `....tr[id]*` pattern).

As for `syl5` and `syl6`, I guess it is the numbers which mostly put me off those. Plus the new names are more evocative and easier to remember (at least if one buys into the `id` and `di` naming conventions and is already using a bunch of the other theorems using those conventions). 

> (I could do without the numbers though, what even happened to 1,2,3,4?)

Well, since you ask we can get an answer (or at least some information) from the git history.  The oldest version in git is what Norm called "stable release, archive name set.mm.1998-03-16-from-vax" and it contains:

```
show statement syl1
55 syl1 $p |- ( ( ph -> ps ) -> ( ( ch -> ph ) -> ( ch -> ps ) ) ) $= ... $.
show statement syl2
56 syl2 $p |- ( ( ph -> ps ) -> ( ( ps -> ch ) -> ( ph -> ch ) ) ) $= ... $.
show statement syl3
58 syl3.1 $e |- ( ph -> ps ) $.
59 syl3 $p |- ( ( ch -> ph ) -> ( ch -> ps ) ) $= ... $.
show statement syl4
62 syl4.1 $e |- ( ph -> ps ) $.
63 syl4 $p |- ( ( ps -> ch ) -> ( ph -> ch ) ) $= ... $.
```

Current set.mm still contains `syl7`, `syl8`, `syl9`, `syl10`, `syl11`, and `syl12` and those are among the ones I mentioned above as ones I'm not proposing any renames for.  This is partly because Norm didn't suggest anything in the proposed renames list, partly because I'm not sure what better names would be.

digama0:
> For what it is worth, this proposal leaves plenty of names which contain syl in them (mostly because they don't obviously fit the ....tr[id]* pattern).

I would say that if there are any `syl*` containing theorems at all remaining, then `syl` should stay named as such. It would be very confusing to only have variants and not have the original.

> As for syl5 and syl6, I guess it is the numbers which mostly put me off those. Plus the new names are more evocative and easier to remember (at least if one buys into the id and di naming conventions and is already using a bunch of the other theorems using those conventions).

I was thinking about suggesting something like `sylid` and `syldi` for these.

jkingdon:
> I was thinking about suggesting something like `sylid` and `syldi` for these.

I've pushed up a change to this pull request to add these to the list tagged with `alternate proposal`.

I also put some language around `syl` based on several people's doubts (I can't quite bear to delete Norm's words on this one - at least not yet - but we can revisit that one later).

changes-set.txt
```diff
@@ -15,11 +15,87 @@ The ones WITH "Notes" may have to be processed manually.
...
+proposed  sylbb1    bitr3iim
+proposed  sylbb2    bitr4iim
+proposed  sylibd    imbitrd
+proposed  sylbid    biimtrd
+proposed  sylibrd   imbitrrd
+proposed  sylbird   biimtrrd
+proposed  syl5com   imtridcom
+proposed  syl5      imtrid      alternate proposal: sylid
+proposed  syl56     imtridi
+proposed  syl5d     imtridd
```
icecream17, line 38: isn't this inconsistent with the interpretation of i and d representing hypotheses?

jkingdon: `sylid` and `syldi` aren't as systematic as `imtrid` and `imtrdi`, agreed.

That was my thinking in putting them down as alternative proposals rather than saying this one is settled.

jkingdon: Agreed, `syl5d` doesn't fit any of the patterns especially well. One question here is whether we are determined to get rid of the numbers (or at least the numbers `5` and `6`), which is mostly pretty doable if we adopt the `id` and `di` conventions, but does leave a few which are harder to name.

icecream17: Hmm. Maybe `d` is an operation that adds a deduction to all hypotheses.

Then for a specific hypothesis adding an antecedent would be `1` or `2`, i.e. the index of the hypothesis.
```
syl5    syl2
syl6    syl1
syl56   syl13 (hmm)
syl5d   syl2d
syl6d   syl1d

syl11   sylcom1 (syl5 with commuted hypothesis)
```
This is somewhat unsatisfying, as syl5 feels "earlier" than syl6 because in the deductive hypothesis, `ph -> ps -> ch`, syl5 replaces ps while syl6 replaces ch

jkingdon:
I'm going to take the lack of objection to "Sounds like I should merge this pull request" and the other positive comments as approval, and merge this (which, after all, does just update the proposed changes list).

Some of the proposed renames (those which replace "tenuous" names) can perhaps go in without much further discussion, others will need more attention before a rename has enough agreement.

I might not get around to doing anything else on this in the near future, so feel free to take up some part of it if you'd like.

changes-set.txt
jkingdon: It is an interesting thought process, but like the one which led me to wonder about `3syl` to `imtriii` as described at https://github.com/metamath/set.mm/pull/3556#issuecomment-1752200990 , I'm not sure it quite falls into place.

I think we can start with the easier cases and realize there are going to be some things which don't fit as readily into a systematic scheme. We know that our names will never be completely systematic (I mean, given our other goals like wanting them short and memorable too).

PR #3570, opened 2023-10-11
Define the class of acyclic graphs and prove some basic theorems about it

BTernaryTau:
I'm planning to eventually define classes for forests (simple acyclic graphs) and trees (connected forests) as well.

df-acycgr

> Define the class of all acyclic graphs.  A graph is called _acyclic_ if it has no (non-trivial) cycles.

dfacycgr1

> An alternate definition of the class of all acyclic graphs that requires all cycles to be trivial.

isacycgr and isacycgr1

> The property of being an acyclic graph.

acycgrcycl

> Any cycle in an acyclic graph is trivial (i.e. has one vertex and no edges).

This change also moves 2exanali out of Andrew Salmon's mathbox.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -25,6 +25,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
  6-Oct-23 vtocl2d   [same]      moved from TA's mathbox to main set.mm
  4-Oct-23 df-cda    df-dju      use |_| instead of +c
  4-Oct-23 cdaval    df-dju      use |_| instead of +c


set.mm
```diff
@@ -501775,6 +501788,75 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
...
+    acycgrcycl $p |- ( G e. AcyclicGraph ->
+              ( F ( Cycles ` G ) P -> F = (/) ) ) $=
+      ( vf vp cacycgr wcel ccycls cfv wbr c0 wceq wi wa cv cvv cwlks w3a adantl
+      wal cycliswlk syl simp2d simp3d breq1 eqeq1 imbi12d breq2 imbi1d sylan9bb
+      wlkv isacycgr1 ibi 19.21bbi adantr vtocl2d ex pm2.43d ) CFGZBACHIZJZBKLZU
+      SVAVAVBMZUSVANDOZEOZUTJZVDKLZMZVCDEBAPPVABPGZUSVACPGZVIAPGZVABACQIJVJVIVK
+      RABCUAABCUKUBZUCSVAVKUSVAVJVIVKVLUDSVDBLZVHBVEUTJZVBMVEALZVCVMVFVNVGVBVDB
+      VEUTUEVDBKUFUGVOVNVAVBVEABUTUHUIUJUSVHVAUSVHDEUSVHETDTDCFEULUMUNUOUPUQUR
+      $.
+  $}
```
tirix, line 501858: In general I prefer statements like
```
|- ( ( G e. AcyclicGraph /\ F ( Cycles ` G ) P ) -> F = (/) ) )
```
Rather than the double implication. I believe we have more theorems that can make use of this form, even if it's just an `~ imp` away.

set.mm
```diff
@@ -501775,6 +501788,75 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
...
+    $d f g p $.
+    $( An alternate definition of the class of all acyclic graphs that requires
+       all cycles to be trivial.  (Contributed by BTernaryTau, 11-Oct-2023.) $)
+    dfacycgr1 $p |- AcyclicGraph =
+              { g | A. f A. p ( f ( Cycles ` g ) p -> f = (/) ) } $=
+      ( cacycgr cv ccycls cfv wbr c0 wne wa wex cab wceq wal df-acycgr 2exanali
+      wn wi df-ne anbi2i 2exbii xchnxbir abbii eqtri ) DAEZCEBEFGHZUFIJZKZCLALZ
+      RZBMUGUFINZSCOAOZBMABCPUKUMBUGULRZKZCLALUMUJUGULACQUIUOACUHUNUGUFITUAUBUC
+      UDUE $.
+  $}
```
tirix, line 501821: These two theorems have the same distinct variable conditions, so maybe they could share the same `${`..`$}` scope.

tirix (approved):
Nice!
And I'm sure there's a ton of applications.
No major comment.

set.mm
```diff
@@ -501775,6 +501788,75 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
...
+  cacycgr $a class AcyclicGraph $.
+
+  ${
+    $d g f p $.
+    $( Define the class of all acyclic graphs.  A graph is called _acyclic_ if
+       it has no (non-trivial) cycles.  (Contributed by BTernaryTau,
+       11-Oct-2023.) $)
+    df-acycgr $a |- AcyclicGraph =
+              { g | -. E. f E. p ( f ( Cycles ` g ) p /\ f =/= (/) ) } $.
+  $}
```
avekens, line 501809: I think it is OK (and conforms with other definitions of special graphs, see, for example, "regular graphs" ~df-rgr) to define "acylic sets" in general ( `g` need not be a "graph" in this definition). If actually graphs are concerned, then it should/can be added in the antecedent, e.g.,: `( ( G e. USGraph /\ G e. AcyclicGraph ) -> ...` ("If `G` is an acyclic simple graph, then ...").

avekens (approved):

avekens:
> I'm planning to eventually define classes for forests (simple acyclic graphs) and trees (connected forests) as well.

Before introducing new definitions, I would like to see more (advanced) theorems for acyclic graphs.

set.mm
BTernaryTau: Done.

set.mm
BTernaryTau: I didn't realize this was thing. I've made the change for both pairs that share distinct variable conditions.

BTernaryTau:
> Before introducing new definitions, I would like to see more (advanced) theorems for acyclic graphs.

Of course. I was just mentioning my plans in case there were any objections that would impact the acyclic graph section.

set.mm
tirix: Thanks!
It's not mandatory and not enforced, it sometimes just helps to group things which belong together, and it's a bit more compact. Usually we do that more for actual `$e` hypotheses.

tirix:
It would be really nice to have things like:
- there exists of at most one path between two nodes in acyclic graphs, 
- given any connected graph, the existence of a spanning tree (i.e. a connected acyclic subtree)

...but I'm sure you already have them on your radar!

Issue #777, opened 2019-03-10
Fix elb4 (at least) to use [y/x] instead of [x/y]


david-a-wheeler:
Per Norm Megill in #775 ...

~~~~
Maybe fix elsb4 to swap them so the comment will be consistent. (I do it like this: Rename elsb4 to elsb4OLD everywhere, create and prove a new elsb4 that swaps x and y, use 'minimize elsb4/allow' on each proof using elsb4OLD, delete elsb4OLD, and add "(Revised by David A. Wheeler...)" to elsb4.)

[ y / x ] is almost always used except in a few places, and someday we should make them consistent:

MM> sea * '|- ( [ x / y
6103 equsb3lem $p |- ( [ x / y ] y = z <-> x = z )
6108 equsb3 $p |- ( [ x / y ] y = z <-> x = z )
6113 equsb3ALT $p |- ( [ x / y ] y = z <-> x = z )
6118 elsb3 $p |- ( [ x / y ] y e. z <-> x e. z )
6123 elsb4 $p |- ( [ x / y ] z e. y <-> z e. x )
7667 eqsb3lem $p |- ( [ x / y ] y = A <-> x = A )
7674 eqsb3 $p |- ( [ x / y ] y = A <-> x = A )
7681 clelsb3 $p |- ( [ x / y ] y e. A <-> x e. A )
9319 sbralie $p |- ( [ x / y ] A. x e. y ph <-> A. y e. x ps )
95855 clelsb3f $p |- ( [ x / y ] y e. A <-> x e. A )
109959 subsym1 $p |- ( [ x / y ] [ x / y ] F. -> [ x / y ] ph )
(the ones in my mathbox aren't important - they will be deleted someday)
~~~~

PR #3571, opened 2023-10-11
Replaced `|- ( [ x / y ]` with `|- ( [ y / x ]` in set.mm, iset.mm and nf.mm

GinoGiotto:
Towards https://github.com/metamath/set.mm/issues/777. I replaced all occurrences of  `|- ( [ x / y ]` with `|- ( [ y / x ]` in all databases that have them.

The only question I have is whether Norman Megill intended more generally to replace all occurrences of `[ x / y ]` or only the occurrences of `|- ( [ x / y ]`. Based on the search function he used https://github.com/metamath/set.mm/pull/775#issuecomment-471202097, it appears to be the latter.

In case he meant the first interpretation then I will continue with other PRs, in case he meant the second interpretation then this PR should suffice for closing #777.

### DIFF
--- a/nf.mm
+++ b/nf.mm
@@ -16746,39 +16746,39 @@ $)
   $}
 
   ${
-    $d y z $.  $d x y $.
+    $d x z $.  $d x y $.
     $( Lemma for ~ equsb3 .  (Contributed by Raph Levien and FL, 4-Dec-2005.)
        (Proof shortened by Andrew Salmon, 14-Jun-2011.) $)
-    equsb3lem $p |- ( [ x / y ] y = z <-> x = z ) $=
-      ( weq nfv equequ1 sbie ) BCDACDZBAHBEBACFG $.
+    equsb3lem $p |- ( [ y / x ] x = z <-> y = z ) $=
+      ( weq nfv equequ1 sbie ) ACDBCDZABHAEABCFG $.
   $}
 
   ${
-    $d w y z $.  $d w x $.
+    $d w x z $.  $d w y $.
     $( Substitution applied to an atomic wff.  (Contributed by Raph Levien and
        FL, 4-Dec-2005.) $)
-    equsb3 $p |- ( [ x / y ] y = z <-> x = z ) $=
-      ( vw weq wsb equsb3lem sbbii nfv sbco2 3bitr3i ) BCEZBDFZDAFDCEZDAFLBAFAC
-      EMNDADBCGHLBADLDIJADCGK $.
+    equsb3 $p |- ( [ y / x ] x = z <-> y = z ) $=
+      ( vw weq wsb equsb3lem sbbii nfv sbco2 3bitr3i ) ACEZADFZDBFDCEZDBFLABFBC
+      EMNDBADCGHLABDLDIJDBCGK $.
   $}
 
   ${
-    $d w y z $.  $d w x $.
+    $d w x z $.  $d w y $.
     $( Substitution applied to an atomic membership wff.  (Contributed by NM,
        7-Nov-2006.)  (Proof shortened by Andrew Salmon, 14-Jun-2011.) $)
-    elsb3 $p |- ( [ x / y ] y e. z <-> x e. z ) $=
-      ( vw wel wsb nfv sbco2 elequ1 sbie sbbii 3bitr3i ) DCEZDBFZBAFMDAFBCEZBAF
-      ACEZMDABMBGHNOBAMODBODGDBCIJKMPDAPDGDACIJL $.
+    elsb3 $p |- ( [ y / x ] x e. z <-> y e. z ) $=
+      ( vw wel wsb nfv sbco2 elequ1 sbie sbbii 3bitr3i ) DCEZDAFZABFMDBFACEZABF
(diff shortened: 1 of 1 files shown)


tirix (commented):
Ok. It's also more natural to me to think about substituting `x` with `y`, rather than the converse.

(Of course this does not change the definition)

tirix (approved):

iset.mm
```diff
@@ -22409,14 +22409,14 @@ deduction version (Contributed by Thierry Arnoux, 30-Jan-2017.) $)
...
     $( Implicit to explicit substitution that swaps variables in a quantified
        expression.  (Contributed by NM, 5-Sep-2004.) $)
-    sbralie $p |- ( [ x / y ] A. x e. y ph <-> A. y e. x ps ) $=
-      ( vz cv wral wsb cbvralsv sbbii nfv raleq sbie bitri sbco2 ralbii ) ACDGZ
-      HZDCIZACFIZFCGZHZBDUBHZTUAFRHZDCIUCSUEDCACFRJKUEUCDCUCDLUAFRUBMNOUCUAFDIZ
-      DUBHUDUAFDUBJUFBDUBUFACDIBACDFAFLPABCDBCLENOQOO $.
+    sbralie $p |- ( [ y / x ] A. y e. x ph <-> A. x e. y ps ) $=
+      ( vz cv wral wsb cbvralsv sbbii nfv raleq sbie bitri sbco2 equcoms ralbii
+      wb ) ADCGZHZCDIZADFIZFDGZHZBCUDHZUBUCFTHZCDIUEUAUGCDADFTJKUGUECDUECLUCFTU
+      DMNOUEUCFCIZCUDHUFUCFCUDJUHBCUDUHADCIBADCFAFLPABDCBDLABSCDEQNOROO $.
```
avekens, line 22419: Why does this proof become so much longer?

GinoGiotto: Because I made the cosmetic choice of keeping `x = y` in the hypothesis instead of switching into `y = x`. Do you prefer to have `y = x`, but with the original proof length? (For me it's fine both ways).

Some potentially useful data include the fact that currently `x = y` appears 486 times in iset.mm, while `y = x` appears only 19 times (and most of those 19 times are there to show commutation properties).

avekens: @GinoGiotto thank you for the explanation. We should not switch the antecedent in the hypothesis (`x = y ` should be the preferred way), so we can keep the longer proof.

avekens (approved):

jkingdon (approved):
As Norm said in the linked issue, "[ y / x ] is almost always used except in a few places, and someday we should make them consistent" so this is a good idea.

iset.mm
benjub: I don't really have an opinion about these renamings, as long as they are **only** renamings.
When we write `x = y -> ( ph <-> ps )`, this generally goes with DV(x,ps) and/or DV(y,ph), so if you swap x and y, do it consistently. (I haven't checked the present case.)

benjub:
This MR was **not** a mere renaming. In the case I checked (sbralie, from the above thread), there was more than a renaming, this is actually a different theorem, and the implicit substitution hypothesis is now written in an unusual way.

I think this should be fixed.

I haven't checked other cases.

iset.mm
GinoGiotto: > When we write `x = y -> ( ph <-> ps )`, this generally goes with DV(x,ps) and/or DV(y,ph).

I didn't know that x is coupled with ps and y is coupled with ph. I'll try to fix this in a next PR.

GinoGiotto:
> This MR was **not** a mere renaming. In the case I checked (sbralie, from the above thread), there was more than a renaming, this is actually a different theorem, and the implicit substitution hypothesis is now written in an unusual way.

Sbralie is actually the only statement for which I didn't make a trivial adjustment, because of the choice of keeping `x = y`. So unless the coupling issue mentioned before isn't the only problem, the only theorem to fix is sbralie.

benjub:
> > This MR was **not** a mere renaming. In the case I checked (sbralie, from the above thread), there was more than a renaming, this is actually a different theorem, and the implicit substitution hypothesis is now written in an unusual way.
> 
> Sbralie is actually the only statement for which I didn't make a trivial adjustment, because of the choice of keeping `x = y`. So unless the coupling issue mentioned before isn't the only problem, the only theorem to fix is sbralie.

Thanks for the checking.

iset.mm
avekens: I do not know what @benjub wants to say with his comment, and I don't see any need for a fix/change: DV(x,ps) means that x does not occur in ps, but of course y can (and usually should) occur in ps, analogously with ph. So we (usually) have ps(y) and ph(x). If the theorem is valid without DV conditions, they should not be added.

benjub: Of course, DV conditions should not be added if they are not necessary. What I mean is that when they are needed, it is usual to write that kind of hypotheses as `x = y -> (ph(x) <-> ps(y))` (or sometimes  `x = y -> (ph(x,y) <-> ps(y))` or `x = y -> (ph(x) <-> ps(x,y))`), where I made explicit the variables that may occur instead of the variables that do not.

PR #3572, opened 2023-10-15
Theorems for Godel-sets (2)

avekens:
To understand Mario's definitions about Godel-sets in more detail, I have proved some more theorems about them (still in MC's mathbox). Unfortunately, I got stuck with proving that ``( ( M Sat E ) ` N )`` is a function, as indicated in the definition ~df-sat:

> This recursive construction builds up a **function over wff codes**

Until now, I was able to prove this for the first level (basic formulas):

```
satfv0fun $p |- ( ( M e. V /\ E e. W ) -> Fun ( ( M Sat E ) ` (/) ) ) $=
```

Furthermore, the following theorems are already available:

```
satfrel $p |- ( ( M e. V /\ E e. W /\ N e. _om ) -> Rel ( ( M Sat E ) ` N ) ) $=
satfdmfmla $p |- ( ( M e. V /\ E e. W /\ N e. _om ) -> dom ( ( M Sat E ) ` N ) = ( Fmla ` N ) ) $=
satfrnmapom $p |- ( ( M e. V /\ E e. W /\ N e. _om ) -> ran ( ( M Sat E ) ` N ) C_ ~P ( M ^m _om ) ) $=
```

If ``Fun( ( M Sat E ) ` N )`` was proven, we would immediately have:

```
satff $p |- ( ( M e. V /\ E e. W /\ N e. _om ) -> ( ( M Sat E ) ` N  ) : ( Fmla ` N ) --> ~P ( M ^m _om ) ) $=
```

which would give a full picture about the satisfaction predicate.

Does anybody has an idea/hint how to prove it? It seems to be obvious, and either it is very easy to prove it, or it is difficult, at least tedious (distinction of many cases).

Details:
main:
* ~fnssresd moved from GS's mathbox
* new theorems ~difelpw, ~rabelpw added

MC's mathbox:
* ~goeleq12bg added
* more theorems for satisfaction predicate (Sat) added: ~satfsucom, ~satfn, ~satom, ~satfvsucom, ~satfv0, ~satfvsuc, ~satfrel, ~satfdm, ~satfrnmapom, ~satfv0fun
* basic theorem for the simplified version of the satisfaction predicate (SatE) added: ~sate0
* more theorems for Goedel formulas (Fmla) added: ~isfmlasuc, ~satfdmfmla

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -25,6 +25,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+15-Oct-23 fnssresd  [same]      moved from GS's mathbox to main set.mm
 11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
  6-Oct-23 vtocl2d   [same]      moved from TA's mathbox to main set.mm
  4-Oct-23 df-cda    df-dju      use |_| instead of +c


digama0:
It is indeed a biggish case disjunction. I will abbreviate ``( M Sat E )`` as $sat$.

* Lemma 1: $m \le n\to sat(m)\subseteq sat(n)$: very obvious proof by induction on $n$.

* Lemma 2: 

$$
\begin{aligned}
&\langle x,y\rangle\in sat(n+1)\to\\
&\quad (\exists i \in \mathbb{N},\exists j \in \mathbb{N},\ x=\ulcorner i\in j\urcorner\wedge y=\textlbrackdbl\overline{\in}\textrbrackdbl(i,j))\lor{}\\
&\quad (\exists a \in sat(n),\exists b \in sat(n),\ x=\ulcorner a_1\overline{\wedge}b_1\urcorner\wedge y=\textlbrackdbl\overline{\wedge}\textrbrackdbl(a_2,b_2))\lor{}\\
&\quad (\exists i \in \mathbb{N},\exists b \in sat(n),\ x=\ulcorner \forall i,b_2\urcorner\wedge y= \textlbrackdbl\overline{\forall}\textrbrackdbl(i,b_2))
\end{aligned}
$$

(The meanings of $\ulcorner a_1\overline{\wedge}b_1\urcorner$ and $\textlbrackdbl\overline{\wedge}\textrbrackdbl(a_2,b_2)$ should hopefully be clear from context, denoting the respective expressions in the definition of $sat$.) We prove this by induction on $n$. Given $\langle x,y\rangle\in sat(n+1)$, either $\langle x,y\rangle\in sat(n)$ in which case you can apply the inductive hypothesis and lemma 1, or else one of the latter two disjuncts holds.

* Theorem: $sat(n)$ is functional. (Thus $sat(\omega)$ is also functional, being an increasing union of functions.)

By induction on $n$, you already have the base case. Let $a,b\in sat(n+1)$ and $a_1=b_1$. Then by lemma 2 we have 9 cases. Observe that if $a$ is in the first case then $a_{11}=0$, and likewise $a_{11}=1$ and $a_{11}=2$ in the other two cases; by this we can eliminate all the off-diagonal cases. For the three remaining cases:
* $a_{11}=b_{11}=0$: You have already done this for the base case, but in short $a_2=\textlbrackdbl\overline{\in}\textrbrackdbl(a_{12})=\textlbrackdbl\overline{\in}\textrbrackdbl(b_{12})=b_2$.
* $a_{11}=b_{11}=1$: In this case we know there exists $u,v,u',v'\in sat(n)$ such that $a_1=b_1=\langle 1,\langle u_1,v_1\rangle\rangle=\langle 1,\langle u'_1,v'_1\rangle\rangle$ and $a_2=\textlbrackdbl\overline{\wedge}\textrbrackdbl(u_2,v_2)$, $b_2=\textlbrackdbl\overline{\wedge}\textrbrackdbl(u'_2,v'_2)$. By injectivity of pair, $u_1=u'_1$ and $v_1=v'_1$, so by the IH $u_2=u'_2$ and $v_2=v'_2$, hence $a_2=b_2$.
* $a_{11}=b_{11}=2$: Similar to case 2.

tirix (approved):
Nice to see progress towards Gödel's Incompleteness Theorem (hopefully)!

The new theorems `difelpw` and `rabelpw` are really nice and useful.
Theorem `fnssresd` is just a 2-steps proof but OK, it might help shortening other proofs.

tirix:
@avekens I guess you now have a better view of what `Sat` and `Fmla` represent.
What will you need besides this formalization in order to prove Gödel's Incompleteness Theorem?
E.g. what would a Gödel Sentence look like for ZF expressed in Gödel sets?

avekens:
@digama0 thank you for your hints

>     * Lemma 1: m≤n→sat(m)⊆sat(n): very obvious proof by induction on n.

I had this lemma also on my mind, I will prove it next. 

>     * Lemma 2:

I have proved this lemma already, but did not pubish it yet.

>     * Theorem: sat(n) is functional. (Thus sat(ω) is also functional, being an increasing union of functions.)

This is the tedious task. I have started with it already, but, as I said before, I got stuck. I hope your hints will help me to go on...

avekens:
> @avekens I guess you now have a better view of what `Sat` and `Fmla` represent. What will you need besides this formalization in order to prove Gödel's Incompleteness Theorem? E.g. what would a Gödel Sentence look like for ZF expressed in Gödel sets?

Unfortunately, I am far from being able to say how `Sat` and `Fmla` can be used to prove Gödel's Incompleteness Theorem? And I still have no idea how Gödel Sentence can look like.

avekens:
> @digama0 thank you for your hints
> 
> > ```
> > * Lemma 1: m≤n→sat(m)⊆sat(n): very obvious proof by induction on n.
> > ```
> 
> I had this lemma also on my mind, I will prove it next.
> 
> > ```
> > * Lemma 2:
> > ```
> 
> I have proved this lemma already, but did not pubish it yet.

I have just added these two theorems to set.mm.

benjub:
I haven't followed this and the previous threads closely, but why do you define https://us.metamath.org/mpeuni/df-goal.html as
```|- A.g N U = <. 2o , <. N , U >. >.```
and not as
```|- A.g U = ( x e. _om |-> <. 2o , <. x , U >. >. )```
or even the more compact (even though maybe a bit less convenient to use)
```|- A.g U = ( x e. _om |-> <. suc suc x , U >. )```
?

By the way: what you call "Gödel set" is usually called "Gödel code", if I remember correctly. We should follow as much as possible standard terminology.

**Comment by digama0** [comment]
I don't call it a code because codes are generally natural numbers and this is a finitely branching tree. It has the general behavior of a code though.

I don't use `A.g U` because then it would make formulas look horrible, ``( A.g U ` N )`` has 3 extra tokens and the binder variable is completely in the wrong place. It's hard enough to read them given that the variables are all numbers. (An alternative design if we didn't care about formula readability would be to use de Bruijn variables instead; then you wouldn't even need a variable argument and could get away with just `A. g U`.)

**Comment by benjub** [comment]
Since the standard term for Gödel codes which are numbers is "Gödel number" (https://en.wikipedia.org/wiki/G%C3%B6del_numbering), this leaves "Gödel code" free to use for more general cases, like the present one. The phrase "Gödel set" is not very informative (since "everything is a set"), and a bit confusing ("with each formula, we associate the set of what?"). Also, it may hint to the different topic of the constructible universe.

As for defining $\ulcorner \forall v_N \varphi \urcorner$ as ``( A.g U ` N )`` where `U` is the code of $\varphi$, I cannot tell whether the advantages or the drawbacks win.

In your proposal using De Bruijn indices, you need to $\alpha$-rename variables within `U`, so I'm not sure how you would go ?

**Comment by savask** [comment]
I'm a bit out of the loop, but what's the reasoning about using sets as codes for formulae? If I recall correctly, the whole point of coding things in natural numbers is that they can be operated in Peano arithmetic.

**Comment by digama0** [comment]
Well, operating in peano arithmetic isn't really our goal, is it? It's easier to work with plain pairs. Besides, they live in HF (hereditarily finite sets) which is equiconsistent with PA anyway, in case we wanted to prove something in that direction. The point here is not to follow Goedel's footsteps, it is to define syntax and this necessarily entails defining some kind of finitistic structural object, whether or not it is done in the natural numbers.

**Comment by savask** [comment]
> Well, operating in peano arithmetic isn't really our goal, is it? 

Correct me if I'm wrong, but usually Godel's theorem is formulated along the lines of "every consistent recursively enumerable theory containing Peano's arithmetic contains a sentence which cannot be proved or disproved". Usually "recursively enumerable" is defined for natural numbers, also if I recall correctly Godel's sentence is also formulated using arithmetic (there's even a polynomial version due to Matiyasevich).

**Comment by digama0** [comment]
The purpose of this section was not to prove Goedel's theorem, it was to develop proof theory and model theory. I think that formalizing that big sequence of definitions inside an embedded PA is not something that is generally useful, it will only be good for that particular statement of Goedel's theorem. You can prove lots of compactness and completeness and incompleteness results, including analogues of Goedel's incompleteness theorems, without ever using the natural numbers directly for encoding.

File: set.mm [hunk 1]
```diff
@@ -507709,6 +508166,37 @@ required as (atomic) formulas, can be introduced as a defined notion in
       CUWTUVOUWOTUYNUXAWPUXGUWOWQWRXBXDXEWRXBXFXGXFXHXIXJXMXKXNXLXO $.
   $}
 
+  ${
+    $d F f i u v $.  $d N f i u v $.
+    $( The characterization of a Godel formula of heigth at least 1.
```
icecream17, line 508171:
```suggestion
    $( The characterization of a Godel formula of height at least 1.
```

icecream17 (approved):

savask:
> The purpose of this section was not to prove Goedel's theorem

Ah, I was under the impression that AV's goal is to prove (or at least move towards to) Goedel's theorem.

> I think that formalizing that big sequence of definitions inside an embedded PA is not something that is generally useful, it will only be good for that particular statement of Goedel's theorem.

I think that substituting the classical theorem statement with something else and then proving it is a bit cheating, but nevertheless, I think an encoding of formulae into numbers can be used in other decidability results (like "first-order theory of algebraically closed fields is decidable" type of stuff).

digama0:
> I think an encoding of formulae into numbers

No matter what we do, we won't be working in PA, we are working in ZFC and sets exist in the ambient context in which we are doing these proofs. If you want to really be hardcore reverse mathematics, you have to throw away the whole infrastructure and work in something like [peano.mm1](https://github.com/digama0/mm0/blob/master/examples/peano.mm1) instead. Given that we are not doing that, working in "finite ordinals" or "hereditarily finite sets" or "elements of `NN0`" doesn't make any difference, all of these sets are computably bijective with the natural numbers of Peano and Goedel.

savask:
> No matter what we do, we won't be working in PA, we are working in ZFC and sets exist in the ambient context in which we are doing these proofs.

I agree, but my point is that if you want to talk about decidability of theories etc, you will have to convert your "Goedel sets" into "Goedel numbers". Unless I'm missing something, this just adds another layer of indirection with little benefit.

digama0:
> I agree, but my point is that if you want to talk about decidability of theories etc, you will have to convert your "Goedel sets" into "Goedel numbers". 

I don't see why you would... A proper decidability infrastructure should handle arbitrary enumerable sets. You really want to hide the coding tricks as much as possible or it gets very unpleasant to work with because the natural way to express a thing is very rarely as a literal natural number.

savask:
> A proper decidability infrastructure should handle arbitrary enumerable sets.

I'm not aware of such an infrastructure, but if you are, then fine! :-) The way I know involves usual computability theory which works with subsets of natural numbers, functions from naturals to naturals etc., and then if you want to study decidable theories or computable structures, you have to encode things into naturals effectively. There's a certain amount of trickery involved when working with computable [numberings](https://en.wikipedia.org/wiki/Numbering_(computability_theory)) though.

digama0:
> > A proper decidability infrastructure should handle arbitrary enumerable sets.
> 
> I'm not aware of such an infrastructure, but if you are, then fine! :-)

I happen to have written [a paper](https://drops.dagstuhl.de/opus/volltexte/2019/11067/pdf/LIPIcs-ITP-2019-12.pdf) about it :) The relevant section is section 2 on encodable sets.

PR #3573, opened 2023-10-15
Prove that some small graphs are acyclic

BTernaryTau:
acycgr0v

> A null graph (with no vertices) is an acyclic graph.

acycgr1v

> A multigraph with one vertex is an acyclic graph.

acycgr2v

> A simple graph with two vertices is an acyclic graph.

prclisacycgr

> A proper class (representing a null graph, see ~ vtxvalprc ) has the property of an acyclic graph (see also ~ acycgr0v ).

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -25,6 +25,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
  6-Oct-23 vtocl2d   [same]      moved from TA's mathbox to main set.mm
  4-Oct-23 df-cda    df-dju      use |_| instead of +c
  4-Oct-23 cdaval    df-dju      use |_| instead of +c
--- a/discouraged
+++ b/discouraged
@@ -17267,6 +17267,7 @@ New usage of "poml4N" is discouraged (3 uses).
 New usage of "poml5N" is discouraged (1 uses).
 New usage of "poml6N" is discouraged (1 uses).
 New usage of "prcdnq" is discouraged (14 uses).
+New usage of "prclisacycgr" is discouraged (0 uses).
 New usage of "preleqALT" is discouraged (0 uses).
 New usage of "prlem934" is discouraged (3 uses).
 New usage of "prlem936" is discouraged (1 uses).


icecream17 (approved):

tirix (approved):

set.mm
```diff
@@ -501875,6 +501875,69 @@ it has no (non-trivial) cycles.  (Contributed by BTernaryTau,
       NUOUPUQURUS $.
   $}
 
+  ${
+    $d f g p $.  $d f G p $.  $d f V p $.
```
avekens, line 501879: The class variables are usually either at the beginning or at the end. I do not know any case where they are in the middle of the sequence of setvar variables in a DV condition.

set.mm
```diff
@@ -501875,6 +501875,69 @@ it has no (non-trivial) cycles.  (Contributed by BTernaryTau,
...
+
+  ${
+    $d f g p $.  $d f G p $.  $d f V p $.
+    prclisacycgr.1 $e |- V = ( Vtx ` G ) $.
+    $( A proper class (representing a null graph, see ~ vtxvalprc ) has the
+       property of an acyclic graph (see also ~ acycgr0v ).  (Contributed by
+       BTernaryTau, 11-Oct-2023.) $)
+    prclisacycgr $p |- ( -. G e. _V ->
+              -. E. f E. p ( f ( Cycles ` G ) p /\ f =/= (/) ) ) $=
+      ( vg cvv wcel wn c0 wceq cv ccycls cfv wbr wa wex cvtx df-br nexdv syl5eq
```
avekens, line 501932: I do not know if it is a good idea to build additional theorems on the degenerate cases of (null) graphs. Probably there will be no application for it. You can leave it in your mathbox, but it should not go to main sometime.

set.mm
BTernaryTau: I've just been copying the distinct variable conditions from mm-lamp, so all of my theorems use whatever order it uses. I don't see any mention of this in https://us.metamath.org/mpeuni/conventions.html, so either I'm missing something (definitely possible) or this isn't documented. Either way, that's probably why mm-lamp doesn't respect it. If you want, I can create a separate PR to update all my distinct variable conditions, and also file a feature request for mm-lamp to make sure other users don't run into this issue.

set.mm
BTernaryTau: Would it be better to just mark it as new usage is discouraged? I created it to be analogous to [prcliscplgr](https://us.metamath.org/mpeuni/prcliscplgr.html) in main, but not to actually build off of it. I don't know what the policy is on tagging a theorem as discouraged vs just not including the theorem.

set.mm
tirix: For the record, @avekens has [posted on the metamath mailing list](https://groups.google.com/g/metamath/c/vRXuz7Qs5wY) on the topic.
I agree with @digama0's answer that users should not worry about order of the DV conditions, i.e. tools should handle it transparently for us.

set.mm
tirix: If `~vtxvalprc` deserves a place in the sun I don't see why `~prclisacycgr` would not, but in both cases my question would be. what math theorem do we ultimately want to prove from those?

avekens: "new usage is discouraged" is OK (for now).

set.mm
avekens: According to @digama0's "class/wff variables should come at the beginning", this should be done manually, until the tools take care for it. This holds already for the cases in this PR. The setvar variables should be ordered alphabetically.

set.mm
BTernaryTau: > Since a proper class cannot be an element of ` AcyclicGraph `, a sethood antecedent like ` G e. W ` is necessary in theorems such as ~ acycgr0v .

I can add this to the description as well if that more clearly indicates why this theorem is relevant even if nothing depends on it?

set.mm
BTernaryTau: All I see regarding manual ordering is:
> But also, this is not a thing that users should ever have to think about.

I'd prefer to wait to make any changes until a consensus has been reached regarding placing class/wff vars at the beginning vs at the end of DV conditions.

digama0: I agree with @BTernaryTau , until we have formatting support let's just not worry about it, it won't make the formatter PR any different.

avekens: For now, it can remain as it is, because it is in a mathbox. I do not know how long it takes until a formatter is availabe, and at least the main body of set.mm should not contain additional variants of ordering the variables in a DV statement. I plan to add a correspondig convention in section "Conventions" soon.

digama0: To put it another way: until we have a formatter, I think the convention should be that the $d declarations are in arbitrary order, and if people want guidance on what order to write things in we have a suggestion but there is no requirement that they abide by it, in main or otherwise.

set.mm
avekens: > > Since a proper class cannot be an element of `AcyclicGraph`, a sethood antecedent like `G e. W` is necessary in theorems such as ~ acycgr0v .

Where does this quote come from? I miss any context for it.

> I can add this to the description as well if that more clearly indicates why this theorem is relevant even if nothing depends on it?

This is obvious and therefore no need to add this to the description. There are a lot of such cases in set.mm.

BTernaryTau: > Where does this quote come from? I miss any context for it.

This wasn't really a quote, just a modified snippet from the description of prcliscplgr.

> This is obvious and therefore no need to add this to the description. There are a lot of such cases in set.mm.

Sounds good to me.

avekens: I see, so it was a quote from myself ;-).

After "(New usage is discouraged.)" is added, as you proposed, I will accept this PR, so it can be merged immediately afterwards.

BTernaryTau: Done.

icecream17 (approved):

avekens (approved):

wlammen:
I don't see a good reason to keep this mathbox work longer on hold.

PR #3574, opened 2023-10-15
Fixed sbralie

GinoGiotto:
Requested by @benjub in https://github.com/metamath/set.mm/pull/3571#discussion_r1358691873

### DIFF
--- a/nf.mm
+++ b/nf.mm
@@ -24273,14 +24273,15 @@ $)
   $}
 
   ${
-    $d x y z $.  $d x z ph $.  $d y z ps $.
+    $d x y z $.  $d x z ps $.  $d y z ph $.
     sbralie.1 $e |- ( x = y -> ( ph <-> ps ) ) $.
     $( Implicit to explicit substitution that swaps variables in a quantified
        expression.  (Contributed by NM, 5-Sep-2004.) $)
-    sbralie $p |- ( [ y / x ] A. y e. x ph <-> A. x e. y ps ) $=
-      ( vz cv wral wsb cbvralsv sbbii nfv raleq sbie bitri sbco2 eqcoms ralbii
-      wb ) ADCGZHZCDIZADFIZFDGZHZBCUDHZUBUCFTHZCDIUEUAUGCDADFTJKUGUECDUECLUCFTU
-      DMNOUEUCFCIZCUDHUFUCFCUDJUHBCUDUHADCIBADCFAFLPABDCBDLABSTUDEQNOROO $.
+    sbralie $p |- ( A. x e. y ph <-> [ y / x ] A. y e. x ps ) $=
+      ( vz cv wral wsb cbvralsv sbbii nfv raleq sbie sbco2 wb weq equcoms bitri
+      bicomd ralbii 3bitrri ) BDCGZHZCDIBDFIZFUCHZCDIUEFDGZHZACUGHZUDUFCDBDFUCJ
+      KUFUHCDUHCLUEFUCUGMNUHUEFCIZCUGHUIUEFCUGJUJACUGUJBDCIABDCFBFLOBADCADLBAPC
+      DCDQABETRNSUASUB $.
   $}
 
   ${


benjub (commented):
Does this MR restore the original proofs ? I don't understand: in https://github.com/metamath/set.mm/pull/3571#discussion_r1356712207 @avekens noticed that the proof became longer, and now it seems that it becomes even longer instead of going back to the original ?

GinoGiotto:
> Does this MR restore the original proofs ? I don't understand: in [#3571 (comment)](https://github.com/metamath/set.mm/pull/3571#discussion_r1356712207) @avekens noticed that the proof became longer, and now it seems that it becomes even longer instead of going back to the original ?

Ah, maybe I misinterpreted your request. What I did here is coupling DV(x,ps) and DV(y,ph) as you mentioned, but I still kept the new `[ y / x ]` notation. Restoring the original statement and proof would mean to go back to the `[ x / y ]` notation, which is against the request of issue #777 (sbralie is in the list of #777).

Also avekens accepted the new proof after I explained the reason of why it is longer: https://github.com/metamath/set.mm/pull/3571#discussion_r1357218015 https://github.com/metamath/set.mm/pull/3571#discussion_r1357761876

EDIT: if you mean to restore the original proof only, I'm not sure that is possible because of the fact that I kept `x = y` in the hypothesis, so as you mentioned in #3571 this is a little more than a simple variable renaming.

EDIT2: same thing for ph and ps. I kept `ph <-> ps` in the hypothesis instead of switching into `ps <-> ph` and this caused an additional proof longening.

set.mm
```diff
@@ -30818,14 +30818,15 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
...
        expression.  (Contributed by NM, 5-Sep-2004.) $)
-    sbralie $p |- ( [ y / x ] A. y e. x ph <-> A. x e. y ps ) $=
-      ( vz cv wral wsb cbvralsv sbbii raleq sbievw bitri sbco2vv eqcoms ralbii
-      wb ) ADCGZHZCDIZADFIZFDGZHZBCUCHZUAUBFSHZCDIUDTUFCDADFSJKUFUDCDUBFSUCLMNU
-      DUBFCIZCUCHUEUBFCUCJUGBCUCUGADCIBADCFOABDCABRSUCEPMNQNN $.
+    sbralie $p |- ( A. x e. y ph <-> [ y / x ] A. y e. x ps ) $=
+      ( vz cv wral wsb cbvralsv sbbii raleq sbievw sbco2vv wb weq equcoms bitri
+      bicomd ralbii 3bitrri ) BDCGZHZCDIBDFIZFUBHZCDIUDFDGZHZACUFHZUCUECDBDFUBJ
+      KUEUGCDUDFUBUFLMUGUDFCIZCUFHUHUDFCUFJUIACUFUIBDCIABDCFNBADCBAOCDCDPABESQM
+      RTRUA $.
```
avekens, line 30829: Now I understand: the previous version was actually a different theorem, because of the reversed DV conditions. The current version seems to the same as the old one, only with all variables exchanged ( x <-> y ; ps <-> ph ). I also wonder, as @benjub did, why the proof became again longer.

I tried it by myself, and I got a proof which ist still longer than the previous one, but a little bit shorter than the current one (it does not use ~3bitrri):
```
      ( vz cv wral wsb cbvralsv sbbii raleq sbievw bitri sbco2vv wb weq equcoms
      bicomd ralbii ) BDCGZHZCDIZBDFIZFDGZHZACUEHZUCUDFUAHZCDIUFUBUHCDBDFUAJKUH
      UFCDUDFUAUELMNUFUDFCIZCUEHUGUDFCUEJUIACUEUIBDCIABDCFOBADCBAPCDCDQABESRMNT
      NN $.
```
The reason is that both sides of the hypothesis have to be swapped (by ~equcoms and ~ bicomd). I think this is acceptable, especially because the proof of ~tfinds2 became shorter.

GinoGiotto: > The reason is that both sides of the hypothesis have to be swapped (by ~equcoms and ~ bicomd).

Exactly, the original proof length is certainly restorable  if the hypothesis is written like `|- ( y = x -> ( ps <-> ph ) )`, but since that's not cosmetic I opted to keep the original one instead. This is why the new version has no guarantee to have a proof of the same length of the old one.

> I tried it by myself, and I got a proof which ist still longer than the previous one, but a little bit shorter than the current one (it does not use ~3bitrri):

I believe you proved `|- ( [ y / x ] A. y e. x ps <-> A. x e. y ph )`, but notice that the statement in this PR is commuted `|- ( A. x e. y ph <-> [ y / x ] A. y e. x ps )`, because usually the statement with ph is on the left, while the one with ps is on the right. I can bring this back tho if you prefer.

avekens: > I believe you proved `|- ( [ y / x ] A. y e. x ps <-> A. x e. y ph )`, but notice that the statement in this PR is commuted `|- ( A. x e. y ph <-> [ y / x ] A. y e. x ps )`, because usually the statement with ph is on the left, while the one with ps is on the right. I can bring this back tho if you prefer.

That's true -> so your version is OK.

avekens (approved):

set.mm
benjub: Then I guess the original version was better: in that case, it seems that having a shorter proof is an indication that the statement was more natural. I don't see anything wrong with having `[ x / y ]` when it's handier, and rigidly sticking to conventions about variable naming is probably not ideal. But as you prefer.

GinoGiotto: > I don't see anything wrong with having `[ x / y ]` when it's handier, and rigidly sticking to conventions about variable naming is probably not ideal. But as you prefer.

My only purpose here is to close #777, and I aim to do so in a manner that satisfies everyone. So I'm not too rigid if we agree that #777 is concluded. 

Btw I've not received an answer to the question in https://github.com/metamath/set.mm/pull/3571#issue-1938929703. What are the criteria for closing #777? I've already converted every occurrence of `|- ( [ x / y ]` in #3571, so every theorem in the list of #777 is now in the form of `|- ( [ y / x ]`. But did Norman Megill intended more generally to convert every occurrence of `[ x / y ]`? Because that is a bigger task.

benjub: I think we can revert sbralie to the original and "comment-and-close" #777 with the comment "Fixed unless there is good reason to keep the `[ x / y ]` idiom".
Sorry that I arrived a bit late in the discussion.

avekens: I have no strong opinion on that, but I would suggest to keep the current version, and to add the old version as ~sbralieALT as compromise. This theorem is used in ~tfinds only, so it has no significant meaning.

I would llike, however, to hear a third opinion: @david-a-wheeler what do you think?

GinoGiotto: @benjub @avekens  I have an alternative proposal, why not keep the current version as is and adding:
<pre>
  ${
    $d x y z $.  $d x z ps $.  $d y z ph $.
    sbralier.1 $e |- ( y = x -> ( ps <-> ph ) ) $.
    $( Version of ~ sbralie with commuted varibables in the hypothesis.
      (Contributed by NM, 5-Sep-2004.) (Revised by Gino Giotto, 18-10-2023.) $)
    sbralier $p |- ( A. x e. y ph <-> [ y / x ] A. y e. x ps ) $=
      ( vz cv wral wsb cbvralsv sbbii raleq sbievw sbco2vv bitri ralbii 3bitrri
      ) BDCGZHZCDIBDFIZFRHZCDITFDGZHZACUBHZSUACDBDFRJKUAUCCDTFRUBLMUCTFCIZCUBHU
      DTFCUBJUEACUBUEBDCIABDCFNBADCEMOPOQ $.
  $}
</pre>

Whose proof has the uppercase section shorter than the original by 5 bytes? This way with don't have to keep exceptions for #777 . The "r" suffix stands for reversed, which is used according to conventions. Notice that `sbralier` is equivalent to the original with the difference that I swapped the biconditional in the main statement. This not only looks more natural, but also shorten the original proof. Adding an ALT version is not ideal because it requires "proof modification is discouraged" and "new usage is discouraged" which is not necessary in this case since we can let the minimizer choose whose version is better. Second we agreed in #3565 that the ALT suffix should be reserved to alternative proofs only, not for alternative versions.

Also @benjub notice another thing, the swapping with ~equcoms and ~bicomd in the current proof of `sbralie` happened in same way in the proof of `tfinds2` with the old `[ x / y ]` version. That's why the new proof of `tfinds2` is shorter. The old version wasn't actually shortening anything, nor was more natural, it was just delegating the swapping job to `tfinds2` instead.

I would even argue that the original version was actually unnatural, not only for the variable naming, but also because it had a longer upper case proof than `sbralier`.

benjub: I haven't followed closely, so choose the statement you think is most natural (probably one statement is enough).

GinoGiotto: > I haven't followed closely, so choose the statement you think is most natural (probably one statement is enough).

Then I'm inclined to keep this PR unchanged, but I want to hear a third opinion for a (hopefully) final agreement.

icecream17 (approved):

tirix (approved):
Good for me as-is.

PR #3575, opened 2023-10-16
All 3-cycles are conjugate in the alternating group of degree 5 or more

tirix:
A few shortenings of previous theorems about cyclic permutations.
New result is that all 3-cycles are conjugate in the alternating group of degree 5 or more, moving towards simplicity of $A_n$.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -288,6 +288,7 @@
 "4syl" is used by "cvmlift3lem9".
 "4syl" is used by "cvmliftlem11".
 "4syl" is used by "cvmseu".
+"4syl" is used by "cycpmconjslem2".
 "4syl" is used by "cygznlem3".
 "4syl" is used by "dchr1".
 "4syl" is used by "dchr1re".
@@ -424,6 +425,8 @@
 "4syl" is used by "swrdccat2".
 "4syl" is used by "symgtrinv".
 "4syl" is used by "tmsxms".
+"4syl" is used by "tocycfvres1".
+"4syl" is used by "tocycfvres2".
 "4syl" is used by "tposss".
 "4syl" is used by "trer".
 "4syl" is used by "tskwe".
@@ -13397,7 +13400,7 @@ New usage of "4atex2-0aOLDN" is discouraged (1 uses).
 New usage of "4atex2-0bOLDN" is discouraged (0 uses).
 New usage of "4atex2-0cOLDN" is discouraged (0 uses).
 New usage of "4ipval2" is discouraged (2 uses).
-New usage of "4syl" is discouraged (189 uses).
+New usage of "4syl" is discouraged (192 uses).
 New usage of "5oai" is discouraged (0 uses).
 New usage of "5oalem1" is discouraged (1 uses).
 New usage of "5oalem2" is discouraged (2 uses).


tirix:
This adds 3 usages of `4syl` which is tagged "New usage is discouraged", but the minimizer still run in a reasonable time.

set.mm
```diff
@@ -467607,6 +467823,274 @@ real number multiplication operation (this has to be defined in the main
...
+        XKVUIVUKVULYSVUIVUKVULUXFWJVWSVULVUKYLVPUXGXSUXHXSUYPVWQVXAWFLVWPUYAVUL
+        UCYTUEYTUXIUYEVWPUHZUYFVWQUYOVXAUYDDUYEVWPUXJWXJUYIVWTUYNWXJUYHVWSUYEVW
+        PWXJUYGVWRGUYEVWPUXKYOWXJWKYMUXLUXMUXNWTUXOAUCVUBVUCGIUCIUXPAVVBVUBBIYC
+        IVUBYBUAUDBIDHVOPNQUXQVUBBIYDWJAGCIVUCVGUBMUXRUXSUXT $.
+
+      $d .+ p q t $.  $d .- p q t $.  $d B p q t $.  $d D p q t $.  $d D u w $.
+      $d M p q t $.  $d M u $.  $d N p q t $.  $d P p q t $.  $d Q p q t $.
+      $d Q u $.  $d T p q t $.  $d p ph q t $.
+      cycpmconjs.t $e |- ( ph -> T e. C ) $.
+      $( All cycles of the same length are conjugate in the permutation group.
```
savask, line 467974: Better to write "in the symmetric group", since clearly cycles of the same length need not be conjugate in an arbitrary permutation group.

tirix: Yes! I've fixed that, and also corrected the "order of the group" (number of permutations in the group) to "degree of the group", (the number of elements in the underlying set).

avekens: I cannot find the definition of "degree of the group" as "the number of elements in the underlying set" in set.mm (and in [Lang} neither). In Wikipedia, however, I found: "The alternating group on a set of n elements is called the alternating group of degree n, ..." (https://en.wikipedia.org/wiki/Alternating_group). But the term "degree" is defined for "symmetric groups" only, not for arbitrary groups.

If this term is used in set.mm, it should be defined somewhere (at least in the commt for ~cyc3conja where it is used).

benjub: Or simply:
> All 3-cycles are conjugate in the alternating group A_n for n>= 5.

or
> All 3-cycles are conjugate in the alternating group A_n on n >= 5 elements.

savask: > But the term "degree" is defined for "symmetric groups" only, not for arbitrary groups.

It is defined for permutation groups, see for instance https://en.wikipedia.org/wiki/Permutation_group#Basic_properties_and_terminology.

tirix: I originally wrote _order_, which was wrong, and which I corrected into _degree_ which is indeed the correct term.
However it's true that it's not defined in `set.mm`, so I'll go with @benjub's proposal.

benjub (approved):

avekens (approved):

savask (approved):

PR #3576, opened 2023-10-17
relcnvtrg

mazsa:
general form of ~ relcnvtr


icecream17 (approved):

set.mm
```diff
@@ -55413,15 +55418,22 @@ singleton of the first member (with no sethood assumptions on ` B ` ).
       ZUTLZGMVJVGVRVTGVQVSUTFUNUQBCVPGTPUGQGUNUSAULVPVORVFVTGVDUTFUHQSSUI $.
   $}
 
+  $( A relation is transitive iff its converse is transitive, general form of
+     ~ relcnvtr .  (Contributed by Peter Mazsa, 17-Oct-2023.) $)
```
icecream17, line 55422:
```suggestion
  $( General form of ~ relcnvtr .  (Contributed by Peter Mazsa,
     17-Oct-2023.) $)
```

actually, does "transitive" apply here? (much more merely a suggestion due to my non-knowledgeable-ness)
https://en.wikipedia.org/wiki/Transitive_relation

is transitive:
xRy and yRz then xRz

relcnvtr
xRy and yRz then xRz <-> z \`R y and y \`R x then z \`R x

relcnvtrg
xRy and ySz then xTz <-> z \`R y and y \`S x then z \`T x

mazsa: Thank you @icecream17 , accepted. I would prefer the solution used in the comment of ~ ineqcom , but unfortunately it is not applicable here.

avekens (approved):

PR #3577, opened 2023-10-18
prove mo4 without ax-10 .. ax-12

wlammen:
1. mo4 can be proven from ax-mp .. ax-7 only.  The lengthy proof has certainly headroom for shortenings (I already spotted one).  I currently have no time to look through it again in detail.  Those watching #3485 may in particular be interested in the result.
2.  Remove an OLD theorem from AV's mathbox, that blocks other OLD theorems.  There are more like dfich2ai, but I don't tamper with mathboxes, unless there is an urgent need.  The same for signstfveq0OLD, signsvtn0OLD, iwrdsplitOLD in TA's mathbox.
4. Remove all OLD theorems in main as of 18-Oct-2023.
5. Remove "Subspaces" from the Hilbert space section.  It's last theorem is gone now.
6. Some comments referenced OLD theorems that are deleted now.  Try to find a saner replacement.  Affected are swrdccatin12lem1, swrdccatin12lem2a, and swrdccatin12lem2c.

This PR will only be touched again in case of revision issues.


set.mm
```diff
@@ -22705,11 +22655,31 @@ derived from that of uniqueness ( ~ df-mo ).  (Contributed by Wolf
...
+      QURCDFUSVBUTQPRVAQVCVDRVEWJCVFVGBDFMZVHBDVISWBWJVMWHWAWFFWGVJWCWMVKSVL $.
+  $}
+
+  ${
+    $d x y $.  $d y ph $.  $d x ps $.
+    mo4OLD.1 $e |- ( x = y -> ( ph <-> ps ) ) $.
+    $( Obsolete version of ~ mo4 as of 18-Oct-2023.  (Contributed by NM,
+       26-Jul-1995.)  (New usage is discouraged.)
+       (Proof modification is discouraged.) $)
+    mo4OLD $p |- ( E* x ph <-> A. x A. y ( ( ph /\ ps ) -> x = y ) ) $=
```
avekens, line 22682: Since the proof is extremely shorter, I would like to keep this theorem as ~mo4ALT theorem.

set.mm
```diff
@@ -151491,7 +150711,7 @@ computer programs (as last() or lastChar()), the terminology used for
       XNXSUUNUVBUVEUVKUVIHUVCEACDUVDXNXOXPXQXBXBXR $.
   $}
 
-  $( Lemma 1 for ~ swrdccatin12OLD .  (Contributed by Alexander van der Vekens,
+  $( Lemma 1 for ~ pfxccatin12 .  (Contributed by Alexander van der Vekens,
```
avekens, line 150714: Then the label of the theorem should also be changed to ~pfxccatin12lem1, shouldn't it?

avekens: This also holds for the following changes of this kind.

set.mm
benjub: I'm not sure it's necessary to have an ALT proof here since that proof adds no insight (that is, proving ~xxx from ~xxxf using ~nfv is obvious and common: it should be kept only for the very first cases).

@wlammen : I would move your new ~mo4 before ~mo4f.

wlammen: I hinted in the comment to mo4f, and moved mo4 before mo4f

avekens (approved):

benjub (approved):

PR #3567, opened 2023-10-10
Reduced axiom usage with BJ and SN mathboxes

GinoGiotto:
This is the announced PR from #3551. I aimed to get the most out of what was available, and this added some complexity. Let's break it down step by step:

* Moved `bj-chvarvv` into main, removed discouraged tag and renamed it into `chvarvv` -> needed for `bj-axrep1` and used to shorten `axext3`.

* `bj-axext3` has a shorter proof than `axext3`, but axiom usage is the same. I found out that `bj-axext3` was submitted before the revision from WL https://github.com/metamath/set.mm/pull/1307#issue-535120208, therefore I added a "revised by BJ" tag instead of "proof shortened by BJ", but I'm not so sure how to handle the credits in this case. What do you think? @wlammen @benjub

* `axext3ALT` uses more axioms than `bj-axext3`, but the length of their upper case section is the same -> deleted `axext3ALT`.

* Replaced `axrep1` proof with the shorter one of `bj-axrep1`. Even if ax-13 has been already dropped by me in #3551, I kept the "revised by BJ" tag instead of the "shortened by BJ" tag because BJ is the first to have found a proof of this theorem without ax-13.

* Replaced `axrep2` proof with the one of `bj-axrep2` to drop ax-13 and minimized it (the proof returned by the minimizer is similar to the old one, using `chvarfv` instead of `chvar`).

* Replaced `axrep3` proof with the one of `bj-axrep3` to drop ax-13 and minimized it (here as well the resulting proof uses `chvarfv` instead of `chvar`).

* Proof of `axrep4` is identical to `bj-axrep4` so doesn't need replacement. Usage of ax-13 is dropped automatically by the previous replacements. Deleted `bj-axrep4` since it's redundant.

* Proof of `axrep5` is identical to `bj-axrep5` so doesn't need replacement . Usage of ax-13 is dropped automatically by the previous replacements. Deleted `bj-axrep5` since it's redundant.

* Moved `sn-axrep` into main. This is needed for `sn-axsep`. In [#3551 (comment)](https://github.com/metamath/set.mm/pull/3551#discussion_r1344632277) @icecream17  proposed to rename it "axrep1", but that's already taken. I renamed it `axrep6` instead.

* Replaced `axsep` proof with the one of `sn-axsep` to drop axioms from ax-9 to ax-13.

* `bj-axsep` has a longer proof and uses more axioms than `sn-axsep` -> deleted `bj-axsep`.

[...]

PR #3578, opened 2023-10-18
Remove an unnecessary DV condition from axsep

benjub:
Remove an unnecessary DV condition from axsep and moot DV conditions from ax-sep and axsep2.
(minimal simplification, as promised in https://github.com/metamath/set.mm/pull/3567#pullrequestreview-1677931284)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -25,6 +25,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+22-Oct-23 axsep2    axsepg
 11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
  6-Oct-23 vtocl2d   [same]      moved from TA's mathbox to main set.mm
  4-Oct-23 df-cda    df-dju      use |_| instead of +c
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -15,16 +15,93 @@ The ones WITH "Notes" may have to be processed manually.
 PROPOSED:
 Date      Old       New         Notes
 proposed  syl       imtri       (analogous to *bitr*, sstri, etc.)
-proposed  sylib     imbitri     etc.
-proposed  sylbid    biimtrd     etc.
-proposed  sylbird   biimtrrd    etc.
-proposed  syl5*     *trid       (syl5bi -> biimtrid; syl5eqel -> eqeltrid;etc.)
-proposed  syl6*     *trdi
+                                there is less agreemnt on renaming syl
+                                than others here
+proposed  syld      imtrd
+proposed  syldc     imtrdcom
+proposed  syldd     imtrdd
+proposed  sylbi     biimtri
+proposed  sylib     imbitri
+proposed  sylbb     bitriim
+proposed  sylibr    imbitrri
+proposed  sylbir    biimtrri
+proposed  sylbbr    bitrriim
+proposed  sylbb1    bitr3iim
+proposed  sylbb2    bitr4iim
+proposed  sylibd    imbitrd
+proposed  sylbid    biimtrd
+proposed  sylibrd   imbitrrd
+proposed  sylbird   biimtrrd
+proposed  syl5com   imtridcom
+proposed  syl5      imtrid      alternate proposal: sylid
+proposed  syl56     imtridi
+proposed  syl5d     imtridd
+proposed  syl5bi    biimtrid
+proposed  syl5bir   biimtrrid
+proposed  syl5ib    imbitrid
+proposed  syl5ibcom imbitridcom
+proposed  syl5ibr   imbitrrid
+proposed  syl5ibrcom imbitrridcom
+proposed  syl5bb    bitrid      compare to bitri or bitrd
+proposed  syl5rbb   bitridcom
+proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
+proposed  syl5rbbr  bitr3idcom
+proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
+proposed  syl5req   eqtridcom
+proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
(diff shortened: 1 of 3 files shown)


benjub:
Ah, the DV conditions of `axeee $p` and `ax-eee $a` are probably required to be identical:
```
 ?Warning: The mandatory distinct variable pairs for statements "ax-sep" and
  "axsep" are different or have a different order.
```
and
```
Run metamath-knife/target/release/metamath-knife --verify \
warning: Invalid axiom restatement
     --> ./set.mm:46740:5
      |
46740 |     axsep $p |- E. y A. x ( x e. y <-> ( x e. z /\ ph ) ) $=
      |     ----- This ax* theorem does not match the corresponding ax-*
      |
     ::: ./set.mm:46754:5
      |
46754 |     ax-sep $a |- E. y A. x ( x e. y <-> ( x e. z /\ ph ) ) $.
      |     ------ note: Axiom ax-* here
      |
1 diagnostics issued.
Error: Process completed with exit code 1.
```
Maybe the best solution would be that the metamath-exe and metamath-knife requirements be to have an inclusion of DVs, and not an equality ?

sctfn:
Personally, I’d rather have it be exact, even if it’s not strictly necessary from a proof point of view. On Oct 18, 2023, at 5:59 PM, Benoit ***@***.***> wrote:﻿
Ah, the DV conditions of axeee $p and ax-eee $a are probably required to be identical:
 ?Warning: The mandatory distinct variable pairs for statements "ax-sep" and
  "axsep" are different or have a different order.

and
Run metamath-knife/target/release/metamath-knife --verify \
warning: Invalid axiom restatement
     --> ./set.mm:46740:5
      |
46740 |     axsep $p |- E. y A. x ( x e. y <-> ( x e. z /\ ph ) ) $=
      |     ----- This ax* theorem does not match the corresponding ax-*
      |
     ::: ./set.mm:46754:5
      |
46754 |     ax-sep $a |- E. y A. x ( x e. y <-> ( x e. z /\ ph ) ) $.
      |     ------ note: Axiom ax-* here
      |
1 diagnostics issued.
Error: Process completed with exit code 1.

Maybe the best solution would be that the metamath-exe and metamath-knife requirements be to have an inclusion of DVs, and not an equality ?

GinoGiotto:
An alternative idea could be to rename [axsep](https://us.metamath.org/mpeuni/axsep.html) into something else (like `axsepg` since it's more general?). Notice that not all axioms have their corresponding identical theorem. For example [ax-ext](https://us.metamath.org/mpeuni/ax-ext.html) doesn't have one, and there is also a similar case here where [axext3](https://us.metamath.org/mpeuni/axext3.html) is a generalization of [ax-ext](https://us.metamath.org/mpeuni/ax-ext.html) because of the elimination of a DV condition.

Then I would also rename [axsep2](https://us.metamath.org/mpeuni/axsep2.html) into `axsepgALT` since with the changes in this PR [axsep](https://us.metamath.org/mpeuni/axsep.html) and [axsep2](https://us.metamath.org/mpeuni/axsep2.html) would become exacly identical, with the proof being the only difference between the two. Currently the comment of [axsep2](https://us.metamath.org/mpeuni/axsep2.html) states `A less restrictive version of the Separation Scheme axsep`, but with this PR it wouldn't be less resctrictive anymore, so the comment [axsep2](https://us.metamath.org/mpeuni/axsep2.html) should be updated as well.

sctfn:
Given that axsep is provable, though, we should really include the exact statement. This is a case where we’re trying to match the literature to verify the literature, so we can and should be pedantic. On Oct 18, 2023, at 7:41 PM, GinoGiotto ***@***.***> wrote:﻿
An alternative idea could be to rename axsep into something else (like axsepg since it's more general?). Notice that not all axioms have their correspoding identical theorem. For example ax-ext doesn't have one, and there is also a similar case here where axext3 is a generalization of ax-ext because of the elimination of a DV condition.
Then I would also rename axsep2 into axsepgALT since with the changes in this PR axsep and axsep2 would become exacly identical, with the proof being the only difference between the two. Currently the comment of axsep2 states A less restrictive version of the Separation Scheme axsep, but with this PR it wouldn't be less resctrictive anymore, so the comment axsep2 should be updated as well.

digama0:
> Maybe the best solution would be that the metamath-exe and metamath-knife requirements be to have an inclusion of DVs, and not an equality ?

I'd rather we didn't. The purpose of the check is to ensure that it is indeed a restatement of the axiom, and it should be as strict as we can practically make it (including e.g. variable names) so as to make it as easy as possible for readers to agree that it is indeed the same theorem.

benjub:
Ok, then let's keep `axsep` identical to `ax-sep`, including DVs.

I'll update the MR to have the theorem without DV relabeled as, `sepfromrep`, then prove its instance `axsep` in one line from it, and also relabel `axsep2` to `axsepg`.

The reason I'd like to have `sepfromrep` is that we are currently saying "this proof actually proves the more general statement xxx", which is a very common thing to do in the math literature, but here, in Metamath, the main principle is to be explicit in what we do, so instead of saying "we can prove xxx", we just do it.

GinoGiotto:
> I'll update the MR to have the theorem without DV relabeled as, `sepfromrep`, then prove its instance `axsep` in one line from it, and also relabel `axsep2` to `axsepg`.

Maybe I would name it `sepgfromrep`, or maybe we could imitate what has been done for ax-6, whose weakened version was derived from ax-sep in [ax6vsep](https://us.metamath.org/mpeuni/ax6vsep.html) (so following this analogy our case would be `axsepgrep` -> "derive a more general version of ax-sep from ax-rep").

benjub:
I added ~axsepgfromrep (it's ok to have a longer label since it won't be used much). I also adapted the comments of ~axsep, ~ax-sep, ~axsepg with care, in order to make the terminology more uniform and add some precisions, with minimal changes.

avekens (approved):

PR #3579, opened 2023-10-19
split set-main.mm into subfiles

### LINKED ISSUE #3520: split set-main.mm into subfiles
When running the `write source set.mm /split` command, the largest file `set-main.mm` is 20 megabytes long which is about half of set.mm

To make this easier to work with I'm proposing to split `set-main.mm` further into subfiles

https://us.metamath.org/mpeuni/mmtheorems.html

<details><summary>As an initial suggestion I would say</summary>

+ chapters 2 to 4: `set-settheory.mm`
+ chapters 5 and 6: `set-numbers.mm`
+ chapters 7 to 10: `set-structures.mm`
+ remaining (11 to 17): stay in `set-main.mm`

(note that chapters 15 and 16 are already separate files)
(chapter 11 might also go in structures?)

</details>

Edit: will be per-chapter

icecream17:
as suggested in #3520
resolves #3520


set.mm
```diff
@@ -423544,6 +423582,7 @@ number of vertices in a friendship graph is (k(k-1)+1), see
 $( End $[ set-graphth.mm $] $)
```
avekens, line 423583:
```
$( End $[ set-tarskigeom.mm $] $)
$( Beginn $[ set-graphth.mm $] $)
```

are missing!?

icecream17: The start and end tarskigeom and graphth tags already exist

avekens (approved):

set.mm
icecream17: It's because these files were already separate before this pr, so I didn't actually change anything

PR #3580, opened 2023-10-19
Revise vtocl2d

BTernaryTau:
This change revises vtocl2d to be shorter and require fewer distinct variables, without changing which axioms it depends on. It also moves vtocl2d to just after vtocld since the new proof does not depend on class substitution.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18024,6 +18024,7 @@ New usage of "vk15.4jVD" is discouraged (0 uses).
 New usage of "vmcn" is discouraged (1 uses).
 New usage of "vsfval" is discouraged (4 uses).
 New usage of "vtocl2OLD" is discouraged (0 uses).
+New usage of "vtocl2dOLD" is discouraged (0 uses).
 New usage of "vtocl2gOLD" is discouraged (0 uses).
 New usage of "vtoclALT" is discouraged (0 uses).
 New usage of "vtoclgftOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


avekens (approved):

wlammen (approved):

wlammen:
Cascading effect: Did you check whether theorems using vtocl2d now need fewer dv conditions as well?

tirix (approved):

BTernaryTau:
> Cascading effect: Did you check whether theorems using vtocl2d now need fewer dv conditions as well?

Good point. This change actually doesn't affect any of the 3 theorems using vtocl2d, though I did find that the mathbox theorem submateq has a DV condition for a nonexistent variable M.

tirix:
It looks like `~submateq` and `~submatminr1` share a block and the corresponding DV conditions. 
The DV on variable `M` is related to `~submatminr1` only, it could indeed be moved either just before that theorem or into its block.

I think someone has built a tool to detect unused/unnecessary DV condition, but I can't remember where to find it.

icecream17:
https://github.com/metamath/set.mm/blob/develop/scripts/check-dvs.py checks for variables that don't exist in any proof used in the same scope

Of course, not all extra dv conditions are found by this script, (in addition to this circumstance, a variable may exist but not need a dv condition) and I haven't heard of any scripts that do

icecream17 (approved):

tirix:
Thank you @icecream17 ! I've had a try now and it found 57 such cases.

PR #3582, opened 2023-10-20
Prove upgracycumgr

BTernaryTau:
lfuhgr

> A hypergraph is loop-free if and only if every edge connects at least two vertices.

lfuhgr2

> A hypergraph is loop-free if and only if every edge is not a loop.

lfuhgr3

> A hypergraph is loop-free if and only if none of its edges connect to only one vertex.

loop1cycl

> A hypergraph has a cycle of length one if and only if it has a loop.

acycgrislfgr

> An acyclic hypergraph is a loop-free hypergraph.

upgracycumgr

> An acyclic pseudograph is a multigraph.


set.mm
```diff
@@ -498438,6 +498438,60 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
       TZBWPTZRZYCBWSTZRSAYDYFYCAWPWSBYBWFWGAYEYDYCRSYDYCWHHWIWJWK $.
   $}
 
+  ${
+    $d x G $.  $d x V $.
+    lfuhgr.1 $e |- V = ( Vtx ` G ) $.
+    lfuhgr.2 $e |- I = ( iEdg ` G ) $.
+    $( A hypergraph is loop-free if and only if every edge connects at least
```
wlammen, line 498445: On the Wikipedia page https://en.wikipedia.org/wiki/Hypergraph your notion of 'loop-free' is close (but not exactly the same) to a 'simple' hypergraph.  Does it make sense to use consistently the Wikipedia nomenclature? Also a loop can be misunderstood as a closed cycle within the hypergraph.

wlammen (approved):

set.mm
BTernaryTau: I'm just reusing the definition from [section 16.2.3](https://us.metamath.org/mpeuni/mmtheorems266.html#mm26530s) of set.mm.

PR #3584, opened 2023-10-22
shorten spimew

wlammen:
1. The spimew (former spimeh) has a proof of same byte length, but 1 proof line less now.
2. Move spimew before speiv.  The reason is that speiv is a convenience theorem with a weakened premise.  Strengthening it leads to wl-speiv, which is derivable from spimew without a proof size penalty.  exgen (see wl-exgen) is then proven from ax6v -> spimew -> wl-speiv -> exgen, step by step.
The spim* series is elaborating on ax6v, providing basic proof techniques in conjunction with a frequent x = y -> ( ph -> ps ) expression.  More general results like exgen, 19.3v etc. can use these basics.  Using the somewhat artificial speiv breaks the above chain, but I think the underlying idea still persists.
3. spimw and spimew are conjugate theorems, so keep them together, and name them similarly.
4. Delete outdated OLD theorems.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -102,6 +102,7 @@ make a github issue.)
 DONE:
 Date      Old       New         Notes
 22-Oct-23 spimeh    spimew      show relation to spimw
+22-Oct-23 spimev    spimevv     same as with spv/spvv
 15-Oct-23 fnssresd  [same]      moved from GS's mathbox to main set.mm
 11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
  6-Oct-23 vtocl2d   [same]      moved from TA's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -103,6 +103,7 @@ DONE:
 Date      Old       New         Notes
 22-Oct-23 spimeh    spimew      show relation to spimw
 22-Oct-23 spimev    spimevv     same as with spv/spvv
+22-Oct-23 axsep2    axsepg
 15-Oct-23 fnssresd  [same]      moved from GS's mathbox to main set.mm
 11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
  6-Oct-23 vtocl2d   [same]      moved from TA's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -101,6 +101,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+22-Oct-23 spimeh    spimew      show relation to spimw
 15-Oct-23 fnssresd  [same]      moved from GS's mathbox to main set.mm
 11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
  6-Oct-23 vtocl2d   [same]      moved from TA's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -17402,6 +17402,7 @@ New usage of "spcgvOLD" is discouraged (0 uses).
 New usage of "speccl" is discouraged (0 uses).
 New usage of "specval" is discouraged (1 uses).
 New usage of "speimfwALT" is discouraged (0 uses).
+New usage of "spimehOLD" is discouraged (0 uses).
 New usage of "spimtOLD" is discouraged (0 uses).
 New usage of "spimvALT" is discouraged (0 uses).
 New usage of "sps-o" is discouraged (7 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -534636,6 +534629,37 @@ extends this idea to multiple additions (TODO).
...
+  ${
+    $d x y $.
+    wl-speiv.1 $e |- ( x = y -> ( ps -> ph ) ) $.
+    wl-speiv.2 $e |- ps $.
+    $( Theorem ~ speiv is not minimalistic in the sense it uses only necessary
+       hypotheses.  The first can be simplified to an implication, so ~ speiv
+       specializes this one for convenience, similar to ~ spvv vs. ~ spimw .
+       (Contributed by Wolf Lammen, 22-Oct-2023.) $)
+    wl-speiv $p |- E. x ph $=
+      ( wex hbth spimew ax-mp ) BACGFBACDBCFHEIJ $.
```
benjub, line 534647: Indeed. I was wondering the same thing after your previous PR. I wrote speiv with a biconditional in the hypothesis to mimic ~spei, and because it is the general idiom for implicit substitution (even if here, there are no associated DVs, so it's more general).

Maybe you can add this (in addition to speiv, so find another label), and similarly add the strengthened version of spei. And then prove exgen as you do in wl-exgen.

benjub:
General remark regarding these and related theorems between ax-6 and ax-7: equality is not yet fully axiomatized, and many theorems with a hypothesis
```
$e |- ( x = y -> ... ) $.
```
are actually forms of a corresponding theorem that could be stated before ax-6 (and often before ax-5) of the form
```
$e |- ( ph -> ... ) $.  $( where ph is a fresh variable and the other ones have been renamed $)
$e |- E. x ph $.        $( that hypothesis is later fulfilled using ~ax-6 $)
```
The more natural of these theorems could be added, for extra clarity of what rests on what.

wlammen:
The idea behind this all is aleximi, with the first two antecedents swapped, i.e (∃𝑥 𝑥 = 𝑦 → (∀𝑥𝜑 → ∃𝑥𝜓)) from (𝑥 = 𝑦 → (𝜑 → 𝜓)). In the presence of ∃𝑥 𝑥 = 𝑦 you arrive at various simplifications, depending on what dv conditions are assumed. Replacing 𝑥 = 𝑦 with 𝜒 gets you to similar variations of aleximi.  It is perhaps more sensible to have these around than the spim* specializations, that have limited application range, in particular with a ¬ ∀𝑥 ¬ 𝑥 = 𝑦 antecedent.

benjub:
> The idea behind this all is aleximi, with the first two antecedents swapped, i.e (∃𝑥 𝑥 = 𝑦 → (∀𝑥𝜑 → ∃𝑥𝜓)) from (𝑥 = 𝑦 → (𝜑 → 𝜓)). In the presence of ∃𝑥 𝑥 = 𝑦 you arrive at various simplifications, depending on what dv conditions are assumed. Replacing 𝑥 = 𝑦 with 𝜒 gets you to similar variations of aleximi. It is perhaps more sensitive to have these around than the spim* specializations, that have limited application range, in particular with a ¬ ∀𝑥 ¬ 𝑥 = 𝑦 antecedent.

Indeed, I agree. It's more sensible (and less sensitive!). I agree that the antecedent `|- ( -. A. x -. x = y -> ... ` could often be replaced by `|- ( E. x x = y -> ... ` when the statement with the former form is not true intuitionistically (I don't remember how @jkingdon handled these in iset.mm). The idiom `-. A. x -. x = y` was probably kept to be identical to the lemmas of the Kalish--Montague article.

set.mm
benjub: Thanks for speiv/speivv. Actually, maybe a better label for speivv would be speivw since it's weaker than speiv but has no additional DV condition (sorry I didn't come up with this label sooner).
Can you do the same with spei ? (that is, relabel it speiw and add the stronger form as spei) ?

wlammen:
Reopened as #3587

jkingdon:
> I agree that the antecedent `|- ( -. A. x -. x = y -> ... ` could often be replaced by `|- ( E. x x = y -> ... ` when the statement with the former form is not true intuitionistically (I don't remember how @jkingdon handled these in iset.mm).

Don't really remember but as far as I can tell from a relatively cursory look at git history this sort of thing - at least https://us.metamath.org/ileuni/exim.html - was already in place when I started working on iset.mm. For example see https://github.com/metamath/set.mm/pull/174 which removes a non-intuitionistic `exim` proof. Oh, looks like something similar is an axiom in iset.mm - https://us.metamath.org/ileuni/ax-ie2.html . Which in turn is related to `∃-GEN` at https://en.wikipedia.org/wiki/Intuitionistic_logic (and I presume other axiomizations of intuitionistic logic).

Might be a bit hard to apply this directly to set.mm, where `E.` is defined in terms of `A.` and negation, but I'll leave that to people who are looking at this more closely than I am.

PR #3586, opened 2023-10-22
(i)set.mm: minor edits

benjub:


iset.mm
```diff
@@ -5,11 +5,11 @@
...
    https://us.metamath.org and
    https://github.com/metamath/set.mm, and feel free to ask questions at
-   http://groups.google.com/group/metamath. $)
+   https://groups.google.com/group/metamath. $)
 
 $( New users may want to read https://us.metamath.org/ileuni/conventions.html
-   to understand the label naming conventions used in iset.mm.  See also "help
-   verify markup" in the Metamath program for markup conventions. $)
+   to understand the label naming conventions used in iset.mm.  See also the
+   Metamath program command "MM> HELP VERIFY MARKUP" for markup conventions. $)
```
jkingdon, line 12: Hmm, I'm not really a fan of this style - it looks like the user is supposed to type the `MM>` and/or just seems kind of cluttered to my eyes.

But I guess that's what we do in a lot of set.mm comments, so :shrug:

jkingdon (approved):

iset.mm
benjub: It's true there is that drawback. On the other hand, it makes it explicit that one has to type this at the Metamath prompt (and since there are other prompts, `MM-PA> ` and `TOOLS> `, it distinguishes between them, e.g. when one mentions the `MM-PA> MINIMIZE__WITH *` command).

PR #3587, opened 2023-10-22
reopen PR #3584 after a merge conflict

wlammen:
1. The spimew (former spimeh) has a proof of same byte length, but 1 proof line less now.
2. Move spimew before speivw. The reason is that speivw is a convenience theorem with a weakened premise. Strengthening it leads to speiv, which is derivable from spimew without a proof size penalty. exgen is then proven from ax6v -> spimew -> speiv -> exgen, step by step.
The spim* series is elaborating on ax6v, providing basic proof techniques in conjunction with a frequent x = y -> ( ph -> ps ) expression. More general results like exgen, 19.3v etc. can use these basics. Using the somewhat artificial speivw breaks the above chain, but I think the underlying idea still persists.
3. spimw and spimew are conjugate theorems, so keep them together, and name them similarly.
4. Rename speiv -> speivw,  the w at the end stands for weak.
5. Create speiv, a strengthened version of speivw.  exgen is now based on it.
6. Delete outdated OLD theorems.

See also discussion in #3584 .

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -101,6 +101,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+22-Oct-23 spimeh    spimew      show relation to spimw
+22-Oct-23 spimev    spimevv     same as with spv/spvv
 22-Oct-23 axsep2    axsepg
 15-Oct-23 fnssresd  [same]      moved from GS's mathbox to main set.mm
 11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -102,7 +102,7 @@ make a github issue.)
 DONE:
 Date      Old       New         Notes
 22-Oct-23 spimeh    spimew      existential version of spimw
-22-Oct-23 speiv     speivv      same as with spv/spvv
+22-Oct-23 speiv     speivw      weak form of speiv
 22-Oct-23 axsep2    axsepg
 15-Oct-23 fnssresd  [same]      moved from GS's mathbox to main set.mm
 11-Oct-23 2exanali  [same]      moved from ATS's mathbox to main set.mm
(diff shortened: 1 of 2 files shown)


avekens:
I see a lot of changes for the axsep theorems modified by @benjub recently - was the merging performed correctly?

wlammen:
No, I have to do it again.

changes-set.txt
```diff
@@ -101,6 +101,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+22-Oct-23 spimeh    spimew      show relation to spimw
```
benjub, line 104:
```suggestion
22-Oct-23 spimeh    spimew      existential version of spimw
```

changes-set.txt
```diff
@@ -101,6 +101,8 @@ make a github issue.)
...
+22-Oct-23 spimev    spimevv     same as with spv/spvv
```
benjub, line 105: I see no "spimevv" in the diffs ?
Did you mean to write the change for "speiv" ? In that case, I think the new label should be "speivw" for "weak form of speiv", since there is no additional DV condition.

wlammen: That's right

benjub: ~spvv has that label because it adds a DV condition to ~spv, which is not the case of ~speivv with respect to ~speiv. This is why I think ~speivw is a better label.

avekens (approved):

wlammen:
For me it is an open question whether all the spim* theorems are necessary and what role they play in set.mm.  A comment in #3984 states they somehow pinpoint where exactly intuitionistic logic forks from classical logic.  I am not familiar with this, so I tend to keep these theorems, although they are not very insightful in classical logic alone.  I think in the end the comments should somehow denote the purpose.
Anyway, further development can advance from here, the changes are quite standard.  So I think the PR can be merged as is.

PR #3588, opened 2023-10-22
Add txmetcnp to iset.mm

jkingdon:
Stated as in set.mm.

The set.mm proof relies on a bunch of machinery which iset.mm doesn't have (at least yet), so the proof is basically a new one. In very broad terms it is similar, in the sense that both proofs are based on using the maximum metric (Chebyshev distance).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10570,6 +10570,18 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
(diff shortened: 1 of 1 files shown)


avekens (approved):

tirix (approved):
Nice!

jkingdon:
> Nice!

Thanks! It took a while to figure out how this should work but once I found a path, the pieces fell into place fairly nicely.

PR #3589, opened 2023-10-23
shorten mo4, equsexvw; move bj-alequexv to Main

wlammen:
1. shorten mo4 by a proof line.  Since the proof was created just a few days ago, no tag, no OLD version.
2. shorten equsexvw by a proof line
3. move bj-alequexv to Main, and minimize proofs with it.
4. reorder exsbim after equs4v, as it is provable from equs4v (although in fact not done because of shortenings).
5. delete an outdated OLD theorem

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -101,7 +101,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
-23-Oct-23 bj-alequexv alequexv  moved from BJ's mahbox to main set.mm
+23-Oct-23 bj-alequexv alequexv  moved from BJ's mathbox to main set.mm
 22-Oct-23 spimeh    spimew      existential version of spimw
 22-Oct-23 speiv     speivw      weak form of speiv
 22-Oct-23 axsep2    axsepg
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -101,6 +101,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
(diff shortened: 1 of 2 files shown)


benjub (approved):

wlammen:
@benjub I noticed your approval too late, otherwise I would have opened another PR.

icecream17 (approved):
wl-equs4vv reminds me of exsbim

benjub:
> wl-equs4vv reminds me of exsbim

Indeed, this is the one but last step of exsbim (with same prooof), and DV(y,ph) is not necessary. For that last reason, this could be added to Main right before exsbim. Maybe as `sbime` for "substituted formula implies existence" ?

wlammen:
I added wl-equs4vv (a duplicate of bj-alequexv) as a reminder for me to keep an eye on possible applications.  It was not meant to be added to Main without an application besides exsbim.  It is, by the way, a consequence of equs4v and exsimpr, but that proof turned out to be longer.  It still suggests the following order along from simple to complex: equs4v -> bj-alequexv -> exsbim.  As for the name: see also spsbe (can be shortened by bj-alequexv, so it is maybe justified to move it to Main).

Since one dv condition on wl-equs4vv can in fact be dropped, its name is a misnomer.  Thanks for the hint.

avekens (approved):

benjub:
Thank you @wlammen for noticing ~alequexv in my mathbox (I had forgotten about it), and for the associated work.

PR #3590, opened 2023-10-23
Prove upgracycusgr

BTernaryTau:
2cycld and 2cycl2d

> Construction of a 2-cycle from two given edges in a graph.

umgr2cycllem

> Lemma for ~ umgr2cycl .

umgr2cycl

> A multigraph with two distinct edges that connect the same vertices has a 2-cycle.

dff15

> A one-to-one function in terms of different arguments never having the same function value.

umgracycusgr

> An acyclic multigraph is a simple graph.

upgracycusgr

> An acyclic pseudograph is a simple graph.


set.mm
```diff
@@ -498824,6 +498824,20 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
     UJUMLZULUIUKUPULUIUKIABNOZUJUIUJMZUMULUIUKUQABPQUIULUJURLUKUIUJRSURUQUMUIUJ
     UQUMABUAUBTUCUDTUEUFUGUH $.
 
+  ${
+    $d x y A $.  $d x y F $.
+    $( A one-to-one function in terms of different arguments never having the
+       same function value.  (Contributed by BTernaryTau, 17-Oct-2023.) $)
+    dff15 $p |- ( F : A -1-1-> B <->
```
jkingdon, line 498831: I'll admit to being partial to `dff13` but that's probably just because that's what I see in constructive math textbooks. No harm in having more than one equivalence.

jkingdon (approved):

set.mm
wlammen: It is indeed a simple rewrite of dff13 using sth like https://us.metamath.org/mpeuni/2nexaln.html .

tirix: Do you mean [ralnex2](https://us.metamath.org/mpeuni/ralnex2.html)?

tirix (approved):

set.mm
wlammen: yes, thanks for the hint.  So dff15 is in fact a combination of dff13 and ralnex2, and can be easily replaced.

BTernaryTau: I was able to shorten the proof by 4 steps using ralnex2, but the resulting proof still has more steps than those for dff12 and dff14b, so I think it's worth keeping. I've found that I work much better with multiple shorter proofs than a single longer proof, so I prefer to split sections like dff15 out whenever they are sensible on their own. (This is also why I have a separate umgr2cycllem instead of just stuffing it in umgr2cycl.)

wlammen: The tricky part is to also convert a x =/= y to a -. x = y.  That is why you have to drill down the expression to the innermost parts, which lengthens the proof.
OK, getting rid of dff15 completely is not the right idea here.

wlammen (approved):

set.mm
avekens: > I've found that I work much better with multiple shorter proofs than a single longer proof, so I prefer to split sections like dff15 out whenever they are sensible on their own. (This is also why I have a separate umgr2cycllem instead of just stuffing it in umgr2cycl.)

This is also my preferred way to prove theorems, and using lemmata is a recommended way to shorten proofs. From time to time, such lemmata become theorems which are/can be reused by other proofs.

jkingdon: Although `ralnex2` holds in intuitionistic logic (I'll add it to iset.mm in my next pull request), `iman` does not. So it doesn't look like `dff15` would be provable in iset.mm.

Not that this is much of a consideration for set.mm, just that if someone intuitionizes these proofs there'll need to be some suitable adjustment (`A` being finite seems like it might suffice, but there can be a closer look when/if someone tries to move this to iset.mm).

PR #3592, opened 2023-10-25
intuitionize addcn , subcn , and mulcn

jkingdon:
These are relatively straightforward now that we have #3588 . Most of this follows set.mm fairly closely.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -74,6 +74,7 @@
 "ax-mulass" is used by "mulass".
 "ax-mulcl" is used by "mulcl".
 "ax-mulcom" is used by "mulcom".
+"ax-mulf" is used by "mulcncntop".
 "ax-mulrcl" is used by "remulcl".
 "ax16ALT" is used by "dvelimALT".
 "ax16ALT" is used by "dvelimfv".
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -78885,10 +78929,23 @@ sequence rather than the limit of the offsetted sequence (see
...
+     NM, 2-May-1996.) $)
   0ncn $p |- -. (/) e. CC $=
     ( c0 cc wcel cnr cxp 0nelxp df-c eleq2i mtbir ) ABCADDEZCDDFBJAGHI $.
 
+  ${
+    $d A u v x $.
+    $( A complex number is an inhabited set.  Note: do not use this after the
+       real number axioms are developed, since it is a construction-dependent
+       property.  (Contributed by Jim Kingdon, 23-Oct-2023.) $)
+    cnm $p |- ( A e. CC -> E. x x e. A ) $=
```
tirix, line 78942: So should this get a "new usage discouraged" tag?

tirix, line 78942: This is new compared to `set.mm`, but it looks like utlimately you have not used it?

jkingdon: Yes. Added.

tirix (approved):

iset.mm
tirix: Understood, thanks!

Issue #2818, opened 2022-09-07
Add joint denial (NOR) to set.mm


benjub:
Among the "less common" logical binary connectives, set.mm already has a section for alternative denial (NAND) and a section for exclusive disjunction (XOR). It would be nice to add a section for joint denial (NOR). Like these other two sections, it is not meant to have many applications (stating more advanced theorems with these connectives would generally make them less readable), but simply to state the basic properties of that connective (the sections introducing NAND and XOR have 15 theorems each, organized in neatly parallel fashions). (**edit:** Additionally, the section "Truth tables" should receive a new subsection "Joint denial".)

This would make a nice first work for someone who wants to get acquainted to Metamath and set.mm.

david-a-wheeler:
Makes sense to me. I have an electronics engineering degree, focusing on digital electronics, so for me NOR is key basic stuff :-).

benjub:
> Makes sense to me. I have an electronics engineering degree, focusing on digital electronics, so for me NOR is key basic stuff :-).

I just learnt on wikipedia (https://en.wikipedia.org/wiki/NOR_gate) that apparently, the first Apollo lunar modules were based on NOR gates ! (But since this issue would be a nice way to get *acquainted* to Metamath and set.mm, you are banned from doing it !)

walkingsophie:
I think I'll work on this as my first direct contribution to Metamath.

walkingsophie:
Sorry, had some issues with setting stuff up, then real life stuff happen, then University happened, so I couldn't work on it. I'm happy for someone else to work on it.

lesyna:
Jeff Machado and I had already discussed doing NOR before I noticed this issue. Any objections if we take it up?

benjub:
Please, go on !

benjub:
Closed by #3594.

PR #3594, opened 2023-10-26
Add joint denial (NOR)

remimimimimi:
Implementation of #2818. 

What should I do with symbol representation? I was not able to find free symbol for it.


tirix:
Looks good, very nice first submission!

I think the symbol would be `&barvee;`, a vee with an overbar: ⊽
That's `&#8893;` in HTML encoding.
The LaTeX symbol would be `\barvee`, [but I'm not sure if we have it](https://tex.stackexchange.com/questions/201027/in-math-mode-i-have-access-to-veebar-but-not-barvee-why).

remimimimimi:
Thanks!

The current included latex packages are amssymb, amsmath, mathtools, mathrsfs, phonetic, and graphicx. However, we don't have unicode-latex which contains `\barvee`.

I think that adding a whole new latex package deserves its own discussion, and for now `\bar{\vee}` will be fine.

If we want to have proper displaying (coherent with `\veebar`) we can add just that symbol manually, but that seems quite dirty.

tirix:
Yes this looks good for now!

I'll let @benjub comment about whether we take this submission directly into the "main" part of set.mm or if he prefers a submission into a mathbox first, since he was the one who initially suggested adding NOR.

set.mm
```diff
@@ -431454,6 +431526,12 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
   althtmldef "\/_" as " &#8891; ";
     /* 2-Jan-2016 reverted sans-serif */
   latexdef "\/_" as "\veebar";
+htmldef "-\/" as
+    " <IMG SRC='vee.gif' WIDTH=9 HEIGHT=19 ALT=' -\/' TITLE='-\/' " +
+    " STYLE='text-decoration:overline;'> ";
+  althtmldef "-\/" as
+    ' <span>&#8893;</span> ';
+  latexdef "-\/" as "\bar{\vee}";
```
benjub, line 431534:
```suggestion
  latexdef "-\/" as "\mathrel{\overline{\vee}}";
```

set.mm
```diff
@@ -431454,6 +431526,12 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
   althtmldef "\/_" as " &#8891; ";
     /* 2-Jan-2016 reverted sans-serif */
   latexdef "\/_" as "\veebar";
```
benjub, line 431528:
```suggestion
  latexdef "\/_" as "\mathrel{\underline{\vee}}";
```

set.mm
```diff
@@ -11655,6 +11656,52 @@ This definition (in the form of ~ dfifp2 ) appears in Section II.24 of
...
+  $( Define joint denial ("not-or" or "nor").  Represented by a vertical bar,
+     but we use a different symbol to avoid ambiguity with other uses of the
+     vertical bar.  After we define the constant true ` T. ` ( ~ df-tru ) and
+     the constant false ` F. ` ( ~ df-fal ), we will be able to prove these
+     truth table values: ` ( ( T. -\/ T. ) <-> F. ) ` ( ~ trunortru ),
+     ` ( ( T. -\/ F. ) <-> F. ) ` ( ~ trunorfal ), ` ( ( F. -\/ T. ) <-> F. ) `
+     ( ~ falnortru ), and ` ( ( F. -\/ F. ) <-> T. ) ` ( ~ falnorfal ).
+     Contrast with ` /\ ` ( ~ df-an ), ` \/ ` ( ~ df-or ), ` -> ` ( ~ wi ),
+     ` -/\ ` ( ~ df-nan ), and ` \/_ ` ( ~ df-xor ).  (Contributed by Remi,
+     25-Nov-2023.) $)
```
avekens, line 11680: This is in the future ;-)

set.mm
```diff
@@ -11655,6 +11656,52 @@ This definition (in the form of ~ dfifp2 ) appears in Section II.24 of
...
+  df-nor $a |- ( ( ph -\/ ps ) <-> -. ( ph \/ ps ) ) $.
+
+  $( The connector ` -\/ ` is commutative.  (Contributed by Remi,
+     25-Oct-2023.) $)
+  norcom $p |- ( ( ph -\/ ps ) <-> ( ps -\/ ph ) ) $=
+    ( wo wn wnor orcom notbii df-nor 3bitr4i ) ABCZDBACZDABEBAEJKABFGABHBAHI $.
+
+  $( ` -. ` is expressable via ` -\/ ` .  (Contributed by Remi,
```
avekens, line 11688: "expressible"?

set.mm
```diff
@@ -12167,6 +12214,31 @@ logical connectives we have introduced (either axiomatically or by a
     ( wfal wxo wtru wn wb df-xor falbifal xchbinx nottru bitri ) AABZCDAKAAECAA
     FGHIJ $.
 
+$(
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+  Joint denial
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  $( A ` -\/ ` identity.  (Contributed by Remi, 25-Oct-2023.) $)
```
avekens, line 12223: Maybe we should be a little bit more verbose here (I know the other theorems of this kind have also these short comments). I would suggest:

``A ` -\/ ` identity: true nor true is false.``

set.mm
```diff
@@ -11655,6 +11656,52 @@ This definition (in the form of ~ dfifp2 ) appears in Section II.24 of
...
+
+  $( Declare connective for joint denial ("nor"). $)
+  $c -\/ $.  $( Overlined "vee" (read:  "nor") $)
+
+  $( Extend wff definition to include joint denial ("nor"). $)
+  wnor $a wff ( ph -\/ ps ) $.
+
+  $( Define joint denial ("not-or" or "nor").  Represented by a vertical bar,
+     but we use a different symbol to avoid ambiguity with other uses of the
+     vertical bar.  After we define the constant true ` T. ` ( ~ df-tru ) and
```
benjub, line 11673: That second sentence was there for df-nan, but should be removed for df-nor.

set.mm
benjub: In general, I favor precise comments, like the one @avekens wrote, but in this particular case, I think the current comment is sufficient, similar to other comments in the section on truth tables.

benjub (approved):
Thanks, that's a nice submission.

Note to anyone: associativity is not proved yet !

benjub (changes requested):
Sorry, I shouldn't have approved: there is my comment above about removing the superfluous sentence from a comment.

remimimimimi:
> Note to anyone: associativity is not proved yet !

Because [it's a non-associative operation](https://proofwiki.org/wiki/NOR_is_not_Associative).

benjub:
> > Note to anyone: associativity is not proved yet !
> 
> Because [it's a non-associative operation](https://proofwiki.org/wiki/NOR_is_not_Associative).

Ha, that's a fair reason !  It should be some result like ~nanass.

benjub (approved):

jkingdon (approved):
Part of me wonders if we should have a section header like "other connectives" to group together the propositional logic connectives other than the main five ones (negation, implication, or, and, and biconditional). Looks like such a section would have about six (or so) subsections. I know Norm was a bit worried about proliferating these things, and I guess from my point of view that's largely a question of organization, rather than whether to have them at all.

But I'm fine with that being food for thought, I don't see any need to delay merging this pull request.

benjub:
> Part of me wonders if we should have a section header like "other connectives" to group together the propositional logic connectives other than the main five ones (negation, implication, or, and, and biconditional). Looks like such a section would have about six (or so) subsections. I know Norm was a bit worried about proliferating these things, and I guess from my point of view that's largely a question of organization, rather than whether to have them at all.

I think a dedicated section is not needed. As for the connectives, I think this MR added the final one ? Sure, technically, there is "does not imply", "is (not) implied by", and degenerated ones ((negated) projections, true/false as unary and binary operators). As for higher arity, I think ifp, had, cad are enough as well. (This is for classical logic.)

PR #3596, opened 2023-10-28
Theorems for Godel-sets (3)

avekens:
The proof that ( ( M Sat E ) ` N ) is a function, as indicated in the definition ~df-sat, which was missing in PR #3572, is available now (see ~satffun and ~satff).

The proof was really hard and tedious, requiring a lot of auxiliary theorems and lemmata. The crucial part was to find a representation of a disjoint extention of formulas/satisfaction predicates of height `N` to formulas/satisfaction predicates of height `N+1`, see ~satfvsucsuc, ~fmla0disjsuc, ~fmlasucdisj.

Details:

Changes in main set.mm:
* ~bj-2ex moved as ~2oex from BJ's mathbox
* ~ralrexbid, ~elneeldif, ~rexdifi added
* ~dmopabelb, ~dmopab2rex added
* ~releldmdifi, ~funfv1st2nd, ~funelss, ~funeldmdif added
* ~omsucne, ~1one2o added

Changes in MC's mathbox:
* theorems for "Godel-set for the Sheffer stroke NAND" added (~gonafv, ~gonanegoal)
* ~satfvsucsuc added
* theorems for Godel-formulas added (~fmlasssuc, ~fmlan0, ~gonan0, ~goaln0, ~gonar, ~goalr)
* theorems for disjoint sets of Godel-formulas added (~fmla0disjsuc, ~fmlasucdisj)
* main theorem ~satffun added (together with some lemmata)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -101,6 +101,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+28-Oct-23 bj-2ex    2oex        moved from BJ's mathbox to main set.mm
 23-Oct-23 bj-alequexv alequexv  moved from BJ's mathbox to main set.mm
 23-Oct-23 axext2    axexte      existential form of ax-ext
 23-Oct-23 axext3    axextg      more general form of ax-ext


set.mm
```diff
@@ -52644,6 +52670,44 @@ Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
...
+    $d A x y z $.  $d B i x y z $.  $d C x y z $.  $d D x y z $.  $d I x y z $.
+    $d U i x y z $.  $d V i x y z $.  $d W z $.  $d X i z $.  $d i u x y z $.
+    $d i v x y z $.
+    $( The domain of an ordered pair class abstraction with two nested
+       restricted existential quantifiers.  (Contributed by AV,
+       23-Oct-2023.) $)
+    dmopab2rex $p |- ( A. u e. U ( A. v e. V B e. X /\ A. i e. I D e. W )
+                       -> dom { <. x , y >. | E. u e. U ( E. v e. V
+                         ( x = A /\ y = B ) \/ E. i e. I ( x = C /\ y = D ) ) }
+                = { x | E. u e. U ( E. v e. V x = A \/ E. i e. I x = C ) } ) $=
```
tirix, line 52695: I don't have issues with the other theorems added to main, but this one seems to be too specific. I suppose you don't want to add such generic theorems directly to @digama0 's mathbox?

set.mm
```diff
@@ -72597,6 +72668,61 @@ Power Set ( ~ ax-pow ).  (Contributed by Mario Carneiro, 20-May-2013.)
...
+    GVSNZVNVRWAWHVJVRWAWHVIWAVKWHVIWAKVKVTAFZWHWAVKWIQVICVTARUKVIWIWHJWAVLVSAUI
+    SULUJTUMSUNWBBVMUOZVNWFWDQVRVJWJWAVRVJKBVRVJBDZVIVJWKJVKVJVIWKBAUPUQSTVBURV
+    MVLVSBUSPUTWBVOWDQZVNWAVRWLVJCVTBRVASVCVDVEVFVGVH $.
+
+  ${
+    $d A x $.  $d B x $.  $d C x $.
+    $( Two ways of expressing membership in the difference of domains of two
+       nested functions.  (Contributed by AV, 27-Oct-2023.) $)
+    funeldmdif $p |- ( ( Fun A /\ B C_ A ) -> ( C e. ( dom A \ dom B )
+                                   <-> E. x e. ( A \ B ) ( 1st ` x ) = C ) ) $=
```
tirix, line 72715: I believe this is also valid if `A` is a relation.

set.mm
```diff
@@ -72597,6 +72668,61 @@ Power Set ( ~ ax-pow ).  (Contributed by Mario Carneiro, 20-May-2013.)
...
+                      -> ( F ` ( 1st ` X ) ) = ( 2nd ` X ) ) $=
+    ( wfun wcel wa c1st cfv c2nd cop wceq wrel funrel 1st2nd sylan eleq1 adantl
+    wb wi funopfv adantr sylbid impancom mpd ) ACZBADZEBBFGZBHGZIZJZUFAGUGJZUDA
+    KUEUIALBAMNUDUIUEUJUDUIEUEUHADZUJUIUEUKQUDBUHAOPUDUKUJRUIUFUGASTUAUBUC $.
+
+  $( If the first component of an element of a function is in the domain of a
+     subset of the function, the element is a member of this subset.
+     (Contributed by AV, 27-Oct-2023.) $)
+  funelss $p |- ( ( Fun A /\ B C_ A /\ X e. A )
+                  -> ( ( 1st ` X ) e. dom B -> X e. B ) ) $=
```
tirix, line 72699: I believe this is also valid if `A` is a relation.

set.mm
```diff
@@ -504787,7 +504923,7 @@ required as (atomic) formulas, can be introduced as a defined notion in
      formulas, see ~ fmla .  (Contributed by Mario Carneiro, 14-Jul-2013.) $)
   df-fmla $a |- Fmla = ( n e. suc _om |-> dom ( ( (/) Sat (/) ) ` n ) ) $.
 
-$( --- theorems for ` ( I e.g J ) ` and ` A.g M A ` --- $)
+$( --- theorems for ` ( I e.g J ) `, ` ( A |g B ) ` and ` A.g M A ` --- $)
```
tirix, line 504926: I see you're only updating a pre-existing comment.
Did you consider turning this into a (sub-sub)-section?

set.mm
```diff
@@ -505542,7 +505759,7 @@ codes is an increasing chain (with respect to inclusion).  (Contributed
 
   ${
     $d f i j n u v x y $.
-    $( The set of all valid Goedel formulas.  (Contributed by AV,
+    $( The set of all valid Godel formulas.  (Contributed by AV,
```
tirix, line 505762: Ah, so you prefer `Godel` over `Goedel`? :)
Too bad we can't have Gödel!

set.mm
```diff
@@ -505773,6 +506254,424 @@ codes is an increasing chain (with respect to inclusion).  (Contributed
...
+    NVOVPWBVQVRVEUUBYFUUCWCUUAXHXMXSYCVSVGVTWAWDWEWFWBWGWHWIWHWJ $.
+
+  ${
+    $d E i j s u x y z $.  $d E i r s u x y z $.  $d E j s u v x y z $.
+    $d M i j s u x y z $.  $d M i r s u x y z $.  $d M j s u v x y z $.
+    $d N i j s u x y z $.  $d N i r s u x y z $.  $d N j s u v x y z $.
+    $d f i j s u y z $.  $d f i r s u y z $.  $d f j s u v y z $.
+    $d i j k s u y z $.  $d k r s u v y z $.
+    $( Lemma for ~ satffunlem1 .  (Contributed by AV, 17-Oct-2023.) $)
+    satffunlem1lem1 $p |- ( Fun ( ( M Sat E ) ` N )
```
tirix, line 506291: I understand what you're doing with `lem1lem1`, `lem2lem1`, usw. 
I think usually we are simply numbering `lem1` through `lem6` instead. 
Anyway I'm guilty of too (I've used `lema`, etc.)

tirix:
That's impressive, and obviously it was hard work!

set.mm
mazsa: Update "HTML code for accented names:"

tirix: Indeed we have `G&ouml;del`!

jkingdon: Yeah, some day perhaps we'll UTF-8-ize everything but that's a big project so unless/until we do that, `&#xf6;` or `&ouml;` is the way to go.

set.mm
jkingdon: I wouldn't especially push the numbers (they are kind of awkward especially if lemmas get added or removed) but I would advocate a name which avoids multiple levels of `lem`. Maybe `satfunlem1a` or `satfunlemsat` or something.

set.mm
avekens: I also was torn between putting it into main and @digama0 's mathbox (just before ~satffunlem1lem2, where it is used), but finally I thought it is kind of generally interesting (how to cope with existential quantifiers in an ordered pair abstraction), so I have put it into main. If there is no agreement to have it in main, I can move it into MC's mathbox.

~ dmopab3rexdif is a similar theorem, but this is really too specific and should be regarded as lemma for ~satffunlem2lem2.

set.mm
avekens: Unfortunately not (and this caused a lot of additional work): Let A = { <a,a'>, <a,a''>, <b,b'> } (A is a relation, but not a function) and B = { <a,a'> }, then B C_ A, dom A = { a , b }, dom B = { a }  and  ( dom A \ dom B ) = { b }, but ( A \ B ) = { <a,a''>, <b,b'> }. Let x = <a,a''>, then ( 1st ` x ) = a, but a e/ ( dom A \ dom B )!

set.mm
avekens: No, see example for ~funeldmdif .

set.mm
avekens: Since I am working within MC's mathbox in this case, I do not want to change the structure of it. This should be decided by @digama0. Or the whole section should be moved to main...

set.mm
avekens: I also would prefer Gödel or at least Goedel, but @digama0 definitions were about "Godel-sets", so for consistency, and because it is in MC's mathbox, I adhered to this terminology.

set.mm
avekens: Reading the conventions about lemmas:

>        A proof with the sole purpose of supporting a final proof is a lemma;
>        the naming convention for a lemma is the final proof label followed by
>        "lem", and a number if there is more than one.  E.g., ~ sbthlem1 is the
>        first lemma for ~ sbth .  The comment should begin with "Lemma for",
>        followed by the final proof label, so that it can be suppressed in
>        theorem lists (see the Metamath program "MM> WRITE THEOREM_LIST"
>        command).

my names obey these rules: the final proof (theorem) label is ~satffunlem1, and the sole purpose of ~satffunlem1lem1 is to support ~satffunlem1. 

Having a lot of lemmas, I think it is helpful for the understanding to allow at least one level of nesting (lemmas of lemmas). I agree that a deeper nesting (lemmas of lemmas of lemmas, etc.) would be more confusing than helpful.

set.mm
avekens: PS: one direction of the biconditional holds for arbitrary relations, see ~releldmdifi

set.mm
jkingdon: > Having a lot of lemmas, I think it is helpful for the understanding to allow at least one level of nesting (lemmas of lemmas). I

In that case I think we should go with your preference on that. Especially once we hit "lem" I think we can be flexible on these names.

set.mm
jkingdon: > Or the whole section should be moved to main...

Makes sense to me. Especially once we have multiple people working on it, keeping it in a mathbox doesn't seem all that appealing.

set.mm
tirix: I believe we should use `G&ouml;del` everywhere, that is what will give the correct rendering (although it impairs reading the ASCII source). 
In any case we can handle that separately. For now, keeping it consistent is what matters.

tirix (approved):

set.mm
avekens: So how/where should it be placed in main? I think a new "part" (as PART 17, following PART 16 GRAPH THEORY) is required for it, which could be titled "MODEL THEORY" (according to @digama0's comment in #3572). Its (currently one and only) section would be "Gödel-sets of formulas", containing the subsections "Primary Gödel-sets", "The Satisfaction predicate", "Gödel formulas", "The "proves" relation", "Additional Gödel-sets".

Additional candidates for this new PART would be the following sections in MC's mathbox:

- 20.6.12  Models of ZF  
- 20.6.13  Metamath formal systems  
- 20.6.14  Grammatical formal systems  
- 20.6.15  Models of formal systems  

I also detected section "Provability logic" in BJ's mathbox, where a "provability predicate" is defined and theorems for Gödel's second incompleteness theorem ( ~ bj-babygodel ) and Löb's  theorem ( ~ bj-babylob ) are provided. Would this also fit into this new PART?

@benjub @digama0 @david-a-wheeler what is your opinion?

I will open a new PR for a correspondig proposal, if nobody objects.

benjub: @avekens can you open an issue for this ? (you can simply paste your text there, as a starting point).

avekens: Issue #3599 is opened to discuss this (and finally to make a decision).

icecream17 (approved):

PR #3597, opened 2023-10-28
pm2.5

benjub:
This started with a simple shortening in iset, but I then noticed that several pm2.5* theorems are actually special instances of theorems, which I added.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -311,6 +311,7 @@ New usage of "opelopabsbALT" is discouraged (2 uses).
 New usage of "peano1nnnn" is discouraged (1 uses).
 New usage of "peano2nnnn" is discouraged (1 uses).
 New usage of "peano5nnnn" is discouraged (1 uses).
+New usage of "pm2.521dcALT" is discouraged (0 uses).
 New usage of "r19.2mOLD" is discouraged (0 uses).
 New usage of "ralxfrALT" is discouraged (0 uses).
 New usage of "rdgivallem" is discouraged (1 uses).
@@ -324,6 +325,7 @@ New usage of "sbieh" is discouraged (7 uses).
 New usage of "spimeh" is discouraged (0 uses).
 New usage of "spimh" is discouraged (1 uses).
 New usage of "spimth" is discouraged (1 uses).
+New usage of "stabtestimpdcOLD" is discouraged (0 uses).
 New usage of "stoic2a" is discouraged (0 uses).
 New usage of "stoic2b" is discouraged (0 uses).
 New usage of "strcollnfALT" is discouraged (0 uses).
@@ -438,6 +440,7 @@ Proof modification of "nn0ge2m1nnALT" is discouraged (48 steps).
 Proof modification of "nnindALT" is discouraged (15 steps).
 Proof modification of "opelopabsbALT" is discouraged (106 steps).
 Proof modification of "peano5set" is discouraged (60 steps).
+Proof modification of "pm2.521dcALT" is discouraged (21 steps).
 Proof modification of "r19.2mOLD" is discouraged (46 steps).
 Proof modification of "ralxfrALT" is discouraged (85 steps).
 Proof modification of "resfunexgALT" is discouraged (76 steps).
@@ -445,6 +448,7 @@ Proof modification of "ruALT" is discouraged (35 steps).
 Proof modification of "sbc8g" is discouraged (55 steps).
 Proof modification of "sbsbc" is discouraged (21 steps).
 Proof modification of "speano5" is discouraged (50 steps).
+Proof modification of "stabtestimpdcOLD" is discouraged (62 steps).
 Proof modification of "strcollnfALT" is discouraged (79 steps).
 Proof modification of "tfri1dALT" is discouraged (247 steps).
 Proof modification of "uzind4ALT" is discouraged (16 steps).


wlammen:
set.mm:  I think this is overdoing it a bit.  In my eyes conax1 and simplim are enough to describe the expression -. ( ph -> ps ) sufficiently in the implication/not-only section, maybe followed by this variation of pm2.27 explaining its construction: ( ph -> ( -. ps -> -. ( ph -> ps ) ) ) .  Negated implication in antecedent position is hardly used, so convenience theorems do not really pay off here.  As a compromise with literature like Whitehead & Russel some theorems like 2.51 are added, but I wouldn't go much further.

benjub:
> set.mm: I think this is overdoing it a bit. In my eyes conax1 and simplim are enough to describe the expression -. ( ph -> ps ) sufficiently in the implication/not-only section, maybe followed by this variation of pm2.27 explaining its construction: ( ph -> ( -. ps -> -. ( ph -> ps ) ) ) . Negated implication in antecedent position is hardly used, so convenience theorems do not really pay off here. As a compromise with literature like Whitehead & Russel some theorems like 2.51 are added, but I wouldn't go much further.

I agree that it's a bit overkill, and also that a negated implication as antecedent is not very useful. I wonder why pm2.51 and pm2.52 were added to Principia Mathematica. But since they were published in that famous book, we added them to set.mm, and now that they are here, we "have to"* put its general instance conax1k (which indeed I wouldn't have added if pm2.51, pm2.52 weren't there).

I think we may also add your proposal `|- ( ph -> ( -. ps -> -. ( ph -> ps ) ) )`.

By the "have to" above, I mean that, in my opinion, (i)set.mm (and actually, most databases), should follow the rule:

If a theorem is in the database and is an instance of a more general theorem, then all the most general instances of that theorem should be in the database.

wlammen:
> If a theorem is in the database and is an instance of a more general theorem, then all the most general instances of that theorem should be in the database.

This is a lofty goal.  It demands that you should provide a general *f version for any theorem with dv-conditions, where each `$d x y $.` is replaced with a distinctor, and each `$d ph z` is replaced with a `F/ z ph` antecedent.  Only then you will easily derive versions with distinctors, inequalities or dv-conditions, at will.

benjub:
> > If a theorem is in the database and is an instance of a more general theorem, then all the most general instances of that theorem should be in the database.
> 
> This is a lofty goal. It demands that you should provide a general *f version for any theorem with dv-conditions, where each `$d x y $.` is replaced with a distinctor, and each `$d ph z` is replaced with a `F/ z ph` antecedent. Only then you will easily derive versions with distinctors, inequalities or dv-conditions, at will.

Oh, I should have been clearer: by "instance", I mean "substitution instance" (possibly followed by addition of DV conditions).

wlammen:
You still need to explain what to do with convenience theorems.  They are per definition _special instances_ , deliberately created besides the general case.

benjub:
> You still need to explain what to do with convenience theorems. They are per definition _special instances_ , deliberately created besides the general case.

What do I need to explain about them ? (**edit:** if it's "what to do" about them: I think special instances don't need in general to be added, but it may depend on cases; I don't have strong opinions on that.)

jkingdon:
> now that they are here, we "have to"* put its general instance conax1k (which indeed I wouldn't have added if pm2.51, pm2.52 weren't there).

I'm not sure I'm sold on that premise, especially when `conax1k` is just three steps from existing theorems/axioms.

I mean, I'm not even sure we need to keep `pm2.51` and `pm2.52` but if we want them because they are in Principia, I would think that logic would apply to those exact theorems, not to variations thereof.

I don't know how much to lean on this, but there is always "a theorem should only be added if doing so reduces the total size of the database" (I forget the exact wording and can't find it right now but perhaps you remember the guideline I'm thinking of). Perhaps that gives at least some idea of when we should have more variations of existing theorems and when we should just stick with what we have.

jkingdon:
Of course if you wanted to put these things in your mathbox, then there would be far less worry about whether they are overkill.

avekens:
In my opinion, we should add the general versions proposed by @benjub in this PR as main theorems, and keep  ~pm2.51 and ~pm2.52 because of their historical relevance (containment in  Principia Mathematica). Maybe we can tag ~pm2.51 and ~pm2.52 with "(New usage is discouraged.)", and use the general theorems instead:
* pm2.51 is used in the proof of pm5.12 only
* pm2.52 is not used in any proof

icecream17:
Given that these theorems must exist there's not really any reason to limit proofs to not use them in [the rare scenario they are used](https://us.metamath.org/mpeuni/stoic4a.html). For example the entire df-nan df-xor df-nor section is _not_ discouraged. Rather, it just so happens that further development in less-used theorem structures is not necessary.

icecream17:
Ignoring conax1, conax1k, pm2.5gdc, pm2.521gdc, the rest of the changes are good.

(I'm unsure about replacing 'or' with disjunction; the comment of stabtestimpdc is about the same to me)

benjub:
I agree that if Principia Mathematica did not contain statements `pm2.51` and `pm2.52*`, then  the statements `pm2.51*` and `pm2.52*` and `conax1k` shouldn't be added to `(i)set.mm`. (Actually, I wonder why they made it into PM, because they indeed are not very natural, and as Wolf says, a negated implication as antecedent is not very useful).

But, if we want to respect the two rules:
1. Every theorem of PM (up to section xxx) should be in `set.mm` (and `iset.mm` when provable).
2. The database `(i)set.mm` should contain every most general instance (provable from axioms in `(i)set.mm`) of any statement it contains.

then we are more or less led to the additions proposed in this PR.

I think (1.) is a historical requirement, and now that it's here we might as well keep it, and in my opinion (2.) is a sensible rule, that could actually be extended to other databases.

avekens:
I think @benjub's arguments are plausible enough to accept this PR. As for rule (2.), I understand it in the way that if a more general theorem is proven (and used in other proofs), then it should be added to (i)set.mm. We do it already often if there is a theorem with dv restrictions only, but the corresponding thorem is needed with "not-free" condition (see also @wlammen's remark). This does not mean that all theorems with dv restrictions should be transformed now: this should really be done only if required.

benjub:
> As for rule (2.), I understand it in the way that if a more general theorem is proven (and used in other proofs), then it should be added to (i)set.mm.

I mean something more specific, limited, and precise: as I wrote above, but without detailing things enough: by "instance", I mean "substitution instance possibly followed by addition of DV conditions". Namely, Statement B is an instance of Statement A if there exists a substitution that turns A into B (and additionally, B may have additional DV conditions). This is equivalent to Statement B having a proof in exactly one essential non-hypothesis step, and that step is labeled A.

For example, if `|- ( ( ph -> ph ) -> ( ph -> ph ) )` was in set.mm, then we would need to add its two most general true instances, which are  `|- ( ( ph -> ps ) -> ( ph -> ps ) )` and  `|- ( ph -> ( ps -> ps ) )`.

See for instance `bijust0`/`bijust` or `monothetic`/`trujust` or `a1i`/`sptruw`. I added to many comments in the past precisions of that kind. See for instance `MM> SEARCH * "instance of" / COMMENTS`.

I acknowledge that this general rule should be discusssed and added to the conventions.

icecream17 (approved):
Given the convention-to-add (general forms regarding just substitution) I think that covers everything, then.

jkingdon:
I'll abstain on this one, since I don't really have enough of an opinion to say much more. There's no great disaster in adding more theorems than necessary, because they will (usually) tend to remain unused if they really are superfluous.

benjub:
> I'll abstain on this one, since I don't really have enough of an opinion to say much more. There's no great disaster in adding more theorems than necessary, because they will (usually) tend to remain unused if they really are superfluous.

I do not think they are superfluous, and I'm glad to learn that "there's no great disaster in" the careful and selfless work of:
* correcting the false statement "only holds for decidable propositions" [of course it's an "if", not an "only if"], to the correct and more precise "holds exactly for stable propositions, and in particular for decidable propositions"
* adding links between related theorems (e.g., the comment of ~expt now links to the related statements ~pm3.3 and ~ex) for ease of browsing
* being careful in improving some comments, without adding credit:
  * "Triple negation." --> "Triple negation is equivalent to negation.": correct sentence and more descriptive
  * "Every formula of the form `-. ph`" --> "Every negated formula": more concise
  * "Based on theorem *2.61" --> "Theorem *2.61 under a decidability condition" (three times): more precise
  * "'or'" --> "disjunction": grammar
  * "Logical 'or'" --> "Disjunction": grammar
  * "x = y" --> " ` x = y ` " : formatting
  * ""Stable and testable" is equivalent to decidable." --> "A formula is decidable if and only if its negation is decidable and it is stable (that is, it is testable and stable).": grammar, and clearer to readers with less background
* correcting the false statement "Importation theorem ~ imp" to "Importation theorem ~ pm3.1 (closed form of ~ imp )"
* adding general instances of existing theorems, and doing so, being careful to be systematic, being careful to synchronize things in set.mm and iset.mm, being careful to write clear, concise, descriptive, precise comments, with links to related theorems when I think they are useful to the reader, adding a new credit tag only when necessary (the new theorems iset/pm2.5gdc, set/pm2.5g, set/pm2.521g2, set/pm5.11g have no credits to me)
* making it clear from its proof that iset/pm2.52 is valid in minimal calculus
* shortening the proof of stabtestimpdc.

jkingdon:
Heh, sounds like I need to abstain even harder. All I really meant was that I don't object to anything here.

wlammen:
I hope nobody is waiting for me here.  I can only partially approve this, as I am not familiar with intuitionistic logic, and do not see the consequences there.  As far as set.mm is concerned, I do not object to merging this PR.

jkingdon:
One approval and no objections, so merging.

PR #3604, opened 2023-10-31
spimv1 -> spimfv

wlammen:
1. Labeling of spimfv consistent with spimefv.
2. Move spimfv and variants right after 19.45 (needs 19.36).

spimefv belongs to this group, too, but before nf5rd was moved, this is not possible.  Extra preparations are needed in subsequent PRs.


benjub (approved):
Easier to review indeed !

PR #3605, opened 2023-10-31
Prove acycgrsubgr

BTernaryTau:
subgrwlk

> If a walk exists in a subgraph of a graph ` G ` , then that walk also exists in ` G ` .

subgrtrl

> If a trail exists in a subgraph of a graph ` G ` , then that trail also exists in ` G ` .

subgrpth

> If a path exists in a subgraph of a graph ` G ` , then that path also exists in ` G ` .

subgrcycl

> If a cycle exists in a subgraph of a graph ` G ` , then that cycle also exists in ` G ` .

acycgrsubgr

> The subgraph of an acyclic graph is also acyclic.


tirix (approved):
Perfect submission as always.

icecream17 (approved):

wlammen (approved):

PR #3608, opened 2023-11-02
the union of countably many countable sets is countable in iset.mm

### LINKED ISSUE #3394: is the union of countably many countable sets countable?

jkingdon:
As discussed in https://github.com/metamath/set.mm/issues/3394#issuecomment-1786278814 and the `ctiunct` comment being added via this pull request, this is provable only based on a certain way of stating the countability of the sets in the collection.

Also proves that the union of two countable sets is countable which does not have those complications.

Also brings iset.mm into closer alignment with set.mm in a few places, some of which are used in this proof, some of which I just noticed while doing that.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4992,6 +4992,11 @@ this implies excluded middle</TD>
 </tr>
 
 <tr>
+  <td>unctb</td>
+  <td>~ unct</td>
+</tr>
+
+<tr>
   <td>infdjuabs</td>
   <td><i>none</i></td>
   <td>the set.mm proof uses sbth and other theorems we don't have</td>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2073,6 +2073,12 @@ favor of theorems in deduction form.</TD>
   <TD>~ iinrabm</TD>
 </TR>
 
+<tr>
+  <td>iunxdif3</td>
+  <td><i>none</i></td>
+  <td>the set.mm proof relies on inundif</td>
+</tr>
+
 <TR>
   <TD>iinuni</TD>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -27453,6 +27453,28 @@ practical reasons (to avoid having to prove sethood of ` A ` in every use
...
 
+  $( Demonstrate by witnesses that two classes lack a subclass relation.
+     (Contributed by Stefan O'Rear, 5-Feb-2015.) $)
+  nelss $p |- ( ( A e. B /\ -. A e. C ) -> -. B C_ C ) $=
+    ( wcel wss ssel com12 con3dimp ) ABDZBCEZACDZJIKBCAFGH $.
+
+  ${
+    ssrexf.1 $e |- F/_ x A $.
+    ssrexf.2 $e |- F/_ x B $.
+    $( restricted existential quantification follows from a subclass
```
icecream17, line 27464:
```suggestion
    $( Restricted existential quantification follows from a subclass
```
(this is also the case in set.mm, optional)

icecream17 (approved):

iset.mm
jkingdon: Yeah, it wasn't new, but sure, I'll change it.

PR #3614, opened 2023-11-05
Reword comment for ctiunct

jkingdon:
The comment being revised here tried to distinguish between this theorem and the one which would use ∃ for each countable set, but hopefully this wording is better.

Inspired by a Mastodon post by @andrejbauer which is at https://mathstodon.xyz/@andrejbauer/111359031129727739 which contains the words "a sequence of enumerations gives an enumeration of the union" but everything else here is my attempt to elaborate on it. Further suggestions for wording tweaks would be appreciated.

The original wording from @digama0 was "The union of countably many (constructively) countable sets is countable." but at least for my tastes that doesn't do enough to explain the difference between the "sequence of enumerations"  and the "countably many countable sets" version.

Our intended audience includes people who don't (yet) know much of the relevant mathematics, and certainly includes people who don't (yet) know anything about constructive mathematics, but we also don't want to dumb things down or include inaccurate oversimplifications. The "~ unct" syntax will turn into a link to https://us.metamath.org/ileuni/unct.html once this is merged.


iset.mm
```diff
@@ -129911,11 +129911,16 @@ then the Limited Principle of Omniscience (LPO) implies excluded middle.
...
+    $( A sequence of enumerations gives an enumeration of the union.  We refer
+       to "sequence of enumerations" rather than "countably many countable
+       sets" because the hypothesis does not provide countability for each
+       ` B ( x ) ` in isolation (independent of how it is enumerated), but in
+       terms of the ` B ( x ) ` together with the ` G ( x ) ` which enumerates
+       it.  Countable choice would be needed to turn the "countably many
+       countable sets" version into the "sequence of enumerations" version.
+
+       Compare with when we have two sets instead of countably many, as seen at
+       ~ unct , in which we can express countability using ` E. ` .
```
benjub, line 129923:
```suggestion
       Compare with the case of two sets instead of countably many, as seen at
       ~ unct , in which case we can express countability using ` E. ` .
```

iset.mm
```diff
@@ -129911,11 +129911,16 @@ then the Limited Principle of Omniscience (LPO) implies excluded middle.
...
-    $( The union of countably many countable sets is countable.  The
-       construction here is constructive in the sense that the hypothesis
-       provides a specific class for ` G ` in terms of ` x ` , as opposed to a
-       construction of the form ` A. x e. A E. g ` .  Countable choice would be
-       needed to turn the latter into the former.
+    $( A sequence of enumerations gives an enumeration of the union.  We refer
+       to "sequence of enumerations" rather than "countably many countable
+       sets" because the hypothesis does not provide countability for each
+       ` B ( x ) ` in isolation (independent of how it is enumerated), but in
+       terms of the ` B ( x ) ` together with the ` G ( x ) ` which enumerates
```
benjub, line 129918: I'm not sure I understand the wording. I think you refer to "sequence of enumerations", simply because this is what the hypotheses gives. You may even write "an A-indexed sequence of enumerations" to be more explicit.
The hypothesis actually does provide countability of each B(x) in isolation.

iset.mm
```diff
@@ -129911,11 +129911,16 @@ then the Limited Principle of Omniscience (LPO) implies excluded middle.
...
+       it.  Countable choice would be needed to turn the "countably many
+       countable sets" version into the "sequence of enumerations" version.
```
benjub, line 129920: Maybe "To obtain a sequence of enumerations from countably many enumerations requires the axiom of countable choice." ?

iset.mm
jkingdon: I too am not sure whether "isolation" is the clearest way to say it, but this quote from Bauer's Mastodon post is the sense in which "countability of each B(x)" is not what the hypothesis says:

> The problem is, if you make countability a structure (a set together with a specific enumeration) as opposed to a property (a set for which there exists an enumeration), then you also change equality: the same set with two different given enumerations becomes two different things.

iset.mm
jkingdon: Suggestion accepted and pull request updated.

iset.mm
benjub: Oh, I see. Indeed, it is not what the hypothesis says. The hypothesis says (strictly) more, so the "does not provide" above is misleading.

iset.mm
jkingdon: At least to my eyes that wording would draw attention away from the existential quantifier in the unct-style (as opposed to ctiunct-style) version.

The existential quantifier is part of what we mean by countability in places like https://us.metamath.org/ileuni/ctm.html and https://us.metamath.org/ileuni/ctssdc.html

iset.mm
jkingdon: Ah yes, good catch about "does not provide" - that's not a good way to say it when we mean "proves more than just". I've pushed up a change to revise the wording to hopefully make it clear that the ctiunct hypotheses provides as much as the other one and more. While I was at it, I explicitly stated the alternate hypothesis - hopefully that is clearer than letting the reader try to figure it out.

iset.mm
benjub: I agree. I actually wanted to be more precise. In the expression "countably many countables", you have a non-constructive (or "non-explicit") notion at two places, and in "enumeration of enumerations", you have a constructive notion at two places. Therefore, to go from one to the other, you have two steps. The expression I wrote above, "contably many enumerations", is midway. And I think the critical step, the one which requires the axiom of coutable choice, is the step from that midway to the "enumeration of enumerations" version. I hope this makes sense ? (and that it is correct !)

jkingdon: Take a look with a fresh pair of eyes (that is, not tonight) - for one thing I did make a revision here after you wrote that.

iset.mm
andrejbauer: I have on occasion found it useful to distinguish between "enumerable set" (a set for which an enumeration exists) and "enumerated set" (a set together with a specific enumeration). It may make sense to make the distinction precise in your formalization if you're working with both kinds of objects. The theorem in question then becomes "the union of an enumerated family of enumerated sets is enumerated". The other theorem (which requires countable choice) is "the union of an enumerable family of enumerable sets is enumerable".

P.S. You could replace "enumerable" with "countable", but then "enumerated" should be replaced with "counted" – which for some reason sounds strange (to me).

jkingdon: > I have on occasion found it useful to distinguish between "enumerable set" (a set for which an enumeration exists) and "enumerated set" (a set together with a specific enumeration). It may make sense to make the distinction precise in your formalization if you're working with both kinds of objects.

Well it is precise in the sense that we define ` G ` one way in https://us.metamath.org/ileuni/ctiunct.html and in #3618 we replace `G` using `∃ g`. We aren't always keen to introduce new notation (partly for metamath-specific reasons), so we're probably working with existing notation somehow. I guess we could have an ordered pair of the set and its enumeration. Maybe worth thinking about but I'd probably hold off until we are more sure we need it (for one thing there would be multiple ways to define the enumeration, given theorems like https://us.metamath.org/ileuni/ctssdc.html ).

As for the terminology, I agree that "counted" is ..... well maybe not a horrible idea but presumably too innovative for a project which tries to follow the literature. And for similar reasons I doubt we'd banish "countable" from our vocabulary entirely.

As for whether to add versions other than the https://us.metamath.org/ileuni/ctiunct.html one and the one proposed at #3618 , I experimented a bit with replacing ` ∃ℎ ℎ:ω–onto→(∪ 𝑥 ∈ 𝐴 𝐵 ⊔ 1o))` in the conclusion with a hypothesis H = _something_ and making the conclusion `H : ω–onto→(∪ 𝑥 ∈ 𝐴 𝐵 ⊔ 1o))`. This would be analogous to https://us.metamath.org/ileuni/ctssdccl.html but it seemed to make the theorem very verbose to state, somewhat awkward to prove, and I wasn't getting excited about whether the whole exercise was especially satisfying unless we have a specific proof which needs that form. If there are any other variations which are different in any important way, I didn't spot them: `𝐹:ω–onto→(𝐴 ⊔ 1o))` versus `∃f f:ω–onto→(𝐴 ⊔ 1o))` in the hypothesis is a trivial difference in light of theorems like https://us.metamath.org/ileuni/exlimddv.html .

digama0: > As for whether to add versions other than the https://us.metamath.org/ileuni/ctiunct.html one and the one proposed at https://github.com/metamath/set.mm/issues/3618 , I experimented a bit with replacing  ∃ℎ ℎ:ω–onto→(∪ 𝑥 ∈ 𝐴 𝐵 ⊔ 1o)) in the conclusion with a hypothesis H = something and making the conclusion H : ω–onto→(∪ 𝑥 ∈ 𝐴 𝐵 ⊔ 1o)). This would be analogous to https://us.metamath.org/ileuni/ctssdccl.html but it seemed to make the theorem very verbose to state, somewhat awkward to prove, and I wasn't getting excited about whether the whole exercise was especially satisfying unless we have a specific proof which needs that form. If there are any other variations which are different in any important way, I didn't spot them: 𝐹:ω–onto→(𝐴 ⊔ 1o)) versus ∃f f:ω–onto→(𝐴 ⊔ 1o)) in the hypothesis is a trivial difference in light of theorems like https://us.metamath.org/ileuni/exlimddv.html .

Using an existential is fundamentally a weaker statement, you may not be able to use exlimdv exactly because this might be part of a larger construction (e.g. `F` depends on some other parameter in scope) and you need a parameterized family of counting functions so that you don't get stuck with a forall-exists statement which can't be turned into an exists-forall statement after the fact (this is exactly the issue that arises in this very theorem). An example where I had to go to great lengths to do this in set.mm is https://us.metamath.org/mpeuni/infxpenc.html . Yes it looks horrible to describe the function, but you actually need this when you are constructing a parameterized family. In type theory this is encoded by using Sigma types in place of "mere existentials".

jkingdon: Good point about F depending on a parameter, I had missed that detail.

Agreed about both `Yes it looks horrible to describe the function` and `you actually need this when you are constructing a parameterized family. In type theory this is encoded by using Sigma types in place of "mere existentials".` This is the technique I used at https://us.metamath.org/ileuni/ctssdccl.html

tirix: Would it be worth to experiment with a new syntax "there exists a constructible" ?

iset.mm
```diff
@@ -129911,11 +129911,18 @@ then the Limited Principle of Omniscience (LPO) implies excluded middle.
...
+       sets" because the hypothesis provides more than countability for each
+       ` B ( x ) ` : it refers to ` B ( x ) ` together with the ` G ( x ) `
+       which enumerates it.
+
+       The "countably many countable sets" version could be expressed as
+       ` ( ph /\ x e. A ) -> E. g g : _om -onto-> ( B |_| 1o ) ` and countable
+       choice would be needed to derive the current hypothesis from that.
+
+       Compare with the case of two sets instead of countably many, as seen at
+       ~ unct , in which case we can express countability using ` E. ` .
```
icecream17, line 129925:
```suggestion
       ~ unct , in which case we express countability using ` E. ` .
```

This is merely a nitpick. It may be clearer to explicitly say that countability is the existence of an enumeration

icecream17 (approved):
Overall the change is clear and more accurate.

Note that the above review on "countability ... in isolation" is outdated and those words do not exist anymore.

(Incidentally while "counted" may be too novel, it is a great solution and I could see this being used if this situation comes up more often)

iset.mm
jkingdon: The word "can" was intended to emphasize that it isn't possible in other cases, but :shrug: suggestion accepted.

jkingdon:
> Overall the change is clear and more accurate.

I think that's sufficient for me for now. Thanks to those who provided suggestions - they have improved the wording.

> Note that the above review on "countability ... in isolation" is outdated and those words do not exist anymore.

Yeah it can be hard to miss the word "outdated" in the github user interface, but when the conversation has moved on a bit from the original specific issue you don't always want to hide the comment either.

> (Incidentally while "counted" may be too novel, it is a great solution and I could see this being used if this situation comes up more often)

Maybe this innovation will come to a paper of @andrejbauer some day.

iset.mm
benjub: The various comments since last time clarified things, and therefore the replacement I proposed at the top of this thread is indeed more precise (namely, "To obtain a sequence of enumerations from countably many enumerations requires the axiom of countable choice.")

PR #3620, opened 2023-11-08
unineqg

mazsa:


icecream17:
Is unineqg going to be used?

mazsa:
> Is unineqg going to be used?

Yes, I'll use it directly to show that
|- (  ( dom R = dom S /\ ran R = ran S ) -> ( ( ( dom R X. ran R ) \ R ) = ( ( dom S X. ran S ) \ S ) <-> R = S ) ) etc.

icecream17 (approved):

mazsa:
@icecream17 I have changed the status of this PR to Draft, because I made a cut&paste mistake and gave the wrong answer to your question.

What I can prove with this is unfortunately not what I wrote (and would need), but

|- ( ( Rel R /\ Rel S ) -> ( ( ( dom R X. ran R ) \ R ) = ( ( dom S X. ran S ) \ S ) ->
      ( ( dom R X. ran R ) = ( dom S X. ran S ) <-> R = S ) ) )

(via my ~ relundmrn  |- ( Rel R -> ( R u. ( dom R X. ran R ) ) = ( dom R X. ran R ) ) )

The generalization is still valid and this still shows that there may be a use case for it, but it is far from what I intended it to be used for. What do you suggest what should happen next with this PR?

icecream17:
If it's not going to be used, I guess closing or moving to a mathbox would be fine. It could always be added later

mazsa:
@icecream17 thanks for your advice. I decided to close this, because I also do not like to see free floating theorems, even in my Mathbox.

PR #3623, opened 2023-11-11
Intuitionize limccnp2

jkingdon:
The theorem https://us.metamath.org/mpeuni/limccnp2.html can be proved as stated (with one minor change in notation). However, the proof is via several epsilon-delta constructions, rather than the set.mm proof which is comparing real numbers for equality a lot.

Includes rpmaxcl , limcmpted and ellimc3apf which are similar to theorems iset.mm already has.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10847,7 +10847,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
-  <td>limcmpt , limcmpt2</td>
+  <td>limcmpt</td>
+  <td>~ limcmpted</td>
+</tr>
+
+<tr>
+  <td>limcmpt2</td>
   <td><i>none</i></td>
 </tr>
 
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -140477,6 +140525,169 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
+                                                   ( ( abs o. - )
+                                                   |` ( Y X. Y ) )
+                                                   s )
+                                                 <
+                                                 L )
+                                           -> ( ( C H D )
+                                                ( abs o. - )
+                                                ( r H s ) )
+                                              <
+                                              E ) ) $.
```
avekens, line 140567: this sems to be mmj2 style. In set.mm, this hypothesis (and the following ones) should be compressed in fewer lines.

jkingdon: Yeah, I'll fix it, just forgot.

jkingdon: Now fixed.

avekens (approved):

PR #3624, opened 2023-11-11
Rename syl5breqr to breqtrrid

jkingdon:
This is in both iset.mm and set.mm

As proposed in changes-set.txt


tirix (approved):
Looks good to me!

icecream17 (approved):

PR #3625, opened 2023-11-12
Mathbox for RP.

arpie-steele:
29 new theorems on ordinals and sets of small cardinality.
Also, iscard4 laments the definition that might have been.

A separate commit minimizes other theorems in mathbox.


avekens (approved):
Thre are some interesting new theorems which I may use for my current work. If I can use them, I will move them to main (thanks for the hints where they can be placed).

icecream17 (approved):

PR #3627, opened 2023-11-13
simplify df-selv

icecream17:
justified by [selvval2lem2](https://us.metamath.org/mpeuni/selvval2lem2.html) (deleted in this pr)

update corresponding theorems

remove remark about no theorems using the definitions


tirix (approved):
Nice!

PR #3630, opened 2023-11-14
Repair typos in dates.

arpie-steele:
Fix for #3628 (which explains how corrected dates were determined.)

dfcoeleqvrels, dfcoeleqvrel -> 28-Nov-2022
2mulprm -> 8-Jun-2023

Issue #3628, opened 2023-11-13
3 Bad dates


arpie-steele:
```
Theorem       Bad Date     Blame Date   BlameRev
dfcoeleqvrels 22-Nov-2028  21-Jul-2023  29b322f0c22c8ab889d6236537b80f5137e822e8
dfcoeleqvrel  22-Nov-2028  21-Jul-2023  29b322f0c22c8ab889d6236537b80f5137e822e8
2mulprm        8-Jun-2030  23-Jul-2023  bc266806bf678083cdebd55364c186284f57feb8
                            1-Jul-2023  0ccfa19b47ec9074801a002db000a8428f2aed81
```

Suggestion: 2mulprm -> 8-Jun-2023 (based on 0ccfa19b47ec9074801a002db000a8428f2aed81 and assuming year only is wrong)
Suggestion: dfcoeleqvrels -> 28-Nov-2022 (same as df-coeleqvrels, which is also in 29b322f0c22c8ab889d6236537b80f5137e822e8 and assuming transposed day/year digits)
Suggestion: dfcoeleqvrel -> 28-Nov-2022

arpie-steele:
Request for comment from @mazsa  and @avekens

mazsa:
>  Suggestion: dfcoeleqvrels -> 28-Nov-2022
>  Suggestion: dfcoeleqvrel -> 28-Nov-2022

@arpie-steele Good catches. If you open a PR for this, I'll accept your suggestions, or I'll fix it in one of my next PRs. Thank you.

avekens:
> Suggestion: 2mulprm -> 8-Jun-2023 

OK for me, please correct it in #3630

arpie-steele:
Upon further review, two more typos were found with obvious likely replacements.

Found: 15-Oct-1013 starting in fda18f6019 so likely it should be 15-Oct-2013
Found: 23-Apr-1017 starting in 0f3533bfd5 so likely it should be 23-Apr-2017

This is all I found in today's `set.mm` and `nf.mm` by looking for out-of-range years. Looking for out-of-range month or day-of-month yielded no results for dates that matched [0-9]{1,2}-[A-Z][a-z][a-z]-[0-9]{4}.

jkingdon:
Looks like everything mentioned here is fixed.

If there are unresolved issues go ahead and re-open this issue (or perhaps better, create a new issue with whatever is left).

PR #3632, opened 2023-11-15
Repair old marker usage of 1013 and 1017

arpie-steele:
Repair usage of 11th century dates with dates determined to be consistent with those added simultaneously in old history imports or the date of the original commits. Follow up for PR #3630, additional fix for Issue #3628.

See commits fda18f6019 and 0f3533bfd5 for where these dates were added originally to `set.mm` before `nf.mm` was created. This was once a historical practice used to promote the contribution of others on the list of new theorems.

### DIFF
--- a/nf.mm
+++ b/nf.mm
@@ -12538,7 +12538,7 @@ $)
     spimfw.2 $e |- ( x = y -> ( ph -> ps ) ) $.
     $( Specialization, with additional weakening to allow bundling of ` x ` and
        ` y ` .  Uses only Tarski's FOL axiom schemes.  (Contributed by NM,
-       23-Apr-1017.)  (Proof shortened by Wolf Lammen, 7-Aug-2017.) $)
+       23-Apr-2017.)  (Proof shortened by Wolf Lammen, 7-Aug-2017.) $)
     spimfw $p |- ( -. A. x -. x = y -> ( A. x ph -> ps ) ) $=
       ( weq wn wal wex speimfw df-ex con1i sylbi syl6 ) CDGHCIHACIBCJZBABCDFKPB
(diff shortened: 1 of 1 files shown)


wlammen (approved):

digama0:
By the way, the usage of 11th century dates is not a typo. I will pull up the exact quote if I can find it, but I recall Norm saying that he would use this as a way to commit new theorems (and record the date) without having them show up on mmrecent, potentially swamping the results of other contributors.

EDIT: here's a quote from a message sent to me by NM in 2014:

> > PS: You appear to have a bug in your processing code or something that
> > keeps putting year "1014" in the dates of your theorems (and somehow my
> > attribution comment in dvhlveclem got changed to year 3014). Unless
> > metamath is much older than I suspected, this is probably a mistake, and
> > I've been correcting them as I see them, but they keep coming back so I
> > suspect that some automation is doing the deed. Do you know what might
> > be causing this?
> 
> There is no bug.  I have been manually changing recent additions to my
mathbox to 1014 so they won't dominate the Most Recent page.  I don't
want to bury others' contributions in my noise, and anyway my mathbox
isn't really ready for public consumption yet.  Periodically I change
them back to 2014 when they roll off of the 200-theorem list.  A few
days ago I changed 1014 to 3014 on a lark, but decided to go back to
1014 - someday I may include future dates on the Most Recent page since
Wolf tends to submit things in early morning Germany time, when for us
it is tommorrow.

icecream17 (approved):

tirix (approved):

arpie-steele:
Thank you @digama0 for the research. I have updated the PR comment and title to remove "typo."

PR #3636, opened 2023-11-17
extract rnasclassa from selvval2lem3 + remove extra hypothesis from psrbagev2 + prove rnrhmsubrg

icecream17:
miscellaneous changes I made while trying to figure out the value of df-evls
commit by commit

I was going to add evlsval3 based on evlslem1 but then I noticed evlslem1 (and evlslem6) have an extra hypothesis. So I got distracted in trying to remove them but it turns out doing this is such a drag (I don't have access to mmj2 right now).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15124,6 +15124,7 @@ New usage of "enrbreq" is discouraged (3 uses).
 New usage of "enreceq" is discouraged (9 uses).
 New usage of "enrer" is discouraged (8 uses).
 New usage of "enrex" is discouraged (9 uses).
+New usage of "ensucne0OLD" is discouraged (0 uses).
 New usage of "epelgOLD" is discouraged (0 uses).
 New usage of "eqeq1dALT" is discouraged (0 uses).
 New usage of "eqeqan12dALT" is discouraged (0 uses).
@@ -18503,6 +18504,7 @@ Proof modification of "elunirnALT" is discouraged (38 steps).
 Proof modification of "en3lpVD" is discouraged (147 steps).
 Proof modification of "en3lplem1VD" is discouraged (95 steps).
 Proof modification of "en3lplem2VD" is discouraged (267 steps).
+Proof modification of "ensucne0OLD" is discouraged (82 steps).
 Proof modification of "epelgOLD" is discouraged (136 steps).
 Proof modification of "eqeq1dALT" is discouraged (62 steps).
 Proof modification of "eqeqan12dALT" is discouraged (23 steps).
--- a/discouraged
+++ b/discouraged
@@ -13326,6 +13326,7 @@ New usage of "ancomstVD" is discouraged (0 uses).
 New usage of "anmp" is discouraged (11 uses).
 New usage of "archnq" is discouraged (1 uses).
 New usage of "arglem1N" is discouraged (0 uses).
+New usage of "ascldimulOLD" is discouraged (0 uses).
 New usage of "atabs2i" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -250868,22 +250876,13 @@ such that every prime ideal contains a prime element (this is a
     $( The scalar injection is a ring homomorphism.  (Contributed by Mario
        Carneiro, 8-Mar-2015.) $)
     asclrhm $p |- ( W e. AssAlg -> A e. ( F RingHom W ) ) $=
```
avekens, line 250878: A "Proof shortened by" tag should be added here (and the old proof shoulds be kept as OLD theorem)

set.mm
```diff
@@ -254514,12 +254510,23 @@ This function maps an element of the formal polynomial algebra (with
...
+    psrbagev2.g $e |- ( ph -> G : I --> C ) $.
+    psrbagev2.i $e |- ( ph -> I e. _V ) $.
     $( Closure of a sum using a bag of multipliers.  (Contributed by Stefan
        O'Rear, 9-Mar-2015.)  (Proof shortened by AV, 18-Jul-2019.) $)
     psrbagev2 $p |- ( ph -> ( T gsum ( B oF .x. G ) ) e. C ) $=
-      ( cof co cvv wf cfsupp wbr psrbagev1 simpld simprd gsumcl ) AICBHFSTZEUAJ
-      LNORAICUIUBZUIJUCUDZABCDEFGHIJKLMNOPQRUEZUFAUJUKULUGUH $.
+      ( cof co cvv c0g cfv eqid wf cfsupp wbr psrbagev1 simpld simprd gsumcl )
+      AICBHFQRZESETUAZKUKUBZMPAICUJUCZUJUKUDUEZABCDEFGHIUKJKLULMNOPUFZUGAUMUNUO
+      UHUI $.
```
avekens, line 254529: Why is this proof longer than the old one?

icecream17: the hypothesis [psrbagev1.z](https://us.metamath.org/mpeuni/psrbagev2.html) was replaced with an eqid

set.mm
icecream17: In the case of asclrhm, all that was changed was replacing [step 54](https://us.metamath.org/mpeuni/asclrhm.html) with ascldimul (+ 3expb)

Maybe ascldimulOLD (from asclrhm) should be added?

tirix (approved):

set.mm
wlammen: It seems the unresolved issue here blocks the finishing of this PR for many days now.  Are you satisfied with this answer @avekens ?

avekens: I still do not understand why no "Proof shortened by" tag should be added for ~asclrhm, and the old proof should be kept as theorem asclrhmOLD (according to our rules/conventions), because the proof changed significantly.

And the part of the proof replaced by ascldimul is an alternate proof for ascldimul (and should be kept as (ascldimulOLD)? If there is nothing interesting in this proof, it needs not be kept.

tirix (approved):

tirix:
Still approving! :-)

set.mm
icecream17: The proof is pretty much exactly the same, analogous to extracting a subproof (which here is ascldimul) so that's why no old version was made. Indeed if 3expb wasn't needed the auto minimizer would probably have used ascldimul.

But I will add the ascldimulOLD for comparison

avekens (approved):

PR #3637, opened 2023-11-17
Minimized proofs with greater depth

GinoGiotto:
I rescanned the first 1000 theorems, but this time with the `improveDepth` flag set to 1. This allows to find further shortenings at the "moderate" cost of a 125x increase in minimization time, so most likely I won't use this version ever again.


icecream17 (approved):

wlammen:
Not all results are convincing.  Here's the most difficult one (pm4.71i)
old:
pm4.71i.1 ⊢ (𝜑 → 𝜓)
pm4.71 ⊢ ((𝜑 → 𝜓) ↔ (𝜑 ↔ (𝜑 ∧ 𝜓)))
mpbi ⊢ (𝜑 ↔ (𝜑 ∧ 𝜓))

new:
pm4.71i.1 ⊢ (𝜑 → 𝜓)
ancli ⊢ (𝜑 → (𝜑 ∧ 𝜓))
simpl ⊢ ((𝜑 ∧ 𝜓) → 𝜑)
impbii ⊢ (𝜑 ↔ (𝜑 ∧ 𝜓))

The new needs more proof lines, and uses more symbols.  So what exactly is the advantage?

In some instances like im2anan9 the proofs look visually more or less the same size, but unlike the example above there is no visual downgrade.  The majority of changes are improvements, both in compressed proof size and visually.

GinoGiotto:
For this PR I looked at the number of upper case characters of the compressed format, throwing away all the ones that kept this amount unchanged.  The new compressed proof of `pm4.71i` has one less uppercase character, however I actually agree with you that one more essential step doesn't look good. So reverted.

The new proof of `im2anan9` has the uppercase section shortened by two characters (making the whole compressed proof shorter), also the amount of labels in the label list is reduced (which is a criteria present in conventions). If there are no trade-offs I don't see why discard it.

wlammen:
Here is the proof of im2anan9 for comparison:

old:
im2an9.1 ⊢ (𝜑 → (𝜓 → 𝜒))
adantr ⊢ ((𝜑 ∧ 𝜃) → (𝜓 → 𝜒))
im2an9.2 ⊢ (𝜃 → (𝜏 → 𝜂))
adantl ⊢ ((𝜑 ∧ 𝜃) → (𝜏 → 𝜂))
anim12d ⊢ ((𝜑 ∧ 𝜃) → ((𝜓 ∧ 𝜏) → (𝜒 ∧ 𝜂)))

new:
im2an9.1 ⊢ (𝜑 → (𝜓 → 𝜒))
adantrd ⊢ (𝜑 → ((𝜓 ∧ 𝜏) → 𝜒))
im2an9.2 ⊢ (𝜃 → (𝜏 → 𝜂))
adantld ⊢ (𝜃 → ((𝜓 ∧ 𝜏) → 𝜂))
anim12ii ⊢ ((𝜑 ∧ 𝜃) → ((𝜓 ∧ 𝜏) → (𝜒 ∧ 𝜂)))

The new one has indeed a 2 bytes smaller compressed proof line, but the proof idea is virtually the same.  Whether you start with the first or the second antecedent of the result before doing the rest in a final step, is of no importance to a casual human reader.  Very deep under the hood one might detect some implementation dependent technical difference.

OK, so one proof is exchanged with a more or less equivalent one, so what is the problem then?  Very deep under the hood of my cerebrum I sense a feeling of uneasiness.  The author of the original proof picked one variant perhaps just because it sprang to his mind first, but maybe he had some other reason, too.  Is one variant more his style than the other?  Does he follow a common proof pattern that is now disturbed?  We don't know.  Like with written books authorship gives you some freedom of expression.  It is in part in respect of this freedom why I think one should not lightly change the contents covered by a contributed by... tag.

At least I want to stop for a moment here to see what others think.

avekens (approved):

avekens:
Any change of a proof affects its style, so the only criteria for changing a proof should be that it becomes shorter or uses less axioms (our current rules in short). If a contributor wants that his/her proof should remain unchanged, she/he should tag it with "(Proof modification is discouraged.)". Therefore, I see no problem with ~im2anan9.

wlammen (approved):
One should still note that one proof was reverted, although it was an acceptable shorter proof, and another proof, adding nothing from a pure mathematical point of view, supersedes its predecessor.  This PR tests the limits of our current rules for proof changes, in my eyes.  One should keep an eye on the situation, in particular if automatic proving becomes more frequent. The reverted proof is by the way a candidate again for any future automatic minimization, and may well be replaced then, despite the decision made here.

I think, for most of the audience of Metamath the visual impression on the Web Site is indicative of how short a proof is.  In my profession we speak of exposing implementation details, if results do not strictly follow the intended semantics of its domain.  Do we see examples here?  I'd say so.

That all said, I don't want to block this PR any longer, as it is not bad in total, I think.

PR #3638, opened 2023-11-18
prove cnplimc in iset.mm

### LINKED ISSUE #3634: Can cnplimc be proved in iset.mm?
As discussed in #3223 we defined `limCC` (and from there derivatives) by basically taking the classical epsilon-delta definition and changing not equal to apart. This has gotten us part of the way to making these concepts useful, including for example https://us.metamath.org/ileuni/limccnp2cntop.html and https://us.metamath.org/ileuni/dvid.html .

However, one of the key sticking points relates continuity to the limit of a function, namely that "A function is continuous at 𝐵 iff its limit at 𝐵 equals the value of the function there" that is https://us.metamath.org/mpeuni/cnplimc.html . The forward direction, proved at https://us.metamath.org/ileuni/cnplimcim.html , is easy, because continuity applies at all points in an interval and the limit of a function needs basically the same property at a subset of that interval (points apart from B).

In set.mm the reverse direction is proved using a lot of looking at decidable equality between B and other real numbers. Can this somehow be proved in iset.mm? On the one hand, you might think so because the limit of a function does tell us things about points arbitrarily close to B. On the other hand, it isn't clear how to replace the technique of combining B with non-B points to make an interval. I tried working with two points, B and another point apart from B but sufficiently close, and relatedly wondering if `cnlimc` is any easier than `cnplimc`, but I haven't figured out anything like that which works yet.

This has implications for theorems like https://us.metamath.org/mpeuni/dvcn.html where the set.mm proof uses the reverse direction of `cnplimc` in an apparently essential way.

jkingdon:
Fixes #3634 (see discussion there)

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10880,12 +10880,6 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
-  <td>cnlimc</td>
-  <td>~ cnlimcim</td>
-  <td>should be provable, see ~ cnplimccntop</td>
-</tr>
-
-<tr>
   <td>limccnp</td>
   <td>~ limccnpcntop</td>
 </tr>
@@ -11003,8 +10997,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 <tr>
   <td>dvcnp2</td>
   <td><i>none</i></td>
-  <td>The set.mm proof relies on limccnp2
-  and the converse of ~ cnplimcim</td>
+  <td>should now be provable</td>
 </tr>
 
 <tr>
@@ -11016,15 +11009,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 <tr>
   <td>dvaddbr</td>
   <td><i>none</i></td>
-  <td>the set.mm proof relies on restntr , undif1 , limcres , and
-  limccnp2</td>
+  <td>the set.mm proof relies on restntr , undif1 , and limcres</td>
 </tr>
 
 <tr>
   <td>dvmulbr</td>
   <td><i>none</i></td>
-  <td>the set.mm proof relies on restntr , undif1 , limcres , and
-  limccnp2</td>
+  <td>the set.mm proof relies on restntr , undif1 , and limcres</td>
 </tr>
 
 <tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10876,15 +10876,13 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <tr>
   <td>cnplimc</td>
-  <td>~ cnplimcim</td>
-  <td>The converse is conjectured to also be provable, but
-  would require a more involved proof.</td>
+  <td>~ cnplimccntop</td>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -140474,6 +140474,79 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
+        YBLVLAYLUUMUUOUUAYBVSYMUUMUUOUUAYBVTUUTUURYNYPWAYQYTUURYNYPWBUUPUUAYBYN
+        YPWCWMUURXTYOYASUURXTXSCXKPZYOUURXSCXKBUUTUVBWEUURXSRQCRQZUVFYOVEUURBRX
+        SUVEUUTVCAUVGYLUUMUUOUUAYBUULVLXSCXKUVCVPVQWFUUQYBWGWHWIWJWNWKWLWMWOAXL
+        BWPOQZXKRWPOQZUVAXQXRYKUPXCAUVIUVDUVHWQIXKBRWRWSUVIAWQWTKUANMXLXKCDXMFB
+        RXMWDZGXAXDXBACXPXNAEXMFUIAEFBXEPZXMHAUVIUVDUVKXMVEWQIXKXLFXMRBXLWDGUVJ
+        XFWSXGXHXIXJ $.
+    $}
+
+    $( A function is continuous at ` B ` iff its limit at ` B ` equals the $d x
+       A $. $d x B $. $d x F $. $d x K $. value of the function there.
```
avekens, line 140541: What is the meaning of the disjoint variables restrictions here?

jkingdon: Oh copy paste mishap, thanks for noticing.

Should be fixed now.

GinoGiotto:
> Specifically, use xmetres2 in place of xmetres .  This takes the size
in SHOW PROOF/SIZE from 3263 characters to 3125.

A little note: SHOW PROOF /SIZE is bugged https://github.com/metamath/metamath-exe/issues/137#issuecomment-1643995026. If you want to know the true size of a compressed proof I recommend using SHOW PROOF /COMPRESSED instead (at the end of the proof metamath-exe always shows the proof length). So in your case the size of `limccnp2cntop` was reduced from 2984 bytes to 2854 bytes.

jkingdon:
> > Specifically, use xmetres2 in place of xmetres .  This takes the size
> > in SHOW PROOF/SIZE from 3263 characters to 3125.
> 
> A little note: SHOW PROOF /SIZE is bugged [metamath/metamath-exe#137 (comment)](https://github.com/metamath/metamath-exe/issues/137#issuecomment-1643995026). If you want to know the true size of a compressed proof I recommend using SHOW PROOF /COMPRESSED instead (at the end of the proof metamath-exe always shows the proof length). So in your case the size of `limccnp2cntop` was reduced from 2984 bytes to 2854 bytes.

Ah yeah, thanks for the analysis. Most of the proof shortenings I do will end up shorter using any of our various ways of measuring it, but if I do post numbers next time I'll try to remember about SHOW PROOF /COMPRESSED.

avekens (approved):

PR #3613, opened 2023-11-05
Theorems for Godel-sets (4)

avekens:
Additional theorems for "Godel-sets of formulas"

Main results:
* The satisfaction predicate (for any Godel-Formula) is a function: ~ satfun
*  An element of the satisfaction predicate at a wff code is a valuation ` S `  of the variables so that the wff is true under the assignment ` S ` : ~satfvel
* Some examples showing the satisfaction predicate and valuations for Godel-Formulas of height 0: ~sategoelfvb and ~sategoelxmpl

In detail:
Changes in main set.mm:
* ~fiunlem, ~fiun added, proof of ~fun11iun shortened using ~fiunlem
* ~fviunfun added

Changes in MC's mathbox:
* some rearrangements in sections "Godel-sets of formulas"
* theorems ~satfun, ~satfvel, ~satfv0fvfmla0 added
* theorems for ` ( M SatE U ) ` added: ~satefv, ~satef, ~sate0fv0, ~satefvfmla0
* exaples for simple wff codes added: ~sategoelfvb, ~sategoelfv, ~sategoelelxmpl, ~sategoelxmpl
* theorem for ` M |= U ` added: ~prv

Issue #3599, opened 2023-10-29
New PART 17 MODEL THEORY


avekens:
In PR #3596, it was discussed to move section "Godel-sets of formulas" from MC's mathbox to the main body of set.mm.

I think a new "part" (as PART 17, following PART 16 GRAPH THEORY) would be required for it, which could be titled "MODEL THEORY" (according to @digama0's comment in PR #3572). Its (currently one and only) section would be "Gödel-sets of formulas", containing the subsections "Primary Gödel-sets", "The Satisfaction predicate", "Gödel formulas", "The "proves" relation", "Additional Gödel-sets".

Additional candidates for this new PART would be the following sections in MC's mathbox:

*    20.6.12 Models of ZF
*    20.6.13 Metamath formal systems
*    20.6.14 Grammatical formal systems
*    20.6.15 Models of formal systems

I also detected section "Provability logic" in BJ's mathbox, where a "provability predicate" is defined and theorems for Gödel's second incompleteness theorem ( ~ bj-babygodel ) and Löb's theorem ( ~ bj-babylob ) are provided. Would this also fit into this new PART?

Side remark:
* In comments, Gödel should be written as `G&ouml;del`, and as `goedel` in labels.

digama0:
Sounds good to me. I would shorten `goedel` the label segment to just `goe` although I don't think we have any examples of such other than `bj-babygodel`?

Regarding the usage of the umlaut, feel free to put it everywhere you see Godel, I was just being lazy when I wrote those sections.

benjub:
Maybe you could do a first PR with only the first subsection, "Gödel codes of formulas" ?  (I think "Gödel code" is more explanatory and less confusing than "Gödel set", as I wrote in https://github.com/metamath/set.mm/pull/3572#issuecomment-1763494610). Maybe that subsection could contain both the primary definitions of Gödel codes, and also the definitions for derived formulas.

Maybe `gc` for "Gödel code" could be used in labels ?

As for bj-babygodel, bj-babylob, these can stay in my mathbox for now. They are really the tiny final part of the proof, once everything has been established, whereas the part you are working on about Gödel codes ("arithmetization of logic") is really about building the infrastructure for that proof (among other applications).

digama0:
[...]

PR #3640, opened 2023-11-18
Theorems for Godel-sets (5)

avekens:
The main result of this PR is ~elnanelprv, giving an example of how the universal validity of a theorem in set.mm (here ~elnanel) can be shown for its corresponding Godel formula:

`|- ( ( M e. V /\ A e. _om /\ B e. _om ) -> M |= ( ( A e.g B ) |g ( B e.g A ) ) )`

Details:

Changes in main set.mm:
* ~bj-rabbida (as ~rabbida), ~bj-rabbid (as ~rabbid) ~inex2g (as ~inex2g) and ~elmapresaun moved from mathboxes
* ~2ex2rexrot, ~rexopabb, ~elnanel added
* ~fun11iun renamed ~f1iun

Changes in MC's mathbox:
* review remarks in PR #3613
* ~satfv1 and lemmas added
* ~satfv1fvfmla1, ~2goelgoanfmla1, ~satefvfmla1 added
* ~elnanelprv added

Changes in GS's mathbox:
* ~rabbida removesd (dublicate of bj-rabbida which is older


set.mm
```diff
@@ -91095,7 +91095,7 @@ is that it denies the existence of a set containing itself ( ~ elirrv ).
     ( cvv wcel wa wn cep wfr zfregfr efrn2lp mpan elex anim12i con3i pm2.61i )
     ACDZBCDZEZABDZBADZEZFZCGHRUBCICABJKUARSPTQABLBALMNO $.
 
-  $( A class is not an element of another class and vice versa.  (Contributed
+  $( Two classes are not elements of each other simultaneously.  (Contributed
      by AV, 5-Nov-2023.) $)
   elnanel $p |- ( A e. B -/\ B e. A ) $=
```
benjub, line 91100: I think that the alternative denial operator should be very seldom used outside of the NAND section. Here, elnanel is simply a rewriting of en2lp using a less standard and lesser known connective. There are many (many) theorems where NAND, NOR and XOR could be used to save a few symbols, but this would be at the cost of increasing the number of statements and the knowledge required from readers.

Maybe it is justified here since it serves as a basic example for Godel codes. If that's the case and we want to keep it, then we should probably add a comment to that effect and discourage its usage.

set.mm
```diff
@@ -47129,6 +47160,11 @@ holding in an empty domain (see Axiom A5 and Rule R2 of [LeBlanc]
       ACOAIPQGOABJKOBDLMN $.
   $}
 
+  $( Sufficient condition for an intersection relation to be a set, reversed
+     form of ~ inex1g .  (Contributed by Peter Mazsa, 19-Dec-2018.) $)
```
benjub, line 47164:
```suggestion
  $( Sufficient condition for an intersection to be a set.  Commuted form of
     ~ inex1g .  (Contributed by Peter Mazsa, 19-Dec-2018.) $)
```

set.mm
```diff
@@ -507309,6 +507630,23 @@ codes is an increasing chain (with respect to inclusion).  (Contributed
       NUIUABJKUGUHOPQFERT $.
   $}
 
+  ${
+    $d A a $.  $d B a $.  $d M a $.  $d V a $.
+    $( The wff ` ( A e. B -/\ B e. A ) ` encoded as ` ( ( A e.g B ) `
+       ` |g ( B e.g A ) ) ` is true in any model ` M ` , i.e., if for every
```
icecream17, line 507636:
```suggestion
       ` |g ( B e.g A ) ) ` is true in any model ` M ` , i.e., for every
```

icecream17 (approved):
approve given changes

set.mm
avekens: The i.e. refers to "in any model", and this is equivalent with "if for every ...". Therefore, the "if" is correct here (or may even be replaced by "iff", see comment for ~prv).

set.mm
avekens: done

set.mm
benjub: I would remove the "i.e." clause altogether, since the comment is sufficiently clear without it.

tirix (approved):

avekens:
Besides the revision according to BJ's review remark, I added 5 new theorems, and want to ask for an additional review/approval. Sorry for that, but I think I will stop my activities regarding Godel sets and model theory for a while after this PR.

set.mm
avekens: Done.

set.mm
```diff
@@ -507303,20 +507522,171 @@ codes is an increasing chain (with respect to inclusion).  (Contributed
...
+  $}
+
+  ${
+    ex-sategoelel12.s $e |- S = ( x e. _om |-> if ( x = 2o , 1o , 2o ) ) $.
+    $( Example of a valuation of a simplified satisfaction predicate over a
+       proper pair (of ordinal numbers) as model for a Godel-set of membership
+       using the properties of a successor:
+       ` ( S `` 2o ) = 1o e. 2o = ( S `` 2o ) ` .  Remark: the indices ` 1o `
+       and ` 2o ` are intentionally reversed to distinguish them from elements
+       of the model: ` ( 2o e.g 1o ) ` should not be confuse with
```
icecream17, line 507671:
```suggestion
       of the model: ` ( 2o e.g 1o ) ` should not be confused with
```
in retrospect these typo comments are unimportant and nonblocking; in fact I may do a pr soon regarding df-ich so such things may be fixed there

icecream17 (approved):

set.mm
avekens: Typos fixed.

benjub (approved):

set.mm
```diff
@@ -505661,6 +505729,150 @@ required as (atomic) formulas, can be introduced as a defined notion in
...
+                          /\ F e. ( M ^m V ) /\ Z e. M )
+      -> ( ( { <. N , Z >. } u. ( F |` ( V \ { N } ) ) ) ` J ) = ( F ` J ) ) $=
+    ( wcel wceq wn w3a cmap co cop csn cfv wfn cin c0 wa sylibr cdif cres simp1
+    cun anim1i 3adant2 fnsng syl elmapfn 3ad2ant2 difssd neldifsnd incom eqeq1i
+    fnssresd disjsn bitri neqne anim2i 3adant1 3ad2ant1 eldifsn fvun2 syl112anc
+    wne fvresd eqtrd ) DEGZBEGZBDHIZJZACEKLGZFCGZJZBDFMNZAEDNZUAZUBZUDOZBVROZBA
+    OVNVOVPPZVRVQPVPVQQZRHZBVQGZVSVTHVNVHVMSZWAVKVMWEVLVKVHVMVHVIVJUCUEUFDFECUG
+    UHVNEVQAVLVKAEPVMACEUIUJVNEVPUKUOVNDVQGIZWCVNDEULWCVQVPQZRHWFWBWGRVPVQUMUNV
+    QDUPUQTVNVIBDVEZSZWDVKVLWIVMVIVJWIVHVJWHVIBDURUSUTVABEDVBTZVPVQVOVRBVCVDVNB
+    VQAWJVFVG $.
```
tirix, line 505753: It looks like we already have `fvsnun` and `fvsnun2` which are very similar and use less hypotheses. I'm wondering if they could have been used here instead.

Also, as a note, elsewhere, modifying a function by changing its value at one point is formalized using `if`, like in `cantnfp1.f`, `symextf.e`, `df-limc`, `fourierdlem32.y`, `fdmdifeqresdif`. Should we recommend one vs. the other? (I believe the `if` works with proper classes, the function restriction not)

set.mm
```diff
@@ -507326,6 +507696,61 @@ codes is an increasing chain (with respect to inclusion).  (Contributed
...
+       AV, 19-Nov-2023.) $)
+    prv0 $p |- ( U e. ( Fmla ` _om ) -> (/) |= U ) $=
+      ( vx com cfmla cfv wcel c0 cprv wbr csate co wceq csat sate0 cv wn peano1
+      wa cvv 0ex wal wf n0ii intnan a1i f00 sylnibr pm3.2i satfvel mp3an1 mtand
+      alrimiv eq0 sylibr eqtrd cmap prv mpan wne ne0ii map0b mp1i eqeq2d mpbird
+      wb bitrd ) ACDEZFZGAHIZGAJKZGLZVHVJACGGMKEEZGAVGNVHBOZVLFZPZBUAVLGLVHVOBV
+      HVNCGVMUBZVHVMGLZCGLZRZVPVSPVHVRVQGCQUCUDUECVMUFUGGSFZVTRVHVNVPVTVTTTUHVM
+      AGGSSUIUJUKULBVLUMUNUOVHVIVJGCUPKZLZVKVTVHVIWBVETAGSVGUQURVHWAGVJCGUSWAGL
+      VHGCQUTCVAVBVCVFVD $.
+  $}
```
tirix, line 507726: The set `U` does not seem to be special, so I think the comment should rather be: _every wff is true in an empty model_?

set.mm
```diff
@@ -507326,6 +507696,61 @@ codes is an increasing chain (with respect to inclusion).  (Contributed
...
+
+  ${
+    $d I a $.  $d J a $.  $d V a $.  $d X a $.
+    $( No wff encoded as a Godel-set of membership is true in a model with only
+       one element.  (Contributed by AV, 19-Nov-2023.) $)
+    prv1n $p |- ( ( I e. _om /\ J e. _om /\ X e. V )
+                  -> -. { X } |= ( I e.g J ) ) $=
```
tirix, line 507733: Is my understanding of this theorem correct?
> In `set.mm`, because of `eirrv`, which is derived from the axiom of regularity, one cannot have `A e. A`.
> And because our model/satisfiability theory `Sat` and `|=` is based on `set.mm` ZFC classes, this also cannot hold in any of our models.

But then, what is the use of `AxReg`, defined in `~ df-gzreg`?

The same is obviously valid for all axioms, I used `ax-reg` here because this is the one linked with this theorem.

tirix (approved):
Another great addition about model theory, too bad you cannot continue!

Nothing blocking in my comments, especially as this is in a mathbox: I have merely a few open questions.

Sorry it took me a bit long to review.
If needed I can make up by helping with the merge.

set.mm
benjub: The formula `E. x T.` is certainly not true in the empty model, so either there is a contradiction in set.mm, or, more probably, the comments do not describe the associated proofs/definitions correctly.

benjub: Ok, so after a quick look at df-prv, we see that it is not adapted to empty models. So this limitation should be indicated there, and after that, probably no use of the phrase "empty model" should be made. (The statement can stay, but of course it doesn't mean anything for the "empty model".) A comment could be:

Since `|=` is defined in terms of the interpretations making the given formula true, it is not defined on the empty model, since there, there are no interpretations. In particular, the empty set on the LHS `|=` should not be interpreted as the empty model.  Statement `~ prv0` shows that our definition  yields `(/) |= U` for all formulas, though of course the formula `E. x x = x` is not satisfied on the empty model.

set.mm
benjub: I guess proving theorems like `prv1n` will be pretty hard (not mathematically, but in terms of tracking axiom dependencies) with current tools, because you want to avoid to use all ZF axioms "on the syntactic side", but all intermediate theorems we used to build the machinery of model theory drag with them many axiom dependencies... (but maybe not the axiom of regularity if we were careful?).

Probably, the theorem we want to prove is something like
```
 prv1n $p |- ( ( I e. _om /\ J e. _om /\ X e. V ) ->
       ( -. { X } |= AxReg -> -. { X } |= ( I e.g J ) ) ) $=
```
while avoiding the "genuine" uses of ax-reg

set.mm
avekens: > It looks like we already have fvsnun and fvsnun2 which are very similar and use less hypotheses. I'm wondering if they could have been used here instead.

You are right, fvsnun and fvsnun2 can be used directly in satfv1lem, so the lemmata satfv1lemlem1 and satfv1lemlem2 are not required/useful and can be removed. I will do this in my next commit.

avekens: > Also, as a note, elsewhere, modifying a function by changing its value at one point is formalized using `if`, like in `cantnfp1.f`, `symextf.e`, `df-limc`, `fourierdlem32.y`, `fdmdifeqresdif`. Should we recommend one vs. the other? (I believe the `if` works with proper classes, the function restriction not)

This comes from the definition df-sat:
``Sat = ( m e. _V , e e. _V |->...  y = { a e. ( m ^m _om ) | A. z e. m ( { <. i , z >. } u.  ( a |` ( _om \ { i } ) ) ) e. ( 2nd ` u ) } ...``

Since we have no function values here, I think using an `if` expression would make it more complicated.

set.mm
avekens: I will enhance the comments of df-prv and prv0 accordingly.

set.mm
avekens: > But then, what is the use of `AxReg`, defined in `~ df-gzreg`?
> 
> The same is obviously valid for all axioms, I used `ax-reg` here because this is the one linked with this theorem.

That is a really good question. I also have no idea how the axioms could/would be integrated into the framework of the current definitions. This would be a task to be done in the context of issue #3599.  And this should be done before all the stuff is moved to main.

benjub: I don't understand: I can see my answer from yesterday above, but maybe it's because I'm logged in, or there is a delay.  I'm copying it below and hopefully it will work this time:

[
I guess proving theorems like prv1n will be pretty hard (not mathematically, but in terms of tracking axiom dependencies) with current tools, because you want to avoid to use all ZF axioms "on the syntactic side", but all intermediate theorems we used to build the machinery of model theory drag with them many axiom dependencies... (but maybe not the axiom of regularity if we were careful?).

Probably, the theorem we want to prove is something like
```
 prv1n $p |- ( ( I e. _om /\ J e. _om /\ X e. V ) ->
       ( -. { X } |= AxReg -> -. { X } |= ( I e.g J ) ) ) $=
```
while avoiding the "genuine" uses of ax-reg
]

What I called "theorems we used to build the machinery of model theory" is what Mario calls the metatheory.

This does not prevent merging of this MR, since this is a longer term work.

benjub (approved):

set.mm
tirix: > This does not prevent merging of this MR, since this is a longer term work.

Yes, I completely agree. 
My point was more about trying to understand what we are proving here.

tirix (approved):

avekens:
I think the discussion should be continued in issue #3599 and could result in separate PRs. therefore, I will merge this PR now.

PR #3644, opened 2023-11-19
shorten pm2.61i

wlammen:
1. shorten pm2.61i
2. rename theorems in my mathbox according to proposed changes
3. reorder theorems
    Recent proof changes allow moving theorems closer to each other that form a group.  In particular
    - move nsyl2 next to nsyl
    - move con1, con1i next to con1d
    - ja does not interrupt the sequence of pm2.61* theorems any longer.


set.mm
```diff
@@ -2175,6 +2160,7 @@ In classical logic (our logic) this is always true.  In intuitionistic
       ( wn pm2.61d2 pm2.61i ) ACEAGBCDFHI $.
   $}
 
+
```
benjub, line 2163:
```suggestion
```

benjub (approved):

benjub (approved):

icecream17 (approved):

PR #3652, opened 2023-11-22
Propose renaming sseqtr4i and sseqtr4d

jkingdon:
The pattern is to use the 3/4 naming convention if both hypotheses are symmetric, but "r" if only one hypothesis is symmetric.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -80,6 +80,8 @@ proposed  syl6eleq  eleqtrdi    compare to eleqtri or eleqtrd
 proposed  syl6eleqr eleqtrrdi   compare to eleqtrri or eleqtrrd
 proposed  syl6ss    sstrdi      compare to sstri or sstrd
 proposed  syl6sseq  sseqtrdi    compare to sseqtri or sseqtrd
+proposed  sseqtr4i  sseqtrri
+proposed  sseqtr4d  sseqtrrd
 proposed  syl6sseqr sseqtrrdi
 proposed  syl6eqss  eqsstrdi    compare to eqsstri or eqsstrd
 proposed  syl6eqssr eqsstrrdi


wlammen (approved):

icecream17 (approved):

icecream17:
Well, some of the proposed renames above use "com", like syl6rbb --> bitrdicom. This is mostly unrelated to replacing "syl5/syl6" but I wonder what the difference between "com" and "r" is. The conventions mention "r" ~~ right and "com" ~~ commutativity.

icecream17:
One explanation is that "r" applies to hypotheses and "com" to conclusions, but after further thought this is also represented with the (1)/2/3/4 system:
```
(proposed is same unless noted)

bitrdi (currently: syl6bb)
ph 1 2, 2 3 => 1 3 <--
1 2, ph 3 1 => 3 2

bitr2di (currently: syl6rbb) (proposed: bitrdicom)
ph 1 2, 2 3 => 3 1 <--
1 2, ph 3 1 => 2 3

bitr3di (currently: syl5rbbr) (proposed: bitr3idcom)
ph 1 2, 1 3 => 2 3 <--
1 2, ph 1 3 => 3 2

bitr3id (currently: syl5bbr)
ph 1 2, 1 3 => 3 2
1 2, ph 1 3 => 2 3 <--

bitr4di (currently: syl6bbr)
ph 1 2, 3 2 => 1 3 <--
1 2, ph 3 2 => 3 1

bitr4id (currently: syl6rbbr) (proposed: bitr4dicom)
ph 1 2, 3 2 => 3 1
1 2, ph 3 2 => 1 3 <--

bitr2id (currently: syl5rbb) (proposed: bitridcom)
ph 1 2, 3 1 => 2 3
1 2, ph 2 3 => 3 1 <--

bitrid (currently: syl5bb)
ph 1 2, 3 1 => 3 2
1 2, ph 2 3 => 2 3 <--
```

And on further thought some of these r vs com issues would probably be in another pr. (The corrections to syl*bb* don't affect this pr, jkingdon's original reasoning above still holds)

jkingdon:
> One explanation is that "r" applies to hypotheses and "com" to conclusions

That has been my thinking, yes.

> but after further thought this is also represented with the (1)/2/3/4 system:

Hmm, interesting. I especially found examples like `bitr2i` and `entr2i` which would have been called `bitricom` and `entricom` based on the (over?)use of `com`. But maybe we can/should find ways to avoid the `com` names.

> bitrdi (currently: syl6bb)

Agreed, looks good as proposed in changes-set.txt.

> bitr2di (currently: syl6rbb) (proposed: bitrdicom)

Yes, `bitr2di` does appear to be a better proposal, and we could say `compare to bitr2i or bitr2d`

> bitr3di (currently: syl5rbbr) (proposed: bitr3idcom)

So then would you change the order of the two hypotheses?

I notice that there were a bunch of changes from Feb-2013 described with "reordered hypotheses for better logical flow" which I guess I take mostly as precedent that we can consider hypothesis order as well as names if we'd like.

> bitr3id (currently: syl5bbr)

Agreed, looks good as proposed in changes-set.txt.

> bitr4di (currently: syl6bbr)

Agreed, looks good as proposed in changes-set.txt.

> bitr4id (currently: syl6rbbr) (proposed: bitr4dicom)

bitr4id would work if the order of the hypotheses is changed.

> bitr2id (currently: syl5rbb) (proposed: bitridcom)

Yes, `bitr2id` does seem to be a better proposal and could be described with `compare to bitr2i or bitr2d`

> bitrid (currently: syl5bb)

Agreed, looks good as proposed in changes-set.txt.

> And on further thought some of these r vs com issues would probably be in another pr. (The corrections to syl_bb_ don't affect this pr, jkingdon's original reasoning above still holds)

Yeah, it seems like we could merge this one (which just has two proposed renames) and make another pull request for the follow up. Did you want to make the follow up pull request or should I?

icecream17:
I went ahead and made the follow up pr

PR #3655, opened 2023-11-24
Revise eq/bitr*id/di proposal with (1)/2/3/4 pattern

icecream17:
Completes a tangent in #3652

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
 proposed  syl5ibr   imbitrrid
 proposed  syl5ibrcom imbitrridcom
 proposed  syl5bb    bitrid      compare to bitri or bitrd
-proposed  syl5rbb   bitr2id
+proposed  syl5rbb   bitr2id     compare to bitr2i or bitr2d
 proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
-proposed  syl5rbbr  bitr3di
+proposed  syl5rbbr  bitr3di     hypotheses would be switched
 proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
-proposed  syl5req   eqtr2id
+proposed  syl5req   eqtr2id     compare to eqtr2i or eqtr2d
 proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
-proposed  syl5reqr  eqtr3di
+proposed  syl5reqr  eqtr3di     hypotheses would be switched
 proposed  syl5eqel  eqeltrid    compare to eqeltri or eqeltrd
 proposed  syl5eqelr eqeltrrid   compare to eqeltrri or eqeltrrd
 proposed  syl5eleq  eleqtrid    compare to eleqtri or eleqtrd
@@ -67,13 +67,13 @@ proposed  syl6ibr   imbitrrdi
 proposed  syl6bi    biimtrdi
 proposed  syl6bir   biimtrrdi
 proposed  syl6bb    bitrdi      compare to bitri or bitrd
-proposed  syl6rbb   bitr2di
+proposed  syl6rbb   bitr2di     compare to bitr2i or bitr2d
 proposed  syl6bbr   bitr4di     compare to bitr4i or bitr4d
-proposed  syl6rbbr  bitr4id
+proposed  syl6rbbr  bitr4id     hypotheses would be switched
 proposed  syl6eq    eqtrdi      compare to eqtri or eqtrd
-proposed  syl6req   eqtr2di
+proposed  syl6req   eqtr2di     compare to eqtr2i or eqtr2d
 proposed  syl6eqr   eqtr4di     compare to eqtr4i or eqtr4d
-proposed  syl6reqr  eqtr4id
+proposed  syl6reqr  eqtr4id     hypotheses would be switched
 proposed  syl6eqel  eqeltrdi    compare to eqeltri or eqeltrd
 proposed  syl6eqelr eqeltrrdi   compare to eqeltrri or eqeltrrd
 proposed  syl6eleq  eleqtrdi    compare to eleqtri or eleqtrd
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
 proposed  syl5ibr   imbitrrid
 proposed  syl5ibrcom imbitrridcom
 proposed  syl5bb    bitrid      compare to bitri or bitrd
-proposed  syl5rbb   bitridcom
+proposed  syl5rbb   bitr2id
 proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
-proposed  syl5rbbr  bitr3idcom
+proposed  syl5rbbr  bitr3di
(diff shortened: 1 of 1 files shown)


changes-set.txt
```diff
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
 proposed  syl5ibr   imbitrrid
 proposed  syl5ibrcom imbitrridcom
 proposed  syl5bb    bitrid      compare to bitri or bitrd
-proposed  syl5rbb   bitridcom
+proposed  syl5rbb   bitr2id
 proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
-proposed  syl5rbbr  bitr3idcom
+proposed  syl5rbbr  bitr3di
```
jkingdon, line 48: And would you change the order of the hypotheses? If so we could add "change order of hypotheses from syl5rbbr" to the notes column.

changes-set.txt
```diff
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
 proposed  syl5ibr   imbitrrid
 proposed  syl5ibrcom imbitrridcom
 proposed  syl5bb    bitrid      compare to bitri or bitrd
-proposed  syl5rbb   bitridcom
+proposed  syl5rbb   bitr2id
```
jkingdon, line 46: Is it OK to add `compare to bitr2i or bitr2d` to the notes column? I find those notes helpful when I'm preparing pull requests to actually rename some of these.

changes-set.txt
```diff
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
...
 proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
-proposed  syl5rbbr  bitr3idcom
+proposed  syl5rbbr  bitr3di
 proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
-proposed  syl5req   eqtridcom
+proposed  syl5req   eqtr2id
```
jkingdon, line 50: This can also be described with `compare to bitr2i or bitr2d`

changes-set.txt
```diff
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
...
 proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
-proposed  syl5reqr  eqtr3idcom
+proposed  syl5reqr  eqtr3di
```
jkingdon, line 52: This is another one with the issue of the order of the hypotheses.

changes-set.txt
```diff
@@ -67,13 +67,13 @@ proposed  syl6ibr   imbitrrdi
 proposed  syl6bi    biimtrdi
 proposed  syl6bir   biimtrrdi
 proposed  syl6bb    bitrdi      compare to bitri or bitrd
-proposed  syl6rbb   bitrdicom
+proposed  syl6rbb   bitr2di
```
jkingdon, line 70: could be described with `compare with bitr2i or bitr2d`

changes-set.txt
```diff
@@ -67,13 +67,13 @@ proposed  syl6ibr   imbitrrdi
...
 proposed  syl6bbr   bitr4di     compare to bitr4i or bitr4d
-proposed  syl6rbbr  bitr4dicom
+proposed  syl6rbbr  bitr4id
```
jkingdon, line 72: this one has the issue with the order of the hypotheses

changes-set.txt
```diff
@@ -67,13 +67,13 @@ proposed  syl6ibr   imbitrrdi
...
 proposed  syl6eq    eqtrdi      compare to eqtri or eqtrd
-proposed  syl6req   eqtrdicom
+proposed  syl6req   eqtr2di
```
jkingdon, line 74: could be described with `compare to eqtr2i or eqtr2d`

changes-set.txt
```diff
@@ -67,13 +67,13 @@ proposed  syl6ibr   imbitrrdi
...
 proposed  syl6eqr   eqtr4di     compare to eqtr4i or eqtr4d
-proposed  syl6reqr  eqtr4dicom
+proposed  syl6reqr  eqtr4id
```
jkingdon, line 76: this has the issue with the order of the hypotheses

jkingdon:
I realize I commented on a bunch of lines here but there are only two issues, (1) on roughly half the changes, whether to describe them a bit more verbosely in changes-set.txt, and (2) on the other half, whether we want to change the order of the hypotheses so that the name fits the patterns better.

jkingdon (approved):

avekens:
I did not follow this sequence of changes in detail, so I want to ask where the "(1)/2/3/4 pattern" is described/documented.

icecream17:
If I recall correctly, it isn't documented anywhere, however the pattern is based on these examples:
eqtri, eqtr2i, eqtr3i, eqtr4i
bitri, bitr2i, bitr3i, bitr4i

```
1. 01 12 => 02
2. 21 10 => 02
3. 10 12 => 02
4. 01 21 => 02
```

changes-set.txt
```diff
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
...
+proposed  syl5rbb   bitr2id     compare to bitr2i or bitr2d
 proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
-proposed  syl5rbbr  bitr3idcom
+proposed  syl5rbbr  bitr3di     hypotheses would be switched
 proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
-proposed  syl5req   eqtridcom
+proposed  syl5req   eqtr2id     compare to eqtr2i or eqtr2d
 proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
-proposed  syl5reqr  eqtr3idcom
+proposed  syl5reqr  eqtr3di     hypotheses would be switched
```
avekens, line 52: Currently, syl5reqr follows the pattern 12 10 02, which does not have a number yet (No. 5 -> eqtr5id?). What does "hypotheses would be switched" mean? Switching their order would result in exchanging the "i" and "d" at the end ("id" means first hypothesis without antecedent, and second hypothesis with antecedent, doesn't it?).

icecream17: The statement "hypotheses would be switched" means that compared to syl5reqr, the hypotheses would be in a different order:

| image | image |

The name eqtr3di is mostly unrelated to syl5reqr 

syl5reqr: (syl5 --> the B in the second hypothesis gets switched out, the two "r" --> the first hypothesis and the conclusion are commuted)

eqtr3di: (eqtr3 --> the left symbols are the same, di --> the first hypothesis has an antecedent, the second doesn't)

The hypotheses are switched because the alternate suggestion here is `eqtr3idcom`, so commuting gives more consistency.

avekens (approved):

avekens:
Maybe the patterns (as shown above) should be added before the proposed renamings (to justify them).

jkingdon:
> Maybe the patterns (as shown above) should be added before the proposed renamings (to justify them).

If anyone wants to try to come up with a way of formatting these which would be more clarifying than confusing, see what you can do.

(personally, I'm not sure that whatever additional text we add would be clearer than the references to existing theorems, but perhaps that's just limited imagination on my part).

PR #3657, opened 2023-11-25
Relax the third hypothesis of ~ liminfgelimsupuz to support sequences of extended reals

glacode:


wlammen (approved):
This is a mathbox change, so no objections to the increase of proof length in total

tirix (approved):
That's the example of your Yamma 11 post!  ;-)

PR #3658, opened 2023-11-27
The sum rule for derivatives (iset.mm)

jkingdon:
The thing which makes the intuitionizing go smoothly here is to realize that while set.mm allows X and Y to be different in theorems like https://us.metamath.org/mpeuni/dvaddbr.html , it isn't using that generality if you follow a few iterations of "This theorem is referenced by". Furthermore, the set.mm proof of dvaddbr is using excluded middle to cover exactly that point.

Includes a few new/revised theorems for the `oF` syntax.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11037,8 +11037,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <tr>
   <td>dvaddf</td>
-  <td><i>none</i></td>
-  <td>the set.mm proof relies on dvaddbr and dvadd</td>
+  <td>~ dviaddf</td>
 </tr>
 
 <tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3142,8 +3142,15 @@ that we are evaluating functions within their domains.
 <TD>~ ofvalg </TD>
 </TR>
 
+<tr>
+  <td>offn</td>
+  <td>~ off</td>
+  <td>the set.mm proof of offn uses ovex and it isn't clear whether
+  anything can be proved with weaker hypotheses than ~ off</td>
+</tr>
+
 <TR>
-<TD>offn , offveq , caofid0l , caofid0r , caofid1 , caofid2</TD>
+<TD>offveq , caofid0l , caofid0r , caofid1 , caofid2</TD>
 <TD><I>none</I></TD>
 <TD>Assuming we really need to add conditions that the operations
 are functions being evaluated within their domains, there would be a
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -3149,8 +3149,13 @@ that we are evaluating functions within their domains.
   anything can be proved with weaker hypotheses than ~ off</td>
 </tr>
 
+<tr>
+  <td>offveq</td>
+  <td>~ offeq</td>
+</tr>
+
 <TR>
-<TD>offveq , caofid0l , caofid0r , caofid1 , caofid2</TD>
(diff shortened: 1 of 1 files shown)


digama0:
I would have assumed the general version of dvaddbr follows from the same-domain version because ``( F oF + G ) = ( ( F |` ( X i^i Y ) ) oF + ( G |` ( X i^i Y ) ) )``. Or is this not true in iset.mm for some reason?

jkingdon:
> I would have assumed the general version of dvaddbr follows from the same-domain version because `` ( F oF + G ) = ( ( F |` ( X i^i Y ) ) oF + ( G |` ( X i^i Y ) ) ) ``. Or is this not true in iset.mm for some reason?

Not using the set.mm proof, which uses `undif1` in the form of `((𝑋 ∖ {𝐶}) ∪ {𝐶}) = (𝑋 ∪ {𝐶})` in https://us.metamath.org/mpeuni/dvaddbr.html where it is doing the restriction to `X i^i Y` that you describe (see for example steps 85, 92, and 127). The proof also uses https://us.metamath.org/mpeuni/restntr.html which looks non-intuitionistic to me (just to eyeball it, not that I've derived a taboo or anything). We also haven't intuitionized https://us.metamath.org/mpeuni/limcres.html although I've been operating under the assumption that is possible.

Now, there might be a whole different approach, but none of what I describe in the previous paragraph seemed like something which could be patched with small tweaks.

digama0:
I am suspecting limcres as the source of the uncomfortable derivation. The assumption limcres.i looks pretty useless / over-strong in an intuitionistic setting. It is being used here to argue that $F\restriction(X\cap Y)\setminus \\{C\\}$ and $F\restriction X\setminus \\{C\\}$ have the same limit points, which is surely true intuitionistically (with the $\ne$ replaced by $\\#$ as appropriate).

I think the statement can be fixed simply by assuming the input sets already have the limit point, instead of including them in the theorem statement. Here's a variation on the statement of limcres that avoids the union, although possibly not all of these assumptions are needed for the proof:

```
$e ⊢ (𝜑 → F : A ⟶ ℂ)
$e ⊢ (𝜑 → C ⊆ A)
$e ⊢ (𝜑 → A ⊆ ℂ)
$e ⊢ (𝜑 → C' ⊆ A')
$e ⊢ (𝜑 → A' ⊆ ℂ)
$e ⊢ (𝜑 → B e. C')
$e ⊢ (𝜑 → A \\ {B} ⊆ A')
$e ⊢ (𝜑 → C \\ {B} ⊆ C')
$e ⊢ K = (TopOpen‘ℂfld)
$e ⊢ J = (K ↾t A')
$e ⊢ (𝜑 → B ∈ ((int‘J) ‘ C'))
$p ⊢ (𝜑 → ((F ↾ C) limℂ B) = (F limℂ B))
```
Here `A \\ {B}` is the set `{ x e. A | x # B }` (which I'm surprised you haven't defined already).

jkingdon:
> I am suspecting limcres as the source of the uncomfortable derivation.

I have little intuition on this topic, but that seems plausible, I've moved this to #3661 since it seems like it would need some working out and we don't seem to need it just yet.

> Here `A \\ {B}` is the set `{ x e. A | x # B }` (which I'm surprised you haven't defined already).

New notations are often a double edged sword but I have thought a bit harder about things like whether to make a theorem `z e. { x e. A | x # B } -> z # B ` which saves three steps (`breq1`, `elrab`, and `simprbi`). Or other such theorems if they are out there.

tirix (approved):

PR #3663, opened 2023-11-28
Proofs about Permutation Cycles

tirix:
This includes several proofs which were pending on my side, but no major new result.

One interesting result might be `~symgcntz`, which states that any set of permutations with disjoint orbits commute, and the corollary for permutation cycles, `~tocyccntz`.

A few theorems were moved from other mathboxes to the main part.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -93,6 +93,11 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+28-Nov-23 unitresl  [same]      moved from GM's mathbox to main set.mm
+28-Nov-23 unitresr  [same]      moved from GM's mathbox to main set.mm
+28-Nov-23 nel02     [same]      moved from PM's mathbox to main set.mm
+28-Nov-23 fvco3d    [same]      moved from SP's mathbox to main set.mm
+28-Nov-23 fzossz    [same]      moved from GS's mathbox to main set.mm
 26-Nov-23 syl5sseq  sseqtrid    compare to sseqtri or sseqtrd
 21-Nov-23 syl5sseqr sseqtrrid
 21-Nov-23 eqsstr3d  eqsstrrd
--- a/discouraged
+++ b/discouraged
@@ -418,6 +418,7 @@
 "4syl" is used by "swrdccat2".
 "4syl" is used by "symgtrinv".
 "4syl" is used by "tmsxms".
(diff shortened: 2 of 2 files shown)


savask (approved):
Nice work! Always good seeing more group theory in set.mm.

set.mm
```diff
@@ -457631,6 +457714,36 @@ Class abstractions (a.k.a. class builders)
...
+    $( Rewrite an empty intersection with a pair.  (Contributed by Thierry
+       Arnoux, 20-Nov-2023.) $)
+    inpr0 $p |- ( ( A i^i { B , C } ) = (/) <-> ( -. B e. A /\ -. C e. A ) ) $=
+      ( vx cv wne wa wral cpr cin c0 wceq wcel wn r19.26 wi wal wb 3bitr4i nelb
+      cvv nelpr elv imbi2i albii disj1 df-ral anbi12i ) DEZBFZUICFZGZDAHZUJDAHZ
+      UKDAHZGABCIZJKLZBAMNZCAMNZGUJUKDAOUIAMZUIUPMNZPZDQUTULPZDQUQUMVBVCDVAULUT
+      VAULRDUIBCUAUBUCUDUEDAUPUFULDAUGSURUNUSUODBATDCATUHS $.
+  $}
+
+  $( The class ` A ` is not in ` ( C \ { A , B } ) ` .  (Contributed by Thierry
```
avekens, line 457735: This is just the statement slightly rewritten. Maybe "An element of a pair is not an element of a difference with this pair."

set.mm
```diff
@@ -524408,6 +524956,15 @@ A Fne if ( S = (/) , { X } , U. S ) ) $=
   mofal $p |- E* x F. $=
     ( wfal wex wmo nexfal exmo mtpor ) BACBADAEBAFG $.
 
+  ${
+    rmonot.1 $e |- ( x e. A -> -. ph ) $.
+    $( "At most one" restricted existential quantifier for a statement never
+       true.  (Contributed by Thierry Arnoux, 27-Nov-2023.) $)
```
avekens, line 524962: "... which is never true."

set.mm
tirix: Yes! I was not very inspired.
I'll write "_The first element of a pair is not an element of a difference with this pair._" and similarly for the second such theorem.

avekens (approved):

PR #3664, opened 2023-11-30
Dropped ax-ext and ax-9 from reu6

GinoGiotto:
This automatically drops ax-ext and ax-9 from ~reu3 ~reu7 ~reu8 ~2reuimp0 as well.

Also the new proof is 12 bytes shorter than the original.


icecream17 (approved):

tirix (approved):

avekens (approved):

wlammen (commented):
I think an OLD version should have been created, according to our conventions.

PR #3667, opened 2023-12-01
Prove a taboo for the proposition that all subcountable sets are countable

### LINKED ISSUE #3493: If every subcountable set is countable and Markov principle holds, then excluded middle holds
This is mentioned in the comment at https://mathoverflow.net/a/453340/489586

It is Proposition 2.6 of [Every metric space is separable in function realizability](https://doi.org/10.23638/LMCS-15%282%3A14%292019) by Andrej Bauer and Andrew Swan (which is already in the iset.mm bibliography as [BauerSwan]).

We already have a few theorems for countability at places like https://us.metamath.org/ileuni/ctm.html (and that whole section). Subcountability should be easily defined. The bit about "every truth value p is ¬¬-stable" implying excluded middle would be one piece which I don't think we have proved yet (we are close with https://us.metamath.org/ileuni/nnexmid.html but ideally we'd prove this expressed in terms of `EXMID`).

jkingdon:
This theorem, Proposition 2.6 of [BauerSwan], is mentioned at the end of the comment at https://mathoverflow.net/questions/453312/is-bauer-hanson-s-result-there-is-a-topos-where-the-dedekind-reals-are-countabl/453340#453340

Fixes #3493


iset.mm
```diff
@@ -7109,6 +7109,19 @@ Many theorems of this section actually hold for stable propositions (see
...
+     (Contributed by BJ, 9-Oct-2019.) $)
+  nnexmid $p |- -. -. ( ph \/ -. ph ) $=
+    ( wn wo wa pm3.24 ioran mtbir ) AABZCBHHBDHEAHFG $.
+
+  $( Double negation of decidability of a formula.  Intuitionistic logic
+     refutes undecidability (but does not prove decidability) of any formula.
+     (Contributed by BJ, 9-Oct-2019.) $)
+  nndc $p |- -. -. DECID ph $=
+    ( wdc wn wo nnexmid df-dc notbii mtbir ) ABZCAACDZCAEIJAFGH $.
+
```
benjub, line 7124: I think these belong the "Decidable propositions" section, not the "Theorems of decidable propositions" section (the latter is typically for theorems having decidability as an antecedent, and a classical tautology as a consequent).

jkingdon: I was wondering about that too but I wasn't sure.

Moved.

benjub (approved):

icecream17 (approved):

PR #3672, opened 2023-12-02
added ~ xlimlimsupleliminf:

glacode:
A sequence of extended reals converges if and only if its superior limit is smaller than or equal to its inferior limit.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -43,13 +43,13 @@ proposed  syl5ibcom imbitridcom
 proposed  syl5ibr   imbitrrid
 proposed  syl5ibrcom imbitrridcom
 proposed  syl5bb    bitrid      compare to bitri or bitrd
-proposed  syl5rbb   bitridcom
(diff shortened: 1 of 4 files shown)


glacode:
mathbox only

wlammen (approved):

tirix (commented):
Wouldn't it be more practical to go one step more and prove the equality?

glacode:
> Wouldn't it be more practical to go one step more and prove the equality?

it should be the theorem just above, isn't it?  (I use it in the trivial proof of this one)

This one is just to "skip" the obvious part of the equality (often times I've seen video lectures where the teacher proves convergence by proving that limsup <= liminf)

avekens:
~xlimlimsupleliminf is just the combination of ~xlimliminflimsup (which is about equality) and ~liminfgelimsupuz.

avekens (approved):

tirix (approved):

PR #3673, opened 2023-12-02
Prove revwlk

BTernaryTau:
> The reverse of a walk is a walk.


set.mm
```diff
@@ -500210,6 +500210,53 @@ have become an indirect lemma of the theorem in question (i.e. a lemma
       DUGUEBAIZMUIBEAPUJUFUEBAQRSTUAABCDEFGUBUC $.
   $}
 
+  ${
+    $d I x $.  $d P k x $.  $d F k x $.  $d G k x $.
+    revwlk.1 $e |- V = ( Vtx ` G ) $.
+    revwlk.2 $e |- I = ( iEdg ` G ) $.
```
avekens, line 500216: These hypothesies are not necessary, because V and I do not appear in the statement.

avekens: Therefore, they sould be removed (`V` and `I` replaced by ``( Vtx ` G )`` and ``( iEdg ` G )`` in the proof).

BTernaryTau: Done, thanks!

tirix (approved):

avekens (approved):
Using ~revrev, it should be possible to show the equivalence

`` revwlkb $p |- ( F ( Walks ` G ) P <-> ( reverse ` F ) ( Walks ` G ) ( reverse ` P ) ) ``

But this can be done in a subsequent PR.

PR #3675, opened 2023-12-03
Rename syl5eqss to eqsstrid

jkingdon:
This is in set.mm and iset.mm

As proposed in changes-set.txt.


benjub (approved):

jkingdon:
I realize this is a quick merge but that seems best to avoid conflicts

PR #3682, opened 2023-12-06
clean up aev section

wlammen:
1. Rename hbnaevg -> naev2.  Reason: this theorem is for ¬ ∀𝑥 𝑥 = 𝑦 what aev2 is for ∀𝑥 𝑥 = 𝑦.
2. Drop an unnecessary dv condition from naev2 (former hbnaevg)
3. aev2ALT -> aev2.  Allows dropping hbaevg.
4. Drop hbaevg.  This theorem is an instance of aev2.  Its application saved just a single proof byte in another theorem.
5. Replacement of hbaevg is now the former aev2ALT
6. Delete outdated OLD theorems

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -90,6 +90,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 6-Dec-23 hbaevg                deleted
+ 6-Dec-23 hbnaevg   naev2       similarity to aev2
  4-Dec-23 syl5ss    sstrid      compare to sstri or sstrd
  3-Dec-23 syl5eqss  eqsstrid    compare to eqsstri or eqsstrd
  1-Dec-23 syl5eqssr eqsstrrid   compare to eqsstrri or eqsstrrd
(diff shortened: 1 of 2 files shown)


set.mm
```diff
@@ -16950,24 +16950,39 @@ variables and no wff metavariables (see ~ ax12wdemo for an example of
...
     $( Version of ~ hbae with a disjoint variable condition, requiring fewer
-       axioms.  Instance of ~ hbaevg and ~ aev2 .  (Contributed by Wolf Lammen,
+       axioms.  Instance of ~ aev2 .  (Contributed by Wolf Lammen,
        22-Mar-2021.) $)
     hbaev $p |- ( A. x x = y -> A. z A. x x = y ) $=
-      ( hbaevg ) ABCBAD $.
+      ( aev2 ) ABCBAAD $.
     $( $j usage 'hbaev' avoids 'ax-8' 'ax-9' 'ax-10' 'ax-11' 'ax-12'
        'ax-13'; $)
   $}
```
tirix, line 16988: What is the need to keep `hbaev`?

tirix: (same question for `hbnaev`)

tirix: Ok, I did not see that those have the `v.vs` hypothesis in the scope, while the `hb` theorems, not.

tirix (approved):

benjub (approved):

PR #3684, opened 2023-12-07
Prove swrdwlk

BTernaryTau:
revpfxsfxrev

> The reverse of a prefix of a word is equal to the same-length suffix of the reverse of that word.

swrdrevpfx

> A subword expressed in terms of reverses and prefixes.

pfxwlk

> A prefix of a walk is a walk.

swrdwlk

> Two matching subwords of a walk also represent a walk.


icecream17 (approved):

tirix (approved):
Very nice!

benjub (approved):

PR #3695, opened 2023-12-14
Initial version of Igor Ieskov's mathbox.

expln:


jkingdon (approved):

tirix (approved):

set.mm
```diff
@@ -751469,4 +751469,97 @@ with simplest antecedents (i.e., in the corresponding ordering of binary
...
+      RSIRBCLZITIQICRSIQIMDEBCNO $.
+  $}
+
+  ${
+    abc3i.1 $e |- A e. CC $.
+    abc3i.2 $e |- B e. CC $.
+    abc3i.3 $e |- C e. CC $.
+    $( Cube of sum of three numbers.  (Contributed by Igor Ieskov,
+       14-Dec-2023.) $)
+    abc3i $p |- ( ( ( A + B ) + C ) ^ 3 ) =
```
avekens, line 751494: I propose to call this theorem cu3addi ("cu" fore "cube" as in ~cu2, "3add" for sum of three numbers" as in ~lt3addmuld)

set.mm
```diff
@@ -751469,4 +751469,97 @@ with simplest antecedents (i.e., in the corresponding ordering of binary
...
+      LNMVTWCVMUNVRWBVSUNUPVOVPYKAVNDYOOBVNEYOOQLNM $.
+  $}
+
+  ${
+    abc3d.1 $e |- ( ph -> A e. CC ) $.
+    abc3d.2 $e |- ( ph -> B e. CC ) $.
+    abc3d.3 $e |- ( ph -> C e. CC ) $.
+    $( Cube of sum of three numbers.  (Contributed by Igor Ieskov,
+       14-Dec-2023.) $)
+    abc3d $p |- ( ph -> ( ( ( A + B ) + C ) ^ 3 ) =
```
savask, line 751526: Maybe I'm missing something, but doesn't it make more sense to prove the deduction version first and then derive `abc3i` as a corollary?

expln: I have changed the proof and reordered the theorems, so now it looks like the deduction was proved first and then the inference was derived from it. 

In reality, I proved the inference first, because it was easier for me. After that I created the proof for the deduction by "mechanically" wrapping each line of the inference proof into `ph -> ( ... )` and reproving everything. But reproving everything was not hard, because I had to remove all justifications and then the proof assistant found new ones. I only had to create binom2d and add a few new steps for this to work.

set.mm
expln: Renamed. Thank you.

tirix (approved):

savask (approved):

icecream17 (approved):

avekens (approved):

set.mm
```diff
@@ -751469,4 +751469,88 @@ with simplest antecedents (i.e., in the corresponding ordering of binary
...
+    + ( ( ( 3 x. 2 ) x. ( A x. B ) ) x. C ) )
+    + ( 3 x. ( ( B ^ 2 ) x. C ) ) ) )
+    + ( ( ( 3 x. ( A x. ( C ^ 2 ) ) )
+    + ( 3 x. ( B x. ( C ^ 2 ) ) ) ) + ( C ^ 3 ) ) )
+    $=
+      ( caddc co c3 cexp c2 cmul wceq wn cc wcel a1i cu3addd pm2.18i ) ABGZHCTH
+      IZJZHAUAUBHUAAKZUBHZBLZHUEHTHUAABUCUBHZUEHUEHBUAUBHTHTHUAUDCUEHUEHUAUCUEH
+      ABUEHUEHCUEHTHUAUFCUEHUEHTHTHUAACUCUBHZUEHUEHUABUGUEHUEHTHCUAUBHTHTHMZUHN
+      ZABCAOZPUIDQBUJPUIEQCUJPUIFQRS $.
+  $}
```
avekens, line 751542: Interesting, that ~pm2.18i can be used instead of ~mptru (and therefore to avoid the usage ov "T.").

expln: This was found by mm-lamp. I even didn't pay attention to this.

david-a-wheeler: @avekens - I agree it's interesting, though in retrospect it's not surprising. Obviously this trick doesn't work in general in intuitionistic logic (iset.mm).

jkingdon: The version with `mptru`` is shorter (199 bytes in SHOW PROOF/COMPRESSED compared with 206 bytes for the pm2.18i one). I guess the reason the difference is so small is that the compressor can reuse the long expression. Ah yes, and in SHOW PROOF/SIZE the difference is 227 characters versus 278).

Sorry for geeking out on this sort of minutia, @expln .  Us metamath folks are vulnerable to this sort of rabbit trap.

expln: Does this mean I have to change the proof? Or it is fine to use pm2.18i?

jkingdon: Have to? Not on my account.

If you want to, feel free to make a follow up pull request, I guess I don't have a strong opinion on the matter although we do tend to prefer shorter proofs.

expln: > Have to?

Sorry, I don't know English very well, sometimes I use incorrect wording :)

As for the length of the proof. mm-lamp uses BFS to find a proof, where each tree node is an assertion application. If there are two possible proofs:

```
Proof1:
h1:: a short hypothesis
h2:: a short hypothesis
h3:: a short hypothesis
1:h1,h2,h3,h1,h2,h3,h1,h2,h3,h1,h2,h3,h1,h2,h3:asrt1: the goal statement
```

```
Proof2:
h1:: a short hypothesis
h2:: a short hypothesis
h3:: a short hypothesis
1:h1,h2:asrt2: a short intermediate statement
2:h3,1:asrt3: the goal statement
```

Then mm-lamp will find the first proof and stop searching, even though the first proof may be much longer then the second one. And I am not familiar enough with set.mm assertions and usual approaches, so I just submitted what the assistant found (after checking correctness of the proof by metamath.exe). The only assertion I remembered after creating these proofs is `addassi` because the assistant could not use it automatically somehow. But I will produce shorter proofs as I get more familiar with set.mm :)

expln: Just recently I looked at the difference between usage of pm2.18i and mptru and understood what's going on. I will submit a shorter proof soon.

PR #3686, opened 2023-12-10
Mathbox

benjub:
The diff looks huge. Not sure why (I moved around a big section within my mathbox). Locally, `git diff --patience` reduces it. Anyway, it's mathbox-only.

PR #3697, opened 2023-12-15
Added equs5av,sb4av and update BJ mathbox

GinoGiotto:
This is a small correction. Theorems `bj-equs45fv` and `bj-hbsb2av` were meant to not use ax-13, but it was later added in their axiom dependencies.  This PR fixes it by adding two convenience theorems.

More context here: https://github.com/metamath/set.mm/pull/3686#discussion_r1425357716


benjub (approved):
Thanks !

icecream17 (approved):

wlammen (approved):
nice!

PR #3698, opened 2023-12-15
fix df-selv

### LINKED ISSUE #3666: df-selv is incorrect?
image

Let `G` be the function highlighted here. (The last argument to `evalSub` in `selectVars`)

<details><summary>images of theorems</summary>

image
image
image

</details>

By selvval2lemn and rhmf, ``(((I evalSub S) ` ran C) : ( Base ` W ) --> ( Base ` T )``

By selvval2lem4 we have ``(C o. F) e. ( Base ` W )``

By pwsbas and elmapg, ``( Base ` T ) = ( B ^m ( B ^m I ) )``, so:

``((((I evalSub S) ` ran C) ` (C o. F)) : (anyfunction I --> B) --> B``.

Which is inconsistent for two reasons. 

1. First the domain doesn't match: `G` is not a function `I --> B`
    <details><summary>~ mvrcl image</summary>
    
    image

    Note that `C` maps `BaseR` to `BaseS` as seen in step 20 of https://us.metamath.org/mpeuni/selvval2lem4.html

    </details>

    ...rather it is `I --> Base(J mPoly S) or Base(I-J mPoly S)` (where S is `I-J mPoly R`)
 
   1. This could be fixed by changing any result of ``(G`x)`` to ``((G`x)`x)``. That way it would collapse to `I --> Base(S)` i.e. `I --> B`

2. And lastly the purpose of df-selv is to take a subset of indices `J` and a polynomial `F : I --> R` and return a polynomial
    `? : J --> I-J --> R`

    But instead we get a polynomial in ``B = ( Base ` S )`` i.e.
    `? : I-J --> R`, so the range doesn't match either. I'm not sure how this could be fixed

icecream17:
update corresponding theorems (selv*val*)

extract issubassa3 from issubassa
move issubassa2 for use in rnasclassa

fixes #3666


tirix (approved):
Looks good.
I guess we will need more to formally confirm that the formula is correct.

set.mm
```diff
@@ -250723,6 +250723,22 @@ such that every prime ideal contains a prime element (this is a
     $d x y z A $.  $d x y z L $.  $d x y z S $.  $d x y z W $.
     issubassa.s $e |- S = ( W |`s A ) $.
     issubassa.l $e |- L = ( LSubSp ` W ) $.
+    $( A subring that is also a subspace is a subalgebra.  The key theorem is
```
wlammen, line 250726: How does this comment come about?  I don't find a matching theorem from MC from 7-Jan-2015.

icecream17: It's a modified version of the comment from ~ issubassa where instead of a biconditional, the sentences states one direction only.

The part about the key theorem would've been like a note to past me that would explain how LSubSp relates to AssAlg (although now looking at it again, this theorem could just use A e. LMod instead of A e. L)

tirix:
@wlammen Was your comment addressed?
Could we consider merging this contribution?

wlammen (commented):
I found the conclusion of issubassa3 in issubassa, and it is therefore a subproof of the latter proof, as SN wrote.  I am not sure whether the Contributed by... tag should be a copy of that of the enclosing theorem.  I'd rather see something like "Based on ideas of MC" in the comment.  But this is not blocking, given the long time this PR is already open.

EDIT: see ax13lem1 for example.

benjub:
> I found the conclusion of issubassa3 in issubassa, and it is therefore a subproof of the latter proof, as SN wrote. I am not sure whether the Contributed by... tag should be a copy of that of the enclosing theorem. I'd rather see something like "Based on ideas of MC" in the comment. But this is not blocking, given the long time this PR is already open.

As for me, it's fine either way, as @icecream17 prefers.

benjub (approved):

wlammen:
Merging this PR created a conflict that was corrected in #3732.

I am still not a fan of adding sth on behalf of someone else.  We have means of attributing to prior art, not written out as a rule, but found as common habit in set.mm.  That said I have nothing to object from a technical perspective, and there are two approvals.  So lets move on.

PR #3702, opened 2023-12-17
mathbox: direct image

benjub:
Apart from a typo and two minor formatting changes, mathbox-only.


set.mm
```diff
@@ -534240,14 +534252,153 @@ Identity relation (complements)
...
+  $( Token for the (object component of the) direct image functor. $)
+  $c ~P_* $.
+
+  $( Syntax for the (object component of the) direct image functor. $)
+  cimdir $a class ~P_* $.
+
+  ${
+    $d a b r x y $.
+    $( Definition of the (object component of the) direct image functor, which
+       maps a binary relation between to given sets to its associated direct
```
icecream17, line 534342:
```suggestion
       maps a binary relation between two given sets to its associated direct
```

set.mm
```diff
@@ -534240,14 +534252,153 @@ Identity relation (complements)
...
+$)
+
+  ${  $d A x y $.  $d B x y $.  $d ch x y $.
+    opelopabd.x $e |- ( ph -> A. x ph ) $.
+    opelopabd.y $e |- ( ph -> A. y ph ) $.
+    opelopabd.exa $e |- ( ph -> A e. U ) $.
+    opelopabd.exb $e |- ( ph -> B e. V ) $.
+    opelopabd.is $e |- ( ( ph /\ ( x = A /\ y = B ) ) -> ( ps <-> ch ) ) $.
+    $( Membership of a couple in a class abstraction of couples.  Note: this
+       could be proved fomr an anlogous version "copsex2d" of ~ copsex2t .
```
icecream17, line 534288:
```suggestion
       could be proved from an analogous version "copsex2d" of ~ copsex2t .
```

icecream17 (approved):

set.mm
```diff
@@ -531826,6 +531823,21 @@ to being a member of a class abstraction (that every class is equal to a
                 ( A e. B <-> ( A e. _V /\ ph ) ) ) $=
     ( wcel cvv elex biadani ) BCDBEDABCFG $.
 
+  $( A set is a binary relation if and only if it belongs to the powerclass of
```
avekens, line 531826: Is it "powerclass" or "power class"? We have  "powerclass" 13 times and "power class" 85 times in set.mm.

tirix:
I might be taking some time to review this PR since I'd like to get a grasp of the math that it opens to. In any case, I have no issue with this PR since it is only impacting your mathbox.
Is the [wikipedia page](https://en.wikipedia.org/wiki/Direct_image_functor) the right entry point?

According to that page the direct image product is a functor between categories of sheaves.
We do have a definition of functors between categories, [`~df-func`](https://us.metamath.org/mpeuni/df-func.html), but your definition does not match such a functor.

Should we consider the more generic definition as a functor between categories of sheaves?
What do we miss if we limit ourselves to sets?

set.mm
benjub: If we want to uniformize, then I'd vote for "powerclass", because it's one word for one concept.

benjub:
> Is the [wikipedia page](https://en.wikipedia.org/wiki/Direct_image_functor) the right entry point?

Ahh, you're making me realize I wrote nonsense ! I wanted to write "the morphism component" ! Actually, what I'm defining is much simpler than the sheaf operation: I will write "the functionalized direct image, which is the morphism component of the covariant powerset functor".

The "covariant powerset functor" is the following endofunctor of the category of sets and functions (and there is one ("the same") for the category of sets and relations as well):
* the set $X$ is sent to $\mathcal{P}X$,
* the morphism $f \colon A \to B$ is sent to $f_* \colon \mathcal{P}A \to \mathcal{P}B, X \mapsto f(X)$,

which in pseudo-metamath is written a bit like $( (A \mathcal{P}_* B) \text{ ` } f ) : X \mapsto ( f \text{ " }X)$.

You have the functorial property ``(f \circ g)_* = f_* \circ g_*``.

Then, there is also a contravariant endofunctor, the inverse image $f^*$.

At that level, I think it is not worth using the heavy machinery of categories of set.mm, at least for the moment, and if it is added later, it would be in addition to this plain non-categorical version, not as a replacement.

set.mm
```diff
@@ -531826,6 +531823,21 @@ to being a member of a class abstraction (that every class is equal to a
...
+     the cartesian square of the universal class.  (Contributed by BJ,
+     16-Dec-2023.) $)
+  relelpw $p |- ( A e. V -> ( Rel A <-> A e. ~P ( _V X. _V ) ) ) $=
```
mazsa, line 531829: This is very similar to ~ elrelsrel (except that it does not use the associated nullary class constant ~ df-rels as recommended in point 2 of our Guidelines).

tirix (approved):
Thanks for the explanations and precisions!

PR #3703, opened 2023-12-17
shorten trunorfal, falnorfal

wlammen:
1. shorten trunorfal
2. shorten falnorfal
3. shorten norass by another proof line (no OLD version, follow up of a same day contribution)

norass contains a proof line
((𝜑 ↔ 𝜒) ↔ ((𝜑 → 𝜒) ↔ (𝜒 → 𝜑)))
with an operation ↔ in the RHS weaker than that of dfbi2
((𝜑 ↔ 𝜓) ↔ ((𝜑 → 𝜓) ∧ (𝜓 → 𝜑)))

Shall I extract this as a lemma, so it can be found, in case someone needs it?  Is it important enough, what do you think?

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17626,6 +17626,7 @@ New usage of "trsspwALT2" is discouraged (0 uses).
 New usage of "trsspwALT3" is discouraged (0 uses).
 New usage of "truniALT" is discouraged (0 uses).
 New usage of "truniALTVD" is discouraged (0 uses).
+New usage of "trunorfalOLD" is discouraged (0 uses).
 New usage of "trunortruOLD" is discouraged (0 uses).
 New usage of "ubth" is discouraged (1 uses).
 New usage of "ubthlem1" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


benjub:
> Shall I extract this as a lemma, so it can be found, in case someone needs it? Is it important enough, what do you think?

I think it is worth adding it.

benjub (approved):

tirix (approved):
I'm not sure it's worth adding "((𝜑 ↔ 𝜒) ↔ ((𝜑 → 𝜒) ↔ (𝜒 → 𝜑)))".
Do you have any other example in mind where it could be useful?

wlammen:
I don't have a specific application in mind.  When designing the proof of norass I looked out for such a theorem, to no avail of course.  There are two aspects that I had in my mind:
1.  Use the most general expression when possible in a theorem, since this is usually slightly easier to prove and has a broader usage.
2. The proof of norass is tricky, a lemma extracting a particular idea renders it more readable.

benjub:
I agree with both of @wlammen's points. Furthemore, that statement is noteworthy enough: it is short, organic, a most general instance, uses only two connectives and two variables, is a classical tautology but not an intuitionistic one.

tirix:
Yes, the statement might be interesting by itself, even if its single use is the shortening of ~ norass.
Maybe there's actually no need to find other uses to justify it.

PR #3706, opened 2023-12-18
shorten hadcoma

wlammen:
1. extend the description of df-had and add its purpose
2. shorten hadcoma

The hadd* series of theorems is based on the ⊻operator, starting with a definition
(hadd(𝜑, 𝜓, 𝜒) ↔ ((𝜑 ⊻ 𝜓) ⊻ 𝜒)) .
This is somewhat weird, since hadbi gives an equivalent definition
(hadd(𝜑, 𝜓, 𝜒) ↔ ((𝜑 ↔ 𝜓) ↔ 𝜒))
in terms of the bi-conditional.
The advantage of ⊻ is that the purpose of hadd is a bit easier to spot,  and the half-adding with carry in electronics is often demonstrated (implemented) using xor-gates.  The disadvantage is that ↔ is covered way more broadly, allowing for shortenings.  If you want to keep the ⊻ relevance for pedagogical reasons or so, tag the theorems as not modifiable.

EDIT:  The reference to a half-adder for the definition and the following theorems is, strictly speaking, incorrect.  See https://en.wikipedia.org/wiki/Adder_(electronics)#Half_adder .  This is somewhere in between a half and a full adder.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15369,6 +15369,7 @@ New usage of "h2hnm" is discouraged (2 uses).
 New usage of "h2hsm" is discouraged (9 uses).
 New usage of "h2hva" is discouraged (9 uses).
 New usage of "h2hvs" is discouraged (2 uses).
+New usage of "hadcomaOLD" is discouraged (0 uses).
 New usage of "halfnq" is discouraged (1 uses).
 New usage of "hatomic" is discouraged (1 uses).
 New usage of "hatomici" is discouraged (6 uses).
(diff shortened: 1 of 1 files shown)


tirix (approved):

icecream17 (approved):

benjub:
> EDIT: The reference to a half-adder for the definition and the following theorems is, strictly speaking, incorrect. See https://en.wikipedia.org/wiki/Adder_(electronics)#Half_adder . This is somewhere in between a half and a full adder.

I rewrote the section comments in #1773 and I think that did clear up the ambiguities. It is true that the old labels `had*` stayed, which is not great, but I think that's ok, given that the subsection titles and section comment give details.

benjub (approved):

PR #3707, opened 2023-12-18
rewrite norass to use lemmas

wlammen:
Restructure the proof of norass such that repeated steps can be moved to lemmas.  One lemma subsumes the idea introduced in #3703 .

EDIT: The total number of proof bytes reduces from 341 to 320, proof lines from 44 to 40.


set.mm
```diff
@@ -11796,21 +11796,40 @@ This definition (in the form of ~ dfifp2 ) appears in Section II.24 of
     ( wnor wo wn notnotb df-nor notbii nornot 3bitr2ri bicomi ) ABCZLCZABDZNNEZ
     ELEMNFLOABGHLIJK $.
 
+  $( This lemma of ~ norass connects the bi-conditional to implications in a
+     different way than ~ dfbi2 does.  (Contributed by Wolf Lammen,
+     18-Dec-2023.) $)
+  norass-lem1 $p |- ( ( ph <-> ps ) <-> ( ( ph -> ps ) <-> ( ps -> ph ) ) ) $=
+    ( wb wi wa dfbi2 pm5.1 sylbi impbi wn conax1 pm5.21im syl2imc bija impbii )
+    ABCZABDZBADZCZPQRESABFQRGHQRPABIRJAJQJBJPBAKABKABLMNO $.
```
benjub, line 11804: To me, this statement is interesting in its own right, so you could move it much earlier, with a specific label. `dfbi4` ?

set.mm
```diff
@@ -11796,21 +11796,40 @@ This definition (in the form of ~ dfifp2 ) appears in Section II.24 of
...
+
+  $( This lemma shows the equivalence of two expressions, used in ~ norass .
+     (Contributed by Wolf Lammen, 18-Dec-2023.) $)
+  norass-lem2 $p |- ( ( ( ph \/ ps ) -> ch ) <-> ( ( ph -\/ ps ) \/ ch ) ) $=
```
benjub, line 11808: Our labeling conventions want no dash (so `norasslem1`, etc.).

set.mm
wlammen: Unfortunately it is not a definition, since <-> appears on both sides.  The name should not start with df.

wlammen: I decided on impimprbi. To avoid any confusion with biimp/biimpr the bi was suffixed.

set.mm
```diff
@@ -7067,6 +7067,14 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
   pm4.45im $p |- ( ph <-> ( ph /\ ( ps -> ph ) ) ) $=
     ( wi ax-1 pm4.71i ) ABACABDE $.
 
+  $( An implication and its reverse are equivalent only when both operands are
+     equivalent.  The right hand side resembles that of ~ dfbi2 , but ` <-> `
+     is a weaker opertor than ` /\ ` .  (Contributed by Wolf Lammen,
```
icecream17, line 7072:
```suggestion
     is a weaker operator than ` /\ ` .  (Contributed by Wolf Lammen,
```

set.mm
```diff
@@ -11796,21 +11804,33 @@ This definition (in the form of ~ dfifp2 ) appears in Section II.24 of
     ( wnor wo wn notnotb df-nor notbii nornot 3bitr2ri bicomi ) ABCZLCZABDZNNEZ
     ELEMNFLOABGHLIJK $.
 
+  $( This lemma shows the equivalence of two expressions, used in ~ norass .
+     (Contributed by Wolf Lammen, 18-Dec-2023.) $)
+  norasslem1 $p |- ( ( ( ph \/ ps ) -> ch ) <-> ( ( ph -\/ ps ) \/ ch ) ) $=
+    ( wo wi wn wnor imor df-nor orbi1i bitr4i ) ABDZCELFZCDABGZCDLCHNMCABIJK $.
+
+  $( This lemma specializes ~ biimt suitable for the proof of ~ norass .
```
icecream17, line 11812:
```suggestion
  $( This lemma specializes ~ biimt suitably for the proof of ~ norass .
```

set.mm
```diff
@@ -11796,21 +11804,33 @@ This definition (in the form of ~ dfifp2 ) appears in Section II.24 of
...
+     (Contributed by Wolf Lammen, 18-Dec-2023.) $)
+  norasslem2 $p |- ( ph -> ( ps <-> ( ( ph \/ ch ) -> ps ) ) ) $=
+    ( wo wi wb orc biimt syl ) AACDZBJBEFACGJBHI $.
+
+  $( This lemma specializes ~ biorf suitable for the proof of ~ norass .
```
icecream17, line 11817:
```suggestion
  $( This lemma specializes ~ biorf suitably for the proof of ~ norass .
```

icecream17 (approved):

set.mm
wlammen: done

set.mm
```diff
@@ -7069,8 +7069,9 @@ _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
 
   $( An implication and its reverse are equivalent only when both operands are
      equivalent.  The right hand side resembles that of ~ dfbi2 , but ` <-> `
-     is a weaker operator than ` /\ ` .  (Contributed by Wolf Lammen,
-     18-Dec-2023.) $)
+     is a weaker operator than ` /\ ` .  Note that an implication and its
+     reverse can never be simultaniously false, because of ~ pm2.521 .
```
benjub, line 7073:
```suggestion
     reverse can never be simultaneously false, because of ~ pm2.521 .
```
also, I would change the "only when" to "exactly when", three lines above.

wlammen: done

icecream17 (approved):

PR #3709, opened 2023-12-21
Main: pwvrel, pwvabrel; mathbox: bj-pwvrelb

benjub:
As noticed by @mazsa in https://github.com/metamath/set.mm/pull/3702#discussion_r1430748820 I recently introduced a theorem very similar to one of his. Therefore, this MR moves it to Main: this is pwvrel. @mazsa : are the credits fine with you ? I'm also adding pwvabrel, for its own sake. I'm also relabeling bj-pwvrelb, but keeping it in my mathbox.

@mazsa : I think it is not worth introducing `Rels` since it is not much longer to write `~P ( _V X. _V )`, which also saves the reader from learning a new definition.


mazsa:
> As noticed by @mazsa in [#3702 (comment)](https://github.com/metamath/set.mm/pull/3702#discussion_r1430748820) I recently introduced a theorem very similar to one of his. Therefore, this MR moves it to Main: this is pwvrel. @mazsa : are the credits fine with you ? 

@benjub yes, thank you.

> @mazsa : I think it is not worth introducing `Rels` since it is not much longer to write `~P ( _V X. _V )`, which also saves the reader from learning a new definition.

I don't think there's any need for an introduction here either: we'll probably discuss it in a different context.

mazsa (approved):

avekens (approved):

PR #3710, opened 2023-12-22
Add chain rule for derivatives to iset.mm

jkingdon:
The main catch here is that there is a condition on the function G (roughly, that it is strictly monotonic, but this is complex numbers and apartness rather than real numbers and less-than, so that might not be exactly the word).

With that proviso, the set.mm proof isn't especially hard to adapt.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11098,10 +11098,7 @@ intuitionistic and it is lightly used in set.mm</TD>
 
 <tr>
   <td>dvcobr</td>
-  <td><i>none</i></td>
-  <td>the set.mm proof depends on limcco and also would
-  need a lot of revision for topics related to complex
-  number decidability and apart versus not equal.</td>
+  <td>~ dvcoapbr</td>
 </tr>
 
 <tr>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -86216,6 +86216,38 @@ would be equivalent to negated equality and this would follow readily (for
...
+      WFYHYIVMVIVJVNVOVPVQYPWNBQYPYQYRWNBMYTUUAABTTVRVGYPWNWRQYOWMWSVMYOWRQRZWT
+      YNUUCYKYNWOWQYLWOQRYMWOVSVTYNKWPUUBYNVKVLYMWPQRYLWPVSWAVNVOWAVTVPVQWBWCWD
+      WDWE $.
+  $}
+
+  ${
+    $d A x y $.  $d B x y $.
+    $( The points apart from a given point are complex numbers.  (Contributed
+       by Jim Kingdon, 19-Dec-2023.) $)
+    apsscn $p |- { x e. A | x =//= B } C_ CC $=
```
avekens, line 86246: Do we need `x e. A`? I think already `{ x | x =//= B } C_ CC` would be true.

jkingdon: Yes `{ x | x =//= B } C_ CC` would also be provable but `{ x e. A | x =//= B }` is the one we are likely to use more. This is in the literature using names like "apartmess space" and sometimes has its own notation.

avekens (approved):

PR #3713, opened 2023-12-25
alternate proof of 0 + 0 = 0

icecream17:
<details><summary>Extra commentary</summary>

This seems like somewhat significant progress on whether ~ ax-mulcom is independent, as almost all theorems about numbers end up using ~ 00id eventually. Some links:

[Any proof of ~ ax-mulcom would have to use most complex number axioms](https://gist.github.com/icecream17/933f95d820e0b8f1cab0d4293b68eaf9).

[Bullet point 2 in my todo list further explains how to save axioms in the current 00id (but not ~ ax-mulcom)](https://gist.github.com/icecream17/c85a1288ccebd09e21c7609943a9a8d6#axioms). But I still have doubts about the custom version of adddir etc so this change won't happen for a while. It's comparable to https://us.metamath.org/mpeuni/nnne0.html / https://github.com/metamath/set.mm/pull/3018. With similar reasoning, most of my mathbox work regarding complex number axioms will stay in my mathbox.

</details>

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -19284,6 +19284,7 @@ Proof modification of "simplbi2comtVD" is discouraged (42 steps).
 Proof modification of "sineq0ALT" is discouraged (986 steps).
 Proof modification of "smgrpassOLD" is discouraged (54 steps).
 Proof modification of "smgrpismgmOLD" is discouraged (22 steps).
+Proof modification of "sn-00id" is discouraged (50 steps).
 Proof modification of "snelpwrVD" is discouraged (37 steps).
 Proof modification of "snex" is discouraged (64 steps).
 Proof modification of "snexALT" is discouraged (48 steps).


benjub:
When not only the statement, but also the proof is important, I advise to put a "(Proof modification is discouraged.)" tag, preceded by what property of the proof should be kept (e.g., not using this or that axiom).

I haven't looked into the details. You write you made significant progress, but you don't say in which direction ? Towards independence ?

benjub (approved):

icecream17:
Towards redundancy in this case, since this greatly expands what can be done without ax-mulcom.

To clarify ax-mulcom's independence is still an open problem.
And clarify that I _also_ tried to prove towards independence, which is what the [first link in "Extra commentary"](https://gist.github.com/icecream17/933f95d820e0b8f1cab0d4293b68eaf9) above was. A surprising amount of progress was made in both directions so I still think this question could go either way.

avekens (approved):

PR #3714, opened 2023-12-25
Labels

benjub:
Two relabelings (both in set.mm and iset.mm):
* selpw-->velpw (the "v" for "variable" as in ~cv, ~vex, etc.)
* prth-->anim12 to follow conventional naming

Add 'v' label convention and reformat some entries.

Best reviewed by commits.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -75,16 +75,16 @@ proposed  sseqtr4i  sseqtrri
 proposed  sseqtr4d  sseqtrrd
 proposed  syl6sseqr sseqtrrdi
 proposed  syl6eqss  eqsstrdi    compare to eqsstri or eqsstrd
-proposed  syl6eqssr eqsstrrdi
+proposed  syl6eqssr eqsstrrdi   compare to eqsstrri or eqsstrrd
 proposed  syl6eqbr  eqbrtrdi    compare to eqbrtri or eqbrtrd
 proposed  syl6eqbrr eqbrtrrdi   compare to eqbrtrri or eqbrtrrd
 proposed  syl6breq  breqtrdi    compare to breqtri or breqtrd
-proposed  syl6breqr breqtrrdi   compare to breqtrri or breqtrrd
 (Please send any comments on these proposals to the mailing list or
 make a github issue.)
 
 DONE:
 Date      Old       New         Notes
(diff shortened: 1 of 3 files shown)


tirix (approved):

icecream17 (approved):

set.mm
```diff
@@ -6908,8 +6908,8 @@ use disjunction (although this is not required since definitions are
      It was proved by Leibniz, and it evidently pleased him enough to call it
      _praeclarum theorema_ (splendid theorem).  (Contributed by NM,
      12-Aug-1993.)  (Proof shortened by Wolf Lammen, 7-Apr-2013.) $)
-  prth $p |- ( ( ( ph -> ps ) /\ ( ch -> th ) )
-       -> ( ( ph /\ ch ) -> ( ps /\ th ) ) ) $=
+  anim12 $p |- ( ( ( ph -> ps ) /\ ( ch -> th ) )
+              -> ( ( ph /\ ch ) -> ( ps /\ th ) ) ) $=
     ( wi id im2anan9 ) ABEZABCDEZCDHFIFG $.
```
avekens, line 6913: Shall we really replace the "historical" name praeclarum theorema (~prth in short) by the technical name ~anim12? Some time ago, I suggested to replace the names of some "stoic" theorems by  their technical counterparts, but this was strongly rejected by @david-a-wheeler I think. Furthermore, it is exercised practice in set.mm to use a corresponding label if the (mostly famous) theorem has a special name (e.g., ~cramer for Cramer's rule, and not ~sollineqmat or something similar).

benjub: It is not obvious that the label "prth" stands for "praeclarum theorema". And then, what that "brilliant theorem" could be is not clear (Leibniz could receive a "poorest naming ever" for this, by the way). On the other hand, you could guess the statement of "anim12" just by its label.

Actually, here is how I discovered this problem: I needed "anim12" in a proof, so I typed "ASSIGN LAST anim12" but that was rejected by MM, so I went to add it to set.mm, and then, minimizing its proof, I realized that it was already in set.mm with label "prth". I had to look at its comment to understand why it was thus labeled.

benjub: Merging after two days, for fear of further merge conflicts.
To clarify what I wrote above: the name "praeclarum theorema" should stay in the comment, since it somehow stuck, but the label "prth" does not really hint at that name (unless we already know it).

PR #3716, opened 2023-12-26
opabresid / mptresid

benjub:
For these two theorems, use the usual definitional form "short expr = long expr" instead of the computational form "long expr = result".
The result should be squashed: the commits were made only for clarity. For the updating of the proofs, I used a script roughly as follows. The important part is to do on each proof:
```
MM-PA> expand OLD
MM-PA> minimize */except OLD
```
(after having proved the OLD version from the NEW version).  In full:
```
! MM> read set.mm
! MM> open log 1.tmp
! MM> show usage OLD
! MM> close log
! MM> submit "update"
! MM> tools
! TOOLS> substitute 1.tmp xxx OLD a ""
! TOOLS> exit
! MM> open log log.txt                (recommended, to review result later)
! MM> submit 1.tmp
! MM> write source set.mm /rewrap
! MM> close log

tools
match 1.tmp '(None)' n
match 1.tmp 'MM>' n
match 1.tmp '  ' y
break 1.tmp ''
unduplicate 1.tmp
add 1.tmp 'prove ' '$expand xxx$min */exc xxx$save new_proof/compressed$_exit_pa$'
! If you want to do a "dry run" without saving proofs, change above line
! to:
! add 1.tmp 'prove ' '$min xxx/no_new ax-*$_exit_pa/force$'
substitute 1.tmp '$' '\n' a ''
quit
```

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16328,6 +16328,7 @@ New usage of "mobiiOLD" is discouraged (0 uses).
 New usage of "moeuOLD" is discouraged (0 uses).
 New usage of "moimiOLD" is discouraged (0 uses).
 New usage of "mpteq12dvOLD" is discouraged (0 uses).
+New usage of "mptresidOLD" is discouraged (0 uses).
 New usage of "mptssALT" is discouraged (0 uses).
 New usage of "mpv" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


benjub:
Closing in favor of #3720.

PR #3718, opened 2023-12-27
Rename syl6breqr to breqtrrdi

jkingdon:
This is in set.mm and iset.mm

There's also a small tweak to the changes-set.txt entry for syl6eqssr


benjub (approved):
It's convenient to change to a label with same length !

jkingdon:
> It's convenient to change to a label with same length !

Indeed. A consequence of "syl6" being the same number of characters as "tr" plus "di"

PR #3724, opened 2023-12-29
Theorems about magmas, semigroups and monoids

avekens:
main set.mm:
* ~uzidd moved from GS's mathbox
* minor adjustment of comment for ~dfgsum

Maybe some of the following theorems can be moved to main immediately (candidates marked with (*))

AV's mathbox (new theorems):
* ~lidridid(*) (left and right identity elements are identical)
* ~mndbn0(*) (generalization of ~grpn0, whose proof cannot be shortened)
* ~nn0mnd (the set of nonnegative integers is a monoid)
* ~gsumsplit1r(*) (splitting off the rightmost summand of a group sum, regardless of `G` being a group, monoid or even a magma)
* ~gsumsgrpccat(*) (generalization of ~gsumccat for semigroups; proof of ~gsumccat can be shortened)
* ~gsumreidx (Re-indexing a finite group sum, variant of ~gsumf1o)
* ~gsumfsupp (restriction of a group sum to the support, variant of ~gsumres)
* ~gsumxp2(*) (group sum over a cartesian product)
* ~mulgnngsum, ~mulgnn0gsum(*) (group multiple (exponentiation) expressed by a group sum)
* ~cycsubmel, ~cycsubmcl, ~cycsubm(*) (cyclic monoids generated by the element, analogous to  ~cycgrpmcl)
* ~smndlsmidm, mndlsmidm(*) (the direct product is idempotent for submonoids, generalization of ~lsmidm)
** Remark: The "inner (or better: internal?) direct product" (see comment of ~df-lsm) operator is called `LSSum` (what does it mean? What does the label fragment "lsm" mean?) and is often called "subgroup sum", which can easily be confused with "group sum". Maybe it should be called "internal direct sum" (see https://en.wikipedia.org/wiki/Direct_sum#Internal_and_external_direct_sums), and the symbol for the operator could be `DSum`. It should not be called "product", because we have already a definition for internal direct products (DProd, see ~df-dprd).

AV's mathbox (misc):
* some rearrangements of theorems
* some alignments of section headers

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -84,6 +84,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+29-Dec-23 uzidd     [same]      moved from GS's mathbox to main set.mm
 26-Dec-23 syl6breqr breqtrrdi   compare to breqtrri or breqtrrd
 25-Dec-23 prth      anim12
 25-Dec-23 selpw     velpw
--- a/discouraged
+++ b/discouraged
@@ -18793,6 +18793,7 @@ Proof modification of "ghomlinOLD" is discouraged (161 steps).
 Proof modification of "grpinvfvalALT" is discouraged (161 steps).
 Proof modification of "grposnOLD" is discouraged (291 steps).
 Proof modification of "grpsubfvalALT" is discouraged (189 steps).
+Proof modification of "gsumccatNEW" is discouraged (310 steps).
 Proof modification of "hadcomaOLD" is discouraged (37 steps).
 Proof modification of "hashge3el3dif" is discouraged (229 steps).
 Proof modification of "hba1-o" is discouraged (34 steps).
@@ -18885,6 +18886,7 @@ Proof modification of "latmassOLD" is discouraged (309 steps).
 Proof modification of "lcmftp" is discouraged (1057 steps).
 Proof modification of "lediv2aALT" is discouraged (264 steps).
 Proof modification of "leibpilem1OLD" is discouraged (254 steps).
+Proof modification of "lsmidmNEW" is discouraged (25 steps).
 Proof modification of "luk-1" is discouraged (73 steps).
 Proof modification of "luk-2" is discouraged (52 steps).
 Proof modification of "luk-3" is discouraged (20 steps).


set.mm
```diff
@@ -214931,7 +214941,7 @@ is an important property of monoids (see ~ mndid ), and therefore also for
 
 $(
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
-  Ordered sums in a magma
+  Group sum operation for magmas
```
benjub, line 214944: A clearer naming for that operation would be "Iterated composition" (to use also at other places).
And "*in* magmas" sounds better.

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+
+  ${
+    gsumfsupp.b $e |- B = ( Base ` G ) $.
+    gsumfsupp.z $e |- .0. = ( 0g ` G ) $.
+    gsumfsupp.s $e |- I = ( F supp .0. ) $.
+    gsumfsupp.g $e |- ( ph -> G e. CMnd ) $.
+    gsumfsupp.a $e |- ( ph -> A e. V ) $.
+    gsumfsupp.f $e |- ( ph -> F : A --> B ) $.
+    gsumfsupp.w $e |- ( ph -> F finSupp .0. ) $.
+    $( A group sum is identical with the group sum restricted to the finite
```
benjub, line 736020:
```suggestion
    $( A group sum is identical with the group sum restricted to the
```

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+       support of its index set.  This corresponds to the definition of an
+       (infinite) product in [Lang] p. 5, two last formulas.  (Contributed by
```
benjub, line 736022:
```suggestion
       (infinite) product in [Lang] p. 5, last two formulas.  (Contributed by
```

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+      FGAFLUOAFVTQZFVPQZFDQAFUPSZWBWCUJLCVPFVCTAFVODAVSFVOSMFGUQTURUSMAOUTZFUIU
+      BPGRPZSVAWEVODAWFVOWEAWDWFVOUKLFGVBTVDURVEVFVGVH $.
+  $}
+
+  ${
+    $d x y z B $.  $d x y z G $.  $d x y z .+ $.  $d x y z W $.  $d x y z X $.
+    gsumsgrpccat.b $e |- B = ( Base ` G ) $.
+    gsumsgrpccat.p $e |- .+ = ( +g ` G ) $.
+    $( Homomorphic property of not empty composites of a group sum over a
+       semigroup.  Formerly part of proof for ~ gsumccat .  (Contributed by AV,
```
benjub, line 735893: It's more of an associative property, so:
"Associative property of the semigroup law: the composition of the iterated compositions of two nonempty finite sequences is equal to the iterated composition of their concatenation. (Recall that "words" are finite sequences, that is, functions on integer intervals ` ( 0 .. N ) ` for some ` N e. NN0 ` .)"
(I'm quoting the last parenthesis from memory, please check.)

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+      IABECOYSUCFGUUNUWHYIUWNAETZUWIAETYIYDUWRYAYCYDYHXTAEWGSYIUWNUWIAEUWPWNWOX
+      QXOXS $.
+  $}
+
+  ${
+    gsumccatNEW.b $e |- B = ( Base ` G ) $.
+    gsumccatNEW.p $e |- .+ = ( +g ` G ) $.
+    $( Homomorphic property of composites.  Second formula in [Lang] p. 4.
+       (Contributed by Stefan O'Rear, 16-Aug-2015.)  (Revised by Mario
+       Carneiro, 1-Oct-2015.)  (Proof shortened by AV, 26-Dec-2023.)
```
benjub, line 735934: As above, removing nonemptiness condition.

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+$)
+
+  ${
+    $d B i x $.  $d F i $.  $d N i x $.  $d X i x $.
+    mulgnngsum.b $e |- B = ( Base ` G ) $.
+    mulgnngsum.t $e |- .x. = ( .g ` G ) $.
+    mulgnngsum.f $e |- F = ( x e. ( 1 ... N ) |-> X ) $.
+    $( Group multiple (exponentiation) operation at a positive integer
+       expressed by a group sum.  (Contributed by AV, 28-Dec-2023.) $)
+    mulgnngsum $p |- ( ( N e. NN /\ X e. B ) -> ( N .x. X ) = ( G gsum F ) ) $=
```
benjub, line 736068: It looks like a lemma for `mulgnn0gsum`. Maybe it can even be inlined.

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+      ( vi cn wcel wa cfv c1 cseq co adantr cv cplusg csn cxp cuz elnnuz biimpi
+      cgsu cfz cmpt wceq a1i eqidd simpr fvmptd elfznn fvconst2g syl2an seqfveq
+      weq eqtr4d cvv eqid elfvex eleq2s adantl fmptd gsumval2 mulgnn 3eqtr4rd
+      cbs ) FLMZGBMZNZFEUAOZDPQOFVNLGUBUCZPQZOEDUGRFGCRVMVNKDVOPFVKFPUDOMZVLVKV
+      QFUEUFSZVMKTZPFUHRZMZNZVSDOGVSVOOZWBAVSGGVTDBDAVTGUIUJWBJUKWBAKUSNGULVMWA
+      UMVMVLWAVKVLUMZSUNVMVLVSLMWCGUJWAWDVSFUOLGVSBUPUQUTURVMBVNDEPFVAHVNVBZVLE
+      VAMZVKWFGEVJOBGEVJVCHVDVEVRVMAVTGBDVMVLATVTMWDSJVFVGBVNVPCEFGHWEIVPVBVHVI
+      $.
+
+    $( Group multiple (exponentiation) operation at a nonnegative integer
```
benjub, line 736078: It depends whether the law is written additively of multiplicatively. Probably "iterated group law". Then, you can add: "When the law is written multiplicatively, the iterated group law is also called exponentiation." (I realize these changes should happen at other places already in the main section. This can be done later.)

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+       of an element ` A ` of a monoid.  (Contributed by AV, 28-Dec-2023.) $)
+    cycsubmel $p |- ( X e. C <-> E. i e. NN0 X = ( i .x. A ) ) $=
+      ( wcel cv wceq cn0 wrex co ovex crn cfv eleq2i wfn wb fvelrnb ax-mp oveq1
+      fnmpti fvmpt eqeq1d eqcom syl6bb rexbiia 3bitri ) IDNIGUAZNZFOZGUBZIPZFQR
+      ZIURBESZPZFQRDUPIMUCGQUDUQVAUEAQAOZBESZGVDBETLUIFQIGUFUGUTVCFQURQNZUTVBIP
+      VCVFUSVBIAURVEVBQGVDURBEUHLURBETUJUKVBIULUMUNUO $.
+
+    $d A i $.  $d B i $.  $d .x. i $.
+    $( The set of nonnegative integer powers of an element ` A ` of a monoid
+       contains ` A ` .  (Contributed by AV, 28-Dec-2023.) $)
```
benjub, line 736107: Deduce this from the analogous statement for semigroups ? (later)

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+    cycsubmcl $p |- ( A e. B -> A e. C ) $=
+      ( vi wcel cv co wceq cn0 wrex c1 1nn0 wb oveq1 eqeq2d adantl mulg1 eqcomd
+      a1i rspcedvd cycsubmel sylibr ) BCMZBLNZBEOZPZLQRBDMUKUNBSBEOZPZLSQSQMUKT
+      UGULSPZUNUPUAUKUQUMUOBULSBEUBUCUDUKUOBCEGBHIUEUFUHABCDELFGBHIJKUIUJ $.
+
+    $d A a b j k $.  $d B a b j k x $.  $d C a b $.  $d F j k $.
+    $d G a b i j k x $.  $d .x. j k $.
+    $( The set of nonnegative integer powers of an element ` A ` of a monoid
+       forms a submonoid containing ` A ` ( see ~ cycsubmcl ), called the
+       cyclic monoid generated by the element ` A ` .  This corresponds to the
```
benjub, line 736117: Deduce this from the semigroup analogue? (later)

set.mm
```diff
@@ -735799,6 +735836,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+  Direct products (extension)
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+$)
+
+  ${
+    $d G x y $.  $d U x y $.
+    smndlsmidm.p $e |- .(+) = ( LSSum ` G ) $.
+    $( The direct product is idempotent for monoids.  (Contributed by AV,
+       27-Dec-2023.) $)
```
benjub, line 736149: Also answering your question in the PR: ` ( LSSum ` G ) ` is simply the group sum (or magma law) considered on subsets (in other words, it is the associated "direct image" operation). So I would simply say "the group sum (magma law) as an operation on subsets".

The present statement is not an idempotence property of that operation. I'd state it as ` A .(+) U = U ` when ` 0g e. A C_ U ` and I'd say "A submonoid absorbs any of its subsets containing the identity under the monoid law operation on subsets."

savask:
> The "inner ... direct product" (see comment of ~df-lsm) operator is called LSSum ...  and is often called "subgroup sum", which can easily be confused with "group sum". Maybe it should be called "internal direct sum" ... It should not be called "product", because we have already a definition for internal direct products (DProd, see ~df-dprd).

It took me some time and a comment of benjub to understand what `df-lsm` really means, I think the comment could indeed be something like "Define a product of subsets in a group". Calling it an inner direct product (or sum) is misleading, since `df-lsm` doesn't even seem to assume its operands to be subgroups.

In group theory one often uses the term "direct product" for all groups and "direct sum" specifically for abelian groups (although even in that case the "product" term can be used). I think set.mm should try to adhere to that convention.

set.mm
avekens: I think this section header was changed already some time ago: In 2020, it was "Ordered group sum operation" (see #1457), but the term "ordered" was (and is still) not mentioned in the comment for df-gsum. The title "Ordered sums in a magma" was also discussed already in this Issue. The term "ordered", however, seems to be too restrictive, because gsum is also defined for arbitrary (unordered) finite sets, see ~gsumval3 (this may make sense only if all involved elements commute). Therefore, I would like to keep the term "group sum operation" for the moment, not only being applicable for groups (like the "group operation" `+g` is also not used for groups only, but also for magmas, monoids, etc. Therefore, I just would change the section header to "Group sum operation in magmas".

For a complete renaming, we should have a broader discussion (within a separate Issue which I will open soon). 

I think the proposed term "composition" comes from Bourbaki (Algebra I, definition 4 on page 3) Here, also the term "ordered" is used: "composition ... of the ordered sequence ...". Unlike `gsum`, however, this definition cannot be used for arbitrary finite sets. 

According to Wikipedia, "Iterated binary operation" would be an appropriate name (see https://en.wikipedia.org/wiki/Iterated_binary_operation), this would be consistent if  `+g` would be called "binary operation" (see also https://en.wikipedia.org/wiki/Binary_operation) instead of "group operation" (however, the term "binary operation" is already used in set.mm for arbitrary functions ( S X S ) -> M, not only for functions ( S X S ) -> S). In https://planetmath.org/binaryoperation, the term "internal composition" is used for  `+g`, whereas in https://www.mathyma.com/mathsNotes/index.php?lng=en&trg=S1C2_Struct_IntBinOp it is called "internal binary operation".

avekens: Issue #3726 opened to continue this discussion.

set.mm
```diff
@@ -735176,6 +735179,40 @@ their group (addition) operations are equal for all pairs of elements of
...
+    lidridid.b $e |- B = ( Base ` G ) $.
+    lidridid.p $e |- .+ = ( +g ` G ) $.
+    lidridid.o $e |- .0. = ( 0g ` G ) $.
+    $( If a magma has a left and right identity element, both identity elements
+       are equal and are equal to "the" identity element.  Generalization of
+       statement in [Lang] p. 3: it is sufficient that "e" is a left identity
+       element and "e``" is a right identity element instead of both being
+       (two-sided) identity elements.  (Contributed by AV, 26-Dec-2023.) $)
+    lidridid $p |- ( ( L e. B /\ R e. B )
+                     -> ( A. x e. B ( ( L .+ x ) = x /\ ( x .+ R ) = x )
```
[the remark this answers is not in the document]
avekens: Because of ~r19.26, both versions are equivalent. I will, however, rewrite this theorem in deduction style, with to separated hypotheses.

avekens:
> > The "inner ... direct product" (see comment of ~df-lsm) operator is called LSSum ...  and is often called "subgroup sum", which can easily be confused with "group sum". Maybe it should be called "internal direct sum" ... It should not be called "product", because we have already a definition for internal direct products (DProd, see ~df-dprd).
> 
> It took me some time and a comment of benjub to understand what `df-lsm` really means, I think the comment could indeed be something like "Define a product of subsets in a group". Calling it an inner direct product (or sum) is misleading, since `df-lsm` doesn't even seem to assume its operands to be subgroups.
> 
> In group theory one often uses the term "direct product" for all groups and "direct sum" specifically for abelian groups (although even in that case the "product" term can be used). I think set.mm should try to adhere to that convention.

This discussion should be continued in Issue #3727 - for this PR, it is irrelevant.

set.mm
avekens: Why should "finite" be removed? It corresponds to the hypothesis `( ph -> F finSupp .0. )`, and I think that this hypothesis is necessary.

set.mm
avekens: I would like not to chane the comment until the wording for `gsum`is clarified. The term "Homomorphic property" is taken from the original theorem ~gsumccat. 

And yes, that the sequences are finite is essential here.

avekens: > (Recall that "words" are finite sequences, that is, functions on integer intervals `( 0 .. N )` for some `N e. NN0` .)" (I'm quoting the last parenthesis from memory, please check.)

No, words are functions on half-open integer intervals `( 0 ..^ N )`, so that empty words are possible. However, empty words must be excluded here (because we cannot assume the existence of an identity element).

set.mm
avekens: Here (for monoids) are no nonemptiness conditions, but the cases for empty words must be handles separately (as in the original proof).

set.mm
avekens: Yes, ~mulgnngsum is used in the proof of ~mulgnn0gsum, but it should be kept separate so that it can be used for magmas/semigroups not being monoids.

set.mm
avekens: "Group multiple (exponentiation)" is the term currently used in set.mm, so I will keep it here for the moment.

set.mm
avekens: "monoid" can be replaced by "magma" in the comment without changing the theorem itself - it already does not depend on `G` being a monoid. The same holds for ~cycsubmel. I will change the comments accordingly.

set.mm
avekens: If we had definitions for submagmas (only within my mathbox) and subsemigroups, we could have corresponding theorems for magmas and semigroups which could be used for the proof. Without these definitions, such theorems would be too clumsy.

set.mm
benjub: I should have explained indeed. We restrict the sum to "the support of ..." which is a notion defined earlier. That support happens to be finite (and it is a necessary hypothesis indeed). "Having finite support" is a property which is also defined earlier.  However, there is no such thing as "the finite support of" a family.

Also, I missed that you wrote "support of the index set" but, it's actually the support of the family.

So, the comment could be phrased:
> The group sum of a family can be restricted to the support of that family without changing its value, provided that that support is finite."

(no typo, "that" appears twice).

Could you prove the same statement in a unital magma ?

set.mm
avekens: ~smndlsmidm is (and was intended to be) a generalization of the already existing ~lsmidm. I think for the given restrictions the operation is idempotent. You are suggesting a broader generalization which could be proven in a separate theorem.

set.mm
benjub: Ok since it is in your mathbox... But if you want to keep "homomorphic", I'm going to ask you which morphism you're talking about ;) I admit that you can say that the composition law is a morphism when it is associative, but this is a bit contrived.

set.mm
benjub: That's why I wrote "removing the nonemptiness condition" !

set.mm
benjub: You're right... I was misled by the word "group" appearing twice in the comment ! (We should fix that soon.)

set.mm
avekens: > Could you prove the same statement in a unital magma ?
I thought about this, too. But I think a lot of preparatory work is necessary for this.

set.mm
benjub: Thanks for opening the issue. So you'll keep that section title unchanged for now ? (I think "group sum operation for magmas" sounds bizarre.)

benjub:
> Maybe some of the following theorems can be moved to main immediately (candidates marked with (*))

Probably yes, most/all of the starred theorems should be moved to Main, but probably better in a separate PR. These are nice additions (which is why I am a bit picky in my above comments). I'll try to do a second review of them after this PR is merged and before you do the PR to move them to Main, and maybe other people too may prefer to see the new theorem webpages first.

set.mm
```diff
@@ -214980,7 +214990,7 @@ is an important property of monoids (see ~ mndid ), and therefore also for
 
 $(
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
-  Ordered sums in a magma
+  Group sum operation in magmas
```
benjub, line 214993: In the meantime, can you keep the existing titles ("Ordered sums in a magma") ?  (group sum in magmas sounds really weird)

set.mm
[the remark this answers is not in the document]
benjub: I know it's equivalent, but it's better that the formal statement be as close as possible to the informal statement. Your new statement is better.

set.mm
```diff
@@ -735822,6 +735867,340 @@ their group (addition) operations are equal for all pairs of elements of
...
+    cycsubmel $p |- ( X e. C <-> E. i e. NN0 X = ( i .x. A ) ) $=
+      ( wcel cv wceq cn0 wrex co ovex crn cfv eleq2i wfn wb fvelrnb ax-mp oveq1
+      fnmpti fvmpt eqeq1d eqcom syl6bb rexbiia 3bitri ) IDNIGUAZNZFOZGUBZIPZFQR
+      ZIURBESZPZFQRDUPIMUCGQUDUQVAUEAQAOZBESZGVDBETLUIFQIGUFUGUTVCFQURQNZUTVBIP
+      VCVFUSVBIAURVEVBQGVDURBEUHLURBETUJUKVBIULUMUNUO $.
+
+    $d A i $.  $d B i $.  $d .x. i $.
+    $( The set of nonnegative integer powers of an element ` A ` of a magma
+       contains ` A ` .  (Contributed by AV, 28-Dec-2023.) $)
+    cycsubmcl $p |- ( A e. B -> A e. C ) $=
```
benjub, line 736139: Sorry I didn't cacth this earlier, but in both theorems, you should write "unital magma". Indeed, the formal statement is still true without assuming it, because df-gsum has to give a value for the empty sum even when the structure is not unital, but this does not really make sense. You may also write some explanation like that in your theorems, to at least warn the reader.

set.mm
benjub: Oh, I understand what you meant: "A submonoid/subgroup is idempotent for the group-like operation on subsets." (using my proposed terminology).

More generally: the equation $p + p = p$ expresses that " $p$ is idempotent for $+$".  But when you say "that operation is idempotent", it means "that operation applied twice is like applied once", or "a second application of that operation has no new effect", as in $f(f(x)) = f(x)$ for any $x$.

The two notions are related: adding an idempotent on the right is an idempotent operation. Indeed, if you define the function $f \colon G \to G, x \mapsto x + p$ where $p$ is idempotent for $+$, then $f$ is idempotent, that is, $f^2 = f$.

set.mm
benjub: Resolving this conversation as subsumed by https://github.com/metamath/set.mm/pull/3724#discussion_r1438717280

set.mm
avekens: Since neither "ordered sum" nor "iterated sum" is defined yet, I would propose the titles "Iterated sums in a magma" and "Iterated sums in a monoid". This would be a first step into the direction of the new naming. 

And I would slightly enhance the comment of df-gsum accordingly: 

> Define the group sum **(also called "iterated sum")** for the structure ` G ` of a finite sequence

I want to keep the term "sum" for the moment because of the label/symbol `gsum`.

@benjub do you agree?

set.mm
avekens: OK, I will revise the comment accordingly.

set.mm
avekens: According to Wikipedia [Idempotence](https://en.wikipedia.org/wiki/Idempotence):

> The binary operation ⋅ is said to be idempotent if x ⋅ x = x for all x ∈ S.


So we can say "The direct product is idempotent for submonoids". Iwill adjust the comments accordingly.

set.mm
benjub: I agree for this PR. For a long term solution, there is #3726.

set.mm
benjub: Yes, since a binary operation is called idempotent when all its elements are idempotent, you can express "submonoids are idempotent for +" as "+ restricted to submonoids is idempotent". It's not false, but sounds a bit contrived. Your mathbox, so as you prefer.

benjub (approved):

PR #3725, opened 2023-12-29
Shorten ncanth and edit comment.

benjub:


set.mm
```diff
@@ -34332,7 +34332,9 @@ something like (wi (wceq (cv vx) (cv vy)) wph) ) into just (wcdeq vx vy
        "collection of all sets which are not members of themselves" is a set.
        However, here we prove ` { x | x e/ x } e/ _V ` .  This contradiction
        was discovered by Russell in 1901 (published in 1903), invalidating the
-       Comprehension Axiom and leading to the collapse of Frege's system.
+       Comprehension Axiom and leading to the collapse of Frege's system, which
+       Frege acknowledged in the second edition of his _Grundgesetze der
+       Arithmetik_.
```
benjub, line 34337: We don't want to be an encyclopedia in the history of mathematics, but I think that fact is important enough to be mentioned (and is an example of intellectual integrity from Frege).

set.mm
```diff
@@ -65959,14 +65966,20 @@ complete lattice has a (least) fixpoint.  Here we specialize this
...
-     that are "too large."  This theorem gives some intuition behind that
-     failure: in NF the universal class is a set, and it equals its own power
-     set.  (Contributed by NM, 29-Jun-2004.) $)
+     universe onto its power class.  Compare ~ canth that works for sets.
+
+     This failure comes from a limitation of the collection principle (which is
+     necessary to avoid Russell's paradox ~ ru ): ` P _V ` , being a class,
+     cannot contain proper classes, so ` ~P _V ` , so it is no larger than
+     ` _V ` , which is why ` _I ` "succeeds" in being surjective onto ` ~P _V `
+     (see ~ pwv ).  See also the remark in ~ ru about NF, in which Cantor's
```
benjub, line 65975: I think this (and the similar remark in ~ pwv) are necessary to avoid possible misunderstandings from a new reader, and to hint at the root cause of the "problem", namely, the necessary restriction of the collection principle to avoid Russell's paradox in Zermelo-style set theories.

tirix (approved):

Issue #3687, opened 2023-12-11
CCHOICE and countably infinite versus countable


jkingdon:
Most literature references that I've seen regarding countable choice either (a) assume excluded middle, for example ZF, and thus are not directly applicable to iset.mm, or (b) just express countable choice in terms of a family of inhabited sets which is countably infinite (that is, equinumerous to ω) and don't mention at all whether it also applies to a family which is countable (that is, ∃𝑓 𝑓:ω–onto→(𝐴 ⊔ 1o)) as well.

Right now, iset.mm has a definition of `CCHOICE` which applies to the countably infinite case - https://us.metamath.org/ileuni/ccfunen.html

The question is whether `CCHOICE` also implies a similar theorem for countable collections, for example:

```
    ccfun.cc $e |- ( ph -> CCHOICE ) $.
    ccfun.a $e |- ( ph -> E. g g : _om -onto-> ( A |_| 1o ) ) $.
    ccfun.m $e |- ( ph -> A. x e. A E. w w e. x ) $.
    ccfun $p |- ( ph -> E. f ( f Fn A /\ A. x e. A ( f ` x ) e. x ) ) $=
      ? $.
```

Although this might seem like a simple matter of finding a countably infinite set which contains the given one and applying `ccfun` to that, my attempts to do so weren't working in ways that have me a bit worried.

My first approach was to wonder whether `( A |_| _om )` is countably infinite, but I didn't see a proof for that.

My second approach was to construct:

```
    ccfunlem.f $e |- ( ph -> F : _om -onto-> ( A |_| 1o ) ) $.
    ccfunlem.g $e |- G = ( n e. _om |-> ( { n } X. ( K ` ( F ` n ) ) ) ) $.
    ccfunlem.k $e |- K = case ( ( _I |` A ) , ( 1o X. { { (/) } } ) ) $.
```

Now `ran G` is a countably infinite collection of inhabited sets, so one can apply `ccfun` to get a choice function on it.

But we are looking for a choice function on ` A `, and given an element of ` A `, there might be multiple natural numbers which ` F ` maps to that element and no way to choose one of them that I noticed.

A construction similar to the one in https://us.metamath.org/ileuni/ennnfone.html (in which we build up our answer one element at a time, adding a new element only when necessary) seems like the sort of thing we'd need, but that relies on decidable equality.

[...]

Issue #3618, opened 2023-11-06
The union of countably many countable sets is countable (using countable choice)


jkingdon:
The statement would be similar to https://us.metamath.org/ileuni/ctiunct.html but would instead be:

```
name-me.a $e |- ( ph -> E. f f : _om -onto-> ( A |_| 1o ) ) $.
name-me.b $e |- ( ( ph /\ x e. A )
  -> E. g g : _om -onto-> ( B |_| 1o ) ) $.
name-me.ch $e |- ( ph -> CCHOICE ) $.
name-me $p |- ( ph -> E. h h : _om -onto-> ( U_ x e. A B |_| 1o ) ) $=
```

This represents an intuitionization of https://us.metamath.org/mpeuni/unictb.html but as in other parts of iset.mm we express countability differently than set.mm.

Depends on #3617

jkingdon:
To be honest I'm not completely sure whether

(1) https://us.metamath.org/ileuni/omiunct.html is roughly the best we can get (where `A` is `_om` rather than any countable set), or

(2) there is some way to prove the theorem at the top of this issue, using `f` as a way of letting `_om` index `A` as described at https://github.com/metamath/set.mm/issues/3687#issuecomment-1868186916 . Perhaps the proof would be able to use https://us.metamath.org/ileuni/cc3.html or https://us.metamath.org/ileuni/cc4.html .

digama0:
(2) is the case (it is possible to prove the OP theorem):

Take `f : _om -onto-> ( A |_| 1o )`, and consider the set `S(n) := { g | g : _om -onto-> ( K(f(n)) |_| 1o ) }`, where `K(inl a) = B(a)` and `K(inr ()) = (/)`. Note that for all `n e. _om`, `S(n)` is nonempty. Use countable choice to define `g(n) e. S(n)`. Now `h(n) := g(fst n)(snd n)` defines a function to `( U_ x e. A B |_| 1o )` which is onto because given `y e. B(x)` where `inl x = f(i)`, `g(i) : _om -onto-> ( B(x) |_| 1o )` so `inl y = g(i)(j)` and thus `inl y = h(i,j)`. In the inr case, we can take `inr () = g(0)(j)` and conclude `inr () = h(0,j)`.

[...]

PR #3728, opened 2023-12-30
a few countable set results for iset.mm

jkingdon:
I haven't made much progress on anything related to #3687 , #3618 , etc, but I thought I should send in a few small results related to countable sets in iset.mm.

* ω is countable
* A countable class is a set
* Countability is invariant relative to equinumerosity


icecream17 (approved):

iset.mm
```diff
@@ -68631,6 +68632,25 @@ property of disjoint unions (see ~ updjud in the case of functions).
...
+      ( vx com c1o cdju cv wfo wex cid cres wf1o f1oi f1ofo cvv wcel omex ax-mp
+      resiexg mp2b c0 foeq1 spcev wb peano1 elex2 ctm mpbir ) CCDEAFZGAHZCCUHGZ
+      AHZCCICJZKCCULGZUKCLCCULMUJUMAULCNOULNOPCNRQCCUHULUAUBSTCOBFCOBHUIUKUCUDB
+      TCUEBCAUFSUG $.
+  $}
+
+  ${
+    $d A f $.
+    $( A countable class is a set.  (Contributed by Jim Kingdon,
+       25-Dec-2023.) $)
```
benjub, line 68648: Maybe a back-link to ctex ?

benjub (approved):

avekens (approved):

iset.mm
jkingdon: I've been back and forth on this a little bit, both before and after your comment.

The case against a link to ctex is that `A ~<_ _om` is a dead end constructively and there's no point in drawing attention to it when people have landed on `E. f f : _om -onto-> ( A |_| 1o )` .

The case for is, well, why not it is just a link.

I guess for now I'll merge without the link but we can always add it if we decide we want it later.

PR #3731, opened 2024-01-01
Remove same-alphabet requirement from eqwrd

jamesjer:
I will post to the mailing list about this PR, but the idea is to not constrain U and W to be over the same alphabet.  The proof does not require that.  Making this change enables a large reduction in size of the proof of pfx2.

I have not added any "Proof shortened by" comments because I didn't really change the essential structure of any proofs (other than cutting off a branch of the pfx2 proof).  I just rearranged proof steps to give the minimizer more opportunities to find valid reductions in proof size.


icecream17 (approved):
eqwrd is more general and proofs are overall shorter

jkingdon:
> I will post to the mailing list about this PR, but the idea is to not constrain U and W to be over the same alphabet. The proof does not require that.

Makes sense to me. I could easily imagine cases where U is a subset of W (for example).

> I have not added any "Proof shortened by" comments

I don't think there is any need for them in this sort of situation.

jkingdon (approved):

avekens (approved):
The proof of ~wwlkseq becomes a little bit longer (it corresponds exactly to the former ~eqwrd), but this does not matter.

PR #3735, opened 2024-01-01
fsplit

benjub:
As commit messages say.

### DIFF
--- a/scripts/min.cmd
+++ b/scripts/min.cmd
@@ -37,7 +37,7 @@ match 1.tmp 'MM>' n
 match 1.tmp '  ' y
 break 1.tmp ''
 unduplicate 1.tmp
-add 1.tmp 'prove ' '$min xxx/allow */no_new ax-*/may$min *$save new_proof/compressed$_exit_pa$'
+add 1.tmp 'prove ' '$min xxx/allow */no_new ax-*$save new_proof/compressed$_exit_pa$'
 ! If you want to do a "dry run" without saving proofs, change above line
 ! to:
 ! add 1.tmp 'prove ' '$min xxx/no_new ax-*$_exit_pa/force$'
--- a/scripts/min.cmd
+++ b/scripts/min.cmd
@@ -37,7 +37,7 @@ match 1.tmp 'MM>' n
 match 1.tmp '  ' y
 break 1.tmp ''
 unduplicate 1.tmp
-add 1.tmp 'prove ' '$min xxx/allow */no_new ax-*$save new_proof/compressed$_exit_pa$'
+add 1.tmp 'prove ' '$min xxx/allow */no_new ax-*/may$min *$save new_proof/compressed$_exit_pa$'
 ! If you want to do a "dry run" without saving proofs, change above line
 ! to:
 ! add 1.tmp 'prove ' '$min xxx/no_new ax-*$_exit_pa/force$'
--- a/discouraged
+++ b/discouraged
@@ -4730,7 +4730,7 @@
 "dfhnorm2" is used by "normval".
 "dfich2ai" is used by "dfich2OLD".
 "dfich2bi" is used by "dfich2OLD".
-"dfid2" is used by "fsplit".
+"dfid2" is used by "fsplitOLD".
 "dfid3" is used by "dfid2".
 "dfiop2" is used by "hoico1".
 "dfiop2" is used by "hoico2".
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -74348,13 +74347,34 @@ composition with a (2-place) operation to build compound operations such
...
+      anbi1i id wrel relcnv dfrel4v mpbi mptv 3eqtr4i ) AEZBEZFGUAZUBZHZABIZWBW
+      AWAJZKZABIWDALWGUJWEWHABWEWBWAWCHWBGMZWBWAFHZNZWHWAWBWCARZBRZUCGWBWAFWLUD
+      WKCEZWAKZWBWNWNJZKZNZCSZWHWBFUEZWAKZWQNZCSXAWQCSZNWSWKXAWQCUFXBWRCWQXAWOW
+      QWTWNWAWNWNWBCRZXDUGUHUIOXAWJXCWIFLUKZWBLMXAWJULLLFUMXEUNLLFUOUPWMLWBWAFU
+      QURWIWBWNDEZKZCDIZMWBWNXFJZKZXGNZDSZCSXCGXHWBCDUSUTXGCDWBVAXLWQCXLXFWNKZW
+      QNZDSWQXKXNDXKXGXJNXMXJNXNXJXGVCXGXMXJCDVBVMXMXJWQXMXIWPWBXFWNWNVDPVEQOWQ
+      WQDCXMWQVFTVGOVHVIVJWQWHCAWOWPWGWBWOWNWAWNWAWOVNZXOVKPTVGQVLWDVOWDWFKWCVP
+      ABWDVQVRABWGVSVT $.
+  $}
+
```
avekens, line 74371: Neither ~fpar nor ~fsplit are currently used in proofs. So what are they good for? The given examples are also not very enlightening.

scripts/min.cmd
```diff
@@ -37,7 +37,7 @@ match 1.tmp 'MM>' n
 match 1.tmp '  ' y
 break 1.tmp ''
 unduplicate 1.tmp
-add 1.tmp 'prove ' '$min xxx/allow */no_new ax-*$save new_proof/compressed$_exit_pa$'
+add 1.tmp 'prove ' '$min xxx/allow */no_new ax-*/may$min *$save new_proof/compressed$_exit_pa$'
```
avekens, line 40: What does this enhancement mean?

icecream17 (approved):
fpar and fsplit provide equivalences between two classes: one in "non-setvar notation" and the other in mpt/mpo notation. The only use I can see for this is to show alternate definitions in the sense of what [df-in](https://us.metamath.org/mpeuni/df-in.html) says about dfin2 and dfin4.

Regardless of usefulness the changes are improvements and bj-bixor is mathbox.

The example in fpar says,
F = sqrt & G = abs => ``(x (+ o. H) y) = (sqrt`x) + (abs`y)``

The example in fsplit says, (in combination with fpar)
F = sqrt & G = abs => ``(((+ o. H) o. ◡(1st ↾ I )) ` x) = (sqrt`x) + (abs`x)``

scripts/min.cmd
benjub: Sorry, this was meant to be temporary and not added to the PR. I'm going to remove it.

tirix:
As a remark, it might also be possible to get the same result using Mario's function operation map `oF`:
``( sqrt oF + abs ) = ( x e. CC |-> ( ( sqrt ` x ) + ( abs ` x ) ) )``

set.mm
benjub: I haven't really considered applications. When looking at dfid2, I saw it had one discouraged use, so I just set on to remove it.

set.mm
```diff
@@ -65983,9 +65981,9 @@ complete lattice has a (least) fixpoint.  Here we specialize this
      universe onto its power class.  Compare ~ canth that works for sets.
 
      This failure comes from a limitation of the collection principle (which is
-     necessary to avoid Russell's paradox ~ ru ): ` P _V ` , being a class,
-     cannot contain proper classes, so ` ~P _V ` , so it is no larger than
-     ` _V ` , which is why ` _I ` "succeeds" in being surjective onto ` ~P _V `
+     necessary to avoid Russell's paradox ~ ru ): ` ~ P _V ` , being a class,
```
tirix, line 65984: Why split the `~` from the `P` here?
I suppose it should be `` ` ~P _V ` ``.

benjub:
> As a remark, it might also be possible to get the same result using Mario's function operation map `oF`: `` ( sqrt oF + abs ) = ( x e. CC |-> ( ( sqrt ` x ) + ( abs ` x ) ) ) ``

Actually, it's rather https://us.metamath.org/mpeuni/df-of.html (the pointwise operation) which could be defined in terms of the diagonal embedding $x \mapsto (x, x)$, that is, if I call that embedding $\delta$, then
$(x \mapsto f(x) R g(x) ) = ( f (o_f R) g ) \circ \delta$.

set.mm
benjub: The main reason to do that is because it's 1:30am !

benjub: Fixed.

scripts/min.cmd
benjub: Done.

tirix (approved):

PR #3737, opened 2024-01-01
mathbox: add bj-elpwg

benjub:


set.mm
```diff
@@ -752285,6 +752294,7 @@ with simplest antecedents (i.e., in the corresponding ordering of binary
 
 $( (End of Adhemar's mathbox.) $)
```
avekens, line 752295: I think  Adhemar's and  Igor Ieskov's mathboxes are misplaced: they should be moved up so that the mathboxes are alphabetically sorted.

benjub: You're right. Can we do that in another PR ?

avekens: > You're right. Can we do that in another PR ?

yes, of course.

icecream17 (approved):

PR #3739, opened 2024-01-03
Reorganize Mathbox Algebra

tirix:
This simply re-organizes the "Algebra" parts of my mathbox.
- theorems about centralizers and centers are moved ahead;
- theorems about the finitely supported group sums (or iterated compositions, if renamed) are moved ahead;
- theorems about the permutation group are moved ahead and grouped together.

No new theorem is introduced, no proof is changed.


tirix:
As I guessed, GitHub is not very clever about displaying the changes and it all looks like a big mess, this is why this PR is only about moving things around - it would have been a nightmare to review otherwise.

To check the differences in this commit, I recommend checking out the branch locally, and running:
```
git diff --patience 1c589b88 -- set.mm
```
The `--patience` switch generates a better diff where you can actually see the blocks moved around.

avekens (approved):
Scrolling quickly through the changes, I see no problem with them.
But I detected some interesting theorems, on which I will have a close look after this PR is merged.

wlammen (approved):
I simply compared the github diff to the current version of set.mm in a text editor.  All the moved theorems are exact copies of those in my editor.  What strikes me weird a bit is that the number of lines increased by 2.

I did not check whether the new order matches the titles.

tirix:
> What strikes me weird a bit is that the number of lines increased by 2.

That might be because of the different chapter organization (I mean, newlines and headings), but I also could not pinpoint where this is coming from.

PR #3740, opened 2024-01-03
minor adjustments for Graph Theory

avekens:
* ~ rspc2vd moved up
* ~df-frgr and ~isfrgr revised (~frgrusgrfrcond became obsolete)
* symbols "RegGraph" and "RegUsGraph" surrounded by spaces (are used in "operation style")


icecream17 (approved):

set.mm
```diff
@@ -418128,45 +418146,33 @@ edge remains odd if it was odd before (regarding the subgraphs induced
        _friendship graph_ if every pair of its vertices has exactly one common
        neighbor.  This condition is called the _friendship condition_ , see
        definition in [MertziosUnger] p. 152.  (Contributed by Alexander van der
-       Vekens and Mario Carneiro, 2-Oct-2017.)  (Revised by AV,
-       29-Mar-2021.) $)
-    df-frgr $a |- FriendGraph = { g | ( g e. USGraph
-                      /\ [. ( Vtx ` g ) / v ]. [. ( Edg ` g ) / e ].
-                         A. k e. v A. l e. ( v \ { k } )
-                            E! x e. v { { x , k } , { x , l } } C_ e ) } $.
+       Vekens and Mario Carneiro, 2-Oct-2017.)  (Revised by AV, 29-Mar-2021.)
```
jkingdon, line 418149: Not a blocker (indeed, I won't refrain from merging), but the conventions say there should normally be only one Revised by line, and in a case like this where both of them have the same author, seems to me like there is a strong case for just updating the date and not keeping the old `Revised by`.

jkingdon (approved):

set.mm
avekens: > Not a blocker (indeed, I won't refrain from merging), but the conventions say there should normally be only one Revised by line, and in a case like this where both of them have the same author, seems to me like there is a strong case for just updating the date and not keeping the old `Revised by`.

@jkingdon where is this convention documented? I cannot find it neither in set.mm (section "Conventions") nor in the Metamath book (section 4.4.1, paragraph "Parentheticals"). I can remember a discussion about this, but the result was, as far as I can remember, that there usually should be only one revision tag, but to have more than one is not forbidden.

jkingdon: > > Not a blocker (indeed, I won't refrain from merging), but the conventions say there should normally be only one Revised by line, and in a case like this where both of them have the same author, seems to me like there is a strong case for just updating the date and not keeping the old `Revised by`.
> 
> @jkingdon where is this convention documented? I cannot find it neither in set.mm (section "Conventions") nor in the Metamath book (section 4.4.1, paragraph "Parentheticals"). I can remember a discussion about this, but the result was, as far as I can remember, that there usually should be only one revision tag, but to have more than one is not forbidden.

Section "Usage of parentheticals" at https://us.metamath.org/mpeuni/conventions-comments.html and you are correct that this is not a hard requirement (per your use of the word "usually" and my use of the word "normally").

PR #3741, opened 2024-01-03
Move theorems about simple groups into main

savask:
There are three commits.

1. Added a theorem to my mathbox that a simple group either has trivial center, or it's abelian. This is definitely not the most groundbreaking result, but it served as a little testing ground for the metamath-lamp proof assistant (it turned out to be better than I thought).
2. In #3498 it was suggested that the section on simple groups from @CatsAreFluffy's mathbox should be moved into main. The formal criteria are met, since a theorem from my mathbox and a theorem by @tirix (which I moved into his mathbox) use some results of Rohan.
I put the "Simple groups" section after the "Abelian groups" section. In my opinion both of these should rather be subsections in "10.2 Groups", but I want to know what others think. I also had to move some helper theorems about finite sets etc into main. I think some of those should be renamed (the ones starting with the `rr-` prefix), but I didn't attempt that.
3. At the end I realized that my theorem relies on some facts from Thierry's mathbox, so I moved those into main in the third commit.


tirix:
> 2\. I think some of those should be renamed (the ones starting with the `rr-` prefix), but I didn't attempt that.

Yes! I suppose this should not be harder than doing a global replacement for the theorem names, and then rewrapping:
```
scripts/rewrap
```
is ideal if you have access to a shell, otherwise it's just:
```
metamath "read set.mm" "write source set.mm /rewrap" \
  'erase' "read set.mm" 'save proof */compressed/fast' \
  "write source set.mm" quit
```

set.mm
```diff
@@ -237805,6 +238091,472 @@ factorization into prime power factors (even if the exponents are
...
+  $c SimpGrp $.
+
+  $( Extend class notation with the class of simple groups. $)
+  csimpg $a class SimpGrp $.
+
+  $( Define class of all simple groups.  A simple group is a group ( ~ df-grp )
+     with exactly two normal subgroups.  These are always the subgroup of all
+     elements and the subgroup containing only the identity ( ~ simpgnsgbid ).
+     (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
+  df-simpg $a |- SimpGrp = { g e. Grp | ( NrmSGrp ` g ) ~~ 2o } $.
```
tirix, line 238109: Maybe I'm a bit pedantic here, but since this moves stuff to main, I'll have a go.

I think it is worth deciding what is our preferred way to write "there's exactly a pair".
Searching through `set.mm`, I see 44 occurrences of ` ~~ 2o`, but 119 occurrences of ``( # ` ... ) = 2``.
To me, the latter is more natural. Rohan chose the other way and there would be a handful of theorems to change if we were to decide to change.

tirix:
Overall : thanks for doing the move!
No problem on my side for moving the one theorem back to my mathbox, of course.

CatsAreFluffy:
`rr-phpd` is unused and doesn’t need to be moved to the mainspace.

set.mm
CatsAreFluffy: I used `~~ 2o` since that’s how ~df-prm does it.

tirix: Good point, and it's important to keep the comparison with ~df-prm.
Also, `~~ 2o` is actually shorter than the other version, and it is more elementary.

But on the other hand, I believe that, compared to equinumerosity and ordinal numbers, the notion of the cardinality of a set, and the number 2 are better known and more accessible.

Anyway this is probably better discussed in a different forum. Maybe I'll open a generic issue for that instead.

set.mm
```diff
@@ -32717,6 +32724,16 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
       ( wrex rspcedv mpd ) ACBDFJIABCDEFGHKL $.
   $}
 
+  ${
+    $d x ph $.  $d x B $.  $d x A $.
+    rr-rspce.1 $e |- ( ( ph /\ x = A ) -> ps ) $.
+    rr-rspce.2 $e |- ( ph -> A e. B ) $.
+    $( Prove a restricted existential.  (Contributed by Rohan Ridenour,
+       3-Aug-2023.) $)
+    rr-rspce $p |- ( ph -> E. x e. B ps ) $=
```
avekens, line 32733: the labels starting with rr must be revised. the prefix rr are the initials of Rohan Ridenour, and were OK in his mathbox. In main, however, we do not have such prefixed labels.

avekens (changes requested):
the labels in main starting with rr-must be revised.

set.mm
savask: I removed the `rr-` prefix in the latest commit, and moved `rr-phpd` back into RR's mathbox. I hope labels I chose for some of the `rr-` theorems are okay.

set.mm
avekens: Yes, we should discuss this further in a separate Issue. I also had/have the need to handle sets with exactly 2 elements - therefore I defined a new symbol `PrPairs` for them in my mathbox.

set.mm
```diff
@@ -53537,6 +53554,29 @@ of the relation (see ~ op1stb ).  (Contributed by NM, 28-Jul-2004.) $)
...
 
+  ${
+    $d D x $.  $d x A $.  $d x C $.  $d x ph $.
+    elrnmptspd.1 $e |- F = ( x e. A |-> B ) $.
+    elrnmptspd.2 $e |- ( ph -> C e. A ) $.
+    elrnmptspd.3 $e |- ( ph -> D e. V ) $.
+    elrnmptspd.4 $e |- ( ( ph /\ x = C ) -> D = B ) $.
+    $( Elementhood in an image set.  (Contributed by Rohan Ridenour,
+       3-Aug-2023.) $)
+    elrnmptspd $p |- ( ph -> D e. ran F ) $=
```
avekens, line 53565: This theorem should be called ~elrnmptd, but this label is already used in GS's mathbox. So either the theorem in GS's mathbox should be changed, or we change ~elrnmptspd to ~elrnmptdv (according to our conventions: "Additionally, we sometimes suffix with "v" the label of a theorem adding a disjoint variable condition").

Furthermore, the term "image set" is not specified appropriately (although it is used often): According to ~imadmrn, "**the** image set" of a function means "the image of the domain" of a function, and is synonymous with "range". But `ran F` is not always a set (it can be a proper class), therefore "**the** image" of a function would be a better synonym for "range".

set.mm
avekens: I am not happy with this label: this is not the deduction form of ~rspce, and it is not in deduction style at all (although it looks like). Looking at ~rspcimedv, this seems to be the deduction form of the theorem above, so it may be called ~rspcime.

set.mm
avekens: I agree. But looking at the comment of ~ elrnmpt, the comment should be "Elementhood in the range of a function in maps-to notation, deduction form."

set.mm
```diff
@@ -53537,6 +53554,29 @@ of the relation (see ~ op1stb ).  (Contributed by NM, 28-Jul-2004.) $)
...
+      ( crn wcel wceq wrex rspced wb elrnmpt syl mpbird ) AFGMNZFDOZBCPZAUCBECL
+      JQAFHNUBUDRKBCDFGHISTUA $.
+  $}
+
+  ${
+    $d x C $.
+    elrnmpt2d.1 $e |- F = ( x e. A |-> B ) $.
+    elrnmpt2d.2 $e |- ( ph -> C e. ran F ) $.
+    $( Elementhood in an image set.  (Contributed by Rohan Ridenour,
+       3-Aug-2023.) $)
```
avekens, line 53575: Comment to be revised, see above.

set.mm
```diff
@@ -86206,6 +86246,19 @@ excluded middle (LEM), and in ILE the LEM is not accepted as necessarily
   php5 $p |- ( A e. _om -> -. A ~~ suc A ) $=
     ( com wcel csuc csdm wbr cen wn php4 sdomnen syl ) ABCAADZEFALGFHAIALJK $.
 
+  ${
+    php2d.1 $e |- ( ph -> A e. Fin ) $.
+    php2d.2 $e |- ( ph -> B C_ A ) $.
+    php2d.3 $e |- ( ph -> A ~~ B ) $.
+    $( Strengthening of ~ php expressed without negation.  (Contributed by
+       Rohan Ridenour, 3-Aug-2023.) $)
+    php2d $p |- ( ph -> A = B ) $=
```
avekens, line 86255: This is not the deduction form of ~php2, so it should be renamed. Maybe ~phpeqd. I also would change the comment to "Corollary of Pigeonhole Principle using equality. Strengthening of ~ php expressed without negation."

set.mm
```diff
@@ -180023,6 +180111,30 @@ obviously not a bijection (by Cantor's theorem ~ canth2 ), and in fact
     ( cz wcel c1 cgcd co wceq 1z gcdcom mpan gcd1 eqtrd ) ABCZDAEFZADEFZDDBCMNO
     GHDAIJAKL $.
 
+  ${
+    gcdmuld.1 $e |- ( ph -> M e. NN0 ) $.
+    gcdmuld.2 $e |- ( ph -> N e. ZZ ) $.
+    $( The greatest common divisor of a nonnegative integer and a multiple of
+       it is itself.  (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
+    gcdmuld $p |- ( ph -> ( M gcd ( N x. M ) ) = M ) $=
```
avekens, line 180119: This is a generalzation of ~gcdmultiple and ~gcdmultiplez. Since it is already placed before ~gcdmultiple, it should be renamed ~gcdmultipled, and the proofs of ~gcdmultiple and ~gcdmultiplez can be shortened accordingly.

avekens: Comment should be "The greatest common divisor of a nonnegative integer and a multiple of it is the nonnegative integer itself."

set.mm
```diff
@@ -180023,6 +180111,30 @@ obviously not a bijection (by Cantor's theorem ~ canth2 ), and in fact
...
+      CKLBKLFKLUFUIMEABDNZAOCBFPQABSLUFBMDBRTAUHUGBGAUGAUGACBEUJUAUBUCUDUE $.
+  $}
+
+  ${
+    gcddvdsd.1 $e |- ( ph -> M e. NN ) $.
+    gcddvdsd.2 $e |- ( ph -> N e. ZZ ) $.
+    gdcdvdsd.3 $e |- ( ph -> M || N ) $.
+    $( The greatest common divisor of a positive integer and another integer it
+       divides is itself.  (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
+    gcddvdsd $p |- ( ph -> ( M gcd N ) = M ) $=
```
avekens, line 180131: not deduction form of ~gcddvds. Maybe ~dvdsgcdidd (analogous to ~nndvdslegcd ).

set.mm
```diff
@@ -215793,6 +215905,28 @@ everywhere defined internal operation (see ~ mndcl ), whose operation is
       PMN $.
   $}
 
+  ${
+    hashelne0d.1 $e |- ( ph -> B e. A ) $.
+    hashelne0d.2 $e |- ( ph -> A e. V ) $.
+    $( A set with an element has nonzero size.  (Contributed by Rohan Ridenour,
+       3-Aug-2023.) $)
+    hashelne0d $p |- ( ph -> -. ( # ` A ) = 0 ) $=
```
avekens, line 215913: Theorem should be placed near after ~hasheq0.

avekens (changes requested):
Some changes are required for the common theorems moved from RR's mathbox. Suggestions see my inline comments (not yet finished). Maybe someone else should approve these suggestions, or provide alternatives.

set.mm
```diff
@@ -220815,6 +220971,33 @@ by a normal subgroup (resp. two-sided ideal).  (Contributed by Mario
...
+
+  ${
+    $d x B $.  $d x G $.  $d x ph $.  $d x .0. $.
+    trivsubgd2.1 $e |- B = ( Base ` G ) $.
+    trivsubgd2.2 $e |- .0. = ( 0g ` G ) $.
+    trivsubgd2.3 $e |- ( ph -> G e. Grp ) $.
+    trivsubgd2.4 $e |- ( ph -> B = { .0. } ) $.
+    $( The only subgroup of a trivial group is itself.  (Contributed by Rohan
+       Ridenour, 3-Aug-2023.) $)
+    trivsubgd2 $p |- ( ph -> ( SubGrp ` G ) = { B } ) $=
```
avekens, line 220995: maybe ~trivsubgsnd

set.mm
```diff
@@ -226317,6 +226550,16 @@ operation is a permutation group (group consisting of permutations), see
       L $.
   $}
 
+  ${
+    odcld.1 $e |- B = ( Base ` G ) $.
+    odcld.2 $e |- O = ( od ` G ) $.
+    odcld.3 $e |- ( ph -> A e. B ) $.
+    $( Deduction associated with ~ odcl .  (Contributed by Rohan Ridenour,
```
avekens, line 226557: Comment: "The order of a group element is always a nonnegative integer, deduction form."

set.mm
```diff
@@ -231100,6 +231343,14 @@ an extension of the previous (inserting an element and its inverse at
   ablgrp $p |- ( G e. Abel -> G e. Grp ) $=
     ( cabl wcel cgrp ccmn isabl simplbi ) ABCADCAECAFG $.
 
+  ${
+    ablgrpd.1 $e |- ( ph -> G e. Abel ) $.
+    $( Deduction associated with ~ ablgrp .  (Contributed by Rohan Ridenour,
+       3-Aug-2023.) $)
```
avekens, line 231349: Comment: "An Abelian group is a group, deduction form"

set.mm
```diff
@@ -237805,6 +238079,472 @@ factorization into prime power factors (even if the exponents are
...
+    ( csimpg wcel cgrp cnsg cfv c2o cen wbr issimpg simprbi ) ABCADCAEFGHIAJK
+    $.
+
+  ${
+    trivnsimpgd.1 $e |- B = ( Base ` G ) $.
+    trivnsimpgd.2 $e |- .0. = ( 0g ` G ) $.
+    trivnsimpgd.3 $e |- ( ph -> G e. Grp ) $.
+    trivnsimpgd.4 $e |- ( ph -> B = { .0. } ) $.
+    $( Trivial groups are not simple.  (Contributed by Rohan Ridenour,
+       3-Aug-2023.) $)
```
avekens, line 238144: Why? Because they have one and only one normal subgroup (because the subgroup of all elements **is** the subgroup containing only the identity). Maybe this can be added to the comment.

set.mm
```diff
@@ -237805,6 +238079,472 @@ factorization into prime power factors (even if the exponents are
...
+      UK $.
+  $}
+
+  ${
+    enpr2d.1 $e |- ( ph -> A e. C ) $.
+    enpr2d.2 $e |- ( ph -> B e. D ) $.
+    enpr2d.3 $e |- ( ph -> -. A = B ) $.
+    $( A pair with distinct elements is equinumerous to ordinal two.
+       (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
+    enpr2d $p |- ( ph -> { A , B } ~~ 2o ) $=
```
avekens, line 238181: This theorem should be moved up (near after ~en2sn?)

set.mm
```diff
@@ -237805,6 +238079,472 @@ factorization into prime power factors (even if the exponents are
...
+$)
+
+  ${
+    $d ph n $.  $d A n $.
+    cycsubggend.1 $e |- B = ( Base ` G ) $.
+    cycsubggend.2 $e |- .x. = ( .g ` G ) $.
+    cycsubggend.3 $e |- F = ( n e. ZZ |-> ( n .x. A ) ) $.
+    cycsubggend.4 $e |- ( ph -> A e. B ) $.
+    $( The cyclic subgroup generated by ` A ` includes its generator.
+       (Contributed by Rohan Ridenour, 3-Aug-2023.) $)
```
avekens, line 238265: "Although this theorem holds for any class ` G ` , the definition of ` F ` is only meaningful if ` G ` is a group." should be added to the comment (see also ~ cycsubmcl).

avekens:
I have finished my review now...

avekens:
> I put the "Simple groups" section after the "Abelian groups" section. In my opinion both of these should rather be subsections in "10.2 Groups", but I want to know what others think.

I agree with this proposal. Then ​section "10.3  Abelian groups" would become subsection "10.2.13  Abelian groups", and the subsections 10.3.x of section "10.3  Abelian groups" become subsubsections 10.2.13.x. Finally, we would have subsection "10.2.14  Simple groups".

set.mm
savask: Named it ` rspcime `.

set.mm
savask: Named it ` elrnmptdv ` and used your comment suggestion.

set.mm
savask: Used the same comment as in ` elrnmptdv `.

set.mm
savask: Renamed to ` phpeqd ` and used your comment suggestion.

set.mm
```diff
@@ -147960,6 +148027,27 @@ are used instead of sets because their representation is shorter (and more
       YI $.
   $}
 
+  ${
+    phphashd.1 $e |- ( ph -> A e. Fin ) $.
+    phphashd.2 $e |- ( ph -> B C_ A ) $.
+    phphashd.3 $e |- ( ph -> ( # ` A ) = ( # ` B ) ) $.
+    $( Equivalent of ~ php2d expressed using the hash function.  (Contributed
+       by Rohan Ridenour, 3-Aug-2023.) $)
+    phphashd $p |- ( ph -> A = B ) $=
```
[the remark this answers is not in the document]
savask: Updated the comment.

set.mm
```diff
@@ -147960,6 +148027,27 @@ are used instead of sets because their representation is shorter (and more
...
+      BCDEABGHCGHIZBCJKZFABLMCLMTUANDABCDEOBCPQRS $.
+  $}
+
+  ${
+    phphash2d.1 $e |- ( ph -> B e. Fin ) $.
+    phphash2d.2 $e |- ( ph -> A C_ B ) $.
+    phphash2d.3 $e |- ( ph -> ( # ` A ) = ( # ` B ) ) $.
+    $( Equivalent of ~ phphashd with reversed arguments.  (Contributed by Rohan
+       Ridenour, 3-Aug-2023.) $)
+    phphash2d $p |- ( ph -> A = B ) $=
```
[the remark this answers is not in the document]
savask: Renamed to ` phphashrd ` and updated the comment.

set.mm
savask: Renamed to ` gcdmultipled `. In the comment I referenced `M` to avoid repeating "nonnegative integer" twice.

set.mm
savask: Renamed to ` dvdsgcdidd `.

set.mm
savask: Placed it after `hash0`, I think it's a bit more logical that way.

set.mm
savask: Renamed to `trivsubgsnd`.

set.mm
savask: Used your comment suggestion.

set.mm
savask: Done. The better way would be "abelian group" (with lowercase), but set.mm seems to use both "Abelian" and "abelian" at different places.

set.mm
savask: I think this trivial (sic) remark can be avoided :-)

set.mm
savask: Moved it right after the block where `en2sn` is.

set.mm
savask: Added that remark.

savask:
> I agree with this proposal. ...

I also changed sections "Abelian groups" and "Simple groups" into subsections in the last commit, as discussed.

set.mm
avekens: What about using ~gcdmultipled to shorten the proofs of ~gcdmultiple and ~gcdmultiplez? I have pout it oin my TODO list and will look at it later,

set.mm
avekens: OK

avekens (approved):
All review remarks are considered, so that I can approve this PR now.

set.mm
savask: I haven't been able to shorten those with the automatic minimize tool, so perhaps that requires some manual labor.

tirix (approved):

jkingdon:
Looks like this is ready to go (two approvals, no outstanding comments, looked good to me in a quick browse).

If anyone sees any issues, please open a github issue or a new pull request.

avekens:
> Looks like this is ready to go (two approvals, no outstanding comments, looked good to me in a quick browse).
> 
> If anyone sees any issues, please open a github issue or a new pull request.

The update of changes-set.txt is missing (sorry that I did not mention this earlier). Each theorem moved from a mathbox to main should be recorded there. For the theorems in the completely moved section "Simple groups", however, only one entry should be sufficient (see, for example, 14-Jan-20 section "The Friendship Theorem" moved from AV's mathbox to main set.mm) . Can this be done in a separate PR?

PR #3742, opened 2024-01-04
delete outdated OLD theorems

wlammen:
janitorial work, as the title says.


digama0 (approved):

avekens:
There are still some outdated OLD theorems from 2022:
* ~ wfrlem4OLD
* ~ cnvoprabOLD
* ~ eubiOLD

Is there a reason why they were not deleted?

wlammen:
wfrlem4OLD is an oversight.  The other theorems are part of mathboxes, where I see the responsibility primarily with their owners.  I usually scan the file for "Jan-2023", for example, to find recent candidates for deletion.  Most often the hits are part of ordinary tags.  A small portion mark a day of obsolescence, and theorems get deleted when the 1-year-threshold is passed.   The number of search hits in total can reach a hundred or so, whence I stop at mathboxes to save effort.  From there on OLD theorems are rare, the reason for tracking previous versions are usually not given, or special.

Thanks for checking.  If you don't delete wfrlem4OLD,  I will do so in my next commit.

PR #3743, opened 2024-01-04
Words and cyclic permutations

tirix:
This adds several new theorems related to words, prefixes, concatenation and the "splice" operation.

The final theorem in this PR is cycpmco2, which proves that the composition of a cyclic permutation and a transposition of one element in the orbit, and one outside of the orbit, results in a new cyclic permutation.

One theorem "unima" moved to main.
One proof ~elfzom1p1elfzo shortened.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -83,6 +83,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 3-Jan-24 unima     [same]      moved from GS's mathbox to main set.mm
 29-Dec-23 uzidd     [same]      moved from GS's mathbox to main set.mm
 28-Dec-23 eqri      [same]      moved from TA's mathbox to main set.mm
 28-Dec-23 domep     dmep        moved from SF's mathbox to main set.mm


set.mm
```diff
@@ -462181,6 +462245,18 @@ its graph has a given second element (that is, function value).
       WJXAWRFOWPEWKWLWMWQFOWNWO $.
   $}
 
+  ${
+    subne0nn.1 $e |- ( ph -> M e. CC ) $.
+    subne0nn.2 $e |- ( ph -> N e. CC ) $.
+    subne0nn.3 $e |- ( ph -> ( M - N ) e. NN0 ) $.
+    subne0nn.4 $e |- ( ph -> M =/= N ) $.
+    $( A nonnegative difference is an integer if the two numbers are not equal.
```
savask, line 462253: Maybe better to say "A nonnegative difference is positive if the two numbers are not equal."?

set.mm
```diff
@@ -466126,6 +466419,224 @@ C_ dom ( ( ( T ` { I , J } ) o. F ) \ _I ) ) $=
...
+    $d W w $.  $d i ph $.
+    cycpmco2.c $e |- M = ( toCyc ` D ) $.
+    cycpmco2.s $e |- S = ( SymGrp ` D ) $.
+    cycpmco2.d $e |- ( ph -> D e. V ) $.
+    cycpmco2.w $e |- ( ph -> W e. dom M ) $.
+    cycpmco2.i $e |- ( ph -> I e. ( D \ ran W ) ) $.
+    cycpmco2.j $e |- ( ph -> J e. ran W ) $.
+    cycpmco2.e $e |- E = ( ( `' W ` J ) + 1 ) $.
+    cycpmco2.1 $e |- U = ( W splice <. E , E , <" I "> >. ) $.
+    $( The composition of a cyclic permutation and a well chosen 2-cycle is
```
savask, line 466433: Maybe it's better to be more precise, like you were in the pull request description: "The composition of a cyclic permutation and a transposition of one element in the cycle and one outside the cycle, results in a cyclic permutation with one more element in its orbit.

Also, this is a humongous proof! metamath-exe has some trouble loading the proof, let alone minimzing it, so maybe it's worth splitting it into smaller lemmata.

set.mm
tirix: Changed in upcoming commit.

set.mm
tirix: I've updated the comment in the last commit.

Yes it's a long proof, I will try to split it up. It also took ages for me to minimize.
(one obvious split is along the conjunction of the final statement)

set.mm
```diff
@@ -457709,6 +457718,25 @@ Class abstractions (a.k.a. class builders)
...
 $)
 
+  $( Elementhood to a union with a singleton.  (Contributed by Thierry Arnoux,
+     14-Dec-2023.) $)
+  elunsn $p |- ( A e. V -> ( A e. ( B u. { C } ) <-> ( A e. B \/ A = C ) ) )
+      $=
+    ( csn cun wcel wo wceq elun elsng orbi2d syl5bb ) ABCEZFGABGZANGZHADGZOACIZ
+    HABNJQPROACDKLM $.
+
+  $( Negated membership to an union.  (Contributed by Thierry Arnoux,
```
avekens, line 457728: "to a union" or maybe better "for a union".

set.mm
```diff
@@ -461739,6 +461767,24 @@ its graph has a given second element (that is, function value).
       EIKUHUIUOWIWK $.
   $}
 
+  $( Elementhood in a open integer interval, except its lower bound.
+     (Contributed by Thierry Arnoux, 1-Jan-2024.) $)
```
avekens, line 461771: "in an open ..."

set.mm
```diff
@@ -461739,6 +461767,24 @@ its graph has a given second element (that is, function value).
...
+  fzne1 $p |- ( ( K e. ( M ... N ) /\ K =/= M ) -> K e. ( ( M + 1 ) ... N ) )
+      $=
+    ( wne cfz co wcel wceq wn c1 caddc df-ne wo cuz cfv elfzuz2 elfzp12 syl ibi
+    wb orcanai sylan2b ) ABDABCEFGZABHZIABJKFCEFGZABLUCUDUEUCUDUEMZUCCBNOGUCUFT
+    ABCPABCQRSUAUB $.
+
+  $( Elementhood in an half-open interval, except the lower bound.
```
avekens, line 461778: "in a half-open ..."

set.mm
avekens: `( M ... N )` is not an open interval. It's a "finite set of sequential integers", or a "finite sequence of integers " or a "finite interval of integers".

set.mm
avekens: The comment is somehow misleading, maybe: "Elementhood of an integer and its predecessor in finite intervals of integers."

set.mm
```diff
@@ -466126,6 +466419,469 @@ C_ dom ( ( ( T ` { I , J } ) o. F ) \ _I ) ) $=
...
+    cycpmco2.c $e |- M = ( toCyc ` D ) $.
+    cycpmco2.s $e |- S = ( SymGrp ` D ) $.
+    cycpmco2.d $e |- ( ph -> D e. V ) $.
+    cycpmco2.w $e |- ( ph -> W e. dom M ) $.
+    cycpmco2.i $e |- ( ph -> I e. ( D \ ran W ) ) $.
+    cycpmco2.j $e |- ( ph -> J e. ran W ) $.
+    cycpmco2.e $e |- E = ( ( `' W ` J ) + 1 ) $.
+    cycpmco2.1 $e |- U = ( W splice <. E , E , <" I "> >. ) $.
+    $( The composition of a cyclic permutation and a well-chosen transposition
+       is injective.  (Contributed by Thierry Arnoux, 4-Jan-2024.) $)
```
savask, line 466434: I have a hard time understanding what's going on, but judging from the comment alone, isn't the composition of any two permutations injective? I.e. the transposition doesn't have to be well-chosen at all.

tirix: Yes, I need to rewrite this comment. 

The cyclic permutation in question is not `U`, but ``( M ` U )``, and it's `U` which is injective.

`U` is actually the representation of a cycle. For example a word `W = <" A B C ">` would correspond to the cyclic permutation ``( M ` W )`` mapping `A` → `B`, `B` → `C`, `C` → `A`, `D` → `D`. 
In order for it to represent a cycle, the same element should not appear twice in the word, so it has to be injective.

tirix: Maybe simply:
> The word `U` used in `~ cycpmco2` is injective, so it can represent a cycle and form a cyclic permutation ``( M ` U )``.

savask: I think that's better, yeah.

savask (approved):

avekens (approved):

PR #3744, opened 2024-01-05
Miscellaneous

avekens:
Enhancement of comment for df-rn (Definition of `ran`), see also discussion in PR #3741

Conventions:
* Revision of section "Distinctness or freeness": order of elements in a $d-condition (see also discussion in PR #3573)

Mathboxes:
* ~mptima moved from GS's mathbox to main
* ~fproj and ~fimaproj moved from TA's mathbox to main
* ~offval0 removed from AV's mathbox (duplicate of ~offval3)

Usage of ~fpar and ~fsplit, see also discussion in PR #3735
* Example ~ex-fpar for ~fpar added
* combination ~fsplitfpar of ~ fsplit and ~ fpar  added
* connection to function operation map ` oF ` added (~offsplitfpar)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -83,6 +83,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 4-Jan-24 fimaproj  [same]      moved from TA's mathbox to main set.mm
+ 4-Jan-24 fproj     [same]      moved from TA's mathbox to main set.mm
+ 4-Jan-24 mptima    [same]      moved from GS's mathbox to main set.mm
 29-Dec-23 uzidd     [same]      moved from GS's mathbox to main set.mm
 28-Dec-23 eqri      [same]      moved from TA's mathbox to main set.mm
 28-Dec-23 domep     dmep        moved from SF's mathbox to main set.mm


set.mm
```diff
@@ -51013,8 +51013,12 @@ We have not yet defined relations ( ~ df-rel ), but here we introduce a few
        ` F = { <. 2 , 6 >. , <. 3 , 9 >. } -> ran F = { 6 , 9 } ` ( ~ ex-rn ).
        Contrast with domain (defined in ~ df-dm ).  For alternate definitions,
        see ~ dfrn2 , ~ dfrn3 , and ~ dfrn4 .  The notation " ` ran ` " is used
-       by Enderton; other authors sometimes use script R or script W.
-       (Contributed by NM, 1-Aug-1994.) $)
+       by Enderton; other authors sometimes use script R or script W. The range
```
benjub, line 51016:
```suggestion
       by Enderton. The range
```
I think this is superfluous (and not before long, we would have to add the whole alphabet).

set.mm
```diff
@@ -74357,6 +74371,61 @@ of related elements (Contributed by Alexander van der Vekens,
...
+      UVOUWBUWAUVOQZIUWCYFUWCQYKUWBUWHXHYFRBXDYKUWBUWHYKUWBLZUWHYRUVOQZUWIYFYFB
+      BYKUWBXEZUWKXFUWBUWHUWJWFYKUWAYRUVOXITWLXJWOXKXLSXMUVOBFCXNXOXRYCRQZABPUV
+      NXRUWLABUWLXRUVTLYAYBUSSWEABYCYDRUVLVNWOIBXSYDXPXQWL $.
+
+    $d C a $.  $d O a $.  $d V a $.  $d W a $.
+    $( Express the function operation map ` oF ` by the functions defined in
+       ~ fsplit and ~ fpar .  (Contributed by AV, 4-Jan-2024.) $)
+    offsplitfpar $p |- ( ( ( F Fn A /\ G Fn A ) /\ ( F e. V /\ G e. W )
+                           /\ ( O Fn C /\ ( ran F X. ran G ) C_ C ) )
+                         -> ( O o. ( H o. S ) ) = ( F oF O G ) ) $=
```
benjub, line 74416: Thanks for formalizing that, that's nice !
Only: the choice of "O" is a bit strange (that letter is seldom used since it can be confused with zero); Mario chose "R" in https://us.metamath.org/mpeuni/df-of.html and that looks fine (it's indeed an operation, not a relation, but that's fine in my opinion).

set.mm
```diff
@@ -74725,6 +74794,60 @@ of related elements (Contributed by Alexander van der Vekens,
...
   $}
 
+  ${
+    $d A a b x y z $.  $d B a b x y z $.  $d F a b c x y z $.
+    $d G a b c x y z $.  $d X a b c z $.  $d Y a b c z $.
+    fvproj.h $e |- H = ( x e. A , y e. B |-> <. ( F ` x ) , ( G ` y ) >. ) $.
+    ${
+      fvproj.x $e |- ( ph -> X e. A ) $.
+      fvproj.y $e |- ( ph -> Y e. B ) $.
+      $( Value of a function on pairs, given two projections ` F ` and ` G ` .
```
benjub, line 74804:
```suggestion
      $( Value of a function on ordered pairs with values expressed as ordered pairs. Note that ` F ` and ` G ` are the projections of ` H ` to the first and second coordinate respectively.
```
Justification:
* it is important to say that the values are ordered pairs and are given explicitly as ordered pairs
* I avoided "given two projections", because the hypothesis does not literally say "here are two projections". We humans simply note, outside the formal systems, that these are the projections.

set.mm
```diff
@@ -74725,6 +74794,60 @@ of related elements (Contributed by Alexander van der Vekens,
...
+    $}
+
+    ${
+      $d H a b c x y z $.  $d ph a b c z $.
+      fimaproj.f $e |- ( ph -> F Fn A ) $.
+      fimaproj.g $e |- ( ph -> G Fn B ) $.
+      fimaproj.x $e |- ( ph -> X C_ A ) $.
+      fimaproj.y $e |- ( ph -> Y C_ B ) $.
+      $( Image of a cartesian product for a function on pairs, given two
+         projections ` F ` and ` G ` .  (Contributed by Thierry Arnoux,
```
benjub, line 74822: Same remarks.

set.mm
```diff
@@ -422891,24 +423014,30 @@ varying definitions (some start from 0, others start from 1), but
...
 
-       <li>"$d x A $." should be read "Assume x is not a variable occurring in
-       class A."</li>
+       <li>"$d ` A x ` $." should be read "Assume ` x ` is not a variable
+       occurring in class ` A `."</li>
+
+       <li>"$d ` A x ` $.  $d ` ps x ` $.
+       $e |- ` ( x = A -> ( ph <-> ps ) ) ` $." is an idiom often used instead
+       of explicit substitution, meaning "Assume ` psi ` results from the
+       proper substitution of ` A ` for ` x ` in ` phi `."</li>
```
benjub, line 423033: Can you take that opportunity to add something like: "Therefore, we often use the term "implicit substitution" for such a hypothesis." ? Indeed, that term is used in several parts of set.mm ?

set.mm
```diff
@@ -422891,24 +423014,30 @@ varying definitions (some start from 0, others start from 1), but
...
-       <li>"$d x A $.  $d x ps $.  $e |- ` ( x = A -> ( ph <-> ps ) ) ` $."
-       is an idiom
-       often used instead of explicit substitution, meaning "Assume psi results
-       from the proper substitution of A for x in phi."</li>
+       <li>Class and wff variables should appear at the beginning of distinct
+       variable conditions, and setvars should be in alphabetical order.
+       E.g., "$d ` Z x y ` $.",  "$d ` ps a x ` $.".  This convention should
+       be applied for new theorems (formerly, the class and wff variables
+       mostly appear at the end) and will be assured by a formatter in the
+       future.</li>
```
benjub, line 423040: It will be a pain to enforce as long as we don't have that formatter...

set.mm
avekens: > I think this is superfluous (and not before long, we would have to add the whole alphabet).

Maybe "W" comes from the German "Wertebereich". But you are right, thee is not much value in this hint, so I will remove it.

set.mm
avekens: I have chosen "O" intentionally (for "operation" or "operator"), because I always get confused with using R as operation. ` A R B ` (binary relation, see ~df-br) and ` ( A R B ) ` (operation value, df-ov) can be easily confused, especially if embedded within a larger statement. Maybe we should choose a special connective symbol like `.+`, e.g.: `( .+ o. ( H o. S ) ) = ( F oF .+ G )`. There are already such cases in set.mm., see ~seqof or ~ghmplusg.

set.mm
avekens: I can change this, but I would like to hear @tirix `s opinion because it is from his mathbox.

set.mm
avekens: Yes, good idea.

set.mm
avekens: Already existing theorems should not be touched, and for new theorems this should not be much efforts (metamath.exe does not create distinct variable conditions, and mmj2 and lamp already respect this convention.

set.mm
benjub: Then maybe leave it like that. Color and font will help distinguish it from zero anyway.

set.mm
benjub: Was this discussed before, actually ?

Which alphabetical order to use ? For the Latin alphabet, it's commonly agreed that the first three letters are x, y, z. But already for the fourth letter, I think the most common is 't' but there are other conventions. Similarly for the Greek alphabet: phi, psi, chi are the first three, but after that there is no dominating convention.

If an order had to be adopted, I think it should be the order in which they are declared (with the $v statment), or better, the order of the $f-statements that type them (remembering these are local).

set.mm
tirix: To me, the point of the comment is exactly to provide the information that we, as humans, need to better understand what the theorem means, and what it's used for. I usually prefer that to simply "reading alout" what the math formula says.

That's why I mentioned "projection" prominently in the description. 

It's like saying "The X operation commutes" rather than "the result of the X operation is unchanged when operands are swapped".

I'm fine with any rewriting which keeps this spirit.

set.mm
```diff
@@ -425366,6 +425495,35 @@ Norman Megill (2007) section 1.1.3.  Megill then states, "A number of
       WO $.
   $}
 
+  ${
+    $d x y A $.  $d x y B $.  $d x y F $.  $d x y G $.
```
tirix, line 425499: The comment added above says _Class and wff variables should appear at the beginning of distinct variable conditions_, it looks like this distinct variable condition (in the same PR!) does not adhere to it!

avekens: > The comment added above says _Class and wff variables should appear at the beginning of distinct variable conditions_, it looks like this distinct variable condition (in the same PR!) does not adhere to it!

You are right, I just copied the distinct variable conditions from ~fpar. I will change it for ~ex-fpar.

set.mm
avekens: > Was this discussed before, actually ?

Yes, see discussion in PR #3573, but also  https://github.com/expln/metamath-lamp/issues/172 and  https://groups.google.com/g/metamath/c/vRXuz7Qs5wY/m/eaxltrMsBwAJ 

The current convention concerns the order within a distinct variable condition: here, only the small Latin letters (for setvar variables) need to be sorted (alphabetically in the usual meaning, i.e. a b c ... x y z). This seems to be a convention which was silently followed already before (see, for example,  ~csbiebg: `$d a x A $.`).

Another issue is the ordering of the distinct variable conditions themselves. I think in most cases, the DV conditions (starting) with a class variable come first, ordered according to the class variables which are upper case Latin letters ( `$d A x y $.` ...` $d Z a b $.`), followed by  DV conditions (starting) with other class variables (special symbols like `.+` or `.(+)`), followed by DV conditions (starting) with wwf variables (greek letters, how ordered?) ending with DV conditions without class and wff variables (consisting of setvar variables only). This issue, however, was not discussed yet (at least as far as I know), and should be treated separately.

set.mm
benjub: Your example with commutation is not the same. Here, if I began by reading your comment, I would formalize it as
```
$e |- H Fn ( A X. B ) $.
$e |- F = ( 1st o. H ) $.
$e |- G = ( 2nd o. H ) $.
...
```
In other words, in order to read `fvproj` as the formalization of your comment, we need the result
```
$e |- H = ( x e. A , y e. B |-> <. ( F ` x ) , ( G ` y ) >. ) $.
$e |- F Fn A $.
$e |- G Fn B $.
$p |- ( F = ( 1st o. H ) /\ G = ( 2nd o. H ) ) $= ? $.
```

set.mm
avekens: Done.

set.mm
avekens: I replaced `O` by `.+`now.

set.mm
avekens: Done.

set.mm
avekens: Done.

benjub:
I'll review ASAP but for the moment I'm getting (as very often when trying to see later commits of a PR) the "angry unicorn" page ("This page is taking too long to load. Sorry about that. Please try refreshing and contact us if the problem persists.").

set.mm
tirix: Yes, sure!

tirix (approved):

benjub (approved):

PR #3746, opened 2024-01-06
The exponential function is continuous (iset.mm)

jkingdon:
The theorems added here are efcn , sincn , and coscn .

There are also a number of notes added to mmil.html about what would be involved in intuitionizing various theorems in this section and the next one (up to `pire` which is the key theorem in this area).

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11201,6 +11201,25 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>the set.mm proof uses reeff1olem</td>
 </tr>
 
+<tr>
+  <td>reefiso</td>
+  <td><i>none</i></td>
+  <td>the set.mm proof uses reeff1o and the converse of
+  ~ efltim</td>
+</tr>
+
+<tr>
+  <td>efcvx</td>
+  <td><i>none</i></td>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -142684,6 +142684,79 @@ S C_ ( P ( ball ` D ) T ) ) $=
...
+  $}
+
+
+$(
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  Properties of pi = 3.14159...
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+$)
+
+  $( Lemma for pire , pigt2lt4 and sinpi .  (Contributed by Mario Carneiro,
```
avekens, line 142750: do not forget to insert "~" here to reference the mentioned theorems as soon as they are available.

avekens (approved):

tirix (approved):

benjub (approved):

PR #3753, opened 2024-01-10
Mathbox for theorems not requiring ax-13

GinoGiotto:
This is the first step towards https://groups.google.com/g/metamath/c/OB2_9sYgDfA.

This mathbox contains the first 84 of the 129 lemmas that I used in https://github.com/GinoGiotto/set.mm/tree/ax-13-complete to reduce ax-13 dependencies. I believe that 129 is the minimum amount of needed lemmas to not significantly affect the proof length caused by automatic minimizations with /MAY_GROW.

There shouldn't be redundancies with theorems in main, since I did some heavy pruning. However there are a few redundancies with @benjub mathbox:

`cbval2w` is identical to [bj-cbval2v](https://us.metamath.org/mpeuni/bj-cbval2v.html)
`cbvex2w` is identical to [bj-cbvex2v](https://us.metamath.org/mpeuni/bj-cbvex2v.html)
`dral1w` is identical to [bj-dral1v](https://us.metamath.org/mpeuni/bj-dral1v.html)
`drex1w` is identical to [bj-drex1v](https://us.metamath.org/mpeuni/bj-drex1v.html)
`drnf1w` is identical to [bj-drnf1v](https://us.metamath.org/mpeuni/bj-drnf1v.html)

I checked that the proofs are the same as well except for the cbval2w, bj-cbval2v pair, where cbval2w is shorter.

Probably the easiest thing to do is to move these five BJ versions to main, even if one of his proofs is longer. My guess is that his proof will eventually turn into my proof after the other lemmas will be moved into main and automatic shortenings will be performed.


set.mm
```diff
@@ -629242,6 +629242,963 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+  ~ ax-13 . Similarly to ~ ax13w , the theorems proved here are weakened by
+  introducing one or more additional dv conditions, hence labeled as xxxw
+  (the pattern xxxv creates naming conflicts with main theorems, despite having
+  different statements. If xxxv is preferred, then the versions in main should
+  be renamed as xxxvv since they all have more dv conditions). These weaker
+  versions can replace the proof steps occupied by their original counterparts
+  with the command "MINIMIZE_WITH xxxw /MAY_GROW". Since these theorems
+  follow the already present proof chain, automatic minimizations should not
+  cause significant proof changes or lengthenings, hence OLD proofs are not
+  necessary (and they would also be unpractical).
```
GinoGiotto, line 629268: The lack of OLD/ALT proofs caused some people to frawn in the past, so I'm opening this thread to discuss about it. I can change/delete this text if you don't agree with it, but you should be aware that it would discourage me from getting further (unless I have some kind of automation tool, which would be possible in this case). We are talking about thousands changed proofs here, which would correspond to thousands of added OLD/ALT proofs. Please consider that it's not a trivial request to ask (also it would be for relatively minor changes). You can check commit https://github.com/metamath/set.mm/commit/6fc6153e0e4df76bf73bfeef76151d5354bc972c to get an idea of the size of this thing.

tirix:
So what would be your program?

Is it something like:
1. Add most new lemmas to mathbox (this PR)
2. Move new lemmas to main next to the original theorems
3. Complete with @benjub's lemmas
4. Use the new lemmas in many (thousands) of existing theorems in order to match [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete)
5. Confirm ax-13 usage.

tirix (approved):
Ok for me as this is mathbox-only and follows the path already discussed.

tirix:
I'd be interested to know if some of the remaining theorems using ax-13 are in my mathbox. If so, I'd be willing to help resolving it.

GinoGiotto:
> 2. Move new lemmas to main next to the original theorems

In [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete) I placed most theorems in the ax-12 section, since it is the last axiom they use. After that the following lemmas are placed next to their original versions.

> 4. Use the new lemmas in many (thousands) of existing theorems in order to match [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete).

There are a few occurrences where the necessary lemmas are already present in main. So in these cases it's enough to minimize with them as mentioned in the header comment of my mathbox (example: replace [rgen2a](https://us.metamath.org/mpeuni/rgen2a.html) with [rgen2](https://us.metamath.org/mpeuni/rgen2.html) whenever possible). I think I will start from these even before adding my lemmas.

The rest will be similar to what you described. There is the somewhat inconvenient problem that ax-13 usage depends on the overall proof chains in set.mm, and those will always change. So probably the necessary steps won't be exactly the same as [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete) and I will have to adapt to the ever evolving situation. Still I think starting from the beginning and progressively going downwards should be a safe approach.

> I'd be interested to know if some of the remaining theorems using ax-13 are in my mathbox. If so, I'd be willing to help resolving it.

I checked: your mathbox is clean except for opreu2reuALT and xrge0tmdALT, but these are ALT proofs so I wasn't going to change them anyway.

set.mm
wlammen: I'm not sure whether you need this passage at all.  Our conventions page states
`When a proof is shortened automatically (...), then it is not necessary to keep the old proof, nor to add credit for the shortening.`
As for adding lemmas to eliminate ax-13 usage:  Fine from my point of view.  It is our goal to reduce axiom usage - even it comes at the cost of a complete sort of theory developed for that purpose.

GinoGiotto: >  `When a proof is shortened automatically (...), then it is not necessary to keep the old proof, nor to add credit for the shortening.`

This is specific for proof shortenings, but here the new proofs won't be shorter. I think I'll leave this passage unless the conventions will be extended for such cases.

wlammen (approved):

set.mm
```diff
@@ -629242,6 +629242,963 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+  $}
+
+  ${
+    $d x y $.
+    $( Version of ~ sb2 with a disjoint variable condition, which does not
+       require ~ ax-13 .  (Contributed by Gino Giotto, 10-Jan-2024.) $)
+    sb2w $p |- ( A. x ( x = y -> ph ) -> [ y / x ] ph ) $=
+      ( weq wal wi wsb alim stdpc4 syl6com wn sb4bw biimprd pm2.61i ) BCDZBEZOA
+      FBEZABCGZFQPABEROABHABCIJPKRQABCLMN $.
+  $}
```
benjub, line 629330: This is (one direction of) https://us.metamath.org/mpeuni/sb6.html . So it can be proved from it and ~biimpri, but more importantly, I think it does not deserve its own statement.

benjub:
I should take some time to read this and your other messages through, hopefully this weekend (this doesn't block this PR, since it is mathbox-only).

I definitely want to move the needed lemmas to Main in order to reduce ax-13 dependencies, but I'm not sure moving all these lemmas in a bloc is necessary. For instance, there is a long chain of lemmas starting with this PR's ~sb4bw, each used only once, so I do not know if it's worth adding them all.  In Section "20.15.4.12  Removing dependencies on ax-13 (and ax-11)", which is made of two parts: versions of lemmas with DV, and then ax-13-free proofs of theorems from Main, I had tried to select only lemmas which provided the biggest gains.  I did not carry on with that approach and I am glad you are taking this over, but we should go step-by-step when moving the various parts to the main section.

tirix:
For those theorem which are used only once, what about keeping only the version which does not depend on ax-13?

set.mm
GinoGiotto: It's very strange that the minimizer doesn't catch this shortening, I even used [the stronger version](https://github.com/metamath/set.mm/pull/3619#issuecomment-1801361982) of it and still doesn't see it. However I should be careful before deleting one of these theorems, because I need to make sure that such deletion won't cause proof lengthenings (which would make the overall minimization process more difficult). For those 5 redundant ones with your mathbox it's easy because they are identical, but otherwise I need a bit of time to verify that.

Still that's a good catch, any other redundancy that you can find is appreciated.

GinoGiotto:
> I definitely want to move the needed lemmas to Main in order to reduce ax-13 dependencies, but I'm not sure moving all these lemmas in a bloc is necessary.

I won't do that, I was thinking around 5 per PR (maybe less or maybe more, but around that number). Of course, it would be easier for me to do less PRs and fixing imperfections later, but I assumed that would be frowned by most people.


> For instance, there is a long chain of lemmas starting with this PR's ~sb4bw, each used only once, so I do not know if it's worth adding them all.

Can you tell me which ones are used only once? I did check the usage of my lemmas with `SHOW USAGE xxxw /RECURSIVE` in [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete) and most of the early ones are used by thousands of theorems. There are a few ones that are used by a few dozens of theorems and some other by less than 10. It's ok if we don't want them, but it may come at the cost of proof lengthenings. Then obviously it's more likely for later ones to become less pervasive (btw "show usage" is actually the trick I used to prune away unnecessary lemmas).

> In Section "20.15.4.12 Removing dependencies on ax-13 (and ax-11)", which is made of two parts: versions of lemmas with DV, and then ax-13-free proofs of theorems from Main

New theorems for proofs that don't require additional dv conditions are definitely unnecessary, and I purposely threw away all of them. It's always possible to find an earlier theorem that does require dv conditions that can be used directly on the already exisiting theorem.

> I had tried to select only lemmas which provided the biggest gains.

I agree with this approach and that's my intent as well. Again I tried to do my best while pruning, but it can always be something that I missed. Any help in this regard is appreciated.

set.mm
GinoGiotto: Ok, I checked and can safely get rid of `sb4bw` and `sb2w` at the minor cost of making `hbsb2w` a bit longer. I can also remove `hbsb2w` as well by lengthening `nfsb2w.` Again this just depends on what's the tolerance for proof lengthenings here.

GinoGiotto:
> For instance, there is a long chain of lemmas starting with this PR's ~sb4bw, each used only once, so I do not know if it's worth adding them all.

I think I understand now which ones you are talking about, you are referring to this chain:

sb4bw -> sb2w -> hbsb2w -> nfsb2w -> nfsb4tw -> nfsbdw -> nfabdw -> 11 theorems -> (from here it explodes)

We can eliminate some intermediate links by lenghtening the following ones, however the overall chain is still needed because recursively `nfabdw` is used by 29,901 theorems.

benjub:
> > For instance, there is a long chain of lemmas starting with this PR's ~sb4bw, each used only once, so I do not know if it's worth adding them all.
> 
> I think I understand now which ones you are talking about, you are referring to this chain:
> 
> sb4bw -> sb2w -> hbsb2w -> nfsb2w -> nfsb4tw -> nfsbdw -> nfabdw -> 11 theorems -> (from here it explodes)
> 
> We can eliminate some intermediate links by lenghtening the following ones, however the overall chain is still needed because recursively `nfabdw` is used by 29,901 theorems.

@GinoGiotto : yes, that's it. There might be other chains like that, were the first 2 to 6 theorems form a "simple chain" like above, and multi-usages only come at the 3rd to 6th step. In these cases, proving only the technical lemma associated with that step (and not the previous ones) might be more economical.

PR #3756, opened 2024-01-12
Removed sb2w,sb4bw

GinoGiotto:
As suggested by https://github.com/metamath/set.mm/pull/3753#discussion_r1448605402.

Before removing one of my lemmas I always have to check that it would be safe to do in [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete) as well. In this case it's safe.


set.mm
```diff
@@ -629570,8 +629561,8 @@ standardize vectors to a length (norm) of one, but such definitions require
     $( Version of ~ hbsb2 with a disjoint variable condition, which does not
        require ~ ax-13 .  (Contributed by Gino Giotto, 10-Jan-2024.) $)
     hbsb2w $p |- ( -. A. x x = y -> ( [ y / x ] ph -> A. x [ y / x ] ph ) ) $=
-      ( weq wal wn wsb wi sb4bw sb2w axc4i syl6bi ) BCDZBEFABCGZMAHZBENBEABCION
-      BABCJKL $.
+      ( weq wal wn wsb sb4bw alim stdpc4 syl6com biimprd pm2.61i axc4i syl6bi
```
wlammen, line 629564: The idea of BJ was to replace sb2w with one direction of sb6 (i.e. sb6 followed by biimpri).  Is there a technical reason for using alim stdpc4 syl6.com biimprd, pm2.61i instead?

GinoGiotto: I forgot, thanks.

GinoGiotto: > EDIT: sb4bw is also a weakened form of sb6, that should be fully replaced, if this will not hamper the minimizing process.

I'm going to check out this one as well.

GinoGiotto: Removed sb4bw.

wlammen (approved):

set.mm
```diff
@@ -629568,10 +629546,12 @@ standardize vectors to a length (norm) of one, but such definitions require
...
     $( Version of ~ hbsb2 with a disjoint variable condition, which does not
-       require ~ ax-13 .  (Contributed by Gino Giotto, 10-Jan-2024.) $)
+       require ~ ax-13 .  (Contributed by Gino Giotto, 10-Jan-2024.)  (Proof
+       shortened by BJ, 11-Jan-2024.)  (Revised by Wolf Lammen,
+       12-Jan-2024.) $)
     hbsb2w $p |- ( -. A. x x = y -> ( [ y / x ] ph -> A. x [ y / x ] ph ) ) $=
-      ( weq wal wn wsb wi sb4bw sb2w axc4i syl6bi ) BCDZBEFABCGZMAHZBENBEABCION
-      BABCJKL $.
+      ( weq wal wn wsb wi wb sb6 a1i biimpri axc4i syl6bi ) BCDZBEFZABCGZOAHZBE
+      ZQBEQSIPABCJZKRQBQSTLMN $.
```
benjub, line 629554: Replace this with hbs1.

More generally, if you see an antecedent `|- ( -. A. x x = y -> ` then the version with DV(x,y) will not need that antecedent.

set.mm
benjub: Yes, @wlammen is right: this is what I meant in general. In this particular case, we can actually do without hbsb2w altogether and use hbs1

set.mm
GinoGiotto: I'm preparing a new PR to remove the distinctors from my lemmas (and delete hbsb2w). Probably I'm going to publish it tomorrow (I won't add the commit here).

benjub: So: do you want this PR merged as is ? or deleted ?  (Fine by me either way.)

GinoGiotto: Merge as is. This takes a bit of time because I have make some revisions to [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete)

benjub (approved):

PR #3759, opened 2024-01-12
Mathboxes ordering

benjub:
and some janitorial work.


set.mm
```diff
@@ -393403,9 +393404,9 @@ then any point of the half line ( ` R ` ` A ` ) also lies opposite to
     isinag.c $e |- ( ph -> C e. P ) $.
     ${
       isinag.g $e |- ( ph -> G e. V ) $.
-      $( Property for point ` X ` to lie in the angle ` <" A B C "> ` Defnition
-         11.23 of [Schwabhauser] p. 101.  (Contributed by Thierry Arnoux,
-         15-Aug-2020.) $)
+      $( Property for point ` X ` to lie in the angle ` <" A B C "> `
```
wlammen, line 393407: I'd rather see this sentence end with a full stop.  Also possible is a comma, but then definition should be in lowercase.

benjub: I spotted the typo, but not the missing period. I'm going to fix it.

wlammen (approved):

PR #3768, opened 2024-01-14
Uniformize mathboxes for use of 'WRITE SOURCE set.mm / SPLIT'.

benjub:
Gérard warned me that I messed up the ordering of mathboxes in #3759. It's a hassle to order them now that we have over 40 mathboxes, and the alphabetical order, with full names mixed with initials and nicknames, is not clear.

Therefore, I first uniformized the mathboxes so that if we use `WRITE SOURCE set.mm / SPLIT`, we can then reorder mathboxes as simply as reordering lines in `set-mbox.mm` (not in this PR).  I uniformized them using the most common convention, which contains a `$( Skip $[ set-main.mm $] $)` at the beginning of each mathbox.  This foster collaboration: if a contributor uses `WRITE SOURCE set.mm / SPLIT`, then his mathbox will be placed after all other mathboxes, so that he can use any theoremd by others (and also minimize using them, by typing `MINIMIZE_WITH */ INCLUDE_MATBOXES`).


avekens:
I'm a little bit torn between the advantages and disadvantages of splitting off all mathboxes. The advantage is the uniformity, and the possibility to use this uniformity to clean up the mathboxes and to support the work with them, the disadvantage is that it creates a lot of files if the splitting is performed.

I once used the split feature, but I  do not use it anymore at the moment, because it creates a lot of overhead. Since my mathbox is almost at the end, I have the possibility to use most of the theorems in other's mathboxes. Furthermore, I usually use "working files" xxx.mm which include set.mm at the beginning (`$[ set.mm $]`) to create new theorems, so that I have access to all mathboxes.

I like the idea to facilitate the usage of theorems in other's mathboxes , especially to identify theorems of common interest, which can (and must, if used) be moved to main.

benjub:
Contributors do not have to use the `WRITE SOURCE set.mm/SPLIT` command. This is optional.

As for the number of files created, this is not a problem since they are in a single folder and automatically managed. If for a reason I don't see now we want to limit their number, we can decide that each contributor's mathbox should not itself be subdivided into multiple files.

tirix (approved):
Why not.

I'm not using the feature anymore, so I'm fine if the subdivisions inside my mathbox need to removed.

avekens:
> As for the number of files created, this is not a problem since they are in a single folder and automatically managed. If for a reason I don't see now we want to limit their number, we can decide that each contributor's mathbox should not itself be subdivided into multiple files.

If we have already a lot of files, it does not matter to have some more. So everybody should be free to use the splitting of his/her mathbox (of course excessive abuse should not be allowed, e.g., putting each single theorem into a separate file...).

avekens (approved):

PR #3770, opened 2024-01-15
A version of the least upper bound property for positive reals in iset.mm

jkingdon:
The larger mission here is to prove the Monotonic Intermediate Value Theorem (Theorem 5.5 of [Bauer], p. 494). But since that is several steps away, it seemed better to submit what I have so far.

This is one of the most common versions of the least upper bound property when excluded middle is not available. It says that an inhabited, bounded-above, and located set of reals has a supremum (in the case of this pull request, reals means positive reals as in `P.`). Located means that given ` x < y ` , either the set contains an element greater than ` x ` , or ` y ` is an upper bound for the set.

The proof has a bunch of lemmas but pretty much the expected ones for a proof that something (in this case, the supremum) is a positive real.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -5475,13 +5475,12 @@ biconditional or which involve proper subsets.</TD>
 <TD>~ recexprlemss1l , ~ recexprlemss1u , ~ recexprlemex </TD>
 </TR>
 
-<TR>
-<TD>supexpr , suplem1pr , suplem2pr</TD>
-<TD>~ caucvgprpr</TD>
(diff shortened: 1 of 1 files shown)


tirix:
Nice! 
Is proving this theorem in `P.` a first step for later proving it in `RR`?

jkingdon:
> Nice! Is proving this theorem in `P.` a first step for later proving it in `RR`?

Yes. That part should be relatively close to set.mm.

tirix (approved):

PR #3780, opened 2024-01-20
One direction of sb5 is provable from fewer axioms

wlammen:
1. One direction of sb5 is provable from axioms up to ax-7 only.  Add this theorem as sb1v, matching the similar sb1 in name.
2. Delete outdated OLD theorems

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16109,6 +16109,7 @@ New usage of "lsatfixedN" is discouraged (1 uses).
 New usage of "lshpinN" is discouraged (0 uses).
 New usage of "lshpnel2N" is discouraged (0 uses).
 New usage of "lshpset2N" is discouraged (1 uses).
+New usage of "lsmidmOLD" is discouraged (0 uses).
 New usage of "ltaddnq" is discouraged (7 uses).
 New usage of "ltaddpr" is discouraged (5 uses).
 New usage of "ltaddpr2" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -541596,6 +541577,14 @@ dependency is expressed in our hypothesis (called implicit
       UGCBUGCLMABCNOUCUEABUECBHAICJZUCAACBPUCUHAACBNQRSTUA $.
   $}
 
+  ${
+    $d x y $.
+    $( One direction of ~ sb5 is provable from fewer axioms.  See also ~ sb1 .
+       (Contributed by Wolf Lammen, 20-Jan-2024.) $)
```
benjub, line 541583:
```suggestion
    $( One direction of ~ sb5 , provable from fewer axioms.  Version of
        ~ sb1 with disjoint variables, requiring fewer axioms.
       (Contributed by Wolf Lammen, 20-Jan-2024.) $)
```

benjub:
I think wl-sb1v should be moved to Main. Can it be used in other theorems (possibly adapting the proofs), in particularly the ones using sb1/sb5 ?

benjub (approved):

wlammen:
`Can it be used in other theorems (possibly adapting the proofs), in particularly the ones using sb1/sb5 ?`

I did a quick check to see whether it helps in other theorems, but I found no instance.

benjub:
> `Can it be used in other theorems (possibly adapting the proofs), in particularly the ones using sb1/sb5 ?`
> 
> I did a quick check to see whether it helps in other theorems, but I found no instance.

To me, even without applications, it's worth it in its own sake. So, now or in a future PR, as you prefer.

icecream17 (approved):

tirix (approved):

set.mm
```diff
@@ -229301,17 +229293,6 @@ U C_ ( T .(+) U ) ) $=
       ( csubg cfv wcel csubmnd co wceq subgsubm smndlsmidm syl ) BCEFGBCHFGBBAI
       BJBCKABCDLM $.
 
-    $( Obsolete proof of ~ lsmidm as of 13-Jan-2023.  Subgroup sum is
```
avekens, line 229304: Sorry, my fault: I replaced the proof in one of my latest commits, so the date should have been 13-Jan-2024. Could you add this theorem again (with corrected date), please?

avekens (approved):

PR #3781, opened 2024-01-20
Reduce ax-13 usage by adding dv conditions

GinoGiotto:
I used a similar approach to https://github.com/metamath/set.mm/pull/3778 to remove ax-13 dependencies from the theorems mentioned below. Most of these were in the list of proofs that I would have added as *w version, but in this case it's possible (and more convenient) to just add dv conditions directly to them.

I added ALT versions for each item of the list. The main reason is that the variables involved were present in the final statements, but perhaps there are theorems where ALT versions wouldn't be really necessary, for example for [bnj1441](https://us.metamath.org/mpeuni/bnj1441.html) and [nfiund](https://us.metamath.org/mpeuni/nfiund.html), which have no current usage.

Since lemma [hbsb](https://us.metamath.org/mpeuni/hbsb.html) is part of the ax-13 section, we can't add dv conditions directly to it to reduce that dependency, therefore [hbsbw](https://us.metamath.org/mpeuni/hbsbw.html) is moved to main from my mathbox. Theorem [nfabdw](https://us.metamath.org/mpeuni/nfabdw.html) is moved to main to be used by [nfiund](https://us.metamath.org/mpeuni/nfiund.html), but it will be used more later.

Current usage of the edited theorems:

* Statement [nfrex](https://us.metamath.org/mpeuni/nfrex.html) is directly used by 71 theorems.
* Statement [cbviunv](https://us.metamath.org/mpeuni/cbviunv.html) is directly used by 69 theorems.
* Statement [nfab](https://us.metamath.org/mpeuni/nfab.html) is directly used by 56 theorems.
* Statement [cbviun](https://us.metamath.org/mpeuni/cbviun.html) is directly used by 29 theorems.
* Statement [nfiun](https://us.metamath.org/mpeuni/nfiun.html) is directly used by 22 theorems.
* Statement [nfiin](https://us.metamath.org/mpeuni/nfiin.html) is directly used by 18 theorems.
* Statement [cbviin](https://us.metamath.org/mpeuni/cbviin.html) is directly used by 11 theorems.
* Statement [nfaba1](https://us.metamath.org/mpeuni/nfaba1.html) is directly used by 8 theorems.
* Statement [cbviinv](https://us.metamath.org/mpeuni/cbviinv.html) is directly used by 8 theorems.
* Statement [hbab](https://us.metamath.org/mpeuni/hbab.html) is directly used by 3 theorems.
* Statement [nfsab](https://us.metamath.org/mpeuni/nfsab.html) is directly used by 3 theorems.
* Statement [nfrexd](https://us.metamath.org/mpeuni/nfrexd.html) is directly used by 3 theorems.
* Statement [hblem](https://us.metamath.org/mpeuni/hblem.html) is directly used by [bnj1311](https://us.metamath.org/mpeuni/bnj1311.html).
* Statement [bnj1441](https://us.metamath.org/mpeuni/bnj1441.html) is not used by anyone.
* Statement [nfiund](https://us.metamath.org/mpeuni/nfiund.html) is not used by anyone.

To make things more clear, I provide the chain shapes of these theorems:
<pre>nfabdw → nfiund
           ↑
         nfrexd → nfrex → nfiun
                           ↑
 * hbsbw → hbab → nfsab → nfab → nfaba1
     ↓      ↓              ↓
   hblem  bnj1441        nfiin

 * cbviun → cbviunv

 * cbviin → cbviinv
</pre>

This PR contains 3 independent chains. The first one has the purpose to drop ax-13 from [nfrex](https://us.metamath.org/mpeuni/nfrex.html) and [nfab](https://us.metamath.org/mpeuni/nfab.html), which are the most used ones in their chain. The other two are meant to drop ax-13 from [cbviunv](https://us.metamath.org/mpeuni/cbviunv.html) and [cbviinv](https://us.metamath.org/mpeuni/cbviinv.html).

The theorems in the chain used directly or indirectly by [nfab](https://us.metamath.org/mpeuni/nfab.html) don't really have high usage. An alternative approach could be to move [nfsabw](https://us.metamath.org/mpeuni/nfsabw.html) to main with a longer proof, although we would loose the (relavitely small) contributions of [hbsbw](https://us.metamath.org/mpeuni/hbsbw.html), [hbab](https://us.metamath.org/mpeuni/hbab.html), [nfsab](https://us.metamath.org/mpeuni/nfsab.html), [hblem](https://us.metamath.org/mpeuni/hblem.html) for reducing ax-13 usage (their recursive usage is around a few dozens of theorems if we cut off [nfab](https://us.metamath.org/mpeuni/nfab.html) from the chain).

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -79,6 +79,21 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+20-Jan-24 nfiund    nfiundg
+20-Jan-24 bnj1441   bnj1441g
+20-Jan-24 cbviinv   cbviinvg
+20-Jan-24 cbviunv   cbviunvg
+20-Jan-24 cbviin    cbviing
+20-Jan-24 cbviun    cbviung
+20-Jan-24 nfiin     nfiing
+20-Jan-24 nfiun     nfiung
+20-Jan-24 nfrex     nfrexg
+20-Jan-24 nfrexd    nfrexdg
+20-Jan-24 nfaba1    nfaba1g
+20-Jan-24 nfab      nfabg
+20-Jan-24 hblem     hblemg
+20-Jan-24 nfsab     nfsabg
+20-Jan-24 hbab      hbabg 
 17-Jan-24 mteqand   [same]      moved from SN's mathbox to main set.mm
 15-Jan-24 sseqtr4d  sseqtrrd
 14-Jan-24 sseqtr4i  sseqtrri
--- a/discouraged
+++ b/discouraged
@@ -2998,6 +2998,8 @@
 "cbncms" is used by "ubthlem2".
 "cbvexsv" is used by "onfrALTlem1".
 "cbvexsv" is used by "onfrALTlem1VD".
+"cbviinALT" is used by "cbviinvALT".
+"cbviunALT" is used by "cbviunvALT".
 "cbvmptALT" is used by "cbvmptvALT".
 "cbvmptfALT" is used by "cbvmptALT".
 "cbvopab1ALT" is used by "cbvmptfALT".
@@ -6342,6 +6344,8 @@
 "hba1-o" is used by "axc711toc7".
 "hba1-o" is used by "dvelimf-o".
 "hba1-o" is used by "nfa1-o".
+"hbabALT" is used by "bnj1441ALT".
+"hbabALT" is used by "nfsabALT".
 "hbae-o" is used by "aev-o".
 "hbae-o" is used by "dral1-o".
 "hbae-o" is used by "dral2-o".
@@ -9525,6 +9529,13 @@
 "nfa1-o" is used by "ax12eq".
 "nfa1-o" is used by "ax12v2-o".
 "nfa1-o" is used by "axc11n-16".
+"nfabALT" is used by "nfaba1ALT".
+"nfabALT" is used by "nfiinALT".
+"nfabALT" is used by "nfiunALT".
+"nfrexALT" is used by "nfiunALT".
+"nfrexdALT" is used by "nfiundALT".
+"nfrexdALT" is used by "nfrexALT".
+"nfsabALT" is used by "nfabALT".
 "nfsb2ALT" is used by "nfsb4tALT".
 "nfsb4ALT" is used by "sbco2ALT".
 "nfsb4tALT" is used by "nfsb4ALT".
@@ -13814,6 +13825,7 @@ New usage of "bnj1423" is discouraged (1 uses).
 New usage of "bnj1424" is discouraged (2 uses).
 New usage of "bnj1436" is discouraged (12 uses).
 New usage of "bnj1441" is discouraged (0 uses).
+New usage of "bnj1441ALT" is discouraged (0 uses).
 New usage of "bnj1442" is discouraged (1 uses).
 New usage of "bnj1444" is discouraged (1 uses).
 New usage of "bnj1445" is discouraged (1 uses).
@@ -14063,6 +14075,10 @@ New usage of "cbvalvOLD" is discouraged (0 uses).
 New usage of "cbveuALT" is discouraged (0 uses).
 New usage of "cbvexsv" is discouraged (2 uses).
 New usage of "cbvexvOLD" is discouraged (0 uses).
+New usage of "cbviinALT" is discouraged (1 uses).
+New usage of "cbviinvALT" is discouraged (0 uses).
+New usage of "cbviunALT" is discouraged (1 uses).
+New usage of "cbviunvALT" is discouraged (0 uses).
 New usage of "cbvmptALT" is discouraged (1 uses).
 New usage of "cbvmptfALT" is discouraged (1 uses).
 New usage of "cbvmptvALT" is discouraged (0 uses).
@@ -15385,6 +15401,7 @@ New usage of "hatomic" is discouraged (1 uses).
 New usage of "hatomici" is discouraged (6 uses).
 New usage of "hatomistici" is discouraged (1 uses).
 New usage of "hba1-o" is discouraged (8 uses).
+New usage of "hbabALT" is discouraged (2 uses).
 New usage of "hbae-o" is discouraged (4 uses).
 New usage of "hbalg" is discouraged (1 uses).
 New usage of "hbalgVD" is discouraged (0 uses).
@@ -15393,6 +15410,7 @@ New usage of "hbexg" is discouraged (0 uses).
 New usage of "hbexgVD" is discouraged (0 uses).
 New usage of "hbimpg" is discouraged (0 uses).
 New usage of "hbimpgVD" is discouraged (0 uses).
+New usage of "hblemALT" is discouraged (0 uses).
 New usage of "hbnae-o" is discouraged (3 uses).
 New usage of "hbntal" is discouraged (3 uses).
 New usage of "hbra2VD" is discouraged (0 uses).
@@ -16385,14 +16403,22 @@ New usage of "nelne2OLD" is discouraged (0 uses).
 New usage of "nf5rOLD" is discouraged (0 uses).
 New usage of "nf5riOLD" is discouraged (0 uses).
 New usage of "nfa1-o" is discouraged (4 uses).
+New usage of "nfabALT" is discouraged (3 uses).
+New usage of "nfaba1ALT" is discouraged (0 uses).
 New usage of "nfabd2OLD" is discouraged (0 uses).
 New usage of "nfabdOLD" is discouraged (0 uses).
 New usage of "nfceqiOLD" is discouraged (0 uses).
 New usage of "nfcvfOLD" is discouraged (0 uses).
 New usage of "nfequid-o" is discouraged (0 uses).
 New usage of "nfeu1ALT" is discouraged (0 uses).
+New usage of "nfiinALT" is discouraged (0 uses).
+New usage of "nfiunALT" is discouraged (0 uses).
+New usage of "nfiundALT" is discouraged (0 uses).
 New usage of "nfopdALT" is discouraged (0 uses).
+New usage of "nfrexALT" is discouraged (1 uses).
+New usage of "nfrexdALT" is discouraged (2 uses).
 New usage of "nfsab1OLD" is discouraged (0 uses).
+New usage of "nfsabALT" is discouraged (1 uses).
 New usage of "nfsb2ALT" is discouraged (1 uses).
 New usage of "nfsb4ALT" is discouraged (1 uses).
 New usage of "nfsb4tALT" is discouraged (1 uses).
@@ -18204,6 +18230,7 @@ Proof modification of "bj-xpima1snALT" is discouraged (25 steps).
 Proof modification of "bj-xpima2sn" is discouraged (23 steps).
 Proof modification of "bj-xpnzex" is discouraged (71 steps).
 Proof modification of "bj-zfauscl" is discouraged (65 steps).
+Proof modification of "bnj1441ALT" is discouraged (30 steps).
 Proof modification of "brfi1indALT" is discouraged (789 steps).
 Proof modification of "brfvidRP" is discouraged (92 steps).
 Proof modification of "c0exALT" is discouraged (15 steps).
@@ -18217,6 +18244,10 @@ Proof modification of "cbvalvOLD" is discouraged (53 steps).
 Proof modification of "cbveuALT" is discouraged (48 steps).
 Proof modification of "cbvexsv" is discouraged (29 steps).
 Proof modification of "cbvexvOLD" is discouraged (38 steps).
+Proof modification of "cbviinALT" is discouraged (64 steps).
+Proof modification of "cbviinvALT" is discouraged (13 steps).
+Proof modification of "cbviunALT" is discouraged (64 steps).
+Proof modification of "cbviunvALT" is discouraged (13 steps).
 Proof modification of "cbvmptALT" is discouraged (15 steps).
 Proof modification of "cbvmptfALT" is discouraged (166 steps).
 Proof modification of "cbvmptvALT" is discouraged (13 steps).
@@ -18785,6 +18816,7 @@ Proof modification of "gsumccatOLD" is discouraged (995 steps).
 Proof modification of "hadcomaOLD" is discouraged (37 steps).
 Proof modification of "hashge3el3dif" is discouraged (229 steps).
 Proof modification of "hba1-o" is discouraged (34 steps).
+Proof modification of "hbabALT" is discouraged (22 steps).
 Proof modification of "hbae-o" is discouraged (85 steps).
 Proof modification of "hbalg" is discouraged (31 steps).
 Proof modification of "hbalgVD" is discouraged (53 steps).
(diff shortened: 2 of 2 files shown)


icecream17 (approved):

tirix (approved):

avekens (approved):
Since @benjub 's mathbox is affected, he should have a look at this.

wlammen:
I am not quite convinced of tagging the former versions of theorems as ALT here. Our https://us.metamath.org/mpeuni/conventions.html page wants this suffix for shorter/clearer proofs of the same theorem, at the expense of axiom usage.  But here the theorems change considerably by having extra dv conditions.  What particularly hurts, is that the usage of the more general ALT version is now discouraged.  Why?  I am very much in favor of using a *w or *v suffix instead.

GinoGiotto:
> Why? I am very much in favor of using a *w or *v suffix instead.

The point is to avoid doing exactly that. The direct usage of these theorems in [ax-13-complete](https://github.com/GinoGiotto/set.mm/tree/ax-13-complete) shows that it would mean minimizing more than 100 theorems. So instead, with this PR I do basically same thing but faster and with less amount of changes.

> What particularly hurts, is that the usage of the more general ALT version is now discouraged.

I was thinking that instead of keeping original versions as ALT, maybe we could add a different suffix to them, that way we wouldn't be forced to add discouraged tags (the verifiers fail if you don't add them where an ALT suffix is present). Is there an appropriate suffix for the inverse of `*w` i.e. "Stronger version of ~ xxx without a dv condition"? (note: for consistency I would change the ALT versions of the [previus PR](https://github.com/metamath/set.mm/pull/3778) as well).

Then as I mentioned in the description of this PR, perhaps not every theorem in the above list deserves it's own permanent copy, some of the less used original versions could simply be reported as OLD instead.

wlammen:
`...shows that it would mean minimizing more than 100 theorems...`
If it is meant as a dirty trick to fool verifiers etc. temporarily into thinking everything is fine, while it is intended to fix the state in a final sweep, I am OK with that.  If it is meant to be the final state, we should discuss that before.
Maybe this was done already, and I missed it.  Could you provide a link then?

GinoGiotto:
> If it is meant as a dirty trick to fool verifiers etc. temporarily into thinking everything is fine, while it is intended to fix the state in a final sweep, I am OK with that.

What do you mean by that? I'm not intending to fool any verifier. Perhaps I didn't explain it well.

 > Maybe this was done already, and I missed it. Could you provide a link then?

The thing is, either you add *w versions and minimize hundreds of theorems, or as suggested by @tirix in https://github.com/metamath/set.mm/pull/3773#pullrequestreview-1823075033 you make the weaker versions the default. In that discussion it was originally proposed to keep the original versions as OLD. Then I was asked to keep them as ALT instead https://github.com/metamath/set.mm/pull/3773#issuecomment-1896206781, this is what lead us to the point we are right now.

So, to address your concerns my new proposal is to keep the original ones with a different suffix instead of ALT, which would represent the inverse meaning of the *w suffix, and maybe keep [bnj1441](https://us.metamath.org/mpeuni/bnj1441.html) and [nfiund](https://us.metamath.org/mpeuni/nfiund.html) as OLD, since they are never used anyway. This way I am happy because I avoid a significant minimization activity and you are happy because non-conventional ALT versions are not generated. Does this sound good to you?

wlammen:
Yes

GinoGiotto:
> Yes

What suffix would you suggest for "Stronger version of ~ xxx without a dv condition"?

wlammen:
We use the suffix g for more general theorems occasionally.

What still nags me a bit is that we do not follow our habits (so far) of providing the general theorem before the specialized one.  Is that minimizing process you speak of a one-time shot that once done will not occur further?  Or is it a repeating event, that will disturb development again and again?  In the latter case we should find a means of circumventing it.  Just a few ideas that sprang to my mind.

benjub:
What is the problem with suffixing a `v` or `w` to a label when we add a DV condition ? If the only problem is a one-time change in proofs, I think that does not justify breaking a convention which has been consistently followed up to now.

set.mm
```diff
@@ -25076,23 +25085,45 @@ yield an eliminable and weakly (that is, object-level) conservative
...
     nfsab $p |- F/ x z e. { y | ph } $=
       ( cv cab wcel nf5ri hbab nf5i ) DFACGHBABCDABEIJK $.
   $}
 
+  ${
+    $d x z $.
+    nfsabALT.1 $e |- F/ x ph $.
+    $( Alternate version of ~ nfsab , with less dv conditions, but requiring
+       ~ ax-13 .  (Contributed by Mario Carneiro, 11-Aug-2016.)
+       (Proof modification is discouraged.)  (New usage is discouraged.) $)
```
benjub, line 25122: I would keep nfsab as the standard version (with fewer DV conditions), and add the version with an additional DV condition as nfsabv, as usual, with comment "Version of nfsab with an additional disjoint variable condition, which requires fewer axioms." (or: "which does not require ~ ax -13" if it saves only that axiom).

benjub: (same with hbab and the new hbabv)

icecream17:
This isn't a one time occurrence, this happened before when I added an extra DV to cbvabv and cbvrabv

GinoGiotto:
First of all (I didn't mentioned before) this PR drops ax-13 from 106 theorems. Full axiom usage of the changes (last commit included) can be consulted here https://github.com/metamath/set.mm/commit/abc58210b2b1fcba3cef32e1a3219fbf6e5d2d5b.

I pushed a new commit renaming *ALT versions as *g versions, the approach I used goes as follows:

- The *g versions keep the original comment and original credits. Proof chains remain immutated with the original versions. Some labels in the proofs look different only because they are in the set of those theorems that I renamed with *g, but they are still the same theorems as before.

- The new versions have an additional dv conditon. They initially keep the original comment and author, then they are followed by a revised tag preceded by a description: "Added dv condition to avoid ~ ax-13, see ~ xxxg for a less restrictive version requiring more axioms." This way the user can always find the corresponding original version.

Note that *g versions are not needed in principle, they are only used between each other. The only reason I'm keeping them is because I've been asked to do so to show how it looks without additional dvs. For future proofs, the new versions should be preferred. I believe the *g versions could actually have a "new usage is discouraged" tag, because you wouldn't want to repropagate ax-13 in a future ax-13 free database.

As I mentioned before, it would take a lot more than one PR to achieve the same result with the *w versions, because we would have to shift hundreds of proofs. Note that adding dv as I did here is not always applicable.  In fact, most of the time we have to follow the long approach anyway, because usually dv conditions propagate too much. But when the advantage is clear, such as in this PR, I think we should take the opportunity to get the highest results, while making less changes as possible.

GinoGiotto:
The description of [hblem](https://us.metamath.org/mpeuni/hblem.html) declares to be a lemma for [nfcrii](https://us.metamath.org/mpeuni/nfcrii.html), but this doesn't seem to be true at all ([hblem](https://us.metamath.org/mpeuni/hblem.html) is not even indirectly used by [nfcrii](https://us.metamath.org/mpeuni/nfcrii.html)). Therefore I removed that sentence from both versions with the last commit.

set.mm
GinoGiotto: > ... and same for the other ones

If your request is specific for `nfsab` and `hbab` then it's not a problem, because they are not used that much. If your suggestion was speaking in general terms (as I interpreted it from the last sentence) then not only I'm against it in principle, but I couldn't even commit it in this PR because of how large it would become.

wlammen:
...`you wouldn't want to repropagate ax-13 in a future ax-13 free database`.
If you can dispense of ax-13 totally (i.e. replace it with ax-1 to ax-12), this is fine with me.  But if you can't: the *g variants are still valid theorems in the Metamath universe, and since they require less, they have in principal the potential for a more general theory.  For now we may collect them in some niche, no use right now, but  I am against prohibiting their application totally, even in a case that can foreseeably in an optimisation step be reproved without ax-13.

GinoGiotto:
> If you can dispense of ax-13 totally (i.e. replace it with ax-1 to ax-12), this is fine with me. But if you can't: the *g variants are still valid theorems in the Metamath universe, and since they require less, they have in principal the potential for a more general theory.

This is reasonable. We can think about it later in a different phase. For now there is still a lot of work to be done.

GinoGiotto:
> Since @benjub 's mathbox is affected, he should have a look at this.

@avekens  BJ mathbox was never affected, the bnj* series belongs to Jonathan Ben-Naim's mathbox.

set.mm
benjub: I haven't followed everything, but what I say is:

* do not change the labels of existing theorems,
* if you add a weakened form of an existing theorem obtained by adding one or several DV conditions, then use a suffix `v` or `vv` or `w` to its label.

This is what we have been doing for a long time and I do not see a reason to change.

benjub: A relabeling cannot generate that much work and it is entirely automatic:
```
MM> SET WIDTH 9999
MM> READ set.mm
MM> SAVE PROOF * / NORMAL
MM> WRITE SOURCE set.mm
MM> ERASE
[in your text editor, do a basic search and replace " old_label " --> " new_label "; note
the spaces, and of course you may need a temporary label if you want to swap two labels]
MM> SET WIDTH 79
MM> READ set.mm
MM> SAVE PROOF * / COMPRESSED
MM> WRITE SOURCE set.mm / REWRAP
```

So, no, it would not complicate things much, and no time would be wasted, or maybe a handful of minutes top.

That said, if the existing stronger theorems are not used much any more and are not interesting in themselves, they may indeed be considered the variant while the ones you introduce here are considered the main ones.

> > do not change the labels of existing theorems
>
> The original theorems are not even needed at all and I would happy to just make them as OLD, the only reason they are kept with a different name is because I was asked to do so.

Then I'm very happy someone asked not to deprecate them.

GinoGiotto: > A relabeling cannot generate that much work and it is entirely automatic:

No! The method you provided doesn't work. You have to use the minimizer. Why? Two reasons:

- First and most important: not every replacement is admitted. You have to check for dv violations, and those can create domino effects that mess up everything. The minimizer automatically checks for them, so that I don't have to worry about them. Can you imagine using your method and then do `verify proof *` and getting hundreds of dv violations as consequence? It's just impractical to work with.

- Second: I cannot provide a PR of an edited proof without dropping ax-13 to it. Or at least it would be frown upon. The thing is, most theorems depend on ax-13 by multiple routes, so if I just remove one, that theorem would still depend on ax-13 and a reviewer may ask "Why making this change at all?". So I have to be more clever about it, and in fact I'm currently doing minimizations of combinations of theorems.

> the existing stronger theorems are not used much any more

The existing theorems are not used at all, they are used only between each other. This PR nullify their usage the same way I did with the previous one https://github.com/metamath/set.mm/pull/3778.

> So, no, it would not complicate things much, and no time would be wasted, or maybe a handful of minutes top.

Of course it would. A PR with hundreds of shifted proofs is not small at all, and you know that.

benjub: If the new labels have already been agreed, then fine, but just two remarks because it could help in the future, and save time:

No, a relabeling does not require the minimizer. (To arrive at this PR, you may have used a variety of tools, but once we arrived at this PR, the difference between our two proposals is a mere relabeling.)

No, it would not take much time to review: when you have to review a huge diff which was the result of an automatic script, you do not review manually the diff: you run the same script locally and you check that the diff is zero.

wlammen:
The goal is accepted, but lofty: Removing ax-13 from theorems on a large scale.  It seems that the correct (i.e. according to our established rules) transformation is difficult, either in general, or for @GinoGiotto.  So the question is whether we can be generous for the moment, and in a final step rename all affected theorems to our standards again.  That way we have some progress, and in the end we get both: the reduced axiom usage, and labels that please the majority of us.

Of course, we must not allow the same procedure for any odd small scale improvement, but perhaps this is big enough in size to make an exception.

GinoGiotto:
> So the question is whether we can be generous for the moment, and in a final step rename all affected theorems to our standards again.

Yes. Treating them independently would be fine for me (I would still fear merge conflicts regarding [cbvmptv](https://us.metamath.org/mpeuni/cbvmptv.html), which is directly used by more than 400 theorems, but at least it's an issue that can be handled without hurry).

> Of course, we must not allow the same procedure for any odd small scale improvement

Correct. This makes sense for theorems (or to be more precise theorem chains) which are used a lot. Prioritizing these first in a simple manner is not only more practical but also faster in achieving the expected results.

avekens:
> > Since @benjub 's mathbox is affected, he should have a look at this.
> 
> @avekens BJ mathbox was never affected, the bnj* series belongs to Jonathan Ben-Naim's mathbox.

That's true, sorry for my mistake.

avekens:
About relabelling: for me it's fine, but it should be documented in changes-set.txt.

Example:

```
20-Jan-24 hbab      hbabg
```

set.mm
benjub: > Which would generate a PR so big that...

See the second of my two remarks above.

> What script are you referring to?

See the first of my two remarks above (and my previous message in this thread).

By the way: this MR is good to merge: I made suggestions in this conversation, and not blocking requests. I don't know why you got suddenly so defensive. I already said my appreciation of your work on multiple topics, and in particular on this one of reducing dependences on ax-13, and I'm saying it again: this is going in the good direction and I hope you keep going.

GinoGiotto: > By the way: this MR is good to merge: I made suggestions in this conversation, and not blocking requests. I don't know why you got suddenly so defensive. I already said my appreciation of your work on multiple topics, and in particular on this one of reducing dependences on ax-13, and I'm saying it again: this is going in the good direction and I hope you keep going.

In retrospect, maybe I overreacted. This project requires to keep track of a lot of information and I guess I perceived your comment as a threat to my workflow or as something that would have forced me to repeat some steps again (turns out, eradicating ax-13 in a collaborative environment is much harder than in a database where I have full control). But maybe my harsh response wasn't justified. I'm sorry.

For my own sanity, I would still prefer to keep renaming as a separate task, maybe I'll note it somewhere.

wlammen:
Please update changes-set.txt so that the merge conflicts are resolved.  If no one objects, I will merge this PR afterwards.

tirix:
No problem for me to merge, I already approved.

My understanding is that this project will ultimately allow to remove ax-13 from the whole of `set.mm`, except for some exploratory theorems at the beginning. This is done at the price of additional DV conditions but the statements and hypotheses of most theorems remain the same.

Then from my point of view, the theorems without ax-13 become "the new normal" and their names should not need to be longer than today. I encouraged that direction, sorry I did not also push for an open discussion on the mailing list first!

In any case that's a laudable goal and I think one everyone agrees on. I also think Norm would definitely have agreed on.

@GinoGiotto you have all my support!

tirix:
> I also think Norm would definitely have agreed on this goal.

Actually Norm mentioned [in this thread](https://groups.google.com/g/metamath/c/1wi1s6qBYqY/m/Dpjx6dqhAwAJ):

> I am a little puzzled by "but it would add around a hundred more theorems to the FOL part of [set.mm](http://set.mm/)".  **Wouldn't most of these replace existing theorems**, along with a few additional ones that more or less amount to "lemmas"?  Or do you really need 100 additional theorems to accomplish the ax-13 removal?  I understand that some dv versions would be used to eliminate ax-13 from some proofs via dummy variables, but it seems surprising we need that many.

My interpretation is that he would be a proponent of replacing the current version of the theorems with the version without ax-13.

PR #3760, opened 2024-01-13
Theorems about magmas, semigroups and monoids (continued)

avekens:
Follow-up of PR #3724: theorems moved from AV's mathbox to main as announced/discussed in #3724 and other minor adjustments

Subsection "Identity elements":
* ~grprinvlem, ~grprinvd, ~grpridd moved from subsection "Operations"
* ~lidrididd ~lidrideqd moved from AV's mathbox to main

Subsection "Iterated sums in a magma":
* ~gsumsplit1r moved from AV's mathbox to main

Subsection "Definition and basic properties of monoids":
*  ~mndbn0 moved from AV's mathbox to main
*  proofs of ~slmdbn0 , ~slmdsn0 in TA's mathbox shortened

Subsection "Iterated sums in a monoid":
* ~gsumsgrpccat moved from AV's mathbox to main
* proof of ~gsumccat shortened

Subsection "Group multiple operation":
* ~mulgnn0gsum, ~mulgnngsum moved from AV's mathbox to main

New subsection "Cyclic monoids and groups":
* ~cycsubgcl, ~cycsubgss, ~cycsubg, ~cycsubg2, ~cycsubg2cl, ~cycsubggend, ~cycsubgcld  moved from other places
* ~cycsubm ~cycsubmcl ~cycsubmel moved from AV's mathbox to main

New subsection header "Direct products (extension)":
* ~smndlsmidm, ~mndlsmidm moved from AV's mathbox to main
* proof of ~lsmidm shortened

Subsection "Abelian groups":
* new theorem ~invmod (variant of ~caovmo) added

Subsubsection "Group sum operation"
* ~gsumreidx, ~gsumxp2 moved from AV's mathbox to main
* ~gsumcom3, ~gsumcom3fi moved up from section "Matrix multiplication"

Miscellaneous:
* ~gcdmultiplez,  ~gcdmultiple moved up and proofs shortened
* Header for subsubsection "Definition and basic properties" of subsection "Simple groups" added
* Header for subsubsection "Convert operation laws using setvar variables to class notation" of subsection "Operations" added

PR #3782, opened 2024-01-21
Theorems about monoids (continued)

avekens:
In the context of PR #3724 and PR  #3760, Gérard Lang asked me to proof that a "cyclic monoid must be commutative". This is mainly done with ~ cycsubmcmn ("The set of nonnegative integer powers of an element ` A ` of a monoid forms a commutative monoid."). In principle, the proof was contained already in the proof of ~cygabl, but had to be modified to be usable for monoids. Therefore, I extracted the common part as ~cyccom and shortened ~cygabl accordingly.

Furthermore, Gérard Lang asked why it is necessary to have a commutative monoid in ~invmod. This is necessary because only right inverses are considered (as already done in the already long existing ~caovmo, which I wanted to move to PART 10  BASIC ALGEBRAIC STRUCTURES first, but then recognized that this theorem is used alredy before - in ~recmulnq) . If this theorem is transformed to (two-sided) inverses (see ~mndinvmod), then it is not neccessary that the monoid is commutative. I renamed ~invmod to ~rinvmod, and could shorten its proof using ~mndinvmod.  

Changes in detail:

Uniqueness of inverses:
* ~ mndinvmod added, ~ invmod (->rinvmod) revised 
Commutativity within sets of powers of an element of a monoid
* ~cyccom, cycsubmcom, cycsubmcmn added; proof of ~cygabl shortened

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -14490,6 +14490,7 @@ New usage of "cvr2N" is discouraged (1 uses).
 New usage of "cvrletrN" is discouraged (0 uses).
 New usage of "cvrnrefN" is discouraged (0 uses).
 New usage of "cvrval4N" is discouraged (0 uses).
+New usage of "cygablOLD" is discouraged (0 uses).
 New usage of "dalem31N" is discouraged (0 uses).
 New usage of "daraptiALT" is discouraged (0 uses).
 New usage of "dariiALT" is discouraged (0 uses).
@@ -18273,6 +18274,7 @@ Proof modification of "currysetlem" is discouraged (49 steps).
 Proof modification of "currysetlem1" is discouraged (71 steps).
 Proof modification of "currysetlem2" is discouraged (20 steps).
 Proof modification of "currysetlem3" is discouraged (82 steps).
+Proof modification of "cygablOLD" is discouraged (379 steps).
 Proof modification of "daraptiALT" is discouraged (23 steps).
 Proof modification of "dariiALT" is discouraged (19 steps).
 Proof modification of "dedtOLD" is discouraged (19 steps).


savask:
> Gérard Lang asked why it is necessary to have a commutative monoid in ~invmod. This is necessary because only right inverses are considered 

Would be nice to have an example of a monoid with several left inverses but no right inverse formalized in set.mm. I don't think the section "Examples and counterexamples for magmas, semigroups and monoids" has one.

avekens:
> > Gérard Lang asked why it is necessary to have a commutative monoid in ~invmod. This is necessary because only right inverses are considered
> 
> Would be nice to have an example of a monoid with several left inverses but no right inverse formalized in set.mm. I don't think the section "Examples and counterexamples for magmas, semigroups and monoids" has one.

I do not know if there is a published example with this requirements. Starting with a set of four elements `n a b c ` and an operation `*` so that `n` is the identity (`n*n=n, a*n=n*a=a, b*n=n*b=b, c*n=n*c=c`), and saying that `a` has two left inverses, but no right inverse (`b*a=c*a=n` and `a*x =/= n` for `x=a,b,c`), we must select results for `a*a, a*b, a*c` (not being `n`), and for `b*b, b*c, c*b , c*c`  (arbitrary) so that `*` becomes associative. I do not know, however, if such a selection is possible.

icecream17:
```
b + a = 0
b + a + a = a
a =/= a + a

If a + a = b
b + a + a = a
b + a + a = b + b
b + b = a
b + b + a = a + a = a
b + 0 = a
b = a
b = a + a
a + a = a

a + a is not b
a + a is not c by same reasoning

if a + a = 0
b + a + a = 0 + a = a
          = b + 0 = b
so a + a =/= 0
```

So a + a is not 0, a, b, c; so a counterexample must have more than 4 elements

benjub:
> > Would be nice to have an example of a monoid with several left inverses but no right inverse formalized in set.mm. I don't think the section "Examples and counterexamples for magmas, semigroups and monoids" has one.
> 
> I do not know if there is a published example with this requirements. Starting with a set of four elements `n a b c ` and an operation `*` so that `n` is the identity (`n*n=n, a*n=n*a=a, b*n=n*b=b, c*n=n*c=c`), and saying that `a` has two left inverses, but no right inverse (`b*a=c*a=n` and `a*x =/= n` for `x=a,b,c`), we must select results for `a*a, a*b, a*c` (not being `n`), and for `b*b, b*c, c*b , c*c` (arbitrary) so that `*` becomes associative. I do not know, however, if such a selection is possible.

For an example, take a monoid of endofunctions under composition and an injective non-surjective function. For instance, take the successor function in $\mathbb{N}$. It is not surjective so does not have right inverses, but it has infinitely many left inverses: the predecessor function on nonzero naturals, and zero sent to anything.

tirix (approved):

avekens:
> > > Would be nice to have an example of a monoid with several left inverses but no right inverse formalized in set.mm. I don't think the section "Examples and counterexamples for magmas, semigroups and monoids" has one.
> > 
> > 
> > I do not know if there is a published example with this requirements. Starting with a set of four elements `n a b c ` and an operation `*` so that `n` is the identity (`n*n=n, a*n=n*a=a, b*n=n*b=b, c*n=n*c=c`), and saying that `a` has two left inverses, but no right inverse (`b*a=c*a=n` and `a*x =/= n` for `x=a,b,c`), we must select results for `a*a, a*b, a*c` (not being `n`), and for `b*b, b*c, c*b , c*c` (arbitrary) so that `*` becomes associative. I do not know, however, if such a selection is possible.
> 
> For an example, take a monoid of endofunctions under composition and an injective non-surjective function. For instance, take the successor function in N. It is not surjective so does not have right inverses, but it has infinitely many left inverses: the predecessor function on nonzero naturals, and zero sent to anything.

That's a good point, I will work on it in a separate PR.

langgerard:
@savask  Consider any set A having at least two member elements equipped with the law (a,b) --> a@b=a.
This law is associative because every (non-empty, finite product is equal to its leftmost factor, so that we have:
(a@b)@c = a@b = a = a@(b@c). 
The law is not commutative because a@b= a and b@a =b are different if a and b are distinct.
Every element a is idempotent because a@a = a, and is also a left (?) unit because b@a = b.
But no element a is a right (?) unit, because this would mean that we have a:= a@b = b for every b, that is possible only in the case that the monoid A has only one unique element a (that is evidently a bilateral unit).

jkingdon:
> That's a good point, I will work on it in a separate PR.

Sounds like there is no need to delay merging this one then.

PR #3785, opened 2024-01-21
Delete and add theorems in set-mbox-ii.

expln:
`cu3addi` and `cu3add` have been deleted because they can be easily derived from `cu3addd` when needed.

Three new theorems have been added - `sqnegd`, `anpnaneq0d`, and `rexlimdvv3d`. They will be used in the proof of `3cubes` (which I will add in one of my next pull requests).

Please help me to come up with a good name for `anpnaneq0d` (if its current name is not good enough). Currently it stands for “**A** to the power of **N** **p**lus **n**egative **A** to the power of **N** **eq**uals **0**”.

Update:
`anpnaneq0d` is renamed to `negexpidd`.


set.mm
```diff
@@ -630918,37 +630918,41 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+    sqnegd $p |- ( ph -> ( -u A ^ 2 ) = ( A ^ 2 ) ) $=
+      ( cc wcel cneg c2 cexp co wceq sqneg syl ) ABDEBFGZHZIBMNIJCBKL $.
+  $}
+
+  ${
+    anpnaneq0d.1 $e |- ( ph -> A e. RR ) $.
+    anpnaneq0d.2 $e |- ( ph -> N e. NN0 ) $.
+    anpnaneq0d.3 $e |- ( ph -> -. 2 || N ) $.
+    $( The sum of a real number to the power of N and the negative of the
+       number to the power of N equals zero if N is a nonnegative even integer.
```
savask, line 630933: Should be "nonnegative **odd** integer".

expln: Fixed. Thanks.

icecream17 (approved):
I think it's fine, but maybe `negexpid` (a la https://us.metamath.org/mpeuni/negid.html ) if necessary

And this gives me some ideas to shorten oexpreposd, awesome

expln:
> I think it's fine, but maybe `negexpid` (a la https://us.metamath.org/mpeuni/negid.html ) if necessary
> 
> And this gives me some ideas to shorten oexpreposd, awesome

`negexpid` is easier to remember, so I renamed. Thanks.

set.mm
```diff
@@ -630918,37 +630918,41 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+
+  ${
+    $d B z $.  $d A y z $.  $d ph x y z $.  $d ch x y z $.
+    rexlimdvv3d.1 $e |- ( ph ->
+      ( ( x e. A /\ ( y e. B /\ z e. C ) ) -> ( ps -> ch ) ) ) $.
+    $( An extended version of ~ rexlimdvv to include three set variables.
+       (Contributed by Igor Ieskov, 21-Jan-2024.) $)
+    rexlimdvv3d $p |- ( ph -> ( E. x e. A E. y e. B E. z e. C ps -> ch ) ) $=
+      ( wrex cv wcel wa wi expdimp rexlimdvv rexlimdva ) ABFIKEHKCDGADLGMZNBCEF
+      HIASELHMFLIMNBCOJPQR $.
```
tirix, line 630953: The `v` was doubled in `rexlimdvv` because there are two set variables.
Here you don't need to both add the `3` and keep the doubled `v`, you can either use a triple `v` or keep the `3` with a single `v`. I would prefer the latter.

tirix: I usually prefer grouping left first, as in `( ( x e. A /\ y e. B ) /\ z e. C )`, but that's a matter of personal preference.
You could also have simply used the 3-conjunct `( x e. A /\ y e. B /\ z e. C )`.

tirix (approved):

savask (approved):

set.mm
expln: Both items have been fixed. The theorem was renamed to `rexlimdv3d` and now it uses `( x e. A /\ y e. B /\ z e. C )`.

PR #3786, opened 2024-01-21
Explanation of DV conditions

benjub:
Commit messages say it all.

### DIFF
--- a/mmset.raw.html
+++ b/mmset.raw.html
@@ -671,7 +671,7 @@ yielding the final answer (<I>b</I> - 7) + <I>c</I> = <I>c</I> +
 the direct replacement of a variable with an expression.  The more complex
 proper substitution of <A HREF="#traditional">traditional logic</A> is a
 derived concept in Metamath, broken down into multiple primitive steps.
-<A HREF="#distinct">Distinct variable</A> provisos, which accompany certain
+<A HREF="#distinct">Distinct variable</A> conditions, which accompany certain
 axioms and are inherited by theorems, forbid unsound substitutions.]
 </FONT>
 </P>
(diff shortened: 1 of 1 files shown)


tirix (approved):

mmset.raw.html
```diff
@@ -3683,69 +3683,69 @@
 variable pair is officially called a "disjoint variable pair" in the Metamath
 specification.  We present the rule as three separate cases above for clarity.
 In the set theory database file, set.mm, we adopt the convention that at least
-one setvar variable always appears in a distinct variable pair, so these are
+one setvar variable always appear in a distinct variable pair, so these are
```
avekens, line 3686: What's wrong with "one setvar variable always appears"?

avekens (commented):
only one minor remark.

mmset.raw.html
benjub: I think subjonctive sounds more natural, but I guess it can be either way.

benjub: (For example: "we adopt the convention that it be blue", not "that it is blue".)

tirix: Benoît seems to be correct, from a grammar point of view.
Maybe native speakers can confirm whether this is natural for them?

icecream17: I'm pretty sure "appears" is grammatically correct, since if one removes "always" then it fits perfectly
image

tirix: I had to try with the full sentence, and other cases.
image

jkingdon: This native speaker suggests "at least one setvar variable should always appear in a distinct variable pair" (word "should" added, "appear" in place of "appears"). Or some other rewording. I wouldn't rely too heavily on the subjunctive in English. In this sentence, "appear" without "should" doesn't really read to me as a subjunctive (at least, not on first reading) and my general understanding is that the subjunctive in English is much less used than in romance languages.

jkingdon (approved):

icecream17 (approved):
i agree that it is non-blocking; good stuff

mmset.raw.html
benjub: Done in https://github.com/metamath/set.mm/pull/3786/commits/346c93b756c19bb58432ae0ebec5995a30660992.

avekens (approved):
Becomes #3754 obsolete now by these changes? @david-a-wheeler what do you think?

PR #3787, opened 2024-01-22
Maximal Ideals

tirix:
Definition and theorems about maximal ideals ported from Jeff Madsen's mathbox.
I also added a chapter about the [sumset](https://en.wikipedia.org/wiki/Sumset) of two sets, which is useful for dealing with operations on ideals.


set.mm
```diff
@@ -472151,6 +472186,240 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+$(
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+  Sumset / Minkowski sum
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+
+  The sumset (also called the Minkowski sum) of two subsets ` A ` and ` B ` ,
+  is defined to be the set of all sums of an element from ` A ` with an element
+  from ` B ` .
+
+  For an operation ` .+ ` This can be expressed as ` ( .+ " ( A X. B ) ) ` .
```
savask, line 472198: Can you comment on this new notation for a sumset vs the already existing `LSSum`? The latter works for any group, and can be used with a more readable (in my opinion) notation
```
.(+) = ( LSSum ` G )
...
( A .(+) B )
```
rather than ` ( .+ " ( A X. B ) ) `.

set.mm
```diff
@@ -472151,6 +472186,240 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+
+  ${
+    $d .+^ x y $.  $d B x y $.  $d E x y $.  $d F x y $.  $d Z x y $.
+    $d ph x y $.
+    elgrpsumset.1 $e |- B = ( Base ` G ) $.
+    elgrpsumset.2 $e |- .+ = ( +g ` G ) $.
+    elgrpsumset.3 $e |- .+^ = ( +f ` G ) $.
+    elgrpsumset.4 $e |- ( ph -> E C_ B ) $.
+    elgrpsumset.5 $e |- ( ph -> F C_ B ) $.
+    $( Membership to a sumset for a group operation.  (Contributed by Thierry
```
savask, line 472217: Maybe "Membership __in__ a sumset", and similarly everywhere.

set.mm
```diff
@@ -472151,6 +472186,240 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+  ${
+    $d .X. x y $.  $d .x. x y $.  $d B x y $.  $d E x y $.  $d F x y $.
+    $d R x y $.  $d Z x y $.  $d ph x y $.
+    elringsumset.1 $e |- B = ( Base ` R ) $.
+    elringsumset.2 $e |- .x. = ( .r ` R ) $.
+    elringsumset.3 $e |- G = ( mulGrp ` R ) $.
+    elringsumset.4 $e |- .X. = ( +f ` G ) $.
+    elringsumset.5 $e |- ( ph -> E C_ B ) $.
+    elringsumset.6 $e |- ( ph -> F C_ B ) $.
+    $( Membership to a sumset for a ring operation.  (Contributed by Thierry
```
savask, line 472297: I think the usage of the term "sumset" here is a bit confusing, since `( .X. " ( E X. F ) )` is actually a product set. I think it will more clear to spell things out as "Membership in a product of two subsets of a ring".

set.mm
```diff
@@ -472151,6 +472186,240 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+  ${
+    $d .X. x y $.  $d A x $.  $d B x y $.  $d K x y $.  $d R x y $.
+    $d X x y $.  $d h ph $.  $d h x $.  $d ph x y $.
+    sumsetsnpridl.1 $e |- B = ( Base ` R ) $.
+    sumsetsnpridl.2 $e |- G = ( mulGrp ` R ) $.
+    sumsetsnpridl.3 $e |- .X. = ( +f ` G ) $.
+    sumsetsnpridl.4 $e |- K = ( RSpan ` R ) $.
+    sumsetsnpridl.5 $e |- ( ph -> R e. Ring ) $.
+    sumsetsnpridl.6 $e |- ( ph -> X e. B ) $.
+    $( The sumset of a ring with a single element is the element's principal
```
savask, line 472327: Here, in my opinion, the term "sumset" is even more confusing :-) Maybe something like "The product of the ring with a single element is equal to the principal ideal generated by that element".

set.mm
```diff
@@ -472151,6 +472186,240 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+  ${
+    $d .+^ f x y z $.  $d .0. i $.  $d B f i x y $.  $d I f i x y z $.
+    $d J f i x y z $.  $d K f i x y z $.  $d R f i x y $.  $d f i ph x y z $.
+    sumsetidl.1 $e |- B = ( Base ` R ) $.
+    sumsetidl.3 $e |- .+^ = ( +f ` R ) $.
+    sumsetidl.4 $e |- K = ( RSpan ` R ) $.
+    sumsetidl.5 $e |- ( ph -> R e. Ring ) $.
+    sumsetidl.6 $e |- ( ph -> I e. ( LIdeal ` R ) ) $.
+    sumsetidl.7 $e |- ( ph -> J e. ( LIdeal ` R ) ) $.
+    $( The sumset sum of two ideals is the ideal generated by their union.
```
savask, line 472357: The "sumset sum" is excessive, I think it's better to write "The sum of two ideals ...".

set.mm
```diff
@@ -472429,6 +472712,198 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+      $( One is not contained in any maximal ideal.  (Contributed by Jeff
+         Madsen, 17-Jun-2011.)  (Revised by Thierry Arnoux, 19-Jan-2024.) $)
+      mxidln1 $p |- ( ( R e. Ring /\ M e. ( MaxIdeal ` R ) ) -> -. .1. e. M )
+      $=
+        ( crg wcel cmxidl cfv wa wn mxidlnr clidl wceq wb mxidlidl eqid lidl1el
+        wne syldan necon3bbid mpbird ) BGHZDBIJHZKZCDHZLDATABDEMUFUGDAUDUEDBNJZ
+        HUGDAOPABDEQABUHCDUHREFSUAUBUC $.
+    $}
+
+    $( A ring with a maximal ideal is a nonzero ring.  (Contributed by Jeff
```
savask, line 472802: If I recall correctly, every nonzero ring (with 1) has a maximal ideal, so maybe it's worth strengthening this theorem.

savask (commented):
Great work! I'm especially happy to see some theorems on sumsets. I think the theory of sumsets is very worth formalizing in set.mm (after agreeing on definitions, see my comment on `LSSum`). It would be great to see some results from additive combinatorics formalized, like [Ruzsa's triangle inequality](https://en.wikipedia.org/wiki/Ruzsa_triangle_inequality).

tirix:
I'm converting this PR to draft until it is reworked using `LSSum`, then I will ask for a review again.
In the mean time it's Ok if other PR like e.g. #3789 are merged, I can adapt (this happened before).

set.mm
```diff
@@ -472151,6 +472186,240 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+  For an operation ` .+ ` This can be expressed as ` ( .+ " ( A X. B ) ) ` .
+  This notation is concise, so no new notation needs to be introduced.
+
+  General theorems exist about this construct (see e.g. ~ ovelimab ), however
+  they are not directly applicable to groups, where the group operation is not
+  guaranteed to be a function.  The theorems in this section help in that case.
+
+  The sumset operation can be used for both group (additive) operations and
+  ring (multiplicative) operations.
+$)
```
benjub, line 472207:
```suggestion
  ring (multiplicative) operations.

$)
```

set.mm
```diff
@@ -472151,6 +472186,240 @@ commutative monoid (=vectors) together with a semiring (=scalars) and a
...
+
+  For an operation ` .+ ` This can be expressed as ` ( .+ " ( A X. B ) ) ` .
+  This notation is concise, so no new notation needs to be introduced.
+
+  General theorems exist about this construct (see e.g. ~ ovelimab ), however
+  they are not directly applicable to groups, where the group operation is not
+  guaranteed to be a function.  The theorems in this section help in that case.
+
+  The sumset operation can be used for both group (additive) operations and
+  ring (multiplicative) operations.
```
benjub, line 472206: What does that mean ?

tirix: I meant it's call "**sum**set" but actually we can use it for any group operations, not only "sums".

set.mm
tirix: Ok, this is finally resolved with the last commit!

set.mm
tirix: Done with last commit.

set.mm
tirix: Done with last commit.

set.mm
tirix: Done with last commit.

set.mm
tirix: Gérard also suggested to prove a similar statement, that for any proper ideal $I$ of a ring $R$, there is a maximal ideal of $R$ containing $I$.
I'll try to include that in a subsequent PR (using Zorn's Lemma), and it should then be possible to derive [Krull's theorem](https://en.wikipedia.org/wiki/Krull%27s_theorem), which is what you stated.

savask (approved):
Nice changes, as always! The definition of a spectrum makes me wonder if you have an eye on algebraic geometry.

I think it's also worth mentioning that some earlier theorems in your pull request, like `lsmidllsp`, work for `Rng`-s (non-unital rings). The benefit here is that ideals in an `Rng` are `Rng`-s themselves, and that simplifies arguments in some cases.

tirix:
> Nice changes, as always! The definition of a spectrum makes me wonder if you have an eye on algebraic geometry.

Thanks! Yes, I had been looking a some algebraic geometry at the time I originally wrote this PR (in January), and I thought that the spectrum of a ring was a relatively low-hanging fruit to pick.

> I think it's also worth mentioning that some earlier theorems in your pull request, like lsmidllsp, work for Rng-s (non-unital rings). The benefit here is that ideals in an Rng are Rng-s themselves, and that simplifies arguments in some cases.

Indeed. Since this PR has been pending for a long time, maybe it's better to just merge as-is, and refine later?

set.mm
tirix: See #3912 for a follow-up on this.

PR #3789, opened 2024-01-22
Rename syl6sseqr to sseqtrrdi

jkingdon:
This is in set.mm and iset.mm


wlammen:
The diff looks ok  and would usually be approved by me.  In the current situation with so many open pull request, I am bit afraid that merging it may cause a lot of conflicts.

jkingdon:
> The diff looks ok and would usually be approved by me. In the current situation with so many open pull request, I am bit afraid that merging it may cause a lot of conflicts.

My general thought process is that merging pull requests quickly when they are ready to go tends to reduce merge conflicts.

But in this specific case, if there is any pull request you particularly think should be merged first, we could do that. I just merged one and looked over the rest (which seemed to me to be in various stages of not being ready), but I'm willing to take requests especially if another pull request seems like it can be merged soon.

wlammen:
Perhaps #3781 first?

jkingdon:
> Perhaps #3781 first?

Sounds good, we'll see how things go with that one.

wlammen (commented):
Approved

PR #3792, opened 2024-01-23
more number properties without ax-mulcom

icecream17:
Changes:
+ move and shorten nelb
+ prove re0m0e0 readdid2 sn-addid2 remul02 etc.

Comments:
I was going to use nelb to prove a version of 0nnn, but I ended up messing up and decided to only prove sn-0ne2

readdid2 is trivial to prove from sn-00id but readdid1 takes a complicated proof

~~**Draft**: https://us.metamath.org/mpeuni/readdid2addid1d.html and https://us.metamath.org/mpeuni/reneg0addid1.html are inconsistent with addid1~~

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -79,6 +79,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+23-Jan-24 nelb      [same]      moved from TA's mathbox to main set.mm
 17-Jan-24 mteqand   [same]      moved from SN's mathbox to main set.mm
 15-Jan-24 sseqtr4d  sseqtrrd
 14-Jan-24 sseqtr4i  sseqtrri
--- a/discouraged
+++ b/discouraged
@@ -16378,6 +16378,7 @@ New usage of "n2dvds3OLD" is discouraged (0 uses).
 New usage of "naecoms-o" is discouraged (1 uses).
 New usage of "natded" is discouraged (0 uses).
 New usage of "negexsr" is discouraged (0 uses).
+New usage of "nelbOLD" is discouraged (0 uses).
 New usage of "nelne1OLD" is discouraged (0 uses).
 New usage of "nelne2OLD" is discouraged (0 uses).
 New usage of "nf5rOLD" is discouraged (0 uses).
@@ -18956,6 +18957,7 @@ Proof modification of "n0lpligALT" is discouraged (74 steps).
 Proof modification of "n2dvds1OLD" is discouraged (37 steps).
 Proof modification of "n2dvds3OLD" is discouraged (40 steps).
 Proof modification of "naecoms-o" is discouraged (19 steps).
+Proof modification of "nelbOLD" is discouraged (81 steps).
 Proof modification of "nelne1OLD" is discouraged (26 steps).
 Proof modification of "nelne2OLD" is discouraged (26 steps).
 Proof modification of "nf5rOLD" is discouraged (22 steps).


benjub (approved):

set.mm
```diff
@@ -29355,6 +29355,12 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
     risset $p |- ( A e. B <-> E. x e. B x = A ) $=
       ( cv wcel wceq wa wex wrex exancom df-rex dfclel 3bitr4ri ) ADZCEZNBFZGAH
       POGAHPACIBCEOPAJPACKABCLM $.
+
+    $( A definition of ` -. A e. B ` .  (Contributed by Thierry Arnoux,
+       20-Nov-2023.)  (Proof shortened by SN, 23-Jan-2024.) $)
+    nelb $p |- ( -. A e. B <-> A. x e. B x =/= A ) $=
+      ( cv wne wral wcel wn wceq wrex df-ne ralbii ralnex bitri risset xchbinxr
+      bicomi ) ADZBEZACFZBCGZHTRBIZACJZUATUBHZACFUCHSUDACRBKLUBACMNABCOPQ $.
```
wlammen, line 29363: This is an example of where the current definition of df-ral introduces some maybe unwanted trickery.  Neither A nor B need to be independent of x, so what is expressed in such a situation?  It would be clearer in my eyes to replace x with a z that is disjoint from B, and A.  Or do you need this (for me) hard to understand case where x is allowed to appears in all classes?

tirix:
Just a note that this is a follow-up of the project described in #3713, with the idea that `ax-mulcom` might not be an independent axiom.

set.mm
tirix: Actually `x` needs to be independent of `A` and `B`, at least in the original proof:
```
  ${
    $d A x $.  $d B x $.
    $( A definition of ` -. A e. B ` .  (Contributed by Thierry Arnoux,
       20-Nov-2023.) $)
    nelb $p |- ( -. A e. B <-> A. x e. B x =/= A ) $=
      ( cv wceq wn wral wcel wi wal wne df-ral df-ne ralbii wa wex dfclel alnex
      notbii imnan ancom bitr2i albii 3bitr2i 3bitr4ri ) ADZBEZFZACGUFCHZUHIZAJ
      ZUFBKZACGBCHZFZUHACLULUHACUFBMNUNUGUIOZAPZFUOFZAJUKUMUPABCQSUOARUQUJAUJUI
      UGOZFUQUIUGTURUOUIUGUASUBUCUDUE $.
  $}
```
The same conditions appear just before `risset`, and in this PR `nelb` is moved to the same frame as `risset`, so it's still the case in the shortened version.

tirix: The distinct variable condition is actually visible just below before what becomes `nelbOLD`

tirix (approved):

set.mm
wlammen: ok, yes the dv conditions are placed in front of risset.

avekens (approved):

savask:
@icecream17 That's a really cool project! Maybe this remark will be useful to you: there exist ordered noncommutative division rings (skew Laurent series over complex numbers, for example). After adding an imaginary unit (`ax-i2m1` and `ax-cnre`) you should get a ring satisfying all the axioms besides `axsup` (and `ax-mulcom`, of course). I don't know if the ordering of this ring is complete, if not, then this would indicate that `axsup` is required to prove `ax-mulcom`.

tirix:
Interesting!

So the question boils down to: Are there example of ordered, complete, non-commutative division rings which are not $\mathbb {R}$? IMO it's worth asking e.g. on [MathExchange](https://math.stackexchange.com/questions/1645969/bbb-r-bbb-c-bbb-h-are-the-only-complete-normed-division-rings).

savask:
> So the question boils down to: Are there example of ordered, complete, non-commutative division rings which are not R?

One should probably mention that in our case the _ordering_ is complete (sup exists), while one often speaks of a "complete ring" if it is a complete metric space.

If we fantasize a little, maybe the proof of `ax-mulcom` should go like this:
1) Prove that numbers of the form 0, 1, 1+1, 1+1+1, ... -1, -1-1, ... form a commutative ring (integers)
2) Prove that the integers together with inverses also generate a commutative ring (rationals)
3) Prove the Archimedean property [arch](https://us.metamath.org/mpeuni/arch.html) and that every element of our ring can be approximated by rationals.

Commutativity should follow from 3).

As far as I understand, the difficulty is that we don't even assume the addition to be commutative (but maybe icecream17 already overcame this).

tirix:
Commutativity for the natural numbers [follow from associativity and closure, using induction](https://us.metamath.org/mpeuni/nnaddcom.html).

icecream17:
(real numbers)
if addition is commutative between A and B and A and C, then it is also commutative between

A and B + C
A and B - C (C + A + B-C = C + B-C + A, and readdcan)

so all integer multiples of any specific real is commutative

savask:
Then what about this proof idea. Suppose $a > 0$ and $b > 0$ do not commute, so without loss of generality $ab > ba$. Multiply it by $a$ inverse from the right to get $aba^{-1} > b$. If you multiply it by $a$ from the left and $a^{-1}$ from the right, you get $a^2 b a^{-2} > aba^{-1} > b$. Hence for any natural $n$ you have $a^{n+1} b a^{-(n+1)} > a^n b a^{-n} $.

Set $d = aba^{-1} - b > 0$. It follows from the Archimedean property that there exists some natural number $m$ such that $d > 1/m > 0$. Hence $a^{n+1} b a^{-(n+1)} > 1/m + a^n b a^{-n}$ (here we assume that $1/m$ commutes with all reals), therefore $a^n b a^{-n} > n/m$ for all $n$.

By Archimedean property applied to $b$, there exists a natural number $N > b$. Then $N = a^n N a^{-n} > a^n b a^{-n}$, hence $N > n/m$ for all $n$, and this is a contradiction.

tirix:
Great! 
Obviously using associativity with $(-1)\cdot a \cdot b$, the choice of $a > 0$ and $b > 0$ is justified.
The fact that that all natural numbers $n$ commute with all reals should be easily obtainable by induction. 
Then we have $(1/m)a = (1/m)a \cdot m(1/m) = (1/m)m \cdot a(1/m) = a(1/m)$, so the assumption that $1/m$ commutes with all reals is also valid.

icecream17:
~~I don't currently see how natural numbers commuting with reals is obtainable by induction
The base case would be 1 + r = r + 1, for any real r~~

EDIT: The base case is actually $1 \cdot r = r \cdot 1$ per below

But assuming this, I think everything checks out

tirix:
Above I was thinking about mutliplicative commutativity. That should just be distributivity of the addition: $a \cdot (n+1) = a \cdot n + a \cdot 1 = n \cdot a + 1 \cdot a = (n+1) \cdot a$ for $a \in \mathbb {R}$ and $n \in \mathbb {N}$.

icecream17:
Oops. Well, we don't currently have $(a + b)c = ac + bc$, only $c(a + b) = ca + cb$

We currently have $a \cdot 1 = a$ per ax-1rid, <s>but not $1 \cdot a = a$</s> (multiplication with 0 commutes, so the induction can start with 0 if needed)

However both of these seem potentially very possible...

----
We have right inverse so say (ax = 1, xy = 1, yw = 1)
axyw = a1w = aw
axyw = 11 = 1

aw = 1
yw = 1

aw = yw
a = y

So a left inverse is a right inverse

savask:
> I don't currently see how natural numbers commuting with reals is obtainable by induction

I have indeed used that rational numbers commute with reals with respect to addition :-)

We know that our `RR` with `+` is a left-ordered group (this is given by `axltadd`, i.e. $a < b$ and $0 < c$ implies $c+a < c+b$). The sup axiom should give us that it's Archimedean, but this should be checked thoroughly. Assuming that, it can be shown that addition (of reals) is commutative without invoking axioms involving multiplication.

There's a paper of Conrad "Right-ordered groups" which shows that a left-ordered Archimedean group is also right-ordered (i.e. $a < b$ and $0 < c$ implies $a+c < b+c$). The argument goes as follows, take some $x, y > 0$. There exists some natural number $n$ such that $x < n \cdot y$; here $n \cdot y = y + y + \dots + y$. Take the smallest such $n$ and suppose that $x > y + x$. Then $n \cdot y > y + x$ and $(n-1) \cdot y > y$ which contradicts the minimality of $n$. Hence $x < y + x$ and this can be easily seen to be equivalent to $a+c < b+c$ if one takes $x = c$ and $y = (-a) + b$.

Then there's an old theorem of Hölder which says that a left- and right-ordered Archimedean group is isomorphic to a subgroup of the real line, and hence is commutative. I think it shouldn't be very hard to take only the "commutativity" part of his proof and adapt it to metamath.

Of course, all of this rests on the Archimedean property, it should be proved without `ax-mulcom` first (if possible).

tirix:
> There's a paper of Conrad "Right-ordered groups" which shows that a left-ordered Archimedean group is also right-ordered

That's interesting! I have formalized the theorem showing that left- and right-ordered Archimedean group are abelian, see [~archiabl](https://metamath.tirix.org/mpests/archiabl). 
Does this mean that any left- **_or_** right-ordered Archimedean group is abelian?

savask:
> I have formalized the theorem showing that left- and right-ordered Archimedean group are abelian, see [~archiabl](https://metamath.tirix.org/mpests/archiabl).

Funny enough, it relies on `ax-mulcom`, but that must be just a technicality.

>  Does this mean that any left- or right-ordered Archimedean group is abelian?

Yes, that's correct.

tirix:
> Funny enough, it relies on `ax-mulcom`, but that must be just a technicality.

It appears to come from [`~mulgdir`](https://us.metamath.org/mpeuni/mulgdir.html). It is a statement about group "power", so it involves operations in integers, as in $a^n \cdot a^m = a^{m+n}$.

Actually even the closure of addition in the integers, [`~zaddcl`](https://us.metamath.org/mpeuni/zaddcl.html), requires `~ax-mulcom`!

PR #3794, opened 2024-01-26
Add axpre-suploc to iset.mm

jkingdon:
This is suplocsr but expressed in terms of RR rather than R.

Includes ax-pre-suploc which is a restatement using $a .

Includes lemma axpre-suploclemres

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -266,6 +266,7 @@ New usage of "axpre-lttrn" is discouraged (0 uses).
 New usage of "axpre-ltwlin" is discouraged (0 uses).
 New usage of "axpre-mulext" is discouraged (0 uses).
 New usage of "axpre-mulgt0" is discouraged (0 uses).
+New usage of "axpre-suploc" is discouraged (0 uses).
 New usage of "axprecex" is discouraged (2 uses).
 New usage of "axresscn" is discouraged (4 uses).
 New usage of "axrnegex" is discouraged (0 uses).
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -5580,12 +5580,14 @@ but "not equal to zero" would need to be changed to
 <TD>~ axpre-ltirr , ~ axpre-ltwlin , ~ ax-pre-ltirr , ~ ax-pre-ltwlin </TD>
 </TR>
 
+<tr>
+  <td>axpre-sup , ax-pre-sup</td>
+  <td>~ axpre-suploc , ~ ax-pre-suploc</td>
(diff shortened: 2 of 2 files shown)


tirix (approved):

icecream17 (approved):

iset.mm
```diff
@@ -81452,6 +81452,104 @@ natural numbers (notably, ~ axcaucvg ).  (Contributed by Jim Kingdon,
...
+      DYPRWAWLVRXQXRWIXCVPXSUUFUYLUYOQBUWESUXCYSUYLUUEUYOUXCYRUYKCFUXCYQUWLUXDW
+      SWBUXCUUDUYNCSUXCYTUYMUUCYOUWEYPRWAXTWBYAWLYBVS $.
+  $}
+
+  ${
+    $d A a b c d w $.  $d A a b c d x y z $.
+    $( An inhabited, bounded-above, located set of reals has a supremum.
+
+       Locatedness here means that given ` x < y ` , either there is an element
+       of ` A ` greater than ` x ` , or ` y ` is an upper bound.
```
benjub, line 81522: @jkingdon We don't know what `A` is. I'd phrase it:
```
A set of reals ` A ` is located if given any reals ... upper bound of ` A ` .
```

iset.mm
```diff
@@ -81701,6 +81799,28 @@ A Cauchy sequence (as defined here, which has a rate convergence built
...
+  ${
+    $d A x y z $.
+    $( An inhabited, bounded-above, located set of reals has a supremum.
+
+       Locatedness here means that given ` x < y ` , either there is an element
+       of ` A ` greater than ` x ` , or ` y ` is an upper bound.
+
+       Although this and ~ ax-caucvg are both completeness properties,
+       countable choice would probably be needed to derive this from
+       ~ ax-caucvg .
```
benjub, line 81811: But this Dedekind completeness implies Cauchy completeness in IZF (even CZF), correct ? (else, I'd have to correct https://us.metamath.org/ileuni/ax-ddkcomp.html)

iset.mm
jkingdon: I think we can just change A to "the set". I'll include that in my next pull request.

Agreed that what is there now isn't very good wording.

iset.mm
jkingdon: I think so, although most of the sources don't exactly address this expressed in quite this way, so I've been a little bit weasel wording statements of this sort unless we prove something.

PR #3800, opened 2024-01-28
Fix some latexdef's.

benjub:
* consistently begin a new line for new definitions
* add some `\mathbin` for more consistent spacing
* upright omega
* calligraphic $\mathcal{C}^n$ for continuity classes


wlammen (approved):

benjub:
(Let me add that the upright omega is used when it does not stand for a variable. Similarly, italic C is avoided because it is not a variable.)

PR #3810, opened 2024-01-31
janitorial stuff

wlammen:
1. equs5av is a weak version of sb56, from which it can be easily derived.  So move it next to sb56.  It remains to be seen whether it adds value to the database at all.
2. Improve navigation by adding links to sb56 and from equs5a.  This level of cross-linking is standard in our database for years now (see spv/spvv for a random example 5 yrs ago) , but was abandoned in recent contributions.  It would be nice to see some (perhaps slow but steady) rework in that direction, too.


tirix (approved):

GinoGiotto:
> It remains to be seen whether it adds value to the database at all.

Historically `equs5av` was added by me to fix an issue in BJ mathbox https://github.com/metamath/set.mm/pull/3686#discussion_r1425357716 https://github.com/metamath/set.mm/pull/3697, but in retrospect maybe it was unnecessary because its axiom usage is the same of `sb56`, and `bj-equs45fv` can be proven with a one step proof directly from it `bj-equs45fv $p |- ( E. x ( x = y /\ ph ) <-> A. x ( x = y -> ph ) ) $= ( sb56 ) ABCE $.`

The original `equs5a` is directly used by  `equs45f` and `sb4aALT` (and recursively used by only 10 statements), so I'm skeptical its *v version would be of great use.

Therefore I'm not against deleting `equs5av`, but we should consult @benjub first because this would affect his mathbox.

PR #3816, opened 2024-02-04
Add Dedekind cut theorem to iset.mm

jkingdon:
This says that given a Dedekind cut, there is a unique real number which separates the lower and upper cuts.

As in many sources which don't assume excluded middle, the defining property of a Dedekind cut is that it is inhabited (bounded), rounded, disjoint, and located.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -6126,8 +6126,16 @@ favor of theorems in deduction form.</TD>
   <TD>See fimaxre</TD>
 </TR>
 
+<tr>
+  <td>sup2</td>
+  <td><i>none</i></td>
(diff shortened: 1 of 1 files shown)


tirix (approved):
Nice!
Why would it not work on rational numbers?

jkingdon:
> Nice! Why would it not work on rational numbers?

It would, but both kinds are potentially useful and the real number one is what I will need for the Monotonic intermediate value theorem.

https://us.metamath.org/mpeuni/dedekind.html is another example of a theorem of this sort where the upper and lower cut are made up of reals.

avekens (approved):

PR #3827, opened 2024-02-13
ltp1, evlsgsumadd

icecream17:
evlsgsumadd is the corresponding version of
evls1gsumadd

prove ltp1 without ax-mulcom

#### comments
copying from evls1 makes things a lot easier

mathbox changes cont'd from #3792


set.mm
```diff
@@ -258381,6 +258381,34 @@ This function maps an element of the formal polynomial algebra (with
...
+    evlsgsumadd.p $e |- P = ( S ^s ( K ^m I ) ) $.
+    evlsgsumadd.k $e |- K = ( Base ` S ) $.
+    evlsgsumadd.b $e |- B = ( Base ` W ) $.
+    evlsgsumadd.i $e |- ( ph -> I e. V ) $.
+    evlsgsumadd.s $e |- ( ph -> S e. CRing ) $.
+    evlsgsumadd.r $e |- ( ph -> R e. ( SubRing ` S ) ) $.
+    evlsgsumadd.y $e |- ( ( ph /\ x e. N ) -> Y e. B ) $.
+    evlsgsumadd.n $e |- ( ph -> N C_ NN0 ) $.
+    evlsgsumadd.f $e |- ( ph -> ( x e. N |-> Y ) finSupp .0. ) $.
+    $( Polynomial evaluation maps (additive) group sums to group sums.
```
wlammen, line 258399: It seems hypotheses of theorems in this section are sort of unordered.  Without knowing much about general conventions in algebra, I would prefer have them ordered such that one introducing a term is mentioned before (or consequently the other way round) another using that term.  So the primitive ones (like evlsgsumadd.i, evlsgsumadd.s, evlsgsumadd.n) come first, followed by evlsgsumadd.r, evlsgsumadd.k, and so on.

icecream17: I just went with what evls1gsumadd did, with the exception of moving `K` right after `P`

Based on looking at various theorems around, it seems like:

+ "variable definitions" go at the start
+ deductive hypotheses go at the end
+ structure definitions go before their components

I do agree that there is still some ambiguity with the ordering, even if this was an explicit convention somewhere

avekens: I think the current ordering is fine:
* "variable definitions" are at the start
* followed by deductive hypotheses with membership
* followed by other deductive hypotheses

avekens (approved):

PR #3837, opened 2024-02-23
Intermediate value theorem for strictly monotonic functions

jkingdon:
As described in [Bauer], Theorem 5.5, iset.mm can prove the intermediate value theorem in the case of a strictly monotonic function.

This pull request includes both the increasing the decreasing cases. The proof follows the proof in [Bauer] pretty closely.

### DIFF
--- a/mm_100.html
+++ b/mm_100.html
@@ -560,7 +560,11 @@ by others.</li>
 
 <!-- 14th added to list -->
 <li><a name="79">79</a>.  The Intermediate Value Theorem (<a
-href="mpeuni/ivth.html">ivth</a>, by Paul Chapman, 2008-01-22)</li>
+href="mpeuni/ivth.html">ivth</a>, by Paul Chapman, 2008-01-22).
+The intuitionistic logic explorer (iset.mm) database also includes <a
+href="https://us.metamath.org/ileuni/ivthinc.html">ivthinc</a> which
+is the theorem for a strictly monotonic function
+(by Jim Kingdon, 2024-02-20).</li>
 
 <!-- 18th added to list -->
 <li><a name="80">80</a>.  Fundamental Theorem of Arithmetic (<a
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10881,6 +10881,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>ivth</td>
+  <td>~ ivthinc</td>
+</tr>
+
+<tr>
   <td>df-limc</td>
   <td>~ df-limced</td>
   <td>~ df-limced is adapted from ellimc3 in set.mm</td>
@@ -11229,9 +11234,8 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>reeff1olem</td>
   <td><i>none</i></td>
   <td>The set.mm proof depends on ivth (the Intermediate
-  Value Theorem).  Apparently, the Monotonic Intermediate
-  Value Theorem (Theorem 5.5 of [Bauer], p. 494) would
-  suffice if we can prove that.</td>
+  Value Theorem).  Apparently, ~ ivthinc would suffice
+  so this may be provable.</td>
 </tr>
 
 <tr>
@@ -11272,9 +11276,9 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>pilem3</td>
   <td><i>none</i></td>
   <td>The set.mm proof depends on ivth2 (the Intermediate
-  Value Theorem).  Apparently, the Monotonic Intermediate
-  Value Theorem (Theorem 5.5 of [Bauer], p. 494) could
-  replace it if we can prove that.  Additionally, the set.mm
+  Value Theorem).  We should be able to use ~ ivthinc or
+  a variation thereof if we show that sine is strictly
+  monotonic between 2 and 4 .  Additionally, the set.mm
   proof uses infrelb , infrecl , pilem2 , and infregelb .</td>
 </tr>
 
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -10886,6 +10886,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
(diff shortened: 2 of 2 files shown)


iset.mm
```diff
@@ -142195,40 +142195,412 @@ Dedekind cut is that it is inhabited (bounded), rounded, disjoint, and
...
-      TUMXOVTUMYBYMVAYJXFYJXMURXFAUUOYHXTUUPVNZYIYDXTUUKWAVOWBYJXOYJXMURXOUURYI
-      YEXTUUMWAVOWBXFXOWCVMWDYJYKYBYJYGXKXHXTTZTZTZYKYJAYFXKTZTZUUSTZUVAYJUVCXH
-      TZXTTUVDYIUVEXTYIAUVBXHTZTUVEYHUVFAYHYFXKXHTZTUVFXLUVGYFXHXKWEWFYFXKXHWIW
-      GWFAUVBXHWIWGWHUVCXHXTWIWJUVDYGXKTZUUSTUVAUVCUVHUUSUVHUVCAYFXKWIWKWHYGXKU
-      USWIWJWJUUTYAYGUUTUVGXTTYAXKXHXTWIUVGXLXTXKXHWEWHWLWFWJWMWNWOWPXLXTBUAXMY
-      BXHXQXKXSYBXGXPHFXFXOXEUCWQWRYBXJXRGEXFXOXIUCWTWRWSXAXBXLBXMXCXD $.
+    ${
+      $d A q r x y $.  $d B q r x y $.  $d L q r x y $.  $d U q r x y $.
+      $d ph q r x y $.
+      $( Lemma for ~ dedekindicc .  The result, except that we merely show it
```
avekens, line 142201: "The result" of what?

iset.mm
```diff
@@ -142195,40 +142195,412 @@ Dedekind cut is that it is inhabited (bounded), rounded, disjoint, and
...
+        VDYGXKTZUUSTUVAUVCUVHUUSUVHUVCAYFXKWGWKWIYGXKUUSWGWJWJUUTYAYGUUTUVGXTTY
+        AXKXHXTWGUVGXLXTXKXHWEWIWLWFWJWMWNWOWPXLXTBUAXMYBXHXQXKXSYBXGXPHFXFXOXE
+        UCWQWSYBXJXRGEXFXOXIUCWRWSWTXAXBXLBXMXCXD $.
+    $}
+
+    ${
+      $d A q r x $.  $d B q r x $.  $d L a q r x $.  $d U b q r x $.
+      $d ph q r x $.
+      $( A Dedekind cut identifies a unique real number.  Similar to ~ df-inp
+         except that the the Dedekind cut is formed by sets of reals (rather
```
avekens, line 142238: "the the"

iset.mm
jkingdon: Hmm, interesting, the overall context sorta answers that but when I look at the sentence more closely I realize that "the result" refers both to the theorem dedekindicc and to the value `x`, in different parts of it.

I guess we can go with "Same as ~ dedekindicc , except that we merely show ` x ` to be an element of ` ( A [,] B ) `"

iset.mm
jkingdon: Fixed.

avekens (approved):
Great! Another Metamath 100 proof in iset.mm!

jkingdon:
> Great! Another Metamath 100 proof in iset.mm!

Indeed! There are other variations of the intermediate value theorem which could be proved in iset.mm (or, perhaps, even impossibility proofs for those which can't be proved) but I think this is close enough to declare victory in a Metamath 100 context.

avekens:
If I counted correctly, this is the 14th Metamath 100 proof in iset.mm. Maybe iset.mm could/should be added to the official "Formalizing 100 Theorems" page (http://www.cs.ru.nl/%7Efreek/100/). There it would be at position 10 in the ranking. @david-a-wheeler could you contact Freek for this?

jkingdon:
> If I counted correctly, this is the 14th Metamath 100 proof in iset.mm. Maybe iset.mm could/should be added to the official "Formalizing 100 Theorems" page (http://www.cs.ru.nl/%7Efreek/100/). There it would be at position 10 in the ranking. @david-a-wheeler could you contact Freek for this?

It looks like there is a convention on that page of using italics for constructive proofs. Perhaps that would fit here? I didn't see any discussion there about, say, lean3 versus lean4, or Isabelle/ZF versus Isabelle/HOL, so I'm not sure how fine grained he'd want to go. Of course, Freek is the decider on such matters so perhaps it is better just to ask than to have me speculate.

david-a-wheeler:
Freek & I discussed this years ago. At least at the time, Freek didn't want to go that fine-grained into exactly how different tools were used to create proofs. Many tools have different versions or options (e.g., lean3 vs. lean4, etc.). He just notes if there is *any* formal proof for a given tool, then links elsewhere for the details.

So consistent with that direction, if there is *any* proof in Metamath, it goes into the Metamath 100 list, regardless of the database used. Usually it's set.mm, though sometimes iset.mm is also there. If there's more than one proof, all are listed. For example, here's the entry for <https://us.metamath.org/mm_100.html#66>:

> 66. Sum of a Geometric Series (geoser, by Norman Megill, 2006-05-09). The intuitionistic logic explorer (iset.mm) database includes this proof as geoserap (added 2022-10-24).

Freek does have a convention that "Formalizations of constructive proofs are in italics". So if there was already a non-constructive proof in Metamath, and we add a constructive one, we *should* ask Freek to italicize that entry. That might be worth double-checking; I may have forgotten to tell Freek about some of them (sorry if I did!). If we tell Freek "please italicize Metamath entries X, Y, and Z", I'm sure he'd be happy to do so.

jkingdon:
> if there is _any_ proof in Metamath, it goes into the Metamath 100 list, regardless of the database used.

Sounds good. In particular, this is a better place than Freek's list to include nuances like "this is the case of strictly monotonic functions" or "this is a negative result". It would be odd to see an italicized link in Freek's list for the Intermediate Value Theorem or Schoeder-Bernstein, when those are both known to be unprovable constructively.

Issue #3840, opened 2024-02-23
The sine function is decreasing over [2,4]


jkingdon:
That is, we would like to prove:

```
  $( The sine function is decreasing between two and four.
     (Contributed by Jim Kingdon, 21-Feb-2024.) $)
  sindec $p |- ( ( A e. ( 2 [,] 4 ) /\ B e. ( 2 [,] 4 ) /\ A < B )
      -> ( sin ` B ) < ( sin ` A ) ) $=
    ? $.
```

using only the theorems now in iset.mm (which include double angle formulas, half angle formulas, formulas for the sine of a sum or difference, etc).

My thought process was that `B - A` is in the range (0,2] and so you'd think that would help us apply theorems like https://us.metamath.org/ileuni/sin02gt0.html . However, unless I missed something, I didn't see how to just expand ``( sin ` ( A + ( B - A ) ) )`` using `sinadd` and have everything just work easily.

Oh and in case the context helps, the larger project is to intuitionize https://us.metamath.org/mpeuni/pilem3.html (or some similar lemma) using the strict monotonic intermediate value theorem from #3837

digama0:
The original definition was to use the first positive root of cos(x) instead of the first positive root of sin(x), which is a bit annoyingly far away for these estimates. Nevertheless, it is possible to do using double angle formulas etc:

* $\cos x$ is negative for $2 < x < 4$, because $\cos(2 + 2x) = \cos 2 \cos^2x - 2 \sin 2 \cos x \sin x - \cos 2 \sin^2 x$ and all terms are negative
* $\sin b<\sin a$ when $2 &lt; a < b < 4$ because if we let $x=\frac{a+b}2$ and $y=\frac{b-a}2$ then $2 &lt; x<4$ and $0 &lt; y<1$ and $\sin a-\sin b=\sin(x+y)-\sin(x-y)=2\cos x\sin y$ which is negative.

You can probably make the bounds weak inequalities, but it would be useful to strengthen cos01gt0 to $-1\le x \le 1$ for that.

jkingdon:
>  $\cos(2 + 2x) = \cos 2 \cos^2x - 2 \sin 2 \cos x \sin x - \cos 2 \sin^2 x$ and all terms are negative

This doesn't seem to be correct as written and I'm not sure how to fix it.

Consider the third term. $\cos 2$ is negative, $\sin^2 x$ is positive, and this term has a negative sign in front of it so the net contribution of this term is positive, not negative.

[...]

PR #3847, opened 2024-02-27
Add sin24dec to iset.mm

jkingdon:
Show that sine is decreasing from two to four.

Includes lemmas sin24declemx24 , sin24declemy01 , sin24declemsub , and sin24declemres .

This is not quite all of #3840  - it does not yet include the part that cosine is negative between 2 and 4. But by including that as a hypothesis, it does show that's the only remaining piece.


iset.mm
```diff
@@ -150551,6 +150551,92 @@ Limited Principle of Omniscience (LPO) implies excluded middle, so we
...
+        RAFWSUPUSZWTVAAWHWFWIWGQAEFPOZMNZWHWFAXADMAXADCPOZRUCOZWRPOZDAEXDFWRPAE
+        WKXDKAWJXCRUCACDWNWOVBVCVDZFWRVJALVEZVFADSTZCSTZXEDVJZWOWNXHXIVGZXJXDWR
+        QOZCVJZDCVHZVIVKVLVMAESTZFSTZXBWFVJWPWSEFVNVKVOAEFQOZMNZWIWGAXQCMAXQXLC
+        AEXDFWRQXFXGVFAXHXIXMWOWNXKXJXMXNVPVKVLVMAXOXPXRWGVJWPWSEFVQVKVOVFAWEWT
+        VRVS $.
+
+      $d X t $.
+
+      $( Lemma for ~ sin24dec .  (Contributed by Jim Kingdon, 26-Feb-2024.) $)
+      sin24declemres $p |- ( ph -> ( sin ` B ) < ( sin ` A ) ) $=
```
avekens, line 150616: What is the meaning of the suffix "res" of the label here?

jkingdon: It stands for "result".

There's often a need to have a lemma which is the same as the overall theorem but has some additional hypotheses which are used for convenience (here, especially `X = ( ( A + B ) / 2 )` and `Y = ( ( B - A ) / 2 )`). I've been naming these with names ending in `lemres` fairly widely but if there is another convention I could switch to that. I'd sort of rather avoid numbered lemmas, not because I see a huge value in making lemma names descriptive, but because it seems to lead to a fair bit of renumbering or `lem2a` style names or the like.

avekens: According to our conventions, "res" stands for "restriction" (e.g. ~opelres), that's why I was a little confused. 
Maybe sin24declemlt (lt for "less than") would be clearer.

jkingdon: `sin24declemlt` it is. Changed.

avekens (approved):

PR #3853, opened 2024-03-01
Minimize ax-13 usage 18

GinoGiotto:
Part of [groups.google.com/g/metamath/c/OB2_9sYgDfA](https://groups.google.com/g/metamath/c/OB2_9sYgDfA).

Continuation of https://github.com/metamath/set.mm/pull/3796.

This PR drops ax-13 from 117 theorems. Full axiom usage here: https://github.com/metamath/set.mm/commit/6fcc67284a11ef196d19617ef8b70c0f5b9bed9c


avekens (approved):
only @tirix's mathbox concerned, so maybe he should have a look at it.

GinoGiotto:
A few days have passed. I'm afraid many more mathboxes of currently inactive users will be affected. I'd say to include these changes in the maintenance edits as described in the conventions and moving forward.

From https://us.metamath.org/mpeuni/mathbox.html:

> Even though in a sense your mathbox belongs to you, it is still part of the shared body of knowledge contained in set.mm, and occasionally other people may make maintenance edits to your mathbox for things like keeping it synchronized with the rest of set.mm, reducing proof lengths, moving your theorems to the main part of set.mm when needed, and fixing typos or other errors.

This case would fall into "keeping it synchronized with the rest of set.mm", which is the global avoidance of an axiom dependency.

wlammen:
Given the nature of change, and the time passed waiting for @tirix opinion, I am in favor of merging this now.  If no one objects, I am going to do this by tomorrow latest.  Note that such a merge can be reverted quite easily should an objection arrive lately.
@tirix is a special case here, because he is usually very active and responsive.  Other mathboxes can be handled according to the guidelines, I think, with our usual reaction time of 24 hours or so.

wlammen (approved):

tirix:
Sorry I've taken time to answer. I'm travelling and I certainly hope to have more time for Metamath in the future.
I cannot check the detail of the changes but I know they're of the same nature as the previous PR.
Ok for me.

tirix (approved):

PR #3861, opened 2024-03-10
Add exmidonfin to iset.mm

### LINKED ISSUE #2203: Show that finite ordinals are natural numbers in iset.mm

jkingdon:
We are unable to prove that a finite ordinal is a natural number.  The proof is a formalization of an informal [proof by Andrew W Swan posted on mathstodon.xyz](https://mathstodon.xyz/@aws/112067779907659794).

Includes lemma exmidonfinlem

Fixes #2203

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -4013,10 +4013,16 @@ by ~ exmidsbth </TD>
 </TR>
 
 <TR>
-<TD>onomeneq , onfin , onfin2</TD>
+<TD>onomeneq , onfin</TD>
 <TD><I>none</I></TD>
-<TD>The set.mm proofs rely on excluded middle</TD>
+<TD>Not possible because they would imply onfin2</TD>
 </TR>
+
+<tr>
+  <td>onfin2</td>
+  <td><i>none</i></td>
+  <td>Implies excluded middle as shown as ~ exmidonfin</td>
+</tr>
 
 <TR>
 <TD>nnsdomo , sucdom2 , sucdom , 0sdom1dom , sdom1</TD>
@@ -4722,9 +4728,8 @@ this implies excluded middle</TD>
 <TR>
   <TD>ficardom</TD>
   <TD><I>none</I></TD>
-  <TD>Both the set.mm proof, and perhaps some possible alternative
-  proofs, rely on onomeneq and perhaps other theorems we don't have
-  currently.</TD>
+  <TD>Presumably not possible for the same reasons as
+  onfin2</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -69544,6 +69545,49 @@ elements or fails to hold (is equal to ` (/) ` ) for some element.
...
+      UVLYGYIMYHLZUVLYGXNYIUVLUVPUVNYDYHMPXQYGYGBMYDUVOYGUPUQQXPXRAXSVFXT $.
+  $}
+
+  ${
+    $d x z $.
+    $( If a finite ordinal is a natural number, excluded middle follows.  The
+       converse should also be provable as shown in the Metamath Proof
+       Explorer.  (Contributed by Andrew W Swan and Jim Kingdon,
+       9-Mar-2024.) $)
+    exmidonfin $p |- ( _om = ( On i^i Fin ) -> EXMID ) $=
```
benjub, line 69586: Looking at the informal proof, it looks like `|- ( ( On i^i Fin ) C_ _om -> EXMID )` ?  If that's correct, then I think that statement is clearer. And I presume that  `|- _om C_ ( On i^i Fin )` ? It's probably already in iset, I haven't looked; there should be a link to it from this theorem.

jkingdon: Yeah, it is worth clearing up the comment on this point. I've removed the word "converse" in favor of explicitly talking about excluded middle. And I have linked to the theorems you describe about a natural number being a finite ordinal ( https://us.metamath.org/ileuni/nnfi.html and https://us.metamath.org/ileuni/nnon.html ).

As for the statement of the theorem, it is chosen to match https://us.metamath.org/mpeuni/onfin2.html

avekens (approved):
great stuff!

jkingdon:
> great stuff!

Absolutely and I'll give Andrew W Swan most of the credit, as I doubt I would have come up with this proof in a zillion years.

Having said that, it was nice that the proof formalized so easily. And it does answer a long outstanding question.

PR #3864, opened 2024-03-11
Remove sin24dec from iset.mm

jkingdon:
The most obvious problem is that this depends on a hypothesis, that cosine is negative between 2 and 4, that we don't have a proof for. Since we proved `pire` and related theorems via another mechanism (see #3840 ), we don't need this any more.  If we do bring back something similar to this theorem, or related ones, it will be easier to do it in terms of pi anyway.

Includes lemmas sin24declemx24 , sin24declemy01 , sin24declemsub , and sin24declemlt .


icecream17 (approved):
linking #3863

PR #3868, opened 2024-03-13
Rename syl6eleqr to eleqtrrdi one more place

jkingdon:
This was a semantic merge conflict between two pull requests in flight at the same time.


wlammen:
How is it possible that this is not detected by the checks?

wlammen (approved):

jkingdon:
> How is it possible that this is not detected by the checks?

It was detected by the checks that run on `develop`. Those checks sent me an email which is what led me to make this pull request fixing it.

As for why it wasn't detected by the checks that run on pull requests, the checks passed on https://github.com/metamath/set.mm/pull/3865 (which renamed `syl6eleqr`), and on https://github.com/metamath/set.mm/pull/3863 (which added another usage of `syl6eleqr`).

When it comes time to merge (for example when you click the github button to merge a pull request), git will notice if both sides of the merge change the same line and mark it as a conflict. But in this case, the two pull requests didn't change the same line they were just incompatible. That's what the "semantic" in "semantic merge conflict" means.

There are ways to prevent this situation (merge queues or mandatory rebases before merging) but unless it starts happening more often than it has been, they aren't worth the trouble (in my opinion).

PR #3877, opened 2024-03-17
preimages, decompositions of function into a surjective and an injective function

avekens:
Inspired by Gérard Lang, I proved that an arbitrary function can be decomposed into a surjective and an injective function in two different ways. The variant using the images as domain of the injective function was very easy, the other variant using the preimages as domain of the injective function was not so easy, at least to prove it formally. A lot of auxiliary theorems were necessary to achieve this.

For more information see section header comment of section "Preimages of function values" in my mathbox.

Details:

Main:
* ~iuneqconst added
* ~uniexd  moved from GS' mathbox

AV's mathbox:
* auxiliary theorems for (pre)images
* theorems for the set of all preimages of function values
* decomposition of function into a surjective and an injective function with preimages as domain of the injective function
* decomposition of function into a surjective and an injective function with images as domain of the injective function

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -110,6 +110,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+23-Mar-24 funfvima2d [same]     moved from SP's mathbox to main set.mm
 17-Mar-24 uniexd    [same]      moved from GS's mathbox to main set.mm
 16-Mar-24 syl6eleq  eleqtrdi    compare to eleqtri or eleqtrd
 10-Mar-24 syl6eleqr eleqtrrdi   compare to eleqtrri or eleqtrrd
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -111,6 +111,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+17-Mar-24 uniexd    [same]      moved from GS's mathbox to main set.mm
 10-Mar-24 syl6eleqr eleqtrrdi   compare to eleqtrri or eleqtrrd
 29-Feb-24 syl6ss    sstrdi      compare to sstri or sstrd
 27-Feb-24 syl6sseq  sseqtrdi    compare to sseqtri or sseqtrd
--- a/discouraged
+++ b/discouraged
@@ -15270,6 +15270,7 @@ New usage of "funcringcsetclem7ALTV" is discouraged (1 uses).
 New usage of "funcringcsetclem8ALTV" is discouraged (1 uses).
 New usage of "funcringcsetclem9ALTV" is discouraged (1 uses).
 New usage of "funcrngcsetcALT" is discouraged (0 uses).
+New usage of "fundcmpsurinjALT" is discouraged (0 uses).
 New usage of "fvimacnvALT" is discouraged (0 uses).
 New usage of "gcdmultipleOLD" is discouraged (0 uses).
 New usage of "gcdmultiplezOLD" is discouraged (0 uses).
@@ -18767,6 +18768,7 @@ Proof modification of "frege98" is discouraged (116 steps).
 Proof modification of "frgrwopreglem5ALT" is discouraged (519 steps).
 Proof modification of "fsplitOLD" is discouraged (234 steps).
 Proof modification of "funcrngcsetcALT" is discouraged (765 steps).
+Proof modification of "fundcmpsurinjALT" is discouraged (221 steps).
 Proof modification of "fvilbdRP" is discouraged (27 steps).
 Proof modification of "fvimacnvALT" is discouraged (102 steps).
 Proof modification of "gcdmultipleOLD" is discouraged (348 steps).


set.mm
```diff
@@ -732651,6 +732663,410 @@ The modulo (remainder) operation - extension
...
+  ${
+    $d F x y $.  $d S x y $.  $d X x y $.
+    $( The union of the image of a subset ` S ` of the domain of a Function
+       with elements having the same function value is the function value at
+       one of the elements of ` S ` .  (Contributed by AV, 5-Mar-2024.) $)
+    uniimafveqt $p |- ( ( F : A --> B /\ S C_ A /\ X e. S )
+                       -> ( A. x e. S ( F ` x ) = ( F ` X )
+                          -> U. ( F " S ) = ( F ` X ) ) ) $=
+      ( vy wf wss wcel w3a cv cfv wceq wral cima cuni wa ciun wfun 3ad2ant1 syl
+      ffun adantr funiunfv simp3 fveqeq2 cbvralv biimpi fveq2 iuneqconst syl2an
```
GinoGiotto, line 732731: you can use [cbvralvw](https://us.metamath.org/mpeuni/cbvralvw.html) instead of [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) here, which will let you avoid ax-13 without adding any extra dv conditions:

<pre>
  ${
    $d F x y $.  $d S x y $.  $d X x y $.
    $( The union of the image of a subset ` S ` of the domain of a Function
       with elements having the same function value is the function value at
       one of the elements of ` S ` .  (Contributed by AV, 5-Mar-2024.) $)
    uniimafveqt $p |- ( ( F : A --> B /\ S C_ A /\ X e. S )
                       -> ( A. x e. S ( F ` x ) = ( F ` X )
                          -> U. ( F " S ) = ( F ` X ) ) ) $=
     ( vy wf wss wcel w3a cv cfv wceq wral cima cuni wa ciun wfun 3ad2ant1 syl
      ffun adantr funiunfv simp3 cbvralvw biimpi fveq2 iuneqconst syl2an eqtr3d
      fveqeq2 ex ) BCEHZDBIZFDJZKZALZEMFEMZNZADOZEDPQZUTNURVBRZGDGLZEMZSZVCUTVD
      ETZVGVCNURVHVBUOUPVHUQBCEUCUAUDGDEUEUBURUQVFUTNZGDOZVGUTNVBUOUPUQUFVBVJVA
      VIAGDUSVEUTEUMUGUHGDVFUTFVEFEUIUJUKULUN $.
$}
</pre>

Actually, I think [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) should be discouraged, and we should invite people to use [cbvralvw](https://us.metamath.org/mpeuni/cbvralvw.html) instead.

set.mm
```diff
@@ -732651,6 +732663,410 @@ The modulo (remainder) operation - extension
...
+        CDEFGUHHRUFUIAUHCUDUHUEFSTUAUB $.
+
+      $d P x y $.
+      $( The image of an element of the preimage of a function value is the
+         singleton constisting of the function value at one of its elements.
+         (Contributed by AV, 5-Mar-2024.) $)
+      imaelsetpreimafv $p |- ( ( F Fn A /\ S e. P /\ X e. S )
+                                -> ( F " S ) = { ( F ` X ) } ) $=
+        ( vy wcel w3a cv cfv csn cima wceq wrex wa 3adant3 3ad2ant1 fveq2 sneqd
+        wfn ccnv fvelsetpreimafv imaeq2d eqeq2d cbvrexv sylibr crn cin 3ad2ant3
```
GinoGiotto, line 732879: Same here, use [cbvrexvw](https://us.metamath.org/mpeuni/cbvrexvw.html) instead of [cbvrexv](https://us.metamath.org/mpeuni/cbvrexvw.html) to avoid ax-13.
<pre>
        ( vy wcel w3a cv cfv csn cima wceq wrex wa 3adant3 3ad2ant1 fveq2 sneqd
        wfn ccnv fvelsetpreimafv imaeq2d eqeq2d cbvrexvw sylibr imaeq2 3ad2ant3
        crn cin fnfun funimacnv syl elsetpreimafvbi wi wss fnfvelrn snssd df-ss
        wfun sylib simp3 eqtrd 3expib sylbid imp 3eqtrd rexlimdv3a mpd ) FCUCZE
        DJZGEJZKZEFUDZILZFMZNZOZPZIEQZFEOZGFMZNZPZVMVNWCVOVMVNREVQALZFMZNZOZPZA
        EQWCABCDEFHUEWBWLIAEVRWHPZWAWKEWMVTWJVQWMVSWIVRWHFUAUBUFUGUHUISVPWBWGIE
        VPVREJZWBKWDFWAOZVTFULZUMZWFWBVPWDWOPWNEWAFUJUKVPWNWOWQPZWBVMVNWRVOVMFV
        CWRCFUNVTFUOUPTTVPWNWQWFPZWBVPWNWSVPWNVRCJZVSWEPZRZWSABCDEFGVRHUQVMVNXB
        WSURVOVMWTXAWSVMWTXAKZWQVTWFVMWTWQVTPZXAVMWTRZVTWPUSXDXEVSWPCVRFUTVAVTW
        PVBVDSXCVSWEVMWTXAVEUBVFVGTVHVISVJVKVL $.
</pre>

set.mm
```diff
@@ -732651,6 +732663,410 @@ The modulo (remainder) operation - extension
...
+  $}
+
+  ${
+    $d A g h p x y z $.  $d B g h p x y z $.  $d F g h p x y z $.  $d V x y $.
+    $( Every function ` F : A --> B ` can be decomposed into a surjective and
+       an injective function.  (Contributed by AV, 13-Mar-2024.) $)
+    fundcmpsurinj $p |- ( ( F : A --> B /\ A e. V ) -> E. g E. h E. p
+                   ( g : A -onto-> p /\ h : p -1-1-> B /\ F = ( h o. g ) ) ) $=
+      ( vz vx vy wcel cv wfo wf1 wceq w3a wex cfv csn cima wf wa ccom ccnv wrex
+      cab cvv abrexexg adantl fveq2 sneqd imaeq2d cbvrexv fundcmpsurinjpreimafv
```
GinoGiotto, line 733020: Same here again, there is no tradeoff in using [cbvrexvw](https://us.metamath.org/mpeuni/cbvrexvw.html) instead of [cbvrexv](https://us.metamath.org/mpeuni/cbvrexv.html).
<pre>
      ( vz vx vy wcel cv wfo wf1 wceq w3a wex cfv csn cima wf wa ccom ccnv wrex
      cab cvv abrexexg adantl fveq2 sneqd eqeq2d cbvrexvw fundcmpsurinjpreimafv
      imaeq2d abbii foeq3 f1eq2 3anbi12d 2exbidv elabd exrot3 sylib ) ABEUAZAFK
      ZUBZAGLZCLZMZVGBDLZNZEVJVHUCOZPZDQCQZGQVMGQDQCQVFVNAHLZEUDZILZERZSZTZOZIA
      UEZHUFZVHMZWCBVJNZVLPZDQCQGWCVEWCUGKVDIHAVTFUHUIJHABWCCDEFWBVOVPJLZERZSZT
      ZOZJAUEHWAWKIJAVQWGOZVTWJVOWLVSWIVPWLVRWHVQWGEUJUKUOULUMUPUNVGWCOZVMWFCDW
      MVIWDVKWEVLVGWCAVHUQVGWCBVJURUSUTVAVMGCDVBVC $.
</pre>

As a rule of thumb, you'll basically never need the stronger versions.

set.mm
jkingdon: I like the idea of making this a discouragement tag if we expect people to follow this practice generally.

avekens: Done - I do not know if this was discussed already before (maybe in Google group: https://groups.google.com/g/metamath/c/OB2_9sYgDfA.?), but I would also endorse to tag cbvralv and cbvrexv (are there more?) as "discouraged".

set.mm
avekens: Done

set.mm
avekens: Done

avekens:
In addition to the new theorems, I replaced cbvrexv by cbvrexvw also in fiunlem.

set.mm
wlammen: The discouraged tag is very prohibitive.  The kind of problem given here appears very often: alrimi vs alrimih, nfsb vs nfsbv and so on, leading a very long (and long standing) list.  The idea behind any of them is that more distinct variables allow for a smaller axiom footprint.  Or: more axioms can provide a more general theorem, which is also a valuable goal.
Here we have the situation that the dv conditions already stated allow for the narrower theorem. What you need is a kind of pick list in your proof designer, that displays close variants of a theorem, and where you easily find the best choice.  Without such an assistant all you can do is improve on documentation.  It would be helpful if you see an overview of variants in a particular location, maybe a doc only web page like "mathbox", but dedicated to cbvral* versions.

tirix: @GinoGiotto has spent a lot of effort to remove references to ax-13. 
If the [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) variant is not discouraged, then it's like trying to fill a leaky bucket: references to ax-13 will always come back. So I'd also be in favor of marking [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) as discouraged in profit of [cbvralvw](https://us.metamath.org/mpeuni/cbvralvw.html), instead of checking every new incoming PR.

Said differently, if both theorems are applicable, which would be a reason to use [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) over [cbvralvw](https://us.metamath.org/mpeuni/cbvralvw.html)?

wlammen: `if both theorems are applicable, which would be a reason to use [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) over [cbvralvw](https://us.metamath.org/mpeuni/cbvralvw.html)`
There is no reason to use cbvralv without any advantage.  I looked at a random subset of its applications and found no instance, where it was not replaceable.  So maybe, cbvralv has no other value than stating that one dv condition in theory is unnecessary (but in practice no one cares).
But this is not an isolated problem.  In this PR alone we have another candidate cbvrexv, where we can start the discussion over again. And far more need such considerations, not alone in respect to ax-13, but other axioms as well.  So simply discouraging anything but the variant with most dv restrictions is not a viable option, in my eyes.
As always, you need to look out for optimizations yourself, and within a mathbox you  are free the freedom to just sketch a proof.  Only when you move sth to main, some care is in order.
I myself put a theorem wl-cbvalsbi into my mathbox, with unnecessary ax-10, ax-11 and ax-12 references in its proof, only to improve on it later.
That's the way work is done up to now.
Given that an unnecessary reference to ax-13 is a magnitude greater than to, say, ax-11, what do you think of extending the $j annotation to mark a deliberate use of ax-13? $j need ax-13 $. for example.  And a verifier that checks the simultaneous presence of ax-13 in the axiom list and this annotation.

GinoGiotto: > Done - I do not know if this was discussed already before (maybe in Google group: [groups.google.com/g/metamath/c/OB2_9sYgDfA](https://groups.google.com/g/metamath/c/OB2_9sYgDfA).?)

This hasn't been discussed yet, but looking at the discussion above it seems this topic deserves its own issue.

Given that ax-13 usage has recently reached an all time low, I would like to have some verification system that it will stay low, not only in the near future, but possibily in the years to come. If we don't do anything about it, I can assure you ax-13 usage will rise again. I've been collecting data about ax-13 usage for a while now. The lesson I've derived is: if you don't systematically check for something, undesired trends will show up, even if small or temporary (and they are usually temporary only because there is someone that checks for it afterwards, which should not be taken for granted).

I'm not particularly attached to the discouragement system to check for this metric, it's just the easiest tool it is currently available for us. If someone decides to show up and build a tool just for this problem, I'm not against it, on the contrary. But it has to be automatic and systematic, human review is not enough.

I don't want to give the impression that using ax-13 is prohibited. In the end, axioms are there to be used. I want to avoid the scenario in which ax-13 is used without awareness. So rather than a prohibition issue this is a communication issue. We need to find a way to communicate the user "hey, it seems you're trying to build your proof using ax-13. Did you know that about 97% of theorems in the database don't need it? Perhaps yours doesn't need it too."

Wolf Lammen is right to say that this problem does not only affects [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) or [cbvrexv](https://us.metamath.org/mpeuni/cbvrexv.html), this affects dozens of theorems, so either we are down to add about a hundred or so of discouragement tags (personally I don't have a problem with that), or we build something else. The worst decision in my opinion is to leave the state of the art as it is.

GinoGiotto:
> In addition to the new theorems, I replaced cbvrexv by cbvrexvw also in fiunlem.

Just above [fiunlem](https://us.metamath.org/mpeuni/fiunlem.html) there are already [fiunlemw](https://us.metamath.org/mpeuni/fiunlemw.html) , [fiunw](https://us.metamath.org/mpeuni/fiunw.html) , [f1iunw](https://us.metamath.org/mpeuni/f1iunw.html) which are the correspoinding versions without ax-13. I'd suggest to replace your proofs with mine in the default versions, copy the dv conditions and delete my versions. Keep the original comments and credits (yours and mario), my edit is so small anyway that doesn't require to change them.

set.mm
wlammen: `if you don't systematically check for something, undesired trends will show up, even if small or temporary`
It will take some time, sure, but maybe after a while people will routinely carry out checks of this kind?  There is evidence for this in the history of Metamath, which has seen some evolvement with deprecating patterns.

tirix (approved):
Nice!

avekens:
> > In addition to the new theorems, I replaced cbvrexv by cbvrexvw also in fiunlem.
> 
> Just above [fiunlem](https://us.metamath.org/mpeuni/fiunlem.html) there are already [fiunlemw](https://us.metamath.org/mpeuni/fiunlemw.html) , [fiunw](https://us.metamath.org/mpeuni/fiunw.html) , [f1iunw](https://us.metamath.org/mpeuni/f1iunw.html) which are the correspoinding versions without ax-13. I'd suggest to replace your proofs with mine in the default versions, copy the dv conditions and delete my versions. Keep the original comments and credits (yours and mario), my edit is so small anyway that doesn't require to change them.

Done. I think we should stop additional cleanups about ax-13 in this PR now, and should continue the discussion how to avoid new usages of ax-13 in the future somewhere else. Maybe an issue can/should be opend for this: @GinoGiotto would you open it, please, and move the corresponding discussions from here to there?

GinoGiotto:
> Done. I think we should stop additional cleanups about ax-13 in this PR now, and should continue the discussion how to avoid new usages of ax-13 in the future somewhere else. Maybe an issue can/should be opend for this: @GinoGiotto would you open it, please, and move the corresponding discussions from here to there?

I opened an issue in https://github.com/metamath/set.mm/issues/3879.

avekens:
As Gérard Lang observed, the mapping of the set of all preimages of values of a function onto the range of the function is bijective (see ~ imasetpreimafvbij ). By this, it can be shown that every function ` F : A --> B ` can be decomposed into a surjective, a bijective and an injective function (see ~ fundcmpsurbijinj ). This was also mentioned by Gérard Lang some time ago.

From this, both variants of decompositions of a function into a surjective and an injective function can be derived:

Let ` F = ( ( I o. B ) o. S ) ` be a decomposition of a function into a surjective, a bijective and an injective function, 
then ` F = ( J o. S ) ` with ` J = ( I o. B ) ` (an injective function) is a decomposition into a surjective and an injective function corresponding to ~ fundcmpsurinj ,
and ` F = ( I o. O ) ` with ` O = ( B o. S ) ` (a surjective function) is a  decomposition into a surjective and an injective function corresponding to ~ fundcmpsurinjimaid .

These additional theorems (and the modification of existing theorems necessary or helpful to prove these) are contained in the latest (and hopefully final) commit of this PR.

avekens:
@tirix and @GinoGiotto could you have a look at the latest commit 9851ad8, please?

GinoGiotto (approved):
Checked for ax-13 usage. No new theorems use it.

set.mm
```diff
@@ -732651,6 +732599,519 @@ The modulo (remainder) operation - extension
...
+      wrex wa preimafvsnel n0i ralnex eqcom notbii ralbii bitr3i elsetpreimafvb
+      syl wb 0ex ax-mp sylnibr df-nel ) ECGZHDIZJHDKURHELAMZENOPZQZACUBZUSURVAH
+      QZJZACRZVCJZURVEACURUTCIUCUTVAIVECEUTUDVAUTUEULUAVGVBJZACRVFVBACUFVHVEACV
+      BVDHVAUGUHUIUJSHTIUSVCUMUNABCDHETFUKUOUPHDUQS $.
+
+    ${
+      $d S x z $.  $d X x $.  $d Y x $.
+      $( An element of the preimage of a function value is an element of the
+         domain of the function with the same value as another element of the
+         premimage.  (Contributed by AV, 9-Mar-2024.) $)
```
tirix, line 732786: _preimage_

tirix (approved):

icecream17 (approved):

PR #3883, opened 2024-03-22
add taupi to iset.mm

jkingdon:
The taupi theorem is stated the same as in set.mm (with the same definition df-tau as well).

The proof is slightly different because there didn't seem to be a direct path to intuitionizing https://us.metamath.org/mpeuni/coseq1.html so I proved a few similar theorems instead, which were sufficient for taupi and I suppose might be on the path to coseq1 if we want that some day.


digama0:
Can you elaborate on what the issue with coseq1 is? From a cursory glance it seems like the theorem should still be true, although I'm not sure about what prerequisites you have.

jkingdon:
> Can you elaborate on what the issue with coseq1 is? From a cursory glance it seems like the theorem should still be true, although I'm not sure about what prerequisites you have.

Like you, I'm pretty sure it can be proved.

However the set.mm proof ( https://us.metamath.org/mpeuni/sineq0.html in particular) uses:

* modulo to an irrational modulus (which we could develop - see https://us.metamath.org/ileuni/flapcl.html - but have not yet developed)
* some uses of the floor function which at least at first glance don't appear like they'd work
* a use of https://us.metamath.org/mpeuni/leloe.html which would need to be replaced somehow

On top of all that coseq1 isn't really what the proof of taupi needs most directly. The theorem, proved here, which states ``A e. ( 0 (,) ( 2 x. _pi ) ) -> ( cos ` A ) < 1`` plugs into https://us.metamath.org/ileuni/infminti.html much more directly than coseq1 would.

iset.mm
```diff
@@ -145625,6 +145648,46 @@ mathematician Ptolemy (Claudius Ptolemaeus).  This particular version is
...
+    ZXMBSFZWRSFZWSXKXRXMUJTUMWRCDUNOUKQZBWRAULVDZUOUPXQXTXJXKXLXMUJTOYADWRAUQVD
+    URAUSUTXEXHJZAWTXGUBEFZXCYCXKXAXHYDXEXKXHXONWSXAXHVAXEXHUHWTSFXGSFYDXKXAXHU
+    JTWTRQXGKWTVBRUKZQWTXGAULVDURYDXBBIYDAAWTXGUGVCYDVEYDVFAVGBIGHYDVHPVIUTXEDX
+    GGHZXFXHVJZDKUCEWTGHZYFCKGHZYHVKCLFZBCGHZJKLFZBKGHZJZXQBDGHZJYIYHTYJYKUNVLV
+    MYLYMVBVQVMZXQYOOVNVMCKDVOVPVRXQWTLFZYNYHYFTORYPDWTKVSVPVRXQXGLFXEXKYFYGVTO
+    YEXODXGAWAWBWCWDWSXDJZXBBUDUEIGYRBABBDWEEZFZYRXSDSFBDMHYTUMDOQBDWFOVNWGBDWH
+    VPPYRXKBAMHADMHAYSFWSXKXDXNNZYRBAYRVEUUAWSXRXDWSXKXRXMYBWINZUIYRADUUAXQYROP
+    WSXDUHUIBDAWFOWJURUUBWKWLWMWSWTDGHZXAXDVJZDWNFUUCWODWPWQYQXQWSXKUUCUUDVTROX
+    NWTDAWAWBWCWD $.
+
```
avekens, line 145690: I think these theorems should be added to set.mm, too.

jkingdon: OK with me. I have revised this pull request to add both `cos02pilt1` and `cosq34lt1` .

avekens (approved):

PR #3885, opened 2024-03-22
Link to theorems in theorem sampler

jkingdon:
The main reason to link using the ~ syntax is that the build process will verify that these theorems exist.  Showing the names to the user is maybe good, maybe bad, but is mainly here as a side effect of that verification.

Rename prth to anim12 as this theorem was renamed a while ago.

Here's a screenshot of a portion of one of the revised pages:

image

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -727,115 +727,85 @@ HREF="mmbiblio.html">Bibliographic Cross-Reference</A> useful.</P>
 <P><TABLE BORDER=0><TR><TD VALIGN=TOP WIDTH="50%"><TR><TD VALIGN=TOP WIDTH="50%">
 <B>Propositional calculus</B>
 <MENU>
-<LI>
-<A HREF="idALT.html">Law of identity</A></LI>
+<LI>Law of identity ~ idALT</LI>
 
-<LI>
-<A HREF="prth.html">Praeclarum theorema</A></LI>
+<LI>Praeclarum theorema ~ anim12</LI>
 
-<LI>
-<A HREF="con3.html">Contraposition introduction</A></LI>
+<LI>Contraposition introduction ~ con3</LI>
 
-<LI>
-<A HREF="notnot.html">Double negation introduction</A></LI>
+<LI>Double negation introduction ~ notnot</LI>
 
-<LI>
-<A HREF="notnotnot.html">Triple negation</A></LI>
+<LI>Triple negation ~ notnotnot</LI>
 
-<LI>
-<A HREF="df-xor.html">Definition of exclusive or</A></LI>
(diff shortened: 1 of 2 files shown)


mmil.raw.html
```diff
@@ -727,115 +727,85 @@
...
-<A HREF="apcotr.html">3</A>
-<A HREF="apti.html">4</A></LI>
+~ apirr ~ apsym ~ apcotr ~ apti</LI>
 
-<LI> <A HREF="sqrt2irrap.html">The square root of 2 is irrational</A> (a
-different statement than <A HREF="sqrt2irr.html">The square root of 2
-is not rational</A>)</LI>
+<LI>The square root of 2 is irrational ~ sqrt2irrap (a
+different statement than The square root of 2
+is not rational ~ sqrt2irr )</LI>
```
avekens, line 800: ~ sqrt2irrap does not exist, I think this must be ~ sqrt2irr0.

Maybe ~sqrt2irr should be renamed to "sqrt2nq", then ~sqrt2irr0 can be renamed to "sqrt2irr"

Furthermore, I would propose to add quotation marks:
(a different statement than "The square root of 2 is not rational" ~ sqrt2irr )

avekens: Sorry, this is the theorem sampler for iset! Here we actually have ~sqrt2irrap: ``( Q e. QQ -> ( sqrt ` 2 ) =//= Q )``.
But this is in the same way not precisely the meaning of "is irrational": `A e. ( RR \ QQ )` is the adequate idiom to say "`A` is irrational".

avekens (approved):
I like this approach: It is shorter and safer.

mmil.raw.html
avekens: By the way, why don't we have ``( sqrt ` 2 ) e. ( RR \ QQ )`` in iset.mm?

tirix (approved):

icecream17 (approved):

mmil.raw.html
jkingdon: I've added the quotes, and added `sqrt2irr0` to iset.mm.

But the usual definition of irrational, in the absence of excluded middle, is `sqrt2irrap` rather than `sqrt2irr`. This is discussed in the comments for those theorems and if you want an external site which discusses it, see https://en.wikipedia.org/wiki/Irrational_number#In_constructive_mathematics or https://ncatlab.org/nlab/show/irrational+number

Or if you want an example of where this would matter, look at the antecedent of https://us.metamath.org/ileuni/flapcl.html

The reason I like having this in the iset.mm theorem sampler is, of course, because this may be a new concept to a lot of readers.

Issue #3879, opened 2024-03-18
Discourage indirect usage of ax-13?


GinoGiotto:
This question arised while reviewing #3877. Checkout the discussion from https://github.com/metamath/set.mm/pull/3877#discussion_r1527569500 for more context.

For this issue description I'm going to simply quote myself since what I wrote already effectively communicates my intentions:

> Given that ax-13 usage has recently reached an all time low, I would like to have some verification system that it will stay low, not only in the near future, but possibily in the years to come. If we don't do anything about it, I can assure you ax-13 usage will rise again. I've been collecting data about ax-13 usage for a while now. The lesson I've derived is: if you don't systematically check for something, undesired trends will show up, even if small or temporary (and they are usually temporary only because there is someone that checks for it afterwards, which should not be taken for granted).
> I'm not particularly attached to the discouragement system to check for this metric, it's just the easiest tool it is currently available for us. If someone decides to show up and build a tool just for this problem, I'm not against it, on the contrary. But it has to be automatic and systematic, human review is not enough.
> I don't want to give the impression that using ax-13 is prohibited. In the end, axioms are there to be used. I want to avoid the scenario in which ax-13 is used without awareness. So rather than a prohibition issue this is a communication issue. We need to find a way to communicate the user "hey, it seems you're trying to build your proof using ax-13. Did you know that about 97% of theorems in the database don't need it? Perhaps yours doesn't need it too."
> Wolf Lammen is right to say that this problem does not only affects [cbvralv](https://us.metamath.org/mpeuni/cbvralv.html) or [cbvrexv](https://us.metamath.org/mpeuni/cbvrexv.html), this affects dozens of theorems, so either we are down to add about a hundred or so of discouragement tags (personally I don't have a problem with that), or we build something else. The worst decision in my opinion is to leave the state of the art as it is.

To this I'll add that we also have the tags `$j usage '<theorem>' avoids 'ax-13'`, which can be added to a few early theorems (I'll do this soon). While they help a little if put in strategic places, they don't communicate anything over newly added theorems (unless the newly added theorems come with their own $j tags).

GinoGiotto:
By the way, this proposal from wolf might not be a bad idea:

> Given that an unnecessary reference to ax-13 is a magnitude more likely than to, say, ax-11, what do you think of extending the $j annotation to mark a deliberate use of ax-13? `$j need ax-13 $.` for example. And a verifier that checks the simultaneous presence of ax-13 in the axiom list and this annotation.

[...]

PR #3886, opened 2024-03-22
Add $j tags

GinoGiotto:
Add $j tags to theorems containing claims about dependencies. Particularly focusing on the ones avoiding ax-13.


avekens (approved):
OK, as discussed in #3879

tirix (approved):

PR #3896, opened 2024-03-28
Discourage usage of theorems relying on ax-13

GinoGiotto:
Based on comments from multiple people, it seems that the discouragement mechanism is the preferred tool to address #3879. Therefore in this PR I add discouragement tags to every theorem in main relying on ax-13.

This is just the first step. If this PR lands, in the following ones I will add cross references, a few missing $j tags, revise comments appropriately and maybe rename some theorems for consistency, but those following developments will depend on what will be the result of this PR.

I also removed ax-13 from [iresn0n0](https://us.metamath.org/mpeuni/iresn0n0.html), which I noticed while I was scanning theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -191,11 +191,34 @@
 "1sr" is used by "supsr".
 "1t1e1ALT" is used by "nnmul1com".
 "1t1e1ALT" is used by "remulinvcom".
+"2ax6e" is used by "2sb5rf".
+"2ax6e" is used by "2sb6rf".
+"2ax6e" is used by "2sb6rfOLD".
+"2ax6elem" is used by "2ax6e".
+"2ax6elem" is used by "2ax6eOLD".
+"2eu1" is used by "2eu2".
+"2eu1" is used by "2eu3".
+"2eu1" is used by "2eu3OLD".
+"2eu1" is used by "2eu5OLD".
+"2eu2" is used by "2eu8".
+"2eu7" is used by "2eu8".
+"2euex" is used by "2exeu".
+"2euswap" is used by "2eu1".
+"2euswap" is used by "2eu1OLD".
+"2euswap" is used by "euxfr2".
+"2exeu" is used by "2eu1".
+"2exeu" is used by "2eu1OLD".
+"2exeu" is used by "2eu2".
+"2exeu" is used by "2eu3".
+"2exeu" is used by "2eu3OLD".
 "2llnjN" is used by "2llnm2N".
 "2llnjaN" is used by "2llnjN".
 "2llnm2N" is used by "2llnm3N".
 "2llnma2rN" is used by "cdleme20y".
 "2llnne2N" is used by "2llnneN".
+"2moex" is used by "2eu2".
+"2moex" is used by "2eu5OLD".
+"2moswap" is used by "2euswap".
 "2pm13.193" is used by "2sb5nd".
 "2pm13.193" is used by "2sb5ndALT".
 "2pm13.193" is used by "2sb5ndVD".
@@ -224,6 +247,7 @@
 "2polvalN" is used by "sspmaplubN".
 "2sb5nd" is used by "2uasbanh".
 "2sb5nd" is used by "2uasbanhVD".
+"2sb5rf" is used by "sbel2x".
 "2uasbanh" is used by "2uasban".
 "3atnelvolN" is used by "2atnelvolN".
 "3atnelvolN" is used by "islvol2aN".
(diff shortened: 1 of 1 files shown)


avekens:
If I count correctly, there are about 300 new "Discourage" tags, and about 80 of the affected theorems are not used by other theorems. As proposed in #3879, these 80 theorems could be renamed without much effort so that they can be identified as "use ax-13 but are not used" theorems easily (this can be done in a separate PR if there is an agreement which should be discussed in #3879).

avekens:
> This is just the first step. If this PR lands, in the following ones I will add cross references, a few missing $j tags, revise comments appropriately and maybe rename some theorems for consistency, but those following developments will depend on what will be the result of this PR.

For a proposal regarding comments, see my comment in #3879.

avekens (commented):
As I wrote in #3879, I support this approach. Before I approve this PR, however, I want to hear other's opinions.

jkingdon:
I haven't been following the ax-13 reduction enough to have an especially firm set of opinions, but generally it makes sense to me and it seems like we end up with enough theorems which use distinct variables and enough theorems which use distinctors that we demonstrate both techniques. Beyond that, I think I better defer to the comments at #3879 including some numbers on how much certain things are used.

As for the details, it would be easier for me to work with these theorems in the future if the comment of the discouraged theorem said why it is discouraged. Ideally, this would point to another theorem and say something like "Most usages should prefer ~ xxxx which has an additional distinct variable constraint but which uses fewer axioms". I don't know what people think is practical here. But even something boilerplate like "Discouraged because it uses ax-13" which could be copy-pasted without further elaboration would, I would think, be an improvement over not saying anything.

GinoGiotto:
Ok, I can add comments and cross references here. What phrasing would satisfy you the most?

I was thinking:
<pre>
Usage of this theorem is discouraged because it depends on  ~ ax-13 . Use the weaker ~ xxx if possible.
</pre>

Note that not all theorems have a weaker version (because they are not needed), so most of the time only the first sentence would be present.

GinoGiotto:
I added comments and references to weaker versions. Contrary to what I initially thought, most theorems depending on ax-13 have one, so it took me some time to check  and report them all. I haven't always used boilerplate sentences because in some occurrences they sounded either redundant or out of place, so I revised the wording from time to time. The case of [rgen2](https://us.metamath.org/mpeuni/rgen2.html) vs [rgen2a](https://us.metamath.org/mpeuni/rgen2a.html) is special and I used a stronger language there ([rgen2a](https://us.metamath.org/mpeuni/rgen2a.html) was used more than 200 times before my PR series, but now it's not used anywhere anymore).

I also added a few $j tags and removed ax-13 from [disjxun](https://us.metamath.org/mpeuni/disjxun.html).

jkingdon (approved):

jkingdon:
> I added comments and references to weaker versions.

Thanks! This helps significantly.

I guess comments so far seem to be positive and it seems like we've given people time to weigh in, so I'll mark this as approved. Feel free to use your judgement about whether to merge it now or seek further discussion.

avekens (approved):
OK for now, but the resuls of the discussions of issue #3896 should be incorporated later. For example, I would like to propagate the original contributors to the new theorems which should be used instead of the discouraged theorems, as proposed on March 27.

set.mm
```diff
@@ -21168,14 +21243,18 @@ requires either a disjoint variable condition ( ~ sb56 ) or a non-freeness
   $}
 
   $( Rederivation of axiom ~ ax-12 from ~ ax12v (used only via ~ sp ) ,
-     ~ axc11r , and ~ axc15 (on top of Tarski's FOL).  (Contributed by NM,
-     22-Jan-2007.)  Proof uses contemporary axioms.  (Revised by Wolf Lammen,
-     8-Aug-2020.)  (Proof shortened by BJ, 4-Jul-2021.) $)
+     ~ axc11r , and ~ axc15 (on top of Tarski's FOL).  Since this version
+     depends on ~ ax-13 , usage of the weaker ~ ax12v or ~ ax12w are preferred.
```
wlammen, line 21247: IMO ax12i should be noted here as well.

GinoGiotto: Done

GinoGiotto:
> For example, I would like to propagate the original contributors to the new theorems which should be used instead of the discouraged theorems, as proposed on March 27.

Can you open a separate issue for this? I would like to close https://github.com/metamath/set.mm/issues/3879 after this PR gets merged since a tool to automatically notice new ax-13 usage is implemented.

GinoGiotto:
> Feel free to use your judgement about whether to merge it now or seek further discussion.

I don't have write access. I'd say wait 12 hours max and then merge if nobody says anything.

tirix (approved):

jkingdon:
> > Feel free to use your judgement about whether to merge it now or seek further discussion.
> 
> I don't have write access. I'd say wait 12 hours max and then merge if nobody says anything.

Hope I'm not jumping the gun by 11 hours but seems like this is ready to merge.

As usual, if people still see problems, open a new issue or pull request.

jkingdon:
See https://github.com/metamath/set.mm/pull/3902 for a fix to failures I'm seeing on develop after merging this.

PR #3899, opened 2024-04-01
Intuitionize products from prodex to nfcprod

jkingdon:
Although this is a pretty small pull request, I don't see much point in letting this kind of thing pile up in a branch for more than a day or two.

These are just some of the equality and freeness theorems for `prod_`.

Includes adopting the set.mm naming convention for `nfiotadw` and `nfiotaw` (iset.mm already had these theorems under other names).


avekens (approved):

jkingdon:
The change was to rebase (just in case it interacts with any of the other branches).

PR #3901, opened 2024-04-01
BJ mathbox

GinoGiotto:
* Use ~ axc11rv instead of ~ axc11r in proof of ~ bj-axc11v.

* Drop ax-11, ax-12 , ax-13 from ~ bj-cbvex4vv.

* Drop ax-13 from ~ ax11-pm2.

* Drop ax-13 from ~ bj-sbeqALT (it says that ~ bj-sbeq is the more general version, so it should make sense to reduce axiom usage in the weaker one).


benjub (approved):
Thanks a lot @GinoGiotto !

PR #3902, opened 2024-04-01
Update discouraged file

jkingdon:
Apparently several pull requests crossed. There are fewer discouraged usages than currently shown in the discouraged file.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -3195,7 +3195,6 @@
 "cbvex" is used by "cbveuALT".
 "cbvex" is used by "cbvexv".
 "cbvex" is used by "sb8e".
-"cbvex2vv" is used by "bj-cbvex4vv".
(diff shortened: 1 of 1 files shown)


GinoGiotto (approved):
This is an interesting case where a discouraged conflict is visible only upon merge.

(Since this is a technical issue I'd suggest to merge this immediately.)

Short summary: PR https://github.com/metamath/set.mm/pull/3901 changed the usage of some theorems that were not discouraged. After https://github.com/metamath/set.mm/pull/3896 was merged, those theorems were assigned a discouragement tag. But the checks look only for self consistency. Only when merged togheter it was possible to notice the updated usage (another way to see this would be do add a merge commit to https://github.com/metamath/set.mm/pull/3896).

See https://github.com/metamath/set.mm/pull/3895 to view another example where the same thing happens.

PR #3904, opened 2024-04-02
Countable reals

jkingdon:
The countable reals paper, long awaited (well at least in the `ruc` entry in https://us.metamath.org/ileuni/mmil.html ) is now up on arxiv! The pull request updates that entry and also uses this paper as a literature reference for one theorem which didn't have one.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11408,6 +11408,10 @@ accepting constructive mathematics,&quot; <I>Bulletin (New Series) of
 the American Mathematical Society</I>, 54:481-498 (2017),
 DOI: <A HREF="http://dx.doi.org/10.1090/bull/1556">10.1090/bull/1556</A> .</LI>
 
+<li><a name="BauerHanson"></a>[BauerHanson] Bauer, Andrej and Hanson,
+James E, "The Countable Reals", arXiv:2404.01256, 2024,
+<a href="https://arxiv.org/abs/2404.01256">https://arxiv.org/abs/2404.01256</a></li>
+
 <li><a name="BauerSwan"></a>[BauerSwan] Bauer, Andrej and Swan, Andrew,
 "Every Metric Space is Separable in Function Realizability",
 <i>Logical Methods in Computer Science</i>,
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9212,11 +9212,8 @@ intuitionistic and it is lightly used in set.mm</TD>
 <TR>
   <TD>ruc</TD>
   <TD><I>none</I></TD>
-  <TD>Apparently not provable without countable choice, assuming
-  the following result holds up:
-  Andrej Bauer (12-May-2022), <a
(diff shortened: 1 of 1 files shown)


avekens:
Remarkable result/paper (although I had a look at the abstract only, an understand only very few of it).

But this means that ~ruc could be provable in iset.mm assuming CCHOICE, couldn't it? 

Searching for "countable choice" in iset.mm, I found the following text which should also be updated (it is page 400 in the PDF I downloaded today):

>        As for the axiom of choice, the full axiom of choice implies
>        excluded middle as seen at ~ acexmid , although some authors will
>        use countable choice or dependent choice.  For example, countable
>        choice or excluded middle is needed to show that the Cauchy reals
>        coincide with the Dedekind reals - Corollary 11.4.3 of [HoTT],
>        p. (varies).
> 

Is there any relationtship between the results of the two papers?

avekens (approved):

tirix (approved):

jkingdon:
> Remarkable result/paper (although I had a look at the abstract only, an understand only very few of it).

I don't know much about toposes or realizability either, so most of the details went over my head too. But section 1 (introduction) and parts of section 5 were pretty much understandable to me, and maybe would be to others who are familiar with iset.mm.

And of course the result - that the use of excluded middle in the set.mm proof of ruc (set.mm doesn't use choice) is not avoidable - is easy enough to understand, and relevant to iset.mm.

> But this means that ~ruc could be provable in iset.mm assuming CCHOICE, couldn't it?

Presumably. The BauerHanson paper has a sketch of a proof using countable choice, although I didn't look in detail how hard it would be to formalize (probably not super easy, given how involved the set.mm proof of ruc is, but should be doable).

> Searching for "countable choice" in iset.mm, I found the following text which should also be updated (it is page 400 in the PDF I downloaded today):
> 
> > ```
> >    As for the axiom of choice, the full axiom of choice implies
> >    excluded middle as seen at ~ acexmid , although some authors will
> >    use countable choice or dependent choice.  For example, countable
> >    choice or excluded middle is needed to show that the Cauchy reals
> >    coincide with the Dedekind reals - Corollary 11.4.3 of [HoTT],
> >    p. (varies).
> > ```
> 
> Is there any relationtship between the results of the two papers?

I can think of a lot of ways to update or modify this text, but what did you have in mind? Section 5.3 of the BauerHanson paper, or section 11.3 of the HoTT book, are all about the Cauchy reals, and I don't know what conclusion anyone else comes to but I guess I can quote section 1.2 of BauerHanson:

> The real numbers, whatever they are, ought to form a metrically complete
space. Thus we must disqualify the Cauchy reals, because without the axiom
of countable choice we cannot show that a Cauchy sequence of Cauchy reals
has a limit which is a Cauchy real.

which is, in a nutshell, why iset.mm develops Dedekind reals and doesn't really do much with Cauchy reals.

PR #3905, opened 2024-04-03
Reduce ax-13 usage

GinoGiotto:
* Add ~ brabidgaw, a weaker version of ~ brabidga that does not require ax-13.

* Add ~ rexlimddvcbvw, a weaker version of ~ rexlimddvcbv that does not require ax-13.

* Use the new theorems to reduce axiom usage.

* Remove a redundant dv condition.

* Add tags and revise comments.

* Adjust redundant text of ~ nfiundg.

This PR drops ax-13 from 29 theorems. Full axiom usage here: https://github.com/metamath/set.mm/commit/b6833c370a8b596e7bd3767ded801d9de7f4d681

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -3255,8 +3255,6 @@
 "cbvralv" is used by "cbvral3v".
 "cbvralv" is used by "frgrwopreglem5ALT".
 "cbvralv" is used by "smfsuplem2".
-"cbvralv" is used by "vonicc".
-"cbvralv" is used by "vonioo".
 "cbvreu" is used by "cbvreuv".
 "cbvreu" is used by "cbvrmo".
 "cbvrex" is used by "cbviung".
@@ -4299,12 +4297,7 @@
 "chub2i" is used by "qlaxr3i".
 "chub2i" is used by "sumdmdlem2".
 "chvar" is used by "chvarv".
-"chvar" is used by "vonhoire".
-"chvar" is used by "vonn0icc2".
-"chvar" is used by "vonn0ioo2".
 "chvar" is used by "zfcndrep".
-"chvarv" is used by "vonicc".
-"chvarv" is used by "vonioo".
 "clmgmOLD" is used by "exidcl".
 "cm0" is used by "chirred".
 "cm2j" is used by "cm2ji".
@@ -10116,7 +10109,6 @@
 "nfiota" is used by "nfsum".
 "nfiotad" is used by "nfiota".
 "nfiotad" is used by "nfriotad".
-"nfixp" is used by "vonioo".
 "nfmo" is used by "2euex".
 "nfmo" is used by "2moex".
 "nfmo" is used by "moexex".
@@ -14917,6 +14909,7 @@ New usage of "bnsscmcl" is discouraged (0 uses).
 New usage of "bra0" is discouraged (1 uses).
 New usage of "bra11" is discouraged (6 uses).
 New usage of "braadd" is discouraged (1 uses).
+New usage of "brabidga" is discouraged (0 uses).
 New usage of "brabn" is discouraged (1 uses).
 New usage of "bracl" is discouraged (4 uses).
 New usage of "bracnln" is discouraged (1 uses).
@@ -14987,7 +14980,7 @@ New usage of "cbvral3v" is discouraged (0 uses).
 New usage of "cbvralcsf" is discouraged (2 uses).
 New usage of "cbvralf" is discouraged (2 uses).
 New usage of "cbvralsv" is discouraged (0 uses).
-New usage of "cbvralv" is discouraged (6 uses).
+New usage of "cbvralv" is discouraged (4 uses).
 New usage of "cbvralv2" is discouraged (0 uses).
 New usage of "cbvreu" is discouraged (2 uses).
 New usage of "cbvreucsf" is discouraged (0 uses).
@@ -15296,8 +15289,8 @@ New usage of "chub1i" is discouraged (18 uses).
 New usage of "chub2" is discouraged (14 uses).
 New usage of "chub2i" is discouraged (24 uses).
 New usage of "chunssji" is discouraged (0 uses).
-New usage of "chvar" is discouraged (5 uses).
-New usage of "chvarv" is discouraged (2 uses).
+New usage of "chvar" is discouraged (2 uses).
+New usage of "chvarv" is discouraged (0 uses).
 New usage of "clel5OLD" is discouraged (0 uses).
 New usage of "cleljustALT" is discouraged (0 uses).
 New usage of "cleljustALT2" is discouraged (0 uses).
@@ -17423,7 +17416,7 @@ New usage of "nfiota" is discouraged (1 uses).
 New usage of "nfiotad" is discouraged (2 uses).
 New usage of "nfiundg" is discouraged (0 uses).
 New usage of "nfiung" is discouraged (0 uses).
-New usage of "nfixp" is discouraged (1 uses).
+New usage of "nfixp" is discouraged (0 uses).
 New usage of "nfmo" is discouraged (3 uses).
 New usage of "nfmod" is discouraged (2 uses).
 New usage of "nfmod2" is discouraged (5 uses).
@@ -18211,6 +18204,7 @@ New usage of "rexcomOLD" is discouraged (0 uses).
 New usage of "rexeqOLD" is discouraged (0 uses).
 New usage of "rexeqbi1dvOLD" is discouraged (0 uses).
 New usage of "rexeqbidvOLD" is discouraged (0 uses).
+New usage of "rexlimddvcbv" is discouraged (0 uses).
 New usage of "rexrnmpt" is discouraged (0 uses).
 New usage of "rexsngOLD" is discouraged (0 uses).
 New usage of "rgen2a" is discouraged (0 uses).


set.mm
```diff
@@ -564333,10 +564333,23 @@ relation with an intersection with the same Cartesian product (see also
     FUBBAFDGZUDIZUJUFUBUKUDFZFULUEUMUBUCUKUDABDJKLUBUKUDMNBACDOPUJBUGFZDGUHUIUN
     DBACMQBUGDRPSUBUEUHTUA $.
 
+  ${
+    $d x y $.
+    brabidgaw.1 $e |- R = { <. x , y >. | ph } $.
+    $( Version of ~ brabidga with a disjoint variable condition, which does not
+       require ~ ax-13 .  (Contributed by Gino Giotto, 2-Apr-2024.) $)
+    brabidgaw $p |- ( x R y <-> ph ) $=
```
avekens, line 564341: As mentioned in issue #3879 (which was closed recently), I think it is unfair not to mention the original contributor here, because this theorem will become standard, and the original theorem discouraged. Furthermore, the intention/description of the theorem is missing. Therefore, I would propose to change the comment as follows:

```
    $( The law of concretion for a binary relation.  Special case of ~ brabga . 
       Version of ~ brabidga with a disjoint variable condition, which does not
       require ~ ax-13 .  (Contributed by Peter Mazsa, 24-Nov-2018.)
       (Revised by Gino Giotto, 2-Apr-2024.) $)
```

This remark applies for all similar cases in the following (and previous PRs).

GinoGiotto: Done in e749638. As mentioned [in the previous PR](https://github.com/metamath/set.mm/pull/3896#issuecomment-2030172554) I think the general case of this proposal would be better addressed on its own issue (along with the ALTV suggention next to it).

As a quick note I should mention that not all versions without ax-13 have been contributed by me. For example, [drex1v](https://us.metamath.org/mpeuni/drex1v.html) and [drnf1v](https://us.metamath.org/mpeuni/drnf1v.html) have been contributed by @benjub as weaker versions of [drex1](https://us.metamath.org/mpeuni/drex1.html) and [drnf1](https://us.metamath.org/mpeuni/drnf1.html), which have been contibuted by Norman Megill and Mario Carneiro respectively (now discouraged).  Therefore, in the case of [drnf1](https://us.metamath.org/mpeuni/drnf1.html), the "Contributed by BJ" tag would be changed into "Contributed by Mario Carneiro" and "Revised by BJ" tags. Those technical lemmas used to be in BJ mathbox in a dedicated section that he made with the exact purpose of reducing axiom usage. It's not granted that all people that contributed weaker versions would be ok with this credit change (therefore I believe a separate issue should be opened for this).

GinoGiotto: And also there is currently an open PR https://github.com/metamath/set.mm/pull/3906/commits/3fa74c27b13c9d7adc18cd6811fa7f4ca9c673d7 for iset.mm with theorems containing my credit contributions, so even those would have to be edited for consistency.

avekens: > Done in [e749638](https://github.com/metamath/set.mm/commit/e749638e501e02c5df990b633a4d748bc79c50d5). As mentioned [in the previous PR](https://github.com/metamath/set.mm/pull/3896#issuecomment-2030172554) I think the general case of this proposal would be better addressed on its own issue (along with the ALTV suggention next to it).
> 
> As a quick note I should mention that not all versions without ax-13 have been contributed by me. For example, [drex1v](https://us.metamath.org/mpeuni/drex1v.html) and [drnf1v](https://us.metamath.org/mpeuni/drnf1v.html) have been contributed by @benjub as weaker versions of [drex1](https://us.metamath.org/mpeuni/drex1.html) and [drnf1](https://us.metamath.org/mpeuni/drnf1.html), which have been contibuted by Norman Megill and Mario Carneiro respectively (now discouraged). Therefore, in the case of [drnf1v](https://us.metamath.org/mpeuni/drnf1v.html), the "Contributed by BJ" tag would be changed into "Contributed by Mario Carneiro" and "Revised by BJ" tags. Those technical lemmas used to be in BJ mathbox in a dedicated section that he made with the exact purpose of reducing axiom usage. It's not granted that all people that contributed weaker versions would be ok with this credit change (therefore I believe a separate issue should be opened for this).

I would propose to re-open #3879 for this discussion (because it already contains some suggestions/remarks about this topic), or I want to ask you to create a new issue. Opinions of @benjub and @digama0 are welcome.

avekens (approved):
The changes by this PR are OK. The general discussion about comments/credits should/will be continued elsewhere.

set.mm
digama0: Just a quick comment on these matters:
* When rewriting a theorem to avoid some axiom, that counts as a "Revised by" and you should bring forward the "Contributed by". One way to think about this is that we optimize for the "Contributed by" date to be as early as possible, and try not to lose old dates due to the revise & deprecate cycle.
* We should try to maintain ax-13-free theorems using either "Proof modification discouraged" or the new `usage avoids` tag. I'm not sure if we have support for it yet in tooling but we can go ahead with tagging things anyway.

wlammen: brabidgaw is not exactly the same theorem as brabidga, because it adds a limiting DV condition, allowing for the removal of ax-13.  So both "interface" and semantics are different in detail.  In the past I sometimes reworked proofs avoiding ax-13 and other axioms, but only the proof -- the theorem stayed the same.  In these cases I added a revised tag.  Sometimes I invented a new lemma, (like ax13lem1) where a DV condition was added to something else already existing, but then I viewed myself a contributor.  If a proof had great similarity to something related I added an appreciative note at the end of the comment paragraph (Based on ideas of ...).

jkingdon: > Just a quick comment on these matters:
> 
> One way to think about this is that we optimize for the "Contributed by" date to be as early as possible, and try not to lose old dates due to the revise & deprecate cycle.

That makes sense to me. I have found it is often hard to know exactly what to do (in my case, this usually comes up when I'm doing things across set.mm and iset.mm), including on matters like what is new and what is just carrying over existing theorems/proofs, but I like the idea of trying to keep old dates as a rule of thumb.

> the new `usage avoids` tag. I'm not sure if we have support for it yet in tooling but we can go ahead with tagging things anyway.

There is tooling. See `metamath-knife --verify-usage` which we run in the set.mm/iset.mm checks: https://github.com/metamath/set.mm/blob/98035d55823ffee80d2b45423a9a8aedd07539df/.github/workflows/verifiers.yml#L258

benjub: > the "Contributed by BJ" tag would be changed into "Contributed by Mario Carneiro" and "Revised by BJ" tags.

That is fine for me.

avekens: > brabidgaw is not exactly the same theorem as brabidga, because it adds a limiting DV condition, allowing for the removal of ax-13. So both "interface" and semantics are different in detail. In the past I sometimes reworked proofs avoiding ax-13 and other axioms, but only the proof -- the theorem stayed the same. In these cases I added a revised tag. Sometimes I invented a new lemma, (like ax13lem1) where a DV condition was added to something else already existing, but then I viewed myself a contributor. If a proof had great similarity to something related I added an appreciative note at the end of the comment paragraph (Based on ideas of ...).

"Revised by" does not only mean "Proof revised by" - minor changes of the theorem/definition itself acan also be reasons for a revision. I think there are three cases:

* a theorem is replace by a new version (the label stays the same): here, it is clear that it is a revision only, the former version is usually marked as OLD and deleted after one year.
* a variant of the theorem is created (with a new label): in this case, both variants will be availabe and can be used without restrictions. Here, the creator of this variant is usually the contributor (unless the creator him-/herself decides that the main work was done by the former contributor and wants to honour this).
* a variant of the theorem is created (with a new label), but the former version becomes discouraged (this is the case under discussion): in my opinion, the change cannot be significant enough to change the contributor in this case, so that it should be a revision.

These are, of course, only rules of thumb, and there will always be border cases.  I think, however, the original contributor of a deprecated/discouraged theorem should survive in the theorem which should be used instead.

wlammen: `a variant of the theorem is created (with a new label), but the former version becomes discouraged`

The  former version is not discouraged. In fact, you are encouraged to use it when you want to express ideas without distinct variables (as in ax6e).  The discouragement technique is a temporary workaround to avoid accidental import of ax-13 in theorems that can do without it.  It is a misuse of a method invented for quite a different use case.  One should strive for dismissing it again as soon as possible, in my eyes.

I am afraid we already see the first signs of misinterpretations here, because said misuse is not apparent enough.

I do not mean hereby that adding a dv condition to a theorem, where the proof more or less carries over from the origin, is no candidate for keeping the contributor's name in the theorem.

GinoGiotto: > We should try to maintain ax-13-free theorems using either "Proof modification discouraged" or the new usage avoids tag. I'm not sure if we have support for it yet in tooling but we can go ahead with tagging things anyway.

Over 100 `$j usage` tags have already been added for the relevant theorems in PR https://github.com/metamath/set.mm/pull/3886, so this is done (discouraging proof modification is not a ideal because you still want to allow further minimizations).

> The discouragement technique is a temporary workaround...

Is it? At the moment we have the discouragement tool and I've not seen plans or approvals to add alternative ones. I'm afraid the current state will be very much permanent unless something like https://github.com/metamath/set.mm/issues/3879#issuecomment-2004760917 will be implemented (adding all `$j need` tags is more feasible now).

wlammen: `At the moment we have the discouragement tool and I've not seen plans or approvals to add alternative ones. I'm afraid the current state will be very much permanent...`

That's exactly why I was (and am) not in favor of adding these tags. For that you need to convince me you are up to add an adequate solution to metamath-knife or so, in the foreseeable future, replacing the current hack.  This piece then issues a warning in rewrap.sh that a use of ax-13 should be checked.  As a reaction you either remove its use, or add a 'Noticed' tag to your theorem.

The current state is highly motivated by quickly removing ax-13, and hampering its use, to the point, where I see our established habits been deteriorated.

I can pinpoint the downsides (dirty hack, axc11r), to no avail it seems.  Well, it won't hurt my life, but I was installed as a maintainer to watch out for misuses, and voice my concerns -- permanently, if necessary.

avekens:
As I said already before, the changes by this PR should be accepted and merged, if there are no specific objections. The general discussion about comments/credits should be continued elsewhere.

tirix:
> As I said already before, the changes by this PR should be accepted and merged, if there are no specific objections. The general discussion about comments/credits should be continued elsewhere.

I also agree. 
Peter has been mentioned in the contribution tag so @avekens's remark has been addressed.
The discussion can continue in #3879, even if the corresponding action was completed.

tirix (approved):

PR #3909, opened 2024-04-06
uniex

benjub:
Prove vuniex first, then uniexg, then uniex (like I did with vpwex a while ago).

Discourage use of zfun since it has x nested in itself and it obfuscates ax-un.

Ping @tirix since I shortened a proof in your mathbox, and discouraged one of your theorems (probably you overlooked that the biconditional was already in the main part). I could even obsolete or delete the theorem if you want.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18104,6 +18104,7 @@ New usage of "pwsnALT" is discouraged (0 uses).
 New usage of "pwtrVD" is discouraged (0 uses).
 New usage of "pwtrrVD" is discouraged (0 uses).
 New usage of "pwundifOLD" is discouraged (0 uses).
+New usage of "pwuniss" is discouraged (0 uses).
 New usage of "pwunssOLD" is discouraged (0 uses).
 New usage of "pythi" is discouraged (0 uses).
 New usage of "qexALT" is discouraged (1 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -461998,19 +462001,18 @@ Class abstractions (a.k.a. class builders)
     CAHUAUFAIACJKLMUAUEACBEZFZUCUAUECUDEZUGFUHCUDBNUAUIAUGUAUIAIACOKPQACBRST $.
 
   $( Condition for a class union to be a subset.  (Contributed by Thierry
-     Arnoux, 21-Jun-2020.) $)
+     Arnoux, 21-Jun-2020.)  Use ~ sspwuni instead.
+     (Proof modification is discouraged.)  (New usage is discouraged.) $)
   pwuniss $p |- ( A C_ ~P B -> U. A C_ B ) $=
     ( cpw wss cuni uniss unipw sseqtrdi ) ABCZDAEIEBAIFBGH $.
```
tirix, line 462007: Yes, feel free to delete, thank you!

tirix (approved):

set.mm
benjub: Done in https://github.com/metamath/set.mm/pull/3909/commits/a10ec2ed2dde9e2a0ff863dc2dfc0828cef2b0e7

tirix (approved):

PR #3910, opened 2024-04-06
Mathbox: add quantitative versions of some set theory existence theor…

benjub:
…ems (and edit one commit in Main accordingly).


set.mm
```diff
@@ -50217,9 +50217,10 @@ That theorem bundles the theorems ( ` |- E. x ( x = y -> z e. x ) ` with
   univ $p |- U. _V = _V $=
     ( cvv cpw cuni pwv unieqi unipw eqtr3i ) ABZCACAHADEAFG $.
 
-  $( Membership of a power class.  Exercise 10 of [Enderton] p. 26.
-     (Contributed by NM, 13-Jan-2007.) $)
-  pwel $p |- ( A e. B -> ~P A e. ~P ~P U. B ) $=
+  $( Quantitative version of ~ pwexg : the powerset of an element of a class is
+     an element of the double powerclass of the union of that class.  Exercise
+     10 of [Enderton] p. 26.  (Contributed by NM, 13-Jan-2007.) $)
+  pwel $p |- ( A e. V -> ~P A e. ~P ~P U. V ) $=
```
avekens, line 50223: Why did you change ` B ` to ` V `? We usually use ` V ` for an arbitrary class, especially in the idioms ` A e. V ` to express "`A ` is a set". In this case, however, ` V `is not arbitrary, because it appears at the RHS of the implication ("that class").  Therefore, I thinlk ` B ` is a better choice here.

set.mm
```diff
@@ -538886,6 +538891,29 @@ is a Moore collection ( ~ bj-snmoore ), this can also be stated as
       VAUTMFDRWDVBVCVDVEVF $.
   $}
 
+  ${
+    $d A x $.
+    $( A pair formed of the empty set and any class is a Moore collection (on
+       ` U. { (/) , A } ` , which is ` A ` if it is a set, else ` (/) ` ).
+
+       One also has ` |- ( ( A C_ B /\ A e. V ) -> { A , B } e. Moore_ ) ` ,
```
avekens, line 538899: Do you plan to provide this as theorem soon?

benjub: Yes, it's only a small variant of that theorem (basically, replace in the proof the line $\varnothing \cap A= \varnothing$ by $A \subseteq B \to A \cap B = A$), but I won't have time in this MR.

set.mm
benjub: Done.

avekens (approved):

set.mm
tirix: Wouldn't it be easier to prove the empty set version from the subset version?
I.e. to prove that coming theorem from this version?

benjub: Of course: once I prove the general theorem, the one with the empty set will be a direct consequence (as will be the forward implication in `|- ( A e. _V <-> { A } e. Moore_ )`).

PR #3911, opened 2024-04-10
Describe the current situation in ax-13 and axc11r

wlammen:
1. Explain why ax-13, axc11r is best avoided outside of niches.
2. Remove usage of df-bi from the implication only theorem jcn and shorten its proof.
3. Delete outdated OLD theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -18137,7 +18137,6 @@ New usage of "ralnex2OLD" is discouraged (0 uses).
 New usage of "ralnex3OLD" is discouraged (0 uses).
 New usage of "ralrexbidOLD" is discouraged (0 uses).
 New usage of "ralrnmpt" is discouraged (1 uses).
-New usage of "ralsngOLD" is discouraged (0 uses).
 New usage of "ralxfrALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -20546,6 +20552,17 @@ satisfied directly but only indirectly (outside of Metamath) by the
      with the same variable.  In this case, the first antecedent becomes
      ` -. x = x ` and the axiom still holds.
 
+     This axiom is mostly used to eliminate conditions requiring set variables
+     be distinct (cf. ~ ax6ev and ~ ax6e , for example) in proofs.  In
+     practice, theorems beyond elementary set theory do not really benefit from
+     such eliminations, so direct or indirect application of this axiom is
+     restricted now.  You need to explicitly confirm its use in case you see a
```
tirix, line 20559: Rather than "restricted", maybe "discouraged" ?
"restricted" sounds a bit like it would be impossible.

wlammen: done

tirix (approved):

jkingdon (approved):

PR #3912, opened 2024-04-10
Krull's Theorem

tirix:
Follow-up for #3787's comments. This adds a proof of [Krull's Theorem](https://en.wikipedia.org/wiki/Krull%27s_theorem) as discussed.

Next I'd like to prove that a field only has two ideals (similar to [~isfldidl2](https://us.metamath.org/mpeuni/isfldidl2.html), but using the structure notation). This would allow to derive that if $R$ is a commutative ring with 1, and $M$ is an ideal, then $R/M$ is a field if and only if $M$ is a maximal ideal. (but that part is _not_ in this PR)

### DIFF
--- a/big-unifier.mm
+++ b/big-unifier.mm
@@ -20,7 +20,7 @@ $( !
 This work is waived of all rights, including copyright, according to the CC0
 Public Domain Dedication.  http://creativecommons.org/publicdomain/zero/1.0/
 
-Norman Megill - http://metamath.org
+Norman Megill - https://us.metamath.org
 
 $)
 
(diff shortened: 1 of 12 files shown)


set.mm
```diff
@@ -476170,6 +476181,105 @@ The sumset operation can be used for both group (additive) operations and
...
+      $d B a b i x $.  $d B j p $.  $d I j p $.  $d R a b i x $.  $d R j p $.
+      $d Z a b i x $.  $d Z j p $.  $d a b i j ph x $.
+      ssmxidllem.1 $e |- P = { p e. ( LIdeal ` R ) | ( p =/= B /\ I C_ p ) } $.
+      ssmxidllem.2 $e |- ( ph -> R e. Ring ) $.
+      ssmxidllem.3 $e |- ( ph -> I e. ( LIdeal ` R ) ) $.
+      ssmxidllem.4 $e |- ( ph -> I =/= B ) $.
+      ssmxidllem2.1 $e |- ( ph -> Z C_ P ) $.
+      ssmxidllem2.2 $e |- ( ph -> Z =/= (/) ) $.
+      ssmxidllem2.3 $e |- ( ph -> [C.] Or Z ) $.
+      $( The set used in the proof of ~ ssmxidl satisfies the condition of
```
avekens, line 476196: Which set is this? `P `? Then the comment should start with "The set ` P ` used in the proof... ".

avekens:
Amazing results.

set.mm
tirix: You guessed it, it's `P`! 
Added.

avekens (approved):

icecream17 (approved):

PR #3919, opened 2024-04-15
Rename and shorten unitres*

wlammen:
1. Rename unitresl/r to ol/rcnd.  This better matches naming habits.
2. Shorten olcnd.
3. Fix some typos in my mathbox.
4. Delete an outdated OLD  theorem


tirix (approved):
Indeed those are way better names. 

I'm the one who moved those theorems to main, I should have renamed them at that time.

PR #3921, opened 2024-04-16
Fix comments for structure power theorems

CatsAreFluffy:
Some comments for theorems about structure powers refer to structure products instead, so I fixed them.


tirix (approved):
Good catch! Must have been copy/pasted and kept as-is by mistake.

PR #3922, opened 2024-04-16
dfifp2 comment

wlammen:
@benjub I don't understand the eliminated sentence "This version of the definition uses only primitive
     symbols ( ` -> , -. , A. ` )." in the context of dfifp2.  Obviously the (in our database) non-primitive /\ is used, while the For-all operator is not needed here.  I suggest to eliminate this sentence, but perhaps you had something in mind when adding the said sentence to the theorem's comment.  I somehow have no clue what it is...
Can we be more explanatory here?  Or just erase this sentence...


benjub:
> @benjub I don't understand the eliminated sentence "This version of the definition uses only primitive symbols ( `-> , -. , A.` )." in the context of dfifp2. Obviously the (in our database) non-primitive /\ is used, while the For-all operator is not needed here. I suggest to eliminate this sentence, but perhaps you had something in mind when adding the said sentence to the theorem's comment. I somehow have no clue what it is... Can we be more explanatory here? Or just erase this sentence...

You're right to delete this sentence. It's probably a copy-paste from another statement. The only-primitive-symbol versions of `ifp` do not look particularly nice, so I think there is no need to add another theorem.

By the way, while checking this, I saw the same mistake in the comment of bj-dfnnf2. Could you make the change
```
  $( Alternate definition of ~ df-bj-nnf using only primitive symbols
     ( ` -> ` , ` -. ` , ` A. ` ).  (Contributed by BJ, 20-Aug-2023.) $)
```
to
```
  $( Alternate definition of ~ df-bj-nnf using only primitive symbols
     ( ` -> ` , ` -. ` , ` A. ` ) in each conjunct.  (Contributed by BJ,
     20-Aug-2023.) $)
```
?

wlammen:
@benjub done.
Need an approval now (from you or others)

tirix (approved):

PR #3923, opened 2024-04-16
Edit credits, add comments

GinoGiotto:
This PR applies @avekens request https://github.com/metamath/set.mm/pull/3905#discussion_r1550489013 of adding comments and credits to theorems avoiding ax-13. The criteria goes as follows: scan through the discouraged theorems using ax-13 and check out whether they have a newer weaker version; if they have one then replace the credit of the weaker version with the contributor of the original one. As default I only added the original contributor and not the following revisers, as it's not always clear to me whether they play a role on the new version (specifically it's not always stated whether the revison concerned the old proof only and not the statement itself). This may change with the review process.

Additional changes:

* Replace usages of [ralcom2](https://us.metamath.org/mpeuni/ralcom2.html) and [ralcom2w](https://us.metamath.org/mpeuni/ralcom2w.html) with [ralcom](https://us.metamath.org/mpeuni/ralcom.html), which uses less axioms (and delete my version [ralcom2w](https://us.metamath.org/mpeuni/ralcom2w.html) since it's just redundant and never used).

* Delete my version [cbvrabcsfw](https://us.metamath.org/mpeuni/cbvrabcsfw.html) since it was only added to avoid ax-13, but it's never used.

* Delete my mathbox. The information it contains it's not relevant anymore and I'm not planning to add other stuff into it. So, to avoid waste of space, it can be removed.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12129,7 +12129,6 @@
 "psubspi2N" is used by "pclfinN".
 "psubspi2N" is used by "pclfinclN".
 "qexALT" is used by "reexALT".
-"ralcom2" is used by "tratrbVD".
 "ralrnmpt" is used by "rexrnmpt".
 "rb-ax1" is used by "rblem1".
 "rb-ax1" is used by "rblem2".
@@ -18127,7 +18126,7 @@ New usage of "rabbidvaOLD" is discouraged (0 uses).
 New usage of "ralab2OLD" is discouraged (0 uses).
 New usage of "ralanidOLD" is discouraged (0 uses).
 New usage of "ralbiOLD" is discouraged (0 uses).
-New usage of "ralcom2" is discouraged (1 uses).
+New usage of "ralcom2" is discouraged (0 uses).
 New usage of "ralcom4OLD" is discouraged (0 uses).
 New usage of "raleleqALT" is discouraged (0 uses).
 New usage of "raleqOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


benjub (approved):
I skimmed through it and it looks good to me.
Thanks for that scrupulous work.

tirix (approved):

wlammen:
The way the credits are handled, is not convincing to me, to be honest.  As far as my person is concerned, I do not care much.  But in principle is adding DV conditions and finding a proof for such a specialized theorem an act of ingeniosity in its own right in most cases.  The proofs do not generally carry over (see https://us.metamath.org/mpeuni/spei.html /https://us.metamath.org/mpeuni/speiv.html for example).  This is similar to specializing a group to a symmetric group, for example.  Extra conditions are added, allowing for a different theory, in the end with its own rules and proof techniques.  Why should one specializing a general rule about groups to symmetric ones not appear as contributor of that theorem, but as a reviser of the one in the more general context? Why are DV conditions handled differently to adding/modifying a hypothesis line?

I don't think this comment should block this PR.  But it can serve as a starting point for a discussion after which we adapt our conventions accordingly, and corrections in set.mm can be made.  In case it is supported by some, I will open an issue regarding this topic.

wlammen (approved):
As stated in my comment above.

icecream17 (approved):

PR #3924, opened 2024-04-16
Reduce ax-13 usage (The end)

GinoGiotto:
Last PR of the series, as I cannot find more theorems that can avoid ax-13 usage. I invite anyone to investigate whether they can discover further improvements.

Axiom usage here: bfbf2806ad81539f8f0666f2a3c2db170ce50efa

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -3232,20 +3232,12 @@
 "cbvmptg" is used by "cbvmptvg".
 "cbvopab1g" is used by "cbvmptfg".
 "cbvrab" is used by "cbvrabvOLD".
-"cbvrab" is used by "smfliminf".
-"cbvrab" is used by "smflimsup".
 "cbvrabcsf" is used by "cbvrabv2".
-"cbvrabcsf" is used by "smfinf".
-"cbvrabcsf" is used by "smfinflem".
-"cbvrabcsf" is used by "smfsup".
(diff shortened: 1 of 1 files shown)


benjub (approved):

tirix (approved):

wlammen:
I'd like to thank you for your contributions, too.  The difference to the old state is very noticeable, and valuable, I'd say.  In particular we now better know what role ax-13 plays in set.mm.  There it is reduced to a small area, where reducing distinct variable conditions enable a more general expression.  It hardly effects the great outline of mathematics, as expressed in this database.

wlammen (approved):

icecream17 (approved):
:tada:

I'll probably delete the Alternate definition of substitution section, seeing as how dfsb1 doesn't depend on that section anymore

this happens to further reduce the number of theprems using ax-13

jkingdon:
> 🎉
> 
> I'll probably delete the Alternate definition of substitution section, seeing as how dfsb1 doesn't depend on that section anymore
> 
> this happens to further reduce the number of theprems using ax-13

Hmm, well if that section somehow demonstrates some alternative approach or something, we could keep it for that reason (and say so in comments).

But I'm not sure it does demonstrate any approach distinct enough to keep and mention.

PR #3929, opened 2024-04-18
remove extra hypothesis: rnmptc, qsalrel

icecream17:
both proofs are shorter


wlammen:
I think the dependency on ax-pow is removed from rnmptc now.  You can add this to your reason for revision, too.

wlammen (approved):

tirix:
Tagging @glacode

tirix (approved):

wlammen:
ph and x need not be distinct in rnmptc

PR #3931, opened 2024-04-18
Update discussion of ruc in mmcomplex.html

jkingdon:
The current text has diverged quite a bit from the proof in set.mm (for example, it refers to 39 lemmas, but set.mm has 13).  Update it to reflect the way the set.mm proof now works, but keep the overall structure, and most of the text, intact.

Refer to theorems (including lemmas) by using the ~ syntax which should reduce the chance of this happening again if the set.mm proof is changed again to this extent.

### DIFF
--- a/mmcomplex.raw.html
+++ b/mmcomplex.raw.html
@@ -663,8 +663,7 @@ In mathematics, a countable set is a set with the same cardinality
 A countable set may be infinite, but every element of the set can be
 mapped to a unique natural number.
 The natural numbers and the integers are countably infinite sets.
-Even the rational numbers are a countably infinite set; see
-<A HREF="qnnen.html">qnnen</A>.
+Even the rational numbers are a countably infinite set; see ~ qnnen .
 In contrast, the real numbers are not a countable set, but instead,
 they are an uncountable set.
 This is a remarkable result.
@@ -728,19 +727,18 @@ pentaconta, hexaconta, heptaconta, octaconta, enneaconta, hecta
 -->
 
 However, he ran into some technical complications, such as the fact that some
-numbers have two equivalent decimal representations (e.g.  <A
-HREF="0.999....html">0.999...  = 1.000...</A>), that would have made a
+numbers have two equivalent decimal representations
+(e.g. 0.999... = 1.000..., see ~ 0.999... ), that would have made a
 formal proof somewhat messy.  So he chose another proof that is simpler
 to formalize but which he believes is in the same spirit as Cantor's
 diagonal proof, in the sense that it constructs a real number different
 from all the numbers in a purported list of all real numbers.
 
 <P>Even so, the "simpler" proof is still daunting when worked out in
-complete formal detail, involving some 39 lemmas.  Therefore we will
+complete formal detail, involving many lemmas.  Therefore we will
 first give an informal description of the proof, then describe the key
-formal lemmas that lead to the final result, which is theorem <A
-HREF="ruc.html">ruc</A> (or equivalently <A
-HREF="ruclem39.html">ruclem39</A>).
+formal lemmas that lead to the final result, which is theorem ~ ruc
+(or equivalently ~ ruclem13 ).</P>
 
 <P> <B><FONT COLOR="#006633">The informal
 argument</FONT></B>&nbsp;&nbsp;&nbsp; We will start by claiming that we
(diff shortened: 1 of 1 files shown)


mmcomplex.raw.html
```diff
@@ -726,19 +725,18 @@
 -->
 
 However, he ran into some technical complications, such as the fact that some
-numbers have two equivalent decimal representations (e.g.  <A
-HREF="0.999....html">0.999...  = 1.000...</A>), that would have made a
+numbers have two equivalent decimal representations
+(e.g. ~ 0.999... = 1.000...), that would have made a
```
avekens, line 729: Here, a reference is within an equation now. Is this a good idea? Maybe "(e.g. 0.999... = 1.000..., see ~ 0.999...)" would be clearer,

mmcomplex.raw.html
```diff
@@ -661,8 +661,7 @@
 A countable set may be infinite, but every element of the set can be
 mapped to a unique natural number.
 The natural numbers and the integers are countably infinite sets.
-Even the rational numbers are a countably infinite set; see
-<A HREF="qnnen.html">qnnen</A>.
+Even the rational numbers are a countably infinite set; see ~ qnnen .
```
wlammen, line 664: Using ~ here means that the page described here needs an extra resolution step from files by Metamath or relatives, and cannot be generated by some odd HTML tool any longer.  Why is that sensible?
If it is, there are still other links not converted, like cnpi for example.

digama0: I believe the web site generation process is not doing such resolution currently, although metamath-exe has the capability to do it.

jkingdon: The generation script does this for all files named `*.raw.html` here: https://github.com/metamath/metamath-website-scripts/blob/8d81a0c1ab3e433b335b4e6a3bdc364d277b4946/build-website.sh#L89 (I guess they also need to be copied here: https://github.com/metamath/metamath-website-scripts/blob/8d81a0c1ab3e433b335b4e6a3bdc364d277b4946/build-website.sh#L13 ). This mechanism has been in use quite extensively for mmil.raw.html for a long time and is also in use by mmset.raw.html .

As for why this is a good idea, one reason is to get the ability to typeset metamath notation, so that when editing the web pages we can just write `` ` a e. B ` `` and have it use the current markup from htmldef. Another reason is to check that references to theorems point to theorems which still exist, which is quite helpful given how much we rename things.

As for converting the rest of the file (`cnpi` for example), we'll get there eventually. As the discussion of `0.999...` elsewhere in this pull request shows, this is not a completely mechanical process (although it is mostly mechanical). But we have to start somewhere and the `ruc` text seemed like a good one because it is so out of date.

mmcomplex.raw.html
jkingdon: I wasn't sure about that one either. The page when I started was using a somewhat similar, dubious, way of formatting the link. Anyway, I've adopted this suggestion.

mmcomplex.raw.html
digama0: Oh yes you are right. (The second link is not a mistake, the files are first copied into mpegif, then to mpeuni, then marked up)

jkingdon: In addition, I've been using the https://github.com/metamath/set.mm/blob/develop/scripts/regen-from-raw script as I edit, to see what the formatted page looks like.

wlammen: If it was my page, I'd add a hint like `use regen-from-raw to transscript to pure html` somewhere at the beginning.

jkingdon: I've submitted this at #3945

mmcomplex.raw.html
```diff
@@ -750,39 +748,32 @@
 <I>f</I>(3),... of all reals, thereby falsifying this claim.
 
 <P>Here is how we construct this number.  We construct, in parallel, two
-auxiliary sequences <I>g</I>(1), <I>g</I>(2), <I>g</I>(3),... and
-<I>h</I>(1), <I>h</I>(2), <I>h</I>(3),... derived from <I>f</I>(1),
+auxiliary sequences <I>g</I>(0), <I>g</I>(1), <I>g</I>(2),... and
```
wlammen, line 751: I think the original indices were right, see the definition https://us.metamath.org/mpeuni/df-nn.html

wlammen (dismissed):

mmcomplex.raw.html
jkingdon: `f` is indexed by `NN`, `g` and `h` are indexed by `NN0`.  Look at https://us.metamath.org/mpeuni/ruclem6.html and note especially 𝐹:ℕ⟶ℝ and 𝐺:ℕ0⟶(ℝ × ℝ)

wlammen: you're right

jkingdon:
Is this ready to merge? I believe all the comments have been addressed.

avekens:
I cannot approve that all syntactical changes as well as the the semantical changes (see, for example, the indices) are correct. But I also have no objections against any of the changes.

jkingdon:
> I cannot approve that all syntactical changes as well as the the semantical changes (see, for example, the indices) are correct. But I also have no objections against any of the changes.

If only we had a system which could take a proof and tell us whether it is correct!

More seriously, I can say with high confidence that the text in this pull request is closer to correct (meaning free of error and also that it describes what is in the set.mm proof of ruc) than what is on the web site now. I can say with medium confidence that it is correct and I did take another pass over it to see if I could spot anything (I did find a few needed updates with respect to &lt; versus ≤ which I have now made).

Therefore, I would propose to take the quoted comment as an approval, unless someone else wants to weigh in within the next, oh, day or so I suppose.

wlammen:
In line 688 the text looks suspicious to me:
`Norman Megill's initial plan to prove that the rational numbers are uncountable ...`
I think he wanted to prove the reals are not countable, or that the rational numbers are countable. Edit: looking up the referenced Wikipedia page, the first part seems to be correct here.

I support your idea to rewrite the file as presented here.  Even if it still contains errors of the kind I noted above, it is a necessary change.  Details can be polished later.

wlammen (approved):
This approval does not assert a full error free state, but that the page received a necessary and considerable update that is way more inline with the current proof of ruc.
Edit:  It replaces content that is outdated to a degree that demands urgent measures.

PR #3932, opened 2024-04-18
fix bad HTML in <HTML> sections

### LINKED ISSUE #3928: </table> tags in AS's mathbox are outside <html> blocks
The </table> tags in the descriptions of many theorems in Alan Sare's (eg [~hbra2VD](https://us.metamath.org/mpeuni/hbra2VD.html) are outside of the <html> blocks, which causes problems for software that interprets <html> blocks, like mm-web-rs.

digama0:
Fixes #3928. The new HTML validator (metamath/metamath-knife#157) caught a bunch of bad HTML. This PR was made by taking the HTML output from theorems flagged by the linter and passing it to https://validator.w3.org/nu/ for a more precise error message.


tirix (commented):
LGTM

tirix (approved):

set.mm
```diff
@@ -427887,7 +427879,7 @@ Sections of any level are separated by two blank lines (if there is a
        at the end of lines.
        </li>
        </ul>
-       </p>
+       <p>
```
icecream17, line 427882: _(Non blocking since this has no effect on content and this pr just fixes the html errors)_

`<p>` followed by `</li>` didn't seem right... but it is... ("end tag may be omitted if ... there is no more content in the parent element")

I now know that html is fascinatingly lenient (this gives me lots of ideas), though a future change can probably just remove the (empty) paragraph tag. I didn't check for other cases of leniency.

icecream17 (approved):

set.mm
digama0: I originally just removed this tag, but the resulting HTML was missing a paragraph break after the list. (Apparently, browsers interpreted this `</p>` as `<p></p>`, and likewise `<p>` is interpreted as `<p></p>`.) Note that the original version was actually considered a dangling (unclosed) tag even though there is a `<p>` earlier, because `<p>` tags cannot contain block elements like `<ul>`, so the paragraph is automatically closed there and the `</p>` is orphaned. It's actually somewhat conventional to never use `</p>` and just use `<p>` at the start of every paragraph.

avekens (approved):

set.mm
jkingdon: Relying on your HTML parser to auto-close tags is a bit hazardous, because it doesn't always put the close tag where you think it will (the example above about how a `<ul>` will cause a currently-open `<p>` to be closed is a good example - chances are after you close the ul you want another `<p>` tag but it might not be apparent).

Having said that I don't have super strong opinions about how we write the HTML, except that I agree that having tools to run some validation is a good idea.

icecream17: Now that I think about it more, placing a `<br>` after the list is probably slightly more semantic than having an empty paragraph.
I'll do this in my next pr

PR #3936, opened 2024-04-18
Fix failing checks (urgent)

GinoGiotto:
The verifiers are failing after the merge of https://github.com/metamath/set.mm/pull/3923. I shouldn't have deleted [cbvrabcsfw](https://us.metamath.org/mpeuni/cbvrabcsfw.html). Sorry for this inconvenience.

Summary: cbvrabcsfw had no usages when I made the PR, but it got picked up in the last minimization activity, which was merged before https://github.com/metamath/set.mm/pull/3923, bypassing the verifiers (and unfortunately I didn't notice).


wlammen (approved):

wlammen:
I don't remember something like this happening before.  But you're right, cbvrabv2w uses cbvrabcsfw, the latter is missing, so set.mm is inconsistent.  This is a hole in our security concept.  Hopefully not many besides me forked or downloaded that broken state during the past 4, 5 hours.
Difficult to solve without a grace period, in which theorems are first marked as bound for incompatible modification/deletion.  No risk, no fun... (like this event)

digama0:
We could enable github merge queues, this re-runs the CI on every PR after rebasing it and before merging.

wlammen:
We routinely execute maintenance updates like #3927 that affects all of set.mm, so a conflict potential is out there.  If it comes with no costs, we should address it.

GinoGiotto:
From [github docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue):

> Using a merge queue is particularly useful on branches that have a relatively high number of pull requests merging each day from many different users.

Apparently this feature is supposed to target highly busy repositories, and by reading a few comments around it seems people agree that this added complexity is not worth it in other cases. I propose to do a one month trial, accompanied by a brief announcement on the mailing list. After that trial, people will be asked whether this feature has been an annoyance or if it's worth keeping.

PR #3937, opened 2024-04-19
Mathbox: Add aks subsublemma factorial 2x+3

metakunt:


metakunt:
I'm quite happy with that one. I'll rewrite the result from my mathbox with the newly proven facp2.

icecream17 (approved):

wlammen:
I'd rather see the proof completely developed in your mathbox.  Before moving something to Main we need to know that it is either of principal value, or that it allows for shortenings in at least 7 theorems.  IMO neither criteria is fulfilled so far.  In addition, the details of the theorem, and how (whether?) it is used in your theorems might not be fully decided by now.  This might introduce a sort of instability which we try to avoid in Main.

So please leave theorems developed to support your goals in your mathbox.  If you have convincing reasons to add something to Main, do that in a separate pull request, where we can assess your ideas.

metakunt:
Alright, I can move it to my mathbox

wlammen (approved):

PR #3948, opened 2024-04-22
tweak paragraph breaks in ~conventions

digama0:
Continuation of #3932. The [`conventions`](https://us.metamath.org/mpeuni/conventions.html) page is huge, and there aren't enough paragraph breaks in it, especially in the giant bulleted list. I tried to use a more consistent style which separates paragraph-sized things with actual `<p>` tags.

Also, I removed the `</p>` tags because they are basically not needed at all in this kind of code (because they aren't styled) and they will auto-close in the right place. Adding `</p>` tags manually is error prone because adding block elements into the paragraph causes it to fail (this is the issue that the new HTML parser metamath/metamath-knife#157 found).


wlammen (approved):

tirix (approved):

avekens (approved):

jkingdon:
> Also, I removed the `</p>` tags because they are basically not needed at all in this kind of code (because they aren't styled) and they will auto-close in the right place. Adding `</p>` tags manually is error prone because adding block elements into the paragraph causes it to fail

Hmm, interesting. I guess I don't have a super strong opinion about how to handle this but I suspect they auto-close at the first block level element, rather than auto-close where people might assume they auto-close.

Having said that, if there isn't a bottom margin (or the like) on a `p` element, and there's no javascript reacting to where the `p` tag closes, maybe this isn't something we need to worry about.

set.mm
```diff
@@ -426182,18 +426182,17 @@ number of vertices in a friendship graph is (k(k-1)+1), see
...
+       language (skipping the many cases where they are identical).
        For more specific conventions, see:
        <ul>
        <li> ~ conventions-labels : conventions related to labels</li>
        <li> ~ conventions-comments : conventions related to comments</li>
        </ul>
 
        <ul>
-       <li><b>Notation.</b>
+       <li><p><b>Notation.</b>
```
jkingdon, line 426195: What is the `<p>` here for? Are we better off putting some style on the `<li>`?

set.mm
```diff
@@ -426723,17 +426720,15 @@ to the label of the original theorem (for instance, ~ con3i is the
...
+       <p>"Deduction form" is the preferred form for theorems because this form
        allows us to easily use the theorem in places where (in traditional
        textbook formalizations) the standard Deduction Theorem (see below)
        would be used. We call this approach <b>"deduction style"</b>.
        In contrast, we usually avoid theorems in "inference form" when that
        would end up requiring us to use the deduction theorem.
-       <br>
 
-       Deductions have a label suffix of "d", especially if there are other
+       <p>Deductions have a label suffix of "d", especially if there are other
```
jkingdon, line 426731: Using multiple paragraphs with `p` tags is better than `br` here, agreed.

jkingdon (approved):

set.mm
digama0: (I recommend viewing this in a browser.) It adds paragraph breaks between the list items. We could do it with css, e.g. `style="margin-bottom:5px"` on the `li`, but some of the list items are multiple paragraphs so this also starts interacting with the global styles.

I want to majorly overhaul the way styling is done in metamath and set.mm, but the first step of that is to make mm-web-rs the official web site generation mechanism, so that we can drop all the regex HTML munging.

PR #3955, opened 2024-04-24
Add decbld* to mathbox

metakunt:
I think these will amortize quickly, as I will be calculating bigger values.


wlammen (approved):

avekens (approved):
Is there any (intended) use for these theorems, or are they just exercises?

tirix (approved):
It seems you missed the case ending in `0`?

metakunt:
@avekens There are intended uses, mainly when I want to calculate gcd's
For example I want to calculate gcd of 420 and 8, I can now simply use 4enn, decbld2, decnncl2 instead of having to prove that 2 e. NN I can save an extra step. @tirix No, I don't https://us.metamath.org/mpeuni/decnncl2.html is already implemented.
I am still debating whether I will need the fact that 1 to 9 is an integer later on. This gives me a slightly better autocomplete experience.

I think your point is valid, maybe I will need the fact at some point that 1 to 9 is a natural number anyways for some intermediate calculation. So I think for now I will close this. 

I will be adding inference versions of commonly used theorems for natural numbers. This one will save me sanity

Issue #3947, opened 2024-04-22
Interchanging factors in integrals and sums (Lemmas for AKS)


metakunt:
@digama0 has proven itgfsum. For my lemma, I need to prove that I can interchange integrals and sums provided that the correct conditions hold.

I have already found useful lemmas for future work, like integration by parts. However I'd need to interchange sums and integral.

I have done a writeup of a lemma I want to prove. I don't even know if those conditions are correct. The qed step could have likely a stronger result regarding the sums. I am yet unsure if I need that a stronger result as in itgfsum, but I am confident that I need the result below.
```
h1::itgfsum3factcom12.1 |- ( ph -> A e. dom vol )
h2::itgfsum3factcom12.2 |- ( ph -> B e. Fin )
h3::itgfsum3factcom12.3 |- (  ( ph /\ ( x e. A /\ k e. B ) )
                      -> C e. V )
h4::itgfsum3factcom12.4 |- ( ( ph /\ k e. B ) -> ( x e. A |-> C ) e. L^1 )
h5::itgfsum3factcom12.5 |- (  ( ph /\ ( x e. A /\ k e. B ) )
                      -> D e. W )
h6::itgfsum3factcom12.6 |- ( ( ph /\ k e. B ) -> ( x e. A |-> D ) e. L^1 )
h7::itgfsum3factcom12.7 |- (  ( ph /\ ( x e. A /\ k e. B ) )
                      -> E e. U )
h8::itgfsum3factcom12.8 |- ( ( ph /\ k e. B ) -> ( x e. A |-> E ) e. L^1 )

qed::    |- (  ph ->  S. A ( C x. sum_ k e. B ( D x. E ) ) _d x = sum_ k e. B ( D x. S. A ( C x. E ) _d x  ) )
                      
```

For my lemma
```math
A = ( 0 , 1 )
```
```math
B = \{0,\dots, n-m\}
```
```math
C = x^{m-1}
```
```math
D=(-1)^{k}\cdot\binom{n-m}{k}
```
```math
E=x^{k}
```
My conditions should hold. The functions are all bounded on $(0,1)$ and the sum is finite.
Lemma is on top of page 4: https://www3.nd.edu/~andyp/notes/AKS.pdf
@digama0 has also shown in itgfsum that under those conditions the sum is in $L^1$. I assume similar conditions will hold for this lemma.

[...]

PR #3959, opened 2024-04-26
First lemma in lcm inequality proof

metakunt:
I've figured out how to do https://github.com/metamath/set.mm/issues/3947

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -108,6 +108,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+26-Apr-24 bddiblnc  [same]      moved from BL's mathbox to main set.mm
+26-Apr-24 cnicciblnc  [same]    moved from BL's mathbox to main set.mm
 21-Apr-24 nnindd    [same]      moved from TA's mathbox to main set.mm
 17-Apr-24 syl6eqel  eqeltrdi    compare to eqeltri or eqeltrd
 13-Apr-24 unitresl  olcnd


metakunt:
As promised, I won't push any more changes unless they are requested.

set.mm
```diff
@@ -632529,6 +632529,116 @@ fixed reference functional determined by this vector (corresponding to
...
 
+  ${
+    $d A k x $.  $d A q r x $.  $d B k x $.  $d F q r $.  $d G q r x $.
+    $d H q r $.  $d L q r $.  $d U q r $.  $d k ph x $.  $d ph q x $.
+    3factsumint1.1 $e |- A = ( L [,] U ) $.
+    3factsumint1.2 $e |- ( ph -> B e. Fin ) $.
+    3factsumint1.3 $e |- ( ph -> L e. RR ) $.
+    3factsumint1.4 $e |- ( ph -> U e. RR ) $.
+    3factsumint1.5 $e |- ( ( ph /\ x e. A ) -> F e. CC ) $.
+    3factsumint1.6 $e |- ( ph -> ( x e. A |-> F ) e. ( A -cn-> CC ) ) $.
```
tirix, line 632540: I believe the first hypothesis:
```
|- ( ( ph /\ x e. A ) -> F e. CC )
```
Can be derived from the second:
```
|- ( ph -> ( x e. A |-> F ) e. ( A -cn-> CC ) )
```
See for example ~ cncff

metakunt: Ah, well time to fix it.

tirix (approved):
Wow, that's a lot of progress since last week! Way to go!

metakunt:
Thanks a lot. Yeah metamath has some insane cliffs you have to jump over. Once you understand which theorems help and learn the diverse equality inference theorems it gets easier.

Honestly, I'll keep the hypotheses on the sublemmas and only remove it on the final lemma since the sublemmas shouldn't be used anyways.

set.mm
tirix: I would have fixed them from the last to the first, so that not everything is broken.

For fixing a theorem, just reopen the proof file, remove the "first" hypothesis, and fill the blank.
This should work:
```
h10::         |- ( ph -> ( x e. A |-> F ) e. ( A -cn-> CC ) )
20::cncff            |- ( ( x e. A |-> F ) e. ( A -cn-> CC ) -> ( x e. A |-> F ) : A --> CC )
30:10,20:syl        |- ( ph -> ( x e. A |-> F ) : A --> CC )
40:30:fvmptelrn   |- ( ( ph /\ x e. A ) -> F e. CC )
```

metakunt:
Ah cool, it's fixed. I did it manually by deleting the hypotheses in set.mm and rewriting the proof. It's fine that the hypotheses stay in 1-4, those lemmas shouldn't be used anyways. Now I can prove the equation
```math
\int_0^1 x^{m-1} \sum_{k=0}^{n-m} (-1)^{k} \binom{n-m}{k}\cdot x^k =
 \sum_{k=0}^{n-m} (-1)^{k} \binom{n-m}{k} \int_0^1 x^{m-1}\cdot x^k
```

Now I need to satisfy the hypotheses 1-7:
A = [0,1] Ok
B = ( 0 ... ( N - M ) ) e. Fin Ok
L = 0 -> 0 e. RR Ok
U = 1 -> 1 e. RR Ok
The difficulty will be satisfying the final 3.
I can use the binomial coefficient operation to get  https://us.metamath.org/mpeuni/df-bc.html
I guess G = ( ( ( -u 1 ) ^ k ) x. ( ( N - M ) _C k ) )
so that one should be fine. The problem will be F and H. I will need to prove that both are continuous on the restriction where I've defined it.

Also later on I will need several helpful lemmas regarding differentiation and integration. I know that those already exist, but I still would like to know what the easiest way would be to get started quickly. I need continuous function builders and ways to integrate easily, what are the lemmas I should focus on learning now?

tirix:
> The problem will be F and H. I will need to prove that both are continuous

So you have $F(x) = x^{m-1}$ and $H(x) = x^k$, right?
I believe [`~cxpcncf1`](https://us.metamath.org/mpeuni/cxpcncf1.html) should work for both.
Or simpler: [`~expcncf`](https://us.metamath.org/mpeuni/expcncf.html).

set.mm
```diff
@@ -632529,6 +632529,118 @@ fixed reference functional determined by this vector (corresponding to
...
+    3factsumint.2 $e |- ( ph -> B e. Fin ) $.
+    3factsumint.3 $e |- ( ph -> L e. RR ) $.
+    3factsumint.4 $e |- ( ph -> U e. RR ) $.
+    3factsumint.5 $e |- ( ph -> ( x e. A |-> F ) e. ( A -cn-> CC ) ) $.
+    3factsumint.6 $e |- ( ( ph /\ k e. B ) -> G e. CC ) $.
+    3factsumint.7 $e |- ( ( ph /\ k e. B ) -> ( x e. A |-> H ) e.
+     ( A -cn-> CC ) ) $.
+    $( First equation in lcm inequality proof.  (Contributed by metakunt,
+       26-Apr-2024.) $)
+    3factsumint $p  |- ( ph -> S. A ( F x. sum_ k e. B ( G x. H ) ) _d x =
```
avekens, line 632632: What is the meaning of this equation? It should be described in more detail (especially since the  "lcm inequality proof" is not available yet).

metakunt: I have added documentation outlining the idea behind this theorem

tirix: The link seems to be missing a tilde ~ : [https://www3.nd.edu/~andyp/notes/AKS.pdf](https://www3.nd.edu/~andyp/notes/AKS.pdf)

metakunt: I added the tilde back, but after rewrapping it just looks wrong. I don't know how metamath.exe handles tildes in links

metakunt: I might have to change tilde to %7e, don't I?

wlammen: %7e works, at least

wlammen (approved):

tirix:
This looks good to me, can we merge once we have @avekens's final approval?

set.mm
avekens: To be honest, I did not find this theorem in the referenced documentation. 
But maybe everything becomes clearer after more theorems are available for this topic. 

Since this is currently in a mathbox only, I will accept PR.

avekens (approved):

set.mm
tirix: I saw the equations on which @metakunt is working in the proof of theorem 3.1.

PR #3962, opened 2024-04-29
update wl-df-had

wlammen:
1. Move ifpdfbi with shortened proof from RP's mathbox to Main.  I need it.
2. Add convenience theorem ifpbi23d, and shorten proofs with it.
3. Mathbox: Update df-wl-had
4. Delete outdated OLD theorems

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -108,6 +108,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+29-Apr-24 ifpdfbi   [same]      moved from RP's mathbox to main set.mm
 21-Apr-24 nnindd    [same]      moved from TA's mathbox to main set.mm
 17-Apr-24 syl6eqel  eqeltrdi    compare to eqeltri or eqeltrd
 13-Apr-24 unitresl  olcnd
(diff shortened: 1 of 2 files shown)


tirix (approved):
Ok for me. 

`~ifpbi23d` is a two-step proof, which I usually don't like but here the 2-3 variant makes sense (more than the 1-3 or 2-3 variants).

wlammen:
Fix merge conflict, and apply some corrections to comments.

@tirix In fact all kinds of combinations of ifpbi*d are found in the wild, 1-3 as well as 1-2.  If one thinks from simple to complex, ifpbi1d to ifpbi3d are the primitives, on which ifpbi123d (and the others) is finally built upon.  Using ifpbi123d as a foundation is a one-size-fits-it-all solution.  It is not particularly application friendly, because you rarely need to exchange all of the parameters to if-(...).

metakunt (approved):
Honestly, I like that there are more theorems, even when they appear not to be that useful or overly constraining, especially when they are in deduction form. The more theorems we have, the easier autocomplete will work, and the easier autocomplete works, the faster I can work on new results. Add them all, someone will need it anyways, especially when you have a lot of useful theorems in your mathbox that I can yoink to main so I can save myself 12 hours of embarassement.

PR #3970, opened 2024-05-02
alternative definition of cad

wlammen:
1. Updating comments and reordering of theorems in the wl-df-had section;
2. Adding an if based definition of wcad, wl-df-cad, along the lines of wl-df-had.


set.mm
```diff
@@ -546008,17 +546008,37 @@ extends this idea to multiple additions (TODO).
   $( Alternative definition of ~ whad based on ~ hadifp .  See ~ df-had to
      learn how it is currently introduced.
 
-     Check whether an odd number of inputs are true.  The oddness of two is
-     expressed by the ` \/_ ` operation, and for evenness we can use ` <-> ` .
+     Check whether an odd number of three inputs are true.  The oddness of two
```
avekens, line 546011: "The oddness of two" sounds odd - what does it mean?

wlammen: done

avekens (approved):

PR #3971, opened 2024-05-04
df-ich builders; sbtd

icecream17:
random DV removed

ichan now only uses ax-mp--4 (6 axioms saved)

--- comment
ichtd would currently use ichbidv with `( true <-> ps )`, then icht to eliminate the true
I don't see this being useful, and the df-ich section is in AV's mathbox (sbtd is in my mathbox)

ichim shows how more builders like ichor could be made


avekens:
> --- comment ichtd would currently use ichbidv with `( true <-> ps )`, then icht to eliminate the true I don't see this being useful, and the df-ich section is in AV's mathbox (sbtd is in my mathbox)

Maybe it is time now to move section "Interchangeable setvar variables" from my mathbox into the main body of set.mm. Maybe after section 1.4.8  "Define proper substitution" would be a good place.  Then @icecream17 can continue to work on this topic without depending on me (and the rules that mathboxes must be independent).

avekens:
Why is this PR still pending? Maybe because there are no remarks/opinions about the movement of df-ich and related theorems to main by other reviewers. If there are no objections, then @icecream17 should move df-ich and related theorems to main (maybe in a separate PR, if this is more practicable).
I would kindly ask for others` opinions on this topic.

wlammen:
I was expecting your approval, as this is (for now) affecting your mathbox.

wlammen (approved):
as is, without moving the symmetry stuff to main.

PR #3972, opened 2024-05-04
mathbox: add wl-cadan

wlammen:
1. mathbox: add wl-cadan.  This opens up access to all other cad theorems in main.
2. delete outdated OLD theorems
3. Revise sbcbi2 as suggested by @icecream17


set.mm
```diff
@@ -36182,14 +36182,6 @@ practical reasons (to avoid having to prove sethood of ` A ` in every use
        ( [. A / x ]. ph <-> [. A / x ]. ps ) ) $=
     ( wb wal nfa1 sp sbcbid ) ABEZCFABCDJCGJCHI $.
 
-  $( Obsolete proof of ~ sbcbi2 as of 4-May-2023.  (Contributed by Giovanni
-     Mascellani, 9-Apr-2018.)  (Proof modification is discouraged.)
-     (New usage is discouraged.) $)
-  sbcbi2OLD $p |- ( A. x ( ph <-> ps ) ->
-       ( [. A / x ]. ph <-> [. A / x ]. ps ) ) $=
-    ( wb wal cab wcel wsbc wceq abbi eleq2 sylbi df-sbc 3bitr4g ) ABECFZDACGZHZ
-    DBCGZHZACDIBCDIPQSJRTEABCKQSDLMACDNBCDNO $.
```
icecream17, line 36191: changing this proof to use abbi1 saves ax-10 and ax-12 on sbcbi2
(optional, I can do it in my pr)

icecream17 (approved):

set.mm
wlammen: good catch

tirix (approved):

PR #3975, opened 2024-05-05
zeroes and inverses (ax-mulcom pr 6)

icecream17:
main results:
+ resubeqsub
+ sn-addid1
+ sn-mul01

--- comment

now almost every important fact about 0 is known, except 0a = 0

the eqtrd and eqtr3d in sn-addid0 could be combined probably; an auto-minimize idea for future me


tirix (approved):
Nice!

wlammen (approved):

PR #3978, opened 2024-05-08
mathbox: rename had -> 3xor

wlammen:
1. Mathbox:  The operation https://us.metamath.org/mpeuni/df-had.html is actually a triple xor operation.  The name "had" resembles somewhat "half adder", but https://www.geeksforgeeks.org/half-adder-in-digital-logic/ shows that in electronics such a circuit has 2 inputs and 2 results, not in accordance with df-had.  Later "had" is referred to as "adder sum", but this is a vague term as well.  I avoid this term in my mathbox and replace it with 3xor, for triple xor.
If one want to keep the idea of an addition, "wrap around addition" can be used instead to find a name https://www.gnu.org/software/gnulib/manual/html_node/Wraparound-Arithmetic.html
2. Rewrite the definition such, that its semantics is clearly seen, and that recursion can be used to extend it to more inputs.
3. Delete outdated OLD theorems

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -3162,7 +3162,6 @@
 "cbv3" is used by "cbv1".
 "cbv3" is used by "cbv3h".
 "cbv3" is used by "cbval".
-"cbvab" is used by "cbvabvOLD".
 "cbvab" is used by "cbvrab".
 "cbvab" is used by "cbvrabcsf".
 "cbvab" is used by "cbvsbc".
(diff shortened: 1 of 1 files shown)


tirix (approved):
I agree with your remark about the half adder having 2 inputs and 2 outputs in electronics, and the direction to rename `had` into `3xor`.

Ideally I think we should be consistent and rename the `hadd` symbol everywhere into something more like `3xor`, either `3xor` itself or a syntax similar to `3or` like `( ph \/_ ps \/_ ch )`, dropping the parenthesis.

wlammen:
I'll open an issue soon regarding the name of df-had.

PR #3979, opened 2024-05-08
The Ackermann function

avekens:
The definition of the Ackermann function is provided as proposed by @tirix in the Google group (see https://groups.google.com/g/metamath/c/OC6epyquy5U/m/qoS7HE2gBQAJ):

First, a function iterator `IterComp` regarding composition is defined (see ~df-itco) and related theorems are provided. As an illustration see ~itcoval3: ``( ( IterComp ` F ) ` 3 ) = ( F o. ( F o. F ) ) )``.

Based on this iterator, the Ackermann function `Ack` is defined (see ~df-ack). With this definition and some basic theorems, some initial values of the Ackermann function are calculated (see ~ackval0012, ~ackval1012, ackval2012, ~ackval3012), and that the definition corresponds to Péter's definition is shown (see ~ackval0val, ~ackvalsuc0val, ~ackvalsucsucval).

For additional information see the header comment of section "The Ackermann function".

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -108,6 +108,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 8-May-24 fexd      [same]      moved from GS's mathbox to main set.mm
 29-Apr-24 elunitcn  [same]      moved from TA's mathbox to main set.mm
 29-Apr-24 elunitrn  [same]      moved from TA's mathbox to main set.mm
 29-Apr-24 unitsscn  [same]      moved from TA's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -18391,6 +18391,7 @@ New usage of "sbtrt" is discouraged (1 uses).
 New usage of "sbtvOLD" is discouraged (0 uses).
 New usage of "scmateALT" is discouraged (0 uses).
 New usage of "seq1hcau" is discouraged (0 uses).
+New usage of "seqp1iOLD" is discouraged (0 uses).
 New usage of "setrec1lem1" is discouraged (3 uses).
 New usage of "setrec1lem2" is discouraged (1 uses).
 New usage of "setrec1lem3" is discouraged (1 uses).
@@ -20338,6 +20339,7 @@ Proof modification of "sbtALT" is discouraged (12 steps).
 Proof modification of "sbtT" is discouraged (7 steps).
 Proof modification of "sbtvOLD" is discouraged (31 steps).
 Proof modification of "scmateALT" is discouraged (236 steps).
+Proof modification of "seqp1iOLD" is discouraged (20 steps).
 Proof modification of "sgrpplusgaopALT" is discouraged (82 steps).
 Proof modification of "simplbi2VD" is discouraged (24 steps).
 Proof modification of "simplbi2comtVD" is discouraged (42 steps).


david-a-wheeler:
Awesome! I think the Ackermann function can also serve as an example of how to define these kinds of functions in set.mm.

So I suggest also modifying the `conventions` discussion on **Function definition**. It currently says, "Functions are typically defined by first defining the constant symbol (using $c)...  and at least one simple value (e.g., ~ cos0 )." How about appending something like this to the **Function definition** section?:

> For an example of how to define functions that aren't primitive recursive, see the Ackermann function definition ~ df-ack and the related material discussing it.

savask:
That's a nice contribution! I have one comment.

`seqp1d` is almost identical to `seqp1i`, the only difference is that the hypothesis `seqp1i.2` isn't in the deduction form. Since the proof of `seqp1i` is now an immediate corollary of `seqp1d`, maybe it's worth deleting `seqp1i` altogether and replacing all its uses by `seqp1d`?

avekens:
> Awesome! I think the Ackermann function can also serve as an example of how to define these kinds of functions in set.mm.
> 
> So I suggest also modifying the `conventions` discussion on **Function definition**. It currently says, "Functions are typically defined by first defining the constant symbol (using $c)... and at least one simple value (e.g., ~ cos0 )." How about appending something like this to the **Function definition** section?:
> 
> > For an example of how to define functions that aren't primitive recursive, see the Ackermann function definition ~ df-ack and the related material discussing it.

Good idea, I will add this with my next commit.

avekens:
> That's a nice contribution! I have one comment.
> 
> `seqp1d` is almost identical to `seqp1i`, the only difference is that the hypothesis `seqp1i.2` isn't in the deduction form. Since the proof of `seqp1i` is now an immediate corollary of `seqp1d`, maybe it's worth deleting `seqp1i` altogether and replacing all its uses by `seqp1d`?

I have thought about this, too. ~seqp1i is used 7 times, so it is not much work to replace it. If there are no objections, I will perfom the replacement with my next commit.

set.mm
```diff
@@ -756151,6 +756179,534 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  cack $a class Ack $.
+
+  ${
+    $d f g i j $.
+    $( Define a function (recursively) that returns the n-th iterate of a class
+       (usually a function) with regard to composition.  (Contributed by Tierry
+       Arnoux, 28-Apr-2024.)  (Revised by AV, 2-May-2024.) $)
+    df-itco $a |- IterComp = ( f e. _V
+         |-> seq 0 ( ( g e. _V , j e. _V |-> ( f o. g ) ) ,
+                     ( i e. NN0 |-> if ( i = 0 , ( _I |` dom f ) , f ) ) ) ) $.
```
tirix, line 756244: Very interesting, thanks for the attribution!
Indeed I see now that the initial step I proposed was too simplistic...

set.mm
```diff
@@ -756151,6 +756179,534 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  $}
+
+  ${
+    $d f i j n $.
+    $( Define the Ackermann function (recursively).  (Contributed by Tierry
+       Arnoux, 28-Apr-2024.)  (Revised by AV, 2-May-2024.) $)
```
tirix, line 756250: (it's spelled "Thierry" with a silent H)

set.mm
```diff
@@ -756151,6 +756179,534 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+    df-ack $a |- Ack = seq 0 ( ( f e. _V , j e. _V |-> ( n e. NN0
+                             |-> ( ( ( IterComp ` f ) ` ( n + 1 ) ) ` 1 ) ) ) ,
+                               ( i e. NN0 |-> if ( i = 0 ,
+                                        ( n e. NN0 |-> ( n + 1 ) ) , i ) ) ) $.
```
tirix, line 756254: Interestingly, the part `if ( i = 0 , ( n e. NN0 |-> ( n + 1 ) ) , i )` does not seem to be homogeneous:
* if `i = 0`, this is a function on `NN0`,
* otherwise this is the integer `i`

I think that "else" branch is actually not needed anywhere.

set.mm
```diff
@@ -756151,6 +756179,534 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
+
+  According to Wikipedia ("Ackermann function", 8-May-2024,
+  ~ https://en.wikipedia.org/wiki/Ackermann_function ): "In computability
+  theory, the Ackermann function, named after Wilhelm Ackermann, is one of the
+  simplest and earliest-discovered examples of a total computable function that
+  is not primitive recursive. ... One common version is the two-argument
+  Ackermann-P&eacute;ter function developed by R&oacute;zsa P&eacute;ter and
+  Raphael Robinson.  Its value grows very rapidly; for example, ` A ( 4 , 2 ) `
+  results in ` 2 ^ ; 6 5 5 3 6 - 3 `, an integer of 19,729 decimal digits."
```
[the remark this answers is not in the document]
savask: > Finally, this statement calls for a proof, but I guess it would really be tedious ;-)

Alexander can correct me, but I actually think it's not that tedious. `A(4, 2) = A(3, A(4, 0)) = A(3, A(3, 1))` and this pull request already finds the value of `A(3, 1)` and gives a formula for `A(3, n)`. The power `2 ^ ; 1 6` which appears in the exponent has been already computed in Metamath theorem `2exp16`.

set.mm
avekens: > Interestingly, the part `if ( i = 0 , ( n e. NN0 |-> ( n + 1 ) ) , i )` does not seem to be homogeneous:
> 
>     * if `i = 0`, this is a function on `NN0`,
> 
>     * otherwise this is the integer `i`
> 
> 
> I think that "else" branch is actually not needed anywhere.

Yes, this is exactly the oddity I mentioned in my comment in the Google group.

set.mm
avekens: Sorry, this is a typo - I'll correct it.

set.mm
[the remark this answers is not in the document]
avekens: Yes, i am just about to prove this and provide it with a separate PR. But since it is mentioned here, I will add it to this PR.

tirix (approved):

set.mm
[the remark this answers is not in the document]

PR #3980, opened 2024-05-09
Intuitionize trigonometry to the end of the Properties of pi section

jkingdon:
Mostly this is just noting what would be involved in proving these theorems. A few are proved.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11455,6 +11455,13 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>the set.mm proof uses sineq0</td>
 </tr>
 
+<tr>
+  <td>cosord</td>
+  <td>~ cosordlem , ~ cos11</td>
+  <td>presumably provable but the set.mm proof
+  uses real number trichotomy</td>
+</tr>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11474,6 +11474,12 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>the set.mm proof depends on cosord</td>
 </tr>
 
+<tr>
+  <td>recosf1o</td>
+  <td>~ ioocosf1o</td>
+  <td>the set.mm proof of recosf1o relies on ivthle2</td>
+</tr>
+
 </TABLE>
 
 <HR NOSHADE SIZE=1><A NAME="bib"></A><B><FONT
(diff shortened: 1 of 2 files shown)


mmil.raw.html
```diff
@@ -11455,6 +11461,49 @@
...
+  <td>sinord</td>
+  <td><i>none</i></td>
+  <td>the set.mm proof depends on cosord</td>
+</tr>
+
+<tr>
+  <td>recosf1o</td>
+  <td>~ ioocosf1o</td>
+  <td>the set.mm proof of recosf1o relies on ivthle2</td>
+</tr>
```
tirix, line 11481: Isn't there any other way to go around this?

Since you have the strict version of the intermediate value theorem, can't you e.g. at least prove a similar theorem on open intervals?

tirix (approved):

iset.mm
```diff
@@ -21226,9 +21226,10 @@ effectively bound by a quantifier or something that expands to one (such
     $d y z A $.  $d x z $.
     rgen2a.1 $e |- ( ( x e. A /\ y e. A ) -> ph ) $.
     $( Generalization rule for restricted quantification.  Note that ` x ` and
-       ` y ` needn't be distinct (and illustrates the use of ~ dvelimor ).
+       ` y ` are not required to be disjoint.  This proof illustrates the use
+       of ~ dvelim .  Usage of ~ rgen2 instead is highly encouraged.
        (Contributed by NM, 23-Nov-1994.)  (Proof rewritten by Jim Kingdon,
-       1-Jun-2018.) $)
+       1-Jun-2018.)  (New usage is discouraged.) $)
```
avekens, line 21232: Why not also "(Proof modification is discouraged.)" as in set.mm?

mmil.raw.html
```diff
@@ -11455,6 +11461,49 @@
...
+<tr>
+  <td>tanord1</td>
+  <td><i>none</i></td>
+  <td>the set.mm proof relies on ltord1</td>
+</tr>
+
+<tr>
+  <td>tanord</td>
+  <td><i>none</i></td>
+  <td>the set.mm proof relies on tanord , real number trichotomy, and ltord1</td>
```
avekens, line 11498: tanord relies on tanord?

mmil.raw.html
jkingdon: ioocosf1o IS a similar theorem on open intervals.

I suspect that might not be enough long-term. But it is what I could clearly see how to prove, and included in this pull request.

iset.mm
jkingdon: The proof is totally different from the set.mm proof (unless they are similar by coincidence), so I didn't see any reason to copy that over.

mmil.raw.html
jkingdon: Good eyes. The second one should be `tanord1`. I've updated the pull request with that change.

avekens (approved):

PR #3981, opened 2024-05-11
recursion for had, cad

wlammen:
1. mathbox: define df-had (renamed to df-3xor in my mathbox) as a particular case of an ifp-based recursion scheme.
2. mathbox: define df-cad (renamed df-3min2 in my mathbox) as a particular case of an ifp-based recursion scheme.
3. delete outdated OLD theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17321,8 +17321,6 @@ New usage of "nd2" is discouraged (4 uses).
 New usage of "nd4" is discouraged (1 uses).
 New usage of "negexsr" is discouraged (0 uses).
 New usage of "nelbOLD" is discouraged (0 uses).
-New usage of "nelne1OLD" is discouraged (0 uses).
-New usage of "nelne2OLD" is discouraged (0 uses).
 New usage of "nf5rOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -545938,24 +545825,31 @@ extends this idea to multiple additions (TODO).
...
-     to an odd number.
+     two remaining inputs need to amount to an even (or: not an odd) number,
+     else to an odd number.
+
+     The idea of an odd number of inputs being true carries over to other than
+     3 inputs by recursion:  In an informal notation we depend the case with
+     n+1 inputs, ` ph ` being the additional one, recursively on that of n
+     inputs:  "(n+1)-xor" ` <-> if- ( ph , -. ` "n-xor" ` , ` "n-xor" ` ) ` .
+     The base case is "0-xor" being ` F. ` , because zero inputs never contain
+     an odd number among them.  Then we find, after simplifacations, in our
```
icecream17, line 545836:
```suggestion
     an odd number among them.  Then we find, after simplifications, in our
```
or "simplifying"

icecream17 (approved):
refreshingly understandable recursion/inductions

PR #3982, opened 2024-05-12
Replace syl5rbbr with bitr3di in set.mm and iset.mm

jkingdon:
My thinking is that the next steps for the proposed syl renames are:

```
proposed  syl5bb    bitrid      compare to bitri or bitrd
proposed  syl5rbb   bitr2id     compare to bitr2i or bitr2d
proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
proposed  syl5rbbr  bitr3di     hypotheses would be switched
proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
proposed  syl5req   eqtr2id     compare to eqtr2i or eqtr2d
proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
proposed  syl5reqr  eqtr3di     hypotheses would be switched
proposed  syl6bb    bitrdi      compare to bitri or bitrd
proposed  syl6rbb   bitr2di     compare to bitr2i or bitr2d
proposed  syl6bbr   bitr4di     compare to bitr4i or bitr4d
proposed  syl6rbbr  bitr4id     hypotheses would be switched
proposed  syl6eq    eqtrdi      compare to eqtri or eqtrd
proposed  syl6req   eqtr2di     compare to eqtr2i or eqtr2d
proposed  syl6eqr   eqtr4di     compare to eqtr4i or eqtr4d
proposed  syl6reqr  eqtr4id     hypotheses would be switched
```

and that these ones are not controversial or otherwise in need of further discussion.
(Many of the others I'm not so sure about, so in fact this might be the last batch
of the syl renames, but that discussion can be for another day).

Details about this pull request in particular:

* bitr3di is syl5rbbr with the order of the hypotheses switched.
* Run "minimize_with bitr3di/may_grow" on each of the theorems which had been using syl5rbbr .
* For theorems marked as "Proof modification is discouraged", use /override on PROVE and SAVE NEW_PROOF, but check that the change from syl5rbbr to bitr3di is the only modification printed in the output.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -83,7 +83,6 @@ proposed  syl5ibrcom imbitrridcom
 proposed  syl5bb    bitrid      compare to bitri or bitrd
 proposed  syl5rbb   bitr2id     compare to bitr2i or bitr2d
 proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
-proposed  syl5rbbr  bitr3di     hypotheses would be switched
 proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
 proposed  syl5req   eqtr2id     compare to eqtr2i or eqtr2d
 proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
@@ -108,6 +107,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+13-May-24 syl5rbbr  bitr3di     order of hypotheses is switched
  8-May-24 fexd      [same]      moved from GS's mathbox to main set.mm
 29-Apr-24 elunitcn  [same]      moved from TA's mathbox to main set.mm
 29-Apr-24 elunitrn  [same]      moved from TA's mathbox to main set.mm


set.mm
```diff
@@ -3165,12 +3165,12 @@ intermediate steps that are essentially incomprehensible to humans (other
   $}
 
   ${
-    syl5rbbr.1 $e |- ( ps <-> ph ) $.
```
wlammen, line 3168: non-blocking: We usually let the variables appear in (Greek) alphabetic order.  This helps when looking for a theorem doing text search in the file.  I do this occasionally.

wlammen (approved):

set.mm
jkingdon: Hmm, interesting. I'm not sure I'm committed to alphabetic order as such, but I do like the idea of having the names similar to https://us.metamath.org/mpeuni/bitr3d.html which would argue for

```
    bitr3di.1 $e |- ( ph -> ( ps <-> ch ) ) $.
    bitr3di.2 $e |- ( ps <-> th ) $.
    bitr3di $p |- ( ph -> ( ch <-> th ) ) $=
```

Sometime (probably tomorrow) I'll take a look at whether this would be easy or hard to change.

metakunt:
How did you manage to do it? Did you do it manually or is there a script that automatically renames all occurrences?

david-a-wheeler:
You might find `scripts/mass-rename` useful; it's in `set.mm`.

jkingdon:
> How did you manage to do it? Did you do it manually or is there a script that automatically renames all occurrences?

Mostly automated. I ran `show usage syl5rbbr` in metamath-exe and then copy-pasted the result to a shell to get a command like:

```sh
for i in sbco3 necon2bbid notzfaus frege92 e2ebind stoweidlem28; do
  scripts/minimize $i set.mm
done
```

where `scripts/minimize` has been edited to contain `minimize_with bitr3di/may_grow` in place of the normal minimize. Also, for theorems marked as "Proof modification is discouraged", edit the script to add /override on PROVE and SAVE NEW_PROOF, but check that the change from syl5rbbr to bitr3di is the only modification printed in the output

Because this isn't just a rename (the order of the hypotheses changes), it needs a metamath-aware tool like metamath-exe.

jkingdon:
> You might find `scripts/mass-rename` useful; it's in `set.mm`.

Good to know about (not for this pull request, which isn't just a rename, but for others).

jkingdon:
Since this has an approval and the only thing I changed since then is renaming variables as requested, I'm going to merge this.

PR #3984, opened 2024-05-13
Reduce axiom usage with ~sn-elabg

GinoGiotto:
Theorem [sn-elabg](https://us.metamath.org/mpeuni/sn-elabg.html) (here renamed ~elabgw) from @icecream17 is pretty effective. As he suggested, this PR applies it to ~eldif, ~elun, ~elin, ~elsng, ~elxpi, leading to the following reduction in axiom usage:

* Drop ax-10 from 364 theorems.
* Drop ax-11 from 399 theorems.
* Drop ax-12 from 322 theorems.

Notice that all added dv conditions are with dummy variables, so the power of the edited theorems remains the same.

Personally, it's the first time I see numbers this high for these three axioms, especially with so few changes.

Full axiom usage here: https://github.com/metamath/set.mm/commit/83a0c1b79ab1710b63bb175d4398406d7c51eef0


wlammen (approved):
Nice!

set.mm
```diff
@@ -34155,6 +34155,20 @@ Such interpretation is rarely needed (see also ~ df-ral ).  (Contributed
...
   $}
 
+  ${
+    $d A y $.  $d x y $.  $d ph y $.  $d ch y $.  $d ps x $.
+    elabgw.1 $e |- ( x = y -> ( ph <-> ps ) ) $.
+    elabgw.2 $e |- ( y = A -> ( ps <-> ch ) ) $.
+    $( Membership in a class abstraction, using implicit substitution and an
+       intermediate setvar ` y ` to avoid ~ ax-10 , ~ ax-11 , ~ ax-12 .  It
+       also avoids a disjoint variable condition on ` x ` and ` A ` .  Compare
+       ~ sbievw2 .  (Contributed by SN, 20-Apr-2024.) $)
```
benjub, line 34165: Can you replace the last sentence by e.g.,
> This is to ~ elabg what ~ sbievw2 is to ~sbievw .

which is an economical way to link to these three theorems and explain their relationship.

GinoGiotto: Done

benjub (approved):

PR #3989, opened 2024-05-16
Least common multiple inequality lemma

metakunt:


metakunt:
I will update the description once I have used this result for a different inequality.

icecream17 (approved):

set.mm
```diff
@@ -633441,8 +633548,7 @@ fixed reference functional determined by this vector (corresponding to
     3factsumint.6 $e |- ( ( ph /\ k e. B ) -> G e. CC ) $.
     3factsumint.7 $e |- ( ( ph /\ k e. B ) -> ( x e. A |-> H ) e.
      ( A -cn-> CC ) ) $.
-    $( First equation in lcm inequality proof, Theorem 3.1 from
-       ~ https://www3.nd.edu/%7eandyp/notes/AKS.pdf (Contributed by metakunt,
```
avekens, line 633445: why is this feference removed? was it wrong?

icecream17: It is correctly the first equation/step in the proof of theorem 3.1, but it seems the reference moved to ~ lcmineqlem below which is the entirety of theorem 3.1

metakunt: That's correct. The theorem is now finished and I have moved the reference to the final step.

avekens (approved):

PR #3993, opened 2024-05-18
Express the root of a number in terms of roots of reals.

arpie-steele:
These theorems have been created in **my mathbox**, close to the each last required theorem.
- `sqrtcvallem1` expresses an equivalent condition for fifth hypothesis of `eqsqrtd`.
- `reabsifneg` , `reabsifnpos` , `reabsifpos` , and `reabsifnneg` express the absolute value of a real number using if in various ways, only `reabsifneg` is used below. `reabssgn` was added in to give a parallel proof in terms of `df-sgn`.
- `sqrtcvallem2` , `sqrtcvallem3` , `sqrtcvallem4` , `sqrtcvallem5` relate to a square root having a non-negative real argument thus justifying the universality of the formula. The sum requires more steps than the difference.
- `sqrtcval` and `sqrtcval2` are the main result. A YouTube video by Michael Penn is cited as inspiration.`sqrtcval` is a slightly longer statement but conforms to the format expected by `crred` and `crimd`
- `resqrtval` and  `imsqrtval` give just the real and imaginary parts, respectively. They have examples now.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -15240,7 +15240,6 @@ New usage of "chub2i" is discouraged (24 uses).
 New usage of "chunssji" is discouraged (0 uses).
 New usage of "chvar" is discouraged (2 uses).
 New usage of "chvarv" is discouraged (0 uses).
-New usage of "clel5OLD" is discouraged (0 uses).
 New usage of "cleljustALT" is discouraged (0 uses).
 New usage of "cleljustALT2" is discouraged (0 uses).
 New usage of "clelsb3f" is discouraged (0 uses).
@@ -18047,7 +18046,6 @@ New usage of "qlaxr3i" is discouraged (0 uses).
 New usage of "qlaxr4i" is discouraged (0 uses).
 New usage of "qlaxr5i" is discouraged (0 uses).
 New usage of "quoremnn0ALT" is discouraged (0 uses).
-New usage of "r19.12OLD" is discouraged (0 uses).
 New usage of "r19.21biOLD" is discouraged (0 uses).
 New usage of "r19.27vOLD" is discouraged (0 uses).
 New usage of "r19.28vOLD" is discouraged (0 uses).
@@ -18066,8 +18064,6 @@ New usage of "ralbiOLD" is discouraged (0 uses).
 New usage of "ralcom2" is discouraged (0 uses).
 New usage of "ralcom4OLD" is discouraged (0 uses).
 New usage of "raleleqALT" is discouraged (0 uses).
-New usage of "ralnex2OLD" is discouraged (0 uses).
-New usage of "ralnex3OLD" is discouraged (0 uses).
 New usage of "ralrexbidOLD" is discouraged (0 uses).
 New usage of "ralrnmpt" is discouraged (1 uses).
 New usage of "ralxfrALT" is discouraged (0 uses).
@@ -19284,7 +19280,6 @@ Proof modification of "ccatw2s1p1OLD" is discouraged (155 steps).
 Proof modification of "ceqsalgALT" is discouraged (58 steps).
 Proof modification of "ceqsexgvOLD" is discouraged (10 steps).
 Proof modification of "chordthmALT" is discouraged (440 steps).
-Proof modification of "clel5OLD" is discouraged (46 steps).
 Proof modification of "cleljustALT" is discouraged (25 steps).
 Proof modification of "cleljustALT2" is discouraged (25 steps).
 Proof modification of "clelsb3vOLD" is discouraged (54 steps).
@@ -20113,7 +20108,6 @@ Proof modification of "pwundifOLD" is discouraged (49 steps).
 Proof modification of "pwunssOLD" is discouraged (61 steps).
 Proof modification of "qexALT" is discouraged (64 steps).
 Proof modification of "quoremnn0ALT" is discouraged (360 steps).
-Proof modification of "r19.12OLD" is discouraged (61 steps).
 Proof modification of "r19.21biOLD" is discouraged (21 steps).
 Proof modification of "r19.27vOLD" is discouraged (39 steps).
 Proof modification of "r19.28vOLD" is discouraged (38 steps).
@@ -20131,8 +20125,6 @@ Proof modification of "ralanidOLD" is discouraged (39 steps).
 Proof modification of "ralbiOLD" is discouraged (19 steps).
 Proof modification of "ralcom4OLD" is discouraged (41 steps).
 Proof modification of "raleleqALT" is discouraged (26 steps).
-Proof modification of "ralnex2OLD" is discouraged (46 steps).
-Proof modification of "ralnex3OLD" is discouraged (62 steps).
 Proof modification of "ralrexbidOLD" is discouraged (29 steps).
 Proof modification of "ralxfrALT" is discouraged (85 steps).
 Proof modification of "rb-ax1" is discouraged (32 steps).
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11374,30 +11374,15 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
-  <td>reeff1olem</td>
-  <td><i>none</i></td>
-  <td>The set.mm proof depends on ivth (the Intermediate
-  Value Theorem).  Apparently, ~ ivthinc would suffice
-  so this may be provable.</td>
-</tr>
-
-<tr>
-  <td>reeff1o</td>
-  <td><i>none</i></td>
-  <td>the set.mm proof uses reeff1olem</td>
-</tr>
-
-<tr>
   <td>reefiso</td>
   <td><i>none</i></td>
-  <td>the set.mm proof uses reeff1o and the converse of
-  ~ efltim</td>
+  <td>the set.mm proof uses the converse of ~ efltim</td>
 </tr>
 
 <tr>
   <td>efcvx</td>
   <td><i>none</i></td>
-  <td>the set.mm proof uses reeff1o , reefiso , dvres , dvres3 ,
+  <td>the set.mm proof uses reefiso , dvres , dvres3 ,
   iccntr , and dvcvx</td>
 </tr>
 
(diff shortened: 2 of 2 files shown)


tirix:
Looks good! Why not start off of a Mathbox first ?

arpie-steele:
I don't need it to be in the main section, but I thought these theorems were of general interest and consistent with current content:
1. Re, Im, * (conjugate), abs, and sqrt are all introduced in the same section, so being able prove a closed form for the real and imaginary part of the square root seemed like natural development of the topic ; while `eqsqrtd` lets one check a guess of a square root, `sqrtcval` allows one to calculate it (up to an expressing involving radicals of positive reals). 
2. The proofs of lemmas 2/3 expose a potential asymmetry in how the cases abs + Re and abs - Re  have supporting theorems
3. Newton's method to find the non-negative square root function is well-behaved while without `sqrtcval` a naive implementation of complex square root may converge to the wrong root (notably NIST doesn't have advice for the complex square root at https://dlmf.nist.gov/1.9 ) ; (Wikipedia and Handbook of mathematical functions with formulas, graphs, and mathematical tables (1964) both have this formula.)
4. Finally, having convinced myself that it was a strong candidate to eventually move to main, putting it in my sandbox seemed like wasted effort to document the move

Please let me know if you really want it in my mathbox.

icecream17 (approved):

set.mm
```diff
@@ -161144,6 +161193,15 @@ reflection about the origin (under the map ` x |-> -u x ` ).  (Contributed
     ( cc wcel cre cfv cabs recl cr recnd abscl syl cle wbr leabs absrele letrd
     ) ABCZADEZRFEZAFEAGZQRBCSHCQRTIRJKAJQRHCRSLMTRNKAOP $.
 
+  $( The square of the imaginary component of the square root of a complex
+     number is a non-negative real number.  (Contributed by RP,
```
benjub, line 161197: Wrong comment, probably a copy-paste error ?

set.mm
```diff
@@ -162305,6 +162363,13 @@ the complex numbers (sorting by real part first, then by imaginary part as
       BCKZHABLMNBOPCLMNCOPUAUBQDEFGBCRST $.
   $}
 
+  $( The absolute value of the imaginary component of the square root of a
+     complex number is a real number.  (Contributed by RP, 11-May-2024.) $)
```
benjub, line 162367: wrong comment

set.mm
```diff
@@ -162551,6 +162616,144 @@ the complex numbers (sorting by real part first, then by imaginary part as
       CDEUDUEUF $.
   $}
 
+  $( The square of the real component of the square root of a complex number is
+     a non-negative real number.  (Contributed by RP, 11-May-2024.) $)
```
benjub, line 162620: wrong comment.

set.mm
```diff
@@ -160919,6 +160936,38 @@ reflection about the origin (under the map ` x |-> -u x ` ).  (Contributed
...
+    UBZKZUMULUQUJADLFZKUMULMUJURUPUJADUJUCUJUDUEUFAUGUHTUIT $.
+
+  $( Alternate expression for the absolute value of a real number.
+     (Contributed by RP, 11-May-2024.) $)
+  reabsifnneg $p |- ( A e. RR -> ( abs ` A ) = if ( 0 <_ A , A , -u A ) ) $=
+    ( cr wcel cc0 cle wbr cneg cif cabs cfv wa absid eqcomd wn wi 0re clt ltnle
+    wceq ltle sylbird mpan2 imdistani absnid syl ifeqda ) ABCZDAEFZAAGZHAIJZUGU
+    HAUIUJUGUHKUJAALMUGUHNZKZUJUIULUGADEFZKUJUISUGUKUMUGDBCZUKUMOPUGUNKUKADQFUM
+    ADRADTUAUBUCAUDUEMUFM $.
+
```
avekens, line 160970: Do we really need all of these variants (in main)? As far as I can see, only ~reabsifneg is actually used in the following.

benjub (approved):

benjub:
This is not ready for the main section: wrong comments, mislabelings of lemmas, redundant statements (only one of the four versions of the absolute value of reals is necessary; probably the one with the condition `if 0 <= x$`). And link to a youtube video for such a basic formula ? Does this formula have applications ? That's a real question: it may be interesting and I do not know. I'm asking, because many such formulas can be devised.

It would probably be worth introducing the signum function, if it's not already in set.mm.

tirix:
> It would probably be worth introducing the signum function, if it's not already in set.mm.

We do have [df-sgn](https://us.metamath.org/mpeuni/df-sgn.html). It's a good remark that it could be used in some of those formulas.

set.mm
arpie-steele: But Im[Sqrt[z]]^2 = (Abs[z] - Re[z])/2 holds for all z in the complex plane.

set.mm
arpie-steele: Abs[Im[Sqrt[z]]] = Sqrt[(Abs[z] - Re[z])/2] holds for all z in the complex plane and is a real number.

set.mm
arpie-steele: Re[Sqrt[z]]^2 = (Abs[z] + Re[z])/2 holds for all z.

set.mm
arpie-steele: Because I needed to map numbers <0, =0, >0 to -1, 1, and 1 (and not -1, 0, and 1 as with df-sgn), I needed one of these. The others were for future authors who might have made different choices. But it looks like all of this is going to my sandbox.

arpie-steele:
The formula √z = √[½(|z| + Re[z])] ± i √[½(|z| − Re[z])] with the sign chosen based on the sign of Im[z] (but not signum(Im[z])) because we want +1, not 0, for the case when z is a purely real number) appears as 3.7.27 on   [page 17 of Handbook of mathematical functions with formulas, graphs, and mathematical tables (1965)](https://www.google.com/books/edition/Handbook_of_Mathematical_Functions/MtU8uP7XMvoC?hl=en&gbpv=1&pg=PA17&printsec=frontcover) and probably many other places. The video was my inspiration to formalize this. The lemmas with 2 and 3 in their labels are described with the (equivalent!) expressions that motivate a need for them, as is made clear by the last two theorems. The labels could be improved with a link to those later theorems.
I will work on returning these theorems to my sandbox.

set.mm
benjub: I'm not saying the sentence is false.  I'm saying that
> The square of the imaginary component of the square root of a complex number is a non-negative real number.

cannot be the comment for the statement $A \in \mathbb{C} \to 0 \leq \frac{|A|- Re(A)}{2}$.

To exagerate things a bit, it's as if I commented the statement $\sin \pi = 0$ with the sentence "The limit when $x \to +\infty$ of $\sin(2\arctan x)$ is 0."

After looking at the statement, maybe the comment "Lemma for ~ sqrtcval ." is enough, given the elementary character of the statement (that is: no need IMO to paraphrase it).

avekens:
> I will work on returning these theorems to my sandbox.

I think this is a good idea to avoid any further discussion, and it is good practice to put such theorems in a mathbox first. As soon as these theorems are used by another one, they can be moved (maybe partly) to main.

benjub:
> The formula √z = √[½(|z| + Re[z])] ± i √[½(|z| − Re[z])] with the sign chosen based on the sign of Im[z] (but not signum(Im[z])) because we want +1, not 0, for the case when z is a purely real number) appears as 3.7.27 on [page 17 of Handbook of mathematical functions with formulas, graphs, and mathematical tables (1965)](https://www.google.com/books/edition/Handbook_of_Mathematical_Functions/MtU8uP7XMvoC?hl=en&gbpv=1&pg=PA17&printsec=frontcover) and probably many other places. The video was my inspiration to formalize this. The lemmas with 2 and 3 in their labels are described with the (equivalent!) expressions that motivate a need for them, as is made clear by the last two theorems. The labels could be improved with a link to those later theorems. I will work on returning these theorems to my sandbox.

If it's in that book, then that answers my question above and it can probably be put to Main in a later PR (with that reference).

Indeed, the 2-step process of first mathbox, and then, give other viewers some time to see if/what part can go to main, or should be modified or relabeled or recommented... is a sound process.

set.mm
benjub: Further comment: what you just wrote above is your statement sqrtcval (and its Re and Im corollaries), and now I agree that these statements have some interest.

wlammen (approved):

set.mm
```diff
@@ -651997,6 +651997,281 @@ of all sets ( ~ inex1g ).
...
+This is based on the observation that the real and imaginary parts of a complex
+number can be calculated from the number's absolute and real part and the sign
+of its imaginary part.
+
+<HTML>
+Formalization of the formula in ~ sqrtcval was motivated by a short
+<A HREF="https://www.youtube.com/watch?v=QAhWTQ4Ravw">Michael Penn video</A>.
+</HTML>
+$)
```
avekens, line 652013: Please respect the conventions for "Sectioning and section headers" (see https://us.metamath.org/mpeuni/conventions-comments.html) for the body of the section header comment: [blank line; header comment indented with two spaces; blank line;].

arpie-steele:
> > It would probably be worth introducing the signum function, if it's not already in set.mm.
> 
> We do have [df-sgn](https://us.metamath.org/mpeuni/df-sgn.html). It's a good remark that it could be used in some of those formulas.

`reabssgn` was created folloowing your suggestion.

set.mm
arpie-steele: Now that they are in my sandbox, a fifth theorem based on `df-abs` was added.

tirix (approved):
Thanks for the move to the mathbox, and the new theorem combining `abs` and `sgn`!

set.mm
```diff
@@ -652048,6 +652049,301 @@ of all sets ( ~ inex1g ).
...
+     (Contributed by RP, 11-May-2024.) $)
+  reabsifpos $p |- ( A e. RR -> ( abs ` A ) = if ( 0 < A , A , -u A ) ) $=
+    ( cr wcel cc0 clt wbr cneg cif cabs cfv wa cle wceq 0re ltle mpan imdistani
+    wi absid eqcomd syl wn id 0red lenltd pm5.32i absnid sylbir ifeqda ) ABCZDA
+    EFZAAGZHAIJZUJUKAULUMUJUKKZUMAUNUJDALFZKUMAMUJUKUODBCUJUKUORNDAOPQASUATUJUK
+    UBZKZUMULUQUJADLFZKUMULMUJURUPUJADUJUCUJUDUEUFAUGUHTUIT $.
+
+  $( Alternate expression for the absolute value of a real number.
+     (Contributed by RP, 11-May-2024.) $)
+  reabsifnneg $p |- ( A e. RR -> ( abs ` A ) = if ( 0 <_ A , A , -u A ) ) $=
```
benjub, line 652119: To me, this the one version out of the four that could be moved in a future MR to Main.  Indeed, it seems more natural to leave the argument unchanged when the given condition is true (this rules out the versions with condition A<0 and A<=0), and also to leave the argument unchanged as often as possible (so I prefer A>=0 to A>0).  Equivalently, the largest set on which the absolute value is the identity is the set of nonnegative reals.

benjub: Edit: can you prove it for extended reals (with $\pm \infty$) ?

set.mm
```diff
@@ -652048,6 +652049,301 @@ of all sets ( ~ inex1g ).
...
+    ( cr wcel cc0 cle wbr cneg cif cabs cfv wa absid eqcomd wn wi 0re clt ltnle
+    wceq ltle sylbird mpan2 imdistani absnid syl ifeqda ) ABCZDAEFZAAGZHAIJZUGU
+    HAUIUJUGUHKUJAALMUGUHNZKZUJUIULUGADEFZKUJUISUGUKUMUGDBCZUKUMOPUGUNKUKADQFUM
+    ADRADTUAUBUCAUDUEMUFM $.
+
+  $( Alternate expression for the absolute value of a real number.
+     (Contributed by RP, 22-May-2024.) $)
+  reabssgn $p |- ( A e. RR -> ( abs ` A ) = ( ( sgn ` A ) x. A ) ) $=
```
benjub, line 652127: Nice. Can you prove it for extended reals ?  Isn't there a shorter proof using reabsifnonneg ?

benjub (approved):

set.mm
```diff
@@ -652048,6 +652049,301 @@ of all sets ( ~ inex1g ).
...
+     by ~ crrei and ~ crimi .
+     <HTML>
+     This formula can be found in section 3.7.27 of <I>Handbook
+     of Mathematical Functions</I>, ed. M. Abramowitz and I. A.  Stegun
+     (1965, Dover Press).
+     </HTML>
+     (Contributed by RP, 18-May-2024.) $)
+  sqrtcval $p |- ( A e. CC -> ( sqrt ` A ) =
+                   ( ( sqrt ` ( ( ( abs ` A ) + ( Re ` A ) ) / 2 ) ) + ( _i x.
+                     ( if ( ( Im ` A ) < 0 , -u 1 , 1 ) x.
```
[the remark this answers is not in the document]
arpie-steele: But you can't use sgn(Im(A)) because sgn(Im(-1)) = 0 which would make the sqrt(-1) = 0

set.mm
arpie-steele: df-abs and df-mul only want to work with complex numbers, so the infinities of the extended reals are out-of-bounds.

set.mm
arpie-steele: This has been addressed throughout my sandbox. [47b6591](https://github.com/metamath/set.mm/pull/3993/commits/47b65911b0e91b3ab8d8e96b1dc0cbecf37547d8)

set.mm
benjub: Oh, I see. So it's a longer term project... It would be nice to extend these operations, using for instance https://us.metamath.org/mpeuni/df-bj-mulc.html

set.mm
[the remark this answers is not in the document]
benjub: Ah, of course !  Sorry, I didn't think, but of course, there has to be a discontinuity along a ray !

set.mm
arpie-steele: I did find a shorter proof with `reabsifneg` because it matches the condition in `df-abs`

benjub: You probably mean "the condition in `df-sgn`" ? Well, I think that the definition of signum is the one that should be changed. That's for later, obviously (but if it's changed while still not much used, it will be easier).

wlammen (approved):

wlammen:
The last two commits are shortenings, affecting the internals of a proof only.

david-a-wheeler:
As the perpetrator of [df-sgn](https://us.metamath.org/mpeuni/df-sgn.html) I always encourage its (appropriate) use :-). I do think that at least some of these should eventually move from the mathbox into the "main" area, but we do often start with mathboxes first.

PR #3997, opened 2024-05-19
N-ary functions

avekens:
Definition ~df-naryf of n-ary (endo)functions `NaryF` and related theorems. Additional theorems for nullary and unary functions.

To get a definition which supports arbitrary numbers of arguments,  mappings from the half-open range of nonnegative integers to the domain of the arguments is used (to be understood as indexed arguments).  Alternatively, elements of Cartesian exponentiations ` ( U ^^ N ) ` (see ~ df-finxp ) could have been used to represent multiple arguments.  However, this concept is not fully developed yet (it is within a mathbox), and it is currently based on ordinal numbers, e.g., ` ( U ^^ 2o ) `, instead of integers, e.g., ` ( U ^^ 2 ) `, which is not very practical.

This is a preparation for the definition of primitive recursive functions, which is based on n-ary functions (from nonnegative integers into nonnegative integers).


set.mm
```diff
@@ -757432,6 +757435,319 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  For example, multiplication of integers is a function of two variables, or
+  _bivariate function_, whose domain is the set of all ordered pairs (2-tuples)
+  of integers, and whose codomain is the set of integers.  The same is true for
+  every binary operation.  Commonly, an n-tuple is denoted enclosed between
+  parentheses, such as in ( 1 , 2 , ... , n ).  When using functional notation,
+  one usually omits the parentheses surrounding tuples, writing
+  f ( x_1 , ... , x_n ) instead of f ( ( x_1 , ... , x_n ) ).  Given n sets
+  X_1 , ... , X_n , the set of all n-tuples ( x_1 , ... , x_n ) such that
+  x_1 is element of X_1 , ... , x_n is element of X_n is called the Cartesian
+  product of X_1 , ... , X_n , and denoted X_1 X ... X X_n .
```
tirix, line 757463: Another interesting point is that because of the way things are defined in set.mm, a `2 N-ary` function is not exactly the same thing as an operation. The domain of the former (this function) is the set of words of length 2 (functions from `( 0 ..^ 2 )` to `X` ), while the domain of the latter is set of pairs of elements of `X`, i.e. the cartesian product `( X X. X )`. The same is true for univariate functions. 

It can be noted though that those are equinumerous, see `~naryenef`.

set.mm
```diff
@@ -757432,6 +757435,319 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  Therefore, a multivariate function is a function that has a Cartesian product
+  or a proper subset of a Cartesian product as a domain: ` f : U --> Y ` where
+  where the domain ` U ` has the form
+  ` U C_ ( ( ... ( ( X `` 1 ) X. ( X `` 2 ) ) X. ... ) X. ( X `` n ) ) `."
+
+  In the following, n-ary functions are defined as mappings (see ~ df-map )
+  from a finite sequence of arguments, which themselves are defined as
+  mappings from the half-open range of nonnegative integers to the domain of
+  each argument.  Furthermore, the definition is restricted to endofunctions,
```
tirix, line 757472: I'm afraid that one not-so-distant day we might need a N-ary function with different range and domains (so, not endofunctions).

Actually the most generic definition would have each X different, as in [multivariate functions](https://en.wikipedia.org/wiki/Function_(mathematics)#Multivariate_functions).
We do have a definition of the [cartesian product of a set of sets](https://us.metamath.org/mpeuni/df-ixp.html) which could be of use.

Of course for this definition you have the primitive recursive functions in mind, so endofunctions are sufficient, and this development is already complex.

set.mm
```diff
@@ -757432,6 +757435,319 @@ last digit of the integer part (for ` b = ` 10 the first digit before
...
+  Instead of using indexed arguments (represented by a mapping as described
+  above), elements of Cartesian exponentiations ` ( U ^^ N ) ` (see
+  ~ df-finxp ) could have been used to represent multiple arguments.  However,
+  this concept is not fully developed yet (it is within a mathbox), and it is
+  currently based on ordinal numbers, e.g., ` ( U ^^ 2o ) `, instead of
+  integers, e.g., ` ( U ^^ 2 ) `, which is not very practical.
+
+$)
+
+  $c NaryF $.  $( Letter I for the n-ary functions. $)
```
tirix, line 757499: I don't understand wher "Letter I" comes from. Copy-paste or typo?

avekens: Yes, this was a copy&paste error - fixed.

set.mm
avekens: I have chosen this approach to keep the representation simple (only two parameters `X`and `N` in `( X NaryF N )`. If the range had to be different, then we must add a third parameter, e.g. `( <. X , Y >. NaryF N )`.  If the domains for the arguments had to be arbitrary, the notation would become even more complex...

tirix: Got it. This already has a wide range of applications.

tirix (approved):

set.mm
avekens: I have added ~ naryfvalixp which shows that n-ary (endo)functions can also be expressed with the notation of infinite Cartesian products. Nevertheless, I think using the notation of infinite Cartesian products to allow different domains for the arguments is not so easy, maybe as `NaryF = ( x e. _V , n e. NN0 |-> ( x ^m X_ y e. ( 0 ..^ n ) x) )`, because ` x ` must not depend on ` y ` for the soundness of the definition...

set.mm
avekens: In a next step, I will investigate 2-ary (i.e, binary) functions/Operations in more detail.

savask:
> This is a preparation for the definition of primitive recursive functions, which is based on n-ary functions (from nonnegative integers into nonnegative integers).

This is very exciting news! Given your recent pull request, you must be planning to show that Ackermann's function isn't primitive recursive. Of course it would be great to have a more general theory of _recursive_ functions, recursively enumerable sets, Turing degrees etc. In my opinion set.mm is already ripe for that.

I have two small comments. First, in the last section "Primitive recursive functions" there are phrases like
>   The basic primitive recursive functions are given by ... axioms:
  More complex primitive recursive functions can be obtained by applying the operations given by ... axioms:

What the ellipsis ... is supposed to mean?

Second comment is more general. I noticed that you often quote wikipedia in set.mm sections. In my opinion, this is not always an ideal choice, since wikipedia is generally a bad source for mathematics and more importantly, it is subject to change (for example, links to specific parts of a wikipedia article can stop working if someone edits the original article). Maybe it is best if we try to reference published books in set.mm sections - for computability theory I suggest Soare's "Recursively enumerable sets and degrees" and references therein.

avekens:
> > This is a preparation for the definition of primitive recursive functions, which is based on n-ary functions (from nonnegative integers into nonnegative integers).
> 
> This is very exciting news! Given your recent pull request, you must be planning to show that Ackermann's function isn't primitive recursive. 

That is exactly my intention.

> Of course it would be great to have a more general theory of _recursive_ functions, recursively enumerable sets, Turing degrees etc. In my opinion set.mm is already ripe for that.

I'll see if I can go further in this direction, but others are invited to extend set.mm accordingly.

> I have two small comments. First, in the last section "Primitive recursive functions" there are phrases like
> 
> > The basic primitive recursive functions are given by ... axioms:
> > More complex primitive recursive functions can be obtained by applying the operations given by ... axioms:
> 
> What the ellipsis ... is supposed to mean?

The ellipses ... are omissons of the text from Wikipedia, which I regard as not helpful.
 
> Second comment is more general. I noticed that you often quote wikipedia in set.mm sections. In my opinion, this is not always an ideal choice, since wikipedia is generally a bad source for mathematics and more importantly, it is subject to change (for example, links to specific parts of a wikipedia article can stop working if someone edits the original article). Maybe it is best if we try to reference published books in set.mm sections - for computability theory I suggest Soare's "Recursively enumerable sets and degrees" and references therein.

Maybe it is not scientific enough, but citing Wikipedia has several advantages: it is open to the public, it is immediately available by clicking the link, and it is often more general than contents in correspondig books (depending on the author). I always provide the date when I took the copy, so the citation is always transparent looking into the history of the Wikipedia article. Since I have not enough time (and opportunity - no direct access to a adequate and fully equipped libraries) for a deeper research in literature, I would like to stick with Wikipedia. Others are welcome to add additial references.

icecream17 (approved):
As a future note/add-on to a resolved thread,  other more general cases are represented well enough by df-map and various structures (df-pws, df-frlm, and incidentally df-word)

(Note also this is definitely not redundant because of the simplicity in notation and expanded tools for this case)

PR #3998, opened 2024-05-20
Add apdiff to iset.mm mathbox

jkingdon:
This is a way of characterizing irrationals as exactly those reals that are a different distance from every rational:

```
    apdiff $p |- ( A e. RR -> ( A. q e. QQ A =//= q <-> A. q e. QQ A. r e. QQ
        ( q =/= r -> ( abs ` ( A - q ) ) =//= ( abs ` ( A - r ) ) ) ) ) $=
```

PR #4002, opened 2024-05-21
Add irrdiff to set.mm

jkingdon:
This is similar to apdiff from iset.mm but the theorem defines irrational differently and uses not equal in place of apart.

Includes lemma irrdifflemf .

See discussion at https://github.com/metamath/set.mm/pull/3998#pullrequestreview-2065479763


tirix (approved):

icecream17 (approved):

jkingdon:
There's an informalization of this proof and a few more thoughts at https://sfba.social/@soaproot/112477489483787894 and the rest of that thread.

avekens (approved):

PR #4003, opened 2024-05-21
mathbox: bj-subst

benjub:


set.mm
```diff
@@ -536981,6 +536982,29 @@ standard proof of the implication in modal logic (B5 ` => ` 4).  Its dual
   bj-wnfenf $p |- ( ( E. x ph -> A. x ps ) -> A. x ( E. x ph -> ps ) ) $=
     ( wex wal wi bj-wnf1 bj-19.21bit sylg ) ACDZBCEFZKJBFCABCGJBCHI $.
 
+  $( Equivalent form of the axiom of substitution ~ bj-ax12 .  Although both
+     sides need a DV condition on ` x , t ` (or as in ~ bj-ax12v3 on
+     ` t , ph ` ) to hold, their equivalence holds without DV conditions.  The
+     forward implication is proved in modal (K4) while the reverse implication
```
tirix, line 536988: So `K4` and `T5` are a numbering of the modal logic axioms here?

set.mm
```diff
@@ -536981,6 +536982,29 @@ standard proof of the implication in modal logic (B5 ` => ` 4).  Its dual
...
+     reverse implication holds by ~ equs4 (or ~ equs4v if a DV condition is
+     added on ` x , t ` as in ~ bj-ax12 ).
+
+     The LHS can be read as saying that if there exists a setvar equal to a
+     given term witnessing ` ph ` , then all setvars equal to that term also
+     witness ` ph ` .  An equivalent suggestive form for the LHS is
+     ` -. ( E. x ( x = t /\ ph ) /\ E. x ( x = t /\ -. ph ) ) ` , which
+     expresses that there can be no two variables both equal to a given term,
+     one witnessing ` ph ` and the other witnessing ` -. ph ` .  (Contributed
+     by BJ, 21-May-2024.)  (Proof modification is discouraged. $)
```
tirix, line 537000: We should probably ensure that ~bj-ax12 is not used?

set.mm
```diff
@@ -536981,6 +536982,29 @@ standard proof of the implication in modal logic (B5 ` => ` 4).  Its dual
...
     ( wex wal wi bj-wnf1 bj-19.21bit sylg ) ACDZBCEFZKJBFCABCGJBCHI $.
 
+  $( Equivalent form of the axiom of substitution ~ bj-ax12 .  Although both
+     sides need a DV condition on ` x , t ` (or as in ~ bj-ax12v3 on
+     ` t , ph ` ) to hold, their equivalence holds without DV conditions.  The
+     forward implication is proved in modal (K4) while the reverse implication
+     is proved in modal (T5).  The LHS has the advantage of not involving
+     nested quantifiers on the same variable.  Note that in the LHS, the
+     reverse implication holds by ~ equs4 (or ~ equs4v if a DV condition is
+     added on ` x , t ` as in ~ bj-ax12 ).
```
tirix, line 536992: I know what DV and LHS mean, but I wonder if we have a glossary?

set.mm
benjub: Yes, this is standard terminology in modal logic. Axiom K is ax-4, Axiom T is spec, Axiom 4 is $\forall \to \forall\forall$, and Axiom 5 is $\exists \to \forall \exists$.  I updated some comments in set.mm in recent years, to indicate these correspondences (search "modal" in comments or labels of the main part of set.mm).

The modal logics (K4) and (T5) are the logics using these axioms on top of propositional calculus and the generalization rule.

This is also summarized in the section "Modal logic" in my mathbox.

set.mm
benjub: Statements in a mathbox are by default not used outside of that mathbox (one has to add the option `/INCLUDE_MATHBOXES` to `MINIMIZE_WITH <label>` in order to use them). And in my mathbox, I'm happy with using it later.  Can I keep it that way ?

set.mm
benjub: Maybe we can add a glossary indeed.  If we do, we should strive to keep it very brief, short, terse, with very few entries, only for terms that are not standard outside Metamath.  And especially important: we should not take the glossary as an excuse to use more abbreviations in the text.  On the one hand, I think DV, LHS, RHS, FOL, ZF(C) could be used as is, but on the other hand, I fear that this could go out of hands (e.g.: AC, DC, BIP, PP), so I may spell out DV and LHS if you think it's better.

set.mm
tirix: Yes, fine for me!

set.mm
tirix: Maybe other have different opinions, I'd like to hear them too.

I'm just thinking that we should keep the entry barrier as low as we can, and that came to my mind reading that comment. This is certainly not blocking, and there are actually several other places where DV and LHS are used as acronyms.

tirix (approved):

set.mm
wlammen: In [df-clab](https://us.metamath.org/mpeuni/df-clab.html) the term LHS appears for the first time in set.mm, and it is there explained as "definiendum".  In the rest of the comment it is used a second time.
1.  Abbreviations make sense if a complicated phrase appears multiple times in a text.  But with as few as 2 times, the abbreviated term must be of extraordinary length to justify such a measure IMO (in my opinion).
2. The phrase LHS stands for in this instance is explained as a "definiendum".  I wonder whether many people new to math really understand it better.  At least it is a term that can be looked up in a dictionary.
3. Being at "looking up": A Google search does not point to it within the first page of search results.  Instead you see companies named "LHS" and other stuff.  You better specify the search to "LHS math", then you are pointed to a Wikipedia page [Sides_of_an_equation](https://en.wikipedia.org/wiki/Sides_of_an_equation) with the explanation "In mathematics, LHS is informal shorthand for the left-hand side of an equation."
4.   An astute reader will notice that this is different from definiendum, since it applies to equations, and the left hand side can be any expression, not just something being defined.  The next appearance in [abid1](https://us.metamath.org/mpeuni/abid1.html) supports this view, because here the LHS is not a definiendum in the slightest way.
5. Norm has used LHS a few times, and explained it in [conventions](https://us.metamath.org/mpeuni/conventions.html) again, this time more to the Wikipedia explanation.
6. In main @david-a-wheeler was so far the only user of LHS multiple times.

tirix: Thanks for this extensive study!
Interestingly, if I google "LHS meaning", I immediately get the right answer. 
Anyway this is certainly not blocking this contribution.

tirix: Noting that this is addressed by #4005

Issue #3990, opened 2024-05-16
Feedback from a new user, improvement suggestions.


metakunt:
I don't have a google account and I don't know if it is possible to use the mailing list without one. If it is, please let me know the steps so that I can post it on the mailing list. Otherwise I would be happy if one of you could open a issue on the metamath mailing list on my behalf.  I hope it is as simple as writing an E-mail to metamath@googlegroups.com.

Hey guys, I have been working for a few weeks on the project and as a new user I wish to share some feedback I have. I want to start this discussion to overall improve the state of metamath and make sure that it is easy and straightforward to use.

Many of you are either maintainers and/or are on the project for a long time.
Also I will only be listing things that can be improved from my perspective, things that are positive/working are not listed. Those things include the helpful community and the fact that metamath is pretty fun to work on.

Here are the things that I had issues with and that I wish to improve in some way, listed in order of severity (most severe issues are listed first.)

-Lack of theorems in deduction form. Oh my goodness, I can't explain how much that one annoyed me. Once I have learned how deduction proofs work I fell in love with, it is simple, fast and has a great developer experience. The reason is once you have a theorem in deduction form the unifier will do most work for you, completing substeps automatically which I'd have to painstakingly find myself. It is just great and speeds up the development by one or two orders of magnitude per step. This adds up, my generous estimate would be that using theorems in deduction form is at least twice as fast as theorems in closed form, just by the virtue of autocompleting.
But this is not all. The best thing about deduction proof (similar to an inference proof) is that it absorbs and scopes everything in context. That means that antecedents are all small, with some exceptions being conjunctions that eliminate themself eventually (classical case elimination). But this issue maps to closed form proofs so it is irrelevant.
Point being it is easier for me to recreate a theorem in deduction form to use it, as I did in my mathbox, because it will save me time after just a few usages. Converting a theorem into deduction form with 5 or more conjuncts will amortize after 1 or 2 usages. Some lemmas which have lots of conditions are gcd lemmas, deduction forms are in my mathbox. I'd like to move them to main as they are way easier to use.

I have multiple suggestions here, my most preferred would be to rewrite simple results on complex numbers from closed form to deduction form and deprecate those closed form results and eventually delete them. (Section 5.2 and Section 5.3) I have the opinion that deduction form is superior to closed form in almost every other way and the benefits for moving in that direction would make proof development much faster. Since this is something I might be wrong with I would like to know if there are any downsides that would make this a bad idea.

[...]

Issue #3908, opened 2024-04-05
Newbie here, I created a proof with mmj2. How do I add the proof to the database


metakunt:
So the mmp file format looks different than the mm file format. And for the love of god I couldn't find any utility to convert it.
The mmj2 repo help says that there is no functionality to add a proof automatically.

In the attachment there is a file of a subproof I am hoping is helpful. What I want to show is that a function is a permutation.
[proofsandsupproofs.zip](https://github.com/metamath/set.mm/files/14890283/proofsandsupproofs.zip)

In the zip archive there are three files. One is a subproof that mmj2 already says is correct. (File: noideahowtonameit.mmp)
The second and third one is the actual proof .
Here I am still debating if I can generalize to include the case ( 1 ... 0 ) --> ( 1 ... 0 ) which is a map from the empty to the empty set.
This is a map, which should also be bijective, so it should work.
I want to show that the function defined by F is a permutation. The domain is ( 1 ... N ) where N is in NN0 (here I am debating if NN0 is better than NN, it might make more sense to define it that way, but the only use case will be N in NN). Let the domain be A.
I want to show that F: A->A  is a function from A to A.
Then I planned to use the fact that A is finite together with an injection/surjection (I still don't know which one is easier) -> F bijective.

Let i \in \{1,...,m\}
The function F is defined by two if statements. In Latexlike it would be something like
\{1,...,m\} -> \{1,...,m\}
j |-> m-i+j+\chi(m-i+j,i) if 1 \leq j \leq i-1
j |-> j if i=j
j |-> j-i+\chi(j-i,i) i+1 \leq j \leq m

Eventually I want to show that this is a permutation. For now, I'd be more than happy to show that it is a function with dom=cod=\{1,...,m\}

This is why I developed this new lemma. My hope is that I can show 
(x e. ( 1 ... M ) -> bla)
by showing 
(x e. (1 ... M ) and x = I ) -> bla
(x e. (1 ... M ) and x \neq I  and  x < l) -> bla
(x e. (1 ... M ) and x \neq I  and not ( x < l )) -> bla

where bla is the fact that the evaluation of F(x) is in (1...m)
Then I can use my theorem to eliminate all, such that the unification works. But I didn't find a good way. The only thing I've found is a syllogism rule, which is quite helpful, but I didn't find how I can move antecedents inside and outside of the if statements.

icecream17:
This doesn't solve the file conversion problem, but the following theorems are relevant to move antecedents:

ex
imp
[...]

PR #4007, opened 2024-05-23
Permutation results

metakunt:


set.mm
```diff
@@ -138351,6 +138364,14 @@ the existence of its supremum (see ~ suprcl ).  (Contributed by Paul
   elfzelz $p |- ( K e. ( M ... N ) -> K e. ZZ ) $=
     ( cfz co wcel cuz cfv cz elfzuz eluzelz syl ) ABCDEFABGHFAIFABCJBAKL $.
 
+  ${
+    elfzelzd.1 $e |- ( ph -> K e. ( M ... N ) ) $.
+    $( A member of a finite set of sequential integer is an integer.
```
benjub, line 138369: integers

metakunt: Ah I just copy pasted, then that needs fixing too https://us.metamath.org/mpeuni/elfzelz.html

set.mm
```diff
@@ -634631,6 +634631,42 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+    $d I x $.  $d M x $.  $d X x $.  $d ph x $.
+    metakunt3.1 $e |- ( ph -> M e. NN ) $.
+    metakunt3.2 $e |- ( ph -> I e. NN ) $.
+    metakunt3.3 $e |- ( ph -> I <_ M ) $.
+    metakunt3.4 $e |- A = ( x e. ( 1 ... M ) |-> if (
+       x = I , M , if ( x < I , x , ( x - 1 ) ) ) ) $.
+    metakunt3.5 $e |- ( ph -> X e. ( 1 ... M ) ) $.
+    $( Value of A. (Contributed by metakunt, 23-May-2024.) $)
+    metakunt3 $p |- ( ph -> ( A ` X ) = if (
+       X = I , M , if ( X < I , X , ( X - 1 ) ) ) ) $=
```
benjub, line 634644: Depending on your goals, it may be worth parameterizing this A by I:
```
A = ( i e. ( 1 ... M ) |-> ( x e. ( 1 ... M ) |-> if ( .... ) )
```
and similarly for the A defined in the next statement.

benjub (approved):

set.mm
metakunt: Thank you for your feedback.

Actually, that is an excellent point. I came a long way since I started with metamath a few weeks ago, see https://github.com/metamath/set.mm/issues/3908 what I want to prove. Back then the theorem I want to prove now was way too difficult for me. But now once I have proven the lcm inequality lemma I am confident to once again tackle it.

A is actually parametrised by I, but what I first want to show is that for any I the function is a permutation. I don't see the benefits in parametrising it right now, as the theorems that I need in this stage would be more complex to prove. Are there any drawbacks in keeping it as is or can I later parametrise it if I choose so? This is a sincere question since I am not that experienced in metamath yet and I want to avoid potential pitfalls.

I have also thought about defining it via operation as you could write something like ( I A X ) and then evaluate it.
One theorem would then be ( X A X ) = X for all X in 1 ... M, which is later an important invariant that holds. But so far this just seems syntactic sugar which I am not yet sure if I'll need. I also think that It should be possible to do when needed without much trouble.

My goal is also mentioned in https://github.com/metamath/set.mm/issues/3990 I want to use https://us.metamath.org/mpeuni/2fvidf1od.html and for that both functions defined need to be maps from 1,...m to itself.

benjub: You can probably add the functionalized version later, and then a proof for the new `` ( YourFunction ` i ) `` will be able to reuse the result that you just proved in this PR since it satisfies its hypotesis. And defining the parameterized function will certainly not spare you that work anyway. So maybe it is a bit premature indeed.

And actually, your function could have two parameters, e.g., $F_n^i \in \mathfrak{S}_n$ with $F(i) = n$ and the other values as you wrote. This is how I view it from a mathematical point of view without looking at formalization matters. Maybe it is overkill for the formalization work, and only makes it harder for no immediate benefit. The benefit really depends on what you will do with it later.  If $F$ were really a fundamental object, I would push towards the parameterized version, but given that it's not, maybe it's not worth the hassle.

Indeed, for binary functions, it may be better to use the notation `( x e. A , y e. B |-> ... )` and then use the operation notation `( X YourFunction Y )` like you wrote. It depends on the manipulation lemmas available in set.mm, and I don't know exactly the current situation.

avekens: > Indeed, for binary functions, it may be better to use the notation `( x e. A , y e. B |-> ... )` and then use the operation notation `( X YourFunction Y )` like you wrote. It depends on the manipulation lemmas available in set.mm, and I don't know exactly the current situation.

Yes, this would be the more preferable way.

avekens (approved):

set.mm
metakunt: Thanks for your feedback.

As of now the lemmas are transitionary. What I want is to show existence of a map
```math
f: \{1,\dots,m\} \to \ ( C_{2} \times S_{n})
```
```math
i \mapsto (c_i,d_i)
```
where $d_i(i)=i$ and some other conditions. This lemma constructs this map and shows that this condition and other conditions also hold.
You can think of $d_i = A^{-1}_i\circ B_i \circ A_i$ and so far the I want to show that those maps are inverse of each other and thus are both permutations. Next step is to construct B with the same procedure. Then I show that $d_i$ is a permutation and thus I have constructed $d_i$. Next I show $d_i(i)=i$ and finally I construct f.
But so far I am not there yet. There are still also some open points as I will need a Homomorphism from $\mathbb{Z} \to C_2$ where $C_2$ is identified by $\mathbb{Z}/{2\mathbb{Z}}$

I don't know what lemmas we have for finite fields, as one map is $e: S_n \to C_2$, plugging the constructed $f$ into e should yield $e(f)\equiv 0$

PR #4010, opened 2024-05-24
Intuitionize logarithms from explog to loglt1b

jkingdon:
For the most part, real logarithm functions are proved and complex logarithm functions are noted in the missing theorems list.

One big part is a proof of https://us.metamath.org/mpeuni/eflt.html via continuity (I don't know if there is an easier proof but this one works). Some consequences of `eflt` are included as well.

`logfac` seems provable but a bit involved, so it isn't included here.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11587,6 +11587,16 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>logfac</td>
+  <td><i>none</i></td>
+  <td>Should be provable.  Because of the way seqhomo is used in the
+  set.mm proof, it may be necessary to special-case ` N = 1 ` and
+  start the main proof with two, or prove an analogue of ~ seq3homo
+  which allows using ` RR+ ` on one side of the equation and ` RR `
+  on the other.</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
   <td>~ apdiff</td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -11611,6 +11611,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>logdivlt</td>
+  <td><i>none</i></td>
+  <td>the set.mm proof uses real number trichotomy</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
   <td>~ apdiff</td>
 </tr>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -123757,14 +123757,6 @@ seq m ( x. , F ) ~~> x ) ) \/
...
       VGVNTAVGUNUOVFVPBGVFABVRVFBVLTUPUQURUSUT $.
 
-    $( The exponential function on the reals is nondecreasing.  (Contributed by
-       Mario Carneiro, 11-Mar-2014.)  (Revised by Jim Kingdon, 20-Dec-2022.) $)
-    efler $p |- ( ( A e. RR /\ B e. RR )
-        -> ( ( exp ` A ) <_ ( exp ` B ) -> A <_ B ) ) $=
-      ( cr wcel wa ce cfv clt wbr wn wi efltim ancoms con3d simpl reefcld simpr
-      cle lenltd lenlt 3imtr4d ) ACDZBCDZEZBFGZAFGZHIZJBAHIZJUFUERIABRIUDUHUGUC
-      UBUHUGKBALMNUDUFUEUDAUBUCOPUDBUBUCQPSABTUA $.
-
```
avekens, line 123767: Why is this proof so  much shorter than the proof of ~eflt? It had almost the same size as the new proof (for ~efle) which is based on ~eflt.
Maybe this proof should be kept as ALT version.

mmil.raw.html
```diff
@@ -9221,18 +9221,6 @@
...
-  <TD>~ efltim</TD>
-  <TD>The set.mm proof of the converse relies on ltord1</TD>
-</TR>
-
-<TR>
-  <TD>efle</TD>
-  <TD>~ efler</TD>
-  <TD>The set.mm proof of the converse relies on eflt</TD>
-</TR>
-
```
avekens, line 9235: These theorems were already proven before? Then this is only a clean up of this list?

avekens (approved):

iset.mm
jkingdon: `efle` is a biconditional, but `efler` is only one direction (hence the `r` in the name).

In both cases, `efltim` is the key to one direction (this pull doesn't change that). But this pull request proves the other direction, not via `ltord1` as set.mm does, but via continuity.

mmil.raw.html
jkingdon: No, `eflt`, `efle`, and `reefiso` were not proved before in iset.mm.  They are added by this pull request.

PR #4012, opened 2024-05-24
Show one sided inverse for metakunt main theorem

metakunt:


set.mm
```diff
@@ -633467,6 +633467,21 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
 
+  ${
+    casetrich.1 $e |- ( ph -> A e. RR ) $.
+    casetrich.2 $e |- ( ph -> B e. RR ) $.
+    casetrich.3 $e |- ( ( ph /\ A < B ) -> ch ) $.
+    casetrich.4 $e |- ( ( ph /\ A = B ) -> ch ) $.
+    casetrich.5 $e |- ( ( ph /\ B < A ) -> ch ) $.
+    $( Case elimination based on trichotomy of real numbers.  (Contributed by
+       metakunt, 24-May-2024.) $)
+    casetrich $p |- ( ph -> ch ) $=
```
wlammen, line 633478: 1. Looks a bit like ltlecasei.  I wonder whether a shorter proof based on it exists.
2. Suggest to rename ch to ps, to be more along the lines of our usual patterns.

set.mm
```diff
@@ -634895,6 +634910,264 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+    $d ph x $.  $d ph y $.
+    metakunt5.1 $e |- ( ph -> M e. NN ) $.
+    metakunt5.2 $e |- ( ph -> I e. NN ) $.
+    metakunt5.3 $e |- ( ph -> I <_ M ) $.
+    metakunt5.4 $e |- A = ( x e. ( 1 ... M ) |-> if (
+       x = I , M , if ( x < I , x , ( x - 1 ) ) ) ) $.
+    metakunt5.5 $e |- C = ( y e. ( 1 ... M ) |-> if (
+       y = M , I , if ( y < I , y , ( y + 1 ) ) ) ) $.
+    metakunt5.6 $e |- ( ph -> X e. ( 1 ... M ) ) $.
+    $( C and A are inverse on one side.  (Contributed by metakunt,
```
wlammen, line 634924: This theorem is perhaps better understood, if you explicitly say that A is a primitive permutation that moves the I-th element to the end, and C moves the last element to the I-th position.

set.mm
metakunt: It is, I deleted casetrich, as I didn't want to search the database I thought it is just faster to implement it myself.

set.mm
metakunt: No I've meant right and left inverse (section, retraction) but I always switch the names so I just said one sided inverse so that I don't write the wrong names. Fixed the description of metakunt14 and clarified right/left inverse on the correct places.

wlammen (approved):

tirix (approved):

PR #4013, opened 2024-05-25
Comment edits: consistency around "(ordered-pair) class abstraction".

benjub:


benjub:
It began with shortening bj-opabco, and then, this leading to that, I edited several comments mentioning "class builder" or related terms to be more consistent and use preferably "class abstraction".

Note: this comment is purposefully not the initial comment of this PR to avoid putting it in the git history.

set.mm
```diff
@@ -40557,8 +40555,8 @@ among classes ( ~ eq0 , as opposed to the weaker uniqueness among sets,
 
   ${
     abf.1 $e |- -. ph $.
-    $( A class builder with a false argument is empty.  (Contributed by NM,
-       20-Jan-2012.) $)
+    $( A class abstraction determiend by a false formula is empty.
```
tirix, line 40558: typo: determined

tirix (approved):

set.mm
```diff
@@ -465412,8 +465454,8 @@ its graph has a given second element (that is, function value).
     fnimatp.2 $e |- ( ph -> A e. D ) $.
     fnimatp.3 $e |- ( ph -> B e. D ) $.
     fnimatp.4 $e |- ( ph -> C e. D ) $.
-    $( The image of a triplet under a function.  (Contributed by Thierry
-       Arnoux, 19-Sep-2023.) $)
+    $( The image of a triple under a function.  (Contributed by Thierry Arnoux,
```
icecream17, line 465457: above, "unordered triple" is used
additionally above, "unordered pair" is changed to simply "pair"

personally I prefer concision but changing this (and again once below) to "unordered triplet" is ok too

icecream17 (approved):

set.mm
```diff
@@ -43875,16 +43893,16 @@ will do (e.g., ` O = (/) ` and ` T = { (/) } ` , see ~ 0nep0 ).
 
   ${
     tpnz.1 $e |- A e. _V $.
-    $( A triplet containing a set is not empty.  (Contributed by NM,
```
tirix, line 43878: I see you removed the _unordered_ mention for pairs, why add it for triples?
Can't we, similarly, consider triple means unordered by default ?

tirix: Now I notice @icecream17's remark: this is basically the same...

tirix (approved):

set.mm
benjub: "Pairs", are "by default" unordered, though maybe the default is not that standard, and when the context is not sufficient, it may be more clear to write explicitly (even if redundantly) "unordered pair".  Maybe I removed it in this PR at places where it should have remained for clarity ?

On the other hand, tuples (hence all terms ending in -uple / -iple) or by default ordered, and here the default is really standard (https://en.wikipedia.org/wiki/Tuple).

tirix: Ok, I see!

benjub: I just added a clarification in my latest commit (and hopefully last). Sorry to ask you if you can approve it, @tirix !

set.mm
benjub: Thanks. Added.

PR #4014, opened 2024-05-25
Move polynomial section, fix df-mhp, + mhpvarcl

### LINKED ISSUE #3995: move polynomial section
## reason

[~ df-mhp](https://us.metamath.org/mpeuni/df-mhp.html) is wrong

The sum is over `I` not `NN0`, since `g : I --> NN0`.

This change also means [df-gsum](https://us.metamath.org/mpeuni/df-gsum.html) needs to be used instead of [df-sum](https://us.metamath.org/mpeuni/df-sum.html), since `I` can technically be infinite in size, and is not necessarily a range over the integers. (The defining property of `I` is being finely supported, which `gsum` handles nicely).

And since we are summing numbers, we need a structure whose base set includes `NN0`...

Unfortunately the complex numbers as an algebraic structure (and various substructures) come in the section right after the polynomials

So I propose to move the polynomial section.

## where

While simply moving it right after the number structures is the obvious move, I would actually like to move it somewhere after "vectors and free modules" because both seem potentially useful.

For instance, compare ~ frlmelbas and ~ mplelbas, or note how https://en.wikipedia.org/wiki/Hilbert%27s_Nullstellensatz#Formulation represents the "Assignments" below as a vector

```
evaluated polynomial = ((IndexVars eval Structure)`Polynomial)`(AssignmentsFromIVarsToStructValues)
```

though it could also go right before "Polynomial matrices" two sections after

icecream17:
and + fsuppind (mathbox)

Fixes #3995

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -106,6 +106,8 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+25-May-24 suppcofnd ---         deleted; its sole usage now uses
+                                suppcoss
 23-May-24 syl5bbr   bitr3id     compare to bitr3i or bitr3d
 23-May-24 elfzd     [same]      moved from GS's mathbox to main set.mm
 23-May-24 elfelzd   [same]      moved from GS's mathbox to main set.mm


wlammen:
14,250 additions, 14,087 deletions not shown because the diff is too large. Please use a local Git client to view these changes. 

This is what I get when I want to look at the diff in my browser.  Does it mean I have to download the fork icecream17:move onto my computer to see there what you are requesting?  Can't the move be broken up in pieces small enough to allow usual reviews?

icecream17:
I split the first commit into two parts. As seen by the checks, the resulting code is the same

This should be within the diff limits by github: https://docs.github.com/en/repositories/creating-and-managing-repositories/repository-limits#diff-limits

set.mm
```diff
@@ -77653,21 +77653,25 @@ any sets (which usually are functions) and any element (even not
     EFUPZSVGETUQWCVKVIRMRWCVJRVIWCAFGZVCHQVJRNWCVCWEWDSAETUQPVIURUSUTVAVB $.
 
   ${
-    $d x F $.  $d x G $.  $d x Z $.  $d x U $.  $d x W $.  $d x A $.  $d x B $.
```
wlammen: If something is not important in itself, why not mark it as a lemma of other theorems?  I can see that you were the author of this theorem, so, ok, I am willing to accept this deletion.  In general, though, there is a tendency to stay stable, even if a theorem loses during a development phase its references.

Maybe suppcoss is a sort of lemma as well?

set.mm
```diff
@@ -263599,6 +263599,8092 @@ the same dimension over the same (nonzero) ring.  (Contributed by AV,
     DOPUDUGSUEBBFCNQTRABUECUAUB $.
 
 
+$(
```
wlammen: The deleted part appears here without change.  At the end some hmldefs were also moved.  This reflects a different order of definitions now, I presume.

icecream17:
Can still review, but draft pending conversation at #3995

set.mm
icecream17: yeah, in retrospect suppcofnd should've been marked as a lemma.

I could see suppcoss being marked as a lemma too, though note that I also used it in mplsubglem (which goes similarly to mhpinvcl for the invcl part of subg). Not sure if that changes anything

wlammen: The idea here is that a lemma is tied to a target theorem.  If the proof of this theorem changes, or if it gets deleted altogether, then the lemmas will vanish too.  If you instead have a wider spread application range, i.e. at least two unrelated applications, then it is a more general utility that you want to keep even if one or both applications are dismissed.  I don't know what idea is prevailing.  That's your choice.

icecream17: I'd say it's useful enough not to be a lemma, then. It feels similarly as useful as suppofssd

tirix:
This moves big chapters around, so the chances of conflicts are high (e.g. if we add theorems from mathboxes to the moved sections). Maybe we should not wait for too long with this?

icecream17:
Perhaps it would be helpful to split this pr into two, and have one pr move the sections, while the other fixes df-mhp?

A follow up pr would be necessary anyway since the Associative Algebra section would be moved too (as suggested by @benjub), and I would have to make a version of fsuppind for homogeneous polynomials.

wlammen:
While it is certainly a good idea to do just one thing per pull request, I think it does not help here a lot.  To me, it seems people are awaiting @digama0 's confirmation (https://github.com/metamath/set.mm/issues/3995#issuecomment-2132270652).  The last post of him here on github I remember, is this one https://github.com/metamath/set.mm/issues/3990#issuecomment-2124160362 .  So either he is awfully busy, or on vacation, or...
Is his approval imperative, so we cannot proceed without it?  If possible, we should allow (slow?) development, even if he is known to be an expert in this field.  I don't like the idea, that otherwise contributions get blocked for an unknown, at least long time, before a decision is made.  This holds the more when the pull request is easily revertible.

icecream17:
I propose to just merge for now, per tirix and wlammen, then.

The one week time period has elapsed but it's a gray area since I don't have write access/merge permissions

wlammen:
I've seen no objection to this pull request, other than "Lets wait and see what others say".  On the other side there is confirmation by @tirix and @benjub for it.  If this state does not change until weekend, I will merge it.

PR #4015, opened 2024-05-26
Use elab2gw, elabgw to reduce axiom usage

GinoGiotto:
Most changes in this PR are of the same kind of https://github.com/metamath/set.mm/pull/3984 and https://github.com/metamath/set.mm/pull/3991. Specifically:

* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [elpwg](https://us.metamath.org/mpeuni/elpwg.html) -> Drop ax-10, ax-11
* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [eluni](https://us.metamath.org/mpeuni/eluni.html) -> Drop ax-10, ax-11, ax-12
* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [elintg](https://us.metamath.org/mpeuni/elintg.html) -> Drop ax-10, ax-11, ax-12
* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [elong](https://us.metamath.org/mpeuni/elong.html) -> Drop ax-10, ax-11
* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [isfin2](https://us.metamath.org/mpeuni/isfin2.html) -> Drop ax-10, ax-11
* Use [elabgw](https://us.metamath.org/mpeuni/elabgw.html) in [elgch](https://us.metamath.org/mpeuni/elgch.html) -> Drop ax-10, ax-11
* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [iswun](https://us.metamath.org/mpeuni/iswun.html) -> Drop ax-10, ax-11
* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [elnpi](https://us.metamath.org/mpeuni/elnpi.html) -> Drop ax-10, ax-11
* Use [elab2gw](https://us.metamath.org/mpeuni/elab2gw.html) in [issal ](https://us.metamath.org/mpeuni/issal.html) -> Drop ax-10, ax-11

I also removed usage of ax-10, ax-11, ax-12 from [opeq1](https://us.metamath.org/mpeuni/opeq1.html) and [opeq2](https://us.metamath.org/mpeuni/opeq2.html). For these two I did not apply a simple replacement like the revisions above, but I proved them again from scratch. Therefore revision tags and OLD proofs are provided.

All added DV conditions are with dummy variables.

----

Results of these changes:

Drop ax-10 from 626 theorems.
[...]

PR #4020, opened 2024-05-27
rmv ax-12 from ssel (ax-mulcom pr 7)

icecream17:
remove reference in remulid2 now that being a real number is not necessary to avoid ax-mulcom

(removing ax-12 from ssel at minimum removes ax-12 from elpwg: https://github.com/metamath/set.mm/pull/4015)


wlammen:
While mulid2 involves ax-mulcom, sn-mulid2 needs [ax-mulrcl](file:///home/wolf/metamath/web/ax-mulrcl.html).  What is so special about ax-mulcom, that it is better exchanged for another axiom?

wlammen (approved):

icecream17:
in my mathbox I am investigating whether ax-mulcom is independent, meaning all other axioms may be used.

ax-mulcom "pr 1" is arbitrarily defined as https://github.com/metamath/set.mm/pull/3713

PR #4023, opened 2024-05-28
Permresults

metakunt:
God this was such a pain to prove.....

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -82,7 +82,6 @@ proposed  syl5ibr   imbitrrid
 proposed  syl5ibrcom imbitrridcom
 proposed  syl5bb    bitrid      compare to bitri or bitrd
 proposed  syl5rbb   bitr2id     compare to bitr2i or bitr2d
-proposed  syl5bbr   bitr3id     compare to bitr3i or bitr3d
 proposed  syl5eq    eqtrid      compare to eqtri or eqtrd
 proposed  syl5req   eqtr2id     compare to eqtr2i or eqtr2d
 proposed  syl5eqr   eqtr3id     compare to eqtr3i or eqtr3d
@@ -107,6 +106,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+23-May-24 syl5bbr   bitr3id     compare to bitr3i or bitr3d
 23-May-24 elfzd     [same]      moved from GS's mathbox to main set.mm
 23-May-24 elfelzd   [same]      moved from GS's mathbox to main set.mm
 22-May-24 2exp5     [same]      moved from AV's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -5427,7 +5427,6 @@
 "drnf2" is used by "nfsb4tALT".
 "drnfc1" is used by "bj-nfcsym".
 "drnfc1" is used by "nfabd2".
-"drnfc1" is used by "nfabd2OLD".
 "drnfc1" is used by "nfcvb".
 "drnfc1" is used by "nfriotad".
 "drngoi" is used by "dvrunz".
@@ -6305,6 +6304,10 @@
 "funcringcsetclem7ALTV" is used by "funcringcsetcALTV".
 "funcringcsetclem8ALTV" is used by "funcringcsetcALTV".
 "funcringcsetclem9ALTV" is used by "funcringcsetcALTV".
+"funop" is used by "funopdmsn".
+"funop" is used by "funsndifnop".
+"funopsn" is used by "funop".
+"funopsn" is used by "funop1".
 "gen11" is used by "ax6e2eqVD".
 "gen11" is used by "ax6e2ndeqVD".
 "gen11" is used by "csbfv12gALTVD".
@@ -10114,7 +10117,6 @@
 "nfnae" is used by "distel".
 "nfnae" is used by "dvelimf".
 "nfnae" is used by "nfabd2".
-"nfnae" is used by "nfabd2OLD".
 "nfnae" is used by "nfald2".
 "nfnae" is used by "nfriotad".
 "nfnae" is used by "nfsb4t".
@@ -10188,7 +10190,6 @@
 "nfsbcd" is used by "nfsbc".
 "nfsbcd" is used by "sbcnestgf".
 "nfsbd" is used by "nfabd".
-"nfsbd" is used by "nfabd2OLD".
 "nfsbd" is used by "nfsb".
 "nfsbd" is used by "wl-sb8eut".
 "nic-ax" is used by "lukshef-ax1".
@@ -12199,6 +12200,7 @@
 "reglogleb" is used by "pellfund14".
 "reglogltb" is used by "pellfund14".
 "reglogmul" is used by "pellfund14".
+"relop" is used by "funopg".
 "relrngo" is used by "iscrngo2".
 "relrngo" is used by "isdrngo1".
 "relrngo" is used by "isrngo".
@@ -15758,7 +15760,7 @@ New usage of "drex2" is discouraged (4 uses).
 New usage of "drhmsubcALTV" is discouraged (1 uses).
 New usage of "drnf1" is discouraged (3 uses).
 New usage of "drnf2" is discouraged (3 uses).
-New usage of "drnfc1" is discouraged (5 uses).
+New usage of "drnfc1" is discouraged (4 uses).
 New usage of "drnfc2" is discouraged (0 uses).
 New usage of "drngcatALTV" is discouraged (0 uses).
 New usage of "drngoi" is discouraged (2 uses).
@@ -16026,11 +16028,13 @@ New usage of "elpjhmop" is discouraged (3 uses).
 New usage of "elpjidm" is discouraged (2 uses).
 New usage of "elpjrn" is discouraged (0 uses).
 New usage of "elpqn" is discouraged (24 uses).
+New usage of "elpr2OLD" is discouraged (0 uses).
 New usage of "elprnq" is discouraged (22 uses).
 New usage of "elpwOLD" is discouraged (0 uses).
 New usage of "elpwgOLD" is discouraged (0 uses).
 New usage of "elpwgded" is discouraged (2 uses).
 New usage of "elpwgdedVD" is discouraged (1 uses).
+New usage of "elpwi2OLD" is discouraged (0 uses).
 New usage of "elreal" is discouraged (7 uses).
 New usage of "elreal2" is discouraged (3 uses).
 New usage of "elringchomALTV" is discouraged (1 uses).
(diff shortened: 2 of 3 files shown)


set.mm
```diff
@@ -635310,6 +635400,361 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+      ORAUSSULAUSUMZPZHFUNUHZUTVFVGPBCDEFGHAVAVEVGITAVBVEVGJTAVCVEVGKTLMNAVDVEV
+      GOTAVEVGUOVFVGSUPVFVGUMZPBCDEFGHAVAVEVHITAVBVEVHJTAVCVEVHKTLMNAVDVEVHOTAV
+      EVHUOVFVHSUQURUR $.
+  $}
+
+  ${
+    metakunt24.1 $e |- ( ph -> M e. NN ) $.
+    metakunt24.2 $e |- ( ph -> I e. NN ) $.
+    metakunt24.3 $e |- ( ph -> I <_ M ) $.
+    $( Technical condition such that ~ metakunt17 holds (Contributed by
```
icecream17, line 635687:
```suggestion
    $( Technical condition such that ~ metakunt17 holds.  (Contributed by
```

set.mm
```diff
@@ -633662,6 +633662,96 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
       $.
   $}
 
+  ${
+    $d A x $.  $d F x $.  $d G x $.  $d ph x $.
+    eqfnfv2d.1 $e |- ( ph -> F Fn A ) $.
```
wlammen, line 633667: This is not a deduction version of eqfnfv2, since it proves the reverse direction of the bi-conditional only.  Use a different name (with suffix -2, for example), to allow for a later true deduction version.

set.mm
```diff
@@ -635310,6 +635400,361 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+    $d B y $.  $d C y $.  $d D y $.  $d I x $.  $d M x y $.  $d ph x y $.
+    metakunt25.1 $e |- ( ph -> M e. NN ) $.
+    metakunt25.2 $e |- ( ph -> I e. NN ) $.
+    metakunt25.3 $e |- ( ph -> I <_ M ) $.
+    metakunt25.4 $e |- B = ( x e. ( 1 ... M ) |-> if ( x = M , M ,
+     if ( x < I , ( x + ( M - I ) ) , ( x + ( 1 - I ) ) ) ) ) $.
+    metakunt25.5 $e |- C = ( x e. ( 1 ... ( I - 1 ) ) |->
+     ( x + ( M - I ) ) ) $.
+    metakunt25.6 $e |- D = ( x e. ( I ... ( M - 1 ) ) |->
+     ( x + ( 1 - I ) ) ) $.
```
icecream17, line 635737: C and D are extra hypotheses (they don't appear in the final statement). They may _optionally_ be removed with eqid in the proof itself if you wish.

Example: https://us.metamath.org/mpeuni/mhpdeg.html
Pro is less hypotheses for future theorems, con is longer proof (at minimum by syntax encoding and referencing eqid; in a compressed proof, a hypothesis is one letter unless there are an incredible amount of variables and hypotheses)

wlammen (changes requested):

set.mm
metakunt: Name changed

icecream17 (commented):
Whew, proving that B is a union of three disjoint functions by hand, congrats

<details><summary>brainstorming how to make proving this easier</summary>

It seems like Metamath is particularly unsuited for "unsophisticated" "brute force" proofs where there are multiple disjoint subparts. Now that I think about it in this meta way, Metamath proofs that use some kind of "builder" theorems are much easier... I tried to think of a way to "build" the if, but I don't see a way that's faster than what you did, other than straight out using https://us.metamath.org/mpeuni/mmtheorems139.html#mm13862b, proving permbuilders for each relevant function, and a final domain hack that makes the function go from 1 to M instead of 0 to M

Not sure how my advice here would've applied to [fsuppind](https://github.com/metamath/set.mm/pull/4014/commits/701b67033c7d294aa5eee08659f1ad9222fd7757) though. 

</details>

icecream17 (approved):

set.mm
```diff
@@ -633662,6 +633662,96 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+      IRLZAWICDUHSZPBWIUIUJUKWOWKWPWLWQWOWPWKWOWPWIWEONZWKWOWTBRLZWPXBURXAAXCWJ
+      GQZWIBULUJWOXBWKWOXBMZWICWEACRLZWJXBEUMXEBKWOXCXBXDQXEUNUOWOWTXBXAQWOCWIO
+      NZXBWJXGAWICDUPSQWOXBUSTWKXBWOWICWEUTZSVAVBVCWOWLWQWLWQWOWIBDUPZSWOWQMWIB
+      DWOXCWQXDQADRLZWJWQFUMWOWTWQXAQWOWQUSWOWIDONZWQWJXKAWICDUTSQTVAVDVEAWKWJW
+      LAWKMZWICDAXFWKEQAXJWKFQZWKWTAWICWEUHZSZWKXGAWICWEUPSXLWIWEDXLWIXOPXLBKAW
+      RWKWSQZXLVFVGZXLDXMPZWKXBAXHSXLWEBDXQXPXRAWEBONZWKAVHKONXSVIABKWSAVFVJVKQ
+      ABDONWKIQVLVLTAWLMZWICDAXFWLEQZAXJWLFQWLAWNWTWIWGWFVMWNAWMWTWIWFWGVNZAWMW
+      TAWKWTWLWKWTVOAXNVPWLWTVOAWIBDUHVPWAWBVQVRZXTCBWIXTCYAPAWRWLWSQXTWIYCPACB
+      ONWLHQWLWQAXISVLWLXKAWIBDUTSTVSVAYBVTWC $.
+  $}
```
avekens, line 633753: Why is this proof so long? It should be much shorter using ~fzsplitnd.

wlammen (commented):
Renaming is ok

wlammen (approved):

set.mm
metakunt: I have implemented that suggestion. In fact it was always my goal to not have them in the final step. I must have forgotten it due to the proof being such a colossal pain that I have forgotten I am at the last step.

metakunt:
> Whew, proving that B is a union of three disjoint functions by hand, congrats
> brainstorming how to make proving this easier
> 
> It seems like Metamath is particularly unsuited for "unsophisticated" "brute force" proofs where there are multiple disjoint subparts. Now that I think about it in this meta way, Metamath proofs that use some kind of "builder" theorems are much easier... I tried to think of a way to "build" the if, but I don't see a way that's faster than what you did, other than straight out using https://us.metamath.org/mpeuni/mmtheorems139.html#mm13862b, proving permbuilders for each relevant function, and a final domain hack that makes the function go from 1 to M instead of 0 to M
> 
> Not sure how my advice here would've applied to [fsuppind](https://github.com/metamath/set.mm/pull/4014/commits/701b67033c7d294aa5eee08659f1ad9222fd7757) though.

Honestly, that wasn't even the hardest part. What really, really made me desperate was proving that
```math
( ( \{ 1,\dots,i-1\} \cup \{i,\dots,m-1\})\cup\{m\}) = \{1,\dots,m\}
```
I thought that would be one of the easier theorems, but it was actually the hardest.
Honestly I've tried two approaches, one with A https://us.metamath.org/mpeuni/metakunt14.html , where I've used https://us.metamath.org/mpeuni/2fvidf1od.html , and a different one with B. Safe to say, they broke me both in different ways.

I suspect I might be doing something horribly wrong, but I don't know what it is yet.

set.mm
metakunt: It's because my brain went afk. Jokes aside, I tried to prove a different statement with weaker hypotheses, which would have been nicer to use. Turns out, the statement I've tried to prove was a contradiction, so I couldn't obviously prove it. I had to strengthen the hypotheses which sadly made the theorem that uses this statement ungodly ugly. But it worked and I've completely forgot that the thing I have just proved is completely and utterly useless. Well actually it is useful because I do need to use it, but I wished I could have used my wishful wrong version.

So the wrong theorem was something originally like this.
```
    fzsplitnr.1 $e |- ( ph -> M e. ZZ ) $.
    fzsplitnr.2 $e |- ( ph -> N e. ZZ ) $.
    fzsplitnr.3 $e |- ( ph -> K e. ZZ ) $.
    fzsplitnr.4 $e |- ( ph -> A e. ZZ ) $.
    fzsplitnr.5 $e |- ( ph -> B e. ZZ ) $.
    fzsplitnr.6 $e |- ( ph -> M <_ A ) $.
    fzsplitnr.7 $e |- ( ph -> B <_ N ) $.


    fzsplitnr $p |- ( ph -> ( A ...B ) =
     ( ( A ... ( K - 1 ) ) u. ( K ... B ) ) ) $=
```
Yeah that clearly is wrong, but it would have saved me a lot of steps.

Nonetheless, I have fixed the theorem, it is indeed one line.

metakunt:
> Whew, proving that B is a union of three disjoint functions by hand, congrats
> brainstorming how to make proving this easier
> 
> It seems like Metamath is particularly unsuited for "unsophisticated" "brute force" proofs where there are multiple disjoint subparts. Now that I think about it in this meta way, Metamath proofs that use some kind of "builder" theorems are much easier... I tried to think of a way to "build" the if, but I don't see a way that's faster than what you did, other than straight out using https://us.metamath.org/mpeuni/mmtheorems139.html#mm13862b, proving permbuilders for each relevant function, and a final domain hack that makes the function go from 1 to M instead of 0 to M
> 
> Not sure how my advice here would've applied to [fsuppind](https://github.com/metamath/set.mm/pull/4014/commits/701b67033c7d294aa5eee08659f1ad9222fd7757) though.

Ah just wait until I get to the point where I have to do distinction by 22 cases.

icecream17:
image

fzsplit ($k := i - 1$) and fzsuc ($n := m - 1$) would seem to neatly fit here

---

If distinction by 22 cases is actually necessary please explain it first so that we may find a better solution 😅

metakunt:
Ah crap, well I've missed that one.

To be fair, there is no other easier way. You need to split the set into 3 sets, each of those you need to split 3 more times, 2 of those sets you need to split twice again. That gives you 11 sets. That calculation you then need to do twice. You also need existence of three different values to make the calculation, as in 

```
( ph -> A e. NN )
( ph -> B e. NN )
( ph -> C e. NN )
( ph -> A =/= B )
( ph -> B =/= C  )
( ph -> A =/= C  )
```
and you need to check whether a given condition holds on each of those sets.

tirix (approved):

icecream17:
Incidentally, why does the calculation have to be done twice?

metakunt:
You are constructing two sets based of three indices a,b,c where set 1 depends on index a and b and set 2 depends on index a and c.  Then you split the base set up in 11 small parts that depend on all three.  Now you need to argue whether the a property in the set holds or not. In general those are arbitrary maps but in my specific case there is some stucture. But you still need to verify that the property holds for each of these sets.
There are 11 sets as they are chosen in a way such that most cancel out and the other are balanced in a delicate way that they cancel out by a different argument. So far I am not even close to even stating the theorem as I need to construct the final permutation first.

tirix:
If what you have to prove is symmetric in a/b and b/c, you're probably better off making it a lemma and proving it only once, so I suppose it's not?

metakunt:
Well the argument isn't that complex as for most of the sets the image is the same. For some of those the argument is equivalent, but some are the same by a different argument. And some are just plain different and have to be handled in a different way. But I am still far away from this theorem, I haven't even defined the main permutation yet, and that one is going to be doozy. I conjecture I am only at 20% of the progress and I will take a little bit of time off before I tackle it.

PR #4028, opened 2024-06-01
add immediate version of fndm and shorten proofs with it

wlammen:


wlammen:
The here introduced immediate version of [fndm](https://us.metamath.org/mpeuni/fndm.html) has 67 applications in Main alone, so I think its addition is justified.

Mathboxes were scanned for its application, too, sometimes fndmd was used.

set.mm shrinks by 21 text lines in total.

delete outdated OLD theorem (not counting in the total above).

tirix (approved):

icecream17 (approved):

PR #4030, opened 2024-06-02
Semiring of ideals (1)

tirix:
Basic statements about the semiring of ideals, but still not proving that it's actually a semiring.


set.mm
```diff
@@ -477442,6 +477720,87 @@ The sumset operation can be used for both group (additive) operations and
...
+       Arnoux, 2-Jun-2024.) $)
+    rspecval $p |- ( R e. Ring
+                 -> ( Spec ` R ) = ( ( IDLsrg ` R ) |`s ( PrmIdeal ` R ) ) ) $=
+      ( vr cv cidlsrg cfv cprmidl cress co crg wceq fveq2 oveq12d df-rspec ovex
+      crspec fvmpt ) BABCZDEZQFEZGHADEZAFEZGHIOQAJRTSUAGQADKQAFKLBMTUAGNP $.
+  $}
+
+  ${
+    rspecbas.1 $e |- S = ( Spec ` R ) $.
+    $( The prime ideal form the base of the spectrum of a ring.  (Contributed
```
icecream17, line 477735:
```suggestion
    $( The prime ideals form the base of the spectrum of a ring.  (Contributed
```

icecream17 (approved):

set.mm
```diff
@@ -477442,6 +477720,87 @@ The sumset operation can be used for both group (additive) operations and
...
+  $( Class of unique factorization domains. $)
+  cufd $a class UFD $.
+
+  ${
+    $d i r $.
+    $( Define the class of unique factorization domains.  A unique
+       factorization domain (UFD for short), is a commutative ring with an
+       absolute value (from ~ abvtriv this is equivalent to being a domain)
+       such that every prime ideal contains a prime element (this is a
+       characterization due to Irving Kaplansky).  UFD are also sometimes also
```
avekens, line 477784: 2x also

set.mm
```diff
@@ -477442,6 +477720,87 @@ The sumset operation can be used for both group (additive) operations and
...
+       called a "factorial rings" following the terminology of Bourbaki.
```
avekens, line 477785: " A UFD is sometimes also called a "factorial ring" ..."

set.mm
```diff
@@ -477442,6 +477720,87 @@ The sumset operation can be used for both group (additive) operations and
...
+       (Contributed by Mario Carneiro, 17-Feb-2015.) $)
+    df-ufd $a |- UFD = { r e. CRing | ( ( AbsVal ` r ) =/= (/) /\
+               A. i e. ( PrmIdeal ` r ) ( i i^i ( RPrime ` r ) ) =/= (/) ) } $.
+  $}
+
+  ${
+    $d A r $.  $d R i r $.
+    isufd.1 $e |- A = ( AbsVal ` R ) $.
```
avekens, line 477793: Why is a class variable provided for `( AbsVal `` R )` and not for `( PrmIdeal `` R )`and `( RPrime `` R )`? Since ` A` occurs only once, I think a class variable is not necessary here.

tirix: You're right, this is not coherent.

Do we have any convention about when to use class variables or not?
I tend to try to use a class variable for every structure extractor, so here I might instead add one for both ``( PrmIdeal ` R )`` and ``( RPrime ` R )``!

tirix: Ok so in my last commit, I ended up assigning a class variable to all 3:``( AbsVal ` R )``, ``( PrmIdeal ` R )`` and ``( RPrime ` R )``. This should allow the maximum flexibility (you can always use `eqid` that's not needed).

I hope it's fine too.

avekens (approved):

PR #4031, opened 2024-06-03
iset.mm: fix metamath-knife build error and a few derivative notes/theorems

jkingdon:
* the fix to the `conventions` markup was needed based on a [recent metamath-knife change](https://github.com/metamath/metamath-knife/pull/157)
* notes in mmil.html for various theorems related to derivatives
* add one new derivative theorem which is similar to one in set.mm


tirix (approved):

jkingdon:
Because of the fix for the broken build I'm going to merge this now. The other stuff in this pull request should be pretty minor but as usual, if you see any issues, feel free to bring them up and we can always make a followup pull request as needed.

PR #4033, opened 2024-06-04
use spvv in nfcr, paragraph html fix

icecream17:
Using spvv (instead of spw) in nfcr avoids the `y, A` dv (the proof gets longer but it's necessary)

Mention spvv in spw

nfcriv now doesn't need the `y, A` dv and becomes nfcri but shorter. nfcri becomes OLD, nfcriv renamed to nfcri

I took the time to do minimize_all for nfcriv usages.
For [fedgmullem2](https://us.metamath.org/mpeuni/fedgmullem2.html) it used 3eqtr3rd but otherwise it is a drop in replacement. No cascading dv effects.

---

**sectioning and section headers** section ending

While looking at conventions-comments I noticed a missing paragraph tag

**comments** section ending

This came since I suggested the `<p>` to `<br>` change in https://github.com/metamath/set.mm/pull/3932/files#r1625029195 to be slightly more semantic

It turns out `ul` (unordered list) can only have `li` (list item) as children, so the `br` goes inside the `li`. I think changing `p` to `br` increased the spacing by a pixel or few.

| before | after |
|---|---|
| image |  image |

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -106,6 +106,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 3-Jun-24 nfcriv    nfcri       dv condition was removed
 31-May-24 ringgrpd  [same]      moved from SN's mathbox to main set.mm
 31-May-24 crnggrpd  [same]      moved from SN's mathbox to main set.mm
 31-May-24 crngringd [same]      moved from SN's mathbox to main set.mm
--- a/discouraged
+++ b/discouraged
@@ -17352,6 +17352,7 @@ New usage of "nfceqiOLD" is discouraged (0 uses).
 New usage of "nfcrALT" is discouraged (0 uses).
 New usage of "nfcriOLD" is discouraged (0 uses).
 New usage of "nfcriOLDOLD" is discouraged (0 uses).
+New usage of "nfcriOLDOLDOLD" is discouraged (0 uses).
 New usage of "nfcriiOLD" is discouraged (0 uses).
 New usage of "nfcsb" is discouraged (6 uses).
 New usage of "nfcsbd" is discouraged (1 uses).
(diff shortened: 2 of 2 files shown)


set.mm
```diff
@@ -27556,13 +27561,15 @@ generally appear in a single form (either definitional, but more often
   $}
 
   ${
-    $d x y $.  $d y A $.
-    nfcriv.1 $e |- F/_ x A $.
-    $( Consequence of the not-free predicate, similiar to ~ nfcri .  Requires
-       ` y ` and ` A ` be disjoint, but is not based on ~ ax-13 .  (Contributed
```
avekens, line 27562: Was this theorem not depending on ax-13 before, but is now? Then the old theorem should be kept!

set.mm
```diff
@@ -27622,13 +27628,13 @@ generally appear in a single form (either definitional, but more often
...
-      ( vz cv wcel wsb wal nfcriv nfsbv nf5ri clelsb3 albii 3imtr3i ) EFCGZEBHZ
-      QAIBFCGZRAIQAPEBAAECDJKLEBCMZQRASNO $.
+      ( vz cv wcel wsb wal nfcri nfsbv nf5ri clelsb3 albii 3imtr3i ) EFCGZEBHZQ
+      AIBFCGZRAIQAPEBAAECDJKLEBCMZQRASNO $.
 
     $( Obsolete version of ~ nfcri as of 23-May-2024.  (Contributed by Mario
        Carneiro, 11-Aug-2016.)  (Proof modification is discouraged.)
        (New usage is discouraged.) $)
-    nfcriOLDOLD $p |- F/ x y e. A $=
+    nfcriOLDOLDOLD $p |- F/ x y e. A $=
```
avekens, line 27637: do we really need so many OLD versions? Since all modifications were done in a short time, the "intermediate" versions may not be necessary.

wlammen: He follows our usual pattern, which is what we want in 99% of the cases.  Here, I agree, there is an overkill of OLD variants with little extra insight into what we want to achieve.  Maybe we should simply squash the variants in an extra janitorial move outside of the user supplied contribution.  It should not block merging, in my eyes.

set.mm
wlammen: Neither spw, spvv nor nfcr depends on ax-13.

wlammen (approved):

tirix:
> changing `p` to `br` increased the spacing by a pixel or few.

The new spacing actually looks better! (to me)

> For [fedgmullem2](https://us.metamath.org/mpeuni/fedgmullem2.html) it used 3eqtr3rd but otherwise it is a drop in replacement. No cascading dv effects.

Good, thanks for the shortening.

wlammen:
Possible improvement: x, y need not be distinct in nfcr.

wlammen:
I don't see a good reason to not merge this PR, despite the open discussion about the OLD..OLD versions.  What are the options?
1. Leave it as is as an isolated exotic case, unimportant enough to keep it for a year, even if frowned upon.
2. Update the conventions and add rules for such cases in future.
3. Anyone who is up to it looks into the proofs and sorts out the unimportant versions.

PR #4036, opened 2024-06-07
Add iccioo01 to set.mm mathbox.

jkingdon:
The closed unit interval is equinumerous to the open unit interval, proved using the Schroeder-Bernstein theorem.

The Mastodon thread being referenced is https://mathstodon.xyz/@ProfKinyon/112555509670920854

Although I didn't see any fundamental obstacle to formalizing the alternate proof given there, it also was starting to look like it would be a lot longer than the one via Schroeder-Bernstein.


icecream17 (approved):

wlammen (approved):

tirix (approved):

avekens (approved):
Interesting result, with a relatively short proof.

jkingdon:
> Interesting result, with a relatively short proof.

The Mastodon thread is also worth a read, with discussion of different approaches people might take to this.

PR #4037, opened 2024-06-08
rename NaryF to -aryF and switch args

### LINKED ISSUE #4035: rename symbol NaryF to -aryF
this switches the order of the arguments so many theorems will have to be reproved

`( X NaryF N )` => `( N -aryF X )`

<s>I marked it as good first issue</s> but I'll probably do it soon after #4034 gets merged

edit: mobile github app has bugs

edit 2: **working on it**

icecream17:
fixes #4035

also commuted:
antecedent of naryfvalel (would lengthen if not for sylan9bb)
consequent of naryrcl (prevents extra ancoms)
antecedent of naryfvalelwrdf (lengthens proof)

--- explanatory comments
there are shortenings but these are either minimize_all or "trivial" (commuting or using i/d/g/etc versions of theorems in ways that are just outside current automation)

proofs of 1aryenef & 2aryenef seem lengthened (0 compressed byte change tho) since ovprc2 is added

I don't think changes-set needs to be changed but here's a convenient copy-paste if it does:
```
 7-Jun-24 df-naryf  [same]      operands switched
```


set.mm
```diff
@@ -759865,61 +759865,58 @@ this concept is not fully developed yet (it is within a mathbox), and it is
...
        4-Jun-2024.) $)
-    naryfvalelwrdf $p |- ( ( X e. V /\ N e. NN0 ) -> ( F e. ( X NaryF N )
+    naryfvalelwrdf $p |- ( ( N e. NN0 /\ X e. V ) -> ( F e. ( N -aryF X )
                            <-> F : { w e. Word X | ( # ` w ) = N } --> X ) ) $=
-      ( wcel cn0 wa cnaryf co cc0 cfzo cmap wf cv chash cfv wceq cword crab
-      eqid naryfvalel wrdnval eqcomd feq2d bitrd ) EDFCGFHZBECIJFEKCLJZMJZEBNAO
-      PQCRAESTZEBNBUHCDEUHUAUBUGUIUJEBUGUJUIACEDUCUDUEUF $.
+      ( cn0 wcel wa cnaryf co cc0 cfzo cmap wf cv chash cfv wceq cword crab
+      eqid naryfvalel wrdnval ancoms feq2d bitr4d ) CFGZEDGZHZBCEIJGEKCLJZMJZEB
+      NAOPQCRAESTZEBNBUJCDEUJUAUBUIULUKEBUHUGULUKRACEDUCUDUEUF $.
```
icecream17, line 759919: it looks like it gets longer for no reason, but really there are two changes:
* ancoms (needed since the antecedent was commuted, or if the antecedent wasn't commuted, it would be since naryfvalel doesn't match wrdnval)
* bytes saved from using bitr4d (and avoiding eqcomd)

wlammen (approved):

tirix (approved):

set.mm
```diff
@@ -442237,9 +442237,9 @@ orthogonal vectors (i.e. whose inner product is 0) is the sum of the
 htmldef "digit" as 'digit';
  althtmldef "digit" as 'digit';
  latexdef "digit" as "\mathrm{digit}";
-htmldef "NaryF" as ' NaryF ';
- althtmldef "NaryF" as ' NaryF ';
- latexdef "NaryF" as "\mathrm{NaryF}";
+htmldef "-aryF" as ' -aryF ';
+ althtmldef "-aryF" as ' -aryF ';
```
avekens, line 442241: I think we should remove the the space at the beginning (' -aryF '=> '-aryF '), so that we will see `( N-aryF X )` instead of `( N -aryF X `).

avekens: > I think we should remove the the space at the beginning (' -aryF '=> '-aryF '), so that we will see `( N-aryF X )` instead of `( N -aryF X `).

PS: not in set.mm, but in the the HTML-representation

avekens:
Thank you very much, @icecream17 , for these changes. It is a nice side effect that many of the proofs became shorter, but we should not pay much attention on minimal changes of the lengths here (it does not matter that some proofs became a little bit longer). An entry in changes-set.txt is not necessary, because all changes are within my mathbox.

set.mm
```diff
@@ -759848,15 +759848,15 @@ this concept is not fully developed yet (it is within a mathbox), and it is
...
+  $c -aryF $.  $( Constant symbol for the n-ary functions. $)
 
   $( Extend the definition of a class to include the n-ary functions. $)
-  cnaryf $a class NaryF $.
+  cnaryf $a class -aryF $.
 
   ${
     $d n x $.
     $( Define the n-ary (endo)functions.  (Contributed by AV, 11-May-2024.) $)
-    df-naryf $a |- NaryF = ( x e. _V , n e. NN0
```
avekens, line 759859: @icecream17 if you like, you can add a "Revised by" tag here with your and/or @tirix's name ,

avekens (approved):

PR #4040, opened 2024-06-10
Add WLPO to iset.mm

jkingdon:
We already have Omni and Markov (where `_om e. Omni` is LPO and `_om e. Markov` is MP). Given that precedent, the obvious notation for WLPO is `_om e. WOmni` (well, or some other name for `WOmni` but if there is an established name for what I am calling "weakly omniscient" I haven't seen it).

Includes a few theorems analogous to ones we have for `Omni` and `Markov`, and the usual theorem relating the three, `A e. Omni <-> ( A e. WOmni /\ A e. Markov )`


icecream17 (approved):

jkingdon:
I'm going to merge this now rather than go to greater lengths to solicit feedback (especially about the phrase "weakly omniscient" and the notation WOmni). If we think of something better it shouldn't be too hard to change later.

PR #4049, opened 2024-06-16
The Zariski topology is a topology

tirix:
Includes proofs for:
- the Zariski topology is a topology (~zartop)
- the Zariski topology is T0 (~zart0)
- the Zariski topology is T1 (~zarmxt1) when restricted to the maximal ideals.
Pending:
- the Zariski topology is compact.

The whole "spectrum of a ring" section has been moved inside of the "topology" section of my mathbox.
I found it to be very interesting and instructive to develop.


set.mm
```diff
@@ -481233,6 +481329,394 @@ such that every prime ideal contains a prime element (this is a
...
+  ${
+    zarclsx.1 $e |- V = ( i e. ( LIdeal ` R )
+                                    |-> { j e. ( PrmIdeal ` R ) | i C_ j } ) $.
+
+    ${
+      $d .0. i j $.  $d P i $.  $d R i j $.
+      zarcls0.1 $e |- P = ( PrmIdeal ` R ) $.
+      zarcls0.2 $e |- .0. = ( 0g ` R ) $.
+      $( The closure of the identity ideal in the Zariski topology.
+         (Contributed by Thierry Arnoux, 16-Jun-2024.) $)
```
avekens, line 481418: What "is" a Zariski topology? Is it ` V `? According to Wikipedia, "the Zariski topology is a topology defined on geometric objects called varieties.". A short explanation, and a reference to literature (at least to Wikipedia) would be very helpful.

set.mm
```diff
@@ -481233,6 +481329,394 @@ such that every prime ideal contains a prime element (this is a
...
+        VBVDVAVIVKUDVLFVDVLVAVDVHKZFVDKVAVIVKUFZVJVKVAVMVNVDBUGUHBVHVDFVHUNZIUI
+        UJUKULUMVEDVFUOUPUQBVHFVOIURAQKVAABOHUSSUT $.
+    $}
+
+    ${
+      $d B i j m $.  $d I i j m $.  $d R i j m $.  $d V m $.
+      zarcls1.1 $e |- B = ( Base ` R ) $.
+      $( The intersection of a family of closed sets is closed in the Zariski
+         topology.  (Contributed by Thierry Arnoux, 16-Jun-2024.) $)
+      zarcls1 $p |- ( ( R e. CRing /\ I e. ( LIdeal ` R ) )
```
avekens, line 481433: What is "family of closed sets "? Wahat is its "intersection"? I cannot see them in the theorem (and its hypotheses).

set.mm
```diff
@@ -481233,6 +481329,394 @@ such that every prime ideal contains a prime element (this is a
...
+    zartop $p |- ( R e. CRing -> J e. Top ) $=
+      ( vi vj vk ccrg wcel cprmidl cfv ctopon ctop clidl cv wss crab cmpt wceq
+      crn ccld eqid sseq1 rabbidv cbvmptv zartopn simpld topontop syl ) AIJZCAK
+      LZMLJZCNJUKUMFAOLZFPZGPZQZGULRZSZUACUBLTULABHGCUSDEULUCFHUNURHPZUPQZGULRU
+      OUTTUQVAGULUOUTUPUDUEUFUGUHULCUIUJ $.
+
+    ${
+      $d P i j k $.
+      zartop.3 $e |- P = ( PrmIdeal ` R ) $.
+      $( The base set of the Zariski topology.  (Contributed by Thierry Arnoux,
```
avekens, line 481643: What is ment by "base set" here?

set.mm
```diff
@@ -481233,6 +481329,394 @@ such that every prime ideal contains a prime element (this is a
...
+      zartopon $p |- ( R e. CRing -> J e. ( TopOn ` P ) ) $=
+        ( vi vj vk ccrg wcel ctopon cfv clidl cv wss crab cmpt wceq crn rabbidv
+        ccld sseq1 cbvmptv zartopn simpld ) BKLDAMNLHBONZHPZIPZQZIARZSZUADUCNTA
+        BCJIDUMEFGHJUHULJPZUJQZIARUIUNTUKUOIAUIUNUJUDUBUEUFUG $.
+    $}
+
+    $d J d x y $.  $d R d j k $.  $d R i j x y $.
+    $( The Zariski topology is T_0 .  (Contributed by Thierry Arnoux,
+       16-Jun-2024.) $)
+    zart0 $p |- ( R e. CRing -> J e. Kol2 ) $=
```
avekens, line 481654: "T_0" = "Kol2"? "T_0" = "Kolmogorov spaces" (see df-t0). Maybe this shoulkd be added here.

set.mm
```diff
@@ -481233,6 +481329,394 @@ such that every prime ideal contains a prime element (this is a
...
+      DUPUQVEVFVIXQWQYQPZQZWRYRWTYTUUEWQYQWPXQUUDSZVJUUEWQYQWSUUFVJVKVLVMVSYRXN
+      YSYPYSIWPXFXDWPWSRVOVPVRVTWIWAWBWCWNFHXFXICGWNCXFWDMKZXFCWEPXFABCDEXFWFZW
+      GXFCWHVRWNUUGXICWJMPXFABJICXHDEUUHYHWKWLWMVS $.
+
+    ${
+      $d R i j k l $.  $d R k l m $.  $d T m $.
+      zarmxt1.1 $e |- M = ( MaxIdeal ` R ) $.
+      zarmxt1.2 $e |- T = ( J |`t M ) $.
+      $( The Zariski topology restricted to maximal ideals is T_1 .
+         (Contributed by Thierry Arnoux, 16-Jun-2024.) $)
```
avekens, line 481679: "T_1" = "Fre"? "T_1" = "Fréchet spaces" (see df-t1). Maybe this shoulkd be added here.

avekens:
Very interesting results, but I am missing some explanations (see inline comments).

set.mm
tirix: The topology itself is a collection of _open sets_. This is the `J` in ~ rspectopn and ~ zartop.

But the textbook definitions all start from _closed sets_. This is also the case for [Wikipedia](https://en.wikipedia.org/wiki/Zariski_topology#Spectrum_of_a_ring).

The closed sets are generated from the ideals using the function `V`, so that the closed sets are actually `ran V`: see ~ zatopn .

The bridge between the closed set `ran V` and the open sets ` J ` is made in ~ zarcls, simply taking the complements.

set.mm
tirix: Sorry, this is a copy/paste error. Changing to the correct comment:
> The unit ideal ` B ` is the only ideal which closure in the Zariski topology is the empty set.

set.mm
tirix: This is meant in the same sense as in [~ tpsuni](https://us.metamath.org/mpeuni/tpsuni.html) : the underlying "set of points" are prime ideals.

The comment is modelled after e.g. [~xkotopon](https://us.metamath.org/mpeuni/xkotopon.html).

set.mm
tirix: Yes, $T_0$ spaces are also called [Kolmogorov spaces](https://en.wikipedia.org/wiki/Separation_axiom), and this is formalized as `Kol2` in set.mm.

The right place to explain this is indeed in [df-t0](https://us.metamath.org/mpeuni/df-t0.html). I did not dig about why the "2" in `Kol2`.

set.mm
tirix: Similar to previous comment, $T_1$ spaces are also called Fréchet spaces, and this is formalized as `Fre` in set.mm.
To me, the right place to explain this is indeed in [df-t1](https://us.metamath.org/mpeuni/df-t1.html). 

The prime spectrum restricted to maximal ideals is called the [maximal spectrum](https://ncatlab.org/nlab/show/maximal+spectrum) and denoted $mSpec$, $Spec_m$ or sometimes $Spec_{max}$. That would correspond to ``( S |`s M )``, and its topology would be the `` ( J |`t M ) `` of this theorem. But it does not have its own Wikipedia page. Maybe @savask can help tell if it's an important concept?

tirix:
Note: this PR is a follow-up to [this remark](https://github.com/metamath/set.mm/pull/3787#pullrequestreview-1991112143) from @savask.

set.mm
```diff
@@ -476838,8 +476915,29 @@ The sumset operation can be used for both group (additive) operations and
     ${
       $d .X. a $.  $d .X. e i $.  $d B e i $.  $d E a e i $.  $d G e i $.
       $d I a e i $.  $d R e i $.  $d a ph $.  $d e i ph $.
-      ringlsmss2.1 $e |- ( ph -> E C_ B ) $.
-      ringlsmss2.2 $e |- ( ph -> I e. ( LIdeal ` R ) ) $.
+      ringlsmss1.1 $e |- ( ph -> R e. CRing ) $.
+      ringlsmss1.2 $e |- ( ph -> E C_ B ) $.
+      ringlsmss1.3 $e |- ( ph -> I e. ( LIdeal ` R ) ) $.
+      $( The product to the left with an ideal of a commutative ring is a
```
savask, line 476921: I think the phrase "The product to the left" is a bit confusing. Maybe say something like "The product of an ideal ` I ` of a commutative ring with some set ` E ` is a subset of the ideal." ?

In general, I find the theorem a bit weird, in a commutative ring left ideals are two-sided ideals, see [crng2idl](https://us.metamath.org/mpeuni/crng2idl.html). Perhaps you need this result for something and it was easier to prove it that way?

set.mm
```diff
@@ -477711,6 +477837,26 @@ The sumset operation can be used for both group (additive) operations and
...
 
+  ${
+    idlsrgmulrss1.1 $e |- S = ( IDLsrg ` R ) $.
+    idlsrgmulrss1.2 $e |- B = ( LIdeal ` R ) $.
+    idlsrgmulrss1.3 $e |- .(x) = ( .r ` S ) $.
+    idlsrgmulrss1.4 $e |- .x. = ( .r ` R ) $.
+    idlsrgmulrss1.5 $e |- ( ph -> R e. CRing ) $.
+    idlsrgmulrss1.6 $e |- ( ph -> I e. B ) $.
+    idlsrgmulrss1.7 $e |- ( ph -> J e. B ) $.
+    $( In a commutative ring, the product of two ideals is a subset of the
```
savask, line 477848: Of course it is also a subset of the second one as well and hence a subset of the intersection :-)

savask (commented):
Great work! My remarks are minor and non-blocking in any way.

set.mm
tirix: Indeed, the proof would have been shorter using ~crng2idl!
I'll update the comment. I wanted to keep short proofs and this seemed like an interesting closure-like theorem.

set.mm
savask: I'm out of my depth here, but it seems that the concept is important enough (wikipedia does mention it in passing), but it doesn't behave that well as `Spec`.

set.mm
tirix: Yes, I'll add a theorem about the intersection, that's an interesting one.

set.mm
tirix: I've updated this comment, with:
> The points of the Zariski topology are the prime ideals.

set.mm
tirix: I've added a short explanation in the chapter's heading.

set.mm
tirix: Fixed in last commit.

set.mm
tirix: I'll leave it aside for now. 
From the Wikipedia page, it looks like it was Grothendieck's idea to generalize from the maximal spectrum $Spec_m$ to the prime spectrum $Spec$.
In our case we have all the tools to restrict, and many properties are inherited in that case. So it's worth noting that the maximal spectrum is easy to construct with the available formalization.

avekens (approved):

set.mm
```diff
@@ -476842,8 +476919,30 @@ The sumset operation can be used for both group (additive) operations and
     ${
       $d .X. a $.  $d .X. e i $.  $d B e i $.  $d E a e i $.  $d G e i $.
       $d I a e i $.  $d R e i $.  $d a ph $.  $d e i ph $.
-      ringlsmss2.1 $e |- ( ph -> E C_ B ) $.
-      ringlsmss2.2 $e |- ( ph -> I e. ( LIdeal ` R ) ) $.
+      ringlsmss1.1 $e |- ( ph -> R e. CRing ) $.
+      ringlsmss1.2 $e |- ( ph -> E C_ B ) $.
+      ringlsmss1.3 $e |- ( ph -> I e. ( LIdeal ` R ) ) $.
+      $( The product of an ideal ` I ` of a commutative ring ` R ` with some
+         set E is a subset of the ideal.  (Contributed by Thierry Arnoux,
```
icecream17, line 476926: would it be more accurate to say ``a subset ` E ` ``? I might be missing something because it looks like ringlsmss1 can be directly proved from ringlsmss2, and since ringlsmss2's proof didnt change I would assume it doesnt use ringlsmss1.

set.mm
```diff
@@ -481237,6 +481348,405 @@ such that every prime ideal contains a prime element (this is a
...
+        eqid rabid2 sylibr syl5req lidl0 fvexi fvmptd ) BJKZCFLZCMZDMZNZDBOPZUA
+        ZABUEPZEQECVHVGUBRVAGSVAVCVBRZTZAVFVGHVJVEDVFUCVFVGRVJVEDVFVJVDVFKZTZVC
+        VBVDVAVIVKUDVLFVDVLVAVDVHKZFVDKVAVIVKUFZVJVKVAVMVNVDBUGUHBVHVDFVHUNZIUI
+        UJUKULUMVEDVFUOUPUQBVHFVOIURAQKVAABOHUSSUT $.
+    $}
+
+    ${
+      $d B i j m $.  $d I i j m $.  $d R i j m $.  $d V m $.
+      zarcls1.1 $e |- B = ( Base ` R ) $.
+      $( The unit ideal ` B ` is the only ideal which closure in the Zariski
```
icecream17, line 481460:
```suggestion
      $( The unit ideal ` B ` is the only ideal whose closure in the Zariski
```

icecream17 (approved):
minor, this is very useful progress

set.mm
tirix: The big difference between ~ringlsmss1 and ~ringlsmss2 is that the former requires the ring to be commutative, while the latter does not.

With $I$ being a left-ideal and $E$ any set in the base set, $EI$ is always a left-ideal, but $IE$ is only an ideal if the ring is commutative. @savask noted that if the ring is commutative, we don't need to speak about left- or -right ideals anymore, since all ideals are two-sided by [crng2idl](https://us.metamath.org/mpeuni/crng2idl.html).

wlammen (approved):

PR #4057, opened 2024-06-22
Decidability of real number equality implies WLPO

### LINKED ISSUE #4041: decidability of real number equality implies WLPO
Or in the notation of #4040 , `A. x e. RR A. y e. RR DECID x = y -> _om e. Womni`

This theorem is taken from "Analytic WLPO" at https://ncatlab.org/nlab/show/principle+of+omniscience#analytic

jkingdon:
This proof turns out to be quite similar to https://us.metamath.org/ileuni/trilpo.html and in fact can use many of the same lemmas.

* Includes `iswomninn` which is similar to `isomninn` but for weak
omniscience (including lemma `iswomninnlem` ).

* Includes `enwomni` which says that weak omniscience is invariant with respect to equinumerosity.

Fixes #4041


iset.mm
```diff
@@ -154327,6 +154363,119 @@ equivalent way to state real trichotomy (see further discussion at
...
+       choice to prove the converse.
+
+       Here's the outline of the proof.  Given an infinite sequence F of zeroes
+       and ones, we need to show the sequence is all ones or it is not.
+
+       Construct a real number A whose representation in base two consists of a
+       zero, a decimal point, and then the numbers of the sequence.  This real
+       number will equal one if and only if the sequence is all ones
+       ( ~ redcwlpolemeq1 ).  Because (by hypothesis) real number equality is
+       decidable, it is decidable whether the sequence is all ones.
```
avekens, line 154460: If I interpret this hypothesis correctly, it is not proven that real number equality is decidable (in iset.mm or in general?) yet, is it? That means that it is unclear if a theorem like ~ qdceq for rationals exists for reals, isn't it?
Maybe this should be documented somewhere...

jkingdon: Ah thanks for saying something, that shows me that I didn't explain the point of these results very well.

The idea here is that LPO and WLPO are constructive taboos (or to put it less colorfully, that they cannot be derived from the axioms of IZF). I've pushed up another commit which adds some language to that effect and also re-words the sentence you are reacting to here (the use of "hypothesis" here was supposed to be akin to a proof which derives a contradiction but it was confusing wording, so I just reworked the whole sentence).

Not sure what you mean by "in general" as contrasted with iset.mm, but if you mean set.mm, all propositions are decidable there (including real number equality of course).

avekens (approved):

icecream17 (approved):

PR #4061, opened 2024-06-26
ax-mulcom pr 9 (mulgt0 results)

icecream17:
mathbox only. This strengthens mulgt0 to mulgt0b2d, so overall a very minor pr


icecream17:
With the hint from https://github.com/metamath/set.mm/pull/3792#issuecomment-1909467443 I found https://math.stackexchange.com/questions/2964961/is-mathbbr-the-only-complete-ordered-abelian-group

So multiplication is commutative if addition is commutative, since we have a complete, ordered, nontrivial, but not necessarily Abelian group.

tirix (approved):
Still, that is some progress in the project!

PR #4063, opened 2024-06-26
Several utility theorems

tirix:
This adds several utility theorems to my mathbox, and moves corresponding required theorems to main.

The two main theorems are:
- `gsumpart`, which rewrites a group sum as a double sum, grouping along a possibly infinite partition (but the sum is over a finite support, so the overall sum remains finite)
- `gsumhashmul`, which expresses a group sum by grouping by nonzero values, multiplied by the number of times those values appear in the sum.

So, nothing exciting in this PR, but some ground work which will hopefully be useful in the upcoming ones.

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -105,6 +105,14 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+26-Jun-24 rnresss   [same]      moved from GS's mathbox to main set.mm
+26-Jun-24 fndmexd   [same]      moved from RR's mathbox to main set.mm
+26-Jun-24 iunssf    [same]      moved from GS's mathbox to main set.mm
+26-Jun-24 rp-imass  resssxp     moved from RP's mathbox to main set.mm
+26-Jun-24 sscon34b  [same]      moved from RP's mathbox to main set.mm
+26-Jun-24 rcompleq  [same]      moved from RP's mathbox to main set.mm
+26-Jun-24 funresd   [same]      moved from GS's mathbox to main set.mm
+25-Jun-24 cmnmndd   [same]      moved from SN's mathbox to main set.mm
 16-Jun-24 syl6eqr   eqtr4di     compare to eqtr4i or eqtr4d
 16-Jun-24 elrnmptd  [same]      moved from GS's mathbox to main set.mm
 10-Jun-24 bj-intss  intss2      moved from BJ's mathbox to main set.mm


set.mm
```diff
@@ -464521,6 +464601,17 @@ Class abstractions (a.k.a. class builders)
...
       adantl syl5bb ) ABJZCKZLZIBCUFMZDNOZUIPZDIJZUKKUFULQUJKZUHULDKZUJUFULBRIR
       SUGUMUNTAUMUGUNBCDULUAUBUDUEUC $.
+
+    $d A x z $.  $d B z $.  $d C z $.  $d ph z $.  $d x y z $.
+    iunsnima2.1 $e |- F/_ x C $.
+    iunsnima2.2 $e |- ( x = y -> B = C ) $.
+    $( Version of ~ iunsnima with different variables.  (Contributed by Thierry
+       Arnoux, 22-Jun-2024.) $)
+    iunsnima2 $p |- ( ( ph /\ y e. A ) -> ( U_ x e. A ( { x } X. B ) " { y } )
+       = C ) $=
```
avekens, line 464611: I think the setvar variable `y` can be replaced by a class variable `Y`, since `Y` is actually a set because of `Y e. A`.

set.mm
```diff
@@ -465681,6 +465847,107 @@ its graph has a given second element (that is, function value).
...
+                            -> ( F supp Z ) = ( `' F " ( ran F \ { Z } ) ) ) $=
+    ( wfun wcel w3a ccnv crn cima csn cdif cdm csupp co wceq cnvimarndm difeq1d
+    a1i difpreima 3ad2ant1 wss suppssdm dfss4 mpbi suppiniseg difeq2d 3eqtr4rd
+    syl5eqr ) AEZABFZDCFZGZAHZAIZJZUNDKZJZLZAMZURLZUNUOUQLJZADNOZUMUPUTURUPUTPU
+    MAQSRUJUKVBUSPULUOUQATUAUMVCUTUTVCLZLZVAVCUTUBVEVCPADUCVCUTUDUEUMVDURUTABCD
+    UFUGUIUH $.
+
+  $( Finite support implies finite range, when the zero is excluded.
+     (Contributed by Thierry Arnoux, 24-Jun-2024.) $)
+  fsupprnfi $p |- ( ( ( Fun F /\ F e. V ) /\ ( .0. e. W /\ F finSupp .0. ) )
```
avekens, line 465942: The range is still finite when the zero is not excluded, isn't it?

set.mm
tirix: You're right! I'll do the change.

set.mm
tirix: Correct, that's a much better and simpler statement, thanks. I'll update the theorem.

tirix:
Thanks to @avekens for very relevant remarks, both are taken into account.

set.mm
```diff
@@ -465722,6 +465894,108 @@ its graph has a given second element (that is, function value).
...
+     zero.  (Contributed by Thierry Arnoux, 24-Jun-2024.) $)
+  supppreima $p |- ( ( Fun F /\ F e. V /\ Z e. W )
+                            -> ( F supp Z ) = ( `' F " ( ran F \ { Z } ) ) ) $=
+    ( wfun wcel w3a ccnv crn cima csn cdif cdm csupp co wceq cnvimarndm difeq1d
+    a1i difpreima 3ad2ant1 wss suppssdm dfss4 mpbi suppiniseg difeq2d 3eqtr4rd
+    syl5eqr ) AEZABFZDCFZGZAHZAIZJZUNDKZJZLZAMZURLZUNUOUQLJZADNOZUMUPUTURUPUTPU
+    MAQSRUJUKVBUSPULUOUQATUAUMVCUTUTVCLZLZVAVCUTUBVEVCPADUCVCUTUDUEUMVDURUTABCD
+    UFUGUIUH $.
+
+  $( Finite support implies finite range, when the zero is excluded.
```
avekens, line 465987: The comment was not updated,

tirix: Done!

avekens (approved):

icecream17 (approved):

PR #4068, opened 2024-06-30
0a = 0 (mulcom pr 10)

icecream17:
mathbox

---

comments:

I'm glad this is possible, though it's quite involved

In the middle of proving 0tie0 I messed up. mmj2 cant edit incomplete metamath.exe proofs, but the `show new /lemmon /ren /no_repeat` command was similar enough to mmj2's syntax that I could get it to work with some editing, copying, and pasting. At minimum, this suggests it would be relatively easy to make a script for the use case of editing incomplete metamath.exe proofs in mmj2 (and also yamma).

Also I suspect that windows wraps at 79 columns instead of 80 because in windows, metamath.exe uses CRLF line endings instead of LF line endings. (I did `git diff` and the different line endings resulted in a failed memory allocation. Luckily my computer only lagged at 100% disk for a minute)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -189,7 +189,7 @@
 "1sr" is used by "axi2m1".
 "1sr" is used by "axicn".
 "1sr" is used by "supsr".
-"1t1e1ALT" is used by "0tie0".
+"1t1e1ALT" is used by "sn-0tie0".
 "1t1e1ALT" is used by "nnmul1com".
 "1t1e1ALT" is used by "remulinvcom".
 "2ax6e" is used by "2sb5rf".
--- a/discouraged
+++ b/discouraged
@@ -189,6 +189,7 @@
 "1sr" is used by "axi2m1".
 "1sr" is used by "axicn".
 "1sr" is used by "supsr".
+"1t1e1ALT" is used by "0tie0".
 "1t1e1ALT" is used by "nnmul1com".
 "1t1e1ALT" is used by "remulinvcom".
 "2ax6e" is used by "2sb5rf".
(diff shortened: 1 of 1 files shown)


wlammen (approved):

wlammen:
It should be part of the specification, what the kind of line ending is, to avoid any hassle with different operating system.  If metamath(.exe) is generous enough to accept all kinds of endings, (CR on Mac Classic, CR LF on Windows, LF on Linux) then the usable space for text before this ending should stay fixed at 80 characters.

tirix (approved):
Nice! 
`0tie0` looks impressive, unfortunately it is not so easy to guess the proof details at the compressed proof...

set.mm
```diff
@@ -640139,31 +640229,80 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
   $( $j usage 'sn-inelr' avoids 'ax-mulcom'; $)
 
-  $(
-    Using ~ sn-inelr we can show:
-    ` ( r + _i s ) = ( t + _i u ) <-> ( r = t /\ s = u ) ` but so far there is
-    no use for such a theorem.
-  $)
+  ${
+    $d R x $.
+    $( ` _i ` times a real is non-real iff the real is non-zero.  (Contributed
```
avekens, line 640249: Why is the comment negated? "` _i ` times a real is real iff the real is zero." sounds better.

set.mm
```diff
@@ -639994,6 +640001,89 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
     $( $j usage 'remulcand' avoids 'ax-mulcom'; $)
   $}
 
+  ${
+    $d a b x y $.
+    $( Lemma for ~ sn-mul02 . ` 0 x. _i = 0 ` .  (Contributed by SN,
+       30-Jun-2024.) $)
+    0tie0 $p |- ( 0 x. _i ) = 0 $=
```
avekens, line 640008: Why is this proof so long? The theorem could easliy be proven using ~mul02 and ~ax-icn. I think the usage of some axioms should be avoided here. Since this is a part of/lemma for ~sn-mul02, it should be named ~sn-mul02lem.
The label ~0tie0 should be used/reserved for the direct proof using ~mul02 and ~ax-icn.

icecream17: Since ~ 0tie0 is also used in ~ retinre (now ~ retire) i ended up renaming this to ~ sn-0tie0, a la ~ sn-it0e0

avekens (approved):

wlammen (approved):

PR #4072, opened 2024-07-04
Revise ficardun and related theorems

BTernaryTau:
This change revises nnadju, ficardun, and ficardun2 so that they no longer depend on ax-rep. It also introduces ficardadju as a new theorem.

> The disjoint union of finite sets is equinumerous to the ordinal sum of the cardinalities of those sets.

I kept the current nnadju proof around as nnadjuALT since it's much shorter and uses a different approach, though I'm not sure how strong the requirements are for justifying keeping an alternate proof.

The revision to ficardun in particular affects a lot of other theorems through hashun, including hashbc (Metamath 100 proof #&#8203;58).

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16172,6 +16172,7 @@ New usage of "fh2" is discouraged (3 uses).
 New usage of "fh2i" is discouraged (1 uses).
 New usage of "fh3i" is discouraged (1 uses).
 New usage of "fh4i" is discouraged (0 uses).
+New usage of "ficardun2OLD" is discouraged (0 uses).
 New usage of "ficardunOLD" is discouraged (0 uses).
 New usage of "fimadmfoALT" is discouraged (0 uses).
 New usage of "findOLD" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


set.mm
```diff
@@ -102838,19 +102838,67 @@ Because we use a disjoint union for cardinal addition (as explained in the
...
     BOEFABGZOEABHZPIJQOEABKABOOLPQMN $.
 
-  $( The cardinal and ordinal sums of finite ordinals are equal.  (Contributed
-     by Paul Chapman, 11-Apr-2009.)  (Revised by Mario Carneiro,
-     6-Feb-2013.) $)
-  nnadju $p |- ( ( A e. _om /\ B e. _om ) ->
+  ${
+    $d B x $.  $d A x y $.
+    $( The cardinal and ordinal sums of finite ordinals are equal.  For a
+       shorter proof using ~ ax-rep , see ~ nnadjuALT .  (Contributed by Paul
```
jkingdon, line 102844: Since the idea of this `nnadju` proof is to avoid ax-rep , I'd add

```
$( $j usage 'nnadju' avoids 'ax-rep'; $)
```

which will make the automated checks fail if the proof starts using ax-rep for whatever reason. Looks like we have been putting this after the proof but before the `$}`

BTernaryTau: I didn't know that was a thing, thanks!

jkingdon:
> I kept the current nnadju proof around as nnadjuALT since it's much shorter and uses a different approach, though I'm not sure how strong the requirements are for justifying keeping an alternate proof.

Shorter proof and different approach generally would be a reason to have an ALT proof.

If someone who has looked more closely at this case than I have has an opinion, feel free to speak up, but seems like a case where it seems like a good idea.

jkingdon (approved):

wlammen (approved):

icecream17 (approved):

tirix (approved):

PR #4073, opened 2024-07-06
Replace syl6reqr with eqtr4id

jkingdon:
They are the same except for the naming of the variables and order of hypotheses.

Here's what I did:

1. Add eqtr4id with the desired naming of variables and order of hypotheses.
2. edit local copy of scripts/minimize to use minimize_with eqtr4id/may_grow in place of the normal minimize
3. Run "show usage syl6reqr" and copy-paste the output to a shell command like for i in x y z; do scripts/minimize $i set.mm end
4. Look for theorems which had been marked "Proof modification is discouraged" (If so, add /OVERRIDE on PROVE and SAVE NEW_PROOF in the script, but check that the change from syl6reqr to eqtr4id was the only modification printed in the output) (step 4 was not needed this time, but recording it here for future reference)
5. Remove syl6reqr which is now unused
6. Repeat for iset.mm
7. update changes-set.txt

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -98,12 +98,12 @@ proposed  syl6rbb   bitr2di     compare to bitr2i or bitr2d
 proposed  syl6bbr   bitr4di     compare to bitr4i or bitr4d
 proposed  syl6rbbr  bitr4id     hypotheses would be switched
 proposed  syl6eq    eqtrdi      compare to eqtri or eqtrd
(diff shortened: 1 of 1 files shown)


icecream17 (approved):

jkingdon:
As usual for these renames, will merge promptly on the idea that will reduce merge conflicts.

PR #4074, opened 2024-07-06
Rename syl6eq to eqtrdi

jkingdon:
This is in set.mm and iset.mm


tirix (approved):
I did not check all 5000 lines, just blindly trusting Jim here.

GinoGiotto (approved):
Checked by creating a copy with VS Code search and replace feature. The diff was zero.

tirix:
That's much better!

wlammen:
I checked all lines by shortly glancing at it, to see whether something unexpected was pushed.

This pull request could have been split in two halves, by moving syl6eqr unchanged to the end of main, and then perform the replacement in the first half first, and proceed with the rest in a second PR.

wlammen (approved):

jkingdon:
Thanks to all the reviewers.

As for splitting it, I guess maybe I will if there is another one this big. I haven't checked the size of the remaining proposed changes but I suppose some of them are probably also large.

PR #4078, opened 2024-07-09
Intuionize logb through logbgt0b

jkingdon:
Given that `log` is only defined for positive reals in iset.mm, it is pretty clear how `logb` will work - the base must be a positive real apart from one and the logarithm to that base is defined on positive reals. The graphical illustration at https://en.wikipedia.org/wiki/Logarithm#/media/File:Log4.svg shows what the function looks like with a base less than one or a base greater than one.

Pretty much everything in the section intuitionizes given the restriction in domain described above. This is all of this section except for some theorems about irrationality which I will leave for later..

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -12092,6 +12092,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>logb1</td>
+  <td>~ rplogb1</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
   <td>~ apdiff</td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -12127,6 +12127,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>relogbmulexp</td>
+  <td>~ rprelogbmulexp</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
   <td>~ apdiff</td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -12171,6 +12171,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>cxplogb</td>
+  <td>~ rpcxplogb</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
   <td>~ apdiff</td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -12132,6 +12132,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>relogbdiv</td>
+  <td>~ rprelogbdiv</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
   <td>~ apdiff</td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -12097,6 +12097,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>elogb</td>
+  <td>~ rpelogb</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
   <td>~ apdiff</td>
 </tr>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -12137,6 +12137,11 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>relogbexp</td>
+  <td>~ relogbexpap</td>
+</tr>
+
+<tr>
   <td>irrdiff</td>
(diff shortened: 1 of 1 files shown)


tirix (approved):

icecream17 (approved):

iset.mm
```diff
@@ -148169,6 +148242,320 @@ mathematician Ptolemy (Claudius Ptolemaeus).  This particular version is
...
+    $d x y $.
+    $( Define the ` logb ` operator.  This is the logarithm generalized to an
+       arbitrary base.  It can be used as ` ( B logb X ) ` for "log base B of
+       X".  In the most common traditional notation, base B is a subscript of
+       "log".  The definition will only be useful where ` x ` is a positive
+       real apart from one and where ` y ` is a positive real, so the choice of
+       ` ( CC \ { 0 , 1 } ) ` and ` ( CC \ { 0 } ) ` is somewhat arbitrary (we
+       adopt the definition used in set.mm).  (Contributed by David A. Wheeler,
+       21-Jan-2017.) $)
+    df-logb $a |- logb = ( x e. ( CC \ { 0 , 1 } ) , y e. ( CC \ { 0 } ) |->
```
avekens, line 148283: to be consistent, `x` should be replaced by `b`, and `y` by `x `.

iset.mm
```diff
@@ -148169,6 +148242,320 @@ mathematician Ptolemy (Claudius Ptolemaeus).  This particular version is
...
+        -> ( B logb X ) = ( ( log ` X ) / ( log ` B ) ) ) $=
+    ( c2 cuz cfv wcel crp wa c1 cap wbr clogb co clog cdiv wceq cc0 wne adantr
+    cz zgt1rpn0n1 simp1d simp3d wb eluzelz zapne syl2anc mpbird simpr rplogbval
+    1zzd syl3anc ) ACDEFZBGFZHAGFZAIJKZUNABLMBNEANEOMPUMUOUNUMUOAQRZAIRZAUAZUBS
+    UMUPUNUMUPURUMUOUQURUSUCUMATFITFUPURUDCAUEUMUKAIUFUGUHSUMUNUIABUJUL $.
+
+  $( Closure of the general logarithm with a positive real base on positive
+     reals.  (Contributed by Stefan O'Rear, 19-Sep-2014.)  (Revised by Thierry
+     Arnoux, 27-Sep-2017.) $)
+  relogbclap $p |- ( ( B e. RR+ /\ X e. RR+ /\ B =//= 1 )
```
avekens, line 148367: This is the same as ~ rplogbcl, isn't it?

iset.mm
```diff
@@ -148169,6 +148242,320 @@ mathematician Ptolemy (Claudius Ptolemaeus).  This particular version is
...
+    eqtrd cmul syl3anc ccxp zred logcxpd cn cxpexpnn fveq2d eqtr3d oveq1d recnd
+    eluz2nn relogcld logrpap0d divcanap4d 3eqtr2d wo wn wdc zdceq mpan2 exmiddc
+    0z syl df-ne orbi2i sylibr adantl mpjaodan ) ACUADEZBFEZGZBHIZAABJKZLKZBIBH
+    MZWOWPGZANLKZHWRBWTAOEZANUBUCZXAHIWOXBWPWOXBAHMZANMZWMXBXDXEUDWNAUEZSUFZSWM
+    XCWNWPWMXCXEWMXBXDXEXFUGWMAFENFEXCXEUHCAUIWMUJANULPUKZQAUMPWTWQNALWTWQAHJKN
+    WTBHAJWOWPUNZUOWTAWOAUPEWPWOAXGUQSURVCUOXIUSWOWSGZWRWQRDZARDZTKZBXLVDKZXLTK
+    BXJXBXCWQOEWRXMIWOXBWSXGSZWMXCWNWSXHQZXJABXOWMWNWSUTZVAAWQVBVEXJXNXKXLTXJAB
+    VFKZRDXNXKXJABXOXJBXQVGZVHXJXRWQRXJAVIEZWNXRWQIWMXTWNWSAVOQXQABVJPVKVLVMXJB
+    XLXJBXSVNXJXLXJAXOVPVNXJAXOXPVQVRVSWNWPWSVTZWMWNWPWPWAZVTZYAWNWPWBZYCWNHFEY
+    DWFBHWCWDWPWEWGWSYBWPBHWHWIWJWKWL $.
```
avekens, line 148478: I think this proof can be shortened using ~relogbexpap .

iset.mm
jkingdon: So it is. Will fix.

iset.mm
jkingdon: Done

iset.mm
jkingdon: I wouldn't object if someone wants to do that (in set.mm and iset.mm, presumably), but there's definitely precedent for using x and y in definitions even if we use other letters elsewhere.

jkingdon:
I'm going to go ahead and merge this since I made 2 of the 3 suggested changes and we can always make the other one later if we decide we want to do that.

PR #4080, opened 2024-07-11
The induced map on prime spectra is continuous

tirix:
In this PR, actually only ~ rhmpreimacnlem is required to prove ~ rhmpreimacn.
The other theorems are interesting but unrelated.


avekens (approved):
Very interesting result!

wlammen:
These are changes in TA's mathbox, nothing more.  In my eyes this can be merged now.

PR #4085, opened 2024-07-16
Intuitionize the rest of the section "Complex products"

jkingdon:
There's no big mystery here, because this is mostly well-trodden with sums (in general terms at least - some details are different between products and sums).

There is, however, a fair bit of making a variety of changes to some relatively large proofs (while keeping the overall structure unchanged).

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -113,6 +113,12 @@
 "bj-axemptylem" is used by "bj-axempty2".
 "cnm" is used by "axaddf".
 "cnm" is used by "axmulf".
+"csbco" is used by "csbcow".
+"csbco" is used by "csbnest1g".
+"csbco" is used by "csbvarg".
+"csbco" is used by "fsum3".
+"csbco" is used by "fsumsplitf".
+"csbco" is used by "zsumdc".
 "df-div" is used by "divfnzn".
 "df-div" is used by "divvalap".
 "df-ilim" is used by "dflim2".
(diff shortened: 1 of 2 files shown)


iset.mm
```diff
@@ -26573,14 +26573,25 @@ practical reasons (to avoid having to prove sethood of ` A ` in every use
...
   $}
 
+  ${
+    $d z A $.  $d x y z $.  $d y z B $.
+    $( Composition law for chained substitutions into a class.  Version of
+       ~ csbco with a disjoint variable condition.  (Contributed by NM,
+       10-Nov-2005.)  (Revised by Gino Giotto, 10-Jan-2024.) $)
+    csbcow $p |- [_ A / y ]_ [_ y / x ]_ B = [_ A / x ]_ B $=
+      ( csbco ) ABCDE $.
+  $}
```
avekens, line 26593: This makes no sense if the proof uses ~csbco. In set.mm, the proof of ~csbcow does not depend on ~csbco, and therefore requires less axioms.

jkingdon: The reasons are:

1. Quite aside from how it is proved, `csbcow` in this pull request does have the distinct variable constraint that `csbco` does not. This ensures that proofs using it do not violate that constraint.
2. We need to have `csbcow` to follow the guideline "Theorems which are the same as in set.mm should be named the same" from https://us.metamath.org/ileuni/conventions.html . My experience in intuitionizing is that this guideline is quite important. It greatly increases the number of proofs which I can copy-paste between iset.mm and set.mm without change, and even when the entire proof doesn't work, it is very helpful if no changes are needed to the parts of the proof which can work as-is.
3. I expect the `ax-13` removal to come to iset.mm some day, since as far as I can tell the issues with ax-13 in set.mm are only slightly different from the issues with https://us.metamath.org/ileuni/ax-bndl.html in iset.mm. Although we don't yet have a volunteer to pursue this on a larger scale, even incremental steps in the direction of making proofs avoid `csbco` get us a tiny bit closer.

avekens: Then this should be mentioned/indicated in the comment of this theorem, at least.

jkingdon: I've added "Although currently the proof is a direct reference to ~ csbco , we expect that the additional distinct variable condition will eventually enable us to remove usage of ~ ax-bndl here ."

avekens (approved):

icecream17 (approved):

PR #4088, opened 2024-07-20
Axiom Sorry

metakunt:
This adds the sorry axiom and its invocations. This is motivated by lean4, which extensively uses sorry for proof stubbing.
Main motivation of this axiom is to develop and formalise statements with an opaque api, to quickly iterate proofs without hypothesis blowup.

To prove any statement you invoke sorry1,sorry2,sorry3,sorry4.

We also add opaque definitions for finite fields, which operate only on the properties of their elements and are irrespective of its construction. As of now the definitions are in @digama0 mathbox. The api there is not yet developed and the task of developing an api is not particularily fun, as that may lead to a rabbit hole. My idea is to work backwards. Develop the api first and see if the statements are usable for the formalisation of AKS, if that is successful construct the mathematical objects and prove the statements in the api.

We also give a simple Frobenius endomorphism property for the algebraic closure of the finite field. Here the definition is not yet fixed. We likely need to develop the homomorphism $ a^n = a\cdot\dots\cdot a$ to define exponentiation on any multiplicative monoid. Here the question is the domain of the homomorphism. 
```math
\mathbb{N} \to \text{multiplicative semigroup}
```
```math
\mathbb{N}_0 \to \text{multiplicative monoid}
```
```math
\mathbb{Z} \to \text{multiplicative group}
```

As far as I have checked the theory seems not yet developed. I have found additive homomorphisms but not one that respects the multiplicative structure.

This references ongoing work https://github.com/tirix/metamath-blueprints/pull/7 and https://github.com/tirix/metamath-blueprints/issues/8

Some statements would benefit from ax-sorry, examples being 
https://us.metamath.org/mpeuni/ax-hgprmladder.html
https://us.metamath.org/mpeuni/ax-bgbltosilva.html
https://us.metamath.org/mpeuni/ax-tgoldbachgt.html
https://us.metamath.org/mpeuni/ax-hgt749.html
https://us.metamath.org/mpeuni/ax-ros335.html
https://us.metamath.org/mpeuni/ax-ros336.html


wlammen:
What is the reason to not simply introduce
`ax-sorry $a |- ph $.` ?

metakunt:
That's simple. If you try it in mmj2 every statement is provable, so it would autocomplete to ax-sorry, which is undesirable.

I thought 4 statements are enough to block any autocompletes. The triple negation is rare and a combination of triple negation as antecedent with an conjunction as a consequent should be so rare that it never autocompletes, yet it is enough to easily prove a false statement in its desired form, namely any $\varphi$.

It is meant that if any statement is proven via ax-sorry that it only involves the four steps sorry1 to sorry4 in that order with no other statements. This gives an easy template to quickly generate statements with stubbable proofs.

My hope is helps @tirix and me to develop an API more interactively, as we can focus on small blocks, figure out the minimal amount of good API theorems to prove fundamental result of the theory and then develop and construct the mathematical objects and connect them to other objects.

I assume we will need to develop some theory, as my venture in lean shows that we need good understanding of irreducibles over finite fields, direct limits, etc... But my hope is that a high level api is enough to prove the statements.

For example the proof of https://us.metamath.org/mpeuni/itgparts.html uses a huge amount of theory and the internals leak sometimes into the proof. That way some statements are in such a general form, this is not a clean API, this is a leaky abstraction.

But before we develop the theory I'd like to figure out which statements to collect for a clean api. Also I am not yet sure if the api with the extensible structures is the best idea.

We certainly will need the generalisation of exponentiation for arbitrary multiplicative structures anyways to formulate frobenius. Most of the work will be working in $\mathbb{Z}[X]$ anyways and all other results should lift to the desired algebra, either $\mathbb{F}_p[X]$, $\mathbb{F}_q$ or the algebraic closure.

digama0:
I don't think this is a good idea. Every proof assistant ecosystem has to deal with how to expose a "sorry"-like system, for writing incomplete proofs, but this is largely an infrastructural question, and every piece of the system has to be on the same page about how it works.

Metamath already chose its approach a while ago: there is no literal `ax-sorry` axiom, but instead you can put `?` in proofs and it will not break the parser. (In MM-PA you can even have multiple `?` in a single proof, and this is used to store proofs in progress, but mmj2 generally sticks to just one `?` at the top level of the proof.) As you and Wolf point out, if you actually add `ax-sorry` you run into troubles because the tooling is smart and uses it to prove everything. In Lean, the tooling has been built from the beginning to treat `sorryAx` as special, and it will produce warnings if you use it and library search tools will automatically avoid it.

Note though that set.mm itself does not allow proofs using `?` at all, just as mathlib does not allow proofs using `sorry`. If you have an incomplete proof, keep it on your machine, or comment it out completely before PR'ing it.

tirix:
As far as Frobenius is concerned, you can check [`~freshmansdream`](https://us.metamath.org/mpeuni/freshmansdream.html), and [`~frobrhm`](https://us.metamath.org/mpeuni/frobrhm.html), which respectively correspond to the statement in this PR, and to the fact that the Frobenius endomorphism indeed is an endomorphism.

metakunt:
Ah I see, the idea was mostly to do mathbox like collaborations. In particular I wanted to work top to bottom as the api for finite fields in not developed.

Do you have any clue what I would have to develop to have a basic theory of finite fields suitable for proving the [ Agrawal–Kayal–Saxena algorithm](https://www3.nd.edu/~andyp/notes/AKS.pdf)

In particular there seem to be the following statements that are of interest.
Existence of algebra homomorphisms that map
```math
\mathbb{Z}_n[X]/(X^r-1) \to algcl(\mathbb{F}_p)
```
the class [x] to its $r$-th unity.

If you know which statements make sense to formalise we could work on extending finite field theory. In particular, what do I need to develop to state and prove the following theorem.

Let $n\in \mathbb{N}$, $n\geq 3$, p a prime natural number that divides n. Let $a\in A$, where $1\leq a \leq a_0$ for some natural number $a_0$. Let r be a natural number such that
```math
(x+a)^n \equiv x^n+a \mod (x^r-1)
```
for all a.
Let $\phi$ be a ring homomorphism from $\mathbb{Z}_n[X]/(X^r-1)$ to $algcl(\mathbb{F}_p)$, taking $\mathbb{Z}_n$ to $\mathbb{F}_p$ and mapping $x$ to a $r$-th primitive unity of $algcl(\mathbb{F}_p)$, called $\mu$.

Then in the algebraic closure the following elements are equal
```math
(\mu-a)^n = \mu^n-a
```

digama0:
I think for longer term collaborations, if you don't want to use commented theorems, you should use axioms, but not a blanket axiom, just an $a in place of the $p theorem you want to assume itself.

metakunt:
Alright, that works for me.

PR #4089, opened 2024-07-21
wl-ifp-ncond..

wlammen:
Richard Penner's mathbox has a lot of material around the conditional operator.  These are somehow missing.

1. Mathbox: add wl-ifp-ncond{1,2}: If one branch in an if-expression does not hold, the other automatically does.
2. Delete outdated OLD theorems.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -12611,7 +12611,6 @@
 "sbimvOLD" is used by "sbanvOLD".
 "sbimvOLD" is used by "sbbivOLD".
 "sblbisALT" is used by "sbieALT".
-"sblbisvOLD" is used by "sbievOLD".
 "sbnALT" is used by "sbanALT".
 "sbnALT" is used by "sbi2ALT".
 "sbrimALT" is used by "sbiedALT".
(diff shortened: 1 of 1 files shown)


tirix (approved):

set.mm
```diff
@@ -22511,10 +22483,10 @@ proposition with a distinct variable (closed form of ~ nfsb4 ).  Usage of
 
   ${
     $d x y $.
-    $( Obsolete version of ~ sbal as of 13-Jul-2023.  A theorem used in
-       elimination of disjoint variable restriction on ` x ` and ` y ` by
-       replacing it with a distinctor ` -. A. x x = z ` .  (Contributed by NM,
-       15-May-1993.)  (Proof shortened by Wolf Lammen, 3-Oct-2018.)
+    $( Check out ~ sbal for a version not dependent on ~ ax-13 .  A theorem
+       used in elimination of disjoint variable restriction on ` x ` and ` y `
```
icecream17, line 22487:
```suggestion
       used in elimination of disjoint variable restriction on ` x ` and ` z `
```

presumably a typo that carried over

icecream17 (approved):
Note: Only WL's mathbox is changed

PR #4095, opened 2024-07-22
Inequality involving fundamental theorem of calculus

metakunt:
Useful lemma to do some analysis


icecream17 (approved):

set.mm
```diff
@@ -637098,6 +637098,72 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+  $}
+
+  ${
+    $d A k $.  $d N k $.  $d k ph $.
+    aks4d1p1p1.1 $e |- ( ph -> A e. RR+ ) $.
+    aks4d1p1p1.2 $e |- ( ph -> N e. NN ) $.
+    $( Exponential law for finite products, special case.  (Contributed by
+       metakunt, 22-Jul-2024.) $)
+    aks4d1p1p1 $p |- ( ph -> prod_ k e. ( 1 ... N ) ( A ^c k ) =
+     ( A ^c sum_ k e. ( 1 ... N ) k ) ) $=
```
tirix, line 637156: I believe this time, this is generic enough to get a name without the `aks`, but I'm unsure what name to suggest.

tirix (approved):

set.mm
metakunt: Yeah, if someone needs it and it is moved outside of the mathbox. I agree.

PR #4097, opened 2024-07-23
prove mhpsclcl, mhpmulcl (, offun)

icecream17:
also shorten mpofun

offun used to auto shorten theorems using both offn and fnfun:
+ frlmsslsp
+ lcomfsupp
+ mndpfsupp
+ mndpsuppss
+ psrbagev1

ffund ended up removing its usage in mhpmulcl

(TODO, future pr: shorten mhpvarcl using ismhp3)

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -17270,6 +17270,7 @@ New usage of "mobiiOLD" is discouraged (0 uses).
 New usage of "moexex" is discouraged (2 uses).
 New usage of "moexexv" is discouraged (0 uses).
 New usage of "mpjao3danOLD" is discouraged (0 uses).
+New usage of "mpofunOLD" is discouraged (0 uses).
 New usage of "mpteq12dvOLD" is discouraged (0 uses).
 New usage of "mptresidOLD" is discouraged (0 uses).
 New usage of "mptssALT" is discouraged (0 uses).
@@ -19975,6 +19976,7 @@ Proof modification of "mo4OLD" is discouraged (9 steps).
 Proof modification of "mobidvALT" is discouraged (48 steps).
 Proof modification of "mobiiOLD" is discouraged (17 steps).
 Proof modification of "mpjao3danOLD" is discouraged (29 steps).
+Proof modification of "mpofunOLD" is discouraged (95 steps).
 Proof modification of "mpteq12dvOLD" is discouraged (18 steps).
 Proof modification of "mptresidOLD" is discouraged (28 steps).
 Proof modification of "mptssALT" is discouraged (57 steps).


tirix (approved):

set.mm
```diff
@@ -268465,7 +268503,11 @@ Additional definitions for (multivariate) polynomials
     mhp0cl.i $e |- ( ph -> I e. V ) $.
     mhp0cl.r $e |- ( ph -> R e. Grp ) $.
     mhp0cl.n $e |- ( ph -> N e. NN0 ) $.
-    $( The zero polynomial is homogeneous.  (Contributed by SN,
+    $( The zero polynomial is homogeneous.  Under the current definition, it
```
benjub, line 268506: Can you link to the "current definition" ?

set.mm
```diff
@@ -268465,7 +268503,11 @@ Additional definitions for (multivariate) polynomials
...
+       has any (nonnegative integer) degree which loosely corresponds to the
+       value "undefined".  The values ` -oo ` and ` 0 ` are also used in
+       Metamath (by ~ df-mdeg and ~ df-dgr respectively) and the literature:
```
benjub, line 268509: Zero sounds like a strange choice. Could we modify df-dgr by simply replacing NN with ZZ ?

set.mm
```diff
@@ -268475,6 +268517,35 @@ Additional definitions for (multivariate) polynomials
...
+    $d ph d y $.  $d C d y $.  $d A d $.  $d R d h y $.  $d I d h y $.
+    $d V k $.  $d I k $.  $d V y $.  $d K y $.
+    mhpsclcl.h $e |- H = ( I mHomP R ) $.
+    mhpsclcl.p $e |- P = ( I mPoly R ) $.
+    mhpsclcl.a $e |- A = ( algSc ` P ) $.
+    mhpsclcl.k $e |- K = ( Base ` R ) $.
+    mhpsclcl.i $e |- ( ph -> I e. V ) $.
+    mhpsclcl.r $e |- ( ph -> R e. Ring ) $.
+    mhpsclcl.c $e |- ( ph -> C e. K ) $.
+    $( A scalar (or constant) polynomial has degree 0.  Compare ~ deg1scl .
```
benjub, line 268530: A **nonzero** constant polynomial...

set.mm
```diff
@@ -268503,6 +268574,96 @@ Additional definitions for (multivariate) polynomials
...
+    mhpmulcl.y $e |- Y = ( I mPoly R ) $.
+    mhpmulcl.t $e |- .x. = ( .r ` Y ) $.
+    mhpmulcl.i $e |- ( ph -> I e. V ) $.
+    mhpmulcl.r $e |- ( ph -> R e. Ring ) $.
+    mhpmulcl.m $e |- ( ph -> M e. NN0 ) $.
+    mhpmulcl.n $e |- ( ph -> N e. NN0 ) $.
+    mhpmulcl.p $e |- ( ph -> P e. ( H ` M ) ) $.
+    mhpmulcl.q $e |- ( ph -> Q e. ( H ` N ) ) $.
+    $( A product of homogeneous polynomials is a homogeneous polynomial whose
+       degree is the sum of the degrees of the factor polynomials.  Compare
```
benjub, line 268590: ... of the factors.

("factor" is a noun, not an adjective)

benjub (commented):
Belated suggestions, @icecream17

set.mm
icecream17: the zero polynomial also has degree zero, it just also has all other degrees

set.mm
icecream17: df-sup simplifies to `U. { x e. ZZ | A. y e. ZZ -. y < x }` when `A = (/), B = ZZ, R = <` (`A` is the set of coefficients which is `(/)` for a zero polynomial)

so the degree would become `(/)` since unlike NN0, ZZ does not have a minimum

set.mm
benjub: Yes, and indeed you wrote that above. But as you wrote, this is due to our special convention: a reader who lands on this page might be a bit puzzled. So that first sentence without "nonzero" sounds a bit strange, even if our special definition makes it work also for the zero polynomial.

set.mm
benjub: Oh, I didn't check the set.mm definitions. Then it's not $\mathbb{Z}$ one should use, but  $\overline{\mathbb{Z}}$, of course.

benjub: Thanks for #4102 ! Resolving.

Issue #3757, opened 2024-01-12
Existence of a certain non-constant function implies the limited principle of omnisience


jkingdon:
This is a bit counterintuitive to me, but certain kinds of functions cannot even be shown to exist in iset.mm. Exercise 11.6(ii) of [HoTT] can be translated into our notation as:

```
ncflpo.f $e |- ( ph -> F : RR --> NN ) $.
ncflpo.0 $e |- ( ph -> ( F ` 0 ) = 0 ) $.
ncflpo.rp $e |- ( ( ph /\ x e. RR+ ) -> ( F ` x ) =/= 0 ) $.
ncflpo $p |- ( ph -> _om e. Omni ) $.
```

Proving this would be a nice companion to https://us.metamath.org/ileuni/trilpo.html .

Perhaps relatedly, does `A. x e. RR A. y e. RR DECID x =//= y` imply `_om e. Omni` ? _Answer: yes, see https://us.metamath.org/ileuni/trilpo.html and https://us.metamath.org/ileuni/triap.html_ Does `A. x e. RR A. y e. RR DECID x =/= y` ? _Answer: no, it implies `_om e. WOmni` as described in https://github.com/metamath/set.mm/issues/4041_

benjub:
If `A. x e. RR DECID x =//= 0`, then you can construct a function as in the hypothesis of `ncflop`: take it to be 1 if x is apart from zero.

benjub:
I think I got it: Let $P \colon \omega \to 2$ be a property. Define the number $x := \sum_n P_n 2^{-n}$. You can decide whether the integer $f(x)$ is zero or not. But this amounts to P being everywhere zero or not.

digama0:
> Does `A. x e. RR A. y e. RR DECID x =/= y` ?

This one also follows similar to @benjub 's first answer: take `f(x)` to be `1` if `x != 0`. Then `f(0) = 0` (because `-. ( 0 =/= 0 )`) and `f(x) =/= 0` if `x > 0` because this implies `x != 0`.

jkingdon:
> Perhaps relatedly, does `A. x e. RR A. y e. RR DECID x =//= y` imply `_om e. Omni` ?

I just now noticed this is already proven - https://us.metamath.org/ileuni/triap.html plus https://us.metamath.org/ileuni/trilpo.html

jkingdon:
[...]

PR #4098, opened 2024-07-24
Strengthen dcapncf from apartness to equality

jkingdon:
This is in iset.mm.

Rename to dceqnconst and change the antecedent from decidable apartness (analytic LPO) to decidable equality for real numbers (analytic WLPO).

There's probably more to do on this topic ( see https://github.com/metamath/set.mm/issues/3757#issuecomment-2229600853 ) but I wanted to start with the low hanging fruit.


tirix (approved):

icecream17 (approved):

benjub:
Hi @jkingdon , a few questions/remarks:
* I think this would have deserved that the OLD version stays a while in order to compare the proofs in a human-readable format. If this is not too much work, it may be useful to put it back, even in a mathbox.
* I don't remember which reals are used in iset.mm (?), but as we both know there are several interesting versions in constructive analysis, some of them included in others. Therefore, it may be interesting to formulate this theorem for arbitrary $A$, typically included in $\mathbb{R}$ and containing 0. This way, the theorem would apply when $A$ is another kind of reals included in $\mathbb{R}$.
* As a general rule, when an existence is proven by constructing a witness (which is always doable, and practically often done, in constructive math), it may be useful to formulate the strong form of the theorem.

Combining the last two remarks, this would give something like:
```
phi -> A. x e. A DECID x = 0
phi -> 0 e. A
phi -> F = ( y e. A |-> if(y = 0, 0, 1) )
=>
phi -> ( F : A --> NN0 /\ ( F(0) = 0 /\ A. z e. ( A \cap RR_+) F(z) \neq 0 ) )
```
(where I strengthened by taking NN0 as codomain and where I avoided the triple conjunction by using a natural parenthesization).

jkingdon:
> Hi @jkingdon , a few questions/remarks:
> 
> I think this would have deserved that the OLD version stays a while in order to compare the proofs in a human-readable format. If this is not too much work, it may be useful to put it back, even in a mathbox.

OK with me. I can put this in my next pull request. I'll bring back `dcapncf`, with the previous proof. In addition, that pull request will have a few theorems which say that analytic LPO implies analytic WLPO so I'll add a bit of text (or, probably, an ALT proof) saying that would be another way to prove `dcapncf`.

In the meantime one way to see the previous proof is to use git to get an older iset.mm and run `SHOW STATEMENT dcapncf/ALT_HTML` using a local copy of metamath.exe.

> I don't remember which reals are used in iset.mm (?), but as we both know there are several interesting versions in constructive analysis, some of them included in others. Therefore, it may be interesting to formulate this theorem for arbitrary A, typically included in R and containing 0. This way, the theorem would apply when A is another kind of reals included in R.

My next pull request will add the following paragraph to the section header which describes the iset.mm construction of the reals:

> When working without excluded middle, there are several possible definitions
of real numbers.  Here we adopt the most common definition, as two-sided
Dedekind cuts with the properties described at ~ df-inp .  The Cauchy reals
(without countable choice) fail to satisfy ~ ax-caucvg and the MacNeille
reals fail to satisfy ~ axltwlin , and we do not develop them
here.  For more on differing definitions of the reals, see the
introduction to Chapter 11 in [HoTT] or section 1.2 of [BauerHanson].

In addition to the two sources cited there, there's a really nice exposition at https://mathoverflow.net/a/453340/489586 by Andrej Bauer.

As for how/whether to develop the other reals further in iset.mm, I'm not too sure. Working with a subset of the Dedekind reals immediately raises the question of what closure properties the subset has, and the MacNeille reals are a superset.

> As a general rule, when an existence is proven by constructing a witness (which is always doable, and practically often done, in constructive math), it may be useful to formulate the strong form of the theorem.

We do this in some cases, like https://us.metamath.org/ileuni/ctssdccl.html . I guess my instinct is to do so where we have a concrete need to use it in a proof.

It definitely is a good technique to keep in mind, as you do correctly point out that sometimes we need the witness and not just existence.

benjub:
Thanks for restoring the other proof and for your answer.  As for the comments to add to the database itself, I'm not sure how much of standard math should be repeated in the database (but your proposal above looks nice and the references are certainly welcome).  I guess that simply writing in https://us.metamath.org/ileuni/df-r.html "Define the set of Dedekind reals." instead of simply "Define the set of reals." would already help a lot.  (Actually, that's the first page I checked while writing the above post, and I didn't go further.)

So, the proposal to make the theorem more generic with a class `A` as defined above still stands, especially since the $\mathbb{R}_{Dedek}$ of the database is among the larger kinds of reals.  Hopefully, one day, some version of the Cauchy reals or the HOTT reals will also be studied in the database !

jkingdon:
> I'm not sure how much of standard math should be repeated in the database (but your proposal above looks nice and the references are certainly welcome)

It is a good caveat. As a general matter I don't think we do a very good job when we try to go too far beyond the actual theorems we prove.

I guess if this text looks good we can keep it, though (basically, if it looks like a brief literature reference more so than "repeat standard math without formalized proofs").

I hope to make a pull request tonight with the material I have discussed. Timing will depend on how quickly I make progress on finishing the one proof which  is needed to complete this pull request.

jkingdon:
> I hope to make a pull request tonight with the material I have discussed.

In case it isn't obvious, that was a reference to what I submitted as https://github.com/metamath/set.mm/pull/4100

Issue #4058, opened 2024-06-22
Lesser limited principle of omniscience


jkingdon:
This is one of the usual constructive principles between nothing and excluded middle (excluded middle is known as the principle of omniscience because of https://us.metamath.org/ileuni/exmidomni.html ).

We can define/prove:

1. ` A e. LOmni <-> ∀𝑓 ∈ (2o ↑𝑚 𝐴) ∀g ∈ (2o ↑𝑚 𝐴)  ( -. 1o e. ( ran f i^i ran g ) -> ( -. 1o e. ran f \/ -. 1o e. ran g ) ) ` (see "Stated set-theoretically, the lesser limited principle of omniscience for A" at https://ncatlab.org/nlab/show/principle+of+omniscience#the_lesser_limited_principle_of_omniscience ).
2. analogue to https://us.metamath.org/ileuni/enomni.html for `LOmni`
3.  ``NN e. LOmni <-> A. f e. ( { 0 , 1 } ^m NN ) ( E* n e. NN ( f ` n ) = 1 -> A. n e. NN ( f ` ( 2 x. n ) ) = 0 \/ A. n e. NN ( f ` ( ( 2 x. n ) + 1 ) ) = 0 )`` . This is from https://awswan.github.io/teaching/intuitionisticlogic/lecturenotes/section3.pdf or https://plato.stanford.edu/entries/mathematics-constructive/ (search for "LLPO")
4. `A. x e. RR A. y e. RR ( x <_ y \/ y <_ x ) -> _om e. LOmni` as described at https://ncatlab.org/nlab/show/principle+of+omniscience#analytic
5. `A e. Omni -> A e. LOmni` (mentioned at https://ncatlab.org/nlab/show/principle+of+omniscience#analytic at least for A = ω but I suppose maybe true for any A)

PR #4100, opened 2024-07-28
results involving LPO, WLPO, analytic LPO, and analytic WLPO

jkingdon:
There are a few small results which are in support of the following but basically this is:

* changes discussed in https://github.com/metamath/set.mm/pull/4098#issuecomment-2254188340 regarding (a) our real number definition, and (b) restoring of a weaker version of https://us.metamath.org/ileuni/dceqnconst.html and including an ALT proof which shows how it is weaker
* formalize proof from https://github.com/metamath/set.mm/issues/3757#issuecomment-2229600853 which is #3757 except for the issue of LPO versus WLPO

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -325,6 +325,7 @@ New usage of "cnm" is discouraged (2 uses).
 New usage of "condcOLD" is discouraged (0 uses).
 New usage of "conventions" is discouraged (0 uses).
 New usage of "csbco" is discouraged (6 uses).
+New usage of "dcapnconstALT" is discouraged (0 uses).
 New usage of "dcnnOLD" is discouraged (0 uses).
 New usage of "demoivreALT" is discouraged (0 uses).
 New usage of "df-div" is discouraged (2 uses).
@@ -476,6 +477,7 @@ Proof modification of "bj-vprc" is discouraged (53 steps).
 Proof modification of "bj-zfpair2" is discouraged (80 steps).
 Proof modification of "brab2ga" is discouraged (79 steps).
 Proof modification of "condcOLD" is discouraged (39 steps).
+Proof modification of "dcapnconstALT" is discouraged (83 steps).
 Proof modification of "dcnnOLD" is discouraged (33 steps).
 Proof modification of "demoivreALT" is discouraged (1087 steps).
 Proof modification of "difidALT" is discouraged (20 steps).


iset.mm
```diff
@@ -70783,6 +70789,15 @@ to define the cut as the lower set only (as Metamath Proof Explorer does),
 but we define the cut as a pair of both the lower and upper sets, as [HoTT]
 does. There are also differences in how we handle order and replacing
 "not equal to zero" with "apart from zero".
+
+When working without excluded middle, there are several possible definitions
```
benjub, line 70793: Should we add "nor the axiom of choice" or "nor choice principles" or "nor the axiom of choice or its weaker forms" ? Or not ? You be the judge.

iset.mm
```diff
@@ -155556,6 +155607,53 @@ WLPO is known to not be provable in IZF (and most constructive
...
   $}
 
+  ${
+    $d x y $.
+    $( Real trichotomy implies decidability of real number equality.  Or in
+       other words, analytic LPO implies analytic WLPO (see ~ trilpo and
+       ~ redcwlpo ).  Thus, this is an analytic analogue to ~ lpowlpo .
+       (Contributed by Jim Kingdon, 24-Jul-2024.) $)
+    tridceq $p |- ( A. x e. RR A. y e. RR ( x < y \/ x = y \/ y < x )
+        -> A. x e. RR A. y e. RR DECID x = y ) $=
```
benjub, line 155617: I'm wondering if in all these theorems, by using symmetry by translation, we can weaken the hypothesis/antecedent to `A. x e. RR ( x < 0 \/ x = 0 \/ 0 < x )` ?

iset.mm
```diff
@@ -155556,6 +155607,53 @@ WLPO is known to not be provable in IZF (and most constructive
...
+      AMUSURUMVAUKULNOPUKULTZUNVBQZVAUQVBUNUAZULUKUBUKULUCZUDRUNUQMUTUNVCUQUNVB
+      UEVESUFUTUOVBUQUSUOVBMURUSUOVBULUKNOUGVBVCUQVDVESRUHUIUJ $.
+  $}
+
+  ${
+    $d x y z $.
+    $( Two ways to express decidability of real number equality.  (Contributed
+       by Jim Kingdon, 23-Jul-2024.) $)
+    redc0 $p |- ( A. x e. RR A. y e. RR DECID x = y <->
+        A. z e. RR DECID z = 0 ) $=
```
benjub, line 155630: I guess this statement confirms the above possibility.

benjub (approved):
Nice results !

iset.mm
jkingdon: My initial draft was just "When working constructively" - maybe that is better. It hardly seems like the place to go deeply into this topic - the HoTT book mentions both countable choice and power sets in this context, and the BauerHanson paper has a little bit of this sort of thing too.

iset.mm
jkingdon: Yes, although `redc0` is for analytic WLPO. `reap0` is the one for analytic LPO which is what you had asked about.

iset.mm
jkingdon: Yes, that would work, and the pull request has it, stated in a slightly different form, as `reap0`.

iset.mm
benjub: "When working constructively" is probably better indeed.

iset.mm
benjub: Perfect. What I mean is that this theorem is probably better stated as
```
    tridceq $p |- ( A. x e. RR ( x < 0 \/ x = 0 \/ 0 < x )
        -> A. x e. RR A. y e. RR DECID x = y ) $=
```
since the antecedent is both simpler and a priori weaker (it only turns out to be equivalent after a bit of thought).

iset.mm
jkingdon: Changed.

benjub (approved):

iset.mm
jkingdon: Since the various choices here are equivalent (and in many cases we have proved them to be equivalent), this is for the most part a potentially hard to resolve style question.

At the risk of getting more into notation topics that I might prefer, I'll list what we have now:

Regular omniscience principles:

* `_om e. Omni` (LPO)
* `_om e. WOmni` (WLPO)
* `_om e. Markov` (MP)
* proposed for LLPO at #4058 

Analytic omniscience principles:

* `∀𝑥 ∈ ℝ ∀𝑦 ∈ ℝ (𝑥 < 𝑦 ∨ 𝑥 = 𝑦 ∨ 𝑦 < 𝑥)` (analytic LPO)
* `∀𝑥 ∈ ℝ ∀𝑦 ∈ ℝ DECID 𝑥 = 𝑦`  (analytic WLPO)
* `A. x e. RR A. y e. RR ( x =/= y -> x =//= y )` (analytic MP)
* `A. x e. RR A. y e. RR ( x <_ y \/ y <_ x )` (analytic LLPO)

jkingdon:
I'll give this a little more time (maybe until tomorrow morning Pacific time) in case anyone else wants to weigh in, but expect to merge it without much further ado.

iset.mm
benjub: That is a nice synthetic view, which may be worth adding somewhere in iset.mm, maybe in some section header comment, with links to related definition for the regular principles, and main theorem for the analytic principles.

I think the difference between the two versions of tridceq in this thread is more than only style (two nested quantifiers is more complicated than one), but this is non-blocking, even less so since it's in a mathbox.

jkingdon: > may be worth adding somewhere in iset.mm, maybe in some section header comment, with links to related definition for the regular principles, and main theorem for the analytic principles.

I've put a version of this in a branch for my next pull request.

Issue #3893, opened 2024-03-25
Hypotheses with an "is a set" property using the universal class _V


avekens:
The conventions for set.mm state that hypotheses of theorems in deduction form with an "is a set" property should not use the universal class `_V`: 

> Notice that this does not hold for hypotheses of theorems in deduction form: Here still ` |- ( ph -> A e. V ) ` should be used rather than ` |- ( ph -> A e. _V ) `.

Currently, there are 173 hypotheses using the universal class `_V` in such cases, e.g., ~euotd. The number of affected theorems is smaller, because some theorems have more than one of these hypotheses.

The question is: should these theorems be modified, replacing `_V` by `V`? If there are more than one of these hypotheses for a theorem, then different class variables (`V`, `W`, `U` etc.) should be used.

The replacement can be performed semiautomatically by using a script (expand the OLD theorem and perform a minimization for each affected theorem afterwards).

I can remember a discussion about this some time ago (which resulted in the above mentioned convention), but I cannot find anything written about it. Does anybody know where to find something about it?

tirix:
I agree with the direction here.

You might think about #944 ?

avekens:
#944 is about '( ? e. _V -> ' , that means `e. _V` is used in the antecedent of an implication, not as consequent (of a hypothesis) as mentioned above.

tirix:
> #944 is about '( ? e. _V -> ' , that means e. _V is used in the antecedent of an implication, not as consequent (of a hypothesis) as mentioned above.

Indeed! 
I don't remember any other such discussion, and could not find any when searching.

jkingdon:
Sounds like a good change to me.  If you are able to get to it, doing this in iset.mm as well would be appreciated. Because iset.mm has https://us.metamath.org/ileuni/elex.html and other theorems relevant to this discussion, the tradeoffs in iset.mm are the same as in set.mm.

avekens:
With PR #4137, this issue is done: there are no hypotheses in the form |- ( ph -> * e. _V ) $. anymore in set.mm (main body). 

There are still some of such hypotheses in the mathboxes: these have to be converted before the corresponding theorems are moved to main.

Furthermore, there are still hypotheses in a similar form, like:

    en2i.3:  ( x e. A -> C e. _V )
    en2i.4: ( y e. B -> D e. _V )
    elfvmptrab.v: ( X e. V -> M e. _V )
[...]

PR #4101, opened 2024-07-28
Revision of hypotheses with an "is a set" property using the universal class _V (Part 3)

avekens:
As discussed in issue https://github.com/metamath/set.mm/issues/3893, hypotheses with an "is a set" property using the universal class _V were revised so that membership in an arbitrary class only is required (according to our conventions). Continuation of #4083.

* revision of fpwwe2.2 and fpwwe.2
* revision of wemaplem2.a and wemapsolem.2: wemaplem2.a and wemapsolem.2 could have been removed entirely, as well as some antecedents ` A e. V `
* revision of ordtypelem9.1 and mapfien.*


icecream17 (approved):
fpwwe2lem5 onwards were renamed fpwwe2lem4 etc

Issue #4091, opened 2024-07-21
double-negated ordinal trichotomy in iset.mm


jkingdon:
We should be able to prove each of the following

* `-. -. EXMID <-> -. -. A. x e. On A. y e. On ( x e. y \/ x = y \/ y e. x )`
* `-. -. EXMID <-> -. -. A. x e. On A. y e. On ( x C_ y \/ y C_ x )`
* `-. -. EXMID <-> A. x e. On A. y e. On -. -. ( x e. y \/ x = y \/ y e. x )`
* `-. -. EXMID <-> A. x e. On A. y e. On -. -. ( x C_ y \/ y C_ x )`

There is more discussion and a proof which looks like it shouldn't be hard to formalize in iset.mm, in [this mathoverflow comment by James E Hanson](https://mathoverflow.net/questions/452678/does-negative-trichotomy-hold-for-constructive-ordinals/475265#475265)

Oh and #4090 is used in the proof so if that isn't already proved, prove that first.

PR #4108, opened 2024-08-03
theorems related to doubly negated ordinal trichotomy

### LINKED ISSUE #4090: Strengthen exmidpw from equinumerosity to equality
[exmidpw](https://us.metamath.org/ileuni/exmidpw.html) is `EXMID ↔ 𝒫 1o ≈ 2o`

But we could easily prove `EXMID ↔ 𝒫 1o = 2o` as well. This is a passing remark in https://mathoverflow.net/questions/452678/does-negative-trichotomy-hold-for-constructive-ordinals/475265#475265 (worded as "Note that 𝖫𝖤𝖬 is equivalent to the statement Ω=2").

The proof of exmidpw already has most of what is needed to prove this, so it shouldn't be hard.

jkingdon:
This formalizes most of the theorems from #4091 (with the one gap being that we don't have #739 yet and thus include it as a hypothesis for the moment).

Fixes #4090 

I put down James E Hanson as a co-contributor because this is a rather direct formalization of the informal proofs at https://mathoverflow.net/questions/452678/does-negative-trichotomy-hold-for-constructive-ordinals/475265#475265

Includes a bit of rearranging in my mathbox (reordering things and the like) but nothing too big.


tirix (approved):

avekens (approved):

benjub:
@jkingdon Several of these theorems are of the form -. PHI -> -. PSI (I include those of the form -. PHI -> A =/= B ). Can you, for these, write in the comments whether the stronger form PSI -> PHI is known to be non-provable intuitionistically or if there is a TODO to add the proof ?

benjub:
Also, concerning ~exmidpweq , that is added in this MR, I would like to quote my comment from https://github.com/metamath/set.mm/issues/4090#issuecomment-2269278029 and your answer:

> > @jkingdon : It would probably be more economical to prove exmidpweq in terms of exmidpw or conversely. Especially since both current proofs prove independently each implication. You only need to prove three implications, and not four, and one of them is simply "equality implies equinumerosity".
>
> I don't have much of an opinion about how we prove these things. So I'm unlikely to object if someone wants to revise the proofs.

Since the theorem is in the main part (and not in a mathbox), I think that the shortening I propose should be done, and actually should have been done before the MR was approved. The current situation is: ~exmidpweq and ~exmidpw
prove respectively `|- ( PHI <-> PSI )` and `|- ( PHI <-> PSI' )` with `PSI` and `PSI'` very close to each other and `|- ( PSI -> PSI')` a 2-line proof from ~eqeng. But both theorems each prove both implications, so you prove four implications whereas it suffices to prove only `PHI -> PSI` and `PSI' -> PHI`. Furthermore, there are currently no links in the comments between these two theorems.

I do not know what approvers @tirix and @avekens think, but to me, since this is a collective project, we have a shared responsibility to strive for a high quality standard of the main parts of the Metamath databases hosted in this repository.

jkingdon:
> @jkingdon Several of these theorems are of the form -. PHI -> -. PSI (I include those of the form -. PHI -> A =/= B ). . . . whether the stronger form PSI -> PHI is known to be non-provable intuitionistically

* From pw1nel3: `~P 1o e. 3o -> EXMID` should follow from exmidpweq , pw1ne0 , and pw1ne1
* From sucpw1ne3: `suc ~P 1o = 3o -> EXMID`. Seems provable from suc11 and exmidpweq
* From sucpw1nss3: `suc ~P 1o C_ 3o -> EXMID`. Not too sure about this one. I was looking at things like https://us.metamath.org/ileuni/onsucsssucexmid.html but I'm not sure how much they help.
* From 3nsssucpw1: `3o C_ suc ~P 1o - EXMID`. Likewise.

digama0:
> From sucpw1nss3: `suc ~P 1o C_ 3o -> EXMID`. Not too sure about this one. I was looking at things like https://us.metamath.org/ileuni/onsucsssucexmid.html but I'm not sure how much they help.

Doesn't `suc ~P 1o C_ 3o` imply `~P 1o C_ 2o` by ~onsucsssucr, and then this should imply `EXMID` because of ~ssfilem.

> From 3nsssucpw1: `3o C_ suc ~P 1o - EXMID`. Likewise.

The assumption `-. EXMID` in 3nsssucpw1 is nonsense, this assumption is false and so everything follows from it. That is, you can prove `( -. EXMID -> ph )`, because `-. -. EXMID` is provable.

For the non-negated version in the question, this is easy because if you look at the proof of 3nsssucpw1 you will notice that this is how it is already proven.

jkingdon:
> Doesn't suc ~P 1o C_ 3o imply ~P 1o C_ 2o by ~onsucsssucr, and then this should imply EXMID because of ~ssfilem.

This looks right. If `~P 1o C_ 2o` then a subset of 1o is either (/) or 1o which is pretty directly the right hand side of https://us.metamath.org/ileuni/exmid01.html .

> `-. -. EXMID` is provable

Don't think so. Most directly stated as "¬𝖫𝖤𝖬 holds in sheaves over Cantor space and the reals. ¬𝖫𝖤𝖬 also holds in the effective topos/realizability model of 𝖨𝖹𝖥." in https://mathoverflow.net/questions/452678/does-negative-trichotomy-hold-for-constructive-ordinals/475265#475265 .

In fact most of the answers and comments at https://mathoverflow.net/questions/452678/does-negative-trichotomy-hold-for-constructive-ordinals/475265 give good insight into what the whole point of these theorems is.

> For the non-negated version in the question, this is easy because if you look at the proof of 3nsssucpw1 you will notice that this is how it is already proven.

Ah yes, now that I go look again at the proof this is correct.

benjub:
Indeed, $\not\vdash \neg\neg \mathrm{exmid}$, yet $\vdash \neg\neg \mathrm{decid} \varphi$ (https://us.metamath.org/ileuni/nndc.html). This is a pitfall that maybe we should mention in the comment of https://us.metamath.org/ileuni/df-exmid.html ?

**edit:** oh, and my comment in https://us.metamath.org/ileuni/nnexmid.html should be changed, and indeed the label too !

jkingdon:
> Indeed, ⊬ ¬ ¬ exmid , yet ⊢ ¬ ¬ decid φ

Whoa, this is kind of wild.

The difference, if I'm reading this right, is that ` DECID φ ` is about decidability of a particular (although arbitrary) proposition, whereas ` EXMID ` is about decidability of all propositions.

Or to say it in terms of ` EXMID <-> A. x e. ~P 1o DECID x = 1o ` (proved at #4116 ), we can prove ` ¬ ¬ DECID x = 1o ` but to turn that into ` ¬ ¬ EXMID ` would require the converse of https://us.metamath.org/ileuni/nnral.html , which we don't have.

As for labels, I'm not sure we need to change labels like https://us.metamath.org/mpeuni/exmid.html or https://us.metamath.org/ileuni/exmiddc.html but I see your point about https://us.metamath.org/ileuni/nnexmid.html (well the comment for sure, maybe the label too).

benjub:
Yes, it's an example of the implication in https://us.metamath.org/ileuni/nnral.html or https://us.metamath.org/ileuni/bj-nnal.html not being a biconditional:

If we could quantify over formulas, then, EXMID could be expressed as $\forall \varphi \mathrm{decid} \varphi$. By generalization over $\varphi$ of $\vdash \neg\neg \mathrm{decid} \varphi$, we would have $\vdash \forall \varphi \neg\neg \mathrm{decid} \varphi$, but on the other hand, $\not\vdash \neg\neg \mathrm{exmid}$, that is, $\not\vdash \neg\neg \forall \varphi \mathrm{decid} \varphi$.

PR #4109, opened 2024-08-04
Correspondence theorem

tirix:
Proof of the [correspondence theorem](https://en.wikipedia.org/wiki/Correspondence_theorem), using [Galois Connections](https://en.wikipedia.org/wiki/Galois_connection).

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -101,6 +101,9 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 4-Aug-24 rnmpt0    rnmpt0f     moved from GS's mathbox to main set.mm
+ 4-Aug-24 rnmptn0   [same]      moved from GS's mathbox to main set.mm
+31-Jul-24 f1oeq1d   [same]      moved from GS's mathbox to main set.mm
 24-Jul-24 grpmndd   [same]      moved from SN's mathbox to main set.mm
 24-Jul-24 zringsubgval [same]   moved from AV's mathbox to main set.mm
 21-Jul-24 syl6rbbr  bitr4id     order of hypotheses is switched


icecream17 (approved):
Nice! I just stumbled upon the [first isomorphism theorem](https://en.wikipedia.org/wiki/Isomorphism_theorems#Groups) the other day; it's fascinating how everything is connected

tirix:
> Nice! I just stumbled upon the [first isomorphism theorem](https://en.wikipedia.org/wiki/Isomorphism_theorems#Groups) the other day; it's fascinating how everything is connected

Yes, those are quite fundamental. Ultimately we'll need all 4 isomorphism theorems.
Actually for the first theorem, we already have almost the same content, scattered among several theorems:
* We have [~ghmker](https://us.metamath.org/mpeuni/ghmker.html) which is part (i) of the first isomorphism for groups, and I have recently proved [~kerlidl](https://us.metamath.org/mpeuni/kerlidl.html), which is the similar theorem for rings.
* There is [~ghmima](https://us.metamath.org/mpeuni/ghmima.html) which is nearly part (ii) of the first isomorphism for groups, and [~rhmima](https://us.metamath.org/mpeuni/rhmima.html) for rings. 
* Similarly, for (iii), we have [~qusghm](https://us.metamath.org/mpeuni/qusghm.html) and [~qusrhm](https://us.metamath.org/mpeuni/qusrhm.html).

It might be interesting to add this in the respective theorem's comments.

savask:
That's a very nice result!

Maybe worth mentioning here a useful generalization. Say, a group A acts on a group G by automorphisms. If some normal subgroup N of G is A-invariant, then all A-invariant subgroups of G containing N are in correspondence with A-invariant subgroups of G/N.

One application, which is mentioned in the wiki article you linked, is that normal subgroups containing N are in correspondence with the normal subgroups of G/N. To see this, one can take A = G acting on G by conjugation. Of course, similar stuff works for rings, modules etc, I wonder if there's a good enough formalism to cover all these special cases in one sweep.

tirix:
> Similar stuff works for rings, modules etc, I wonder if there's a good enough formalism to cover all these special cases in one sweep.

Of course, one formalism to cover them all is Category Theory. But there is also a different approach, called _Universal Algebra_, which tries to do just that. Its [Wikipedia page](https://en.wikipedia.org/wiki/Universal_algebra) also cites the isomorphism theorems as one application. But that seemed to veer off too much of my shorter term goals, so I chose not to go into that direction.

On the other hand, Galois connections seemed to be more interesting, because they are related to Galois theory, and because they are in the field of Order Theory, which is more fundamental and for which we already had some formalism. This explains my digression into Galois connections, and why I chose to prove this theorem using Galois connections.

In the end, the proof is probably longer than a direct proof, and it misses the generalization you mention, so it has downsides too.

set.mm
```diff
@@ -477620,6 +477788,376 @@ The sumset operation can be used for both group (additive) operations and
...
+      wa rexsn 3bitr4g bitrd eqrdv ) DLMZABNZAOUAZUBZIEUCZACPZAVPIQZVRMZVSJQZKQ
+      ZDUDUEZPZRZKASZJVQSZVSAMZVMVNVTWGUFZVOVMVNVQBNZWIVMEBBDEFHUGUHVMWJVNWIVMW
+      JVNWIJKBWCCVQADLVSFWCUQZGUIUJUKULUMVPVSEWBWCPZRZKASZWBVSRZKASWGWHVPWMWOKA
+      VPWBAMZVHZWMVSWBRWOWQWLWBVSWQVMWBBMWLWBRVMVNVOWPUNVPABWBVMVNVOUOUPBWCDWBE
+      FWKHURUSTIKUTVAVBWFWNJEEDVCHVDWAERZWEWMKAWRWDWLVSWAEWBWCVETVFVIKVSAVGVJVK
+      VL $.
+
+    $( Sumset with the identity singleton is the original set.  (Contributed by
+       Thierry Arnoux, 27-Jul-2024.) $)
+    subglsm0l $p |- ( ( G e. Grp /\ A e. ( SubGrp ` G ) )
```
avekens, line 477812: Maybe "Sumset of a subgroup with the identity singleton is the original subgroup."

I think the hypothesis  `` B = ( Base ` G ) `` is not needed here, and the antecedent `G e. Grp`is also not required (follows from ``A e. ( SubGrp ` G )`` by ~issubg).

set.mm
```diff
@@ -477620,6 +477788,376 @@ The sumset operation can be used for both group (additive) operations and
...
+  $}
+
+  ${
+    $d .(+) x $.  $d A a o x $.  $d B a o $.  $d G a o x $.  $d X a o x $.
+    grplsmid.b $e |- B = ( Base ` G ) $.
+    grplsmid.p $e |- .(+) = ( LSSum ` G ) $.
+    $( The direct sum of an element ` X ` of a subgroup ` A ` is the subgroup
+       itself.  (Contributed by Thierry Arnoux, 27-Jul-2024.) $)
+    grplsmid $p |- ( ( G e. Grp /\ A e. ( SubGrp ` G ) /\ X e. A )
+                  -> ( { X } .(+) A ) = A ) $=
```
avekens, line 477826: See comment on ~subglsm0l.

set.mm
```diff
@@ -477620,6 +477788,376 @@ The sumset operation can be used for both group (additive) operations and
...
+  ${
+    $d B i j k $.  $d G i j k $.  $d S i j k $.  $d X i j k $.  $d i j k ph $.
+    quslsm.b $e |- B = ( Base ` G ) $.
+    quslsm.p $e |- .(+) = ( LSSum ` G ) $.
+    quslsm.g $e |- ( ph -> G e. Grp ) $.
+    quslsm.n $e |- ( ph -> S e. ( SubGrp ` G ) ) $.
+    quslsm.s $e |- ( ph -> X e. B ) $.
+    $( Express the image by the quotient map in terms of direct sum.
+       (Contributed by Thierry Arnoux, 27-Jul-2024.) $)
+    quslsm $p |- ( ph -> [ X ] ( G ~QG S ) = ( { X } .(+) S ) ) $=
```
avekens, line 477858: See comment on ~subglsm0l

set.mm
tirix: ...and then this theorem becomes exactly [`~lsm02`](https://us.metamath.org/mpeuni/lsm02.html)!
Thanks, many functions operating on structures work on `_V`, and I never noticed `SubGrp` only accepts groups.
I'll also apply that "trick" to `~nsgqus0` just above, and now I notice hypotheses in others of my previous theorems can be simplified, too. I'll do that as I encounter them.

tirix: Done! I've completely removed this theorem in favor of `~lsm02`, and reworked the other theorems in this PR to remove the `G e. Grp` hypothesis.

avekens (approved):

PR #4111, opened 2024-08-05
Reduce ax-pow usage via fvprc

BTernaryTau:
This change creates an alternate version of fvprc that uses ax-sep and ax-pr instead of ax-pow. It also uses fvprcALT to remove ax-pow dependencies for some theorems that already depend on ax-sep and ax-pr.

> A function's value at a proper class is the empty set.  See ~ fvprcALT for a proof that uses ~ ax-sep and ~ ax-pr instead of ~ ax-pow .

> Alternate proof of ~ fvprc using ~ ax-sep and ~ ax-pr instead of ~ ax-pow .  While this proof requires more axioms, it can still be useful for eliminating dependencies on the Axiom of Power Sets in proofs that already depend on Separation and Pairing.  The "discouraged" tags are here only because this is an *ALT result.

Changes in axiom usage can be found here: https://github.com/metamath/set.mm/commit/7a2bde57a9a49023020c85c64e21bf84e4afb6b4

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -5425,6 +5425,7 @@
 "drsb1" is used by "iotaeq".
 "drsb1" is used by "sb2ae".
 "drsb1" is used by "sbco3".
+"dtruALT2" is used by "fvprcALT".
 "dvadiaN" is used by "diarnN".
 "dveel1" is used by "distel".
 "dveel2" is used by "axc14".
@@ -15750,7 +15751,7 @@ New usage of "drngcatALTV" is discouraged (0 uses).
 New usage of "drngoi" is discouraged (2 uses).
 New usage of "drsb1" is discouraged (3 uses).
 New usage of "dtruALT" is discouraged (0 uses).
-New usage of "dtruALT2" is discouraged (0 uses).
+New usage of "dtruALT2" is discouraged (1 uses).
 New usage of "dtrucor2" is discouraged (0 uses).
 New usage of "dummylink" is discouraged (0 uses).
 New usage of "dvadiaN" is discouraged (1 uses).
@@ -16194,6 +16195,7 @@ New usage of "funop" is discouraged (2 uses).
 New usage of "funopg" is discouraged (0 uses).
 New usage of "funopsn" is discouraged (2 uses).
 New usage of "fvimacnvALT" is discouraged (0 uses).
+New usage of "fvprcALT" is discouraged (0 uses).
 New usage of "gcdmultipleOLD" is discouraged (0 uses).
 New usage of "gcdmultiplezOLD" is discouraged (0 uses).
 New usage of "gen11" is discouraged (19 uses).
@@ -19799,6 +19801,7 @@ Proof modification of "funcrngcsetcALT" is discouraged (765 steps).
 Proof modification of "fundcmpsurinjALT" is discouraged (221 steps).
 Proof modification of "fvilbdRP" is discouraged (27 steps).
 Proof modification of "fvimacnvALT" is discouraged (102 steps).
+Proof modification of "fvprcALT" is discouraged (259 steps).
 Proof modification of "gcdmultipleOLD" is discouraged (348 steps).
 Proof modification of "gcdmultiplezOLD" is discouraged (230 steps).
 Proof modification of "gen11" is discouraged (27 steps).
--- a/discouraged
+++ b/discouraged
@@ -6300,6 +6300,27 @@
 "funop" is used by "funsndifnop".
 "funopsn" is used by "funop".
 "funopsn" is used by "funop1".
+"fvprcALT" is used by "1stval".
+"fvprcALT" is used by "2idlval".
+"fvprcALT" is used by "2ndval".
+"fvprcALT" is used by "bafval".
+"fvprcALT" is used by "brfvopab".
+"fvprcALT" is used by "brfvopabrbr".
+"fvprcALT" is used by "dffv3".
+"fvprcALT" is used by "edgval".
+"fvprcALT" is used by "fvmptopab".
+"fvprcALT" is used by "fvmptrabfv".
+"fvprcALT" is used by "fvrn0".
+"fvprcALT" is used by "fvunsn".
+"fvprcALT" is used by "grpidval".
+"fvprcALT" is used by "iedgval".
+"fvprcALT" is used by "indislem".
+"fvprcALT" is used by "rnfvprc".
+"fvprcALT" is used by "rrgval".
+"fvprcALT" is used by "s1prc".
+"fvprcALT" is used by "smfval".
+"fvprcALT" is used by "vafval".
+"fvprcALT" is used by "vtxval".
 "gen11" is used by "ax6e2eqVD".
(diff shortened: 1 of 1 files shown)


icecream17 (approved):

tirix (approved):
For those interested, [this Mathoverflow question](https://mathoverflow.net/questions/133597/what-would-remain-of-current-mathematics-without-axiom-of-power-set) asks about ZFC without the axiom of powerset, so it is an interesting goal to pursue to find out what can be proven without that axiom.

GinoGiotto:
> Changes in axiom usage can be found here: https://github.com/metamath/set.mm/commit/7a2bde57a9a49023020c85c64e21bf84e4afb6b4

I suppose it took you an eternity to generate this file with metamath-exe. It might interest you that [metamath-knife](https://github.com/metamath/metamath-knife) provides the same information in less than a second with the command `metamath-knife set.mm -X ax.txt`.

There is also the command `metamath-knife set.mm --stmt-use usage.txt syl,bezout` to show direct and indirect usage of any list of statements you like (in this example it shows usages of theorems [syl](https://us.metamath.org/mpeuni/syl.html) and [bezout](https://us.metamath.org/mpeuni/bezout.html) and prints them in `usage.txt` file).

BTernaryTau:
@GinoGiotto thanks, I was wondering what you used. For metamath-exe it takes a couple hours for each run, so it's definitely workable, but far less convenient.

set.mm
```diff
@@ -62489,20 +62489,45 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
...
     $d x A $.  $d x F $.
-    $( A function's value at a proper class is the empty set.  (Contributed by
-       NM, 20-May-1998.) $)
+    $( A function's value at a proper class is the empty set.  See ~ fvprcALT
+       for a proof that uses ~ ax-sep and ~ ax-pr instead of ~ ax-pow .
+       (Contributed by NM, 20-May-1998.) $)
     fvprc $p |- ( -. A e. _V -> ( F ` A ) = (/) ) $=
       ( vx cvv wcel wn cv wbr weu cfv c0 wceq brprcneu tz6.12-2 syl ) ADEFACGBH
       CIFABJKLCABMCABNO $.
   $}
```
avekens, line 62498: So we can add "$( $j usage 'fvprc' avoids 'ax-sep' 'ax-pr' ; $)" here?

set.mm
```diff
@@ -62489,20 +62489,45 @@ empty set when it is not meaningful (as shown by ~ ndmfv and ~ fvprc ).
...
+  ${
+    $d A x y $.  $d F x y $.
+    $( Alternate proof of ~ fvprc using ~ ax-sep and ~ ax-pr instead of
+       ~ ax-pow .  While this proof requires more axioms, it can still be
+       useful for eliminating dependencies on the Axiom of Power Sets in proofs
+       that already depend on Separation and Pairing.  (Contributed by
+       BTernaryTau, 3-Aug-2024.)  The "discouraged" tags are here only because
+       this is an *ALT result.  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    fvprcALT $p |- ( -. A e. _V -> ( F ` A ) = (/) ) $=
```
avekens, line 62509: Maybe we can call this theorem ~fvprc2 to avoid the suffix ALT and the "discouraged" tags.

BTernaryTau: That would definitely make things a lot simpler. We could also create a different suffix to differentiate between identical theorems with different proofs/axioms and merely similar theorems. (If the existing suffix was something like DUP instead of ALT, I would suggest ALT as the new suffix.) Maybe something like fvprcAX2, or fvprcV2?

set.mm
BTernaryTau: Done.

set.mm
avekens: I made a similar suggestion some years ago (I proposed suffix ALTV), but this was not accepted in general, but only for mathboxes (see "Conventions").

benjub:
I think that the theorem introduced in this PR should be the "main" fvprc, since ax-pow is among the more "powerful" axioms of ZF, and certainly more than the "staple" axioms { ax-sep , ax-pr }.  The comments could be simply:

> $( The value of a function at a proper class is the empty set.  See ~ fvprcALT
       for a shorter proof requiring ~ ax-pow instead of ~ ax-sep and ~ ax-pr .
       (Contributed by NM, 20-May-1998.)  Remove dependency on ~ ax-pow .
       (Revised by BTernaryTau, 3-Aug-2024.) $)

and

> $( Alternate proof of ~ fvprc , shorter but using ~ ax-pow instead of ~ ax-sep and ~ ax-pr .
       (Contributed by NM, 20-May-1998.)  (Proof modification is discouraged.)

(No need for another suffix than ALT, and no need to write that the discouragement tags are here because of the ALT suffix, since this requirement is in the conventions.)

jkingdon:
> For those interested, [this Mathoverflow question](https://mathoverflow.net/questions/133597/what-would-remain-of-current-mathematics-without-axiom-of-power-set) asks about ZFC without the axiom of powerset, so it is an interesting goal to pursue to find out what can be proven without that axiom.

That mathoverflow question is an interesting read because I was more familiar with CZF vs IZF (the latter has power set, the former something weaker).

A full investigation of this would also change other axioms (ZFC<sup>&minus;</sup> has an axiom called "collection" and I didn't read the whole paper but I assume it is https://us.metamath.org/ileuni/ax-coll.html or conceivably https://us.metamath.org/ileuni/ax-strcoll.html ). But that doesn't affect the current pull request.

avekens:
> I think that the theorem introduced in this PR should be the "main" fvprc, since ax-pow is among the more "powerful" axioms of ZF, and certainly more than the "staple" axioms { ax-sep , ax-pr }. 

I agree! We have already a similar situation for ~fnex and ~fnexALT : Here we have the same theorem with different proofs, and the proof of ~fnexALT uses ax-pow, while ~fnex does not.

benjub:
> A full investigation of this would also change other axioms (ZFC− has an axiom called "collection" and I didn't read the whole paper but I assume it is https://us.metamath.org/ileuni/ax-coll.html or conceivably https://us.metamath.org/ileuni/ax-strcoll.html ). But that doesn't affect the current pull request.

Yes, this is that axiom of collection. The "strong" version is stronger in that it gives you not only the existence of a set containing what you want and possibly more: it gives you the existence of a set containing exactly what you want and nothing more (see comment of https://us.metamath.org/ileuni/ax-strcoll.html). As a consequence, the strong version is typically necessary when you do not have the axiom of separation (like in CZF where you only have bounded separation). Indeed, when you have separation, you can deduce strong collection from collection: you simply "separate", as in "cut out", what you want. On the need of stronger axioms when full separation is not available, see also the comment of https://us.metamath.org/ileuni/ax-infvn.html.

avekens:
Converted to "draft" since this PR becomse obsolete if #4115 is accepted and merged.

benjub:
Closing as superseded by #4115.

PR #4112, opened 2024-08-07
iset: characteristic functions

benjub:
FYI @jkingdon : bj-charfun* look like the general statements that are used in several theorems, e.g., dceqnconst.


iset.mm
```diff
@@ -40722,17 +40722,22 @@ arbitrary sets or real numbers will not be provable (without the law of
...
-    $( A member of a natural number is a natural number.  (Contributed by NM,
-       21-Jun-1998.) $)
-    elnn $p |- ( ( A e. B /\ B e. _om ) -> A e. _om ) $=
-      ( vy vx com wcel wss cv c0 csuc sseq1 0ss wa csn cun unss vex snss anbi2i
-      df-suc sseq1i 3bitr4i biimpi expcom finds ssel2 ancoms sylan2 ) BEFABFZBE
-      GZAEFZCHZEGIEGDHZEGZUMJZEGZUJCDBULIEKULUMEKULUOEKULBEKELUNUMEFZUPUNUQMZUP
-      UNUMNZEGZMUMUSOZEGURUPUMUSEPUQUTUNUMEDQRSUOVAEUMTUAUBUCUDUEUJUIUKBEAUFUGU
-      H $.
+    $d A x y $.
+    $( A finite ordinal is, as a set, included in the set of finite ordinals.
```
jkingdon, line 40726: Can't use the word finite ordinal here (or in any similar contexts) in light of https://us.metamath.org/ileuni/exmidonfin.html

jkingdon (changes requested):
Except for the terminology around finite ordinals, looks good. A nice exploration around decidable subsets and similar topics.

iset.mm
benjub: Oh, good catch. I just had a look at https://us.metamath.org/ileuni/df-iom.html.

Since we have two different sets, $\omega$ and $\mathbb{N}$, and the term "natural number" looks better suited for the latter, we should pick a name for the former, to use it consistently in the database. Since there are ordinals, a candidate is "natural number ordinal" (since "natural ordinal" may be too obscure"). What do you think ?

jkingdon: That's what I ended up picking at #4113 so although I think there might be contexts where we say something else, this seems like a good default wording.

jkingdon (approved):

iset.mm
```diff
@@ -153349,12 +153490,12 @@ Inductive classes and the class of natural numbers (finite ordinals)
 
 $(
 -.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-.-
-  The set of natural numbers (finite ordinals)
+  The set of natural numbers ordinals
```
icecream17, line 153493:
```suggestion
  The set of natural number ordinals
```

icecream17 (approved):

PR #4114, opened 2024-08-08
eliminate 17 antecedents in polynomial bag theorems

icecream17:
and
* elrabi, elrn*, eldm* ax-10,11,12 reductions
* `g`-suffixed variants of theorems with ` F e. V ` instead of ` I e. V ` to avoid ax-rep (only a handful of ax-rep usages saved... it's likely that that the theorems I add here can be used more, also most of the df-map section can have versions avoiding ax-rep, but that's out of the scope of this pr)

commit by commit (though a few things are out of order)

Axiom changes: https://github.com/metamath/set.mm/commit/6f430cd8583700f75c3a512dcd88539172cc33a3 (does not include savings from ssrab2)

---

I added a revision tag and OLD version if I removed an antecedent, this gets repetitive but it seems to be the process: https://github.com/metamath/set.mm/pull/4101

(if nothing changed except the proof uses a newer theorem, I don't add tags or OLD versions)

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -103,8 +103,8 @@ Date      Old       New         Notes
  7-Aug-24 tdeglem4  [same]      revised - eliminated unnecessary antecedent
  7-Aug-24 tdeglem3  [same]      revised - eliminated unnecessary antecedent
  7-Aug-24 tdeglem1  [same]      revised - eliminated unnecessary antecedent
- 7-Aug-24 psrbagev2 [same]      revised - eliminated unnecessary antecedent
- 7-Aug-24 psrbagev1 [same]      revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrbagev2 [same]      revised - eliminated unnecessary hypothesis
+ 7-Aug-24 psrbagev1 [same]      revised - eliminated unnecessary hypothesis
  7-Aug-24 psrbagfsupp [same]    revised - eliminated unnecessary antecedent
  7-Aug-24 psrbagaddcl [same]    revised - eliminated unnecessary antecedent
  7-Aug-24 psrass1lem [same]     revised - eliminated unnecessary antecedent
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -107,9 +107,9 @@ Date      Old       New         Notes
  7-Aug-24 psrbagev1 [same]      revised - eliminated unnecessary hypothesis
  7-Aug-24 psrbagfsupp [same]    revised - eliminated unnecessary antecedent
  7-Aug-24 psrbagaddcl [same]    revised - eliminated unnecessary antecedent
- 7-Aug-24 psrass1lem [same]     revised - eliminated unnecessary antecedent
- 7-Aug-24 gsumbagdiag [same]    revised - eliminated unnecessary antecedent
- 7-Aug-24 gsumbagdiaglem [same] revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrass1lem [same]     revised - eliminated unnecessary hypothesis
+ 7-Aug-24 gsumbagdiag [same]    revised - eliminated unnecessary hypothesis
+ 7-Aug-24 gsumbagdiaglem [same] revised - eliminated unnecessary hypothesis
  7-Aug-24 psrbagconf1o [same]   revised - eliminated unnecessary antecedent
  7-Aug-24 psrbagconcl [same]    revised - eliminated unnecessary antecedent
  7-Aug-24 psrbaglefi [same]     revised - eliminated unnecessary antecedent
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -100,6 +100,23 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+ 7-Aug-24 tdeglem4  [same]      revised - eliminated unnecessary antecedent
+ 7-Aug-24 tdeglem3  [same]      revised - eliminated unnecessary antecedent
+ 7-Aug-24 tdeglem1  [same]      revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrbagev2 [same]      revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrbagev1 [same]      revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrbagfsupp [same]    revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrbagaddcl [same]    revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrass1lem [same]     revised - eliminated unnecessary antecedent
+ 7-Aug-24 gsumbagdiag [same]    revised - eliminated unnecessary antecedent
+ 7-Aug-24 gsumbagdiaglem [same] revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrbagconf1o [same]   revised - eliminated unnecessary antecedent
+ 7-Aug-24 psrbagconcl [same]    revised - eliminated unnecessary antecedent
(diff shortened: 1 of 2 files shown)


set.mm
```diff
@@ -33978,10 +33978,23 @@ generally appear in a single form (either definitional, but more often
   $}
 
   ${
-    $d A x $.  $d V x $.
+    $d A y $.  $d V x y $.  $d ph y $.
     $( Implication for the membership in a restricted class abstraction.
-       (Contributed by Alexander van der Vekens, 31-Dec-2017.) $)
+       (Contributed by Alexander van der Vekens, 31-Dec-2017.)  Remove disjoint
+       variable condition on ` A ` , ` x ` and avoid ~ ax-10 , ~ ax-11 ,
```
benjub, line 33984:
```suggestion
       variable condition on ` A , x ` and avoid ~ ax-10 , ~ ax-11 ,
```
(That's how I write it for the HTML rendering with, I think, less spacing.)

set.mm
```diff
@@ -54603,18 +54616,31 @@ ordered pairs as classes (in set.mm, the Kuratowski encoding).  A more
...
+      IUDUFCUCBSMNOAECQR $.
 
     $( Membership in a range.  (Contributed by Scott Fenton, 2-Feb-2011.) $)
     elrng $p |- ( A e. V -> ( A e. ran B <-> E. x x B A ) ) $=
       ( wcel crn cv cop wex wbr elrn2g df-br exbii bitr4di ) BDEBCFEAGZBHCEZAIO
       BCJZAIABCDKQPAOBCLMN $.
   $}
 
+  ${
+    $d x y A $.  $d x y B $.
```
benjub, line 54633: Probably there is no `y` in these two proofs ?

set.mm
```diff
@@ -264005,11 +264112,37 @@ the same dimension over the same (nonzero) ring.  (Contributed by AV,
...
+      sylbi ) CAFCBGHIJKFZBLDMNZOZFZDLCPZAUDCEQUECUCFUFUBBCUCRCLDSTUA $.
+
+    $( Obsolete version of ~ psrbag as of 6-Aug-2024.  (Contributed by Mario
+       Carneiro, 29-Dec-2014.)  (Proof modification is discouraged.)
+       (New usage is discouraged.) $)
+    psrbagfOLD $p |- ( ( I e. V /\ F e. D ) -> F : I --> NN0 ) $=
       ( wcel cn0 wf ccnv cn cima cfn psrbag simprbda ) DEGCAGDHCICJKLMGABCDEFNO
       $.
 
+    $( Finite bags have finite nonzero-support.  (Contributed by Stefan O'Rear,
```
benjub, line 264128:
```suggestion
    $( Finite bags have finite support.  (Contributed by Stefan O'Rear,
```

set.mm
```diff
@@ -264005,11 +264112,37 @@ the same dimension over the same (nonzero) ring.  (Contributed by AV,
       CTEUFUGUHUI $.
 
     $( A finite bag is a function.  (Contributed by Mario Carneiro,
-       29-Dec-2014.) $)
-    psrbagf $p |- ( ( I e. V /\ F e. D ) -> F : I --> NN0 ) $=
+       29-Dec-2014.)  Remove extra antecedent.  (Revised by SN,
```
benjub, line 264115: It could be "Remove a sethood antecedent.", to give a bit more information. Since sethood antecedents are quite common in set.mm, this gives a good idea of the nature of the revision.  (The fact that it was extraneous is contained in the fact that you could remove it.)

set.mm
```diff
@@ -264031,34 +264164,80 @@ the same dimension over the same (nonzero) ring.  (Contributed by AV,
...
-      DZMZDAHZULDNOPZQHZUJUKULUMUAUOCNOPZUQUOEICJZUSQHZUOUKUTVAMZUJUKULUMUBUJUK
-      VBRUNABCEFGSTUEUFABCDEFGUGUHUJUPULURMRUNABDEFGSTUI $.
+      wbr w3a mpbid simprd psrbaglesuppOLD ssfid mpbir2and ) EFHZCAHZEIDJZDCKLU
+      CZUDZMZDAHZULDNOPZQHZUJUKULUMUAUOCNOPZUQUOEICJZUSQHZUOUKUTVAMZUJUKULUMUBU
+      JUKVBRUNABCEFGSTUEUFABCDEFGUGUHUJUPULURMRUNABDEFGSTUI $.
 
     $( The sum of two finite bags is a finite bag.  (Contributed by Mario
-       Carneiro, 9-Jan-2015.) $)
-    psrbagaddcl $p |- ( ( I e. V /\ F e. D /\ G e. D ) ->
+       Carneiro, 9-Jan-2015.)  Shorten and remove extra antecedent.  (Revised
```
benjub, line 264225: Shorten **proof** and...

set.mm
```diff
@@ -264077,198 +264256,394 @@ the same dimension over the same (nonzero) ring.  (Contributed by AV,
...
+        CEWIYDUVMXHUVNVWQAJFUWTYGWTALXFYHVVAVWSAVWRUXAYIFUXAVURYJJFUWTUXAUXAVJY
+        KYOWTFUXAWIWIVURYLYMVWHYSAVVPUWGLUNVWOYRYTAUXEUWNLUNAHFUXDUWMUXNUXDLUXQ
+        UNUQUWMUXNUWKDUWKUXCLUXPXEVURSVUSAVVSUXITVPVVLUXRUXNUXCWIURZUXCXRZVWBVV
+        KUXCVURXSUQZUWKXTZUXCVURYAUJUXNVWEUWKWIURVWTVWEUXNVWFWTUWJBEWIYDIUWKNWI
+        YFYNVXAUXNIUWKNYGWTUXNLXFYHVVLVXCUXNVXBUXCYIUWKUXCVURYJIUWKNUXCUXCVJYKY
+        OWTUWKUXCWIWIVURYLYMUYFYSUXNUXQUWLLUNVUNYRYTWHYRWO $.
+    $}
+
+    $( Once these OLD theorems are deleted, the above curly braces become
+       unnecessary, and can be deleted. $)
```
benjub, line 264530: Cannot the curly braces be adjusted now (possibly by having shorter scopes, which is often a better practice anyway since long scopes are error-prone), so that we do not forget it next year when we remove the OLD versions ?

benjub (approved):
LGTM, but I only skimmed through the later changes, so another pair of eyes would be appreciated (the second approver could also be the merger).

set.mm
```diff
@@ -264077,198 +264263,402 @@ the same dimension over the same (nonzero) ring.  (Contributed by AV,
...
+  ${
+    $d D u v w x $.  $d D y $.  $d D x z $.  $d F f $.  $d F u v w x $.
+    $d F y $.  $d F x z $.  $d I f $.  $d I z $.  $d S u v w $.  $d S z $.
+    $d X f $.  $d X u v w x $.  $d X y $.  $d X x z $.  $d Y f $.
+    $d Y u v w x $.  $d Y y $.  $d Y x z $.  $d ph u v w $.  $d ph z $.
+    gsumbagdiag.d $e |- D = { f e. ( NN0 ^m I ) | ( `' f " NN ) e. Fin } $.
+    gsumbagdiag.s $e |- S = { y e. D | y oR <_ F } $.
+    gsumbagdiag.f $e |- ( ph -> F e. D ) $.
+    $( Lemma for ~ gsumbagdiag .  (Contributed by Mario Carneiro, 5-Jan-2015.)
+       Remove a sethood antecedent.  (Revised by SN, 6-Aug-2024.) $)
```
avekens, line 264548: Here, I think it is a sethood hypothesis which was removed.

benjub (approved):

avekens (approved):

PR #4117, opened 2024-08-09
iset: shorten pw1dc1

benjub:
* ss1o0el1 was a step in the proof of exmid01 so I extracted it.
* shorten pw1dc1 using (a reformulation of) it.
* added the reformulation pw1dc0el

There are currently some redundant subproofs because sometimes `{ (/) }` is used and sometimes `1o` is used. I propose to use 1o systematically (which involve making its definition available earlier) to avoid redundancies.


avekens (approved):

jkingdon:
> There are currently some redundant subproofs because sometimes `{ (/) }` is used and sometimes `1o` is used. I propose to use 1o systematically

Some days I want to use ` { (/) } ` when we are emphasizing whether it has elements or subsets, and ` 1o ` when making assertions about whether it is an element of something else.

I don't think we've been doing that consistently, and even if we did I'm not sure how satisfying the result would be, so I would lean towards ` 1o ` systematically. It is fewer characters for one thing.

> (which involve making its definition available earlier)

Yeah, that was the other aspect; I wasn't sure whether to move ` 1o ` in order to use it earlier.

jkingdon (approved):

icecream17 (approved):

PR #4118, opened 2024-08-10
Intuitionize products from fprodntriv to fprodss

jkingdon:
Every proof here needs fairly significant changes for things like apartness or decidability but it is also possible to keep the general form of each theorem and structure of its proof.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9305,6 +9305,11 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>~ prodssdc</td>
 </tr>
 
+<tr>
+  <td>fprodss</td>
+  <td>~ fprodssdc</td>
+</tr>
+
 <TR>
   <TD>eftval</TD>
   <TD>~ eftvalcn</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9300,6 +9300,11 @@ intuitionistic and it is lightly used in set.mm</TD>
   <td>~ prodfct</td>
 </tr>
 
+<tr>
+  <td>prodss</td>
+  <td>~ prodssdc</td>
+</tr>
+
 <TR>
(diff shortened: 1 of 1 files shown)


iset.mm
```diff
@@ -123738,6 +123738,249 @@ seq m ( x. , F ) ~~> x ) ) \/
...
+    $d C n $.  $d f k $.  $d f m $.  $d f n $.  $d f ph $.  $d k m $.
+    $d k n $.  $d k ph $.  $d m n $.  $d m ph $.  $d n ph $.  $d A y $.
+    $d B y p $.  $d C y $.  $d f y $.  $d k p y $.  $d m y $.  $d n y p $.
+    $d ph y p $.  $d j p $.  $d f j $.  $d j k $.
+    fprodss.1 $e |- ( ph -> A C_ B ) $.
+    fprodss.2 $e |- ( ( ph /\ k e. A ) -> C e. CC ) $.
+    fprodssdc.a $e |- ( ph -> A. j e. B DECID j e. A ) $.
+    fprodss.3 $e |- ( ( ph /\ k e. ( B \ A ) ) -> C = 1 ) $.
+    fprodss.4 $e |- ( ph -> B e. Fin ) $.
+    $( Change the index set to a subset in a finite sum.  (Contributed by Scott
```
benjub, line 123932:
```suggestion
    $( Change the index set to a subset in a finite product.  (Contributed by Scott
```

benjub (changes requested):

benjub (approved):

iset.mm
benjub: Actually, I'll fix it in iset.mm and set.mm in my upcoming PR.

jkingdon: Sounds good. I hadn't even seen that, good eyes.

PR #4119, opened 2024-08-10
Typos; small iset-FOL restructuring

benjub:
* iset and mmil.raw.html: ax-12 --> ax12 since it it a $p-statement, not an $a-statement. If we want consistency with set.mm labels, then we can create an alias ax12 of ax-12 in set.mm (and rename the current ax12 to ax12from12v).
* iset.mm: since ax12or exists, use it consistently and discourage ax-i12.
* iset.mm: move wcel/wel to the end of FOL for clearer structure.
* iset.mm: typography fixes
* set.mm: mainly typography, and small rewordings (fyi @tirix @avekens in your mathboxes).

### DIFF
--- a/mmnotes.txt
+++ b/mmnotes.txt
@@ -932,7 +932,7 @@ To update your mathbox etc.,
 1. Make sure your mathbox is compatible with the the set.mm just prior
 to this change, temporarily available here:
 
-  http://us2.metamath.org:88/metamath/set.mm.2015-01-08.bz2
+  https://us.metamath.org/metamath/set.mm.2015-01-08.bz2
 
 1. In a text editor, suffix all references to mpbi*an* in your proofs
 with OLD (e.g. mpbiran to mpbiranOLD).  This will make your proofs
@@ -2600,8 +2600,8 @@ Example - proof of theorem divdivdiv
 ------------------------------------
 
 Compare:
-http://us2.metamath.org:88/mpegif/divdivdiv.html - new
-http://us2.metamath.org:88/mpegif/divdivdivOLD.html - old
+https://us.metamath.org/mpegif/divdivdiv.html - new
+https://us.metamath.org/mpegif/divdivdivOLD.html - old
 
 As an example, I re-proved the existing divdivdiv (which I remember as
 being somewhat tedious to achieve a short proof for using the traditional
@@ -3536,7 +3536,7 @@ longer makes sense to subdivide the axioms into separate groups on the
 MPE Home Page, and I combined them into one big table.  I moved the
 description of the "pure" predicate calculus subsystem to the last entry
 of the subsystem table
-http://us2.metamath.org:8888/mpeuni/mmset.html#subsys
+https://us.metamath.org/mpeuni/mmset.html#subsys
 
 -----
 
@@ -3634,7 +3634,7 @@ as follows:
 logical OR expanded into negation and implication.  pm3.26OLD, pm3.27OLD,
 and pm3.28OLD, which will eventually be deleted, are the erroneous
 versions of these.  This error also found its way into pmproofs.txt
-http://us2.metamath.org:8888/mmsolitaire/pmproofs.txt which has also
+https://us.metamath.org/mmsolitaire/pmproofs.txt which has also
 been corrected.  Going through my backups, I found that this error dates
 back to pre-Metamath in the early 90's when I converted my manually
 typed list of 193 Principia Mathematica theorems into the condensed
@@ -3892,7 +3892,7 @@ Space Explorer axiomatization will then allow us to convert all theorems
 to pure ZFC theorems, with no changes to the theorems themselves,
 whenever we are dealing with a fixed Hilbert space (such as complex
 numbers).  This axiomatization change is described in the comment of
-ax-hilex http://us2.metamath.org:8888/mpegif/ax-hilex.html .
+ax-hilex https://us.metamath.org/mpegif/ax-hilex.html .
 
 I probably will not actually make this change in axiomatization but will
 only describe it.  It is very simple to do for anyone interested.  I
@@ -3901,7 +3901,7 @@ Hilbert Space Explorer Home Page easier to describe and it also allows
 us to see what axioms are used to prove specific theorems.
 
 The Hilbert Space Explorer Home Page
-http://us2.metamath.org:8888/mpegif/mmhil.html was updated to mention
+https://us.metamath.org/mpegif/mmhil.html was updated to mention
 this alternate approach (the first 3 paragraphs of "The Axioms"
 section).
 
@@ -4038,7 +4038,7 @@ discussion at http://planetx.cc.vt.edu/AsteroidMeta/closed_and_closure
 (24-May-2007) axnegex and axrecex are now no longer used by any proof,
 and were renamed to axnegexOLD and axrecexOLD for eventual deletion.
 The axiom list at
-http://us2.metamath.org:8888/mpegif/mmcomplex.html#axioms was updated.
+https://us.metamath.org/mpegif/mmcomplex.html#axioms was updated.
 
 A note on theorem names like msqgt0:  a theorem name such as "msqgt0"
 with "msq" (m=multiplication) means "A x. A", while a name such as
@@ -4090,7 +4090,7 @@ it may take a couple of days to formalize.  These kinds of proofs tend
 to be somewhat long, because we can't make use of future theorems that
 depend on the axioms we are trying to prove.  Eventually axnegex and
 axrecex will be eliminated from the official set of complex number
-axioms at http://us2.metamath.org:8888/mpegif/mmcomplex.html, reducing
+axioms at https://us.metamath.org/mpegif/mmcomplex.html, reducing
 the number of axioms from 27 to 25.
 
 
@@ -4352,7 +4352,7 @@ of proof.
 Even if we can't prove ax-11 from ax11a without ax-16 and ax-17, the
 axiom set ax-1 through ax-15 would still be logically complete in the
 sense described at
-http://us2.metamath.org:8888/mpegif/mmzfcnd.html#distinctors .  The
+https://us.metamath.org/mpegif/mmzfcnd.html#distinctors .  The
 deficiency would be that more theorems would have dummy variables in
 their distinctor antecedents, in particular the old ax-11 proved as a
 theorem.  However, in a way this is only of cosmetic importance, since
@@ -4471,7 +4471,7 @@ related to ax-11.
   about ax-11:
 
     ( -. A. x x = y -> ( x = y -> ( ph -> A. x ( x = y -> ph ) ) ) )
-    http://us2.metamath.org:8888/mpegif/ax-11.html
+    https://us.metamath.org/mpegif/ax-11.html
 
   (You may already know some of this.)  Before Juha proved its
   _metalogical_ independence, I spent some time in the other direction,
@@ -4479,7 +4479,7 @@ related to ax-11.
   ax-11, the "distinct variable elimination theorem" dvelimf2 (which
   pleased me at the time):
 
-    http://us2.metamath.org:8888/mpegif/dvelimf2.html
+    https://us.metamath.org/mpegif/dvelimf2.html
 
   that provides a method for converting "$d x y" to the antecedent
   "-. A. x x = y ->" in some cases.  This theorem can be used to derive,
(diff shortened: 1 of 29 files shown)


set.mm
```diff
@@ -748370,9 +748375,9 @@ be a Fermat number (i.e., a Fermat prime), see ~ 2pwp1prmfmtno .
 
   ${
     $d K m n p $.  $d k m p $.
-    $( For every prime number of the form ` ( ( 2 ^ k ) + 1 ) ` ` k ` must be a
-       power of 2, see Wikipedia "Fermat number", section "Other theorms about
-       Fermat numbers", ~ https://en.wikipedia.org/wiki/Fermat_number ,
+    $( For the positive integer ` ( ( 2 ^ k ) + 1 ) ` to be prime, ` k ` must
+       be a power of 2, see Wikipedia "Fermat number", section "Other theorms
```
icecream17, line 748379:
```suggestion
    $( If ` ( ( 2 ^ k ) + 1 ) ` is prime for positive integer ` k ` , ` k ` is
       a power of 2, see Wikipedia "Fermat number", section "Other theorms
```

assuming I'm interpreting the intent of this change correctly, a fix for the case where $$2^{\log_2{6}} + 1$$ is technically a prime positive integer could be like this

iset.mm
```diff
@@ -287,9 +287,9 @@ described in the Metamath book (or the online HELP).
 or a new relationship among symbols that can be eliminated; they always
 extend the definition of a wff or class.  Metamath blindly treats $a
 statements as new given facts but does not try to justify them.  The
-mmj2 program will justify the definitions as sound, except for 5 (df-bi,
-df-cleq, df-clel, df-clab, df-sbc) that require a more complex metalogical
-justification by hand.
+mmj2 program will justify the definitions as sound, except for five of them
+(df-bi, df-cleq, df-clel, df-clab, df-sbc) that require a more complex
+metalogical justification.
```
icecream17, line 292: It doesn't seem like df-sbc is an exception:

https://github.com/metamath/set.mm/blob/e50dd75c9a8fb079cc6dbb3620b787e247699561/.github/workflows/verifiers.yml#L289

but I'm probably missing something

icecream17 (commented):
LGTM though I skimmed, and there might be future commits, but submitting these early since you are currently making commits so it's easier to incorporate

set.mm
icecream17: After more thinking, this is probably not the intent. You may optionally exclude ``for positive integer ` k ` `` since that's rather inherent/obvious, or simply remove `the positive integer` from the start (primes are positive integers so it's redundant) instead of doing my suggestion

iset.mm
benjub: Actually, I blindy copied the analogous part from set.mm. So one of them has to be fixed. Maybe I "fixed" the wrong one... I'm going to check.

set.mm
benjub: I'll do the latter. My concern was purely English style.

iset.mm
benjub: Done: you were right, actually iset.mm was correct and it is set.mm that had to be fixed. (Well, the correct fix would be to simply have set.mm and iset.mm share that information, by including a common file. That's still in my plans, for later.)

tirix (approved):

set.mm
```diff
@@ -17804,32 +17805,32 @@ disjoint variable condition ( ~ sb2 ).  Theorem ~ sb6f replaces the
...
-     that is irrelevant at this point: the predicate calculus axioms for ` e. `
-     apply to any arbitrary binary predicate symbol.  "Non-logical" means that
-     the predicate is presumed to have additional properties beyond the realm
-     of predicate calculus, although these additional properties are not
-     specified by predicate calculus itself but rather by the axioms of a
-     theory (in our case set theory) added to predicate calculus.  "Binary"
-     means that the predicate has two arguments.
-
-     (Instead of introducing ~ wel as an axiomatic statement, as was done in an
+     symbol ` e. ` into our predicate calculus.  We will eventually use it for
```
avekens, line 17816: Why did you remove "(stylized lowercase epsilon)"?

set.mm
```diff
@@ -101234,10 +101237,9 @@ strictly dominates its argument (or the supremum of all ordinals that
 
   $( The aleph function has at least one fixed point.  Proposition 11.18 of
      [TakeutiZaring] p. 104.  See ~ alephfp for an actual example of a fixed
-     point.  Compare the inequality ~ alephle that holds in general.  Note that
-     if ` x ` is a fixed point, then ` aleph `` aleph `` aleph `` ` ...
-     ` aleph `` x = x ` .  (Contributed by NM, 6-Nov-2004.)  (Revised by Mario
-     Carneiro, 15-May-2015.) $)
+     point.  Compare the inequality ~ alephle that holds in general.
+     (Contributed by NM, 6-Nov-2004.)  (Revised by Mario Carneiro,
```
avekens, line 101241: Why did you remove "Note that if ` x ` is a fixed point, then ` aleph `` aleph `` aleph `` ` ... ` aleph `` x = x ` ."?

set.mm
```diff
@@ -148815,8 +148817,8 @@ notation is used by Donald Knuth for iterated exponentiation (_Science_
       QWSFXAVEUSVFVGWRVISSHXBFAVJVKCWQKDVLVMVNWJKVOZBKVPGWOWMRENWPVQVRETVSAWAZT
       KQXJAEVTXKTKWBWCBWDBCKWEWFWGBFHWHWI $.
 
-    $( ` ``' G ` maps the size function's value to ` card ` .  (Contributed by
-       Paul Chapman, 22-Jun-2011.)  (Revised by Mario Carneiro,
+    $( The function ` ``' G ` maps the size function's value to ` card ` .
```
avekens, line 148820: Then "The converse of function ` G ` ..."

set.mm
benjub: It's in the comment of the $c-statement introducing it.  For most other symbols, there is no description of what they look like (or only in a comment of the $c-statement).  It is displayed in the HTML page, for people to see what it looks like.

That said, I can put it back, though the addition looks a bit out of place to me.

set.mm
```diff
@@ -427289,8 +427290,8 @@ All axiomatic assertions ($a statements)
        a wff or class.  Metamath blindly treats $a statements as new
        given facts but does not try to justify them.  The mmj2 program
        will justify the definitions as sound as discussed below,
-       except for 4 definitions ( ~ df-bi , ~ df-cleq , ~ df-clel , ~ df-clab )
-       that require a more complex metalogical justification by hand.
+       except for four of them ( ~ df-bi , ~ df-clab , ~ df-cleq , ~ df-clel )
+       that require a more complex metalogical justification.
```
avekens, line 427294: Why did you remove "by hand"? How atre they justified otherwise?

benjub: I think "by hand" adds nothing to the sentence. (Also, that "more complex metalogical justification" could be made by a human, or by a more complex algorithm in the future, or be formalized in a "metamath-in-metamath" database or in Mario's metamath0... So, removing "by hand" is more about being agnostic about how that justification may be done.)

set.mm
benjub: Done. The reason of that edit was to avoid starting a sentence with a symbol.

set.mm
```diff
@@ -764894,19 +764899,22 @@ According to Wikipedia ("Ackermann function", 8-May-2024,
...
-  ~ ackval3012 : ` A ( 3 , 0 ) = 5 `,  ` A ( 3 , 1 ) = ; 1 3 `,
-                 ` A ( 3 , 3 ) = ; 2 9 `
+  ~ ackval0012 : ` A ( 0 , 0 ) = 1 ` , ` A ( 0 , 1 ) = 2 ` ,
+                 ` A ( 0 , 2 ) = 3 ` ;
+  ~ ackval1012 : ` A ( 1 , 0 ) = 2 ` , ` A ( 1 , 1 ) = 3 ` ,
+                 ` A ( 1 , 3 ) = 4 ` ;
+  ~ ackval2012 : ` A ( 2 , 0 ) = 3 ` , ` A ( 2 , 1 ) = 5 ` ,
+                 ` A ( 2 , 3 ) = 7 ` ;
+  ~ ackval3012 : ` A ( 3 , 0 ) = 5 ` , ` A ( 3 , 1 ) = ; 1 3 ` ,
+                 ` A ( 3 , 3 ) = ; 2 9 ` .
```
avekens, line 764917: Why did you insert line breaks? I thik they are not helpful at this place.

benjub: I agree it's a pity, but after I added the spaces between the backtick and the comma, the lines were 80-character wide, which is more than the 79 maximum.

set.mm
avekens: Since the comment for the  $c-statement is not dsiplayed in the HTML page, I would prefer to have this put back.

set.mm
avekens: I do not know Norm's intention of this, but probably had one. Since it is not wrong, I would like to keep it.

set.mm
avekens: But how are the justified currently, i.e., within or at least for set. mm? If there were not justified yet, most of set.mm would be meaningless.

set.mm
avekens: OK, knowing the reason now I can accept this.

set.mm
icecream17: I suppose to be more specific, we can add a link explaining the current metalogical justification in more detail

set.mm
```diff
@@ -769972,7 +769980,7 @@ _modus ponens_ (our ~ ax-mp ), but he later qualifies this.
 
 $(
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
-  Formally define terms such as Reflexivity
+  Formally define notions such as reflexivity
```
icecream17, line 769983: (optional, unrelated suggestion)
add ` ~ ` before the link at the bottom of this commit
image

icecream17 (approved):

mmnotes.txt
```diff
@@ -4890,7 +4890,7 @@ http://planetx.cc.vt.edu/AsteroidMeta/metamath ).  It has actually been
 modernized slightly, to remove the requirement that the empty set exist.
 This eliminates the need for the Axiom of Replacement, from which
 empty set existence is derived.  The original can be seen at
-http://de2.metamath.org/metamath/set.mm .
+https://de.metamath.org/metamath/set.mm .
```
icecream17, line 4893: It seems like the new link doesn't work

icecream17 (approved):

mmnotes.txt
benjub: Both old and new versions are broken... It's probably lost forever, and not a big loss. I'd simply remove that sentence, but this is in old notes anyway. I'd leave it like that in this MR.

set.mm
benjub: The justification is in Lévy's Basic Set Theory, Appendix, as written in the homepage mmset.html and discussed on the google discussion group multiple times.

Anyway, I'll revert my change since I don't want to spend more time on two words.

benjub:
I reverted both controversial changes, that is, I restored
> Note that if ` x ` is a fixed point, then ` aleph `` aleph `` aleph `` ` ... ` aleph `` x = x ` .

and changed back
> metalogical justification.

to
>metalogical justification by hand.

set.mm
benjub: Done.

set.mm
benjub: Change reverted.

set.mm
benjub: Change reverted.

avekens:
> I reverted both controversial changes, that is, I restored
> 
> > Note that if `x` is a fixed point, then `` aleph `` aleph `` aleph `` `` ... `` aleph `` x = x `` .
> 
> and changed back
> 
> > metalogical justification.
> 
> to
> 
> > metalogical justification by hand.

Sorry, @benjub, for my possibly annoying comments, but I often hesitate and feel uncertain when information provided by Norm is removed. I always fear that we destroy  some part of his legacy. And unfortunately, we do not know about his intention to provide this information, and, alas, we cannot ask him anymore.

jkingdon:
> I often hesitate and feel uncertain when information provided by Norm is removed. I always fear that we destroy some part of his legacy.

I tend to agree with this, at least as a starting point.

I also tend to default to the assumption that it is worth maintaining Norm's style.

I do not take either of these as absolutes - this isn't a static project and Norm never treated it as such when he was involved - but I do hope that we manage to preserve the good parts of what we were left.

mmnotes.txt
jkingdon: Sounds like it is in the git history somewhere, if I'm correctly following. But it doesn't seem especially important to have a specific link in `mmnotes.txt` so just removing the sentence seems fine (at least for now, probably indefinitely).

PR #4120, opened 2024-08-10
Nnnninf2

benjub:
Commit messages say it all.


iset.mm
```diff
@@ -69275,31 +69275,21 @@ property of disjoint unions (see ~ updjud in the case of functions).
 
 $(
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
-  Omniscient sets
+  The one-point compactification of the natural numbers
 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+
+  This section introduces the one-point compactification of the set of natural
+  numbers, introduced by Escardo as the set of nonincreasing sequences on
```
jkingdon, line 69282: Nice. I'm sure Escardo would appreciate this acknowledgement (I've traded a few messages with him) but I've never been sure how to fit that into our format. This seems like a good way to handle it.

jkingdon (approved):

iset.mm
benjub: I hesitated also adding the article to our bibliography. We may add it in the future.

jkingdon: Been a while since I read the paper, but last time I did, it didn't seem to be quite as good a fit for the theorems we had formalized as the things we are citing now.

PR #4122, opened 2024-08-11
iset: nninf

benjub:
* revise infnninf
* add explanation for ~nnexmid, as proposed in https://github.com/metamath/set.mm/pull/4108#issuecomment-2276026363

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -162,6 +162,7 @@
 "hbs1" is used by "mopick".
 "hbs1" is used by "nfs1v".
 "hbs1" is used by "sb9v".
+"infnninfOLD" is used by "fxnn0nninf".
 "mo3h" is used by "mo2dc".
 "mo3h" is used by "mo3".
 "mo3h" is used by "mo4f".
@@ -350,6 +351,7 @@ New usage of "fnexALT" is discouraged (0 uses).
 New usage of "hbs1" is discouraged (5 uses).
 New usage of "idALT" is discouraged (0 uses).
 New usage of "idi" is discouraged (0 uses).
+New usage of "infnninfOLD" is discouraged (1 uses).
 New usage of "mathbox" is discouraged (0 uses).
 New usage of "mo3h" is discouraged (7 uses).
 New usage of "nn0ge2m1nnALT" is discouraged (0 uses).
@@ -490,6 +492,7 @@ Proof modification of "fnexALT" is discouraged (111 steps).
 Proof modification of "idALT" is discouraged (26 steps).
 Proof modification of "idi" is discouraged (1 steps).
 Proof modification of "idref" is discouraged (94 steps).
+Proof modification of "infnninfOLD" is discouraged (127 steps).
 Proof modification of "nn0ge2m1nnALT" is discouraged (48 steps).
 Proof modification of "nnindALT" is discouraged (15 steps).
 Proof modification of "opelopabsbALT" is discouraged (106 steps).


iset.mm
```diff
@@ -69304,18 +69330,36 @@ property of disjoint unions (see ~ updjud in the case of functions).
...
   $}
 
+  ${
+    $d f i j $.
+    $( The point at infinity in ` NN+oo ` is the constant sequence equal to
+       ` 1o ` .  Note that with our encoding of functions, that constant
+       function can also be written as ` ( _om X. { 1o } ) ` , as ~ fconstmpt
+       shows.  (Contributed by Jim Kingdon, 14-Jul-2022.)  Use maps-to
+       notation.  (Revised by BJ, 10-Aug-2024.) $)
+    infnninf $p |- ( i e. _om |-> 1o ) e. NN+oo $=
```
jkingdon, line 69340: I think currently we have a mix of ` ( i e. _om |-> 1o ) ` and ` ( _om X. { 1o } ) `; I'm fine with standardizing on ` ( i e. _om |-> 1o ) `.

iset.mm
```diff
@@ -155301,26 +155342,28 @@ C_ if ( A. k e. suc J ( Q ` ( i e. _om
       = ( G ` ( i e. _om |-> if ( i e. n , 1o , (/) ) ) ) ) $.
     $( Equality of two functions on ` NN+oo ` which agree at every integer and
        at the point at infinity.  From an online post by Martin Escardo.
-       (Contributed by Jim Kingdon, 4-Aug-2023.) $)
+       Remark/TODO: the last two hypotheses can be grouped into one,
```
jkingdon, line 155345: Hmm, the notation that most people use treats the point at infinity quite separately from an element of ` NN+oo ` corresponding to an integer (I think the latter was underlined integers or something in at least one paper, anyway some small adornment around an integer, as opposed to the rather more elaborate combination of maps-to and ` if ` notation we have here).

Not sure that really tells us what to do, but something to think about in terms of whether we really want to combine those two cases.

set.mm
```diff
@@ -7837,7 +7837,8 @@ an illustration of the conservativity of definitions (definitions do not
     ( wn wo wi orordi imor orbi12i 3bitr4ri ) ADZBCEZEKBEZKCEZEALFABFZACFZEKBCG
     ALHOMPNABHACHIJ $.
 
-  $( A wff disjoined with truth is true.  (Contributed by NM, 23-May-1999.) $)
+  $( A disjunction with a true formula is equivalent to that true formula.
```
jkingdon, line 7840: The replacement wording is much clearer and only slightly longer, so I like it.

iset.mm
```diff
@@ -105304,15 +105345,21 @@ equinumerous to omega (the set of natural numbers as ordinals).
     fxnn0nninf.i $e |- I = ( ( F o. `' G )
       u. { <. +oo , ( _om X. { 1o } ) >. } ) $.
     $( A function from ` NN0* ` into ` NN+oo ` .  (Contributed by Jim Kingdon,
-       16-Jul-2022.) $)
+       16-Jul-2022.)  TODO: use ~ infnninf instead of ~ infnninfOLD .  More
+       generally, this theorem and most theorems in this section could use an
+       extended ` G ` defined by
+       ` G = ( frec ( ( x e. ZZ |-> ( x + 1 ) ) , 0 ) u. <. _om , +oo >. ) `
```
jkingdon, line 105351: I went to some effort to provide a similar mechanism in https://us.metamath.org/ileuni/df-ihash.html and as far as I've noticed we are only using the finite case there.

As for these theorems relating ` NN0* ` into ` NN+oo `, I do sort of like the way they illustrate the comparisons between those two, but on the other hand, there aren't any proofs that use them, so it makes it even harder than usual to me to know which version better serves their purpose (because it is pedagogical).

I guess I'm saying I could see going either way about whether to follow up on the suggestion in the "could use" comment here.

jkingdon (approved):
My comments are either agreeing with what is being done, or are non-blocking.

iset.mm
```diff
@@ -6809,16 +6809,36 @@ particular kinds of propositions may be decidable (for example, the
...
     DNLKCOMKLAEFLKGHLIKIJ $.
 
-  $( Double negation of excluded middle.  Intuitionistic logic refutes the
-     negation of excluded middle (but does not prove excluded middle) for any
-     formula.  Can also be proved quickly from ~ bj-nnor as in ~ bj-nndcALT .
+  $( Double negation of decidability of a formula.  See also comment of
+     ~ nnexmid to avoid a pitfall that could come from the label "nnexmid".
+     This theorem can also be proved from ~ bj-nnor as in ~ bj-nndcALT .
      (Contributed by BJ, 9-Oct-2019.) $)
   nnexmid $p |- -. -. ( ph \/ -. ph ) $=
```
icecream17, line 6816: it says "See also comment of ~ nnexmid" but the theorem here is nnexmid

icecream17 (approved):

iset.mm
benjub: Yes, I think it is better when we really mean the **function** and not a particular **set**. Indeed, that the latter set represents the former constant function is an "implementation detail" (even though a standard one in treatments based on set theory).

iset.mm
benjub: You're right, maybe I'm a bit too assertive; I'll change "Remark/TODO" to "Remark".

Escardo's paper uses $\underbar{n}$ for the sequence with $n$ ones followed by zeroes, that is, the image of $n$ under the canonical embedding. He then defines $\underbar{\mathbb{N}}$ as the set of $\underbar{n}$'s.
I think it makes sense to extend this to $\underbar{+\infty}$ (or, since our treatment uses ordinals, $\underbar{\omega}$) the constant sequence always equal to 1. I mean: not introduce a new definition, but treat it on par with finite integers like in the "Remark" above.  Indeed, we want to "prove as much as possible of the resemblance of NNoo* and NNoo" to pinpoint the last pieces that evade constructionism.

iset.mm
benjub: > I went to some effort to provide a similar mechanism in https://us.metamath.org/ileuni/df-ihash.html

This is also what I (would like to) do in https://us.metamath.org/mpeuni/df-bj-iomnn.html.

There is another point: in https://us.metamath.org/ileuni/fxnn0nninf.html and the following three theorems, the function `I : NN0* --> NN+oo` is constructed as follows:
* define  `G = frec ... `
* define ` F = ( n e. _om |-> ... `
* and then, define `` I = ( F o. `' G ) u. { special-couple } ``

that is, we do not know that `I` can be naturally factorized into two natural functions.  Whereas, if we define
* define  ` G = frec ... u. { special-couple' } `
* define ` F = ( n e. suc _om |-> ... `
* and then, define `` I = ( F o. `' G ) ``

then we see that `I` is the composition of two natural functions.  Relatedly, the form of the special couple is simpler in the latter case: ` <. _om , +oo >. ` instead of ` <. +oo , ( _om X. { 1o } ) >. `.

iset.mm
jkingdon: If you are reading set.mm or iset.mm in order, you run into sets of ordered pairs before you run into the maps-to notation, but it is a fair point that in some foundations, functions are fundamental rather than built on sets, so :shrug: . Once we have introduced the maps-to notation, might as well use it for constant functions as well as others.

iset.mm
jkingdon: "Remark" makes sense at least for now. Like I said, I am having trouble coming up with compelling arguments one way or the other.

iset.mm
jkingdon: Yeah, once we've decided to do things with `NN0*` there isn't much point in getting worried about whether the whole concept doesn't make sense.

Adjoining the special couplet to ` G `, as suggested above, probably is a bit nicer than what we have. The key theorem is https://us.metamath.org/ileuni/inftonninf.html and would be provable either way.

jkingdon (approved):

iset.mm
benjub: It is also the case that functions may be defined slightly differently even in a ZF-based foundation: there are treatments (and it is the one I was taught as a freshman), where a function "is" a triple (A, X, B) where $X \subseteq A \times B$ is such that ... That is, the domain and codomain are explicitly given as A and B. In these foundations, the function `( i e. _om |-> 1o )` would be `<. _om , ( _om X. { 1o } ) , 2o >.` (and the codomain 2o should somehow appear in the maps-to notation too). That encoding has the merit of specifying the codomain, which is not the case with our encoding.

jkingdon: The maps-to notation would need to be modified in that case to specify the codomain. This is mentioned at https://en.wikipedia.org/wiki/Function_(mathematics)#Arrow_notation

Anyway, this is getting pretty far afield from our current issue given that we both agree on what to do.

benjub: Yes, that's what I meant above by my "should somehow appear".

No worries: the point of this exchange is not to convince one way is better than the other: just to exchange information.

PR #4123, opened 2024-08-11
Derivative of the logarithm on an interval

metakunt:


icecream17 (approved):

tirix (approved):

icecream17 (approved):
note: I will consider this as a draft pr, but approve anyway to mark progress reviewing

metakunt:
PR is complete from my point of view. I won't make any additions unless otherwise wished.

set.mm
```diff
@@ -638670,6 +638670,162 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+
+  ${
+    $d A x $.  $d A y $.  $d B x $.  $d B y $.  $d ph x $.  $d ph y $.
+    dvrelog3.1 $e |- ( ph -> A e. RR ) $.
+    dvrelog3.2 $e |- ( ph -> B e. RR ) $.
+    dvrelog3.3 $e |- ( ph -> 0 < A ) $.
+    dvrelog3.4 $e |- ( ph -> A <_ B ) $.
+    dvrelog3.5 $e |- F = ( x e. ( A (,) B ) |-> ( log ` x ) ) $.
+    dvrelog3.6 $e |- G = ( x e. ( A (,) B ) |-> ( 1 / x ) ) $.
+    $( The derivative of the logarithm on an open interval (Contributed by
```
benjub, line 638709:
```suggestion
    $( The derivative of the logarithm on an open interval.  (Contributed by
```

set.mm
```diff
@@ -638670,6 +638670,162 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
   $}
 
+  ${
+    $d A x $.  $d A y $.  $d B x $.  $d B y $.  $d ph x $.  $d ph y $.
+    dvrelog2.1 $e |- ( ph -> A e. RR ) $.
+    dvrelog2.2 $e |- ( ph -> B e. RR ) $.
+    dvrelog2.3 $e |- ( ph -> 0 < A ) $.
+    dvrelog2.4 $e |- ( ph -> A <_ B ) $.
+    dvrelog2.5 $e |- F = ( x e. ( A [,] B ) |-> ( log ` x ) ) $.
+    dvrelog2.6 $e |- G = ( x e. ( A (,) B ) |-> ( 1 / x ) ) $.
```
benjub, line 638680: In this and the following theorems:
* define $F$ on $(A, B)$;
* $0 < A$ can be weakened to $0 \leq A$;
* $A \leq B$ is not needed;
* $A \in \mathbb{R}$ and $B \in \mathbb{R}$ could also be removed, but since that fact relies on our encoding details, it is better to keep them.

set.mm
```diff
@@ -638670,6 +638670,162 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+    $( The derivative of the logarithm, ~ ftc2 version.  (Contributed by
```
benjub, line 638681: What does "ftc2 version" mean ?

metakunt: https://us.metamath.org/mpeuni/ftc2.html

set.mm
metakunt: Actually that is dvrelog3. I need to apply the fundamental theorem of calculus on it and in that case A will be 4 or something like that. Since those are just auxiliary lemmas that I need for that specific case and I wasn't aware of any conditions that will be of future use I implemented two lemmas. It's likely that an unused lemma will be deleted once the dependent theorem is formalised.

metakunt: I'm also not sure what you mean by defining F on (A,B) I assume that means the open interval. In that case dvrelog3 is the theorem that implements that one. In any case it is very likely that one of those theorems will be removed.

set.mm
metakunt: Fixed. For some reason commit suggestion doesn't work.

set.mm
icecream17: comparing the statements, dvrelog2 considers the derivative of the logarithm on a specific range, in contrast to dvrelog

So dvrelog2 is a variant of dvrelog where you can apply ftc2, which is on a specific range

metakunt: Actually now that I think about it it explains why I made the first version like it is. It was like that because my first integration proof used https://us.metamath.org/mpeuni/itgparts.html, which was a colossal pain in the ass to get the hypotheses right.

After checking out most derivative maps to builders I have figured out that most operate on an open set.

My proposed solution would be to change itgparts to use open intervals instead of closed intervals.

Usually I was kinda frustrated at the "bad" API I had to work with.

https://us.metamath.org/mpeuni/lcmineqlem12.html#128

This was in my opinion too complicated. And it also used the derivative of the closed interval.

So it might be that both versions are necessary. However I think that you may be able to get away with just open intervals. I guess I'll figure it out

set.mm
```diff
@@ -638670,6 +638670,162 @@ deduction version (Contributed by metakunt, 22-May-2024.) $)
...
+      QWMWLZYQDWMWLZAYRUUAUUCUUDWNZACMSZDMSZYRUUEWOGHCDYQWPWQWRZWSZYTTCYQYTXAAU
+      UFYRGWTUUIATCWKWLYRIWTYTUUAUUCUUDUUHXBXCXDYQXEXFXGXHYCYCXIZXJUUJAUUFUUGXS
+      YBXMUJUJXOUHGHCDXKWQXLWJAFXRFXRUHALUKWGWJ $.
+  $}
+
+  ${
+    $d A x $.  $d A y $.  $d B x $.  $d B y $.  $d ph x $.  $d ph y $.
+    dvrelog3.1 $e |- ( ph -> A e. RR ) $.
+    dvrelog3.2 $e |- ( ph -> B e. RR ) $.
+    dvrelog3.3 $e |- ( ph -> 0 < A ) $.
```
avekens, line 638705: Then @benjub's remark should be applied here:  0 < A can be weakened to 0 ≤ A

benjub:
I haven't looked at the proofs, so I cannot give opinions on the better way to arrive at the final result, which IMO should simply be about log defined on RR+: you do not need to worry about sub-intervals. Since it's in a mathbox, my remarks are non-blocking.

To me, the simplest proof should use the derivative of a composition: if x > 0, then $1 = \mathrm{id}'(x) = (\exp \circ \log)'(x) = \exp'(\log(x)) \log'(x) = \exp(\log(x)) \log'(x) = x.\log'(x)$.

set.mm
benjub: > https://us.metamath.org/mpeuni/ftc2.html

Yes, I had looked at that theorem. However, if there is a theorem xxx, then writing in the comment of theorem yyy "xxx version" is not enlightening. I genuinely do not know what it is meant to mean.

metakunt:
This is actually using https://us.metamath.org/mpeuni/dvrelog.html. I just don't like the shortness of the statement since that makes it kinda awkward for me to use. I prefer maps-to notation since that is usually more ergonomic.

set.mm
metakunt: Fixed it, although I won't really need the full strength of the statement.

set.mm
metakunt: It just meant that the theorem I've formalised is in a form where there it is significantly easier to use with the fundamental theorem of calculus. That's it.

avekens (approved):

PR #4121, opened 2024-08-11
Surreal work

sctfn:
So, I made a mistake when proving a fundamental theorem of surreals a few years ago, and the consequences finally caught up with me. I fixed the theorem (noeta) and proved some consequences

PR #4125, opened 2024-08-12
fix up surreal comment

sctfn:
Take in @icecream17's comments from the previous PR


icecream17 (approved):

icecream17:
Note: Probably wait until @benjub verifies https://github.com/metamath/set.mm/pull/4121#discussion_r1713490038 is resolved before merging (the 'added antecedent' is actually a way to generalize the conclusion)

Edit: diff

```diff
      $( The full-eta axiom for the surreal numbers.  This is the single most
         important property of the surreals.  It says that, given two sets of
         surreals such that one comes completely before the other, there is a
-        surreal lying strictly between the two.  Furthermore, there is an upper
-        bound on the birthday of that surreal.  Alling's axiom FE. (Contributed
-        by Scott Fenton, 7-Dec-2021.) $)
-     noeta $p |- ( ( ( A C_ No /\ A e. V ) /\ ( B C_ No /\ B e. W ) /\
-             A. x e. A A. y e. B x <s y ) ->
+        surreal lying strictly between the two.  Furthermore, if the birthdays
+        of members of ` A ` and ` B ` are strictly bounded above by ` O ` , then
+        ` O ` non-strictly bounds the separator.  Alling's axiom FE.
+        (Contributed by Scott Fenton, 9-Aug-2024.) $)
+     noeta $p |- ( ( ( ( A C_ No /\ A e. V ) /\ ( B C_ No /\ B e. W ) /\
+             A. x e. A A. y e. B x <s y ) /\
+             ( O e. On /\ ( bday " ( A u. B ) ) C_ O ) ) ->
            E. z e. No ( A. x e. A x <s z /\ A. y e. B z <s y /\
-                        ( bday ` z ) C_ suc U. ( bday " ( A u. B ) ) ) ) $=
+                        ( bday ` z ) C_ O ) ) $=
```

benjub:
> ```diff
> +        surreal lying strictly between the two.  Furthermore, if the birthdays
> +        of members of ` A ` and ` B ` are strictly bounded above by ` O ` , then
> +        ` O ` non-strictly bounds the separator.  Alling's axiom FE.
> ```

This is not what the proved theorem states: the proved theorem is an implication and the bound condition `( O e. On /\ ( bday " ( A u. B ) ) C_ O ) )` is an antecedent of that implication. My proposal for the comment in https://github.com/metamath/set.mm/pull/4121#discussion_r1712991115 seems closer to the stated theorem.

If what you write in the comment is true, then it is probably worth formalizing in a separate theorem. That is, you would have two theorems: one, with no assumed bound condition, proving the existence of a surreal with a weak bound, and another, assuming a bound condition, and proving existence of a surreal with a stronger bound. (Actually, it seems that the new "noeta" does not completely subsumes the old "noeta", so having two statements may be a cleaner solution.)

benjub:
The above being said, this is in a mathbox, so I would say this is non-blocking. Something could be added in the form of a "TODO" for instance.

sctfn:
Take a look at noeta2, though. That's the same statement as the old noeta just proving the generalization. I think that covers the case you were talking about, if I'm not mistaken.

benjub:
> Take a look at noeta2, though. That's the same statement as the old noeta just proving the generalization. I think that covers the case you were talking about, if I'm not mistaken.

Indeed, it covers what I wanted: I had missed it. Apart from the fact that its comment lacks a final period and that I think the other comment could describe the formal statement more faithfully, that's good for me.

tirix (approved):

PR #4128, opened 2024-08-13
Excluded middle implies ordinal trichotomy in iset.mm

### LINKED ISSUE #739: Prove ordinal trichotomy in iset.mm given a suitable decidability hypothesis

jkingdon:
Although the result here wasn't in doubt, because set.mm has had ordinal trichotomy for a long time, it wasn't clear how to prove this in iset.mm

Using the set.mm proof would have involved a large number of theorems regarding proper subsets and ordinals and wouldn't have cleanly separated the part of the argument which uses excluded middle and the part which does not.

After trying quite a few dead ends over the years, what eventually succeeded is to note that the HoTT book has a proof of the desired theorem, and that following that proof closely led to success.

The trickiest part is the induction on A and then on B. I hope something fairly general like "nested induction" gives some idea, because I'm not sure I have quite the words to elaborate much in informal mathematical language, other than to say that what ended up succeeding follows the language in the HoTT book fairly closely.

Also removes the temporary hypotheses in [onntri51](https://us.metamath.org/ileuni/onntri51.html) and [onntri52](https://us.metamath.org/ileuni/onntri52.html) because they were just workarounds for not having this theorem yet.

Fixes #739

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -2514,7 +2514,7 @@ is provable.</TD>
 
 <TR>
 <TD>ordtri3or</TD>
-<TD>~ nntri3or </TD>
+<TD>~ nntri3or , ~ exmidontriim</TD>
 <TD>Ordinal trichotomy implies the law of the excluded middle as shown
 in ~ ordtriexmid .
 </TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1552,7 +1552,7 @@ is double negation elimination.</TD>
 
 <TR>
 <TD>r19.30</TD>
-<TD><I>none</I></TD>
+<TD>~ r19.30dc</TD>
 </TR>
 
 <TR>


icecream17 (approved):

iset.mm
```diff
@@ -70653,6 +70661,102 @@ Similar to the Axiom of Choice (first form) of [Enderton] p. 49.
...
+      eleq1 eleq1w imbi2d wa adantl simpll wem wel simplr simpllr simpr rspcdva
+      mpd ralrimiva cbvralv sylib exmidordtriimlemb exp31 tfis2 impcom ) ADJMZN
+      ZDVEOZVEDNZPZDENZDEOZEDNZPJQEVEEOVFVJVGVKVHVLVEEDUAVEEDRVEEDUESAVIJQVEQNZ
+      AVIAVIUBADKMZNZDVNOZVNDNZPZUBZJKJKUCZVIVRAVTVFVOVGVPVHVQJKDUDVEVNDRJKDUFS
+      UGVMVSKVETZAVIVMWAUHZAUHZBCDVEADQNWBFUIVMWAAUJAUKWBHUIACBULCBUCBCULPBQTCD
+      TWBIUIWCDLMZNZDWDOZWDDNZPZLVETDBMZNZDWIOZWIDNZPZBVETWCWHLVEWCLJULZUHZAWHW
+      BAWNUMWOVSAWHUBKVEWDKLUCZVRWHAWPVOWEVPWFVQWGKLDUDVNWDDRKLDUFSUGVMWAAWNUNW
+      CWNUOUPUQURWHWMLBVELBUCWEWJWFWKWGWLLBDUDWDWIDRLBDUFSUSUTVAVBVCVDURGUP $.
+  $}
+
```
avekens, line 70742: I cannot follow the scheme of naming these lemmas: the suffixes are 30, al, b and o - what does this mean? Why not numbering them consecutively starting with *lem1?

jkingdon: Short answer is I don't aim for much of a scheme, but when I name them I generally don't know how many there will be or what order they'll go in.

Having said that I'll admit these names are even less descriptive than usual. I guess I'll rename to use consecutive numbers.

jkingdon: Done.

avekens (approved):

benjub (approved):
> Using the set.mm proof would have involved a large number of theorems regarding proper subsets and ordinals and wouldn't have cleanly separated the part of the argument which uses excluded middle and the part which does not.

Appreciated effort !

> The trickiest part is the induction on A and then on B. I hope something fairly general like "nested induction" gives some idea, because I'm not sure I have quite the words to elaborate much in informal mathematical language, other than to say that what ended up succeeding follows the language in the HoTT book fairly closely.

"Nested induction" gives a good description IMO, and you may probably add it to the comment, e.g.,

"The proof uses nested induction on both ordinals and follows the proof from the HoTT book fairly closely."

> Also removes the temporary hypotheses in [onntri51](https://us.metamath.org/ileuni/onntri51.html) and [onntri52](https://us.metamath.org/ileuni/onntri52.html) because they were just workarounds for not having this theorem yet.

Nice byproduct.

jkingdon:
> "Nested induction" gives a good description IMO, and you may probably add it to the comment

Maybe, but my head went somewhere else with the term nested induction which didn't involve the `A. y e. On` in the `A. z e. A A. y e. On ( z e. y \/ z = y \/ y e. z )` hypothesis to `exmidontriimlem3` and `exmidontriimlem4`.

It is possible that what's really missing here is a tutorial on how to do induction (somewhat in general, but with particular emphasis on the metamath-specific parts of how to set it up), including patterns like nested induction and whatever other variations there are. No, I'm not volunteering to write such a tutorial (unless perhaps if some kind of group effort might produce one).

PR #4129, opened 2024-08-13
Rename syl6rbb to bitr2di

jkingdon:
This is in iset.mm and set.mm.


benjub (approved):
Thanks for that work... and even more so for choosing the correct number of letters ;)

PR #4130, opened 2024-08-14
illustrate "Projective spaces" section header

icecream17:
2 small commits then the main one

~ 40% of the raw size comes from deduction versions of various structure properties: https://github.com/metamath/set.mm/issues/3990

(Just like how complex number theorems now basically all have deduction versions, a similar thing will probably happen for algebraic structures)

---
https://us.metamath.org/mpeuni/mmtheorems398.html#mm39766s
the resulting theorem corresponds to this part of the section header
image

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16195,6 +16195,7 @@ New usage of "flddivrng" is discouraged (2 uses).
 New usage of "fldhmsubcALTV" is discouraged (0 uses).
 New usage of "fnexALT" is discouraged (0 uses).
 New usage of "fnresiOLD" is discouraged (0 uses).
+New usage of "fnsnfvOLD" is discouraged (0 uses).
 New usage of "footexALT" is discouraged (0 uses).
 New usage of "frgrwopreglem5ALT" is discouraged (0 uses).
 New usage of "fsplitOLD" is discouraged (0 uses).
@@ -19681,6 +19682,7 @@ Proof modification of "fimadmfoALT" is discouraged (108 steps).
 Proof modification of "findOLD" is discouraged (70 steps).
 Proof modification of "fnexALT" is discouraged (111 steps).
 Proof modification of "fnresiOLD" is discouraged (25 steps).
+Proof modification of "fnsnfvOLD" is discouraged (74 steps).
 Proof modification of "footexALT" is discouraged (2161 steps).
 Proof modification of "frege10" is discouraged (27 steps).
 Proof modification of "frege100" is discouraged (31 steps).


set.mm
```diff
@@ -641865,10 +641865,29 @@ D Fn ( I ... ( M - 1 ) ) /\
     ifexd.1 $e |- ( ph -> A e. V ) $.
     ifexd.2 $e |- ( ph -> B e. W ) $.
     $( Conditional operator existence.  (Contributed by SN, 26-Jul-2024.) $)
```
benjub, line 641867: I'd write
> Existence of the conditional operator (deduction form).

because it's easier to parse (with the current proposal, it could be read as "(conditional (operator existence))", and because it looks like it is more common (see `!SEARCH *ex exist / COMMENTS`)

set.mm
```diff
@@ -641865,10 +641865,29 @@ D Fn ( I ... ( M - 1 ) ) /\
...
+    MCDEABCKIL $.
+
+  $( Move a conditional outside a relation.  (Contributed by SN,
+     14-Aug-2024.) $)
+  brif2 $p |- ( C R if ( ph , A , B ) <-> if- ( ph , C R A , C R B ) ) $=
+    ( cif wbr iftrue breq2d wn iffalse casesifp ) ADABCFZEGDBEGDCEGAMBDEABCHIAJ
+    MCDEABCKIL $.
+
+  $( Move two conditionals outside a relation.  (Contributed by SN,
+     14-Aug-2024.) $)
```
benjub, line 641885: "inside and outside" (to emphasize it's a biconditional: you can go both ways).

Actually, it seems more natural to write it the other way:
> Move a relation inside and outside the conditional operator.

set.mm
```diff
@@ -642497,6 +642516,44 @@ D Fn ( I ... ( M - 1 ) ) /\
...
   $}
 
+  ${
+    ringassd.b $e |- B = ( Base ` R ) $.
+    ringassd.t $e |- .x. = ( .r ` R ) $.
+    ringassd.r $e |- ( ph -> R e. Ring ) $.
+    ringassd.x $e |- ( ph -> X e. B ) $.
+    ringassd.y $e |- ( ph -> Y e. B ) $.
+    ringassd.z $e |- ( ph -> Z e. B ) $.
+    $( Associative law for the multiplication operation of a ring.
```
benjub, line 642526: multiplication operation --> multiplication

set.mm
```diff
@@ -645588,6 +645803,128 @@ with respect to a linear (1-dimensional) relation.  (Contributed by BJ
...
+      LXDDHVERXRMUJZXTXAXRYBMXCYCXSWTNUHXRMWKZWLYCWMZYCYAXBXRMGYCXSWTIYDWNYEWOW
+      PUGAXAMXCDVEUGAXBMGVHWQWRWS $.
+
+    $d Y l m x y $.  $d Y b $.  $d B m $.  $d I m $.  $d X m $.  $d .x. m $.
+    $d ph m $.  $d S m x y $.
+    prjspner1.y $e |- ( ph -> Y e. B ) $.
+    prjspner1.1 $e |- ( ph -> ( X ` 0 ) =/= .0. ) $.
+    prjspner1.2 $e |- ( ph -> ( Y ` 0 ) =/= .0. ) $.
+    $( Vectors whose first coordinate is nonzero are equivalent iff they
+       intersect the (n-1)-th dimensional space (offset 1 from the origin) at
```
benjub, line 645878: Two vectors whose zeroth coordinate is nonzero are equivalent if and only if they have the same representative in the (n-1)-dimensional affine subspace `{ x_0 = 1 }`.

set.mm
```diff
@@ -645588,6 +645803,128 @@ with respect to a linear (1-dimensional) relation.  (Contributed by BJ
...
+    prjspner01.w $e |- W = ( K freeLMod ( 0 ... N ) ) $.
+    prjspner01.t $e |- .x. = ( .s ` W ) $.
+    prjspner01.s $e |- S = ( Base ` K ) $.
+    prjspner01.0 $e |- .0. = ( 0g ` K ) $.
+    prjspner01.i $e |- I = ( invr ` K ) $.
+    prjspner01.k $e |- ( ph -> K e. DivRing ) $.
+    prjspner01.n $e |- ( ph -> N e. NN0 ) $.
+    prjspner01.x $e |- ( ph -> X e. B ) $.
+    $( Any vector is equivalent to some vector whose value on the 0th axis is
+       either ` .0. ` or ` .1. ` : equivalence part.  (Contributed by SN,
```
benjub, line 645856: Any vector is equivalent to a vector whose zeroth coordinate is `.0. ` or ` .1. ` (proof of the equivalence).


Cannot these two theorems be grouped ? Maybe there are some common proof steps that could be mutualized.

set.mm
```diff
@@ -645588,6 +645803,128 @@ with respect to a linear (1-dimensional) relation.  (Contributed by BJ
...
+    prjspnfv01.w $e |- W = ( K freeLMod ( 0 ... N ) ) $.
+    prjspnfv01.t $e |- .x. = ( .s ` W ) $.
+    prjspnfv01.0 $e |- .0. = ( 0g ` K ) $.
+    prjspnfv01.1 $e |- .1. = ( 1r ` K ) $.
+    prjspnfv01.i $e |- I = ( invr ` K ) $.
+    prjspnfv01.k $e |- ( ph -> K e. DivRing ) $.
+    prjspnfv01.n $e |- ( ph -> N e. NN0 ) $.
+    prjspnfv01.x $e |- ( ph -> X e. B ) $.
+    $( Any vector is equivalent to some vector whose value on the 0th axis is
+       either ` .0. ` or ` .1. ` : value part.  (Contributed by SN,
```
benjub, line 645820: Any vector is equivalent to a vector whose zeroth coordinate is .0. or .1. (proof of the value of the zeroth coordinate).

set.mm
```diff
@@ -645574,6 +645741,54 @@ with respect to a linear (1-dimensional) relation.  (Contributed by BJ
...
+  ${
+    $d W l x y $.  $d B x y $.  $d S l $.  $d .x. l x y $.  $d ph x y $.
+    prjspner.e $e |- .~ = { <. x , y >. |
+      ( ( x e. B /\ y e. B ) /\ E. l e. S x = ( l .x. y ) ) } $.
+    prjspner.w $e |- W = ( K freeLMod ( 0 ... N ) ) $.
+    prjspner.b $e |- B = ( ( Base ` W ) \ { ( 0g ` W ) } ) $.
+    prjspner.s $e |- S = ( Base ` K ) $.
+    prjspner.x $e |- .x. = ( .s ` W ) $.
+    prjspner.k $e |- ( ph -> K e. DivRing ) $.
+    $( The relation in ` PrjSpn ` is an equivalence relation.  (Contributed by
```
benjub, line 645753: The binary relation used to define `PrjSp` in ~ df-prjsp (hence also indirectly `PrjSpn` in ~ df-prjspn ) is an equivalence relation.

set.mm
```diff
@@ -642557,6 +642711,20 @@ D Fn ( I ... ( M - 1 ) ) /\
     KAQFBQFUCUDUGUFUIUCUDUGABCLABCMNUCUEUHJUCUHUEABCUEUHUEOZUHOZPRSUAUEAUJTUHBU
     KTUB $.
 
+  ${
+    frlm0d.f $e |- F = ( R freeLMod I ) $.
+    frlm0d.0 $e |- .0. = ( 0g ` R ) $.
+    frlm0d.r $e |- ( ph -> R e. Ring ) $.
+    frlm0d.i $e |- ( ph -> I e. W ) $.
+    frlm0d.j $e |- ( ph -> J e. I ) $.
+    $( All coordinates in the zero vector are zero.  (Contributed by SN,
```
benjub, line 642720: All coordinates **of**

set.mm
icecream17: This would work, since 33 steps/46 each would be shared, but since there are different hypotheses, I don't really see a good way to combine them. (Well, the only way I can think of is `ph -> ps /\ ch` which seems clunky)

set.mm
```diff
@@ -641864,25 +641864,27 @@ D Fn ( I ... ( M - 1 ) ) /\
   ${
     ifexd.1 $e |- ( ph -> A e. V ) $.
     ifexd.2 $e |- ( ph -> B e. W ) $.
-    $( Conditional operator existence.  (Contributed by SN, 26-Jul-2024.) $)
+    $( Existence of the conditional operator (deduction form).  EDITORIAL: If
+       this theorem is moved to main, use it to shorten ~ ifex .  (Contributed
+       by SN, 26-Jul-2024.) $)
```
benjub, line 641869: It would be nice if in a later MR, you can move ifexd to Main and see which order of proving ifcl(d) and ifex(d|g|) makes the overall proof shorter.
Actually, it looks like there is a theorem missing, which would be
```
|- ( ( A e. U /\ B e. V ) -> if ( ph , A , B ) e. ( U u. V ) )
```
and its deduction form too.

set.mm
benjub: Oh, I hadn't seen your already existing ~prjsper, which explains your wording. Anyway, the links you added help the reader connect the dots, so I think they're useful.

set.mm
```diff
@@ -645874,12 +645897,12 @@ with respect to a linear (1-dimensional) relation.  (Contributed by BJ
...
     prjspner1.2 $e |- ( ph -> ( Y ` 0 ) =/= .0. ) $.
-    $( Vectors whose first coordinate is nonzero are equivalent iff they
-       intersect the (n-1)-th dimensional space (offset 1 from the origin) at
-       the same point.  For example, vectors in 3D space whose ` z ` coordinate
-       is nonzero are equivalent iff they intersect at the plane ` z = 1 ` at
-       the same point (where ` z ` is the first coordinate; also see section
-       header).  (Contributed by SN, 13-Aug-2023.) $)
+    $( Two vectors whose zeroth coordinate is nonzero are equivalent if and
+       only if they have the same representative in the (n-1)-dimensional
+       subspace { x_0 = 1 } .  For example, vectors in 3D space whose ` x `
```
benjub, line 645902: **affine** subspace

(Because if you simply write "subspace", it means "vector subspace", that is, a subset that is a vector space for the induced structure, and in particular contains the zero vector.)

set.mm
benjub: Then maybe there is a common lemma that can be extracted ? Anyway, that's only a suggestion for a later PR. Resolving.

benjub (approved):

set.mm
benjub: @icecream17 : I'll do this unless you tell me you are working on it.  Are you ?

benjub: ifexd moved to Main in #4139. I took a note to add `|- ( ( A e. U /\ B e. V ) -> if ( ph , A , B ) e. ( U u. V ) )`.

Resolving.

PR #4133, opened 2024-08-15
Add some distributive theorems for class difference

BTernaryTau:
Adds a few theorems for distributing intersections and restrictions over differences. Also shortens indifdir using indifdi.

indifdi

> Distribute intersection over difference.

resdifdi and resdifdir

> Distributive law for restriction over difference.

### DIFF
--- a/discouraged
+++ b/discouraged
@@ -16811,6 +16811,7 @@ New usage of "in2an" is discouraged (1 uses).
 New usage of "in3" is discouraged (13 uses).
 New usage of "in3an" is discouraged (1 uses).
 New usage of "incomOLD" is discouraged (0 uses).
+New usage of "indifdirOLD" is discouraged (0 uses).
 New usage of "indistps2ALT" is discouraged (0 uses).
 New usage of "indistpsALT" is discouraged (0 uses).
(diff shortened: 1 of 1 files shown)


icecream17 (approved):

metakunt:
Well I guess since it was said to me a few times. Might wanna put those in your mathbox first.

benjub:
@metakunt you're right that the default is to put it in the mathbox first. But in this specific case, I think indifdi can indeed go directly to Main, with the shortening it permits (in addition to its interest per se).

As for resdifdi(r), I would say "yes": the statements look natural enough to go to Main.

benjub (approved):

BTernaryTau:
I chose to place these theorems in main because they are similar to existing theorems there ([difindi](https://us.metamath.org/mpeuni/difindi.html), [resundi](https://us.metamath.org/mpeuni/resundi.html), [resindi](https://us.metamath.org/mpeuni/resindi.html), etc.) and because I am planning to use resdifdi in a revision of [dif1en](https://us.metamath.org/mpeuni/dif1en.html) to remove its dependency on ax-pow.

avekens (approved):

PR #4134, opened 2024-08-15
(i)set: ifssun ; iset: ifex(|d|g)

benjub:
FYI @jkingdon : I think it was important to add ifex(|d|g), even though you indicated in the index that you had let it on the side. It may help remove decidability hypotheses in some usages of ifcld(a)cd, but I haven't looked.

### DIFF
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -318,7 +318,7 @@ New usage of "bdsepnfALT" is discouraged (0 uses).
 New usage of "bj-axempty" is discouraged (0 uses).
 New usage of "bj-axempty2" is discouraged (0 uses).
 New usage of "bj-axemptylem" is discouraged (2 uses).
-New usage of "bj-charfunALT" is discouraged (0 uses).
+New usage of "bj-charfundcALT" is discouraged (0 uses).
 New usage of "bj-el2oss1o" is discouraged (0 uses).
 New usage of "bj-nn0sucALT" is discouraged (0 uses).
 New usage of "bj-nndcALT" is discouraged (0 uses).
@@ -422,7 +422,7 @@ Proof modification of "bj-axun2" is discouraged (47 steps).
 Proof modification of "bj-bdfindes" is discouraged (121 steps).
 Proof modification of "bj-bdfindis" is discouraged (84 steps).
 Proof modification of "bj-bdfindisg" is discouraged (44 steps).
-Proof modification of "bj-charfunALT" is discouraged (194 steps).
+Proof modification of "bj-charfundcALT" is discouraged (194 steps).
 Proof modification of "bj-d0clsepcl" is discouraged (111 steps).
 Proof modification of "bj-dcstab" is discouraged (19 steps).
 Proof modification of "bj-el2oss1o" is discouraged (23 steps).
--- a/iset-discouraged
+++ b/iset-discouraged
@@ -318,6 +318,7 @@ New usage of "bdsepnfALT" is discouraged (0 uses).
 New usage of "bj-axempty" is discouraged (0 uses).
 New usage of "bj-axempty2" is discouraged (0 uses).
 New usage of "bj-axemptylem" is discouraged (2 uses).
+New usage of "bj-charfunALT" is discouraged (0 uses).
 New usage of "bj-el2oss1o" is discouraged (0 uses).
 New usage of "bj-nn0sucALT" is discouraged (0 uses).
 New usage of "bj-nndcALT" is discouraged (0 uses).
@@ -421,6 +422,7 @@ Proof modification of "bj-axun2" is discouraged (47 steps).
 Proof modification of "bj-bdfindes" is discouraged (121 steps).
 Proof modification of "bj-bdfindis" is discouraged (84 steps).
 Proof modification of "bj-bdfindisg" is discouraged (44 steps).
+Proof modification of "bj-charfunALT" is discouraged (194 steps).
 Proof modification of "bj-d0clsepcl" is discouraged (111 steps).
 Proof modification of "bj-dcstab" is discouraged (19 steps).
 Proof modification of "bj-el2oss1o" is discouraged (23 steps).
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1875,7 +1875,7 @@ csbifgOLD</TD>
 
 <TR>
   <TD>ifid</TD>
-  <TD>~ ifiddc</TD>
+  <TD>~ ifidss , ~ ifiddc</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1848,7 +1848,7 @@ is double negation elimination.</TD>
 
 <TR>
 <TD>ifclda</TD>
-<TD>~ ifcldacd</TD>
+<TD>~ ifcldadc</TD>
 </TR>
 
 <TR>
@@ -1885,7 +1885,7 @@ csbifgOLD</TD>
 
 <TR>
   <TD>ifcl , ifcld</TD>
-  <TD>~ ifcldcd , ~ ifcldacd</TD>
+  <TD>~ ifcldcd , ~ ifcldadc</TD>
 </TR>
 
 <TR>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1903,11 +1903,6 @@ favor of theorems in deduction form.</TD>
 </TR>
 
 <TR>
-  <TD>ifex , ifexg</TD>
-  <TD>~ ifcldcd , ~ ifcldadc</TD>
-</TR>
-
-<TR>
   <TD>ifpr</TD>
   <TD><I>none</I></TD>
   <TD>Should be provable if the condition is decidable.</TD>
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -1848,7 +1848,7 @@ is double negation elimination.</TD>
 
 <TR>
 <TD>ifclda</TD>
-<TD>~ ifcldadc</TD>
+<TD>~ ifcldacd</TD>
 </TR>
 
 <TR>
@@ -1885,7 +1885,7 @@ csbifgOLD</TD>
 
 <TR>
   <TD>ifcl , ifcld</TD>
-  <TD>~ ifcldcd , ~ ifcldadc</TD>
+  <TD>~ ifcldcd , ~ ifcldacd</TD>
 </TR>
 
 <TR>


iset.mm
```diff
@@ -30751,11 +30768,11 @@ intersection or the difference (that is, this theorem would be equality
   $}
 
   ${
-    ifcldadc.1 $e |- ( ( ph /\ ps ) -> A e. C ) $.
-    ifcldadc.2 $e |- ( ( ph /\ -. ps ) -> B e. C ) $.
-    ifcldadc.dc $e |- ( ph -> DECID ps ) $.
+    ifcldacd.1 $e |- ( ( ph /\ ps ) -> A e. C ) $.
```
jkingdon, line 30771: What are you thinking with this naming? `ifcldadc` is `ifclda` from set.mm plus `dc`. I've often wondering whether it should be `ifcldcda` but I don't know where `ifcldacd` is coming from.

iset.mm
```diff
@@ -39258,6 +39275,30 @@ rather than an implication (but the two are equivalent by ~ bm1.3ii ),
       VHABVFBJZVTVHWDVTTVFVGRZVGVFRZTVHWDWEVTWFVFBUQVGVFURVAVFVGUSUTVBVCVDVE $.
   $}
 
+  $( Existence of a conditional set, quantitative version (closed form).
+     (Contributed by BJ, 15-Aug-2024.) $)
+  ifexg $p |- ( ( A e. V /\ B e. W ) -> if ( ph , A , B ) e. ~P ( A u. B ) ) $=
```
jkingdon, line 39280: I wouldn't call these `ifexg`, `ifexd`, and `ifex`, both because of "Theorems which are different [between set.mm and iset.mm] should be named differently" from https://us.metamath.org/ileuni/conventions.html but also because it is a pretty well established convention in both set.mm and iset.mm that `ex` means set existence, not some flavor of closure (or whatever the word is for `e. ~P ( A u. B )` as opposed to `e. _V`).

I realize that you wanted the stronger theorem to prove `fmelpw1o`, which does seem like a noble goal, but a name like `ifpw` or `ifelpw` would seem fine. Also, given the limited usage of these, I think we probably just need one (the `g` or `d` one, I don't have much of an opinion between those two) rather than three.

Separately, it would be fine to have `ifex` and `ifexd` stated exactly as in set.mm. As far as I could tell from git logs, I didn't have any reason for leaving them out other than that the set.mm proof doesn't work as-is and I didn't have a need for them (well, that I noticed - as you pointed out in the description of this pull request there might be places they could shorten some existing proofs).

iset.mm
```diff
@@ -152516,14 +152557,45 @@ which membership in it is a bounded proposition ( ~ df-bdc ).
       ZADFKZBCLTAUABCHMBCDSFSNOPACESGQR $.
   $}
 
+  $( With a formula ` ph ` one can associate an element of ` ~P 1o ` , which
+     can therefore be thought of as the set of "truth values" (but recall that
+     there are no other genuine truth values than ` T. ` and ` F. ` , by
+     ~ nndc , which translate to ` 1o ` and ` (/) ` respectively by ~ iftrue
+     and ~ iffalse , giving ~ pwtrufal ).  (Contributed by BJ, 15-Aug-2024.) $)
+  fmelpw1o $p |- if ( ph , 1o , (/) ) e. ~P 1o $=
```
jkingdon, line 152565: We'd also be able to prove `if ( ph , 1o , (/) ) = { x e. 1o | ph }`, right?

(Not a blocking comment, and no need to add it to this pull request unless you want, just an observation).

iset.mm
benjub: I'm not sure what happened in my mind... I'm going to revert !

jkingdon: I could try to guess what happened (`ifcldadc` is a bit of a strange name), but reverting sounds fine.

iset.mm
benjub: Ok, I'm going to change the labels. Do you have some proposals ? Maybe ifsspw(|d|g) ?

Then, I'll prove the standard form of ifex(|d|g) from them.

I had chosen the same labels, because to me, this was almost the same as existence: when existence is needed, it is often of the for `... e. V ` where ` V ` is a metavariable, so any version of existence would do.

"Quantitative" versions are  nice because they permit to work inside various kinds of universes: if you have a universe closed under unions and powersets, then you have existence in that universe.

iset.mm
benjub: We may also add it indeed. With `fmelpw1o`, I wanted to expose $\mathcal{P}1$ as a particularly important set.

set.mm
```diff
@@ -427517,18 +427526,18 @@ it can be understood as claiming either uniqueness ( ` E* ` ) or unique
...
-       however, ` |- A e. _V ` is used rather than ` |- A e. V ` (e.g.,
-       ~ difexi ). This is because ` A e. _V ` is almost always satisfied using
-       an existence theorem stating "... ` e. _V `", and a hard-coded ` _V ` in
+       variable if ` V ` is not available) to make it more general.  That way
+       we can often avoid extra uses of ~ elex and ~ syl in the common case
+       where ` I ` is already a member of something.  For hypotheses
+       ($e statement) of theorems (mostly in inference form), however,
+       ` |- A e. _V ` is used rather than ` |- A e. V ` (e.g., ~ difexi ).
+       This is because ` A e. _V ` is almost always satisfied using an
+       existence theorem stating " ` ... e. _V ` ", and a hard-coded ` _V ` in
```
avekens, line 427540: Hm, the "symbol" ` ... ` does not mean "ellipsis", but is a "notation for a contiguous finite set of integers". Therefore, I am not sure to include it in the Metamath expression.

There are already cases in set.mm, however, which use ` ... ` as ellipsis in Methamath expressions, e.g., in the header comment of "Class abstractions":

>  1.  The axiomatic statement ~ df-clab does not define the class
>      abstraction ` { y | ph } ` itself, that is, it does not have the form
>      ` |- { y | ph } = ... ` that a standard definition should have (for a good...

And ´this does not look bad in HTML:
grafik

Therefore, I have no objections against this change, although I think the previous version was also acceptable:
grafik

benjub: You're right: I noticed too that "semantically", using "..." as a math symbol in the sense of an ellipsis is not correct, but as you write, it's better typographically, and it's only in comments, where the meaning is clear.

In the present case, you're right that there is no big change between both versions.  In more complex formulas, as a general rule I try to avoid "switching from text-mode to math-mode" back and forth, e.g.,
`` ( 3 ` x.` 4 ) ` +. ` ... ` +. ` 2 ``
instead of simply
`` ` ( 3 x. 4 ) +. ... +. 2 ` ``.

iset.mm
benjub: Done. I chose to add the deduction version only, but I think the other versions (closed form and inference form) could have been added too. In general, I think we should not shy away from adding these forms: currently, the more limited library of lemmas in iset.mm makes it harder (at least for me) to prove things than in set.mm, because the proof assistant "matches" fewer possibilities (I mean: the `IMPROVE ALL` of `MM-PA` or the `Ctrl+U` of `mmj2` are less likely to finish the proof for you).

iset.mm
benjub: Done. I added the general instance of that theorem.

iset.mm
benjub: Done.

avekens (approved):

set.mm
avekens: Maybe we can add this kind of usage to the comment of ` cfz $a class ... $. `, for example

> $( Extend class notation to include the notation for a contiguous finite set of integers.  Read " ` M ... N ` " as "the set of integers from ` M ` to ` N ` inclusive."  Within comments, this symbol is also used as ellipsis, e.g., `` ` ( 3 x. 4 ) +. ... +. 2 ` ``. $)

icecream17 (approved):

set.mm
benjub: Good idea. Done.

benjub:
(Sorry, I shouldn't keep adding independent minor commits after approvals... I'm done.)

set.mm
```diff
@@ -458949,8 +458949,8 @@ Positive operators (cont.)
     pjima.1 $e |- A e. SH $.
     pjima.2 $e |- B e. CH $.
     $( The image of a projection.  Lemma 5 in Daniel Lehmann, "A presentation
-       of Quantum Logic based on an _and then_ connective"
-       ~ http://www.arxiv.org/pdf/quant-ph/0701113 p. 20.  (Contributed by NM,
+       of Quantum Logic based on an _and then_ connective",
+       ~ https://doi.org/10.48550/arXiv.quant-ph/0701113 .  (Contributed by NM,
```
icecream17, line 458953: deletes page 20?

icecream17 (approved):

set.mm
benjub: Yes. It's more standard to cite that way ("Lemma 5 of [xxx]") and the page is unnecessary (it is also less stable than the lemma number, now that a given paper can be available in various formats).

jkingdon:
> (Sorry, I shouldn't keep adding independent minor commits after approvals... I'm done.)

Agreed. I probably shouldn't have even suggested that as a possibility.

I'll try to look over this pull request at some point but it is a lot to try to keep up with.

benjub:
> > (Sorry, I shouldn't keep adding independent minor commits after approvals... I'm done.)
> 
> Agreed. I probably shouldn't have even suggested that as a possibility.

No no, I was not referring to your remark, which was useful: I was referring to my minor typographical commits, which could have waited another MR.
 
> I'll try to look over this pull request at some point but it is a lot to try to keep up with.

Thanks. Actually, since your first approval, there is little real change.

iset.mm
jkingdon: There isn't any intentional divergence between iset.mm and set.mm on this point. If it is a frequently used theorem, it is fine to have multiple forms.

set.mm
jkingdon: Oh, I didn't know arxiv had that issue too. It is very much an issue with the HoTT book, which has been available in different formats (with different page numbers, but the same theorem numbers) from the start.

jkingdon:
> > I'll try to look over this pull request at some point but it is a lot to try to keep up with.
> 
> Thanks. Actually, since your first approval, there is little real change.

Kind of irrelevant now, but I had never approved it until now.

jkingdon (approved):

set.mm
benjub: Well, maybe I wasn't precise enough: if you select a given version of the arXiv preprint and a given format, say PDF, then the page does not change. But you now you have html versions a newer preprints, where page numbers make no sense, and you may also land via internet search on the journal version, which will most probably have different page numbers, while the lemma/proposition/theorem number is more likely (if the author did put the corrected version on the arXiv) to be the same.
Anyway, these are minor details.

jkingdon: Ah got it, somewhat similar results but due to different mechanisms (in the HoTT Book case, the scripts literally generate multiple formats - PDFs with different page sizes - from the same LaTeX source).

Issue #3483, opened 2023-09-15
Contribution principles


jkingdon:
Managing a collaborative project such as metamath involves, among other things, acknowledging contributions and we have a variety of ways of doing that. This has led to quite a variety of different perspectives and uncertainty about how to proceed in a number of contexts. Some of what I say here was prompted by specific things which happened recently but others have been vexing me (and probably others) for much longer.

My proposal is to write down a few things (e.g. about git logs and the contributor list at the top of set.mm) which I think were clear in some people's minds but maybe not everyone. But also to write a list of principles, on the theory that if we have at least some level of shared understanding about what we are trying to achieve, perhaps that will make it a bit easier for the specific conventions to be applied or refined than if we just have conventions alone.

I suppose if we decide we want this text it would go in CONTRIBUTING.md but whether there is some other place is also up for discussion.

> We acknowledge people who contribute to these proofs in various ways, including
comments on each theorem, statements on the web site, git logs, and others.

> Although there are conventions relating to many of these, they are guided
by some principles:

> 1. We want to make everyone who contributes to metamath to feel appreciated.
This goes double for new contributors! In many cases, their experiences
upon attempting their first submission will determine whether they continue
to participate.

> 2. Although we want every contributor to be acknowledged in some way, where
and how this is done is more flexible.  In some cases contribution
will be mentioned along with the text or proofs that was contributed,
in others it will be in separate historical records.

> 3. A lot of people have built set.mm and iset.mm, going back decades, and
this is worth celebrating! From lists of people to graphical visualizations,
we love showing off our current and past contributors.

> Mechanisms:

> * There is a contributor list at the top of set.mm and iset.mm.
If you find someone with a Contributed, Revised by, or Proof shortened
by line who is not in the contributor list, either it is a mistake
or they are so far undercover that their identity must remain as
an easter egg to be discovered.

> * The git history.  This is the most historical of our records, but is
also hard to revise if something confusing happened during merging.
For that reason if a submission is going to be accepted and then revised,
it is usually better either for the original author to make the
revisions, or merge the original pull request followed a separate pull
request (possibly from a different person) with the revisions.
By "hard to revise" we mean the only practical way to change git history,
once a change has been merged, is to make a commit reverting
that change, followed by another one with new content or different
authorship.

> * The Contributed by, Revised by, and Proof shortened by comments are described
in the Metamath book and https://us.metamath.org/mpeuni/conventions-comments.html

> * The Metamath 100 list gives an author or a small list next to each
[...]

PR #4139, opened 2024-08-18
Move ifexd to Main.

benjub:
* Minor shortening of iset/cnstab.
* Move ifexd to Main and minimize ifexg with it, as announced in https://github.com/metamath/set.mm/pull/4130#discussion_r1719654644

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -115,6 +115,7 @@ Date      Old       New         Notes
 18-Aug-24 reldmprds [same]      revised - compatibility with new df-prds
 18-Aug-24 df-prds   [same]      revised - fixed the comp part
 18-Aug-24 difexd    [same]      moved from SN's mathbox to main set.mm
+18-Aug-24 ifexd     [same]      moved from SN's mathbox to main set.mm
 13-Aug-24 syl6rbb   bitr2di     compare to bitr2i or bitr2d
 12-Aug-24 disjdifr  [same]      moved from TA's mathbox to main set.mm
  7-Aug-24 tdeglem4  [same]      revised - eliminated unnecessary antecedent


icecream17 (approved):

avekens:
>     * Move ifexd to Main and minimize ifexg with it, as announced in [illustrate "Projective spaces" section header #4130 (comment)](https://github.com/metamath/set.mm/pull/4130#discussion_r1719654644)

Are ther other prrof which can be shortened using ifexd?

avekens (approved):

tirix (approved):

avekens (changes requested):
Ups, I forgot something: the movement should be documented in changes-set.txt

benjub:
> Ups, I forgot something: the movement should be documented in changes-set.txt

It's in the commit message, which is what matters for tracking history. It's not breaking anything in Main, so no need to edit the `changes` file.

avekens:
> > Ups, I forgot something: the movement should be documented in changes-set.txt
> 
> It's in the commit message, which is what matters for tracking history. It's not breaking anything in Main, so no need to edit the `changes` file.

It is an established practice for a long time that theorems which were moved from (at least other's) maxboxes to main are documented in changes-set.txt (see, for example, https://groups.google.com/g/metamath/c/dGW5v81srRY/m/-u_Qq1YCAwAJ in the Google group). I do not know the exact reason for it (anymore), but I think it was to give a hint for the mathbox owner why a theorem seems to be missing in his mathbox. Looking up the git history for such changes is very tedious, escpecially because we have a single file set.mm with an enormous amount of changes.

Unfortunately, this seems not to be documented anywhere (neither in changes-set.txt itself nor in set.mm - here we only have "For a chronological list of changes to label names and label deletions, see the changes-set.txt file.  This should help if you have a proof not checked into the main repository and want to update it for recent changes").

icecream17:
since I know that it was moved I'm indifferent to this change

avekens:
This rule is also mentioned in the proposed "Contribution principles" by @jkingdon , see issue #3483.

jkingdon:
> It is an established practice for a long time that theorems which were moved from (at least other's) maxboxes to main are documented in changes-set.txt

Yeah, I've just taken it for granted that this happens. I don't really know if there has been a lot of discussion on it (I don't remember any).

benjub:
So... we're just following a rule because it's a rule even though we do not know why it exists ?

avekens (approved):
Thanks!

tirix:
> So... we're just following a rule because it's a rule even though we do not know why it exists ?

I think the rule is _really_ useful for renames like `syl6bbr` to `bitr4di`. It's also useful when a theorem is renamed at the same time as it is moved to main. In both cases, in order to adapt pending changes to the changes already in the `develop` branch.

Other than that, I find it entertaining and interesting to see _which_ theorems are moved to main (e.g. there is a big recent trend to move deduction variants), but I don't think it's absolutely necessary.

The table and the rule must have originated from the time when Git was not used yet and there was no change tracking. Git makes it less useful for sure.

PR #4141, opened 2024-08-19
Intuitionize finite products from fprodser to fprodshft

jkingdon:
The biggest difference from set.mm is that using `seq` for finite sets involves significant adjustments (typically an additional `if` expression and related changes). Some of these proofs make those adjustments. Others replace a proof using `seq` with one which performs induction using the product notation directly.

For proofs which hadn't used `seq` they intuitionize pretty straightforwardly.

### DIFF
--- a/mmil.raw.html
+++ b/mmil.raw.html
@@ -9311,6 +9311,12 @@ intuitionistic and it is lightly used in set.mm</TD>
 </tr>
 
 <tr>
+  <td>fproddiv</td>
+  <td><i>none</i></td>
+  <td>presumably could be proved with not equal changed to apart</td>
+</tr>
+
+<tr>
   <td>fprodsplit</td>
   <td>~ fprodsplitdc</td>
 </tr>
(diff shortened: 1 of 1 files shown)


tirix (approved):

iset.mm
```diff
@@ -124272,6 +124272,528 @@ seq n ( x. ,
...
+      VHULZXEUVNBUVJVHVKWGWNWOWPWQWRWSWT $.
+  $}
+
+  ${
+    $d K j k $.  $d M j k $.  $d N j k $.  $d Z k $.  $d ph j k $.
+    fprodeq0.1 $e |- Z = ( ZZ>= ` M ) $.
+    fprodeq0.2 $e |- ( ph -> N e. Z ) $.
+    fprodeq0.3 $e |- ( ( ph /\ k e. Z ) -> A e. CC ) $.
+    fprodeq0.4 $e |- ( ( ph /\ k = N ) -> A = 0 ) $.
+    $( Anything finite product containing a zero term is itself zero.
```
avekens, line 124755: Anything -> Any?

jkingdon: Fixed in both set.mm and iset.mm.

avekens (approved):

PR #4146, opened 2024-08-20
Surreals 3

sctfn:
Beginning work on the recursive surreal definitions here.  Mostly proving theorems about recursion and induction, but I did take a stop off and prove some basic properties about surreal addition!

Moved brsnop up to main

### DIFF
--- a/changes-set.txt
+++ b/changes-set.txt
@@ -99,6 +99,7 @@ make a github issue.)
 
 DONE:
 Date      Old       New         Notes
+19-Aug-24 rabeqdca  [same]	moved from SN's mathbox to main set.mm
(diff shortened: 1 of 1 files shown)


icecream17 (approved):
the `First` `Next` `Now` `Finally` pattern is unexpected but it makes sense, nice results!

tirix (approved):
Really nice! On the way to show that surreal numbers form an (ordered) field!
Which proof assistant do you use?

sctfn:
I actually use the MPA plus emacs. I’m a bit old fashioned 😁On Aug 21, 2024, at 6:38 AM, tirix ***@***.***> wrote:﻿
@tirix approved this pull request.

Really nice! On the way to show that surreal numbers form an (ordered) field!
Which proof assistant do you use?

PR #4149, opened 2024-08-21
Fermat's last theorem for exponent 4

icecream17:
<s>Draft <s>since I'm avoiding a merge conflict (deduction of [opthpr](https://us.metamath.org/mpeuni/opthpr.html) will be where ifexd was: #4139)</s>, but everything is proved except for the lemmas</s> Done

<s>In changes-set.txt I replaced two tabs with spaces</s> fixed by #4160

**Edit**: ~~flt4lem5 is impossible to prove. flt4g will become a lemma with an extra `A < B` hypothesis, and flt4 will become nontrivial (will use fltne)...~~

~~**edit 2:** `A < B` is not very useful considering the relevant terms are `A ^ 4` and `B ^ 2`, so instead the parity antecedent from [pythagtriplem18](https://us.metamath.org/mpeuni/pythagtriplem18.html) will be carried over~~

**edit 3**: A [better proof](https://github.com/metamath/set.mm/pull/4149#issuecomment-2305234256) was used


metakunt:
Nice, how much harder is exponent 3?

set.mm
```diff
@@ -647669,29 +647743,326 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+  $}
+
+  ${
+    fltabcoprm.s $e |- ( ph -> S e. NN ) $.
+    fltabcoprm.a $e |- ( ph -> A e. NN ) $.
+    fltabcoprm.b $e |- ( ph -> B e. NN ) $.
+    fltabcoprm.c $e |- ( ph -> C e. NN ) $.
+    fltabcoprm.n $e |- ( ph -> N e. NN0 ) $.
+    fltabcoprm.1 $e |- ( ph -> ( ( A ^ N ) + ( B ^ N ) ) = ( C ^ N ) ) $.
+    $( A counterexample to FLT implies a counterexample to FLT with ` A , B `
```
benjub, line 647808: At this point, you do not know that you obtain a Fermat or Pythagorean triple, since you do not know whether $A \wedge B | C$ (for Pythagorean triples, it turns out to be the case; I do not know in general, but in any case this theorem does not state it).

set.mm
```diff
@@ -647669,29 +647743,326 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+    flt4lem.z $e |- ( y = z -> ( ps <-> th ) ) $.
+    flt4lem.s $e |- ( ph -> S C_ ( ZZ>= ` M ) ) $.
+    flt4lem.1 $e |-
+        ( ( ph /\ ( x e. S /\ ch ) ) -> E. z e. S ( th /\ z < x ) ) $.
+    $( Infinite descent.  The hypotheses say that ` S ` is lower bounded, and
+       that if ` ps ` holds for an integer in ` S ` , it holds for a smaller
+       integer in ` S ` .  By infinite descent, eventually we cannot go any
+       smaller, therefore ` ps ` holds for no integer in ` S ` .  (Contributed
+       by SN, 20-Aug-2024.)  $)
+    flt4lem $p |- ( ph -> { y e. S | ps } = (/) ) $=
```
benjub, line 647850: It's more than a lemma.

set.mm
```diff
@@ -647669,29 +647743,326 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+  ${
+    $d ph k m n $.  $d A k m n $.  $d B k m n $.  $d C k m n $.
+    flt4lem2.a $e |- ( ph -> A e. NN ) $.
+    flt4lem2.b $e |- ( ph -> B e. NN ) $.
+    flt4lem2.c $e |- ( ph -> C e. NN ) $.
+    flt4lem2.1 $e |- ( ph -> A < B ) $.
+    $( ~ pythagtrip in deduction form with the additional hypothesis ` A < B `
+       to separate ` { A , B } = { x , y } ` into ` A = x /\ B = y ` .  This
+       hypothesis can be assumed without loss of generality because of ~ fltne
+       and the commutative property of addition.  (Contributed by SN,
```
benjub, line 647871:
```suggestion
       and commutativity of addition.  (Contributed by SN,
```

icecream17:
FLT3 seems several times as hard as FLT4, judging by https://en.wikipedia.org/wiki/Proof_of_Fermat%27s_Last_Theorem_for_specific_exponents#n_=_3 (though it uses a "crucial lemma" without proving it)

and Lean, which has tactics and probably a better library given its cyclotomic Three import, and which takes 600 lines to prove FLT3
https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/NumberTheory/FLT/Three.lean

<details><summary>worse options I looked at</summary>

I tried to find if anyone found a simple proof, but to no avail.

---

https://www.researchgate.net/publication/269238745_Proof_of_Fermat's_Last_Theorem_for_n3_Using_Tschirnhaus_Transformation 
has many steps introducing very complicated substitutions and transformations, ending with
image
then
image

Which I find rather amusing. Other than the slightly questionable last sentence, it does not seem wrong, but it also does not seem to have garnered any citations.

---

Strangely, google keeps shoving the one vixra paper, which obviously doesn't work (edit: ironic typo). But there's also this arxiv paper: https://arxiv.org/abs/math/0309005 which is wrong because the logic at the end of page 6 doesn't connect (I can't tell what "embedded" means, based on the second last paragraph on page 6 it means two groups of equations can be true at the same time. That would make page 6 last paragraph sentence 2 equivalent to FLT, however the proof of that sentence is not given).

Afaict no one has bothered to refute this paper until now, probably because the grammar + the claim was a red flag.

</details>

metakunt:
What do you mean by questionable? The proof appears to be short to be formalised.
Is there any step that is likely a non-sequitur?

icecream17:
I was referring to "...and thereby this proof suggests that there is a possibility of existence of a simple proof to the Fermat's Last Theorem in general", there doesn't seem to be anything wrong with the actual proof.

set.mm
benjub: I see you added the associated lemma `fltdvdsabdvdsc`: perfect. Resolving the conversation.

set.mm
```diff
@@ -647464,12 +647542,13 @@ standardize vectors to a length (norm) of one, but such definitions require
 
     prjspnenm1.f $e |- F = ??? $.
     prjspnenm1.g $e |- G = ??? $.
-    @( A bijection between an n-dimensional projective space and its
-       (n-1)-dimensional affine and projective spaces.  (Contributed by Steven
-       Nguyen, ??-??-2023.) @)
+    @( The canonical bijection between an n-dimensional projective K-space and
```
benjub, line 647545: I would change "between ... and" to "from ... onto".

metakunt:
So formalising FLT for n=3 is in reach? I think we might even have Cardano in set.mm

set.mm
```diff
@@ -647669,29 +647787,355 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+    $d ph a c $.  $d ph b $.  $d A a $.  $d A c $.  $d B a $.  $d B b c $.
+    $d C c $.  $d a b d e f $.  $d c d e f $.  $d e g h $.   $d e f x z $.
+    $d ph d g h x z $.
+    flt4g.a $e |- ( ph -> A e. NN ) $.
+    flt4g.b $e |- ( ph -> B e. NN ) $.
+    flt4g.c $e |- ( ph -> C e. NN ) $.
+    $( Fermat's last theorem for the exponent four, generalized.  This is
+       equivalent to Fermat's right triangle theorem.  (Contributed by SN,
+       20-Aug-2024.) $)
+    flt4g $p |- ( ph -> ( ( A ^ 4 ) + ( B ^ 2 ) ) =/= ( C ^ 4 ) ) $=
```
avekens, line 648069: Is it legitimate to call this a "generalization of Fermat's last theorem for the exponent four"?

avekens (commented):
Very interesting (intermediate) results, and nice generalizations of existing theorems. Please go ahead.

icecream17:
More delays: I just rediscovered my notes which [links to a proof that's even better than wikipedia's](https://crypto.stanford.edu/pbc/notes/numberfield/fermatn4.html)

set.mm
icecream17: Yes, by taking the contrapositive of $A^4 + B^4 = C^4 \implies A^4 + (B^2)^2 = C^4$

(Note: New proof, this will become $A^4 + B^4 = C^2$ instead)

benjub: It is indeed stronger, but "generalization" is a bit far-fetched. Maybe:
```
$( Fermat's right triangle theorem, which implies Fermat's last theorem for exponent 4.
   (Contributed by SN, 20-Aug-2024.) $)
```
And the "new proof" should actually be another theorem: both theorems should be here.  The other comment can be
```
$( Strengthening of Fermat's last theorem for exponent 4, where the sum is
   only assumed to be a square.  (Contributed by SN, 20-Aug-2024.) $)
```

benjub:
General note: for these theorems, I think the following statements look nicer:
```
( ph -> A e. ZZ) 
...
( ph -> ( ( A ^ 4  ) + ( B ^ 4 ) ) =  ( C ^ 2 ) )
=>
( ph -> ( A x B ) =  0 )
```
And when the exponent is odd, the conclusion is `( ph -> ( ( A x B ) x C ) =  0 )` (or associated the other way, I do not know which we prefer in set.mm)

icecream17:
Almost done, except for proving a lemma, ~~flt4g modified~~, and ~~[flt4g (new label)](https://github.com/metamath/set.mm/pull/4149#discussion_r1727701427)~~

Here's what the new a1ii looks like. I converted the text into a list, used `<code>` for commands, and added the sentence below it (I highlighted the first word).
image

set.mm
```diff
@@ -474,18 +474,24 @@ The Metamath (Metamath-exe) program Proof Assistant requires proofs to
...
+       a1ii.2.</li><li>After the independent subproof is complete, use
+       <code>MM-PA> IMPROVE ALL</code> to assign it automatically to an unknown
+       step in the main proof that matches it.</li><li>After the entire proof
+       is complete, use <code>MM-PA> MINIMIZE_WITH *</code> to clean up
+       (discard) all ~ a1ii references automatically.</ol></HTML>
+
+       This can also be used to apply subproofs from other theorems.  In step
+       2, simply assign the theorem to a1ii.2, and run
+       <HTML><code>MM-PA> EXPAND <theorem></code></HTML> to "import" a subproof
+       from another theorem.
```
benjub, line 494: This looks prettier indeed, and this change is good to me.  As a loose principle, I think there are two things to be wary of:
* do not spend too much time on rich formatting of comments since the further we depart from plain text, the harder the database is to maintain (but of course, when necessary for improving understanding, do use available functionalities);
* if there are general instructions or remarks about "proof engineering" which are not specific to set.mm, they probably have a better place to be that in set.mm.

Again, I think that in the current case this is ok (even though the method with using a1ii is generalizable to various sorts of databases, and these general guidelines could be added somewhere more general).

set.mm
```diff
@@ -184668,7 +184747,8 @@ their least common multiple ( ~ df-lcmf ) are provided.  Both definitions are
 
     $( Two positive integers are not coprime, i.e. there is an integer greater
        than 1 which divides both integers, iff their greatest common divisor is
-       not 1.  (Contributed by AV, 9-Aug-2020.) $)
+       not 1.  See ~ ncoprmgcdne1bgd for a more general version.  (Contributed
+       by AV, 9-Aug-2020.) $)
     ncoprmgcdne1b $p |- ( ( A e. NN /\ B e. NN )
                           -> ( E. i e. ( ZZ>= ` 2 ) ( i || A /\ i || B )
                                <-> ( A gcd B ) =/= 1 ) ) $=
```
benjub, line 184754: True for any $A, B \in \mathbb{Z}$.

set.mm
```diff
@@ -645565,9 +645707,21 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+      IJZHABMNCMNDONTUAPEFGBCDQRS $.
+  $}
+
   $( If ` A ` and ` B ` are relatively prime, then so are ` A ^ N ` and
-     ` B ^ N ` . ~ rppwr extended to nonnegative integers.  (Contributed by
-     Steven Nguyen, 4-Apr-2023.) $)
+     ` B ^ N ` . ~ rppwr extended to nonnegative integers.  Less general than
+     ~ rpexp12i .  (Contributed by Steven Nguyen, 4-Apr-2023.) $)
   nn0rppwr $p |- ( ( A e. NN0 /\ B e. NN0 /\ N e. NN0 ) ->
                     ( ( A gcd B ) = 1 -> ( ( A ^ N ) gcd ( B ^ N ) ) = 1 ) ) $=
```
benjub, line 645726: True for any $A, B \in \mathbb{Z}$.

set.mm
```diff
@@ -645698,6 +645854,20 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
       USUTRACUOSTT $.
   $}
 
+  $( ~ dvdssqlem generalized to positive integer exponents.  (Contributed by
+     SN, 20-Aug-2024.) $)
+  dvdsexpnn $p |- ( ( A e. NN /\ B e. NN /\ N e. NN ) ->
+                                    ( A || B <-> ( A ^ N ) || ( B ^ N ) ) ) $=
```
benjub, line 645860: True for any $A, B \in \mathbb{Z}$.

set.mm
```diff
@@ -186071,6 +186166,27 @@ being prime ( ` Prime = { p e. NN | ... ` ), but even if ` p e. NN0 ` was
     cfv syl2anr wi prmfac1 3exp impcom con3d sylbid ) ACDZBEDZFZBAGHZABIHZJZABK
     SLHZJUHBMDAMDUJULNUGBOUGAAPQBARTUIUMUKUHUGUMUKUAUHUGUMUKABUBUCUDUEUF $.
 
+  ${
+    $d ph i p $.  $d A i p $.  $d B i p $.
+    ncoprmgcdne1bgd.a $e |- ( ph -> A e. NN ) $.
+    ncoprmgcdne1bgd.b $e |- ( ph -> B e. NN ) $.
+    $( Two positive integers are not coprime iff a prime divides both integers.
+       General, deduction version of ~ ncoprmgcdne1b .  (Contributed by SN,
```
benjub, line 186174: Neither more nor less general.

True for $A, B \in \mathbb{Z}$.

set.mm
```diff
@@ -645322,6 +645405,65 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+    mvrlsubd $p |- ( ph -> ( A + C ) = B ) $=
+      ( caddc co cmin cc subcld eqeltrd addcld pncand eqtrd subcan2d ) ABDHIZCD
+      ABDABCDJIZKGACDEFLMZFNEFARDJIBSABDTFOGPQ $.
+  $}
+
+  ${
+    laddrotrd.a $e |- ( ph -> A e. CC ) $.
+    laddrotrd.b $e |- ( ph -> B e. CC ) $.
+    laddrotrd.1 $e |- ( ph -> ( A + B ) = C ) $.
+    $( Rotate the variables right in an equation with addition on the left,
```
benjub, line 645423: I find "rotate" a bit strange. Why not:
> Move a left-addition in the LHS to a subtraction in the RHS, and then use symmetry of equality.

set.mm
```diff
@@ -645322,6 +645405,65 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
     $( $j usage 'nnmulcom' avoids 'ax-mulcom'; $)
   $}
 
+  ${
+    mvrlsubd.a $e |- ( ph -> B e. CC ) $.
+    mvrlsubd.b $e |- ( ph -> C e. CC ) $.
+    mvrlsubd.1 $e |- ( ph -> A = ( B - C ) ) $.
+    $( Move RHS left subtraction to LHS.  Converse of ~ mvlraddd .
```
benjub, line 645412: Or maybe:
> Move a subtraction in the RHS to a right-addition in the LHS.

set.mm
```diff
@@ -645322,6 +645405,65 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
...
+       by SN, 21-Aug-2024.) $)
+    laddrotrd $p |- ( ph -> ( C - A ) = B ) $=
+      ( cmin co mvlladdd eqcomd ) ACDBHIABCDEFGJK $.
+  $}
+
+  ${
+    raddcom12d.b $e |- ( ph -> B e. CC ) $.
+    raddcom12d.c $e |- ( ph -> C e. CC ) $.
+    raddcom12d.1 $e |- ( ph -> A = ( B + C ) ) $.
+    $( Swap the first two variables in an equation with addition on the right,
```
benjub, line 645435: Same kind of things: it seems strange to "swap" the whole LHS with a part of the RHS.
> Move a right-addition to in the RHS to a subtraction in the LHS, and then use symmetry of equality.

Same for the following results.

set.mm
```diff
@@ -645688,8 +645842,10 @@ number axioms (add ~ ax-10 , ~ ax-11 , ~ ax-13 , ~ ax-nul , and remove
     exp11d.3 $e |- ( ph -> N e. ZZ ) $.
     exp11d.4 $e |- ( ph -> N =/= 0 ) $.
     exp11d.5 $e |- ( ph -> ( A ^ N ) = ( B ^ N ) ) $.
-    $( ~ sq11d for positive real bases and nonzero exponents.  (Contributed by
-       Steven Nguyen, 6-Apr-2023.) $)
+    $( ~ sq11d for positive real bases and nonzero exponents.  The base cannot
+       be generalized much further, since if ` N ` is even then we have
+       ` A ^ N = -u A ^ N ` .  TODO-SN: Avoid ~ df-cxp .  (Contributed by SN,
+       6-Apr-2023.) $)
```
benjub, line 645848: What **can** be generalized to complex bases, on the other hand, is the conclusion of equality of the absolute values.  If you are interested, make yourself a TODO!

set.mm
```diff
@@ -647669,29 +647853,497 @@ standardize vectors to a length (norm) of one, but such definitions require
...
+  ${
+    fltdvdsabdvdsc.a $e |- ( ph -> A e. NN ) $.
+    fltdvdsabdvdsc.b $e |- ( ph -> B e. NN ) $.
+    fltdvdsabdvdsc.c $e |- ( ph -> C e. NN ) $.
+    fltdvdsabdvdsc.n $e |- ( ph -> N e. NN ) $.
+    fltdvdsabdvdsc.1 $e |- ( ph -> ( ( A ^ N ) + ( B ^ N ) ) = ( C ^ N ) ) $.
+    $( Any factor of both ` A ` and ` B ` also divides ` C ` .  This
+       establishes the validity of ~ fltabcoprmex .  (Contributed by SN,
+       21-Aug-2024.) $)
+    fltdvdsabdvdsc $p |- ( ph -> ( A gcd B ) || C ) $=
```
benjub, line 647920: As you proved in the theorem above, N can be in NN0. Also, C can be in ZZ, and maybe A and B too.

It would certainly be nice to be as general as possible, but it may be a bit annoying in the proofs. Hopefully the generalizatio of the earlier results to ZZ may reduce the needs to treat zero separately.

The same questions will appear repeatedly below. Since it is your mathbox, these are non-blocking remarks.

set.mm
benjub: I see you did that. Thanks. Resolving.

icecream17:
Note: The force push only changes the commit message to link to a more relevant pr

set.mm
```diff
@@ -184668,7 +184747,8 @@ their least common multiple ( ~ df-lcmf ) are provided.  Both definitions are
 
     $( Two positive integers are not coprime, i.e. there is an integer greater
        than 1 which divides both integers, iff their greatest common divisor is
-       not 1.  (Contributed by AV, 9-Aug-2020.) $)
+       not 1.  See ~ ncoprmgcdne1bgd for a more general version.  (Contributed
+       by AV, 9-Aug-2020.) $)
```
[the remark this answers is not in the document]
icecream17: I'm a bit confused by that, but I reworded it (and changed the label)

set.mm
icecream17: I'll leave this (also) to a later pr since a few parent theorems would also need to be generalized

set.mm
icecream17: the comment should now mention https://us.metamath.org/mpeuni/rpexp12i.html

set.mm
tirix: Maybe it's possible (easier) to first keep the proof in `NN` as a kind of lemma, and then extend to `ZZ` in a subsequent proof?

set.mm
```diff
@@ -178844,20 +178863,36 @@ infinite descent (here implemented by strong induction).  This is
...
-      dvds2subd $p |- ( ph -> K || ( M - N ) ) $=
-        ( cdvds wbr cmin co cz wcel wa wi dvds2sub syl3anc mp2and ) ABCJKZBDJKZ
-        BCDLMJKZFGABNOCNODNOUAUBPUCQEHIBCDRST $.
-    $}
+  ${
+    dvds2addd.1 $e |- ( ph -> K e. ZZ ) $.
+    dvds2addd.2 $e |- ( ph -> K || M ) $.
+    dvds2addd.3 $e |- ( ph -> K || N ) $.
+    dvds2addd.4 $e |- ( ph -> M e. ZZ ) $.
+    dvds2addd.5 $e |- ( ph -> N e. ZZ ) $.
```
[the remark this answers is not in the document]
tirix: Those are even actually redundant by [`dvdszrcl`](https://us.metamath.org/mpeuni/dvdszrcl.html)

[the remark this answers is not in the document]
benjub: @tirix do you think we could discourage that theorem ? For instance as an "implementation detail", or "too fragile" (e.g., if in the future we want to extend it, for instance as in https://us.metamath.org/mpeuni/df-bj-divc.html).

In every textbook, similar statements about divisibility begin with "Let K, M, N be integers".

set.mm
benjub: Oh, sorry. So why add this theorem ? Isn't rpexp12i enough ?

set.mm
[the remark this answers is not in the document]
benjub: The two theorems have the form ph -> (E. x e. A ps <-> ch) and ph -> (E. x e. B ps <-> ch), with say, $A \subsetneq B$. Then, the forward implication of the second is stronger than the forward implication of the first, but for the reverse implications, it's the opposite.

set.mm
benjub: @tirix It may be. I cannot tell from here: it really depends on whether zero has to be treated separately in specific subproofs, or if there is no need for it because the above lemmas have been properly generalized.

set.mm
icecream17: I didn't notice rpexp12i when I made nn0rppwr. Also nn0rppwr doesn't depend on ~ df-prime I guess, so it and the theorems that use it could be placed earlier

set.mm
icecream17: It's easier for my brain to process it as a "rotation" rather than some two step combination (the comment notes two such combinations)

The purpose of consistency is easier understanding and use, so I realize it's a slippery trade off

I could add aliases `mvlladdcomd` etc. as an experiment? (No "usage is discouraged" since that would obviate the experiment, though)

set.mm
icecream17: (from my remark on `fltdvdsabdvdsc` below, this generalization will be left for another pr)

icecream17:
I think this is equivalent to a non-draft state, although https://github.com/metamath/set.mm/pull/4149#discussion_r1730335610 isn't resolved yet

set.mm
benjub: Mathbox, so I won't insist, though I find it very strange to "rotate" across an equality sign (at least in the present case).

benjub (approved):
Good for me, even if I have one labeling remark (see "rotate" above), and I hope that in the future, as many of these statements as possible will have more general and natural hypotheses, typically, A, B, C \in \ZZ (and exponent in \NN0 when the statement remains true without complicating it, or else \NN), and expressing triviality of solutions by ABC=0.  That is, statements of the form:
> All solutions in $\mathbb{Z}^d$ to [this or that Diophantine equation] are trivial.

rather than
> [This or that Diophantine equation] has no solution in [this or that subset of $\mathbb{Z}^d$].

set.mm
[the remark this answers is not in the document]
tirix: > @tirix do you think we could discourage that theorem ?

Oh, I see. Yes, why not ?

benjub:
I gave my approval above (and hopefully my comment to the approval will be useful). There are two other reviewers, @tirix and @avekens. Good for you? Feel free to merge if so.

icecream17:
if you are referring to me, I do not have merge permissions

benjub:
> if you are referring to me, I do not have merge permissions

I was referring to Thierry and Alexander (the requested reviewers are displayed at the top of the right pane on this page).

tirix (approved):
No objections from me!

